PR #1105, opened 2020-01-25
Type fixes

fbs:
Opening this early in case someone  wants to review it  before I do the codegen changes / commit message cleanup. 

Still need to fix struct types, those are a bit more involved, might leave those for a later PR.

PR #1104, opened 2020-01-25
Use a load instruction to access a context field [v2]

mmisono:
This is a follow up of #988.

The previous approach is recursively traversing the AST to see if the Node accesses a context field. But it turns out that this approach has several problems. This PR, instead, introduces Type::ctx to propagate context information in a top-down way (idea by ajor). Type::ctx is basically the same as Type::cast. codegen emit load instructions in FieldAccess and ArrayAccess if the type is Type::ctx.

Please see each commit message for the details. I think all the previously discussed points are solved. 

One thing not previously discussed is the availability of  "ctx" in tracepoint. In this PR, "ctx" is concealed in tracepoint for simplicity. 

Closes #945

PR #1066, opened 2020-01-06
Implement while

fbs:
With the loop supported added in the 5.0 kernel we're now able to
implement C-style while and for loops. Allowing us to write code like:

```
i:s:1 {
        $i = 0;
        while ($i < 100) {
                if ( ($i/10) * 10 == $i ) {
                        printf("\n");
                }
                if ( ($i/6)*6 == $i ){
                        printf("XX ");
                        $i+=1;
                        continue;
                }
                printf("%2d ", $i);
                $i+=1;
                if ($i >= 50) {
                        break;
                }
        }
}
```

To generate:

```
Attaching 1 probe...

XX  1  2  3  4  5 XX  7  8  9
10 11 XX 13 14 15 16 17 XX 19
20 21 22 23 XX 25 26 27 28 29
XX 31 32 33 34 35 XX 37 38 39
40 41 XX 43 44 45 46 47 XX 49
```

Using only a fewinstructions:

```
203: perf_event  name 1  tag 4881cacfe5abad41  gpl
        loaded_at 2020-03-26T18:17:17+0000  uid 0
        xlated 448B  jited 340B  memlock 4096B  map_ids 85
```

Note that LLVM tries to unroll quite aggressively causing a loop the
following be fully unrolled:

```
i:s:1 { $i = 0; while ($i < 30) { @=$i; $i++ } }
```

PR #871, opened 2019-08-26
RFC: Add xlock increase code for count function

olsajiri:
hi,
I tried the global data together with locking increase on count
function and it seems to be working. I think there're probably
other areas this could be put in.

The code still needs some testing/changes, but would be great
to get some input before I dive in more.

thanks,
jirka

---
Changing the count generation code to use eBPF global
data and locked increase to speedup the count function.

Having following code:

```
  # cat count.bt
  tracepoint:syscalls:sys_enter_write
  {
        @write = count();
  }

```
We generate following code even for single value increase:

```
   0: (b7) r1 = 0
   1: (7b) *(u64 *)(r10 -16) = r1
   2: (18) r1 = map[id:536]
   4: (bf) r2 = r10
   5: (07) r2 += -16
   6: (85) call htab_percpu_map_lookup_elem#100624
   7: (b7) r1 = 1
   8: (15) if r0 == 0x0 goto pc+2
   9: (79) r1 = *(u64 *)(r0 +0)
  10: (07) r1 += 1
  11: (7b) *(u64 *)(r10 -8) = r1
  12: (18) r1 = map[id:536]
  14: (bf) r2 = r10
  15: (07) r2 += -16
  16: (bf) r3 = r10
  17: (07) r3 += -8
  18: (b7) r4 = 0
  19: (85) call htab_percpu_map_update_elem#102848
  20: (b7) r0 = 0
  21: (95) exit
```
By adding the possibility to operate directly on the map's
data and locked increase we can use just this code:

```
   0: (18) r1 = map[id:542][0]+0
   2: (b7) r2 = 1
   3: (db) lock *(u64 *)(r1 +0) += r2
   4: (b7) r0 = 0
   5: (95) exit
```
This of course speeds up the processing, without bpftrace involved:

```
  # perf bench sched messaging -l 100000
  # Running 'sched/messaging' benchmark:
  # 20 sender and receiver processes per group
  # 10 groups == 400 processes run

       Total time: 58.194 [sec]
```

With current code:

```
  # bpftrace count.bt -c 'perf bench sched messaging -l 100000'
  Attaching 1 probe...
  # Running 'sched/messaging' benchmark:
  # 20 sender and receiver processes per group
  # 10 groups == 400 processes run

       Total time: 180.127 [sec]

  @write: 400000726
```

With change applied:

```
  # bpftrace count.bt -c 'perf bench sched messaging -l 100000'
  Attaching 1 probe...
  # Running 'sched/messaging' benchmark:
[...]

PR #1108, opened 2020-01-27
Remove version dependence from codegen

fbs:
sorry for the noisy PR. All the magic happens in `codegen/llvm/common.h`

----

The codegen tests are currently quite tedious to update as the "source"
and "result" are defined in the same file which makes it hard to
automatically update.

This commit changes that by:

- Moving the "result" to its own file, `TEST_NAME.ll`
- Introducing a rewriter that tries to automatically fix the LLVM
  version differences
- As rewriting doesn't always work it supports loading results from
  `TEST_NAME_LLVM-X.ll` these will have to be maintained manually but
  at the moment there are only a few (only LLVM 5)


fbs:
This  hopefully makes  adding llvm-9 codegen tests a  lot easier too. I will give that a  try  once (if) this gets merged

tests/codegen/common.h
```diff
@@ -18,17 +18,151 @@ namespace bpftrace {
...
+  // clang-format on
+  auto parts = split_string(line, ' ');
+  std::stringstream buf;
+  std::string alignment = "";
+  for (std::vector<std::string>::iterator it = parts.begin(); it != parts.end();
+       it++)
+  {
+    if (*it == "align")
+    {
+      if (alignment == "")
```
fbs: This is a bit of a hack as we have a few tests with a funky address space, so only the value after the first align is valid:

```
call void @llvm.memcpy.p0i8.p64i8.i64(i8* nonnull align 1 %1, i8 addrspace(64)* align 536870912 null, i64 32, i1 false)
```

tests/codegen/common.h
```diff
@@ -18,17 +18,151 @@ namespace bpftrace {
...
+{
+#if LLVM_VERSION_MAJOR < 7
+  // clang-format off
+  //FROM: call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull %4, i8* nonnull align 1 i32 1, %lookup_elem, i64 32, i1 false)
+  //TO:  call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull %4, i8* nonnull %lookup_elem, i64 32, i32 1, i1 false)
+  // clang-format on
+  auto parts = split_string(line, ' ');
+  std::stringstream buf;
+  std::string alignment = "";
+  for (std::vector<std::string>::iterator it = parts.begin(); it != parts.end();
```
danobi:
```suggestion
  for (auto it = parts.begin(); it != parts.end();
```

tests/codegen/common.h
```diff
@@ -18,17 +18,151 @@ namespace bpftrace {
...
+static std::string rewrite_memcpy_call(const std::string &line)
+{
+#if LLVM_VERSION_MAJOR < 7
+  // clang-format off
+  //FROM: call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull %4, i8* nonnull align 1 i32 1, %lookup_elem, i64 32, i1 false)
+  //TO:  call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull %4, i8* nonnull %lookup_elem, i64 32, i32 1, i1 false)
+  // clang-format on
+  auto parts = split_string(line, ' ');
+  std::stringstream buf;
+  std::string alignment = "";
```
danobi: Default initializes to empty
```suggestion
  std::string alignment;
```

tests/codegen/common.h
danobi: Maybe add a comment mentioning it

danobi: Nit
```suggestion
      if (alignment.empty())
```

tests/codegen/common.h
```diff
@@ -18,17 +18,151 @@ namespace bpftrace {
...
+  parts.insert(parts.end() - 2, std::string("i32 ") + alignment + ",");
+
+  buf << parts.at(0);
+  for (unsigned int i = 1; i < parts.size(); i++)
+  {
+    buf << " " << parts.at(i);
+  }
+  return buf.str();
+#else
+  return std::string(line);
```
danobi:
```suggestion
  return line;
```

tests/codegen/common.h
```diff
@@ -18,17 +18,151 @@ namespace bpftrace {
...
+{
+#if LLVM_VERSION_MAJOR < 7
+  if (line.find("addrspace") != std::string::npos)
+    return std::string(
+        "declare void @llvm.memcpy.p0i8.p64i8.i64(i8* nocapture writeonly, i8 "
+        "addrspace(64)* nocapture readonly, i64, i32, i1) #1");
+  return std::string("declare void @llvm.memcpy.p0i8.p0i8.i64(i8* nocapture "
+                     "writeonly, i8* nocapture readonly, i64, i32, i1) #1");
+#else
+  return std::string(line);
```
danobi:
```suggestion
  return line;
```

tests/codegen/common.h
```diff
@@ -18,17 +18,151 @@ namespace bpftrace {
...
+#endif
+}
+
+static std::string rewrite(const std::string &ir)
+{
+  std::stringstream buf;
+  for (auto &line : split_string(ir, '\n'))
```
danobi:
```suggestion
  for (const auto &line : split_string(ir, '\n'))
```

fbs:
@danobi  thanks for the review. I've updated it to have the test runner update the codegen when `BPFTRACE_UPDATE_TESTS` is set to `1` in the env.  Still not the best user experience when updating tests but it should be a lot better than is it now

danobi:
Oh, also please add documentation. Maybe in the README.md in tests/

fbs:
@danobi / @ajor  I've added docs and added  LLVM9 support (which turned out quite trivial). Can you take a look at last 2 commits? 

https://github.com/iovisor/bpftrace/pull/1108/commits/6036494814b6a6bbd3fc3eb4474f0f763f87c33f

fbs:
@mmisono  any objection to me merging this or would you rather wait until #1104 is merged as it impacts  your work? It's blocking #1105 and testing on fedora >=31 systems.

mmisono:
@fbs
I'm glad if #1104 is merged first, but not sure how long it takes, so please go ahead. This and #1105 are important changes and I also want them to be merged.

By the way, I think it is useful if bpftrace has an option to save LLVM IR to a file like [this --elf option](https://github.com/iovisor/bpftrace/pull/871/commits/c97ce70aada26e63b1a49fe2c36c2475b83).

fbs:
> @fbs
> I'm glad if #1104 is merged first, but not sure how long it takes, so please go ahead. This and #1105 are important changes and I also want them to be merged.

I'll hold it off for the weekend then :)

> By the way, I think it is useful if bpftrace has an option to save LLVM IR to a file like [this --elf option](https://github.com/iovisor/bpftrace/pull/871/commits/c97ce70aada26e63b1a49fe2c36c2475b83).

I had something like this in #1066 but there is some discussion still about how to name the flags.  I got a WIP somewhere that adds `--emit-<ir,opt-ir,elf> [filename]` somewhere, i'll try and clean that up and open it as a rfc for discussion :)

PR #1111, opened 2020-01-27
Support "." in attach point function argument

### LINKED ISSUE #548: functions with dots are not accepted
There are several kernel functions that include dots in their names. I assume the linker renamed / splitted them, but I'm not 100% sure (OT: if someone could point me to why this happens, I will appreciate it, my search-fu failed me 😢)

Aaand `bpftrace` fails when trying to attach a kprobe to them, but `trace.py` seems to be happy with them (don't have example of the errors in this machine, sorry! but IIRC it fails in the lexing).

Anyone against adding support for them? One one hand we are listing them `bpftrace -l 'kprobe*' | grep '\.'` and other tracing tools seem to work ok with them, but on the other hand not sure if this makes sense? 😄

danobi:
golang likes to use "." a lot in symbol names. Before, you would have to
quote your function, ie

    uprobe:/my/go/app:"runtime.main.func1" { 1 }

but that's a bit ugly.

Now you can do:

    uprobe:/my/go/app:runtime.main.func1 { 1 }

This closes #548 .

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -187,6 +187,7 @@ wildcard : wildcard ident    { $$ = $1 + $2; }
          | wildcard MUL      { $$ = $1 + "*"; }
          | wildcard LBRACKET { $$ = $1 + "["; }
          | wildcard RBRACKET { $$ = $1 + "]"; }
+         | wildcard DOT      { $$ = $1 + "."; }
          |                   { $$ = ""; }
          ;
 
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -959,6 +959,14 @@ TEST(Parser, wildcard_path)
        "  int: 1\n");
 }
 
+TEST(Parser, dot_in_func)
+{
+  test("uprobe:/my/go/program:runtime.main.func1 { 1; }",
+       "Program\n"
+       " uprobe:/my/go/program:runtime.main.func1\n"
+       "  int: 1\n");
+}
+
 TEST(Parser, wildcard_func)
 {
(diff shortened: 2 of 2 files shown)


danobi:
1 line fix, PR 1111 :)

brendangregg:
Looks much better, thanks.

It's a little weird, though, or maybe that's just me. In lexel.l we have:

```
ident   [_a-zA-Z][_a-zA-Z0-9]*
```

does adding a dot there break other uses of dot? And the workaround is to add DOT to wildcard, even though it isn't a wildcard? Might need a comment to say "wildcard DOT is not a wildcard, it's for matching golang functions" or something like that.

danobi:
It's because all uprobes that don't have a quoted function match this rule in the parser:
```
| ident PATH wildcard
```

It should instead probably be something like this:
```
| ident PATH ident
| ident PATH wildcard
| ident PATH ident_with_dot
```

(or something like that).

I'm going to try and clean up the attach point parser code in some follow up PRs.
I'm hoping we can rip out all the wildcard bits and parse later in our own C++ code.
Bison isn't really great at dealing with all our variations of "something:something:..."
(all the colons).

danobi:
> does adding a dot there break other uses of dot

I suspect if we added `.` to `ident` a bunch of stuff would parse wrong.
But I have not tested it.

brendangregg:
Ok, thanks; I'm fine with this PR for now.

danobi:
#1117 is the follow up work

Issue #687, opened 2019-05-28
Automatically mangle c++ symbols


danobi:
Sometimes when I uprobe c++ binaries, it's easier to directly type the c++ name I want to trace, eg `Foo::bar`. Does it make sense to let uprobe support something like:
```
probe:/usr/local/bin/dynolog:*folly::EventBase::run*                                                                                                                                                                                                                                      
{                                                                                                                                                                                                                                                                                         
....                                                                                                                                                                                                                                                          
}                    
```

jgkamat:
This would be a fantastic feature, but I'm worried how this would work with probes on other languages, that might demangle differently. If there was a way to tell bpftrace that this name is a c++ symbol, perhaps then demangling can occur automatically.

ajor:
Which languages might cause problems here? Just want to know what we're working with

jgkamat:
Alastair Robertson writes:

> Which languages might cause problems here? Just want to know what we're 
> working with

The languages that I know might need mangling are cpp, D, ocaml, rust, 
haskell, and swift. I think cpp is the one that's the most important though, 
we can probably mostly ignore the others. It's also possible that most 
languages use the same demangling strategy, but I'm not sure.

danobi:
@jgkamat I think we can add more language support as necessary. For example when attaching a uprobe, we could:
```
foreach mangling_strategy:
    n = mangle(symbol)
    if canAttach(n):
        attach(n)
        break
```

Assuming each mangling stratey is unique (needs more research), this feature could be forwards and backwards compatible.

ajor:
Attempting different demangling strategies seems like it would work, but maybe we'd also want to add support for more C++ features at some point, such as the `this` pointer (and other features for other languages too). In that case I wouldn't want to pollute the default builtin namespace with language-specific features.

Following up on @jgkamat's idea to tell bpftrace we've got a C++ symbol, how about saying we're probing a C++ function, then all these language-specific features could be enabled. Something like:

```
probe:/usr/local/bin/dynolog:c++:"*folly::EventBase::run*"
```

dalehamel:
just ran into this today, kinda sucks to have to find the mangled symbol to attach to. Would be great to attach to the dymangled symbol as seen in the source code

dalehamel:
Also if we list symbols with bpftrace -l, i would expect it to list the demangled symbols - so we should be able to attach in the same format as we list IMO.

ajor:
Done in #1116

PR #1116, opened 2020-01-29
Add support for C++ mangled symbols in uprobe names #687

caringi:
Allow to specify uprobe names using demangled C++ format (this is done
with the help of abi::__cxa_demangle() function).

Example:

$ nm -g a.out |grep foo
0000000000401151 T _Z3fooi
0000000000401146 T _Z3foov

$ nm -g --demangle a.out |grep foo
0000000000401151 T foo(int)
0000000000401146 T foo()

$ bpftrace -e 'uprobe:./a.out:foo {printf("ok\n");}'
Attaching 2 probes...
^C

$ bpftrace -e 'uprobe:./a.out:"foo()" {printf("ok\n");}'
Attaching 1 probe...
^C

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1,6 +1,7 @@
 #include <arpa/inet.h>
 #include <cassert>
 #include <ctime>
+#include <cxxabi.h>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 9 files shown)


danobi:
Haven't looked too closely at the code yet, but I like this feature.

One thought: should we make `bpftrace -l uprobe:/path/to/bin` print demanged names as well? Sometimes it can be tricky to guess what the demangled name will be.

caringi:
> One thought: should we make `bpftrace -l uprobe:/path/to/bin` print demanged names as well? Sometimes it can be tricky to guess what the demangled name will be.

I think that's a good idea...

Maybe `bpftrace -l uprobe:/path/to/bin` could print mangled and demangled names side by side...

Somethiing like:

```
uprobe:./a.out:foo() [_Z3foov]
uprobe:./a.out:foo(int) [_Z3fooi]
```

Another option would be to use `-lv` to print demangled names, and leave `-l` printing the symbols without demangling.

dalehamel:
> One thought: should we make bpftrace -l uprobe:/path/to/bin print demanged names as well? Sometimes it can be tricky to guess what the demangled name will be.

I think the default behavior should be that listing uprobes this way should print demangled by default, as we also want to be able to have the list and attach syntax be consistent. (echoing what I said on #687 )

Thanks for doing this @caringi

src/bpftrace.cpp
```diff
@@ -272,8 +293,27 @@ std::set<std::string> BPFtrace::find_wildcard_matches(
     }
 
     if (!wildcard_match(line, tokens, start_wildcard, end_wildcard))
+    {
+      if (!line.rfind("_Z", 0) || !line.rfind("____Z", 0))
```
ajor: If this check is extracted into a function then we'd be sure it always stays in sync with the same check down in `find_symbol_matches()`. Giving it a descriptive name would also help to make it clear it's trying to match C++ symbols (self documenting!)

caringi:
> Looks good to me. Is it still a RFC or ok to merge?

I've added the RFC prefix to emphasize I needed feedback...

I think it's ok to merge it...

dalehamel:
this is awesome @caringi i think this is ok to merge!

PR #1117, opened 2020-01-29
Refactor attachpoint parser

danobi:
It's getting harder to unambiguously add new probe formats into the
parser. There's a bunch of hacks (see CINT) to get around ambiguous
parsing issues.

It's nicer to write our own parsing logic in C++. This way things
stay extensible and modular so new providers can do custom things if
necessary.


src/ast/attachpoint_parser.cpp
```diff
@@ -0,0 +1,426 @@
...
+    if (offset_parts.size() != 2)
+    {
+      errs_ << "Invalid offset" << std::endl;
+      return 1;
+    }
+
+    ap_->func = offset_parts[0];
+    try
+    {
+      ap_->func_offset = std::stoll(offset_parts[1], nullptr, 0);
```
mmisono: This (and other stoll/stoi) allows trailing non-numeric characters. e.g., `k:func+0x100zzz`.  It can be checked by

```cpp
std::size t idx;
ap_->func_offset = std::stoll(offset_parts[1], &idx, 0);
if (idx != offset_parts[1].size()) { /* err */ } 
```

When I implemented the support of attaching kprobe to offset, it took some time to understand attachpoint parser. This looks nice.

src/ast/attachpoint_parser.cpp
```diff
@@ -0,0 +1,426 @@
...
+  }
+
+  return 0;
+}
+
+int AttachPointParser::kprobe_parser(bool allow_offset)
+{
+  if (parts_.size() != 2)
+  {
+    errs_ << ap_->provider << " provider requires 2 arguments" << std::endl;
```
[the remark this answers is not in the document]
danobi: Fixed for other instances too

tests/runtime/probe
```diff
@@ -10,7 +10,7 @@ TIMEOUT 5
 
 NAME kprobe_target
 RUN bpftrace -v -e 'kprobe:syscalls:sys_exit_nanosleep { printf("SUCCESS '$test' %d\n", pid); exit(); }'
-EXPECT kprobes should not have a target
+EXPECT kprobe provider requires 1 argument
```
ajor: This error message took me a while to understand - I think it can improved a little
1. "requires 1 argument" sounds to me like it's missing an argument, not that it has too many
1. I didn't know what it meant by "argument" (I was looking for something like a function call)

I'm just providing criticism here - no ideas are coming to me for better wording at the moment :P

tests/bpftrace.cpp
```diff
@@ -111,7 +111,8 @@ void check_special_probe(Probe &p, const std::string &attach_point, const std::s
 
 TEST(bpftrace, add_begin_probe)
 {
-  ast::AttachPoint a("BEGIN");
+  ast::AttachPoint a("");
```
ajor: Is it possible to just do this instead of the separate assignment? Just to keep the tests a bit cleaner
```suggestion
  ast::AttachPoint a("BEGIN");
```

src/ast/attachpoint_parser.cpp
```diff
@@ -0,0 +1,501 @@
...
+    case ProbeType::interval:
+      return interval_parser();
+    case ProbeType::software:
+      return software_parser();
+    case ProbeType::hardware:
+      return hardware_parser();
+    case ProbeType::watchpoint:
+      return watchpoint_parser();
+    default:
+      errs_ << "Unrecognized provider: " << ap_->provider << std::endl;
```
ajor:
```suggestion
      errs_ << "Unrecognized probe type: " << ap_->provider << std::endl;
```
and in other error messages

tests/bpftrace.cpp
danobi: I looked into the options and here is a summary of my thoughts:
* I would prefer to avoid parsing outside of the parsing unit tests. The reason being unit tests should be as orthogonal to each other as possible.
* I would also prefer to avoid all the overloaded constructors we had before (b/c it was hard to reason about which overload would get selected)
* Designated initializers looked pretty good, but there's too many language restrictions on them in C++ so I don't think it's possible to use it here (no user declared or inherited constructors allowed)

It seems like this is the best we can do unless you have other thoughts.

tests/runtime/probe
danobi: Yeah, I don't have anything better. I'm hoping the error underlining makes these errors more clear.

danobi:
I've incorporated all the suggestions into the latest push.

> uprobe:/tmp/a.out { 1 }

I fixed that issue with this change:
```diff
diff --git a/src/parser.yy b/src/parser.yy
index 074aa53..2dd8164 100644
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -180,6 +180,7 @@ attach_point_def : attach_point_def ident    { $$ = $1 + $2; }
                  | attach_point_def MUL      { $$ = $1 + "*"; }
                  | attach_point_def LBRACKET { $$ = $1 + "["; }
                  | attach_point_def RBRACKET { $$ = $1 + "]"; }
+                 | attach_point_def DIV      { $$ = $1 + "/"; }
                  |                           { $$ = ""; }
                  ;
```

> uprobe:/tmp/a.out:"A::f" { 1 }

Still working on this one. Something is wrong with how `::`'s are handled. Something probably thinks it's a PATH.

danobi:
> > uprobe:/tmp/a.out { 1 }

> I fixed that issue with this change:

Ok that didn't work. Broke predicate parsing. I'll probably leave that alone for now since it still errors with a reasonable message.

danobi:
I fixed the `uprobe:/tmp/a.out:"A::f" { 1 }` parsing issue.

Ready for review again.

src/ast/attachpoint_parser.cpp
```diff
@@ -0,0 +1,547 @@
...
+      if (argument.empty() && remove_empty)
+        continue;
+
+      ret.emplace_back(std::move(argument));
+      // The standard says an std::string in moved-from state is in
+      // valid but unspecified state, so clear() to be safe
+      argument.clear();
+    }
+    else if (c == '"')
+      in_quotes = !in_quotes;
```
mmisono: I found the case an escaped quote is needed. e.g.,

```c
#include <stdio.h>
#include <sys/sdt.h>

int main(){
    int a = 1;
    DTRACE_PROBE1(test, "probe1", a);
    printf("%d\n", a);
    return 0;
}
```
then
```
% readelf -n a.out
[...]
Displaying notes found in: .note.stapsdt
  stapsdt              0x00000032       NT_STAPSDT (SystemTap probe descriptors)
    Provider: test
    Name: "probe1"
    Location: 0x0000000000000659, Base: 0x0000000000000708, Semaphore: 0x0000000000000000
    Arguments: -4@-4(%rbp)
```
```sh
% sudo ./src/bpftrace -e 'usdt:./a.out:test:"\"probe1\"" { printf("%d\n", arg0); }'
couldn't get argument 0 for ./a.out:test:probe1
```
This works on master. I'm not sure but how about split attachpoint in parser?

mmisono: In this case, the string to parse is `usdt:./a.out:test:""probe1""`, and the follwoing is `usdt:./a.out:test:""pro:be1""`.
 
```sh
% sudo ./src/bpftrace -e 'usdt:./a.out:test:"\"pro:be1\"" { printf("%d\n", arg0); }'
stdin:1:1-32: ERROR: usdt probe type requires 2 or 3 arguments
usdt:./a.out:test:"\"pro:be1\"" { printf("%d\n", arg0); }
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Maybe `split_attachpoint` should only remove the first and the last quotes.

danobi: Ah nice find.

> Maybe split_attachpoint should only remove the first and the last quotes.

Hm, that feels a bit brittle. I think lexers typically walk one character at a time and can peek/consume forward as necessary. Let me see if that will work here.

danobi:
Fixed escaped double quote handling

mmisono:
I found the another thing... The current parser allows concatenation of ident and string. e.g., the followings are all the same.
- `usdt:./a.out:test:probe`
- `usdt:./a.out:test:"probe"`
- `usdt:./a.out:test:pro"be"`

This differs the current behavior:

```
% sudo ./src/bpftrace -e 'usdt:./a.out:test:pro"be" { printf("%d\n", arg0); }'
stdin:1:1-26: ERROR: syntax error, unexpected string, expecting {
usdt:./a.out:test:pro"be" { printf("%d\n", arg0); }
```

fbs:
> I found the another thing... The current parser allows concatenation of ident and string. e.g., the followings are all the same.

Would be nice to fix that but not critical imo. 

LGTM

danobi:
> I found the another thing... The current parser allows concatenation of ident and string. e.g., the followings are all the same.

I will put up a follow up PR to fix this.

danobi:
@mmisono , I'm looking at the tests and see `"k:\"fn.abc\"+0x10 {}"` being tested. So it appears we do allow mixing quoted and unquoted text, but only if the quoted text is in the beginning of the argument.

I wonder if it would make things more consistent to generally allow mixing quoted and unquoted text.

Supporting either is quite trivial -- the change to disallow quoted text at end looks like this:
```diff
+    else if (raw[idx] == '"')
+    {
+      // Only allow mixing quoted and unquoted text if quoted text is at beginning of arg
+      if (argument.size() && !in_quotes)
+      {
+        errs_ << "Mixing quoted and unquoted text is not allowed" << std::endl;
+        return 1;
+      }
+
+      in_quotes = !in_quotes;
+    }
```

mmisono:
I see. I prefer to disallow mixing quoted and unquoted text other than when using + (I don't come up with other cases that mixing is needed), but I'm OK with the current behavior since it seems harmless and potentially useful in the future.
(by the way, I think the suggested change allow something like `k:"fn.a"bc+0x10 {}`)

mmisono:
I don't think it is critical, but I also noticed that spaces between idents are allowed, e.g., `k pro be : vfs   _open`.

danobi:
Thanks for finding all these issues, @mmisono .

> I also noticed that spaces between idents are allowed, e.g., k pro be : vfs _open

That's interesting. I thought bison wouldn't allow this. Will investigate.

danobi:
So the reason is b/c the lexer ignores whitespace: https://github.com/iovisor/bpftrace/blob/master/src/lexer.l#L44-L45 when generating tokens.

It appears this was a problem before as well. Consider:
```
kprobe: f s d { if(1){ printf("arg0: %d\n", 1) }}
```

I spent the day exploring solutions but I think it'll be difficult to parse this correctly. The problem is a lack of keywords or delimiters the lexer can anchor to. Without something to anchor to, it becomes quite difficult (without keeping more state) to say a sequence of characters should be an attachpoint definition.

mmisono:
Hmm. parsing is not easy... 
@danobi 
I realized that arbitrary extra colons are allowed other than usdt and watchpoint. e.g.
```
% sudo ./src/bpftrace -e '::k::vfs_open:: {}'
Attaching 1 probe...
```
This is because [remove_empty is true](https://github.com/iovisor/bpftrace/blob/11814f2981c2fcf4c0177720910786c62474e100/src/ast/attachpoint_parser.cpp#L34). Is it desirable?

mmisono:
Even if watchpoint, `:w:0x10000000:8:rw` is allowed.

danobi:
So I spent a bunch of time trying to deal with whitespace. Here were the things I tried:
* create an attachpoint token: https://github.com/danobi/bpftrace/commit/0a3ce8b25a1cb4f1
  * IIRC this didn't work b/c it swallowed up ident tokens too
* use global state to figure out when we're parsing attachpoint definitions: https://github.com/danobi/bpftrace/commit/a8c9409ac4aaf
  * this didn't work b/c we were peeking the stream and once you see an EOF you can't put it back anymore. you get segfaults if you try to continue reading the stream.
  * peeking the stream in the lexer is also wrong
* use flex states to create a substate for attachpoint definition: https://github.com/danobi/bpftrace/commit/38c253013fe9
  * this had a lot of promise (and probably was the best attempted solution) but unfortunately doesn't work b/c there's nothing for the lexer to "anchor" onto to figure out it's seeing an attachpoint definition. ie how do you determine if `#include <asdf>` in the beginning of the file is an attachpoint definition? or how would you parse:
```
kprobe:f // comment
{
  1
}
```

danobi:
It seems to me that it's extremely difficult to parse attachpoint definitions in bpftrace correctly. Maybe if I knew more parsing techniques I could figure it out but I'm completely out of ideas now. If probes looked more like:
```
[kprobe:f] / 1 / { 1 }
```
it'd be really easy to parse.

danobi:
@mmisono let's take the discussion to #1202 so it's easier to search up later.

PR #1121, opened 2020-01-31
cmake: use *_LIBRARIES when testing for libbfd version

cyphar:
On openSUSE, [binutils has really strangely named libbfd and libopcodes
libraries][1]. This is fine (we can define *_LIBRARIES variables), but the
result is that the LIBBFD_DISASM_FOUR_ARGS_SIGNATURE test will always
fail because CHECK_CXX_SOURCE_COMPILES cannot find the right libraries.

The fix is simple -- just use *_LIBRARIES for CMAKE_REQUIRED_LIBRARIES.
We need to find_library(LibOpcodes) for that test, but that isn't too
bad because CMake caches the result.

[1]: https://bugzilla.suse.com/show_bug.cgi?id=1162312

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/cmake/FindLibBfd.cmake
+++ b/cmake/FindLibBfd.cmake
@@ -61,7 +61,8 @@ FIND_PACKAGE_HANDLE_STANDARD_ARGS(LibBfd "Please install the libbfd development 
 mark_as_advanced(LIBBFD_INCLUDE_DIRS LIBBFD_LIBRARIES)
 
 if(${LIBBFD_FOUND})
-SET(CMAKE_REQUIRED_LIBRARIES bfd opcodes)
+find_package(LibOpcodes)
+SET(CMAKE_REQUIRED_LIBRARIES ${LIBBFD_LIBRARIES} ${LIBOPCODES_LIBRARIES})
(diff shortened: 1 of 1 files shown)


danobi:
Seems like this is breaking the static builds. Can you figure out what broke it?

cyphar:
The reason why this was initially necessary was to work around [an issue when packaging on openSUSE](https://bugzilla.suse.com/show_bug.cgi?id=1162312) but it turns out that on openSUSE it's actually recommended to build against the static libraries and thus custom library paths aren't necessary.

However, there is still a bug in `bpftrace`'s CMake setup -- you need to link against `-liberty` when linking against `-lbfd`. When linking against shared libraries this requirement is hidden, but if you're linking against static libs you end up with build failures that don't make any sense (cannot resolve symbol "xmalloc").

danobi:
I agree this is better, but I'm baffled why the CI is failing.

I did a little bit of debugging but I can't imagine why changing the linker options affects compilation.

danobi:
I figured out the CI issues in #1166 . Let's move further discussion there

PR #1122, opened 2020-01-31
0.9.4 release

fbs:
Based on commit d3cb095a6ef19388bc33b3d8811d84353be7fff3.

----

As we have almost 200 commits since 0.9.3 I've modified the changelog a bit to keep it readable. Only changes that directly impact the end users are now in the added/changed/fixed section. Build warning and tests fixes are all placed in the internal group.

I've gone through it a few time but with 200 commits its easy to misplace something.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,204 @@ and this project adheres to
 
 ## Unreleased
 
+## [0.9.4] 2020-02-04
+
+### Highlights
+
+  - New calls: `signal`, `override`, `strncmp`
+  - Support for attaching to `kprobes` at an offset
+  - Support for struct bitfields
+
+### All Changes
+
+#### Added
+  - Add support to attach kprobe to offset (e31e398) by Masanori Misono &lt;m.misono760@gmail.com&gt;
+  - Add `--info` flag (afafbf5) by bas smit &lt;bas@baslab.org&gt;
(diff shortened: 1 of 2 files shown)


mmisono:
There are several commits included in 0.9.3: 
```
  - Add sargX builtin
  - Add support to cast to a pointer of integer
  - Add tests for sargX builtin
  - Avoid leaking state between cmake tests
  - Avoid testing for FOUR_ARGS_SIGNATURE on systems without bfd
  - bfd-disasm: fix LIBBFD_DISASM_FOUR_ARGS_SIGNATURE
  - Change "variable.tracepoint arg casts in predicates" runtime test
  - codegen: fix rhs type check for binop
  - Define PACKAGE to make libbfd happy
  - Fix map field access
  - Fix 'may be used uninitialized' build warning in bfd-disasm.cpp
  - Mention sargX builtin in docs
  - semantic_analyser: fix gcc build error on xenial
  - semantic_analyser: validate use of calls as map keys
  - Unset `CMAKE_REQUIRED_LIBRARIES` to avoid influencing tests
  - Use the same shebang for all tools
```
Relevant commits are 56ad6e8 .. dc200b3 (master) and  7d62627 .. 8b60006 (0.9.3_release).
Maybe hash values were changed due to cherry-picking.

fbs:
ah woops. I only checked for duplicate commit ids, not duplicate lines. Thanks :)

PR #1135, opened 2020-02-06
Allow running tests as non-root (again)

danobi:
Commit 76770de39e ("Report kernel instruction limit") broke non-root
tests again. It was crashing non-root tests with:

    bpf: Failed to load program: Operation not permitted

    unknown file: Failure
    C++ exception with description "stoi" thrown in the test body.

This commit gracefully handles errors.


fbs:
Maybe we should restructure bpffeature in a way that avoids doing the feature testing at all during testing?

Issue #1051, opened 2019-12-31
RFC: Release targets


dalehamel:
This is a fork off of the discussion from #342 @danobi 

The idea here is to use the existing CI build for bpftrace to generate various release targets when the release process is initiated by `git tag`. 

These artifacts should be attached to a Release object from the github API transparently. 

It may also make sense to use this process to drive publishing artifacts to other systems, such as:

- Publishing a debian package to an iovisor-hosted apt repo
- Publishing a flatpak or snapcraft image
- Pushing a docker image to quay.io/iovisor/bpftrace or dockerhub

I think that this wolud help to make bpftrace more accessible.

This issue mostly supposes that #1041 makes sense, and bits like authenticating with quay.io/iovisor images would help to speed up CI runs, which is itself a concern of #1041 

# Implementation

Most of this comes down to someone with credentials on the various package publishing system to encrypt them and store them in the travis.yml

Travis itself provides deploy hooks that are probably flexible enough to initiate almost any process.

Most of these artifacts already exist, and deploy steps just need to be configured to copy them or do some packaging post-processing.

# Side benefit - Faster CI builds

If we can get access to push images from previous runs, we only rebuild layers when necessary. Each build type can have their own docker tag based on their ENV vars. This could help drop CI times significantly, depending on the docker pull speed.

dalehamel:
Sample release process for travis: 

#before_deploy:
#  - FIXME copy binaries out of static builds
# deploy:
#   provider: releases
#   api_key:
#     secure: YOUR_API_KEY_ENCRYPTED
#   file:
#     - "FILE 1"
#     - "FILE 2"
#   skip_cleanup: true
#   on:
#     tags: true
#     condition: $RELEASE_BINARY = ON

dalehamel:
@mmarchini @fbs @danobi and i discussed using github actions https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-workflows

The syntax is here https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions and doesn't seem too different than Travis.

I don't have access to set this up for iovisor/bpftrace, but perhaps I can explore it on my own fork and submit a config file analogous to the existing travis file?

Who would have access to set this up / could review the PR?

dalehamel:
I've played around with github actions, they are incredible.

[...]

PR #1139, opened 2020-02-11
Add Dockerfile.release for bpftrace docker image on quay.io

dalehamel:
For #1051 this sets up a release docker image.

I'm testing this out on my own fork to make sure it pushes as expected before I mark this as ready to review. Locally it produces a ~100MB image with a statically linked (+libc) bpftrace.

I've configured my own quay.io/dalehamel/bpftrace repo to use the same triggers as iovisor/bpftrace and am running a build now.


dalehamel:
bad news, the default build timeout on quay.io is 2 hours, which isn't enough time to build LLVM/Clang as "MinSizeRel", rather than the distro default "RelWithDebInfo". By building only libclang, I can get a 76MB bpftrace exe, but this is more than twice the size of the former.

@4ast it looks like you have access to both quay.io and bpftrace repo settings, are you able to add the secret for the bot account from https://quay.io/organization/iovisor?tab=robots for "bpftrace_buildbot" as per https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets 

This way we can do something similar to what kubect-trace does on travis with encrypted secrets, but with github actions (https://github.com/iovisor/kubectl-trace/blob/master/hack/ci-release-image.sh#L15-L18). If the bot secret were added as "QUAY_TOKEN" then it should be easy to use this same approach.

We already build this artifact on github actions, so it should be almost a no-op to do this build there, but we'll need that secret.

cc @danobi

dalehamel:
>  the default build timeout on quay.io is 2 hours,

It looks like quay.io jobs might actually eventually complete, I do see one successful job and it is able to build and push the smaller image.

We can still merge this if that's the case, it'll just be a as jobs are retried. I think that the QUAY_TOKEN is probably more reliable though, and would still suggest we go this route if possible.

dalehamel:
https://quay.io/repository/dalehamel/bpftrace/build/938bf94d-aaca-455f-8175-af605d6dddd6 here is the docker image build, the resulting layer is 40.6 MB 

The docker pull is pretty fast, and I can run the image with

```
docker run quay.io/dalehamel/bpftrace:quay-release bpftrace
```

dalehamel:
@danobi even though the build is slow, I think we should still merge. I can follow up about the QUAY_TOKEN thing on #1051

docker/Dockerfile.release
```diff
@@ -0,0 +1,69 @@
+# This dockerfile is used for quay.io images
+FROM ubuntu:bionic as builder
+
+ARG bcc_ref="v0.12.0"
+ARG LLVM_VERSION="8"
+ENV LLVM_VERSION=$LLVM_VERSION
+
+# There is no need need for most of this once
+# https://github.com/iovisor/bcc/pull/2677 merges, as this is all to build
```
fbs: https://github.com/iovisor/bcc/pull/2677 looks merged

dalehamel: Sorry comment should say “Released”. The packages on the iovisor apt repo are also out of date. Once bcc 0.13.0 is out we shouldn’t need to build it to get all of the static libs, but for now we still do.

I’ll update the comment.

dalehamel:
I added the /tools directory at /usr/local/bin since the size difference is minimal. Now out of the box you can do things like:

```
docker run -ti --privileged -v /sys/kernel/debug:/sys/kernel/debug:rw -v /lib/modules/:/lib/modules:ro  -v /usr/src:/usr/src:ro quay.io/iovisor/bpftrace tcpconnect.bt
```

Which is pretty cool I think. @fbs this should be good to merge

dalehamel:
thanks @danobi I'll follow up with another PR to use the token + github actions. If that works as I expect we can revert this, but good to have it pushing tags / branches now (even if it is slow)

PR #1107, opened 2020-01-27
Print arbitrary data safely

acj:
Fixes #659. This PR introduces a built-in function `buf` that sanitizes arbitrary data so that it's safe to print. The function is similar in implementation to `str` and shares its constraints on length. The differences mostly stem from not using `CreateProbeReadStr` (can't expect the buffer to end on a zero byte) and sorting out the necessary plumbing.

I couldn't find a straightforward way to pass the buffer's length to userspace independently from the buffer itself, so I've embedded it as the first byte in the buffer. Feels a bit odd, but it's simple and works.

I'm opening this as a draft because it still needs docs and perhaps another test. It's ready for initial feedback, though.

```
sudo src/bpftrace -e 'tracepoint:syscalls:sys_enter_sendto { printf("%s\n", buf(args->buff, args->len)); }' -c 'ping 8.8.8.8 -c1'
Attaching 1 probe...
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
\x08\x00\xb1\x85D\xc3\x00\x01\xceQ.^\x00\x00\x00\x0073\x0f\x00\x00\x00\x00\x00\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&'()*+,-./0123456
64 bytes from 8.8.8.8: icmp_seq=1 ttl=52 time=20.5 ms

--- 8.8.8.8 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 20.474/20.474/20.474/0.000 ms
```

PR #1150, opened 2020-02-16
Add uwrite builtin function

### LINKED ISSUE #28: user address space writes
Perhaps we should have a way to call BPF_FUNC_probe_write_user()

https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=96ae52279594470622ff0585621a13e96b700600

This could either be a call, eg, `write()`, or it could just happen if you write to a user address. As an example of the latter, imagine if this worked (needs #4):

```
./src/bpftrace -e 'uprobe:/bin/bash:readline { *uaddr("ps1_prompt") = "BPF-says-hi> "; }'
```

So your bash prompt becomes "BPF-says-hi> " the text time you hit enter.

ajor:
This resolves #28

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -559,7 +559,6 @@ void CodegenLLVM::visit(Call &call)
     //     char[16] inet6;
     //   }
     // }
-    //}
     std::vector<llvm::Type *> elements = {
       b_.getInt64Ty(), // printf ID
       ArrayType::get(b_.getInt8Ty(), 16)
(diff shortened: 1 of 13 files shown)


danobi:
On a high level, it feels like a missed opportunity to not support writing more complex data. In #28, the example was writing a string, but I could see it being useful for users to write fixed size arbitrary data.

I'm thinking to allow for something like this, you'd want to add another helper to encode arbitrary data, essentially the inverse of #1107 . So maybe `uwrite($dst, encode("123abc\x00\x01"))`, where `encode` takes a string of printable ASCII characters or hex encoded values. We might be able to get away without an `encode` builtin and let `uwrite` take the string directly but I figure if #1107 goes in we can reused the `Type::buffer` and maybe reuse `encode` somewhere else in the future.

The nice thing about `encode` would be you wouldn't need to overload `uwrite` to take multiple types to support things other than ints.

Issue #925, opened 2019-10-09
Warn if in lockdown mode


mmarchini:
> his restricts a number of kernel APIs when you boot from UEFI secure boot with a signed kernel. Those APIs include kprobe and bpf.

If a version of lockdown is going to end up in mainline it would be nice if we could detect this mode and generate an appropriate error, like we do for running without root.

_Originally posted by @fbs in https://github.com/iovisor/bpftrace/issues/853#issuecomment-521531382_

fbs:
linking to https://github.com/iovisor/bcc/issues/2565

dailyherold:
My freshly installed 19.10 laptop with `5.3.0-29-generic` kernel had issues with this. Thanks to @psanford in #853 disabling lockdown mode fixed my issues both with bpftrace and bcc (iovisor/bcc#2676). Definitely think we may need to escalate this if all 19.10 users are going to be experiencing the same thing with a fresh install.

fbs:
@dailyherold which bpftrace version were you using? Do you  have a  stacktrace of when it happens? Bpftrace shouldn't segv when it failes to attach code / create maps.

dailyherold:
@fbs I was getting segfaults for any attempt, even hello world, see https://github.com/iovisor/bpftrace/issues/664#issuecomment-583803813. Those results were built off 2667b8a2e1c22613e731c7b68d66e742a58f2249. I can re-enable lockdown and get some more messages from dmesg, or if there's any other more verbose logging/stacktraces that would be helpful? I also have realized desktop ubuntu isn't the primary use case for these tools, so I don't think "escalating" is really necessary. But I do imagine that people picking up the BPF book (like me!) might be running newer distros and wanting to mess around on their local machine while not at work.

fbs:
It seems like  lockdown mode is  only enabled on uefi systems with secure boot which makes it a bit tricky for me to tests with. If you can make a gdb stacktrace that would be quite helpful as the segv doesn't tell me much

```
[12102.514807] bpftrace[16858]: segfault at 5 ip 00007fe60e4f4ba5 sp 00007fff9d4dc808 error 4 in libc-2.30.so[7fe60e38e000+178000]
[...]

PR #1152, opened 2020-02-17
Detect lockdown

fbs:
The first 4 commits of this PR add map and probe type detection to `BPFfeature` and remove some of the boiler plate. To detect a helper adding ` EMIT_HELPER_TEST(override_return, BPF_PROG_TYPE_KPROBE);` should now be enough (after update bpffeature.h).

The last commit adds lock down detection. The detection is based on these two commits:

- https://github.com/torvalds/linux/commit/a94549dd87f5ea4ca50fee493df08a2dc6256b53
- https://github.com/torvalds/linux/commit/9d1f8be5cf42b497a3bddf1d523f2bb142e9318c

If both kprobes and probe_read support is it is likely that lockdown is in effect. It could still be a false positive but in any case the system is not suitable to run bpftrace.

My only test system is an ubuntu 19.10 workstation. For some reason bpf appears to be fully blocked on it, even map creation fails. This doesn't appear to be part of the original patches, but it only happens when secure boot is enabled. It would be good to have more people test this.

```
sudo strace -e bpf -c  bpftrace  --info
The system appears to be in lockdown mode (kprobes and probe_reads are blocked).
Lockdown mode blocks parts of BPF which makes it impossible for bpftrace to function.
Please see https://github.com/iovisor/bpftrace/blob/master/INSTALL.md#disable-lockdown
for more details on lockdown and how to disable it.
Keep in mind that detection is not entirely accurate, there are other
security mechanisms that can cause this too

Kernel helpers
  get_current_cgroup_id: no
  send_signal: no
  override_return: no

Kernel features
  Instruction limit: -1
  Loop support: no

Map types
  hash: no
  percpu hash: no
  array: no
  percpu array: no
  stack_trace: no
  perf_event_array: no

Probe types
  kprobe: no
  tracepoint: no
  perf_event: no

% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100,00    0,000135           3        40        40 bpf
------ ----------- ----------- --------- --------- ----------------
100.00    0,000135                    40        40 total
```

Relates to: #925

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,6 +15,7 @@
   - (*please add sections for other OSes)*
   - [Using Docker](#using-docker)
   - [Generic build](#generic-build-process)
+- [Disable Lockdown](#disable-lockdown)
 
 # Linux Kernel Requirements
 
@@ -250,3 +251,41 @@ To test that the build works, you can try running the test suite, and a one-line
 ./tests/bpftrace_test
 ./src/bpftrace -e 'kprobe:do_nanosleep { printf("sleep by %s\n", comm); }'
 ```
+
+# Disable Lockdown
+
+From the original patch set description:
+
+> This patchset introduces an optional kernel lockdown feature, intended
+> to strengthen the boundary between UID 0 and the kernel. When enabled,
+> various pieces of kernel functionality are restricted. Applications that
+> rely on low-level access to either hardware or the kernel may cease
+> working as a result - therefore this should not be enabled without
+> appropriate evaluation beforehand.
+>
+> The majority of mainstream distributions have been carrying variants of
+> this patchset for many years now, so there's value in providing a
+> doesn't meet every distribution requirement, but gets us much closer to
+> not requiring external patches.
+>
+>   - https://patchwork.kernel.org/patch/11140085/
+
+When lockdown is enabled and set to 'confidentiality' all methods that can
+extract confidential data from the kernel are blocked. This means that:
+
+- kprobes are blocked
+- tracefs access is blocked
+- probe_read and probe_read_str are blocked
+
+which makes it impossible for bpftrace to function.
+
+There are a few ways to disable lockdown.
+
+1. Disable secure boot in UEFI.
+2. Disable validation using mokutil, run the following command, reboot and
+   follow the prompt.
+```
+$ sudo mokutil --disable-validation
+```
+3. Use the `SysRQ+x` key combination to temporarily lift lockdown (until next
+   boot)
--- a/src/bpffeature.h
+++ b/src/bpffeature.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <memory>
 #include <string>
 
 namespace bpftrace {
@@ -7,37 +8,25 @@ namespace bpftrace {
 class BPFfeature
 {
 public:
-  BPFfeature();
-  bool has_loop(void)
-  {
-    return has_loop_;
-  };
-  bool has_helper_send_signal(void)
-  {
-    return has_signal_;
-  };
-  bool has_helper_get_current_cgroup_id(void)
-  {
-    return has_get_current_cgroup_id_;
-  };
-  bool has_helper_override_return(void)
-  {
-    return has_override_return_;
-  };
-  bool instruction_limit(void)
-  {
-    return insns_limit_;
-  };
+  BPFfeature(){};
+
+  int instruction_limit();
+  bool has_loop();
+
+  bool has_helper_override_return();
+  bool has_helper_get_current_cgroup_id();
+  bool has_helper_send_signal();
+
   std::string report(void);
 
 protected:
-  bool has_loop_;
-  int insns_limit_;
+  std::unique_ptr<bool> has_loop_;
+  std::unique_ptr<int> insns_limit_;
 
   /* Helpers */
-  bool has_signal_;
-  bool has_get_current_cgroup_id_;
-  bool has_override_return_;
+  std::unique_ptr<bool> has_send_signal_;
+  std::unique_ptr<bool> has_get_current_cgroup_id_;
+  std::unique_ptr<bool> has_override_return_;
 };
 
 } // namespace bpftrace
--- a/src/bpffeature.h
+++ b/src/bpffeature.h
@@ -16,6 +16,8 @@ public:
   bool has_helper_override_return();
   bool has_helper_get_current_cgroup_id();
   bool has_helper_send_signal();
+  bool has_helper_probe_read();
+  bool has_helper_probe_read_str();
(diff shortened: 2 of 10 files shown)


INSTALL.md
```diff
@@ -250,3 +251,14 @@ To test that the build works, you can try running the test suite, and a one-line
 ./tests/bpftrace_test
 ./src/bpftrace -e 'kprobe:do_nanosleep { printf("sleep by %s\n", comm); }'
 ```
+
+# Disable Lockdown
+
```
fbs: not sure what to put here, other than disabling secure boot in the bios

fbs:
ah found the issue with ubuntu, they ship this (bug):
```
commit 2a68c65abae66d28e2acb3245cb156ae2ea6eb1d
Author: David Howells <dhowells@redhat.com>
Date:   Mon Mar 25 15:09:50 2019 -0700

    UBUNTU: SAUCE: (efi-lockdown) bpf: Restrict kernel image access functions when the kernel is locked down

    There are some bpf functions can be used to read kernel memory:
    bpf_probe_read, bpf_probe_write_user and bpf_trace_printk.  These allow
    private keys in kernel memory (e.g. the hibernation image signing key) to
    be read by an eBPF program and kernel memory to be altered without
    restriction.

    Completely prohibit the use of BPF when the kernel is locked down.

diff --git a/kernel/bpf/syscall.c b/kernel/bpf/syscall.c
index 84470d1480aa..d6f3b787d90d 100644
--- a/kernel/bpf/syscall.c
+++ b/kernel/bpf/syscall.c
@@ -2593,6 +2593,9 @@ SYSCALL_DEFINE3(bpf, int, cmd, union bpf_attr __user *, uattr, unsigned int, siz
        if (sysctl_unprivileged_bpf_disabled && !capable(CAP_SYS_ADMIN))
                return -EPERM;

+       if (kernel_is_locked_down("BPF"))
+               return -EPERM;
+
```

git://git.launchpad.net/~ubuntu-kernel/ubuntu/+source/linux/+git/disco

brendangregg:
For detection there's also

```
# mokutil --sb-state
SecureBoot disabled
```

If that says enabled on Ubuntu, then lockdown is also on.

INSTALL.md
tyhicks: > not sure what to put here, other than disabling secure boot in the bios

Disabling secure boot in the bios is one option. Another option is `sudo mokutil --disable-validation`, rebooting, and then following the prompts to disable secure boot verification in `shim`.

tyhicks:
> ah found the issue with ubuntu, they ship this (bug):

This is currently being relaxed in Ubuntu to allow `bpf(2)` but `bpf_probe_read` and `bpf_probe_read_str` will still be restricted in lockdown mode. See the following bug for details:

 https://bugs.launchpad.net/bugs/1863234

I know that this still isn't the perfect solution for users of bpf but I wanted to provide the updated status here. Hope it helps some.

src/main.cpp
```diff
@@ -78,6 +78,33 @@ void usage()
   // clang-format on
 }
 
+/*
+  Return true and print a warning message if lockdown is enabled.
+*/
+static bool warn_on_lockdown(BPFfeature* feature)
```
jeremycline: For what it's worth I sent https://lore.kernel.org/lkml/20200220151738.1492852-1-jcline@redhat.com/ so once it's merged you should be able to read /sys/kernel/security/lockdown for the current status as an unprivileged user.

fbs: Ah nice. Bpftrace only supports running as root at the moment so we don't  really need it yet but good to have :)

fbs:
I've pushed a new version which makes use the `/sys/kernel/security/lockdown` file. 

Ubuntu is still a special case for now. We can probably remove the exception around the time 20.04 comes out as most users will probably have updated (19.10 latest or 20.04).

With this new approach the changes to bpffeature aren't required anymore, but I've kept them in as I think that they're good to have anyway.

I still need to update the INSTALL notes though

src/lockdown.cpp
```diff
@@ -0,0 +1,87 @@
...
+static LockdownState read_security_lockdown(void)
+{
+  std::ifstream file("/sys/kernel/security/lockdown");
+  if (file.fail())
+    return LockdownState::Unknown;
+
+  // Format: none [integrity] confidentiality
+  while (!file.fail())
+  {
+    std::string line;
```
danobi: This loop was a little confusing since the temp var was named `line` (also the entire `std::*stream` API is confusing). Maybe use `split_string()` or name this `entry`.

src/bpffeature.h
```diff
@@ -1,43 +1,32 @@
...
+  std::unique_ptr<bool> has_loop_;
+  std::unique_ptr<int> insns_limit_;
 
   /* Helpers */
-  bool has_signal_;
-  bool has_get_current_cgroup_id_;
-  bool has_override_return_;
+  std::unique_ptr<bool> has_send_signal_;
+  std::unique_ptr<bool> has_get_current_cgroup_id_;
+  std::unique_ptr<bool> has_override_return_;
```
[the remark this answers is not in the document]
fbs: > Seems like an inappropriate use of unique_ptr. I know we have our build set to c++14, but how about bumping to c++17 and using `std::optional`?

Agreed it is a bit hacky but should work ok for this purpose. Switching to c++17 means we have to have gcc >= 7, not sure how many issues that it going to cause (e.g. ubuntu xenial would be an issue)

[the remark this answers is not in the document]
danobi: Agreed, won't block PR on this.

As a side note, I think you can install newer GCCs from this PPA: https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test . GCC 7 has been out for almost 3 years now, I think it would be reasonable to depend on it be available.

src/lockdown.cpp
fbs: ill rename it to field an add an extra comment.

Not a fan of `*stream` either but I think this works ok in this case.

src/bpffeature.h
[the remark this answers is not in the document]
fbs: created an issue to look at c++17 (#1170 )

fbs:
@danobi  I've updated it  to:

- delete the copy/move constructor/assignment as I don't think they make sense with this unique_ptr hack
- updated the macro to define both the variable and the helper function

fbs:
I've updated the wording too, @brendangregg  can you take a look at it?

fbs:
c++17 support has been enabled but seems to cause some issues for people building on ubunt 16.04. If we decide to keep 17 ill create a follow up to convert this from unique_ptr to std::optional

PR #1155, opened 2020-02-18
Support pushing docker images to quay.io from github actions

dalehamel:
If this works we can revert #1139 as this approach is better i think.

We can push whatever tags we want using this approach, and we're already building the CI artifacts.

I've tested on my personal fork of bpftrace, and tried to make it work seamlessly for iovisor/bpftrace. We won't know for sure until we see something merged to master (including this PR).

The result of this should be the following artifacts:

- An image tagged with SHA-NAME for all Release builds. This will let us very surgically reference specific builds, even if another build has been pushed to a given tag/branch.
- An image tagged with BRANCHNAME-NAME for all Release builds. This will let us keep a floating reference of the latest build of a given type for a branch name
- An image tagged with BRANCHNAME for the vanilla+glibc2.7 image type. This will let us keep a floating reference to the default release image for a given branch name, as well as have tagged releases
- An image tagged with :latest if the branch is master. 

Note that for tags, it will do one build each for the branch and the tag.

Since tagging is just a book-keeping thing, this only actually builds one image (even though it may push multiple tags).


dalehamel:
I ran this test against my own repo, did a tag and also pushed to dalehamel/bpftrace master branch.

I was able to verify that it pushed the tags as expected. I think this is ready to go.

dalehamel:
@danobi please take a look when you have the chance!

PR #1159, opened 2020-02-23
Type fixes

fbs:
With #1108 merged these are now a lot less tedious to do. This is the first PR of a series to get rid of the IR typing issues we have. 

Once all of them are in we can use llvm-as to validate the IR and avoid these kinds of issues in the future.

My plan is to switch all builtin types from arrays to structs as LLVM will handle the typing in that case and we won't have to update all indices if we switch from a i64 to an i32. usym is the first example of this.

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -181,13 +181,13 @@ void CodegenLLVM::visit(Builtin &builtin)
       offset = arch::arg_offset(arg_num);
     }
 
+    Value *ctx = b_.CreatePointerCast(ctx_, b_.getInt64Ty()->getPointerTo());
     // LLVM optimization is possible to transform `(uint64*)ctx` into
     // `(uint8*)ctx`, but sometimes this causes invalid context access.
     // Mark every context acess to supporess any LLVM optimization.
-    expr_ = b_.CreateLoad(
-        b_.getInt64Ty(),
-        b_.CreateGEP(ctx_, b_.getInt64(offset * sizeof(uintptr_t))),
-        builtin.ident);
+    expr_ = b_.CreateLoad(b_.getInt64Ty(),
+                          b_.CreateGEP(ctx, b_.getInt64(offset)),
+                          builtin.ident);
     // LLVM 7.0 <= does not have CreateLoad(*Ty, *Ptr, isVolatile, Name),
     // so call setVolatile() manually
     dyn_cast<LoadInst>(expr_)->setVolatile(true);
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -545,11 +545,19 @@ void CodegenLLVM::visit(Call &call)
   else if (call.func == "usym")
   {
     // store uint64_t[2] with: [0]: (uint64_t)addr, [1]: (uint64_t)pid
-    AllocaInst *buf = b_.CreateAllocaBPF(call.type, "usym");
-    b_.CREATE_MEMSET(buf, b_.getInt8(0), call.type.size, 1);
+    std::vector<llvm::Type *> elements = {
+      b_.getInt64Ty(), // addr
+      b_.getInt64Ty(), // pid
+    };
+    StructType *usym_t = b_.GetStructType("usym_t", elements, false);
+    AllocaInst *buf = b_.CreateAllocaBPF(usym_t, "usym");
+
     Value *pid = b_.CreateLShr(b_.CreateGetPidTgid(), 32);
-    Value *addr_offset = b_.CreateGEP(buf, b_.getInt64(0));
-    Value *pid_offset = b_.CreateGEP(buf, {b_.getInt64(0), b_.getInt64(8)});
+
+    // The extra 0 here ensures the type of addr_offset will be int64
+    Value *addr_offset = b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(0) });
+    Value *pid_offset = b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(1) });
+
     call.vargs->front()->accept(*this);
     b_.CreateStore(expr_, addr_offset);
     b_.CreateStore(pid, pid_offset);
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -654,8 +654,7 @@ void CodegenLLVM::visit(Call &call)
      * The asyncaction_id informs user-space that this is not a printf(), but is a
      * special asynchronous action. The ID maps to exit().
      */
-    ArrayType *perfdata_type = ArrayType::get(b_.getInt8Ty(), sizeof(uint64_t));
-    AllocaInst *perfdata = b_.CreateAllocaBPF(perfdata_type, "perfdata");
+    AllocaInst *perfdata = b_.CreateAllocaBPF(b_.getInt64Ty(), "perfdata");
     b_.CreateStore(b_.getInt64(asyncactionint(AsyncAction::exit)), perfdata);
     b_.CreatePerfEventOutput(ctx_, perfdata, sizeof(uint64_t));
     b_.CreateLifetimeEnd(perfdata);
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -573,15 +573,13 @@ void CodegenLLVM::visit(Call &call)
     //   }
     // }
     //}
-    std::vector<llvm::Type *> elements = {
-      b_.getInt64Ty(), // printf ID
-      ArrayType::get(b_.getInt8Ty(), 16)
-    };
-    StructType *inet_struct = StructType::create(elements, "inet_t", false);
+    std::vector<llvm::Type *> elements = { b_.getInt64Ty(),
+                                           ArrayType::get(b_.getInt8Ty(), 16) };
+    StructType *inet_struct = b_.GetStructType("inet_t", elements, false);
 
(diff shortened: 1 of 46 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -565,15 +573,17 @@ void CodegenLLVM::visit(Call &call)
     //   }
     // }
     //}
+    // clang-format off
```
ajor: I see clang-format is making an interesting choice about this code. Can its settings be adjusted to get it to format this section how we want? I'd rather not introduce lots of "clang-format off" lines throughout the codebase so maybe we should just suck it up and accept its changes here if not (since it's not awful).

src/ast/irbuilderbpf.cpp
```diff
@@ -603,62 +617,73 @@ CallInst *IRBuilderBPF::CreateGetRandom()
...
+      bpftrace_.stackid_maps_[stack_type]->mapfd_);
+  assert(mapfd->getType()->isIntegerTy());
 
   int flags = 0;
   if (ustack)
     flags |= (1<<8);
   Value *flags_val = getInt64(flags);
 
-  // int bpf_get_stackid(ctx, map, flags)
+  // int bpf_get_stackid(struct pt_regs *ctx, struct bpf_map *map, u64
```
ajor: Looks like the end of the comment got cut off

ajor: Also if the 2nd argument is `struct bpf_map *` why's the type `Int64` below?

src/ast/irbuilderbpf.cpp
```diff
@@ -15,6 +15,20 @@ namespace libbpf {
 namespace bpftrace {
 namespace ast {
 
+StructType *IRBuilderBPF::GetStructType(
```
ajor: What's the advantage of caching the StructTypes like this?

src/ast/irbuilderbpf.cpp
ajor: I'm getting some serious deja vu about this question... sorry if you've already answered months ago on another pull request but I've forgotten :P

fbs: The helper we have returns an int64:

```cpp
CallInst *IRBuilderBPF::CreateBpfPseudoCall(int mapfd)
{
  Function *pseudo_func = module_.getFunction("llvm.bpf.pseudo");
  return CreateCall(pseudo_func, {getInt64(BPF_PSEUDO_MAP_FD), getInt64(mapfd)}, "pseudo");
}
...
Value *map_ptr = CreateBpfPseudoCall(bpftrace_.stackid_maps_[stack_type]->mapfd_);
```

And for bpf its just an int too:

```
r1 = 0xffff9e6075850000
```

Guess we could change it to a i64* or struct*, for LLVM it doesn't matter but it might remove confusion?

src/ast/irbuilderbpf.cpp
fbs: The main advantage is that we don't define a lot of duplicate types in the  the generated ir (before it goes through llvm passes), making it  a bit easier to read. e.g. for  tools/tcpaccept.bt we generate a ton of structs:

```
%printf_t.1 = type { i64, [64 x i8], [64 x i8], [64 x i8], [64 x i8], [64 x i8] }
%printf_t.0 = type { i64, [64 x i8], [64 x i8], [64 x i8] }
%printf_t = type { i64 }
%printf_t.9 = type { i64, i64, i64 }
%printf_t.8 = type { i64, [24 x i8], i64, [24 x i8], i64 }
%printf_t.7 = type { i64, i64, [16 x i8] }
%inet_t.6 = type { i64, [16 x i8] }
%inet_t.5 = type { i64, [16 x i8] }
%inet_t.4 = type { i64, [16 x i8] }
%inet_t.3 = type { i64, [16 x i8] }
%inet_t.2 = type { i64, [16 x i8] }
%inet_t = type { i64, [16 x i8] }
```

And it might remove some extra work LLVM has to do de-duplicating all types.

src/ast/codegen_llvm.cpp
fbs: I'll take it out, as we already have the comment above it the formatting doesn't add much.

src/ast/irbuilderbpf.cpp
```diff
@@ -603,62 +617,73 @@ CallInst *IRBuilderBPF::CreateGetRandom()
...
+
   Value *map_ptr = CreateBpfPseudoCall(bpftrace_.perf_event_map_->mapfd_);
 
   Value *flags_val = CreateGetCpuId();
   Value *size_val = getInt64(size);
 
-  // int bpf_perf_event_output(ctx, map, flags, data, size)
+  // int bpf_perf_event_output(struct pt_regs *ctx, struct bpf_map *map,
+  //                           u64 flags, void *data, u64 size)
+  // clang-format off
```
ajor: I think we can do away with this clang-format off as well

src/ast/irbuilderbpf.cpp
```diff
@@ -603,62 +617,73 @@ CallInst *IRBuilderBPF::CreateGetRandom()
 
 CallInst *IRBuilderBPF::CreateGetStackId(Value *ctx, bool ustack, StackType stack_type)
 {
+  assert(ctx && ctx->getType() == getInt8PtrTy());
   assert(bpftrace_.stackid_maps_.count(stack_type) == 1);
-  Value *map_ptr = CreateBpfPseudoCall(bpftrace_.stackid_maps_[stack_type]->mapfd_);
+
+  Value *mapfd = CreateBpfPseudoCall(
```
ajor:
```suggestion
  Value *map_ptr = CreateBpfPseudoCall(
```
From what I understand BpfPseudo is converting our file descriptor into a pointer to the actual map struct, so the original variable name makes more sense

fbs:
@ajor I've updated it to `map_ptr` and `map_ptr->getType()`.  Updating the pseudo call to return a i64* or even `struct bpf_map *` would be a massive change to the codegen with no real gain. 

I don't want to go through the pain of updating all LLVM codegne again :p

fbs:
@danobi  do you have time to review this one? Got a few more like these blocked on this

src/ast/codegen_llvm.cpp
```diff
@@ -181,13 +181,13 @@ void CodegenLLVM::visit(Builtin &builtin)
       offset = arch::arg_offset(arg_num);
     }
 
+    Value *ctx = b_.CreatePointerCast(ctx_, b_.getInt64Ty()->getPointerTo());
```
danobi: I wonder if we can't just do this when `ctx_` is assigned (grep for `ctx_ = `)

fbs:
> LGTM with some small comments. Just wondering, why can't llvm-as currently verify the IR? I assume you have a good reason as you wouldn't go through all this trouble for nothing.

I'm using llvm-as during testing. But it errors on every issue it finds so I cannot add it to the CI until I've fixed all issues.

src/ast/codegen_llvm.cpp
fbs: I eventually want to do this (also see https://github.com/iovisor/bpftrace/pull/1104#discussion_r371649994). I'll do it as a follow up PR once the current ones are in. It should be easier to review and spot issues when all the other fixes have been applied

src/ast/irbuilderbpf.cpp
```diff
@@ -15,6 +15,20 @@ namespace libbpf {
...
+StructType *IRBuilderBPF::GetStructType(
+    std::string name,
+    const std::vector<llvm::Type *> &elements,
+    bool packed)
+{
+  auto search = structs_.find(name);
+  if (search != structs_.end())
+    return search->second;
+
+  StructType *s = StructType::create(elements, name, packed);
```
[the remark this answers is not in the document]
fbs: I'm not 100% sure about this but I think LLVM does all this. In the case of struct the method called is:

```c++
StructType *StructType::create(ArrayRef<Type*> Elements, StringRef Name,
                                bool isPacked) {
   assert(!Elements.empty() &&
          "This method may not be invoked with an empty list");
   return create(Elements[0]->getContext(), Elements, Name, isPacked);
 }

 StructType *StructType::create(LLVMContext &Context, StringRef Name) {
   StructType *ST = new (Context.pImpl->Alloc) StructType(Context);
   if (!Name.empty())
     ST->setName(Name);
   return ST;
 }
```

So the allocation happens through context from through their [allocator](https://llvm.org/doxygen/classllvm_1_1BumpPtrAllocatorImpl.html), which gets cleaned up at the end. 

```
 LLVMContext::~LLVMContext() { delete pImpl; }
```

src/ast/irbuilderbpf.cpp
```diff
@@ -624,26 +624,28 @@ CallInst *IRBuilderBPF::CreateGetRandom()
...
   Value *flags_val = getInt64(flags);
 
-  // int bpf_get_stackid(ctx, map, flags)
+  // int bpf_get_stackid(struct pt_regs *ctx, struct bpf_map *map, u64 flags)
   // Return: >= 0 stackid on success or negative error
   FunctionType *getstackid_func_type = FunctionType::get(
-      getInt64Ty(),
-      {getInt8PtrTy(), getInt8PtrTy(), getInt64Ty()},
-      false);
+      getInt64Ty(), { getInt8PtrTy(), getInt64Ty(), getInt64Ty() }, false);
```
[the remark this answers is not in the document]
fbs: Looks like this is a previous version as it is marked as outdated. Probably this one:

https://github.com/iovisor/bpftrace/pull/1159/files#diff-1917a98fffb60046a7d50cd41791de3eR635

PR #1161, opened 2020-02-24
Add synchronization point between BEFORE and RUN to avoid race in runtime tests

### LINKED ISSUE #1160: Flakiness in runtime tests
Some runtime tests appear to be flakey, in particular one that attach to a process for testing uprobe or USDT tracing functionality.

I believe that this may be due to a potential race between the `BEFORE` clause and the `RUN` clause:

https://github.com/iovisor/bpftrace/blob/master/tests/runtime/engine/utils.py#L100-L115

The tests failures manifest as being unable to attach to the test process, I believe this is because the bpftrace binary can sometimes win a race and beat the test program to execution. 

A possible workaround could be to:

* Use the `-c` flag to have bpftrace run its test program, which should eliminate the race.
* Modify the runtime test suite to be able to block on the process from `BEFORE` having a pid before executing the RUN clause

@ajor recently reported this for a USDT test case:

```
18:30 <@ajor> looks like a spurious failure I've not seen before:   Command: ../src/bpftrace -l 'usdt:*:tracetest2:*' -p $(pidof usdt_test)
18:30 <@ajor>   option requires an argument -- 'p'
18:30 <@ajor> `pidof usdt_test` must have returned nothing
19:18 <@ajor> uh oh, three failures in a row...
19:18 <@ajor> it's the same test failing every time, but only on one (changing) LLVM version per run
19:21 <@ajor> "usdt probes - filter probes by pid on provider" is failing, but "usdt probes - filter probes by pid and wildcard probe name" passes when it's essentially the
              same
19:22 <@ajor> oh well, hopefully it'll fix itself by tomorrow
```

dalehamel:
This is my attempt to fix #1160 

@ajor @mmarchini can you take a look?

Locally this works, but the real test will be to see how this does in CI.

This is also based on my theory of a race condition between BEFORE and RUN

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -46,6 +46,19 @@ Runtime tests will call the bpftrace executable.
 * Run: `sudo make runtime-tests` inside your build folder
 * By default, runtime-tests will look for the executable in the build folder. You can set a value to the environment variable `BPFTRACE_RUNTIME_TEST_EXECUTABLE` to customize it
 
+If you need to run a test program to probe (eg, uprobe/USDT), you can use the
+`BEFORE` clause. The test scripts will wait for the test program to have a pid.
+
+The BEFORE clause will block up to the TIMEOUT waiting for a PID matching the
+basename of the last space-separated token. For instance, if the BEFORE clause
+is `./testprogs/usdt_test`, it will wait for a processed called `usdt_test`.
+If it is `./testprogs/mountns_wrapper usdt_test` it will also wait for a
+process called `usdt_test`. This approach is invalidated if a test program
+requires arguments in the future, but so far test programs are simple and
+separate minimal programs to test tracing functionality, and argument passing
+hasn't been required. If test programs need arguments, a more sophisticated
+approach will be necessary.
+
 ### Test programs
 
 You can add test programs for your runtime tests by placing a `.c` file corresponding to your test program in `tests/testprogs`.
--- a/tests/runtime/engine/utils.py
+++ b/tests/runtime/engine/utils.py
@@ -3,6 +3,7 @@
 import subprocess
 import signal
 import os
+import time
 from os import environ, uname, devnull
 from distutils.version import LooseVersion
 import re
@@ -99,6 +100,17 @@ class Utils(object):
 
             if test.before:
                 before = subprocess.Popen(test.before, shell=True, preexec_fn=os.setsid)
+                waited=0
+                with open(devnull, 'w') as dn:
+                    # This might not work for complicated cases, such as if
+                    # a test program needs to accept arguments. It covers the
+                    # current simple calls with no arguments
+                    child_name = os.path.basename(test.before.split()[-1])
+                    while subprocess.call(["pidof", child_name], stdout=dn, stderr=dn) != 0:
+                        time.sleep(0.1)
+                        waited+=0.1
+                        if waited > test.timeout:
+                            raise TimeoutError('Timed out waiting for BEFORE %s ', test.before)
 
             bpf_call = Utils.prepare_bpf_call(test)
             env = {'test': test.name}


fbs:
I wonder if the race still happens (noticeable like now) when we spawn commands without a shell. Now it first has to spawn a shell, just to have the shell spawn our binary. Looking at current tests we don't need the shell at all.

dalehamel:
> I wonder if the race still happens (noticeable like now) when we spawn commands without a shell. Now it first has to spawn a shell, just to have the shell spawn our binary. Looking at current tests we don't need the shell at all.

Yeah i was wondering that too, but one argument for keeping the shell is that it's a bit friendlier in case we do need any shell features in tests. It might just be legacy of how the original tests were written?

dalehamel:
Also RE the shell, I think that without something like this `WAITPID` we will still have the possibility of a race, even if spawning a subprocess is made to be faster

dalehamel:
Here is my github actions run for this https://github.com/dalehamel/bpftrace/actions/runs/44425718

Will wait and see if it passes there too, but (so far) travis is looking good

tests/runtime/engine/utils.py
```diff
@@ -99,6 +100,14 @@ def run_test(test):
 
             if test.before:
                 before = subprocess.Popen(test.before, shell=True, preexec_fn=os.setsid)
+                if(test.waitpid):
+                    waited=0
+                    with open(devnull, 'w') as dn:
+                        while subprocess.call("pidof %s" % test.waitpid, shell=True, stdout=dn, stderr=dn) != 0:
+                            time.sleep(0.1)
+                            waited+=0.1
+                            if waited > test.timeout:
```
fbs: don't we get up to 2x test.timeout now? test.Timeout to  spawn this binary and another test.Timeout for bpftrace  to run.

fbs: If you raise an  timeout  error  instead of breaking  you fix ^^ and also avoid running bpftrace

dalehamel: > don't we get up to 2x test.timeout now
yes, I figured going with the existing TIMEOUT value was better than hardcoding. 

> If you raise an timeout error instead of breaking you fix ^^ and also avoid running bpftrace

Smart! I'll make the change.

dalehamel: @fbs unless you feel strongly about a separate TIMEOUT for this, i think leaving it is fine. I think in most cases this should find the pid in less than a second, it shouldn't add much time to the CI run and should pretty much never actually hit the TIMEOUT case.

dalehamel:
CI is red because an LLVM mirror sync is happening -_- will push to trigger a build again later.

dalehamel:
CI is green again after LLVM mirror sync, I've squashed.

Here is the GHA run:

https://github.com/dalehamel/bpftrace/actions/runs/44581320

dalehamel:
Ok, finally all green. @fbs I think this might be worth merging just to see if the timeouts seem to go away. I think this will take a large number of tests to reproduce.

Unless we want to try a different approach?

fbs:
Looks like this PR adds a WAITFOR after each BEFORE, might be easier to just have BEFORE always do a WAITFOR (implied). That way you always know the desired process has started before bpftrace  does it's thing, which is what you expect from before.

dalehamel:
@fbs i had considered that, but the issues I see with it are:

- we don’t know what PID we are expecting from BEFORE, if the before process needs to fork or exec, if we just wait for any PID it may not be the right one, and we still have possibility for a race. So how do we know what process to wait for?
- not all tests need to wait for the PID, the ones I have added it to are ones that are expecting a value for “pidof”. Adding the waiting to other tests might make them slower
- we don’t know if the BEFORE script is supposed to exit or keep running during the RUN. If we wait for a bit but the process is meant to exit, the wait may fail

dalehamel:
actually i think i'm beginning to agree with you @fbs these are the only tests that actually use `BEFORE` anyways:

```
grep -r BEFORE tests/runtime
tests/runtime/engine/parser.py:            elif item_name == 'BEFORE':
tests/runtime/uprobe:BEFORE ./testprogs/uprobe_test
tests/runtime/uprobe:BEFORE ./testprogs/uprobe_test
tests/runtime/uprobe:BEFORE ./testprogs/mountns_wrapper uprobe_test
tests/runtime/uprobe:BEFORE ./testprogs/mountns_wrapper uprobe_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_test
tests/runtime/usdt:BEFORE ./testprogs/usdt_semaphore_test

```

dalehamel:
had to close/reopen to get github to refresh commits...

tests/runtime/engine/utils.py
```diff
@@ -99,6 +100,13 @@ def run_test(test):
 
             if test.before:
                 before = subprocess.Popen(test.before, shell=True, preexec_fn=os.setsid)
+                waited=0
+                with open(devnull, 'w') as dn:
+                    while subprocess.call("pidof $(basename $(echo %s | awk '{print $NF}'))" % test.before, shell=True, stdout=dn, stderr=dn) != 0:
```
fbs: ` $(basename $(echo %s | awk '{print $NF}'`

might be nicer to do that basename thing in python, that way you can avoid forking yet another shell just to run pidof.

dalehamel: I reworked it in python, I also added a note on the limitations of this heuristic. It fits the current cases by just taking the basename of the last shell token.

If we augment tests with wrappers like mountns_wrapper rather than pass arguments, this approach works. If future tests take arguments, this will need to change.

I think the established pattern though is to wrap a binary or just make a separate one to keep them simple and separate, rather than having to do argument parsing.

tests/README.md
```diff
@@ -46,6 +46,15 @@ Runtime tests will call the bpftrace executable.
...
 * By default, runtime-tests will look for the executable in the build folder. You can set a value to the environment variable `BPFTRACE_RUNTIME_TEST_EXECUTABLE` to customize it
 
+If you need to run a test program to probe (eg, uprobe/USDT), you can use the
+`BEFORE` clause. The test scripts will wait for the test program to have a pid.
+
+The BEFORE clause will block up to the TIMEOUT waiting for a PID matching the
+basename of the last space-separated token. For instance, if the BEFORE clause
+is `./testprogs/usdt_test`, it will wait for a processed called `usdt_test`.
+If it is `./testprogs/mountns_wrapper usdt_test` it will also wait for a
+process called `usdt_test`.
```
ajor: It'd be good to also explicitly call out that arguments on testprogs aren't supported, as that's something that might confuse people in the future

ajor:
Is there a difference between these two test patterns?
```
BEFORE ./mytestprog
RUN bpftrace ... -p $(pidof mytestprog)
```
```
RUN bpftrace ... -c ./mytestprog
```
If not then the second would solve this problem (and shave off a line per test!).

We should still fix the race condition like you're doing here no matter what, since having a test for the `-p` flag is useful by itself.

tests/README.md
dalehamel: Done!

dalehamel:
> Is there a difference between these two test patterns?

@ajor I suggested the same thing on IRC, I agree we should have test cases using the `-c` approach. For context though, I think these tests are from before the `-c` patch landed, so the pattern predates its use.

As you suggested, we should probably test both cases. For now I just want to fix the race, we can add runtime tests using the `-c` flag and testing this code path separately in another PR.

dalehamel:
@ajor @fbs either of you mind taking another look?

tests/runtime/engine/utils.py
```diff
@@ -99,6 +100,17 @@ def run_test(test):
 
             if test.before:
                 before = subprocess.Popen(test.before, shell=True, preexec_fn=os.setsid)
+                waited=0
+                with open(devnull, 'w') as dn:
+                    # This might not work for complicated cases, such as if
+                    # a test program needs to accept arguments. It covers the
+                    # current simple calls with no arguments
+                    child_name = os.path.basename(test.before.split()[-1])
+                    while subprocess.call("pidof %s" % child_name, shell=True, stdout=dn, stderr=dn) != 0:
```
fbs: won't this work without shell? Less stuff to spawn

dalehamel: @fbs you're right, done! I've been burned by calls like this not passing PATH properly but seems python is ok.

fbs:
Feels a bit too hacky but it's only the test suite, can always revert it.

dalehamel:
> Feels a bit too hacky but it's only the test suite, can always revert it.

Yeah fair, I just wanted to get something in to stop the bleeding - failing builds on master now means that docker images aren't pushed, and that release artifacts won't be published (once we add that).

The original solution of WAITPID was a bit more flexible because you could specify what process to wait for, the current solution is much more rigid and tries to be transparent. As documented it will break down if our test programs need arguments in the future, we can revisit it then.

PR #1166, opened 2020-02-27
cmake: use *_LIBRARIES when testing for libbfd version

danobi:
On openSUSE, binutils has really strangely named libbfd and libopcodes
libraries[1]. This is fine (we can define *_LIBRARIES variables), but the
result is that the LIBBFD_DISASM_FOUR_ARGS_SIGNATURE test will always
fail because CHECK_CXX_SOURCE_COMPILES cannot find the right libraries.

The fix is simple -- just use *_LIBRARIES for CMAKE_REQUIRED_LIBRARIES.
We need to find_library(LibOpcodes) for that test, but that isn't too
bad because CMake caches the result.

[1]: https://bugzilla.suse.com/show_bug.cgi?id=1162312

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/cmake/FindLibBfd.cmake
+++ b/cmake/FindLibBfd.cmake
@@ -6,6 +6,7 @@
 #  LIBBFD_LIBRARIES - Link these to use libbfd
 #  LIBBFD_DEFINITIONS - Compiler switches required for using libbfd
 #  LIBIBERTY_LIBRARIES - libiberty static library (for static compilation)
+#  LIBZ_LIBRARIES - libz static library (for static compilation)
 
 #if (LIBBFD_LIBRARIES AND LIBBFD_INCLUDE_DIRS)
 #  set (LibBpf_FIND_QUIETLY TRUE)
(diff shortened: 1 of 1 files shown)


danobi:
testing ci again

cmake/FindLibBfd.cmake
```diff
@@ -61,7 +78,18 @@ FIND_PACKAGE_HANDLE_STANDARD_ARGS(LibBfd "Please install the libbfd development
...
 
 if(${LIBBFD_FOUND})
-SET(CMAKE_REQUIRED_LIBRARIES bfd opcodes)
+find_package(LibOpcodes)
+SET(CMAKE_REQUIRED_LIBRARIES ${LIBBFD_LIBRARIES} ${LIBOPCODES_LIBRARIES})
+# libbfd.a is not statically linked with libiberty.a or libz.a so we must manually
+# do it. Furthermore, libbfd uses some libc symbols that we must manually
+# link against if we're not using static libc (which includes such symbols).
+if(STATIC_LINKING)
+  list(APPEND CMAKE_REQUIRED_LIBRARIES ${LIBIBERTY_LIBRARIES} ${LIBZ_LIBRARIES})
```
cyphar: I will have to test if this fixes the issue for building on openSUSE, but my gut instinct is that it won't -- the issue is that we have to create a dynamically-linked binary but statically link against `libbfd` (and thus `libopcodes`, `libiberty` and `zlib`). `if(STATIC_LINKING)` doesn't apply to the openSUSE case (though I'm sure it does fix an actual bug).

danobi: I see. I think a `STATIC_BINUTILS` build flag would be reasonable. But that can go in a different PR.

PR #1167, opened 2020-02-27
Extend info flag with system/build info

fbs:
This makes it easier for users to gather system/build info when opening
a bug report.

```
$ bpftrace  --info
System
  OS: Linux 5.3.0-40-generic #32-Ubuntu SMP Fri Jan 31 20:24:34 UTC 2020
  Arch: x86_64

Build
  version: v0.9.4-38-gd079-dirty
  LLVM: 7
  foreach_sym: yes
  unsafe uprobe: no
  btf: no
  bfd: yes

Kernel helpers
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
```

The ifdefs are a bit verbose but I don't see a better way to do it.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -126,6 +126,51 @@ bool is_root()
     return true;
 }
 
+static int info()
+{
+  struct utsname utsname;
+  uname(&utsname);
+
+  std::cerr << "System" << std::endl
+            << "  OS: " << utsname.sysname << " " << utsname.release << " "
+            << utsname.version << std::endl
+            << "  Arch: " << utsname.machine << std::endl;
+
+  std::cerr << std::endl
+            << "Build" << std::endl
+            << "  version: " << BPFTRACE_VERSION << std::endl
+            << "  LLVM: " << LLVM_VERSION_MAJOR << std::endl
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -126,6 +126,52 @@ bool is_root()
...
+#else
+            << "no" << std::endl
+#endif
+            << "  bfd: "
+#ifdef HAVE_BFD_DISASM
+            << "yes" << std::endl
+#else
+            << "no" << std::endl
+#endif
+            << "  ";
```
danobi: Extra whitespace?

src/main.cpp
```diff
@@ -126,6 +126,52 @@ bool is_root()
     return true;
 }
 
+static int info()
+{
+  struct utsname utsname;
+  uname(&utsname);
```
danobi: Check error code?

fbs: Don't think that is needed, the only error in `uname(2)` is

> ERRORS
>       EFAULT buf is not valid.

As this is a stack var I don't see how EFAULT could happen

PR #1169, opened 2020-02-27
Use host network when building docker image

danobi:
I was getting these errors before:
```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.8/main/x86_64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.8/main: temporary error (try again later)
WARNING: Ignoring APKINDEX.adfa7ceb.tar.gz: No such file or directory
fetch http://dl-cdn.alpinelinux.org/alpine/v3.8/community/x86_64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.8/community: temporary error (try again later)
WARNING: Ignoring APKINDEX.efaa1f73.tar.gz: No such file or directory
ERROR: unsatisfiable constraints:
```

Adding this flag seemed to fix the issue.

### DIFF
--- a/build-debug.sh
+++ b/build-debug.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 set -e
-docker run --rm -it -u $(id -u):$(id -g) -v $(pwd):$(pwd) -e STATIC_LINKING=ON -e STATIC_LIBC=ON -e RUN_TESTS=0 bpftrace-builder-alpine "$(pwd)/build-debug" Debug "$@"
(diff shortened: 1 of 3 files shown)


danobi:
I'm not very familiar with docker -- please let me know if this isn't the right fix.

dalehamel:
that shouldn't be necessary, was this locally or in CI?

Sometimes DNS resolution screws up on my laptop when moving between work and office, docker mounts resolv.conf with a bind mount and it can act as a form of caching...

fbs:
Should be ok to do this. You basically remove all network isolation (namespacing) with this, which we don't need for our builds. Might fix issues in edge cases where somone broke the docker0 bridge somehow.

Should probably update the other *build.sh too though.

danobi:
> that shouldn't be necessary, was this locally or in CI?

This was locally on my laptop.

I'll update the other build.sh too.

PR #1174, opened 2020-02-28
Add "edge" build, push master to :latest and :edge

dalehamel:
This adds a new build type, "edge", which tracks bcc master and is based on the latest LLVM supported by bpftrace (LLVM 9). The main utility of this is to ensure that we catch regressions against bcc that may occur between releases, but also offer a way to try out the latest LLVM version.

I also removed the Dockerfile from #1139, as #1155 allows for a much simpler release dockerfile, previously called `Dockerfile.minimal`. I bumped this release dockerfile from ubuntu bionic to ubuntu focal, as this eliminates the security vulnerabilities in the scanner on quay.io.

Finally, I made it so that pushes to master will update both the `:latest` tag, as well as push a tag for the SHA. I found a bug in the docker push script related to this which I've fixed.


dalehamel:
@danobi mind taking a look?

I've only been able to test on my own github actions / quay namespaces, but it seems to do what I expect

dalehamel:
This should also make it easy for us to bisect issues in the future, as this will also push the SHA as a tag,

Eg you can do quay.io/iovisor/bpftrace:SOMESHA as long as the sha the top commit of a PR merged to master.

dalehamel:
You can see this in action by looking at my tags: https://quay.io/repository/dalehamel/bpftrace?tab=tags as well a s my github actions builds https://github.com/dalehamel/bpftrace/actions

docker/Dockerfile.release
```diff
@@ -1,69 +1,10 @@
...
-RUN apt-get update && apt-get install -y curl gnupg &&\
-    llvmRepository="\n\
-deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic main\n\
-deb-src http://apt.llvm.org/bionic/ llvm-toolchain-bionic main\n\
-deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-${LLVM_VERSION} main\n\
-deb-src http://apt.llvm.org/bionic/ llvm-toolchain-bionic-${LLVM_VERSION} main\n" &&\
-    echo $llvmRepository >> /etc/apt/sources.list && \
-    curl -L https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+# Run security updates
+RUN apt-get update && apt-get upgrade -y && rm -rf /var/lib/apt/lists/*
```
fbs: do we need to do this? Won't ubuntu keep  their images up to date?

dalehamel: When I ran this against Ubuntu bionic it dropped the vulnerability count from 7 to 4, according to the quay scanner, so I guess there is some lag on security updates in upstream images 

I also tried Debian buster but it had a high severity one, so I switched to focal. It might be that focal is just so new that there’s no vulns yet.

In any case it doesn’t matter tremendously, the only thing bpftrace really needs I think is glibc.

dalehamel:
I was thinking of adding some docs before merging, I’ll get those up now.

dalehamel:
@fbs I wrote some docs and added to install.md, would appreciate your feedback on that as well. Thanks very much!

dalehamel:
docs added, this can be merged

fbs:
nice

dalehamel:
thanks for the merge fbs

Issue #1173, opened 2020-02-28
bug: tcp probes failing on master


dalehamel:
I did some quick work to bisect.

It seems like it was introduced between 675727a439e6b8fc5e7a1ecc366ad1e241e7271e and f4e377a1972c7ddb03e1a25edb32f4840c7ec88e

The following works for me:

```
docker run -ti --privileged -v /sys/kernel/debug:/sys/kernel/debug:rw -v /lib/modules/:/lib/modules:ro  -v /usr/src:/usr/src:ro quay.io/iovisor/bpftrace:675727a439e6b8fc5e7a1ecc366ad1e241e7271e-debian_llvm_libclang_glibc2.27 tcpconnect.bt
```

And the following fails:

```
$ docker run -ti --privileged -v /sys/kernel/debug:/sys/kernel/debug:rw -v /lib/modules/:/lib/modules:ro  -v /usr/src:/usr/src:ro quay.io/iovisor/bpftrace:f4e377a1972c7ddb03e1a25edb32f4840c7ec88e-debian_llvm_libclang_glibc2.27 tcpconnect.bt
Attaching 2 probes...
Error loading program: kprobe:tcp_connect (try -v)

```

Output with -v:

```

Program ID: 173186

Bytecode:
0: (bf) r6 = r1
1: (b7) r8 = 0
2: (7b) *(u64 *)(r10 -264) = r8
last_idx 2 first_idx 0
regs=100 stack=0 before 1: (b7) r8 = 0
3: (18) r7 = 0xffff9135c8e32a00
5: (85) call bpf_get_smp_processor_id#8
6: (bf) r4 = r10
7: (07) r4 += -264
8: (bf) r1 = r6
9: (bf) r2 = r7
10: (bf) r3 = r0
11: (b7) r5 = 8
12: (85) call bpf_perf_event_output#25
last_idx 12 first_idx 0
regs=20 stack=0 before 11: (b7) r5 = 8
13: (b7) r1 = 1162692948
14: (7b) *(u64 *)(r10 -256) = r1
15: (b7) r1 = 1
16: (7b) *(u64 *)(r10 -264) = r1
17: (b7) r1 = 4475216
18: (63) *(u32 *)(r10 -192) = r1
19: (63) *(u32 *)(r10 -200) = r8
20: (63) *(u32 *)(r10 -204) = r8
21: (63) *(u32 *)(r10 -208) = r8
22: (63) *(u32 *)(r10 -212) = r8
[...]

PR #1175, opened 2020-02-28
Revert "No need to promote scalars to 64-bit"

dalehamel:
This reverts commit 8af25ae9176766839d98d8cc31e1fb499fae36a4 as per #1173 

As indicated in the issue, this commit breaks a number of the tcp tools, possibly others. We'll need to investigate why and add a regression test when this commit is re-introduced.

@fbs @ajor


dalehamel:
The revert diff failed clang-format, so I clang-formatted the patch

PR #1177, opened 2020-02-29
Remove type qualifiers from a cast_type

### LINKED ISSUE #406: const leads to struct not recognized
Consider the following program

```c
#include <linux/fuse.h>
#include <linux/uio.h>

kretprobe::fuse_dev_read {
        $to = (iov_iter *)arg1;
        $buf = $to->iov->iov_base;
}
```

On it bpftrace gives:

```
root@deco:# bpftrace x.bt 
Unknown struct/union: 'const struct iovec'
```

However if I manually remove const from `iov_iter`  struct:

```c
#include <linux/fuse.h>
#include <linux/uio.h>

struct xiov_iter {
        unsigned int type;
        size_t iov_offset;
        size_t count;
        union {
                /*const*/ struct iovec *iov;                     // <-- NOTE
                /*const*/ struct kvec *kvec;
                /*const*/ struct bio_vec *bvec;
                struct pipe_inode_info *pipe;
        };
        union {
                unsigned long nr_segs;
                struct {
                        int idx;
                        int start_idx;
                };
        };
};

kretprobe::fuse_dev_read {
        $to = (xiov_iter *)arg1;
        $buf = $to->iov->iov_base;
}
```

it compiles ok.

mmisono:
Clang parser does not remove type qualifiers ("const", "volatile, and
"restrict" keywords) from a cast_type of Type::cast, and this results in
a casting error. For example:

```
% sudo ./src/bpftrace  -e "kprobe:vfs_open { @ = ((struct file*)arg1)->f_cred->usage; }"
stdin:1:23-54: ERROR: Unknown struct/union: 'const struct cred'
kprobe:vfs_open { @ = ((struct file*)arg1)->f_cred->usage; }
```

Remove type qualifiers to fix this. libclang C API does not provide a
direct way to remove type qualifiers, so  remove "const" / "volatile" /
"restrict" keywords from a string obtained by `clang_getTypeSpelling()`.

In addition to this, fix BTF::type_of to remove type qualifiers from a
btf_type. Otherwise, BTF::type_of fails to get a struct name (it'll
return "(anon)").

Resolves #406


src/clang_parser.cpp
```diff
@@ -201,10 +202,25 @@ static bool translateMacro(CXCursor cursor, std::string &name, std::string &valu
   return value.length() != 0;
 }
 
+static std::string remove_qualifiers(std::string &&typestr)
+{
+  // libclang prints "const" keyword first
+  // https://github.com/llvm-mirror/clang/blob/65acf43270ea2894dffa0d0b292b92402f80c8cb/lib/AST/TypePrinter.cpp#L137-L157
+  return std::regex_replace(typestr,
+                            std::regex("^(const volatile\\s+)|^(const\\s+)|"
+                                       "^(volatile\\s+)|\\*(\\s*restrict)$"),
```
danobi: I tried pretty hard to come up with something better but I couldn't.

PR #1183, opened 2020-03-04
Fix reading USDT probe arguments on AArch64

nick-arm:
On AArch64 the register names in USDT probe arguments from libbcc are names of fields in `struct pt_regs` like "regs[0]", "regs[1]", etc. rather than the "r0", "r1", names currently used by bpftrace.

Modified `arch::offset()` to accept the pt_regs style names in addition to the existing names so we can generate the correct structure offsets.

Also fixed a small bug with the existing AArch64 register definitions where it defines registers r0..r31 but the `regs` array in `struct pt_regs` only has 31 elements. So the offsets of the registers after r31 are incorrect and e.g. `reg("sp")` actually reads "pc".

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -348,6 +348,13 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx, struct bcc_usdt_argument
   if (argument->valid & BCC_USDT_ARGUMENT_BASE_REGISTER_NAME) {
     int offset = 0;
     offset = arch::offset(argument->base_register_name);
+    if (offset < 0)
+    {
+      std::cerr << "offset for register " << argument->base_register_name
+                << " not known" << std::endl;
+      abort();
+    }
+
     Value* reg = CreateGEP(ctx, getInt64(offset * sizeof(uintptr_t)), "load_register");
     AllocaInst *dst = CreateAllocaBPF(builtin.type, builtin.ident);
     CreateProbeRead(dst, builtin.type.size, reg);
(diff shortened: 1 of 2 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -348,6 +348,12 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx, struct bcc_usdt_argument
   if (argument->valid & BCC_USDT_ARGUMENT_BASE_REGISTER_NAME) {
     int offset = 0;
     offset = arch::offset(argument->base_register_name);
+    if (offset < 0) {
```
fbs: Doesn't the semantic analyser catch this? Semantic analyser gives nicer errors than this here does.

danobi:
> So the offsets of the registers after r31 are incorrect and e.g. reg("sp") actually reads "pc". 

Nice catch.

danobi:
Also, please see clang-format failure for formatting fixes.

src/ast/irbuilderbpf.cpp
nick-arm: No, it silently uses offset -1. I don't think the semantic analyser sees `base_register_name`: it comes from libbcc in the second overload of `CreateUSDTReadArgument` just below.

nick-arm:
> 
> 
> Also, please see clang-format failure for formatting fixes.

Fixed.

src/ast/irbuilderbpf.cpp
fbs: We can probably move it, but it will be quite a lot of work for a slightly nicer error message

fbs:
Thanks!

PR #1184, opened 2020-03-05
Require C++17 to build

### LINKED ISSUE #1170: Switch to C++17
Seems like an inappropriate use of unique_ptr. I know we have our build set to c++14, but how about bumping to c++17 and using `std::optional`?

_Originally posted by @danobi in https://github.com/iovisor/bpftrace/pull/1152_

---

C++17 has some new stuff that is useful (like ^^, std::filesystem looks useful too) but the might cause build issues for users on older distros. According to [wikipedia](https://en.wikipedia.org/wiki/C%2B%2B17):

- GCC since version 7 has complete support for C++17.
- Clang 5 and later implement all the features of C++17
- libstdc++ since version 9.1 has complete support for c++17 (8.1 without Parallelism TS and referring to C99 instead of C11)
- libc++ as of version 9 has partial support for c++17, with the remainder "in progress"

GCC7 shipped in may 2017 so it should be available on most distros by now. Ubuntu 16.04 is the only exception but I'm not sure if we still have users on that.

Can we enable this safely or would it impact users?

danobi:
C++17 has some nice features like std::optional and std::filesystem that
we would like to use.

This closes #1170 .


danobi:
Just trying to see if this will build first.

PR #1186, opened 2020-03-06
Add kfunc support

olsajiri:
Adding support to specify kfunc/kretfunc attach points.

The 'kfunc' attach point is a bpftrace 'code name' for 
function trampolines, which is a new way to attach BPF 
program to kernel functions.

The trampolines are closely binded with BTF kernel data,
and that allows them to know the name and type of arguments.

We are able to use function arguments by name as standard
bpftrace variables with '$' prefix', like:

```
  # bpftrace  -e 'kretfunc:fget { printf("fd %d\n", $fd); }'
  Attaching 1 probe...
  fd 3
  ... 

```
We are also able to access return value via $ret variable
(together with kfunc function arguments), like:

```
  # bpftrace -e 'kretfunc:fget { printf("fd %d name %s\n", $fd, str($ret->f_path.dentry->d_name.name));  }'  
  Attaching 1 probe...
  fd 9 name _journal.cpython-37m-x86_64-linux-gnu.so
  ... 

```
As I already mentioned and presented above the kretfunc has access
to function arguments as well and could speed up some scripts,
that need function arguments at the return probe.

The kfunc and kretfunc is also a starting point for using BTF 
powered argument passing to kernel helpers, which would allow
us to do fast filepath helper, that operates on 'struct file'
pointer directly. I have the code ready both for kernel and 
bpftrace and it will follow once this change in.


fbs:
Nice! Very useful to have.

I'm not convinced by the use of "hidden" variables for this purpose. I can see this cause confusion, e.g. when you start using bpftrace scripts between kernel versions and the kernel function signature change starts clashing with your existing . It also introduces another way to access args. We already have `arg0..9, retval` for k/uprobes and the `args` struct for tracepoints.
Maybe we can use `args` for this too? E.g. `kretfunc:fget {@=args->fd }` instead of `kretfunc:fget { @=$fd }`.

src/ast/semantic_analyser.cpp
```diff
@@ -1695,6 +1698,20 @@ void SemanticAnalyser::visit(AttachPoint &ap)
       }
     }
   }
+  else if (ap.provider == "kfunc" || ap.provider == "kretfunc") {
```
fbs: if the kernel doesn't support these a warning should be emitted too.

src/attached_probe.cpp
```diff
@@ -68,6 +68,8 @@ bpf_prog_type progtype(ProbeType t)
     case ProbeType::software:   return BPF_PROG_TYPE_PERF_EVENT; break;
     case ProbeType::watchpoint: return BPF_PROG_TYPE_PERF_EVENT; break;
     case ProbeType::hardware:   return BPF_PROG_TYPE_PERF_EVENT; break;
+    case ProbeType::kfunc:      return BPF_PROG_TYPE_TRACING; break;
+    case ProbeType::kretfunc:   return BPF_PROG_TYPE_TRACING; break;
```
fbs: should we add these to the libbpf.h we vendored, that way we avoid compile issues

src/attached_probe.cpp
```diff
@@ -174,6 +184,11 @@ AttachedProbe::~AttachedProbe()
     case ProbeType::kretprobe:
       err = bpf_detach_kprobe(eventname().c_str());
       break;
+    case ProbeType::kfunc:
+    case ProbeType::kretfunc:
+      close(tracing_fd_);
+      err = bpf_detach_kfunc(progfd_, NULL);
```
fbs: Looks like this is going to cause issues on systems that don't have this.

tests/semantic_analyser.cpp
```diff
@@ -979,6 +983,21 @@ TEST(semantic_analyser, watchpoint)
   test("watchpoint::0x0:8:rww { 1 }", 1);
 }
 
+TEST(semantic_analyser, kfunc)
+{
+  BTF btf;
+
+  if (!btf.has_function("memcpy"))
```
fbs: can we fake the btf data so that we don't have to skip the tests?

olsajiri:
> I'm not convinced by the use of "hidden" variables for this purpose. I can see this cause confusion, e.g. when you start using bpftrace scripts between kernel versions and the kernel function signature change starts clashing with your existing . It also introduces another way to access args. We already have `arg0..9, retval` for k/uprobes and the `args` struct for tracepoints.
> Maybe we can use `args` for this too? E.g. `kretfunc:fget {@=args->fd }` instead of `kretfunc:fget { @=$fd }`.

ok, that seems like good idea, should not be a problem to do that

src/types.h
```diff
@@ -141,6 +145,8 @@ const std::vector<ProbeItem> PROBE_LIST =
   { "software", "s", ProbeType::software },
   { "hardware", "h", ProbeType::hardware },
   { "watchpoint", "w", ProbeType::watchpoint },
+  { "kfunc", "f", ProbeType::kfunc },
+  { "kretfunc", "rf", ProbeType::kretfunc },
```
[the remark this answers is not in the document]
olsajiri: ok, will change

src/attached_probe.cpp
olsajiri: yep, I noticed that in the CI tests, did not notice out libbpf.h ;-) thnx

src/attached_probe.cpp
olsajiri: ah right.. I'll detect these in cmake

tests/semantic_analyser.cpp
olsajiri: yep, will do

src/list.cpp
```diff
@@ -286,6 +287,8 @@ void list_probes(const BPFtrace &bpftrace, const std::string &search_input)
     std::cout << probe << std::endl;
   }
 
+  // kfuncs
+  bpftrace.btf_.display_funcs();
```
mmisono: Is there a function that kfuncs can trace but kprobe can't? If not, then I think showing parameters of kprobe func is straightforward. e.g., something like

```
% bpftrace -lv 'kprobe:vfs_open'
kprobe:vfs_open
    const struct path___2 * $path;
    struct file___2 * $file;
    int $ret;
```

(by the way, I'm not sure why trailing `___2` is added to some struct name...)

src/btf.cpp
```diff
@@ -303,8 +303,137 @@ std::string BTF::type_of(const btf_type *type, const std::string &field)
...
+
+    if (btf_is_composite(t))
+    {
+      const char *cast = btf_str(btf, t->name_off);
+
+      if (cast)
+      {
+        std::string comp = btf_is_struct(t) ? "struct" : "union";
+
+        stype.type = Type::cast;
```
mmisono: Recently, Tyep::ctx is added (#1104). Type::ctx is almost the same as the Type::cast, but codegen emits a load instruction instead of bpf_probe_read() when accessing a value. In my understanding, the verifier allow for BPF program to access a field of argument directly in kfunc/kretfunc since it can ensure a type of argument thanks to the BTF. So using Type::ctx here (and in the FieldAnalyser) will eliminate one bpf_probe_read().

... but currently Type::ctx value cannot be stored in map (https://github.com/iovisor/bpftrace/commit/b7d2510b9a7a13f3e71685fa35fb52c50e265f4e). So we need to change a semantic analyzer in some ways.

Well, using Type::ctx can be made in another PR.

danobi:
Awesome, was actually just thinking about this. (Haven't looked at code changes yet)

cc @4ast .

danobi:
Would be interesting to see if we can do something about https://github.com/iovisor/bpftrace/blob/master/src/main.cpp#L64 with kfunc. Alexei says setup should be faster but teardown cost is not yet profiled.

src/btf.cpp
```diff
@@ -433,6 +433,82 @@ int BTF::resolve_args(std::string &func, std::map<std::string, SizedType> &varia
...
+      continue;
+
+    const struct btf_param *p;
+    int j;
+
+    for (j = 0, p = btf_params(t); j < btf_vlen(t); j++, p++)
+    {
+      type = std::string("");
+
+      err = btf_dump__emit_type_decl(dump, p->type, &decl_opts);
```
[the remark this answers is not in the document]
olsajiri: true, will change, thanks

src/btf.cpp
```diff
@@ -433,6 +433,82 @@ int BTF::resolve_args(std::string &func, std::map<std::string, SizedType> &varia
...
+    type = std::string("");
+
+    err = btf_dump__emit_type_decl(dump, t->type, &decl_opts);
+    if (err)
+      break;
+
+    std::cout << "    " << type << " $ret;" << std::endl;
+  }
+
+  if (id != max)
```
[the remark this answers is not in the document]
olsajiri: yep, that was buggy

src/list.cpp
olsajiri: yes, I have it like this in new version

src/btf.cpp
olsajiri: right, I'm using Type::ctx in new version

olsajiri:
> Would be interesting to see if we can do something about https://github.com/iovisor/bpftrace/blob/master/src/main.cpp#L64 with kfunc. Alexei says setup should be faster but teardown cost is not yet profiled.

hum, do you mean BPFTRACE_MAX_PROBES?

danobi:
> hum, do you mean BPFTRACE_MAX_PROBES?

Yeah, I think main.cpp changed and my link points to something else now

olsajiri:
the test failures seem ok.. there's clang format check,
which would force me to change code I did not changed,
so I think it's ok

and there's following one:
  tar: invalid tar magic
  The command '/bin/sh -c tar xvf /bcc.tar.gz' returned a non-zero code: 1

olsajiri:
I re-pushed and now it's only the clang error.. so we should be good ;-)

docs/reference_guide.md
```diff
@@ -1319,6 +1321,87 @@ Examples:
...
 ```
 
+## 15. `kfunc`/`kretfunc`: Kernel Functions Tracing
+
+
+Syntax:
+
+```
+kfunc::function
```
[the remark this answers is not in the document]
olsajiri: oops, will change

src/ast/codegen_llvm.cpp
```diff
@@ -1149,6 +1149,15 @@ void CodegenLLVM::visit(FieldAccess &acc)
   assert(type.type == Type::cast || type.type == Type::ctx);
   acc.expr->accept(*this);
 
+  if (type.is_kfarg)
+  {
+    expr_ = b_.CreateLoad(
+        b_.getInt64Ty(),
+        b_.CreateGEP(ctx_, b_.getInt64(acc.type.kfarg_idx * sizeof(__u64))),
```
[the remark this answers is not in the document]
olsajiri: sounds good, will check

CMakeLists.txt
```diff
@@ -107,6 +109,16 @@ set(CMAKE_REQUIRED_LINK_OPTIONS -L${LIBBCC_LIBDIR})
 check_symbol_exists(bcc_prog_load "${LIBBCC_INCLUDE_DIRS}/libbpf.h" HAVE_BCC_PROG_LOAD)
 check_symbol_exists(bcc_create_map "${LIBBCC_INCLUDE_DIRS}/libbpf.h" HAVE_BCC_CREATE_MAP)
 check_symbol_exists(bcc_elf_foreach_sym "${LIBBCC_INCLUDE_DIRS}/bcc_elf.h" HAVE_BCC_ELF_FOREACH_SYM)
+check_symbol_exists(bpf_attach_kfunc "${LIBBCC_INCLUDE_DIRS}/libbpf.h" HAVE_BCC_KFUNC)
+INCLUDE(CheckCXXSourceCompiles)
+CHECK_CXX_SOURCE_COMPILES("#include <linux/bpf.h>
+int main() {
+  return BPF_PROG_TYPE_TRACING;
+}" HAVE_BPF_PROG_TYPE_TRACING)
```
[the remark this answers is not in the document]
olsajiri: ok

src/btf.cpp
```diff
@@ -207,6 +207,10 @@ static std::string btf_type_str(const std::string& type)
 
 std::string BTF::c_def(std::unordered_set<std::string>& set)
 {
+  if (!has_data())
+    return std::string("");
+  ;
```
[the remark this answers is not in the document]
olsajiri: oops, yes..

olsajiri:
rebased..

src/ast/semantic_analyser.cpp
```diff
@@ -1725,6 +1756,31 @@ void SemanticAnalyser::visit(AttachPoint &ap)
...
+  else if (ap.provider == "kfunc" || ap.provider == "kretfunc")
+  {
+#ifdef HAVE_KFUNC
+    bool supported = bpftrace_.btf_.has_data();
+#else
+    bool supported = false;
+#endif
+
+    if (!supported)
+      error("kfunc/kretfunc not available for your kernel version.", ap.loc);
```
fbs:
```suggestion
{
      error("kfunc/kretfunc not available for your kernel version.", ap.loc);
      return;
}
```

To avoid a 'Failed to resolv...' error too

src/list.cpp
fbs: All other list functions support regex based filtering. e.g. `bpftrace -l 'k:get*'`, can you add that here too? (see `search_probe()`)

fbs: did you implement this?

fbs:
other than the issue around listing, LGTM. 

Porting tcpconnect was quite easy,  `s/$sk/args->sk/` and it works :)

src/ast/semantic_analyser.cpp
olsajiri: ok

olsajiri:
it's again only clang format failures

mmisono:
Can we access a return value  if there is a function that use `ret` as an argument name?

src/attached_probe.h
```diff
@@ -41,11 +41,14 @@ class AttachedProbe
...
   void attach_hardware();
   void attach_watchpoint(int pid, const std::string &mode);
+  void attach_kfunc(void);
+  int detach_kfunc(void);
 
   Probe &probe_;
   std::tuple<uint8_t *, uintptr_t> func_;
   std::vector<int> perf_event_fds_;
   int progfd_ = -1;
+  int tracing_fd_ = -1;
```
mmisono, line 51: For suppressing warning about unused variable if kfunc is not available.
```suggestion
#ifdef HAVE_KFUNC
  int tracing_fd_ = -1;
#endif 
```

src/list.cpp
mmisono: I thought we could unite kprobe and kfunc here, but I found that there was no BTF data for static function...
The new version looks good to me.

src/attached_probe.h
olsajiri: right, will add

olsajiri:
> Can we access a return value if there is a function that use `ret` as an argument name?

damn..  you're right :-\
how about using 'retval' as we do for kprobe/uprobe ? I'll check on that

thanks

fbs:
Looks like the listing doesnt fully work yet, it only lists with a regex present now

```
[vagrant@localhost bpftrace]$ sudo bpftrace --version
bpftrace v0.9.4-146-g1940-dirty
[vagrant@localhost bpftrace]$ sudo bpftrace -l 'kfunc:*' | wc -l
21863
[vagrant@localhost bpftrace]$ sudo bpftrace -l  | grep -c kfunc
0
```
The `-lv` from the reference guide doesnt work either.

> > Can we access a return value if there is a function that use `ret` as an argument name?
> 
> damn.. you're right :-\
> how about using 'retval' as we do for kprobe/uprobe ? I'll check on that

Does accessing arguments make sense for a kretfunc? Or does the kernel somehow store the input args until the return probes have been evaluated?

olsajiri:
> Looks like the listing doesnt fully work yet, it only lists with a regex present now
> 
> ```
> [vagrant@localhost bpftrace]$ sudo bpftrace --version
> bpftrace v0.9.4-146-g1940-dirty
> [vagrant@localhost bpftrace]$ sudo bpftrace -l 'kfunc:*' | wc -l
> 21863
> [vagrant@localhost bpftrace]$ sudo bpftrace -l  | grep -c kfunc

oops. sry, will fix

> The `-lv` from the reference guide doesnt work either.

dtto

> > > Can we access a return value if there is a function that use `ret` as an argument name?
> > 
> > 
> > damn.. you're right :-
> > how about using 'retval' as we do for kprobe/uprobe ? I'll check on that
> 
> Does accessing arguments make sense for a kretfunc? Or does the kernel somehow store the input args until the return probes have been evaluated?

it's actually one of the nice kfunc/trampoline features, that you can
access function arguments in return handler, because there's plenty
of tools that capture function arguments and process them in the return
handler.. so you could use just one probe instead of two

src/attached_probe.cpp
```diff
@@ -66,6 +66,16 @@ bpf_prog_type progtype(ProbeType t)
     case ProbeType::software:   return BPF_PROG_TYPE_PERF_EVENT; break;
     case ProbeType::watchpoint: return BPF_PROG_TYPE_PERF_EVENT; break;
     case ProbeType::hardware:   return BPF_PROG_TYPE_PERF_EVENT; break;
+#ifdef HAVE_KFUNC
+    case ProbeType::kfunc:      return BPF_PROG_TYPE_TRACING; break;
+    case ProbeType::kretfunc:   return BPF_PROG_TYPE_TRACING; break;
+#else
+    case ProbeType::kfunc:
+    case ProbeType::kretfunc:
+      std::cerr << "ERRPR: kfunc not available for your kernel version"
```
mmisono, line 75:
```suggestion
      std::cerr << "ERROR: kfunc not available for your kernel version"
```

mmisono:
Using `retval` seems good.

src/attached_probe.cpp
olsajiri: ugh, thanks

mmisono:
I found two problems:

- reg() is allowed to use in kfunc and kretrunc. This should be checked in the semantic analyzer.

```
% sudo ./src/bpftrace -e 'f:vfs_open { printf("%d\n", reg("ip"));  }' -v
Attaching 1 probe...

Error log:
0: (bf) r6 = r1
1: (b7) r1 = 0
2: (7b) *(u64 *)(r10 -16) = r1
last_idx 2 first_idx 0
3: (79) r1 = *(u64 *)(r6 +128)
invalid bpf_context access off=128 size=8

Error loading program: kfunc:vfs_open
```

- `probe` builtin does not work.

```
% sudo ./src/bpftrace -e 'f:vfs_open { printf("%s\n", probe);  }' -v
terminate called after throwing an instance of 'bpftrace::WildcardException'
  what():  Wildcard matches aren't available on probe type 'kfunc'
zsh: abort      sudo ./src/bpftrace -e 'f:vfs_open { printf("%s\n", probe);  }' -v
```

olsajiri:
> sudo ./src/bpftrace -e 'f:vfs_open { printf("%s\n", probe);  }' -v

ok, reg needs to be banned for kfunc.. for probe, I guess it'd be
better to make the expansion working now

olsajiri:
> > sudo ./src/bpftrace -e 'f:vfs_open { printf("%s\n", probe);  }' -v
> 
> ok, reg needs to be banned for kfunc.. for probe, I guess it'd be
> better to make the expansion working now

after some more thought, I'm not enabling kfunc expansion at the moment:
  - expansion is done in code generation and I'll need it already in semantic analyser,
   to get all the arguments.. so not small factoring would be needed
  - we'd need to check that all the expansed functions have the same arguments,
   which in my opinion won't be the case for most of the kernel functions, so
   I don't think it's too useful feature
  - if really needed we could do that in separate PR

mmisono:
> I'm not enabling kfunc expansion at the moment:

I think that is fine for now. 

> we'd need to check that all the expansed functions have the same arguments

Indeed, if we want to attach the same kfunc BPF program to multiple functions, then functions must have the same arguments. On the other hand, bpftrace supports attaching a single program to multiple functions that have different arguments by creating a BPF program for each function. e.g., :

```
% sudo ./src/bpftrace -e 'tracepoint:syscalls:sys_enter* { @[probe] = count(); }'
Attaching 332 probes...
^C
```
(though in this case, I believe there is no need to create a BPF program for each function)

I think this is useful. For example, we would be able to create kfunc version of [vfscount.bt](https://github.com/iovisor/bpftrace/blob/master/tools/vfscount.bt): 

```
% sudo ./src/bpftrace -e 'f:vfs_* { @[probe] = count(); }'
stdin:1:1-8: ERROR: Failed to resolve kfunc args.
f:vfs_* { @[probe] = count(); }
~~~~~~~
```
(By the way, printing like "wildcard expansion is not supported" would be easy to understand the rror.)

olsajiri:
> I think this is useful. For example, we would be able to create kfunc version of [vfscount.bt](https://github.com/iovisor/bpftrace/blob/master/tools/vfscount.bt):
> 
> ```
> % sudo ./src/bpftrace -e 'f:vfs_* { @[probe] = count(); }'
> stdin:1:1-8: ERROR: Failed to resolve kfunc args.
> f:vfs_* { @[probe] = count(); }
> ~~~~~~~
> ```

aaah right.. when we are not using args->xxx we don't need that check ;-)
I'll do the wildcards expansion for kfunc in separate PR

> (By the way, printing like "wildcard expansion is not supported" would be easy to understand the rror.)

ok, changing to this:

```
# ./src/bpftrace -e 'f:vfs_* { @[probe] = count(); }'
stdin:1:1-8: ERROR: wildcard expansion is not supported for kfunc/kretfunc probes
f:vfs_* { @[probe] = count(); }
~~~~~~~
```

mmisono:
Looks good to me. We can make improvements later. kfunc is very interesting!

fbs:
I am tempted to merge this and do the 0.10 release. This one is still has some improvements/issues but those can be done in a later version too. The basics look good to me and this one is getting quite big. 

This is mostly a new code path too, so getting bad regressions into the release is unlikely too (I don't see any).

@danobi  any objections?

src/ast/semantic_analyser.cpp
```diff
@@ -1725,6 +1793,34 @@ void SemanticAnalyser::visit(AttachPoint &ap)
...
+  {
+#ifdef HAVE_KFUNC
+    bool supported = bpftrace_.btf_.has_data();
+#else
+    bool supported = false;
+#endif
+
+    if (!supported)
+    {
+      error("kfunc/kretfunc not available for your kernel version.", ap.loc);
```
danobi, line 1806: As a user, do you think it'd be helpful to see an error message about BTF? Or is that too implementation specific?

src/ast/semantic_analyser.cpp
```diff
@@ -1725,6 +1793,34 @@ void SemanticAnalyser::visit(AttachPoint &ap)
...
+      return;
+    }
+
+    auto ap_map = bpftrace_.btf_ap_args_;
```
danobi, line 1810: I think we can avoid a copy here
```suggestion
    const auto& ap_map = bpftrace_.btf_ap_args_;
```

src/btf.cpp
```diff
@@ -303,8 +312,258 @@ std::string BTF::type_of(const btf_type *type, const std::string &field)
...
+    else
+    {
+      stype.type = Type::integer;
+    }
+  }
+
+  return stype;
+}
+
+int BTF::resolve_args(std::string &func,
```
danobi, line 395:
```suggestion
int BTF::resolve_args(const std::string &func,
```

src/ast/field_analyser.cpp
```diff
@@ -159,6 +187,31 @@ void FieldAnalyser::visit(Predicate &pred)
...
+    if (!bpftrace_.btf_.resolve_args(ap.func,
+                                     ap_args_,
+                                     ap.provider == "kretfunc"))
+    {
+      // store/save args for each kfunc ap for later processing
+      bpftrace_.btf_ap_args_.insert({ ap.provider + ap.func, ap_args_ });
+
+      // pick up cast arguments immediately and let the
+      // FieldAnalyser to resolve args builtin
+      for (auto arg : ap_args_)
```
danobi, line 206:
```suggestion
      for (const auto& arg : ap_args_)
```

danobi:
> @danobi any objections?

Nope, I think this should be fine to merge for 0.10.0.

src/ast/semantic_analyser.cpp
olsajiri: user just used 'kfunc or kretfunc' so the message seems related,
maybe it'd more confusing mentioning BTF, not sure ;-)

src/ast/semantic_analyser.cpp
olsajiri: ok

src/btf.cpp
olsajiri: ok

src/ast/field_analyser.cpp
olsajiri: ok

danobi:
Can you rebase again? Master was broken for a bit.

src/ast/irbuilderbpf.cpp
```diff
@@ -790,5 +790,20 @@ void IRBuilderBPF::CreateOverrideReturn(Value *ctx, Value *rc)
   CreateCall(override_func, { ctx, rc }, "override");
 }
 
+Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
+                                   SizedType &type,
+                                   std::string &name)
+{
+  ctx = CreatePointerCast(ctx, getInt64Ty()->getPointerTo());
+  Value *expr = CreateLoad(getInt64Ty(),
```
fbs, line 798: As work is being done on getting more architectures support; is this always 64 on all architectures?

olsajiri: hm, so the pointer could be 32bit and we would add 64bit number,
not sure how would this be handled.. but if it's a problem we have it
in other places, because I stole this code from retval builtin and
saw it on other places ;-)

Issue #1165, opened 2020-02-26
Support Clang 10.0.0


Leo-Yan:
Hi there,

When built bpftrace with Clang 10.0.0, found below errors.  

~~~
[17/38] /home/leo.yan/volumes/OE-build/poky/build_arm64/tmp/work/aarch64-poky-linux/bpftrace/0.9.4-r0/recipe-sysroot-native/usr/bin/aarch64-poky-linux/aarch64-poky-linux-g++   -I/home/leo.yan/volumes/OE-build/pp
FAILED: src/ast/CMakeFiles/ast.dir/irbuilderbpf.cpp.o                                                                                                                                                              
/home/leo.yan/volumes/OE-build/poky/build_arm64/tmp/work/aarch64-poky-linux/bpftrace/0.9.4-r0/recipe-sysroot-native/usr/bin/aarch64-poky-linux/aarch64-poky-linux-g++   -I/home/leo.yan/volumes/OE-build/poky/builp
/home/leo.yan/volumes/OE-build/poky/build_arm64/tmp/work/aarch64-poky-linux/bpftrace/0.9.4-r0/git/src/ast/irbuilderbpf.cpp: In member function 'llvm::AllocaInst* bpftrace::ast::IRBuilderBPF::CreateAllocaBPFInit:
/home/leo.yan/volumes/OE-build/poky/build_arm64/tmp/work/aarch64-poky-linux/bpftrace/0.9.4-r0/git/src/ast/irbuilderbpf.cpp:85:52: error: no matching function for call to 'bpftrace::ast::IRBuilderBPF::CreateMemS'
   85 |     CreateMemSet(alloca, getInt64(0), stype.size, 1);                                                                                                                                                      
      |                                                    ^                                                                                                                                                       
In file included from /home/leo.yan/volumes/OE-build/poky/build_arm64/tmp/work/aarch64-poky-linux/bpftrace/0.9.4-r0/git/src/ast/irbuilderbpf.h:9,                                                                  
                 from /home/leo.yan/volumes/OE-build/poky/build_arm64/tmp/work/aarch64-poky-linux/bpftrace/0.9.4-r0/git/src/ast/irbuilderbpf.cpp:3:                                                                
[...]

PR #1187, opened 2020-03-06
CreateMemSet: llvm10: Fix compilation errors

ovpanait:
The following llvm commit:
"[Alignment][NFC] CreateMemSet use MaybeAlign" (https://reviews.llvm.org/D71213)
changed the signature of CreateMemSet() function, which causes the following
compilation errors:

...
src/ast/irbuilderbpf.cpp:85:52: error: no matching function for call to 'bpftrace::ast::IRBuilderBPF::CreateMemSet'
   85 |     CreateMemSet(alloca, getInt64(0), stype.size, 1);
      |
...

Use LLVM_VERSION_MAJOR macro to select the appropriate call.

Fixes: #1165

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -41,7 +41,7 @@ void CodegenLLVM::visit(PositionalParameter &param)
         {
           Constant *const_str = ConstantDataArray::getString(module_->getContext(), pstr, true);
           AllocaInst *buf = b_.CreateAllocaBPF(ArrayType::get(b_.getInt8Ty(), pstr.length() + 1), "str");
-          b_.CreateMemSet(buf, b_.getInt8(0), pstr.length() + 1, 1);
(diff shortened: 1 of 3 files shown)


fbs:
In irbuilder.h we've defined a few CREATE_MEMCPY macros to handle LLVM version specific things. Doing it like that avoids having a ton of ifdefs throughout the code.

Also, I thought that llvm deprecated the legacy pass manager with version 10, or does it still work? :)

ovpanait:
Hi,

Thanks for the feedback! I have updated the patch and added the CREATE_MEMSET macro in src/ast/irbuilderbpf.h.

As for the the legacy pass manager deprecation, sorry but I am not really familiar with the llvm api. On my side it builds fine with these changes only.

danobi:
Please see the clang-format failures in the build. Otherwise LGTM

ovpanait:
Updated the patch as per clang-format test indication.

fbs:
nice. Thanks!

PR #1189, opened 2020-03-07
Type fixes 2

fbs:
Follow up of #1159

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -629,10 +629,10 @@ void CodegenLLVM::visit(Call &call)
       abort();
     }
 
-    expr_ = b_.CreateLoad(
-        b_.getInt64Ty(),
-        b_.CreateGEP(ctx_, b_.getInt64(offset * sizeof(uintptr_t))),
-        call.func+"_"+reg_name);
+    Value *ctx = b_.CreatePointerCast(ctx_, b_.getInt64Ty()->getPointerTo());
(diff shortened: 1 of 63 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -1313,7 +1313,9 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
   Value *val, *expr;
   expr = expr_;
   AllocaInst *key = getMapKey(map);
-  if (map.type.type == Type::string || assignment.expr->type.type == Type::inet)
+  if (map.type.type == Type::string ||
+      assignment.expr->type.type == Type::inet ||
+      assignment.expr->type.type == Type::usym)
```
danobi: Just to double check: this is ok b/c the usym is already on the stack, right?

danobi:
Looks like you have some unused variable warnings (`assert`s compiled out in release mode) and LLVM api changes to ifdef over, though

src/ast/codegen_llvm.cpp
fbs: Yes.

Found a different bug though  `:f { @=usym(arg0); @a=@ }'` results in a segv. I've added a commit to fix it :)

fbs:
@danobi  I've (hopefully) fixed the tests and fixed a copy bug around usym. Can you take another look?

danobi:
LGTM

fbs:
I'll merge this monday if there are no further comments.

PR #1190, opened 2020-03-08
Add llvm 10 test

mmisono:


mmisono:
This may conflict #1189.

mmisono:
Should cache regex patterns. cf. #1192

danobi:
Seems reasonable to me. What do you think @fbs?

fbs:
Look OK :)

mmisono:
I'll wait for #1189 before fixing regex patterns. It might eliminate differences in some test cases. Resolving conflicts (if any) will be easier also.

fbs:
@mmisono  can you rebase this on amster?

mmisono:
@fbs 
Is it OK to not wait for #1194? (not sure if it conflicts)

fbs:
> @fbs
> Is it OK to not wait for #1194? (not sure if it conflicts)

I don't think it will conflict but if it does I can easily fix them.  Besides that one has no reviews yet and this one is mostly good to go.

mmisono:
OK, rebased. Changes were little (only changed the type of `probe_read` in *.ll files).
I also updated the code so that it caches regex patterns.

PR #1191, opened 2020-03-08
tests: cmake: Fix build with ninja

ovpanait:
Fix the following error when trying to build with ninja:
$ mkdir build
$ cd build
$ cmake .. -GNinja
$ ninja
ninja: error: '../tests/codegen/*.cpp', needed by 'tests/codegen_includes.cpp', missing and no known rule to make it


danobi:
Nice, thanks.

PR #1192, opened 2020-03-08
Slowfix

fbs:
Avoid calling "slow" regex constructor

On my ubuntu VM it took ~17 seconds before `tcpconnect.bt` was compiled,
most of the time spent in the newly added `remove)qualifiers`:

```
Time spent in remove_qualifiers: 17502 ms
Time spent in remove_qualifiers: 16662 ms
```

Which is mostly caused by the regex constructor being called 27k times.

With this patch the average of 10 runs is down to: 575ms

This fixes a regression introduced by #1177

----

The second commit optimizes the regex a bit, shaving off ~70ms

---

@mmisono  can you take a look at this one? I tested it by adding:

```
std::string x = std::regex_replace(typestr, re, "");
std::cout << x << std::endl;
return x;
```

The output for the original and mine are equal for tcpconnect.bt

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -206,10 +206,9 @@ static std::string remove_qualifiers(std::string &&typestr)
 {
   // libclang prints "const" keyword first
   // https://github.com/llvm-mirror/clang/blob/65acf43270ea2894dffa0d0b292b92402f80c8cb/lib/AST/TypePrinter.cpp#L137-L157
-  return std::regex_replace(typestr,
-                            std::regex("^(const volatile\\s+)|^(const\\s+)|"
-                                       "^(volatile\\s+)|\\*(\\s*restrict)$"),
-                            "");
+  static std::regex re("^(const volatile\\s+)|^(const\\s+)|"
+                       "^(volatile\\s+)|\\*(\\s*restrict)$");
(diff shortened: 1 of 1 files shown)


mmisono:
Initially, I implemented this feature by traversing AST: https://github.com/mmisono/bpftrace/commit/e6a079ba55110255adc273c8215fed957d866ddf
But I changed it to use regex because I thought it was more straightforward and I didn't know well about clang parser (I think the above implementation is OK, but I'm not completely sure).

In my environment, the times spent in `ClangParser::parse()` for `tcp_connect.bt` are (averages in 10 runs):
- 921ms (regex version (this PR version))
- 672ms (https://github.com/mmisono/bpftrace/commit/e6a079ba55110255adc273c8215fed957d866ddf version)

The speed of the latter version is almost the same as before. Considering the performance, I think it's worth to use the latter implementation.
 
Could you take a look the above patch? @fbs 
If it is desirable, I'll create a PR.

------

(By the way, when using BTF, the parse times are around 30ms and 21ms respectively.)

fbs:
> Could you take a look the above patch? @fbs
> If it is desirable, I'll create a PR.

That looks like a better approach.

I'll update this one to just factor out the regex and fix the regression for now, we can use yours to speed it up further.

PR #1194, opened 2020-03-09
codegen: Send map id instead of ident string for async events

### LINKED ISSUE #1188: `print()` uses strings
```
i:s:1 { 
@thisisawaytolongstringname=1; 
print(@thisisawaytolongstringname)
}
```

compiles to:

```
13: (b7) r1 = 30001
14: (7b) *(u64 *)(r10 -56) = r1
15: (18) r1 = 0x6173697369687440
17: (7b) *(u64 *)(r10 -32) = r1
18: (18) r1 = 0x6e6f6c6f74796177
20: (7b) *(u64 *)(r10 -24) = r1
21: (18) r1 = 0x6e676e6972747367
23: (7b) *(u64 *)(r10 -16) = r1
24: (b7) r1 = 6647137
25: (63) *(u32 *)(r10 -8) = r1
26: (7b) *(u64 *)(r10 -40) = r7
27: (7b) *(u64 *)(r10 -48) = r7
28: (18) r7 = 0xffff97eabbb55800
30: (85) call bpf_get_smp_processor_id#8
31: (bf) r4 = r10
32: (07) r4 += -56
33: (bf) r1 = r6
34: (bf) r2 = r7
35: (bf) r3 = r0
36: (b7) r5 = 52
37: (85) call bpf_perf_event_output#25
```

```
>>> "6173697369687440".decode("hex")
'asisiht@'
```

Instead we should just store an integer map identifier, like  we do with asyncid. 

I'll fix this as part of #1091

fbs:
We're currently sending the full map identifier as string to userspace.
This has a lot of overhead as we're sending a lot of potentially very
long strings around.

The approach taking here is to assign each map a unique uint32_t id
during creation. Which can then be used to send events from bpf to user
space.
To make the whole thing llvm-type-safe and remove the confusing pointer
math it now uses structs instead of arrays on both the bpf and the
user space side.

To avoid having to keep track of the "LLVM struct" and the "C struct"
in multiple files and accidentally introduce type issues they've been
defined in a single place. While there is no checking on them, having
them close together makes spotting errors easier.

Fixes #1188

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1,6 +1,7 @@
 #include "codegen_llvm.h"
 #include "arch/arch.h"
 #include "ast.h"
+#include "ast/async_event_types.h"
 #include "bpforc.h"
 #include "parser.tab.hh"
 #include "tracepoint_format_parser.h"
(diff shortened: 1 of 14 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -1779,11 +1783,20 @@ int SemanticAnalyser::create_maps(bool debug)
         Integer &min = static_cast<Integer&>(min_arg);
         Integer &max = static_cast<Integer&>(max_arg);
         Integer &step = static_cast<Integer&>(step_arg);
-        bpftrace_.maps_[map_name] = std::make_unique<bpftrace::Map>(map_name, type, key, min.n, max.n, step.n, bpftrace_.mapmax_);
+        bpftrace_.maps_[map_name] = std::make_unique<bpftrace::Map>(
+            map_name, type, key, min.n, max.n, step.n, bpftrace_.mapmax_);
+        bpftrace_.maps_[map_name]->id = bpftrace_.map_ids_.size();
+        bpftrace_.map_ids_.push_back(map_name);
```
fbs: I'm not really happy with this. I think the long term fix should be to move the map storage out of bpftrace, for now this works ok

src/ast/codegen_llvm.cpp
```diff
@@ -655,87 +657,86 @@ void CodegenLLVM::visit(Call &call)
...
+    b_.CreatePerfEventOutput(ctx_, buf, AsyncEvent::Print().LLVMTypeSize());
+    b_.CreateLifetimeEnd(buf);
     expr_ = nullptr;
   }
   else if (call.func == "clear" || call.func == "zero")
   {
+    auto elements = AsyncEvent::MapEvent().asLLVMType(&b_);
+    StructType *print_struct = b_.GetStructType(call.func + "_t",
+                                                elements,
+                                                false);
```
fbs: these structs should be packed to be able to use `layout_->getAllocSize()` instead of hardcoding the size

src/async_event_types.h
```diff
@@ -0,0 +1,78 @@
...
+  {
+    return {
+      b->getInt64Ty(), // asyncid
+      b->getInt32Ty(), // map id
+    };
+  }
+
+  uint64_t LLVMTypeSize(void)
+  {
+    return 12;
```
fbs: should ensure the struct is packed and use datalayout here

src/ast/codegen_llvm.cpp
```diff
@@ -661,87 +663,92 @@ void CodegenLLVM::visit(Call &call)
...
+      if (call.vargs->size() > i)
+      {
+        call.vargs->at(i)->accept(*this);
+        val = b_.CreateIntCast(expr_, elements.at(i), false);
+      }
+      else
+        val = b_.GetIntSameSize(0, elements.at(i));
+
+      b_.CreateStore(val,
+                     b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(i + 1) }));
```
danobi, line 695: Might be a personal preference, but this was a little confusing for me to read. Might be more clear if there was 2 loops with something like:

```
size_t i;
for (i = 1; i < call.vargs.size(); i++)
{
  <accept real value and use it>
}

// Fill in 0 for remaining
for (; i < 3; i++)
{
  <use 0>
}
```

src/ast/codegen_llvm.cpp
```diff
@@ -661,87 +663,92 @@ void CodegenLLVM::visit(Call &call)
...
-     * The asyncaction_id informs user-space that this is not a printf(), but is a
-     * special asynchronous action. The ID maps to print(). The top argument is either
-     * a value for truncation, or 0 for everything. The div argument divides the output values
-     * by this (eg: for use in nanosecond -> millisecond conversions).
-     * TODO: consider stashing top & div in a printf_args_ like struct, so we don't need to pass
-     * them here via the perfdata output (which is a little more wasteful than need be: I'm using
-     * uint64_t's to avoid "misaligned stack access off" errors when juggling uint32_t's).
-     */
+    auto elements = AsyncEvent::Print().asLLVMType(&b_);
+    StructType *print_struct = b_.GetStructType("print_t", elements, true);
```
danobi, line 667: Maybe do same as below?
```suggestion
    StructType *print_struct = b_.GetStructType(call.func + "_t", elements, true);
```

src/bpftrace.h
```diff
@@ -130,6 +134,9 @@ class BPFtrace
   static volatile sig_atomic_t exitsig_recv;
 
   std::map<std::string, std::unique_ptr<IMap>> maps_;
+  //
```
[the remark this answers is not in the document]
fbs: No i actually wanted to place a comment here, just forgot to actually write it :p. Wanted to avoid adding yet-another-undocumented-variable to bpftrace.h

fbs:
Addressed your comments @danobi

src/bpftrace.h
```diff
@@ -130,6 +134,10 @@ class BPFtrace
   static volatile sig_atomic_t exitsig_recv;
 
   std::map<std::string, std::unique_ptr<IMap>> maps_;
+
+  // Maps a map id back to the map identifier. See get_map_by_id(
```
danobi, line 138: Typo?
```suggestion
  // Maps a map id back to the map identifier. See get_map_by_id()
```

danobi:
Also clang-format

PR #1195, opened 2020-03-10
Strengthen tracepoint format parsing

olsajiri:
There's issue in current RHEL real time kernel with tracepoint format,
which makes bpftrace to return wrong data.

There's a 'not described gap' in the sched_wakeup's format file and
probably in other formats as well:

```
  # cat /sys/kernel/debug/tracing/events/sched/sched_wakeup/format
  name: sched_wakeup
  ID: 310
  format:
          field:unsigned short common_type;       offset:0;       size:2; signed:0;
          field:unsigned char common_flags;       offset:2;       size:1; signed:0;
          field:unsigned char common_preempt_count;       offset:3;       size:1; signed:0;
          field:int common_pid;   offset:4;       size:4; signed:1;
          field:unsigned char common_migrate_disable;     offset:8;       size:1; signed:0;
          field:unsigned char common_preempt_lazy_count;  offset:9;       size:1; signed:0;

          field:char comm[16];    offset:12;      size:16;        signed:1;
          field:pid_t pid;        offset:28;      size:4; signed:1;
          field:int prio; offset:32;      size:4; signed:1;
          field:int success;      offset:36;      size:4; signed:1;
          field:int target_cpu;   offset:40;      size:4; signed:1;


```
There's "common_preempt_lazy_count" field on offset 9 with size 1:
`        common_preempt_lazy_count;  offset:9;       size:1;`

and it's followed by "comm" field on offset 12:
`        field:char comm[16];    offset:12;      size:16;        signed:1;`

which makes 2 bytes gap in between, that might confuse some applications
like bpftrace.

The tracepoint parser makes struct out of the field descriptions,
but does not account for such gaps.

I posted patch to fix this [1] in RT kernel, but that might take a while,
and we could easily fix our tracepoint parser to workaround this issue.

Adding code to detect this gaps and add 1 byte `__pad_X` fields, where X is
the offset number.

The test code for the parser uses zero offset in tests, which will
not happen in real life, but we can live with that and limit the
gap generation only when offset is defined.

Adding new test for `sched:sched_wakeup` with the gap problem.

[1] [https://lore.kernel.org/linux-rt-users/20200221153541.681468-1-jolsa@kernel.org/](https://lore.kernel.org/linux-rt-users/20200221153541.681468-1-jolsa@kernel.org/)

### DIFF
--- a/src/tracepoint_format_parser.cpp
+++ b/src/tracepoint_format_parser.cpp
@@ -116,8 +116,11 @@ std::string TracepointFormatParser::get_struct_name(const std::string &category,
   return "struct _tracepoint_" + category + "_" + event_name;
 }
 
-std::string TracepointFormatParser::parse_field(const std::string &line)
+std::string TracepointFormatParser::parse_field(const std::string &line,
+                                                int *last_offset)
 {
+  std::string extra = "";
+
   auto field_pos = line.find("field:");
   if (field_pos == std::string::npos)
     return "";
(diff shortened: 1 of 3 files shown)


fbs:
Just took a quick glance, didn't test this yet but does this show up in  tracepoint listing too (`-lv`)? If so, should we hide it as it is not a real field you can attach to?

olsajiri:
> Just took a quick glance, didn't test this yet but does this show up in tracepoint listing too (`-lv`)? If so, should we hide it as it is not a real field you can attach to?

I was wondering that too, but we have separate code for that,
so it's not displayed

PR #1198, opened 2020-03-12
Add snapcraft yaml to build bpftrace snap

ColinIanKing:
Add the snapcraft yaml rules to allow bpftrace to be built as
a snap.

If one is building using multipass, one should enable a large
multipass VM using:

SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=8G snapcraft

Alternatively, one can build in a lxd container using:

snapcraft --use-lxd

..and install using something like:

sudo snap install --devmode bpftrace_0.9.4-20200312-1403-11814f2_amd64.snap

Signed-off-by: Colin Ian King <colin.king@canonical.com>

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,122 @@
+#
+# To build:
+#
+# Using multipass, one should enable a large multipass VM using:
+#
+#   SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=8G snapcraft
+#
+# Alternatively, one can build in a lxd container using:
+#
+#   snapcraft --use-lxd
+#
+# Install using something like:
+#
+#   sudo snap install --devmode bpftrace_0.9.4-20200312-1403-11814f2_amd64.snap
+#
+name: bpftrace
+summary: high-level tracing language for Linux eBPF
+description: BPFtrace is a high-level tracing language for
+  Linux enhanced Berkeley Packet Filter available in recent
+  Linux kernels. BPFtrace uses LLVM as a backend to compile scripts
+  to BPF-bytecode and makes use of BCC for interacting with the Linux
+  BPF system, as well as existing Linux tracing capabilities like
+  kernel dynamic tracing, user-level dynamic tracing and tracepoints.
+
+confinement: strict
+grade: stable
+base: core18
+architectures: [amd64]
+assumes: [snapd2.37]
+adopt-info: bpftrace
+
+parts:
+    bcclibs3:
+        source: https://github.com/iovisor/bcc.git
+        source-tag: v0.8.0
+        plugin: cmake
+        configflags:
+            - -DLLVM_DEFINITIONS="-D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS"
+            - -DREVISION_LAST=v0.8.0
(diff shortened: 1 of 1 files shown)


danobi:
So bpftrace is licensed as apache 2. I'm not sure how license mixing works but I think it'd be better if we avoided it.

Can you change the copyright assignment?

ColinIanKing:
I've removed the copyright now.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,107 @@
...
+          - binutils-dev
+          - llvm-7-dev
+          - llvm-7-runtime
+          - libclang-7-dev
+          - clang-7
+        stage-packages:
+          - libbinutils
+          - libclang1-7
+          - libllvm6.0
+          - libllvm7
```
fbs: Does it need both 6 & 7?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,107 @@
...
+        source: https://github.com/iovisor/bpftrace.git
+        plugin: cmake
+        configflags:
+            - -DCMAKE_BUILD_TYPE=Release
+            - -DLIBBCC_INCLUDE_DIRS=$SNAPCRAFT_STAGE/usr/include/bcc
+        override-pull: |
+            snapcraftctl pull
+            version=$(git tag | sort -V | tail -1 | cut -c2-)
+            commits=$(git log --oneline | wc -l)
+            sha=$(git log -1 --oneline | cut -d' ' -f1)
```
fbs: we use `git describe` to determine the version number for our git-builds, it does most of this for you:

```
$ git describe head --tags
v0.9.4-83-g11814f2
```

fbs:
very nice, thank you!

snap/snapcraft.yaml
ColinIanKing: I've updated the commit to now use just use version 7

snap/snapcraft.yaml
ColinIanKing: I've updated the commit to use git describe HEAD --tags

ColinIanKing:
I've refreshed the commit to address the 2 pending queries.

ColinIanKing:
ping. This patch is now refreshed. Is it OK to merge?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,122 @@
...
+grade: stable
+base: core18
+architectures: [amd64]
+assumes: [snapd2.37]
+adopt-info: bpftrace
+
+parts:
+    bcclibs3:
+        source: https://github.com/iovisor/bcc.git
+        source-tag: v0.8.0
```
fbs: is there a specific reason this is still 0.8.0? Or can we safely upgrade it to 0.12 later?

ColinIanKing: I used 0.8.0 as this was the version I had verfied to word for a build using the versions of the support libraries used in the snap. I suspect it can be updated to 0.12 or later if it can be verified to work fine.

fbs:
@ColinIanKing  LGTM. I will look into adding this to the CI.  Would you be ok with us "taking over" the /bpftrace name in the store?

ColinIanKing:
I'm very happy for you to take over the bpftrace name in the store, I'm not sure how one passes the ownership over though.

fbs:
> I'm very happy for you to take over the bpftrace name in the store, I'm not sure how one passes the ownership over though.

Cool! I'll look into the options and will reach out to you.

PR #1200, opened 2020-03-13
Update minimum required version of CMake to 3.8

### LINKED ISSUE #1193: cmake errors: Target "bpftrace" requires the language dialect "CXX17" , but CMake does not know the compile flags to use to enable it.
probably caused by https://github.com/iovisor/bpftrace/pull/1184

My environment:

* Ubuntu 16.04
* cmake 3.5.1

mmisono:
CMake 3.8 is required to use C++17.
cf. https://github.com/Kitware/CMake/commit/ae1a6815b6e2c0a45df51e994b75f9005e1794fe

closes #1193

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.8)
 project(bpftrace)
 
 # bpftrace version number components.


mmisono:
It seems Ubuntu 16.04 has CMake 3.5 by default. So users will need to update CMake to build.
alpine 3.8 has CMake 3.11.

danobi:
I think @fbs brought up downgrading back to C++14 in IRC. Maybe we should do that for this coming release, then go back up to C++17 and add this change too.

What do you think?

Also, please come join the IRC channel :)

fbs:
> I think @fbs brought up downgrading back to C++14 in IRC. Maybe we should do that for this coming release, then go back up to C++17 and add this change too.

My idea was to do the C++17 right after the 0.9.5 release. That way we can make this the last 16.04 release instead of dropping it half way through without using any features of it yet (like  we did now). They can still profit from any improvements that land in 0.9.5 but one that is release 16.04 will be ~4y old so dropping support sounds ok to me

I will add this to the next milestone so we can do it right after we've released

> Also, please come join the IRC channel :)

+1

PR #1203, opened 2020-03-13
parser: Do not remove empty probe arguments

danobi:
There was no reason to remove empty probe arguments. There wasn't a
single probe that needed this behavior.

Also add a regression test.


mmisono:
Trailing colon is ignored now: 
```
% sudo ./src/bpftrace -e 'k:f: {}'
Attaching 1 probe...
```

I think if input is `k:f:` then `parts_` sohuld be `{ "k", "f", ""}`.

danobi:
Updated

danobi:
Going to merge tomorrow if no other comments

PR #1205, opened 2020-03-14
Error if wildcards are used in "category" of tracepoint

mmisono:
Currently, TracepointFormatParser::parse() does not expand wildcards in "category". This leads to an incorrect event name.
https://github.com/iovisor/bpftrace/blob/24f97308895078fe0130a58e0cf5703f4de975a9/src/tracepoint_format_parser.cpp#L73-L75

Check if wildcards are used in "category" and return an error in that case.
For example:

```
% sudo ./src/bpftrace -e 'tracepoint:sys*:*{args;}'
ERROR: wildcard in tracepoint category is not supported: sys*
```

### DIFF
--- a/src/tracepoint_format_parser.cpp
+++ b/src/tracepoint_format_parser.cpp
@@ -41,6 +41,14 @@ bool TracepointFormatParser::parse(ast::Program *program)
         std::string format_file_path = "/sys/kernel/debug/tracing/events/" + category + "/" + event_name + "/format";
         glob_t glob_result;
 
+        if (has_wildcard(category))
+        {
+          std::cerr
+              << "ERROR: wildcards in tracepoint category is not supported: "
+              << category << std::endl;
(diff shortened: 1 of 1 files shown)


mmisono:
I found this when fuzzing bpftrace with [AFL](https://github.com/google/AFL).

```
% sudo ./src/bpftrace -e 'tracepoint:****************:*{args;}'
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 50) > this->size() (which is 49)
zsh: abort      sudo ./src/bpftrace -e 'tracepoint:****************:*{args;}'
```

danobi:
> I found this when fuzzing bpftrace with AFL.

Cool, could you share details on your setup?

mmisono:
> could you share details on your setup?

sure, I'll write up the details later.

PR #1206, opened 2020-03-14
Add apt-transport-https for xenial build

dalehamel:
Master builds have been failing on xenial because of missing https transport.

This appears to resolve the issue


dalehamel:
@fbs @danobi can you take a look please

PR #1208, opened 2020-03-15
Don't check cgroupid string argument length

### LINKED ISSUE #1207: cgroupid string argument length is checked and rejected if too long
Currently, the length of the argument to [`cgroupid()`](https://github.com/iovisor/bpftrace/blob/8bcf0c04063f1b7855ee929ebf4e76a85d6eddf4/docs/reference_guide.md#13-cgroupid-resolve-cgroup-id) is checked and rejected if over 64 characters long by default.

To reproduce from a checkout of master (currently at 8bcf0c04):

```
vagrant up ubuntu-19.10
vagrant ssh ubuntu-19.10 -c 'cmake /vagrant -DCMAKE_INSTALL_PREFIX=/usr/local && make && sudo make install'
vagrant ssh ubuntu-19.10 -c 'sudo bpftrace -e "tracepoint:sched:sched_process_exec /cgroup == cgroupid(\"/sys/fs/cgroup/unified/user.slice/user-1000.slice/session-1.scope\")/ { printf(\"msg\"); }"'
```

the result is

```
stdin:1:48-124: ERROR: String is too long (over 64 bytes): /sys/fs/cgroup/unified/user.slice/user-1000.slice/session-1.scope
tracepoint:sched:sched_process_exec /cgroup == cgroupid("/sys/fs/cgroup/unified/user.slice/user-1000.slice/session-1.scope")/ { printf("msg"); }
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Since the `cgroupid("...")` expression is evaluated to the int cgroup id when we're parsing the script, and the actual string itself is not used at runtime, I would expect that the length of the argument should not be checked.

chrahunt:
Since `cgroupid` is evaluated at compile-time, we don't need to check its arguments length.

Now when analyzing function calls, we keep track of the inner-most function and use it to decide whether to check the string argument length, skipping the check for compile-time functions. This just applies to `cgroupid` for now.

We can extend this to `kaddr` and `uaddr` with a few more lines. If this approach looks OK I can add them here or in another PR.

Fixes #1207.

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -96,7 +96,8 @@ void SemanticAnalyser::visit(PositionalParameter &param)
 
 void SemanticAnalyser::visit(String &string)
 {
-  if (string.str.size() > STRING_SIZE-1) {
+  if (!is_compile_time_func(func_) && string.str.size() > STRING_SIZE - 1)
+  {
     ERR("String is too long (over " << STRING_SIZE << " bytes): " << string.str,
         string.loc);
   }
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -332,6 +332,26 @@ void SemanticAnalyser::visit(Call &call)
           call.loc);
   }
 
+  struct func_setter
+  {
+    func_setter(SemanticAnalyser &analyser, const std::string &s)
+        : analyser_(analyser), old_func_(analyser_.func_)
+    {
+      analyser_.func_ = s;
+    }
+
+    ~func_setter()
+    {
+      analyser_.func_ = old_func_;
+    }
+
+  private:
+    SemanticAnalyser &analyser_;
+    std::string old_func_;
+  };
+
+  func_setter scope_bound_func_setter{ *this, call.func };
+
   if (call.vargs) {
     for (Expression *expr : *call.vargs) {
       expr->accept(*this);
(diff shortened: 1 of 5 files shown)


fbs:
Thank you for working on this!

Wonder if it wouldn't be better to just skip the `->accept` for `call`s that deal with literals? We currently have this at the top of call:

```
  if (call.vargs) {
    for (Expression *expr : *call.vargs) {
      expr->accept(*this);
    }
  }
```

But that would be quite a noisy change. Not sure which one is better

danobi:
> Wonder if it wouldn't be better to just skip the ->accept for calls that deal with literals?

IMO it'd be better to always accept AST nodes in case we need to do fixups laters (eg modify the STRING node in semantic analyser later). It's also nice to have all the properties of a node stored in the `::visit(...)` method so all the properties are in one place, even if that means storing a little more state.

src/ast/semantic_analyser.cpp
```diff
@@ -332,6 +333,26 @@ void SemanticAnalyser::visit(Call &call)
           call.loc);
   }
 
+  struct func_setter
```
danobi: This is fine for now, but if we use this pattern later anywhere else it'd be nice to pull it into util.h and make a generic RAII helper.

chrahunt: Agreed. A few more examples will make that easier.

PR #1217, opened 2020-03-17
cmake: Control manpages generation

ovpanait:
Introduce ENABLE_MAN cmake option to control manpages generation. It is enabled
by default.


fbs:
Looks ok to me. Just curious, what is your usecase for this? :)

ovpanait:
I am working on integrating bcc/bpftrace in Yocto (which is a build system for creating distributions for embedded systems) and we like to have fine-grained control over what gets built/installed :) 

By default manpages are not needed on embedded systems, so it would be great to have a standard way to control manpage generation/installation.

Issue #1091, opened 2020-01-21
llvm type issues


fbs:
We have quite a few typing issues in our codegen, mostly around pointer types not matching:

```
llvm-as-7: ../buf-opt-bad.ll:20:16: error: explicit pointee type doesn't match operand's pointee type
  %arg0 = load i64, i8* %1, align 8
```

While it works in most cases these do cause issues with LLVM in edge cases leading to out-of-memory issues (#711) or segmentation faults:

```
#include <linux/skbuff.h>
#include <linux/icmp.h>

k:icmp_echo {
        $skb = (struct sk_buff *)arg0;
        $sk = $skb->sk;
        $hdr = (struct icmphdr *) $skb->transport_header;
        @=$hdr->un.echo.sequence;
}
```
(not entirely sure this example is actually an ir issue, but it crashes in the codegen and fixing the IR manually seems to make llc happy)

I've got a few changes ready that should fix most of them. Once they're merged I want to add a new test step which uses `llvm-as` to verify  the generated IR. 

TODO:

- [x] Merge #1194
- [x] Merge #1159
- [x] Merge #1189 
- [x] Fix printf issues, #873 
- [x] Test IR
- [x] Check issue list of issues that likely are fixed
   - [x]  #875
   - [x] #1173
   - [x] #1502 
   - [x] #1501
   - [x] #873
   - [x] #1173

fbs:
#1736 \o/

fbs:
all linked tickets seem fixed, time to finally close this one

PR #1226, opened 2020-03-20
Don't over-read usdt arguments

danobi:
bpftrace internally represents USDT arguments as 64 bit types. However,
actual USDT arguments can be less than that. This commit ensures we only
read the actual size while zero-ing out unused bits.

See runtime test for where the previous behavior can go wrong.

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -386,15 +386,29 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx, struct bcc_usdt_argument
       abort();
     }
 
+    // Argument size must be 1, 2, 4, or 8. See
+    // https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation
+    int abs_size = std::abs(argument->size);
+    assert(abs_size == 1 || abs_size == 2 || abs_size == 4 || abs_size == 8);
+
+    // bpftrace's args are internally represented as 64 bit integers. However,
+    // the underlying argument (of the target program) may be less than 64
+    // bits. So we must be careful to zero out unused bits.
     Value* reg = CreateGEP(ctx, getInt64(offset * sizeof(uintptr_t)), "load_register");
     AllocaInst *dst = CreateAllocaBPF(builtin.type, builtin.ident);
-    CreateProbeRead(dst, builtin.type.size, reg);
-    result = CreateLoad(dst);
     if (argument->valid & BCC_USDT_ARGUMENT_DEREF_OFFSET) {
-      Value *ptr = CreateAdd(
-          result,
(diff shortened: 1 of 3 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -362,15 +362,30 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx, struct bcc_usdt_argument
...
     Value* reg = CreateGEP(ctx, getInt64(offset * sizeof(uintptr_t)), "load_register");
     AllocaInst *dst = CreateAllocaBPF(builtin.type, builtin.ident);
-    CreateProbeRead(dst, builtin.type.size, reg);
-    result = CreateLoad(dst);
     if (argument->valid & BCC_USDT_ARGUMENT_DEREF_OFFSET) {
+      // If we are dereferencing a pointer for the offset, we want to
+      // read native pointer size (assumed to be 64-bit here)
+      assert(builtin.type.size == 8);
+      CreateProbeRead(dst, builtin.type.size, reg);
+      result = CreateLoad(dst);
```
mmisono: I think it is possible to load reg directly because it's in context memory.

```suggestion
      result = CreateLoad(getInt64Ty(), reg)
```

src/ast/irbuilderbpf.cpp
```diff
@@ -362,15 +362,30 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx, struct bcc_usdt_argument
...
       Value *ptr = CreateAdd(
           result,
           getInt64(argument->deref_offset));
-      CreateProbeRead(dst, builtin.type.size, ptr);
+      // Zero out `dst` here in case we read less than 64 bits
+      CreateStore(getInt64(0), dst);
+      CreateProbeRead(dst, argument->size, ptr);
```
mmisono: Signed types have [a negative value](https://sourceware.org/systemtap/wiki/UserSpaceProbeImplementation)

```suggestion
      CreateProbeRead(dst, std::abs(argument->size), ptr);
```

I think it would be better to check `argument->size` in case a binary contains an invalid value.

src/ast/irbuilderbpf.cpp
```diff
@@ -362,15 +362,30 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx, struct bcc_usdt_argument
...
+      result = CreateLoad(dst);
+    }
+    else
+    {
+      // Zero out `dst` in case we read less than 64 bits
+      CreateStore(getInt64(0), dst);
+      CreateProbeRead(dst, argument->size, reg);
```
mmisono: Same as the first.

fbs:
LGTM. Looks I missed this one in #1091 as it has some typing issues. I will look into them

PR #1232, opened 2020-03-24
Positional fixes

fbs:
A few fixes for integer positional parameters. With this it is now possible to use different bases (octal, hex) and support different numbers:

```
# bpftrace  -e 'BEGIN { printf("got: %d %d %d\n", $1, $2, $3); exit() }' -- -123 0775 0x123
got: -123 509 291
```

Where before it would error:

```
# bpftrace  -e 'BEGIN { printf("got: %d %d %d\n", $1, $2, $3); exit() }' -- -123 0775 0x123
stdin:1:35-37: ERROR: 1.35-36 $1 used numerically << but given "-123". Try using str($1).
BEGIN { printf("got: %d %d %d\n", $1, $2, $3); exit() }
                                  ~~
stdin:1:43-45: ERROR: 1.43-44 $3 used numerically << but given "0x123". Try using str($3).
BEGIN { printf("got: %d %d %d\n", $1, $2, $3); exit() }
                                          ~~
```

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -800,4 +800,30 @@ bool symbol_has_cpp_mangled_signature(const std::string &sym_name)
     return false;
 }
 
+pid_t parse_pid(const std::string &str)
+{
+  try
+  {
+    constexpr ssize_t pid_max = 4 * 1024 * 1024;
+    std::size_t idx = 0;
+    auto pid = std::stol(str, &idx, 10);
+    // Detect cases like `13ABC`
+    if (idx < str.size())
+      throw InvalidPIDException(str, "is not a valid decimal number");
+    if (pid < 1 || pid > pid_max)
+      throw InvalidPIDException(str,
+                                "out of valid pid range [1," +
+                                    std::to_string(pid_max) + "]");
+    return pid;
+  }
+  catch (const std::out_of_range &e)
+  {
+    throw InvalidPIDException(str, "outside of integer range");
+  }
+  catch (const std::invalid_argument &e)
+  {
+    throw InvalidPIDException(str, "is not a valid decimal number");
+  }
+}
+
 } // namespace bpftrace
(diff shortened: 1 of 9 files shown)


fbs:
Also improved the validation around the pid argument a bit. The change in `is_numeric` behaviour broke it somewhat.

tests/semantic_analyser.cpp
```diff
@@ -1139,6 +1139,14 @@ TEST(semantic_analyser, positional_parameters)
...
   test(bpftrace, "kprobe:f { printf(\"%d\", $#); }", 0);
   test(bpftrace, "kprobe:f { printf(\"%s\", str($#)); }", 10);
+
+  Driver driver(bpftrace);
+  test(driver, "k:f { $1 }", 0);
+  auto stmt = static_cast<ast::ExprStatement *>(
+      driver.root_->probes->at(0)->stmts->at(0));
+  auto pp = static_cast<ast::PositionalParameter *>(stmt->expr);
+  EXPECT_EQ(SizedType(Type::integer, 8, true), pp->type);
+  EXPECT_TRUE(pp->is_literal);
```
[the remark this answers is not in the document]
fbs: I think this is the best way to test that it actually behaves as we want. This also verifies that it is a signed 64 bits number which is hard to verify otherwise.

I agree it is a bit ugly.

fbs:
Thanks, updated :)

src/utils.h
```diff
@@ -131,6 +148,7 @@ std::string str_join(const std::vector<std::string> &list,
                      const std::string &delim);
 bool is_numeric(const std::string &str);
 bool symbol_has_cpp_mangled_signature(const std::string &sym_name);
+pid_t parse_pid(std::string str);
```
danobi, line 151:
```suggestion
pid_t parse_pid(const std::string& str);
```

src/utils.cpp
```diff
@@ -791,4 +800,30 @@ bool symbol_has_cpp_mangled_signature(const std::string &sym_name)
     return false;
 }
 
+pid_t parse_pid(std::string str)
```
danobi, line 803:
```suggestion
pid_t parse_pid(const std::string& str)
```

PR #1229, opened 2020-03-23
Check a type of condition of if and ternary

mmisono:
Some functions (e.g., `exit()`) return Type::none but if and ternary does not check it and that results in a NULL pointer dereference.

```
% sudo ./src/bpftrace -e 'k:f { exit() ? 0 : 0 }'
AddressSanitizer:DEADLYSIGNAL
=================================================================
==14010==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000850854 bp 0x7fffffffa450 sp 0x7fffffffa180 T0)
==14010==The signal is caused by a READ memory access.
==14010==Hint: address points to the zero page.
    #0 0x850853 in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Ternary&) /home/ubuntu/work/bpftrace/src/ast/codegen_llvm.cpp
    #1 0x861f29 in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Probe&) /home/ubuntu/work/bpftrace/src/ast/codegen_llvm.cpp:1506:13
    #2 0x865256 in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Program&) /home/ubuntu/work/bpftrace/src/ast/codegen_llvm.cpp:1597:12
    #3 0x86e34b in bpftrace::ast::CodegenLLVM::compile(bpftrace::DebugLevel, std::ostream&) /home/ubuntu/work/bpftrace/src/ast/codegen_llvm.cpp:2013:10
    #4 0x77623f in main /home/ubuntu/work/bpftrace/src/main.cpp:605:22
    #5 0x7fffee1451e2 in __libc_start_main /build/glibc-t7JzpG/glibc-2.30/csu/../csu/libc-start.c:308:16
    #6 0x4d1bcd in _start (/disk/work/bpftrace/build_san3/src/bpftrace+0x4d1bcd)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/ubuntu/work/bpftrace/src/ast/codegen_llvm.cpp in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Ternary&)
==1401
```

Fix this by checking a type of cond in the semantic analysis.

PR #1222, opened 2020-03-18
Fix ternary comparison

mmisono:
`CreateICmpNE()` in a ternary unconditionally uses Int64Ty and cannot
use with other types. For example:

```
% sudo ./src/bpftrace -e 'k:f {!nsecs ? 0: 1;}'
bpftrace: /usr/lib/llvm-9/include/llvm/IR/Instructions.h:1151: void llvm::ICmpInst::AssertOK(): Assertion `getOperand(0)->getType() == getOperand(1)->getType() && "Both operands to ICmp instruction are not of the same type!"' failed.
```

The type of `!nsecs` is Integer whose bit width is 1.
Fix this by using the same type of the operand.

PR #1234, opened 2020-03-25
Fix if comparison

mmisono:
if_block compares condition with a value of int64ty and it cannot be
used with other types. Fix this by using the same type of the cond.

Error example:

```
% sudo ./src/bpftrace -e 'k:f { if ( (int32)pid ) { 1 } }'
bpftrace: /usr/lib/llvm-10/include/llvm/IR/Instructions.h:1163: void llvm::ICmpInst::AssertOK(): Assertion `getOperand(0)->getType() == getOperand(1)->getType() && "Both operands to ICmp instruction are not of the same type!"' failed.
zsh: abort      sudo ./src/bpftrace -e 'k:f { if ( (int32)pid ) { 1 } }'
```

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1420,7 +1420,10 @@ void CodegenLLVM::visit(If &if_block)
   if_block.cond->accept(*this);
   Value *cond = expr_;
 
-  b_.CreateCondBr(b_.CreateICmpNE(cond, b_.getInt64(0), "true_cond"), if_true, if_false);
+  Value *zero_value = Constant::getNullValue(cond->getType());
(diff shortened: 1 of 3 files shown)


mmisono:
Similar to #1222 (should have noticed when making that PR)

mmisono:
Predicate casts expr_ before comparison, so it does not have the same problem.
https://github.com/iovisor/bpftrace/blob/23603bfca771c21e9a5ed61cda8478ec41df87b0/src/ast/codegen_llvm.cpp#L1455-L1457

fbs:
are you finding all these by fuzzing? Was bumping into the same things testing #1231 

Can you add a regression test? Our test coverage for all these kind of things is quite low

mmisono:
Added some tests. I'll add tests for #1229 after this is merged.

PR #1236, opened 2020-03-27
Make `exit()` terminate current probe

fbs:
Currently `exit()` emits an async event but does not terminate further
probe processing. E.g. the following prints 2 which can confuse users:

```
$ bpftrace -e 'i:ms:100 { @=1; exit(); @=2 }'
@: 2
```

or a program that appears to conditionally exit but still writes to a
map:

```
$ bpftrace -e 'i:ms:100 { if(1){exit();} @=11 }'
@: 11
```

Due to it's asynchronous behaviour `exit()` is never fully reliable. A
probe might fire again before the runtime has had time to unload a
probe but this should improve the situation a bit.

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -658,6 +658,15 @@ void CodegenLLVM::visit(Call &call)
     b_.CreatePerfEventOutput(ctx_, perfdata, sizeof(uint64_t));
     b_.CreateLifetimeEnd(perfdata);
     expr_ = nullptr;
+    b_.CreateRet(ConstantInt::get(module_->getContext(), APInt(64, 0)));
+
+    // create an unreachable basic block for all the "dead instructions" that
+    // may come after exit(). If we don't, LLVM will emit the instructions
+    // leading to a `unreachable insn` warning from the verifier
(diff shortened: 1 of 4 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -658,6 +658,15 @@ void CodegenLLVM::visit(Call &call)
...
     b_.CreateLifetimeEnd(perfdata);
     expr_ = nullptr;
+    b_.CreateRet(ConstantInt::get(module_->getContext(), APInt(64, 0)));
+
+    // create an unreachable basic block for all the "dead instructions" that
+    // may come after exit()
+    Function *parent = b_.GetInsertBlock()->getParent();
+    BasicBlock *unreach = BasicBlock::Create(module_->getContext(),
+                                             "unreach",
+                                             parent);
```
danobi, line 668: Could you explain why this is necessary in the comment?

fbs: updated it @danobi

fbs:
@olsajiri  @mmisono  any objections to this? I don't expect it will but it could break (or at least change) current scripts in subtle ways.

mmisono:
No, I think the change is reasonable.

PR #1239, opened 2020-04-01
cmake: Add a path /usr/lib64 for LIBBPF_LIBRARIES

Taeung:
The installed libbpf path can be /usr/lib64/
if $(uname -m) is x86_64 so add the path.

### DIFF
--- a/cmake/FindLibBpf.cmake
+++ b/cmake/FindLibBpf.cmake
@@ -27,6 +27,7 @@ find_library (LIBBPF_LIBRARIES
     bpf
   PATHS
     /usr/lib
+    /usr/lib64
     /usr/local/lib
     /opt/local/lib
     /sw/lib


fbs:
I think we should keep these paths consistent between all cmake files we have, would be confusing otherwise.

Looking at the cmake docs I noticed:

> If the global property FIND_LIBRARY_USE_LIB64_PATHS is set all search paths will be tested as normal, with “64/” appended, and with all matches of “lib/” replaced with “lib64/”. This property is automatically set for the platforms that are known to need it if at least one of the languages supported by the PROJECT command is enabled.

I wonder why this isn't set for your machine and what the best way is to set it. Not sure if unconditionally setting it is smart or that it just breaks more than it fixes. If it appends a path (i.e. `/usr/lib` -> `/usr/lib64  and /usr/lib`) it probably isnt worse than manually updating the path everywhere. I'd have to dig deeper into cmake to figure that out.

fbs:
I think we can drop those cmake hints completely and depend on default behaviour. Closing this

Issue #878, opened 2019-09-04
Treat pointers as integers


ajor:
I propose we stop using `Type::cast` for pointers and just make them all `Type::integer`. `Type::cast` (or even `Type::struct`) would still be used for holding actual structs and unions.

This would let us get rid of special cases around pointer-to-integer comparisons and make pointer arithmetic possible.

It would also reduce confusion about what a `Type::cast` is holding, so bugs where a non-pointer struct is treated as a pointer (of which we currently have a few) won't happen.

fbs:
While doing this we should also correctly store the pointee type instead of just the pointee size. A pointer itself should always be an 8 byte integer.

fbs:
Maybe we should take this a bit further and create a few basic types for our "IR", a bit like llvm does too.

- base type, 
- Integer,  size and sign
- a pointer type,  itself an 64b integer and a pointee type
- an array type,  type and element count field
- a struct type

`hist,lhist,sum,min,max,avg,stats` etc can be based on array and struct types.

This should stop the `SizedType` type from growing, it is getting quite big already:

```
  Type type;
  Type elem_type = Type::none
  size_t size;
  StackType stack_type;
  bool is_signed = false;
  std::string cast_type;
  bool is_internal = false;
  bool is_pointer = false;
  bool is_tparg = false;
  AddrSpace addrspace = AddrSpace::none; // new
  size_t pointee_size = 0;
```

ajor:
Sounds much cleaner! I'm definitely with you on all the bullet points. Just not sure about:
> `hist,lhist,sum,min,max,avg,stats` etc can be based on array and struct types.

I think they don't actually have anything in common with arrays/structs at the AST stage - wouldn't it make more sense to just give them types inherited from the base type?

fbs:
I'm going to give this a go

fbs:
Rough plan:

- [ ] Add and implement `SizedType` factory function, CreateInt64, CreateString etc to provider a simpler interface to creating types (without having to worry about which fields to set) (#1316)
- [ ] Create and implement accessor functions to access sizedtype fields instead of access them directly. This way we can error on accessing fields which don't make sense for a type (e.g. accessing pointee_type on an integer)
- [ ] Redefine the pointer type, store  the actual pointee_type instead of its size
- [ ] Deprecate type::cast for type::record  which holds structs and unions. (Cant do type::struct as struct is a C keyword). Pointer too struct will be a type::pointer holding a type::record
[...]

PR #1241, opened 2020-04-03
Bpftrace fixes for s390x

sumanthkorikkar:
The patch series provides register support for s390x. This also solves various issues that can occur in big-endian architectures.
1. bpftrace: Add s390x register support
2. bpftrace: Fix sarg support for s390x and other architectures
3. bpftrace: Fix type conversion in printf() builtin
4. bpftrace: Fix histogram output & stat output for big-endian systems
5. bpftrace: Add architecture specific testcase support
6. bpftrace: Introduce htonl, htons, ntohl, ntohs functions
7. bpftrace: Fix ntop builtin function
8. bpftrace: Fix variable assignment of 32 bit integers

### DIFF
--- a/src/arch/ppc64.cpp
+++ b/src/arch/ppc64.cpp
@@ -3,6 +3,11 @@
 #include <algorithm>
 #include <array>
 
+// For little endian 64 bit, sp + 32 + 8 regs save area + argX
+#define ARG0_STACK_LE 96
+// For big endian 64 bit, sp + 48 + 8 regs save area + argX
+#define ARG0_STACK_BE 112
+
 namespace bpftrace {
 namespace arch {
 
@@ -97,6 +102,15 @@ int pc_offset()
 int sp_offset()
 {
   return offset("r1");
+}
+
+int arg_stack_offset()
+{
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+  return ARG0_STACK_LE / 8;
+#else
+  return ARG0_STACK_BE / 8;
+#endif
 }
 
 std::string name()
--- a/src/arch/s390.cpp
+++ b/src/arch/s390.cpp
@@ -5,6 +5,9 @@
 
 #define REQ_REGISTERS 19
 #define ARG_REGISTERS 5
+// For s390x, r2-r6 registers are used as function arguments, then the extra
+// arguments can be found starting at sp+160
+#define ARG0_STACK 160
 
 namespace bpftrace {
 namespace arch {
@@ -74,6 +77,11 @@ int pc_offset()
 int sp_offset()
 {
   return offset("r15");
+}
+
+int arg_stack_offset()
+{
+  return ARG0_STACK / 8;
 }
 
 std::string name()
--- a/src/arch/s390.cpp
+++ b/src/arch/s390.cpp
@@ -1,0 +1,85 @@
+#include "arch.h"
+
+#include <algorithm>
+#include <array>
+
+#define REQ_REGISTERS 19
+#define ARG_REGISTERS 5
+
+namespace bpftrace {
+namespace arch {
+
+// clang-format off
+static std::array<std::string, REQ_REGISTERS> registers = {
+  // Breakpoint event address
+  "arg",
+  "pswmask",
+  // Instruction address
+  "pswaddr",
+  "r0",
+  "r1",
+  "r2",
+  "r3",
+  "r4",
+  "r5",
+  "r6",
+  "r7",
+  "r8",
+  "r9",
+  "r10",
+  "r11",
+  "r12",
+  "r13",
+  "r14",
+  "r15",
+};
+
+static std::array<std::string, ARG_REGISTERS> arg_registers = {
+  "r2",
+  "r3",
+  "r4",
+  "r5",
+  "r6",
+};
+// clang-format on
+
+int offset(std::string reg_name)
+{
+  auto it = find(registers.begin(), registers.end(), reg_name);
+  if (it == registers.end())
+    return -1;
+  return distance(registers.begin(), it);
+}
+
+int max_arg()
+{
+  return arg_registers.size() - 1;
+}
+
+int arg_offset(int arg_num)
+{
+  return offset(arg_registers.at(arg_num));
+}
+
+int ret_offset()
+{
+  return offset("r2");
+}
+
+int pc_offset()
+{
+  return offset("pswaddr");
+}
+
+int sp_offset()
+{
+  return offset("r15");
+}
+
+std::string name()
+{
+  return std::string("s390x");
+}
+
+} // namespace arch
+} // namespace bpftrace
--- a/tests/runtime/engine/parser.py
+++ b/tests/runtime/engine/parser.py
@@ -3,6 +3,7 @@
 from fnmatch import fnmatch
 from collections import namedtuple
 import os
+import platform
 
 from utils import ERROR_COLOR, NO_COLOR
 
@@ -15,7 +16,7 @@ class UnknownFieldError(Exception):
     pass
 
 
-TestStruct = namedtuple('TestStruct', 'name run expect timeout before after suite kernel requirement env')
+TestStruct = namedtuple('TestStruct', 'name run expect timeout before after suite kernel requirement env arch')
 
 
 class TestParser(object):
@@ -52,7 +53,8 @@ class TestParser(object):
                 if line == '\n' or line_num == len(lines):
                     if test_lines:
                         test_struct = TestParser.__read_test_struct(test_lines, test_suite)
-                        if fnmatch("%s.%s" % (test_suite, test_struct.name), test_filter):
+                        if fnmatch("%s.%s" % (test_suite, test_struct.name), test_filter) and \
+                           (not test_struct.arch or (platform.machine().lower() in test_struct.arch)):
                             tests.append(test_struct)
                         test_lines = []
 
@@ -69,6 +71,7 @@ class TestParser(object):
         kernel = ''
         requirement = ''
         env = {}
+        arch = []
 
         for item in test:
             item_split = item.split()
@@ -95,6 +98,8 @@ class TestParser(object):
                 for e in line.split():
                     k, v = e.split('=')
                     env[k]=v
+            elif item_name == 'ARCH':
+                arch = [x.strip() for x in line.split("|")]
             else:
                 raise UnknownFieldError('Field %s is unknown. Suite: %s' % (item_name, test_suite))
 
@@ -107,4 +112,4 @@ class TestParser(object):
         elif timeout == '':
             raise RequiredFieldError('Test TIMEOUT is required. Suite: ' + test_suite)
 
-        return TestStruct(name, run, expect, timeout, before, after, test_suite, kernel, requirement, env)
+        return TestStruct(name, run, expect, timeout, before, after, test_suite, kernel, requirement, env, arch)
--- a/src/arch/CMakeLists.txt
+++ b/src/arch/CMakeLists.txt
@@ -3,6 +3,9 @@ if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
 elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "ppc64" OR
        CMAKE_SYSTEM_PROCESSOR STREQUAL "ppc64le")
     add_library(arch ppc64.cpp)
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "s390" OR
+       CMAKE_SYSTEM_PROCESSOR STREQUAL "s390x")
+    add_library(arch s390.cpp)
 elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
     add_library(arch x86_64.cpp)
 else()
--- a/src/arch/aarch64.cpp
+++ b/src/arch/aarch64.cpp
@@ -3,6 +3,9 @@
 #include <algorithm>
 #include <array>
 
+// SP points to the first argument that is passed on the stack
+#define ARG0_STACK 0
+
 namespace bpftrace {
 namespace arch {
 
@@ -132,6 +135,11 @@ int pc_offset()
 int sp_offset()
 {
   return offset("sp");
+}
+
+int arg_stack_offset()
+{
+  return ARG0_STACK / 8;
 }
 
 std::string name()
--- a/src/arch/arch.h
+++ b/src/arch/arch.h
@@ -11,6 +11,7 @@ int arg_offset(int arg_num);
 int ret_offset();
 int pc_offset();
 int sp_offset();
+int arg_stack_offset();
 std::string name();
 
 } // namespace arch
--- a/src/arch/x86_64.cpp
+++ b/src/arch/x86_64.cpp
@@ -3,6 +3,9 @@
 #include <algorithm>
 #include <array>
 
+// SP + 8 points to the first argument that is passed on the stack
+#define ARG0_STACK 8
+
 namespace bpftrace {
 namespace arch {
 
@@ -78,6 +81,11 @@ int pc_offset()
 int sp_offset()
 {
   return offset("sp");
+}
+
+int arg_stack_offset()
+{
+  return ARG0_STACK / 8;
 }
 
 std::string name()
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -219,7 +219,9 @@ void CodegenLLVM::visit(Builtin &builtin)
                               "reg_sp");
     dyn_cast<LoadInst>(sp)->setVolatile(true);
     AllocaInst *dst = b_.CreateAllocaBPF(builtin.type, builtin.ident);
-    Value *src = b_.CreateAdd(sp, b_.getInt64((arg_num + 1) * sizeof(uintptr_t)));
+    Value *src = b_.CreateAdd(sp,
+                              b_.getInt64((arg_num + arch::arg_stack_offset()) *
+                                          sizeof(uintptr_t)));
     b_.CreateProbeRead(dst, 8, src);
     expr_ = b_.CreateLoad(dst);
     b_.CreateLifetimeEnd(dst);
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1230,7 +1230,7 @@ int BPFtrace::print_map_hist(IMap &map, uint32_t top, uint32_t div)
   while (bpf_get_next_key(map.mapfd_, old_key.data(), key.data()) == 0)
   {
     auto key_prefix = std::vector<uint8_t>(map.key_.size());
-    int bucket = key.at(map.key_.size());
+    uint64_t bucket = read_data<uint64_t>(key.data() + map.key_.size());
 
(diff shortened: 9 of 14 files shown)


sumanthkorikkar:
Please review and request to merge into master.

Thank you.

src/arch/ppc64.cpp
```diff
@@ -99,6 +104,15 @@ int sp_offset()
   return offset("r1");
 }
 
+int arg_stack_offset()
+{
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+  return ARG0_STACK_LE / 8;
+#else
+  return ARG0_STACK_BE / 8;
+#endif
```
danobi, line 113: So @dalehamel added support for static builds of bpftrace that are built on every commit / release. And @fbs added better support for runtime detection of kernel features.

Perhaps this should be a runtime check as well (ie `is_little_endian()`)? So that we can eventually build a single static bpftrace binary that can run on any linux host. Thoughts @dalehamel @fbs ?

danobi:
Hi @sumanthkorikkar , thanks for the PR. I'm skimming it and it seems reasonable (will look more closely next on monday) .

> Shall I again create a pull requests without the last commit?.

It looks like it's hitting some kind of assertion. Do you think you can figure out what's going on?

If you wanna do partial merges of s390x support we can also go with a build flag that's off by default and fix the abort in another PR.

> Or should I create a pull request per commit, so that it can be reviewed individually?

Usually we do it all in one PR, especially if the commits depend on each other. But see above if you wanna get most of the code in first so you don't have to keep fixing merge conflicts.

Other stuff:
* can you look at the clang-format build job failures and apply the suggestions there?
* might be a good time to decide what direction we wanna take static binaries. See inline comment

sumanthkorikkar:
Hi @danobi 

> Hi @sumanthkorikkar , thanks for the PR. I'm skimming it and it seems reasonable (will look more closely next on monday) .

Ok. thank you.

> 
> > Shall I again create a pull requests without the last commit?.
> 
> It looks like it's hitting some kind of assertion. Do you think you can figure out what's going on?

Sure. I will give it a try. 

> 
> If you wanna do partial merges of s390x support we can also go with a build flag that's off by default and fix the abort in another PR.

Ok. I will separate s390 support and I will fix the abort in another PR

> 
> > Or should I create a pull request per commit, so that it can be reviewed individually?
> 
> Usually we do it all in one PR, especially if the commits depend on each other. But see above if you wanna get most of the code in first so you don't have to keep fixing merge conflicts.
> 
> Other stuff:
> 
> * can you look at the clang-format build job failures and apply the suggestions there?

Ok. Sure I will conform it to clang-format and send. Thanks

> * might be a good time to decide what direction we wanna take static binaries. See inline comment

Ok.

src/arch/ppc64.cpp
fbs: I don't think there are any architectures that mix endianness? Linux is either big or little on an architecture (even if the hardware supports both) right? If that won't change there is no reason to detect it at runtime.

Also a single binary that runs on any platform won't work due to the different CPU architectures.

fbs:
Thanks for working on this! :)

Haven't look closely either yet, but just some remarks:

---

As for the helpers. Having both a  'short/i16' and 'long/i32' versions is a bit different from the way most helpers currently work, which usually look at the input argument and derive stuff from it. Instead of having 4 helpers, cant we have one that just flips the byte order? Looking at the change that seems possible to do.

We're not 1.0 yet, but I'd rather not add any helpers that we don't need. If a good use case pops up we can always add more helpers (calls).

---

Re the current compiler issue:

bpftrace is not as strict with LLVM  typing as LLVM wants us to be. While it "mostly works"  you can run into weird issues like the crashes you're seeing now. A simple way to test that is running `bpftrace -dd` if it shows the before opt code but crashes during optimization it is likely a typing issue.

In this case it is caused by:

```
    $dport = $sk->__sk_common.skc_dport;
    $dport = ntohs($dport);
```

Which compiles down to

```
  %"$dport" = alloca i16
...
  %100 = load i16, i16* %"struct sock_common.skc_dport"
...
  %111 = or i16 %108, %110
  %112 = zext i16 %111 to i64
  store i64 %112, i16* %"$dport"
```

Doing a i64 store in a i16* causes the issue in this case. As a fix:

```
]$ diff /vagrant/tools/tcpconnect.bt tcpconnect.bt
46c46
<     $dport = $sk->__sk_common.skc_dport;
---
>     $dport = (uint64) $sk->__sk_common.skc_dport;
```

Lately I've been fixing most of these (e.g #1189) with #1231 and #1194  still open but there probably still are some left.

In this case the issue is mostly with the semantic analyser however.  Your newly added functions have call.type set to i64 which appearently doesn't get checked properly during variable assignment as it only appears to check the type:

```
1481:  else if (search->second.type != assignment.expr->type.type) {
```

I want to start working on improving on this (#878) once the current codegen issues are fixed but that will take a while.

For now it will probably work if you make the return type of your new helpers match the input type, instead of casting it to i64.

---

Can you add codegen tests for the new helpers (after we've decided what to do with them, see fisrt part of comment). 

---

What are your thoughts about testing this and making sure it keeps working? Are there any CI platforms that do s390. At least being able to do semi-regular (daily, weekly) builds would be a huge help.

fbs:
As for splitting up, I having the first 4 or 5 commits going in first would be easier.

src/arch/ppc64.cpp
danobi: > Also a single binary that runs on any platform won't work due to the different CPU architectures.

Yeah right now you can't. But if we wanted, we could have a fatter binary that switches between the different CPU-specific stuff at runtime, right?

fbs: What I mean is that code (e.g. our static binary) compiled for x86 won't run on ARM.  So LE/BE runtime detection is only useful if there is a platform on which Linux runs as both LE and BE, which I don't think there is.

sumanthkorikkar:
Hi @fbs, 

Thanks for reviewing.

> Thanks for working on this! :)
> 
> Haven't look closely either yet, but just some remarks:
> 
> As for the helpers. Having both a 'short/i16' and 'long/i32' versions is a bit different from the way most helpers currently work, which usually look at the input argument and derive stuff from it. Instead of having 4 helpers, cant we have one that just flips the byte order? Looking at the change that seems possible to do.

something like bswap() instead of hton*(). ?. Based on input type, Let me try that. 
For user readability, I added these functions. 

> 
> We're not 1.0 yet, but I'd rather not add any helpers that we don't need. If a good use case pops up we can always add more helpers (calls).
> 
> Re the current compiler issue:
> 
> bpftrace is not as strict with LLVM typing as LLVM wants us to be. While it "mostly works" you can run into weird issues like the crashes you're seeing now. A simple way to test that is running `bpftrace -dd` if it shows the before opt code but crashes during optimization it is likely a typing issue.
> 
> In this case it is caused by:
> 
> ```
>     $dport = $sk->__sk_common.skc_dport;
>     $dport = ntohs($dport);
> ```
> 
> Which compiles down to
> 
> ```
>   %"$dport" = alloca i16
> ...
>   %100 = load i16, i16* %"struct sock_common.skc_dport"
> ...
>   %111 = or i16 %108, %110
>   %112 = zext i16 %111 to i64
>   store i64 %112, i16* %"$dport"
> ```
> 
> Doing a i64 store in a i16* causes the issue in this case. As a fix:
> 
> ```
> ]$ diff /vagrant/tools/tcpconnect.bt tcpconnect.bt
> 46c46
> <     $dport = $sk->__sk_common.skc_dport;
> ---
> >     $dport = (uint64) $sk->__sk_common.skc_dport;
> ```
> 

Ok. Thanks for mentioning this.

> Lately I've been fixing most of these (e.g #1189) with #1231 and #1194 still open but there probably still are some left.
> 
> In this case the issue is mostly with the semantic analyser however. Your newly added functions have call.type set to i64 which appearently doesn't get checked properly during variable assignment as it only appears to check the type:
> 
> ```
> 1481:  else if (search->second.type != assignment.expr->type.type) {
> ```
> 
> I want to start working on improving on this (#878) once the current codegen issues are fixed but that will take a while.
> 
> For now it will probably work if you make the return type of your new helpers match the input type, instead of casting it to i64.

Ok. I will check.

> 
> Can you add codegen tests for the new helpers (after we've decided what to do with them, see fisrt part of comment).

I will add this. But this is specific to little endian. [ llvm/*] datalayout is different for big endian. I will think about this and see how we improve codegen tests for both arch.
 
> 
> What are your thoughts about testing this and making sure it keeps working? Are there any CI platforms that do s390. At least being able to do semi-regular (daily, weekly) builds would be a huge help.

Ok. I need to check with my colleagues, if there is CI platforms for s390.

Thank you

sumanthkorikkar:
Hi @fbs,

> What are your thoughts about testing this and making sure it keeps working? Are there any CI platforms that do s390. At least being able to do semi-regular (daily, weekly) builds would be a huge help.

I have added bpftrace runtime-tests to our internal CI environment as of now. So, we run daily builds and runtime-tests.

sumanthkorikkar:
Hi @fbs, @danobi 

Made changes to initial 5 commits. Other commits related to hton* and later will be done in another PR. Request to merge this into master, if it is good. Thank you.

src/arch/ppc64.cpp
danobi: Ah yeah I'm going crazy.

tests/runtime/engine/parser.py
```diff
@@ -95,6 +98,8 @@ def __read_test_struct(test, test_suite):
                 for e in line.split():
                     k, v = e.split('=')
                     env[k]=v
+            elif item_name == 'ARCH':
+                arch = line.split("|")
```
danobi, line 102: Maybe strip whitespace here too?
```suggestion
                arch = [x.strip() for x in line.split("|")]
```

danobi:
> I have added bpftrace runtime-tests to our internal CI environment as of now. So, we run daily builds and runtime-tests.

Sounds good. If you can find a way to set it up in the open source CI it would help us prevent future breakages at merge time. Not sure if it would be less work on the long run (depends on how much of our stuff implicitly depends on endianness, I guess).

src/arch/ppc64.cpp
```diff
@@ -3,6 +3,11 @@
 #include <algorithm>
 #include <array>
 
+// For little endian 64 bit, sp + 32 + 8 regs save area + argX
+#define ARG0_STACK_LE 96
+// For big endian 64 bit, sp + 48 + 8 regs save area + argX
+#define ARG0_STACK_BE 112
```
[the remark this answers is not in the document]
sumanthkorikkar: Powerpc64, big endian
https://refspecs.linuxfoundation.org/ELF/ppc64/PPC-elf64abi.html#PARAM-PASS 
- The parameter save area, which is located at a fixed offset of 48 bytes from the stack pointer, is reserved in each stack frame for use as an argument list. A minimum of 8 doublewords is always reserved.
- General registers are used to pass some values. The first eight doublewords mapped to the parameter save area correspond to the registers r3 through r10. 
- All other arguments (or parts thereof) not already covered must be stored in the parameter save area following the first eight doublewords.

src/arch/ppc64.cpp
```diff
@@ -3,6 +3,11 @@
 #include <algorithm>
 #include <array>
 
+// For little endian 64 bit, sp + 32 + 8 regs save area + argX
+#define ARG0_STACK_LE 96
```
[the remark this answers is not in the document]
sumanthkorikkar: Powerpc64, little endian, See Power Architecture 64-Bit ELF V2 ABI Specification
https://members.openpowerfoundation.org/document/dl/576
- See parameter save area
   - When the caller allocates the Parameter Save Area, it will always be automatically quadword aligned  because it must always start at SP + 32. **It shall be at least 8 doublewords in length.** If a function needs to pass more than 8 doublewords of arguments, the Parameter Save Area shall be large enough to spill all register-based parameters and to contain the arguments that the caller stores in it.

src/arch/s390.cpp
```diff
@@ -5,6 +5,9 @@
 
 #define REQ_REGISTERS 19
 #define ARG_REGISTERS 5
+// For s390x, r2-r6 registers are used as function arguments, then the extra
+// arguments can be found starting at sp+160
+#define ARG0_STACK 160
```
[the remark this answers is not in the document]
sumanthkorikkar: - Check Parameter Passing.
	- r2 - r6 - parameters registers
	- the rest of the arguments are passed on the stack 160 bytes above the initial stack pointer.

http://refspecs.linuxfoundation.org/ELF/zSeries/lzsabi0_zSeries.html#PARAMETERPASSING
http://refspecs.linuxfoundation.org/ELF/zSeries/lzsabi0_zSeries.pdf

tests/runtime/engine/parser.py
sumanthkorikkar: ok. Will change it.

src/arch/ppc64.cpp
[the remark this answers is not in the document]
sumanthkorikkar: Hence it is sp + 48 + 8 regs save area + argX

sumanthkorikkar:
> > I have added bpftrace runtime-tests to our internal CI environment as of now. So, we run daily builds and runtime-tests.
> 
> Sounds good. If you can find a way to set it up in the open source CI it would help us prevent future breakages at merge time. Not sure if it would be less work on the long run (depends on how much of our stuff implicitly depends on endianness, I guess).

Hi @danobi , @fbs 

yes. that would be great for community. One of my colleague, referred me to this, So it should be possible.
https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z

I can try this. **But I would need some input from you people about the requirements - (os requirements, services, scripts) to run this.?**
- Some docker scripts, travis yaml file which are currently used to run bpftrace
- The following seems to use LXD container instead of docker and verify if we can run bpftrace as root.

src/arch/ppc64.cpp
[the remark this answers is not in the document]
danobi: Ah, got it. 8 doublewords if callee wants to treat all the args as contiguous memory (requiring  callee stores I'm assuming).

src/arch/ppc64.cpp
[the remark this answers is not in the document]
danobi: Hm I'm a little confused. Seems like the ABI changed starting at power ISA 2.07B. So why is only LE using the new ABI?

Won't block this PR on these questions. Realizing this was an issue is a big win already.

danobi:
> But I would need some input from you people about the requirements - (os requirements, services, scripts) to run this.?

I don't know if we have any hard requirements other than it functioning as expected. I don't think any of us are too picky on how tests get run, just that they run somehow. We have github actions and travis ci at our disposal.

> The following seems to use LXD container instead of docker and verify if we can run bpftrace as root.

This seems like a pain. We also have github actions set up. I looked around and I found this: https://github.com/marketplace/actions/run-on-architecture . Looks much easier to get set up than lxd. Seems like it's powered by https://github.com/multiarch/qemu-user-static which looks quite powerful.

fbs:
Looks mostly good to me. I'm not that sure about "bpftrace: Fix type conversion in printf() builtin". Using `cast_type` probably works ok in most cases but because structs are casts and casts are casts (obviously) but not all types are 8 bytes. E.g. cpid is 4bytes.

I think that the better way to fix this is to get rid of that resize the semantic analyser does and instead make sure everything is properly stored. Which is what I want to do with #1231. 

I think the best we can do is to accept for now and revert it later (e.g. in #1231).

luser1990:
> > But I would need some input from you people about the requirements - (os requirements, services, scripts) to run this.?
> 
> I don't know if we have any hard requirements other than it functioning as expected. I don't think any of us are too picky on how tests get run, just that they run somehow. We have github actions and travis ci at our disposal.
> 
> > The following seems to use LXD container instead of docker and verify if we can run bpftrace as root.
> 
> This seems like a pain. We also have github actions set up. I looked around and I found this: https://github.com/marketplace/actions/run-on-architecture . Looks much easier to get set up than lxd. Seems like it's powered by https://github.com/multiarch/qemu-user-static which looks quite powerful.

Hi @danobi 

My findings:
1. I gave qemu-user-static a try. I tried running bpftrace command - s390x/fedora, priviliged mode docker run.
- Run docker run -t -v /usr/src:/usr/src:ro -v /lib/modules/:/lib/modules:ro -v /sys/kernel/debug/:/sys/kernel/debug:rw --net=host --pid=host --privileged -t test-demo
- Dockerfile contains setup to run sys_open_read() probe 
- Starting the script right shows the following error:
Error creating printf map: Function not implemented
##[error]Process completed with exit code 1.
- Compilation is quite slow and may timeout when running runtime tests.

- Related issue : https://github.com/multiarch/qemu-user-static/issues/17. Tried this, but I could only get the same error message.

2. However the same works without qemu-user-static for x86

3. Using lxd containers may not be an option, as this does not provide access to /sys/kernel/debug:rw, which bpftrace needs 

4. Other options to be discussed with the team. But this would need some time

sumanthkorikkar:
> > But I would need some input from you people about the requirements - (os requirements, services, scripts) to run this.?
> 
> I don't know if we have any hard requirements other than it functioning as expected. I don't think any of us are too picky on how tests get run, just that they run somehow. We have github actions and travis ci at our disposal.
> 
> > The following seems to use LXD container instead of docker and verify if we can run bpftrace as root.
> 
> This seems like a pain. We also have github actions set up. I looked around and I found this: https://github.com/marketplace/actions/run-on-architecture . Looks much easier to get set up than lxd. Seems like it's powered by https://github.com/multiarch/qemu-user-static which looks quite powerful.

Hi @danobi,

My findings:
1. I gave qemu-user-static a try. I tried running bpftrace command,  s390x/fedora with priviliged mode docker run.
- Run docker run -t -v /usr/src:/usr/src:ro -v /lib/modules/:/lib/modules:ro -v /sys/kernel/debug/:/sys/kernel/debug:rw --net=host --pid=host --privileged -t test-demo
- Dockerfile contains setup to run sys_open_read() probe 
- Starting the script:
Error creating printf map: Function not implemented
##[error]Process completed with exit code 1.
- Compilation is quite slow and may timeout when running runtime tests.

-Related issue : https://github.com/multiarch/qemu-user-static/issues/17. Tried this, but same error message.

2. However the same works without qemu-user-static for x86

3. Using lxd containers may not be the option, as this does not provide access to /sys/kernel/debug:rw, which bpftrace needs 

4. Should discuss about other options with the team. But this may need some time.

tests/runtime/engine/parser.py
sumanthkorikkar: Added. Thanks

sumanthkorikkar:
> Looks mostly good to me. I'm not that sure about "bpftrace: Fix type conversion in printf() builtin". Using `cast_type` probably works ok in most cases but because structs are casts and casts are casts (obviously) but not all types are 8 bytes. E.g. cpid is 4bytes.
> 
> I think that the better way to fix this is to get rid of that resize the semantic analyser does and instead make sure everything is properly stored. Which is what I want to do with #1231.
> 
> I think the best we can do is to accept for now and revert it later (e.g. in #1231).

Ok. Thanks

fbs:
+1. Lets get the 0.10 release out first and merge it first thing after

PR #1242, opened 2020-04-03
Add links to examples for probes in reference_guide

n1tram1:
During my discovery of bpftrace I missed the fact there were no links to examples for the probe. So I added links the same way BCC has.

Some probes don't have any links to examples because there are no tools which use them in the bpftrace repo.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1247,6 +1247,9 @@ Attaching 2 probes...
 
 This prints the rate of syscalls per second.
 
+Examples in situ:
+[search /tools](https://github.com/search?q=interval%3A+path%3Atools+extension%3Abt&type=Code)
+
 ## 11. `software`: Pre-defined Software Events
 
 Syntax:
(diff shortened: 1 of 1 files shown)


docs/reference_guide.md
```diff
@@ -1193,6 +1215,8 @@ Attaching 1 probe...
 @[32586]: 98
 @[0]: 579
 ```
+Examples in situ:
+[tool: profile_order.bt](https://github.com/iovisor/bpftrace/blob/master/tests/runtime/scripts/profile_order.bt)
```
danobi, line 1219: I'd rather not link to runtime tests. We probably won't ever move the `scripts/` dir but I wouldn't bet on runtime test stuff staying still forever. And there's no way anyone will remember to update this if it happens.

docs/reference_guide.md
```diff
@@ -1223,6 +1247,9 @@ Attaching 2 probes...
 
 This prints the rate of syscalls per second.
 
+Examples in situ:
+[search /tools](https://github.com/search?q=interval%3A+path%3Atools+extension%3Abt&type=Code)
```
danobi, line 1251: I think this is searching all of github

docs/reference_guide.md
```diff
@@ -1324,6 +1351,10 @@ END
 These are special built-in events provided by the bpftrace runtime. `BEGIN` is triggered before all other
 probes are attached. `END` is triggered after all other probes are detached.
 
+Examples in situ:
+[(BEGIN) search /tools](https://github.com/search?q=BEGIN+path%3A%2Ftools+extension%3Abt&type=Code)
+[(END) search /tools](https://github.com/search?q=END+path%3A%2Ftools+extension%3Abt&type=Code)
```
danobi, line 1356: Same as above

docs/reference_guide.md
n1tram1: You are right, I tried playing with the search query and I messed up, will fix

PR #1243, opened 2020-04-04
Use BPFtrace::error for TracepointFormatParser errors

### LINKED ISSUE #1213: Use `BPFtrace::error` in `TracepointFormatParser`
TracepointFormatParser currently uses std::cerr to print errors. `BPFtrace::error` is a bit nicer b/c it highlights error locations using source location information. Would be nice to have that for tracepoint format parser.

n1tram1:
Fix #1213 

I didn't change BPFtrace::error for the verbose errors because when I tested, it didn't look right.
(Those verbose errors show messages from syscalls, it doesn't make sense to also show a file location).

### DIFF
--- a/src/tracepoint_format_parser.cpp
+++ b/src/tracepoint_format_parser.cpp
@@ -43,8 +43,10 @@ bool TracepointFormatParser::parse(ast::Program *program, BPFtrace &bpftrace)
 
         if (has_wildcard(category))
         {
-            bpftrace.error(std::cerr, ap->loc,
-              "wildcards in tracepoint category is not supported: " + category);
+          bpftrace.error(std::cerr,
+                         ap->loc,
+                         "wildcards in tracepoint category is not supported: " +
+                             category);
           return false;
         }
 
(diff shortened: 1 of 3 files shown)


src/tracepoint_format_parser.cpp
```diff
@@ -98,12 +105,20 @@ bool TracepointFormatParser::parse(ast::Program *program)
...
             if (category == "syscall")
-              std::cerr << "Did you mean syscalls:" << event_name << "?" << std::endl;
+              bpftrace.warning(std::cerr,
+                               ap->loc,
+                               "Did you mean syscalls:" + event_name + "?");
             if (bt_verbose) {
-                std::cerr << strerror(errno) << ": " << format_file_path << std::endl;
+              // Having the location info isn't really useful here, so no
+              // bpftrace.error
+              std::cerr << strerror(errno) << ": " << format_file_path
```
danobi, line 120: Hmm, errno could easily be clobbered here. I read through llvm's libcxx implementation and it seems like errno should still be respected (although I remain skeptical that it's in the spec).

 Can you save a copy of errno right after `if (format_file.fail())` so it's not accidentally clobbered by bpftrace.error() or bpftrace.warning()?

n1tram1: Done

src/tracepoint_format_parser.cpp
```diff
@@ -98,12 +105,26 @@ bool TracepointFormatParser::parse(ast::Program *program)
...
+                           "tracepoint not found: " + category + ":" +
+                               event_name);
             // helper message:
             if (category == "syscall")
-              std::cerr << "Did you mean syscalls:" << event_name << "?" << std::endl;
+              bpftrace.warning(std::cerr,
+                               ap->loc,
+                               "Did you mean syscalls:" + event_name + "?");
+
+            errno = saved_errno;
```
danobi, line 121: Bit strange to re-set errno. Why not just use saved_errno like `strerror(saved_errno)`?

PR #1247, opened 2020-04-07
Only link agains libbpf if it exists

fbs:
Libbpf is optional so we must only link against it when it exists. This
optional linking is already  done for the dynamic build but not the
static:

```
if (LIBBPF_BTF_DUMP_FOUND)
  target_compile_definitions(bpftrace_test PRIVATE HAVE_LIBBPF_BTF_DUMP)
  target_include_directories(bpftrace_test PUBLIC ${LIBBPF_INCLUDE_DIRS})
  target_link_libraries(bpftrace_test ${LIBBPF_LIBRARIES})
endif(LIBBPF_BTF_DUMP_FOUND)
```


fbs:
Looks like this somehow broke static builds in the CI.

PR #1253, opened 2020-04-10
Remove codegen tests warning

fbs:
```
[ RUN      ] codegen.builtin_cpid

GMOCK WARNING:
Uninteresting mock function call - taking default action specified at:
/vagrant/tests/codegen/builtin_cpid.cpp:17:
    Function call: child_pid()
          Returns: 1337
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#knowing-when-to-expect for details.
```

### DIFF
--- a/tests/codegen/builtin_cpid.cpp
+++ b/tests/codegen/builtin_cpid.cpp
@@ -6,15 +6,16 @@ namespace codegen {
 
 class MockBPFtraceCpid : public BPFtrace
 {
-public:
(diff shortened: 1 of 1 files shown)


danobi:
Please clang-format

PR #1254, opened 2020-04-11
0.10.0 release changelog

### LINKED ISSUE #1214: 0.9.5 / 0.10.0 release
I've originally set the release date for the next release to April 1. As we're approaching 100 commits again (60 in the `src` dir) we should probably stick to that.

Is there anything that has to land before we do our next release? The milestone itself is empty (just removed `while` as that is blocked on the type fixes). 

We also still have the version number "issue". IIRC the main argument against a 0.10.0 release is that it looks a lot like an 0.1.0. My preference would be to do a 0.10.0 as sticking to 0.9.x feels like we're doing minor patch releases. But if there is good reason to stick to 0.9.0 thats fine too (or do we skip to 0.11.0?)

@ajor @brendangregg  @danobi  @mmarchini @mmisono

fbs:
Closes #1214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,254 @@ and this project adheres to
 
 ## Unreleased
 
+## [0.10.0] 2020-04-12
+
+### Highlights
+
+#### kfuncs
+
+Improved kprobes which are near zero overhead and use BTF to derive argument
+names and types:
+
+```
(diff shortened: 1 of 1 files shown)


fbs:
Based on some grep magic I don't see any duplicate commits like last time. Looks like all 175 commits (`git log --oneline  v0.9.4..upstream/master --no-merges  | wc -l`) are in there too this time.

Issue #1215, opened 2020-03-17
Drop LLVM 5 codegen tests


fbs:
We currently have codegen tests for LLVM 5 - 9 (10 in a PR).  The code we generate is basically the same for each LLVM version (the only LLVM_VERSION ifdefs are the `mem*` calls) so there is not much to test between versions there. That we're able to automatically rewrite most of the tests confirms that.
The main advantage of having codegen tests for each version is probably that it makes development easy, you can use the LLVM your OS has instead of having everyone use the same version.

Especially the LLVM5 tests mainly feel like a maintenance burden. The automatic rewriting fails in a lot of places (17) and the CI job has a massive test filter too.
Are there objections to dropping LLVM 5 from the codegen tests?  We still have the runtime and other tests to verify that everything is working. This might cause some issues for people that still use LLVM5 as their codegen tests will fail (unless we code the exception into the test runner) but I don't think there are many (any) LLVM5 users left. Even ubunt 16.04 has LLVM8 

Or we could drop LLVM5 support completely, removing it completely (and use the glibc static(ish) builds instead).

danobi:
+1 on dropping

mmisono:
LLVM 10 test is merged: #1190
+1 on dropping

acj:
Another +1 for dropping

> Or we could drop LLVM5 support completely, removing it completely (and use the glibc static(ish) builds instead).

fwiw, #1063 removes the static LLVM 5 builds in favor of LLVM 9, still on Alpine as before.

PR #1260, opened 2020-04-14
Use std::optional in bpffeature

fbs:
As we updated to C++17 we can start using std::optional and remove the
unique_ptr hack.

### DIFF
--- a/src/bpffeature.cpp
+++ b/src/bpffeature.cpp
@@ -107,8 +107,8 @@ bool BPFfeature::detect_map(enum bpf_map_type map_type)
 
 bool BPFfeature::has_loop(void)
 {
-  if (has_loop_)
-    return *has_loop_.get();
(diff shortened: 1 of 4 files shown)


fbs:
Blocked on #1215

src/bpffeature.cpp
```diff
@@ -117,52 +117,52 @@ bool BPFfeature::has_loop(void)
...
+  has_loop_ = std::make_optional<bool>(
       try_load(BPF_PROG_TYPE_TRACEPOINT, insns, ARRAY_SIZE(insns)));
 
   return has_loop();
 }
 
 int BPFfeature::instruction_limit(void)
 {
-  if (!insns_limit_)
+  if (insns_limit_.has_value())
```
danobi, line 128: Wow what a hilariously bad diff from github

Issue #1255, opened 2020-04-11
Improve changelog / release process


fbs:
Our changelog is currently a list of commits grouped into a few sections. To end users this isn't useful as it is hard to make sense of what individual commits do. It would be a lot more useful if we describe what each PR does and how it impacts the end user. But doing this once every two months is a lot of work.

We've discussed this on IRC a bit, the best we came up with is to make this part of every PR. Each PR that add/changes/fixes something that impacts the user must also create a change log entry. Internal changes (refactoring etc) don't need a changelog entry.

To make sure we don't forget this we can add a PR template or even hook it into the CI.

Are there any objections to this?

danobi:
+1

danobi:
So github actions can't comment on a pull request coming from a forked repository: https://github.community/t5/GitHub-Actions/Github-Workflow-not-running-from-pull-request-from-forked/td-p/32979/highlight/true

Quite unfortunate. I think the next best thing is a pull request template

danobi:
PR template: #1299

fbs:
I think the current flow is good enough for now

PR #1261, opened 2020-04-14
Create changelog entries for merged PRs

fbs:
As we've discussed in #1255  we could use a better changelog, one that just shows the important changes instead of every commit that has been done. This is a first attempt at doing this. 

Instead of showing individual commits this groups by change. This means that one change can have multiple PRs linked to it, but also that one PR can have multiple changes in it (like the s390x change set).
I've dropped the contributor info as that makes it a bit too noisy imo, I think we should just create a list of (new?) contributors when doing the release.

Opinions @mmisono  @danobi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,48 @@ and this project adheres to
 
 ## Unreleased
 
+### All Changes
+
+#### Added
+
+- Support for s390x
+  - [#1241](https://github.com/iovisor/bpftrace/pull/1241)
+- `buf` a new function that makes it possible to safely print arbitrary binary data
+  - [#1107](https://github.com/iovisor/bpftrace/pull/1107)
+- A new function, `sizeof`, which returns the size of an expression, similar to  `sizeof` in C
+  - [#1269](https://github.com/iovisor/bpftrace/pull/1269)
+- C style while loop support, `while ($a < 100) { $a++ }`
+  - [#1066](https://github.com/iovisor/bpftrace/pull/1066)
+
+#### Changed
+
+- Require C++17 and CMake 3.8 for building bpftrace
+  - [#1200](https://github.com/iovisor/bpftrace/pull/1200)
+  - [#1259](https://github.com/iovisor/bpftrace/pull/1259)
+
+#### Deprecated
+
+#### Removed
+
+- Drop LLVM 5 support
+  - [#1215](https://github.com/iovisor/bpftrace/issues/1215)
+
+#### Fixed
+
+- Various big endian related fixes
+  - [#1241](https://github.com/iovisor/bpftrace/pull/1241)
+- Type check the `cond` of if and ternary statements
+  - [#1229](https://github.com/iovisor/bpftrace/pull/1229)
+
+#### Tools
+
+#### Documentation
+
+- Clean up README
+  - [#1273](https://github.com/iovisor/bpftrace/pull/1273)
+- Add missing `struct` keyword to examples in the one liner tutorial
+  - [#1275](https://github.com/iovisor/bpftrace/pull/1275)
+
 ## [0.10.0] 2020-04-12
 
 ### Highlights


mmisono:
I like the new format. The link to the PR is mostly we want to look at. As of the contributor, dropping is fine with me. Alternatively, we could use github's username (e.g., @mmisono). (just saw [this](https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md))

The one thing I'm wondering is how to obtain a PR number before submitting a PR, or is there an easy way to embed a PR number in the changelog?

fbs:
> Alternatively, we could use github's username (e.g., @mmisono). (just saw [this](https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md))

I guess we can give it a try. If it turns out that multiple usernames + PRs gets too much we can always drop the users and have a contributor list at the end.

> The one thing I'm wondering is how to obtain a PR number before submitting a PR, or is there an easy way to embed a PR number in the changelog?

I think the only way is to open the PR and then amend the ID into it later. Or we can make the changelog based on issues. But then each PR would have to be linked to an issue and most of the discussion will have to happen in the issue which doesn't usually happen (like we do now ;))

CHANGELOG.md
```diff
@@ -8,6 +8,30 @@ and this project adheres to
 
 ## Unreleased
 
+### All Changes
+
+#### Added
+
+- Support for s390x ([#1241](https://github.com/iovisor/bpftrace/pull/1241))
```
fbs, line 15:
```suggestion
- Support for s390x ([#1241](https://github.com/iovisor/bpftrace/pull/1241) by [@sumanthkorikkar](https://github.com/sumanthkorikkar))
```

CHANGELOG.md
```diff
@@ -8,6 +8,30 @@ and this project adheres to
...
+
+#### Changed
+
+- Require C++17 and CMake 3.8
+  ([#1200](https://github.com/iovisor/bpftrace/pull/1200), [#1259](https://github.com/iovisor/bpftrace/pull/1259))
```
fbs, line 20:
```suggestion
  ([#1200](https://github.com/iovisor/bpftrace/pull/1200) by [@mmisono](https://github.com/mmisono, [#1259](https://github.com/iovisor/bpftrace/pull/1259) by [@fbs](https://github.com/fbs))
```

CHANGELOG.md
```diff
@@ -8,6 +8,30 @@ and this project adheres to
...
+### All Changes
+
+#### Added
+
+- Support for s390x ([#1241](https://github.com/iovisor/bpftrace/pull/1241))
+
+#### Changed
+
+- Require C++17 and CMake 3.8
+  ([#1200](https://github.com/iovisor/bpftrace/pull/1200) by [@mmisono](https://github.com/mmisono, [#1259](https://github.com/iovisor/bpftrace/pull/1259) by [@fbs](https://github.com/fbs))
```
fbs, line 20:
```suggestion
  ([#1200](https://github.com/iovisor/bpftrace/pull/1200) by [@mmisono](https://github.com/mmisono), [#1259](https://github.com/iovisor/bpftrace/pull/1259) by [@fbs](https://github.com/fbs))
```

danobi:
>  I've dropped the contributor info as that makes it a bit too noisy imo, I think we should just create a list of (new?) contributors when doing the release.

Sounds reasonable to me. Git history has everything too.

> The one thing I'm wondering is how to obtain a PR number before submitting a PR, or is there an easy way to embed a PR number in the changelog?

> I think the only way is to open the PR and then amend the ID into it later. Or we can make the changelog based on issues

I looked around and I don't see any easy ways either. We could get complicated and do scripts and stuff but I'd rather keep complexity down. How about using https://github.com/marketplace/actions/changelog-reminder and making it remind PR-openers to edit the changelog and put the PR # into it? Seems like the low tech way that lets people still consciously opt out of changelog entries (eg documentation fix).

fbs:
> How about using https://github.com/marketplace/actions/changelog-reminder and making it remind PR-openers to edit the changelog and put the PR # into it? Seems like the low tech way that lets people still consciously opt out of changelog entries (eg documentation fix).

Looks ok and easy enough to try.

What do you think of the updated PR @danobi ? I think the plain text readability decreased a bit, but I'm not sure whether people still do plain text.

danobi:
Is it necessary to tag names? That information is made available through the PR link.

Also, would be nice to keep lines shorter (easier to skim). Maybe like this?
---
#### Added

- Support for s390x
  - [#1241](https://github.com/iovisor/bpftrace/pull/1241)
- `buf` a new function that makes it possible to safely print arbitrary binary data
  - [#1107](https://github.com/iovisor/bpftrace/pull/1107)

#### Changed

- Require C++17 and CMake 3.8 for building bpftrace
  - [#1200](https://github.com/iovisor/bpftrace/pull/1200)
  - [#1259](https://github.com/iovisor/bpftrace/pull/1259)

fbs:
Updated format and added everything merged recently (i hope)

fbs:
I've added a simple changelog entry check too. https://github.com/iovisor/bpftrace/actions/runs/83256142

danobi:
> I've added a simple changelog entry check too. https://github.com/iovisor/bpftrace/actions/runs/83256142

Do you think you can make it a comment on the PR instead? I think it's worthwhile not to train people to ignore CI checks. Maybe the comment can say something like

```
It looks like this PR doesn't have a CHANGELOG entry attached.

If your change is non-trivial and user-facing, please add an entry following the existing style.
```

danobi:
https://github.com/marketplace/actions/comment-on-pr checks for dupes and has support for a template file. And I think you can use https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths to check for CHANGELOG in modified files. But you probably need to create a new workflow for it.

fbs:
> Do you think you can make it a comment on the PR instead? I think it's worthwhile not to train people to ignore CI checks. Maybe the comment can say something like

Fair point. Was a bit worried that the comment might get lost on a busy PR but  we can just give it a try and see what happens.

If you're happy with the layout I'd like to get this merged so we can start making use of it in other PRs (ill do a manual comment there for now). Keeping it updated after each PR gets a bit annoying :p

danobi:
I'll see if I can get the automation set up. Already pretty familiar with it.

PR #1264, opened 2020-04-14
Fix "Unsupported architecture" error message

caringi:
To show a CMAKE variable value, a '$' prefix should be used.


fbs:
ah derp, thanks :)

PR #1265, opened 2020-04-14
Enable LTO if supported

fbs:


kloczek:
IMO that is extremely bad solution.

Things like LTO after compiler specific and they should be not coder in built framework.
To use LTIO all whgat is necessary is to pass necessary settings in env varables like, AR, RANLIB, NM, CC, CXX CFLAGS, CXXFLAGS and LDFLAGS. In cmake files should none about LTO because LTO can be handled only using those variables.

danobi:
> Things like LTO after compiler specific

Hmm but isn't cmake offering a compiler-independent check here?

> all whgat is necessary is to pass necessary settings in env varables like, AR, RANLIB, NM, CC, CXX CFLAGS, CXXFLAGS and LDFLAGS. In cmake files should none about LTO because LTO can be handled only using those variables.

True as that may be, I don't have those flags memorized. Using the build system to encode things about the build is, well, the point of the system.

You also do not have to use this flag. Notice how no extra work is being done if the flag is off.

kloczek:
> > Things like LTO after compiler specific
> 
> Hmm but isn't cmake offering compiler-independent helpers here?

Again: in case of build framework you don't need yto place any bits of LTO, PGO or any other type of optimisation. Try to think that you will be supporting in cmake support from -Os, to -O3 and all other -ffoo optimisation flags .. pointeless.

> > all whgat is necessary is to pass necessary settings in env varables like, AR, RANLIB, NM, CC, CXX CFLAGS, CXXFLAGS and LDFLAGS. In cmake files should none about LTO because LTO can be handled only using those variables.
> 
> True as that may be, I don't have those flags memorized. Using the build system to encode things about the build is, well, the point of the system.
> 
> You also do not have to use this flag. Notice how no extra work is being done if the flag is off.

I can show you example of env variables which I'm using and exactly the same set of flags is used with ac/am/lt, meson, cmake and all other build framework.
```
]$ rpm -E %set_build_flags

        export CFLAGS="${CFLAGS:--O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -flto=auto -flto-partition=none}";
        export CXXFLAGS="${CXXFLAGS:--O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -flto=auto -flto-partition=none}";
        export FFLAGS="${FFLAGS:--O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -I/usr/lib64/gfortran/modules -flto=auto -flto-partition=none}";
        export FCFLAGS="${FCFLAGS:--O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -I/usr/lib64/gfortran/modules -flto=auto -flto-partition=none}";
        export LDFLAGS="${LDFLAGS:--Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -flto=auto -flto-partition=none -fuse-linker-plugin}";
        export AR="/usr/bin/gcc-ar" RANLIB="/usr/bin/gcc-ranlib" NM="/usr/bin/gcc-nm";
        export CC="gcc" CXX="g++"
```
With that I can control optimisations and compiler type without harcoding those settings in rpm spec files. All could be controlled from outside of all my spec files. With that I can build any of my packages using clang by simple change what that macro provides.

fbs:
> > > Things like LTO after compiler specific
> > 
> > 
> > Hmm but isn't cmake offering compiler-independent helpers here?
> 
> Again: in case of build framework you don't need yto place any bits of LTO, PGO or any other type of optimisation. Try to think that you will be supporting in cmake support from -Os, to -O3 and all other -ffoo optimisation flags .. pointeless.

I think I get your point, if you build packages you don't want to have to keep track of 100 cmake flags that do whatever and might change every version. You want an easy way to control what happens between all packages you build. 
At the same time I think we should provide good defaults for our users and not require them to set 20 flags just to get started (I'm assuming a lot of people still compile their own version as distros are often behind). Especially as not all of those flags work for every platform making it a very complicated story. And as cmake can detect those things it seems like the best option.

Maybe we should introduce a single global flag (EXPERT_MODE?) that acts as a global toggle? If its set cmake will only set  the required flags, all the `add_compile_options()` will be skipped. Individual things like LTO can still be behind their own flag too, like `if (LTO and !EXPERT_MODE) { enable_lto }`

Or we could reverse it, have an `OPTIMIZE_BUILD` flag that defaults to false.

fbs:
Just realized that our test progs will have to be compiled with a few specific flags `-O0 -no-pie` to avoid issues in the test suite, not sure how well that combines with setting `*FLAGS`.

kloczek:
> At the same time I think we should provide good defaults for our users and not require them to set 20 flags just to get started (I'm assuming a lot of people still compile their own version as distros are often behind). Especially as not all of those flags work for every platform making it a very complicated story. And as cmake can detect those things it seems like the best option

Please don't do that because none of the other project does that.

Please do not try to solve Wold Famine Problems using bpftrace cmake files :)

If everything will be like it is now you will be able in bpftrace CIs control building with/without LTO, using GCC and CLang. All **without touching cmake options/switches**.

Really please leave choosing optimisation and compiler to packagers :)
I really think bpftrace developers time could be better invested :P 

If any source code build framework fiddles around env variables which Ive listed sooner or later it only creates some additional problems/issues.

PR #1268, opened 2020-04-15
Allow user supplied C definitions to reference BTF definitions

danobi:
Before, this would fail even with --btf flag:

    struct foo
    {
	  struct task_struct t;
    }

    BEGIN
    {
    }

This should be allowed since we can get everything we need from BTF
data.

This commit adds another pass on clang AST to figure out which structs
are referenced but not defined. There's also a hack that passes
`-D_LINUX_TYPES_H` to the clang front end. It's pretty bad but it's also
the cleanest way I can think of to avoid redefinition conflicts.

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -14,6 +14,63 @@
 #include "field_analyser.h"
 
 namespace bpftrace {
+namespace {
+const std::vector<CXUnsavedFile> &getDefaultHeaders()
+{
+  static std::vector<CXUnsavedFile> unsaved_files = {
+    {
+        .Filename = "/bpftrace/include/__stddef_max_align_t.h",
+        .Contents = __stddef_max_align_t_h,
+        .Length = __stddef_max_align_t_h_len,
+    },
+    {
+        .Filename = "/bpftrace/include/float.h",
+        .Contents = float_h,
+        .Length = float_h_len,
+    },
+    {
+        .Filename = "/bpftrace/include/limits.h",
+        .Contents = limits_h,
+        .Length = limits_h_len,
+    },
+    {
+        .Filename = "/bpftrace/include/stdarg.h",
+        .Contents = stdarg_h,
+        .Length = stdarg_h_len,
+    },
+    {
+        .Filename = "/bpftrace/include/stddef.h",
+        .Contents = stddef_h,
+        .Length = stddef_h_len,
+    },
+    {
+        .Filename = "/bpftrace/include/stdint.h",
+        .Contents = stdint_h,
+        .Length = stdint_h_len,
+    },
+    {
+        .Filename = "/bpftrace/include/" CLANG_WORKAROUNDS_H,
+        .Contents = clang_workarounds_h,
+        .Length = clang_workarounds_h_len,
+    },
+  };
+
+  return unsaved_files;
+}
+
+std::vector<CXUnsavedFile> getTranslationUnitFiles(
+    const CXUnsavedFile &main_file)
+{
+  std::vector<CXUnsavedFile> files;
+  files.reserve(1 + files.size());
+
+  files.emplace_back(main_file);
+  const auto &dfl = getDefaultHeaders();
+  files.insert(files.end(), dfl.cbegin(), dfl.cend());
+
+  return files;
+}
+} // namespace
 
 static std::string get_clang_string(CXString string)
 {
@@ -422,6 +479,9 @@ std::unordered_set<std::string> ClangParser::get_incomplete_types(
     std::vector<CXUnsavedFile> &unsaved_files,
     const std::vector<const char *> &args)
 {
+  if (input.empty())
+    return {};
+
   ClangParserHandler handler;
   CXErrorCode error;
   {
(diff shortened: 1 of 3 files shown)


src/clang_parser.cpp
```diff
@@ -518,11 +607,33 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
...
+        .Length = from_btf.size(),
+    });
+
+    input = "#include <__btf_dependent_decls.h>\n" + input;
+    unsaved_files[0].Contents = input.c_str();
+    unsaved_files[0].Length = input.size();
+
+    // Prevent BTF generated header from redefining stuff found
+    // in <linux/types.h>
+    args.push_back("-D_LINUX_TYPES_H");
```
danobi, line 627: The hack in question. Without it, you'll get stuff like:
```
/bpftrace/include/__btf_dependent_decls.h:12:3: error: typedef redefinition with different types ('struct atomic_t' vs 'struct atomic_t')
/lib/modules/5.5.15-200.fc31.x86_64/source/include/linux/types.h:173:3: note: previous definition is here
/bpftrace/include/__btf_dependent_decls.h:35:8: error: redefinition of 'list_head'
/lib/modules/5.5.15-200.fc31.x86_64/source/include/linux/types.h:181:8: note: previous definition is here
/bpftrace/include/__btf_dependent_decls.h:179:8: error: redefinition of 'hlist_head'
/lib/modules/5.5.15-200.fc31.x86_64/source/include/linux/types.h:185:8: note: previous definition is here
/bpftrace/include/__btf_dependent_decls.h:264:8: error: redefinition of 'hlist_node'
/lib/modules/5.5.15-200.fc31.x86_64/source/include/linux/types.h:189:8: note: previous definition is here
/bpftrace/include/__btf_dependent_decls.h:417:3: error: typedef redefinition with different types ('struct atomic64_t' vs 'struct atomic64_t')
/lib/modules/5.5.15-200.fc31.x86_64/source/include/linux/types.h:178:3: note: previous definition is here
/bpftrace/include/__btf_dependent_decls.h:432:8: error: redefinition of 'callback_head'
/lib/modules/5.5.15-200.fc31.x86_64/source/include/linux/types.h:219:8: note: previous definition is here
```

olsajiri:
nice, I wonder we could make it go through btf.c_def just once, so we would skip the types iterations, could we perhaps:

- compile user input first to get missing types and add them to bpftrace.btf_set_ 
- then we would call parse_btf_definitions and add the output also to unsaved_files later

there could be some unwanted definitions in unsaved_files this way, but it should be ok

just an idea to skip one btf.c_def call, it could have other flaws I can't see now ;-)
other that that your change looks ok

danobi:
@olsajiri I added a commit to bring down clang AST passes and BTF passes. I think the file reads a bit cleaner now as well.

danobi:
Will merge tomorrow if no other comments

fbs:
It looks ok to me, but I'm not familiar with the clang api. But builds pass so :)

mmisono:
It seems there is a case that `--include` and `-b` does not work together. For example:

```
% sudo ./src/bpftrace  -e \
'tracepoint:syscalls:sys_enter_writev { @ = args->vec->iov_len; }
kprobe:vfs_open { @a =  ((struct path *)arg0)->dentry; }' \
--include /usr/include/linux/uio.h -b
/usr/include/linux/uio.h:20:2: error: unknown type name '__kernel_size_t'
```

Here, `--include` is needed for tracepoint (`struct io_vec`) and `-b` for kprobe (`struct path`) (currently FieldAnalyzer does not analyze tracepoint structs, so we need to manually define or include structs used in tracepoints). This works on master.

danobi:
Thanks for catching that @mmisono. I think the right fix here is to analyze the tracepoint struct correctly so we can avoid the `--include`.

mmisono:
@danobi 
> I think the right fix here is to analyze the tracepoint struct correctly so we can avoid the --include.

Agreed. Previously, I tried to do it: https://github.com/mmisono/bpftrace/commit/953367f433cad26772ea0ee3281cf5695a486db3
For now, I don't have enough time to review this to see if this works with this PR. I'll do it at night (~12hr later from now), or please feel free to use the commit if you want to (I'm not sure the above is the best way to do this).

By the way, does this PR prohibit using both `--include` and `-b` simultaneously? I think sometimes we want do `--include` to get `#define`-d value even if BTF is available.

danobi:
> I'm not sure the above is the best way to do this

I pushed a commit that fixes the issue in 2 LoC. I think it should cover all the cases too.

> By the way, does this PR prohibit using both --include and -b simultaneously?

It does not. But it is possible to run into type issues because we're forcing exclusion of <uapi/linux/types.h>. I'm hopeful that we (FB kernel team & I) can fix the #define flag issue in the kernel. There's been some more discussion upstream since I posted my RFC.

mmisono:
> I pushed a commit that fixes the issue in 2 LoC. 

oh great, I confirmed it worked. :)

> It does not. But it is possible to run into type issues because we're forcing exclusion of <uapi/linux/types.h>.

I see. Thanks for the explanation. I don't see any problems for now.

PR #1269, opened 2020-04-16
Implement sizeof() function

### LINKED ISSUE #73: support sizeof
Add support for sizeof().

```
# bpftrace -e 'BEGIN { printf("%d\n", sizeof(int)); }'
1.34: syntax error, unexpected ), expecting (
```

It will be handy to have, but not critical.

danobi:
It's often useful to know the size of a struct. As a bonus, this
implementation also works with all other known types and
expressions. eg:

    struct Foo
    {
        int x;
    }

    BEGIN
    {
        printf("%d\n", sizeof(struct Foo));
        printf("%d\n", sizeof(1 == 1));
        printf("%d\n", sizeof(((struct Foo)0).x));
    }

This closes #73 .

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -84,6 +84,7 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [19. `strncmp()`: Compare first n characters of two strings](#19-strncmp-compare-first-n-characters-of-two-strings)
     - [20. `override()`: Override return value](#20-override-override-return-value)
     - [21. `buf()`: Buffers](#21-buf-buffers)
+    - [22. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
 - [Map Functions](#map-functions)
     - [1. Builtins](#1-builtins-2)
     - [2. `count()`: Count](#2-count-count)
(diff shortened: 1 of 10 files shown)


tests/runtime/engine/utils.py
```diff
@@ -9,6 +9,7 @@
 import re
 
 BPF_PATH = environ["BPFTRACE_RUNTIME_TEST_EXECUTABLE"]
+ENV_PATH = environ["PATH"]
```
danobi, line 12: This is stolen from #1268 . Whichever PR goes in second I'll take it out from.

danobi:
cc @olsajiri for the BTF fix

docs/reference_guide.md
```diff
@@ -2538,6 +2539,38 @@ Datagram bytes: \x08\x00+\xb9\x06b\x00\x01Aen^\x00\x00\x00\x00KM\x0c\x00\x00\x00
 rtt min/avg/max/mdev = 19.426/19.426/19.426/0.000 ms
 ```
 
+## 22. `sizeof()`: Size of type or expression
+
+Syntax:
+- `sizeof(TYPE)`
+- `sizeof(EXPRESSION)`
+
+Returns size of the argument in bytes. Similar to C/C++ `sizeof` operator.
```
fbs, line 2548:
```suggestion
Returns size of the argument in bytes. Similar to C/C++ `sizeof` operator. Note that the expression does not get evaluated.
```

PR #1273, opened 2020-04-16
Clean up README

danobi:
Smaller READMEs are nicer for learning about a new project.

This commit:
* Removes mini language reference. This was getting outdated and we the
  reference guide is our source of truth anyways.
* Removes examples. We have plenty of examples in the one-liners and
  tools section.
* Remove internals graphic. This was outdated. It's also present in the
  internal development doc.


fbs:
Maybe place the oneliners before the install section? That way they're first thing you see.

PR #1274, opened 2020-04-16
btf: Allow referencing an enum value to pull in enum definition

danobi:
It's useful to be able to reference an enum value to have bpftrace pull
in the enum definition from BTF. This is especially useful for anonymous
enums which cannot be correctly referenced.


danobi:
This PR depends on #1268 and #1269. Will rebase away all but the last commit when the previous ones are in.

olsajiri:
nice result of that refactoring you did before, I don't see any problem

Issue #464, opened 2019-03-01
need to warn when exceeding map key limits


tyroguru:
This may well be an existing issue as I remember reading something from @brendangregg  on this recently in an issue. Please close this if it is a duplicate.

Currently, associative arrays are statically sized to a hard coded value ('max_entries' in map.cpp which is currently 128). When a script exceeds this number of active entries a  write to the hash appears to work and a read from that hash entry returns 0. An example with map.cpp:max_entries set to 8:

```
# cat broken-maps.bt
BEGIN
{
        @["1"] = nsecs;
        @["2"] = nsecs;
        @["3"] = nsecs;
        @["4"] = nsecs;
        @["5"] = nsecs;
        @["6"] = nsecs;
        @["7"] = nsecs;
        @["8"] = nsecs;
        @["9"] = nsecs;
        @["10"] = nsecs;

        printf("nsecs = %llu\n", @["10"]);
}

# ./bpftrace ./broken-maps.bt
Attaching 1 probe...
nsecs = 0
^C

@[1]: 1380839892949659
@[2]: 1380839892958637
@[3]: 1380839892959973
@[4]: 1380839892961084
@[5]: 1380839892962178
@[6]: 1380839892963548
@[7]: 1380839892964913
@[8]: 1380839892966208

```

Increasing the size of 'max_entries' would be good but it is just a band-aid. The silent write and a read returning 0 causes logic errors in scripts that can be extremely difficult to debug.

If the number of entries is to remain static then we must inform the user that the size has been exceeded and allow them to tune the map size accordingly. Obviously, the most preferable solution would be to have a dynamic map size.

brendangregg:
There was another case where we increased the defaults for stack maps. I think in this case we should make that 1024 (at least) to start with, and follow up to see how to print warnings when exceeded.

tyroguru:
Re; the "enhancement" label. Of course, any bug can be seen as an enhancement of current behaviour :-) . However, this is actually a bug and quite a nasty one at that as it leads to highly misleading script behaviour owing to the way we just silently drop the write and return 0 for a read. Really needs fixing ASAP.

brendangregg:
changed title. At the very least, on program termination we could iterate over all maps and if any has a key count that equals max_entries then print:

```WARNING: map %s has reached the BPFTRACE_MAP_KEYS_MAX limit, possible lost entries.```

[...]

PR #1276, opened 2020-04-17
Report helper function errors

mmisono:
Some helper functions return error codes when failing, but currently,
bpftrace does not check and discards it. Several times I had to use bcc
to check error codes for debugging.

This patch adds an error checking mechanism, and helper function errors
are reported in runtime using '-k' option. For example:

```
% sudo BPFTRACE_MAP_KEYS_MAX=2 ./src/bpftrace -e \
  'BEGIN { @[0] = 0; @[1] = 1; @[2] = 2;} -k'
Attaching 1 probe...
stdin:1:35-36: WARNING: Failed to map_update_elem: Argument list too long (-7)
BEGIN { @[0] = 0; @[1] = 1; @[2] = 2;}
                                 ~

^C

@[0]: 0
@[1]: 1
```

This mechanism is implemented as an async action. By default, errors of
`bpf_probe_read()`-related functions and `bpf_map_lookup_elem()` are not
checked. This is because when these functions fail, bpftrace returns
zero as a result, and some scripts already check it, for example, by
using predicate `/ @[tid] /`. `-kk` option forces to check all errors
including these functions.

Note that the error check increases code size. All codegen tests are
performed with the error check disabled.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -144,6 +144,8 @@ OPTIONS:
     -p PID         enable USDT probes on PID
     -c 'CMD'       run CMD and enable USDT probes on resulting process
     -v             verbose messages
+    -k             emit a warning when a bpf helper returns an error (except read functions)
+    -kk            check all bpf helper functions
     --version      bpftrace version
 
 ENVIRONMENT:
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1,9 +1,12 @@
 #include <iostream>
 
-#include "irbuilderbpf.h"
 #include "arch/arch.h"
+#include "ast/async_event_types.h"
+#include "bpftrace.h"
+#include "irbuilderbpf.h"
 #include "utils.h"
 
+#include <llvm/IR/DataLayout.h>
 #include <llvm/IR/Module.h>
 
 namespace libbpf {
@@ -212,22 +215,33 @@ CallInst *IRBuilderBPF::createMapLookup(int mapfd, AllocaInst *key)
   return CreateCall(lookup_func, { map_ptr, key }, "lookup_elem");
 }
 
-CallInst *IRBuilderBPF::CreateGetJoinMap(Value *ctx __attribute__((unused)))
+CallInst *IRBuilderBPF::CreateGetJoinMap(Value *ctx, const location &loc)
 {
   AllocaInst *key = CreateAllocaBPF(getInt32Ty(), "key");
   CreateStore(getInt32(0), key);
 
   CallInst *call = createMapLookup(bpftrace_.join_map_->mapfd_, key);
+  CreateHelperErrorCond(ctx, call, libbpf::BPF_FUNC_map_lookup_elem, loc, true);
   return call;
 }
 
-Value *IRBuilderBPF::CreateMapLookupElem(Map &map, AllocaInst *key) {
+Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
+                                         Map &map,
+                                         AllocaInst *key,
+                                         const location &loc)
+{
+  assert(ctx && ctx->getType() == getInt8PtrTy());
   int mapfd = bpftrace_.maps_[map.ident]->mapfd_;
-  return CreateMapLookupElem(mapfd, key, map.type);
+  return CreateMapLookupElem(ctx, mapfd, key, map.type, loc);
 }
 
-Value *IRBuilderBPF::CreateMapLookupElem(int mapfd, AllocaInst *key, SizedType &type)
+Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
+                                         int mapfd,
+                                         AllocaInst *key,
+                                         SizedType &type,
+                                         const location &loc)
 {
+  assert(ctx && ctx->getType() == getInt8PtrTy());
   CallInst *call = createMapLookup(mapfd, key);
 
   // Check if result == 0
@@ -265,6 +279,7 @@ Value *IRBuilderBPF::CreateMapLookupElem(int mapfd, AllocaInst *key, SizedType &
     CREATE_MEMSET(value, getInt8(0), type.size, 1);
   else
     CreateStore(getInt64(0), value);
+  CreateHelperError(ctx, getInt32(0), libbpf::BPF_FUNC_map_lookup_elem, loc);
   CreateBr(lookup_merge_block);
 
   SetInsertPoint(lookup_merge_block);
@@ -274,10 +289,15 @@ Value *IRBuilderBPF::CreateMapLookupElem(int mapfd, AllocaInst *key, SizedType &
   return CreateLoad(value);
 }
 
-void IRBuilderBPF::CreateMapUpdateElem(Map &map, AllocaInst *key, Value *val)
+void IRBuilderBPF::CreateMapUpdateElem(Value *ctx,
+                                       Map &map,
+                                       AllocaInst *key,
+                                       Value *val,
+                                       const location &loc)
 {
   Value *map_ptr = CreateBpfPseudoCall(map);
 
+  assert(ctx && ctx->getType() == getInt8PtrTy());
   assert(key->getType()->isPointerTy());
   assert(val->getType()->isPointerTy());
 
@@ -294,11 +314,18 @@ void IRBuilderBPF::CreateMapUpdateElem(Map &map, AllocaInst *key, Value *val)
       Instruction::IntToPtr,
       getInt64(libbpf::BPF_FUNC_map_update_elem),
       update_func_ptr_type);
-  CreateCall(update_func, { map_ptr, key, val, flags }, "update_elem");
+  CallInst *call = CreateCall(update_func,
+                              { map_ptr, key, val, flags },
+                              "update_elem");
+  CreateHelperErrorCond(ctx, call, libbpf::BPF_FUNC_map_update_elem, loc);
 }
 
-void IRBuilderBPF::CreateMapDeleteElem(Map &map, AllocaInst *key)
+void IRBuilderBPF::CreateMapDeleteElem(Value *ctx,
+                                       Map &map,
+                                       AllocaInst *key,
+                                       const location &loc)
 {
+  assert(ctx && ctx->getType() == getInt8PtrTy());
   assert(key->getType()->isPointerTy());
   Value *map_ptr = CreateBpfPseudoCall(map);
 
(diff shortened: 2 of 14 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -812,5 +913,93 @@ Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
...
+  BasicBlock *helper_success_block = BasicBlock::Create(module_.getContext(),
+                                                        "helper_success",
+                                                        parent);
+  BasicBlock *helper_failure_block = BasicBlock::Create(module_.getContext(),
+                                                        "helper_failure",
+                                                        parent);
+  BasicBlock *helper_merge_block = BasicBlock::Create(module_.getContext(),
+                                                      "helper_merge",
+                                                      parent);
+  auto *ret = CreateIntCast(return_value, getInt32Ty(), true);
```
mmisono, line 989: I found that a return type of most helper functions was Int64Ty but they actually returned int and we needed to use Int32Ty value to check if a return value was negative. I cast the value here to keep code changes of this PR as small as possible.

mmisono:
I think this solves #464.

I forgot to mention but this uses perf_event_output() for error reporting, so an error of perf_event_output() is not reported. An error message is printed using `bpftrace.error()` and is not included in json if an output format is json.

mmisono:
This is the comparison of processed insn sizes (not the actual code size. Since the verifier explorers all possible code path, processed insn size becomes larger than the code size). Multiple values mean that the scripts have multiple functions.

So, basically error checking increases the number of processed insn by tens or hundreds, but some scripts (especially which have many map updates) increase code size by thousands (e.g., capable.bt). I also measured the compilation times but I didn't see noticeable differences with and without error checking. I didn't measure runtime speed.

I'm still wondering what is the best default behavior, and I'd appreciate any feedback.

| tools             | default                            | --check-all-helper                 | --disable-helper-check             |
| -------           | ---------                          | --------------------               | ------------------------           |
| tcpsynbl.bt       | 14,14,185,185                      | 14,14,478,478                      | 14,14,139,139                      |
| bashreadline.bt   | 161,14                             | 161,14                             | 161,14                             |
| tcpaccept.bt      | 394,391                            | 394,1166                           | 394,331                            |
| tcpconnect.bt     | 362,14                             | 362,14                             | 362,14                             |
| runqlen.bt        | 14,92                              | 14,251                             | 14,70                              |
| writeback.bt      | 2094,34,516,14                     | 2094,34,546,14                     | 828,14,453,14                      |
| setuids.bt        | 270,154,154,272,200,201,314        | 270,154,154,272,298,299,513        | 270,45,45,71,123,124,197           |
| mdflush.bt        | 149,14                             | 149,14                             | 149,14                             |
| loads.bt          | 14,66                              | 14,215                             | 14,66                              |
| biosnoop.bt       | 159,14,253,198                     | 159,14,352,198                     | 159,14,182,42                      |
| cpuwalk.bt        | 14,62                              | 14,77                              | 14,41                              |
| opensnoop.bt      | 211,36,36,266,266,14               | 211,36,36,318,318,14               | 211,15,15,193,193,14               |
| capable.bt        | 10463,14,365                       | 10515,14,379                       | 3382,14,335                        |
| naptime.bt        | 256,182                            | 256,633                            | 256,154                            |
| dcsnoop.bt        | 193,175,14,42                      | 193,223,14,86                      | 193,121,14,21                      |
| biostacks.bt      | 14,14,278,278,132                  | 14,14,341,341,132                  | 14,14,185,185,27                   |
| xfsdist.bt        | 14,250,250,250,250,14,94,94,94,94  | 14,305,305,305,305,14,94,94,94,94  | 14,150,150,150,150,14,30,30,30,30  |
| pidpersec.bt      | 14,42,41,14                        | 14,57,41,14                        | 14,21,41,14                        |
| biolatency.bt     | 14,14,191,34                       | 14,14,236,34                       | 14,14,134,14                       |
| runqlat.bt        | 14,36,36,363,14                    | 14,36,36,388,14                    | 14,16,16,135,14                    |
| tcpretrans.bt     | 3262,14,875                        | 3262,14,1769                       | 1229,14,875                        |
| threadsnoop.bt    | 117,14                             | 117,14                             | 117,14                             |
| statsnoop.bt      | 181,36,36,36,36,161,161,161,161,14 | 181,36,36,36,36,195,195,195,195,14 | 181,15,15,15,15,108,108,108,108,14 |
| tcpdrop.bt        | 3278,14,975                        | 3279,14,1915                       | 1246,14,894                        |
| vfsstat.bt        | 14,41,14,42,42,42,42,42,42,42,42   | 14,41,14,57,57,57,57,57,57,57,57   | 14,41,14,21,21,21,21,21,21,21,21   |
| oomkill.bt        | 14,14                              | 14,14                              | 14,14                              |
| syncsnoop.bt      | 177,70,72,70,72,72,72              | 177,70,72,70,72,72,72              | 177,43,44,43,44,44,44              |
| syscount.bt       | 14,162,14                          | 14,203,14                          | 14,53,14                           |
| swapin.bt         | 41,14                              | 41,14                              | 41,14                              |
| execsnoop.bt      | 101,218                            | 101,707                            | 101,218                            |
| killsnoop.bt      | 227,93,171                         | 227,93,270                         | 227,29,115                         |
| bitesize.bt       | 14,198,14                          | 14,221,14                          | 14,177,14                          |
| gethostlatency.bt | 193,14,216,216,216,94,94           | 193,14,275,275,275,94,94           | 193,14,163,163,163,30,30           |
| tcplife.bt        | 490,14,1015                        | 490,14,2110                        | 490,14,635                         |

docs/reference_guide.md
```diff
@@ -141,6 +141,8 @@ OPTIONS:
     -p PID         enable USDT probes on PID
     -c 'CMD'       run CMD and enable USDT probes on resulting process
     -v             verbose messages
+    --check-all-helper     check return codes of all helper functions
```
fbs, line 144:
```suggestion
    --check-helper-errors    emit an error when a bpf helper returns an error
```

I think this is a bit more descriptive.

Also alignment :)

src/ast/irbuilderbpf.cpp
```diff
@@ -812,5 +913,93 @@ Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
...
 }
 
+static bool is_probe_read_funcs(libbpf::bpf_func_id func_id)
+{
+  static auto probe_read_funcs = {
+    libbpf::BPF_FUNC_probe_read,        libbpf::BPF_FUNC_probe_read_str,
+    libbpf::BPF_FUNC_probe_read_kernel, libbpf::BPF_FUNC_probe_read_kernel_str,
+    libbpf::BPF_FUNC_probe_read_user,   libbpf::BPF_FUNC_probe_read_user_str
+  };
+  return std::find(std::begin(probe_read_funcs),
```
fbs, line 923: Using a simple case statement might be more readable than this, e.g:

```
switch (func_id) {
  case libbpf::BPF_FUNC_probe_read:
  case libbpf::BPF_FUNC_probe_read_kernel:
    return true
  default:
    return false;
}
```

src/ast/irbuilderbpf.cpp
```diff
@@ -812,5 +913,93 @@ Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
...
+void IRBuilderBPF::CreateHelperError(Value *ctx,
+                                     Value *return_value,
+                                     libbpf::bpf_func_id func_id,
+                                     const location &loc)
+{
+  assert(ctx && ctx->getType() == getInt8PtrTy());
+  assert(return_value && return_value->getType() == getInt32Ty());
+  if (bpftrace_.disable_helper_check_)
+    return;
+  if (!bpftrace_.check_all_helper_ &&
```
fbs, line 937: This could use a comment (next to the one in the commit msg)

src/ast/irbuilderbpf.cpp
```diff
@@ -812,5 +913,93 @@ Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
...
+  BasicBlock *helper_merge_block = BasicBlock::Create(module_.getContext(),
+                                                      "helper_merge",
+                                                      parent);
+  auto *ret = CreateIntCast(return_value, getInt32Ty(), true);
+  Value *condition;
+  if (compare_zero)
+    condition = CreateICmpNE(ret, Constant::getNullValue(ret->getType()));
+  else
+    condition = CreateICmpSGE(ret, Constant::getNullValue(ret->getType()));
+  CreateCondBr(condition, helper_success_block, helper_failure_block);
```
fbs, line 995:
```suggestion
  CreateCondBr(condition, helper_merge_block, helper_failure_block);
```
Don't think the success block is needed.

src/main.cpp
```diff
@@ -57,14 +57,16 @@ void usage()
   std::cerr << "    --unsafe       allow unsafe builtin functions" << std::endl;
   std::cerr << "    -v             verbose messages" << std::endl;
   std::cerr << "    --info         Print information about kernel BPF support" << std::endl;
+  std::cerr << "    --check-all-helper     check return codes of all helper functions" << std::endl;
```
fbs, line 60: same comment and alignment thing here

fbs:
Nice, this is useful :)

Seems like the processed instruction count is actually reduced in the `--disable-helper-check` case, what is going on there?

mmisono:
Thanks for the review, will fix.

"default" in the above table means the default behavior when this patch applied (so it checks errors other than bpf_probe_read_xxx / bpf_map_lookup_elem). `--disable-helper-check` does not check error at all, and this is the same behavior as the current.

mmisono:
I examined the generated code. Basically, this error check creates one branch and adds 10 instructions per helper function call. A problem is a branch creates another state and the verifier needs to explore paths both that take a branch and not. This is the reason the processed instructions sharply increase in scripts that call many helper functions. The verifier prunes a search if two states are equivalent, but unfortunately not.

I'm wondering if we can implement this functionality using [bpf-to-bpf call](https://lwn.net/Articles/741773/). I guess it reduces states. I'll give it a try later. (there is little information about this..)

mmisono:
I fixed the reviewed points and changed the option since the initial version was a bit confusing:

- `-k` : emit a warning when a bpf helper returns an error, except read functions (bpf_probe_read_xxx(), bpf_map_lookup_elem()).
   - The reason excludes these functions is described in the commit message.
- `-kk`: check all helper function errors.

Therefore, if these options are not given, program behavior is the same as the current. What do you think about this? @fbs

(As I wrote, I'm seeing if bpf-to-bpf call is usable, but I don't plan to include it in this PR even if it works.)

src/bpftrace.h
```diff
@@ -145,6 +151,7 @@ class BPFtrace
   std::vector<std::string> join_args_;
   std::vector<std::string> time_args_;
   std::vector<std::tuple<std::string, std::vector<Field>>> cat_args_;
+  std::map<int64_t, struct HelperErrorInfo> helper_error_info_;
```
danobi, line 154:
```suggestion
  std::unordered_map<int64_t, struct HelperErrorInfo> helper_error_info_;
```
Do we need the ordering? unordered_map should be faster lookups

src/ast/irbuilderbpf.cpp
danobi: Can you leave that as a comment in the code? Will help next person who wonders why there's a cast.

danobi:
Nice, this looks really useful. Code looks good to me. Can you add some codegen tests with the error checks and also a runtime test or two? Just so we don't accidentally break this later.

Sorry about the delay, didn't see this PR. In the future feel free to ping me if PR get delayed too much.

mmisono:
Updated:
- Rebased onto master
- Use unordered map instead of map
- Add comments about IntCast
- Add codegen and runtime tests

Please take a look when you have time @danobi

tests/runtime/other
```diff
@@ -174,3 +174,8 @@ NAME map_assign_map_ptr
 RUN bpftrace -e 'i:ms:100 { @ = curtask; @a = @; printf("%d\n", @a); exit(); }'
 EXPECT -?[0-9]+
 TIMEOUT 1
+
+NAME runtime_error_check
+RUN bpftrace -kk -e 'i:ms:100 { @[1] = 1; printf("%d\n", @[2]); exit(); }'
+EXPECT 0
```
danobi, line 180: Wouldn't it make more sense to look for the warning message?

mmisono: Yes, I updated it. (I didn't realize that the runtime engine redirected stderr to stdout and simply did regex match using EXPECT.)

PR #1279, opened 2020-04-17
Move process monitoring out of `BPFtrace`

fbs:
This moves the process monitoring logic out of bpftrace itself and
instead provide a simple interface bpftrace can use.

If pidfds are supported by the system they're used, else it falls back
to polling. Detection is done at runtime.

The way the 'pid to trace' and 'child pid' are accessed through the
bpftrace class is still a bit messy and confusing. But that's for later.

### DIFF
--- a/src/procmon.cpp
+++ b/src/procmon.cpp
@@ -1,0 +1,116 @@
+#include <cerrno>
+#include <fcntl.h>
+#include <poll.h>
+#include <stdexcept>
+#include <sys/syscall.h>
+#include <system_error>
+#include <unistd.h>
+
+#include "procmon.h"
+#include "utils.h"
+
+namespace bpftrace {
+
+#ifndef __NR_pidfd_open
+#define __NR_pidfd_open 434
+#endif
+
+static std::system_error SYS_ERROR(std::string msg)
+{
+  return std::system_error(errno, std::generic_category(), msg);
+}
+
+static inline int pidfd_open(int pid, unsigned int flags)
+{
+  return syscall(__NR_pidfd_open, pid, flags);
+}
+
+ProcMon::ProcMon(const std::string& pid)
+{
+  setup(parse_pid(pid));
+}
+
+ProcMon::ProcMon(pid_t pid)
+{
+  setup(pid);
+}
+
+void ProcMon::setup(pid_t pid)
+{
+  pid_ = pid;
+
+  int pidfd = pidfd_open(pid, 0);
+  // Fall back to polling if pidfds or anon inodes are not supported
+  if (pidfd >= 0)
+  {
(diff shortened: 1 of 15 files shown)


src/procmon.cpp
```diff
@@ -0,0 +1,116 @@
...
+  setup(pid);
+}
+
+void ProcMon::setup(pid_t pid)
+{
+  pid_ = pid;
+
+  int pidfd = pidfd_open(pid, 0);
+  // Fall back to polling if pidfds or anon inodes are not supported
+  if (pidfd < 0 && errno != ENOSYS && errno != ENODEV)
```
brauner, line 44: Just because this caught my eye, we've made anonymous inodes unconditional even prior to landing `pidfd_open()`:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5dd50aaeb1853ee0953b60fa6d1143d95429ae7b

fbs: Ah thanks. Is nodev something that can happen or is that a docs issue? I based my code on http://man7.org/linux/man-pages/man2/pidfd_open.2.html

brauner:
On April 18, 2020 4:16:09 PM GMT+02:00, bas smit <notifications@github.com> wrote:
>@fbs commented on this pull request.
>
>
>
>> +{
>+  setup(parse_pid(pid));
>+}
>+
>+ProcMon::ProcMon(pid_t pid)
>+{
>+  setup(pid);
>+}
>+
>+void ProcMon::setup(pid_t pid)
>+{
>+  pid_ = pid;
>+
>+  int pidfd = pidfd_open(pid, 0);
>+  // Fall back to polling if pidfds or anon inodes are not supported
>+  if (pidfd < 0 && errno != ENOSYS && errno != ENODEV)
>
>Ah thanks. Is nodev something that can happen or is that a docs issue?
>I based my code on
>http://man7.org/linux/man-pages/man2/pidfd_open.2.html 

Hm, I think the manpage is wrong on this one, i.e. it shouldn't happen.
Sigh, hopefully I'll find time to update it or just report to Michael that it's not up to date.

src/procmon.cpp
brauner: Replying by mail sucks apparently so let me comment here again :)
ENODEV can't happen with `pidfd_open()`. It has been released with 5.3 while we made anonymous inodes unconditional in 5.2. So manpages need updating it seems. :)

fbs: Clear, I'll update it. Thanks for the review (and for pidfds) ! :)

brauner: Thank your for the bpftrace tool. I can't tell you how often I'll help it to debug bugs in my kernel patches. :)

PR #1280, opened 2020-04-17
Move CI over to GH actions

danobi:
Travis is slower and has a clunkier integration with github. GH actions also supports way more features and has community plugins.

This also unblocks #1255 to use GH actions to remind people to update changelogs.


danobi:
Will squash before merging.

danobi:
cc @dalehamel

PR #1284, opened 2020-04-18
Install C++17 compatible compiler on xenial embedded builds

danobi:
The build is failing on missing C++17 support. Xenial is still using
GCC5 which is too old. According to:
https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.201z ,
std::optional was not available until gcc7-ish.

### DIFF
--- a/docker/Dockerfile.ubuntu-glibc
+++ b/docker/Dockerfile.ubuntu-glibc
@@ -63,6 +63,16 @@ RUN if [ "${DISTRO}" = "xenial" ]; then \
     apt install -y cmake \
     ; fi
 
+# Install C++17 compatible compiler if running on xenial
+RUN if [ "${DISTRO}" = "xenial" ]; then \
+    apt-get install -y software-properties-common \
+    && add-apt-repository -y ppa:ubuntu-toolchain-r/test \
+    && apt update \
+    && apt install -y g++-7 \
+    && update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 \
+                           --slave /usr/bin/g++ g++ /usr/bin/g++-7 \
+    ; fi
+
 # Build BCC and install static libs
 RUN mkdir -p /src && git clone https://github.com/$bcc_org/bcc /src/bcc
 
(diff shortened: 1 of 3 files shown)


danobi:
@dalehamel

dalehamel:
The implication of dropping this is that older distros won’t be supported, as the glibc 2.23 is the real goal of this build.

If we drop this, we could ideally have a solution for still building for older glibc

Also cc @alexeldeib as he wrote this one

danobi:
We could try to install a newer GCC in the container: https://gist.github.com/jlblancoc/99521194aba975286c80f93e47966dc5

danobi:
Took a while but I finally got this to work. Ready for review now.

docker/Dockerfile.ubuntu-glibc
```diff
@@ -63,6 +63,17 @@ RUN if [ "${DISTRO}" = "xenial" ]; then \
...
 
+# Install C++17 compatible compiler if running on xenial
+RUN if [ "${DISTRO}" = "xenial" ]; then \
+    apt-get install -y software-properties-common \
+    && add-apt-repository ppa:ubuntu-toolchain-r/test \
+    && apt update \
+    && apt install g++-7 -y \
+    && update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 60 \
+                           --slave /usr/bin/g++ g++ /usr/bin/g++-7 \
+    && update-alternatives --config gcc \
```
fbs, line 74: Don't you need:

```
       --set name path
              Set the program path as alternative for name.  This is equivalent to --config but is non-interactive and thus scriptable.
```

Instead of:


```
       --config name
              Show available alternatives for a link group and allow the user to interactively select which one to use. The link group is updated.
```

danobi: Ah, seems this isn't even necessary. When I tried it in a vm:
```
$ update-alternatives --config gcc
There is only one alternative in link group gcc (providing /usr/bin/gcc): /usr/bin/gcc-7
Nothing to configure.
```

danobi:
Going to merge this b/c I'm tired of getting emails about broken builds. Feel free to revert if there's issues.

dalehamel:
Lgtm anwyays thats @danobi

alexeldeib:
Thanks a lot for making this work on older builds @danobi!

PR #1286, opened 2020-04-19
Unroll improvements

mmisono:
Add support of using positional params in unroll and increase the unroll limit to 100.

Example:

```
% sudo ./src/bpftrace -e 'BEGIN { unroll($1) { printf("hi\n"); } }' 5
Attaching 1 probe...
hi
hi
hi
hi
hi
^C
```

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1408,10 +1408,43 @@ void SemanticAnalyser::visit(If &if_block)
 
 void SemanticAnalyser::visit(Unroll &unroll)
 {
+  unroll.expr->accept(*this);
+
+  unroll.var = 0;
+
+  if (auto *integer = dynamic_cast<Integer *>(unroll.expr))
+  {
+    unroll.var = integer->n;
+  }
+  else if (auto *param = dynamic_cast<PositionalParameter *>(unroll.expr))
+  {
+    if (param->ptype == PositionalParameterType::count)
+    {
+      unroll.var = bpftrace_.num_params();
+    }
+    else
+    {
+      std::string pstr = bpftrace_.get_param(param->n, param->is_in_str);
+      if (is_numeric(pstr))
+        unroll.var = std::stoll(pstr, nullptr, 0);
+      else
+        error("Invalid positonal params: " + pstr, unroll.loc);
+    }
+  }
+  else
+  {
+    out_ << "Unsupported expression" << std::endl;
+    abort();
+  }
+
   if (unroll.var > 20)
-    error("unroll maximum value is 20", location(0));
+  {
+    error("unroll maximum value is 20", unroll.loc);
+  }
   else if (unroll.var < 1)
-    error("unroll minimum value is 1", location(0));
+  {
+    error("unroll minimum value is 1", unroll.loc);
+  }
 
   for (int i = 0; i < unroll.var; i++)
     accept_statements(unroll.stmts);
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1437,9 +1437,9 @@ void SemanticAnalyser::visit(Unroll &unroll)
     abort();
   }
 
-  if (unroll.var > 20)
+  if (unroll.var > 100)
   {
-    error("unroll maximum value is 20", unroll.loc);
+    error("unroll maximum value is 100", unroll.loc);
   }
   else if (unroll.var < 1)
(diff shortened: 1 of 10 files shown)


danobi:
Do you think it would work to rewrite `unroll()` as a while loop (based on #1066)? I think it would result in less byte code. I can't think of any reasons why it would break existing programs.

danobi:
Hmm except on systems without loop support.

src/parser.yy
```diff
@@ -219,8 +219,8 @@ stmts : semicolon_ended_stmt stmts { $$ = $2; $2->insert($2->begin(), $1); }
       |                            { $$ = new ast::StatementList; }
       ;
 
-block_stmt : if_stmt                  { $$ = $1; }
-           | UNROLL "(" INT ")" block { $$ = new ast::Unroll($3, $5); }
+block_stmt : if_stmt                           { $$ = $1; }
+           | UNROLL "(" int_or_param ")" block { $$ = new ast::Unroll($3, $5, @1 + @4); }
```
fbs, line 223:
```suggestion
           | UNROLL "(" int ")" block { $$ = new ast::Unroll($3, $5, @1 + @4); }
            | UNROLL "(" param ")" block { $$ = new ast::Unroll($3, $5, @1 + @4); }
```

Currently we have the `INT` (lexer) token and the `int` symbol in the parser which handles negative ints too. This would be a `positive_int_or_param`, which I think would add more confusion than it solves. Additionally this moves the detection of negative numbers to the semantic analyser, which can produce nicer errors:

```
stdin:1:17-28: ERROR: unroll minimum value is 1
i:s:1 { $a= 0 ; unroll(-10) { printf("%d\n", $a++); } exit() }
                ~~~~~~~~~~~
```

Compared to

```
stdin:1:17-25: ERROR: syntax error, unexpected -, expecting $# or positional parameter or integer
i:s:1 { $a= 0 ; unroll(-10) { printf("%d\n", $a++); } exit() }
                ~~~~~~~~
```

mmisono:
Fixed parser.yy. Thanks @fbs
probe.interval_order tests failed for some reason, but I believe it wasn't related to this PR.

------

> Do you think it would work to rewrite unroll() as a while loop (based on #1066)? 

I think it would, but potentially unrolling optimize an execution speed so I guess keeping unrolling as is is good.

fbs:
>  probe.interval_order tests failed for some reason, but I believe it wasn't related to this PR.

It always was a bit flaky but the new CI seems to have made it worse, got a commit to remove it in #1066

src/ast/semantic_analyser.cpp
```diff
@@ -1406,16 +1406,46 @@ void SemanticAnalyser::visit(If &if_block)
...
+      if (is_numeric(pstr))
+        unroll.var = std::stoll(pstr, nullptr, 0);
+      else
+        error("Invalid positonal params: " + pstr, unroll.loc);
+    }
+  }
+  else
+  {
+    out_ << "Unsupported expression" << std::endl;
+    abort();
```
fbs, line 1435:
```suggestion
error("unroll can only be used..", unroll.loc);
```

fbs:
Bit curious where the 100 came from (or 20 for that matter). What is your usecase here? :)

src/ast/semantic_analyser.cpp
mmisono: The intent of `abort()` here is indicating this is a bug if this code is executed (this shouldn't happen due to the grammar). Abort is useful when fuzzing because most fuzzers treat SIGABRT as a bug and report it.

mmisono:
Actually there is no particular reason for 100, I just want to benchmark bpftrace (and BPF program) performance and unrolling is an easy way to generate long instructions. I feel 20 is a bit small.  Maybe introducing `BPFTRACE_MAX_UNROLL` env variables is better. I think the limiting is useful because it helps us to prevent unrolling a large number of loop which consequently causes a stack overflow or something (this is especially useful when fuzzing since fuzzer generates unexpected inputs)

By the way, it is easy to circumvent unroll limitation: `unroll(10) { unroll(10) { $i++; } }`.

src/ast/semantic_analyser.cpp
fbs: oh yes, brainfart

fbs:
Reminder to add a changelog entry for this (#1255) before merging

mmisono:
Rebased onto master and updated a changelog.

PR #1287, opened 2020-04-19
Update vagrant config with ubuntu 16.04

danobi:
It's useful to have xenial in the vagrant config since we support ubuntu
16.04 in embedded builds.

Also use llvm/clang 8 in all ubuntu VMs since 16.04/18.04/19.10 all have
llvm/clang 8.

Also update README with more instructions.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,11 +5,11 @@
 #
 # SKIP_BCC_BUILD: Set to skip the building bcc from source
 
-$ubuntu_18_deps = <<EOF
+$ubuntu_deps = <<EOF
 apt-get -qq update
 apt-get -qq install linux-headers-$(uname -r) binutils-dev
 apt-get -qq install bison cmake flex g++ git libelf-dev zlib1g-dev libfl-dev systemtap-sdt-dev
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -5,11 +5,11 @@
 #
 # SKIP_BCC_BUILD: Set to skip the building bcc from source
 
-$ubuntu_18_deps = <<EOF
+$ubuntu_deps = <<EOF
 apt-get -qq update
 apt-get -qq install linux-headers-$(uname -r) binutils-dev
 apt-get -qq install bison cmake flex g++ git libelf-dev zlib1g-dev libfl-dev systemtap-sdt-dev
```
fbs, line 11: Does this work on xenial? #1284 does a bit of effort to get gcc 7 all set up

danobi: Ah crap it probably does. What a day well spent yesterday.

danobi: Hmm well the other builds are using g++ so maybe it'd be better to stick to g++ for all releases for now.

danobi:
Gonna merge b/c it's just dev stuff.

PR #1292, opened 2020-04-20
Push docker image to quay.io only when updating iovisor/bpftrace:master

mmisono:
Other cases do not have a right to complete the action.

### DIFF
--- a/.github/workflows/embedded.yml
+++ b/.github/workflows/embedded.yml
@@ -109,11 +109,11 @@ jobs:
         path: build-embedded/tests/bpftrace_test
 
     - name: Authenticate with docker registry
-      if: matrix.env['TYPE'] == 'Release'
+      if: matrix.env['TYPE'] == 'Release' && github.ref == 'refs/heads/master' && github.repository == 'iovisor/bpftrace'
       env:
         QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
       run: ./docker/scripts/auth.sh ${{ github.repository }}
 
     - name: Package docker image and push to quay.io
-      if: matrix.env['TYPE'] == 'Release'
+      if: matrix.env['TYPE'] == 'Release' && github.ref == 'refs/heads/master' && github.repository == 'iovisor/bpftrace'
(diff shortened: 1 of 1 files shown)


dalehamel:
@danobi @mmisono why was this added? What is wrong with pushing other branche builds? This allows for easily testing out other builds.

Is this to limit images pushed to quay / did we hit some issue?

dalehamel:
the push.sh script is intended to be general, so that you can push your own repo with your own github actions, so this also breaks that functionality.

fbs:
IIRC its to prevents forks from pushing the image . And to avoid having branches other than master end up in quay, we have a few of those

dalehamel:
> IIRC its to prevents forks from pushing the image

This isn't possible, the organization is part of the github.repo:

https://github.com/iovisor/bpftrace/blob/master/docker/scripts/push.sh#L13

Which is passed in from the "github" context, not the user:

https://github.com/iovisor/bpftrace/blob/master/.github/workflows/embedded.yml#L246

So the user can't push to the iovisor repo because their action builds run within their own repo. eg, I can push to "quay.io/dalehamel/bpftrace", but not "quay.io/iovisor/bpftrace". This is pretty fundamental to github actions security model i think. I think filtering on the org is not necessary, and breaks this functionality:

https://github.com/iovisor/bpftrace/blob/master/.github/workflows/embedded.yml#L243

> And to avoid having branches other than master end up in quay, we have a few of those

I can see the argument for not wanting to push other branches, but what is the harm in it? As I said above, it is good for testing purposes. If someone uses a branch tag, they clearly can't expect it to be the same quality of a release tag or :latest, but I don't see any harm in having it available.

If we do want to prevent this, perhaps then we can remove the org/repo filter in #1525 as well. Otherwise, to just keep pushing everything, we can revert #1292

dalehamel:
Actually now that I think of it, i think I realize why this was probably added: builds would *fail* on forks if they don't have quay set up. I think that's a separate issue we should fix, so #1525 is the way forward, keeping the org/repo filter.

mmisono:
@dalehamel 
> builds would fail on forks if they don't have quay set up

Yes, that is the reason for this PR. Previously github action failed on forked repositories (e.g., https://github.com/mmisono/bpftrace/actions/runs/82847409). I've forgotten about the branch on iovisor/bpftrace. #1525 looks good.

PR #1294, opened 2020-04-21
Printer print ast after semantic analysis

fbs:
Extracting all the relevant type information while debugging can be
quite tedious. Walking the AST manually in GDB is a pain and putting
break points in the right places to be able to inspect it isn't much
easier either.

This extends the AST printer to include the type information added by
the semantic analyser:

```
AST after semantic analysis
-------------------
struct A { int a };

Program
 kprobe:f
  =
   variable: $a :: type[struct A*]
   (struct A*)
    int: 0 :: type[int64]
  =
   map: @ :: type[int64]
   .
    dereference
     variable: $a :: type[struct A*]
    a
  =
   map: @a :: type[string[16]]
   builtin: comm :: type[string[16]]
```

### DIFF
--- a/src/ast/printer.cpp
+++ b/src/ast/printer.cpp
@@ -1,6 +1,7 @@
 #include <cctype>
 #include <iomanip>
 #include <regex>
+#include <sstream>
 
 #include "printer.h"
 #include "ast.h"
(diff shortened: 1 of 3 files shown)


fbs:
While trying to get #1141 implemented I was looking for a an easy way to get type / addrspace info without having to mess with gdb. I think this turned out quite useful, useful enough to make enable it by default. 

Extended with as info:

```
Program
 tracepoint:syscalls:sys_enter_openat
  pred
   == :: type[unsigned int64 AS(kernel)]
    call: str :: type[string[64]]
     .
      dereference
       builtin: args :: type[(ctx) struct _tracepoint_syscalls_sys_enter_openat*]
      filename
    builtin: comm :: type[string[16] AS(kernel)]
  =
   map: @ :: type[int64 AS(kernel)]
   int: 1 :: type[int64 AS(kernel)]
```

olsajiri:
yep, seems useful, I dont see any issues

PR #1300, opened 2020-04-22
Use direct increase instead of map update for count function

olsajiri:
So the code generated for count follows this logic:

```
  val = map_lookup(key)
  if (!val)
    val = 0;
  val += 1
  map_update(key, val)
```
The idea of the change is to skip map_update call and do the 
increase directly in the memory, like:

```
  ptr = map_lookup(key)
  if (!ptr)
    map_update(key, 1)
  else
   *ptr += 1
```

I couldn't find interface to add `inc ptr` or `add ptr` op, 
I only found locked locked increase, which is not necessary for this case, but works ;-)

Any idea if there's interface for non locked add directly in memory?

The newly generated code looks like:
```
0: (b7) r1 = 0 
1: (7b) *(u64 *)(r10 -16) = r1
2: (18) r1 = 0xffff8898052cc800
4: (bf) r2 = r10 
5: (07) r2 += -16 
6: (85) call bpf_map_lookup_elem#1
7: (15) if r0 == 0x0 goto pc+3
```
new in memory increase:
```
8: (b7) r1 = 1 
9: (db) lock *(u64 *)(r0 +0) += r1
10: (05) goto pc+10
```
slow path for first item:
```
11: (b7) r1 = 1 
12: (7b) *(u64 *)(r10 -8) = r1
13: (18) r1 = 0xffff8898052cc800
15: (bf) r2 = r10 
16: (07) r2 += -16 
17: (bf) r3 = r10 
18: (07) r3 += -8
19: (b7) r4 = 0 
20: (85) call bpf_map_update_elem#2
21: (b7) r0 = 0 
22: (95) exit
```

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -270,14 +270,7 @@ void CodegenLLVM::visit(Call &call)
   {
     Map &map = *call.map;
     AllocaInst *key = getMapKey(map);
-    Value *oldval = b_.CreateMapLookupElem(map, key);
-    AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
-    b_.CreateStore(b_.CreateAdd(oldval, b_.getInt64(1)), newval);
-    b_.CreateMapUpdateElem(map, key, newval);
-
-    // oldval can only be an integer so won't be in memory and doesn't need lifetime end
-    b_.CreateLifetimeEnd(key);
-    b_.CreateLifetimeEnd(newval);
+    b_.CreateMapCount(map, key);
     expr_ = nullptr;
   }
   else if (call.func == "sum")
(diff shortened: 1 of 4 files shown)


fbs:
Sounds good. Looks like this need a somewhat modern kernel to work?

olsajiri:
> Sounds good. Looks like this need a somewhat modern kernel to work?

hm, I think the value was always writeable directly, or for long time now,
but I will check on that

Any idea if there's interface for non locked add directly in memory?

to do:
`9: (db)  *(u64 *)(r0 +0) += r1
`
instead of:
`9: (db) lock *(u64 *)(r0 +0) += r1`

because lok is not needed at this point, but I couldn't find the interface to do that,
just for locked increase.

fbs:
>  `9: (db) *(u64 *)(r0 +0) += r1`

Does that instruction exist? I don't see an xadd without lock

olsajiri:
> > `9: (db) *(u64 *)(r0 +0) += r1`
> 
> Does that instruction exist? I don't see an xadd without lock

ea, I can't see any and I thought it was strange, so wasn't sure ;-) thanks

PR #1311, opened 2020-04-28
Cache symbol resolvers by PID

dimakuz:
Symbol resolver cache is currently using executable paths as key, this leads to bad results when enabled with ASLR and bad performance when disabled.
This patch changes the cache a bit to store resolvers by PID.

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1762,15 +1762,21 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
     if (cache_user_symbols_)
     {
       std::string pid_exe = get_pid_exe(pid);
-      if (exe_sym_.find(pid_exe) == exe_sym_.end())
-      {
+      auto itr = exe_sym_.find(pid_exe);
+
+      if (itr != exe_sym_.end() && itr->second.first == pid) {
+        // FIXME check if that's the same process, e.g. ctime on /proc/pid/exe link
+        psyms = exe_sym_[pid_exe].second;
+      }
+
+      if (!psyms) {
+        if (itr != exe_sym_.end()) {
+          bcc_free_symcache(itr->second.second, itr->second.first);
+          exe_sym_.erase(itr);
+        }
         // not cached, create new ProcSyms cache
         psyms = bcc_symcache_new(pid, &symopts);
         exe_sym_[pid_exe] = std::make_pair(pid, psyms);
-      }
-      else
-      {
-        psyms = exe_sym_[pid_exe].second;
       }
     }
     else
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -91,10 +91,10 @@ int format(char * s, size_t n, const char * fmt, std::vector<std::unique_ptr<IPr
 
 BPFtrace::~BPFtrace()
 {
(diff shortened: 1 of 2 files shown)


mmisono:
This increases memory usage If tracee processes are short-lived. So I think it would be better to limit a total number of pid to cache. Also, can we make `chache_user_symbols_=1` by default? (then documentation (docs/reference_guid.md and help message in main.cpp) need to be updated.)

https://github.com/iovisor/bpftrace/blob/e7e87336d7606128bd075f39280fb184cfa59da3/src/main.cpp#L527-L530

fbs:
Thanks for looking into this :)

This looks like it's basically a revert of 63ff8b0d12e9a1a3f903834ee40b13c9086e20e4, which as was done to reduce memory usage when tracking a lot of forks.

> this leads to bad results when enabled with ASLR 

Even with this patch we can still bump into ASLR issues. If the pid exits before it has been cached results will still be off.

I'm not sure if there is one caching solution that will work, maybe we should look at multiple strategies depending on the situation (let the user control).

> and bad performance when [ASLR is] disabled.

How so?

dimakuz:
@fbs

> Thanks for looking into this :)
> 
> This looks like it's basically a revert of [63ff8b0](https://github.com/iovisor/bpftrace/commit/63ff8b0d12e9a1a3f903834ee40b13c9086e20e4), which as was done to reduce memory usage when tracking a lot of forks.

Oh, I wasn't aware it was this way before.

> 
> > this leads to bad results when enabled with ASLR
> 
> Even with this patch we can still bump into ASLR issues. If the pid exits before it has been cached results will still be off.

Yes I agree, but I'm not sure there's much to be done other that pro actively resolve each time e.g. ustack is referenced.

> 
> I'm not sure if there is one caching solution that will work, maybe we should look at multiple strategies depending on the situation (let the user control).
> 
> > and bad performance when [ASLR is] disabled.
> 
> How so?

I meant the bad performance when caching is disabled, due to constant module loading by bcc.

@mmisono 

> This increases memory usage If tracee processes are short-lived. So I think it would be better to limit a total number of pid to cache. Also, can we make `chache_user_symbols_=1` by default? (then documentation (docs/reference_guid.md and help message in main.cpp) need to be updated.)
> 
> https://github.com/iovisor/bpftrace/blob/e7e87336d7606128bd075f39280fb184cfa59da3/src/main.cpp#L527-L530

If this change is desired, some policy can be implemented, e.g. cache up to BPFTRACE_CACHE_SIZE resolvers and evict the ones used last. While each process can take a different amount of memory, this can still impose some ballpark limit.

danobi:
> If this change is desired, some policy can be implemented, e.g. cache up to BPFTRACE_CACHE_SIZE resolvers and evict the ones used last. While each process can take a different amount of memory, this can still impose some ballpark limit.

Yeah that sounds pretty reasonable to me.

It's been a while since I thought about this issue but IIRC I remember thinking it was possible to store base addr of each pid and then normalize each addr from pids sharing running the same exe before symbolizing. That way, we could even do offline symbolization if the pid exits and we haven't symbolized yet (by normalizing the addr and then symbolizing directly from exe on disk). This has been on my todo list to explore for a long time now but unfortunately I'm still a bit short on time to try it out.

PR #1313, opened 2020-04-29
Allow overriding whether usym module is shown

dimakuz:
I found it useful to on several occasions to display explicitly where the specific symbol is coming from. This patch allows overriding whether usym module is shown by setting an env variable.


danobi:
Hmm I wonder if we should make this an optional argument for `usym()` instead. This doesn't seem like something that _has_ to be applied script-wide.

Any thoughts @fbs ?

dimakuz:
@danobi I am actually using while printing ustack.

danobi:
Can't you use `ustack(perf)`, then?

dimakuz:
@danobi, that somehow that escaped my attention when I read the ref guide.
Closing.

PR #1317, opened 2020-05-05
Add --usdt-file-activation to activate USDT semaphores by filename

danobi:
This PR implements USDT semaphore probe activation. Previously, you had
to pass in `-p $PID` or `-c $CMD` to activate usdt semaphores on a process.
This has the limitation of single binaries. It's also difficult to make work in a
distributed tracing environment where you don't know the PIDs beforehand are
attaching to running processes.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1,9 +1,11 @@
 #include <arpa/inet.h>
 #include <cassert>
+#include <cstdio>
 #include <cstring>
 #include <ctime>
 #include <cxxabi.h>
 #include <fstream>
+#include <glob.h>
 #include <iomanip>
 #include <iostream>
 #include <sstream>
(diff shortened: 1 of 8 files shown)


danobi:
cc @dalehamel

danobi:
Hmm clang format is failing but I don't think I touched those files. Gonna ignore it and hope it goes away. Otherwise I'll put up another PR later to format over it.

dalehamel:
Thanks for this will take a look soon @danobi

src/main.cpp
```diff
@@ -192,6 +195,7 @@ int main(int argc, char *argv[])
   option long_options[] = {
     option{ "help", no_argument, nullptr, 'h' },
     option{ "version", no_argument, nullptr, 'V' },
+    option{ "usdt-file-activation", no_argument, nullptr, '$' },
```
fbs, line 198: `$` seems like a weird flag, or it is a special thing in getopt long?

src/bpftrace.cpp
```diff
@@ -724,8 +726,86 @@ void perf_event_lost(void *cb_cookie, uint64_t lost)
...
+      return ret;
+    }
+    std::string resolved(p);
+
+    for (size_t i = 0; i < globbuf.gl_pathc; ++i)
+    {
+      std::string path(globbuf.gl_pathv[i]);
+      std::ifstream file(path);
+      if (file.fail())
+        throw std::runtime_error("failed to open " + path);
```
fbs, line 760: Won't this cause an issue if a pid exits between the glob and now?

src/bpftrace.cpp
```diff
@@ -724,8 +726,86 @@ void perf_event_lost(void *cb_cookie, uint64_t lost)
...
+    int pid,
+    bool file_activation)
+{
+  std::vector<std::unique_ptr<AttachedProbe>> ret;
+
+  // File activation works by scanning through /proc/*/maps and seeing
+  // which processes have the target executable in their address space
+  // with execute permission. If found, we will try to attach to each
+  // process we find.
+  if (file_activation && probe.path.size())
```
fbs, line 741:
```suggestion
  if (!(file_activation && probe.path.size()))
  {
       ret.emplace_back(std::make_unique<AttachedProbe>(probe, func, pid));
       return ret;
   }
```

Avoiding the big if statement might improve readability a bit, nit.

src/bpftrace.cpp
```diff
@@ -724,8 +726,86 @@ void perf_event_lost(void *cb_cookie, uint64_t lost)
...
+        {
+          pid_parsed = std::stoi(pid_str);
+        }
+        catch (const std::exception &ex)
+        {
+          throw std::runtime_error("failed to parse pid=" + pid_str);
+        }
+
+        ret.emplace_back(
+            std::make_unique<AttachedProbe>(probe, func, pid_parsed));
```
fbs, line 789:
```suggestion
            std::make_unique<AttachedProbe>(probe, func, pid_parsed));
            break;
```

Don't think we have to process the rest of the file now

src/bpftrace.cpp
```diff
@@ -724,8 +726,86 @@ void perf_event_lost(void *cb_cookie, uint64_t lost)
...
+          continue;
+
+        auto parts = split_string(line, ' ');
+        if (parts.at(1).find('x') == std::string::npos)
+          continue;
+
+        // Remove `/proc/` prefix
+        std::string pid_str(globbuf.gl_pathv[i] + 6);
+        // Remove `/maps` suffix
+        for (int i = 0; i < 5; ++i)
```
fbs, line 775: I think you can skip this as stoi ignores anything after the number

src/main.cpp
danobi: It's only used for the `switch/case`. No short option for this flag.

src/bpftrace.cpp
danobi: Hmm good point, guess we have to fail silently here

danobi:
I added another runtime tests for multi-process too

danobi:
Any objections?

src/bpftrace.cpp
```diff
@@ -724,8 +726,89 @@ void perf_event_lost(void *cb_cookie, uint64_t lost)
...
+    ret.emplace_back(std::make_unique<AttachedProbe>(probe, func, pid));
+    return ret;
+  }
+
+  // File activation works by scanning through /proc/*/maps and seeing
+  // which processes have the target executable in their address space
+  // with execute permission. If found, we will try to attach to each
+  // process we find.
+  glob_t globbuf;
+  if (::glob("/proc/[0-9]*/maps", GLOB_NOSORT, nullptr, &globbuf))
```
dalehamel, line 748: I wonder how this will work for containers?

If we have different process namespaces, they will have their own view of /proc, so there are probably edge cases with a different process / mount namespace that we need to think through.

Provided bpftrace is in the same process / mount namespace you want to target though, this should work. If bpftrace runs from the root pid namespace i think this will be fine, but we may want to document this quick

dalehamel:
makes sense, i added a note about mount namespaces but that can probably be solved later (and there are already lots of issues with USDT and separate mount/process namespaces, so people are probably used to quirks here already)

src/bpftrace.cpp
danobi: So if bpftrace runs in root host I believe all processes running in containers can be activated. If bpftrace runs inside a pid/mount namespace then only the processes visible inside that container will be activated. This seems like fairly desirable and expected behavior.

Let me confirm that this is the behavior.

danobi:
I added another runtime test for running inside pid + mount namespaces. I also added a note in the documentation about how doing funny things w/ private mount namespaces and bind mounts can bread usdt file activation. Lots of quirks indeed.

danobi:
Removed the container test. Worked on all my machines but fails in CI. No idea.

dalehamel:
There are actions available to allow ssh in to GitHub actions, maybe that can provide some insight into the failure?

It might also be because actions runs in docker maybe? Can’t remember if it runs with host pid, I would have assumed so. (Checking from phone sorry)

danobi:
I'll put up another PR with the additional test and debug there

PR #1318, opened 2020-05-06
btf: Do not use <linux/types.h> for resolving tracepoint defs

danobi:
Passing --btf should mean bpftrace does not use system headers at all.
The only exception should be if the user manually does an `#include`.

To that end, this patch has TracepointFormatParser use the existing
BTF infrastructure to resolve basic types we used to get from
<linux/types.h> instead of doing header inclusions.

This is particularly important in container environments where kernel
headers may not be installed.

##### Checklist


olsajiri:
makes sense and looks good

PR #1320, opened 2020-05-07
types: Fix abort in semantic analysis

danobi:
Commit f68ef85005c ("types: Add basic type wrappers") added an assert
for integer sizes. However, it's possible during the first pass of
semantic analysis that integer size is unknown. On later passes the
correct size is determined.

The easy fix is to allow 0 sized integers and rely on the fact that
another error check later will catch the invalid integer if a zero sized
integer actually manages to get through semantic analysis.

##### Checklist


fbs:
woops, good to have a test for this one :)

PR #1325, opened 2020-05-11
bpftrace: Fix usdt reads in various architecture

sumanthkorikkar:
In USDT tracing, bpftrace uses bcc library function
bcc_usdt_get_argument to retrieve the arg passed to the DTRACE_PROBE().
The value present in the arg string is "gprs[0-15]" in s390. However it
is named as "r[0-15]" in bpftrace. This leads to offset for register
"gprs[0-15]" not known error.

Fix this by creating an alias of each register for s390x/ppc64. aarch64 already does it.
x86 does have the same convention in bpftrace and bcc.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to
   - [#1241](https://github.com/iovisor/bpftrace/pull/1241)
 - Type check the `cond` of if and ternary statements
   - [#1229](https://github.com/iovisor/bpftrace/pull/1229)
+- Fix usdt reads in various architecture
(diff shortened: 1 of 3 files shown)


sumanthkorikkar:
Hi All,

Note: Address space  separation patch (probe read split) still needed on s390 for uprobes/usdt.  However, this should fix the failure, once the probe read split is integrated to master. 

Thank you.

danobi:
Can you add an entry in the changelog?

Other than that, LGTM.

sumanthkorikkar:
> Can you add an entry in the changelog?
> 
Added
> Other than that, LGTM.
Let me check. Thank you

PR #1326, opened 2020-05-11
Add tuples to language

danobi:
This PR adds tuple support to the bpftrace language.

Tuples are useful for a couple of reasons:
* There is already a way to associate multiple values to a map key
(`@[1,2] = 3`) which is useful. Likewise, it would be very useful to
associate multiple values with a map value (ie `@[1] = (2,3)`).
Scripts currently work around this by using multiple maps. That's
fine but it'd be nicer to have the language support common use
cases.
* Tuples provide a way to structure data together in output.
Especially for distributed tracing, one common feature request is
to have multiple pieces of data appear as one "row" in wherever
the data is collected. You could work around this by using `printf`
and requiring a certain format (like CSV), but that's not very
reliable and can be easily broken by typos.

This PR allows stuff like the following:
```
$t = (1, 2, "string");
printf("%d %d %s\n", $t.0, $t.1, $t.2);

@m = ("str", (1, 2));
printf("%s %d %d\n", @m.0, @m.1.0, @m.1.1);
print(@m);
```

I'd rather get this proof of concept out to get feedback first, so
still on the todo list:
- [x] `print()`ing tuples
  * so stuff like `@ = (1,2); print(@);` comes out correctly
- [ ] modifying a tuple after creation (probably for another PR)
  * is this even desired? for example python tuples are immutable
  * would look something like: `$t = (0,1); $t.0 = 5;`

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -26,7 +26,7 @@ namespace {
 bool shouldBeOnStackAlready(SizedType &type)
 {
   return type.IsStringTy() || type.IsBufferTy() || type.IsInetTy() ||
-         type.IsUsymTy();
+         type.IsUsymTy() || type.IsTupleTy();
 }
 } // namespace
 
@@ -1311,12 +1311,25 @@ void CodegenLLVM::visit(Ternary &ternary)
 void CodegenLLVM::visit(FieldAccess &acc)
 {
   SizedType &type = acc.expr->type;
-  assert(type.IsCastTy() || type.IsCtxTy());
+  assert(type.IsCastTy() || type.IsCtxTy() || type.IsTupleTy());
   acc.expr->accept(*this);
 
   if (type.is_kfarg)
   {
     expr_ = b_.CreatKFuncArg(ctx_, acc.type, acc.field);
+    return;
+  }
+  else if (type.IsTupleTy())
+  {
+    Value *src = b_.CreateGEP(expr_,
+                              { b_.getInt32(0), b_.getInt32(acc.index) });
+    SizedType &elem_type = type.tuple_elems[acc.index];
+
+    if (shouldBeOnStackAlready(elem_type))
+      expr_ = src;
+    else
+      expr_ = b_.CreateLoad(b_.GetType(elem_type), src);
+
     return;
   }
 
@@ -1477,9 +1490,31 @@ void CodegenLLVM::visit(Cast &cast)
   }
 }
 
-void CodegenLLVM::visit(Tuple &tuple __attribute__((unused)))
+void CodegenLLVM::visit(Tuple &tuple)
 {
-  // XXX implement
+  // Store elements on stack
+  llvm::Type *tuple_ty = b_.GetType(tuple.type);
+  AllocaInst *buf = b_.CreateAllocaBPF(tuple_ty, "tuple");
+  for (size_t i = 0; i < tuple.elems->size(); ++i)
+  {
+    Expression *elem = tuple.elems->at(i);
+    elem->accept(*this);
+
+    Value *dst = b_.CreateGEP(buf, { b_.getInt32(0), b_.getInt32(i) });
+
+    if (shouldBeOnStackAlready(elem->type))
+    {
+      b_.CREATE_MEMCPY(dst, expr_, elem->type.size, 1);
+      if (!elem->is_variable && dyn_cast<AllocaInst>(expr_))
+        b_.CreateLifetimeEnd(expr_);
+    }
+    else
+    {
+      b_.CreateStore(expr_, dst);
+    }
+  }
+
+  expr_ = buf;
 }
 
 void CodegenLLVM::visit(ExprStatement &expr)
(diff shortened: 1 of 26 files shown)


src/parser.yy
```diff
@@ -314,10 +314,17 @@ expr : int                                      { $$ = $1; }
      | MINUS expr                               { $$ = new ast::Unop(token::MINUS, $2, @1); }
      | MUL  expr %prec DEREF                    { $$ = new ast::Unop(token::MUL,  $2, @1); }
      | expr DOT ident                           { $$ = new ast::FieldAccess($1, $3, @2); }
+     | expr DOT INT                             { $$ = new ast::FieldAccess($1, $3, @3); }
```
fbs, line 317: For the parser  I think  `map_or_var DOT INT` makes more sense here.  The downside is the error it generates

```
stdin:1:9-16: ERROR: syntax error, unexpected integer
i:s:1 { (1+2).2 }
        ~~~~~~~
```

Although I don't like the current one that much either:

```
stdin:1:9-16: ERROR: Can not access field '' on expression of type 'int64'
i:s:1 { (1+2).2 }
        ~~~~~~~
```

src/ast/codegen_llvm.cpp
```diff
@@ -1273,14 +1273,32 @@ void CodegenLLVM::visit(Ternary &ternary)
...
     return;
   }
+  else if (type.type == Type::tuple)
+  {
+    // Get offset to start of the indexed element
+    size_t offset = 0;
+    for (ssize_t i = 0; i < acc.index; ++i)
+      offset += type.tuple_elems[i].size;
+
+    Value *src = b_.CreateAdd(expr_, b_.getInt64(offset));
```
fbs, line 1292: I  wonder if llvm will be happier and codegen simpler if we emit  a struct for  each tuple. These  5 lines of codegen could then be replaced a single CreateGEP as llvm does the offset math.

src/output.cpp
```diff
@@ -305,6 +318,39 @@ void TextOutput::attached_probes(uint64_t num_probes) const
...
+
+  for (const SizedType &elemtype : ty.tuple_elems)
+  {
+    if (first)
+      first = false;
+    else
+      ret += ", ";
+
+    std::vector<uint8_t> elem_value(value.begin() + offset,
+                                    value.begin() + offset + ty.size);
```
mmisono, line 339:
```suggestion
                                    value.begin() + offset + elemtype.size);
```
AddressSanitizer detects heap-buffer-overflow.

mmisono:
Can we have a more user-friendly error message?
```
% ./src/bpftrace -e 'kprobe:vfs_read { @bytes[comm] = (avg(arg2), sum(arg2)); }'
stdin:1:34-44: ERROR: avg() should be assigned to a map
kprobe:vfs_read { @bytes[comm] = (avg(arg2), sum(arg2)); }
                                 ~~~~~~~~~~
stdin:1:46-55: ERROR: sum() should be assigned to a map
kprobe:vfs_read { @bytes[comm] = (avg(arg2), sum(arg2)); }
                                             ~~~~~~~~~
```

src/parser.yy
danobi: I think this would bread nested tuple indexing:
```
[  FAILED  ] basic.basic tuple
        Command: ../src/bpftrace -e 'BEGIN { $v = 99; $t = (0, 1, "str", (5, 6), $v); printf("%d %d %s %d %d %d\n", $t.0, $t.1, $t.2, $t.3.0, $t.3.1, $t.4); exit(); }'
        Expected: 0 1 str 5 6 99
        Found: stdin:1:98-104: ERROR: syntax error, unexpected integer
BEGIN { $v = 99; $t = (0, 1, "str", (5, 6), $v); printf("%d %d %s %d %d %d\n", $t.0, $t.1, $t.2, $t.3.0, $t.3.1, $t.4); exit(); }
```

src/ast/codegen_llvm.cpp
danobi: Ok, I'll take a look

src/output.cpp
danobi: Nice catch

src/parser.yy
danobi: I think I can make the error message better too

danobi:
> Can we have a more user-friendly error message?

I added the word "directly". Now it's:

```
# bpftrace -e 'kprobe:vfs_read { @bytes[comm] = (avg(arg2), sum(arg2)); }'
stdin:1:34-44: ERROR: avg() should be directly assigned to a map
kprobe:vfs_read { @bytes[comm] = (avg(arg2), sum(arg2)); }
                                 ~~~~~~~~~~
stdin:1:46-55: ERROR: sum() should be directly assigned to a map
kprobe:vfs_read { @bytes[comm] = (avg(arg2), sum(arg2)); }
```

I thought about adding a hint like "Is the call value being used in a tuple?" but I worry about making the common case more confusing. I feel like more people will do things like `$v = avg($asdf)` than `$v = (avg($asdf), "string")`.

src/ast/codegen_llvm.cpp
danobi: Done

PR #1327, opened 2020-05-11
tests: Remove flakey hardware_order runtime test

danobi:
This test is always flakey on my machine.

I'm not even sure the kernel guarantees two identically configured perf
events to be scheduled onto the same hardware PMU. IIUC modern x86 CPUs
have something like 4 PMUs. That means if our probe goes onto two
different PMUs we could race (which would explain the flakiness).

##### Checklist


danobi:
cc @olsajiri who probably knows more about PMUs

olsajiri:
yea, I don't think you can test the order via cache-events,
I think that could work ok with instructions

PR #1329, opened 2020-05-13
Add environment variable to tune perf ring buffer size

danobi:
Some scripts sample high frequency events and can overflow the ring
buffer. We should let the user be able to tune the ring buffer size as
it's a pretty classical throughput / space tradeoff.

##### Checklist


czardoz:
Would be nice to have this documented in the [reference guide](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#9-environment-variables) too!

javierhonduco:
LGTM!! Curious to see what happens if we pass a number that's not a power of 2, does it round it to the nearest value or would it fail?

danobi:
> LGTM!! Curious to see what happens if we pass a number that's not a power of 2, does it round it to the nearest value or would it fail?

I'll fail with a message like "unable to create perf buffer". I'll add a note about that in the docs

PR #1339, opened 2020-05-18
sizedtype accessors

fbs:
This adds and uses accessors for the `is_signed` and `type` fields of `sizedtype` as part of the type system improvement.

These make it easier to improve the type system, e.g. we could add `assert(IsIntTy()` to the `IsSigned` method to ensure that sign checking is only done on integers, without having to modify the code in 1000 places (like happens now)

This is largely an automatic refactor.

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cassert>
 #include <memory>
 #include <ostream>
 #include <sstream>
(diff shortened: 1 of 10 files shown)


fbs:
Link to #878

fbs:
@danobi  can you take a look at this one when you have time? It's size makes this one annoying

src/types.h
```diff
@@ -102,6 +100,14 @@ class SizedType
...
   bool operator==(const SizedType &t) const;
   bool operator!=(const SizedType &t) const;
+
+  bool IsSigned(void) const;
+
+  friend std::ostream &operator<<(std::ostream &, const SizedType &);
+  friend std::ostream &operator<<(std::ostream &, Type);
+
+private:
+  bool is_signed = false;
```
danobi: Style nit
```suggestion
  bool is_signed_ = false;
```

fbs: I'll sneak this in with the follow up :)

PR #1344, opened 2020-05-21
Reimplement `Type::array`

fbs:
This is the first in the series, so if we want to change function name style or something this is probably the best time.

Related to #878 

---

Arrays used the `size`, `elem_type` and `pointee_size` fields to store
array type information. This is confusing as `pointee_size` indicates a
pointer and also incomplete as an array of pointers cannot be stored
without losing pointee size information.

This fixes that by storing the complete type (`SizedType *`) instead of
just a few fields.

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -321,7 +321,7 @@ static SizedType get_sized_type(CXType clang_type)
     case CXType_ConstantArray:
     {
       auto elem_type = clang_getArrayElementType(clang_type);
-      auto size = clang_getArraySize(clang_type);
+      auto size = clang_getNumElements(clang_type);
       if (elem_type.kind == CXType_Char_S || elem_type.kind == CXType_Char_U)
       {
         return CreateString(size);
@@ -330,8 +330,8 @@ static SizedType get_sized_type(CXType clang_type)
       // Only support one-dimensional arrays for now
       if (elem_type.kind != CXType_ConstantArray)
       {
-        auto type = get_sized_type(elem_type);
-        return CreateArray(size, type.type, type.size, type.IsSigned());
+        auto elem_stype = get_sized_type(elem_type);
+        return CreateArray(size, elem_stype);
       } else {
         return CreateNone();
       }
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -24,12 +24,11 @@ std::ostream &operator<<(std::ostream &os, const SizedType &type)
   }
   else if (type.IsIntTy())
   {
-    os << (type.is_signed_ ? "" : "unsigned ") << "int" << 8*type.size;
+    os << (type.is_signed_ ? "" : "unsigned ") << "int" << 8 * type.size;
   }
   else if (type.IsArrayTy())
   {
-    os << (type.is_signed_ ? "" : "unsigned ") << "int" << 8 * type.pointee_size;
-    os << "[" << type.size << "]";
+    os << type.GetElementTy() << "[" << type.GetNumElements() << "]";
   }
   else if (type.IsStringTy() || type.IsBufferTy())
   {
@@ -257,14 +256,11 @@ SizedType CreateCTX(size_t size, std::string name)
   return SizedType(Type::ctx, size, name);
 }
 
-SizedType CreateArray(size_t num_elem,
-                      Type elem_type,
-                      size_t elem_size,
-                      bool elem_is_signed)
+SizedType CreateArray(size_t num_elements, const SizedType &element_type)
 {
-  auto ty = SizedType(Type::array, num_elem, elem_is_signed);
-  ty.elem_type = elem_type;
-  ty.pointee_size = elem_size;
+  auto ty = SizedType(Type::array, num_elements);
+  ty.num_elements_ = num_elements;
+  ty.element_type_ = new SizedType(element_type);
   return ty;
 }
 
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -24,11 +24,11 @@ std::ostream &operator<<(std::ostream &os, const SizedType &type)
   }
   else if (type.IsIntTy())
   {
-    os << (type.is_signed ? "" : "unsigned ") << "int" << 8*type.size;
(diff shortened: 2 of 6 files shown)


src/types.h
```diff
@@ -233,10 +264,7 @@ SizedType CreateUInt32();
 SizedType CreateUInt64();
 
 SizedType CreateString(size_t size);
-SizedType CreateArray(size_t num_elem,
-                      Type elem_type,
-                      size_t elem_size,
-                      bool elem_is_signed);
+SizedType CreateArray(size_t, SizedType);
```
danobi, line 267: Style nit:
```suggestion
SizedType CreateArray(size_t num_elements, SizedType element_type);
```
At a glance this looks like a single parameter type `size_t` with name `SizedType`. Also the API is less obvious if I just read the header

src/types.h
```diff
@@ -216,6 +243,10 @@ class SizedType
 
   friend std::ostream &operator<<(std::ostream &, const SizedType &);
   friend std::ostream &operator<<(std::ostream &, Type);
+
+  // Factories
+
+  friend SizedType CreateArray(size_t, SizedType);
```
danobi, line 249: Same as in decl

src/clang_parser.cpp
```diff
@@ -330,8 +330,7 @@ static SizedType get_sized_type(CXType clang_type)
       // Only support one-dimensional arrays for now
       if (elem_type.kind != CXType_ConstantArray)
       {
-        auto type = get_sized_type(elem_type);
-        return CreateArray(size, type.type, type.size, type.IsSigned());
+        return CreateArray(size, get_sized_type(elem_type));
```
danobi, line 333: Is `size` correct? Shouldn't it be `size / <elem_size>`?

src/types.cpp
```diff
@@ -257,14 +256,11 @@ SizedType CreateCTX(size_t size, std::string name)
   return SizedType(Type::ctx, size, name);
 }
 
-SizedType CreateArray(size_t num_elem,
-                      Type elem_type,
-                      size_t elem_size,
-                      bool elem_is_signed)
+SizedType CreateArray(size_t num_elements, SizedType element_type)
```
danobi, line 259:
```suggestion
SizedType CreateArray(size_t num_elements, const SizedType &element_type)
```

We can probably take by const ref

src/types.h
```diff
@@ -97,6 +98,9 @@ class SizedType
 
 private:
   bool is_signed_ = false;
+  std::shared_ptr<SizedType> element_type_; // for "container" and pointer
```
danobi, line 101: Don't think this needs to be a shared_ptr b/c I don't see SizedType sharing ownership of this. I think we can just store it directly.

fbs: I was looking for a good way to handle container types. In this case the copy constructor still works but once we introduce a pointer type it would have to recursively copy the whole chain. I was looking at ensuring that types are immutable and that we only "create" a type once (calling `CreateInt(8)` twice should return the same object).

Anyway all that isn't really relevant here so I'll take it out :)

src/clang_parser.cpp
fbs: I don't think so as size already was the number of elements, that bit hasn't changed.

Looking at the clang src itself this method is equal to `clang_getNumElements` for a constant array.
```
long long clang_getNumElements(CXType CT) {
  long long result = -1;
  QualType T = GetQualType(CT);
  const Type *TP = T.getTypePtrOrNull();

  if (TP) {
    switch (TP->getTypeClass()) {
    case Type::ConstantArray:
      result = cast<ConstantArrayType> (TP)->getSize().getSExtValue();
      break;
    case Type::Vector:
      result = cast<VectorType> (TP)->getNumElements();
      break;
    case Type::ExtVector:
      result = cast<ExtVectorType>(TP)->getNumElements();
      break;
    default:
      break;
    }
  }
  return result;
}

long long clang_getArraySize(CXType CT) {
  long long result = -1;
  QualType T = GetQualType(CT);
  const Type *TP = T.getTypePtrOrNull();

  if (TP) {
    switch (TP->getTypeClass()) {
    case Type::ConstantArray:
      result = cast<ConstantArrayType> (TP)->getSize().getSExtValue();
      break;
    default:
      break;
    }
  }
  return result;
}
```

I'll switch it to `clang_getNumElements` to make it less confusing

src/types.h
```diff
@@ -97,6 +98,9 @@ class SizedType
 
 private:
   bool is_signed_ = false;
+  SizedType *element_type_ = nullptr; // for "container" and pointer
```
danobi, line 101: Isn't this a memory leak?

Hmm, I'd still rather store by value. It looks like `SizedType` could just have some `= default`'d copy constructors and the compiler should be able to do the right thing.

fbs: > Hmm, I'd still rather store by value. It looks like SizedType could just have some = default'd copy constructors and the compiler should be able to do the right thing.

As in `SizedType elment_type_` or something else? 

> Isn't this a memory leak?

Yes it is. I could go (back) to shared_ptr as we allow copying this object. Or we could go to a unique_ptr (or normal pointer) and copy the object in the copy constructor and make sure it is deleted properly.

danobi: > As in SizedType elment_type_ or something else?

Yeah, like that.

> Yes it is. I could go (back) to shared_ptr as we allow copying this object. Or we could go to a unique_ptr (or normal pointer) and copy the object in the copy constructor and make sure it is deleted properly.

I'd prefer the method mentioned above. But I also don't wanna block this PR b/c how we store `element_type_` isn't really that important. Whatever seems reasonable to you is fine w/ me.

fbs: > Yeah, like that.

Hmm I don't think that will be possible as an object storing itself would make it endless nested.

> Whatever seems reasonable to you is fine w/ me.

I think this will work for now. Not entirely sure what the end result will be yet but it should be easy to refactor if needed :)

danobi: > Hmm I don't think that will be possible as an object storing itself would make it endless nested.

Ah right. Whoops

Issue #305, opened 2019-01-01
str() to support large strings


Birch-san:
(Discussion extracted from https://github.com/iovisor/bpftrace/pull/299)

This issue is to track "how could str() support kilobyte/megabyte/arbitrary string sizes?"

str() allocates strings on the BPF stack, so we reach the 512-byte limit very quickly (especially since we pay the toll again when printf() composes a perf event output buffer).

In fact, we're currently limited to tracing ~240-byte strings.

For an idea of "when would I want to trace a string larger than 240 bytes?":  
My favourite dtrace command was one that was frequently guilty of maxing out the default string buffer: tracing what SQL queries were submitted to MySQL. I'd usually grow the buffer to 1024 bytes for good measure.

=====

I explored the possibility of storing the strings off-stack. The main problem there is that probe_read_str() will only write upon on-stack memory. I have ideas for workarounds (and some code progress towards those workarounds), but they're too big to be included in the scope of this pull request.

=====

Here's some code that I built to explore the possibility of storing strings off-stack:

https://github.com/iovisor/bpftrace/compare/bd038f9110009f282c657c393fae60fdb7dd50ba...Birch-san:c5088c9a5d731165e1005a6a800e611bb63f3dc3

It introduces (in codegen) a "StrCall" specialization of "Call".  
StrCall has a `PERCPU_ARRAY` BPF map, to hold large data off-stack.

We cannot use probe_read_str() to output directly into the BPF map, but maybe we could allocate a small on-stack buffer, and ferry bytes (one buffer at a time) from BPF stack into BPF map.

Next, printf() poses a few problems:

- composes the perf event output struct on the BPF stack
- doesn't use runtime information to decide length of strings inside the perf event output struct
  - composes a fixed-size structure
  - determines size of structure during semantic analysis

I _think_ PerfEventOutput can transfer data from _BPF map_ to userspace (that's what "join" does), so we could compose the perf event output struct inside a BPF map.

====

We may find that we want a dynamically-sized perf event output structure. Like, imagine we see that args->count is just 3 bytes. We wouldn't want to send MAX_STRLEN bytes to userspace, if we know at runtime a smaller, more specific amount.

Here's some (horrifying) code I wrote to make print decide the size of its MEMCPY based on runtime information rather than semantic analysis:

https://github.com/iovisor/bpftrace/compare/master...Birch-san:6c1f19a3a4c3cf253aa23db79f5f637b6823e5b2

[...]

PR #750, opened 2019-06-11
RFC: allocate printf arguments in a map instead of the stack

mmarchini:
While working on #26 I hit the 512 stack limit, since printf arguments are stored in the stack and join will have a fixed size of `8 + 8 + (16 * 1024)`. To avoid bloating the stack, we can move printf arguments to a map (like we do for the list of strings that `join`  take).

`runtime-tests` passed (with the adjusts I made here) and tools parsing+codegen is working as well. Only a few codegen unit tests broke with this chages.

This is also one small step towards large string support.

PR #1346, opened 2020-05-25
RFC: Store strings off-stack

Birch-san:

We can now set BPFTRACE_STRLEN as high as we like. str() reads strings into a map, and gives printf() a descriptor of how to locate the string (map_fd, map_key, str_len, cpu_id).

=====

I suspect a data race is possible (similar to what @mmarchini theorised in https://github.com/iovisor/bpftrace/pull/750) since reading perf events is async.

Happy path:  
1. write string into map key 0
2. output perf event "you should read map key 0"
3. read perf event "you should read map key 0"

Data race:  
1. write string into map key 0
2. output perf event "you should read map key 0"
3. write string into map key 0
4. output perf event "you should read map key 0"
5. read perf event "you should read map key 0"
6. read perf event "you should read map key 0"

Proposed solution:  
1. write string into map key 0
2. output perf event "you should read map key 0"
3. write string into map key 1
4. output perf event "you should read map key 1"
5. read perf event "you should read map key 0"
6. indicate "map key 0 is now free to re-use"
7. read perf event "you should read map key 1"

In other words, a ring buffer. I haven't implemented this yet, but you'll see the idea referenced in the code so far.

We could determine "how likely is this data race to happen" by outputting a numeric sequence via frequent events, and check whether we encounter gaps in that sequence when printed.

Such a data race is already presumed to affect join(), but nobody's reported its happening in the wild.

=====

I think I would want each str() call to have an additional per-CPU array to encode 2 values: "start_used_map_ix" and "end_used_map_ix". This should be enough for the BPF program and userspace to coordinate "where's a free slot for me to write my strings".

This would solve the data race on str(). We'd need to do something similar for printf() too. And both would need error-handling to communicate when there's no free slots left.

=====

We could also avoid the data race on str() entirely (and be left with just the data race on printf() to worry about).

I chose to give printf "a descriptor of where to lookup the string", but it would've also been possible to put the string itself straight into printf's perf event. This would eliminate the need to coordinate userspace to read our strings out of maps.

You'd have to "always allocate space for max strlen" or "compose a struct of dynamic size". I don't know whether it's problematic for perf events to be big. I do know that it's hard to create structs of dynamic sizes.

Visitor pattern makes it hard for the str() call to probe read bytes directly into printf()'s buffer; the obvious implementation involves str() probe-reading into one map, and then printf() probe-reading the same bytes into its perf event buffer.

=====

Whilst thrashing around, I made a ring indexer in user-space, but I think the logic will need to move to BPF-space instead. Could be a useful reference though:

https://github.com/iovisor/bpftrace/blob/4dc04e781050036f08ab5493edfa4e4e3a457d9a/src/ring_indexer.h  
https://github.com/iovisor/bpftrace/blob/4dc04e781050036f08ab5493edfa4e4e3a457d9a/src/ring_indexer.cpp

=====

We'll also need to decide "should all strings be off-stack?" (e.g. string literals). In this prototype, printf() only supports printing strings produced by str().

I also expect that if we have anything like a "concatenate strings" operator, or a "compare string" function: those will now be incompatible with the output of str().


danobi:
Thanks for working on this, @Birch-san . I haven't looked too closely at the code yet -- just some high level thoughts.

> I think I would want each str() call to have an additional per-CPU array to encode 2 values: "start_used_map_ix" and "end_used_map_ix". This should be enough for the BPF program and userspace to coordinate "where's a free slot for me to write my strings".

> I chose to give printf "a descriptor of where to lookup the string", but it would've also been possible to put the string itself straight into printf's perf event. This would eliminate the need to coordinate userspace to read our strings out of maps.

Slot lookups / ring buffer is one approach but I worry about the complexity of coordination between userspace and bpf prog. Do you think it'd work if we had a per-cpu scratch space for storing results from `bpf_probe_read_str`? Then you send the data via perf event ring buffer. It's an extra copy but I suspect it'll be ok. The advantage of this approach is that we don't need to do additional head/tail coordination to find empty slots to write strings b/c (almost all[0]) bpf progs run in uninterruptible context.

> You'd have to "always allocate space for max strlen" or "compose a struct of dynamic size". I don't know whether it's problematic for perf events to be big. 

perf events are allowed to be pretty sizeable. I believe perf does last-branch-record based stack walks by copying sections of the running task's stack to userspace via the perf ring buffer. And if users have really huge strings they can use a05e9ef71cda0d to tune rb size.

> I do know that it's hard to create structs of dynamic sizes.

Not sure what you mean. Can you elaborate?

> We'll also need to decide "should all strings be off-stack?" (e.g. string literals). In this prototype, printf() only supports printing strings produced by str().

I think fully off stack would be a good idea.

> I also expect that if we have anything like a "concatenate strings" operator, or a "compare string" function: those will now be incompatible with the output of str().

Why would they be incompatible? Strings in a map can be referenced with a pointer just like strings on the stack.

---

[0]: There is one tracepoint that is inside an nmi handler but I think we can ignore it. I'm told there won't be more in the future.

Birch-san:
Thanks @danobi for reviewing. It's good to hear a clear evaluation of the design alternatives I didn't pursue.

> Do you think it'd work if we had a per-cpu scratch space for storing results from bpf_probe_read_str? Then you send the data via perf event ring buffer. It's an extra copy but I suspect it'll be ok.

Yup, should work. The simplest version of this is fairly easy to implement. If we're okay with the extra copy, and if we're okay reserving max string length in cases where the string length is dynamic.

The way to eliminate the extra copy and scratch buffer is: printf()'s codegen should — before visiting each str() argument — provide a pointer into the printf args buffer. it could assign this to a member variable on the visitor; str() could probe_read into that location.

Ah, but then it would become harder to compose the output of str() with other functions like string compare or string concatenate. Giving str() its own buffer means we have something we can assign to `expr_`, which maintains the interface we have with other callers.

> Not sure what you mean [regarding dynamically-sized structs]. Can you elaborate?

The perf event that we send to userspace is composed via `StructType::create(elements, …)`, where `elements` is a `std::vector<llvm::Type *> elements = { b_.getInt64Ty() };`.  
Each type specified here has to have a size known at C++ compile-time. I can do:

```c++
int MAX_STRLEN = strcall.maxStrSize.value();
elements.push_back(ArrayType::get(b_.getInt8Ty(), MAX_STRLEN))
```

But I cannot create a type from a length that's determined at runtime:

```c++
Value* dynamic_strlen = expr_;
elements.push_back(ArrayType::get(b_.getInt8Ty(), dynamic_strlen))
```

We could work around this by giving up on the struct, and just allocate some untyped length of memory (note: this would involve doing visiting each of our arguments to ask for their length, before we're ready to visit them again to ask for their content).
Or we could be "mostly a struct", but then wherever the struct ends we start treating as a dynamic-length string space:

```
# some 64-bit integers
00000010: [0000 0000 0000 0000] [0000 0000 0000 0000]
# some 16-bit and 32-bit integers (aligned to 64-bit words)
00000020: [0000]0000 0000 0000  [0000 0000]0000 0000
# some static-length char[]s (aligned to 64-bit words)
00000030: [CAFE BABE]0000 0000  [DEAD BEEF]0000 0000
# our struct ends; our dynamic-length string space begins
00000040:  FACE CAFE 00BA BA00   FAEC 1800
```

But if "perf events are allowed to be pretty sizeable", then it's probably overkill to try this hard to save bytes.

I guess a more reasonable way to save bytes is: allow every str() call to specify a MAX_STRLEN as an argument, rather than relying on one global MAX_STRLEN. So you could save your budget for just the str() calls you're expecting free-text in.

> [moving all strings] fully off stack would be a good idea  
> Why would [stack strings and map strings] be incompatible? Strings in a map can be referenced with a pointer just like strings on the stack.

Ah, I was tripping up over the distinction I created in this pull request: the new "map string" type meant that what we store in expr_ is a descriptor of how to find the string (in a different map), not the string itself.

But I guess if we return to "put string content into expr_", then we maintain the original interface.

The simpler design is starting to sound pretty appealing.

Birch-san:
Rebased; review continues in https://github.com/iovisor/bpftrace/pull/1357

PR #1347, opened 2020-05-25
Add lsm probe support

olsajiri:
Adding new 'lsm' attach point. The 'lsm' stands for linux security module and lsm probe allows to trace lsm kernel hooks.

```
  # bpftrace -e 'lsm:* { @[comm,probe] = kstack; }'
  Attaching 215 probes...
  @[vim, lsm:inode_getattr]:
        cleanup_module+1646461
        bpf_get_stackid_raw_tp+78
        cleanup_module+1646461
        cleanup_module+1636212
        bpf_lsm_inode_getattr+5
        security_inode_getattr+48
        vfs_getattr+29
        vfs_statx+150
        __do_sys_newstat+57
        do_syscall_64+91
        entry_SYSCALL_64_after_hwframe+68
```

It's also possible to change the return value of the hook if it's available:
```
  # bpftrace -e 'lsm:perf_event_open { return -1234; }'
  Attaching 1 probe...

  $ sudo perf record ls
  perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error 1234 (INTERNAL ERROR: strerror_r(1234, [buf], 128)=22)
  perf_event_open(..., 0) failed unexpectedly with error 1234 (INTERNAL ERROR: strerror_r(1234, [buf], 128)=22)
```

You need following bcc change to run this:
https://github.com/iovisor/bcc/pull/2934


docs/reference_guide.md
```diff
@@ -781,6 +783,8 @@ Loops can be short circuited by using the `continue` and `break` keywords.
 The `return` keyword is used to exit the current probe. This differs from
 `exit()` in that it doesn't exit bpftrace.
 
+By default zero value is returned, but user can specify expression to be returned.
```
fbs, line 786: I think it would be good to provide a bit more detail here as this might give users the impression that the return value has a meaning. Atm it is only relevant for the LSM and for the few kprobes that support it `override()` can be used

docs/reference_guide.md
```diff
@@ -1503,6 +1507,64 @@ fd 3 name libselinux.so.1
...
+
+**WARNING**: this feature is experimental and may be subject to interface changes.
+
+Syntax:
+
+```
+lsm:function
+```
+
+These are LSM (Lisnux Security Module) hooks probes implemented via eBPF trampolines.
```
danobi, line 1521:
```suggestion
These are LSM (Linux Security Module) hooks probes implemented via eBPF trampolines.
```

docs/reference_guide.md
```diff
@@ -1503,6 +1507,64 @@ fd 3 name libselinux.so.1
...
+```
+
+
+## 18. `lsm`: LSM Hooks Tracing Arguments
+
+Syntax:
+
+```
+kfunc:function      args->NAME  ...
+kretfunc:function   args->NAME ... retval
```
danobi, line 1555: Typo?
```suggestion
lsm:function      args->NAME  ...
lsm:function   args->NAME ... retval
```

docs/reference_guide.md
```diff
@@ -1503,6 +1507,64 @@ fd 3 name libselinux.so.1
...
+```
+
+Arguments can be accessed the same way as for kfunc probes (described in Chapter 16)
+
+It's possible to change the return value of the hook, by returning the new value, like:
+
+  # bpftrace -e 'lsm:perf_event_open { return -1234; }'
+
```
danobi, line 1563: Dupe of below?
```suggestion
```

docs/reference_guide.md
olsajiri: ok, will add

docs/reference_guide.md
olsajiri: will fix

docs/reference_guide.md
olsajiri: yep, will fix

docs/reference_guide.md
olsajiri: will check, thanks

mmisono:
Is `bpf_get_stackid` usable in lsm probe?

```
% sudo ./src/bpftrace -e 'l:socket_create { @=kstack; }' -v
Attaching 1 probe...

Error log:
0: (18) r2 = 0xffffb2edc25b1000
2: (b7) r6 = 0
3: (b7) r3 = 0
4: (85) call bpf_get_stackid#27
unknown func bpf_get_stackid#27

Error loading program: lsm:socket_create
```

It seems currently it is not allowed in the mainline kernel.

- https://github.com/torvalds/linux/blob/86852175b016f0c6873dcbc24b93d12b7b246612/kernel/bpf/bpf_lsm.c#L52
- https://github.com/torvalds/linux/blob/86852175b016f0c6873dcbc24b93d12b7b246612/kernel/trace/bpf_trace.c#L813

olsajiri:
> Is `bpf_get_stackid` usable in lsm probe?
> 
> ```
> % sudo ./src/bpftrace -e 'l:socket_create { @=kstack; }' -v
> BTF: using data from /sys/kernel/btf/vmlinux
> Attaching 1 probe...
> 
> Error log:
> 0: (18) r2 = 0xffffb2edc25b1000
> 2: (b7) r6 = 0
> 3: (b7) r3 = 0
> 4: (85) call bpf_get_stackid#27
> unknown func bpf_get_stackid#27
> processed 4 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
> 
> Error loading program: lsm:socket_create
> ```
> 
> It seems currently it is not allowed in the mainline kernel.
> 
>     * https://github.com/torvalds/linux/blob/86852175b016f0c6873dcbc24b93d12b7b246612/kernel/bpf/bpf_lsm.c#L52
> 
>     * https://github.com/torvalds/linux/blob/86852175b016f0c6873dcbc24b93d12b7b246612/kernel/trace/bpf_trace.c#L813

right I forgot I have patched kernel for that, sry.. I'll post the change
and put the link in here

olsajiri:
posted the kernel change to add missing lsm helpers:
[https://lore.kernel.org/bpf/20200531154255.896551-1-jolsa@kernel.org/T/#u](https://lore.kernel.org/bpf/20200531154255.896551-1-jolsa@kernel.org/T/#u)

src/ast/semantic_analyser.cpp
```diff
@@ -2121,23 +2124,27 @@ void SemanticAnalyser::visit(AttachPoint &ap)
       }
     }
   }
-  else if (ap.provider == "kfunc" || ap.provider == "kretfunc")
+  else if (ap.provider == "kfunc" || ap.provider == "kretfunc" ||
+           ap.provider == "lsm")
   {
 #ifndef HAVE_BCC_KFUNC
     error("kfunc/kretfunc not available for your linked against bcc version.",
```
mmisono, line 2131: I think "lsm not available ..." is better when lsm is not available since there is a case where kfunc is supported but lsm is not.

mmisono:
Something goes wrong when trying to attach non-existing kfunc probe with existing probe. For example

``` 
sudo ./src/bpftrace -e 'f:vfs_open,f:aaa { @[probe] = count(); }' -v
Attaching 2 probes...

Error log:

Error loading program: kfunc:aaa
```

------

Previously
```
% sudo ./src/bpftrace -e 'f:vfs_open,f:aaa { @[probe] = count(); }'
stdin:1:12-17: ERROR: Failed to resolve kfunc args.
f:vfs_open,f:aaa { @[probe] = count(); }
           ~~~~~
```

docs/reference_guide.md
```diff
@@ -1519,6 +1524,61 @@ fd 3 name libselinux.so.1
 ```
 And as you can see in above example it's also possible to access function arguments on `kretfunc` probes.
 
+
+## 17. `lsm`: LSM Hooks Tracing
```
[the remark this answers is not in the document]
olsajiri: good idea, will add

docs/reference_guide.md
```diff
@@ -782,6 +784,9 @@ Loops can be short circuited by using the `continue` and `break` keywords.
 The `return` keyword is used to exit the current probe. This differs from
 `exit()` in that it doesn't exit bpftrace.
 
+By default zero value is returned, but user can specify expression to be returned,
+which is currently supported (has effect) only in lsm probes.
+
```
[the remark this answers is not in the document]
olsajiri: ok

src/ast/semantic_analyser.cpp
olsajiri: yes, will change

olsajiri:
> Something goes wrong when trying to attach non-existing kfunc probe with existing probe. For example
> 
> ```
> sudo ./src/bpftrace -e 'f:vfs_open,f:aaa { @[probe] = count(); }' -v
> BTF: using data from /sys/kernel/btf/vmlinux
> Attaching 2 probes...
> 
> Error log:
> 
> Error loading program: kfunc:aaa
> ```
> 
> Previously
> 
> ```
> % sudo ./src/bpftrace -e 'f:vfs_open,f:aaa { @[probe] = count(); }'
> stdin:1:12-17: ERROR: Failed to resolve kfunc args.
> f:vfs_open,f:aaa { @[probe] = count(); }
>            ~~~~~
> ```

good catch, I'll add test for it, thanks

olsajiri:
> Isn't `prog_lsm_` needed to be defined in mocks.h?
> 
> ```diff
> diff --git a/tests/mocks.h b/tests/mocks.h
> index f550fdf3..eaec07a5 100644
> --- a/tests/mocks.h
> +++ b/tests/mocks.h
> @@ -67,6 +67,7 @@ public:
>      has_get_current_cgroup_id_ = std::make_optional<bool>(has_features);
>      has_override_return_ = std::make_optional<bool>(has_features);
>      prog_kfunc_ = std::make_optional<bool>(has_features);
> +    prog_lsm_ = std::make_optional<bool>(has_features);
>      has_loop_ = std::make_optional<bool>(has_features);
>    };
>  };
> ```
> 
> By the way, what will happen if multiple BPF program that attached to the same lsm probe return different values? Is only the last value valid?
> 
> edit: According to the documentation (https://www.kernel.org/doc/html/latest/bpf/bpf_lsm.html), "ret" holds the return value from the previous BPF program, but the following example seems only executing the first program. Is it expected?
> 
> ```
> % sudo ./src/bpftrace -e 'lsm:perf_event_open { return -1234; } lsm:perf_event_open { printf("%d\n", retval); return 0; }'
> ```
> 
> ```
> % sudo ./perf record ls
> perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error 1234 (INTERNAL ERROR: strerror_r(1234, [buf], 128)=22)
> ```

I thought it'd fail on first error value.. I'll check and get back to you

brendangregg:
I like it. It's obvious in this case, but something I'm thinking about is how bpftrace will look on BSD and Windows. Having some functions or probe types that look generic but have Linux-isms baked in by accident is something we'd want to avoid. In this case, it's pretty clear it's for Linux only (lsm), so other OSes simply wouldn't have this probe type.

Oh, and I've been calling them "probe types" everywhere, not attach points.

olsajiri:
> I like it. It's obvious in this case, but something I'm thinking about is how bpftrace will look on BSD and Windows. Having some functions or probe types that look generic but have Linux-isms baked in by accident is something we'd want to avoid. In this case, it's pretty clear it's for Linux only (lsm), so other OSes simply wouldn't have this probe type.

interesting.. so is this as a plan? bpftrace ports to other OSes?

> 
> Oh, and I've been calling them "probe types" everywhere, not attach points.

ok, I'll change that to probe types, thanks

olsajiri:
> By the way, what will happen if multiple BPF program that attached to the same lsm probe return different values? Is only the last value valid?

I think so, BPF program gets return value of the previous one
and the last BPF program return value is the one that overwrites
the hook's return value

> 
> edit: According to the documentation (https://www.kernel.org/doc/html/latest/bpf/bpf_lsm.html), "ret" holds the return value from the previous BPF program, but the following example seems only executing the first program. Is it expected?

the example below attaches just one program to  lsm:perf_event_open hook,
so it gets executed and changes the hook's return value 
 
> 
> ```
> % sudo ./src/bpftrace -e 'lsm:perf_event_open { return -1234; } lsm:perf_event_open { printf("%d\n", retval); return 0; }'
> ```
> 
> ```
> % sudo ./perf record ls
> perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error 1234 (INTERNAL ERROR: strerror_r(1234, [buf], 128)=22)
> ```

mmisono:
As for return values, I checked the code and found that the execution of BPF programs stopped if a BPF program returned other than zero.
 
https://github.com/torvalds/linux/commit/ae24082331d9bbaae283aafbe930a8f0eb85605a#diff-b2c1ad58525d1bb2fb7e87c1508e2accR1496-R1507

I think the current kernel documentation is somewhat misleading. Anyway, the behavior of 

> 
> ```
> % sudo ./src/bpftrace -e 'lsm:perf_event_open { return -1234; } lsm:perf_event_open { printf("%d\n", retval); return 0; }'
> ```
> ```
> % sudo ./perf record ls
> perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error 1234 (INTERNAL ERROR: strerror_r(1234, [buf], 128)=22)
> ```

seems correct.

src/ast/field_analyser.cpp
```diff
@@ -206,22 +229,99 @@ void FieldAnalyser::visit(Predicate &pred)
...
+  }
+
+  // check if we already stored arguments for this probe
+  auto it = bpftrace_.btf_ap_args_.find(probe_->name());
+
+  if (it != bpftrace_.btf_ap_args_.end())
+  {
+    // we did, and it's different.. save the state and
+    // triger the error if there's args->xxx detected
+    if (!compare_args(it->second, ap_args_))
```
mmisono, line 300: It seems this check is too restrictive. For example, `vfs_open(const struct path *path, struct file *file)` and `vfs_write(struct file *file, ...)` have the same arg `struct file *file`, but cannot create the program to access the arg.

```
% sudo ./src/bpftrace -e 'f:vfs_open, f:vfs_write { @ = args->file->f_count; }'
stdin:1:12-24: ERROR: Failed: probe has attach points with mixed arguments
f:vfs_open, f:vfs_write { @ = args->file->f_count; }
           ~~~~~~~~~~~~
```
This works on the current master.

mmisono:
I encountered the following error.

```
% sudo ./src/bpftrace -e 'lsm:path_chown { }'  -v
Attaching 1 probe...

Error log:
The function bpf_lsm_path_chown arg1 type STRUCT is unsupported.
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

Error loading program: lsm:path_chown
```
Run on Linux 5.7 with [a patch](https://lore.kernel.org/bpf/20200531154255.896551-1-jolsa@kernel.org/T/#u). Maybe this is a problem with the kernel.

olsajiri:
> path_chown

right, I see that too.. it's because of the kuid_t type being throught struct:

```
typedef struct {
        uid_t val;
} kuid_t;
```
and kernel verifier so far supports only pointer to struct as an argument,
I'll see if we can somehow add exception to cases like this and handle
it as scalar value

mmisono:
bcc introduced `bpf_attach_lsm()` in addition to `bpf_attach_kfunc`(in iovisor/bcc#2979), though there is no essential differences between them.

olsajiri:
> `bpf_attach_kfunc`

nice, the more the merrier ;)

willfindlay:
> > path_chown
> 
> right, I see that too.. it's because of the kuid_t type being throught struct:
> 
> ```
> typedef struct {
>         uid_t val;
> } kuid_t;
> ```
> 
> and kernel verifier so far supports only pointer to struct as an argument,
> I'll see if we can somehow add exception to cases like this and handle
> it as scalar value

Any updates on this?

danobi:
Also bumping for @mmisono latest review comments

olsajiri:
> > > path_chown
> > 
> > 
> > right, I see that too.. it's because of the kuid_t type being throught struct:
> > ```
> > typedef struct {
> >         uid_t val;
> > } kuid_t;
> > ```
> > 
> > 
> > and kernel verifier so far supports only pointer to struct as an argument,
> > I'll see if we can somehow add exception to cases like this and handle
> > it as scalar value
> 
> Any updates on this?

sry for delay.. there were some comments I need to address, but I think it'll go through eventually

however I decided to split this code and first submit the kfunc wildcard support.. which I'm about to do right now, then I'll figure out the kernel change and get back to lsm support

jirka

olsajiri:
> 
> however I decided to split this code and first submit the kfunc wildcard support.. which I'm about to do right now, ...

https://github.com/iovisor/bpftrace/pull/1410

fbs:
can you add a changelog item for this?

bitvijays:
@olsajiri Hope you are doing well. Thank you for working on this and bringing LSM support to bcc. I came across the PR https://github.com/iovisor/bcc/pull/2934 which provides some info to load LSM programs using `bpf_attach_kfunc` and `bpf_detach_kfunc` functions. Is there any documentation/ blogs which provide examples on loading simple LSM program via `bpf_attach_kfunc` Thank you for your support.

danobi:
Hi @bitvijays , you might have more luck asking in bcc repo or on the iovisor mailing list. Your question seems specific to bcc and not bpftrace, so this might be the wrong audience.

willfindlay:
@bitvijays I added automatic LSM probe loading to bcc in https://github.com/iovisor/bcc/pull/2979. All you need to do now is just use the `LSM_PROBE(hook, [arguments...])` syntax in your BPF program. The probe will be automatically loaded.

bitvijays:
@willfindlay Hope you are doing well. Thank you for your support. Tried re-using your example in issue https://github.com/iovisor/bcc/issues/2961 . Have a query https://github.com/iovisor/bcc/issues/2961#issuecomment-663927793 Would it be possible for you to have a look? Thanks

olsajiri:
> @bitvijays I added automatic LSM probe loading to bcc in [iovisor/bcc#2979](https://github.com/iovisor/bcc/pull/2979). All you need to do now is just use the `LSM_PROBE(hook, [arguments...])` syntax in your BPF program. The probe will be automatically loaded. Of course this is bcc only but it sounds like you are asking a bcc question, as danobi said.

@bitvijays sorry for late reply, IMO best is to follow @willfindlay advice,
I added  bpf_attach_kfunc/bpf_detach_kfunc to support bpftrace,
not to be used directly in bcc programs

willfindlay:
@olsajiri Any updates on this? Would be quite interested in having LSM support in newer builds of bpftrace.

danobi:
Closing due to age. This would be good to get in -- tracking this feature in https://github.com/iovisor/bpftrace/issues/2783

PR #1356, opened 2020-05-28
usdt: Fix attaching to probes in different binaries

danobi:
USDTHelper caching logic was previously incorrect. It would mark the
cache as loaded if any pid or path was processed. This would cause
scripts that place usdt probes on different binaries to silently fail
(because a default usdt_probe_entry would be returned).

See the runtime test for an example of how this can occur.

##### Checklist


danobi:
cc @dalehamel

dalehamel:
Sorry for the delay, i'll take a look shortly danobi

PR #1357, opened 2020-05-28
RFC: Store strings off-stack (rebased)

Birch-san:
**This is just a rebase/squash of https://github.com/iovisor/bpftrace/pull/1346, intended to make it easier to branch off and try suggestions from the review.**

We can now set BPFTRACE_STRLEN as high as we like. str() reads strings into a map, and gives printf() a descriptor of how to locate the string (map_fd, map_key, str_len, cpu_id).

=====

I suspect a data race is possible (similar to what @mmarchini theorised in https://github.com/iovisor/bpftrace/pull/750) since reading perf events is async.

Happy path:  
1. write string into map key 0
2. output perf event "you should read map key 0"
3. read perf event "you should read map key 0"

Data race:  
1. write string into map key 0
2. output perf event "you should read map key 0"
3. write string into map key 0
4. output perf event "you should read map key 0"
5. read perf event "you should read map key 0"
6. read perf event "you should read map key 0"

Proposed solution:  
1. write string into map key 0
2. output perf event "you should read map key 0"
3. write string into map key 1
4. output perf event "you should read map key 1"
5. read perf event "you should read map key 0"
6. indicate "map key 0 is now free to re-use"
7. read perf event "you should read map key 1"

In other words, a ring buffer. I haven't implemented this yet, but you'll see the idea referenced in the code so far.

We could determine "how likely is this data race to happen" by outputting a numeric sequence via frequent events, and check whether we encounter gaps in that sequence when printed.

Such a data race is already presumed to affect join(), but nobody's reported its happening in the wild.

=====

I think I would want each str() call to have an additional per-CPU array to encode 2 values: "start_used_map_ix" and "end_used_map_ix". This should be enough for the BPF program and userspace to coordinate "where's a free slot for me to write my strings".

This would solve the data race on str(). We'd need to do something similar for printf() too. And both would need error-handling to communicate when there's no free slots left.

=====

We could also avoid the data race on str() entirely (and be left with just the data race on printf() to worry about).

I chose to give printf "a descriptor of where to lookup the string", but it would've also been possible to put the string itself straight into printf's perf event. This would eliminate the need to coordinate userspace to read our strings out of maps.

You'd have to "always allocate space for max strlen" or "compose a struct of dynamic size". I don't know whether it's problematic for perf events to be big. I do know that it's hard to create structs of dynamic sizes.

Visitor pattern makes it hard for the str() call to probe read bytes directly into printf()'s buffer; the obvious implementation involves str() probe-reading into one map, and then printf() probe-reading the same bytes into its perf event buffer.

=====

Whilst thrashing around, I made a ring indexer in user-space, but I think the logic will need to move to BPF-space instead. Could be a useful reference though:

https://github.com/iovisor/bpftrace/blob/4dc04e781050036f08ab5493edfa4e4e3a457d9a/src/ring_indexer.h  
https://github.com/iovisor/bpftrace/blob/4dc04e781050036f08ab5493edfa4e4e3a457d9a/src/ring_indexer.cpp

=====

We'll also need to decide "should all strings be off-stack?" (e.g. string literals). In this prototype, printf() only supports printing strings produced by str().

I also expect that if we have anything like a "concatenate strings" operator, or a "compare string" function: those will now be incompatible with the output of str().

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -1,6 +1,7 @@
 #include "ast.h"
 #include "parser.tab.hh"
 #include <iostream>
+#include <cstdlib>
 
 namespace bpftrace {
 namespace ast {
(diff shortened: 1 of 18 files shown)


Birch-san:
Copying the feedback from @danobi here from previous MR: https://github.com/iovisor/bpftrace/pull/1346#issuecomment-634312784

image

fbs:
Being able to read in loooooooong strings would be great but I don't think we should change the current behaviour of `str()`, as a construct like `@[str(arg0)]=arg1` should continue to work. I'd rather see a new call for this async version of str.

Using the perf infrastructure would have my preference over trying to implement our own ringbuffer in bpf

Birch-san:
Closed in favour of https://github.com/iovisor/bpftrace/pull/1360, which iterates on this work in light of review feedback.

PR #1358, opened 2020-05-29
Remove RLIMIT_AS limit

### LINKED ISSUE #1355: bpftrace fails on arm64 with many CPUs
### bpftrace --info
System
  OS: Linux 4.18.0-206.el8.aarch64 #1 SMP Tue May 26 18:40:29 UTC 2020
  Arch: aarch64

Build
  version: v0.10.0-143-g1033
  LLVM: 10
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: no
  probe_read_user_str: no
  probe_read_kernel: no
  probe_read_kernel_str: no
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no

### What reproduces the bug?
Running any bpftrace tool on a arm64 machine with enough CPUs.

$ /usr/local/share/bpftrace/tools/biolatency.bt
Attaching 4 probes...
mmap: Cannot allocate memory
Failed to open perf buffer

What's happening is that we hit the 1 GB address space limit set in cap_memory_limits(). There are 2 factors at play here: first, there are many cpus on this machine (256) and second, the page size is 64k. So in setup_perf_events() we're trying to open 256 perf buffer of  64 pages (that is 4 MB per cpu).

The memory limits in cap_memory_limits() should probably be dependant on the memory available in the machine. Also, do we need address space limit at all? The RSS limit should be enough to prevent OOM and it's quite common that the address space is significantly bigger than RSS. Just removing the RLIMIT_AS line allows bpftrace to work on that machine and the RSS stays well below the 1 GB limit.

danobi:
We haven't seen any OOM issues in a while so I suspect either our type
fixes in bpftrace or upstream llvm changes have resolved the issue.

This closes #1355.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to
   - [#1328](https://github.com/iovisor/bpftrace/pull/1328)
 - Only list uprobe and usdt probes when `-p` is given
   - [#1340](https://github.com/iovisor/bpftrace/pull/1340)
+- Remove address space memory limit
+  - [#1358](https://github.com/iovisor/bpftrace/pull/1358)
 
 #### Deprecated
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -102,7 +102,6 @@ if (BUILD_ASAN)
     # target_link_options is supported in CMake 3.13 and newer
     message("Please use CMake 3.13 or newer to enable ASAN")
   endif()
-  target_compile_definitions(bpftrace PRIVATE BUILD_ASAN)
   target_compile_options(bpftrace PUBLIC "-fsanitize=address")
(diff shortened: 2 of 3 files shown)


javierhonduco:
LGTM!

(Maybe we could set up some alarms in our infra to see how OOMs / memory usage our bpftrace invocations get?)

Issue #1067, opened 2020-01-07
Duplicates in string keyed map


terrelln:
```
> bpftrace --version
bpftrace v0.9.3
> sudo bpftrace -e 'tracepoint:block:block_rq_issue { @bytes[args->rwbs] = sum(args->bytes); }'
Attaching 1 probe...
^C

@bytes[WS]: 2
@bytes[R]: 4
@bytes[R]: 5
@bytes[RA]: 6
@bytes[RA]: 9
@bytes[R]: 16
@bytes[RA]: 30
@bytes[R]: 32
```

There are duplicates in the string keyed map. It seems like maybe the end of the string isn't being zeroed, and then the whole buffer gets strcmped, but that is just a guess. I'm running a patched version of Linux 4.16.

mmisono:
I guess your guess is right. Maybe the following (terrible) workaround will work.

```
#include <linux/types.h>

struct arg_1 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[1]; }
struct arg_2 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[2]; }
struct arg_3 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[3]; }
struct arg_4 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[4]; }
struct arg_5 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[5]; }
struct arg_6 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[6]; }
struct arg_7 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[7]; }
struct arg_8 { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes; char rwbs[8]; }
struct arg   { long pad; dev_t dev; sector_t sector; unsigned int nr_sector; unsigned int bytes;  int rwbs[8]; }

tracepoint:block:block_rq_issue
{
  $i = 0;
  $a = (struct arg*)args;
  if($a->rwbs[0] != 0) { $i = 1; }
  if($i == 1 && $a->rwbs[1] != 0) { $i = 2; }
[...]

Issue #342, opened 2019-01-14
RFC: alternatives to static build


mmarchini:
During the BPF Microconference at LPC, we got feedback that installing llvm/clang/libbpfcc in production servers just to use `bpftrace` would be overkill and sometimes unfeasible. An alternative would be linking to those dependencies statically, but most distributions don't provide the static library `libclang.a` on their libclang packages.

I started to think about alternatives to a static build, and two ideas came up:

## Have an option to compile a `.bt` file into an executable

In this approach, we would add a new option to `bpftrace` (let's say, `-o`), which instead of attaching to the probes we provided, would build an executable binary with our bpftrace program. We could then send this binary to a given server without having to install llvm/clang libraries there.

Downsides would be:

  - bpftrace would get more complex
  - I think we would still need libbpfcc installed on the server to attach to probes

## Compile a `.bt` file into an intermediate format, and run with a minimal bpftrace binary

Similar to the previous idea, but instead of creating an executable binary we would have an intermediate format with our probes as well as BPF programs for each probe. We could then use a minimal bpftrace binary (not linked against llvm/clang) to run this intermediate file.

On the upside, I think this would be more maintainable than the previous approach (internally we could also use the intermediate format). Downsides are the same as the previous approach.

---

I still think it's worth getting our static builds to work though, but if we can't do that reliably, we could start investigating those alternatives.

danobi:
> I think we would still need libbpfcc installed on the server to attach to probes

Curious, why?

> we would have an intermediate format with our probes as well as BPF programs for each probe

It sounds like we could even consider taking it a step further and create a custom bpf IR + compiler. FWIW, I've been tossing a similar idea around the last couple months. Clang is a bit heavy weight for the kind of bpf programs that bpftrace enables.

brendangregg:
> It sounds like we could even consider taking it a step further and create a custom bpf IR + compiler. 

Discussed that with @4ast ?

Right now I'm hoping to stabilize the API and fix bugs; changing the compiler chain may be worthwhile, but not high on my list at the moment!

dalehamel:
> most distributions don't provide the static library `libclang.a` on their libclang packages.

I think the main issue is with the alpine static build, and the way that it links to musl, per my investigation in #266. Are there any other problems with a static build we should be trying to solve for?

[...]

PR #1360, opened 2020-05-30
RFC: Store strings off-stack (in per-CPU scratch buffer)

Birch-san:
I've iterated on https://github.com/iovisor/bpftrace/pull/1357 based on review feedback.

The design is a lot simpler now. We allocate a per-CPU array map, big enough for your largest string. str() probe-reads strings into this map. Then, printf() probe-reads strings out of that map into its own map, which is submitted as a perf event.

Compared to the previous proposal: this design does not have the data race and coordination with userspace, nor does it introduce a "different" string structure, so should interoperate better with the existing string functions.

I found that I had to change printf()'s argvalue memcpy() into a probe_read(). We may find other interoperability problems like that between string functions, but they're easily solved.

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1,4 +1,5 @@
 #include "codegen_llvm.h"
+#include "allocation_helper.h"
 #include "arch/arch.h"
 #include "ast.h"
 #include "ast/async_event_types.h"
@@ -978,6 +979,9 @@ void CodegenLLVM::visit(Variable &var)
 {
   if (needMemcpy(var.type))
   {
+    // TODO: where needMapStorage(var.type) is true, we could use
+    // CreateMapLookupElem here instead. would there be any advantage
+    // to doing so?
     expr_ = variables_[var.ident];
   }
   else
@@ -1687,7 +1691,8 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
 
   // any map-backed variable will have already had storage allocated in entry
   // block
-  if (variables_.find(var.ident) == variables_.end() && !needMemcpy(var.type))
+  if (variables_.find(var.ident) == variables_.end() &&
+      !needMapStorage(var.type))
   {
     // alloca is hoisted to start of program, so doesn't matter which block
     // we're in
@@ -1697,7 +1702,9 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
 
   if (needMemcpy(var.type))
   {
-    // consider a CreateMapUpdateElem here
+    // TODO: where needMapStorage(var.type) is true, we could use
+    // CreateMapUpdateElem here instead. would there be any advantage
+    // to doing so?
     b_.CreateCopy(
         ctx_, variables_[var.ident], expr_, var.type.size, assignment.loc);
   }
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -15,6 +15,7 @@
 #include <csignal>
 #include <ctime>
 #include <fstream>
+#include <string_view>
 
 #include <llvm/Support/TargetRegistry.h>
 #include <llvm/IR/Constants.h>
@@ -91,12 +92,15 @@ void CodegenLLVM::visit(PositionalParameter &param)
 
 void CodegenLLVM::visit(String &string)
 {
-  string.str.resize(string.type.size-1);
-  Constant *const_str = ConstantDataArray::getString(module_->getContext(), string.str, true);
-  AllocaInst *buf = b_.CreateAllocaBPF(string.type, "str");
-  b_.CreateStore(const_str, buf);
+  int key = bpftrace_.str_map_keys_[static_cast<Node *>(&string)];
+  CallInst *buf = b_.CreateGetStrMap(ctx_, key, string.loc);
+  b_.CreateZeroInit(ctx_, buf, bpftrace_.strlen_, string.loc);
+  size_t space_for_null_len = std::min<size_t>(string.str.size(),
+                                               string.type.size - 1);
+  std::string_view shorter_str = { string.str.data(), space_for_null_len };
+  b_.CreateStoreConstStr(ctx_, shorter_str, buf, string.loc);
+
   expr_ = buf;
-  expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
 }
 
 // NB: we do not resolve identifiers that are structs. That is because in
@@ -310,21 +314,20 @@ void CodegenLLVM::visit(Call &call)
   if (call.func == "count")
   {
     Map &map = *call.map;
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
     b_.CreateStore(b_.CreateAdd(oldval, b_.getInt64(1)), newval);
     b_.CreateMapUpdateElem(ctx_, map, key, newval, call.loc);
 
     // oldval can only be an integer so won't be in memory and doesn't need lifetime end
-    b_.CreateLifetimeEnd(key);
     b_.CreateLifetimeEnd(newval);
     expr_ = nullptr;
   }
   else if (call.func == "sum")
   {
     Map &map = *call.map;
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
 
@@ -337,14 +340,13 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateMapUpdateElem(ctx_, map, key, newval, call.loc);
 
     // oldval can only be an integer so won't be in memory and doesn't need lifetime end
-    b_.CreateLifetimeEnd(key);
     b_.CreateLifetimeEnd(newval);
     expr_ = nullptr;
   }
   else if (call.func == "min")
   {
     Map &map = *call.map;
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
 
@@ -367,14 +369,13 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateBr(lt);
 
     b_.SetInsertPoint(lt);
-    b_.CreateLifetimeEnd(key);
     b_.CreateLifetimeEnd(newval);
     expr_ = nullptr;
   }
   else if (call.func == "max")
   {
     Map &map = *call.map;
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
 
@@ -394,7 +395,6 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateBr(lt);
 
     b_.SetInsertPoint(lt);
-    b_.CreateLifetimeEnd(key);
     b_.CreateLifetimeEnd(newval);
     expr_ = nullptr;
   }
@@ -502,16 +502,15 @@ void CodegenLLVM::visit(Call &call)
   {
     auto &arg = *call.vargs->at(0);
     auto &map = static_cast<Map&>(arg);
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
-    b_.CreateLifetimeEnd(key);
     expr_ = nullptr;
   }
   else if (call.func == "str")
   {
     AllocaInst *strlen = b_.CreateAllocaBPF(b_.getInt64Ty(), "strlen");
-    b_.CREATE_MEMSET(strlen, b_.getInt8(0), sizeof(uint64_t), 1);
     if (call.vargs->size() > 1) {
+      b_.CreateStore(b_.getInt64(0), strlen);
       auto scoped_del = accept(call.vargs->at(1));
       Value *proposed_strlen = b_.CreateAdd(expr_, b_.getInt64(1)); // add 1 to accommodate probe_read_str's null byte
 
@@ -527,16 +526,21 @@ void CodegenLLVM::visit(Call &call)
     } else {
       b_.CreateStore(b_.getInt64(bpftrace_.strlen_), strlen);
     }
-    AllocaInst *buf = b_.CreateAllocaBPF(bpftrace_.strlen_, "str");
-    b_.CREATE_MEMSET(buf, b_.getInt8(0), bpftrace_.strlen_, 1);
+
+    int key = bpftrace_.str_map_keys_[static_cast<Node *>(&call)];
+    CallInst *str_map = b_.CreateGetStrMap(ctx_, key, call.loc);
+    b_.CreateZeroInit(ctx_, str_map, bpftrace_.strlen_, call.loc);
     auto arg0 = call.vargs->front();
     auto scoped_del = accept(call.vargs->front());
-    b_.CreateProbeReadStr(
-        ctx_, buf, b_.CreateLoad(strlen), expr_, arg0->type.GetAS(), call.loc);
+    b_.CreateProbeReadStr(ctx_,
+                          str_map,
+                          b_.CreateLoad(strlen),
+                          expr_,
+                          arg0->type.GetAS(),
+                          call.loc);
     b_.CreateLifetimeEnd(strlen);
 
-    expr_ = buf;
-    expr_deleter_ = [this,buf]() { b_.CreateLifetimeEnd(buf); };
+    expr_ = str_map;
   }
   else if (call.func == "buf")
   {
@@ -563,7 +567,7 @@ void CodegenLLVM::visit(Call &call)
       auto &arg = *call.vargs->at(0);
       fixed_buffer_length = arg.type.GetNumElements() *
                             arg.type.GetElementTy()->size;
-      length = b_.getInt8(fixed_buffer_length);
+      length = b_.getInt32(fixed_buffer_length);
     }
 
     auto elements = AsyncEvent::Buf().asLLVMType(b_, fixed_buffer_length);
@@ -572,7 +576,12 @@ void CodegenLLVM::visit(Call &call)
     StructType *buf_struct = b_.GetStructType(dynamic_sized_struct_name,
                                               elements,
                                               false);
-    AllocaInst *buf = b_.CreateAllocaBPF(buf_struct, "buffer");
+    int struct_size = layout_.getTypeAllocSize(buf_struct);
+
+    auto buf_struct_ptr_ty = PointerType::get(buf_struct, 0);
+    int key = bpftrace_.buf_map_keys_[static_cast<Node *>(&call)];
+    CallInst *buf = b_.CreateGetBufMap(ctx_, key, buf_struct_ptr_ty, call.loc);
+    b_.CreateZeroInit(ctx_, buf, struct_size, call.loc);
 
     Value *buf_len_offset = b_.CreateGEP(buf,
                                          { b_.getInt32(0), b_.getInt32(0) });
@@ -581,22 +590,13 @@ void CodegenLLVM::visit(Call &call)
 
     Value *buf_data_offset = b_.CreateGEP(buf,
                                           { b_.getInt32(0), b_.getInt32(1) });
-    b_.CREATE_MEMSET(buf_data_offset,
-                     b_.GetIntSameSize(0, elements.at(0)),
-                     fixed_buffer_length,
-                     1);
 
     auto scoped_del = accept(call.vargs->front());
     auto arg0 = call.vargs->front();
-    b_.CreateProbeRead(ctx_,
-                       static_cast<AllocaInst *>(buf_data_offset),
-                       length,
-                       expr_,
-                       arg0->type.GetAS(),
-                       call.loc);
+    b_.CreateProbeRead(
+        ctx_, buf_data_offset, length, expr_, arg0->type.GetAS(), call.loc);
 
     expr_ = buf;
-    expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
   }
   else if (call.func == "kaddr")
   {
@@ -633,25 +633,8 @@ void CodegenLLVM::visit(Call &call)
     AllocaInst *second = b_.CreateAllocaBPF(b_.getInt64Ty(),
                                             call.func + "_second");
     Value *perfdata = b_.CreateGetJoinMap(ctx_, call.loc);
-    Function *parent = b_.GetInsertBlock()->getParent();
-
-    BasicBlock *zero = BasicBlock::Create(module_->getContext(),
-                                          "joinzero",
-                                          parent);
-    BasicBlock *notzero = BasicBlock::Create(module_->getContext(),
-                                             "joinnotzero",
-                                             parent);
-
-    b_.CreateCondBr(b_.CreateICmpNE(perfdata,
-                                    ConstantExpr::getCast(Instruction::IntToPtr,
-                                                          b_.getInt64(0),
-                                                          b_.getInt8PtrTy()),
-                                    "joinzerocond"),
-                    notzero,
-                    zero);
 
     // arg0
-    b_.SetInsertPoint(notzero);
     b_.CreateStore(b_.getInt64(asyncactionint(AsyncAction::join)), perfdata);
     b_.CreateStore(b_.getInt64(join_id_),
                    b_.CreateGEP(perfdata, b_.getInt64(8)));
@@ -687,10 +670,6 @@ void CodegenLLVM::visit(Call &call)
         perfdata,
         8 + 8 + bpftrace_.join_argnum_ * bpftrace_.join_argsize_);
 
-    b_.CreateBr(zero);
-
-    // done
-    b_.SetInsertPoint(zero);
     expr_ = nullptr;
   }
   else if (call.func == "ksym")
@@ -991,13 +970,8 @@ void CodegenLLVM::visit(Call &call)
 
 void CodegenLLVM::visit(Map &map)
 {
-  AllocaInst *key = getMapKey(map);
-  Value *value = b_.CreateMapLookupElem(ctx_, map, key, map.loc);
-  expr_ = value;
-
-  if (dyn_cast<AllocaInst>(value))
-    expr_deleter_ = [this, value]() { b_.CreateLifetimeEnd(value); };
-  b_.CreateLifetimeEnd(key);
+  auto [key, scoped_key_deleter] = getMapKey(map);
+  expr_ = b_.CreateMapLookupElem(ctx_, map, key, map.loc);
 }
 
 void CodegenLLVM::visit(Variable &var)
@@ -1073,7 +1047,7 @@ void CodegenLLVM::visit(Binop &binop)
                                left_as,
                                right_string,
                                right_as,
-                               len + 1,
+                               len,
                                binop.loc,
                                inverse);
     }
@@ -1217,7 +1191,7 @@ void CodegenLLVM::visit(Unop &unop)
         if (unop.expr->is_map)
         {
           Map &map = static_cast<Map&>(*unop.expr);
-          AllocaInst *key = getMapKey(map);
+          auto [key, scoped_key_deleter] = getMapKey(map);
           Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, unop.loc);
           AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_newval");
           if (is_increment)
@@ -1225,7 +1199,6 @@ void CodegenLLVM::visit(Unop &unop)
           else
             b_.CreateStore(b_.CreateSub(oldval, b_.getInt64(1)), newval);
           b_.CreateMapUpdateElem(ctx_, map, key, newval, unop.loc);
-          b_.CreateLifetimeEnd(key);
 
           if (unop.is_post_op)
             expr_ = oldval;
@@ -1311,12 +1284,18 @@ void CodegenLLVM::visit(Ternary &ternary)
   BasicBlock *right_block = BasicBlock::Create(module_->getContext(), "right", parent);
   BasicBlock *done = BasicBlock::Create(module_->getContext(), "done", parent);
   // ordering of all the following statements is important
-  Value *result = ternary.type.IsNoneTy()
-                      ? nullptr
-                      : b_.CreateAllocaBPF(ternary.type, "result");
-  AllocaInst *buf = ternary.type.IsNoneTy()
-                        ? nullptr
-                        : b_.CreateAllocaBPF(ternary.type, "buf");
+  AllocaInst *result;
+  CallInst *buf;
+  if (ternary.type.IsIntTy())
+  {
+    result = b_.CreateAllocaBPF(ternary.type, "result");
+  }
+  else if (ternary.type.IsStringTy())
+  {
+    int key = bpftrace_.str_map_keys_[static_cast<Node *>(&ternary)];
+    buf = b_.CreateGetStrMap(ctx_, key, ternary.loc);
+    b_.CreateZeroInit(ctx_, buf, bpftrace_.strlen_, ternary.loc);
+  }
   Value *cond;
   auto scoped_del = accept(ternary.cond);
   cond = expr_;
@@ -1352,12 +1331,12 @@ void CodegenLLVM::visit(Ternary &ternary)
     // copy selected string via CreateMemCpy
     b_.SetInsertPoint(left_block);
     auto scoped_del_left = accept(ternary.left);
-    b_.CREATE_MEMCPY(buf, expr_, ternary.type.size, 1);
+    b_.CreateCopy(ctx_, buf, expr_, ternary.type.size, ternary.left->loc);
     b_.CreateBr(done);
 
     b_.SetInsertPoint(right_block);
     auto scoped_del_right = accept(ternary.right);
-    b_.CREATE_MEMCPY(buf, expr_, ternary.type.size, 1);
+    b_.CreateCopy(ctx_, buf, expr_, ternary.type.size, ternary.right->loc);
     b_.CreateBr(done);
 
     b_.SetInsertPoint(done);
@@ -1647,28 +1626,26 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
   if (!expr_) // Some functions do the assignments themselves
     return;
 
-  Value *val, *expr;
-  expr = expr_;
-  AllocaInst *key = getMapKey(map);
+  Value *val;
   if (shouldBeOnStackAlready(assignment.expr->type))
   {
-    val = expr;
+    val = expr_;
   }
   else if (map.type.IsRecordTy())
   {
     if (assignment.expr->type.is_internal)
     {
-      val = expr;
+      val = expr_;
     }
     else
     {
-      // expr currently contains a pointer to the struct
+      // expr_ currently contains a pointer to the struct
       // We now want to read the entire struct in so we can save it
       AllocaInst *dst = b_.CreateAllocaBPF(map.type, map.ident + "_val");
       b_.CreateProbeRead(ctx_,
                          dst,
                          map.type.size,
-                         expr,
+                         expr_,
                          assignment.expr->type.GetAS(),
                          assignment.loc);
       val = dst;
@@ -1680,12 +1657,13 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
     // expr currently contains a pointer to the struct
     // and that's what we are saving
     AllocaInst *dst = b_.CreateAllocaBPF(map.type, map.ident + "_ptr");
-    b_.CreateStore(expr, dst);
+    b_.CreateStore(expr_, dst);
     val = dst;
     self_alloca = true;
   }
   else
   {
+    Value *expr = expr_;
     if (map.type.IsIntTy())
     {
       // Integers are always stored as 64-bit in map values
@@ -1695,8 +1673,8 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
     b_.CreateStore(expr, val);
     self_alloca = true;
   }
+  auto [key, scoped_key_deleter] = getMapKey(map);
   b_.CreateMapUpdateElem(ctx_, map, key, val, assignment.loc);
-  b_.CreateLifetimeEnd(key);
   if (self_alloca)
     b_.CreateLifetimeEnd(val);
 }
@@ -1707,15 +1685,21 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
 
   auto scoped_del = accept(assignment.expr);
 
-  if (variables_.find(var.ident) == variables_.end())
+  // any map-backed variable will have already had storage allocated in entry
+  // block
+  if (variables_.find(var.ident) == variables_.end() && !needMemcpy(var.type))
   {
-    AllocaInst *val = b_.CreateAllocaBPFInit(var.type, var.ident);
+    // alloca is hoisted to start of program, so doesn't matter which block
+    // we're in
+    Value *val = b_.CreateAllocaBPFInit(var.type, var.ident);
     variables_[var.ident] = val;
   }
 
   if (needMemcpy(var.type))
   {
-    b_.CREATE_MEMCPY(variables_[var.ident], expr_, var.type.size, 1);
+    // consider a CreateMapUpdateElem here
+    b_.CreateCopy(
+        ctx_, variables_[var.ident], expr_, var.type.size, assignment.loc);
   }
   else
   {
@@ -1924,7 +1908,18 @@ void CodegenLLVM::generateProbe(Probe &probe,
   {
     auto scoped_del = accept(probe.pred);
   }
+
   variables_.clear();
+  // hoist declarations for all map-backed variables
+  // (currently this is any type for which needMemcpy() returns true)
+  for (auto &[var_name, var] : bpftrace_.vars_[&probe])
+  {
+    CallInst *val = b_.CreateGetVarMap(ctx_, var_name, var);
+    b_.CreateZeroInit(
+        ctx_, val, var.semantic.sized_type.size, var.semantic.loc);
+    variables_[var_name] = static_cast<Value *>(val);
+  }
+
   for (Statement *stmt : *probe.stmts)
   {
     auto scoped_del = accept(stmt);
@@ -1934,6 +1929,7 @@ void CodegenLLVM::generateProbe(Probe &probe,
 
 void CodegenLLVM::visit(Probe &probe)
 {
+  probe_ = &probe;
   FunctionType *func_type = FunctionType::get(
       b_.getInt64Ty(),
       {b_.getInt8PtrTy()}, // struct pt_regs *ctx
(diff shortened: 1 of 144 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -2318,6 +2324,41 @@ int SemanticAnalyser::create_maps(bool debug)
...
   }
 
+  size_t max_zero_buffer_size_ = 0;
+
+  if (needs_fmtstr_map_)
+  {
+    std::string map_ident = "fmtstr";
+
+    size_t printf_struct_size = sizeof(size_t) + max_fmtstr_args_size_;
+    SizedType type = SizedType(Type::fmtstr, printf_struct_size);
```
danobi, line 2334: Do we really need a a Type::fmtstr? Isn't a format string still a string?

src/ast/codegen_llvm.h
```diff
@@ -87,6 +87,7 @@ class CodegenLLVM : public Visitor {
   DataLayout layout_;
   Value *expr_ = nullptr;
   std::function<void()> expr_deleter_; // intentionally empty
+  bool expr_points_to_map_value = false;
```
danobi, line 90: style nit
```suggestion
  bool expr_points_to_map_value_ = false;
```

src/ast/codegen_llvm.cpp
```diff
@@ -2297,23 +2320,44 @@ void CodegenLLVM::createFormatStringCall(Call &call, int &id, CallArgs &call_arg
...
     Value *offset = b_.CreateGEP(fmt_args, {b_.getInt32(0), b_.getInt32(i)});
+    // TODO: consider checking whether expr_ is a nullptr to map value
     if (arg.type.IsAggregate())
     {
-      b_.CREATE_MEMCPY(offset, expr_, arg.type.size, 1);
+      if (expr_points_to_map_value) {
+        b_.CreateProbeRead(ctx_, offset, arg.type.size, expr_, call.loc);
+      } else {
+        b_.CREATE_MEMCPY(offset, expr_, arg.type.size, 1);
+      }
```
danobi, line 2360: Shouldn't these arms be flipped? If it's in a map we can do a regular memcpy right?

danobi:
Also please clang-format and get the builds to pass

fbs:
Noticed my comment from the last version got lost:

> Being able to read in loooooooong strings would be great but I don't think we should change the current behaviour of str(), as a construct like @[str(arg0)]=arg1 should continue to work. I'd rather see a new call for this async version of str.

src/parser.yy
```diff
@@ -359,9 +359,9 @@ ident : IDENT         { $$ = $1; }
       ;
 
 call : CALL "(" ")"                 { $$ = new ast::Call($1, @$); }
-     | CALL "(" vargs ")"           { $$ = new ast::Call($1, $3, @$); }
+     | CALL "(" vargs ")"           { $$ = new ast::Call($1, @$, $3); }
      | CALL_BUILTIN  "(" ")"        { $$ = new ast::Call($1, @$); }
-     | CALL_BUILTIN "(" vargs ")"   { $$ = new ast::Call($1, $3, @$); }
+     | CALL_BUILTIN "(" vargs ")"   { $$ = new ast::Call($1, @$, $3); }
```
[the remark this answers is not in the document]
Birch-san: my previous design, https://github.com/iovisor/bpftrace/pull/1357, used specialisations of Call (i.e. a StrCall that had state upon it such as an IMap).

I found during that effort that the four overloads of Call's constructor could be simplified to just 1, if the order of arguments was shuffled to put optional args on the end.

but ultimately this change isn't required to make this feature work. should I remove it?

src/bpftrace.h
```diff
@@ -85,6 +87,15 @@ struct HelperErrorInfo
...
 
+struct PerfEventCallbackCookie {
+  PerfEventCallbackCookie(BPFtrace* bpftrace, int cpu)
+  : bpftrace(bpftrace)
+  , cpu(cpu)
+  {}
+  BPFtrace* bpftrace;
+  int cpu;
+};
+
```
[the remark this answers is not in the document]
Birch-san: my previous design, https://github.com/iovisor/bpftrace/pull/1357, coordinated map lookups between the BPF program and userspace. `cpu` was used to determine where to locate string content in the per-CPU map.

this change isn't required any more, but I kept it because it seemed like the correct way to expose CPU in case a future feature has the same needs (userspace lookup of entry from a per-CPU map when only the key is known). should I remove it for now?

src/ast/semantic_analyser.cpp
Birch-san: this was inherited from @mmarchini's https://github.com/iovisor/bpftrace/pull/750.

fmtstr is the structure composed by `CreateFormatString()` (e.g. printf, join) and emitted as a perf event.

I think now that str is handled the same way by map.cpp, fmtstr could be changed to str without breaking anything. but that risks being disingenuous; it's not a "char array of length n" — it's a struct (whose length depends on the lengths of the operands to printf()).

Since it's a struct not a string: I think it still makes sense to continue calling it fmtstr.

src/ast/codegen_llvm.h
Birch-san: agreed; will incorporate

src/ast/codegen_llvm.cpp
Birch-san: I was surprised, but MEMCPY from str()'s map to printf()'s map failed as invalid. something to do with `inv`. I can grab the message if you'd like to see? but this way round was definitely necessary to fix it.

Birch-san:
> Being able to read in loooooooong strings would be great but I don't think we should change the current behaviour of str(), as a construct like @[str(arg0)]=arg1 should continue to work. I'd rather see a new call for this async version of str.

@fbs thanks for carrying over this comment, and sorry to not address it directly. I felt that this was a very valid criticism of the previous RFC (https://github.com/iovisor/bpftrace/pull/1357); that version certainly could be called async (BPF emitted a map coordinate via perf, and userspace had to lookup that map entry later). and yes, in that version the output of str() was not compatible with other function calls.

But the design in this review iteration is far simpler; when str() returns, `expr_` just points to string contents stored in a map value. this _should_ give it interoperability with other functions which consume strings.

Do you have a program handy that uses a syntax like `@[str(arg0)]=arg1`? I'd like to try it out and see if the new str() is still compatible. I may need to tweak some MEMCPY() calls similarly to what I did with printf:

```c++
      if (expr_points_to_map_value) {
        b_.CreateProbeRead(ctx_, offset, arg.type.size, expr_, call.loc);
      } else {
        b_.CREATE_MEMCPY(offset, expr_, arg.type.size, 1);
      }
```

Moreover, @danobi suggested in https://github.com/iovisor/bpftrace/pull/1346#issuecomment-634312784 that it'd be okay to change _all_ strings to be map-backed. so, perhaps str() won't be a special case.

Certainly, it seems to me at the moment that it should be possible to implement this such that the user notices no difference to the bpftrace language (i.e. no additional str() call required, and all string-using functions continue to interoperate).

Birch-san:
> Also please clang-format and get the builds to pass

@danobi I think I'm not quite ready to write tests for this, since I have to resolve the TODOs around "what happens if map_lookup_elem returns a nullptr" (i.e. during str() or during printf()). I'm planning to make it Return from the BPF program and emit a perf event to produce an error message (dropped event). Progress so far on that: https://github.com/Birch-san/bpftrace/compare/long-strings-4...Birch-san:long-strings-5

Birch-san:
> Also please clang-format and get the builds to pass

@danobi I've had a stab at that, but maybe I'm doing something wrong?

```bash
git --no-pager diff --name-only master | xargs clang-format -i
```

The output looks correct, but it changes 4000 lines:

https://github.com/Birch-san/bpftrace/compare/107371e...long-strings-4-clang-format

There's plenty of stuff absent from my own changes that got caught by using clang-format in this way. Is there a way to just target the new lines I've added? Do we normally run clang-format?

Birch-san:
Update: I've now added handling for the case where str() or printf() fails to acquire a map element. I figured it made sense to re-use the existing helper_error infrastructure, except make it a bit harsher. Because this isn't a user-managed map where it could be normal to encounter missing entries -- it's one of bpftrace's maps, where we really expect the contents to exist. I don't expect a user to be interested in sniffing an error to decide how to continue, I think one of their invariants has been violated and it makes sense to return from the BPF program.

I'm even tempted to make that harsher still, and detach the probe if this ever happens. Since you could easily be in an invalid state. Imagine having logic after a str() or printf() — you'd expect this logic to get hit every time, but if str()/printf() fails to acquire a map entry then we'll have broken one of your guarantees.

fbs:
> Do you have a program handy that uses a syntax like `@[str(arg0)]=arg1`? I'd like to try it out and see if the new str() is still compatible. I may need to tweak some MEMCPY() calls similarly to what I did with printf:

Not really. #1067  is somewhat related but does not currently work

> Moreover, @danobi suggested in [#1346 (comment)](https://github.com/iovisor/bpftrace/pull/1346#issuecomment-634312784) that it'd be okay to change _all_ strings to be map-backed. so, perhaps str() won't be a special case.

`str()` is currently the only thing that `probe_read_str` into a stack var, I'm not convinced that we  should remove that completely. There is some room for improvement as `@[str()]` should read directly into  the map,  not through a stack buffer.

> > Also please clang-format and get the builds to pass
> 
> @danobi I've had a stab at that, but maybe I'm doing something wrong?
> 
> ```shell
> git --no-pager diff --name-only master | xargs clang-format -i
> ```
> 
> The output looks correct, but it changes 4000 lines:

You  need git-clang-format to only format changed lines, your line clang-formats the whole file. Can use the ci  job as example

src/parser.yy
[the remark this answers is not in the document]
Birch-san: removed to make the review smaller.

src/bpftrace.h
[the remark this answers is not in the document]
Birch-san: removed to make the review smaller.

src/ast/codegen_llvm.h
Birch-san: fixed

Birch-san:
> I'm wondering now what happens if you do something like:
> 
> ```
> $x = str(one);
> $y = str(two);
> printf("%s\n", $x);
> ```
> 
> Does $y clobber $x?

@danobi conceptually it shouldn't; I expect variable assignment to do a MEMCPY (will need to change to be a probe read now), so as soon as line 1 finishes, the shared scratch space that str() calls use is free to be re-used.

Do you know a good probe I could invoke that would give me two different strings to try `str()` upon (so I can try and make one clobber the other)?

It looks like map value assignment needs a bit more attention — they're stack-bounded somewhere, and they blow up if I try to copy into them a string longer than the map is sized for.

> Also please clang-format and get the builds to pass

I've now clang-formatted and removed any code that was unrelated to the strings change. I think before I update tests, I probably need to figure out the limits around map assignment and whether they're regressions.

src/ast/semantic_analyser.cpp
Birch-san: removed to make the review smaller.

src/ast/codegen_llvm.cpp
danobi: Looking at this again, this looks right. Not sure what I was thinking before

danobi:
@Birch-san 

> Do you know a good probe I could invoke that would give me two different strings to try str() upon (so I can try and make one clobber the other)?

rename(2) seems like it could do the trick:
```
tracepoint:syscalls:sys_enter_rename
    int __syscall_nr;
    const char * oldname;
    const char * newname;
```

> CI still seems dissatisfied. Not sure what I'm missing.

I usually just run `git clang-format master` on my topic branch and it "just" works. I also do it per-commit like this: `git clang-format HEAD~1`.

danobi:
@fbs 

> str() is currently the only thing that probe_read_str into a stack var, I'm not convinced that we should remove that completely. There is some room for improvement as @[str()] should read directly into the map, not through a stack buffer.

Correct me if I'm wrong -- it seems like there's 2 concerns here:

1. Concern that the user may have a use case where they need a string on stack
2. The proposed implementation in this PR could be further optimized

Does that sound right? 

I read the reference docs on `str()` again and it doesn't seem like there's promises the strings will be on stack. My feeling is if map-based strings can be pulled off without regressions it would be a win and the user experiences no visible side effects (other than max string size being increased)

src/ast/codegen_llvm.cpp
```diff
@@ -495,14 +499,15 @@ void CodegenLLVM::visit(Call &call)
     } else {
       b_.CreateStore(b_.getInt64(bpftrace_.strlen_), strlen);
     }
-    AllocaInst *buf = b_.CreateAllocaBPF(bpftrace_.strlen_, "str");
-    b_.CREATE_MEMSET(buf, b_.getInt8(0), bpftrace_.strlen_, 1);
     call.vargs->front()->accept(*this);
-    b_.CreateProbeReadStr(ctx_, buf, b_.CreateLoad(strlen), expr_, call.loc);
+    // zero it out first
+    b_.CreateProbeRead(ctx_, str_map, bpftrace_.strlen_,
+                      ConstantExpr::getCast(Instruction::IntToPtr, zeroed_area_ptr, b_.getInt8PtrTy()), call.loc);
```
danobi, line 505: Would a memset work here instead? I see the intention is to clear out map entry before reading the string in but it feels a little funky.

I did a small test and it loaded ok on my system:
http://ix.io/2oRP
diffed on https://github.com/danobi/libbpf-rs

src/ast/codegen_llvm.cpp
```diff
@@ -2300,20 +2305,31 @@ void CodegenLLVM::createFormatStringCall(Call &call, int &id, CallArgs &call_arg
...
-  AllocaInst *fmt_args = b_.CreateAllocaBPF(fmt_struct, call_name + "_args");
-  b_.CREATE_MEMSET(fmt_args, b_.getInt8(0), struct_size, 1);
+  Value *fmt_args = b_.CreateGetFmtStrMap(ctx_, fmt_struct, call.loc);
+
+  auto fmt_struct_ptr_ty = PointerType::get(fmt_struct, 0);
+
+  auto zeroed_area_ptr = b_.getInt64(reinterpret_cast<uintptr_t>(bpftrace_.zero_buffer_->data()));
+
+  b_.CreateProbeRead(ctx_, fmt_args, struct_size,
+                     ConstantExpr::getCast(Instruction::IntToPtr, zeroed_area_ptr, fmt_struct_ptr_ty), call.loc);
```
danobi, line 2315: Same as above

src/ast/semantic_analyser.cpp
danobi: Hmm, I'm not fully convinced on this point. I'm not really seeing in which cases the type system would need to disambiguate between a format string and a regular string. For `printf`, `system`, etc, the semantic analyser + codegen knows it's a format string based on the position of the argument. I'm not really sure why we would need additional help to know it's a format string.

For example in C, a format string is still a `char *`. `sprintf` takes a `char *` + some additional arguments and "creates" and new `char *`.

My concern with new types is that each new type creates edge cases and strange interactions between types. For example, we have a `count` type that cannot be compared to `int` type. (That one has a technical reason but it'd be awesome if `count` would just be an `int`).

Is there something I'm missing?

danobi:
> I'm even tempted to make that harsher still, and detach the probe if this ever happens. Since you could easily be in an invalid state. Imagine having logic after a str() or printf() — you'd expect this logic to get hit every time, but if str()/printf() fails to acquire a map entry then we'll have broken one of your guarantees.

I'd tend to agree but even harsher still. I'm looking at the kernel implementation ( https://github.com/torvalds/linux/blob/master/kernel/bpf/arraymap.c#L240-L250 ) and I don't really see how a lookup on a percpu map can fail unless there's a bug in codegen. So I propose abort()ing bpftrace so it's clear it's a bug.

fbs:
> @fbs
> Does that sound right?
> 
> I read the reference docs on `str()` again and it doesn't seem like there's promises the strings will be on stack. My feeling is if map-based strings can be pulled off without regressions it would be a win and the user experiences no visible side effects (other than max string size being increased)

I've tried to come up with a use case that would require keeping it but I couldn't.  If someone comes up with a good usecase we can always add it back in, so let's not block this on it :)

src/ast/codegen_llvm.cpp
Birch-san: The same question was asked of @mmarchini in his original RFC on storing printf args in a map: https://github.com/iovisor/bpftrace/pull/750#discussion_r292685014

> memset will only work for small values (even if we're using a map). I'm on another branch right now, I'll try to share more information later and I'll add a comment here explaining why this is implemented this way.

Unfortunately I don't know more detail than that! I'd definitely appreciate some detail from @mmarchini on this one.

src/ast/semantic_analyser.cpp
Birch-san: I suspect the entire motivation of the original code was "the type system determines what type of Map you will get from map.cpp".

the additional changes in my PR (since the original printf RFC https://github.com/iovisor/bpftrace/pull/750) ensure that `Types::str` gets a per-CPU array from map.cpp, so on that basis I think we no longer need fmtstr — regular str should serve the same purpose now.

Birch-san:
> I'd tend to agree but even harsher still. I'm looking at the kernel implementation ( https://github.com/torvalds/linux/blob/master/kernel/bpf/arraymap.c#L240-L250 ) and I don't really see how a lookup on a percpu map can fail unless there's a bug in codegen. So I propose abort()ing bpftrace so it's clear it's a bug.

yes, I agree "stop everything" makes sense for a "well this should never happen anyway".

when you say abort(), do you mean like the abort() calls in codegen_llvm.cpp? (https://github.com/iovisor/bpftrace/blob/9515965/src/ast/codegen_llvm.cpp#L64) I think we can only use those to abort compilation of the LLVM IR. sadly I think we don't find out the result of map_lookup_elem until BPF runtime, so we would not be able to use an abort().

I think we'd instead have to handle it the same way exit() works: send a perf event, return from the current BPF function, (https://github.com/iovisor/bpftrace/blob/9515965/src/ast/codegen_llvm.cpp#L759), and then have userspace do a `bpftrace->request_finalize()` (https://github.com/iovisor/bpftrace/blob/f38b144/src/bpftrace.cpp#L466). easy enough; just needs `AsyncAction::helper_error` to have a "fatal error" boolean.

Birch-san:
> @Birch-san
> 
> > Do you know a good probe I could invoke that would give me two different strings to try str() upon (so I can try and make one clobber the other)?
> 
> rename(2) seems like it could do the trick:
> 
> ```
> tracepoint:syscalls:sys_enter_rename
>     int __syscall_nr;
>     const char * oldname;
>     const char * newname;
> ```

Brilliant; I'll give that a try.

> > CI still seems dissatisfied. Not sure what I'm missing.
> 
> I usually just run `git clang-format master` on my topic branch and it "just" works. I also do it per-commit like this: `git clang-format HEAD~1`.

okay, I'll give it another stab with one of those approaches and see if it goes any better.

Birch-san:
> There is some room for improvement as @[str()] should read directly into the map, not through a stack buffer.

@fbs I wonder whether an LLVM optimization pass could be introduced to elide this copy in the general case, rather than complicating the codegen.

there'll be a lot of cases like this. for example, with `printf("%s", str())`:  
str() writes into a map as scratch space, then printf copies that content into another map, which perf_event_output exfiltrates. there's a wasted step (i.e. str() doesn't need its own scratch space -- it could output directly into printf's map). and I think when we used `AllocaInst()` for scratch space, LLVM knew how to optimize away these copies (the optimized LLVM IR looks like str() writes directly into an offset of the printf buffer).

I think if we annotated these particular maps as "we use this like stack-allocated memory", an optimization pass could be written… somehow. :)

src/ast/codegen_llvm.cpp
Birch-san: this is a wild guess, but IIRC I observed that the LLVM IR output for memset() was something like:

```
store a 0 at offset +16
store a 0 at offset +8
store a 0 at offset +0
```

so maybe you reach limits of BPF program size if you memset() a large enough value? whereas I think a call to probe_read() would do the work outside of our BPF program.

admittedly I'm not satisfied with that guess, since BPF programs are allowed to be pretty big nowadays.

danobi: Maybe llvm unrolls the memset on small enough regions of memory. And maybe verifier doesn't like loop-based memsets.

danobi:
> I think we'd instead have to handle it the same way exit() works: send a perf event, return from the current BPF function, (https://github.com/iovisor/bpftrace/blob/9515965/src/ast/codegen_llvm.cpp#L759), and then have userspace do a bpftrace->request_finalize() (https://github.com/iovisor/bpftrace/blob/f38b144/src/bpftrace.cpp#L466). easy enough; just needs AsyncAction::helper_error to have a "fatal error" boolean.

Yeah that's what I meant

src/ast/semantic_analyser.cpp
Birch-san: removed fmtstr type.

Birch-san:
> I'd tend to agree but even harsher still.

@danobi I've now made map_lookup_elem fatal, removed Types::fmtstr and run clang-format.

Now need to investigate how well this interoperates with other string functionality (e.g. storing in maps, assigning to variables).

Birch-san:
> I'm wondering now what happens if you do something like:
> 
> ```
> $x = str(one);
> $y = str(two);
> printf("%s\n", $x);
> ```
> 
> Does $y clobber $x?

@danobi I've now tried that with rename, and it works fine:

```bash
sudo bpftrace -e 'tracepoint:syscalls:sys_enter_renameat2 {
  $x=str(args->oldname);
  $y=str(args->newname);
  printf("<%s; %s>\n", $x, $y);
}'
Attaching 1 probe...
</var/tmp/rename_me; /var/tmp/rename_me2>
```

Triggered with:

```bash
mv /var/tmp/rename_me /var/tmp/rename_me2
```

Birch-san:
> a construct like `@[str(arg0)]=arg1` should continue to work

@fbs I just tried this map assignment, and it looks like off-stack strings broke it. I got a segfault during semantic analysis, because this case came out as a nullptr:

https://github.com/iovisor/bpftrace/blob/9f975da07f1019d51e18d7362ecc93492dc00860/src/ast/codegen_llvm.cpp#L1982-L1984

The specific program I tried was this (press Ctrl+C :

```bash
bpftrace -e 'tracepoint:syscalls:sys_enter_renameat2 {
  @[str(args->oldname)]=1;
}
interval:s:1 {
  print(@);
  clear(@);
}'
```

Triggered with:

```bash
mv /var/tmp/rename_me /var/tmp/rename_me2
```

Until now: map keys would've always been pointers to AllocaInst memory. But strings are now CallInst memory provided by map_lookup_elem.

So, I needed to update all map functions to accept keys of either AllocaInst or CallInst. This actually went pretty well:  
https://github.com/Birch-san/bpftrace/compare/long-strings-4...Birch-san:long-strings-7

And it works; I got the expected output when my probe fired:

```
@[/var/tmp/rename_me]: 1
```

But I'm wondering whether `std::variant<AllocaInst *, CallInst *>` is overkill; maybe I should just downcast everything; make `getMapKey()` return `Value *`, make all Map helpers accept `Value *`, and then have the helpers sniff what value they receive:

```c++
if (auto *AI = dyn_cast<AllocaInst>(Val))
```

Depends whether we want "specific types + compile-time safety" of std::variant, or "smaller diff, reliance on RTTI" of dynamic cast.

I guess I'd like to know "why do our map helpers ask for a key as specific as AllocaInst?" -- I notice that we call the key's `getType()`. AllocaInst overloads getType() (to return a PointerTy, rather than a vague llvm::Type), so I suppose this is the reason.

Anyway, any thoughts on whether to simplify this to use just `Value *` and a `dyn_cast<AllocaInst>`?

Birch-san:
I tried changing the CreateMap helpers to accept `Value *`, and that went fine. I think they weren't relying on any specific features of `AllocaInst *`.

The assertion `key->getType()->IsPointerTy()` succeeds without needing to know the subtype. CallInst memory passes this assertion fine too.

I'm not sure whether there's any consequences here:

```c++
// does it matter that key is now a Value* instead of an AllocaInst*?
// key->getType() will no longer use AllocaInst*'s overload (which casts to more specific `PointerType`)
FunctionType *lookup_func_type = FunctionType::get(
      getInt8PtrTy(), { map_ptr->getType(), key->getType() }, false);
```

Seems to be working fine though.

So, regarding "make str() keys work again", here's the `std::variant` approach (from my previous comment): https://github.com/Birch-san/bpftrace/compare/9f975da...long-strings-7  
and here's the simplification I've applied to do it without `std::variant` (i.e. downcast to `Value*` instead): https://github.com/Birch-san/bpftrace/compare/long-strings-7...long-strings-7-simpler

I think the simpler version is what I'll merge into this PR, since it's the smaller change.

Birch-san:
I've fixed another regression with associative arrays: storing string values.

```bash
sudo ./src/bpftrace -e 'tracepoint:syscalls:sys_enter_renameat2 {
  @[str(args->oldname)]="test";
  exit();
}'
Attaching 1 probe...

@[/var/tmp/rename_me]: test
```

Triggered with:

```bash
mv /var/tmp/rename_me /var/tmp/rename_me2
```

The fix is d371a38cf7ffdd4c5180a5d10837c981f2ee46d0, but it'll need a better design.  
The problem was that "all maps that people store strings into" became per-CPU array maps.  
Whereas I only wanted my two scratch maps — "str" and "fmtstr" — to become per-CPU array maps. In all other cases, the usual hash map suffices.

=====

The next problem I've found is:

```bash
sudo ./src/bpftrace -e 'tracepoint:syscalls:sys_enter_renameat2 {
  @[str(args->oldname)]=str(args->newname);
  exit();
}'
Attaching 1 probe...

@[/var/tmp/rename_me]: /var/tmp/rename_me
```

Triggered with:

```bash
mv /var/tmp/rename_me /var/tmp/rename_me2
```

Unfortunately the key and the value _both_ become `/var/tmp/rename_me`, which means that the key "clobbers" the value.

I think that makes sense; the map assignment consumes one str() without making a copy of its contents, then consumes another str() without copying too. They use the same scratch space, so both key and value will point to the same string.  
Guess I'll need to introduce a copy.

=====

Another known problem is that constant strings are sized to `STRING_SIZE` rather than `bpftrace_.strlen_`. This limits how big printf()'s format string can be.

=====

Another problem: variables are stored on-stack, so you can't assign big strings to them. Not sure whether to solve that; would it be weird for variables to become map-backed?

=====

I also still need to check whether string functions like strncmp still work.

Birch-san:
I noticed that when I originally tested the limits of BPFTRACE_STRLEN, I left a comment:

```
~1024= <A call to built-in function 'memset' is not supported.>
```

I wonder if that's why @mmarchini used probe read instead of memset to zero out the map-backed printf buffer.

If there's a precise threshold at which memset ceases to work, we could build an IRBuilder helper for zeroing buffers (e.g. memset for small buffers, probe read for big buffers).

Birch-san:
ac648a0f2f1cf256c935e6e4af12bfcd75acad43 + 52fc17d8fd34dd6c7088067e85e8682c32bc578f We can now assign values to maps with composite keys comprised of strings. There's a lot of special branches here which exist purely for performance… it'd be simpler if getMapKey() just _always_ used a map-backed buffer. And if we're still concerned with performance, we should consider the extent to which we can mitigate this by writing an LLVM optimizer pass.

e072e8f67f1c167a4a09fdefa61def654c52f4c0 Whilst looking at composite keys e.g. `@[comm, pid]`: I found that MEMCPY() works just fine to copy contents from one map to another. I revisited createFormatString and tried the same there (to move str() content into printf buffer), and that worked fine too. I can no longer reproduce the problem that made me resort to probe_read, so let's enjoy the simpler code.

Birch-san:
Things that accept strings that now accept map-backed strings:

- join()
- buf()
- ternary expressions
- strncmp()
- join()
- printf()
  - hardcoded length limit now removed
- time()
  - hardcoded length limit now removed
- associative array keys
  - composite keys too
- assignment of string to associative array
- removed hardcoded length limit from probe names…
  - …but they're backed by string literals, so see the following:

I just noticed that string literals are still stack-allocated. seems easy enough to make them map-allocated, and this should enable longer probe names (important for C++ uprobes) and longer format strings for printf.

I think the only case left is variables, and there's an argument that those don't *need* to be in this PR. I'll check what the effort is… the solution will be different to the stuff I've done so far (i.e. passing a scratch map down the stack). I'm thinking I'll use one per-CPU hash keyed on variable name, with values sized to strlen_, and clear its contents every time our probe is entered.

I'm also seeing that some care is needed to safely receive contents of scratch maps via accept(). You don't need to lifetime end them (they're reusable). But the re-use introduces a nuance: you need to make sure you've copied the contents somewhere safe (or finished using them) before the scratch map gets re-used (i.e. if you accept() multiple args that will write over the same scratch space). I've dealt with all of these cases already, but I'm wondering if there's any helper function I can introduce so that you don't need to think about the difference between receiving an AllocaInst or a scratch CallInst via accept(). think I'll look at all the cases I've done so far and see whether they have anything in common.

another option is "rather than one map shared by all str() calls: use the map key to distinguish between instances of str(), and so have a map entry per str() call". this way there'd be no re-use of map entries… so you don't need to work out "do I need to make a copy of this content". it'd increase our memory use (by n times the number of times we call any given function) and add (n times) the number of distinct map lookups (and error handling conditions). maybe LLVM can optimize that away… they have a lot of instructions in common. but the upside would be simpler, safer codegen. and we'd be avoiding copies (probe read copies, which I suspect LLVM can't elide), so it might be better for perf as well as simplicty (albeit with marginally higher memory allocation… well, same amount of allocations as when we stack-allocated… but into maps now, and potentially multiplied by number of CPUs). hm, I'm tempted to try this option.

martinetd:
(Just passing by to say thanks for the work! This has been annoying me occasionally.)

FWIW I tested this, doesn't seem to break anything I'm using (nothing very sophisticated).

There's still a size limitation (trying to set BPFTRACE_STRLEN to > 31kish (so 32k onwards)) fails but it's a very nice step forward!
If there's anything a bystander can do to help (review on #1943?) feel free to say

Birch-san:
@martinetd thanks for trying out the branch 🙂 

yeah, I think this implementation supports everything, and doesn't break anything else. for the platform I was testing on, I think I got it down to either "all tests passed" or "all tests passed, until I rebased". probably there were other-platform tests failing, which could presumably be made to pass by regenerating fixtures.

hah, I never tried a number as large as 32k. that's interesting; I _suspect_ that's a limitation of `probe_read()`, and wonder whether it could be bypassed by emitting multiple `probe_read()`s in a loop.

>  If there's anything a bystander can do to help (review on #1943?) feel free to say

Assuming "1943"=https://github.com/iovisor/bpftrace/pull/1493: I haven't done everything I could to get that over the line — i.e. I haven't implemented @fbs's suggestion of "reduce test churn by leaving the lines closer to their original order, or splitting that into a separate review".  
I agree with it (reviewing 93 bytecode fixtures is hard enough as it is, and I'm happy there are people willing to do so), but at the same time it opened my eyes to how much effort it's going to take to make this reasonably reviewable.

There are a few reasons I'm not sure how to move forward with this:

- `probe_read()` generates substantially messier bytecode than memcpy
  - i.e. because we have to verify whether it succeeded — which creates a conditional branch, whose negative case has to send an error message to userspace and end the BPF program
    - our strategy for exiting the BPF program is "return from the current function" — which relies on being the outermost function… frequently true, but I'd prefer `exit` if I knew a way to do so. [0]
    - we emit a "verify `probe_read()` outcome, and handle negative case" _per_ copy, which can mount up to **a lot**… but I wonder whether we could define it as a function and re-use it? again, this would require some way to do `exit` (returning from a function doesn't end our program).
- variables stored off-stack require allocation of maps — again, messier bytecode
  - and again, we have the same "verify whether it succeeded" messiness mentioned above
  - though, I wonder whether we could verify _all_ BPF map allocations (i.e. all bpftrace arrays, sufficiently large variables, and all strings) in one condition at the start of the program? i.e. AND all of their pointers together, and require a non-nullptr result.
  - dunno whether the BPF verifier is clever enough to understand that one statement is capable of verifying multiple variables?
- the test churn would actually be **far** more than the large amount already evident in this PR. I'm currently taking advantage of "for strings smaller than 128 bytes, we use `memcpy()`", so currently the tests don't exercise the new `probe_read()`-based Copy/ZeroInit strategy **at all** (though they **do** test the new "strings and sufficiently large variables are stored in BPF maps)
  - for full coverage, I think that requires adding an entire dimension to our test matrix (i.e. duplicate almost every existing test, but for "big string" mode)
  - tests involving `probe_read()` are harder, because you have to mock out the memory address to be deterministic
- I wonder whether there's a performance penalty for using BPF maps instead of on-stack variables, or for using `probe_read()` instead of memset/memcpy.
  - in which case, should there be a threshold over which we switch to BPF map + `probe_read()`?
    - this could make codegen and testing complicated
  - would we even go as far to say "not all strings/buffers have to be the same length"?
    - this could make codegen complicated (becomes harder to assign a string to a variable — the variable may not be big enough to hold it)
      - though, I do think there's enough information available at the time of semantic analysis to prove that a variable only ever has small constants assigned to it. otherwise you could fall back to "make it big enough for largest known string".
- LLVM's BPF backend is slow at emitting (or optimizing a program which contains) lots of 1-byte store instructions — which is how we use memset/memcpy (note: we switch to `probe_read()` strategy for strings larger than 128 bytes)
  - well, maybe it's faster in newer LLVM; I only tried LLVM 6.
  - I think LLVM _knows_ this is slow, which is why they set a "do not allow more than 128 stores in a row" limit (you can't even get around it by calling memcpy/memset multiple times).
    - by the way, if LLVM's memset/memcpy **could** be made fast (and the limit removed): then I would _strongly_ consider using that _instead of_ the `probe_read()` helper (to avoid the difficulty of having to verify whether it succeeded). that said, there's still an argument that `probe_read()` makes more sense for large values — emitting one call to a helper (albeit with lots of messy verification/error-handling) rather than emitting many, _many_ stores. dunno what the cost of "lots of stores" is — makes the BPF program longer, and more instructions for the verifier to scrutinize. but once verified: I think _executing_ many stores is probably very performant, **if** they match the CPU's native word size.
  - 8-byte store instructions (i.e. machine's native word size) would be far faster, and I expect result in a substantially more performant BPF program
    - to be honest, I actually tried this (align all `CREATE_MEMSET()` and `CREATE_MEMCPY()` to 8 bytes), and almost all tests passed. it's definitely something we would benefit from pursuing. the specific approach I'd recommend is "Copy and ZeroInit helpers should be given full information of the layout of structure that they're clearing, so that they can choose the appropriate byte alignment).
  - there are probably some `memset()`s that we _don't even need to do_. but I think we've lost the knowledge of "in what situations is memset necessary" (there was some issue on some old kernel, whose BPF verifier wasn't as clever as nowadays) https://github.com/iovisor/bpftrace/issues/1392
- we haven't defined how variables are scoped in bpftrace (i.e. block-local versus program-local). it actually turns out that they're program-local, because we hoist all Allocas to the start of the program. this creates a guarantee (was this intentional??) that no matter which condition or block you're inside: the variable has already been declared.
  - thus, in this PR: when I changed variables to be off-stack: I hoisted them such that the BPF maps for them get allocated, verified and zero-inited at the start of the program. I think though that that's not quite how it worked before (i.e. the zero-init used to be block-local? something like that). anyway, this'll have resulted in **yet more test churn**.

So… overall there's **a lot** of fronts on which I'm not sure of how to proceed. some of it is "how do we implement this" and some of it is "how do we test this".

I did begin a (non-exhaustive) checklist of how to carve up the PR:

- getMapKey() safety
- remove join type
- remove STRING_SIZE
- printf() use scratch buffer
  - "fatal" support on probe_helper
  - align_to?
  - probe_read use Value* ?
- remove conditions in join() codegen
  - moved into CreateGetJoinMap()'s fatal probe_helper
- str() use scratch buffer
- buf() use scratch buffer
  - 32-bit size
  - align_to?
  - printers
- variables
- plenty more stuff, probably

But that checklist was "how to get it in in its _current_ state", and doesn't address "should we have implemented/tested it like this".

[0] I think `exit` _does_ exist in BPF (but not in the LLVM BPF backend). I saw that Alessandro Decina's succeeded in doing it via inline ASM. I had a stab at doing the same; here's what I tried, but LLVM told me it wasn't supported:

```cpp
#include <llvm/IR/InlineAsm.h>
CallInst   *CreateExit();
/*
 * The LLVM BPF target does not support core::intrinsics::abort().
 * Thanks to Alessandro Decina for explaining how to emit an exit instruction
 * https://blog.redsift.com/labs/oxidised-ebpf-ii-taming-llvm/
 * https://github.com/redsift/redbpf/blob/320e3d1/cargo-bpf/src/llvm.rs#L83
 */
CallInst* IRBuilderBPF::CreateExit() {
  FunctionType *func_ty = FunctionType::get(getVoidTy(), false);
  InlineAsm *inlineAsm = InlineAsm::get(func_ty, "exit", {}, false);
  CallInst *call = CreateCall(inlineAsm);
  call->addAttribute(AttributeList::FunctionIndex, Attribute::NoUnwind);
  // call->addAttribute(AttributeList::FunctionIndex, Attribute::NoInline);
  // call->addAttribute(AttributeList::FunctionIndex, Attribute::ReadNone);
  call->addAttribute(AttributeList::FunctionIndex, Attribute::NoReturn);
  return call;
}

// and… maybe we have to add this to the CMake config? I tried that too though.
add_compile_options("-D__ASM_SYSREG_H")
```

martinetd:
wow. Thanks for the very detailled answer, I didn't expect quite that much!

I don't know if that helps, but the reason I tried >32k was for a uprobe actually -- in there it's frequent enough to pass large arrays in arguments and I wanted to dump SSL read/writes as described in https://confused.ai/posts/intercepting-zoom-tls-encryption-bpf-uprobes (for which they implemented https://github.com/alessandrod/snuffy based on https://github.com/redsift/redbpf -- I boasted that was overkill and I could do the same in a few minutes with bpftrace and well that wasn't quite so easy due to the size limitation!)
honestly at this point performance isn't so critical, but I assume the similar could be done for read/write kernel calls if required :P

I'm not too familiar with it but the magic seems to be in https://github.com/alessandrod/snuffy/blob/master/snuffy-probes/src/snuffy/main.rs#L254 where they iterate `BUF_LEN` (368 bytes) at a time through `bpf_probe_read`+corresponding event -- pretty close to what you were first suggesting; they don't care about checking everytime to break the loop ; maybe a helper like they have could help keep code sane-ish.
performance-wise it would obviously need testing but intuitively `probe_read` should be faster than byte per byte copy on large buffers, I honestly don't know... :/


Well, that was more of a play thing than an actual need, so let's put >32k aside for now if that helps.

Regarding the question, would printf("%s", str()) be good enough for me? Probably, I'm not sure.
For that particular example I gave, I'd probably want to print it a bit more nicely (e.g. hexdump or something, people send weird things over tls!), but that would probably be more practical to postprocess than print in bpftrace itself? So just format <length><string><length><string> and postprocess would work I guess.
However the most common usecase I'd actually have that often requires more than 200 is just dumping paths (think opensnoop for example); 200 is too short with a deep filesystem like I often have to deal with, I'd need at least 500 and if possible 4k. filenames can have linefeeds so just printing "is bad" but in the context of debugging it's ok for me.
I was thinking that wouldn't work because I usually want to print at sys_exit the arg I got at sys_enter, but for things like openat it's fine storing the address of the buffer at enter and not the string, so it works ok... There might be syscalls where I'd want to store the actual data on enter like write?
Well, it's safe to say it would cover most use cases for sure anyway.


All in all if you don't know I don't either :D Please do what you think is best within the time you have ;) and thanks again!

Birch-san:
> detailled answer

ah, that was something long overdue — I'd left it to languish, so wanted to at least document the blockers.

> do the same in a few minutes with bpftrace

🙂 always the dream. seems like a perfect use-case to try it on.

> intuitively probe_read should be faster than byte per byte copy on large buffers

Aye, though I suspect the lesson here is "it'd be really nice to copy word-by-word instead".

> I'd probably want to print it a bit more nicely (e.g. hexdump or something)

Okay, so probably `printf("Datagram bytes: %r\n", buf(args->buff, args->len))` is more your speed — that's fine; it works very similar to `str()`, so still achievable.

> However the most common usecase I'd actually have that often requires more than 200 is just dumping paths (think opensnoop for example)

If you just want to _print_ them, then this `printf("%s", str())` proposal should be able to do that without storing anything on stack.

But if you need to store it in a variable _on-stack_, do some calculations and then `printf()` something afterwards… well, at least moving the `printf()` buffer off-stack gives you _a bit more room_ to play with that string on-stack. 500~4k would need off-stack variables at that point.

martinetd:
Just a note that with bpf ring buffers now available that might be easier to use than per cpu scratch buffers, albeit a bit less performant perhaps? It doesn't remove the need to check for potential failures and bail out though so that might not actually help at all for the problems you listed :/

For my own self-centered needs, I was able to straight-forwardly update bcc's libbpf-tools/opensnoop program to handle paths > 254 that way, since ring buffers can be used as a mostly drop in replacement. It does require a recent kernel and is far from as flexible as bpftrace but I'll try to waddle through like that for now when the need occasionally arises.

Birch-san:
interesting! presumably that eliminates the need to allocate maps? unfortunately that's not the hard part 😅 

yeah, the hard part is the validation/bailout for acquiring a map. as you say: we'd still have that problem.

and I think we'd still have the problem of "now that you have this big storage: how do you do big copies/clears onto it?" presumably that'd still be "do a bpf_probe_read, with validation/bailout".

and yeah, as you say: it's new-kernel only. if we wanted to enable it conditionally, that adds a condition to the codegen and adds another dimension to the test matrix.

so, glad to know about it. but my initial reaction is that the hardest problems are still there. 😅

Birch-san:
@mmarchini hey, good to see you back. 🙂 I'm not actively developing Big Strings. still interested.

what approach are you thinking of taking? there's a spectrum of how far it can be taken.

- [map-backed scratch buffer](https://github.com/iovisor/bpftrace/pull/1801) for formatting strings (e.g. `printf`)
- map-backed `string` datatype (affects every expression which consumes strings)
  - here's [most of them](https://github.com/iovisor/bpftrace/pull/1360#issuecomment-647198245)
- map-backed `buffer` datatype
- map-backed variables
- implement copying via `probe_read()`

this PR demonstrates a working implementation of all of those.

hard parts:

- all this map validation adds a lot of blocks to the IR
  - lots of churn in the test fixtures
  - hard to read, maintain or review
- using `probe_read()` to do large `memset`/`memcpy` creates challenges for tests
  - the address changes on each run, so you need mocking or suchlike to overlook that
  - I cheated here; only utilised `probe_read()` for _large_ strings (the tests use small strings)

some ideas:

- there may be alternatives to `probe_read()` for large strings ([>1014 bytes](https://github.com/iovisor/bpftrace/blob/17679cee98ace80c9a19765c0a8250683b6bee57/src/bpftrace.h#L192-L201))
  - [emit many store instructions](https://github.com/iovisor/bpftrace/discussions/1800#discussioncomment-830253) (I haven't tried this, and don't know whether LLVM tries to fuse them into a memset)
  - there is actually an LLVM option [`-bpf-expand-memcpy-in-order`](https://github.com/iovisor/bpftrace/discussions/1800) to turn off the "max stores" limitation, but I don't know how to enable it (or whether it's exposed at all via programmatic API).
    - it may be disabled for a good reason (e.g. causing some optimization pass to go too slow)
- we may need a solution for "how to keep our IR simple when we have many maps to validate"
  - [proposal](https://github.com/iovisor/bpftrace/discussions/1809) of how to lay out our blocks
    - [sub-proposal](https://github.com/iovisor/bpftrace/discussions/1808) (stop emitting allocas in reverse order)
  - is there a way to make re-usable "map validation failure" handling?
    - "send perf event explaining the error" is a lot of IR, and we do this **per** map.
    - I did try (in a branch somewhere) defining this sequence as a function, but LLVM inlined it to each call-site so I got duplicate IR regardless.
    - maybe another option is to `&&` every map validation together, so we only need to handle one failure for `n` maps?
      - you'd get poorer error feedback (i.e. which map failed, and which AST node is to blame), but better IR
      - only possible if the BPF verifier is smart enough
- once you have big strings, you get a new problem: tiny loads/stores.
  - with (our default) 1-byte-aligned accesses: a bpftrace program where "all strings are big" takes _super_ long to compile (tens of seconds)
  - with 8-byte-aligned accesses: the same program can compile sub-second.
    - not sure whether this was because it's simpler to compile once word-aligned, or instead because there were fewer instructions to run the optimizer on
    - pretty much nothing broke when I set everything to 8-byte-aligned (maybe 1 runtime test? maybe an alignment-specific one?)
    - LLVM can compute a common alignment given a source and destination address. in this PR, I refactored a few places where we weren't yet using `StructTy` or were using vaguer types than we needed to. I think it's achievable to give our `memset`/`memcpy` calls enough information to pick an optimal alignment. this would be a good general improvement for bpftrace, and I think isn't too hard to implement — just not sure how to test.

another consideration is the [proposal of a non-LLVM backend](https://github.com/iovisor/bpftrace/issues/1845) (which I assume is what you were getting at with "[when installing llvm is unfeasible](https://github.com/iovisor/bpftrace/issues/342)"). if there's any part of shipping this feature that's made harder _because of LLVM_ (e.g. the fixture-based testing), then we do have the option of supporting it via another backend.

danobi:
This work is now ongoing in https://github.com/iovisor/bpftrace/pull/2746 . There's been a lot of great activity in this PR, but due to age I'm closing it out.

PR #1365, opened 2020-06-01
Fix calling ntop on tracepoint args

danobi:
Commit f05b4cda691e82ab ("Introduce Type::ctx") introduced a `ctx` type
for values that come from the bpf prog context. This subtly broke ntop()
because ntop was not yet aware of Type::ctx.

Fix this by relaxing semantic analyser check to include Type::ctx.
Similarly relax for codegen.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,8 @@ and this project adheres to
   - [#1341](https://github.com/iovisor/bpftrace/pull/1341)
 - Fix `KBUILD_MODNAME`
   - [#1352](https://github.com/iovisor/bpftrace/pull/1352)
+- Fix `ntop()` not accepting tracepoint arguments
+  - [#1365](https://github.com/iovisor/bpftrace/pull/1365)
 
 #### Tools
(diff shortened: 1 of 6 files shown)


danobi:
I wonder if Type::ctx should be limited to just the `ctx` builtin. Reasoning is that a value coming from the prog context is more an attribute and not a type unto itself. For example, `saddr_v6` can be _both_ an array and from the prog ctx. It doesn't necessarily have to be one or the other.

@mmisono any thoughts?

fbs:
> I wonder if Type::ctx should be limited to just the `ctx` builtin. Reasoning is that a value coming from the prog context is more an attribute and not a type unto itself.

I'm working on replacing cast with a pointer and a record type. As ctx can currently be both and more which makes it confusing I'm going to remove ctx too and make it an attribute instead. 

https://github.com/fbs/bpftrace/tree/pointer_type

danobi:
Sounds good. Looking forward to seeing that merged :)

PR #1370, opened 2020-06-05
usdt: Decrement usdt semaphore value during probe detach

danobi:
Before, I think we were intentionally leaking the bcc usdt context so
that usdt semaphore values would not go back to zero. Reason is that
bcc_usdt_close() decrements usdt semaphore value.

However, this causes issues if we don't eventually decrement the counter
as the tracee will forever be on more expensive code paths.

This commit defers semaphore decrement until probe is detached.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@ and this project adheres to
   - [#1365](https://github.com/iovisor/bpftrace/pull/1365)
 - Fix attaching to usdt probes in multiple binaries
   - [#1356](https://github.com/iovisor/bpftrace/pull/1356)
+- Decrement usdt semaphore count after bpftrace execution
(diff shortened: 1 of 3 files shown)


dalehamel:
LGTM, as we discussed on IRC, keeping track of the semaphore in bpftrace is asking for trouble if bpftrace crashes. The kernel API i'm working on wrapping in https://github.com/iovisor/bcc/pull/2738 i think avoids this, but obviously that needs to be finished first, and we'll need to support both approaches for a while, to maintain support for older kernels.

danobi:
Yeah, kernel support would be optimal. For correctness, I think bpftrace still needs to support both approaches until we're pretty sure everyone is running a new-enough kernel.

PR #1374, opened 2020-06-05
usdt: Use new usdt_addsem_probe* helpers

danobi:
While debugging a high memory consumption issue, I noticed that a
USDT::Context object can take ~10M per instance [0]. Along with the new
--usdt-file-activation, bpftrace can potentially hold onto many dozens
of USDT:Context instances, causing memory issues.

This patch makes use of the new bcc usdt semaphore helpers (
iovisor/bcc#2953 ) to avoid holding onto
USDT::Context instances for the lifetime of the tracing session.

The new algorithm is:

1. create a USDT::Context
2. increment the semaphore count for the probe we care about
3. destroy the USDT::Context
4. repeat 1-3 for all probes we want to attach to
5. do our tracing
6. create a USDT::Context for the probe we care about
7. decrement the semaphore count
8. destroy the USDT::Context
9. repeat 6-8 for all the probes we're attached to

This approach also has the benefit of 1 USDT::Context instance being
alive at a time which can help keep memory high watermark low.

[0]: Through gdb single stepping and /proc/pid/status. Exact process is
not described here b/c memory usage probably varies based on tracee
binary.

##### Checklist


danobi:
Relies on #1370 .

danobi:
Ready for review

src/attached_probe.cpp
```diff
@@ -796,13 +796,44 @@ void AttachedProbe::attach_usdt(int pid)
   struct bcc_usdt_location loc = {};
   int err;
   void *ctx;
+  // TODO: fn_name may need a unique suffix for each attachment on the same
```
fbs, line 799: does this need to be tested?

fbs: oh nvm, the line just got moved up.

danobi:
> Should we add a tracking ticket for CI testing this?

I think this'll get testing when the new bcc version is released and the CI starts building against the new bcc api. I'm not too sure it'd be worth it to set up a new CI step just to build/test custom versions of bcc.

Issue #1335, opened 2020-05-15
`interval` and `profile` probes should support same time units


danobi:
### Is your feature request related to a problem? Please describe.
If you look at `interval` and `profile` probe in [reference guide](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#9-profile-timed-sampling-events), you'll notice that profile has `hz` and `us` and interval does not.

### Describe the solution you'd like
Support for `hz` and `us` in interval probe.

### Describe alternative solutions or features you've considered
n/a

PR #1377, opened 2020-06-10
Add time units for `interval` probe

ember526:
This patch adds support for new time units `us` and `hz` for probe `interval`.

Such changes make the time units of `interval` consistent with those of `profile`. To address [#issue1335](https://github.com/iovisor/bpftrace/issues/1335)

##### Checklist


fbs:
I think it should be doable to build some "footgun" detection into the semantic analyser, like unconditional `print()`ing in `us` and maybe even `ms` interval /profile probes.

danobi:
> @danobi do you think it would be worth it to add some runtime tests here?

Meh, might be more trouble than it's worth. Cuz the test framework doesn't do multi line matches IIRC and then you'd need to print stuff on a single line and that might get flakey with output buffering.

PR #1380, opened 2020-06-12
Disable reg() for kfunc

mmisono:
"reg()" is not usable in kfunc since its context data does not contain
struct pt_regs. Disable it in the semantic analyzer.

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -670,8 +670,11 @@ void SemanticAnalyser::visit(Call &call)
     if (check_nargs(call, 1)) {
       for (auto &attach_point : *probe_->attach_points) {
         ProbeType type = probetype(attach_point->provider);
-        if (type == ProbeType::tracepoint) {
-          error("The reg function cannot be used with 'tracepoint' probes",
+        if (type == ProbeType::tracepoint || type == ProbeType::kfunc ||
+            type == ProbeType::kretfunc)
+        {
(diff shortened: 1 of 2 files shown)


mmisono:
This also applies to the lsm probe. I'll update this PR after #1347 is merged.

src/ast/semantic_analyser.cpp
```diff
@@ -670,8 +670,11 @@ void SemanticAnalyser::visit(Call &call)
     if (check_nargs(call, 1)) {
       for (auto &attach_point : *probe_->attach_points) {
         ProbeType type = probetype(attach_point->provider);
-        if (type == ProbeType::tracepoint) {
-          error("The reg function cannot be used with 'tracepoint' probes",
+        if (type == ProbeType::tracepoint || type == ProbeType::kfunc ||
```
javierhonduco, line 673: Should we gate for probe types that we know this should work instead of the ones we know it will not work?

danobi:
LGTM. Added do-not-merge tag to indicate it's blocked on LSM PR

fbs:
@mmisono dont think lsm is ready any time soon, shall we just add this in?

mmisono:
Yes, I'll work on over the weekend.

src/ast/semantic_analyser.cpp
mmisono: I agree with you, will try that.

mmisono:
I created a #1646 for continuation.

Issue #1332, opened 2020-05-15
tcpconnect.bt prints weird destination port


rhdxmr:
Hi,

I am using Pop!_OS 20.04
```
03:05:52 15 ❯  lsb_release  -a
No LSB modules are available.
Distributor ID: Pop
Description:    Pop!_OS 20.04 LTS
Release:        20.04
Codename:       focal
```

kernel version is 5.4.0

```
03:08:54 15 ❯  uname -a
Linux pop-os 5.4.0-7626-generic #30~1588169883~20.04~bbe668a-Ubuntu SMP Wed Apr 29 21:00:02 UTC  x86_64 x86_64 x86_64 GNU/Linux
```

I installed bpftrace by apt.

```
sudo apt install bpftrace
```

But tcpconnect.bt prints weird destination port.

At a terminal, run curl to get google.com:80

And at another terminal, run tcpconnect.bt
```

03:04:10 15 ❯  /usr/bin/bpftrace /sbin/tcpconnect.bt
Attaching 2 probes...
Tracing tcp connections. Hit Ctrl-C to end.
TIME     PID      COMM             SADDR                                   SPORT  DADDR                                   DPORT
15:04:13 980915   curl             192.168.0.9                             60490  172.217.175.14                          13429
^C
```

DPORT shows 13429, not 80.

bpftrace depends on LLVM 9.
Maybe is it a reason of this problem?

```
03:04:15 15 ❯  apt depends bpftrace
bpftrace
  Depends: libbpfcc (>= 0.12.0)
  Depends: libc6 (>= 2.27)
  Depends: libclang1-9 (>= 1:9~svn359771-1~)
  Depends: libgcc-s1 (>= 3.0)
  Depends: libllvm9 (>= 1:9~svn298832-1~)

```

INSTALL.md suggests to use LLVM 7.

So I built bpftrace from source with LLVM 7. And tcpconnect.bt works as expected. yay

```
03:11:44 15 ❯  ldd /home/esrse/.local/bin/bpftrace
        linux-vdso.so.1 (0x00007ffd2757f000)
        libbfd-2.34-system.so => /lib/x86_64-linux-gnu/libbfd-2.34-system.so (0x00007f661a382000)
        libopcodes-2.34-system.so => /lib/x86_64-linux-gnu/libopcodes-2.34-system.so (0x00007f661a224000)
        libbcc.so.0 => /lib/x86_64-linux-gnu/libbcc.so.0 (0x00007f6616878000)
[...]

PR #1381, opened 2020-06-12
Support non-map print()

### LINKED ISSUE #356: print() for non maps
#56 may extended printf() to support maps as well: a print formatted for maps.

This issue is to extend print() to support non maps as well: a print without formatting for all types. We'd then have:
- printf() to print anything with a format string
- print() to print anything in a default way

Eg, something like this should work:

```
# bpftrace -e 'k:do_nanosleep { print(comm); }'
Attaching 1 probe...
bash
sleep
sleep
```

Instead of always having to do:

```
# bpftrace -e 'k:do_nanosleep { printf("%s\n", comm); }'
```

danobi:
This implements non-map prints as described in #356.

Note that non-map `print()`s are synchronous, meaning they do not
work the same as map `print()`s. I think this is a pretty nice feature.

Example:
```
# bpftrace -e 'BEGIN { $t = (1, "string"); print(123); print($t); print(comm) }'
Attaching 1 probe...
123
(1, string)
bpftrace
^C
```
We have a growing number of types in the type system so I may have
missed a corner case somewhere. I think I got most of it right, though.

This closes #356.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ and this project adheres to
   - [#1367](https://github.com/iovisor/bpftrace/pull/1367)
 - Add support for time units `us` and `hz` for probe `interval`
   - [#1377](https://github.com/iovisor/bpftrace/pull/1377)
+- Add support for non-map print()
+  - [#1381](https://github.com/iovisor/bpftrace/pull/1381)
 
 #### Changed
 
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -88,6 +88,7 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [20. `override()`: Override return value](#20-override-override-return-value)
     - [21. `buf()`: Buffers](#21-buf-buffers)
     - [22. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
+    - [23. `print()`: Print Value](#23-print-print-value)
 - [Map Functions](#map-functions)
     - [1. Builtins](#1-builtins-2)
     - [2. `count()`: Count](#2-count-count)
@@ -2656,6 +2657,31 @@ Attaching 1 probe...
 8
 ```
 
+## 23. `print()`: Print Value
+
+Syntax: ```print(value)```
+
+The `print()` function can print a non-map value with default formatting.
+
+For example, local variables and most builtins can be printed:
+
+```
+# bpftrace -e 'BEGIN { $t = (1, "string"); print(123); print($t); print(comm) }'
+Attaching 1 probe...
+123
+(1, string)
+bpftrace
+^C
+```
+
+It is important to note that printing values is different than printing maps.
+Both printing maps and printing values are asynchronous: the kernel queues the
+event but some time later it is processed in userspace. For values, the event
+contains the memcopy'd value so the value at `print()` invocation time will be
+printed.  However for maps, only the handle to the map is queued up, so the
+printed map may be different than the map at `print()` invocation.
+
+
 # Map Functions
 
 Maps are special BPF data types that can be used to store counts, statistics, and histograms. They are
@@ -2700,11 +2726,12 @@ END
 - `lhist(int n, int min, int max, int step)` - Produce a linear histogram of values of n
 - `delete(@x[key])` - Delete the map element passed in as an argument
 - `print(@x[, top [, div]])` - Print the map, optionally the top entries only and with a divisor
+- `print(value)` - Print a value
 - `clear(@x)` - Delete all keys from the map
 - `zero(@x)` - Set all map values to zero
 
 Some of these are asynchronous: the kernel queues the event, but some time later (milliseconds) it is
-processed in user-space. The asynchronous actions are: `print()`, `clear()`, and `zero()`.
+processed in user-space. The asynchronous actions are: `print()` on maps, `clear()`, and `zero()`.
 
 ## 2. `count()`: Count
 
@@ -3000,6 +3027,9 @@ This one-liner sums the vfs_read() durations as nanoseconds, and then does the d
 when printing. Without this capability, should one try to divide to milliseconds when summing (eg,
 `sum((nsecs - @start[tid]) / 1000000)`), the value would often be rounded to zero, and not accumulate as
 it should.
+
+Note that printing maps is different than printing values. See the explanation
+in [`print()`: Print Value](#23-print-print-value).
 
 # Output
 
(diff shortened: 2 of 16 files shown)


mmisono:
Looks very useful. I found some issues:

- `comm` in a tuple is not correct.
```
% sudo ./src/bpftrace -e 'BEGIN { print(comm); print((1,comm)); }'
Attaching 1 probe...
bpftrace
(1, 7u)
^C
```

Consulting the generated IR, it seems this is the same bug as #1332. The buffer holding the result of `bpf_get_current_comm` is overwritten before use. (The async id of print is 30007 = 0x7537 = 'u' and '7'.) This bug is really annoying..

- 

```
$ sudo ./src/bpftrace -e 'BEGIN { print(exit()); }'
AddressSanitizer:DEADLYSIGNAL
==20081==ERROR: AddressSanitizer: SEGV on unknown address 0x000000004e71 (pc 0x7f39655c2080 bp 0x7ffc781c73b0 sp 0x7ffc781c72c8 T0)
==20081==The signal is caused by a READ memory access.
    #0 0x7f39655c2080 in llvm::Value::getContext() const (/usr/lib/x86_64-linux-gnu/libLLVM-10.so.1+0xac1080)
    #1 0x7f3965563c8f in llvm::StoreInst::StoreInst(llvm::Value*, llvm::Value*, bool, llvm::MaybeAlign, llvm::AtomicOrdering, unsigned char, llvm::Instruction*) (/usr/lib/x86_64-linux-gnu/libLLVM-10.so.1+0xa62c8f)
    #2 0x7f3965563bc7 in llvm::StoreInst::StoreInst(llvm::Value*, llvm::Value*, bool, llvm::Instruction*) (/usr/lib/x86_64-linux-gnu/libLLVM-10.so.1+0xa62bc7)
    #3 0x7ae58f in llvm::IRBuilder<llvm::ConstantFolder, llvm::IRBuilderDefaultInserter>::CreateStore(llvm::Value*, llvm::Value*, bool) /usr/lib/llvm-10/include/llvm/IR/IRBuilder.h:1699:23
    #4 0x7a20b7 in bpftrace::ast::CodegenLLVM::createPrintNonMapCall(bpftrace::ast::Call&, int&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:2375:8
    #5 0x79d618 in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Call&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:777:7
    #6 0x78ea90 in bpftrace::ast::Call::accept(bpftrace::ast::Visitor&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/ast.cpp:131:5
    #7 0x7a858a in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::ExprStatement&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:1489:14
    #8 0x78f9a3 in bpftrace::ast::ExprStatement::accept(bpftrace::ast::Visitor&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/ast.cpp:291:5
    #9 0x7aa0bf in bpftrace::ast::CodegenLLVM::generateProbe(bpftrace::ast::Probe&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, llvm::FunctionType*, bool) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:1776:11
    #10 0x7aa4ba in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Probe&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:1807:5
    #11 0x790133 in bpftrace::ast::Probe::accept(bpftrace::ast::Visitor&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/ast.cpp:383:5
    #12 0x7aafab in bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Program&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:1910:12
    #13 0x7901e3 in bpftrace::ast::Program::accept(bpftrace::ast::Visitor&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/ast.cpp:392:5
    #14 0x7ad528 in bpftrace::ast::CodegenLLVM::compile(bpftrace::DebugLevel, std::ostream&) /home/ubuntu/work/bpftrace/bpftrace/src/ast/codegen_llvm.cpp:2409:10
    #15 0x710f8d in main /home/ubuntu/work/bpftrace/bpftrace/src/main.cpp:650:19
    #16 0x7f39617e1b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310
    #17 0x4830c9 in _start (/home/ubuntu/work/bpftrace/bpftrace/build_dev/src/bpftrace+0x4830c9)

```

fbs:
Looks good, the synchronous part got me a bit confused tho, I think it needs some clarification to not confuse users. 
The whole thing is still async in the sense that the bpf program (likely) terminates before the event handles, all events are "queued" waiting for userspace. The difference between map and this is that with maps we send a mapid (pointer) to be printed, which means the value could have changed before the printing, while for values it sends the value.

> Consulting the generated IR, it seems this is the same bug as #1332. The buffer holding the result of `bpf_get_current_comm` is overwritten before use. (The async id of print is 30007 = 0x7537 = 'u' and '7'.) This bug is really annoying..

Ugh this seems to bite us everywhere now. I need to get it finshed

danobi:
> Consulting the generated IR, it seems this is the same bug as #1332. 

Ok, will leave that alone for now. I'll try to take a look if I get time.

> $ sudo ./src/bpftrace -e 'BEGIN { print(exit()); }'
AddressSanitizer:DEADLYSIGNAL

Fixed. Forgot to exclude `Type::none` from printable types.

> Looks good, the synchronous part got me a bit confused tho, I think it needs some clarification to not confuse users.

Updated. Now reads:
```
It is important to note that printing values is different than printing maps.
Both printing maps and printing values are asynchronous: the kernel queues the
event but some time later it is processed in userspace. For values, the event
contains the memcopy'd value so the value at `print()` invocation time will be
printed.  However for maps, only the handle to the map is queued up, so the
printed map may be different than the map at `print()` invocation.
```

fbs:
LGTM, should we squash this or leave it as is?

danobi:
I'll squash; commits only seem useful for code review

PR #1386, opened 2020-06-16
codegen: Introduce safer accept()

danobi:
While looking at codegen, I noticed that it's quite easy to forget to
end the lifetime of an alloca buffer. This PR proposes an interface that
should make it harder to forget ending lifetimes.

This patchset is best viewed one commit at a time in the order I've
laid them out.

Commits 1-3 should be fairly straightforward. The final commit,
"Update all callsites to use safer accept()" was quite tricky to write
so I'm not sure if I've fully considered all the cases. The final commit
should warrant some serious scrutiny, as marking a lifetime longer
than actually is doesn't result in logical bugs -- only higher memory
use. However, marking a lifetime shorter than it really is can cause
corruption.

Fortunately, accidentally shortening lifetimes is pretty rare in our
codebase. It's pretty much only used in FieldAccess.

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -241,7 +241,11 @@ void CodegenLLVM::visit(Builtin &builtin)
     dyn_cast<LoadInst>(expr_)->setVolatile(true);
 
     if (builtin.type.IsUsymTy())
+    {
       expr_ = b_.CreateUSym(expr_);
+      Value *expr = expr_;
+      expr_deleter_ = [this, expr]() { b_.CreateLifetimeEnd(expr); };
+    }
   }
   else if (!builtin.ident.compare(0, 4, "sarg") && builtin.ident.size() == 5 &&
       builtin.ident.at(4) >= '0' && builtin.ident.at(4) <= '9')
@@ -326,7 +330,7 @@ void CodegenLLVM::visit(Call &call)
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
 
-    call.vargs->front()->accept(*this);
+    auto scoped_del = accept(call.vargs->front());
     // promote int to 64-bit
     expr_ = b_.CreateIntCast(expr_,
                              b_.getInt64Ty(),
@@ -349,7 +353,7 @@ void CodegenLLVM::visit(Call &call)
     // Store the max of (0xffffffff - val), so that our SGE comparison with uninitialized
     // elements will always store on the first occurrence. Revent this later when printing.
     Function *parent = b_.GetInsertBlock()->getParent();
-    call.vargs->front()->accept(*this);
+    auto scoped_del = accept(call.vargs->front());
     // promote int to 64-bit
     expr_ = b_.CreateIntCast(expr_,
                              b_.getInt64Ty(),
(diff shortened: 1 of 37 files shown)


danobi:
IMO this is kind of a risky PR but if we can manage to prevent regressions I think it should make things less painful long term.

Birch-san:
I think this will help. It's not strict visitor pattern any more… but using the return value of `::accept()` is the only way to give the deleter the same lifetime as the block it's returned into.

I wonder whether to tie expr_ and expr_deleter_ closer together (i.e. so that they're guaranteed to be talking about the same thing)?

```c++
std::tuple<Value*, CodegenLLVM::ExprDeleter> CodegenLLVM::accept(Node *node);

auto [expr, del] = accept(call.vargs->front());
```

Although the counterargument to this is, "we have plenty of member variables on the visitor; why is expr_ so special that it deserves to be returned by `::accept()`?" (and I can't make the same compelling argument about variable lifetime there).

Birch-san:
ah… I guess one area liable to cause mistakes is that ExprDeleter will get triggered at the end of the current C++ block, not at the end of the current LLVM IR block…

```c++
Value *perfdata = b_.CreateSomeKindOfMap(ctx_, call.loc);
Function *parent = b_.GetInsertBlock()->getParent();

BasicBlock *zero = BasicBlock::Create(module_->getContext(),
                                      "zero",
                                      parent);
BasicBlock *notzero = BasicBlock::Create(module_->getContext(),
                                          "notzero",
                                          parent);

b_.CreateCondBr(b_.CreateICmpNE(perfdata,
                                ConstantExpr::getCast(Instruction::IntToPtr,
                                                      b_.getInt64(0),
                                                      b_.getInt8PtrTy()),
                                "zerocond"),
                notzero,
                zero);
b_.SetInsertPoint(notzero);
// we'd have to be careful to remember to set up a block scope here…
{
  auto del = accept(call.vargs->front());
  // do stuff with expr_…
  
  // we've reached the end of the C++ block, so del's ~ExprDeleter will emit a lifetime end
}
b_.CreateBr(zero);
b_.SetInsertPoint(zero);

**Comment by danobi** [comment]
About to go off the grid for the rest of the week so will respond next week.

BTW @Birch-san, not sure if you're in the IRC but a bunch of us are in the oftc/#bpftrace IRC/discord chat. Come visit :)

**Comment by danobi** [comment]
>  // if we don't set up the block-scope, then del's ~ExprDeleter will emit a lifetime end here instead
// which won't make sense in the "go straight to InsertPoint zero" case (ending lifetime of a variable not used in our conditional branch)

That's a good point. Hmm, I'm kind of torn on this PR now.

On the one hand, it's annoying always adding a
```
if (expr_deleter_)
  expr_deleter_()
```
after every accept. However, it's not dangerous to forget to add it.

On the other hand, properly scoping the deleter so we're in the right basic block is also easy to forget and can lead to danger.

@mmisono , @fbs , do you have any opinions?

**Comment by Birch-san** [comment]
Maybe this is just a branding problem? What if we rename the class to `ScopedExprDeleter`, and set a good precedent for variable naming:

```c++
auto scoped_del = accept(call.vargs->front());
```

People are familiar with scoped pointers and scoped mutexes, so maybe this will evoke that kind of feeling?

danobi:
Updated to used "Scoped" as much as possible to remind people it's scoped.

I'm feeling pretty good about this now. I think it's worth a shot to see how it turns out over time.

fbs:
Gotta find the time/energy to stare at that codegen output for a bit to make sure it makes sense but at first glance it looks good :)

fbs:
Code looks ok. I'll take a look at the codegen once it has been rebased onto master, with the codegen now happening before opt  this might be a bit easier to look at.

danobi:
Rebased

PR #1387, opened 2020-06-16
Add support for strftime() builtin

### LINKED ISSUE #1337: Add a way for users to get current time

`time()` printing a string to terminal is fine for some use cases but it's more useful in others to be able to `printf()` a timestamp.

Making `time()` return a string would be API breaking and also cause issues with `strncmp` (because we cannot resolve the timestamp string in kernel) so we probably don't wanna change that.

Perhaps we can make a `epoch` builtin that returns seconds since epoch? I think seconds as granularity should be enough b/c high precision timing could just use `nsecs` instead.

`epoch` builtin to return seconds since epoch.

We could insert the epoch time at system boot in a map during prog load and then use `bpf_ktime_get_ns()` helper to compute seconds since epoch.

Making `timestamp()` return a string

suyuee:
This PR introduces a new builtin function `strftime`. This function makes it possible to print the current time with `printf`:

```
bpftrace -e 'i:s:1 { printf("%s\n", strftime("%H:%M:%S", nsecs)); }'
```

Although `time` can print current time on its own, printing timestamp with `printf` can produce more concise and more readable output. Resolves #1337 .
##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to
 ## Unreleased
 
 #### Added
+- Add builtin function: `strftime`
+  - [#1387](https://github.com/iovisor/bpftrace/pull/1387)
 - Fix `printf` not allowing format specifiers to be directly followed by
   alphabetic characters
   - [#1414](https://github.com/iovisor/bpftrace/pull/1414)
@@ -72,6 +74,7 @@ and this project adheres to
   - [#1404](https://github.com/iovisor/bpftrace/pull/1404)
 - Enable the `ternary` operator to evaluate builtin calls
   - [#1405](https://github.com/iovisor/bpftrace/pull/1405)
+
 
 #### Changed
 
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -89,6 +89,7 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [21. `buf()`: Buffers](#21-buf-buffers)
     - [22. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
     - [23. `print()`: Print Value](#23-print-print-value)
+    - [24. `strftime()`: Formatted timestamp](#24-strftime-formatted-timestamp)
 - [Map Functions](#map-functions)
     - [1. Builtins](#1-builtins-2)
     - [2. `count()`: Count](#2-count-count)
@@ -2701,6 +2702,28 @@ contains the memcopy'd value so the value at `print()` invocation time will be
 printed.  However for maps, only the handle to the map is queued up, so the
 printed map may be different than the map at `print()` invocation.
 
+## 24. `strftime()`: Formatted timestamp
+
+Syntax:
+- `strftime(const char *format, int nsecs)`
+
+This returns a formatted timestamp that is printable with `printf`. The format
+string must be supported by `strftime(3)`. Use format specifier "%s" when
+printing the return value. Note that `strftime` does not actually return a
+string in bpf (kernel), the formatting happens in userspace.
+
+Examples:
+
+```
+# bpftrace -e 'i:s:1 { printf("%s\n", strftime("%H:%M:%S", nsecs)); }'
+Attaching 1 probe...
+13:11:22
+13:11:23
+13:11:24
+13:11:25
+13:11:26
+^C
+```
 
 # Map Functions
 
--- a/src/ast/async_event_types.h
+++ b/src/ast/async_event_types.h
@@ -76,6 +76,20 @@ struct Time
   }
 } __attribute__((packed));
 
+struct Strftime
+{
+  uint64_t strftime_id;
+  uint64_t nsecs_since_boot;
+
+  std::vector<llvm::Type*> asLLVMType(ast::IRBuilderBPF& b)
+  {
+    return {
+      b.getInt64Ty(), // strftime id
+      b.getInt64Ty(), // strftime arg, time elapsed since boot
+    };
+  }
+} __attribute__((packed));
+
 struct Buf
 {
   uint8_t length;
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -3,6 +3,7 @@
 #include "ast.h"
 #include "ast/async_event_types.h"
 #include "bpforc.h"
+#include "codegen_helper.h"
 #include "parser.tab.hh"
 #include "tracepoint_format_parser.h"
 #include "types.h"
@@ -21,14 +22,6 @@
 
 namespace bpftrace {
 namespace ast {
-
-namespace {
-bool shouldBeOnStackAlready(SizedType &type)
-{
-  return type.IsStringTy() || type.IsBufferTy() || type.IsInetTy() ||
-         type.IsUsymTy() || type.IsTupleTy();
-}
-} // namespace
 
 void CodegenLLVM::visit(Integer &integer)
 {
@@ -834,6 +827,25 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateLifetimeEnd(buf);
     expr_ = nullptr;
   }
+  else if (call.func == "strftime")
+  {
+    auto elements = AsyncEvent::Strftime().asLLVMType(b_);
+    StructType *strftime_struct = b_.GetStructType(call.func + "_t",
+                                                   elements,
+                                                   true);
+
+    AllocaInst *buf = b_.CreateAllocaBPF(strftime_struct, call.func + "_args");
+    b_.CreateStore(b_.GetIntSameSize(strftime_id_, elements.at(0)),
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(0) }));
+    strftime_id_++;
+    Expression &arg = *call.vargs->at(1);
+    arg.accept(*this);
+    b_.CreateStore(expr_,
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(1) }));
+    expr_ = buf;
+    expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
+    // b_.CreateLifetimeEnd(buf);
+  }
   else if (call.func == "kstack" || call.func == "ustack")
   {
     Value *stackid = b_.CreateGetStackId(
@@ -934,13 +946,13 @@ void CodegenLLVM::visit(Map &map)
 
 void CodegenLLVM::visit(Variable &var)
 {
-  if (!var.type.IsAggregate())
+  if (needMemcpy(var.type))
   {
-    expr_ = b_.CreateLoad(variables_[var.ident]);
+    expr_ = variables_[var.ident];
   }
   else
   {
-    expr_ = variables_[var.ident];
+    expr_ = b_.CreateLoad(variables_[var.ident]);
   }
 }
 
@@ -1586,15 +1598,15 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
     variables_[var.ident] = val;
   }
 
-  if (!var.type.IsAggregate())
-  {
-    b_.CreateStore(expr_, variables_[var.ident]);
-  }
-  else
+  if (needMemcpy(var.type))
   {
     b_.CREATE_MEMCPY(variables_[var.ident], expr_, var.type.size, 1);
     if (!assignment.expr->is_variable && dyn_cast<AllocaInst>(expr_))
       b_.CreateLifetimeEnd(expr_);
+  }
+  else
+  {
+    b_.CreateStore(expr_, variables_[var.ident]);
   }
 }
 
@@ -1845,6 +1857,7 @@ void CodegenLLVM::visit(Probe &probe)
     int starting_cat_id = cat_id_;
     int starting_system_id = system_id_;
     int starting_time_id = time_id_;
+    int starting_strftime_id = strftime_id_;
     int starting_join_id = join_id_;
     int starting_helper_error_id = b_.helper_error_id_;
     int starting_non_map_print_id = non_map_print_id_;
@@ -1854,6 +1867,7 @@ void CodegenLLVM::visit(Probe &probe)
       cat_id_ = starting_cat_id;
       system_id_ = starting_system_id;
       time_id_ = starting_time_id;
+      strftime_id_ = starting_strftime_id;
       join_id_ = starting_join_id;
       b_.helper_error_id_ = starting_helper_error_id;
       non_map_print_id_ = starting_non_map_print_id;
@@ -2316,7 +2330,7 @@ void CodegenLLVM::createFormatStringCall(Call &call, int &id, CallArgs &call_arg
     expr_deleter_ = nullptr;
     arg.accept(*this);
     Value *offset = b_.CreateGEP(fmt_args, {b_.getInt32(0), b_.getInt32(i)});
-    if (arg.type.IsAggregate())
+    if (needMemcpy(arg.type))
     {
       b_.CREATE_MEMCPY(offset, expr_, arg.type.size, 1);
       if (!arg.is_variable && dyn_cast<AllocaInst>(expr_))
@@ -2406,7 +2420,7 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
   // Store content
   Value *content_offset = b_.CreateGEP(buf, { b_.getInt32(0), b_.getInt32(2) });
   b_.CREATE_MEMSET(content_offset, b_.getInt8(0), arg.type.size, 1);
-  if (arg.type.IsAggregate())
+  if (needMemcpy(arg.type))
     b_.CREATE_MEMCPY(content_offset, expr_, arg.type.size, 1);
   else
     b_.CreateStore(expr_, content_offset);
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -4,6 +4,7 @@
 #include "arch/arch.h"
 #include "ast/async_event_types.h"
 #include "bpftrace.h"
+#include "codegen_helper.h"
 #include "irbuilderbpf.h"
 #include "utils.h"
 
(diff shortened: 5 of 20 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -879,20 +879,47 @@ void SemanticAnalyser::visit(Call &call)
...
-    if (check_varargs(call, 0, 1)) {
-      if (is_final_pass()) {
-        if (call.vargs && call.vargs->size() > 0) {
+    if (check_varargs(call, 0, 2))
+    {
+      if (is_final_pass())
+      {
+        if (call.vargs == nullptr || call.vargs->size() == 0)
+        {
+          std::string fmt_default = "%H:%M:%S\n";
```
danobi, line 888: This should be a constant. Maybe
```
constexpr static auto default_time_fmt = "%H...";
```
at top of file.

src/ast/semantic_analyser.cpp
```diff
@@ -879,20 +879,47 @@ void SemanticAnalyser::visit(Call &call)
...
+          bpftrace_.time_args_.push_back(fmt_default.c_str());
+          warning("Using `time()` to print time directly is now deprecated.",
```
danobi, line 890: "directly" is a bit confusing in this context. Maybe this is more clear: "Using `time()` to implicitly print time is deprecated. Print the returned value instead".

src/ast/semantic_analyser.cpp
```diff
@@ -879,20 +879,47 @@ void SemanticAnalyser::visit(Call &call)
...
       }
     }
+    if (call.vargs == nullptr || call.vargs->size() < 2)
+    {
+      call.type = CreateNone();
+    }
+    else
+    {
+      call.type = CreateTimestamp();
+    }
```
danobi, line 922: Can't you just have the `call.type = CreateTimestamp()` right after `bpftrace_.timestamp_args_.push_back(fmt.str)` and omit this entirely?

src/ast/semantic_analyser.cpp
```diff
@@ -879,20 +879,47 @@ void SemanticAnalyser::visit(Call &call)
...
             String &fmt = static_cast<String&>(fmt_arg);
             bpftrace_.time_args_.push_back(fmt.str);
+            warning("Using `time()` to print time directly is now deprecated.",
+                    call.loc);
+          }
+        }
+        else
+        {
+          if (check_arg(call, Type::string, 0, true) &&
+              check_arg(call, Type::integer, 1, false))
```
danobi, line 906: Use an `else if` instead

tests/semantic_analyser.cpp
```diff
@@ -511,6 +510,30 @@ TEST(semantic_analyser, call_time)
...
+  test("kprobe:f { time(\"%M:%S\", 1); }", 0);
+  test("kprobe:f { time(\"%M:%S\", nsecs); }", 0);
+  test("kprobe:f { time(\"%M:%S\", \"\"); }", 10);
+  test("kprobe:f { time(1, nsecs); }", 10);
+  test("kprobe:f { time(\"%M:%S\", 1, 1); }", 1);
+  test("kprobe:f { time(1, 1, 1); }", 1);
+  test("kprobe:f { time(\"%M:%S\", \"\", 1); }", 1);
+  test("kprobe:f { $ts = time(\"%M:%S\", 1); }", 1);
+  test("kprobe:f { @ts = time(\"%M:%S\", nsecs); }", 1);
+  test("kprobe:f { @[time(\"%M:%S\", nsecs)] = 1; }", 1);
```
danobi, line 524: I think all 3 of these need to work

src/ast/semantic_analyser.cpp
ember526: I was thinking that doing so (putting `call.type = CreateTimestamp()` inside the `if (is_final_pass())` block) would result in `call.type` having 'Type::none' in previous passes. I assumed it would be prudent to let it have the correct type for all passes.

danobi: Ah right. Can you move the `is_final_pass()` into the inner arms of if-branches, then? It'd be nice to keep all the type-dependent logic close to each other.

brendangregg:
What is this?

```
TEST(semantic_analyser, call_timestamp_warnings)
{
  test_for_warning("kprobe:f { time(); }", "deprecated");
  test_for_warning("kprobe:f { time(\"%M:%S\"); }", "deprecated");
  test_for_warning("kprobe:f { time(\"%M:%S\", nsecs); }", "deprecated", true);
}
```

Does this deprecate/break the old behavior? I don't see a compelling reason why we should. It's in ~10k printed copies of the BPF book that we cannot update, so no.

I also saw this (there's no reference docs, so it's not clear what you're doing):

```
          warning("Using time() to implicitly print time is deprecated. Print "
                  "the returned value instead.",
```

I called it time() because it was not strftime(). This sounds like you're turning it into strftime(). If you want strftime(), add a new thing called strftime() and leave time() alone.

brendangregg:
Why create a new printf() format character? Why not "%s"?

See printf.cpp: we convert a lot of types to String when printed.

```
    if (arg_type == Type::ksym || arg_type == Type::usym || arg_type == Type::probe ||
        arg_type == Type::username || arg_type == Type::kstack || arg_type == Type::ustack ||
        arg_type == Type::inet)
      arg_type = Type::string; // Symbols should be printed as strings
```

ember526:
@brendangregg  The reason we hesitated to use "%s" was that we were afraid It would send a message to user that `strftime()` returns a string and might cause some confusion. It's tempting to link "strftime()" with "string()". For example, people may end up using 'strncmp()` on the return value, which doesn't make sense.

Using "%ts" doesn't have that issue. But the problem with it is that creating a new specifier for the old friend `printf()` is not actually user-friendly either. It's hard to remember and lacks simplicity.

I think the async nature of `strftime()` makes it hard to make perfect decisions for users, since people might not always be aware that BPF programs come with many limitations.

brendangregg:
A) If it's called strftime() then people will use it with %s. (Where did %ts come from anyway? POSIX? I'm not familiar with it.)

B) I don't really see the use case other than with printf().

C) That ship has sailed anyway: we have things like kstack() and username() that use %s, but you can't use with strcmp().

It's a tracing language, not an application language. It only needs to support enough for common case tools -- anything more complex can be deferred to BCC. The existence of BCC allows us to keep bpftrace simple.

src/types.cpp
```diff
@@ -369,7 +369,11 @@ SizedType CreateBuffer(size_t size)
 
 SizedType CreateTimestamp()
 {
-  return SizedType(Type::timestamp, 64);
+  auto tp = SizedType(Type::timestamp, 24);
+  tp.tuple_elems.emplace_back(Type::integer, 8);
+  tp.tuple_elems.emplace_back(Type::integer, 8);
+  tp.tuple_elems.emplace_back(Type::integer, 8);
```
danobi, line 375: I don't think using `tuple_elems` is correct. Why is this necessary?

src/types.h
```diff
@@ -254,7 +254,7 @@ class SizedType
   };
   bool IsTupleTy(void) const
   {
-    return type == Type::tuple;
+    return type == Type::tuple || type == Type::timestamp;
```
danobi, line 257: timestamp is not a tuple type. I don't think you can do this. Why was this necessary?

docs/reference_guide.md
```diff
@@ -2681,6 +2682,18 @@ contains the memcopy'd value so the value at `print()` invocation time will be
 printed.  However for maps, only the handle to the map is queued up, so the
 printed map may be different than the map at `print()` invocation.
 
+# 24. `strftime()`: Formatted timestamp
+
+Syntax:
+- `strftime(const char *format, int nsecs)`
+
+This returns a formatted timestamp that is printable with `printf`. The format string should be supported by `strftime(3)`. Use format specifier "%s" when printing the return value. Note that `strftime` does not actually return a string.
```
danobi, line 2690: The "should" makes it sound optional. "must" is more accurate. See IETF RFC 2119: http://www.faqs.org/rfcs/rfc2119.html
```suggestion
This returns a formatted timestamp that is printable with `printf`. The format string must be supported by `strftime(3)`. Use format specifier "%s" when printing the return value. Note that `strftime` does not actually return a string.
```

src/ast/async_event_types.h
```diff
@@ -76,6 +76,22 @@ struct Time
   }
 } __attribute__((packed));
 
+struct Strftime
+{
+  uint64_t action_id;
```
danobi, line 81: I think we can drop the action_id. Don't think it's necessary

src/ast/codegen_llvm.cpp
```diff
@@ -834,6 +834,28 @@ void CodegenLLVM::visit(Call &call)
...
+
+    b_.CreateStore(b_.GetIntSameSize(time_id_, elements.at(1)),
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(1) }));
+    strftime_id_++;
+    Expression &arg = *call.vargs->at(1);
+    arg.accept(*this);
+    b_.CreateStore(expr_,
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(2) }));
+    expr_ = buf;
+    // b_.CreateLifetimeEnd(buf);
```
danobi, line 857: You'll want to put this in `expr_deleter_`: https://github.com/iovisor/bpftrace/blob/master/src/ast/codegen_llvm.h#L91 .

Callers should be running `expr_deleter_`. My #1386 should make it harder to forget about calling the deleter.

src/ast/codegen_llvm.cpp
```diff
@@ -834,6 +834,28 @@ void CodegenLLVM::visit(Call &call)
...
+    StructType *strftime_struct = b_.GetStructType(call.func + "_t",
+                                                   elements,
+                                                   true);
+
+    AllocaInst *buf = b_.CreateAllocaBPF(strftime_struct, call.func + "_t");
+    b_.CreateStore(b_.GetIntSameSize(asyncactionint(AsyncAction::strftime),
+                                     elements.at(0)),
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(0) }));
+
+    b_.CreateStore(b_.GetIntSameSize(time_id_, elements.at(1)),
```
danobi, line 849: Typo?
```suggestion
    b_.CreateStore(b_.GetIntSameSize(strftime_id_, elements.at(1)),
```

docs/reference_guide.md
```diff
@@ -2681,6 +2682,18 @@ contains the memcopy'd value so the value at `print()` invocation time will be
...
+
+Examples:
+
+```
+# bpftrace -e 'i:s:1 { printf("%s\n", strftime("%H:%M:%S", nsecs)); }'
```
danobi, line 2695: Maybe include some of the output too

src/bpftrace.cpp
```diff
@@ -1673,6 +1680,52 @@ std::string BPFtrace::resolve_uid(uintptr_t addr) const
...
+{
+  auto fmt = strftime_args_[strftime_id].c_str();
+  char timestr[STRING_SIZE];
+  struct tm tmp;
+  std::ifstream file("/proc/stat");
+  if (!file)
+  {
+    std::cerr << "Fail to open file /proc/stat: " << std::strerror(errno)
+              << std::endl;
+    return "";
```
danobi, line 1694: I'm wondering if it'd be more helpful to return `"(?)"`. Any thoughts?

ember526: Yeah, totally agree. It could be confusing from users' side to see nothing printed out.

src/types.cpp
ember526: I made it this way to avoid an error "invalid indirect read from stack off -40+8 size 16". I'll look for another solution I guess.

ember526:
Ready for review.

src/ast/irbuilderbpf.cpp
```diff
@@ -273,7 +273,7 @@ Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
   CreateCondBr(condition, lookup_success_block, lookup_failure_block);
 
   SetInsertPoint(lookup_success_block);
-  if (type.IsAggregate())
+  if (type.IsAggregate() || type.IsTimestampTy())
```
fbs, line 276: I think we should either put timestamp in IsAggregate or find a new name to describe them both (and all future async events which should be handled like this).

src/bpftrace.cpp
```diff
@@ -1673,6 +1686,52 @@ std::string BPFtrace::resolve_uid(uintptr_t addr) const
   return username;
 }
 
+std::string BPFtrace::resolve_timestamp(uint32_t strftime_id,
```
fbs, line 1689: Looks like this currently gives us 1s accuracy right?  I was hoping for to get closer to 100us but that should probably go  in a follow up ticket.

src/bpftrace.cpp
```diff
@@ -1673,6 +1686,52 @@ std::string BPFtrace::resolve_uid(uintptr_t addr) const
...
+    ss >> field;
+    if (field == "btime")
+    {
+      ss >> btime;
+      if (ss.fail())
+        btime = 0;
+      break;
+    }
+  }
+  if (btime == 0)
```
fbs, line 1716: btime should be a "kernel constant", making it possible to read its value once and storing it (maybe before code gets loaded into the kernel even). That would remove some useless string parsing from the "hot path" .

src/ast/async_event_types.h
```diff
@@ -76,6 +76,20 @@ struct Time
...
 
+struct Strftime
+{
+  uint64_t strftime_id;
+  uint64_t nsecs_since_boot;
+
+  std::vector<llvm::Type*> asLLVMType(ast::IRBuilderBPF& b)
+  {
+    return {
+      b.getInt64Ty(), // strftime id
```
fbs, line 87:
```suggestion
      b.getInt64Ty(), // asyncid
```
nit: to keep it consistent with the othrs

docs/reference_guide.md
```diff
@@ -2701,6 +2702,25 @@ contains the memcopy'd value so the value at `print()` invocation time will be
 printed.  However for maps, only the handle to the map is queued up, so the
 printed map may be different than the map at `print()` invocation.
 
+# 24. `strftime()`: Formatted timestamp
+
+Syntax:
+- `strftime(const char *format, int nsecs)`
+
+This returns a formatted timestamp that is printable with `printf`. The format string must be supported by `strftime(3)`. Use format specifier "%s" when printing the return value. Note that `strftime` does not actually return a string.
```
fbs, line 2710: can you align this to the rest of the doc, 80 (or 100?) char lines

docs/reference_guide.md
```diff
@@ -89,6 +89,7 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [21. `buf()`: Buffers](#21-buf-buffers)
     - [22. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
     - [23. `print()`: Print Value](#23-print-print-value)
+    - [24. `strftime()`: Formatted timestamp](#23-formatted-timestamp)
```
fbs, line 92:
```suggestion
    - [24. `strftime()`: Formatted timestamp](#24-formatted-timestamp)
```

src/bpftrace.cpp
suyuee: Yes. If we want someting better, we may have to choose something more accurate than btime. `clock_gettime(CLOCK_REALTIME)` might help though I'm not so sure if that's what we want; and if we use it, it's likely we cannot allow `strftime` to accept user specified timestamp.

src/bpftrace.cpp
suyuee: Sure, will do.

src/ast/async_event_types.h
suyuee: Actually, it's not `asyncid` since `strftime()` doesn't call `CreatePerfEventOutput` directly and we don't need the `asyncid` to indicate what the data is in `perf_event_printer`. Instead, it's the index in the `bpftrace.strftime_args_` which stores the format strings of strftime.

docs/reference_guide.md
suyuee: sure

src/ast/irbuilderbpf.cpp
suyuee: I'd agree. I think `IsCompound` could be a nice name here since C++ seems to use similar term: [`std::is_compound`](https://en.cppreference.com/w/cpp/types/is_compound) and it includes array, tuple, class and so on (`Type::timestamp` seems like a class type to me).  Btw there could be some other options like `IsCollectionTy`, `IsComposite` and `IsCombined` though I'm not sure if there are as good.

suyuee: Discussed with @danobi and thought that it might be good to add a helper function instead.

docs/reference_guide.md
```diff
@@ -2701,6 +2702,28 @@ contains the memcopy'd value so the value at `print()` invocation time will be
...
 
+## 24. `strftime()`: Formatted timestamp
+
+Syntax:
+- `strftime(const char *format, int nsecs)`
+
+This returns a formatted timestamp that is printable with `printf`. The format
+string must be supported by `strftime(3)`. Use format specifier "%s" when
+printing the return value. Note that `strftime` does not actually return a
+string.
```
fbs, line 2713:
```suggestion
string in bpf  (kernel), the formatting happens in userspace.
```

src/ast/codegen_llvm.cpp
```diff
@@ -26,7 +27,7 @@ namespace {
 bool shouldBeOnStackAlready(SizedType &type)
```
fbs, line 27: should we move this to codegen_helper too then?

docs/reference_guide.md
```diff
@@ -2701,6 +2702,28 @@ contains the memcopy'd value so the value at `print()` invocation time will be
 printed.  However for maps, only the handle to the map is queued up, so the
 printed map may be different than the map at `print()` invocation.
 
+## 24. `strftime()`: Formatted timestamp
+
```
fbs, line 2706: Wonder if we should mark this experimental as we're so close to the 0.11 release What do you think @danobi

docs/reference_guide.md
suyuee: It looks much better, thanks.

src/ast/codegen_llvm.cpp
suyuee: I'd agree. Sounds reasonable.

CHANGELOG.md
```diff
@@ -50,11 +50,14 @@ and this project adheres to
   - [#1377](https://github.com/iovisor/bpftrace/pull/1377)
 - Add support for non-map print()
   - [#1381](https://github.com/iovisor/bpftrace/pull/1381)
+- Add builtin function: `strftime`
+  - [#1387](https://github.com/iovisor/bpftrace/pull/1387)
```
fbs, line 54: needs to move up due to 0.11 release

Issue #1282, opened 2020-04-18
Question: uprobe on arbitrary address (without defined symbol)


mzr:
Can I place uprobe on an address that doesn't relate to any symbol?

[Here](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#3-uprobeuretprobe-dynamic-tracing-user-level) is a suggestion that it is possible:

>Syntax:
>```
>uprobe:library_name:function_name[+offset]
>uprobe:library_name:address
>uretprobe:library_name:function_name
>```

But when I run: `sudo bpftrace -e 'uprobe:/full/path/to/binary:0x40076a { printf("called"); }` I get:
```
Attaching 1 probe...
Could not resolve address: /full/path/to/binary:0x40076a
```

This error message corresponds to [this code](https://github.com/iovisor/bpftrace/blob/8cd398b9bc5021557db40da5a3b057f7eb92c15b/src/attached_probe.cpp#L413):
```
  if (symbol.empty())
  {
    sym.address = probe_.address;
    bcc_elf_foreach_sym(probe_.path.c_str(), sym_address_cb, &option, &sym);

    symbol = sym.name;
    func_offset = probe_.address - sym.start;

    if (!sym.start) {
      std::stringstream ss;
      ss << "0x" << std::hex << probe_.address;
      throw std::runtime_error("Could not resolve address: " + probe_.path + ":" + ss.str());
    }
  }
```

And there if I understand correctly `bcc_elf_foreach_sym` is looking for a symbol defined at `sym_address_cb`.

The binary `/full/path/to/binary` is being ptraced by gdb.

So, is it possible to define uprobe on arbitrary address without any symbol defined on it?

danobi:
I guess it could be done but what is the use case? Why can you not reference the address by function+offset?

mzr:
TBH my use case is CTF right now, but what if one tries to set uprobes inside stripped binaries?

danobi:
Oh interesting. I suppose we could allow it since we're checking instruction boundaries already. What do you think @fbs , @tyroguru ?

>  but what if one tries to set uprobes inside stripped binaries (eg. running on prod)?

Generally stripping binaries is a pretty bad idea b/c it breaks a lot of tooling. Like stack traces.

mzr:
> Generally stripping binaries is a pretty bad idea b/c it breaks a lot of tooling. Like stack traces.
[...]

PR #1388, opened 2020-06-18
Allow uprobe placement on arbitrary addresses when --unsafe is used.

mzr:
Follow-up to #1282.
When there is no symbol found for requested address of uprobe placement and `--unsafe` is set, warning is emitted, and uprobe is placed as normally.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1070,6 +1070,20 @@ Attaching 1 probe...
 Unsafe uprobe in the middle of the instruction: /bin/bash:main+1
 ```
 
+Using --unsafe option you can also place uprobes on arbitrary addresses.
+This might come in handy when the binary is stripped.
+```
+$ echo 'int main(){return 0;}' | gcc -xc -o bin -
+$ nm bin | grep main
+...
+0000000000001119 T main
+...
+$ strip bin
+# bpftrace --unsafe -e 'uprobe:bin:0x1119 { printf("main called\n"); }'
+Attaching 1 probe...
+WARNING: could not determine instruction boundary for uprobe:bin:4377 (binary appears stripped). Misaligned probes can lead to tracee crashes!
+```
+
 Examples in situ:
 [(uprobe) search /tools](https://github.com/iovisor/bpftrace/search?q=uprobe%3A+path%3Atools&type=Code)
 [(uretprobe) /tools](https://github.com/iovisor/bpftrace/search?q=uretprobe%3A+path%3Atools&type=Code)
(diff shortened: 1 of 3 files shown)


fbs:
nice, thanks for working on this! 

`Unsafe uprobe:bin:4377 not matching any symbol at address 0x1119`

I think the message can be improved a bit. Matching a symbol isn't a requirement for bpftrace. All it wants to know is that it is not inserting a probe in the middle of an instruction and then crashing the tracee. Maybe something along the lines of `WARNING: could not determine instruction boundary (binary appears stripped),  misaligned probes can lead to tracee crashes`

mzr:
I don't think that's the correct warning either. At least it doesn't fit into the control flow in the source code. Originally it failed because there was no symbol defined at that address (`if (!sym.start)`).
Also as far as I understand, alignment within an instruction is checked later by calling the `check_alignment`.

How about 
```
WARNING: Unsafe uprobe:bin:4377 not matching any symbol at address 0x1119. Skipping all alignment checks such as instruction boundary, Misaligned probes can lead to tracee crashes.
```
?

fbs:
> I don't think that's the correct warning either. At least it doesn't fit into the control flow in the source code. Originally it failed because there was no symbol defined at that address (`if (!sym.start)`).

Having a `sym.start` doesn't mean that there exists a symbol at that specific address. What it does, through the `sym_address_cb`, is finding a "function" that contains that address (`func_start <= address <= func_end`). That way it can use `func_start` as a starting point for checking instruction boundaries as `func_start` will always be aligned properly.

`!sym.start` can mean that the file is (partly) stripped from the required symbols or that the address is actually outside of function boundaries the current code doesn't really know which it is.

Maybe we could add some checks that at least verify that the address is not completely out of bounds, but that can be done in a later pr too.

fbs:
Can you squash the commits into one? :)

fbs:
@danobi what do you think of the warning message?

docs/reference_guide.md
```diff
@@ -1063,6 +1063,20 @@ Attaching 1 probe...
...
+```
+$ echo 'int main(){return 0;}' | gcc -xc -o bin -
+$ nm bin | grep main
+...
+0000000000001119 T main
+...
+$ strip bin
+# bpftrace --unsafe -e 'uprobe:bin:0x1119 { printf("main called\n"); }'
+Attaching 1 probe...
+Unsafe uprobe:bin:4377 not matching any symbol at address 0x1119
```
danobi, line 1077: Update output?

mzr: Yup. I forgot about that. Now it's changed.

danobi:
Thanks!

PR #1393, opened 2020-06-21
Codegen before optimizing

### LINKED ISSUE #1368: test codegen before opt
Currently we test codegen _after_ optimization. this has 2 issues: 

- We have to rewrite generated code to deal with LLVM versions which adds a lot of noise
- We don't actually test the code we generate, we test what LLVM does to it. This makes harder to relate a change in the codegen to a change in the LLVM output.

Are there any objections to switching the tests to the code we generate? It will be a bit more verbose but we wont have to deal with llvm version specific optimizations anymore, only with version specific function call differences.

fbs:
Based on #1389 
Fixes #1368


mmisono:
Looks good but I found that I got the slightly different output from `codegen.call_lhist` when building with clang (use LLVM 10): 

<details>

```diff
@@ -74,7 +74,7 @@
   store i64 %2, i64* %6
   store i64 %3, i64* %5
-  %14 = load i64, i64* %7
-  %15 = load i64, i64* %8
-  %16 = icmp slt i64 %15, %14
+  %14 = load i64, i64* %8
+  %15 = load i64, i64* %7
+  %16 = icmp slt i64 %14, %15
   br i1 %16, label %lhist.lt_min, label %lhist.ge_min

@@ -83,15 +83,15 @@

 lhist.ge_min:                                     ; preds = %entry
-  %17 = load i64, i64* %6
-  %18 = load i64, i64* %8
-  %19 = icmp sgt i64 %18, %17
+  %17 = load i64, i64* %8
+  %18 = load i64, i64* %6
+  %19 = icmp sgt i64 %17, %18
   br i1 %19, label %lhist.gt_max, label %lhist.le_max

 lhist.le_max:                                     ; preds = %lhist.ge_min
-  %20 = load i64, i64* %5
+  %20 = load i64, i64* %8
   %21 = load i64, i64* %7
-  %22 = load i64, i64* %8
-  %23 = sub i64 %22, %21
-  %24 = udiv i64 %23, %20
+  %22 = sub i64 %20, %21
+  %23 = load i64, i64* %5
+  %24 = udiv i64 %22, %23
   %25 = add i64 %24, 1
   store i64 %25, i64* %4

@@ -100,9 +100,9 @@
 lhist.gt_max:                                     ; preds = %lhist.ge_min
-  %27 = load i64, i64* %5
+  %27 = load i64, i64* %6
   %28 = load i64, i64* %7
-  %29 = load i64, i64* %6
-  %30 = sub i64 %29, %28
-  %31 = udiv i64 %30, %27
+  %29 = sub i64 %27, %28
+  %30 = load i64, i64* %5
+  %31 = udiv i64 %29, %30
   %32 = add i64 %31, 1
   store i64 %32, i64* %4
```

</details>

When building with gcc, the test passes. I'm not sure what is a cause.

mmisono:
This is the result of CI with clang: https://github.com/mmisono/bpftrace/runs/806809132

mmisono:
It seems the problem is the order of evaluation of function arguments. For example, 
https://github.com/iovisor/bpftrace/blob/f2a6bdeac94e2dbef9a6190de54d5f01802c54ab/src/ast/codegen_llvm.cpp#L2265
then IR differs depending on which load is evaluated first.

tests/codegen/common.h
```diff
@@ -248,13 +260,30 @@ static std::string rewrite_attrs(const std::string &line)
...
+static std::string rewrite_function_hdr(const std::string &line)
 {
 #if LLVM_VERSION_MAJOR >= 10
-  // FROM: define i64 @BEGIN(i8*) local_unnamed_addr section \"s_BEGIN_1\" {
-  // TO:   define i64 @BEGIN(i8* %0) local_unnamed_addr section \"s_BEGIN_1\" {
-  static std::regex re("(@[^\\(]+)\\(([^\\)]+)\\)");
-  return std::regex_replace(line, re, "$1($2 %0)");
+  // FROM: define internal i64 @linear(i64, i64, i64, i64) #1 section
+  // \"helpers\" { TO:   define internal i64 @linear(i64 %0, i64 %1, i64 %2, i64
+  // %3) #1 section \"helpers\" {
```
danobi, line 268: Any reason not to newline like this?
```suggestion
  // FROM: define internal i64 @linear(i64, i64, i64, i64) #1 section \"helpers\" {
  // TO:   define internal i64 @linear(i64 %0, i64 %1, i64 %2, i64 %3) #1 section \"helpers\" {
```

danobi:
Any reason we need to version freeze this? This seems pretty safe from a end user perspective. It's only our test infra that's changed.

fbs:
> Any reason we need to version freeze this? This seems pretty safe from a end user perspective. It's only our test infra that's changed.

It should be safe, I just don't want to complicate the PRs that we want to land for 0.11. Once 0.11 is released this can go in

fbs:
rebasing this on #1425 (if we decide to merge that) should work around the llvm build issue for now

fbs:
The reordering was quite annoying but I think i've fixed it :)

fbs:
ugh, seems like we have some more typing issues to fix. I've excluded those for now to avoid bloating this PR futher. I will fix those as part of #1091

PR #1394, opened 2020-06-22
Pointer type

fbs:
This is based on #1393 which adds a bit of noise

This PR replaces the `cast` and `ctx` types with `record` (struct/union) and `pointer` types.

The `cast` and `ctx` types are confusing as they can be either pointers or structures or more, depending on some optional attributes.

I still have to do more testing but the original test suite passes with minimal changes, which is a good sign.

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1222,10 +1222,10 @@ void CodegenLLVM::visit(Unop &unop)
       case bpftrace::Parser::token::MUL:
       {
         int size = type.size;
-        if (type.is_pointer)
+        if (type.IsPtrTy())
         {
           // When dereferencing a 32-bit integer, only read in 32-bits, etc.
-          size = type.pointee_size;
+          size = type.GetPointeeTy()->size;
         }
         AllocaInst *dst = b_.CreateAllocaBPF(SizedType(type.type, size), "deref");
         b_.CreateProbeRead(ctx_, dst, size, expr_, unop.loc);
@@ -1240,15 +1240,17 @@ void CodegenLLVM::visit(Unop &unop)
         abort();
     }
   }
-  else if (type.IsCastTy() || type.IsCtxTy())
+  else if (type.IsPtrTy())
   {
-    switch (unop.op) {
+    switch (unop.op)
+    {
       case bpftrace::Parser::token::MUL:
       {
-        if (type.is_pointer && unop.type.IsIntTy())
+        if (unop.type.IsIntegerTy())
         {
-          int size = unop.type.size;
-          AllocaInst *dst = b_.CreateAllocaBPF(unop.type, "deref");
+          auto *et = type.GetPointeeTy();
+          int size = et->GetIntBitWidth() / 8;
+          AllocaInst *dst = b_.CreateAllocaBPF(*et, "deref");
           b_.CreateProbeRead(ctx_, dst, size, expr_, unop.loc);
           expr_ = b_.CreateIntCast(b_.CreateLoad(dst),
                                    b_.getInt64Ty(),
@@ -1257,8 +1259,7 @@ void CodegenLLVM::visit(Unop &unop)
         }
         break;
       }
-      default:
-        ; // Do nothing
+      default:; // Do nothing
     }
   }
   else
@@ -1349,8 +1350,14 @@ void CodegenLLVM::visit(Ternary &ternary)
 void CodegenLLVM::visit(FieldAccess &acc)
 {
   SizedType &type = acc.expr->type;
-  assert(type.IsCastTy() || type.IsCtxTy() || type.IsTupleTy());
+  assert(type.IsRecordTy() || type.IsCastTy() || type.IsCtxTy() || type.IsTupleTy());
   auto scoped_del = accept(acc.expr);
+
+  bool is_ctx = type.IsCtxAccess();
+  bool is_tparg = type.is_tparg;
+  bool is_internal = type.is_internal;
+  bool is_kfarg = type.is_kfarg;
+  assert(type.IsRecordTy() || type.IsTupleTy());
 
   if (type.is_kfarg)
   {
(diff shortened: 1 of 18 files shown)


fbs:
BTF related tests pass too now.

src/ast/codegen_llvm.cpp
```diff
@@ -1442,16 +1486,17 @@ void CodegenLLVM::visit(FieldAccess &acc)
       Value *masked = b_.CreateAnd(shifted, field.bitfield.mask);
       expr_ = masked;
     }
-    else if (type.IsCtxTy() &&
-             (field.type.IsIntTy() ||
-              (field.type.IsCastTy() && field.type.is_pointer)))
+    else if ((field.type.IsIntTy() || field.type.IsPtrTy()) &&
+             type.IsCtxAccess())
     {
+      // TODO XXX fixme
```
fbs, line 1492: fixme

src/ast/semantic_analyser.cpp
```diff
@@ -1071,10 +1099,11 @@ void SemanticAnalyser::visit(Map &map)
         map.vargs->at(i) = cast;
         expr = cast;
       }
-      else if (expr->type.IsCtxTy())
+      else if (expr->type.IsCtxAccess())
       {
         // map functions only accepts a pointer to a element in the stack
-        error("context cannot be used as a map key", map.loc);
+        // TODO xx checkme
+        // error("context cannot be used as a map key", map.loc);
```
fbs, line 1106: fixme

src/printf.cpp
```diff
@@ -37,7 +37,8 @@ std::string verify_format_string(const std::string &fmt, std::vector<Field> args
         arg_type == Type::username || arg_type == Type::kstack || arg_type == Type::ustack ||
         arg_type == Type::inet)
       arg_type = Type::string; // Symbols should be printed as strings
-    if (arg_type == Type::cast && args.at(i).type.is_pointer)
+    // TODO XXX
```
fbs, line 40: fixme

danobi:
Looks like this PR is blocked on #1393.

danobi:
I took a quick look at the commits and they look quite reasonable. It'll be really nice to have these proper types in.

fbs:
One thing I did't really look at yet is how bad the memory leaks are with this. As we're doing 10 passes this might quite a few f objects. I'm not sure how to best address that yet

fbs:
With #1368  merged this should be a bit eaiser

fbs:
Looks like clang-format 6 and 9 don't like each other. I'll try and fix the clang-format check once we're happy about the code

fbs:
rebased onto master

src/types.h
```diff
@@ -313,6 +314,11 @@ SizedType CreateUInt64();
 SizedType CreateString(size_t size);
 SizedType CreateArray(size_t num_elements, const SizedType &element_type);
 SizedType CreatePointer(const SizedType &pointee_type);
+/**
+   Create a record type
+
+   \param size size in bits
```
danobi, line 305: A bit confusing that `size` is in bits for Type::record. The comment next to `size_t size` in `struct SizedType` says size is in bytes. Anything we can do about that?

src/types.cpp
```diff
@@ -57,8 +59,31 @@ std::ostream &operator<<(std::ostream &os, const SizedType &type)
   return os;
 }
 
+bool SizedType::IsSameType(const SizedType &t) const
```
danobi, line 59: It isn't really obvious to me the difference between `IsSameType` and `IsEqual`. Is the only difference that type size isn't checked? Maybe `IsEqualIgnoringSize` then?

src/types.cpp
```diff
@@ -295,7 +319,7 @@ SizedType CreatePointer(const SizedType &pointee_type)
 
 SizedType CreateRecord(size_t size, const std::string &name)
 {
-  auto ty = SizedType(Type::record, size);
+  auto ty = SizedType(Type::record, size / 8);
```
danobi, line 306: I'll leave the previous comments up for my own benefit later, but why go through all the trouble of passing in bits if it gets stored as bytes later? If you really want these semantics, maybe name the constructor parameter `size_bits` to be extra clear.

src/types.h
```diff
@@ -170,7 +162,7 @@ class SizedType
 
   bool IsPtrTy() const
   {
-    return ((type == Type::pointer) || (IsIntTy() && is_pointer));
+    return type == Type::pointer;
```
danobi, line 165: Ah I see it's removed now. Leaving the previous comment up for my own benefit.

src/ast/codegen_llvm.cpp
```diff
@@ -1373,26 +1380,32 @@ void CodegenLLVM::visit(FieldAccess &acc)
...
-  type = SizedType(type);
-  type.size = cstruct.size;
-  type.cast_type = cast_type;
+  // This overwrites the stored type!
+  type = CreateRecord(cstruct.size, cast_type);
+  if (is_ctx)
+    type.MarkCtxAccess();
+  type.is_tparg = is_tparg;
+  type.is_internal = is_internal;
+  type.is_kfarg = is_kfarg;
```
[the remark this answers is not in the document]
fbs: > I realize it might be out of the scope of this commit, but couldn't all this be done in the semantic analyser?

It should be. I think this is the result of probe expansion happening in all phases instead of early on. 

> Also, why are the variables copied and reassigned? Doesn't look like changes are being made to them.

`type` is a references so the `CreateRecord()` call overwrites the attributes. Which causes issues with the wildcard expansion

src/types.cpp
fbs: Yes that can  do with a  bit of a cleanup, I'll give it a go

I think that having  sizes in bits eventually makes more sense. As booleans will be impossible otherwise

fbs:
@mmisono @olsajiri can you take a look at this one please? I think this is a improvement but as it's quite a fundamental change some consensus would be nice.

fbs:
I'll merge this tuesday if there are no  comments

mmisono:
@fbs sorry for the late reply & thanks for the improvement! I found an issue so I create #1464

PR #1395, opened 2020-06-23
Fixes for big endian arch

sumanthkorikkar:
Hi @fbs, @danobi, @mmisono 

This patchset tries to reduce failures in big endian architectures. The problem occurs due to type conversion issues. I have tried to put detailed analysis in the commit, so that we can avoid these in future. Let me know, if it is fine. Thank you!!

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -546,7 +546,7 @@ void CodegenLLVM::visit(Call &call)
                                               false);
     AllocaInst *buf = b_.CreateAllocaBPF(buf_struct, "buffer");
 
-    b_.CreateStore(length,
+    b_.CreateStore(b_.CreateIntCast(length, b_.getInt8Ty(), false),
                    b_.CreateGEP(buf, { b_.getInt32(0), b_.getInt32(0) }));
     Value *buf_data_offset = b_.CreateGEP(buf,
                                           { b_.getInt32(0), b_.getInt32(1) });
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1219,7 +1219,9 @@ void CodegenLLVM::visit(Unop &unop)
         }
         AllocaInst *dst = b_.CreateAllocaBPF(SizedType(type.type, size), "deref");
         b_.CreateProbeRead(ctx_, dst, size, expr_, unop.loc);
-        expr_ = b_.CreateLoad(dst);
+        expr_ = b_.CreateIntCast(b_.CreateLoad(dst),
+                                 b_.getInt64Ty(),
+                                 type.IsSigned());
         b_.CreateLifetimeEnd(dst);
         break;
       }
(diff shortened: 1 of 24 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -1487,6 +1494,9 @@ void CodegenLLVM::visit(ArrayAccess &arr)
     AllocaInst *dst = b_.CreateAllocaBPF(stype, "array_access");
     b_.CreateProbeRead(ctx_, dst, element_size, src, arr.loc);
     expr_ = b_.CreateLoad(dst);
+    expr_ = b_.CreateIntCast(b_.CreateLoad(dst),
```
fbs, line 1497: I'm confused about this one (double load here). If  dst is an i32, how does loading it first and then casting it to i64 fix it?

sumanthkorikkar: - bpftrace -v -e 'struct MyStruct { int y[4]; } uprobe:./testprogs/array_access:test_struct { $s = (struct MyStruct *)arg0; $x = $s->y[0]; printf("Result: %d\n", $x); exit(); }'
- lets say, the $s->y[0] has value of 1.
 
1. In codegen, expr_ is of type 32 bit.
In LE, (0x1 00 00 00) | (00 00 00 00)
In BE, (0x0 00 00 01) | (00 00 00 00)
      (Converted 32 bit) 
		 
2. Now arg.type.type is 4 bytes for $s->y[0] (not promoted to 64 bits)

3. With my previous patch, "bpftrace: Fix type conversion in printf() builtin"
   arg.type.cast_type is not set:
   	Hence: arg_values.push_back(std::make_unique<PrintableInt>(                  
              *reinterpret_cast<uint64_t *>(arg_data + arg.offset)));

 4. In format(), this will take lower 32 bits, which is zero for BE.
 5. When %lx format specifier in printf is used, we do get 0x1 00 00 00 00 in BE
 
 By converting to 64 bit, we ensure it is written as 0x0 00 00 00 00 00 00 01 in BE.

src/ast/codegen_llvm.cpp
```diff
@@ -1494,6 +1494,9 @@ void CodegenLLVM::visit(ArrayAccess &arr)
     AllocaInst *dst = b_.CreateAllocaBPF(stype, "array_access");
     b_.CreateProbeRead(ctx_, dst, element_size, src, arr.loc);
     expr_ = b_.CreateLoad(dst);
```
sumanthkorikkar, line 1496: Yes. I should remove the previous one. I missed it. Thanks.

src/ast/codegen_llvm.cpp
sumanthkorikkar: Yes. I should remove the previous load. I missed it. Thanks.

fbs: oh so the issue is with printf and not with bpf loading from the wrong address?

sumanthkorikkar: Yes. In this case, it is an issue with printf. ( As type size is actually 4 bytes). One kind of resolution is to change it here. In other cases, where integers are promoted to 8 bytes( For example: FieldAccess/Unop for 32 bit types, then promoting the types in codegen is necessary)

fbs: Ah ok that makes sense. We need to get rid of that upcast, but untill we do that this looks like an OK fix.

sumanthkorikkar: Yes. even I think in long term, it's better to remove the promotion and have some other solution. This can probably make the type conversion easier to understand and can prevent future issues.

sumanthkorikkar: Removed the double load. Fixed.

sumanthkorikkar:
So anything that needs to be done from my side ?. Thanks.

PR #1404, opened 2020-06-30
Enable printf, cat, system to have more args

### LINKED ISSUE #1336: printf() only supports 7 arguments
It would be useful to have support for more arguments. I don't think there's a technical limit.

More args

n/a

suyuee:
This PR enables `printf`, `cat` and `system` to have more than 7 arguments. It can be more user-friendly to allow any reasonable number of arguments for these builtin calls which use formatted strings. Resolves #1336.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -48,53 +48,69 @@ namespace bpftrace {
 DebugLevel bt_debug = DebugLevel::kNone;
 bool bt_verbose = false;
 volatile sig_atomic_t BPFtrace::exitsig_recv = false;
+const int FMT_BUF_SZ = 512;
 
-int format(char * s, size_t n, const char * fmt, std::vector<std::unique_ptr<IPrintable>> &args) {
+std::string format(std::string fmt,
+                   std::vector<std::unique_ptr<IPrintable>> &args)
+{
+  std::string retstr;
+  auto buffer = std::vector<char>(FMT_BUF_SZ);
+  auto check_snprintf_ret = [](int r) {
+    if (r < 0)
+    {
+      std::cerr << "format() error occurred: " << std::strerror(errno)
+                << std::endl;
+      abort();
+    }
+  };
   // Args have been made safe for printing by now, so replace nonstandard format
   // specifiers with %s
-  std::string str = std::string(fmt);
   size_t start_pos = 0;
-  while ((start_pos = str.find("%r", start_pos)) != std::string::npos)
+  while ((start_pos = fmt.find("%r", start_pos)) != std::string::npos)
   {
-    str.replace(start_pos, 2, "%s");
+    fmt.replace(start_pos, 2, "%s");
     start_pos += 2;
   }
-  fmt = str.c_str();
 
-  int ret = -1;
-  switch(args.size()) {
-    case 0:
-      ret = snprintf(s, n, "%s", fmt);
-      break;
-    case 1:
-      ret = snprintf(s, n, fmt, args.at(0)->value());
-      break;
-    case 2:
-      ret = snprintf(s, n, fmt, args.at(0)->value(), args.at(1)->value());
-      break;
-    case 3:
-      ret = snprintf(s, n, fmt, args.at(0)->value(), args.at(1)->value(), args.at(2)->value());
-      break;
-    case 4:
-      ret = snprintf(s, n, fmt, args.at(0)->value(), args.at(1)->value(), args.at(2)->value(), args.at(3)->value());
-      break;
-    case 5:
-      ret = snprintf(s, n, fmt, args.at(0)->value(), args.at(1)->value(), args.at(2)->value(),
-        args.at(3)->value(), args.at(4)->value());
-      break;
-    case 6:
-      ret = snprintf(s, n, fmt, args.at(0)->value(), args.at(1)->value(), args.at(2)->value(),
-        args.at(3)->value(), args.at(4)->value(), args.at(5)->value());
-      break;
-    default:
-      std::cerr << "format() can only take up to 7 arguments (" << args.size() << ") provided" << std::endl;
-      abort();
+  auto tokens_begin = std::sregex_iterator(fmt.begin(),
+                                           fmt.end(),
+                                           format_specifier_re);
+  auto tokens_end = std::sregex_iterator();
+
+  // replace format string tokens with args one by one
+  int literal_text_pos = 0; // starting pos of literal text (text that is not
+                            // format specifier)
+  int i = 0;                // args index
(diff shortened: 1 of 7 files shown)


src/bpftrace.cpp
```diff
@@ -49,52 +49,46 @@ DebugLevel bt_debug = DebugLevel::kNone;
...
-    default:
-      std::cerr << "format() can only take up to 7 arguments (" << args.size() << ") provided" << std::endl;
+  const std::regex re("%-?[0-9]*(\\.[0-9]+)?[a-zA-Z]+");
+  auto tokens_begin = std::sregex_iterator(fmt.begin(), fmt.end(), re);
+  auto tokens_end = std::sregex_iterator();
+
+  int text_pos = 0;
+  int i = 0;
+  std::string str;
+  while (tokens_begin != tokens_end)
```
danobi, line 73: Can you add a comment describing the general algorithm? A bit confusing to figure out from the code alone.

src/bpftrace.cpp
```diff
@@ -573,66 +587,52 @@ void perf_event_printer(void *cb_cookie, void *data, int size __attribute__((unu
...
-    bpftrace->out_->message(MessageType::printf, std::string(buffer), false);
-  } else {
-    auto buf = std::make_unique<char[]>(required_size+1);
-    // if for some reason the size is still wrong the string
-    // will just be silently truncated
-    format(buf.get(), required_size, fmt, arg_values);
-    bpftrace->out_->message(MessageType::printf, std::string(buf.get()), false);
-  }
+  std::string retstr = format(fmt, arg_values);
+  bpftrace->out_->message(MessageType::printf, retstr, false);
```
danobi, line 635: Not sure a temporary named `retstr` improves readablility. Maybe just:
```suggestion
  bpftrace->out_->message(MessageType::printf, format(fmt, arg_values), false);
```

src/bpftrace.cpp
```diff
@@ -573,66 +587,52 @@ void perf_event_printer(void *cb_cookie, void *data, int size __attribute__((unu
...
+    if (sz >= FMT_BUF_SZ)
     {
       std::cerr << "cat() command to long (" << sz << " bytes): ";
-      std::cerr << buffer << std::endl;
+      std::cerr << retstr << std::endl;
       return;
     }
     std::stringstream buf;
-    cat_file(buffer, bpftrace->cat_bytes_max_, buf);
+    cat_file(retstr.c_str(), bpftrace->cat_bytes_max_, buf);
```
danobi, line 623: I think we can drop the FMT_BUF_SZ check. I don't believe it does anything useful anymore.

So instead it might look like:
```
std:stringstream buf;
cat_file(format(fmt, arg_vaules).c_str(), bpftrace->cat_bytes_max_, buf);
```

src/bpftrace.cpp
```diff
@@ -573,66 +587,52 @@ void perf_event_printer(void *cb_cookie, void *data, int size __attribute__((unu
...
     auto arg_values = bpftrace->get_arg_values(args, arg_data);
 
-    const int BUFSIZE = 512;
-    char buffer[BUFSIZE];
-    int sz = format(buffer, BUFSIZE, fmt, arg_values);
-    // Return value is required size EXCLUDING null byte
-    if (sz >= BUFSIZE)
+    std::string retstr = format(fmt, arg_values);
+    auto sz = retstr.length();
+    if (sz >= FMT_BUF_SZ)
```
danobi, line 596: Same as `cat`

src/bpftrace.cpp
```diff
@@ -48,53 +48,67 @@ namespace bpftrace {
...
 bool bt_verbose = false;
 volatile sig_atomic_t BPFtrace::exitsig_recv = false;
+const int FMT_BUF_SZ = 512;
 
-int format(char * s, size_t n, const char * fmt, std::vector<std::unique_ptr<IPrintable>> &args) {
+std::string format(std::string fmt,
+                   std::vector<std::unique_ptr<IPrintable>> &args)
+{
+  std::string retstr;
+  int n = FMT_BUF_SZ;
```
danobi, line 57: I prefer avoiding ambiguously named temporaries like `n`. It's ok as a loop counter but usually not as a function-scope variable. It forces the reader to keep track of unnecessary details.

I think using `buffer.capacity()` would be better.

src/bpftrace.cpp
```diff
@@ -48,53 +48,67 @@ namespace bpftrace {
...
+  auto buffer = std::vector<char>(n);
+  char *str = buffer.data();
```
danobi, line 59: Drop this temporary too. It's more clear to the reader when they see `buffer.data()` than `str`.

src/bpftrace.cpp
```diff
@@ -48,53 +48,67 @@ namespace bpftrace {
...
+  auto check_snprintf_ret = [](int r) {
+    if (r < 0)
+    {
+      std::cerr << "format() error occurred: " << std::strerror(errno)
+                << std::endl;
+      abort();
+    }
+    return;
```
danobi, line 67: No need to return from a void lambda
```suggestion
```

src/bpftrace.cpp
```diff
@@ -48,53 +48,67 @@ namespace bpftrace {
...
-    abort();
-  }
-  return ret;
+  auto tokens_begin = std::sregex_iterator(fmt.begin(),
+                                           fmt.end(),
+                                           format_specifier_re);
+  auto tokens_end = std::sregex_iterator();
+
+  // replace format string tokens with args one by one
+  int interval_pos = 0; // starting pos of interval strings (text that is not
```
danobi, line 84: Maybe `literal_text` instead of `interval`. `interval` doesn't really make sense.

ember526:
Conflicts Resolved.

src/bpftrace.cpp
```diff
@@ -48,53 +48,69 @@ namespace bpftrace {
...
+    // take out the literal text
+    retstr += fmt.substr(literal_text_pos,
+                         tokens_begin->position() - literal_text_pos);
+    // replace current specifier with an arg
+    int r = snprintf(buffer.data(),
+                     buffer.capacity(),
+                     tokens_begin->str().c_str(),
+                     args.at(i)->value());
+    check_snprintf_ret(r);
+    if ((unsigned)r >= buffer.capacity())
```
danobi, line 95: Style nit
```suggestion
    if (static_cast<size_t>(r) >= buffer.capacity())
```

danobi:
One more thing: can you put the non-checklist portions of the PR description in the commit message?

suyuee:
Sure. Already done.

PR #1407, opened 2020-07-02
Use llvm-as to validate generated IR

fbs:
llvm-as emits warnings when the IR has typing issues:

```
 error: '%2' defined with type 'i8*'
  call void @llvm.lifetime.start.p0i8(i64 -1, i64* nonnull %2)
```

We can use this to validate that the IR we generate is valid and avoid
regressions (like the OOM error).

As the codegen reference output is LLVM 7 based this is currently only
enabled for LLVM 7. Testing the version specific things is not
necessary as we don't have any version specific logic in the codegen.

The tests that do not pass are ignored. Eventually we shouldn't have to
ignore any of them. Adding new entries to this list should be avoided,
unless the test overlaps with something currently on the list.

builtin_func_1.ll
builtin_func_uprobe_2.ll
call_buf_implicit_size.ll
call_buf_size_literal.ll
call_buf_size_nonliteral.ll
call_print_non_map.ll
call_printf.ll
if_else_printf.ll
if_nested_printf.ll
if_printf.ll

Related to #1091


fbs:
Ready for review

PR #1410, opened 2020-07-07
Add wildcard support for kfunc probe types

olsajiri:
Adding wildcard support for kfunc probe types. It's now possible to use following probes:
    
`  # bpftrace -e 'kfunc:ksys_* { @[comm, probe] = kstack; }'`


mmisono:
@olsajiri 
What do you think about this comment?

> It seems this check is too restrictive. For example, `vfs_open(const struct path *path, struct file *file)` and `vfs_write(struct file *file, ...)` have the same arg `struct file *file`, but cannot create the program to access the arg.
> 
> ```
> % sudo ./src/bpftrace -e 'f:vfs_open, f:vfs_write { @ = args->file->f_count; }'
> stdin:1:12-24: ERROR: Failed: probe has attach points with mixed arguments
> f:vfs_open, f:vfs_write { @ = args->file->f_count; }
>            ~~~~~~~~~~~~
> ```
> 
> This works on the current master.
> https://github.com/iovisor/bpftrace/pull/1347#discussion_r436262720

Well, I think it is OK to cope with it later in another PR. Other than that looks good to me.

olsajiri:
> @olsajiri
> What do you think about this comment?

oops, did I overlook this comment somewhere? sry

> 
> > It seems this check is too restrictive. For example, `vfs_open(const struct path *path, struct file *file)` and `vfs_write(struct file *file, ...)` have the same arg `struct file *file`, but cannot create the program to access the arg.
> > ```
> > % sudo ./src/bpftrace -e 'f:vfs_open, f:vfs_write { @ = args->file->f_count; }'
> > stdin:1:12-24: ERROR: Failed: probe has attach points with mixed arguments
> > f:vfs_open, f:vfs_write { @ = args->file->f_count; }
> >            ~~~~~~~~~~~~
> > ```
> > 
> > 
> > This works on the current master.
> > [#1347 (comment)](https://github.com/iovisor/bpftrace/pull/1347#discussion_r436262720)
> 
> Well, I think it is OK to cope with it later in another PR. Other than that looks good to me.

the issue here is that the argument you want to use need to have the same position, because that's how we access arguments in kfunc program - through its index into the argument array

so when you have single code for multiple probes, that's accessing an argument, that argument must be in the same place in arguments array for each probe, so the example above would not work anyway:

```
vfs_open(const struct path *path, struct file *file)
vfs_write(struct file *file, ...)
```

because 'file' has different argument index in each function

we could make the check more permissive to allow this, I thought this to be a corner case which might not be worth to add so far.. I'll check on it

however I'm planning to add 'outside' wildcard matching that should solve all these cases.. meaning that bpftrace will create extra separate probe code for every matched probe, which can't be shared with others

olsajiri:
rebased, @danobi @fbs , could you guys please check on it?

thanks

src/ast/field_analyser.cpp
```diff
@@ -5,6 +5,16 @@
 namespace bpftrace {
 namespace ast {
 
+void FieldAnalyser::error(const std::string &msg, const location &loc)
+{
+  bpftrace_.error(err_, loc, msg);
```
danobi, line 10: Why store in `err_` and not use `out_` like `warning()`?

src/ast/field_analyser.cpp
```diff
@@ -216,22 +229,59 @@ void FieldAnalyser::visit(Predicate &pred)
   pred.expr->accept(*this);
 }
 
-void FieldAnalyser::visit(AttachPoint &ap __attribute__((unused)))
+bool FieldAnalyser::compare_args(std::map<std::string, SizedType> args1,
+                                 std::map<std::string, SizedType> args2)
```
danobi, line 233: I think we can save a copy here
```suggestion
bool FieldAnalyser::compare_args(const std::map<std::string, SizedType>& args1,
                                 const std::map<std::string, SizedType>& args2)
```

src/ast/field_analyser.cpp
```diff
@@ -40,6 +40,15 @@ void FieldAnalyser::visit(Jump &jump __attribute__((unused)))
 {
 }
 
+void FieldAnalyser::check_kfunc_args(void)
+{
+  if (has_kfunc_probe_ && has_mixed_args_)
+  {
+    error("Failed: probe has attach points with mixed arguments",
```
danobi, line 47: Having `Failed: ` seems unnecessary since there should already be a big `ERROR: ` printed. My 2c

fbs:
can you add a changelog entry for this?

src/ast/field_analyser.cpp
olsajiri: yep, should be out_, thanks

src/ast/field_analyser.cpp
olsajiri: true, will change

src/ast/field_analyser.cpp
olsajiri: ok, will change, thanks

src/ast/field_analyser.cpp
mmisono: @olsajiri 
I believe `err_` should be used here since FieldAnalyzer later checks if an error occurs by consulting `err_` length:
https://github.com/iovisor/bpftrace/blob/f0701114e22709ad4828c8c83f84f467e558e670/src/ast/field_analyser.cpp#L369
Now `semantic_analyser_btf.kfunc` test fails because field analyzer always returns zero.

PR #1413, opened 2020-07-08
Ignore trailing kernel module annotation

### LINKED ISSUE #1036: kprobe pattern matching returns "Invalid argument" for some probes
Kernel: 5.3.11-200.fc30.x86_64
OS: Fedora 30
bpftrace v0.9.2

If I look for all probes using the `'*pmem*'` search pattern, I get 43 hits:

```
# bpftrace -l 'kprobe:*pmem*' | wc -l
43
```

When enabling these 43 probes using the following I get errors with 33 them:

```
# bpftrace -e 'kprobe:*pmem* { @[func] = count(); }'
Attaching 43 probes...
Warning: could not attach probe kprobe:write_pmem [nd_pmem], skipping.
Warning: could not attach probe kprobe:release_free_pmem [libnvdimm], skipping.
Warning: could not attach probe kprobe:pmem_should_map_pages [libnvdimm], skipping.
Warning: could not attach probe kprobe:pmem_sector_size [libnvdimm], skipping.
Warning: could not attach probe kprobe:pmem_rw_page [nd_pmem], skipping.
Warning: could not attach probe kprobe:pmem_release_queue [nd_pmem], skipping.
Warning: could not attach probe kprobe:pmem_release_disk [nd_pmem], skipping.
[...snip...]
```

However, if I fully qualify the probes, it works, eg:

```
# bpftrace -e 'kprobe:write_pmem, kprobe:release_free_pmem { @[func] = count(); }'
Attaching 2 probes...
```

It's unclear why the glob pattern matching fails, but the fully qualified approach works.  This issue doesn't appear to be specific to the `*pmem*` pattern/modules as I see the same behaviour with other search patterns such as `*dax*` and `*vfs*` for example.

The symptom is similar to #68 except the example provided works on my environment.

danobi:
Before, `find_wildcard_matches` would choke on lines
available_filter_functions like:

    ehci_disable_ASE [ehci_hcd]

This patch has k[ret]probes ignore the module annotation.

This closes #1036 .

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,8 @@ and this project adheres to
   - [#1374](https://github.com/iovisor/bpftrace/pull/1374)
 - Remove registers that are not in struct pt_regs (x86-64)
   - [#1383](https://github.com/iovisor/bpftrace/issues/1383)
+- Ignore trailing kernel module annotation for k[ret]probe's
+  - [#1413](https://github.com/iovisor/bpftrace/pull/1413)
 
 #### Tools
 
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -148,11 +148,7 @@ void SemanticAnalyser::builtin_args_tracepoint(AttachPoint *attach_point,
    * 2. sets is_tparg so that codegen does the real type setting after
    *    expansion.
    */
-  auto symbol_stream = bpftrace_.get_symbols_from_file(
-      "/sys/kernel/debug/tracing/available_events");
-  auto matches = bpftrace_.find_wildcard_matches(attach_point->target,
-                                                 attach_point->func,
-                                                 *symbol_stream);
+  auto matches = bpftrace_.find_wildcard_matches(*attach_point);
   if (!matches.empty())
   {
     auto &match = *matches.begin();
@@ -1611,11 +1607,7 @@ void SemanticAnalyser::visit(FieldAccess &acc)
         continue;
       }
 
-      auto symbol_stream = bpftrace_.get_symbols_from_file(
-          "/sys/kernel/debug/tracing/available_events");
-      auto matches = bpftrace_.find_wildcard_matches(attach_point->target,
-                                                     attach_point->func,
-                                                     *symbol_stream);
+      auto matches = bpftrace_.find_wildcard_matches(*attach_point);
       for (auto &match : matches) {
         std::string tracepoint_struct =
             TracepointFormatParser::get_struct_name(attach_point->target,
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -51,10 +51,15 @@ namespace {
  * If an optional prefix is provided, lines must start with it to count as a
  * match, but the prefix is stripped from entries in the result set.
  * Wildcard tokens ("*") are accepted in func.
+ *
+ * If `ignore_trailing_module` is true, will ignore trailing kernel module.
+ * For example, `[ehci_hcd]` will be ignored in:
+ *     ehci_disable_ASE [ehci_hcd]
  */
 std::set<std::string> find_wildcard_matches_internal(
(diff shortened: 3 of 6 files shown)


fbs:
Ah nice you beat me to it. Wonder if it makes sense to define kprobes like `kprobe:module:func` to be able to attach to "module functions" that overlap in name

danobi:
Yeah, I think `kprobe:module:func` would work. And I think it could be done as a followup PR.

Do you think the following rules make sense?
* `kprobe:foo*`
  * ignore module name (module implicitly wildcarded)
* '`kprobe:*:foo*`
  * ignore module name (module explicitly wildcarded)
* `kprobe:bar*:foo*`
  * don't ignore module name (not full wildcard)
* non-wildcard cases work as you'd expect -- full text match only

Or is this too complicated? The alternative is to abandon the logic changes in this PR (the cleanup commits I think are still good) and always require module param to be specified if you want to include module kprobes.

fbs:
> Or is this too complicated? 

Sounds good

> The alternative is to abandon the logic changes in this PR (the cleanup commits I think are still good) and always require module param to be specified if you want to include module kprobes.

I think we should land this asap, we can always refactor it later. There is quite a lot of duplicate code between the listing and the function finding and kaddr/ksym expansion that we can clean up and simplify.

zhangjaycee:
> Yeah, I think `kprobe:module:func` would work. And I think it could be done as a followup PR.
> 
> Do you think the following rules make sense?
> 
> * `kprobe:foo*`
>   
>   * ignore module name (module implicitly wildcarded)
> * '`kprobe:*:foo*`
>   
>   * ignore module name (module explicitly wildcarded)
> * `kprobe:bar*:foo*`
>   
>   * don't ignore module name (not full wildcard)
> * non-wildcard cases work as you'd expect -- full text match only
> 
> Or is this too complicated? The alternative is to abandon the logic changes in this PR (the cleanup commits I think are still good) and always require module param to be specified if you want to include module kprobes.

Hi, I wonder if there has been any progress or plans on this feature? It is exactly what we need recently. Thanks!

zhangjaycee:
> 

Functionally, `kfunc:module:func` is good. But I found it requires relatively new kernel version support. So I wonder if anyone plans to implement `kprobe:module:func`.  If it is not in process,  I would also appreciate it if anyone could give some suggestions on how to implement it. Thanks!

viktormalik:
> Functionally, `kfunc:module:func` is good. But I found it requires relatively new kernel version support. So I wonder if anyone plans to implement `kprobe:module:func`. If it is not in process, I would also appreciate it if anyone could give some suggestions on how to implement it. Thanks!

Yes, it requires kernel 5.5+ with BTF support, if you call that new :-). 

Since we have this functionality, we don't plan to duplicate it for kprobes. If you want to do it yourself, you can get inspired by the kfunc implementation for some parts. The main difference will be in the way probes are attached. While kfunc specifies the target module/function upon BPF program loading, kprobes do that during the attachment itself. So, you'll have to check if (and how) BCC's `bpf_attach_kprobe` supports specifying the module to attach to.

zhangjaycee:
> > Functionally, `kfunc:module:func` is good. But I found it requires relatively new kernel version support. So I wonder if anyone plans to implement `kprobe:module:func`. If it is not in process, I would also appreciate it if anyone could give some suggestions on how to implement it. Thanks!
> 
> Yes, it requires kernel 5.5+ with BTF support, if you call that new :-).
> 
> Since we have this functionality, we don't plan to duplicate it for kprobes. If you want to do it yourself, you can get inspired by the kfunc implementation for some parts. The main difference will be in the way probes are attached. While kfunc specifies the target module/function upon BPF program loading, kprobes do that during the attachment itself. So, you'll have to check if (and how) BCC's `bpf_attach_kprobe` supports specifying the module to attach to.

Thanks for the information :-). I will give feedback if there is any progress on this in the future.

PR #1416, opened 2020-07-10
Enable top and div args for printing Type::avg maps

### LINKED ISSUE #1385: top and divisor arguments don't work when printing an avg() map
### bpftrace --info
```
System
  OS: Linux 5.7.2-arch1-1 #1 SMP PREEMPT Wed, 10 Jun 2020 20:36:24 +0000
  Arch: x86_64

Build
  version: v0.10.0-182-gadf93
  LLVM: 10
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  libbpf: yes
  libbpf btf dump: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  lsm: yes
```

### What reproduces the bug?

Not exactly sure if this should be a bug report or feature request, as it might be intended behavior.
When printing a map whose values have been generated with `avg()`, I should be able to specify a top and div argument like `print(@, 10, 1000)` to print the top 10 results divided by 1000. The current implementation does not support this.

An example use case would be measuring average execution times for system calls.

Minimal example:

```d
t:raw_syscalls:sys_enter {
    @start[tid] = nsecs;
}

t:raw_syscalls:sys_exit {
    $start_t = @start[tid];
    $id = args->id;

    if ($id != $start_id) {
        return;
    }

    $dur = (nsecs - $start_t);
    @avg_t[$id] = avg($dur);
}

END {
    clear(@start);

    printf("Printing top 10 average times (us):\n");
    print(@avg_t, 10, 1000);
    clear(@avg_t);
}
```

Expected output:
```
Printing top 10 average times (us):
@avg_t[28]: 45
@avg_t[230]: 74
@avg_t[35]: 82
@avg_t[217]: 117
@avg_t[202]: 89641762
@avg_t[7]: 630540603
@avg_t[232]: 761679852
@avg_t[281]: 1390053830
@avg_t[271]: 3706822648
@avg_t[23]: 3971615699
```

Actual output:
```
Printing top 10 average times (us):
@avg_t[38]: 2276
@avg_t[89]: 2362
@avg_t[4]: 2374
@avg_t[24]: 2485
@avg_t[286]: 3485
@avg_t[39]: 4216
@avg_t[233]: 4322
@avg_t[102]: 4345
..... (omitted)
@avg_t[28]: 45934
@avg_t[230]: 74438
@avg_t[35]: 82891
@avg_t[217]: 117918
@avg_t[202]: 89641762244
@avg_t[7]: 630540603371
@avg_t[232]: 761679852015
@avg_t[281]: 1390053830002
@avg_t[271]: 3706822648732
@avg_t[23]: 3971615699056
```

suyuee:
The top and div arguments of `print()` currently does not work for avg maps
because it wasn't implemented. I think it makes sense to have them supported.

Resolves #1385.
##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to
 - Fix `printf` not allowing format specifiers to be directly followed by
   alphabetic characters
   - [#1414](https://github.com/iovisor/bpftrace/pull/1414)
+- Fix `top` and `div` arguments of `print()` not working for Type::avg maps
+  - [#1416](https://github.com/iovisor/bpftrace/pull/1416)
 
 #### Changed
 
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -275,14 +275,27 @@ void TextOutput::map_hist(BPFtrace &bpftrace, IMap &map, uint32_t top, uint32_t 
   }
 }
 
-void TextOutput::map_stats(BPFtrace &bpftrace, IMap &map,
-                           const std::map<std::vector<uint8_t>, std::vector<int64_t>> &values_by_key,
-                           const std::vector<std::pair<std::vector<uint8_t>, int64_t>> &total_counts_by_key) const
+void TextOutput::map_stats(
+    BPFtrace &bpftrace,
+    IMap &map,
+    uint32_t top,
+    uint32_t div,
+    const std::map<std::vector<uint8_t>, std::vector<int64_t>> &values_by_key,
+    const std::vector<std::pair<std::vector<uint8_t>, int64_t>>
+        &total_counts_by_key) const
 {
+  uint32_t i = 0;
   for (auto &key_count : total_counts_by_key)
   {
     auto &key = key_count.first;
     auto &value = values_by_key.at(key);
+
+    if (map.type_.IsAvgTy() && top)
+    {
+      if (i++ < (values_by_key.size() - top))
+        continue;
+    }
+
     out_ << map.name_ << map.key_.argument_value_list_str(bpftrace, key) << ": ";
 
     int64_t count = (int64_t)value.at(0);
@@ -295,7 +308,7 @@ void TextOutput::map_stats(BPFtrace &bpftrace, IMap &map,
     if (map.type_.IsStatsTy())
       out_ << "count " << count << ", average " <<  average << ", total " << total << std::endl;
     else
-      out_ << average << std::endl;
+      out_ << average / div << std::endl;
   }
 
   out_ << std::endl;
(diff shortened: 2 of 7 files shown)


src/output.cpp
```diff
@@ -586,11 +604,18 @@ void JsonOutput::map_stats(BPFtrace &bpftrace, IMap &map,
     out_ << "{";
 
   uint32_t i = 0;
+  uint32_t j = 0;
   for (auto &key_count : total_counts_by_key)
   {
     auto &key = key_count.first;
     auto &value = values_by_key.at(key);
 
+    if (map.type_.IsAvgTy() && top)
```
danobi, line 613: Any reason we restrict to `IsAvgTy` here? Seems stats maps is still broken. Maybe reword the PR so both `avg` and `stats` is fixed.

suyuee: I suppose `top` might be ambiguous for `stats`. `stats` returns the average, the sum and the count but `top` does not specify which statistic it's based on. I noticed that `print_map_stats` would sort its entries based on the average but the user might not necessarily expect that. As for `div`, it would somehow be strange, too: if use `div` on all three statistics, then the output does not have avg * count == sum anymore; also it's likely that count become 0 after dividing div. I'm thinking if we should disallow top&div for stats maps during semantic checking.

suyuee: And I wonder if there is an easy way to test it. I'm thinking about have the following to work as the testcase:
```
bpftrace -e 'i:ms:1 {@[nsecs / 1000000 % 10] = avg(1);} i:s:1 { print(@, 3, 0); clear(@); exit();}'
```
Not sure if I could come up with a regular expression to verify the output.

I notice there doesn't seem to be runtime tests for `print(map, top, div)` (or I missed them out); there could be another pr for runtime tests if there aren't. Any thoughts?

CHANGELOG.md
```diff
@@ -98,6 +98,8 @@ and this project adheres to
   - [#1383](https://github.com/iovisor/bpftrace/issues/1383)
 - Ignore trailing kernel module annotation for k[ret]probe's
   - [#1413](https://github.com/iovisor/bpftrace/pull/1413)
+- Fix `top` and `div` arguments of `print()` not working for Type::avg maps
+  - [#1416](https://github.com/iovisor/bpftrace/pull/1416)
```
fbs, line 102: needs to move up due to 0.11 release

suyuee:
Add runtime tests.

danobi:
Seems like the test has an off-by-one issue

suyuee:
Fixed the issue and add a runtime test for json format output. Seems like some tests are queued and won't run. Guess I'll push it later to rerun the tests.

suyuee:
Push again to rerun the tests.

suyuee:
Sure, already done:) #1433.

PR #1417, opened 2020-07-14
0.11 release

fbs:


fbs:
If there are no objection i'll do the release tomorrow (~24h from now)

danobi:
+1

mmisono:
Could you upload static build binaries as release assets? Having static download links would be useful :)

Issue #1080, opened 2020-01-10
Break up `class BPFtrace` into smaller classes


danobi:
The class has gotten too big and does way too many things now. We should break it up into smaller components.

### Work items:
- [x] child process management
  * `cmd_` and `pid_` are used and abused too often b/c they are public
- [x] logging

fbs:
Child process management might   be one to move. 

logging can be moved out of it too

jordalgo:
The two work items were completed. Going to close

PR #1418, opened 2020-07-14
Add a logging framework to bpftrace

### LINKED ISSUE #1411: Improve logging of bpftrace
Got these ideas from @fbs and @danobi.

Generally, it would be nice to provide logging facilities to developers and users so that they could have better visibility of bpftrace internal states.
1. bpftrace currently doesn't provide tools for developers to print debug messages. It would be helpful if we can have some handy tools.
2. There should be semantic warnings for users when they try to do things unproperly and having a consistent way to print the warnings can make it easier to do so. Also, bpftrace should allow users to enable and disable warnings like gcc does. 
3. bpftrace should provide options about what kind of debug information to print, as in [#131](https://github.com/iovisor/bpftrace/issues/131).

Implementing a logging framework for bpftrace. It would be much easier to deal with these issues with the right infrastructure. 

- [x] Add the logging infra. #1418 
- [x] Replace callsites of class BPFtrace logging facilities (warning() & error()). #1443
- [x] Add the option to turn off warnings (maybe with `-w` as in gcc).#1444
- [ ] Add the option to show certain debugging information (ast, ir).
- [x] Change the error/warning messages sent through std::cerr.#1449 

N/A

suyuee:
This PR introduces a stream-style logging infrastructure. It supports the basic
functionalities for now and can be extended with future PRs.

There are several benefits for including it in bpftrace:
- It would be more convinient for developers to keep track of inner status of
bpftrace and debug.
- It can help to print consistent error/warning messages.
- Future work will offer users the option to toggle on/off certain messages.

Some usage examples and their output:
```
LOG(ERROR, call.loc, std::cerr)
    << "Non-map print() only takes 1 argument, " << call.vargs->size()
    << " found";
stdin:1:7-18: ERROR: Non-map print() only takes 1 argument, 2 found
i:s:1 {print(1,2)}
      ~~~~~~~~~~~

LOG(WARNING) << "this is a warning!";
WARNING: this is a warning!

LOG(DEBUG, std::cout) << "this is a message.";
DEBUG: [/home/yue/bpftrace/src/ast/semantic_analyser.cpp:861]
this is a message.
```
Related to #1411

##### Checklist

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -1,0 +1,196 @@
+#include "log.h"
+
+namespace bpftrace {
+
+std::string logtype_str(LogType t)
+{
+  switch (t)
+  {
+    // clang-format off
+    case LogType::DEBUG   : return "DEBUG";
+    case LogType::INFO    : return "INFO";
+    case LogType::WARNING : return "WARNING";
+    case LogType::ERROR   : return "ERROR";
+    // clang-format on
+    default:
+      std::cerr << "Invalid log type." << std::endl;
+      abort();
+  }
+}
+
+Log& Log::get()
+{
+  static Log log;
+  return log;
+}
+
+void Log::take_input(LogType type,
+                     const std::optional<location>& loc,
+                     std::ostream& out,
+                     std::string&& input)
+{
+  auto print_out = [&]() {
+    out << logtype_str(type) << ": " << input << std::endl;
+  };
+
+  if (loc)
+  {
+    if (src_.empty())
+    {
+      std::cerr << "Log: cannot resolve location before calling set_source()."
+                << std::endl;
+      print_out();
+    }
+    else if (loc->begin.line == 0)
+    {
+      std::cerr << "Log: invalid location." << std::endl;
+      print_out();
+    }
+    else if (loc->begin.line > loc->end.line)
+    {
+      std::cerr << "Log: loc.begin > loc.end: " << loc->begin << ":" << loc->end
+                << std::endl;
+      print_out();
+    }
+    else
+    {
+      log_with_location(type, loc.value(), out, input);
+    }
+  }
+  else
+  {
+    print_out();
+  }
+}
+
+const std::string Log::get_source_line(unsigned int n)
+{
+  // Get the Nth source line (N is 0-based). Return an empty string if it
+  // doesn't exist
+  std::string line;
+  std::stringstream ss(src_);
+  for (unsigned int idx = 0; idx <= n; idx++)
+  {
+    std::getline(ss, line);
+    if (ss.eof() && idx == n)
+      return line;
+    if (!ss)
+      return "";
+  }
+  return line;
+}
+
+void Log::log_with_location(LogType type,
+                            const location& l,
+                            std::ostream& out,
+                            const std::string& m)
+{
+  if (filename_.size())
+  {
+    out << filename_ << ":";
+  }
+
+  std::string msg(m);
+  const std::string& typestr = logtype_str(type);
+
+  if (!msg.empty() && msg.back() == '\n')
+  {
+    msg.pop_back();
+  }
+
+  /* For a multi line error only the line range is printed:
+     <filename>:<start_line>-<end_line>: ERROR: <message>
+  */
+  if (l.begin.line < l.end.line)
+  {
+    out << l.begin.line << "-" << l.end.line << ": " << typestr << ": " << msg
+        << std::endl;
+    return;
+  }
+
+  /*
+    For a single line error the format is:
+
+    <filename>:<line>:<start_col>-<end_col>: ERROR: <message>
+    <source line>
+    <marker>
+
+    E.g.
+
+    file.bt:1:10-20: error: <message>
+    i:s:1   /1 < "str"/
+            ~~~~~~~~~~
+  */
+  out << l.begin.line << ":" << l.begin.column << "-" << l.end.column;
+  out << ": " << typestr << ": " << msg << std::endl;
+
+  // for bpftrace::position, valid line# starts from 1
+  std::string srcline = get_source_line(l.begin.line - 1);
+
+  if (srcline == "")
+    return;
+
+  // To get consistent printing all tabs will be replaced with 4 spaces
+  for (auto c : srcline)
+  {
+    if (c == '\t')
+      out << "    ";
+    else
+      out << c;
+  }
+  out << std::endl;
+
+  for (unsigned int x = 0;
+       x < srcline.size() && x < (static_cast<unsigned int>(l.end.column) - 1);
+       x++)
+  {
+    char marker = (x < (static_cast<unsigned int>(l.begin.column) - 1)) ? ' '
+                                                                        : '~';
+    if (srcline[x] == '\t')
+    {
+      out << std::string(4, marker);
+    }
+    else
+    {
+      out << marker;
+    }
+  }
+  out << std::endl;
+}
+
+LogStream::LogStream(const std::string& file,
+                     int line,
+                     LogType type,
+                     std::ostream& out)
+    : sink_(Log::get()),
+      type_(type),
+      loc_(std::nullopt),
+      out_(out),
+      log_file_(file),
+      log_line_(line)
+{
+}
+
+LogStream::LogStream(const std::string& file,
+                     int line,
+                     LogType type,
+                     const location& loc,
+                     std::ostream& out)
+    : sink_(Log::get()),
+      type_(type),
+      loc_(loc),
+      out_(out),
+      log_file_(file),
+      log_line_(line)
+{
+}
+
+LogStream::~LogStream()
+{
+  std::string prefix = "";
+  if (type_ == LogType::DEBUG)
+    prefix = "[" + log_file_ + ":" + std::to_string(log_line_) + "]\n";
+  sink_.take_input(type_, loc_, out_, prefix + buf_.str());
+}
+
+}; // namespace bpftrace
--- a/src/log.h
+++ b/src/log.h
@@ -1,0 +1,92 @@
+#pragma once
+
+#include <iostream>
+#include <optional>
+#include <sstream>
+
+#include "bpftrace.h"
+
+namespace bpftrace {
+
+// clang-format off
+enum class LogType
+{
+  DEBUG,
+  INFO,
+  WARNING,
+  ERROR
+};
+// clang-format on
+
+class Log
+{
+public:
+  static Log& get();
+  void take_input(LogType type,
+                  const std::optional<location>& loc,
+                  std::ostream& out,
+                  std::string&& input);
+  inline void set_source(const std::string& filename, const std::string& source)
+  {
+    src_ = source;
+    filename_ = filename;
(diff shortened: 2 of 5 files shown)


src/log.cpp
```diff
@@ -0,0 +1,101 @@
+#include "log.h"
+
+namespace bpftrace {
+
+static std::unordered_map<LogType, std::string> log_levels = {
+  { LogType::debug, "INFO" },
+  { LogType::semantic, "INFO" },
+  { LogType::semantic_warning, "WARNING" },
+  { LogType::semantic_error, "ERROR" },
+};
```
danobi, line 10: I think an enum would be better here b/c std::unordered_map requires runtime initialization. And since it's not protected by a meyers singleton we open ourselves up to SIOF (which is more likely b/c logging can be expected to be used anywhere).

suyuee: I wanted this structure to give a string about log level, which is used by BPFtrace::log_with_location. I guess enum might not help with that. Maybe I would write a function which uses a switch statement to do the mapping. Not sure if that is good though. Btw, will taking away `static` solve the problem? Or it would be an issue as long as it's a global object.

fbs:
Nice!

> ```
> // can indicate which 'channel' to put the logs into
> // LLOG is for logs with loc information, need bpftrace_ to help with the
> // formatting (bpftrace.filename_, bpftrace._src)
> LLOG(LogType::semantic_error, bpftrace_, call.loc)
>     << "Non-map print() only takes 1 argument, " << call.vargs->size()
>     << " found" << LOGFLUSH;
> 
> // DLOG is for debug information
> // logs will be buffered if without LOGFLUSH
> DLOG << "Hello" << " there " << 123;
> // can indicate ostream
> DLOG << " End" << LOGFLUSH_OUT(std::cout);
> 
> // output from Log
> # bpftrace -e 'BEGIN {print(1, 2)}' 
> stdin:1:7-19: ERROR: Non-map print() only takes 1 argument, 2 found
> BEGIN {print(1, 2)}
>       ~~~~~~~~~~~~
> [/home/yue/bpftrace/src/ast/semantic_analyser.cpp:860]Hello there123
> [/home/yue/bpftrace/src/ast/semantic_analyser.cpp:861] End
> ```

Some thoughts:

I'm not sure about the logflush part. I'm not sure in which situation you would write `LLOG() << X << LOGFLUSH << Y << LOGFLUSH` over `LLOG() << x; LLOG() << y`. Maybe we can just flush it as soon as the object goes out of scope?

I think we can avoid passing bpftrace around if we setup the logger with the source lines in main and lift all the printing logic out of it (doesn't belong in bpftrace anyway). E.g.  `Log::Get().SetSource(script)` in main.

Maybe we can put the output stream in the constructor? E.g. `LLOG(Level = LogType::semantic_error, ostream = std::cout, location = 0) << x`

suyuee:
Thanks for the input!
> I think we can avoid passing bpftrace around if we setup the logger with the source lines in main and lift all the printing logic out of it (doesn't belong in bpftrace anyway). E.g. Log::Get().SetSource(script) in main.

I agree with you about this one. I was thinking about the same thing when I saw your discussion in #1080.

Then about this:
> I'm not sure about the logflush part. I'm not sure in which situation you would write LLOG() << X << LOGFLUSH << Y << LOGFLUSH over LLOG() << x; LLOG() << y. Maybe we can just flush it as soon as the object goes out of scope?

I now agree on that. I was actually thinking that there could be cases when we don't want to flush the log until some point like this part in semantic_analyser.cpp, it gathers all the error logs in one round which could come from more than one places; and check on them, print them out all together:
```
int SemanticAnalyser::analyse()
{
  // Multiple passes to handle variables being used before they are defined
  std::string errors;

  for (pass_ = 1; pass_ <= num_passes_; pass_++) {
    root_->accept(*this);
    errors = err_.str();
    if (!errors.empty()) {
      out_ << errors;
      return pass_;
    }
  }
  return 0;
}
```
But I thought about it again, maybe we can just use `err_` as the output stream of `LLOG`. Wasn't thinking straight.
Now that we can rid of the `LOGFLUSH` thing, it would be reasonable to move the output stream to the constructor.

suyuee:
Make some changes and add tests. Ready for review.

src/log.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+Log::Log()
+{
+}
+
+Log::~Log()
+{
+}
```
danobi, line 27: Let the compiler do this for you instead. In log.h:
```
Log() = default;
~Log() = default;
```

src/log.cpp
```diff
@@ -0,0 +1,182 @@
...
+  switch (t)
+  {
+    // clang-format off
+    case LogType::debug   : return "DEBUG";
+    case LogType::info    : return "INFO";
+    case LogType::warning : return "WARNING";
+    case LogType::error   : return "ERROR";
+    // clang-format on
+    default:
+      std::cerr << "Log type not found." << std::endl;
```
danobi, line 16: "Not found" reads kinda funny. How about this instead:
```suggestion
      std::cerr << "Invalid log type." << std::endl;
```

src/log.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+  return line;
+}
+
+void Log::log_with_location(LogType type,
+                            const location& l,
+                            std::ostream& out,
+                            const std::string& m)
+{
+  if (filename_ != "")
```
danobi, line 75: Style nit: "positive" conditions tend to be easier to read
```suggestion
  if (filename_.size())
```

src/log.h
```diff
@@ -0,0 +1,104 @@
...
+  }
+  ~LogLocStream();
+
+private:
+  Log& sink_;
+  LogType type_;
+  const location& loc_;
+  std::ostream& out_;
+  std::ostringstream buf_;
+};
```
danobi, line 93: Is a second class for this necessary? Can't you just have an overloaded constructor for `LogStream`?

src/log.h
```diff
@@ -0,0 +1,104 @@
...
+
+#define LOG(type, out) LogStream(type, out)
+// log with location
+#define LLOG(type, loc, out) LogLocStream(type, loc, out)
+// debugging log, use std::cerr by default
+#define DLOG                                                                   \
+  LogStream(LogType::debug, std::cerr)                                         \
+      << "[" << __FILE__ << ":" << __LINE__ << "]"
+#define DLOG_OUT(out)                                                          \
+  LogStream(LogType::debug, out) << "[" << __FILE__ << ":" << __LINE__ << "]"
```
danobi, line 103: I'd prefer if we don't have `DLOG` and `DLOG_OUT`. In general it's good to keep API surface area small. Is there a strong reason we need these? I think it'd be better if LOG(LogType::debug) always printed line numbers anyways.

src/log.h
```diff
@@ -0,0 +1,104 @@
...
+
+private:
+  Log& sink_;
+  LogType type_;
+  const location& loc_;
+  std::ostream& out_;
+  std::ostringstream buf_;
+};
+
+#define LOG(type, out) LogStream(type, out)
```
danobi, line 95: So right now users of the logging API will have to type `LOG(LogType::Blah)`. Can you find a way to make this just `LOG(Blah)`? I don't see any reason we need to make the user type more characters than necessary. Should be a fun exercise w/ C macros for you.

suyuee:
Improved codes, cleaned up APIs (only have `LOG` now and don't require `LogType::` prefix anymore).

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+class Log
+{
+public:
+  static Log& get();
+  void take_input(LogType type,
+                  const location& loc,
+                  bool with_loc,
+                  std::ostream& out,
+                  std::string&& input);
+  inline void set_source(std::string filename, std::string source)
```
danobi, line 29: Nit
```suggestion
  inline void set_source(const std::string& filename, std::string source)
```

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+namespace bpftrace {
+
+// clang-format off
+enum class LogType
+{
+  debug,
+  info,
+  warning,
+  error
+};
```
danobi, line 17: Typically enums are all caps (eg https://github.com/torvalds/linux/blob/master/include/uapi/linux/bpf.h#L122 ). I think that makes the log api look nicer too. So it'd be:
```
LOG(DEBUG) << "whatever";
```
That's also how it looks for glog (google's log library)

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+  void log_with_location(LogType,
+                         const location&,
+                         std::ostream&,
+                         const std::string&);
+};
+
+class LogStream
+{
+public:
+  LogStream(const std::string file,
```
danobi, line 58: Nit: doesn't matter in this case (compiler will probably RVO), but IIUC const-ref is still superior in case the api gets used differently later.
```suggestion
  LogStream(const std::string file,
```

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+            int line,
+            LogType type,
+            std::ostream& out = std::cerr);
+  LogStream(const std::string file,
```
danobi, line 62: Same as above

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+  bool with_loc_;
+  std::ostream& out_;
+  std::string log_file_;
+  int log_line_;
+  std::ostringstream buf_;
+};
+
+// Usage examples:
+// 1. LOG(warning) << "this is a " << "warning!"; (this goes to std::cerr)
+// 2. LOG(debug, std::out) << "this is a " << " message.";
```
danobi, line 88: Typo?
```suggestion
// 2. LOG(debug, std::cout) << "this is a " << " message.";
```

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+// 3. LOG(error, call.loc, std::cerr) << "this is a semantic error";
+// Note: LogType::debug will prepend __FILE__ and __LINE__ to the debug message
+
+#define LOG(...) LogStream(__FILE__, __LINE__, LogType::__VA_ARGS__)
```
danobi, line 92: Nice, very clever.

danobi:
PR is looking really good. Just some small comments

src/log.h
suyuee: I'd agree, will change.

src/log.h
suyuee: I guess we may want  `const std::string& source`, too?

suyuee:
Thanks for the review! Made some changes.

src/log.h
```diff
@@ -0,0 +1,93 @@
...
+    buf_ << v;
+    return *this;
+  }
+  ~LogStream();
+
+private:
+  Log& sink_;
+  LogType type_;
+  const location loc_;
+  bool with_loc_;
```
danobi, line 79: I think it would be cleaner to use an optional instead:
```suggestion
  const std::optional<location> loc_;
```

suyuee: Yeah, looks cool.

suyuee:
Use std::optional.

PR #1421, opened 2020-07-15
Fix some runtime tests relying on vfs_read

suyuee:
Some runtime tests in builtin and probe expect the system to generate vfs_read
events for them, which in some cases might cause timeout error. This PR
explicitly generates the events for these tests with testprog/syscall.

##### Checklist

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -2,11 +2,13 @@ NAME kprobe
 RUN bpftrace -v -e 'kprobe:vfs_read { printf("SUCCESS '$test' %d\n", pid); exit(); }'
 EXPECT SUCCESS kprobe [0-9][0-9]*
 TIMEOUT 5
+AFTER ./testprogs/syscall read
 
 NAME kprobe_short_name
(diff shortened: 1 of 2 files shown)


tests/runtime/probe
```diff
@@ -23,21 +25,25 @@ NAME kprobe_offset
 RUN bpftrace -v -e 'kprobe:vfs_read+0 { printf("SUCCESS '$test' %d\n", pid); exit(); }'
 EXPECT SUCCESS kprobe_offset [0-9][0-9]*
 TIMEOUT 5
+AFTER ./testprogs/syscall read
 
 NAME kprobe_offset_fail_size
 RUN bpftrace -v -e 'kprobe:vfs_read+1000000 { printf("SUCCESS '$test' %d\n", pid); exit(); }'
 EXPECT Offset outside the function bounds \('vfs_read' size is*
 TIMEOUT 5
+AFTER ./testprogs/syscall read
```
danobi, line 34: This bpftrace invocation should fail before it even loads so I don't think we need it here

suyuee: Yeah, I'd agree.

PR #1425, opened 2020-07-16
codegen: Only run codegen tests for llvm 7

fbs:
Looking at the way our codegen tests are used now I don't think testing
multiple LLVM versions has a clear benefit for us. The codegen tests are
already too large to spot subtle bugs, instead they are a way to
verify that you (small) change doesn't trigger unexpected changes. We
don't have any version specific code paths either (excluding the
function signature changes).

Limiting codegen tests to just LLVM7 frees us from maintaining
version specific tests and the codegen rewriter (which turned out a bit
more work than I originally hoped for).


danobi:
So my development machines are fedora (which I keep pretty up to date, so 32 right now) and arch linux (which is always up to date). Both distros are shipping llvm 10. Does that mean with this change I won't be able to run codegen tests locally?

fbs:
Hmm fair point. I keep llvm7 installed on fedora 32 as it is the easiest way to update codegen tests when needed. But if you just want to run codegen tests to be sure it is a bit annoying to have to rebuild onto LLVM7 first.

danobi:
Would it be possible to base codegen off the latest llvm release? So 10 in this case. Is there anything special about 7 other than most of the generated IR is currently 7?

fbs:
> Would it be possible to base codegen off the latest llvm release? So 10 in this case. Is there anything special about 7 other than most of the generated IR is currently 7?

No, I picked 7 at the time as its available on most distros and iirc we didn't have 9/10 support either. 

With llvm (and fedora) doing 2 releases a year we will always be behind on them, so you'd probably always end up running a not-latests version on your dev system.

I'll take a look at how other llvm based projects deal with multiple versions, there must be a better way.

danobi:
> With llvm (and fedora) doing 2 releases a year we will always be behind on them, so you'd probably always end up running a not-latests version on your dev system.

I think arch linux keeps up really close with upstream releases. But yeah, you're right, in general doing constant catch up is gonna be a PITA long term.

> I'll take a look at how other llvm based projects deal with multiple versions, there must be a better way.

Yeah, curious how rust deals with it. I remember @yonghong-song mentioned that llvm tests use wildcards or regex patterns or something. You think that'd work for us? Instead of rewriting into llvm 7 format maybe we could wildcard anything that isn't uniform across versions?

PR #1426, opened 2020-07-17
Add feature specific runtime testcase support

mmisono:
This adds `REQUIRES_FEATURE` field for a runtime test script. This allows
a feature specific testcase without manually writing
`REQUIRES bpftrace --info | grep xxxx`. The current supported features
are "loop" and "btf".

WIth #1424, btf related tests won't be run because the condition
"REQUIRES bpftrace --info 2>&1 | grep "btf: yes" never match. This fixes it.

##### Checklist

### DIFF
--- a/tests/runtime/engine/utils.py
+++ b/tests/runtime/engine/utils.py
@@ -7,6 +7,7 @@ import time
 from os import environ, uname, devnull
 from distutils.version import LooseVersion
 import re
+from functools import lru_cache
 
 BPF_PATH = environ["BPFTRACE_RUNTIME_TEST_EXECUTABLE"]
 ENV_PATH = environ["PATH"]
@@ -42,6 +43,7 @@ class Utils(object):
     TIMEOUT = 3
     SKIP_REQUIREMENT_UNSATISFIED = 4
     SKIP_ENVIRONMENT_DISABLED = 5
+    SKIP_FEATURE_REQUIREMENT_UNSATISFIED = 6
 
     @staticmethod
     def failed(status):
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to
 - Fix `printf` not allowing format specifiers to be directly followed by
   alphabetic characters
   - [#1414](https://github.com/iovisor/bpftrace/pull/1414)
+- Add feature specific runtime testcase support
```
fbs, line 17: I don't think we should add internal things like these test changes to the changelog as they'ree not relevant to end users.

tests/runtime/engine/utils.py
```diff
@@ -105,6 +109,29 @@ def run_test(test):
                         print(warn("[   SKIP   ] ") + "%s.%s" % (test.suite, test.name))
                         return Utils.SKIP_REQUIREMENT_UNSATISFIED
 
+            if test.feature_requirement:
+                cmd = "bpftrace --info"
+                p = subprocess.Popen(
```
fbs, line 114: nit: maybe we can cache this output

CHANGELOG.md
mmisono: OK, I'll remove it.

tests/runtime/engine/utils.py
mmisono: I'll try it.

mmisono:
The test failed due to the connection error. It passed on my branch: https://github.com/mmisono/bpftrace/actions/runs/175079253

danobi:
Awesome

Issue #614, opened 2019-05-09
[RFC] prepare for bpf_probe_read() split into user/kernel


brendangregg:
As was discussed at LSFMM, bpf_probe_read() may be split in the future into bpf_probe_read_kernel() and bpf_probe_read_user(), to support other architectures (like SPARC) where the kernel/user address space overlaps, and the pointer isn't sufficient to identify the mode. This will be a major change.

We currently read memory addresses in at least these two ways:

- `*addr`: dereference
- `str(addr)`: fetch NULL-terminated string

bpftrace will need a way to determine whether to call the _kernel() or _user() version of bpf_probe_read(). Here's what I propose, which breaks the fewest tools:

- `*addr`: dereference kernel
- `str(addr)`: fetch NULL-terminated kernel string
- `ucopy(addr)`: dereference user
- `ustr(addr)`: fetch NULL-terminated user string

Right now, ucopy() and ustr() would be dummy functions that do nothing, and later on can be switched to call bpf_probe_read_user(). Adding these dummy functions now will help us prevent breaking the bpftrace API in the future.

I picked "u"something() since we have other user functions that start with "u". ucopy() is short for copy_from_user(), and ustr() is the user version of str(). I think these names make sense.

If we are in agreement, we should add these dummy functions sooner rather than later for API stabilization.

Note that #28 calls for a uwrite() function, so if we want symmetry then we could call it uread() instead of ucopy().

Edit: this is option A.

ajor:
If we're in a kprobe/kernel tracepoint will all pointers be to kernel memory, and vice versa for uprobes/usdts?

brendangregg:
@ajor Ah right, this was suggested as one possible solution. I'll check all my tools to see how feasible that is.

Following that approach, we could leave these with their current syntax and the following meanings, with their behavior changing when bpf_probe_read() is split:

- `*addr`: dereference event-context memory (kernel/user based on event)
- `str(addr)`: fetch NULL-terminated event-context string (kernel/user based on event)

And add these optional qualified versions (going with the *read() version here):

- `kread(addr)`: dereference kernel
- `kstr(addr)`: fetch NULL-terminated kernel string
- `uread(addr)`: dereference user
- `ustr(addr)`: fetch NULL-terminated user string

danobi:
What about if you kprobed
```
ssize_t ksys_write(unsigned int fd, const char __user *buf, size_t count);
```
And tried to read from `buf`?

brendangregg:
[...]

PR #1427, opened 2020-07-17
Addrspace basics

fbs:
To correctly read data on architectures with overlapping address spaces `probe_read` got split up in a `_kernel` and a `_user` variant. As we're responsible for selecting the right helper we have to keep track of the address space a pointer is in.

This PR implements the basics  required to do so, it;

- updates the pointer type to support address space information
- extends the probe_read irbuilder methods to select the right helper based on the addrspace info passed
- extends the ast printer to include addrspace info
- adds 2 new calls, kptr() and uptr() that change the addrspace of a pointer. This is left undocumented for now.

I initially tried to implement this in one go (#1141) but that had some corner cases that where hard to fix. Getting the basics in first makes reviewing easier and makes it easier for others to help out with it.

---

### warning

as this depends on a new kernel  the new behaviour won't be covered by the runtime tests in the ci

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -223,6 +223,7 @@ void CodegenLLVM::visit(Builtin &builtin)
                                           arg_num,
                                           builtin,
                                           bpftrace_.pid(),
+                                          AddrSpace::none,
                                           builtin.loc);
         return;
       }
@@ -266,7 +267,7 @@ void CodegenLLVM::visit(Builtin &builtin)
     Value *src = b_.CreateAdd(sp,
                               b_.getInt64((arg_num + arch::arg_stack_offset()) *
                                           sizeof(uintptr_t)));
-    b_.CreateProbeRead(ctx_, dst, 8, src, builtin.loc);
+    b_.CreateProbeRead(ctx_, dst, 8, src, builtin.type.GetAS(), builtin.loc);
     expr_ = b_.CreateLoad(dst);
     b_.CreateLifetimeEnd(dst);
   }
@@ -528,8 +529,10 @@ void CodegenLLVM::visit(Call &call)
     }
     AllocaInst *buf = b_.CreateAllocaBPF(bpftrace_.strlen_, "str");
     b_.CREATE_MEMSET(buf, b_.getInt8(0), bpftrace_.strlen_, 1);
+    auto arg0 = call.vargs->front();
     auto scoped_del = accept(call.vargs->front());
-    b_.CreateProbeReadStr(ctx_, buf, b_.CreateLoad(strlen), expr_, call.loc);
+    b_.CreateProbeReadStr(
+        ctx_, buf, b_.CreateLoad(strlen), expr_, arg0->type.GetAS(), call.loc);
     b_.CreateLifetimeEnd(strlen);
 
     expr_ = buf;
@@ -584,10 +587,12 @@ void CodegenLLVM::visit(Call &call)
                      1);
 
     auto scoped_del = accept(call.vargs->front());
+    auto arg0 = call.vargs->front();
     b_.CreateProbeRead(ctx_,
                        static_cast<AllocaInst *>(buf_data_offset),
                        length,
                        expr_,
+                       arg0->type.GetAS(),
                        call.loc);
 
     expr_ = buf;
@@ -620,7 +625,9 @@ void CodegenLLVM::visit(Call &call)
   }
   else if (call.func == "join")
   {
-    auto scoped_del = accept(call.vargs->front());
+    auto arg0 = call.vargs->front();
+    auto scoped_del = accept(arg0);
+    auto addrspace = arg0->type.GetAS();
     AllocaInst *first = b_.CreateAllocaBPF(b_.getInt64Ty(),
                                            call.func + "_first");
     AllocaInst *second = b_.CreateAllocaBPF(b_.getInt64Ty(),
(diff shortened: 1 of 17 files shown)


fbs:
ready for review  @sumanthkorikkar, @danobi , @mmisono

sumanthkorikkar:
Will check. Thanks

src/ast/irbuilderbpf.h
```diff
@@ -96,9 +165,16 @@ class IRBuilderBPF : public IRBuilder<>
...
+                                const location &loc);
   CallInst   *createMapLookup(int mapfd, AllocaInst *key);
-  Constant   *createProbeReadStrFn(llvm::Type * dst, llvm::Type * src);
+  Constant *createProbeReadStrFn(llvm::Type *dst,
+                                 llvm::Type *src,
+                                 AddrSpace as);
+  libbpf::bpf_func_id selectProbeReadHelper(AddrSpace as, bool str);
 
   std::map<std::string, StructType *> structs_;
   // clang-format on
```
danobi, line 180: Delete this too?

src/ast/codegen_llvm.cpp
```diff
@@ -985,6 +985,11 @@ void CodegenLLVM::visit(Call &call)
     expr_ = b_.CreateIntCast(expr_, b_.getInt64Ty(), arg.type.IsSigned());
     b_.CreateOverrideReturn(ctx_, expr_);
   }
+  else if (call.func == "kptr" || call.func == "uptr")
+  {
+    auto &arg = *call.vargs->at(0);
+    arg.accept(*this);
```
danobi, line 991:
```suggestion
    auto arg = call.vargs->at(0);
    accept(arg);
```

sumanthkorikkar:
In my opinion, Having runtime-tests would be great for all platforms. May be we should work on the subsequent commits after this integration.

mmisono:
It might be good to add FieldAccess support (e.g., `kptr(args)->x`) to this PR. Otherwise, we cannot do any useful things.

fbs:
> In #614 the plan was to have default addr spaces for pointers in certain probes (ie kprobe ptr args are kernel, uprobe ptr args are user, etc). Is the plan still to do that? We might be able to get away with not doing that and let sparc/ppc/s390x use the kptr and uptr calls.

My concern is that if we don't do it bpftrace would be CPU dependent, people would have to maintain patched tools and it would make non-x86 archs even more a 2nd class citizen.

Overal I don't think it will be *that* hard to properly implement it. The original PR suffered from the type system, it should be easier now. 

> In my opinion, Having runtime-tests would be great for all platforms. May be we should work on the subsequent commits after this integration.

I'll add some basic tests, even if though they don't run in the CI they will be useful.

> It might be good to add FieldAccess support (e.g., `kptr(args)->x`) to this PR. Otherwise, we cannot do any useful things.

Hmm ill look into it

danobi:
> Overal I don't think it will be that hard to properly implement it. The original PR suffered from the type system, it should be easier now.

Just to confirm: this PR doesn't implement it but the plan is to follow up with it, right?

danobi:
>> In my opinion, Having runtime-tests would be great for all platforms. May be we should work on the subsequent commits after this integration.
>>
> I'll add some basic tests, even if though they don't run in the CI they will be useful.

@sumanthkorikkar I think GH actions supports hosted runners ( https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners ). This could be one way to get additional architecture support for the CI.

I don't think any of us mind if you come and fix regressions after they occur but you could probably save yourself work in the long term if we can find a way to upstream your CI setup.

fbs:
> > Overal I don't think it will be that hard to properly implement it. The original PR suffered from the type system, it should be easier now.
> 
> Just to confirm: this PR doesn't implement it but the plan is to follow up with it, right?

Yes this just the basics framework to keep track of the AS and to get some agreement on how it should look.

sumanthkorikkar:
Hi @fbs,

Few things:
- Tested few cases on s390x. kptr() and uptr() resolves to bpf_probe_read_kernel() and bpf_probe_read_user.
- -v does not seems to show the bpf bytecode. It is shown only during error. This is not relevant to this PR. However, It makes me to use bpftool each time to see the bpf bytecode.
- Tried to change bashreadline.bt 
    * printf("%-6d %s\n", pid, str(uptr(retval)));
    * ERROR: uptr() only supports pointer arguments (integer provided)
    *  printf("%-6d %s\n", pid, str(uptr((uint32 *)retval))); works
       * Do we have somekind of auto type resolution available in the future.?
- Now all the deferences without kptr() or uptr() are bpf_probe_read(), [ which just fails on s390x, but this is for good purpose ].

fbs:
>     * Tested few cases on s390x. kptr() and uptr() resolves to bpf_probe_read_kernel() and bpf_probe_read_user, great!!
> 
>     * -v does not seems to show the bpf bytecode. It is shown only during error. This is not relevant to this PR. However, It makes me to use bpftool each time to see the bpf bytecode.
> 

Weird, cannot reproduce this

```
sudo bpftrace -e 'k:f { @=kptr((int8*)123) }' -v
Attaching 1 probe...

Program ID: 2481

Bytecode:
0: (b7) r1 = 0
1: (7b) *(u64 *)(r10 -16) = r1
...
Running...
^C
````

>     * Tried to change bashreadline.bt
>       
>       * printf("%-6d %s\n", pid, str(uptr(retval)));
>       * ERROR: uptr() only supports pointer arguments (integer provided)
>       * printf("%-6d %s\n", pid, str(uptr((uint32 *)retval))); works
>         
>         * Do we have somekind of auto type resolution available in the future.?

I guess we can add it for the `str()` case, internally cast it to a `(char*)`.

sumanthkorikkar:
> > ```
> > * Tested few cases on s390x. kptr() and uptr() resolves to bpf_probe_read_kernel() and bpf_probe_read_user, great!!
> > 
> > * -v does not seems to show the bpf bytecode. It is shown only during error. This is not relevant to this PR. However, It makes me to use bpftool each time to see the bpf bytecode.
> > ```
> 
> Weird, cannot reproduce this
> 
ok. Then I can check this case...
> ```
> sudo bpftrace -e 'k:f { @=kptr((int8*)123) }' -v
> Attaching 1 probe...
> 
> Program ID: 2481
> 
> Bytecode:
> 0: (b7) r1 = 0
> 1: (7b) *(u64 *)(r10 -16) = r1
> ...
> Running...
> ^C
> ```
> 
> > ```
> > * Tried to change bashreadline.bt
> >   
> >   * printf("%-6d %s\n", pid, str(uptr(retval)));
> >   * ERROR: uptr() only supports pointer arguments (integer provided)
> >   * printf("%-6d %s\n", pid, str(uptr((uint32 *)retval))); works
> >     
> >     * Do we have somekind of auto type resolution available in the future.?
> > ```
> 
> I guess we can add it for the `str()` case, internally cast it to a `(char*)`.

ok.

fbs:
I'll merge this on friday (~40h from now) if there are no objections, that way @sumanthkorikkar can start building on this

PR #1429, opened 2020-07-19
codegen: add option to emit elf file

fbs:
While debugging it is sometimes useful to see the BPF created by LLVM.
Currently the only way to do that is to use bpftool to dump the program
after it has been loaded into the kernel, which is quite a lot of
hassle.

This adds a simple pass that emits an elf file. This pass can
(currently) only be called after the optimization pass as that does some
setup. To prevent some programmer errors I've added some assert() to
make sure all passes are called in the correct order.

LLVM 6 doesn't support the raw_fd_ostream API so that uses the svector
based one, that part is basically a copy of
`tools/perf/util/c++/clang.cpp` in the linux kernel.

This new flag hasn't been documented as I don't want to expose it to
users.

### DIFF
--- a/scripts/compare_tool_elf.sh
+++ b/scripts/compare_tool_elf.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Compare the IR generated for the shipped
+# Compare the ELF generated for the shipped
 # tools between two bpftrace builds
 #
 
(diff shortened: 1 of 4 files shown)


fbs:
Example: 
```
sudo bpftrace /vagrant/tools/loads.bt  --emit-elf /tmp/loads.elf -d >/dev/null; llvm-objdump-6.0 -S /tmp/loads.elf | grep section -A5
Disassembly of section s_BEGIN_1:
BEGIN:
       0:	bf 16 00 00 00 00 00 00 	r6 = r1
       1:	b7 01 00 00 00 00 00 00 	r1 = 0
       2:	7b 1a f8 ff 00 00 00 00 	*(u64 *)(r10 - 8) = r1
       3:	18 17 00 00 01 00 00 00 00 00 00 00 00 00 00 00 	ld_pseudo	r7, 1, 1
--
Disassembly of section s_interval:s:1_1:
interval:s:1:
       0:	7b 1a c0 ff 00 00 00 00 	*(u64 *)(r10 - 64) = r1
       1:	bf a1 00 00 00 00 00 00 	r1 = r10
       2:	07 01 00 00 c8 ff ff ff 	r1 += -56
       3:	b7 02 00 00 08 00 00 00 	r2 = 8
```

fbs:
As it's supposed to be a debug tool for devs I don't think we should put it in the help or refguide. If we had a proper manpage that would be the place. I should start on that man page..

PR #1431, opened 2020-07-21
[RFC] Link to discourse server

danobi:
Discourse is a somewhat popular open source forum software. For example,
rust uses it: https://users.rust-lang.org/ .

I think it'd be nice if we had a bpftrace forum so commonly asked questions
can be catalogued and searched. It seems like discourse also supports
mailing lists so email-lovers can continue with that if they'd like. It also looks
like discourse has support for admin/moderator boards, so we can have
more private contributor discussions there instead of via email.


fbs:
Maybe we should add a `question/issue`  issue template too that asks people to go to discourse when its not a bug/feature?

danobi:
> Maybe we should add a question/issue issue template too that asks people to go to discourse when its not a bug/feature?

Yeah that's a good idea

danobi:
It looks like this now:
issue menu

PR #1432, opened 2020-07-22
Disable failing kfunc

### LINKED ISSUE #1366: Inline functions are listed as traceable in kfunc
### bpftrace --info
```
System
OS: Linux 5.7.0+ SMP Mon Jun 1 12:50:01 JST 2020
Arch: x86_64

Build
version: v0.10.0-151-g7a73
LLVM: 10
foreach_sym: yes
unsafe uprobe: no
bfd: yes
bpf_attach_kfunc: yes

Kernel helpers
probe_read: yes
probe_read_str: yes
probe_read_user: yes
probe_read_user_str: yes
probe_read_kernel: yes
probe_read_kernel_str: yes
get_current_cgroup_id: yes
send_signal: yes
override_return: yes

Kernel features
Instruction limit: 1000000
Loop support: yes
btf: yes

Map types
hash: yes
percpu hash: yes
array: yes
percpu array: yes
stack_trace: yes
perf_event_array: yes

Probe types
kprobe: yes
tracepoint: yes
perf_event: yes
kfunc: yes
```

### What reproduces the bug?
```
% sudo ./src/bpftrace -lv 'f:vfs_stat_set_lookup_flags'
kfunc:vfs_stat_set_lookup_flags
    unsigned int * lookup_flags;
    int flags;
    unsigned int retval;
```
```
% sudo ./src/bpftrace -e 'f:vfs_stat_set_lookup_flags {}'
Attaching 1 probe...
bpf_attach_raw_tracepoint (kfunc): Device or resource busy
Error attaching probe: kfunc:vfs_stat_set_lookup_flags
```

`vfs_stat_set_lookup_flags()` is [a inline function](https://github.com/torvalds/linux/blob/f359287765c04711ff54fbd11645271d8e5ff763/fs/stat.c#L156), so I think it is not traceable (the symbol name is not presented in /sys/kernel/debug/tracing/available_filter_functions, but BTF have its definition). I think bpftrace should filter out these functions.

------

I found this when trying (#1347 is needed to use wildcard in kfunc)

```
% sudo ./src/bpftrace -e 'f:vfs_* { @[probe] = count(); }'
Attaching 63 probes...
bpf_attach_raw_tracepoint (kfunc): Device or resource busy
Error attaching probe: kfunc:vfs_stat_set_lookup_flags
```

CC: @olsajiri

viktormalik:
There is a number of functions that are listed as traceable in kfunc, but their tracing crashes bpftrace. This PR disables tracing of 2 kinds of such functions:
1. Functions not listed in `/sys/kernel/debug/tracing/available_filter_functions` (e.g. inlined functions or functions marked as "notrace"). These should not be traceable even though there is BTF info for them (which makes them appear on the kfunc list). This fixes #1366.
2. Functions having more than 6 parameters. Kernel currently doesn't support BPF trampolines for functions with more than 6 parameters.

##### Checklist

### DIFF
--- a/src/btf.h
+++ b/src/btf.h
@@ -44,9 +44,11 @@ private:
   std::unique_ptr<std::istream> get_funcs(std::regex* re,
                                           bool params,
                                           std::string prefix) const;
+  bool is_traceable_func(const std::string& func_name) const;
 
   struct btf* btf;
   enum state state = NODATA;
+  std::unordered_set<std::string> traceable_funcs_;
 };
 
 inline bool BTF::has_data(void) const
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -13,6 +13,7 @@
 #include <tuple>
 #include <unistd.h>
 
+#include "list.h"
 #include "utils.h"
 #include <bcc/bcc_elf.h>
 
@@ -737,6 +738,33 @@ std::string hex_format_buffer(const char *buf, size_t size)
   s[offset] = '\0';
 
   return std::string(s);
+}
+
+std::unordered_set<std::string> get_traceable_funcs()
+{
+  // Try to get the list of functions from BPFTRACE_AVAILABLE_FUNCTIONS_TEST env
+  const char *path = std::getenv("BPFTRACE_AVAILABLE_FUNCTIONS_TEST");
+
+  // Use kprobe list as default
+  if (!path)
+    path = kprobe_path.c_str();
+
+  std::ifstream available_funs(path);
+  if (available_funs.fail())
+  {
+    if (bt_debug != DebugLevel::kNone)
+    {
+      std::cerr << "Error while reading traceable functions from "
+                << kprobe_path << ": " << strerror(errno);
+    }
+    return {};
+  }
+
+  std::unordered_set<std::string> result;
+  std::string line;
+  while (std::getline(available_funs, line))
+    result.insert(line);
+  return result;
 }
 
 } // namespace bpftrace
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to
 #### Deprecated
 
 #### Removed
+- Disable some kfunc probes whose tracing crashes
+  - [#1432](https://github.com/iovisor/bpftrace/pull/1432)
 
 #### Tools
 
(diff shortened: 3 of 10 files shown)


mmisono:
Thanks for fixing this. Looks good to me.

src/btf.h
```diff
@@ -44,9 +44,11 @@ class BTF
   std::unique_ptr<std::istream> get_funcs(std::regex* re,
                                           bool params,
                                           std::string prefix) const;
+  bool is_traceable_func(std::string& func_name) const;
```
danobi, line 47: Nit
```suggestion
  bool is_traceable_func(const std::string& func_name) const;
```

src/utils.cpp
```diff
@@ -739,4 +740,31 @@ std::string hex_format_buffer(const char *buf, size_t size)
...
+  if (!path)
+    path = kprobe_path.c_str();
+
+  std::ifstream available_funs(path);
+  if (available_funs.fail())
+  {
+    if (bt_debug != DebugLevel::kNone)
+    {
+      std::cerr << "Error while reading traceable functions from "
+                << kprobe_path << ": " << strerror(errno);
```
danobi, line 758: Hmm, this seems unlikely to fail. But when it does fail, kfuncs are disabled. I think this message would help user figure out what went wrong in that case. Not sure if it should be hidden behind `-d` flag.

viktormalik: That's true, although I don't think that the message should be shown every time the user lists probes and the file is not available. I'd suggest leaving it behind `-d` flag and displaying a similar message when user tries to attach a kfunc and it fails due to this error. I'd implement that as a separate PR that will improve error messages for kfuncs in general.

fbs: Why not raise an exception? That way the caller can figure out what to do with it which makes more sense for a "library" function like this imo.

If kfuncs depend on this we should probably hook it into `--info` too, would be confusing if bpftrace reports that kfuncs should work when it doesn.t

viktormalik: > Why not raise an exception? That way the caller can figure out what to do with it which makes more sense for a "library" function like this imo.

I agree that raising an exception is cleaner, however, the caller is always the same - the constructor of `BTF`, called from the constructor of `BPFtrace`, so we would end up with the same problem again. A solution could be to postpone reading of the file to the phase when the function list is actually queried, but the function doing it (`BTF::get_funcs`) is marked as `const` and I didn't want to break that.

> If kfuncs depend on this we should probably hook it into `--info` too, would be confusing if bpftrace reports that kfuncs should work when it doesn.t

I agree, will do.

PR #1433, opened 2020-07-22
Add a semantic warning for printing stats maps with top/div args

suyuee:
Both `top` and `div` are ambiguous when printing stats maps. `top` does not
specify which statistic it's based on and it would cause confusion to define a
default behavior for it. As for `div`, it would also somehow be strange: if use
`div` on all three statistics, then the output does not have avg * count == sum
anymore; if only on sum and avg, it might be misleading. 

Related to #1416.

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -836,6 +836,12 @@ void SemanticAnalyser::visit(Call &call)
             check_arg(call, Type::integer, 1, true);
           if (call.vargs->size() > 2)
             check_arg(call, Type::integer, 2, true);
+          if (map.type.IsStatsTy() && call.vargs->size() > 1)
+          {
+            warning("print()'s top and div arguments are ignored when used on "
+                    "stats() maps.",
+                    call.loc);
+          }
         }
       }
       // Note that IsPrintableTy() is somewhat disingenuous here. Printing a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to
   - [#1416](https://github.com/iovisor/bpftrace/pull/1416)
 
 #### Changed
+- Warn if using `print` on `stats` maps with top and div arguments
+  - [#1433](https://github.com/iovisor/bpftrace/pull/1433)
 
 #### Deprecated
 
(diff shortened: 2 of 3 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -836,6 +836,12 @@ void SemanticAnalyser::visit(Call &call)
             check_arg(call, Type::integer, 1, true);
           if (call.vargs->size() > 2)
             check_arg(call, Type::integer, 2, true);
+          if (map.type.IsStatsTy() && call.vargs->size() > 1)
+          {
+            warning("When printing stats maps, arguments top and div are "
```
danobi, line 841: My gut feeling is this should be an error. Are there advantages to keeping it a warning?

suyuee:
Something I'm not sure of:
1. If this should be recorded in CHANGLOG.md? If so, which section it belongs to then.
2. I make it a warning, not sure if it's strong enough.

src/ast/semantic_analyser.cpp
suyuee: Actually I felt the same way at first, but I somehow felt it was too strong since it has always been allowed. But I guess error is more reasonable. Btw should this pr be included in CHANGELOG.md? I don't know if there is a section it belongs to.

danobi: That's a good point about not breaking previous programs. Yeah let's keep it as a warning.

And yeah there should probably be a "Changed" entry

src/ast/semantic_analyser.cpp
```diff
@@ -836,6 +836,12 @@ void SemanticAnalyser::visit(Call &call)
             check_arg(call, Type::integer, 1, true);
           if (call.vargs->size() > 2)
             check_arg(call, Type::integer, 2, true);
+          if (map.type.IsStatsTy() && call.vargs->size() > 1)
+          {
+            ERR("Arguments top and div are disallowed when printing stats maps "
+                "because of the ambiguity.",
```
danobi, line 842: Strange to say "disallowed" when it's actually allowed. Maybe the following wording?

```
print()'s top and div arguments are ignored when used on stats() maps
```

suyuee: Eh, I thought we would use `error` instead of `warning` so I changed from 'ignored' to 'disallow'. Sorry about the mess.

suyuee:
Updated CHANGELOG.md

PR #1437, opened 2020-07-23
Fix: print() no output when used on hist() maps with large top args

suyuee:
Currently, using `print` with a large `top` arg (any number larger than number
of map keys) on `hist` maps would result in a negative overflow bug, which will
make `print` fail to produce any output. This patch resolves this issue.
```
output.cpp
void TextOutput::map_hist
{
...
    if (top)
    {
        std::cerr << "values_by_key.size()       = " << values_by_key.size()       << std::endl;
        std::cerr << "                       top = " <<                        top << std::endl;
        std::cerr << "values_by_key.size() - top = " << values_by_key.size() - top << std::endl;
        if (i++ < (values_by_key.size() - top))
            continue;
    }
...
}

// output
# bpftrace -e 'kretprobe:vfs_read { @bytes = hist(retval); } i:s:5 {print(@bytes, 2, 1); exit();} END { clear(@bytes); }'
values_by_key.size()       = 1
                       top = 2
values_by_key.size() - top = 18446744073709551615
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to
 #### Removed
 
 #### Fixed
+- Fix `print` outputs nothing when used on hist() maps with large top args
+  - [#1437](https://github.com/iovisor/bpftrace/pull/1437)
 
 #### Tools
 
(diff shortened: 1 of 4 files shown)


suyuee:
Update CHANGLOG.md

Issue #954, opened 2019-11-08
BEGIN probe not working


danobi:
```
$ sudo ./src/bpftrace -e 'BEGIN { printf("hello\n"); }'
Attaching 1 probe...
Could not resolve symbol: /proc/self/exe:BEGIN_trigger
```

Built from master at time of writing.

danobi:
So that was on my fedora box. It works ok on my arch linux box.

caringi:
On my fedora it's working...

```
$ cat /etc/fedora-release 
Fedora release 30 (Thirty)
$ uname -r
5.2.9-200.fc30.x86_64+debug
$ git rev-parse HEAD
2e73e04e27c439cf87d3687be24c77cce2cf65ae
$ sudo ./src/bpftrace -e 'BEGIN { printf("hello\n"); }'
Attaching 1 probe...
hello
^C
```

danobi:
More info on my box:
```
$ cat /etc/fedora-release
Fedora release 30 (Thirty)
$ uname -r
5.2.16-200.fc30.x86_64
```

brendangregg:
Here's a clue:

/usr/local/bin/bpftrace doesn't work:

```
(root) /mnt/src/bpftrace/build # bpftrace -e 'BEGIN { @++ }'
Attaching 1 probe...
Could not resolve symbol: /proc/self/exe:BEGIN_trigger
(root) /mnt/src/bpftrace/build # which bpftrace
/usr/local/bin/bpftrace
```

But build/src/bpftrace does:

```
(root) /mnt/src/bpftrace/build # ./src/bpftrace -e 'BEGIN { @++ }'
Attaching 1 probe...
^C

@: 1
```

They are the same version(!):

```
(root) /mnt/src/bpftrace/build # bpftrace --version
bpftrace v0.9.2-154-gd95a

(root) /mnt/src/bpftrace/build # ./src/bpftrace --version
bpftrace v0.9.2-154-gd95a
```

but they are not quite the same binary:

```
(root) /mnt/src/bpftrace/build # file /usr/local/bin/bpftrace src/bpftrace 
/usr/local/bin/bpftrace: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/l, for GNU/Linux 3.2.0, BuildID[sha1]=9a58fc11648c14151a3c49d8a7cc592bf21bd16b, with debug_info, not stripped
[...]

PR #1438, opened 2020-07-24
Ensure symbols are exported for bpftrace executable

vincentbernat:
bpftrace executable needs to lookup `BEGIN_trigger` and `END_trigger`
symbols. Since the bump of the minimal versin of CMake in
1a1a4a92f7bcb154282df228d080d5c3083e04f8, these symbols are stripped
from bpftrace: the executable is built without the `-rdynamic` option
and the symbol is not present anymore (tested with GCC 10). To restore
this compile option, we can instruct CMake to export symbols for
bpftrace executable.

Also, see #954.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,6 +29,7 @@ add_executable(bpftrace
   utils.cpp
   ${BFD_DISASM_SRC}
 )
+set_property(TARGET bpftrace PROPERTY ENABLE_EXPORTS 1)
 
 if(HAVE_NAME_TO_HANDLE_AT)
   target_compile_definitions(bpftrace PRIVATE HAVE_NAME_TO_HANDLE_AT=1)


dalehamel:
i think this broke master for embedded builds, just looking at CI history, and also suspicious is linked by #1440 

Maybe we need to back this out while we figure it out. Is this fix just needed for GCC 10?

> from bpftrace: the executable is built without the -rdynamic

I think we don't have this by default in the embedded builds, but i'll have to check the flags.

danobi:
Yeah, let's revert for now so we get CI back

danobi:
Revert in #1441 1441

vincentbernat:
❦ 28 juillet 2020 11:20 -07, Dale Hamel:

> Maybe we need to back this out while we figure it out. Is this fix just needed for GCC 10?

If useful, I can test with GCC 9.

PR #1441, opened 2020-07-28
Revert "Ensure symbols are exported for bpftrace executable"

danobi:
This reverts commit 38f1c69b38b5f53969883d9cd0fd77a3dfb52ecc.

It was breaking the embedded builds.

##### Checklist

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,7 +29,6 @@ add_executable(bpftrace
   utils.cpp
   ${BFD_DISASM_SRC}
 )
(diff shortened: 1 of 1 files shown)


dalehamel:
:+1: let's just ensure it indeed fixes CI then we can figure out a proper fix-forward

dalehamel:
looking at https://cmake.org/cmake/help/v3.14/prop_tgt/ENABLE_EXPORTS.html it's not clear to me why this would break things either, and @vincentbernat 's logic in #1438 does seem correct, so there's some dragons here somewhere...

PR #1442, opened 2020-07-28
ci: Run embedded build on PRs

danobi:
The embedded build / test takes a while to run but it's probably better
than having master randomly break. Every time that breaks it takes at
least a couple days to get a fix up.

##### Checklist


dalehamel:
we should expect this to fail on embedded builds now since they are broken on master without #1441

PR #1443, opened 2020-07-29
Move logging out of `class BPFtrace`

suyuee:
Logging in bpftrace currently relies on facilities provided by class BPFtrace.
Now that we have a new logging infrastructure #1418, we can change the callsites
to new logging APIs. This patch helps with breaking up class BPFtrace #1080 and
improving logging in bpftrace #1411 .

##### Checklist

### DIFF
--- a/src/tracepoint_format_parser.cpp
+++ b/src/tracepoint_format_parser.cpp
@@ -4,9 +4,10 @@
 #include <iostream>
 
 #include "ast.h"
+#include "bpftrace.h"
+#include "log.h"
 #include "struct.h"
 #include "tracepoint_format_parser.h"
-#include "bpftrace.h"
 
 namespace bpftrace {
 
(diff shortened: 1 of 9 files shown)


mmisono:
Can we also use this logging infrastructure where not using class bpftrace logging (i.e., in utils.cpp, semantic analyzer, codegen, etc.)?

suyuee:
> Can we also use this logging infrastructure where not using class bpftrace logging (i.e., in utils.cpp, semantic analyzer, codegen, etc.)?

Sure. That's part of the plan. I hope to leave those to another pr since there are quite some callsites of std::cerr/cout and I'm not sure if all these should be replaced.

src/tracepoint_format_parser.cpp
```diff
@@ -109,18 +105,15 @@ bool TracepointFormatParser::parse(ast::Program *program, BPFtrace &bpftrace)
...
             // helper message:
             if (category == "syscall")
-              bpftrace.warning(std::cerr,
-                               ap->loc,
-                               "Did you mean syscalls:" + event_name + "?");
+              LOG(WARNING, ap->loc, std::cerr)
+                  << "Did you mean syscalls:" << event_name << "?";
             if (bt_verbose) {
               // Having the location info isn't really useful here, so no
               // bpftrace.error
```
mmisono, line 119: I guess you are thinking to fix this in another PR, but here wen can use `LOG(DEBUG, std::cerr)`.

mmisono, line 119: I guess you are thinking to fix this in another PR, but I think we can use `LOG(DEBUG, std::cerr)` here.

javierhonduco:
Nice one! LGTM

Issue #1267, opened 2020-04-15
How to disable semantic analyser warnings?


ex3me0:
**version:** `bpftrace  v0.9.4-97-20200317-1417-gef39ed0` installed via snap

Hey
Is it possible to disable such kind of warnings, or what u can advice?
I have updated from bpftrace 0.9.1 to 0.9.4 and these messages are slightly annoying 
```
script.bt:7:7-9: WARNING: comparison of integers of different signs: 'unsigned int64' and 'int64' can lead to undefined behavior
/ pid == $1 /
```

`script.bt` example
```
tracepoint:syscalls:sys_enter_read,
tracepoint:syscalls:sys_enter_write
/ pid == $1 /
{
  @fd = args->fd;
  @buf_raw = args->buf;
  @size = args->count;
}
```

fbs:
There currently is none, but it would be nice to have :)

This particular warning should be fixed in the 0.10 release btw

mmisono:
To fix the warning, you can cast a value.

```
tracepoint:syscalls:sys_enter_read,
tracepoint:syscalls:sys_enter_write
/ pid == (uint64)$1 /
{
  @fd = args->fd;
  @buf_raw = args->buf;
  @size = args->count;
}
```

(in this case, upgrading bpftrace would be better though)

ex3me0:
Got it, thanks for your support!

PR #1444, opened 2020-07-29
Logging: add an option to disable warning messages.

suyuee:
It would be nice for bpftrace to provide an option which hides warning messages.
This patch proposes using `-w` to diable warnings, just like in gcc.
Related to #1267 and #1411 .
##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -442,12 +442,14 @@ open path: .com.google.Chrome.R1234s
 
 ## 8. Other Options
 
-The `--version` option prints the bpftrace version:
+- The `--version` option prints the bpftrace version:
 
 ```
 # bpftrace --version
 bpftrace v0.8-90-g585e-dirty
 ```
+
+- The `--no-warnings` option disables warnings.
 
 ## 9. Environment Variables
 
--- a/src/log.h
+++ b/src/log.h
@@ -3,6 +3,7 @@
 #include <iostream>
 #include <optional>
 #include <sstream>
+#include <unordered_map>
 
 #include "bpftrace.h"
 
@@ -41,8 +42,21 @@ public:
   Log(const Log& other) = delete;
   Log& operator=(const Log& other) = delete;
 
+  inline void enable(LogType type)
+  {
+    enabled_map_[type] = true;
+  }
+  inline void disable(LogType type)
+  {
+    enabled_map_[type] = false;
+  }
+  inline bool is_enabled(LogType type)
+  {
+    return enabled_map_[type];
+  }
+
 private:
-  Log() = default;
+  Log();
   ~Log() = default;
   std::string src_;
   std::string filename_;
@@ -50,6 +64,7 @@ private:
                          const location&,
                          std::ostream&,
                          const std::string&);
+  std::unordered_map<LogType, bool> enabled_map_;
 };
 
 class LogStream
@@ -67,7 +82,8 @@ public:
   template <typename T>
   LogStream& operator<<(const T& v)
   {
-    buf_ << v;
+    if (sink_.is_enabled(type_))
+      buf_ << v;
     return *this;
   }
   ~LogStream();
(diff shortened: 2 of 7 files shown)


suyuee:
Just realized that it relies on #1443 to actually take effects.

javierhonduco:
Ohhh, neat! LGTM!

Let's please add a runtime tests so we make sure the CLI (e.g. accidentally removing the argument parsing for this flag) won't regress!

suyuee:
Update: add runtime test

src/main.cpp
```diff
@@ -325,6 +328,9 @@ int main(int argc, char *argv[])
           return 1;
         }
         break;
+      case 'w':
```
fbs, line 331: a `-w` flag that disables warning logging feels a bit counter intuitive.  A long option like `--no-warnings`  feels more descriptive to me, I don't think we need a short flag for this (you probably should fix the warnings)

Others compilers have the `-Wxxx` and `-Wno-xxx` flags, would be nice to keep  a similar option open in the future when we add that kind of granularity to our warnings.

suyuee: Yeah, `-w` feels strange to me, too and I got the idea since it seems gcc does so. I think `--no-warnings` is better.
> (you probably should fix the warnings)

You mean the failing tests in embedded builds?

mmisono:
How about using log levels to control outputs? I guess many other logging libraries do that. (e.g., [python's logging](https://docs.python.org/3/howto/logging.html))

suyuee:
> How about using log levels to control outputs? I guess many other logging libraries do that. (e.g., [python's logging](https://docs.python.org/3/howto/logging.html))

That would look nice. But I think having an option to disable warnings for now is good. We can introduce log levels when we have the granularity. Currently, most loggings are ERROR and the rest are WARNINGS.

suyuee:
Update: change the option from `-w` to `--no-warnings`

docs/reference_guide.md
```diff
@@ -442,13 +442,15 @@ open path: .com.google.Chrome.R1234s
...
 
-The `--version` option prints the bpftrace version:
+- The `--version` option prints the bpftrace version:
 
 ```
 # bpftrace --version
 bpftrace v0.8-90-g585e-dirty
 ```
 
+- The `-w` option disables warnings.
```
danobi, line 452: Outdated?

src/main.cpp
```diff
@@ -64,7 +65,9 @@ void usage()
   std::cerr << "    --info         Print information about kernel BPF support" << std::endl;
   std::cerr << "    -k             emit a warning when a bpf helper returns an error (except read functions)" << std::endl;
   std::cerr << "    -kk            check all bpf helper functions" << std::endl;
-  std::cerr << "    -V, --version  bpftrace version" << std::endl << std::endl;
+  std::cerr << "    -V, --version  bpftrace version" << std::endl;
+  std::cerr << "    -w             disable all warning messages" << std::endl;
```
danobi, line 69: Outdated?

src/log.h
```diff
@@ -89,4 +105,9 @@ class LogStream
 // Note: LogType::DEBUG will prepend __FILE__ and __LINE__ to the debug message
 
 #define LOG(...) LogStream(__FILE__, __LINE__, LogType::__VA_ARGS__)
+
+#define DISABLE_LOG(type) bpftrace::Log::get().disable(LogType::type)
+#define ENABLE_LOG(type) bpftrace::Log::get().enable(LogType::type)
+#define LOG_ENABLED(type) bpftrace::Log::get().is_enabled(LogType::type)
```
danobi, line 111: Let's add these as needed. I don't see any callers yet. I also can't see a reason why callers would need this.
```suggestion
```

tests/runtime/basic
```diff
@@ -145,3 +145,13 @@ NAME tuple print
 RUN bpftrace -e 'BEGIN { @ = (1, 2, "string", (4, 5)); exit(); }'
 EXPECT ^@: \(1, 2, string, \(4, 5\)\)$
 TIMEOUT 5
+
+NAME disable warning: without --no-warnings
+RUN bpftrace  -e 'BEGIN { @x = stats(10); print(@x, 2); clear(@x); exit();}' 2>&1| grep -c  "WARNING"
+EXPECT ^1$
+TIMEOUT 1
```
danobi, line 152: We should already have tests for warning output. Runtime tests are relatively expensive b/c of how long they take to run. Not sure we should have a duplicated test.

PR #1449, opened 2020-07-31
Logging: change error/warning info to use new logging infra

suyuee:
This patch changes error/warning messages that are sent though std::cerr to use
new logging infra introduced in #1418.

##### Checklist

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -1,4 +1,5 @@
 #include "ast.h"
+#include "log.h"
 #include "parser.tab.hh"
 #include <iostream>
 
@@ -429,8 +430,7 @@ std::string opstr(Binop &binop)
     case bpftrace::Parser::token::BOR:   return "|";
     case bpftrace::Parser::token::BXOR:  return "^";
     default:
-      std::cerr << "unknown binary operator" << std::endl;
-      abort();
+      LOG(FATAL) << "unknown binary operator";
   }
 }
 
@@ -444,8 +444,7 @@ std::string opstr(Unop &unop)
     case bpftrace::Parser::token::INCREMENT: return "++";
     case bpftrace::Parser::token::DECREMENT: return "--";
     default:
-      std::cerr << "unknown unary operator" << std::endl;
-      abort();
+      LOG(FATAL) << "unknown unary operator";
   }
 }
 
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -432,6 +432,7 @@ std::string opstr(Binop &binop)
     default:
       LOG(FATAL) << "unknown binary operator";
   }
+  // lgtm[cpp/missing-return]
 }
 
 std::string opstr(Unop &unop)
@@ -446,6 +447,7 @@ std::string opstr(Unop &unop)
     default:
       LOG(FATAL) << "unknown unary operator";
   }
+  // lgtm[cpp/missing-return]
 }
 
 std::string AttachPoint::name(const std::string &attach_target,
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -477,11 +477,13 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
   // TODO (mmarchini): Handle base + index * scale addressing.
   // https://github.com/iovisor/bcc/pull/988
   if (argument->valid & BCC_USDT_ARGUMENT_INDEX_REGISTER_NAME)
-    std::cerr << "index register is not handled yet [" << argument->index_register_name << "]" << std::endl;
+    LOG(ERROR) << "index register is not handled yet ["
+               << argument->index_register_name << "]";
   if (argument->valid & BCC_USDT_ARGUMENT_SCALE)
-    std::cerr << "scale is not handled yet [" << argument->scale << "]" << std::endl;
+    LOG(ERROR) << "scale is not handled yet [" << argument->scale << "]";
   if (argument->valid & BCC_USDT_ARGUMENT_DEREF_IDENT)
-    std::cerr << "defer ident is not handled yet [" << argument->deref_ident << "]" << std::endl;
+    LOG(ERROR) << "defer ident is not handled yet [" << argument->deref_ident
+               << "]";
 
   if (argument->valid & BCC_USDT_ARGUMENT_CONSTANT)
     return getInt64(argument->constant);
@@ -547,7 +549,8 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
   }
 
   if (usdt == nullptr) {
-    std::cerr << "failed to initialize usdt context for probe " << attach_point->target << std::endl;
+    LOG(ERROR) << "failed to initialize usdt context for probe "
+               << attach_point->target;
     exit(-1);
   }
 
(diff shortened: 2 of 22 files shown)


src/map.cpp
```diff
@@ -62,16 +63,16 @@ Map::Map(const std::string &name, const SizedType &type, const MapKey &key, int
...
 }
 
 Map::Map(const SizedType &type) {
 #ifdef DEBUG
   // TODO (mmarchini): replace with DCHECK
   if (!type.IsStack()) {
-    std::cerr << "Map::Map(SizedType) constructor should be called only with stack types" << std::endl;
+    LOG(ERROR) << "Map::Map(SizedType) constructor should be called only with "
+                  "stack types";
     abort();
```
fbs, line 76: Would a `fatal` log level make sense? It would remove a lot of `abort()` calls

suyuee: Yeah, it sounds reasonable.

suyuee:
Update: introduce FATAL log level. But I think it can't remove all the abort calls since the compiler would be unhappy about "control reaches end of non-void function".

src/ast/ast.cpp
```diff
@@ -429,7 +430,7 @@ std::string opstr(Binop &binop)
     case bpftrace::Parser::token::BOR:   return "|";
     case bpftrace::Parser::token::BXOR:  return "^";
     default:
-      std::cerr << "unknown binary operator" << std::endl;
+      LOG(FATAL) << "unknown binary operator"; // will abort here
       abort();
```
danobi, line 434: Assuming you left the abort() in b/c of the static analysis warning, I'd rather have a comment to silence static analysis warning (if it exists, which it really should)

src/ast/ast.cpp
```diff
@@ -444,7 +445,7 @@ std::string opstr(Unop &unop)
     case bpftrace::Parser::token::INCREMENT: return "++";
     case bpftrace::Parser::token::DECREMENT: return "--";
     default:
-      std::cerr << "unknown unary operator" << std::endl;
```
danobi, line 447: Same as above

src/ast/irbuilderbpf.cpp
```diff
@@ -195,7 +196,8 @@ llvm::Type *IRBuilderBPF::GetType(const SizedType &stype)
         ty = getInt8Ty();
         break;
       default:
-        std::cerr << stype.size << " is not a valid type size for GetType" << std::endl;
+        LOG(FATAL) << stype.size
+                   << " is not a valid type size for GetType"; // will abort
         abort();
```
danobi, line 201: Same as above

src/main.cpp
```diff
@@ -187,9 +188,8 @@ static uint64_t get_btime(void)
   std::ifstream file("/proc/stat");
   if (!file)
   {
-    std::cerr << "Fail to open file /proc/stat: " << std::strerror(errno)
-              << std::endl
-              << "Builtin function strftime won't work properly." << std::endl;
+    LOG(ERROR) << "Fail to open file /proc/stat: " << std::strerror(errno)
```
danobi, line 191:
```suggestion
    LOG(ERROR) << "Failed to open file /proc/stat: " << std::strerror(errno)
```

src/main.cpp
```diff
@@ -208,9 +208,8 @@ static uint64_t get_btime(void)
   }
   if (btime == 0)
   {
-    std::cerr << "Fail to read btime from /proc/stat. Builtin function "
-                 "strftime won't work properly."
-              << std::endl;
+    LOG(ERROR) << "fail to read btime from /proc/stat. Builtin function "
```
danobi, line 211:
```suggestion
    LOG(ERROR) << "failed to read btime from /proc/stat. Builtin function "
```

src/ast/ast.cpp
suyuee: Yea, my intention was to silence the 'control may reach end of non-void function'.

> I'd rather have a comment to silence static analysis warning

Hmm, I'm not sure what you mean. Do you mean leaving a comment here to specify the intention, or is there actually a way to silence the warning?

src/main.cpp
suyuee: There are many other occurences of 'Fail to' in current error/warning messages, would you think we should replace them all with 'Failed to'?

src/ast/ast.cpp
danobi: > Hmm, I'm not sure what you mean. Do you mean leaving a comment here to specify the intention, or is there actually a way to silence the warning?

https://lgtm.com/help/lgtm/alert-suppression#c-c++

src/main.cpp
danobi: Let's just change the ones we're touching for now. Not a big deal, just grammar

src/ast/ast.cpp
suyuee: I assume it doesn't silence warnings that local compiler gives, right? Is it alright to have the compiler warnings? I mean Idk if it's somehow annoying to people.

danobi: No compiler warnings either, please. There should be a way to silence that too

src/log.h
```diff
@@ -98,13 +101,38 @@ class LogStream
...
 // 2. LOG(DEBUG, std::cout) << "this is a " << " message.";
 // 3. LOG(ERROR, call.loc, std::cerr) << "this is a semantic error";
 // Note: LogType::DEBUG will prepend __FILE__ and __LINE__ to the debug message
 
-#define LOG(...) LogStream(__FILE__, __LINE__, LogType::__VA_ARGS__)
+// clang-format off
+#define LOGSTREAM_DEBUG(...) bpftrace::LogStream(__FILE__, __LINE__, __VA_ARGS__)
+#define LOGSTREAM_INFO(...) bpftrace::LogStream(__FILE__, __LINE__, __VA_ARGS__)
+#define LOGSTREAM_WARNING(...) bpftrace::LogStream(__FILE__, __LINE__, __VA_ARGS__)
+#define LOGSTREAM_ERROR(...) bpftrace::LogStream(__FILE__, __LINE__, __VA_ARGS__)
```
danobi, line 131: Can you combine the definitions into a common macro? Not sure if possible, but would be nice if it is.

ie.
```
#define LOGSTREAM_COMMON(...) ...
#define LOGSTREAM_DEBUG(...) LOGSTREAM_COMMON(__VA_ARGS__)
```

src/log.cpp
```diff
@@ -201,7 +203,15 @@ LogStream::~LogStream()
...
     sink_.take_input(type_, loc_, out_, prefix + buf_.str());
+    if (type_ == LogType::FATAL)
+      abort();
   }
 }
 
+[[noreturn]] LogStreamFatal::~LogStreamFatal()
+{
+  sink_.take_input(type_, loc_, out_, buf_.str());
+  abort();
```
danobi, line 214: Shouldn't the body of this destructor be empty? The base class destructor already handles all the work, right?

src/log.h
```diff
@@ -86,7 +88,8 @@ class LogStream
       buf_ << v;
     return *this;
   }
-  ~LogStream();
+  virtual ~LogStream();
+  friend class LogStreamFatal;
```
danobi, line 92: Friend classes are generally a code smell. Why not make `private` members `protected` instead?

src/log.cpp
suyuee: Oh that's my bad, the base destructor shouldn't have done this:
```
    if (type_ == LogType::FATAL)
      abort();
```
`~LogStreamFatal` should call `abort` instead. Otherwise the compiler would complain `function declared 'noreturn' should not return [-Winvalid-noreturn]`

suyuee:
About the lgtm alerts:

>  the reason why the alerts are still being reported is because pull request changes to alert suppression, file classification (that is, whether a file is included in the LGTM analysis or excluded) and result display (showing or hiding the results of specific queries) are not taken into account during pull request checks. They are only effective once the pull request is merged on to the default branch. The reason for this is that we don't want an introduction of a bug being unnoticed if a contributor excludes it in the same pull request.

https://github.com/github/codeql/issues/2361#issuecomment-554991763
https://lgtm.com/help/lgtm/automated-code-review-faqs#why-automated-code-review-reports-misses-alert

Issue #841, opened 2019-07-24
Improve error messages (make them more informative and consistent)


mmarchini:
From #819:

```
print() expects a map to be provided
```

```
printf: %d specifier expects a value of type integer (sum supplied)
```

The formatting is different (`()` vs `:`) and one message is more informative than the other (printf outputs the received type, print doesn't).

In this case, the `print` message could be:

```
print: first parameters must be a map (sum supplied)
```

There's probably other inconsistencies across our error messages. We should have a unified format and we should try to keep them as informative as possible.

fbs:
I want to continue working on  adding line number support (see #757) the other PRs are done which should add a bit more context to the error. As I will have to touch most errors in the semantic analyser anyway I will look into this too.

jordalgo:
This is a very old issue that probably is outdated. closing for now.

PR #1451, opened 2020-08-04
Improve error messages for kfunc

viktormalik:
Most of the errors for kfunc (including, e.g., detecting a non-existing or a non-traceable function) occur when resolving kfunc arguments. This PR proposes to print more informative messages (instead of a generic "resolving kfunc arguments failed") that could help user to identify the problem.

Related to #841 

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -1,6 +1,7 @@
 #include "btf.h"
 #include "arch/arch.h"
 #include "bpftrace.h"
+#include "list.h"
 #include "types.h"
 #include "utils.h"
 #include <cstring>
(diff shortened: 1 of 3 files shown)


src/btf.cpp
```diff
@@ -439,23 +440,37 @@ int BTF::resolve_args(const std::string &func,
 
     t = btf__type_by_id(btf, t->type);
     if (!btf_is_func_proto(t))
-      return -1;
+    {
+      throw std::runtime_error("not a function\n");
```
fbs, line 444:
```suggestion
      throw std::runtime_error("not a function");
```

PR #1453, opened 2020-08-04
Devdocs

fbs:
##### Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,3 +8,4 @@ Then please complete the checklist.
 
 - [ ] Language changes are updated in `docs/reference_guide.md`
 - [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
+- [ ] The new behaviour is covered by tests
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -1,0 +1,38 @@
+# bpftrace development guide
+
+## Code style
+
+We use clang-format with our custom config for formatting code. This was
+[introduced](https://github.com/iovisor/bpftrace/pull/639) after a lot of code
+was already written. Instead of formatting the whole code base at once and
+breaking `git blame` we're taking an incremental approach, each new/modified bit
+of code needs to be formatted.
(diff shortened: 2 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,4 @@ Then please complete the checklist.
 
 - [ ] Language changes are updated in `docs/reference_guide.md`
 - [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
+- [ ] The new behaviour is covered by tests
```
javierhonduco, line 11: 😍

javierhonduco:
LGTM! Do we wanna ask folks to add the formatting changes in a commit below their changes?

fbs:
> LGTM! Do we wanna ask folks to add the formatting changes in a commit below their changes?

What do you mean by 'below'? All the PRs we do should have the individual commits  correctly formatted, having a fixup commit makes git blame harder to use. I had it in the document like

> We want to avoid `fix formatting` commits. Instead every commit should be
formatted correctly.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,4 @@ Then please complete the checklist.
```
fbs, line 8: I  think it will be good to link the document  here somehow

javierhonduco:
What I meant is that instead of including the changes + formatting in one commit, which might make reviewing it harder, maybe doing the formatting in one commit + actual changes in another commit could be better. We can always squash them at merge time

fbs:
> What I meant is that instead of including the changes + formatting in one commit, which might make reviewing it harder, maybe doing the formatting in one commit + actual changes in another commit could be better. We can always squash them at merge time

Hmm, I'm not a big fan of that. If we forget to squash we still end up with that 'fix formatting' commit. And on a PR that has more than one commit it won't work either.

danobi:
> What I meant is that instead of including the changes + formatting in one commit, which might make reviewing it harder,

Hmm, `git clang-format` only formats the lines you touched. I've never had any issue with unreadability. And "formatting" commit after the "real" commits also kinda clobbers git blame.

PR #1457, opened 2020-08-06
Fix array indexing regression

danobi:
We regressed on array indexing because the CreateArray() helper forgot
to set pointee_size. This caused all array indexes to silently index to
index 0.

##### Checklist


danobi:
#1394 will kill `pointee_size`. It'd be nice to merge this in now so the pointer PR will be tested against the regression test.

PR #1458, opened 2020-08-06
Add the compile option to run with BPF ring buffer

suyuee:
BPF ring buffer is a new BPF map type introduced in linux 5.8 and it has the
potential to improve memory efficiency and solve event re-ordering problems
comparing with BPF perf buffer. This patch adds a compile option
`USE_BPF_RINGBUF` that enables bpftrace to run with BPF ring buffer. This
patch is experimental and follow-ups can benchmark between BPF ring buffer and
BPF perf buffer. The second commit adds a simple benchmark test on event loss
rate.
##### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,3 +169,7 @@ add_subdirectory(tools)
 if (ENABLE_MAN)
 add_subdirectory(man)
 endif(ENABLE_MAN)
+
+if(USE_BPF_RINGBUF)
+unset(USE_BPF_RINGBUF CACHE)
+endif(USE_BPF_RINGBUF)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -138,3 +138,7 @@ if(NOT "${retcode}" STREQUAL "0")
 endif()
 
 add_definitions("-DBPFTRACE_VERSION=\"${BPFTRACE_VERSION}\"")
+
+if(USE_BPF_RINGBUF)
+  target_compile_definitions(bpftrace PRIVATE USE_BPF_RINGBUF)
+endif(USE_BPF_RINGBUF)
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -68,3 +68,7 @@ else()
   endif()
   target_link_libraries(ast libclang)
 endif()
+
+if(USE_BPF_RINGBUF)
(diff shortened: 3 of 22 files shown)


suyuee:
Some results of the event loss rate benchmark test.

test_1
test_2

fbs:
Nice, what are your plans for this? Any reason it has to stay hidden behind a compile flag and not behind runtime detection (maybe a runtime feature flag tho)?

suyuee:
> Nice, what are your plans for this? Any reason it has to stay hidden behind a compile flag and not behind runtime detection (maybe a runtime feature flag tho)?

BPF ring buffer is still very new and most hosts don't have the feature yet. I was thinking maybe it is still premature to be exposed to users and it might be good to gather some thoughts from developments first. Tbh I'm not very sure about this part.

danobi:
Runtime detection sounds reasonable. I originally suggested compile time flag b/c it was faster to iterate with than writing runtime detection code.

Since this touches on pretty core parts of bpftrace, I kind of envision using bpf ringbuf in these phases:
* default perf ringbug, flag to force bpf ringbug (maybe `--bpfringbuf`)
* use bpf ringbuf when available, detected via runtime test, with flag to fallback to perf buf (maybe `--no-bpfringbuf`)

My only concern is that we'll have to support two code paths for pretty much forever. But saving memory would be really nice. Especially b/c perf ringbuf memory usage scales linearly with # cpus (which totally sucks).

fbs:
Would be nice to land this. 

Having feature detection + a flag  (`--bpfringbuf`) sounds like a good way forward.

aktau:
Ping. It's 2022 now, and the feature far more ubiquitous.

xh4n3:
We use bpftrace to analyse root causes of intermittent packet delay or drop. When the server is busy, we're likely to miss some key packets due to the event loss.

This feature should be helpful. Are you available to land this? @suyuee

suyuee:
> We use bpftrace to analyse root causes of intermittent packet delay or drop. When the server is busy, we're likely to miss some key packets due to the event loss.
> 
> This feature should be helpful. Are you available to land this? @suyuee

suyuee:
Unfortunately I don't have the time to land this; feel free to fork it if you find it useful

xh4n3:
@suyuee 😃 Thank you for letting us know. Let me see if I could get it going.

PR #1460, opened 2020-08-11
biosnoop: Output disk name, offset and length

lathiat:
- Add support for outputting disk name, offset and length
- Outputs the system time and startup to compare to the relative timestamps
- Requires bpftrace v0.11.0 due to using more than 7 arguments to printf

##### Checklist

- [N/A] Language changes are updated in `docs/reference_guide.md`
- [Yes] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [No] The new behaviour is covered by tests - there currently seems to be no test framework for the example tools

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ and this project adheres to
   - [#1457](https://github.com/iovisor/bpftrace/pull/1457)
 
 #### Tools
+- biosnoop: Added output of disk name, offset and length
+  - [#1460](https://github.com/iovisor/bpftrace/pull/1460)
 
 #### Documentation
 
--- a/man/man8/biosnoop.8
+++ b/man/man8/biosnoop.8
@@ -4,9 +4,10 @@ biosnoop.bt \- Block I/O tracing tool, showing per I/O latency. Uses bpftrace/eB
 .SH SYNOPSIS
 .B biosnoop.bt
 .SH DESCRIPTION
-This is a basic block I/O (disk I/O) tracing tool, showing each I/O event
-along with the issuing process ID, and the I/O latency. This can be used to
-investigate disk I/O performance issues.
+This is a basic block I/O (disk I/O) tracing tool, showing each I/O event along
+with the issuing process ID, name, I/O latency, disk, offset and length. This
+can be used to investigate disk I/O performance issues. It also outputs the
+system time at startup to reference against the TIME field of each output row.
 
 This tool currently works by dynamic tracing of the blk_account*() kernel
 functions, which will need updating to match any changes to these functions
@@ -17,7 +18,7 @@ Since this uses BPF, only the root user can use this tool.
 CONFIG_BPF and bpftrace.
 .SH EXAMPLES
 .TP
-Trace block I/O events, printing per-line summaries:
+Trace block I/O events, printing one line per operation:
 #
 .B biosnoop.bt
 .SH FIELDS
@@ -31,8 +32,19 @@ Issuing process name. This often identifies the issuing application process, but
 PID
 Issuing process ID. This often identifies the issuing application process, but I/O may be initiated from kernel threads only.
 .TP
-ARGS
-Process name and arguments (16 word maximum).
+DISK
+The name of the target disk, be aware that this will generally show the
+underlying disk receiving the operation if some kind of translation layer such
+as RAID or bcache is used.
+.TP
+OFFSET
+The offset of the I/O operation in bytes.
+.TP
+LEN
+The length of the I/O operation in bytes.
+.TP
+LAT(ms)
+The latency of the I/O operation in milliseconds (ms).
 .SH OVERHEAD
 Since block device I/O usually has a relatively low frequency (< 10,000/s),
 the overhead for this tool is expected to be negligible. For high IOPS storage
@@ -55,5 +67,6 @@ Linux
 Unstable - in development.
 .SH AUTHOR
 Brendan Gregg
+Trent Lloyd
 .SH SEE ALSO
 opensnoop(8)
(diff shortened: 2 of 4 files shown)


lathiat:
CI failure seems to be from the CI environment during build of the docker container:
ERROR: Failed to create usr/libexec/gcc/x86_64-alpine-linux-musl/9.2.0/cc1plus: I/O error

tools/biosnoop.bt
```diff
@@ -1,43 +1,64 @@
...
 
 kprobe:blk_account_io_start
 {
+	$req = ((struct request *)arg0);
+
 	@start[arg0] = nsecs;
 	@iopid[arg0] = pid;
 	@iocomm[arg0] = comm;
+	@iolen[arg0] = ((struct request *)arg0)->__data_len;
+	@iostart[arg0] = $req->__sector * $req->q->limits.logical_block_size
```
brendangregg, line 32: Why stash these on start? Can't these fields be read on done, which would avoid needing to allocate the BPF maps? (The other fields were stashed on start since they are process context and time, and aren't available in the done arguments.)

tools/biosnoop.bt
```diff
@@ -1,43 +1,64 @@
...
+ *       			Output systm time at startup. Requires bpftrace
+ *      			> 0.11.0 for more than 6 printf arguments.
  */
 
+#include <linux/blkdev.h>
+
 BEGIN
 {
-	printf("%-12s %-16s %-6s %7s\n", "TIME(ms)", "COMM", "PID", "LAT(ms)");
+	time("Start Time: %F %T\n");
```
brendangregg, line 20: I would not add this start time. It's clutter, and there's the risk people will add it (and more) to all tools. This is the kind of thing I put behind an option flag in the bcc version; we don't have option flags in bpftrace yet, although even if we did, I'm not sure we want the bpftrace tools to grow this much.

I sometimes automate these tools and want details such as the date and other details stored with the output: so I simply run commands like "date; uname -a; uptime" before the tool.

tools/biosnoop_example.txt
```diff
@@ -4,33 +4,20 @@ Demonstrations of biosnoop, the Linux BPF/bpftrace version.
...
-802          cksum            4179        32
-[...]
+Start Time: 2020-07-31 16:03:09
+TIME(ms)     COMM             PID     DISK     OFFSET        LEN    LAT(ms)
+720          bash             1319718 nvme0n1  777994240     4096         0
+724          cksum            1321648 nvme0n1  680353792     16384        13
+725          cksum            1321648 nvme0n1  2021421056    4096         10
+725          cksum            1321648 nvme0n1  2021355520    4096         6
+725          cksum            1321648 nvme0n1  777986048     4096         10
+726          cksum            1321648 nvme0n1  1331789824    4096         15
```
brendangregg, line 15: The LAT(ms) column does not match the format ("%7d").

brendangregg:
In general this is a welcome improvement, thanks. And I'm glad we're still <80 chars.

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ and this project adheres to
   - [#1457](https://github.com/iovisor/bpftrace/pull/1457)
 
 #### Tools
+- biosnoop: Added output of disk name, offset and length
+  - Also print time and startup to correlate with the offset time
```
fbs, line 50:
```suggestion
  - [#1460](https://github.com/iovisor/bpftrace/pull/1460)
```

The link to the PR can only be added after opening it, a bit annoying but it helps to create a useful changelog :) (please squash this into your current commit)

tools/biosnoop.bt
lathiat: I originally tried that but the __data_len specifically gets zerod out during the request processing and reads 0 during blk_account_io_done. Looking at some kernel commits they renamed them with the private __ prefix in part because of some of that behaviour breaking external code relying on it but that's as far as I looked into that.

While the offset doesn't seem to do that at first glance I was concerned since it was the related private field it may also sometimes change and decided to stash that also -- but I didn't comprehensively look at the kernel paths to figure out if that was likely or not.

tools/biosnoop.bt
lathiat: The problem I had here is I sometimes see multi-second delays to the BPF startup that makes that method quite innacurate for correlating against other logs.

I'll remove this but perhaps we could solve that by adding a generic "print timestamp at startup" option to bpftrace itself that correctly correlates with the elapsed counter.

tools/biosnoop_example.txt
lathiat: Fixed in the latest commit, I modified the times to keep the existing example sensible and apparently I am not a good printf parser :)

CHANGELOG.md
lathiat: Done

man/man8/biosnoop.8
```diff
@@ -4,9 +4,10 @@ biosnoop.bt \- Block I/O tracing tool, showing per I/O latency. Uses bpftrace/eB
 .SH SYNOPSIS
 .B biosnoop.bt
 .SH DESCRIPTION
-This is a basic block I/O (disk I/O) tracing tool, showing each I/O event
-along with the issuing process ID, and the I/O latency. This can be used to
-investigate disk I/O performance issues.
+This is a basic block I/O (disk I/O) tracing tool, showing each I/O event along
+with the issuing process ID, name, I/O latency, disk, offset and length. This
+can be used to investigate disk I/O performance issues. It also outputs the
+system time at startup to reference against the TIME field of each output row.
```
danobi, line 10: Start time is gone now, right?

tools/biosnoop_example.txt
```diff
@@ -2,36 +2,22 @@ Demonstrations of biosnoop, the Linux BPF/bpftrace version.
...
-717          cksum            4179         6
-728          cksum            4179        10
-735          cksum            4179         6
-751          cksum            4179        10
-758          cksum            4179        17
-783          cksum            4179        12
-796          cksum            4179        25
-802          cksum            4179        32
-[...]
+Start Time: 2020-07-31 16:03:09
```
danobi, line 10: Same as above

fbs:
@lathiat poke

Issue #1454, opened 2020-08-05
Drop as many capabilities as we can


danobi:
### Is your feature request related to a problem? Please describe.
CAP_PERFMON and CAP_BPF are landed in 5.8 kernel ( https://lwn.net/Articles/827820/ ). We should consider dropping CAP_SYS_ADMIN b/c less caps are always safer.

### Describe the solution you'd like
Drop all our unused caps. We can use bpftrace on bpftrace to figure out which caps are actually used.

### Describe alternative solutions or features you've considered
seccomp?

fbs:
Ill give this a go, getting some basic capabilities and seccomp in shouldn't be too hard.

Some questions:

- How do we want this to behave? Should this be an essential part (so libcap, libseccomp are require for build and runtime) or is it opt-in?
- Should we terminate on failure or should we warn and continue (like we do with rlimit)
- How do we run child procs? Do we start them before or after we drop capabilities?

Xyene:
I ran into [the root check](https://github.com/bpftrace/bpftrace/blob/279301df50c74123f311c9939d74cb726d79636f/src/run_bpftrace.cpp#L29-L35
) today after setting up an unprivileged user with ambient `cap_perfmon` and `cap_bpf` permissions in the hopes of running bpftrace without granting full root / setuid-ing the bpftrace binary.

After commenting out that check, I also ran into two other issues, documenting them here in case they're useful in the future.

- `/sys/fs/bpf` was mounted with `700` permissions, had to relax this to `750`.
- `/sys/kernel/debug/tracing` was mounted with more restrictive permissions than `/sys/kernel/tracing`, but bpftrace only knows to look for the debugfs path. This was also easily workaroundable, but maybe bpftrace should try the tracefs path first?

Having done that, things seem to Just Work, at least for the use-cases I had.

Is there a reason bpftrace requires root and this ticket is about dropping root perms, as opposed to going the other way and saying that a certain set of capabilities (whether they're granted by setuid, ambiently, or otherwise) sufficient?

Of course, pre-5.8 that set is just `cap_sys_admin` since there's no separate `cap_bpf`; not sure if there are other capabilities that bpftrace needs that don't have separate capabilities.

PR #1461, opened 2020-08-11
RFC: drop capabilities

fbs:
First attempt at #1454 

This drops our capabilities down to the minimum required to run bpftrace. As a "command" (`-c`) might need more permissions than the fork happens before the capabilities are dropped[1][2]. The `system()` call is impacted by this. I'm not sure if anyone uses it to do "dangerous stuff", you probably shouldn't but if we do this it will break existing behaviour.

The capabilities are currently static with a fixed "profile" for all. But using the `system_has_cap` function we could auto detect what the kernel has and use that instead. But that has the same problem as we had with bpffeature. If we want to be portable and run on newer kernel we need to make sure those symbols are defined, so that means pulling (part of) `capability.h` into our tree, which I'm not a big fan of (it's ok now, but not if we keep adding files).

This seems to work well in ubuntu-19.10 vagrant, but fails to list probes with `-l -p` on the fedora-32 box.  Not sure what the issues is there yet. Might be related to the slightly different pidfd code path.

I'm not entirely convinced by this yet. Over al I don't think it hurts to have this but I don't think it protects us against a real exploit either. Seccomp might be a better approach, we could add an earlier analyser pass that analyses what we need for the run and use only that.

[1]: . If we want to limit the childproc further we should add flags to change user and drop caps, libcap can parse strings to cap values, so maybe `--child-capabilities CAP_CHOWN,CAP_NET_RAW+ep`
[2]: The `watchpoint` tests fail if capabilties are dropped first

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,7 @@ include_directories(SYSTEM ${LIBBCC_INCLUDE_DIRS})
 find_package(LibElf REQUIRED)
 include_directories(SYSTEM ${LIBELF_INCLUDE_DIRS})
 
+find_package(LibCap REQUIRED)
 find_package(BISON REQUIRED)
 find_package(FLEX REQUIRED)
 bison_target(bison_parser src/parser.yy ${CMAKE_BINARY_DIR}/parser.tab.cc VERBOSE)
--- a/cmake/FindLibCap.cmake
+++ b/cmake/FindLibCap.cmake
@@ -1,0 +1,15 @@
+find_path (LIBCAP_INCLUDE_DIRS
+  NAMES
+    capability.h
+  PATH_SUFFIXES
+    sys
+)
+
+find_library (LIBCAP_LIBRARIES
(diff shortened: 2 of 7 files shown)


fbs:
when doing this we can probably remove the root check, modify it to something like : `  sorry, you must be root or get CAP_NET_ADMIN capability to do this`

danobi:
> I'm not entirely convinced by this yet. Over al I don't think it hurts to have this but I don't think it protects us against a real exploit either.

I'm feeling somewhat the same now. I think it sounds good on paper but in practice we'd probably get a lot of regressions and also not get much security benefit. Especially b/c CAP_SYS_ADMIN is pretty much root.

mmisono:
Can we also drop CAP_SYS_ADMIN of `bpftrace_test` to ensure that the test runs without root?

fbs:
Don't think this is worth while right now but ill leave the branch around in  case anyone wants to pick it up.

PR #1463, opened 2020-08-12
Feature-detect bpf_attach_kprobe signature

### LINKED ISSUE #1027: Remove cast_signature hack in attached_probe.cpp
See https://github.com/iovisor/bpftrace/blob/master/src/attached_probe.cpp#L548-L561

This works ok for calling conventions that pass the first 6 arguments in registers but things would start to get weird if bpftrace gets compiled with a calling convention that places all args on the stack. 

It's more robust to do a feature test. See https://github.com/iovisor/bpftrace/blob/master/cmake/FindLibBfd.cmake#L64-L78 for an example of a feature test.

cc @mmarchini

viktormalik:
This removes cast_signature hack in attached_probe.cpp and fixes #1027.

The detection is in CMake similarly to [FindLibBfd.cmake#L80-L110](https://github.com/iovisor/bpftrace/blob/master/cmake/FindLibBfd.cmake#L80-L110). 

I tried to follow the same approach of using `*_LIBRARIES` variables for static linking. This may cause some packages/libraries to be searched for multiple times, which should be fine since CMake caches results.

##### Checklist


cmake/FindLibBcc.cmake
```diff
@@ -70,3 +70,38 @@ include (FindPackageHandleStandardArgs)
...
+
+# Check bpf_attach_kprobe signature
+if(${LIBBCC_FOUND})
+if(STATIC_LINKING)
+  # libbcc.a is not statically linked with libbpf.a, libelf.a, and libz.a.
+  # If we do a static bpftrace build, we must link them in.
+  find_package(LibBpf)
+  find_package(LibElf)
+  find_library(LIBZ_LIBRARIES
+    NAMES
```
fbs, line 82: this seems to be a copy of the one in libbfd, i think we should move this to FindLibZ.cmake

viktormalik: Right, thanks, done.

cmake/FindLibBcc.cmake
```diff
@@ -70,3 +70,26 @@ include (FindPackageHandleStandardArgs)
...
+endif()
+INCLUDE(CheckCXXSourceCompiles)
+CHECK_CXX_SOURCE_COMPILES("
+#include <bcc/libbpf.h>
+
+int main(void) {
+  bpf_attach_kprobe(0, BPF_PROBE_ENTRY, \"\", \"\", 0, 0);
+  return 0;
+}
+" LIBBCC_ATTACH_KPROBE_SIX_ARGS_SIGNATURE)
```
fbs, line 94:
```suggestion
" LIBBCC_ATTACH_KPROBE_SIX_ARGS_SIGNATURE)
  SET(CMAKE_REQUIRED_DEFINITIONS)
  SET(CMAKE_REQUIRED_LIBRARIES)
```

iirc you need this to avoid leaking this global var into other Cmake tests. See 6c3768d06218418e140a2d5a4edfec9c564df972

viktormalik: Correct, I overlooked this. But I think that it should suffice to set `CMAKE_REQUIRED_LIBRARIES` since I didn't change `CMAKE_REQUIRED_DEFINITIONS`, shouldn't it?

fbs: Yes it  should be, I don't think it would hurt to leave it in but up to you :)

viktormalik: Ok, I prefer to set what was changed only.

PR #1464, opened 2020-08-13
Fix kfarg access

mmisono:
Fix kfarg access so that a chain of dereference works.

Example:

```
% bpftrace -e 'kfunc:vfs_open { printf("%s\n", str(args->path->dentry->d_name.name));}'
Attaching 1 probe...
cmdline
^C
```

##### Checklist

### DIFF
--- a/src/ast/field_analyser.cpp
+++ b/src/ast/field_analyser.cpp
@@ -193,6 +193,7 @@ void FieldAnalyser::visit(FieldAccess &acc)
         type_ = "";
     }
 
+    bpftrace_.btf_set_.insert(type_);
     has_builtin_args_ = false;
   }
   else if (!type_.empty())
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1846,6 +1846,12 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
   auto search = variable_val_.find(var_ident);
   assignment.var->type = assignment.expr->type;
 
+  auto *builtin = dynamic_cast<Builtin *>(assignment.expr);
+  if (builtin && builtin->ident == "args" && builtin->type.is_kfarg)
+  {
+    LOG(ERROR, assignment.loc, err_) << "args cannot be assigned to a variable";
+  }
+
   if (search != variable_val_.end()) {
     if (search->second.IsNoneTy())
     {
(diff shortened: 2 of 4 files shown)


fbs:
Oops. Seems like I messed up during all the rebases that had to happen.  

Can you add a test for this? Looks like we don't have  any kfunc  tests  yet.

mmisono:
Assigning `args` to a variable does not work.

```
% sudo ./src/bpftrace -e 'kfunc:vfs_open { $a = args;  printf("%s\n", str($a->path->dentry->d_name.name));}'
stdin:1:45-65: ERROR: Unknown struct/union: 'struct path'
kfunc:vfs_open { $a = args;  printf("%s\n", str($a->path->dentry->d_name.name));}
                                            ~~~~~~~~~~~~~~~~~~~~
```

I'll take a look over the weekend. Maybe disabling the assignment of args is a simple way to cope with this.

fbs:
> Assigning `args` to a variable does not work.
> 
> ```
> % sudo ./src/bpftrace -e 'kfunc:vfs_open { $a = args;  printf("%s\n", str($a->path->dentry->d_name.name));}'
> stdin:1:45-65: ERROR: Unknown struct/union: 'struct path'
> kfunc:vfs_open { $a = args;  printf("%s\n", str($a->path->dentry->d_name.name));}
>                                             ~~~~~~~~~~~~~~~~~~~~
> ```
> 
> I'll take a look over the weekend. Maybe disabling the assignment of args is a simple way to cope with this.

Did this work prior to the struct type? I don't think this is a regression. Also, isn't the issues not with it not knowning `struct dentry`, the `$a->path` part seems to work ok

mmisono:
> Can you add a test for this? Looks like we don't have any kfunc tests yet.

There is a `TEST_F(semantic_analyser_btf, kfunc)`. I added a test case for a chain of a dereference. 9bffa55

> Did this work prior to the struct type?  I don't think this is a regression.

Previously, storing "args" in tracepiont is OK, but storing "args" in kfunc is an error.

> Also, isn't the issues not with it not knowning struct dentry, the $a->path part seems to work ok

The problem is FieldAnalyzer does not handle this case properly and information that the variable is kfarg will be lost. I disabled an assignment of kfarg in a semantic analyzer.  a218b8c

PR #1468, opened 2020-08-17
add time function show millisecond time by %f format,for example,time("%H:%M:%S.%f");

thinkingning:
add time function show millisecond time by %f format,for example,time("%H:%M:%S.%f");

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -38,6 +38,10 @@
 #include "triggers.h"
 #include "utils.h"
 
+#include <string.h>
+#include <sys/timeb.h>
+#include <time.h>
+
 namespace libbpf {
 #define __BPF_NAME_FN(x) #x
 const char *bpf_func_name[] = { __BPF_FUNC_MAPPER(__BPF_NAME_FN) };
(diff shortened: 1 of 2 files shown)


danobi:
To clarify, what I mean to say is that millisecond timestamps seem like they would be totally inaccurate if used with time() builtin. Since strftime() takes an explicit timestamp (nsecs), it makes more sense to me to add millisecond precision there.

And also, thanks for the PR

thinkingning:
when I print time,I need to know the trace point millisecond time difference,so I extend time function in millisecond format.

thinkingning:
> What's the use case for this? time() is async which means there's an unbounded amount of time between when a bpftrace script calls time() and the bpftrace runtime actually prints the time. So by design it's not really going to be significant to events you're tracing.
> 
> If you need this, we should instead look at extending the `strftime()` builtin.

when I print time,I need to know the trace point millisecond time difference,so I extend time function in millisecond format.

fbs:
Thanks for looking into this @thinkingning 

The issue is not so much with milisecond time accuracy, that makes sense, it's with the implementation. What happens  with a program like `i:s:1 { time() }` is that once a second the bpf program in the kernel sends a "hey print the current time" event to the bpftrace userspace application. Bpftrace continuously polls for those events and when they arrive it grabs the current time and prints that. The issue is that there is no way of knowing what the delay between the kernel sending the event and the userspace application handling it is. On an idle system that might be a few us, but on a loaded system it might be a second, nobody knows. As we cannot set an upper limit on the delay the milisecond accuracy is useless.

The `strftime()` call is different, it sends a timestamp with the event "hey print this timestamp: 1323233232" so it doesn't suffer from the delay. THe only issue with it is that userspace can only get the time since boot in seconds, so again miliseconds don't work. But if you can find a way to improve that we can add miliseconds (or better!) to strftime :)

danobi:
I'm asking around to see if there's a way to get high precision boot time. So far I don't see any but it looks pretty easy to add another field to /proc/stat: http://ix.io/2uxX

danobi:
#1471 gives us sub millisecond precision on boot time now.

brendangregg:
Sounds like we also need to add a note to the reference guide to say something like: time() is rough (async); for accurate millisecond timestamps, use strftime().

PR #1471, opened 2020-08-18
Calculate boot time with sub-second precision

danobi:
If we do the "triple vdso sandwich" a couple of times and take the
results from the iteration that has the lowest time delta between the
first and last clock_gettime(), we can get pretty good precision.

My test showed ~30ns of accuracy.

Test plan:
```
$ sudo bpftrace -e 'i:s:1 { printf("%s\n", strftime("%D %H:%M:%S %P %Z", 0)); }'
Attaching 1 probe...
08/17/20 18:28:19 pm PDT
```

Which corresponds to my `last reboot` time:
```
$ last reboot | head -n1
reboot   system boot  5.7.12-arch1-1   Mon Aug 17 18:28   still running
```

##### Checklist

- [NA] Language changes are updated in `docs/reference_guide.md`
- [NA] User-visible and non-trivial changes updated in `CHANGELOG.md`

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1677,17 +1677,16 @@ std::string BPFtrace::resolve_uid(uintptr_t addr) const
 std::string BPFtrace::resolve_timestamp(uint32_t strftime_id,
                                         uint64_t nsecs_since_boot)
 {
-  if (btime == 0)
+  if (!boottime_)
   {
-    LOG(ERROR)
-        << "Cannot resolve the timestamp returned by strftime due to the "
-           "previous failure to get btime from /proc/stat.";
+    LOG(ERROR) << "Cannot resolve timestamp due to failed boot time calcuation";
     return "(?)";
   }
   auto fmt = strftime_args_[strftime_id].c_str();
   char timestr[STRING_SIZE];
   struct tm tmp;
-  time_t time = btime + nsecs_since_boot / 1e9;
+  time_t time = boottime_->tv_sec +
+                ((boottime_->tv_nsec + nsecs_since_boot) / 1e9);
   if (!localtime_r(&time, &tmp))
   {
     LOG(ERROR) << "localtime_r: " << strerror(errno);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,11 +1,14 @@
+#include <array>
 #include <csignal>
 #include <cstdio>
 #include <cstring>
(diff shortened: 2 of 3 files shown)


danobi:
cc @suyuee

suyuee:
Awesome, looks so nice!

src/main.cpp
```diff
@@ -185,37 +188,67 @@ static int info()
...
   }
-  if (btime == 0)
-  {
-    std::cerr << "Fail to read btime from /proc/stat. Builtin function "
-                 "strftime won't work properly."
-              << std::endl;
-  }
-  return btime;
+
+  if (ret && lowest_delta >= 1e6)
```
fbs, line 247: I wonder if we can get away with `1e5` here, that way you know the timing is accurate to 1ms +- 0.1ms.

src/main.cpp
```diff
@@ -185,37 +188,67 @@ static int info()
...
+    LOG(WARNING) << ">1ms skew detected when calculating boot time. strftime() "
```
fbs, line 248:
```suggestion
    LOG(WARNING) << (lowest_delta / 1e3) >> "us skew detected when calculating boot time. strftime() "
                         "builtin may be inaccurate";
```

fbs: might be useful to know how bad it really is

src/bpftrace.cpp
```diff
@@ -1680,7 +1680,7 @@ std::string BPFtrace::resolve_uid(uintptr_t addr) const
 std::string BPFtrace::resolve_timestamp(uint32_t strftime_id,
                                         uint64_t nsecs_since_boot)
 {
-  if (btime == 0)
+  if (!boottime_)
   {
     std::cerr << "Cannot resolve the timestamp returned by strftime due to the "
```
fbs, line 1685: needs updating i think

src/main.cpp
fbs: Did a quick test in a small virtualbox vm on a compiling macbook. Seems like everything completes in 100ns.  Worst delay was 2335450 nsecs, at that point the system was already unusable but the 2nd iteration still completed in a few 100ns.

danobi: Yeah, 1e5 should be fine too

src/main.cpp
danobi: Good idea

PR #1475, opened 2020-08-19
Switch `nsecs` to `ktime_get_boot_ns`

### LINKED ISSUE #1474: Switch `nsecs` to  `bpf_ktime_get_boot_ns`
This was discussed a bit in #1337 but never happened. I don't see a downside to switching it. The `clock_gettime` manpage even says:

```
CLOCK_BOOTTIME (since Linux 2.6.39; Linux-specific)
    Identical to CLOCK_MONOTONIC, except it also includes any time that the system is suspended. This allows applications to get a suspend-aware monotonic clock without having to deal with the complications of CLOCK_REALTIME, which may have discontinuities if the time is changed using settimeofday(2). 
```

```
       u64 bpf_ktime_get_ns(void)

              Description
                     Return the time elapsed since system boot, in
                     nanoseconds.  Does not include time the system was
                     suspended.  See: clock_gettime(CLOCK_MONOTONIC)

              Return Current ktime.


       u64 bpf_ktime_get_boot_ns(void)

              Description
                     Return the time elapsed since system boot, in
                     nanoseconds.  Does include the time the system was
                     suspended.  See: clock_gettime(CLOCK_BOOTTIME)

              Return Current ktime.
```

fbs:
Fixes #1474

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -117,7 +117,7 @@ void CodegenLLVM::visit(Builtin &builtin)
 {
   if (builtin.ident == "nsecs")
   {
-    expr_ = b_.CreateGetNs();
+    expr_ = b_.CreateGetNs(bpftrace_.feature_.has_helper_ktime_get_boot_ns());
   }
   else if (builtin.ident == "elapsed")
   {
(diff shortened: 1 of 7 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -117,7 +117,7 @@ void CodegenLLVM::visit(Builtin &builtin)
 {
   if (builtin.ident == "nsecs")
   {
-    expr_ = b_.CreateGetNs();
+    expr_ = b_.CreateGetNs(bpftrace_.feature_.has_helper_ktime_get_boot_ns());
   }
   else if (builtin.ident == "elapsed")
```
fbs, line 122: depending on the discussion in #1474  this elapsed might have to be updated too

PR #1485, opened 2020-08-26
Resolve unknown/undefined types for tracepoints

### LINKED ISSUE #730: Provide better error message for tracepoints with undefined struct types
I think we can somehow provide better information for users when there are missing includes in tracepoint usage. For example

```
# bpftrace -e 'tracepoint:syscalls:sys_enter_io_setup {printf("%u %x %p\n", pid, args->nr_events, args->ctxp)}'
definitions.h:10:3: error: unknown type name 'aio_context_t'

```

The program above needs `uapi/linux/aio_abi.h` to work, but the error message don't give anything actionable to the user (not even a tip that maybe they should include a Kernel header).

_Originated from https://github.com/iovisor/bpftrace/issues/604#issuecomment-499699673_

### LINKED ISSUE #1154: tracepoint unknown type name 'sigset_t'
```
➜  ~ bpftrace -e 'tracepoint:syscalls:sys_enter_io_uring_enter {printf("%d\n",args->fd)}'
definitions.h:13:9: error: unknown type name 'sigset_t'
```
the tracepoint is :

```
➜  ~ cat /sys/kernel/debug/tracing/events/syscalls/sys_enter_io_uring_enter/format
name: sys_enter_io_uring_enter
ID: 916
format:
	field:unsigned short common_type;	offset:0;	size:2;	signed:0;
	field:unsigned char common_flags;	offset:2;	size:1;	signed:0;
	field:unsigned char common_preempt_count;	offset:3;	size:1;	signed:0;
	field:int common_pid;	offset:4;	size:4;	signed:1;

	field:int __syscall_nr;	offset:8;	size:4;	signed:1;
	field:unsigned int fd;	offset:16;	size:8;	signed:0;
	field:u32 to_submit;	offset:24;	size:8;	signed:0;
	field:u32 min_complete;	offset:32;	size:8;	signed:0;
	field:u32 flags;	offset:40;	size:8;	signed:0;
	field:const sigset_t * sig;	offset:48;	size:8;	signed:0;
	field:size_t sigsz;	offset:56;	size:8;	signed:0;
```

it seems bpftrace can't found type define for the tracepoint's some args field
Env :
```
➜  ~ uname -a
Linux lambda 5.3.0-25-generic #27-Ubuntu SMP Wed Dec 4 13:36:32 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
➜  ~ bpftrace --version
bpftrace v0.9.2
```

viktormalik:
If a type is defined by a typedef located in a non-included header, clang parser will fail. This change analyses error messages from the clang parser and retrieves the missing typedefs from BTF (if it is used). If all types cannot be resolved, it provides the user a hint: either to include BTF (running with `--btf`) or to include relevant headers.

Resolves #1154, resolves #730. It should also help with iovisor/kubectl-trace#109.

I was also wondering, if it would make sense to get rid of the `--btf` option completely. AFAIU, it is used for tracepoints only. Using BTF automatically whenever it is available would make most of the "unknown type name" errors disappear in default bpftrace setup.

##### Checklist

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -386,13 +386,6 @@ bool ClangParser::ClangParserHandler::check_diagnostics(
   return true;
 }
 
-bool ClangParser::ClangParserHandler::check_diagnostics(
-    const std::string &input)
-{
-  std::vector<std::string> msgs;
-  return check_diagnostics(input, msgs, true);
-}
-
 CXCursor ClangParser::ClangParserHandler::get_translation_unit_cursor() {
   return clang_getTranslationUnitCursor(translation_unit);
 }
@@ -529,15 +522,11 @@ std::unordered_set<std::string> ClangParser::get_incomplete_types(
   //   unknown type name 'type_t'
   // that imply an unresolved typedef of type_t. This cannot be done below in
   // clang_visitChildren since clang does not have the unknown type names.
-  const std::string unknown_type_msg = "unknown type name \'";
-  for (auto &msg : diag_msgs)
+  for (const auto &msg : diag_msgs)
   {
-    if (msg.find(unknown_type_msg) == 0)
-    {
-      type_data.incomplete_types.emplace(
-          msg.substr(unknown_type_msg.length(),
-                     msg.length() - unknown_type_msg.length() - 1));
-    }
+    auto unknown_type = get_unknown_type(msg);
+    if (unknown_type)
+      type_data.incomplete_types.emplace(unknown_type.value());
   }
 
   CXCursor cursor = handler.get_translation_unit_cursor();
@@ -667,11 +656,40 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     return false;
   }
 
-  if (!handler.check_diagnostics(input))
+  std::vector<std::string> error_msgs;
+  if (!handler.check_diagnostics(input, error_msgs, true))
+  {
+    for (auto &msg : error_msgs)
+    {
+      if (get_unknown_type(msg) != "" && !bpftrace.force_btf_)
+      {
+        LOG(ERROR) << "Try running with --btf to force BTF processing or "
+                      "include headers with missing type definitions";
+      }
+    }
     return false;
+  }
 
   CXCursor cursor = handler.get_translation_unit_cursor();
   return visit_children(cursor, bpftrace);
+}
+
+/*
+ * Parse the given Clang diagnostics message and if it has the form:
+ *   unknown type name 'type_t'
+ * return type_t.
+ */
+std::optional<std::string> ClangParser::ClangParser::get_unknown_type(
+    const std::string &diagnostic_msg)
+{
+  const std::string unknown_type_msg = "unknown type name \'";
+  if (diagnostic_msg.find(unknown_type_msg) == 0)
+  {
+    return diagnostic_msg.substr(unknown_type_msg.length(),
+                                 diagnostic_msg.length() -
+                                     unknown_type_msg.length() - 1);
+  }
+  return {};
 }
 
 } // namespace bpftrace
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -363,20 +363,34 @@ CXErrorCode ClangParser::ClangParserHandler::parse_translation_unit(
 
 bool ClangParser::ClangParserHandler::check_diagnostics(
     const std::string &input,
+    std::vector<std::string> &error_msgs,
     bool bail_on_error)
 {
   for (unsigned int i=0; i < clang_getNumDiagnostics(get_translation_unit()); i++) {
     CXDiagnostic diag = clang_getDiagnostic(get_translation_unit(), i);
     CXDiagnosticSeverity severity = clang_getDiagnosticSeverity(diag);
+    std::string msg = clang_getCString(clang_getDiagnosticSpelling(diag));
+    error_msgs.push_back(msg);
+
     if ((bail_on_error && severity == CXDiagnostic_Error) ||
         severity == CXDiagnostic_Fatal)
     {
+      // Do not fail on "too many errors"
+      if (!bail_on_error && msg == "too many errors emitted, stopping now")
+        return true;
       if (bt_debug >= DebugLevel::kDebug)
         LOG(ERROR) << "Input (" << input.size() << "): " << input;
       return false;
     }
   }
   return true;
+}
+
+bool ClangParser::ClangParserHandler::check_diagnostics(
+    const std::string &input)
+{
+  std::vector<std::string> msgs;
+  return check_diagnostics(input, msgs, true);
 }
 
 CXCursor ClangParser::ClangParserHandler::get_translation_unit_cursor() {
@@ -461,7 +475,8 @@ bool ClangParser::visit_children(CXCursor &cursor, BPFtrace &bpftrace)
 std::unordered_set<std::string> ClangParser::get_incomplete_types(
     const std::string &input,
     std::vector<CXUnsavedFile> &unsaved_files,
-    const std::vector<const char *> &args)
+    const std::vector<const char *> &args,
+    const std::unordered_set<std::string> &complete_types)
 {
   if (input.empty())
     return {};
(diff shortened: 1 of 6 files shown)


mmisono:
I found that if a struct has a member whose type is not in the BTF, then bpftrace hangs (though probably this does not happen in most cases.)

```
% ./src/bpftrace -e 'struct foo {aaa a;};  BEGIN { @ = ((struct foo*)0)->a}' -b
// hang
```

> I was also wondering, if it would make sense to get rid of the --btf option completely. 

I also think using BTF whenever it is available is reasonable. Is there any downside? possible slowdown?

viktormalik:
> I found that if a struct has a member whose type is not in the BTF, then bpftrace hangs (though probably this does not happen in most cases.)

Correct, thanks for noticing. Even though this shouldn't happen for tracepoints, hanging is not good. I fixed it, now it terminates and shows the corresponding error message.

> I also think using BTF whenever it is available is reasonable. Is there any downside? possible slowdown?

I didn't notice any slowdown, in my local setup, the run-times are almost equal with and without BTF. Also, we're using BTF for all other probe categories, where a larger number of probes can be attached at a time (thus more BTF lookups), so I don't see any reason not to use it for tracepoints, too.

src/clang_parser.cpp
```diff
@@ -633,11 +661,40 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     return false;
   }
 
-  if (!handler.check_diagnostics(input))
+  std::vector<std::string> error_msgs;
+  if (!handler.check_diagnostics(input, error_msgs, true))
+  {
+    for (auto &msg : error_msgs)
+    {
+      if (get_unknown_type(msg) != "")
```
mmisono, line 669:
```suggestion
      if (get_unknown_type(msg) != "" && !bpftrace.force_btf_)
```
just to avoid a redundant message (if we leave `force_btf_` option.)

mmisono (commented):
Looks good to me, though it's better to have another review. I think it is OK to remove `--btf`.

danobi (commented):
LGTM overall with small comments, but my only concern is the string parsing of error messages. Not sure if error strings are part of clang API but I'd be worried if it wasn't. I didn't look around too hard in clang for alternative solutions so I'll take it on faith this is the best way.

danobi:
> I was also wondering, if it would make sense to get rid of the --btf option completely. AFAIU, it is used for tracepoints only. Using BTF automatically whenever it is available would make most of the "unknown type name" errors disappear in default bpftrace setup.

Tentative +1. I remember I tried doing this a couple times but I always hit some kind of realization that it would cause an issue. Can't seem to remember right now. It's possible that I've chipped away at reducing internal usage of `--btf` that there's no issue anymore.

src/clang_parser.cpp
```diff
@@ -672,11 +661,40 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
...
   CXCursor cursor = handler.get_translation_unit_cursor();
   return visit_children(cursor, bpftrace);
 }
 
+/*
+ * Parse the given Clang diagnostics message and if it has the form:
+ *   unknown type name 'type_t'
+ * return type_t.
+ */
+std::string ClangParser::ClangParser::get_unknown_type(
```
danobi, line 687: Returning an optional instead of empty string would be cleaner API
```suggestion
std::optional<std::string> ClangParser::ClangParser::get_unknown_type(
```

src/clang_parser.cpp
```diff
@@ -584,46 +619,50 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
...
-        "definitions.h",
-        args.data(),
-        args.size(),
-        input_files.data(),
-        input_files.size(),
-        CXTranslationUnit_DetailedPreprocessingRecord);
+      };
+      // If additional BTF types were found, we need to repeat the process since
+      // that might have introduced some new unresolved typedefs.
+      check_additional_types = types_cnt != bpftrace.btf_set_.size();
```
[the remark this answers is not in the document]
viktormalik: Not really, we need to check if a **new** incomplete type was found (one that wasn't in `bpftrace.btf_set_` already), otherwise the loop could not terminate (see the comment above for example).

src/clang_parser.cpp
```diff
@@ -584,46 +619,50 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
...
-                       .Contents = btf_and_input.c_str(),
-                       .Length = btf_and_input.size() });
-
-    auto btf_cdef = bpftrace.btf_.c_def(bpftrace.btf_set_);
-    btf_and_input_files.emplace_back(CXUnsavedFile{
+    auto incomplete_types = get_incomplete_types(
+        input, input_files, args, bpftrace.btf_set_);
+    unsigned types_cnt = bpftrace.btf_set_.size();
+    bpftrace.btf_set_.insert(incomplete_types.cbegin(),
+                             incomplete_types.cend());
```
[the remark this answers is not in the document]
viktormalik: They weren't there before this PR, but I agree - `bpftrace.btf_set_` should only be used when BTF is on.

viktormalik:
> my only concern is the string parsing of error messages. Not sure if error strings are part of clang API but I'd be worried if it wasn't. I didn't look around too hard in clang for alternative solutions so I'll take it on faith this is the best way.

I don't like this either. I tried to find an API for this, but it seems that there is none. Clang parser replaces all unknown types by `int` and it seems that it looses the information about the original type name. The only option could be extracting the information from [Clang diagnostics](https://clang.llvm.org/docs/InternalsManual.html#the-diagnostics-subsystem) (which is what we're doing here, anyway), but it doesn't seem that other parts of the diagnostics are more stable than the message itself.

viktormalik:
> Tentative +1. I remember I tried doing this a couple times but I always hit some kind of realization that it would cause an issue. Can't seem to remember right now. It's possible that I've chipped away at reducing internal usage of `--btf` that there's no issue anymore.

The problem will come if the user tries to define a type that is also parsed from BTF. For example this program:

    # bpftrace -e 'struct task_struct {int x;} i:ms:1 { printf("%d\n", curtask->x); }'

won't work because using `curtask` will force parsing `struct task_struct` from BTF and it will be then defined twice. I don't see a reason for doing this, but I may be missing something.

Anyway, I have a proposal implementation ready. Once this PR is merged, I'll open a new PR where we can discuss further.

mmisono (approved):

src/clang_parser.cpp
```diff
@@ -584,46 +607,46 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
...
+    btf_cdef = bpftrace.btf_.c_def(bpftrace.btf_set_);
+    input_files.back() = CXUnsavedFile{
+      .Filename = "/bpftrace/include/__btf_generated_header.h",
+      .Contents = btf_cdef.c_str(),
+      .Length = btf_cdef.size(),
+    };
+
+    // If additional BTF types were found, we need to repeat the process since
+    // that might have introduced some new unresolved typedefs.
+    check_additional_types = types_cnt != bpftrace.btf_set_.size();
```
mmisono, line 638: @viktormalik 
I found this PR causes a slowdown if an unknown struct is big. For example:

before this PR

```
% time sudo  ./src/bpftrace --btf -e 'struct f { struct task_struct x; } BEGIN{exit();}'
Attaching 1 probe..

0.94s user 0.20s system 80% cpu 1.420 tota
```

after
```
% time sudo ./src/bpftrace --btf -e 'struct f { struct task_struct x; } BEGIN{exit();}'
Attaching 1 probe...

7.76s user 0.59s system 96% cpu 8.646 total
```

The problem is this commented part, and for `task_struct`, this loop iterates 9 times. I wonder if we can reduce the iteration. Actually, I'm not sure when we need to iterate until the condition is satisfied. At least, the above example works without any iteration. Do you have an example?

viktormalik: Thanks for noticing this problem. The truth is that I don't have any example that would require multiple iterations, but it seems to me that it could, in theory, occur.

However, what I realized now, is that we run this loop to make sure that the Clang parser succeeds. So I believe that we can stop collecting the incomplete types once the Clang parser runs without errors. I implemented a simple fix in [this branch](https://github.com/viktormalik/bpftrace/tree/undef-types-speedup). Could you please try it and confirm that the slowdown is gone? I'll run some tests myself and then I can open a PR.

mmisono: Yes, it drastically improves speed. 

```
% time sudo ./src/bpftrace --btf -e 'struct f { struct task_struct x; } BEGIN{exit();}'
Attaching 1 probe...


0.47s user 0.27s system 72% cpu 1.009 total
```

Please open a PR and let's discuss this further there.

viktormalik: Great, thanks. 
It's not that easy after all, because this simple fix breaks one regression test. The problem is that the type resolution must be run at least once. I'll come with a better fix and open a PR.

PR #1486, opened 2020-08-26
ustd: better handling of empty namespaces

### LINKED ISSUE #1466: Code not generated for probe: usdt:/full-path:testprobe:testprobe
I write a example, and the executable file has tracepint
```
$ readelf -n ./tracetest 

Displaying notes found at file offset 0x00000254 with length 0x00000020:
  GNU                  0x00000010       NT_GNU_ABI_TAG (ABI version tag)
    OS: Linux, ABI: 2.6.32

Displaying notes found at file offset 0x00000274 with length 0x00000024:
  GNU                  0x00000014       NT_GNU_BUILD_ID (unique build ID bitstring)
    Build ID: b324052bdf781fe5348d1322a4beb249904d4ad8

Displaying notes found at file offset 0x00001080 with length 0x00000048:
  stapsdt              0x00000034       NT_STAPSDT (SystemTap probe descriptors)
    Provider: tracetest
    Name: testprobe
    Location: 0x0000000000400612, Base: 0x00000000004006e4, Semaphore: 0x0000000000000000
    Arguments: -8@%rax
```

I run bpftrace, show error
```
$ sudo bpftrace -e 'usdt:/home/ace/c/dtrace/tracetest:testprobe {printf("%d: time is %d\n", pid, arg0); }'  
Attaching 1 probe...
Code not generated for probe: usdt:/home/ace/c/dtrace/tracetest:testprobe:testprobe
```

My environment is ubuntu 16.04, and install bpftrace by snap, I not find answer on google

### LINKED ISSUE #1469: usdt: what the default namespace should be?
This is a question that follows #1466.

According to the documentation, it is possible to omit namespace when specifying usdt probe:
>Syntax:
>
>```
>usdt:binary_path:probe_name
>usdt:binary_path:[probe_namespace]:probe_name
>usdt:library_path:probe_name
>usdt:library_path:[probe_namespace]:probe_name
>```
>
>Where the `probe_namespace` is optional, and will default to the basename of the binary or library path.

This suggests that in such case, namespace is set to the basename of the binary. 

On the other hand, the code in [bpftrace.cpp#L366-L367](https://github.com/iovisor/bpftrace/blob/722c5c73ebdc3b87ea525dcf3f8573179bf1d7c8/src/bpftrace.cpp#L366-L367) indicates that the empty namespace should be expanded as if `*` was provided. This behaviour is also supported by a test in [tests/bpftrace.cpp#L541](https://github.com/iovisor/bpftrace/blob/722c5c73ebdc3b87ea525dcf3f8573179bf1d7c8/tests/bpftrace.cpp#L541).

My question is, what do you think the correct behaviour should be? I'd suggest the expansion, but that currently doesn't work since it is not handled correctly in all places in the code, causing the mentioned bug.

Any thoughts?

viktormalik:
If an empty namespace is given, it is deduced automatically. We assume that a single probe with the given name exists. If multiple probes having the same name but different namespaces are matched, no probes are attached to and the user is told to use an explicit wildcard for the namespace.

This solves dicrepancy between the tests (which treat an empty namespace as a wildcard) and the reference guide (which says that an empty namespace is replaced by the binary name).

Resolves #1469 and fixes #1466.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to
   - [#1439](https://github.com/iovisor/bpftrace/pull/1439)
 - Improve error messages for kfunc probe types
   - [#1451](https://github.com/iovisor/bpftrace/pull/1451)
+- Better handling of empty usdt namespaces
+  - [#1486](https://github.com/iovisor/bpftrace/pull/1486)
 
 #### Deprecated
 
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1224,7 +1224,7 @@ usdt:library_path:probe_name
 usdt:library_path:[probe_namespace]:probe_name
 ```
 
-Where the `probe_namespace` is optional, and will default to the basename of the binary or library path.
+Where `probe_namespace` is optional if `probe_name` is unique within the binary.
 
 Examples:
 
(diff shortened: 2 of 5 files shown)


fbs:
nice

PR #1490, opened 2020-08-28
execsnoop.bt: hook up onto all execve calls

Hi-Angel:
Besides execve there's also execveat event. Hook up onto both of them and also any other that could get possibly added by using a `*`.

##### Checklist

- [ x ] Language changes are updated in `docs/reference_guide.md`
- [ ✓ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ x ] The new behaviour is covered by tests


Hi-Angel:
> LGTM, thanks, but shouldn't the changelog entry be in "Tools"?

Thanks, done!

PR #1492, opened 2020-08-30
Add path builtin function

olsajiri:
Adding dpath builtin function that returns full path
referenced by struct path pointer in argument.

`  dpath(struct path *path)`

Example:

```
  # bpftrace  -e 'kfunc:filp_close { printf("%s\n", dpath(args->filp->f_path)); }'
  Attaching 1 probe...
  /proc/sys/net/ipv6/conf/eno2/disable_ipv6
  /proc/sys/net/ipv6/conf/eno2/use_tempaddr
  socket:[23276]
  /proc/sys/net/ipv6/conf/eno2/disable_ipv6
  socket:[17655]
  /sys/devices/pci0000:00/0000:00:1c.5/0000:04:00.1/net/eno2/type
  socket:[38745]
  /proc/sys/net/ipv6/conf/eno2/disable_ipv6
```

The dpath returns the string with the path or empty string
on error. I don't think it's necessary to return error code
at the moment, however it can be added later in the second
optional argument.

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -90,6 +90,7 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [22. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
     - [23. `print()`: Print Value](#23-print-print-value)
     - [24. `strftime()`: Formatted timestamp](#24-strftime-formatted-timestamp)
+    - [25. `path()`: Return full path](#25-path-return-full-path)
 - [Map Functions](#map-functions)
     - [1. Builtins](#1-builtins-2)
     - [2. `count()`: Count](#2-count-count)
@@ -1958,6 +1959,7 @@ Tracing block I/O sizes > 0 bytes
 - `signal(char[] signal | u32 signal)` - Send a signal to the current task
 - `strncmp(char *s1, char *s2, int length)` - Compare first n characters of two strings
 - `override(u64 rc)` - Override return value
+- `path(struct path *path)` - Return full path
 
 Some of these are asynchronous: the kernel queues the event, but some time later (milliseconds) it is
 processed in user-space. The asynchronous actions are: `printf()`, `time()`, and `join()`. Both `ksym()`
@@ -2750,6 +2752,33 @@ Attaching 1 probe...
 13:11:25
 13:11:26
 ^C
+```
+
+## 25. `path()`: Return full path
+
+Syntax:
+- `path(struct path *path)`
+
+Return full path referenced by struct path pointer in argument.
+There's list of allowed kernel functions, that can use this
+helper in probe.
+
+Examples:
+```
+# bpftrace  -e 'kfunc:filp_close { printf("%s\n", path(args->filp->f_path)); }'
+Attaching 1 probe...
+/proc/sys/net/ipv6/conf/eno2/disable_ipv6
+/proc/sys/net/ipv6/conf/eno2/use_tempaddr
+socket:[23276]
+/proc/sys/net/ipv6/conf/eno2/disable_ipv6
+socket:[17655]
+/sys/devices/pci0000:00/0000:00:1c.5/0000:04:00.1/net/eno2/type
+socket:[38745]
+/proc/sys/net/ipv6/conf/eno2/disable_ipv6
+
+# bpftrace  -e 'kretfunc:dentry_open { printf("%s\n", path(retval->f_path)); }'
+Attaching 1 probe...
+/dev/pts/1 -> /dev/pts/1
 ```
 
 # Map Functions
--- a/src/bpffeature.cpp
+++ b/src/bpffeature.cpp
@@ -66,14 +66,15 @@ static bool try_load(const char* name,
 
 static bool try_load(enum libbpf::bpf_prog_type prog_type,
                      struct bpf_insn* insns,
-                     size_t len)
+                     size_t len,
+                     const char* name = nullptr)
 {
   constexpr int log_size = 4096;
   char logbuf[log_size] = {};
 
   // kfunc / kretfunc only for now. We can refactor if more attach types
   // get added to BPF_PROG_TYPE_TRACING
-  if (prog_type == libbpf::BPF_PROG_TYPE_TRACING)
+  if (prog_type == libbpf::BPF_PROG_TYPE_TRACING && !name)
   {
     // List of available functions must be readable
     std::ifstream traceable_funcs(kprobe_path);
@@ -88,7 +89,7 @@ static bool try_load(enum libbpf::bpf_prog_type prog_type,
                "kretfunc__strlen", prog_type, insns, len, 0, logbuf, log_size);
   }
 
-  return try_load(nullptr, prog_type, insns, len, 0, logbuf, log_size);
+  return try_load(name, prog_type, insns, len, 0, logbuf, log_size);
 }
 
 bool BPFfeature::detect_helper(enum libbpf::bpf_func_id func_id,
@@ -278,6 +279,31 @@ bool BPFfeature::has_map_batch()
 #endif
 }
 
+bool BPFfeature::has_d_path(void)
+{
+  if (has_d_path_.has_value())
+    return *has_d_path_;
+
+  struct bpf_insn insns[] = {
+    BPF_LDX_MEM(BPF_W, BPF_REG_1, BPF_REG_1, 0),
+    BPF_MOV64_REG(BPF_REG_2, BPF_REG_10),
+    BPF_ALU64_IMM(BPF_ADD, BPF_REG_2, -8),
+    BPF_MOV64_IMM(BPF_REG_6, 0),
+    BPF_STX_MEM(BPF_DW, BPF_REG_2, BPF_REG_6, 0),
+    BPF_LD_IMM64(BPF_REG_3, 8),
+    BPF_RAW_INSN(BPF_JMP | BPF_CALL, 0, 0, 0, libbpf::BPF_FUNC_d_path),
+    BPF_MOV64_IMM(BPF_REG_0, 0),
+    BPF_EXIT_INSN(),
+  };
+
+  has_d_path_ = std::make_optional<bool>(try_load(libbpf::BPF_PROG_TYPE_TRACING,
+                                                  insns,
+                                                  ARRAY_SIZE(insns),
+                                                  "kfunc__dentry_open"));
+
+  return *has_d_path_;
+}
+
 std::string BPFfeature::report(void)
 {
   std::stringstream buf;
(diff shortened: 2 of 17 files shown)


src/bpffeature.cpp
```diff
@@ -232,6 +233,29 @@ int BPFfeature::instruction_limit(void)
   return *insns_limit_;
 }
 
+bool BPFfeature::has_d_path(void)
+{
+  if (has_d_path_.has_value())
+    return *has_d_path_;
+
+  struct bpf_insn insns[] = {
```
fbs, line 241: Why does this need a custom program? doesn't the helper detection work here?

fbs:
Nice! Glad this helper finally landed

src/bpffeature.cpp
olsajiri: aah, I try to load the working program and check that try_load succeeds,
but I did not noticed that the detection stuff checks for 'invalid' string,
so the load can fail for other reasons but still detect the helper.. that might
actually work.. I'll try it ;-)

thanks

danobi:
Couple things:
* are we locked into "dpath" name? I feel like "path" makes more sense here
* last I heard bpf_d_path() works for whitelisted functions in the kernel. is that still the case? if so, I think it'd be important to return error code

wilsonwr:
> * are we locked into "dpath" name? I feel like "path" makes more sense here

Personally, I appreciated it being called "dpath" when I saw this pull request. It helped me quickly understand that this function is like [d_path()](https://www.kernel.org/doc/htmldocs/filesystems/API-d-path.html) of the Linux kernel's filesystem API. Just my two cents.

danobi:
> > * are we locked into "dpath" name? I feel like "path" makes more sense here
> 
> Personally, I appreciated it being called "dpath" when I saw this pull request. It helped me quickly understand that this function is like [d_path()](https://www.kernel.org/doc/htmldocs/filesystems/API-d-path.html) of the Linux kernel's filesystem API. Just my two cents.

Yeah, I buy that. @olsajiri also made a good point in the IRC that fd_path() might still come one day.

brendangregg:
If possible, we shouldn't wed functions to Linux internals. There's already interest to see bpftrace on BSD and other OSes.

For example, I think an fdpath() function makes sense across kernels, since they all have a notion of an FD (POSIX).

This is trickier since it's for Linux struct dentry. I guess options are:

A) We go with dpath() and the docs say "this is a Linux specific function." BSD might add a vpath() to its port, etc.

B) We go with path() and the docs say "On Linux, this turns a dentry to a path; on BSD, this tuns a vnode to a path." etc.

C) The docs say "dpath() is an internal function that you should not use directly," and then we add a fdpath() wrapper that maps to code for doing the FD to dentry and then dpath() call (I'm sure I have examples of FD->dentry in the BPF book). Then we document fdpath() and it should work once other OSes have their own implementation.

D) We turn this PR into fdpath() with the FD->dentry code built in. :) 

(C/D) Solves the problem of forking scripts and documentation over something trivial: path lookups, which will be used by a bunch of one-liners. It would be nice if basic scripts/documentation didn't fork so quickly (it will be necessary for deeper things of course).

Just thinking of a one-liner one might see in a hello-world-level tutorial:

```
# bpftrace -e 't:syscalls:sys_enter_read { @paths[fdpath(args->fd)] = count(); }'
```

(probe types [t for tracepoint] and their arguments are of course currently Linux-influenced; although if other OSes just use the abbreviations it mostly works: k: for kernel dynamic instrumentation, u: for user-level dynamic instrumentation, U: for user-level static instrumentation, and t: for...hang on, perhaps we should add a K: alternate alias for for kernel static instrumentation, and then we have an OS-agnostic set!).

@olsajiri what ended up in the whitelist? syscall tracepoints, ...?

javierhonduco:
This is very exciting! 😄 cc @tyroguru

src/bpffeature.cpp
olsajiri: hm, the tracing program fails earlier with:
  Tracing programs must provide btf_id

so we need to have special function for that

olsajiri:
The latest version renames  'dpath' to 'path', could you please check? thanks  @fbs @danobi @brendangregg

fbs:
Looks ok

Can you add a runtime test for this? By doing something like 5441336f78fd9b9ae527e9c104d02da8cd433285 you can limit the tests to systems that support it:

```
NAME openat uptr
RUN bpftrace -v -e 't:syscalls:sys_enter_openat /comm == "syscall"/ { print(str(uptr(args->filename))) }' -c "./testprogs/syscall openat"
EXPECT bpftrace_runtime_test_syscall
REQUIRES_FEATURE probe_read_kernel
TIMEOUT 5
```

olsajiri:
> Looks ok +1
> 
> Can you add a runtime test for this? By doing something like [5441336](https://github.com/iovisor/bpftrace/commit/5441336f78fd9b9ae527e9c104d02da8cd433285) you can limit the tests to systems that support it:
> 
> ```
> NAME openat uptr
> RUN bpftrace -v -e 't:syscalls:sys_enter_openat /comm == "syscall"/ { print(str(uptr(args->filename))) }' -c "./testprogs/syscall openat"
> EXPECT bpftrace_runtime_test_syscall
> REQUIRES_FEATURE probe_read_kernel
> TIMEOUT 5
> ```

ah, did not see REQUIRES_FEATURE, nice.. will add

fbs (approved):

fbs:
@brendangregg @mmisono @danobi . Code looks ok (but I don't have a good btf test system atm). Any thoughts on this one?

mmisono:
BPF_FUNC_d_path is not merged in mainline, isn't it?

`bpftrace_test` does not run w/o root.

```
/home/ubuntu/work/bpftrace/bpftrace/tests/semantic_analyser.cpp:70: Failure
  fields.analyse()
    Which is: 1
  expected_field_analyser
    Which is: 0

Input:
kfunc:filp_close { $k = path( args->filp->f_path ) }

Output:
stdin:1:1-17: ERROR: kfunc:filp_close: could not read traceable functions from /sys/kernel/debug/tracing/available_filter_functions (is debugfs mounted?)
kfunc:filp_close { $k = path( args->filp->f_path ) }
~~~~~~~~~~~~~~~~

/home/ubuntu/work/bpftrace/bpftrace/tests/semantic_analyser.cpp:70: Failure
  fields.analyse()
    Which is: 1
  expected_field_analyser
    Which is: 0

Input:
kretfunc:fget { $k = path( retval->f_path ) }

Output:
stdin:1:1-14: ERROR: kfunc:fget: could not read traceable functions from /sys/kernel/debug/tracing/available_filter_functions (is debugfs mounted?)
kretfunc:fget { $k = path( retval->f_path ) }
~~~~~~~~~~~~~

[  FAILED  ] semantic_analyser.call_path (788 ms)
```

mmisono:
@olsajiri 
Could you include the error checking code? https://github.com/mmisono/bpftrace/commit/bddb3c24215b5aed5544821bcb0ffeabdb6db600

olsajiri:
> @olsajiri
> Could you include the error checking code? [mmisono@bddb3c2](https://github.com/mmisono/bpftrace/commit/bddb3c24215b5aed5544821bcb0ffeabdb6db600)
> (I assume that dpath return <0 if error)

sure, will fix that, thnx

tests/semantic_analyser.cpp
```diff
@@ -791,6 +791,24 @@ TEST(semantic_analyser, call_stack)
   test("kprobe:f { @x = 3; ustack(perf, @x) }", 10);
 }
 
+TEST(semantic_analyser, call_path)
+{
+  BPFfeature feature;
```
fbs, line 796:
```suggestion
```

MockBPFFeature handles this part. Allows us to run tests without root.

olsajiri:
> > @olsajiri
> > Could you include the error checking code? [mmisono@bddb3c2](https://github.com/mmisono/bpftrace/commit/bddb3c24215b5aed5544821bcb0ffeabdb6db600)
> > (I assume that dpath return <0 if error)
> 
> sure, will fix that, thnx

I checked and we don't check for error in there, it returns void,
but I should have used createCall call instead

mmisono:
> > > @olsajiri
> > > Could you include the error checking code? [mmisono@bddb3c2](https://github.com/mmisono/bpftrace/commit/bddb3c24215b5aed5544821bcb0ffeabdb6db600)
> > > (I assume that dpath return <0 if error)
> > 
> > 
> > sure, will fix that, thnx
> 
> I checked and we don't check for error in there, it returns void,
> but I should have used createCall call instead

The commit is for reporting helper errors if any (the functionality introduced in #1276.) Or do you mean `bpf_d_path()` helper returns void?

olsajiri:
> > > > @olsajiri
> > > > Could you include the error checking code? [mmisono@bddb3c2](https://github.com/mmisono/bpftrace/commit/bddb3c24215b5aed5544821bcb0ffeabdb6db600)
> > > > (I assume that dpath return <0 if error)
> > > 
> > > 
> > > sure, will fix that, thnx
> > 
> > 
> > I checked and we don't check for error in there, it returns void,
> > but I should have used createCall call instead
> 
> The commit is for reporting helper errors if any (the functionality introduced in #1276.) Or do you mean `bpf_d_path()` helper returns void?

ah I misunderstood, sry.. yes, it's good change, I'll add it, thanks

tests/semantic_analyser.cpp
```diff
@@ -1892,6 +1892,21 @@ TEST_F(semantic_analyser_btf, short_name)
   test("fr:func_1 { 1 }", 0);
 }
 
+TEST_F(semantic_analyser_btf, call_path)
```
mmisono, line 1895: The test runs only when HAVE_LIBBPF_BTF_DUMP. So I think it would be better to split this into for kfuncs and for others.

olsajiri: ok

docs/reference_guide.md
```diff
@@ -2700,6 +2701,25 @@ Attaching 1 probe...
...
+socket:[23276]
+/proc/sys/net/ipv6/conf/eno2/disable_ipv6
+socket:[17655]
+/sys/devices/pci0000:00/0000:00:1c.5/0000:04:00.1/net/eno2/type
+socket:[38745]
+/proc/sys/net/ipv6/conf/eno2/disable_ipv6
+
+# bpftrace  -e 'kretfunc:dentry_open { printf("%s\n", path(retval->f_path)); }'
+Attaching 1 probe...
+/dev/pts/1 -> /dev/pts/1
```
danobi, line 2722: This ends up in the example codeblock of sizeof(). Also this should probably go at the end as number 25 (instead of reusing number 23)

danobi (approved):

docs/reference_guide.md
olsajiri: right, I reformatted it, thanks

fbs:
Can you rebase this on master? The addrspace stuff just landed.

docs/reference_guide.md
```diff
@@ -2752,6 +2754,31 @@ Attaching 1 probe...
 ^C
 ```
 
+## 25. `path()`: Return full path
+
+Syntax:
+- `path(struct path *path)`
+
+Return full path referenced by struct path pointer in argument.
```
danobi, line 2762: One more thing: maybe it'd be good to document there's a whilelist of kernel functions you can use this on. Could be confusing for users who get load errors when they have a correct handle to `struct path`.

olsajiri: right, I added a sentence about that.. I think we need some way to get the whitelist info from kernel, I'll try to add something

PR #1493, opened 2020-08-30
Safer getMapKey()

Birch-san:
_This is an attempt to carve off bits from https://github.com/iovisor/bpftrace/pull/1360 to make it easier to review._

Use `ScopedExprDeleter` to guarantee that key allocated by `getMapKey()` is freed after use. Relieve consumer from having to know details of how key was allocated.

##### Checklist

  - No language changes
  - No user-visible changes
  - Changes are trivial
  - Behaviour is equivalent to original

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -310,21 +310,20 @@ void CodegenLLVM::visit(Call &call)
   if (call.func == "count")
   {
     Map &map = *call.map;
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
     b_.CreateStore(b_.CreateAdd(oldval, b_.getInt64(1)), newval);
     b_.CreateMapUpdateElem(ctx_, map, key, newval, call.loc);
 
     // oldval can only be an integer so won't be in memory and doesn't need lifetime end
-    b_.CreateLifetimeEnd(key);
     b_.CreateLifetimeEnd(newval);
     expr_ = nullptr;
   }
   else if (call.func == "sum")
   {
     Map &map = *call.map;
-    AllocaInst *key = getMapKey(map);
+    auto [key, scoped_key_deleter] = getMapKey(map);
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
     AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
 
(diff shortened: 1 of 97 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -1647,28 +1640,26 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
   if (!expr_) // Some functions do the assignments themselves
     return;
 
-  Value *val, *expr;
-  expr = expr_;
-  AllocaInst *key = getMapKey(map);
```
fbs, line 1652: What is the reason for moving this mapkey down? It seems to cause a lot of codegen changes making it harder to review. Like tests/codegen/llvm/args_multiple_tracepoints_category_wild.ll  seems to only a order change, not a functional one

fbs:
Nice, trying to break up the other one is useful.

Expand  the commit message for the codegen change a bit? I'd like to keep the commit log useul

src/ast/codegen_llvm.cpp
Birch-san: yup, it's just an order change. I wanted to make the lifetime of the map key as short as possible. should I delete it? it's only a nice-to-have.

I think either way I expect a lot of test churn, because the scoped_deleter emits its lifetime-end in a later location than when we did it manually. should I try to control where it emits its lifetime-end, to further reduce test churn?

Birch-san:
> Expand the commit message for the codegen change a bit? I'd like to keep the commit log useul

Okay sure. think I'll first see what decision we come to about "changing the codegen to reduce the test churn", to know how to describe the commits.

Birch-san:
btw, any idea why the [LLVM 10 Release build](https://github.com/iovisor/bpftrace/pull/1493/checks?check_run_id=1048808208) failed? it's saying on `codegen.call_uptr`:

```diff
@@ -9,6 +9,6 @@
 define i64 @\"kprobe:f\"(i8* %0) section \"s_kprobe:f_1\" {
 entry:
-  %\"@_val\" = alloca i64
-  %\"@_key\" = alloca i64
+  %\"@_key\" = alloca i64
+  %\"@_val\" = alloca i64
   %deref = alloca i16
   %1 = bitcast i8* %0 to i64*
```

So, "move the `@val` alloca beneath the `@key` alloca". but that's actually exactly what I did:

https://github.com/iovisor/bpftrace/pull/1493/files#diff-5916ace37f833dad59e548448a2e1f56R9-R14

image

is the CI wrong? it's telling me I need to change the code, but the code already looks exactly as it suggests.

fbs:
Looks like you need to update call_uptr_1_LLVM-10.ll

src/ast/codegen_llvm.cpp
fbs: > yup, it's just an order change. I wanted to make the lifetime of the map key as short as possible. should I delete it? it's only a nice-to-have.

If it has a significant impact on the amoun t  of changes in this PR it might be better to split it to a separate pr (or commit) to make reviewing easier.

`scripts/compare_tool_elf.sh` might be useful here too, just to make sure existing tools aren't impacted in a bad way.

> I think either way I expect a lot of test churn, because the scoped_deleter emits its lifetime-end in a later location than when we did it manually. should I try to control where it emits its lifetime-end, to further reduce test churn?

Don't think thats worth the effort.

Birch-san:
Closed in favour of https://github.com/iovisor/bpftrace/pull/1780

PR #1497, opened 2020-08-31
ci: Cache docker images

danobi:
We're seeing more often issues with apt repos syncing and our CI
spuriously failing. We don't really need super up to date docker CI
images -- one a week is probably good enough.

This commit caches the docker build for a week if files in `docker/` are
not changed.

I tested this works by pushing this commit my personal repo, letting GH
actions run, then pushing a dummy commit and verifying cache works as
expected.

Also note that github only gives 5G of cache for free. Each image is
~300MB so that gives us ~15 images in cache. We'll probably see some
churn but hey, that's how caches work, right? Better than nothing.

##### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,14 +97,37 @@ jobs:
           BASE: bionic
     steps:
     - uses: actions/checkout@v2
+    - name: Get date
+      id: get-date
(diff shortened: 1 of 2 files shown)


danobi:
Alright I think it works. Ready for review

fbs:
Nice work :)

Restoring the container from cache seems quite slow, but still a decent speed up over building it from scratch.

Issue #1263, opened 2020-04-14
use std::filesystem


fbs:
Now that we have C++17 we can start making use of [std::filesystem](https://en.cppreference.com/w/cpp/filesystem) and start replacing  the `sprintf()` path concat and stat/access stuff with this.

Horki:
@fbs @brendangregg please test #1704

Horki:
@fbs this can be closed now right?

fbs:
I still see some stat/access and path concat happening in various places

jordalgo:
I think there may just be one left:
https://github.com/search?q=repo%3Aiovisor%2Fbpftrace+%3A%3Astat&type=code

PR #1499, opened 2020-09-02
Support multi-matched globbed target files for uprobe and usdt

### LINKED ISSUE #1113: Support multi-matched globbed target files for uprobe and usdt
Right now we require attach points target files to match a unique binary.
So `uprobe:/bin/globme*:asdf { printf("hit\n"); }` does not work if there exists
/bin/globme1 and /bin/globme2.

We should probably have uprobe/usdt attach to all globbed binaries b/c
it's a strange restriction to allow globs but disallow multiple attach.

viktormalik:
The change is pretty straightforward, it's mostly achieved by changing the form of strings returned by `find_wildcard_matches` for these probe types (the new form is `target:function` for uprobe and `target:namespace:probename` for usdt) and handling the new form where needed.

A noticeable behavior is for usdt probes when a process PID is specified. By default, all binaries and libraries executed by the process are searched for usdt probes. If a target path is specified along with a PID, only the binaries/libraries matching the path are searched. I wonder if this information should be put into the Reference Guide.

Resolves #1113.

Note: this change introduces usage of `std::filesystem`. This requires some changes, in particular adding the `-lstdc++fs` linker flag for some compiler versions and using GCC version 8 in Docker images (since GCC 7 does not support `std::filesystem`). These changes will probably be needed for #1263, too.

##### Checklist

### DIFF
--- a/.github/workflows/embedded.yml
+++ b/.github/workflows/embedded.yml
@@ -98,7 +98,7 @@ jobs:
           EMBED_BINUTILS: OFF
           RUN_ALL_TESTS: 1
           TEST_GROUPS_DISABLE: "tools-parsing-test"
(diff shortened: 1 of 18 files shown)


fbs:
Look ok! Adding std::filesystem should be fine, shame the detection has to happen like this but I don't see a better way either.

danobi (approved):

PR #1504, opened 2020-09-08
Set addrspace info for various builtins

sumanthkorikkar:
1. Usecases are described on https://github.com/iovisor/bpftrace/issues/614
2. https://github.com/iovisor/bpftrace/pull/1427

Implemented usecases:
- *addr(), str(addr) -> Use bpf_probe_read_kernel or
   bpf_probe_read_user based on addrspace context.
   Addrspace context:
   - kprobe, kretprobe, kfunc, kretfunc, tracepoints without syscalls
        -> bpf_probe_read_kernel
   - usdt, uprobes, uretprobes, usdt, tracepoint with syscalls
        -> bpf_probe_read_user
- *uptr(addr), str(uptr(addr)) -> use bpf_probe_read_user
- *kptr(addr), str(kptr(addr)) -> use bpf_probe_read_kernel

- argX, retval on kernel context -> bpf_probe_read_kernel
- argX, retval  on user context -> bpf_probe_read_user
- args->field -> bpf_probe_read_user

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -222,7 +222,7 @@ void CodegenLLVM::visit(Builtin &builtin)
                                           arg_num,
                                           builtin,
                                           bpftrace_.pid(),
-                                          AddrSpace::none,
+                                          AddrSpace::user,
                                           builtin.loc);
         return;
       }
@@ -587,11 +587,15 @@ void CodegenLLVM::visit(Call &call)
 
     auto scoped_del = accept(call.vargs->front());
     auto arg0 = call.vargs->front();
+    // arg0 is already on the bpf stack -> use probe kernel
+    // otherwise ->  addrspace of arg0->type
+    // case : struct MyStruct { char b[4]; };
+    // $s = (struct MyStruct *)arg0; buf($s->b, 4)
     b_.CreateProbeRead(ctx_,
                        static_cast<AllocaInst *>(buf_data_offset),
                        length,
                        expr_,
-                       arg0->type.GetAS(),
+                       find_addrspace_stack(arg0->type),
                        call.loc);
 
     expr_ = buf;
@@ -1388,6 +1392,7 @@ void CodegenLLVM::visit(Ternary &ternary)
 void CodegenLLVM::visit(FieldAccess &acc)
 {
   SizedType &type = acc.expr->type;
+  AddrSpace addrspace = acc.expr->type.GetAS();
   assert(type.IsRecordTy() || type.IsTupleTy());
   auto scoped_del = accept(acc.expr);
 
@@ -1430,6 +1435,9 @@ void CodegenLLVM::visit(FieldAccess &acc)
   type.is_tparg = is_tparg;
   type.is_internal = is_internal;
   type.is_kfarg = is_kfarg;
+  // Restore the addrspace info
+  // struct MyStruct { const int* a; };  $s = (struct MyStruct *)arg0;  $s->a
+  type.SetAS(addrspace);
 
   auto &field = cstruct.fields[acc.field];
 
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -115,7 +115,7 @@ void CodegenLLVM::visit(Builtin &builtin)
 {
   if (builtin.ident == "nsecs")
   {
-    expr_ = b_.CreateGetNs(bpftrace_.feature_.has_helper_ktime_get_boot_ns());
+    expr_ = b_.CreateGetNs(bpftrace_.feature_->has_helper_ktime_get_boot_ns());
   }
   else if (builtin.ident == "elapsed")
   {
@@ -126,7 +126,7 @@ void CodegenLLVM::visit(Builtin &builtin)
     auto type = CreateUInt64();
     auto start = b_.CreateMapLookupElem(
         ctx_, map->mapfd_, key, type, builtin.loc);
-    expr_ = b_.CreateGetNs(bpftrace_.feature_.has_helper_ktime_get_boot_ns());
+    expr_ = b_.CreateGetNs(bpftrace_.feature_->has_helper_ktime_get_boot_ns());
     expr_ = b_.CreateSub(expr_, start);
     // start won't be on stack, no need to LifeTimeEnd it
     b_.CreateLifetimeEnd(key);
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -48,7 +48,7 @@ libbpf::bpf_func_id IRBuilderBPF::selectProbeReadHelper(AddrSpace as, bool str)
 {
   libbpf::bpf_func_id fn;
   // Assume that if a kernel has probe_read_kernel it has the other 3 too
-  if (bpftrace_.feature_.has_helper_probe_read_kernel())
+  if (bpftrace_.feature_->has_helper_probe_read_kernel())
   {
     if (as == AddrSpace::kernel)
     {
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -65,6 +65,12 @@ libbpf::bpf_func_id IRBuilderBPF::selectProbeReadHelper(AddrSpace as, bool str)
       // if the kernel has the new helpers but AS is still none it is a bug
       // in bpftrace, assert catches it for debug builds.
       // assert(as != AddrSpace::none);
+      static bool warnonce = false;
+      if (!warnonce)
+      {
+        warnonce = true;
+        LOG(WARNING) << "Addrspace is not set";
+      }
       fn = str ? libbpf::BPF_FUNC_probe_read_str : libbpf::BPF_FUNC_probe_read;
     }
   }
@@ -594,7 +600,7 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
     {
       // Zero out `dst` in case we read less than 64 bits
       CreateStore(getInt64(0), dst);
-      CreateProbeRead(ctx, dst, abs_size, reg, as, loc);
+      CreateProbeRead(ctx, dst, abs_size, reg, AddrSpace::kernel, loc);
       result = CreateLoad(dst);
     }
     CreateLifetimeEnd(dst);
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -598,10 +598,8 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
     }
     else
     {
-      // Zero out `dst` in case we read less than 64 bits
-      CreateStore(getInt64(0), dst);
-      CreateProbeRead(ctx, dst, abs_size, reg, AddrSpace::kernel, loc);
-      result = CreateLoad(dst);
+      result = CreateLoad(GetType(CreateInt(abs_size * 8)), reg);
+      result = CreateIntCast(result, getInt64Ty(), argument->size < 0);
     }
     CreateLifetimeEnd(dst);
   }
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -60,6 +60,9 @@ void SemanticAnalyser::visit(PositionalParameter &param)
               << "$" << param.n << " used numerically but given \"" << pstr
               << "\". Try using str($" << param.n << ").";
         }
+        // string allocated in bpf stack. See codegen.
+        if (!is_numeric(pstr))
+          param.type.SetAS(AddrSpace::kernel);
       }
       break;
     case PositionalParameterType::count:
@@ -82,6 +85,8 @@ void SemanticAnalyser::visit(String &string)
                                  << " bytes): " << string.str;
   }
   string.type = CreateString(STRING_SIZE);
+  // @a = buf("hi", 2). String allocated on bpf stack. See codegen
+  string.type.SetAS(AddrSpace::kernel);
 }
 
 void SemanticAnalyser::visit(StackMode &mode)
@@ -137,8 +142,10 @@ void SemanticAnalyser::builtin_args_tracepoint(AttachPoint *attach_point,
     std::string tracepoint_struct = TracepointFormatParser::get_struct_name(
         match);
     Struct &cstruct = bpftrace_.structs_[tracepoint_struct];
-
-    builtin.type = CreatePointer(CreateRecord(cstruct.size, tracepoint_struct));
+    AddrSpace as = (attach_point->target == "syscalls") ? AddrSpace::user
+                                                        : AddrSpace::kernel;
+    builtin.type = CreatePointer(CreateRecord(cstruct.size, tracepoint_struct),
+                                 as);
     builtin.type.MarkCtxAccess();
     builtin.type.is_tparg = true;
   }
@@ -162,6 +169,35 @@ ProbeType SemanticAnalyser::single_provider_type(void)
   return type;
 }
 
+AddrSpace SemanticAnalyser::find_addrspace(ProbeType pt)
+{
+  switch (pt)
+  {
+    case ProbeType::kprobe:
+    case ProbeType::kretprobe:
+    case ProbeType::kfunc:
+    case ProbeType::kretfunc:
+    case ProbeType::tracepoint:
+      return AddrSpace::kernel;
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+    case ProbeType::usdt:
+      return AddrSpace::user;
+    // case : i:ms:1 (struct x*)ctx)->x
+    // Cannot decide the addrspace. Provide backward compatibility,
+    // if addrspace cannot be detected.
+    case ProbeType::invalid:
+    case ProbeType::profile:
+    case ProbeType::interval:
+    case ProbeType::software:
+    case ProbeType::hardware:
+    case ProbeType::watchpoint:
+      // Will trigger a warning in selectProbeReadHelper.
+      return AddrSpace::none;
+  }
+  return {}; // unreached
+}
+
 void SemanticAnalyser::visit(Builtin &builtin)
 {
   if (builtin.ident == "ctx")
@@ -181,7 +217,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
     {
       case BPF_PROG_TYPE_KPROBE:
         builtin.type = CreatePointer(
-            CreateRecord(bpftrace_.structs_["pt_regs"].size, "struct pt_regs"));
+            CreateRecord(bpftrace_.structs_["pt_regs"].size, "struct pt_regs"),
+            AddrSpace::kernel);
         builtin.type.MarkCtxAccess();
         break;
       case BPF_PROG_TYPE_TRACEPOINT:
@@ -191,7 +228,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
       case BPF_PROG_TYPE_PERF_EVENT:
         builtin.type = CreatePointer(
             CreateRecord(bpftrace_.structs_["bpf_perf_event_data"].size,
-                         "struct bpf_perf_event_data"));
+                         "struct bpf_perf_event_data"),
+            AddrSpace::kernel);
         builtin.type.MarkCtxAccess();
         break;
       default:
@@ -221,9 +259,10 @@ void SemanticAnalyser::visit(Builtin &builtin)
   else if (builtin.ident == "curtask")
   {
     /*
-     * Retype curtask to its original type: struct task_truct.
+     * Retype curtask to its original type: struct task_struct.
      */
-    builtin.type = CreatePointer(CreateRecord(0, "struct task_struct"));
+    builtin.type = CreatePointer(CreateRecord(0, "struct task_struct"),
+                                 AddrSpace::kernel);
   }
   else if (builtin.ident == "retval")
   {
@@ -250,6 +289,9 @@ void SemanticAnalyser::visit(Builtin &builtin)
           << (type == ProbeType::tracepoint ? " (try to use args->ret instead)"
                                             : "");
     }
+    // For kretprobe, kfunc, kretfunc -> AddrSpace::kernel
+    // For uretprobe -> AddrSpace::user
+    builtin.type.SetAS(find_addrspace(type));
   }
   else if (builtin.ident == "kstack") {
     builtin.type = CreateStack(true, StackType());
@@ -261,6 +303,9 @@ void SemanticAnalyser::visit(Builtin &builtin)
   }
   else if (builtin.ident == "comm") {
     builtin.type = CreateString(COMM_SIZE);
+    // comm allocated in the bpf stack. See codegen
+    // Case: @=comm and strncmp(@, "name")
+    builtin.type.SetAS(AddrSpace::kernel);
   }
   else if (builtin.ident == "func") {
     for (auto &attach_point : *probe_->attach_points)
@@ -279,6 +324,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
   }
   else if (!builtin.ident.compare(0, 3, "arg") && builtin.ident.size() == 4 &&
       builtin.ident.at(3) >= '0' && builtin.ident.at(3) <= '9') {
+    ProbeType pt = probetype((*probe_->attach_points)[0]->provider);
+    AddrSpace addrspace = find_addrspace(pt);
     for (auto &attach_point : *probe_->attach_points)
     {
       ProbeType type = probetype(attach_point->provider);
@@ -294,9 +341,12 @@ void SemanticAnalyser::visit(Builtin &builtin)
       LOG(ERROR, builtin.loc, err_)
           << arch::name() << " doesn't support " << builtin.ident;
     builtin.type = CreateUInt64();
+    builtin.type.SetAS(addrspace);
   }
   else if (!builtin.ident.compare(0, 4, "sarg") && builtin.ident.size() == 5 &&
       builtin.ident.at(4) >= '0' && builtin.ident.at(4) <= '9') {
+    ProbeType pt = probetype((*probe_->attach_points)[0]->provider);
+    AddrSpace addrspace = find_addrspace(pt);
     for (auto &attach_point : *probe_->attach_points)
     {
       ProbeType type = probetype(attach_point->provider);
@@ -314,6 +364,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
       }
     }
     builtin.type = CreateUInt64();
+    builtin.type.SetAS(addrspace);
   }
   else if (builtin.ident == "probe") {
     builtin.type = CreateProbe();
@@ -350,7 +401,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
     }
     else if (type == ProbeType::kfunc || type == ProbeType::kretfunc)
     {
-      builtin.type = CreatePointer(CreateRecord(0, "struct kfunc"));
+      builtin.type = CreatePointer(CreateRecord(0, "struct kfunc"),
+                                   AddrSpace::kernel);
       builtin.type.MarkCtxAccess();
       builtin.type.is_kfarg = true;
     }
@@ -555,6 +607,8 @@ void SemanticAnalyser::visit(Call &call)
         }
       }
 
+      // Required for cases like strncmp(str($1), str(2), 4))
+      call.type.SetAS(t.GetAS());
       if (is_final_pass() && call.vargs->size() > 1) {
         check_arg(call, Type::integer, 1, false);
       }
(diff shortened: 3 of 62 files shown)


sumanthkorikkar:
I will do few more minor changes here, before review. Thank you.

sumanthkorikkar:
Right now, I could successfully resolve to either bpf_probe_read_kernel/user for newer kernel. 
Two exceptions:
- bpftrace -e 'usdt::tracetest:testprobe { printf("%s\n", arg1 ); exit(); }' -p $(pidof usdt_semaphore_test)
- bpftrace -e 'usdt:./testprogs/usdt_semaphore_test:tracetest:testprobe { printf("%s\n", str(arg1) ); exit(); }' --usdt-file-activation

src/ast/semantic_analyser.cpp
```diff
@@ -701,14 +748,16 @@ void SemanticAnalyser::visit(Call &call)
...
-    call.type = CreateUInt64();
+    // Case : $a=*reg("ip"). Check provider and decide the addrspace.
+    ProbeType pt = probetype((*probe_->attach_points)[0]->provider);
+    call.type = CreatePointer(CreateUInt64(), find_addrspace(pt));
   }
   else if (call.func == "kaddr") {
     if (check_nargs(call, 1)) {
       check_arg(call, Type::string, 0, true);
     }
     call.type = CreateUInt64();
```
fbs, line 759:
```suggestion
```

src/ast/codegen_llvm.cpp
```diff
@@ -589,11 +589,15 @@ void CodegenLLVM::visit(Call &call)
 
     auto scoped_del = accept(call.vargs->front());
     auto arg0 = call.vargs->front();
+    // arg0 is already on the bpf stack -> use probe kernel
+    // otherwise ->  addrspace of arg0->type
+    // case : 'struct MyStruct { char b[4]; };
+    // $s = (struct MyStruct *)arg0; buf($s->b, 4)
```
fbs, line 595: We should get rid of the probe read here, doing it for onstack data is wrong (out of scope here tho)

src/ast/semantic_analyser.cpp
```diff
@@ -165,6 +173,25 @@ ProbeType SemanticAnalyser::single_provider_type(void)
...
+    case ProbeType::kfunc:
+    case ProbeType::kretfunc:
+    case ProbeType::tracepoint:
+      return AddrSpace::kernel;
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+    case ProbeType::usdt:
+      return AddrSpace::user;
+    default:
+      return AddrSpace::none;
```
fbs, line 191: should we abort here in case someone adds a new type but doesn't update this?

fbs (changes requested):
Nice! The inline comments are useful.

Does this handle syscall tracepoints correctly too? Data for those is usually coming from userspace, e.g. `t:syscalls:sys_enter_openat { @=str(args->filename); }`

sumanthkorikkar:
> Nice! The inline comments are useful.
> 
> Does this handle syscall tracepoints correctly too? Data for those is usually coming from userspace, e.g. `t:syscalls:sys_enter_openat { @=str(args->filename); }`

Yes, this is handled. Added a case in builtin_args_tracepoint()

src/ast/codegen_llvm.cpp
sumanthkorikkar: Right

src/ast/semantic_analyser.cpp
sumanthkorikkar: Ok. Will change the call.type to uint64 and also set the addrspace here. Needed in cases like uprobe:fn { @=sum(*(uint16*)(reg("sp")+8)); }

src/ast/semantic_analyser.cpp
sumanthkorikkar: Agree.

src/ast/semantic_analyser.cpp
sumanthkorikkar: Ok. Forgot to remove that.

sumanthkorikkar:
Question: @fbs, I used single_provider_type() to get the provider in certain cases and resolve the addrspace. If you can share some scenarios/examples where this could be different, that would be great.. Thnx

src/ast/semantic_analyser.cpp
fbs: This might cause a regression, #1507

src/ast/semantic_analyser.cpp
danobi: The compiler should be able to catch it for us (no `default` case to force exhaustive switch). I think we had some weird `-W` settings on that prevented doing that before. Should look again.

danobi: b20320b028ff is in now so we should drop the `default` here

sumanthkorikkar: ok. Will remove it.

sumanthkorikkar:
@fbs, @danobi 
bpftrace -e 'kretprobe:vfs_read,uretprobe:sample:read  { printf("%s ", retval); }'
- In this case, it is not possible to find the addrspace.
- So, when the probe type is invalid (single_provider_type()), I will have to set the AddrSpace to none (which get resolved to bpf_probe_read). Atleast, here we wont introduce regression for x86, if any applications are using this assumptions. In this case, we cant help for arch with split addrspace.
- I think, we should also define what combinations are allowed. From my point of view, we should ideally not allow this combination to happen and its not possible to know the addrspace.  
  - One can separtely use bpftrace -e kretprobe:vfs_read {} followed by bpftrace -e uretprobe:sample:read {}

Thanks.

sumanthkorikkar:
Ok. I see single_provider_type and checks are performed for "args" and "retval"

sumanthkorikkar:
- "reg" and "kaddr" use same type as before. Just to prevent regression, if any.
- AddrSpace::none is used as default, when addrspace cannot be detected. Also to provide backward compatibility. There might be some usecase, old applications are still using.
  - Also see tools/runqlen.bt, builtin.ctx runtime test.
- Now the codegen tests should work fine for both older kernel and newer kernel. Hence, we can probably enable probe_read_user/probe_read_kernel in ci.

src/ast/semantic_analyser.cpp
sumanthkorikkar: Kept the same type.

src/ast/semantic_analyser.cpp
sumanthkorikkar: Kept AddrSpace::none for now. See tools/runqlen.bt, builtin.ctx runtime test, i:ms:1 (struct x*)ctx)->x.

src/ast/irbuilderbpf.cpp
```diff
@@ -593,7 +593,8 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
     {
       // Zero out `dst` in case we read less than 64 bits
       CreateStore(getInt64(0), dst);
-      CreateProbeRead(ctx, dst, abs_size, reg, as, loc);
+      // TODO : probe_read_user does not work here.
```
sumanthkorikkar, line 596: Having a probe_read_kernel, works here. Not sure why.. (See usdt_semaphore test)

sumanthkorikkar:
- Kind ping. Request for review @fbs, @danobi . Thank you. 
- W.r.t qemu tcg + s390x, I tried booting and packaging using debootstrap.  Simple tasks works, Yet to come up with full working state along with bpftrace. I will try to come up with the solution soon. May be cross compilation of bpftrace would also help, if compilation takes long in emulated platform.

sumanthkorikkar:
Well, compilation does take a lot of time with qemu-tcg. Also runtime tests may timeout and might need some adjustment.

mmisono:
```
sudo ./src/bpftrace -e 'f:vfs_open { printf("open path: %s\n", str(args->path->dentry->d_name.name)); }'
```

This uses `bpf_probe_read()` and `bpf_probe_read_str()`. Isn't it supposed to use `bpf_probe_read_kern()` and `bpf_probe_read_kern_str()`?

<details> <summary> bytecode </summary>

```
Bytecode:
0: (bf) r6 = r1
1: (b7) r1 = 0
2: (7b) *(u64 *)(r10 -144) = r1
last_idx 2 first_idx 0
3: (7b) *(u64 *)(r10 -136) = r1
4: (7b) *(u64 *)(r10 -128) = r1
5: (7b) *(u64 *)(r10 -120) = r1
6: (7b) *(u64 *)(r10 -112) = r1
7: (7b) *(u64 *)(r10 -104) = r1
8: (7b) *(u64 *)(r10 -96) = r1
9: (7b) *(u64 *)(r10 -88) = r1
10: (7b) *(u64 *)(r10 -80) = r1
11: (7b) *(u64 *)(r10 -72) = r1
12: (7b) *(u64 *)(r10 -64) = r1
13: (7b) *(u64 *)(r10 -56) = r1
14: (7b) *(u64 *)(r10 -48) = r1
15: (7b) *(u64 *)(r10 -40) = r1
16: (7b) *(u64 *)(r10 -32) = r1
17: (7b) *(u64 *)(r10 -24) = r1
18: (7b) *(u64 *)(r10 -16) = r1
19: (79) r3 = *(u64 *)(r6 +0)
func 'vfs_open' arg0 has btf_id 1006 type STRUCT 'path'
20: (07) r3 += 8
21: (bf) r1 = r10
22: (07) r1 += -8
23: (b7) r2 = 8
24: (85) call bpf_probe_read#4
last_idx 24 first_idx 0
regs=4 stack=0 before 23: (b7) r2 = 8
25: (79) r3 = *(u64 *)(r10 -8)
26: (07) r3 += 40
27: (bf) r1 = r10
28: (07) r1 += -8
29: (b7) r2 = 8
30: (85) call bpf_probe_read#4
last_idx 30 first_idx 0
regs=4 stack=0 before 29: (b7) r2 = 8
31: (79) r3 = *(u64 *)(r10 -8)
32: (bf) r1 = r10
33: (07) r1 += -72
34: (b7) r2 = 64
35: (85) call bpf_probe_read_str#45
last_idx 35 first_idx 31
regs=4 stack=0 before 34: (b7) r2 = 64
36: (79) r1 = *(u64 *)(r10 -72)
37: (7b) *(u64 *)(r10 -136) = r1
38: (79) r1 = *(u64 *)(r10 -64)
39: (7b) *(u64 *)(r10 -128) = r1
40: (79) r1 = *(u64 *)(r10 -56)
41: (7b) *(u64 *)(r10 -120) = r1
42: (79) r1 = *(u64 *)(r10 -48)
43: (7b) *(u64 *)(r10 -112) = r1
44: (79) r1 = *(u64 *)(r10 -40)
45: (7b) *(u64 *)(r10 -104) = r1
46: (79) r1 = *(u64 *)(r10 -32)
47: (7b) *(u64 *)(r10 -96) = r1
48: (79) r1 = *(u64 *)(r10 -24)
49: (7b) *(u64 *)(r10 -88) = r1
50: (79) r1 = *(u64 *)(r10 -16)
51: (7b) *(u64 *)(r10 -80) = r1
52: (18) r7 = 0xffff942999432200
54: (85) call bpf_get_smp_processor_id#8
55: (bf) r4 = r10
56: (07) r4 += -144
57: (bf) r1 = r6
58: (bf) r2 = r7
59: (bf) r3 = r0
60: (b7) r5 = 72
61: (85) call bpf_perf_event_output#25
last_idx 61 first_idx 55
regs=20 stack=0 before 60: (b7) r5 = 72
62: (b7) r0 = 0
63: (95) exit
processed 63 insns (limit 1000000) max_states_per_insn 0 total_states 2 peak_states 2 mark_read 2
```

</details>

mmisono:
Another issue:

```
% sudo ./src/bpftrace  -e 'uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }' -kk
Attaching 1 probe...
stdin:1:49-54: WARNING: Failed to probe_read_kernel: Bad address (-14)
uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }
                                                ~~~~~
stdin:1:49-106: WARNING: Failed to probe_read_str: Bad address (-14)
uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }
                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
PS1:
```

sumanthkorikkar:
> ```
> sudo ./src/bpftrace -e 'f:vfs_open { printf("open path: %s\n", str(args->path->dentry->d_name.name)); }'
> ```
> 
> This uses `bpf_probe_read()` and `bpf_probe_read_str()`. Isn't it supposed to use `bpf_probe_read_kern()` and `bpf_probe_read_kern_str()`?
> bytecode

Nice. I will add a case in FieldAccess to resolve this.

sumanthkorikkar:
> Another issue:
> 
> ```
> % sudo ./src/bpftrace  -e 'uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }' -kk
> Attaching 1 probe...
> stdin:1:49-54: WARNING: Failed to probe_read_kernel: Bad address (-14)
> uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }
>                                                 ~~~~~
> stdin:1:49-106: WARNING: Failed to probe_read_str: Bad address (-14)
> uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }
>                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> PS1:
> ```
> 
> edit: I realized that since `curtask` is in kaddrspace, we need to use `uptr()` in this case.

Ok, Will change it. 
- Based on the provider type, curtask addrspace shall be selected.  
- If we use different providers in this case, then we end up in AddrSpace::none, which would translate to bpf_probe_read in x86.
- If this is not allowed, we can also add a additional check. 

Thank you

mmisono:
@sumanthkorikkar 
To be clear, it is OK that `curtask` has kaddrspace. The problem here is `curtask->mm->mmap->vm_start + uaddr("ps1_prompt")`  has kaddrspace but it should do uaddrspace (btw, `curtask->mm->mmap->vm_start` is added to cope with ASLR). I'm not sure how we should handle kaddrspace + uaddrspace in general, but making it addrspace::none would preserve the current behavior.

sumanthkorikkar:
> @sumanthkorikkar
> To be clear, it is OK that `curtask` has kaddrspace. The problem here is `curtask->mm->mmap->vm_start + uaddr("ps1_prompt")` has kaddrspace but it should do uaddrspace (btw, `curtask->mm->mmap->vm_start` is added to cope with ASLR). I'm not sure how we should handle kaddrspace + uaddrspace in general, but making it addrspace::none would preserve the current behavior.

Ok, correct. Could we have something like, For binop: ?
- If lhs or rhs is none, then take the addrspace which is set
- if both lhs, rhs is none, then addrspace is none.
- if lhs or rhs has different addrspace (not none), then set the addrspace to none -> to preserve the behaviour for x86. May be also provide a warning?

mmisono:
> Ok, correct. Could we have something like, For binop: ?
> 
> * If lhs or rhs is none, then take the addrspace which is set
> * if both lhs, rhs is none, then addrspace is none.
> * if lhs or rhs has different addrspace (not none), then set the addrspace to none -> to preserve the behaviour for x86. May be also provide a warning?

Seems reasonable to me. +1 for adding a warning if addrspace becomes none.

sumanthkorikkar:
@mmisono . Added changes. 
- For Binop, added usecases for mixed types. User is provided with warning once, if there is a addrspace mismatch for binop and the addrspace is set to None.
- When selecting the probe: If the new probe read functions are available and Addrspace is still None, then the warning is issued only once, instead of repeated warnings. [ example multiple deferences ]
- Added a missing usecase in FieldAccess [ kfarg ]


Thanks

src/ast/irbuilderbpf.cpp
mmisono: Here, we read the value (`reg`) in `struct pt_regs`, so we need to use `probe_read_kernel()`. But it is a BPF context data so I think we can simply load the value, i.e.,

```cpp
{
       result = CreateLoad(reg);
       result = CreateIntCast(result, getInt64Ty(), false);
}
```

sumanthkorikkar: :+1:  I will  use your load suggestion here.

fbs:
I'm not a fan of the ` bpftrace/tests: Rewrite codegen tests for probe read ` commit. How does it know when to use user over the _kernel version of the helper? 

With mockbpffeature we can control the `has_helper_probe_read_kernel()` behaviour. I think we can have it on by default.

sumanthkorikkar:
> I'm not a fan of the `bpftrace/tests: Rewrite codegen tests for probe read` commit. How does it know when to use user over the _kernel version of the helper?
> 
> With mockbpffeature we can control the `has_helper_probe_read_kernel()` behaviour. I think we can have it on by default.
- I agree. I didnt know about mockbpffeature :)
- But I suspect releases like bionic still dont have this helper. In this case, codegen tests would just fail. Hence, I had to somewhat manually add entries like probe_read|probe_read_user in **expected ir code**.  Let me know your thoughts. Thanks

fbs:
>     * But I suspect releases like bionic still dont have this helper. In this case, codegen tests would just fail. Hence, I had to somewhat manually add entries like probe_read|probe_read_user in **expected ir code**.  Let me know your thoughts. Thanks

Codegen tests aren't loaded into the kernel, so they shouldn't fail :).

sumanthkorikkar:
> > ```
> > * But I suspect releases like bionic still dont have this helper. In this case, codegen tests would just fail. Hence, I had to somewhat manually add entries like probe_read|probe_read_user in **expected ir code**.  Let me know your thoughts. Thanks
> > ```
> 
> Codegen tests aren't loaded into the kernel, so they shouldn't fail :).

Yes :) As, I was thinking about runtime select probe read helper... I will remove this commit and use mockbpffeature as suggested by you. Makes things much simpler then..

src/ast/irbuilderbpf.cpp
sumanthkorikkar: I tried the simple load here. But it didnt work for usdt_semphore_test - str(arg). However, reading reg using probe_read_kernel() did work.

mmisono: What errors do you have? Actually, in my environment (x64), the USDT arguments of usdt_semaphore_test are placed on a stack, so this if block are not executed.

```
% readelf --notes ./tests/testprogs/usdt_semaphore_test
[...]
Displaying notes found in: .note.stapsdt
  stapsdt              0x00000047       NT_STAPSDT (SystemTap probe descriptors)
    Provider: tracetest
    Name: testprobe
    Location: 0x00000000004005d1, Base: 0x0000000000400687, Semaphore: 0x0000000000601038
    Arguments: -8@-136(%rbp) 8@-144(%rbp)
```

But the following program executes the block,  and simple load works.

```c
#include <stdio.h>
#include <sys/sdt.h>
void f(int a){
  DTRACE_PROBE1(foo, bar, a);
}
int main(){
    f(3);
    return 0;
}
```

```
% readelf --notes ./a.out
[...]
Displaying notes found in: .note.stapsdt
  stapsdt              0x00000028       NT_STAPSDT (SystemTap probe descriptors)
    Provider: foo
    Name: bar
    Location: 0x0000000000000610, Base: 0x00000000000006a4, Semaphore: 0x0000000000000000
    Arguments: -4@%edi
```

```
% sudo ./src/bpftrace -e 'usdt:./a.out:foo:bar { printf("val: %d\n", arg0); }' -c ./a.out
Attaching 1 probe...
val: 3
```

```
Bytecode:
0: (bf) r6 = r1
1: (b7) r1 = 0
2: (7b) *(u64 *)(r10 -16) = r1
last_idx 2 first_idx 0
3: (71) r1 = *(u8 *)(r6 +112)
[...]
```

mmisono: ah,  load instruciton should be `result = CreateLoad(getInt64Ty(), reg);`

sumanthkorikkar: This works. Great. cant we just have signed IntCast here?

fbs: Eventually we need to move this to the semantic analysis so that we can get rid of all these casts. I'm doing that in my 1091_integer_size branch, so for now this should be ok (but use `CreateInt()` instead of `SizedType()`)

sumanthkorikkar: Ok. Will do.

sumanthkorikkar:
For codegen tests: As per my understanding, Setting  has_probe_read_kernel_ in mockbpffeature seems to be not enough.

- In common.h, test() -> we do use mockbpffeature feature and pass by ref to semantics. However, it does not set bpftrace.feature_.
- In irbuilderbpf.c -> we do use bpftrace.feature_.has_probe_read_kernel(). [ Which is not in sync with mockbpffeature ].

sumanthkorikkar:
Hi @fbs, 

 Copy is restricted in bpffeature for good sake,  This doesnt look so nice. But I had to make this for codegen.
- bpftrace.feature_ and MockBPFfeature are different instances and dont sync.
- First point wont be a problem for semantic analyser though, as we have a ref to MockBPFfeature.

If it is fine for now, I will add this. Thanks

```
--- a/src/bpffeature.h
+++ b/src/bpffeature.h
@@ -30,7 +30,7 @@ protected:                                                                     \
                                                                                \
 public:                                                                        \
   bool has_helper_##name(void)                                                 \
-  {                                                                            \
+  {                                                                            \
     if (!has_##name##_.has_value())                                            \
       has_##name##_ = std::make_optional<bool>(                                \
           detect_helper(libbpf::BPF_FUNC_##name, (progtype)));                 \
@@ -92,6 +92,11 @@ public:
   DEFINE_PROG_TEST(tracepoint, libbpf::BPF_PROG_TYPE_TRACEPOINT);
   DEFINE_PROG_TEST(perf_event, libbpf::BPF_PROG_TYPE_PERF_EVENT);
   DEFINE_PROG_TEST(kfunc, libbpf::BPF_PROG_TYPE_TRACING);
+  void manual_set_probe_read_kernel()
+  {
+    has_probe_read_kernel_ = std::optional<bool>(true);
+  }
+
 
 protected:
   std::optional<bool> has_loop_;
diff --git a/tests/codegen/common.h b/tests/codegen/common.h
index a003d84..9544bcf 100644
--- a/tests/codegen/common.h
+++ b/tests/codegen/common.h
@@ -377,6 +377,8 @@ static void test(BPFtrace &bpftrace,
   ASSERT_EQ(driver.parse_str(input), 0);
 
   MockBPFfeature feature;
+  // By default, we set the codegen tests to use probe split variant
+  bpftrace.feature_.manual_set_probe_read_kernel();
   ast::SemanticAnalyser semantics(driver.root_, bpftrace, feature);
   ASSERT_EQ(semantics.analyse(), 0);
   ASSERT_EQ(semantics.create_maps(true), 0);
```

fbs:
Ah yeah we shuffeled with that a bit. I think it makes more sense to let `class bpftrace` have a `unique_ptr` to `bpffeature` that way we can override it easily.

sumanthkorikkar:
> Ah yeah we shuffeled with that a bit. I think it makes more sense to let `class bpftrace` have a `unique_ptr` to `bpffeature` that way we can override it easily.

Ok. will try now.

sumanthkorikkar:
Hi @fbs, @mmisono,

Done with the changes.

- Use only one instance of bpffeature using unique_ptr. Needed for codegen tests
- Added suggestion by mmisono.

Thanks.

sumanthkorikkar:
Hi @fbs, @danobi, @mmisono,

Could please check if it looks good now. Looking forward to your comments and integration of this for our platform.  Thank you everyone

src/ast/semantic_analyser.cpp
```diff
@@ -1060,8 +1120,16 @@ void SemanticAnalyser::visit(Call &call)
   {
     if (!check_nargs(call, 1))
       return;
-    if (!check_arg(call, Type::pointer, 0))
+
+    // kptr should accept both integer or pointer. Consider case: kptr($1)
+    auto &arg = *call.vargs->at(0);
```
fbs, line 1125: What is the use case for this? 

Setting the addrspace should only be valid on pointers as pointers should be the only thing that we have to probe read. 

If we need `kptr($1)` to read data from a user specified  the resulting type should be `uint64 *` as per #1507, not just an int type.

fbs: Ah i guess `kptr(arg0)` is a valid use case, in that case we should do the implicit cast e.g. `kptr(arg0)` --> `kptr((uint64*) arg0)`

fbs:
Small comment but I think its good for now. This probably has a bug or two but its a good base to build on imo

src/ast/semantic_analyser.cpp
sumanthkorikkar: Ok. Cast looks better. I was thinking the following: If we know some kernel address and store that in a variable, then perform kptr($a). We would not need cast. But it it conflicts with the language, then we can surely change it.

sumanthkorikkar:
> Small comment but I think its good for now. This probably has a bug or two but its a good base to build on imo

Ok sure. Thank you.

src/ast/semantic_analyser.cpp
```diff
@@ -1396,6 +1468,29 @@ void SemanticAnalyser::visit(Binop &binop)
...
+  auto addr_lhs = binop.left->type.GetAS();
+  auto addr_rhs = binop.right->type.GetAS();
+
+  // if lhs or rhs has different addrspace (not none), then set the
+  // addrspace to none. This preserves the behaviour for x86.
+  if (addr_lhs != addr_rhs && addr_lhs != AddrSpace::none &&
+      addr_rhs != AddrSpace::none)
+  {
+    if (is_final_pass())
+      LOG(WARNING) << "Addrspace mismatch";
```
mmisono, line 1481:
```suggestion
      LOG(WARNING, binop.loc, out_) << "Addrspace mismatch";
```

mmisono (commented):
Functionally it looks ok, though I only tested this on x86_64. Regarding addrspace warnings, we still get a warning even if using uptr/kptr explicitly.

```
 % sudo ./src/bpftrace  -e 'uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*uptr(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }'
stdin:1:87-88: WARNING: Addrspace mismatch
uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*uptr(curtask->mm->mmap->vm_start + uaddr("ps1_prompt")))); }
                                                                                      ~
WARNING: Addrspace is not set
```

Edit: ah, in the above case, should we use `str(*(uptr(curtask->mm->mmap->vm_start) + uaddr("ps1_prompt"))`? It feels a bit weird though... In addition, still get a warning

```
% sudo ./src/bpftrace  -e 'uprobe:/bin/bash:readline { printf("PS1: %s\n", str(*(uptr(curtask->mm->mmap->vm_start) + uaddr("ps1_prompt")))); }'
WARNING: Addrspace is not set
```

src/ast/semantic_analyser.cpp
sumanthkorikkar: Done.

mmisono: I just remembered previously I discussed this topic: https://github.com/iovisor/bpftrace/pull/1141#discussion_r418320486. In short, pointer type would help to eliminate outer uptr() but it could break the compatibility... https://github.com/iovisor/bpftrace/issues/1507#issuecomment-696369158. I think the current behavior is OK for now.

sumanthkorikkar:
Kind ping.

fbs:
can you fix the conflict?

sumanthkorikkar:
Conflicts resolved - CHANGELOG.md.

fbs:
@mmisono @danobi   any last comments? I don't think we're doing a release any time soon so we will still have time to land fixes and improve the addrsspace situation.

fbs:
can you rebase onto master @sumanthkorikkar, i just merged #1544  to get rid of the test failure here

sumanthkorikkar:
Done.

sumanthkorikkar:
Minor change. Since recent commit commit be8fafc ("Positional params: handle numbers as strings") was introduced after this review.

sumanthkorikkar:
Done.

sumanthkorikkar:
Info: This PR fixes 26 runtime testcases on our platform. Also few tools.  This should also fix problems for ARM arch. Thanks.

mmisono (approved):

fbs:
\o/

Issue #1470, opened 2020-08-18
use batch map operators


fbs:
https://github.com/torvalds/linux/commit/990bca1fc8ad added support for batch operations. As something like `i:s:1 { print(@); clear(@); }` is quite common in tracing scripts this might be useful for us.  We can use `lookup_batch` and `delete_batch` to do the whole batch in one go.

Even better if we batch the two, i.e. we detect the `print(@);clear(@);` and merge it into `print_and_clear(@)` internally which does the `lookup_and_delete_batch`.

fbs:
[poc](https://github.com/fbs/bpftrace/tree/map_batch)

danobi:
Can't see any downsides. Seems like a pure win.

PR #1505, opened 2020-09-08
Map batch prep

fbs:
While implementing batch operators for #1470 I saw some things that could use a cleanup. This is that part, the map batch implementation will follow at some point.

This does 3 things:

- detect batch operator, both in libbpf and runtime
- deduplicate the fakemap/map creation code in the semanticanalyser
- move the map storage from `BPFtrace` into a std::map like container type.

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -2251,6 +2251,17 @@ int SemanticAnalyser::analyse()
 
 int SemanticAnalyser::create_maps(bool debug)
 {
+  // This exists to make sure both versions are defined
+  // when compiling this into a .o.
+  if (debug)
+    return create_maps_impl<bpftrace::FakeMap>();
+  else
+    return create_maps_impl<bpftrace::Map>();
+}
+
+template <typename T>
+int SemanticAnalyser::create_maps_impl(void)
+{
   uint32_t failed_maps = 0;
   auto is_invalid_map = [](int a) -> uint8_t { return a < 0 ? 1 : 0; };
   for (auto &map_val : map_val_)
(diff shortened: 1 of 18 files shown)


src/mapmanager.h
```diff
@@ -0,0 +1,97 @@
...
+  };
+  auto end()
+  {
+    return maps_by_id_.end();
+  };
+
+  /**
+     Internal maps
+  */
+  enum class Type
```
fbs, line 56: Not entirely sure about this part of it. Storing it as `elapsed_map_` etc like bpftrace did might be a better interface, with `std::optional<Map&> GetElapsed()` as getter?

src/map.cpp
```diff
@@ -137,4 +138,93 @@ Map::~Map()
...
+{
+  auto name = map->name_;
+  if (name.empty())
+    throw std::runtime_error("Map without name cannot be stored");
+
+  auto count = maps_by_name_.count(name);
+  if (count > 0)
+    throw std::runtime_error("Map with name: @" + name + " already exists");
+
+  std::shared_ptr<IMap> smap = std::move(map);
```
danobi, line 151: A little strange to use shared_ptrs here. Might be semantically more correct to store 1 unique ptr and the rest raw pointers, as MapManager already "owns" the map.

src/map.cpp
```diff
@@ -137,4 +138,93 @@ Map::~Map()
     close(mapfd_);
 }
 
+void MapManager::Add(std::unique_ptr<IMap> map)
+{
+  auto name = map->name_;
+  if (name.empty())
+    throw std::runtime_error("Map without name cannot be stored");
```
danobi, line 145: I think it'd be nicer to return a bool from this function instead of throwing an exception

src/map.cpp
```diff
@@ -137,4 +138,93 @@ Map::~Map()
...
+{
+  return maps_by_name_.find(name) != maps_by_name_.end();
+}
+
+IMap &MapManager::Lookup(const std::string &name)
+{
+  auto search = maps_by_name_.find(name);
+  if (search == maps_by_name_.end())
+  {
+    throw std::runtime_error("No map with name @" + name + " defined");
```
danobi, line 170: Might make more sense to return an `std::optional` instead of throwing exceptions so callers can decide how to handle errors. A missing map isn't really an exceptional situation. Same for other lookup functions below.

src/mapmanager.h
```diff
@@ -0,0 +1,97 @@
...
+namespace bpftrace {
+
+/**
+   A container for all maps created during program execution.
+
+*/
+class MapManager
+{
+public:
+  MapManager(){};
```
danobi, line 17:
```suggestion
  MapManager() = default;
```

src/map.cpp
```diff
@@ -137,4 +138,93 @@ Map::~Map()
...
+{
+  switch (t) {
+    case MapManager::Type::PerfEvent:
+      return "perf_event";
+    case MapManager::Type::Join:
+      return "join";
+    case MapManager::Type::Elapsed:
+      return "elapsed";
+    default:
+      throw std::runtime_error("BUG: Unknown MapManager::type");
```
danobi, line 227: Since b20320b0 is in now
```suggestion
    
    return {}; // unreached
```

src/mapmanager.h
danobi: No strong opinion here. Seems easy enough to modify if we change our minds later.

danobi (commented):
Nice cleanup. Biggest comment from me is that I don't like exceptions very much. I'd rather force callers to acknowledge that they're going to ignore an error at compile time (via std::optional or similar). It's too easy to not set up exception handling and get crashes after some changes some time later.

src/ast/semantic_analyser.cpp
```diff
@@ -2257,6 +2257,17 @@ int SemanticAnalyser::analyse()
 }
 
 int SemanticAnalyser::create_maps(bool debug)
+{
+  // This exists to make sure both versions are defined
+  // when compiling this into a .o.
```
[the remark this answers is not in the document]
fbs: No was just a comment for when someone wonders why we don't do the template in main()

src/map.cpp
fbs: Yeah sounds good

src/map.cpp
fbs: This is here to indicate a bug. All user maps should be named, for internal maps Set() should be used. I think that exceptions are the most appropriate thing we have atm for bugs as you want to keep the message somewhere.

src/ast/semantic_analyser.cpp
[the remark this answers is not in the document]
danobi: Ah I see now. Mind putting that in the comment? Wasn't sure what the comment meant and it might scare people off from modifying the code later.

src/map.cpp
danobi: Sure, makes sense

fbs:
Thanks! I think I've addressed them all.

danobi (approved):

PR #1506, opened 2020-09-09
Remove the --btf option

viktormalik:
Parse BTF by default if it is available. Actually, this was already done for all probe types, except tracepoints.

The only problem is that now the user cannot override types if BTF is available. This won't work:
```
bpftrace -e 'struct task_struct {int x;} i:ms:1 { printf("%d\n", curtask->x); }'
```
I don't see why someone would want to do that, but I may be missing something.

##### Checklist

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -592,8 +592,11 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     args.push_back(flag.c_str());
   }
 
-  bool process_btf = program->c_definitions.empty() ||
-                     (bpftrace.force_btf_ && bpftrace.btf_.has_data());
+  // Don't parse BTF for programs containing userspace probes if the user
+  // defined some custom data types, since those may conflict with BTF types.
+  bool process_btf = bpftrace.btf_.has_data() &&
+                     (!program->has_userspace_probes() ||
+                      program->c_definitions.empty());
 
   // We set these args early because some systems may not have <linux/types.h>
   // (containers) and fully rely on BTF.
@@ -661,10 +664,10 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
   {
     for (auto &msg : error_msgs)
     {
-      if (get_unknown_type(msg) != "" && !bpftrace.force_btf_)
+      if (get_unknown_type(msg))
       {
-        LOG(ERROR) << "Try running with --btf to force BTF processing or "
-                      "include headers with missing type definitions";
+        LOG(ERROR) << "Include headers with missing type definitions or "
+                      "install BTF information to your system";
       }
     }
     return false;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -53,7 +53,6 @@ void usage()
   std::cerr << "    -o file        redirect bpftrace output to file" << std::endl;
   std::cerr << "    -d             debug info dry run" << std::endl;
   std::cerr << "    -dd            verbose debug info dry run" << std::endl;
-  std::cerr << "    -b             force BTF (BPF type format) processing" << std::endl;
   std::cerr << "    -e 'program'   execute this program" << std::endl;
   std::cerr << "    -h, --help     show this help message" << std::endl;
   std::cerr << "    -I DIR         add the directory to the include search path" << std::endl;
@@ -259,7 +258,6 @@ int main(int argc, char *argv[])
   std::string cmd_str;
   bool listing = false;
   bool safe_mode = true;
-  bool force_btf = false;
   bool usdt_file_activation = false;
   int helper_check_level = 0;
   std::string script, search, file_name, output_file, output_format, output_elf;
@@ -350,7 +348,6 @@ int main(int argc, char *argv[])
         safe_mode = false;
         break;
       case 'b':
-        force_btf = true;
         break;
       case 'h':
         usage();
@@ -437,7 +434,6 @@ int main(int argc, char *argv[])
 
   bpftrace.usdt_file_activation_ = usdt_file_activation;
   bpftrace.safe_mode_ = safe_mode;
-  bpftrace.force_btf_ = force_btf;
   bpftrace.helper_check_level_ = helper_check_level;
   bpftrace.boottime_ = get_boottime();
 
--- a/src/types.h
+++ b/src/types.h
@@ -399,7 +399,7 @@ std::string addrspacestr(AddrSpace as);
 std::string typestr(Type t);
 std::string probetypeName(const std::string &type);
 std::string probetypeName(ProbeType t);
-bool is_userspace_probe(std::string &probe_name);
+bool is_userspace_probe(const std::string &probe_name);
 
 struct Probe
 {
--- a/src/types.h
+++ b/src/types.h
@@ -399,6 +399,7 @@ std::string addrspacestr(AddrSpace as);
 std::string typestr(Type t);
 std::string probetypeName(const std::string &type);
 std::string probetypeName(ProbeType t);
+bool is_userspace_probe(std::string &probe_name);
 
 struct Probe
 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ and this project adheres to
 #### Removed
 - Disable some kfunc probes whose tracing crashes
   - [#1432](https://github.com/iovisor/bpftrace/pull/1432)
+- Disable the --btf option
+  - [#1506](https://github.com/iovisor/bpftrace/pull/1506)
 
 #### Fixed
 - Fix negative overflow bug and unstable tests in PR #1416
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -138,7 +138,6 @@ OPTIONS:
     -B MODE        output buffering mode ('line', 'full', or 'none')
     -d             debug info dry run
     -dd            verbose debug info dry run
-    -b             force BTF (BPF type format) processing
     -e 'program'   execute this program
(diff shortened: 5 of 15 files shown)


mmisono:
I realized that this affects the user program that has a struct whose name is also in BTF. For example:

```c
#include <stdio.h>

struct path {
    int x;
};

void foo(struct path* path){
    printf("%d\n", path->x);
}

int main(){
    struct path p = { .x = 123 };
    foo(&p);
    return 0;
}
```
then
```bash
% sudo ./src/bpftrace -e 'struct path { int x; }; u:./a.out:foo { printf("%d\n", ((struct path*)arg0)->x); }' -c ./a.out
definitions.h:2:8: error: redefinition of 'path'
/bpftrace/include/__btf_generated_header.h:5:8: note: previous definition is here
ERROR: Include headers with missing type definitions or install BTF information to your system
```

though we can use a different name as a workaround (i.e. `'struct path2 { int x; }; u:./a.out:foo { printf("%d\n", ((struct path2*)arg0)->x); }' `). Considering BTF contains many structs, I'm not sure but there might be some chances of conflict.

So, how about processing BTF if probes are not uprobes? Something like

```cpp
bool process_btf =  (program does not contain uprobe/uretprobe/usdt probes)
                                 && bpftrace.btf_.has_data() ;
```

viktormalik:
> So, how about processing BTF if probes are not uprobes?

Seems like a good idea to me.

viktormalik:
In the end, I came with a less restrictive condition:
```cpp
bool process_btf = bpftrace.btf_.has_data() &&
                   !(program->has_userspace_probes() &&
                     !program->c_definitions.empty());
```
i.e., BTF is not processed if probes are uprobes and the user specified custom types.

viktormalik:
Tests need restart

src/clang_parser.cpp
```diff
@@ -663,8 +666,8 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     {
       if (get_unknown_type(msg) != "" && !bpftrace.force_btf_)
```
mmisono, line 667: I think we can remove `bpftrace.force_btf_`.

mmisono (approved):

src/clang_parser.cpp
viktormalik: Ah, I overlooked this one. Thanks

danobi:
Just an idea I had: what if we give user defined types precedence over BTF types? In other words, if we detect a definition mismatch we take the user definition.

That seems a little more predictable for users, keeps the rules simpler, and is backwards compatible.

src/main.cpp
```diff
@@ -349,9 +346,6 @@ int main(int argc, char *argv[])
       case 'u':
         safe_mode = false;
         break;
-      case 'b':
-        force_btf = true;
-        break;
```
danobi, line 354: Can we make this a noop instead? A noop with this PR is the same behavior as before right? And we can't reuse '--btf` anyways so we might as well leave it in to not break existing invocations people might have saved.

viktormalik: Yeah, that sounds reasonable, will do it.

viktormalik:
> Just an idea I had: what if we give user defined types precedence over BTF types? In other words, if we detect a definition mismatch we take the user definition.
> 
> That seems a little more predictable for users, keeps the rules simpler, and is backwards compatible.

I already had the same idea and I agree that it would be the nicest solution. The implementation would be a bit complicated, though, since we don't parse user data types separately. So at the time we extract types from BTF, we don't know which types the user defined. I can see two possible solutions:

1. Parse the user definitions (and included files) first, collect all the defined type names and then do not take those from BTF.

2. Try to detect a definition mismatch, but I haven't been able to find a libclang API that would allow to retrieve both conflicting type definitions (to check that one truly comes from BTF). We could parse the error message as we do [here](https://github.com/iovisor/bpftrace/blob/b20320b028ffa12ab27306f50f68cfe945755922/src/clang_parser.cpp#L677-L693) but that's not a nice approach.

I can implement one of the above methods (I'd prefer the first one), the question is whether it is worth it. It would certainly be useful for backwards compatibility, but I can see the only actual use-case for programs that combine usage of userspace and kernel probes and in the same time have some user data types.

danobi:
> I can implement one of the above methods (I'd prefer the first one), the question is whether it is worth it. It would certainly be useful for backwards compatibility, but I can see the only actual use-case for programs that combine usage of userspace and kernel probes and in the same time have some user data types.

Yeah I think you're right. I tried coming up with a counter example where someone might want to redefine a kernel type but they could just name the type something else. eg `struct my_task_struct { .. }`.

src/types.h
```diff
@@ -399,6 +399,7 @@ std::string addrspacestr(AddrSpace as);
 std::string typestr(Type t);
 std::string probetypeName(const std::string &type);
 std::string probetypeName(ProbeType t);
+bool is_userspace_probe(std::string &probe_name);
```
danobi, line 402: Nit: 
```suggestion
bool is_userspace_probe(const std::string &probe_name);
```

src/clang_parser.cpp
```diff
@@ -592,8 +592,11 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     args.push_back(flag.c_str());
   }
 
-  bool process_btf = program->c_definitions.empty() ||
-                     (bpftrace.force_btf_ && bpftrace.btf_.has_data());
+  // Don't parse BTF for programs containing userspace probes if the user
+  // defined some custom data types, since those may conflict with BTF types.
+  bool process_btf = bpftrace.btf_.has_data() &&
+                     !(program->has_userspace_probes() &&
+                       !program->c_definitions.empty());
```
danobi, line 599: Easier to read if we avoid double negative
```suggestion
  bool process_btf = bpftrace.btf_.has_data() &&
                     (!program->has_userspace_probes() ||
                       program->c_definitions.empty());
```

danobi (approved):

viktormalik:
Maybe we should address #1511 before closing this as it is related.

Including `sys/types.h` for userspace probes will cause redefinition errors since `linux/types.h` is included by default. Even if BTF was used instead of `linux/types.h`, there could still be conflicts with types parsed from BTF.

Implementing preference of user-defined types over those from BTF would partially resolve this problem, but it still wouldn't suffice for systems without BTF. Alternatively, we could stop including `linux/types.h` for programs with userspace probes or we could add a CLI option to disable including of `linux/types.h`.

danobi:
Maybe a new `--no-system-headers` flag that disables processing any system headers including those from BTF is the easiest way out. Then the user can be sure that nothing will conflict with any header files they include.

danobi:
I put up #1545 -- we can continue discussing there. I think we're probably good to merge this PR b/c it doesn't make #1511 any worse.

src/clang_parser.cpp
```diff
@@ -592,8 +592,11 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     args.push_back(flag.c_str());
   }
 
-  bool process_btf = program->c_definitions.empty() ||
-                     (bpftrace.force_btf_ && bpftrace.btf_.has_data());
+  // Don't parse BTF for programs containing userspace probes if the user
+  // defined some custom data types, since those may conflict with BTF types.
+  bool process_btf = bpftrace.btf_.has_data() &&
+                     (!program->has_userspace_probes() ||
+                      program->c_definitions.empty());
```
mmisono, line 599: I found that this could cause the redefinition errors. For example:

```
% sudo ./src/bpftrace ../tools/tcplife.bt
/lib/modules/5.9.0-rc6+/source/include/net/tcp_states.h:13:2: error: redefinition of enumerator 'TCP_ESTABLISHED'
/bpftrace/include/__btf_generated_header.h:9418:2: note: previous definition is here
/lib/modules/5.9.0-rc6+/source/include/net/tcp_states.h:14:2: error: redefinition of enumerator 'TCP_SYN_SENT'
/bpftrace/include/__btf_generated_header.h:9419:2: note: previous definition is here
[...]
```

We need to think of another condition to solve this.

danobi: Ouch, not very nice to break a tool we ship. I think we should revert this PR until we can make it work in all cases.

Continuing discussion from https://github.com/iovisor/bpftrace/pull/1506#issuecomment-690995775 , @viktormalik , do you think you could take a look at option 1?

viktormalik: Yeah, I'll look into that. It seems to be the cleanest solution and looks like it has use cases, after all.

viktormalik: So, after some experiments, I discovered a problem with this approach  - when asking BTF to dump a type definition, it also dumps definitions of all other types needed for successful compilation. If one of those types is overridden by the user, a re-definition error occurs (and it's quite difficult to discover which types were dumped).

This leads me to another thought - do we really want to allow user to do this? I'd say that allowing overriding a part of kernel types while taking the rest from BTF will always eventually blow up somewhere. I'd suggest enabling only the following scenarios:
- there are no user-defined types -> use BTF
- there are user-defined types but they do not override kernel types -> use BTF
- there are user-defined types conflicting with kernel types -> don't use BTF

This should allow user to define his own types that use existing kernel types without any problems, but if he wants to use types with conflicting names (e.g. for userspace tracing), he must define/include all the necessary types manually (or use them from `linux/types.h`, which is included in such case). This approach should be also backwards compatible for scripts that include kernel headers (such as the above `tcplife.bt`) as those must have already specified all the necessary types.

Any thoughs before I start implementing this?

fbs: I've moved the discussion to #1511

PR #1510, opened 2020-09-10
Turn off -Wswitch-default

danobi:
Requiring all `switch` statements have a `default` case is not that
sensible. It's better to let the developers omit the `default` case so
that the compiler can check that a `switch` on an `enum` is exhaustive.

This way, when new enum types are added, the developer can lean on the
compiler to catch all the unhandled cases instead of doing a manual
audit.

This commit removes the `-Wswitch-default` compiler flag and also
codemods the codebase to omit `default` cases where appropriate.

Note that the unreached return statements are necessary to avoid
warnings. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87951 for
in-depth discussion.

##### Checklist

- [n/a] Language changes are updated in `docs/reference_guide.md`
- [n/a] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [n/a] The new behaviour is covered by tests


danobi:
Now with this diff:
```diff
diff --git a/src/types.cpp b/src/types.cpp
index d1477d64..af2a2f2a 100644
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -222,7 +222,6 @@ std::string probetypeName(ProbeType t)
     case ProbeType::usdt:        return "usdt";        break;
     case ProbeType::tracepoint:  return "tracepoint";  break;
     case ProbeType::profile:     return "profile";     break;
-    case ProbeType::interval:    return "interval";    break;
     case ProbeType::software:    return "software";    break;
     case ProbeType::hardware:    return "hardware";    break;
     case ProbeType::watchpoint:  return "watchpoint";  break;
```

you get this warning (I had `-Werror` on for the build below):
```
/home/daniel/dev/bpftrace/src/types.cpp: In function ‘std::string bpftrace::probetypeName(bpftrace::ProbeType)’:
/home/daniel/dev/bpftrace/src/types.cpp:215:11: error: enumeration value ‘interval’ not handled in switch [-Werror=switch]
  215 |    switch (t)
```

PR #1514, opened 2020-09-14
Improve work with positional parameters

### LINKED ISSUE #1168: Inadequate semantic analysis of positional parameters
The following is an error (as intended; because `params.is_in_str` is true)

```sh
% sudo ./src/bpftrace -e 'BEGIN {printf("%s\n", str($1));}' 123
stdin:1:23-29: ERROR: $1 used in str(), but given numeric value: 123. Try $1 instead of str($1).
BEGIN {printf("%s\n", str($1));}
                      ~~~~~~
```

but the following is not (`params.is_in_str` is false)

```
% sudo ./src/bpftrace -e 'BEGIN {printf("%s\n", str($1+1));}' 123
Attaching 1 probe...
```

We need to propagate `is_in_str` information in some ways.

--------

In addition,  this is OK

```
% sudo ./src/bpftrace -e 'BEGIN {printf("%s\n", str($1));}' abc
Attaching 1 probe...
```

but the following is not

```
% sudo ./src/bpftrace -e 'BEGIN {printf("%s\n", str($1+1));}' abc
stdin:1:23-29: ERROR: 1.23-28$1 used numerically << but given "abc". Try using str($1).
BEGIN {printf("%s\n", str($1+1));}
                      ~~~~~~
```

This is the error reason for brendangregg/bpf-perf-tools-book#12.
If a positional parameter $1 is a string, $1 is already [in the stack](https://github.com/iovisor/bpftrace/blob/d0792c0619e0a024c1c443b978ae8c6db808f977/src/ast/codegen_llvm.cpp#L42-L46). If we don't support pointer arithmetic for $1, then we can treat a type of $1 as Type::string and eliminate `str()`.

### LINKED ISSUE #1218: Support using positional params as string literal
Positional params cannot be used as an argument of a function that takes string literals. For example:

```
% sudo ./src/bpftrace -e 'k:f { cgroupid($1); }' abc
stdin:1:7-19: ERROR: cgroupid() expects a string literal (integer provided)
k:f { cgroupid($1); }
      ~~~~~~~~~~~~
```
and
```
% sudo ./src/bpftrace -e 'k:f { cgroupid(str($1)); }' abc
stdin:1:7-24: ERROR: cgroupid() expects a string literal (string provided)
k:f { cgroupid(str($1)); }
      ~~~~~~~~~~~~~~~~~
```

Related: #1168,  positional params has some problems.

viktormalik:
This brings 3 improvements to positional parameters:
- improved semantic analysis that allows to use pointer arithmetic on string params (should help with brendangregg/bpf-perf-tools-book#12)
- handling numerical params as strings if specified in a `str()` call (fixes #1168)
- using positional params in calls that require string literals (fixes #1218)

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1862,18 +1862,26 @@ be used as a string in the `str()` call. If a parameter is used that was not pro
 zero for numeric context, and "" for string context. Positional parameters may also be used in probe
 argument and will be treated as a string parameter.
 
+If a positional parameter is used in `str()`, it is interpreted as a pointer to the actual given string 
+literal, which allows to do pointer arithmetic on it. Only addition of a single constant, less or equal to
+the length of the supplied string, is allowed.
+
 `$#` returns the number of positional arguments supplied.
 
 This allows scripts to be written that use basic arguments to change their behavior. If you develop a
 script that requires more complex argument processing, it may be better suited for bcc instead, which
 supports Python's argparse and completely custom argument processing.
 
-One-liner example:
+One-liner examples:
 
 ```
 # bpftrace -e 'BEGIN { printf("I got %d, %s (%d args)\n", $1, str($2), $#); }' 42 "hello"
 Attaching 1 probe...
 I got 42, hello (2 args)
+
+# bpftrace -e 'BEGIN { printf("%s\n", str($1 + 1)) }' "hello"
+Attaching 1 probe...
+ello
 ```
 
 Script example, bsize.d:
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -538,17 +538,21 @@ void SemanticAnalyser::visit(Call &call)
       call.type = CreateString(bpftrace_.strlen_);
       if (has_pos_param)
       {
-        auto binop = dynamic_cast<Binop *>(arg);
-        if (!(dynamic_cast<PositionalParameter *>(arg) ||
-              (binop && (dynamic_cast<PositionalParameter *>(binop->left) ||
-                         dynamic_cast<PositionalParameter *>(binop->right)))))
+        if (dynamic_cast<PositionalParameter *>(arg))
+          call.is_literal = true;
+        else
         {
-          // Only str($1), str($1 + CONST), or str(CONST + $1) are allowed
-          LOG(ERROR, call.loc, err_)
-              << call.func << "() only accepts positional parameters"
-              << " directly or with a single constant offset added";
+          auto binop = dynamic_cast<Binop *>(arg);
+          if (!(binop && (dynamic_cast<PositionalParameter *>(binop->left) ||
+                          dynamic_cast<PositionalParameter *>(binop->right))))
+          {
+            // Only str($1), str($1 + CONST), or str(CONST + $1) are allowed
+            LOG(ERROR, call.loc, err_)
+                << call.func << "() only accepts positional parameters"
+                << " directly or with a single constant offset added";
+          }
+          has_pos_param = false;
         }
-        has_pos_param = false;
       }
 
       if (is_final_pass() && call.vargs->size() > 1) {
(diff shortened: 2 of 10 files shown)


src/bpftrace.cpp
```diff
@@ -2124,4 +2124,26 @@ void BPFtrace::sort_by_key(std::vector<SizedType> key_args,
   }
 }
 
+std::string BPFtrace::get_string_literal(const ast::Expression *expr) const
```
fbs, line 2127: bit of a nit: I don't think this belongs in the bpfrace class. bpftrace already does way to much and this is very specific to the ast.  `ast/ast.h` might be a better place for it

viktormalik: I thought the same, but then I realized that it is specific to BPFtrace class since the class holds values of positional parameters that this function needs. I can move it to `ast/ast.h`, but then I'd have to add a reference to bpftrace as a new parameter (which I don't like too much as it breaks encapsulation).

fbs: Hmm fair point. Neither of the options are great. Passing "bpftrace" along feels a bit less wrong to me as its a global state we pass along everywhere anyway (although a name like context might be better). But it's easy to refactor this later on.

@danobi  any strong feelings?

mmisono:
This is a really nice fix to have. I think it is nice if we can limit the calculation in `str()` only for addition within a string length. For example, currently, the semantic analyzer allows the following expression (which resulting errors).

- `'BEGIN { printf("%s\n", str($1+5)); }' abc`
- `'BEGIN { printf("%s\n", str($1*10)); }' abc`
- `'BEGIN { printf("%s\n", str($1/10)); }' abc`

I also think we need to add documentation about this pointer arithmetic behavior of `str()`.

viktormalik:
> I think it is nice if we can limit the calculation in `str()` only for addition within a string length.
> I also think we need to add documentation about this pointer arithmetic behavior of `str()`.

Good points, I added a corresponding semantic check and a documentation entry. For now, these are specific to positional params within `str()` only, since general pointer arithmetic is not allowed yet (it is being discussed in #1505).

src/ast/semantic_analyser.cpp
```diff
@@ -1337,6 +1330,21 @@ void SemanticAnalyser::visit(Binop &binop)
               << "(cast to unsigned to silence warning)";
         }
       }
+
+      auto pos_param = dynamic_cast<PositionalParameter *>(left);
```
mmisono, line 1334: `right` can also be positional params (e.g., `str(1 + $1)`). Looks good to me other than that.

mmisono: I realized that we cannot write something like `str($1 + 2*4)`. I also found that we can bypass the check of addition length by using multiple add (e.g., `str($1 + 3 + 3)`.) But I guess it is probably hard to cope with these problems, so I think the current behavior is OK.

fbs: what is the usecase for this?

I don't think we should allow arbitrary complex expressions in this. A single literal integer as offset should be enough. If we ever land the option to reduce expressions (e.g. `1+2*3` into `7`) we can always add it back in.

mmisono: I agree, a literal integer offset should be enough.

viktormalik: I also agree, a single constant offset should be sufficient. I tried to make the semantic checks more robust though, so currently only `str($1)`, `str($1 + CONST)`, or `str(CONST + $1)` are supported, everything else should cause a semantic error.

mmisono (approved):

docs/reference_guide.md
```diff
@@ -1862,18 +1862,26 @@ be used as a string in the `str()` call. If a parameter is used that was not pro
...
 
-One-liner example:
+One-liner examples:
 
 ```
 # bpftrace -e 'BEGIN { printf("I got %d, %s (%d args)\n", $1, str($2), $#); }' 42 "hello"
 Attaching 1 probe...
 I got 42, hello (2 args)
+
+# bpftrace -e 'BEGIN { printf("%s\n", str($1 + 1)) }' "hello"
```
fbs, line 1882: can you add this as runtime test

fbs (approved):
Extra test would be good but otherwise LGTM

docs/reference_guide.md
viktormalik: Sure, I also added some tests for semantic analyzer, including tests with invalid expressions

danobi:
#1548  fixes a state leak issue this PR introduces

PR #1516, opened 2020-09-15
Plug leak in perf buffers

oazizi000:
Small change to remove a memory leak in the perf buffers. Otherwise, the missing `perf_reader_free()` triggers ASAN leak check.

##### Checklist


fbs:
Nice!

Issue #1047, opened 2019-12-31
Fix memory leaks in bpftrace


danobi:
If you run a trivial bpftrace script with ASAN on (see #1046), you'll see ~34 memory leaks: http://ix.io/2628 .

Build with asan by:
```
$ cd ~/dev/bpftrace
$ mkdir build && cd build
$ cmake .. -DBUILD_ASAN && make -j4
$ sudo ./src/bpftrace -e 'BEGIN { printf("hello\n"); }'
[...]
```

I'm sure we can find a lot more by running some of the scripts in `tools/` under asan mode.

rockdaboot:
BTW, cmake 3.15.4 on Debian unstable needs `cmake .. -DBUILD_ASAN=1`.

I struggled a bit since I'm normally not using cmake :-)

markdrayton:
Almost all of the leaks in the mentioned paste (and now, when I retry) are from parsing the program code. Presumably fixing them requires some Bison tinkering to `delete` all the stuff created with `new`. This looks tricky. The others:

```
$ sudo ./src/bpftrace -e 'BEGIN { printf("hello\n"); }' 2> /tmp/o
Attaching 1 probe...
hello
^C

$ awk 'BEGIN { RS = "\n\n"; ORS = RS } !/::parse/' /tmp/o


=================================================================
==36689==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1691 byte(s) in 329 object(s) allocated from:
    #0 0x7f034e8dd459 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f0348ff2e3c  (/usr/lib/libclang.so.10+0x3cde3c)

Direct leak of 576 byte(s) in 8 object(s) allocated from:
    #0 0x7f034e8dd639 in __interceptor_calloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x7f034dc90fa8 in perf_reader_new /home/mark/code/bcc/src/cc/perf_reader.c:57
    #2 0x7f034dc9075d in bpf_open_perf_buffer /home/mark/code/bcc/src/cc/libbpf.c:1216
    #3 0x55ba6aece22e in bpftrace::BPFtrace::setup_perf_events() /home/mark/code/bpftrace/src/bpftrace.cpp:1076
[...]

PR #1517, opened 2020-09-16
Free sym->module when resolving offset

markdrayton:
Free symbol module name to avoid leaking it (see #1047). Checklist doesn't seem appropriate for this fix.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -328,6 +328,11 @@ resolve_offset(const std::string &path, const std::string &symbol, uint64_t loc)
   if (bcc_resolve_symname(path.c_str(), symbol.c_str(), loc, 0, nullptr, &bcc_sym))
      throw std::runtime_error("Could not resolve symbol: " + path + ":" + symbol);
 
+  // Have to free sym.module, see:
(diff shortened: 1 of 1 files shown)


src/attached_probe.cpp
```diff
@@ -328,6 +328,10 @@ resolve_offset(const std::string &path, const std::string &symbol, uint64_t loc)
   if (bcc_resolve_symname(path.c_str(), symbol.c_str(), loc, 0, nullptr, &bcc_sym))
      throw std::runtime_error("Could not resolve symbol: " + path + ":" + symbol);
 
+  // https://github.com/iovisor/bcc/blob/master/src/cc/bcc_syms.h#L98-L99
```
fbs, line 331:
```suggestion
  // Have to free sym.module, see:
  // https://github.com/iovisor/bcc/blob/ba73657cb8c4dab83dfb89eed4a8b3866255569a/src/cc/bcc_syms.h#L98-L99
```

Links to master tend to break when someone updates the file, you can select a permalink from the "menu" after selecting line numbers

markdrayton: Ah, fair point. Will update. Thanks!

fbs:
Looks good!. Could you squash the commits for me? Don't think i can squash and rebase with github.

markdrayton:
Squashed!

PR #1518, opened 2020-09-16
Add LIBBPF_LIBRARIES to bpf_attach_kprobe build

olsajiri:
When using libbcc-no-libbpf.so as bcc library,
we need to link the test also with libbpf.so,
otherwise the test compilation fails on missing
libbpf symbols.

/usr/bin/c++   -DLIBBCC_ATTACH_KPROBE_SIX_ARGS_SIGNATURE  \
 CMakeFiles/cmTC_ddd8d.dir/src.cxx.o  -o cmTC_ddd8d       \
 -Wl,-rpath,/opt/bcc/lib64 /opt/bcc/lib64/libbcc-no-libbpf.so
/usr/bin/ld: /opt/bcc/lib64/libbcc-no-libbpf.so: undefined reference to `ring_buffer__poll@LIBBPF_0.0.9'
/usr/bin/ld: /opt/bcc/lib64/libbcc-no-libbpf.so: undefined reference to `ring_buffer__new@LIBBPF_0.0.9'
/usr/bin/ld: /opt/bcc/lib64/libbcc-no-libbpf.so: undefined reference to `ring_buffer__add@LIBBPF_0.0.9'
/usr/bin/ld: /opt/bcc/lib64/libbcc-no-libbpf.so: undefined reference to `ring_buffer__consume@LIBBPF_0.0.9'
/usr/bin/ld: /opt/bcc/lib64/libbcc-no-libbpf.so: undefined reference to `ring_buffer__free@LIBBPF_0.0.9'


fbs:
weird ci failure, ill rerun it

```
 Post Cache docker image
2m 13s
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5)
Post job cleanup.
/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/bpftrace/bpftrace --files-from manifest.txt
[warning]uploadChunk (start: 67108864, end: 100663295) failed: Cache service responded with 503
/home/runner/work/_actions/actions/cache/v2/dist/save/index.js:3136
                        throw new Error(`Cache upload failed because file read failed with ${error.message}`);
                        ^

Error: Cache upload failed because file read failed with EBADF: bad file descriptor, read
    at ReadStream.<anonymous> (/home/runner/work/_actions/actions/cache/v2/dist/save/index.js:3136:31)
    at ReadStream.emit (events.js:210:5)
    at internal/fs/streams.js:167:12
    at FSReqCallback.wrapper [as oncomplete] (fs.js:470:5)
```

fbs:
I wonder if it makes sense to start making libbpf a requirement now that it is available as separate tree: https://github.com/libbpf/libbpf. If we vendor (git submodule) the libbpf version we need we can clean up some of the library mess and  I think it will clean up some of our library mess and libbpf related ifdefs:

HAVE_LIBBPF
HAVE_LIBBPF_BTF_DUMP
HAVE_LIBBPF_BTF_DUMP_EMIT_TYPE_DECL
HAVE_LIBBPF_MAP_BATCH

olsajiri:
> I wonder if it makes sense to start making libbpf a requirement now that it is available as separate tree: https://github.com/libbpf/libbpf. If we vendor (git submodule) the libbpf version we need we can clean up some of the library mess and I think it will clean up some of our library mess and libbpf related ifdefs:

do you mean add libbpf as submodule and compile it in? I'd be strongly against ;-)

it's already added in bcc like that and we ended up with multiple libbpf loaded
in process, causing issues during dynamic linking and more pain if those 2 libbpf
libraries were on different versions

that's why in fedora we link bpftrace with libbcc-no-libbpf.so and libbpf.so to be clear
and have just one libbpf.so loaded

fbs:
> that's why in fedora we link bpftrace with libbcc-no-libbpf.so and libbpf.so to be clear
> and have just one libbpf.so loaded

Yes we would have to be sure that everyone does that.  I guess it probably causes more issues than it solves :)

PR #1055, opened 2020-01-01
Fix memory leaks

danobi:
This PR fixes some memory leaks in bpftrace. I've only run the asan
build under some very basic programs so this doesn't prove that bpftrace
is totally leak free or anything.

This PR is also rebased on top of #1050. I'll rebase this PR onto master 
once #1050 merges.

This fixes some of #1047.

PR #1519, opened 2020-09-16
Fix leaks from Bison memory allocations

oazizi000:
The Bison code creates a lot of objects with the `new` operator. These all end up causing memory leaks, because there is no corresponding call to `delete`. This change resolves this issue by using `std::unique_ptr` to create managed objects that will be destroyed once they are no longer needed.

This is a large commit, but is mostly mechanical changes, no logical change is intended.

##### Checklist

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -118,13 +118,15 @@ Call::Call(const std::string &func, location loc)
 {
 }
 
-Call::Call(const std::string &func, ExpressionList *vargs)
-    : func(is_deprecated(func)), vargs(vargs)
+Call::Call(const std::string &func, std::unique_ptr<ExpressionList> vargs)
+    : func(is_deprecated(func)), vargs(std::move(vargs))
 {
 }
 
-Call::Call(const std::string &func, ExpressionList *vargs, location loc)
-    : Expression(loc), func(is_deprecated(func)), vargs(vargs)
+Call::Call(const std::string &func,
+           std::unique_ptr<ExpressionList> vargs,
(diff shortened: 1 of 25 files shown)


fbs:
Thanks for looking into this? Have you seen the discussion in #1055 ?

oazizi000:
> Thanks for looking into this! Have you seen the discussion in #1055 ?

@fbs I had not seen #1055, so thanks for pointing that out! Looks like I've gone down a very similar road to what was discussed in that thread, so happy to see that (provides some confirmation that it's the right approach).

This particular PR still has some issues that I have not been able to reproduce locally, so it's not passing the checks. Once I get things figured out, I can push out an update community to review and discuss, as a potential fix. If there's a pending fix from #1055, I'm obviously happy to go with that as well.

fbs:
Looks like the build issue is exactly the thing discussed in #1055, a bison doesn't move http://www.comrite.com/wp/c11-stdunique_ptr-as-semantic-value-in-bison-c-mode/ . 

It looks like we'd loose ubuntu 18.04 (lts) if we do this as it still has 3.0.x and no backports, so this might have to wait a few more years.

I think the delete approach from 1055 is better for now. Its a bit ugly but easy enough to implement.

fbs:
The  manual memory management is in  so I think thats the way to go for now. Moving to smart pointers would be nice but bumping bison versions means dropping ubuntu which we can't do just yet.

Maybe we can have AST constructors dedicated to the parser  that convert the raw pointers into smart pointers untill we can update bison.

PR #1520, opened 2020-09-16
Align perf event data to avoid UBSAN errors

oazizi000:
This change copies the data being passed from the perf buffers back into the user space into 8-byte aligned memory. Note that the data from the perf buffers is not generally aligned because of the perf event header (more details below).

Without this change, an undefined behavior sanitizer complains about alignment issues:

```
third_party/bpftrace/src/bpftrace.cpp:723:17: runtime error: reference binding to misaligned address 0x7f5bbc95d014 for type 'uint64_t' (aka 'unsigned long'), which requires 8 byte alignment
0x7f5bbc95d014: note: pointer points here
  00 00 00 00 5f bf 78 b4  a6 df 00 00 00 00 00 00  09 00 00 00 01 00 20 00  14 00 00 00 00 00 00 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior third_party/bpftrace/src/bpftrace.cpp:723:17 in 
/usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/unique_ptr.h:962:58: runtime error: reference binding to misaligned address 0x7f5bbc95d014 for type 'unsigned long', which requires 8 byte alignment
0x7f5bbc95d014: note: pointer points here
  00 00 00 00 5f bf 78 b4  a6 df 00 00 00 00 00 00  09 00 00 00 01 00 20 00  14 00 00 00 00 00 00 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/unique_ptr.h:962:58 in 
/usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/move.h:77:14: runtime error: reference binding to misaligned address 0x7f5bbc95d014 for type 'unsigned long', which requires 8 byte alignment
0x7f5bbc95d014: note: pointer points here
  00 00 00 00 5f bf 78 b4  a6 df 00 00 00 00 00 00  09 00 00 00 01 00 20 00  14 00 00 00 00 00 00 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/move.h:77:14 in 
/usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/unique_ptr.h:962:38: runtime error: load of misaligned address 0x7f5bbc95d014 for type 'unsigned long', which requires 8 byte alignment
0x7f5bbc95d014: note: pointer points here
  00 00 00 00 5f bf 78 b4  a6 df 00 00 00 00 00 00  09 00 00 00 01 00 20 00  14 00 00 00 00 00 00 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/unique_ptr.h:962:38 in
```

Extra details for the curious:

 perf_submit() uses PERF_RECORD_SAMPLE with PERF_SAMPLE_RAW, which has the following structure.
```
      struct {
        struct perf_event_header {
          __u32   type;
          __u16   misc;
          __u16   size;
        } header;
        u32    size;        /* if PERF_SAMPLE_RAW */
        char  data[size];   /* if PERF_SAMPLE_RAW */
      };
```
 The entire struct as a whole is 12-bytes + data[size].
 This means the data member is 4-bytes aligned.

 If data is reinterpreted as any struct with 8-byte members,
 access to the struct's members can cause misalignment errors.
 For this reason, the data needs to be copied to an 8-byte boundary before being used.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -537,11 +537,21 @@ void BPFtrace::request_finalize()
     child_->terminate();
 }
 
-void perf_event_printer(void *cb_cookie, void *data, int size __attribute__((unused)))
+void perf_event_printer(void *cb_cookie, void *data, int size)
 {
+  // The perf event data is not aligned, so we use memcpy to copy the data and
+  // avoid UBSAN errors. Using an std::vector guarantees that it will be aligned
+  // to the largest type. See:
+  // https://stackoverflow.com/questions/8456236/how-is-a-vectors-data-aligned.
+  std::vector<uint8_t> data_aligned;
+  data_aligned.resize(size);
+  memcpy(data_aligned.data(), data, size);
+
   auto bpftrace = static_cast<BPFtrace*>(cb_cookie);
-  auto printf_id = *static_cast<uint64_t*>(data);
-  auto arg_data = static_cast<uint8_t*>(data);
+  auto arg_data = data_aligned.data();
+
+  auto printf_id = *reinterpret_cast<uint64_t *>(arg_data);
+
   int err;
 
   // Ignore the remaining events if perf_event_printer is called during finalization


fbs:
Thanks! Can you put all this info into the commit itself? Our policy is to rebase so it would get 'lost' otherwise.

oazizi000:
> Thanks! Can you put all this info into the commit itself? Our policy is to rebase so it would get 'lost' otherwise.

No problem. Done :)

PR #1525, opened 2020-09-20
Push tags to quay as well as master

### LINKED ISSUE #1522: Container image for 0.11.0
### What reproduces the bug?
Was looking into updating kubectl-trace to the latest bpftrace 0.11.0 however I can't find the `quay.io/iovisor/bpftrace:v0.11.0` image that we would need to use as base image.

I see that there's automation setup here to publish the images and in fact we do have the images for `v0.9.4` and `v.0.10.0` for example however the `v0.11.0` one is missing.

Anyone has idea why? I'm here to help if there's anything I can do to get the image published.

dalehamel:
WIP to fix #1522 from coming up with future releases

This is difficult to test because it needs to be run on the iovisor repo to be sure.

The syntax here should work, using the logical grouping operator `()` and logical OR, `||`, and the "startsWith") builtin: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#startswith

### DIFF
--- a/.github/workflows/embedded.yml
+++ b/.github/workflows/embedded.yml
@@ -230,7 +230,7 @@ jobs:
     - name: Authenticate with docker registry
       if: >
         matrix.env['TYPE'] == 'Release' &&
-        github.ref == 'refs/heads/master' &&
+        (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/')) &&
         github.repository == 'iovisor/bpftrace'
       env:
         QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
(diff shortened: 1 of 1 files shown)


fbs:
Nice, you've beat me to it :)

dalehamel:
@fbs can you pull this, then push it as a dummy tag to the iovisor repo (eg, quay-tag-test)? This should trigger a build that will show this works.

I can also try to test this myself by reverting the org/repo filter, then reverting that revert.

fbs:
I've added this as tag,  build here: https://github.com/iovisor/bpftrace/actions/runs/264151000

dalehamel:
It looks like the test for printing the version fails on tags that don't match a version structure

dalehamel:
Looks like it just has to start with v:

https://github.com/iovisor/bpftrace/blob/master/tests/runtime/basic#L3

dalehamel:
https://github.com/dalehamel/bpftrace/actions/runs/264163688 if this succeeds, we can try appending a "v" to the tagname to test on the iovisor repo (eg https://github.com/iovisor/bpftrace/tree/push-tags-quay can be "vpush-tags-quay")

dalehamel:
Looks like it works, can you push a tag that starts with "v" to verify, then I think this is OK to merge @fbs

fbs:
https://github.com/iovisor/bpftrace/actions/runs/264891900

dalehamel:
@fbs looks good I think!

PR #1526, opened 2020-09-22
release v0.11.1

fbs:
Our quay.io builds had two issues:

- END trigger was broken due to binary stripping
- v0.11.0 was never pushed due to a CI issue

To fix that we need a v0.11.1 release with fixes for both. This means that we won't get all the v0.X images we usually have but I don't think anyone uses them. 

See #1522 for details.

The 0.11_release branch is based on the current v0.11.0 tag.  This PR backports the fixes required for the quay.io image into that branch and prepares for the 0.11.1 release.

After this has been merged we need still need to:

- go through the release process, create a v0.11.1 tag etc
- merge the `v0.11.1 release` commit into master, to keep the changelog in sync

### DIFF
--- a/.github/workflows/embedded.yml
+++ b/.github/workflows/embedded.yml
@@ -207,7 +207,7 @@ jobs:
     - name: Authenticate with docker registry
       if: >
         matrix.env['TYPE'] == 'Release' &&
-        github.ref == 'refs/heads/master' &&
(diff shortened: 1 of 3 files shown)


fbs:
@dalehamel

danobi (approved):
clang-format seems confused. Can probably ignore that

dalehamel (approved):
Yeah looks good, same comment that CI seems confused because of how the clang format check works.

PR #1530, opened 2020-09-23
src/attached_probe: fix wrong setting of vmlinux_location.raw

lostjeffle:
find_vmlinux() will skip checking if vmlinux_location.raw is true, and
thus the wrong setting of vmlinux_location.raw in
resolve_offset_kprobe() will always cause the following error when
BPFTRACE_VMLINUX env exists:

"ERROR: Could not resolve symbol XXXXX. Use BPFTRACE_VMLINUX
env variable to specify vmlinux path. Compile bpftrace with
ALLOW_UNSAFE_PROBE option to force skip the check."

Signed-off-by: Jeffle Xu <jefflexu@linux.alibaba.com>

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ and this project adheres to
   - [#1457](https://github.com/iovisor/bpftrace/pull/1457)
 - Fix type resolution for struct field access via variables
   - [#1450](https://github.com/iovisor/bpftrace/pull/1450)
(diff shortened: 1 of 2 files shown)


mmisono (approved):
I've missed it. Thanks for the fix!

fbs:
Can you add a changelog entry? :)

lostjeffle:
CHANGELOG.md updated.

PR #1534, opened 2020-09-23
Support old C++ compilers that don't have <filesystem>

kzys:
Old C++ compilers such as g++ 7.x doesn't have `<filesystem>`, but they sometimes
have `<experimental/filesystem>` instead.

##### Checklist

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -3,7 +3,6 @@
 #include <cmath>
 #include <cstring>
 #include <fcntl.h>
-#include <filesystem>
 #include <fstream>
 #include <glob.h>
 #include <link.h>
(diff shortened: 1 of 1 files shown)


fbs:
Thanks looking into this!

Looks like the alpine build fails on this, guess the `__cpp_lib_filesystem` isn't set there?

kzys:
The alpine issue is odd. Since it was working before my change, it should have `std::filesystem` but `__cpp_lib_filesystem` seems not defined. Let me dig in.

kzys:
`__cpp_lib_filesystem` is defined in `<filesystem>`. I think I had to include `<version>` first to have the feature macro.

https://en.cppreference.com/w/cpp/header/version

kzys:
But <version> is not available prior to C++20... So I've changed the include section to use has_include which is available > C++17.

https://en.cppreference.com/w/cpp/preprocessor/include

fbs:
thanks!

PR #1538, opened 2020-09-26
Do not check string size for printf()'s format string

yzhao1012:
And create String with its original size, without truncation.

This allows having longer format string in printf() statement,
and does not affect existing functionality.

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -76,6 +76,14 @@ void SemanticAnalyser::visit(PositionalParameter &param)
 
 void SemanticAnalyser::visit(String &string)
 {
+  // Skip check for printf()'s format string (1st argument) and create the string
+  // with the original size. This is because format string is not part of bpf byte code.
+  if (func_ == "printf" && func_arg_idx_ == 0)
+  {
+    string.type = CreateString(string.str.size());
+    return;
+  }
+
   if (!is_compile_time_func(func_) && string.str.size() > STRING_SIZE - 1)
   {
     LOG(ERROR, string.loc, err_) << "String is too long (over " << STRING_SIZE
@@ -388,6 +396,7 @@ void SemanticAnalyser::visit(Call &call)
     ~func_setter()
     {
       analyser_.func_ = old_func_;
+      analyser_.func_arg_idx_ = -1;
     }
 
   private:
@@ -401,7 +410,11 @@ void SemanticAnalyser::visit(Call &call)
     is_in_str = true;
 
   if (call.vargs) {
-    for (Expression *expr : *call.vargs) {
+    for (size_t i = 0; i < call.vargs->size(); ++i)
+    {
+      auto &expr = (*call.vargs)[i];
+      func_arg_idx_ = i;
+
       expr->accept(*this);
     }
   }
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -76,8 +76,9 @@ void SemanticAnalyser::visit(PositionalParameter &param)
 
 void SemanticAnalyser::visit(String &string)
 {
-  // Skip check for printf()'s format string (1st argument) and create the string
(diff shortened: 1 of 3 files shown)


yzhao1012:
Not a language change.

Need help with where to add the tests.

src/ast/semantic_analyser.cpp
```diff
@@ -76,6 +76,14 @@ void SemanticAnalyser::visit(PositionalParameter &param)
 
 void SemanticAnalyser::visit(String &string)
 {
+  // Skip check for printf()'s format string (1st argument) and create the string
+  // with the original size. This is because format string is not part of bpf byte code.
+  if (func_ == "printf" && func_arg_idx_ == 0)
```
fbs, line 81: Bit of a shame we cannot use `is_compiel_time_func` here. I've created #1539 to make that possible.

fbs:
Thanks!

> Need help with where to add the tests.

You generally want to add a test for the pass you modified. So in this case in tests/semantic_analyser.cpp you can extend `﻿﻿TEST(semantic_analyser, printf)` with a case for this. And please add a runtime test too

src/ast/semantic_analyser.h
```diff
@@ -100,7 +100,13 @@ class SemanticAnalyser : public Visitor {
   void accept_statements(StatementList *stmts);
 
   Probe *probe_;
+
+  // Temporarily record the function currently being visited by this SemanticAnalyser.
```
danobi, line 104: Wording nit. Same for below
```suggestion
  // Holds the function currently being visited
```

yzhao1012:
> Thanks!
> 
> > Need help with where to add the tests.
> 
> You generally want to add a test for the pass you modified. So in this case in tests/semantic_analyser.cpp you can extend `﻿﻿TEST(semantic_analyser, printf)` with a case for this. And please add a runtime test too.
> 
> Could you also run [clang format](https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#using-clang-format) and squash all the commits together?

Fixed clang-format issue.

Working on test now.

src/ast/semantic_analyser.h
yzhao1012: Done.

fbs:
@yzhao1012, can I help you with the tests? This would be good to get this in :)

yzhao1012:
> @yzhao1012, can I help you with the tests? This would be good to get this in :)

It would be great if you can shoulder the task of writing tests.

I am not sure what form of collaboration is the best here. It's totally OK to me if you can just copy or rewrite this PR in a new one that you create from scratch.

Or you were thinking to give specific instructions on writing tests?

fbs:
I was thinking simple something like this

```
TEST(semantic_analyser, call_printf)
{
  std::stringstream prog;

  prog << "i:ms:100 { printf(\"" << std::string("a", 200) << " %d\\n\", 1); }"
  test(prog, 0);
}
```

yzhao1012:
> > Thanks!
> > > Need help with where to add the tests.
> > 
> > 
> > You generally want to add a test for the pass you modified. So in this case in tests/semantic_analyser.cpp you can extend `﻿﻿TEST(semantic_analyser, printf)` with a case for this. And please add a runtime test too.
> > Could you also run [clang format](https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#using-clang-format) and squash all the commits together?
> 
> Fixed clang-format issue.
> 
> Working on test now.

How could I run `semantic_analyser.cpp`?

I was reading https://github.com/iovisor/bpftrace/tree/master/tests but could not find the exact instructions on how to run an individual test.

fbs:
if you want an individual test you can do `./tests/bpftrace_test --gtest_filter="semantic_analyser.call_printf"` but you can just run the whole testsuite.

fbs:
@yzhao1012 if you don't mind I'll steal this, rebase and add the tests and then get it merged :). Its nice to have

PR #770, opened 2019-06-14
[WIP] tracepoint __data_loc support

rjenkins:
N.B, Questions below, sharing for early feedback.

Per Issue https://github.com/iovisor/bpftrace/issues/385 accessing `__data_loc` fields via `args->` should not result in the following error `Struct/union of type '_tracepoint_irq_irq_handler_entry' does not contain a field named 'name'`

I've started working on this, with the initial approach of detecting the `__data_loc` field in `clang_parser.cpp`, marking `is_data_loc=true` in `SizedType` and dual registering it in the fields with the prefix stripped. Additionally we extract the type in `tracepoint_format_parser.cpp` and convert `char[]` to `const char *`. This results in the following.

```
rjenkins@lappy:~/bpftrace/build$ sudo ./src/bpftrace -e 't:irq:irq_handler_entry { @[args->name] = count(); }'
Attaching 1 probe...
^C

@[15605576393127534]: 1
@[14479676486284910]: 2
@[15324101416416878]: 3
@[14198201509574254]: 3
@[8097308509766972513]: 12
@[7233740185486452856]: 13
@[892418409]: 58
```

However as noted by @brendangregg we still require special handling for this field as usage with the `str` built in results in null values at `@[]`

```
rjenkins@lappy:~/bpftrace/build$ sudo ./src/bpftrace -e 't:irq:irq_handler_entry { @[str(args->name)] = count(); }'
Attaching 1 probe...
^C

@[]: 121
```

Questions: 
Is it OK to return `const char *` here?
Is dual registering the field with and without `data_loc_` an acceptable approach or should we prefer something else?

I plan on working on digging into how these calls eventually get passed into `irbuilderbpf.cpp` and see if we can perform `bpf_probe_read` with the proper offset data as Sasha did for bcc https://github.com/iovisor/bpftrace/issues/385#issuecomment-495006035.

Issue #1396, opened 2020-06-24
Improve type::record (struct)


fbs:
#1394 introduces a simple `record` type that is either a struct or a union. This type currently has a name and a size. For info about its fields it falls back to bpftrace.structs_ which uses the Struct and Field types. Would be nice if we can combine the whole thing and have type::record contain everything.

bpftrace.structs_ can then hold the record type directly or maybe we can get rid of it completely and solve that some other way.

PR #1542, opened 2020-09-29
tracepoint: Support __data_loc fields in tracepoints

### LINKED ISSUE #385: tracepoint arguments are missing __data_loc char strings
This one-liner shouldn't fail like this:

```
# bpftrace -ve 't:irq:irq_handler_entry { @[args->name] = count(); }'
Struct/union of type '_tracepoint_irq_irq_handler_entry' does not contain a field named 'name'

# cat /sys/kernel/debug/tracing/events/irq/irq_handler_entry/format 
name: irq_handler_entry
ID: 149
format:
	field:unsigned short common_type;	offset:0;	size:2;	signed:0;
	field:unsigned char common_flags;	offset:2;	size:1;	signed:0;
	field:unsigned char common_preempt_count;	offset:3;	size:1;	signed:0;
	field:int common_pid;	offset:4;	size:4;	signed:1;

	field:int irq;	offset:8;	size:4;	signed:1;
	field:__data_loc char[] name;	offset:12;	size:4;	signed:1;

print fmt: "irq=%d name=%s", REC->irq, __get_str(name)
```

I don't think the tracepoint args parser is handling the "__data_loc char[] name" entry.

danobi:
This PR teaches bpftrace to read `__data_loc` arguments from tracepoints.

Note that this PR causes a minor compatibility break where fields previously
named `args->data_loc_name` are now named `args->name`.

This closes #385 . This supercedes #770 .

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1539,6 +1539,27 @@ void CodegenLLVM::visit(FieldAccess &acc)
       Value *masked = b_.CreateAnd(shifted, field.bitfield.mask);
       expr_ = masked;
     }
+    else if (field.type.IsIntTy() && field.is_data_loc)
+    {
+      // `is_data_loc` should only be set if field access is on `args` which
+      // has to be a ctx access
+      assert(type.IsCtxAccess());
+      assert(ctx_->getType() == b_.getInt8PtrTy());
+      // Parser needs to have rewritten field to be a u64
+      assert(field.type.IsIntTy());
+      assert(field.type.GetIntBitWidth() == 64);
+
+      // Top 2 bytes are length (which we'll ignore). Bottom two bytes are
+      // offset which we add to the start of the tracepoint struct.
+      expr_ = b_.CreateLoad(
+          b_.getInt32Ty(),
+          b_.CreateGEP(b_.CreatePointerCast(ctx_,
+                                            b_.getInt32Ty()->getPointerTo()),
+                       b_.getInt64(field.offset / 4)));
+      expr_ = b_.CreateIntCast(expr_, b_.getInt64Ty(), false);
+      expr_ = b_.CreateAnd(expr_, b_.getInt64(0xFFFF));
+      expr_ = b_.CreateAdd(expr_, b_.CreatePtrToInt(ctx_, b_.getInt64Ty()));
+    }
     else if ((field.type.IsIntTy() || field.type.IsPtrTy()) &&
              type.IsCtxAccess())
     {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to
   - [#1499](https://github.com/iovisor/bpftrace/pull/1499)
 - Positional parameters: support numbers as strings and params as string literals
   - [#1514](https://github.com/iovisor/bpftrace/pull/1514)
+- Support for tracepoint __data_loc fields
+  - [#1542](https://github.com/iovisor/bpftrace/pull/1542)
 
 #### Changed
 - Warn if using `print` on `stats` maps with top and div arguments
@@ -44,6 +46,8 @@ and this project adheres to
   - [#1486](https://github.com/iovisor/bpftrace/pull/1486)
 - Switch `nsecs` to `ktime_get_boot_ns`
   - [#1475](https://github.com/iovisor/bpftrace/pull/1475)
+- Tracepoint __data_loc fields are renamed from `args->data_loc_name` to `args->name`
+  - [#1542](https://github.com/iovisor/bpftrace/pull/1542)
 
 #### Deprecated
 
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -391,6 +391,42 @@ CXCursor ClangParser::ClangParserHandler::get_translation_unit_cursor() {
   return clang_getTranslationUnitCursor(translation_unit);
 }
 
+namespace {
+// Get annotation associated with field declaration `c`
+std::optional<std::string> get_field_decl_annotation(CXCursor c)
+{
+  assert(clang_getCursorKind(c) == CXCursor_FieldDecl);
+
+  std::optional<std::string> annotation;
+  clang_visitChildren(c,
+                      [](CXCursor c,
+                         CXCursor __attribute__((unused)) parent,
+                         CXClientData data) {
+                        // The header generation code can annotate some struct
+                        // fields with additional information for us to parse
+                        // here. The annotation looks like:
+                        //
+                        //    struct Foo {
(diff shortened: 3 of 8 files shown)


danobi:
There's still one thing I'm really confused about. So the typical case works ok:

```
$ sudo ./build/src/bpftrace -e 't:irq:irq_handler_entry { print(str(args->name)) }' | head -10
Attaching 1 probe...
nvidia
nvidia
nvidia
nvidia
nvidia
nvidia
nvidia
nvidia
```

But printing the address fails:
```
$ sudo ./build/src/bpftrace -e 't:irq:irq_handler_entry { print(args->name) }' -v
Attaching 1 probe...

Error log:
0: (bf) r6 = r1
1: (61) r1 = *(u32 *)(r6 +12)
2: (b7) r2 = 30007
3: (7b) *(u64 *)(r10 -24) = r2
4: (57) r1 &= 65535
5: (0f) r1 += r6
last_idx 5 first_idx 0
regs=2 stack=0 before 4: (57) r1 &= 65535
regs=2 stack=0 before 3: (7b) *(u64 *)(r10 -24) = r2
regs=2 stack=0 before 2: (b7) r2 = 30007
regs=2 stack=0 before 1: (61) r1 = *(u32 *)(r6 +12)
6: (7b) *(u64 *)(r10 -8) = r1
7: (b7) r1 = 0
8: (7b) *(u64 *)(r10 -16) = r1
last_idx 8 first_idx 0
regs=2 stack=0 before 7: (b7) r1 = 0
9: (18) r7 = 0xffff89de0a4b1600
11: (85) call bpf_get_smp_processor_id#8
12: (bf) r4 = r10
13: (07) r4 += -24
14: (bf) r1 = r6
15: (bf) r2 = r7
16: (bf) r3 = r0
17: (b7) r5 = 24
18: (85) call bpf_perf_event_output#25
invalid indirect read from stack off -24+16 size 24
processed 18 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

ERROR: Error loading program: tracepoint:irq:irq_handler_entry
```
As far as I can tell, the "-24+16 size 24" error is bogus. Instructions 3, 6 and 8 completely initialize the 24 byte region starting at -24 + 16 = 8.

If anyone has ideas I would be ecstatic to hear them.

fbs:
Maybe its unhappy about adding ctx into the mix? The ir and output look valid in both cases, but this seems to fix it:

```
diff --git a/src/ast/codegen_llvm.cpp b/src/ast/codegen_llvm.cpp
index 7c72ce3e..a32e9af1 100644
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1552,7 +1552,7 @@ void CodegenLLVM::visit(FieldAccess &acc)
                             b_.CreateGEP(ctx_, b_.getInt64(field.offset)));
       expr_ = b_.CreateIntCast(expr_, b_.getInt64Ty(), false);
       expr_ = b_.CreateAnd(expr_, b_.getInt64(0xFFFF));
-      expr_ = b_.CreateAdd(expr_, b_.CreatePtrToInt(ctx_, b_.getInt64Ty()));
+      // expr_ = b_.CreateAdd(expr_, b_.CreatePtrToInt(ctx_, b_.getInt64Ty()));
     }
```

And this works too

```
diff --git a/src/ast/codegen_llvm.cpp b/src/ast/codegen_llvm.cpp
index 7c72ce3e..b2d3d6b3 100644
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1552,7 +1552,8 @@ void CodegenLLVM::visit(FieldAccess &acc)
                             b_.CreateGEP(ctx_, b_.getInt64(field.offset)));
       expr_ = b_.CreateIntCast(expr_, b_.getInt64Ty(), false);
       expr_ = b_.CreateAnd(expr_, b_.getInt64(0xFFFF));
-      expr_ = b_.CreateAdd(expr_, b_.CreatePtrToInt(ctx_, b_.getInt64Ty()));
+      // expr_ = b_.CreateAdd(expr_, b_.CreatePtrToInt(ctx_, b_.getInt64Ty()));
+      expr_ = b_.CreateAdd(expr_, b_.getInt64(1234));
     }
```

But here it does an i16 load to avoid the masking:

```
Bytecode:
0: (bf) r6 = r1
1: (69) r1 = *(u16 *)(r6 +12)
2: (b7) r2 = 30007
3: (7b) *(u64 *)(r10 -24) = r2
4: (07) r1 += 1234
5: (7b) *(u64 *)(r10 -8) = r1
6: (b7) r1 = 0
7: (7b) *(u64 *)(r10 -16) = r1
last_idx 7 first_idx 0
```

danobi:
> Maybe its unhappy about adding ctx into the mix? The ir and output look valid in both cases, but this seems to fix it:

Commenting out the ctx + offset fixed it for me. Not sure there's anything we can do from our side. It seems like a kernel issue to me.

src/ast/codegen_llvm.cpp
```diff
@@ -1539,6 +1539,21 @@ void CodegenLLVM::visit(FieldAccess &acc)
...
+    else if (field.type.IsIntTy() && field.is_data_loc)
+    {
+      // `is_data_loc` should only be set if field access is on `args` which
+      // has to be a ctx access
+      assert(type.IsCtxAccess());
+      assert(ctx_->getType() == b_.getInt8PtrTy());
+
+      // Top 2 bytes are length (which we'll ignore). Bottom two bytes are
+      // offset which we add to the start of the tracepoint struct.
+      expr_ = b_.CreateLoad(b_.getInt32Ty(),
```
fbs, line 1551: Loading an i32 from an i8* here which makes llvm complain. Inserting a  `b_.CreatePointerCast(ctx_, b_.getInt32Ty()->getPointerTo()` shoud fix it

fbs: Oops, comment should go on the line below.

Also requires `field.offset/4` as creategep will now work with 32b steps.

#1396 will help clean this up

danobi:
Chatted w/ Alexei. When we add an offset to ctx and store the result to the stack, the verifier thinks it's a register spill and doesn't let anyone read from the location on stack.

We'll probably have to live with the unprintable data_loc integers. Shouldn't be too bad functionally b/c it's not like userspace could do anything with that pointer.

src/ast/codegen_llvm.cpp
danobi: Fixed

fbs:
Just landed the new tuple code, so this might need a rebase

danobi:
Rebased.

Also moved the type rewrite from semantic analyser into clang parser. Probably better to reduce the spread of the logic. Also made it easier to write a better clang parser test.

fbs (approved):

PR #1543, opened 2020-09-29
clang-parser: Remove extraneous redefinition check

danobi:
We already check for redefinition errors (among other things) at
`handler.check_diagnostics(..)`. The diagnostic check occurs before we
walk the AST. The removed check is done while walking the AST.

Furthermore, the removed redefinition check is wrong. I don't think the
boolean operators are correct (most of the `&&`s should be `||`s).

##### Checklist


danobi:
Hmm, test failures. Will look again

danobi:
Ok

PR #1544, opened 2020-09-29
tests: Remove profile_order runtime test

danobi:
This test is consistently flakey in the CI. And to be honest, I'm not
quite sure what exactly the test is testing.

##### Checklist


fbs:
I think it is testing that probes fire in the order theyre created in. Probably related to the comment in bpftrace.cpp

```
  // The kernel appears to fire some probes in the order that they were
  // attached and others in reverse order. In order to make sure that blocks
  // are executed in the same order they were declared, iterate over the probes
  // twice: in the first pass iterate forward and attach the probes that will
  // be fired in the same order they were attached, and in the second pass
  // iterate in reverse and attach the rest.
```

But as its flaky were better off removing it

PR #1548, opened 2020-10-07
semantic_analyser: Fix state leak between str() processing

danobi:
Before, if the semantic analyser encountered a str() call on a
positional parameter, it would leak state into following `str()` calls
and still think it was processing a positional parameter.

##### Checklist


danobi:
Gonna skip the changelog entry b/c the offending commit is still unreleased

danobi:
cc @viktormalik

viktormalik (approved):

PR #1549, opened 2020-10-08
Refactor wildcard matching

### LINKED ISSUE #565: List and attach syntax are inconsistent
We implement the parsing for matching probes in a list here:
https://github.com/iovisor/bpftrace/blob/master/src/list.cpp#L116-L139 

And this logic is basically duplicated for filtering probes to attach:
https://github.com/iovisor/bpftrace/blob/master/src/bpftrace.cpp#L152

These should use the same syntax and provide the same filtering functionality.

#557 and #508 both touch this

### LINKED ISSUE #1489: Support quotes in probe listing
We require quotes for "complex names", like:

```
$ bpftrace  -e 'uprobe:"/usr/lib/x86_64-linux-gnu/libstdc++.so.6":_ZSt18uncaught_exceptionv { printf("HIT\n"); }'
Attaching 1 probe...
^C
```

But using quotes breaks the listing:

```
$ bpftrace -l 'u:"/usr/lib/x86_64-linux-gnu/libstdc++.so.6":*'
ERROR: uprobe target '"/usr/lib/x86_64-linux-gnu/libstdc++.so.6"' does not exist or is not executable
```

_Originally posted by @fbs in https://github.com/iovisor/bpftrace/issues/1487#issuecomment-682074262_

### LINKED ISSUE #1581: bpftrace -l lists duplicate probles
### What reproduces the bug?
```
$ sudo ./src/bpftrace -l | sort | uniq -c -d | sort -n -r
    147 kprobe:__ia32_sys_old_shmctl
    119 kprobe:__ia32_compat_sys_io_pgetevents_time32
     42 kprobe:__ia32_sys_ipc
     32 kprobe:__ia32_sys_uselib
     21 kprobe:__ia32_sys_subpage_prot
     18 kprobe:type_show
     17 kprobe:name_show
     17 kprobe:modalias_show
     15 kprobe:sys_ni_syscall
     13 kprobe:__ia32_sys_old_semctl
     13 kprobe:__ia32_sys_old_msgctl
     12 kprobe:__ia32_sys_ppoll_time32
     11 kprobe:init_once
      9 kprobe:print_cpus_offline
      8 kprobe:state_show
      7 kprobe:event_show
      6 kprobe:umask_show
      6 kprobe:new_id_store
      6 kprobe:__ia32_sys_pciconfig_iobase
      5 kprobe:version_show
[...]
```

In total 1407 probes are duplicated (happens with `kprobe` and `kfunc`). Is this expected?

I looked in the code, and it turns out the kprobe duplicates are coming from `/sys/kernel/debug/tracing/available_filter_functions`. Not sure if they should have been de-duplicated there already, or should be de-duplicated in bpftrace?

### bpftrace --info
```
System
  OS: Linux 5.8.15-201.fc32.x86_64 #1 SMP Thu Oct 15 15:56:44 UTC 2020
  Arch: x86_64

Build
  version: v0.11.0-212-g4808
  LLVM: 10.0.1
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  libbpf: yes
  libbpf btf dump: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
```

viktormalik:
The main goal of this work is to unify code used for wildcard matching into one place. Currently, probe listing uses different code to resolve wildcards than probe attachment does, which results in possible discrepancies (such as in #565). The idea is that if some probes are listed, they can be attached with the same syntax.

There are two primary changes done:
- all code related to probe matching is moved into a new class `ProbeMatcher`,
- when listing probes, we use the standard Bison parser and `AttachPointParser` to parse the listing query. This required a change in the parser grammar: different rule for attach points for listing (when attach point is a listing query) and for probe attachement.

Some details worth mentioning:
- Listing allows to use wildcards in probe type, while attachment originally doesn't. To cope with this, `AttachPointParser` tries to expand the probe type and when multiple types are matched, it creates new `AttachPoint` objects that replace the original `AttachPoint` object created by the Bison parser.
- On the other hand, this PR allows to attach a single program to multiple attach points from different providers using a wildcard (e.g. `k*:vfs_open`). This should not break anything, though, since it was possible before by just simply enumerating the attach points (`kprobe:vfs_open,kfunc:vfs_open`).
- One round of semantic analyser is run when listing probes to make the errors consistent for listing and attachement.
- All usages of regular expressions should be now removed.

Resolves #565, resolves #1489, fixes #1581.

##### Checklist


tests/runtime/usdt
```diff
@@ -1,5 +1,5 @@
 NAME "usdt probes - list probes by file"
-RUN bpftrace -l 'usdt:./testprogs/usdt_test'
+RUN bpftrace -l 'usdt:./testprogs/usdt_test:*'
```
fbs, line 2: why is this needed?

viktormalik: This is the normal syntax to "attach to all probes for that binary". Since I wanted to unify attachment and listing syntax, we either have to require the explicit wildcard for listing or make the wildcard implicit for attachment. I prefer the first option. Also, we would probably have to make the wildcard implicit for all probe types, which I don't really like.

fbs: We try to avoid breaking compatibility with existing documentation (books) when possible.  But the current behaviour is inconsistent too:

```
$ sudo bpftrace_master  -l 'u:/bin/bash' | head -n2
uprobe:/bin/bash:__libc_csu_fini
uprobe:/bin/bash:__libc_csu_init
$ sudo bpftrace_master  -l 'k:' | head -n2
$
````

So I'm ok with breaking it as long as all probe types end up having consistent behaviour. @brendangregg  how do you feel about it.

But I do not like  the error this throws currently, its confusing:
```
$ sudo bpftrace -l 'u:/bin/bash'
stdin:1:13-14: ERROR: syntax error, unexpected {
u:/bin/bash {}
            ~
```

And can even segfault:

```
$ sudo bpftrace -l 'k:'
Segmentation fault
```

Comment (viktormalik) [reply tests/runtime/usdt:2]: > We try to avoid breaking compatibility with existing documentation (books) when possible.
> ...
> So I'm ok with breaking it as long as all probe types end up having consistent behaviour.

The current behaviour in this PR is that every probe part should be specified and non-empty (at least `*` should be given). The only part that can be empty/skipped is namespace for usdt (per documentation) and binary name for uprobe/usdt if PID is specified.

On a second thought, I would agree with empty part of a probe behaving as `*`. That would keep the backwards compatibility (at least for the above cases). Also, AFAIK, DTrace uses a similar syntax.

> 
> But I do not like the error this throws currently, its confusing:
> ...
> And can even segfault:

I'm currently working on this, this is why the PR is in draft state. I'm thinking about running one iteration of semantic analyser, which already has good error messages (and they would be consistent, too).

**Comment by viktormalik** [comment]
The PR is ready for review. 

The remaining thing to decide is: how do we treat empty probe parts (both for listing and attachment)? Currently, probe listing allows empty probe parts (by treating them as '*') but probe attachment doesn't. Since we want to unify this, we have to choose one of the approaches. I'd prefer treating empty probe parts as '*', which will keep the compatibility and make the attach point/listing specs shorter. @fbs, @brendangregg WDYT?

**Comment by mmisono** [comment]
I don't review the code yet but it seems something wrong with struct listing.

before:
```
% time sudo ./src/bpftrace -lv 'k:vfs_open'
kprobe:vfs_open
a0.10s user 0.33s system 91% cpu 0.473 total
```

after:
```
% time sudo ./src/bpftrace -lv 'struct path'
struct path {
        struct vfsmount *mnt;
        struct dentry *dentry;
};
struct path_cond {
        kuid_t uid;
        umode_t mode;
};
11.08s user 0.21s system 85% cpu 13.137 total  // too slow
```

**Comment by mmisono** [comment]
Anther issuse: 

```
% sudo ASAN_OPTIONS=detect_leaks=0 ./src/bpftrace -e 't*a:a{}'
FATAL: program type invalid\
zsh: abort      sudo ASAN_OPTIONS=detect_leaks=0 ./src/bpftrace -e 't*a:a{}'
```

The program shouldn't abort, should print an error.

**Comment by viktormalik** [comment]
> I don't review the code yet but it seems something wrong with struct listing.
> 
> before:
> 
> ```
> % time sudo ./src/bpftrace -lv 'struct path'
> BTF: using data from /sys/kernel/btf/vmlinux
> struct path {
>         struct vfsmount *mnt;
>         struct dentry *dentry;
> };
> 0.13s user 0.33s system 92% cpu 0.505 total
> ```
> 
> after:
> 
> ```
> % time sudo ./src/bpftrace -lv 'struct path'
> BTF: using data from /sys/kernel/btf/vmlinux
> struct path {
>         struct vfsmount *mnt;
>         struct dentry *dentry;
> };
> struct path_cond {
>         kuid_t uid;
>         umode_t mode;
> };
> 11.08s user 0.21s system 85% cpu 13.137 total  // too slow
> ```

Thanks for noticing, this should be fixed now.

**Comment by viktormalik** [comment]
> Anther issuse:
> 
> ```
> % sudo ASAN_OPTIONS=detect_leaks=0 ./src/bpftrace -e 't*a:a{}'
> FATAL: program type invalid\
> zsh: abort      sudo ASAN_OPTIONS=detect_leaks=0 ./src/bpftrace -e 't*a:a{}'
> ```
> 
> The program shouldn't abort, should print an error.

Yes, the problem here is that `t*a` doesn't match any probe type, which I don't handle properly (hence the abort). But this reveals another inconsistence in bpftrace - how should we handle wildcarded attach points that have no matches?

From the current master:
```
# bpftrace -e 't:*:nonsense {}'
stdin:1:1-13: ERROR: tracepoints not found: *:nonsense
t:*:nonsense {}
~~~~~~~~~~~~
```
but on the other hand:
```
bpftrace -e 'k:nonsense* {}'
No probes to attach
```
I'd say that this should be unified - either always print an error or always ignore that attach point.

**Comment by mmisono** [comment]
I agree with you, it's better if the behavior is unified. But I guess the reason currently trying to attach nonexistent kprobe is allowed is to make the script work as much as possible because kernel function names can change from version to version. For example, in my environment (kernel 5.9), one of the kprobe in biostacks.bt does not exist.

```
% sudo ./src/bpftrace ../tools/biostacks.bt
Attaching 5 probes...
WARNING: could not attach probe kprobe:blk_start_request, skipping.
```

or more simple example is

```
% sudo ./src/bpftrace -e "k:a,k:f {}"
Attaching 2 probes...
WARNING: could not attach probe kprobe:a, skipping.
```

We also want no to break existing scripts as much as possible. So I think we should keep the current behavior, that means error if wildcard matches any probe type, error if wildcard attach point exists except kprobe (I'm not sure about other than kprobe and tracepoint, so we need check other probes)

**Comment by viktormalik** [comment]
> We also want no to break existing scripts as much as possible. So I think we should keep the current behavior, that means error if wildcard matches any probe type, error if wildcard attach point exists except kprobe (I'm not sure about other than kprobe and tracepoint, so we need check other probes)

This sounds like a good way to go. I'd say that we shoudn't fail on non-existing probes for probe types that are not stable: kprobe, kfunc, uprobe, usdt. To sum it up:
- error if probe type doesn't exist/match,
- ignore non-existing attach points for non-stable probe types (kprobe, kfunc, uprobe, usdt),
- error if attach point doesn't exist/match for all other probe types.

File: tests/runtime/basic [hunk 2]
```diff
@@ -96,6 +96,12 @@ EXPECT kretfunc
 REQUIRES_FEATURE btf
 TIMEOUT 1
 
+NAME listing with wildcarded probe type
```
mmisono, line 99: Just to make sure, these newly added tests are worked as the same as before? or some behavior are changed?

mmisono:
Looks like this works before: `sudo ./src/bpftrace -lv "*:*:*"`

src/parser.yy
```diff
@@ -167,7 +167,22 @@ probes : probes probe { $$ = $1; $1->push_back($2); }
...
+                                   {
+                                     error(@$, "unexpected listing query format");
+                                     YYERROR;
+                                   }
+                                 }
+      | attach_points END { if (driver.listing_)
+                              $$ = new ast::Probe($1, nullptr, nullptr);
+                            else
+                            {
+                              error(@$, "unexpected end of file, expected {");
```
mmisono, line 182:
```
% sudo ./src/bpftrace -l 'k:vfs_open // {}'
stdin:1:12-13: ERROR: syntax error, unexpected end predicate, expecting {
k:vfs_open // {}
           ~
```

In this case, "unexpected listing query format" is more appropriate?

src/btf.cpp
```diff
@@ -585,48 +607,33 @@ std::unique_ptr<std::istream> BTF::get_funcs(std::regex *re,
...
+  return params;
+#else
+  LOG(ERROR) << "Could not get kfunc arguments "
+                "(HAVE_LIBBPF_BTF_DUMP_EMIT_TYPE_DECL is not set)" return {};
+  return {};
+#endif
 }
 
-void BTF::display_structs(std::regex *re) const
+std::set<std::string> BTF::get_structs() const
```
mmisono, line 630: nit: whichever is ok, but the function is `get_all_structs`?

mmisono:
I realized that now 
```
% sudo ./src/bpftrace -l 'k:vfs_write,k:vfs_open'
kprobe:vfs_write
kprobe:vfs_open
```
works. It is good thing, but it seems `-v` option does not work. `./src/bpftrace -lv 'k:vfs_write,k:vfs_open'` gives the same output as the above.

mmisono:
I'll check the remaining code when I have time.

tests/runtime/basic
viktormalik: This one works the same as before, but e.g. listing userspace probes by wildcarded probe type didn't work.

viktormalik:
> I realized that now
> 
> ```
> % sudo ./src/bpftrace -l 'k:vfs_write,k:vfs_open'
> kprobe:vfs_write
> kprobe:vfs_open
> ```
> 
> works. It is good thing, but it seems `-v` option does not work. `./src/bpftrace -lv 'k:vfs_write,k:vfs_open'` gives the same output as the above.

That's expected, kprobes do not list arguments with `-v` (neither they did before). It's kfuncs doing that:
```
# src/bpftrace -lv 'f:vfs_write,f:vfs_open'
kfunc:vfs_write
    struct file * file
    [...]
kfunc:vfs_open
    const struct path * path
    [...]
```

mmisono:
> > I realized that now
> > ```
> > % sudo ./src/bpftrace -l 'k:vfs_write,k:vfs_open'
> > kprobe:vfs_write
> > kprobe:vfs_open
> > ```
> > 
> > 
> > works. It is good thing, but it seems `-v` option does not work. `./src/bpftrace -lv 'k:vfs_write,k:vfs_open'` gives the same output as the above.
> 
> That's expected, kprobes do not list arguments with `-v` (neither they did before). It's kfuncs doing that:
> 
> ```
> # src/bpftrace -lv 'f:vfs_write,f:vfs_open'
> kfunc:vfs_write
>     struct file * file
>     [...]
> kfunc:vfs_open
>     const struct path * path
>     [...]
> ```

my bad,  I forgot about it.

src/btf.cpp
viktormalik: correct, fixed

viktormalik:
> Looks like this works before: `sudo ./src/bpftrace -lv "*:*:*"`

Right, thanks. This should now work as before.

src/parser.yy
viktormalik: True, but since this is an automatic error message by bison, I haven't yet found a way to implement it in a short way. And I don't like the idea of making large changes to the parser (potentially introducing errors) just for the sake of fixing one error message.

mmisono: I see. I also don't familiar with bison, and the current behavior is OK to me.

mmisono:
```
% sudo ./src/bpftrace -l "k*:vfs_open"
kfunc:vfs_open
kprobe:vfs_open
kretfunc:vfs_open
````
this is ok but 

```
% sudo  ./src/bpftrace -l "f*:vfs_open"
stdin:1:1-12: ERROR: No probe type matched
f*:vfs_open
~~~~~~~~~~~

% sudo  ./src/bpftrace -lv "*:vfs_open"
stdin:1:1-1: ERROR: vfs_open is not a hardware probe
*:vfs_open

stdin:1:1-1: ERROR: vfs_open is not a software probe
*:vfs_open

```

**Comment by viktormalik** [comment]
> ```
> % sudo  ./src/bpftrace -l "f*:vfs_open"
> stdin:1:1-12: ERROR: No probe type matched
> f*:vfs_open
> ~~~~~~~~~~~
> 
This is intentional since `f*` does not match any probe type. I improved the error message, though:
```
stdin:1:1-12: ERROR: No probe type matched for f*
```

> % sudo  ./src/bpftrace -lv "*:vfs_open"
> stdin:1:1-1: ERROR: vfs_open is not a hardware probe
> *:vfs_open
> 
> stdin:1:1-1: ERROR: vfs_open is not a software probe
> *:vfs_open
> ```

This is now fixed:
```
# src/bpftrace -lv "*:vfs_open"
kfunc:vfs_open
    const struct path * path
    [...]
kprobe:vfs_open
```
I also removed `kretfunc` as it is just a duplicate of `kfunc` (similarly, `kretprobe` are not listed).

**Comment by mmisono** [comment]
OK^

It seems listing can only work without other queries. This is not supported previously, but I think it would be useful. What do you think?

```
sudo  ./src/bpftrace -lv "f:vfs_open, struct path"
stdin:1:12-12: ERROR: Unrecognized probe type:
f:vfs_open, struct path
```

or

```
sudo ./src/bpftrace -lv "struct path,struct file"
```

**Comment by viktormalik** [comment]
> It seems listing can only work without other queries. This is not supported previously, but I think it would be useful. What do you think?

Might be useful, but I'd rather not extend this PR with additional features, it's quite large already. We can add this later in a separate PR.

File: src/probe_matcher.h [hunk 5]
```diff
@@ -0,0 +1,117 @@
...
+struct ProbeListItem
+{
+  std::string path;
+  std::string alias;
+  uint32_t type;
+  uint64_t defaultp;
+};
+
+// clang-format off
+const std::vector<ProbeListItem> SW_PROBE_LIST = {
```
mmisono, line 26: I guess we need to use a singleton to thread-safe initialization. something like: https://github.com/iovisor/bpftrace/blob/77a25cf9eb76f1107240ee26cca8b7223bb5184f/src/ast/semantic_analyser.cpp#L22-L34
@danobi probably know this.

mmisono (commented):
Overall looks good and good improvements to me, but definitely more reviewers are needed because of its size. Also, I wonder if we can have tests for ProbeMatcher.

viktormalik:
> Overall looks good and good improvements to me, but definitely more reviewers are needed because of its size. Also, I wonder if we can have tests for ProbeMatcher.

I already thought about this, but then I realized that matching is already tested in `tests/bpftrace.cpp`, `tests/parser.cpp`, and in runtime tests. It might make sense to group these into a separate test class for ProbeMatcher (although I would, again, postpone it to another PR).

fbs:
Overal it looks good to me, nice cleanup.

I'm not that happy with the way this hooks into semantic analysis. It's another thing bolted into the semantic analyser that makes it a bit more complex. A separate simpeler pass seems more appropriate to me. But as its a nice cleanup we shouldn't block on it imo

fbs:
@danobi any feedback?

I want to merge this soon but it needs a rebase @viktormalik .

viktormalik:
Rebased, ready to merge.

There's still an unresolved issue - how to treat empty probe parts? See [this review](https://github.com/iovisor/bpftrace/pull/1549#pullrequestreview-505181521) and [this comment](https://github.com/iovisor/bpftrace/pull/1549#issuecomment-712754036) for details. Currently, this PR requires that each probe part is specified, which is not backwards compatible with listing (previously, listing allowed to skip probe parts). I think that this is not a big issue (since it's only listing), I'd suggest merging this PR and then maybe adding a possibility to omit empty probe parts as another PR.

viktormalik:
One embedded build test time-outed, but it passes on my fork branch ([check here](https://github.com/viktormalik/bpftrace/runs/1516055475)). I guess a restart could solve this.

viktormalik:
Is this waiting for me? I think that the CI fail should be resolved by a restart and otherwise the PR should be ready for merge.

fbs:
Looks ok to me :)

fbs (approved):

danobi:
Needs a rebase

viktormalik:
Rebased

PR #1557, opened 2020-10-13
lang: Add support for pointer to pointer

danobi:
Previously, you could not do something like:
```
$pp = (char **) arg0;
print(**$pp);
```

Instead, you had to do something like:
```
$p = *arg0;
$val = *((char *)$p);
print($val);
```

Notice how in the second example we had to load a pointer as
an integer, then cast it back to a pointer and dereference again.

The first example is better because it's more clear what's going
on.

##### Checklist

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -333,8 +333,9 @@ expr : int                                      { $$ = $1; }
      | expr DOT INT                             { $$ = new ast::FieldAccess($1, $3, @3); }
      | expr PTR ident                           { $$ = new ast::FieldAccess(new ast::Unop(token::MUL, $1, @2), $3, @$); }
      | expr "[" expr "]"                        { $$ = new ast::ArrayAccess($1, $3, @2 + @4); }
-     | "(" IDENT ")" expr %prec CAST            { $$ = new ast::Cast($2, false, $4, @1 + @3); }
-     | "(" IDENT MUL ")" expr %prec CAST        { $$ = new ast::Cast($2, true, $5, @1 + @4); }
+     | "(" IDENT ")" expr %prec CAST            { $$ = new ast::Cast($2, false, false, $4, @1 + @3); }
+     | "(" IDENT MUL ")" expr %prec CAST        { $$ = new ast::Cast($2, true, false, $5, @1 + @4); }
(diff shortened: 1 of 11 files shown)


src/parser.yy
```diff
@@ -330,8 +330,9 @@ expr : int                                      { $$ = $1; }
      | expr DOT INT                             { $$ = new ast::FieldAccess($1, $3, @3); }
      | expr PTR ident                           { $$ = new ast::FieldAccess(new ast::Unop(token::MUL, $1, @2), $3, @$); }
      | expr "[" expr "]"                        { $$ = new ast::ArrayAccess($1, $3, @2 + @4); }
-     | "(" IDENT ")" expr %prec CAST            { $$ = new ast::Cast($2, false, $4, @1 + @3); }
-     | "(" IDENT MUL ")" expr %prec CAST        { $$ = new ast::Cast($2, true, $5, @1 + @4); }
+     | "(" IDENT ")" expr %prec CAST            { $$ = new ast::Cast($2, false, false, $4, @1 + @3); }
+     | "(" IDENT MUL ")" expr %prec CAST        { $$ = new ast::Cast($2, true, false, $5, @1 + @4); }
+     | "(" IDENT MUL MUL ")" expr %prec CAST    { $$ = new ast::Cast($2, true, true, $6, @1 + @5); }
```
fbs, line 335: Wouldn't doing `ast::Cast(ast::Cast(...), ...)` be possible, that way we can easily do N depth pointers (although I never seen it going deeper than 3 so it might not be useful)

danobi: I'm not sure that works. `ast::Cast(ast::Cast(...), ...)` is equivalent to `(uint8 *) (uint8 *) $foo` instead of `(uint8 **) $foo`, right?

fbs: Hmm you're right. If we ever need more nesting we could do something like 


```
Cast::Cast(const std::string &type,
           bool is_pointer,
           int   depth,
           Expression *expr) {
....
   for (i < depth)
      cast.type = CreatePointer(cast.type)

```

Or even lift it to the CreatePointer method.

But I don't see a reason for this now

danobi: Yeah, it should be fairly simple. I grepped around linux codebase and there's only a handful of triples pointers and no quadruple pointers.

fbs (approved):
needs rebase tho

Issue #1560, opened 2020-10-14
Add support for assigning one tuple element value


rdmello:
### Is your feature request related to a problem? Please describe.
I'm using a `tuple` to store some data after looking at the syntax [here](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#14----tuples).

This allows me to access the data in the tuple using a convenient `@tuple.1` syntax. However, I get an error message when I try to assign to it using `@tuple.1 = 1;`.

I'm running bpftrace v0.11.1 on Gentoo. Here's a one-liner that reproduces this:

```bash
root# bpftrace -e 'interval:s:1 { @x = (3, 3); @x.1 = 1; printf("%d\n", @x.1); }'
stdin:1:34-35: ERROR: syntax error, unexpected =, expecting }
interval:s:1 { @x = (3, 3); @x.1 = 1; printf("%d\n", @x.1); }
                                 ~
```

### Describe the solution you'd like

It would be convenient if assigning to a `tuple` element just worked:

```bash
root# bpftrace -e 'interval:s:1 { @x = (3, 3); @x.1 = 1; printf("%d\n", @x.1); }'
Attaching 1 probe...
1
1
^C
```

### Describe alternative solutions or features you've considered

* The current solution of updating a tuple by assigning all the tuple values to it does work fine:

```bash
root# bpftrace -e 'interval:s:1 { @x = (3, 3); @x = (@x.0, 1); printf("%d\n", @x.1); }'
Attaching 1 probe...
1
1
^C
```

but allowing direct assignment would be a little more convenient.

* If this is not possible due to syntax limitations, it would be good to document this in the reference guide.

danobi:
Right, tuples are currently immutable once created. Just like python. Seems like python can get away with doing that so I'd like to see how long we can get away from it as well.

fbs:
Agree on keeping immutable (atleast until we have a valid real world usecase)

But we should improve error message, so reopening this. 

```
stdin:1:34-35: ERROR: syntax error, unexpected =, expecting }
interval:s:1 { @x = (3, 3); @x.1 = 1; printf("%d\n", @x.1); }
```

PR #1563, opened 2020-10-15
Provide better error message when assigning to tuple element

rdmello:
##### Description

This change addresses the second part of #1560 by adding an explicit error message when assigning to a tuple element:

```
./src/bpftrace -e 'i:s:1 { $x = (1, 2); $x.1 = 1; }'
stdin:1:22-30: ERROR: Tuples are immutable. Consider creating a new tuple and assigning it instead.
i:s:1 { $x = (1, 2); $x.1 = 1; }
                     ~~~~~~~~
```

I went with the approach of making this a parse-time error for simplicity. I ran all the unit tests and verified that the number of shift-reduce conflicts haven't changed after implementing this.

An alternative approach is to allow tuple assignment to parse successfully, and then make tuple assignment error later during semantic analysis. This would be a little more complex, but would also provide a better error message (since it can disambiguate between assigning to tuples vs. assigning to maps). Let me know if this approach is preferable.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ and this project adheres to
   - [#1542](https://github.com/iovisor/bpftrace/pull/1542)
 - Change a part of the message of '-v' output
   - [#1553](https://github.com/iovisor/bpftrace/pull/1553)
+- Improve tuple assignment error message
+  - [#1563](https://github.com/iovisor/bpftrace/pull/1563)
 
 #### Deprecated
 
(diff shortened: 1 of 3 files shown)


danobi (commented):
LGTM, but can you add a couple more tests for other conditions?

eg `(1,0).1 = 0`, `((1,0),3).0.0 = 3`, `((1,0),3).0 = (0,1)`. You can just check that parsing failed for these, no need to check error message.

danobi:
Oh, also update the changelog, please.

rdmello:
@danobi Thank you for reviewing this...I added a few more tests and updated the changelog.

danobi (approved):

danobi:
Thanks for the PR!

PR #1565, opened 2020-10-16
Fix llvm errors of PositonalParameter

mmisono:
The first commit fixes the error reported by AddressSanitizer. As of the second commit, I didn't encounter a specific problem but I think it is a correct fix. Please see each commit message for the details.

##### Checklist


danobi (approved):
Nice, thanks

PR #1568, opened 2020-10-21
Error if Positional Params num is zero

mmisono:
Positional parameter zero ($0) is invalid. Previously, using $0 causes
std::out_of_range error due to the underflow and confuses users.

```
% sudo ./src/bpftrace -e 'k:$0 {}' vfs_open
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 18446744073709551615) >= this->size() (which is 1)
zsh: abort      sudo ./src/bpftrace -e 'k:$0 {}' vfs_open
```

This commit changes this and in the parser check the positional params
num is not zero. This can be also done in the lexical analyzer, but to
print verbose an error message do this in the parser. Now the error message is

```
% sudo ./src/bpftrace -e 'k:$0 {}' vfs_open
stdin:1:1-5: ERROR: param $0 is out of integer range [1, 9223372036854775807]
k:$0 {}
~~~
```

##### Checklist

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -215,7 +215,9 @@ ternary : expr QUES expr COLON expr { $$ = new ast::Ternary($1, $3, $5, @$); }
 
 param : PARAM      {
                      try {
-                       $$ = new ast::PositionalParameter(PositionalParameterType::positional, std::stol($1.substr(1, $1.size()-1)), @$);
+                       long n = std::stol($1.substr(1, $1.size()-1));
+                       if (n == 0) throw std::exception();
+                       $$ = new ast::PositionalParameter(PositionalParameterType::positional, n, @$);
                      } catch (std::exception const& e) {
                        error(@1, "param " + $1 + " is out of integer range [1, " +
                              std::to_string(std::numeric_limits<long>::max()) + "]");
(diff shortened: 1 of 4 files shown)


mmisono:
I realized that currently, if the positional parameter is not given, it resulted in an error. For example:

```
% sudo ./src/bpftrace -e 'i:ms:$1 {}'
stdin:1:1-8: ERROR: Failed to parse '': stoi
i:ms:$1 {}
~~~~~~~
```

But iirc some scripts expects that if positional parameters are not defined, the value becomes zero. I'll work on it later.

src/parser.yy
```diff
@@ -215,7 +215,9 @@ ternary : expr QUES expr COLON expr { $$ = new ast::Ternary($1, $3, $5, @$); }
 
 param : PARAM      {
                      try {
-                       $$ = new ast::PositionalParameter(PositionalParameterType::positional, std::stol($1.substr(1, $1.size()-1)), @$);
+                       long n = std::stol($1.substr(1, $1.size()-1));
+                       if (n == 0) throw std::exception();
```
danobi, line 219: Why throw an exception over calling `error(...); YYERROR`?

mmisono: The intention is to let the catch block do the work.

https://github.com/iovisor/bpftrace/blob/a03b5c4bbb12a25345a2fdec78374963ff936131/src/parser.yy#L217-L223

But I don't have a strong opinion, which do you prefer?

danobi: Ah I didn't notice that line was already in a try/catch block. It's fine then

danobi (approved):

mmisono:
@fbs 
Is It OK to merge this? My fixes' CHANGELOG would conflict, so I'll fix it one by one.

fbs (approved):

PR #1570, opened 2020-10-23
Fix LNOT

mmisono:
CreateUint() expects bit size, not byte size. This fixes the following
error.

```
% sudo ./src/bpftrace -e 'BEGIN { !(int32)0 }'
bpftrace: /home/ubuntu/work/bpftrace/src/types.cpp:251: bpftrace::SizedType bpftrace::CreateInteger(size_t, bool): Assertion `bits == 0 || bits == 8 || bits == 16 || bits == 32 || bits == 64' failed.
zsh: abort      sudo ./src/bpftrace -e 'BEGIN { !(int32)0 }'
```

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1594,7 +1594,19 @@ void SemanticAnalyser::visit(Unop &unop)
     }
   }
   else if (unop.op == Parser::token::LNOT) {
-    unop.type = CreateUInt(type.size);
+    // CreateUInt() abort if a size is invalid, so check the size here
+    if (!(type.size == 0 || type.size == 1 || type.size == 2 ||
+          type.size == 4 || type.size == 8))
+    {
+      LOG(ERROR, unop.loc, err_)
+          << "The " << opstr(unop)
+          << " operator can not be used on expressions of type '" << type
+          << "'";
+    }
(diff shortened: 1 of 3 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -1594,7 +1594,19 @@ void SemanticAnalyser::visit(Unop &unop)
...
+          type.size == 4 || type.size == 8))
+    {
+      LOG(ERROR, unop.loc, err_)
+          << "The " << opstr(unop)
+          << " operator can not be used on expressions of type '" << type
+          << "'";
+    }
+    else
+    {
+      unop.type = CreateUInt(8 * type.size);
```
fbs, line 1608: shouldn't we just return a fixed size int here?  Logical not should give a boolean so u64 might be better?

But not sure hwo much that will break.

mmisono: I'm not sure but with #1532, there is a chance that bytecode size becomes smaller?

fbs: Not sure. I Wouldn't mind  keeping this PR as is an investigate it further in #1532 & friends

mmisono:
rebased

PR #1571, opened 2020-10-23
Optimize unknown/incomplete types resolution

viktormalik:
Following the discussion in [#1485](https://github.com/iovisor/bpftrace/pull/1485#discussion_r506187878), I implemented a fix that will speed up the resolution of incomplete types and unknown typedefs. The main goal is to limit the number of iterations of type resolution.

The proposed solution is:

- Since incomplete types mostly occur for tracepoints, the number of iterations is given by the number of nested field accesses in tracepoint probes. This will allow to handle, e.g.:
```
src/bpftrace -e 't:napi:napi_poll { printf("%s\n", args->napi->dev->name); }' -b
```
which needs at least 2 iterations.

- Unknown typedefs (see #1485 for details) must be resolved completely, since they will make the Clang parser fail, even if the types are not used in the program. Usually, no more than 3 iterations are needed.
An example that needs 3 iterations:
```
src/bpftrace -e 't:syscalls:sys_enter_* { @[args->__syscall_nr] = count(); } i:ms:1 { exit(); }' -b
```

Still, there may be cases that I missed which require more iterations but in such case, bpftrace will fail with "unknown type" error and the user will have to include appropriate headers.

Any thoughts on this approach? Ideas when this could not work?

##### Checklist

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -529,8 +529,7 @@ bool ClangParser::visit_children(CXCursor &cursor, BPFtrace &bpftrace)
 std::unordered_set<std::string> ClangParser::get_incomplete_types(
     const std::string &input,
     std::vector<CXUnsavedFile> &unsaved_files,
-    const std::vector<const char *> &args,
-    const std::unordered_set<std::string> &complete_types)
+    const std::vector<const char *> &args)
 {
   if (input.empty())
     return {};
@@ -565,8 +564,7 @@ std::unordered_set<std::string> ClangParser::get_incomplete_types(
   }
 
   // Don't bail on errors (ie incomplete structs) because our goal is to
-  // enumerate all such errors. Instead, collect error messages for later
-  // analysis.
+  // enumerate all such errors.
   std::vector<std::string> diag_msgs;
   if (!handler.check_diagnostics(input, diag_msgs, false))
     return {};
@@ -576,19 +574,6 @@ std::unordered_set<std::string> ClangParser::get_incomplete_types(
     std::unordered_set<std::string> complete_types;
     std::unordered_set<std::string> incomplete_types;
   } type_data;
-  // Initialize to already defined types
-  type_data.complete_types = complete_types;
-
-  // Search for error messages of the form:
-  //   unknown type name 'type_t'
-  // that imply an unresolved typedef of type_t. This cannot be done below in
-  // clang_visitChildren since clang does not have the unknown type names.
-  for (const auto &msg : diag_msgs)
-  {
-    auto unknown_type = get_unknown_type(msg);
-    if (unknown_type)
-      type_data.incomplete_types.emplace(unknown_type.value());
-  }
 
   CXCursor cursor = handler.get_translation_unit_cursor();
   clang_visitChildren(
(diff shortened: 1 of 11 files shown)


mmisono:
Do you have some benchmark of a type resolution speed?

viktormalik:
> Do you have some benchmark of a type resolution speed? (I wonder if 3 iterations are accepted or not)

There's no benchmark set, the goal is to efficiently minimize the number of sufficient iterations. If the computed number is high, then the resolution will take some time, but I believe that this should not happen often. 3 iterations should still be fast, I guess.

mmisono:
By rebasing this onto master, it is relatively easy to compare processing times like the following:

```
$ cat a.bt
struct f { struct task_struct x; }
BEGIN{exit();}
$ cat b.bt
t:napi:napi_poll { printf("%s\n", args->napi->dev->name); }
$ cat c.bt
t:syscalls:sys_enter_* { @[args->__syscall_nr] = count(); }
i:ms:1 { exit(); }
```


```
% ls *.bt | sort > name
% ls *.bt | sort | parallel "/usr/bin/time -f '%e' sudo ./build_dev_release/src/bpftrace --btf --test semantic --no-warnings {} 2>> dev"
% ls *.bt | sort | parallel "/usr/bin/time -f '%e' sudo ./build_master_release/src/bpftrace --btf --test semantic --no-warnings {} 2>> master"
% paste name dev master
a.bt    0.57    0.81
b.bt    0.68    1.40
c.bt    0.76    1.56
```

So this gives up to about one-second speed up, nice.

mmisono (approved):
I think this does not break existing scripts, so this is OK for me. Is it good to have an environment variable (like BPFTRACE_MAX_STRING) to control the maximum iterations number? I'm not sure.

fbs:
Looks good

> I think this does not break existing scripts, so this is OK for me. Is it good to have an environment variable (like BPFTRACE_MAX_STRING) to control the maximum iterations number? I'm not sure.

I think having an undocumented var for that would be ok. That way we can ask the user to try something in case anyone ever bumps into an issue

viktormalik:
Ok, I added the env variable (separate commit) and rebased.

src/clang_parser.cpp
```diff
@@ -676,29 +661,68 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
...
+    // Resolution of incomplete types must run at least once, maximum should be
+    // the number of levels of nested field accesses for tracepoint args.
+    // The maximum number of iterations can be also controlled by the
+    // BPFTRACE_MAX_TYPE_RES_ITERATIONS env variable (0 is unlimited).
+    unsigned field_lvl = 1;
+    for (auto &probe : *program->probes)
+      if (probe->tp_args_structs_level > (int)field_lvl)
+        field_lvl = probe->tp_args_structs_level;
+
+    unsigned max_iterations = bpftrace.max_type_res_iterations > 0 &&
```
fbs, line 675: Wouldn't std::max(bpftrace.max, field_lvl) be enough here?

viktormalik: Of course it would :-) I just over-complicated it, thanks for noticing

viktormalik:
Looks like there are some issues with downloading packages. Tests need restart.

fbs:
I've triggered the build but CI seems to fail on a missing field now

viktormalik:
Yes, I rebased onto the current master.

PR #1573, opened 2020-10-24
Check string comparison size

mmisono:
In the debug build, the following assertion failure occurs:

```
% sudo ./src/bpftrace -e 'BEGIN /comm == "aaaaaaaaaaaaaaaaaaaaaaaaa"/ {}'
bpftrace: /home/ubuntu/work/bpftrace/bpftrace/src/ast/irbuilderbpf.cpp:683: llvm::Value *bpftrace::ast::IRBuilderBPF::CreateStrncmp(llvm::Value *, llvm::Value *, bpftrace::AddrSpace, std::string, uint64_t, const bpftrace::location &, bool): Assertion `valp->getElementType()->isArrayTy() && valp->getElementType()->getArrayNumElements() >= n && valp->getElementType()->getArrayElementType() == getInt8Ty()' failed.
```

The problem part is https://github.com/iovisor/bpftrace/blob/a03b5c4bbb12a25345a2fdec78374963ff936131/src/ast/irbuilderbpf.cpp#L682
and the reason is that the the size of "comm" is 16, but the size of
string literal is bigger than that.

To fix the problem, check the string size in the semantic analyzer and
error if the size of string literal exeeds the that of the other. Now
the above exmaple becomes

```
% sudo ./src/bpftrace -e 'BEGIN /comm == "aaaaaaaaaaaaaaaaaaaaaaaaa"/ {}'
stdin:1:7-43: ERROR: The right string literal has the longer length of the size of the left (16 < 25)
BEGIN /comm == "aaaaaaaaaaaaaaaaaaaaaaaaa"/ {}
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

##### Checklist

---------------

I wonder if this could cause a regression. I don't think so, but...

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1499,6 +1499,21 @@ void SemanticAnalyser::visit(Binop &binop)
           << " operator can not be used on expressions of types " << lhs << ", "
           << rhs;
     }
+    else if (binop.op == Parser::token::EQ &&
+             ((!binop.left->is_literal && binop.right->is_literal) ||
+              (binop.left->is_literal && !binop.right->is_literal)))
+    {
+      auto *lit = binop.left->is_literal ? binop.left : binop.right;
+      auto *str = lit == binop.left ? binop.right : binop.left;
+      auto lit_len = bpftrace_.get_string_literal(lit).size();
+      auto str_len = str->type.GetNumElements();
+      if (lit_len > str_len)
+      {
+        LOG(WARNING, binop.left->loc + binop.loc + binop.right->loc, out_)
+            << "The literal is longer than the variable string (size="
+            << str_len << "), condition will always be false";
+      }
+    }
   }
 
   bool is_signed = lsign && rsign;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,8 @@ and this project adheres to
   - [#1570](https://github.com/iovisor/bpftrace/pull/1570)
 - Fix invalid cast handling in tuple
   - [#1572](https://github.com/iovisor/bpftrace/pull/1572)
+- Check string comparison size
+  - [#1573](https://github.com/iovisor/bpftrace/pull/1573)
 
 #### Tools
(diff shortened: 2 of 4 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -1499,6 +1499,32 @@ void SemanticAnalyser::visit(Binop &binop)
           << " operator can not be used on expressions of types " << lhs << ", "
           << rhs;
     }
+    else if (binop.op == Parser::token::EQ && !binop.left->is_literal &&
```
fbs, line 1502: I feel like emitting a warning and always return false should be better here. Don't want to block a valid script.

src/ast/semantic_analyser.cpp
```diff
@@ -1499,6 +1499,32 @@ void SemanticAnalyser::visit(Binop &binop)
...
+             binop.right->is_literal)
+    {
+      auto llen = binop.left->type.GetNumElements();
+      auto rlen = bpftrace_.get_string_literal(binop.right).size();
+      if (rlen > llen)
+      {
+        LOG(ERROR, binop.left->loc + binop.loc + binop.right->loc, err_)
```
fbs, line 1509: Dont think we have to keep two cases here. Something simple like this should be enough:

```
Literal is longer than the variable string, condition will always be false
```

fbs (changes requested):

src/ast/semantic_analyser.cpp
mmisono: sure, will fix.

mmisono:
Updated:
- Changed the semantic analyzer so that it emits a warning instead of making an error. Also, make the check simple.
- In the CreateStrcmp, skip comparison if the condition is always false. Also, fix assertion so that it does not assert that case.

fbs (approved):
Looks good.

Having a warning test might be useful tho, just to catch this case in a test somewhere.

mmisono:
Added test for string comparisons

PR #1576, opened 2020-10-26
NFC: Split binop codegen to improve readability

fbs:
split from #1532 as rebasing this every time is a bit error prone


mmisono (approved):
Maybe a silly question, but what does NFC stand for?

fbs:
non functional change in this case :)

PR #1580, opened 2020-10-28
Fix a possible integer overflow

mmisono:
When I tried to use UndefinedSanitizer (by
CXXFLAGS="-fsanitize=undefined"), I found the following error.

```
% sudo ./src/bpftrace -e 'BEGIN{curtask}'
/home/ubuntu/work/bpftrace_master/src/clang_parser.cpp:225:22: runtime error: shift exponent 64 is too large for 32-bit type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/ubuntu/work/bpftrace_master/src/clang_parser.cpp:225:22 in
/home/ubuntu/work/bpftrace_master/src/clang_parser.cpp:225:44: runtime error: signed integer overflow: -2147483648 - 1 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/ubuntu/work/bpftrace_master/src/clang_parser.cpp:225:44 in
```

It warned about integer overflows. This patch fixes this.

##### Checklist

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -221,8 +221,19 @@ static bool getBitfield(CXCursor c, Bitfield &bitfield)
 
   size_t bitfield_offset = clang_Cursor_getOffsetOfField(c) % 8;
   size_t bitfield_bitwidth = clang_getFieldDeclBitWidth(c);
+  size_t bitfield_bitdidth_max = sizeof(uint64_t) * 8;
 
-  bitfield.mask = (1 << bitfield_bitwidth) - 1;
+  if (bitfield_bitwidth > bitfield_bitdidth_max)
+  {
+    LOG(WARNING) << "bitfiled bitwidth " << bitfield_bitwidth
+                 << "is not supporeted."
+                 << " Use bitwidth " << bitfield_bitdidth_max;
+    bitfield_bitwidth = bitfield_bitdidth_max;
+  }
+  if (bitfield_bitwidth == bitfield_bitdidth_max)
+    bitfield.mask = std::numeric_limits<uint64_t>::max();
+  else
+    bitfield.mask = (1ULL << bitfield_bitwidth) - 1;
   bitfield.access_rshift = bitfield_offset;
   // Round up to nearest byte
   bitfield.read_bytes = (bitfield_offset + bitfield_bitwidth + 7) / 8;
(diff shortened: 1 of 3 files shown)


src/clang_parser.cpp
```diff
@@ -222,7 +222,10 @@ static bool getBitfield(CXCursor c, Bitfield &bitfield)
   size_t bitfield_offset = clang_Cursor_getOffsetOfField(c) % 8;
   size_t bitfield_bitwidth = clang_getFieldDeclBitWidth(c);
 
-  bitfield.mask = (1 << bitfield_bitwidth) - 1;
+  if (bitfield_bitwidth == 64)
```
fbs, line 225: Shouldn't this be `>=32` as the warning is:

shift exponent 64 is too large for 32-bit type 'int'

fbs: of maybe 

```
if (width >= sizeof(size_t)) {
  mask = SSIZE_MAX
}
```

mmisono: > Shouldn't this be >=32 as the warning is:

Isn't bitfield_bitwidth >= 32 possible? something like `struct {long x: 48; long y: 16}`

> of maybe
> 
> ```
> if (width >= sizeof(size_t)) {
>   mask = SSIZE_MAX
> }
> ```

ah, this seems better. will fix.

fbs: Shouldn't we make bitfield.mask an uint64_t then, to avoid issues on 32bit?

mmisono: It makes sense. I also noticed that both clang and gcc support `__int128`. So I added the support of it.

mmisono:
hmm.. it seems the error reason for the embedded build is not related to this.

````
/home/runner/work/bpftrace/bpftrace/src/attached_probe.cpp: In member function 'void bpftrace::AttachedProbe::attach_uprobe(bool)':
/home/runner/work/bpftrace/bpftrace/src/attached_probe.cpp:706:51: error: too few arguments to function 'int bpf_attach_uprobe(int, bpf_probe_attach_type, const char*, const char*, uint64_t, pid_t, uint32_t)'
                                         probe_.pid);
                                                   ^
[...]
```

src/clang_parser.cpp
mmisono: I realized that bpftrace does not support `__int128`. So, change it to use `uint64_t`

PR #1586, opened 2020-11-03
codegen: Strip trailing bytes off of str() call

danobi:
It turns out that bpf_probe_read_*str() may copy an additional couple
bytes after the nul terminator. The most likely reason is that the
kernel does the copy in strides and not byte-by-byte for efficiency
reasons. Usually this doesn't matter b/c string handling routines will
stop when they see a nul byte. However, it *does* matter when strings
are used as map keys b/c map keys are byte-by-byte compared and *not*
string aware.

This commit fixes the issue by ensuring that our string copy does not
leave any trailing bytes.

##### Checklist


danobi:
Posted the issue to bpf list; seems like a kernel fix will come soon: https://lore.kernel.org/bpf/20201103182432.yjgfzf25prdhhnmf@ast-mbp.dhcp.thefacebook.com/T/#r81a5969279b0bfeec34ee4799304c68077395651

I'm not sure what the best way to work around this issue is. The "clean" workaround requires a fairly recent kernel. The ugly workaround requires BPFTRACE_STRLEN to be pretty close to a power of 2 (so we can &= 0b111111... and please the verifier).

mmisono:
I wonder what kernel versions are affected by this issue.

danobi:
> I wonder what kernel versions are affected by this issue.

Alexei said it was a recent optimization but when I looked at kernel commit history it seems it's been there for at least a few years, if not since 2012 (before bpf). At least to me it's not obvious when it changed or if it was always subtly buggy.

mmisono:
I tried in Linux 5.9

```c
#include <stdio.h>

void foo(char *a) {
    printf("%s\n", a);
}

int main(){
    char a[] = { '1', '2', '\0', '1' };
    char b[] = { '1', '2', '\0', '2' };
    char c[] = { '1', '2', '\0', '3' };

    foo(&a[0]);
    foo(&b[0]);
    foo(&c[0]);
}
```

```
% sudo ./src/bpftrace -e 'u:./a.out:foo { @[str(arg0)] = count(); }' -c ./a.out
Attaching 1 probe...
12
12
12


@[12]: 1
@[12]: 1
@[12]: 1
```

I hope it doesn't affect too many...

mmisono (approved):
Merging this would depend on how much impact the bug has, but this fix itself looks good.

I checked this on Ubuntu 20.04 (kernel 5.4), but the bug does not exist.

danobi:
If the kernel fix goes in ( https://lore.kernel.org/bpf/cover.1605642949.git.dxu@dxuuu.xyz/T/#t ) I think I'll abandon this PR.

PR #1587, opened 2020-11-03
scripts: Add script to update codegen tests

danobi:
This script helps update codegen changes. Basically all it does is
compile bpftrace against LLVM 8 and run bpftrace_test with the right
invocation.

##### Checklist

### DIFF
--- a/scripts/update_codegen_tests.sh
+++ b/scripts/update_codegen_tests.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Updates codegen tests' expected LLVM IR
+#
+
+set -eu
+
+# Change dir to project root
+cd "$(dirname "${BASH_SOURCE[0]}")"
+cd ..
+
+# Build docker image
+pushd docker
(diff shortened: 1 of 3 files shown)


scripts/update_codegen_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Change dir to project root
+cd "$(dirname "${BASH_SOURCE[0]}")"
+cd ..
+
+# Build docker image
+pushd docker
+docker build                  \
+  --network host              \
+  --build-arg LLVM_VERSION=8  \
```
fbs, line 16: I think i based it on 7, but 8 is identical from what I've seen

danobi: 8 seemed to work OK. Could do 7 but maybe 8 will have a longer shelf life. Up to you

fbs: 8 seem to work fine so lets stick to it :)

PR #1601, opened 2020-11-08
Add basic options and documentations for fuzzing

mmisono:
Several months ago and the last month, I fuzzed bftrace with AFL and found several bugs. This PR adds the documentation about it and some options useful for fuzzing. There are many fuzzers other than AFL and several techniques. I've been trying to fuzz various ways in my spare time and find some useful ways. I plan to update the document later. Also, I hope for more people to try to fuzz and find bugs and make bpftrace safer :)

Please read each commit messages and docs/fuzzing.md for the details.

##### Checklist


src/main.cpp
```diff
@@ -269,6 +269,7 @@ int main(int argc, char *argv[])
   bool force_btf = false;
   bool usdt_file_activation = false;
   int helper_check_level = 0;
+  int test_mode = 0;
```
danobi, line 274: I think an enum instead of an int would make it more readable.

src/ast/callback_visitor.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace bpftrace {
+namespace ast {
+
+using callback = std::function<void(Node *, void *)>;
+
+class CallbackVisitor : public Visitor
+{
+public:
+  explicit CallbackVisitor(callback func, void *data) : func_(func), data_(data)
```
danobi, line 14: I think `explicit` is only necessary if you have a single arg constructor and you want to prevent implicit conversions.

src/ast/callback_visitor.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "ast.h"
+#include <functional>
+
+namespace bpftrace {
+namespace ast {
+
+using callback = std::function<void(Node *, void *)>;
```
danobi, line 9: Absent a strong need to have C function pointers as callbacks, I'd suggest:
```
 using callback = std::function<void(Node *)>; 
```
and letting callers pass a lambda that captures whatever it needs by reference.

src/main.cpp
```diff
@@ -732,6 +738,23 @@ int main(int argc, char *argv[])
     std::cout << std::endl;
   }
 
+  // Count AST nodes
+  uint64_t node_count = 0;
+  ast::CallbackVisitor counter([](ast::Node* node __attribute__((unused)),
+                                  void* data) { *((uint64_t*)data) += 1; },
+                               &node_count);
```
danobi, line 745: See previous comment. This could just be:
```c++
ast::CallbackVisitor counter([&](ast::Node* node __attribute__((unused))) { data += 1; });
```

docs/fuzzing.md
```diff
@@ -0,0 +1,114 @@
...
+     BPFTRACE_NODE_MAX=200 \
+     taskset -c ${CPU} \
+     $FUZZER -M 0 -m none -i ./input -o ./output -t 3000 -- \
+     ./src/bpftrace --test codegen --no-warnings @@
+```
+
+I describe several important things:
+
+- `-i` is the input directory, and `-o` is the output directory. In the input directory, you need to put
+  something to start fuzzing. The most simple example is `echo a > input/a`. More sofisticated inputs can
```
danobi, line 90: Typo for "sophisticated"

docs/fuzzing.md
```diff
@@ -0,0 +1,114 @@
...
+  be created by extracting the bpftrace program from the source code directory (especially fromt tests
```
danobi, line 91: fromt -> from

danobi (commented):
Thanks for the detailed write up!

src/ast/callback_visitor.h
mmisono: ok, remove it.

src/ast/callback_visitor.h
mmisono: ah, it looks good. will fix.

src/ast/callback_visitor.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace bpftrace {
+namespace ast {
+
+using callback = std::function<void(Node *)>;
+
+class CallbackVisitor : public Visitor
+{
+public:
+  CallbackVisitor(callback func) : func_(func)
```
danobi, line 14: Actually, now I think you'll want the `explicit` :p

mmisono: oops!

danobi (approved):

fbs (approved):
Nice

fbs:
Maybe we can try https://github.com/google/oss-fuzz too

PR #1602, opened 2020-11-10
Use pid relative to bpftrace's pidns when resolving user symbols

danobi:
We currently pass the root namespace pid from the bpf prog to userspace.
bpftrace then takes the pid and reads from /proc/pid/exe to symbolize.

When bpftrace is run inside a container, the root namespace pid is
meaningless. We want to always use the pid relative to the pid namespace
bpftrace is in.

This commit has bpftrace generate progs that pass the pid relative to
bpftrace's pid namespace (whenever the helper is available).

##### Checklist


fbs:
Looks good. Might be good to add documentation on how to use this in the guide. Something like `nsenter -t nsref -p bpftrace ..` i guess?

danobi:
> Looks good. Might be good to add documentation on how to use this in the guide. Something like nsenter -t nsref -p bpftrace .. i guess?

Does this need documentation? Things should "just work" now. I don't think this is really a feature, more of a bug fix.

fbs:
> > Looks good. Might be good to add documentation on how to use this in the guide. Something like nsenter -t nsref -p bpftrace .. i guess?
> 
> Does this need documentation? Things should "just work" now. I don't think this is really a feature, more of a bug fix.

True

 Stating that `pid/tid` are relative to the bpftrace pidns would still be nice but im not sure that really fits the refguide currently.

danobi:
> Stating that pid/tid are relative to the bpftrace pidns would still be nice but im not sure that really fits the refguide currently.

Isn't `pid`/`tid` relative to the root pidns currently? I decided not to change that in this PR to avoid any compat issues. The stuff I changed here is not user visible.

fbs:
> > Stating that pid/tid are relative to the bpftrace pidns would still be nice but im not sure that really fits the refguide currently.
> 
> Isn't `pid`/`tid` relative to the root pidns currently? I decided not to change that in this PR to avoid any compat issues. The stuff I changed here is not user visible (other than symbolization being fixed)

Oh woops I totally misread that.Thought it fixed both (usymbols and builtins). NVM me

javierhonduco (approved):
LGTM 🚢

danobi:
Turns out the kernel's API is designed very strangely. So this PR doesn't actually work (it actually regresses symbolization).

I'm asking some questions here: https://lore.kernel.org/bpf/C71MVCBQMCPF.1CCKLFRTGYD0D@maharaja/T/#u

danobi:
Can't remember what this does anymore. Looks like it didn't work anyways.

Issue #1603, opened 2020-11-10
Improve AST / allow mutation


fbs:
I think we can simplify our big AST passes (esp semantic) a bit if we make it easier to build new passes for analysis or mutation. E.g. we could have a pass that replaces all positional parameters with their actual value (and cases like `str($1+1)`) with their constant value. This should not only improve the semantic pass, but also make codegen better as we currently emit such cases onto the stack and probe read them. There are more passes like these we could add.

Adding new passes is currently a bit tricky. You have to copy a lot of boilerplate to visit the tree add the pass into main and to debug the pass you might want to add the printer before/after. This also means we have to extend every pass if we ever add an AST node.

What I want to add:

- A Visitor base class that handles the tree iteration. Allowing us to write a pass like:

```
class MyThing : public ASTVisitor {
   MyThing(Node * root, args..) :  ASTVisitor(root);

   void visit(Integer &n) { dosomthing(); ASTVistor::visit(n); }
}
```
- A mutator base class which makes it easy to return a modified copy of the tree, similar to the ASTVisitor
- A pass manager to make it easier to add/disable passes and collect info(runtime, new tree etc).

And related:

- Tree destructor to clean up the old copy of the ast.
- Disable the copy constructors to avoid accidental shallow copies of pointers

fbs:
Got a working thing [here](https://github.com/fbs/bpftrace/commits/ast_walker) but I'm not that happy with how the double dispatch makes it hard to customize the return value (would need copies of accept() in the node classes), which is now handled by helpers in the astmutator class (Return, Value). I'm currentl looking at dropping double dispatch and implementing our own vtable dispatch based on type id. The added benefit is that we can have a pattern like:

```
visit(call c)
{
    for (arg in c.args)
        visitor(arg)
}

visitor(node c)
{
    return vtable(c)
}
```

As visitor is visited for every node we can also override visitor to run a single function on each node, similar to what #1601 does now.

brendangregg:
Maybe this will fix ttysnoop.bt, and another tool I'm writing that is supposed to accept an optional filename argument; e.g.:

```if ($2 == 0 || str($2) == str($filename)) {```

fbs:
iirc we did some fixes to allow `str($1 + 3)` again but i see it now fails on the ==0 check.

We should be able to make it work with the pos param removing pass and maybe another. That way we can write it as  `if($1)` and keep `if($1 == 0)` supported for legacy purposes

PR #1607, opened 2020-11-11
RFC: improved AST visiting/manipulation

fbs:
This fixes most of #1603, see that issue for some additional context.

The first two commits deduplicate the existing AST passes, instead of having to re-implement the tree traversing every time we can just inherit from a class that does that and call the parent function when needed. This doesn't do much for complex passes like semantic and codegen, but for simpler ones that don't touch all nodes (like tracepoint) it removes quite a bit of logic.

Then two cleanup commits, one to remove all the constructors without location info, as we use location everywhere we don't need those anymore. And one to dedupe the `accept()` function in every node, making it easier to modify if needed.

The next 5 are destructors for ast nodes based on #1055 and related test changes and the definition of copy constructors. All copies are shallow, settings all pointers to nullptr to avoid ownership issues. And hiding the copy constructor, forcing the use of the `clone()` method when needed to avoid accidental copies, which can corrupt the tree (due to the nullptr).

`AST: Implement custom dispatcher for node walking ` is the interesting part imo. Instead of using double dispatch it implements a custom dispatcher which  makes it eaiser to use custom return types and to write callback functions. The commit message has all the details.

The last two rewrite the CallbackVisitor to the new ASTNodeVisitor class and a basic implementation of a postional parameter removing class, which should simplify the semantic phase.

----

The whole things is quite big, a lot of test refactoring and things that add to the total lines. When/if we decide to merge this I'll split it into parts to make it a bit easier.

Opening this as RFC to get some feedback on the idea itself, i.e. the new dispatching and the idea of adding more/simpler passes.


mmisono:
Just skimmed through the commits, but implementing CallbackVisitor was a bit of pain, so the new one looks nice.

fbs:
I've fixed some bugs and rebased it. 

The positional param pass should be mostly done now. The unit tests are broken ofcourse but it passes the runtime test suite. Some stuff is still missing but it should be good enough to show the concept.

fbs:
@danobi how do you feel about this one? I want to do some tests to see if i can get the dispatch overhead down a bit but I don't think it is an issue for the amount of passes we will have anyway.

danobi (commented):
I like the new API. I'm a little concerned about the complexity of the internals but I think I understand it. I think it'd also be worth writing some short docs on how to extend the API and why things are designed the way they are. Maybe inline where `NodeDispatcher` is defined.

fbs:
> I like the new API. I'm a little concerned about the complexity of the internals but I think I understand it. I think it'd also be worth writing some short docs on how to extend the API and why things are designed the way they are. Maybe inline where `NodeDispatcher` is defined.

Yeah I'll add some inline comments and things to make it a bit easier to follow.

Thanks

fbs:
After messing with it a bit more I think its best to keep both versions. Double dispatch is faster and if we dispatch trough `Visit()` instead of calling `node->accept()` directly we also get the generic node to dispatch through.

For mutators we  can keep the vtable based dispatch, its a bit trickier and higher overhead but more flexible.

I've also added a basic pass manager. The idea is that every pass adds a function that indicates how to run the pass. This will make it easier to do diagnostics between passes (e.g. print ast, print time it took) and should simplify main. All config is stored in a PassCtx object, its a bit ugly but works for now.  Too many options in there indicates that we have to many special passes.

Hope to get it finished in the next few days.

fbs:
#1655 does most of this. The posparam transformer will be part of a follow up once that one lands

PR #1609, opened 2020-11-12
0.11.3 release

fbs:
0.11.3 release to support bcc 0.17.

The changelog item isn't really a user impact thing, but it details why we release so I figured its good to have in


fbs:
@vincentbernat  can you give this a go? :)

vincentbernat:
Works fine for me. Tested on Debian unstable with the packaging rules from Debian.

danobi (approved):

PR #1616, opened 2020-11-13
add -q option for quiet

### LINKED ISSUE #446: --silent (or --quiet)
This option would disable any output to stdout which doesn't come from the bpftrace program being executed. I think the only message that would be suppressed right now is `Attaching X probes...`, since we output other warnings to stderr. The goal is to make it easier for external applications to parse bpftrace's output.

graphaelli:
Add -q option for quiet.  

Closes #446

##### Checklist


danobi:
Seems reasonable to me. Wish we didn't have a bunch of easily forgettable if statements, though.

Wonder if we can improve this in the future with the LOG() api.

graphaelli:
thanks @danobi, what happens now with this pr?

fbs:
> thanks @danobi, what happens now with this pr?

It gets merged :)

PR #1617, opened 2020-11-14
Fuzzing improvements

mmisono:
This PR includes three changes. Basically, add a main function for fuzzing for convenience.

1
```
Although the current bpftrace can be fuzzed with AFL, there are several
possible optimizations for fuzzing. This add -DBUILD_FUZZ CMake options
and it defined FUZZ. We can use it in the source code to optimize bpftrace
for fuzzing. Of course, such bpftrace would not work for tracing.
```

2
```
Add a simple main for fuzzing. This would make it easier to optimize some
processing for fuzzing and integrate with another fuzzer.
There are several duplicated lines in the original main function. I'll
plan to unify them later.
```

3
```
I realized that reading debug/tracing/available_filter_functions takes
time, and that slow down fuzzing severely. Especially, BTF's constructor
always call the function no matter if it does not use the function. So this
patches changes it, and now BTF's constructor does not call the
function, and always report tracee function is available. It should be
OK since during fuzzing we don't load the program in the kernel and
trace. Now average AFL's exec speed becomes from 1 to up to 20 in my machine.
```

I think we can add more several things to improve fuzzing, but I'll make this PR for the time being to make review easy.

##### Checklist


src/main.cpp
```diff
@@ -875,3 +877,137 @@ int main(int argc, char *argv[])
...
+    llvm.optimize();
+    bpforc = llvm.emit();
+  }
+  catch (const std::system_error& ex)
+  {
+    return 1;
+  }
+  catch (const std::exception& ex)
+  {
+    // failed to compile
```
mmisono, line 1003: I'm not sure we should abort here, and let the fuzzer notify us. Does invalid IR result in an exception?

fbs: llvm doesnt really care. Unless you really screwed up, in that case it segvs or goes oom. Thats why we have the codegen-validator check to find our IR mistakes

src/main.cpp
```diff
@@ -875,3 +877,137 @@ int main(int argc, char *argv[])
 
   return 0;
 }
+
+#else
+
+#if !defined(NODE_MAX)
```
fbs, line 883: I think i'd prefer a special fuzz only binary instead of doing ifdefs everywhere. That way you can just disable BTF support completely instead of what happens now. 

So in `src/CMakelist.txt`

```
add_executable(bpftrace_fuzz, ..files.. fuzz_main.cpp ...)
```

mmisono: I think we also want to fuzz BTF related part, but the separate a binary seems good to me. I'll try it.

src/main.cpp
mmisono: I see. then I leave it as is.

src/main.cpp
mmisono: I added fuzz_main.cpp, and now when building fuzz binary that name is bpftrace_fuzz. I'll rebase commits if it's desirable.

mmisono:
Ah, before merging I need to update the document. I'll do it later.

mmisono:
Updated.

src/CMakeLists.txt
```diff
@@ -2,7 +2,15 @@ if(HAVE_BFD_DISASM)
   set(BFD_DISASM_SRC bfd-disasm.cpp)
 endif()
 
-add_executable(bpftrace
+if(BUILD_FUZZ)
+  set(MAIN_SRC fuzz_main.cpp)
+  set(BPFTRACE bpftrace_fuzz)
+else()
+  set(MAIN_SRC main.cpp)
+  set(BPFTRACE bpftrace)
```
fbs, line 10: nit: might be good to unset (`set(MAIN_SRC "") iirc`) these at the end to avoid them leaking into cmake state.

fbs (approved):

PR #1625, opened 2020-11-16
Ast cleanup

fbs:
Split from #1607 to make it less huge. This,

- removes constructors that don't have a location. Location is required for error messages, having the old constructors make it possible to forget to specify a location.
- deduplicates some visitor code by implementing a base visitor that just walks the tree and the moving exisiting visitors to that. That way its possible to just override the nodes you want instead of having to duplicate the whole tree.


danobi (approved):
Awesome, thanks for the improvements

PR #1626, opened 2020-11-16
Update docs

danobi:
Small bug fixes to documentation. Also document the missing kptr() and uptr()
functions. See individual commits for more details.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1242,8 +1242,8 @@ hi
 ^C
 ```
 
-The namespace of the probe is deduced automatically. If the binary `/root/tick` contained multiple probes 
-with the name `loop` (e.g. `tick:loop` and `tock:loop`), no probe would be attached. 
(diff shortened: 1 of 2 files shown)


docs/reference_guide.md
```diff
@@ -2821,6 +2829,39 @@ Attaching 1 probe...
...
 ```
 
+## 26. `uptr()`: Annotate userspace pointer
+
+Syntax:
+- `uptr(void *p)`
+
+Annotate `p` as a pointer belonging to userspace address space.
+
+This instructs bpftrace to generate calls to `bpf_probe_read_user*()` when
```
mmisono, line 2839: I wonder if we should write about internal things (`bpf_probe_read_user`). Maybe just writing bpftrace infers pointer address space but sometimes fails so in this case use it is enough?

danobi: Yeah, that makes sense

mmisono (approved):

PR #1627, opened 2020-11-17
Ast mem leaks

fbs:
Also split from #1607 

These fix some of the memory leaks we have, mostly a rebase of #1055. 

To avoid memory leaks and ownership issue this also defines shallow copy constructors for each node and marks them as private. This makes  it impossible to accidentally copy the object and have two objects point to the same  memory, which can lead to double frees. If you really want a shallow copy you can use the `copy()` method. A deep copy is possible, but not implemented as I didn't need it.

##### Checklist
 
### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -43,18 +43,6 @@ MAKE_ACCEPT(Program)
 
 #undef MAKE_ACCEPT
 
-Node::Node() : loc(location())
-{
-}
-
-Node::Node(location loc) : loc(loc)
-{
-}
-
-Expression::Expression(location loc) : Node(loc)
-{
-}
-
 Integer::Integer(long n, location loc) : Expression(loc), n(n)
 {
   is_literal = true;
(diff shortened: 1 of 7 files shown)


mmisono:
```sh
[ RUN      ] bpftrace_btf.add_probes_kfunc
munmap_chunk(): invalid pointer
zsh: abort (core dumped)  ./tests/bpftrace_test --gtest_filter="*btf*"
```

I guess semantic analyzer tests also need to be updated like test/ast.cpp.

-------------

LeakSanitizer less complains about leaks, nice. Maybe you know, but for example, `sudo ./src/bpftrace ../tools/runqlat.bt` still have memory leaks. Is it a different issue?

mmisono:
really sorry, I mistakenly thought it was my fuzzing PR...

fbs:
> ```shell
> [ RUN      ] bpftrace_btf.add_probes_kfunc
> munmap_chunk(): invalid pointer
> zsh: abort (core dumped)  ./tests/bpftrace_test --gtest_filter="*btf*"
> ```
> 
> I guess semantic analyzer tests also need to be updated like test/ast.cpp.
> 
> LeakSanitizer less complains about leaks, nice. Maybe you know, but for example, `sudo ./src/bpftrace ../tools/runqlat.bt` still have memory leaks. Is it a different issue?

ah woops, forgot to test the btf path.

PR #1628, opened 2020-11-17
bpftrace: Fix bitfield access for Big endian

sumanthkorikkar:
This resolves bitfield access for big endian systems. 
Also added testcases to cover few more cases.

Thanks

##### Checklist

- [x ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x ] The new behaviour is covered by tests

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1611,7 +1611,14 @@ void CodegenLLVM::visit(FieldAccess &acc)
         raw = b_.CreateLoad(dst);
         b_.CreateLifetimeEnd(dst);
       }
-      Value *shifted = b_.CreateLShr(raw, field.bitfield.access_rshift);
+      size_t rshiftbits;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+      rshiftbits = field.bitfield.access_rshift;
+#else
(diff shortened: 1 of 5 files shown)


src/ast/codegen_llvm.cpp
```diff
@@ -1611,7 +1611,16 @@ void CodegenLLVM::visit(FieldAccess &acc)
         raw = b_.CreateLoad(dst);
         b_.CreateLifetimeEnd(dst);
       }
-      Value *shifted = b_.CreateLShr(raw, field.bitfield.access_rshift);
+      size_t rshiftbits;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+      rshiftbits = field.bitfield.access_rshift;
+#else
+      // Number of bytes to shift
```
fbs, line 1618:
```suggestion
      // Number of bits to shift
```

fbs: but i'd leave the comment out, or explain the logic behind how this shifting works

sumanthkorikkar: Ok, Sure. I will explain with an example in the commit message then. Thanks.

sumanthkorikkar: Added in the commit message. Thanks

fbs (approved):

PR #1629, opened 2020-11-17
bpftrace: Fix buf builtin testcases for big endian

sumanthkorikkar:
- buf builtin displays hex bytes according to the byte order of the hosts.
  Hence fix the testcase
- Also add additional cases for buf builtin

Thanks

##### Checklist

- [x ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x ] The new behaviour is covered by tests

### DIFF
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -50,9 +50,16 @@ EXPECT P: /*.
 TIMEOUT 5
 
 NAME buf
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -104,6 +104,8 @@ and this project adheres to
   - [#1580](https://github.com/iovisor/bpftrace/pull/1580)
 - Printing of small integers with `printf`
   - [#1532](https://github.com/iovisor/bpftrace/pull/1532)
+- Fix buf builtin testcases for big endian
```
fbs, line 107: We try to keep the changelog for end users, so internal test changes are not relevant

fbs (changes requested):
need to remove the changelog entry but except for that lgmt

CHANGELOG.md
sumanthkorikkar: Ok. Will remove the changelog then. Thanks

fbs (approved):

PR #1631, opened 2020-11-17
Revert "Ast mem leaks"

mmisono:
Reverts iovisor/bpftrace#1627


mmisono:
Really sorry, I accidentally cricked.

PR #1621, opened 2020-11-15
Error if using negative length in str() and buf()

mmisono:
This disables a negative length in `str()` and `buf`(), such as `buf("aaa", -1)`.

##### Checklist

- [ ] Language changes are updated in `docs/reference_guide.md`
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x] The new behaviour is covered by tests

Issue #1618, opened 2020-11-14
Can't use IS_MODULE in script (or flex scanner push-back overflow)


mmisono:
### What reproduces the bug?
I've trying to fuzz using libfuzzer and found an interesting case.

```
% sudo ./src/bpftrace -e "BEGIN { IS_MODULE }" 
flex scanner push-back overflow
```

Only `IS_MODULE` has this problem.  After some investigation, apparently, the following part is related.

https://github.com/iovisor/bpftrace/blob/330953c11190c23a12f98d374d1b5eb6d7aad1fe/src/lexer.l#L182-L199

and in this case, `yytext` is `IS_MODULE` and `s` is `(option)__is_defined(option##_MODULE)`.  I'm not sure why the macro is defined nor how to fix this.

### bpftrace --info
```
System
  OS: Linux 5.9.0+ #24 SMP Tue Nov 10 16:05:24 JST 2020
  Arch: x86_64

Build
  version: v0.11.0-240-g3309
  LLVM: 11.0.1
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: no
  bcc bpf_attach_uprobe refcount: no
  libbpf: yes
  libbpf btf dump: yes
  libbpf btf dump type decl: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf (depends on Build:libbpf): yes
  map batch (depends on Build:libbpf): yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes

```

mmisono:
As the comment says, macros with operators go into infinite loops. I'm not sure why the macro is defined (maybe from system headers?), but if the macro has an operator, shouldn't we error it? since bpftrace does not support an macro operation if I understand correctly.

PR #1633, opened 2020-11-17
Add support of libFuzzer

mmisono:
[LibFuzzer](https://llvm.org/docs/LibFuzzer.html) is a coverage-guided fuzzer developed with llvm/clang.
LibFuzzer's main target is a function, and the biggest difference
between AFL and the libfuzzer is that the former re-executes the program
after each execution (actually it does fork), while the libFuzzer keeps
calling the target function in the main loop.

The main target of the libfuzzer is a function, but with a few
changes, it can be fuzzed from the beginning of the program, like the
AFL.

Because the current bpftrace is known to have memory leaks and there
might be probably some global states, it may not be suitable for
libfuzzer. But libFuzzer already found several bugs (#1618, #1621). Also,
it seems oss-fuzz expects the libfuzzer interface, so support it is a good thing.

##### Checklist


src/ast/field_analyser.cpp
```diff
@@ -294,7 +294,11 @@ int FieldAnalyser::analyse()
   std::string errors = err_.str();
   if (!errors.empty())
   {
+#ifdef FUZZ
+    (void)out_;
```
fbs, line 298: Both the field analyser and semantic allow setting `out` in the constructor. Setting it to `/dev/null` from fuzz_main.cpp seems cleaner than doing ifdefs everywhere

mmisono: I didn't notice it. Thanks, fixed it.

src/log.cpp
```diff
@@ -196,7 +196,16 @@ LogStream::LogStream(const std::string& file,
 
 LogStream::~LogStream()
 {
+#ifdef FUZZ
+  // When fuzzing, we don't want to output error messages. However, some
```
danobi, line 200: Why do we need to suppress error messages? Is it a speed thing? I thought fuzzing engines look for program crashes -- the output shouldn't matter to them.

mmisono: libfuzzer prints the program's stdout/stderr to a terminal by default, and it makes reading libfuzzer's output almost impossible. We can change the behavior `-close_fd_mask` flags of libfuzzer( https://llvm.org/docs/LibFuzzer.html). But I think suppressing in bpftrace is better. And maybe another fuzzer behaves like libfuzzer.

docs/fuzzing.md
```diff
@@ -104,6 +104,29 @@ I describe several important things:
...
+and bpftrace can be fuzzed with it.
+
+### Compile with libFuzzer
+To use the libFuzzer, use clang and compile it as follows.
+
+```
+CC=clang-10 CXX=clang++-10 cmake .. -DBUILD_ASAN=1 -DBUILD_FUZZ=1 -DUSE_LIBFUZZER=1 -DBUILD_TESTING=0
+```
+
+`-DBUILD_FUZZ=1` and `-DUSE_LIBFUZZE=1` is necessary.
```
mmisono, line 118: ah, I'll fix typo

src/log.cpp
danobi: Ah, got it.

CMakeLists.txt
```diff
@@ -42,14 +43,23 @@ endif()
...
-    message(STATUS "Fuzzing seantic analyzer")
-    add_compile_options("-DTEST_SEMANTIC")
-elseif(FUZZ_TARGET STREQUAL "codegen")
-    message(STATUS "Fuzzing codegen")
-else()
-    message(FATAL_ERROR "Unsupported FUZZ_TARGET")
+
+    if(FUZZ_TARGET STREQUAL "semantic")
+        message(STATUS "Fuzzing seantic analyzer")
+        add_compile_options("-DTEST_SEMANTIC")
```
danobi, line 49: I think `target_compile_options()` is better b/c it doesn't proliferate the macro everywhere. Also probably more portable.

fbs (approved):

CMakeLists.txt
mmisono: I didn't know that option. Now I move to fuzz related settings to src/CMakeList.txt and use it.

mmisono: ah, target_compile_definitions looks better.

mmisono: hmm.. I tried to use target_compile_definitions but it does not work the same as the target_compile_options (encountered linking error with libfuzzer.a). CMake is difficult... I'll keep the current code.

PR #1635, opened 2020-11-18
Ast mem leaks

fbs:
#1627 was accidentally merged

----

These fix some of the memory leaks we have, mostly a rebase of #1055.

To avoid memory leaks and ownership issue this also defines shallow copy constructors for each node and marks them as private. This makes it impossible to accidentally copy the object and have two objects point to the same memory, which can lead to double frees. If you really want a shallow copy you can use the copy() method. A deep copy is possible, but not implemented as I didn't need it.

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -605,5 +605,17 @@ If::If(const If &other) : Statement(other)
 {
 }
 
+AssignVarStatement::AssignVarStatement(const AssignVarStatement &other)
+    : Statement(other)
+{
+  compound = other.compound;
+};
+
+AssignMapStatement::AssignMapStatement(const AssignMapStatement &other)
+    : Statement(other)
+{
+  compound = other.compound;
+};
+
 } // namespace ast
 } // namespace bpftrace
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -43,18 +43,6 @@ MAKE_ACCEPT(Program)
 
 #undef MAKE_ACCEPT
 
-Node::Node() : loc(location())
-{
-}
-
-Node::Node(location loc) : loc(loc)
-{
-}
-
-Expression::Expression(location loc) : Node(loc)
-{
-}
-
 Integer::Integer(long n, location loc) : Expression(loc), n(n)
 {
   is_literal = true;
@@ -105,7 +93,6 @@ Call::Call(const std::string &func, ExpressionList *vargs, location loc)
 {
 }
 
-
 Map::Map(const std::string &ident, location loc)
     : Expression(loc), ident(ident), vargs(nullptr)
 {
@@ -122,13 +109,11 @@ Map::Map(const std::string &ident, ExpressionList *vargs, location loc)
   }
 }
 
-
 Variable::Variable(const std::string &ident, location loc)
     : Expression(loc), ident(ident)
 {
   is_variable = true;
 }
-
 
 Binop::Binop(Expression *left, int op, Expression *right, location loc)
     : Expression(loc), left(left), right(right), op(op)
@@ -194,10 +179,6 @@ Tuple::Tuple(ExpressionList *elems, location loc)
 {
 }
 
-
-Statement::Statement(location loc) : Node(loc)
-{
-}
 
 ExprStatement::ExprStatement(Expression *expr, location loc)
     : Statement(loc), expr(expr)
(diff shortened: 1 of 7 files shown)


fbs:
> ```shell
> [ RUN      ] bpftrace_btf.add_probes_kfunc
> munmap_chunk(): invalid pointer
> zsh: abort (core dumped)  ./tests/bpftrace_test --gtest_filter="*btf*"
> ```
> 
> I guess semantic analyzer tests also need to be updated like test/ast.cpp.
> 
> LeakSanitizer less complains about leaks, nice. Maybe you know, but for example, `sudo ./src/bpftrace ../tools/runqlat.bt` still have memory leaks. Is it a different issue?

fbs:
ill look into the mem leak and gotta update one of the commit messages still.

fbs:
@mmisono which leak did you see? All i see is sizedtype related leaks

mmisono:
@fbs ah, yes, all leaks seem to be related to sizedtype and not mem ast mem leaks.

mmisono:
with #1636, now only synsnoop.bt complains in tools. It seems a different issue.

```
% sudo ./build_llvm10/src/bpftrace ./tools/syncsnoop.bt --test codegen

==1067785==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 96 byte(s) in 6 object(s) allocated from:
    #0 0x54d8b9 in realloc (/disk/work/bpftrace_master/build_llvm10/src/bpftrace+0x54d8b9)
    #1 0x7fffed65ad5e in glob_in_dir /build/glibc-ZN95T4/glibc-2.31/posix/../posix/glob.c:1410:4
    #2 0x7fffed6bb375 in glob64 /build/glibc-ZN95T4/glibc-2.31/posix/../posix/glob.c:985:13
    #3 0x4f8328 in glob (/disk/work/bpftrace_master/build_llvm10/src/bpftrace+0x4f8328)
    #4 0x7d44b6 in bpftrace::TracepointFormatParser::parse(bpftrace::ast::Program*, bpftrace::BPFtrace&) /home/shinalab/work/bpftrace_master/src/tracepoint_format_parser.cpp:49:21
    #5 0x787095 in main /home/shinalab/work/bpftrace_master/src/main.cpp:676:7
    #6 0x7fffed57e0b2 in __libc_start_main /build/glibc-ZN95T4/glibc-2.31/csu/../csu/libc-start.c:308:16

Indirect leak of 450 byte(s) in 6 object(s) allocated from:
    #0 0x54d59d in malloc (/disk/work/bpftrace_master/build_llvm10/src/bpftrace+0x54d59d)
    #1 0x7fffed65a74d in prefix_array /build/glibc-ZN95T4/glibc-2.31/posix/../posix/glob.c:1221:19

SUMMARY: AddressSanitizer: 546 byte(s) leaked in 12 allocation(s).
```

fbs:
ah nice, ready for review then :)

fbs:
Fixed a few small issues in the copy constructors

src/ast/ast.h
```diff
@@ -16,6 +16,12 @@ class Visitor;
 
 #define DEFINE_ACCEPT void accept(Visitor &v) override;
 
+#define DEFINE_CLONE(T)                                                        \
+  T *clone()                                                                   \
```
mmisono, line 20: I feel `clone()` indicates deepcopy, but may it's just me.

fbs: What would you prefer? I wanted to go with copy and deepcopy, but if you think shallowcopy and deepcopy is better im happy with that too

mmisono: I think copy and deepcopy is fine.

mmisono: but this function does not copy pointers and all pointers become nullptr. So copy_without_pointers()? it's a bit longer though...

fbs: Maybe:

- copy(), shallow copy, copies everything by value and points to the same subtree as  orig did.
- deepcopy(), creates a copy of the whole subtree.  
- leafcopy(), like shallow copy but not copying the pointers to other nodes, effectively turning this into a leaf node.

fbs: Or nullcopy

mmisono: +1 for leafcopy

fbs: done :)

src/ast/ast.cpp
```diff
@@ -388,5 +369,80 @@ void Probe::set_index(int index) {
...
+}
+
+Call::Call(const Call &other) : Expression(other)
+{
+  func = other.func;
+}
+
+Binop::Binop(const Binop &other) : Expression(other)
+{
+  left = nullptr;
```
mmisono, line 550: just wonder why the only binop assign nullptr explicitly?

mmisono (approved):
Looks good to me. At least, ast memleks of tools/*.bt are gone.

src/ast/ast.cpp
fbs: left over from when the ast didnt have a default value yet, ill remove these

PR #1636, opened 2020-11-18
types: (hacky) fix sizedtype memleak with shared_ptr

fbs:
Switching to shared_pointers is a bit of a hack and not really doing
proper memory management. We're still risk accidentally modifying the
object we're pointing too and changing the containing type of an
array/pointer. But it shouldn't be any worse than it is now an it
reduces the amount of leaks significantly, which make using the ASAN
output a lot easier.

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -101,8 +101,8 @@ public:
 private:
   size_t size_; // in bytes
   bool is_signed_ = false;
-  SizedType *element_type_ = nullptr; // for "container" and pointer
-                                      // (like) types
+  std::shared_ptr<SizedType> element_type_; // for "container" and pointer
+                                            // (like) types
   size_t num_elements_;               // for array like types
   std::string name_; // name of this type, for named types like struct
   bool ctx_ = false; // Is bpf program context
@@ -204,13 +204,13 @@ public:
   const SizedType *GetElementTy() const
   {
     assert(IsArrayTy());
-    return element_type_;
+    return element_type_.get();
   }
 
   const SizedType *GetPointeeTy() const
   {
     assert(IsPtrTy());
-    return element_type_;
+    return element_type_.get();
   }
 
   bool IsBoolTy() const
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -328,7 +328,7 @@ SizedType CreateArray(size_t num_elements, const SizedType &element_type)
 {
   auto ty = SizedType(Type::array, num_elements);
   ty.num_elements_ = num_elements;
-  ty.element_type_ = new SizedType(element_type);
+  ty.element_type_ = std::make_shared<SizedType>(element_type);
   return ty;
 }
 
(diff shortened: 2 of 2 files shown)


src/types.h
```diff
@@ -101,8 +101,8 @@ class SizedType
 private:
   size_t size_; // in bytes
   bool is_signed_ = false;
-  SizedType *element_type_ = nullptr; // for "container" and pointer
-                                      // (like) types
+  std::shared_ptr<SizedType> element_type_; // for "container" and pointer
```
danobi, line 104: Why not unique_ptr? We're not sharing ownership anywhere.

danobi (commented):
I actually don't think this is a hack.

SizedType semantically "owns" `element_type_`. And it's fine to hand out `const SizedType *` pointers b/c they don't imply ownership.

src/types.h
fbs: We do a lot of copies of types everywhere (see snippet below). For most types it doesn't really matter but for array/pointer types we need to recursively copy the type. The hacky fix for now is to use a sharedptr to manage the ownership mess for us. 



```
/vagrant/src/ast/semantic_analyser.cpp: In member function ‘virtual void bpftrace::ast::SemanticAnalyser::visit(bpftrace::ast::AssignVarStatement&)’:
/vagrant/src/ast/semantic_analyser.cpp:2099:43: error: use of deleted function ‘bpftrace::SizedType& bpftrace::SizedType::operator=(const bpftrace::SizedType&)’
 2099 |   assignment.var->type = assignment.expr->type;
      |                                           ^~~~
/vagrant/src/ast/semantic_analyser.cpp:2114:43: error: use of deleted function ‘bpftrace::SizedType& bpftrace::SizedType::operator=(const bpftrace::SizedType&)’
 2114 |         search->second = assignment.expr->type;
      |                                           ^~~~
/vagrant/src/ast/semantic_analyser.cpp:2128:49: error: use of deleted function ‘bpftrace::SizedType& bpftrace::SizedType::operator=(const bpftrace::SizedType&)’
 2128 |     variable_val_[var_ident] = assignment.expr->type;
      |                                                 ^~~~
/vagrant/src/ast/semantic_analyser.cpp:2129:45: error: use of deleted function ‘bpftrace::SizedType& bpftrace::SizedType::operator=(const bpftrace::SizedType&)’
 2129 |     assignment.var->type = assignment.expr->type;
      |                                             ^~~~
/vagrant/src/ast/semantic_analyser.cpp:2175:23: error: use of deleted function ‘bpftrace::SizedType::SizedType(const bpftrace::SizedType&)’
 2175 |       auto var_type = storedTy;
      |                       ^~~~~~~~
/vagrant/src/ast/semantic_analyser.cpp:2176:24: error: use of deleted function ‘bpftrace::SizedType::SizedType(const bpftrace::SizedType&)’
 2176 |       auto expr_type = assignTy;
      |                        ^~~~~~~~
/vagrant/src/ast/semantic_analyser.cpp: In member function ‘void bpftrace::ast::SemanticAnalyser::assign_map_type(const bpftrace::ast::Map&, const bpftrace::SizedType&)’:
/vagrant/src/ast/semantic_analyser.cpp:2821:26: error: use of deleted function ‘bpftrace::SizedType& bpftrace::SizedType::operator=(const bpftrace::SizedType&)’
 2821 |         search->second = type;
```

danobi: Ah, I see. I guess this PR is the simplest fix for now.

danobi (approved):

PR #1639, opened 2020-11-19
Support clear() and delete() on a count()-based map without key

### LINKED ISSUE #1350: clear() on count()-based map fails with a key error
### bpftrace --info

```
root@ubuntu18:~# bpftrace --info
System
  OS: Linux 5.7.0-rc5-default #361 SMP Tue May 26 12:34:40 EEST 2020
  Arch: x86_64

Build
  version: v0.10.0-124-g147a
  LLVM: 6
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
```

### What reproduces the bug?
Following bpftrace program exits with an error: 

`bpftrace -e 'kprobe:btrfs_commit_transaction { @exec = count() } interval:s:3 { print(@exec); clear(@exec); } '`  

I get: 

```
what():  Could not clear map with ident "@exec", err=-1
```

My expectation is to print the number of times this kprobe has fired and clear it up for the next interval. I've literally coded this after an example in the reference guide.

mmisono:
439a6bc changes so that count()-based map with no arguments uses
BPF_MAP_TYPE_PERCPU_ARRAY for the performance. The problem is crear()
uses bpf_map_delete_elem(), but ARRAY-typed maps cannot be deleted.
To solve this, suggested by fbs in #1350 (comment),
use zero() internally in clear() if a map type is array.

Like clear(), delete() also does not work for count()-based map (please
see the previous commit message for the details.) To fix this, instead
calling bpf_map_delete_elem() when delete(), call bpf_mpa_updat_elem()
and update the map value with zero.

Closes #1350 

##### Checklist

### DIFF
--- a/src/imap.h
+++ b/src/imap.h
@@ -27,6 +27,11 @@ public:
     return map_type_ == BPF_MAP_TYPE_PERCPU_HASH ||
            map_type_ == BPF_MAP_TYPE_PERCPU_ARRAY;
   }
+  bool is_clearable() const
+  {
+    return map_type_ != BPF_MAP_TYPE_ARRAY &&
+           map_type_ != BPF_MAP_TYPE_PERCPU_ARRAY;
+  }
 
   // unique id of this map. Used by (bpf) runtime to reference
   // this map
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,8 @@ and this project adheres to
   - [#1658](https://github.com/iovisor/bpftrace/pull/1658)
 - Don't create a tuple if an element size if zero
   - [#1653](https://github.com/iovisor/bpftrace/pull/1653)
+- Support clear() and delete() on a count()-based map without a key
+  - [#1639](https://github.com/iovisor/bpftrace/pull/1639)
 
 #### Tools
 - Hook up execsnoop.bt script onto `execveat` call
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -502,7 +502,19 @@ void CodegenLLVM::visit(Call &call)
     auto &arg = *call.vargs->at(0);
     auto &map = static_cast<Map&>(arg);
     AllocaInst *key = getMapKey(map);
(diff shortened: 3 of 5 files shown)


mmisono:
So the difference between before and after 439a6bc is a count()-based map is an array, so it'll be printed out even if we clear() or delete() it. For example: 


```
sudo ./src/bpftrace -e 'BEGIN { @ = 1; clear(@); exit(); }'
Attaching 1 probe...

// no output
```
but
```
% sudo ./src/bpftrace -e 'BEGIN { @ = count(); clear(@); exit(); }' probe...


@: 0
```

But I guess most people want to know the final value of the count.

src/imap.h
```diff
@@ -27,6 +27,11 @@ class IMap
     return map_type_ == BPF_MAP_TYPE_PERCPU_HASH ||
            map_type_ == BPF_MAP_TYPE_PERCPU_ARRAY;
   }
+  bool is_not_clearable() const
```
danobi, line 30: Can we name this `is_clearable()`? Would be confusing to read `!is_not_clearable()` later on (the double negation).

danobi (commented):
LGTM, small nit

src/imap.h
mmisono: That makes sense. will fix.

mmisono: done.

danobi (approved):

danobi:
Needs a rebase

PR #1640, opened 2020-11-19
Glob memleak

fbs:
This fixes a glob memory leak. We short circuit the loop but don't clean up the glob object. This is only triggered for scripts that have a wildcard tracepoint and  `need_tp_args_structs == false`.

As the whole function was a bit long and hard to read with all the nesting I've split it into parts.

### DIFF
--- a/src/tracepoint_format_parser.cpp
+++ b/src/tracepoint_format_parser.cpp
@@ -71,7 +71,10 @@ bool TracepointFormatParser::parse(ast::Program *program, BPFtrace &bpftrace)
           }
 
           if (probe->tp_args_structs_level <= 0)
+          {
+            globfree(&glob_result);
             continue;
+          }
 
           for (size_t i = 0; i < glob_result.gl_pathc; ++i) {
             std::string filename(glob_result.gl_pathv[i]);


src/tracepoint_format_parser.cpp
```diff
@@ -23,11 +29,14 @@ bool TracepointFormatParser::parse_wildcard_ap(ast::Probe *probe,
   std::string format_file_path = "/sys/kernel/debug/tracing/events/" +
                                  category + "/" + event_name + "/format";
 
-  glob_t glob_result;
+  // Use a unique_ptr to have to system call globfree() for us instead of having
+  // to do it ourselves. The braces after `new glob_t()` are required to init
+  // the struct. `globfree()` on random data is unsafe
+  std::unique_ptr<glob_t, decltype(&freeglob)> glob_result(new glob_t(),
```
danobi, line 35: Custom deleter on a unique_ptr works, but in the past I've used the scope guard pattern (it works well and is really resuable).

Source: https://github.com/facebookincubator/oomd/blob/master/src/oomd/util/ScopeGuard.h
Example: https://github.com/facebookincubator/oomd/blob/7bffe9e0d4143ca8acacc4bc640888d6e14e36b4/src/oomd/util/ScopeGuardTest.cpp#L27-L37

Maybe copy paste that into bpftrace?

fbs: I guess we can give that a go :)

fbs:
I've updated it to just include a glob free. Doing the refactoring and scope guarding would be nice but not all that important right now and its tedious to keep rebasing onto other changes. This is good enough for now

danobi (approved):

Issue #1137, opened 2020-02-10
ARM64: several test case failure in runtime-tests.sh


lostjeffle:
OS: 4.19.48
arch: aarch64
bpftrace: v0.9.3

I got several test case failure when testing the bpftrace rpm package, which is built by myself.

All failed test cases include:
- array.array_element_access_-_assign_to_map
- array.array_element_access_-_assign_to_var
- builtin.sarg
- call.str
- intptrcast.Casting_ints
- intptrcast.Sum_casted_value
- json-output.join_delim
- regression.logical_and_or_different_sizes

For example, the error log of "regression.logical_and_or_different_sizes" is:

```
[ RUN      ] regression.logical_and_or_different_sizes
[  FAILED  ] regression.logical_and_or_different_sizes
	Command: /usr/bin/bpftrace -v -e 'struct Foo { int m; } uprobe:./testprogs/simple_struct:func { $foo = (struct Foo*)arg0; printf("%d %d %d %d %d¥n", $foo->m, $foo->m && 0, 1 && $foo->m, $foo->m || 0, 0 || $foo->m); exit(); }'
	Expected: 2 0 1 1 1
	Found: Attaching 1 probe...

Program ID: 1103

Bytecode: 
0: (bf) r6 = r1
1: (79) r7 = *(u64 *)(r6 +0)
2: (b7) r8 = 0
3: (7b) *(u64 *)(r10 -48) = r8
4: (7b) *(u64 *)(r10 -56) = r8
5: (bf) r1 = r10
6: (07) r1 += -4
7: (b7) r2 = 4
8: (bf) r3 = r7
9: (85) call bpf_probe_read#4
10: (61) r1 = *(u32 *)(r10 -4)
11: (63) *(u32 *)(r10 -48) = r1
12: (bf) r1 = r10
13: (07) r1 += -4
14: (b7) r2 = 4
15: (bf) r3 = r7
16: (85) call bpf_probe_read#4
17: (7b) *(u64 *)(r10 -40) = r8
18: (bf) r1 = r10
19: (07) r1 += -4
20: (b7) r2 = 4
21: (bf) r3 = r7
22: (85) call bpf_probe_read#4
23: (b7) r8 = 1
24: (61) r2 = *(u32 *)(r10 -4)
25: (b7) r1 = 1
26: (55) if r2 != 0x0 goto pc+1
[...]

PR #1641, opened 2020-11-20
tests/semantic: Add architecture specific testcase support

lostjeffle:
commit 48d405133160 ("bpftrace: Add architecture specific testcase
support") added architecture specific support for runtime tests.
Still there are architecture specific testcases in semantic_analyser,
where 'ip' register is hard coded in semantic_analyser.call_reg and
semantic_analyser.builtin_functions.

These testcases are all c/c++ files and thus we need conditional compiling.

Signed-off-by: Jeffle Xu <jefflexu@linux.alibaba.com>

#1137 had ever noted that some architecture specific test cases are hard coded, and thus will fail in some architectures not supported.

- some runtime-test had been fixed by  commit 48d405133160 ("bpftrace: Add architecture specific testcase support")
- opensnoop/statsnoop in tools-parsing-test had been fixed by commit 330953c11190c23a12f98d374d1b5eb6d7aad1fe ("fix build failed on aarch64")

This patch is supposed to fix the failure in semantic_analyser.builtin_functions, semantic_analyser.call_reg.

##### Checklist

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -120,6 +120,18 @@ if(LIBBCC_ATTACH_UPROBE_SEVEN_ARGS_SIGNATURE)
   target_compile_definitions(bpftrace_test PRIVATE LIBBCC_ATTACH_UPROBE_SEVEN_ARGS_SIGNATURE)
 endif(LIBBCC_ATTACH_UPROBE_SEVEN_ARGS_SIGNATURE)
 
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+  target_compile_definitions(bpftrace_test PRIVATE ARCH_AARCH64)
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "ppc64" OR
+       CMAKE_SYSTEM_PROCESSOR STREQUAL "ppc64le")
+  target_compile_definitions(bpftrace_test PRIVATE ARCH_PPC64)
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "s390" OR
+       CMAKE_SYSTEM_PROCESSOR STREQUAL "s390x")
+  target_compile_definitions(bpftrace_test PRIVATE ARCH_S390)
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+  target_compile_definitions(bpftrace_test PRIVATE ARCH_X86_64)
+endif()
+
 target_link_libraries(bpftrace_test arch ast parser resources)
 
 target_link_libraries(bpftrace_test ${LIBBCC_LIBRARIES})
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -106,6 +106,8 @@ and this project adheres to
   - [#1580](https://github.com/iovisor/bpftrace/pull/1580)
 - Printing of small integers with `printf`
   - [#1532](https://github.com/iovisor/bpftrace/pull/1532)
+- Add architecture specific testcase support for semantic_analyser
```
fbs, line 109: We only put the enduser facing changes in the changelog, internal test changes are not relevant for them so you can remove this :)

tests/CMakeLists.txt
```diff
@@ -2,6 +2,19 @@ add_compile_options("-Wno-undef")
 add_compile_options("-Wno-switch-default")
 add_compile_options("-Wno-switch-enum")
 
+# Architecture-specific preprocessor macro for *.cpp/*.c
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+  add_definitions(-DARCH_AARCH64)
```
danobi, line 7: I think it's a bit nicer to use the target_compile_definitions
```suggestion
  target_compile_definitions(bpftrace_test PRIVATE ARCH_AARCH64)
```

CHANGELOG.md
lostjeffle: Regards.

tests/CMakeLists.txt
lostjeffle: Actually 'target_compile_definitions' is used in the original patch, but failed with several automatic build test triggered by GitHub, warning that "target_compile_definitions is not defined". Maybe it's related to the version of cmake, since target_compile_definitions is added since cmake3?

fbs: Do you have a link to the failed builds? It should work. We use it in a quite a few places and we require cmake 3.something anyway:

```
tests/CMakeLists.txt
69:  target_compile_definitions(bpftrace_test PRIVATE HAVE_LIBBPF_MAP_BATCH)
73:  target_compile_definitions(bpftrace_test PRIVATE HAVE_NAME_TO_HANDLE_AT=1)
76:  target_compile_definitions(bpftrace_test PRIVATE HAVE_BCC_PROG_LOAD)
79:  target_compile_definitions(bpftrace_test PRIVATE HAVE_BCC_CREATE_MAP)
```

Maybe you placed it before the bpftrace_test target definition? Needs to go on line 69 or later

> I'm not familiar with cmake and really look forward you guys' professional opinion. Thanks.

We all struggle with cmake ;)

lostjeffle:
@fbs @danobi 
Changes since v1:
- remove changes in CHANGELOG.md
- use 'target_compile_definitions' instead
- skip some arch specific codegen test cases when arch not supports

fbs:
thanks!

If we add a lot of cases it might be nice to change the expected value instead of just removing the test but this is fine for now :)

PR #1644, opened 2020-11-22
bpftrace: Fix list probes by pids tests

sumanthkorikkar:
List by pid fails on Fedora 33, as two pids are returned for
uprobe_test, usdt_test.

root      173886  173874  0 10:12 ?        00:00:00 [uprobe_test] \<defunct\>
root      173896  173874  0 10:12 ?        00:00:00 ./testprogs/uprobe_test

pidof uprobe_test returns : 173896 173886

I could see these two pids in x86 as well. However, on s390x, pidof sometimes returns 2 pids, which result in failure

This results in Usage error for listing usecase, as getopt_long expects
just one pid in -p.

Fix this by returning just one pid. pidof -s seems to return the latest
pid. i.e testprogs/uprobe_test.

##### Checklist

### DIFF
--- a/tests/runtime/engine/utils.py
+++ b/tests/runtime/engine/utils.py
@@ -161,13 +161,16 @@ class Utils(object):
                     # a test program needs to accept arguments. It covers the
                     # current simple calls with no arguments
                     child_name = os.path.basename(test.before.split()[-1])
-                    while subprocess.call(["pidof", child_name], stdout=dn, stderr=dn) != 0:
+                    while subprocess.call(["pidof", "-s", child_name], stdout=dn, stderr=dn) != 0:
                         time.sleep(0.1)
                         waited+=0.1
                         if waited > test.timeout:
                             raise TimeoutError('Timed out waiting for BEFORE %s ', test.before)
 
             bpf_call = Utils.prepare_bpf_call(test)
+            if test.before:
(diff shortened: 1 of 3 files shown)


fbs:
I wonder whether its nicer to have the runtime test runner keep track of the pid and use a special variable that it subsititutes, so the test would be like:

```
 RUN bpftrace -l 'usdt:*' -p {{BEFORE_PID}}
```

It should work as long as the process itself doesn't fork again, not sure how often that happens.

sumanthkorikkar:
ok sure. I will try and see.

sumanthkorikkar:
Ok. Strange. Seems to work in my system. Will check.

sumanthkorikkar:
Ok, in ubuntu seems to return the id of the shell process, with subprocess.Popen() and Shell = True.

sumanthkorikkar:
> I wonder whether its nicer to have the runtime test runner keep track of the pid and use a special variable that it subsititutes, so the test would be like:
> 
> ```
>  RUN bpftrace -l 'usdt:*' -p {{BEFORE_PID}}
> ```
> 
> It should work as long as the process itself doesn't fork again, not sure how often that happens.

Added this behaviour. Thanks.

fbs:
Bit uglier than i hoped for, but still good to have it in one place instead of everywhere :)

fbs (approved):

sumanthkorikkar:
:)

sumanthkorikkar:
If it is fine, could we please have this in master. (which shows 16 failures due to this reason). Thank you.

PR #1645, opened 2020-11-22
Fix several undefined behaviors

mmisono:
UBSan (Undefined Behavior Sanitizer) found several errors when running bpftrace_test. This commits fix them.
Please see the commit messages for the details.

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -627,22 +627,22 @@ void SemanticAnalyser::visit(Call &call)
         }
       }
 
-      if (call.vargs->size() == 2)
+      if (is_final_pass() && call.vargs->size() == 2 &&
+          check_arg(call, Type::integer, 1, false))
       {
         auto &size_arg = *call.vargs->at(1);
-        long value = static_cast<Integer &>(size_arg).n;
-        if (value < 0)
+        if (size_arg.is_literal)
         {
-          LOG(ERROR, call.loc, err_)
-              << call.func << "cannot use negative length (" << value << ")";
+          auto &integer = static_cast<Integer &>(size_arg);
+          long value = integer.n;
+          if (value < 0)
+            LOG(ERROR, call.loc, err_)
+                << call.func << "cannot use negative length (" << value << ")";
         }
       }
 
       // Required for cases like strncmp(str($1), str(2), 4))
       call.type.SetAS(t.GetAS());
-      if (is_final_pass() && call.vargs->size() > 1) {
-        check_arg(call, Type::integer, 1, false);
-      }
     }
     has_pos_param_ = false;
   }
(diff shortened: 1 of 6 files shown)


src/ast/semantic_analyser.cpp
```diff
@@ -607,11 +607,15 @@ void SemanticAnalyser::visit(Call &call)
       if (call.vargs->size() == 2)
       {
         auto &size_arg = *call.vargs->at(1);
-        long value = static_cast<Integer &>(size_arg).n;
-        if (value < 0)
+        auto integer = dynamic_cast<Integer *>(&size_arg);
+        if (integer)
```
danobi, line 611: I think this check was wrong to begin with. We should only be downcasting and checking the value after we know it's a literal, right? Otherwise we could be reading junk. So in `is_final_pass()` and after the `check_arg()`. Same for the `buf` case. I believe that would solve this static_cast issue as well.

danobi (changes requested):

src/ast/semantic_analyser.cpp
mmisono: That makes sense. But I found in the case of buf(), the length (buffer_size) is used in other places, and we cannot check the length in the final pass.. So I decided to use dynamic_cast here. 735b944 please let me know if there is a better way.

src/ast/semantic_analyser.cpp
```diff
@@ -604,7 +604,7 @@ void SemanticAnalyser::visit(Call &call)
         }
       }
 
-      if (call.vargs->size() == 2)
+      if (is_final_pass() && call.vargs->size() == 2)
```
danobi, line 607: I think you have to check the type before the downcast otherwise you don't know what you're casting
```suggestion
      if (is_final_pass() && check_arg(..) && call.vargs->size() == 2)
```

And then delete the `check_arg()` below

src/ast/semantic_analyser.cpp
danobi: Can't buf() just use `max_buffer_size` until the final pass? Then on the final pass it can do all the error checks and calculations.

src/ast/semantic_analyser.cpp
mmisono: I see. I missed the `check_arg()` below. thanks.

src/ast/semantic_analyser.cpp
mmisono: I first tried it and failed. Then I realized that AssignMap did not update the map size. I fixed it and now I think it's OK: bfded77

danobi: Is update_assign_map_type() fixing a separate issue? I'm not seeing why it's related.

mmisono: without it, https://github.com/iovisor/bpftrace/blob/7ca62cb0aca9d6125d8a96fdc0238f7e07354097/src/ast/semantic_analyser.cpp#L2067 is failed because it thinks map size is `max_buffer_size`, but `expr_size` become smaller in the final pass.

mmisono:
hmm.. still there are some dowancast issues

```
% sudo ./src/bpftrace -e "kprobe:f { str(arg0, arg1); }" 
/disk/work/bpftrace2/src/ast/semantic_analyser.cpp:634:22: runtime error: downcast of address 0x611000002d40 which does not point to an object of type 'bpftrace::ast::Integer'
0x611000002d40: note: object is of type 'bpftrace::ast::Builtin'
 11 00 00 10  58 17 3a 01 00 00 00 00  00 00 00 00 00 00 00 00  01 00 00 00 16 00 00 00  00 00 00 00
              ^~~~~~~~~~~~~~~~~~~~~~~
              vptr for 'bpftrace::ast::Builtin
```

mmisono:
the problem is Type::integer does not necessarily mean that is ast::Integer.

mmisono:
ah, sorry I forgot about `is_literal`... now the problem is almost solved

danobi (approved):

PR #1646, opened 2020-11-22
Check a call function is available for a probe type

mmisono:
Some call functions (e.g., `reg()`) is not available for some probe types. To check this, add `check_available(call, type)` function. This function returns true if the call is available for that type. Also, disable `reg()` in k(ret)func using this function. Previously that is mistakenly allowed and resulted in invalid BPF codes.

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -456,6 +456,15 @@ void SemanticAnalyser::visit(Call &call)
     }
   }
 
+  for (auto &ap : *probe_->attach_points)
+  {
+    if (!check_available(call, *ap))
+    {
+      LOG(ERROR, call.loc, err_) << call.func << " can not be used with \""
+                                 << ap->provider << "\" probes";
+    }
+  }
+
   if (call.func == "hist") {
     check_assignment(call, true, false, false);
     check_nargs(call, 1);
@@ -766,16 +775,6 @@ void SemanticAnalyser::visit(Call &call)
   }
   else if (call.func == "reg") {
     if (check_nargs(call, 1)) {
-      for (auto &attach_point : *probe_->attach_points) {
-        ProbeType type = probetype(attach_point->provider);
-        if (!check_available(call, type))
-        {
-          LOG(ERROR, call.loc, err_)
-              << "The reg function cannot be used with 'tracepoint' probes";
-          continue;
-        }
-      }
-
       if (check_arg(call, Type::string, 0, true)) {
         auto reg_name = bpftrace_.get_string_literal(call.vargs->at(0));
         int offset = arch::offset(reg_name);;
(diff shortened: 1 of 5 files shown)


mmisono:
This is a continuation of #1380. kfunc reg() tests would conflicts #1641.

src/ast/semantic_analyser.cpp
```diff
@@ -2820,6 +2818,83 @@ bool SemanticAnalyser::check_symbol(const Call &call, int arg_num __attribute__(
...
+      case ProbeType::invalid:
+      case ProbeType::interval:
+      case ProbeType::software:
+      case ProbeType::hardware:
+      case ProbeType::watchpoint:
+        return false;
+    }
+  }
+
+  if (type == ProbeType::invalid)
```
fbs, line 2892: can we move this to the top and then remove all the invalid cases from the switch statements? Or does that trigger a missing case warning?

mmisono: yes, -Wswitch emits warnings. `warning: enumeration value 'invalid' not handled in switch [-Wswitch]`

mmisono:
rebased and update tests since #1641 is just merged.

danobi:
We can make a call to `check_available(..)` at the beginning of the function and delete the other callsites right? Would reduce duplicated code.

fbs:
I wonder if its doable to write a simple spec for each call  that handles  this and all the other standard check things for calls (and builtins). E.g.

```
CallSpec {
 .name = "str",
 .probe_types = any,
 .required_features = {somefeature},
 .min_args = 1,
 .max_args = 2,
 .arg_types = [
  [{ type = integer, literal = true}, { type = pointer } ], // arg0
  [{ type = integer }], // arg1
  .func = &semantic_analyser::call_str
]}
```

It would remove quite a bit of basically duplicate code everywhere, the deep nested long functions and would help standardize error messages. It should also make it easier to get a quick overview of the calls we have an what kind of arguments they take, might even be possible to generate nice help format output from it.

But there probably are some cases that make it hard to do

src/ast/semantic_analyser.cpp
```diff
@@ -810,8 +811,7 @@ void SemanticAnalyser::visit(Call &call)
     for (auto &ap : *probe_->attach_points)
     {
       ProbeType type = probetype(ap->provider);
-      if (type != ProbeType::usdt && type != ProbeType::uretprobe &&
-          type != ProbeType::uprobe)
+      if (!check_available(call, type))
       {
         LOG(ERROR, call.loc, err_)
             << "uaddr can only be used with u(ret)probes and usdt probes";
```
fbs, line 817: Feels like we now need to keep track of this in two places. We have to update check_available to reflect the probe types and then update the error message too. Can't we update check_available to output the err msg, the other checks do it as well.

Changing the signature to `check_available(Call &c, std::vector<probetype &> t)` might make that easier too.

mmisono:
fbs's proposal seems great. For now, I implemented danobi's suggestion.

src/ast/semantic_analyser.cpp
mmisono: I changed so that the error messages are printed out when check_available() failed.
https://github.com/iovisor/bpftrace/pull/1646/commits/6fcf1a72913db4ea699ce8cbd7e7215cd9e70f77#diff-042d9701e67f79452452faac3c625c3ca83e7692abd3274956abe30d58ae0c39R459-R467

danobi (approved):

mmisono:
rebased

PR #1647, opened 2020-11-22
Add built-in function for formatting MAC addresses

### LINKED ISSUE #1575: mac address printer
We have `ntop` to print ip addresses but nothing for mac addresses. Printing macs in the usual format without support from a builtin call is somewhat tricky to do, for v0.11 I came up with:

```
struct mac {
  char common[3];
  char h1;
  char h2;
  char h3;
};

kprobe:arp_create {
  $sip = arg4;
  $dip = arg2;
  $smac = (struct mac*) sarg0;
  $dmac = (struct mac*)sarg1;
  time("%H:%M:%S ");
  printf("\033[32mSEND:\033[0m   SRC: %16s %hhX:%hhX:%hhX", ntop($sip), $smac->h1, $smac->h2, $smac->h3);
  printf(" -> DST: %16s %hhX:%hhX:%hhX\n", ntop($dip), $dmac->h1, $dmac->h2, $dmac->h3);
}
```

Should be easy to implement. Just needs a good name.

acj:
This adds a helper function named `macaddr` (tentative name) that formats MAC addresses in the canonical hex style, e.g. `00:1C:42:B5:20:09`. Resolves #1575.

WIP, but feedback is welcome and appreciated.

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1016,6 +1016,23 @@ void CodegenLLVM::visit(Call &call)
     auto arg = call.vargs->at(0);
     auto scoped_del = accept(arg);
   }
+  else if (call.func == "macaddr")
+  {
+    // MAC addresses are presented as char[6]
+    AllocaInst *buf = b_.CreateAllocaBPFInit(call.type, "macaddr");
+    auto macaddr = call.vargs->front();
+    auto scoped_del = accept(macaddr);
+
+    b_.CreateProbeRead(ctx_,
+                       static_cast<AllocaInst *>(buf),
+                       macaddr->type.GetSize(),
+                       expr_,
+                       macaddr->type.GetAS(),
+                       call.loc);
+
+    expr_ = buf;
+    expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
+  }
   else
   {
     LOG(FATAL) << "missing codegen for function \"" << call.func << "\"";
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1187,6 +1187,26 @@ void SemanticAnalyser::visit(Call &call)
     call.type = call.vargs->front()->type;
     call.type.SetAS(as);
   }
+  else if (call.func == "macaddr")
+  {
+    if (!check_nargs(call, 1))
+      return;
+
+    auto &arg = call.vargs->at(0);
+
+    if (!arg->type.IsIntTy() && !arg->type.IsArrayTy() &&
+        !arg->type.IsByteArray() && !arg->type.IsPtrTy())
+      LOG(ERROR, call.loc, err_)
+          << call.func << "() only supports array or pointer arguments"
+          << " (" << arg->type.type << " provided)";
+
+    auto type = arg->type;
+    if ((type.IsArrayTy() || type.IsByteArray()) && type.GetSize() != 6)
+      LOG(ERROR, call.loc, err_)
+          << call.func << "() argument must be 6 bytes in size";
+
+    call.type = CreateMacAddress();
+  }
   else
   {
     LOG(ERROR, call.loc, err_) << "Unknown function: '" << call.func << "'";
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -845,6 +845,11 @@ std::vector<std::unique_ptr<IPrintable>> BPFtrace::get_arg_values(const std::vec
         arg_values.push_back(std::make_unique<PrintableInt>(
             *reinterpret_cast<uint64_t *>(arg_data + arg.offset)));
         break;
+      case Type::mac_address:
+        arg_values.push_back(
+            std::make_unique<PrintableString>(resolve_mac_address(
+                reinterpret_cast<uint8_t *>(arg_data + arg.offset))));
+        break;
         // fall through
       default:
         LOG(FATAL) << "invalid argument type";
@@ -1452,6 +1457,8 @@ std::string BPFtrace::map_value_to_str(const SizedType &stype,
         reinterpret_cast<AsyncEvent::Strftime *>(value.data())->strftime_id,
         reinterpret_cast<AsyncEvent::Strftime *>(value.data())
             ->nsecs_since_boot);
+  else if (stype.IsMacAddressTy())
+    return resolve_mac_address(value.data());
   else
     return std::to_string(read_data<int64_t>(value.data()) / div);
 }
@@ -1974,6 +1981,22 @@ int BPFtrace::resolve_uname(const std::string &name,
   sym->size = 8;
   return 0;
 #endif
+}
+
+std::string BPFtrace::resolve_mac_address(const uint8_t *mac_addr) const
+{
+  const size_t SIZE = 18;
+  char addr[SIZE];
+  snprintf(addr,
+           SIZE,
+           "%02X:%02X:%02X:%02X:%02X:%02X",
+           mac_addr[0],
+           mac_addr[1],
+           mac_addr[2],
+           mac_addr[3],
+           mac_addr[4],
+           mac_addr[5]);
+  return std::string(addr);
 }
 
 #ifdef HAVE_BCC_ELF_FOREACH_SYM
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -206,6 +206,7 @@ TEST(semantic_analyser, builtin_functions)
   test("kprobe:f { cat(\"/proc/uptime\") }", 0);
   test("uprobe:/bin/bash:main { uaddr(\"glob_asciirange\") }", 0);
   test("kprobe:f { cgroupid(\"/sys/fs/cgroup/unified/mycg\"); }", 0);
+  test("kprobe:f { macaddr(0xffff) }", 0);
 }
 
 TEST(semantic_analyser, undefined_map)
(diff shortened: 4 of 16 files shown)


acj:
The codegen validator is currently failing with an error that I don't understand:

```
4: /usr/bin/llvm-as-7: /home/runner/work/bpftrace/bpftrace/tests/codegen/llvm/call_macaddr.ll:24:9: error: stored value and pointer type do not match
4: store i64* %macaddr, [6 x i8]* %"@x_val"
4:         ^
```

Fiddling with the codegen types, I can even get this:

```
4: /usr/bin/llvm-as-7: /vagrant/tests/codegen/llvm/call_macaddr.ll:24:9: error: stored value and pointer type do not match
4:   store [8 x i8]* %macaddr, [8 x i8]* %"@x_val"
4:         ^
```

... which seems odd. If I remove the `@ = ` assignment from the codegen test, then everything passes, but I think it should be possible for the current test to pass. Any tips?

src/ast/codegen_llvm.cpp
```diff
@@ -1004,6 +1004,24 @@ void CodegenLLVM::visit(Call &call)
     auto arg = call.vargs->at(0);
     auto scoped_del = accept(arg);
   }
+  else if (call.func == "macaddr")
+  {
+    // Mac addresses are presented as char[6]
+    AllocaInst *buf = b_.CreateAllocaBPF(b_.getInt64Ty(), "macaddr");
```
fbs, line 1010:
```suggestion
    AllocaInst *buf = b_.CreateAllocaBPFInit(call.type, "macaddr");
```

You create a i64 type with this, storing that in an `[6*i8]` later causes issues. `IRBuilderBPF::GetType` knows how to handle array types so it shoudl work ok

src/bpftrace.cpp
```diff
@@ -1959,6 +1966,20 @@ int BPFtrace::resolve_uname(const std::string &name,
 #endif
 }
 
+std::string BPFtrace::resolve_mac_address(const uint8_t *mac_addr) const
+{
+  char addr[18];
+  sprintf(addr,
```
fbs, line 1972:
```suggestion
  snprintf(addr, 18
```

Not needed but just to be safe anyway

tests/semantic_analyser.cpp
```diff
@@ -802,6 +803,21 @@ TEST(semantic_analyser, call_stack)
...
+TEST(semantic_analyser, call_macaddr)
+{
+  std::string structs = "struct mac { char addr[6]; }; ";
+
+  test(structs + "kprobe:f { macaddr((struct mac*)arg0); }", 0);
+  test(structs + "kprobe:f { macaddr(((struct mac*)arg0)->addr); }", 0);
+
+  test(structs + "kprobe:f { @x = macaddr((struct mac*)arg0); }", 0);
+
+  test(structs + "kprobe:f { @x[macaddr((struct mac*)arg0)] = 1; }", 0);
```
fbs, line 815: having a test for  macaddr in printf would be good

fbs (changes requested):
Sweet! Some comments.

Don't really like the name, don't really know anything better either.

src/ast/codegen_llvm.cpp
acj: Thanks. The latest commit has this change, but I'm still seeing this in my local build:

```
4: /usr/bin/llvm-as-7: /vagrant/tests/codegen/llvm/call_macaddr.ll:28:9: error: stored value and pointer type do not match
4:   store [6 x i8]* %macaddr, [6 x i8]* %"@x_val"
4:         ^
```

I also tried using `macaddr->type` instead of `call.type` but got the same result

fbs: This comes from the AssignMap codegen. It thinks its dealing with a simpel integer value so it creates a new allocation and then stores it into it. shouldBeOnStackAlready needs updating for the new type

acj: Aha, that did it. Thank you :)

src/ast/semantic_analyser.cpp
```diff
@@ -1190,6 +1190,26 @@ void SemanticAnalyser::visit(Call &call)
...
+  else if (call.func == "macaddr")
+  {
+    if (!check_nargs(call, 1))
+      return;
+
+    auto &arg = call.vargs->at(0);
+
+    if (!arg->type.IsIntTy() && !arg->type.IsArray() && !arg->type.IsPtrTy())
+      LOG(ERROR, call.loc, err_)
+          << "() only supports int, array, or pointer arguments"
```
fbs, line 1202:
```suggestion
          << "() only supports array or pointer arguments"
```

Using an actual int as mac doesn't work as you're probe_reading them. Only ints that are actually pointers (like arg0 can be) would work.

src/ast/codegen_llvm.cpp
```diff
@@ -1004,6 +1004,24 @@ void CodegenLLVM::visit(Call &call)
...
   }
+  else if (call.func == "macaddr")
+  {
+    // MAC addresses are presented as char[6]
+    AllocaInst *buf = b_.CreateAllocaBPFInit(call.type, "macaddr");
+    b_.CREATE_MEMSET(buf, b_.getInt8(0), 6, 1);
+    auto macaddr = call.vargs->front();
+    auto scoped_del = accept(macaddr);
+
+    b_.CreateProbeRead(ctx_,
```
fbs, line 1015: IIRC arrays can end up on stack already so this would probe read from the stack in that case

src/ast/codegen_llvm.cpp
```diff
@@ -1004,6 +1004,24 @@ void CodegenLLVM::visit(Call &call)
     auto arg = call.vargs->at(0);
     auto scoped_del = accept(arg);
   }
+  else if (call.func == "macaddr")
+  {
+    // MAC addresses are presented as char[6]
+    AllocaInst *buf = b_.CreateAllocaBPFInit(call.type, "macaddr");
+    b_.CREATE_MEMSET(buf, b_.getInt8(0), 6, 1);
```
fbs, line 1011: allocainit already inits. I'm not sure that memset is really needed here anyway as we proberead 6 bytes into a 6 byte buffer

tests/semantic_analyser.cpp
```diff
@@ -802,6 +803,23 @@ TEST(semantic_analyser, call_stack)
   test("kprobe:f { @x = 3; ustack(perf, @x) }", 1);
 }
 
+TEST(semantic_analyser, call_macaddr)
+{
+  std::string structs = "struct mac { char addr[6]; }; ";
+
+  test(structs + "kprobe:f { macaddr((struct mac*)arg0); }", 0);
```
fbs, line 810: need some tests (and runtime tests) for the other supported cases too, like

macaddr(arg0)
macaddr(*(struct X*)y)
macaddr(somekindofarray)

fbs (commented):
Looks good but I think it will have some issues with less common types

Changelog and refguide entries need to be added at some point too

src/ast/semantic_analyser.cpp
acj: Ah yep, the int was from an earlier iteration and doesn't apply now

src/ast/codegen_llvm.cpp
acj: How should I check for that case? I see that `shouldBeOnStackAlready` doesn't include arrays, so it wouldn't be enough to branch on that. But I'm thinking that we need to branch on something and then store (instead of probe read) if the arg is already on the stack.

tests/semantic_analyser.cpp
acj: I'm not sure what you meant for these two:

>macaddr((struct X)y)

How would it differ from this line? Do you mean a cast to a non-pointer struct type?

>macaddr(somekindofarray)

The line below this one is testing an array, no?

acj:
@fbs ping

tests/semantic_analyser.cpp
fbs: oh woops, codeblock makes the difference

```
macaddr(arg0)
macaddr(*(struct mac*)y)
macaddr(somekindofarray)
```

> > macaddr(somekindofarray)
> 
> The line below this one is testing an array, no?

I'm not sure, I just want to make sure it works for data that has already been loaded onto the stack (so no probe_read required).

fbs:
Sorry, missed this one, busy times :(

src/ast/codegen_llvm.cpp
fbs: Yes it would have to load/memcpy instead of probe reading.

I dont know the best way to check, arrays need some love. I'd be ok with this probe reading for now too, can always fix it in the future.

acj:
All good! I should have time this week to finish testing the different cases, update the docs, etc.

acj:
@fbs ready for another look when you have a moment

fbs (approved):

fbs:
@acj sorry for all the delays. Can you rebase this so we can get it in?

acj:
@fbs All good. Here's the diff addressing the type changes and fixing a logging typo:

```diff
--- src/ast/semantic_analyser.cpp
+++ src/ast/semantic_analyser.cpp
@@ -1194,14 +1194,14 @@ void SemanticAnalyser::visit(Call &call)

     auto &arg = call.vargs->at(0);

-    if (!arg->type.IsIntTy() && !arg->type.IsArray() && !arg->type.IsPtrTy())
+    if (!arg->type.IsIntTy() && !arg->type.IsArrayTy() &&
+        !arg->type.IsByteArray() && !arg->type.IsPtrTy() &&
+        !arg->type.IsRecordTy())
       LOG(ERROR, call.loc, err_)
-          << "() only supports array or pointer arguments"
+          << call.func << "() only supports array or pointer arguments"
           << " (" << arg->type.type << " provided)";

     auto type = arg->type;
-
-    if (arg->type.IsArray() && type.GetSize() != 6)
+    if ((type.IsArrayTy() || type.IsByteArray()) && type.GetSize() != 6)
       LOG(ERROR, call.loc, err_)
           << call.func << "() argument must be 6 bytes in size";
```

src/ast/semantic_analyser.cpp
```diff
@@ -1187,6 +1187,27 @@ void SemanticAnalyser::visit(Call &call)
...
+  else if (call.func == "macaddr")
+  {
+    if (!check_nargs(call, 1))
+      return;
+
+    auto &arg = call.vargs->at(0);
+
+    if (!arg->type.IsIntTy() && !arg->type.IsArrayTy() &&
+        !arg->type.IsByteArray() && !arg->type.IsPtrTy() &&
+        !arg->type.IsRecordTy())
```
fbs, line 1199: `type::record` should only be structs not struct pointers so this should be removed right?

acj: Yep, good catch. Fixed in latest commit

PR #1652, opened 2020-11-25
bpftrace: Correct ntop tests for BE and add more tests

sumanthkorikkar:
- Ideally ntop should take the arguments in network byte order. However,
ntop function is normally used in tcp_connect or so, which already
provides the address in network byte order. Hence just adjust the
testcase.

Thanks

##### Checklist


sumanthkorikkar:
I tested the ntop in both little endian and big endian and could correctly examine the ipv4 and ipv6 addresses in the right order. So no functional change needed in ntop builtin.

mmisono:
Do BE architecture users use IP address 0.0.0.1, 1.0.0.127, 255.255.0.0, 0.0.255.255?

sumanthkorikkar:
> Do BE architecture users use IP address 0.0.0.1, 1.0.0.127, 255.255.0.0, 0.0.255.255? I mean, do they write IP address reverse order?

Ok. Changed. :)

mmisono (approved):

PR #1654, opened 2020-11-27
Add 'elapsed' to man page

ema:
Include 'elapsed' in the list of builtin variables documented in
bpftrace.8. Also clarify that the variable is not a timestamp but rather
the number of nanoseconds elapsed since bpftrace was initialized.


danobi (approved):
Thanks!

danobi:
Happy 2000th commit too

Issue #1539, opened 2020-09-26
Handle printf literal arguments at compile time


fbs:
For a simple program like:

```
sudo bpftrace -e 'BEGIN { printf("%s %s\n", "abc", "def"); }'
```

We currently emit 140 instructions. 90% of that is just useless string handling (strings are 64 bytes) and moving data to the stack for a perf_event.

The whole thing can be reduced to `BEGIN{ printf("abc def\n") }`, which emits a single 64bit perf event in a total of 14 instructions.

Applies to integer literal and other format strings calls (system) as well

fbs:
Once this lands we can address https://github.com/iovisor/bpftrace/pull/1538/files#r495452630

fbs:
working on this right now, the amount of runtime code some scripts generates get in my way

ajor:
Reopening as this doesn't appear to have been implemented

PR #1655, opened 2020-12-03
Improve AST walking / mutating and pass manager

fbs:
This is a follow up of #1607 / #1603 

---

This adds some of the infrastructure required to easily add new AST walking and manipulating passes. By adding new passes instead of bolting more features into existing ones we can keep the code simpeler. E.g. we could have a pass that checks for statements that are always hold (e.g. `if (1<2)`) and either warn the user or simplify the generated code. To make this easy this adds:

- A `Visit()` function to all visitors to do indirect dispatching. Instead of calling `node->accept(this)` we call `Visit()`. This makes it easier to write a generic visitor (like the node counter)
- A `Mutator()` which visits the current tree and returns a new modified tree. To do this it implements its own virtual dispatch table.
- A pass manager that makes it easier to run passes and do some diagnostics (e.g. print the tree) after each pass.

And some refactoring to make everything a bit easier to read.

#1603 has an example mutator pass that removes all `PositionalParam` nodes from the AST and replaces them with constants which simplifies the semantic analyser and generates efficient code. Its not fully up to date, once this is in I'll finish that one. (it should allow us to remove useless probereads as well). I want to use this for  #1539  too.

The diff is a bit ugly but each commit has details :)


fbs:
Looks like clang doesn't like my code, should be a small fix

fbs:
@danobi this is ready for review. I'll fix the clangformat issue when everyone agrees on this change. My local clangformat doesn't agree with the one we use here on some minor syntax things :/

fbs:
Bump. Like to get this in so we can add some simplifying passes like the `str($1)` case, that generates some horrible code atm.

src/ast/pass_manager.h
```diff
@@ -0,0 +1,129 @@
...
+};
+
+/**
+   AST modifying passes
+ */
+class MutatePass : public Pass
+{
+public:
+  MutatePass(std::string name, PassFPtr fn) : Pass(name, fn){};
+};
```
danobi, line 113: Why does the pass manager need to be aware of _what_ each pass does? I'd expect the pass manager to just manage passes -- what each pass semantically does is up to pass. Furthermore, the caller shouldn't be using a pass unless they know what the pass does. Seems differentiating passes in the pass manager just adds friction.

Am I missing something?

danobi (commented):
LGTM overall

src/ast/pass_manager.h
fbs: Yeah I don't think we really need it. I initially differentiated between them as mutating passes return a new root which needs handling, but ended up handling that by passing an optional arg to Success(). And because it would make it easy to fiilter  out modifying passes when debugging, but as passes are named we could just add a `skip pass` option.

The whole pass manager thing is experimental, not sure what interface will best suite our future needs. Just hoping we can keep it simple so refactoring in the future is easy.

Ill remove it :)

fbs:
any objectiosn to landing this @danobi ? :)

danobi (approved):
Nope, let's get this in

PR #1656, opened 2020-12-04
Array improvements

### LINKED ISSUE #1052: Support `Type::Array` as a map key
```
./src/bpftrace -e 'BEGIN { @[curtask->numa_faults_locality] = 1; }' 
Attaching 1 probe...
^C

@invalid mapkey argument type
zsh: abort      sudo ./src/bpftrace -e 'BEGIN { @[curtask->numa_faults_locality] = 1; }
```
The type type of `numa_faults_locality` is [`unsigned long[3]`](https://github.com/torvalds/linux/blob/738d2902773e30939a982c8df7a7f94293659810/include/linux/sched.h#L1124).

-----

Originally reported in https://github.com/iovisor/bpftrace/pull/1009#discussion_r361824791

### LINKED ISSUE #1057: Can't assign Type::Array value to a local variable
```
% sudo ./src/bpftrace -e 'BEGIN { $a = curtask->numa_faults_locality; @ = $a[0];}' -v
Attaching 1 probe...

Error log:
0: (b7) r6 = 0
1: (73) *(u8 *)(r10 -18) = r6
last_idx 1 first_idx 0
regs=40 stack=0 before 0: (b7) r6 = 0
2: (6b) *(u16 *)(r10 -20) = r6
3: (85) call bpf_get_current_task#35
4: (71) r1 = *(u8 *)(r0 +3482)
R0 invalid mem access 'inv'
processed 5 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

Error loading program: BEGIN
```

This is because [MUL](https://github.com/iovisor/bpftrace/blob/374ca46cac9b3ee8c927c3cc8c926571e4210088/src/ast/codegen_llvm.cpp#L1023) and [FieldAccess](https://github.com/iovisor/bpftrace/blob/374ca46cac9b3ee8c927c3cc8c926571e4210088/src/ast/codegen_llvm.cpp#L1152) do not read if the type is `Type::Array` ([ArrayAccess](https://github.com/iovisor/bpftrace/blob/374ca46cac9b3ee8c927c3cc8c926571e4210088/src/ast/codegen_llvm.cpp#L1203) does) and [AssignVar](https://github.com/iovisor/bpftrace/blob/374ca46cac9b3ee8c927c3cc8c926571e4210088/src/ast/codegen_llvm.cpp#L1294) perform memcpy (not proberead). (Current implementation tries to raed all array contents when assigning value.)

I also found `IRBuilderBPF::GetType` only consider [int8 type](https://github.com/iovisor/bpftrace/blob/374ca46cac9b3ee8c927c3cc8c926571e4210088/src/ast/irbuilderbpf.cpp#L113). (It seems these part only consider` Type::string`.)

viktormalik:
This improves work with constant arrays. Until now, user could only use `[]` to directly access particular array elements.
This PR adds new possibilities to:
- assign an array to a local variable (resolves #1057)
- assign an array to a map
- use an array as a map key (resolves #1052).

If an array is assigned into a variable, operator `[]` can be later used on that variable to access the array elements.

For local variables, only the pointer to the array is stored. For maps, the array is fully copied into the map (or into the map key) since it may be accessed from another probe where the original array may not exist anymore, or may have a different value. I'm not sure if this is a correct approach, feel free to suggest something better.

The last commit also slightly changes the type of `alloca` for map key - instead of always allocating an array of bytes, it allocates the appropriate type (e.g. `i64` instead of `[8 * i8]` for integer keys). This caused some changes in existing codegens.

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -2294,10 +2294,32 @@ AllocaInst *CodegenLLVM::getMapKey(Map &map)
       }
       else
       {
-        key = b_.CreateAllocaBPF(expr->type.GetSize(), map.ident + "_key");
-        b_.CreateStore(
-            b_.CreateIntCast(expr_, b_.getInt64Ty(), expr->type.IsSigned()),
-            b_.CreatePointerCast(key, expr_->getType()->getPointerTo()));
+        key = b_.CreateAllocaBPF(expr->type, map.ident + "_key");
+        if (expr->type.IsArrayTy())
+        {
+          // expr currently contains a pointer to the array
+          if (expr->type.is_internal)
+          {
+            // The array is already in the BPF memory - memcpy it
+            b_.CREATE_MEMCPY(key, expr_, expr->type.GetSize(), 1);
+          }
+          else
+          {
+            // We need to read the entire array and save it
+            b_.CreateProbeRead(ctx_,
+                               key,
+                               expr->type.GetSize(),
+                               expr_,
+                               expr->type.GetAS(),
+                               expr->loc);
+          }
+        }
+        else
+        {
+          b_.CreateStore(
+              b_.CreateIntCast(expr_, b_.getInt64Ty(), expr->type.IsSigned()),
+              b_.CreatePointerCast(key, expr_->getType()->getPointerTo()));
+        }
       }
     }
     else
@@ -2318,15 +2340,29 @@ AllocaInst *CodegenLLVM::getMapKey(Map &map)
         Value *offset_val = b_.CreateGEP(
             key, { b_.getInt64(0), b_.getInt64(offset) });
 
-        if (shouldBeOnStackAlready(expr->type))
+        if (shouldBeOnStackAlready(expr->type) ||
+            (expr->type.IsArrayTy() && expr->type.is_internal))
           b_.CREATE_MEMCPY(offset_val, expr_, expr->type.GetSize(), 1);
         else
         {
-          // promote map key to 64-bit:
-          b_.CreateStore(
-              b_.CreateIntCast(expr_, b_.getInt64Ty(), expr->type.IsSigned()),
-              b_.CreatePointerCast(offset_val,
-                                   expr_->getType()->getPointerTo()));
+          if (expr->type.IsArrayTy())
+          {
+            // Read the array into the key
+            b_.CreateProbeRead(ctx_,
+                               offset_val,
+                               expr->type.GetSize(),
+                               expr_,
+                               expr->type.GetAS(),
+                               expr->loc);
+          }
+          else
+          {
+            // promote map key to 64-bit:
+            b_.CreateStore(
+                b_.CreateIntCast(expr_, b_.getInt64Ty(), expr->type.IsSigned()),
+                b_.CreatePointerCast(offset_val,
+                                     expr_->getType()->getPointerTo()));
+          }
         }
         offset += expr->type.GetSize();
       }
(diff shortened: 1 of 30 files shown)


viktormalik:
CI failed on fetching packages, needs restart

src/ast/codegen_llvm.cpp
```diff
@@ -1672,7 +1672,10 @@ void CodegenLLVM::visit(ArrayAccess &arr)
   size_t element_size = type.GetElementTy()->GetSize();
 
   auto scoped_del_expr = accept(arr.expr);
-  array = expr_;
+  if (expr_->getType()->isPointerTy())
+    array = b_.CreatePtrToInt(expr_, b_.getInt64Ty());
```
danobi, line 1676: Why is this necessary? Doesn't `expr_` already have the right type?

src/ast/codegen_llvm.cpp
```diff
@@ -2275,10 +2275,32 @@ AllocaInst *CodegenLLVM::getMapKey(Map &map)
       }
       else
```
danobi, line 2276: So with this code, you can have arrays as the only key to a map entry. What about support for 2 or more values as a map key?

danobi:
Can you also add some runtime tests for more complex array types? eg an array of structs and nested arrays.

src/ast/codegen_llvm.cpp
viktormalik: Not really, `expr_` has the right (integer) type if the array is read (proberead) from the kernel. On the other hand, if it is loaded from a map, the array has a pointer type.

src/ast/codegen_llvm.cpp
viktormalik: Good point, I'll add support for array used as a part of multi-value key, it should be quite straightforward.

viktormalik:
> Can you also add some runtime tests for more complex array types? eg an array of structs and nested arrays.

Done. It required some small changes, so there's one more commit to review.

I also added support for array as a part of multi-key.

mmisono:
Nice changes, but it seems there is a problem with handling an array in a tuple.

```
% sudo ./src/bpftrace -e 'BEGIN {@a = curtask->numa_faults_locality; @b = (1, curtask->numa_faults_locali
ty);}'
Attaching 1 probe...
^C

@a: [0,7,0]

@b: (1, [18446619453034875792,0,0])
```

viktormalik:
> Nice changes, but it seems there is a problem with handling an array in a tuple.

Right, good observation. In fact, neither structs can be used in tuples. Also, a good implementation wouldn't be exactly simple, so I'd prefer disabling this (array or struct in a tuple) for now and implementing it later.

src/ast/codegen_llvm.cpp
```diff
@@ -1884,7 +1894,14 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
   }
   else
   {
-    b_.CreateStore(expr_, variables_[var.ident]);
+    Value *val = expr_;
+    if (assignment.expr->type.IsArrayTy())
+    {
+      val = b_.CreatePtrToInt(expr_, b_.getInt64Ty());
+      // Since only the pointer is copied, we need to extend lifetime of RHS
+      scoped_del.disarm();
```
danobi, line 1912: Are you sure this is what you want? Lifetime extension is usually done with:
```
expr_deleter_ = scoped_del.disarm();
```

danobi:
> Right, good observation. In fact, neither structs can be used in tuples. Also, a good implementation wouldn't be exactly simple, so I'd prefer disabling this (array or struct in a tuple) for now and implementing it later.

Sure, sounds reasonable. Can you add a check to semantic analyser, then?

src/ast/codegen_llvm.cpp
viktormalik: That will only extend lifetime to the end of the current statement but I need to extend it to the end of the probe (since other statements may want to read from that pointer). I haven't found any other way to do that, so this seemed like the simplest solution.

Anyway, I plan to improve work with structs and arrays (to avoid unnecessary memcpys), so I'll probably come with a cleaner solution, then.

danobi (approved):

mmisono (approved):

PR #1476, opened 2020-08-19
language: support scientific notation for literals

fbs:
As bpf doesn't support floats this is quite simple and e.g. `1e-10`
isn't supported.

```
bpftrace -e 'i:ms:100 { print(1e6); exit(); }'
1000000
```

##### Checklist

- [ ] Language changes are updated in `docs/reference_guide.md` (not sure where to put this)
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x] The new behaviour is covered by tests

PR #1660, opened 2020-12-10
Add DISK field to biosnoop

DavidVentura:
Add disk field to biosnoop.

I tried to also add `req->part->part_no` but I felt I was not getting what I expected on it when writing to certain partitions. 

##### Checklist

### DIFF
--- a/tools/biosnoop.bt
+++ b/tools/biosnoop.bt
@@ -13,7 +13,7 @@
 
 BEGIN
 {
-	printf("%-12s %-16s %-16s %-6s %7s\n", "TIME(ms)", "DISK", "COMM", "PID", "LAT(ms)");
+	printf("%-12s %-7s %-16s %-6s %7s\n", "TIME(ms)", "DISK", "COMM", "PID", "LAT(ms)");
 }
 
 kprobe:blk_account_io_start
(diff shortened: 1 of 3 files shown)


brendangregg:
Thanks for testing on different platforms. I think the field should be 7-chars wide to match the BCC version, plus as this grows to approach the BCC version (if it does) we should keep the default output <80 chars.

tools/biosnoop.bt
```diff
@@ -12,32 +13,36 @@
...
 /@start[arg0] != 0 && @iopid[arg0] != 0 && @iocomm[arg0] != ""/
+
 {
 	$now = nsecs;
-	printf("%-12u %-16s %-6d %7d\n",
-	    elapsed / 1e6, @iocomm[arg0], @iopid[arg0],
-	    ($now - @start[arg0]) / 1e6);
+	printf("%-12u %-7s %-16s %-6d %7d\n",
+	    elapsed / 1000000, @disk[arg0], @iocomm[arg0], @iopid[arg0],
+	    ($now - @start[arg0]) / 1000000);
```
danobi, line 34: What's the reason for the `1e6` -> `1000000` changes?

DavidVentura: my compiler refused to parse 1e6 -- i'll revert this change if you think i should

danobi: You're probably on an older bpftrace. Scientific notation was merged in #1476.

Upstream tools should be targeting upstream bpftrace - we don't guarantee upstream tools are compatible across old versions of bpftrace.

danobi (approved):
Thanks!

PR #1664, opened 2020-12-24
tests: fix semantic_analyser.type_ctx failed on aarch64

xuchunmei000:
char is default to signed on x86_64, but default to unsigned on aarch64,
ppc and s390.

Signed-off-by: Chunmei Xu <xuchunmei@linux.alibaba.com>

##### Checklist

### DIFF
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -1776,11 +1776,17 @@ TEST(semantic_analyser, type_ctx)
   var = static_cast<ast::Variable *>(unop->expr);
   EXPECT_TRUE(var->type.IsPtrTy());
 
+#if ARCH_X86_64
+  auto chartype = CreateInt8();
+#else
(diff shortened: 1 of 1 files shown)


mmisono:
Looks ok, but I think the following is simpler.
```c
#if ARCH_X86_64
auto chartype = CreateInt8()
#else
auto chartype = CreateUInt8()
#endif

EXPECT_EQ(chartype, assignment->var->type);
[...]
```

xuchunmei000:
> Looks ok, but I think the following is simpler.
> 
> ```c
> #if ARCH_X86_64
> auto chartype = CreateInt8()
> #else
> auto chartype = CreateUInt8()
> #endif
> 
> EXPECT_EQ(chartype, assignment->var->type);
> [...]
> ```

yes, and done as you suggested.

danobi (approved):
Thanks! Sorry for delay

PR #1670, opened 2021-01-09
tests: Ubuntu 20.04 no longer has python (python 2) by default

gfx:
This PR changes the use of  `python(1)` (i.e. Python v2.x) to `python3(1)` in tests, because `python(1)` is no longer installed by default in, for example, Ubuntu 20.04 because Python v2.x has been EOL.

##### Checklist

- [ ] <del>Language changes are updated in `docs/reference_guide.md`</del>
- [ ] <del>User-visible and non-trivial changes updated in `CHANGELOG.md`</del>
- [ ] <del>The new behaviour is covered by tests</del>

### DIFF
--- a/tests/runtime/json-output
+++ b/tests/runtime/json-output
@@ -4,52 +4,52 @@ EXPECT ^ERROR: Invalid output format "jsonx"$
 TIMEOUT 5
 
 NAME scalar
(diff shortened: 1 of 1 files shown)


danobi (approved):
Thanks!

PR #1673, opened 2021-01-12
Use kfunc__sched_fork instead of kfunc__strlen for kfunc detection

olsajiri:
The strlen function can be compiled out in some kernels.


danobi:
Can you give it a clang format?

danobi (approved):

PR #1677, opened 2021-01-17
Move LibBpf detection before bcc symbols check

olsajiri:
So we can use detected libbpf in libbcc checks.


mmisono:
The CI failed to find `bcc_prog_load()` for some reasons.

olsajiri:
> The CI failed to find `bcc_prog_load()` for some reasons.

yep, I forgot to put the condition on LIBBPF_FOUND, should be ok now

danobi (approved):

PR #1679, opened 2021-01-25
Add support for iterator probes

olsajiri:
Adding support for following iterator probes:

- iter:task

```
# bpftrace -e 'iter:task { seq_printf("%s:%d\n", ctx->task->comm, ctx->task->pid); }'
Attaching 1 probe...
systemd:1
kthreadd:2
rcu_gp:3
rcu_par_gp:4
kworker/0:0H:6
mm_percpu_wq:8
```

- iter:task_file

```
# bpftrace -e 'iter:task_file { seq_printf("%s:%d %s\n", ctx->task->comm, ctx->task->pid, path(ctx->file->f_path)); }'
Attaching 1 probe...
systemd:1 /dev/null
systemd:1 /dev/null
systemd:1 /dev/null
systemd:1 /dev/kmsg
systemd:1 anon_inode:[eventpoll]
systemd:1 anon_inode:[signalfd]
```
It's also possible to pin the iterator with --pin option:

```
# bpftrace -e 'iter:task_file { seq_printf("%s:%d %s\n", ctx->task->comm, ctx->task->pid, path(ctx->file->f_path)); }' --pin /sys/fs/bpf/files
Attaching 1 probe...
Program pinned to /sys/fs/bpf/files
```
That makes  it available in bpffs like:

```
# cat /sys/fs/bpf/files
systemd:1 anon_inode:inotify
systemd:1 anon_inode:[timerfd]
...
systemd-journal:849 /dev/kmsg
systemd-journal:849 anon_inode:[eventpoll]
...
sssd:1146 /var/log/sssd/sssd.log
sssd:1146 anon_inode:[eventpoll]
...
NetworkManager:1155 anon_inode:[eventfd]
NetworkManager:1155 /var/lib/sss/mc/passwd (deleted)
```

I still need to clear few thing, but would like to get some thoughts/ideas on this,

thanks,
jirka


fbs:
I like it, useful :)

danobi:
Few initial thoughts:
* Can we make `printf` just do the right thing in `iter` probes?
* Not sure I like adding global `--pin` flag for a specific probe type. Is there any way to keep it probe specific?

olsajiri:
> Few initial thoughts:
> 
>     * Can we make `printf` just do the right thing in `iter` probes? Seems like `seq_printf` is an implementation detail -- not sure we need another helper for that.

so seq_printf is new helper that is used in iterator programs

I think we should keep it to be consistent with other iterator programs (bpf selftests)
we could overload 'printf' for iterator probes to call seq_printf, but I think that
would add more confusion

> 
>     * Not sure I like adding global `--pin` flag for a specific probe type. Is there any way to keep it probe specific? Maybe `iter:foo:pin`?

iter:task:pin
iter:task_file:pin

ok, I like that more than the --pin option

thanks

danobi:
> I think we should keep it to be consistent with other iterator programs (bpf selftests) we could overload 'printf' for iterator probes to call seq_printf, but I think that would add more confusion

To clarify, confusion for who? I understand as a bpftrace/kernel developers it might be a bit confusing. But it's not clear to me why we need to expose users to the details. It seems to me `seq_printf` is an implementation detail.

I guess to put it differently, is there any case where a user would need to use both `printf` and `seq_printf` in an `iter` probe?

olsajiri:
> > I think we should keep it to be consistent with other iterator programs (bpf selftests) we could overload 'printf' for iterator probes to call seq_printf, but I think that would add more confusion
> 
> To clarify, confusion for who? I understand as a bpftrace/kernel developers it might be a bit confusing. But it's not clear to me why we need to expose users to the details. It seems to me `seq_printf` is an implementation detail.

my concern was that when iterators are used in bpf selftests and 
will be used in bcc tools, they are/will be used with bpf_seq_printf,
so I wanted to stay consistent

> 
> I guess to put it differently, is there any case where a user would need to use both `printf` and `seq_printf` in an `iter` probe?

also iterators use tracing type programs, so I think it's actually possible to use
perf event ring buffer in them, so we could in theory use classic printf.. but that
does not seems useful at the moment

but let's try it and repent later.. I like that idea, because it's less typing ;-)

danobi:
> my concern was that when iterators are used in bpf selftests and
> will be used in bcc tools, they are/will be used with bpf_seq_printf,
> so I wanted to stay consistent

IIRC we already diverged a little from the kernel with `k[ret]func` and `fentry/fexit` probes.

I remember Andrii was a little confused :). But that being said, I think it's ok if we diverge
a little if it makes things easier for the user. Many bpftrace users I've worked with are
not kernel developers.

src/map.cpp
```diff
@@ -37,7 +37,12 @@ Map::Map(const std::string &name, const SizedType &type, const MapKey &key, int
   if (key_size == 0)
     key_size = 8;
 
-  if (type.IsCountTy() && !key.args_.size())
+  if (type.IsBufferTy())
+  {
+    map_type_ = BPF_MAP_TYPE_ARRAY;
```
fbs, line 42: doesn't this break cases like `@thing[pid] = buf(1,1)`?

docs/reference_guide.md
```diff
@@ -1644,6 +1645,118 @@ fd 3 name libselinux.so.1
 ```
 And as you can see in above example it's also possible to access function arguments on `kretfunc` probes.
 
+## 17. `iter`: Iterators Tracing
+
```
fbs, line 1650: Adding a 

'Kernel: 5.4`

would be helpful as a quick ref, some calls/builtins already have it

src/ast/semantic_analyser.cpp
```diff
@@ -2692,6 +2702,52 @@ int SemanticAnalyser::create_maps_impl(void)
...
+    // and create map with that
+    MapKey key;
+    auto map = std::make_unique<T>("data", type, key, 1);
+    failed_maps += is_invalid_map(map->mapfd_);
+    bpftrace_.maps.Set(MapManager::Type::Data, std::move(map));
+
+    // copy all the formats to data array and store
+    // indexes and lengths so we can later access it
+    uint32_t idx = 0;
+    char *data = new char[size];
```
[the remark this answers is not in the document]
olsajiri: ok, will change

src/map.cpp
olsajiri: it does :-\ I'll add separate type for that

docs/reference_guide.md
olsajiri: sure, will add

src/map.cpp
fbs: Can you add a test case for the buf thing too?

src/bpftrace.cpp
```diff
@@ -288,6 +291,15 @@ int BPFtrace::add_probe(ast::Probe &p)
       probe.mode = attach_point->mode;
       probe.async = attach_point->async;
 
+      if (has_iter)
+      {
+        LOG(ERROR) << "Only one iter probe allowed";
+        return 1;
```
danobi, line 298: And also I don't think codegen_llvm's call to `add_probe()` checks for errors. We should probably make it check and bail on error

src/types.cpp
```diff
@@ -176,6 +176,7 @@ std::string typestr(Type t)
     case Type::tuple:    return "tuple";    break;
     case Type::timestamp:return "timestamp";break;
     case Type::mac_address: return "mac_address"; break;
+    case Type::data:     return "data";     break;
```
danobi, line 179: I'd really prefer to avoid adding new types if we can help it. I think complexity grows exponentially with every new type.

It seems like Type::data is currently used to signal to map creation code it needs a specific type of map. Could you perhaps do that with a new map constructor?

src/bpffeature.cpp
```diff
@@ -134,10 +134,11 @@ bool BPFfeature::detect_helper(enum libbpf::bpf_func_id func_id,
          (strstr(buf, "unknown func ") == nullptr);
 }
 
-bool BPFfeature::detect_prog_type(enum libbpf::bpf_prog_type prog_type)
+bool BPFfeature::detect_prog_type(enum libbpf::bpf_prog_type prog_type,
+                                  const char* name = NULL)
```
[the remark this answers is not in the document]
olsajiri: true, will remove

src/attached_probe.cpp
```diff
@@ -146,6 +153,40 @@ int AttachedProbe::detach_kfunc(void)
...
+{
+  close(linkfd_);
+  return 0;
+}
+#else
+void AttachedProbe::attach_iter(void)
+{
+  throw std::runtime_error(
+      "Error attaching probe: " + probe_.name +
+      ", iter API is not available for linked bpf version");
```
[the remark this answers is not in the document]
olsajiri: right, will fix

src/bpftrace.cpp
```diff
@@ -288,6 +291,15 @@ int BPFtrace::add_probe(ast::Probe &p)
...
+      }
+
+      if (probe.type == ProbeType::iter)
+        has_iter = true;
```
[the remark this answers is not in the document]
olsajiri: I need to find better way for this.. the idea is that you can use only single iterator probe for program, not to be mixed with other probes or iterators

src/bpftrace.cpp
olsajiri: right, I'll check on that

src/types.cpp
olsajiri: The new constructor seems tricky for this.. but perhaps I can reuse Array or Buffer, I'll check, thanks

src/bpftrace.cpp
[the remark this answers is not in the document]
danobi: Maybe check in semantic analysis? Could iterate over all the probes in a program

src/types.cpp
danobi: Thought about it overnight again. It doesn't look like `Type::data` is user visible (ie a user cannot assign a variable to type Type::data). So IMO it should not be a part of the type system.

olsajiri:
the new version tries to address the review comments from @danobi and adds 2 unrelated very small fixes

fbs (approved):

danobi (approved):
Small nit but lgtm

src/mapmanager.h
```diff
@@ -59,6 +59,7 @@ class MapManager
     PerfEvent,
     Join,
     Elapsed,
+    Data,
```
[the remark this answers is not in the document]
olsajiri: ok, first I was thinking this can be generic, but the usage is actually quite specific,
so I think SeqPrintfData is better, will change

olsajiri:
I had to make some changes because of the rebase on current master,
also one extra patch so the libbpf check can use custom kernel headers

olsajiri:
test fails because of this:
The server is temporarily unable to service your request due to maintenance downtime or capacity problems. Please try again later.

let's see later ;-)

danobi:
Yes, see: https://lists.debian.org/debian-infrastructure-announce/2021/03/msg00000.html

fbs:
conflicts again but go ahead and merge right after fixing them!

brendangregg:
Looks like a promising capability, thanks!

PR #1681, opened 2021-01-27
Fix attaching to multiple usdt probe locations

rsanger:
Fixes a bug where all usdt BPF programs were attached to the first probe point
location when there are multiple locations of the same probe (inlining
etc).
Also, modifies the existing test to test the correct probes are installed.

Here is a minimal test example to show this behaviour:
```
// test-probe-loc.c
#include <sys/sdt.h>
int main() {
   DTRACE_PROBE1(test, probe, 1); // probe 1
}
int unused() {
   DTRACE_PROBE1(test, probe, 2); // probe 2
}
```
```
# gcc test-probe-loc.c -o test-probe-loc
# bpftool -e 'usdt:./test:test:probe { printf("%d\n", arg0); }' -c ./test-probe-loc
# tplist-bpfcc -vv -l ./test-probe-loc
b'test':b'probe' [sema 0x0]
  location #1 b'./test-probe-loc' 0x1131
    argument #1 4 signed   bytes @ 1
  location #2 b'./test-probe-loc' 0x1141
    argument #1 4 signed   bytes @ 2
```

Without the fix, ```probe 1``` is hit twice (but with different BPF programs running, hence both 1 and 2 are printed). _Or it might never get called depending on the order that probe locations are listed in test-probe-loc._
```
# bpftrace_v.11.4 -e 'usdt:./test-probe-loc:test:probe { printf("%d\n", arg0); }' -c ./test-probe-loc
Attaching 2 probes...
1
2
```

With the fix, ```probe 1``` is only called once with the correct BPF program
```
# bpftrace_master -e 'usdt:./test-probe-loc:test:probe { printf("%d\n", arg0); }' -c ./test-probe-loc
Attaching 2 probes...
1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,8 @@ and this project adheres to
   - [#1650](https://github.com/iovisor/bpftrace/pull/1650)
 - Fix attaching to usdt probes in shared libraries
   - [#1600](https://github.com/iovisor/bpftrace/pull/1600)
+- Fix attaching to multiple usdt probe locations with the same label
+  - [#1681](https://github.com/iovisor/bpftrace/pull/1681)
 
 #### Tools
 - Hook up execsnoop.bt script onto `execveat` call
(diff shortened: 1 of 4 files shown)


danobi:
Nice fix, thanks! Kind of upset I bungled my "fix".

Just one thing: could you update the CHANGELOG with a Fixed entry?

rsanger:
No problem, I've added a changelog entry.

PR #1684, opened 2021-02-03
Support usdt args with index register and scale

rsanger:
Adds support usdt args with index register and scale.
Adds tests for all support usdt argument types including register+scale
arguments. The tests also verify the correct sign extension of
signed arguments.
Additionally, this patch corrects sign extension of usdt arguments
in the case of constant and addressed arguments.

Currently, I have disabled 64-bit constant tests as these are waiting on https://github.com/iovisor/bcc/pull/3258

##### Checklist

- [x] N/A Language changes are updated in `docs/reference_guide.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to
   - [#1656](https://github.com/iovisor/bpftrace/pull/1656)
 - Add builtin function: `macaddr`
   - [#1647](https://github.com/iovisor/bpftrace/pull/1647)
+- Add support for usdt arguments utilising the index register and scale
+  - [#1684](https://github.com/iovisor/bpftrace/pull/1684)
 
 #### Changed
 - Warn if using `print` on `stats` maps with top and div arguments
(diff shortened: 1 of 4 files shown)


danobi:
Looks reasonable but didn't look very closely yet. Curious: what kind of code generates usdt probes like this?

rsanger:
Only GCC (with optimisation -O1 or higher) seems to generate code with index register and scale components.

On occasion, GCC will generate a probe with an index register and scale component when the argument is an element in an array. It seems rare to encounter as you need a case where 1) both the address of the array and index cannot be inferred, and 2) the element is not already loaded into a register.

```
void update_item(int *array, int index) {
    DTRACE_PROBE1(test, probe1, array[index]);
}
```
As written will trigger it. However, to trigger it in actual code you need the right circumstance where array[index] is not loaded into a register.

danobi (approved):
LGTM

Could you fix up the clang-format error? Thanks

rsanger:
No problem, done, I missed that one.

PR #1697, opened 2021-02-14
bpftrace: Fix build error

sumanthkorikkar:
Add required header to avoid build errors.

The compilation fails for s390x and probably powerpc as well.

Fixes: 290dadea ("watchpoint: Permit specifying function+function_arg
XOR absolute address")

##### Checklist

### DIFF
--- a/src/arch/arch.h
+++ b/src/arch/arch.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


danobi (commented):
LGTM, but I think I only added watchpoint support for x86-64 and aarch64 in 290dadeabc6 .

I mentioned in the commit message:

> Note we only support x86 and arm for now. It should be trivial to add
other archs later (just implement `arch::invalid_watchpoint_modes()`).

sumanthkorikkar:
Ok. Changed the tests. Run it for - x86-64 and aarch64.

PR #1698, opened 2021-02-15
Update INSTALL.md

notmorpheus:
##### Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -192,7 +192,7 @@ set -e
 KERNEL_VERSION="${KERNEL_VERSION:-$(uname -r)}"
 kernel_version="$(echo "${KERNEL_VERSION}" | awk -vFS=- '{ print $1 }')"
 major_version="$(echo "${KERNEL_VERSION}" | awk -vFS=. '{ print $1 }')"
-
+apt-get update
 apt-get install -y build-essential bc curl flex bison libelf-dev
 
 mkdir -p /usr/src/linux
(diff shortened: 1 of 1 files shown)


danobi (changes requested):
Not sure this is correct. According to the text just above this code block:

> this script provides a generic means to get bpftrace kernel headers:

Your change seems to be ubuntu specific and also not correct. I think the idea is to get the headers from the tarball -- not the package manager.

notmorpheus:
> Your change seems to be ubuntu specific and also not correct
This change is debian based. Also, the script is using apt package manager without the change anyway, so I am not sure what you mean by that.

> I think the idea is to get the headers from the tarball -- not the package manager.
I did not know that tarball had some sort of preference in this case. Can I ask why it's not correct to install headers from the package manager, if found?

danobi:
> "Your change seems to be ubuntu specific and also not correct"
> This change is debian based. Also, the script is using apt package manager without the change anyway, so I am not sure what you mean by that.

I didn't write the instructions, but my interpretation is that the apt-get is there to express the fact that a linux kernel build needs those dependencies. I don't believe they imply that the instructions are for debian/ubuntu.

> " I think the idea is to get the headers from the tarball -- not the package manager."
> Did not know that tarball had some sort of preference in this case. Can I ask why it's not correct to install headers from the package manager, if found?

B/c right above the code snippet says the following: `Usually kernels headers can be installed from a system package manager. In some cases though, this may not be an option, and headers aren't easily available.` ... `For older kernels, and on distributions where headers may not be available, this script provides a generic means to get bpftrace kernel headers:`

So the instructions are for when the package manager does **not** contain kernel headers to install. Having steps to use the package manager headers does not make sense in those instructions.

PR #1700, opened 2021-02-17
Remove bpftrace.next_probe_id

omeranson:
The member function 'bpftrace.next_probe_id' is only used to assign new IDs
to probe names. However, this is redundant. Probe IDs are taken from
their location in the vector 'bpftrace.probe_ids_'. This change modifies the
code to take new IDs from the same behaviour.

There are no language, user-visible, or behaviour changes. This is a minor
code improvement.

##### Checklist

- [N/A] Language changes are updated in `docs/reference_guide.md` 
- [N/A] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [N/A] The new behaviour is covered by tests


danobi (approved):
Thanks!

PR #1704, opened 2021-02-22
1263: use filesystem; get_pid_exe

Horki:
##### Checklist

@fbs #1263 small update in utils::get_pid_exe; using filesystem's path instead of `sprintf`, all tests passed, and it's backward compatible

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -151,19 +151,23 @@ bool get_uint64_env_var(const std::string &str, uint64_t &dest)
 
 std::string get_pid_exe(pid_t pid)
 {
-  char proc_path[512];
-  char exe_path[4096];
-  int res;
+  std::error_code ec;
+  std_filesystem::path proc_path{ "/proc" };
+  proc_path /= std::to_string(pid);
+  proc_path /= "exe";
 
-  sprintf(proc_path, "/proc/%d/exe", pid);
-  res = readlink(proc_path, exe_path, sizeof(exe_path));
-  if (res == -1)
+  if (!std_filesystem::exists(proc_path, ec) ||
+      !std_filesystem::is_symlink(proc_path, ec))
     return "";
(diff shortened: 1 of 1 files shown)


src/utils.cpp
```diff
@@ -151,19 +151,27 @@ bool get_uint64_env_var(const std::string &str, uint64_t &dest)
...
+
+    std::string exe_path = std_filesystem::read_symlink(proc_path).string();
+    if (exe_path.length() >= 4096)
+    {
+      throw std::runtime_error(
+          "executable path exceeded maximum supported size of 4096 characters");
+    }
+    return exe_path;
+  }
+  catch (const std_filesystem::filesystem_error &)
```
fbs, line 171: I'd prefer having a `catch runtime_errror &err { throw err }` here to make it clear that you want to propagete the original error. Or maybe use the non-throwing alternatives for these filesystem calls

fbs:
Nice! Small comment and can you squash the commits?

Horki:
Hi @fbs changes have been updated + squashed

src/utils.cpp
```diff
@@ -151,19 +151,27 @@ bool get_uint64_env_var(const std::string &str, uint64_t &dest)
...
+      return "";
+
+    std::string exe_path = std_filesystem::read_symlink(proc_path).string();
+    if (exe_path.length() >= 4096)
+    {
+      throw std::runtime_error(
+          "executable path exceeded maximum supported size of 4096 characters");
+    }
+    return exe_path;
+  }
```
fbs, line 170: don't you still want to catch the `filesystem::error` here? These don't seem to be the non-throwing functions?

Now this whole try/catch can be removed as it throws the only exception it catches, rest is propagated anyway

Horki: @fbs Yes you are right, now I've used `noexcept` methods, and remove try/catch block. but leave 4096 exception

fbs:
Thanks!

PR #1599, opened 2020-11-07
add mips64 reg

George-Hu:
##### Checklist

- [ ] Language changes are updated in `docs/reference_guide.md`
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #1708, opened 2021-02-23
add mips64 reg

fbs:
Closes #1599


fbs:
Tests would be nice but I don't have a machine to run them on. As its still a nice addition / starting point I've added the changelog

PR #1711, opened 2021-03-01
docs: Add a manual / language reference / man page

fbs:
The reference_guide was meant to be a short guide to quickly see how a
feature works. It was never meant to be the place where all technical
details are discussed but as there was no better place that is where it
ended up.

This introduces a new place for all the technical details. It is
written in asciidoc as that, improves plain text readability (slighly
personal) and it is easy exportable to a lot of other formats which
makes it possible to use this for our man-page and website.

----

This is still work in progress but hoping to get some feedback on it.

Open questions:

- do we want to render the asciidoc into man and add it to the repository or do we ask packages to add asciidoctor as a dependency?
- do we want this to be one manpage or should we split it up into `bpftrace` and `bpftrace-language` (i'd prefer one page for now)
- is there a reason to not use asciidoc and go to back to man instead? I chose adoc as its easier to read/write imo and it is readable in github.

TODO:

- finish calls
- finish probes
- go over styling of html and man
- strip details from the ref guide

Note that I don't expect this to be completely finished when the todo items are done. I'm hoping its good enough for a first version so we can expand later. We also have some behaviour that isn't clearly defined, I hope we can define them before we put it in the manpage.


danobi:
> do we want to render the asciidoc into man and add it to the repository or do we ask packages to add asciidoctor as a dependency?

I'd prefer if we only had source files. Less confusion and less stuff to forget to update. Also it looks like most distros have it already: https://pkgs.org/search/?q=asciidoctor

> do we want this to be one manpage or should we split it up into bpftrace and bpftrace-language (i'd prefer one page for now)

One seems fine to me

> is there a reason to not use asciidoc and go to back to man instead? I chose adoc as its easier to read/write imo and it is readable in github.

I'd really rather not write raw man pages. That sounds too painful. adoc seems appealing b/c it can generate reasonable .htmls and man pages.

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+  This flag enables the detection of errors (except for errors from 'probe_read_*').
+  When errors occurs bpftrace will log an error containing the source location and the error code:
+-----
+stdin:48-57: WARNING: Failed to probe_read_user_str: Bad address (-14)
+u:lib.so:"fn(char const*)" { printf("arg0:%s\n", str(arg0));}
+                                                 ~~~~~~~~~
+-----
+
+*-kk*::
+  Similar to '-k' but also includes the errors from 'probe_read_*' helpers.
```
danobi, line 103: I think "same" better implies `-kk` is a superset. WDYT?
```suggestion
  Same to '-k' but also includes the errors from 'probe_read_*' helpers.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+
+=== Process management
+
+*-p* _PID_::
+  'Attach' to the process with _PID_ as ID. If the process terminates bpftrace terminates as well.
```
danobi, line 109: Why is `'Attach'` in quotes?

danobi, line 109: More active phrasing
```suggestion
  'Attach' to the process with _PID_ as ID. If the process terminates, bpftrace will also terminate.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+  When using USDT probes they will be attached to only this process.
+
+*-c* _COMMAND_::
+  Run '_COMMAND_' as a child process.
```
danobi, line 113: Should COMMAND be in quotes? It's already highlighted

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+  When the child terminates bpftrace stops as well, as if 'exit()' has been called.
```
danobi, line 114:
```suggestion
  When the child terminates, bpftrace will also stop (as if 'exit()' has been called).
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+  If bpftrace terminates before the child process does the child process will be terminated with a SIGTERM.
```
danobi, line 115:
```suggestion
  If bpftrace terminates before the child process, the child process will be terminated with a SIGTERM.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+  If used, 'USDT' probes these will only be attached to the child process.
+  To avoid a race condition when using 'USDTs' the child is stopped after 'execve' using 'ptrace(2)' and continued when all 'USDT' probes are attached. +
+  The child PID is available to programs as the 'cpid' builtin. +
+  The child process runs with the same privileges as bpftrace itself (usually root).
+
+*--usdt-file-activation*::
+  activate usdt semaphores based on file path
+
```
danobi, line 123: I suggest just take this out or mention the new alternative. #1567 is much more reliable and does not require the flag. Newer kernels + newer bcc should have it
```suggestion
```

danobi: If you don't take it out, you're missing the capitalization at the beginning and period at the end.

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+=== *Miscellaneous*
+
+*--info*::
+  Print detailed information about features supported by the kernel and the bpftrace build.
+
+*-h, --help*::
+  Print the help summary
```
danobi, line 130:
```suggestion
  Print the help summary.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+*-V, --version*::
+  Print bpftrace version information
```
danobi, line 133:
```suggestion
  Print bpftrace version information.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+*-v*::
+  verbose messages
```
danobi, line 136:
```suggestion
  Turn on verbose logs.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+*-d*::
+  debug mode
```
danobi, line 139:
```suggestion
  Turn on debug logs.
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+*-dd*::
+  verbose debug mode
```
danobi, line 142:
```suggestion
  Turn on all debug logs.
```

man/adoc/bpftrace.adoc
danobi: Also, what does "pid as id" mean? Maybe just "Attach to process PID"?

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+== Environment Variables
+
+Some behaviour can only be controlled through environment variables.
+This section lists all those variables.
+
+=== *BPFTRACE_STRLEN*
+
+Default: 64
+
+Number of bytes allocated on the BPF stack for the string returned by str().
```
danobi, line 153: Do we want to highlight builtins (ie `str()`)? I guess they're keywords so maybe.

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+Make this larger if you wish to read bigger strings with str().
+
+Beware that the BPF stack is small (512 bytes), and that you pay the toll again inside printf() (whilst
```
danobi, line 157: I'm not sure what "and that you pay the toll again inside printf() (whilst it ....)". Is this referring to the string constant or the scratch space needed for submitting an event to the perf event buffer?

If it's the latter, not sure we need to describe implementation details (as it might change later)

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+Number of pages to allocate per CPU for perf ring buffer.
+The value must be a power of 2.
+
+If you're getting a lot of dropped events bpftrace may not be processing events in the ring buffer fast enough.
+It may be useful to bump the value higher so more events can be queued up.
+The tradeoff is that bpftrace will use more memory.
+
+== Supported architectures
+
+x86, arm64 and s390x
```
danobi, line 225: We don't support 32-bit ;)
```suggestion
x86-64, arm64 and s390x
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+
+//TODO: is this true?
```
danobi, line 227: I think we have support for other arch's too. But unsure if it's tested much. Could list some tiers of support here. Maybe x86-64, arm64, and s390x best supported, everything may or may not be well tested, but we'll take bug fixes.

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1500 @@
...
+|uint64
+|Unsigned 64 bit integer
+
+|int64
+|Signed 64 bit integer
+|===
+
+==== Floating-point
+
+Floating-point numbers are not supported by BPF and therefor not by bpftrace.
```
danobi, line 350:
```suggestion
Floating-point numbers are not supported by BPF and therefore not by bpftrace.
```
Actually I think they're adding support: https://lore.kernel.org/bpf/20210226202256.116518-1-iii@linux.ibm.com/ . Unsure what it actually enables, though

danobi (commented):
Will review some more later

brendangregg:
Yes, the reference guide is intended to be brief. As a general rule, I don't like to see a wall of text anywhere: there should be at least one heading visible on the screen at all times. This forces all sections to be less than one screenful.

I wouldn't delete things entirely from the reference guide, as it'll still be the number one place people do searches on. So things like BPFTRACE_STRLEN should still be mentioned, even if it's just a list of keywords and a reference. E.g., "For environment variables BPFTRACE_STRLEN, ..., ..., ..., see the xxx.doc". Then people can still search on it.

I don't understand bpftrace.adoc: it says its a man page. I think that's the right place for man page things, but not "all technical details." Right now we have:

- Reference guide: brief summary of everything
- Tool example files: where you can discuss anything about the tools.
- internal_development.md: where you can discuss anything about the code.

So if we have a bunch of technical details including code snippits or command outputs that doesn't fit into those locations, then we can create another file for it. I hope we can come up with a name that makes sense. I'm not sure what though. The "bpftrace Appendix?" The "bpftrace Encyclopedia"?

It could also be a more of a community developed wiki, where people are free to dump stuff in there that might be useful for searching later on. Whereas we carefully review what goes into the official reference guide.

fbs:
> Yes, the reference guide is intended to be brief. As a general rule, I don't like to see a wall of text anywhere: there should be at least one heading visible on the screen at all times. This forces all sections to be less than one screenful.
> 
> I wouldn't delete things entirely from the reference guide, as it'll still be the number one place people do searches on. So things like BPFTRACE_STRLEN should still be mentioned, even if it's just a list of keywords and a reference. E.g., "For environment variables BPFTRACE_STRLEN, ..., ..., ..., see the xxx.doc". Then people can still search on it.

Yes i want to link to the env section of the manpage

> I don't understand bpftrace.adoc: it says its a man page. I think that's the right place for man page things, but not "all technical details." Right now we have:
> 
>     * Reference guide: brief summary of everything
> 
>     * Tool example files: where you can discuss anything about the tools.
> 
>     * internal_development.md: where you can discuss anything about the code.
> 
> 
> So if we have a bunch of technical details including code snippits or command outputs that doesn't fit into those locations, then we can create another file for it. I hope we can come up with a name that makes sense. I'm not sure what though. The "bpftrace Appendix?" The "bpftrace Encyclopedia"?

The manpage  I wrote documents:

- flags
- env vars
- all the language details (builtins, calls, probetypes, types & pointers) 
- all the details about using bpftrace, address space, etc

It doesn't say anything about the internals or about the tools we ship, we have separate docs for it.

> It could also be a more of a community developed wiki, where people are free to dump stuff in there that might be useful for searching later on. Whereas we carefully review what goes into the official reference guide.

Sounds good

fbs:
thanks @danobi

man/adoc/bpftrace.adoc
fbs: this is copy/pasta from the refguide, ill just leave it out :)

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1900 @@
...
+}
+----
+
+This script has two action blocks and a total of 3 probes.
+The first action block uses the special `BEGIN` probe, which fires once during `bpftrace` startup.
+This probe is used to print a header, indicating that the tracing has started.
+
+The second action block uses two probes, one for `open` and one for `openat`, and defines an action that prints the file being `open`ed as well as the `pid` and `comm` of the process that execute the syscall.
+See the <<Probe types>> section for details on the available probe types.
+
```
fbs, line 291: Feel like more words describing actions should go here, but i lack inspration

fbs:
Took a while but I think im almost there. I still need to do the details of a few probe types but other than that I think its good enough for a first version.

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,1900 @@
...
+== Functions
+
+Functions that are marked *async* are asynchronous which can lead to unexpected behaviour, see the 'sync and async' section for more information. // TODO: LINK
+
+*compile time* functions are evaluated at compile time, a static value will be compiled into the program.
+
+*unsafe* functions can have dangerous side effects and should be used with care, the `--unsafe` flag is required for use.
+
+=== buf_t buf(void * data, [int64 length])
```
fbs, line 1118: this seems to break cross references. I think doing

```
=== buf

.variants
* `buf_t buf...
```

etc for all functions is better

man/adoc/bpftrace.adoc
fbs: this seems to be btf only

fbs:
Added the int syntax from #1900

fbs:
I fixed the section naming thing so linking is easier now. I uploaded a html version as well http://dump.madhax.nl/bpftrace.html#functions-reg , this is what it would look like if we put it on bpftrace.org

fbs:
Ready for review @danobi @viktormalik @brendangregg . There is always more to add but I think its good enough for a first man page

```
$ asciidoctor  man/adoc/bpftrace.adoc
$ open man/adoc/bpftrace.html
$ asciidoctor man/adoc/bpftrace.adoc  -b manpage
$ man man/adoc/bpftrace.8
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,2213 @@
...
+| get_prandom_u32
+| Random number
+
+| retval
+| int64
+| n/a
+| n/a
+| Value returned by the function being traced (kretprobe, uretprobe, kretfunc)
+
+| `arg0`, `arg1`, `...argn`
```
fbs, line 933:
```suggestion
| `sarg0`, `sarg1`, `...sargn`
```

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,2213 @@
...
+.Supported probes
+* kprobe
+* uprobe
+
+Get the contents of the register identified by `name`.
+Valid names depend on the CPU architecture:
+
+[cols="~,~"]
+|===
+|*Arch*
```
fbs, line 1346: needs updating, maybe leave it out?

viktormalik (commented):
Just a bunch of formatting/spelling suggestions. The content itself looks good.

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,2213 @@
...
+////
+
+:idprefix: _
+
+== Sync and Async
+
+While BPF in the kernel can do a lot there are still things that can only be done from user space, like the outputting (printing) of data.
+The way bpftrace handles this is by sending events from the BPF program which user-space will pick up some time in the future (usually in milliseconds).
+
+This can lead to some unexpected behavior,
```
danobi, line 798: There should be more here, right?

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,2213 @@
...
+Number of pages to allocate per CPU for perf ring buffer.
+The value must be a power of 2.
+
+If you're getting a lot of dropped events bpftrace may not be processing events in the ring buffer fast enough.
+It may be useful to bump the value higher so more events can be queued up.
+The tradeoff is that bpftrace will use more memory.
+
+== Supported architectures
+
+x86_64, arm64 and s390x
```
danobi, line 223: Make these bullet points instead?

man/adoc/bpftrace.adoc
```diff
@@ -0,0 +1,2213 @@
...
+}
+----
+
+----
+\x00\x03\x00\x00\x00\x00\x00\x00
+\xc2\x02\x00\x00\x00\x00\x00\x00
+----
+
+[#functions-cat]
+=== void cat(string namefmt, [...args])
```
danobi, line 1171: When compiled to html, I find the full function signature in the heading a ugly to read in the table of contents. Any thoughts on leaving the heading `cat()` and providing full signature(s) in the body?

danobi (commented):
Also I think we need a cmake rule to generate man page using asciidoctor + install it

fbs:
Addressed the comments.

- all typos fixes are applied
- new strftime has been added
- function heading has changed from buf_t buf(...) to just `buf`
- all other small comments have been addressed too


If there are no objections im hoping to merge this as is. CI integration etc can be done in a follow up. Once this is in we can update it as part of PRs instead of having to sync this PR the whole time :)

man/adoc/bpftrace.adoc
fbs: yeah, applied

danobi:
Yeah, let's get this in now so we can all commit changes to it. Looks fairly complete already

danobi (approved):

PR #1713, opened 2021-03-01
1263: replaced all stats in utils.cpp

Horki:
##### Checklist

@fbs Hi, replaced all `stats` in utils.cpp; please test

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -301,12 +301,9 @@ std::vector<std::string> get_kernel_cflags(
 
 bool is_dir(const std::string& path)
 {
-  struct stat buf;
-
-  if (::stat(path.c_str(), &buf) < 0)
-    return false;
-
-  return S_ISDIR(buf.st_mode);
+  std::error_code ec;
+  std_filesystem::path buf{ path };
+  return std_filesystem::is_directory(buf, ec);
 }
 
 namespace {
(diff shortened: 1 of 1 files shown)


Horki:
part of #1263

src/utils.cpp
```diff
@@ -604,62 +604,46 @@ namespace, it will throw an exception
...
   }
 
-  if (fstat(self_fd, &self_stat))
+  if (target_path.string().length() >= 64)
   {
-    errmsg << "fstat(self_fd): " << strerror(errno);
-    goto error;
+    throw MountNSException(
+        "Failed to compare mount ns with PID " + std::to_string(pid) +
+        ". The error was Reading mountNS would overflow buffer.");
```
danobi, line 634: Do we still need this check? Isn't the 64 length limit b/c of `char buf[64]` which is deleted now?

Horki: @danobi Hi, just want to be backward-compatible, so the same exceptions are thrown

Horki: @danobi Hi, I have removed the overflow exception

danobi: > Hi, just want to be backward-compatible, so the same exceptions are thrown

I don't think anyone will mind if there are less errors than before :)

danobi (approved):

danobi:
Thanks!

PR #1714, opened 2021-03-02
1263: fs; bfd_disasm

Horki:
##### Checklist

@fbs related to #1263 for `bfd_disasm`

### DIFF
--- a/src/bfd-disasm.cpp
+++ b/src/bfd-disasm.cpp
@@ -1,7 +1,5 @@
-#include <cstdlib>
-#include <cstring>
 #include <fcntl.h>
-#include <iostream>
+#include <memory>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -9,6 +7,7 @@
 // PACKAGE is defined. Some distros patch this check out.
 #define PACKAGE "bpftrace"
 #include "bfd-disasm.h"
+#include "utils.h"
 #include <bcc/bcc_elf.h>
 #include <bcc/bcc_syms.h>
 #include <bfd.h>
@@ -34,18 +33,6 @@ BfdDisasm::~BfdDisasm()
     close(fd_);
 }
 
-static void get_exec_path(char *tpath, size_t size)
-{
-  const char *path = "/proc/self/exe";
-  ssize_t len;
-
-  len = readlink(path, tpath, size - 1);
-  if (len < 0)
-    len = 0;
-
-  tpath[len] = 0;
-}
-
 static int fprintf_nop(void *out __attribute__((unused)), const char *fmt __attribute__((unused)), ...)
 {
   return 0;
@@ -55,13 +42,11 @@ static AlignState is_aligned_buf(void *buf, uint64_t size, uint64_t offset)
 {
   disassembler_ftype disassemble;
   struct disassemble_info info;
-  char tpath[4096];
+  std::string tpath = get_pid_exe("self");
   bfd *bfdf;
 
-  get_exec_path(tpath, sizeof(tpath));
-
-  bfdf = bfd_openr(tpath, NULL);
-  if (bfdf == NULL)
+  bfdf = bfd_openr(tpath.c_str(), nullptr);
+  if (bfdf == nullptr)
     return AlignState::Fail;
 
   if (!bfd_check_format(bfdf, bfd_object))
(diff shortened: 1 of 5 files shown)


src/bfd-disasm.cpp
```diff
@@ -16,34 +14,17 @@
...
-  fd_ = open(path.c_str(), O_RDONLY);
-
-  if (fd_ >= 0) {
-    struct stat st;
-
-    if (fstat(fd_, &st) == 0)
-      size_ = st.st_size;
-  }
+  std::error_code ec;
+  size_ += std_filesystem::file_size(fs_, ec);
```
danobi, line 20: Can we check `ec` for errors just in case? Don't see anything on cppreference on return value if there's an error.

src/bfd-disasm.h
```diff
@@ -1,19 +1,27 @@
...
 #include "disasm.h"
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace std_filesystem = std::filesystem;
+#elif __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace std_filesystem = std::experimental::filesystem;
+#else
+#error "neither <filesystem> nor <experimental/filesystem> are present"
+#endif
```
danobi, line 12: Can this be extracted out into another file, maybe `filesystem_wrapper.h`? I see another copy paste of this in `utils.cpp`. Would be great to not repeat this and also avoid using transitively included code.

src/bfd-disasm.cpp
```diff
@@ -16,34 +14,17 @@
...
-  ssize_t len;
-
-  len = readlink(path, tpath, size - 1);
-  if (len < 0)
-    len = 0;
-
-  tpath[len] = 0;
+  std::error_code ec;
+  std_filesystem::path path{ "/proc/self/exe" };
+  return std_filesystem::read_symlink(path, ec);
```
danobi, line 27: Same as above

src/bfd-disasm.h
Horki: @danobi sure

src/bfd-disasm.cpp
Horki: @danobi sure

src/filesystem_wrapper.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "disasm.h"
```
danobi, line 3: This shouldn't be here

src/bfd-disasm.cpp
```diff
@@ -114,7 +89,7 @@ AlignState BfdDisasm::is_aligned(uint64_t offset, uint64_t pc)
   uint64_t size = std::min(pc + 100, size_);
   void *buf;
 
-  if (fd_ < 0)
+  if (!std_filesystem::exists(fs_))
```
danobi, line 92: If we're just going to open the fd later again I think it's better we use `open()` to avoid TOCTOU issues.

Horki: @danobi replaced with buffer as a unique pointer, and `ifstream` instead of C-style open, pread

src/filesystem_wrapper.h
Horki: @fbs @danobi Hi, removed with the latest commit

src/bfd-disasm.cpp
```diff
@@ -1,49 +1,27 @@
...
 #include <bcc/bcc_syms.h>
 #include <bfd.h>
 #include <dis-asm.h>
 
 namespace bpftrace {
 
-BfdDisasm::BfdDisasm(std::string &path) : size_(0)
+BfdDisasm::BfdDisasm(std::string &path) : fs_{ path }, size_{ 0 }
 {
-  fd_ = open(path.c_str(), O_RDONLY);
```
danobi, line 21: I strongly feel we should leave the FD based stuff in. This is an easily preventable TOCTOU bug waiting to happen that nobody enjoys debugging.

src/bfd-disasm.cpp
danobi: I do not believe that solves the problem

Horki: @danobi Hi, revered to FD

src/bfd-disasm.h
```diff
@@ -7,13 +7,13 @@ namespace bpftrace {
...
 public:
-  BfdDisasm(std::string &path);
+  BfdDisasm(std::string &);
   ~BfdDisasm();
 
   AlignState is_aligned(uint64_t offset, uint64_t pc);
 
 private:
-  int fd_ = -1;
+  int fd_;
```
danobi, line 16: Please revert these changes -- no point creating noise in git blame

Horki: @danobi reverted

danobi:
Needs a rebase

Horki:
@danobi hi, rebased

fbs:
We try to avoid merge commits in our, it makes history confusing and can make diffs weird like happens here. What I usually do is adding a `--rebase` to the pull command to have it rebase instead of merge. It can be bit more work but it keeps history clean.

The PR is currently unmergeable due to this, can you fix it? One way to do it is by:

```
# replace upstream with w/e your remote is
$ git fetch upstream
$ git checkout -b 1263_bfd_disasm_v2 1263_bfd_disasm  #easy backup just in case
$ git reset upstream/master
$ git add <your changes>
$ git commit -c 30e95232830919a9ebfc94cf63fce0d4e5495c70
$ git push yourreporemotename 1263_bfd_disasm_v2:1263_bfd_disasm --force-with-lease
```

Sorry for the noise, thanks!

PR #1715, opened 2021-03-02
Change pipe to eventfd in ChildProc

inwardvessel:
Using a pipe as a signaling mechanism between parent and child works fine as is. The thought here is to save a file descriptor by using an eventfd instead.

##### Checklist

- [x] Language changes are updated in `docs/reference_guide.md` (n/a)
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md` (n/a)
- [x] The new behaviour is covered by tests (covered within "childproc" suite)

### DIFF
--- a/src/child.cpp
+++ b/src/child.cpp
@@ -150,7 +150,8 @@ ChildProc::ChildProc(std::string cmd)
 
 ChildProc::~ChildProc()
 {
-  if (child_event_fd_ >= 0) {
+  if (child_event_fd_ >= 0)
+  {
     close(child_event_fd_);
   }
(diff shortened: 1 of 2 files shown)


danobi (approved):
LGTM, thanks!

Will merge in a few days if noone has any objections

danobi:
In case anyone else was also wondering:

> eventfd()  is  available  on  Linux  since kernel 2.6.22.  Working support is provided in glibc since version 2.8

which is well before ebpf was added to the kernel.

fbs (approved):

PR #1716, opened 2021-03-03
LLVM ORCv2 + LLVM12

fbs:
This adds orcv2 support which we need for llvm12.

ORCv2 was introduced with LLVM 7 so I was hoping we could ditch llvm6 and only do orcv2 but a lof of llvm apis have changed and making it all compatible is a lot of pain for no gain. Instead I've tried to keep the API for both somewhat consistent and remove the ifdef mess by putting the implementation in separate files.

It's not great. I have some plans for future refactors but as llvm12 might fix that memory bug I'd rather have this in a bit ugly than not at all.


src/ast/codegen_llvm.cpp
```diff
@@ -2966,8 +2954,26 @@ void CodegenLLVM::optimize()
 std::unique_ptr<BpfOrc> CodegenLLVM::emit(void)
 {
   assert(state_ == State::OPT);
-  orc_->compileModule(move(module_));
+  orc_->compile(move(module_));
   state_ = State::DONE;
+
+#ifdef LLVM_ORC_V2
```
fbs, line 2960: so because I cannot figure out how to tell llvm to materialize/codegen the whole module right away this kludge triggers the compile by calling lookup. It looks like llvm compiles the whole module when you ask for one func, so we can exit early.
For this to work the probe_ variables need to be public in bpftace. 

I have some ideas to clean this up in the future, by dumping the function list in the module and then asking it to materialize that. I also want to move the probe and attach logic out of bpftrace to a separate module, so Ill tackle both in a follow up.

Unless someone knows how to emit the code right away without creating a `MaterializationResponsibility` that list all symbol we need.

src/ast/codegen_llvm.cpp
```diff
@@ -19,39 +19,25 @@
...
 #include <llvm/Transforms/IPO/PassManagerBuilder.h>
 
 namespace bpftrace {
 namespace ast {
 
+static std::unique_ptr<BpfOrc> CreateOrc()
+{
+  return BpfOrc::Create();
+}
+
```
danobi, line 32: Is this function necessary?

src/bpforc.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  return std::make_unique<BpfOrc>(TM);
+}
+
+std::optional<std::tuple<uint8_t *, uintptr_t>> BpfOrc::getSection(
+    const std::string &name) const
+{
+  auto sec = sections_.find(name);
+  if (sec == sections_.end())
+    return {};
```
danobi, line 36: Nit: I prefer `std::nullopt` b/c it's more obvious what will happen

danobi (commented):
I'm not too familiar with LLVM APIs but the overall design of this PR looks good to me.

src/ast/codegen_llvm.cpp
fbs: oops, no

src/bpforc.cpp
fbs: ah ogood one

fbs:
ah nice the CI is crashing due to a memleak, need to investigate

fbs:
Apparently the memleaks were just my system, I cannot reproduce them in a different VM anymore.

The segv the CI sees I can, it happens <20% of the time and only in `codegen.struct*`, which are also the tests that don't pass the llvm-as test  as the have some invalid IR in them. It looks like the bug is introduced by `llvm: Move initialization logic to orc class` but I'm not sure why. Running just 1 IR test 100x in a row never triggers it, running in GDB never triggers it and the stacktrace I managed to get with a segv handler looks like:

```
#3  0x000056546d17ae01 in segvHandler (x=11) at /mnt/volume_ams3_01/bpftrace/tests/main.cpp:9
#4  <signal handler called>
#5  0x00007f49fce5c92c in llvm::Type::isSizedDerivedType(llvm::SmallPtrSetImpl<llvm::Type*>*) const () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#6  0x00007f49fd67c426 in ?? () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#7  0x00007f49fd632902 in ?? () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#8  0x00007f49fd6338bc in ?? () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#9  0x00007f49fd63482c in llvm::InstructionCombiningPass::runOnFunction(llvm::Function&) () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#10 0x00007f49fce1fd76 in llvm::FPPassManager::runOnFunction(llvm::Function&) () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#11 0x00007f49fce1fff3 in llvm::FPPassManager::runOnModule(llvm::Module&) () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#12 0x00007f49fce204a0 in llvm::legacy::PassManagerImpl::run(llvm::Module&) () from /usr/lib/llvm-10/lib/libLLVM-10.so.1
#13 0x000056546d33a409 in bpftrace::ast::CodegenLLVM::optimize (this=0x7ffeda107800) at /mnt/volume_ams3_01/bpftrace/src/ast/codegen_llvm.cpp:2949
#14 0x000056546d2172d9 in bpftrace::test::codegen::test (bpftrace=...,
    input="struct Foo { short x; }kprobe:f{  $foo = *(struct Foo*)arg0;  @x = $foo.x;}", name="struct_short_1")
``` 

But ubuntu seems to be missing the debug symbols I need to dig deeper.

We've seen weird llvm crashes before due to invalid IR so I'm going to test this change set rebased on #1705, if that fixes it we might just have to land that one first.

fbs:
Rebased on master, I'll rerun CI a couple of times to make sure it doesn't segv again

danobi (approved):
I've been rebased on this PR for a few days now and it works great.

PR #1719, opened 2021-03-06
bpftrace: Add kprobe, uprobe and DEBUG_FS configs to kernel options

seehearfeel:
Add the configs about kprobe and uprobe to Linux Kernel Requirements,
and also add CONFIG_DEBUG_FS to be able to read and write the files
under /sys/kernel/debug/tracing.

Signed-off-by: Tiezhu Yang <yangtiezhu@loongson.cn>


danobi (approved):
Thanks!

Issue #1305, opened 2020-04-24
Reading pointers off stack broken


danobi:
### bpftrace --info
```
System
  OS: Linux 5.6.4-arch1-1 #1 SMP PREEMPT Mon, 13 Apr 2020 12:21:19 +0000
  Arch: x86_64

Build
  version: v0.10.0-63-g4b37
  LLVM: 10
  foreach_sym: yes
  unsafe uprobe: no
  btf: no
  bfd: yes
  bpf_attach_kfunc: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
```
### What reproduces the bug?
This works:
```
#include <linux/fs.h>
#include <linux/blk_types.h>
#include <linux/blkdev.h>
#include <linux/blk-cgroup.h>
#include <linux/cgroup-defs.h>
#include <linux/sched.h>

kprobe:blk_update_request
{
        $req = (struct request *)arg0;
        printf("cgroup name: %p\n", $req->bio->bi_blkg->blkcg->css.cgroup->kn->parent);
}
```
with output like this:
```
cgroup name: 0xffffa02fdd173480
cgroup name: 0xffffa02fdd173480
cgroup name: (nil)
cgroup name: (nil)
```

But this script does not work:
```
#include <linux/fs.h>
#include <linux/blk_types.h>
#include <linux/blkdev.h>
#include <linux/blk-cgroup.h>
#include <linux/cgroup-defs.h>
#include <linux/sched.h>

kprobe:blk_update_request
{
        $req = (struct request *)arg0;
        $kn = $req->bio->bi_blkg->blkcg->css.cgroup->kn;
        printf("cgroup name: %p\n", $kn->parent);
}
```

with output like:
```
cgroup name: (nil)
cgroup name: (nil)
cgroup name: (nil)
cgroup name: (nil)
cgroup name: (nil)
cgroup name: (nil)
cgroup name: (nil)
cgroup name: (nil)
[...]

PR #1737, opened 2021-03-17
Llvm12 default

fbs:
This makes the LLVM12 release the default image on quay.io as it should fix the allocation bug (#1305).

Needs some careful testing before we do it.

### DIFF
--- a/docker/scripts/push.sh
+++ b/docker/scripts/push.sh
@@ -5,7 +5,7 @@ set -e
 # You must run login.sh before running this script.
 
 DEFAULT_DOCKER_REPO="quay.io"
-DEFAULT_RELEASE_TARGET="vanilla_llvm+clang+glibc2.27"
+DEFAULT_RELEASE_TARGET="vanilla_llvm12+clang+glibc2.27"
 
 # Currently only support pushing to quay.io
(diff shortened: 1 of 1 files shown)


.github/workflows/embedded.yml
```diff
@@ -52,6 +52,21 @@ jobs:
...
+          STATIC_LINKING: ON
+          STATIC_LIBC: OFF
+          EMBED_LLVM: ON
+          EMBED_CLANG: ON
+          EMBED_LIBCLANG_ONLY: OFF
+          EMBED_BCC: OFF
+          EMBED_LIBELF: OFF
+          EMBED_BINUTILS: OFF
+          RUN_ALL_TESTS: 1
+          RUNTIME_TEST_DISABLE: builtin.cgroup,probe.kprobe_offset_fail_size
```
danobi, line 67: Didn't your other PR remove `builtin.cgroup`?
```suggestion
          RUNTIME_TEST_DISABLE: probe.kprobe_offset_fail_size
```

danobi (approved):

.github/workflows/embedded.yml
fbs: yes, looks like i forgot to update this flie.

Issue #1507, opened 2020-09-09
Define pointer arithmetic


fbs:
Currently adding to pointers doesn't work:

```
stdin:1:37-38: ERROR: The + operator can not be used on expressions of types cast, integer
BEGIN { $a = (uint16*) 123; $b = $a + 5; }
```

In C adding to a pointer uses the pointee size to increment, so:

```
int *c = (int *)0;
printf("%lx %lx\n", c, ++c);
```

prints:

```
0 4
```

I think we should stick to that behaviour.

fbs:
One issue:

loads.bt currently does this:

```
	$avenrun = kaddr("avenrun");
	$load1 = *$avenrun;
	$load5 = *($avenrun + 8);
	$load15 = *($avenrun + 16);
```

This works because kaddr returns an uint64 while it should return a pointer(uint64* by default). If we we're to fix that and implement pointer arithmetic this would have to be changed to `$(avenrun + 1)`. Which means breaking existing scripts. Looking at existing tools the 2 I've found that rely on int like arithmetic for kaddr are:

```
originals/Ch14_Kernel/loads.bt
34:	$avenrun = kaddr("avenrun");
35-	$load1 = *$avenrun;

originals/Ch11_Security/eperm.bt
24:	@EACCESS[comm, ksym(*(kaddr("sys_call_table") + args->id * 8))] =
25-	    count();
--
31:	@EPERM[comm, ksym(*(kaddr("sys_call_table") + args->id * 8))] =
32-	    count();
```

So changing it might not be too bad. Kernel symbols are not a stable api so they might have to change over time anyway.


If we don't switch `kaddr` to a pointer we would have to support dereferencing integers,  `$a= *123` would be treated as `$a = * kptr((uint64*) 123)`. Which feels wrong and might lead to subtle bugs, either accidentally using an int when wanting a pointer or a probe_read that doesn't work.
Another downside is that  that if we ever get to option to infer the size of a kaddr (like we do for uaddr) we cannot do that as it's not a pointer type.

fbs:
linking this to #1091.

fbs:
@brendangregg @danobi @mmisono

danobi:
In general, making `kaddr()` return a `uint64 *` makes a lot of sense tome. However, a compatibility break here could easily cause previously-ok scripts to return bogus data. It's hard to say how much stuff in the wild relies on the current behavior, though.

[...]

PR #1739, opened 2021-03-18
Array indexing on pointers

### LINKED ISSUE #1717: Support array indexing on pointers
Array indexing is only supported on constant length arrays. I can't think of any reasons why we can't let users index through a pointer as long as the pointer isn't `void *`.

Support array index through pointers.

eg.
```
$p = (uint32 *)arg0;
$val = $p[5];
```

You can currently workaround it by doing pointer math and printing individual elements.

Or you can declare a dummy struct with a constant length array, eg:
```
struct Foo {
  uint32_t arr[16];
}

... {
  $p = (struct Foo *)arg0;
  $val = $p->arr[5];
}
```

viktormalik:
Resolves #1717.

The generated codegen is exactly the same as for indexing on arrays. This basically only enables the `[]` operator on pointer type (except for `void *`) in semantic analyser.

This is closely related to #1507 which has not been fully resolved, yet, but since this works the same way as array indexing, it should not be a problem.

##### Checklist

### DIFF
--- a/src/ast/codegen_llvm.cpp
+++ b/src/ast/codegen_llvm.cpp
@@ -1623,7 +1623,8 @@ void CodegenLLVM::visit(FieldAccess &acc)
 void CodegenLLVM::visit(ArrayAccess &arr)
 {
   SizedType &type = arr.expr->type;
-  auto elem_type = *type.GetElementTy();
+  auto elem_type = type.IsArrayTy() ? *type.GetElementTy()
+                                    : *type.GetPointeeTy();
   size_t elem_size = elem_type.GetSize();
 
   auto scoped_del_expr = accept(arr.expr);
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -1385,22 +1385,31 @@ void SemanticAnalyser::visit(ArrayAccess &arr)
   SizedType &indextype = arr.indexpr->type;
 
   if (is_final_pass()) {
-    if (!type.IsArrayTy())
+    if (!type.IsArrayTy() && !type.IsPtrTy())
     {
-      LOG(ERROR, arr.loc, err_)
-          << "The array index operator [] can only be used on arrays, found "
-          << type.type << ".";
+      LOG(ERROR, arr.loc, err_) << "The array index operator [] can only be "
+                                   "used on arrays and pointers, found "
+                                << type.type << ".";
       return;
+    }
+
+    if (type.IsPtrTy() && type.GetPointeeTy()->GetSize() == 0)
+    {
+      LOG(ERROR, arr.loc, err_) << "The array index operator [] cannot be used "
+                                   "on a pointer to an unsized type (void *).";
     }
 
     if (indextype.IsIntTy() && arr.indexpr->is_literal)
     {
-      Integer *index = static_cast<Integer *>(arr.indexpr);
+      if (type.IsArrayTy())
+      {
+        Integer *index = static_cast<Integer *>(arr.indexpr);
 
-      if ((size_t)index->n >= type.GetNumElements())
-        LOG(ERROR, arr.loc, err_)
-            << "the index " << index->n
-            << " is out of bounds for array of size " << type.GetNumElements();
+        if ((size_t)index->n >= type.GetNumElements())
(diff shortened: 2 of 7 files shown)


viktormalik:
There's a problem with downloading files from salsa.debian.org, which makes embedded builds to fail. They'll need restart once the  site is available again.

src/ast/semantic_analyser.cpp
```diff
@@ -1385,30 +1385,44 @@ void SemanticAnalyser::visit(ArrayAccess &arr)
...
-            << " is out of bounds for array of size " << type.GetNumElements();
+        if ((size_t)index->n >= type.GetNumElements())
+          LOG(ERROR, arr.loc, err_) << "the index " << index->n
+                                    << " is out of bounds for array of size "
+                                    << type.GetNumElements();
+      }
     }
     else {
       LOG(ERROR, arr.loc, err_) << "The array index operator [] only "
                                    "accepts literal integer indices.";
```
danobi, line 1416: I'm ok with this restriction for both constant pointers and arrays for now but perhaps in the future we can allow expressions as the array index. And if we happen to know both the size of the array and the index, we can warn if they're reading out of bounds.

For stuff like this I'd prefer to trust the user. But again, we can do it if someone asks later.

src/ast/codegen_llvm.cpp
```diff
@@ -1623,7 +1623,8 @@ void CodegenLLVM::visit(FieldAccess &acc)
 void CodegenLLVM::visit(ArrayAccess &arr)
 {
   SizedType &type = arr.expr->type;
-  auto elem_type = *type.GetElementTy();
+  auto elem_type = type.IsArrayTy() ? *type.GetElementTy()
+                                    : *type.GetPointeeTy();
```
danobi, line 1627: Beautiful

danobi (approved):

src/ast/semantic_analyser.cpp
odinuge: Doing dynamic index would be awesome! I have a script where I loop through all cpus on the machine, and prints one line per cpu, and use the cpu index as the array (it is a pointer tho). This works awesome when removing this validation, but when sharing the "script" I have to concat probes in eg. bash and use one probe per cpu core.

Should I open an issue about this?

odinuge: But thank you for this @viktormalik!

btw. the smart git merge logic added the changelog to `v0.12`, but it is not part of that release... :sweat_smile:

danobi: It should be part of the release. Can you file a bug report if it's not working?

odinuge: Hey @danobi!

The code works as a charm, so that is not a problem! The only way (currently) to use it is to compile master yourself. It is not a part of `0.21.0` [0] or `0.21.1`, even though the changelog in master says so [1].

If you look at the code in this PR, it is added to the `Unreleased` block, but when it is merged in b89f7611f5737eae [2], it is added to `0.21.0`, probably due to the git merging algorithm. :)


Again, thanks for your awesome work! :)

```
[0]: https://github.com/iovisor/bpftrace/commits/v0.12.1
[1]: https://github.com/iovisor/bpftrace/blob/master/CHANGELOG.md
[2]: https://github.com/iovisor/bpftrace/commit/b89f7611f5737eae37daefdecf949c3ea66b8464
```

fbs: yes something went wonky there

```
$ git log --oneline upstream/master
...
118bf1e BTF: set correct size for struct/union types
b89f761 Support array indexing on pointers
373d43e Replace '@' prefix in map name with 'AT_' when creating map
d8d9ace (tag: v0.12.0, release) v0.12.0 release bump
```

Let's just do a .13 release in a few weeks, this one was looooong overdue

viktormalik: @odinuge thanks for the feedback! Please open an issue for "expressions as array indices". Also, if you have a concrete usage example, add it to the issue so that we can evaluate the implementation. I'll try to get to that in near future.

odinuge: Will do @viktormalik! :smile:

PR #1741, opened 2021-03-18
build: Share core bpftrace code with test binary

danobi:
Before, we had a lot of duplicated cmake code between `bpftrace_test`
and `bpftrace`. This was, as they say in the industry, doubly bad,
because you had to write the same cmake code twice and the build would
compile the same code twice.

With this change, there is less duplicated cmake code and the build runs
much faster (both cold starts):

Before:
    609.40s user 30.70s system 329% cpu 3:14.13 total

After:
    422.81s user 21.66s system 300% cpu 2:28.07 total

This commit also fixes a cyclic dependency: ast.a depended on functions
in libbpftrace.a (`signal_name_to_num` and `FakeMap`). Better not to
have cyclic dependencies (ugly and may cause issues), so fix by moving
source into ast/ directory.

##### Checklist


fbs (approved):
Nice!

PR #1745, opened 2021-03-20
tools: handle renamed lookup_fast function in dcsnoop

jeromemarchand:
The lookup_fast function can be rename lookup_fast.constprop.x by gcc
constant propagation optimization and that breaks dcsnoop. Let's look
for both name.

##### Checklist


danobi (approved):

danobi:
@fbs seems fine to take even though we're freezing master -- WDYT?

fbs:
I think tool changes are OK to merge

Issue #1752, opened 2021-03-25
map names do not appear in bpftool


omeranson:
### What reproduces the bug?

Running a simple bpftrace command:
```
bpftrace -e 'BEGIN { @ppid = -1 } tracepoint:syscalls:sys_enter_nanosleep /pid != (uint64)@ppid/ { print((nsecs, pid, tid, comm, probe)); }'
```

And then bpftool:
```
vagrant@ubuntu1804:~$ sudo bpftool map
3705: lpm_trie  flags 0x1
	key 8B  value 8B  max_entries 1  memlock 4096B
3706: lpm_trie  flags 0x1
	key 20B  value 8B  max_entries 1  memlock 4096B
3707: lpm_trie  flags 0x1
	key 8B  value 8B  max_entries 1  memlock 4096B
3708: lpm_trie  flags 0x1
	key 20B  value 8B  max_entries 1  memlock 4096B
3709: lpm_trie  flags 0x1
	key 8B  value 8B  max_entries 1  memlock 4096B
3710: lpm_trie  flags 0x1
	key 20B  value 8B  max_entries 1  memlock 4096B
5244: hash  flags 0x0
	key 8B  value 8B  max_entries 4096  memlock 331776B
5245: perf_event_array  name printf  flags 0x0
	key 4B  value 4B  max_entries 1  memlock 4096B
```

The name of the hash map (ID 5244) is not displayed.

Expected output:
```
3705: lpm_trie  flags 0x1
	key 8B  value 8B  max_entries 1  memlock 4096B
3706: lpm_trie  flags 0x1
	key 20B  value 8B  max_entries 1  memlock 4096B
3707: lpm_trie  flags 0x1
	key 8B  value 8B  max_entries 1  memlock 4096B
3708: lpm_trie  flags 0x1
	key 20B  value 8B  max_entries 1  memlock 4096B
3709: lpm_trie  flags 0x1
	key 8B  value 8B  max_entries 1  memlock 4096B
3710: lpm_trie  flags 0x1
	key 20B  value 8B  max_entries 1  memlock 4096B
5247: hash  name ppid  flags 0x0
	key 8B  value 8B  max_entries 4096  memlock 331776B
5248: perf_event_array  name printf  flags 0x0
	key 4B  value 4B  max_entries 1  memlock 4096B

```

PR #1753, opened 2021-03-25
Replace '@' prefix in map name with 'AT_' when creating map

omeranson:
When calling the API function to create the map, e.g., bcc_create_map,
replace the '@' prefix, if it exists, with the string 'AT_'. Otherwise,
map creation with the name fails, and (e.g.,) BCC falls back to
creating the map without a name.

Closes: #1752

##### Checklist

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1,7 +1,7 @@
 #include <cstring>
 #include <iostream>
-#include <unistd.h>
 #include <linux/version.h>
+#include <unistd.h>
 
 #include "bpftrace.h"
 #include "log.h"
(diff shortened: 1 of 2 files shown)


src/map.cpp
```diff
@@ -13,11 +13,15 @@
...
+  std::string name_ = name;
+  if (name_[0] == '@') {
+    name_ = "AT_" + name.substr(1);
+  }
 #ifdef HAVE_BCC_CREATE_MAP
-  return bcc_create_map(map_type, name, key_size, value_size, max_entries, flags);
+  return bcc_create_map(map_type, name_.c_str(), key_size, value_size, max_entries, flags);
 #else
-  return bpf_create_map(map_type, name, key_size, value_size, max_entries, flags);
+  return bpf_create_map_name(map_type, name_.c_str(), key_size, value_size, max_entries, flags);
```
danobi, line 24: The `#ifdef HAVE_BCC_CREATE_MAP` is to handle when bcc renamed its own `bpf_create_map` -> `bcc_create_map`. Changing this to libbpf's `bpf_create_map_name` may have unintended consequences when bpftrace is compiled without libbpf.

All that to say: please undo this change :)

src/map.cpp
```diff
@@ -13,11 +13,15 @@
 
 namespace bpftrace {
 
-int Map::create_map(enum bpf_map_type map_type, const char *name, int key_size, int value_size, int max_entries, int flags) {
+int Map::create_map(enum bpf_map_type map_type, const std::string & name, int key_size, int value_size, int max_entries, int flags) {
+  std::string name_ = name;
```
danobi, line 17: Style nit:
```suggestion
  std::string fixed_name = name;
```
We generally use trailing `_` for member variables, not locals.

src/map.cpp
```diff
@@ -13,11 +13,15 @@
...
+  if (name_[0] == '@') {
```
danobi, line 18: Should check `!fixed_name.empty()` before indexing

danobi (changes requested):
Thanks for the fix; few comments.

Also please address the clang-format lint -- our policy is to re-format old code when the old code is modified.

src/map.cpp
omeranson: Done.

src/map.cpp
omeranson: Done

src/map.cpp
omeranson: Done. Although C++11 allows indexing the one-passed-the-end character of a string.

omeranson:
Thanks for the review. I went ahead and passed map.cpp via clang-format, since I didn't want to guess the exact formatting.

danobi (approved):
Thank you!

PR #1768, opened 2021-04-03
tests: Use system gtest and gmock

### LINKED ISSUE #1129: 0.9.4: build process clones gtest
It would be good to switch to use system installed gtest because it will not work if build system is cut off from public network.

danobi:
This makes packagers happier b/c the build should no longer require
internet access.

I had to copy paste a FindGMock.cmake file from the internet b/c cmake
has yet to ship a gmock module [0].

Note that if we run into new compatability / versioning issues when
writing tests I am highly inclined to revert this commit. I believe it
is a bigger win (in tradeoff land) to let bpftrace developers easily and
painlessly write tests than enabling packagers to build/run tests at
package time.

This closes #1129.

[0]: https://gitlab.kitware.com/cmake/cmake/-/issues/17365

##### Checklist


danobi:
A nice side effect is that clean bpftrace builds no long spew all that gtest junk

fbs:
Nice. Looks ok and the CI is happy.

PR #1776, opened 2021-04-06
Split out build info into another module

danobi:
7c40e2a3 ("build: Share core bpftrace code with test binary")
broke the libbpf build info b/c libbpftrace target wasn't given
the libbpf cmake vars.

This commit splits out the buildinfo such that the buildinfo is only
needed inside libbpftrace target. This is nicer b/c then we don't have
to pass cmake vars to multiple targets. It's also nicer b/c main.cpp
is less concerned with internal details like build flags.

Also note that this commit also defines LLVM_VERSION_* preprocessor
variables. I'm not sure where those variables were coming from (maybe
one of the llvm header files). It's probably better we define it
ourselves so things are more explicit.

##### Checklist


danobi:
Gonna merge this tomorrow if there's no objections. The broken `--info` output is confusing

fbs:
Looks ok, go for it:)

PR #1778, opened 2021-04-08
Refactor output

viktormalik:
Following the discussion in https://github.com/iovisor/bpftrace/pull/1705#discussion_r590817157, I tried to do some refactoring of the `Output` class. Main changes in this PR are:
- moved all methods doing conversion of values into strings into `Output`
- removed code duplicated between `TextOutput` and `JsonOutput` by moving it into the base `Output` class
- `Output` now defines an interface that all output classes must define
- at the same time, `Output` implements some methods that are likely to behave the same between all output formatters
- one of the effects is that JSON printing now handles nested structs

There's surely room for more improvement, so feel free to suggest more changes that could be done here.

##### Checklist


fbs:
need to stare at it a bit more but looks good

fbs:
looks good.

@andreasgerstmayr, are you interested in giving this a go? I think you're the most active user of this.

andreasgerstmayr:
looks good, I'll try it out next week.

andreasgerstmayr:
I tested it today, works fine!

One thing I noticed is that now, instead of sending the output direct to the output stream `out_`, it's first buffering it in `std::vector<std::string>` and then joining into a new string (looking at `Output::map_to_str`). It's certainly more elegant, but requires more memory. `str_join()` keeps appending to a `std::string` - afaics that used to be very inefficient (allocating new strings every time), but I also read that recent C++ versions optimize that, so it shouldn't be an issue anymore?

fwiw, I didn't notice an impact for my application. I don't know if this might become an issue with very big maps, what do you think @fbs? If we continue on this route (building data structures before printing), we could also consider vendoring a JSON library in the future?

viktormalik:
Thanks for the review @andreasgerstmayr!

Good point about the string efficiency. I don't expect bpftrace users to create maps so huge that this would actually matter, so I prefer the elegant solution rather than a highly optimized one. But if someone thinks that we should address the issue, it shouldn't be a problem to re-implement this. I'm not sure how efficient modern compilers are in terms of string concatenation, but we could probably pre-reserve enough memory in `str_join` to prevent repetitive re-allocation. Or we can print to `out_` directly, which would be even more efficient (not requiring the buffer at all).

andreasgerstmayr:
I'm fine with both ways, I'll leave it to @fbs if the PR should be updated regarding the string allocations or not.

fbs:
Would be nice to have but not blocking. Lets just get this in :)

fbs:
@viktormalik  can you fix the changelog iissue?

viktormalik:
Fixed, ready for rebase.

Should I open an issue for the string allocation optmization?

PR #1779, opened 2021-04-10
Posparm

fbs:
Opening this  to get some feedback on the testing part / the idea in general.

This introduces a new pass that improves that handling of positional params by replacing them with their value. This way we can drop dynamic_cast checks from the semantic phase (still need to do that) and generate more efficient code. The way it mutates the ast:

```
bpftrace -e 'i:s:1 { print(str($1 + 1)); }' "blaat" -dd

AST after: parser
-------------------
Program
 interval:s:1
  call: print
   call: str
    +
     param: $1
     int: 1


AST after: RemovePositionalParams
-------------------
Program
 interval:s:1
  call: print
   string: laat
```

But this makes testing a bit harder to do, e.g.

```
/vagrant/tests/semantic_analyser.cpp:114: Failure
  expected_result
    Which is: 10
  res
    Which is: 1 <------ return code from posparam pass

Input:
kprobe:f { printf("%d", $2); }

Output:
stdin:1:25-27: ERROR: $2 used numerically but given "hello". Try using str($2).
kprobe:f { printf("%d", $2); }
                        ~~
```

It used to be the semantic analyser that caught this, but now its the new pass. So where do we want to test it? Should the pass get its own test suite (I think it should) and should we then move this case there. Or do we keep it in both and have the semant test then test the whole chain of passes that have to happen before it (it kinda already does with field analyser) as well. We would need a way to indicate that we expect the failure to happen in some named pass, so `test_expect_fail('invalid prog', 'PassName')`

I think it makes sense to have a test suite per pass but also to duplicate the test to the semantic phase so we can make sure everything stil fails the way we expect it to. I don't think we should have to indicate which pass it failed in, we just want to be sure that it fails.


viktormalik:
Overall looks good to me - the code is much nicer and cleaner than before. I scanned through the code, but it surely needs some more reviews.

I agree with the testing approach - we should have a separate test suite for each pass, but also duplicate (at least some of) the tests into the semantic analyser to see that it all works together. It shouldn't be necessary to indicate which pass failed - the only reason I can see here would be for cases when a pass changes the result of another pass.

fbs:
this would break posparam + aot and I'm not convinced its the way to go. It does have some useful commits so might pick them into seperate MRs at some point

PR #1795, opened 2021-04-17
Remove join type

Birch-san:
join() does not return a value, so there was never any way that another function in the AST would consume join()'s output as an argument.

I recall from one of the "big string" pull requests (which had tried to introduce a "fmtstr" type to describe a scratch map just like the one join() uses) that it was desirable for _fewer_ types to exist, to limit the number of types for which each function had to implement support.

##### Checklist

  - No language changes
  - No user-visible changes
  - No new behaviour. ran `join()` example from reference guide; it has not regressed. AST still prints fine with `-dd`.

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -2748,11 +2748,9 @@ int SemanticAnalyser::create_maps_impl(void)
   {
     // join uses map storage as we'd like to process data larger than can fit on
     // the BPF stack.
-    std::string map_ident = "join";
-    SizedType type = CreateJoin(bpftrace_.join_argnum_,
-                                bpftrace_.join_argsize_);
-    MapKey key;
-    auto map = std::make_unique<T>(map_ident, type, key, 1);
+    int value_size = 8 + 8 + bpftrace_.join_argnum_ * bpftrace_.join_argsize_;
+    auto map = std::make_unique<T>(
+        "join", BPF_MAP_TYPE_PERCPU_ARRAY, 4, value_size, 1, 0);
     failed_maps += is_invalid_map(map->mapfd_);
     bpftrace_.maps.Set(MapManager::Type::Join, std::move(map));
   }
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -71,12 +71,6 @@ Map::Map(const std::string &name,
   {
     map_type_ = BPF_MAP_TYPE_PERCPU_HASH;
   }
-  else if (type.IsJoinTy())
-  {
-    map_type_ = BPF_MAP_TYPE_PERCPU_ARRAY;
-    max_entries = 1;
-    key_size = 4;
-  }
   else
     map_type_ = BPF_MAP_TYPE_HASH;
 
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -166,7 +166,6 @@ std::string typestr(Type t)
     case Type::string:   return "string";   break;
     case Type::ksym:     return "ksym";     break;
     case Type::usym:     return "usym";     break;
-    case Type::join:     return "join";     break;
     case Type::probe:    return "probe";    break;
     case Type::username: return "username"; break;
     case Type::inet:     return "inet";     break;
@@ -435,11 +434,6 @@ SizedType CreateUSym()
 SizedType CreateKSym()
 {
   return SizedType(Type::ksym, 8);
-}
-
-SizedType CreateJoin(size_t argnum, size_t argsize)
-{
-  return SizedType(Type::join, 8 + 8 + argnum * argsize);
 }
 
 SizedType CreateBuffer(size_t size)
--- a/src/types.h
+++ b/src/types.h
@@ -36,7 +36,6 @@ enum class Type
   string,
   ksym,
   usym,
-  join,
   probe,
   username,
   inet,
@@ -285,11 +284,6 @@ public:
   {
     return type == Type::usym;
   };
-
-  bool IsJoinTy(void) const
-  {
-    return type == Type::join;
-  };
   bool IsProbeTy(void) const
   {
     return type == Type::probe;
@@ -389,7 +383,6 @@ SizedType CreateLhist();
 SizedType CreateHist();
 SizedType CreateUSym();
 SizedType CreateKSym();
-SizedType CreateJoin(size_t argnum, size_t argsize);
 SizedType CreateBuffer(size_t size);
 SizedType CreateTimestamp();
 SizedType CreateMacAddress();


src/ast/semantic_analyser.cpp
```diff
@@ -2748,11 +2748,10 @@ int SemanticAnalyser::create_maps_impl(void)
   {
     // join uses map storage as we'd like to process data larger than can fit on
     // the BPF stack.
-    std::string map_ident = "join";
     SizedType type = CreateJoin(bpftrace_.join_argnum_,
```
fbs, line 2751: I think we should remove this function as well. Bit weird to have a 'create type x' that doesn't correspond to a type.

Birch-san: I agree — had wondered about that. though, I'm not sure what to replace it with? the minimum I can do is inline the arithmetic.

but I think really this should be modelled as a (packed) struct in [`async_event_types.h`](https://github.com/iovisor/bpftrace/blob/master/src/ast/async_event_types.h). semantic_analyser could ask LLVM what the byte size of that struct is.

likewise, codegen could code against the LLVM struct type instead of calculating pointer offsets.

Birch-san: for now I've inlined the arithmetic, which is probably a good place to start.

still wondering whether it'd benefit from developing against a struct. admittedly that can be done in a separate PR.

Birch-san: @fbs not sure what to make of the CI pipeline results. I haven't touched codegen, yet one of the embedded builds believes that codegen changed.

I expect this is flakiness? it was passing previously.

I don't have any way to re-run the tests (is that a privileges thing?) — I can try pushing empty commits at it to get a second opinion?

danobi: I think adding a struct in `async_event_types.h` is definitely the right move. Calculating manual offsets is not very nice.

danobi:
LGTM.

> that it was desirable for fewer types to exist

Yes, we should only be adding user visible types to the type system. User visible means the user can assign the type to a variable.

Feel free to define the async_event_types struct in a followup PR. This is a good cleanup on its own.

danobi (approved):
Any other comments @fbs ?

src/ast/semantic_analyser.cpp
Birch-san: yes, I think defining a struct in `async_event_types.h`, allocating such a struct during codegen and populating it by accessing its members would be nice.

however, I am not sure how `semantic_analyser.cpp` will be able to consume `async_event_types.h` to determine how much memory to allocate in the map? sure, we'd have a `std::vector<llvm::Type*>`. but I think in order to measure the size of the struct, we'd need the LLVMContext and DataLayout?

```c++
llvm::DataLayout &layout = module_.getDataLayout();
LLVMContext &ctx;
std::vector<llvm::Type*> elements;
StructType *struct = StructType::get(ctx, elements, /*packed=*/false);
uint64_t struct_size = layout.getTypeAllocSize(helper_error_struct);
```

we don't have LLVMContext and DataLayout at the time of semantic analysis… is it reasonable to create a way to access it?

currently semantic_analyser keeps itself pretty separated from LLVM concerns, preferring to rely on memory arithmetic in `types.cpp` (like `CreateJoin()`). to be honest I think those functions would benefit from collaborating with the LLVM type system. is there a good way to organise this?

Birch-san: the `LLVMContext` is created in the constructor of `BpfOrc`:  
https://github.com/iovisor/bpftrace/blob/master/src/bpforcv2.cpp#L14

the `DataLayout` is created in the static factory of `BpfOrc`:  
https://github.com/iovisor/bpftrace/blob/master/src/bpforcv2.cpp#L38

`BpfOrc` _could_ invert control of its construction, and ask for `main.cpp` to acquire a `shared_ptr<LLVMContext>` and `shared_ptr<DataLayout>`, which it could then supply to `semantic_analyser.cpp` also? or semantic_analyser could consume from a new file, `llvm_types.cpp` (which main.cpp would construct)?

danobi: What about `sizeof(AsyncEvent::Join)`?

Birch-san: hm… `sizeof()` couldn't accommodate the `join_argnum_ * join_argsize_`. I think unless you use the LLVM types, you're gonna have to use arithmetic.

danobi: Ah yeah I suppose not.

It looks like join_argnum_ and join_argsize_ are constant anyways. Maybe turn those values into preprocessor constant? With the current code it's not like you could modify those values after codegen anyways. Downside would be user can't customize the values. But nobody's complained yet and `join()` has been there for a while.

Then in async_event_types.h you could have:

```
struct Join
{
  uint64_t something;  // Not sure what this type actually should be; just example
  uint64_t something_else; // Ditto
  uint8_t data[kJoinArgNum][kJoinArgSize];
  
  // Similar llvm definition here
  ...
} __attribute__((packed));
```

Birch-san: hm, I think that would work, and yeah looks like it's not exposed to the user yet, and the defaults aren't bothering anyone.

but I'm still reticent to take away the future ability to tune that. it's tuned to "generous amount of memory", and (probably) "people are using this to trace CLI options". fair assumptions, but opinionated. a good way to limit the impact of "opinionated" features is to keep them tunable.

in any case, I'm more interested in a general solution (for all types) that can help `semantic_analyser.cpp` allocate maps of the correct size, and help `bpftrace.cpp` unmarshal perf events correctly, without hardcoding maths into `types.cpp`. any solution based on `sizeof` forces us to use packed structs in our codegen. I think worst-case scenario of that is stack misalignment.

anyway, let's say we do use packed structs and we do model the types using C++ structs. but maybe compromising slightly on the arithmetic by treating the variable-length section as a special case:

```c++
struct Join
{
  uint64_t something;  // Not sure what this type actually should be; just example
  uint64_t something_else; // Ditto
  uint8_t content[0]; // variable-length section
  
  // Similar llvm definition here
  ...
} __attribute__((packed));
```

if semantic_analyser wanted to calculate how much map memory to allocate, it could use [`offsetof()`](https://www.cplusplus.com/reference/cstddef/offsetof/) to determine where the variable-length section starts:

```c++
int map_size = offsetof(struct AsyncEvent::Join, content) + join_argnum_ * join_argsize_;
```

danobi: > but I'm still reticent to take away the future ability to tune that.

I don't think you'd be taking away anything -- it's all implementation details we can change later. 

> in any case, I'm more interested in a general solution (for all types) that can help semantic_analyser.cpp allocate maps of the correct size,

What are the other use cases? I'm really hesitant to add a general solution for N=1 use cases, especially if the general solution is complicated.

> anyway, let's say we do use packed structs and we do model the types using C++ structs. but maybe compromising slightly on the arithmetic by treating the variable-length section as a special case:

Why do we need a VLA for `Join`? It looks like the current code in `BPFtrace::perf_event_printer()` assumes a fixed length.

fbs (approved):

PR #1805, opened 2021-04-21
Disable build tests when install bpftrace.

haosdent:
Disable build tests when install bpftrace to address this error
```
CMake Error at
/usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146
(message):
  Please install the bcc library package, which is required.  Depending
on
  your distro, it may be called bpfcclib or bcclib (Ubuntu), bcc-devel
  (Fedora), or something else.  If unavailable, install bcc from source
  (github.com/iovisor/bcc).  (missing: LIBBCC_LIBRARIES
LIBBCC_INCLUDE_DIRS)
Call Stack (most recent call first):
  /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393
(_FPHSA_FAILURE_MESSAGE)
  cmake/FindLibBcc.cmake:72 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:86 (find_package)
```

##### Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -249,7 +249,7 @@ sudo apt-get install -y bison cmake flex g++ git libelf-dev zlib1g-dev libfl-dev
 sudo apt-get install -y llvm-7-dev llvm-7-runtime libclang-7-dev clang-7
 git clone https://github.com/iovisor/bpftrace
 mkdir bpftrace/build; cd bpftrace/build;
-cmake -DCMAKE_BUILD_TYPE=Release ..
+cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF ..
 make -j8
 sudo make install
 ```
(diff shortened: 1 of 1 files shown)


haosdent:
@danobi would you like to take a look this minor changes?

fbs:
I'm a bit confused, how does this fix the error you see? We _need_ bcc for a lot of code loading and uprobe/elf handling, without it it just won't work. Disabling tests shouldn't remove that dependency

danobi:
Yeah, I'm not sure how disabling testing fixing the missing bcc dependency.

BTW, probably update all the other `cmake` call sites in that doc too to have `-DBUILD_TESTING=OFF` to make it consistent.

haosdent:
@fbs @danobi sry, I paste wrong error, error is missing GTEST

fbs:
If gtest is missing you should install the libs or run with VENDOR_GTEST=1, not building tests is not recommended

PR #1817, opened 2021-04-26
embedded: limit support to llvm 8 and 12

fbs:
This logic should've matched on version exactly instead of defaulting to
8, even when we specify LLVM12. As 9 never worked and 12 fixes a bug we
can limit it to 8 and 12.


fbs:
ah this fails because for the debian build we only do the LIBCLANG_ONLY thing which uses find_package(LLVM) and shadows the LLVM_VERSION flag we set on the CLI.

fbs:
This one is ready to go in

PR #1819, opened 2021-04-27
Reduce LLVM build options

fbs:
Builds on #1817 

----

We currently have a lot of build options around LLVM. We can build LLVM, with or without Clang and even only libclang. This is hard to maintain, we need to keep track of  all LLVM releases and all patches to make sure it works. Currently only LLVM/Clang 8 really works, for the others the sources or patches are not in sync.

Imo we only really care about LLVM8 & 12. 8 is what we have now and mostly works, 12 fixes an important bug and is going to be our next base for dev and embedded. By reducing the options the setup is a lot simpler and easier to maintain.

I'm also not sure if the libclang only stuff has a real use case for us

The new options will be:

- normal, dynamic
- static
- static + static_libc
- static + llvm&clang
- static + llvm&clang + static libc


fbs:
@dalehamel how do you feel about doing this?

danobi (commented):
I'm +1 on these changes. The number of different options before was difficult/impossible to maintain with the manpower we have available.

The hope when we merged all this in the beginning was the maintaining this wouldn't be too bad. I think time has shown that it is **really** hard to maintain.

dalehamel:
@fbs changes LGTM.

When the linking options were originally exposed, doing them separately made sense from a modularity perspective, but in practice I don't think we need the modularity. Rolling them all up into the option we actually use (EMBED_BUILD_LLVM)

> I'm also not sure if the libclang only stuff has a real use case for us

The only benefit of it that I recall was that it built much faster by using the LLVM package deps, and only builds the one or two libs it is missing IIRC. If we cache the LLVM build there is no benefit, and it adds complexity. It is fine to drop.

Thanks for doing this. I'm hoping the work my team will be completing soon will allow me to focus on some missing bpftrace fixes and refactors i've had on my to-do list for a long time.

PR #1823, opened 2021-04-29
tools: Make tools rely on BTF instead of header files.

### LINKED ISSUE #1820: Some tools don't work without kernel-devel package when the kernel is build with ikheaders and BTF
### What reproduces the bug?

biosnoop.bt, dcsnoop.bt, mdflush.bt, runqlat.bt, tcpaccept.bt, tcpconnect.bt, rcpdrop.bt and tcpretrans.bt won't work when no kernel-devel package is installed, even though the kernel is build with CONFIG_IKHEADERS=m.
Ever since commit 17bbcd2b0773 ("kheader: Do not unpack kheaders if system has BTF"), we skip unpacking kheaders when BTF is available, but it seems those tools rely on data present in kheaders and not BTF. Reverting commit 17bbcd2b0773 fixes the issue.

Note that for some tools like biosnoop.bt, it just a matter of not finding the header file and removing the  #include line fixes the issue.
```
# /usr/share/bpftrace/tools/biosnoop.bt
definitions.h:2:10: fatal error: 'linux/blkdev.h' file not found
```

For other like dcsnoop.bt, it's complaining about incomplete type:
```
# /usr/share/bpftrace/tools/dcsnoop.bt 
definitions.h:5:25: error: field has incomplete type 'struct path'
definitions.h:5:16: note: forward declaration of 'struct path'
definitions.h:6:25: error: field has incomplete type 'struct qstr'
definitions.h:6:16: note: forward declaration of 'struct qstr'
```

### bpftrace --info
System
  OS: Linux 4.18.0-304.3.el8.bz1928463.x86_64 #1 SMP Wed Apr 14 11:46:12 EDT 2021
  Arch: x86_64

Build
  version: v0.12.1
  LLVM: 11.0.0
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  bcc bpf_attach_uprobe refcount: no
  libbpf: yes
  libbpf btf dump: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  iter:task: yes
  iter:task_file: yes

jeromemarchand:
Many distribution already ship BTF and this is more robust that
parsing header files.

Fixes #1820

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -3522,7 +3522,8 @@ Histograms can also be printed on-demand, using the `print()` function. Eg:
 # BTF Support
 
 If kernel has BTF, kernel types are automatically available and there is no need to include additional headers
-to use them.
+to use them. To allow users to detect this situation in scripts, the preprocessor macro `BPFTRACE_HAVE_BTF` 
+is defined if BTF is detected. See tools/ for examples of its usage.
 
 Requirements for using BTF:
 
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -748,6 +748,9 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     // Since we're omitting <linux/types.h> there's no reason to
     // add the wokarounds for it
     args.push_back("-D__CLANG_WORKAROUNDS_H");
+    // Let script know we have BTF -- this is useful for prewritten tools to
+    // conditionally include headers if BTF isn't available.
+    args.push_back("-DBPFTRACE_HAVE_BTF");
 
     if (handler.parse_file("definitions.h", input, args, input_files, false) &&
         handler.has_redefinition_error())
@@ -778,6 +781,7 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
   {
     // There is a conflict (redefinition) between user-supplied types and types
     // taken from BTF. We cannot use BTF in such a case.
+    args.pop_back();
     args.pop_back();
     args.pop_back();
     input_files.back() = get_empty_btf_generated_header();
--- a/tools/biosnoop.bt
+++ b/tools/biosnoop.bt
@@ -1,5 +1,4 @@
 #!/usr/bin/env bpftrace
-#include <linux/blkdev.h>
 /*
  * biosnoop.bt   Block I/O tracing tool, showing per I/O latency.
  *               For Linux, uses bpftrace, eBPF.
@@ -10,6 +9,10 @@
  *
  * 15-Nov-2017	Brendan Gregg	Created this.
  */
+
+#ifndef BPFTRACE_HAVE_BTF
+#include <linux/blkdev.h>
+#endif
 
 BEGIN
 {
--- a/tools/dcsnoop.bt
+++ b/tools/dcsnoop.bt
@@ -15,6 +15,7 @@
  * 08-Sep-2018	Brendan Gregg	Created this.
  */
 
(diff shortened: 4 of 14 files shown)


sevan:
Which of the required kernel features listed in [INSTALL.md](/iovisor/bpftrace/blob/master/INSTALL.md#linux-kernel-requirements) are now redundant?

jeromemarchand:
> Which of the required kernel features listed in [INSTALL.md](/iovisor/bpftrace/blob/master/INSTALL.md#linux-kernel-requirements) are now redundant?
None. The issue is with kernel header from kernel-devel package vs ikheaders vs BTF (https://github.com/iovisor/bpftrace/blob/master/INSTALL.md#kernel-headers-install).

Currently, bpftrace doesn't unpack ikheaders when BTF is available, which make some tools fail when the kernel-devel package isn't installed, which shouldn't be needed if the kernel is build with ikheaders.

fbs:
The tools check is still an important CI step imo. Its the only check that real life scripts at least work. But the ci doesn't support btf yet (atleast not last time i checked, maybe it does now?). Might make sense to support have a btf and non-btf version of the tools. 

Maybe we cna have a conditional. `#ifndef BTF  ... include headers ... #endif`

danobi:
> Maybe we cna have a conditional. #ifndef BTF ... include headers ... #endif

Yeah, maybe this is the best way to deal with it. bpftrace can do a `-D__BPFTRACE_BTF` (or something) when parsing headers

viktormalik:
Taking this over from @jeromemarchand. Should be ready for review.

fbs:
Looks good. My only concern is the `__` in the name. The `__` always makes it feel like something internal to me but if we do this we make it 'part of the language' // the official 'macro' for BTF support. 

Maybe just dropping the `__` is enough?

@brendangregg

viktormalik:
> Looks good. My only concern is the `__` in the name. The `__` always makes it feel like something internal to me but if we do this we make it 'part of the language' // the official 'macro' for BTF support.

Fair enough, so if we make it "official", should we also document it somewhere?

> Maybe just dropping the `__` is enough?

I'm fine with that.

danobi:
Dropping `__` sounds good to me. And yes, please document

viktormalik:
Updated and documented.

docs/reference_guide.md
```diff
@@ -3522,7 +3522,8 @@ Histograms can also be printed on-demand, using the `print()` function. Eg:
 # BTF Support
 
 If kernel has BTF, kernel types are automatically available and there is no need to include additional headers
-to use them.
+to use them. To allow users to detect this situation in scripts, `BPFTRACE_HAVE_BTF` is defined if BTF is
```
danobi, line 3525:
```suggestion
to use them. To allow users to detect this situation in scripts, the preprocessor macro `BPFTRACE_HAVE_BTF` is defined if BTF is
```

danobi (approved):

fbs:
yeah lets get it in

PR #1827, opened 2021-05-01
embedded: use llvm container

fbs:
This let's the embedded build make use of the LLVM image we've created instead of building LLVM from scratch every time.

The only noticable change is that we have to bump the cmake version again to 3.13, which is ["modern"](https://cmake.org/pipermail/cmake/2018-November/068637.html). This might cause an issue for packages but cmake is available as pre built binary and if you really want to these lines can easily be patched out as it only impacts the static build.


fbs:
The linker flag stuff took some time to figure out. Still not entirely happy with manually specifying the dynamic lm,ldl,lrt and static lz, they should be added with add_library but it works and the flags make sense now

fbs:
Needs #1829

fbs:
ready for a review @danobi @dalehamel . I tried to keep history readable.

Once this is in I can try the LLVM12 build.

dalehamel:
@fbs great work on this! This is so much faster. I'm sorry that I never got to it but it looks like you've done a fantastic job.

fbs:
> @fbs great work on this! This is so much faster. I'm sorry that I never got to it but it looks like you've done a fantastic job.

haha, can't blame you. It's been a lot of waiting the past weeks.

Would be nice if we could cache all containers at some point so we don't have to install anything, but thats for later.

PR #1829, opened 2021-05-04
embedded: do a MinSizeRel llvm build

fbs:
We're currently doing a Debug build (cmake default) which gets mapped to
a `RelWithDebInfo` build in cmake/embed/embed_llvm.cmake. But for our
release we want the smallest build possible.


fbs:
Screenshot 2021-05-04 at 17 36 59
 
quite a size diff..

fbs:
killed the embedded builds to get some CI time for 1827

PR #1830, opened 2021-05-04
embedded: llvm 12 build

fbs:
builds on #1827 
----


fbs:
33->46MB, quite a size diff but im happy that llvm12 finally builds. Would be nice if someone can figure out how to reduce it

PR #1835, opened 2021-05-10
Warn if attaching a kprobe to a non-traceable function

viktormalik:
Currently, when trying to attach a kprobe a non-traceable function, `bpf_attach_kprobe` will fail and we're relying on libbpf to print an error message. This error messgage may be confusing (see example in #1818) though, so it's better if we print our own warning to the user.

##### Checklist

### DIFF
--- a/src/btf.h
+++ b/src/btf.h
@@ -14,6 +14,8 @@ struct btf_type;
 
 namespace bpftrace {
 
+class BPFtrace;
+
 class BTF
 {
   enum state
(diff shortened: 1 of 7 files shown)


src/btf.h
```diff
@@ -24,6 +26,10 @@ class BTF
 
 public:
   BTF();
```
danobi, line 28: Any other callers of this? I'd rather BTF() only have one constructor so nobody forgets to pass `BPFtrace*`

danobi (commented):
LGTM overall

src/btf.h
viktormalik: Unfortunately yes, it's called from `BPFfeature` which doesn't have access to `BPFtrace` ([bpffeature.cpp:199](https://github.com/iovisor/bpftrace/blob/master/src/bpffeature.cpp#L199))

PR #1836, opened 2021-05-10
Make kaddr() report failure for unknown kernel symbols

danobi:
Previously, bad kaddr() calls would silently fail and return 0. This is
bad b/c you silently get bad data. Let's report and error and fail
instead.

##### Checklist


fbs (approved):

src/ast/codegen_llvm.cpp
```diff
@@ -612,6 +612,8 @@ void CodegenLLVM::visit(Call &call)
     uint64_t addr;
     auto name = bpftrace_.get_string_literal(call.vargs->at(0));
     addr = bpftrace_.resolve_kname(name);
+    if (!addr)
+      throw std::runtime_error("Failed to resolve kernel symbol: " + name);
```
fbs, line 616: we should be able to use `LOG` here

danobi: Not sure about that. codegen isn't hooked up to check for error messages like semantic_analyser is. Could `LOG(FATAL)` but it'd be better if we had a nice message instead

PR #1839, opened 2021-05-13
build: Cleanup cmake search paths

fbs:
Whlie looking into another cmake issue I found these. I'm pretty confident that it works (see commit message), but if someone can try a build that would be great

### DIFF
--- a/cmake/FindLibBcc.cmake
+++ b/cmake/FindLibBcc.cmake
@@ -21,25 +21,12 @@ find_path (LIBBCC_INCLUDE_DIRS
   NAMES
     bcc/libbpf.h
   PATHS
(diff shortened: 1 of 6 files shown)


fbs:
random test failure in ci, build was fine

danobi:
Hmm, I'll try to run this on a suse container tomorrow. IIRC suse had some weird pathing

fbs:
> Hmm, I'll try to run this on a suse container tomorrow. IIRC suse had some weird pathing

Ah I never use suse, can't hurt to test it. Although 'weird' shouldn't matter as long as it's in line with the rest of the system libs.

danobi (approved):

PR #1848, opened 2021-05-18
Embedded fix

fbs:


fbs:
This seems to work in my fork, but actions have been quite unreliable today so havent done proper testing. Guess we could just revert it if it doesn't help

Issue #1845, opened 2021-05-15
New BPF compiler backend


ajor:
This ticket tracks the creation of a new compiler backend for bpftrace to replace LLVM.

## What's wrong with LLVM?
LLVM produces great bytecode, but it is a very large library. This makes it difficult to use bpftrace on low-resource systems.

## Goals of new compiler backend
- Small compiled size on disk
- Low runtime memory usage
- Reasonable BPF code generation
- Faster compilations than LLVM as currently used. We'll have far fewer abstractions and optimisation passes, so we'd better be faster!

These are vague goals but hopefully will help guide us as we start implementing. The idea is to be a minimal compiler backend that can still produce reasonable BPF bytecode. It does not need to support all features of LLVM or produce optimal BPF bytecode.

## Implementation

- A BPF-specific intermediate representation (IR)
  - Three Address Code, Single Static Assignment IR
  - Nodes directly mapping to each BPF instruction (load, add, branch, etc.)
  - Minimal abstractions as there is only a single target and few users of this backend
- LLVM IR generation
  - This will allow us to convert bpftrace over to using the new API before the entire backend is ready, which may let us spot things we haven't thought of
  - Gives bpftrace the option to use LLVM when available to produce potentially more efficient BPF programs, without us having to maintain two separate "codegen_xxx.cpp" files
- Passes over IR
  - Register allocation
  - Stack allocation
  - "mem2reg"
  - Convert 3-address instructions to 2-address instructions
  - Other optimisation passes TBD
- BPF code generation

Some optimisations are absolutely required, such as on memory placement to avoid blowing the stack, so direct BPF-bytecode emission without an IR is not an option.

fbs:
Would be great if we could offer it as library other projects can use. That way they can target our IR and use it as an easier way to get LLVM or our compiler as light weight target.

danobi:
> Would be great if we could offer it as library other projects can use. That way they can target our IR and use it as an easier way to get LLVM or our compiler as light weight target.

As long as we:
* have a cmake target that supports building a `.so` or `.a` + installing library and header file
* be careful with what we put in the bpfir headers

it should be pretty easy

danobi:
Since we're changing codegen and thus also the test fixtures, it's a good time to consider if we wanna keep the codegen testing framework as-is.

Do we wanna do something fancier like https://www.llvm.org/docs/CommandGuide/FileCheck.html ?

Advantage of doing something like that is:
[...]

PR #1850, opened 2021-05-20
Improve work with struct (record) types

### LINKED ISSUE #1396: Improve type::record (struct)

viktormalik:
This PR attempts to improve (mainly internal) work with record types in several ways:
- Adds reference to the struct layout (the `Struct` type) into `SizedType`. This allows to access the record's fields directly from the expression type, without the need to fall back to `bpftrace.structs_`. In fact, the work with `bpftrace.structs_` is now completely hidden from the users of `SizedType`.
- Preserves the original order of struct fields. Since we now allow to print structs, changing the field order could confuse the user.
- Unifies the internal structure of records and tuples (classes `Struct` and `Tuple` are merged). The only difference is that tuple fields have empty names.

This should resolve #1396.

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -393,7 +393,11 @@ SizedType BTF::get_stype(__u32 id)
     const char *cast = btf_str(btf, t->name_off);
     assert(cast);
     std::string comp = btf_is_struct(t) ? "struct" : "union";
-    stype = CreateRecord(t->size, comp + " " + cast);
+    std::string name = comp + " " + cast;
+    // We're usually resolving types before running ClangParser, so the struct
+    // definitions are not yet pulled into the struct map. We initialize them
+    // now and fill them later.
+    stype = CreateRecord(name, bpftrace_->structs.LookupOrAdd(name, t->size));
   }
   else if (btf_is_ptr(t))
   {
--- a/src/struct.h
+++ b/src/struct.h
@@ -42,6 +42,19 @@ struct Struct
 {
   int size; // in bytes
   FieldsMap fields;
+
+  explicit Struct(int size) : size(size)
+  {
+  }
+
+  bool HasField(const std::string &name) const;
+  const Field &GetField(const std::string &name) const;
+  void AddField(const std::string &field_name,
+                const SizedType &type,
+                ssize_t offset,
+                bool is_bitfield,
+                const Bitfield &bitfield,
+                bool is_data_loc);
 };
 
 struct Tuple
@@ -56,4 +69,17 @@ struct Tuple
 };
 
 std::ostream &operator<<(std::ostream &os, const TupleFields &t);
+
+class StructManager
+{
+public:
+  void Add(const std::string &name, size_t size);
+  std::shared_ptr<Struct> Lookup(const std::string &name) const;
+  std::shared_ptr<Struct> LookupOrAdd(const std::string &name, size_t size);
+  bool Has(const std::string &name) const;
+
+private:
+  std::map<std::string, std::shared_ptr<Struct>> struct_map_;
+};
+
 } // namespace bpftrace
--- a/src/struct.h
+++ b/src/struct.h
@@ -20,6 +20,7 @@ struct Bitfield
 
 struct Field
 {
+  std::string name;
   SizedType type;
   ssize_t offset;
 
@@ -35,13 +36,12 @@ struct Field
   bool is_data_loc = false;
 };
 
-using FieldsMap = std::map<std::string, Field>;
-using TupleFields = std::vector<Field>;
+using Fields = std::vector<Field>;
 
 struct Struct
 {
   int size; // in bytes
-  FieldsMap fields;
+  Fields fields;
 
   explicit Struct(int size) : size(size)
   {
@@ -62,13 +62,13 @@ struct Tuple
   size_t size; // in bytes
   int align;   // in bytes
   bool padded = false;
-  TupleFields fields;
+  Fields fields;
 
   static std::unique_ptr<Tuple> Create(std::vector<SizedType> fields);
   void Dump(std::ostream &os);
 };
 
-std::ostream &operator<<(std::ostream &os, const TupleFields &t);
+std::ostream &operator<<(std::ostream &os, const Fields &t);
 
 class StructManager
 {
--- a/src/struct.h
+++ b/src/struct.h
@@ -41,6 +41,8 @@ using Fields = std::vector<Field>;
 struct Struct
 {
   int size; // in bytes
+  int align = 1; // in bytes, used for tuples only
+  bool padded = false;
   Fields fields;
 
   explicit Struct(int size) : size(size)
@@ -55,16 +57,8 @@ struct Struct
                 bool is_bitfield,
                 const Bitfield &bitfield,
                 bool is_data_loc);
-};
 
-struct Tuple
-{
-  size_t size; // in bytes
-  int align;   // in bytes
-  bool padded = false;
-  Fields fields;
-
-  static std::unique_ptr<Tuple> Create(std::vector<SizedType> fields);
+  static std::unique_ptr<Struct> CreateTuple(std::vector<SizedType> fields);
   void Dump(std::ostream &os);
 };
 
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -2,6 +2,7 @@
 #include <cassert>
 #include <iostream>
 
+#include "bpftrace.h"
 #include "log.h"
 #include "struct.h"
 #include "types.h"
@@ -355,10 +356,11 @@ SizedType CreatePointer(const SizedType &pointee_type, AddrSpace as)
   return ty;
 }
 
-SizedType CreateRecord(size_t size, const std::string &name)
+SizedType CreateRecord(const std::string &name, std::shared_ptr<Struct> record)
 {
-  auto ty = SizedType(Type::record, size);
+  auto ty = SizedType(Type::record, record ? record->size : 0);
   ty.name_ = name;
+  ty.record = record;
   return ty;
 }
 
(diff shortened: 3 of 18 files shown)


viktormalik:
One embedded test failed, a restart should resolve it.

src/types.cpp
```diff
@@ -355,13 +356,26 @@ SizedType CreatePointer(const SizedType &pointee_type, AddrSpace as)
   return ty;
 }
 
-SizedType CreateRecord(size_t size, const std::string &name)
+SizedType CreateRecord(const std::string &name, size_t size, BPFtrace &bpftrace)
 {
-  auto ty = SizedType(Type::record, size);
+  auto record = bpftrace.structs_.insert(
```
fbs, line 361: I don't think we should tie the typesystem to BPFtrace like this. A type is just a
description so we know how to deal with it. Instance of the type (all the structs coming
from btf/clang) are not relevant to the type system, they should be dealt with at another level.

This also makes #1845 harder. Right now we can use the type system and sill keep it simple, if we add BPFtrace in you have
to pull in all the stuff defined in BPFtrace (i.e. everything we have) which makes it way harder to use as lib.

This structmap type should really only exists between the clang,btf and semantic phases. clang,btf define structs and store
them in the map. The semantic phase finds the types it need, which it knows from either pointer cast or btf,  and assigns it
to ast nodes. All unused types can be discarded at that point. 
Keeping it in BPFtrace is easier though, a single object that keeps all the state instead of having to pass state between phases.

Having a simple abstraction to access this instead of dealing with bpftrace_.structs_ directly makes sense. Maybe a `StructManager`
similar to `MapManager`. But that doesn't belong in the typesystem.

Not needing bpftrace for all the type related stuff should also keep the btf & clang code simpler, don't have to keep passing bpftrace around.

fbs (changes requested):
Thanks for working on this. It's a nice improvement overall, improving the type system makes a lot of things easier.

src/btf.cpp
```diff
@@ -372,7 +372,7 @@ const struct btf_type *BTF::btf_type_skip_modifiers(const struct btf_type *t)
   return t;
 }
 
-SizedType BTF::get_stype(__u32 id)
+SizedType BTF::get_stype(__u32 id, BPFtrace &bpftrace)
```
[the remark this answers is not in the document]
viktormalik: Sure we can, I just forgot that we already have the pointer in BTF.

src/types.cpp
viktormalik: Hm, this sounds reasonable. If I understand correctly, you're suggesting to keep the `bpftrace.structs_` map which will be populated from ClangParser (the way it is done now) and its entries will be copied into AST nodes during semantic analysis (perhaps using an abstraction layer). In other words, instead of holding a pointer to the struct map, `SizedType` will contain a `Struct` definition by value.

This will introduce some struct copying, which I wanted to avoid, but I understand that having the type system independent from BPFtrace is more important.

fbs: A lot of sizedtype handling is currently by value anyway. It would be nice if types are immutable so we only have to store on copy and avoid corrupting it, but thats a different issue.

I think we can have a simple `StructManager` class that handles add /lookup.

So in clang_parser you can write something like:

1. bpftrace.structs.Add(name)
1. bpftrace.structs.AddField(name, field...)
1. bpftrace.structs.Lookup(name)

1,2 would be relevant to the clang/btf phase for populating types.
3. would be used in the semantic phase to assign types to ast nodes

I *think* that should work ok. Its not that different from what you have now. It would just move the logic from CreateRecord to a wrapper class.

viktormalik: `StructManager` sounds good, it shouldn't be a problem to implement it.

What will be a problem is storing full field definitions inside `SizedType` by value. There are 2 issues:

1. How do we handle fields that are pointers to structs? For pointers, we store the pointed type inside, however, this will be likely impossible with recursive types.
2. There are some struct types that are created before their definition appears in the struct map. These are in particular (1) kfunc args types (created from `FieldAnalyser`) and (2) types of struct fields that are pointers to forward-declared structs (created in `ClangParser`). 
 
I'm afraid that these will prevent us from storing full field definitions inside `SizedType`. We could solve no. 2 by updating the relevant types after all struct types definitions are retrieved, but no.1 will be problematic unless we somehow rework the way pointer types are stored.

Or we leave `SizedType` to contain a reference to the struct map (since it's a `shared_ptr`, it should survive even if the entry in the struct map is deleted later) and just move the logic of adding entries to the struct map from `CreateRecord` to `StructManager`.

fbs: Ah yeah recursive types are a pain. But as long as every type is unique (e.g.  every int8 points to the same immutable  type) that shouldn't be an issue. All we'd need is a way to forward declare structs and  extend them later. As everything is pointing to the same Type object we wouldn't have to do anything later. When the  fields of `Struct Y` get filled `Struct S { Y y }` will be updated automatically as everything references the same Y. But if we do that we do need some kind of type storage, not just for structs but for all types.

Thats also how LLVM handles its types, every type is unique within the context. Basic types are pre-allocated, https://llvm.org/doxygen/classllvm_1_1LLVMContextImpl.html  , https://llvm.org/doxygen/classllvm_1_1Type.html#aa75984a442f2379de0c66018201fa628  and structs must be unique

> There are two different kinds of struct types: Literal structs and Identified structs.

> Literal struct types (e.g. { i32, i32 }) are uniqued structurally, and must always have a body when created. You can get one of these by using one of the StructType::get() forms.

> Identified structs (e.g. foo or %42) may optionally have a name and are not uniqued. The names for identified structs are managed at the LLVMContext level, so there can only be a single identified struct with a given name in a particular LLVMContext. Identified structs may also optionally be opaque (have no body specified). You get one of these by using one of the StructType::create() forms.

I think that would make sense for us too. No more type copying, only assignment. Getting there will be quite some work but maybe this PR can be a first step. Then we can leave it as is and extend the type storage later. If we keep the type storage as a separate type we don't need to depend on bpftrace.h, but that's easy to refactor later.

viktormalik: Actually, this is pretty much how it is implemented right now (in fact, it's inspired by the LLVM type system). All struct types are unique and stored in `bpftrace.struct_map`. `SizedType` then contains just a reference to the unique type (and thanks to that, the actual definitions can be forward-declared and populated later).

So the only improvement of the current implementation will be to use a `StructManager` instead of passing `BPFtrace` to `CreateRecord`.

fbs: Yeah, but doing it for all types and making sure no copies happen (so we can drop the copy constructor) is a bit more work. 


 > So the only improvement of the current implementation will be to use a StructManager instead of passing BPFtrace to CreateRecord.

Yes, sounds good

viktormalik: > Yeah, but doing it for all types and making sure no copies happen (so we can drop the copy constructor) is a bit more work.

Of course, hopefully, we'll be able to build on this approach (e.g. eventually transforming `StructManager` into `TypeManager`).

Anyway, I implemented the discussed changes, PR should be ready for another review.

fbs (approved):
nice!

src/struct.h
```diff
@@ -56,4 +63,26 @@ struct Tuple
 };
 
 std::ostream &operator<<(std::ostream &os, const TupleFields &t);
+
+class StructManager
+{
+public:
+  using StructMap = std::map<std::string, std::shared_ptr<Struct>>;
```
danobi, line 64: Nit: I'm not a fan of single use typedefs. Unless `StructMap` is used in multiple places, I suggest remove it b/c it just adds an extra layer of things to look at when reading code

src/types.cpp
```diff
@@ -510,4 +512,16 @@ ssize_t SizedType::GetAlignment() const
...
+const std::map<std::string, Field> &SizedType::GetStructFields() const
+{
+  assert(IsRecordTy() && record);
+  return record->fields;
+}
+
+const Field &SizedType::GetField(const std::string &name) const
+{
+  assert(IsRecordTy());
+  return record->fields[name];
```
danobi: `Struct::GetField()` throws exception on missing field (via `.at()`). Do that here too?

tests/mocks.cpp
```diff
@@ -75,67 +75,58 @@ void setup_mock_probe_matcher(MockProbeMatcher &matcher)
...
-    .size = 16,
-    .fields = { { "saddr_v6",
-                  Field{
-                      .type = CreateArray(16, CreateUInt(8)),
-                      .offset = 8,
-                      .is_bitfield = false,
-                      .bitfield = {},
-                  } } },
-  };
+  bpftrace.structs.Add("struct _tracepoint_sched_sched_one", 8);
```
danobi, line 78: Kinda weird now that I think about it to have `struct` keyword in key. It's it's a huge diff to change that then don't bother.

src/struct.h
```diff
@@ -56,4 +63,26 @@ struct Tuple
...
+
+  void Add(const std::string &name, size_t size);
+  void AddField(const std::string &struct_name,
+                const std::string &field_name,
+                const SizedType &type,
+                ssize_t offset,
+                bool is_bitfield,
+                const Bitfield &bitfield,
+                bool is_data_loc);
```
danobi, line 73: Wouldn't it make more sense for `AddField` be a method on `Struct` and not `StructManager`?

danobi (commented):
Nice cleanup overall :)

src/types.cpp
danobi: Nvm, this is fixed in a later commit

src/struct.h
viktormalik: Yeah, that makes more sense, fixed.

tests/mocks.cpp
viktormalik: We're using the original type name collected by `ClangParser` so we'd have to have a special handling for tracepoint defs, which I think doesn't make much sense.

danobi (approved):

PR #1857, opened 2021-05-25
tracepoint: Return 1 from tracepoint probes

danobi:
Returning 0 from a tracepoint prog causes the perf subsystem to ignore
the tracepoint event. We should return 1 to be a better citizen.

And yes, everyone in the kernel thinks it's weird too. But it's too late
to change.

##### Checklist


fbs (approved):
lame :( but looks ok

PR #1858, opened 2021-05-25
Re-enable BTF tests

viktormalik:
When unifying core and tests code, we removed definition of `HAVE_LIBBPF_BTF_DUMP` from `tests/CMakeLists.txt` which causes BTF tests not to be compiled.

After re-enabling BTF tests, some CI builds failed, especially if system BTF was not available or if BCC did not have kfunc probes. This PR hence disables the failing tests for such cases.

##### Checklist


fbs:
The same change also caused a mem leak due to flags missing. Would it make sense to refactor these flags so they're global

so instead of

```
target_compile_definitions(bpftrace_test PRIVATE HAVE_BCC_KFUNC)
```

we do this:

 main cmakelist
```
set(BPFTRACE_FLAGS "${BPFTRACE_FLAGS}" HAVE_BCC_KFUNC)
```

and in src & tst

```
target_compile_definitions(bpftrace_test PRIVATE $BPFTRACE_FLAGS)
```

or whatever the correct cmake syntax is. That way we don't have to worry about flag mixup when we add one.

cc @danobi

viktormalik:
Good idea, I agree with this. There are many existing flags that are not used in tests now, but once they start to be used, we'd have to add them one-by-one to `src/CMakeLists.txt`. This way, we'll have all flags both in src/ and tests/.

I'll do this refactoring if there are no objections against this approach.

tests/semantic_analyser.cpp
```diff
@@ -2214,6 +2214,8 @@ class semantic_analyser_btf : public test_btf
 {
 };
 
+#ifdef HAVE_BCC_KFUNC
```
fbs, line 2217: do we really need these guards? I thought we had some mocking in place. Was hoping the 'unit' test suite could be independent of syslibs/kernel features. It wasn't failing in the CI right?

danobi: Agreed, in general unit tests should not rely on environment at all and instead use mocked data. From looking at the semantic analyser tests, I think this should already be mocked out.

tests/clang_parser.cpp
```diff
@@ -667,6 +667,9 @@ TEST(clang_parser, btf_unresolved_typedef)
   // size_t is defined in stddef.h, but if we have BTF, it should be possible to
   // extract it from there
   BPFtrace bpftrace;
+  if (!bpftrace.btf_.has_data())
+    return;
```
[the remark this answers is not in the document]
viktormalik: Unfortunately, GTEST_SKIP is available from version 1.10.0 and we're currently using 1.8.1 (for vendored gtest). If you're fine with upgrading to 1.10.0 then using GTEST_SKIP is certainly a better approach.

tests/semantic_analyser.cpp
viktormalik: This guard checks if BCC is able to attach to kfuncs because the same guard is used in our semantic analyser [here](https://github.com/iovisor/bpftrace/blob/339ff4b116ad63f7cd546e47a6a71d1425e362de/src/ast/semantic_analyser.cpp#L2580-L2584). If it is not defined, then these tests fail (which also happens in CI).

tests/clang_parser.cpp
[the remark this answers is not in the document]
danobi: I've bumped the gtest version to 1.10.0 (da9dbe893d2) -- please rebase and use GTEST_SKIP

tests/semantic_analyser.cpp
danobi: Ah, makes sense then.

danobi (approved):

PR #1867, opened 2021-06-01
parser: rewrite to remove most conflicts

fbs:
R/R conflicts should be avoided in LALR parsers (like ours) but we had
31 of them.

This completely restructures the parser to follow a more common
approach [1][2] which removes most of the ambiguities and *should* make
the parser easier to extend and understand.

The attach point part still has a r/r conflict which will be for a
follow up.

This still has a few s/r conflicts but those aren't as big of an issue

```
 warning: 4 shift/reduce conflicts [-Wconflicts-sr]
 warning: 1 reduce/reduce conflict [-Wconflicts-rr]

State 10 conflicts: 1 reduce/reduce
State 41 conflicts: 1 shift/reduce
State 51 conflicts: 1 shift/reduce
State 81 conflicts: 2 shift/reduce
```

[1]: http://www.quut.com/c/ANSI-C-grammar-y-2011.html#inclusive_or_expression
[2]: https://github.com/dtrace4linux/linux/blob/master/libdtrace/dt_grammar.y

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -114,28 +114,28 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %token <long> INT "integer"
 %token <std::string> STACK_MODE "stack_mode"
 
-%type <std::string> c_definitions
-%type <ast::ProbeList *> probes
-%type <ast::Probe *> probe
-%type <ast::Predicate *> pred
-%type <ast::Ternary *> ternary
-%type <ast::StatementList *> block stmts block_or_if
-%type <ast::Statement *> if_stmt block_stmt stmt semicolon_ended_stmt compound_assignment jump_stmt loop_stmt tuple_assignment
-%type <ast::Expression *> expr
-%type <ast::Call *> call
-%type <ast::Map *> map
-%type <ast::Variable *> var
-%type <ast::ExpressionList *> vargs
-%type <ast::AttachPointList *> attach_points
-%type <ast::AttachPoint *> attach_point
-%type <std::string> attach_point_def
-%type <ast::PositionalParameter *> param
-%type <std::string> ident
-%type <ast::Expression *> map_or_var
-%type <ast::Expression *> pre_post_op
-%type <ast::Integer *> int
 
-%right ASSIGN
+%type <int> unary_op compound_op
+%type <std::string> attach_point_def c_definitions ident
+
+%type <ast::AttachPoint *> attach_point
+%type <ast::AttachPointList *> attach_points
+%type <ast::Call *> call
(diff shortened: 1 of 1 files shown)


fbs:
I also changed the styling a bit, its now

```
node:
                     rule1
              |      rule2 {}
                     ;
```

Mostly because emacs did it for me but I think it works quite well. All rules start on the same column, independent of how ever long the node name is.

src/parser.yy
```diff
@@ -218,193 +218,293 @@ attach_point_def : attach_point_def ident    { $$ = $1 + $2; }
...
+                tuple_access_expr ASSIGN expr
+                {
+                  error(@1 + @3, "Tuples are immutable once created. Consider creating a new tuple and assigning it instead.");
+                  YYERROR;
+                }
+        |       map ASSIGN expr      { $$ = new ast::AssignMapStatement($1, $3, false, @2); }
+        |       var ASSIGN expr      { $$ = new ast::AssignVarStatement($1, $3, false, @2); }
+        |       map compound_op unary_expr
+                {
+                auto b = new ast::Binop($1, $2, $3, @2);
```
danobi, line 309: indentation
```suggestion
                  auto b = new ast::Binop($1, $2, $3, @2);
```

src/parser.yy
```diff
@@ -218,193 +218,293 @@ attach_point_def : attach_point_def ident    { $$ = $1 + $2; }
...
+                  $$ = new ast::AssignMapStatement($1, b, true, @$);
+                }
+        |       var compound_op unary_expr
+                {
+                auto b = new ast::Binop($1, $2, $3, @2);
```
danobi, line 314: indentation
```suggestion
                  auto b = new ast::Binop($1, $2, $3, @2);
```

src/parser.yy
```diff
@@ -218,193 +218,293 @@ attach_point_def : attach_point_def ident    { $$ = $1 + $2; }
...
-    ;
-
-map_or_var : var { $$ = $1; }
-           | map { $$ = $1; }
-           ;
-
-vargs : vargs "," expr { $$ = $1; $1->push_back($3); }
-      | expr           { $$ = new ast::ExpressionList; $$->push_back($1); }
-      ;
+basic_expr:
```
danobi, line 319: Nit: I believe it's usually called a "primary expression" -- maybe we should stick to that? No strong opinion here

danobi (commented):
At a glance this looks like a nice improvement. The precedence ladder (not sure if that's the right term) looks more like a traditional grammar which I'm sure will help people familiar with the traditional stuff.

src/parser.yy
fbs: yeah forgot to clean that up, primary makes more sense :)

danobi (approved):

PR #1868, opened 2021-06-03
fix typos

ivanov:
I came across a typo in the reference guide, and decided to fix any others that I found. in the docs directory. 

Thanks for making and maintaining bpftrace and friends!


danobi (approved):
Thanks!

PR #1873, opened 2021-06-09
Forbid casting of struct types

### LINKED ISSUE #1869: bpftrace type casting problem
I wrote a bpftrace script as below:

```
#!/usr/local/bin/bpftrace

#include <net/neighbour.h>

k:___neigh_create
{
	$tbl = (struct neigh_table *)arg0;
	$va1 = (uint64)$tbl->gc_thresh1;
	$va2 = (uint64)$tbl->gc_entries;
	printf("aaa: %lx  %lx \n", $va1, $va2);
}
```

I want to take the values of gc_thresh1 and gc_entries of arp_tbl, where gc_entries1 is correct, but when I print gc_entries, I get the address of gc_entries instead of the value.

Looking at the bytecode, the results are as follows:

```
\# bpftool prog dump xlated id 53
0: (bf) r6 = r1
1: (79) r7 = *(u64 *)(r6 +112)
2: (bf) r3 = r7
3: (07) r3 += 244
4: (bf) r1 = r10
5: (07) r1 += -24
6: (b7) r2 = 4
7: (85) call bpf_probe_read#-52880
8: (61) r1 = *(u32 *)(r10 -24)
9: (67) r1 <<= 32
10: (c7) r1 s>>= 32
11: (7b) *(u64 *)(r10 -16) = r1
12: (07) r7 += 420          # Locate offset to gc_entries here
13: (7b) *(u64 *)(r10-8) = r7 # Assignment of the result
14: (b7) r1 = 0
15: (7b) *(u64 *)(r10 -24) = r1
```

In my opinion, C does not allow casts between compound types and primitive types, so it is better to report an error when this is done:
error: aggregate value used where an integer was expected
Instead of giving an address out of nowhere.

```
bpftrace --info

System
  OS: Linux 5.4.56.XXX
  Arch: x86_64

Build
  version: XXX-v0.11.3
  LLVM: 7
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: no
  libbpf: no
  libbpf btf dump: no
  libbpf btf dump type decl: no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: no
  probe_read_user_str: no
  probe_read_kernel: no
  probe_read_kernel_str: no
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
```

viktormalik:
C does not allow casting struct types by value, so we should do the same in BPFtrace.

Resolves #1869.

##### Checklist

### DIFF
--- a/src/ast/semantic_analyser.cpp
+++ b/src/ast/semantic_analyser.cpp
@@ -2007,6 +2007,12 @@ void SemanticAnalyser::visit(Cast &cast)
 {
   cast.expr->accept(*this);
 
+  if (cast.expr->type.IsRecordTy())
+  {
+    LOG(ERROR, cast.loc, err_)
+        << "Cannot cast from struct type \"" << cast.expr->type << "\"";
+  }
+
   bool is_ctx = cast.expr->type.IsCtxAccess();
   auto &intcasts = getIntcasts();
   auto k_v = intcasts.find(cast.cast_type);
@@ -2066,12 +2072,8 @@ void SemanticAnalyser::visit(Cast &cast)
   }
   else
   {
-    auto &etype = cast.expr->type;
-    if (etype.IsIntegerTy() || etype.IsPtrTy())
-    {
-      LOG(ERROR, cast.loc, err_) << "Cannot convert " << etype << " to struct";
-    }
-    cast.type = struct_type;
+    LOG(ERROR, cast.loc, err_)
+        << "Cannot cast to struct type \"" << cast.cast_type << "\"";
   }
   if (is_ctx)
     cast.type.MarkCtxAccess();
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -1363,21 +1363,34 @@ TEST(semantic_analyser, map_cast_types)
 
 TEST(semantic_analyser, variable_casts_are_local)
 {
-  std::string structs = "struct type1 { int field; } struct type2 { int field; }";
-  test(structs + "kprobe:f { $x = (struct type1*)cpu } kprobe:g { $x = (struct "
-                 "type2*)cpu; }",
+  std::string structs =
+      "struct type1 { int field; } struct type2 { int field; }";
+  test(structs + "kprobe:f { $x = *(struct type1 *)cpu } "
+                 "kprobe:g { $x = *(struct type2 *)cpu; }",
        0);
 }
 
 TEST(semantic_analyser, map_casts_are_global)
 {
-  std::string structs = "struct type1 { int field; } struct type2 { int field; }";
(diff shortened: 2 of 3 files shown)


tests/semantic_analyser.cpp
```diff
@@ -1380,6 +1380,14 @@ TEST(semantic_analyser, cast_unknown_type)
   test("kprobe:f { (struct faketype)cpu }", 1);
 }
 
+TEST(semantic_analyser, cast_struct)
+{
+  // Casting struct by value is forbidden
+  test("struct type { int field; }"
+       "kprobe:f { $s = (struct type)cpu; $u = (uint32)$s; }",
```
fbs, line 1387: this seems to test the wrong thing

```
stdin:1:59-74: ERROR: Cannot convert unsigned int64 to struct
struct type { int field; } struct b { int a }; k:f { $s = (struct type )cpu; }
                                                          ~~~~~~~~~~~~~~~
```

src/ast/semantic_analyser.cpp
```diff
@@ -2007,6 +2007,12 @@ void SemanticAnalyser::visit(Cast &cast)
 {
   cast.expr->accept(*this);
 
+  if (cast.expr->type.IsRecordTy())
```
fbs, line 2010: This also blocks struct to struct casts, should we block those?

src/ast/semantic_analyser.cpp
```diff
@@ -2007,6 +2007,12 @@ void SemanticAnalyser::visit(Cast &cast)
...
+  {
+    LOG(ERROR, cast.loc, err_)
+        << "Cannot cast struct type \"" << cast.expr->type << "\"";
```
fbs, line 2013: would be nice to include the target type as well, but that means shuffling a bit

src/ast/semantic_analyser.cpp
viktormalik: I'd say yes, C doesn't allow them either.

src/ast/semantic_analyser.cpp
viktormalik: right, that'll also break some existing tests but we should do that nevertheless

viktormalik:
Updated to block both casts *from* and *to* struct types.

danobi (approved):

PR #1887, opened 2021-06-18
tests: Actually test data_loc output

danobi:
I realized that the previous test would match against "Attaching 1
probe..." output. Which makes the test meaningless. This makes the test
actually test output.

##### Checklist


danobi:
Annoying b/c we can't control if there are any interrupts

PR #1900, opened 2021-06-29
Expand int syntax

### LINKED ISSUE #1897: Support scientific notation in probe
```
stdin:1:1-26: ERROR: Found trailing non-numeric characters: 1e6
hardware:cache-misses:1e7 { @[pid] = count(); }
~~~~~~~~~~~~~~~~~~~~~~~~~
```

Writing 10000000 makes it hard to spot the number

### LINKED ISSUE #1898: Support underscores in numbers as sep, e.g. 1_000_000
A small separator makes all the diff (if used correctly)
```
100000000000
100_000_000_000
```

Rust and ruby both support this kind of syntax, would be nice if we do too :)

fbs:
This will allow us use underscores as int digit separator to improve readability, e.g. 1_000_000 and unifies the int parsing logic allowing us to use the same syntax in attach points. E.g. hardware:cache-misses:1e6 or 1_000 are now valid instead.

The exponent parsing thing feels a bit clunky, c++ supports it out of the box but for doubles only, might be easier to just use that instead. But its hidden behind a simple interface so we can always change it if needed.

Fixes #1897 
Fixes #1898

### DIFF
--- a/src/ast/attachpoint_parser.cpp
+++ b/src/ast/attachpoint_parser.cpp
@@ -7,11 +7,38 @@
 #include <unordered_map>
 #include <vector>
 
+#include "int_parser.h"
 #include "log.h"
 #include "types.h"
 
 namespace bpftrace {
 namespace ast {
+
+std::optional<uint64_t> AttachPointParser::stoull(const std::string &str)
+{
+  try
+  {
+    return int_parser::to_uint(str, 0);
+  }
+  catch (const std::exception &e)
+  {
+    errs_ << e.what();
+    return std::nullopt;
+  }
+}
+
+std::optional<int64_t> AttachPointParser::stoll(const std::string &str)
+{
+  try
+  {
+    return int_parser::to_uint(str, 0);
+  }
+  catch (const std::exception &e)
+  {
+    errs_ << e.what();
+    return std::nullopt;
+  }
+}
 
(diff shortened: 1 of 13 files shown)


fbs:
TODO:

- [] refguide
- [] manpage
- [] changelog

src/ast/attachpoint_parser.cpp
```diff
@@ -5,14 +5,42 @@
 #include <functional>
 #include <string>
 #include <unordered_map>
+#include <variant>
```
danobi, line 8: Accidental include? Don't see any variant usage in this commit

src/ast/int_parser.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <variant>
+
+#include "int_parser.h"
+
+namespace {
+
+template <typename T>
+T _parse_int(const std::string &num, size_t *idx, int base)
+{
+  throw std::runtime_error("parse_int not implemented for type");
```
danobi, line 14: I think we can make this a compile time error instead. Example: https://godbolt.org/z/sEs3xhac4

fbs: Ah yeah that works :)

danobi (approved):

PR #1903, opened 2021-06-30
tools: add new tool dropwatch.bt

liuhangbin:
This is a bpftrace version of dropwatch. It's used to show where are
packets are dropped. Both software and hardware stack are supported.

Signed-off-by: Hangbin Liu <liuhangbin@gmail.com>

### DIFF
--- a/tools/hwdrop.bt
+++ b/tools/hwdrop.bt
@@ -1,0 +1,52 @@
+#!/usr/bin/bpftrace
+/*
+ * hwdrop.bt	Trace kernel packets hardware drops.
+ *		For Linux, uses bpftrace and eBPF.
+ *
+ * USAGE: hwdrop.bt
+ *
+ * This provides kernel packets hardware drops with interface and trap info.
+ *
+ * Copyright (c) 2021 Hangbin Liu.
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ */
+
+#include <linux/devlink.h>
+#include <net/devlink.h>
+
+// from net/core/devlink.c
+struct devlink_trap_item {
+        struct devlink_trap *trap;
+        // [...]
+};
+
+BEGIN
+{
+	printf("Tracing kernel hardware packets drops. Hit Ctrl-C to end.\n");
+}
+
+kprobe:devlink_trap_report
+{
+	$devlink = (struct devlink *)arg0;
+	$skb = (struct sk_buff*)arg1;
+	$trap_item = (struct devlink_trap_item *)arg2;
+	$devlink_port = (struct devlink_port *)arg4;
+
+	/* The same with dropwatch, we only ignore CONTROL type, see
+	 * include/uapi/linux/devlink.h: enum devlink_trap_type
+	*/
+	if ($trap_item->trap->type == DEVLINK_TRAP_TYPE_CONTROL) {
+		return;
+	}
+
+	if ($devlink_port->type == DEVLINK_PORT_TYPE_ETH) {
+		$dev = (struct net_device *)$devlink_port->type_dev;
+		printf("drop at %s, input port name %s, ifindex %d, dev driver %s, protocol 0x%02x, skb len %u\n",
+		       str($trap_item->trap->name), $dev->name, $dev->ifindex,
+		       str($devlink->dev->driver->name), $skb->protocol, $skb->len);
+	} else {
+		printf("drop at %s, dev driver %s, protocol 0x%02x, skb len %u\n",
+		       str($trap_item->trap->name), str($devlink->dev->driver->name),
+		       $skb->protocol, $skb->len);
+	}
+}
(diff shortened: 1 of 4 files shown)


liuhangbin:
v2: use tracepoint:skb:kfree_skb to catch skb drops in net_tx_action()

fbs:
Looks ok to me :)

fbs:
@danobi @viktormalik  // @brendangregg . We don't often merge new tools but this looks ok to me and I dont see why we shouldn't include it, so lets make a decision on this

viktormalik:
LGTM, but there seems to be a problem with positional parameters which causes incorrect behaviour of the script:
```
$ bpftrace tools/dropwatch.bt
Attaching 5 probes...
Tracing kernel hw packets drops. Hit Ctrl-C to end.
```
while the default should be tracing of software drops. This seems to be a bpftrace issue, I'll investigate it and post a fix.

Also, I'm not very familiar with the used attachpoints, so someone else should review to confirm that the tool does what is expected.

brendangregg:
This prints many "dropped" packets on my idle laptop. E.g.:

```
@[
    kfree_skb+118
    kfree_skb+118
    hci_event_packet+495
    hci_rx_work+397
    process_one_work+488
    worker_thread+77
    kthread+276
    ret_from_fork+31
]: 25

```

Is this a problem?...I don't think so. I don't think these are "dropped," I think this is just showing freed packets. AFAIK "dropped" is a network term where a device drops a packet instead of transmitting it, causing retransmits and performance issues. So yes, we like to study dropped packets. But I don't think that's what this tool shows. Please explain how you filter out normal packet transmission and only filter on problem packets.

tcpdrop.py from bcc is meant to focus on (actually) dropped packets. I don't think it works properly either, but I've proposed how to fix it (ensuring tcp_drop() is called when needed, and adding a tracepoint with a reason field) and that's the direction we are going. See https://github.com/iovisor/bcc/issues/3533 .  Unless someone can explain why dropwatch is actually working, I think the right direction is to close this PR and port tcpdrop from bcc once it's done (it'll be dependent on a later kernel version).

(There's a lot of observability tools out there that are and have always been broken.)

liuhangbin:
> This prints many "dropped" packets on my idle laptop. E.g.:
> 
> ```
> @[
>     kfree_skb+118
>     kfree_skb+118
>     hci_event_packet+495
>     hci_rx_work+397
>     process_one_work+488
>     worker_thread+77
>     kthread+276
>     ret_from_fork+31
> ]: 25
> ```
> 
> Is this a problem?...I don't think so. I don't think these are "dropped," I think this is just showing freed packets.

I'm not familiar with Bluetooth, maybe it should use [consume_skb()](https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/tree/net/core/skbuff.c?h=v5.14-rc5#n896) if it's not intended to drop the skb? Because kfree_skb() assumes that the frame is being dropped after a failure. There may have false positives due to some kernel code. But it is still the best place to track dropped pkgs. And the the kernel [dropwatch[(https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/tree/net/core/drop_monitor.c?h=v5.14-rc5#n1134) does the same.

> 
> tcpdrop.py from bcc is meant to focus on (actually) dropped packets. I don't think it works properly either, but I've proposed how to fix it (ensuring tcp_drop() is called when needed, and adding a tracepoint with a reason field) and that's the direction we are going. See [iovisor/bcc#3533](https://github.com/iovisor/bcc/issues/3533) . Unless someone can explain why dropwatch is actually working, I think the right direction is to close this PR and port tcpdrop from bcc once it's done (it'll be dependent on a later kernel version).

tcpdrop.bt is more Accurate and specific. But users may like some more generic drops watch on like IP, UDP layer.

Thanks
Hangbin

brendangregg:
Ok, so if someone can reproduce in real life, then the devlink_trap_report() trace sounds useful. But I'd rename the tool to hwdrop.bt, to make it clear that it's watching hardware drops (or devdrop.bt for device drops).

Anyone have an idea about reproducing hardware drops? Ideally I like to see: given these 23 known events, the tracing tool has 23 lines of output. That would give me confidence it works.

danobi:
I have not reviewed the tools for correctness but I think this is pretty useful. It's unclear what the path forward here should be.

IIRC during office hours last week we discussed creating a bpftrace org and having a tool-shed for tools like this. Perhaps that's the way forward?

liuhangbin:
> IIRC during office hours last week we discussed creating a bpftrace org and having a tool-shed for tools like this. Perhaps that's the way forward?

If you just want to try a bpftrace tool with same function as dropwatch. You can try [retis](https://github.com/retis-org/retis) . See the [Drop monitoring](https://github.com/retis-org/retis#drop-monitoring) usage.

danobi:
https://github.com/iovisor/bpftrace/pull/2437 is a good candidate

danobi:
Please see https://github.com/iovisor/bpftrace/issues/2829

danobi:
Repo has been created: https://github.com/bpftrace/user-tools

Please send PRs!

PR #1911, opened 2021-07-02
Fix time in biolatency_example.txt

xiaoxuanzi:
From the source code, I found that the unit of data in usecs is microsecond, so I think 1k represents 1000. Is it right? I not sure since i am beginner.

### DIFF
--- a/tools/biolatency_example.txt
+++ b/tools/biolatency_example.txt
@@ -21,7 +21,7 @@ Tracing block device I/O... Hit Ctrl-C to end.
 [128K, 256K)          10 |@                                                   |
 
 While tracing, this shows that 426 block I/O had a latency of between 1K and 2K
-usecs (1024 and 2048 microseconds), which is between 1 and 2 milliseconds.
+usecs (1000 and 2000 microseconds), which is between 1 and 2 milliseconds.
 There are also two modes visible, one between 1 and 2 milliseconds, and another
 between 8 and 16 milliseconds: this sounds like cache hits and cache misses.
 There were also 10 I/O with latency 128 to 256 ms: outliers. Other tools and


fbs:
The `hist` function creates a log2 histogram of the data. So 1024..2048 as its 2^10 and 2^11.

This shows it:

```
sudo bpftrace -e 'BEGIN { @=hist(1024); @=hist(1025); @=hist(2047); exit(); }' -q

@:
[1K, 2K)               3 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|

---

sudo bpftrace -e 'BEGIN { @=hist(1023); @=hist(1025); @=hist(2049); exit(); }' -q

@:
[512, 1K)              1 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[1K, 2K)               1 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[2K, 4K)               1 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
```

But you're right, `Ki` would've been better here.

Closing this MR as its incorrect but as docs can always be improved maybe you can add a note to `hist` or map printing section to clarify that `K` here is 2^10?

Thanks!

xiaoxuanzi:
@fbs  Thank you so much for your detailed explanation.<br>
But I had another question, I didn't understand why you say: "which is between 1 and 2 milliseconds."<br>
I maintain that "1024 and 2048 microseconds" is not between 1 and 2 milliseconds.You mean approximation?  <br>

PR #1913, opened 2021-07-02
Add script that generates release assets

fbs:
Bit hacky but atleast we'll have consistent tars. Future version could
get the binary for us.

### DIFF
--- a/docs/release_process.md
+++ b/docs/release_process.md
@@ -42,9 +42,11 @@ You must do these things to formally release a version:
    as "commits"), then "Draft a new release". The tag version and release title
    should be the same and in `vX.Y.Z` format. The tag description should
    be the same as what you added to `CHANGELOG.md`.
-1. Attach a tar of the tools to the release.
 1. Once the release tag pipeline has finished, extract the bpftrace binary from
    the release build on and attach it to the release.
   -  `docker run -v
 $(pwd):/output quay.io/iovisor/bpftrace:vXX.YY.ZZ /bin/bash -c "cp
 /usr/bin/bpftrace /output"`
+1. Run `scripts/create-assets.sh` from bpftrace root dir and attach the
+   generated archives to the release.
+
--- a/scripts/create-assets.sh
+++ b/scripts/create-assets.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+ZSTDFLAGS="-19"
+
+function err() {
+  echo >&2 "$(tput setaf 1)ERROR: $*$(tput sgr0)"
+  exit 1
+}
+function info() {
+  echo "$(tput bold)$*$(tput sgr0)"
+}
+
+[[ -d tools ]] || err "'tools' directory not found, run script from bpftrace root dir"
+[[ -d man ]] || err "'man' directory not found, run script from bpftrace root dir"
+[[ -f bpftrace ]] || err "bpftrace binary not found, download the build artifact"
+command -v zstd >/dev/null 2>&1 || err "zstd command not found, required for release"
+
+OUT=$(mktemp -d) || err "Failed to create temp dir"
+TMP="${OUT}/tmp"
+
+echo "Using '$OUT' as assert dir"
+
+set -e
+
+info "Creating tools archive"
+
+# bit of copying to avoid confusing tar flags not great but works
+mkdir -p "$TMP/bin"
+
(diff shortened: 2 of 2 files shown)


scripts/create-assets.sh
```diff
@@ -0,0 +1,49 @@
...
+
+info "Creating man archive"
+mkdir -p "$TMP/share/man/man8"
+cp man/man8/*.8 "$TMP/share//man/man8/"
+gzip "$TMP/share/man/man8/"*
+tar --xz -cf "$OUT/man.tar.xz" -C "$TMP/share" man
+
+info "Creating bundle"
+ln bpftrace "$TMP/bin/bpftrace"
+tar --xz -cf "$OUT/bundle.tar.xz" -C "$TMP" bin share
```
fbs, line 46:
```suggestion
tar --xz -cf "$OUT/binary+tools+man.tar.xz" -C "$TMP" bin share
```

not sure..

scripts/create-assets.sh
```diff
@@ -0,0 +1,49 @@
...
+[[ -f bpftrace ]] || err "bpftrace binary not found, download the build artifact"
+
+OUT=$(mktemp -d) || err "Failed to create temp dir"
+TMP="${OUT}/tmp"
+
+echo "Using '$OUT'"
+
+
+set -e
+
```
fbs, line 23: considering `xz`ing the binary as well, so we have bpftrace.xz as artifact.

scripts/create-assets.sh
```diff
@@ -0,0 +1,49 @@
...
+  exit 1
+}
+
+function info() {
+  echo "$(tput bold)$*$(tput sgr0)"
+}
+
+[[ -d tools ]] || err "'tools' directory not found, run script from bpftrace root dir"
+[[ -d man ]] || err "'man' directory not found, run script from bpftrace root dir"
+[[ -f bpftrace ]] || err "bpftrace binary not found, download the build artifact"
```
danobi, line 14: Might be better to figure out repo root from script path (`$0`) so script can be run from another dir.

docs/release_process.md
```diff
@@ -42,9 +42,11 @@ You must do these things to formally release a version:
    as "commits"), then "Draft a new release". The tag version and release title
    should be the same and in `vX.Y.Z` format. The tag description should
    be the same as what you added to `CHANGELOG.md`.
-1. Attach a tar of the tools to the release.
 1. Once the release tag pipeline has finished, extract the bpftrace binary from
    the release build on and attach it to the release.
   -  `docker run -v
 $(pwd):/output quay.io/iovisor/bpftrace:vXX.YY.ZZ /bin/bash -c "cp
 /usr/bin/bpftrace /output"`
+1. Run `scripts/create-assets.sh` from bpftrace root dir  and attach the
```
danobi, line 50:
```suggestion
1. Run `scripts/create-assets.sh` from bpftrace root dir and attach the
```

scripts/create-assets.sh
danobi: I think verbose is better

scripts/create-assets.sh
danobi: If we're going to compress might as well go w/ best algorithm (zstd maybe?)

fbs: guess we could just do both. Zstd is not that common yet, will probably confuse people. If we XZ it we still have a nice saving

scripts/create-assets.sh
fbs: I considered it but was hoping to keep this script minimal in options. I only need this a few times a year, copying the binary/symlinking the binary isn't that much work

fbs:
Added zstd version of the bundle and binary for the coolkids that prefer zstd. 

Size diff is minimal

```
 13M	binary_tools_man-bundle.tar.xz
 14M	binary_tools_man-bundle.tar.zst
 46M	bpftrace
 13M	bpftrace.xz
 14M	bpftrace.zst
 40K	man.tar.xz
 24K	tools-with-help.tar.xz
 12K	tools.tar.xz
```

```
26736	binary_tools_man-bundle.tar.xz
28800	binary_tools_man-bundle.tar.zst
93880	bpftrace
26720	bpftrace.xz
28800	bpftrace.zst
```

danobi:
Ok let's stick to xz then

danobi (approved):

PR #1920, opened 2021-07-13
Memory leak tests

viktormalik:
This PR introduces a simple test suite for memory leaks. It is implemented as a bash script (`tests/memleak-tests.sh`) which expects bpftrace built with address sanitizer (`-fsanitize=address`) and simply checks if the given test cases fail.

Currently, the test cases are included in (and should be added directly to) the script.

The tests are added to the CI with some configurations being skipped:
- LLVM 12, as it seems to contain a memory leak
- embedded Alpine, as it doesn't seem to be able to build with ASAN

Since we need to rebuild bpftrace with ASAN, the CI is prolonged.

Suggestions for further improvements/changes are welcome :-)

##### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,20 @@ jobs:
           BASE: bionic
           VENDOR_GTEST: ON
           BUILD_LIBBPF: ON
+        - NAME: Memleak test (LLVM 11 Debug)
+          TYPE: Debug
+          LLVM_VERSION: 11
+          BASE: bionic
+          RUN_MEMLEAK_TEST: 1
+          RUN_TESTS: 0
+          VENDOR_GTEST: ON
+        - NAME: Memleak test (LLVM 11 Release)
+          TYPE: Release
+          LLVM_VERSION: 11
+          BASE: bionic
+          RUN_MEMLEAK_TEST: 1
+          RUN_TESTS: 0
+          VENDOR_GTEST: ON
     steps:
     - uses: actions/checkout@v2
     - name: Get date
@@ -177,7 +191,9 @@ jobs:
         -v /sys/kernel/debug:/sys/kernel/debug:rw
         -v /lib/modules:/lib/modules:ro
         -v /usr/src:/usr/src:ro
+        -e RUN_TESTS=${RUN_TESTS}
         -e RUN_ALL_TESTS=${RUN_ALL_TESTS}
+        -e RUN_MEMLEAK_TEST="${RUN_MEMLEAK_TEST}"
         -e CMAKE_EXTRA_FLAGS="${CMAKE_EXTRA_FLAGS}"
         -e RUNTIME_TEST_DISABLE="${RUNTIME_TEST_DISABLE}"
         -e VENDOR_GTEST="${VENDOR_GTEST}"
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -11,6 +11,7 @@ EMBED_BUILD_LLVM=${EMBED_BUILD_LLVM:-OFF}
 ALLOW_UNSAFE_PROBE=${ALLOW_UNSAFE_PROBE:-OFF}
 DEPS_ONLY=${DEPS_ONLY:-OFF}
 RUN_TESTS=${RUN_TESTS:-1}
+RUN_MEMLEAK_TEST=${RUN_MEMLEAK_TEST:-0}
 VENDOR_GTEST=${VENDOR_GTEST:-OFF}
 CI_TIMEOUT=${CI_TIMEOUT:-0}
 BUILD_LIBBPF=${BUILD_LIBBPF:-OFF}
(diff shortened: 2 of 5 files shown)


.github/workflows/embedded.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           EMBED_BUILD_LLVM: OFF
           EMBED_USE_LLVM: ON
           RUN_ALL_TESTS: 1
+          RUN_MEMLEAK_TEST: 1
```
mmisono, line 43: Should we build release assets with ASAN?

viktormalik: The way it's done now is that two binaries are built: *bpftrace* and *bpftrace-asan*. We could possibly run `make clean` after the memleak tests which should remove unnecessary files built with ASAN.

mmisono: I only see one build [here](https://github.com/iovisor/bpftrace/actions/runs/1026678998) (for example, `bpftrace-Release-vanilla_llvm+clang+glibc2.27`, which is built with ASAN I guess), or am I missing? I think it's OK to publish two binaries unless it costs too much (e.g., increasing publish time largely or hitting storage limits or something.)

viktormalik: Those artifacts just package the *bpftrace* binary which is built without ASAN (see [here](https://github.com/iovisor/bpftrace/blob/master/.github/workflows/embedded.yml#L147-L154)) so there should be no problem, I believe.

mmisono: ah, I missed that `RUN_MEMLEAK_TEST` let `build.sh` build not only `bpftrace` but also `bpftrace-asan`. Then it seems fine.

mmisono (approved):
Many bugs I fixed before were found with ASAN. Integrating it in CI is nice and this looks good to me.

docker/build.sh
```diff
@@ -56,3 +57,10 @@ if [ $RUN_TESTS = 1 ]; then
     ./tests/bpftrace_test $TEST_ARGS;
   fi
 fi
+
+# Memleak test needs the bpftrace binary with -fsanitize-address
+if [ $RUN_MEMLEAK_TEST = 1 ]; then
+    cmake -DBUILD_ASAN=ON -DBPFTRACE=bpftrace-asan ..
+    make clean && make "$@" -j $(nproc)
+    BPFTRACE_ASAN=$PWD/src/bpftrace-asan ASAN_OPTIONS=alloc_dealloc_mismatch=0 ./tests/memleak-tests.sh
```
danobi, line 65: Should  ASAN_OPTIONS=... instead be inside memleak-tests.sh? That way when someone runs the script manually they get consistent results w/ CI

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to
 #### Added
 - Build time dependency on cereal
   - [#1893](https://github.com/iovisor/bpftrace/pull/1893)
+- Memory leak regression tests
+  - [#1920](https://github.com/iovisor/bpftrace/pull/1920)
```
danobi, line 15: Changelog is for users to read. This change (while great!) should not go into changelog.

danobi:
> The tests are added to the CI with some configurations being skipped:
>
> * LLVM 12, as it seems to contain a memory leak
> * embedded Alpine, as it doesn't seem to be able to build with ASAN

Can they be added to the .yml files but commented out with a note? So that it's easier to figure out why they're not running in the future. And so someone can fix it one day if they see the note.

CHANGELOG.md
viktormalik: Ok, I wasn't sure here, removed.

viktormalik:
> Can they be added to the .yml files but commented out with a note? So that it's easier to figure out why they're not running in the future. And so someone can fix it one day if they see the note.

Good idea. I didn't comment them out but set `RUN_MEMLEAK_TEST: 0` with an appropriate comment.

danobi (approved):

.github/workflows/ci.yml
```diff
@@ -103,34 +113,39 @@ jobs:
           CC: clang-10
           CXX: clang++-10
           RUN_ALL_TESTS: 1
+          RUN_MEMLEAK_TEST: 1
```
fbs, line 116: what do we gain from running this test on every llvm version? This test seems to have a somewhat high overhead as we have to recompile bpftrace, think it be better to have the memleak test as a separate job/workflow.

viktormalik: LLVM is known to contain memory leaks, so this way we know if they affect us on particular versions when we add a new feature. But I agree that the overhead is quite high which is probably more important than the benefit.

viktormalik:
Changed CI to run memleak tests as a separate job for a single LLVM version (11) only, so that we don't have a large overhead. Currently, the tests are not run for embedded builds since that would require larger changes in the workflow YAML file.

fbs:
nice!

PR #1946, opened 2021-07-27
parser improvements

fbs:
- fixes the last r/r
- error when the parser has an r/r
- make it easier to enable debug printing


src/parser.yy
```diff
@@ -20,6 +21,7 @@
 #include <limits>
 #include <regex>
 
+
```
danobi, line 25: Unnecessary change?

danobi (approved):

PR #1948, opened 2021-07-28
parser: move tokens to ast

fbs:
As the parser generates the tokens used in the AST everything depends on
the parser. Any small change in the parser will regenerate the parser
headers and in turn for the ast the recompile which in turn causes all
of bpftrace to recompile. This makes the dev cycle for parser changes
long and painful.

There are work arounds, `make parser` followed by manually linking works
but it would be nice if it just works out of the box.

This is a bit of an overhaul but it reverses the relationship between
the parser and ast. Parser changes now only force the relinking of the
binary but ast changes force the recreation of the parser as the parser
uses Tokens from the ast.

The downside is that we now have to keep both lists in sync, but adding
new symbols to the parser is unlikely and updating is easy so it
shouldn't be an issue.

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -27,7 +27,44 @@ class VisitorBase;
     return new T(*this);                                                       \
   };
 
-class Node {
+enum class JumpType
+{
+  INVALID = 0,
+  RETURN,
+  CONTINUE,
+  BREAK,
+};
+
+enum class Operator
+{
+  INVALID = 0,
+  ASSIGN,
+  EQ,
+  NE,
+  LE,
+  GE,
+  LEFT,
+  RIGHT,
+  LT,
+  GT,
+  LAND,
+  LOR,
+  PLUS,
+  INCREMENT,
+  DECREMENT,
+  MINUS,
+  MUL,
+  DIV,
+  MOD,
+  BAND,
+  BOR,
+  BXOR,
+  LNOT,
+  BNOT,
+};
+
+class Node
+{
 public:
   Node() = default;
   Node(location loc) : loc(loc){};
@@ -210,13 +247,13 @@ public:
   DEFINE_ACCEPT
   DEFINE_LEAFCOPY(Binop)
 
-  Binop(Expression *left, int op, Expression *right, location loc);
+  Binop(Expression *left, Operator op, Expression *right, location loc);
 
   ~Binop();
 
   Expression *left = nullptr;
   Expression *right = nullptr;
-  int op;
+  Operator op;
 
 private:
   Binop(const Binop &other);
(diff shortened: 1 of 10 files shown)


src/ast/ast.h
```diff
@@ -27,7 +27,75 @@ class VisitorBase;
...
   };
 
-class Node {
+
+/* Update with
+
+   echo INVALID=0,; sed -n '/\%token$/,/;$/p' src/parser.yy | awk 'NF == 2 { print $1 "," }'
+
+ */
+enum class Token
```
danobi, line 36: Should `Token` be in ast/ dir? Would it make more sense to have a parser/ dir with all the lexing and parsing files too?

fbs: yeah maybe. Or we could limit the tokens to binop and unop symbols only to make ti less confusing

danobi: From commit message: 

>  it reverses the relationship between the parser and ast

I think we can get rid of weird relationship ordering by just creating a `Tokens.h` file, right?

> Or we could limit the tokens to binop and unop symbols only to make ti less confusing

IMO that sounds more confusing. Probably better to keep it all in one place

fbs: > From commit message:
> 
> > it reverses the relationship between the parser and ast
> 
> I think we can get rid of weird relationship ordering by just creating a `Tokens.h` file, right?

The parser would still use it's own tokens for the lexer->parser communication (also needs locations) so we can't get rid of that completely and still need to keep the list in 2 places which isn't great (not a real issue either)


> > Or we could limit the tokens to binop and unop symbols only to make ti less confusing
> 
> IMO that sounds more confusing. Probably better to keep it all in one place

I don't think it would be bad if we rename it from token to operator.

```
namespace ast {
   enum class Operator
     INCREMENT = 0,
     DECREMENT,
...
...
  }
}
```  

So in the parser it would be:

```
         |       DECREMENT map_or_var { $$ = new ast::Unop(ast::Operator::DECREMENT, $2, @1); } 
```

And usage in the ast

```
  switch (binop.op) {
    case Operator::EQ:
```

danobi: > I don't think it would be bad if we rename it from token to operator.

I guess that makes sense. Just to be clear, this is a separate concern from build times now, right?

fbs: Yes my goal is to reduce the build time after modifying the parser. Although not having to include parser.tab.hh in a lot of places is also a nice cleanup.

src/parser.yy
```diff
@@ -35,10 +35,18 @@ class Node;
...
 
 #include "driver.h"
+#include "lexer.h"
 
 void yyerror(bpftrace::Driver &driver, const char *s);
 %}
 
+/*
+!!!!!
+Any changes here must be applied to `enum class Token` too!
```
danobi, line 45: Outdated comment?

fbs: thanks fixed

PR #1949, opened 2021-07-28
Add some assert()s to StdioSilencer

danobi:
Spent most of a day debugging why `--info` output was truncated. Turns
out the root cause was libbpf[0] closing FD 0 which it did not own.
StdioSilencer's saved stderr FD was sometimes 0 and was being closed
behind its back. This mean that once silenced, stderr could not be
unsilenced.

The asserts helped nail down the issue. Since there's buggy libbpf's in
the wild, these asserts might be helpful for other people.

[0]: https://lore.kernel.org/bpf/5969bb991adedb03c6ae93e051fd2a00d293cf25.1627513670.git.dxu@dxuuu.xyz/T/#u

##### Checklist

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -117,9 +117,13 @@ void StdioSilencer::silence()
 {
   fflush(ofile);
   int fd = fileno(ofile);
+  assert(fd >= 0);
   old_stdio_ = dup(fd);
+  assert(old_stdio_ >= 0);
   int new_stdio_ = open("/dev/null", O_WRONLY);
-  dup2(new_stdio_, fd);
+  assert(new_stdio_ >= 0);
+  int ret = dup2(new_stdio_, fd);
+  assert(ret >= 0);
   close(new_stdio_);
 }
 
@@ -129,7 +133,9 @@ StdioSilencer::~StdioSilencer()
   {
     fflush(ofile);
     int fd = fileno(ofile);
-    dup2(old_stdio_, fd);
+    assert(fd >= 0);
+    int ret = dup2(old_stdio_, fd);
+    assert(ret >= 0);
     close(old_stdio_);
     old_stdio_ = -1;
   }


fbs:
What if we just call dup() a 2nd time if `0 <= old_stdio_ <= 3` like

```
old_stdio_ = dup(fd);
if (old_stdio_ >= 0 && old_stdio_ < 4) {
  int tmp = old_stdio_;
  old_stdio_ = dup(fd);
  close(tmp);
}
```

Its not great either but it's better than just crashing. And it should help against buggy programs closing the usualy streams.

We could even add a warning after the libbpf call by doing something like

```
silencer.silence();
do_libbpf_thingies()
if (fcntl(0, F_GETFD) < 0 && errno = EBADFD)
  cout << "Libbpf bug, please upgrade";
```

> StdioSilencer's saved stderr FD was sometimes 0 and was being closed
behind its back

How does this happen? Do we close stdin somewhered?

danobi:
> What if we just call dup() a 2nd time if 0 <= old_stdio_ <= 3 like

Hmm, but the buggy libbpfs will always close 0, making the next allocated fd 0. So if we really wanted to work around this issue we'd have to do checks like this everywhere we create a new fd.

> Its not great either but it's better than just crashing.

With the assert()s it only crashes in debug builds. End users should have the assert()s compiled out.

> And it should help against buggy programs closing the usualy streams.

Do you mean buggy libraries? bpftrace is an end-user binary so nobody can close our FDs 

> How does this happen? Do we close stdin somewhered?

Most likely libbpf closed fd 0 in a previous iteration of `try_load()`)

danobi:
I'd say this bug in libbpf is really difficult to work around -- there's almost nothing we can do against it other than require newer libbpf be used with bpftrace. This would be one argument in favor of vendoring. We could apply my fix and ship bpftrace with the fix.

The other option is to require the latest libbpf which has BTF_KIND_FLOAT support. This only covers up the bug, though, by making sure parsing btf succeeds.

fbs:
Ah yeah good point, we have way to many calls into libbpf nowdays to work around this. Vendor all the things

PR #1954, opened 2021-07-30
tests: Use less hacky mechanism to synchronize AFTER directive

danobi:
Before, the AFTER directive implicitly required `-v` to be passed to
bpftrace executable. This is hacky bc it's not obvious `-v` was used as
a synchronization mechanism. It also put unnecessary burden on the test
writer to know about `-v` being required.

Commit 2 fixes some bugs the `-v` synchronization covered up. Commit 3
removes `-v` from all the runtime tests.
##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1190,8 +1190,9 @@ int BPFtrace::run(BpfBytecode bytecode)
     }
   }
 
-  if (bt_verbose)
-    std::cerr << "Running..." << std::endl;
+  // Used by runtime test framework to know when to run AFTER directive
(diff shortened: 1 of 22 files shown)


src/bpftrace.cpp
```diff
@@ -1190,8 +1190,9 @@ int BPFtrace::run(BpfBytecode bytecode)
     }
   }
 
-  if (bt_verbose)
-    std::cerr << "Running..." << std::endl;
```
fbs, line 1194: Any reason to remove the running msg completely? It can be nice to know tracing has started. You can add a BEGIN for that ofc, but for cli oneliners thats annoying

fbs (approved):

PR #1964, opened 2021-08-07
NFC: move ast passes to ast/passes dir

fbs:
As the amount of passes grows a bit of structure becomes useful.


src/ast/CMakeLists.txt
```diff
@@ -1,18 +1,20 @@
 add_library(ast
+
```
danobi, line 2: Drop whitespace change?

danobi (approved):

PR #1965, opened 2021-08-09
Visitor improvements

fbs:
Split from #1779.


src/ast/pass_manager.cpp
```diff
@@ -47,25 +47,25 @@ std::unique_ptr<Node> PassManager::Run(std::unique_ptr<Node> node,
...
 
-PassResult PassResult::Error()
+PassResult PassResult::Error(const std::string &pass)
 {
-  return Error("");
+  return PassResult(pass);
 }
 
-PassResult PassResult::Error(const std::string &msg)
+PassResult PassResult::Error(const std::string &pass, int code)
```
danobi: If the idea is to do error checking on specific passes from later passes or in the pass manager, I think it'd be better to make `pass` an enum so we get protection from typos.

That might require a master list of passes, though. I think it'd be ok, what do you think?

fbs: Mostly just to be able to generate better error messages and to make all passes consistent. If we have optional/mayfail passes then a enum would def be useful.

Don't mind adding it in now

danobi: If it's only for error messages then let's keep it simple for now. Once there's a read API then list would be good

danobi (approved):

Issue #1901, opened 2021-06-29
[RFC] Reading debug info from userspace binaries


viktormalik:
### Is your feature request related to a problem? Please describe.

I'm wondering if it makes sense to implement parsing of debug information from binaries.
 
Being able to do so would improve the experience of tracing (not only) userspace programs. Examples of features that would be possible:
- accessing (and printing) `uprobe` arguments by name (issue #1742)
- no need to include userspace headers to access struct fields
- `uprobe:line-number` probes (issue #1743)

### Describe the solution you'd like

Implement support for debug information. While DWARF seems to be the default choice, its support may have some downsides (at the very least introducing a new dependency on a parser library). As an alternative, we could support at least parsing of BTF for userspace programs, which shouldn't introduce any additional dependencies.

Has this been discussed before? Any opinions on this?

danobi:
I think we've had some brief discussion before. IIRC a DWARF -> BTF converter was considered but IMO that's pretty hacky and not likely to work well.

In my opinion, DWARF is the way to go here. We'll support more languages and get access to stuff like location of local variables in frames. Could even uprobe by file and line number and access local variables.

Downside to DWARF is that libdwarf is quite slow and we'll take on another dependency.

fbs:
+1 would be a nice addition to have.

It is another dependency but if we do it right it should be possible to just have it as a cmake flag without riddling the code with ifdefs, this logic can stay contained.

viktormalik:
Ok then, I'll look into this and try to integrate DWARF support.

What about using libdw (from elfutils) instead of libdwarf? It's supposed to be faster, has a nicer API, and we're already using libelf so we have a dependency on elfutils anyway.

danobi:
> What about using libdw (from elfutils) instead of libdwarf? It's supposed to be faster, has a nicer API, and we're already using libelf so we have a dependency on elfutils anyway.

Sounds perfect, then.

mmisono:
FYI perf-probe has dwarf support, and it uses dwarf for not only arguments access but also stack unwinding. IIRC it works both kprobes and uprobes. (though I think we can't refer to the implementation since its license is GPL.)

danobi:
I wrote up an initial design for DWARF based stack walks: https://dxuuu.xyz/stack-symbolize.html

mmisono:
I forgot to mention, but perf-probe also utilizes DWARF to access local variables.

Example:

The source file
```c
#include <stdio.h>
int main(){
[...]

PR #1966, opened 2021-08-09
Initial support for DWARF parsing (listing uprobe args)

viktormalik:
This PR introduces support for parsing DWARF information from traced binaries. More on this topic can be found in #1901. I didn't implement all the mentioned features at once, rather I wanted to create an initial version of a DWARF parser to see if this is a correct way to go. Therefore, this PR only introduces a very basic feature - listing `uprobe` arguments (along with their types).

For example:
```
# bpftrace -lv 'uprobe:tests/testprogs/uprobe_test:function*'
uprobe:tests/testprogs/uprobe_test:function1
    int* n
uprobe:tests/testprogs/uprobe_test:function2
    struct Foo* foo1
    struct Foo* foo2
```

The parsing is implemented on top of the libdw library from elfutils. If the library is not present on the system, bpftrace will still compile, only some features won't work (similarly to when BTF is not available).

So far, there are only runtime tests as I haven't come with a reasonable way to mock DWARF data for GTest, yet.

I'd prefer merging this as it is and then start implementing various features on the top of it as separate PRs. If you feel that this PR should already provide more functionality, let me know.

##### Checklist


src/dwarf_parser.h
```diff
@@ -0,0 +1,76 @@
...
+class Dwarf
+{
+public:
+  Dwarf(const std::string &file_path);
+  virtual ~Dwarf();
+
+  std::vector<std::string> get_function_params(
+      const std::string &function) const;
+
+  bool has_data;
```
danobi, line 22: Maybe nicer as read-only getter function?

src/dwarf_parser.h
```diff
@@ -0,0 +1,76 @@
...
+
+private:
+  static bool warned;
+};
+
+bool Dwarf::warned = false;
+
+} // namespace bpftrace
+
+#endif
```
danobi, line 76: Suggestion:
```
#endif // HAVE_LIBDW
```

src/dwarf_parser.h
```diff
@@ -0,0 +1,76 @@
...
+
+  std::string get_type_name(Dwarf_Die &type_die) const;
+
+  Dwfl *dwfl = nullptr;
+  std::string file_path;
+};
+
+} // namespace bpftrace
+
+#else
```
danobi, line 38: Suggestion:
```
#else // HAVE_LIBDW
```

src/dwarf_parser.cpp
```diff
@@ -0,0 +1,166 @@
...
+  for (auto &param : function_param_dies(function))
+  {
+    result.push_back(get_type_name(param.second) + " " + param.first);
+  }
+  return result;
+}
+
+} // namespace bpftrace
+
+#endif
```
danobi, line 166: Suggestion:
```
#endif // HAVE_LIBDW
```

src/dwarf_parser.h
```diff
@@ -0,0 +1,76 @@
...
+    return {};
+  }
+
+  bool has_data = false;
+
+private:
+  static bool warned;
+};
+
+bool Dwarf::warned = false;
```
danobi, line 72: Not sure you can do this in a header. Might be an ODR violation. Move to .cpp instead?

See https://stackoverflow.com/questions/185844/how-to-initialize-private-static-members-in-c

src/probe_matcher.cpp
```diff
@@ -439,6 +441,23 @@ void ProbeMatcher::list_probes(ast::Program* prog)
...
+            if (dwarf.has_data)
+              param_lists.emplace(match, dwarf.get_function_params(fun));
+            else
+            {
+              if (warned_paths.insert(path).second)
+                LOG(WARNING) << "No DWARF found for \"" << path << "\""
+                             << ", cannot show parameter info";
+            }
+          }
+        }
```
danobi, line 460: Extract out into helper function like the other ones?

danobi (changes requested):
Looks like a great start! I don't know much about DWARF so I only skimmed those bits. Seemed reasonable to me. The only thought I had was if the code should use "???" or something for all unknown data. Given that DWARF supports many languages, I wouldn't be surprised if users bumped up against all kinds of weird stuff. Would be nice to have uniform "unknown" markers.

And one other thing: I think libdw-dev should be added to all the dockerfiles (for CI consistency).

BTW, wasn't there mention of libdw being included in elfutils? So I thought we didn't need to install more packages.

viktormalik:
> Looks like a great start! I don't know much about DWARF so I only skimmed those bits. Seemed reasonable to me. The only thought I had was if the code should use "???" or something for all unknown data. Given that DWARF supports many languages, I wouldn't be surprised if users bumped up against all kinds of weird stuff. Would be nice to have uniform "unknown" markers. 

I used `<unknown type>` here, not sure if we deal with potentially unknown data somewhere else. If you prefer using "???" instead, I have no problem with it.

> And one other thing: I think libdw-dev should be added to all the dockerfiles (for CI consistency).

Right, I added it to Dockerfiles for Ubuntu and Fedora (no need to add it to Alpine, see below). 

There's a problem with static builds though. Linking against static libdw is not recommended ([see here](https://launchpad.net/ubuntu/bionic/+package/libdw-dev)) as it requires also linking against libebl.a. That one is usually included in the same package, so it should be doable, but I wanted to check your opinion first - do you prefer including libdw in static builds even for the price of having to add one more library to CMake?

> BTW, wasn't there mention of libdw being included in elfutils? So I thought we didn't need to install more packages.

Yeah, it's part of elfutils, but some distributions (Fedora, Ubuntu) ship in separately while others (Alpine) ship all elfutils in a single package.

src/dwarf_parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  Dwarf_Addr bias;
+  has_data_ = dwfl_nextcu(dwfl, NULL, &bias) != NULL;
+}
+
+Dwarf::~Dwarf()
+{
+  dwfl_end(dwfl);
+}
+
+int get_func_die_cb(Dwarf_Die *func_die, void *arg)
```
danobi, line 46:
```suggestion
static int get_func_die_cb(Dwarf_Die *func_die, void *arg)
```

src/dwarf_parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  while ((cudie = dwfl_nextcu(dwfl, cudie, &cubias)) != nullptr)
+  {
+    if (dwarf_getfuncs(cudie, get_func_die_cb, &func_info, 0) != 0)
+      return func_info.die;
+  }
+
+  return std::nullopt;
+}
+
+Dwarf_Die type_of(Dwarf_Die &die)
```
danobi, line 72:
```suggestion
static Dwarf_Die type_of(Dwarf_Die &die)
```

src/dwarf_parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  dwfl = dwfl_begin(&callbacks);
+  if (dwfl_report_offline(dwfl, file_path.c_str(), file_path.c_str(), -1) ==
+      NULL)
+  {
+    LOG(WARNING) << "Could not read DWARF info from " + file_path;
+  }
+  dwfl_report_end(dwfl, NULL, NULL);
+
+  Dwarf_Addr bias;
+  has_data_ = dwfl_nextcu(dwfl, NULL, &bias) != NULL;
```
danobi, line 38: Nit incoming:

As an alternative to `has_data()`, we can make the `Dwarf` constructor private and have a `friend` factory method return a `std::optional<Dwarf>`. If the factory method returns an instance, then the callers may assume it contains data. This has a nice property where `Dwarf` methods can assume the instance contains data and skip `has_data()` checks.

This alternative has its own drawbacks in that the class contract is a bit different. And `BTF` class also does the `has_data()` thing. So I don't mind either way -- just something to consider.

src/dwarf_parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+Dwarf_Die type_of(Dwarf_Die &die)
+{
+  Dwarf_Attribute attr;
+  Dwarf_Die type_die;
+  dwarf_formref_die(dwarf_attr_integrate(&die, DW_AT_type, &attr), &type_die);
+  return type_die;
+}
+
+std::map<std::string, Dwarf_Die> Dwarf::function_param_dies(
```
danobi, line 80: If you don't need ordering, prefer `std::unordered_map` -- it'll be faster. And I don't think we need ordering here.

danobi:
> I used <unknown type> here, not sure if we deal with potentially unknown data somewhere else. If you prefer using "???" instead, I have no problem with it.

Let's keep `<unknown type>` for now -- better to be verbose I guess.

src/dwarf_parser.cpp
viktormalik: I like this since it is more transparent for users of the `Dwarf` class.

But, in order to manage libdw objects correctly, we don't want to allow copying `Dwarf` instances around. So, I prefer to return `std::unique_ptr<Dwarf>` instead of `std::optional<Dwarf>`.

src/dwarf_parser.cpp
```diff
@@ -0,0 +1,164 @@
...
+      NULL)
+  {
+    LOG(WARNING) << "Could not read DWARF info from " + file_path;
+  }
+  dwfl_report_end(dwfl, NULL, NULL);
+}
+
+std::unique_ptr<Dwarf> Dwarf::GetFromBinary(const std::string &file_path)
+{
+  std::unique_ptr<Dwarf> dwarf(new Dwarf(file_path));
```
danobi, line 33: Nit: more idiomatic and cuts down memory allocations from 2 -> 1 (not that we really care about perf here)
```suggestion
  auto dwarf = std::make_unique<Dwarf>(file_path);
```

danobi (approved):

src/dwarf_parser.cpp
viktormalik: Unfortunately, `make_unique` cannot access the private constructor from a static method.

fbs (approved):

src/dwarf_parser.cpp
danobi: Ah too bad

danobi (approved):

Issue #1933, opened 2021-07-21
remove support for old bcc


fbs:
we still try to support old bcc versions (we have some objdump parsing code) but I'm not convinced it works well. I suggest we drop support for bcc < 0.1x, similar to #1874  . This should also ~fix~ prevent #1929

jordalgo:
Closing now as this is old and we're planning on removing bcc completely. 
https://github.com/orgs/bpftrace/projects/11

PR #1971, opened 2021-08-11
uprobes: automatic resolution of library paths

viktormalik:
This introduces an option to attach to library functions by specifying just the library name instead of a full path. For example:
```
# bpftrace -e 'uprobe:libc:malloc { printf("Allocated %d bytes\n", arg0); }'
Allocated 4 bytes                              
...  
```

One of the use-cases is a better portability of scripts that attach to library functions (as the full library paths may differ across systems). For example, the current `tools/gethostlatency.bt` doesn't work on Fedora.

Unfortunately, our shipped tools cannot be updated right away since this feature relies on BCC's `bcc_procutils_which_so` which is not available on old BCC versions, especially not those used in the CI. If #1933 was resolved, it would be possible.

##### Checklist


vchuravy:
Does this also work for libraries that are dlopened?

viktormalik:
> Does this also work for libraries that are dlopened?

The library paths are searched for in `/etc/ld.so.cache` so it depends whether the dlopen'ed library's there. According to the [man page](https://man7.org/linux/man-pages/man8/ldconfig.8.html), the cache should contain libraries from standard paths and from directories listed in `/etc/ld.so.conf`.

docs/reference_guide.md
```diff
@@ -1108,6 +1108,14 @@ Attaching 1 probe...
 WARNING: could not determine instruction boundary for uprobe:bin:4377 (binary appears stripped). Misaligned probes can lead to tracee crashes!
 ```
 
+When tracing libraries, it is sufficient to specify the library name instead of
+a full path.
```
danobi, line 1112: Maybe also mention the lookup is done via ldcache? This may be important for systems that have multiple versions of a library installed.

tests/runtime/engine/parser.py
```diff
@@ -115,7 +115,7 @@ def __read_test_struct(test, test_suite):
             elif item_name == 'ARCH':
                 arch = [x.strip() for x in line.split("|")]
             elif item_name == 'REQUIRES_FEATURE':
-                features = {"loop", "btf", "probe_read_kernel", "dpath", "uprobe_refcount", "signal",  "iter:task", "iter:task_file"}
+                features = {"loop", "btf", "probe_read_kernel", "dpath", "uprobe_refcount", "signal",  "iter:task", "iter:task_file", "libpath_resolv"}
```
danobi, line 118: Could you break this up into multiple lines, please?

tests/runtime/probe
```diff
@@ -86,6 +86,13 @@ RUN {{BPFTRACE}} -e 'uprobe:/bin/bash:10 { printf("arg0: %d\n", arg0); exit();}'
 EXPECT Could not resolve address: /bin/bash:0xa
 TIMEOUT 5
 
+NAME uprobe_library
+RUN {{BPFTRACE}} -e 'uprobe:libc:write { printf("size: %d\n", arg0); exit(); }'
+EXPECT size: [0-9]+
+TIMEOUT 5
+AFTER cat /etc/passwd > /dev/null
```
danobi, line 93: Can you write a testprog for this instead? Don't want this test to flake out on us in the future if cat changes.

danobi (commented):
Great feature!

docs/reference_guide.md
fbs: dont forget to mirror this to the manpage too

viktormalik: Good points. Updated, please check.

danobi (approved):

PR #1979, opened 2021-08-23
Unify and fix string comparison codegen

### LINKED ISSUE #1978: Incorrect string literal comparisons
It seems that string equality comparisons are incorrect when one of the operands is a literal as they succeed if one string is a prefix of the another (including when one of the strings is empty).

The reason for this behaviour is that bpftrace uses `strncmp` which ends the comparison with success once one of the compared strings ends.

### What reproduces the bug?

```
$ bpftrace -e 'BEGIN { if ("hello" == "hell") { printf("error\n"); } }'
Attaching 1 probe...
error

$ bpftrace -e 'BEGIN { if ("hell" == "hello") { printf("error\n"); } }'
Attaching 1 probe...
error

$ bpftrace -e 'BEGIN { $x = "hello"; if ($x == "hell") { printf("error\n"); } }'
Attaching 1 probe...
error

bpftrace -e 'BEGIN { $x = "hello"; if ($x == "") { printf("error\n"); } }'
Attaching 1 probe...
error
```

I believe that this is not an expected behaviour.

The only thing that works is comparison of 2 variables:
```
bpftrace -e 'BEGIN { $x = "hello"; $y = "hell"; if ($x == $y) { printf("error\n"); } }'
Attaching 1 probe...
```

viktormalik:
Fixes #1978.

There were two versions of string comparison codegen (`CreateStrncmp`) - one for comparing two on-stack strings and one for comparing an on-stack string with a string literal. The version for comparing with literals was wrong and caused the linked bug.

This PR unifies the two functions into one and fixes the bug.

##### Checklist


tests/runtime/regression
```diff
@@ -52,5 +52,20 @@ TIMEOUT 1
 
 NAME strncmp_checks_for_nul
 RUN {{BPFTRACE}} -e 'BEGIN { printf("_%d_\n", strncmp("one", "one ", 4)); exit() }'
-EXPECT _0_
+EXPECT _1_
```
viktormalik, line 55: I believe that the expected result was incorrect as the two strings are not equal. `strncmp` in C also returns a non-zero result in this case.

danobi (approved):
Nice! Thanks

tests/runtime/regression
danobi: Oops, sorry

src/ast/irbuilderbpf.cpp
```diff
@@ -832,15 +738,38 @@ Value *IRBuilderBPF::CreateStrncmp(Value *ctx,
...
+  if (auto constString2 = dyn_cast<ConstantDataArray>(val2))
+    literal2 = constString2->getAsString();
+  else if (isa<ConstantAggregateZero>(val2))
+    literal2 = "";
+  else
+    literal2 = std::nullopt;
 
-  assert(val1p->getElementType()->isArrayTy() &&
-         val1p->getElementType()->getArrayElementType() == getInt8Ty());
+#ifndef NDEBUG
```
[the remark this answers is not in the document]
viktormalik: Hm, I just used what was originally there, but I think that it's correct: if `NDEBUG` is defined (i.e. debug mode is off), don't run the assertions.

fbs (approved):

PR #1982, opened 2021-08-24
Support positional parameters as integer literals

### LINKED ISSUE #1980: strncmp length parameter uses positional parameter's index instead of its value
When using a positional parameter as the length argument to the strncmp function, the positional parameter's index is used instead of its value.

### What reproduces the bug?

This should detect a difference between $1 and $2, but compares only the first 3 characters, so returns 0 when it should return 1:

    # bpftrace -e 'BEGIN {printf("%d\n", strncmp(str($1), str($2), $3)); exit() }' -- aaaa aaab 4
    Attaching 1 probe...
    0

This should detect no difference between the first three characters in $1 and $2, but returns 1 when it should return 0:

    # bpftrace -e 'BEGIN {printf("%d\n", strncmp(str($1), str($2), $4)); exit() }' -- aaaa aaab x 3
    Attaching 1 probe...
    1

### bpftrace --info

```
System
  OS: Linux 5.13.12-arch1-1 #1 SMP PREEMPT Wed, 18 Aug 2021 20:49:03 +0000
  Arch: x86_64

Build
  version: v0.13.0
  LLVM: 12.0.1
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  libbpf: yes
  libbpf btf dump: no
  libbpf btf dump type decl: no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf (depends on Build:libbpf): no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  iter:task: yes
  iter:task_file: yes```

### PR DESCRIPTION
**Pull request by viktormalik**
In all places where an integer literal is expected, we now also accept an integer positional parameter.
These are namely:
- builtin function calls (`lhist`, `buf`, `signal`, `strncmp`)
- array indices.

Fixes #1980.

Note: the only function that is not supported is `str`, since the solution conflicts with the way `str` currently handles positional parameters.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -2196,6 +2196,36 @@ std::string BPFtrace::get_string_literal(const ast::Expression *expr) const
   return "";
 }
 
+std::optional<long> BPFtrace::get_int_literal(const ast::Expression *expr) const
+{
+  if (expr->is_literal)
+  {
+    if (auto *integer = dynamic_cast<const ast::Integer *>(expr))
+      return integer->n;
+    else if (auto *pos_param = dynamic_cast<const ast::PositionalParameter *>(
+                 expr))
+    {
+      if (pos_param->ptype == PositionalParameterType::positional)
+      {
+        auto param_str = get_param(pos_param->n, false);
+        if (is_numeric(param_str))
+          return std::stol(param_str);
+        else
+        {
+          LOG(ERROR, pos_param->loc)
+              << "$" << pos_param->n << " used numerically but given \""
+              << param_str << "\"";
+          return std::nullopt;
+        }
+      }
+      else
+        return (long)num_params();
+    }
+  }
+
+  return std::nullopt;
+}
+
 bool BPFtrace::is_traceable_func(const std::string &func_name) const
 {
 #ifdef FUZZ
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/bpftrace.cpp [hunk 1]
```diff
@@ -2196,6 +2196,37 @@ std::string BPFtrace::get_string_literal(const ast::Expression *expr) const
...
+              << param_str << "\"";
+          return std::nullopt;
+        }
+      }
+      else
+        return (long)num_params();
+    }
+  }
+
+  LOG(ERROR) << "Expected integer literal, got " << expr->type;
```
fbs, line 2226: I think we should drop this error, i see every user of this already has its own error handling. Adding an extra error without location on top of that doesn't sound helpful. 

Or we could return the error too and append that to the error message, that would be useful but somewhat tricky to do. Std::variant works somewhat ok

viktormalik: I agree with dropping the error, it doesn't really say much.

Returning the error and appending it to the message could be useful for the other error in this function (line 2215) as it always prints to the default `stderr` even if the function is called from semantic analyser (which normally prints to its own `err_` stream). The question is whether it is worth the effort.

fbs (approved):

PR #1990, opened 2021-08-27
output: fix non_map_print of struct w/ bitfields

davemarchevsky:
When using `print()` to dump a struct with bitfield members, the value of
the bitfield members was incorrect. For example, using a script to trace
`perf_event_open`, which takes a `struct perf_event_attr` as input:

`bpftrace -e 'tracepoint:syscalls:sys_enter_perf_event_open {print(*args->attr_uptr); print(args->attr_uptr->precise_ip)}'`

Results in output like

```
{ .type = 0, .size = 112, .config = 0, .sample_period = 10000000, .sample_freq = 10000000, .sample_type = 60, .read_format = 8, .disabled = 65537, .inherit = 65537, .pinned = 65537, <SNIP...>, .precise_ip = 256, <SNIP...> }
2
```

on events where only the `disabled` and `precise_ip` bit fields are set (to 1
and 2, respectively). The values output for the `perf_event_attr`
bitfields in this example are doubly incorrect in that they're too large
for the fields in question (`disabled` and most other `perf_event_attr`
bitfields are 1 bit, `precise_ip` is 2).

Printing the bit field directly does work, though, as bpf codegen
generates some special bitfield handling to read only the specific field - see `is_bitfield` check in codegen_llvm.cpp

This patch applies the same bitfield handling logic when printing the
entire struct. As a result, example script output is now consistent and
correct:

```
{ .type = 0, .size = 112, .config = 0, .sample_period = 10000000, .sample_freq = 10000000, .sample_type = 60, .read_format = 8, .disabled = 1, .inherit = 0, .pinned = 0, <SNIP...>, .precise_ip = 2, <SNIP...> }
2
```

Note: This is my first bpftrace commit, so could use some advice as to where to add a test to properly exercise this.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -196,14 +196,43 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
   else if (type.IsRecordTy())
   {
     std::vector<std::string> elems;
+    SizedType field_type;
+    std::vector<uint8_t> elem_data;
+    uint64_t bitfield_data;
     for (auto &field : type.GetFields())
     {
-      std::vector<uint8_t> elem_data(value.begin() + field.offset,
-                                     value.begin() + field.offset +
-                                         field.type.GetSize());
+      if (field.is_bitfield)
+      {
+        elem_data = std::vector<uint8_t>(value.begin() + field.offset,
+                                         value.begin() + field.offset +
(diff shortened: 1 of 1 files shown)


danobi:
Closing due to inactivity. Tracking in #2801

Issue #1742, opened 2021-03-19
Support argument access by name, show arguments in probe list


ema:
### Is your feature request related to a problem? Please describe.
While in the case of static kernel tracepoints arguments [can be accessed by name](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#6-tracepoint-static-tracing-kernel-level-arguments), when it comes to dynamic kernel and userspace tracing (kprobe, uprobe) arguments can only be accessed with the arg0, ... argN positionals. Argument names are more user friendly, make tracing scripts slightly less implemetation-dependant, and improve readability.
 
### Describe the solution you'd like
SystemTap supports accessing arguments by name for all type of probes, and shows which arguments are in scope  together with the data type:

```
$ sudo stap -L 'process("/bin/ls").function("print_name_with_quoting")' 
process("/bin/ls").function("print_name_with_quoting@src/ls.c:4727") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t
```
It would be great if bpftrace could do the same too.

danobi:
We have support for kernel arguments: https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#16-kfunckretfunc-kernel-functions-tracing-arguments

User arguments would require some cost/benefit analysis. Seems like it'd require deep language support (for anything other than C).

ema:
> We have support for kernel arguments: https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#16-kfunckretfunc-kernel-functions-tracing-arguments

Oh, I did not know that, thanks for pointing it out. The distinction between kprobe and kfunc confused me. It seems that kfunc support isn't available on my system?

```
$ sudo bpftrace -e 'kfunc:fget { printf("fd %d\n", args->fd);  }'
stdin:1:1-11: ERROR: kfunc/kretfunc not available for your kernel version.
kfunc:fget { printf("fd %d\n", args->fd);  }
~~~~~~~~~~
stdin:1:32-40: ERROR: Can't find a field
kfunc:fget { printf("fd %d\n", args->fd);  }
```
```
$ sudo bpftrace --info
System
  OS: Linux 5.10.0-5-amd64 #1 SMP Debian 5.10.24-1 (2021-03-19)
  Arch: x86_64

Build
  version: v0.11.3
  LLVM: 11
  foreach_sym: yes
  unsafe uprobe: no
  bfd: no
[...]

PR #1994, opened 2021-09-07
Access to uprobe arguments by name

viktormalik:
Allows to reference uprobe arguments by name, with automatic type resolution:
```
# bpftrace -lv 'uprobe:tests/testprogs/uprobe_test:function1'
uprobe:tests/testprogs/uprobe_test:function1
    int* n
    char c

# bpftrace -e 'uprobe:tests/testprogs/uprobe_test:function1 { printf("%d\n", *(args->n)); }'
Attaching 1 probe...
13
```

Arguments are accessed via the `args` builtin (same as with tracepoints and kfuncs). This feature relies on DWARF to obtain argument names and types, so the traced binary must be compiled with DWARF.

Currently supports only basic types (integers, chars, enums) and pointers to them. Support for C structs will follow, but it'll require more extensive changes, so I'd prefer doing it in a separate PR.

Related issue #1742 (will close it once at least all C types are fully supported).

This PR also introduces unit testing for DWARF-based features.

##### Checklist


tests/data/dwarf_data.h
```diff
@@ -0,0 +1,738 @@
...
+//  }
+//
+// int main(void)
+// {
+//   func_1(0, 0, 0);
+//   return 0;
+// }
+
+// generated by:
+//  % cat a.c
```
fbs, line 50: would be nice to have cmake do this

tests/dwarf_common.h
```diff
@@ -0,0 +1,40 @@
...
+class test_dwarf : public ::testing::Test
+{
+protected:
+  void SetUp() override
+  {
+    char *bin = strdup("/tmp/dwarf_dataXXXXXX");
+    int fd = mkstemp(bin);
+    if (fd < 0)
+    {
+      std::remove(bin);
```
fbs, line 15: is this correct?

fbs:
Nice! Need to find the time to dive into this and test it. But it looks good

tests/dwarf_common.h
viktormalik: It's not, `mkstemp` doesn't create a file on failure...I just followed what is used in `tests/btf_common.h`.

tests/data/dwarf_data.h
viktormalik: Good idea, I tried to do so in the last commit.

Would be nice if we could do the same for BTF data in `tests/data/bft_data.h`. I actually have it prepared, the problem is that it requires pahole version >=1.17 which is not available in our CI environment.

fbs (approved):

src/ast/passes/semantic_analyser.cpp
```diff
@@ -449,11 +449,20 @@ void SemanticAnalyser::visit(Builtin &builtin)
       builtin.type.MarkCtxAccess();
       builtin.type.is_funcarg = true;
     }
+    else if (type == ProbeType::uprobe)
+    {
+      builtin.type = CreatePointer(CreateRecord("struct uprobe_args",
+                                                bpftrace_.structs.Lookup(
+                                                    "struct uprobe_args")),
```
danobi, line 456: This lookup confused me for a bit. Maybe leave a comment along the lines of "Add a dummy struct type here. When the eventual FieldAccess is resolved it, it special cases the argument lookup".

src/dwarf_parser.h
```diff
@@ -11,28 +13,37 @@
...
+                                              const std::string &file_path);
 
   std::vector<std::string> get_function_params(
       const std::string &function) const;
 
+  ProbeArgs resolve_args(const std::string &function);
+
 private:
-  explicit Dwarf(const std::string &file_path);
+  explicit Dwarf(BPFtrace *bpftrace, const std::string &file_path);
```
danobi, line 32: I don't see `bpftrace` ptr being used anywhere in this class. Is this intentional?

tests/CMakeLists.txt
```diff
@@ -43,6 +43,14 @@ add_executable(bpftrace_test
   ${CODEGEN_SRC}
 )
 
+# Regenerate DWARF data for unit tests
+add_custom_target(testing_debuginfo_data
+        COMMAND
+        ${CMAKE_COMMAND}
+        -DDATA_SOURCE_DIR="${CMAKE_SOURCE_DIR}/tests/data"
+        -P ${CMAKE_SOURCE_DIR}/tests/data/generate_debuginfo_data.cmake)
```
danobi, line 51: I'm not sure this is correct. Will this regen the final output if the `data_source.c` is changed?

Instead, how about using [add_custom_command()](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command) with a shell script? The output file is the generated header. The input files are the shell script and `data_source.c`. And then the following line `add_dependencies` stays the same and I think it will "just work". Should be a littler easier to grok/program too.

src/dwarf_parser.h
viktormalik: Ah, good catch, it's not necessary here. I'll be needed once we support struct types, removing for now.

tests/CMakeLists.txt
viktormalik: > I'm not sure this is correct. Will this regen the final output if the `data_source.c` is changed?

Yes, it will. CMake will handle that.

> Instead, how about using [add_custom_command()](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command) with a shell script? The output file is the generated header. The input files are the shell script and `data_source.c`. And then the following line `add_dependencies` stays the same and I think it will "just work". Should be a littler easier to grok/program too.

In fact, using add_custom_command() was my original approach, but I wasn't able to make it work. I just couldn't make CMake create the correct rule for the target file.

Using a shell script instead of the CMake script is possible, but CMake allows to use configure_file() which gives quite a nice control of the contents of the generated header.

danobi: Any chance this example might help? https://github.com/libbpf/libbpf-bootstrap/blob/67a29e511cc9d0a570d4d3b9797827f3a08ccdb5/tools/cmake/FindBpfObject.cmake#L163-L170

I'm ok to merge this as is too

viktormalik: The example suggests to use target_sources() to create the dependency between the target (`bpftrace_test`) and the generated header, but unfortunately this doesn't rebuild `bpftrace_test` when `data_source.c` is changed.

It's probably possible to make it work somehow, but likely it will be more complicated than the current approach.

danobi (approved):

PR #2002, opened 2021-09-09
build: Improve error message on wrong LLVM version

tbaederr:
Instead of just printing $LLVM_INCLUDE_DIRS, additionally print the
version we found and the allowed versions.

Sample output:

```
CMake Error at CMakeLists.txt:171 (message):
  Unsupported LLVM version found via /usr/include: 13


CMake Error at CMakeLists.txt:172 (message):
  Only versions between 6 and 12 are supported


CMake Error at CMakeLists.txt:173 (message):
  Specify an LLVM major version using LLVM_REQUESTED_VERSION=<major version>

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,8 +164,12 @@ else()
     find_package(LLVM REQUIRED)
   endif()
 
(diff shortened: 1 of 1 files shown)


viktormalik (approved):

danobi (approved):
Thanks!

PR #2008, opened 2021-09-16
codegen: Fix interaction with verifier when accessing ctx field twice

danobi:
The verifier doesn't like it if we modify the ctx pointer twice before
dereferencing it. In other words, ctx+8 is OK, but ctx+4+4 is NOT.
The reasoning in kernel land is due to interactions w/ socket buffer
(skb) access rewrites. Without this limitation, the kernel may let the
prog write past the end of the skb.

We trip over this issue when accessing the same tracepoint field twice.
LLVM decides to save ctx+8 in a temporary register and reuse the value
when indexing into the struct in the unrolled memcpy loop.

Fix this issue by using bpf_probe_read_*() so that we only modify the
ctx pointer once.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ and this project adheres to
   - [#1972](https://github.com/iovisor/bpftrace/pull/1972)
 - Fix string comparison codegen
   - [#1979](https://github.com/iovisor/bpftrace/pull/1979)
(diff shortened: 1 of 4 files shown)


danobi:
@viktormalik what do you think?

viktormalik (approved):
LGTM - replacing `memcpy` by `probe_read` when reading strings from ctx should be a safe thing to do.

PR #2010, opened 2021-09-17
Fix AOT build/tests with older BCC versions

danobi:
Recent bcc change: iovisor/bcc@7191918
made libbcc_bpf.so the runtime-only components of bcc. So no LLVM. The
bpftrace 'runtime' target tried to use the runtime-only version of bcc
if available. If not availble, it links to the regular libbcc.so. AOT
runtime currently links against libLLVMSupport.a which causes issues
when also linked against libbcc.so (duplicate registered command line
error on startup).

This PR fixes the AOT build / AOT tests when built against an older BCC.
##### Checklist


danobi:
cc @sumanthkorikkar . I reproduced the original issue in a fedora34 docker image. This PR seems to fix it. Can you check as well?

sumanthkorikkar:
Hi @danobi, 

I tested it both with upstream bcc and fedora bcc. Works.

Tests are skipped for fedora bcc, which is expected as per PR.

Thank you.

PR #2015, opened 2021-09-24
add kvmpagefault.bt to capture KVM page fault

Rtoax:
##### Checklist

### DIFF
--- a/tools/kvmpagefault.bt
+++ b/tools/kvmpagefault.bt
@@ -1,0 +1,30 @@
+#!/usr/bin/env bpftrace
+/*
+ * kvmpagefault.bt	Trace kvm_page_fault
+ *		For Linux, uses bpftrace and eBPF.
+ *
+ * Also a basic example of bpftrace.
+ *
+ * USAGE: kvmpagefault.bt
+ *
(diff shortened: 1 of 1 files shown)


brendangregg:
G'Day; Thanks, although my guess is that this is too niche for bpftrace: we could write 10,000+ similar tools (tracepoint/kprobe + args + stacks), whereas in the bpftrace tools directory we'd only want the top 100 tools, those that are most useful. Having too many tools creates a maintenance and clutter (learnability) burden. If you think it'd be in that top 100, please explain your thoughts. KVM + page faults is a bit niche.

I have my own BPF repo for tools that don't belong in bcc/bpftrace. I've thought about creating an official one for BPF; I'll kick of some discussions about this. The idea is that we have a repo that accepts everything, and then we have the curated "top 100" that make it into the bcc and bpftrace repos.

Rtoax:
> KVM + page faults is a bit niche.

Thank you for your reply, I think you are right. KVM page faults are indeed relatively niche. By the way, you are my idol.

thefallentree:
closing due to previous discussion

PR #2029, opened 2021-10-04
Stop warnings about variables only used in assert statements

ZekeMedley:
Release builds were emitting warnings about unused variables when a
variable was only used in an assert statement. For example:

```cpp
int res = // ...
assert(res >= 0);
```

C++17 has a nice [[maybe_unused]] attribute which we can use here to
silence the warning.

##### Checklist


viktormalik (approved):

danobi (approved):
Thanks!

PR #2034, opened 2021-10-12
Type resolution from DWARF

### LINKED ISSUE #1742: Support argument access by name, show arguments in probe list

viktormalik:
When tracing userspace programs, this extension automatically resolves types from DWARF (if it is available). This has mostly 2 advantages:
- bpftrace program doesn't need to define the types (or include the headers) used in the traced binary
- uprobe arguments of struct type can be accessed by name

Version 2:
Parsing of structure types from DWARF is done in `FieldAnalyser`, which directly fills structs in `StructManager`. `ClangParser` may override these types if user defined some custom types or included some headers.

Version 1 (abandoned):
Since we're relying on Clang parser to do the actual type resolution, we need to dump types from DWARF into compilable C format. Even though there probably exist libraries that can do this (e.g. libdwarves), I decided to implement this by hand, to have more control over the supported types and to avoid yet another dependency.
We now have two sources of type information - BTF and DWARF. To avoid potential collisions, for now we use DWARF for userspace tracing and BTF for kernel tracing.

I also believe that this resolves #1742.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to
   - [#2147](https://github.com/iovisor/bpftrace/pull/2147)
 - Improve include paths resolution
   - [#2149](https://github.com/iovisor/bpftrace/pull/2149)
+- Automatic type resolution from DWARF
+  - [#2034](https://github.com/iovisor/bpftrace/pull/2034)
 - Add builtin function: `bswap`
   - [#2166](https://github.com/iovisor/bpftrace/pull/2166)
 
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1193,8 +1193,7 @@ Syntax:
 uprobe: args->NAME
 ```
 
-The arguments can be accessed by dereferencing `args` and accessing the argument's `NAME`. Currently, only
-simple types (integer, char, bool, enum) and pointers to them are supported.
+The arguments can be accessed by dereferencing `args` and accessing the argument's `NAME`.
 
 The list of function's arguments can be retrieved using the verbose list option:
 ```
--- a/src/ast/passes/field_analyser.cpp
+++ b/src/ast/passes/field_analyser.cpp
@@ -93,14 +93,20 @@ void FieldAnalyser::visit(Map &map)
 
   auto it = var_types_.find(map.ident);
   if (it != var_types_.end())
-    type_ = it->second;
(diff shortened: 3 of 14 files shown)


viktormalik:
> > (e.g. if the user used some of them in his custom types).
> 
> Hmm, is this a likely scenario? What types could the user be defining that are not already available in DWARF? I'm not necessarily against going to C then to clang parser, but I think it should be a conscious tradeoff. If it's significantly less complex to skip C, then we ought to consider that. If it's similar complexity, then going to C has theoretical advantages. What do you think?

I agree that this seems to be a very unlikely situation (in fact, I haven't come with a reasonable example so far). What I'd suggest is:
- if there are no user-defined types, skip the ClangParser and resolve all from DWARF (and we can eventually do the same for BTF, too)
- if there are some user-defined types or includes, use ClangParser with DWARF->C or BTF->C dumping

Does it make sense?

danobi:
Sure, that sounds fine to me

viktormalik:
After some time, coming with version 2.

The main change is that parsing of structure types from DWARF is now done in `FieldAnalyser`, which directly fills structs in `StructManager`. `ClangParser` is not changed by this PR, so it will run if there is something to parse (i.e. user provided some type definitions or included some headers). Since `ClangParser` is run after `FieldAnalyser`, the effect of this is that user may override DWARF-parsed types, if he wishes to.

danobi (approved):
Not very familiar w/ DWARF but nothing stands out to me. Great work!

PR #2036, opened 2021-10-14
Allow resizing empty string assigned to a map

### LINKED ISSUE #2025: BUG: Struct size mismatch
The problem is the size of `""` is 64 (STRING_SIZE) but the size of `comm` is 16 (COMM_SIZE). 
https://github.com/iovisor/bpftrace/blob/45d23c82c36d2467e322e96b532edee33720c997/src/ast/passes/semantic_analyser.cpp#L97 https://github.com/iovisor/bpftrace/blob/45d23c82c36d2467e322e96b532edee33720c997/src/ast/passes/semantic_analyser.cpp#L346

(The difference comes from the fact that type of `""` is `char *` but the `comm` is `char[16].)

A workaround is using `comm` instead of `""` in the `BEGIN{}`.
`sudo bpftrace -e 'BEGIN {@prev = (comm,"")} u:/usr/lib/libc.so.6:printf,u:/usr/lib/libc.so.6:puts / @prev != (comm, str(arg0)) / { printf("%s %s: %s\n", probe, comm, str(arg0)); @prev = (comm, str(arg0)) }`

_Originally posted by @mmisono in https://github.com/iovisor/bpftrace/discussions/2023#discussioncomment-1419674_

---

We should catch this sooner

viktormalik:
Type of an empty string literal is by default `string[64]`, however this may cause problems when assigning it to a map that contains string of a different size.

For example:
```
# bpftrace -e 'BEGIN { @ = "" } k:vfs_read { @ = comm }'
```

Here, `comm` is `string[16]` which causes bpftrace to fail (see #2025 for a real case).

This PR attempts to resolve the above problem by resizing the empty string to the actual size stored inside the map (since empty string fits any string size), including the case when the string is inside a tuple.

The problem is that the implementation is rather complex (introducing a new `SizedType` field, semantic analysis changing an expression type from `visit` of another expression, ...) considering the fact that it only solves an edge case. If we don't want to introduce this amount of complexity, an alternative would be to just emit a correct error in semantic analyser (since the above example now fails during runtime).

Fixes #2025.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to
 #### Changed
 - Prevent LLVM from unrolling loops
   - [#1967](https://github.com/iovisor/bpftrace/pull/1967)
+- Allow resizing empty string assigned to a map
+  - [#2036](https://github.com/iovisor/bpftrace/pull/2036)
 
 #### Deprecated
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -95,6 +95,7 @@ void SemanticAnalyser::visit(String &string)
                                  << " bytes): " << string.str;
   }
   string.type = CreateString(STRING_SIZE);
+  string.type.SetRealSize(string.str.size());
   // @a = buf("hi", 2). String allocated on bpf stack. See codegen
   string.type.SetAS(AddrSpace::kernel);
 }
@@ -2225,6 +2226,8 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
   const std::string &map_ident = assignment.map->ident;
   auto type = assignment.expr->type;
 
+  update_expr_assigned_to_map(map_val_[map_ident], assignment.expr);
+
   if (type.IsRecordTy())
   {
     std::string ty = assignment.expr->type.GetName();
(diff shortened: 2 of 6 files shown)


mmisono:
I wonder if it is easy to use the largest string size assigned to a value as a string size. Then we also could write something like `bpftrace -e 'BEGIN { @ = comm } k:vfs_read { @ = "abc" }'`. (Though in that case, we need to codegen changes to make a string null terminated when copying a string shorter than the max buffer size.)

danobi:
> Type of an empty string literal is by default string[64], however this may cause problems

Could you elaborate on what the exact problem is? Having trouble figuring it out from the PR and I figure you already figured it out :)

viktormalik:
> I wonder if it is easy to use the largest string size assigned to a value as a string size. Then we also could write something like `bpftrace -e 'BEGIN { @ = comm } k:vfs_read { @ = "abc" }'`. (Though in that case, we need to codegen changes to make a string null terminated when copying a string shorter than the max buffer size.)

I was thinking the same. In fact, I was wondering if there's a reason that string literals are always `string[64]` (apart from being unified with other strings). If no, then we could set their size to the actual string length and then set a variable/map type to the larges string being assigned to it.

PR #2039, opened 2021-10-16
Disable llvm 8 build

fbs:
LLVM8 image build fails, we don't really need it anyway so lets just get rid of it.


danobi:
I wonder why we need/want to support multiple LLVM versions for embedded builds anyways. If we always tell users to use the statically linked binary, then why confuse them with multiple options? Seems easier for users and for us to maintain 1 embedded build.

thefallentree:
> Yes this is just the docker build on quay.io. The main CI tests just install the LLVM packages and build on that, that won't be going anytime soon.
> 
> Help would be appreciated but if you have time left we have more important issues than this docker image. As long as nobody deletes the image from quay it will keep on working so very low priority.

so, does embedded here just mean static linking to libLLVM and libClang?  As far as I can tell, llvm packages in ubuntu provides both dynamic lib and static lib, so maybe we can just change this test to use static linking on ubuntu image instead of using alpine image?

Issue #1500, opened 2020-09-03
Getting a cgroup path from cgroup ID


natchaphon-r:
### Is your feature request related to a problem? Please describe.
Use case:
My host machine has multiple running docker containers. Inside a docker container, there is a medic process that can report container health.

I'm looking to extend oomkill.bt to notify the medic process whenever there is an OOM event. bpftrace provides a built-in variable "cgroup" which is a cgroup ID. I'm looking to convert cgroup ID to a full cgroup path. Since the full cgroup path has docker container ID, I can use the path to notify the medic process in the correct container.

### Describe the solution you'd like
A new BPF helper function that calls [task_cgroup_path](https://elixir.bootlin.com/linux/v4.9.12/source/kernel/cgroup.c#L2366)

### Describe alternative solutions or features you've considered
An alternative is to pass cgroup id to a separate program that can convert cgroup id to cgroup path using a similar approach.

danobi:
Getting filesystem paths through bpf is notoriously tricky b/c you need to take locks. IIUC from following the kernel discussions, the main issue is deadlocks b/c the bpf prog could try to take a lock in an already locked context. @olsajiri probably has the most insight into feasibility of a cgroup_path() helper.

cgroup id is just inode # in cgroupfs so one workaround could look like this:
```
$ sudo ./build/src/bpftrace -e 'BEGIN { print(cgroup) }'
Attaching 1 probe...
5505
^C

$ find /sys/fs/cgroup -inum 5505
/sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/gnome-launched-alacritty-9823.scope
```

bwplotka:
> cgroup id is just inode # in cgroupfs so one workaround could look like this:

Hm @danobi .. it does not make much sense, but my `find /sys/fs/cgroup -inum 1730` for one service returns multiple paths.. 🤔 Why? I thought inode number is unique.

```
/sys/fs/cgroup/blkio/system.slice/fwupd.service/blkio.throttle.io_service_bytes_recursive
/sys/fs/cgroup/cpu,cpuacct/system.slice/systemd-logind.service/cpuacct.usage_percpu
/sys/fs/cgroup/cpu,cpuacct/system.slice/systemd-logind.service/cpuacct.usage_percpu
/sys/fs/cgroup/unified/system.slice/docker.service
 ```
 
 Code https://github.com/bwplotka/ebpf_exporter/blob/exp/decoder/cgroup.go#L38

[...]

PR #2055, opened 2021-11-01
Add cgroup_path built-in function

lenticularis39:
The function takes a cgroup id and sends it to userspace via a perf
event, where it gets converted to a cgroup path.

By combining the generic print mechanism with finding the cgroup path in userspace this provides an easy workaround to #1500, avoiding the possibility of an in-kernel deadlock. A limitation is that due to the conversion occuring in userspace, the cgroup path string cannot be manipulated.

I'm also not sure if `print_cgroup_path` is the best name for the function, maybe it would be better to name it `cgrouppath` to be consistent with other such printing functions like `join` that don't include "print" in their name.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -96,6 +96,7 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [26. `uptr()`: Annotate userspace pointer](#26-uptr-annotate-userspace-pointer)
     - [27. `kptr()`: Annotate kernelspace pointer](#27-kptr-annotate-kernelspace-pointer)
     - [28. `macaddr()`: Convert MAC address data to text](#28-macaddr-convert-mac-address-data-to-text)
+    - [29. `cgroup_path()`: Convert cgroup id to cgroup path](#29-cgroup_path-convert-cgroup-id-to-cgroup-path)
 - [Map Functions](#map-functions)
     - [1. Builtins](#1-builtins-2)
     - [2. `count()`: Count](#2-count-count)
@@ -3090,6 +3091,25 @@ Example:
 # bpftrace -e 'kprobe:arp_create { printf("SRC %s, DST %s\n", macaddr(sarg0), macaddr(sarg1)); }'
 SRC 18:C0:4D:08:2E:BB, DST 74:83:C2:7F:8C:FF
 ^C
+```
+
+## 29. `cgroup_path`: Convert cgroup id to cgroup path
+
+Syntax: `cgroup_path(int cgroupid, string filter)`
+
+Converts the given cgroup id into the corresponding cgroup path for each cgroup hierarchy the id
+appears in. Because the conversion is done in user space, the resulting object can only be used for
+printing.
+
+Optionally a string literal may be passed as the second argument to filter cgroup hierarchies to look
+in (interpreted as a wildcard expression).
+
+Example:
+
+```
+# bpftrace -e 'BEGIN { print(cgroup_path(5386)); }'
+Attaching 1 probe...
+unified:/user.slice/user-1000.slice/session-3.scope
 ```
 
 # Map Functions
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1287,6 +1287,34 @@ t:syscalls:sys_enter_execve {
 55f683ee2000-55f683ee3000 rw-p 00000000 00:00 0
 ----
 
+=== cgroup_path
+
+.variants
+* `cgroup_path cgroup_path(int cgroupid, string filter)`
+
+Convert cgroup id to cgroup path.
+This is done asynchronously in userspace when the cgroup_path value is printed,
+therefore it can resolve to a different value if the cgroup id gets reassigned.
+This also means that the returned value can only be used for printing.
+
+A string literal may be passed as an optional second argument to filter cgroup
+hierarchies in which the cgroup id is looked up by a wildcard expression (cgroup2
+is always represented by "unified", regardless of where it is mounted).
+
+The currently mounted hierarchy at /sys/fs/cgroup is used to do the lookup. If
+the cgroup with the given id isn't present here (e.g. when running in a Docker
+container), the cgroup path won't be found (unlike when looking up the cgroup
+path of a process via /proc/.../cgroup).
+
+----
+BEGIN {
+  $cgroup_path = cgroup_path(3436);
+  print($cgroup_path);
+  print($cgroup_path); /* This may print a different path */
+  printf("%s %s", $cgroup_path, $cgroup_path); /* This may print two different paths */
+}
+----
+
 === cgroupid
 
 .variants
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -895,6 +895,32 @@ void CodegenLLVM::visit(Call &call)
     else
       createPrintNonMapCall(call, non_map_print_id_);
   }
+  else if (call.func == "cgroup_path")
+  {
+    auto elements = AsyncEvent::CgroupPath().asLLVMType(b_);
+    StructType *cgroup_path_struct = b_.GetStructType(call.func + "_t",
+                                                      elements,
+                                                      true);
+    AllocaInst *buf = b_.CreateAllocaBPF(cgroup_path_struct,
+                                         call.func + "_args");
+
+    // Store cgroup path event id
+    b_.CreateStore(b_.GetIntSameSize(cgroup_path_id_, elements.at(0)),
+                   b_.CreateGEP(cgroup_path_struct,
+                                buf,
+                                { b_.getInt64(0), b_.getInt32(0) }));
+    cgroup_path_id_++;
+
+    // Store cgroup id
+    auto arg = call.vargs->at(0);
+    auto scoped_del = accept(arg);
+    b_.CreateStore(expr_,
+                   b_.CreateGEP(cgroup_path_struct,
+                                buf,
+                                { b_.getInt64(0), b_.getInt32(1) }));
+
+    expr_ = buf;
+  }
   else if (call.func == "clear" || call.func == "zero")
   {
     auto elements = AsyncEvent::MapEvent().asLLVMType(b_);
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -228,6 +228,25 @@ bool wildcard_match(const std::string &str, std::vector<std::string> &tokens, bo
   return true;
 }
 
+/*
+ * Splits input string by '*' delimiter and return the individual parts.
+ * Sets start_wildcard and end_wildcard if input starts or ends with '*'.
+ */
+std::vector<std::string> get_wildcard_tokens(const std::string &input,
+                                             bool &start_wildcard,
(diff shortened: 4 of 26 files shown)


fbs:
nice thanks for adding this!

I don't think we should add new printing functions to the language, they're annoying to use. E.g. what if I want to print the cgroup path along side with the pid? I'd have to do someting like

```
printf("Cgroup path: "); cgroup_path(..); printf(" PID: %d\n", pid)
```

Thats quite horrible and `cgroup_path` probably spits out a new line as well.

`time()` and `join()` also suffer from this.

I'd suggest something like we did to fix `time`, a new type that can only be used for printing. In the kernel it's just an integer you but when sent to userspace it gets translated properly. This is also how `strftime` works. The time formatting happens in userspace but you treat it just like a string in bpftrace code:

```
BEGIN { $ts = strftime("%m/%d/%y", nsecs); printf("%s\n", $ts); exit(); }
```

lenticularis39:
I also thought about such solution, but I missed it already having been implemented for `time()`, I saw only the behavior of `join()` and thought that this problem should be fixed for all these functions in another PR. I'll look at how `strftime()` is implemented and try to use it to fix the cgroup path conversion.

lenticularis39:
> I'll look at how `strftime()` is implemented and try to use it to fix the cgroup path conversion.

Done.

fbs:
looks good, will do a proper review later. Can you rebase on master to fix ci?

lenticularis39:
Rebased.

viktormalik (commented):
Nice! LGTM, could you add some runtime test as well?

lenticularis39:
Added runtime tests.

man/adoc/bpftrace.adoc
```diff
@@ -1287,6 +1287,25 @@ t:syscalls:sys_enter_execve {
...
 ----
 
+=== cgroup_path
+
+.variants
+* `cgroup_path cgroup_path(int cgroupid)`
+
+Convert cgroup id to cgroup path.
+This is done asynchronously in userspace when the cgroup_path value is printed,
+therefore it can resolve to a different value if the cgroup id gets reassigned.
```
fbs, line 1297: Just curious, how likely is this?

lenticularis39: Very unlikely, the reassignment would have to happen exactly between the two perf event handler calls, perhaps not possible at all if cgroup ids don't get reused.

viktormalik:
> It looks like the way I implemented the runtime tests depends on the cgroup path reported by /proc/.../cgroup to be an existing file in /sys/fs/cgroup, which is not the case in Alpine Linux (or maybe is /sys/fs/cgroup not mounted there?). 

Why not just run `bpftrace -e 'BEGIN { printf("%s", cgroup_path(cgroup)); }'` and check that it starts with `/sys/fs/cgroup`? I think that it could be sufficient here.

> Also the test will fail if run in an environment with no cgroup at all, since such cgroup id won't be found in /sys/fs/cgroup.

Similarly to the `cgroup` builtin [runtime test](https://github.com/iovisor/bpftrace/blob/master/tests/runtime/builtin#L137-L141), you can limit the test to the minimal kernel version that supports cgroups.

lenticularis39:
> Why not just run `bpftrace -e 'BEGIN { printf("%s", cgroup_path(cgroup)); }'` and check that it starts with `/sys/fs/cgroup`? I think that it could be sufficient here.

The `cgroup` built-in returns an invalid value on my RHEL 8 VM for some reason:

```
$  sudo /opt/ktrace/bin/bpftrace -e 'BEGIN { print(cgroup); }'
Attaching 1 probe...
4294974185
^C
```

It might work on the CI, though, I'll try it.

lenticularis39:
> Similarly to the `cgroup` builtin [runtime test](https://github.com/iovisor/bpftrace/blob/master/tests/runtime/builtin#L137-L141), you can limit the test to the minimal kernel version that supports cgroups.

That makes sense, although I assume it would still not work on systems that don't use cgroups (i.e. all their processes are in the root cgroup).

lenticularis39:
> It might work on the CI, though, I'll try it.

It looks like it didn't work. (It does work on my openSUSE Leap 15.2 system, though.

lenticularis39:
I'm reworking the code to better support cgroupv1 and mixed hierarchy (currently it just picks the cgroup path it finds first and prints it relative to /sys/fs/cgroup, which is likely not the expected behavior, e.g. compared to the contents of /proc/.../cgroup), so I'm marking it as draft for now. Will revert it when I'm done.

fbs:
woops forgot about this one. The code looks good on first glance :)

Is there a way I can help with tests?

lenticularis39:
> woops forgot about this one. The code looks good on first glance :)

Thanks for looking at it. I updated it a bit, so now it should support all kinds of cgroup hierarchies (see the updated docs).

> Is there a way I can help with tests?

The issue appears to be related to cgroups on GitHub Actions, where bpftrace doesn't find the cgroup id in /sys/fs/cgroup, causing the runtime test for `cgroup_path` to fail. It doesn't happen when the tests are run locally in Docker.

A solution would be disabling the runtime test, but to do that properly it would be good to know the exact condition to check for (e.g. nothing is mounted at /sys/fs/cgroup) - I could try to run some checks inside the CI to see how it looks there. Do you have any other ideas what could be causing it?

fbs:
> 
> The issue appears to be related to cgroups on GitHub Actions, where bpftrace doesn't find the cgroup id in /sys/fs/cgroup, causing the runtime test for `cgroup_path` to fail. It doesn't happen when the tests are run locally in Docker.
> 
> A solution would be disabling the runtime test, but to do that properly it would be good to know the exact condition to check for (e.g. nothing is mounted at /sys/fs/cgroup) - I could try to run some checks inside the CI to see how it looks there. Do you have any other ideas what could be causing it?

Nope :(, also never looked at the insides of a actions VM.

lenticularis39:
I figured it out, it's caused by Docker on cgroupv1 Ubuntu 20.04: the css_set's default cgroup is set to host containerd service, which is not present in the container's cgroup hierarchy:
```
root@69f5958e04f6:/home/tglozar/bpftrace# bpftrace -e 'BEGIN { printf("%d", cgroup); }'
Attaching 1 probe...
^C2071

tglozar@ubuntu-test:~$ find /sys/fs/cgroup/ -inum 2071
/sys/fs/cgroup/memory/system.slice/dbus.socket/memory.kmem.tcp.limit_in_bytes
/sys/fs/cgroup/unified/system.slice/containerd.service

root@69f5958e04f6:/home/tglozar/bpftrace# ls /sys/fs/cgroup/unified/system.slice/containerd.service /sys/fs/cgroup/memory/system.slice/dbus.socket/memory.kmem.tcp.limit_in_bytes
ls: cannot access '/sys/fs/cgroup/unified/system.slice/containerd.service': No such file or directory
ls: cannot access '/sys/fs/cgroup/memory/system.slice/dbus.socket/memory.kmem.tcp.limit_in_bytes': No such file or directory
```

This is a design flaw of the cgroupv1 hierarchy for Docker containers. I added a note about this into the manual page and added a condition to the runtime test that disables it when cgroupv2 filesystem is not mounted.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -895,6 +895,28 @@ void CodegenLLVM::visit(Call &call)
...
+    auto elements = AsyncEvent::CgroupPath().asLLVMType(b_);
+    StructType *cgroup_path_struct = b_.GetStructType(call.func + "_t",
+                                                      elements,
+                                                      true);
+    AllocaInst *buf = b_.CreateAllocaBPF(cgroup_path_struct,
+                                         call.func + "_args");
+
+    // Store cgroup path event id
+    b_.CreateStore(b_.GetIntSameSize(cgroup_path_id_, elements.at(0)),
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(0) }));
```
viktormalik, line 909: Specify explicit type for `CreateGEP`, this version is deprecated on LLVM 13. See #2125.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -895,6 +895,28 @@ void CodegenLLVM::visit(Call &call)
...
+    cgroup_path_id_++;
+
+    // Store cgroup id
+    auto arg = call.vargs->at(0);
+    auto scoped_del = accept(arg);
+    b_.CreateStore(expr_,
+                   b_.CreateGEP(buf, { b_.getInt64(0), b_.getInt32(1) }));
```
viktormalik, line 916: Same as above.

src/utils.cpp
```diff
@@ -304,6 +304,121 @@ std::vector<std::string> get_kernel_cflags(
...
+  mounts_file.close();
+  return result;
+}
+
+std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
+    uint64_t cgroupid,
+    std::string filter)
+{
+  // TODO: Rewrite using std::views when C++20 support becomes common
+  auto roots = get_cgroup_hierarchy_roots();
```
viktormalik, line 369: What's the expected size of this vector? There's a lot of copying below so it could be more efficient to rewrite it by a loop.

docs/reference_guide.md
```diff
@@ -3092,6 +3093,25 @@ SRC 18:C0:4D:08:2E:BB, DST 74:83:C2:7F:8C:FF
...
+## 29. `cgroup_path`: Convert cgroup id to cgroup path
+
+Syntax: `cgroup_path(int cgroupid, string filter)`
+
+Converts the given cgroup id into the corresponding cgroup path for each cgroup hierarchy the id
+appears in. Because the conversion is done in user space, the resulting object can only be used for
+printing.
+
+Optionally a string literal may be passed as the second argument to filter cgroup hierarchies to look
+in (interpreted as a regex).
```
viktormalik, line 3105: I'm wondering how complex we expect the regex expressions to be. If it's mostly about using `.*`, then it would be better to allow just `*` as a placeholder for a sequence of any chars (instead of full regexes) as we do in other places (e.g. probe matcher). You could use the `wildcard_match` function then.

lenticularis39: I don't expect them to be used too much, most of the time the user will probably want to use either `.*` or an exact name of the hierarchy. `wildcard_match` supports only `*`, though, and not something like `(unified|cpu)`, unlike the regex.

viktormalik: I believe that we'll eventually get a support for that into our matcher so I'd still prefer going with `wildcard_match` for now. Users are also more used to specify path patterns using plain `*`, I'd say.

lenticularis39: OK, I'll move `get_tokens` to utils.cpp (currently it's private to probe_matcher.cpp) and swap the regex match with `wildcard_match` (which can be extended in another PR). I'd normally use glob(3) to match directly against the path, but cgroup2 can be mounted directly into `/sys/fs/cgroup`, which would break it...

src/utils.cpp
lenticularis39: On my system there are 18 cgroup hierarchy roots, which is not much, nevertheless I could improve the code.

fbs (approved):

viktormalik (approved):

PR #2056, opened 2021-11-02
Headers organization in AST

oazizi000:
Cleaning up `#includes` in the ast directory.

Heads up that some of these changes are stylistic, and I'm happy to change those parts if there are other preferences. Mostly, I was shooting for consistency.

With that said, this PR offers the following:
1) Make include paths relative to `src`. Since includes can be relative, some of this is arguably optional. However, there are also instances where it is required for correctness. For example, there were cases where files in `src/ast/passes` included files from `src/ast` without giving the correct path.

2) Include order consistency: This is purely stylistic, but I went ahead and tried to make the include orders a bit more consistent across files. Current order is: (1) header for the cpp file, if applicable, (2) system includes, (3) library includes (like llvm), (4) local includes.

I limited the changes to the `ast` subdirectory for now. I can expand out to other files in a future PR if there is any desire for that.

##### Checklist

### DIFF
--- a/src/ast/passes/field_analyser.h
+++ b/src/ast/passes/field_analyser.h
@@ -1,10 +1,11 @@
 #pragma once
 
-#include "bpftrace.h"
-#include "visitors.h"
 #include <iostream>
 #include <string>
 #include <unordered_set>
+
+#include "ast/visitors.h"
+#include "bpftrace.h"
 
 namespace bpftrace {
 namespace ast {
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -1,7 +1,9 @@
-#include "ast.h"
-#include "log.h"
-#include "visitors.h"
+#include "ast/ast.h"
+
 #include <iostream>
+
+#include "ast/visitors.h"
+#include "log.h"
 
 namespace bpftrace {
 namespace ast {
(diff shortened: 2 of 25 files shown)


src/ast/passes/field_analyser.h
```diff
@@ -1,11 +1,12 @@
 #pragma once
 
-#include "bpftrace.h"
-#include "visitors.h"
```
oazizi000, line 4: Just highlighting an example of where the include path was neither truly relative to the file itself nor to the project source root, in case it helps the review.

`field_analyser` is in `src/ast/passes`, but the included `visitors.h` is in `src/ast`

oazizi000:
> I suppose this paves the way towards running IWYU in the future?

Yup. That would be the ideal way to go.

My main motivation for now was just to fix the cases where includes were neither relative to the file nor relative to the ToT. And just couldn't help myself in fixing some style consistency. :)

danobi (approved):
This x100. Would be great if we could get this consistency for rest of code base too.

cc @fbs @viktormalik

fbs:
> This x100. Would be great if we could get this consistency for rest of code base too.
> 
> cc @fbs @viktormalik

yes! We def need some consistency (I know im horrible at it)

PR #2057, opened 2021-11-03
Use unique_ptr for Driver::root_

oazizi000:
Change to use `std::unique_ptr` instead of raw pointer for `Driver::root_`, for the sake of more modern memory management.

##### Checklist

### DIFF
--- a/src/driver.cpp
+++ b/src/driver.cpp
@@ -17,11 +17,6 @@ Driver::Driver(BPFtrace &bpftrace, std::ostream &o)
 {
 }
 
-Driver::~Driver()
-{
-  delete root_;
-}
-
 void Driver::source(std::string filename, std::string script)
 {
   Log::get().set_source(filename, script);
@@ -36,10 +31,8 @@ int Driver::parse_str(std::string script)
 
 int Driver::parse()
 {
-  // Ensure we free memory allocated the previous parse if we parse
-  // more than once
-  delete root_;
-  root_ = nullptr;
+  // Reset previous state if we parse more than once
+  root_.reset();
 
   // Reset source location info on every pass
   loc.initialize();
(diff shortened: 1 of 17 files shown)


src/main.cpp
```diff
@@ -397,16 +397,16 @@ static std::optional<struct timespec> get_boottime()
...
   err = driver.parse();
   if (err)
     return nullptr;
 
-  auto ast = driver.root_;
-  driver.root_ = nullptr;
-  return std::unique_ptr<ast::Node>(ast);
+  std::unique_ptr<ast::Node> retval = std::move(driver.root_);
+  driver.root_.reset();
+  return retval;
```
thefallentree, line 409: I feel this could be just `return (ast::Node *)driver.root_.release();`  ?

thefallentree (approved):

thefallentree (changes requested):
pls fix CI, looks like it needs a cast

src/main.cpp
oazizi000: Yup, this can be simplified. Changing it to `return std::move(driver.root_)`.

thefallentree (approved):

src/driver.cpp
```diff
@@ -19,7 +19,6 @@ Driver::Driver(BPFtrace &bpftrace, std::ostream &o)
 
 Driver::~Driver()
 {
-  delete root_;
 }
```
danobi, line 22: Just delete the destructor entirely?

fbs:
Looks good.

To be nitpicky, we should rename root_ to root. We try to use _ for private members

src/driver.cpp
oazizi000: Indeed...that's cleaner.

oazizi000:
> Looks good.
> 
> To be nitpicky, we should rename root_ to root. We try to use _ for private members



> Looks good.
> 
> To be nitpicky, we should rename root_ to root. We try to use _ for private members

Went ahead and changed `root_` to `root`.

PR #2058, opened 2021-11-05
Add a static build script

thefallentree:
This PR adds a static (embed) build script that can be used locally to generate static linked binaries through docker.

Also adds instructions in README.md.


danobi (approved):
Maybe as a followup, we should move all the `build*` scripts to `scripts/`? Not sure if anyone uses the docker builds for local development

PR #2059, opened 2021-11-06
src: fuzz_main.cpp: Fix includes

jsmolic:
This fixes the following build issues:
```
fuzz_main.cpp:20:10: fatal error: ast/clang_parser.h: No such file or
directory
fuzz_main.cpp:25:10: fatal error: codegen_llvm.h: No such file or
director
```
Bug: https://bugs.gentoo.org/801385

##### Checklist

### DIFF
--- a/src/fuzz_main.cpp
+++ b/src/fuzz_main.cpp
@@ -17,12 +17,12 @@
 #include <unistd.h>
 
 #include "ast/bpforc/bpforc.h"
-#include "ast/clang_parser.h"
 #include "ast/passes/callback_visitor.h"
+#include "ast/passes/codegen_llvm.h"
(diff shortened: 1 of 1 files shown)


jsmolic:
Compilation still fails with
```
src/fuzz_main.cpp:164:19: error: ‘class bpftrace::ast::SemanticAnalyser’ has no member named ‘create_maps’
  164 |   err = semantics.create_maps(true);
```

fbs:
thanks!

we should really add this to the CI so we know it builds, else it will keep on breaking.

Can you fix the clang-formt issue in fuzz_main.c? Not sure why the others are in there as well, might be fixed after a rebase.

jsmolic:
@fbs Thanks!

PR #2060, opened 2021-11-08
Mark CreateCounterPass() as inline

oazizi000:
`CreateCounterPass()` should be marked as inline since it is implemented in the header.

##### Checklist


danobi (approved):
Thanks!

For anyone else curious, the c++ guidelines say this too: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf2-a-h-file-must-not-contain-object-definitions-or-non-inline-function-definitions

PR #2062, opened 2021-11-11
poll_perf_buffers: make epollfd_ member variable

oazizi000:
Two items around `poll_perf_events`:
1) Make epollfd a member variable. It's state that is used in multiple functions, so seems like a good candidate for that.
2) Expose poll_perf_events interface polling period via `timeout_ms`

##### Checklist

### DIFF
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -214,12 +214,13 @@ private:
       int pid,
       bool file_activation);
   int setup_perf_events();
-  void poll_perf_events(int epollfd, bool drain = false);
+  void poll_perf_events(bool drain = false);
   int print_map_hist(IMap &map, uint32_t top, uint32_t div);
   int print_map_stats(IMap &map, uint32_t top, uint32_t div);
   static uint64_t read_address_from_output(std::string output);
   std::vector<uint8_t> find_empty_key(IMap &map, size_t size) const;
   bool has_iter_ = false;
+  int epollfd_ = -1;
 
   std::unordered_map<std::string, std::unique_ptr<Dwarf>> dwarves_;
 };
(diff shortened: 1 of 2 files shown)


src/bpftrace.h
```diff
@@ -214,12 +214,13 @@ class BPFtrace
       int pid,
       bool file_activation);
   int setup_perf_events();
-  void poll_perf_events(int epollfd, bool drain = false);
+  void poll_perf_events(bool drain = false, int timeout_ms = 100);
```
danobi, line 217: Any reason to expose `timeout_ms`? Doesn't look like there's any callers that set !100. Maybe make it a constant instead?

oazizi000: Made it a constant in this PR.

It could be useful to have it be a parameter that is controllable via arguments. I may put out a separate PR for exposing the timeout; we can discuss the merits of it at that time. For now, better to keep PRs focused :)

danobi: If this were a library, I'd agree. But for internal use only, we should only expose APIs that are actively used. Adding theoretically useful stuff leads to cruft (see YAGNI)

danobi (approved):

oazizi000:
I can't tell if the test failure is related to these changes. At first glance, it looks like it's not, but if it is related please let me know, and I'll investigate further.

fbs:
can you rebase?

oazizi000:
Rebased.

The failures appear unrelated (e.g. `fatal: The remote end hung up unexpectedly`).

fbs:
CI should work again, can you rebase your work?

Sorry for the CI struggle, seems like everything broke in a short window :/

viktormalik (approved):

PR #2066, opened 2021-11-15
parse_args: Move argument parsing into a function

oazizi000:
Refactored some code to consolidate argument parsing into a `parse_args` function. Basic idea was to keep all argument parsing (including processing of `optind`) into a single function that returns a struct of all the parsed args. 

Throwing this out there to see if the maintainers want it. If you like these sort of changes, there are few other code pieces in `main()` that can be moved into functions to help with organization. Let me know what you think...I'd be happy to put out a few other contributions. 

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -141,7 +141,7 @@ bool is_root()
     return true;
 }
 
-static int info()
+static void info()
 {
   struct utsname utsname;
   uname(&utsname);
@@ -156,8 +156,6 @@ static int info()
 
   std::cerr << std::endl;
   std::cerr << BPFfeature().report();
-
-  return 0;
 }
 
 static std::optional<struct timespec> get_boottime()
@@ -427,9 +425,8 @@ ast::PassManager CreateAotPM(std::string __attribute__((unused)))
   return pm;
 }
 
-int main(int argc, char* argv[])
+struct Args
 {
-  int err;
   std::string pid_str;
   std::string cmd_str;
   bool listing = false;
@@ -437,11 +434,23 @@ int main(int argc, char* argv[])
   bool usdt_file_activation = false;
   int helper_check_level = 0;
   TestMode test_mode = TestMode::UNSET;
-  std::string script, search, file_name, output_file, output_format, output_elf,
-      aot;
+  std::string script;
+  std::string search;
+  std::string filename;
+  std::string output_file;
+  std::string output_format;
+  std::string output_elf;
+  std::string aot;
   OutputBufferConfig obc = OutputBufferConfig::UNSET;
   BuildMode build_mode = BuildMode::DYNAMIC;
-  int c;
+  std::vector<std::string> include_dirs;
+  std::vector<std::string> include_files;
+  std::vector<std::string> params;
+};
+
+Args parse_args(int argc, char* argv[])
+{
+  Args args;
 
   const char* const short_options = "dbB:f:e:hlp:vqc:Vo:I:k";
   option long_options[] = {
@@ -458,8 +467,8 @@ int main(int argc, char* argv[])
     option{ "aot", required_argument, nullptr, 2004 },
     option{ nullptr, 0, nullptr, 0 }, // Must be last
   };
-  std::vector<std::string> include_dirs;
-  std::vector<std::string> include_files;
+
+  int c;
   while ((c = getopt_long(
               argc, argv, short_options, long_options, nullptr)) != -1)
   {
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -509,110 +521,177 @@ int main(int argc, char* argv[])
...
+    }
+
+    // Load positional parameters before driver runs so positional
+    // parameters used inside attach point definitions can be resolved.
+    while (optind < argc)
+    {
+      args.params.push_back(argv[optind]);
+      optind++;
+    }
+  }
```
oazizi000, line 654: This is mostly copied from various places within the original `main()`. There should be no new functionality here.

src/main.cpp
```diff
@@ -509,110 +521,177 @@ int main(int argc, char* argv[])
...
+
+  return args;
+}
+
+int main(int argc, char* argv[])
+{
+  int err;
+
+  const Args args = parse_args(argc, argv);
```
oazizi000, line 663: Main parse_args function, which returns all the parsed arguments into a const struct.

src/main.cpp
```diff
@@ -627,29 +706,22 @@ int main(int argc, char* argv[])
...
 
   if (!parse_env(bpftrace))
     return 1;
 
-  // Difficult to serialize flex generated types
-  if (helper_check_level && build_mode == BuildMode::AHEAD_OF_TIME)
-  {
-    LOG(ERROR) << "Cannot use -k[k] with --aot";
-    return 1;
-  }
```
oazizi000, line 641: Moved into parse_args.

src/main.cpp
```diff
@@ -673,33 +745,23 @@ int main(int argc, char* argv[])
...
 
-    if (optind == argc || std::string(argv[optind]) == "*")
-      script = "*:*";
-    else if (optind == argc - 1)
-      script = argv[optind];
-    else
-    {
-      usage();
-      return 1;
-    }
```
oazizi000, line 689: Moved into parse_args, as it deals with processing a positional parameter. See `args.search`

src/main.cpp
```diff
@@ -717,18 +779,11 @@ int main(int argc, char* argv[])
...
 
-  if (script.empty())
+  if (!args.filename.empty())
   {
-    // Script file
-    if (argv[optind] == nullptr)
-    {
-      LOG(ERROR) << "USAGE: filename or -e 'program' required.";
-      return 1;
-    }
```
oazizi000, line 727: More argument parsing that has been moved into `parse_args`, which either populates `args.script` (for `-e`) or `args.filename` (for a positional parameter).

src/main.cpp
```diff
@@ -752,26 +807,22 @@ int main(int argc, char* argv[])
...
 
-  // Load positional parameters before driver runs so positional
-  // parameters used inside attach point definitions can be resolved.
-  while (optind < argc)
+  for (const auto& param : args.params)
   {
-    bpftrace.add_param(argv[optind]);
-    optind++;
+    bpftrace.add_param(param);
   }
```
oazizi000, line 826: More positional parameter parsing moved into `parse_args`

oazizi000 (commented):
A few explanatory comments to help the review process.

src/main.cpp
```diff
@@ -744,34 +799,30 @@ int main(int argc, char* argv[])
...
+      std::ifstream file(args.filename);
       if (file.fail())
       {
-        LOG(ERROR) << "failed to open file '" << filename
+        LOG(ERROR) << "failed to open file '" << args.filename
                    << "': " << std::strerror(errno);
         return -1;
       }
 
+      filename = args.filename;
```
viktormalik, line 810: I believe that `filename` could be completely removed in favor of `args.filename`.

src/main.cpp
```diff
@@ -427,21 +425,32 @@ ast::PassManager CreateAotPM(std::string __attribute__((unused)))
...
   std::string cmd_str;
   bool listing = false;
   bool safe_mode = true;
   bool usdt_file_activation = false;
   int helper_check_level = 0;
   TestMode test_mode = TestMode::UNSET;
-  std::string script, search, file_name, output_file, output_format, output_elf,
-      aot;
+  std::string script;
+  std::string search;
```
viktormalik, line 438: If I'm not mistaken, `script` and `search` are mutually exclusive so we could merge them together (I'd use `script` here).

viktormalik (commented):
Personally, I like these kinds of refactorings, so I'm in favor of accepting it. Especially here as our `main` is huge.

Few more related refactorings are suggested in the comments below.

viktormalik:
A rebase could help with the CI

src/main.cpp
oazizi000: I debated this too. So far, I've taken the view that the parsed args should be constant:

```
const Args args = parse_args(argc, argv);
```

The rationale is to try to nudge manipulations of arguments into the `parse_args()` function, rather than doing them in the body. Of course, this turns out to be a counter-example because it ends up causing the creation of another variable.

Anyways, I'm not rigid on this. If the preference is still to switch this, I'm happy to do so.

src/main.cpp
oazizi000: They are indeed mutually exclusive, but since they are logically different, I figured it helps readability to keep separate variables.

Again, if the preference is still to merge them, let me know and I'll do it.

oazizi000:
@viktormalik Thanks for the feedback! Rebased and also left a few comments. Let me know what you think.

src/main.cpp
viktormalik: Me neither, I think that we can leave it is.

src/main.cpp
viktormalik: Ok, let's leave them separate for now, perhaps they'll get merged in future.

viktormalik (approved):
It'd be nice to have an opinion from one of the other maintainers, too. 

@fbs @danobi could you at least check if you find this refactoring meaningful?

fbs:
I like the idea, code looks good too.

Seems like this stupid build just keeps breaking on us :/ . Will try to fix it again once I find the time.
Looks like it might be time to get off ubuntu bionic (18) and go modern

viktormalik:
> Seems like this stupid build just keeps breaking on us :/ . Will try to fix it again once I find the time. Looks like it might be time to get off ubuntu bionic (18) and go modern

I already opened an issue for that (#2084)

viktormalik:
CI is fixed now, please rebase.

PR #2067, opened 2021-11-15
Fix minor sanitizer issue

oazizi000:
`AttachedProbe` has a reference to a `Probe` inside `RequiredResources`. Thus, strictly speaking, `RequiredResources resources` should be destroyed after `std::vector<std::unique_ptr<AttachedProbe>> attached_probes_`.

This means `std::vector<std::unique_ptr<AttachedProbe>> attached_probes_` should be declared after `RequiredResources resources;` in the struct, so that the AttachedProbes are destroyed first on `~BPFtrace`.

This was caught by an independent ASAN build. It doesn't affect much in practice, but changing it for strict correctness.

##### Checklist

### DIFF
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -143,7 +143,6 @@ public:
   Dwarf *get_dwarf(const std::string &filename);
   Dwarf *get_dwarf(const ast::AttachPoint &attachpoint);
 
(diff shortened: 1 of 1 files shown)


fbs:
LGTM but can you rebase onto master?

viktormalik (approved):

PR #2073, opened 2021-11-22
docker: Add xz-utils to the release docker image

aidanharris:
This doesn't bloat the image too much and allows bpftrace to extract the
kernel headers from /sys/kernel/kheaders.tar.xz when IKHEADERS is
enabled.

### DIFF
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -4,7 +4,7 @@ FROM ubuntu:$BASE
 ARG build_dir=build-embedded
 
 # Run security updates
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.release
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:$BASE
 ARG build_dir=build-embedded
 
 # Run security updates
```
thefallentree, line 6: maybe update the comments here

thefallentree (approved):

viktormalik (approved):

danobi (approved):

PR #2077, opened 2021-11-26
Fix runtime test color display

lenticularis39:
Don't display colors in runtime test output when stdout is not a tty.
This works because only the stdout output messages are colored.

Just a quick workaround of a probably forgotten todo I noticed. This doesn't handle the uncommon case of a terminal without color support, but it does handle the much more common case of stdout redirection.

##### Checklist

### DIFF
--- a/tests/runtime/engine/runner.py
+++ b/tests/runtime/engine/runner.py
@@ -2,6 +2,7 @@
 
 import subprocess
 import signal
+import sys
 import os
 import time
 from os import environ, uname, devnull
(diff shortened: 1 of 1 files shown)


tests/runtime/engine/runner.py
```diff
@@ -22,9 +23,8 @@
 ERROR_COLOR = '\033[91m'
 NO_COLOR = '\033[0m'
 
-# TODO(mmarchini) only add colors if terminal supports it
```
viktormalik, line 25: I'd leave the TODO here since this situation (terminal without color support) is still not covered.

lenticularis39: Done. (Also fixed typo that I forgot to commit and that caused the tests to fail.)

viktormalik (approved):

PR #2086, opened 2021-12-08
doc: added Arch linux install instructions

eze-kiel:
As an Arch Linux user, I would like to easily know how I can install `bpftrace` on my system.


dxuuu (approved):
Thanks!

danobi (approved):

PR #2087, opened 2021-12-09
Fix CI

danobi:
Move from bionic docker images to focal. More details in individual commits.

##### Checklist


.github/workflows/embedded.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           EMBED_USE_LLVM: ON
           RUN_ALL_TESTS: 1
           RUNTIME_TEST_DISABLE: builtin.cgroup,probe.kprobe_offset_fail_size
-          BASE: bionic
```
fbs, line 19: can we keep this on bionic for now? We're dynamic linking against libc so that will probably means that the binary from docker won't work on older systems anymore.

Also can't do this because the LLVM build is broken

danobi: Sounds good

danobi:
Whoops, didn't create the branch on my fork so now it's running double the CI jobs...

viktormalik (approved):

fbs:
Does this mean we're deprecating LLVM 6-8 support? Our track record of keeping untested things  working is not great.

fbs (approved):

danobi:
I'm in favor of dropping llvm 6-8. We ought to make the note at the same time we remove all #ifdef code, though

PR #2088, opened 2021-12-09
llvm-embedded: Add focal build

danobi:
When the CI moves away from bionic we'll need focal images to build
against.

##### Checklist

### DIFF
--- a/.github/workflows/embedded_llvm.yml
+++ b/.github/workflows/embedded_llvm.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         llvm_version: [12]
-        base: [xenial, bionic]
+        base: [xenial, bionic, focal]
     steps:
     - name: Checkout repo
       uses: actions/checkout@v2


viktormalik (commented):
LLVM builds seem to be failing, is this expected?

fbs:
sadly it has been broken for a whlie :(

The LLVM build actually pulls in LLVM & CLang from cmake and does the whole build magic stuff. For some reason  it cannot find LLVM just after building it

```
[100%] Built target embedded_llvm
[ 50%] Built target embedded_llvm
[ 57%] Creating directories for 'embedded_clang'
 CMake Error at CMakeLists.txt:66 (find_package):
  Could not find a package configuration file provided by "LLVM" with any of
  the following names:

    LLVMConfig.cmake
    llvm-config.cmake
```

It was (like many things :/ ) on my fix list but as we had an image that worked I found it low priority. Also because building LLVM is a huge pain and my hardware isn't that great. I ended up getting a big box at digital ocean for it, but thats adding $$ fast if I cant dedicate time.

Luckily we don't have to rush with this just yet. The embedded build builds BCC from source too (see docker/Dockerfile.ubuntu-glibc) so that will work (hopefully for a while). 
And our build is not fully static, libc is still dynamically linked so keeping the bionic build is beneficial for older systems

viktormalik:
Duplicate of #2372.

PR #2096, opened 2021-12-21
parser: fix operator precedence

fbs:
`1 + 2 *3` is currently evaluated as `(1+2)*3` while it should be `1 +
(2*3)`. This is because the additive and multiplicative operators we're
mixed in the same rule.

Tests are only implemented in runtime right now as doing this in the
parser is a bit tedious to write and it doesn't actually guaranteed the
correct output.

Fixes #2095

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to
 #### Deprecated
 #### Removed
 #### Fixed
+
+- Fix precedence of multiplicative operations
+  - [#2096](https//github.com/iovisor/bpftrace/pull/2096)
+
 #### Tools
 #### Documentation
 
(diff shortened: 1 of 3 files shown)


fbs:
<.< another build issue with libbpf. I guess we can get this in regardless, as the regular build passes

fbs:
#2097 is the backport for a v0.14.1 release

fbs:
imo we should also backport to 0.13.1, thoughts @danobi  @viktormalik?

danobi (approved):
Maybe add a test that uses all the operators? eg `1 + 2 - 3 << 4 >> 5` or something

fbs:
agreed lets add those later

Issue #2075, opened 2021-11-24
libc.so.6 shouldn't be fixed.


FingerLiu:
### What reproduces the bug?

`tools/gethostlatency.bt` does not work on my centOS machine.

After some research, found it was because libc.so.6 path was not the same with my machine.
```bash
# whereis libc.so.6
> libc.so: /usr/lib64/libc.so.6
```

Quick workaround is to change the path to /usr/lib64/libc.so.6 in my case.
Is there a method to dynamicly find the right location of libc.so on the running machine/

```bash
-uprobe:/lib/x86_64-linux-gnu/libc.so.6:getaddrinfo,
-uprobe:/lib/x86_64-linux-gnu/libc.so.6:gethostbyname,
-uprobe:/lib/x86_64-linux-gnu/libc.so.6:gethostbyname2
+uprobe:/usr/lib64/libc.so.6:getaddrinfo,
+uprobe:/usr/lib64/libc.so.6:gethostbyname,
+uprobe:/usr/lib64/libc.so.6:gethostbyname2
```

### bpftrace --info

System
  OS: Linux 5.15.4-1.el7.elrepo.x86_64 #1 SMP Sat Nov 20 08:10:25 EST 2021
  Arch: x86_64

Build
  version: v0.13.0
  LLVM: 12.0.0
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  bcc bpf_attach_uprobe refcount: yes
  libbpf: yes
  libbpf btf dump: yes
  libbpf btf dump type decl: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf (depends on Build:libbpf): no
  map batch (depends on Build:libbpf): yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
[...]

Issue #2065, opened 2021-11-14
"No probes to attach" with uretprobe in a shared library


phmarek:
### What reproduces the bug?

All these as root:

```
$ lsof -p 615444 | grep libssl
...
sbcl    615444 user  mem    REG               0,26          13945975 /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (path dev=0,27)

$ bpftrace -l 'uretprobe:/usr/lib/x86_64-linux-gnu/libssl.so.1.1:SSL_*' | wc -l
456
$ bpftrace -p 615444 -v -e 'uretprobe:/usr/lib/x86_64-linux-gnu/libssl.so.1.1:SSL_* { printf("%7d %-40s %12d 0x%8x\n", tid, func, retval, retval); }'
INFO: node count: 10
No probes to attach

$ bpftrace -l 'uretprobe:/usr/lib/x86_64-linux-gnu/libssl.so.1.1:SSL_write'
uretprobe:/usr/lib/x86_64-linux-gnu/libssl.so.1.1:SSL_write
$ bpftrace -p 615444 -v -e 'uretprobe:/usr/lib/x86_64-linux-gnu/libssl.so.1.1:SSL_write { printf("%7d %-40s %12d 0x%8x\n", tid, func, retval, retval); }'
INFO: node count: 10
No probes to attach
```

Seems that `uprobes` in a shared library aren't found for this process.

There _are_ probes, though -- but only ones in the binary:

```
$ bpftrace -p 615444 -l | cut -f 1-2 -d: | sort | uniq -c
    582 uprobe:/proc/615444/root/.../sbcl
```

### bpftrace --info

```
ERROR: bpftrace currently only supports running as the root user.
```

That might be another thing to fix... can't I get `--info` as a user??

```
System
  OS: Linux 5.14.0-2-amd64 #1 SMP Debian 5.14.9-2 (2021-10-03)
  Arch: x86_64

Build
  version: v0.14.0
  LLVM: 11.1.0
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: no
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  bcc bpf_attach_uprobe refcount: yes
  bcc library path resolution: yes
  libbpf: yes
  libbpf btf dump: yes
  libbpf btf dump type decl: yes
  libdw (DWARF support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
[...]

PR #2117, opened 2022-01-07
Add sslsnoop and ssllatency tools

xutao323:
Add sslsnoop and ssllatency tools for SSL/TLS handshake performance analysis.

Use case and background is described here: https://xutao323.github.io/2022/01/06/tls-acceleration-with-new-CPU.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to
 #### Added
 #### Docs
 #### Tools
+- Add sslsnoop and ssllatency tools
+  - [#2117](https://github.com/iovisor/bpftrace/pull/2117)
 
 ## [0.15.0] 2022-05-24
 
--- a/tools/ssllatency.bt
+++ b/tools/ssllatency.bt
@@ -1,0 +1,78 @@
+#!/usr/bin/bpftrace
+/*
+ * ssllatency	Trace SSL/TLS handshake for OpenSSL.
+ * 		For Linux, uses bpftrace and eBPF.
+ *
+ * ssllatency shows handshake latency stats and distribution.
+ *
+ * Copyright (c) 2021 Tao Xu.
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ *
+ * 17-Dec-2021	Tao Xu	created this.
+ */
+
+BEGIN
+{
+	printf("Tracing SSL/TLS handshake... Hit Ctrl-C to end.\n");
+}
+
+uprobe:libssl:SSL_read,
+uprobe:libssl:SSL_write,
+uprobe:libssl:SSL_do_handshake
+{
+	@start_ssl[tid] = nsecs;
+	@func_ssl[tid] = func; // store for uretprobe
+}
+
+uretprobe:libssl:SSL_read,
+uretprobe:libssl:SSL_write,
+uretprobe:libssl:SSL_do_handshake
+/@start_ssl[tid] != 0/
+{
+	$lat_us = (nsecs - @start_ssl[tid]) / 1000;
+	if ((int8)retval >= 1) {
+		@hist[@func_ssl[tid]] = lhist($lat_us, 0, 1000, 200);
+		@stat[@func_ssl[tid]] = stats($lat_us);
+	} else {
+		@histF[@func_ssl[tid]] = lhist($lat_us, 0, 1000, 200);
+		@statF[@func_ssl[tid]] = stats($lat_us);
+	}
+	delete(@start_ssl[tid]); delete(@func_ssl[tid]);
+}
+
+// need debug symbol for ossl local functions
+uprobe:libcrypto:rsa_ossl_public_encrypt,
+uprobe:libcrypto:rsa_ossl_public_decrypt,
+uprobe:libcrypto:rsa_ossl_private_encrypt,
+uprobe:libcrypto:rsa_ossl_private_decrypt,
+uprobe:libcrypto:RSA_sign,
+uprobe:libcrypto:RSA_verify,
+uprobe:libcrypto:ossl_ecdsa_sign,
+uprobe:libcrypto:ossl_ecdsa_verify,
+uprobe:libcrypto:ecdh_simple_compute_key
+{
+	@start_crypto[tid] = nsecs;
+	@func_crypto[tid] = func; // store for uretprobe
+}
+
+uretprobe:libcrypto:rsa_ossl_public_encrypt,
+uretprobe:libcrypto:rsa_ossl_public_decrypt,
+uretprobe:libcrypto:rsa_ossl_private_encrypt,
+uretprobe:libcrypto:rsa_ossl_private_decrypt,
+uretprobe:libcrypto:RSA_sign,
+uretprobe:libcrypto:RSA_verify,
+uretprobe:libcrypto:ossl_ecdsa_sign,
+uretprobe:libcrypto:ossl_ecdsa_verify,
+uretprobe:libcrypto:ecdh_simple_compute_key
+/@start_crypto[tid] != 0/
+{
+	$lat_us = (nsecs - @start_crypto[tid]) / 1000;
+	@hist[@func_crypto[tid]] = lhist($lat_us, 0, 1000, 200);
+	@stat[@func_crypto[tid]] = stats($lat_us);
+	delete(@start_crypto[tid]); delete(@func_crypto[tid]);
+}
+
+END
+{
+	printf("\nLatency distribution in microsecond:");
+}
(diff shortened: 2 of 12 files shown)


brendangregg:
Thanks; I think these are useful tools, but I don't think they'll work for everyone without edits to library paths. They just error for me as the paths can't be found. These would be the only tools in bpftrace that failed to run for most users, and would create a lot of tickets for people asking why. For that reason I wouldn't include them as is, as it will just give a lot of people a bad experience.

I'm not sure what to do, but I've been thinking along a few of different lines:

A) Include them in an examples directory instead and not the tools directory, and make it clear that this isn't expected to work without edits. bcc has an examples directory. I'm not sure this will be enough, people will still try to run them and complain.

B) Rewrite these in bcc libbpf-tools, where you can include a lot of other logic to search for the correct library paths and have the tool make a best effort to automatically do ssl tracing for you. 

C) Write this as a bash script that does the library and path checks, and finishes by invoking an embedded bpftrace program with the correct paths (or throws a better error message to say it can't find your ssl libraries). Programs that build programs are hard to debug and maintain, so my approach is to separate them as much as possible so it looks like this:

```
#!/bin/bash
...
bash logic
...
bpftrace -e '
...
bpftrace logic
...
'
```

So the tool source is in two halves, the top being bash and the bottom being bpftrace. A did this model years ago and maintaining it worked ok, for example: https://github.com/opendtrace/toolkit/blob/master/execsnoop

xutao323:
Thanks for the reiview and suggestions. I'll add patch checks for the new tools.

At first I found sslsniff in bcc-tools will trace system ssl library (libssl.so) and it can resolve the library path automatically, so I also added SSL_do_handshake() trace to show crypto overhead in sslsniff [PR 3799](https://github.com/iovisor/bcc/pull/3799). As bpftrace is simpler to write and modify, I added those extra crypto function traces in bpftrace here as sslsnoop.bt and ssllatency.bt.

I'll try the "C" approach to combine library patch check and bpftrace script.

fbs:
> Thanks; I think these are useful tools, but I don't think they'll work for everyone without edits to library paths. They just error for me as the paths can't be found. These would be the only tools in bpftrace that failed to run for most users, and would create a lot of tickets for people asking why. For that reason I wouldn't include them as is, as it will just give a lot of people a bad experience.

We don't have to specify the full path anymore so just doing 'u:libssl.so' should be enough :). We just haven't converted tools yet: #2075  it also has some caveats still: #2065

xutao323:
> > Thanks; I think these are useful tools, but I don't think they'll work for everyone without edits to library paths. They just error for me as the paths can't be found. These would be the only tools in bpftrace that failed to run for most users, and would create a lot of tickets for people asking why. For that reason I wouldn't include them as is, as it will just give a lot of people a bad experience.
> 
> We don't have to specify the full path anymore so just doing 'u:libssl.so' should be enough :). We just haven't converted tools yet: #2075 it also has some caveats still: #2065

Sounds great. Then system library path like libssl.so and libcrypto.so can be automatically resolved. And existing uprobe tracing tools with same problem can be converted too like gethostlatency and threadsnoop.

One question about non-default library or binary, if they don't exist yet on some system (some already have), like libcrypto_mb.so or qatengine.so in /lib64/engines-1.1/, can we just skip those uprobes and only trace remaining common funcitons in libssl.so or libcrypto.so ? Otherwise, I'll need to check path or remove non-default uprobes in the proposed new tools which is the main analysis target with TLS performance optimization.

fbs:
The docker container we build might not have it. Can you try adding `openssl` to the packages to install? You can add it to `docker/Dockerfile.focal` and `docker/Dockerfile.ubuntu-glibc`

xutao323:
Embedded build test failed even after installed `openssl-dev` on `alpine`. I tried the docker build locally, and found it's a common `ldconfig` problem of `alpine linux` (ubuntu is fine). So lib name can't be resolved properly by `ldconfig` in `alpine`, it's suggested to `apk add libc6-compat or gcompat`. Shall we fix it here or open a new issue for it?

Test steps:
1. Build the docker `alpine` image according to the [build log](https://github.com/iovisor/bpftrace/runs/5123757531?check_suite_focus=true#step:3:2):
```
# docker build -t bpftrace-embedded-alpine -f docker/Dockerfile.alpine --build-arg bcc_ref= --build-arg BASE=alpine --build-arg LLVM_VERSION=9 --build-arg ALPINE_VERSION=3.11 docker/
```
2. Manually run test and it failed with no libs found: 
```
# docker run --privileged -v $(pwd):$(pwd) -w $(pwd) --entrypoint /bin/bash bpftrace-embedded-alpine -c "/root/bpftrace/build-embedded/src/bpftrace /root/bpftrace/build-embedded/tools/ssllatency.bt"
/root/bpftrace/build-embedded/tools/ssllatency.bt:17-19: ERROR: uprobe target file 'libssl' does not exist or is not executable
...
/root/bpftrace/build-embedded/tools/ssllatency.bt:41-44: ERROR: uprobe target file 'libcrypto' does not exist or is not executable
...
```
3. Use absolute path is OK:
```
# docker run --privileged -v $(pwd):$(pwd) -w $(pwd) --entrypoint /bin/bash bpftrace-embedded-alpine -c "/root/bpftrace/build-embedded/src/bpftrace -e 'uprobe:/lib/libssl.so.1.1:SSL_do_handshake { printf(\"enter handshake\n\"); }'"
Attaching 1 probe...
```
4. Even the glibc example in [manual](https://github.com/iovisor/bpftrace/blob/master/man/adoc/bpftrace.adoc#probes-uprobe) failed due to same problem:
```
# docker run --privileged -v $(pwd):$(pwd) -w $(pwd) --entrypoint /bin/bash bpftrace-embedded-alpine -c "/root/bpftrace/build-embedded/src/bpftrace -e 'uprobe:libc:malloc { printf(\"Allocated %d bytes\n\", arg0); }'"
stdin:1:1-19: ERROR: uprobe target file 'libc' does not exist or is not executable
uprobe:libc:malloc { printf("Allocated %d bytes\n", arg0); }
~~~~~~~~~~~~~~~~~~
# docker run --privileged -v $(pwd):$(pwd) -w $(pwd) --entrypoint /bin/bash bpftrace-embedded-alpine -c "/root/bpftrace/build-embedded/src/bpftrace -e 'uprobe:/lib/libc.so.6:malloc { printf(\"Allocated %d bytes\n\", arg0); }'"
Attaching 1 probe...
Allocated 4 bytes
Allocated 8 bytes
Allocated 8 bytes
Allocated 48 bytes
```
Next:
Install the suggested `libc6-compat or gcompat` , so that ldconfig can work properly in `alpine`.

fbs:
for the alpine one we should probably just skip it, alpine is always a pain.  I think we already have the code to skip it but I'd have to check

xutao323:
> for the alpine one we should probably just skip it, alpine is always a pain. I think we already have the code to skip it but I'd have to check

Any update ? Thanks in advance.

xutao323:
#2156 is merged, so next step is to rerun the build tests ? if everything is ok, this new ssl/tls tool can be used like in bcc.

viktormalik:
> #2156 is merged, so next step is to rerun the build tests ? if everything is ok, this new ssl/tls tool can be used like in bcc.

You'll need to disable tests for the new tools on Alpine (similarly to #2181).

viktormalik:
Could you please rebase onto master and remove all the unnecessary merge commits? Thanks!

xutao323:
Rebased onto most recent master and cleaned up the commits, please help to review, thanks.

tools/ssllatency.bt
```diff
@@ -0,0 +1,78 @@
...
+		@hist[@func_ssl[tid]] = lhist($lat_us, 0, 1000, 200);
+		@stat[@func_ssl[tid]] = stats($lat_us);
+	} else {
+		@histF[@func_ssl[tid]] = lhist($lat_us, 0, 1000, 200);
+		@statF[@func_ssl[tid]] = stats($lat_us);
+	}
+	delete(@start_ssl[tid]); delete(@func_ssl[tid]);
+}
+
+// need debug symbol for ossl local functions
```
viktormalik, line 43: Just wondering - what does this comment mean? Do you need some info from DWARF?

viktormalik (approved):

tools/ssllatency.bt
xutao323: The uprobes with "rsa_ossl" prefix are static functions in libcrypto.so that needs DWARF debug info. So *openssl-libs-debuginfo* should be installed for those uprobes in CentOS/Fedora , and *libssl-dev-dbgsym* in Ubuntu.

The rest uprobes in the same *action block* are fine without DWARF. bpftrace allows wildcard to match multiple attach points, and I tested run with no error if no DWARF package installed. I think it shall be fine as far as there's at least one matching uprobe found.

xutao323:
any update? waiting for approval or any additional commennts need to be addressed? thanks.

viktormalik:
@brendangregg the tools should now work for most of the systems, do you agree with merging this?

fbs:
Sorry for the radio silence here. I think they're useful so lets just get them in. In the end they're only example tools we ship so we shouldn't make it too difficult, worst case we can delete them again.

Can you rabse to fix the few conflicts?

CHANGELOG.md
```diff
@@ -67,7 +67,10 @@ The 0.15.0 release has basic support for LLVM 14 but not all features work yet,
 #### Tools
 - Update biosnoop.bt for kernel >=5.17
   - [#2207](https://github.com/iovisor/bpftrace/pull/2207)
+- Add sslsnoop and ssllatency tools
+  - [#2117](https://github.com/iovisor/bpftrace/pull/2117)
```
fbs, line 71: argh sorry for the extra work but can you move this entry up? We did the 0.15 release and this needs to go into `unreleased`

xutao323: done

Issue #2135, opened 2022-01-31
Accessing the current probe name uprobes specified with an address causes some pathological behaviour


javierhonduco:
### What reproduces the bug?
Accessing the current probe name (`probe`) in uprobes specified with an address causes some pathological behaviour.

For example:
```dtrace
BEGIN {
    printf("start...\n");
}

uprobe:/tmp/test_binary_opt:0x24c91c0
{
    printf("probe=%s\n", probe);
    @[probe] = count();
}
```

Doesn't print anything after minutes of execution. A quick profile fetched with `perf -g` shows that we are spending a lot of time in `std::find` which looks like it could be an infinite loop, but haven't had time to dig deeper into this.

```
-   13.42%  bpftrace.real    bpftrace.real                                  [.] std::find<__gnu_cxx::__normal_iterator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*, std:◆
   - std::find<__gnu_cxx::__normal_iterator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::alloc▒
      - 13.42% bpftrace::ast::CodegenLLVM::visit                                                                                                                                                             ▒
           bpftrace::ast::CodegenLLVM::accept                                                                                                                                                                ▒
           bpftrace::ast::CodegenLLVM::createFormatStringCall                                                                                                                                                ▒
           bpftrace::ast::CodegenLLVM::visit                                                                                                                                                                 ▒
           bpftrace::ast::CodegenLLVM::accept                                                                                                                                                                ▒
           bpftrace::ast::CodegenLLVM::visit                                                                                                                                                                 ▒
           bpftrace::ast::CodegenLLVM::accept                                                                                                                                                                ▒
           bpftrace::ast::CodegenLLVM::generateProbe                                                                                                                                                         ▒
           bpftrace::ast::CodegenLLVM::visit                                                                                                                                                                 ▒
           bpftrace::ast::CodegenLLVM::accept                                                                                                                                                                ▒
           bpftrace::ast::CodegenLLVM::visit                                                                                                                                                                 ▒
           bpftrace::ast::CodegenLLVM::accept                                                                                                                                                                ▒
           bpftrace::ast::CodegenLLVM::generate_ir                                                                                                                                                           ▒
           main                                                                                                                                                                                              ▒
           __libc_start_main                                                                                                                                                                                 ▒
[...]

PR #2141, opened 2022-02-04
Limit number of generated BPF programs

viktormalik:
Originates from [this comment](https://github.com/iovisor/bpftrace/issues/2135#issuecomment-1027307705).

In some cases (e.g. when using the `probe` builtin), the number of generated BPF functions may be very large which may cause bpftrace to hang. This is a safeguard that sets a limit of 512 BPF functions by default. The limit can be changed using the
`BPFTRACE_MAX_BPF_PROGS` environment variable.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2281,6 +2281,18 @@ void CodegenLLVM::visit(Probe &probe)
         matches = bpftrace_.probe_matcher_->get_matches_for_ap(*attach_point);
       }
 
+      probe_count_ += matches.size();
+      if (probe_count_ > bpftrace_.max_programs_)
+      {
+        throw std::runtime_error(
+            "Your program is trying to generate more than " +
+            std::to_string(probe_count_) +
+            " BPF programs, which exceeds the current limit of " +
+            std::to_string(bpftrace_.max_programs_) +
+            ".\nYou can increase the limit through the BPFTRACE_MAX_BPF_PROGS "
+            "environment variable.");
+      }
+
       tracepoint_struct_ = "";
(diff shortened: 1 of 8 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -2281,6 +2281,19 @@ void CodegenLLVM::visit(Probe &probe)
         matches = bpftrace_.probe_matcher_->get_matches_for_ap(*attach_point);
       }
 
+      probe_count_ += matches.size();
+      if (probe_count_ > bpftrace_.max_programs_)
+      {
+        throw std::runtime_error(
```
fbs, line 2287: Using `LOG(FATAL)` here might be nicer for the end user

viktormalik: I wouldn't say, the exception is caught and translated to `LOG(ERROR)` while `LOG(FATAL)` will terminate the program with `SIGABRT` which is not very user-friendly.

fbs: ah we catch it, nice :)

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2281,6 +2281,19 @@ void CodegenLLVM::visit(Probe &probe)
...
+            "Your program is trying to generate more than " +
+            std::to_string(probe_count_) +
+            " BPF programs, which exceeds the current limit of " +
+            std::to_string(bpftrace_.max_programs_) +
+            " and it could take a long time.\n"
```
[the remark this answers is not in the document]
viktormalik: Changed "take a long time" to "take a lot of resources" in docs. Should be ready for another review.

PR #2146, opened 2022-02-21
[main] use enum for getopt return values

mmarchini:
##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -60,6 +60,21 @@ enum class BuildMode
   // Compile script into portable executable
   AHEAD_OF_TIME,
 };
+
+enum Options
+{
+  INFO = 2000,
+  EMIT_ELF,
+  NO_WARNING,
+  TEST,
+  AOT,
+  HELP,
+  VERSION,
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -556,18 +571,21 @@ Args parse_args(int argc, char* argv[])
...
         break;
       case 'b':
+      case Options::BTF:
         break;
       case 'h':
+      case Options::HELP:
         usage();
         exit(0);
       case 'V':
+      case Options::VERSION:
```
mmarchini, line 588: I wasn't sure about keeping these three unchanged vs using the enum. I went with using the enum for consistency, but happy to revert if folks think that makes things more complicated.

viktormalik (approved):
Could you format according to `clang-format` suggestions so that CI passes?

src/main.cpp
viktormalik: I prefer consistency so I'm fine with the current solution.

danobi (approved):

PR #2148, opened 2022-02-22
[cli] add --emit-llvm iption to write IR to file

mmarchini:
Adding a new CLI option to write LLVM IR to files so those can be opened
in editors with syntax highlight.

In theory the new option could be used on `compare_tool_codegen.sh` and `update_codegen_tests.sh`, not sure if that's worth it though.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -64,7 +64,6 @@ enum class BuildMode
 enum Options
 {
   INFO = 2000,
-  EMIT_ELF,
   NO_WARNING,
   TEST,
   AOT,
@@ -74,6 +73,8 @@ enum Options
   UNSAFE,
   BTF,
   INCLUDE,
+  EMIT_ELF,
+  EMIT_LLVM,
 };
 } // namespace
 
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -74,10 +73,12 @@ enum Options
   UNSAFE,
   BTF,
   INCLUDE,
+  EMIT_ELF,
+  EMIT_LLVM,
 };
 } // namespace
 
-void usage()
+void usage(bool extraHelp = false)
```
fbs, line 81: this isn't used right?

viktormalik (approved):

src/main.cpp
mmarchini: Good catch, I was going to make these options hidden by default and available with --full-help or something, but figured showing everything was probably a better user experience, especially since we don't have that many command line options.

fbs: yeah lets leave it ou

fbs (approved):
+1, would be nice if that help function argument can be removed

viktormalik:
@mmarchini could you please remove that unused argument so we can merge this? Thanks!

mmarchini:
done

viktormalik:
You'll also need to rebase to the latest master to make the CI pass.

fbs:
rebased

PR #2149, opened 2022-02-24
Improve resolution of Clang system include paths

### LINKED ISSUE #2150: Cannot find system headers on some systems
Debian-like systems (such as Ubuntu, which we use as a base for our Docker builds) moved system headers into arch-specific directories, e.g., `/usr/include/x86_64-linux-gnu`. On such systems, bpftrace is not able to find the headers and ClangParser fails to parse the script. 

This affects some of the shipped tools and makes CI fail.

### What reproduces the bug?

```
# bpftrace tools/tcp_accept.bt
```
 when BTF is available (as it tries to include `sys/socket.h`).

### Possible solutions

1. Use Clang to determine the list of system include paths. This can be done by running:
    ```
    $ clang -Wp,-v -x c -fsyntax-only /dev/null
    ```
    and parsing out the paths. While this is a good solution, it requires Clang to be installed, which is not a dependency of bpftrace and 
    is not available, e.g., when doing static builds.

2. It should be possible to obtain the list of include paths using Clang as a library, but we'd have to depend on `libclang-cpp` as the C API in `libclang` doesn't provide such functionality.

3. The simplest solution would be to hard-code `/usr/include/x86_64-linux-gnu` to the list of system include directories in ClangParser. This would only help on Ubuntu on x86, but at least it would make CI pass again. I implemented this one in #2149.

Do you think we should implement some more robust solution to this?

viktormalik:
Ubuntu and other Debian-like systems use arch-specific directories for their system include paths. This adds the default path (`/usr/include/<arch>-linux-gnu`) to Clang include directories so that ClangParser can locate system headers.

At the same time, this also adds automatic detection of system include paths (returned by `clang`). Resolves #2150.

##### Checklist

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -722,6 +722,10 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     "-isystem", "/usr/include",
   };
   // clang-format on
+  std::string arch_path = get_arch_include_path();
+  args.push_back("-isystem");
+  args.push_back(arch_path.c_str());
+
   for (auto &flag : extra_flags)
   {
     args.push_back(flag.c_str());
@@ -884,6 +888,13 @@ CXUnsavedFile ClangParser::get_empty_btf_generated_header()
     .Contents = btf_cdef.c_str(),
     .Length = btf_cdef.size(),
   };
+}
+
+std::string ClangParser::get_arch_include_path()
+{
+  struct utsname utsname;
+  uname(&utsname);
+  return "/usr/include/" + std::string(utsname.machine) + "-linux-gnu";
 }
 
 } // namespace bpftrace
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -715,13 +715,13 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
       .Length = input.size(),
   });
 
-  // clang-format off
-  args = {
-    "-isystem", "/usr/local/include",
-    "-isystem", "/bpftrace/include",
-    "-isystem", "/usr/include",
-  };
-  // clang-format on
+  args = { "-isystem", "/bpftrace/include" };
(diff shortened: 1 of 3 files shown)


src/clang_parser.cpp
```diff
@@ -720,6 +720,7 @@ bool ClangParser::parse(ast::Program *program, BPFtrace &bpftrace, std::vector<s
     "-isystem", "/usr/local/include",
     "-isystem", "/bpftrace/include",
     "-isystem", "/usr/include",
+    "-isystem", "/usr/include/x86_64-linux-gnu",
```
danobi, line 723: Would this be an issue on !x86 systems w/ x86 headers installed? Should we get host arch and craft an arch-specific path?

viktormalik: Right, yeah, that could be an issue. I'll try to craft the path from the host arch.

viktormalik:
I updated the PR based on the above comment.
In addition, I added automatic detection of system include paths, so this now resolves #2150.

src/clang_parser.cpp
```diff
@@ -886,4 +890,33 @@ CXUnsavedFile ClangParser::get_empty_btf_generated_header()
...
+  uname(&utsname);
+  LOG(INFO) << utsname.machine;
+  return "/usr/include/" + std::string(utsname.machine) + "-linux-gnu";
+}
+
+std::vector<std::string> ClangParser::system_include_paths()
+{
+  auto clang = "clang-" + std::to_string(LLVM_VERSION_MAJOR);
+  auto cmd = clang + " -Wp,-v -x c -fsyntax-only /dev/null 2>&1";
+  auto check = exec_system(cmd.c_str());
```
dxuuu, line 905: This will throw if `clang` does not exist, right? We need to catch `std::runtime_error` here and handle it

viktormalik: It won't, `exec_system` calls `popen` which doesn't fail on non-existing command. Anyway, it may fail in other situations, so I added exception handling.

danobi (approved):

PR #2153, opened 2022-03-02
Store lhist bounds inside map

### LINKED ISSUE #2121: lhist misprint labels based on step argument of last lhist call made
### What reproduces the bug?
Run a script to two lhist calls that use different step values. The printouts of both will use the step value of the last called one.

Eg, running 

    bpftrace -e 'kprobe:do_nanosleep {@[0] = lhist(pid, 0, 100000, 1000); @[1] = lhist(tid, 0, 100000, 10000); }'

returns

    @[0]: 
    [10000, 20000)      2106 |@@@@@@@@@                                           |
    [20000, 30000)         0 |                                                    |
    [30000, 40000)       613 |@@                                                  |
    [40000, 50000)         0 |                                                    |
    [50000, 60000)       768 |@@@                                                 |
    [60000, 70000)         0 |                                                    |
    [70000, 80000)      1453 |@@@@@@                                              |
    [80000, 90000)         9 |                                                    |
    [90000, 100000)        0 |                                                    |
    [100000, ...)        512 |@@                                                  |
    
    @[1]: 
    [0, 10000)          4934 |@@@@@@@@@@@@@@@@@@@                                 |
    [10000, 20000)     13156 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
    [20000, 30000)         0 |                                                    |
    [30000, 40000)        15 |                                                    |

Note that the data in @[0] is correct, but the labels are wrong. This cosmetic error can create confusion and even formatting errors, since the padding is computed for the correct labels (eg [1000,2000), [2000, 3000), ...)

### bpftrace --info
System
  OS: Linux 5.4.0-92-generic #103~18.04.2-Ubuntu SMP Wed Dec 1 16:50:36 UTC 2021
  Arch: x86_64

Build
  version: v0.14.1
  LLVM: 12.0.0
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  libbpf: yes
  libbpf btf dump: yes
  libdw (DWARF support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: no
  probe_read_user_str: no
  probe_read_kernel: no
  probe_read_kernel_str: no
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: no
  dpath: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  map batch (depends on Build:libbpf): no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
  iter:task: no
  iter:task_file: no

lenticularis39:
Move lhist bounds (min, max, step) from map resource metadata into the
map itself. Data is stored separately for each key prefix, enabling
storing multiple histograms in one map.

Fixes #2121

Makes use of negative indices with the map to avoid collision with histogram buckets (which are represented only by non-negative numbers).

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ and this project adheres to
   - [#2138](https://github.com/iovisor/bpftrace/pull/2138)
 - Fix tools to work on new kernel versions
   - [#2136](https://github.com/iovisor/bpftrace/pull/2136)
+- Fix misprinting of lhist labels
+  - [#2153](https://github.com/iovisor/bpftrace/pull/2153)
 
 #### Tools
 #### Documentation
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -457,6 +457,45 @@ void CodegenLLVM::visit(Call &call)
                                   { value, min, max, step },
                                   "linear");
 
+    // write lhist bounds into map
+    std::vector<Value *> lhist_bounds;
+    for (auto key_suffix :
+         std::vector<lhist_bound_index>{ lhist_bound_index_min,
+                                         lhist_bound_index_max,
+                                         lhist_bound_index_step })
+    {
+      AllocaInst *bound_key = getHistMapKey(
+          map,
+          Constant::getIntegerValue(b_.getInt64Ty(),
+                                    APInt(64, key_suffix, true)));
+      Value *val = nullptr;
+      std::string name;
+      switch (key_suffix)
+      {
+        case lhist_bound_index_min:
+          val = min;
+          name = "min";
+          break;
+        case lhist_bound_index_max:
+          val = max;
+          name = "max";
+          break;
+        case lhist_bound_index_step:
+          val = step;
+          name = "step";
+          break;
+      }
+      if (val)
+      {
+        AllocaInst *val_alloca = b_.CreateAllocaBPF(map.type,
+                                                    map.ident + "_" + name);
+        b_.CreateStore(val, val_alloca);
+        b_.CreateMapUpdateElem(ctx_, map, bound_key, val_alloca, call.loc);
+        b_.CreateLifetimeEnd(val_alloca);
+      }
+      b_.CreateLifetimeEnd(bound_key);
+    }
+
     AllocaInst *key = getHistMapKey(map, linear);
 
     Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -116,21 +116,6 @@ void ResourceAnalyser::visit(Call &call)
     resources_.join_args.push_back(delim);
     resources_.needs_join_map = true;
   }
-  else if (call.func == "lhist")
-  {
-    Expression &min_arg = *call.vargs->at(1);
-    Expression &max_arg = *call.vargs->at(2);
-    Expression &step_arg = *call.vargs->at(3);
-    Integer &min = static_cast<Integer &>(min_arg);
-    Integer &max = static_cast<Integer &>(max_arg);
-    Integer &step = static_cast<Integer &>(step_arg);
-
-    resources_.lhist_args[call.map->ident] = LinearHistogramArgs{
-      .min = min.n,
-      .max = max.n,
-      .step = step.n,
-    };
-  }
   else if (call.func == "time")
   {
     if (call.vargs && call.vargs->size() > 0)
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1552,12 +1552,12 @@ int BPFtrace::print_map_hist(IMap &map, uint32_t top, uint32_t div)
   }
   auto key(old_key);
 
-  std::map<std::vector<uint8_t>, std::vector<uint64_t>> values_by_key;
+  std::map<std::vector<uint8_t>, HistData> hists_by_key;
 
   while (bpf_get_next_key(map.mapfd_, old_key.data(), key.data()) == 0)
   {
     auto key_prefix = std::vector<uint8_t>(map.key_.size());
-    uint64_t bucket = read_data<uint64_t>(key.data() + map.key_.size());
+    int64_t bucket = read_data<int64_t>(key.data() + map.key_.size());
 
     for (size_t i=0; i<map.key_.size(); i++)
       key_prefix.at(i) = key.at(i);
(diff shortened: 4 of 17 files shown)


fbs:
Thanks for working on this!

If I understand the code correctly this basically adds a few extra (hidden) keys to the map that store the bounds right? And this basically happens every time?

Looking at the bpf for a simple `@ = lhist()` program:

```
2027: perf_event  name interval__1  tag 3fa72bdc138bbef1  gpl
	loaded_at 2022-03-12T13:06:51+0000  uid 0
	xlated 464B  jited 256B  memlock 4096B  map_ids 2223
2028: perf_event  name interval__1  tag d36e95831c5b0e61  gpl
	loaded_at 2022-03-12T13:07:14+0000  uid 0
	xlated 176B  jited 106B  memlock 4096B  map_ids 2226
```

We're basically going from 21 to 57 instructions.

While creating the 'optimal' bpf is not our goal I do think we should keep an eye on what we generate.

If instead of storing 3 values we store one, a unique ID that identifies the bounds used, we can reduce 3 stores to 1. As an optimization on top of that you can consider only doing that when its actually needed.

But more importantly, I'm not convinced this needs fixing. What is the use case for using different `lhist()`s in an associative array over just using multiple maps?

With this change in you can do `@=lhist(A); @=lhist(B)`, mixing two different bounds in the same element producing a totally garbage result. Detecting and blocking a situation like that is not easy as dynamic map keys are valid, e.g.

```
i:s:1 { @[nsecs/1e6] = lhist(123, 0, 1e3, 50); }
i:s:1 { @[nsecs/1e6] = lhist(123, 0, 1e6, 5*1e3); }
```

This will quite likely produce garbage but detecting the situation is probably not possible at compile time. So we'd have to add runtime checking for bounds but that would blow up code size (similar to the `-k` flag).

My preference would be to make `@ = lhist(A); @ = lhist(b)` a error. If there is a good use case I think doing something like `@[1] = lhist(a); @[2] = lhist(b)` is acceptable too as long as we can prove the map key is unique in all cases.

lenticularis39:
> If I understand the code correctly this basically adds a few extra (hidden) keys to the map that store the bounds right? And this basically happens every time?
> 

Yes, these keys are added every time the histogram is updated. It could of course check if the keys are already there, but there would be little benefit from doing this, since it would still have to access the map, and the check would add even more instructions.

> We're basically going from 21 to 57 instructions.
> 
> While creating the 'optimal' bpf is not our goal I do think we should keep an eye on what we generate.
> 
> If instead of storing 3 values we store one, a unique ID that identifies the bounds used, we can reduce 3 stores to 1. As an optimization on top of that you can consider only doing that when its actually needed.
> 

The optimal solution would of course be storing all constant data used only for output in userspace, including the linear histogram bounds in case they are constant. This would have to be done by introducing an abstraction for such values in the code, and is out of the scope of this patch.

I considered keeping the original code storing the bounds in userspace along with the new code for cases when both the bounds and the index are constant, but after seeing how it was implemented I decided to remove it and possibly reimplement it in a way that won't require having the bounds variables in every map. I can add this to the PR if you prefer to have it together to prevent the possible breakage of exisiting programs using lhist.

> But more importantly, I'm not convinced this needs fixing. What is the use case for using different `lhist()`s in an associative array over just using multiple maps?
> 

One might want to create histograms with an dynamic index, e.g. one for each PID:

```
# bpftrace -e 'kfunc:vfs_read { @[pid] = lhist(args->count, 0, 100000, 10000); }'
Attaching 1 probe...
^C

@[74200]: 
[10000, 20000)         2 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|

@[74480]: 
[0, 10000)             2 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|

@[1040]: 
[0, 10000)             3 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
```

One might also want to dynamically set the bounds, e.g. we might know that a specific group of PIDs do an operation more frequently than others. This could be done either by calling `lhist` once for each set of bounds or by setting the bounds dynamically (the latter would save some instructions, so it would be worth it to also allow non-constant bounds).

> With this change in you can do `@=lhist(A); @=lhist(B)`, mixing two different bounds in the same element producing a totally garbage result. Detecting and blocking a situation like that is not easy as dynamic map keys are valid, e.g.
> 
> ```
> i:s:1 { @[nsecs/1e6] = lhist(123, 0, 1e3, 50); }
> i:s:1 { @[nsecs/1e6] = lhist(123, 0, 1e6, 5*1e3); }
> ```
> 
> This will quite likely produce garbage but detecting the situation is probably not possible at compile time. So we'd have to add runtime checking for bounds but that would blow up code size (similar to the `-k` flag).
> 

This is also the case without the change (unless I'm mistaken), only the resulting garbage is slightly different, since only one set of bounds is used, so the data makes slightly more sense, but still is incorrect. I understand that allowing creating lhists at dynamic indices makes this hard to detect, however IMO avoiding this should be the responsibility of the programmer, the code generating garbage data being essentially equal to this in pseudocode with more standard syntax:

```
set_lhist_bounds(@[idx], 0, 10000, 1000);
add_value_to_lhist(@[idx], value);
set_lhist_bounds(@[idx], 100, 1000, 50); // This clearly invalidates the contents of the lhist
add_value_to_lhist(@[idx], another_value);
```

> My preference would be to make `@ = lhist(A); @ = lhist(b)` a error. If there is a good use case I think doing something like `@[1] = lhist(a); @[2] = lhist(b)` is acceptable too as long as we can prove the map key is unique in all cases.

Alternatively we could just check for that when possible, and leave the rest to the programmer, mentioning it in the manual. Also a future idea could be improving the non-standard hist/lhist syntax (especially the fact that the equals sign doesn't represent what it usually does in most programming languages), which seems to me like the main source of this problem.

lenticularis39:
Closing the PR for now in favor of a simpler fix after a discussion with @viktormalik.

I have some WIP code to re-introduce saving the bounds into resources if a use case for having multiple histograms with different bounds in a single map (like the one below) appears, so it can be revisited later when that happens.

> One might also want to dynamically set the bounds, e.g. we might know that a specific group of PIDs do an operation more frequently than others. This could be done either by calling lhist once for each set of bounds or by setting the bounds dynamically (the latter would save some instructions, so it would be worth it to also allow non-constant bounds).

PR #2154, opened 2022-03-02
Fix ScopedExprDeleter

michalgr:
When move assigning/constructing std::function the function moved from
is left in a valid but unspecified state. Care needs to be taken to make
sure ScopedExprDeleter accounts for that.

This commit fixes lifetime issues I observed in bpftrace compiled with
Android's r23 ndk.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.h
+++ b/src/ast/passes/codegen_llvm.h
@@ -89,8 +89,17 @@ private:
 
     ScopedExprDeleter(const ScopedExprDeleter &other) = delete;
     ScopedExprDeleter &operator=(const ScopedExprDeleter &other) = delete;
-    ScopedExprDeleter(ScopedExprDeleter &&other) = default;
-    ScopedExprDeleter &operator=(ScopedExprDeleter &&other) = default;
+
+    ScopedExprDeleter(ScopedExprDeleter &&other)
+    {
+      *this = std::move(other);
+    }
+
+    ScopedExprDeleter &operator=(ScopedExprDeleter &&other)
+    {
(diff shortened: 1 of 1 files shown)


src/ast/passes/codegen_llvm.h
```diff
@@ -89,8 +89,17 @@ class CodegenLLVM : public Visitor
 
     ScopedExprDeleter(const ScopedExprDeleter &other) = delete;
     ScopedExprDeleter &operator=(const ScopedExprDeleter &other) = delete;
-    ScopedExprDeleter(ScopedExprDeleter &&other) = default;
-    ScopedExprDeleter &operator=(ScopedExprDeleter &&other) = default;
+
+    ScopedExprDeleter(ScopedExprDeleter &&other)
+    {
+      *this = std::move(other);
```
Birch-san, line 95: is this just a shorthand that says "[move constructor](https://en.cppreference.com/w/cpp/language/move_constructor) delegates to [move assignment operator](https://en.cppreference.com/w/cpp/language/move_assignment)"?

michalgr: Yes. When move-constructing ScopedExprDeleter the compiler will first default initialize deleter_ to be an empty function and then the constructor will delegate to operator= to handle disarming of other.

michalgr:
I see that the failed check [keeps on failing on other commits as well](https://github.com/iovisor/bpftrace/runs/5444505840?check_suite_focus=true), with the same missing header error message. I'll assume it's not related to this pull request unless you tell me otherwise.

PR #2160, opened 2022-03-07
Add support for llvm 15

chenhengqi:
##### Checklist


src/ast/irbuilderbpf.cpp
```diff
@@ -282,7 +282,11 @@ CallInst *IRBuilderBPF::createCall(Value *callee,
 {
 #if LLVM_VERSION_MAJOR >= 11
   auto *calleePtrType = cast<PointerType>(callee->getType());
+#if LLVM_VERSION_MAJOR >= 15
+  auto *calleeType = cast<FunctionType>(calleePtrType->getPointerElementType());
```
viktormalik, line 286: `Type::getPointerElementType` has been in LLVM for a long time, so we should be able to use it for all versions. In fact, here you don't even need to do the above cast to `PointerType`.

This also applies to the rest of changes in this PR.

src/ast/irbuilderbpf.cpp
```diff
@@ -382,8 +386,13 @@ Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
     CREATE_MEMCPY(value, call, type.GetSize(), 1);
   else
   {
+#if LLVM_VERSION_MAJOR >= 15
+    assert(value->getType()->isPointerTy() &&
+           (value->getType()->getPointerElementType() == getInt64Ty()));
```
viktormalik, line 391: Same as above: use `getPointerElementType` for all versions.

viktormalik (changes requested):

viktormalik:
`clang-format` reports some issues, please fix them.

chenhengqi:
Done.

viktormalik (approved):

PR #2164, opened 2022-03-10
avoid regex 'parsing' format strings every time

fbs:
bpftrace::format spends about 60% of its time in the regex_iterator we
use to split the string.
The FormatString type avoids this by splitting the string on first use
and storing the result for re-use.

I've kept the code mostly as is, just moving and minor modifications.
The whole output code can do with a refactoring but this should be a
nice performance gain.

Test case is running opensnoop.bt while running `find /` in another
terminal:

original: bpftrace::perf_event_printer spent 75% of the time in format
new: bpftrace::perf_event_printer spent 15% of the time in format

##### Checklist


fbs:
before:


profile-pre

after:

profile-post

src/format_string.h
```diff
@@ -0,0 +1,79 @@
...
+{
+private:
+  /**
+   * Split the format string on format specifiers, e.g.
+   * 'foo %s bar' -> [ 'foo %s', 'bar' ]
+   */
+  void split();
+
+public:
+  FormatString() = default;
```
fbs, line 31: shouldn't need this but got some confusing template errors. Will try to remove it

fbs: Don't see a way to avoid this, unless we change the `vector<tuple<FormatString, vector<Fields>>` into something with a custom serializer that does it correctly. Unless I'm doing something wrong.

Having this also keeps the serializer for this class easy.

src/printf.cpp
```diff
@@ -6,70 +6,6 @@
 
 namespace bpftrace {
 
-std::string verify_format_string(const std::string &fmt, std::vector<Field> args)
```
viktormalik, line 9: Should be also removed from `src/printf.h`.

viktormalik (commented):
Just one comment, LGTM otherwise.

viktormalik (approved):

PR #2170, opened 2022-03-17
Add BPFTRACE_KERNEL_BUILD env var for out-of-tree kernel build

ThomsonTan:
The existing BPFTRACE_KERNEL_SOURCE supports passing kernel source tree
from environment variable to bpftrace, but it assumes in-tree build as
it returns the previous env var value as both ksrc and kobj, which
causes compilation error for out-of-tree Linux kernel build when loading
BTF.

This PR adds optional BPFTRACE_KERNEL_BUILD to pass out-of-tree build
dir, and fallback to BPFTRACE_KERNEL_SOURCE if it is not provided which
makes it still work for in-tree build.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -3631,4 +3631,5 @@ bpftrace requires kernel headers for certain features, which are searched for by
 /lib/modules/$(uname -r)
 ```
 
-The default search directory can be overridden using the environment variable `BPFTRACE_KERNEL_SOURCE`.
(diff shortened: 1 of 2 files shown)


viktormalik:
Could you document this? At least in reference guide (where `BPFTRACE_KERNEL_SOURCE` is mentioned), but ideally also in manpage and help.

viktormalik (approved):

viktormalik:
> Should I look into below test failure for the latest commit? The first commit passed all tests, and the appended one is only for formatting though.

No need, a restart helped. Please squash the commits and we can merge this.

PR #2172, opened 2022-03-20
Fix error on accessing field of struct request when without BTF

ideal:
Without including blk-mq.h, we will get: ERROR: Struct/union
of type 'struct request' does not contain a field named 'rq_disk'

##### Checklist

### DIFF
--- a/tools/biosnoop.bt
+++ b/tools/biosnoop.bt
@@ -12,6 +12,7 @@
 
 #ifndef BPFTRACE_HAVE_BTF
 #include <linux/blkdev.h>
+#include <linux/blk-mq.h>
 #endif
 
 BEGIN


tools/biosnoop.bt
```diff
@@ -12,6 +12,7 @@
 
 #ifndef BPFTRACE_HAVE_BTF
 #include <linux/blkdev.h>
+#include <linux/blk-mq.h>
```
viktormalik, line 15: This is a potential problem for systems which don't have `linux/blk-mq.h` but it's been added in kernel 3.13 which is over 8 years old, so I guess that we're fine to add it.

viktormalik (approved):

PR #2176, opened 2022-03-27
Correct handling of BPFTRACE_MAX_TYPE_RES_ITERATIONS env var

gedimitr:
When the env var BPFTRACE_MAX_TYPE_RES_ITERATIONS was malformed,
the env parsing function would erroneously return a 'success'
return code

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -293,10 +293,6 @@ static std::optional<struct timespec> get_boottime()
 
   if (!get_uint64_env_var("BPFTRACE_MAX_TYPE_RES_ITERATIONS",
                           bpftrace.max_type_res_iterations))
-    return 1;
-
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -291,10 +291,6 @@ static std::optional<struct timespec> get_boottime()
   if (!get_uint64_env_var("BPFTRACE_PERF_RB_PAGES", bpftrace.perf_rb_pages_))
     return false;
 
-  if (!get_uint64_env_var("BPFTRACE_MAX_TYPE_RES_ITERATIONS",
-                          bpftrace.max_type_res_iterations))
-    return 1;
```
viktormalik, line 296: This should just change the return to `return false`, not remove the entire block, shouldn't it?

viktormalik: Ah, didn't see that, right. Thanks for noticing and fixing this!

viktormalik (approved):

PR #2178, opened 2022-03-31
Use std::string to manipulate bpf program name

michalgr:
name buffer is not zero initialized and strncpy does not append
terminating zero character when there is no place for it. As a result,
on Android which checks for correctness of strrchr calls I see the
following crash:

0000000000026c28  /system/lib64/libc.so (syscall+24)
0000000000029805  /system/lib64/libc.so (abort+101)
0000000000050ed9  /system/lib64/libc.so (__fortify_fatal(char const*, ...)+153)
0000000000051815  /system/lib64/libc.so (__strrchr_chk+53)
00000000003dcbac  /data/local/tmp/bpftools/bin/bpftrace
00000000003dc2cb  /data/local/tmp/bpftools/bin/bpftrace
00000000003f64bc  /data/local/tmp/bpftools/bin/bpftrace
00000000003fcdca  /data/local/tmp/bpftools/bin/bpftrace
00000000003aa431  /data/local/tmp/bpftools/bin/bpftrace
00000000000c288c  /system/lib64/libc.so (__libc_init+92)

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -676,9 +676,8 @@ void AttachedProbe::load_prog()
 
   uint64_t log_buf_size = probe_.log_size;
   auto log_buf = std::make_unique<char[]>(log_buf_size);
-  char name[STRING_SIZE];
-  const char *namep;
-  std::string tracing_type, tracing_name;
+  std::string name;
+  std::string tracing_type;
 
   {
     // Redirect stderr, so we don't get error messages from BCC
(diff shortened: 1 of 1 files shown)


src/attached_probe.cpp
```diff
@@ -694,16 +693,13 @@ void AttachedProbe::load_prog()
 
     // Use orig_name for program name so we get proper name for
     // wildcard probes, replace wildcards with '.'
```
viktormalik, line 695: Can you adjust this comment accordingly to the change? There's no more `orig_name` identifier.

michalgr: Doesn't orig_name refer to probe_.orig_name in this context ?

viktormalik: Ah, that's probably correct, sorry about the confusion.

viktormalik (approved):

PR #2185, opened 2022-04-07
tests/runtime: Restrict kfunc attach test to x86.

sumanthkorikkar:
The kernel function arch_prepare_bpf_trampoline() is implemented only for x86.

kfunc attach tests would fail irrespective of whether has_prog_kfunc()
is true or false on non-x86 platforms


viktormalik:
We should resolve this one in a better way. If `has_prog_kfunc` is not sufficient to determine if kernel has BPF trampolines, then we should improve it instead of limiting all `kfunc` tests to x86 only.

viktormalik:
We should resolve this one in a better way. If `has_prog_kfunc` is not sufficient to determine if kernel has BPF trampolines, then we should improve it instead of limiting all `kfunc` tests to x86 only.

sumanthkorikkar:
Ok. agree. Will work on it. Thanks

PR #2187, opened 2022-04-08
Add test max kernel option

sumanthkorikkar:
Add MAX_KERNEL option to ensure that testcases are still executed for older kernels, but skipped for newer kernels.

Usecase:
urandom_read tracepoint is no longer present since kernel v5.17-rc5. Ensure testing is performed until kernel 5.17.

### DIFF
--- a/tests/runtime/variable
+++ b/tests/runtime/variable
@@ -52,6 +52,12 @@ TIMEOUT 5
 MAX_KERNEL 5.16
 AFTER dd if=/dev/urandom bs=3 count=1
 
+NAME 32-bit tracepoint arg openat_flags
+PROG tracepoint:syscalls:sys_enter_openat /comm == "syscall"/ { $i = args->flags; printf("openflags: %d\n", $i); if ($i == 64) { exit() } }
+EXPECT openflags: 64
+TIMEOUT 5
+AFTER ./testprogs/syscall openat
+
 NAME tracepoint arg casts in predicates
 RUN {{BPFTRACE}} -e 'tracepoint:syscalls:sys_enter_wait4 /args->ru/ { @ru[tid] = args->ru; } tracepoint:syscalls:sys_exit_wait4 /@ru[tid]/ { @++; exit(); }' -c ./testprogs/wait4_ru
 EXPECT @: 1
--- a/tests/runtime/variable
+++ b/tests/runtime/variable
@@ -49,6 +49,7 @@ NAME 32-bit tracepoint arg
 PROG tracepoint:random:urandom_read { $i = args->got_bits; printf("bits: %d\n", $i); if ($i == 24) { exit() } }
 EXPECT bits: 24
 TIMEOUT 5
+MAX_KERNEL 5.16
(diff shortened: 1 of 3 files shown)


tests/runtime/variable
```diff
@@ -49,6 +49,7 @@ NAME 32-bit tracepoint arg
 PROG tracepoint:random:urandom_read { $i = args->got_bits; printf("bits: %d\n", $i); if ($i == 24) { exit() } }
 EXPECT bits: 24
 TIMEOUT 5
+MAX_KERNEL 5.17
```
viktormalik, line 52: If `tracepoint:random:urandom_read` is not available since 5.17, then this should be set to 5.16, shouldn't it?

sumanthkorikkar: If the current kernel is 5.17.0-rc5, Then

 LooseVersion("5.17.0-rc5") > LooseVersion("5.17") returns True. 

This means kernel 5.17.0-rc4 wouldnt be ignored. However, kernel 5.17.0-rc6 would fail.

sumanthkorikkar:
The urandom_read tests the 32-bit tp args usecase. Let me check, if there are any other similar 32 bit args tracepoint which is present in older kernel as well.

tests/runtime/variable
sumanthkorikkar: Added the suggestions

viktormalik:
> Also added additional tests to ensure that tracepoint 32 bit tests runs for latest kernel

Unfortunately, the new test failed for embedded build. Could you please check and fix it?

sumanthkorikkar:
Fixed. Thanks

tests/runtime/variable
```diff
@@ -49,8 +49,15 @@ NAME 32-bit tracepoint arg
...
 EXPECT bits: 24
 TIMEOUT 5
+MAX_KERNEL 5.16
 AFTER dd if=/dev/urandom bs=3 count=1
 
+NAME 32-bit tracepoint arg mode
+PROG tracepoint:syscalls:sys_enter_access /comm == "tp_access"/ { if (args->mode == 2) { printf("mode: %d\n", args->mode);  exit(); } }
+EXPECT mode: 2
+TIMEOUT 5
+AFTER ./testprogs/tp_access
```
viktormalik, line 59: If you used a different syscall than `access` (e.g. `openat`), you could use `AFTER ./testprogs/syscall openat [<ARGS>]` and wouldn't need to write another testprog.

viktormalik, line 59: If you used a different syscall than `access` (e.g. `openat`), you could use `AFTER ./testprogs/syscall openat [<ARGS>]` and wouldn't need to write another testprog.

viktormalik, line 59: If you used a different syscall than `access` (e.g. `openat`), you could use `AFTER ./testprogs/syscall openat [<ARGS>]` and wouldn't need to write another testprog.

sumanthkorikkar:
added

viktormalik (approved):

PR #2194, opened 2022-04-17
Add support for multi kprobe interface

olsajiri:
Adding support for multi kprobe bpf link, that allows
to attach kprobe program to multiple functions in very
fast way.

```
  # perf stat -r 5 -e cycles ./src/bpftrace -e 'kprobe:x* {  } i:ms:1 { exit(); }'
  ...
  0.5661 +- 0.0275 seconds time elapsed  ( +-  4.85% )

```
The kprobe multi link was recently added to the kernel:
  https://lore.kernel.org/bpf/20220316122419.933957-1-jolsa@kernel.org/

When the feature is detected it is used instead of current
wildcard support, so there's no syntax change and kprobes
with wildcards like 'kprobe:ksys_*' work as before.

Apart from attachment and detachment speed another notable
difference is that with kprobe multi link support we load
just single program for all the attached functions.

Using '-vv' option to display functions we attach to:

```
  # bpftrace -vv  -e 'kprobe:ksys_* { @[kstack] = count(); } '
  INFO: node count: 7
  Attaching 1 probe...

  Program ID: 1940

  The verifier log:
  processed 31 insns (limit 1000000) max_states_per_insn 0 total_states 2 peak_states 2 mark_read 1

  Attaching kprobe:ksys_*
  Attaching to 27 functions
   ksys_dup3
   ksys_fadvise64_64
   ksys_fallocate
   ksys_fchown
   ksys_ioperm
   ksys_lseek
   ksys_mmap_pgoff
   ksys_msgctl.constprop.0
   ksys_msgget
   ksys_msgrcv
   ksys_msgsnd
   ksys_pread64
   ksys_pwrite64
   ksys_read
   ksys_readahead
   ksys_semctl.constprop.0
   ksys_semget
   ksys_semtimedop
   ksys_setsid
   ksys_shmctl.constprop.0
   ksys_shmdt
   ksys_shmget
   ksys_sync
   ksys_sync_file_range
   ksys_sync_helper
   ksys_unshare
   ksys_write
  ^C

  @[
      ksys_write+5
      do_syscall_64+56
      entry_SYSCALL_64_after_hwframe+68
  ]: 2
  @[
      ksys_read+5
      do_syscall_64+56
      entry_SYSCALL_64_after_hwframe+68
  ]: 26

```
Signed-off-by: Jiri Olsa <jolsa@kernel.org>

##### Checklist


src/types.h
```diff
@@ -531,6 +531,9 @@ struct Probe
             address,
             func_offset);
   }
+
+public:
+  std::vector<std::string> funcs;
```
danobi, line 536: This can go w/ the other fields in `struct Probe`. Fields in a struct are public by default.

Also please add to the `archive()` call so the field gets serialized for AOT (hope to return to that soon...)

src/libbpf/bpf.h
```diff
@@ -112,6 +112,11 @@ enum bpf_attach_type {
 	BPF_XDP_CPUMAP,
 	BPF_SK_LOOKUP,
 	BPF_XDP,
+	BPF_SK_SKB_VERDICT,
+	BPF_SK_REUSEPORT_SELECT,
+	BPF_SK_REUSEPORT_SELECT_OR_MIGRATE,
+	BPF_PERF_EVENT,
+	BPF_TRACE_KPROBE_MULTI,
```
danobi, line 119: Mind syncing up the rest of the file too? So file doesn't get skewed too much.

src/bpftrace.cpp
```diff
@@ -52,6 +52,7 @@ namespace bpftrace {
 DebugLevel bt_debug = DebugLevel::kNone;
 bool bt_quiet = false;
 bool bt_verbose = false;
+bool bt_verbose2 = false;
```
danobi, line 55: Nit: just make `bt_verbose` an integer?

src/attached_probe.cpp
```diff
@@ -803,8 +819,66 @@ void AttachedProbe::load_prog()
...
+static inline uint64_t ptr_to_u64(const void *ptr)
+{
+  return (uint64_t)(unsigned long)ptr;
+}
+
+void AttachedProbe::attach_multi_kprobe(void)
+{
+  DECLARE_LIBBPF_OPTS(bpf_link_create_opts, opts);
+
+  const char **syms = new const char *[probe_.funcs.size()];
```
danobi, line 832: How about directly using `probe_.funcs.data()` instead of allocating a new array? I think `.data()` satisfies the C API's needs here (contiguous memory)

src/types.h
danobi: Also please add a comment describing what this field is used for (multiattach kprobes)

tests/codegen/general.cpp
```diff
@@ -35,6 +35,11 @@ class MockBPFtrace : public BPFtrace
   {
     return true;
   }
+
+  bool has_kprobe_multi(void)
+  {
+    return (*feature_).has_kprobe_multi();
```
danobi, line 41: Nit: `feature_->has_kprobe_multi()`

tests/mocks.h
```diff
@@ -79,6 +79,11 @@ class MockBPFtrace : public BPFtrace
     mock_probe_matcher = dynamic_cast<MockProbeMatcher *>(probe_matcher_.get());
   }
 
+  bool has_kprobe_multi(void)
+  {
+    return (*feature_).has_kprobe_multi();
```
danobi, line 84: Same as above

danobi (commented):
Can you also please add some runtime tests that check that only a single prog is loaded for a wildcard multiattach k[ret]probe? And that the multi-attach functions as expected

src/attached_probe.cpp
danobi: Oh, it needs an array of C strings. Still, how about using a `std::vector<char *>` instead of raw pointer? So we don't need to worry about freeing memory here (atm there's a leak)

olsajiri:
> Can you also please add some runtime tests that check that only a single prog is loaded for a wildcard multiattach k[ret]probe? And that the multi-attach functions as expected

ok, will add

src/types.h
olsajiri: ok, will check, thanks

src/libbpf/bpf.h
olsajiri: sure, will add the rest, thanks

src/bpftrace.cpp
olsajiri: I like adding bt_verbose2 because it's simple ;-) there are many bt_verbose condition instances of that would need to change.. maybe I'm biased from perf code but that seems more clear to me than adding other machinery/funcs to print messages with proper verbose level

src/attached_probe.cpp
olsajiri: ah right, will check on that, thanks

tests/mocks.h
olsajiri: ok, thanks

tests/codegen/general.cpp
olsajiri: ok

olsajiri:
I did the requested changes, but I still need to figure out some specific tests for this

fbs (approved):

danobi (approved):

PR #2197, opened 2022-04-21
Fix cgroup_path builtin

sumanthkorikkar:
bpftrace -e '{ print(cgroup_path(cgroup)); exit(); }'

The addrspace is not set and bpf code would resolve it  to
bpf_probe_read, which is invalid on s390.

As cgroup_path is already allocated on the bpf stack, a normal copy
should be enough. No need to set the addrspace or use the
CreateProbeRead variant.


viktormalik (approved):
Makes sense to me, there's no reason to proberead here as everything is on the stack.

cc @lenticularis39

lenticularis39:
Nice catch, it's a structure built on the stack in codegen, so there's no reason to proberead here :) I probably just forgot to add it there, since my original implementation was different.

PR #2203, opened 2022-04-29
Print all maps to stdout on SIGUSR1

andreasgerstmayr:
This commit adds the functionality to print all eBPF maps to stdout
when the bpftrace process receives a SIGUSR1 signal. This feature is
especially useful for long-running bpftrace programs, or to inspect
running programs without stopping them.

Main use cases:
* users want to see the current contents of all maps at a user-choosen time (i.e. no external probe or fixed interval), without stopping the bpftrace program
* observability tools want to start long-running bpftrace programs and get the contents of their maps on demand (this PR also works with JSON output by using `-f json`)

Open questions:
* The `sigusr1` runtime test can be flaky, depending how long bpftrace takes to start the program. Any ideas how to improve it?
* I've updated the reference guide, but I'm not sure in which section I should add this feature to `man/adoc/bpftrace.adoc` - any recommendations?

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -958,10 +958,8 @@ int main(int argc, char* argv[])
   sigaction(SIGTERM, &act, NULL);
 
   // Signal handler that prints all maps when SIGUSR1 was received.
-  static auto sigusr1_handler = [&]() { bpftrace.print_maps(); };
-  struct sigaction sigusr1_act = {};
-  sigusr1_act.sa_handler = [](int) { sigusr1_handler(); };
-  sigaction(SIGUSR1, &sigusr1_act, NULL);
+  act.sa_handler = [](int) { BPFtrace::sigusr1_recv = true; };
+  sigaction(SIGUSR1, &act, NULL);
 
   err = bpftrace.run(bytecode);
   if (err)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -957,6 +957,12 @@ int main(int argc, char* argv[])
   sigaction(SIGINT, &act, NULL);
   sigaction(SIGTERM, &act, NULL);
 
+  // Signal handler that prints all maps when SIGUSR1 was received.
+  static auto sigusr1_handler = [&]() { bpftrace.print_maps(); };
+  struct sigaction sigusr1_act = {};
+  sigusr1_act.sa_handler = [](int) { sigusr1_handler(); };
+  sigaction(SIGUSR1, &sigusr1_act, NULL);
+
   err = bpftrace.run(bytecode);
   if (err)
     return err;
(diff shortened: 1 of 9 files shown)


src/main.cpp
```diff
@@ -957,6 +957,12 @@ int main(int argc, char* argv[])
   sigaction(SIGINT, &act, NULL);
   sigaction(SIGTERM, &act, NULL);
 
+  // Signal handler that prints all maps when SIGUSR1 was received.
+  static auto sigusr1_handler = [&]() { bpftrace.print_maps(); };
```
danobi, line 961: How about setting a flag in the signal handler and running `bpftrace.print_maps()` inside main loop? I highly doubt  `print_maps()` is signal safe

danobi (commented):
Idea seems reasonable to me. Have a few points on the impl

> The sigusr1 runtime test can be flaky, depending how long bpftrace takes to start the program. Any ideas how to improve it?

Maybe try running `kill ...` in the `AFTER` directive. I believe the runtime test engine synchronizes with `RUN` attaching to probes before running `AFTER`

src/main.cpp
andreasgerstmayr: Ah, I didn't realize we're in racy terrain when using signals. I've updated the handler to set a flag (similar to the current exit signal handler) and print the maps in the main loop.

andreasgerstmayr:
Thanks for the review!

> > The sigusr1 runtime test can be flaky, depending how long bpftrace takes to start the program. Any ideas how to improve it?
> 
> Maybe try running `kill ...` in the `AFTER` directive. I believe the runtime test engine synchronizes with `RUN` attaching to probes before running `AFTER`

Good idea. Yep, looking at the code, `AFTER` waits until all probes are attached. I've added a new placeholder variable `{{BPFTRACE_PID}}` and I'm using this now in the `AFTER` directive. In case I should not introduce another placeholder variable, I could also use `pidof` in `AFTER` (but that might be a problem if there are other bpftrace processes already running while running the test).

andreasgerstmayr:
Sorry about the noise - turns out shell=True in runner.py doesn't spawn a shell on my system, but does so in CI (I guess the Python version makes the difference). When it spawns a shell, p.pid contains the PID of the shell, and the new test case fails.

I've reverted this change, and now the test case passes reliably on my system and CI. Please squash the commits when merging this PR.

tests/runtime/sigusr1
```diff
@@ -0,0 +1,5 @@
+NAME print all maps
+RUN {{BPFTRACE}} -f json -e 'BEGIN { @scalar = 5; } interval:s:2 { exit(); }'
+AFTER kill -s USR1 $(pidof bpftrace)
+EXPECT @scalar[\s\S]+@scalar
+TIMEOUT 5
```
viktormalik, line 5: Could you merge this file with `tests/runtime/sigint` into something like `tests/runtime/signals`? No need to have a separate suite for each signal that we handle.

andreasgerstmayr: Yep, sure! I've updated the branch.

danobi (approved):

andreasgerstmayr:
Thanks for the review & merging! :)

PR #2204, opened 2022-05-04
Disallow different lhist bounds

lenticularis39:
bpftrace doesn't currently support having linear histograms with different bounds inside a single map (issue #2121). PR #2153 fixes this issue, but at the cost of a significant increase in BPF program size.

This patch adds detecting this unsupported situation and throwing an error into ResourceAnalyzer.

##### Checklist


CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ and this project adheres to
 
 #### Deprecated
 #### Removed
+- Disallow different lhist bounds in a single map
+  - [#2204](https://github.com/iovisor/bpftrace/pull/2204)
+
```
viktormalik, line 35: Move this to "Fixed" please. It has never been supported so it shouldn't be in "Removed".

src/ast/passes/resource_analyser.cpp
```diff
@@ -29,16 +30,22 @@ std::string get_literal_string(Expression &expr)
...
   Visit(*root_);
   prepare_seq_printf_ids();
 
-  return std::move(resources_);
+  if (!err_.str().empty())
+  {
+    out_ << err_.str();
+  }
+
+  return { std::move(resources_), !err_.str().empty() };
```
viktormalik, line 48: Since this is returning `bool`, shouldn't the return type be `std::pair<RequiredResources, bool>`? Or, maybe even better, `std::optional<RequiredResources>`?

viktormalik (commented):
Would be nice to have a test checking that `lhist` with different bounds fails.

src/ast/passes/resource_analyser.cpp
lenticularis39: All other analysers (`FieldAnalyser`, `PortabilityAnalyser`, `SemanticAnalyser`) return `int`. In case of `SemanticAnalyser` it's because it returns the number of the pass it failed in, the other ones I assume are `int` was chosen to easily allow multiple error codes in the future if necessary. Either way, wouldn't it be better to keep it consistent here?

viktormalik: Well, consistency is nice but we don't really make use of the error codes anyway. And since `ResourceAnalyser` already returns something different, I wouldn't try too hard here to be similar to other analysers.

lenticularis39: OK, that makes sense.

fbs: Also the passes are handled by the pass manager nowdays and that has its own error handling (even though its a bit shitty too). So what you do here doesn't matter too much.

lenticularis39:
Moved the change to Fixed in changelog, changed `ResourceAnalyser` return value type to `std::optional<RequiredResources>` and added a unit test for `ResourceAnalyser` to check for the new behavior.

tests/resource_analyser.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "driver.h"
+#include "mocks.h"
+
+namespace bpftrace {
+namespace test {
+namespace resource_analyser {
+
+using ::testing::_;
+
+void test(BPFtrace &bpftrace, const std::string &input, int expected_result = 0)
```
viktormalik, line 17: Shouldn't be:
```suggestion
void test(BPFtrace &bpftrace, const std::string &input, bool expected_result = true)
```

lenticularis39: Of course, I must have missed that :)

viktormalik (approved):

fbs:
woops been slow and now it needs a changelog fixup. Can you do that so we can get this in?

fbs (approved):

lenticularis39:
Changelog fixed.

PR #2208, opened 2022-05-10
Fixed the issue when using wildcards in kfunc

fengyoulin:
The kfunc and kretfunc probes that contains a wildcard always failed to attach:

```
$ sudo src/bpftrace -ve 'kfunc:vfs_* { @[probe] = count(); }'
INFO: node count: 7
Attaching 67 probes...

Error log: 

ERROR: Error loading program: kfunc:vfs_writev
$ sudo src/bpftrace -ve 'kfunc:vfs_writev* { @[probe] = count(); }'
INFO: node count: 7
Attaching 1 probe...

Error log: 

ERROR: Error loading program: kfunc:vfs_writev
```

The [bcc_prog_load_xattr](https://github.com/iovisor/bpftrace/blob/38c099dff9100bafeaaa7cee865f4dfda58134ac/src/attached_probe.cpp#L746) function internally rely on the [libbpf_find_vmlinux_btf_id](https://github.com/iovisor/bcc/blob/0064febd47972cc6c79de64639790f5889387d03/src/cc/libbpf.c#L736) function to find btf id for the kfunc and kretfunc probe, which need a exact probe name. So, I use the attach_point as program name when probe type is kfunc or kretfunc.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -691,19 +691,15 @@ void AttachedProbe::load_prog()
     if (bt_verbose)
       log_level = 1;
 
-    if (probe_.type == ProbeType::kfunc || probe_.type == ProbeType::kretfunc)
-    {
-      // Use attach_point for program name when probe type is kfunc or
-      // kretfunc, otherwise will get a ENOENT when using a wildcard
-      name = probe_.attach_point;
-    }
-    else
+    if (probe_.type == ProbeType::kprobe || probe_.type == ProbeType::kretprobe)
     {
       // Use orig_name for program name so we get proper name for
       // wildcard probes, replace wildcards with '.'
       name = probe_.orig_name;
       std::replace(name.begin(), name.end(), '*', '.');
     }
+    else
+      name = probe_.name;
 
     // bpf_prog_load rejects colons in the probe name,
     // so start the name after the probe type, after ':'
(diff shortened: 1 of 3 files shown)


src/attached_probe.cpp
```diff
@@ -691,10 +691,19 @@ void AttachedProbe::load_prog()
...
+      // kretfunc, otherwise will get a ENOENT when using a wildcard
+      name = probe_.attach_point;
+    }
+    else
+    {
+      // Use orig_name for program name so we get proper name for
+      // wildcard probes, replace wildcards with '.'
+      name = probe_.orig_name;
+      std::replace(name.begin(), name.end(), '*', '.');
+    }
```
viktormalik, line 706: Since the commit that caused this is 2b7e38ab60bf288ed8681ae08135710b5f5f6b1e, I'd suggest something like:
```suggestion
    if (probe_.type == ProbeType::kprobe || probe_.type == ProbeType::kretprobe)
    {
        // Use orig_name for program name so we get proper name for
        // wildcard probes, replace wildcards with '.'
        name = probe_.orig_name;
        std::replace(name.begin(), name.end(), '*', '.');
    }
    else
        name = probe_.name
```

viktormalik (commented):
A good catch, thanks!

Could you perhaps add a runtime test attaching a kfunc with wildcard to make sure that this doesn't happen again? Also please add a CHANGELOG entry.

src/attached_probe.cpp
fengyoulin: Yeah, only use the orig_name for kprobe and kretprobe, this is better.

viktormalik (approved):
LGTM

PR #2216, opened 2022-05-16
Add offsetof() builtin

Rtoax:
Get the offset of the element in the struct.

```
struct Foo {
    int x;
    long l;
    char c;
}
BEGIN
{
    printf("%ld, %ld\n", offsetof(struct Foo, c), sizeof(struct Foo));
    printf("%ld\n", offsetof(struct Foo, c));
    printf("%ld\n", offsetof(struct Foo, x));
    printf("%ld\n", offsetof(struct task_struct, comm));
    printf("%ld\n", offsetof(struct task_struct, flags));
    printf("%ld\n", offsetof(struct task_struct, mm));
    printf("%ld\n", offsetof(struct list_head, next));
    printf("%ld\n", offsetof(struct list_head, prev));
}
```

Result:

```
Attaching 1 probe...
16, 24
16
0
3216
44
2504
0
8
```

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -2222,6 +2222,7 @@ Tracing block I/O sizes > 0 bytes
 - `kptr(void *p)` - Annotate as kernelspace pointer
 - `macaddr(char[6] addr)` - Convert MAC address data
 - `bswap(uint[8|16|32|64] n)` - Reverse byte order
+- `offsetof(struct, element)` - Offset of element in structure
 
 Some of these are asynchronous: the kernel queues the event, but some time later (milliseconds) it is
 processed in user-space. The asynchronous actions are: `printf()`, `time()`, and `join()`. Both `ksym()`
@@ -3135,6 +3136,32 @@ Example:
 # bpftrace -e 'BEGIN { $i = (uint32)0x12345678; printf("Reversing byte order of 0x%x ==> 0x%x\n", $i, bswap($i)); }'
 Attaching 1 probe...
 Reversing byte order of 0x12345678 ==> 0x78563412
+```
+
+# 31. `offsetof`: Offset of element in structure
+
+Syntax: `offsetof(struct, element)`
+
+Get the offset of the element in the struct.
+
+Examples:
+
+```
+#!/usr/bin/env bpftrace
+#include <linux/sched.h>
+
+BEGIN
+{
+    printf("Offsetof %ld\n", offsetof(struct task_struct, flags));
+    printf("Offsetof %ld\n", offsetof(struct task_struct, comm));
+    exit();
+}
+```
+
+```
+Attaching 1 probe...
+Offsetof 44
+Offsetof 3216
 ```
 
 # Map Functions
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1026,6 +1026,13 @@ void CodegenLLVM::visit(Call &call)
   {
     expr_ = b_.getInt64(call.vargs->at(0)->type.GetSize());
   }
+  else if (call.func == "offsetof")
+  {
+    auto &element_arg = *call.vargs->at(1);
+    String &element = static_cast<String&>(element_arg);
+    auto &field = call.vargs->at(0)->type.GetField(element.str);
+    expr_ = b_.getInt64(field.offset);
+  }
   else if (call.func == "strncmp") {
     uint64_t size = (uint64_t)*bpftrace_.get_int_literal(call.vargs->at(2));
     const auto& left_arg = call.vargs->at(0);
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -126,6 +126,11 @@ void SemanticAnalyser::visit(Identifier &identifier)
     identifier.type = CreateInt(
         std::get<0>(getIntcasts().at(identifier.ident)));
   }
+  else if (func_ == "offsetof")
+  {
+    // Handling the second arg of offsetof (a field name)
+    identifier.type = CreateNone();
+  }
   else {
     identifier.type = CreateNone();
     LOG(ERROR, identifier.loc, err_)
@@ -1132,6 +1137,26 @@ void SemanticAnalyser::visit(Call &call)
 
     call.type = CreateUInt64();
   }
+  else if (call.func == "offsetof")
+  {
+    check_nargs(call, 2);
+    auto &element_arg = *call.vargs->at(1);
+    String &element = static_cast<String&>(element_arg);
+    auto &arg = *call.vargs->at(0);
+    // check_arg() cannot be applied to offsetof(), because
+    // Type::record is first arg of offsetof(). check_arg()
+    // return true anyway.
+    if (arg.type.type != Type::record) {
+      LOG(ERROR, call.loc, err_)
+          << call.func << "() only supports " << Type::record << " arguments ("
+          << arg.type.type << " provided)";
+    }
+    // Only case Type::record can call HasField().
+    else if (!call.vargs->at(0)->type.HasField(element.str))
+      LOG(ERROR, call.loc, err_)
+          << "struct has no this field";
+    call.type = CreateUInt64();
+  }
   else if (call.func == "path")
   {
     if (!bpftrace_.feature_->has_d_path())
@@ -1976,6 +2001,13 @@ void SemanticAnalyser::visit(FieldAccess &acc)
 {
   // A field access must have a field XOR index
   assert((acc.field.size() > 0) != (acc.index >= 0));
+
+  if (func_ == "offsetof")
+  {
+    LOG(ERROR, acc.loc, err_)
+		<< "offsetof not support subfield yet";
+    return;
+  }
 
   acc.expr->accept(*this);
 
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -38,7 +38,7 @@ vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#\*])+
 builtin  arg[0-9]|args|cgroup|comm|cpid|cpu|ctx|curtask|elapsed|func|gid|nsecs|pid|probe|rand|retval|sarg[0-9]|tid|uid|username
-call     avg|buf|cat|cgroupid|clear|count|delete|exit|hist|join|kaddr|kptr|ksym|lhist|macaddr|max|min|ntop|override|print|printf|cgroup_path|reg|signal|sizeof|stats|str|strftime|strncmp|sum|system|time|uaddr|uptr|usym|zero|path|unwatch|bswap
+call     avg|buf|cat|cgroupid|clear|count|delete|exit|hist|join|kaddr|kptr|ksym|lhist|macaddr|max|min|ntop|override|print|printf|cgroup_path|reg|signal|sizeof|stats|str|strftime|strncmp|sum|system|time|uaddr|uptr|usym|zero|path|unwatch|bswap|offsetof
 
 /* Don't add to this! Use builtin OR call not both */
 call_and_builtin kstack|ustack
@@ -170,10 +170,10 @@ bpftrace|perf           { return Parser::make_STACK_MODE(yytext, loc); }
 
 struct|union|enum       yy_push_state(STRUCT, yyscanner); buffer.clear(); struct_type = yytext;
 <STRUCT,BRACE>{
-  "*"|")"               {
+  "*"|")"|","           {
                           if (YY_START == STRUCT)
                           {
-                            // Finished parsing the typename of a cast
+                            // Finished parsing the typename of a cast or a call arg
                             // Put the cast type into a canonical form by trimming
                             // and then inserting a single space.
                             yy_pop_state(yyscanner);
--- a/tests/codegen/call_offsetof.cpp
+++ b/tests/codegen/call_offsetof.cpp
@@ -1,0 +1,16 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_offsetof)
+{
+  test("struct Foo { int x; long l; char c; } BEGIN { @x = offsetof(struct Foo, x); }",
+
+       NAME);
+}
+
+} // namespace codegen
+} // namespace test
+} // namespace bpftrace
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -193,3 +193,8 @@ PROG BEGIN { printf("size=%d\n", sizeof(struct task_struct)); exit(); }
 EXPECT ^size=
 TIMEOUT 5
 REQUIRES_FEATURE btf
+
+NAME offsetof
+PROG struct Foo { int x; long l; char c; } BEGIN { printf("%ld\n", offsetof(struct Foo, x)); exit(); }
+EXPECT ^0$
+TIMEOUT 5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to
   - [#2166](https://github.com/iovisor/bpftrace/pull/2166)
 - Print all maps to stdout on `SIGUSR1`
   - [#2203](https://github.com/iovisor/bpftrace/pull/2203)
+- Add new function, `offsetof`, get the offset of the element in the struct
+  - [#2216](https://github.com/iovisor/bpftrace/pull/2216)
 
 #### Changed
 - Use auto-resolution of library paths for tools
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -336,6 +336,7 @@ primary_expr:
         |       int                { $$ = $1; }
         |       STRING             { $$ = new ast::String($1, @$); }
         |       STACK_MODE         { $$ = new ast::StackMode($1, @$); }
+        |       CALL               { $$ = new ast::String($1, @$); }
         |       BUILTIN            { $$ = new ast::Builtin($1, @$); }
         |       CALL_BUILTIN       { $$ = new ast::Builtin($1, @$); }
         |       LPAREN expr RPAREN { $$ = $2; }
--- a/tests/codegen/llvm/call_offsetof.ll
+++ b/tests/codegen/llvm/call_offsetof.ll
@@ -1,0 +1,45 @@
+; ModuleID = 'bpftrace'
+source_filename = "bpftrace"
+target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n32:64-S128"
+target triple = "bpf-pc-linux"
+
+; Function Attrs: nounwind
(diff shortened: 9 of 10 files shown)


tests/codegen/call_offsetof.cpp
```diff
@@ -0,0 +1,16 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_offsetof)
+{
+  test("struct Foo { int x; long l; char c; } BEGIN { @x = offsetof(c, struct Foo) }",
```
viktormalik, line 9: This is a bit tricky as the test may fail on some arches due to different ABI. Since we don't have a way to filter codegen tests per-arch, I'd suggest testing the offset of `x` which will be 0 pretty much everywhere.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -126,6 +126,10 @@ void SemanticAnalyser::visit(Identifier &identifier)
     identifier.type = CreateInt(
         std::get<0>(getIntcasts().at(identifier.ident)));
   }
+  else if (func_ == "offsetof")
+  {
+    identifier.type = CreateString(identifier.ident.size() + 1);
+  }
```
viktormalik, line 132: I'm not sure what the purpose of this code is. Why not set the args types when analysing the call?

viktormalik (changes requested):

viktormalik (changes requested):
I'm missing more tests here, too. There should be unit tests checking for different valid and invalid combinations of arguments as well as runtime tests checking for the actual functionality.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1026,6 +1026,13 @@ void CodegenLLVM::visit(Call &call)
   {
     expr_ = b_.getInt64(call.vargs->at(0)->type.GetSize());
   }
+  else if (call.func == "offsetof")
+  {
+    auto &element_arg = *call.vargs->at(0);
+    String &element = static_cast<String&>(element_arg);
+    auto &field = call.vargs->at(1)->type.GetField(element.str);
```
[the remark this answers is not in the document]
Rtoax: > What if the struct doesn't have the given field? This will throw an exception.

```
struct Foo {
    int x;
    long l;
    char c;
}
BEGIN
{
    printf("%ld\n", offsetof(y, struct Foo));
}
```

Will

```
ERROR: Failed to compile: struct has no field named y
```

docs/reference_guide.md
```diff
@@ -3137,6 +3138,32 @@ Attaching 1 probe...
 Reversing byte order of 0x12345678 ==> 0x78563412
 ```
 
+# 31. `offsetof`: Offset of element in structure
+
+Syntax: `offsetof(element, struct)`
```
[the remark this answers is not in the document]
Rtoax: > You should respect the common order of args for `offsetof` which is `(struct, element)`.

There is no corresponding `vargs` in `src/parser.yy` to handle `struct xxx, xxx`, Need to add a new type?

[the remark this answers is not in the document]
viktormalik: It should be sufficient to do:
```diff
diff --git a/src/lexer.l b/src/lexer.l
index 2868859c..b778cf26 100644
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -170,7 +170,7 @@ bpftrace|perf           { return Parser::make_STACK_MODE(yytext, loc); }
 
 struct|union|enum       yy_push_state(STRUCT, yyscanner); buffer.clear(); struct_type = yytext;
 <STRUCT,BRACE>{
-  "*"|")"               {
+  "*"|")"|","           {
                           if (YY_START == STRUCT)
                           {
                             // Finished parsing the typename of a cast

```

fbs:
> may i ask you a question about #2216,
> 
> i just add `offsetof(element, struct)`, and it's OK, but add `offsetof(struct, element)`, it's failed.
> 
> in `src/parser.yy` add:
> 
> ```
> struct:
>                 STRUCT IDENT { $$ = new ast::Identifier("struct " + $2, @$); }
>                 ;
> 
> vargs:
>                 vargs "," expr { $$ = $1; $1->push_back($3); }
>         |       vargs "," struct { $$ = $1; $1->push_back($3); }
>         |       struct
>                 {
>                   $$ = new ast::ExpressionList;
>                   $$->push_back($1);
>                   std::cout << ">>>3:" << std::endl;
>                 }
>         |       expr           { $$ = new ast::ExpressionList; $$->push_back($1); }
>                 ;
> ```
> 
> is not helpful. what should i do to fix this.

What is the exact issue you bump into?

docs/reference_guide.md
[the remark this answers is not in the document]
Rtoax: > ```diff
> "*"|")"|","
> ```

Thanks a lot.

Rtoax:
> 

Yes, i just Solved with the guidance of @viktormalik , thanks.

Rtoax:
> I'm missing more tests here, too. There should be unit tests checking for different valid and invalid combinations of arguments as well as runtime tests checking for the actual functionality.

Where should i add more tests.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1026,6 +1026,15 @@ void CodegenLLVM::visit(Call &call)
   {
     expr_ = b_.getInt64(call.vargs->at(0)->type.GetSize());
   }
+  else if (call.func == "offsetof")
+  {
+    auto &element_arg = *call.vargs->at(1);
+    String &element = static_cast<String&>(element_arg);
+    if (!call.vargs->at(0)->type.HasField(element.str))
+      LOG(ERROR) << "struct has no field named '" + element.str + "'";
```
viktormalik, line 1034: Could this check be in semantic analyser? It'd make more sense there and you can add location info to the error.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -126,6 +126,10 @@ void SemanticAnalyser::visit(Identifier &identifier)
     identifier.type = CreateInt(
         std::get<0>(getIntcasts().at(identifier.ident)));
   }
+  else if (func_ == "offsetof")
+  {
+    identifier.type = CreateNone();
+  }
```
viktormalik, line 132: Why is this necessary?

src/lexer.l
```diff
@@ -170,7 +170,7 @@ bpftrace|perf           { return Parser::make_STACK_MODE(yytext, loc); }
 
 struct|union|enum       yy_push_state(STRUCT, yyscanner); buffer.clear(); struct_type = yytext;
 <STRUCT,BRACE>{
-  "*"|")"               {
+  "*"|")"|","           {
```
viktormalik, line 173: Could you please adjust the comment below, too? Something like:
```
// Finished parsing the typename of a cast or a call arg
// ...
```

tests/codegen/call_offsetof.cpp
```diff
@@ -0,0 +1,23 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_offsetof)
+{
+  test("struct Foo { int x; long l; char c; } BEGIN { @x = offsetof(struct Foo, x); @c = offsetof(struct Foo, c); }",
```
viktormalik, line 9: I'd just leave `@x = offsetof(struct Foo, x)` here, the other offset depends on ABI, so it may not pass on some arches.

tests/codegen/call_offsetof.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+       NAME);
+}
+
+TEST(codegen, call_offsetof_task_struct)
+{
+  test("#include <linux/sched.h> \
+  BEGIN { @x = offsetof(struct Foo, comm); }",
```
viktormalik, line 17: I'm afraid that this will be very flaky - a change of `struct task_struct` layout in kernel will break it. I don't think that we need this test.

tests/runtime/builtin
```diff
@@ -193,3 +193,8 @@ PROG BEGIN { printf("size=%d\n", sizeof(struct task_struct)); exit(); }
 EXPECT ^size=
 TIMEOUT 5
 REQUIRES_FEATURE btf
+
+NAME offsetof
+PROG struct Foo { int x; long l; char c; } BEGIN { printf("%ld %ld\n", offsetof(struct Foo, c), offsetof(struct Foo, c)); }
+EXPECT ^0 16$
```
viktormalik, line 199: Same as with the codegen test, the value '16' depends on ABI, so I'd stick with checking the offset of `x`.

viktormalik (changes requested):

viktormalik:
> Where should i add more tests.

Some unit tests in `tests/semantic_analyser.cpp` would be useful. Something similar to what we have for other calls.

src/ast/passes/semantic_analyser.cpp
Rtoax: > Why is this necessary?

I am also curious, if there is no such sentence, it will report an error

```
./offsetof.bt:14:43-44: ERROR: Unknown identifier: 'x'
    printf("%ld\n", offsetof(struct Foo, x));
                                          ~
./offsetof.bt:16:51-56: ERROR: Unknown identifier: 'flags'
    printf("%ld\n", offsetof(struct task_struct, flags));
                                                  ~~~~~
./offsetof.bt:17:51-53: ERROR: Unknown identifier: 'mm'
    printf("%ld\n", offsetof(struct task_struct, mm));
                                                  ~~
./offsetof.bt:18:49-53: ERROR: Unknown identifier: 'next'
    printf("%ld\n", offsetof(struct list_head, next));
                                                ~~~~
./offsetof.bt:19:49-53: ERROR: Unknown identifier: 'prev'
    printf("%ld\n", offsetof(struct list_head, prev));
```

viktormalik: Ah, makes sense, the field is an unknown identifier. I'm fine with keeping it as it is now, maybe just add a comment, something like:
```
// Handling the second arg of offsetof (a field name)
```

tests/codegen/call_offsetof.cpp
fbs: instead a test for the more exotic structs might be good. What about function pointers, anonymous structs and other nesting?

Rtoax: Add `union`, `function pointers`, `anonymous structs` unit test in `tests/semantic_analyser.cpp`, looks like works:

```
$ sudo ./bpftrace_test --gtest_filter=semantic_analyser.call_offsetof
Note: Google Test filter = semantic_analyser.call_offsetof
[==========] Running 1 test from 1 test suite.
[----------] 1 test from semantic_analyser
[ RUN      ] semantic_analyser.call_offsetof
[       OK ] semantic_analyser.call_offsetof (1594 ms)
[----------] 1 test from semantic_analyser (1594 ms total)

[==========] 1 test from 1 test suite ran. (1594 ms total)
[  PASSED  ] 1 test.
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1132,6 +1137,26 @@ void SemanticAnalyser::visit(Call &call)
...
+  else if (call.func == "offsetof")
+  {
+    check_nargs(call, 2);
+    auto &element_arg = *call.vargs->at(1);
+    String &element = static_cast<String&>(element_arg);
+    auto &arg = *call.vargs->at(0);
+    // check_arg() cannot be applied to offsetof(), because
+    // Type::record is first arg of offsetof(). check_arg()
+    // return true anyway.
+    if (arg.type.type != Type::record) {
```
viktormalik, line 1149: This comment is very confusing to me. What is the problem here? `check_arg` should do the same comparison as you do here.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1132,6 +1137,26 @@ void SemanticAnalyser::visit(Call &call)
...
+      LOG(ERROR, call.loc, err_)
+          << call.func << "() only supports " << Type::record << " arguments ("
+          << arg.type.type << " provided)";
+    }
+    // Only case Type::record can call HasField().
+    else if (!call.vargs->at(0)->type.HasField(element.str))
+      LOG(ERROR, call.loc, err_)
+          << "struct has no this field";
```
viktormalik, line 1157:
```suggestion
          << "struct \'" << arg.type.GetName() << "\' has no field \'" 
          << ... << "\'";
```
Let's make the error messages as informative as possible.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1977,6 +2002,13 @@ void SemanticAnalyser::visit(FieldAccess &acc)
   // A field access must have a field XOR index
   assert((acc.field.size() > 0) != (acc.index >= 0));
 
+  if (func_ == "offsetof")
+  {
+    LOG(ERROR, acc.loc, err_)
+		<< "offsetof not support subfield yet";
+    return;
+  }
```
viktormalik, line 2010: This can be covered when handling the call. Also it's not entirely correct as it'd prevent using field access in the first argument.

viktormalik (commented):
Does this support passing expressions to `offsetof`? Something like `offsetof($var, field)`?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1132,6 +1137,26 @@ void SemanticAnalyser::visit(Call &call)
 
     call.type = CreateUInt64();
   }
+  else if (call.func == "offsetof")
+  {
+    check_nargs(call, 2);
+    auto &element_arg = *call.vargs->at(1);
+    String &element = static_cast<String&>(element_arg);
```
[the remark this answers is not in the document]
Rtoax: > I don't think that this should be a `String` as it'd have to be quoted. Struct field names should follow the rules for identifiers, so I think that the proper thing to do here is to `dynamic_cast` the first arg to `Identifier *` and then check if it hasn't returned `nullptr` (which would also cover the case then `f1.f2` is used).

I try

```suggestion
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1029,8 +1029,8 @@ void CodegenLLVM::visit(Call &call)
   else if (call.func == "offsetof")
   {
     auto &element_arg = *call.vargs->at(1);
-    String &element = static_cast<String&>(element_arg);
-    auto &field = call.vargs->at(0)->type.GetField(element.str);
+    Identifier &element = dynamic_cast<Identifier&>(element_arg);
+    auto &field = call.vargs->at(0)->type.GetField(element.ident);
     expr_ = b_.getInt64(field.offset);
   }
   else if (call.func == "strncmp") {
```

 but get error:

```
ERROR: Failed to compile: std::bad_cast
```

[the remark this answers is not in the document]
viktormalik: I'd try:
```cpp
auto element = dynamic_cast<Identifier *>(call.vargs->at(1));
if (element == nullptr)
   ...
```
notice using a pointer (which allows to check for NULL) instead of a reference.

Rtoax:
> Does this support passing expressions to `offsetof`? Something like `offsetof($var, field)`?

```
printf("task_struct: %ld\n", offsetof(*$task, comm));
```

same as:

```
printf("task_struct: %ld\n", offsetof(struct task_struct, comm));
```

however, `printf("task_struct: %ld\n", offsetof($task, comm));` is not ok.

src/ast/passes/semantic_analyser.cpp
Rtoax: > This can be covered when handling the call. Also it's not entirely correct as it'd prevent using field access in the first argument.

because `offsetof(record, x)` is different with all other `call`, `builtin`(special vargs), Some places require special handling, I'll try a generic way to see if it works

src/ast/passes/semantic_analyser.cpp
Rtoax: > This comment is very confusing to me. What is the problem here? `check_arg` should do the same comparison as you do here.

because `offsetof(record, x)` is different with all other `call`, `builtin`(special vargs), when `Type::record ` in `check_arg()` must be `is_final_pass()`, obviously, for `obbsetof()` `check_arg()` return `true` anyway.

viktormalik: Sorry, I'm still a bit confused. IIUC, the problem is that the subsequent access to the field must be done on record types only and `check_arg` only does that check in the final pass. This problem can be solved by doing:
```cpp
else if (is_final_pass() && !call.vargs->at(0)->type.HasField(element.str))
```
below

src/ast/passes/semantic_analyser.cpp
viktormalik: If the `dynamic_cast` approach that I suggested above works, then you can throw an error when the second arg is not an identifier in the call visitor.

viktormalik:
Yeah, `offsetof` in kernel requires a type, but for us, supporting expressions is fairly easy. As you noted, it already works with the current approach, so I'd suggest adding tests for it and making it explicitly supported in documentation.

Rtoax:
This PR may not satisfy `offsetof(struct Foo, a.b.c)`, I try to use other methods, this PR is temporarily suspended here.

BurntBrunch:
I have a concrete use case for this functionality. I have a script that needs to do the equivalent of `container_of($val, struct foo, bar)`. This is possible by manually adjusting the pointer via a fixed offset. However, when deploying this across all our servers, we run into compatibility issues due to changing struct layout. 

`offsetof` as a builtin would suffice to make this compatible across all kernel versions that have the struct and field pair.

Further, I don't see exactly why we need compound expressions (a.b.c) when that's just a sum of individual offsets and can be expressed, however inconveniently, via the basic single-member form.

What would it take to merge the single-member form and iterate on the compound expressions at a later time?

Rtoax:
@BurntBrunch @viktormalik @fbs @FooBarWidget Do we need a `offsetof()`  support `offsetof($s, a)` only, not support `offsetof($s, a.b.c)`, With my current analysis, it is difficult to support the application scenario of `a.b.c`, Like @BurntBrunch said above.

viktormalik:
@Rtoax mind giving this a push so that we can merge it? It would be fine just to support single-member form for now.

Rtoax:
> not support subfield yet

I'll try to start another new PR.

Rtoax:
> @Rtoax mind giving this a push so that we can merge it? It would be fine just to support single-member form for now.

Hi, @viktormalik , Please help!! I got a problem

```bash
struct offsetof {
  int offsetof;
};
BEGIN {
  @x = offsetof(struct offsetof, offsetof);
}
```

How should we solve this situation?! Please give me some advise, Thank you!

viktormalik:
That should be a valid expression, returning the offset of the `offsetof` field in the `struct offsetof` type, i.e. 0. Does parser have a problem with that?

Rtoax:
> That should be a valid expression, returning the offset of the `offsetof` field in the `struct offsetof` type, i.e. 0. Does parser have a problem with that?

I just open a new PR https://github.com/iovisor/bpftrace/pull/2565, please review, thanks.

PR #2224, opened 2022-05-19
Fix sizeof() reference guide.

Rtoax:
ERROR: Cannot cast to struct type "struct Foo"
struct Foo { int x; char c; } BEGIN { printf("%d\n", sizeof(((struct Foo)0).c)); }
                                                     ~~~~~~~~~~~~~~~~~~~~~
```
$ sudo bpftrace -e 'struct Foo { int x; char c; } BEGIN { printf("%d\n", sizeof(((struct Foo)0).c)); }'
stdin:1:54-75: ERROR: Cannot cast to struct type "struct Foo"
struct Foo { int x; char c; } BEGIN { printf("%d\n", sizeof(((struct Foo)0).c)); }
                                                     ~~~~~~~~~~~~~~~~~~~~~~
```

```
$ sudo bpftrace -e 'struct Foo { int x; char c; } BEGIN { printf("%d\n", sizeof(((struct Foo*)0)->c)); }'
Attaching 1 probe...
1
```

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -2952,7 +2952,7 @@ Examples:
 Attaching 1 probe...
 8
 
-# bpftrace -e 'struct Foo { int x; char c; } BEGIN { printf("%d\n", sizeof(((struct Foo)0).c)); }'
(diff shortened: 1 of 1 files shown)


fbs:
thanks

PR #2225, opened 2022-05-22
Add undump tool to tracing UNIX domain socket packages receive.

Rtoax:
This is a bpftrace version of the bcc examples/tracing of the same name, [undump](https://github.com/iovisor/bcc/blob/master/examples/tracing/undump.py)

* `tcpdump` not support `unix domain socket`;
* first tool use `buf()` builtin;

# Usage

Terminal 1, UNIX Socket Server:

```
$ nc -lU /var/tmp/unixsocket
# receive from Client
Hello, world
123abc
```

Terminal 2, UNIX socket Client:

```
$ nc -U /var/tmp/unixsocket
# Input some lines
Hello, world
123abc
```

Terminal 3, receive tracing:

```
$ sudo ./undump.bt
Attaching 3 probes...
Dump UNIX socket packages RX. Ctrl-C to end
TIME     COMM             PID      SIZE     DATA
20:40:11 nc               139071   13       Hello, world\x0a
20:40:14 nc               139071   7        123abc\x0a
^C
```

### DIFF
--- a/tools/undump.bt
+++ b/tools/undump.bt
@@ -1,0 +1,34 @@
+#!/usr/bin/env bpftrace
+/*
+ * undump	Trace unix domain socket package receive.
+ *		For Linux, uses bpftrace and eBPF.
+ *
+ * Also a basic example of bpftrace.
+ *
+ * This is a bpftrace version of the bcc examples/tracing of the same name.
+ *
+ * USAGE: undump.bt
+ *
+ * Copyright 2022 CESTC, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ *
+ * 22-May-2022	Rong Tao	Created this.
+ */
+#include <linux/skbuff.h>
+
+BEGIN
+{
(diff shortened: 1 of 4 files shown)


tools/undump.bt
```diff
@@ -0,0 +1,34 @@
...
+BEGIN
+{
+	printf("Dump UNIX socket packages RX. Ctrl-C to end\n");
+	printf("%-8s %-16s %-8s %-8s %-s\n", "TIME", "COMM", "PID", "SIZE", "DATA");
+}
+
+kprobe:unix_stream_read_actor
+{
+	$skb = (struct sk_buff *)arg0;
+	time("%H:%M:%S ");
```
fbs, line 28: would be nice if we can get higher accuracy then seconds, do we have mili/micro seconds in strftime yet?

Rtoax: > would be nice if we can get higher accuracy then seconds, do we have mili/micro seconds in strftime yet?

Too bad it's not supported mili/micro seconds

fbs:
can you rebase to fix the changelog?

PR #2236, opened 2022-05-25
Handle enum values in tracepoint format defs

viktormalik:
It may happen that a tracepoint format definition contains enum values instead of constants for array sizes. Not sure if this is a kernel bug or a legit behaviour but it has happened at my system (5.17.4-200.fc35.x86_64 kernel):
```
# cat /sys/kernel/debug/tracing/events/block/block_rq_issue/format
[...]
	field:char comm[TASK_COMM_LEN];	offset:40;	size:16;	signed:1;
[...]
```
breaking `tools/bitesize.bt` (fails with `error: use of undeclared identifier 'TASK_COMM_LEN'`).

This PR fixes the problem by inserting the enum value into `bpftrace.btf_set_` to get the definition from BTF.

##### Checklist


danobi:
LGTM, but can we add a unit test for this?

viktormalik:
> LGTM, but can we add a unit test for this?

Good idea, I extended an existing unit test for tracepoint parsing.

PR #2242, opened 2022-05-26
man: Rename *.8 to *.bt.8 to avoid conflicts with bcc

klemensn:
- man: Rename *.8 to *.bt.8 to avoid conflicts with bcc
- man: biolatency.bt: Fix biosnoop reference


viktormalik (approved):
Makes sense to me, thanks!

danobi:
How are naming conflicts usually handled for other packages? Are files renamed downstream?

klemensn:
> How are naming conflicts usually handled for other packages? Are files renamed downstream?

bpftrace/bcc is a simple case where upstream simply made the mistake to put their tools into a new namespace but kept squatting the old manual page namespace.

---
The old [vis(1)](https://man.openbsd.org/vis.1) (originated in 4.4BSD) and the comparatively newborn text editor [vis(1)](https://github.com/martanne/vis) for example came up with the same three letter acronym.

in OpenBSD, one has been a standard part of the operating system for decades, while the other is optional third-party software.
Thus the project simply ships the latter as [vise(1)](https://github.com/openbsd/ports/blob/4283c21e786ae0325a549a9df91bdab18cbc87ed/editors/vis/Makefile#L51) to avoid conflicts and most certainly unavoidable regressions.

This type of conflict is only relevant/possible if system administrators decide to mix well defined namespaces (all of `/bin`, `/sbin`, `/usr/bin` and `/usr/sbin`) with an ever-changing collection of third-party software that is maintained on a best-effort basis (`/usr/local/bin` for ports/packages and whatever site-local paths are used).


nixpkgs on the other hand provides the editor and some some audio visualizing software under the same namespace:
```
$ vis
The program 'vis' is not in your PATH. It is provided by several packages.
You can make it available in an ephemeral shell by typing one of the following:
  nix-shell -p cli-visualizer
  nix-shell -p vis
```
OpenBSD's obvious arguments don't apply here, nixpkgs `PATH` handling is special compared to a traditional namespace hierachy, but I'd still argue that nixpkgs rename of the tools to avoid using both.

---
Just like the `vis` case, `xsd` on NixOS also comes to mind:
```
xsd
The program 'xsd' is not in your PATH. It is provided by several packages.
You can make it available in an ephemeral shell by typing one of the following:
  nix-shell -p mono
  nix-shell -p mono4
  nix-shell -p mono5
  nix-shell -p xsd
```

These are the [Mono Project](https://github.com/mono/mono/blob/41d4eb187476f9f983a2a7ec6f851bac6cd5e44e/man/xsd.1) (.NET) and [XSD: XML Data Binding for C++](https://www.codesynthesis.com/products/xsd/).

There does not seem to be an easy tie breaker for conflicts of this type.
Operating systems/package distributions can prioritise one package, e.g. due to significance or age, and rename other packages.

FreeBSD for example renames the latter to [xsdxx(1)](https://cgit.freebsd.org/ports/tree/devel/xsd/files/patch-xsd_xsd_makefile?id=f484b5cb288e4d041d65a6ea3d5d25b4dc2b64cb), which seems reasonable.

danobi:
Thanks for the detailed explanation. Sounds like renaming our man files is the easiest way to go then.

danobi (approved):

PR #2243, opened 2022-05-27
Libbpf note that btf__get_nr_types() == btf__type_cnt() - 1.

Rtoax:
See also: https://github.com/libbpf/libbpf/blob/master/src/btf.h#L135

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -31,7 +31,7 @@ namespace bpftrace {
 static __u32 type_cnt(const struct btf *btf)
 {
 #ifdef HAVE_LIBBPF_BTF_TYPE_CNT
-  return btf__type_cnt(btf);
+  return btf__type_cnt(btf) - 1;
 #else
   return btf__get_nr_types(btf);
 #endif


viktormalik:
I'm wondering what's the proper way of indexing through BTF types now. It was `1 <= i <= btf__get_nr_types()`, does it still start with 1 after this change?

Rtoax:
> I'm wondering what's the proper way of indexing through BTF types now. It was `1 <= i <= btf__get_nr_types()`, does it still start with 1 after this change?

in https://github.com/libbpf/libbpf/blob/master/src/btf.c#L441, `btf__get_nr_types = btf__type_cnt - 1`, whether `1 <= i <= btf__get_nr_types()` is start with 1 or not

```c
__u32 btf__get_nr_types(const struct btf *btf)
{
	return btf->start_id + btf->nr_types - 1;
}

__u32 btf__type_cnt(const struct btf *btf)
{
	return btf->start_id + btf->nr_types;
}
```

viktormalik:
Ok, it still seems to start with 1, so we don't need to change anything else.

viktormalik (approved):

danobi:
Looking at https://github.com/libbpf/libbpf/commit/596c9a2d771683889c8c4989acebece0f8e94416, this change seems right to me

danobi (approved):

PR #2244, opened 2022-05-27
Add executable permission to tools.

Rtoax:
It is best to add executable permissions to the toolset, which is more convenient to use.


danobi (approved):
Good idea

PR #2256, opened 2022-06-03
tools: Allow skipping examples

klemensn:
Default behaviour stays unchanged but distribution package maintainers
can now pass `-DINSTALL_TOOL_DOCS=OFF` to not ship them.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,6 +1,11 @@
+option(INSTALL_TOOL_DOCS "Install example text files demonstrating tools" ON)
+
+if (INSTALL_TOOL_DOCS)
+  file(GLOB TXT_FILES *.txt)
+  list(REMOVE_ITEM TXT_FILES ${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.txt)
+  install(FILES ${TXT_FILES} DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/bpftrace/tools/doc)
+endif()
+
 file(GLOB BT_FILES *.bt)
-file(GLOB TXT_FILES *.txt)
-list(REMOVE_ITEM TXT_FILES ${CMAKE_CURRENT_SOURCE_DIR}/CMakeLists.txt)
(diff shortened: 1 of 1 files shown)


tools/CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
+option(INSTALL_EXAMPLES "" ON)
```
martinetd, line 1: Is it worth adding a description to the option that describes what exactly is installed?

From a very high level all the .bt files themselves also are examples, the whole purpose of bpftrace is to have hackable scripts, so it's not really clear that examples means just the script outputs here.

klemensn:
> Is it worth adding a description to the option that describes what exactly is installed?

Obviously, just forgot to fill it in eventually...

Will fix that once I can.

> From a very high level all the .bt files themselves also are examples, the whole purpose of bpftrace is to have hackable scripts, so it's not really clear that examples means just the script outputs here.

Well yes, they are all example usages for the same program, but I used the term to refer their ouput as shown in the .txt files, because these are examples... of output/results.

I'm all in for a better word if someone comes up with it.

tools/CMakeLists.txt
klemensn: > Is it worth adding a description to the option that describes what exactly is installed?

Obviously, just forgot to fill it in eventually...

Will fix that once I can.

> From a very high level all the .bt files themselves also are examples, the whole purpose of bpftrace is to have hackable scripts, so it's not really clear that examples means just the script outputs here.

Well yes, they are all example usages for the same program, but I used the term to refer their ouput as shown in the .txt files, because these are examples... of output/results.

I'm all in for a better word if someone comes up with it.

danobi: At the risk of bike shedding, how about something like `INSTALL_TOOL_DOCS` ?

klemensn: Sure, why not.
Will update in a bit.

martinetd (approved):

danobi (approved):
Thanks!

Issue #2248, opened 2022-05-30
Support IP comparison


xh4n3:
### Is your feature request related to a problem? Please describe.
One example:
```
kfunc:dev_queue_xmit {
        $skb = args->skb;
        $iph = (struct iphdr *)($skb->head + $skb->network_header);
        $sip = ntop($iph->saddr);
        // I want to filter skbs by sip here.
        printf("%-8d %14s\n", pid, $sip);
}
```
This script may output a lot skbs, IP filtering should be added here. 

Workaround 1
```
        if (strncmp($sip, "127.0.0.1", 9)) {
                return;
        }
```
This prints error message: ERROR: strncmp() only supports string arguments (inet provided)

Workaround 2
```
        if (strncmp(str($sip), "127.0.0.1", 9)) {
                return;
        }
```
This prints error message: ERROR: str() expects an integer or a pointer type as first argument (inet provided)

Workaround 3
```
        if ($iph->saddr == 0x100007f ) {
                return;
        }

```
This actually worked, but if the script isn't very human-readable.

### Describe the solution you'd like
Solution 1: Add a new built-in func like pton, so we can compare numbers
```
        if ($iph->saddr == pton(AF_INET, "127.0.0.1)) {
                return;
        }
```
Then this script can be hand over to anyone unfamiliar with hex to dig into network issues, IMHO, this would work great with the command line arguments https://github.com/iovisor/bpftrace/issues/1945 

### Describe alternative solutions or features you've considered
Somehow fixing the case in workaround 1/2, but strncmp takes a length argument, which would be hard to determine.

viktormalik:
This would be a nice feature, thanks for the suggestion!

I think that the solution using `pton` is the only one possible here as the `ntop` string is resolved asynchronously, making it impossible to be compared inside the BPF program (using e.g. `strncmp`).

xh4n3:
> This would be a nice feature, thanks for the suggestion!

May I draft a PR regarding this? 😃

viktormalik:
> May I draft a PR regarding this? smiley

Yes, please do, that would be great :-)

xh4n3:
I'm able to convert string-format IPv6 address to char array now, but I think we may need the ability to compare char array in bpftrace language, otherwise we can only compare them char by char.

What I'm having now:
```
tracepoint:tcp:tcp_retransmit_skb {
  printf("real daddr %s\n", ntop(args->daddr_v6));

  // can convert back and forth
[...]

PR #2261, opened 2022-06-09
RFC: allow uint8 array parsed as string

xh4n3:
Context: https://github.com/iovisor/bpftrace/issues/2248#issuecomment-1144512575

I'm looking for ways to compare two char array in bpftrace, some idea:
1. use `str()` convert two char array to string, and do `strncmp()` on them, it works, but seems cumbersome
2. implement char array binop comparison
3. parse char array as string, and compare them directly by `==` or `strncmp()`

This PR is a try of **the 3rd idea**, not sure if this is appropriate, please leave your comments. :)

In the following example, the `daddr_v6` is defined as `__u8 daddr_v6[16]` in the tracepoint format. 
```
tracepoint:tcp:tcp_retransmit_skb {
  $addr = args->daddr_v6;
}
```

When clang_parser parses this field, it firstly recognize this `daddr_v6[16]` field as `CXType_ConstantArray`, then its element type is `CXType_UChar`, a similar but different type from `CXType_Char_U`.

https://github.com/iovisor/bpftrace/blob/master/src/clang_parser.cpp#L333-L338

To achieve my goal with the 3rd idea, I added the `CXType_UChar` and `CXType_SChar` to the if-statement before `CreateString()`. Now the `args->daddr_v6` can be recognized as a `Type:string`.

I don't think there is much difference between `CXType_UChar` and `CXType_Char_U`, how do you think?

Some info found: https://lists.llvm.org/pipermail/cfe-dev/2012-February/019902.html

##### Side effect
This pr may cause some of the existing char array becoming strings, as the changes in the test files shown.
There are other cases may fail, e.g., c_array_indexing https://github.com/iovisor/bpftrace/blob/master/tests/runtime/regression#L26

##### Checklist

- [-] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [*] The new behaviour is covered by tests

### DIFF
--- a/tests/runtime/regression
+++ b/tests/runtime/regression
@@ -24,8 +24,8 @@ EXPECT 0
 TIMEOUT 5
 
 NAME c_array_indexing
-RUN {{BPFTRACE}} -e 'struct Foo { int a; uint8_t b[10]; } uprobe:testprogs/uprobe_test:function2 { $foo = (struct Foo *)arg0; printf("%c %c %c %c %c\n", $foo->b[0], $foo->b[1], $foo->b[2], $foo->b[3], $foo->b[4]) }' -c ./testprogs/uprobe_test
-EXPECT h e l l o
+RUN {{BPFTRACE}} -e 'struct Foo { int a; uint8_t b[10]; int c[3]; } uprobe:testprogs/uprobe_test:function2 { $foo = (struct Foo *)arg0; printf("%d %d %d\n", $foo->c[0], $foo->c[1], $foo->c[2]) }' -c ./testprogs/uprobe_test
+EXPECT 1 2 3
 TIMEOUT 5
 
 # https://github.com/iovisor/bpftrace/issues/1773
--- a/.github/workflows/embedded.yml
+++ b/.github/workflows/embedded.yml
@@ -52,6 +52,7 @@ jobs:
           DISTRO: alpine
           ALPINE_VERSION: 3.11
           VENDOR_GTEST: ON
+          GTEST_FILTER: "-clang_parser.uint8_array"
         - TYPE: Release
           NAME: alpine
           LLVM_VERSION: 9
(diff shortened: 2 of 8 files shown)


xh4n3:
Hi @viktormalik , I've include the potential side effects in addition to our prior discussion. Could you please comment on this? Thanks!

tests/runtime/regression
```diff
@@ -24,8 +24,8 @@ EXPECT 0
 TIMEOUT 5
 
 NAME c_array_indexing
-RUN {{BPFTRACE}} -e 'struct Foo { int a; uint8_t b[10]; } uprobe:testprogs/uprobe_test:function2 { $foo = (struct Foo *)arg0; printf("%c %c %c %c %c\n", $foo->b[0], $foo->b[1], $foo->b[2], $foo->b[3], $foo->b[4]) }' -c ./testprogs/uprobe_test
```
fbs, line 27: why is this test changed instead of adding a new one? This is a regression test added for a specific case so removing it seems weird

xh4n3: Because the change is at language level, this is what I meant in the side effect section.

> This pr may cause some of the existing char array becoming strings, as the changes in the test files shown.
There are other cases may fail, e.g., c_array_indexing https://github.com/iovisor/bpftrace/blob/master/tests/runtime/regression#L26

In this testcase, the `b` field in `Foo` is now parsed to string type, the original test won't pass, I use the `c` field to replace, it will verify the c array indexing functionality as well.

viktormalik:
At first, this seemed to me like a good way to go, but now that I'm looking at it more, I'm not sure if treating any byte array as a string is good. Strings get some special handling in bpftrace and so this change, e.g., introduces unnecessary operations in codegen and breaks existing regression tests.

Maybe implementing a specialized operation for comparing fixed-size arrays (MACs, IPs, etc.) would be a better approach.

xh4n3:
@viktormalik Yeah, I was worried about the breaking changes as well. Let me close this.

PR #2263, opened 2022-06-11
Add 'has_field()' and 'get_field' call.

Rtoax:
##  `get_field`: Get elements of struct/union

Syntax: `get_field(struct/union X, field)`

Get elements of struct/union, `get_field()` usually used with `has_field()`.
This interface adapts to CO-RE, This interface is adapted to `CO-RE`,
For example: since linux-5.14, the process state field of the `task_struct` structure has changed from
`state` to `__state`, and this interface can adapt to the change of this kernel structure field.

Note that when field does not exist, `get_field` will return `none`, so when __state does not exist,
`printf("%d.\n", get_field(*$task, "__state"))` will report an error, which should be avoided.

Examples:

```
if (has_field(struct task_struct, "state")) {
  $state = get_field(*$task, "state");
} else if (has_field(struct task_struct, "__state")) {
  $state = get_field(*$task, "__state");
}
```

##  `has_field`: Returns true if field exists in struct/union

Syntax: `has_field(struct/union X, field)`

Returns true if field exists in struct/union.

Examples:

```
$ sudo bpftrace -e 'struct Foo { int x; char c; } BEGIN { @x = has_field(struct Foo, "x"); exit(); }'
Attaching 1 probe...

@x: 1
```

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -2223,6 +2223,8 @@ Tracing block I/O sizes > 0 bytes
 - `kptr(void *p)` - Annotate as kernelspace pointer
 - `macaddr(char[6] addr)` - Convert MAC address data
 - `bswap(uint[8|16|32|64] n)` - Reverse byte order
+- `has_field(struct/union X, field)` - Returns true if field exists in struct/union
+- `get_field(struct/union X, field)` - Get elements of struct/union
 
 Some of these are asynchronous: the kernel queues the event, but some time later (milliseconds) it is
 processed in user-space. The asynchronous actions are: `printf()`, `time()`, and `join()`. Both `ksym()`
@@ -3136,6 +3138,44 @@ Example:
 # bpftrace -e 'BEGIN { $i = (uint32)0x12345678; printf("Reversing byte order of 0x%x ==> 0x%x\n", $i, bswap($i)); }'
 Attaching 1 probe...
 Reversing byte order of 0x12345678 ==> 0x78563412
+```
+
+## 31. `has_field`: Returns true if field exists in struct/union
+
+Syntax: `has_field(struct/union X, field)`
+
+Returns true if field exists in struct/union.
+
+Examples:
+
+```
+$ sudo bpftrace -e 'struct Foo { int x; char c; } BEGIN { @x = has_field(struct Foo, "x"); exit(); }'
+Attaching 1 probe...
+
+@x: 1
+```
+
+## 31. `get_field`: Get elements of struct/union
+
+Syntax: `get_field(struct/union X, field)`
+
+Get elements of struct/union, `get_field()` usually used with `has_field()`.
+This interface adapts to CO-RE.
+For example: since linux-5.14, the process state field of the `task_struct` structure has changed from
+`state` to `__state`, and this interface can adapt to the change of this kernel structure field.
+
+Note that when field does not exist, `get_field` will return `none`, so when __state does not exist,
+`printf("%d.\n", get_field(*$task, "__state"))` will report an error, which should be avoided.
+
+Examples:
+
+```
+$task = (struct task_struct *)curtask;
+if (has_field(struct task_struct, "state")) {
+  $state = get_field(*$task, "state");
+} else if (has_field(struct task_struct, "__state")) {
+  $state = get_field(*$task, "__state");
+}
 ```
 
 # Map Functions
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1261,6 +1261,26 @@ Functions that are marked *async* are asynchronous which can lead to unexpected 
 `bswap` reverses the order of the bytes in integer `n`. In case of 8 bit integers, `n` is returned without being modified.
 The return type is an unsigned integer of the same width as `n`.
 
+=== has_field
+
+.variants
+* `has_field(STRUCT, "x")`
+* `has_field(UNION, "x")`
+
+*compile time*
+
+Returns true if field exists in struct/union.
+
+=== get_field
+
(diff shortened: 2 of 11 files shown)


docs/reference_guide.md
```diff
@@ -3138,6 +3140,44 @@ Attaching 1 probe...
...
+
+@x: 1
+```
+
+## 31. `get_field`: Get elements of struct/union
+
+Syntax: `get_field(struct/union X, field)`
+
+Get elements of struct/union, `get_field()` usually used with `has_field()`.
+This interface adapts to CO-RE, This interface is adapted to `CO-RE`,
```
fbs, line 3163:
```suggestion
This interface adapts to CO-RE,
```

same thing twice

fbs:
I might be missing some info but what is the use of `get_field` why not use the regular struct field access `a.b` or `a->`?

Rtoax:
> I might be missing some info but what is the use of `get_field` why not use the regular struct field access `a.b` or `a->`?

The `has_field` and `get_field` used like bcc api https://github.com/iovisor/bcc/blob/master/tools/runqlat.py#L251

```
if BPF.kernel_struct_has_field(b'task_struct', b'__state') == 1:
    bpf_text = bpf_text.replace('STATE_FIELD', '__state')
else:
    bpf_text = bpf_text.replace('STATE_FIELD', 'state')
```

if bpftrace sctips like:

```
if (has_field(struct task_struct, "state")) {
    $state = $task->state;
} else if (has_field(struct task_struct, "__state")) {
    $state = $task->__state;
}
```

Whether `state` exists or `__state` exists, the other will report an error, causing the script to fail to run, so we need `get_field()`.

viktormalik:
I can see the use-case for this (portability of bpftrace scripts) but I wonder if we can come with some more compact syntax (which I believe also @fbs meant by his comment). Something like:
```
if (isfield($task->state))
    $state = $task->state; // semantic/field analyser needs to check that we're inside guarded block here
else if (...)
```

Also this feature could be helpful for AOT so maybe @danobi has some thoughts.

fbs:
ah yes why you added it makes sense now. The semantic analyser doesn't see it as a guard.

I do agree that `get_field` is a lot easier to do but for a high level language it makes sense if usual field access methods just work.

Rtoax:
indeed

danobi:
I think this would make sense for AOT. AOT could ship both branches and patch out the irrelevant branches at load time. Verifier would be OK with this too b/c of dead branch pruning (so no load errors).

> I think it's hard to solve $task->state when state field not exist.

I think I understand your point, but I think it's still doable. Just like how `sizeof(<expr>)` does not actually evaluate the expression, maybe semantic analyser can be smart enough to check for the presence of a field rather than actually access the field. Does that make sense?

Rtoax:
> I think this would make sense for AOT. AOT could ship both branches and patch out the irrelevant branches at load time. Verifier would be OK with this too b/c of dead branch pruning (so no load errors).
> 
> > I think it's hard to solve $task->state when state field not exist.
> 
> I think I understand your point, but I think it's still doable. Just like how `sizeof(<expr>)` does not actually evaluate the expression, maybe semantic analyser can be smart enough to check for the presence of a field rather than actually access the field. Does that make sense?

Yes, however, i don't find a good way to fix this yet.

danobi:
Looking this over again, it's not clear what `get_field()` is supposed to do. It looks like there are various language-level concerns in this thread. If someone wants to revive this effort, I think a short design doc would be necessary. To make sure we make this fit into the language correctly.

shunghsiyu:
I know its been awhile, but would be nice to see the `has_field()` work revived again. Would make working with issues like #3898 even simpler.

jordalgo:
@shunghsiyu Agreed. I was thinking about it a little this past week and it might be a bit of work as we have to do branch analysis and AST pruning to avoid errors in field analyser. Maybe something that would simplify things is limiting it to being inside it's own conditional e.g. this would be ok:
```
if (has_field(struct task_struct, "state")) {
    $state = $task->state;
} else if (has_field(struct task_struct, "__state")) {
    $state = $task->__state;
}
```
But this would not:
```
if (has_field(struct task_struct, "state") && $x == whatever) {
    $state = $task->state;
} else if (has_field(struct task_struct, "__state")) {
    $state = $task->__state;
}
```
cc @amscanne

amscanne:
I was also thinking in the past about `constexpr` style branch pruning. I think there are many cases where it makes sense, e.g. multi-architecture support:

```
if (arch() == "amd64") {
   ... // amd64-specific code
} else {
   ... // non-amd64-specific code
}
```

I think the complexity creeps in with long-term considerations for AOT. If we're pruning branches based on the *current system*, then any compiled script is not portable and we're throwing some more debt on the AOT pile. I'm not making a value judgement about that, but we should be intentional. (For architecture, this is less of a thing because you need to ship the program with the runtime, so this decision must be early bound.)

The general issue *is* solvable in some cases via suitable CO-RE annotations and calls, e.g. `bpf_core_field_exists`, `bpf_core_type_exists`, etc. but I don't think this can be done incrementally. In an ideal world, we would just implement these. It is possible to have two separate implementations (e.g. pruning for non-AOT cases, and eventually you support the annotations that allow these to get pruned by the verifier/JIT), but I'm not sure it's a good idea to support two different mechanisms. If we did want to do pruning for the short-term, at least a principle there would be that we should only build out this kind of `constexpr` pruning for cases supported by BPF CO-RE? This also provides a naming scheme to follow (e.g. `field_exists` and `type_exists` to line up with the standardized macros).

This could also all apply to the kernel version and kernel config that are supported by libbpf (inlined/appropriate extern symbols inserted depending on strategy).

PR #2264, opened 2022-06-12
BEGIN/END: Add more reliable implementation

danobi:
Before, we implemented BEGIN/END probes as uprobes, attached the probes
to trigger functions in /proc/self/exe, and triggered the prog by calling
the trigger functions. This works well most of the time except for when
distros decide to strip symbols from bpftrace.

This commit makes BEGIN/END probes more reliable by implementing special
probes as BPF_PROG_TYPE_RAW_TRACEPOINT and calling BPF_PROG_RUN on them.
This is ok b/c we don't allow users to access any associated `ctx` in
speical probes anyways. And we choose raw_tp prog type b/c it has
BPF_PROG_RUN implemented.

I did try to implement BPF_PROG_RUN on BPF_PROG_TYPE_KPROBE in the
kernel but upstream didn't like it too much:
https://lore.kernel.org/bpf/cover.1653861287.git.dxu@dxuuu.xyz/ .
So we have raw_tp instead.

##### Checklist


viktormalik (commented):
I really like this, handling of BEGIN/END has been a bit of pain so far.

Just one thing: there are some places where BEGIN and END are treated as special cases of uprobes. Those could be cleaned as a part of this PR.

danobi:
> Just one thing: there are some places where BEGIN and END are treated as special cases of uprobes. Those could be cleaned as a part of this PR.

Which parts are you referring to? 

I left the old codepath in b/c BPF_PROG_RUN for raw_tp was only added in late 2020: https://lore.kernel.org/bpf/20200925205432.1777-1-songliubraving@fb.com/ . So I'm concerned we'll leave some of our users hanging if we only use this new approach.

viktormalik:
> Which parts are you referring to?
> 
I found 2 occurrences that could be removed:
- https://github.com/iovisor/bpftrace/blob/master/src/ast/passes/field_analyser.cpp#L199-L201
- https://github.com/iovisor/bpftrace/blob/master/src/probe_matcher.cpp#L336

danobi:
Done

ajor (approved):
Thanks for doing this, looks much better!

viktormalik (approved):

PR #2279, opened 2022-06-24
Better helper verifier error handling

lenticularis39:
Currently, if a helper is rejected by the verifier, bpftrace displays a generic message:

```
# bpftrace -e 'kfunc:vfs_read { print(path(args->file->f_path)); }'
Attaching 1 probe...
ERROR: Error loading program: kfunc:vfs_read (try -v)
```

If run in verbose mode, the message includes the name of the unallowed helper, including its location in the compiled BPF program. This patch improves the error message to point to the location in the source core causing the generation of the helper:

```
# bpftrace -e 'kfunc:vfs_read { print(path(args->file->f_path)); }'
Attaching 1 probe...
stdin:1:18-48: ERROR: helper bpf_d_path not supported in probe
kfunc:vfs_read { print(path(args->file->f_path)); }
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

In addition to this user-visible improvement, helper call generation in `IRBuilderBPF` was refactored to go through a single `CreateHelperCall` function, which optionally takes a location argument. The location at which the first call of a helper function is generated is recorded into a map, hence it can be displayed to the user should the helper call be rejected by the verifier.

All helper creating methods were extended to pass down the location (in case of `CreatePerfEventOutput`, having no location is permitted, since that function is also used outside of parser code).

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -639,6 +639,11 @@ void AttachedProbe::cache_progfd(void)
   cached_prog_fds_[probe_.orig_name] = progfd_;
 }
 
+namespace {
+const std::regex helper_verifier_error_re(
+    "[0-9]+: \\([0-9]+\\) call ([a-z|A-Z|0-9|_]+)#([0-9]+)");
+}
+
 void AttachedProbe::load_prog(BPFfeature &feature)
 {
   if (use_cached_progfd())
@@ -770,6 +775,21 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         throw std::runtime_error(errmsg.str());
       }
     }
+
+    // try to identify an illegal helper call
+    if (std::string(log_buf.get())
+            .find("helper call is not allowed in probe") != std::string::npos)
+    {
+      // get helper function name and id
+      std::cmatch m;
+      if (std::regex_search(log_buf.get(), m, helper_verifier_error_re))
+      {
+        throw HelperVerifierError(static_cast<libbpf::bpf_func_id>(
+                                      std::stoi(m[2].str())),
+                                  m[1].str());
+      }
+    }
+
     throw std::runtime_error("Error loading program: " + probe_.name + (bt_verbose ? "" : " (try -v)"));
   }
 
--- a/src/attached_probe.h
+++ b/src/attached_probe.h
@@ -89,4 +89,18 @@ private:
   BTF &btf_;
 };
 
+class HelperVerifierError : public std::runtime_error
+{
+public:
+  const libbpf::bpf_func_id func_id_;
+  const std::string helper_name_;
+  explicit HelperVerifierError(libbpf::bpf_func_id func_id,
+                               std::string helper_name)
+      : std::runtime_error("helper invalid in probe"),
+        func_id_(func_id),
+        helper_name_(helper_name)
+  {
+  }
+};
+
 } // namespace bpftrace
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -316,6 +316,14 @@ REQUIRES_FEATURE dpath kfunc
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
+NAME path_in_unsupported_kfunc
+PROG kfunc:vfs_read { print(path(args->file->f_path)); }
+EXPECT \nstdin:1:18-48: ERROR: helper bpf_d_path not supported in probe\n
+REQUIRES_FEATURE dpath
+TIMEOUT 5
+WILL_FAIL
+AFTER ./testprogs/syscall read
+
 NAME macaddr
 RUN {{BPFTRACE}} -e 'struct MyStruct { const char* ignore; char mac[6]; }; u:./testprogs/complex_struct:func { $s = ((struct MyStruct *)arg0); printf("P: %s\n", macaddr($s->mac)); exit(); }' -c ./testprogs/complex_struct
 EXPECT P: 05:04:03:02:01:02
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to
 #### Added
 - Add builtin: `numaid`
   - [#2177](https://github.com/iovisor/bpftrace/pull/2177)
+- Add helper verifier error handling
+  - [#2279](https://github.com/iovisor/bpftrace/pull/2279)
 
 #### Changed
 - Move from BCC to libbpf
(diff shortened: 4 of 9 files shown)


viktormalik (commented):
How is this related to the `-k` option? Is it that the `-k` option will print error details if a helper fails while this PR handles situations when the program is revoked by the verifier due to an unsupported helper?

lenticularis39:
Yes, as I understand it from the comments and the source code, `-k` is used to report runtime helper errors for helpers where the user cannot easily determine whether an error happened:
```cpp
static bool return_zero_if_err(libbpf::bpf_func_id func_id)
{
  switch (func_id)
  {
    /*
     * When these function fails, bpftrace stores zero as a result.
     * A user script can check an error by seeing the value.
     * Therefore error checks of these functions are omitted if
     * helper_check_level == 1.
     */
```
`-kk` (which sets `helper_check_level` to 1) enables this for all functions, including these where the value from which the error can be detected is returned to the user.

viktormalik (approved):

src/attached_probe.h
```diff
@@ -84,4 +84,16 @@ class AttachedProbe
   std::function<void()> usdt_destructor_;
 };
 
+class HelperVerifierError
```
danobi, line 87: Should inherit from `std::runtime_error` here. So catching by `std::exception &` will work

src/attached_probe.cpp
```diff
@@ -804,6 +804,22 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         throw std::runtime_error(errmsg.str());
       }
     }
+
+    // try to identify an illegal helper call
+    if (std::string(log_buf.get())
+            .find("helper call is not allowed in probe") != std::string::npos)
+    {
+      // get helper function name and id
+      std::regex re("[0-9]+: \\([0-9]+\\) call ([a-z|A-Z|0-9|_]+)#([0-9]+)");
```
danobi, line 813: Maybe pull this into a global in an anon namespace? So we don't reconstruct the regex every time this code path is hit. Making this static also works but IIRC the compiler has to generate some locks to synchronize construction between multiple threads

tests/runtime/call
```diff
@@ -316,6 +316,14 @@ REQUIRES_FEATURE dpath
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
+NAME path_invalid_helper
+PROG kfunc:vfs_read { print(path(args->file->f_path)); }
+EXPECT \nstdin:1:18-48: ERROR: helper bpf_d_path not supported in probe\n
+REQUIRES_FEATURE dpath
```
danobi, line 322: I'm confused about this test. If `REQUIRES_FEATURE dpath` condition is met, then it means dpath is supported and loads correctly (see `BPFfeature::has_d_path()`). I suspect the CI is passing b/c the condition is not met so the test is noop. But on a machine with dpath, this test will fail, right?

danobi (commented):
nicely done; few small comments

tests/runtime/call
lenticularis39: The program is rejected by verifier because it uses the `path` helper in a kfunc not on the list of functions where `bpf_d_path` is allowed (see `btf_allowlist_d_path` in `kernel/trace/bpf_trace.c`). On systems without `d_path` support, the test is not run, because `d_path` always fails there, hence it's pointless.

danobi: Ah, I see. I thought you were testing that path() was not supported on the machine at all.

Mind changing the name of the test case to something like "use_path_in_unsupported_function" ?

lenticularis39: I renamed it to `path_in_unsupported_kfunc`.

danobi (approved):

rmilkowski:
I'm getting a similar issue:

# bpftrace -e 'kfunc:vfs_open {printf("%s\n", path(args->path));}'
Attaching 1 probe...
ERROR: Error loading program: kfunc:vfs_open (try -v)

# bpftrace -e 'kfunc:vfs_open {printf("%s\n", path(args->path));}' -v
INFO: node count: 10
Attaching 1 probe...

Error log:
0: (bf) r6 = r1
1: (b7) r1 = 0
2: (7b) *(u64 *)(r10 -136) = r1
last_idx 2 first_idx 0
3: (7b) *(u64 *)(r10 -8) = r1
4: (7b) *(u64 *)(r10 -16) = r1
5: (7b) *(u64 *)(r10 -24) = r1
6: (7b) *(u64 *)(r10 -32) = r1
7: (7b) *(u64 *)(r10 -40) = r1
8: (7b) *(u64 *)(r10 -48) = r1
9: (7b) *(u64 *)(r10 -56) = r1
10: (7b) *(u64 *)(r10 -64) = r1
11: (79) r1 = *(u64 *)(r6 +0)
func 'vfs_open' arg0 has btf_id 627 type STRUCT 'path'
12: (bf) r2 = r10
13: (07) r2 += -64
14: (b7) r3 = 64
15: (85) call bpf_d_path#147
helper call is not allowed in probe
processed 16 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

ERROR: Error loading program: kfunc:vfs_open


Is it supposed to work?

viktormalik:
No, this is not supposed to work. The `path` helper is not available for the `kfunc:vfs_open` probe. What this PR did is that it added a better error message for the user.

If you ran the program with bpftrace containing commits from this PR (should be part of version 0.16.0), you should see something like this:
```
# bpftrace -e 'kfunc:vfs_open {printf("%s\n", path(args->path));}'
Attaching 1 probe...
stdin:1:32-48: ERROR: helper bpf_d_path not supported in probe
kfunc:vfs_open {printf("%s\n", path(args->path));}
                               ~~~~~~~~~~~~~~~~
```

rmilkowski:
Right... thanks for super quick reply.

I see:

kernel/trace/bpf_trace.c:
...
BTF_SET_START(btf_allowlist_d_path)
#ifdef CONFIG_SECURITY
BTF_ID(func, security_file_permission)
BTF_ID(func, security_inode_getattr)
BTF_ID(func, security_file_open)
#endif
#ifdef CONFIG_SECURITY_PATH
BTF_ID(func, security_path_truncate)
#endif
BTF_ID(func, vfs_truncate)
BTF_ID(func, vfs_fallocate)
BTF_ID(func, dentry_open)
BTF_ID(func, vfs_getattr)
BTF_ID(func, filp_close)
BTF_SET_END(btf_allowlist_d_path)
...

Given that flip_close and vfs_open both provide "struct file *", is there a reason why vfs_open() is not on the allowed list?

viktormalik:
In this case, I'd say that it's missing from the list b/c there's already `dentry_open` which calls `vfs_open`.

PR #2286, opened 2022-07-02
Remove useless ELF_GETSHDRSTRNDX test

Rtoax:


danobi (commented):
Wrong file? Maybe you meant cmake/FindLibElf.cmake?

Also I wonder why we're even testing for `elf_getshdrstrndx`. Doesn't look like any parts of the build use it. Maybe some other cmake module transitively requires it...

fbs:
I think the whole file is copied from bcc

viktormalik:
I don't think that we use it. It'd be better to remove the test completely.

Rtoax:
> I don't think that we use it. It'd be better to remove the test completely.

OK

cmake/FindLibElf.cmake
```diff
@@ -43,15 +43,4 @@ FIND_PACKAGE_HANDLE_STANDARD_ARGS(LibElf "Please install the libelf development
...
-CHECK_CXX_SOURCE_COMPILES("#include <libelf.h>
-int main() {
-  Elf *e = (Elf*)0;
-  size_t sz;
-  elf_getshdrstrndx(e, &sz);
-  return 0;
-}" ELF_GETSHDRSTRNDX)
-SET(CMAKE_REQUIRED_LIBRARIES)
-
 mark_as_advanced(LIBELF_INCLUDE_DIRS LIBELF_LIBRARIES ELF_GETSHDRSTRNDX)
```
viktormalik, line 46: Should be removed here, too.

danobi (approved):

PR #2289, opened 2022-07-05
add pton builtin

xh4n3:
##### Description 

This pull request addressed the proposal in https://github.com/iovisor/bpftrace/issues/2248, added a new built-in function `pton`.

The `pton` converts text-formatted IP addresses to byte arrays, which is exactly reversed-version of `ntop`. It should come in handy when we need to select packets with certain IP addresses.

##### Plan

The current implementation only support comparison byte by byte, direct byte array comparison will be implemented in future pull requests. 

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -2212,6 +2212,7 @@ Tracing block I/O sizes > 0 bytes
 - `kstack([StackMode mode, ][int level])` - Kernel stack trace
 - `ustack([StackMode mode, ][int level])` - User stack trace
 - `ntop([int af, ]int|char[4|16] addr)` - Convert IP address data to text
+- `pton(const string *addr)` - Convert text IP address to byte array
 - `cat(char *filename)` - Print file content
 - `signal(char[] signal | u32 signal)` - Send a signal to the current task
 - `strncmp(char *s1, char *s2, int length)` - Compare first n characters of two strings
@@ -3179,6 +3180,36 @@ dropped privs to tcpdump
 10:23:44.674087 IP 22.128.74.231.63175 > 192.168.0.23.22: Flags [.], ack 3513221061, win 14009, options [nop,nop,TS val 721277750 ecr 3115333619], length 0
 10:23:45.823194 IP 100.101.2.146.53 > 192.168.0.23.46619: 17273 0/1/0 (130)
 10:23:45.823229 IP 100.101.2.146.53 > 192.168.0.23.46158: 45799 1/0/0 A 100.100.45.106 (60)
+```
+
+## 32. `pton()`: Convert text IP address to byte array
+
+Syntax: `pton(const string *addr)`
+
+This converts a text representation of an IPv4 or IPv6 address to byte array.
+`pton` infers the address family based on `.` or `:` in the given argument.
+`pton` comes in handy when we need to select packets with certain IP addresses.
+
+Examples:
+
+```
+# bpftrace -e 'tracepoint:tcp:tcp_retransmit_skb {
+  if (args->daddr_v6[0] == pton("::1")[0]) {
+    printf("first octet matched\n");
+  }
+}'
+Attaching 1 probe...
+first octet matched
+^C
+
+# bpftrace -e 'tracepoint:tcp:tcp_retransmit_skb {
+  if (args->daddr[0] == pton("127.0.0.1")[0]) {
+    printf("first octet matched\n");
+  }
+}'
+Attaching 1 probe...
+first octet matched
+^C
 ```
 
 # Map Functions
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1461,6 +1461,18 @@ SRC 18:C0:4D:08:2E:BB, DST 74:83:C2:7F:8C:FF
 If an integer or `char[4]` is given, ntop assumes IPv4, if a `char[16]` is given, ntop assumes IPv6.
 You can also pass the address type (e.g. AF_INET) explicitly as the first parameter.
 
+=== pton
+
+.variants
+* `char addr[4] pton(const string *addr_v4)`
+* `char addr[16] pton(const string *addr_v6)`
+
+*compile time*
+
+`pton` converts a text representation of an IPv4 or IPv6 address to byte array.
+`pton` infers the address family based on `.` or `:` in the given argument.
+`pton` comes in handy when we need to select packets with certain IP addresses.
+
 === override
 
 .variants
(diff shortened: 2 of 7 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -817,6 +818,43 @@ void SemanticAnalyser::visit(Call &call)
...
+    if (call.vargs->size() == 2)
+    {
+      check_arg(call, Type::integer, 0);
+      auto af_type_val = bpftrace_.get_int_literal(arg);
+      if (af_type_val.has_value())
+      {
+        af_type = *af_type_val;
+      }
+      arg = call.vargs->at(1);
+    }
```
danobi, line 837: Is it possible to infer ipv4/ipv6 from the string instead of the optional param?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -817,6 +818,43 @@ void SemanticAnalyser::visit(Call &call)
...
+
+    if (!arg->type.IsStringTy())
+    {
+      LOG(ERROR, call.loc, err_)
+          << call.func << "() expects an string argument, got "
+          << arg->type.type;
+      return;
+    }
```
danobi, line 845: I think we need to check if the addr is a string literal, not just a string.

tests/semantic_analyser.cpp
```diff
@@ -698,6 +702,31 @@ TEST(semantic_analyser, call_ntop)
   test(structs + "kprobe:f { ntop(((struct inet*)0)->invalid); }", 1);
 }
 
+TEST(semantic_analyser, call_pton)
```
danobi, line 705: Please add some tests that fail. For example invalid arg types. And non-literal strings

docs/reference_guide.md
```diff
@@ -3181,6 +3182,37 @@ dropped privs to tcpdump
...
+Syntax: `pton([int af, ]char* addr)`
+
+This converts a text representation of an IPv4 or IPv6 address to byte array. 
+`pton` will parse `addr` as address type given by `af`. If `af` is not specified, `pton` assumes `addr` as IPv4 type.
+`pton` comes in handy when we need to select packets with certain IP addresses.
+
+Examples:
+
+```
+# bpftrace --include /usr/include/arpa/inet.h \
```
danobi, line 3196: Ideally we do not require import header files at all. BTF should give us nearly everything we need. In this case I think we can omit the `AF_INET[4|6]` optional params by inferring AF from the string literal.

man/adoc/bpftrace.adoc
```diff
@@ -1461,6 +1461,17 @@ SRC 18:C0:4D:08:2E:BB, DST 74:83:C2:7F:8C:FF
 If an integer or `char[4]` is given, ntop assumes IPv4, if a `char[16]` is given, ntop assumes IPv6.
 You can also pass the address type (e.g. AF_INET) explicitly as the first parameter.
 
+=== pton
+
```
fbs, line 1465:
```suggestion
*compile time*
```

I think it needs this as its strictly compile time only

man/adoc/bpftrace.adoc
```diff
@@ -1461,6 +1461,17 @@ SRC 18:C0:4D:08:2E:BB, DST 74:83:C2:7F:8C:FF
...
+.variants
+* `char addr[4] pton(char *addr)`
```
fbs, line 1467:
```suggestion
* `char addr[4] pton(const string *addr)`
```

All the others that only except static strings use the same

src/ast/passes/codegen_llvm.cpp
```diff
@@ -790,6 +790,59 @@ void CodegenLLVM::visit(Call &call)
...
+    }
+
+    auto addr_str = bpftrace_.get_string_literal(addr);
+    char dst[addr_size];
+    Value *octet;
+
+    auto ret = inet_pton(af_type, addr_str.c_str(), &dst);
+    if (ret != 1)
+    {
+      LOG(FATAL) << "inet_pton() call returns " << ret;
```
viktormalik, line 832: This should be a standard `ERROR` as it may occur quite normally when the user supplies an invalid address string. Ideally, it should be detected in the semantic analyser and reported along with the call location.

xh4n3:
@danobi @fbs @viktormalik Thanks for the suggestions!
I've updated the commits to address those issues. 😃

src/ast/passes/semantic_analyser.cpp
```diff
@@ -817,6 +818,31 @@ void SemanticAnalyser::visit(Call &call)
 
     call.type = CreateInet(buffer_size);
   }
+  else if (call.func == "pton")
+  {
+    if (!check_nargs(call, 1))
+      return;
+    std::string addr = bpftrace_.get_string_literal(call.vargs->at(0));
```
danobi, line 825: This might be an empty string. Is the current error (one that `get_string_literal()` may print) good enough or can we give a nicer one?

xh4n3: Thanks @danobi . If `addr` is an empty string, this would be catched by the `else` branch at line 840, which prints `ERROR: pton() expects an string argument of an IPv4/IPv6 address, got`. Should I add another check above?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -790,6 +790,46 @@ void CodegenLLVM::visit(Call &call)
...
+    {
+      buf = b_.CreateAllocaBPF(array_t, "addr4");
+    }
+
+    char dst[addr_size];
+    Value *octet;
+    auto ret = inet_pton(af_type, addr.c_str(), &dst);
+    if (ret != 1)
+    {
+      LOG(ERROR) << "inet_pton() call returns " << ret;
```
viktormalik, line 820: I think that it's too late to detect this error here. It currently works like this:
```
# bpftrace -e 'BEGIN { printf("%d\n", pton("1920.168.0.1")[0]); }'
ERROR: inet_pton() call returns 0
Attaching 1 probe...
16
```
which I don't think is a correct behaviour.

If the user enters an invalid address, he should be properly notified what (and where) is wrong without continuing to run the program. I'd suggest to run `inet_pton` in semantic analyser, too (it should be cheap, so no problem running it twice) and detect this failure there.

viktormalik: And now this can (and should) be a `FATAL` error.

tests/semantic_analyser.cpp
viktormalik: And also some tests for invalid addresses.

xh4n3:
@viktormalik @danobi The inet_pton check has been added to semantic analyser, it should find empty string or illegal address in the first place now.

src/ast/passes/semantic_analyser.cpp
danobi: Ah, I see. No, this is fine

danobi (approved):

viktormalik (approved):

PR #2295, opened 2022-07-12
Replace commas in BPF program name

### LINKED ISSUE #2294: BEGIN,END { ... } does not work
This appears to be a regression introduced in #2265

### What reproduces the bug?
With bpftrace built from fc940d31be140d54ceb4c7ae47371785f0b26b65 this script fails:
```
# ./src/bpftrace -e 'BEGIN,END { printf("hi\n") }' 
Attaching 2 probes...
ERROR: Error loading program: BEGIN,END (try -v)
```

Go back one commit to 746a53c35afafb5a782b49cd16692ca6fb0b6618 and it works as expected:
```
# ./src/bpftrace -e 'BEGIN,END { printf("hi\n") }' 
Attaching 2 probes...
hi
^Chi


```

viktormalik:
`bpf_prog_load` does not accept commas in the program name, which may cause failures when attaching, e.g. to `BEGIN,END`. This fixes the issue by replacing commas by dots, which are accepted.

Fixes #2294.

##### Checklist

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -256,6 +256,11 @@ PROG interval:s:1 { exit(); } END { printf("end"); }
 EXPECT end
 TIMEOUT 2
 
+NAME BEGIN,END
+PROG BEGIN,END { printf("Hello\n"); exit(); }
+EXPECT Hello\nHello
+TIMEOUT 2
+
 # The "probe" builtin forces bpftrace to generate one BPF function for each
 # match and so we should fail on exceeding BPFTRACE_MAX_BPF_PROGS.
 NAME bpf_programs_limit
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -680,8 +680,9 @@ void AttachedProbe::load_prog(BPFfeature &feature)
     // start the name after the probe type, after ':'
     if (auto last_colon = name.rfind(':'); last_colon != std::string::npos)
       name = name.substr(last_colon + 1);
-    // replace '+' by '.'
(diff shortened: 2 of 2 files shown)


viktormalik:
> a small regression test for it would be nice, can see this special probe combination being an issue in the future

Done.

tests/runtime/probe
```diff
@@ -256,6 +256,11 @@ PROG interval:s:1 { exit(); } END { printf("end"); }
 EXPECT end
 TIMEOUT 2
 
+NAME BEGIN,END
+PROG BEGIN,END { printf("Hello\n"); exit(); }
+EXPECT Hello
```
ajor, line 261: Should there be two "Hello"s here? Can this test framework handle that?

viktormalik: Good catch, it can, I updated the test.

tests/runtime/probe
```diff
@@ -256,6 +256,11 @@ PROG interval:s:1 { exit(); } END { printf("end"); }
 EXPECT end
 TIMEOUT 2
 
+NAME BEGIN,END
+PROG BEGIN,END { printf("Hello\n"); exit(); }
```
fbs, line 260:
```suggestion
PROG BEGIN,END { print(probe); exit(); }
```

maybe better? That way youre sure you didnt trigger BEGIN twice or swapped ordering

fbs: ah good point :)

dxuuu (approved):

danobi (approved):

Issue #2297, opened 2022-07-12
Detect and fix invalid LLVM IR generated by bpftrace


lenticularis39:
When running bpftrace on LLVM 14, an issue related to bpftrace generating invalid LLVM IR were encountered (see #2228, #2222 and #2296). Furthermore, other failing tests in LLVM 14 also generate invalid LLVM IR, implying this could be a more widespread issue in the codegen.

As suggested by @fbs in #2222, a verifier pass detecting LLVM IR in bpftrace output could be added to all tests to ensure these issues are identified and fixed both in the present code and in the future.

lenticularis39:
Thank you for the suggestion @BurntBrunch, I implemented a solution in the linked PR by calling `VerifierPass` directly and doing output redirection, using `verifyModule` is a simpler solution.

PR #2298, opened 2022-07-15
Add option to verify LLVM IR generated by bpftrace

lenticularis39:
As proposed in #2297, this PR implements an option to verify generated LLVM IR via setting the `BPFTRACE_VERIFY_LLVM_IR` environemntal variable to `1`. An environmental variable is used to easily allow this to be set in tests.

~It also enables this option for runtime tests, detecting invalid code that has to be fixed before merging this PR to make the tests pass.~

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -243,7 +243,7 @@ static std::optional<struct timespec> get_boottime()
   return ret;
 }
 
-[[nodiscard]] static bool parse_env(BPFtrace& bpftrace)
+[[nodiscard]] static bool parse_env(BPFtrace& bpftrace, bool& verify_llvm_ir)
 {
   if (!get_uint64_env_var("BPFTRACE_STRLEN", bpftrace.strlen_))
     return false;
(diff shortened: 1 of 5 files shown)


fbs:
nice, lots of issues we still have to fix :D 

I suggest we merge it but dont enable it yet, that way we can start fixing them

src/main.cpp
```diff
@@ -125,6 +125,7 @@ void usage()
   std::cerr << "    BPFTRACE_PERF_RB_PAGES      [default: 64] pages per CPU to allocate for ring buffer" << std::endl;
   std::cerr << "    BPFTRACE_NO_USER_SYMBOLS    [default: 0] disable user symbol resolution" << std::endl;
   std::cerr << "    BPFTRACE_CACHE_USER_SYMBOLS [default: auto] enable user symbol cache" << std::endl;
+  std::cerr << "    BPFTRACE_VERIFY_LLVM_IR     [default: 0] verify generated LLVM IR" << std::endl;
```
danobi, line 128: Not exactly a user facing env var IMO. Maybe document in test docs somewhere instead?

lenticularis39: Good point, I removed it from the help.

lenticularis39:
I removed the commit that enables the LLVM IR verification in CI, so that it can be merged without breaking the CI. This also means there is no documentation for the option, since it is not user facing and is not used by the tests yet.

lenticularis39:
Simplified `CodegenLLVM::verify` to make use of `llvm::verifyModule` instead of capturing the output of `VerifierPass`.

viktormalik (approved):

viktormalik:
@fbs @danobi are you fine with merging this?

fbs (approved):

Issue #2300, opened 2022-07-22
Add support for bpf_trace_printk


xh4n3:
### Is your feature request related to a problem? Please describe.
I need `bpf_trace_printk` while developing bpftrace features, and I don't have enough confidence about the `printf` output because I may have broken these codes.

### Describe the solution you'd like
Add a built-in function `debugf`, which calls `bpf_trace_printk`. Compared with current `printf`, it has following benefits:
* The `bpf_trace_printk` is helpful while debugging bpftrace itself, e.g. the `printf` function.
* Since `debugf` doesn't rely on the async mechanism, it would help if we're debugging with async events.
* For bpftrace script developers, `debugf` prints to /sys/kernel/debug/tracing/trace, the output content won't get mixed with stdout to confuse script users.

### Describe alternative solutions or features you've considered
No.

I'm willing to submit a PR about this. Please leave comments if any, thanks!

viktormalik:
I'm not against adding a new builtin function, but we should be cautious about this one. Looking at the helper description from the [bpf-helpers manpage](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html):

>  Also, note that bpf_trace_printk() is slow, and
should only be used for debugging purposes. For
this reason, a notice block (spanning several
lines) is printed to kernel logs and states that
the helper should not be used "for production use"

So, our reference guide should strongly advise not to use this inside shipped scripts, which may be used in production.

Still, this can be useful as a replacement for `printf` when its async nature is a problem. So, feel free to proceed with the PR, but maybe an actual example where this is better than `printf` would be nice.

Thanks!

xh4n3:
Thanks, I will add the warning part in the docs, and think about what scenarios for this to be suitable.

danobi:
Sounds useful for debugging. Maybe we can leave it undocumented and print a loud warning any time it's used?

PR #2307, opened 2022-07-28
add debugf builtin

xh4n3:
This PR implements the builtin function `debugf`, which we've discussed earlier in https://github.com/iovisor/bpftrace/issues/2300

This PR is currently in RFC status, comments are welcomed. At least we should decide whether documentation is needed.

##### Example
```bash
[root@test build]# bpftrace -e 'i:s:1 { debugf("hello"); exit();}'
stdin:1:9-24: WARNING: The debugf() builtin is not recommended for production use. For more information see bpf_trace_printk in bpf-helpers(7).
i:s:1 { debugf("hello"); exit();}
        ~~~~~~~~~~~~~~~
Attaching 1 probe...


[root@test build]# cat /sys/kernel/debug/tracing/trace
          <idle>-0       [000] d.h. 880121.845861: bpf_trace_printk: hello
```

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1042,6 +1042,28 @@ void IRBuilderBPF::CreatePerfEventOutput(Value *ctx,
                    loc);
 }
 
+void IRBuilderBPF::CreateTracePrintk(Value *fmt_ptr,
+                                     Value *fmt_size,
+                                     const std::vector<Value *> &values,
+                                     const location &loc)
+{
+  std::vector<Value *> args = { fmt_ptr, fmt_size };
+  for (auto val : values)
+  {
+    args.push_back(val);
+  }
+
+  // long bpf_trace_printk(const char *fmt, u32 fmt_size, ...)
+  FunctionType *traceprintk_func_type = FunctionType::get(
+      getInt64Ty(), { getInt8PtrTy(), getInt32Ty() }, true);
+
+  CreateHelperCall(libbpf::BPF_FUNC_trace_printk,
+                   traceprintk_func_type,
+                   args,
+                   "trace_printk",
+                   &loc);
+}
+
 void IRBuilderBPF::CreateSignal(Value *ctx, Value *sig, const location &loc)
 {
   // long bpf_send_signal(u32 sig)
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -955,7 +955,8 @@ void SemanticAnalyser::visit(Call &call)
     }
     call.type = CreateUInt64();
   }
-  else if (call.func == "printf" || call.func == "system" || call.func == "cat")
+  else if (call.func == "printf" || call.func == "system" ||
+           call.func == "cat" || call.func == "debugf")
   {
     check_assignment(call, false, false, false);
     if (check_varargs(call, 1, 128))
@@ -989,12 +990,18 @@ void SemanticAnalyser::visit(Call &call)
                   },
           });
         }
-        std::string msg = validate_format_string(fmt.str, args);
+        std::string msg = validate_format_string(fmt.str, args, call.func);
         if (msg != "")
         {
           LOG(ERROR, call.loc, err_) << msg;
         }
       }
+    }
+    if (call.func == "debugf" && is_final_pass())
+    {
+      LOG(WARNING, call.loc, out_)
+          << "The debugf() builtin is not recommended for production use. For "
+             "more information see bpf_trace_printk in bpf-helpers(7).";
     }
 
     call.type = CreateNone();
--- a/src/format_string.cpp
+++ b/src/format_string.cpp
@@ -5,9 +5,12 @@
 namespace bpftrace {
 
 const int FMT_BUF_SZ = 512;
+// bpf_trace_printk cannot use more than three arguments, see bpf-helpers(7).
+const int PRINTK_MAX_ARGS = 3;
 
 std::string validate_format_string(const std::string &fmt,
-                                   std::vector<Field> args)
+                                   std::vector<Field> args,
+                                   const std::string call_func)
 {
   std::stringstream message;
 
@@ -20,14 +23,22 @@ std::string validate_format_string(const std::string &fmt,
   int num_args = args.size();
   if (num_args < num_tokens)
   {
-    message << "printf: Not enough arguments for format string (" << num_args
-            << " supplied, " << num_tokens << " expected)" << std::endl;
+    message << call_func << ": Not enough arguments for format string ("
+            << num_args << " supplied, " << num_tokens << " expected)"
+            << std::endl;
     return message.str();
   }
   if (num_args > num_tokens)
   {
-    message << "printf: Too many arguments for format string (" << num_args
-            << " supplied, " << num_tokens << " expected)" << std::endl;
+    message << call_func << ": Too many arguments for format string ("
+            << num_args << " supplied, " << num_tokens << " expected)"
+            << std::endl;
+    return message.str();
+  }
+  if (call_func == "debugf" && num_args > PRINTK_MAX_ARGS)
+  {
+    message << call_func << ": Cannot use more than " << PRINTK_MAX_ARGS
+            << " conversion specifiers" << std::endl;
     return message.str();
   }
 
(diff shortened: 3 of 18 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -949,13 +950,19 @@ void SemanticAnalyser::visit(Call &call)
...
         {
           LOG(ERROR, call.loc, err_) << msg;
         }
       }
     }
+    if (call.func == "debugf" && is_final_pass())
+    {
+      LOG(WARNING, call.loc, out_)
+          << "The debugf() is highly disrecommended for production use, check "
+             "bpf_trace_printk in bpf-helpers(7).";
```
ajor, line 964:
```suggestion
          << "The debugf() builtin is not recommended for production use. For more information see "
             "bpf_trace_printk in bpf-helpers(7).";
```
Some re-wording that I think reads better (I don't think "disrecommend" is a word)

src/format_string.cpp
```diff
@@ -5,9 +5,11 @@
 namespace bpftrace {
 
 const int FMT_BUF_SZ = 512;
+const int PRINTK_MAX_ARGS = 3;
```
ajor, line 8: Where does this number come from?

src/required_resources.cpp
```diff
@@ -142,6 +133,41 @@ int RequiredResources::create_maps_impl(BPFtrace &bpftrace, bool fake)
   return failed_maps;
 }
 
+template <typename T>
+std::unique_ptr<T> RequiredResources::prepareFormatStringDataMap(
+    std::vector<std::tuple<FormatString, std::vector<Field>>> args,
```
ajor, line 138:
```suggestion
    const std::vector<std::tuple<FormatString, std::vector<Field>>> &args,
```

src/required_resources.cpp
```diff
@@ -93,36 +93,27 @@ int RequiredResources::create_maps_impl(BPFtrace &bpftrace, bool fake)
...
-      idx += len + 1;
+      failed_maps += 1;
     }
-
-    // store the data in map
-    uint64_t id = 0;
-    int ret = bpf_update_elem(map->mapfd_, &id, formats.data(), 0);
-
-    if (is_invalid_map(map->mapfd_) || (!fake && ret == -1))
+    else
```
ajor, line 101: I'm not familiar enough with this code to know why it's like this myself, but why doesn't this new code follow the style in the other branches of always calling `bpftrace.maps.Set()`?

If this way is better then it'd be best to change them all - the mismatch in styles adds extra confusion (to me at least!)

src/required_resources.cpp
```diff
@@ -142,6 +133,41 @@ int RequiredResources::create_maps_impl(BPFtrace &bpftrace, bool fake)
   return failed_maps;
 }
 
+template <typename T>
+std::unique_ptr<T> RequiredResources::prepareFormatStringDataMap(
+    std::vector<std::tuple<FormatString, std::vector<Field>>> args,
+    bool fake)
```
ajor, line 139: I think it'd be nicer to take the `is_invalid_map` lambda as an argument instead of repeating the logic using the `fake` bool

src/ast/irbuilderbpf.cpp
```diff
@@ -1042,6 +1042,28 @@ void IRBuilderBPF::CreatePerfEventOutput(Value *ctx,
                    loc);
 }
 
+void IRBuilderBPF::CreateTracePrintk(Value *fmt_ptr,
+                                     Value *fmt_size,
+                                     std::vector<Value *> values,
```
ajor, line 1047:
```suggestion
                                     const std::vector<Value *> &values,
```

src/printf_format_types.h
```diff
@@ -73,6 +73,22 @@ const std::unordered_map<std::string, Type> printf_format_types = {
   {"tp", Type::integer}
 };
 
+// trace_printk_format_types is a subset of printf_format_types that contains valid types for bpf_trace_printk()
+// see iovisor/bcc BTypeVisitor::checkFormatSpecifiers
+const std::unordered_map<std::string, Type> trace_printk_format_types = {
```
ajor, line 78: No %i ?

tests/semantic_analyser.cpp
```diff
@@ -1156,6 +1156,33 @@ TEST(semantic_analyser, printf)
   }
 }
 
+TEST(semantic_analyser, debugf)
```
ajor, line 1159: Is there a way we can test that a warning is displayed when using this builtin?

ajor (commented):
A few suggestions and comments, but it looks good to me overall. (A review from someone else would be appreciated though, given the length of time since I last looked at most of this)

xh4n3:
Hi @ajor , thank you for reviewing this!
I've replied to some of your comments, others should be fixed now.

ajor:
I actually liked the factoring out of `prepareFormatStringDataMap()` you had in your previous version! It was nice to not have that section of duplicated code. Any reason you removed it?

xh4n3:
> I actually liked the factoring out of `prepareFormatStringDataMap()` you had in your previous version! It was nice to not have that section of duplicated code. Any reason you removed it?

Hi @ajor , I thought I was always calling always calling `bpftrace.maps.Set()`, and misunderstood your comment on that 😢 

Luckily I still have the `prepareFormatStringDataMap()` codes.

> I think it'd be nicer to take the is_invalid_map lambda as an argument instead of repeating the logic using the fake bool

I think the `fake` argument is still needed for the if-condition even if we have `is_invalid_map()` lambda as argument, so I moved this if-condition out. I think it'd look nicer.

src/required_resources.cpp
```diff
@@ -93,37 +93,20 @@ int RequiredResources::create_maps_impl(BPFtrace &bpftrace, bool fake)
...
-    int ret = bpf_update_elem(map->mapfd_, &id, formats.data(), 0);
-
+    int ret;
+    auto map = prepareFormatStringDataMap<T>(seq_printf_args, &ret);
     if (is_invalid_map(map->mapfd_) || (!fake && ret == -1))
       failed_maps += 1;
-
     bpftrace.maps.Set(MapManager::Type::SeqPrintfData, std::move(map));
   }
+  if (needs_data_map)
```
viktormalik, line 102: The same `if` is above, it'd make more sense to move the code to that block.

viktormalik (commented):
Couple of nits, plus I have one question/suggestion: wouldn't it make sense to merge `debugfs_ids` with `seq_printf_ids`? Most of the code using them is exactly the same so you'd completely avoid duplication.

xh4n3:
>  wouldn't it make sense to merge `debugfs_ids` with `seq_printf_ids`? Most of the code using them is exactly the same so you'd completely avoid duplication.

Thanks @viktormalik 

IMHO, these ids are used to reference format arguments in the data maps, and there are different kind of maps for arguments. It is like we have a one-to-one correspondence relation: id <-> data_map <-> arguments.

As there are different kinds of arguments https://github.com/iovisor/bpftrace/blob/master/src/required_resources.h#L78-L85, should we combine these two ids?

Correct me if I'm wrong 😃

tests/runtime/call
```diff
@@ -412,3 +412,15 @@ EXPECT OK
 REQUIRES_FEATURE skboutput
 TIMEOUT 5
 MIN_KERNEL 5.5
+
+NAME debugf
+REQUIRES echo > /sys/kernel/debug/tracing/trace
```
viktormalik, line 417: `REQUIRES` should be used to check for existence of resources, use `BEFORE` to prepare a test environment.

viktormalik:
> IMHO, these ids are used to reference format arguments in the data maps, and there are different kind of maps for arguments. It is like we have a one-to-one correspondence relation: id <-> data_map <-> arguments.
> 
> As there are different kinds of arguments https://github.com/iovisor/bpftrace/blob/master/src/required_resources.h#L78-L85, should we combine these two ids?

Well,  I meant ofc to use both `seq_printf_args`and `seq_printf_ids` since they are used exactly in the same way as `debugfs_args`/`_ids`, the only difference is in the builtin/helper name. If we support another helper in future which is processed the same way, we'll have to copy portions of code again.

tests/runtime/call
```diff
@@ -412,3 +412,21 @@ EXPECT OK
 REQUIRES_FEATURE skboutput
 TIMEOUT 5
 MIN_KERNEL 5.5
+
+NAME debugf
+BEFORE ./testprogs/empty_trace_buffer
```
viktormalik, line 417: It's not very practical to write a separate program for this when you can just use `echo > /sys/kernel/debug/tracing/trace`. I know I suggested to use the `BEFORE` directive, but since it doesn't allow running programs with arguments, the original version (running the `echo` command as a part of `RUN`) seems much better to me.

viktormalik (approved):

danobi:
@xh4n3 can you take a look at the CI failures on master? Looks like legit errors

xh4n3:
> @xh4n3 can you take a look at the CI failures on master? Looks like legit errors

Thanks, fixed in https://github.com/iovisor/bpftrace/pull/2341

PR #2315, opened 2022-08-03
BTF for kernel modules

viktormalik:
This introduces support for BTF information for kernel modules.

BTF for kernel modules is nowadays shipped by many distros (it is exposed via `/sys/kernel/btf/<module-name>`). Using it enables BTF-depending features for modules, in particular:
- k(ret)func probes for module functions including access to function arguments by name
- automatic type resolution of types defined in modules
- listing of types defined in modules

This is quite tricky to test, though, as it either requires to insmod a custom module before running the tests or to rely on some common module that should be available on most systems. In the end, I went with the second way - the runtime tests hook into the `nft_trans_alloc_gfp` function from the `nf_tables` module which is pretty common and is invoked by the `nft` utility. The module is now also loaded in CI to make sure that the tests are executed there.

This PR also features one important change - FieldAnalyser now fully resolves structure types from BTF/DWARF (including struct fields) and Clang parser is only called if there are some user-defined types/includes or if FieldAnalyser was not able to resolve something (for instance, it doesn't support bitfields at the moment). This should bring some execution speedup and is also crucial for type resolution from kernel modules since BTF dumping doesn't work properly for them.

In addition, there are several minor changes done:
- BTF parsing is refactored to use as many functions from libbpf as possible
- runtime tests now allow multiple `REQUIRES` fields (all must be satisfied for the test to run)
- unit tests for `FieldAnalyser` were moved into a separate test suite (they were previously scattered between `ClangParser` and `SemanticAnalyser` tests)

##### Checklist


src/attached_probe.cpp
```diff
@@ -753,7 +753,9 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         std::string btf_name = name;
         if (probe_.type == ProbeType::iter)
           btf_name = "bpf_iter_" + btf_name;
-        opts.attach_btf_id = btf_.get_btf_id(btf_name);
+        auto btf_id = btf_.get_btf_id_fd(btf_name);
+        opts.attach_btf_id = btf_id.first;
+        opts.attach_prog_fd = btf_id.second;
```
BurntBrunch, line 758: Can we use the other union member, `attach_btf_obj_fd` instead? In the libbpf headers, they're actually two separate members.

Further, I actually think we should explicitly convert the ID to an FD here, since this is the only place where we need it. Right now, the code is acquiring FDs for everything and not using them in 99% of cases.

src/btf.cpp
```diff
@@ -749,9 +709,119 @@ std::set<std::string> BTF::get_all_structs() const
...
-  return btf__find_by_name_kind(btf, name.c_str(), BTF_KIND_FUNC);
+  __s32 id = btf__find_by_name_kind(vmlinux_btf, func.c_str(), BTF_KIND_FUNC);
+  if (id >= 0)
+    return { id, 0 };
+
+  for (auto &mod_btf : modules_btf)
+  {
+    id = btf__find_by_name_kind(mod_btf.btf, func.c_str(), BTF_KIND_FUNC);
+    if (id >= 0)
+      return { id, bpf_btf_get_fd_by_id(mod_btf.id) };
```
BurntBrunch, line 735: Isn't this leaking the FD? Who's closing it?
Overall, I feel bpftrace is long overdue for a RAII FD type.

BurntBrunch (commented):
Functionality-wise, this is great! I do worry about how it scales with modules/complexity of BTF though, a lot of the work is eager/iterates over everything. There may be a future "make all this lazy" issue, especially if we start dumping module types.

src/btf.cpp
```diff
@@ -369,7 +369,7 @@ const struct btf_type *BTF::btf_type_skip_modifiers(const struct btf_type *t,
   return t;
 }
 
-SizedType BTF::get_stype(const BTFId &btf_id)
+SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
```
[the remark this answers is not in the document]
viktormalik: This is useful when resolving fields of structs. If a field is a pointer to foo, we only want to resolve it when it is actually accessed.

tests/runtime/engine/runner.py
```diff
@@ -72,7 +72,7 @@ def skip_reason(test, status):
         if status == Runner.SKIP_KERNEL_VERSION_MAX:
             return "max Kernel: %s" % test.kernel_max
         elif status == Runner.SKIP_REQUIREMENT_UNSATISFIED:
-            return "unmet condition: '%s'" % test.requirement
+            return "unmet condition: '%s'" % ' && '.join(test.requirement)
```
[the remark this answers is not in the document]
viktormalik: We're not using the f-syntax anywhere, so wouldn't change that for one case.

src/btf.cpp
viktormalik: Yes, the FD is leaked. I wasn't sure when it is supposed to be closed - is if ok to close it right after the the BPF program is loaded?

src/attached_probe.cpp
viktormalik: > Can we use the other union member, `attach_btf_obj_fd` instead? In the libbpf headers, they're actually two separate members.

Well, it won't work for libbpf<0.6 which we still support. But I'm fine dropping that support since we plan to move to higher libbpf anyway.

> Further, I actually think we should explicitly convert the ID to an FD here, since this is the only place where we need it. Right now, the code is acquiring FDs for everything and not using them in 99% of cases.

In fact we do, it's just hidden inside `get_btf_id_fd` which is only called from here (and from `bpffeature` when trying to load some programs). But I agree that it may be cleaner to do the conversion in this function, especially if the function will be closing it (see my question about closing the FD below).

viktormalik:
> I do worry about how it scales with modules/complexity of BTF though, a lot of the work is eager/iterates over everything. There may be a future "make all this lazy" issue, especially if we start dumping module types.

Indeed, I have the same impression. I gave it some more thought and came with some observations/ideas:

- We have to iterate everything at least once to find out which module (or let's say "BTF source") contains the function that we're attaching to. After that, we could store that BTF source within the probe and use it for the rest of the BTF queries (e.g. type resolution).
- This may be a problem if one probe attaches to multiple modules (e.g. via wildcards). Not sure if we want to allow something like that, but we could store a list of BTF sources, which could be fine as it'll usually be of size 1.
- In general, two modules may define functions/types with the same name. We should probably handle that, so I'd suggest to:
  - print a warning when user tries to access a function/type defined in multiple modules
  - allow the user to explicitly specify which module he wants to access (e.g. `kfunc:modname:function`)

Does this make sense? Any ideas how we could do even better?

BurntBrunch:
Okay, how about this:
1) Parse kallsyms and keep the module names that symbols come from
2) When resolving wildcards, resolve against kallsyms, remembering the module, then filter to only attachable.
3) When we're ready to attach, we now know the module names we need, so we don't have to parse _everything_.

This would save us from walking all BTF types, though we'd still have to iterate over all modules because of the ID, FD tuple.
The downside is that kallsyms gets pretty big but honestly, I think it's useful enough to be worth it.

> allow the user to explicitly specify which module he wants to access (e.g. `kfunc:modname:function`)

I think there's value in this syntax extension regardless! However, I think it should be `function:modname` so you don't have to edit the middle of the probe spec when clarifying the module name.

viktormalik:
> Okay, how about this:
> 
> 1. Parse kallsyms and keep the module names that symbols come from
> 2. When resolving wildcards, resolve against kallsyms, remembering the module, then filter to only attachable.

And what about parsing `sys/kernel/debug/tracing/available_filter_functions`? It also contains the module name and we're parsing it anyway to get the list of traceable functions. And it's also considerably smaller than kallsyms.

> I think there's value in this syntax extension regardless! However, I think it should be `function:modname` so you don't have to edit the middle of the probe spec when clarifying the module name.

While I see your point, I think we should stay consistent with other probe types. E.g. tracepoints have the form `tracepoint:subsystem:event`, which corresponds more to `kfunc:module:function`.

BurntBrunch:
> And what about parsing sys/kernel/debug/tracing/available_filter_functions? 

That's what I meant by "filter to attachable" above. I didn't realize it contains the module name, so, sure, we don't need kallsyms for this filtering.

Most reasons to look at kallsyms are relevant for kprobe anyway, which doesn't matter here.

> While I see your point, I think we should stay consistent with other probe types.

Sure, my opinion is very weakly held. :smiley: Besides, this module specifier would be mostly useful for wildcard resolution (duplicate symbols are not something that people would normally encounter), in which case function/pattern at the end makes sense.

viktormalik:
Version 2 with several major improvements:
- before parsing BTF, collect names of all modules that we're attaching to from `available_filter_functions`, then parse BTF only for those modules (vmlinux BTF is always parsed), which reduces the amount of BTF that must be iterated
- enable `kfunc:module:function` syntax which allows to attach to functions from specific module(s) only, supports also wildcards
- minimal required libbpf is now 0.6.0
- addressed all the PR comments and added several more fixes

More implementation details can be found in commit messages. The PR is now quite large, so I'm happy to break it if it simplifies reviewing.

Note: there is a BPF verifier bug which causes incorrect attachment when dealing with multiple functions from different modules (or vmlinux) having the same name. In practice, it will always attach to the first function of that name found in kallsyms. I'm working on a fix but I think that we can proceed with this independently as the situation should occur rarely. If someone finds it useful, we can show a warning or mention it in the reference guide.

src/attached_probe.cpp
```diff
@@ -734,32 +731,59 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
+        opts.name = name.c_str();
+        opts.insns = reinterpret_cast<struct bpf_insn *>(insns);
+        opts.insns_cnt = prog_len / sizeof(struct bpf_insn);
+        opts.license = license;
+        progfd_ = bpf_load_program_xattr(&opts, log_buf.get(), log_buf_size);
 #endif
+      }
+
+      if (opts.attach_prog_fd > 0)
+        close(opts.attach_prog_fd);
```
BurntBrunch, line 786: :heart: thanks for fixing the fd leak

src/btf.cpp
```diff
@@ -38,151 +39,116 @@ static __u32 type_cnt(const struct btf *btf)
...
+{
+  fprintf(stderr, "BTF: (%d) ", level);
+  return vfprintf(stderr, msg, ap);
+}
 
-  size = st.st_size;
+BTF::BTF(const std::set<std::string> &modules) : state(NODATA)
+{
+  if (bt_debug != DebugLevel::kNone)
+    libbpf_set_print(libbpf_print);
```
BurntBrunch, line 58: This belongs at the main.cpp level, not in here. If I use btf.cpp in a library, I shouldn't get libbpf behavior altered from under me.

src/btf.cpp
```diff
@@ -38,151 +39,116 @@ static __u32 type_cnt(const struct btf *btf)
...
-      btf = btf_raw(path);
-    else
-      btf = btf__parse_elf(path, nullptr);
-
-    int err = libbpf_get_error(btf);
+    // Get BTF object info - needed to determine if this is a kernel module BTF
+    char name[64];
+    struct bpf_btf_info info;
+    __u32 info_len = sizeof(info);
+    memset(&info, 0, info_len);
```
BurntBrunch, line 116: Prefer C++ default init: 
```suggestion
    struct bpf_btf_info info{};
```

src/btf.cpp
```diff
@@ -38,151 +39,116 @@ static __u32 type_cnt(const struct btf *btf)
...
+  state = OK;
 }
 
-static int libbpf_print(enum libbpf_print_level level, const char *msg, va_list ap)
+BTF::~BTF()
 {
-  fprintf(stderr, "BTF: (%d) ", level);
-  return vfprintf(stderr, msg, ap);
+  for (auto &btf_obj : btf_objects)
+    btf__free(btf_obj.btf);
```
BurntBrunch, line 85: If you wanna get C++ fancy, you can put them in 
`std::unique_ptr<struct btf, void (*)(struct btf *)>` like this:
```
std::unique_ptr<struct btf, void (*)(struct btf *)> btf_obj{btf, &btf__free};
```

src/btf.h
```diff
@@ -24,9 +25,35 @@ class BTF
...
+  struct BTFObj
+  {
+    struct btf* btf;
+    __u32 id;
+    std::string name;
+
+    BTFObj(struct btf* btf, __u32 id, const std::string& name)
+        : btf(btf), id(id), name(name)
+    {
+    }
```
BurntBrunch, line 39: Why do you need this constructor? Designated initializers are better for plain structs:
```
BTFObj foo{.btf = btf, .id = id, .name = name};
```

src/probe_matcher.cpp
```diff
@@ -557,4 +567,24 @@ void ProbeMatcher::list_structs(const std::string& search)
     std::cout << match << std::endl;
 }
 
+// Transform the kernel syntax (function [module]) into bpftrace syntax
+// (module:function).
+void ProbeMatcher::adjust_kernel_modules(
+    std::unique_ptr<std::istream>& symbol_list) const
```
BurntBrunch, line 573: This should return a new istream and not modify a unique_ptr taken by reference - unique_ptr& is a pretty unidiomatic construct in the first place.

src/struct.h
```diff
@@ -2,6 +2,7 @@
 
 #include <map>
 #include <memory>
+#include <unordered_set>
```
BurntBrunch, line 5: is this necessary here or in utils.h/cpp?

src/utils.cpp
```diff
@@ -1210,4 +1210,18 @@ int get_kernel_ptr_width()
...
+{
+  if (!symbol_has_module(symbol))
+    return { symbol, "" };
+
+  size_t idx = symbol.rfind(" [");
+  if (idx == std::string::npos)
+    return { symbol, "" };
+
+  return { symbol.substr(0, idx),
+           symbol.substr(idx + 2, symbol.length() - idx - 3) };
```
BurntBrunch, line 1224: What are the three symbols at the end? "] \0"? 
Consider doing something like
```suggestion
symbol.substr(idx + strlen(" ["), symbol.length() - idx - strlen("] ") - 1)
```
Compilers can inline this these days.

BurntBrunch (approved):
I like the as-lazy-as-possible approach! The new interactions between FieldAnalyser and ClangParser were a little hard to figure out but I think that's just me not being familiar with the older code.

> More implementation details can be found in commit messages.

I wish there was a nice architectural diagram somewhere that we could change as part of this rework. Oh well.

I'm very much of the opinion that we should merge this and bake it pre-release instead of trying to iron out every possible edge case in the PR.

src/btf.cpp
viktormalik: While I'm usually a fan of this kind of solutions, it would just unnecessarily complicate the code in this case.

src/struct.h
viktormalik: It's completely superfluous here, removed.

src/utils.cpp
viktormalik: In fact, it's `" []"` (to get the length of the module name). But your suggestion is still very valid, I used it. Thanks!

.github/workflows/ci.yml
```diff
@@ -152,6 +152,9 @@ jobs:
       if: steps.docker-cache.outputs.cache-hit == 'true'
       run: >
         docker load --input /tmp/docker-save/i.tar
+    - name: Load kernel modules
+      # nf_tables is necessary for testing kernel modules BTF support
+      run: sudo modprobe nf_tables
```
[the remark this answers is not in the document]
viktormalik: That would be something :smile: Probably worth trying.

viktormalik:
> I wish there was a nice architectural diagram somewhere that we could change as part of this rework. Oh well.

There is a nice picture in [internals_development.md](https://github.com/iovisor/bpftrace/blob/master/docs/internals_development.md), although it will be partially outdated by now. Still, I can have a look into it while updating the development guide.

danobi:
I read through most of this PR and it looks quite reasonable.

viktormalik:
Thanks for the reviews, let's not prolong this and merge.

PR #2316, opened 2022-08-04
Fix invalid LLVM IR as detected by tests

### LINKED ISSUE #2297: Detect and fix invalid LLVM IR generated by bpftrace

lenticularis39:
With the option to verify LLVM IR generated by runtime tests (implemented in #2298), several issues with codegen were found. This PR includes commits to fix them, along with enabling LLVM IR verification in tests and enabling remaining failing LLVM 14 tests. The reasoning behind each fix is explained in the individual commits.

Closes #2297

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -612,7 +612,14 @@ void CodegenLLVM::visit(Call &call)
     AllocaInst *buf = b_.CreateAllocaBPF(bpftrace_.strlen_, "path");
     b_.CREATE_MEMSET(buf, b_.getInt8(0), bpftrace_.strlen_, 1);
     call.vargs->front()->accept(*this);
-    b_.CreatePath(ctx_, buf, expr_, call.loc);
+    b_.CreatePath(ctx_,
+                  buf,
+                  b_.CreateCast(expr_->getType()->isPointerTy()
+                                    ? Instruction::BitCast
+                                    : Instruction::IntToPtr,
+                                expr_,
+                                b_.getInt8PtrTy()),
+                  call.loc);
     expr_ = buf;
     expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
   }
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2581,7 +2581,7 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
     else
     {
       // Two or more values as a map key (e.g, @[comm, pid] = 1;)
-      key = getMultiMapKey(map, {}, 0);
+      key = getMultiMapKey(map, {});
     }
   }
   else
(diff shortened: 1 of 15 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -612,7 +612,8 @@ void CodegenLLVM::visit(Call &call)
     AllocaInst *buf = b_.CreateAllocaBPF(bpftrace_.strlen_, "path");
     b_.CREATE_MEMSET(buf, b_.getInt8(0), bpftrace_.strlen_, 1);
     call.vargs->front()->accept(*this);
-    b_.CreatePath(ctx_, buf, expr_, call.loc);
+    b_.CreatePath(
+        ctx_, buf, b_.CreateIntToPtr(expr_, b_.getInt8PtrTy()), call.loc);
```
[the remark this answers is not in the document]
lenticularis39: I'm not aware of any case where `expr_` is not either a `int8 *` (in that case nothing happens, since `CreateCast` called by `CreateIntToPtr` returns its argument unchanged if it already is of the target type) or an `uint64` (in that case the `inttoptr` is valid). I also checked that both `path(args->path)` in `vfs_truncate` and `path(args->file->path)` in `vfs_fallocate` pass LLVM IR verifier.

However, it might still be of use to create a pointer cast should `expr_` be a pointer, since the user might theoretically use of `path` on a different argument that is a pointer different from `int8 *`, hence I'll fix that.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2658,9 +2658,14 @@ AllocaInst *CodegenLLVM::getMultiMapKey(Map &map,
                                      key,
```
fbs, line 2667: IIRC CreateGEP creates the right type based on its arguments so if we have to cast `offset_val` the type of `key_type` must be wrong?

viktormalik: IIUC, `key_type` is a byte array (to comprise all possible key types) so we have to cast the result of the GEP to the type that we want to store. The question here is if the cast to `i64 *` is correct, as the extra key may not always be an `i64` (although currently it always is).

Btw. there's a bug in this loop - `offset` is never incremented so it won't work with multiple `extra_keys` (at this point, at most one extra key is passed). @lenticularis39 perhaps you could fix that as a part of this PR?

lenticularis39: The extra key is currently only used in histogram maps, where it is always `i64`. I guess we can cast to `extra_key->getType()->getPointerTo()` to accommodate for possible future uses of extra key.

As of the missing `offset` incrementation: the extra keys are passed as a vector of `Value`s, whose type has to be inferred from `DataLayout` (unlike the type of non-extra keys). `extra_keys_size` only contains the sum of the sizes of the extra keys (which is now the same as the size of the extra key, since there is only one used). Do you think it would be better to pass a vector of sizes instead of `extra_keys_size` or to drop the argument altogether and infer the sizes from `DataLayout`?

viktormalik: Let's infer it from data layout, if possible.

lenticularis39: OK, I pushed both changes into the PR branch.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2658,9 +2670,15 @@ AllocaInst *CodegenLLVM::getMultiMapKey(Map &map,
...
     if (aligned)
-      b_.CreateStore(extra_key, offset_val);
+      b_.CreateStore(extra_key,
+                     b_.CreatePointerCast(
+                         offset_val, extra_key->getType()->getPointerTo()));
     else
-      b_.createAlignedStore(extra_key, offset_val, 1);
+      b_.createAlignedStore(
+          extra_key,
+          b_.CreatePointerCast(offset_val, b_.getInt64Ty()->getPointerTo()),
```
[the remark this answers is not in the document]
lenticularis39: Oops, I forget the cast is also in the other branch, of course it should be the same for both branches.

viktormalik (approved):

fbs:
good stuff!

PR #2326, opened 2022-08-12
Pipe helper errors into normal text and json output

BurntBrunch:
Helper errors (-k or -kk) are currently only visible on stderr.
Change that behavior to emit them to text output (-o) and in JSON
format.

This allows tools that process the JSON output to account for the
errors, similarly to lost perf events.

##### Checklist


viktormalik (commented):
Good idea, I like this. There are some clang-format issues that need to be fixed, though.

BurntBrunch:
Fixed clang format and added a changelog entry.

danobi (approved):

viktormalik (approved):

PR #2329, opened 2022-08-16
Add strerror builtin

### LINKED ISSUE #2114: builtin for int to errno constant string translation
Per discussion in #2082, it would be nice to have a way to translate an errno integer into the corresponding constant string.

Could make a builtin function that does the translation. You'd probably want to store the `int -> string` mapping as a readonly bpf map that the prog can access. And only load the map if the script uses that builtin.

_Originally posted by @danobi in https://github.com/iovisor/bpftrace/discussions/2082#discussioncomment-1801628_

lenticularis39:
The new builtin behaves just like strerror in C, with the expection
of the conversion being done in userspace when the value returned
by the helper is printed.

Closes #2114

##### Checklist

### DIFF
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -375,6 +375,11 @@ REQUIRES grep -q '^cgroup2' /proc/mounts
 TIMEOUT 5
 MIN_KERNEL 4.18
 
+NAME strerror
+RUN {{BPFTRACE}} --include "errno.h" -e 'BEGIN { print(strerror(EPERM)); exit(); }'
+EXPECT Operation not permitted
+TIMEOUT 5
+
 NAME bswap_int64
 RUN {{BPFTRACE}} -e 'BEGIN { printf("%lx", bswap(0x1122334455667788)); exit(); }'
 EXPECT 8877665544332211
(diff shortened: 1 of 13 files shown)


tests/runtime/call
```diff
@@ -375,6 +375,11 @@ REQUIRES grep -q '^cgroup2' /proc/mounts
 TIMEOUT 5
 MIN_KERNEL 4.18
 
+NAME strerror
+PROG BEGIN { print(strerror(1)); exit(); }
```
viktormalik, line 379: Would this work?
```suggestion
RUN {{BPFTRACE}} --include "stderr.h" -e 'BEGIN { print(strerror(EPERM)); exit(); }'
```
Seems more reliable to me.

lenticularis39: Yes (with `errno.h`).

fbs:
I like the feature but I'm not sure about the name. I get that it's inspired by C but wonder if we can't do something better that makes more sense to those not from a C background.

lenticularis39:
My first idea was to simply name the function `errno`, but it's not very clear what it does from that name (other than it's related to errno) and it is confusing for C programmers, being the same as the `errno` variable. 

Another name could be `errno_to_str` or something like that, I decided to go for `strerror`, since the function is used the same way as C's `strerror`, and we already have many builtins named after C functions, e.g. `printf`, `strncmp`, `sizeof`, `strftime` and so on (some of these also don't make sense to someone without C background).

I also think that users of bpftrace almost always have some C or C++ background, given it's a tool to trace C and C++ code.

viktormalik (approved):

dxuuu (approved):
Yeah, I find that argument about C/C++ familiarity convincing

danobi:
Pls rebase -- will merge after

PR #2330, opened 2022-08-17
Add cmake TARGET uninstall

Rtoax:
usage:

```bash
 $ cd bpftrace
 $ mkdir build
 $ cd build
 $ cmake ..
 $ sudo make install
 $ sudo make uninstall <<=== New Target
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,3 +302,13 @@ add_subdirectory(tools)
 if (ENABLE_MAN)
   add_subdirectory(man)
 endif(ENABLE_MAN)
+
+if(NOT TARGET uninstall)
+  configure_file(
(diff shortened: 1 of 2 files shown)


danobi (commented):
The code looked a bit complicated but I realized that it was a copy paste from https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake .

I guess if the actual cmake wiki recommends it, it probably works.

How about adding a link to the wiki so there's some context?

Rtoax:
Yep, add https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake as ref

danobi (approved):

PR #2332, opened 2022-08-18
Userspace kprobe offset verification can now fail

### LINKED ISSUE #2327: bpftrace cannot attach kprobe to module symbol+offset
### What reproduces the bug?
Trying to attach to an offset in a module function always fails. (`icmp_match` is in `ip_tables.ko`)

```
sudo src/bpftrace -e 'kprobe:icmp_match {print("enter\n"); exit()}'
Attaching 1 probe...
^C
```

```
sudo src/bpftrace -e 'kprobe:icmp_match+0x1 {print("enter\n"); exit()}'
Attaching 1 probe...
ERROR: Could not resolve symbol icmp_match. Use BPFTRACE_VMLINUX env variable to specify vmlinux path. Compile bpftrace with ALLOW_UNSAFE_PROBE option to force skip the check.
```

### bpftrace --info
```
System
  OS: Linux 5.18.17-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Aug 11 14:36:06 UTC 2022
  Arch: x86_64

Build
  version: v0.15.0-85-g14ec
  LLVM: 14.0.0
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bcc_usdt_addsem: yes
  libbpf btf dump: yes
  libbpf bpf_prog_load: yes
  libbpf bpf_map_create: yes
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes
  skboutput: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  map batch: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  iter:task: yes
  iter:task_file: yes
  kprobe_multi: no
  raw_tp_special: yes
```

### Discussion

This fails because we try to verify the kprobe offset ourselves in order to disallow offsets in the middle of instructions or outside of functions. When we do that, we end up parsing symbol tables for the kernel (thus requiring kernel debug symbols to be present, which is not documented anywhere afaict) but not for any modules.

The discussion in #956 shows that this was chosen because it produces better sanity checks, even though it does not add any safety per se. The kprobe machinery in the kernel already verifies that the location is viable in more ways than just "is at instruction boundary."

Given that this logic breaks module attachment with offsets and there's no viable way to extend it (we can't reasonably parse all modules' debug info), I think we can
1) save us a bunch of file system lookups
2) save us a bunch of symbol scanning and disassembly and
3) remove the need for kernel debug symbols
by just letting perf_event_open fail with its message? Perhaps we can extend the user-visible error string with diagnostics on possible failure modes?

@mmisono, @danobi, I'd love to hear you input!

BurntBrunch:
Previously, this logic could only fail in unsafe mode. Unfortunately,
this prevented attachment to module functions with an offset, as
the module binary would never be examined.

This PR makes the logic entirely optional and improves the error
messages to provide a similar user experience.

Fixes #2327 

##### Checklist

- [x] The new behaviour is covered by runtime tests using the nf_tables module that #2315 uses as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to
   - [#2264](https://github.com/iovisor/bpftrace/pull/2264)
 - Helper errors (-k, -kk options) are now emitted to text or json output
   - [#2326](https://github.com/iovisor/bpftrace/pull/2326)
+- kprobe offset verification is now optional, without requiring --unsafe
+  - [#2332](https://github.com/iovisor/bpftrace/pull/2332)
 
 #### Deprecated
 #### Removed
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -76,6 +76,23 @@ EXPECT Offset outside the function bounds \('vfs_read' size is*
 TIMEOUT 5
 WILL_FAIL
 
+NAME kprobe_offset_module
+RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x5 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
+EXPECT hit
+TIMEOUT 5
+REQUIRES lsmod | grep '^nf_tables'
+REQUIRES /usr/sbin/nft --help
+AFTER nft delete table bpftrace
+
+NAME kprobe_offset_module_error
+RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x1 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
+EXPECT Possible attachment attempt in the middle of an instruction, try a different offset.
+TIMEOUT 5
+REQUIRES lsmod | grep '^nf_tables'
+REQUIRES /usr/sbin/nft --help
+AFTER nft delete table bpftrace
+WILL_FAIL
+
 NAME kretprobe
 PROG kretprobe:vfs_read { printf("SUCCESS %d\n", pid); exit(); }
 EXPECT SUCCESS [0-9][0-9]*
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -586,27 +586,15 @@ void AttachedProbe::resolve_offset_kprobe(bool safe_mode)
   std::string path = find_vmlinux(locs, sym);
   if (path.empty())
   {
-    if (safe_mode)
+    if (bt_verbose)
     {
-      std::stringstream buf;
-      buf << "Could not resolve symbol " << symbol << ".";
-      buf << " Use BPFTRACE_VMLINUX env variable to specify vmlinux path.";
-#ifdef HAVE_UNSAFE_PROBE
-      buf << " Use --unsafe to skip the userspace check.";
-#else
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to
   - [#2264](https://github.com/iovisor/bpftrace/pull/2264)
 - Helper errors (-k, -kk options) are now emitted to text or json output
   - [#2326](https://github.com/iovisor/bpftrace/pull/2326)
+- kprobe offset verification is now optional, without requiring --unsafe
+  - [#2331](https://github.com/iovisor/bpftrace/pull/2331)
```
viktormalik, line 27: Should be #2332.

tests/runtime/probe
```diff
@@ -76,6 +76,23 @@ EXPECT Offset outside the function bounds \('vfs_read' size is*
 TIMEOUT 5
 WILL_FAIL
 
+NAME kprobe_offset_module
+RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x5 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
+EXPECT hit
+TIMEOUT 5
+REQUIRES lsmod | grep '^nf_tables'
+REQUIRES /usr/sbin/nft --help
```
viktormalik, line 84: This won't work until #2315 is merged (currently only the second `REQUIRES` is checked). I'll factor that out and submit as a separate PR not to block this.

viktormalik: See #2333.

BurntBrunch:
Thanks @viktormalik. I just fixed the changelog, will wait for #2333 to land, so I can rebase and get clean CI (even though it passed with the partial REQUIRES)

tests/runtime/probe
```diff
@@ -76,6 +76,23 @@ EXPECT Offset outside the function bounds \('vfs_read' size is*
 TIMEOUT 5
 WILL_FAIL
 
+NAME kprobe_offset_module
```
danobi, line 79: Maybe add comment that this test relies on lack of module instruction checking support?

BurntBrunch: @danobi, it's just punting the check to the kernel - if you get the offset wrong, the kernel won't let the program attach. After all, if we implemented module disassembly and offset checking, this test would _still_ pass, so it's clearly not relying on the absence of anything.

danobi:
Off by 1 in changelog link?

BurntBrunch:
> Off by 1 in changelog link?

Sorry, forgot to push.

viktormalik:
> > Off by 1 in changelog link?
> 
> Sorry, forgot to push.

The description is correct, but the link is still wrong.

viktormalik (approved):

danobi (approved):

PR #2333, opened 2022-08-19
Runtime tests: allow multiple REQUIRES

viktormalik:
Some tests may require multiple conditions to be run. Using multiple `REQUIRES` fields is more convenient than just a single `REQUIRES` with multiple commands joint with `&&`.

This was originally a part of #2315 but #2332 will need it earlier, so extracting it as a separate PR.

##### Checklist

### DIFF
--- a/tests/runtime/engine/parser.py
+++ b/tests/runtime/engine/parser.py
@@ -105,7 +105,7 @@ class TestParser(object):
         after = ''
         kernel_min = ''
         kernel_max = ''
-        requirement = ''
+        requirement = []
         env = {}
         arch = []
         feature_requirement = set()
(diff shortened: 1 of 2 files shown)


fbs (approved):

tests/runtime/engine/parser.py
```diff
@@ -105,7 +105,7 @@ def __read_test_struct(test, test_suite):
         after = ''
         kernel_min = ''
         kernel_max = ''
-        requirement = ''
+        requirement = set()
```
danobi, line 108: Is set ordered? Should we use a list to make it more predictable?

viktormalik: Good point, it may come useful to exec the conditions in the same order they were supplied. Fixed.

Issue #1338, opened 2020-05-18
bpftrace looking for /sys/kernel/debug/tracing instead of /sys/kernel/tracing


avdi:
Today I ran into an issue where bpftrace reported that it couldn't find the `available_filter_functions`:

```
root@b40be1747ab9:/workspaces/idw# bpftrace -l
...a very short list, followed by...
No such file or directory: /sys/kernel/debug/tracing/available_filter_functions
```

It turned out that `debugfs` had come unmounted for some reason. But I went down a rabbit hole because I found [this documentation ](https://www.kernel.org/doc/Documentation/trace/ftrace.txt)saying that tracing lives at `/sys/kernel/tracing` not `/sys/kernel/debug/tracing` in recent kernels. And indeed, `/sys/kernel/tracing` WAS available.

Anyway like I said, re-mounting `debugfs` fixed the issue. But it might be nice to have an error message that suggests mounting `debugfs`.

javierhonduco:
Good idea! Wonder if it would make sense to have some subcommand like `bpftrace doctor` to check for various things to be in the "right" place.

Maybe also do some Kernel checks like [this script](https://github.com/iovisor/bpftrace/blob/master/scripts/check_kernel_features.sh)?

fbs:
> Good idea! Wonder if it would make sense to have some subcommand like `bpftrace doctor` to check for various things to be in the "right" place.

I think we can do that in the `--info` flag, displaying either the path to tracefs or `no`.

For detection we could extend (and possibly rename to kernelfeature) bpffeature.

brendangregg:
G'Day @avdi! Glad you're trying bpftrace!

avdi:
Hi @brendangregg! Since diving into the world of BPF I've been having some "oh it was THAT Brenden Gregg that I hung out with in Australia" moments 😅

Davis-A:
I've just run into this issue.  We don't mount debugfs which makes this somewhat awkward

danobi:
We really should just parse mount table and not hard code these paths

neomafo88:
if it helps anyone, what worked for me wanting to build https://github.com/facebookexperimental/ExtendedAndroidTools, which uses `bpftrace`, was:

from root `/ExtendedAndroidTools/projects`,

```
sed -E -i 's|/sys/kernel/debug/tracing/|/sys/kernel/tracing/|g' bcc/sources/docs/reference_guide.md
[...]

PR #2351, opened 2022-09-04
use tracefs directly when debugfs is not available

michalgr:
On some platforms tracefs is available as "/sys/kernel/tracing" and not "/sys/kernel/debug/tracing". This pull request makes bpftrace use "/sys/kernel/tracing" when "/sys/kernel/debug/tracing" is not accessible.

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -3,6 +3,7 @@
 #include "bpftrace.h"
 #include "log.h"
 #include "probe_matcher.h"
+#include "tracefs.h"
 #include "types.h"
 #include "utils.h"
 #include <cstring>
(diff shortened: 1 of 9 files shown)


src/btf.cpp
```diff
@@ -478,7 +479,8 @@ int BTF::resolve_args(const std::string &func,
     {
       if (bpftrace_->traceable_funcs_.empty())
         throw std::runtime_error("could not read traceable functions from " +
-                                 kprobe_path + " (is debugfs mounted?)");
+                                 tracefs::available_filter_functions() +
+                                 " (is debugfs mounted?)");
```
viktormalik, line 483:
```suggestion
                                 " (is tracefs mounted?)");
```

viktormalik (commented):
Looks good, thanks! Just a nit.

viktormalik:
Related to #1338.

viktormalik (approved):
Ok with me. Parsing the mount table would still be better, but this is a nice step forward.

PR #2355, opened 2022-09-09
Disable runtime test failing in CI

viktormalik:
Runtime test `variable.32-bit tracepoint arg` requires max kernel 5.16. It is currently failing in the CI since it has the kernel version 5.15.0-1029-azure (i.e. <= 5.16) which doesn't have the required tracepoint.

This disables the test in the CI until a better solution is found.

##### Checklist

### DIFF
--- a/tests/runtime/variable
+++ b/tests/runtime/variable
@@ -46,13 +46,6 @@ TIMEOUT 5
 AFTER ./testprogs/syscall read
 
 NAME 32-bit tracepoint arg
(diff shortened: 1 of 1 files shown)


ajor:
Maybe we should just delete this test. By the looks of it, we've already got a replacement for it in "32-bit tracepoint arg openat_flags" anyway.

ajor (approved):

PR #2361, opened 2022-09-16
Improve codegen for 32-bit architectures

tnovak:
This is a follow-up to #2360. The main issues I've run into on armv7 have to do with pointers in bpftrace being 64-bit, which results in misaligned loads and/or wrong offsets; for example:
```
# bpftrace -ve 't:syscalls:sys_enter_open { printf("%p\n", args->filename); }'
0: (b7) r2 = 0
1: (7b) *(u64 *)(r10 -16) = r2
2: (79) r2 = *(u64 *)(r1 +12)
invalid bpf_context access off=12 size=8
``` 
(individual commit messages go into more details)

This changeset tackles this by modifying the size of ctx-relative pointer loads and `bpf_probe_read` calls so it depends on the target arch (4 bytes on armv7, unchanged on all other supported architectures) and adjusting ctx offsets when reading register values.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1340,5 +1340,30 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
 #endif
 }
 
+llvm::Type *IRBuilderBPF::getPointerStorageTy(AddrSpace as)
+{
+  switch (as)
+  {
+    case AddrSpace::user:
+      return getUserPointerStorageTy();
+    default:
+      return getKernelPointerStorageTy();
+  }
+}
+
+llvm::Type *IRBuilderBPF::getKernelPointerStorageTy()
+{
+  static int ptr_width = get_kernel_ptr_width();
+
+  return getIntNTy(ptr_width);
+}
+
+llvm::Type *IRBuilderBPF::getUserPointerStorageTy()
+{
+  // TODO: we don't currently have an easy way of determining the pointer size
+  // of the uprobed process, so assume it's the same as the kernel's for now.
+  return getKernelPointerStorageTy();
+}
+
 } // namespace ast
 } // namespace bpftrace
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -481,16 +481,6 @@ void IRBuilderBPF::CreateMapDeleteElem(Value *ctx,
 
 void IRBuilderBPF::CreateProbeRead(Value *ctx,
                                    Value *dst,
-                                   size_t size,
-                                   Value *src,
-                                   AddrSpace as,
-                                   const location &loc)
-{
-  return CreateProbeRead(ctx, dst, getInt32(size), src, as, loc);
-}
-
-void IRBuilderBPF::CreateProbeRead(Value *ctx,
-                                   Value *dst,
                                    llvm::Value *size,
                                    Value *src,
                                    AddrSpace as,
@@ -672,7 +662,7 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
       {
         ptr = CreateAdd(ptr, index_offset);
       }
-      CreateProbeRead(ctx, dst, abs_size, ptr, as, loc);
+      CreateProbeRead(ctx, dst, getInt32(abs_size), ptr, as, loc);
       result = CreateLoad(getIntNTy(abs_size * 8), dst);
     }
     else
@@ -1365,7 +1355,7 @@ void IRBuilderBPF::CreateProbeRead(Value *ctx,
   AddrSpace as = addrSpace ? addrSpace.value() : type.GetAS();
 
   if (!type.IsPtrTy())
-    return CreateProbeRead(ctx, dst, type.GetSize(), src, as, loc);
+    return CreateProbeRead(ctx, dst, getInt32(type.GetSize()), src, as, loc);
 
   // Pointers are internally always represented as 64-bit integers, matching the
   // BPF register size (BPF is a 64-bit ISA). This helps to avoid BPF codegen
@@ -1383,7 +1373,7 @@ void IRBuilderBPF::CreateProbeRead(Value *ctx,
   if (ptr_size != type.GetSize())
     CREATE_MEMSET(dst, getInt8(0), type.GetSize(), 1);
 
-  CreateProbeRead(ctx, dst, ptr_size, src, as, loc);
+  CreateProbeRead(ctx, dst, getInt32(ptr_size), src, as, loc);
 }
 
 llvm::Value *IRBuilderBPF::CreateDatastructElemLoad(
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1162,16 +1162,31 @@ Value *IRBuilderBPF::CreateRegisterRead(Value *ctx, const std::string &builtin)
   else // argX
     offset = arch::arg_offset(atoi(builtin.substr(3).c_str()));
 
-  Value *ctx_ptr = CreatePointerCast(ctx, getInt64Ty()->getPointerTo());
+  return CreateRegisterRead(ctx, offset, builtin);
+}
+
+Value *IRBuilderBPF::CreateRegisterRead(Value *ctx,
+                                        int offset,
+                                        const std::string &name)
+{
+  // Bitwidth of register values in struct pt_regs is the same as the kernel
+  // pointer width on all supported architectures.
+  llvm::Type *registerTy = getKernelPointerStorageTy();
+  Value *ctx_ptr = CreatePointerCast(ctx, registerTy->getPointerTo());
   // LLVM optimization is possible to transform `(uint64*)ctx` into
   // `(uint8*)ctx`, but sometimes this causes invalid context access.
   // Mark every context access to suppress any LLVM optimization.
-  Value *result = CreateLoad(getInt64Ty(),
-                             CreateGEP(getInt64Ty(), ctx_ptr, getInt64(offset)),
-                             builtin);
+  Value *result = CreateLoad(registerTy,
+                             CreateGEP(registerTy, ctx_ptr, getInt64(offset)),
+                             name);
   // LLVM 7.0 <= does not have CreateLoad(*Ty, *Ptr, isVolatile, Name),
   // so call setVolatile() manually
   dyn_cast<LoadInst>(result)->setVolatile(true);
+  // Caller expects an int64, so add a cast if the register size is different.
+  if (result->getType()->getIntegerBitWidth() != 64)
+  {
+    result = CreateIntCast(result, getInt64Ty(), false);
+  }
   return result;
 }
 
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1355,6 +1355,62 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
 #endif
 }
 
+void IRBuilderBPF::CreateProbeRead(Value *ctx,
+                                   Value *dst,
+                                   const SizedType &type,
+                                   Value *src,
+                                   const location &loc,
+                                   std::optional<AddrSpace> addrSpace)
+{
+  AddrSpace as = addrSpace ? addrSpace.value() : type.GetAS();
+
+  if (!type.IsPtrTy())
+    return CreateProbeRead(ctx, dst, type.GetSize(), src, as, loc);
+
+  // Pointers are internally always represented as 64-bit integers, matching the
+  // BPF register size (BPF is a 64-bit ISA). This helps to avoid BPF codegen
+  // issues such as truncating PTR_TO_STACK registers using shift operations,
+  // which is disallowed (see https://github.com/iovisor/bpftrace/pull/2361).
+  // However, when reading pointers from kernel or user memory, we need to use
+  // the appropriate size for the target system.
+  const size_t ptr_size = getPointerStorageTy(as)->getIntegerBitWidth() / 8;
+
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+  // TODO: support 32-bit big-endian systems
+  assert(ptr_size == type.GetSize());
+#endif
+
+  if (ptr_size != type.GetSize())
+    CREATE_MEMSET(dst, getInt8(0), type.GetSize(), 1);
+
+  CreateProbeRead(ctx, dst, ptr_size, src, as, loc);
+}
+
+llvm::Value *IRBuilderBPF::CreateDatastructElemLoad(
+    const SizedType &type,
+    llvm::Value *ptr,
+    bool isVolatile,
+    std::optional<AddrSpace> addrSpace)
+{
+  AddrSpace as = addrSpace ? addrSpace.value() : type.GetAS();
+  llvm::Type *ptr_storage_ty = getPointerStorageTy(as);
+
+  if (!type.IsPtrTy() || ptr_storage_ty == getInt64Ty())
+    return CreateLoad(GetType(type), ptr, isVolatile);
+
+  assert(GetType(type) == getInt64Ty());
+
+  // Pointer size for the given address space doesn't match the BPF-side
+  // representation. Use ptr_storage_ty as the load type and cast the result
+  // back to int64.
+  llvm::Value *expr = CreateLoad(
+      ptr_storage_ty,
+      CreatePointerCast(ptr, ptr_storage_ty->getPointerTo()),
+      isVolatile);
+
+  return CreateIntCast(expr, getInt64Ty(), false);
+}
+
 llvm::Type *IRBuilderBPF::getPointerStorageTy(AddrSpace as)
 {
   switch (as)
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -271,10 +271,19 @@ void CodegenLLVM::visit(Builtin &builtin)
     int arg_num = atoi(builtin.ident.substr(4).c_str());
     Value *sp = b_.CreateRegisterRead(ctx_, sp_offset, "reg_sp");
     AllocaInst *dst = b_.CreateAllocaBPF(builtin.type, builtin.ident);
-    Value *src = b_.CreateAdd(sp,
-                              b_.getInt64((arg_num + arch::arg_stack_offset()) *
-                                          sizeof(uintptr_t)));
-    b_.CreateProbeRead(ctx_, dst, 8, src, builtin.type.GetAS(), builtin.loc);
+
+    // Pointer width is used when calculating the SP offset and the number of
+    // bytes to read from stack for each argument. We pass a pointer SizedType
+    // to CreateProbeRead to make sure it uses the correct read size while
+    // keeping builtin.type an int64.
+    size_t arg_width =
+        b_.getPointerStorageTy(builtin.type.GetAS())->getIntegerBitWidth() / 8;
+    SizedType arg_type = CreatePointer(CreateInt8(), builtin.type.GetAS());
+    assert(builtin.type.GetSize() == arg_type.GetSize());
+
+    Value *src = b_.CreateAdd(
+        sp, b_.getInt64((arg_num + arch::arg_stack_offset()) * arg_width));
+    b_.CreateProbeRead(ctx_, dst, arg_type, src, builtin.loc);
     expr_ = b_.CreateLoad(b_.GetType(builtin.type), dst);
     b_.CreateLifetimeEnd(dst);
   }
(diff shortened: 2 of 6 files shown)


tnovak:
Considering that BPF is a 64-bit ISA, are there any fundamental issues with bpftrace doing 32-bit pointer loads/stores I may have missed?

There's one corner case I've hit where adding casts to read 4-bytes from `ctx` generates incorrect BPF instructions with LLVM 14.0.6. IR that bpftrace builds looks roughly like this:
```
  %1 = ptrtoint i8* %0 to i64     ; ctx
  %cast = trunc i64 %1 to i32     ; cast to Foo *, which is 4 bytes
  %2 = inttoptr i32 %cast to i32*
  %3 = load volatile i32, i32* %2, align 4
``` 
Because of that `trunc` instruction, this compiles into the following BPF bytecode:
```
  r1 <<= 32
  r1 >>= 32
  r1 = *(u32 *)(r1 + 0)
```
However, this is not valid BPF code: register `r1` is of type `PTR_TO_CTX`, and shifts on non-scalar registers are [not allowed](https://github.com/torvalds/linux/blob/a335366bad1364a07f49df9da1fdfa6d411a5f39/kernel/bpf/verifier.c#L8270). After the first shift, `r1` becomes `SCALAR_VALUE` so the BPF verifier rejects the memory access. I'm still investigating what's the best way to fix this.

src/ast/irbuilderbpf.cpp
```diff
@@ -392,33 +388,33 @@ Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
       "map_lookup_cond");
   CreateCondBr(condition, lookup_success_block, lookup_failure_block);
 
+  assert(value->getType()->isPointerTy());
+  llvm::Type *value_ptr_ty = value->getType()->getPointerElementType();
```
ajor, line 392: The name `value_ptr_ty` is a little confusing for me - it's the element (i.e. pointee) type not the pointer type, right?

As for the functionality, bpftrace stores all integers in maps as 64-bit: https://github.com/iovisor/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L3232

Does this function not always return Int64Ty because of this?

On another note, apparently getPointerElementType is being deprecated: https://llvm.org/doxygen/classllvm_1_1Type.html#ae61f17d62c8a79bc5cffa63a2f152311
So if it's actually doing something other than always returning Int64Ty then we should try to find another way of doing it.

ajor (commented):
I'm only a small part of the way through so this is a partial review, but I love the `CreateRegisterRead()` cleanup!

I still need to think some more about using 32-bit pointers inside 64-bit BPF - I'll have to try and get myself a 32-bit ARM environment to play around with.

src/ast/irbuilderbpf.cpp
tnovak: `value_ptr_ty` is equivalent to `GetType(type)` here. I agree it's confusing; perhaps something like this would be better?

```
llvm::Type *value_ty = GetType(type);
AllocaInst *value = CreateAllocaBPF(value_ty, "lookup_elem_val");
// [...]
```

> Does this function not always return Int64Ty because of this?

Case where `type.IsIntTy() == true` should be unaffected; the issue here were maps containing pointers (for example, in `@a[0] = (struct Foo *) arg0` the map contains native pointers, so 4-byte values on arm32).

Hm, let me think some more whether it might be better to store pointers in maps as 64-bit ints. My concern originally was that this could lead to too many special cases (already had to do something similar in [ffebc5c](https://github.com/iovisor/bpftrace/pull/2361/commits/ffebc5c876892cba3d99bce36affdb494f258c59) to get around a BPF codegen issue).

> On another note, apparently getPointerElementType is being deprecated

I saw a comment in one of the LLVM headers about moving to opaque pointers (after which `getPointerElementType` will no longer work). We can avoid it in this method with the code above, but there are several other places in the codebase that would need to be updated. Not sure if there's a good alternative other than explicitly tracking T every time we create pointer-to-T, which might require quite a bit of refactoring. :/

tnovak:
> I still need to think some more about using 32-bit pointers inside 64-bit BPF - I'll have to try and get myself a 32-bit ARM environment to play around with.

Another option could be a 32-bit x86 VM -- I suspect we'd run into same issues there.

I debated between these two approaches initially:
1. Make the `Type::pointer` size equivalent to `sizeof(void*)` and fix (hopefully not too many) type mismatches by leveraging the existing test suite and LLVM asserts.
2. Keep pointers 64-bit internally, but modify various load and store operations, bpf_probe_read call etc. to use the arch-dependent pointer size.

I went with the first approach because the discrepancy between internal and target pointer types seemed like it would otherwise make codegen more complicated and messy. Unfortunately, with some hacks like [ffebc5c](https://github.com/iovisor/bpftrace/pull/2361/commits/ffebc5c876892cba3d99bce36affdb494f258c59) that were needed to avoid LLVM emitting incorrect BPF code (shifts on pointer registers; 32-bit spills of pointer registers to stack), this implementation ended up being quite involved too.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1926,11 +1926,30 @@ void CodegenLLVM::visit(Cast &cast)
   {
     // If casting to a pointer and the size of the integer expression doesn't
     // match the pointer size (e.g. "(struct Foo*)(uint8)42"), add an explicit
-    // cast.
+    // conversion.
     auto pointer_ty = b_.GetType(cast.type);
     if (expr_->getType() != pointer_ty)
     {
-      expr_ = b_.CreateIntCast(expr_, pointer_ty, false, "cast");
+      // Rather than keeping expr_ as an integer expression and casting it,
```
BurntBrunch, line 1933: CreateIntCast -> CreateIntToPtr feels like the right call, given that it expresses the semantics better (a pointer cast should return a pointer!).

However, I am not 100% sure I follow why CreateIntCast casts up to i64 given that it's going off of pointer_ty, which should be 32-bit.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3487,6 +3515,8 @@ void CodegenLLVM::probereadDatastructElem(Value *src_data,
     // below triggering an assert in LLVM
     src_data = b_.CreateIntCast(src_data, b_.getInt64Ty(), false);
   }
+  else if (src_ty->isPointerTy())
```
BurntBrunch, line 3518: Note to other readers: 
This worked before this PR because the values were being passed around as integers. Now that this PR makes some of the values retain pointer types, we need explicit casts of the inputs to proberead.

BurntBrunch (commented):
This looks great! I only have a few concerns:
1) There's confusion between the bitness of the userspace process and the kernel, that needs a more considered approach.
2) There's a lot of implicit casts which can discard information now. I think these should be handled at the semantic analysis level and required to be explicit, as opposed to silently inserted into the IR, generating potentially unexpected data.

Additionally, I think 32-bit bpftrace on 64-bit kernel is an interesting (and easier!) test path for this work - if all goes well, that should work as seamlessly as 32-bit bpftrace on 32-bit kernel. 

The only way I can think of testing the latter would be running qemu inside CI. Which is certainly appealing, given how much random base kernel changes break CI!

> Unfortunately, with some hacks like [ffebc5c](https://github.com/iovisor/bpftrace/pull/2361/commits/ffebc5c876892cba3d99bce36affdb494f258c59) that were needed to avoid LLVM emitting incorrect BPF code (shifts on pointer registers; 32-bit spills of pointer registers to stack), this implementation ended up being quite involved too.

I wonder if we can gain more confidence in this approach by requiring that AST nodes with pointer types have corresponding llvm pointer types for their expr_ as well. Then, we can do this exhaustively (instead of cherrypicking what to fix for the test suite to pass) and the design is just as clean to reason about. WDYT?

src/ast/irbuilderbpf.cpp
```diff
@@ -1337,5 +1346,25 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
 #endif
 }
 
+llvm::Type *IRBuilderBPF::getPointerStorageTy()
+{
+  switch (sizeof(uintptr_t))
```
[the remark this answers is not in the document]
tnovak: Good catch! The case where bpftrace runs in compat mode on a 64-bit kernel isn't something I've considered, but it could be useful to support it -- or at least, as you noted, log an error and abort rather than producing wrong results.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2040,6 +2063,16 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
     // expr currently contains a pointer to the struct
     // and that's what we are saving
     AllocaInst *dst = b_.CreateAllocaBPF(map.type, map.ident + "_ptr");
+
+    auto expr_ty = expr->getType();
+    if (expr_ty->isIntegerTy() && expr_ty != b_.GetType(map.type))
```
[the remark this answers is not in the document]
tnovak: Yup -- for a given `LLVMContext`, there's only one instance of each distinct type (e.g. [Type::getInt64Ty](https://github.com/llvm/llvm-project/blob/7e22179d38c438fedb0d9bb0cff1585843bd7082/llvm/lib/IR/Type.cpp#L240)).

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2040,6 +2063,16 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
...
+    {
+      // Pointer assignment may require a cast if the integer expression has a
+      // different size. For example, in "@ = curtask" expr is always i64, but
+      // GetType(map.type) depends on the pointer size.
+      expr = b_.CreateIntCast(expr, b_.GetType(map.type), false);
```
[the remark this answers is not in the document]
tnovak: `curtask()` returns an `i64` (matches `bpf_get_current_task`'s signature) but this is another case where `CreateIntToPtr` might work well.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1906,6 +1918,17 @@ void CodegenLLVM::visit(Cast &cast)
...
   }
+  else if (cast.type.IsPtrTy() && expr_->getType()->isIntegerTy())
+  {
+    // If casting to a pointer and the size of the integer expression doesn't
+    // match the pointer size (e.g. "(struct Foo*)(uint8)42"), add an explicit
+    // cast.
+    auto pointer_ty = b_.GetType(cast.type);
+    if (expr_->getType() != pointer_ty)
+    {
+      expr_ = b_.CreateIntCast(expr_, pointer_ty, false, "cast");
```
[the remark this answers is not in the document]
tnovak: There are several places currently where `expr_` evaluates to an `i64` but may represent a pointer, for example `ctx()`, `reg()`, `sarg()` built-ins.

`ctx()` in particular must not be truncated to 32 bits (that would generate invalid BPF opcodes: shifts on the `PTR_TO_CTX` BPF register), but perhaps we could not cast it to an int at all (i.e. keep it as a pointer). Others could presumably be changed to return `i32`s on 32-bit platforms -- I'll test that to make sure this doesn't cause any new issues with BPF codegen.

tnovak:
Thanks for the detailed review, @BurntBrunch!

> I wonder if we can gain more confidence in this approach by requiring that AST nodes with pointer types have corresponding llvm pointer types for their expr_ as well. Then, we can do this exhaustively (instead of cherrypicking what to fix for the test suite to pass) and the design is just as clean to reason about. WDYT?

That seems like a good idea to me in general, but I'm not sure if it would be sufficient to eliminate all BPF codegen issues. :/ Regardless of whether `expr_` is a llvm pointer or integer type, there are cases where making the assigned-to variable 64-bit wide avoids incorrect BPF code. From one of the test cases:

```
struct A { int x[4]; }
uprobe:foo { 
    @a[0] = ((struct A *) arg0)->x;
    $x = @a[0]; 
    printf("Result: %d\n", $x[0]);
}
```

In `$x = @a[0]`, the RHS expression is a llvm pointer: `%lookup_elem_val = alloca [4 x i32], align 4`. However, if `$x` is 4 bytes long, in my testing LLVM 14 compiled `$x[0]` into invalid BPF code that uses LSH + RSH to truncate the `PTR_TO_STACK` register (`R10-n`, i.e. the value of `%lookup_elem_val`) to 32 bits before the load.

```
call bpf_map_lookup_elem#1

; copy 16 bytes (4 x i32) from the map to the BPF stack
r2 = *(u8 *)(r0 +9)      
[...]
*(u64 *)(r10 -32) = r2
*(u64 *)(r10 -40) = r4
[...]

; load 4 bytes from r10-40
r1 = r10
r1 += -40
r1 <<= 32                ; ugh
r1 >>= 32
r1 = *(u32 *)(r1 +0)     ; invalid mem access 'inv'

**Comment by BurntBrunch** [comment]
Okay, so that counter-example is useful to expose the fundamental confusion that I've been struggling to put my finger on. 

When instantiating the clang frontend and parsing types, we don't know what context we're using them in. This is clear if you consider a 64-bit kernel, a 32-bit bpftrace and a 64-bit uprobed program - you'll get the wrong types then (and if you don't, you would for a 32-bit uprobed program). This is really hard to untangle and would require parsing with multiple target triples to get right. In the current setup, we have to bail on any bitness mismatch - we need kernel, bpftrace and uprobed program to agree on bitness first and foremost.

That said, I think this _can_ be handled correctly. The more I think about it, the more it feels like approach 2 above would work more cleanly and potentially allow bitness mismatches down the road.

Ultimately, I really don't think we want bpf-internal llvm pointers in a 32-bit bpftrace to be 4 bytes long. After all, the 32-bitness only matters for probes into kernel/user memory and ctx access.  We might be able to abuse `AddressSpace` to express a distinction between pointers outside the program/maps and internal ones. If `llvm::AddressSpace` doesn't work, the internal `AddrSpace` is also an option (I'm not convinced `AddrSpace::none` is what we're looking for but it _might_ be).

If we keep everything in bpf land to 64-bit, then the boundary where we care about pointer size and type conversions is only the probes into kernel/user memory and ctx access. This, along with a strictly divided Addr(ess)Space for internal (maps, stack, etc) pointers, should make it much easier to gain confidence in the completeness and soundness of the type conversions.

Right now, it's hard to feel confident that the conversions happen in all the possible locations or that they chain correctly. Further, we're entrenching the bitness equality deep into codegen and making it hard to ever unwind it (which would be a useful feature for Android platforms for example).

I hate massive rewrites, so if you have thoughts on how to gain confidence that this work is complete and sound, I'd love to hear them! Ultimately, I can't prove it to myself, even with additions like enabling the 32-bit ALU instructions.

**Comment by tnovak** [comment]
> If we keep everything in bpf land to 64-bit, then the boundary where we care about pointer size and type conversions is only the probes into kernel/user memory and ctx access.

(y) Let me take another stab at approach 2. From my initial cursory exploration it seemed the surface area would still be fairly large, but I agree this would be cleaner (and shouldn't require workarounds for BPF codegen issues).

**Comment by tnovak** [comment]
I've rewritten this to keep the internal `Type::pointer` 8 bytes long, but use the correct pointer size when reading pointer types from kernel/user memory (approach 2 described in earlier comments). This simplified things quite a bit in the end (no llvm type mismatches to deal with and no BPF codegen workarounds).

File: src/ast/irbuilderbpf.cpp [hunk 8]
```diff
@@ -1340,5 +1353,89 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
...
+                                      bool isVolatile)
+{
+  llvm::Type *ptr_storage_ty = getPointerStorageTy(as);
+
+  if (!type.IsPtrTy() || ptr_storage_ty == getInt64Ty())
+    return CreateLoad(GetType(type), ptr, isVolatile);
+
+  assert(GetType(type) == getInt64Ty());
+
+  // Pointer size for the given address space doesn't match the internal
```
BurntBrunch, line 1397: s/internal/bpf-side/

src/ast/passes/codegen_llvm.cpp
```diff
@@ -264,17 +264,22 @@ void CodegenLLVM::visit(Builtin &builtin)
...
+    Value *sp = b_.CreateRegisterRead(ctx_, sp_offset, "reg_sp");
     AllocaInst *dst = b_.CreateAllocaBPF(builtin.type, builtin.ident);
-    Value *src = b_.CreateAdd(sp,
-                              b_.getInt64((arg_num + arch::arg_stack_offset()) *
-                                          sizeof(uintptr_t)));
-    b_.CreateProbeRead(ctx_, dst, 8, src, builtin.type.GetAS(), builtin.loc);
+
+    // Pointer width is used when calculating the SP offset and the number of
+    // bytes to read from stack for each argument (e.g. on 32-bit systems sargN
+    // 4 bytes). We pass a pointer SizedType to CreateProbeRead to make sure it
```
BurntBrunch, line 272: The part in the parentheses needs rewording but also the formula is right there in the code so the example is redundant

src/ast/passes/codegen_llvm.cpp
```diff
@@ -705,8 +706,14 @@ void CodegenLLVM::visit(Call &call)
                        b_.CreateGEP(b_.getInt8Ty(), perfdata, b_.getInt64(8)),
                        b_.getInt64Ty()->getPointerTo()));
     join_id_++;
+
+    SizedType elem_type = CreatePointer(CreateInt8(), addrspace);
+    size_t ptr_width = b_.getPointerStorageTy(addrspace)->getIntegerBitWidth();
+    assert(b_.GetType(elem_type) == b_.getInt64Ty());
```
BurntBrunch, line 712: Kind of a weird assert to throw in here but sure

BurntBrunch (approved):
Looks good to me, thanks for taking the time to isolate the changes in this way. I think I can convince myself way more reliably that this is in fact correct.

We really need to exercise this code in CI, in 32-bit mode, but that's for another PR.

Approved with minor nits inline.

src/ast/irbuilderbpf.cpp
```diff
@@ -1340,5 +1353,89 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
...
+  // Pointers are internally always represented as 64-bit integers, matching the
+  // BPF register size (BPF is a 64-bit ISA). This helps to avoid BPF codegen
+  // issues such as truncating PTR_TO_STACK registers using shift operations,
+  // which is disallowed (see https://github.com/iovisor/bpftrace/pull/2361).
+  // However, when reading pointers from kernel or user memory, we need to use
+  // the appropriate size for the target system.
+  const size_t ptr_size = getPointerStorageTy(as)->getIntegerBitWidth() / 8;
+
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+  // TODO: support 32-bit big-endian systems
```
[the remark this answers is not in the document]
tnovak: AFAICT big-endian powerpc is one of the supported configurations.

The "32-bit" part is currently a runtime check; we could have a static assert on the `uinptr_t` size, but that wouldn't cover the case where bpftrace is a 32-bit binary running on a 64-bit kernel (which is supported and wouldn't trip the runtime assert below). I'd lean towards not breaking the build on the off chance that someone builds it that way, but don't feel strongly about it.

src/ast/irbuilderbpf.cpp
```diff
@@ -1340,5 +1353,89 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
...
+llvm::Type *IRBuilderBPF::getUserPointerStorageTy()
+{
+  // TODO: we don't currently have an easy way of determining the pointer size
+  // of the uprobed process, so assume it's the same as the kernel's for now.
+  return getKernelPointerStorageTy();
+}
+
+llvm::Type *IRBuilderBPF::getRegisterTy()
+{
+  // Bitwidth of registers values in struct pt_regs is the same as the kernel
```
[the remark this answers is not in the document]
tnovak: I had this as a separate function for readability, but since it's local to `IRBuilderBPF` anyway I'll switch it to `getKernelPointerStorageTy`.

src/ast/irbuilderbpf.cpp
```diff
@@ -1340,5 +1355,82 @@ StoreInst *IRBuilderBPF::createAlignedStore(Value *val,
 #endif
 }
 
+void IRBuilderBPF::CreateProbeRead(Value *ctx,
+                                   Value *dst,
+                                   const SizedType &type,
+                                   Value *src,
+                                   AddrSpace as,
```
ajor, line 1362: We don't need to take address space as an argument since it can just be derived from type (and that's what's happening at every call site now anyway). Same for `CreateLoad`

ajor (commented):
This looks pretty good overall, definitely easier to understand.

One thing that I'm a little concerned about though is people using the wrong version of CreateProbeRead / CreateLoad in the future.

For CreateProbeRead, I'm wondering if we should get rid of the version that takes `size_t size` and move them all over to this new version. Almost all the call sites are just doing `SizedType.GetSize()` and you've already converted most of the rest.

src/ast/irbuilderbpf.cpp
tnovak: In `probereadDatastructElem()`, the type passed to `CreateProbeRead()` is `elem_type`, but `AddrSpace` comes from `data_type`. Let me verify that both have the same `AddrSpace`.

tnovak: There are currently cases where the type's `AddrSpace` isn't set (e.g. fields of a struct have it set to `none`, while the struct's AS is `user` or `kernel`) and we need to pass in a different AS; for now I've kept the extra argument, but made it optional (defaults to `type.GetAS()`) and added a comment.

ajor (approved):

PR #2363, opened 2022-09-21
Fix probe builtin probe expansion bug

### LINKED ISSUE #2362: probe builtin breaks interval and profile probes
When using the `probe` bultin within an interval or profile probe, the program fails to compile:
```
# bpftrace -e 'interval:ms:1000 { print(probe); }'
ERROR: Failed to compile: Wildcard matches aren't available on probe type 'interval'
# bpftrace -e 'profile:ms:1000 { print(probe); }'
ERROR: Failed to compile: Wildcard matches aren't available on probe type 'profile'
```
This happens because SemanticAnalyzer always sets `need_expansion` when `probe` builtin is encountered, regardless of whether it actually contains a wildcard or not.

lenticularis39:
Expand probes using probe builtin only when they contain a wildcard. This prevents an error in cases where the probe builtin is used in a probe that cannot be expanded, e.g. interval probe.

Fixes #2362

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -408,7 +408,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
   }
   else if (builtin.ident == "probe") {
     builtin.type = CreateProbe();
-    probe_->need_expansion = true;
+    if (has_wildcard(probe_->name()))
+      probe_->need_expansion = true;
   }
   else if (builtin.ident == "username") {
     builtin.type = CreateUsername();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to
 #### Deprecated
 #### Removed
 #### Fixed
+- Fix probe builtin probe expansion bug
+  - [#2363](https://github.com/iovisor/bpftrace/pull/2363)
+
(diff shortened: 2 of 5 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -408,7 +408,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
   }
   else if (builtin.ident == "probe") {
     builtin.type = CreateProbe();
-    probe_->need_expansion = true;
+    if (has_wildcard(probe_->name()))
+      probe_->need_expansion = true;
```
viktormalik, line 412: This doesn't feel like a sufficient solution. Compare:
```
(master) # bpftrace -e 'kprobe:vfs_read,kprobe:vfs_open { print(probe); exit(); }'
Attaching 2 probes...
kprobe:vfs_open

(probe-builtin-expansion) # bpftrace -e 'kprobe:vfs_read,kprobe:vfs_open { print(probe); exit(); }'
WARNING: Symbols not found: [ kprobe:vfs_read ]
Attaching 2 probes...
kprobe:vfs_read,kprobe:vfs_open
```
(this problem is also reflected by the codegen test change)

viktormalik (changes requested):

Issue #2334, opened 2022-08-19
libbpf as a main part of program loading


BurntBrunch:
This is a first draft of the refactoring steps discussed in #2322.

The overall goal is to delegate more of the work to libbpf and gradually stop dealing with details that libbpf can already handle.

## Program and map loading

1. Emit in-memory ELF and build the BpfBytecode map out of it via ELF
1.1. This step remove the ORC usage but keeps the current bag-of-sections approach.
3. Emit global variables for all the maps in the libbpf format, which requires BTF. As I found out in #2321, this requires LLVM 10.
4. Import libbpf 0.8.1 as a git submodule
  3.1 Configure local builds to use it by default.
  3.2 Create PACKAGING.md and document the minimum LLVM and libbpf versions. 
    3.2.1 Optionally document every single llvm dependency (individual library names)
5. Use `bpf_object__open_mem` to iterate over the programs but pass the instructions directly to the syscall
  4.1. Remove the non-`bpf_prog_load` paths around this time
  4.2. Ensure we call `bpf_program__set_autoattach(false)` on all programs
  4.3. If necessary (?), call `bpf_program__set_expected_attach_type`
6. Be able to cleanly `bpf_object__load` the programs and maps

At this point, program relocation and loading happens with libbpf and we can use `bpf_program__fd` to pass the resulting fd into the existing attachment infrastructure.
An interesting angle here is that an external program can at this point easily load the output of `--emit-elf`. This might be interesting for AOT though I haven't gone through that code yet.

## Program attachment

From here on out, the path is roughly:
  * For each program type:
    * Use libbpf to actually attach the program via `bpf_program__attach*`
    * (Optionally) Migrate to the libbpf naming convention for that attachment type and use autoattach
  * In terms of new features, at this point we can potentially use the ksyscall logic to automatically attach to the right architecture-specific syscall handler via kfunc (instead of tracepoints).

### USDT suport

The biggest question mark in my mind is the USDT support in libbpf. While libbpf has helpers we can't really use for access to usdt params in BPF, it also requires a ["usdt manager" to be active](https://github.com/libbpf/libbpf/blob/master/src/usdt.c#L24).

Overall, this area needs more investigation - it might be the case that bpftrace itself would embed this BPF-side logic and load it when it encounters usdt probes. Or, it might be the case that bpftrace doesn't fit into this model of working with USDT.

anakryiko:
> pass the instructions directly to the syscall

and 

[...]

PR #2370, opened 2022-09-26
Clean up unnecessary symbol existence checks

BurntBrunch:
(This is based on top of #2369, look at the libbpf: and libbcc: commits only in this PR)
(Part of #2334 work)

Now that we have strict requirements for minimum library versions, this PR removes unnecessary libbpf and bcc symbol checks in the bpftrace codebase. Now that these are cleaned up, we may even be able to set up libbpf and bcc as ExternalProjects, if we so desired (though I'd personally prefer to keep them as git submodules).


BurntBrunch:
Alright, I *think* found a fix for the flaky usdt test. For reasons I don't understand, I was only able to fix it by writing a new C binary that forks and execs the necessary two instances. Doing so from bash made the BEFORE phase time out (wtf GH Actions; I suspect some layer of security hooks or auditing slowing things down). Doing so from a bash script doesn't work because `pidof` will see `bash` as the name.

viktormalik (approved):
LGTM, just one thing comes to mind: are we documenting the minimum required libbpf and bcc versions somewhere?

BurntBrunch:
> are we documenting the minimum required libbpf and bcc versions somewhere

I'll document it in INSTALL.md and README.md and send a separate PR

PR #2375, opened 2022-09-28
Support static linking with Android NDK toolchain

tnovak:
A few minor tweaks to make -DSTATIC_LINKING work when cross-compiling for Android:

- remove -lrt and -lpthread from linker flags; those libraries don't exist on Android (libc provides the relevant functionality)
- libz.so is provided by the platform so link against it dynamically
- use libclang_static (defined in ClangTargets.cmake) in target_link_libraries() instead of libclang.a; the latter (without the full path) was causing cmake to not be able to find the static lib in my [ExtendedAndroidTools](https://github.com/facebookexperimental/ExtendedAndroidTools) builds and to link with libclang.so instead

##### Checklist


src/ast/CMakeLists.txt
```diff
@@ -58,6 +57,9 @@ if (STATIC_LINKING)
     target_link_libraries(ast "-Wl,--start-group" ${CLANG_EMBEDDED_CMAKE_TARGETS} ${LLVM_EMBEDDED_CMAKE_TARGETS} "-Wl,--end-group")
   else()
     llvm_map_components_to_libnames(llvm_libs ${llvm_lib_names})
+
+    find_package(LibClang_Static)
```
BurntBrunch, line 61: This feels like a bug in your setup. Given that STATIC_LINKING changes the suffix to .a already, getting libclang.so here is making me think something is wrong with your prefix or your paths. 

Try `cmake --debug-find -DSTATIC_LINKING=ON .. |& less` to figure out where it's going wrong maybe?
Ultimately, we should not be hardcoding llvm internal names in this way and should be using the LLVM-Config.cmake macros instead. Maybe we need judicious use of llvm_config macro to do the right thing here?

tnovak: @BurntBrunch: in the initial revision I used `libclang_static` as the target name, which is defined in `ClangTargets.cmake`. However, this broke one of the embedded CI builds (llvm_clang, Alpine, llvm9) -- presumably because there's no `ClangTargets.cmake` in that setup (and llvm/clang libs are found in the default library search path).

I'm not very familiar with cmake internals, but the docs for [target_link_libraries](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) suggest that the name should be a previously-defined cmake target, a plain name (e.g. `clang`) or a full path; not sure how it works with `libclang.a` as written here originally.

Based on the suffix being set to .a, I'd expect just using `clang` instead of `libclang.a` to work, but that didn't seem to be sufficient in my setup. Let me check again with `--debug-find`.

tnovak: > However, this broke one of the embedded CI builds (llvm_clang, Alpine, llvm9) -- presumably because there's no ClangTargets.cmake in that setup (and llvm/clang libs are found in the default library search path).

I checked the clang-dev-9.0.0-r1 package in alpine -- there's no libclang_static target in ClangTargets.cmake there (it must've been added after 9.0).

tnovak:
Reverted back to using `libclang_static` (name of the target defined in `ClangTargets.cmake` when built with `LIBCLANG_BUILD_STATIC` set), with the fallback to `libclang.a` for LLVM 9.

src/ast/CMakeLists.txt
```diff
@@ -58,6 +57,15 @@ if (STATIC_LINKING)
     target_link_libraries(ast "-Wl,--start-group" ${CLANG_EMBEDDED_CMAKE_TARGETS} ${LLVM_EMBEDDED_CMAKE_TARGETS} "-Wl,--end-group")
   else()
     llvm_map_components_to_libnames(llvm_libs ${llvm_lib_names})
+
+    if(TARGET libclang_static)
```
BurntBrunch, line 61: That makes things a lot cleaner, thanks!

cmake/FindLibBcc.cmake
```diff
@@ -60,8 +60,15 @@ if(STATIC_LINKING)
   # If we do a static bpftrace build, we must link them in.
   find_package(LibBpf)
   find_package(LibElf)
-  find_package(LibZ)
-  SET(CMAKE_REQUIRED_LIBRARIES ${LIBBCC_BPF_LIBRARIES} ${LIBBPF_LIBRARIES} ${LIBELF_LIBRARIES} ${LIBZ_LIBRARIES})
+
+  if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
```
BurntBrunch, line 64: I don't think you have to interpolate this.
`if (CMAKE_SYSTEM_NAME STREQUAL "Android")` should work?
Even better, the ndk toolchain defines `ANDROID`, so `if (ANDROID)` should be enough.

src/CMakeLists.txt
```diff
@@ -204,10 +204,15 @@ if (USE_LIBFUZZER)
 endif()
 
 if (STATIC_LINKING)
-  target_link_libraries(libbpftrace "-Wl,-Bdynamic" "-lrt" "-lpthread" "-ldl" "-lm")
-  target_link_libraries(libbpftrace "-Wl,-Bstatic" "-lz")
-  target_link_libraries(runtime "-Wl,-Bdynamic" "-lrt" "-lpthread" "-ldl" "-lm")
-  target_link_libraries(runtime "-Wl,-Bstatic" "-lz")
+  if(${CMAKE_SYSTEM_NAME} STREQUAL "Android")
```
BurntBrunch, line 207: Same as above

BurntBrunch (approved):
This looks pretty straight-forward! Minor nits about Android conditionals

ajor (approved):

PR #2376, opened 2022-09-28
Remove MCJIT/bpforc usage

BurntBrunch:
As part of #2334, this PR removes the Orc usage entirely and replaces it with <50 lines of in-memory ELF parsing. We produce in-memory ELF files, parse them into the bag-of-sections and carry on as before.

In the process, we also unify the `CodegenLLVM::emit_elf` and `CodegenLLVM::emit` paths to do the exact same thing but to different streams.

The middleware optimization passes remain the same, we just change the final backend code - the codegen fixtures are all the same.


fbs (approved):
LGTM!

viktormalik (approved):

PR #2379, opened 2022-10-01
tcpdrop: Fix: ERROR: Error attaching probe: 'kprobe:tcp_drop'

Rtoax:
kernel commit 8fbf195798b5('tcp_drop() is no longer needed.') remove the kprobe:tcp_drop, bcc commit 16eab39171eb('Add
tracepoint:skb:kfree_skb if no tcp_drop() kprobe.') already fix this problem.

CI old kernel is too old and not support the 'reason' field, move the old tools/tcpdrop.bt into tools/old/tcpdrop.bt and set the CI to use it.

ERROR log:
```bash
 $ sudo ./tcpdrop.bt
 ./tcpdrop.bt:49-51: WARNING: tcp_drop is not traceable (either non-existing, inlined, or marked as "notrace"); attaching to it will likely fail
 Attaching 3 probes...
 ERROR: Error attaching probe: 'kprobe:tcp_drop'
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
           LLVM_VERSION: 10
           RUN_ALL_TESTS: 1
           RUNTIME_TEST_DISABLE: probe.kprobe_offset_fail_size
-          TOOLS_TEST_OLDVERSION: biosnoop.bt
+          TOOLS_TEST_OLDVERSION: biosnoop.bt tcpdrop.bt
           BASE: focal
           VENDOR_GTEST: ON
         - NAME: LLVM 10 Release
@@ -42,7 +42,7 @@ jobs:
           LLVM_VERSION: 10
           RUN_ALL_TESTS: 1
           RUNTIME_TEST_DISABLE: probe.kprobe_offset_fail_size
-          TOOLS_TEST_OLDVERSION: biosnoop.bt
+          TOOLS_TEST_OLDVERSION: biosnoop.bt tcpdrop.bt
           BASE: focal
           VENDOR_GTEST: ON
         - NAME: LLVM 10 Clang Debug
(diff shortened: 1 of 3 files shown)


viktormalik:
The tool is crashing in the CI, can you debug it?

Rtoax:
> The tool is crashing in the CI, can you debug it?

OK, i'll try to fix that.

viktormalik:
> Hi, @viktormalik , My env is kernel 5.15 on centos-stream9, CI env has a wrong kernel version? should i fix CI kernel version, and how to modify it?

Hi, unfortunately, we cannot control the CI kernel (yet). The likely problem here is that the kernel is too old and the `reason` field, as well as the corresponding enum, don't exist. The preferred way is to make the tool as generic as possible, but if that's not an option, we can have multiple implementations of the tool - just move the old one into `tools/old` and set the CI to use it (see e.g. `biosnoop.bt`).

Rtoax:
Is this done right?

viktormalik:
> @viktormalik Hi, Is this done right?

Yes, but the comments in the tools headers need to be adjusted. Please add a specific kernel version from which the new version is necessary.

Rtoax:
@viktormalik Hi, i specific kernel version in comment.

Useful hint:

```
Since 5.17 support trace_kfree_skb(skb, ..., reason) 'reason' field.
Since 5.19 remove tcp_drop() function.
```

tools/tcpdrop.bt
```diff
@@ -10,15 +10,18 @@
  *
  * This provides information such as packet details, socket state, and kernel
  * stack trace for packets/segments that were dropped via tcp_drop().
-
+ *
  * WARNING: this script attaches to the tcp_drop kprobe which is likely inlined
  *          on newer kernels and not replaced by anything else, therefore
  *          the script will stop working
```
viktormalik, line 16: This should go away in the new version of the script.

tools/tcpdrop.bt
```diff
@@ -10,15 +10,18 @@
  *
  * This provides information such as packet details, socket state, and kernel
  * stack trace for packets/segments that were dropped via tcp_drop().
```
viktormalik, line 12: `tcp_drop()` is not called in the new version so the description should be updated.

viktormalik (changes requested):

viktormalik (approved):

PR #2381, opened 2022-10-05
Fix kprobe multi-attachment

viktormalik:
Tests for kprobe multi-attachment are not executed in the CI, so we missed a couple of bugs. This PR fixes them (see individual commits for details).

Also, some runtime tests (including those for kprobe multi) require bpftool, so it is now installed in Docker. Unfortunately, kprobe_multi is stilll not available in the CI kernel.

##### Checklist

### DIFF
--- a/docker/Dockerfile.focal
+++ b/docker/Dockerfile.focal
@@ -41,6 +41,7 @@ RUN apt-get update && apt-get install -y \
       llvm-${LLVM_VERSION}-dev \
       llvm-${LLVM_VERSION}-runtime \
       libllvm${LLVM_VERSION} \
+      linux-tools-$(uname -r) \
       systemtap-sdt-dev \
       python3 \
       xxd \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to
 #### Fixed
 - Fix pointer/register loads on 32-bit architectures
   - [#2361](https://github.com/iovisor/bpftrace/pull/2361)
+- Fix kprobe multi-attachment
+  - [#2381](https://github.com/iovisor/bpftrace/pull/2381)
+
 #### Docs
 #### Tools
 
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -835,8 +835,7 @@ void AttachedProbe::attach_multi_kprobe(void)
   opts.kprobe_multi.syms = syms.data();
   opts.kprobe_multi.cnt = syms.size();
   opts.kprobe_multi.flags = probe_.type == ProbeType::kretprobe
-                                ? static_cast<enum ::bpf_attach_type>(
-                                      libbpf::BPF_TRACE_KPROBE_MULTI)
+                                ? BPF_F_KPROBE_MULTI_RETURN
                                 : 0;
 
   if (bt_verbose)
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -138,7 +138,7 @@ int BPFtrace::add_probe(ast::Probe &p)
       attach_funcs.insert(attach_funcs.end(), matches.begin(), matches.end());
 
       if (feature_->has_kprobe_multi() && has_wildcard(attach_point->func) &&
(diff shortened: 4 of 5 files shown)


BurntBrunch (approved):
Looks pretty straight-forward.

docker/Dockerfile.focal
```diff
@@ -41,6 +41,7 @@ RUN apt-get update && apt-get install -y \
       llvm-${LLVM_VERSION}-dev \
       llvm-${LLVM_VERSION}-runtime \
       libllvm${LLVM_VERSION} \
+      linux-tools-$(uname -r) \
```
BurntBrunch, line 44: @viktormalik this actually breaks this docker file for non-ubuntu kernels, which makes the dockerfile useless for local development on e.g. Fedora. There's also no guarantee that a package for the host kernel would exist in the ubuntu repo for cross-distro situations. 

E.g.:
```
E: Unable to locate package linux-tools-5.19.16-200.fc36.x86_64
E: Couldn't find any package by glob 'linux-tools-5.19.16-200.fc36.x86_64'
E: Couldn't find any package by regex 'linux-tools-5.19.16-200.fc36.x86_64'
```

> A version corresponding to the host kernel must be installed.

I'm not sure this is true. bpftool only uses syscalls and libbpf, so it should fall back reasonably well.

Speaking of bpftool, I've noticed tests failing on Fedora because they hardcode old bpftool text output - this is clearly a problem for reproducible tests. The approach in a lot of libbpf-related projects is to vendor bpftool and honestly I think that's the only viable option if we want to rely on the not-meant-for-machines text output.

Ultimately, I think aiming for less dependence on the environment would help us with updates going forward. (I'm looking at you Alpine static build)

viktormalik: > @viktormalik this actually breaks this docker file for non-ubuntu kernels, which makes the dockerfile useless for local development on e.g. Fedora. 

Well, since the file is called `Dockerfile.focal`, I wouldn't expect it to work on anything else than Ubuntu Focal ;-) It certainly won't work on Fedora as it uses `apt` (we have `Dockerfile.fedora30` which is probably very outdated).

> Speaking of bpftool, I've noticed tests failing on Fedora because they hardcode old bpftool text output - this is clearly a problem for reproducible tests. The approach in a lot of libbpf-related projects is to vendor bpftool and honestly I think that's the only viable option if we want to rely on the not-meant-for-machines text output.

Yeah, vendoring bpftool in tests seems like the only viable option. I'm completely fine with that, it's good to have the test env under control as much as possible.

I'm also thinking if it'd make sense to add some support for running bpftool in tests. Something like instead of:
```
RUN {{BPFTRACE}} --unsafe -e 'kretprobe:ksys_* { system("/usr/sbin/bpftool prog | grep kprobe | grep ksys_ | wc -l"); exit(); }'
EXPECT 1
```
we'd have
```
RUN {{BPFTRACE}} --unsafe -e 'kretprobe:ksys_* { {{BPFTOOL_TEST}} exit(); }'
BPFTOOL_TEST {{BPFTOOL}} prog | grep kprobe | grep ksys | wc -l
EXPECT 1
```

which is at least more readable and we'll be able to, e.g., change the bpftool binary without affecting the tests. Still, it'll require some work so I'm not sure if it's worth it. Atm, we have 5 usages of bpftool throughout the tests, but I'd expect more to come.

> Ultimately, I think aiming for less dependence on the environment would help us with updates going forward. (I'm looking at you Alpine static build)

+1

BurntBrunch: > Well, since the file is called Dockerfile.focal, I wouldn't expect it to work on anything else than Ubuntu Focal ;-) It certainly won't work on Fedora as it uses apt (we have Dockerfile.fedora30 which is probably very outdated).

That's not what I mean. It's a docker container with a focal userspace but when instantiated on a fedora _kernel_, it fails because the uname propagates to the container. Notice the fc part in `linux-tools-5.19.16-200.fc36.x86_64`, that's coming from the `uname -r`, which talks to the fedora kernel. It doesn't make sense to me that the container would also require the host to be the same distro.

PR #2382, opened 2022-10-05
Codegen: initialize LLVM targets

viktormalik:
The initialization is required by the new in-memory ELF parsing that has replaced ORC usage (#2376). At this point, it is executed inside `clang_createIndex` but if ClangParser did not run for some reason (as #2315 proposes), codegen would fail.

This PR does the initialization transparently from codegen.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -24,6 +24,8 @@
 #include <llvm/Support/TargetRegistry.h>
 #endif
 
+#include <llvm/Support/TargetSelect.h>
+
 #include "arch/arch.h"
 #include "ast.h"
 #include "ast/async_event_types.h"
@@ -45,6 +47,9 @@ CodegenLLVM::CodegenLLVM(Node *root, BPFtrace &bpftrace)
       module_(std::make_unique<Module>("bpftrace", *context_)),
       b_(*context_, *module_, bpftrace)
 {
+  llvm::InitializeAllTargets();
+  llvm::InitializeAllTargetMCs();
+  llvm::InitializeAllAsmPrinters();
   std::string error_str;
   auto target = llvm::TargetRegistry::lookupTarget(LLVMTargetTriple, error_str);
   if (!target)


tnovak:
> At this point, it is executed inside `clang_createIndex` but if ClangParser did not run for some reason (as https://github.com/iovisor/bpftrace/pull/2315 proposes), codegen would fail.

It can also fail if LLVM is _statically_ linked, but libclang isn't (this is how [ExtendedAndroidTools](https://github.com/facebookexperimental/ExtendedAndroidTools) builds bpftrace, at least until #2375 is merged). In that case `llvm::TargetRegistry` populated via `clang_createIndex` is the one local to `libclang.so`, and not visible to the rest of bpftrace code. 

I hit this issue today and `error_str` was "Unable to find target for this triple (no targets are registered)". Thanks for fixing it!

BurntBrunch:
Hmm, is it possible at all to selectively initialize only the bpf target? Anything with "all" in the name sounds like a startup regression to me :) 

Also, any suggestions on how to extend CI to catch this?

viktormalik:
> Hmm, is it possible at all to selectively initialize only the bpf target? Anything with "all" in the name sounds like a startup regression to me :)

I wondered the same, but it seems to me that `InitializeAll...` is the preferred way to go. There are `LLVMInitializeBPF...` functions but those are not exposed by the LLVM API.

> Also, any suggestions on how to extend CI to catch this?

#2315 will have tests that skip ClangParser and hence need this to work.

BurntBrunch (approved):
Sounds good! 
Somehow this didn't trigger the full CI here but it has all the checks in your repo.

tnovak:
> Hmm, is it possible at all to selectively initialize only the bpf target? Anything with "all" in the name sounds like a startup regression to me :)

Initializing all targets also increases the size of the `bpftrace` binary quite a bit when linking statically.

Switching to `LLVMInitializeBPF*` calls cuts the size of my armv7 build from ~91MB to ~45MB. Note that this also required modifying `libclang` (since `CIndex.cpp` also initializes all targets), so it's unfortunately not something we can do entirely in bpftrace. :(

Issue #2284, opened 2022-06-30
Race condition in `usym` resolution


viktormalik:
Originally discovered by @xutao323 in [this comment](https://github.com/iovisor/bpftrace/issues/2118#issuecomment-1017188844).

As `usym` is resolved asynchronously, a race condition may occur in case the traced binary exits before its symbols are resolved. This causes, e.g., the `func` builtin to return address instead of the actual function name.

### What reproduces the bug?

Usually can be reproduced using `func` inside an `uprobe/uretprobe` for a function close to the end of the traced binary:

```
# cat test.c
int fun() { }

int main() {
    fun();
}

# bpftrace -e 'uprobe:test:fun { printf("%s\n", func); }
Attaching 1 probe...
0x401106
```

Inserting a `sleep` fixes the problem:
```
# cat test.c
#include <unistd.h>
int fun() { sleep(1); }

int main() {
    fun();
}

# bpftrace -e 'uprobe:test:fun { printf("%s\n", func); }
Attaching 1 probe...
fun
```

Another way to reproduce is to run bpftrace in a debugger.

jordalgo:
Is this still an issue?

viktormalik:
I believe that we managed to solve the issue in most of the situations. There are still a few where we don't get the symbol name but I don't think that we can do much about them.

@lenticularis39 knows better, though, so if he can confirm my thoughts, I'd suggest closing this issue.

lenticularis39:
There are still cases where usym resolution fails but IMO these should be tracked in separate, specific issues if they turn out to be problematic. 

I agree with closing this issue, if anyone hits a race condition regarding usym resolution, please open a new issue and describe the exact situation where it happened.

PR #2386, opened 2022-10-11
Improve user symbol resolution

lenticularis39:
This PR contains three related improvements of usym resolution (edit: added purpose of the changes):
- Caching BCC symcache per process if ASLR is enabled (edit: optional after changes, by default disabled)
  - Enables faster symbol resolution and symcache preloading (see below)
- Resolving symbols directly from ELF (only if ASLR is disabled)
  - Enables resolving symbols that are not dynamically loaded even after process exit
- Preloading BCC symcache for running instances of program
  - Enables resolving symbols after program exists even with ASLR enabled, given that the process exists at probe attachment time.

The latter two provide a partial fix of #2284. 

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -210,12 +210,13 @@ Increasing the value will consume more memory, increase startup times and can in
 
 === BPFTRACE_CACHE_USER_SYMBOLS
 
-Default: 0 if ASLR is enabled on system and `-c` option is not given; otherwise 1
+Default: PER_PROGRAM if ASLR disabled or `-c` option given, PER_PID otherwise.
 
-By default, bpftrace caches the results of symbols resolutions only when ASLR (Address Space Layout Randomization) is disabled.
-This is because the symbol addresses change with each execution with ASLR.
-However, disabling caching may incur some performance penalty.
-Set this env variable to 1 to force bpftrace to cache.
+Caching strategy for user symbols. Valid values are:
+
+* PER_PROGRAM - each program has its own cache. If there are more processes with enabled ASLR for a single program, this might produce incorrect results.
+* PER_PID - each process has its own cache. This is accurate for processes with ASLR enabled, and enables bpftrace to preload caches for processes running at probe attachement time. If there are many processes running, it will consume a lot of a memory.
+* NONE - caching disabled. This saves the most memory, but at the cost of speed.
 
 === BPFTRACE_VMLINUX
 
@@ -1788,7 +1789,7 @@ The pointer type is left unchanged.
 
 Equal to <<functions_ksym>> but resolves user space symbols.
 
-If ASLR is enabled, it works only when the process is running at either the time of the symbol resolution or the time of the probe attachment. The latter might not work with older versions of BCC. A similar limitation also applies to dynamically loaded symbols.
+If ASLR is enabled, user space symbolication only works when the process is running at either the time of the symbol resolution or the time of the probe attachment. The latter requires `BPFTRACE_CACHE_USER_SYMBOLS` to be set to `PER_PID`, and might not work with older versions of BCC. A similar limitation also applies to dynamically loaded symbols.
 
 ----
 uprobe:/bin/bash:readline
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1786,8 +1786,9 @@ The pointer type is left unchanged.
 * uretprobes
 
 
-Equal to <<functions_ksym>> but resolves user space symbols
+Equal to <<functions_ksym>> but resolves user space symbols.
 
+If ASLR is enabled, it works only when the process is running at either the time of the symbol resolution or the time of the probe attachment. The latter might not work with older versions of BCC. A similar limitation also applies to dynamically loaded symbols.
 
 ----
 uprobe:/bin/bash:readline
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -140,15 +140,50 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
   // Kernel stacks should not be differentiated by tid, since the kernel
   // address space is the same between pids (and when aggregating you *want*
   // to be able to correlate between pids in most cases). User-space stacks
-  // are special because of ASLR and so we do usym()-style packing.
+  // are special because of ASLR, hence we also store the pid; probe id is
+  // stored for cases when only ELF resolution works (e.g. ASLR disabled and
+  // process exited).
   if (ident == "ustack")
   {
-    // pack uint64_t with: (uint32_t)stack_id, (uint32_t)pid
-    Value *pidhigh = b_.CreateShl(b_.CreateGetPidTgid(loc), 32);
-    stackid = b_.CreateOr(stackid, pidhigh);
-  }
+    std::vector<llvm::Type *> elements = {
+      b_.getInt64Ty(), // stack id
+      b_.getInt32Ty(), // pid
+      b_.getInt32Ty(), // probe id
+    };
+    StructType *stack_struct = b_.GetStructType("stack_t", elements, false);
+    AllocaInst *buf = b_.CreateAllocaBPF(stack_struct, "stack_args");
 
-  expr_ = stackid;
+    // store stack id
+    b_.CreateStore(
+        stackid,
+        b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(0) }));
+    // store pid
+    b_.CreateStore(
+        b_.CreateTrunc(b_.CreateGetPidTgid(loc), b_.getInt32Ty()),
+        b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
+    // store probe id
+    b_.CreateStore(
+        b_.GetIntSameSize(get_probe_id(), elements.at(2)),
+        b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(2) }));
+
+    expr_ = buf;
+  }
+  else
+  {
+    expr_ = stackid;
+  }
+}
+
+int CodegenLLVM::get_probe_id()
+{
+  auto begin = bpftrace_.resources.probe_ids.begin();
+  auto end = bpftrace_.resources.probe_ids.end();
+  auto found = std::find(begin, end, probefull_);
+  if (found == end)
+  {
+    bpftrace_.resources.probe_ids.push_back(probefull_);
+  }
+  return std::distance(begin, found);
 }
 
 void CodegenLLVM::visit(Builtin &builtin)
@@ -264,7 +299,7 @@ void CodegenLLVM::visit(Builtin &builtin)
 
     if (builtin.type.IsUsymTy())
     {
-      expr_ = b_.CreateUSym(expr_, builtin.loc);
+      expr_ = b_.CreateUSym(expr_, get_probe_id(), builtin.loc);
       Value *expr = expr_;
       expr_deleter_ = [this, expr]() { b_.CreateLifetimeEnd(expr); };
     }
@@ -299,13 +334,7 @@ void CodegenLLVM::visit(Builtin &builtin)
   }
   else if (builtin.ident == "probe")
   {
-    auto begin = bpftrace_.resources.probe_ids.begin();
-    auto end = bpftrace_.resources.probe_ids.end();
-    auto found = std::find(begin, end, probefull_);
-    builtin.probe_id = std::distance(begin, found);
-    if (found == end) {
-      bpftrace_.resources.probe_ids.push_back(probefull_);
-    }
+    builtin.probe_id = get_probe_id();
     expr_ = b_.getInt64(builtin.probe_id);
   }
   else if (builtin.ident == "args" || builtin.ident == "ctx")
@@ -774,7 +803,7 @@ void CodegenLLVM::visit(Call &call)
   else if (call.func == "usym")
   {
     auto scoped_del = accept(call.vargs->front());
-    expr_ = b_.CreateUSym(expr_, call.loc);
+    expr_ = b_.CreateUSym(expr_, get_probe_id(), call.loc);
   }
   else if (call.func == "ntop")
   {
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -65,6 +65,13 @@ void ResourceAnalyser::visit(Builtin &builtin)
   {
     resources_.stackid_maps.insert(StackType{});
   }
+
+  if (uses_usym_table(builtin.ident))
+  {
+    // mark probe as using usym, so that the symbol table can be pre-loaded
+    // and symbols resolved even when unavailable at resolution time
+    resources_.probes_using_usym.insert(probe_);
+  }
 }
 
 void ResourceAnalyser::visit(Call &call)
@@ -194,6 +201,13 @@ void ResourceAnalyser::visit(Call &call)
     resources_.skboutput_args_.emplace_back(file.str, offset.n);
     resources_.needs_perf_event_map = true;
   }
+
+  if (uses_usym_table(call.func))
+  {
+    // mark probe as using usym, so that the symbol table can be pre-loaded
+    // and symbols resolved even when unavailable at resolution time
+    resources_.probes_using_usym.insert(probe_);
+  }
 }
 
 void ResourceAnalyser::visit(Map &map)
@@ -215,6 +229,11 @@ void ResourceAnalyser::prepare_mapped_printf_ids()
     resources_.mapped_printf_ids.push_back({ idx, len + 1 });
     idx += len + 1;
   }
+}
+
+bool ResourceAnalyser::uses_usym_table(const std::string &fun)
+{
+  return fun == "usym" || fun == "func" || fun == "ustack";
 }
 
 Pass CreateResourcePass()
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -60,6 +60,12 @@ BPFtrace::~BPFtrace()
       bcc_free_symcache(pair.second.second, pair.second.first);
   }
 
+  for (const auto& pair : pid_sym_)
+  {
+    if (pair.second)
+      bcc_free_symcache(pair.second, pair.first);
+  }
+
   if (ksyms_)
     bcc_free_symcache(ksyms_, -1);
 }
@@ -2246,7 +2252,7 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
 
   if (resolve_user_symbols_)
   {
-    if (cache_user_symbols_)
+    if (cache_user_symbols_per_program_)
     {
       std::string pid_exe = get_pid_exe(pid);
       if (exe_sym_.find(pid_exe) == exe_sym_.end())
@@ -2262,7 +2268,17 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
     }
     else
     {
-      psyms = bcc_symcache_new(pid, &symopts);
+      // cache user symbols per pid
+      if (pid_sym_.find(pid) == pid_sym_.end())
+      {
+        // not cached, create new ProcSyms cache
+        psyms = bcc_symcache_new(pid, &symopts);
+        pid_sym_[pid] = psyms;
+      }
+      else
+      {
+        psyms = pid_sym_[pid];
+      }
     }
   }
 
@@ -2283,9 +2299,6 @@ std::string BPFtrace::resolve_usym(uintptr_t addr, int pid, bool show_offset, bo
     if (show_module)
       symbol << " ([unknown])";
   }
-
-  if (psyms && !cache_user_symbols_)
-    bcc_free_symcache(psyms, pid);
 
   return symbol.str();
 }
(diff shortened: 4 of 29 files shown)


viktormalik:
CI failed, can you debug it?

lenticularis39:
I looked into that, one issue is codegen tests, because there is now one more attribute of the usym event (that one is easily fixable), other issue is ustack runtime tests failing with segfault (that one will be a little bit more difficult, since it doesn't happen on my system).

Also there is another problem with the PR, that is that the necessary BCC fix was reverted in iovisor/bcc#4270 (it turned that `/proc/.../exe` is not a simple symlink, but it also reflects mount namespaces, meaning the patch broke symbol resultion for processes running in a different namespace). That has to be resolved in order for the last of this PR to work.

viktormalik (commented):
Overall, these are very nice changes!

What I'm missing here, though, are better explanations of *why* the given approaches are necessary - this should appear in the PR description, in commit messages, and in code comments, too. Perhaps even a single-place summary of all the considered scenarios and corresponding solutions could be useful.

src/bpftrace.cpp
```diff
@@ -2129,23 +2157,40 @@ bool BPFtrace::is_aslr_enabled(int pid)
...
     {
-      std::string pid_exe = get_pid_exe(pid);
+      if (!pid_exe.empty() && !show_offset && !show_module)
+      {
+        // try to resolve symbol directly from binary
+        std::map<uint64_t, std::string> &symbol_table =
+            symbol_table_cache_.find(pid_exe) != symbol_table_cache_.end()
+                ? symbol_table_cache_[pid_exe]
+                : (symbol_table_cache_[pid_exe] = get_symbol_table_for_elf(
+                       pid_exe));
```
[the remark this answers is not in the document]
lenticularis39: The issue with using insert is that under eager evaluation this always calls `get_symbol_table_for_elf`, making it less efficient.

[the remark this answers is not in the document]
viktormalik: Ah, right. This looks ugly, but seems like we don't have much of a better option.
Perhaps this could be an alternative, but I'm not strongly against leaving the original one:
```suggestion
        if (symbol_table_cache_.find(pid_exe) == symbol_table_cache_.end())
            symbol_table_cache_[pid_exe] = get_symbol_table_for_elf(pid_exe)
        auto &symbol_table = symbol_table_cache_[pid_exe];
```

lenticularis39:
Rebased and changed caching by PID, which is required for per-process symbol cache preloading, to optional (see the last commit for details).

tests/runtime/builtin
```diff
@@ -108,6 +108,13 @@ EXPECT SUCCESS .*
 AFTER ./testprogs/uprobe_negative_retval
 TIMEOUT 5
 
+# BCC code it depends on is prone to race condition
+#NAME func_uprobe_exited_process
+#PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(func); exit(); }
+#EXPECT test
+#BEFORE ./testprogs/uprobe_symres_exited_process
+#TIMEOUT 5
```
viktormalik, line 117: For other tests, we use
```
REQUIRES bash -c "exit 1"
```
to explicitly disable the test.

lenticularis39: Ah, nice, I was looking for a way to disable a test globally and I didn't find anything.

viktormalik (approved):
Looks good but it's quite a complex change so I'd like to have another review before merging.

man/adoc/bpftrace.adoc
```diff
@@ -1761,8 +1765,9 @@ The pointer type is left unchanged.
 * uretprobes
 
 
-Equal to <<functions_ksym>> but resolves user space symbols
+Equal to <<functions_ksym>> but resolves user space symbols.
 
+If ASLR is enabled, it works only when the process is running at either the time of the symbol resolution or the time of the probe attachment. The latter requires `BPFTRACE_CACHE_USER_SYMBOLS` to be set to `PER_PID`, and might not work with older versions of BCC. A similar limitation also applies to dynamically loaded symbols.
```
ajor, line 1770:
```suggestion
If ASLR is enabled, user space symbolication only works when the process is running at either the time of the symbol resolution or the time of the probe attachment. The latter requires `BPFTRACE_CACHE_USER_SYMBOLS` to be set to `PER_PID`, and might not work with older versions of BCC. A similar limitation also applies to dynamically loaded symbols.
```
Nit: Reworded to sound better to me

src/bpftrace.cpp
```diff
@@ -2196,6 +2197,11 @@ std::string BPFtrace::resolve_usym(uintptr_t addr,
         psyms = pid_sym_[pid];
       }
     }
+    else
+    {
+      // no user symbol caching, create new bcc cache
+      psyms = bcc_symcache_new(pid, &symopts);
```
ajor, line 2203: The old code had a matching `bcc_free_symcache` when caching was disabled - I assume we'll leak here without it

src/main.cpp
```diff
@@ -305,21 +305,44 @@ static std::optional<struct timespec> get_boottime()
...
-    return false;
-
-  if (!get_bool_env_var("BPFTRACE_CACHE_USER_SYMBOLS",
-                        bpftrace.cache_user_symbols_))
-    return false;
+  if (const char* env_p = std::getenv("BPFTRACE_CACHE_USER_SYMBOLS"))
+  {
+    std::string s(env_p);
+    // Note: options 0 and 1 are for compatibility with older versions of
+    // BPFtrace
```
ajor, line 312:
```suggestion
    // bpftrace
```
The old name really refuses to go away! 😂

src/main.cpp
```diff
@@ -305,22 +305,44 @@ static std::optional<struct timespec> get_boottime()
...
+    else
+    {
+      LOG(ERROR)
+          << "Env var 'BPFTRACE_CACHE_USER_SYMBOLS' did not contain a valid "
+             "value: valid values are PER_PID, PER_PROGRAM, and NONE.";
+      return false;
+    }
+  }
 
   if (!std::getenv("BPFTRACE_CACHE_USER_SYMBOLS"))
```
ajor, line 335: Can this just be an "else" or am I missing something?

man/adoc/bpftrace.adoc
```diff
@@ -210,12 +210,16 @@ Increasing the value will consume more memory, increase startup times and can in
 
 === BPFTRACE_CACHE_USER_SYMBOLS
 
-Default: 0 if ASLR is enabled on system and `-c` option is not given; otherwise 1
+Default: PER_PROGRAM if ASLR disabled, PER_PID if ASLR enabled and -p option used; NONE otherwise.
```
ajor, line 213: This doesn't match the implemented behaviour, although the way it's described here sounds good to me

ajor:
I love this PR, thanks so much for doing it - it solves so many problems I've had with user symbols over the years!

I'd suggest that we make "PER_PID" the default caching behaviour when ASLR is enabled, and leave it up to users to disable the cache if they feel it's using too much memory.
1. Having ASLR enabled is not too uncommon, and often users will only be tracing a single process. I'd think the large amount of forking described in #816 is a less common scenario
2. It's simpler for users to understand disabling a cache than it is to work out which type of caching to enable

src/bpftrace.cpp
```diff
@@ -2099,23 +2134,44 @@ bool BPFtrace::is_aslr_enabled(int pid)
...
+    std::string pid_exe = get_pid_exe(pid);
+    if (pid_exe.empty() && probe_id != -1)
+    {
+      // sometimes program cannot be determined from PID, typically when the
+      // process does not exist anymore; in that case, try to get program name
+      // from probe
+      auto probe_full = resolve_probe(probe_id);
+      size_t start = probe_full.find(':') + 1;
+      size_t end = probe_full.find(':', start);
+      pid_exe = probe_full.substr(start, end - start);
```
ajor, line 2158: Is the caching just expected to fail when using wildcards in probes, e.g. "uprobe:/bin/gc*:*"

A comment explaining the expected behaviour would be useful for future developers

src/ast/passes/resource_analyser.cpp
```diff
@@ -222,6 +236,11 @@ void ResourceAnalyser::prepare_mapped_printf_ids()
   }
 }
 
+bool ResourceAnalyser::uses_usym_table(const std::string &fun)
+{
+  return fun == "usym" || fun == "func" || fun == "ustack";
```
ajor, line 241: ustack is listed here, but it doesn't use the usym table, right?

src/required_resources.h
```diff
@@ -116,6 +116,9 @@ class RequiredResources
   std::vector<Probe> special_probes;
   std::vector<Probe> watchpoint_probes;
 
+  // List of probes using userspace symbol resolution
+  std::unordered_set<std::string> probes_using_usym;
```
ajor, line 120:
```suggestion
  std::unordered_set<std::reference_wrapper<Probe>> probes_using_usym;
```
Storing a reference will be a bit faster and use less memory than basing this set on strings. Theoretically we could also get probe name collisions, which references would prevent.

src/bpftrace.cpp
```diff
@@ -303,6 +315,29 @@ int BPFtrace::add_probe(ast::Probe &p)
...
+              symbol_table_cache_.end())
+        symbol_table_cache_[attach_point->target] = get_symbol_table_for_elf(
+            attach_point->target);
+
+      if (user_symbol_cache_type_ == UserSymbolCacheType::per_pid)
+        // preload symbol tables from running processes
+        // this allows symbol resolution for processes that are running at probe
+        // attach time, but not at symbol resolution time, even with ASLR
+        // enabled, since BCC symcache records the offsets
+        for (int pid : get_pids_for_program(attach_point->target))
```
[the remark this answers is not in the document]
lenticularis39: It shouldn't be too much, e.g. `ls -l /proc/*/exe` takes 20ms on a system with around 350 processes, and the iteration only happens once per probe (which can be further reduced to just once with a better implementation).

src/ast/passes/resource_analyser.cpp
lenticularis39: `ustack` uses `resolve_usym` to resolve addresses in the stack to symbols, as shown in your comment above, or did you mean something else?

man/adoc/bpftrace.adoc
lenticularis39: I don't see any difference between the behavior described here and the behavior implemented in `src/main.cpp`. Anyway, I'll change both to reflect the suggestion to default to PER_PID.

src/bpftrace.cpp
lenticularis39: Yes, since the program is not expanded, we don't know which executable the perf event came from if a wildcard is used. I'll add a comment explaining this.

lenticularis39:
Changed default setting of `BPFTRACE_CACHE_USER_SYMBOLS` to `PER_PROGRAM` without ASLR and `PER_PID` per ASLR, updated documentation, resolved comments, rebased to master and updated changelog with the new changes.

ajor:
My comments about "ustack" come from it not passing a "probe_id" value back from the kernel like "usym" now does. The code is currently passing in "-1" as a probe_id for ustacks, which means they don't get all the features of usym (post-termination symbol resolution).

On another note: I'm wondering if we need to do some probe expansion when using usym now, i.e. similar to `probe_->need_expansion`.

I assume that would solve the wildcard issue I commented on before, and I'm also worried that we can potentially resolve symbols incorrectly against the wrong binary as it stands.

e.g. say we have a probe:
```probe:myprog1:funcA,uprobe:myprog2:funcB { @sym = usym) }```

If the process terminates before we resolve the symbol, I believe the current code will always pick "myprog1" as the pid_exe. If it was actually an address from myprog2, then we could end up with completely the wrong symbol. (not tested, just assuming)

Expansion would only really be needed when there are multiple different exes sharing a single probe. Multiple functions from the same exe sharing a probe would not need this. Multiple-function-single-exe is going to be a much more common case, and over-expanding probes would have some significant overheads when tracing a large number of functions, so is something we should avoid.

I realise that adding in this logic would not necessarily be trivial, and the rest of your changes in this PR look great, so if you want we could merge this with the post-process-termination logic temporarily disabled and follow up with a separate PR. Up to you.

**Comment by ajor** [comment]
Or maybe just explicitly disabling the post-process-termination logic for probes specifying multiple target paths (with either wildcards or commas) would be good enough as a permanent solution. That'd leave it working fast in the common case (probes with a single target path), and ensure we still get correct results, albeit slower, in the less common case.

**Comment by brendangregg** [comment]
I just heard about https://github.com/libbpf/blazesym which Meta is working on. At LSFMMBPF there was talk about whether it made sense for bcc and bpftrace to use blazesym (talk as in "does someone want to explore doing this?").

**Comment by lenticularis39** [comment]
I disabled ELF symbol resulution for probes with multiple attach points as suggested, since we don't know which binary is the correct one. `kstack` now passed `probe_id` to `resolve_usym`; I also updated the ELF symbol resolution logic to work with addresses with non-zero offsets, and removed the code for legacy BCC in `get_symbol_table_for_elf`, since we now ship BCC as a submodule.

(The failing tests are just a missing cast, I'll fix that soon.)

**Comment by viktormalik** [comment]
> I just heard about https://github.com/libbpf/blazesym which Meta is working on. At LSFMMBPF there was talk about whether it made sense for bcc and bpftrace to use blazesym (talk as in "does someone want to explore doing this?").

@brendangregg we were discussing this at one of the bpftrace office hours meetings. We're slowly trying to get rid of dependency on bcc and symbolization is one of those parts where we depend on it a lot. I'd keep that discussion outside of this PR, though.

File: src/bpftrace.h [hunk 12]
```diff
@@ -223,7 +236,11 @@ class BPFtrace
                         BpfBytecode &bytecode,
                         void (*trigger)(void));
   void* ksyms_{nullptr};
+  // note: exe_sym_ is used when layout is same for all instances of program
   std::map<std::string, std::pair<int, void *>> exe_sym_; // exe -> (pid, cache)
+  std::map<int, void *> pid_sym_;                         // pid -> cache
+  std::map<std::string, std::map<uint64_t, elf_symbol, std::greater<uint64_t>>>
```
ajor, line 242: I think uintptr_t would be a better type for storing symbols. It's what we use everywhere else at least

tests/runtime/builtin
```diff
@@ -108,6 +108,15 @@ EXPECT SUCCESS .*
...
 
+# Disabled, since BCC code it depends on is prone to race condition
+NAME func_uprobe_exited_process
+ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PID
+PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(func); exit(); }
+EXPECT test
+BEFORE ./testprogs/uprobe_symres_exited_process
+TIMEOUT 5
+REQUIRES bash -c "exit 1"
+
```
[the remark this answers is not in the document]
ajor: Ok, makes sense. I think there'd be a race condition between the target process exiting and bpftrace printing, but it'd be so extremely unlikely to happen that it doesn't matter for this.

What's the race condition you mention in the comment about this test being disabled though?

ajor (commented):
> I also updated the ELF symbol resolution logic to work with addresses with non-zero offsets

Did you have an example of where this was causing symbolication problems? It'd be useful to check that in as a test-case.

I have a slight concern about the impact this pre-caching will have on startup time when tracing huge programs. I'll do a little bit of testing on this to satisfy myself.

src/utils.cpp
```diff
@@ -1335,4 +1337,49 @@ std::tuple<std::string, std::string, std::string> split_addrrange_symbol_module(
                          symbol.length() - idx2 - strlen(" []")) };
 }
 
+std::map<uint64_t, elf_symbol, std::greater<uint64_t>> get_symbol_table_for_elf(
```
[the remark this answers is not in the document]
lenticularis39: That is because the semantics of `std::map::lower_bound` and `std::map::upper_bound`, citing CPPreference:
- `std::map::lower_bound`: Returns an iterator pointing to the first element that is not less than (i.e. greater or equal to) key.
- `std::map::upper_bound`:  Returns an iterator pointing to the first element that is greater than key.

I want the element that is *less or equal* to the key (possible start address of the symbol), therefore I have to sort the elements of the map the other way.

tests/runtime/builtin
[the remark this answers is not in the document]
lenticularis39: Just for clarification, the sequence of events that happens 99%+ of the time:
- `uprobe_symres_exited_process` starts and sleeps for 3 seconds.
- bpftrace attaches probe on `uprobe_symres_exited_process:test` and pre-loads the cache with the use of util function `get_pids_for_program` and `bcc_symcache_new`.
- the process wakes up, executes `test()` and immediately exists
- the probe fires and `bcc_symcache_resolve` is run on the stored cache (note that `symbol_table_cache_` is not used here, since `BPFTRACE_CACHE_USER_SYMBOLS` is set to `PER_PID`)

Now what happens in `bcc_symcache_resolve` is that the binary of `uprobe_symres_exited_process` is opened to look for the symbol. However, since in a general case the binary can be in a different mount namespace, BCC uses a file descriptor of `/proc/<pid>/root` that was pre-opened in `bcc_symcache_now` to access the file using `openat` (see iovisor/bcc#4319). The issue is that this approach with `openat` sometimes does not work due to current implementation issues, hence BCC checks if the process is running, and if yes, it does not use `openat`. This is prone to a race condition, since the process may exit *after* this check is done.

This race condition is much rarer than the original one from #2284 linked above (which in my case happened 100% of the time), but still happens and would cause CI randomly failing, hence the test being disabled.

lenticularis39:
> > I also updated the ELF symbol resolution logic to work with addresses with non-zero offsets
> 
> Did you have an example of where this was causing symbolication problems? It'd be useful to check that in as a test-case.

The issue is in general `ustack` addresses that are not resolvable using `bcc_symcache_resolve`, but are resolvable using `symbol_table_cache_`. More specifically, it is `ustack` addresses in exited processes with ASLR disabled - I didn't write any tests for this because of the requirement of disabling ASLR (also for `usym` resolution), now when I'm thinking about it I could use `personality(ADDR_NO_RANDOMIZE)` to disable ASLR in a test program and test both.

> I have a slight concern about the impact this pre-caching will have on startup time when tracing huge programs. I'll do a little bit of testing on this to satisfy myself.

OK. The bcc symcache pre-loading is not a issue, since it only loads the symbols when `bcc_symcache_resolve` is called (since e.g. dynamically loaded modules might not be present at the time of calling `bcc_symcache_new`), what could impact the startup time is `symbol_table_cache_`, it might be useful to have an option to disable it, like we do for the symcache.

src/types.cpp
```diff
@@ -119,9 +119,10 @@ bool SizedType::operator==(const SizedType &t) const
 
 bool SizedType::IsByteArray() const
 {
-  return type == Type::string || type == Type::usym || type == Type::inet ||
-         type == Type::buffer || type == Type::timestamp ||
-         type == Type::mac_address || type == Type::cgroup_path;
+  return type == Type::string || type == Type::usym || type == Type::ustack ||
```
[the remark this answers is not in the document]
lenticularis39: Wouldn't that then also be a case for `ustack`, which is already defined as a byte array and realized with an 8-aligned struct in codegen?

src/bpftrace.h
lenticularis39: Makes sense, on 32-bit platforms it will save some memory.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -146,9 +146,43 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
     // pack uint64_t with: (uint32_t)stack_id, (uint32_t)pid
     Value *pidhigh = b_.CreateShl(b_.CreateGetPidTgid(loc), 32);
     stackid = b_.CreateOr(stackid, pidhigh);
+
+    // create structure to hold both stackid and probe id
+    std::vector<llvm::Type *> elements = {
+      b_.getInt64Ty(), // stack id
+      b_.getInt64Ty(), // probe id
+    };
```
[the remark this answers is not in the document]
lenticularis39: That shouldn't be too hard, I'll try to add it with improvements from the other comments, if it turns out to be too difficult to quickly implement, I'll add a comment.

src/types.cpp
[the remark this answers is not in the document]
ajor: Good point, and I don't know...

@viktormalik Can you shed any light on this?

tests/testprogs/disable_aslr.c
```diff
@@ -0,0 +1,8 @@
+#include <sys/personality.h>
+#include <unistd.h>
+
+int main(int argc __attribute__((unused)), char **argv)
+{
+  personality(ADDR_NO_RANDOMIZE);
+  execv(argv[1], argv + 1);
```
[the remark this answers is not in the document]
lenticularis39: I am wondering if this can be solved by an existing program, I don't know any.

lenticularis39:
Oops, I forgot to change the main offset in the test to a more generic pattern, causing the test to fail with a different compiler.

tests/testprogs/disable_aslr.c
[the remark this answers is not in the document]
viktormalik: Since it's in tests, there's no need to fix this. Dismissed.

src/types.cpp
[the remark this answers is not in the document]
viktormalik: So, it turns out that the `SizedType::GetAlignment` function is only used to compute alignment in tuples. I tried to change it to 8 but that causes some failures, so I'll need to investigate more.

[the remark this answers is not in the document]
viktormalik: This is now addressed in #2625.

ajor (approved):
Looks great, thank you!

lenticularis39:
Rebased and fixed failing test added by #2625.

PR #2387, opened 2022-10-12
support integer arrays comparison

xh4n3:
This PR implements the binop comparison for integer arrays, it is a part of #2248 , a follow-up to #2261 .

By using this binop and the `pton` built-in together, we can now select packets filtered by addresses. Example:

```
tracepoint:tcp:tcp_retransmit_skb {
  $addr = pton("::1");
  if (args->daddr_v6 == $addr) {
    printf("localhost ipv6 retransmit\n");
  } else {
    printf("non-localhost ipv6 retransmit\n");
  }
  return;
}
```

Comments are welcome.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -974,6 +974,233 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
   return CreateHelperCall(fn, gettime_func_type, {}, "get_ns", &loc);
 }
 
+Value *IRBuilderBPF::CreateIntegerArrayCmpUnrolled(Value *ctx,
+                                                   Value *val1,
+                                                   Value *val2,
+                                                   const SizedType &val1_type,
+                                                   const SizedType &val2_type,
+                                                   const bool inverse,
+                                                   const location &loc)
+{
+  /*
+   // This function compares each character of the two arrays.
+   // It returns true if all are equal and false if any are different
+   // cmp([]char val1, []char val2)
+   {
+      for (size_t i = 0; i < n; i++)
+      {
+        if (val1[i] != val2[i])
+        {
+          return false;
+        }
+      }
+      return true;
+   }
+*/
+  auto elem_type = *val1_type.GetElementTy();
+  const size_t num = val1_type.GetNumElements();
+
+  Value *val1_elem_i, *val2_elem_i, *cmp;
+  AllocaInst *v1 = CreateAllocaBPF(elem_type, "v1");
+  AllocaInst *v2 = CreateAllocaBPF(elem_type, "v2");
+
+  Function *parent = GetInsertBlock()->getParent();
+  AllocaInst *store = CreateAllocaBPF(getInt1Ty(), "arraycmp.result");
+  CreateStore(getInt1(inverse), store);
+  BasicBlock *arr_ne = BasicBlock::Create(module_.getContext(),
+                                          "arraycmp.false",
+                                          parent);
+  BasicBlock *done = BasicBlock::Create(module_.getContext(),
+                                        "arraycmp.done",
+                                        parent);
+
+  Value *ptr_val1 = CreateIntToPtr(val1, GetType(val1_type)->getPointerTo());
+  Value *ptr_val2 = CreateIntToPtr(val2, GetType(val2_type)->getPointerTo());
+
+  for (size_t i = 0; i < num; i++)
+  {
+    BasicBlock *arr_eq = BasicBlock::Create(module_.getContext(),
+                                            "arraycmp.loop",
+                                            parent);
+    auto *ptr_val1_elem_i = CreateGEP(GetType(val1_type),
+                                      ptr_val1,
+                                      { getInt32(0), getInt32(i) });
+    if (onStack(val1_type))
+    {
+      val1_elem_i = CreateLoad(GetType(elem_type), ptr_val1_elem_i);
+    }
+    else
+    {
+      CreateProbeRead(ctx,
+                      v1,
+                      getInt32(elem_type.GetSize()),
+                      ptr_val1_elem_i,
+                      val1_type.GetAS(),
+                      loc);
+      val1_elem_i = CreateLoad(GetType(elem_type), v1);
+    }
+
+    auto *ptr_val2_elem_i = CreateGEP(GetType(val2_type),
+                                      ptr_val2,
+                                      { getInt32(0), getInt32(i) });
+    if (onStack(val2_type))
+    {
+      val2_elem_i = CreateLoad(GetType(elem_type), ptr_val2_elem_i);
+    }
+    else
+    {
+      CreateProbeRead(ctx,
+                      v2,
+                      getInt32(elem_type.GetSize()),
+                      ptr_val2_elem_i,
+                      val2_type.GetAS(),
+                      loc);
+      val2_elem_i = CreateLoad(GetType(elem_type), v2);
+    }
+    cmp = CreateICmpNE(val1_elem_i, val2_elem_i, "arraycmp.cmp");
+
+    CreateCondBr(cmp, arr_ne, arr_eq);
+    SetInsertPoint(arr_eq);
+  }
+  CreateBr(done);
+
+  SetInsertPoint(arr_ne);
+  CreateStore(getInt1(!inverse), store);
+  CreateBr(done);
+
+  SetInsertPoint(done);
+  Value *result = CreateLoad(getInt1Ty(), store);
+  CreateLifetimeEnd(store);
+  CreateLifetimeEnd(v1);
+  CreateLifetimeEnd(v2);
+  result = CreateIntCast(result, getInt64Ty(), false);
+  return result;
+}
+
+Value *IRBuilderBPF::CreateIntegerArrayCmp(Value *ctx,
+                                           Value *val1,
+                                           Value *val2,
+                                           const SizedType &val1_type,
+                                           const SizedType &val2_type,
+                                           const bool inverse,
+                                           const location &loc,
+                                           MDNode *metadata)
+{
+  /*
+   // This function compares each character of the two arrays.
+   // It returns true if all are equal and false if any are different
+   // cmp([]char val1, []char val2)
+   {
+      for (size_t i = 0; i < n; i++)
+      {
+        if (val1[i] != val2[i])
+        {
+          return false;
+        }
+      }
+      return true;
+   }
+*/
+  auto elem_type = *val1_type.GetElementTy();
+  const size_t num = val1_type.GetNumElements();
+
+  Value *val1_elem_i, *val2_elem_i, *cmp;
+  AllocaInst *v1 = CreateAllocaBPF(elem_type, "v1");
+  AllocaInst *v2 = CreateAllocaBPF(elem_type, "v2");
+  AllocaInst *store = CreateAllocaBPF(getInt1Ty(), "arraycmp.result");
+  CreateStore(getInt1(inverse), store);
+
+  Function *parent = GetInsertBlock()->getParent();
+  BasicBlock *while_cond = BasicBlock::Create(module_.getContext(),
+                                              "while_cond",
+                                              parent);
+  BasicBlock *while_body = BasicBlock::Create(module_.getContext(),
+                                              "while_body",
+                                              parent);
+  BasicBlock *arr_ne = BasicBlock::Create(module_.getContext(),
+                                          "arraycmp.false",
+                                          parent);
+  BasicBlock *done = BasicBlock::Create(module_.getContext(),
+                                        "arraycmp.done",
+                                        parent);
+
+  Value *ptr_val1 = CreateIntToPtr(val1, GetType(val1_type)->getPointerTo());
+  Value *ptr_val2 = CreateIntToPtr(val2, GetType(val2_type)->getPointerTo());
+  AllocaInst *i = CreateAllocaBPF(getInt32Ty(), "i");
+  AllocaInst *n = CreateAllocaBPF(getInt32Ty(), "n");
+  CreateStore(getInt32(0), i);
+  CreateStore(getInt32(num), n);
+  CreateBr(while_cond);
+
+  SetInsertPoint(while_cond);
+  auto *cond = CreateICmpSLT(CreateLoad(getInt32Ty(), i),
+                             CreateLoad(getInt32Ty(), n),
+                             "size_check");
+  Instruction *loop_hdr = CreateCondBr(cond, while_body, done);
+  loop_hdr->setMetadata(LLVMContext::MD_loop, metadata);
+
+  SetInsertPoint(while_body);
+  BasicBlock *arr_eq = BasicBlock::Create(module_.getContext(),
+                                          "arraycmp.loop",
+                                          parent);
+  auto *ptr_val1_elem_i = CreateGEP(GetType(val1_type),
+                                    ptr_val1,
+                                    { getInt32(0),
+                                      CreateLoad(getInt32Ty(), i) });
+  if (onStack(val1_type))
+  {
+    val1_elem_i = CreateLoad(GetType(elem_type), ptr_val1_elem_i);
+  }
+  else
+  {
+    CreateProbeRead(ctx,
+                    v1,
+                    getInt32(elem_type.GetSize()),
+                    ptr_val1_elem_i,
+                    val1_type.GetAS(),
+                    loc);
+    val1_elem_i = CreateLoad(GetType(elem_type), v1);
+  }
+
+  auto *ptr_val2_elem_i = CreateGEP(GetType(val2_type),
+                                    ptr_val2,
+                                    { getInt32(0),
+                                      CreateLoad(getInt32Ty(), i) });
+  if (onStack(val2_type))
+  {
+    val2_elem_i = CreateLoad(GetType(elem_type), ptr_val2_elem_i);
+  }
+  else
+  {
+    CreateProbeRead(ctx,
+                    v2,
+                    getInt32(elem_type.GetSize()),
+                    ptr_val2_elem_i,
+                    val2_type.GetAS(),
+                    loc);
+    val2_elem_i = CreateLoad(GetType(elem_type), v2);
+  }
+
+  cmp = CreateICmpNE(val1_elem_i, val2_elem_i, "arraycmp.cmp");
+  CreateCondBr(cmp, arr_ne, arr_eq);
+
+  SetInsertPoint(arr_eq);
+  CreateStore(CreateAdd(CreateLoad(getInt32Ty(), i), getInt32(1)), i);
+  CreateBr(while_cond);
+
+  SetInsertPoint(arr_ne);
+  CreateStore(getInt1(!inverse), store);
+  CreateBr(done);
+
+  SetInsertPoint(done);
+  Value *result = CreateLoad(getInt1Ty(), store);
+  CreateLifetimeEnd(store);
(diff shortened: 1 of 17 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -1336,6 +1336,38 @@ void CodegenLLVM::binop_string(Binop &binop)
...
+               << "\"";
+  }
+
+  std::string string_literal;
+
+  // cmp returns 0 when arrays are equal
+  bool inverse = binop.op == Operator::EQ;
+
+  auto left_array_ptr = getString(binop.left);
+  auto right_array_ptr = getString(binop.right);
```
viktormalik, line 1353: This doesn't seem correct, if you want to work with integer arrays the why use `getString`?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1336,6 +1336,38 @@ void CodegenLLVM::binop_string(Binop &binop)
...
 
+void CodegenLLVM::binop_integer_array(Binop &binop)
+{
+  if (binop.op != Operator::EQ && binop.op != Operator::NE)
+  {
+    LOG(FATAL) << "missing codegen to array operator \"" << opstr(binop)
+               << "\"";
+  }
+
+  std::string string_literal;
```
viktormalik, line 1347: Unused variable?

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,71 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+  CreateStore(getInt1(!inverse), store);
+  for (size_t i = 0; i < n; i++)
+  {
+    BasicBlock *char_eq = BasicBlock::Create(module_.getContext(),
+                                             "arraycmp.loop",
+                                             parent);
+    offset = CreateMul(getInt64(i), getInt64(elem_size));
+    l_addr = CreateAdd(val1, offset);
+    r_addr = CreateAdd(val2, offset);
+    CreateProbeRead(ctx, l, getInt32(1), l_addr, AddrSpace::kernel, loc);
```
viktormalik, line 879: If the value is on stack already (e.g. when using `pton`) then you should use GEP instead of proberead.

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,71 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+   }
+*/
+  size_t elem_size = elem_type.GetSize();
+  Value *offset, *l_addr, *r_addr, *ll, *rr;
+  AllocaInst *l = CreateAllocaBPF(elem_type, "ll");
+  AllocaInst *r = CreateAllocaBPF(elem_type, "rr");
+
+  Function *parent = GetInsertBlock()->getParent();
+  AllocaInst *store = CreateAllocaBPF(getInt1Ty(), "arraycmp.result");
+  BasicBlock *str_ne = BasicBlock::Create(module_.getContext(),
```
viktormalik, line 863: `arr_ne` would be a better name here

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,71 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+                                          "arraycmp.false",
+                                          parent);
+  BasicBlock *done = BasicBlock::Create(module_.getContext(),
+                                        "arraycmp.done",
+                                        parent);
+
+  CreateStore(getInt1(!inverse), store);
+  for (size_t i = 0; i < n; i++)
+  {
+    BasicBlock *char_eq = BasicBlock::Create(module_.getContext(),
```
viktormalik, line 873: Same as above: `arr_eq` is a more suitable name.

tests/runtime/array
```diff
@@ -142,3 +142,15 @@ RUN {{BPFTRACE}} -e 'struct A { int x[4]; } uprobe:./testprogs/array_access:test
 EXPECT Result: 1
 TIMEOUT 5
 AFTER ./testprogs/array_access
+
+NAME array compare eq
+PROG struct A { int x[4]; } uprobe:./testprogs/array_access:test_arrays { $a = (struct A *) arg0; $b = (struct A *) arg0; if ($a == $b) { printf("two int arrays are equal.\n"); } exit(); }
```
viktormalik, line 147: This is not correct, you're comparing pointers (`struct A *`) instead of arrays. Should be `$a->x == $b->x`.

viktormalik (commented):
It feels like we're missing some checks in the semantic analyser when comparing arrays - in particular equality of inner types and of the array sizes.

src/ast/irbuilderbpf.cpp
xh4n3: @viktormalik Thanks! So the proper way is, use a GEP when the value is on stack, use a proberead when it is not. Am I understanding it correctly? How should determine if the value is on stack, use IsCtxAccess?

xh4n3:
@viktormalik Thank you for reviewing this. 😃 

> It feels like we're missing some checks in the semantic analyser when comparing arrays - in particular equality of inner types and of the array sizes.

I've added this part.

> If the value is on stack already (e.g. when using pton) then you should use GEP instead of proberead.

I've updated this part also, not sure if I understand it correctly. Could you please help me check it?
https://github.com/iovisor/bpftrace/pull/2387/files#diff-3d80d60dba9cfc34c2fe79c01df26172886685f76475c9b53b372136e49f1af0R876-R915

src/ast/irbuilderbpf.cpp
viktormalik: Sorry for a late answer. There's a function `shouldBeOnStackAlready` which should be used to detect this (although you'll have to add `pton` there for your usecase).

xh4n3: @viktormalik Cool, will try that. Besides, may i know what does the CtxAccess mean?

viktormalik: Context is a special structure that is passed to each BPF program. The exact information passed within the context depends on the program type.

xh4n3: Thanks. I found `shouldBeOnStackAlready` only recognizes those specific `SizedType`, while `pton` is returning byte arrays only. May I use `is_internal` instead?

viktormalik: Yeah, `is_internal` makes sense. And you can use `onStack` instead of `shouldBeOnStackAlready`, then.

xh4n3: @viktormalik Great, now the code is using `is_internal`. Please let me if any other issues. Thanks!

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,103 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+                                        "arraycmp.done",
+                                        parent);
+
+  CreateStore(getInt1(!inverse), store);
+  for (size_t i = 0; i < n; i++)
+  {
+    BasicBlock *arr_eq = BasicBlock::Create(module_.getContext(),
+                                            "arraycmp.loop",
+                                            parent);
+    if (onStack(array_type))
```
viktormalik, line 876: This doesn't seem correct. One of the arrays may be on stack while the other one is not.

I'm also wondering if it'd make sense to make use of the `readDatastructElemFromStack` and `probereadDatastructElem` functions.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1675,6 +1676,29 @@ void SemanticAnalyser::binop_int(Binop &binop)
...
+        << "Arrays support EQ(==) and NE(!=) operators only.";
+  }
+
+  if (!lht.GetElementTy()->IsIntegerTy() || !rht.GetElementTy()->IsIntegerTy())
+  {
+    LOG(ERROR, binop.loc, err_)
+        << "Only arrays of integer support comparison operators.";
+  }
+
+  if (lht.GetSize() != rht.GetSize())
```
viktormalik, line 1695: This should compare the number of elements and the element size separately.

viktormalik (changes requested):
This feature should also have a codegen test.

src/ast/irbuilderbpf.cpp
xh4n3: > if it'd make sense to make use of the readDatastructElemFromStack and probereadDatastructElem functions.

This is doable, then we should be moving `readDatastructElemFromStack`  and `probereadDatastructElem` to `irbuilderbpf.cpp`. But the IRBuilderBPF doesn't have the `current_attach_point` variable, so I guess I need to pass this as a parameter. Is that alright?

viktormalik: Yeah, that would be necessary and also, `expr_deleter_` is missing. So I guess that it's best to just not use those functions (they anyway do more functionality than what is necessary here).

tests/codegen/array_integer_not_equal_comparison.cpp
```diff
@@ -0,0 +1,25 @@
+#include "common.h"
```
viktormalik, line 1: No need to have two (almost) identical codegen tests.

xh4n3:
> This feature should also have a codegen test.

Done.

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,101 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+      rr = CreateLoad(GetType(elem_type), ptr_r);
+    }
+    else
+    {
+      auto *ptr_r = CreateAdd(
+          val2, CreateMul(getInt64(i), getInt64(elem_type.GetSize())));
+      CreateProbeRead(
+          ctx, r, getInt32(elem_type.GetSize()), ptr_r, val1_type.GetAS(), loc);
+      rr = CreateLoad(GetType(elem_type), r);
+    }
```
viktormalik, line 908: TBH, I'm not very keen that this is basically a repetition of the above code. As we discussed, this is a special use-case which doesn't need full-fledged element accessing functions, so I guess I'm ok with this. But still, if you found a way to factor this into a reasonable function, that would be great.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1336,6 +1336,42 @@ void CodegenLLVM::binop_string(Binop &binop)
...
+  if (binop.op != Operator::EQ && binop.op != Operator::NE)
+  {
+    LOG(FATAL) << "missing codegen to array operator \"" << opstr(binop)
+               << "\"";
+  }
+
+  // integer array compare returns 0 when arrays are equal
+  bool inverse = binop.op == Operator::EQ;
+
+  auto scoped_del_right = accept(binop.left);
```
viktormalik, line 1350:
```suggestion
  auto scoped_del_left = accept(binop.left);
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1336,6 +1336,42 @@ void CodegenLLVM::binop_string(Binop &binop)
...
+  Value *left_array_val = expr_;
+
+  auto scoped_del_left = accept(binop.right);
```
viktormalik, line 1353:
```suggestion
  auto scoped_del_right = accept(binop.right);
```

src/ast/passes/codegen_llvm.cpp
xh4n3: Sorry, I was being stupid..

src/ast/irbuilderbpf.cpp
viktormalik: It does look better, but now there's a confusion whether to use `CreateGetElemInArrayOnStack` or `readDatastructElemFromStack`. They do similar things, although the new one is specialized for non-context integer-only accesses (same applies for proberead variants).

It's probably the best to leave the original expanded code here - it's a single function after all. Or, alternatively, you could move it to `codegen_llvm.cpp` in which case you'll be able to use the current element accessing functions.

xh4n3: > Or, alternatively, you could move it to codegen_llvm.cpp in which case you'll be able to use the current element accessing functions.

Do you mean that I could move the `CreateGetElemInArrayOnStack` function into `codegen_llvm.cpp`?
Then `IRBuilderBPF`'s `CreateIntegerArrayCmp` would be calling `CodegenLLVM`'s methods, which I don't think it is a good idea.

xh4n3: I see, that would leave `CreateIntegerArrayCmp` and `CreateStrncmp` in separate files, kind of strange. Let me keep the expanded one for this PR.

viktormalik (approved):

viktormalik:
This has quite a lot of code, I'd like to have one more review.

@danobi, @ajor, @fbs, @BurntBrunch

danobi:
Missing semantic analysis unit tests? Maybe also test semantic analysis errors out correctly for:
* arrays of different length
* arrays of different type
* array compared to !array

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,101 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+                                          "arraycmp.false",
+                                          parent);
+  BasicBlock *done = BasicBlock::Create(module_.getContext(),
+                                        "arraycmp.done",
+                                        parent);
+
+  CreateStore(getInt1(!inverse), store);
+  for (size_t i = 0; i < n; i++)
+  {
+    BasicBlock *arr_eq = BasicBlock::Create(module_.getContext(),
```
BurntBrunch, line 874: Similarly to the strstr discussion in #2393, I think this implementation would be better if we don't unroll the loop.

danobi: Seems reasonable to me to gate this helper based on bounded loop presence.

BurntBrunch: > based on bounded loop support

That would be.. [5.3, the patch is from Jun 2019](https://lwn.net/Articles/794934/). Runtime checks for this are annoying but possible.

danobi: You mean a feature probe like this? https://github.com/iovisor/bpftrace/blob/86b7843b0e56c96dc91c9fa0354ffe1507245ad8/src/bpffeature.cpp#L191-L207

xh4n3:
Hi @danobi , thanks for the review!
I've added semantic analysis tests and an unrolling-disabled variation for the `CreateIntegerArrayCmp` method.

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,238 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+  AllocaInst *i = CreateAllocaBPF(getInt32Ty(), "i");
+  AllocaInst *n = CreateAllocaBPF(getInt32Ty(), "n");
+  CreateStore(getInt32(0), i);
+  CreateStore(getInt32(num), n);
+
+  CreateBr(while_cond);
+  SetInsertPoint(while_cond);
+  auto *cond = CreateICmpSLT(CreateLoad(getInt32Ty(), i),
+                             CreateLoad(getInt32Ty(), n),
+                             "true_cond");
```
viktormalik, line 996: Maybe
```suggestion
                             "size_check");
```
`true_cond` doesn't seem like a good name.

viktormalik (commented):
Thanks for the update! The loop variant looks good, although I've couple of nits to improve the IR. Feel free to ignore those that don't make sense to you.

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,238 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+  SetInsertPoint(arr_ne);
+  Value *result = CreateLoad(getInt1Ty(), store);
+  CreateLifetimeEnd(store);
+  CreateLifetimeEnd(v1);
+  CreateLifetimeEnd(v2);
+  result = CreateIntCast(result, getInt64Ty(), false);
+  return result;
+}
+
+Value *IRBuilderBPF::CreateIntegerArrayCmp(Value *ctx,
```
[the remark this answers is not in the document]
xh4n3: I renamed the unrolled function to `CreateIntegerArrayCmpUnrolled`. I think if this is the right way to go, we probably should refactor other irbuilder functions as well, e.g. the `strncmp`.

src/ast/irbuilderbpf.cpp
```diff
@@ -830,6 +830,238 @@ CallInst *IRBuilderBPF::CreateGetNs(bool boot_time, const location &loc)
...
+  cmp = CreateICmpNE(val1_elem_i, val2_elem_i, "arraycmp.cmp");
+  CreateCondBr(cmp, arr_ne, arr_eq);
+
+  SetInsertPoint(arr_eq);
+  CreateStore(CreateAdd(CreateLoad(getInt32Ty(), i), getInt32(1)), i);
+
+  CreateBr(while_cond);
+  SetInsertPoint(while_end);
+
+  CreateBr(done);
```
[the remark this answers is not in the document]
xh4n3: I have to confess that I borrowed some lines of codes from other irbuilder functions, including those empty line code style.
Thanks for letting me understand this. Let me create a PR to clean up these bad code style.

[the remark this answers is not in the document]
viktormalik: I'm not a big fan of code-style-only changes (makes blaming harder). Let's leave the rest as is and change it once we're doing functional changes (which we'll probably do for some, e.g. `strncmp`).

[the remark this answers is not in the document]
xh4n3: Okay, I see.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1682,6 +1683,35 @@ void SemanticAnalyser::binop_int(Binop &binop)
...
+{
+  auto &lht = binop.left->type;
+  auto &rht = binop.right->type;
+  if (binop.op != Operator::EQ && binop.op != Operator::NE)
+  {
+    LOG(ERROR, binop.loc, err_)
+        << "The " << opstr(binop) << " operator cannot be used on arrays.";
+  }
+
+  if (!lht.GetElementTy()->IsIntegerTy() || !rht.GetElementTy()->IsIntegerTy())
```
fbs, line 1696: maybe we can just do:

```
  if (lht.GetNumElements() != rht.GetNumElements())
  {
    LOG(ERROR, binop.loc, err_)
     << "Arrays must be of the same length. " << lth << " != " << rht;
  }

if (!lht.GetElementTy()->IsIntegerTy()  || lht != rht) { // uses the IsEqual method on sized type
  LOG(ERROR, binop.loc, err_)
     << "Array elements must be of the same integer type. " << lth << " != " << rht;
}
```

This removes 2 related checks and also indicates what the mismatching types are

fbs (changes requested):
looks good!

src/ast/passes/semantic_analyser.cpp
xh4n3: Sure!

xh4n3:
@fbs @danobi Hi, I've done the changes regarding to your comments. Could I have an approval? Thanks!

fbs (approved):

fbs:
great work! I'll merge it soon

fbs:
woops the other merge caused a conflict :(

xh4n3:
> woops the other merge caused a conflict :(

fixed :p

fbs:
thanks!

Issue #1151, opened 2020-02-17
Add new builtin to check if currently in user or kernel mode


ajor:
`kstack` will be empty if a probe (e.g. `profile`) fires while the CPU was in user-mode. Adding a builtin to check what mode we were in will make it easier to filter out invalid stacks.

See #1124
Platform-dependent solution: https://github.com/iovisor/bpftrace/issues/1124#issuecomment-582347103

ryantimwilson:
Per @tyroguru, dtrace encoded this as `arg1` for profile/interval probes vs. a separate builtin: https://docs.oracle.com/cd/E19253-01/817-6223/chp-profile/index.html

We can also go with that approach as well!

4ast:
hmm. kstack should never be empty. bpf prog always runs in the kernel context.

danobi:
There was the bug Yonghong fixed for kstack in syscall tracepoint right?

4ast:
> There was the bug Yonghong fixed for kstack in syscall tracepoint right?

yeah. original "lack of stack" issue could have been that bug.

ajor:
Yonghong's patch fixed kstack in syscall tracepoints specifically (#442), but that's not the same as the issue here, which is that `profile` probes (and similar perf_events probe types, e.g. `hardware:cpu-cycles`) can return empty kernel stack traces.

I can still get empty kernel stack traces today with this repro:
```c
#include <stdio.h>

double foo(double a, double b)
{
  return a/b;
}

int main()
{
  while(1)
    foo(999999999.0, 2135413462343.0);
}
```

```c
profile:hz:99 /comm=="a.out"/ { @[kstack, ustack] = count() }
```

Output shows various ustack values but kstack is always empty:
```
@[
,
        main+4
        __libc_start_call_main+128
]: 9
@[
,
        main+28
        __libc_start_call_main+128
]: 15
@[
,
        f1+19
        main+33
        __libc_start_call_main+128
]: 18
@[
,
        f1+4
        main+33
        __libc_start_call_main+128
]: 19
@[
,
        f1+24
        main+33
        __libc_start_call_main+128
]: 21

```

PR #2392, opened 2022-10-14
Do not produce empty stack strings for error stacks

BurntBrunch:
Currently the following output is possible:
```
^C
...
@[]: 1234
```

This is incredibly misleading and hard to diagnose. What's worse, the error LOG is printed _before_ any map output, so it gets immediately lost in the scrollback.

This PR changes the output format to contain the error value and pid:

```
@[
    <error: -17, pid: -1>
]: 14
```

(pid: -1 is a bug, will be fixed separately)

I'm open to ideas on how to write a deterministic runtime test for this behavior.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1758,11 +1758,30 @@ std::string BPFtrace::get_stack(uint64_t stackidpid, bool ustack, StackType stac
                             stack_trace.data());
   if (err)
   {
-    // ignore EFAULT errors: eg, kstack used but no kernel stack
-    if (stackid != -EFAULT)
+    if (stackid == -EFAULT && stack_type.type == StackType::ukstack::kstack)
+    {
+      // EFAULT kstacks are expected when the task is in usermode.
+      return "";
+    }
+    else
+    {
       LOG(ERROR) << "failed to look up stack id " << stackid << " (pid " << pid
                  << "): " << err;
-    return "";
+    }
+
+    std::ostringstream errstr;
+    switch (stack_type.mode)
+    {
+      case StackMode::bpftrace:
+        errstr << "\n" << std::string(indent, ' ');
+        break;
+      case StackMode::perf:
+        errstr << "\n\t";
+        break;
+    }
+
+    errstr << "<error: " << stackid << ", pid: " << pid << ">\n";
+    return errstr.str();
   }
 
   std::ostringstream stack;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to
 - Support for 32-bit ARM systems
   - [#2360](https://github.com/iovisor/bpftrace/pull/2360)
 #### Changed
+- Output for error stacks has been improved to include the error code and pid
+  - [#2392](https://github.com/iovisor/bpftrace/pull/2392)
 #### Deprecated
 #### Removed
 #### Fixed
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -61,9 +61,15 @@ void ResourceAnalyser::visit(Builtin &builtin)
   {
     resources_.needs_elapsed_map = true;
   }
-  else if (builtin.ident == "kstack" || builtin.ident == "ustack")
+  else if (builtin.ident == "kstack")
   {
-    resources_.stackid_maps.insert(StackType{});
+    resources_.stackid_maps.insert(
+        StackType{ .type = StackType::ukstack::kstack });
+  }
+  else if (builtin.ident == "ustack")
+  {
+    resources_.stackid_maps.insert(
+        StackType{ .type = StackType::ukstack::ustack });
   }
 }
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -336,10 +336,12 @@ void SemanticAnalyser::visit(Builtin &builtin)
     builtin.type.SetAS(find_addrspace(type));
   }
   else if (builtin.ident == "kstack") {
-    builtin.type = CreateStack(true, StackType());
+    builtin.type = CreateStack(true,
+                               StackType{ .type = StackType::ukstack::kstack });
   }
   else if (builtin.ident == "ustack") {
-    builtin.type = CreateStack(false, StackType());
+    builtin.type = CreateStack(false,
+                               StackType{ .type = StackType::ukstack::ustack });
   }
   else if (builtin.ident == "comm") {
     builtin.type = CreateString(COMM_SIZE);
@@ -1365,7 +1367,8 @@ void SemanticAnalyser::check_stack_call(Call &call, bool kernel)
     return;
   }
 
-  StackType stack_type;
+  StackType stack_type{ .type = kernel ? StackType::ukstack::kstack
+                                       : StackType::ukstack::ustack };
   if (call.vargs)
   {
     switch (call.vargs->size())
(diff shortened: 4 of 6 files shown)


viktormalik:
> I'm open to ideas on how to write a deterministic runtime test for this behavior.

Perhaps you could somehow (not sure how, though) delete the stack map from the probe? bpftool probably doesn't allow it, but there may be other ways.

Anyways, I'd be fine with letting this go without a runtime test if we can't come with something reasonably simple.

viktormalik (approved):

BurntBrunch:
The best I've got is a uprobe on a program whose stack takes two pages and one of them has been madvised(DONT_NEED). This should result in EFAULT.

But that setup is frankly a bit much for a test of an error message.

ajor:
I can't build bpftrace to test myself right now, but how does this affect the examples in #428? That's the PR that first started hiding these errors

BurntBrunch:
> I can't build bpftrace to test myself right now, but how does this affect the examples in https://github.com/iovisor/bpftrace/pull/428? That's the PR that first started hiding these errors

```
> sudo src/bpftrace -e 'profile:hz:99 {@[kstack, pid] = count();} END {print(@, 5); clear(@);}'
Attaching 2 probes...
^C@[
    <error: -14, pid: -1>
, 45416]: 8
@[
    <error: -14, pid: -1>
, 48386]: 9
@[
    <error: -14, pid: -1>
, 2048386]: 15
@[
    cpuidle_enter_state+216
    cpuidle_enter+41
    do_idle+438
    cpu_startup_entry+25
    rest_init+203
    arch_call_rest_init+10
    start_kernel+2376
    secondary_startup_64_no_verify+229
, 0]: 313
@[
    cpuidle_enter_state+216
    cpuidle_enter+41
    do_idle+438
    cpu_startup_entry+25
    start_secondary+258
    secondary_startup_64_no_verify+229
, 0]: 4738
```

I can hide them for EFAULT in particular but I would rather not - EFAULT for ustacks is interesting. I would further argue that #428 was entirely the wrong "fix." 

As things got refactored, the error messages ended up *before* the map output, not interleaved with it, which in practice means people don't see them and can't associate them with the empty output. Having the error as a substitute for the stack string (in json output too) makes pipelines built out of bpftrace output more robust, as you're not losing errors anymore.

BurntBrunch:
> Would checking if this is a kstack we're printing and only hiding EFAULT in that case work?

I can give that a shot, it should be doable.

BurntBrunch:
Added tracking of ukstack to StackType and used that to hide only the kstack EFAULT errors.
# Sample output

## ustack
```
> sudo src/bpftrace -e 'profile:hz:99 {@[ustack, pid] = count();} END {print(@, 5); clear(@);}'
Attaching 2 probes...
^CERROR: failed to look up stack id -14 (pid -1): -1
ERROR: failed to look up stack id -14 (pid -1): -1
@[
    __memcmp_avx2_movbe+765
    0x55f85d57712e
    0x55f85d575e36
    0x55f85d6d1f61
    0x55f85d6d1aea
    0x55f85d6d1783
    0x55f85e1f6a8b
    0x55f85d00694a
    0x55f8625af166
    0x55f85e4b9e61
    0x55f85e4b9315
    0x55f85f7f9f46
    0x55f85e4b8de8
    0x55f85e053dce
    start_thread+717
, 142823]: 1
@[
    __libc_read+119
, 495767]: 1
@[
    <error: -14, pid: -1>
, 1301801]: 1
@[
    0x7fdc2a7c888a
    0x7fdc2a7b09ec
    0x7fdc2a794a37
    0x5647dfaa5985
    0x5647dc40394a
    0x5647deac71ee
    0x5647dd8b6e61
    0x5647dd8b6315
    0x5647dd7255a1
    0x5647dd895384
    0x5647ddcd8a4a
    0x5647ddcd6641
    0x7fdc3b50b550
, 458040]: 1
@[
    <error: -14, pid: -1>
, 0]: 8261
```

## kstack
```
> sudo src/bpftrace -e 'profile:hz:99 {@[kstack, pid] = count();} END {print(@, 5); clear(@);}'
Attaching 2 probes...
^C@[, 143119]: 3
@[, 142823]: 3
@[
    update_blocked_averages+1548
    run_rebalance_domains+76
    __softirqentry_text_start+248
    __irq_exit_rcu+197
    sysvec_apic_timer_interrupt+158
    asm_sysvec_apic_timer_interrupt+22
    cpuidle_enter_state+216
    cpuidle_enter+41
    do_idle+438
    cpu_startup_entry+25
    start_secondary+258
    secondary_startup_64_no_verify+229
, 0]: 4
@[
    cpuidle_enter_state+216
    cpuidle_enter+41
    do_idle+438
    cpu_startup_entry+25
    rest_init+203
    arch_call_rest_init+10
    start_kernel+2376
    secondary_startup_64_no_verify+229
, 0]: 193
@[
    cpuidle_enter_state+216
    cpuidle_enter+41
    do_idle+438
    cpu_startup_entry+25
    start_secondary+258
    secondary_startup_64_no_verify+229
, 0]: 2924
```

src/bpftrace.cpp
```diff
@@ -1758,11 +1758,30 @@ std::string BPFtrace::get_stack(uint64_t stackidpid, bool ustack, StackType stac
                             stack_trace.data());
   if (err)
   {
-    // ignore EFAULT errors: eg, kstack used but no kernel stack
-    if (stackid != -EFAULT)
+    if (stackid == -EFAULT && stack_type.type == StackType::ukstack::kstack)
```
ajor, line 1761: We've actually got the ustack / kstack information already available as an argument to this function: `ustack`. I'm not opposed to a potential refactor, but there's some unnecessary duplication with this method and I'm not sure how to get rid of it nicely

ajor (commented):
Sorry, I didn't think about the reverse at all before but I think ustack should also silently return an empty string when called from a pure-kernel context, for the same reason of having combined stack traces work.

Why is it that you think EFAULTs shouldn't be hidden? Are there other situations that they could occur in?

I suppose one option could be to check for pid == 0 with ustacks, but then I wonder if we're just ignoring all EFAULTs anyway, just with more steps.

Another option could be to change codegen to not even try collecting stacks in these situations (kstack in user-mode, ustack in pure kernel-mode) and returning a sentinal value for `stackidpid`. That way we wouldn't expect any errors at all so we can just print them all. We'd need a way of determining whether we're in user or kernel mode first: #1151. No idea how easy that'd be to make cross-platform.

(Assuming `stackid` doesn't already have any kind of special value when it's pointing to an invalid stack, prior to calling bpf_lookup_elem. I still haven't got my bpftrace build working to check this now)

BurntBrunch:
@ajor, ping given the above constraints (multiple probe reads to filter kernel tasks from the bpf side).

ajor:
> @ajor, ping given the above constraints (multiple probe reads to filter kernel tasks from the bpf side).

Yeah, sounds like the selective filtering of errors I was suggesting isn't the way to go for now.

> > Why is it that you think EFAULTs shouldn't be hidden? Are there other situations that they could occur in?
> 
> Yes, you can have stack pages paged out.
> 
> Overall, I don't think we can make the correct calls on errors, users should make the call in the vast majority of cases. (I also have opinions about -kk not being default 👀 )

Paging @brendangregg for your thoughts on all this (related to your change #428).

viktormalik:
This still seems like a relevant feature so it'd worth to take over the branch and finish it.

@ajor you seem to have been following this more closely than I was, would you find some time to push it forward?

src/bpftrace.cpp
ajor, line 1761: Am I going crazy or should we be checking `err == -EFAULT` instead of `stackid`? It's been this way for years

ajor:
Agreed that we need to improve these error messages. I'll see if I can get to it soonish

src/bpftrace.cpp
viktormalik: It certainly looks so, checking `stackid` doesn't make any sense.

danobi:
Captured in #3093. Closing for now

Issue #2306, opened 2022-07-28
Add support for strstr() builtin


liutingjieni:
I have a requirement, I hope to be able to check which process deleted the "cores" related files. 
First of all, because my need is to write a script that is deployed on thousands of servers, it is difficult for us to confirm its configuration and the directory it is mounted on because of the change of the deployment version, so we need fuzzy matching. 

The only existing str function that supports bpftrace is strncmp, which has no way to obfuscate. And there is no way to traverse strings in bpftrace, so there is no way to use strncmp to write a strstr by hand.

### Describe the solution you'd like

Solution 1: Add a new built-in func like strstr()
` bpftrace  -e 'tracepoint:syscalls:sys_enter_unlink / strstr( str(args->pathname), "cores") > 0 / {printf("comm: %s(%d) delete file: %s\n", comm, pid, str(args->pathname)); printf("%s\n", ustack());}'`

I think strstr this is a generic requirement. There may be many times when there is no way to confirm the absolute path of the file or the full name of the process or...

### Describe alternative solutions or features you've considered

Another way is to get all the information in kernel mode and then filter it in user mode. But this results in a lot of pointless copying from kernel mode to user mode. I have tested that attaching the bpftrace program to the specified function and frequently triggering it. Copying all the information to user mode will cause a CPU to reach 100% utilization. Copying after kernel-state filtering has almost no impact on performance.

I'm willing to submit a PR about this. Please leave comments if any, thanks!

viktormalik:
It could be nice to have `strstr` and since the BPF verifier now allows bounded loops and we have strings of bounded size, it should be possible to implement it. The implementation may still be tricky, but feel free to give it a try.

liutingjieni:
I have started to develop this function now, but I have not developed bpftrace before, and I am not familiar with the related libraries of llvm. I have some difficulties in development and debugging. Do you have relevant information here to help me get started quickly?

liutingjieni:
I am currently developing based on  0.12.1_release version, is this branch ok? If not, is it a recommended branch?

fbs:
All development happens against the master branch. 0.12 is quite an old release.

[...]

PR #2344, opened 2022-08-28
add strstr()

liutingjieni:
Add a new built-in func like strstr()

Add the strstr () function to make fuzzy comparisons when there is no way to confirm the absolute path of the file or the full name of the process or

usage:
`sudo bpftrace -e 't:syscalls:sys_enter_execve /strstr(comm,"sh") == 0/ { printf("%s called %s\n", comm, str(args->filename));}'`

Problem background: https://github.com/iovisor/bpftrace/issues/2306

##### Checklist

- [x] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x] The new behaviour is covered by tests

PR #2393, opened 2022-10-18
Implement strstr

liutingjieni:
Add a new built-in func like strstr()

Add the strstr () function to make fuzzy comparisons when there is no way to confirm the absolute path of the file or the full name of the process or

usage:
`sudo bpftrace -e 't:syscalls:sys_enter_execve /strstr(comm,"sh") == 0/ { printf("%s called %s\n", comm, str(args->filename));}'`

Problem background: https://github.com/iovisor/bpftrace/issues/2306

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -87,20 +87,21 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [17. `cat()`: Print file content](#17-cat-print-file-content)
     - [18. `signal()`: Send a signal to the current task](#18-signal-send-a-signal-to-current-task)
     - [19. `strncmp()`: Compare first n characters of two strings](#19-strncmp-compare-first-n-characters-of-two-strings)
-    - [20. `override()`: Override return value](#20-override-override-return-value)
-    - [21. `buf()`: Buffers](#21-buf-buffers)
-    - [22. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
-    - [23. `print()`: Print Value](#23-print-print-value)
-    - [24. `strftime()`: Formatted timestamp](#24-strftime-formatted-timestamp)
-    - [25. `path()`: Return full path](#25-path-return-full-path)
-    - [26. `uptr()`: Annotate userspace pointer](#26-uptr-annotate-userspace-pointer)
-    - [27. `kptr()`: Annotate kernelspace pointer](#27-kptr-annotate-kernelspace-pointer)
-    - [28. `macaddr()`: Convert MAC address data to text](#28-macaddr-convert-mac-address-data-to-text)
-    - [29. `cgroup_path()`: Convert cgroup id to cgroup path](#29-cgroup_path-convert-cgroup-id-to-cgroup-path)
-    - [30. `bswap`: Reverse byte order](#30-bswap-reverse-byte-order)
-    - [31. `skb_output`: Write `skb` 's data section into a PCAP file](#31-skb_output-write-skb-s-data-section-into-a-pcap-file)
-    - [32. `pton()`: Convert text IP address to byte array](#32-pton-convert-text-ip-address-to-byte-array)
-    - [33. `strerror`: Get error message for errno value](#33-strerror-get-error-message-for-errno-value)
+    - [20. `strcontains()`: Compares whether the string haystack contains the string needle.](#20-Compare-whether-the-string-haystack-contains-the-string-needle)
+    - [21. `override()`: Override return value](#20-override-override-return-value)
+    - [22. `buf()`: Buffers](#21-buf-buffers)
+    - [23. `sizeof()`: Size of type or expression](#22-sizeof-size-of-type-or-expression)
+    - [24. `print()`: Print Value](#23-print-print-value)
+    - [25. `strftime()`: Formatted timestamp](#24-strftime-formatted-timestamp)
+    - [26. `path()`: Return full path](#25-path-return-full-path)
+    - [27. `uptr()`: Annotate userspace pointer](#26-uptr-annotate-userspace-pointer)
+    - [28. `kptr()`: Annotate kernelspace pointer](#27-kptr-annotate-kernelspace-pointer)
+    - [29. `macaddr()`: Convert MAC address data to text](#28-macaddr-convert-mac-address-data-to-text)
+    - [30. `cgroup_path()`: Convert cgroup id to cgroup path](#29-cgroup_path-convert-cgroup-id-to-cgroup-path)
+    - [31. `bswap`: Reverse byte order](#30-bswap-reverse-byte-order)
+    - [32. `skb_output`: Write `skb` 's data section into a PCAP file](#31-skb_output-write-skb-s-data-section-into-a-pcap-file)
+    - [33. `pton()`: Convert text IP address to byte array](#32-pton-convert-text-ip-address-to-byte-array)
+    - [34. `strerror`: Get error message for errno value](#33-strerror-get-error-message-for-errno-value)
 - [Map Functions](#map-functions)
     - [1. Builtins](#1-builtins-2)
     - [2. `count()`: Count](#2-count-count)
@@ -2218,6 +2219,7 @@ Tracing block I/O sizes > 0 bytes
 - `cat(char *filename)` - Print file content
 - `signal(char[] signal | u32 signal)` - Send a signal to the current task
 - `strncmp(char *s1, char *s2, int length)` - Compare first n characters of two strings
+- `strcontains(const char *haystack, const char *needle)` - Compares whether the string haystack contains the string needle.
 - `override(u64 rc)` - Override return value
 - `buf(void *d [, int length])` - Returns a hex-formatted string of the data pointed to by d
 - `sizeof(...)` - Return size of a type or expression
@@ -2884,7 +2886,28 @@ Attaching 320 probes...
 @[mpv, tracepoint:syscalls:sys_enter_futex]: 403868
 ```
 
-## 20. `override()`: Override return value
+## 20. `strcontains()`: Compare whether the string haystack contains the string needle.
+
+Syntax: `strcontains(const char *haystack, const char *needle)`
+
+Return true if the string haystack contains the string needle, and zero otherwise.
+
+Examples:
+
+```
+bpftrace -e 't:syscalls:sys_enter_execve /strcontains(str(args->filename),"bin")/ { @[comm, str(args->filename)] = count(); }'  
+Attaching 1 probe...
+
+@[sh, /usr/bin/which]: 2
+@[sh, /home/liuting.0xffff/.vscode-server/bin/3b889b090b5ad5793f524b5]: 2
+@[cpuUsage.sh, /usr/bin/sleep]: 2
+@[sh, /usr/bin/ps]: 2
+@[cpuUsage.sh, /usr/bin/sed]: 4
+@[node, /bin/sh]: 6
+@[cpuUsage.sh, /usr/bin/cat]: 12
+```
+
+## 21. `override()`: Override return value
 
 Syntax: `override(u64 rc)`
 
@@ -2911,7 +2934,7 @@ ioctl(PERF_EVENT_IOC_SET_BPF): Invalid argument
 Error attaching probe: 'kprobe:vfs_read'
 ```
 
-## 21. `buf()`: Buffers
+## 22. `buf()`: Buffers
 
 Syntax: `buf(void *d [, int length])`
 
@@ -2942,7 +2965,7 @@ Datagram bytes: \x08\x00+\xb9\x06b\x00\x01Aen^\x00\x00\x00\x00KM\x0c\x00\x00\x00
 rtt min/avg/max/mdev = 19.426/19.426/19.426/0.000 ms
 ```
 
-## 22. `sizeof()`: Size of type or expression
+## 23. `sizeof()`: Size of type or expression
 
 Syntax:
 - `sizeof(TYPE)`
@@ -2975,7 +2998,7 @@ Attaching 1 probe...
 8
 ```
 
-## 23. `print()`: Print Value
+## 24. `print()`: Print Value
 
 Syntax: ```print(value)```
 
@@ -2999,7 +3022,7 @@ contains the memcopy'd value so the value at `print()` invocation time will be
 printed.  However for maps, only the handle to the map is queued up, so the
 printed map may be different than the map at `print()` invocation.
 
-## 24. `strftime()`: Formatted timestamp
+## 25. `strftime()`: Formatted timestamp
 
 Syntax:
 - `strftime(const char *format, int nsecs)`
@@ -3036,7 +3059,7 @@ Attaching 1 probe...
 ^C
 ```
 
-## 25. `path()`: Return full path
+## 26. `path()`: Return full path
 
 Syntax:
 - `path(struct path *path)`
@@ -3063,7 +3086,7 @@ Attaching 1 probe...
 /dev/pts/1 -> /dev/pts/1
 ```
 
-## 26. `uptr()`: Annotate userspace pointer
+## 27. `uptr()`: Annotate userspace pointer
 
 Syntax:
 - `uptr(void *p)`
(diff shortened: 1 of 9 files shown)


liutingjieni:
Pr was here before ( [add strstr() by liutingjieni · Pull Request #2344 · iovisor/bpftrace](https://github.com/iovisor/bpftrace/pull/2344) ). Because of the conflict with master, so based on the latest master commit changes

liutingjieni:
I see a check error, but I don't quite understand。what do I need to do now?

viktormalik:
> Pr was here before ( [add strstr() by liutingjieni · Pull Request #2344 · iovisor/bpftrace](https://github.com/iovisor/bpftrace/pull/2344) ). Because of the conflict with master, so based on the latest master commit changes

In future, please keep updating the same PR (by force-push into the branch), opening a new one after each rebase is not a good idea.

> I see a check error, but I don't quite understand。what do I need to do now?

Your new test is failing with segmentation fault, so you'll have to debug. If you want to reproduce the CI environment, you can build and use the prepared Docker images. Check `.github/workflows/ci.yml` for exact Docker commands.

liutingjieni:
I found that the error is build_test (LLVM 15 Release). I tried to remove the commit about it( Latest Commit) and I found that it works.

viktormalik:
> I found that the error is build_test (LLVM 15 Release). I tried to remove the commit about it( Latest Commit) and I found that it works.

Which could mean that your code doesn't work with LLVM 15 (very likely it doesn't work with opaque pointers). Please check and fix that.

liutingjieni:
all checks have passed. what else do I need to do now?

viktormalik:
This is a terrible solution. You can't just arbitrarily revert a commit from master if it doesn't suit you.

You have to make your code work with [opaque pointers](https://llvm.org/docs/OpaquePointers.html). Once you do that, the checks will pass.

liutingjieni:
> > This is a terrible solution. You can't just arbitrarily revert a commit from master if it doesn't suit you.
> > You have to make your code work with [opaque pointers](https://llvm.org/docs/OpaquePointers.html). Once you do that, the checks will pass.
> 
> I have changed my code with opaque pointers. The check for llvm15 has passed. The last two commit hashes from master have changed due to cherry-pick, but the code content has not changed.

liutingjieni:
I want the code to be checked

viktormalik:
> I have changed my code with opaque pointers. The check for llvm15 has passed. The last two commit hashes from master have changed due to cherry-pick, but the code content has not changed.

Ah, right, sorry. I'll review when I have time.

src/ast/irbuilderbpf.cpp
```diff
@@ -819,6 +819,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  else
+    literal1 = std::nullopt;
+
+  std::optional<std::string> literal2;
+  if (auto constString2 = dyn_cast<ConstantDataArray>(val2))
+    literal2 = constString2->getAsString();
+  else if (isa<ConstantAggregateZero>(val2))
+    literal2 = "";
+  else
+    literal2 = std::nullopt;
```
BurntBrunch, line 870: We should handle both strings being literal and not even emit the branches below.

src/ast/irbuilderbpf.cpp
```diff
@@ -819,6 +819,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+      auto *ptr_str = CreateGEP(ArrayType::get(getInt8Ty(), str1_size),
+                                val1,
+                                { getInt32(0), getInt32(j) });
+      str_c = CreateLoad(getInt8Ty(), ptr_str);
+    }
+
+    for (size_t i = 0; i < str2_size; i++)
+    {
+      BasicBlock *two_loop = BasicBlock::Create(module_.getContext(),
+                                                "two.loop",
```
BurntBrunch, line 905: Basic block names would ideally be unique. Can you append the idx in the first string here?

src/ast/irbuilderbpf.cpp
```diff
@@ -819,6 +819,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+                                                parent);
+      BasicBlock *loop_cmp_char = BasicBlock::Create(module_.getContext(),
+                                                     "strstr.loop_cmp_char",
```
BurntBrunch, line 908: Same

BurntBrunch (commented):
This looks fine for what it is but I have a meta question - why are we unrolling the loop?

For a char[32] vs char[32] call, this emits ~500 instructions.
```
> src/bpftrace -vv -e 't:mac80211:wake_queue {if (strstr(ar
gs->wiphy_name, args->wiphy_name)) {exit()}}'
INFO: node count: 13
Attaching 1 probe...

Program ID: 1422

The verifier log: 
processed 522 insns (limit 1000000) max_states_per_insn 0 total_states 32 peak_states 32 mark_read 24
```

Given that bpf has had backward branches for ages, why don't we just emit this as an actual loop with known bounds and let the verifier do the hard work? It would dramatically improve complexity while potentially also allowing probed strings to work (by emitting the loop against the probe size limit or even against the runtime string length, assuming we can satisfy the verifier).

tests/semantic_analyser.cpp
```diff
@@ -2035,6 +2035,24 @@ TEST(semantic_analyser, strncmp_posparam)
...
 
+TEST(semantic_analyser, strstr)
+{
+  // Test strncmp builtin
+  test("i:s:1 { $a = \"bar\"; strstr(\"foo\", $a) }", 0);
+  test("i:s:1 { strstr(\"foo\", \"bar\") }", 0);
+  test("i:s:1 { strstr(1) }", 1);
+  test("i:s:1 { strstr(1,1) }", 10);
+  test("i:s:1 { strstr(\"a\",1) }", 10);
+  test("i:s:1 { strstr(\"a\",\"a\") }", 0);
```
viktormalik, line 2046: Isn't this basically a duplicate of line 2042?

man/adoc/bpftrace.adoc
```diff
@@ -1633,6 +1633,16 @@ bpftrace doesn't read past the length of the shortest string.
 
 The use of the `==` and `!=` operators is recommended over calling `strncmp` directly.
 
+=== strstr
+
+.variants
+* `int64 strstr(char * s1, char * s2)`
```
fbs, line 1639: maybe call them needle/haystack?

man/adoc/bpftrace.adoc
```diff
@@ -1633,6 +1633,16 @@ bpftrace doesn't read past the length of the shortest string.
...
+
+`strstr` compares whether the string val1 contains the string val2.
+If s2 is contained `0` is returned, else a non-zero value is returned.
```
fbs, line 1642: I feel like this was discussed before but I'm not convinced by the `strstr` name.  This function doesn't actually behave the same as `strstr` does in the C libs.

Maybe `strcontains` is a better name?

From the `strstr` manpages:

>      If needle is an empty string, haystack is returned; if needle occurs nowhere in haystack, NULL is returned; otherwise a pointer to the first character of the first occurrence of needle is returned.

This one behaves completely different, it returns 0 *when found*.

fbs:
> This looks fine for what it is but I have a meta question - why are we unrolling the loop?
> 
> For a char[32] vs char[32] call, this emits ~500 instructions.
> 
> ```
> > src/bpftrace -vv -e 't:mac80211:wake_queue {if (strstr(ar
> gs->wiphy_name, args->wiphy_name)) {exit()}}'
> INFO: node count: 13
> Attaching 1 probe...
> 
> Program ID: 1422
> 
> The verifier log: 
> processed 522 insns (limit 1000000) max_states_per_insn 0 total_states 32 peak_states 32 mark_read 24
> ```
> 
> Given that bpf has had backward branches for ages, why don't we just emit this as an actual loop with known bounds and let the verifier do the hard work? It would dramatically improve complexity while potentially also allowing probed strings to work (by emitting the loop against the probe size limit or even against the runtime string length, assuming we can satisfy the verifier).

Agree but maybe we shouldn't block on it.

For a future improvement it would also be nice if we compare more bytes at once

docs/reference_guide.md
```diff
@@ -2884,7 +2886,28 @@ Attaching 320 probes...
 @[mpv, tracepoint:syscalls:sys_enter_futex]: 403868
 ```
 
-## 20. `override()`: Override return value
+## 20. `strcontains()`: Compare whether the string s1 contains the string s2.
+
+Syntax: `strcontains(const char *haystack, const char *needle)`
+
+Return zero if the string val1 contains the string val2, and non-zero otherwise.
```
viktormalik, line 2893: The description doesn't fit the parameter names.

man/adoc/bpftrace.adoc
```diff
@@ -1633,6 +1633,16 @@ bpftrace doesn't read past the length of the shortest string.
 
 The use of the `==` and `!=` operators is recommended over calling `strncmp` directly.
 
+=== strcontains
+
+.variants
+* `int64 strcontains(const char *haystack, const char *needle)`
+
+`strcontains` compares whether the string val1 contains the string val2.
```
viktormalik, line 1641: Same as above.

viktormalik (changes requested):
Plus there are several unresolved comments (and some that were incorrectly resolved) from the previous reviews.

man/adoc/bpftrace.adoc
fbs: @BurntBrunch @viktormalik  how do you feel about this rename? it got implemented now but I wanted to start a discussion :)

viktormalik: I think that it's fine, although I'd expect it to return 1 if the substring is found, so that you can test for the positive case instead of the negative one when using it in a condition.

Another option would be to introduce a new operator for this (`in`?) but that's a bigger change and would have to be discussed more, I guess.

viktormalik: Ah, sorry, I prefer `strcontains` since, as you said, `strstr` has a different semantics in C.

src/ast/irbuilderbpf.cpp
liutingjieni: sorry, I don't understand what you mean. here I just mimic the implementation of strncmp.

viktormalik: If both strings are literal, you can directly determine the result here (e.g. by using `std::string::find`) and return it.

viktormalik: Yes, it could use the same optimization.

liutingjieni:
> Plus there are several unresolved comments (and some that were incorrectly resolved) from the previous reviews.

Are there any incorrectly resolved now?

src/ast/irbuilderbpf.cpp
```diff
@@ -820,6 +819,153 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+
+  if (literal1 && literal2)
+  {
+    std::string s1 = literal1.value();
+    std::string s2 = literal2.value();
+    LOG(INFO) << "string s1: " << s1;
+    LOG(INFO) << "string s2: " << s2;
+    std::string::size_type position = s1.find(s2);
+
+    if (position != std::string::npos)
```
viktormalik, line 867: What about the other option, when `s2` is not a substring?

src/ast/irbuilderbpf.cpp
```diff
@@ -820,6 +819,153 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+    {
+      AllocaInst *t = CreateAllocaBPF(getInt1Ty(), "result");
+      CreateStore(getInt1(0), t);
+      Value *ret = CreateLoad(getInt1Ty(), t);
+      CreateLifetimeEnd(t);
+      ret = CreateIntCast(ret, getInt64Ty(), false);
+      return ret;
```
viktormalik, line 874: Isn't this way too complicated? Can't we just do:
```
return getInt64(0);
```

man/adoc/bpftrace.adoc
```diff
@@ -1633,6 +1633,16 @@ bpftrace doesn't read past the length of the shortest string.
 
 The use of the `==` and `!=` operators is recommended over calling `strncmp` directly.
 
+=== strcontains
+
+.variants
+* `int64 strcontains(const char *haystack, const char *needle)`
+
+`strcontains` compares whether the string haystack contains the string needle.
+If needle is contained `0` is returned, else a non-zero value is returned.
```
viktormalik, line 1642: Looking at this now, why return 0 when the substring is found? I'd expect a function called `strcontains` returns true if it succeeds (so that you can test for the positive case, i.e. `if (strcontains(s1, s2)) ...`

viktormalik (commented):
Could you also merge the two commits and cleanup the commit message?
Thanks!

src/ast/irbuilderbpf.cpp
```diff
@@ -820,6 +819,151 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  std::optional<std::string> literal2 = ValToString(val2);
+
+  if (literal1 && literal2)
+  {
+    std::string s1 = literal1.value();
+    std::string s2 = literal2.value();
+    s2 = s2.substr(0, s2.size() - 1);
+    std::size_t position = s1.find(s2);
+
+    if (position != s1.npos)
```
viktormalik, line 866:
```suggestion
    if (position != std::string::npos)
```

src/ast/irbuilderbpf.cpp
```diff
@@ -820,6 +819,151 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+    {
+      return getInt64(1);
+    }
+    else
+    {
+      return getInt64(2);
```
viktormalik, line 872:
```suggestion
      return getInt64(0);
```

viktormalik (approved):
Since there has been a lot of work, I'd like to have one more review before merging this.

viktormalik:
cc'ing @danobi and @ajor explicitly since this was opened before we added the CODEOWNERS file.

xh4n3:
Since this is similar to #2387, I guess you may need to add the codegen tests and change the `CHANELOG.md`.
And maybe a little change to the empty lines, see https://github.com/iovisor/bpftrace/pull/2387#discussion_r1040584036

tests/runtime/call
```diff
@@ -326,6 +326,13 @@ REQUIRES_FEATURE dpath kfunc
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
+NAME path
```
fbs, line 329: can we have a more descriptive name for this test?

tests/runtime/call
```diff
@@ -326,6 +326,13 @@ REQUIRES_FEATURE dpath kfunc
...
+RUN {{BPFTRACE}} -ve 'kfunc:filp_close { $f = path(args->filp->f_path); if (!strcontains($f, "tmp")) { printf("OK\n"); exit(); } }'
+EXPECT OK
+REQUIRES_FEATURE dpath kfunc
+TIMEOUT 5
+AFTER ./testprogs/syscall read
+
```
fbs, line 335: would also be nice to have a runtime test that just tests with a few combinations of constants to make sure that codepath works

fbs (changes requested):
looks good! Just a few suggestions on the testing side

tests/runtime/call
liutingjieni: Do you have  an example of runtime test？

fbs: `RUN {{BPFTRACE}} -ve 'BEGIN { if (!strcontains("abc", "a")) { printf("OK\n"); exit(); } }'`

tests/runtime/call
```diff
@@ -326,6 +326,18 @@ REQUIRES_FEATURE dpath kfunc
...
 AFTER ./testprogs/syscall read
 
+NAME strcontains
+RUN {{BPFTRACE}} -ve 'kfunc:filp_close { $f = path(args->filp->f_path); if (strcontains($f, "tmp")) { printf("OK\n"); exit(); } }'
+EXPECT OK
+REQUIRES_FEATURE dpath kfunc
+TIMEOUT 5
+AFTER ./testprogs/syscall read
+
+NAME strcontains runtime
```
fbs, line 336: we try to keep the test name descriptive of what the test does. 

```suggestion
NAME strcontains literals
```

Issue #2373, opened 2022-09-27
Improvements to CI


BurntBrunch:
I've spent quite a few hours dealing with bpftrace's CI and the matrix of supported builds.

As a result, I have some thoughts on possible improvements and would like to open the floor to a discussion on where to take things.

## Current setup

1. Base distros are focal, bionic, xenial, and alpine. The ubuntu flavors can run against multiple llvm versions thanks to llvm-X packages.
2. Some of the dockerfiles support building libbpf and bcc from source, others don't (this will change soon as both will get vendored)
3. Github uses kernels we don't control and so we get issues like #2356
4. Because of the way the Dockerfiles are expressed, builds are not hermetic and pull arbitrary artifacts from upstream repos.
5. Because of the way the Dockerfiles are expressed, the images are hard to use for local development (i.e. entrypoint directly builds the script instead of having the workflow `docker exec` the build script in CI)

## Possible changes

There are a few changes we could make:
* Double down on docker but use quay.io to cache and explicitly rotate the base images. This would make the builds more reproducible, while retaining much of the current setup.
* Remove entrypoints from Dockerfiles, so distrobox/toolbox can easily use the images. This is purely a developer experience type of change.
* Given that Github Runners do not support nested virtualization, either use qemu in software mode or find a sponsor for dedicated Github Runner hosts, so we can control the kernel that the tests run against.
* (Speculative!) Use nix tooling (nix-build, nix-shell) instead of docker base images to express the dependency matrix (docker would still be used for caching probably). I'm not clear how achievable this is given that nixpkgs moves linearly in time. It's not obvious to me how to get e.g. old glibc but new llvm.
  * It's possible that a from-source distro like Arch or Gentoo would make a better base distro here as we would be able to dictate the full dependency set.
  * Guix may actually be better at this (judging by [this](https://guix.gnu.org/manual/en/html_node/Package-Transformation-Options.html)) but package availability is worse than the other options.

Overall, I think using real distros is important for sanity checking but is not the best way to express the parameter space that bpftrace aims to cover. I'd love to hear people's opinions on how to control these parameters better!

BurntBrunch:
> As a matter of that, we could also provide a script which would fill the necessary docker run parameters, especially mount the necessary volumes.

I would rather not, that's literally what distrobox does (it's a series of shell scripts around podman/docker). If you read through it, there's a bunch of gotchas you need to get right.

[...]

PR #2396, opened 2022-10-24
docker: fall back to a known bpftool version

BurntBrunch:
The `linux-tools-$(uname -r)` construct that 1f0ff4c6 added only works if the host kernel is one of Ubuntu's.

If the install fails, fall back to a known package version and replace the /sbin/bpftool script with a direct symlink.

@viktormalik


viktormalik (approved):
I'm ok with merging this as a temp fix, but I'd prefer vendoring bpftool in tests.

BurntBrunch:
> I'm ok with merging this as a temp fix, but I'd prefer vendoring bpftool in tests.

Honestly, I'm a bit burned out from trying to vendor things ;) Let's land this - it helps a little bit - and we can figure out the long term CI story in #2373 and merge this in the mean time?

PR #2390, opened 2022-10-13
Sanitise program names

stevenjohnstone:
bpf objects may only have names containing alphanumeric characters, '.' and '_'. Golang programs, in particular, have symbol names which cause problems e.g. "os.(*File).Read".

Fixes #2388.

##### Checklist

- [x] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x] The new behaviour is covered by tests

PR #2399, opened 2022-10-27
runtime tests: improve test for probe name sanitization

viktormalik:
#2390 introduced a runtime test attaching to a uprobe of `/usr/bin/docker` which is normally a Golang program.

That is not always the case, as Fedora uses Podman and `/usr/bin/docker` is just a compatibility shell script. So, this PR improves the test requirement that `/usr/bin/docker` is an ELF file (instead of just checking that it exists).

The test still depends on a non-standard binary (it's not run in the CI either) so it'd be probably better implement our own Golang program to test the functionality.

##### Checklist

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -285,3 +285,9 @@ PROG k:* { @[probe] = count(); }
 EXPECT BPFTRACE_MAX_BPF_PROGS
 WILL_FAIL
 TIMEOUT 2
+
+NAME sanitise probe name
+PROG uprobe:./testprogs/uprobe_namesan:fn* { printf("ok\n"); exit(); }
+EXPECT ^ok
+AFTER ./testprogs/uprobe_namesan
+TIMEOUT 5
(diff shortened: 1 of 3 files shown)


viktormalik:
cc @stevenjohnstone

stevenjohnstone:
Hi @viktormalik! I wonder if you could save some pain related to finding/building a golang executable by compiling a C test program with a character in a function name which would be rejected by the kernel if we tried to use it as an object name? For example,

```c
#include <stdio.h>
void __attribute__ ((noinline)) fn\u03b1() {
    printf("%s\n", __func__);
}
int main(int argc, const char **argv) {
    fn\u03b1();
}
```
would fail 
```
$ bpftrace -e 'uprobe:./a.out:"fnα" {}'
Attaching 1 probe...
ERROR: Error loading program: uprobe:./a.out:fnα (try -v)
```
with versions before the sanitisation changes.

viktormalik:
Good idea! I replaced the test by your suggested one.

viktormalik:
In the end, the non-ASCII character caused problems for the Python code parsing output of tests, so I went with a dollar sign inside the function name which is permitted by GNU C but still requires the sanitisation change.

danobi (commented):
LGTM

PR #2407, opened 2022-11-02
Document min library versions and vendoring

BurntBrunch:
Anywhere else we need to document these?


viktormalik (approved):
This is sufficient, I think.

PR #2411, opened 2022-11-03
Fix compile error: bcc_elf.c: undefined reference to `debuginfod_begin'

Rtoax:
If libdebuginfod is installed on the system and bpftrace is compiled with static linking to BCC (STATIC_BPF_BCC), then libbcc.a requires the dynamic library -ldebuginfod, in which case a compilation error will be reported if bpftrace is compiled without linking libdebuginfod. We just need to add bcc's FindLibDebuginfod.cmake to bpftrace to solve this problem.

## How to reproduce the error:

```
$ sudo dnf install -y elfutils-debuginfod*
$ cd bpftrace
$ mkdir build && cd build
$ ../build-libs.sh
$ cmake -DCMAKE_INSTALL_PREFIX=/usr/ -DCMAKE_BUILD_TYPE=Debug .. $ make -j$(nproc)
...
[ 91%] Linking CXX executable bpftrace
[ 92%] Linking CXX executable bpftrace_test
/usr/bin/ld: ../build-libs/lib64/libbcc.a(bcc_elf.c.o): in function `find_debug_file': bcc_elf.c:(.text+0x10df): undefined reference to `debuginfod_begin' /usr/bin/ld: bcc_elf.c:(.text+0x10fd): undefined reference to `debuginfod_find_debuginfo' /usr/bin/ld: bcc_elf.c:(.text+0x1110): undefined reference to `debuginfod_end' collect2: error: ld returned 1 exit status
make[2]: *** [src/CMakeFiles/bpftrace.dir/build.make:121: src/bpftrace] Error 1 make[1]: *** [CMakeFiles/Makefile2:1328: src/CMakeFiles/bpftrace.dir/all] Error 2 make[1]: *** Waiting for unfinished jobs....
/usr/bin/ld: ../build-libs/lib64/libbcc.a(bcc_elf.c.o): in function `find_debug_file': bcc_elf.c:(.text+0x10df): undefined reference to `debuginfod_begin' /usr/bin/ld: bcc_elf.c:(.text+0x10fd): undefined reference to `debuginfod_find_debuginfo' /usr/bin/ld: bcc_elf.c:(.text+0x1110): undefined reference to `debuginfod_end' collect2: error: ld returned 1 exit status
[...]

PR #2413, opened 2022-11-03
Update bcc and pass ENABLE_LIBDEBUGINFOD=0

BurntBrunch:
As an alternative to iovisor/bpftrace#2411, this updates to the exact commit in bcc that adds ENABLE_LIBDEBUGINFOD and passes that flag in our local build.

Because we're not using any functionality not in v0.25, this technically does not raise the minimum bcc version for the bpftrace package - distros would build against a dynamic, system bcc anyway.


danobi (approved):

Rtoax:
OK, you're faster than me!!  :)

Issue #2402, opened 2022-10-28
flaky usdt test


xh4n3:
### What reproduces the bug?
Trigger CI on the master branch.
https://github.com/iovisor/bpftrace/actions/runs/3343605763/jobs/5536998903
https://github.com/iovisor/bpftrace/actions/runs/3341593002/jobs/5532960039

```bash
2: [ RUN      ] usdt."usdt probes - file based semaphore activation multi process"
2: [  TIMEOUT ] usdt."usdt probes - file based semaphore activation multi process"
2: 	Command: ../src//bpftrace runtime/scripts/usdt_file_activation_multiprocess.bt --usdt-file-activation
2: 	Timeout: 5
2: 	Current output: Attaching 3 probes...
2: __BPFTRACE_NOTIFY_PROBES_ATTACHED
2: found 1 processes
```

BurntBrunch:
I'm fighting with it in https://github.com/iovisor/bpftrace/pull/2370/commits/4eb280b07ee683f62176e06bdbb594c146a72379 but I can't make it stable yet. Now the pidof invocation to get the child pid after it has started is failing. I may just make that logic only run if the test needs BEFORE_PID.

viktormalik:
I think that this should be resolved by #2370

BurntBrunch:
Unfortunately, I don't think it was fully fixed, just made a bit better. I still don't fully understand how this happens (and why it happens on _that_ test specifically) but here's a run after that PR that timed out again - https://github.com/BurntBrunch/bpftrace/actions/runs/3381817256/jobs/5616467787

BurntBrunch:
Okay, I give up.  #2414 doesn't help either, so this is not about the test runner racing the BEFORE calls. Maybe it's something to do with --usdt-file-activation and whatever kernel the GH runners run?

jordalgo:
I don't see this anymore - perhaps @danobi fixed it. Closing for now.

PR #2414, opened 2022-11-03
Teach runtime runner about multiple BEFORE clauses 

BurntBrunch:
As noted in https://github.com/iovisor/bpftrace/issues/2402, usdt flakiness was made better by https://github.com/iovisor/bpftrace/commit/508538a0eaa596995cbf8ce261f05c90b08c50df but not fully fixed.

This commit is what I should have done all along: it allows the test runner to parse and wait for multiple BEFORE clauses and thus ensures the processes have started before the test runs.

There are two minor changes:

  1. The check for child processes is now `ps --ppid` based to eventually allow parallel process runs in the same environment.
  2. Because of the `ps` usage, the name check is now truncated to 15 chars, which will fail if TASK_COMM_LEN is not 16. That looks like a constant in the kernel, so I think we're good.


BurntBrunch:
While this is a good idea, it doesn't help with the usdt test flakiness. I'm starting to think the issue is in how long bcc or the kernel take.

viktormalik:
Even though this didn't resolve the flakiness problem, it's still a useful feature. So, I rebased the PR onto the master and it's now ready for review.

danobi (approved):
lgtm. But pls also update docs (https://github.com/iovisor/bpftrace/blob/master/tests/README.md#runtime-test-directives) that `BEFORE` clause can be repeated

viktormalik:
> lgtm. But pls also update docs (https://github.com/iovisor/bpftrace/blob/master/tests/README.md#runtime-test-directives) that `BEFORE` clause can be repeated

Good point, updated.

PR #2424, opened 2022-11-13
tools: Add filelife tools to track the lifecycle of a file

Rtoax:
Usage:

```
  $ sudo ./filelife.bt
  Attaching 6 probes...
  Tracing the lifespan of short-lived files, Hit Ctrl-C to end
  TIME     PID      COMM             AGE(us)        FILE
  20:01:26 109927   rm               507814         a.txt
  20:01:28 109934   rm               507448         a.txt
  ^C
```

  Test command line:

```
  $ touch a.txt && sleep 0.5 && rm a.txt
  ...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ and this project adheres to
 
 #### Docs
 #### Tools
+- Add filelife tool
+  - [#2424](https://github.com/iovisor/bpftrace/pull/2424)
 
 ## [0.16.0] 2022-08-30
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ bpftrace contains various tools, which also serve as examples of programming in 
 - tools/[cpuwalk.bt](tools/cpuwalk.bt): Sample which CPUs are executing processes. [Examples](tools/cpuwalk_example.txt).
 - tools/[dcsnoop.bt](tools/dcsnoop.bt): Trace directory entry cache (dcache) lookups. [Examples](tools/dcsnoop_example.txt).
 - tools/[execsnoop.bt](tools/execsnoop.bt): Trace new processes via exec() syscalls. [Examples](tools/execsnoop_example.txt).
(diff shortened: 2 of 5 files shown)


brendangregg:
From 2019: https://github.com/brendangregg/bpf-perf-tools-book/blob/master/originals/Ch08_FileSystems/filelife.bt

A) This version is a bit different: Kernel changes?

I did not put it (and dozens of other examples I created) in the bpftrace repo since the plan was maintain bcc libbpf/CO-RE/BTF tools as the eBPF-observability-toolkit, and not duplicate maintenance effort by shipping similar versions in bpftrace, leaving bpftrace with a smaller set as examples. I know, I know, it's such a short thing, how hard could maintenance be? see (A). :-)

I think an argument could be made that bpftrace/CO-RE/BTF/no-libllvm with 100 text tools will be a smaller install than the equivalent libbpf/CO-RE/BTF versions, with the added advantage of being customizable on the fly. In which case, we're doing it wrong by pouring effort into libbpf, and we should abandon the libbpf versions and back bpftrace. I'm interested to hear what the Facebook folk think. We should have this argument once bpftrace is buildable without llvm, and you can drop a small static bpftrace binary on an Android phone and it just works.

Rtoax:
Thanks, Gregg, actually, i start two PR in bcc to fix filelife with CO-RE way, https://github.com/iovisor/bcc/pull/4339, https://github.com/iovisor/bcc/pull/4340.

viktormalik:
So the conclustion is that we close this?

Rtoax:
I'd like to merge this PR, which is a good change https://github.com/brendangregg/bpf-perf-tools-book/blob/master/originals/Ch08_FileSystems/filelife.bt, but it's up to you and Gregg to merge or not. :)

Rtoax:
Thanks @brendangregg @viktormalik , I just close this PR. Just like Gregg said

> I did not put it (and dozens of other examples I created) in the bpftrace repo since the plan was maintain bcc libbpf/CO-RE/BTF tools as the eBPF-observability-toolkit.

Thanks again.

PR #2433, opened 2022-11-17
Probe matcher: don't fail on missing files

minipli-oss:
Even though we rely on ftrace's function tracing functionality to list available probe functions, bpftrace can work without these, e.g. when making use of tracepoints only.

Right now `bpftrace -l` will die with a `std::runtime_error` exception if called on a kernel that lacks `CONFIG_FUNCTION_TRACER`. This isn't exactly user friendly.

Instead of dying by throwing an exception, emit a warning and keep on going.

This makes bpftrace usable on systems that lack `CONFIG_FUNCTION_TRACER` and/or `CONFIG_EVENT_TRACING` -- with limited functionality, of course!

Signed-off-by: Mathias Krause <minipli@grsecurity.net>

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to
   - [#2397](https://github.com/iovisor/bpftrace/pull/2397)
 - Fix segfault for invalid AssignVarStatement visit
   - [#2423](https://github.com/iovisor/bpftrace/pull/2423)
+- Better handling of missing function trace support files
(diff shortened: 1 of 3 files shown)


viktormalik (commented):
Looks good, thanks! This is also worth a CHANGELOG entry, can you add it?

minipli-oss:
Thanks for the review!

I added the changelog entry and updated the branch.

viktormalik (approved):

PR #2440, opened 2022-11-30
Fix offset calculation for arrays of pointers on 32-bit architectures

tnovak:
When indexing into an array of pointers, or doing arithmetic on pointers to pointers, make sure the correct element/pointee size is used for offset calculation. For example, `*((int32 **)$ptr + 1)` should generate a read at offset 4 from `$ptr` on a 32-bit system.

This is a follow-up to 02b0f077 ("Use arch-dependent size when reading pointers from kernel/user memory").

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.h
+++ b/src/ast/irbuilderbpf.h
@@ -183,6 +183,13 @@ public:
                        Value *data_len,
                        const location &loc);
 
+  // For a type T, creates an integer expression representing the byte offset
+  // of the element at the given index in T[]. Used for array dereferences and
+  // pointer arithmetic.
+  llvm::Value *CreatePtrOffset(const SizedType &type,
+                               llvm::Value *index,
+                               AddrSpace as);
+
   StoreInst *createAlignedStore(Value *val, Value *ptr, unsigned align);
   // moves the insertion point to the start of the function you're inside,
   // invokes functor, then moves the insertion point back to its original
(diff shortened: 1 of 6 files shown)


tests/runtime/array
```diff
@@ -142,3 +142,9 @@ RUN {{BPFTRACE}} -e 'struct A { int x[4]; } uprobe:./testprogs/array_access:test
 EXPECT Result: 1
 TIMEOUT 5
 AFTER ./testprogs/array_access
+
+NAME array of pointers element access
+PROG struct A { int *x[4]; } uprobe:./testprogs/array_access:test_ptr_array { @x = *((struct A*)arg0)->x[1]; exit(); }
```
ajor, line 147: Just to match the naming in array_access.c (I got confused on the first read of this test):
```suggestion
PROG struct C { int *z[4]; } uprobe:./testprogs/array_access:test_ptr_array { @x = *((struct C*)arg0)->z[1]; exit(); }
```

src/ast/irbuilderbpf.h
```diff
@@ -183,6 +183,14 @@ class IRBuilderBPF : public IRBuilder<>
                        Value *data_len,
                        const location &loc);
 
+  // For a type T, creates an integer expression representing the byte offset
+  // of the element at the given index in T[]. Used for array dereferences and
+  // pointer arithmetic.
+  llvm::Value *CreatePtrOffset(
+      const SizedType &type,
+      llvm::Value *index,
+      std::optional<AddrSpace> addrSpace = std::nullopt);
```
ajor, line 192: Any reason to make this parameter optional when it's always being passed in? Unless it's needed, I'd do without it for simplicity and let someone make it optional in the future if it's useful

ajor (commented):
Good spot! Thanks for fixing this

ajor (approved):
Looks good!

PR #2454, opened 2022-12-13
INSTALL.md: Add more details about git submodules

sevan:
Need to init before update.


viktormalik (commented):
Thanks! Just a small rewording, making it a bit simpler.

viktormalik (approved):

PR #2456, opened 2022-12-13
Fix attaching to multiple USDT probes using the same wildcard

Kaufi-Jonas:
Hi,
this is my first pull request in a public repository. Let me know what I can improve :)

Short description of the issue:
Attaching to multiple USDT probes with different number of locations by the same wildcard can produce the error 'code not generated for probe'. The reason is that a single AttachPoint has multiple matching probes in this case but we only store the USDT information for a single probe in AttachPoint. Later, the number of locations from the USDT information stored in AttachPoint is used to look up the generated code for every probe to attach to, which can differ between the USDT probes and cause the above mentioned error.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -221,6 +221,14 @@ int BPFtrace::add_probe(ast::Probe &p)
         attach_point->target = target;
         attach_point->ns = ns;
         attach_point->func = func_id;
+        // Necessary for correct number of locations if wildcard expands to
+        // multiple probes.
+        std::optional<usdt_probe_entry> usdt;
+        if (attach_point->need_expansion &&
+            (usdt = USDTHelper::find(this->pid(), target, ns, func_id)))
+        {
+          attach_point->usdt = *usdt;
+        }
       }
       else if (probetype(attach_point->provider) == ProbeType::tracepoint ||
                probetype(attach_point->provider) == ProbeType::uprobe ||
--- a/tests/runtime/usdt
+++ b/tests/runtime/usdt
@@ -142,7 +142,7 @@ TIMEOUT 5
 
 NAME "usdt probes - attach to probe on multiple files by wildcard"
 PROG usdt:./testprogs/usdt*::* { printf("here\n" ); exit(); }
-EXPECT Attaching 41 probes...
+EXPECT Attaching 48 probes...
 TIMEOUT 5
 
 NAME "usdt probes - attach to probe on multiple providers by wildcard and pid"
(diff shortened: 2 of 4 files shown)


tests/runtime/usdt
```diff
@@ -152,6 +152,11 @@ TIMEOUT 5
 BEFORE ./testprogs/usdt_test
 REQUIRES ./testprogs/usdt_test should_not_skip
 
+NAME "usdt probes - attach to multiple probes with different number of locations by wildcard"
+RUN {{BPFTRACE}} -e 'usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe* { printf("here\n" ); exit(); }' -c ./testprogs/usdt_multiple_locations
+EXPECT Attaching 6 probes...
+TIMEOUT 5
```
viktormalik, line 158:
```suggestion
TIMEOUT 5
REQUIRES ./testprogs/usdt_multiple_locations should_not_skip
```
This is necessary so that the check for SystemTap headers in `main` works.

tests/runtime/usdt
```diff
@@ -152,6 +152,11 @@ TIMEOUT 5
 BEFORE ./testprogs/usdt_test
 REQUIRES ./testprogs/usdt_test should_not_skip
 
+NAME "usdt probes - attach to multiple probes with different number of locations by wildcard"
+RUN {{BPFTRACE}} -e 'usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe* { printf("here\n" ); exit(); }' -c ./testprogs/usdt_multiple_locations
```
viktormalik, line 156: This is the preferred syntax:
```suggestion
PROG usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe* { printf("here\n" ); exit(); }'
BEFORE ./testprogs/usdt_multiple_locations
```

src/bpftrace.cpp
```diff
@@ -221,6 +221,13 @@ int BPFtrace::add_probe(ast::Probe &p)
         attach_point->target = target;
         attach_point->ns = ns;
         attach_point->func = func_id;
+        // Necessary for correct number of locations if wildcard expands to
+        // multiple probes.
+        auto usdt = USDTHelper::find(this->pid(), target, ns, func_id);
+        if (usdt)
+        { // make existing unittests pass
```
viktormalik, line 228: Unnecessary comment, I believe.

viktormalik (commented):
Hi, thanks for the PR!

I've couple of comments in the code, plus one general question: what will happen if we don't give a wildcard but still attach to a probe with multiple locations (`usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe`)?

Also I'd prefer all the commits squashed into one since the changes (in the code and in the tests) are coupled together.

Kaufi-Jonas:
> I've couple of comments in the code, plus one general question: what will happen if we don't give a wildcard but still attach to a probe with multiple locations (`usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe`)?

In that case everything works as expected. An `AttachPoint` is created for `usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe` and the USDT information stored in it is correct meaning we use the correct number of locations in https://github.com/iovisor/bpftrace/blob/517fb1a05012ae47663ae5a30c166e4f57d6e576/src/bpftrace.cpp#L273-L281

The issue I am fixing only arises when a single `AttachPoint` refers to multiple USDT probes.

viktormalik (approved):

PR #2464, opened 2022-12-21
Add support for string literal concatenation

### LINKED ISSUE #2458: String Literal Concatenation
It would be nice to support string literal concatenation like in C, i.e.
```c
printf("Hello " "world");
```
The preprocessor concatenated these string literals and the program prints "Hello world".

But with `bpftrace`, I get `ERROR: syntax error, unexpected string, expecting ) or ","`.

This isn't a very critical feature to have, but it seems like it wouldn't be extremely complicated to implement and it would provide some nice quality of life improvements where you could do things like:
```c
#define LOG_PREFIX "[+] logprefix: "
...
printf(LOG_PREFIX "my log message");
```

gsingh93:
Adds support for string literal concatenation. For example, the text `"Foo" "Bar"` will be converted to `"FooBar"` during parsing.

I'm not familiar with the codebase and it's been many years since using flex/bison, so apologies if this isn't the right approach to do this. Let me know if you have any suggestions or feedback.

##### Checklist

Fixes #2458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to
   - [#2315](https://github.com/iovisor/bpftrace/pull/2315)
 - Add %rh option to print buffer as hex without \x
   - [#2445](https://github.com/iovisor/bpftrace/pull/2445)
+- Add support for string literal concatenation
+  - [#2464](https://github.com/iovisor/bpftrace/pull/2464)
 #### Changed
 - Raise minimum versions for libbpf and bcc and vendor them for local builds
   - [#2369](https://github.com/iovisor/bpftrace/pull/2369)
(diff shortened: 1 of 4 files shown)


gsingh93:
Seems like this doesn't work with macros, which was the main use case for it :(

I'll add a test, in the meantime any suggestions for how to fix would be appreciated.

viktormalik:
> Seems like this doesn't work with macros, which was the main use case for it :(

Interesting, I'd expect this approach to work with the way macros are handled at the moment. 

Macros are expanded inside the second pass through [lexer.l](https://github.com/iovisor/bpftrace/blob/master/src/lexer.l#L221-L233) so I'd expect that the identifier (macro name) is properly replaced by the respective string and the concatenation would work.

> I'll add a test, in the meantime any suggestions for how to fix would be appreciated.

An alternative approach could be to introduce a "string concatenation" operation (either by a `+` operator or by a new `strcat` builtin) and then do the concatenation either internally (for literals only) or inside the BPF program.

danobi:
Closed due to inactivity. Looks like someone else is gonna give this a go

PR #2475, opened 2023-01-18
Support ringbuf output

xh4n3:
This PR supports `bpf_ringbuf_output` in bpftrace, as a continuation of PR #1458 .

For most async builtins, it uses feature detection to decide whether use perf-event or ringbuf. However, `bpf_skb_output` cannot send to ringbuf, so we have to keep ringbuf and perf-event both running to consume events. In this case, I use a thread to poll events from ringbuf.

A couple of details we should discuss beforehand:
1. Should we have a event_loss_counter like the former PR does? One alternative is to add return value for functions like printf(), so user can use it on their own.
2. There are a bunch of codegen tests need to be updated, I updated one of them as an example in the commit, is it okay? One alternative is to add feature variation to codegen tests, like [this branch](https://github.com/xh4n3/bpftrace/tree/codegen-tests-variation) does.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1336,14 +1336,107 @@ void IRBuilderBPF::CreateGetCurrentComm(Value *ctx,
   CreateHelperErrorCond(ctx, call, libbpf::BPF_FUNC_get_current_comm, loc);
 }
 
+void IRBuilderBPF::CreateOutput(Value *ctx,
+                                Value *data,
+                                size_t size,
+                                const location *loc)
+{
+  assert(ctx && ctx->getType() == getInt8PtrTy());
+  assert(data && data->getType()->isPointerTy());
+
+  if (bpftrace_.feature_->has_map_ringbuf())
+  {
+    CreateRingbufOutput(data, size, loc);
+  }
+  else
+  {
+    CreatePerfEventOutput(ctx, data, size, loc);
+  }
+}
+
+void IRBuilderBPF::CreateRingbufOutput(Value *data,
+                                       size_t size,
+                                       const location *loc)
+{
+  Value *map_ptr = CreateBpfPseudoCallId(
+      bpftrace_.maps[MapManager::Type::Ringbuf].value()->id);
+
+  // long bpf_ringbuf_output(void *ringbuf, void *data, u64 size, u64 flags)
+  FunctionType *ringbuf_output_func_type = FunctionType::get(
+      getInt64Ty(),
+      { map_ptr->getType(), data->getType(), getInt64Ty(), getInt64Ty() },
+      false);
+
+  Value *ret = CreateHelperCall(libbpf::BPF_FUNC_ringbuf_output,
+                                ringbuf_output_func_type,
+                                { map_ptr, data, getInt64(size), getInt64(0) },
+                                "ringbuf_output",
+                                loc);
+
+  Function *parent = GetInsertBlock()->getParent();
+  BasicBlock *loss_block = BasicBlock::Create(module_.getContext(),
+                                              "event_loss_counter",
+                                              parent);
+  BasicBlock *merge_block = BasicBlock::Create(module_.getContext(),
+                                               "counter_merge",
+                                               parent);
+  Value *condition = CreateICmpSLT(ret, getInt64(0), "ringbuf_loss");
+  CreateCondBr(condition, loss_block, merge_block);
+
+  SetInsertPoint(loss_block);
+  CreateAtomicIncCounter(
+      bpftrace_.maps[MapManager::Type::RingbufLossCounter].value()->id,
+      bpftrace_.rb_loss_cnt_key_);
+  CreateBr(merge_block);
+
+  SetInsertPoint(merge_block);
+}
+
+void IRBuilderBPF::CreateAtomicIncCounter(int mapid, uint32_t idx)
+{
+  AllocaInst *key = CreateAllocaBPF(getInt32Ty(), "key");
+  CreateStore(getInt32(idx), key);
+
+  CallInst *call = createMapLookup(mapid, key);
+  Function *parent = GetInsertBlock()->getParent();
+  BasicBlock *lookup_success_block = BasicBlock::Create(module_.getContext(),
+                                                        "lookup_success",
+                                                        parent);
+  BasicBlock *lookup_failure_block = BasicBlock::Create(module_.getContext(),
+                                                        "lookup_failure",
+                                                        parent);
+  BasicBlock *lookup_merge_block = BasicBlock::Create(module_.getContext(),
+                                                      "lookup_merge",
+                                                      parent);
+
+  Value *condition = CreateICmpNE(
+      CreateIntCast(call, getInt8PtrTy(), true),
+      ConstantExpr::getCast(Instruction::IntToPtr, getInt64(0), getInt8PtrTy()),
+      "map_lookup_cond");
+  CreateCondBr(condition, lookup_success_block, lookup_failure_block);
+
+  SetInsertPoint(lookup_success_block);
+  Value *val = CreatePointerCast(call, getInt64Ty()->getPointerTo());
+  CREATE_ATOMIC_RMW(AtomicRMWInst::BinOp::Add,
+                    val,
+                    getInt64(1),
+                    1,
+                    AtomicOrdering::SequentiallyConsistent);
+  CreateBr(lookup_merge_block);
+
+  SetInsertPoint(lookup_failure_block);
+  // ignore lookup failure
+  CreateBr(lookup_merge_block);
+
+  SetInsertPoint(lookup_merge_block);
+  CreateLifetimeEnd(key);
+}
+
 void IRBuilderBPF::CreatePerfEventOutput(Value *ctx,
                                          Value *data,
                                          size_t size,
                                          const location *loc)
 {
-  assert(ctx && ctx->getType() == getInt8PtrTy());
-  assert(data && data->getType()->isPointerTy());
-
   Value *map_ptr = CreateBpfPseudoCallId(
       bpftrace_.maps[MapManager::Type::PerfEvent].value()->id);
 
@@ -1584,7 +1677,7 @@ void IRBuilderBPF::CreateHelperError(Value *ctx,
 
   auto &layout = module_.getDataLayout();
   auto struct_size = layout.getTypeAllocSize(helper_error_struct);
-  CreatePerfEventOutput(ctx, buf, struct_size, &loc);
+  CreateOutput(ctx, buf, struct_size, &loc);
   CreateLifetimeEnd(buf);
 }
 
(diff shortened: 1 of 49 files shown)


xh4n3:
@viktormalik Thanks! I've updated the commits according to your suggestions.

src/map.cpp
```diff
@@ -139,16 +139,29 @@ Map::Map(const SizedType &type) : IMap(type)
   }
 }
 
-Map::Map(libbpf::bpf_map_type map_type) : IMap(map_type)
+Map::Map(libbpf::bpf_map_type map_type, uint64_t perf_rb_pages) : IMap(map_type)
```
viktormalik, line 142: This whole thing of using different `Map` constructors for different kinds of maps is terribly ugly, but you could probably follow it and create a separate constructor `Map(libbpf::bpf_map_type, uint64_t)` for ringbuf maps 
while using the original constructor `Map(libbpf::bpf_map_type)` for perf events map.

Anyways, this should be refactored as it took me a while to figure out what's going on.

tests/codegen/llvm/array_integer_equal_comparison.ll
```diff
@@ -32,84 +35,102 @@ entry:
...
-  %21 = bitcast i1* %arraycmp.result to i8*
-  call void @llvm.lifetime.end.p0i8(i64 -1, i8* %21)
-  %22 = bitcast i32* %ll to i8*
-  call void @llvm.lifetime.end.p0i8(i64 -1, i8* %22)
-  %23 = bitcast i32* %rr to i8*
-  call void @llvm.lifetime.end.p0i8(i64 -1, i8* %23)
-  %24 = zext i1 %20 to i64
-  %true_cond = icmp ne i64 %24, 0
-  br i1 %true_cond, label %if_body, label %if_end
+while_cond:                                       ; preds = %arraycmp.loop, %entry
```
viktormalik, line 67: This looks like it's using the loop version instead of the unrolled one which should be used by this test, shouldn't it?

src/bpftrace.cpp
```diff
@@ -1308,10 +1325,60 @@ int BPFtrace::setup_perf_events()
...
+  }
+  // Calls perf_reader_free() on all open perf buffers.
+  open_perf_buffers_.clear();
+}
+
+void BPFtrace::poll_output(bool drain)
+{
+  std::thread t;
+  if (feature_->has_map_ringbuf())
+    t = std::thread(&bpftrace::BPFtrace::poll_ringbuf, this, drain);
```
viktormalik, line 1372: I'm wondering if we could spawn a new thread only for the case when both ringbuf and perf_events are in use (i.e. `skboutput` is used). We could also do similar for setup.

Out of curiosity: is this async approach even necessary? Isn't it possible to poll ringbuf and perf_events one after another?

xh4n3:
@viktormalik Thank you! I've updated these commits as you suggest, except below:

> Out of curiosity: is this async approach even necessary? Isn't it possible to poll ringbuf and perf_events one after another?

I found this might be tricky, as these two have different condition for exiting. Might lost event?

viktormalik (approved):
LGTM, thanks!

I'd still like to have one more review as I'm not that knowledgeable in this area.

src/bpftrace.h
```diff
@@ -194,6 +196,8 @@ class BPFtrace
   int helper_check_level_ = 0;
   uint64_t ast_max_nodes_ = 0; // Maximum AST nodes allowed for fuzzing
   std::optional<struct timespec> boottime_;
+  uint32_t rb_loss_cnt_key_ = 0;
+  uint64_t rb_loss_cnt_val_ = 0;
```
danobi, line 200: Pls mark as `constexpr static` to indicate these are constants.

src/ast/irbuilderbpf.cpp
```diff
@@ -1336,14 +1336,107 @@ void IRBuilderBPF::CreateGetCurrentComm(Value *ctx,
...
+                    AtomicOrdering::SequentiallyConsistent);
+  CreateBr(lookup_merge_block);
+
+  SetInsertPoint(lookup_failure_block);
+  // ignore lookup failure
+  CreateBr(lookup_merge_block);
+
+  SetInsertPoint(lookup_merge_block);
+  CreateLifetimeEnd(key);
+}
```
danobi, line 1433: This looks fine and is probably useful for other stuff in the future, but an alternative and possibly simpler design is to use a per-cpu map.

src/bpftrace.cpp
```diff
@@ -1308,10 +1331,75 @@ int BPFtrace::setup_perf_events()
...
+            maps[MapManager::Type::RingbufLossCounter].value()->mapfd_,
+            &rb_loss_cnt_key_,
+            &value))
+    {
+      LOG(ERROR) << "fail to get ringbuf loss counter";
+    }
+    if (value)
+    {
+      out_->lost_events(value);
+    }
```
danobi, line 1368: Looks like this will only print at the end of the trace session. Perhaps fine, but I wonder if we should be sampling the loss counter every N events or something. So that lost events are displayed to the user during the trace session (rather than at the end).

src/bpftrace.cpp
```diff
@@ -1308,10 +1331,75 @@ int BPFtrace::setup_perf_events()
...
+  if (feature_->has_map_ringbuf())
+  {
+    if (resources.needs_perf_event_map)
+    {
+      // poll events from ringbuf and perf_event at the same time.
+      std::thread t;
+      t = std::thread(&bpftrace::BPFtrace::poll_ringbuf, this, drain);
+      poll_perf_events(drain);
+      t.join();
+    }
```
danobi, line 1387: Rather than do the thread dance, please check if we can register the ringbuf fd with the epoll in `poll_perf_events()`. Then we can just rename `poll_perf_events()` and reduce the duplicated code and remove multi-threading. I'd rather not have to think about thread safety if possible.

danobi (changes requested):

danobi:
Hi @xh4n3 , sorry about the delay. It looks like I didn’t get notified about your commit pushes (my email filters must be buggy or something). Is this PR ready for another review pass?

xh4n3:
> Is this PR ready for another review pass?

Hi @danobi , apart from the per-cpu loss counter part (which need your comments), I've already made all the other changes. Please help reviewing them, thanks!

src/bpftrace.cpp
```diff
@@ -1264,14 +1270,30 @@ int BPFtrace::run(BpfBytecode bytecode)
...
+{
+  // when ringbuf is available, setup ringbuf for build-ins like printf, cat.
+  if (feature_->has_map_ringbuf())
+  {
+    int err = setup_ringbuf();
+    if (err)
+      return err;
+  }
+  // when ringbuf is unavailable or built-in skboutput is used, setup perf_event
+  if (!feature_->has_map_ringbuf() || resources.needs_perf_event_map)
```
danobi, line 1290: How about wrapping `!feature_->has_map_ringbuf() || resources.needs_perf_event_map` and `feature_->has_map_ringbuf()` into two helper functions instead? I see these conditionals used in a couple places and it would make it less error prone and easier to change to have a wrapper.

src/bpftrace.cpp
```diff
@@ -1363,6 +1432,49 @@ void BPFtrace::poll_perf_events(bool drain)
...
+          const_cast<uint32_t *>(&rb_loss_cnt_key_),
+          &value))
+  {
+    LOG(ERROR) << "fail to get ringbuf loss counter";
+  }
+  if (value)
+  {
+    if (reset_rb_loss_counter())
+      LOG(ERROR) << "fail to reset ringbuf loss counter";
+    out_->lost_events(value);
```
danobi, line 1474: Seems it's possible to lose loss counts here (prog increments counter between when userspace reads and resets to 0). How about keeping track of the last read count and printing the delta each time? I think that should prevent races, right?

src/bpftrace.cpp
```diff
@@ -1313,37 +1335,84 @@ int BPFtrace::setup_perf_events()
...
+        continue;
+      }
+      if (BPFtrace::exitsig_recv || (ready == 0 && (drain || finalize_)))
+      {
+        // Stop if either
+        //   * exit signal is received
+        //   * There's no events left and we've been instructed to drain or
+        //     finalization has been requested through exit() builtin.
+        do_poll_ringbuf = false;
+      }
```
danobi, line 1411: This is duplicated from above. Any way to avoid duplication? Maybe put the each conditional into a local variable in the loop body?

src/ast/irbuilderbpf.cpp
```diff
@@ -1336,14 +1336,107 @@ void IRBuilderBPF::CreateGetCurrentComm(Value *ctx,
...
+
 void IRBuilderBPF::CreatePerfEventOutput(Value *ctx,
```
danobi, line 1435: Change this method to private? That would help statically enforce nobody inside codegen_llvm.cpp accidentally misses using `CreateOutput`

danobi (commented):
Few small comments. Looking good overall

src/ast/irbuilderbpf.cpp
danobi: Just to be clear, I was thinking out loud. Not necessary to do this for this PR

xh4n3:
@danobi Thank you for your patient review.

These are really good suggestions and I have made changes to the commits.

src/bpftrace.cpp
```diff
@@ -1313,37 +1333,90 @@ int BPFtrace::setup_perf_events()
...
+        // Some callbacks returned errors, skip and run again
+        continue;
+      }
+      if (BPFtrace::exitsig_recv || (ready == 0 && should_stop))
+      {
+        // Stop if either
+        //   * exit signal is received
+        //   * no events left and we should stop
+        do_poll_ringbuf = false;
+      }
```
danobi, line 1415: As far as I can tell the only difference between these conditionals for ringbuf and the conditionals for perf is that perf explicitly handles the EINTR case which doesn't seem necessary. I think the logic can be deduplicated with a lambda. Maybe something like:

```
auto should_retry = [](int ready) {
  return ready < 0 && errno == EINTR && !BPFtrace::exitsig_recv;
};
auto should_stop = [](int ready) {
  return ready < 0 || (ready == 0 && (drain || finalize_));
};
```

Then you could have something like:

```
if (do_poll_perf_event) {
  ...
  if (should_retry(ready)
    if (!do_poll_ringbuf)
          continue;  

  if (should_stop(ready)
    do_poll_perf_event = false;
}

if (do_poll_ringbuf) {
  ...
  if (should_retry(ready)
          continue;  

  if (should_stop(ready)
    do_poll_ringbuf = false;
}
```

I think this would be a lot cleaner. Do you think this would work or am I missing something?

danobi (approved):
lgtm! thank you for persevering this long. this is a good feature addition.

(note to whoever merges: probably squash the commits)

viktormalik:
Looks good, it's great that you managed to get rid of the multithreading :-)

I'm going to squash the commits if you don't mind.

xh4n3:
@danobi @viktormalik Thank you for reviewing this, much appreciated! I've squash the commits. 😃

viktormalik:
Even better :-) Merged.

Issue #2474, opened 2023-01-17
CI failed on LLVM 14


xh4n3:
The LLVM 14 Release test keeps failing, both on Github Action of my forked master branch and on my ubuntu machine. Because the docker image is cached, all actions on this repo are not affected yet.

Error message: 
```bash
[ 93%] Building C object src/cc/CMakeFiles/bcc-shared.dir/bcc_proc.c.o
make[3]: *** No rule to make target '/usr/lib/llvm-14/lib/libPolly.a', needed by 'src/cc/libbcc.so.EAD-HASH-NOTFOUND'.  Stop.
make[3]: *** Waiting for unfinished jobs....
[ 96%] Building CXX object src/cc/CMakeFiles/bcc-shared.dir/common.cc.o
make[2]: *** [CMakeFiles/Makefile2:430: src/cc/CMakeFiles/bcc-shared.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:437: src/cc/CMakeFiles/bcc-shared.dir/rule] Error 2
make: *** [Makefile:244: bcc-shared] Error 2
```

It seems related to this [issue](https://discourse.llvm.org/t/ubuntu-packages-with-llvm-14-then-15-suddenly-looking-for-libpolly-a/67655), I tried downgrade the llvm-14 to a earlier version, the build passed.

### What reproduces the bug?
Run LLVM-14 Release CI tests locally.

viktormalik:
I noticed the same today. Probably an LLVM bug, let's see if they resolve it within few days.

If not or if we need to merge something, we can temporarily disable LLVM 14 in the CI.

Issue #682, opened 2019-05-26
Should require struct/union prefix when casting


ajor:
We're allowed to skip the struct/union prefix when casting as a shorthand at the moment, but it can cause problems:

test.h:
```
struct Foo
{
  int x;
};

typedef union
{
  int a;
  int b;
} Foo;
```

This script allows some invalid field accesses:
```
#include "test.h"

BEGIN
{
  ((Foo*)0)->a;
  ((Foo*)0)->b;
  ((Foo*)0)->x;
}
```

This is what should be required:
```
#include "test.h"

BEGIN
{
  ((Foo*)0)->a;
  ((Foo*)0)->b;
  ((struct Foo*)0)->x;
}
```

mmarchini:
@brendangregg thoughts?

PR #2477, opened 2023-01-19
Allow working with all probe params (kfunc, uprobe)

### LINKED ISSUE #2436: get all arguments values of a function as a string in uprobe and kprobe. something like $$parms in systemtap do.
I want to print all argument values of many traced functions with uprobe and kprobe. 

provide api to get function's all argument values as a string. some like $$parms in systemtap. https://sourceware.org/systemtap/SystemTap_Beginners_Guide/targetvariables.html

Or simpler just provide a variable for the number of argument of each traced function. Then I can just write a loop myself to construct the arguments in a string.

viktormalik:
Allows to use `*args` in multiple constructions such as:
- printing (resolves #2436)
```
# bpftrace -e 'kfunc:vfs_read { print(*args); exit(); }'
Attaching 1 probe...
{ .file = 0xffff9fbe928a7e00, .buf = 0x7ffddb859ee0, .count = 16, .pos = 0xffffb203068afee0 }
```
- storing in map
```
# bpftrace -e 'kfunc:vfs_read { @ = *args; exit(); }'
Attaching 1 probe...

@: { .file = 0xffff9fbe928a7e00, .buf = 0x7ffddb859ee0, .count = 16, .pos = 0xffffb203068afe88 }
```
- map keys
```
# bpftrace -e 'kfunc:vfs_read { @[*args] = count(); } i:ms:100 { exit(); }'
Attaching 2 probes...

[...]
@[{.file=0xffff9fbe8164df00,.buf=0x3434008bd400,.count=1024,.pos=0xffffb2030f24ff00}]: 1
@[{.file=0xffff9fbe4496e400,.buf=0x7ffe84009f10,.count=64,.pos=0xffffb20300757e60}]: 2
@[{.file=0xffff9fbf0c666100,.buf=0x7fff1d7d8577,.count=1,.pos=0x0}]: 2

```

The main change is that probe args are now represented using a special record with `struct <probename>_args` type and hence can be treated as any other record type. Construction of the type is different for kfuncs and uprobes:
- kfunc: the `ctx` pointer points to the struct of args which has exactly the same layout as our representation, so the implementation is straightforward.
- uprobes: args are stored in registers, so when `*args` occurs in the code, the args record is constructed on BPF stack and filled from registers. Note: this is done every time, even when only `args->x` is used (and it would be sufficient to read one register). The reason for this is that creating entire `*args` elegantly fits the current implementation and since there are at most 6 args supported, this should be quite a cheap operation.

This required some minor changes, especially to codegen tests, see commit messages for more details.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1589,6 +1589,40 @@ Value *IRBuilderBPF::CreateRawTracepointArg(Value *ctx,
   return expr;
 }
 
+Value *IRBuilderBPF::CreateUprobeArgsRecord(Value *ctx,
+                                            const SizedType &args_type)
+{
+  assert(args_type.IsRecordTy());
+
+  auto *args_t = UprobeArgsType(args_type);
+  AllocaInst *result = CreateAllocaBPF(args_t, "args");
+
+  for (auto &arg : args_type.GetFields())
+  {
+    assert(arg.type.is_funcarg);
+    Value *arg_read = CreateRegisterRead(
+        ctx, "arg" + std::to_string(arg.type.funcarg_idx));
+    if (arg.type.GetSize() != 8)
+      arg_read = CreateTrunc(arg_read, GetType(arg.type));
+    CreateStore(arg_read,
+                CreateGEP(args_t,
+                          result,
+                          { getInt64(0), getInt32(arg.type.funcarg_idx) }));
+  }
+  return result;
+}
+
+llvm::Type *IRBuilderBPF::UprobeArgsType(const SizedType &args_type)
+{
+  auto type_name = args_type.GetName();
+  type_name.erase(0, strlen("struct "));
+
+  std::vector<llvm::Type *> arg_types;
+  for (auto &arg : args_type.GetFields())
+    arg_types.push_back(GetType(arg.type));
+  return GetStructType(type_name, arg_types, false);
+}
+
 Value *IRBuilderBPF::CreateRegisterRead(Value *ctx, const std::string &builtin)
 {
   int offset;
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -436,6 +436,9 @@ void SemanticAnalyser::visit(Builtin &builtin)
       auto type_name = probe_->args_typename();
       builtin.type = CreateRecord(type_name,
                                   bpftrace_.structs.Lookup(type_name));
+      if (builtin.type.GetFieldCount() == 0)
+        LOG(ERROR, builtin.loc, err_) << "Cannot read function parameters";
+
       builtin.type.MarkCtxAccess();
       builtin.type.is_funcarg = true;
       builtin.type.SetAS(type == ProbeType::uprobe ? AddrSpace::user
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -1483,7 +1483,7 @@ void SemanticAnalyser::visit(Map &map)
         map.vargs->at(i) = cast;
         expr = cast;
       }
-      else if (expr->type.IsCtxAccess())
+      else if (expr->type.IsPtrTy() && expr->type.IsCtxAccess())
       {
         // map functions only accepts a pointer to a element in the stack
         LOG(ERROR, map.loc, err_) << "context cannot be used as a map key";
@@ -3204,8 +3204,11 @@ void SemanticAnalyser::assign_map_type(const Map &map, const SizedType &type)
 {
   const std::string &map_ident = map.ident;
 
-  if (type.IsRecordTy() && (type.is_funcarg || type.is_tparg))
-    LOG(ERROR, map.loc, err_) << "Storing args in maps is not supported";
+  if (type.IsRecordTy() && type.is_tparg)
+  {
+    LOG(ERROR, map.loc, err_)
+        << "Storing tracepoint args in maps is not supported";
+  }
 
   auto *maptype = get_map_type(map);
   if (maptype)
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2445,12 +2445,6 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
 
   auto &assignTy = assignment.expr->type;
 
-  auto *builtin = dynamic_cast<Builtin *>(assignment.expr);
-  if (builtin && builtin->ident == "args" && builtin->type.is_funcarg)
-  {
-    LOG(ERROR, assignment.loc, err_) << "args cannot be assigned to a variable";
-  }
-
   if (search != variable_val_[probe_].end())
   {
     if (search->second.IsNoneTy())
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -443,6 +443,9 @@ void SemanticAnalyser::visit(Builtin &builtin)
       builtin.type.is_funcarg = true;
       builtin.type.SetAS(type == ProbeType::uprobe ? AddrSpace::user
                                                    : AddrSpace::kernel);
+      // We'll build uprobe args struct on stack
+      if (type == ProbeType::uprobe)
+        builtin.type.is_internal = true;
     }
     else if (type != ProbeType::tracepoint) // no special action for tracepoint
     {
@@ -1966,6 +1969,7 @@ void SemanticAnalyser::visit(Unop &unop)
       if (type.IsCtxAccess())
         unop.type.MarkCtxAccess();
       unop.type.is_btftype = type.is_btftype;
+      unop.type.is_internal = type.is_internal;
       unop.type.SetAS(type.GetAS());
     }
     else if (type.IsRecordTy())
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to
   - [#2588](https://github.com/iovisor/bpftrace/pull/2588)
 - Support all iterators
   - [#2630](https://github.com/iovisor/bpftrace/pull/2630)
+- Improve working with all probe params (kfunc, uprobe)
+  - [#2477](https://github.com/iovisor/bpftrace/pull/2477)
 #### Changed
 - Make `args` a structure (instead of a pointer)
   - [#2578](https://github.com/iovisor/bpftrace/pull/2578)
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1902,6 +1902,9 @@ NetworkManager:1155 /var/lib/sss/mc/passwd (deleted)
 - `sarg0`, `sarg1`, ..., `sargN`. - Arguments to the traced function (for programs that store arguments
   on the stack); assumed to be 64 bits wide
 - `retval` - Return value from traced function
+- `args` - The struct with all arguments of the traced function.
+  Available in `tracepoint`, `kfunc`, and `uprobe` (with DWARF) probes. Use
+  `args.x` to access argument `x` or `args` to get a record with all arguments.
 - `func` - Name of the traced function
 - `probe` - Full name of the probe
 - `curtask` - Current task struct as a u64
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -932,6 +932,12 @@ For string arguments use the `str()` call to retrieve the value.
 | n/a
 | nth argument passed to the function being traced. These are extracted from the CPU registers. The amount of args passed in registers depends on the CPU architecture. (kprobes, uprobes, usdt).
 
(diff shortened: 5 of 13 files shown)


viktormalik:
ab7681e will go away when #2474 is resolved (do not merge before that).

fbs:
looks good!

viktormalik:
Rebased on master, ready for merge.

viktormalik:
@fbs you didn't put an explicit approval - can we merge this?

@danobi, @ajor want to have a look before that?

Thanks!

ajor (changes requested):
First of all I just want to say that I really like the idea behind this change - the examples look great!

Some of the commits in this PR are simple and unobjectionable and could be merged independently - that would make the core change here a lot less intimidating to go through :)

### Language Design
I know that in the current implementation, `args` is treated as a pointer, but does it need to be?

I don't remember, but I'm assuming I made it like that just because it was easiest - just pass the context pointer directly around and call it "args".

Ideally the bpftrace language won't become full of cruft because of lazy decisions made in the past. It seems to me that a pointer doesn't actually make sense, since we don't allow using it without first dereferencing.

What do you think about allowing use of args without dereferencing? e.g.:
```
kfunc:vfs_read { @a = args }
kfunc:vfs_read { @b = args.file }
```

The old syntax of "args->file" could be kept for backwards compatibility (at least for a while).

### Bugs
I came across a couple of bugs when playing around with this new builtin:
#### Error dereferencing args in tracepoint
```
# ./build/src/bpftrace -e 'tracepoint:syscalls:sys_enter_read { @ = *args }'
Attaching 1 probe...                                                                                      
ERROR: failed to create map: '@': Invalid argument                                                        
```

#### Segfault dereferencing args when function is not defined
```
void foo(int a, int b);
int main() { }
```
```
# ./build/src/bpftrace -e 'uprobe:/tmp/a.out:foo { @ = *args }'
Segmentation fault
```

fbs:
> ### Language Design
> 
> I know that in the current implementation, `args` is treated as a pointer, but does it need to be?
> 
> I don't remember, but I'm assuming I made it like that just because it was easiest - just pass the context pointer directly around and call it "args".
> 
> Ideally the bpftrace language won't become full of cruft because of lazy decisions made in the past. It seems to me that a pointer doesn't actually make sense, since we don't allow using it without first dereferencing.
> 
> What do you think about allowing use of args without dereferencing? e.g.:
> 
> ```
> kfunc:vfs_read { @a = args }
> kfunc:vfs_read { @b = args.file }
> ```
> 
> The old syntax of "args->file" could be kept for backwards compatibility (at least for a while).

Maybe we should just drop the `->` syntax completely? Languages like rust and go are fine without it. I think the type system is good enough to know when we have pointers by now. `$x.a` can just be transformed into `(*$x).a` without the user having to remember?

viktormalik:
> My main concern is unintentionally breaking some part of the C language, like happened when we didn't require the "struct" prefix when casting: #682 A minor concern is confusing C programmers, but I'm sure they'll get used to it.

Well, we'd have to support `->` for backwards compatibility so that shouldn't be an issue. Basically `->` and `.` would have exactly the same effect.

What would change is that users wouldn't be able to rely on the type system to warn them if they use `.` to access pointer variables and vice-versa. But I doubt that anyone relies on that in bpftrace.

viktormalik:
Per agreement with @ajor, I'm splitting this into smaller PRs.

The first three commits are now separated into #2556 and #2557.

ajor:
I'm having second thoughts about the move from "->" to "." now. What would we do about dereferencing pointers when not accessing fields?

C++ and Rust are two languages that mainly use the "." syntax to access struct fields, but that's because they mainly use references instead of pointers. Reading a reference gives the value being pointed to rather than the address it holds (it's implicitly dereferenced). We wouldn't want this behaviour for pointers in bpftrace though, since users will often want and expect an address if they're reading a pointer.

Maybe we should keep pointers and references as separate ideas - we can introduce reference types for higher level language support later, but it feels like this change would be creating something half way between a reference and a pointer. Opinions?

I'm also not sure that it will help with this PR for printing args - if args remains as a pointer type, then printing it will still require `*args`, right? But that dereference was the part I was hoping we could get rid of.

danobi:
I didn't take a very close look at the PR, but the runtime test examples look great.


> I'm having second thoughts about the move from "->" to "." now. What would we do about dereferencing pointers when not accessing fields?

I think simplicity is a good north star for the bpftrace language. While I like the idea of a unifying `.` operator to dereference pointers, I think it is too late to deprecate `->`. We have countless examples and scripts out in the wild. While we could retain support but "soft deprecate it", I suspect that'll just increase confusion. At least for me, I get distrustful of docs/examples that use deprecated features.

> C++ and Rust are two languages that mainly use the "." syntax to access struct fields, but that's because they mainly use references instead of pointers. Reading a reference gives the value being pointed to rather than the address it holds (it's implicitly dereferenced). We wouldn't want this behaviour for pointers in bpftrace though, since users will often want and expect an address if they're reading a pointer.
> 
> Maybe we should keep pointers and references as separate ideas - we can introduce reference types for higher level language support later, but it feels like this change would be creating something half way between a reference and a pointer. Opinions?

I think reference types would require some more discussion, perhaps at an office hours. I have lots of thoughts on that :).

> I'm also not sure that it will help with this PR for printing args - if args remains as a pointer type, then printing it will still require `*args`, right? But that dereference was the part I was hoping we could get rid of.

I don't think we can get rid of it easily. For example rust even supports `*ptr` syntax.

viktormalik:
> > I'm having second thoughts about the move from "->" to "." now. What would we do about dereferencing pointers when not accessing fields?
> 
> I think simplicity is a good north star for the bpftrace language. While I like the idea of a unifying `.` operator to dereference pointers, I think it is too late to deprecate `->`. We have countless examples and scripts out in the wild. While we could retain support but "soft deprecate it", I suspect that'll just increase confusion. At least for me, I get distrustful of docs/examples that use deprecated features.

I agree, deprecating `->` doesn't seem like a good idea. I'd still consider allowing `.` where `->` can be used, though. It may be a bit confusing but we'd basically allow the `.` operator to dereference if applied to a pointer, for the sake of language simplicity.

> I think reference types would require some more discussion, perhaps at an office hours. I have lots of thoughts on that :).

Yes, let's discuss on the next office hours :-)

> > I'm also not sure that it will help with this PR for printing args - if args remains as a pointer type, then printing it will still require `*args`, right? But that dereference was the part I was hoping we could get rid of.
> 
> I don't think we can get rid of it easily. For example rust even supports `*ptr` deref syntax.

But do we need to keep `args` to be a pointer? It's a language builtin so I don't see a problem making it a struct (and perhaps allowing `->` on it, for backwards compatibility).

Anyways, allowing `.` to dereference when necessary would resolve the problem even if `args` remains a pointer.

viktormalik:
Rebased onto master which now includes #2578 which started treating `args` as a struct. This PR now adds capabilities to use `args` in variables and maps (for `kfunc` and `uprobe` probes).

danobi (approved):
nice!

src/ast/irbuilderbpf.cpp
```diff
@@ -1589,6 +1589,37 @@ Value *IRBuilderBPF::CreateRawTracepointArg(Value *ctx,
...
+  }
+  return result;
+}
+
+llvm::Type *IRBuilderBPF::UprobeArgsType(const SizedType &args_type)
+{
+  std::vector<llvm::Type *> arg_types;
+  for (auto &arg : args_type.GetFields())
+    arg_types.push_back(GetType(arg.type));
+  return GetStructType("args_t", arg_types, false);
```
ajor, line 1620: This is causing name collisions - see my top-level comment for details

ajor (changes requested):
I have a couple of questions to help me understand this area better:
**Q1**: Will we able to do something similar to this for tracepoint args in the future or is there some fundamental limitation for them?

**Q2**: Why are uprobe args handled differently from kfunc args? (put on the stack)

I also found a few more problems which we'll need to sort out before merging:

### uprobe arg collisions
```
#include <stdio.h>

void __attribute__((noinline)) bar(int a, int b, int c, int d, int e, int f) {                                                                                        
  printf("hi\n");
}
void __attribute__((noinline)) baz(char g, char h, char i) {                                                                                                          
  printf("hi\n");
}

int main() {                                                                                                                                                          
  bar(11,12,13,14,15,16);
  baz(21, 22, 23);
}
```
`gcc test.c -g -O0`

Works fine with just one uprobe:
```
sudo ./src/bpftrace -e 'uprobe:/tmp/a.out:baz { print(args); }'                                                
Attaching 1 probe...                                                                                                                                                  
{ .g = 21, .h = 22, .i = 23 }                                                                                                                                         
^C
```

Wrong results with multiple uprobes:
```
sudo ./src/bpftrace -e 'uprobe:/tmp/a.out:bar { print(args); } uprobe:/tmp/a.out:baz { print(args); }'         
Attaching 2 probes...                                                                                                                                                 
{ .a = 11, .b = 12, .c = 13, .d = 14, .e = 15, .f = 16 }                                                                                                              
{ .g = 21, .h = 0, .i = 0 }                                                                                                                                           
^C 
```

Crash when uprobe order is swapped:
```
sudo ./src/bpftrace -e 'uprobe:/tmp/a.out:baz { print(args); } uprobe:/tmp/a.out:bar { print(args); }' 
bpftrace: /usr/include/llvm/IR/Instructions.h:922: llvm::Type* llvm::checkGEPType(llvm::Type*): Assertion `Ty && "Invalid GetElementPtrInst indices for type!"' failed.
Aborted
```

### Error message
The error message for uprobes with lots of arguments isn't very nice:
```
sudo ./src/bpftrace -e 'uprobe:/tmp/a.out:foo { print(args); }'
ERROR: Failed to compile: array::at: __n (which is 6) >= _Nm (which is 6)
```

Repro:
```
void __attribute__((noinline)) foo(int a, int b, int c, int d, int e, int f, int g) {                                                                                 
}

int main() {                                                                                                                                                          
  foo(1,2,3,4,5,6,7);                                                                                                                                            
}
```
`gcc test.c -g -O0`

src/ast/irbuilderbpf.cpp
viktormalik: Nice catch, thanks!

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1483,7 +1483,7 @@ void SemanticAnalyser::visit(Map &map)
         map.vargs->at(i) = cast;
         expr = cast;
       }
-      else if (expr->type.IsCtxAccess())
+      else if (expr->type.IsPtrTy() && expr->type.IsCtxAccess())
```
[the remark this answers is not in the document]
viktormalik: Indeed. In fact, only now the original comment underneath matches the condition:
```cpp
        else if (expr->type.IsPtrTy() && expr->type.IsCtxAccess())
        // map functions only accepts a pointer to a element in the stack
```
The check is really about pointers and `args` is not a pointer anymore, so I wouldn't state it explicitly. We can wrap this into some helper like `isNonStackPointer` but I don't think that it would help much.

viktormalik:
Thanks for the review! I added fixup commits to simplify another reviewing and will squash them before merge.

> I have a couple of questions to help me understand this area better: 
>
>**Q1**: Will we able to do something similar to this for tracepoint args in the future or is there some fundamental 
limitation for them?

We should be. Tracepoint args are passed in a similar way to kfuncs, but they have some specialties, like extra fields at the beginning of the struct which we will want to omit or `__data_loc` fields. Once this is merged, I'll open an issue to track support for tracepoints.

> 
> **Q2**: Why are uprobe args handled differently from kfunc args? (put on the stack)

That's b/c for kfuncs, the BPF program gets passed a pointer to a struct containing all the arguments (pretty much exactly what we want). OTOH, uprobe args are passed via registers (the passed context is pointer to `pt_regs`) so we need to build that "struct of args" on stack to have the same codegen representation of `args` for both kfuncs and uprobes (and eventually tracepoints).

src/ast/irbuilderbpf.cpp
```diff
@@ -1589,6 +1589,39 @@ Value *IRBuilderBPF::CreateRawTracepointArg(Value *ctx,
...
+                CreateGEP(args_t,
+                          result,
+                          { getInt64(0), getInt32(arg.type.funcarg_idx) }));
+  }
+  return result;
+}
+
+llvm::Type *IRBuilderBPF::UprobeArgsType(const SizedType &args_type)
+{
+  auto type_name = args_type.GetName().substr(strlen("struct "));
```
ajor, line 1617:
```suggestion
  auto type_name = args_type.GetName();
  type_name.erase(0, strlen("struct "));
```
Removes the prefix in-place to avoid creating a temporary std::string with substr(), eliminating an unnecessary malloc/memcpy/free

ajor (approved):
Nice!

src/ast/irbuilderbpf.cpp
viktormalik: Good idea, fixed.

viktormalik:
Squashed the fixup commits, ready for merge.

PR #2478, opened 2023-01-20
opensnoop.bt: 0 is a valid file descriptor

flx42:
Signed-off-by: Felix Abecassis <fabecassis@nvidia.com>

Consider the following C program:
```c
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

int main(void) {
        int fd;

        close(0);
        fd = open("/proc/self/status", O_RDONLY);
        if (fd < 0) {
                perror("open");
                return 1;
        }
        printf("fd: %d\n", fd);
}
```
The current version of opensnoop.bt prints the following, transforming `0` to `-1`:
```console
288501 a.out              -1   0 /proc/self/status
```

### DIFF
--- a/tools/opensnoop.bt
+++ b/tools/opensnoop.bt
@@ -32,8 +32,8 @@ tracepoint:syscalls:sys_exit_openat
 /@filename[tid]/
 {
 	$ret = args->ret;
-	$fd = $ret > 0 ? $ret : -1;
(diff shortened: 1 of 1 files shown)


viktormalik (approved):
Looks good, thanks!

viktormalik:
CI is failing due to #2474 but this should be a safe change, merging.

Issue #2094, opened 2021-12-21
unknown type name 'xfs_ino_t'


Rtoax:
### What reproduces the bug?

when i execute the following bpftrace scription:

```
#!/usr/bin/env bpftrace

/**
 *	fs/xfs/xfs_linux.h 
 *
 *	See https://github.com/iovisor/bpftrace/issues/2094
 */
//typedef unsigned long long	xfs_ino_t;

BEGIN
{
	printf("Tracing XFS calls... Hit Ctrl-C to end.\n");
}

tracepoint:xfs:xfs_setfilesize
{
	$offset = args->offset;
	$size = args->size;
	printf("%-8d %-16s %-8d %-8d \n", pid, comm, $offset, $size);
}

END
{
	printf("Bye.\n");
}
```

get

```
$ sudo ./xfs_setfilesize-tp.bt
definitions.h:13:3: error: unknown type name 'xfs_ino_t'
ERROR: Include headers with missing type definitions or install BTF information to your system.
```

### bpftrace --info

```
$ sudo bpftrace --info
System
  OS: Linux 5.14.0 #7 SMP Thu Nov 11 15:15:29 CST 2021
  Arch: x86_64

Build
  version: v0.14.0-13-gd431-dirty
  LLVM: 12.0.1
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  bcc bpf_attach_uprobe refcount: yes
  bcc library path resolution: yes
  libbpf: yes
  libbpf btf dump: yes
  libbpf btf dump type decl: yes
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf (depends on Build:libbpf): yes
  map batch (depends on Build:libbpf): yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
[...]

PR #2479, opened 2023-01-20
BTF support for tracepoints defined in modules

viktormalik:
For tracepoints defined in vmlinux, we support usage of BTF for type parsing which can automatically resolve types from tracepoint defs. Thanks to this feature, users do not have to include headers with necessary types manually.
    
This PR enables the same for tracepoints defined in kernel modules. Since parsing of BTF for modules is already supported for kfuncs (#2315) , this change is as simple as adding the subsystem name to the list of modules whose BTF is parsed.

At the moment, this is not supported for tracepoints using wildcarded subsystems. The reason is that tracepoints require dumping of C definitions from BTF which is not possible for multiple modules at once (as it could introduce type redefinitions) so this feature is limited to single-subsystem tracepoints only.

Check #2094 to see where this is useful.

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -251,9 +251,16 @@ std::string BTF::c_def(const std::unordered_set<std::string> &set) const
   if (!has_data())
     return std::string("");
 
-  // Definition dumping from modules would require to resolve type conflicts,
-  // so we dump from vmlinux only for now.
+  // Definition dumping from multiple modules would require to resolve type
+  // conflicts, so we allow dumping from a single module or from vmlinux only.
   std::unordered_set<std::string> to_dump(set);
+  if (btf_objects.size() == 2)
+  {
+    auto *mod_btf = btf_objects[0].btf == vmlinux_btf ? btf_objects[1].btf
+                                                      : btf_objects[0].btf;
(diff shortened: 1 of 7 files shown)


src/btf.cpp
```diff
@@ -250,9 +250,16 @@ std::string BTF::c_def(const std::unordered_set<std::string> &set) const
   if (!has_data())
     return std::string("");
 
-  // Definition dumping from modules would require to resolve type conflicts,
-  // so we dump from vmlinux only for now.
+  // Definition dumping from multiple modules would require to resolve type
+  // conflicts, so we allow dumping from a single module or from vmlinux only.
   std::unordered_set<std::string> to_dump(set);
+  if (btf_objects.size() == 2)
```
danobi, line 256: It looks like if `bpf_objects.size() > 2`, then only vmlinux btf is dumped. And user will get an ambiguous error about incomplete/missing definitions. Do we want to have a more explicit error check?

viktormalik: Good idea. It'll happen only if probes from multiple modules are used at once, but adding the error message should be relatively easy.

dxuuu (approved):

danobi (approved):

PR #2482, opened 2023-01-23
Simplify and fix probe index assignment

### LINKED ISSUE #2447: Can't attach multiple kprobes to multiple function
### What reproduces the bug?
Attach multi-attachment to **TWO** functions.

```bt
kprobe:generic_file_read_iter,
kprobe:filemap_fault
{
    printf("fst attachment on %s\n", func);
}

kprobe:generic_file_read_iter
{
    printf("snd attachment on %s\n", func);
}

kprobe:filemap_fault
{
    printf("snd attachment on %s\n", func);
}
```

The output dumped the following.

```
Attaching 4 probes...
snd attachment on generic_file_read_iter
snd attachment on filemap_fault
...
```

**fst statements are missing.**

#### Also describe what you expect to happen.

The output should dump the following.
```
Attaching 4 probes...
fst attachment on generic_file_read_iter
snd attachment on generic_file_read_iter
fst attachment on filemap_fault
snd attachment on filemap_fault
...
```

Besides, I've tried attach multi-attachment to **ONE** function.

```
kprobe:filemap_fault
{
    printf("fst attachment on %s\n", func);
}

kprobe:filemap_fault
{
    printf("snd attachment on %s\n", func);
}
```

And it works as expected.

```
Attaching 2 probes...
fst attachment on filemap_fault
snd attachment on filemap_fault
```

### bpftrace --info

`bpftrace v0.16.0-97-g50565`

lenticularis39:
Assign a unique index for each probe or attach point in generateProbe (the latter if expansion field is set to true). A new index is generated per each call of generateProbe, unless the probe/attach point already has an index; in the case, the old index is kept and used for the section. (this case happens when a wildcard is expanded).

This makes sure that each section in the generated bytecode is uniquely paired to a pair of AST probe/attach point and the actual probe name, preventing the program being attached to a wrong probe.

Fixes #2447

See the included test for a more minimal example of the issue. 

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -72,7 +72,7 @@ Probe BPFtrace::generateWatchpointSetupProbe(const std::string &func,
   setup_probe.path = ap.target;
   setup_probe.attach_point = func;
   setup_probe.orig_name = get_watchpoint_setup_probe_name(probe.name());
-  setup_probe.index = ap.index(func) > 0 ? ap.index(func) : probe.index();
+  setup_probe.index = ap.index() > 0 ? ap.index() : probe.index();
 
   return setup_probe;
 }
@@ -91,8 +91,8 @@ int BPFtrace::add_probe(ast::Probe &p)
       probe.orig_name = p.name();
       probe.name = p.name();
       probe.loc = 0;
-      probe.index = attach_point->index(probe.name) > 0 ?
-          attach_point->index(probe.name) : p.index();
+      probe.index = attach_point->index() > 0 ? attach_point->index()
+                                              : p.index();
       resources.special_probes.push_back(probe);
       continue;
     }
@@ -266,8 +266,8 @@ int BPFtrace::add_probe(ast::Probe &p)
       probe.address = attach_point->address;
       probe.func_offset = attach_point->func_offset;
       probe.loc = 0;
-      probe.index = attach_point->index(func) > 0 ? attach_point->index(func)
-                                                  : p.index();
+      probe.index = attach_point->index() > 0 ? attach_point->index()
+                                              : p.index();
       probe.len = attach_point->len;
       probe.mode = attach_point->mode;
       probe.async = attach_point->async;
@@ -282,8 +282,8 @@ int BPFtrace::add_probe(ast::Probe &p)
         {
           Probe probe_copy = probe;
           probe_copy.usdt_location_idx = i;
-          probe_copy.index = attach_point->index(func + "_loc" +
-                                                 std::to_string(i));
+          probe_copy.index = attach_point->index() > 0 ? attach_point->index()
+                                                       : p.index();
 
           resources.probes.emplace_back(std::move(probe_copy));
         }
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -30,6 +30,15 @@ EXPECT SUCCESS [0-9][0-9]*
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
+# https://github.com/iovisor/bpftrace/issues/2447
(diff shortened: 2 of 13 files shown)


src/bpftrace.cpp
```diff
@@ -282,8 +282,8 @@ int BPFtrace::add_probe(ast::Probe &p)
         {
           Probe probe_copy = probe;
           probe_copy.usdt_location_idx = i;
-          probe_copy.index = attach_point->index(func + "_loc" +
-                                                 std::to_string(i));
+          probe_copy.index = attach_point->index() > 0 ? attach_point->index()
+                                                       : p.index();
```
viktormalik, line 286: Shouldn't this assign different indices to different locations?

tests/runtime/probe
```diff
@@ -30,6 +30,13 @@ EXPECT SUCCESS [0-9][0-9]*
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
+# https://github.com/iovisor/bpftrace/issues/2447
+NAME kfunc_multiple_attach_point_multiple_functions
+PROG BEGIN { @a = 2; @b = 1; @c = 1 } i:s:1 { exit() } kfunc:vfs_read, kfunc:vfs_open /@a != 0/ { @a -= 1 } kfunc:vfs_read /@b == 1/ { @b = 0 } kfunc:vfs_open /@c == 1/ { @c = 0 }
+EXPECT @a: 0\n@b: 0\n@c: 0
+TIMEOUT 5
```
viktormalik, line 37:
```suggestion
TIMEOUT 5
REQUIRES_FEATURE btf
```

tests/runtime/probe
```diff
@@ -30,6 +30,13 @@ EXPECT SUCCESS [0-9][0-9]*
...
+AFTER ./testprogs/syscall read
```
viktormalik, line 38: We should probably add a comment that this also opens the file which calls `vfs_open`.

src/bpftrace.cpp
lenticularis39: The sections generated for the USDT probe clone are distinguished by name, see this passage in `CodegenLLVM::visit(Probe &probe)`:
```cpp
          current_usdt_location_index_ = 0;
          for (int i = 0; i < attach_point->usdt.num_locations; ++i)
          {
            reset_ids();

            std::string full_func_id = match + "_loc" + std::to_string(i);
            generateProbe(probe, full_func_id, probefull_, func_type, true, i);
            current_usdt_location_index_++;
          }
```
Therefore, they don't need different indices (also they can't have different indices, otherwise they wouldn't be properly matched to the section, which has the attach point index common to the clone and the original program).  Actually, the previous implementation assignes indices unique in the context of a probe name (it was made to fix an issue with multiple identical probes, see #100), hence its index is even less unique globally.

tests/runtime/probe
lenticularis39: I wonder why the other test cases (e.g. the one I copied mine from) don't have this, I thought kfunc usually (always?) require BTF?

viktormalik: It does, if some test is using kfuncs and missing the BTF requirement, it's a mistake.

viktormalik:
Needs rebase to master to fix the CI.

src/bpftrace.cpp
viktormalik: Thanks, that makes sense.

lenticularis39:
I know, I am struggling a little with getting bpftrace to build with LLVM 12 to fix the codegen tests, CMake can be a pain sometimes :D

viktormalik (approved):

PR #2506, opened 2023-02-15
Add CodeQL workflow for GitHub code scanning

lenticularis39:
The PR builds on and closes #2446. CodeQL is configured to avoid scanning submodules (bcc/, libbpf/) and avoid checking the CMake template file tests/runtime/engine/cmake_vars.py. It fixes a few CodeQL errors, mainly in tests.

The remaining errors are because of CodeQL being unable to detect the pattern `switch(...) case ...: ...; case ...: ...; default: LOG(ERROR, ...)`, complaining about no return value or uninitialized variable. These have to be dismissed manually in the PR, or more comfortably in Security tab.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,3 @@
+paths-ignore:
+  - bcc/
+  - libbpf/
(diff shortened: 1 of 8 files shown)


viktormalik (commented):
I dismissed all the non-relevant alerts and left only several notes that can be fixed in future.

Besides that, there are two errors remaining which I believe could be fixed in this PR:
- https://github.com/iovisor/bpftrace/security/code-scanning/523
- https://github.com/iovisor/bpftrace/security/code-scanning/524

viktormalik (approved):

PR #2518, opened 2023-03-07
arm64: define the KASAN_SHADOW_SCALE_SHIFT macro

viktormalik:
arm64 defines the `KASAN_SHADOW_SCALE_SHIFT` macro [from Makefile](https://elixir.bootlin.com/linux/v6.2/source/arch/arm64/Makefile#L147) instead of defining it in a header file as is the case for other architectures. This causes ClangParser not to find it and fail on arm64 systems with `CONFIG_KASAN=y`.

This resolves the issue by defining the macro manually via CFLAGS. The definition is taken from kernel's `arch/arm64/Makefile` and it depends on the running kernel configuration. Therefore, the first commit adds support for parsing the kernel config.

This fixes the `runqlat.bt`, `tcpdrop.bt`, and `undump.bt` tools on `aarch64+debug` kernel which previously failed with:

```
# /usr/share/bpftrace/tools/runqlat.bt 
/lib/modules/5.14.0-283.el9.aarch64+debug/source/arch/arm64/include/asm/memory.h:308:9: error: use of undeclared identifier 'KASAN_SHADOW_SCALE_SHIFT'
/lib/modules/5.14.0-283.el9.aarch64+debug/source/arch/arm64/include/asm/memory.h:308:9: error: use of undeclared identifier 'KASAN_SHADOW_SCALE_SHIFT'
/lib/modules/5.14.0-283.el9.aarch64+debug/source/arch/arm64/include/asm/memory.h:308:9: error: returning 'void' from a function with incompatible result type 'phys_addr_t' (aka 'unsigned long long')
```

##### Checklist


danobi:
It seems like kconfig values are stored in BTF: https://github.com/libbpf/libbpf/blob/b5e9722ec268702b2ec62863222277148e4fe656/src/libbpf.c#L3773-L3798

Is it possible for us to get it from BTF instead?  I think it'll be more likely to be the source of truth, as reading from `/boot/..` can suffer from corner cases when a new kernel is installed but the system hasn't rebooted yet.

src/utils.cpp
```diff
@@ -178,6 +179,54 @@ StdioSilencer::~StdioSilencer()
...
+          option = option.substr(0, option.length() - 1);
+
+        auto split = option.find("=");
+        if (split == std::string::npos)
+          continue;
+
+        config.emplace(option.substr(0, split), option.substr(split + 1));
+      }
+    }
+    gzclose(file);
```
danobi, line 226: Should there be a `break` after this? If both locations are available, probably best not to duplicate effort.

src/utils.cpp
```diff
@@ -178,6 +179,54 @@ StdioSilencer::~StdioSilencer()
...
+  if (path_env)
+    config_locs = { std::string(path_env) };
+  else
+  {
+    struct utsname utsname;
+    if (uname(&utsname) < 0)
+      return;
+    config_locs = {
+      "/boot/config-" + std::string(utsname.release),
+      "/proc/config.gz",
```
danobi, line 198: Maybe order procfs option first. The /boot/.. configs can suffer from skew when installing new kernels.

danobi:
> > It seems like kconfig values are stored in BTF: https://github.com/libbpf/libbpf/blob/b5e9722ec268702b2ec62863222277148e4fe656/src/libbpf.c#L3773-L3798
> > Is it possible for us to get it from BTF instead? I think it'll be more likely to be the source of truth, as reading from `/boot/..` can suffer from corner cases when a new kernel is installed but the system hasn't rebooted yet.
> 
> I may be missing something, but IIUC, this is a libbpf CO-RE feature when it can pass values of kernel configs to a BPF program via external symbols. The values are loaded from kconfig files upon program load in the same way as I used here.

Oh, ha. I did not dig that far. Please ignore me :)

tests/utils.cpp
```diff
@@ -222,6 +222,25 @@ TEST(utils, get_cgroup_path_in_hierarchy)
   }
 }
 
+TEST(utils, parse_kconfig)
+{
+  char path[] = "/tmp/configXXXXXX";
+  int fd = mkstemp(path);
```
danobi, line 228: Might be nice to unlink this after use

danobi (approved):
Sorry about delay. Missed the notification

tests/utils.cpp
viktormalik: Done and resolved the changelog conflict.

PR #2527, opened 2023-03-14
bpftrace: Load vmlinux BTF using btf__load_vmlinux_btf()

chenhengqi:
Using libbpf API btf__load_vmlinux_btf() to load vmlinux BTF. This avoid an issue on old kernel which does NOT support emitting BTF object name and fail on bpf_obj_get_info_by_fd().

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -72,6 +72,16 @@ BTF::~BTF()
 
 void BTF::load_kernel_btfs(const std::set<std::string> &modules)
 {
+  vmlinux_btf = btf__load_vmlinux_btf();
+  if (libbpf_get_error(vmlinux_btf))
+  {
+    if (bt_debug != DebugLevel::kNone)
+      LOG(ERROR) << "BTF: failed to find BTF data for vmlinux, errno " << errno;
+    return;
+  }
+  btf_objects.push_back(
+      BTFObj{ .btf = vmlinux_btf, .id = 0, .name = "vmlinux" });
+
   // Note that we cannot parse BTFs from /sys/kernel/btf/ as we need BTF object
   // IDs, so the only way is to iterate through all loaded BTF objects
   __u32 id = 0;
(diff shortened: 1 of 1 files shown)


src/btf.cpp
```diff
@@ -72,6 +72,16 @@ BTF::~BTF()
 
 void BTF::load_kernel_btfs(const std::set<std::string> &modules)
 {
+  vmlinux_btf = btf__load_vmlinux_btf();
+  if (libbpf_get_error(vmlinux_btf))
+  {
+    if (bt_debug != DebugLevel::kNone)
+      LOG(ERROR) << "BTF: failed to find BTF data for vmlinux";
```
danobi, line 79: Also log the errno from `libbpf_get_error()` here?

chenhengqi: Done.

danobi:
Looks like some runtime tests are failing

chenhengqi:
Will investigate.

src/btf.cpp
```diff
@@ -116,19 +126,10 @@ void BTF::load_kernel_btfs(const std::set<std::string> &modules)
 
     if (mod_name == "vmlinux")
     {
-      btf_objects.push_back(BTFObj{ .btf = btf__load_from_kernel_by_id(id),
-                                    .id = id,
-                                    .name = "vmlinux" });
-      vmlinux_btf = btf_objects.back().btf;
+      btf_objects.front().id = id;
```
viktormalik, line 129: Maybe add a check that the first (and only) entry in `btf_objects` is for vmlinux?

viktormalik (approved):
Looks good to me, just a nit below.

src/btf.cpp
chenhengqi: It seems unnecessary, if we reach `BTF::load_kernel_btfs()`, `vmlinux` should be the first entry. Do you mean adding an assertion ?

chenhengqi: > what if BTF for vmlinux is not the first entry returned by bpf_btf_get_next_id()?

This is OK since we check if the name matched first.

viktormalik: You're right, it's not a problem. Thanks for the PR!

PR #2528, opened 2023-03-14
cmake: Raise max llvm major version to 16

kraj:


viktormalik:
Could we add LLVM 16 to the CI?

kraj:
> Could we add LLVM 16 to the CI?

Once it releases then yes, it is in RC4 right now. Perhaps final release will happen any time

viktormalik:
> > Could we add LLVM 16 to the CI?
> 
> Once it releases then yes, it is in RC4 right now. Perhaps final release will happen any time

Ok, thanks for the info. I'd prefer waiting until that happens. I don't fancy the idea of having support for something that isn't properly tested.

danobi:
I believe llvm 16 is released now

viktormalik:
@kraj please add LLVM 16 to the CI as a part of this PR.

kraj:
> @kraj please add LLVM 16 to the CI as a part of this PR.

Sadly, we are running into https://github.com/llvm/llvm-project/issues/61436 which will be fixed in 16.0.1 point release hopefully, since the fix is already attached to the issue. Until then clang-16 loop wont be able to run.

viktormalik:
> > @kraj please add LLVM 16 to the CI as a part of this PR.
> 
> Sadly, we are running into [llvm/llvm-project#61436](https://github.com/llvm/llvm-project/issues/61436) which will be fixed in 16.0.1 point release hopefully, since the fix is already attached to the issue. Until then clang-16 loop wont be able to run.

Ok, hopefully they release it soon. I suggest to wait until then.

viktormalik:
16.0.1 has been out for two weeks but the issue doesn't seem to be fixed (at least restarting the CI didn't help)

viktormalik:
We should try our best to have this in the 0.18 release (currently planned for May 1st).

danobi:
While upstream LLVM should still be fixed, we can work around the bug by doing something like bcc does:

https://github.com/iovisor/bcc/commit/a1525b76c8297b10ab9c2a227e0c9a5b2e701f44

viktormalik:
> While upstream LLVM should still be fixed, we can work around the bug by doing something like bcc does:
> 
> [iovisor/bcc@a1525b7](https://github.com/iovisor/bcc/commit/a1525b76c8297b10ab9c2a227e0c9a5b2e701f44)

Sounds good. Let me give it a try so that we can do a release with LLVM 16 support.

viktormalik:
Ok, I have it ready and it passes the CI. @kraj do you mind if I push it directly to this PR (which would be your master branch) or should I open a new one?

kraj:
> Ok, I have it ready and it passes the CI. @kraj do you mind if I push it directly to this PR (which would be your master branch) or should I open a new one?

please go ahead.

Issue #341, opened 2019-01-14
raw tracepoint program type


brendangregg:
Raw tracepoint program types were added in Linux 4.17. At some point we should support them, perhaps with a probe type: rawtracepoint, alias "rt".

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c4f6699dfcb8558d138fe838f741b2c10f416cf9

They are faster than normal tracepoint programs, since they skip argument processing and give raw access to the arguments.

brendangregg:
This came up today; we have a use case where we want the pre-translated tracepoint arguments.

qais-yousef:
It'd be nice to have this feature. Some tracepoints are only accessible as raw tracepoints [1]

[1] https://lore.kernel.org/lkml/20210116182133.2286884-1-qais.yousef@arm.com/

gghh:
Hello, I write here to be counted: adding access to raw tracepoints from bpftrace would be very desirable for my usecase.

@brendangregg above mentions the issue of performance (accessing raw tracepoints is faster), but as @qais-yousef says, tracepoints for the scheduler metric "PELT"[1] are available only in raw form.

Part of my work is on improving the frequency scaling governor "schedutil", which was born for ARM but is getting momentum on x86 too. Schedutil feeds on data from PELT and currently the only way to debug PELT is to load an external module to make the tracepoints visible or use BCC (with the addition of a recent kernel patch by Yousef for 5.12).

Brendan's latest book popularized bpftrace and this tool is getting a lot of traction; in some cases it is way more convenient to use than building and loading a module or writing a BCC script. It surely is for asking users to run a quick one-liner and report the result.

[1] Per-entity load tracking https://lwn.net/Articles/531853/

fbs:
From a users perspective, how do we use this? I see some options that could work, without extending the list of builtins.

1:
```
rawtracepoint:something {
   @ = args[0]
}
```

2:
```
rawtracepoint:something {
   @ = args->raw[0]
}
```

3:
```
rawtracepoint:something {
   @ = arg0
}
```

1 & 2 stay closer to the tp syntax, using the `args` builtin. 2 would be easier to implement but 1 might be nicer to have. 3 makes it more like a kprobe, not sure if there is a limit to the amount of args a tracepoint can have but we can probably support `arg133030` if we want to.

@brendangregg @gghh

qais-yousef:
[...]

PR #2542, opened 2023-03-27
Support raw tracepoint program type

zf1575192187:
Complete bpftrace support for raw tracepoint.

Most of the logic follows kprobe or tracepoint. It's just that the parameters of bpf prog are specially handled, and it only needs to be simply offset according to the X of argX.

https://github.com/iovisor/bpftrace/issues/341

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -46,17 +46,19 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [4. `uprobe`/`uretprobe`: Dynamic Tracing, User-Level Arguments](#4-uprobeuretprobe-dynamic-tracing-user-level-arguments)
     - [5. `tracepoint`: Static Tracing, Kernel-Level](#5-tracepoint-static-tracing-kernel-level)
     - [6. `tracepoint`: Static Tracing, Kernel-Level Arguments](#6-tracepoint-static-tracing-kernel-level-arguments)
-    - [7. `usdt`: Static Tracing, User-Level](#7-usdt-static-tracing-user-level)
-    - [8. `usdt`: Static Tracing, User-Level Arguments](#8-usdt-static-tracing-user-level-arguments)
-    - [9. `profile`: Timed Sampling Events](#9-profile-timed-sampling-events)
-    - [10. `interval`: Timed Output](#10-interval-timed-output)
-    - [11. `software`: Pre-defined Software Events](#11-software-pre-defined-software-events)
-    - [12. `hardware`: Pre-defined Hardware Events](#12-hardware-pre-defined-hardware-events)
-    - [13. `BEGIN`/`END`: Built-in events](#13-beginend-built-in-events)
-    - [14. `watchpoint`/`asyncwatchpoint`: Memory watchpoints](#14-watchpointasyncwatchpoint-memory-watchpoints)
-    - [15. `kfunc`/`kretfunc`: Kernel Functions Tracing](#15-kfunckretfunc-kernel-functions-tracing)
-    - [16. `kfunc`/`kretfunc`: Kernel Functions Tracing Arguments](#16-kfunckretfunc-kernel-functions-tracing-arguments)
-    - [17. `iter`: Iterators Tracing ](#17-iter-iterators-tracing)
+    - [7. `rawtracepoint`: Static Tracing, Kernel-Level](#7-rawtracepoint-static-tracing-kernel-level)
+    - [8. `rawtracepoint`: Static Tracing, Kernel-Level Arguments](#8-rawtracepoint-static-tracing-kernel-level-arguments)
+    - [9. `usdt`: Static Tracing, User-Level](#9-usdt-static-tracing-user-level)
+    - [10. `usdt`: Static Tracing, User-Level Arguments](#10-usdt-static-tracing-user-level-arguments)
+    - [11. `profile`: Timed Sampling Events](#11-profile-timed-sampling-events)
+    - [12. `interval`: Timed Output](#12-interval-timed-output)
+    - [13. `software`: Pre-defined Software Events](#13-software-pre-defined-software-events)
+    - [14. `hardware`: Pre-defined Hardware Events](#14-hardware-pre-defined-hardware-events)
+    - [15. `BEGIN`/`END`: Built-in events](#15-beginend-built-in-events)
+    - [16. `watchpoint`/`asyncwatchpoint`: Memory watchpoints](#16-watchpointasyncwatchpoint-memory-watchpoints)
+    - [17. `kfunc`/`kretfunc`: Kernel Functions Tracing](#17-kfunckretfunc-kernel-functions-tracing)
+    - [18. `kfunc`/`kretfunc`: Kernel Functions Tracing Arguments](#18-kfunckretfunc-kernel-functions-tracing-arguments)
+    - [19. `iter`: Iterators Tracing ](#19-iter-iterators-tracing)
 - [Variables](#variables)
     - [1. Builtins](#1-builtins)
     - [2. `@`, `$`: Basic Variables](#2---basic-variables)
@@ -1284,7 +1286,45 @@ listed first, the members are specific to the tracepoint.
 Examples in situ:
 [search /tools](https://github.com/iovisor/bpftrace/search?q=tracepoint%3A+path%3Atools&type=Code)
 
-## 7. `usdt`: Static Tracing, User-Level
+## 7. `rawtracepoint`: Static Tracing, Kernel-Level
+
+The hook point triggered by `tracepoint` and `rawtracepoint` is the same. The difference is that raw tracepoints are faster since no argument processing is done (see below).
+
+Syntax: `rawtracepoint:name`
+
+These use tracepoints (a Linux kernel capability).
+
+```
+# bpftrace -e 'rawtracepoint:block_rq_insert { printf("block I/O created by %d\n", tid); }'
+Attaching 1 probe...
+block I/O created by 189126
+[...]
+```
+
+## 8. `rawtracepoint`: Static Tracing, Kernel-Level Arguments
+
+`tracepoint` and `rawtracepoint` are nearly identical in terms of functionality. The only difference is in the program context. `rawtracepoint` offers raw arguments to the tracepoint while `tracepoint` applies further processing to the raw arguments. The additional processing is defined inside the kernel.
+
+Example:
+
+```
+# bpftrace -e 'rawtracepoint:block_rq_insert { printf("%llx %llx\n", arg0, arg1); }'
+Attaching 1 probe...
+ffff88810977d6f8 ffff8881097e8e80
+[...]
+```
+
+The available arguments for each tracepoint can be found in the relative path of the kernel source code include/trace/events/. For example:
+```
+include/trace/events/block.h
+DEFINE_EVENT(block_rq, block_rq_insert,
+	TP_PROTO(struct request_queue *q, struct request *rq),
+	TP_ARGS(q, rq)
+);
+```
+The individual args are accessed by their order via `argN` builtins. Each arg is a 64-bit integer.
+
+## 9. `usdt`: Static Tracing, User-Level
 
 Syntax:
 
@@ -1356,7 +1396,7 @@ matches based on file path. This means that if bpftrace runs from the root host,
 as expected if there are processes `execve`d from private mount namespaces or bind mounted directories.
 One workaround is to run bpftrace inside the appropriate namespaces (ie the container).
 
-## 8. `usdt`: Static Tracing, User-Level Arguments
+## 10. `usdt`: Static Tracing, User-Level Arguments
 
 Examples:
 
@@ -1379,7 +1419,7 @@ my string: 6
 ^C
 ```
 
-## 9. `profile`: Timed Sampling Events
+## 11. `profile`: Timed Sampling Events
 
 Syntax:
 
@@ -1403,7 +1443,7 @@ Attaching 1 probe...
 @[0]: 579
 ```
 
-## 10. `interval`: Timed Output
+## 12. `interval`: Timed Output
 
 Syntax:
 
@@ -1437,7 +1477,7 @@ This prints the rate of syscalls per second.
 Examples in situ:
 [search /tools](https://github.com/iovisor/bpftrace/search?q=interval+extension%3Abt+path%3Atools&type=Code)
 
-## 11. `software`: Pre-defined Software Events
+## 13. `software`: Pre-defined Software Events
 
 Syntax:
 
@@ -1490,7 +1530,7 @@ Attaching 1 probe...
 This roughly counts who is causing page faults, by sampling the process name for every one in one hundred
 faults.
 
-## 12. `hardware`: Pre-defined Hardware Events
+## 14. `hardware`: Pre-defined Hardware Events
 
 Syntax:
 
@@ -1526,7 +1566,7 @@ bpftrace -e 'hardware:cache-misses:1000000 { @[pid] = count(); }'
 
 That would fire once for every 1000000 cache misses. This usually indicates the last level cache (LLC).
 
-## 13. `BEGIN`/`END`: Built-in events
+## 15. `BEGIN`/`END`: Built-in events
 
 Syntax:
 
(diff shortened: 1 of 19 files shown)


src/ast/irbuilderbpf.h
```diff
@@ -188,6 +188,7 @@ class IRBuilderBPF : public IRBuilder<>
   Value *CreateRegisterRead(Value *ctx, const std::string &builtin);
   Value *CreateRegisterRead(Value *ctx, int offset, const std::string &name);
   Value      *CreatKFuncArg(Value *ctx, SizedType& type, std::string& name);
+  Value *CreatRawTracepointArg(Value *ctx, const std::string &builtin);
```
viktormalik, line 191:
```suggestion
  Value *CreateRawTracepointArg(Value *ctx, const std::string &builtin);
```

src/ast/irbuilderbpf.cpp
```diff
@@ -1474,6 +1474,16 @@ Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
...
 }
 
+Value *IRBuilderBPF::CreatRawTracepointArg(Value *ctx, const std::string &builtin)
+{
+  int offset;
+
+  // argX
+  offset = atoi(builtin.substr(3).c_str());
+
+  return CreateRegisterRead(ctx, offset, builtin);
```
viktormalik, line 1484: TBH, I don't like that this calls `CreateRegisterRead` when there are in fact no registers. In addition, `CreateRegisterRead` assumes read from `pt_regs` whose elements have the size of a pointer while raw tracepoint args are always u64. We should probably factor out a piece of `CreateRegisterRead` and use that from here.

danobi:
Adding raw tracepoint support seems good to me. Please also update the appropriate docs (reference_guide and man page)

src/ast/irbuilderbpf.h
zf1575192187: Done

src/ast/irbuilderbpf.cpp
zf1575192187: Yes, it has been re-implemented.

zf1575192187:
Will do, added.

docs/reference_guide.md
```diff
@@ -1829,6 +1831,42 @@ NetworkManager:1155 /var/lib/sss/mc/passwd (deleted)
 
 ```
 
+## 18. `rawtracepoint`: Static Tracing, Kernel-Level
+
+Syntax: `rawtracepoint:name` or `rt:name`
+
+These use tracepoints (a Linux kernel capability).
```
viktormalik, line 1838: This should mention the differences from `tracepoint`. Also, since these two are quite related, it'd make sense to put them next to each other in the reference guide.

docs/reference_guide.md
```diff
@@ -1829,6 +1831,42 @@ NetworkManager:1155 /var/lib/sss/mc/passwd (deleted)
...
+```
+include/trace/events/block.h
+
+DEFINE_EVENT(block_rq, block_rq_insert,
+
+	TP_PROTO(struct request_queue *q, struct request *rq),
+
+	TP_ARGS(q, rq)
+);
+```
```
viktormalik, line 1868: Mention here that the individual args are accessed by their order via `argN` builtins and that each arg is 64-bit.

docs/reference_guide.md
```diff
@@ -1829,6 +1831,42 @@ NetworkManager:1155 /var/lib/sss/mc/passwd (deleted)
...
+Example:
+
+```
+# bpftrace -e 'rawtracepoint:block_rq_insert { printf("%llx %llx\n", arg0, arg1); }'
+Attaching 1 probe...
+ffff88810977d6f8 ffff8881097e8e80
+[...]
+```
+
+The available members for each tracepoint can be can be found in the relative path of the kernel source code include/trace/events/. For example:
```
viktormalik, line 1858:
```suggestion
The available arguments for each tracepoint can be can be found in the relative path of the kernel source code include/trace/events/. For example:
```
These are not members as we are not accessing them via `args->...`.

src/ast/irbuilderbpf.cpp
```diff
@@ -1474,6 +1474,24 @@ Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
...
+  llvm::Type *type = getInt64Ty();
+
+  ctx = CreatePointerCast(ctx, type->getPointerTo());
+  Value *expr = CreateLoad(type,
+                             CreateGEP(type, ctx, getInt64(offset)),
+                             builtin);
+
+  // LLVM 7.0 <= does not have CreateLoad(*Ty, *Ptr, isVolatile, Name),
+  // so call setVolatile() manually
+  dyn_cast<LoadInst>(expr)->setVolatile(true);
```
viktormalik, line 1490: We don't support LLVM below 10 anymore, so it's fine to drop this in favor of `CreateLoad(*Ty, *Ptr, isVolatile, Name)` call.

viktormalik (commented):
More tests would be useful. Some tips:
- semantic analyser (various valid and invalid raw tracepoint invocations)
- runtime (actually attaching to a raw tracepoint)
- codegen (accessing raw tracepoint arguments)
- all of the above in combination with wildcards (if we want to support them)

docs/reference_guide.md
```diff
@@ -1241,7 +1248,23 @@ block I/O created by 28941
...
+```
+# bpftrace -e 'rawtracepoint:block_rq_insert { printf("block I/O created by %d\n", tid); }'
+Attaching 1 probe...
+block I/O created by 189126
+[...]
+```
+
+## 6. `tracepoint`/`rawtracepoint`: Static Tracing, Kernel-Level Arguments
+
+The bpf ctx parameters of tracepoint and rawtracepoint are different, rawtracepoint is lower level, and tracepoint is processed by the kernel on the basis of the parameters of rawtracepoint.
```
danobi, line 1260: I understand what you mean here but probably only b/c I already know what raw tracepoints are. This might be too ambiguous for people new to it. Perhaps something like:

```suggestion
`tracepoint` and `rawtracepoint` are nearly identical in terms of functionality. The only difference is in the program context. `rawtracepoint` offers raw arguments to the tracepoint while `tracepoint` applies further processing to the raw arguments. The additional processing is defined inside the kernel.
```

docs/reference_guide.md
```diff
@@ -1284,6 +1307,24 @@ listed first, the members are specific to the tracepoint.
...
+
+The available arguments for each tracepoint can be found in the relative path of the kernel source code include/trace/events/. For example:
+```
+include/trace/events/block.h
+DEFINE_EVENT(block_rq, block_rq_insert,
+	TP_PROTO(struct request_queue *q, struct request *rq),
+	TP_ARGS(q, rq)
+);
+```
+The individual args are accessed by their order via argN builtins and that each arg is 64-bit.
```
danobi, line 1326:
```suggestion
The individual args are accessed by their order via argN builtins. Each arg is a 64-bit integer.
```

danobi (commented):
Please also add docs to `man/adoc/bpftrace.adoc`. I understand it's mostly duplicated from reference_guide.md. It's on our todo list to consolidate it. But for now please add the extra docs so it's easier to consolidate later.

Please also add a codegen test. 1 simple one should be sufficient. They're useful to get a diff of codegen if/when someone changes it. See: https://github.com/iovisor/bpftrace/tree/master/tests#codegen-tests

docs/reference_guide.md
viktormalik: I'd still consider separating raw tracepoints section from tracepoints (just keep them next to each other). They are quite different from the usage point of view (raw tracepoints don't have categories, arguments are accessed differently). But that's just an idea, not insisting on it.

viktormalik: I'd still consider separating raw tracepoints section from tracepoints (just keep them next to each other). They are quite different from the usage point of view (raw tracepoints don't have categories, arguments are accessed differently). But that's just an idea, not insisting on it.

viktormalik: But I really like your rewording here, no problem using that.

danobi: I don't have a strong opinion on separate or together. Separate is probably cleaner from a docs perspective.

zf1575192187: Will do. I'll revise it next week and add test case, glad to get your suggestions.

docs/reference_guide.md
```diff
@@ -1284,7 +1286,45 @@ listed first, the members are specific to the tracepoint.
 Examples in situ:
 [search /tools](https://github.com/iovisor/bpftrace/search?q=tracepoint%3A+path%3Atools&type=Code)
 
-## 7. `usdt`: Static Tracing, User-Level
+## 7. `rawtracepoint`: Static Tracing, Kernel-Level
+
+The hook point triggered by `tracepoint` and `rawtracepoint` is the same.
```
viktormalik, line 1291: Maybe also add:
```suggestion
The hook point triggered by `tracepoint` and `rawtracepoint` is the same. The difference is that raw tracepoints are faster since no argument processing is done (see below).
```

docs/reference_guide.md
```diff
@@ -1284,7 +1286,45 @@ listed first, the members are specific to the tracepoint.
...
+
+The available arguments for each tracepoint can be found in the relative path of the kernel source code include/trace/events/. For example:
+```
+include/trace/events/block.h
+DEFINE_EVENT(block_rq, block_rq_insert,
+	TP_PROTO(struct request_queue *q, struct request *rq),
+	TP_ARGS(q, rq)
+);
+```
+The individual args are accessed by their order via argN builtins. Each arg is a 64-bit integer.
```
viktormalik, line 1325:
```suggestion
The individual args are accessed by their order via `argN` builtins. Each arg is a 64-bit integer.
```

man/adoc/bpftrace.adoc
```diff
@@ -2397,6 +2397,29 @@ snmpd /proc/net/dev
 .Additional information
 * https://www.kernel.org/doc/html/latest/trace/tracepoints.html
 
+.variants
```
viktormalik, line 2400:
```suggestion
[#probes-rawtracepoint]
=== rawtracepoint

.variants
```

man/adoc/bpftrace.adoc
```diff
@@ -2397,6 +2397,29 @@ snmpd /proc/net/dev
...
+* `rawtracepoint:event`
+
+.shortnames
+* `rt`
+
+The hook point triggered by `tracepoint` and `rawtracepoint` is the same.
+`tracepoint` and `rawtracepoint` are nearly identical in terms of functionality. The only difference is in the program context. `rawtracepoint` offers raw arguments to the tracepoint while `tracepoint` applies further processing to the raw arguments. The additional processing is defined inside the kernel.
+
+Tracepoint arguments are available in the `argN` which can be found in the relative path of the kernel source code `include/trace/events/` for more details.
```
viktormalik, line 2409:
```suggestion
Tracepoint arguments are available via the `argN` builtins. The available arguments can be found in the relative path of the kernel source code `include/trace/events/`.
```

tests/semantic_analyser.cpp
```diff
@@ -1460,6 +1460,11 @@ TEST(semantic_analyser, tracepoint)
   test("tracepoint:category:event { 1 }", 0);
 }
 
+TEST(semantic_analyser, rawtracepoint)
+{
+  test("rawtracepoint:event { 1 }", 0);
```
viktormalik, line 1465: Can we add at least one negative test-case?
```suggestion
  test("rawtracepoint:event { 1 }", 0);
  test("rawtracepoint:category:event { 1 }", 1);
```

src/ast/attachpoint_parser.cpp
```diff
@@ -778,5 +781,24 @@ AttachPointParser::State AttachPointParser::iter_parser()
...
+      return SKIP;
+
+    errs_ << ap_->provider << " probe type requires 1 argument" << std::endl;
+    return INVALID;
+  }
+
+  ap_->func = parts_[1];
+
+  if (ap_->func.find('*') != std::string::npos)
+    ap_->need_expansion = true;
```
viktormalik, line 798: If we want to support wildcards, we should probably add tests for them. Perhaps it would be easier to omit wildcard support for now and add it as a follow-up PR?

src/probe_matcher.cpp
```diff
@@ -126,7 +126,7 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
       break;
     }
     case ProbeType::tracepoint:
-    {
+    case ProbeType::rawtracepoint: {
```
viktormalik, line 129: I don't think that this is correct. `/sys/kernel/tracing/available_events` contains entries in the form `category:event` while raw tracepoints only use the `event` part.

Listing all raw tracepoints shows an empty list:
```
# bpftrace -l 'rawtracepoint:*`
#
```
As I said above, I suggest removing wildcard support for now and adding it in a follow-up PR.

viktormalik (commented):
Couple of comments below plus two remarks:
- I'm still missing at least one runtime test. Perhaps you could add something simple into `tests/runtime/probe`?
- Attaching to a non-existing raw tracepoint gives a generic error:
```
# bpftrace -e 'rawtracepoint:something { printf("hit\n"); }'
Attaching 1 probe...
ERROR: Error attaching probe: rawtracepoint:something
```
Would it be possible to improve the error in such case?

src/ast/attachpoint_parser.cpp
```diff
@@ -778,5 +781,21 @@ AttachPointParser::State AttachPointParser::iter_parser()
...
+  if (parts_.size() != 2)
+  {
+    if (ap_->ignore_invalid)
+      return SKIP;
+
+    errs_ << ap_->provider << " probe type requires 1 argument" << std::endl;
+    return INVALID;
+  }
+
+  ap_->func = parts_[1];
```
viktormalik, line 795: We should explicitly disable wildcards for now:
```suggestion
  ap_->func = parts_[1];
  
  if (has_wildcard(ap_->func))
  {
    if (ap_->ignore_invalid)
      return SKIP;

    errs_ << ap_->provider << " probe type does not support wildcards"
          << std::endl;
    return INVALID;
  }
```

src/attached_probe.cpp
```diff
@@ -147,6 +148,22 @@ int AttachedProbe::detach_iter(void)
...
 }
 
+void AttachedProbe::attach_raw_tracepoint(void)
+{
+  if (progfd_ < 0)
+    // Errors for raw_tracepoint are handled in load_prog, ignore here
+    return;
+  tracing_fd_ = bpf_raw_tracepoint_open(probe_.attach_point.c_str(), progfd_);
+  if (tracing_fd_ < 0)
+    throw std::runtime_error("Error attaching probe: " + probe_.name);
```
viktormalik, line 158: We could detect tracepoint existence here:
```suggestion
  if (tracing_fd_ < 0)
  {
    if (tracing_fd_ == -ENOENT)
      throw std::runtime_error("Probe does not exist: " + probe_.name);
    else
      throw std::runtime_error("Error attaching probe: " + probe_.name);
  }
```

viktormalik:
Semantic analyser tests are segfaulting in the CI. Can you debug please?

zf1575192187:
Will do.

viktormalik (approved):

tests/runtime/probe
```diff
@@ -248,6 +248,24 @@ PROG tracepoint:irq:irq_handler_entry { print(str(args->name)); exit(); }
...
+
+NAME rawtracepoint_short_name
+PROG rt:sys_exit { printf("SUCCESS %d\n", gid); exit(); }
+EXPECT SUCCESS [0-9][0-9]*
+AFTER ./testprogs/syscall nanosleep 1e8
+TIMEOUT 5
+
+NAME rawtracepoint_missing
+PROG rawtracepoint:nonsense { printf("hit"); exit(); }
+EXPECT ERROR: Error attaching probe: rawtracepoint:nonsense
```
viktormalik, line 265: This needs to be fixed to the new error message.

zf1575192187:
It has been revised, please take the time to help review it, thanks.

viktormalik:
LGTM, thanks!

danobi (approved):

zf1575192187:
@danobi @viktormalik 
The rawtracepoint implemented by this PR does not yet support -l viewing, and there are wildcards. After this merge, I will support it in the next PR.

viktormalik:
Merged. @zf1575192187 if you want to track the following steps (listing and wildcards), feel free to open an issue(s). 
Thanks a lot for your work!

PR #2555, opened 2023-04-06
Include DWARF support in static build.

haosdent:
##### Changes

* Disable test binary building in static build
* Include `libdw.a` and `libebl.a` libraries to make bpftrace static build support DWARF.
* Remove unnecessary `-it` flag in `build-static.sh` to avoid build error in CI. (Because almost CI don't accept TTY)

##### Checklist


docker/build.sh
```diff
@@ -10,6 +10,7 @@ EMBED_USE_LLVM=${EMBED_USE_LLVM:-OFF}
 EMBED_BUILD_LLVM=${EMBED_BUILD_LLVM:-OFF}
 ALLOW_UNSAFE_PROBE=${ALLOW_UNSAFE_PROBE:-OFF}
 DEPS_ONLY=${DEPS_ONLY:-OFF}
+BUILD_TESTING=${BUILD_TESTING:-OFF}
```
danobi, line 13: I think this needs to default to `ON`. Otherwise it disables all the tests in the CI.

```suggestion
BUILD_TESTING=${BUILD_TESTING:-ON}
```

danobi (changes requested):

docker/build.sh
haosdent: I see, let me enable back

PR #2559, opened 2023-04-08
Add trailer to truncated strings

danobi:
This PR adds support for string truncation trailers. See runtime tests and included
documentation for details.

I tried two other implementations before landing on this one:

1. Detect truncation in the prog and add the trailer there
2. Modify the layout of `Type::string` to include the attempted read length as part of an async event

(1) was fairly complex and probably isn't as flexible as the current approach.
The upside is it would be easier to support truncation trailers for `str($p, $n)`
calls, where `$n` causes truncation.

(2) was an attempt to simplify and address the `str($p, $n)` problem, but it
also suffered from complexity as other helpers (like `strncmp()` and
`strcontains()`) depended on the c-string layout. I suppose you could add
extra metadata _after_ the NUL, but that seemed like a bit of a hack.

The final (and current) approach was the simplest and easiest to maintain
IMO. It does suffer from the downside that `str($p, $n)` truncation is not
properly handled, depending on how you look at it.

My take is that if the user manually requests truncation, then they probably
don't need or want the trailer. It's usually when users are expecting the full
string (via `str($p)`) that the trailer can be a useful visual identifier.

##### Checklist


viktormalik:
> My take is that if the user manually requests truncation, then they probably don't need or want the trailer. It's usually when users are expecting the full string (via `str($p)`) that the trailer can be a useful visual identifier.

I have the same view about this.

I don't expect users to hit this often so this simple approach seems like a nice and easy-to-maintain solution.

viktormalik (approved):

viktormalik:
@ajor you might want to have a look since you're the one who suggested this solution.

danobi:
Going to go ahead and merge. Just lemme know if you want changes after the fact

PR #2563, opened 2023-04-12
Add pahole dep to ubuntu build requirements.

rcorre:
Building `main` on Ubuntu 22.04 ARM gave me:

```
CMake Error at tests/CMakeLists.txt:50 (find_program):
  Could not find PAHOLE using the following names: pahole
```

`apt install pahole` resolved it.

##### Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -295,7 +295,8 @@ sudo apt-get install -y \
   libpcap-dev \
   libgtest-dev \
   libgmock-dev \
-  asciidoctor
+  asciidoctor \
+  pahole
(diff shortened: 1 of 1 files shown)


danobi (commented):
Hi, thanks for catching this! Looks like this went into the fedora section rather than ubuntu. Mind adding it to ubuntu as well?

Unrelated to your PR but we should probably drop all the developer packages (like pahole) from INSTALL.md. We can document those deps in docs/developers.md.

viktormalik (approved):

viktormalik:
Thanks!

PR #2565, opened 2023-04-14
Add offsetof() builtin

Rtoax:
Add the offsetof() function, the syntax follows the kernel offsetof() macro definition, and currently only supports single-member format parameters such as offsetof(struct task_struct, comm). The function returns a long integer with the output format '%ld'.

This is the second PR for offsetof(), the first version was not merged, and the discussion of offsetof() can be found in the first PR [0].

Basic use:

```
    #include <linux/sched.h>

    struct Foo {
        int a;
        long b;
        char c;
        struct {
            int d;
        };
        struct {
            int a;
        } e;
        int offsetof;
    }
    struct offsetof {
        int offsetof;
        int bswap;
	int comm;
	int kstack;
	int ustack;
	int pid;
	int ctx;
	int arg;
    }

    BEGIN {
        printf("%ld\n", offsetof(struct Foo, c));
        printf("%ld\n", offsetof(struct Foo, d));
        printf("%ld\n", offsetof(struct Foo, offsetof));

        printf("%ld\n", offsetof(struct offsetof, offsetof));
        printf("%ld\n", offsetof(struct offsetof, bswap));

        printf("%ld\n", offsetof(struct task_struct, comm));
        printf("%ld\n", offsetof(*curtask, comm));

        exit();
    }
```

  Get offset:
```
    16
    20
    28
    0
    4
    3240
    3240
```

Wrong use:

```
    # ERROR: offsetof() requires 2 arguments (1 provided)
    printf("%ld\n", offsetof(struct Foo));

    # ERROR: The first argument of offsetof() only supports record (pointer provided)
    printf("%ld\n", offsetof(curtask, d));

    # ERROR: struct 'struct Foo' has no field 'xyz'
    printf("%ld\n", offsetof(struct Foo, xyz));

    # ERROR: The first argument of offsetof() only supports record (none provided)
    printf("%ld\n", offsetof(c, struct Foo));

    # ERROR: offsetof() unsupport subfield format.
    printf("%ld\n", offsetof(struct Foo, e.a));

    # ERROR: offsetof() unsupport subfield format.
    printf("%ld\n", offsetof(struct Foo, e->a));

    # ERROR: struct 'struct Foo' has no field '@e'
    printf("%ld\n", offsetof(struct Foo, @e));

    # ERROR: struct 'struct Foo' has no field '$e'
    printf("%ld\n", offsetof(struct Foo, $e));

    # ERROR: offsetof() unsupport expression format.
    printf("%ld\n", offsetof(struct Foo, e+a));
```

[0] https://github.com/iovisor/bpftrace/pull/2216

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -2240,6 +2240,7 @@ Tracing block I/O sizes > 0 bytes
 - `kptr(void *p)` - Annotate as kernelspace pointer
 - `macaddr(char[6] addr)` - Convert MAC address data
 - `bswap(uint[8|16|32|64] n)` - Reverse byte order
+- `offsetof(struct, element)` - Offset of element in structure
 
 Some of these are asynchronous: the kernel queues the event, but some time later (milliseconds) it is
 processed in user-space. The asynchronous actions are: `printf()`, `time()`, and `join()`. Both `ksym()`
@@ -3260,6 +3261,31 @@ Example:
 BEGIN { print(strerror(EPERM)); }'
 Attaching 1 probe...
 Operation not permitted
+```
+
+# 35. `offsetof`: Offset of element in structure
+
+Syntax: `offsetof(struct, element)`
+
+Get the offset of the element in the struct.
+
+Examples:
+
+```
+#!/usr/bin/env bpftrace
+
+BEGIN
+{
+	printf("Offset of flags: %ld\n", offsetof(struct task_struct, flags));
+	printf("Offset of comm: %ld\n", offsetof(struct task_struct, comm));
+	exit();
+}
+```
+
+```
+Attaching 1 probe...
+Offset of flags: 44
+Offset of comm: 3216
 ```
 
 # Map Functions
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1147,6 +1147,14 @@ void CodegenLLVM::visit(Call &call)
   {
     expr_ = b_.getInt64(call.vargs->at(0)->type.GetSize());
   }
+  else if (call.func == "offsetof")
+  {
+    auto identifier = dynamic_cast<Identifier *>(call.vargs->at(1));
+    assert(identifier && "offsetof() 2nd argument not an identifier");
+
+    auto &field = call.vargs->at(0)->type.GetField(identifier->ident);
+    expr_ = b_.getInt64(field.offset);
+  }
   else if (call.func == "strerror")
   {
     auto scoped_del = accept(call.vargs->front());
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -126,6 +126,11 @@ void SemanticAnalyser::visit(Identifier &identifier)
     identifier.type = CreateInt(
         std::get<0>(getIntcasts().at(identifier.ident)));
   }
+  else if (func_ == "offsetof")
+  {
+    // Handling the second arg of offsetof(): (field name)
+    identifier.type = CreateNone();
+  }
   else {
     identifier.type = CreateNone();
     LOG(ERROR, identifier.loc, err_)
@@ -514,7 +519,39 @@ void SemanticAnalyser::visit(Call &call)
       auto &expr = (*call.vargs)[i];
       func_arg_idx_ = i;
 
-      expr->accept(*this);
+      // Handling offsetof() 2nd argument
+      // Handle the second parameter offsetof(record, field), this parameter
+      // is treated as an Identifier, of course, it may be call or builtin, we
+      // convert them all to Identifiers. For example:
+      //
+      //  struct foo {
+      //    int offsetof;   > offsetof(struct foo, offsetof)
+      //    int kstack;     > offsetof(struct foo, kstack)
+      //    int ctx;        > offsetof(struct foo, ctx)
+      //    int comm;       > offsetof(struct foo, comm)
+      //  }
+      if (func_ == "offsetof" && func_arg_idx_ == 1)
+      {
+        auto _builtin = dynamic_cast<Builtin *>(call.vargs->at(1));
+        auto _call = dynamic_cast<Call *>(call.vargs->at(1));
+
+        if (_builtin)
+        {
+          delete expr;
+          expr = new Identifier(_builtin->ident, call.loc);
+        }
+        else if (_call)
+        {
+          delete expr;
+          expr = new Identifier(_call->func, call.loc);
+        }
+
+        expr->accept(*this);
+      }
+      else
+      {
+        expr->accept(*this);
+      }
     }
   }
 
@@ -1177,6 +1214,33 @@ void SemanticAnalyser::visit(Call &call)
     check_nargs(call, 1);
 
     call.type = CreateUInt64();
+  }
+  else if (call.func == "offsetof")
+  {
+    if (check_nargs(call, 2))
+    {
+      // Handle Identifier only, see earlier in this function where we accept()
+      // call arguments.
+      auto element = dynamic_cast<Identifier *>(call.vargs->at(1));
+      assert(element && "offsetof() 2nd argument not an identifier");
+
+      auto &arg = *call.vargs->at(0);
+
+      if (arg.type.type != Type::record)
+      {
+        LOG(ERROR, call.loc, err_)
+            << "The first argument of " << call.func << "() only supports "
+            << Type::record << " (" << arg.type.type << " provided)";
+      }
+      else if (!arg.type.HasField(element->ident))
+      {
+        LOG(ERROR, call.loc, err_)
+            << "struct \'" << arg.type.GetName() << "\' has no field \'"
+            << element->ident << "\'";
+      }
+
+      call.type = CreateUInt64();
+    }
   }
   else if (call.func == "path")
   {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to
   - [#2479](https://github.com/iovisor/bpftrace/pull/2479)
 - Add trailer to truncated strings
   - [#2559](https://github.com/iovisor/bpftrace/pull/2559)
+- Add new function, `offsetof`, get the offset of the element in the struct
+  - [#2565](https://github.com/iovisor/bpftrace/pull/2565)
 #### Changed
 - Improve attaching to uprobes with size 0
   - [#2562](https://github.com/iovisor/bpftrace/pull/2562)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1569,6 +1569,17 @@ Returns size of the argument in bytes.
 Similar to C/C++ `sizeof` operator.
 Note that the expression does not get evaluated.
 
+=== offsetof
+
+.variants
+* `offsetof(STRUCT, FIELD)`
+
+*compile time*
+
+Returns offset of the field offset bytes in struct.
+Similar to kernel `offsetof` operator.
+Note that subfields are not yet supported.
+
 === str
 
 .variants
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -38,7 +38,7 @@ vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#\*])+
 builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|nsecs|pid|probe|rand|retval|sarg[0-9]|tid|uid|username
-call     avg|buf|cat|cgroupid|clear|count|delete|exit|hist|join|kaddr|kptr|ksym|lhist|macaddr|max|min|ntop|override|print|printf|cgroup_path|reg|signal|sizeof|stats|str|strerror|strftime|strncmp|strcontains|sum|system|time|uaddr|uptr|usym|zero|path|unwatch|bswap|skboutput|pton|debugf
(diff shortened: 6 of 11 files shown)


docs/reference_guide.md
```diff
@@ -3255,6 +3256,32 @@ Attaching 1 probe...
...
+
+Syntax: `offsetof(struct, element)`
+
+Get the offset of the element in the struct.
+
+Examples:
+
+```
+#!/usr/bin/env bpftrace
+#include <linux/sched.h>
```
danobi, line 3269: Recommend dropping the `#include`. Most linux distros ship with BTF built in now; bpftrace can automatically extract all necessary definitions from it.

docs/reference_guide.md
```diff
@@ -3255,6 +3256,32 @@ Attaching 1 probe...
...
+
+BEGIN
+{
+    printf("Offsetof %ld\n", offsetof(struct task_struct, flags));
+    printf("Offsetof %ld\n", offsetof(struct task_struct, comm));
```
danobi, line 3274: Style nit

```suggestion
    printf("Offset of flags: %ld\n", offsetof(struct task_struct, flags));
    printf("Offset of comm: %ld\n", offsetof(struct task_struct, comm));
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -514,7 +519,51 @@ void SemanticAnalyser::visit(Call &call)
...
+        {
+          LOG(ERROR, call.loc, err_)
+              << call.func << "() unsupport subfield format.";
+        }
+        // a+b in offsetof(struct Foo, a+b) is illegal
+        else if (binop != nullptr)
+        {
+          LOG(ERROR, call.loc, err_)
+              << call.func << "() unsupport expression format.";
+        }
```
danobi, line 550: It looks like you're checking for very specific invalid usage. Is that necessary? Maybe it'd be simpler if we blindly coerce the second argument to an identifier. And then later when `offsetof` is processed below it'll just error and say something like "struct foo has no field 'a+b'". That error seems good enough and I can't really imagine anyone thinking `offsetoff(struct task_struct, foo+bar)` being  valid usage.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1178,6 +1227,37 @@ void SemanticAnalyser::visit(Call &call)
...
+  else if (call.func == "offsetof")
+  {
+    if (check_nargs(call, 2))
+    {
+      auto element = dynamic_cast<Identifier *>(call.vargs->at(1));
+      if (element == nullptr)
+      {
+        LOG(ERROR, call.loc, err_)
+            << call.func << "() only support single word right now.";
+      }
```
danobi, line 1239: This branch is an internal error, no? We do error checking and coercion above. Maybe it should just panic or throw an internal exception.

danobi (commented):
pls also update https://github.com/iovisor/bpftrace/blob/master/man/adoc/bpftrace.adoc . I know it's annoying to update 2 things every time but we'll fix it at some point (hopefully soon).

also looks like there's a codegen test failure:
```
4: /usr/local/bin/llvm-as: /home/runner/work/bpftrace/bpftrace/tests/codegen/llvm/call_offsetof.ll:9:19: error: expected type
4: define i64 @BEGIN(ptr %0) section "s_BEGIN_1" {
4:                   ^
4: [ FAILED ]	/home/runner/work/bpftrace/bpftrace/tests/codegen/llvm/call_offsetof.ll
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -514,7 +519,51 @@ void SemanticAnalyser::visit(Call &call)
...
+        // a+b in offsetof(struct Foo, a+b) is illegal
+        else if (binop != nullptr)
+        {
+          LOG(ERROR, call.loc, err_)
+              << call.func << "() unsupport expression format.";
+        }
+        else
+        {
+          auto &field_arg = *call.vargs->at(1);
+          String &field_str = static_cast<String &>(field_arg);
```
danobi, line 554: Hmm, not sure this works. The second argument doesn't _have_ to be a string -- this might just segfault if it's something else. In fact I think it cannot be a string -- a string would be `"field"` (note the quotes). This might only be working b/c `String` has a similar layout to `Identifier` which is what you actually want here. And I think this needs to be a `dynamic_cast` to `Identifier` similar to the checks above

Rtoax:
> pls also update https://github.com/iovisor/bpftrace/blob/master/man/adoc/bpftrace.adoc . I know it's annoying to update 2 things every time but we'll fix it at some point (hopefully soon).
> 
> also looks like there's a codegen test failure:
> 
> ```
> 4: /usr/local/bin/llvm-as: /home/runner/work/bpftrace/bpftrace/tests/codegen/llvm/call_offsetof.ll:9:19: error: expected type
> 4: define i64 @BEGIN(ptr %0) section "s_BEGIN_1" {
> 4:                   ^
> 4: [ FAILED ]	/home/runner/work/bpftrace/bpftrace/tests/codegen/llvm/call_offsetof.ll
> ```

I think this is due to the difference in LL format due to different LLVM versions?

src/ast/passes/semantic_analyser.cpp
Rtoax: `offsetof(struct task_struct, comm)`'s `comm` will `dynamic_cast` to `nullptr`, cause `comm` is a builtin. same as `offsetof(struct Foo, offsetof)`'s `offsetof`, `offsetof(struct Foo, bswap)`'s `bswap`. I don't find a better way to fix these situation.

src/ast/passes/semantic_analyser.cpp
Rtoax: I just remove it in newest force push.

src/ast/passes/semantic_analyser.cpp
Rtoax: I just force push a resolve, please review.

src/ast/passes/semantic_analyser.cpp
Rtoax: Fix it.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -514,7 +519,48 @@ void SemanticAnalyser::visit(Call &call)
...
+      //    int offsetof;   > offsetof(struct foo, offsetof)
+      //    int kstack;     > offsetof(struct foo, kstack)
+      //    int ctx;        > offsetof(struct foo, ctx)
+      //    int comm;       > offsetof(struct foo, comm)
+      //  }
+      if (func_ == "offsetof" && func_arg_idx_ == 1)
+      {
+        auto identifier = dynamic_cast<Identifier *>(call.vargs->at(1));
+        auto builtin = dynamic_cast<Builtin *>(call.vargs->at(1));
+        auto _call = dynamic_cast<Call *>(call.vargs->at(1));
```
Rtoax, line 537: @danobi We must handle the builtins and calls, like `offsetof(struct foo, offsetof)` and `offsetof(struct foo, comm)`, because builtin and call will `dynamic_cast` to `nullptr`.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1147,6 +1147,15 @@ void CodegenLLVM::visit(Call &call)
   {
     expr_ = b_.getInt64(call.vargs->at(0)->type.GetSize());
   }
+  else if (call.func == "offsetof")
+  {
+    auto identifier = dynamic_cast<Identifier *>(call.vargs->at(1));
+    if (identifier)
```
danobi, line 1153: Perhaps instead of:

```
if (identifier)
{
  ...
}
```

Do a:

```
assert(identifier);
...
```

At this point in codegen it's an internal bug if the argument is not an identifier

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1178,6 +1224,33 @@ void SemanticAnalyser::visit(Call &call)
 
     call.type = CreateUInt64();
   }
+  else if (call.func == "offsetof")
+  {
+    if (check_nargs(call, 2))
+    {
+      // Handle Identifier only, see visit(Call) of offsetof()
```
danobi, line 1231: Comment is a bit confusing. Suggestion:
```suggestion
      // Handle Identifier only, see earlier in this function where we accept()  call arguments
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1178,6 +1224,33 @@ void SemanticAnalyser::visit(Call &call)
...
+      auto element = dynamic_cast<Identifier *>(call.vargs->at(1));
+      if (element)
```
danobi, line 1233: Similar to above, probably better to use an `assert()` here to document the invariant. It's an internal bug if the first arg is not an `Identifier`

danobi (commented):
Overall looks good to me.

Regarding your comments about handling the `offsetof(struct task_struct, comm)` edge case where `comm` is also a builtin, I think an alternative approach is to add `offsetof()` directly into the parser. Similar to what https://github.com/iovisor/bpftrace/pull/2560 does for `sizeof()`.  That would probably make the AST less ambiguous.

But I am also happy enough with the current approach b/c changes are fairly localized to `Call` handling rather than being spread out across the code base. @lenticularis39 do you have any thoughts? Maybe I am missing other edge cases.

src/ast/passes/semantic_analyser.cpp
danobi: Yes, good catch

src/ast/passes/semantic_analyser.cpp
Rtoax: Fix it, thanks.

src/ast/passes/codegen_llvm.cpp
Rtoax: Fixed, thank you!!

src/ast/passes/semantic_analyser.cpp
```diff
@@ -514,7 +519,48 @@ void SemanticAnalyser::visit(Call &call)
...
+      {
+        auto identifier = dynamic_cast<Identifier *>(call.vargs->at(1));
+        auto builtin = dynamic_cast<Builtin *>(call.vargs->at(1));
+        auto _call = dynamic_cast<Call *>(call.vargs->at(1));
+
+        if (identifier)
+        {
+          expr = new Identifier(identifier->ident, call.loc);
+          expr->accept(*this);
+        }
```
viktormalik, line 543: This entire branch doesn't seem necessary.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -514,7 +519,48 @@ void SemanticAnalyser::visit(Call &call)
...
+        else if (builtin)
+        {
+          expr = new Identifier(builtin->ident, call.loc);
```
viktormalik, line 546: Isn't this introducing a memleak? Who deletes the previous `Identifier`?

src/ast/passes/semantic_analyser.cpp
Rtoax: Yes, you are right, i just remove the whole `ident` branch. thanks!

src/ast/passes/semantic_analyser.cpp
Rtoax: I just `delete` the prev identifier, thanks for the review, please review again.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -514,7 +519,39 @@ void SemanticAnalyser::visit(Call &call)
...
+        if (_builtin)
+        {
+          delete expr;
+          expr = new Identifier(_builtin->ident, call.loc);
+        }
+        else if (_call)
+        {
+          delete expr;
+          expr = new Identifier(_call->func, call.loc);
+        }
```
viktormalik, line 547:
```suggestion
        }
        auto _ident = dynamic_cast<Identifier *>(call.vargs->at(1));
        if (!_ident)
        {
          LOG(ERROR) << "...";
        }
```
Not sure if that can happen, though.

lenticularis39:
> Regarding your comments about handling the `offsetof(struct task_struct, comm)` edge case where `comm` is also a builtin, I think an alternative approach is to add `offsetof()` directly into the parser. Similar to what #2560 does for `sizeof()`. That would probably make the AST less ambiguous.
>

IMO adding `offsetof()` as a new expression type directly into the parser makes sense, it is not a builtin function in the true sense, since its arguments are not expressions (like `sizeof`).  I suggest adding a "foreign identifier" non-terminal symbol of type `std::string` into the grammar that will expand to either an identifier, a call, a builtin or a call-builtin token. Then this new non-terminal symbol can be easily used used as an argument of `offsetof()`.

Also after merging #2560, the current approach will no longer work, since `struct ...` expressions now gets parsed as a type, which is distinct from an identifier (similarly to what happened with `sizeof`).

src/ast/passes/semantic_analyser.cpp
Rtoax: Unfortunately, this will not happen, because it will be processed as `assert(ident)` and will be printed directly:

```
Assertion `element && "offsetof() 2nd argument not an identifier"' failed.
```

Rtoax:
@lenticularis39 Thank you for the advise, @danobi @viktormalik What do you think? If we just consider single-field situation, i think this PR is OK.

danobi:
@Rtoax mind rebasing and checking if this approach still works? With the AST changes in #2560 now merged, it sounds like it might break this PR. Apologies for the churn (if any) :(

Rtoax:
> @Rtoax mind rebasing and checking if this approach still works? With the AST changes in #2560 now merged, it sounds like it might break this PR. Apologies for the churn (if any) :(

Unfortunately, after rebase to master, This PR's offsetof() does not work. I tried to implement Offsetof like Sizeof, and submit a new PR as soon as possible. Thanks.

Rtoax:
@danobi @viktormalik I just start a new PR https://github.com/iovisor/bpftrace/pull/2579, please review.

PR #2568, opened 2023-04-18
[RFC] Add docker image to build and run bpftrace for both amd64, arm64 and powerpc

eiffel-fl:
Hi.

With this PR, I added a new `dockerfile` which is used to first build `bpftrace` and then make it usable for both `amd64` and `arm64`.
Note that, for `arm64`, `bpftrace` is cross-compiled so it does not take more than time than building it for `amd64`.

You can build the image for both architecture with:

```bash
$ docker buildx build --platform=linux/amd64,linux/arm64 -t bpftrace -f docker/Dockerfile.cross
```

Then you can run it with:

```bash
$ docker buildx build --load --platform=linux/amd64 -t bpftrace-amd -f docker/Dockerfile.cross
$ docker run --rm -ti --privileged bpftrace-amd
root@d32016dabfb4:/# bpftrace -e 'kprobe:do_nanosleep { printf("PID %d sleeping...\n", pid); }'
PID 2209 sleeping...
PID 62649 sleeping...
PID 2209 sleeping...
^C
```

You can of course test the `bpftrace-arm64` with the same way on an `arm64` computer:

```bash
$ lscpu | head -1
Architecture:           aarch64
$ docker run --rm --privileged -ti bpftrace-arm
root@78893e525312:/# bpftrace -e 'kprobe:do_nanosleep { printf("PID %d sleeping...\n", pid); }'
PID 4581 sleeping...
PID 3754 sleeping...
^C
```

I would like to get your opinion on this contribution as I think it can replace several dockerfiles here.
Nonetheless, I also would like to get specific reviews, particularly on the `cmake` changes and also on how to use `STATIC_LINKING` as I failed to use it with both `libclang-dev` installed from `debian` repo as I should miss including some static libraries and with `EMBED_USE_LLVM` and `LLVM_VERSION=12` as I got an error regarding missing `clang-c/Index.h`.

Best regards and thank you in advance.

### DIFF
--- a/cmake/embed/embed_llvm.cmake
+++ b/cmake/embed/embed_llvm.cmake
@@ -173,7 +173,11 @@ endif()
 
 set(LLVM_EMBEDDED_CMAKE_TARGETS "")
 
-include_directories(SYSTEM ${EMBEDDED_LLVM_INSTALL_DIR}/include)
+# The below .. is needed otherwise the Demangle header is not found:
+#0 81.26 /bpftrace/src/cxxdemangler/cxxdemangler_llvm.cpp:3:10: fatal error: llvm/Demangle/Demangle.h: No such file or directory
+#0 81.26     3 | #include <llvm/Demangle/Demangle.h>
+#0 81.26       |
+include_directories(SYSTEM ${EMBEDDED_LLVM_INSTALL_DIR}/../include)
 
 foreach(llvm_target IN LISTS LLVM_LIBRARY_TARGETS)
   list(APPEND LLVM_EMBEDDED_CMAKE_TARGETS ${llvm_target})
--- a/docker/Dockerfile.cross
+++ b/docker/Dockerfile.cross
@@ -1,0 +1,79 @@
+ARG BUILDER_IMAGE=debian:bullseye
+ARG BASE_IMAGE=debian:bullseye
+ARG LLVM=ghcr.io/iovisor/llvm
+FROM ${LLVM} as llvm
+FROM --platform=${BUILDPLATFORM} ${BUILDER_IMAGE} as builder
+ARG LLVM_VERSION=12
+ARG BUILD_TYPE=Release
+ARG TARGETARCH
+ARG BUILDARCH
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+ENV TARGETARCH=${TARGETARCH}
+ENV BUILDARCH=${BUILDARCH}
+
+COPY --from=llvm /usr/local /usr/local
+
+RUN if [ "${TARGETARCH}" != "${BUILDARCH}" ]; then \
+    PACKAGEARCH="${TARGETARCH}"; \
+    if [ "${TARGETARCH}" = 'ppc64le' ]; then \
+      PACKAGEARCH='ppc64el'; \
+    fi; \
+    dpkg --add-architecture ${PACKAGEARCH}; \
+  fi \
+  && apt-get update \
+  && apt-get install -y \
+    bison \
+    cmake \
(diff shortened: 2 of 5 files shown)


danobi:
This looks fairly reasonable at a glance. 

> I would like to get your opinion on this contribution as I think it can replace several dockerfiles here.

Which dockerfiles do you think this can replace? My main concern is that we'll change dependencies over time and this dockerfile will remain unexercised and be hard to debug later. I think before merging this we need to find a way to ensure this dockerfile doesn't bitrot.

Also it would be good to add a wrapper script into `scripts/` that encodes the invocation (ie `docker buildx build --load --platform=linux/amd64 -t bpftrace-amd -f docker/Dockerfile.cross`). Either that or documentation. B/c as it stands it's not obvious the purpose of the dockerfile.

CMakeLists.txt
```diff
@@ -193,8 +193,8 @@ add_definitions(-DLLVM_VERSION_PATCH=${LLVM_VERSION_PATCH})
 if(EMBED_USE_LLVM)
   include(embed_clang)
 else()
-  find_package(Clang REQUIRED)
-  include_directories(SYSTEM ${CLANG_INCLUDE_DIRS})
+  find_package(LibClang REQUIRED)
```
viktormalik, line 196: Is this going to work with static builds? It's using e.g. [CLANG_EXPORTED_TARGETS](https://github.com/iovisor/bpftrace/blob/master/src/ast/CMakeLists.txt#L78) which is defined in `ClangConfig.cmake` somewhere in the system path.

viktormalik:
> I was a bit confused about how the Dockerfile is this project are used. They are more used as building environment where the actual building of `bpftrace` occurs in a running container rather than in the building container.

You're right. I'm wondering if we could take this opportunity and clean the Dockerfiles up. Ideally separate building containers from the running containers.

> Regarding the bitroting of it, I understand your concern but if we add a job to the CI to publish this Dockerfile on every release, we will then be aware if the Dockerfile is no more up to date due to some added dependencies.

Running the container build in the CI would be good. Perhaps we could run it for every commit into the master (which would be `bpftrace:latest`) and then for each release (`bpftrace:v0.xx`). All of these could be published.

> Also, do you have any clue on the root cause of the missing `<clang-c/Index.h>` problem I am getting while buildin with `-DEMBED_USE_LLVM:BOOL=ON -DEMBED_BUILD_LLVM:BOOL=ON -DSTATIC_LINKING:BOOL=ON -DEMBED_LLVM_VERSION=12`? I think this is due because this files belongs to `libclang-common` and maybe something in the whole `cmake` stuff cannot find it? Sadly, my `cmake`-fu is quited limited sweat_smile.

Not sure but I'd guess that `find_package(Clang)` did more than your `FindLibClang.cmake` is doing.

CMakeLists.txt
viktormalik: It looks like the Alpine build is failing on many undefined reference to Clang functions. This change is probably the reason.

danobi:
Thanks for doing all this research! Being able to easily cross compile statically linked bpftrace binaries seems really useful.

I’m thinking about this again and my main concern is maintenance. An unreferenced dockerfile will certainly bitrot over time. So I think this new dockerfile would need to replace/integrate with the existing embedded build workflow (https://github.com/iovisor/bpftrace/blob/master/.github/workflows/embedded.yml). 

I think the existing embedded workflow only builds for x86. So I think in theory your cross compile dockerfile should be usable.

danobi:
I'm going to close this for now. Sorry about all the long term churn here :(

PR #2569, opened 2023-04-20
CMake: make libebl independent of libdw

viktormalik:
#2555 added a search for libebl into `FindLibDw.cmake`. The problem is that distros only provide libebl as a static library which makes the search for libdw fail when linking dynamically.

Since libebl is necessary for static builds only, this separates its search into `FindLibEbl.cmake` and require it from static builds.

In addition, libebl is not necessary on all systems (e.g. Alpine doesn't have it), so we make its requirement optional. This may cause some trouble when running a static build on a system *with* libdw but *without* libebl but since those two are usually packaged in the same package, that shouldn't occur often.

##### Checklist


viktormalik:
cc @haosdent

PR #2570, opened 2023-04-20
17 1 release

fbs:
@viktormalik


fbs:
it should merge into the v0.17.1_release branch, we can then tag that as v0.17.1 and create the release entry

viktormalik (approved):
I agreed with @olsajiri that he'll update libbpf in F37 but I would still proceed with this in case other systems are affected.

PR #2573, opened 2023-04-21
Use blacklist to filter unavailable kprobes

Rtoax:
Some functions in the kernel are explicitly designated as non-kprobe, such as do_trap() in the kernel with such a declaration:

  NOKPROBE_SYMBOL(do_trap);

So the file /sys/kernel/debug/kprobes/blacklist contains the do_trap() function. Now, the -l parameter of bpftrace can be found in this kprobe:

  $ sudo bpftrace -l | grep do_trap
  kprobe:do_trap

If attach kprobe:do_trap will report an error directly, which is not the result we are seeing. We should filter out this kprobe as early as possible.

  $ sudo bpftrace -e 'kprobe:do_trap {}'
  Attaching 1 probe...
  ERROR: Error attaching probe: 'kprobe:do_trap'

Because do_trap() is in the available_filter_functions:

  $ sudo grep do_trap /sys/kernel/debug/tracing/available_filter_functions
  do_trap

So using a blacklist to filter out do_trap() is a good solution:

  $ sudo grep do_trap /sys/kernel/debug/kprobes/blacklist
  0xffffffff846258d0-0xffffffff846259f0	do_trap

### DIFF
--- a/src/probe_matcher.cpp
+++ b/src/probe_matcher.cpp
@@ -112,8 +112,7 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     case ProbeType::kprobe:
     case ProbeType::kretprobe:
     {
-      symbol_stream = get_symbols_from_file_safe(
-          tracefs::available_filter_functions());
+      symbol_stream = get_symbols_from_traceable_funcs();
       ignore_trailing_module = true;
       break;
     }
@@ -206,6 +205,24 @@ std::unique_ptr<std::istream> ProbeMatcher::get_symbols_from_file(
   }
 
   return file;
+}
+
+std::unique_ptr<std::istream> ProbeMatcher::get_symbols_from_traceable_funcs(
+    void) const
+{
+  std::string funcs;
+  for (auto& func_mod : bpftrace_->traceable_funcs_)
+  {
+    if (func_mod.second.empty() || *func_mod.second.begin() == "vmlinux")
+    {
+      funcs += func_mod.first + "\n";
+    }
+    else
+    {
+      funcs += func_mod.first + " [" + *func_mod.second.begin() + "]\n";
+    }
+  }
+  return std::make_unique<std::istringstream>(funcs);
 }
 
 std::unique_ptr<std::istream> ProbeMatcher::get_symbols_from_file_safe(
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -20,6 +20,7 @@
 #include <unistd.h>
 
 #include "bpftrace.h"
+#include "debugfs.h"
 #include "log.h"
 #include "probe_matcher.h"
 #include "tracefs.h"
@@ -1071,6 +1072,19 @@ FuncsModulesMap get_traceable_funcs()
       func_mod.second = "vmlinux";
     result[func_mod.first].insert(func_mod.second);
   }
+
+  // Filter out functions from the kprobe blacklist.
+  const std::string kprobes_blacklist_path = debugfs::kprobes_blacklist();
+  std::ifstream kprobes_blacklist_funs(kprobes_blacklist_path);
+  while (std::getline(kprobes_blacklist_funs, line))
+  {
+    auto addr_func_mod = split_addrrange_symbol_module(line);
+    if (result.find(std::get<1>(addr_func_mod)) != result.end())
+    {
+      result.erase(std::get<1>(addr_func_mod));
+    }
+  }
+
   return result;
 #endif
 }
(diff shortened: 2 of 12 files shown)


danobi (commented):
looks good overall; thanks for the PR!

just to confirm: `-l` still lists blacklisted kprobes, right? But now attaching to them will generate a better error message?

danobi:
also looks like you need a clang-format

src/utils.cpp
```diff
@@ -1296,4 +1310,26 @@ std::pair<std::string, std::string> split_symbol_module(
                          symbol.length() - idx - strlen(" []")) };
 }
 
+// Usually the /sys/kernel/debug/kprobes/blacklist file.
+// Format example:
+// 0xffffffff85201511-0xffffffff8520152f	first_nmi
+// 0xffffffffc17e9373-0xffffffffc17e94ff	vmx_vmexit [kvm_intel]
+std::tuple<std::string, std::string, std::string> split_addrrange_symbol_module(
+    const std::string &symbol)
```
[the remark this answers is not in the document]
Rtoax: The reason for this use is to be consistent with the existing split_symbol_module() usage, just to deal with the parsing of /sys/kernel/debug/kprobes/blacklist, I wrote detailed input and output comments in front of the function, and added unit tests. Please review again, thanks!

Rtoax:
> looks good overall; thanks for the PR!
> 
> just to confirm: `-l` still lists blacklisted kprobes, right? But now attaching to them will generate a better error message?

The `-l` parameter of this PR will no longer lists blacklisted kprobes.

```
$ sudo bpftrace -l | grep do_trap
# Get nothing
```

Rtoax:
> also looks like you need a clang-format

There is still a clang-format alarm that is not related to this PR, does it need to be modified?

```diff
diff --git a/src/probe_matcher.cpp b/src/probe_matcher.cpp
index 8dd9999..081dfb2 100644
--- a/src/probe_matcher.cpp
+++ b/src/probe_matcher.cpp
@@ -[5](https://github.com/iovisor/bpftrace/actions/runs/4771176620/jobs/8482932890?pr=2573#step:8:6)[6](https://github.com/iovisor/bpftrace/actions/runs/4771176620/jobs/8482932890?pr=2573#step:8:7)4,[7](https://github.com/iovisor/bpftrace/actions/runs/4771176620/jobs/8482932890?pr=2573#step:8:8) +564,[8](https://github.com/iovisor/bpftrace/actions/runs/4771176620/jobs/8482932890?pr=2573#step:8:9) @@ std::set<std::string> ProbeMatcher::expand_probetype_userspace(
     const std::string& probe_type)
 {
   if (has_wildcard(probe_type))
-    return get_matches_in_stream(probe_type, false, true, *userspace_probe_list());
+    return get_matches_in_stream(
+        probe_type, false, true, *userspace_probe_list());
   else
     return { probe_type };
 }
```

danobi (approved):
lgtm! 

> There is still a clang-format alarm that is not related to this PR, does it need to be modified?

Yes please do.
 
I think you did modify that line (adding another bool). Which is why clang format CI job is picking up on it. In general if the CI says to fix formatting please do -- we decided a long time ago (for better or for worse) to gradually format the code base. Meaning any time old unformatted code is touched we require formatting be fixed. As opposed to a big formatting commit that you would throw into `.git-blame-ignore-revs`

Rtoax:
> lgtm!
> 
> > There is still a clang-format alarm that is not related to this PR, does it need to be modified?
> 
> Yes please do.
> 
> I think you did modify that line (adding another bool). Which is why clang format CI job is picking up on it. In general if the CI says to fix formatting please do -- we decided a long time ago (for better or for worse) to gradually format the code base. Meaning any time old unformatted code is touched we require formatting be fixed. As opposed to a big formatting commit that you would throw into `.git-blame-ignore-revs`

Hi, @danobi  , i just fix the clang-format alarm. Please review, thanks!!

src/utils.cpp
```diff
@@ -1296,4 +1310,29 @@ std::pair<std::string, std::string> split_symbol_module(
...
 
+// Usually the /sys/kernel/debug/kprobes/blacklist file.
+// Format example:
+// 0xffffffff85201511-0xffffffff8520152f	first_nmi
+// 0xffffffffc17e9373-0xffffffffc17e94ff	vmx_vmexit [kvm_intel]
+// The outputs are:
+// { "0xffffffff85201511-0xffffffff8520152f", "first_nmi", "" }
+// { "0xffffffffc17e9373-0xffffffffc17e94ff", "vmx_vmexit", "kvm_intel" }
+std::tuple<std::string, std::string, std::string> split_addrrange_symbol_module(
+    const std::string &symbol)
```
viktormalik, line 1321: Do we really need such a complicated implementation when we practically need only the function name?

src/probe_matcher.cpp
```diff
@@ -115,6 +123,7 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
       symbol_stream = get_symbols_from_file_safe(
           tracefs::available_filter_functions());
```
viktormalik, line 124: If we changed this to get symbols from `bpftrace_.traceable_funcs_`, we could completely avoid addition of the `ignore_kprobes_blacklist` parameter.

Rtoax: The method you said is a good idea, but the amount of code may be larger and more complex, because you need to convert FuncsModulesMap to std::istream, add ignore_kprobes_blacklist is more concise, what do you think? Thank you for your review!!

src/utils.cpp
Rtoax: Here I want to be consistent with `split_symbol_module()`, the blacklist contains ('func', 'module') information, which may be used in subsequent development. Moreover, this function does not require additional maintenance costs.

viktormalik: Ok, the module name will probably be used eventually, so let's keep it as it is.

src/probe_matcher.cpp
viktormalik: Shouldn't be that hard:
```cpp
std::string funcs;
for (auto &mod : bpftrace_.traceable_funcs_)
{
  for (auto &fun : mod->second)
    funcs += fun->second + "\n";
}
return std::make_unique<std::istringstream>(funcs);
```
Or you can do:
```cpp
    funcs += fun->second + " [" + mod->first + "]\n";
```
if you want the module, too (but we're ignoring it for kprobes anyway).

An advantage is that this way, we don't parse `/sys/kernel/tracing/available_filter_functions` twice.

Rtoax: @viktormalik Thanks very much, i'm not familiar with C++, I change the code according to your suggestion, please review.

danobi (approved):

viktormalik (approved):

PR #2578, opened 2023-04-24
Make `args` a structure (instead of a pointer)

viktormalik:
Historically, `args` has been treated as a pointer as it was closest to the implementation (as a context of the BPF program). From the user point of view, this adds an unnecessary complexity (need to dereference) as the value of the raw pointer itself is never read.
    
This PR changes `args` to a standard structure type. This allows to access the individual arguments with the dot operator (`args.arg`). For backwards compatibility, dereferencing args is still possible, it simply has no effect.
    
Also some refactoring is done which now allows to print all arguments of kfuncs using `print(args)`.
    
The change has to be reflected in several places:
- Tests: I was considering transforming all `args->` tests to use `args.` but that would break Git history and decouple the tests from the original changes that introduced them. Therefore, I just duplicated some of the existing tests to use `args.` instead of `args->`.
- Documentation: I think that plain dot access should now be the canonical way to access probe arguments, so I replaced all occurrences of `args->` by `args.` in the reference guide, manpage, and one-liners tutorial. I executed all of the modified examples and found no issue. Unfortunately, I don't have the capacity to update the Chinese and Japanese tutorials so if we decide to merge this change, I'd suggest pinging people who originally contributed them to do the update.
- Tools: same as with the documentation, all `args->` have been replaced by `args.`.

This is quite a significant change but I kind of like it as it simplifies the language and doesn't break the existing examples/tools. Still, I'll be happy to see other's opinions.

##### Checklist


src/btf.cpp
```diff
@@ -456,16 +456,19 @@ void BTF::resolve_args(const std::string &func,
     SizedType stype = get_stype(BTFId{ .btf = func_id.btf, .id = p->type });
     stype.funcarg_idx = j;
     stype.is_funcarg = true;
-    args.insert({ str, stype });
+    args.AddField(str, stype, args.size, std::nullopt, false);
+    args.size += stype.GetSize();
```
danobi, line 460: Do we need to consider padding inside the struct?

src/btf.cpp
```diff
@@ -456,16 +456,19 @@ void BTF::resolve_args(const std::string &func,
...
   }
 
   if (ret)
   {
     SizedType stype = get_stype(BTFId{ .btf = func_id.btf, .id = t->type });
     stype.funcarg_idx = j;
     stype.is_funcarg = true;
-    args.insert({ "$retval", stype });
+    args.AddField("$retval", stype, args.size, std::nullopt, false);
+    args.size += stype.GetSize();
```
danobi, line 469: Same as above

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ and this project adheres to
 #### Changed
 - Improve attaching to uprobes with size 0
   - [#2562](https://github.com/iovisor/bpftrace/pull/2562)
+- Make `args` a structure (instead of a pointer)
+  - [#2578](https://github.com/iovisor/bpftrace/pull/2578)
+- Make `args` a structure (instead of a pointer)
```
danobi, line 27: Duplicate entry?

danobi (commented):
> Tests: I was considering transforming all args-> tests to use args. but that would break Git history and decouple the tests from the original changes that introduced them. Therefore, I just duplicated some of the existing tests to use args. instead of args->.

I don't feel very strongly about it, but I would prefer all the tests to moved to use `args.` syntax. While it's additional friction to retrieve the git history, I don't think it's a huge deal -- it's a matter of tooling. You could even add it to a `.git_blame_ignore_rev` if you want ( https://www.michaelheap.com/git-ignore-rev/ ).

The upside to migrating to `args.` syntax is the tests reflect the canonical way of using bpftrace. A lot of folks (including me) look to tests to see how to use a tool.

viktormalik:
> The upside to migrating to `args.` syntax is the tests reflect the canonical way of using bpftrace. A lot of folks (including me) look to tests to see how to use a tool.

This is a good argument. In addition, runtime tests are composed of multiple lines so the link to the original commit will likely be preserved.

src/btf.cpp
viktormalik: Good catch. In fact, kprobe (fentry/fexit) arguments are stored in a u64 array, so we always need to add 8 here. It's not a problem that we represent them using a struct b/c we use `funcarg_idx` to access them in codegen, but I added a comment explaining that to prevent confusion in future.

src/btf.cpp
viktormalik: Same as above :-)

src/dwarf_parser.cpp
```diff
@@ -295,17 +295,19 @@ std::vector<std::string> Dwarf::get_function_params(
...
   for (auto &param_die : function_param_dies(function))
   {
     Dwarf_Die type_die = type_of(param_die);
     SizedType arg_type = get_stype(type_die);
     arg_type.is_funcarg = true;
     arg_type.funcarg_idx = i++;
-    result.emplace(dwarf_diename(&param_die), arg_type);
+    result.AddField(
+        dwarf_diename(&param_die), arg_type, result.size, std::nullopt, false);
+    result.size += arg_type.GetSize();
```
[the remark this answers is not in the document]
viktormalik: Not at this moment. The struct is never built (it's just a placeholder type) and the individual args are read from the registers. Once I rebase the support for working with the entire `args` for uprobes (#2477), the struct will be built on stack, so this will matter. Still, I'll try to build the struct without padding to save stack space.

danobi (approved):

viktormalik:
This is quite a significant change so I'd like to have more approvals (@ajor, @fbs).

ajor (changes requested):
I found a couple of bugs when playing around with this branch:

### Error reading args in tracepoint
```
# ./build/src/bpftrace -e 'tracepoint:syscalls:sys_enter_read { @ = args }'
Attaching 1 probe...
ERROR: failed to create map: '@': Invalid argument
```

### Abort when reading args with no names
bpftest.cpp:
```
extern "C" {
void foo(int, int) { }
}
int main() { }
```
`g++ -g bpftest.c -O0`
```
# ./build/src/bpftrace -e 'uprobe:/tmp/a.out:foo { @ = args }' -d
libbpf: (2) libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
libbpf: (2) libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted
```

src/ast/ast.cpp
```diff
@@ -580,6 +580,11 @@ std::string Probe::name() const
   return n;
 }
 
+std::string Probe::args_typename() const
+{
+  return "struct " + name() + "_args";
```
ajor, line 585: I don't understand the code fully myself yet, but I had a couple of questions about this at first glance:
- Does this name get exposed to users anywhere? e.g. in casting
- Could it conflict with a user-defined type name? e.g. if they have a struct called "input_args" and a function called "input"

ajor (commented):
Sorry for the delay in reviewing this. I'm in France this week and don't have much time, but I will try and go over this fully when I'm back on Monday.

src/struct.cpp
```diff
@@ -204,4 +209,14 @@ size_t StructManager::GetTuplesCnt() const
...
 }
 
+const Field *StructManager::GetProbeArg(const ast::Probe &probe,
+                                        const std::string &arg_name)
+{
+  auto args = Lookup(probe.args_typename());
+  if (args.expired() || !args.lock()->HasField(arg_name))
+    return nullptr;
+
+  return &args.lock()->GetField(arg_name);
```
ajor, line 219: I think there's a potential race condition in this code (even if it won't be hit in the current implementation, if we're using shared_ptrs we should prepare for this as someone else might change the code in the future).

It'd be safer to do something like this instead:
```
if (auto observe = weak.lock()) {
  // use *observe
} else {
  return nullptr;
}
```
https://en.cppreference.com/w/cpp/memory/weak_ptr/lock

src/ast/passes/field_analyser.cpp
```diff
@@ -289,21 +271,17 @@ bool FieldAnalyser::resolve_args(Probe &probe)
     }
 
     // check if we already stored arguments for this probe
-    auto it = bpftrace_.ap_args_.find(probe_->name());
-
-    if (it != bpftrace_.ap_args_.end())
+    auto args = bpftrace_.structs.Lookup(probe.args_typename());
+    if (!args.expired() && *args.lock() != probe_args)
```
ajor, line 275: Also potentially racey

src/struct.cpp
```diff
@@ -173,6 +173,11 @@ void StructManager::Add(const std::string &name,
   struct_map_[name] = std::make_unique<Struct>(size, allow_override);
 }
 
+void StructManager::Add(const std::string &name, Struct &&record)
+{
+  struct_map_[name] = std::make_unique<Struct>(record);
```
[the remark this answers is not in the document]
ajor: Also, it looks like this could be changed to `make_shared`, to be clearer about what it's doing and it'd be a bit more efficient with its allocations.

src/ast/ast.cpp
viktormalik: - If a user tries to cast `args` then the error message will contain this type name:
```
# bpftrace -e 'kfunc:vfs_read { printf("%d\n", ((struct foo)args).count); exit(); }'
stdin:1:33-47: ERROR: Cannot resolve unknown type "struct foo"
kfunc:vfs_read { printf("%d\n", ((struct foo)args).count); exit(); }
                                ~~~~~~~~~~~~~~
stdin:1:33-47: ERROR: Cannot cast from struct type "struct kfunc:vmlinux:vfs_read_args"
kfunc:vfs_read { printf("%d\n", ((struct foo)args).count); exit(); }
```
- `name()` returns the full probe name, so the type would be called `struct uprobe:bin:input_args` which is not a valid C identifier, so it cannot conflict a user type.

viktormalik:
> Sorry for the delay in reviewing this. I'm in France this week and don't have much time, but I will try and go over this fully when I'm back on Monday.

No worries, took me a while to get to this, too. Thanks for finding these problems, they should be fixed now (I hope I got the race conditions right). Feel free to run more checks, thanks!

ajor (approved):
Looks good to me!

PR #2582, opened 2023-04-27
Fix the problem of kfunc loading failure in 5.10 kernel which doesn't support module btf

chengshuyi:
The first patch add function has_module_btf to check if the kernel supports module btf.
 
The second patch fix the problem of kfunc loading failure in 5.10 kernel which doesn't support module btf. And this patch does two things:

- Modified the load_kernel_btfs function: when the kernel does not support the module btf feature,
    it is not necessary to read the module btf.
   
 - The get_btf_id_fd function is modified: when the number of btfojb is 0, it is not necessary to
  obtain its fd through bpf_btf_get_fd_by_id. This avoids assigning wrong values to attach_btf_obj_fd.
  When the value of attach_btf_obj_fd is 0, the kernel will find the corresponding kernel symbol
  from the vmlinux btf according to the btf id.

Best regards and thank you in advance.

##### Checklist

- [] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [] The new behaviour is covered by tests

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -82,6 +82,9 @@ void BTF::load_kernel_btfs(const std::set<std::string> &modules)
   btf_objects.push_back(
       BTFObj{ .btf = vmlinux_btf, .id = 0, .name = "vmlinux" });
 
+  if (bpftrace_ && !bpftrace_->feature_->has_module_btf())
+    return;
+
   // Note that we cannot parse BTFs from /sys/kernel/btf/ as we need BTF object
   // IDs, so the only way is to iterate through all loaded BTF objects
   __u32 id = 0;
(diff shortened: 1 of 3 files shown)


src/btf.cpp
```diff
@@ -722,8 +725,11 @@ std::pair<int, int> BTF::get_btf_id_fd(const std::string &func,
       continue;
 
     auto id = find_id_in_btf(btf_obj.btf, func, BTF_KIND_FUNC);
-    if (id >= 0)
+    if (id >= 0) {
+      if (btf_obj.id == 0)
+        return { id, 0 };
       return { id, bpf_btf_get_fd_by_id(btf_obj.id) };
```
viktormalik, line 731: Perhaps more compact variant:
```suggestion
      return { id, btf_obj.id ? bpf_btf_get_fd_by_id(btf_obj.id) : 0 };
```

viktormalik (commented):
Thanks for the PR! Looks mostly good to me, just a couple of suggestions below.

viktormalik:
Also please apply clang-format to the final version.

viktormalik (approved):

PR #2584, opened 2023-04-27
Replace with correct macro HAVE_UNSAFE_PROBE

dubeyabhishek:
With reference to change e31e398f4, "bpftrace --info" was always reporting 'no' for 'Build: unsafe uprobe'.


src/build_info.cpp
```diff
@@ -13,7 +13,7 @@ std::string BuildInfo::report()
       << "  LLVM: " << LLVM_VERSION_MAJOR << "." << LLVM_VERSION_MINOR << "."
       << LLVM_VERSION_PATCH << std::endl
       << "  unsafe uprobe: "
```
viktormalik, line 15: Also please change this to:
```suggestion
      << "  unsafe probes: "
```

viktormalik (commented):
Good catch! One suggestion below.

viktormalik (approved):

PR #2586, opened 2023-05-03
Add BPFTRACE_STACK_MODE env variable

jordalgo:
Summary:
This is so users can more easily set the global output for their stacks rather than repeating the same string for every invocation of ustack or kstack

Test Plan:
Tested locally.
```
BPFTRACE_STACK_MODE=tomato sudo ./src/bpftrace -e 'kprobe:do_nanosleep {
printf("Jordan\n%s\n", ustack());  }'
ERROR: Env var 'BPFTRACE_STACK_MODE' did not contain a valid StackMode:
tomato
```

```
BPFTRACE_STACK_MODE=perf sudo ./src/bpftrace -e 'kprobe:do_nanosleep {
printf("Jordan\n%s\n", ustack());  }'
Attaching 1 probe...
Jordan

	7f088cae85b3 __clock_nanosleep+99
(/usr/local/fbcode/platform010-compat/lib/libc.so.6)
	7f088caed253 __nanosleep+19
(/usr/local/fbcode/platform010-compat/lib/libc.so.6)
	1f3418b
scribe::common::sink::QueueSink::sendLoop(scribe::common::sink::detail::BatchQueue*,
int)+2971 (/usr/local/fbprojects/scribe/bin/scribed)
	6187e12 folly::ThreadedExecutor::work(folly::Function<void
()>&)+82 (/usr/local/fbprojects/scribe/bin/scribed)
	7f088cedf2e5 execute_native_thread_routine+21
(/usr/local/fbcode/platform010-compat/lib/libstdc++.so.6.0.29)
	7f088ca979bf start_thread+559
(/usr/local/fbcode/platform010-compat/lib/libc.so.6)
	7f088cb288ac __GI___clone3+44
(/usr/local/fbcode/platform010-compat/lib/libc.so.6)
```

Reviewers:

Subscribers:

Tasks:

Tags:

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -176,6 +176,7 @@ ENVIRONMENT:
     BPFTRACE_CACHE_USER_SYMBOLS [default: auto] enable user symbol cache
     BPFTRACE_VMLINUX            [default: none] vmlinux path used for kernel symbol resolution
     BPFTRACE_BTF                [default: none] BTF file
+    BPFTRACE_STACK_MODE         [default: bpftrace] Output format for ustack and kstack builtins
 
 EXAMPLES:
 bpftrace -l '*sleep*'
@@ -553,6 +554,12 @@ Default: `..`
 
 Trailer to add to strings that were truncated. Set to empty string to disable truncation trailers.
 
+### 9.11 `BPFTRACE_STACK_MODE`
+
+Default: bpftrace
+
+Output format for ustack and kstack builtins. Available modes/formats: `bpftrace`, `perf`, and `raw`.
+This can be overwritten at the call site.
 
 ## 10. Clang Environment Variables
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -128,6 +128,7 @@ void usage()
   std::cerr << "    BPFTRACE_VMLINUX            [default: none] vmlinux path used for kernel symbol resolution" << std::endl;
   std::cerr << "    BPFTRACE_BTF                [default: none] BTF file" << std::endl;
   std::cerr << "    BPFTRACE_STR_TRUNC_TRAILER  [default: '..'] string truncation trailer" << std::endl;
+  std::cerr << "    BPFTRACE_STACK_MODE         [default: bpftrace] Output format for ustack and kstack builtins" << std::endl;
   std::cerr << std::endl;
   std::cerr << "EXAMPLES:" << std::endl;
   std::cerr << "bpftrace -l '*sleep*'" << std::endl;
@@ -333,6 +334,21 @@ static std::optional<struct timespec> get_boottime()
 
   if (!get_bool_env_var("BPFTRACE_VERIFY_LLVM_IR", verify_llvm_ir))
     return false;
+
+  if (const char* stack_mode = std::getenv("BPFTRACE_STACK_MODE"))
+  {
+    auto found = STACK_MODE_MAP.find(stack_mode);
+    if (found != STACK_MODE_MAP.end())
+    {
+      bpftrace.stack_mode_ = found->second;
+    }
+    else
+    {
+      LOG(ERROR) << "Env var 'BPFTRACE_STACK_MODE' did not contain a valid "
+                    "StackMode: "
+                 << stack_mode;
+    }
+  }
 
   return true;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ and this project adheres to
   - [#2579](https://github.com/iovisor/bpftrace/pull/2579)
 - Add 'StackMode::raw' for ustack and kstack formatting
   - [#2581](https://github.com/iovisor/bpftrace/pull/2581)
+- Add 'BPFTRACE_STACK_MODE' env variable
+  - [#2586](https://github.com/iovisor/bpftrace/pull/2586)
 #### Changed
 - Improve attaching to uprobes with size 0
   - [#2562](https://github.com/iovisor/bpftrace/pull/2562)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -258,6 +258,13 @@ Default: `..`
 
 Trailer to add to strings that were truncated. Set to empty string to disable truncation trailers.
 
+=== BPFTRACE_STACK_MODE
+
+Default: bpftrace
+
+Output format for ustack and kstack builtins. Available modes/formats: `bpftrace`, `perf`, and `raw`.
+This can be overwritten at the call site.
+
 ////
 !
 !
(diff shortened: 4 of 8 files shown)


jordalgo:
Not sure where to add unit tests for this.

src/main.cpp
```diff
@@ -334,6 +335,28 @@ static std::optional<struct timespec> get_boottime()
   if (!get_bool_env_var("BPFTRACE_VERIFY_LLVM_IR", verify_llvm_ir))
     return false;
 
+  if (const char* stack_mode = std::getenv("BPFTRACE_STACK_MODE"))
+  {
+    if (strcmp(stack_mode, "bpftrace") == 0)
+    {
+      bpftrace.stack_mode_ = StackMode::bpftrace;
+    }
```
viktormalik, line 343: I'd prefer the C++ way:
```suggestion
  if (const char* env_p = std::getenv("BPFTRACE_STACK_MODE"))
  {
    std::string stack_mode(env_p);
    if (stack_mode == "bpftrace")
    {
      bpftrace.stack_mode_ = StackMode::bpftrace;
    }
```
Also, I'd suggest using `switch` rather than `if ... else if ... `.

viktormalik:
> Not sure where to add unit tests for this.

I think that runtime tests (e.g. in `tests/runtime/call`) would be sufficient.

docs/reference_guide.md
```diff
@@ -553,6 +554,11 @@ Default: `..`
 
 Trailer to add to strings that were truncated. Set to empty string to disable truncation trailers.
 
+### 9.11 `BPFTRACE_STACK_MODE`
```
danobi, line 557: Please also update [bpftrace.adoc](https://github.com/iovisor/bpftrace/blob/master/man/adoc/bpftrace.adoc#environment-variables) . (We really have to unify it one of these days)

src/main.cpp
jordalgo: Switch doesn't work on strings but I"ll update the ==

jordalgo:
> > Not sure where to add unit tests for this.
> 
> I think that runtime tests (e.g. in `tests/runtime/call`) would be sufficient.

I don't see a way to verify the ustack or kstack output easily. The current ustack test just checks that the probe is attached. I could add something like this just to ensure bpftrace doesn't vomit but I don't know how useful it is:
```
NAME ustack_stack_mode_env
PROG k:do_nanosleep { printf("%s\n%s\n", ustack(), ustack(1)); exit(); }
ENV BPFTRACE_STACK_MODE=raw
EXPECT Attaching 1 probe
TIMEOUT 5
AFTER ./testprogs/syscall nanosleep  1e8
```

viktormalik:
> I don't see a way to verify the ustack or kstack output easily. The current ustack test just checks that the probe is attached. I could add something like this just to ensure bpftrace doesn't vomit but I don't know how useful it is:

Could you verify that **any** kstack/ustack was printed in the correct format?
```
# bpftrace format
EXPECT ^\s+[a-zA-Z0-9_]+
# perf format
EXPECT ^\s+[0-9a-f]+ [a-zA-Z0-9_]+
# raw format
EXPECT ^\s+[0-9a-f]+$
```

jordalgo:
> Could you verify that **any** kstack/ustack was printed in the correct format?
> 
> ```
> # bpftrace format
> EXPECT ^\s+[a-zA-Z0-9_]+
> # perf format
> EXPECT ^\s+[0-9a-f]+ [a-zA-Z0-9_]+
> # raw format
> EXPECT ^\s+[0-9a-f]+$
> ```

I worry that the output might change slightly and then the regex would fail e.g. here is the output for this:
```
NAME ustack_stack_mode_env
PROG k:do_nanosleep { printf("%s", ustack(1)); exit(); }
ENV BPFTRACE_STACK_MODE=perf
EXPECT ^\s+[0-9a-f]+$
TIMEOUT 5
AFTER ./testprogs/syscall nanosleep  1e8
```

```
Attaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\n\n\t7f088cae85b3 __clock_nanosleep+99 (/usr/local/fbcode/platform010-compat/lib/libc.so.6)\n\n\n
```

viktormalik:
> > Could you verify that **any** kstack/ustack was printed in the correct format?
> > ```
> > # bpftrace format
> > EXPECT ^\s+[a-zA-Z0-9_]+
> > # perf format
> > EXPECT ^\s+[0-9a-f]+ [a-zA-Z0-9_]+
> > # raw format
> > EXPECT ^\s+[0-9a-f]+$
> > ```
> 
> I worry that the output might change slightly and then the regex would fail e.g. here is the output for this:
> 
> ```
> NAME ustack_stack_mode_env
> PROG k:do_nanosleep { printf("%s", ustack(1)); exit(); }
> ENV BPFTRACE_STACK_MODE=perf
> EXPECT ^\s+[0-9a-f]+$
> TIMEOUT 5
> AFTER ./testprogs/syscall nanosleep  1e8
> ```
> 
> ```
> Attaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\n\n\t7f088cae85b3 __clock_nanosleep+99 (/usr/local/fbcode/platform010-compat/lib/libc.so.6)\n\n\n
> ```

But you're using the `raw` regex for `perf` format.

jordalgo:
> But you're using the `raw` regex for `perf` format.

I was just using that as an example of the output. As the regexes for all would need to start with something that looked like this
```
^Attaching 1 probe...\\n__BPFTRACE_NOTIFY_PROBES_ATTACHED[\\n\\t]
```

Feels fragile

jordalgo:
> I don't think so, the regex should be matched against every line of the output.

Well, I don't quite understand HOW it passes but it does :) E.g. why doesn't the check fail for the line "Attaching 1 probe..." against this regex `^\s+[0-9a-f]+$`. Maybe it just needs to match against at least ONE line in the output?

danobi:
I think the regex only has to match a single line:

https://github.com/iovisor/bpftrace/blob/8e9853fb015e62968cde853dc8c1cdca8c68fec8/tests/runtime/engine/runner.py#L251
https://docs.python.org/3/library/re.html#re.search

which would match my experience (IIRC)

So I’m not sure having perf/raw/bpftrace regex is correct.

jordalgo:
> So I’m not sure having perf/raw/bpftrace regex is correct.

I guess it depends how much extra noise we get outside of the single stack trace that we think might pollute the results. The tests I added seem to do the right thing and fail if we change the StackMode...

viktormalik:
> > So I’m not sure having perf/raw/bpftrace regex is correct.
> 
> I guess it depends how much extra noise we get outside of the single stack trace that we think might pollute the results. The tests I added seem to do the right thing and fail if we change the StackMode...

Agreed with @jordalgo, the added regexes are specific enough to avoid matching anything outside of the expected stack trace.

viktormalik (approved):

danobi (approved):

PR #2587, opened 2023-05-03
Fix segfaults in dwarf_parser

viktormalik:
Programs may contain DWARF entities with no name (missing DW_AT_name attribute). This fixes segfaults for programs containing two kinds of such entities:
- subprogram entries (unnamed functions),
- function parameters.

Fixes the following issue (taken from https://github.com/iovisor/bpftrace/pull/2578#pullrequestreview-1405996930):
bpftest.cpp:
```
extern "C" {
void foo(int, int) { }
}
int main() { }
```
`g++ -g bpftest.cpp -O0`
```
# ./build/src/bpftrace -e 'uprobe:/tmp/a.out:foo { @ = args }' -d
libbpf: (2) libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
libbpf: (2) libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ and this project adheres to
   - [#2546](https://github.com/iovisor/bpftrace/pull/2546)
 - arm64: define the KASAN_SHADOW_SCALE_SHIFT macro
   - [#2518](https://github.com/iovisor/bpftrace/pull/2518)
+- Fix segfaults in dwarf_parser
+  - [#2587](https://github.com/iovisor/bpftrace/pull/2587)
 #### Docs
 #### Tools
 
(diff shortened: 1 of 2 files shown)


viktormalik:
> I think we'll need to do the same in `get_function_params` too:

Thanks, fixed.

> Are there no unit tests for dwarf_parser that these test cases could be added to?

Unfortunately not, DWARF features are tested in the FieldAnalyser unit tests. Since it's not possible to test probe listing there, I ended up adding two runtime tests for these fixes.

viktormalik:
Argh, C standard doesn't allow unnamed function params. While it works on my system, it doesn't in the CI. We could add some C++ tests but that's more work (CMake changes, etc.) and I'm not sure if it's worth it for this corner case.

I'm removing the tests for now.

viktormalik:
@ajor are you ok merging this without tests or should I try to add something?

danobi (approved):
No opinion, seems inoffensive to me

ajor (approved):
Not necessary for now, this change looks good, but I think having the ability of adding C++ test programs would be useful generally in the future

PR #2590, opened 2023-05-09
Always set libbpf print function

tnovak:
Libbpf's default print function writes all non-debug messages to stderr[1], which can be noisy: for example, as of cd8c1b188 ("bpftrace: Load vmlinux BTF using btf__load_vmlinux_btf()"), bpftrace logs a warning[2] on startup every time on older kernels that don't have BTF support.

Move the debug level check into libbpf_print() and set it unconditionally; this way libbpf messages are printed only if -d is specified.

[1] https://github.com/libbpf/libbpf/blob/v1.0.0/src/libbpf.c#L209
[2] https://github.com/libbpf/libbpf/blob/v1.0.0/src/btf.c#L4685

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -668,11 +668,27 @@ Args parse_args(int argc, char* argv[])
   return args;
 }
 
+static const char* libbpf_print_level_string(enum libbpf_print_level level)
+{
+  switch (level)
+  {
+    case LIBBPF_WARN:
+      return "WARN";
+    case LIBBPF_INFO:
+      return "INFO";
+    default:
+      return "DEBUG";
+  }
+}
+
 static int libbpf_print(enum libbpf_print_level level,
                         const char* msg,
                         va_list ap)
 {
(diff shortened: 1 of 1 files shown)


tnovak:
This patch could suppress some libbpf messages that previously would have been logged even without `-d`, but given that

- some error logging in bpftrace is already gated on `bt_debug`
- we go out of our way to prevent libbpf logging in certain cases (e.g. the use of `StderrSilencer` before calling `bpf_prog_load`)

... this may be fine?

viktormalik:
> This patch could suppress some libbpf messages that previously would have been logged even without `-d`, but given that
> 
> * some error logging in bpftrace is already gated on `bt_debug`
> * we go out of our way to prevent libbpf logging in certain cases (e.g. the use of `StderrSilencer` before calling `bpf_prog_load`)
> 
> ... this may be fine?

Yeah, I think that we shouldn't pollute the default output with libbpf log messages, unless the user wants to. I like this change.

PR #2592, opened 2023-05-10
LLVM 16

viktormalik:
This is a follow-up PR for #2528. It includes a fix for the CMake issue mentioned in https://github.com/iovisor/bpftrace/pull/2528#issuecomment-1487614582.


viktormalik:
Once this is merged, we can do a release. @fbs will you handle it?

CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.13.0)
 project(bpftrace)
 
+if(${CMAKE_VERSION} VERSION_EQUAL 3.3.0 OR ${CMAKE_VERSION} VERSION_GREATER 3.3.0)
```
ajor, line 4: We set the minimum required CMake version to 3.13 above, so I don't think this check is needed?

ajor (approved):

CMakeLists.txt
danobi: Yeah, agreed

viktormalik: Ah, of course, I just copied the thing from BCC.

viktormalik:
Merged, we're now ready to do a release.

Issue #2322, opened 2022-08-10
bpftrace should have a strong dependency on libbpf


BurntBrunch:
While working on #2321, I had to manually - and poorly - reimplement functionality that's already in libbpf in a more complete form. (namely, subprogram and func_info relocation)

I'm opening this issue to discuss making libbpf a required dependency, especially given its imminent 1.0 release. This dependency would allow us to

1) Pass in-memory ELF files to libbpf, removing all of the current orc jit hacks
2) Not worry about relocations and other BTF minutiae
3) Make the object files emitted by --emit-elf loadable by other tools (!)
4) Open a clean path to CO-RE usage (!)
5) Potentially simplify USDT handling by relying on libbpf's support
6) Simplify feature probing by relying on libbpf's probes and fallbacks

Obeying the libbpf layout requires *some* work (e.g. BTF-based map definitions and the very particular section naming conventions) but honestly that's probably more shallow a rabbit hole than what I endured to get #2321 to the finish line.

I'm opening this issue primarily because I don't know the project's policy on build dependencies and would love to discuss any objections to this change.

viktormalik:
Hi @BurntBrunch, thanks for opening this!

libbpf is currently required for building bpftrace (since #2265) so you can freely use functions that it provides (although we currently support also older libbpf versions). On the other hand, many of the ideas above are great and we should consider implementing them eventually. I'd suggest to continue discussion in this thread and then open separate issues for each of the features.

BurntBrunch:
Alright, I'm glad this is well-received! :)

How do we go about formalizing the min version? Is that strictly in the cmake config or do we do it anywhere else?
I'd love to make the version requirements pretty high (e.g. 0.8.1 or the upcoming 1.0.0) to avoid distros getting by with much older libbpf than is advisable. 

Also, does this mean we can clean up all the HAVE_LIBBPF_* checks and fallbacks? I really don't think we care about a libbpf version that doesn't have bpf_prog_load, for example.

fbs:
While from a development perspective only bleeding edge would be great we do try to support a wide range of kernels and library versions. Stable distros as still quite popular, so I'd say as low as we can reasonably go.

It's tricky, bpf moves way to fast for the distros (and us) to keep up

danobi:
I am open to this as well. I evaluated the same idea ~1 year ago but stopped short of implementing for a few reasons:

[...]

PR #2595, opened 2023-05-15
RFC: Nix flake based build

danobi:
This is an RFC for the nix build infra we discussed during office hours.

Attached in this PR is documentation on how to interact with the nix build.
Please see that for reference.

If we decide to move forward with this, we can at the very least replace/remove:

* All the docker based build stuff in the CI
* `build-libs.sh`
* `STATIC_BPF_BCC` in cmake config
* The vendored libbpf + bcc repos

We also get for free:

* Reproducible builds / tests (modulo kernel until `vmtest` comes) between CI
  and your development host. Ideally it'll just be `nix develop --command run-tests.sh`
  that's both run in CI and locally.
* The ability to patch libbpf, bcc, llvm, really any of our deps, completely through
  `flake.nix` configs. This is something that cannot be done through traditional,
  non-functional package managers. You can sort of see in the current `flake.nix`
  how I'm patching nixpkg's upstream libbpf and bcc. This is easily extended to
  include `.patch` as well.

Downsides:

* Nix's DSL is quite difficult to understand. You can probably already see that.
  * That being said, I got somewhat productive with it in one weekend. I'd also
    be happy to walk people through what I've learned.
  * It's also quite beautiful (the semantics, not the syntax) when you piece it all together.
* One more thing to learn (but hopefully not maintain, as we'll delete stuff if this goes in)
  * But the industry is moving towards this direction so hopefully another tool in your tool kit

What's left to do:

* Fix all the tests
* Hook up to CI
* Delete all the will-be-deprecated stuff

##### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,28 @@ $ vagrant up $YOUR_CHOICE
 $ vagrant ssh $YOUR_CHOICE
 ```
 
+### Nix
+
+There is experimental support for building and testing with [Nix flakes](https://nixos.wiki/wiki/Flakes).
+The quickstart follows but more detailed documentation can be found in [nix.md](docs/nix.md).
+
+First, install `nix`: https://nixos.org/download.html. Then enable Nix flake
+support:
+
+```
+$ mkdir -p ~/.config/nix
+$ echo "experimental-features = nix-command flakes" >> ~/.config/nix/nix.conf
+```
+
+Now build and run:
+
+```
+$ nix build
+$ sudo ./result/bin/bpftrace --info
+```
+
+The above steps are theoretically guaranteed to work on any linux x86-64/arm system.
+
 ## License
 
 Copyright 2019 Alastair Robertson
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,72 @@
+{
+  description = "High-level tracing language for Linux eBPF";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-22.11";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils, ... }:
+    # This flake only supports 64-bit linux systems.
+    # Note bpftrace support aarch32 but for simplicity we'll omit it for now.
+    flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ]
+      (system:
+        let
+          pkgs = import nixpkgs { inherit system; };
+        in
+        {
+          # Set formatter for `nix fmt` command
+          formatter = pkgs.nixpkgs-fmt;
+
+          # Define package set
+          packages = rec {
+            # Default package is bpftrace binary
+            default = bpftrace;
+
+            # Derivation for bpftrace binary
+            bpftrace =
+              with pkgs;
+              pkgs.stdenv.mkDerivation rec {
+                name = "bpftrace";
+
+                src = self;
+
+                nativeBuildInputs = [ cmake ninja bison flex gcc12 ];
+
+                # TODO: use submoduled bcc + libbpf
+                buildInputs = with llvmPackages_15;
+                  [
+                    asciidoctor
+                    bcc
+                    cereal
+                    elfutils
+                    gtest
+                    libbpf_1
+                    libbfd
+                    libclang
+                    libelf
+                    libffi
+                    libopcodes
+                    libpcap
+                    libsystemtap
+                    llvm
+                    pahole
+                    xxd
+                    zlib
+                  ];
+
+                # Release flags
+                cmakeFlags = [
+                  "-DCMAKE_BUILD_TYPE=Release"
+                  "-DUSE_SYSTEM_BPF_BCC=ON"
+                ];
+              };
+          };
+
+          # Define apps that can be run with `nix run`
+          apps.default = {
+            type = "app";
+            program = "${self.packages.${system}.default}/bin/bpftrace";
+          };
+        });
+}
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,46 @@
     flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ]
       (system:
         let
-          pkgs = import nixpkgs { inherit system; };
+          # Overlay to specify build should use the specific libbpf we want
+          libbpfVersion = "1.2.0";
+          libbpfOverlay =
+            (self: super: {
+              libbpf_1 = super.libbpf_1.overrideAttrs (old: {
+                version = libbpfVersion;
+                src = super.fetchFromGitHub {
+                  owner = "libbpf";
+                  repo = "libbpf";
+                  rev = "v${libbpfVersion}";
+                  # If you don't know the hash the first time, set:
+                  # hash = "";
+                  # then nix will fail the build with such an error message:
+                  # hash mismatch in fixed-output derivation '/nix/store/m1ga09c0z1a6n7rj8ky3s31dpgalsn0n-source':
+                  # specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
+                  # got:    sha256-173gxk0ymiw94glyjzjizp8bv8g72gwkjhacigd1an09jshdrjb4
+                  sha256 = "sha256-NimK4pdYcai21hZHdP1mBX1MOlNY61iDJ+PDYwpRuVE=";
+                };
+              });
+            });
+
+          # Overlay to specify build should use the specific bcc we want
+          bccVersion = "0.27.0";
+          bccOverlay =
+            (self: super: {
+              bcc = super.bcc.overridePythonAttrs (old: {
+                version = bccVersion;
+                src = super.fetchFromGitHub {
+                  owner = "iovisor";
+                  repo = "bcc";
+                  rev = "v${bccVersion}";
+                  sha256 = "sha256-+RK5RZcoNHlgMOFPgygRf2h+OZGxR9gJ+fTbYjDB6Ww=";
+                };
+                # Seems like these extra tools are needed to build bcc
+                nativeBuildInputs = old.nativeBuildInputs ++ [ pkgs.python310Packages.setuptools pkgs.zip ];
+              });
+            });
+
+          # We need to use two overlays so that bcc inherits the our pinned libbpf
+          pkgs = import nixpkgs { inherit system; overlays = [ libbpfOverlay bccOverlay ]; };
         in
         {
           # Set formatter for `nix fmt` command
@@ -33,7 +72,6 @@
 
                 nativeBuildInputs = [ cmake ninja bison flex gcc12 ];
 
-                # TODO: use submoduled bcc + libbpf
                 buildInputs = with llvmPackages_15;
                   [
                     asciidoctor
--- a/flake.nix
+++ b/flake.nix
@@ -52,18 +52,10 @@
 
           # We need to use two overlays so that bcc inherits the our pinned libbpf
           pkgs = import nixpkgs { inherit system; overlays = [ libbpfOverlay bccOverlay ]; };
-        in
-        {
-          # Set formatter for `nix fmt` command
-          formatter = pkgs.nixpkgs-fmt;
 
-          # Define package set
-          packages = rec {
-            # Default package is bpftrace binary
-            default = bpftrace;
-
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -153,6 +153,26 @@ $ vagrant up $YOUR_CHOICE
...
 
+### Nix
+
+There is experimental support for building and testing with [Nix flakes](https://nixos.wiki/wiki/Flakes).
+The quickstart follows but more detailed documentation can be found in [nix.md](docs/nix.md).
+
+First, install `nix` (see https://nixos.org/download.html):
+
+```
+$ sh <(curl -L https://nixos.org/nix/install) --daemon
```
JakeHillion, line 164: I'm not sure it's great to include the install command here, particularly as it ends up using high privilege at points. Probably best to link out to the Nix download page as you have above.

flake.nix
```diff
@@ -0,0 +1,169 @@
...
+                  rev = "v${bccVersion}";
+                  sha256 = "sha256-+RK5RZcoNHlgMOFPgygRf2h+OZGxR9gJ+fTbYjDB6Ww=";
+                };
+                # Seems like these extra tools are needed to build bcc
+                nativeBuildInputs = old.nativeBuildInputs ++ [ pkgs.python310Packages.setuptools pkgs.zip ];
+              });
+            });
+
+          # We need to use two overlays so that bcc inherits the our pinned libbpf
+          pkgs = import nixpkgs { inherit system; overlays = [ libbpfOverlay bccOverlay ]; };
```
JakeHillion, line 53: Do we need these overlays? It appears to at least build fine with the current versions in the `nixpkgs` referred to by the `flake.lock` file. I agree that they may be needed at some point in the future so perhaps it is better to have the infrastructure, and I understand the motivation to keep them synced with the submodules. Worth a discussion at least.

flake.nix
```diff
@@ -0,0 +1,169 @@
...
+
+          # Define the development shell you get when you run `nix develop`
+          devShells.default =
+            with import nixpkgs { inherit system; };
+            pkgs.mkShell rec {
+              packages = with pkgs; [
+                runTests
+                git
+                self.packages.x86_64-linux.default.nativeBuildInputs
+                self.packages.x86_64-linux.default.buildInputs
```
JakeHillion, line 165: If these stay, `s/x86_64-linux/"${system}"/`

flake.nix
```diff
@@ -0,0 +1,169 @@
+{
+  description = "High-level tracing language for Linux eBPF";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/release-22.11";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils, ... }:
+    # We only support 64-bit linux systems
```
JakeHillion, line 10: I think it's reasonable to limit this to tested systems in the beginning, but I don't think this is true of bpftrace - it supports quite a few architectures.

JakeHillion (commented):
Thank you for doing this, more Nix flake adoption in upstream repos is great to see! I commented on some more specific parts but generally.
- Rather than including the tests in the binary output they should really be included as flake checks (`outputs.checks."${system}".tests = ...`). This will make the whole local testing/CI run with `nix flake check`.
- I'm not entirely convinced by the default dev build mode. My assumption would be that the build is a prod build, with a user dropping into a dev shell (`nix develop`) if they need anything special. Worth discussing.
- Having `bpftrace` in the `apps` part of the output would be nice, allowing `nix run`. This wouldn't interact very well with `sudo` though. Again, worth discussing.
- I don't think we need `devShells.default` at all in its current form, as it's effectively set to the default (`+runTests`, which can be covered with `nix flake check`). Not sure how other projects have handled this though.
- I think there are some dependencies missing as `cmake -B build; cd build; make` didn't work on my NixOS system in `nix develop` (a really clean external path). This is minutia though.

The output schema is [here](https://nixos.wiki/wiki/Flakes#Output_schema).

Great starting point, and thanks for offering up this RFC!

flake.nix
RyanGibb: `s/x86_64-linux/${system}/` should be fine (the quotes aren't necessary)

README.md
danobi: Ack. Best to let the nix download page sell the mechanism.

flake.nix
danobi: Yes, I think we need these. Based on the discussion in #2334 and #2322 there is consensus that bpftrace should have a strong dependency on libbpf, especially to ease development and migration away from bcc. It is too difficult to support every libbpf under the sun and perform possibly the largest refactor this project has seen. During release time we will simply bump the required version to the one we developed against (which solve the packaging problem too).

flake.nix
danobi: Ack. Good catch.

flake.nix
danobi: I think nix only supports aarch and x86 on linux and osx, right? We do support quite a few architectures but none 32 bit. And osx has no bpf support.

So that limits nix support to aarch64-linux and x86-64-linux, right? Or am I misunderstanding something?

danobi:
Hi @JakeHillion, @RyanGibb , thank you for the reviews! It is much appreciated b/c I am quite new to nix.

> Thank you for doing this, more Nix flake adoption in upstream repos is great to see! I commented on some more specific parts but generally.
> 
>     * Rather than including the tests in the binary output they should really be included as flake checks (`outputs.checks."${system}".tests = ...`). This will make the whole local testing/CI run with `nix flake check`.

Thanks for the tip. I will look into that. Looks like it's uncovering a few (maybe bad) assumptions already b/c it's running in a hermetic environment.

>     * I'm not entirely convinced by the default dev build mode. My assumption would be that the build is a prod build, with a user dropping into a dev shell (`nix develop`) if they need anything special. Worth discussing.

So the goal of this flake is to address some the pain points in CI we have. I'd like to start off with fixing development pain first and then if possible, make it a flake suitable for downstream users. Unless there's an easy way to satisfy both that I'm missing (and I probably am), I'd prefer to stick with dev in mind first.

>     * Having `bpftrace` in the `apps` part of the output would be nice, allowing `nix run`. This wouldn't interact very well with `sudo` though. Again, worth discussing.

Ack. Might as well add it even if really only --help works.

>     * I don't think we need `devShells.default` at all in its current form, as it's effectively set to the default (`+runTests`, which can be covered with `nix flake check`). Not sure how other projects have handled this though.

Yeah makes sense. Will address it when I fix `nix flake check`.

>     * I think there are some dependencies missing as `cmake -B build; cd build; make` didn't work on my NixOS system in `nix develop` (a really clean external path). This is minutia though.

Hmm, that's weird. It's the same deps as `nix build` right? And that seems to work ok.

[...]

danobi:
This patch makes it so tests are run (but they fail for various reasons): https://pastes.dxuuu.xyz/f8tnu6

Running into this issue: 
```
$ nix flake check --allow-new-privileges
warning: Git tree '/home/dxu/dev/bpftrace' is dirty
error: builder for '/nix/store/b2px0hqb5i8l3m52jii5za99rg7bmy0z-bpftrace.drv' failed with exit code 1;
       last 10 log lines:
       > [192/196] Linking CXX static library src/ast/libast.a
       > [193/196] Linking CXX static library src/libbpftrace.a
       > [194/196] Linking CXX executable src/bpftrace
       > /nix/store/9n9gjvzci75gp2sh1c4rh626dhizqynl-binutils-2.39/bin/ld: warning: libbpf.so.0, needed by /nix/store/qbrgk50h4z64vxfdx5z7a31knwszsd9n-bcc-0.27.0/lib/libbcc_bpf.so, may conflict with libbpf.so.1
       > [195/196] Building CXX object tests/CMakeFiles/bpftrace_test.dir/semantic_analyser.cpp.o
       > [196/196] Linking CXX executable tests/bpftrace_test
       > /nix/store/9n9gjvzci75gp2sh1c4rh626dhizqynl-binutils-2.39/bin/ld: warning: libbpf.so.0, needed by /nix/store/qbrgk50h4z64vxfdx5z7a31knwszsd9n-bcc-0.27.0/lib/libbcc_bpf.so, may conflict with libbpf.so.1
       > running tests
       > sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
       > sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.
       For full logs, run 'nix-store -l /nix/store/b2px0hqb5i8l3m52jii5za99rg7bmy0z-bpftrace.drv'.
```

Can't seem to get sudo to work even with `--allow-new-privileges`

flake.nix
JakeHillion: I think you would still have precise control as there is control over when the lock and hence `nixpkgs` dependency is updated, but this definitely makes it more explicit. Fine by me.

JakeHillion (approved):
Thanks for the great discussion! With this enabled in the CI and the tests fixed this looks good to me, and gives us a fantastic base to add new Nix features in the future.

danobi:
> I think the best solution to this long term is to run the tests in VMs using https://nixos.org/manual/nixos/stable/index.html#sec-nixos-tests. However this has a bit of a conflict of interest with your existing https://github.com/iovisor/bpftrace/pull/2607. As far as I can tell, it would be possible to achieve the same thing by pinning kernels in the NixOS testing framework.

Looks neat. Will take a look. That looks like a great option when the entire workflow has already been onboarded to nix.

That being said, there's 1 big blocker in the way of all the VM testing stuff: nested virtualization support. Right now it took over 6 hours to run the runtime tests (and it still didn't finish) in a VM in standard GHA. GH supports large runners with nested virt now but we need a paid plan.

danobi:
I asked around in the nix matrix room and thought about it overnight. Basically they're saying that flake checks cannot run anything that requires root. Disappointing but understandable.

In light of this, I think we should scale back the ambitions of the nix flake build a bit. I've pushed changes to make it more of a dependency manager rather than an all-in-one CI thing.  This will still immediately solve some developer pain. We'll just have multiple steps in the CI that runs stuff inside nix developer shells rather than just with `nix flake check`.

I've also incorporated the feedback about making the flake geared towards downstream users. The cmake flags and such will be manually chosen in the developer shell now.

danobi:
@viktormalik @fbs @ajor any thoughts? At the very least this would provide an easy way to get started with developing on bpftrace.

I am planning on replacing `ci.yml` workflow using this nix flake (to get better local reproducibility). I am also hoping we can replace the embedded builds as well with the help of https://github.com/rsms/llvmbox . But I am less certain about embedded builds.

I have no strong opinion on whether we should wait for ci.yml replacement work to be done before merging this, but like I mentioned, the flake here is enough for smoother onboarding of new developers.

JakeHillion:
Ack these changes, I think this looks a lot nicer now. Agreed on the scaled back ambitions too, as everybody gets a clean development environment with consistent dependencies and can work from there.

With luck fixing the sudo issues can be solved with some form of vm testing, but if this goes in as is it'll be an excellent base for both new developers and the CI - extensions can come later.

flake.nix
ajor: bpftrace does support 32-bit ARM now, FWIW

ajor (commented):
I know no Nix, but I like the sound of it. I'd say to go ahead and merge this before fiddling with the CI - what we've got here is useable right now.

I think the Docker-based build system will still be useful for confirming that bpftrace builds on various Linux distros, so wouldn't get rid of it entirely. It seems fair enough to move the general CI builds and the LLVM matrix across to Nix though.

danobi:
> I think the Docker-based build system will still be useful for confirming that bpftrace builds on various Linux distros, so wouldn't get rid of it entirely. It seems fair enough to move the general CI builds and the LLVM matrix across to Nix though.

I agree it's useful in principle. But I'd like to point out in practice that `ci.yml` only builds with ubuntu 20.04 and `embedded.yml` builds with ubuntu 18.04, 20.04, and alpine 3.12. So at minimum reworking `ci.yml` is status-quo.

Long term I'd like to revisit the embedded llvm builds. We pay a heavy maintenance cost for that. I am hoping (praying perhaps) that we can externalize that cost to llvmbox. But that will come later, if at all.

flake.nix
```diff
@@ -0,0 +1,118 @@
...
+                ];
+              };
+        in
+        {
+          # Set formatter for `nix fmt` command
+          formatter = pkgs.nixpkgs-fmt;
+
+          # Define package set
+          packages = {
+            # Default package is bpftrace with llvm 15
```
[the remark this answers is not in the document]
danobi: Ack.

PR #2599, opened 2023-05-16
Support identifying btf type for btf-based tracing

chengshuyi:
k(ret)func and iter are btf-based tracing, which can detect whether dereferencing is safe through btf, so dereferencing does not need to call the helper function bpf_probe_read. This patch mainly modifies the semantic and codegen modules

semantic:
- Handle the ctx of iter, mark is_btftype as true
- When dealing with ArrayAccess, Unop, FieldAccess, inherit is_btftype attribute

codegen:
- If it is a string type, use CREATE_MEMCPY directly to load the data
- For other types, load data through CreateDatastructElemLoad instead of CreateProbeRead

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1846,6 +1846,7 @@ void CodegenLLVM::visit(FieldAccess &acc)
   bool is_tparg = type.is_tparg;
   bool is_internal = type.is_internal;
   bool is_funcarg = type.is_funcarg;
+  bool is_btftype = type.is_btftype;
   assert(type.IsRecordTy() || type.IsTupleTy());
 
   if (type.is_funcarg)
@@ -1887,6 +1888,7 @@ void CodegenLLVM::visit(FieldAccess &acc)
   type.is_tparg = is_tparg;
   type.is_internal = is_internal;
   type.is_funcarg = is_funcarg;
+  type.is_btftype = is_btftype;
   // Restore the addrspace info
   // struct MyStruct { const int* a; };  $s = (struct MyStruct *)arg0;  $s->a
   type.SetAS(addrspace);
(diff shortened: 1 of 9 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -3561,6 +3570,14 @@ void CodegenLLVM::probereadDatastructElem(Value *src_data,
...
       }
     }
+    else if (data_type.is_btftype)
+    {
+      expr_ = b_.CreateDatastructElemLoad(
+          elem_type,
+          b_.CreateIntToPtr(src, dst_type->getPointerTo()),
+          true,
+          data_type.GetAS());
+    }
```
viktormalik, line 3580: This is exactly the same call as on lines 3545-3549, can we merge them?

chengshuyi: Ok.

chengshuyi:
> E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/j/javascript-common/javascript-common_11_all.deb  Could not connect to archive.ubuntu.com:80 (91.189.91.38), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.36), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.91.39), connection timed out Could not connect to archive.ubuntu.com:80 (185.125.190.39), connection timed out [IP: 185.125.190.39 80]

It seems ubuntu package server is not stable which causes ci fail.

viktormalik:
I'm surprised that this didn't break any codegen test. It looks like we don't have a test for kfunc+dereference. Maybe this is a good time to add one?

chengshuyi:
> I'm surprised that this didn't break any codegen test. It looks like we don't have a test for kfunc+dereference. Maybe this is a good time to add one?

Ok, will do

danobi (approved):
great patch! lgtm

viktormalik (approved):
Looks good, thanks!

PR #2604, opened 2023-05-16
Adding feature dependency in runtime tests

dubeyabhishek:


tests/runtime/call
```diff
@@ -367,7 +367,7 @@ TIMEOUT 1
 NAME path_in_unsupported_kfunc
 PROG kfunc:vfs_read { print(path(args->file->f_path)); }
 EXPECT \nstdin:1:18-48: ERROR: helper bpf_d_path not supported in probe\n
-REQUIRES_FEATURE dpath
+REQUIRES_FEATURE dpath kfunc
```
viktormalik, line 370: We prefer this style:
```suggestion
REQUIRES_FEATURE dpath
REQUIRES_FEATURE kfunc
```
It makes it easier to spot that the test depends on multiple features.

viktormalik (commented):
Thanks for cleaning this!

I believe that support for kfuncs presumes support for BTF, so we can leave only `REQUIRES_FEATURE kfunc` for all tests containing kfuncs.

tests/runtime/call
dubeyabhishek: Ack.

danobi (approved):

viktormalik (approved):

PR #2605, opened 2023-05-16
Fix kprobe_offset_module test with correct offset for ppc64

dubeyabhishek:
The instruction size in ppc64 is 4 bytes, but the entry-point to the local functions is located at an offset of 8 bytes. Therefore, a kprobe can be attached at offsets that are multiples of 4 bytes beyond the initial 8-byte offset.

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -104,17 +104,28 @@ NAME kprobe_offset_module
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x5 { printf("hit\n"); exit(); }'
 AFTER /usr/sbin/nft add table bpftrace
 EXPECT hit
+ARCH x86_64|aarch64
 TIMEOUT 5
 REQUIRES lsmod | grep '^nf_tables'
 REQUIRES /usr/sbin/nft --help
-CLEANUP nft delete table bpftrace
+CLEANUP /usr/sbin/nft delete table bpftrace
+
+# Local entry point to nft_trans_alloc_gfp is located at offset of 8 bytes in ppc64.
+NAME kprobe_offset_module
+RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x8 { printf("hit\n"); exit(); }'
+AFTER /usr/sbin/nft add table bpftrace
+EXPECT hit
+ARCH ppc64|ppc64le
+TIMEOUT 5
+REQUIRES lsmod | grep '^nf_tables'
+REQUIRES /usr/sbin/nft --help
+CLEANUP /usr/sbin/nft delete table bpftrace
 
 NAME kprobe_offset_module_error
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x1 { printf("hit\n"); exit(); }'
 EXPECT Possible attachment attempt in the middle of an instruction, try a different offset.
 TIMEOUT 5
 REQUIRES lsmod | grep '^nf_tables'
-REQUIRES /usr/sbin/nft --help
 WILL_FAIL
 
 NAME kretprobe


tests/runtime/probe
```diff
@@ -100,6 +100,17 @@ WILL_FAIL
 NAME kprobe_offset_module
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x5 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
 EXPECT hit
+ARCH x86_64|aarch64
+TIMEOUT 5
+REQUIRES lsmod | grep '^nf_tables'
+REQUIRES /usr/sbin/nft --help
+AFTER nft delete table bpftrace
```
danobi, line 107: I think `AFTER` is wrong here. See: https://github.com/iovisor/bpftrace/blob/31d7a35816fa4993ae2f1208153ce38c632be567/tests/runtime/engine/runner.py#L245

It's run right after probe attachment. I think a `CLEANUP` directive is needed

dubeyabhishek: CLEANUP directive PR sent.

dubeyabhishek: @viktormalik Working on the issue.

tests/runtime/probe
```diff
@@ -100,18 +100,28 @@ WILL_FAIL
...
+REQUIRES /usr/sbin/nft --help
+CLEANUP nft delete table bpftrace
 
 NAME kprobe_offset_module_error
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x1 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
 EXPECT Possible attachment attempt in the middle of an instruction, try a different offset.
 TIMEOUT 5
 REQUIRES lsmod | grep '^nf_tables'
 REQUIRES /usr/sbin/nft --help
-AFTER nft delete table bpftrace
```
danobi, line 114: Should be replaced with a CLEANUP here, right?

dubeyabhishek: Here we expect the failure due to incorrect offset in between instruction. I was not sure of CLEANUP error getting reported in a test where we expect other reason.

danobi: Ah, that is a good point. How about deleting the above `-c '/usr/sbin/nft add table bpftrace'` in the test just so we're certain there's nothing to clean up

viktormalik:
It looks like there could be a race condition in the `kprobe_offset_module` test - it has failed in one CI run (the probe was never hit) but was fixed by a restart. It's strange that we never hit that in the past as that part shouldn't be changed in this PR.

tests/runtime/probe
```diff
@@ -100,18 +100,27 @@ WILL_FAIL
 NAME kprobe_offset_module
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x5 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
 EXPECT hit
+ARCH x86_64|aarch64
 TIMEOUT 5
 REQUIRES lsmod | grep '^nf_tables'
 REQUIRES /usr/sbin/nft --help
-AFTER nft delete table bpftrace
+CLEANUP nft delete table bpftrace
```
viktormalik, line 107:
```suggestion
CLEANUP /usr/sbin/nft delete table bpftrace
```

tests/runtime/probe
```diff
@@ -100,18 +100,27 @@ WILL_FAIL
...
+
+# Local entry point to nft_trans_alloc_gfp is located at offset of 8 bytes in ppc64.
+NAME kprobe_offset_module
+RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x8 { printf("hit\n"); exit(); }' -c '/usr/sbin/nft add table bpftrace'
+EXPECT hit
+ARCH ppc64|ppc64le
+TIMEOUT 5
+REQUIRES lsmod | grep '^nf_tables'
+REQUIRES /usr/sbin/nft --help
+CLEANUP nft delete table bpftrace
```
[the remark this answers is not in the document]
dubeyabhishek: @viktormalik please restart the checks. Again its kprobe_offset_module test.

tests/runtime/probe
viktormalik: Still valid.

viktormalik:
> > @viktormalik please restart the checks. Again its kprobe_offset_module test.
> 
> It seems like the PR is introducing a race condition into the test for some reason. I'll try to have a look, could you also try to debug?

So, I can reproduce this locally on master, so it's not related to this PR.

dubeyabhishek:
> So, the problem is not in this PR but in the way `nft` commands are executed in tests (but this PR made the possibility to hit the race condition much higher). I submitted a fix in #2622, I suggest to wait until it's merged and then rebase this work on top of this.

Sure.

viktormalik:
#2622 is merged, please rebase, thanks!

viktormalik (approved):

PR #2606, opened 2023-05-17
Fixing absolute address test issue across platforms

dubeyabhishek:
mmap to the address 0x10000000 works fine on few arch, i.e x86_64. However, on ppc64, this mmap fails on the hinted address, leading to an untriggered watchpoint on 0x10000000. This patch dynamically replaces the allocated address from mmap, causing the watchpoint to be triggered on write at the correct mapped address.

### DIFF
--- a/tests/runtime/watchpoint
+++ b/tests/runtime/watchpoint
@@ -1,7 +1,9 @@
 NAME watchpoint - absolute address
-RUN {{BPFTRACE}} -e 'watchpoint:0x10000000:8:w { printf("hit!\n"); exit() }' -c ./testprogs/watchpoint
+BEFORE ./testprogs/watchpoint
+RUN {{BPFTRACE}} -e "watchpoint:$(awk '{print $1}' /tmp/watchpoint_mem):8:w { printf(\"hit!\n\"); exit() }" -p {{BEFORE_PID}}
 EXPECT hit!
 ARCH aarch64|ppc64|ppc64le|x86_64
+CLEANUP rm -f /tmp/watchpoint_mem
 TIMEOUT 5
 
 NAME kwatchpoint - knl absolute address
--- a/tests/testprogs/watchpoint.c
+++ b/tests/testprogs/watchpoint.c
@@ -1,6 +1,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 int main() {
@@ -17,9 +18,22 @@ int main() {
     return 1;
   }
 
+  mode_t old_umask = umask(S_IWGRP | S_IROTH | S_IWOTH);
+  FILE* addr_fp = fopen("/tmp/watchpoint_mem", "w");
+
+  if (!addr_fp)
+    perror("failed to open file in /tmp");
+
+  fprintf(addr_fp, "%p", addr);
+  fclose(addr_fp);
 
   uint8_t i = 0;
-  while (1) {
+  while (i < 10)
+  {
     *((volatile uint8_t*)addr) = i++;
+    // 250ms*10 sleep, enough for watchpoint trigger
+    usleep(250 * 1000);
   }
+
+  umask(old_umask);
 }


tests/runtime/watchpoint
```diff
@@ -1,7 +1,8 @@
 NAME watchpoint - absolute address
-RUN {{BPFTRACE}} -e 'watchpoint:0x10000000:8:w { printf("hit!\n"); exit() }' -c ./testprogs/watchpoint
+RUN install -m 644 /dev/null /tmp/watchpoint_mem; ./testprogs/watchpoint & > /dev/null; {{BPFTRACE}} -e "watchpoint:$(awk '{print $1}' /tmp/watchpoint_mem):8:w { printf(\"hit!\n\"); exit() }"
 EXPECT hit!
 ARCH aarch64|ppc64|ppc64le|x86_64
+AFTER /bin/bash -c "rm -f /tmp/watchpoint_mem"
```
danobi, line 5: Unfortunately the `AFTER` directive does not do what you think it does here. It's not a cleanup directive. Rather, it runs right after the `RUN` command is spawned. So using it like this is racy. Feel free to add a `CLEANUP` directive to the test runner -- shouldn't be difficult.

tests/runtime/watchpoint
```diff
@@ -1,7 +1,8 @@
 NAME watchpoint - absolute address
-RUN {{BPFTRACE}} -e 'watchpoint:0x10000000:8:w { printf("hit!\n"); exit() }' -c ./testprogs/watchpoint
+RUN install -m 644 /dev/null /tmp/watchpoint_mem; ./testprogs/watchpoint & > /dev/null; {{BPFTRACE}} -e "watchpoint:$(awk '{print $1}' /tmp/watchpoint_mem):8:w { printf(\"hit!\n\"); exit() }"
```
danobi, line 2: Hmm, by removing the `-c` param, this turns it into a kernel watchpoint, no?

Docs say:
> In the first form, an absolute address is monitored. If a pid (-p) or a command (-c) is provided, bpftrace takes the address as a userspace address and monitors the appropriate process. If not, bpftrace takes the address as a kernel space address.

tests/testprogs/watchpoint.c
```diff
@@ -17,9 +17,19 @@ int main() {
     return 1;
   }
 
+  FILE* addr_fp = fopen("/tmp/watchpoint_mem", "w");
+
+  if (!addr_fp)
+    perror("failed to open file in /tmp");
+
+  fprintf(addr_fp, "%p", addr);
+  fclose(addr_fp);
```
danobi, line 26: How about making this testprog print the address to stdout? Then run it in a `BEFORE` clause and cat it to a file, ie:

```
BEFORE ./testprogs/watchpoint > /tmp/watchpoint_mem &
```

Then in the `RUN` directive you can tell bpftrace about the testprog using `-p {{BEFORE_PID}}`.

danobi (changes requested):

tests/testprogs/watchpoint.c
dubeyabhishek: Above command in BEFORE get stuck and timeout happens. Other issue is, redirection of o/p must be handled in subprocess.Popen for BEFORE directive(changes needed).

If everything falls in place, still the race between bpftrace invocation in RUN and flushing the address to /tmp/watchpoint_mem exists.

dubeyabhishek: @danobi please review new changes

tests/runtime/watchpoint
```diff
@@ -1,7 +1,8 @@
 NAME watchpoint - absolute address
-RUN {{BPFTRACE}} -e 'watchpoint:0x10000000:8:w { printf("hit!\n"); exit() }' -c ./testprogs/watchpoint
+RUN ./testprogs/watchpoint & > /dev/null; {{BPFTRACE}} -e "watchpoint:$(awk '{print $1}' /tmp/watchpoint_mem):8:w { printf(\"hit!\n\"); exit() }"
 EXPECT hit!
 ARCH aarch64|ppc64|ppc64le|x86_64
+AFTER /bin/bash -c "rm -f /tmp/watchpoint_mem"
```
viktormalik, line 5: Shouldn't be
```suggestion
CLEANUP /bin/bash -c "rm -f /tmp/watchpoint_mem"
```

tests/runtime/watchpoint
viktormalik: This still applies, right? It seems to me that it would be better to run `testprogs/watchpoint` from the `BEFORE` clause and then use `-p {{BEFORE_PID}}` not to transform this into a kernel watchpoint.

dubeyabhishek: My bad, I missed that. Now implemented.

tests/runtime/watchpoint
dubeyabhishek: Ack

tests/runtime/watchpoint
```diff
@@ -1,7 +1,9 @@
 NAME watchpoint - absolute address
-RUN {{BPFTRACE}} -e 'watchpoint:0x10000000:8:w { printf("hit!\n"); exit() }' -c ./testprogs/watchpoint
+BEFORE ./testprogs/watchpoint
+RUN {{BPFTRACE}} -e "watchpoint:$(awk '{print $1}' /tmp/watchpoint_mem):8:w { printf(\"hit!\n\"); exit() }" -p {{BEFORE_PID}}
 EXPECT hit!
 ARCH aarch64|ppc64|ppc64le|x86_64
+CLEANUP /bin/bash -c "rm -f /tmp/watchpoint_mem"
```
danobi, line 6: I think CLEANUP already runs in a shell, no? So I think you can drop the `/bin/bash -c ...` bit

danobi (commented):
lgtm overall.

viktormalik (approved):

dubeyabhishek:
@danobi @viktormalik can we close this change?

danobi (approved):

PR #2607, opened 2023-05-17
ci: Run tests under controlled kernels

danobi:
This PR teaches CI how to run tests under different and tightly controlled
kernels. Under the hood it uses `vmtest`.

More details about vmtest can be found here:

* https://github.com/danobi/vmtest
* https://dxuuu.xyz/vmtest.html

Note we use kernels that are built and provided by vmtest. See:

* https://github.com/danobi/vmtest/releases/tag/test_assets
* https://github.com/danobi/vmtest/blob/master/.github/workflows/kernels.yml

While I'm not making any official guarantees about provided kernel stability,
I do promise not to break bpftrace CI if at all possible.

##### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,17 @@ jobs:
           TOOLS_TEST_OLDVERSION: biosnoop.bt tcpdrop.bt
           BASE: focal
           VENDOR_GTEST: ON
+        - NAME: Latest LLVM + Upstream kernel
+          TYPE: Release
+          LLVM_VERSION: 16
+          RUN_RUNTIME_TESTS: 1
+          RUN_ALL_TESTS: 0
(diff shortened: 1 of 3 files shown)


danobi:
Current status and thoughts:

* It looks like vmtest functions correct. But lack of nested virt makes the runtime tests take way too long (>6hr)
* To address the above, my current plan is to make only the runtime tests that have `FEATURE` gates run. This strikes a balance between speed and effectiveness.
* vmtest is also hamstringed by the docker usage in CI. See: https://github.com/danobi/vmtest/blob/master/docs/faqs.md#how-do-i-run-docker-inside-vmtest . This can/will be addressed by the nix-ified CI.

danobi:
I'm traveling and also a bit under the weather. Will try to come back to this ASAP

danobi:
Outdated attempt. Need to integrate with ci.py now

PR #2608, opened 2023-05-18
testdata: Rewrite test header generation code

danobi:
Before, the old generation could would write temporary data in the source directory. Technically this works, but it was causing issues with vmtest, as vmtest was only mounting the build directory read/write. This rewrite makes it so all temporary file generation occurs in the build directory.

Furthermore, this rewrite correctly declares all the input/output dependencies so the test headers are only regenerated when absolutely necessary.

##### Checklist


viktormalik (approved):
Thanks! This is now much cleaner and more idiomatic.

PR #2613, opened 2023-05-20
Fixing runtime test filter option: --filter

dubeyabhishek:
This patch handles following observed issues:

1. Inconsistency with test names, some have double qoutes('"'), whereas others don't. Extra pre-processing is required in runtime bash script for such inconsistent names, before they are supplied as args to python script.

2. Test name nomenclature for --test-filter option is missing. Need to manually figure out from code.

### DIFF
--- a/tests/runtime-tests.sh
+++ b/tests/runtime-tests.sh
@@ -17,4 +17,4 @@ echo "bpftrace --info:"
 echo "===================="
 "${BPFTRACE_RUNTIME_TEST_EXECUTABLE}/bpftrace" --info
 
-python3 runtime/engine/main.py $@
+python3 runtime/engine/main.py "$@"
--- a/tests/runtime/engine/main.py
+++ b/tests/runtime/engine/main.py
@@ -72,11 +72,11 @@ def main(test_filter, run_aot_tests):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Runtime tests for bpftrace.')
-    parser.add_argument('--filter', dest='tests_filter',
-                        help='filter runtime tests')
+    parser.add_argument('--filter', type=str, dest='test_filter',
+                        help='Run only specified runtime test. Format should be "<test feature/test group>.<testcase name>"')
     parser.add_argument('--run-aot-tests', action='store_true',
                         help='Run ahead-of-time compilation tests. Note this would roughly double test time.')
 
     args = parser.parse_args()
 
-    main(args.tests_filter, args.run_aot_tests)
+    main(args.test_filter, args.run_aot_tests)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
           TYPE: Debug
           LLVM_VERSION: 10
           RUN_ALL_TESTS: 1
-          RUNTIME_TEST_DISABLE: probe.kprobe_offset_fail_size,usdt."usdt probes - file based semaphore activation multi process"
+          RUNTIME_TEST_DISABLE: probe.kprobe_offset_fail_size,usdt.usdt probes - file based semaphore activation multi process
           TOOLS_TEST_OLDVERSION: biosnoop.bt tcpdrop.bt
           BASE: focal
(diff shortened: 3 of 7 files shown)


tests/runtime-tests.sh
```diff
@@ -17,4 +17,4 @@ echo "bpftrace --info:"
 echo "===================="
 "${BPFTRACE_RUNTIME_TEST_EXECUTABLE}/bpftrace" --info
 
-python3 runtime/engine/main.py $@
+python3 runtime/engine/main.py "$@"
```
danobi, line 20: I think we actually want the word splitting here. Otherwise if there were multiple args then main would incorrectly only see 1 parameter, right?

danobi (commented):
Nice cleanup. Pls take a look at the ci failures though

tests/runtime-tests.sh
dubeyabhishek: Without double quotes, cmd like : ./tests/runtime-tests.sh --test-filter "probe.kprobe_wildcard probe builtin" throws error:

main.py: error: unrecognized arguments: probe builtin

Without quotes works fine for test names which doesn't contain "space" or "dash", otherwise fail due to command split at space/dash.

With quotes, multiple args passed as single string to python works fine.

dxuuu: Ah, I see. "$@" is special cased: https://stackoverflow.com/questions/3811345/how-to-pass-all-arguments-passed-to-my-bash-script-to-a-function-of-mine

danobi: Ah, I see. "$@" is special cased: https://stackoverflow.com/questions/3811345/how-to-pass-all-arguments-passed-to-my-bash-script-to-a-function-of-mine

danobi (commented):
CI still fails. Perhaps you need to update the disabled test strings in this file? https://github.com/iovisor/bpftrace/blob/fb1e082a4f9f2b756f813156b58ee2a5af12747e/.github/workflows/embedded.yml#L18

viktormalik:
> CI still fails. Perhaps you need to update the disabled test strings in this file?
> 
> https://github.com/iovisor/bpftrace/blob/fb1e082a4f9f2b756f813156b58ee2a5af12747e/.github/workflows/embedded.yml#L18

Agreed, you have to remove the quotes.

viktormalik:
> > file based semaphore activation multi process
> 
> This test is actually skipped. So I was not suspecting same. Will try now.

Not only here, you also have to remove the quotes from the alpine builds (those are the ones failing b/c the tests are not properly ignored).

dubeyabhishek:
> > > file based semaphore activation multi process
> > 
> > 
> > This test is actually skipped. So I was not suspecting same. Will try now.
> 
> Not only here, you also have to remove the quotes from the alpine builds (those are the ones failing b/c the tests are not properly ignored).

@viktormalik Got the idea. Thanks!

tests/runtime/engine/main.py
```diff
@@ -72,11 +72,11 @@ def main(test_filter, run_aot_tests):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Runtime tests for bpftrace.')
-    parser.add_argument('--filter', dest='tests_filter',
-                        help='filter runtime tests')
+    parser.add_argument('--test-filter', type=str,
```
viktormalik, line 75: I don't see a reason for renaming the option and it will require to rewrite existing scripts that people may have, so I'd prefer sticking with the old name.

dubeyabhishek: Correct. Actually we support filtering a single test right now, so thought of removing 's' from option.

viktormalik: Feel free to rename the internal variable to `test_filter` but don't change the user-facing CLI please.

viktormalik: To make it clear: I prefer sticking with the original CLI option name, i.e. `--filter`.

viktormalik (approved):
Very useful changes, thanks!

danobi (approved):

PR #2615, opened 2023-05-22
resources: generate std::string_view instead of c strings and lengths

JakeHillion:
This PR makes two changes:
- Informs CMake that it should reconfigure if any of the headers turned into strings are changed, meaning changes will be picked up on the next build.
- Uses C++ `std::string_view` rather than a `c_str` and a length parameter. This stores the same data but in a neatly packaged way.

One subtlety here is `std::string_view::data()` does not guarantee the string is null-terminated, and I think in this case it isn't. However the Clang docs describe the argument `.Contents` as "A buffer containing the unsaved contents of this file.", making no reference to it being null-terminated, plus the tests pass. It is possible to use `std::string` and get the `c_str` method, but it uses more memory and this seems a neater use of `std::string_view`.

##### Checklist


ajor (approved):
Nice!

PR #2617, opened 2023-05-22
Fixing args-type in subprocess.Popen against f4bc83e

dubeyabhishek:
As per subprocess.Popen docs: If shell=True, it is recommended to pass args as string rather than as
a sequence.


danobi (approved):
thanks!

Issue #2512, opened 2023-03-01
Builtin for getting real time in nanoseconds


viktormalik:
### Is your feature request related to a problem? Please describe.

As was suggested in discussion #2486, there may be applications that require bpftrace programs to get the current time in nanoseconds. Since kernel 6.2, there is the [bpf_ktime_get_tai_ns helper](https://elixir.bootlin.com/linux/v6.2/source/kernel/bpf/helpers.c#L204) that allows BPF programs to access the TAI clock.

### Describe the solution you'd like

A new builtin variable, e.g. `nsecs_tai` or simply `tai` that would call the `bpf_ktime_get_tai_ns` helper (if it is available on the current kernel) and return the TAI in nanoseconds. Then, `strftime` can be used to create a formatted timestamp.

chengshuyi:
Hi, i'm learning the bpftrace code, I can add this function if you don't mind.

viktormalik:
Please do! Thanks :-)

ajor:
With this new TAI clock helper, is there any reason anyone would want to use the old nsecs-since-boot timestamp? Can we just switch the "nsecs" builtin across to using this new clock?

chengshuyi:
The kernel currently provides four methods for obtaining timestamps, namely bpf_ktime_get_ns, bpf_ktime_get_boot_ns, bpf_ktime_get_coarse_ns and bpf_ktime_get_tai_ns. They can be roughly divided into two categories:
1. bpf_ktime_get_ns, bpf_ktime_get_boot_ns and bpf_ktime_get_coarse_ns are  the time elapsed since system boot》
2. bpf_ktime_get_tai_ns is to get the time of CLOCK_TAI.

So, I think it may be necessary to provide methods to get these two different types of timestamps.

Thanks!

ajor:
What would the use cases be in bpftrace for exposing both these categories of timestamps?

chengshuyi:
I don't have a specific use case, and I don't see any special requirements for the time elapsed since system boot from the issues and pr of bpftrace. On the contrary, #1471  and #2486 have demand for tai time. Among them, #1471 uses the "triple vdso sandwich" method to obtain approximate tai time.

viktormalik:
> With this new TAI clock helper, is there any reason anyone would want to use the old nsecs-since-boot timestamp? Can we just switch the "nsecs" builtin across to using this new clock?
> 
[...]

PR #2619, opened 2023-05-23
Introduce builtin nsecs_tai which is nanoseconds of CLOCK_TAI 

chengshuyi:
introduce new builtin nsecs_tai，on kernels that version >= 6.1, get the nanoseconds through `ktime_get_tai_ns`, otherwise use the "triple vdso sandwich" method to get the nanoseconds of CLOCK_TAI, and its accuracy is about tens of nanoseconds.A typical case is to use necs_tai to obtain the delay of network packets between machines.

issues: #2512 

##### Checklist

- [] User-visible and non-trivial changes updated in `CHANGELOG.md`

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1146,11 +1146,17 @@ void CodegenLLVM::visit(Call &call)
         b_.GetIntSameSize(strftime_id_, elements.at(0)),
         b_.CreateGEP(strftime_struct, buf, { b_.getInt64(0), b_.getInt32(0) }));
     strftime_id_++;
+    b_.CreateStore(
+        b_.GetIntSameSize(
+            static_cast<std::underlying_type<TimestampMode>::type>(
+                call.type.ts_mode),
+            elements.at(1)),
+        b_.CreateGEP(strftime_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
     Expression *arg = call.vargs->at(1);
     auto scoped_del = accept(arg);
     b_.CreateStore(
         expr_,
-        b_.CreateGEP(strftime_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
+        b_.CreateGEP(strftime_struct, buf, { b_.getInt64(0), b_.getInt32(2) }));
     expr_ = buf;
   }
   else if (call.func == "kstack" || call.func == "ustack")
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1319,10 +1319,17 @@ void CodegenLLVM::visit(Call &call)
   }
   else if (call.func == "nsecs")
   {
-    if (call.type.ts_mode == TimestampMode::boot)
+    if (call.type.ts_mode == TimestampMode::boot ||
+        call.type.ts_mode == TimestampMode::sw_tai)
     {
       expr_ = b_.CreateGetNs(bpftrace_.feature_->has_helper_ktime_get_boot_ns(),
                              call.loc);
+      if (call.type.ts_mode == TimestampMode::sw_tai)
+      {
+        uint64_t delta = bpftrace_.delta_taitime_->tv_sec * 1e9 +
+                         bpftrace_.delta_taitime_->tv_nsec;
+        expr_ = b_.CreateAdd(expr_, b_.getInt64(delta));
+      }
     }
     else if (call.type.ts_mode == TimestampMode::tai)
     {
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1311,6 +1311,14 @@ void CodegenLLVM::visit(Call &call)
     expr_ = ret;
     skb_output_id_++;
   }
+  else if (call.func == "nsecs")
+  {
+    if (call.type.ts_mode == TimestampMode::boot)
+    {
+      expr_ = b_.CreateGetNs(bpftrace_.feature_->has_helper_ktime_get_boot_ns(),
+                             call.loc);
+    }
+  }
   else
   {
     LOG(FATAL) << "missing codegen for function \"" << call.func << "\"";
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1324,6 +1324,10 @@ void CodegenLLVM::visit(Call &call)
       expr_ = b_.CreateGetNs(bpftrace_.feature_->has_helper_ktime_get_boot_ns(),
                              call.loc);
     }
+    else if (call.type.ts_mode == TimestampMode::tai)
+    {
+      expr_ = b_.CreateGetTaiNs(call.loc);
+    }
   }
   else
   {
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -144,6 +144,10 @@ void SemanticAnalyser::visit(Identifier &identifier)
     {
       identifier.type.ts_mode = TimestampMode::tai;
     }
+    else if (identifier.ident == "sw_tai")
+    {
+      identifier.type.ts_mode = TimestampMode::sw_tai;
+    }
     else
     {
       goto err;
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -1113,9 +1113,13 @@ void SemanticAnalyser::visit(Call &call)
   else if (call.func == "strftime")
   {
     call.type = CreateTimestamp();
-    check_varargs(call, 2, 2) && is_final_pass() &&
+    if (check_varargs(call, 2, 2) && is_final_pass() &&
         check_arg(call, Type::string, 0, true) &&
-        check_arg(call, Type::integer, 1, false);
+        check_arg(call, Type::integer, 1, false))
+    {
+      auto &arg = *call.vargs->at(1);
+      call.type.ts_mode = arg.type.ts_mode;
+    }
   }
   else if (call.func == "kstack") {
     check_stack_call(call, true);
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -140,6 +140,10 @@ void SemanticAnalyser::visit(Identifier &identifier)
     {
       identifier.type.ts_mode = TimestampMode::boot;
     }
+    else if (identifier.ident == "tai")
+    {
+      identifier.type.ts_mode = TimestampMode::tai;
+    }
     else
     {
       goto err;
@@ -1361,6 +1365,13 @@ void SemanticAnalyser::visit(Call &call)
           check_arg(call, Type::timestamp_mode, 0))
       {
         call.type.ts_mode = call.vargs->at(0)->type.ts_mode;
+      }
+
+      if (call.type.ts_mode == TimestampMode::tai &&
+          !bpftrace_.feature_->has_helper_ktime_get_tai_ns())
+      {
+        LOG(ERROR, call.loc, err_)
+            << "Kernel does not support tai timestamp, please try sw_tai";
       }
     }
   }
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -133,7 +133,20 @@ void SemanticAnalyser::visit(Identifier &identifier)
     identifier.type = CreateInt(
         std::get<0>(getIntcasts().at(identifier.ident)));
   }
+  else if (func_ == "nsecs")
+  {
+    identifier.type = CreateTimestampMode();
+    if (identifier.ident == "boot")
+    {
+      identifier.type.ts_mode = TimestampMode::boot;
+    }
+    else
+    {
+      goto err;
+    }
+  }
   else {
+  err:
     identifier.type = CreateNone();
     LOG(ERROR, identifier.loc, err_)
         << "Unknown identifier: '" + identifier.ident + "'";
@@ -1333,6 +1346,19 @@ void SemanticAnalyser::visit(Call &call)
       }
     }
     call.type = CreateUInt32();
+  }
+  else if (call.func == "nsecs")
+  {
+    if (check_varargs(call, 0, 1))
+    {
+      call.type = CreateUInt64();
+      call.type.ts_mode = TimestampMode::boot;
+      if (call.vargs && call.vargs->size() == 1 &&
+          check_arg(call, Type::timestamp_mode, 0))
+      {
+        call.type.ts_mode = call.vargs->at(0)->type.ts_mode;
+      }
+    }
   }
   else
   {
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -2004,16 +2004,26 @@ std::string BPFtrace::resolve_timestamp(uint32_t mode,
 {
   static const auto usec_regex = std::regex("%f");
   TimestampMode ts_mode = static_cast<TimestampMode>(mode);
+  struct timespec zero = {};
+  struct timespec *basetime = &zero;
 
-  if (!boottime_ && ts_mode == TimestampMode::boot)
+  if (ts_mode == TimestampMode::boot)
   {
-    LOG(ERROR) << "Cannot resolve timestamp due to failed boot time calcuation";
-    return "(?)";
+    if (!boottime_)
+    {
+      LOG(ERROR)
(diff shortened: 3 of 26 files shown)


chengshuyi:
#2512

ajor (changes requested):
Thanks for working on this!

I've got a few of high-level concerns about `nsecs_tai` which probably need some discussion.

### What is the use-case?
I can see timestamps being useful for when you leave a script running and come back to it later, just to eyeball when an event happened, but didn't our existing time functions let us do this?
If the goal here is to allow correlating events across machines, then we'll need to be more careful. Will the VDSO-sandwich fallback be accurate enough or would it be better to disable tai timestamps on older kernels?

### More builtins
In general I'm not keen on adding more builtins and bloating the language.

Maybe having nsecs accept an argument like ustack would be nicer, e.g. `nsecs(monotonic)`,`nsecs(boot)`,`nsecs(tai)`.

### Handling multiple timestamps
If we're now allowing multiple different timestamps, we will need to ensure that we don't get incorrect results from mixing them up with each other.

For example, as you [spotted in the issue](https://github.com/iovisor/bpftrace/issues/2512#issuecomment-1541579300), strftime is now returning the wrong result for this new builtin. We will need to return the correct result here for all timestamps, or at the very least have strftime reject unsupported variants.

I think the best way of achieving this would probably be to add some new "type" to bpftrace's semantic analyser, one for each of the timestamps. However, we'd need to maintain backwards compatibility and allow these timestamp types to be used as integers everywhere too.

chengshuyi:
Thank you for your detailed review

> ### What is the use-case?
>  I can see timestamps being useful for when you leave a script running and come back to it later, just to eyeball when an event happened, but didn't our existing time functions let us do this?

There are three main considerations
- The precision of time fucntion is not enough. Generally speaking, the precision of microseconds is needed
- strftime (fmt, nsecs) can get the timestamp, but the semantics of nsecs here is no longer nsecs-since-boot. So here is a deviation from the usual understanding.
- Sometimes an integer timestamp is required because it is easier to handle.

>Will the VDSO-sandwich fallback be accurate enough or would it be better to disable tai timestamps on older kernels?

At present, most of the kernel versions in the production environment stay at 4.19 and 5.10, so we may need to support tai timestmaps on older kernels. For network diagnostics, an accuracy of tens of nanoseconds is sufficient. Of course, we can provide something like nsecs(tai_sw) to let users make their own choices.

> ### More builtins
> In general I'm not keen on adding more builtins and bloating the language.
> 
> Maybe having nsecs accept an argument like ustack would be nicer, e.g. `nsecs(monotonic)`,`nsecs(boot)`,`nsecs(tai)`.
> 

Great idea, i like this.

> ### Handling multiple timestamps
> If we're now allowing multiple different timestamps, we will need to ensure that we don't get incorrect results from mixing them up with each other.
> 
> For example, as you [spotted in the issue](https://github.com/iovisor/bpftrace/issues/2512#issuecomment-1541579300), strftime is now returning the wrong result for this new builtin. We will need to return the correct result here for all timestamps, or at the very least have strftime reject unsupported variants.
> 
> I think the best way of achieving this would probably be to add some new "type" to bpftrace's semantic analyser, one for each of the timestamps. However, we'd need to maintain backwards compatibility and allow these timestamp types to be used as integers everywhere too.

I don’t know if it is possible to modify strftime, let strftime(fmt, nsecs) print nsecs timestamp instead of nsecs+boottime. In this case, strftime(fmt, nsecs(tai)) can print normal timestamp

src/libbpf/bpf.h
```diff
@@ -279,7 +279,71 @@ enum bpf_attach_type {
...
+	FN(dynptr_write),		\
+	FN(dynptr_data),		\
+	FN(tcp_raw_gen_syncookie_ipv4),		\
+	FN(tcp_raw_gen_syncookie_ipv6),		\
+	FN(tcp_raw_check_syncookie_ipv4),		\
+	FN(tcp_raw_check_syncookie_ipv6),		\
+	FN(ktime_get_tai_ns),		\
+	FN(user_ringbuf_drain),		\
+	FN(cgrp_storage_get),		\
+	FN(cgrp_storage_delete),
```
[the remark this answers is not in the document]
chengshuyi: fix

src/ast/passes/codegen_llvm.cpp
```diff
@@ -158,6 +158,30 @@ void CodegenLLVM::visit(Builtin &builtin)
...
+  {
+    if (bpftrace_.feature_->has_helper_ktime_get_tai_ns())
+    {
+      expr_ = b_.CreateGetTaiNs(builtin.loc);
+    }
+    else
+    {
+      if (bpftrace_.delta_taitime_.has_value())
+      {
+        uint64_t delta = ((uint64_t)bpftrace_.delta_taitime_->tv_sec) * 1e9 +
```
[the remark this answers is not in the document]
chengshuyi: ok, using too much c language :)

src/ast/passes/codegen_llvm.cpp
```diff
@@ -158,6 +158,30 @@ void CodegenLLVM::visit(Builtin &builtin)
...
+                         bpftrace_.delta_taitime_->tv_nsec;
+        expr_ = b_.CreateGetNs(
```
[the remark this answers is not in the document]
chengshuyi: ktime_get_ns is also bootime.

chengshuyi:
This may not be the final solution, so test and documentation will be added after the solution is determined

chengshuyi:
```shell
1: With diff:
1: @@ -55,5 +55,5 @@
1:  lookup_success:                                   ; preds = %event_loss_counter
1:    %13 = bitcast i8* %lookup_elem to i64*
1: -  %14 = atomicrmw add i64* %13, i64 1 seq_cst, align 1
1: +  %14 = atomicrmw add i64* %13, i64 1 seq_cst
1:    br label %lookup_merge
1:  
1: @@ -70,9 +70,9 @@
1:  declare void @llvm.lifetime.start.p0i8(i64 immarg %0, i8* nocapture %1) #1
1:  
1: -; Function Attrs: argmemonly nofree nounwind willreturn writeonly
1: +; Function Attrs: argmemonly nofree nosync nounwind willreturn writeonly
1:  declare void @llvm.memset.p0i8.i64(i8* nocapture writeonly %0, i8 %1, i64 %2, i1 immarg %3) #2
1:  
1: -; Function Attrs: argmemonly nofree nounwind willreturn
1: -declare void @llvm.memcpy.p0i8.p0i8.i64(i8* noalias nocapture writeonly %0, i8* noalias nocapture readonly %1, i64 %2, i1 immarg %3) #3
1: +; Function Attrs: argmemonly nofree nosync nounwind willreturn
1: +declare void @llvm.memcpy.p0i8.p0i8.i64(i8* noalias nocapture writeonly %0, i8* noalias nocapture readonly %1, i64 %2, i1 immarg %3) #1
1:  
1:  ; Function Attrs: argmemonly nofree nosync nounwind willreturn
1: @@ -81,4 +81,3 @@
1:  attributes #0 = { nounwind }
1:  attributes #1 = { argmemonly nofree nosync nounwind willreturn }
1: -attributes #2 = { argmemonly nofree nounwind willreturn writeonly }
1: -attributes #3 = { argmemonly nofree nounwind willreturn }\n
1: +attributes #2 = { argmemonly nofree nosync nounwind willreturn writeonly }\n
1: 
1: the following program failed: 'kprobe:f { printf("%s", strftime("%M:%S", nsecs)); }'

```
I don't know why this test failed. I'll check it tormorrow

src/bpftrace.cpp
```diff
@@ -1957,21 +1959,45 @@ std::string BPFtrace::resolve_uid(uintptr_t addr) const
...
-    return "(?)";
+    if (!boottime_)
+    {
+      LOG(ERROR)
+          << "Cannot resolve timestamp due to failed boot time calcuation";
+      return "(?)";
+    }
+    else
+    {
+      basetime = boottime_.operator->();
```
viktormalik, line 1981: Maybe better
```suggestion
      basetime = &boottime_.value();
```

viktormalik (commented):
I like this approach better than introducing new builtins. Giving users an explicit choice between TAI obtained from BPF helper and from the VDSO-sandwich seems like a reasonable approach to me.

src/bpftrace.cpp
chengshuyi: OK, edited. In addition, the `support sw_tai timestamp` patch has also been modified. Because currently nsecs("sw_tai") will return boot time instead of tai time. So put the delta of the timestamp in the codegen stage instead of resolve_timestamp.

viktormalik:
Mostly LGTM.

> I don’t know if it is possible to modify strftime, let strftime(fmt, nsecs) print nsecs timestamp instead of nsecs+boottime. In this case, strftime(fmt, nsecs(tai)) can print normal timestamp

The problem is backwards compatibility here: this change will basically change meaning of all existing `strftime` calls. I agree with @ajor that we should either disable usage of `strftime` for tai timestamps or not implicitly add boottime for them to get (roughly) the same number from `strftime(fmt, nsecs)` and `strftime(fmt, nsecs("tai"))`.

ajor:
Also, can we have the nsecs function accept an identifier rather than a string? e.g. `nsecs(tai)`, without the `"`

This would match kstack/ustack and look a bit cleaner.

I'm not sure if it really needs a new lexer token like `STACK_MODE`, or if just using `IDENT` would work (this would seem preferable if possible).

ajor:
I merged the first two of your commits (updating libbpf/bpf.h) into master as they're useful for something else I'm working on - thanks!

chengshuyi:
Mainly two changes were made:
1. According to @ajor 's suggestion, use identifier to indicate the type of clock instead of string.
2. Added semantic, codegen and runtime tests. No parser tests were added because there may not have been a syntax change involved.

> but also make sure that timestamps with different sources can't be compared against each other / subtracted / etc., to help avoid buggy scripts

I understand that different clock types really shouldn't be compared, but nsecs() returns uint64 type, so from the uint64 type, being able to compare should also be acceptable. I also have relevant examples in the documentation. see patch: [add nsecs() guide](https://github.com/iovisor/bpftrace/pull/2619/commits/390de233136f39e7de4105363befcd115198d54d)

src/ast/passes/semantic_analyser.cpp
```diff
@@ -133,7 +133,28 @@ void SemanticAnalyser::visit(Identifier &identifier)
...
+      identifier.type.ts_mode = TimestampMode::tai;
+    }
+    else if (identifier.ident == "sw_tai")
+    {
+      identifier.type.ts_mode = TimestampMode::sw_tai;
+    }
+    else
+    {
+      goto err;
+    }
```
viktormalik, line 154: I think that transforming this into a `switch` would make it more resistant to errors - the static analyser should report when an enum case is not covered.

src/types.h
```diff
@@ -209,7 +218,8 @@ class SizedType
   bool IsPrintableTy()
   {
     return type != Type::none && type != Type::stack_mode &&
-           (!IsCtxAccess() || is_funcarg); // args builtin is printable
+           (!IsCtxAccess() || is_funcarg) &&
+           type != Type::timestamp_mode; // args builtin is printable
```
viktormalik, line 222: More readable and also the comment is well-placed:
```suggestion
           type != Type::timestamp_mode &&
           (!IsCtxAccess() || is_funcarg); // args builtin is printable
           
```

tests/runtime/call
```diff
@@ -511,3 +511,18 @@ NAME debugf_with_seq_printf
...
+PROG i:ms:1 { printf("SUCCESS %d\n", nsecs()); exit(); }
+EXPECT SUCCESS -?[0-9][0-9]*
+TIMEOUT 5
+
+NAME nsecs_boot
+PROG i:ms:1 { printf("SUCCESS %d\n", nsecs(boot)); exit(); }
+EXPECT SUCCESS -?[0-9][0-9]*
+TIMEOUT 5
+
+NAME nsecs_sw_tai
```
viktormalik, line 525: We could also add a runtime test for `nsecs(tai)` which would require the TAI support by kernel (feature detection for TAI would have to be added to the test runner).

tests/semantic_analyser.cpp
```diff
@@ -2702,6 +2703,16 @@ TEST(semantic_analyser, string_size)
   ASSERT_EQ(var_assign->var->type.GetField(0).type.GetSize(), 6UL);
 }
 
+TEST(semantic_analyser, call_nsecs)
+{
+  test("BEGIN { $ns = nsecs(); }", 0);
+  test("BEGIN { $ns = nsecs(boot); }", 0);
+  // The CI kernel doesn't support tai related helper function
+  // test("BEGIN { $ns = nsecs(tai); }", 0);
```
viktormalik, line 2711: It should be possible to add a conditional test for this:
```cpp
TEST(semantic_analyser, call_nsecs_tai)
{
  BPFfeature feature;
  if (!feature.has_helper_ktime_get_tai_ns())
    GTEST_SKIP();

  test("BEGIN { $ns = nsecs(tai); }", 0);
}
```

tests/codegen/call_nsecs.cpp
```diff
@@ -0,0 +1,14 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_nsecs)
+{
+  test("k:f { @x = nsecs(); }", NAME);
```
viktormalik, line 9: I don't think that it's necessary to have a test for the `nsecs()` call since we already have a codegen test for `nsecs` as a builtin. Maybe rather add codegen test for `nsecs(tai)`?

tests/semantic_analyser.cpp
ajor: It'd be better to use `MockBPFfeature` here - the unit tests shouldn't depend on the environment.

@chengshuyi you'll need to add your new feature to MockBPFfeature, like I am doing here: https://github.com/iovisor/bpftrace/pull/2645/files#diff-aa3b9f9492e765d109d2fd769d283c0fd6b0a76ffcd11cd02054041ae0598948R120

The semantic analyser tests will run with this feature enabled once you do this.

(Independently, I'm refactoring this mocking so that we won't need to remember to update this separate class every time a test requires a new feature)

src/ast/passes/semantic_analyser.cpp
chengshuyi: C++ should not be able to perform `switch case` on strings, maybe I failed to understand what you mean. Thanks.

tests/runtime/call
chengshuyi: Update to patch: [add runtime test for nsecs()](https://github.com/iovisor/bpftrace/pull/2619/commits/ba492b0887d7bc1bb3b7d8a97d0e04e8f89c945e).Thanks.

tests/semantic_analyser.cpp
chengshuyi: Update to patch: [add semantic analyser test for nsecs()](https://github.com/iovisor/bpftrace/pull/2619/commits/f899a1523da41757255ccb03945016f66db7bcc5). Thanks.

tests/codegen/call_nsecs.cpp
chengshuyi: Update to patch: [add codegen test for nsecs()](https://github.com/iovisor/bpftrace/pull/2619/commits/edea0147485c4b1764b642b8c400625bd5d39053). Thanks.

src/ast/passes/semantic_analyser.cpp
viktormalik: Ah, my bad, sorry.

tests/codegen/call_nsecs.cpp
```diff
@@ -0,0 +1,14 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_nsecs)
```
viktormalik, line 7: nit
```suggestion
TEST(codegen, call_nsecs_tai)
```
same for the file name

viktormalik (approved):
LGTM, many thanks for this work!

danobi:
Added the waiting-on-feedback label since it looks like another +1 would be good.

src/types.h
```diff
@@ -93,6 +94,13 @@ struct StackType
   }
 };
 
+enum class TimestampMode
```
ajor, line 97:
```suggestion
enum class TimestampMode : uint8_t
```
Nit: I updated the other enums in this file to keep the size of SizedType down, so let's do the same to this one

ajor (approved):
Looks good to me, thanks!

ajor:
There are some test failures in the intermediate commits of this stack, so I'll squash-merge this tomorrow unless there's any objection

PR #2621, opened 2023-05-24
Align byte-arrays to 8 bytes

viktormalik:
Several types defined as byte-arrays (e.g. `usym`) are represented using 64-bit integers in codegen. Since 64-bit ints require 8-byte alignment, set that to be the default for byte arrays, too.

##### Checklist

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -528,7 +528,7 @@ ssize_t SizedType::GetAlignment() const
     return GetSize();
   else if (IsArrayTy())
     return element_type_->GetAlignment();
-  else if (IsByteArray() || GetSize() <= 4)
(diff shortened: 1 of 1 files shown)


danobi (approved):
Hmm, why doesn't this result in any codegen changes?

viktormalik:
> Hmm, why doesn't this result in any codegen changes?

That's b/c `SizedType::GetAlignment` is only used for computing alignments in tuples and we don't have any codegen test which would have a tuple with a byte-array not aligned to 8 bytes.

Anyways, I just discovered that this change breaks some programs so I'll have to solve this more properly.

Closing for now and will post another PR.

PR #823, opened 2019-07-15
print: skip keys which were removed during iteration

andreasgerstmayr:
I'm consuming bpftrace maps with another program by printing them to stdout every 1s (`interval:s:1`). I just ran into the following issue:
In `int BPFtrace::print_map(IMap &map, uint32_t top, uint32_t div)` there is an iteration over all keys with `bpf_get_next_key`, and a few lines below they key is accessed with `bpf_lookup_elem`. In case the key is removed during `bpf_get_next_key` and `bpf_lookup_elem` by the eBPF program, bpftrace aborts with
```
Error looking up elem: -1
terminate called after throwing an instance of 'std::runtime_error'
  what():  Could not print map with ident "@qtime", err=-1
```

I'd suggest to ignore key not found errors here, and just continue the iteration.
Thoughts?

Looking at the man page of `bpf_lookup_elem`, it can return only 0 for success or -1 for key not found. I left the error handling code intact, in case the API returns another error code for a new type of error at some point.

PR #2623, opened 2023-05-24
Fix crashes caused by improper error handling on bpf_look/update/delete_elem

maokelong:
bpf_*_elem returns -errno when error occured. These functions will return -ENOENT(-2) when they failed to find a entry. It's a harmless error that could be found when doing delete and print/zero/clear operations on the same map simultaneously. It's not reasonable to stop the whole program at this moment.

A simple script could reproduce this annoying situation:
```
i:us:1 {
    @data[rand % 100] = count();
    if (rand % 5 == 0) {
        delete(@data[rand % 10]);
    }
}

i:ms:1 {
    clear(@data);
    // or print(@data);
    // or zero(@data);
}
```

This script crashed almost immediately and reports errors likes:

```
Attaching 2 probes...
ERROR: failed to look up elem: -2
  what():  Could not clear map with ident "@data", err=-1
fish: 'sudo bpftrace ../../../workspac…' terminated by signal SIGABRT (Abort)
```

But It's accepable for users that some entries are deleted asynchronously when tring to clear/print/zero a map.

##### Checklist

### DIFF
--- a/tests/runtime/scripts/parallel_map_access.bt
+++ b/tests/runtime/scripts/parallel_map_access.bt
@@ -1,0 +1,24 @@
+// Regression testing the map's thread safety.
+// Bpftrace shouldn't crash during map's print/clear/zero operations
+// while delete is called at the same time.
+//
+// For more information, please refer to:
+// https://github.com/iovisor/bpftrace/pull/2623
+
+i:us:1 {
+    @data[rand % 100] = count();
+    if (rand % 5 == 0) {
+        delete(@data[rand % 10]);
+    }
+}
+
+i:ms:1 {
+    print(@data);
+    clear(@data);
+    zero(@data);
+}
+
+i:s:1 {
+    print("SUCCESS");
+    exit();
+}
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1,5 +1,6 @@
 #include <arpa/inet.h>
 #include <cassert>
+#include <cerrno>
 #include <cstdio>
 #include <cstring>
 #include <ctime>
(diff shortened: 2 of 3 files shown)


danobi (commented):
Change lgtm. But could you please add a runtime test so we can avoid future regressions?

The script you provided in your PR description seems reasonable to use.

Here's some hints on runtime tests: https://github.com/iovisor/bpftrace/blob/master/tests/README.md#runtime-tests

maokelong:
> Change lgtm. But could you please add a runtime test so we can avoid future regressions?
> 
> The script you provided in your PR description seems reasonable to use.
> 
> Here's some hints on runtime tests: https://github.com/iovisor/bpftrace/blob/master/tests/README.md#runtime-tests


@danobi Done. And here's the test report on my pc.  : )

Without my patch:
image

With my patch:
image

tests/runtime/scripts/parallel_map_access.bt
danobi, line 1: Please add a short description of the bug this script is checking for. Will be helpful in the future when we forget the bug.

danobi (commented):
Small nit but lgtm otherwise

tests/runtime/scripts/parallel_map_access.bt
maokelong: done

danobi (approved):
thanks!

viktormalik (approved):
Looks good, thanks! Could you also add a CHANGELOG entry?

danobi:
I'll push a changelog entry up for this

ajor:
Thanks for the fix!

This did confuse me as I thought we'd already handled at least one of these cases years ago: https://github.com/iovisor/bpftrace/pull/823

With a bit of digging it turns out that libbpf changed how they returned errors in 2021: https://github.com/libbpf/libbpf/commit/7c7ba067fc82a1228e344dfdbc35e9184dfeedaf

Issue #2516, opened 2023-03-03
RFC: Support for user-defined functions


lenticularis39:
Following from #2459: The idea is to implement a function feature into the bpftrace language.

Implementation-wise, newer versions of the kernel and libbpf support [BPF subprogs](https://lwn.net/Articles/741773/) ([kernel commit](https://github.com/torvalds/linux/commit/cc8b0b92a1699bc32f7fec71daa2bfc90de43a4d)), generated by LLVM from LLVM IR functions; this can be used for bpftrace functions. For older kernels (or when wanted by the user), inlining can be used, either explicitely done by the codegen or using LLVM's inline pass.

Syntactically, this would mean a new type of top-level structure besides probes. The keyword could be `function` like in JavaScript, or `fn` for a shorter name (or possibly both, but not `func`, which is already used as a built-in). I think `$arg` makes sense for function arguments, since we already use it for local (scratch) variables. Example:

```
function add($a, $b) {
    return $a + $b;
}

BEGIN {
   printf("1 + 1 = %d\n", add(1, 1));
}
```

As of typing, there are multiple options:
- Specify type in the code. The design of bpftrace currently is to synthesize types from values, avoiding explicit type names in the code, if this option is chosen, this would have to be changed.
-  Derive types from first call, thrown an error when the type doesn't match.
-  Generate one function for each valid assignment of argument types among function calls. This is the most flexible, but it has the disadvantage of possibly generating large amounts of code without the programmer noticing.

A similar situation occurs with built-ins. We might either pass the required context from the caller (and then either generate one function for each type of probe from which the function is called, or allow such a function to be called from only one probe type), or simply disallow built-ins in functions and require the user to pass the arguments by hand. The latter is simpler, but will take additional stack space and copy instructions even when not needed.

fbs:
> Implementation-wise, newer versions of the kernel and libbpf support BPF subprogs (kernel commit), generated by LLVM from LLVM IR functions; this can be used for bpftrace functions. For older kernels (or when wanted by the user), inlining can be used, either explicitely done by the codegen or using LLVM's inline pass.

'newer' is 2017 right? 

> As of typing, there are multiple options:

I would like to see type signatures in the function. It would make the implementation a lot simpler (even though we have to decide syntax (i quite like ($x: typename)) as we don't have to analyse the function body and we can just create simple errors in the semantic analyser.

[...]

PR #2624, opened 2023-05-25
Add user function definition support

lenticularis39:
A new structure for defining subprograms is added on the same level as probes:
```
fn <function_name>(<type0> <arg0>, ...): <return_type> {
  <body>
}
```
The existing return statement is extended to also allow returning a specified value. SemanticAnalyser checks if the type is correct, while CodegenLLVM goes through the generated code to ensure all code flow paths return a value in a non-void function.

Only a limited number of builtins are allowed to be used inside functions due to current implementation restrictions; these can be addressed in later development.

----

This is the first step in implementing #2516 that can be either merged separately or together with the rest of the implementation. Note that this does not implement a way to call the functions nor a way to generate the correct BPF code for that.

(I'm opening this draft PR, since I believe there is quite a few non-trivial changes that it makes sense to review while I work on the rest.)

##### Checklist

Note: this checklist is only for the function definition part, excluding the BPF code generator and function calls.

- [x] Tests for parser
- [x] Tests for semantic analyser
- [x] Tests for codegen
- [x] Function argument support

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -44,6 +44,8 @@ MAKE_ACCEPT(While)
 MAKE_ACCEPT(Config)
 MAKE_ACCEPT(Jump)
 MAKE_ACCEPT(Probe)
+MAKE_ACCEPT(SubprogArg)
+MAKE_ACCEPT(Subprog)
 MAKE_ACCEPT(Program)
 
 #undef MAKE_ACCEPT
@@ -179,6 +181,12 @@ Unroll::~Unroll()
   delete stmts;
   stmts = nullptr;
 }
+Jump::~Jump()
+{
+  if (return_value)
+    delete return_value;
+  return_value = nullptr;
+}
 Predicate::~Predicate()
 {
   delete expr;
@@ -209,6 +217,15 @@ Config::~Config()
   delete stmts;
 }
 
+Scope::~Scope()
+{
+  if (stmts)
+    for (auto *stmt : *stmts)
+      delete stmt;
+  delete stmts;
+  stmts = nullptr;
+}
+
 Probe::~Probe()
 {
   if (attach_points)
@@ -219,16 +236,22 @@ Probe::~Probe()
 
   delete pred;
   pred = nullptr;
+}
 
-  if (stmts)
-    for (Statement *s : *stmts)
-      delete s;
-  delete stmts;
-  stmts = nullptr;
+Subprog::~Subprog()
+{
+  if (args)
+    for (SubprogArg *a : *args)
+      delete a;
+  delete args;
 }
 
 Program::~Program()
 {
+  if (functions)
+    for (Subprog *s : *functions)
+      delete s;
+  delete functions;
   if (probes)
     for (Probe *p : *probes)
       delete p;
@@ -427,17 +450,46 @@ Unroll::Unroll(Expression *expr, StatementList *stmts, location loc)
 {
 }
 
+Scope::Scope(StatementList *stmts) : stmts(stmts)
+{
+}
+
 Probe::Probe(AttachPointList *attach_points,
              Predicate *pred,
              StatementList *stmts)
-    : attach_points(attach_points), pred(pred), stmts(stmts)
+    : Scope(stmts), attach_points(attach_points), pred(pred)
+{
+}
+
+SubprogArg::SubprogArg(std::string name, SizedType type)
+    : type(std::move(type)), name_(std::move(name))
+{
+}
+
+std::string SubprogArg::name() const
+{
+  return name_;
+}
+
+Subprog::Subprog(std::string name,
+                 SizedType return_type,
+                 SubprogArgList *args,
+                 StatementList *stmts)
+    : Scope(stmts),
+      args(args),
+      return_type(std::move(return_type)),
+      name_(std::move(name))
 {
 }
 
 Program::Program(const std::string &c_definitions,
                  Config *config,
+                 SubprogList *functions,
                  ProbeList *probes)
-    : c_definitions(c_definitions), config(config), probes(probes)
+    : c_definitions(c_definitions),
+      config(config),
+      functions(functions),
+      probes(probes)
 {
 }
 
@@ -669,11 +721,27 @@ Cast::Cast(const Cast &other) : Expression(other)
 {
 }
 
-Probe::Probe(const Probe &other) : Node(other)
+SubprogArg::SubprogArg(const SubprogArg &other) : Node(other)
+{
+  name_ = other.name_;
+  type = other.type;
+}
+
+Subprog::Subprog(const Subprog &other) : Scope(other)
+{
+  name_ = other.name_;
+}
+
+Probe::Probe(const Probe &other) : Scope(static_cast<const Scope &>(other))
 {
   need_expansion = other.need_expansion;
   tp_args_structs_level = other.tp_args_structs_level;
   index_ = other.index_;
+}
+
+std::string Subprog::name() const
+{
+  return name_;
 }
 
 bool Probe::has_ap_of_probetype(ProbeType probe_type)
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -494,12 +494,18 @@ public:
   DEFINE_ACCEPT
   DEFINE_LEAFCOPY(Jump)
 
-  Jump(JumpType ident, location loc = location()) : Statement(loc), ident(ident)
+  Jump(JumpType ident, Expression *return_value, location loc = location())
+      : Statement(loc), ident(ident), return_value(return_value)
   {
   }
-  ~Jump() = default;
+  Jump(JumpType ident, location loc = location())
+      : Statement(loc), ident(ident), return_value(nullptr)
+  {
+  }
+  ~Jump();
 
   JumpType ident = JumpType::INVALID;
+  Expression *return_value;
 
 private:
   Jump(const Jump &other) = default;
@@ -566,6 +572,14 @@ private:
   Config(const Config &other);
 };
 
+class Scope : public Node {
+public:
+  Scope(StatementList *stmts);
+  virtual ~Scope() = 0;
+
+  StatementList *stmts;
+};
+
 class AttachPoint : public Node {
 public:
   DEFINE_ACCEPT
@@ -613,7 +627,7 @@ private:
 };
 using AttachPointList = std::vector<AttachPoint *>;
 
-class Probe : public Node {
+class Probe : public Scope {
 public:
   DEFINE_ACCEPT
   DEFINE_LEAFCOPY(Probe)
@@ -623,7 +637,6 @@ public:
 
   AttachPointList *attach_points = nullptr;
   Predicate *pred = nullptr;
-  StatementList *stmts = nullptr;
 
   std::string name() const;
   std::string args_typename() const;
@@ -642,17 +655,58 @@ private:
 };
 using ProbeList = std::vector<Probe *>;
 
+class SubprogArg : public Node {
+public:
+  DEFINE_ACCEPT
+  DEFINE_LEAFCOPY(SubprogArg)
+
+  SubprogArg(std::string name, SizedType type);
+
+  std::string name() const;
+  SizedType type;
+
+private:
+  SubprogArg(const SubprogArg &other);
+  std::string name_;
+};
+using SubprogArgList = std::vector<SubprogArg *>;
+
+class Subprog : public Scope {
+public:
+  DEFINE_ACCEPT
+  DEFINE_LEAFCOPY(Subprog)
+
+  Subprog(std::string name,
+          SizedType return_type,
+          SubprogArgList *args,
+          StatementList *stmts);
+  ~Subprog();
+
+  SubprogArgList *args = nullptr;
+  SizedType return_type;
+
+  std::string name() const;
+
+private:
+  Subprog(const Subprog &other);
+  std::string name_;
+};
+using SubprogList = std::vector<Subprog *>;
+
 class Program : public Node {
 public:
   DEFINE_ACCEPT
   DEFINE_LEAFCOPY(Program)
 
-  Program(const std::string &c_definitions, Config *config, ProbeList *probes);
-
+  Program(const std::string &c_definitions,
+          Config *config,
+          SubprogList *functions,
+          ProbeList *probes);
   ~Program();
 
   std::string c_definitions;
   Config *config = nullptr;
+  SubprogList *functions = nullptr;
   ProbeList *probes = nullptr;
 
 private:
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -10,15 +10,14 @@
 #if LLVM_VERSION_MAJOR <= 16
 #include <llvm-c/Transforms/IPO.h>
 #endif
+#include <llvm/CodeGen/UnreachableBlockElim.h>
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/GlobalVariable.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/LegacyPassManager.h>
 #include <llvm/IR/Metadata.h>
 #include <llvm/IR/Verifier.h>
-#if LLVM_VERSION_MAJOR >= 14
 #include <llvm/Passes/PassBuilder.h>
-#endif
 #include <llvm/Transforms/IPO.h>
 #if LLVM_VERSION_MAJOR <= 16
 #include <llvm/Transforms/IPO/PassManagerBuilder.h>
@@ -884,7 +883,8 @@ void CodegenLLVM::visit(Call &call)
     expr_ = b_.CreateRegisterRead(ctx_, offset, call.func + "_" + reg_name);
   } else if (call.func == "printf") {
     // We overload printf call for iterator probe's seq_printf helper.
-    if (probetype(current_attach_point_->provider) == ProbeType::iter) {
+    if (!inside_subprog_ &&
+        probetype(current_attach_point_->provider) == ProbeType::iter) {
       auto nargs = call.vargs->size() - 1;
 
       int ptr_size = sizeof(unsigned long);
@@ -2225,7 +2225,11 @@ void CodegenLLVM::visit(Jump &jump)
   switch (jump.ident) {
     case JumpType::RETURN:
       // return can be used outside of loops
-      createRet();
+      if (jump.return_value) {
+        auto scoped_del = accept(jump.return_value);
+        createRet(expr_);
+      } else
+        createRet();
       break;
     case JumpType::BREAK:
       b_.CreateBr(std::get<1>(loops_.back()));
@@ -2383,11 +2387,59 @@ void CodegenLLVM::generateProbe(Probe &probe,
         func_type, section_name, current_attach_point_->address, index);
 }
 
+void CodegenLLVM::visit(Subprog &subprog)
+{
+  std::vector<llvm::Type *> arg_types;
+  // First argument is for passing ctx pointer for output, rest are proper
+  // arguments to the function
+  arg_types.push_back(b_.getInt8PtrTy());
+  std::transform(subprog.args->begin(),
+                 subprog.args->end(),
+                 std::back_inserter(arg_types),
+                 [this](SubprogArg *arg) { return b_.GetType(arg->type); });
+  FunctionType *func_type = FunctionType::get(b_.GetType(subprog.return_type),
+                                              arg_types,
+                                              0);
+
+  Function *func = Function::Create(
+      func_type, Function::InternalLinkage, subprog.name(), module_.get());
+  BasicBlock *entry = BasicBlock::Create(module_->getContext(), "entry", func);
+  b_.SetInsertPoint(entry);
+
+  variables_.clear();
+  ctx_ = func->arg_begin();
+  inside_subprog_ = true;
+
+  int arg_index = 0;
+  for (SubprogArg *arg : *subprog.args) {
+    auto alloca = b_.CreateAllocaBPF(b_.GetType(arg->type), arg->name());
+    b_.CreateStore(func->getArg(arg_index + 1), alloca);
+    variables_.insert({ arg->name(), alloca });
+    ++arg_index;
+  }
+
+  for (Statement *stmt : *subprog.stmts) {
+    auto scoped_del = accept(stmt);
+  }
+  if (subprog.return_type.IsVoidTy())
+    createRet();
+
+  FunctionPassManager fpm;
+  FunctionAnalysisManager fam;
+  llvm::PassBuilder pb;
+  pb.registerFunctionAnalyses(fam);
+  fpm.addPass(UnreachableBlockElimPass());
+  fpm.run(*func, fam);
+}
+
 void CodegenLLVM::createRet(Value *value)
 {
   // If value is explicitly provided, use it
   if (value) {
     b_.CreateRet(value);
+    return;
+  } else if (inside_subprog_) {
+    b_.CreateRetVoid();
     return;
   }
 
@@ -2441,6 +2493,7 @@ void CodegenLLVM::visit(Probe &probe)
 
   bool generated = false;
   current_attach_point_ = attach_point;
+  inside_subprog_ = false;
 
   /*
    * Most of the time, we can take a probe like kprobe:do_f* and build a
@@ -2590,6 +2643,9 @@ void CodegenLLVM::visit(Probe &probe)
 
 void CodegenLLVM::visit(Program &program)
 {
+  if (program.functions)
+    for (Subprog *subprog : *program.functions)
+      auto scoped_del = accept(subprog);
   for (Probe *probe : *program.probes)
     auto scoped_del = accept(probe);
 }
--- a/src/ast/passes/codegen_llvm.h
+++ b/src/ast/passes/codegen_llvm.h
@@ -58,7 +58,9 @@ public:
   void visit(Predicate &pred) override;
   void visit(AttachPoint &ap) override;
   void visit(Probe &probe) override;
+  void visit(Subprog &subprog) override;
   void visit(Program &program) override;
+
   AllocaInst *getHistMapKey(Map &map, Value *log2);
   int getNextIndexForProbe();
   Value *createLogicalAnd(Binop &binop);
@@ -175,7 +177,7 @@ private:
 
   // Create return instruction
   //
-  // If null, return value will depend on current attach point
+  // If null, return value will depend on current attach point (void in subprog)
   void createRet(Value *value = nullptr);
 
   // Every time we see a watchpoint that specifies a function + arg pair, we
@@ -248,6 +250,7 @@ private:
   int next_probe_index_ = 1;
   // Used if there are duplicate USDT entries
   int current_usdt_location_index_{ 0 };
+  bool inside_subprog_ = false;
 
   std::map<std::string, AllocaInst *> variables_;
   int printf_id_ = 0;
--- a/src/ast/passes/printer.cpp
+++ b/src/ast/passes/printer.cpp
@@ -367,6 +367,11 @@ void Printer::visit(Jump &jump)
 {
   std::string indent(depth_, ' ');
   out_ << indent << opstr(jump) << std::endl;
+  if (jump.return_value) {
+    ++depth_;
+    jump.return_value->accept(*this);
+    --depth_;
+  }
 }
 
 void Printer::visit(Predicate &pred)
@@ -401,6 +406,27 @@ void Printer::visit(Probe &probe)
   --depth_;
 }
 
+void Printer::visit(Subprog &subprog)
+{
+  std::string indent(depth_, ' ');
+  out_ << indent << subprog.name() << ": " << subprog.return_type;
+
+  out_ << "(";
+  for (size_t i = 0; i < subprog.args->size(); i++) {
+    auto &arg = subprog.args->at(i);
+    out_ << arg->name() << " : " << arg->type;
+    if (i < subprog.args->size() - 1)
+      out_ << ", ";
+  }
+  out_ << ")" << std::endl;
+
+  ++depth_;
+  for (Statement *stmt : *subprog.stmts) {
+    stmt->accept(*this);
+  }
+  --depth_;
+}
+
 void Printer::visit(Program &program)
 {
   if (program.c_definitions.size() > 0)
@@ -416,6 +442,8 @@ void Printer::visit(Program &program)
   }
 
   ++depth_;
+  for (Subprog *subprog : *program.functions)
+    subprog->accept(*this);
   for (Probe *probe : *program.probes)
     probe->accept(*this);
   --depth_;
--- a/src/ast/passes/return_path_analyser.cpp
+++ b/src/ast/passes/return_path_analyser.cpp
@@ -1,0 +1,94 @@
+#include "return_path_analyser.h"
+#include "log.h"
+
+namespace bpftrace {
+namespace ast {
+
+ReturnPathAnalyser::ReturnPathAnalyser(Node *root, std::ostream &out)
+    : root_(root), out_(out)
+{
+}
+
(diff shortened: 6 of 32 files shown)


tests/parser.cpp
```diff
@@ -1971,6 +1971,121 @@ TEST(Parser, invalid_provider)
   test_parse_failure("asdf:xyz:www { }");
 }
 
+TEST(Parser, subprog_only)
+{
+  test("fn f1(): void {}",
+       "Program\n"
+       " f1: void\n");
```
viktormalik, line 1978: Nit: would it be possible to add `()` here so that it is visible that it's a function?
```suggestion
       " f1: void()\n");
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2465,6 +2472,72 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+
+  for (Statement *stmt : *subprog.stmts)
+  {
+    auto scoped_del = accept(stmt);
+  }
+  createRet();
+
+  if (!subprog.type.IsVoidTy())
+  {
+    // Check that every reachable return is non-dummy (for dummy returns, see
```
viktormalik, line 2512: This piece of code would deserve to be enclosed in a separate function, e.g. `checkMissingReturns` or so.

viktormalik (commented):
Looks great, the implementation is pretty straightforward. Just a couple of comments.

I'd advocate for merging this before implementing support for function calls.

lenticularis39:
Resolved comments and rebased.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2465,6 +2472,72 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+      auto term = bb->getTerminator();
+      for (unsigned i = 0, e = term->getNumSuccessors(); i != e; i++)
+      {
+        auto succ = term->getSuccessor(i);
+        if (bfs_visited.find(succ) != bfs_visited.end())
+          continue;
+        bfs_visited.insert(succ);
+        bfs_queue.push_front(succ);
+      }
+    }
```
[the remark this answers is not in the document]
viktormalik: This still applies, right?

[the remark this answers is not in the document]
lenticularis39: Ah yes, I somehow missed that you are suggesting to run unreachable code elimination. The codegen already generates dummy returns for probes, they are just simpler, since the return value of probes is either void or some kind of integer, so there is no need for an empty alloca.

On the other hand, running dead code elimination here might slightly impact the tests (e.g. the pass might fail on invalid LLVM IR or errors in unreachable code won't be caught in verification). I will try to put the pass there and see what it does.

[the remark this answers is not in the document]
lenticularis39: I tried running `UnreachableBlockElimPass` at the end of `void CodegenLLVM::visit(Subprog &subprog)` and it looks good. I think it might make sense to do that in a separate PR together with the same pass for probes rather than here, as I wrote above, this is not an issue specific to user-defined functions, it is just the most visible there.

src/ast/visitors.cpp
ajor, line 1: I think Mutator would need to visit jump.return_value to be fully correct.
Doesn't it also need to visit subprog.args?

However, it seems like Mutator & Dispatcher aren't used anywhere, so this is all dead code anyway. This also means we can't test it.

I don't want to merge in potentially incorrect code, so I think we'd be better off deleting these classes entirely.

cc @fbs

src/lexer.l
```diff
@@ -183,19 +191,41 @@ struct|union|enum       {
...
+  "{"                   {
+                          yy_pop_state(yyscanner);
+                          unput(yytext[0]);
+                          return Parser::make_IDENT(struct_type + " " + trim(buffer), loc);
                         }
+  .                     { buffer += yytext[0]; }
+  \n                    buffer += '\n'; loc.lines(1); loc.step();
+}
+<STRUCT,BRACE>{
+  "*"|")"|","|{space}+"$" {
```
ajor, line 216: Why are the space changes needed in lexer.l (this line and also in the `AFTER_COLON` section)? Why is space handled differently in this section?

ajor (changes requested):
Love the tests!

The one main structural comment I've got is that something like checking all paths return a value should come under semantic analysis rather than codegen. Codegen should be able to assume that what it's been given is valid, and not worry about checking things itself.

Slightly off topic: We need to start caring more about the performance of the parser. Our approach previously was to not worry because it's not on the tracing hot-path, but the unit tests taking multiple seconds to run indicates that we've taken this too far. I did a quick profile and we're spending a lot of our time just copying strings around. I left a few comments on the new changes here, but we'll need to revisit the rest of ast.h at some point.

danobi:
Going to set aside some time this weekend to take a look

tests/parser.cpp
```diff
@@ -1971,6 +1971,121 @@ TEST(Parser, invalid_provider)
   test_parse_failure("asdf:xyz:www { }");
 }
 
+TEST(Parser, subprog_only)
```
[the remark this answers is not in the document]
lenticularis39: Oops, I missed that.

src/ast/ast.h
```diff
@@ -541,6 +547,16 @@ class While : public Statement
   While(const While &other);
 };
 
+class Scope : public Node
+{
+public:
+  Scope(StatementList *stmts);
+  Scope(const Scope &other) = default;
```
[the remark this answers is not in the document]
lenticularis39: Ah, that is true, and I know that - no idea why I put the line there, might have been copied from somewhere and changed because of compiler warnings. Thanks!

src/ast/ast.h
```diff
@@ -541,6 +547,16 @@ class While : public Statement
...
+  virtual ~Scope();
```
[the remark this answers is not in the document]
lenticularis39: I agree, that was the intention - I was just focusing more on the functional side of things and forgot to change it.

src/lexer.l
lenticularis39: `AFTER_COLON` is an `%x` section, IIUC the usual rules do not apply there. Here the space handling should be a part of the logic ensuring `struct test $x` (for example) is parsed correctly.

I can take a closer look at it if that did not make much sense, that's what I remember from writing this code.

src/ast/visitors.cpp
lenticularis39: I'm not sure myself what the Mutator class is supposed to do exactly, you are probably right.

lenticularis39:
> Love the tests!

Thank you! I had to write thorough tests, the area of possibility of bugs here is almost unlimited :D

> 
> The one main structural comment I've got is that something like checking all paths return a value should come under semantic analysis rather than codegen. Codegen should be able to assume that what it's been given is valid, and not worry about checking things itself.
> 

I also thought that initially, however, checking if all paths return a value is actually not semantic analysis, but control flow analysis, since semantic analysis operates on the *AST*, while this checks for a property of the *CFG*. Implementing it in semantic analyser would require to construct the control flow inside the AST, which would duplicate work that is already done in codegen.

> Slightly off topic: We need to start caring more about the performance of the parser. Our approach previously was to not worry because it's not on the tracing hot-path, but the unit tests taking multiple seconds to run indicates that we've taken this too far. I did a quick profile and we're spending a lot of our time just copying strings around. I left a few comments on the new changes here, but we'll need to revisit the rest of ast.h at some point.

Good point.

src/ast/visitors.cpp
ajor: After thinking about this some more and discussing the Mutator with @fbs, I think we're going to remove it entirely for now as it's not used. (it can be brought back when needed)

Basically: don't worry about the mutator in this PR

lenticularis39: OK, thanks.

ajor:
> I also thought that initially, however, checking if all paths return a value is actually not semantic analysis, but control flow analysis, since semantic analysis operates on the _AST_, while this checks for a property of the _CFG_. Implementing it in semantic analyser would require to construct the control flow inside the AST, which would duplicate work that is already done in codegen.

In languages with more complex control flow (e.g. `goto` statements) I think this makes sense, but is it necessary for bpftrace, which only has structured control flow?

I feel like something similar to this pseudo-code should be able to detect missing return statements:
```
fn all_paths_return(block):
  for stmt in block:
    if stmt == "return":
      return true
    if stmt == "break" or stmt == "continue":
      return false
    if stmt == "if":
      if all_paths_return(stmt.true_branch) && all_paths_return(stmt.false_branch):
        return true
    if stmt has children: (e.g. a loop)
      if all_paths_return(stmt.children):
        return true
  return false
```

src/ast/visitors.cpp
fbs: goal of the mutator is to write ast modifying passes, it mutates one ast into another. But never got around to writing mutating passes

src/ast/ast.h
```diff
@@ -615,16 +631,60 @@ class Probe : public Node {
...
+  DEFINE_LEAFCOPY(Subprog)
+
+  Subprog(std::string name,
+          SizedType type,
+          SubprogArgList *args,
+          StatementList *stmts);
+  ~Subprog();
+
+  SubprogArgList *args = nullptr;
+  SizedType type;
```
danobi, line 665: nit: maybe `return_type` instead of `type`? To differentiate between type of function and return type

src/ast/passes/codegen_llvm.h
```diff
@@ -230,6 +241,7 @@ class CodegenLLVM : public Visitor
   int next_probe_index_ = 1;
   // Used if there are duplicate USDT entries
   int current_usdt_location_index_{ 0 };
+  bool inside_subprog_;
```
danobi, line 244: nit: just to make it explicit
```suggestion
  bool inside_subprog_ = false;
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2537,6 +2603,33 @@ void CodegenLLVM::createRet(Value *value)
...
+    for (unsigned i = 0, e = term->getNumSuccessors(); i != e; i++)
+    {
+      auto succ = term->getSuccessor(i);
+      if (bfs_visited.find(succ) != bfs_visited.end())
+        continue;
+      bfs_visited.insert(succ);
+      bfs_queue.push_front(succ);
+    }
+  }
+}
```
danobi, line 2631: Same comment as @ajor -- doesn't feel right to do this in codegen

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2502,6 +2550,24 @@ void CodegenLLVM::createRet(Value *value)
...
+    // rejected later; dummy value is needed for the LLVM IR to be correct
+    auto ret_type = b_.GetInsertBlock()->getParent()->getReturnType();
+    if (ret_type->isVoidTy())
+      b_.CreateRetVoid();
+    else
+    {
+      auto dummy_alloca = b_.CreateAlloca(ret_type, nullptr, "ret_dummy");
+      auto dummy = b_.CreateLoad(ret_type, dummy_alloca);
+      auto ret_inst = b_.CreateRet(dummy);
+      ret_inst->setMetadata("ret_dummy", MDTuple::get(b_.getContext(), {}));
```
danobi, line 2567: Does moving return checks into semantic analyser get rid of the ret_dummy stuff? Not immediately clear to me what this does

tests/semantic_analyser.cpp
```diff
@@ -2702,6 +2702,20 @@ TEST(semantic_analyser, string_size)
...
+{
+  test("fn f(): void { return; }", 0);
+  test("fn f(): void { return 1; }", 1);
+  test("fn f(): int64 { return; }", 1);
+  test("fn f(): int64 { return 1; }", 0);
+}
+
+TEST(semantic_analyser, subprog_arguments)
+{
+  test("fn f(int64 $a): int64 { return $a; }", 0);
```
danobi, line 2715: Sorry if this was discussed elsewhere already, but the argument syntax feels a bit odd to me.

The function syntax is `fn <idents and stuff>: <ret type>`. It seems to me it would follow that argument syntax should also be `<ident>: <type>`, but here it's reversed.

IIRC we also added type annotations for local variables as `$var: type = <blah>` right?

danobi: In that case I am in favor of `$arg: type` syntax

fbs:
> > I also thought that initially, however, checking if all paths return a value is actually not semantic analysis, but control flow analysis, since semantic analysis operates on the _AST_, while this checks for a property of the _CFG_. Implementing it in semantic analyser would require to construct the control flow inside the AST, which would duplicate work that is already done in codegen.
> 
> In languages with more complex control flow (e.g. `goto` statements) I think this makes sense, but is it necessary for bpftrace, which only has structured control flow?
> 
> I feel like something similar to this pseudo-code should be able to detect missing return statements:
> 
> ```
> fn all_paths_return(block):
>   for stmt in block:
>     if stmt == "return":
>       return true
>     if stmt == "break" or stmt == "continue":
>       return false
>     if stmt == "if":
>       if all_paths_return(stmt.true_branch) && all_paths_return(stmt.false_branch):
>         return true
>     if stmt has children: (e.g. a loop)
>       if all_paths_return(stmt.children):
>         return true
>   return false
> ```

I'd put something like this in its own pass. Is not semantic but also shouldn't go in the codegen. Passes should be fairly easy to add as all the boilerplate is handled (see e.g. nodecount.h).

Might be a good usecase for the `Dispatcher`  visitor as it allows for a return value which makes it a bit easier to write. 

Might be good to do it in a separte PR to avoid some noise in this one

lenticularis39:
> I'd put something like this in its own pass. Is not semantic but also shouldn't go in the codegen. Passes should be fairly easy to add as all the boilerplate is handled (see e.g. nodecount.h).
> 
> Might be a good usecase for the `Dispatcher` visitor as it allows for a return value which makes it a bit easier to write.
> 
That sounds like the best idea to me, that will avoid the check being in codegen and having it returning bool, and also complicating the code of semantic analyzer.

Also, there is a direct benefit of doing it on the AST as  @ajor and @danobi suggested: we can print the exact place where a return instruction should be and isn't, which will make it easier for the programmer.

> Might be good to do it in a separte PR to avoid some noise in this one
There might be just an issue with the fact that without the check, bpftrace will generate invalid LLVM IR; however, I don't think that is that big of an issue, since the function feature is not officially supported yet. I can also make it a separate commit, that might be better.

fbs:
The idea behind the passmanager and all the visitor stuff is that it should make new passes easy to add, to avoid increasing the complexity of existing passes, the semantic analyser is quite tricky in places.

If you do want to try the dispather visitor, it will be the first usecase and its a bit hacky so it might not work that  well :(


Also, really nice work here :)

lenticularis39:
Thank you all for the helpful comments, I feel like it cleared a lot of things up and will increase the quality of the PR a lot.

tests/semantic_analyser.cpp
fbs: I wonder if this type syntax can be expanded to other cases, like:

```
kprobe:some_kernel_fn(a: int, b: int): int
```

Basically adding the option to specify function signatures in places (althoug with btf and dwarf that might be a bit late). In that case, i wonder if having `->` like rust,python will make parsing easier

lenticularis39:
Resolved (hopefully) all of the comments. There are some minor issues with the code like leftover declarartions, I will clean it after rebasing to current master.

danobi:
Top of my todo list to review

tests/return_path_analyser.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+TEST(return_path_analyser, simple_no_return)
+{
+  test("fn test(): int64 { $x = 0; }", 1);
+}
+
+TEST(return_path_analyser, if_else)
+{
+  test("fn test($x: int64): int64 { if ($x > 0) { return 0; } else { return 1; "
+       "} }",
```
viktormalik, line 61: Can we do something with the ugly linebreaks or is it clang-format enforcing this?
```suggestion
  test("fn test($x: int64): int64 {"
       "  if ($x > 0) { return 0; } else { return 1; }"
       "}",
```
Same below.

viktormalik (commented):
Looks good! I mostly have some minor suggestions.

src/ast/passes/return_path_analyser.cpp
```diff
@@ -0,0 +1,88 @@
...
+namespace ast {
+
+ReturnPathAnalyser::ReturnPathAnalyser(Node *root, std::ostream &out)
+    : root_(root), out_(out)
+{
+}
+
+bool ReturnPathAnalyser::visit(Program &prog)
+{
+  for (Subprog *subprog : *prog.functions)
```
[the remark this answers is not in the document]
lenticularis39: Not only this file, it's a common ugliness of my code :D

tests/return_path_analyser.cpp
lenticularis39: I don't think clang-format is enforcing this, I just let it auto-format the file and it did it like that.

src/ast/ast.cpp
```diff
@@ -204,7 +221,14 @@ Probe::~Probe()
...
 
+Subprog::~Subprog()
+{
+  if (args)
+    for (SubprogArg *a : *args)
+      delete a;
+  delete args;
   if (stmts)
     for (Statement *s : *stmts)
       delete s;
```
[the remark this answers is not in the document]
lenticularis39: Ah I see, there is no double free, since I set `stmts` to `nullptr` so I didn't notice. Nice catch!

danobi:
Took a close look at ast, parser, lexer, return_path_analyser. They look really nice. Will try to make another pass tmr morning over semantic analyser and codegen details.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2160,7 +2197,21 @@ void SemanticAnalyser::visit(Jump &jump)
   switch (jump.ident)
   {
     case JumpType::RETURN:
-      // return can be used outside of loops
+      if (jump.return_value)
+        jump.return_value->accept(*this);
+      if (auto subprog = dynamic_cast<Subprog *>(scope_))
+      {
+        if ((subprog->return_type.IsVoidTy() !=
+             (jump.return_value == nullptr)) ||
```
danobi, line 2205: Nit: easier to read if `is_void && (return_value != nullptr)` IMO

danobi (commented):
semantic analysis lgtm; one small nit

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2582,6 +2587,50 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+                                              arg_types,
+                                              0);
+
+  Function *func = Function::Create(
+      func_type, Function::InternalLinkage, subprog.name(), module_.get());
+  BasicBlock *entry = BasicBlock::Create(module_->getContext(), "entry", func);
+  b_.SetInsertPoint(entry);
+
+  variables_.clear();
+  ctx_ = nullptr;
```
danobi, line 2607: Would it make sense to hoist this out to a helper function? I think `generateProbe()` shares the `variables_.clear()` bit. It would help self-document the intent and make future extensions less error prone

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2582,6 +2587,50 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+  inside_subprog_ = true;
+
+  int arg_index = 0;
+  for (SubprogArg *arg : *subprog.args)
+  {
+    auto alloca = b_.CreateAllocaBPF(b_.GetType(arg->type), arg->name());
+    b_.CreateStore(func->getArg(arg_index), alloca);
+    variables_.insert({ arg->name(), alloca });
+    ++arg_index;
+  }
```
danobi, line 2617: Is it necessary to put arguments on stack? I know `variables_` value is an allocainst. But I'm not sure it's strictly necessary (unless I'm mistaken which I could easily be). AFAIU should be OK to keep variables in register.

src/ast/passes/codegen_llvm.h
```diff
@@ -166,7 +168,7 @@ class CodegenLLVM : public Visitor
 
   // Create return instruction
   //
-  // If null, return value will depend on current attach point
+  // If null, return value will depend on current attach point (void in subprog)
```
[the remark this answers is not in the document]
lenticularis39: Subprogs are allowed to return a value, the comment only refers to `createRet()` being called without an argument/with null argument.

src/ast/passes/codegen_llvm.cpp
lenticularis39: Parts of it are shared, but parts of it are also different, subprogs process arguments, probes do not, probe do additional setup subprogs don't.

danobi (approved):

src/ast/passes/codegen_llvm.h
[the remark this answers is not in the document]
danobi: Ah yeah. Misread the comment

lenticularis39:
Just fixed a small readability issue (moving check for void return value to `ReturnPathAnalyser::visit(Subprog &subprog)`).

src/ast/passes/return_path_analyser.cpp
```diff
@@ -0,0 +1,100 @@
...
+  {
+    if (subprog->return_type.type != Type::voidtype && !visit(*subprog))
```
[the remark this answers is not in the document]
lenticularis39: Fixed.

viktormalik (approved):
Thanks!

@ajor wants to review this, too, so let's wait for that before merging.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -377,7 +391,8 @@ void SemanticAnalyser::visit(Builtin &builtin)
     builtin.type.SetAS(AddrSpace::kernel);
   }
   else if (builtin.ident == "func") {
-    for (auto &attach_point : *probe_->attach_points)
+    auto probe = dynamic_cast<Probe *>(scope_);
+    for (auto &attach_point : *probe->attach_points)
```
ajor, line 395: I'm guessing these bits of code are the reason for disallowing certain builtins in `check_available_in_subprog()`?

I think it'd be safer if we included the checks closer to the point where they're actually needed (here) - I can imagine in the future we'll forget about this implicit link, modify the list in `check_available_in_subprog` and end up with a SEGV from this dynamic_cast returning a null pointer.

i.e. something like:
```suggestion
    auto *probe = dynamic_cast<Probe *>(scope_);
    if (!probe)
    {
      LOG(ERROR, builtin.loc, err_) << builtin.ident << " cannot be used in functions";
      return;
    }
    for (auto &attach_point : *probe->attach_points)
```

ajor (changes requested):
This is really looking pretty good now, thanks!

I just found some bugs that we should fix before merging.

Crash using disallowed builtins in functions:
```
$ sudo ./bpftrace -e 'fn foo(): int64 { return func; }' 
Segmentation fault
```

Crash writing to a map in a function:
```
$ sudo ./bpftrace -e 'fn bar(): void { @x = 1; }'
bpftrace: /home/ajor/src/bpftrace/src/ast/irbuilderbpf.cpp:444: void bpftrace::ast::IRBuilderBPF::CreateMapUpdateElem(llvm::Value*, bpftrace::ast::Map&, llvm::Value*, llvm::Value*, const bpftrace::location&): Assertion `ctx && ctx->getType() == getInt8PtrTy()' failed.
Aborted
```

Crash using map's value in a function:
```
$ sudo ./bpftrace -e 'fn bar(): uint64 { $res = @x + 1; return $res; } BEGIN { @x = 1; }'
bpftrace: /home/ajor/src/bpftrace/src/ast/irbuilderbpf.cpp:379: llvm::Value* bpftrace::ast::IRBuilderBPF::CreateMapLookupElem(llvm::Value*, bpftrace::ast::Map&, llvm::Value*, const bpftrace::location&): Assertion `ctx && ctx->getType() == getInt8PtrTy()' failed.
Aborted
```

Poor error message reading a map from a function:
```
$ sudo ./bpftrace -e 'fn bar(): int64 { return @x; } BEGIN { @x = 1; }'
stdin:1:20-29: ERROR: Function bar is of type int64, cannot return none
fn bar(): int64 { return @x; } BEGIN { @x = 1; }
                   ~~~~~~~~~
```

src/ast/passes/printer.cpp
```diff
@@ -377,6 +383,32 @@ void Printer::visit(Probe &probe)
...
 }
 
+void Printer::visit(Subprog &subprog)
+{
+  std::string indent(depth_, ' ');
+  out_ << indent << subprog.name() << ": " << subprog.return_type;
+
+  out_ << "(";
+  if (subprog.args->size() > 0)
+  {
```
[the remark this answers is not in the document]
lenticularis39: Also I noticed the function outputs the old syntax of arguments:
```
# build/src/bpftrace -de 'fn test($x: int64): void {  } BEGIN {}'
[DEBUG] libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0
[DEBUG] libbpf: loading kernel BTF '/sys/kernel/btf/vmlinux': 0

AST after: parser
-------------------
Program
 test: void(int64 $x)
 BEGIN
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2582,6 +2587,50 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+  if (subprog.return_type.IsVoidTy())
+    createRet();
+
+  FunctionPassManager fpm;
+  FunctionAnalysisManager fam;
+  llvm::PassBuilder pb;
+  pb.registerFunctionAnalyses(fam);
+  fpm.addPass(UnreachableBlockElimPass());
+  fpm.run(*func, fam);
```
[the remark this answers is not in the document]
lenticularis39: Unlike probes which have a default return value, subprograms might leave unreachable blocks that are not finished with a return instruction, these have to be removed for the IR to be correct.

lenticularis39:
Fixed maps and casts in subprogs. Also added more tests to cover this functionality.

Subprog argument implementation in codegen was changed to include ctx as the first argument, which is needed for perf-event-based builtin functions (like print) to work properly. This was tested at runtime with the in-progress function call PR.

With this all feedback should be resolved.

danobi:
Are there any outstanding work items left? Or are we missing a final round of reviews?

Side note, this behavior is a bit unexpected:

```
$ sudo ./build/src/bpftrace -e 'fn foo($a: int32): int32 { return $a + 4; }'
stdin:1:30-43: ERROR: Function foo is of type int32, cannot return int64
fn foo($a: int32): int32 { return $a + 4; }

```

I suspect this is more related to how we represent all ints as int64 internally. Not a show stopper -- IMO fixing this relaxes some rules so this is ok to do later.

danobi:
Also another bit of fix-later feedback:

```
sudo ./build/src/bpftrace -e 'fn foo(): void { return; }'
No probes to attach

$ sudo ./build/src/bpftrace -e 'fn foo() { return; }'
stdin:1:10-11: ERROR: syntax error, unexpected {, expecting :
fn foo() { return; }

```

Going with bpftrace's philosophy of brevity, perhaps good to allow `void` return types to be omitted. But I think this is debatable

lenticularis39:
> Are there any outstanding work items left? Or are we missing a final round of reviews?
> 
No, just waiting for reviews.

> Side note, this behavior is a bit unexpected:
> 
> ```
> $ sudo ./build/src/bpftrace -e 'fn foo($a: int32): int32 { return $a + 4; }'
> stdin:1:30-43: ERROR: Function foo is of type int32, cannot return int64
> fn foo($a: int32): int32 { return $a + 4; }
>                              ~~~~~~~~~~~~~
> ```
> 
> I suspect this is more related to how we represent all ints as int64 internally. Not a show stopper -- IMO fixing this relaxes some rules so this is ok to do later.
> 
> Also the error underline seems to be off.

I know about this, it's because types were not explicitly given anywhere previously (besides casts). Not sure why the underlining is off.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2372,12 +2376,60 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+}
+
 void CodegenLLVM::createRet(Value *value)
 {
   // If value is explicitly provided, use it
   if (value) {
     b_.CreateRet(value);
     return;
+  } else if (inside_subprog_) {
+    b_.CreateRetVoid();
```
ajor, line 2431: I'm assuming this is just a temporary thing until your next PR?

ajor (approved):
Sorry for taking so long to review this. Really appreciate the work you've put it!

The code's looking good to me - the comments are just a bunch of minor things that can be fixed later. Let's just get this one through without more delay!

The only issue I found was that we're allowing multiple functions to be defined with the same name. Again, this can be fixed later:
```
fn aaa(): void { print(1); }
fn aaa(): void { print(2); }
```

src/ast/passes/codegen_llvm.cpp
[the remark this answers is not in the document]
ajor: Just been thinking about this some more and I realise that I don't actually understand. How can a subprogram generate an unreachable block without a return instruction?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2105,8 +2182,13 @@ void SemanticAnalyser::visit(Cast &cast)
   // case : BEGIN { @foo = (struct Foo)0; }
   // case : profile:hz:99 $task = (struct task_struct *)curtask.
   if (cast.type.GetAS() == AddrSpace::none) {
-    ProbeType type = single_provider_type();
-    cast.type.SetAS(find_addrspace(type));
+    if (auto probe = dynamic_cast<Probe *>(scope_)) {
+      ProbeType type = single_provider_type(probe);
+      cast.type.SetAS(find_addrspace(type));
+    } else {
+      // Assume kernel space for data in subprogs
```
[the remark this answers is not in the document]
lenticularis39: Yes, the issue is this probably requires a new language feature, since in the context of a subprog, there is no way to tell if you are accessing userspace or kernel space data based on pointers alone.

src/ast/passes/codegen_llvm.cpp
lenticularis39: This is because unlike with probes, there is not any default return value for subprogs. createRet() without argument is only used to end unreachable code that is then removed in subprogs.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2372,12 +2376,60 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+  // First argument is for passing ctx pointer for output, rest are proper
+  // arguments to the function
+  arg_types.push_back(b_.getInt8PtrTy());
+  std::transform(subprog.args->begin(),
+                 subprog.args->end(),
+                 std::back_inserter(arg_types),
+                 [this](SubprogArg *arg) { return b_.GetType(arg->type); });
+  FunctionType *func_type = FunctionType::get(b_.GetType(subprog.return_type),
+                                              arg_types,
+                                              0);
```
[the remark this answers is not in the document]
lenticularis39: Yeah I must have copied this from somewhere :D

src/ast/passes/codegen_llvm.cpp
```diff
@@ -879,7 +878,8 @@ void CodegenLLVM::visit(Call &call)
     expr_ = b_.CreateRegisterRead(ctx_, offset, call.func + "_" + reg_name);
   } else if (call.func == "printf") {
     // We overload printf call for iterator probe's seq_printf helper.
-    if (probetype(current_attach_point_->provider) == ProbeType::iter) {
+    if (!inside_subprog_ &&
```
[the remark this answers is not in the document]
lenticularis39: Currently subprogs do not track if they are called from an iterator probe so printf from subprog from iterator probe is treated like outside iterator probe. This should be probably fixed eventually.

lenticularis39:
The PR should be now rebased and ready for merge.

ajor:
Thanks @lenticularis39!

tests/semantic_analyser.cpp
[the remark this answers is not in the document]
ajor: Added in #3062

PR #2629, opened 2023-05-26
New option to run specific runtime test suite: --filter-suite

dubeyabhishek:
This new option serves following purpose:
1. Run only feature-specific tests.
2. Quick runs save development time on specific features. Complete runs can be used for checking regressions.

Usage:
./runtime-tests.sh --filter-suite <suite_name>

<suite_name> can be a regex expression with/without quotes.

### DIFF
--- a/tests/runtime/engine/main.py
+++ b/tests/runtime/engine/main.py
@@ -9,9 +9,11 @@ from parser import TestParser, UnknownFieldError, RequiredFieldError
 from runner import Runner, ok, fail, warn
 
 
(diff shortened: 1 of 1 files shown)


dubeyabhishek:
@viktormalik @danobi Your thoughts for this patch, I am finding this useful.

dubeyabhishek:
> Is this currently possible with `./runtime-tests --filter "suite_name.*"` ?

Great catch! With working --filter now, no need for this. Thanks

Issue #2620, opened 2023-05-24
RFC: support all supported iter attaching type


chengshuyi:
### Is your feature request related to a problem? Please describe.

bpftrace currently supports iter:task, iter:task_file and iter:task_vma through hardcoding. There are other not yet supported, such as iter:tcp, iter:udp, etc.

### Describe the solution you'd like

We can read the iter type supported by the current kernel from btf like supporting kfunc. According to the [kernel documentation](https://www.kernel.org/doc/html/latest/bpf/bpf_iterators.html#how-to-use-bpf-iterators),  we should be able to find the currently supported iter type through the `bpf_iter__` prefix

> from kernel doc: Any struct name in vmlinux.h in the format bpf_iter__<iter_name> represents a BPF iterator. The suffix <iter_name> represents the type of iterator.

### Describe alternative solutions or features you've considered

not yet.

Any suggestions are welcome. Thanks!

chengshuyi:
done. #2630

PR #2630, opened 2023-05-30
support all iters

chengshuyi:
This pr mainly improves bpftrace's support for iter

- `bpftrace -l iter:*` print the iter supported by the current kernel
- Support the iter supported by the current kernel, no need to hard code

issues: #2620

##### Checklist


src/btf.cpp
```diff
@@ -714,6 +714,39 @@ std::set<std::string> BTF::get_all_structs() const
   return structs;
 }
 
+std::set<std::string> BTF::get_all_iters_from_btf(const struct btf *btf) const
```
danobi, line 717: Nit: use `std::unordered_set` for performance since we do not need ordering here

src/btf.cpp
```diff
@@ -714,6 +714,39 @@ std::set<std::string> BTF::get_all_structs() const
...
+    if (name.size() >= prefix.size() &&
+        name.compare(0, prefix.size(), prefix) == 0)
+    {
+      iter_set.insert(name.substr(prefix.size()));
+    }
+  }
+  return iter_set;
+}
+
+std::set<std::string> BTF::get_all_iters() const
```
danobi, line 739: Same as above

src/probe_matcher.cpp
```diff
@@ -162,7 +157,25 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
...
+        {
+          // second check
+          if (bpftrace_->feature_->has_iter(iter))
+          {
+            ret += iter + "\n";
+          }
+          else
+          {
+            LOG(WARNING) << "The kernel broke the 'bpf_iter__' rule, please "
+                            "report this bug to us";
```
danobi, line 174: Error message is a bit ambiguous. How about something like:

```
The kernel contains bpf_iter__$ITER struct but does not support loading an iterator program against it. Please report this bug.
```

where `$ITER` is the loop variable `iter`

src/probe_matcher.cpp
```diff
@@ -162,7 +157,25 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     }
     case ProbeType::iter:
     {
-      symbol_stream = get_iter_symbols();
+      if (bpftrace_->has_btf_data())
+      {
```
danobi, line 161: Nit: reduce indentation with an early break. ie:

```
if (!bpftrace_->has_btf_data())
  break;

...
```

danobi (commented):
Woah, didn't realize it would be this easy to support all iters. Lgtm overall, just some small nits. Great PR!

danobi:
Can we also delete https://github.com/iovisor/bpftrace/blob/446109acdef586fb0f0b5d40196dedb946b613ed/src/bpffeature.h#L111-L122 and https://github.com/iovisor/bpftrace/blob/446109acdef586fb0f0b5d40196dedb946b613ed/src/bpffeature.h#L57-L58 ?

src/btf.cpp
chengshuyi: fix.

src/probe_matcher.cpp
chengshuyi: Great.

chengshuyi:
> Can we also delete
> 
> https://github.com/iovisor/bpftrace/blob/446109acdef586fb0f0b5d40196dedb946b613ed/src/bpffeature.h#L111-L122
> 
> and
> https://github.com/iovisor/bpftrace/blob/446109acdef586fb0f0b5d40196dedb946b613ed/src/bpffeature.h#L57-L58
> 
> ?

OK, it has been removed. In addition, a `support to get all iter params` commit has been added, thanks for your review

src/btf.cpp
```diff
@@ -714,6 +714,40 @@ std::set<std::string> BTF::get_all_structs() const
...
+  __s32 id, max = (__s32)type_cnt(btf);
+  for (id = start_id(btf); id <= max; id++)
+  {
+    const struct btf_type *t = btf__type_by_id(btf, id);
+
+    if (!t || !(btf_is_struct(t)))
+      continue;
+
+    const std::string name = btf_str(btf, t->name_off);
+    if (name.size() >= prefix.size() &&
```
viktormalik, line 731: Although it shouldn't occur in BTF, unnamed iterator is not correct, right?
```suggestion
    if (name.size() > prefix.size() &&
```

viktormalik (approved):
Nice!

danobi (approved):

danobi:
Can you update the changelog please?

chengshuyi:
> Can you update the changelog please?

Done, thanks.

PR #2640, opened 2023-06-07
Make parallel map access test less flakey

ajor:
The new test introduced in #2623 was a little flakey at detecting the bad behaviour on my machine.

This change increases the frequency at which missing map keys will be encountered and makes the test consistently fail when the patch from #2623 is reverted. It still passes with the patch applied.


ajor:
Hmm, seems like this new frequency might be too high for the CI machines

danobi:
I think occasional failures are probably good enough. Race conditions are always tricky to reliably reproduce.

Issue #2285, opened 2022-06-30
`func` in `uretprobe` returns caller name


viktormalik:
Originally discovered by @xutao323.

Using the `func` builtin inside an `uretprobe` gives the name of the caller function. However, the documentation says:

> func - Name of the traced function

which suggests that the name of the callee (one that the `uretprobe` is attached to) should be returned.

The reason for this is that `func` reads the `ip` register which is at that time set to the return instruction in the caller function. This should be either fixed or at least marked in documentation.

### What reproduces the bug?

```
# cat test.c
#include <unistd.h>

int fun() { }

int main() {
    fun();
    sleep(1);  // this is necessary due to other bug, see issue #2284
}

# bpftrace -e 'uretprobe:test:fun { printf("%s\n", func); }'
Attaching 1 probe...
main
```

ajor:
Some discussion here: https://lore.kernel.org/bpf/ZKuyKj8jrEPzWYMM@krava/T/#u

viktormalik:
FYI, kernel now supports `get_func_ip` for uprobes (https://github.com/torvalds/linux/commit/a3c485a5d8d4)

xutao323:
Got a chance to revisit. I tried recent Fedora `kernel-6.7.0-0.rc4.35.fc40` with #2645, and this issue can be resolved.

My steps:

1. Install latest Fedora [kernel](https://src.fedoraproject.org/rpms/kernel): the `get_func_ip` change is in newer kernels after 6.6, I used [kernel-6.7.0-0.rc4.35.fc40](https://koji.fedoraproject.org/koji/buildinfo?buildID=2327764)
2. Build #2645 locally: a small change to pass `get_func_ip` for both uprobe and kprobe on top of https://github.com/ajor/bpftrace2/commit/b8a14f44051ce4ece7e34b0062ca9be58e3ab9ea (test still fails, i don't know):
```
diff --git a/src/ast/passes/codegen_llvm.cpp b/src/ast/passes/codegen_llvm.cpp
index 92ea1fe7..db288ab9 100644
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -278,7 +278,7 @@ void CodegenLLVM::visit(Builtin &builtin)
     // But only use the it for kretprobes, as the helper will fail for kprobes
     // placed within a function (not at the entry).
     auto probe_type = probetype(current_attach_point_->provider);
[...]

PR #2645, opened 2023-06-16
Fix func builtin in kretprobes and uretprobes

### LINKED ISSUE #68: func only matches kretprobe_trampoline on kretprobes
```
# bpftrace -e 'kretprobe:vfs_* { @[func] = count(); }'
Attaching 53 probes...
Warning: could not attach probe kretprobe:vfs_dedupe_get_page.isra.21, skipping.
^C

@[kretprobe_trampoline]: 438
```

Would be nice if `func` could match the function on kretprobe, intsead of only ever returning `kretprobe_trampoline`. If that's not possible, it should be documented in the reference guide.

ajor:
For ret probes, the IP register will not be pointing to the function we want to trace. It may point to a kernel trampoline, or it may point to the caller of the traced function, as it fires after the "ret" instruction has executed.
    
Switch to using the get_func_ip BPF helper in [ku]retprobes.
    
**Disable func in [ku]retprobes** when this helper is not available, as it previously returned incorrect results.

This closes #68 and #2285.

Old behaviour:
```
# bpftrace -e 'kretprobe:vfs_read { printf("func: '%s'\n", func); exit(); }' 
Attaching 1 probe...
func: ksys_read
```

New behaviour:
```
# bpftrace -e 'kretprobe:vfs_read { printf("func: '%s'\n", func); exit(); }' 
Attaching 1 probe...
func: vfs_read
```

##### Checklist


lenticularis39:
> @lenticularis39 Do you have any idea why this test failed in CI?
> 
> ```
> 2: [  FAILED  ] builtin.func_uprobe
> 2: 	Command: ../src//bpftrace -e 'uprobe:./testprogs/uprobe_negative_retval:function1 { printf("func: \"%s\"\n", func); exit(); }'
> 2: 	Expected: func: "function1"
> 2: 	Found: Attaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\nfunc: "0x400492"\n\n\n
> ```
> 
> I assume the target process must have exited before we got a chance to symbolise, but didn't your recent changes handle this case?

My fix in #2386 is not for the general case (that is why the PR doesn't close the associated issues), it is for cases where either:
- ASLR is disabled, or
- the process is running at the time of *attaching* the probe

Neither of these is true for `func_uprobe`: the process is run *after* attaching the probe, meaning that bpftrace doesn't get a chance to pre-load the correct symbol. If you change the `after` to `before` (and put a time delay into the test program), you should get the correct output most of the time (sometimes it still fails due to the bcc race condition I mentioned in https://github.com/iovisor/bpftrace/pull/2386#discussion_r1196056779).

This is also documented in the manpage: "If ASLR is enabled, user space symbolication only works when the process is running at either the time of the symbol resolution or the time of the probe attachment."

ajor:
Doh, thank you!

tests/runtime/builtin
```diff
@@ -96,18 +96,32 @@ EXPECT SUCCESS .*
...
-NAME func
-PROG k:vfs_read { printf("SUCCESS %s\n", func); exit(); }
-EXPECT SUCCESS .*
+NAME func_kprobe
+PROG k:vfs_read { printf("func: \"%s\"\n", func); exit(); }
+EXPECT func: "vfs_read"
+TIMEOUT 5
+AFTER ./testprogs/syscall read
+
+NAME func_kretprobe
```
viktormalik, line 105: Shouldn't this test have a requirement for the `get_func_ip` helper?

danobi: Yeah if you add a requirement on the feature then we don't need to disable it. It'll only run if the host supports the correct version.

ajor:
> Also closes #2285, right?

I suppose it does fix the incorrect results described in this ticket, but I'd still like to get func working properly for uretprobes instead of just disabling it.

ajor:
Added REQUIRES_FEATURE to runtime test and hopefully fixed the symbol resolution test issue

danobi:
Looks like CI still failing

ajor:
The symbolication issues must be the PID-namespacing problem (the tests run in Docker)! I was going crazy trying to understand what I was doing wrong.

Looks like we didn't have any PID-based symbolication tests that actually checked the results before this.

viktormalik:
FWIW, kernel now supports `bpf_get_func_ip` for uretprobes:

https://git.kernel.org/bpf/bpf-next/c/a3c485a5d8d4

jordalgo:
@ajor - do you know what other work is needed here to get this merged?

ajor:
> @ajor - do you know what other work is needed here to get this merged?

Yes, sorry - I've got another PR in the works which adds support for PID-namespaces and is required to support these tests in CI. It just requires a little cleaning up before I submit. I'll try and get through my development backlog in January, starting next week.

src/bpffeature.h
```diff
@@ -123,7 +123,7 @@ class BPFfeature {
   DEFINE_HELPER_TEST(probe_read_kernel_str, libbpf::BPF_PROG_TYPE_KPROBE);
   DEFINE_HELPER_TEST(ktime_get_boot_ns, libbpf::BPF_PROG_TYPE_KPROBE);
   DEFINE_HELPER_TEST(ktime_get_tai_ns, libbpf::BPF_PROG_TYPE_KPROBE);
-  DEFINE_HELPER_TEST(get_func_ip, libbpf::BPF_PROG_TYPE_TRACING);
+  DEFINE_HELPER_TEST(get_func_ip, libbpf::BPF_PROG_TYPE_KPROBE);
```
ajor, line 126: get_func_ip for kprobes was added 1 commit after it was added for tracing programs, so this is a stricter test, but realistically won't make a difference

tests/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ if(${LLVM_VERSION_MAJOR} VERSION_EQUAL 12)
   set(CODEGEN_SRC ${CMAKE_BINARY_DIR}/tests/codegen_includes.cpp)
 else()
   set(CODEGEN_SRC "")
-  message(STATUS "Disabled codegen test for LLVM != 12")
+  message(STATUS "Disabled codegen test for LLVM != 12 (found ${LLVM_VERSION_MAJOR})")
```
ajor, line 17: Drive-by fix to help me debug my environment

src/ast/passes/semantic_analyser.cpp
```diff
@@ -346,6 +346,14 @@ void SemanticAnalyser::visit(Builtin &builtin)
         LOG(ERROR, builtin.loc, err_)
             << "The func builtin can not be used with '"
             << attach_point->provider << "' probes";
+
+      if ((type == ProbeType::kretprobe || type == ProbeType::uretprobe) &&
+          !bpftrace_.feature_->has_helper_get_func_ip()) {
+        LOG(ERROR, builtin.loc, err_)
+            << "The 'func' builtin is not available for " << type
+            << "s on kernels without the get_func_ip BPF feature. Consider "
+               "using 'probe' instead.";
```
jordalgo, line 355: nit: can we say either 'kprobe' or 'uprobe' depending on the context ('probe' by itself isn't a thing) ?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -327,10 +327,37 @@ void CodegenLLVM::visit(Builtin &builtin)
...
+    // the get_func_ip helper to get the instruction pointer.
+    //
+    // For [ku]retprobes, the IP register will not be pointing to the function
+    // we want to trace. It may point to a kernel trampoline, or it may point to
+    // the caller of the traced function, as it fires after the "ret"
+    // instruction has executed.
+    //
+    // The get_func_ip helper resolves these issues for us.
+    //
+    // But do not use the it for non-ret [ku]probes, as the helper will fail for
```
jordalgo, line 341: Curious, do you mean if folks pass an offset or if the function specified has been inlined?

jordalgo (commented):
One small nit but otherwise LGTM!

src/ast/passes/semantic_analyser.cpp
viktormalik: I think it means the `probe` builtin ;-)

viktormalik (approved):
LGTM!

src/ast/passes/semantic_analyser.cpp
ajor: Hopefully it's clearer now!

jordalgo (approved):

PR #2652, opened 2023-06-22
Fix one-liners args access

Wv5twkFEKh54vo4tta9yu7dHa3:
These two one-liners were accessing the args struct members without dereferencing the pointer first, causing errors like

`stdin:1:39-49: ERROR: Can not access field 'bytes' on type '(ctx) struct _tracepoint_block_block_rq_issue *'. Try dereferencing it first, or using '->'`

##### Checklist

There is no new behaviour, this is just documentation


mmarchini (approved):

viktormalik:
This is not a bug, it's a feature. Since #2578, it is possible and recommended to treat `args` as a structure instead of a pointer. The errors that you observed cannot be reproduced on the latest master, you were probably trying the one-liners with the system installed bpftrace.

@mmarchini are you ok if we revert this patch?

mmarchini:
oh yeah, i was using an older version of bpftrace when i tested it. my bad, #2655 will revert it

PR #2657, opened 2023-06-27
uprobes: make C++ symbol demangling explicit

### LINKED ISSUE #2186: Mangled symbols overlap cannot be disambiguated
I see in #1116 that mangled symbols are now supported as probe names. However, this causes issues and unexpected behavior if there is overlap in mangled/non-mangled (C) function names. (Causes rubrikinc/wachy#6)

### What reproduces the bug?

```
~/Developer/rust/wachy/demo$ cat test.cpp
bool main_test() {
  return false;
}

int main() {
  return 0;
}
~/Developer/rust/wachy/demo$ c++ test.cpp
~/Developer/rust/wachy/demo$ nm -g a.out
0000000000201010 B __bss_start
                 w __cxa_finalize@@GLIBC_2.2.5
0000000000201000 D __data_start
0000000000201000 W data_start
0000000000201008 D __dso_handle
0000000000201010 D _edata
0000000000201018 B _end
0000000000000684 T _fini
                 w __gmon_start__
00000000000004b8 T _init
0000000000000690 R _IO_stdin_used
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
0000000000000680 T __libc_csu_fini
0000000000000610 T __libc_csu_init
                 U __libc_start_main@@GLIBC_2.2.5
0000000000000605 T main
00000000000004f0 T _start
0000000000201010 D __TMC_END__
00000000000005fa T _Z9main_testv
~/Developer/rust/wachy/demo$ sudo bpftrace -e 'uprobe:./a.out:main {}'
Attaching 2 probes...
^C

~/Developer/rust/wachy/demo$ sudo bpftrace -e 'uprobe:./a.out:"main()" {}'
No probes to attach
```

I don't see any way to specify that I just want to trace `main`, not `main_test`. `BPFTRACE_NO_CPP_DEMANGLE` doesn't help - it seems to affect the output, not input command line.

Is there some env var/probe option to disable mangled support?

### bpftrace --info
```
System
  OS: Linux 4.15.0-147-generic #151-Ubuntu SMP Fri Jun 18 19:21:19 UTC 2021
  Arch: x86_64

Build
  version: v0.14.1
  LLVM: 12.0.0
  ORC: v2
  foreach_sym: yes
  unsafe uprobe: no
  bfd: yes
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  libbpf: yes
  libbpf btf dump: yes
  libdw (DWARF support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: no
  probe_read_user_str: no
  probe_read_kernel: no
  probe_read_kernel_str: no
  get_current_cgroup_id: no
  send_signal: no
  override_return: no
  get_boot_ns: no
  dpath: no

Kernel features
  Instruction limit: -1
  Loop support: no
  btf (depends on Build:libbpf): no
  map batch (depends on Build:libbpf): no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
  iter:task: no
  iter:task_file: no
```

viktormalik:
Introduce a new part of uprobe specification which allows to explicitly enable language-specific features. For now, C++ prefixes (`:c++` and `:cpp`) are supported. Specifying one of these turns on symbol demangling which is now disabled by default.

Some examples of the current behaviour
```
# cat cpptest.cpp
extern "C" int fun() { return 1; }
void fun(int x) {}
void function() {}
class Foo { void fun() {} };
class Bar { void fun() {} };

# nm cpptest
0000000000401126 T fun
0000000000401148 T _Z3funi
0000000000401131 T _Z8functionv
0000000000401184 W _ZN3Bar3funEv
0000000000401178 W _ZN3Foo3funEv

# 
bpftrace -e 'u:./cpptest:fun { exit(); }'
Attaching 1 probe...    (C "fun" only)

bpftrace -e  'u:./cpptest:c++:fun { exit(); }
Attaching 3 probes...   ("fun", "fun(int)", "function")

bpftrace -e 'u:./cpptest:c++"fun(int)" { exit(); }' 
Attaching 1 probe...    ("fun(int)")

bpftrace -e 'u:./cpptest:c++:*fun { exit(); }'
Attaching 5 probes...   (all the functions)

bpftrace -e 'u:./cpptest:c++:"*::fun" { exit(); }'
Attaching 2 probes...   ("Foo::fun" and "Bar::fun")
```

While this is not perfect (e.g. the second example should not match `function`), it's better than what it was before (see #2186). I'd prefer using this as a starting point and then following with more PRs rather than doing everything at once.

Missing features include:
- more sophisticated matching of mangled symbols (i.e. fix the above 2nd example)
- better support for namespaces (e.g. support `::fun`)
- support for demangling in probe listing

Note that this is a breaking change b/c demangling is not done by default now. However, since it was never documented and it sometimes had incorrect and unexpected behaviour, I believe that this is the correct step forward.

Resolves #2186.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1148,6 +1148,14 @@ Allocated 4 bytes
 ...
 ```
 
+When tracing C++ programs, it is possible to turn on automatic symbol demangling
+by using the `:cpp` prefix:
+```
+# bpftrace -e 'u:src/bpftrace:cpp:"bpftrace::BPFtrace::add_probe" { print("adding probe\n"); }'
+Attaching 1 probe...
+adding probe
+```
+
 Examples in situ:
 [(uprobe) search /tools](https://github.com/iovisor/bpftrace/search?q=uprobe%3A+path%3Atools&type=Code)
 [(uretprobe) /tools](https://github.com/iovisor/bpftrace/search?q=uretprobe%3A+path%3Atools&type=Code)
--- a/src/probe_matcher.cpp
+++ b/src/probe_matcher.cpp
@@ -101,7 +101,8 @@ std::set<std::string> ProbeMatcher::get_matches_in_stream(
 std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     const ProbeType& probe_type,
     const std::string& target,
-    const std::string& search_input)
+    const std::string& search_input,
+    bool demangle_symbols)
 {
   std::unique_ptr<std::istream> symbol_stream;
   bool ignore_trailing_module = false;
@@ -189,9 +190,8 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
   }
 
   if (symbol_stream)
-    return get_matches_in_stream(search_input,
-                                 *symbol_stream,
-                                 ignore_trailing_module);
+    return get_matches_in_stream(
+        search_input, *symbol_stream, ignore_trailing_module, demangle_symbols);
   else
     return {};
 }
@@ -571,7 +571,8 @@ std::set<std::string> ProbeMatcher::get_matches_for_ap(
 
   return get_matches_for_probetype(probetype(attach_point.provider),
                                    attach_point.target,
-                                   search_input);
+                                   search_input,
+                                   attach_point.lang == "cpp");
 }
 
 std::set<std::string> ProbeMatcher::expand_probetype_kernel(
--- a/src/probe_matcher.cpp
+++ b/src/probe_matcher.cpp
@@ -67,17 +67,31 @@ std::set<std::string> ProbeMatcher::get_matches_in_stream(
         if (symbol_has_cpp_mangled_signature(fun_line))
         {
           char* demangled_name = cxxdemangle(fun_line.c_str());
-          if (demangled_name)
+          if (!demangled_name)
+            continue;
+
+          // Match against the demanled name.
+          // Since demangled_name contains function arguments, we need to remove
+          // them unless the user specified '(' in the search input (i.e. wants
+          // to match against the arguments explicitly).
+          std::string match_line = prefix + demangled_name;
+          if (std::all_of(tokens.begin(),
+                          tokens.end(),
+                          [&](const std::string& token) {
+                            return token.find("(") == std::string::npos;
+                          }))
           {
-            if (!wildcard_match(prefix + demangled_name, tokens, true, true))
-            {
-              free(demangled_name);
-            }
-            else
-            {
-              free(demangled_name);
-              goto out;
-            }
+            match_line = match_line.substr(0, match_line.find_last_of("("));
+          }
+
+          if (!wildcard_match(match_line, tokens, start_wildcard, end_wildcard))
+          {
+            free(demangled_name);
+          }
+          else
+          {
+            free(demangled_name);
+            goto out;
           }
         }
       }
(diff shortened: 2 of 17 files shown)


mmarchini:
overall LGTM, definitely a breaking change as you said (but that seems fine in this case). Have you looked into how much work it would be to fix the 2nd example? That's the only missing thing I think would be worth including in this PR if the fix is not a large patch.

viktormalik:
> Have you looked into how much work it would be to fix the 2nd example? That's the only missing thing I think would be worth including in this PR if the fix is not a large patch.

For this particular case, it's as simple as adding an extra `(` to the end of the query. There are other cases, though, that'll require more thinking. Let me have a look, if we manage to review and merge this in the meantime, I'll send the fix as a follow-up PR ASAP.

viktormalik:
> > Have you looked into how much work it would be to fix the 2nd example? That's the only missing thing I think would be worth including in this PR if the fix is not a large patch.
> 
> For this particular case, it's as simple as adding an extra `(` to the end of the query. There are other cases, though, that'll require more thinking. Let me have a look, if we manage to review and merge this in the meantime, I'll send the fix as a follow-up PR ASAP.

It turned out not to be too complicated to fix the bugs that I know of, so I updated this PR (see 17beb2c).

src/probe_matcher.cpp
```diff
@@ -69,7 +69,22 @@ std::set<std::string> ProbeMatcher::get_matches_in_stream(
...
+                                             cpp_tokens.end(),
+                                             [&](const std::string& token) {
+                                               return token.find("(") ==
+                                                      std::string::npos;
+                                             }))
+            {
+              cpp_tokens.back() += "(";
+            }
+            if (!wildcard_match(
+                    prefix + demangled_name, cpp_tokens, true, true))
```
danobi, line 87: Feels kinda odd to add a `(` to the end of the search string. Why not remove the `(args...)` from the symbol instead?

src/parser.yy
```diff
@@ -280,6 +280,7 @@ attach_point_def:
         |       attach_point_def MUL      { $$ = $1 + "*"; }
         |       attach_point_def LBRACKET { $$ = $1 + "["; }
         |       attach_point_def RBRACKET { $$ = $1 + "]"; }
+        |       attach_point_def INCREMENT { $$ = $1 + "++"; }
```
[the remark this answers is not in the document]
viktormalik: Ok, makes sense, I'm ok with sticking to `cpp` only.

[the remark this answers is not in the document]
viktormalik: Done, only `:cpp` is supported now.

src/probe_matcher.cpp
viktormalik: Good idea, it's even better as it allows to get rid of the artificially added leading and trailing whitespace which may potentially lead to incorrect matches.

docs/reference_guide.md
```diff
@@ -1,4 +1,3 @@
-# bpftrace Reference Guide
```
danobi, line 1: Accidental deletion?

danobi (approved):

docs/reference_guide.md
viktormalik: Yeah :-( Fixed, thanks.

PR #2660, opened 2023-07-02
Support kprobe:module:function

zhangjaycee:
As mentioned in the issue #1413 ，the kprobe type currently does not support selecting kernel modules.

This draft MR supports it by allowing the 'kprobe:module:function' syntax with little code changes. Note that wildcard is not supported yet in this MR. I hope anyone who is interested in this feature can make use of this draft patch, or even improve it and make it mergeable in the future.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -935,11 +935,15 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
     return;
   }
 
+  std::string funcname = probe_.attach_point;
+  if (probe_.path.length() > 0)
+    funcname = probe_.path + ":" + funcname;
+
   resolve_offset_kprobe(safe_mode);
   int perf_event_fd = bpf_attach_kprobe(progfd_,
                                         attachtype(probe_.type),
                                         eventname().c_str(),
-                                        probe_.attach_point.c_str(),
+                                        funcname.c_str(),
                                         offset_,
                                         0);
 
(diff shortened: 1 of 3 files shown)


src/attached_probe.cpp
```diff
@@ -935,11 +935,15 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  std::string funcname = probe_.attach_point;
+  if (probe_.path.length() > 0)
+    funcname = probe_.path + ":" + funcname;
+
   resolve_offset_kprobe(safe_mode);
   int perf_event_fd = bpf_attach_kprobe(progfd_,
                                         attachtype(probe_.type),
                                         eventname().c_str(),
-                                        probe_.attach_point.c_str(),
+                                        funcname.c_str(),
```
viktormalik, line 946: I don't think that this is going to work. BCC expects a kprobe name here and likely won't handle `module:function` syntax.

zhangjaycee: IIUC, the kernel interface [1] accepts the name with the [MOD:] prefix. And I tested this patch it works. Thanks!

[1] https://www.kernel.org/doc/Documentation/trace/kprobetrace.rst

viktormalik: Nice! Would you be able to finish the PR, then? It's just missing a couple of things, mainly documentation and tests.

danobi:
Tracking in #2826

PR #2662, opened 2023-07-04
Allow uprobes to work across filesystem namespace boundaries

ivankelly:
When attaching to a uprobe, bpftrace will try to cache symbols for all pids using that program. It does this by canonicalizing the executable path and scanning /proc for processes whose /exe matches.

For containers, this will not work as the canonicalize command will resolve ```/proc/123/root/foo/bar``` to ```/foo/bar``` in the host filesystem and then error trying to canonicalize that because it doesn't exist on the host filesystem.

The proposed solution is, when getting the pids for a program, if the path is of the form ```/proc/<pid>/<root>/<path>```, don't bother with the search and just extract the pid from the path itself.

##### Checklist

- [n/a] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`


tests/CMakeLists.txt
```diff
@@ -124,7 +124,13 @@ endif(NOT STATIC_LINKING)
 # Compile all testprograms, one per .c file for runtime testing
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/testprogs/)
 file(GLOB testprogs testprogs/*.c)
+list(FILTER testprogs EXCLUDE REGEX "uprobe_test_static\\.c")
 set(compiled_testprogs "")
+
+add_executable (uprobe_test_static testprogs/uprobe_test_static.c)
+set_target_properties(uprobe_test_static PROPERTIES LINK_SEARCH_START_STATIC FALSE LINK_SEARCH_END_STATIC FALSE RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/testprogs/ COMPILE_FLAGS "-g -O0" LINK_FLAGS "-no-pie -static")
```
viktormalik, line 131: IIUC, the only difference for `uprobe_test_static` is the `-static` link flag. Perhaps we could just do:
```cmake
if (${testprog} STREQUAL "uprobe_test_static")
    set_target_properties(${bin_name} PROPERTIES LINK_FLAGS "-no-pie -static")
endif()
```
and avoid all other special treatment of the testprog which contains a lot of duplicated lines.

tests/runtime/uprobe
```diff
@@ -52,6 +52,16 @@ EXPECT Attaching 1 probe...
 TIMEOUT 5
 BEFORE ./testprogs/mountns_wrapper uprobe_test
 
+# Note that we don't expect the exact function1 name because CI runs in a docker
+# container and symbols don't currently resolved correctly in this scenario
```
viktormalik, line 56: nit
```suggestion
# container and symbols don't currently get resolved correctly in this scenario
```

tests/runtime/uprobe
```diff
@@ -52,6 +52,16 @@ EXPECT Attaching 1 probe...
...
+# as the event received by bpftrace contains the pid of the process in the root
+# pid namespace, not that of the docker container's namespace.
+NAME uprobes - attach to probe for executable in a pivot_root'd mount namespace
+RUN {{BPFTRACE}} -e 'uprobe:/proc/{{BEFORE_PID}}/root/uprobe_test_static:function1 { printf("func %s\n", func); exit(); }'
+EXPECT func [0-9a-fx]*
```
viktormalik, line 61: perhaps better regex:
```suggestion
EXPECT func 0x[0-9a-f]\+
```

tests/runtime/uprobe
```diff
@@ -52,6 +52,16 @@ EXPECT Attaching 1 probe...
 TIMEOUT 5
 BEFORE ./testprogs/mountns_wrapper uprobe_test
 
+# Note that we don't expect the exact function1 name because CI runs in a docker
```
viktormalik, line 55: Does this mean that when this is not running in a container, this test will fail?

tests/testprogs/mountns_pivot_wrapper.c
```diff
@@ -0,0 +1,88 @@
...
+
+  // Recursively set the mount namespace to private, so caller can't see
+  if (mount(NULL, "/", NULL, MS_REC | MS_PRIVATE, NULL) != 0)
+    errExit("Failed to make mount private");
+
+  // make a tempdir and bind mount containing testprog folder to it
+  if (mkdir(private_mount, 0770) != 0 && (errno != EEXIST))
+    errExit("Failed to make private mount dir");
+
+
```
viktormalik, line 65: nit: extra blank line

tests/runtime/engine/runner.py
```diff
@@ -159,14 +159,16 @@ def run_test(test):
...
         bpftrace = None
         after = None
         cleanup = None
+        bpf_call = None
         try:
             result = None
             timeout = False
             output = ""
-
+
```
viktormalik, line 171: nit: superfluous whitespace

tests/runtime/engine/runner.py
```diff
@@ -320,11 +327,20 @@ def run_test(test):
...
+                    err = err.decode("utf-8")
+                    print(f"stdout: {out}\nstderr stderr: {err}")
+
         if p and p.returncode != 0 and not test.will_fail and not timeout:
             print(fail("[  FAILED  ] ") + "%s.%s" % (test.suite, test.name))
-            print('\tCommand: ' + bpf_call)
+            print('\tCommand: ' + (bpf_call or "unset"))
             print('\tUnclean exit code: ' + str(p.returncode))
             print('\tOutput: ' + output.encode("unicode_escape").decode("utf-8"))
+            print_befores_output()
```
viktormalik, line 343: This looks like a useful feature which is not exactly specific to this PR. Would you mind moving it into a separate PR, preferably with something similar for the `AFTER` directive?

viktormalik (commented):
Thanks! Looks ok to me, just a couple of nits and suggestions, mainly for the tests.

@lenticularis39 mind having a quick look?

src/utils.cpp
```diff
@@ -1370,17 +1370,29 @@ std::map<uintptr_t, elf_symbol, std::greater<>> get_symbol_table_for_elf(
...
   {
-    std::string filename = process.path().filename().string();
-    if (!std::all_of(filename.begin(), filename.end(), ::isdigit))
-      continue;
-    std::error_code ec;
-    std_filesystem::path pid_program = std_filesystem::read_symlink(
-        process.path() / "exe", ec);
-    if (!ec && program_abs == pid_program)
-      pids.emplace_back(std::stoi(filename));
+    pids.emplace_back(std::stoi(match.str(1)));
```
lenticularis39, line 1379: I'm not sure if this is the right idea, I can trace e.g. `/proc/123/root/bin/bash`, and 123 does not have to be the ID of the bash process, it can be e.g. the root of the namespace or another process in the namespace. Since the cache is loaded by PID, it won't cause incorrect symbol resolution, but perhaps a smarter solution could be implemented (see comment below).

src/utils.cpp
```diff
@@ -1370,17 +1370,29 @@ std::map<uintptr_t, elf_symbol, std::greater<>> get_symbol_table_for_elf(
...
+  {
+    auto program_abs = std_filesystem::canonical(program);
+
+    for (const auto &process : std_filesystem::directory_iterator("/proc"))
+    {
+      std::string filename = process.path().filename().string();
+      if (!std::all_of(filename.begin(), filename.end(), ::isdigit))
+        continue;
+      std::error_code ec;
+      std_filesystem::path pid_program = std_filesystem::read_symlink(
```
lenticularis39, line 1391: Idea: I am not sure what this does for programs which are in a different mount namespace. Maybe also trying `std_filesystem::absolute` next to `std_filesystem::canonical` could work.

tests/runtime/uprobe
ivankelly: No, it will pass if the regex is left as is, because if symbol resolution works, as it would when not running in a container, then the output would be function1, which ```[0-9a-fx]*``` matches.

viktormalik: Well, technically it does, but only b/c it's not matching the entire line and a single letter is sufficient. I'd prefer something more robust like `^(0x[0-9a-f]+|function1)$` (you may need to escape some characters).

src/utils.cpp
ivankelly: I don't think absolute is correct either, as the absolute path would simply be the path passed in. Perhaps weakly_canonical would work though. I'll give it a go.

ivankelly: nope
```
[  FAILED  ] uprobe.uprobes - attach to probe for executable in a pivot_root'd mount namespace
        Command: ../src//bpftrace -e 'uprobe:/proc/20424/root/uprobe_test_static:function1 { printf("func %s\n", func); exit(); }'
        Unclean exit code: 1
        Output: ERROR: failed to write elf: filesystem error: cannot make canonical path: No such file or directory [/proc/20424/root/uprobe_test_static]\n
        ```
Which is strange, because that path does exist, but maybe the problem is that its reading across a filesystem boundary or something.

A solution could be to just catch the exception and return no pids if it gets a file not found. This method is only used to cache symbols for pids at probe attach time. if it doesn't cache anything, that's no big whoop, they can be resolved at output time.

ivankelly:
@viktormalik @lenticularis39 Addressed comments. Now get_pids_for_program just bails if it cannot get the canonical path.

viktormalik (approved):
Sorry for a late review, looks good to me now, thanks!

Could you please fix the formatting issues reported by clang-format and resolve the merge conflict in CHANGELOG? Thank you!

ajor:
Closing in favour of #2802

PR #2663, opened 2023-07-04
Reduce size of SizedType class

ajor:
- Store all sizes internally as bits, converting back to bytes as needed
- Delete unused field num_elements_
- Shrink StackType::limit
- Shrink enum underlying types
- Reorder fields from largest to smallest (roughly)

Before:
```
sizeof(SizedType) = 136
```

After:
```
sizeof(SizedType) = 96
```

SizedType objects are created and copied a lot during compilation, so reducing the size should help with performance.


danobi (approved):
nice!

PR #2664, opened 2023-07-04
AST: Various cleanups

ajor:
I'll wait for #2624 to go into master before I merge this, as there'll be conflicts.

- Use move ctors for std::string and SizedType
- Remove explicit keywords on multi-arg ctors (they were a hold-over from when the ctors were single-argument, before locations were added)
- Use in-class initialisers instead of member initialisers for constants

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -226,14 +226,14 @@ Integer::Integer(int64_t n, location loc) : Expression(loc), n(n)
   is_literal = true;
 }
 
-String::String(const std::string &str, location loc) : Expression(loc), str(str)
+String::String(std::string str, location loc)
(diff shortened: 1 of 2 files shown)


src/ast/ast.cpp
```diff
@@ -633,7 +618,7 @@ Map::Map(const Map &other) : Expression(other)
 }
 
 FieldAccess::FieldAccess(const FieldAccess &other)
-    : Expression(other), expr(nullptr)
+    : Expression(other) // TODO why not copy expr??
```
ajor, line 621: @fbs Do you remember if `expr` wasn't copied for a reason?

fbs: nope :(

danobi:
closing due to inactivity

PR #2665, opened 2023-07-05
docs: Fix typo.

mofelee:
CleanShot 2023-07-05 at 17 07 22@2x


ajor:
I appreciate the effort to keep the docs accurate, but in this case it's an intentional change. In #2578 we converted `args` from a pointer to a struct.

The docs reflect the most up-to-date version of bpftrace, but it'll take a while for this version to roll out to all Linux distributions.

This isn't the first time someone has spotted this discrepancy: #2652. I suppose we might need to start releasing versioned documentation for bpftrace to avoid this problem in the future.

PR #2667, opened 2023-07-06
Adjust to build with llvm 17

kraj:
PassManagerBuilder has been removed in llvm 17
itaniumDemangle() API signature has changed too

##### Checklist


danobi (commented):
lgtm, but maybe we should wait until llvm17 is released so we can add CI support at the same time

viktormalik (changes requested):
This will also need to bump the bcc submodule to include at least https://github.com/iovisor/bcc/commit/594ab9a9b0e3dfa7cf8fc7f28923ba0eb33a5c66. Probably libbpf will have to be updated, too.

viktormalik:
Hm, LLVM 17 was already branched and is now in qualification phase, so it has a separate [APT package](https://apt.llvm.org/). IOW, we can already run LLVM 17 in CI even though it hasn't been released, yet. I can see 2 advantages in doing so: (1) we allow building bpftrace in staging environments and (2) we catch problems early. The disadvantage is that CI may "randomly" fail if the release candidate breaks something.

@danobi @fbs @ajor WDYT?

danobi:
Putting it in early and removing the job if it gets flaky sounds reasonable to me

viktormalik:
@kraj any chance you'd have time to follow up on this?

Rtoax:
Seems you forgot modify LLVM MAJOR [CMakeLists.txt](https://github.com/iovisor/bpftrace/pull/2755/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a)

viktormalik:
LLVM 17 release date is approaching, let me pick this up.

viktormalik:
So, unfortunately, LLVM 17 is not yet in [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs), yet, so we don't have a way to run it with the CI. I guess that we'll have to wait for the official release, then.

lenticularis39:
Note: Fedora 39 release date is approaching (October 17, 2023), which comes with LLVM 17, it would be ideal if we could get it merged before that.

danobi:
It looks like it might be a bit until nix packages llvm17: https://github.com/NixOS/nixpkgs/issues/257654

I would be open to merging this and keeping an eye out for nixos package

viktormalik:
Agreed, doesn't seem like we have much choice.

I'd even consider releasing 0.19.1 to make the distro packagers work easier.

viktormalik (approved):

danobi:
That sounds fine to me

danobi (approved):

PR #2669, opened 2023-07-07
arm: Determine target arch at runtime

### LINKED ISSUE #2654: check for 32-bit in src/arch/CMakeLists.txt seems wrong
Trying to build on Ubuntu armhf with an Ubuntu arm64 kernel fails, with this error:

  Unsupported architecture: armv7l

It looks like the only value of CMAKE_SYSTEM_PROCESSOR (which is the same as `uname -m` ttbomk) that is accepted for 32-bit ARM is armv7-a. But unless I am missing something (certainly possible), the possible values of `uname -m` on 32-bit arm are "arm", "armv7l" and "armv8l" (well and armv7b and armv8b for the linux32 personality running on big endian aarch64 but well). So I think the check should probably be changed.

It's possible I'm confused about all this though.

tnovak:
As of 9d9cb58c9 ("Add helper methods to obtain pointer width of traced code") we're already checking if the kernel is 32-bit, and this can be extended to figuring out the list of available registers, the layout of `struct pt_regs` etc. dynamically rather than having separate armv7 and aarch64 implementations selected at compile time.

This makes it possible to use the same 32-bit build of bpftrace on both arm and arm64 kernels.

Fixes #2654

##### Checklist


src/arch/utils.cpp
```diff
@@ -0,0 +1,27 @@
...
+  // processes (e.g. "armv8l" instead of "aarch64"; see COMPAT_UTS_MACHINE), so
+  // make sure this is taken into account.
+  struct utsname utsname;
+  if (uname(&utsname) >= 0)
+  {
+    if (!strncmp(utsname.machine, "armv", 4) && utsname.machine[4] < '8')
+      return 32;
+  }
+  return 64;
+}
```
danobi, line 24: Should this be defined in the per-arch implementations instead? This function seems to be making a global assumption that the only 32-bit arch support we have is on arm (which is technically true for now).

Seems like it'd be cleaner to do it per-arch and have the 64-bit archs just return 64. So that `get_kernel_ptr_width()` is implemented like the rest of the helpers declared in `arch.h`.

tnovak: Good idea, will update in a bit!

viktormalik (approved):

danobi (approved):

PR #2674, opened 2023-07-12
RFC: Use correct format string length modifiers when printing integers

tnovak:
Integer arguments to printf-style functions are represented as 64-bit ints, both in generated BPF code [1] and in bpftrace-side printing logic [2]. Therefore, a common pattern such as `printf("%d", pid)` is buggy -- it uses the conversion specifier for ints to print an int64_t value.

While in practice this works as expected on most platforms (the value is truncated to 4 bytes), it's by accident -- it depends on the architecture, toolchain, calling convention, etc. For example, the above breaks on arm with gcc where the compiler emits the following code for PrintableInt:

```
   0x000d6240 <+8>:     vldr    d7, [r12, #8]    # this->value_
   [...]
   0x000d6250 <+24>:    vstr    d7, [sp]
   0x000d6254 <+28>:    mov     r2, r3           # fmt
   0x000d6258 <+32>:    bl      0x1e000 <snprintf@plt>
```

(Note that the first 3 arguments to snprintf are passed via r0-r2, but `value_` is pushed onto the stack; `snprintf` expects the "%d" argument to be in r3, so it ends up printing the value of `fmt` instead)

With this patch, bpftrace internally replaces int format specifiers with their 64-bit variants (e.g. %x -> %llx on arm) in order to match int64_t values passed to snprintf.

[1] https://github.com/iovisor/bpftrace/blob/34cf5c2b53b82cddc9111055cd499d6e35763f3c/src/ast/passes/codegen_llvm.cpp#L3167
[2] https://github.com/iovisor/bpftrace/blob/34cf5c2b53b82cddc9111055cd499d6e35763f3c/src/printf.h#L82

##### Checklist

### DIFF
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -63,21 +63,21 @@ TIMEOUT 5
 AFTER ./testprogs/syscall read
 
 NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
+PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", (int32)sarg0, (int32)sarg1); exit(); }
 EXPECT SUCCESS 32 64
 ARCH x86_64
 TIMEOUT 5
 AFTER ./testprogs/stack_args
 
 NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
+PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", (int32)sarg0, (int32)sarg1); exit(); }
 EXPECT SUCCESS 128 256
 ARCH aarch64|ppc64|ppc64le
 TIMEOUT 5
 AFTER ./testprogs/stack_args
 
 NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
+PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", (int32)sarg0, (int32)sarg1); exit(); }
 EXPECT SUCCESS 16 32
 ARCH s390x
 TIMEOUT 5
--- a/src/format_string.cpp
+++ b/src/format_string.cpp
@@ -2,11 +2,42 @@
 #include "log.h"
 #include "struct.h"
 
+#include <cassert>
+#include <cinttypes>
+#include <regex>
+#include <unordered_map>
+
 namespace bpftrace {
 
 const int FMT_BUF_SZ = 512;
 // bpf_trace_printk cannot use more than three arguments, see bpf-helpers(7).
 const int PRINTK_MAX_ARGS = 3;
+
+namespace {
+
+// Matches conversion specifiers without explicit length modifiers.
+const std::regex int_format_re("%(-?[0-9.]*)([diuoxX])");
+const std::unordered_map<std::string, std::string> int_format_replacement = {
+  { "d", PRId64 }, { "i", PRIi64 }, { "u", PRIu64 },
+  { "o", PRIo64 }, { "x", PRIx64 }, { "X", PRIX64 },
+};
+
+// Performs in-place replacement of int specifiers with their 8-byte versions.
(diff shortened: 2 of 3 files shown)


tests/runtime/builtin
```diff
@@ -63,21 +63,21 @@ TIMEOUT 5
 AFTER ./testprogs/syscall read
 
 NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
+PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", (int32)sarg0, (int32)sarg1); exit(); }
```
tnovak, line 66: This is an interesting example where `%d` was masking the fact that `sarg<N>` reads 8 bytes from the stack (at address `SP + N*8`).

`too_many_args()` is a function that takes `int` arguments, and clang generates the following code:

```
[...]
   0x00000000004005a1 <+17>:    mov    $0x1,%esi
   0x00000000004005a6 <+22>:    mov    $0x2,%edx
   0x00000000004005ab <+27>:    mov    $0x4,%ecx
   0x00000000004005b0 <+32>:    mov    $0x8,%r8d
   0x00000000004005b6 <+38>:    mov    $0x10,%r9d
   0x00000000004005bc <+44>:    movl   $0x20,(%rsp)
   0x00000000004005c3 <+51>:    movl   $0x40,0x8(%rsp)
   0x00000000004005cb <+59>:    movl   $0x80,0x10(%rsp)
   0x00000000004005d3 <+67>:    movl   $0x100,0x18(%rsp)
   0x00000000004005db <+75>:    callq  0x400560 <too_many_args>
```

It only does a 4-byte write (`movl`) to `%rsp`, so the value returned by `sarg0` also includes whatever was already on the stack (the other 4 bytes at `%rsp`).

`%d` truncated previously, but with this change we'd display the full 64-bit value.

tnovak (commented):
Submitting this as an RFC since this is a change in behavior -- `%d`, `%x`, etc. now by default print the full 8 byte value. This may break scripts relying on the previous truncating behavior (see also inline comment about `sarg`).

A couple of alternatives:
- Instead of replacing `%d` with `%lld`, cast `value_` explicitly to the type that matches the length modifier (e.g. for "%ld" we'd do `snprintf(..., static_cast<long int>(value_))`). This may be a bit more involved/less elegant.
- Do nothing. This requires bpftrace scripts to use different format specifiers ("%lld" instead of "%d") on some platforms (armv7) -- okay as a workaround perhaps, but not ideal since it requires knowledge of bpftrace internals (the fact that the values in `Printable(S)Int` are always 64 bit).

Interestingly, I've been using armv7 builds of bpftrace on Android for quite a while now and haven't noticed this before. Turns out this was because with `FORTIFY` enabled, `snprintf` is replaced with a hardened variant like `__snprintf_chk` that takes additional arguments, which means that `value_` is expected to be on the stack anyway. However, testing on Linux with gcc/glibc, `value_` is the 4th argument so "%d" will try to get it from `r3` whereas "%lld" will correctly read it from the stack.

viktormalik:
I'm not convinced that this is the best approach. Users shouldn't be required to know bpftrace internals but also we shouldn't do anything unexpected under the hood. If the user uses `%d` and passes an `int32` value, we should print an `int32` value, no matter if we represent it as `int64` internally.

> A couple of alternatives:
> 
> * Instead of replacing `%d` with `%lld`, cast `value_` explicitly to the type that matches the length modifier (e.g. for "%ld" we'd do `snprintf(..., static_cast<long int>(value_))`). This may be a bit more involved/less elegant.

This sounds like the best approach to me. In addition, we could check if the passed argument type (width) matches the corresponding specifier and if not, print a warning (and do the implicit conversion). IIUC, the `sarg` runtime test would then work the same way as before, it would just print an additional warning (which is IMHO fine).

viktormalik:
> I'm leaning towards that as well -- I'll try implementing it to see what it ends up looking like.

Thanks!

> Warnings sound useful, but may be too noisy if enabled by default; for example, something like `printf("%d\n", nsecs - @start)` is presumably a common pattern, and the second argument is a `uint64`.

Right, good point. I'm now going back and forth on this, too :-). It feels like the warning could be useful in some situations (e.g. for the `sarg` case or when user passes some value read from memory) but we definitely don't want it in others (for values coming from some builtins). Unfortunately, it's not easy to differentiate these two, unless we introduce a new internal type for the relevant builtins (and that feels like over-complicating things in this case).

Issue #2635, opened 2023-06-02
Test compilation fails with missing libdw-devel


wcunning:
Just cloned fresh on a new Fedora 38 install, and other than finding a couple of additional libraries needed installing (will submit INSTALL.md PR later today), I now get the following error: 

```
[ 34%] Generating btf_data.h, dwarf_data.h
CMake Error at /home/wcunning/projects/bpf/bpftrace/tests/data/configure_headers.cmake:7 (file):
  file READ must be called with at least two additional arguments


make[2]: *** [tests/data/CMakeFiles/debuginfo_data.dir/build.make:76: tests/data/btf_data.h] Error 1
make[1]: *** [CMakeFiles/Makefile2:2611: tests/data/CMakeFiles/debuginfo_data.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

Google does not indicate immediately what might be going wrong on this compilation, and I didn't see any recently opened or closed issues that might help with the error message. Anybody got any ideas?

fbs:
Do you have libdw (drawf) installed?

wcunning:
That does not seem to be a package I can install. There's a libdwarf, but nothing seems to exist for drawf. Am I missing something in how I should be searching?

wcunning:
@fbs You were correct -- I had the relevant library, but not the development version. I have successfully compiled now, and have created a pull request to update INSTALL.md to reflect the additional dependencies on Fedora 38. 

https://github.com/iovisor/bpftrace/pull/2637

fbs:
Thanks. If you want to attempt to fix the build issue. https://github.com/iovisor/bpftrace/blob/master/tests/data/CMakeLists.txt#L46-L60 this defines the var that is used https://github.com/iovisor/bpftrace/blob/master/tests/data/configure_headers.cmake .  Looks like the depends https://github.com/iovisor/bpftrace/blob/master/tests/data/CMakeLists.txt#L79 doesn't work as expected

xpahos:
Looks like the same problem with Ubuntu.

PR #2675, opened 2023-07-13
Add libdw-dev to ubuntu install in INSTALL.md

paulzhn:
According to It #2635, it seems like we also need `libdw-dev` to build the bpftrace in ubuntu 22.04, or the same error (file READ must be called with at least two additional arguments) to the issue occurs.

##### Checklist


viktormalik (approved):
Thanks!

PR #2678, opened 2023-07-17
Use nix infra for ci.yml jobs

### LINKED ISSUE #2627: Refactor `ci.yml` to use nix flake
Should be doable. Goal is CI runs a single shell command. This will make it easy for developers to run it locally.

I've also created https://app.cachix.org/cache/bpftrace-ci to help cache the nix builds.

The cool thing with nix is that since it's fully reproducible builds, all build deps can be trivially cached.

danobi:
This refactors ci.yml to make it use the new nix infra.

This should make it considerably easier to reproduce
the CI results locally.

This closes #2627.

##### Checklist


danobi:
Ready for review! There is quite some trickiness with running the tests as root, but at least it's all in one place and well documented with comments now. Other than that, I think this greatly simplifies the CI and makes local runs a breeze.

I intend to look at simplifying the embedded CI next.

tests/codegen/llvm/kfunc_dereference.ll
```diff
@@ -6,7 +6,7 @@ target triple = "bpf-pc-linux"
 ; Function Attrs: nounwind
 declare i64 @llvm.bpf.pseudo(i64 %0, i64 %1) #0
 
-define i64 @"kfunc:vmlinux:tcp_sendmsg"(i8* %0) section "s_kfunc:vmlinux:tcp_sendmsg_1" {
+define i64 @"kfunc:tcp_sendmsg"(i8* %0) section "s_kfunc:tcp_sendmsg_1" {
```
viktormalik, line 9: Any idea why this is happening? I've encountered the same using the old `update_codegen_tests.sh` on my machine. It seems that the function -> module mapping is not computed correctly and IIRC, it should be taken from `/sys/kernel/debug/tracing/available_filter_functions`.

Maybe it's the same reason as with the other changed codegen test?

tests/codegen/llvm/array_integer_equal_comparison.ll
```diff
@@ -63,80 +64,71 @@ if_body:                                          ; preds = %arraycmp.done
 if_end:                                           ; preds = %deadcode, %arraycmp.done
   ret i64 0
 
-arraycmp.false:                                   ; preds = %arraycmp.loop7, %arraycmp.loop3, %arraycmp.loop, %entry
+while_cond:                                       ; preds = %arraycmp.loop, %entry
```
viktormalik, line 67: I don't think that this is correct. Looking at [array_integer_equal_comparison.cpp](https://github.com/iovisor/bpftrace/blob/master/tests/codegen/array_integer_equal_comparison.cpp#L10-L13):
```cpp
  if (bpftrace->has_loop())
  {
    return;
  }
```
This test should be skipped when the machine has the `bpf_loop` helper, so, IIUC, the codegen should have the loop unrolled, which is what the original version has.

FWIW, I think that we have a bug somewhere in the testing logic b/c I've seen this on my machine, too, with the old `update_codegen_tests.sh`.

viktormalik (commented):
This incredibly helps debugging the CI, thanks!

I've some concerns about the codegen test changes (comments inline) and also a couple of high-level comments:
- Could you also update the [Continuous intergation](https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#continuous-integration) section of the development guide? Also it may be worth mentioning that the first run of `.github/include/ci.py` takes a considerable amount of time to build the nix environment (I was puzzled for a while b/c the command just seemed to be stuck).
- As we discussed in the last office hours, we want to be particular about making the tests pass after every commit to simplify bisecting. Here, the failing tests are fixed only after the new infra is introduced, could you do it in the opposite order? FWIW, I don't think that squashing the entire PR would be useful here so I'd rather just reorder the commits.

tests/codegen/llvm/kfunc_dereference.ll
danobi: Took a look. It's cuz codegen tests looked at the host's real tracefs for available_filter_functions. It "worked" in CI cuz codegen tests run as root in the container and we bind mount in debugfs. But locally it silently fails. The correct fix is to mock out the access (perhaps more correct is to use fakes and dependency inject, but a different discussion).

I've put up the fix in #2712

tests/codegen/llvm/array_integer_equal_comparison.ll
danobi: Good catch. I concur, I think there is a bug in that conditional test logic. I've put up #2714 which should improve the situation.

danobi:
Ack on both high level comments. Will update this PR when https://github.com/iovisor/bpftrace/pull/2712 and #2714 are in.

viktormalik (approved):
LGTM, thanks!

danobi:
Did a final rebase to make sure no breakage occurred.

PR #2681, opened 2023-07-17
Print stdout and stderr of befores and after on fail

ivankelly:
If bpftrace is waiting for before to do something and it doesn't do it, it's good to know why.

##### Checklist

  * not user visible
  * Is part of testing

### DIFF
--- a/tests/runtime/engine/runner.py
+++ b/tests/runtime/engine/runner.py
@@ -156,9 +156,12 @@ class Runner(object):
 
         p = None
         befores = []
+        befores_output = []
         bpftrace = None
         after = None
+        after_output = None
         cleanup = None
+        bpf_call = "[unknown]"
         try:
             result = None
             timeout = False
@@ -196,7 +199,11 @@ class Runner(object):
 
             if test.befores:
                 for before in test.befores:
-                    before = subprocess.Popen(before.split(), start_new_session=True)
+                    before = subprocess.Popen(before.split(),
+                                              start_new_session=True,
+                                              universal_newlines=True,
+                                              stdout=subprocess.PIPE,
+                                              stderr=subprocess.STDOUT)
                     befores.append(before)
 
                 with open(os.devnull, 'w') as dn:
@@ -264,7 +271,11 @@ class Runner(object):
                     attached = True
                     signal.alarm(test.timeout or DEFAULT_TIMEOUT)
                     if test.after:
-                        after = subprocess.Popen(test.after, shell=True, start_new_session=True)
+                        after = subprocess.Popen(test.after, shell=True,
+                                                 start_new_session=True,
+                                                 universal_newlines=True,
+                                                 stdout=subprocess.PIPE,
+                                                 stderr=subprocess.STDOUT)
 
             signal.alarm(0)
             output += p.stdout.read()
@@ -298,14 +309,23 @@ class Runner(object):
         finally:
             if befores:
                 for before in befores:
+                    try:
+                        befores_output.append(before.communicate(timeout=1)[0])
+                    except subprocess.TimeoutExpired:
+                        pass # if timed out getting output, there is effectively no output
                     if before.poll() is None:
                         os.killpg(os.getpgid(before.pid), signal.SIGKILL)
 
             if bpftrace and bpftrace.poll() is None:
                 os.killpg(os.getpgid(p.pid), signal.SIGKILL)
 
-            if after and after.poll() is None:
-                os.killpg(os.getpgid(after.pid), signal.SIGKILL)
+            if after:
+                try:
+                    after_output = after.communicate(timeout=1)[0]
+                except subprocess.TimeoutExpired:
+                        pass # if timed out getting output, there is effectively no output
+                if after.poll() is None:
+                    os.killpg(os.getpgid(after.pid), signal.SIGKILL)
 
         if test.cleanup:
             try:
@@ -317,11 +337,21 @@ class Runner(object):
                 print('\tCLEANUP error: %s' % e.stderr)
                 return Runner.FAIL
 
+        def print_befores_and_after_output():
+            if len(befores_output) > 0:
+                for out in befores_output:
+                    out = out.encode("unicode_escape").decode("utf-8")
+                    print(f"\tBefore cmd output: {out}")
+            if after_output is not None:
+                out = after_output.encode("unicode_escape").decode("utf-8")
+                print(f"\tAfter cmd output: {out}")
+
         if p and p.returncode != 0 and not test.will_fail and not timeout:
             print(fail("[  FAILED  ] ") + "%s.%s" % (test.suite, test.name))
             print('\tCommand: ' + bpf_call)
             print('\tUnclean exit code: ' + str(p.returncode))
             print('\tOutput: ' + output.encode("unicode_escape").decode("utf-8"))
+            print_befores_and_after_output()
             return Runner.FAIL
 
         if result:
@@ -335,4 +365,5 @@ class Runner(object):
             print('\tCommand: ' + bpf_call)
             print('\tExpected: ' + test.expect)
             print('\tFound: ' + output.encode("unicode_escape").decode("utf-8"))
+            print_befores_and_after_output()
             return Runner.FAIL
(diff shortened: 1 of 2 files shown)


ivankelly:
@viktormalik I've moved this into it's own PR. Will revise the other one on top of this

tests/runtime/engine/runner.py
```diff
@@ -321,11 +335,26 @@ def run_test(test):
                 print('\tCLEANUP error: %s' % e.stderr)
                 return Runner.FAIL
 
+        def print_befores_and_after_output():
+            if len(befores_output) > 0:
+                print("\tBefore cmd output:")
+                for (out, err) in befores_output:
+                    out = out.decode("utf-8")
+                    err = err.decode("utf-8")
+                    print(f"stdout: [{out}]\nstderr: [{err}]")
```
viktormalik, line 344: Perhaps would be nicer with indentation? Also, I'm not sure what the purpose of the square brackets is, I'd remove them:
```suggestion
                    print(f"\t\tstdout: {out}\n\t\tstderr: {err}")
```
Same below for `AFTER`.

viktormalik (commented):
Good idea separating this! Just a formatting comment, otherwise LGTM.

tests/runtime/engine/runner.py
```diff
@@ -321,11 +335,26 @@ def run_test(test):
                 print('\tCLEANUP error: %s' % e.stderr)
                 return Runner.FAIL
 
+        def print_befores_and_after_output():
+            if len(befores_output) > 0:
+                print("\tBefore cmd output:")
+                for (out, err) in befores_output:
+                    out = out.decode("utf-8").replace('\n', '\n\t\t\t')
+                    err = err.decode("utf-8").replace('\n', '\n\t\t\t')
```
viktormalik, line 343: I wouldn't do this. Sometimes, it's useful to see the exact output of the commands and adding artificial whitespace doesn't help much.

viktormalik (commented):
Now, I'm getting this:
```
[ RUN      ] probe.uretprobe
[  FAILED  ] probe.uretprobe
	Command: ../src//bpftrace -e 'uretprobe:/bin/bash:echo_builtin { printf("readline: %d\n", retval); exit();}'
	Expected: readlin: [0-9]*
	Found: Attaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\nreadline: 0\n\n\n
	After cmd output:
stdout:
			lala
			
stderr:
		
```
I'd prefer something like this:
```
[ RUN      ] probe.uretprobe
[  FAILED  ] probe.uretprobe
	Command: ../src//bpftrace -e 'uretprobe:/bin/bash:echo_builtin { printf("readline: %d\n", retval); exit();}'
	Expected: readlin: [0-9]*
	Found: Attaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\nreadline: 0\n\n\n
	After cmd output:
		stdout: lala
		stderr:	
```
Obviously, it won't be that nice if there are linebreaks in stdout/stderr, but that's IMHO ok.

tests/runtime/engine/runner.py
```diff
@@ -303,14 +308,23 @@
         finally:
             if befores:
                 for before in befores:
+                    try:
+                        befores_output.append(before.communicate(timeout=5))
+                    except:
```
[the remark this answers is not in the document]
danobi: Warning looks reasonable. Perhaps only catch the timeout exception that communicate() can raise

tests/runtime/engine/runner.py
```diff
@@ -303,14 +308,23 @@
...
 
             if bpftrace and bpftrace.poll() is None:
                 os.killpg(os.getpgid(p.pid), signal.SIGKILL)
 
-            if after and after.poll() is None:
-                os.killpg(os.getpgid(after.pid), signal.SIGKILL)
+            if after:
+                try:
+                    after_output = after.communicate(timeout=5)
+                except:
```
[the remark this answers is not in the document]
danobi: same as above

tests/runtime/engine/runner.py
danobi: +1. I think we should follow the example below:

```
print('\tFound: ' + output.encode("unicode_escape").decode("utf-8"))
```

That is, encode all the whitespace as escaped characters so it doesn't scribble all over the screen. Makes output tidier.

tests/runtime/engine/runner.py
```diff
@@ -321,11 +335,26 @@ def run_test(test):
...
+                print("\tAfter cmd output:")
+                (out, err) = after_output;
+                out = out.decode("utf-8").replace('\n', '\n\t\t\t')
+                err = err.decode("utf-8").replace('\n', '\n\t\t\t')
+                print(f"stdout:\n\t\t\t{out}\nstderr:\n\t\t\t{err}")
+
         if p and p.returncode != 0 and not test.will_fail and not timeout:
             print(fail("[  FAILED  ] ") + "%s.%s" % (test.suite, test.name))
-            print('\tCommand: ' + bpf_call)
+            print('\tCommand: ' + (bpf_call or "unset"))
```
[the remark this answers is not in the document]
ivankelly: previously it could have been, but I guess this is redundant since I added a declaration of bpf_call at the top level

tests/runtime/engine/runner.py
ivankelly: Changed to use the unicode escape

tests/runtime/engine/runner.py
```diff
@@ -262,7 +266,9 @@ def run_test(test):
                     attached = True
                     signal.alarm(test.timeout or DEFAULT_TIMEOUT)
                     if test.after:
-                        after = subprocess.Popen(test.after, shell=True, start_new_session=True)
+                        after = subprocess.Popen(test.after, shell=True, start_new_session=True,
+                                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
danobi, line 270: The other shellouts we do we combine stdout and stderr into one stream. I typically find that helpful cuz a lot of programs will interleave stdout and stderr in a way where separating them makes it harder to understand. WDYT about combining the streams here too?

tests/runtime/engine/runner.py
```diff
@@ -296,14 +302,23 @@ def run_test(test):
         finally:
             if befores:
                 for before in befores:
+                    try:
+                        befores_output.append(before.communicate(timeout=5))
```
danobi, line 306: By here we already intend on killing the process. Does it make sense to wait up to 5s if it's still running? Maybe 1s or less is more prudent.

tests/runtime/engine/runner.py
```diff
@@ -296,14 +302,23 @@ def run_test(test):
...
                         os.killpg(os.getpgid(before.pid), signal.SIGKILL)
 
             if bpftrace and bpftrace.poll() is None:
                 os.killpg(os.getpgid(p.pid), signal.SIGKILL)
 
-            if after and after.poll() is None:
-                os.killpg(os.getpgid(after.pid), signal.SIGKILL)
+            if after:
+                try:
+                    after_output = after.communicate(timeout=5)
```
danobi, line 317: Same as above

ivankelly:
Going to churn on this, feel free to pick it up.

viktormalik:
I rebased the work and addressed all the remaining comments from reviews. I also used this opportunity to add one commit removing some unnecessary debugging prints from the runtime runner.

Ready for review @danobi

tests/runtime/engine/runner.py
```diff
@@ -317,11 +335,23 @@ def run_test(test):
                 print('\tCLEANUP error: %s' % e.stderr)
                 return Runner.FAIL
 
+        def print_befores_and_after_output():
+            if len(befores_output) > 0:
+                print("\tBefore cmd output: ", end="")
+                for out in befores_output:
+                    out = out.encode("unicode_escape").decode("utf-8")
+                    print(out)
+            if after_output is not None:
```
xh4n3, line 344: nit: Since bpftrace support multiple BEFORE clause, do we need to print the exact command executed here to diffierenciate?

xh4n3 (approved):

tests/runtime/engine/runner.py
viktormalik: Good point. I wouldn't bother with printing the BEFORE command, though. I just fixed the code to show one `Before cmd output: ...` line per BEFORE command and they can be matched to the respective commands by the order.

Feel free to re-review.

xh4n3: yeah, this should be sufficient

danobi (approved):

PR #2682, opened 2023-07-18
utils: resolve_binary_path: Support symbolic links in ENV

Rtoax:
On some systems `/bin` is a symbolic link to `/usr/bin` (`/bin -> /usr/bin`), When using the '`-c`' parameter to specify the executable program in the environment variable, the problem of "multiple binary files matching" occurs, for example:

```
```

```
```

This commit fixes the issue.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -308,6 +308,16 @@ std::vector<std::string> split_string(const std::string &str,
   return elems;
 }
 
+std::string realpath(std::string const &path)
+{
+  int exist = ::access(path.c_str(), F_OK);
+  if (exist == -1)
+  {
+    return path;
+  }
+  return std::filesystem::canonical(path);
+}
+
 /// Erase prefix up to the first colon (:) from str and return the prefix
 std::string erase_prefix(std::string &str)
 {
@@ -842,6 +852,7 @@ resolve_binary_path(const std::string &cmd, const char *env_paths, int pid)
   if (cmd.find("*") != std::string::npos)
     candidate_paths = ::expand_wildcard_paths(candidate_paths);
 
+  std::unordered_set<std::string> abs_path;
   std::vector<std::string> valid_executable_paths;
   for (const auto &path : candidate_paths)
   {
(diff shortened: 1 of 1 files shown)


src/utils.cpp
```diff
@@ -849,10 +857,17 @@ resolve_binary_path(const std::string &cmd, const char *env_paths, int pid)
...
+      rel_path = realpath(path);
+
     if (bcc_elf_is_exe(rel_path.c_str()) ||
         bcc_elf_is_shared_obj(rel_path.c_str()))
-      valid_executable_paths.push_back(rel_path);
+    {
+      auto it = std::find(valid_executable_paths.begin(),
+                          valid_executable_paths.end(), rel_path);
+      if (it == valid_executable_paths.end()) {
+        valid_executable_paths.push_back(rel_path);
```
viktormalik, line 868: Instead of doing this manually, make `valid_executable_paths` a `std::unordered_set`.

src/utils.cpp
```diff
@@ -308,6 +308,14 @@ std::vector<std::string> split_string(const std::string &str,
   return elems;
 }
 
+std::string realpath(std::string const& path) {
+  int exist = ::access(path.c_str(), F_OK);
+  if (exist != 0) {
```
viktormalik, line 313: nit: [access](https://man7.org/linux/man-pages/man2/access.2.html) returns -1 on error so let's check it directly:
```suggestion
  if (exist == -1) {
```

viktormalik (commented):
Looks like a useful feature, thanks!

Please, look at the CI failures and fix clang-format issues.

src/utils.cpp
Rtoax: Fix it , thanks.

src/utils.cpp
```diff
@@ -842,19 +852,24 @@ resolve_binary_path(const std::string &cmd, const char *env_paths, int pid)
   if (cmd.find("*") != std::string::npos)
     candidate_paths = ::expand_wildcard_paths(candidate_paths);
 
+  std::unordered_set<std::string> abs_path;
```
viktormalik, line 855: Why can't we just change `valid_executable_paths` (and consequently the return type of this function) to `std::unordered_set` and avoid the copy below?

viktormalik (commented):
This seems to break the tests in several ways. Probe listing tests are broken b/c abs path is reported now. Also several tests are now segfaulting.

You may need to take another approach to resolve this issue.

src/utils.cpp
Rtoax: It's wired, On my `CentOS-Stream9`, there are no failed tests at all.

```
[==========] 401 tests from 22 test suites ran. (610238 ms total)
[  PASSED  ] 401 tests.
```

I'll try another approach, maybe submit new V2 patch, thanks.

Rtoax:
@viktormalik I just close this PR, and start V2 (https://github.com/iovisor/bpftrace/pull/2689)

PR #2686, opened 2023-07-19
Support casting int <-> int array

### LINKED ISSUE #2634: IP address comparison is incomplete
### What reproduces the bug?

I  tried the following code.
```
kprobe:tcp_connect
{
        $sk = ((struct sock *) arg0);
        if ( $sk->__sk_common.skc_daddr == pton("10.244.166.137")){
                printf("find");
        }
}
```
As `pton(const string *addr)` described in doc, it could help me to select packets.
But it failed as following error:
```
./find-gid.bt:40:7-59: ERROR: Type mismatch for '==': comparing 'unsigned int32' with 'unsigned int8[4]'
    if ( $sk->__sk_common.skc_daddr == pton("10.244.166.137")){
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I find the pr #2248, which tried to solve this problem. But it doesn't seems quite finished.

### bpftrace --info

[root@node1 bpf-probe]# bpftrace --info
System
  OS: Linux 5.14.0-316.el9.x86_64 #1 SMP PREEMPT_DYNAMIC Fri May 19 11:26:13 UTC 2023
  Arch: x86_64

Build
  version: v0.17.0
  LLVM: 15.0.7
  unsafe uprobe: no
  bfd: yes
  libdw (DWARF support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes
  skboutput: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  map batch: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  iter:task: yes
  iter:task_file: yes
  kprobe_multi: yes
  raw_tp_special: yes

viktormalik:
This PR adds support for conversions between integers and integer arrays. The main purpose is to allow conversion from/to byte arrays.

A good use-case for this are scripts working with IPv4 addresses which the kernel sometimes stores as 32-bit integers (e.g. in [struct sock_common](https://elixir.bootlin.com/linux/v6.4/source/include/net/sock.h#L167)) and sometimes as 4-byte arrays (e.g. in [TCP tracepoints](https://elixir.bootlin.com/linux/v6.4/source/include/trace/events/tcp.h#L64)). This feature, in combination with array comparison added in #2387, allows to compare such representations:
```
kfunc:tcp_connect
{
    @ = args->sk->__sk_common.skc_daddr;
}
tracepoint:tcp:tcp_rcv_space_adjust
{
    // map values are always 64-bit so first cast to uint32
    // and then to the target array
    if ((uint8[4])(uint32)@ == args->daddr)
    {
        ...
    }
}
```

Similarly, it is now possible to compare the int IPv4 addresses with absolute addresses obtained from the `pton` builtin:
```
kfunc:tcp_connect
{
    if (args->sk->__sk_common.skc_daddr == (uint32)pton("127.0.0.1"))
        ...
}
```

Both the cast and the destination type must have the same size. When casting to an array, it is possible to omit the size which will be determined automatically from the size of the cast value.

This also does some refactoring of type parsing. See individual commit messages for more implementation details.

Resolves #2634.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -36,9 +36,10 @@ discussion to other files in /docs, the /tools/\*\_examples.txt files, or blog p
     - [10. `++ and --`: increment operators](#10--and----increment-operators)
     - [11. `[]`: Array access](#11--array-access)
     - [12. Integer casts](#12-integer-casts)
-    - [13. Looping constructs](#13-looping-constructs)
-    - [14. `return`: Terminate Early](#14-return-terminate-early)
-    - [15. `( , )`: Tuples](#15----tuples)
+    - [13. Array casts](#13-array-casts)
+    - [14. Looping constructs](#14-looping-constructs)
+    - [15. `return`: Terminate Early](#15-return-terminate-early)
+    - [16. `( , )`: Tuples](#16----tuples)
 - [Probes](#probes)
     - [1. `kprobe`/`kretprobe`: Dynamic Tracing, Kernel-Level](#1-kprobekretprobe-dynamic-tracing-kernel-level)
     - [2. `kprobe`/`kretprobe`: Dynamic Tracing, Kernel-Level Arguments](#2-kprobekretprobe-dynamic-tracing-kernel-level-arguments)
@@ -839,7 +840,40 @@ Attaching 1 probe...
 ^C
 ```
 
-## 13. Looping Constructs
+## 13. Array casts
+
+It is possible to cast between integer arrays and integers. Both the source and
+the destination type must have the same size. The main purpose of this is to
+allow casts from/to byte arrays.
+
+Example:
+
+```
+# bpftrace -e 'BEGIN { $a = (int8[8])12345; printf("%x %x\n", $a[0], $a[1]); printf("%d\n", (uint64)$a); }'
+Attaching 1 probe...
+39 30
+12345
+^C
+```
+
+When casting to an array, it is possible to omit the size which will be
+determined automatically from the size of the cast value.
+
+This feature is especially useful when working with IP addresses since various
+libraries, builtins, and parts of the kernel use different approaches to
+represent addresses (usually byte arrays vs. integers). Array casting allows
+seamless comparison of such representations:
+
+```
+# cat ip_cmp.bt
+kfunc:tcp_connect
+{
+    if (args->sk->__sk_common.skc_daddr == (uint32)pton("127.0.0.1"))
+        ...
+}
+```
+
+## 14. Looping Constructs
 
 **Experimental**
 
@@ -853,12 +887,12 @@ bpftrace supports C style while loops:
 
 Loops can be short circuited by using the `continue` and `break` keywords.
 
-## 14. `return`: Terminate Early
+## 15. `return`: Terminate Early
 
 The `return` keyword is used to exit the current probe. This differs from
 `exit()` in that it doesn't exit bpftrace.
 
-## 15. `( , )`: Tuples
+## 16. `( , )`: Tuples
 
 N-tuples are supported, where N is any integer greater than 1.
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2355,12 +2355,30 @@ void SemanticAnalyser::visit(Cast &cast)
 
   if (!cast.type.IsIntTy() && !cast.type.IsPtrTy() &&
       !(cast.type.IsPtrTy() && !cast.type.GetElementTy()->IsIntTy() &&
-        !cast.type.GetElementTy()->IsRecordTy()))
+        !cast.type.GetElementTy()->IsRecordTy()) &&
+      // we support casting integers to int arrays
+      !(cast.type.IsArrayTy() && cast.type.GetElementTy()->IsIntTy()))
   {
     LOG(ERROR, cast.loc, err_) << "Cannot cast to \"" << cast.type << "\"";
   }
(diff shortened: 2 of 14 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -2336,12 +2336,30 @@ void SemanticAnalyser::visit(Cast &cast)
...
+    {
+      LOG(ERROR, cast.loc, err_) << "Bit arrays are not supported";
+      return;
+    }
+
+    if (rhs.IsIntTy())
+      cast.type.is_internal = true;
+  }
+
+  if ((cast.type.IsIntTy() && !rhs.IsIntTy() && !rhs.IsPtrTy() &&
```
danobi, line 2370: Is it useful to enable casting a pointer to an array? Would it make more sense to let the user do `(uint8[8])(uint64)$ptr`?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2351,6 +2351,18 @@ void SemanticAnalyser::visit(Cast &cast)
...
 
+    if (cast.type.GetNumElements() == 0)
+    {
+      if (cast.type.GetElementTy()->GetSize() == 0)
+        LOG(ERROR, cast.loc, err_) << "Could not determine size of the array";
+      else
+      {
+        // cast to unsized array (e.g. int8[]), determine size from RHS
+        auto num_elems = rhs.GetSize() / cast.type.GetElementTy()->GetSize();
+        cast.type = CreateArray(num_elems, *cast.type.GetElementTy());
```
danobi, line 2362: This relies on the later size check right? Cuz otherwise we need to check if the array elem size can divide the cast-ee size correctly.

src/ast/passes/semantic_analyser.cpp
danobi, line 1: neat use of FLA semantics.

docs/reference_guide.md
```diff
@@ -842,7 +843,26 @@ Attaching 1 probe...
...
+allow casts from/to byte arrays.
+
+Example:
+
+```
+# bpftrace -e 'BEGIN { $a = (int8[8])12345; printf("%x %x\n", $a[0], $a[1]); printf("%d\n", (uint64)$a); }'
+Attaching 1 probe...
+39 30
+12345
+^C
```
danobi, line 859: Throw in the pton example too? I find that to be a good motivating example

src/ast/passes/semantic_analyser.cpp
danobi: Ah I misread. This is the pre-existing ptr -> int check

src/ast/passes/semantic_analyser.cpp
viktormalik: It does, but an explicit check here wouldn't hurt and would make things easier to understand.

src/ast/passes/semantic_analyser.cpp
viktormalik: Yeah, only int <-> array casts for now. I'm going to refactor this whole thing in a follow-up PR b/c it's becoming too hard to read (and it's even missing few checks).

docs/reference_guide.md
viktormalik: Good idea, will add it.

src/ast/passes/semantic_analyser.cpp
viktormalik: Hm, I just realized that this can't really happen as we only support integers and integer arrays. The only error we can get is when the array element size is larger than the cast-ee size.

Still, I added the check to get a better error message and capture these cases if we add more types in future.

viktormalik:
Thanks for the review @danobi! I addressed all the comments and added them as new commits to simplify reviewing. Will fixup the commits afterwards.

src/ast/passes/semantic_analyser.cpp
danobi: Ah right. All powers of 2 :)

viktormalik:
This is quite a big feature and could use another review.

@fbs, @ajor

tests/runtime/array
```diff
@@ -166,3 +166,30 @@ PROG struct D { int x; int y[0]; } uprobe:./testprogs/array_access:test_variable
 EXPECT @y: 1
 TIMEOUT 5
 AFTER ./testprogs/array_access
+
+NAME cast pid to int array
+PROG BEGIN { printf("pid byte: %02x\n", ((uint8[8])pid)[0]); exit(); }
+EXPECT pid byte: [0-9a-f][0-9a-f]
```
ajor, line 172: Regexes for expected results are a common pattern in our runtime tests, but in my opinion this kind of test doesn't really add anything.

Checking that we can cast a `pid` to an array and that we can print an array element are both things that unit tests can handle. Checking that the results are correct is what the runtime tests should be doing and a regex doesn't do that.

ajor (approved):
Really nice - the code's a lot simpler than I'd expected!

tests/runtime/array
viktormalik: Good point, I dropped the test since there's another one testing correctness of int -> array casting.

PR #2689, opened 2023-07-20
child.cpp: validate_cmd: Support symbolic links in environment variables

Rtoax:
On some systems `/bin` is a symbolic link to `/usr/bin` (`/bin -> /usr/bin`), When using the '-c' parameter to specify the executable program in the environment variable, the problem of "multiple binary files matching" occurs, for example:

```
```

```
```

This commit fixes the issue. This is the V2 submition, V1(https://github.com/iovisor/bpftrace/pull/2682)

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -788,6 +788,18 @@ std::string exec_system(const char *cmd)
   return result;
 }
 
+std::string get_absolute_path(std::string const &path)
+{
+  std::error_code ec;
+  std_filesystem::path p(path);
+  std::string abs_path = std_filesystem::canonical(p, ec);
+  if (ec || errno == ENOENT)
+  {
+    return path;
+  }
+  return abs_path;
+}
+
 /*
 Original resolve_binary_path API defaulting to bpftrace's mount namespace
 */
--- a/src/child.cpp
+++ b/src/child.cpp
@@ -95,18 +95,26 @@ static int childfn(void* arg)
 static void validate_cmd(std::vector<std::string>& cmd)
 {
   auto paths = resolve_binary_path(cmd[0]);
-  switch (paths.size())
+
+  // /usr/bin/ls and /bin/ls may be the same file.
+  std::unordered_set<std::string> abs_path;
+  for (const auto& str : paths)
+  {
+    abs_path.insert(get_absolute_path(str.c_str()));
+  }
+
+  switch (abs_path.size())
   {
     case 0:
       throw std::runtime_error("path '" + cmd[0] +
                                "' does not exist or is not executable");
     case 1:
-      cmd[0] = paths.front().c_str();
+      cmd[0] = *(abs_path.begin());
       break;
     default:
       throw std::runtime_error("path '" + cmd[0] +
                                "' must refer to a unique binary but matched " +
-                               std::to_string(paths.size()) + " binaries");
+                               std::to_string(abs_path.size()) + " binaries");
       return;
   }
 
--- a/src/utils.h
+++ b/src/utils.h
@@ -195,6 +195,7 @@ bool is_unsafe_func(const std::string &func_name);
 bool is_compile_time_func(const std::string &func_name);
 bool is_supported_lang(const std::string &lang);
 std::string exec_system(const char *cmd);
+std::string get_absolute_path(std::string const &path);
 std::vector<std::string> resolve_binary_path(const std::string &cmd);
 std::vector<std::string> resolve_binary_path(const std::string &cmd, int pid);
 std::string path_for_pid_mountns(int pid, const std::string &path);
--- a/tests/child.cpp
+++ b/tests/child.cpp
@@ -129,7 +129,8 @@ TEST(childproc, stop_cont)
   int status = 0;
 
   child->run();
-  msleep(25);
+  // Give children enough time to spawn.
+  msleep(250);
   EXPECT_TRUE(child->is_alive());
 
   if (kill(child->pid(), SIGSTOP))
--- a/tests/childhelper.h
+++ b/tests/childhelper.h
@@ -28,8 +28,8 @@ static std::unique_ptr<ChildProc> getChild(std::string cmd)
   {
     StderrSilencer es;
     StdoutSilencer os;
-    os.silence();
-    es.silence();
+    // os.silence();
+    // es.silence();
     child = std::make_unique<ChildProc>(cmd);
   }
   EXPECT_NE(child->pid(), -1);


Rtoax:
It's wired, the CI/CD's error:

```bash
1: /home/runner/work/bpftrace/bpftrace/tests/child.cpp:133: Failure
1: Value of: child->is_alive()
1:   Actual: false
1: Expected: true
1: /home/runner/work/bpftrace/bpftrace/tests/child.cpp:136: Failure
1: Failed
1: kill(SIGSTOP)
...
1: [----------] Global test environment tear-down
1: [==========] 403 tests from 22 test suites ran. (470850 ms total)
1: [  PASSED  ] 402 tests.
1: [  FAILED  ] 1 test, listed below:
1: [  FAILED  ] childproc.stop_cont
1:
1:  1 FAILED TEST
1/3 Test #1: bpftrace_test ....................***Failed  470.89 sec
test 2
    Start 2: runtime_test
```

However, on my env all is well:

```bash
$ sudo ./bpftrace_test --gtest_filter=childproc*
Note: Google Test filter = childproc*
[==========] Running 12 tests from 1 test suite.
[----------] 12 tests from childproc
[ RUN      ] childproc.exe_does_not_exist
[       OK ] childproc.exe_does_not_exist (0 ms)
[ RUN      ] childproc.too_many_arguments
[       OK ] childproc.too_many_arguments (0 ms)
[ RUN      ] childproc.child_exit_success
[       OK ] childproc.child_exit_success (10 ms)
[ RUN      ] childproc.child_exit_err
[       OK ] childproc.child_exit_err (10 ms)
[ RUN      ] childproc.terminate
[       OK ] childproc.terminate (10 ms)
[ RUN      ] childproc.destructor_destroy_child
[       OK ] childproc.destructor_destroy_child (25 ms)
[ RUN      ] childproc.child_kill_before_exec
[       OK ] childproc.child_kill_before_exec (10 ms)
[ RUN      ] childproc.stop_cont
[       OK ] childproc.stop_cont (35 ms)
[ RUN      ] childproc.ptrace_child_exit_success
[       OK ] childproc.ptrace_child_exit_success (10 ms)
[ RUN      ] childproc.ptrace_child_exit_error
[       OK ] childproc.ptrace_child_exit_error (10 ms)
[ RUN      ] childproc.ptrace_child_kill_before_execve
[       OK ] childproc.ptrace_child_kill_before_execve (0 ms)
[ RUN      ] childproc.ptrace_child_term_before_execve
[       OK ] childproc.ptrace_child_term_before_execve (11 ms)
[----------] 12 tests from childproc (138 ms total)

[==========] 12 tests from 1 test suite ran. (138 ms total)
[  PASSED  ] 12 tests.
```

Maybe `msleep(25);` is not enough?

```c
TEST(childproc, stop_cont)
{
  // STOP/CONT should not incorrectly mark the child
  // as dead
  auto child = getChild(TEST_BIN_SLOW);
  int status = 0;

  child->run();
  msleep(25);
  EXPECT_TRUE(child->is_alive());
```

src/utils.cpp
```diff
@@ -788,6 +788,16 @@ std::string exec_system(const char *cmd)
...
 }
 
+std::string get_absolute_path(std::string const &path)
+{
+  int exist = ::access(path.c_str(), F_OK);
+  if (exist == -1)
+  {
+    return path;
+  }
+  return std_filesystem::canonical(path);
```
danobi, line 798: TOCTOU here. How about using:

path canonical( const [std::filesystem::path](http://en.cppreference.com/w/cpp/filesystem/path)& p, [std::error_code](http://en.cppreference.com/w/cpp/error/error_code)& ec );

form of `canonical()` and explicitly checking for ENOENT?

danobi (commented):
> Maybe msleep(25); is not enough?

Yeah, seems racy. How about putting that into a loop and retrying up to ~10 times for child to be alive? That puts an upper bound at ~1/4 of a second. Should be enough for child to spawn.

danobi:
I kicked the job again. Let's see if it reproduces

Rtoax:
> I kicked the job again. Let's see if it reproduces

Seems the error happen again in CICD :

```
1: [ RUN      ] childproc.stop_cont
1: /home/runner/work/bpftrace/bpftrace/tests/child.cpp:133: Failure
1: Value of: child->is_alive()
1:   Actual: false
1: Expected: true
1: /home/runner/work/bpftrace/bpftrace/tests/child.cpp:136: Failure
1: Failed
1: kill(SIGSTOP)
1: [  FAILED  ] childproc.stop_cont (25 ms)
```

I'll try `putting that into a loop and retrying up to ~10 times for child to be alive`, thank you.

src/utils.cpp
Rtoax: Fix it, please review, thanks.

Rtoax:
> > Maybe msleep(25); is not enough?
> 
> Yeah, seems racy. How about putting that into a loop and retrying up to ~10 times for child to be alive? That puts an upper bound at ~1/4 of a second. Should be enough for child to spawn.

I just add some time to `msleep`, is this OK? Please review.

```patch
diff --git a/tests/child.cpp b/tests/child.cpp
index 8dedca07300e..4fe8c2e62e29 100644
--- a/tests/child.cpp
+++ b/tests/child.cpp
@@ -129,7 +129,8 @@ TEST(childproc, stop_cont)
   int status = 0;

   child->run();
-  msleep(25);
+  // Give children enough time to spawn.
+  msleep(250);
   EXPECT_TRUE(child->is_alive());
```

danobi (commented):
I wonder if this PR has a bug. My guess is that somehow the command is corrupted. Maybe try disabling the stderr and stdout silencer in getChild() and seeing if there's any output.

Rtoax:
> I wonder if this PR has a bug. My guess is that somehow the command is corrupted. Maybe try disabling the stderr and stdout silencer in getChild() and seeing if there's any output.

Could you give me some hint, How to enable stdout and stderr to `getChild()` in bpftrace CICD test system, thanks.

Rtoax:
> @Rtoax these lines disable stdout/stderr:
> 
> https://github.com/iovisor/bpftrace/blob/0b3392baa881f501ce684637acbd4136f8a29ed3/tests/childhelper.h#L29-L32
> 
> Perhaps if you comment them out then any errors from the child get printed out to console. It should be picked up by the CI.

Thanks, I just submit a commit, turn on stdio. let's see cicd print. thanks.

Rtoax:
> @Rtoax these lines disable stdout/stderr:
> 
> https://github.com/iovisor/bpftrace/blob/0b3392baa881f501ce684637acbd4136f8a29ed3/tests/childhelper.h#L29-L32
> 
> Perhaps if you comment them out then any errors from the child get printed out to console. It should be picked up by the CI.

CICD print `applet not found`, it's wired.

```
1: [ RUN      ] childproc.stop_cont
1: 10: applet not found
1: /home/runner/work/bpftrace/bpftrace/tests/child.cpp:134: Failure
1: Value of: child->is_alive()
1:   Actual: false
1: Expected: true
1: /home/runner/work/bpftrace/bpftrace/tests/child.cpp:137: Failure
1: Failed
1: kill(SIGSTOP)
1: [  FAILED  ] childproc.stop_cont (250 ms)
```

Rtoax:
> Ah, I don't think it's a problem with busybox. Rather, it's b/c you're resolving `ls` to the busybox binary. I think busybox expects `argv[0]` to be `ls` in this case (which is how it resolves applets). And you're probably providing `busybox` as `argv[0]`.
> 
> This would indeed be a bug with this PR. We don't want to break busybox based systems

Maybe there is another way to solve `/usr/bin/ls` `/bin/ls` problem, let me try. Thanks!!

Rtoax:
> Ah, I don't think it's a problem with busybox. Rather, it's b/c you're resolving `ls` to the busybox binary. I think busybox expects `argv[0]` to be `ls` in this case (which is how it resolves applets). And you're probably providing `busybox` as `argv[0]`.
> 
> This would indeed be a bug with this PR. We don't want to break busybox based systems

Hello, I just submit a v3 PR (https://github.com/iovisor/bpftrace/pull/2708), It is hoped that the child execution problem can be solved on the premise of satisfying CICD (busybox). Please review  (https://github.com/iovisor/bpftrace/pull/2708), thanks.

Rtoax:
Just close this PR.

PR #2698, opened 2023-07-24
Fixing runtime tests for ppc

dubeyabhishek:
In ppc, "nip" is used for instruction pointer.
Other fix is general dependency issue.

### DIFF
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -231,6 +231,6 @@ TIMEOUT 5
 NAME print args in kfunc
 PROG kfunc:vfs_open { print(args); exit(); }
 EXPECT { .path = 0x[0-9a-f]+, .file = 0x[0-9a-f]+ }
-REQUIRES_FEATURE btf
+REQUIRES_FEATURE kfunc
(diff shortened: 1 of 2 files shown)


tests/runtime/builtin
```diff
@@ -231,6 +231,6 @@ TIMEOUT 5
 NAME print args in kfunc
 PROG kfunc:vfs_open { print(args); exit(); }
 EXPECT { .path = 0x[0-9a-f]+, .file = 0x[0-9a-f]+ }
-REQUIRES_FEATURE btf
+REQUIRES_FEATURE kfunc
```
viktormalik, line 234: AFAIK should be:
```suggestion
REQUIRES_FEATURE btf
REQUIRES_FEATURE kfunc
```

dubeyabhishek: in following commit we used same approach f2353019927351f0808befc5518c912e96c98f69

viktormalik: Ok, let's keep consistent.

viktormalik (approved):
Thanks!

PR #2703, opened 2023-07-28
Fixing stack unwind regex for ppc

dubeyabhishek:
Stack unwinding in x86 skips the test2 function,
whereas unwinding on ppc lists function test2.
The current regex is not having test2 match-string, hence test failure on ppc.

### DIFF
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -246,6 +246,15 @@ ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
 PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(ustack); exit(); }
 EXPECT ^\s+test\+0\s+main\+[1-9][0-9]*
 AFTER ./testprogs/disable_aslr ./testprogs/uprobe_symres_exited_process
+ARCH x86_64
+TIMEOUT 5
+
+NAME ustack_elf_symtable
+ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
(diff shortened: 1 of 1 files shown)


tests/runtime/call
```diff
@@ -246,6 +246,15 @@ ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
 PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(ustack); exit(); }
 EXPECT ^\s+test\+0\s+main\+[1-9][0-9]*
 AFTER ./testprogs/disable_aslr ./testprogs/uprobe_symres_exited_process
+ARCH x86_64
+TIMEOUT 5
+
+NAME ustack_elf_symtable
+ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
+PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(ustack); exit(); }
+EXPECT ^\s+test\+0\s+test2\+[1-9][0-9]*\s+main\+[1-9][0-9]*\s+0x[0-9a-fA-F]+\s+0x[0-9a-fA-F]+$
```
viktormalik, line 255: I don't think that the part after `main\+[1-9][0-9]*` is necessary. We should only check that the stacktrace contains symbols from the original program.

viktormalik (commented):
I'm wondering why the x86 variant skips `test2` even though it's marked as `((noinline))`.

@lenticularis39 can you shed any light on this?

tests/runtime/call
dubeyabhishek: Done

viktormalik (commented):
LGTM but still would like to know the reason for `test2` being missing on x86.

PR #2705, opened 2023-07-30
Removing unsupported options of subprocess.run()

dubeyabhishek:
"capture_output" and "text" options are supported beyond python-3.7, whereas some instances of current CI pipeline are using python-3.6 and facing failures. The "stderr" option is supported across python versions.

### DIFF
--- a/tests/runtime/engine/runner.py
+++ b/tests/runtime/engine/runner.py
@@ -307,7 +307,8 @@ class Runner(object):
 
         if test.cleanup:
             try:
-                cleanup = subprocess.run(test.cleanup, shell=True, capture_output=True, text=True)
+                cleanup = subprocess.run(test.cleanup, shell=True, stderr=subprocess.PIPE,
+                                         stdout=subprocess.PIPE, universal_newlines=True)
                 cleanup.check_returncode()
             except subprocess.CalledProcessError as e:
                 print(fail("[  FAILED  ] ") + "%s.%s" % (test.suite, test.name))


dubeyabhishek:
CI run for reference:
https://github.com/iovisor/bpftrace/actions/runs/5706338400/job/15461931821?pr=2606

tests/runtime/engine/runner.py
```diff
@@ -307,7 +307,7 @@ def run_test(test):
 
         if test.cleanup:
             try:
-                cleanup = subprocess.run(test.cleanup, shell=True, capture_output=True, text=True)
+                cleanup = subprocess.run(test.cleanup, shell=True, stderr=subprocess.PIPE)
```
danobi, line 310: All the other callsites use `universal_newlines=True` which is equivalent to `text=True`. So maybe set that too?

Also how about capturing both stdout and stderr? I see below we only print stderr, but buffering both stdout & stderr would help reduce noise for successful CLEANUPs

tests/runtime/engine/runner.py
```diff
@@ -307,7 +307,8 @@ def run_test(test):
 
         if test.cleanup:
             try:
-                cleanup = subprocess.run(test.cleanup, shell=True, capture_output=True, text=True)
+                cleanup = subprocess.run(test.cleanup, shell=True, stderr=subprocess.PIPE,
+                                         stdout=subprocess.PIPE, text=True)
```
danobi, line 311: Isn't `text=True` breaking python 3.6 installations.

dubeyabhishek: surprisingly no. So, I kept it same. In such case what is your view "text or universal_newlines" ?

dubeyabhishek: Not sticking with "text" option, although its passing. Replacing with universal_newlines=True

dubeyabhishek:
@danobi Can we go ahead with latest change?

PR #2708, opened 2023-08-03
child.cpp: validate_cmd: Support symbolic links in environment variables

Rtoax:
On some systems /bin is a symbolic link to /usr/bin (/bin -> /usr/bin), When using the '-c' parameter to specify the executable program in the environment variable, the problem of "multiple binary files matching" occurs, for example:
```bash
```
```bash
```
This commit fixes the issue. The v1[0] will cause CICD busybox error.

[0] https://github.com/iovisor/bpftrace/pull/2689

### DIFF
--- a/src/child.cpp
+++ b/src/child.cpp
@@ -120,7 +120,7 @@ static void validate_cmd(std::vector<std::string>& cmd)
       if (uniq_abs_path.size() == 1)
       {
         cmd[0] = paths.front().c_str();
-        return;
+        break;
       }
       else
       {
@@ -128,7 +128,6 @@ static void validate_cmd(std::vector<std::string>& cmd)
             "path '" + cmd[0] + "' must refer to a unique binary but matched " +
             std::to_string(paths.size()) + " binaries");
       }
-      return;
   }
 
   if (cmd.size() >= (maxargs - 1))
--- a/src/child.cpp
+++ b/src/child.cpp
@@ -104,9 +104,30 @@ static void validate_cmd(std::vector<std::string>& cmd)
       cmd[0] = paths.front().c_str();
       break;
     default:
-      throw std::runtime_error("path '" + cmd[0] +
-                               "' must refer to a unique binary but matched " +
-                               std::to_string(paths.size()) + " binaries");
+      // /bin maybe is a symbolic link to /usr/bin (/bin -> /usr/bin), there
+      // may be worse cases like:
+      // $ realpath /usr/bin/ping /bin/ping /usr/sbin/ping /sbin/ping
+      // /usr/bin/ping
+      // /usr/bin/ping
+      // /usr/bin/ping
+      // /usr/bin/ping
+      std::unordered_set<std::string> uniq_abs_path;
+      for (unsigned int i = 0; i < paths.size(); i++)
+      {
+        uniq_abs_path.insert(abs_path(paths[i]).value());
+      }
+
+      if (uniq_abs_path.size() == 1)
+      {
+        cmd[0] = paths.front().c_str();
+        return;
+      }
+      else
+      {
+        throw std::runtime_error(
+            "path '" + cmd[0] + "' must refer to a unique binary but matched " +
+            std::to_string(paths.size()) + " binaries");
+      }
       return;
   }
 
--- a/tests/child.cpp
+++ b/tests/child.cpp
@@ -1,5 +1,7 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
+#include <fstream>
+#include <iostream>
 #include <sstream>
 #include <string>
 
@@ -11,6 +13,16 @@
 #include "child.h"
 #include "childhelper.h"
 #include "utils.h"
+
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace std_filesystem = std::filesystem;
+#elif __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace std_filesystem = std::experimental::filesystem;
+#else
+#error "neither <filesystem> nor <experimental/filesystem> are present"
+#endif
 
 namespace bpftrace {
 namespace test {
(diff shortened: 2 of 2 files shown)


Rtoax:
@danobi Seems all is well. please review. thanks.

src/child.cpp
```diff
@@ -103,6 +103,14 @@ static void validate_cmd(std::vector<std::string>& cmd)
     case 1:
       cmd[0] = paths.front().c_str();
       break;
+    case 2:
+      // /bin maybe is a symbolic link to /usr/bin (/bin -> /usr/bin)
+      if (get_absolute_path(paths[0]) == get_absolute_path(paths[1]))
```
viktormalik, line 108: Use an existing helper function `abs_path`:
```suggestion
      if (abs_path(paths[0]) == abs_path(paths[1]))
```

viktormalik (changes requested):

src/child.cpp
Rtoax: Sorry, Just fix it, thank you.

viktormalik:
Can we create a test for this?

Rtoax:
> 

Is this necessary? This PR only solves the error when /usr/bin/sleep and /bin/sleep are the same executable file. How should this situation be deployed in the CICD environment?

```
$ ls -l /bin
lrwxrwxrwx. 1 root root 7 Aug 10  2021 /bin -> usr/bin
$ realpath /bin/sleep
/usr/bin/sleep
```

This problem may not exist in different environments (including CICD), so I suggest not to add a test case. What do you think?

Rtoax:
I just force push a modification. There may be worse cases like

```
  $ realpath /usr/bin/ping /bin/ping /usr/sbin/ping /sbin/ping
  /usr/bin/ping
  /usr/bin/ping
  /usr/bin/ping
  /usr/bin/ping

  $ sudo bpftrace -e 'kprobe:do_nanosleep { printf("%d sleeping\n", pid); }' -c 'ping -c 4 0'
  ERROR: Failed to fork child: path 'ping' must refer to a unique binary but matched 4 binaries
```
So instead of using `case 2` to solve it, it is better to use `std::unordered_set<std::string>+abs_path()` to solve this problem.

danobi:
The fix looks good to me. But I agree with Viktor -- test here would be good. But rather than depend on CI environment (which is not reliable like you mention), we could create a tempdir and then populate it with a an executable shell script and some symlinks to it. I think that should be reliable.

Rtoax:
> The fix looks good to me. But I agree with Viktor -- test here would be good. But rather than depend on CI environment (which is not reliable like you mention), we could create a tempdir and then populate it with a an executable shell script and some symlinks to it. I think that should be reliable.

I have a question, shall we add a test example in C, such as the following code in tests/child.cpp:

```c
TEST(childproc, match_multi_exe)
{
  char path[] = "rootfs/";
  // make temp rootfs directory
  // make /usr/bin/ under rootfs
  // create tmp shell script "sleep.sh" (with x permission)
  // create link: /bin -> /usr/bin
  // put the rootfs/usr/bin and rootfs/bin to env
  // THEN, test sleep.sh
}
```

right? I'm not sure if my idea is accurate, if not, please give me some advice. thanks.

danobi:
Yeah that looks right to me

Rtoax:
> Yeah that looks right to me

OK, I'll try it.

Rtoax:
@danobi @viktormalik I just add test `childproc.multi_exec_match`, please review.

```
        $ sudo ./tests/bpftrace_test --gtest_filter=childproc.multi_exec_match
        Note: Google Test filter = childproc.multi_exec_match
        [==========] Running 1 test from 1 test suite.
        [----------] 1 test from childproc
        [ RUN      ] childproc.multi_exec_match
        [       OK ] childproc.multi_exec_match (10 ms)
        [----------] 1 test from childproc (10 ms total)
    
        [==========] 1 test from 1 test suite ran. (10 ms total)
        [  PASSED  ] 1 test.
```

tests/child.cpp
```diff
@@ -204,6 +206,88 @@ TEST(childproc, ptrace_child_term_before_execve)
...
 
+TEST(childproc, multi_exec_match)
+{
+  std::string tmpdir = "/tmp/bpftrace-test-child-XXXXXX";
+
+  // Create directory for test
+  if (::mkdtemp(&tmpdir[0]) == nullptr)
+  {
+    throw std::runtime_error("creating temporary path for tests failed");
+  }
```
danobi, line 217: Prefer:

```
EXPECT_NE(::mkdtemp(..), nullptr);
```

tests/child.cpp
```diff
@@ -204,6 +206,88 @@ TEST(childproc, ptrace_child_term_before_execve)
...
+  // Create symbol link: bin -> usr/bin
+  char cwd[512];
+  if (::getcwd(cwd, sizeof(cwd)) == NULL)
+  {
+    throw std::runtime_error("Failed to getcwd().");
+  }
+  if (::chdir(path.c_str()))
+  {
+    throw std::runtime_error("Failed to chdir().");
+  }
```
danobi, line 243: I don't think chdir is necessary, right? Can symlink two absolute paths

src/child.cpp
```diff
@@ -104,9 +104,30 @@ static void validate_cmd(std::vector<std::string>& cmd)
...
+      std::unordered_set<std::string> uniq_abs_path;
+      for (unsigned int i = 0; i < paths.size(); i++)
+      {
+        uniq_abs_path.insert(abs_path(paths[i]).value());
+      }
+
+      if (uniq_abs_path.size() == 1)
+      {
+        cmd[0] = paths.front().c_str();
+        return;
```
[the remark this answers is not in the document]
Rtoax: - Replace `return` with `break`.
- `case 1` will running line 134(`if (cmd.size() >= (maxargs - 1))`), right?

tests/child.cpp
Rtoax: Thanks.

tests/child.cpp
Rtoax: If `EXPECT_EQ(::symlink(usr_bin.c_str(), symlnk_bin.c_str()), 0);`, will got this:
```
# sudo tree /tmp/bpftrace-test-child-*
/tmp/bpftrace-test-child-GIcs6T
├── bin -> /tmp/bpftrace-test-child-GIcs6T/usr/bin
└── usr
    └── bin
        └── mysleep
```

However, we should keep the 'rootfs' this way:

```
# sudo tree /tmp/bpftrace-test-child-*
/tmp/bpftrace-test-child-rMpfBC
├── bin -> usr/bin
└── usr
    └── bin
        └── mysleep
```

Like the `/` did (`bin -> usr/bin`):

```
$ ls -l /bin
lrwxrwxrwx. 1 root root 7 Aug 10  2021 /bin -> usr/bin
```
Thus, i just `chdir -> link -> chdir`. What do you think?

src/utils.h
```diff
@@ -14,6 +14,16 @@
...
 
+#if __has_include(<filesystem>)
+#include <filesystem>
+namespace std_filesystem = std::filesystem;
+#elif __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace std_filesystem = std::experimental::filesystem;
+#else
+#error "neither <filesystem> nor <experimental/filesystem> are present"
+#endif
```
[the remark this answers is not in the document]
Rtoax: Fixed, thanks.

danobi (approved):
thanks for iterating on this!

PR #2712, opened 2023-08-13
test: codegen: Make kfunc targets determinstic

danobi:
Before, codegen tests tried to look at the host's actual tracefs for available_filter_functions. This would silently fail in most developer envs as we don't usually run codegen tests as root. However, in CI, they actually ran as root so reads succeeded. This caused difficult-to-debug skew.

Fix by mocking out the filesystem access so tests are consistent no matter where they are run.

##### Checklist


ajor (approved):
Nice 👌

PR #2714, opened 2023-08-13
test: codegen: Test both codegen paths for array_integer_equal_comparison

danobi:
Before, there was some feature checking logic that did not look
correct. In either case, only one codegen path was being tested.

This PR refactors some infra and then in the final commit fixes
`array_integer_equal_comparison` to test both possibilities.

##### Checklist


ajor (approved):
Nice!

PR #2715, opened 2023-08-14
add notes on built-ins' invocation mode

### LINKED ISSUE #2646: Spurious lost events warnings + duplicate/out-of-order/silent lost events
After the changes in #2475, I'm seeing a lot of spurious lost event warnings.

We should sort this out before pushing the next release.

### What reproduces the bug?
```
# bpftrace -e 'kretprobe:vfs_read { printf("func: '%s'\n", func); exit(); }' 
Attaching 1 probe...
func: vfs_read
Lost 4595 events
```

xh4n3:
The current implementation of `bpf_ringbuf_output` is causing some different behaviors from the old `bpf_perf_event_output`. However, these differences seem unavoidable for performance and bpf API reasons.

This PR adds notes on built-in functions' invocation modes, to explain why these seemingly erroneous behaviors occur, and how to avoid them.

This should close #2646 .

##### Checklist


fbs:
Isn't this already in the manpage, if not we should add it there and maybe link to it?

The change itself is ok, just worried we keep putting more and more in the 'quick reference guide' which instead belongs in the manpage (or maybe multiple manpages, idk)

xh4n3:
@fbs Thanks for the comment! I saw the reference guide more like a glossary rather than a quick guide, 🤣 .

I agree that we should put it somewhere else. But I'm not sure if we should put an manpage link inside a markdown page. Could it be put into the `internals_development.md`?

fbs:
internals_development is meant for contributors, describing internals.

A link should be fine, github renders the adoc manpage quite well. For those viewing offline its not that useful, but we could just say `see bpftrace(8)`

xh4n3:
@fbs I've updated the commits.

viktormalik (approved):
> @fbs Thanks for the comment! I saw the reference guide more like a glossary rather than a quick guide, 🤣 .

We've already discussed this in the office hours and I believe that the consensus was that both the reference guide and the manpage should be built from the same source. But we can surely iterate on that decision, let's discuss again next time and ideally come with clear instructions on what to put where.

danobi (approved):

PR #2719, opened 2023-08-22
add a MapError async event type for join()

xh4n3:
This PR borrowed some content from #1801 by @Birch-san.

I added a MapError event type, which will be raised and terminate execution when join() failed to find the join map.

This PR is one of the prerequisites for the mentioned refactor of the printf function in #2651. I'll be implementing the refactor. But I'm splitting it into smaller PRs for quicker review and merging, if you don't mind.

I haven't update test cases. If we all found it feasible, I will update the test cases accordingly. Thanks!

##### Checklist

### DIFF
--- a/src/ast/async_event_types.cpp
+++ b/src/ast/async_event_types.cpp
@@ -104,5 +104,13 @@ std::vector<llvm::Type*> SkbOutput::asLLVMType(ast::IRBuilderBPF& b)
   };
 }
 
+std::vector<llvm::Type*> MapError::asLLVMType(ast::IRBuilderBPF& b)
+{
+  return {
+    b.getInt64Ty(), // asyncid
+    b.getInt32Ty(), // map id
+  };
+}
+
 } // namespace AsyncEvent
 } // namespace bpftrace
--- a/src/ast/async_event_types.h
+++ b/src/ast/async_event_types.h
@@ -127,5 +127,13 @@ struct SkbOutput
   std::vector<llvm::Type*> asLLVMType(ast::IRBuilderBPF& b);
 } __attribute__((packed));
 
+struct MapError
+{
+  uint64_t action_id;
+  uint32_t map_id;
+
+  std::vector<llvm::Type*> asLLVMType(ast::IRBuilderBPF& b);
+} __attribute__((packed));
+
 } // namespace AsyncEvent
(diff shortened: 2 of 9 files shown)


viktormalik (commented):
Agreed with @danobi here, do we really need to introduce the `MapError` as an async event?

I like the idea of factoring out a helper to get a "scratch" map but don't think that we need to propagate its errors in such a complex way.

viktormalik:
> In essence, I think this is primarily for bpftrace developers now.

Maybe we could hide the error path behind some debugging flag or env variable, then?

But yes, we should absolutely cover this with tests rather than propagating internal bpftrace errors to the user.

xh4n3:
> Maybe we could hide the error path behind some debugging flag or env variable

This is a viable solution, but may bring some complexity.

How do you think? @danobi

danobi:
Sorry about delay -- was out the last couple days.

I agree that debugability is important. If we agree on not propagating errors to user, how about a percpu counter on lookup failure or something? We could make the event loop check the counters only if `-v` (or something) is used.

xh4n3:
@danobi @viktormalik Thank you, so we reached the consensus that:
1. debugability is important, i.e. we should do something when bpftrace hit a bug and cannot find the corresponding map value
2. let's not complicate user's interface, i.e. do not output errors that user should not care about

> If we agree on not propagating errors to user, how about a percpu counter on lookup failure or something? We could make the event loop check the counters only if -v (or something) is used.

Sorry but IMO, this will let one error path relying on another possible error path, since they're both looking up the potentially missing map value.

Combining your ideas, I wonder if I can do this:
* use the trace pipe (via bpf_trace_printk) for these debugging purpose output, instead of using the MapError async method
* when we pass "-v" (or a new Env var) to the bpftrace, poll outputs from the trace pipe

danobi:
Makes sense about layering dependencies. Good point.

Using trace pipe also seems reasonable. It'd be good to check the exact semantics of the trace pipe. For example, if some other application was using the pipe, will bpftrace polling from it clear events from that third party?

xh4n3:
@danobi Thanks for the comments, I implemented this in the new PR https://github.com/iovisor/bpftrace/pull/2746

I've also included its side effects when running with some third party applcations.

PR #2720, opened 2023-08-23
Add FUNDING.yml

danobi:
As discussed during office hours, it would be useful if we have a way for sponsors to donate money. We may need to pay down some infra costs in the near term.

I looked at a few options and it looked like OpenCollective was the most straightforward. They can do fiscal hosting which means they'll deal with all the tax stuff. But in exchange they take 10%. Not too bad (but also not 0%).

This is the collective: https://opencollective.com/bpftrace

##### Checklist

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+open_collective: bpftrace


ajor (approved):
Sounds reasonable to me

PR #2725, opened 2023-08-25
tests: updating dependency for builtin tests

dubeyabhishek:
These dependency failures were caught on other OS variants.


danobi (approved):
lgtm, thanks!

PR #2726, opened 2023-08-25
docs/usdts: add info on targeting all shared libraries for a process

jordalgo:
Just a docs update.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1359,7 +1359,7 @@ usdt:library_path:probe_name
 usdt:library_path:[probe_namespace]:probe_name
 ```
 
-Where `probe_namespace` is optional if `probe_name` is unique within the binary.
+Where `probe_namespace` is optional if `probe_name` is unique within the binary. You can also target all shared libraries and the main executable in a process's address space if you use the `*` wildcard in place of the `binary_path`/`library_path` AND pass in a pid command line argument e.g. `bpftrace -e 'usdt:*:loop { printf("hi\n"); }' -p 1234`.
 
 Examples:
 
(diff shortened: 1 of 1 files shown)


docs/reference_guide.md
```diff
@@ -2990,7 +2990,7 @@ Return true if the string haystack contains the string needle, and zero otherwis
 Examples:
 
 ```
-bpftrace -e 't:syscalls:sys_enter_execve /strcontains(str(args.filename),"bin")/ { @[comm, str(args.filename)] = count(); }'
```
viktormalik, line 2993: Don't update lines not relevant for the PR please.

docs/reference_guide.md
```diff
@@ -1359,7 +1359,7 @@ usdt:library_path:probe_name
 usdt:library_path:[probe_namespace]:probe_name
 ```
 
-Where `probe_namespace` is optional if `probe_name` is unique within the binary.
+Where `probe_namespace` is optional if `probe_name` is unique within the binary. You can also target all shared libraries and the main executable in a process's address space if you use the `*` wildcard in place of the `binary_path`/`library_path` AND pass in a pid command line argument e.g. `bpftrace -e 'usdt:*:loop { printf("hi\n"); }' -p 1234`.
```
viktormalik, line 1362: This applies to uprobes, too, doesn't it? Shouldn't it be placed elsewhere, then?

jordalgo: Good point. How about in the probes intro section or perhaps I should add a new "-p" item in the Usage section that points this out.

viktormalik: Either *Probes* intro or a new subsection of *Probes* seems the best place to me. I'm also wondering if we should reflect this in manpage, too (we should really put some work to unify these two).

jordalgo: Actually, this method of targeting doesn't work for uprobes/uretprobes. Anyway, going to abandon in favor of [this PR](https://github.com/iovisor/bpftrace/pull/2734)

PR #2734, opened 2023-08-30
USDT: Enable targeting all pids/paths on a host

jordalgo:
This utilizes the single wildcard `*` for the binary_path/library_path value to target all running pids on the host.

##### Checklist

### DIFF
--- a/src/usdt.cpp
+++ b/src/usdt.cpp
@@ -1,5 +1,6 @@
 #include "usdt.h"
 #include "log.h"
+#include "utils.h"
 
 #include <signal.h>
 
@@ -88,9 +89,9 @@ std::optional<usdt_probe_entry> USDTHelper::find(int pid,
   }
 }
 
-usdt_probe_list USDTHelper::probes_for_pid(int pid)
+usdt_probe_list USDTHelper::probes_for_pid(int pid, bool print_error)
 {
-  read_probes_for_pid(pid);
+  read_probes_for_pid(pid, print_error);
 
   usdt_probe_list probes;
   for (auto const &path : usdt_pid_to_paths_cache[pid])
@@ -100,6 +101,19 @@ usdt_probe_list USDTHelper::probes_for_pid(int pid)
       probes.insert(probes.end(),
                     usdt_probes.second.begin(),
                     usdt_probes.second.end());
+    }
+  }
+  return probes;
+}
+
+usdt_probe_list USDTHelper::probes_for_all_pids()
+{
+  usdt_probe_list probes;
+  for (int pid : bpftrace::get_all_running_pids())
+  {
+    for (auto &probe : probes_for_pid(pid, false))
+    {
+      probes.push_back(std::move(probe));
     }
   }
   return probes;
@@ -119,7 +133,7 @@ usdt_probe_list USDTHelper::probes_for_path(const std::string &path)
   return probes;
 }
 
-void USDTHelper::read_probes_for_pid(int pid)
+void USDTHelper::read_probes_for_pid(int pid, bool print_error)
 {
   if (pid_cache.count(pid))
     return;
@@ -129,10 +143,13 @@ void USDTHelper::read_probes_for_pid(int pid)
     void *ctx = bcc_usdt_new_frompid(pid, nullptr);
     if (ctx == nullptr)
     {
-      LOG(ERROR) << "failed to initialize usdt context for pid: " << pid;
+      if (print_error)
+      {
+        LOG(ERROR) << "failed to initialize usdt context for pid: " << pid;
 
-      if (kill(pid, 0) == -1 && errno == ESRCH)
-        LOG(ERROR) << "hint: process not running";
+        if (kill(pid, 0) == -1 && errno == ESRCH)
+          LOG(ERROR) << "hint: process not running";
+      }
 
       return;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to
   - [#2692](https://github.com/iovisor/bpftrace/pull/2692)
 - Support casting int <-> int array
   - [#2686](https://github.com/iovisor/bpftrace/pull/2686)
+- Support targeting all running processes for USDTs
+  - [#2734](https://github.com/iovisor/bpftrace/pull/2734)
 #### Changed
 - Make `args` a structure (instead of a pointer)
   - [#2578](https://github.com/iovisor/bpftrace/pull/2578)
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -1398,6 +1398,8 @@ usdt:library_path:[probe_namespace]:probe_name
 
 Where `probe_namespace` is optional if `probe_name` is unique within the binary.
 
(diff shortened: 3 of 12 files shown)


jordalgo:
In terms of the API for targeting all pids, the other option I was considering was with the `-p` argument (perhaps by setting this to -1 but that seems less intuitive). 

Also, in terms of testing, I added a runtime test but not sure where else I should add some tests.

src/usdt.cpp
```diff
@@ -129,10 +143,13 @@ void USDTHelper::read_probes_for_pid(int pid)
     void *ctx = bcc_usdt_new_frompid(pid, nullptr);
     if (ctx == nullptr)
     {
-      LOG(ERROR) << "failed to initialize usdt context for pid: " << pid;
+      if (print_error)
```
jordalgo, line 146: Added this because the error logs are very noisy when targeting all processes on a host.

viktormalik (commented):
Thanks, this looks useful.

I found an issue when testing the PR locally:
```
# bpftrace -e 'usdt:*:tracetest:testprobe { printf("hit\n"); exit(); }'
Parse error:
    -4@%ebx 8@seq_iter.6(%rip)
----------------------^
Parse error:
    -4@%ebx 8@seq_iter.6(%rip)
----------------------^
Parse error:
    -4@%ebx 8@seq_iter.6(%rip)
----------------------^
Parse error:
    -4@%ebx 8@seq_iter.6(%rip)
----------------------^
[...]
^C
```
Looks like the problem is that you're not handling wildcarded target in [semantic_analyser.cpp:2749](https://github.com/iovisor/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L2749) and it tries to find probes for all binary paths on the system.

As for the additional tests, you could add one in `tests/bpftrace.cpp` to verify that all the expected probes are created. You'll need to mock some of the new USDT helper functions, though. See e.g. `add_probes_usdt_wildcard` in that file for inspiration.

jordalgo:
@viktormalik - Thanks for the review. Added a unit test and also updated semantic_analyser.cpp. I wasn't seeing that parser error in my local testing but adding the check for the single wildcard there makes sense.

viktormalik (approved):
LGTM, although I'll wait for another review before merging this.

Thanks!

danobi:
I was under the impression if you did not provide a PID the kernel will effectively attach the probe to all processes that use the underlying binary. For example, see this runtime test: https://github.com/iovisor/bpftrace/blob/420b053730846960f8c6afca3a934a602448251b/tests/runtime/usdt#L65-L70

Or maybe I am misremembering or missing something?

danobi:
What is the use case for this feature? IIRC I saw you comment on another ticket about supporting multiple `-p` arguments to bpftrace. Wondering cuz I'd like to understand the tradeoffs. I'm a bit worried that creating a probe for each pid on the system will be very slow or perhaps cause bpftrace to use a lot of memory and cause issues.

jordalgo:
> What is the use case for this feature?

The way I see it, it's a feature of convenience. If you have one or more USDTs that may be active in multiple pids (or even just one) and you don't want to have to find the path then you can just use the single wildcard and bpftrace does the right thing. 

> I'm a bit worried that creating a probe for each pid on the system will be very slow or perhaps cause bpftrace to use a lot of memory and cause issues

I haven't found it to be slow in my local testing but I also don't know enough about how much memory this might take up or the strain it puts on other processes (in terms of hogging resources). I can do some light profiling of bpftrace when using the single wildcard - what metrics should I be looking at aside from heap size or rss for bpftrace?

danobi:
> > What is the use case for this feature?
> 
> The way I see it, it's a feature of convenience. If you have one or more USDTs that may be active in multiple pids (or even just one) and you don't want to have to find the path then you can just use the single wildcard and bpftrace does the right thing.

Ack.
 
> > I'm a bit worried that creating a probe for each pid on the system will be very slow or perhaps cause bpftrace to use a lot of memory and cause issues
> 
> I haven't found it to be slow in my local testing but I also don't know enough about how much memory this might take up or the strain it puts on other processes (in terms of hogging resources). I can do some light profiling of bpftrace when using the single wildcard - what metrics should I be looking at aside from heap size or rss for bpftrace?

Max RSS would be a good one I think. Have had issues in past with that. Also maybe check if `read_probes_for_pid()` gets hot on cpu?

jordalgo:
> Max RSS would be a good one I think. Have had issues in past with that. Also maybe check if read_probes_for_pid() gets hot on cpu?

Not noticing much of a difference here for the different target types:
```
sudo bpftrace -e 'usdt:*:jordalgo_tools:* { printf("hit\n"); }'

VmPeak:	  193232 kB
VmSize:	  193232 kB
VmHWM:	  101432 kB
VmRSS:	  101432 kB
RssAnon:	   46048 kB
RssFile:	   55384 kB


sudo bpftrace -e 'usdt:*:jordalgo_tools:* { printf("hit\n"); }' -p 1815647

VmPeak:	  189308 kB
VmSize:	  189308 kB
VmHWM:	  101380 kB
VmRSS:	  101380 kB
RssAnon:	   42664 kB
RssFile:	   58716 kB


sudo bpftrace -e 'usdt:/proc/1815647/exe:jordalgo_tools:* { printf("hit\n"); }'

VmPeak:	  189308 kB
VmSize:	  189308 kB
VmHWM:	  101528 kB
VmRSS:	  101528 kB
RssAnon:	   42664 kB
RssFile:	   58864 kB
```

For `read_probes_for_pid` - Correct me if I'm wrong but I think the real slow-down comes from `get_symbols_from_usdt`. 
For this program (on my local machine) with pid targeting:
```
sudo bpftrace -e 'usdt:*:jordalgo_tools:* { printf("hit\n"); }' -p 1815647
```
387 candidate probes are generated. Whereas without:
```
sudo bpftrace -e 'usdt:*:jordalgo_tools:* { printf("hit\n"); }'
```
777,481 candidate probes are generated and fed into the stream that then goes to `get_matches_in_stream`. But even then we only create 2 probes because they are the only 2 that match. Sure, if someone were do something bonkers like `usdt:*:*:*` this would be a problem but I don't consider this a foot-gun - it's just increasing flexibility.

viktormalik:
> Sure, if someone were do something bonkers like `usdt:*:*:*` this would be a problem but I don't consider this a foot-gun - it's just increasing flexibility.

Agreed, we already have several ways to make attachment take forever (try `kfunc:*`) but it's up to the user to avoid such situations. If the probe matcher is fast enough to process the candidate entries (and it seems that it is), then I'm in favor of merging this as it may simplify user's lives.

jordalgo:
@danobi No problem. Thanks for the review!

BTW, expect another PR to add similar functionality for uprobes/uretprobes as I feel like they should work similarly to USDTs.

jordalgo:
> Looks like we're getting some flakiness on other PRs:
> 
> ```
> 
> 2: [ RUN      ] usdt.usdt probes - attach to fully specified probe all pids
> 2: [  TIMEOUT ] usdt.usdt probes - attach to fully specified probe all pids
> 2: 	Command: ../src//bpftrace -e 'usdt:*:tracetest:testprobe { printf("here\n" ); exit(); }'
> 2: 	Timeout: 5
> 2: 	Current output: 
> ```

Bummer. Think just increasing the timeout to 10 or 20 would fix?

dubeyabhishek:
@danobi @viktormalik just bumping the timeout in usdt test is not sufficient on my system. I need to increase ATTACH_TIMEOUT in engine/runner.py to 30 seconds for test to work.

danobi:
Could we run that particular test in a separate pid namespace? It would still exercise the business logic without varying widely between different systems.

jordalgo:
> Could we run that particular test in a separate pid namespace?

Good idea. Let me try that.

jordalgo:
@danobi I might need some assistance with this. I tried this, which works locally, but seems to have issues in CI - also it feels pretty hacky: https://github.com/iovisor/bpftrace/pull/2765/files

Issue #2721, opened 2023-08-23
Re-examine the embedded builds


danobi:
The embedded builds were added by @dalehamel quite a while ago. It looks like our CI supports a few different embedded configurations. I am no longer sure which configurations people are actually using downstream. To the best of my knowledge, the one big desirable downstream is the fully static bpftrace binary. So maybe this will be the only embedded configuration we support in the future?
    
This task is to re-examine the embedded builds and see which can be removed or reworked. I am cautiously optimistic that we can use https://github.com/rsms/llvmbox to build a fully static bpftrace binary.

It looks like the embedded build only supports LLVM 10 and LLVM 12. I am sure we are all seriously dreading the day when we need to upgrade it.

danobi:
So `nix bundle` functions but has two downsides:

1. It takes ~30s for first run (due to self-extraction). Subsequent runs take ~2-3s.
2. It's a ~300 MB binary (~10x size of static binary today)

I looked around more and found that you can use a bundlers: https://github.com/ralismark/nix-appimage

With the appimage, the size is still roughly the same but the first run is down to ~1s. Subsequent runs are identical to the first, so still ~1s.

I think 10x binary footprint is a worthy tradeoff for how much code we can delete. The static builds are immensely complicated and I don't think we can maintain it long term.

NB: all the above options produce a fully hermetic "static" binary that can run on any linux system.

viktormalik:
> I think 10x binary footprint is a worthy tradeoff for how much code we can delete. The static builds are immensely complicated and I don't think we can maintain it long term.

I completely agree. Static builds are notoriously painful so if it's "only" the 10x binary footprint growth that we need to pay to get rid of them, I'm in favor of doing it.

fbs:
Would that mean shipping a 300MB binary with every release?

danobi:
Yes, as well as for every commit on master (our quay images). The exact mechanism for master builds distribution is TBD.

But note that 300MB is the _starting_ size. If you inspect the appimage, there is a lot of useless stuff held inside:

```
$ ./result --appimage-mount
/tmp/.mount_resultMJnObk
```

In another terminal:

```
$ cd /tmp/.mount_resultMJnObk

$ du -hs .
1.8G    .

$ ls -lh
total 46K
-r-xr-xr-x 1 root root 46K Dec 31  1979 AppRun
[...]

PR #2736, opened 2023-08-31
Support embedded nix builds

danobi:
This adds support for building bpftrace as an appimage. See the
documentation update in nix.md in the commit for usage.

This is intended to replace the current semi-static embedded build.
This approach has both pros and cons over the semi-static embedded build.
For one, the appimage is actually fully static (as it bundles the
necesssary libc). So it's more likely to work on more systems.

The downside is that the appimage is a ~10x larger binary than the
semi-static binary. The appimage also requires fuse to work (which most
systems should support). It's takes ~1s longer to start up which I don't
think is a huge deal.

Note that I had to apply some patches to nix-appimage (upstreamed
in https://github.com/ralismark/nix-appimage/pull/8). As mentioned in
that PR description, I'm not 100% sure yet why the changes work.

I'd like to spend some time deep diving on that before we merge this.
But in the meantime, please give this a try.

This closes https://github.com/iovisor/bpftrace/issues/2721.

##### Checklist

### DIFF
--- a/docs/nix.md
+++ b/docs/nix.md
@@ -46,6 +46,17 @@ $ sudo ./result/bin/bpftrace --info 2>&1 | grep LLVM
   LLVM: 13.0.1
 ```
 
+### Build bpftrace as a statically linked binary
+
+```
+$ nix build .#appimage
+$ ldd ./result
+        not a dynamic executable
+$ sudo ./result -e 'BEGIN { print("static!"); exit() }'
+Attaching 1 probe...
+static!
+```
+
 ### Don't use Nix to build, but rather only manage dependencies
 
 ```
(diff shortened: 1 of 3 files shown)


danobi:
Note that if we decide to go this direction, I will try to delete all the embedded build code as a followup. I expect the deletions to be quite substantial.

viktormalik:
Looks good to me. I was able to build the appimage with `nix build .#appimage` and it executed on a freshly installed VM without any problem. Nice :-)

As I already wrote elsewhere, the footprint size increase is a reasonable price to pay for the simplicity and robustness this brings.

danobi:
I understand why the nix-appimage PR works now. See the comment I left on it. I'll also have a longer form write up in a while. An interesting problem :)

danobi:
Since this is relatively harmless, I shall merge after the weekend if nobody else objects.

Issue #2626, opened 2023-05-25
ci: Nix-ify CI


danobi:
The goal of this project is to make the CI easier to maintain and also trivially
runnable locally. The existing CI has evolved over the years to be quite complex
and difficult to reproduce or modify.

To that end, this project explores Nix as a replacement for most of the docker
based CI.

Task list:
- Set up initial nix flake (#2595)
- #2627
  - Ideally all the CI logic is held outside of the YAML. So GHA only runs one shell command that can be easily done locally. 
- #2743 
  - So we have precise dependency control
- #2721
  - The embedded builds were added by @dalehamel quite a while ago. It looks like our CI supports a few different embedded configurations. I am no longer sure which configurations people are actually using downstream. To the best of my knowledge, the one big desirable downstream is the fully static bpftrace binary. So maybe this will be the only embedded configuration we support in the future?
  - This task is to re-examine the embedded builds and see which can be removed or reworked. I am cautiously optimistic that we can use https://github.com/rsms/llvmbox to build a fully static bpftrace binary.
  - It looks like the embedded build only supports LLVM 10 and LLVM 12. I am sure we are all seriously dreading the day when we need to upgrade it.
- #2741
  - Non-comprehensive list of things to look at:
    - `build-debug.sh`
    - `build-docker-image.sh`
    - `build-libs.sh`
    - `build-release.sh`
    - `build-static.sh`
    - `build.sh`
    - `docker/build.sh`
    - `docker/Dockerfile.ubuntu-glibc`
    - `bcc/` (submodule)
    - `libbpf/` (submodule)
- Update user facing docs with instructions on downloading bpftrace build artifact
- #2609
  - Need to update `scripts/create-assets.sh` to build and use appimage
  - Just create a workflow that listens for releases and then uploads assets
- #2756 
  - The Dockerfiles served as good living documentation on the required dependencies to build bpftrace
  - We should keep them updated to the latest distro release
  - Can probably condition docker builds in CI on changes to cmake files (to reduce unnecessary builds in PRs)

danobi:
Yay, it's done

PR #2742, opened 2023-09-02
embedded: Delete support

### LINKED ISSUE #2741: Delete deprecated scripts / tools
Non-comprehensive list of things to look at:

    build-debug.sh
    build-docker-image.sh
    build-libs.sh
    build-release.sh
    build-static.sh
    build.sh
    docker/build.sh
    docker/Dockerfile.ubuntu-glibc
    bcc/ (submodule)
    libbpf/ (submodule)

danobi:
## PR description

This builds on #2736 in that we will rely on the appimage to do for us
what the embedded build used to do. That is, to provide a static binary.

Supporting and maintaining the embedded builds was very difficult,
as the combinations between distros were essentially endless. This
led to a maze of difficult-to-understand-and-modify build and ci code.
Unifying on the nix build will provide both simplicity and reproducibility.

## A farewell

But let us not go without saying how awesome the embedded binaries
were. They were extremely useful on countless occasions for myself alone.
As I'm sure they were for other people. But engineering is full of tradeoffs
and I think it's time to let the embedded builds ride off into the sunset.

Hopefully the appimage can live up to the legacy left behind by the static
build.

This closes #2741.

##### Checklist


danobi:
I've pulled out two of the more independent changes into https://github.com/iovisor/bpftrace/pull/2744 and https://github.com/iovisor/bpftrace/pull/2745 .

viktormalik (commented):
This is an awesome cleanup!

Main README also contains mentions of `build-static.sh` (section Development), you'll want to remove that, too.

Also, I'd like to clarify what is the currently recommended development workflow. Previously, we had the `bcc` and `libbpf` submodules, `build-libs.sh` to build them locally, and we statically linked bpftrace against them. This made sure that we were always developing for a single BCC/libbpf version (the latest one) and didn't have to deal with supporting multiple versions that we did before.

Do we have something similar with nix? Can we document that somewhere?

danobi:
> This is an awesome cleanup!
> 
> Main README also contains mentions of `build-static.sh` (section Development), you'll want to remove that, too.

Ack. Good catch.


> Also, I'd like to clarify what is the currently recommended development workflow. Previously, we had the `bcc` and `libbpf` submodules, `build-libs.sh` to build them locally, and we statically linked bpftrace against them. This made sure that we were always developing for a single BCC/libbpf version (the latest one) and didn't have to deal with supporting multiple versions that we did before.
> 
> Do we have something similar with nix? Can we document that somewhere?

The flake defines the exact libbpf and bcc version to use: https://github.com/iovisor/bpftrace/blob/e6e4318a2005992df121c6d25412c725e1cfb6d1/flake.nix#L24-L60

It downloads and builds from github.

For people on bleeding edge distros (like me on arch linux), I don't expect the flake will be required. The distro libbpf + bcc should be new enough. But for folks on more stable distros, maybe we could recommend them use `nix develop` shell? 

They are also welcome to install libbpf + bcc from source too.

If that sounds reasonable I can document that in the developers.md doc

danobi:
I updated the README and removed all the references to `build-libs.sh` and recursive cloning in INSTALL.md. Also removed the docker bits in INSTALL.md.

viktormalik (commented):
Couple of comments:
- Yesterday at the office hours, we discussed that the installation instructions for various distros in `INSTALL.md` are often outdated b/c some packages were renamed, etc. The original purpose of some Dockerfiles (e.g. `Dockerfile.fedora30`) was to simplify maintaining those instructions. While they're probably outdated, too, we plan to update them to serve their original purpose so I'm not sure that deleting them is the best approach. I think we should eventually end up with something like `Dockerfile.fedora`, `Dockerfile.ubuntu`, `Dockerfile.alpine`, etc., each for the latest version of the distro.
- I may be missing something but I believe that this breaks our current process of building and pushing images to quay. Should we update `docs/release_process.md` until we fix it?

viktormalik:
> For people on bleeding edge distros (like me on arch linux), I don't expect the flake will be required. The distro libbpf + bcc should be new enough. But for folks on more stable distros, maybe we could recommend them use `nix develop` shell?
> 
> They are also welcome to install libbpf + bcc from source too.
> 
> If that sounds reasonable I can document that in the developers.md doc

Sounds good, would be nice to add it to `developers.md`.

danobi:
> Couple of comments:
> 
>     * Yesterday at the office hours, we discussed that the installation instructions for various distros in `INSTALL.md` are often outdated b/c some packages were renamed, etc. The original purpose of some Dockerfiles (e.g. `Dockerfile.fedora30`) was to simplify maintaining those instructions. While they're probably outdated, too, we plan to update them to serve their original purpose so I'm not sure that deleting them is the best approach. I think we should eventually end up with something like `Dockerfile.fedora`, `Dockerfile.ubuntu`, `Dockerfile.alpine`, etc., each for the latest version of the distro.

Agreed, Dockerfiles were good documentation. However, in the current (now deleted) state, they reflected fairly old distro releases with quite a few manually built packages (like libcereal and in certain cases pahole).

I've tacked https://github.com/iovisor/bpftrace/issues/2756 onto https://github.com/iovisor/bpftrace/issues/2626 . Mind if I take that on in a followup PR?

>     * I may be missing something but I believe that this breaks our current process of building and pushing images to quay. Should we update `docs/release_process.md` until we fix it?

Yes, you are right. My plan was to upload the appimage as a workflow artifact (see https://docs.github.com/en/actions/managing-workflow-runs/downloading-workflow-artifacts?tool=webui). I don't think it's too difficult -- let me add that to this PR and update docs

danobi:
I cleaned up the commits a bit as well as addressing feedback

viktormalik (approved):
Beautiful, this is so much simpler than the embedded builds!

FWIW, I think that this is quite an important change so I'd like to have at least one more review.

ajor:
So much simpler! I'll have to have a go building it on my computer, but the changes look reasonable to me.

The one concern I have is the size of the appimage vs the old static build and how that will work for embedded devices. Are we talking an appimage size of 500MB-1GB now?

Pinging @tnovak who's done the work on Android static builds to weigh in.

danobi:
@ajor As of https://github.com/iovisor/bpftrace/pull/2749, the appimage is down to ~140M. It would be pretty heavy for an embedded device, but ~45M (static binary) is in the same order of magnitude.

tnovak:
> Hopefully the appimage can live up to the legacy left behind by the static build.

I worry that the removal of `STATIC_LINKING` is a step back when it comes to Android support, where AppImage isn't a viable alternative.

Android platform doesn't provide all the necessary dependencies (such as a relatively recent LLVM) so we need to build them from source, and static linking allows us to produce a single binary with no external dependencies that can be pushed to any Android device that has root access and runs a fairly recent kernel.

A bit more context:

---

There are two ways of building bpftrace for Android that I'm familiar with:

- Integrating it into the AOSP build (i.e. building it as part of the Android system image). For example: [bpftrace_local_manifest](https://github.com/simpleton/bpftrace_local_manifest)
  - This method relies on manually written [Android blueprint](https://source.android.com/docs/setup/build) files (e.g. [Android.bp](https://github.com/simpleton/bpftrace/blob/0.17.0-o12/src/Android.bp)) for bpftrace and its dependencies; it doesn't use CMake at all. As such it's not officially supported by either the bpftrace project or AOSP, so upgrading can be painful (that might be why it's still on LLVM8).
- Cross-compiling using the NDK toolchain, which is the approach taken by the [ExtendedAndroidTools](https://github.com/facebookexperimental/ExtendedAndroidTools) project (see [cmake.mk](https://github.com/facebookexperimental/ExtendedAndroidTools/blob/c2bb54afc927043630cf969e84daf206c376e5ea/toolchain/cmake.mk) and [bpftrace/build.mk](https://github.com/facebookexperimental/ExtendedAndroidTools/blob/c2bb54afc927043630cf969e84daf206c376e5ea/projects/bpftrace/build.mk#L32)).
  - `ExtendedAndroidTools` builds libbpf and bcc separately, so it sets `USE_SYSTEM_BPF_BCC` instead of using vendored versions (partly because it predates vendored lib changes in bpftrace, but presumably also because some of these deps are not used only by bpftrace). We'll need to switch to vendored libs if/when `USE_SYSTEM_BPF_BCC` goes away.
  - It supports linking bpftrace both statically and dynamically. If doing the latter, it creates a package that contains bpftrace and all the libraries, as well as wrappers to set up `LD_LIBRARY_PATH` and other stuff. This is handy when other tools (bcc, python) are needed as well, but for bpftrace, static linking is way more convenient in my experience.

Bpftrace 0.18.1 prebuilts for Android built using ExtendedAndroidTools can be found [here](https://github.com/facebookexperimental/ExtendedAndroidTools/actions/runs/6096182516). In terms of size, statically-linked binaries for armv7 and arm64 are ~33MB and ~49MB, respectively. Keeping these small is important (especially for devices that have limited storage and memory[*]), so these are stripped (with the exception of BEGIN/END triggers) and LLVM is built with only BPF target support.

[*] Of course, larger binary size doesn't necessarily translate into higher memory footprint, especially if most of the extra stuff is in sections that don't get loaded (e.g. debug info). On my armv7 device, a simple bpftrace one-liner causes around ~14.5MB out of 33MB to be resident. Even that can be problematic on low-RAM devices if the extra memory pressure caused by running bpftrace affects measurements.

danobi:
Thanks for the feedback @tnovak . Two things I'd like to clarify:

1. Why is appimage not viable for android? 
2. Are you using our pre-built static binaries at all? Or just the `STATIC_LINKING` build flag?

ajor (changes requested):
@danobi It seems like we'd be able to get a lot of the build simplifications by just removing the embedded build and leaving the `STATIC_LINKING` option alone. Would that make sense?

@tnovak When the `USE_SYSTEM_BPF_BCC` option goes away, it'll be in favour of always using the system libraries rather than the other way around, so you won't need to change anything for this part. (I think we could look into using CMake's FetchContent to get the best of both worlds here: use system libraries if installed, otherwise automatically vendor them)

tnovak:
Hey @danobi!

> Why is appimage not viable for android?

I'm not too familiar with AppImage, but a cursory search indicates that Android isn't supported out of the box; presumably because Android is just different enough from a typical Linux distribution that making it work would be non-trivial.

For example, Android uses its own dynamic linker (installed as `/system/bin/linker`) and `libc` so an AppImage built against `glibc` wouldn't run. (Even if one were to link against `glibc` statically, some of its functionality may not work -- for example, `getaddrinfo()` in Android is implemented by talking to a separate `netd` process).

In general, there are only [a few libraries](https://developer.android.com/ndk/guides/stable_apis) that are guaranteed to be available.

> Are you using our pre-built static binaries at all? Or just the STATIC_LINKING build flag?

Just `STATIC_LINKING` (and `USE_SYSTEM_BPF_BCC`). [This file](https://github.com/facebookexperimental/ExtendedAndroidTools/blob/c2bb54afc927043630cf969e84daf206c376e5ea/projects/bpftrace/build.mk) has the flags passed to CMake, in addition to those from `cmake.mk` linked above.

danobi:
> @danobi It seems like we'd be able to get a lot of the build simplifications by just removing the embedded build and leaving the `STATIC_LINKING` option alone. Would that make sense?

Yeah, that's was I was thinking too. I wouldn't mind keeping the `STATIC_LINKING` support in and having a light CI job that does a static build. That seems like it has a worthwhile value to maintenance ratio.

danobi:
Btw, we have an office hours this week if anyone wants to drop in and chat about this. Should be links to the meeting info in the main readme

danobi:
As discussed during office hours last week, we will keep `STATIC_LINKING` support in the build. And we will exercise it with a CI job (nix if possible, docker otherwise). But still remove embedded build pipeline.

danobi:
Ready for another look. I'm fairly pleased with what we've arrived at

.github/include/static.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Build the base image
+docker build -t "$IMAGE" -f docker/Dockerfile.static docker/
+
+# Perform bpftrace static build
+docker run -v $(pwd):$(pwd) -w $(pwd) -i "$IMAGE" <<'EOF'
+set -eux
+BUILD_DIR=build-static
+cmake -B build-static -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTING=OFF -DSTATIC_LINKING=ON
+make -C build-static -j$(nproc)
```
viktormalik, line 21:
```suggestion
cmake -B "$BUILD_DIR" -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_TESTING=OFF -DSTATIC_LINKING=ON
make -C "$BUILD_DIR" -j$(nproc)
```

viktormalik (approved):
Looks good, great cleanup! Just one nit, otherwise I'd be happy to get this in ASAP, a problem with embedded builds is blocking me from opening a PR.

.github/include/static.sh
danobi: Oops. Fixed.

danobi:
Sounds like this is final call then. I'll wait 48 hrs before merging.

There will also probably be a while where we can do a clean revert.

ajor (approved):
Looks good to me

PR #2746, opened 2023-09-04
Add the GetScratchMap and DebugOutput utilities

xh4n3:
This PR reimplements https://github.com/iovisor/bpftrace/pull/2719. The GetScratchMap part remains identical with the old PR.

I added a DebugOutput utility in this PR, for outputting unexpected + bpftrace internal error messages. The missing map value in the GetScratchMap utility is a typical use of it.

How to use:
```bash
# export BPFTRACE_DEBUG_OUTPUT=1
# bpftrace -e 'tracepoint:syscalls:sys_enter_execve { printf("hello world\n"); join(args.argv); }'
Attaching 1 probe...
DEBUG:            <...>-144298  [001] d...1 961288.443108: bpf_trace_printk: [BPFTRACE_DEBUG_OUTPUT] unable to find the scratch map value for join
hello world
^C
```

Note: This DEBUG utility *consumes* messages from the trace_pipe, so if some other applications are using the pipe, they may receive bpftrace's message or lost their own message. Since this is a *debug* purpose utility, and disabled by default, I think this is okay, better than the case that bpftrace hits error but nothing printed.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -405,7 +405,10 @@ CallInst *IRBuilderBPF::createGetScratchMap(int mapid,
   CreateCondBr(condition, lookup_merge_block, lookup_failure_block);
 
   SetInsertPoint(lookup_failure_block);
-  // TODO: should we return a loud error?
+  if (bpftrace_.debug_output_)
+    CreateDebugOutput("unable to find the scratch map value for " + name,
+                      std::vector<Value *>{},
+                      loc);
   CreateBr(failure_callback);
 
   SetInsertPoint(lookup_merge_block);
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -338,7 +338,17 @@ CallInst *IRBuilderBPF::CreateBpfPseudoCallValue(Map &map)
   return CreateBpfPseudoCallValue(mapid);
 }
 
-CallInst *IRBuilderBPF::createMapLookup(int mapid, Value *key)
+CallInst *IRBuilderBPF::createMapLookup(int mapid,
+                                        Value *key,
+                                        const std::string &name)
+{
+  return createMapLookup(mapid, key, getInt8PtrTy(), name);
+}
+
+CallInst *IRBuilderBPF::createMapLookup(int mapid,
+                                        Value *key,
+                                        PointerType *val_ptr_ty,
+                                        const std::string &name)
 {
   Value *map_ptr = CreateBpfPseudoCallId(mapid);
   // void *map_lookup_elem(struct bpf_map * map, void * key)
@@ -346,24 +356,59 @@ CallInst *IRBuilderBPF::createMapLookup(int mapid, Value *key)
 
   assert(key->getType()->isPointerTy());
   FunctionType *lookup_func_type = FunctionType::get(
-      getInt8PtrTy(), { map_ptr->getType(), key->getType() }, false);
+      val_ptr_ty, { map_ptr->getType(), key->getType() }, false);
   PointerType *lookup_func_ptr_type = PointerType::get(lookup_func_type, 0);
   Constant *lookup_func = ConstantExpr::getCast(
       Instruction::IntToPtr,
       getInt64(libbpf::BPF_FUNC_map_lookup_elem),
       lookup_func_ptr_type);
-  return createCall(
-      lookup_func_type, lookup_func, { map_ptr, key }, "lookup_elem");
+  return createCall(lookup_func_type, lookup_func, { map_ptr, key }, name);
 }
 
-CallInst *IRBuilderBPF::CreateGetJoinMap(Value *ctx, const location &loc)
+CallInst *IRBuilderBPF::CreateGetJoinMap(BasicBlock *failure_callback,
+                                         const location &loc)
 {
-  AllocaInst *key = CreateAllocaBPF(getInt32Ty(), "key");
-  CreateStore(getInt32(0), key);
+  return createGetScratchMap(bpftrace_.maps[MapManager::Type::Join].value()->id,
+                             "join",
+                             getInt8PtrTy(),
+                             loc,
+                             failure_callback);
+}
+
+// createGetScratchMap will jump to failure_callback if it cannot find the map
+// value
+CallInst *IRBuilderBPF::createGetScratchMap(int mapid,
+                                            const std::string &name,
+                                            PointerType *val_ptr_ty,
+                                            const location &loc,
+                                            BasicBlock *failure_callback,
+                                            int key)
+{
+  AllocaInst *keyAlloc = CreateAllocaBPF(getInt32Ty(),
+                                         nullptr,
+                                         "lookup_" + name + "_key");
+  CreateStore(getInt32(key), keyAlloc);
 
   CallInst *call = createMapLookup(
-      bpftrace_.maps[MapManager::Type::Join].value()->id, key);
-  CreateHelperErrorCond(ctx, call, libbpf::BPF_FUNC_map_lookup_elem, loc, true);
+      mapid, keyAlloc, val_ptr_ty, "lookup_" + name + "_map");
+  CreateLifetimeEnd(keyAlloc);
+
+  Function *parent = GetInsertBlock()->getParent();
+  BasicBlock *lookup_failure_block = BasicBlock::Create(
+      module_.getContext(), "lookup_" + name + "_failure", parent);
+  BasicBlock *lookup_merge_block = BasicBlock::Create(
+      module_.getContext(), "lookup_" + name + "_merge", parent);
+  Value *condition = CreateICmpNE(
+      CreateIntCast(call, getInt8PtrTy(), true),
+      ConstantExpr::getCast(Instruction::IntToPtr, getInt64(0), getInt8PtrTy()),
+      "lookup_" + name + "_cond");
+  CreateCondBr(condition, lookup_merge_block, lookup_failure_block);
+
+  SetInsertPoint(lookup_failure_block);
+  // TODO: should we return a loud error?
+  CreateBr(failure_callback);
+
+  SetInsertPoint(lookup_merge_block);
   return call;
 }
 
@@ -1550,6 +1595,24 @@ void IRBuilderBPF::CreatePerfEventOutput(Value *ctx,
                    { ctx, map_ptr, flags_val, data, size_val },
                    "perf_event_output",
                    loc);
+}
+
+void IRBuilderBPF::CreateDebugOutput(std::string fmt_str,
+                                     const std::vector<Value *> &values,
+                                     const location &loc)
+{
+  fmt_str = "[BPFTRACE_DEBUG_OUTPUT] " + fmt_str;
+  Constant *const_str = ConstantDataArray::getString(module_.getContext(),
+                                                     fmt_str,
+                                                     true);
+  AllocaInst *fmt = CreateAllocaBPF(
+      ArrayType::get(getInt8Ty(), fmt_str.length() + 1), "fmt_str");
+  CREATE_MEMSET(fmt, getInt8(0), fmt_str.length() + 1, 1);
+  CreateStore(const_str, fmt);
+  CreateTracePrintk(CreatePointerCast(fmt, getInt8Ty()->getPointerTo()),
+                    getInt32(fmt_str.length() + 1),
+                    values,
+                    loc);
 }
 
 void IRBuilderBPF::CreateTracePrintk(Value *fmt_ptr,
--- a/src/ast/irbuilderbpf.h
+++ b/src/ast/irbuilderbpf.h
@@ -191,6 +191,9 @@ public:
                         Value *key,
                         Value *val,
                         const location &loc);
+  void CreateDebugOutput(std::string fmt_str,
+                         const std::vector<Value *> &values,
+                         const location &loc);
   void CreateTracePrintk(Value *fmt,
                          Value *fmt_size,
                          const std::vector<Value *> &values,
--- a/src/ast/irbuilderbpf.h
+++ b/src/ast/irbuilderbpf.h
@@ -170,7 +170,7 @@ public:
   CallInst *CreateGetRandom(const location &loc);
   CallInst   *CreateGetStackId(Value *ctx, bool ustack, StackType stack_type, const location& loc);
   CallInst *CreateGetFuncIp(const location &loc);
-  CallInst   *CreateGetJoinMap(Value *ctx, const location& loc);
+  CallInst *CreateGetJoinMap(BasicBlock *failure_callback, const location &loc);
   CallInst *CreateHelperCall(libbpf::bpf_func_id func_id,
                              FunctionType *helper_type,
                              ArrayRef<Value *> args,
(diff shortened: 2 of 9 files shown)


src/bpftrace.h
```diff
@@ -282,6 +285,8 @@ class BPFtrace
   std::vector<uint8_t> find_empty_key(IMap &map, size_t size) const;
   bool has_iter_ = false;
   int epollfd_ = -1;
+  int debug_epollfd_ = -1;
+  char debug_output_buffer_[256 + 1];
```
danobi, line 289: Nit: maybe keep this as a local var in the function it's used? IIUC stack vars are basically free. But keeping it as a member makes class instance bigger

src/bpftrace.cpp
```diff
@@ -1338,6 +1344,34 @@ int BPFtrace::setup_output()
...
+int BPFtrace::set_debug_output()
+{
+  debug_epollfd_ = epoll_create1(EPOLL_CLOEXEC);
+  if (debug_epollfd_ == -1)
+  {
+    LOG(ERROR) << "Failed to create debug epollfd: " << strerror(errno);
+    return -1;
+  }
+
+  int trace_pipe_fd = open("/sys/kernel/debug/tracing/trace_pipe", O_RDONLY);
```
danobi, line 1356: Use `src/debugfs.cpp` helpers? At some point we should dynamically look at mount table to find the mntpts

src/bpftrace.cpp
```diff
@@ -1503,6 +1551,30 @@ int BPFtrace::poll_perf_events()
   return ready;
 }
 
+int BPFtrace::poll_debug_output()
+{
+  auto events = std::vector<struct epoll_event>(1);
+  int ready = epoll_wait(debug_epollfd_, events.data(), 1, timeout_ms);
```
danobi, line 1557: Not sure if `timeout_ms` should be used here. It unnecessarily delays the event loop right? Is it possible to do no timeout?

src/bpftrace.h
```diff
@@ -282,6 +285,8 @@ class BPFtrace
   std::vector<uint8_t> find_empty_key(IMap &map, size_t size) const;
   bool has_iter_ = false;
   int epollfd_ = -1;
+  int debug_epollfd_ = -1;
```
danobi, line 288: Need to `close()` in destructor?

danobi (commented):
sorry about the delay in reviewing

src/bpftrace.cpp
xh4n3: @danobi I originally planned to use a new thread to read event data independently, but this would complicate printing to standard output. So I switched to using the epoll, which avoids thread switching and may be less prone to problems.

I think the epoll_wait timeout is compulsory, otherwise it will block indefinitely until a debug event comes in.

src/ast/irbuilderbpf.cpp
```diff
@@ -338,32 +338,85 @@ CallInst *IRBuilderBPF::CreateBpfPseudoCallValue(Map &map)
...
+  return createGetScratchMap(bpftrace_.maps[MapManager::Type::Join].value()->id,
+                             "join",
+                             getInt8PtrTy(),
+                             loc,
+                             failure_callback);
+}
+
+// createGetScratchMap will jump to failure_callback if it cannot find the map
+// value
+CallInst *IRBuilderBPF::createGetScratchMap(int mapid,
```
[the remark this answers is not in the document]
xh4n3: @danobi If bpftrace cannot find the map value, it should always skip the processing logic of the join, IIUC, so there will always be a conditional statement.

Same as https://github.com/iovisor/bpftrace/blob/v0.18.1/src/ast/passes/codegen_llvm.cpp#L706-L712

src/bpftrace.cpp
danobi: According to the man page:

> Specifying a timeout of -1
> causes epoll_wait() to block indefinitely, while specifying a timeout equal to zero cause epoll_wait()
> to return immediately, even if no events are available.

So maybe 0 will work

src/ast/irbuilderbpf.cpp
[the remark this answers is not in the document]
danobi: Ah right. My bad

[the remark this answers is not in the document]
danobi: But still might be good to avoid printing to trace printk buffer unless debugging is enabled

src/bpftrace.cpp
xh4n3: Oops, how careless i am...

xh4n3:
@danobi Thanks for the nice suggestions! This PR is now ready for another review.

src/bpftrace.cpp
```diff
@@ -1338,6 +1345,34 @@ int BPFtrace::setup_output()
   return 0;
 }
 
+int BPFtrace::set_debug_output()
```
danobi, line 1348: nit: set_debug_output() -> setup_debug_output()

danobi (approved):
sorry about the delay! this lgtm with one small naming nit

xh4n3:
@danobi Thanks, I've update the commits.

danobi:
@ajor is interested in building on top of this work to migrate calls of `bpf_get_stackid()` to `bpf_get_stack()`. I'll wait 1 more day for comments before merging

viktormalik:
Putting this to my priority list, I'll do a review tomorrow morning.

src/ast/irbuilderbpf.cpp
```diff
@@ -338,32 +338,85 @@ CallInst *IRBuilderBPF::CreateBpfPseudoCallValue(Map &map)
...
 
   CallInst *call = createMapLookup(
-      bpftrace_.maps[MapManager::Type::Join].value()->id, key);
-  CreateHelperErrorCond(ctx, call, libbpf::BPF_FUNC_map_lookup_elem, loc, true);
+      mapid, keyAlloc, val_ptr_ty, "lookup_" + name + "_map");
+  CreateLifetimeEnd(keyAlloc);
+
+  Function *parent = GetInsertBlock()->getParent();
+  BasicBlock *lookup_success_block = BasicBlock::Create(
+      module_.getContext(), "lookup_" + name + "success", parent);
```
viktormalik, line 398: nit
```suggestion
      module_.getContext(), "lookup_" + name + "_success", parent);
```

src/ast/irbuilderbpf.cpp
```diff
@@ -338,32 +338,85 @@ CallInst *IRBuilderBPF::CreateBpfPseudoCallValue(Map &map)
...
+  BasicBlock *lookup_failure_block = BasicBlock::Create(
+      module_.getContext(), "lookup_" + name + "failure", parent);
```
viktormalik, line 400: nit
```suggestion
      module_.getContext(), "lookup_" + name + "_failure", parent);
```

src/ast/irbuilderbpf.cpp
```diff
@@ -1496,6 +1549,24 @@ void IRBuilderBPF::CreatePerfEventOutput(Value *ctx,
                    loc);
 }
 
+void IRBuilderBPF::CreateDebugOutput(std::string fmt_str,
```
viktormalik, line 1552: Why is this not used for `debugf`, too? IIUC, if you now want to dump `debugf` output to stderr, you always need to add `BPFTRACE_DEBUG_OUTPUT` substring, is that intentional? It seems so to me when I'm looking at the new runtime test.

viktormalik (commented):
Very nice!

Just one question and a couple of nits below.

src/ast/irbuilderbpf.cpp
xh4n3: > IIUC, if you now want to dump debugf output to stderr, you always need to add BPFTRACE_DEBUG_OUTPUT substring, is that intentional?

Yes, but I'm not quite sure about it.

The current setting is:
* the built-in `debugf` is for the convenience of our bpftrace development, never for end users. This is implemented just like `printf`, we can customize its output. The drawback is we have to look into the trace-pipe by ourselves.
* the BPFTRACE_DEBUG_OUTPUT is something like tracepoints that we intentionally left in the bpftrace codes, to provide diagnostic information under unexpected circumstances. These messages should be well-formatted, easily-understandable.

Anyway, I'm okay with using the `CreateDebugOutput` to achieve the built-in `debugf` if you say so.

src/ast/irbuilderbpf.h
```diff
@@ -185,6 +185,9 @@ class IRBuilderBPF : public IRBuilder<>
                     size_t size,
                     const location *loc = nullptr);
   void CreateAtomicIncCounter(int mapfd, uint32_t idx);
+  void CreateDebugOutput(std::string fmt_str,
```
ajor, line 188:
```suggestion
  void CreateDebugOutput(const std::string &fmt_str,
```
Or use `std::string_view`, but that would probably require a little more work on the caller's side

ajor (commented):
What's the reason for introducing a new pipe for BPF <=> userspace communication? It seems like a lot of extra code that we'll need to support

tests/codegen/call_join.cpp
```diff
@@ -0,0 +1,27 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_join)
+{
+  test("struct arg { char **argv } kprobe:f { @x = (struct arg *) 0; "
+       "join(@x->argv); }",
```
ajor, line 10:
```suggestion
  test("struct arg { char **argv } kprobe:f { $x = (struct arg *) 0; "
       "join($x->argv); }",
```
Nit: To keep the unit tests as specific as possible - this will mean the only maps in the LLVM IR are the scratch maps

xh4n3:
> What's the reason for introducing a new pipe for BPF <=> userspace communication?

@ajor Theoretically, there is a possibility that the GetScratchMap method may fail to retrieve the map value, although most of the time it could be due to bugs in bpftrace during development. I believe that in such cases we should throw a loud error instead of failing silently, as [it is currently](https://github.com/iovisor/bpftrace/blob/v0.19.1/src/ast/passes/codegen_llvm.cpp#L752-L813.

Since future PRs will utilize GetScratchMap for built-in functions like printf, the existing BPF <-> userspace communication will all relies the GetScratchMap. If we throw a loud error in the GetScratchMap, we should not in reverse depend on GetScratchMap itself. So I came up with this idea, using a side-channel (i.e. trace_pipe) for these internal error message, to address this circular dependency issue.

PR #2719 also provides some contexts on this.

ajor:
Two points:
1. Are there any cases when these errors could happen other than during bpftrace development? i.e. does it need to be enabled in release builds?
2. I see the first approach was to use the existing ring/perf buffer output - was this new approach taken with the idea that it's lighter weight? It feels more complicated to me.

The reason for my opposition to this new side-channel is that I'd like to keep the BPF <=> userspace interface as simple as possible, to make it easy to create other frontends to bpftrace.

xh4n3:
@ajor Thank you for the comments!

> Are there any cases when these errors could happen other than during bpftrace development? i.e. does it need to be enabled in release builds? 

As I understand it, there isn't one case currently (though I can never be sure about end-user's runtime environment). If we're going to utilize this pipe in more built-in, we may want to enable this in the release builds as a way to collect diagnostic infos.

> Could it just dump to this tracing pipe and have bpftrace developers check the output separately, rather than having it show in the terminal?

Yes, this is an option.

> Was this new approach taken with the idea that it's lighter weight? It feels more complicated to me.

The new approach is more complicated, but it provides a common utility could be useful for other built-ins in the future. To be honest, I'm quite happy if we use the implementation in the previous PR, since it doesn't introduce any user-facing changes, only introduce [a new error message MapError](https://github.com/iovisor/bpftrace/pull/2719/files#diff-6a5f2f25f035d434182f7b0074b3a1551b810a8a57bc190fcd26d6e2aa732039R107-R114). Alternatives are [a new field of the async error message](https://github.com/iovisor/bpftrace/pull/1801/commits/53d5728f17a42ed332d71642672eff797565a392#diff-17a44cdf8c7c18875c6ed057c572cec4db6175d9d7348ed1e7f475ef8cc34fd8R111-R127), or [just a plain HelperError](https://github.com/iovisor/bpftrace/blob/v0.19.1/src/ast/irbuilderbpf.cpp#L419).

danobi:
@ajor My original thought/suggestion was that we have the prog print to trace pipe the debug logs. And then developer can access it from there. I figured that was a simple enough mechanism.

But b/c the userspace polling code is completely additive, I am thinking we have the flexibility to delete it later if we want. Or we can delete it now if you think it's not worth it.

I'd prefer if we merge this PR cuz it'll unblock a lot of important fixes/features that need to be addressed in bpftrace (like big string support). And we can sort it out in follow up PRs.

ajor:
Ok on reading the code more carefully I see that this new debug output is only activated when the env var is set, so it won't be affecting production traces (as I'd feared). Also since it's opt-in, I think it'd be fine to expect bpftrace developers to check the trace_pipe manually when this logging is enabled.

I'd suggest that we keep the trace pipe logging as-in, but delete the bpftrace.cpp code that dumps it to stderr. It is just additive, but every extra bit of code adds complexity and more maintenance burden that I don't really want to take on for rarely-used debug output.

I'm also considering how to make it possible to embed bpftrace scripts in other applications, which would require a stable BPF <=> userspace interface. Although this debug output doesn't have to be part of that, it'd be nice if bpftrace.cpp could be a reference example of how to do this integration, without extra debug code adding confusion. I can talk more about this in the next office hours (once I have more concrete ideas myself!)

xh4n3:
I agree that we should keep it simple. I've updated the commits. Could you please taking another look? @ajor @danobi @viktormalik

src/ast/irbuilderbpf.cpp
viktormalik: Right, makes sense. Let's keep it as it is now.

src/ast/irbuilderbpf.cpp
```diff
@@ -338,32 +338,85 @@ CallInst *IRBuilderBPF::CreateBpfPseudoCallValue(Map &map)
...
+  BasicBlock *lookup_merge_block = BasicBlock::Create(
+      module_.getContext(), "lookup_" + name + "_merge", parent);
+  Value *condition = CreateICmpNE(
+      CreateIntCast(call, getInt8PtrTy(), true),
+      ConstantExpr::getCast(Instruction::IntToPtr, getInt64(0), getInt8PtrTy()),
+      "lookup_" + name + "_cond");
+  CreateCondBr(condition, lookup_success_block, lookup_failure_block);
+
+  SetInsertPoint(lookup_success_block);
+  CreateBr(lookup_merge_block);
```
viktormalik, line 410: Are there plans to add something to the `lookup_success_block`? Any reason why not to jump directly to `lookup_merge_block`? Or are we relying on LLVM to optimize this away?

viktormalik:
Overall looks good to me.

Agreed with @ajor that it's probably better to just dump to the trace pipe and leave out the code for polling the trace pipe and printing to stderr, for simpler maintenance.

src/ast/irbuilderbpf.cpp
xh4n3: Nice, let me clean up this branch.

ajor (approved):
Looks good!

Issue #2753, opened 2023-09-06
Improving the documentations


xh4n3:
Consensus reached from bpftrace meeting on Sep 5th, 2023
> * Let’s keep everything in one place: manual page
> * It’s generated from asciidoc which is rendered nicely on GitHub website
> * Reference guide should be either deleted or shortened (sort of a “quick start guide”, inspired by 6-page awk manual starting on page 106 of [UNIX programmer's manual](https://9p.io/7thEdMan/v7vol2b.pdf))

There will be several stages marching towards the desired final state, which includes:
1. Ensuring that the manual page covers everything that the reference guide currently has.
2. Creating a new quick start guide, either from scratch or by modifying the existing reference guide.
3. Developing a website to host the manual pages.

I'm creating this issue to track progresses. I will keep this updated.

Please leave comments if any ideas, thanks!

jordalgo:
@xh4n3 Happy New Year and gentle ping :)

xh4n3:
@jordalgo Thanks for the reminder :)

I'm sorry that due to a recent change in my job position, I am unable to continue contributing to this project 😞. I've truly learned a lot from this project and from collaborating with all of you, especially @viktormalik @danobi . Thank you so much ❤️ 

This particular improvement task might still be valuable, so I'll leave it open. Please feel free to take https://github.com/iovisor/bpftrace/pull/2754 branch and contiune.

danobi:
Thanks for you contributions @xh4n3 ! You're welcome back any time :)

jordalgo:
I'm sure there are still things we can do to improve documentation but we've done an initial pass and merged the adoc with the reference doc.

PR #2754, opened 2023-09-06
[Draft] Moving contents to manual pages from reference guide

xh4n3:
I'm creating this PR to demonstrate the concept of stage 1 mentioned in #2753 .

I've deleted some sections from the reference guide, and moved them into the manual page. There is a live manual page preview at [my working branch](https://github.com/xh4n3/bpftrace/blob/update-man-page/man/adoc/bpftrace.adoc).

Please leave your valuable feedback on this change, thanks! If there is no objections, I'll go moving the remaining sections shortly.

**Note: The deletions are meant to show which parts are being moved. If we all agree on this change, I'll revert the deletions before merging this.**


danobi:
If I understand correctly, the plan is to complete the manual, then start deleting from the reference guide (or replace the reference guide)?

docs/reference_guide.md
```diff
@@ -9,137 +9,6 @@ Also, please keep these docs as terse as possible to maintain their brevity (ins
...
-    - [9. `lhist()`: Linear Histogram](#9-lhist-linear-histogram)
-    - [10. `print()`: Print Map](#10-print-print-map)
-- [Output](#output)
-    - [1. `printf()`: Per-Event Output](#1-printf-per-event-output)
-    - [2. `interval`: Interval Output](#2-interval-interval-output)
-    - [3. `hist()`, `printf()`: Histogram Printing](#3-hist-print-histogram-printing)
-    - [4. `SIGUSR1`: On-Demand Output](#4-sigusr1-on-demand-output)
-- [BTF Support](#btf-support)
-- [Advanced Tools](#advanced-tools)
-- [Errors](#errors)
```
danobi, line 125: I'm a bit confused about the deleted entries here. It doesn't look like it was moved into the man page. Is that intentional?

xh4n3: Yes, this is intentional.

I'm moving contents those should be included in the man page, and deleting those should not. So when the reference guide become empty, the man page is complete.

xh4n3:
> If I understand correctly, the plan is to complete the manual, then start deleting from the reference guide (or replace the reference guide)?

@danobi Yes, the deletions from this PR is just for easy review, will be reverted before merge.

After merging this, the next stage is to create a new quick start guide and delete the whole reference page.

docs/reference_guide.md
danobi: Ah ok, I see. Your intention is to "delete" the table of contents in this PR. But the actual contents will be moved in subsequent PRs.

danobi:
I gave the man pages changes a skim -- they lgtm

jordalgo:
I do find the markdown a bit easier to navigate on github but maybe it's worth deleting it so there is a single source of truth for using bpftrace.

man/adoc/bpftrace.adoc
```diff
@@ -53,104 +105,225 @@ Custom prefix to easily link section
 
 === Output format
 
-*-B* _MODE_::
+==== *-B* _MODE_
   Set the buffer mode for stdout. Valid values are::
```
viktormalik, line 109: Remove indentation as in the rest of the section?

man/adoc/bpftrace.adoc
```diff
@@ -53,104 +105,225 @@ Custom prefix to easily link section
...
 
     *none* No buffering. Each I/O is written as soon as possible +
     *line* Data is written on the first newline or when the buffer is full.
     This is the default mode. +
     *full* Data is written once the buffer is full.
 
-*-f* _FORMAT_::
+==== *-f* _FORMAT_
   Set the output format. Valid values are::
```
viktormalik, line 117: Same as above.

viktormalik (commented):
Looks good to me, just two nits.

I think that you can go on with converting all the parts missing from the manual.

PR #2763, opened 2023-09-14
Various fixes for CI

viktormalik:
- GitHub Actions runners seem to have updated kernel and the old version of `biosnoop.bt` is no longer necessary. 
Unfortunately, it looks like this is the case for some runners only, so we cannot universally use neither the old nor the new variant. Let's disable the test for now and revisit later when all runners are updated.

- We're also observing CI failure in `memleak-test.sh`. It's not related to the memleak test itself, the bug is in `ClangParser` which one of the memleak tests executes:

        # cat script.bt
        #include <linux/skbuff.h>
        BEGIN {}
    
        # bpftrace script.bt
        [...]
        /lib/modules/[...]/ibt.h:71:10: error: use of undeclared identifier 'true'
        [...]
        /lib/modules/[...]/le.h:36:2: error: use of undeclared identifier 'uintptr_t'
    
  This is fixed by improving our parsing of undefined typedefs from Clang diagnostic messages to be able to pull those typedefs from BTF.

##### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,54 +39,64 @@ jobs:
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm10
           RUNTIME_TEST_DISABLE: probe.kprobe_offset_fail_size,usdt.usdt probes - file based semaphore activation multi process
-          TOOLS_TEST_OLDVERSION: biosnoop.bt tcpdrop.bt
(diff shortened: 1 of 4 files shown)


danobi (approved):

viktormalik:
There's one more new CI failure (probably also caused by the kernel update), I'll add it to this PR,  soon.

viktormalik:
@danobi I added one more CI fix, could you please re-check? Thanks!

danobi (approved):
lgtm!

PR #2765, opened 2023-09-17
Fix USDT run_time test targeting all pids

jordalgo:
This creates a new pid namespace (and mounts proc) to run bpftrace inside of this namespace.

##### Checklist

### DIFF
--- a/tests/runtime/engine/runner.py
+++ b/tests/runtime/engine/runner.py
@@ -12,7 +12,7 @@ from functools import lru_cache
 import cmake_vars
 
 BPF_PATH = os.environ["BPFTRACE_RUNTIME_TEST_EXECUTABLE"]
-ATTACH_TIMEOUT = 5
+ATTACH_TIMEOUT = 10
 DEFAULT_TIMEOUT = 5
 
 
@@ -83,22 +83,24 @@ class Runner(object):
             raise ValueError("Invalid skip reason: %d" % status)
 
     @staticmethod
-    def prepare_bpf_call(test):
-        bpftrace_path = "{}/bpftrace".format(BPF_PATH)
-        bpftrace_aotrt_path = "{}/aot/bpftrace-aotrt".format(BPF_PATH)
+    def prepare_bpf_call(test, nsenter=[]):
+        bpftrace_path = os.path.abspath(f"{BPF_PATH}/bpftrace")
+        bpftrace_aotrt_path = os.path.abspath(f"{BPF_PATH}/aot/bpftrace-aotrt")
+
+        nsenter_prefix = (" ".join(nsenter) + " ") if len(nsenter) > 0 else ""
 
         if test.run:
             ret = re.sub("{{BPFTRACE}}", bpftrace_path, test.run)
             ret = re.sub("{{BPFTRACE_AOTRT}}", bpftrace_aotrt_path, ret)
 
-            return ret
+            return nsenter_prefix + ret
         else:  # PROG
             # We're only reusing PROG-directive tests for AOT tests
             if test.suite == 'aot':
-                return "{} -e '{}' --aot /tmp/tmpprog.btaot && {} /tmp/tmpprog.btaot".format(
+                return nsenter_prefix + "{} -e '{}' --aot /tmp/tmpprog.btaot && {} /tmp/tmpprog.btaot".format(
                     bpftrace_path, test.prog, bpftrace_aotrt_path)
             else:
-                return "{} -e '{}'".format(bpftrace_path, test.prog)
+                return nsenter_prefix + "{} -e '{}'".format(bpftrace_path, test.prog)
 
     @staticmethod
     def __handler(signum, frame):
@@ -108,7 +110,7 @@ class Runner(object):
     @lru_cache(maxsize=1)
     def __get_bpffeature():
         p = subprocess.Popen(
-            [f"{BPF_PATH}/bpftrace", "--info"],
+            [os.path.abspath(f"{BPF_PATH}/bpftrace"), "--info"],
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
             start_new_session=True,
@@ -137,6 +139,29 @@ class Runner(object):
         bpffeature["jiffies64"] = output.find("jiffies64: yes") != -1
         return bpffeature
 
+    
+    @staticmethod
+    def __wait_for_children(parent_pid, timeout, ps_format, condition):
+        with open(os.devnull, 'w') as dn:
+            waited=0
+            while waited <= timeout:
+                run_cmd = ["ps", "--ppid", str(parent_pid), "--no-headers", "-o", ps_format]
+                children = subprocess.run(run_cmd,
+                                          check=False,
+                                          stdout=subprocess.PIPE,
+                                          stderr=subprocess.PIPE)
+                if children.returncode == 0 and children.stdout:
+                    lines = [line.decode('utf-8') for line in children.stdout.splitlines()]
+                    if (condition(lines)):
+                        return lines
+                else:
+                    print(f"__wait_for_children error: {children.stderr}. Return code: {children.returncode}")
+
+                time.sleep(0.1)
+                waited+=0.1
+        return None
+
+
     @staticmethod
     def run_test(test):
         current_kernel = LooseVersion(os.uname()[2])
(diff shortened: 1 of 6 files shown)


danobi (commented):
The issue in the regular (non-embedded) CI is that it looks like `unshare` is not present in the nix flake config. Should be fairly easy to add it there: https://github.com/iovisor/bpftrace/blob/12c463e8ff71bb1423243cbc3d23a7c4d59d5b8b/flake.nix#L110

The other issue in the embedded CI kinda looks like a namespacing issue (guessing that's what the `[unknown]` bit is). I'm suspecting it'd be more reliable to add pid namespace awareness to the runtime test harness itself rather than having tests shell out to `unshare`. Cuz it requires a bit of care to run the BEFORE progs in the same pidns as bpftrace. And it looks like the infra already has to have some awareness.

Could use a marker directive like existing `WILL_FAIL`. That way other tests could easily reuse this. I see a lot of value for the other existing ones that walk procfs.

jordalgo:
@danobi Let me know what you think of this first draft

danobi (commented):
Should `AFTER` scripts run in the pid ns too? Might be ok to omit until actually necessary.

But docs should be updated with `NEW_PIDNS` as well as whatever we land on for `AFTER` behavior

tests/runtime/engine/runner.py
```diff
@@ -196,10 +198,53 @@ def run_test(test):
...
+                    while waited <= test.timeout:
+                        get_unshare_children = ["ps", "--ppid", str(unshare_out.pid), "--no-headers", "-o", "pid"]
+                        unshare_children = subprocess.run(get_unshare_children,
+                                                          check=False,
+                                                          stdout=subprocess.PIPE,
+                                                          stderr=subprocess.PIPE)
+
+                        if unshare_children.returncode == 0 and unshare_children.stdout:
+                            lines = [line.decode('utf-8') for line in unshare_children.stdout.splitlines()]
+                            lines = sorted((line.strip() for line in lines))
```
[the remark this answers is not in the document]
jordalgo: Good call. Although at this point there really should only be one child.

jordalgo:
> Should AFTER scripts run in the pid ns too? Might be ok to omit until actually necessary.

That's a good call. I think they should. I'll make that update.

tests/runtime/engine/runner.py
```diff
@@ -136,6 +138,31 @@ def __get_bpffeature():
...
+                children = subprocess.run(run_cmd,
+                                          check=False,
+                                          stdout=subprocess.PIPE,
+                                          stderr=subprocess.PIPE)
+                if children.returncode == 0 and children.stdout:
+                    lines = [line.decode('utf-8') for line in children.stdout.splitlines()]
+                    if (condition(lines)):
+                        return True
+                else:
+                    print(f"__wait_for_children error: {children.stderr}")
```
danobi, line 157:
```suggestion
                    print(fail(f"__wait_for_children error: {children.stderr}"))
```

tests/runtime/engine/runner.py
```diff
@@ -196,42 +232,64 @@ def run_test(test):
...
+                    # Use the first BEFORE as the first process in the new pid namespace
+                    unshare_out = subprocess.Popen(["unshare", "--fork", "--pid", "--mount-proc", "-r", "--kill-child"] + test.befores[0].split(),
+                                                   start_new_session=True, universal_newlines=True,
+                                                   stdout=subprocess.PIPE,
+                                                   stderr=subprocess.STDOUT)
+
+                    def has_children(lines):
+                        if len(lines) == 0:
+                            return False
+                        nsenter.extend(["nsenter", "-p", "-m", "-t", lines[0].strip()])
```
danobi, line 250: A bit confusing that the "conditional" callback mutates state in the caller

danobi:
Ah didn't notice PR was in draft. Feel free to ignore comments while you debug

tests/runtime/engine/runner.py
jordalgo: I know. I didn't love this either. Maybe I could have this function return the lines if the conditional is true.

jordalgo:
> Ah didn't notice PR was in draft. Feel free to ignore comments while you debug

No, please continue. These comments are helpful. I also still can't figure out why the Embedded builds are failing.

tests/runtime/engine/runner.py
jordalgo: This is just for debugging. I don't think we want it to have the fail colors.

jordalgo:
@danobi This is ready for a proper review

tests/runtime/engine/runner.py
```diff
@@ -162,7 +187,16 @@ def run_test(test):
         after = None
         after_output = None
         cleanup = None
+        # This is only populated if the NEW_PIDNS directivce is set
```
danobi, line 190:
```suggestion
        # This is only populated if the NEW_PIDNS directive is set
```

tests/runtime/engine/runner.py
```diff
@@ -162,7 +187,16 @@ def run_test(test):
...
+        # and is used to enter the newly created pid namespace for all BEFOREs,
+        # the prirmary RUN or PROG command, and the AFTER.
```
danobi, line 192:
```suggestion
        # the primary RUN or PROG command, and the AFTER.
```

tests/runtime/engine/runner.py
```diff
@@ -162,7 +187,16 @@ def run_test(test):
...
+        nsenter = []
         bpf_call = "[unknown]"
+
+        def get_pid_ns_cmd(cmd):
+            cwd = "." + os.getcwd() + "/"
+            return nsenter + [x.replace("./", cwd) for x in cmd.split()]
```
danobi, line 198: Use abspath() like above? Also handles relative paths
```suggestion
            return nsenter + [os.path.abspath(x) for x in cmd.split()]
```

tests/runtime/engine/runner.py
```diff
@@ -137,6 +139,29 @@ def __get_bpffeature():
         bpffeature["jiffies64"] = output.find("jiffies64: yes") != -1
         return bpffeature
 
+    
+    @staticmethod
+    def __wait_for_children(parent_pid, timeout, ps_format, condition):
+        with open(os.devnull, 'w') as dn:
```
danobi, line 145: Unused?

tests/runtime/engine/runner.py
```diff
@@ -196,42 +232,64 @@ def run_test(test):
...
+                        return True
+
+                    if not Runner.__wait_for_children(unshare_out.pid, test.timeout, "pid", has_children):
+                        raise TimeoutError(f'Timed out waiting create a new PID namespace')
+
+                    # This is the only one we need to add to befores as killing the
+                    # unshare process will kill the whole process subtree with "--kill-child"
+                    befores.append(unshare_out)
+                    for before in test.befores[1:]:
+                        child_name = os.path.basename(before.strip().split()[-1])[:15]
```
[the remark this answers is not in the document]
danobi: Ah I see. It's assuming there are no arguments passed to the BEFORE program. For example with this diff things break:

```diff
diff --git a/tests/runtime/usdt b/tests/runtime/usdt
index 72f1da3c..5eb1b03f 100644
--- a/tests/runtime/usdt
+++ b/tests/runtime/usdt
@@ -90,7 +90,7 @@ NEW_PIDNS

 NAME usdt probes - attach to fully specified library probe by pid
 RUN {{BPFTRACE}} -e 'usdt:./testlibs/libusdt_tp.so:tracetestlib:lib_probe_1 { printf("here\n" ); exit(); }' -p $(pidof usdt_lib)
-BEFORE ./testprogs/usdt_lib
+BEFORE ./testprogs/usdt_lib la
 EXPECT here
 TIMEOUT 5
 REQUIRES ./testprogs/usdt_lib should_not_skip
```

jordalgo:
> Ah I see. It's assuming there are no arguments passed to the BEFORE program. For example with this diff things break:

Good catch. Will fix.

tests/runtime/engine/runner.py
jordalgo: I stole this from the original code. TBH, I'm not sure what it's doing but I was afraid to just remove it.

danobi (approved):

PR #2769, opened 2023-09-19
Add jiffies builtin with bpf_jiffies64

heterog:
Add jiffies builtin, one of the use case is to calculate the `sched_clock` indirectly within some old kernels.

Result:

```bash
bpftrace --unsafe -e 'BEGIN { system("grep jiffies: /proc/timer_list | head -1"); print(jiffies); exit() }'
Attaching 1 probe...
jiffies: 4325204698
4325204697
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to
 ## Unreleased
 
 #### Added
+- Add a `jiffies` builtin for advanced usages
+  - [#2769](https://github.com/iovisor/bpftrace/pull/2769)
 #### Changed
 #### Deprecated
 #### Removed
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1045,6 +1045,12 @@ For string arguments use the `str()` call to retrieve the value.
 | ktime_get_ns / ktime_get_boot_ns
 | nanoseconds since kernel boot. On kernels that support `ktime_get_boot_ns` this includes the time spent suspended, on older kernels it does not.
 
+| jiffies
+| uint64
+| 5.9
+| get_jiffies_64
+| Jiffies of the kernel. In 32-bit system, using this builtin might be slower.
+
 | pid
 | uint64
 | 4.2
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -999,6 +999,15 @@ CallInst *IRBuilderBPF::CreateGetNs(TimestampMode ts, const location &loc)
   return CreateHelperCall(fn, gettime_func_type, {}, "get_ns", &loc);
 }
 
+CallInst *IRBuilderBPF::CreateJiffies64(const location &loc)
+{
+  // u64 bpf_jiffies64()
+  // Return: jiffies (BITS_PER_LONG == 64) or jiffies_64 (otherwise)
+  FunctionType *jiffies64_func_type = FunctionType::get(getInt64Ty(), false);
+  return CreateHelperCall(
+      libbpf::BPF_FUNC_jiffies64, jiffies64_func_type, {}, "jiffies64", &loc);
+}
+
 Value *IRBuilderBPF::CreateIntegerArrayCmpUnrolled(Value *ctx,
                                                    Value *val1,
                                                    Value *val2,
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -276,3 +276,10 @@ EXPECT @[{.n=0x[0-9a-f]+,.c=120}]: 1
 REQUIRES_FEATURE dwarf
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
+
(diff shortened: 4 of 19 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -1045,6 +1045,12 @@ For string arguments use the `str()` call to retrieve the value.
 | ktime_get_ns / ktime_get_boot_ns
 | nanoseconds since kernel boot. On kernels that support `ktime_get_boot_ns` this includes the time spent suspended, on older kernels it does not.
 
+| jiffies
+| uint64
+| 5.6
```
viktormalik, line 1050: Technically speaking, it's only possible to use this from tracing programs from 5.9 (https://github.com/torvalds/linux/commit/72e2b2b66f9c).

src/ast/irbuilderbpf.cpp
```diff
@@ -999,6 +999,16 @@ CallInst *IRBuilderBPF::CreateGetNs(TimestampMode ts, const location &loc)
   return CreateHelperCall(fn, gettime_func_type, {}, "get_ns", &loc);
 }
 
+CallInst *IRBuilderBPF::CreateJiffies64(const location &loc)
+{
+  FunctionType *jiffies64_func_type = FunctionType::get(getInt64Ty(), false);
```
viktormalik, line 1004: Add comment with the helper signature (see other `Create...` functions in `irbuilder.cpp`).

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to
   - [#2686](https://github.com/iovisor/bpftrace/pull/2686)
 - Support targeting all running processes for USDTs
   - [#2734](https://github.com/iovisor/bpftrace/pull/2734)
+- Add a `jiffies` builtin for advanced usages
+  - [#2769](https://github.com/iovisor/bpftrace/pull/2769)
```
viktormalik, line 25: I just did a new release of bpftrace, please move this to the "Unreleased" section of the changelog.

viktormalik (changes requested):
Thanks a lot for implementing this!

Looks good, a couple of things are missing, though:
- detection if the helper is present in the kernel, it should go to `BPFFeature` and then be called from semantic analyser with a proper error message if the helper is missing,
- semantic analyser tests,
- runtime tests, although I'm not sure if we can come up with something useful here.

See e.g. my recent PR #2692 for inspiration how to implement the above points.

man/adoc/bpftrace.adoc
heterog: Aha yes, I just looked at the https://github.com/torvalds/linux/commit/5576b991e9c1a11d2cc21c4b94fc75ec27603896 and thought it was introduced in 5.6, thanks for pointing out ;)

heterog:
> Thanks a lot for implementing this!
> 
> Looks good, a couple of things are missing, though:
> 
>     * detection if the helper is present in the kernel, it should go to `BPFFeature` and then be called from semantic analyser with a proper error message if the helper is missing,
> 
>     * semantic analyser tests,
> 
>     * runtime tests, although I'm not sure if we can come up with something useful here.
> 
> 
> See e.g. my recent PR #2692 for inspiration how to implement the above points.

I'm quite unfamiliar here, thanks for the advising! Pushed a new commit, fixed those things.

As for the runtime tests, I want to compare the jiffies between bpftrace and `/proc/timer_list`, but neither the test engine nor the bpftrace can achieve that easily (the `system()` in bpftrace cannot be extracted, put compare outside will cause a huge jiffies delay between them), so yes, nothing useful idea has came up :face_in_clouds:.

viktormalik:
Please fix the problems reported by clang-format.

viktormalik (approved):

viktormalik:
Thanks!

danobi (approved):
lgtm, thanks!

tests/runtime/builtin
```diff
@@ -276,3 +276,10 @@ EXPECT @[{.n=0x[0-9a-f]+,.c=120}]: 1
 REQUIRES_FEATURE dwarf
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
+
+NAME jiffies
+PROG i:ms:1 { printf("SUCCESS %d\n", jiffies); exit(); }
+EXPECT SUCCESS -?[0-9][0-9]*
```
danobi, line 282: Is the `-?` bit of the regex cuz you're trying to print a u64 as an signed 32 bit int? Shouldn't the format specifier be `%llu` and the you can drop the `-?`?

heterog: I think yes, other builtins like `nsecs` also has the same problem, shall fix them all later and testing 😶‍🌫️.

danobi:
Thanks for the cleanup!

viktormalik:
Could you squash the first two commits together? Thanks!

heterog:
> Could you squash the first two commits together? Thanks!

Done!

PR #2790, opened 2023-10-06
tools/runqlen.bt: remove 'runnable_weight' field

AboorvaDevarajan:
In this commit, the 'runnable_weight' field has been removed from the 'cfs_rq' struct in runqlen.bt to align with the current  version of Linux, which no longer includes this field.

Reference to the Linux kernel commit (introduced from 5.7-rc1):

https://github.com/torvalds/linux/commit/0dacee1bfa70e171be3a12a30414c228453048d2

-------------------------

This script always return 0 as runqlen even when the it is actually non-zero. 

Force run 4 processes affinitize'd to a single CPU (say, CPU 20) thereby making the runqlen of CPU 20 - 3-4.

##### Workload:

```
# taskset -c 20 stress-ng --cpu 4
stress-ng: info:  [68396] defaulting to a 1 day, 0 secs run per stressor
stress-ng: info:  [68396] dispatching hogs: 4 cpu
..
Running
```

##### Before Patch: (value is always 0)

```
# ./tools/runqlen.bt 
Attaching 2 probes...
Sampling run queue length at 99 Hertz... Hit Ctrl-C to end.
^C

@runqlen: 
[0, 1)             60129 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
```

##### After Patch:

```
 ./tools/runqlen.bt 
Attaching 2 probes...
Sampling run queue length at 99 Hertz... Hit Ctrl-C to end.
^C

@runqlen: 
[0, 1)             57536 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[1, 2)                 2 |                                                    |
[2, 3)                 0 |                                                    |
[3, 4)               454 |                                                    |
```

##### Checklist

### DIFF
--- a/tools/runqlen.bt
+++ b/tools/runqlen.bt
@@ -19,7 +19,6 @@
 // your kernel version. It is from kernel/sched/sched.h:
 struct cfs_rq {
 	struct load_weight load;
-	unsigned long runnable_weight;
 	unsigned int nr_running;
 	unsigned int h_nr_running;
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to
 #### Fixed
 #### Docs
 #### Tools
+- Update runqlen.bt to remove `runnable_weight` field from cfs_rq struct.
+  - [#2790](https://github.com/iovisor/bpftrace/pull/2790)
 
 ## [0.19.0] 2023-09-19
 


tools/runqlen.bt
```diff
@@ -5,6 +5,8 @@
  *
  * This is a bpftrace version of the bcc tool of the same name.
  *
+ * For Linux 5.7+ (see tools/old for script for lower versions).
```
viktormalik, line 8: Technically, this is not correct. If BTF is available, the correct definition of `struct cfs_rq` will be pulled.

That being said, I'm ok with this change, I just wouldn't mention that the script is for kernel 5.7+ only. Also I'd avoid keeping the old version of the tool as the comment below explicitly states that the struct def may require user's intervention (on systems without BTF).

viktormalik (changes requested):

tools/runqlen.bt
AboorvaDevarajan: @viktormalik Thanks, I have made the requested changes, this will be just to ensure that the script remains compatible with the most recent kernel version.

viktormalik: Thanks! 

IMHO, any recent kernel version *should* have BTF enabled but I understand that may not always be the case.

viktormalik (approved):

PR #2791, opened 2023-10-06
Refactor work with BPF bytecode

viktormalik:
The current conversion of "compiled bytecode" to "arguments to `bpf_prog_load`" is pretty straight-forward - just a lookup for the appropriate section in the BpfBytecode map.

Things are going to be more complicated as we (1) introduce new features such as subprograms and (2) move towards libbpf-style of loading programs. So, this creates a new class `BpfProgram` to encapsulate the relationship between the bytecode and what gets sent to `bpf_prog_load`. Maps relocation code is moved from `Relocator` to this new class.

This is a part of #2710 posted separately to simplify reviewing.

Should contain no functional change.

##### Checklist


danobi (approved):
nice cleanup. thanks for splitting this out

src/attached_probe.h
```diff
@@ -81,7 +82,7 @@ class AttachedProbe
   void cache_progfd(void);
 
   Probe &probe_;
-  std::tuple<uint8_t *, uintptr_t> func_;
+  BpfProgram &prog_;
```
ajor, line 85: Please correct me if I'm wrong, but this looks to me like a dangling reference.

BpfProgram is created on the stack in `BPFtrace::attach_probe` from the return value of `BpfProgram::CreateFromBytecode`. Then a reference to this stack value is stored here in AttachedProbe.

src/bpfprogram.h
```diff
@@ -0,0 +1,46 @@
...
+using BpfBytecode = std::unordered_map<std::string, std::vector<uint8_t>>;
+
+class BPFtrace;
+
+class BpfProgram
+{
+public:
+  static std::optional<BpfProgram> CreateFromBytecode(
+      const BpfBytecode &bytecode,
+      const char *name,
```
ajor, line 23:
```suggestion
      std::string_view name,
```
Just a nit and I know it wasn't you, but this is C++!

ajor (changes requested):

ajor:
Thanks a lot for splitting this up by the way, it's so much easier to review

src/bpfprogram.cpp
```diff
@@ -0,0 +1,77 @@
...
+BpfProgram::BpfProgram(const BpfBytecode &bytecode,
+                       const char *name,
+                       MapManager &maps)
+    : bytecode_(bytecode), maps_(maps), name_(name), code_()
+{
+}
+
+const std::vector<uint8_t> &BpfProgram::getCode()
+{
+  if (code_.empty())
```
ajor, line 30: Not wrong specifically, but the flow of the code through `getCode` and `assemble` is a little odd and seems like a recipe for bugs
- why do we check `code_.empty()` so many times?
- right now we never call getCode() without first calling assembel(), which seems reasonable. This invariant isn't defined anywhere and this block of code I'm commenting on would hide violations of this
- we're not catching the exceptions in this code path

As it seems to work and in the interests of moving forward with BPF subprograms, it probably doesn't have to be changed right away though

src/attached_probe.h
viktormalik: You're right, good catch. Let's just store `BpfProgram` directly inside `AttachedProbe` as it's where it belongs.

src/bpfprogram.cpp
viktormalik: This is the original design that I kept but I agree here that we should be always explicitly calling `assemble` before we work with the BPF program, otherwise it's a bug. So let me remove this check.

src/bpfprogram.h
viktormalik: In the end, I used `const std::string &` as the values passed are strings anyways.

src/bpfprogram.cpp
```diff
@@ -0,0 +1,73 @@
...
+  code_.reserve(section.size());
+  code_.assign(section.data(), section.data() + section.size());
+
+  relocateMaps();
+}
+
+void BpfProgram::relocateMaps()
+{
+  struct bpf_insn *insns = reinterpret_cast<struct bpf_insn *>(code_.data());
+  for (uintptr_t i = 0; i < code_.size() / sizeof(struct bpf_insn); ++i)
```
jordalgo, line 48: nit: you could do the division outside the loop to prevent re-calculation (the compiler might optimize this out anyway)

src/bpfprogram.h
```diff
@@ -0,0 +1,46 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <tuple>
+#include <unordered_map>
+#include <vector>
+
+namespace bpftrace {
+
+using BpfBytecode = std::unordered_map<std::string, std::vector<uint8_t>>;
```
jordalgo, line 14: naming nit: BpfBytecodeMap?

src/bpftrace.cpp
```diff
@@ -965,18 +961,14 @@ std::vector<std::unique_ptr<AttachedProbe>> BPFtrace::attach_probe(
...
-  std::get<0>(*section) = relocated.data();
-  auto relocator = Relocator(*section, *this);
-  if (relocator.relocate())
-  {
-    LOG(ERROR) << "Failed to relocate insns for probe: " << probe.name;
+  try
+  {
+    program->assemble();
+  }
+  catch (std::runtime_error &ex)
```
jordalgo, line 968: nit: `const std::runtime_error`

src/bpfprogram.h
viktormalik: IMHO no, it's not a map of "multiple BPF bytecodes", the entire map represents a "single BPF bytecode" as it's a map of its sections to the section contents.

jordalgo (approved):

src/bpfprogram.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+void BpfProgram::assemble()
+{
+  if (!code_.empty())
+    return;
+
+  auto &section = bytecode_.at(name_);
+  code_.reserve(section.size());
+  code_.assign(section.data(), section.data() + section.size());
```
ajor, line 40:
```suggestion
  auto handle = bytecode_.extract(name_);
  code_ = std::move(handle.mapped());
```
To avoid copying the bytecode unnecessarily

src/bpftrace.cpp
```diff
@@ -839,7 +839,7 @@ void perf_event_lost(void *cb_cookie, uint64_t lost)
 
 std::vector<std::unique_ptr<AttachedProbe>> BPFtrace::attach_usdt_probe(
     Probe &probe,
-    std::tuple<uint8_t *, uintptr_t> func,
+    BpfProgram &program,
```
ajor, line 842:
```suggestion
    BpfProgram &&program,
```
`BpfProgram` (no reference) would work as well, but this expresses the intention better and ensures that the callers std::move into it

src/attached_probe.cpp
```diff
@@ -234,11 +234,11 @@ AttachedProbe::AttachedProbe(Probe &probe,
 }
 
 AttachedProbe::AttachedProbe(Probe &probe,
-                             std::tuple<uint8_t *, uintptr_t> func,
+                             BpfProgram &prog,
```
ajor, line 237:
```suggestion
                             BpfProgram &&prog,
```
Same here

ajor (commented):
The logic looks good to me now, just have a few comments about avoiding copies to improve performance.

Again, thanks for doing this work and pushing this through - I know it wasn't even your code in the first place.

viktormalik:
> The logic looks good to me now, just have a few comments about avoiding copies to improve performance.
> 
> Again, thanks for doing this work and pushing this through - I know it wasn't even your code in the first place.

Thanks for the review, these are great suggestions and I'm happy to improve the code as much as possible as this is one of the core parts.

src/bpfprogram.cpp
viktormalik: I'm a bit skeptical about this one, for two reasons:

1. It'd require us to store `BpfBytecode` as a non-const reference in `BpfProgram`, allowing accidental unwanted modifications.
2. There exists a 1-to-N relationship between `BpfBytecode` (essentially representing the entire ELF created by codegen) and `BpfProgram` (representing the instructions attached to a single probe). While it'd be possible to move out the section with probe instructions from `BpfBytecode` to `BpfProgram`, as it's not supposed to be used in another `BpfProgram`, this won't be possible with other sections that we intend to add in near future (`.BTF`, `.BTF.ext`, `.text`) b/c those will be re-used among multiple `BpfProgram`s.

With all this, I'm inclined towards keeping the copy here for the sake of avoiding hard-to-find bugs where one probe messes with the code of another.

ajor: Good points. How about we just make `code_` be a reference into `bytecode_` instead of copying it? It doesn't look like it's modified anywhere

ajor: Good points. Can we just make `code_` a reference into `bytecode_` instead? Then we avoid the copy and the data ownership is essentially the same as what we had before.

The only modification we make to `code_` is relocation, which I think is safe to do multiple times (not that we would currently do this anyway).

viktormalik: ATM yes, but that will change once we add support for subprograms. Most importantly, we'll need to append `.text` to the current `code_` and pass the entire thing to libbpf. So, I don't think that we'll avoid copying it but since I'm adding the support in parts, I can just make it a reference for now.

viktormalik: And we'd still need to make the reference to `BpfBytecode` non-const.

ajor: Ah yes, I see it now looking at the full stack. Fair enough, if it'll make the whole thing simpler then let's just always copy it here for now.

I'm not sure what the purpose of the reserve and assign originally were, can they not be simplified to this?
```suggestion
  code_ = bytecode_.at(name_);
```

viktormalik: It will make the whole thing much simpler for the moment. Also, I'm working on some follow-ups and it's possible that we'll need to introduce more abstraction to `BpfProgram` so I may revisit this later.

> I'm not sure what the purpose of the reserve and assign originally were, can they not be simplified to this?

I don't see a reason why not, I updated it.

ajor (approved):

PR #2793, opened 2023-10-06
Fix JSON output for cgroup_path

### LINKED ISSUE #2792: cgroup_path not being quoted in json output

Simple repro: `sudo bpftrace -f json -e 'kprobe:do_nanosleep { print((pid, cgroup_path(cgroup))); }'`

Output:

{"type": "value", "data": [1423939,unified:/user.slice/user-149108.slice/session-14593.scope]}
{"type": "value", "data": [556752,unified:/user.slice/user-149108.slice/user@149108.service/vscode_lsp_pyre.slice/run-r5cbc86191c244922a3019e345f62cff2.scope]}
{"type": "value", "data": [1423939,unified:/user.slice/user-149108.slice/session-14593.scope]}

Notice how the path doesn't have quotes around it. This breaks tools that parse this json.

viktormalik:
As #2792 noted, for some reason, cgroup path is not surrounded by quotes in the output which poses problems for JSON parsers as it may contain JSON-reserved characters (`:`). This is strange, because the same doesn't happen with other builtins, e.g. `inet`, which also have `:` in the output.

Anyways, add surrounding quotes manually and write a test for this. The test passes the produced JSON to Python's JSON parser to see if it parses correctly. Prior to this change, the test fails.

Fixes #2792.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to
 #### Deprecated
 #### Removed
 #### Fixed
+- Fix JSON output for cgroup_path
+  - [#2793](https://github.com/iovisor/bpftrace/pull/2793)
 #### Docs
 #### Tools
 - Update runqlen.bt to remove `runnable_weight` field from cfs_rq struct.
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -11,7 +11,7 @@ bool is_quoted_type(const SizedType &ty)
 {
   return ty.IsKstackTy() || ty.IsUstackTy() || ty.IsKsymTy() || ty.IsUsymTy() ||
          ty.IsInetTy() || ty.IsUsernameTy() || ty.IsStringTy() ||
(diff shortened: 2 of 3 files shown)


jordalgo:
@viktormalik Wow! Thanks for the very fast fix!

ajor (changes requested):
I think the `is_quoted_type()` function would be the right place to put this fix:
https://github.com/iovisor/bpftrace/blob/281a67a71d06b917d107cd0e85eac7b23ace2b42/src/output.cpp#L10-L15

viktormalik:
> I think the `is_quoted_type()` function would be the right place to put this fix:
> 
> https://github.com/iovisor/bpftrace/blob/281a67a71d06b917d107cd0e85eac7b23ace2b42/src/output.cpp#L10-L15
> 
> This would explain how inet works already.

Ah, of course, I completely missed that.

Many thanks! Fixed.

ajor (approved):
Looks good, thank you!

PR #2795, opened 2023-10-08
Use direct increment for count builtin instead of map update

jordalgo:
##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1478,6 +1478,53 @@ void IRBuilderBPF::CreateAtomicIncCounter(int mapid, uint32_t idx)
   CreateLifetimeEnd(key);
 }
 
+void IRBuilderBPF::CreateMapElemInc(Value *ctx,
+                                    Map &map,
+                                    Value *key,
+                                    const location &loc)
+{
+  int mapid = bpftrace_.maps[map.ident].value()->id;
+  CallInst *call = createMapLookup(mapid, key);
+  SizedType &type = map.type;
+
+  Function *parent = GetInsertBlock()->getParent();
+  BasicBlock *lookup_success_block = BasicBlock::Create(module_.getContext(),
+                                                        "lookup_success",
+                                                        parent);
+  BasicBlock *lookup_failure_block = BasicBlock::Create(module_.getContext(),
+                                                        "lookup_failure",
+                                                        parent);
+  BasicBlock *lookup_merge_block = BasicBlock::Create(module_.getContext(),
+                                                      "lookup_merge",
+                                                      parent);
+
+  AllocaInst *value = CreateAllocaBPF(type, "lookup_elem_val");
+  Value *condition = CreateICmpNE(
+      CreateIntCast(call, getInt8PtrTy(), true),
+      ConstantExpr::getCast(Instruction::IntToPtr, getInt64(0), getInt8PtrTy()),
+      "map_lookup_cond");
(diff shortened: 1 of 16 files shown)


jordalgo:
I suck at LLVM IR - also not sure about the name `CreateAtomicMapInc` and if it should be returning anything.

viktormalik (approved):
LGTM, thanks!

> I suck at LLVM IR - also not sure about the name `CreateAtomicMapInc` and if it should be returning anything.

IMHO it shouldn't and the name is fine :-)

Just one nit - `CreateAtomicMapInc` shares a large piece of code with `CreateMapLookupElem`. I'm wondering if we could do some refactoring to make them use the same code.

src/ast/irbuilderbpf.cpp
```diff
@@ -1478,6 +1477,60 @@ void IRBuilderBPF::CreateAtomicIncCounter(int mapid, uint32_t idx)
...
+                    getInt64(1),
+                    1,
+                    AtomicOrdering::Monotonic);
+
+  CreateBr(lookup_merge_block);
+
+  SetInsertPoint(lookup_failure_block);
+
+  // map item is not found, create one with value 1
+  CreateStore(getInt64(0), value);
```
xh4n3, line 1524: what is the use of this line?

src/ast/irbuilderbpf.cpp
```diff
@@ -1478,6 +1477,60 @@ void IRBuilderBPF::CreateAtomicIncCounter(int mapid, uint32_t idx)
...
+
+  // createMapLookup  returns an u8*
+  auto *cast = CreatePointerCast(call, value->getType(), "cast");
+
+  // map item is found, increment the value directly
+  CREATE_ATOMIC_RMW(AtomicRMWInst::BinOp::Add,
+                    cast,
+                    getInt64(1),
+                    1,
+                    AtomicOrdering::Monotonic);
```
danobi, line 1517: I'm not sure this should be an atomic op. I believe this'll generate code that takes exclusive access to the cache line. Which could be expensive at runtime.

Since `count()` already uses per-cpu maps, we can do a regular add operation on the map value. That should be equivalent and also faster for this use case IIUC.

src/ast/irbuilderbpf.cpp
jordalgo: Bad copy-pasta. Will remove.

src/ast/irbuilderbpf.cpp
```diff
@@ -1442,7 +1442,6 @@ void IRBuilderBPF::CreateAtomicIncCounter(int mapid, uint32_t idx)
 {
   AllocaInst *key = CreateAllocaBPF(getInt32Ty(), "key");
   CreateStore(getInt32(idx), key);
-
```
jordalgo, line 1445: Will fix.

jordalgo:
> Just one nit - CreateAtomicMapInc shares a large piece of code with CreateMapLookupElem. I'm wondering if we could do some refactoring to make them use the same code.

Good call. Maybe let me try that in a follow-up PR. As mentioned I'm still very much a noob at getting the LLVM IR right :)

danobi (approved):
lgtm. btw, would be good to put the motivation for this change into the commit msg.

jordalgo:
I realized I could use this same technique for `sum`: https://github.com/iovisor/bpftrace/pull/2798

Let me know if I should just abandon this diff and squash the two commits in 2798. I'm going to look into expanding this abstraction more but it's not as simple as just using it for `count` and `sum` so figured I'd start there.

PR #2796, opened 2023-10-09
Load `/sys/kernel/tracing/available_filter_functions` lazily (and speedup tests)

viktormalik:
I was wondering why unit tests are taking so long. After a bit of profiling, I discovered that most of the time is spent in reading `/sys/kernel/tracing/available_filter_functions`. The file is parsed in the `BPFtrace` constructor which is not strictly necessary, b/c we need to query the list for certain probe types only (and almost never in unit tests).

This changes the approach to read the list lazily by creating a wrapper for accessing `BPFtrace::traceable_functions_` which only loads the list when accessed for the first time.

The speedup is huge, especially in unit tests. See below for numbers on my local setup.

Before:
    
    Unit tests:
    [==========] 410 tests from 22 test suites ran. (852498 ms total)
    Runtime tests:
    [==========] 554 tests from 26 test cases ran. (528572 ms total)

After:

    Unit tests:
    [==========] 410 tests from 22 test suites ran. (41140 ms total)
    Runtime tests:
    [==========] 554 tests from 26 test cases ran. (301417 ms total)

The PR also contains a patch to pretty-print elapsed time of runtime tests.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -2525,13 +2525,22 @@ std::optional<int64_t> BPFtrace::get_int_literal(
   return std::nullopt;
 }
 
+const FuncsModulesMap &BPFtrace::get_traceable_funcs() const
+{
+  if (traceable_funcs_.empty())
+    traceable_funcs_ = parse_traceable_funcs();
+
+  return traceable_funcs_;
+}
+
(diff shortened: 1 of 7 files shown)


danobi (approved):
nice!

PR #2797, opened 2023-10-11
elf_parser: Fix invalid-null-argument

chantra:
When running
```
bpftrace -e 'kprobe:do_nanosleep { printf("PID %d sleeping...\n", pid);
```
with ASAN enabled, it was failing asan with `UndefinedBehaviorSanitizer: invalid-null-argument`:

```
src/ast/elf_parser.cpp:46:19: runtime error: null pointer passed as argument 1, which is declared to never be null
```

This change makes sure shdr->sh_size is non-zero before attempting memcpy

### DIFF
--- a/src/ast/elf_parser.cpp
+++ b/src/ast/elf_parser.cpp
@@ -38,7 +38,7 @@ BpfBytecode parseBpfBytecodeFromElfObject(void *const elf)
     std::vector<uint8_t> data;
     data.resize(shdr->sh_size);
 
-    if (shdr->sh_type != SHT_NOBITS)
+    if (shdr->sh_size && shdr->sh_type != SHT_NOBITS)
     {
       // NOBITS sections occupy no size on disk but take up size in
       // memory. Copy the file data for all other sections.


viktormalik:
Is the command correct? It's missing a trailing `}'` so it's not even a valid bpftrace command.

FWIW, if I run
```
bpftrace -e 'kprobe:do_nanosleep { printf("PID %d sleeping...\n", pid); }'
```
with ASAN, I get no errors.

chantra:
> Is the command correct? It's missing a trailing `}'` so it's not even a valid bpftrace command.
> 
> FWIW, if I run
> 
> ```
> bpftrace -e 'kprobe:do_nanosleep { printf("PID %d sleeping...\n", pid); }'
> ```
> 
> with ASAN, I get no errors.

Yeah, I made a typo when I copied/pasted the command.


And ASAN was an gross approximation... it is a UBSAN error: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

viktormalik (approved):
> And ASAN was an gross approximation... it is a UBSAN error: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

Ah, I see. Anyways, the fix makes sense so I'm ok with taking it.

Thanks!

PR #2799, opened 2023-10-12
Add codegen-tests.sh to update or run the tests

jordalgo:
Right now if you go the 'nix' route of updating the codegen tests (because you're not on llvm-12) you have to first figure out that the script builds everything in a different directory "build-codegen-update" then run all the codegen tests separately there. This just adjusts the script so you can either run the tests or regenerate them and puts the script in the same place as the other test runner scripts. Also update the test documentation so it's clearer how to run and update the codegen tests.

##### Checklist

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -27,14 +27,21 @@ compares it (string compare) with the expected result, a file named by the
 second argument. The `NAME` macro holds the test name,  which is `call_avg` in
 this case.
 
+These tests run as part of the normal suite of unit tests if you are running LLVM 12.
+If not, you need to install 'nix' and run these tests via this script:
+`./tests/codegen-tests.sh`.
+
 #### Updating
 
-Run `./scripts/update_codegen_tests.sh` after making codegen changes up update
-the expected LLVM IR.
+**LLVM 12**
 
-Alternatively (if you need more control over which tests are updated), if the
-test is run with `BPFTRACE_UPDATE_TESTS=1` the `test` helper will update the IR
-instead of running the tests.
+If you are running LLVM 12 or want to only update specific tests with `--gtest_filter`
+run `<builddir>/tests/bpftrace_test` with `BPFTRACE_UPDATE_TESTS=1` and the `test`
+helper will update the IR instead of running the tests.
+
+**Not LLVM 12**
+
+Run `./tests/codegen-tests.sh -u`. This updates all the codegen tests.
 
 ## Runtime tests
 
--- a/scripts/update_codegen_tests.sh
+++ b/scripts/update_codegen_tests.sh
@@ -1,23 +1,3 @@
 #!/bin/bash
 
-# Updates codegen tests' expected LLVM IR
-#
-# Example usage:
-#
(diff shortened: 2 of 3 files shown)


viktormalik:
I'm wondering what is the motivation for this change. Maybe mention it in the commit message?

jordalgo:
@viktormalik Right now if you go the 'nix' route of updating the codegen tests (because you're not on llvm-12) you have to first figure out that the script builds everything in a different directory "build-codegen-update" then run all the codegen scripts separately there. This just adjusts the script so you can either run the tests or regenerate them and puts the script in the same place as the other test runner scripts. But I will add this to the commit message.

tests/README.md
```diff
@@ -27,14 +27,21 @@ compares it (string compare) with the expected result, a file named by the
 second argument. The `NAME` macro holds the test name,  which is `call_avg` in
 this case.
 
+These tests run as part of the normal suite of unit tests unless you are not
+running LLVM 12, in which case you need to install 'nix' and run these via this
```
danobi, line 31: nit: easier to read without double negative
```suggestion
These tests run as part of the normal suite of unit tests if you are
running LLVM 12. If not,  you need to install 'nix' and run these via this
```

tests/README.md
```diff
@@ -27,14 +27,21 @@ compares it (string compare) with the expected result, a file named by the
...
+script `./tests/codegen-tests.sh`.
+
 #### Updating
 
-Run `./scripts/update_codegen_tests.sh` after making codegen changes up update
-the expected LLVM IR.
+**LLVM 12**
+
+If you are running LLVM 12 or if you need more control over which tests are updated,
```
danobi, line 38: Unclear what "need more control" means here. I'm guessing you mean you can combine with `--gtest_filter` to only update a subset of codegen tests?

danobi (approved):
Makes sense to me. See nits pls

tests/README.md
jordalgo: Yup, updated the language to make it clearer.

tests/README.md
```diff
@@ -27,14 +27,21 @@ compares it (string compare) with the expected result, a file named by the
...
+If not, you need to install 'nix' and run these tests via this script:
+`./tests/codegen-tests.sh`.
+
 #### Updating
 
-Run `./scripts/update_codegen_tests.sh` after making codegen changes up update
-the expected LLVM IR.
+**LLVM 12**
+
+If you are running LLVM 12 or want to only update specifc tests with `--gtest_filter`
```
viktormalik, line 38: Typo:
```suggestion
If you are running LLVM 12 or want to only update specific tests with `--gtest_filter`
```

viktormalik (approved):
LGTM, just one typo. Also would be nice to respect the 50/72 rule for commit messages.

Thanks for the change!

tests/README.md
jordalgo: Good catch. I gotta run the spell checker more often.

jordalgo:
Also updated the commit message

PR #2806, opened 2023-10-17
[min] update value directly in memory if available

jordalgo:
Similar to changes in 'count' and 'sum', if the element already exists in the map, just update the value in memory directly instead of calling `CreateMapUpdateElem`.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -404,29 +404,57 @@ void CodegenLLVM::visit(Call &call)
   {
     Map &map = *call.map;
     auto [key, scoped_key_deleter] = getMapKey(map);
-    Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
-    AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
+    CallInst *lookup = b_.CreateMapLookup(map, key);
+    SizedType &type = map.type;
 
-    // Store the max of (0xffffffff - val), so that our SGE comparison with uninitialized
-    // elements will always store on the first occurrence. Revent this later when printing.
-    Function *parent = b_.GetInsertBlock()->getParent();
     auto scoped_del = accept(call.vargs->front());
     // promote int to 64-bit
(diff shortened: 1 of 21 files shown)


src/utils.cpp
```diff
@@ -1337,28 +1337,15 @@ std::optional<std::string> abs_path(const std::string &rel_path)
...
-  int64_t val, max = 0, retval;
+  int64_t val, max = 0;
   for (int i = 0; i < nvalues; i++)
   {
     val = read_data<int64_t>(value.data() + i * sizeof(int64_t));
     if (val > max)
       max = val;
   }
 
-  /*
```
jordalgo, line 1348: Since we no longer need to compare to the uninitialized value, we can remove this hack.

jordalgo:
If this looks good, I'll also do this for "max", "avg", "hist", "lhist", etc..

src/ast/passes/codegen_llvm.cpp
```diff
@@ -404,29 +404,59 @@ void CodegenLLVM::visit(Call &call)
...
     b_.SetInsertPoint(ge);
-    b_.CreateStore(inverted, newval);
-    b_.CreateMapUpdateElem(ctx_, map, key, newval, call.loc);
+    b_.CreateStore(expr_, cast);
     b_.CreateBr(lt);
 
     b_.SetInsertPoint(lt);
-    b_.CreateLifetimeEnd(newval);
+
+    b_.CreateBr(lookup_merge_block);
```
danobi, line 450: The `lt` basic block looks extraneous to me. Can the `CreateCondBr()` above jump to `lookup_merge_block` instead of `lt`?

danobi (commented):
lgtm overall, nice cleanup!

danobi:
> If this looks good, I'll also do this for "max", "avg", "hist", "lhist", etc..

Yes, please do

src/ast/passes/codegen_llvm.cpp
jordalgo: Removed. Good suggestion.

danobi (approved):

PR #2810, opened 2023-10-22
Add support to use kernel uprobe multi bpf link

olsajiri:
Adding support to use uprobe_multi kernel for multiple uprobes.

It's useful for attaching bigger amount of uprobes, because it's faster and
saves file descriptors, because it does not create perf event for each uprobe.

##### Checklist


src/bpffeature.cpp
```diff
@@ -401,6 +403,78 @@ bool BPFfeature::has_kprobe_multi()
   return *has_kprobe_multi_;
 }
 
+extern "C" __attribute__((noinline)) __attribute__((unused)) void uprobe_multi_test_func(
+    void)
+{
+}
```
danobi, line 415: I'm afraid this won't work. Packagers typically strip symbols. We've had a lot of issues in the past with BEGIN/END probes being implemented with uprobes inside bpftrace binary. Which is why that moved to raw tracepoint prog type.

Is there another way to feature probe? Perhaps by passing an invalid symbol name and checking for EINVAL (rather than ENOSYS or something).

man/adoc/bpftrace.adoc
```diff
@@ -137,6 +137,12 @@ u:lib.so:"fn(char const*)" { printf("arg0:%s\n", str(arg0));}
 *--info*::
   Print detailed information about features supported by the kernel and the bpftrace build.
 
+*--disable-feature* _feature,feature,..._::
+  Disable detected features, valid values are::
+
+  *uprobe_multi* to disable uprobe_multi link +
+  *kprobe_multi* to disable kprobe_multi link
```
danobi, line 144: How about `--no-feature` instead? It's shorter and is more similar to gcc. I plan on adding `--feature` at some point for the AOT stuff (soon I promise...)

src/attached_probe.cpp
```diff
@@ -960,8 +970,176 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+       it != addrs->end();
+       ++it)
+  {
+    struct addr_offset &a = *it;
+
+    if (a.addr >= v_addr && a.addr < (v_addr + mem_sz))
+    {
+      a.offset = a.addr - v_addr + file_offset;
+    }
+  }
```
danobi, line 1014: Looks familiar to some of the code in bcc. bcc doesn't export any of this functionality?

src/bpffeature.cpp
olsajiri: right, actually we can do that, it's better.. I actually did that in libbpf, will change

man/adoc/bpftrace.adoc
olsajiri: sure, will change

src/attached_probe.cpp
olsajiri: I can't see any.. it's basically what we do for single uprobe but adapted to do it for multiple symbols,
but I guess we could have just one function to do that for both cases then.. I'll check on that

olsajiri: actually I don't think the unifying will make the code easier, I'd rather keep it as it is

src/attached_probe.cpp
```diff
@@ -960,8 +970,176 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+int bcc_load_cb(uint64_t v_addr, uint64_t mem_sz, uint64_t file_offset, void *p)
+{
+  std::vector<struct addr_offset> *addrs =
+      static_cast<std::vector<struct addr_offset> *>(p);
+
+  for (std::vector<struct addr_offset>::iterator it = addrs->begin();
+       it != addrs->end();
+       ++it)
+  {
+    struct addr_offset &a = *it;
```
viktormalik, line 1008: Can we use a range-based loop?
```suggestion
  for (auto &a : addrs)
  {
```

viktormalik (commented):
Overall LGTM, thanks!

I just have a couple of nits, mainly suggestions to use C++ features which make the code more safe and legible.

man/adoc/bpftrace.adoc
```diff
@@ -137,6 +137,12 @@ u:lib.so:"fn(char const*)" { printf("arg0:%s\n", str(arg0));}
 *--info*::
   Print detailed information about features supported by the kernel and the bpftrace build.
 
+*--no-feature* _feature,feature,..._::
+  Disable detected features, valid values are::
+
+  *uprobe_multi* to disable uprobe_multi link +
+  *kprobe_multi* to disable kprobe_multi link
```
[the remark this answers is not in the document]
olsajiri: 1) it's basically the only way now how to attach kprobe or uprobe with the old interface.. as they use different attachment link you might end up in situation you want to attach using the older interface.. like I needed this  to have an easy way to compare them, but maybe also when the new one is broken on your system then the old one is handy, or kprobe and fprobe also have different re-entry checks, so you might want to get and try more restricted kprobe layer.. I think it's handy to have that option

2) right, I will make it more strict and check the actual values

src/attached_probe.cpp
```diff
@@ -960,8 +970,176 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+
+  // Resolve probe_.funcs into offsets and syms vector
+  resolve_offset_uprobe_multi(
+      probe_.path, probe_.name, probe_.funcs, syms, offsets);
+
+  // Attach uprobe through uprobe_multi link
+  DECLARE_LIBBPF_OPTS(bpf_link_create_opts, opts);
+
+  opts.uprobe_multi.path = probe_.path.c_str();
+  opts.uprobe_multi.offsets = &offsets[0];
```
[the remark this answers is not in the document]
olsajiri: I'd think overloaded [] operator screams C++ loud enough ;-) but yes, it looks better

src/attached_probe.cpp
```diff
@@ -960,8 +970,176 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  for (i = 0; i < funcs.size(); i++)
+  {
+    auto pos = funcs[i].find(':');
+
+    if (pos == std::string::npos)
+    {
+      throw std::runtime_error("Error resolving probe: '" + probe_name + "'");
+    }
+
+    syms.push_back(funcs[i].c_str() + pos + 1);
```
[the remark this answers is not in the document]
olsajiri: ok

src/bpftrace.cpp
```diff
@@ -167,6 +167,26 @@ int BPFtrace::add_probe(ast::Probe &p)
...
+        Probe probe;
+        probe.attach_point = attach_point->func;
+        probe.path = attach_point->target;
+        probe.type = probetype(attach_point->provider);
+        probe.log_size = log_size_;
+        probe.orig_name = p.name();
+        probe.name = attach_point->name(attach_point->target,
+                                        attach_point->func);
+        probe.index = p.index();
+        probe.funcs.assign(attach_funcs.begin(), attach_funcs.end());
```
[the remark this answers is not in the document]
olsajiri: ok

src/attached_probe.cpp
olsajiri: ok

src/attached_probe.cpp
```diff
@@ -960,8 +970,176 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  }
+
+  return 0;
+}
+
+static void resolve_offset_uprobe_multi(std::string path,
+                                        std::string probe_name,
+                                        std::vector<std::string> &funcs,
+                                        std::vector<std::string> &syms,
+                                        std::vector<uint64_t> &offsets)
```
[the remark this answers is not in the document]
olsajiri: ok

viktormalik (commented):
One more thing just came across my mind: let's update the libbpf submodule to a version that contains uprobe multi support.

viktormalik (approved):
Nice, thanks!

danobi (approved):

src/bpffeature.cpp
```diff
@@ -401,6 +404,65 @@ bool BPFfeature::has_kprobe_multi()
...
+
+  progfd = bpf_prog_load(static_cast<::bpf_prog_type>(
+                             libbpf::BPF_PROG_TYPE_KPROBE),
+                         "uprobe_multi",
+                         "GPL",
+                         reinterpret_cast<struct bpf_insn*>(insns),
+                         ARRAY_SIZE(insns),
+                         &load_opts);
+
+  if (progfd >= 0)
```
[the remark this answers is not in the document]
olsajiri: you are right, I set err as 0 which should make it ok,
I wonder why there's no compiler warning

PR #2814, opened 2023-10-30
Etc ptr

jordalgo:
Similar to previous PRs, if the element already exists in the map, just update the value in memory directly instead of calling CreateMapUpdateElem.

These 4 commits change "max", "avg", "stats", "hist", and "lhist".

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -590,15 +590,9 @@ void CodegenLLVM::visit(Call &call)
                                   "linear");
 
     AllocaInst *key = getHistMapKey(map, linear);
+    b_.CreateMapElemAdd(ctx_, map, key, b_.getInt64(1), call.loc);
 
-    Value *oldval = b_.CreateMapLookupElem(ctx_, map, key, call.loc);
-    AllocaInst *newval = b_.CreateAllocaBPF(map.type, map.ident + "_val");
-    b_.CreateStore(b_.CreateAdd(oldval, b_.getInt64(1)), newval);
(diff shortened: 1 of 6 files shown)


jordalgo:
@danobi Let me know if I should squash these 4 commits.

jordalgo:
>  could you please wrap the commit messages to 72 chars and maybe unify the titles (enclose hist and lhist prefixes into [])? 

Done. Sorry about that. Need to fix my .gitconfig.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -461,26 +461,59 @@ void CodegenLLVM::visit(Call &call)
...
+    BasicBlock *lookup_merge_block = BasicBlock::Create(module_->getContext(),
+                                                        "lookup_merge",
+                                                        parent);
+
+    AllocaInst *value = b_.CreateAllocaBPF(type, "lookup_elem_val");
+    Value *condition = b_.CreateICmpNE(
+        b_.CreateIntCast(lookup, b_.getInt8PtrTy(), true),
+        ConstantExpr::getCast(Instruction::IntToPtr,
+                              b_.getInt64(0),
+                              b_.getInt8PtrTy()),
```
danobi, line 489: Could probably make a helper for null ptr at some point

danobi (approved):
lgtm

src/ast/passes/codegen_llvm.cpp
jordalgo: Ha. True. I'll do that as a follow-up.

PR #2819, opened 2023-11-03
Fix documentation of Environment variables

jordalgo:
- alphabetize env vars in documentation
- sync adoc and reference_guide
- add some missing env variables
- remove reference to BPFTRACE_NO_USER_SYMBOLS which is not being used

##### Checklist


docs/reference_guide.md
```diff
@@ -168,15 +168,21 @@ OPTIONS:
...
+    BPFTRACE_MAX_BPF_PROGS              [default: 512] max number of generated BPF programs
+    BPFTRACE_MAP_KEYS_MAX               [default: 4096] max keys in a map
+    BPFTRACE_MAX_PROBES                 [default: 512] max number of probes bpftrace can attach to
+    BPFTRACE_MAX_TYPE_RES_ITERATIONS    [default: 0] number of levels of nested field accesses for tracepoint args
+    BPFTRACE_NO_CPP_DEMANGLE            [default: 0] disable C++ symbol demangling
+    BPFTRACE_PERF_RB_PAGES              [default: 64] number of pages to allocate per CPU for perf ring buffer
+    BPFTRACE_STACK_MODE                 [default: bpftrace] Output format for ustack and kstack builtins
+    BPFTRACE_STRLEN                     [default: 64] bytes on BPF stack per str()
+    BPFTRACE_STR_TRUNC_TRAILER          [default: `..`] trailer to add to strings that were truncated
+    BPFTRACE_VMLINUX                    [default: none] vmlinux path used for kernel symbol resolution
```
danobi, line 185: Any reason to only alphabetize in the copy-paste in docs and not the binary too? I have a minor preference to not modify the env vars in the binary so git-blame works better but would also like consistency

docs/reference_guide.md
```diff
@@ -469,68 +475,80 @@ bpftrace v0.8-90-g585e-dirty
 
 ## 9. Environment Variables
 
-### 9.1 `BPFTRACE_STRLEN`
+### 9.1 `BPFTRACE_BTF`
```
danobi, line 478: Alphabetizing here makes sense

danobi (approved):

docs/reference_guide.md
```diff
@@ -168,15 +168,21 @@ OPTIONS:
...
-    BPFTRACE_NO_CPP_DEMANGLE    [default: 0] disable C++ symbol demangling
-    BPFTRACE_MAP_KEYS_MAX       [default: 4096] max keys in a map
-    BPFTRACE_MAX_PROBES         [default: 512] max number of probes bpftrace can attach to
-    BPFTRACE_MAX_BPF_PROGS      [default: 512] max number of generated BPF programs
-    BPFTRACE_CACHE_USER_SYMBOLS [default: auto] enable user symbol cache
-    BPFTRACE_VMLINUX            [default: none] vmlinux path used for kernel symbol resolution
-    BPFTRACE_BTF                [default: none] BTF file
-    BPFTRACE_STACK_MODE         [default: bpftrace] Output format for ustack and kstack builtins
+    BPFTRACE_BTF                        [default: none] BTF file
+    BPFTRACE_CACHE_USER_SYMBOLS         [default: (see below)] enable user symbol cache
```
danobi, line 172: The `see below` is a little misleading cuz this block is intended to be a copy paste of the binary output.

docs/reference_guide.md
jordalgo: I think I just forgot to alphabetize in the binary. Personally, I think it's worth the consistency but don't feel strongly.

docs/reference_guide.md
jordalgo: I don't like it either but the default looks like this "Default: PER_PROGRAM if ASLR disabled or `-c` option given, PER_PID otherwise."

danobi: Maybe "see docs" instead?

docs/reference_guide.md
danobi: Yeah I guess we shouldn't make ux decisions based on git blame. There are other ways to blame an env var addition anyways. Feel free to alphabetize the whole thing

docs/reference_guide.md
jordalgo: Works for me.

jordalgo:
- Alphabetized main.cpp env vars
- changed "see below" to "see docs"
- Added "BPFTRACE_KERNEL_BUILD" and "BPFTRACE_KERNEL_SOURCE"

viktormalik (approved):

Issue #2756, opened 2023-09-06
Revive Dockerfiles


danobi:
The Dockerfiles served as good living documentation on the required dependencies to build bpftrace

We should keep them updated to the latest distro release

Can probably condition docker builds in CI on changes to cmake files (to reduce unnecessary builds in PRs)

PR #2824, opened 2023-11-10
Fixed cmake download path which had arch hardcoded to x86_64

havefuncoding1:
##### Checklist

### DIFF
--- a/docker/Dockerfile.focal
+++ b/docker/Dockerfile.focal
@@ -13,8 +13,10 @@ deb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-${LLVM_VERSION} main\n" 
     curl -L https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4052245BD4284CDD
 
-RUN curl -L --output /tmp/cmake.tar.gz \
-  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-x86_64.tar.gz \
+ARG ARCH
+RUN ARCH=$(uname -m) \
+  && curl -L --output /tmp/cmake.tar.gz \
+  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-${ARCH}.tar.gz \
   && tar -xf /tmp/cmake.tar.gz -C /usr/local/ --strip-components=1
 
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.focal
```diff
@@ -13,8 +13,10 @@ deb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-${LLVM_VERSION} main\n"
     curl -L https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4052245BD4284CDD
 
-RUN curl -L --output /tmp/cmake.tar.gz \
-  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-x86_64.tar.gz \
+ARG ARCH
+RUN ARCH=$(uname -m) \
+  && curl -L --output /tmp/cmake.tar.gz \
+  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-${ARCH}.tar.gz \
```
viktormalik, line 19: I don't think that this is going to do what you intend. The assignment to `ARCH` in `RUN` will override whatever was set by user, so you may as well do
```suggestion

RUN curl -L --output /tmp/cmake.tar.gz \
  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-$(uname -m).tar.gz \
```

docker/Dockerfile.llvm
```diff
@@ -34,8 +34,10 @@ RUN apt update \
                          --slave /usr/bin/g++ g++ /usr/bin/g++-8      \
   && cp /usr/lib/gcc/x86_64-linux-gnu/8/plugin/include/plugin-api.h /usr/local/include
 
-RUN curl -L --output /tmp/cmake.tar.gz \
-  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-x86_64.tar.gz \
+ARG ARCH
+RUN ARCH=$(uname -m) \
+  && curl -L --output /tmp/cmake.tar.gz \
+  https://github.com/Kitware/CMake/releases/download/v3.20.0/cmake-3.20.0-linux-${ARCH}.tar.gz \
```
viktormalik, line 40: Beware that `x86_64` is also used few lines above in this Dockerfile.

viktormalik (commented):
What is your use-case for these Dockerfiles? We're using them in the CI but they may soon go away (see #2742). If people are still using them, we'd like to know how so that we can revive them after CI doesn't need them anymore (see #2756).

docker/Dockerfile.focal
havefuncoding1: I was experimenting changes to get eBPF working on Yocto, and bpftrace is a tool I could use to verify functionalities on target system. So I had to build the project and hit compile errors.

docker/Dockerfile.llvm
havefuncoding1: Thanks for catching it. I searched all dockerfiles for x86_64 and replaced them with $(uname -m) in latest iteration.

danobi (approved):
seems reasonable to me

danobi:
We will probably rewrite the existing dockerfiles in the near future. But will be good to have these fixes in so it can be referenced

PR #2828, opened 2023-11-14
utils: use /data/local/tmp as temprary dir on Android

simpleton:
`/tmp` does not exist on Android and cannot be created by regular applications. 
Use `/data/local/tmp` as an alternate tmpDir.

### before
Test bfptrace on Android device. bpftrace raises runtime_error due to the `/tmp` is not available and creatable.
```bash
$ ./bpftrace  -e 't:timer:hrtimer_start { @[ksym(args->function)] = count(); }'
terminating with uncaught exception of type std::runtime_error: creating temporary path for kheaders.tar.xz failed
Aborted
```

### after
runtime_error got migrated.
```
$ bpftrace  -e 't:timer:hrtimer_start { @[ksym(args->function)] = count(); }'
Attaching 1 probe...
[..]
```

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -714,7 +714,11 @@ namespace {
   std::string unpack_kheaders_tar_xz(const struct utsname& utsname)
   {
     std::error_code ec;
+#if defined(__ANDROID__)
+    std_filesystem::path path_prefix{ "/data/local/tmp" };
+#else
     std_filesystem::path path_prefix{ "/tmp" };
+#endif
     std_filesystem::path path_kheaders{ "/sys/kernel/kheaders.tar.xz" };
     if (const char* tmpdir = ::getenv("TMPDIR")) {
       path_prefix = tmpdir;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to
   - [#2793](https://github.com/iovisor/bpftrace/pull/2793)
 - Fix silent truncation of 64-bit values in hist()
   - [#2822](https://github.com/iovisor/bpftrace/pull/2822)
+- utils: use /data/local/tmp as temprary dir on Android
+  - [#2828](https://github.com/iovisor/bpftrace/pull/2828)
 #### Docs
 #### Tools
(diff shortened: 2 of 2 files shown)


src/utils.cpp
```diff
@@ -714,7 +714,11 @@ namespace {
   std::string unpack_kheaders_tar_xz(const struct utsname& utsname)
   {
     std::error_code ec;
+#if defined(__ANDROID__)
+    std_filesystem::path path_prefix{ "/data/local/tmp" };
+#else
     std_filesystem::path path_prefix{ "/tmp" };
+#endif
     std_filesystem::path path_kheaders{ "/sys/kernel/kheaders.tar.xz" };
     if (const char* tmpdir = ::getenv("TMPDIR")) {
```
danobi, line 723: The above seems reasonable, but I'm wondering why `TMPDIR` is not set. Or if it is, what it's set to on android. Wikipedia says it's a posix standard: https://en.wikipedia.org/wiki/TMPDIR

simpleton: afaik, Android doesn't have global tmp dir due to security considerations. Instead global tmp dir, the apps are able to access the [app-specific cache dir](https://developer.android.com/reference/android/content/Context#getCacheDir()). (/data/data/${APP_ID}/cache or /data/user/0/$APP_ID}/cache). 
For the shell or testing frame, Android gives this `/data/local/tmp` as `TMPDIR`

danobi (commented):
Please add a CHANGELOG entry (see https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#changelog) but lgtm otherwise

danobi (approved):

tnovak:
FYI bpftrace will use the `TMPDIR` env variable if it exists, which is normally[*] set to `/data/local/tmp` on Android:

```
$ adb shell echo \$TMPDIR
/data/local/tmp
```

[*] It's set by `adbd` [here](https://android.googlesource.com/platform/packages/modules/adb/+/refs/tags/android-14.0.0_r15/daemon/shell_service.cpp#277), so this applies to `adb shell` but may not if, say, invoking bpftrace from another process or clearing the environment.

Issue #2817, opened 2023-11-01
Limit uprobes in shared libraries to a single process when `-p` is given


viktormalik:
### Is your feature request related to a problem? Please describe.

When attaching to uprobes in shared libraries, the `-p` option is ignored, i.e. library functions called from all processes are hit. Limiting to the calls from the given process would make more sense, e.g. the `gethostlatency.bt` tool (which attaches to several libc functions) could be used per-process:
```
# bpftrace gethostlatency.bt -p $(pgrep firefox)
```
At the moment, the above command will show host name lookup times for all processes, not just the firefox process. See #2813 for details (the reproducer from there can be used on non-NixOs systems, too).

### Describe the solution you'd like

The `bpf_attach_uprobe` function from BCC that we use has a `pid` parameter. At the moment, we're only using the parameter for the `BEGIN` and `END` probes, this could be expanded to also pass the value from `-p`. We're already doing something similar for USDT probes.

jordalgo:
Working on this one here: https://github.com/jordalgo/bpftrace/tree/uprobe-pid-targeting

PR #2830, opened 2023-11-17
[uprobe] support for pid targeting for shared libs

jordalgo:
This adds support for specifying the pid even when targeting a uprobe/uretprobe in a library shared by multiple pids.

Example:
```
sudo bpftrace -p 1899508 -e 'uprobe:libc:getaddrinfo {
  print((comm, pid));
}
```

[Issue 2817](https://github.com/iovisor/bpftrace/issues/2817)

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -191,7 +191,7 @@ AttachedProbe::AttachedProbe(Probe &probe,
       // If BPF_PROG_TYPE_RAW_TRACEPOINT is available, no need to attach prog
       // to anything -- we will simply BPF_PROG_RUN it
       if (!feature.has_raw_tp_special())
-        attach_uprobe(safe_mode);
+        attach_uprobe(getpid(), safe_mode);
       break;
     case ProbeType::kprobe:
       attach_kprobe(safe_mode);
@@ -199,10 +199,6 @@ AttachedProbe::AttachedProbe(Probe &probe,
     case ProbeType::kretprobe:
       check_banned_kretprobes(probe_.attach_point);
       attach_kprobe(safe_mode);
-      break;
-    case ProbeType::uprobe:
-    case ProbeType::uretprobe:
-      attach_uprobe(safe_mode);
       break;
     case ProbeType::tracepoint:
       attach_tracepoint();
@@ -239,7 +235,8 @@ AttachedProbe::AttachedProbe(Probe &probe,
                              BpfProgram &&prog,
                              int pid,
                              BPFfeature &feature,
-                             BTF &btf)
+                             BTF &btf,
+                             bool safe_mode)
     : probe_(probe), prog_(std::move(prog)), btf_(btf)
 {
   load_prog(feature);
@@ -251,6 +248,10 @@ AttachedProbe::AttachedProbe(Probe &probe,
     case ProbeType::watchpoint:
     case ProbeType::asyncwatchpoint:
       attach_watchpoint(pid, probe.mode);
+      break;
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+      attach_uprobe(pid, safe_mode);
       break;
     default:
       LOG(FATAL) << "invalid attached probe type \""
@@ -1091,7 +1092,7 @@ static void resolve_offset_uprobe_multi(const std::string &path,
   }
 }
 
-void AttachedProbe::attach_multi_uprobe(void)
+void AttachedProbe::attach_multi_uprobe(int pid)
 {
   std::vector<std::string> syms;
   std::vector<uint64_t> offsets;
@@ -1110,6 +1111,11 @@ void AttachedProbe::attach_multi_uprobe(void)
   opts.uprobe_multi.flags = probe_.type == ProbeType::uretprobe
                                 ? BPF_F_UPROBE_MULTI_RETURN
                                 : 0;
+  if (pid != 0)
+  {
+    opts.uprobe_multi.pid = pid;
+  }
+
   if (bt_verbose)
   {
     std::cout << "Attaching to " << probe_.funcs.size() << " functions"
@@ -1135,16 +1141,16 @@ void AttachedProbe::attach_multi_uprobe(void)
   }
 }
 #else
-void AttachedProbe::attach_multi_uprobe(void)
+void AttachedProbe::attach_multi_uprobe(int)
 {
 }
 #endif // HAVE_LIBBPF_UPROBE_MULTI
 
-void AttachedProbe::attach_uprobe(bool safe_mode)
+void AttachedProbe::attach_uprobe(int pid, bool safe_mode)
 {
   if (!probe_.funcs.empty())
   {
-    attach_multi_uprobe();
+    attach_multi_uprobe(pid);
     return;
   }
 
@@ -1156,7 +1162,7 @@ void AttachedProbe::attach_uprobe(bool safe_mode)
                                         eventname().c_str(),
                                         probe_.path.c_str(),
                                         offset_,
-                                        probe_.pid,
+                                        pid == 0 ? -1 : pid,
                                         0);
 
   if (perf_event_fd < 0)
--- a/src/attached_probe.h
+++ b/src/attached_probe.h
@@ -30,7 +30,8 @@ public:
                 BpfProgram &&prog,
                 int pid,
                 BPFfeature &feature,
-                BTF &btf);
+                BTF &btf,
+                bool safe_mode = true);
   ~AttachedProbe();
   AttachedProbe(const AttachedProbe &) = delete;
   AttachedProbe &operator=(const AttachedProbe &) = delete;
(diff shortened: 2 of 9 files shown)


jordalgo:
Note: I looked into making a runtime test for this but it seems pretty tricky in that you'd have to run two programs that use the same shared library and assert you weren't getting samples for one of them.

ajor:
I'm a bit conflicted about this feature. I can definitely see it being useful to retrofit PID-filtering into existing scripts, but I feel like the better way of doing this would be to have the script include a PID predicate itself:
```
uprobe:libc:getaddrinfo
/ $# == 0 || pid == $1 /
{
  print((comm, pid));
}
```

If we sort out script-level argument parsing properly then this could be made more powerful by filtering on multiple PIDs, e.g. `--pid 123 --pid 124` or `--pids 123,124` and `/ @cli_pids[pid] == 1 /`.

ajor:
Seems reasonable to put in at least until we do have proper argument parsing support though.

This new feature applies to all uprobes, not just on shared libs, right?
A runtime test could be done as a program which takes a command line argument, and which is invoked multiple times with different arguments. If we trace this program and print the argument then we can check we've attached to the right PID.

jordalgo:
> If we sort out script-level argument parsing properly then this could be made more powerful by filtering on multiple PIDs

It's definitely more powerful doing it this way but I think this change is (at least for me) about matching expected behavior. Because we already support pid filtering for USDTs and uprobes/uretprobes (when there is a wildcard involved) I feel like it makes sense to do this as well for all uprobe/uretprobe invocations.

> This new feature applies to all uprobes, not just on shared libs, right?

Yeah I believe so.

> A runtime test could be done as a program which takes a command line argument, and which is invoked multiple times with different arguments. If we trace this program and print the argument then we can check we've attached to the right PID.

Maybe I'm confused but I think the problem is not so much attaching to the right pid but testing that we're not attaching to the other pids (or maybe you're saying I should write a NOT_EXPECT predicate). I could write a simple test that passing a pid argument doesn't affect attaching, which ensures we're not breaking existing behavior, but is not testing the new behavior. Maybe an example would help my brain 🧠

ajor:
> > If we sort out script-level argument parsing properly then this could be made more powerful by filtering on multiple PIDs
> 
> It's definitely more powerful doing it this way but I think this change is (at least for me) about matching expected behavior. Because we already support pid filtering for USDTs and uprobes/uretprobes (when there is a wildcard involved) I feel like it makes sense to do this as well for all uprobe/uretprobe invocations.

Yep, makes sense.

For the tests, I was imagining something like this:
```
BEFORE ./a.out 123
BEFORE ./a.out 456 # SOMEHOW PICK THIS ONE
BEFORE ./a.out 789
RUN bpftrace -p $PID -e 'BEGIN { print("BEGIN") } uprobe:a.out:func { printf(arg0) }' END { print("END") }'
EXPECT "BEGIN\n456\nEND"
TIMEOUT 5
```
Although I don't know how well the test framework will cope with this, particularly getting the PID of a process.

jordalgo:
@ajor Thanks for the suggestion on the test. I got around the limitation of having multiple BEFOREs but needing the BEFORE_PID by just spinning up a program that forks. I confirmed that this test fails on master.

src/attached_probe.cpp
```diff
@@ -191,7 +191,7 @@ AttachedProbe::AttachedProbe(Probe &probe,
       // If BPF_PROG_TYPE_RAW_TRACEPOINT is available, no need to attach prog
       // to anything -- we will simply BPF_PROG_RUN it
       if (!feature.has_raw_tp_special())
-        attach_uprobe(safe_mode);
+        attach_uprobe(0, safe_mode);
```
viktormalik, line 194: I'm not sure that this is correct. IIUC, we use process' PID in this case, set in [bpftrace.cpp:1090](https://github.com/iovisor/bpftrace/blob/master/src/bpftrace.cpp#L1090). This should probably use the same approach.

src/attached_probe.cpp
```diff
@@ -1146,7 +1152,7 @@ void AttachedProbe::attach_uprobe(bool safe_mode)
                                         eventname().c_str(),
                                         probe_.path.c_str(),
                                         offset_,
-                                        probe_.pid,
+                                        pid == 0 ? -1 : pid,
```
viktormalik, line 1155: I think that `bpftrace::Probe.pid` is not used anywhere else and we can drop it completely.

tests/runtime/scripts/uprobe_pid_check.bt
```diff
@@ -0,0 +1,18 @@
+BEGIN {
+    @count = 0;
+}
+
+uprobe:./testprogs/uprobe_fork_loop:uprobeFunction1 {
+    if (pid == $1) {
+        // make sure the pid is always what we expect
+        if (@count == 10) {
```
viktormalik, line 8: Just checking that I understand this correctly: `uprobe_fork_loop` creates 2 processes with different PIDs running `uprobeFunction1` and we check here that we hit the function from wanted PID 10 times before hitting the function from the other PID, which effectively means that we're not attached to the wrong PID.

Sounds a reasonable approach to me, maybe we could just add a comment documenting it somewhere? I can totally see myself staring into it in a couple of months and wondering what the hell is going on :-)

tests/testprogs/uprobe_fork_loop.c
```diff
@@ -0,0 +1,36 @@
+#include <unistd.h>
+
+int GLOBAL_A = 0x55555555;
+int GLOBAL_B = 0x88888888;
+int GLOBAL_C = 0x33333333;
+char GLOBAL_D = 8;
```
viktormalik, line 6: Unused?

viktormalik (commented):
>> If we sort out script-level argument parsing properly then this could be made more powerful by filtering on multiple PIDs

> It's definitely more powerful doing it this way but I think this change is (at least for me) about matching expected behavior. Because we already support pid filtering for USDTs and uprobes/uretprobes (when there is a wildcard involved) I feel like it makes sense to do this as well for all uprobe/uretprobe invocations.

Agreed with @jordalgo here, this is the expected behaviour to me. I don't see a situation where I would want to attach binary uprobes to a PID but attach library uprobes system-wide.

Just a couple of remarks, otherwise looks good, thanks!

tests/testprogs/uprobe_fork_loop.c
```diff
@@ -0,0 +1,36 @@
...
+int main(int argc __attribute__((unused)), char **argv __attribute__((unused)))
+{
+  if (fork() == 0)
+  {
+    spin();
+  }
+  else
+  {
+    spin();
+  }
```
ajor, line 33: `spin()` is run in both the true and false cases - can the if-statement be removed entirely?

tests/runtime/scripts/uprobe_pid_check.bt
jordalgo: > we check here that we hit the function from wanted PID 10 times before hitting the function from the other PID

In this test we should never hit the branch for the other PID. The counting till 10 is just kind of a delay to ensure we never hit the other pid during that time. But I'll add a comment to make that more clear.

tests/testprogs/uprobe_fork_loop.c
jordalgo: Good call. It can. This is what I get from just copying from the interwebs

jordalgo:
Updated based on comments.

tests/testprogs/uprobe_fork_loop.c
```diff
@@ -0,0 +1,30 @@
...
+    usleep(500);
+  }
+}
+
+int main(int argc __attribute__((unused)), char **argv __attribute__((unused)))
+{
+  pid_t p = fork();
+  if (p < 0)
+  {
+    perror("fork fail");
```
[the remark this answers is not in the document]
viktormalik: nit: missing `#include <stdio.h>`

tests/testprogs/uprobe_fork_loop.c
```diff
@@ -0,0 +1,30 @@
...
+    exit(1);
```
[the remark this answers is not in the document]
viktormalik: nit: missing `#include <stdlib.h>`

viktormalik (approved):
LGTM, just two nits (see the CodeQL warnings).

src/attached_probe.h
```diff
@@ -30,7 +30,8 @@ class AttachedProbe
                 BpfProgram &&prog,
                 int pid,
                 BPFfeature &feature,
-                BTF &btf);
+                BTF &btf,
+                bool safe_mode = false);
```
danobi, line 34: Why default to non safe mode?

tests/testprogs/uprobe_fork_loop.c
```diff
@@ -0,0 +1,32 @@
...
+    int n = 13;
+    char c = 'x';
+    uprobeFunction1(&n, c);
+    usleep(500);
+  }
+}
+
+int main(int argc __attribute__((unused)), char **argv __attribute__((unused)))
+{
+  pid_t p = fork();
```
danobi, line 23: Should we wait for the child? Otherwise zombie process right

src/attached_probe.h
jordalgo: This default isn't being used by anything checking safe_mode so fine to make it default to true.

tests/testprogs/uprobe_fork_loop.c
jordalgo: We want them to be running at the same time.

danobi: Ah yeah. Thought `spin()` had finite iterations. If we rely on runtime test runner to kill the process group this should be fine

jordalgo:
make `safe_mode` default to true

danobi (approved):

PR #2831, opened 2023-11-17
hist: log2 histograms with finer granularity

luigirizzo:
Allow a second optional argument in hist(n, k) to map each power of 2 into 2^k buckets, thus creating a logarithmic scale with finer granularity and modest runtime overhead (a couple of shifts and add/mask in addition to the original algorithm).

Allowed values of k are 0..5, with 0 as default for backward compatibility.

The implementation follows my earlier code in https://github.com/luigirizzo/lr-cstats

Example below:
```
$ sudo src/bpftrace -e 'kfunc:tick_do_update_jiffies64 { @ = hist((nsecs & 0xff),2); }' Attaching 2 probes...
@:
[0]                    4 |@                                                   |
[1]                    1 |                                                    |
[2]                    3 |@                                                   |
[3]                    2 |                                                    |
[4]                    3 |@                                                   |
[5]                    0 |                                                    |
[6]                    3 |@                                                   |
[7]                    2 |                                                    |
[8, 10)                5 |@                                                   |
[10, 12)               7 |@@                                                  |
[12, 14)               5 |@                                                   |
[14, 16)               6 |@@                                                  |
[16, 20)              11 |@@@                                                 |
[20, 24)              14 |@@@@                                                |
[24, 28)              20 |@@@@@@                                              |
[28, 32)              13 |@@@@                                                |
[32, 40)              40 |@@@@@@@@@@@@@                                       |
[40, 48)              38 |@@@@@@@@@@@@@                                       |
[48, 56)              35 |@@@@@@@@@@@                                         |
[56, 64)              29 |@@@@@@@@@                                           |
[64, 80)              72 |@@@@@@@@@@@@@@@@@@@@@@@@                            |
[80, 96)              64 |@@@@@@@@@@@@@@@@@@@@@                               |
[96, 112)             61 |@@@@@@@@@@@@@@@@@@@@                                |
[112, 128)            67 |@@@@@@@@@@@@@@@@@@@@@@                              |
[128, 160)           124 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@          |
[160, 192)           130 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@        |
[192, 224)           124 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@          |
[224, 256)           152 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
```

Performance:
most of the cost of hist() in bpftrace is in the bpf hash lookup,
and the cost of index calculation is negligible.
One way to measure the overall cost is the following
```
sudo taskset -c 1 src/bpftrace -e 'i:us:1 { $t = nsecs; @A = hist($t); $t = nsecs - $t; @ = lhist($t, 0, 5000, 100);} '
```
and on my AMD 5800 most of the samples are in the 900-100us range;
my estimate for index computation (from lr-cstats) is in the 10-20ns range.

Remember to check `/proc/sys/kernel/perf_` to make sure that the code
can generate a sufficient number of samples per second.

##### Checklist


docs/reference_guide.md
danobi, line 1: Also need update to `man/adoc/bpftrace.adoc` - we are working on deduplicating that and the reference guide

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3054,81 +3061,103 @@ Value *CodegenLLVM::createLogicalOr(Binop &binop)
...
   BasicBlock *is_less_than_zero = BasicBlock::Create(module_->getContext(), "hist.is_less_than_zero", log2_func);
   BasicBlock *is_not_less_than_zero = BasicBlock::Create(module_->getContext(), "hist.is_not_less_than_zero", log2_func);
-  b_.CreateCondBr(b_.CreateICmpSLT(b_.CreateLoad(b_.getInt64Ty(), n_alloc),
-                                   b_.getInt64(0)),
-                  is_less_than_zero,
-                  is_not_less_than_zero);
+
+  // code generation helper macros: Constant, LoaD
+#define _C(_n)		b_.getInt64(_n)
+#define _LD(_n)		b_.CreateLoad(b_.getInt64Ty(), _n)
```
danobi, line 3112: Nit: for consistency with rest of codebase, please avoid these macros

src/ast/passes/semantic_analyser.cpp
```diff
@@ -554,7 +554,16 @@ void SemanticAnalyser::visit(Call &call)
 
   if (call.func == "hist") {
     check_assignment(call, true, false, false);
-    check_nargs(call, 1);
+    check_varargs(call, 1, 2);
+    if (call.vargs->size() == 1) {
+      call.vargs->push_back(new Integer(0, call.loc)); // default bits is 0
+    } else {
+      check_arg(call, Type::integer, 1);
+      const auto bits = bpftrace_.get_int_literal(call.vargs->at(1));
```
danobi, line 562: Missing check for `std::nullopt`?

src/ast/passes/semantic_analyser.cpp
danobi, line 1: I think additional checks need to be added to prevent something like:
```
@ = hist(1,1);
@ = hist(1,2);
```
IOW, `k` should be encoded as part of the map type. Otherwise I think weird stuff will happen with bucket calculation.

danobi (commented):
I didn't look too close as the mathy bits yet. Mostly tried to leave some high level comments. This PR also needs tests (see https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#tests for some hints)

Overall, this feature makes sense to me. I'm taking it on faith that the $$2^{k}$$ restriction (versus taking any old integer) is to make the bit twiddling work efficiently.

danobi:
CI failures look legit. There should be some hints about debugging those in the developers.md doc I linked as well

src/ast/passes/semantic_analyser.cpp
luigirizzo: that should be handled in required_resources.cpp, hist_bits_arg makes sure that all instances of `hist()`  for the same target have the same second argument

src/ast/passes/codegen_llvm.cpp
luigirizzo: ack, will do in next revision

src/ast/passes/semantic_analyser.cpp
danobi: Ah right. Missed that. I have a weak preference for moving the semantic checks to `semantic_analyser` pass. Resource analysis is kinda meant to collect resources, not check for semantic issues. The weak preference is in case it's excessively complicated to do it in semantic analysis.

luigirizzo: will have a look at that. can you point me at an example of similar code that stores data on the map type?

src/ast/passes/semantic_analyser.cpp
luigirizzo: not sure. line 559 should add the second argument in case it is not supplied, am I missing something?

src/ast/passes/semantic_analyser.cpp
danobi: https://github.com/iovisor/bpftrace/blob/3afd12a9de750c54b01df9948b68a297d0f418c5/src/ast/passes/semantic_analyser.cpp#L2451 handles map assignments

My first thought is that you'd want to extend `SizedType` for `k`. Haven't done this in a bit tho, so might be missing some better ideas

src/output.cpp
```diff
@@ -698,24 +681,19 @@ std::string JsonOutput::hist_to_str(const std::vector<uint64_t> &values,
...
-    else
-    {
-      long low = 1ULL << (i - 2);
-      long high = (1ULL << (i - 2 + 1)) - 1;
-      res << "\"min\": " << low << ", \"max\": " << high << ", ";
+      uint32_t power = (i - 1) >> bits, n = 1<<bits, frac = (i - 1) & (n - 1);
+      low = (1ul << (power - 1) ) * (n + frac);
+      power = i >> bits;
+      frac = i & (n - 1);
+      high = (1ul << (power - 1)) * (n + frac);
```
luigirizzo, line 691: need to fix this. the text output prints open intervals [low,high), json uses closed intervals so max should be high-1

src/ast/passes/semantic_analyser.cpp
luigirizzo: actually, to clarify: I implemented the check on `hist()` second argument **exactly as the existing code** that checks `lhist()` arguments.
The code in `semantic_analyser.cpp` (this chunk for `hist()`, one just below for `lhist()`) compares arguments with previously seen values, and `required_resources.cpp` creates the state.
That should resolve the issue.

src/ast/passes/codegen_llvm.cpp
luigirizzo: done

luigirizzo:
> CI failures look legit. There should be some hints about debugging those in the developers.md doc I linked as well



> CI failures look legit. There should be some hints about debugging those in the developers.md doc I linked as well

I have fixed the json format ones and function prototype formatting.
The remaining failure should be in the code generation for hist() which I have not updated yet.

luigirizzo:
> I didn't look too close as the mathy bits yet. Mostly tried to leave some high level comments. This PR also needs tests (see https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#tests for some hints)

two questions here:
1. there seems to be no helper to compare the output of a named test with a file, so instead of
   ```
   NAME foo
   PROG ...
   EXPECT_FILE runtime/output/foo.xyz
   ```
   every test repeats the sequence
   ```
   RUN {{BPFTRACE}} args ... | python3 -c 'import sys,json; print(json.load(sys.stdin) == json.load(open("runtime/outputs/foo.xyz")))'
   EXPECT ^True$
    ```
    Is there any plan to add this kind of helpers in the runner engine ?
  2. unless I am mistaken I see no tests for the plain output format of hist/lhist etc. Shall I focus on json output only ?

src/ast/passes/semantic_analyser.cpp
luigirizzo: Ah, found the problem and fixed

luigirizzo:
> I didn't look too close as the mathy bits yet. Mostly tried to leave some high level comments. This PR also needs tests (see https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#tests for some hints)

Added a few tests in (semantic_analyser, call_hist) and one in runtime/json-output (histogram-finegrain), plus updated codegen for call_hist

src/output.cpp
luigirizzo: fixed and verified tests pass.

luigirizzo:
> CI failures look legit. There should be some hints about debugging those in the developers.md doc I linked as well

They should be fixed now.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -554,7 +554,17 @@ void SemanticAnalyser::visit(Call &call)
 
   if (call.func == "hist") {
     check_assignment(call, true, false, false);
-    check_nargs(call, 1);
+    if (!check_varargs(call, 1, 2))
+      return;
+    if (call.vargs->size() == 1) {
+      call.vargs->push_back(new Integer(0, call.loc)); // default bits is 0
```
danobi, line 560: I don't think we should be modifying the AST in semantic analysis. I see one other occurrence in semantic analysis where we inject a cast, but I think that's bad too. Ideally mutable passes should derive from `Mutator` visitor so it's explicit.

In this case, I think it'd be better to teach the other passes to use a default value of 0 where appropriate

src/output.cpp
```diff
@@ -35,32 +35,19 @@ std::ostream& operator<<(std::ostream& out, MessageType type) {
   return out;
 }
 
-std::string TextOutput::hist_index_label(int power)
+// power must be >= 2<<bits ?
```
danobi, line 38: This is a precondition right? The `?` is confusing - if it's indeed a precondition pls make it an `assert()` in the function

src/output.cpp
```diff
@@ -698,24 +683,19 @@ std::string JsonOutput::hist_to_str(const std::vector<uint64_t> &values,
...
-      res << "\"min\": " << low << ", \"max\": " << high << ", ";
+      uint32_t power = (i - 1) >> bits, n = 1 << bits, frac = (i - 1) & (n - 1);
+      low = (1ul << (power - 1)) * (n + frac);
+      power = i >> bits;
+      frac = i & (n - 1);
+      high = (1ul << (power - 1)) * (n + frac) - 1;
     }
+    if (low != -1)
+      res << "\"min\": " << low << ", ";
+    res << "\"max\": " << high << ", ";
```
danobi, line 697: Please add some documentation or unit tests, or anything else to help explain this. Or even slower/dumber code. I'd prefer dumber code cuz this is not in the fast path. I will probably never get enough time or be smart enough to understand the math here

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3054,81 +3061,99 @@ Value *CodegenLLVM::createLogicalOr(Binop &binop)
...
-  // values less than 0, index 1 is for 0, and indexes 2 onwards is the
-  // power-of-2 histogram index.
+  // Arguments: N (int64), K (0..5)
+  // Maps each power of 2 into 2^K buckets, so we can build fine-grained
+  // histograms with low runtime cost.
+  //
+  // Returns 0 for N < 0, 1..2^K for 0 <= N < 2^K; otherwise
+  // N has the form ..01xxx... where the leftmost is at bit L >= K,
+  // xxx are the following K bits with a value 0 <= X < 2^K.
+  // In this case the function returns 1 + (L-K + 1) * 2^K + X
```
danobi, line 3071: Similar to comment on `output.cpp`, I'm finding this hard to understand. The `..` in `..01` is confusing to me - is this supposed to be two bits or an arbitrary number of bits?

And "leftmost" and "L" are similarly ambiguous. Leftmost what? 

Also the "In this case" bit is confusing. What other cases are there?

Basically what I'm saying is that a very verbose explanation would be really helpful for mere mortals like me (who would otherwise have no hope of ever understanding given any reasonable amount of time)

danobi:
> > I didn't look too close as the mathy bits yet. Mostly tried to leave some high level comments. This PR also needs tests (see https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#tests for some hints)
> 
> two questions here: (but these are orthogonal, so feel free to ignore them...)
> 
>     1. there seems to be no helper to compare the output of a named test with a file, so instead of
>        ```
>        NAME foo
>        PROG ...
>        EXPECT_FILE runtime/output/foo.xyz
>        ```
>        every test repeats the sequence
>        ```
>        RUN {{BPFTRACE}} args ... | python3 -c 'import sys,json; print(json.load(sys.stdin) == json.load(open("runtime/outputs/foo.xyz")))'
>        EXPECT ^True$
>        ```        
>        Is there any plan to add this kind of helpers in the runner engine ?

Yeah that would be a good addition. I filed https://github.com/iovisor/bpftrace/issues/2841 . Feel free to take it if you're interested. Otherwise I will probably pick it up at some point

> 
>     2. unless I am mistaken I see no tests for the plain output format of hist/lhist etc. Shall I focus on json output only ?

The json output happened to be easier to test. In the past we've done regexes to match the text output but that's messy and we often got it wrong. Might be useful in this case to check in some "gold" files and did a regular text compare. Might be useful to have an `EXPECT_TEXT` directive just like with the above suggested `EXPECT_JSON` directive.

Manual testing would be ok for now. I'll file a ticket for `EXPECT_TEXT` and refer to this PR.

src/ast/passes/semantic_analyser.cpp
danobi: I think the code is still coercing the nullopts wrong. For example:

```
$ sudo ./build/src/bpftrace -e 'BEGIN { $k = 3; @ = hist(1,$k); exit(); }'
stdin:1:21-31: ERROR: hist: bits 94594394363600 must be 0..5
BEGIN { $k = 3; @ = hist(1,$k); exit(); }
```

See:
https://cpp.sh/?source=%2F%2F+Example+program%0A%23include+%3Ciostream%3E%0A%23include+%3Cstring%3E%0A%0Aint+main()%0A%7B%0A++std%3A%3Astring+name%3B%0A++std%3A%3Acout+%3C%3C+%22What+is+your+name%3F+%22%3B%0A++getline+(std%3A%3Acin%2C+name)%3B%0A++std%3A%3Acout+%3C%3C+%22Hello%2C+%22+%3C%3C+name+%3C%3C+%22!%5Cn%22%3B%0A%7D

I think you need an explicit check for nullopt

luigirizzo: hopefully fixed, I needed to check that the second argument to `hist()` is a compile time constant integer, for which I haven't found a helper, so I restricted to integer literal, same as it is done for `lhist()`.

I misunderstood how the `check_*()` functions work; based on their usage (typically the return value is not used) I thought they would throw an exception or set some global error variable that would further prevent other processing, but instead termination has to be explicit.

docs/reference_guide.md
```diff
@@ -3577,11 +3577,15 @@ and the total of the argument value. This is similar to using count(), avg(), an
...
 ```
-@histogram_name[optional_key] = hist(value)
+@histogram_name[optional_key] = hist(value[, k])
 ```
 
 This is implemented using a BPF map.
 
+Values are accumulated in 2^k buckets for each power of 2,
+with negative values in their own bucket.
+k can be 0..5, defaults to 0.
```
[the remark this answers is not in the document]
luigirizzo: added one example.

tests/semantic_analyser.cpp
```diff
@@ -387,6 +387,9 @@ TEST(semantic_analyser, compound_bxor)
 TEST(semantic_analyser, call_hist)
 {
   test("kprobe:f { @x = hist(1); }", 0);
+  test("kprobe:f { @x = hist(1, 0); }", 0);
+  test("kprobe:f { @x = hist(1, 5); }", 0);
+  test("kprobe:f { @x = hist(1, 10); }", 1);
```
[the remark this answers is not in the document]
luigirizzo: yes, added some more here and in resource_analyser.cpp

src/output.cpp
luigirizzo: added an assert and a lot of explanation (and one example) on how the function works.
Unfortunately the encoding is not terribly intuitive especially because of the initial entry for negative intervals (which is removed here).

src/output.cpp
luigirizzo: added detailed explanation in the Text version (and a reference to it), plus made a special case for the negative interval and the ones with only one value, so code is similar to the previous one.

src/ast/passes/codegen_llvm.cpp
luigirizzo: I added a description and an example, hopefully more clear now, PTAL

luigirizzo:
> Allow a second optional argument in hist(n, k) to map each power of 2 into 2^k buckets, thus creating a logarithmic scale with finer granularity and modest runtime overhead (a couple of shifts and add/mask in addition to the original algorithm).

anything else left to do here?

danobi:
Code needs a clang-format (https://github.com/iovisor/bpftrace/blob/master/docs/developers.md#code-style). Also would be good to use the new EXPECT_JSON for the runtime test.

Other than that, was waiting for enough coffee to hit the bloodstream to take a look at the algorithm again.

src/output.cpp
ajor: @danobi As a broad rule, I'd consider everything that executes during tracing part of the fast path (including text output). It's not uncommon for me to see dropped events because the bpftrace runtime can't keep up with printing events the BPF program is pumping out. If you just meant histograms, then you're probably right that no one will be printing them at such a high frequency. Documentation is always good though!,

ajor:
Thanks for the PR! Agree with @danobi that this feature looks reasonable. I just want to make sure before merging that it won't slow down histogram calculations for the default case (k=0).

The LLVM IR in the tests looks unoptimised so we'd probably need to look at the generated BPF bytecode to be sure. I can have a check next week if no one gets there before me.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3054,81 +3061,111 @@ Value *CodegenLLVM::createLogicalOr(Binop &binop)
...
 
-  // setup n and result registers
-  Value *arg = log2_func->arg_begin();
+  // storage for arguments
   Value *n_alloc = b_.CreateAllocaBPF(CreateUInt64());
-  b_.CreateStore(arg, n_alloc);
-  Value *result = b_.CreateAllocaBPF(CreateUInt64());
-  b_.CreateStore(b_.getInt64(0), result);
+  b_.CreateStore(log2_func->arg_begin(), n_alloc);
+  Value *k_alloc = b_.CreateAllocaBPF(CreateUInt64());
```
danobi, line 3111: Missing `b_.CreateLifetimeEnd` for `n_alloc` and `k_alloc`

src/ast/passes/codegen_llvm.cpp
danobi: Thanks! The description and example is very helpful

danobi (commented):
other than the outstanding comments (clang-format, semantic analysis ast modification, stack allocation lifetime tracking, EXPECT_JSON, and Alastair's todo), LGTM.

I'm kinda on the fence about the semantic analysis ast modification thing now. Probably fine the way it is.

luigirizzo:
> Thanks for the PR! Agree with @danobi that this feature looks reasonable. I just want to make sure before merging that it won't slow down histogram calculations for the default case (k=0).
> 
> The LLVM IR in the tests looks unoptimised so we'd probably need to look at the generated BPF bytecode to be sure. I can have a check next week if no one gets there before me.

I don't know if you have any better way to measure index computation (other than modifying the code generator to collect timestamps around it) but I suspect most of the time is spent in looking up the bpf hash entry, rather than computing the index.

For measuring the time I use the following:
```
sudo taskset -c 1 /tmp/bpftrace0 -e 'i:us:1 { $t = nsecs; @a = hist($t); $t = nsecs - $t; @ = lhist($t, 0, 5000, 100);} ' 
```
and on  my Beelink SER5 i see most entries in the 900,1000 slot with both the old and new code.

For reference, the same (native) code in `github.com:luigirizzo/lr-cstats` takes less than 20ns.

On the topic:
- it may be preferable to hook the code on some event firing at high frequency but I don't have any good idea (perhaps an external process spinning on a syscall and hooking the call?)
- the rate at which this generates events depends on `/proc/sys/kernel/perf*`, e.g. after installing a couple of perf-related packages I got the following
  ```
  /proc/sys/kernel/perf_cpu_time_max_percent:25
  /proc/sys/kernel/perf_event_max_contexts_per_stack:8
  /proc/sys/kernel/perf_event_max_sample_rate:39750 
  /proc/sys/kernel/perf_event_max_stack:127
  /proc/sys/kernel/perf_event_mlock_kb:516
  /proc/sys/kernel/perf_event_paranoid:-1
  ```
  and was stuck to some 6K events/s. Bumping `perf_event_max_sample_rate` to some 1 million I was able to run the above loop at most at 100K samples/s even with small intervals. `/proc/self/timerslack` did not seem to matter.

- if you make the probes less frequent (say 1ms) and allow the CPU to go into some deep sleep state, you will see values more scattered (in my  case bands are some 500ns apart, around 900, 1400, 1800, 2300, 2700) possibly due to various memory and tlb misses

luigirizzo:
> other than the outstanding comments (clang-format, semantic analysis ast modification, stack allocation lifetime tracking, EXPECT_JSON, and Alastair's todo), LGTM.
> 
> I'm kinda on the fence about the semantic analysis ast modification thing now. Probably fine the way it is.

I think I am done with the modifications from my side (see comment on why I believe lifetime annotations are useless in this case).

danobi:
> I think I am done with the modifications from my side (see comment on why I believe lifetime annotations are useless in this case).

I don't see the comment anywhere - forgot to send?

luigirizzo:
> > I think I am done with the modifications from my side (see comment on why I believe lifetime annotations are useless in this case).
> 
> I don't see the comment anywhere - forgot to send?

This one on line 3109 for codegen_llvm.cpp ?
----

I have looked up the lifetime annotations https://llvm.org/docs/LangRef.html#object-lifetime and I believe that for this function it is much better to leave it to the optimizer.

As an experiment I also tried to access the two call arguments directly from the registers instead of allocating arguments on the stack:

Value  *n = log2_func->arg_begin();
Value *k = log2_func->arg_begin() + 1;
(leaving it to the compiler whether or not to spill variables to the stack).
The unoptimized IR code is slightly better with these changes, but the optimized code is exactly the same in both cases.

danobi:
Hmm, I see `// l -= k;` on line 3109.

But in any case, I think you're right. log2() is a separate function. We can leave it to compiler in this case.

luigirizzo:
> @ajor did you get a chance to take a look? I suspect this will add a few more insns to the hot path. But ALU ops are quite cheap (basically free since it's all in registers) so I'm personally not too worried

FWIW (see other comments and updated description) I believe the hash lookup is probably 20-50 times more expensive than the index computation. This test
```
sudo taskset -c 1 src/bpftrace -e 'i:us:1 { $t = nsecs; @A = hist($t); $t = nsecs - $t; @ = lhist($t, 0, 5000, 100);} '
```
reports 900-1000ns for most samples with both old and new code.

In fact, it would be great if we could specialize the `hist()` and `lhist()` to use a bpf array instead of a map when there are no other keys.

src/ast/passes/resource_analyser.cpp
```diff
@@ -132,6 +132,21 @@ void ResourceAnalyser::visit(Call &call)
...
   }
+  else if (call.func == "hist")
+  {
+    auto &r = resources_.hist_bits_arg;
+    int old, bits = static_cast<Integer *>(call.vargs->at(1))->n;
+    if (r.find(call.map->ident) != r.end() &&
+        (old = r[call.map->ident]) != bits)
+    {
+      LOG(ERROR, call.loc, err_)
+          << "Different bits in a single hist, had " << old << " now " << bits;
```
ajor, line 143:
```suggestion
    int bits = static_cast<Integer *>(call.vargs->at(1))->n;
    if (r.find(call.map->ident) != r.end() &&
        (r[call.map->ident]) != bits)
    {
      LOG(ERROR, call.loc, err_)
          << "Different bits in a single hist, had " << r[call.map->ident] << " now " << bits;
```
The multiple declarations and assignment/comparision in single lines feel too clever for me. Since this is just an error path there's no need to optimise it anyway, so let's just read the map twice to simplify the code.

ajor (commented):
Sorry for the delay. LLVM does optimise the code well and the only extra instructions in the default case are some extra bit twiddling for the higher bits:
```
  hist.is_not_zero.i:                               ; preds = %hist.is_not_less_than_zero.i
+   %a = icmp ugt i64 %input, 4294967295                                                                                                                                                                                                                                                                                                     
+   %b = select i1 %a, i64 32, i64 0                                                                                                                                                                                                                                                                                                         
+   %c = lshr i64 %input, %b                                                                                                                                                                                                                                                                                                                 
    %d = icmp ugt i64 %c, 65535                                                                                                                                                                                                                                                                                                              
    %e = select i1 %d, i64 16, i64 0
    %f = lshr i64 %c, %e                                                                                                                                                                                                                                                                                                                     
    %g = icmp ugt i64 %f, 255
    %h = select i1 %g, i64 8, i64 0
    %i = lshr i64 %f, %h
    %j = icmp ugt i64 %i, 15
    %k = select i1 %j, i64 4, i64 0
    %l = lshr i64 %i, %k
    %m = icmp ugt i64 %l, 3
    %n = select i1 %m, i64 2, i64 0
    %o = lshr i64 %l, %n
    %p = icmp ugt i64 %o, 1
    %q = zext i1 %p to i64
+   %r = or i64 %b, %e                                                                                                                                                                                                                                                                                                                       
    %s = or i64 %r, %h                                                                                                                                                                                                                                                                                                                       
    %t = or i64 %s, %k
    %u = or i64 %t, 2
    %v = add nuw nsw i64 %u, %n
    %w = or i64 %v, %q
    br label %log2.exit
```
So no major performance penalty. Definitely agree that we should switch histograms over to array maps in the future!

What's the reason that this algorithm deals with more bits now though?

luigirizzo:
> 
> What's the reason that this algorithm deals with more bits now though?

Do you mean what is the motivation ? Power of 2 is too coarse to evaluate e.g. changes that cause less than 2x performance differences. With the finer granularity we become able to see down to 3-5% without having to compromise on the dynamic range (e.g. we can differentiate 1000ns and 1100ns and still identify tails in the ms range)

danobi:
I mentioned in commit msg in https://github.com/iovisor/bpftrace/commit/006e68d0f2c0c57175ef8583fe0c42d25e4b43b5 that we were underutilizing the u64 by a lot (~54 bits unused IIUC). So yes, this uses more bits. Not sure if that addresses the question.

danobi:
I'd prefer if we merge this soon-ish rather than let it drag out and potentially eat merge conflicts. I can follow up with the request readability fix.

I'll push the button sometime this weekend if there are no more comments.

janca-ucdavis:
Hi - I just found out about this feature.  I've used partial-power log tables in Dtrace for many years to get fine grained histograms, and I see something here that seems relevant.

The buckets don't appear to be log-sized.  Rather, they're fixed width within each existing histogram.  So, you get (assuming 4 divisions per doubling):

128, 160, 192, 224, 256, 320...

This gives "sizes" that change abruptly at the border: four buckets that are 32-wide, then four that are 64-wide, and so on.  They're linear divisions of a log scale.

If they're log scale all the way down (as the rest of the histogram structure is) then you get boundaries like:

128, 152, 181, 215, 256, 304...

and each bucket boundary is the previous one * 2^(1/N), where N is the number of subdivisions of each doubling.  Bucket sizes go up smoothly throughout the histogram (24, 29, 34, 41, 49).  This gives a smooth distribution without aliasing and is what I expected to see from a log scale histogram.

Thoughts?

Thanks,

janet

PR #2836, opened 2023-11-18
Update tutorial_one_liners.md - Lesson 3 args.filename

Pitterling:
should be args->filename

 bpftrace -e 'tracepoint:syscalls:sys_enter_openat { printf("%s %s\n", comm, str(args.filename)); }'
stdin:1:64-73: ERROR: Can not access field 'filename' on type '(ctx) struct _tracepoint_syscalls_sys_enter_openat *'. Try dereferencing it first, or using '->'

### DIFF
--- a/docs/tutorial_one_liners.md
+++ b/docs/tutorial_one_liners.md
@@ -33,7 +33,7 @@ This prints a welcome message. Run it, then hit Ctrl-C to end.
 # Lesson 3. File Opens
 
 ```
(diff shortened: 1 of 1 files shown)


danobi:
Hi, thanks for the PR! Unfortunately, this was an intentional change. See https://github.com/iovisor/bpftrace/pull/2578 . I'm guessing you are reading documentation from `master` and using a different bpftrace version.

Pitterling:
@danobi  you're right .. using SLES15.5 shipped bpftrace version ...

but i think this one here slipped through - changing header title ?
https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#5---c-struct-navigation

PR #2840, opened 2023-11-21
New builtin for getting the number of map elements

viktormalik:
This adds a new builtin function called `size` which takes a single map argument and returns the number of elements stored inside the map (e.g. `size(@)`). Internally, this uses the `bpf_for_each_map_elem` BPF helper with a custom callback function doing the counting.

The builtin should be useful for eliminating the need for a separate map to do the counting in some use-cases.

One of the reasons for adding this builtin is that it requires support for subprograms (b/c the codegen creates a callback function) which is added as a part of this PR in the first two commits. These commits implement relocations of the BPF program and will be necessary, e.g., for user-defined functions (#2516). The `size` builtin is the simplest use-case of subprograms that I was able to create for the purpose of testing and simpler review of the relocations.

Note: I couldn't decide between names `size` and `mapsize`. In the end, I went with the former, but I recon that it doesn't explicitly state that it's for maps only, so I've no problem of changing it to the latter.

This is the remaining part of #2710 which can be closed afterwards.

##### Checklist


jordalgo:
> Note: I couldn't decide between names size and mapsize. In the end, I went with the former, but I recon that it doesn't explicitly state that it's for maps only, so I've no problem of changing it to the latter.

Are there plans on adding something similar for arrays or supporting arrays for `size`? It would be nicest if `size` just took a map or array or whatever and either returned the size or a not supported error.

src/ast/irbuilderbpf.cpp
```diff
@@ -541,6 +541,44 @@ void IRBuilderBPF::CreateMapDeleteElem(Value *ctx,
...
 
+void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
+                                        Map &map,
+                                        Value *callback,
+                                        Value *callback_ctx,
+                                        const location &loc)
+{
+  Value *map_ptr = CreateBpfPseudoCallId(map);
+
+  // long bpf_for_each_map_elem(struct bpf_map *map, void *callback_fn, void
```
jordalgo, line 552: comment leftovers?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1161,6 +1161,19 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateLifetimeEnd(buf);
     expr_ = nullptr;
   }
+  else if (call.func == "size")
+  {
+    auto &arg = *call.vargs->at(0);
```
jordalgo, line 1166: naming nit: maybe map_id?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3807,6 +3820,50 @@ void CodegenLLVM::createIncDec(Unop &unop)
...
+  };
+
+  FunctionType *callback_type = FunctionType::get(b_.getInt64Ty(), args, false);
+
+  Function *callback = Function::Create(callback_type,
+                                        Function::LinkageTypes::InternalLinkage,
+                                        "map_size_cb",
+                                        module_.get());
+
+  callback->addFnAttr(Attribute::NoInline);
```
jordalgo, line 3846: Just curious, why is this needed?

jordalgo:
LGTM!

ajor (commented):
Thanks again for resurrecting all these old commits!

I'll give this another review to go over the logic some more tomorrow, but will just start with my regular "this is C++" comments :P

ajor:
> > Note: I couldn't decide between names size and mapsize. In the end, I went with the former, but I recon that it doesn't explicitly state that it's for maps only, so I've no problem of changing it to the latter.
> 
> Are there plans on adding something similar for arrays or supporting arrays for `size`? It would be nicest if `size` just took a map or array or whatever and either returned the size or a not supported error.

Are you talking about C-style arrays in tracee code? I think `sizeof()` can deal with them, although it's not as ergonomic.

Regarding the syntax, I'm wondering if we should have certain functions operate as methods on maps, as is done in most other languages, e.g.:
```
@x.size()
@x.clear()
@x.zero()
@x.delete(1)
@x.contains(2) // this would be useful but we don't have it yet
```
(I'm not trying to suggest that this be done in this PR though)

jordalgo:
> Are you talking about C-style arrays in tracee code? I think sizeof() can deal with them, although it's not as ergonomic.

Yeah, it would be nice if `size` was polymorphic (that possible in bpftrace since we know the types up front?) so c-style arrays and other containers we might have in the future.

> we should have certain functions operate as methods on maps

Agreed this would be great.

viktormalik:
> I'll give this another review to go over the logic some more tomorrow, but will just start with my regular "this is C++" comments :P

These are great suggestions, thanks! And sorry for not checking this before posting the PR (again) :grimacing: 

>> Are you talking about C-style arrays in tracee code? I think sizeof() can deal with them, although it's not as ergonomic.

> Yeah, it would be nice if size was polymorphic (that possible in bpftrace since we know the types up front?) so c-style arrays and other containers we might have in the future.

Agreed, we could eventually apply it to arrays or to other containers (e.g. there's `BPF_MAP_TYPE_ARRAY` which we could use one day). In such a case, `size` makes more sense.

> Regarding the syntax, I'm wondering if we should have certain functions operate as methods on maps, as is done in most other languages

At the moment, I'm not a big fan of these. bpftrace is inspired by C (at least per README :)) and introducing "methods" would drive us further from that concept. In addition, it doesn't seem to be applicable outside map functions at this point. That being said, I'm not strictly against it, maybe we could open a separate RFC to discuss.

src/ast/passes/codegen_llvm.cpp
viktormalik: Not really, it's an argument expression which we cast to a `Map` on the following line. This is the naming used in the other map functions (e.g. `clear` and `zero`) so I followed the pattern.

src/ast/passes/codegen_llvm.cpp
viktormalik: Good question, I copied it from the original PR. I believe that the original point is to prevent Clang from inlining the function, although I'd hope that it never does that since it's used as a callback. Still, it's probably no harm leaving this attribute.

src/bpfprogram.cpp
```diff
@@ -35,16 +75,195 @@ const std::vector<uint8_t> &BpfProgram::getBTF()
...
+  const btf_ext_info_sec *text_funcinfo_sec = nullptr;
+  const btf_ext_info_sec *prog_funcinfo_sec = nullptr;
+
+  while (ptr < func_info_end)
+  {
+    auto *info_sec = reinterpret_cast<const struct btf_ext_info_sec *>(ptr);
+    auto *name = reinterpret_cast<const char *>(btfstr +
+                                                info_sec->sec_name_off);
+    if (text_offset_ > 0 && strcmp(name, ".text") == 0)
+      text_funcinfo_sec = reinterpret_cast<const btf_ext_info_sec *>(ptr);
```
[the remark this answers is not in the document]
viktormalik: > I'm finding all casting makes the code harder to read

Same here but I don't see many ways to avoid that. That's the price of manipulating raw data which is the way both libelf and libbpf work.

danobi:
At the risk of starting a naming discussion, but I kinda like `len()`:
* python popularized it
* it's one character shorter
* no visual ambiguity with `sizeof()`

viktormalik:
To me, `len()` always suggested that the underlying container is contiguous, which is not the case of BPF hashmaps, so I'm still a bit in favor of `size`. But it's no hard preference, so no problem of using `len` if other people like it.

jordalgo:
I have a slight preference for `len` for the reasons @danobi listed but mostly because it's more distinguishable from `sizeof`.

tests/codegen/call_size.cpp
```diff
@@ -0,0 +1,14 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, call_size)
+{
+  test("BEGIN { @x[1] = 1; } kprobe:f { printf(\"%d\", size(@x)); }", NAME);
```
ajor, line 9:
```suggestion
  test("BEGIN { @x[1] = 1; } kprobe:f { @y = size(@x)); }", NAME);
```
Not a big deal, but it makes the LLVM IR easier to read if the test cases are doing the bare minimum they need to. The printf adds a bunch of extra basic blocks to handle ring buffer statistics. If there's a way to test this without even assigning size to a map then that'd be even better!

If we can keep the tests cases smaller and more focused then they'll hopefully only be affected by smaller parts of codegen, so won't get updated so often and we'll have an easier time reading through the diffs for suspicious IR changes.

ajor (commented):
The code generally looks ok to me. I do have a slight concern about the lack of error handling in the BpfProgram relocation functions though. There's a lot of assumptions being made there, e.g. section names existing, offsets being valid.

1. Is there some standard which defines which sections LLVM will always produce for a BPF program? The tests should hopefully detect it if LLVM changes this, I'm just interested for myself and it would be useful to include a reference to it in the comments for future developers.
2. If we were to support the loading of BPF programs not written in bpftrace (e.g. a stack walker written in C/libbpf), would they have to pass through this same relocation code? Are they guaranteed to have the same format?

Taking a step back, could someone remind me why we need to do relocation ourselves in bpftrace instead of leaving it up to libbpf?

viktormalik:
> The code generally looks ok to me. I do have a slight concern about the lack of error handling in the BpfProgram relocation functions though. There's a lot of assumptions being made there, e.g. section names existing, offsets being valid.

That's a valid point, I'll have a look and add more checks and better error messages, where applicable.
 
> 1. Is there some standard which defines which sections LLVM will always produce for a BPF program? The tests should hopefully detect it if LLVM changes this, I'm just interested for myself and it would be useful to include a reference to it in the comments for future developers.

Not at the moment but there's one [in progress](https://www.ietf.org/archive/id/draft-thaler-bpf-elf-00.html). The parts relevant for us are not about to change in the final version, so I guess that we could also cite this one from the source (or a newer version, if exists - I'll have a look).

> 2. If we were to support the loading of BPF programs not written in bpftrace (e.g. a stack walker written in C/libbpf), would they have to pass through this same relocation code? Are they guaranteed to have the same format?

Before we do that, I'd like to delegate the entire loading and attachment to libbpf, thanks to which we will drop these manual relocations. Then, both the bytecode produced by bpftrace and loaded externally will be passed to libbpf, which will take care of the relocations. 

> Taking a step back, could someone remind me why we need to do relocation ourselves in bpftrace instead of leaving it up to libbpf?

The main reason is that leaving it up to libbpf requires a lot of plumbing, so this is a temporary minimal-effort solution to unblock other features, especially user-defined functions. Once we fully move to libbpf, the relocation code will be dropped. I believe that it's also the reason why the implementation doesn't bother much with error handling.

tests/codegen/call_size.cpp
viktormalik: This is a good suggestion.

> If there's a way to test this without even assigning size to a map then that'd be even better!

It seems that
```suggestion
  test("BEGIN { @x[1] = 1; } kprobe:f { $s = size(@x); }", NAME);
```
is ~10 instructions shorter, so I'll use that.

viktormalik:
We still have to resolve the `size` vs `len` discussion. We have 2 votes for `len` (@danobi, @jordalgo), 1 vote for `size`. @ajor wdyt?

docs/reference_guide.md
```diff
@@ -3401,6 +3401,7 @@ END
 - `print(value)` - Print a value
 - `clear(@x)` - Delete all keys from the map
 - `zero(@x)` - Set all map values to zero
+- `size(@x)` - Get the number of elements in the map
```
jordalgo, line 3404: nit for language consistency with the adoc: "Return the number of elements in the map."

And super nitpicky, I prefer the present tense "returns" so it reads like "what does 'size' do? It 'returns the number of elements in the map'." - but that would involve changing these other ones so I settle for "return".

src/ast/passes/codegen_llvm.cpp
viktormalik: > I would assume this as well. Is it easy to find out if this is the case or are there different clang build environments/settings that might cause this to happen in the future?

Considering how much LLVM likes to break things, I'm usually mostly concerned about future versions. But you're right that this shouldn't happen, kernel selftests also don't use the `noline` keyword. Removed.

docs/reference_guide.md
viktormalik: Agreed, let's use "return" as everywhere else.

jordalgo (approved):
Changes look good. Just need to settle the `len` 🆚  `size` battle

ajor:
@viktormalik Thanks for the explanations and comments! The code looks good to me.

I'll put my vote behind `len()` because I think it's clearer that it's returning the number of elements, rather than a size in bytes.

`size()` was the obvious one to me at first as it's what used by C++ and Java (+ other JVM languages), but after doing a bit of research it does seem like the trend for newer languages is to use `len()` to get the number of elements in a hash map. There are also a few languages using `count()` (which I'd argue makes the most sense for hash maps), but I think we can discard that option due to conflicts.

We do only have hash maps right now, but in the future we might also want to support array, stack and queue maps, as well as other non-map data structures and it would be good to have consistent naming across all of them.

viktormalik:
`len()` is is, then :-) Updated, ready for final reviews.

danobi:
I think elf sections are documented here: https://libbpf.readthedocs.io/en/latest/program_types.html

I'll do a pass on this over the weekend if you're waiting for me to take a look

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3807,6 +3820,49 @@ void CodegenLLVM::createIncDec(Unop &unop)
...
+  callback->setSection(".text");
+  debug_.createFunctionDebugInfo(*callback);
+
+  auto *bb = BasicBlock::Create(module_->getContext(), "", callback);
+  b_.SetInsertPoint(bb);
+
+  Value *ctx = callback->getArg(3);
+  auto len = b_.CreateBitCast(ctx, b_.getInt64Ty()->getPointerTo());
+  b_.CreateStore(
+      b_.CreateAdd(b_.CreateLoad(b_.getInt64Ty(), len), b_.getInt64(1)), len);
```
danobi, line 3857: This assumes that `len` is a uint64 right? Maybe more generic to get type from llvm? If not, perhaps add an `assert()` on that condition

tests/semantic_analyser.cpp
```diff
@@ -590,6 +591,14 @@ TEST(semantic_analyser, call_zero)
   test("kprobe:f { @x = count(); zero(@x) ? 0 : 1; }", 10);
 }
 
+TEST(semantic_analyser, call_len)
+{
+  test("kprobe:f { @x[0] = 0; len(@x); }", 0);
+  test("kprobe:f { @x[0] = 0; len(); }", 1);
+  test("kprobe:f { @x[0] = 0; len(@x, 1); }", 1);
+  test("kprobe:f { $x = 0; len($x); }", 1);
```
danobi, line 599: Add negative test for `len(@x[2])` ? I think I saw handling for map args in semantic analyser

danobi (commented):
lgtm! this infra will be quite useful.

just small nits

src/bpfprogram.cpp
```diff
@@ -42,9 +47,102 @@ void BpfProgram::assemble()
 
   code_ = bytecode_.at(name_);
 
+  relocateInsns();
   relocateMaps();
 }
 
+// Instruction relocations assume BPF ELF structure as generated by LLVM.
+// This is not standardized, yet, but there is an effort in place [1]. Still,
+// the structure is already assumed by many tools (libbpf in the first place)
```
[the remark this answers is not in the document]
viktormalik: Added, thanks.

tests/semantic_analyser.cpp
viktormalik: Added

src/ast/passes/codegen_llvm.cpp
viktormalik: How would you want to do either of those? The callback receives `void *ctx` so I don't see many ways to check or assert the pointed type.

danobi: Ah yeah you're right.

danobi (approved):
lgtm!

ajor (approved):

PR #2846, opened 2023-11-23
Revive Dockerfiles

### LINKED ISSUE #2756: Revive Dockerfiles

danobi:
Brings back Dockerfiles as executable documentation for building bpftrace on major distros.

This closes #2756.

##### Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,6 +15,7 @@
 - [Building bpftrace](#building-bpftrace)
   - [Ubuntu](#ubuntu)
   - [Fedora](#fedora)
+  - [Debian](#debian)
   - [Amazon Linux](#amazon-linux)
   - (*please add sections for other OSes)*
   - [Generic build](#generic-build-process)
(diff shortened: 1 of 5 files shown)


docker/Dockerfile.fedora
```diff
@@ -0,0 +1,30 @@
+# This Dockerfile is used to both document and test building bpftrace
+# on the latest fedora. We would use `fedora:latest`, but to make the
+# CI more deterministic, we pin to latest release.
```
viktormalik, line 3: I'd argue that using `fedora:latest` is what we want here b/c it will allow CI to catch build problems very early. Maybe we could go even further and use `fedora:rawhide` to try to build against the upcoming release.

docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,33 @@
+# This Dockerfile is used to both document and test building bpftrace
+# on the latest ubuntu. We would use `ubuntu:latest`, but to make the
+# CI more deterministic, we pin to latest release.
```
viktormalik, line 3: Same as with Fedora, what about using `ubuntu:latest` or `ubuntu:running`?

INSTALL.md
```diff
@@ -338,12 +272,15 @@ Use specific OS build sections listed earlier if available (Ubuntu).
...
 - Bison
+- Asciidoctor
 - LLVM & Clang 10.0+ development packages
 - LibElf
 - Binutils development package
 - Libcereal
 - Kernel requirements described earlier
 - Libpcap
+- Systemtap SDT headers
+- Zlib development package
```
viktormalik, line 283: Can we add `LibDw`, too?

viktormalik (commented):
Couple of suggestions, otherwise LGTM.

docker/Dockerfile.fedora
danobi: Rawhide sounds reasonable. Given unstable nature of living at tip (same with `:latest`), I'd probably wanna only run the distro CI post-merge (IOW on master). WDYT?

INSTALL.md
danobi: Ack

docker/Dockerfile.fedora
viktormalik: Yes, good idea, that'll notify us about a distro breakage and not break PRs at the same time.

viktormalik (approved):

PR #2860, opened 2023-11-30
Fix bpftrace -l when kfunc is not available

jordalgo:
A change to add fentry/fexit broke `bpftrace -l` because these new probe types weren't accounted for in `kernel_probe_list`.

This also removes the need for 'is_userspace_probe' with the addition of two new fields to ProbeItem.

Change: https://github.com/iovisor/bpftrace/commit/4151babaf9b616e8006762cfa853c01fde3408ae

##### Checklist

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -8,6 +8,7 @@
 #include <string>
 #include <sys/types.h>
 #include <unistd.h>
+#include <unordered_set>
 #include <vector>
 
 #include <cereal/access.hpp>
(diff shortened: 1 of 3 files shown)


danobi:
lgtm, but two thoughts:
* can we add a test to prevent future regressions?
* can the exclusion list be moved closer to PROBE_LIST so it's more obvious next time?

jordalgo:
> can the exclusion list be moved closer to PROBE_LIST so it's more obvious next time?

I was thinking the same thing. Perhaps just adding an additional field to `ProbeItem` e.g.
```
struct ProbeItem
{
  std::string name;
  std::string abbr;
  ProbeType type;
  bool in_kernel_list;
};
```

Then we can make the check in `kernel_probe_list` simple e.g. ` else if (p.in_kernel_list)`

jordalgo:
Went with a different solution that feels more explicit (maybe more code because I needed to add a unit test).

tests/types.cpp
```diff
@@ -0,0 +1,30 @@
...
+    bool found = false;
+    for (auto& s : PROBE_LIST)
+    {
+      if (s.name == p)
+      {
+        found = true;
+        break;
+      }
+    }
+    ASSERT_TRUE(found);
```
danobi, line 24: Probably a bunch of ways to get fancy here but I think the loop is obvious and simple

danobi (approved):

viktormalik (approved):
Just two nits, otherwise looks good. Great that we added a test!

src/types.h
```diff
@@ -509,6 +510,11 @@ const std::vector<ProbeItem> PROBE_LIST = {
   { "rawtracepoint", "rt", ProbeType::rawtracepoint },
 };
 
+const std::unordered_set<std::string> KERNEL_LIST_PROBES = {
+  "kprobe", "kfunc",      "software",     "hardware",
+  "iter",   "tracepoint", "rawtracepoint"
```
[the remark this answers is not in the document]
jordalgo: Good question. This actually inspired me to do some refactoring.

src/types.h
```diff
@@ -480,33 +481,30 @@ std::ostream &operator<<(std::ostream &os, ProbeType type);
 struct ProbeItem
 {
   std::string name;
-  std::string abbr;
+  std::unordered_set<std::string> aliases;
   ProbeType type;
+  bool show_in_kernel_list = false;
```
danobi, line 486: Would be good to add short comment describing semantics

danobi: Also, should we add `show_in_userspace_list` as well? So that we can get rid of `is_userspace_probe()` for similar reasons

danobi (approved):

danobi:
Since there's a bunch of fields now, maybe using designated initializers would make it easier to read

jordalgo:
> Since there's a bunch of fields now, maybe using designated initializers would make it easier to read

I used this for the two boolean fields that have defaults - seems to be a nice mix of readability and reduction in boilerplate.

viktormalik (approved):
Nice! Much more elegant solution than before, thanks!

PR #2864, opened 2023-12-02
[kfunc/kretfunc] fix arg and retval resolution

jordalgo:
In the case of functions that have arguments that are larger than the size of the registers (e.g. '__sys_bpf' whose second arg, 'bpfptr_t uattr', which has a size of 16 bytes) the 'retval' for kretfunc/fexit is not correct.

What's needed is to resolve the btf type size and use that (by dividing by the register size) to determine the index into the args array.

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -6,6 +6,7 @@
 #include "tracefs.h"
 #include "types.h"
 #include "utils.h"
+#include <cmath>
 #include <cstring>
 #include <fcntl.h>
 #include <iostream>
@@ -450,6 +451,8 @@ Struct BTF::resolve_args(const std::string &func, bool ret)
 
   Struct args(0, false);
   int j = 0;
+  int arg_idx = 0;
+  __u32 type_size = 0;
   for (; j < vlen; j++, p++)
   {
     const char *str = btf_str(func_id.btf, p->name_off);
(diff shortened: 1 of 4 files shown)


jordalgo:
Here we're assuming the register size is 64bits (which we were before) so not sure if we should account for 32bit architectures.

src/btf.cpp
```diff
@@ -450,30 +451,34 @@ Struct BTF::resolve_args(const std::string &func, bool ret)
...
     SizedType stype = get_stype(BTFId{ .btf = func_id.btf, .id = p->type });
-    stype.funcarg_idx = j;
+    stype.funcarg_idx = arg_idx;
     stype.is_funcarg = true;
     args.AddField(str, stype, args.size, std::nullopt, false);
     // kfunc (fentry/fexit) args are stored in a u64 array.
     // Note that it's ok to represent them by a struct as we will use GEP with
     // funcarg_idx to access them in codegen.
-    args.size += 8;
+    type_size = btf__resolve_size(func_id.btf, p->type);
```
anakryiko, line 469: I don't think kernel supports arg-by-value bigger than 16 bytes, don't know if bpftrace should warn/error out for that?

src/btf.cpp
```diff
@@ -450,30 +451,34 @@ Struct BTF::resolve_args(const std::string &func, bool ret)
...
+    args.size += type_size;
+    arg_idx += (type_size > 8) ? std::ceil(type_size / 8) : 1;
```
anakryiko, line 471: std::ceil(type_size / 8) should work for all cases, unless type_size == 0, the latter being an invalid case, no? so no need for ternary operator

anakryiko (commented):
fentry/fexit uses full 8-byte slot for any argument that is <=8 bytes, it should be the case even on 32-bit arches (but testing never hurts)

src/btf.cpp
jordalgo: Good point.

jordalgo:
Having a little trouble building/testing a 32bit kernel in my VM (not sure if that should be a blocker).

src/btf.cpp
```diff
@@ -450,30 +451,34 @@ Struct BTF::resolve_args(const std::string &func, bool ret)
...
+    stype.funcarg_idx = arg_idx;
     stype.is_funcarg = true;
     args.AddField(str, stype, args.size, std::nullopt, false);
     // kfunc (fentry/fexit) args are stored in a u64 array.
     // Note that it's ok to represent them by a struct as we will use GEP with
     // funcarg_idx to access them in codegen.
-    args.size += 8;
+    type_size = btf__resolve_size(func_id.btf, p->type);
+    args.size += type_size;
+    arg_idx += std::ceil((float)type_size / (float)8);
```
danobi, line 471: Is using the float to handle `type_size < 8`? If so, maybe more clear to assert that `type_size == 8 || type_size == 16`. Would be clear if kernel ever accidentally changes its ABI that a breakage occurred.

jordalgo: We already know that the type_size can be 4 (this is expected behavior)

danobi: Oh right, slot size is 8 or 16, not type

viktormalik (approved):

tests/runtime/probe
```diff
@@ -68,6 +68,14 @@ REQUIRES_FEATURE get_func_ip
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
+NAME kretfunc large args
+PROG kretfunc:__sys_bpf { if (args.cmd == 1111 && args.size == 2222 && (int64)retval == -7) { printf("SUCCESS %d\n", pid); exit(); }}
```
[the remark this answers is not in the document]
jordalgo: Good catch. Probably don't need pid. Will remove.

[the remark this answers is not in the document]
viktormalik: Sure, let's leave it, then.

Issue #2829, opened 2023-11-16
Redirect tool PRs to new bpftrace-tool-shed


danobi:
As discussed in office hours, there will be a new repository to hold community contributed bpftrace tools. Until it's created, this ticket holds some state we need to remember.

danobi:
For https://github.com/iovisor/bpftrace/pull/2593, @llaoj has asked to be notified when the new repo is created

danobi:
https://github.com/iovisor/bpftrace/pull/2437 is a good candidate

danobi:
https://github.com/iovisor/bpftrace/pull/1903 is a good candidate

danobi:
https://github.com/iovisor/bpftrace/pull/1903 is good as well

viktormalik:
another candidate: #2898

jordalgo:
I guess we might be waiting on our new github home to do this? (cc @ajor)

llaoj:
waiting for the new repo

danobi:
New repo has been created: https://github.com/bpftrace/user-tools

Please send PRs!

I'll bump the existing PRs as well

PR #2866, opened 2023-12-04
consolidate sublist of storage tools

makelinux:
to make list of all tools more organized


viktormalik (commented):
I'm not sure if we want to go this way. The list of tools is not that large and having them in alphabetical order makes sense in terms of quick search.

Once we have another repo for community contributed tools (see #2829), having them in categories will make more sense.

viktormalik:
Closing based on the above comments.

Issue #2854, opened 2023-11-28
Symbolication is broken when running a 32-bit userspace (and bpftrace binary) on a 64-bit kernel


jeez:
Symbolication is broken on platforms running a 64-bit kernel but 32-bit userspace. For example, this is a simple test case that can reproduce the issue: `$ bpftrace -e 'profile:hz:99 { @[kstack] = count(); }'`

I believe this is happening because bpftrace requests uint64_t addresses to the kernel here https://github.com/iovisor/bpftrace/blob/fb77dae0b0a3d16b9453264401ff778556028576/src/bpftrace.cpp#L1941 but then performs an implicitly conversion to uintptr_t here https://github.com/iovisor/bpftrace/blob/fb77dae0b0a3d16b9453264401ff778556028576/src/bpftrace.cpp#L2076 .

Since bpftrace is a 32-bit binary in this case, then uintptr_t is only 32-bit wide and the higher part of the 64-bit address provided by the kernel is dropped / ignored. This leads to a failure when bcc_symcache_resolve() is called. The implicit conversion should have been flagged by the compiler, but -Wconversion is currently not used by bpftrace: https://github.com/iovisor/bpftrace/blob/79c366d3d3eb93da18b1e75fbe150b0537626ca8/CMakeLists.txt#L70 .

The issue can be fixed if we change resolve_ksym() so it receives an `uint64_t addr` instead, but I wanted to check here first if there is an alternate preferred solution in case I'm missing something. I've only tested and observed this with kernel symbols so far, but I see uintptr_t being used in other locations throughout the project so this might be an issue for a few other use cases too.

Let me know what you think, please.

### `bpftrace --info` output
```
$ ./bpftrace --info
System
  OS: Linux 5.15.119-android13
  Arch: armv8l

Build
  version: v0.18.1
  LLVM: 14.0.6
  unsafe probe: yes
  bfd: no
  libdw (DWARF support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: no
  skboutput: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes
[...]

PR #2869, opened 2023-12-05
Fix symbolication on 32-bit userspace / 64-bit kernel

jeez:
Symbolication is broken on platforms running a 64-bit kernel and a 32-bit userspace. For example, this simple test case can reproduce the issue: $ bpftrace -e 'profile:hz:99 { @[kstack] = count(); }' .

This is happening because bpftrace requests uint64_t addresses to the kernel here but then performs an implicit conversion to uintptr_t here when resolving ksyms, usyms and uids. Fix the issue by changing the signature of these functions so they receive a uint64_t address instead.

Ideally we should uncomment add_compile_options("-Wconversion") in bpftrace/CMakeLists.txt to avoid running into simple issues like this in the future. This patch is one small step towards fixing the current conversion warnings we have.

[Issue 2854](https://github.com/iovisor/bpftrace/issues/2854)

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -2029,7 +2029,7 @@ std::string BPFtrace::get_stack(int64_t stackid,
   return stack.str();
 }
 
-std::string BPFtrace::resolve_uid(uintptr_t addr) const
+std::string BPFtrace::resolve_uid(uint64_t addr) const
 {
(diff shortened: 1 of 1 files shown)


danobi (approved):
lgtm, thanks!

wondering if we should add a `-m32` testprog and add a runtime test to cover this case.

PR #2872, opened 2023-12-07
tests: runtime: fix kprobe_offset_module* on aarch64

dtcccc:
For kprobe_offset_module:
On aarch64, +0x5 is not a correct offset. Since +0x4 is reserved by ftrace, the test should use +0x8, which is the same as ppc64.

For kprobe_offset_module_error:
nft_trans_alloc_gfp is not in vmlinux so usermode offset check is skipped. The kernel will return -EINVAL instead of -EILSEQ during arch_prepare_kprobe() on aarch64, so change the expected string.

BTW, I'm not familiar with other arch like ppc64, so I only keep x86_64 and aarch64 remained in kprobe_offset_module_error. Can anyone give some helps?

##### Checklist

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -153,18 +153,18 @@ NAME kprobe_offset_module
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x5 { printf("hit\n"); exit(); }'
 AFTER /usr/sbin/nft add table bpftrace
 EXPECT hit
-ARCH x86_64|aarch64
+ARCH x86_64
 TIMEOUT 5
 REQUIRES lsmod | grep '^nf_tables'
 REQUIRES /usr/sbin/nft --help
 CLEANUP /usr/sbin/nft delete table bpftrace
 
-# Local entry point to nft_trans_alloc_gfp is located at offset of 8 bytes in ppc64.
+# Local entry point to nft_trans_alloc_gfp is located at offset of 8 bytes in ppc64 and aarch64.
 NAME kprobe_offset_module
 RUN {{BPFTRACE}} -e 'kprobe:nft_trans_alloc_gfp+0x8 { printf("hit\n"); exit(); }'
 AFTER /usr/sbin/nft add table bpftrace
 EXPECT hit
-ARCH ppc64|ppc64le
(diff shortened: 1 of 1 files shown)


danobi (approved):
seems reasonable to me. though unfortunately I don't have any hardware to test this on.

I'm also now wondering why we do userspace offset checks for kprobe. And only for vmlinux. Seems that's the kernel's responsibility.

viktormalik:
> I'm also now wondering why we do userspace offset checks for kprobe. And only for vmlinux. Seems that's the kernel's responsibility.

Agreed. I've also had reports of people failing to run valid bpftrace programs with older versions of bpftrace due to these checks. We should revisit if this makes sense, let's discuss at the office hours.

viktormalik (commented):
> For kprobe_offset_module_error:
nft_trans_alloc_gfp is not in vmlinux so usermode offset check is skipped. The kernel will return -EINVAL instead of -EILSEQ during arch_prepare_kprobe() on aarch64, so change the expected string.

Do we know why kernel returns a different error code? IIUC, it should have nothing to do with the fact whether we do userspace checks or not.

Otherwise LGTM.

dtcccc:
> Do we know why kernel returns a different error code? IIUC, it should have nothing to do with the fact whether we do userspace checks or not.

This is the kernel behavior.

For generic offset error, arch_prepare_kprobe() will return different value:
x86_64: https://elixir.bootlin.com/linux/v6.6/source/arch/x86/kernel/kprobes/core.c#L749
aarch64: https://elixir.bootlin.com/linux/v6.6/source/arch/arm64/kernel/probes/kprobes.c#L99

viktormalik:
> > Do we know why kernel returns a different error code? IIUC, it should have nothing to do with the fact whether we do userspace checks or not.
> 
> This is the kernel behavior.
> 
> For generic offset error, arch_prepare_kprobe() will return different value: x86_64: https://elixir.bootlin.com/linux/v6.6/source/arch/x86/kernel/kprobes/core.c#L749 aarch64: https://elixir.bootlin.com/linux/v6.6/source/arch/arm64/kernel/probes/kprobes.c#L99

Thanks!

viktormalik (approved):

PR #2873, opened 2023-12-07
BTF: Handle array types

### LINKED ISSUE #2861: Segmentation fault when printing bpf_local_storage_elem
Repro:
```
$ sudo bpftrace -e 'BEGIN { $elem = (struct bpf_local_storage_elem*)0; print(*$elem); }'
WARNING: Addrspace is not set
Attaching 1 probe...
Segmentation fault
```

ajor:
This fixes a segmentation fault when attempting to print types containing arrays.

Unhandled BTF types are given Type::none and size 0. value_to_str previously attempted to read an int64_t into a 0-byte buffer allocated for this type, causing the SEGV.

This commit contains two fixes:
- Handle BTF arrays so they have Type::array and can be used & printed correctly in bpftrace
- Don't attempt to read an int64_t when printing a Type::none. Print nothing instead.

This script previously crashed bpftrace:
`BEGIN { $elem = (struct bpf_local_storage_data*)0; print(*$elem); }`

This type is defined as:
```
struct bpf_local_storage_data {
  struct bpf_local_storage_map *smap;
  u8 data[];
};
```

With this commit, we now successfully print the type:
`{ .smap = 0x0, .data = [] }`

Fixes #2861

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -412,6 +412,20 @@ SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
     stype = CreatePointer(
         get_stype(BTFId{ .btf = btf_id.btf, .id = t->type }, false));
   }
+  else if (btf_is_array(t))
+  {
+    auto *array = btf_array(t);
+    const auto &elem_type = get_stype(
+        BTFId{ .btf = btf_id.btf, .id = array->type });
+    if (elem_type.type == Type::integer && elem_type.GetSize() == 1)
+    {
+      stype = CreateString(array->nelems);
+    }
+    else
+    {
+      stype = CreateArray(array->nelems, elem_type);
+    }
+  }
 
   return stype;
 }
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -293,7 +293,10 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
     return res.str();
   }
   else
-    return std::to_string(read_data<int64_t>(value.data()) / div);
+  {
+    assert(type.IsNoneTy());
+    return "";
+  }
 }
 
 std::string Output::array_to_str(const std::vector<std::string> &elems) const
--- a/tests/data/data_source.c
+++ b/tests/data/data_source.c
@@ -41,6 +41,17 @@ struct Foo3 *func_3(int a, int *b, struct Foo1 *foo1)
   return 0;
 }
 
+struct Arrays
+{
+  int int_arr[4];
+  char char_arr[8];
+  void *ptr_arr[2];
+  int multi_dim[3][2];
+  int zero[0];
+  int flexible[];
+};
+struct Arrays arrays;
+
 struct task_struct
 {
   int pid;
--- a/tests/field_analyser.cpp
+++ b/tests/field_analyser.cpp
@@ -127,6 +127,65 @@ TEST_F(field_analyser_btf, btf_types)
   EXPECT_EQ(foo2_field.offset, 8);
 }
 
+TEST_F(field_analyser_btf, btf_arrays)
+{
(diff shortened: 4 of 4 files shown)


ajor:
Should I add a test case to "runtime/btf" using some kernel type containing an array (might not be stable) or do we have a way of generating custom BTF for test cases?

danobi:
We have facilities to generate both dwarf and btf from data_source.c: https://github.com/iovisor/bpftrace/blob/532d4e9d81e0c49033b19d2ae55ba3fd6866e347/tests/data/data_source.c

jordalgo (approved):
LGTM (minus the pending test addition)

src/output.cpp
```diff
@@ -294,7 +294,7 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
     return res.str();
   }
   else
-    return std::to_string(read_data<int64_t>(value.data()) / div);
+    return "";
```
danobi, line 297: might be good to assert type == none here

viktormalik, line 297: I think that this was originally intended to cover situations when we forget to add a handle for certain type. Maybe it makes sense to preserve that and return `""` only for `Type::None` and/or when `value` is empty?

ajor: The problem is that unsupported types won't necessarily always be 8-bytes. If they're smaller then the always-printing-something approach has the potential to be undefined behaviour by reading uninitialised memory. If they're larger then we'll be silently truncating values and people might not notice immediately.

IMO an obvious failure is better than a subtle one and crashes should be avoided at all costs.

The best solution would be for this function to use an exhaustive switch-statement instead of a chain of ifs, but I'll save that refactoring for another time.

viktormalik: Ok, makes sense.

ajor:
Parse BTF arrays of chars as Type::string, to match the Clang and DWARF parsers and fix broken tests.

src/btf.cpp
```diff
@@ -412,6 +412,20 @@ SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
     stype = CreatePointer(
         get_stype(BTFId{ .btf = btf_id.btf, .id = t->type }, false));
   }
+  else if (btf_is_array(t))
+  {
+    auto *array = btf_array(t);
+    const auto &elem_type = get_stype(
+        BTFId{ .btf = btf_id.btf, .id = array->type });
+    if (elem_type.type == Type::integer && elem_type.GetSize() == 1)
```
jordalgo, line 420: Help, I'm confused by this if check. If the element type is an integer with size 1 then create string? Is there a test for this case below and I'm just missing it?

ajor: A `char` is an integer with size 1 (`int8_t`, `uint8_t` etc. are just aliases for char types, so no difference). The "char_arr" member in the new test case covers this.


Doing the conversion in the type parser isn't the nicest way of handling this, but it does work without too many changes. Ideally I think I'd like to parse it as a char-array at this stage and then have an implicit casting system that could convert it into `Type::string` later when needed (e.g. when printing or comparing against a string).

That was too much work for me today though so I copied from our other type parsers:
https://github.com/iovisor/bpftrace/blob/78eebadde8a0000e65c06235ed4c93d5f940907c/src/clang_parser.cpp#L300-L303
https://github.com/iovisor/bpftrace/blob/78eebadde8a0000e65c06235ed4c93d5f940907c/src/dwarf_parser.cpp#L200-L203

jordalgo: Ahh thanks for the explanation. I see "type::string" in the tests now for char_arr

jordalgo (approved):

viktormalik (approved):
Thanks!

Issue #2867, opened 2023-12-05
bpftrace may crash handling anonymous structs or unions in DWARF data


rpigott:
Originally reported in #2863, bpftrace may crash if DWARF data contains references to anonymous structs or unions. bpftrace fails to create a valid name for these members here:

https://github.com/iovisor/bpftrace/blob/289f0c7e028b2317aeddbb0228d542f6250fe08c/src/dwarf_parser.cpp#L183C1-L185

See #2863 for more info.

danobi:
FWIW, when generating type definitions from BTF (similar to DWARF, but simpler), anon unions/structs get a generic anon name: https://github.com/libbpf/libbpf-rs/blob/4ef150551dbb25b18db7368cec79ca7ef9adfe0e/libbpf-cargo/src/gen/btf.rs#L72-L82

That seems like it should work? User probably will never directly reference anon record type. So as long as bpftrace doesn't barf, my feeling is that it should be ok.

@viktormalik would know better tho

danobi:
More details here: https://github.com/libbpf/libbpf-rs/commit/fa5b28cb13e7dd94b5f01972dd0b2a4d2ae7750f

viktormalik:
> FWIW, when libbpf-rs generates type definitions from BTF (similar to DWARF, but simpler), anon unions/structs get a generic anon name: https://github.com/libbpf/libbpf-rs/blob/4ef150551dbb25b18db7368cec79ca7ef9adfe0e/libbpf-cargo/src/gen/btf.rs#L72-L82
> 
> That seems like it should work? I can't think of any valid reasons a user would directly reference an anon record type. So as long as bpftrace doesn't barf, my feeling is that it should be ok.

Agreed. As long as the names are unique (so that different anonymous types don't override each other in the struct manager), I believe that this approach should work.

martinetd:
I was about to open another bug for something similar (although there doesn't seem to be anonymous in the field I was accessing), but perhaps the situation on this has also improved with #5095 ?

bpftrace 0.25.1 crashes with this minimal build of swupdate
```
$ cd /tmp
$ git clone http://github.com/sbabic/swupdate
$ cd swupdate && make no_ssl_defconfig && make -j32
# bpftrace  -v -e 'uprobe:/tmp/swupdate/swupdate_unstripped:copyimage {$img = args->img; if ($img) { printf("copyimage %s\n", $img->fname ); }}'
terminate called after throwing an instance of 'std::logic_error'
[...]

PR #2874, opened 2023-12-08
WIP: Index anonymous structs

rpigott:
In #2863 I tried to use a fixed name for anonymous structs. That was no good so it was suggested to generate unique names for them. This does basically fix my original use case, but it doesn't seem to have the intended effect. I can still reproduce the crash in some of the test cases though I'm not sure why yet.

Also, some types that I don't think should be anonymous are identified as such:

```
$ sudo ./src/bpftrace -e 'uprobe:libc:getaddrinfo {printf("%d\n", args.hints->ai_protocol);}'
stdin:1:41-64: ERROR: Struct/union of type 'struct __anon_1' does not contain a field named 'ai_protocol'
uprobe:libc:getaddrinfo {printf("%d\n", args.hints->ai_protocol);}
```
but hints should have type `struct addrinfo *` [1].

Intended for #2867

[1] https://debuginfod.archlinux.org/buildid/8bfe03f6bf9b6a6e2591babd0bbc266837d8f658/debuginfo

### DIFF
--- a/src/dwarf_parser.cpp
+++ b/src/dwarf_parser.cpp
@@ -17,7 +17,7 @@ struct FuncInfo
 };
 
 Dwarf::Dwarf(BPFtrace *bpftrace, const std::string &file_path)
-    : bpftrace_(bpftrace), file_path_(file_path)
+    : bpftrace_(bpftrace), file_path_(file_path), anon_idx(1)
 {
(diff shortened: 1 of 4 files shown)


danobi:
closing due to inactivity

PR #2875, opened 2023-12-10
biosnoop.bt: use block tracepoints

makelinux:
[since functions blk_account_io_start and blk_account_io_done became static](https://lore.kernel.org/all/20211117061404.331732-9-hch@lst.de/) and can't be used as kprobes.

Using major and minor because unfortunately disk_name is not exported to tracepoints.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,10 @@ and this project adheres to
   - [#2849](https://github.com/iovisor/bpftrace/pull/2849)
 - Add milliseconds to timestamp and align numbers to the right in `killsnoop.bt`
   - [#2936](https://github.com/iovisor/bpftrace/pull/2936)
+- Update biosnoop.bt for kernels >=5.17 and >=4.18
+  - [#2875](https://github.com/iovisor/bpftrace/pull/2875)
+- biosnoop.bt: replace disk name field with major:minor.
+  - [#2875](https://github.com/iovisor/bpftrace/pull/2875)
 
 ## [0.19.0] 2023-09-19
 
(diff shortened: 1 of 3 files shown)


danobi (approved):
seems reasonable to me

danobi:
Ah ok, so I was getting confused about `dev_t` cuz apparently kernel `dev_t` is different than the one userspace sees: https://lwn.net/Articles/50703/ . So that all looks good.

Just one question though: when I run biosnoop during `sudo pacman -Syu`, I see a lot of:

```
10497         259:1  pacman           270981       0
10499         259:1  pacman           270981       0
10499         259:1  pacman           270981       0
10500         259:1  pacman           270981       0
10502         259:1  pacman           270981       0
10502         259:1  pacman           270981       0
10503         259:1  pacman           270981       0
```

But my `lsblk` shows:

```
nvme0n1     259:0    0 953.9G  0 disk
├─nvme0n1p1 259:1    0     1G  0 part  /boot
├─nvme0n1p2 259:2    0    32G  0 part  [SWAP]
└─nvme0n1p3 259:3    0 920.9G  0 part  /
```

Doesn't seem correct that pacman writes/reads from `/boot`. It should be from `/`, right? Any ideas?

makelinux:
Good catch! Fixed.

danobi:
Ha, `&&` vs `&` -- didn't spot that.

With fix, it looks like all activity is going to minor number 0. Is that expected? I'm not too familiar with block layer semantics for this case

makelinux:
> With fix, it looks like all activity is going to minor number 0. Is that expected? I'm not too familiar with block layer semantics for this case

The same major or another?

danobi:
The same major as before (correct one I think)

makelinux:
I don't know.

Issue #2877, opened 2023-12-13
execsnoop.bt script failing with ERROR: Unknown struct/union: 'struct task_struct'


gkamathe:
### What reproduces the bug? Provide code if possible.

A recent pull request #2876 added PPID to execsnoop.bt script, currently its failing with following error

```
$ bpftrace tools/execsnoop.bt
tools/execsnoop.bt:26:10-33: ERROR: Cannot resolve unknown type "struct task_struct"
    $task = (struct task_struct *)curtask;
            ~~~~~~~~~~~~~~~~~~~~~~~
tools/execsnoop.bt:27:49-67: ERROR: Unknown struct/union: 'struct task_struct'
    printf("%-10u %-7d %-7d ", elapsed / 1e6, pid, $task->real_parent->pid);
                                                   ~~~~~~~~~~~~~~~~~~
$ 
```

gkamathe:
Additional information
```

$ ./src/bpftrace --version
bpftrace v0.19.0-87-gd286
$ 
$ cat /etc/fedora-release 
Fedora release 37 (Thirty Seven)
$ 
$ ./src/bpftrace ../tools/execsnoop.bt 
../tools/execsnoop.bt:25:10-33: ERROR: Cannot resolve unknown type "struct task_struct"
    $task = (struct task_struct *)curtask;
            ~~~~~~~~~~~~~~~~~~~~~~~
../tools/execsnoop.bt:26:49-67: ERROR: Unknown struct/union: 'struct task_struct'
    printf("%-10u %-7d %-7d ", elapsed / 1e6, pid, $task->real_parent->pid);
                                                   ~~~~~~~~~~~~~~~~~~
$
```

PR #2878, opened 2023-12-13
tools: fixing execsnoop.bt failure due to PPID addition

gkamathe:
Fixing issue #2877 

##### Checklist

### DIFF
--- a/tools/execsnoop.bt
+++ b/tools/execsnoop.bt
@@ -15,7 +15,9 @@
  * 11-Sep-2018	   "     "	Switched to use join().
  */
 
+#ifndef BPFTRACE_HAVE_BTF
 #include <linux/sched.h>
+#endif
 
 BEGIN
 {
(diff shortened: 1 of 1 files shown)


tools/execsnoop.bt
```diff
@@ -15,6 +15,8 @@
  * 11-Sep-2018	   "     "	Switched to use join().
  */
 
+#include <linux/sched.h>
```
viktormalik, line 18: As in other tools, this is just needed when BTF is not available:
```suggestion
#ifndef BPFTRACE_HAVE_BTF
#include <linux/sched.h>
#endif
```

viktormalik (changes requested):
Thanks for finding this! We just need to make sure that the header is only included if BTF is unavailable.

viktormalik (approved):
Thanks!

PR #2881, opened 2023-12-14
Removed unused ptr_to_u64() function

shunghsiyu:
Hi, this is a minor patch that removes the unused `ptr_to_u64()` static function. As far as I can tell while it was added in commit 5ab53fad, it was not used since its introduction (probably because C++ cast was used instead).

This PR removes the unused `ptr_to_u64()` to silence the compiler warning.


viktormalik (approved):
LGTM, just needed to restart a CI job.

Thanks!

PR #2885, opened 2023-12-20
Add Config class

jordalgo:
This is in preparation for a new config syntax:
https://github.com/iovisor/bpftrace/pull/2815

This consolidates configuration into a single class/object and supports a hierarchy for setting config variables e.g. setting these variables via environment variables will take precedence over setting them from within a script.

##### Checklist

### DIFF
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -16,6 +16,7 @@
 #include "bpfprogram.h"
 #include "btf.h"
 #include "child.h"
+#include "config.h"
 #include "dwarf_parser.h"
 #include "map.h"
 #include "mapmanager.h"
@@ -94,11 +95,13 @@ class BPFtrace
 {
 public:
   BPFtrace(std::unique_ptr<Output> o = std::make_unique<TextOutput>(std::cout),
-           BPFnofeature no_feature = BPFnofeature())
+           BPFnofeature no_feature = BPFnofeature(),
+           Config config = Config())
       : out_(std::move(o)),
         feature_(std::make_unique<BPFfeature>(no_feature)),
         probe_matcher_(std::make_unique<ProbeMatcher>(this)),
-        ncpus_(get_possible_cpus().size())
+        ncpus_(get_possible_cpus().size()),
+        config_(config)
   {
   }
   virtual ~BPFtrace();
@@ -150,7 +153,6 @@ public:
   std::string get_param(size_t index, bool is_str) const;
   size_t num_params() const;
   void request_finalize();
-  bool is_aslr_enabled(int pid);
   std::string get_string_literal(const ast::Expression *expr) const;
   std::optional<int64_t> get_int_literal(const ast::Expression *expr) const;
   std::optional<std::string> get_watchpoint_binary_path() const;
@@ -190,30 +192,11 @@ public:
   std::unique_ptr<Output> out_;
   std::unique_ptr<BPFfeature> feature_;
 
-  uint64_t strlen_ = 64;
-  const char *str_trunc_trailer_ = "..";
-  uint64_t mapmax_ = 4096;
-  size_t cat_bytes_max_ = 10240;
-  uint64_t max_probes_ = 512;
-  uint64_t max_programs_ = 512;
-  uint64_t log_size_ = 1000000;
-  uint64_t perf_rb_pages_ = 64;
-  uint64_t max_type_res_iterations = 0;
-  bool demangle_cpp_symbols_ = true;
   bool resolve_user_symbols_ = true;
-  bool debug_output_ = false;
-  enum class UserSymbolCacheType
-  {
-    per_pid,
-    per_program,
-    none,
-  } user_symbol_cache_type_;
   bool safe_mode_ = true;
   bool has_usdt_ = false;
   bool usdt_file_activation_ = false;
   int helper_check_level_ = 0;
-  uint64_t ast_max_nodes_ = 0; // Maximum AST nodes allowed for fuzzing
-  std::optional<StackMode> stack_mode_;
   std::optional<struct timespec> boottime_;
   std::optional<struct timespec> delta_taitime_;
   static constexpr uint32_t rb_loss_cnt_key_ = 0;
@@ -236,6 +219,7 @@ public:
   }
   int ncpus_;
   int online_cpus_;
+  Config config_;
 
 private:
   int run_special_probe(std::string name,
--- a/src/config.h
+++ b/src/config.h
@@ -1,0 +1,194 @@
+#pragma once
+
+#include <cstdint>
+#include <iostream>
+#include <map>
+#include <variant>
+
+#include "types.h"
+
+namespace bpftrace {
+
+// This is used to determine which source
+// takes precedence when a config key is set
+enum class ConfigSource
+{
+  // the default config value
+  default_,
+  // config set via environment variable
+  env_var,
+};
+
+enum class ConfigKeyBool
+{
+  debug_output,
+  no_cpp_demangle,
+  verify_llvm_ir,
+};
+
+enum class ConfigKeyInt
+{
+  ast_max_nodes,
+  cat_bytes_max,
+  log_size,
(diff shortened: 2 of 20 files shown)


src/config.h
```diff
@@ -0,0 +1,134 @@
...
+
+#include "types.h"
+
+namespace bpftrace {
+
+enum class ConfigSource
+{
+  default_,
+  env_var,
+};
```
danobi, line 16: Would be good to leave some doc comments on each variant

src/config.h
```diff
@@ -0,0 +1,134 @@
...
+  stack_mode,
+  strlen,
+  str_trunc_trailer,
+  user_symbol_cache_type,
+  verify_llvm_ir,
+};
+
+struct ConfigValue
+{
+  ConfigSource source;
```
danobi, line 39: Maybe:

```suggestion
  ConfigSource source = ConfigSource::default_;
```

and then in the default map below you can do:

```
    { ConfigKey::ast_max_nodes, { .value = (uint64_t)0 } },
```

src/config.h
```diff
@@ -0,0 +1,134 @@
...
+  { "bpftrace", StackMode::bpftrace },
+  { "perf", StackMode::perf },
+  { "raw", StackMode::raw },
+};
+
+class Config
+{
+public:
+  explicit Config(bool has_cmd = false, bool bt_verbose = false);
+  ~Config(){};
```
danobi, line 54: Might be interpreting guidelines wrong, but I think it's ok to omit this right?

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c30-define-a-destructor-if-a-class-needs-an-explicit-action-at-object-destruction

src/config.h
```diff
@@ -0,0 +1,134 @@
...
+    try
+    {
+      std::get<T>(it->second.value);
+    }
+    catch (std::bad_variant_access const &ex)
+    {
+      throw std::runtime_error("Type mismatch for config key");
+    }
+
+    it->second.value = (T)val;
```
danobi, line 78: Why is the cast necessary? Looks like it should be a noop cuz of `T val` above

tests/config.cpp
```diff
@@ -0,0 +1,55 @@
...
+  EXPECT_THAT(
+      [&]() {
+        config.set<std::string>(ConfigKey::cat_bytes_max,
+                                "str",
+                                ConfigSource::env_var);
+      },
+      testing::Throws<std::runtime_error>());
+}
+
+TEST(Config, get)
```
danobi, line 30: Maybe also test a get() with wrong template type?

danobi (commented):
Various small comments; lgtm otherwise

src/bpftrace.h
danobi, line 1: Nice, this is a good cleanup!

src/config.h
jordalgo: I like it!

src/config.h
jordalgo: Good catch. Will remove

danobi (commented):
Thinking some more, is there a way to omit the template argument for the `get<..>()`? Cuz using the wrong variant type results in an exception, and it won't be completely obvious during code review if, for example, someone does `get<int64_t>(..)` expecting it to convert from the uint64_t or something.

Seems like in theory we should have enough information with the enum class to deduce the type (if we annotate the knob appropriately).

Wondering if there's some `T::type` technique that's possible here. Something like https://en.cppreference.com/w/cpp/types/is_array . Not sure if that makes sense. But at the same time, I dont' really wanna get too deep into template magic, so maybe it's ok like this for now.

jordalgo:
@danobi I think the caller would still have to specify the type even if we did something like have separate get methods for different types based on the config key. What if we just added a try/catch in the `get` with a clearer message e.g.
```
template <typename T>
  T get(ConfigKey key) const
  {
    auto it = config_map_.find(key);
    if (it == config_map_.end())
    {
      throw std::runtime_error("Unsupported config key");
    }
    try
    {
      return std::get<T>(it->second.value);
    }
    catch (std::bad_variant_access const &ex)
    {
      throw std::runtime_error("Type mismatch for config key");
    }
  }
```

danobi:
Yeah better error message is good for now. Pls consider the template comment as non-blocking. I'll try to spend some time this week playing with it

jordalgo:
@danobi Another option would be to do something like this:
```
bool set(ConfigKeyBool key, bool val, ConfigSource source)
  {
    return set<bool>(key, val, source);
  }
  
  bool set(ConfigKeyInt key, uint64_t val, ConfigSource source)
  {
    return set<uint64_t>(key, val, source);
  }
  
  bool get(ConfigKeyBool key) const
  {
    return get<bool>(key);
  }
  
  uint64_t get(ConfigKeyInt key) const
  {
    return get<uint64_t>(key);
  }
  
  bool set(std::variant<ConfigKeyBool, ConfigKeyInt> key, T val, ConfigSource source)
  {
    auto it = config_map_.find(key);
    if (it == config_map_.end())
    {
      throw std::runtime_error("Unsupported config key");
    }
    if (!can_set(it->second.source, source))
    {
      return false;
    }

    it->second.value = val;
    it->second.source = source;
    return true;
  }
  
  template <typename T>
  T get(std::variant<ConfigKeyBool, ConfigKeyInt> key) const
  {
    auto it = config_map_.find(key);
    if (it == config_map_.end())
    {
      throw std::runtime_error("Unsupported config key");
    }
    try
    {
      return std::get<T>(it->second.value);
    }
    catch (std::bad_variant_access const &ex)
    {
      throw std::runtime_error("Type mismatch for config key");
    }
  }
```

src/utils.h
```diff
@@ -157,8 +160,11 @@ static std::vector<std::string> COMPILE_TIME_FUNCS = { "cgroupid" };
 
 static std::vector<std::string> UPROBE_LANGS = { "cpp" };
 
+bool get_uint64_env_var(const ::std::string &str,
+                        const std::function<void(uint64_t)> &cb);
 bool get_uint64_env_var(const ::std::string &str, uint64_t &dest);
```
viktormalik, line 165: Do we still need this variant?

viktormalik (commented):
I love this, great work! Just one question below.

As for the discussion on the form of getters/setters, no strong opinions here. Both proposed variants require the user to know the type of the config option and I don't see much difference between `bpftrace.config.get<bool>(ConfigKey::X)` and `bpftrace.config.get(ConfigKeyBool::X)`. Perhaps the latter is slightly better b/c it's easier to determine the config type by just looking at the enums.

jordalgo:
>  I don't see much difference between bpftrace.config.get<bool>(ConfigKey::X) and bpftrace.config.get(ConfigKeyBool::X). Perhaps the latter is slightly better b/c it's easier to determine the config type by just looking at the enums.

You also get some compile time checks with the second one in that you can't do `uint64_t myvar = bpftrace.config.get(ConfigKeyBool::X)`. I can also add a unit test to exhaustively check all the ConfigKeys to make sure they don't throw in getting or setting. The only slightly annoying thing with the second options are the 3 ConfigKey types that only have a single enum e.g. ConfigKeyString, ConfigKeyStackMode, ConfigKeyUserSymbolCacheType

danobi:
Yeah variant 2 looks a bit nicer w.r.t. compile time checks. Single variant enums are a little verbose, yeah, but maybe one day it'll inspire someone to unify it all

src/utils.h
jordalgo: Good call. Will remove.

jordalgo:
Ok, implemented version 2 with a few tweaks

viktormalik (approved):
Nice work! 

The single-value enums are a bit unfortunate but I think that we can live with them for now.

danobi (approved):
thanks for iterating! lgtm

viktormalik:
Looks like we can merge.

jordalgo:
Thanks for the reviews @danobi & @viktormalik!

PR #2891, opened 2023-12-26
biostacks.bt: use tracepoints instead of kprobes

makelinux:
because since v5.17-rc1~116^2~83 function blk_account_io_start became static and can't be used as a kprobe.


chenhengqi:
The problem is that the block_io_start tracepoint is relative new and not widely available.

danobi (commented):
Seems like there's already prior art in having a -kp variant of the tool.
So probably fine to continue.

CI failure looks legit (and is what @chenhengqi mentioned). I think 
ci.yml needs to be updated to ignore testing that tool.

viktormalik (commented):
I'd prefer moving `tools/biostacks-kp.bt` to `tools/old/biostacks.bt` which is the way we currently maintain multiple versions of tools for different kernels.

That would also allow to easily resolve the CI issue by adding the tool into `TOOLS_TEST_OLDVERSION` for relevant CI jobs.

viktormalik (approved):
Ah, I forgot to update the manpage, good catch!

makelinux:
@viktormalik, consider updating tools/biostacks_example.txt as well.

viktormalik:
@makelinux good point, could you perhaps update it as a part of this PR?

PR #2893, opened 2023-12-30
docker/debian: Install deps xxd, libgtest-dev and libgmock-dev

Rtoax:
On Debian 12, i just install all deps in Dockerfile.debian, however:

```bash
    $ cmake -DCMAKE_INSTALL_PREFIX=/usr/ -DCMAKE_BUILD_TYPE=Debug \
    	-DBUILD_TESTING=ON -DVENDOR_GTEST=OFF ..
    ...
    -- Found LibXml2: /usr/lib/aarch64-linux-gnu/libxml2.so (found version "2.9.14")
    -- Found LLVM 14.0.6: /usr/lib/llvm-14/lib/cmake/llvm
    -- Disabled codegen test for LLVM != 12
    CMake Error at tests/data/CMakeLists.txt:4 (find_program):
      Could not find XXD using the following names: xxd
```

### DIFF
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -13,11 +13,14 @@ RUN apt-get update && apt-get install -y \
     clang \
     cmake \
     flex \
+    xxd \
     libbpf-dev \
     libbpfcc-dev \
     libcereal-dev \
     libdw-dev \
     libelf-dev \
+    libgmock-dev \
+    libgtest-dev \
     libiberty-dev \
     libpcap-dev \
     llvm-dev \


danobi (commented):
The reason those weren't included was cuz below we have `-DBUILD_TESTING=OFF`. Cuz in my mind the dockerfiles are for building bpftrace as a user, not as a developer. For development, I was thinking folks should refer to `flake.nix`.

But if this is more convenient than flake.nix, we can convert all the dockerfiles to build tests well. WDYT?

viktormalik (commented):
Agreed with @danobi here. Keeping the dependencies is a maintenance cost so we should try to avoid duplicates wherever possible. The purpose of the Dockerfiles is an early detection of problems with bpftrace installation on various distros. The recommended development environment is Nix so I believe that we should only keep the list of dev dependencies in `flake.nix`.

PR #2890, opened 2023-12-26
hist: PR followups

danobi:
Fixes up some follow-on comments from [0].

Also fixes a compile time warning:

```
[25/85] Building CXX object src/CMakeFiles/runtime.dir/output.cpp.o In file included from /usr/include/c++/13.2.1/cassert:44,
                 from /home/dxu/dev/bpftrace/src/log.h:3,
                 from /home/dxu/dev/bpftrace/src/output.cpp:3:
/home/dxu/dev/bpftrace/src/output.cpp: In static member function ‘static std::string bpftrace::TextOutput::hist_index_label(int, int)’: /home/dxu/dev/bpftrace/src/output.cpp:53:16: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const uint32_t’ {aka ‘const unsigned int’} [-Wsign-compare]
   53 |   assert(index >= n); // Smaller indexes are converted directly.
      |          ~~~~~~^~~~
```

[0]: https://github.com/iovisor/bpftrace/pull/2831#discussion_r1417767030

##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #2894, opened 2023-12-30
Fix dangerous compile warning: Compare signed and unsigned int

Rtoax:
```bash
    $ gcc --version
    gcc (Debian 12.2.0-14) 12.2.0

    $ cmake -DCMAKE_INSTALL_PREFIX=/usr/ -DCMAKE_BUILD_TYPE=Debug \
	-DBUILD_TESTING=ON -DVENDOR_GTEST=OFF ..
    $ make
    ...
    [ 75%] Building CXX object src/CMakeFiles/runtime.dir/output.cpp.o
    In file included from /usr/include/c++/12/cassert:44,
                     from /home/rongtao/Git/bpftrace/src/log.h:3,
                     from /home/rongtao/Git/bpftrace/src/output.cpp:3:
    /home/rongtao/Git/bpftrace/src/output.cpp: In static member function
    'static std::string bpftrace::TextOutput::hist_index_label(int, int)':
    /home/rongtao/Git/bpftrace/src/output.cpp:53:16: warning: comparison of
    integer expressions of different signedness: 'int' and 'const uint32_t'
    {aka 'const unsigned int'} [-Wsign-compare]
       53 |   assert(index >= n); // Smaller indexes are converted directly.
          |          ~~~~~~^~~~
```

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -50,7 +50,7 @@ std::ostream& operator<<(std::ostream& out, MessageType type) {
 std::string TextOutput::hist_index_label(int index, int k)
 {
   const uint32_t n = (1 << k), interval = index & (n - 1);
-  assert(index >= n); // Smaller indexes are converted directly.
+  assert(index >= (int)n); // Smaller indexes are converted directly.
   uint32_t power = (index >> k) - 1;
   // Choose the suffix for the largest power of 2^10
   const uint32_t decade = power / 10;


danobi (commented):
I think `n` has a higher chance of overflowing. I have an alternative fix in #2890 where I cast `index`, which is guaranteed to be small and non-negative

ajor:
Closing in favour of #2890. Appreciate the fix though!

PR #2896, opened 2023-12-31
aot: Finish up write side

danobi:
This PR finishes up write side support as described in https://dxuuu.xyz/aot-bpftrace.html .

Basically start cloning the shim and injecting a `.btaot` section into the new binary.
The read side will look in that section later. Next PR.

There's lots and lots still broken about AOT. Lots of language features don't work.
Also it looks like `bpftrace-aotrt` (the shim) is dynamically linked against a crazy
number of libraries again. I'll have to pull that apart and try to enforce it somehow.

Runtime tests were augmented a long time ago to reuse existing test cases against
AOT. Will fix those later as well.

##### Checklist

### DIFF
--- a/src/aot/aot.cpp
+++ b/src/aot/aot.cpp
@@ -1,11 +1,14 @@
 #include "aot.h"
 
 #include <cstdint>
+#include <cstdlib>
 #include <cstring>
 #include <exception>
 #include <fcntl.h>
 #include <fstream>
 #include <istream>
+#include <optional>
+#include <ostream>
 #include <streambuf>
 #include <sys/mman.h>
 #include <sys/stat.h>
@@ -16,10 +19,14 @@
 #include <cereal/types/unordered_map.hpp>
 #include <cereal/types/vector.hpp>
 
+#include "filesystem.h"
 #include "log.h"
 #include "utils.h"
 
+#define AOT_ELF_SECTION ".btaot"
 static constexpr auto AOT_MAGIC = 0xA07;
+static constexpr auto AOT_SHIM_NAME = "bpftrace-aotrt";
+static constexpr auto AOT_SECDATA_TEMPFILE = ".temp_btaot";
 
 // AOT payload will have this header at the beginning. We don't worry about
 // versioning the header b/c we enforce that an AOT compiled script may only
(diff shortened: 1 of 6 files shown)


tests/utils.cpp
```diff
@@ -281,6 +274,83 @@ TEST(utils, sanitiseBPFProgramName)
...
+                     std::function<void()> fn)
+{
+  EXPECT_EQ(::setenv(key.c_str(), val.c_str(), 1), 0);
+  try
+  {
+    fn();
+  }
+  catch (const std::exception &ex)
+  {
+    EXPECT_EQ(::unsetenv(key.c_str()), 0);
```
jordalgo, line 290: Why not unset even on success (perhaps just add this to the top in a scope exit?)

src/aot/aot.cpp
```diff
@@ -99,6 +99,72 @@ int load_bytecode(BPFtrace &bpftrace, uint8_t *ptr, size_t len)
   return 0;
 }
 
+// Generates contents of the .BTAOT ELF section.
+std::optional<std::vector<uint8_t>> generate_section(
```
jordalgo, line 110: naming nit: maybe 'generate_BTAOT_section' (then you could maybe ditch the comment)

src/aot/aot.cpp
```diff
@@ -165,6 +172,59 @@ std::optional<std::vector<uint8_t>> generate_section(
...
+
+  // Resolve objcopy binary to full path
+  auto objcopy_full = find_in_path(objcopy);
+  if (!objcopy_full)
+  {
+    LOG(ERROR) << "Failed to find " << objcopy << " in $PATH";
+    goto out;
+  }
+
+  n = snprintf(cmd,
```
jordalgo, line 204: nit: n is not a great variable name and doesn't seem to get re-used so why not inline this into the `if` check?

src/aot/aot.cpp
```diff
@@ -165,6 +172,59 @@ std::optional<std::vector<uint8_t>> generate_section(
   return out;
 }
 
+// Clones the shim to final destination while also injecting
+// the custom .btaot section.
+int build_binary(const std_filesystem::path &shim,
```
jordalgo, line 177: It feel like this could just return a bool to indicate success or failure as the only useful err is from `std::system(cmd)`, which could just be logged with `LOG(ERROR)` but not a big deal either way.

jordalgo:
I'm not super familiar with the AOT code but overall these changes look good. objcopy is also ubiquitous enough IMHO that I don't think it's a big issue to shell out to it.

tests/utils.cpp
danobi: Yeah, we're missing SCOPE_EXIT style macros in bpftrace. Should probably add it at somepoint

src/aot/aot.cpp
danobi: Yeah good catch. I think we'll still want it to be a local var cuz `< 0` is error and `>= sizeof(cmd)` is also overflow error

src/aot/aot.cpp
danobi: Ah yeah that's simpler. Let's do `bool`

danobi:
Ty for the review!

src/aot/aot.cpp
```diff
@@ -146,44 +151,102 @@ int generate(const RequiredResources &resources,
...
+  int written;
+
+  // Write out section data in a temporary file
+  std::ofstream secdata(AOT_SECDATA_TEMPFILE, std::ios_base::binary);
+  secdata.write(reinterpret_cast<const char *>(section.data()), section.size());
+  secdata.close();
+
+  // Respect user provided BPFTRACE_OBJCOPY if present
+  std::string objcopy = "objcopy";
+  if (auto c = std::getenv("BPFTRACE_OBJCOPY"))
```
jordalgo, line 192: Should this get added to the list of env vars in both the reference doc and man page:
https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#9-environment-variables

danobi: My plan was to leave this (and `--aot` cli flag) undocumented until it's ready to be used. Not necessarily completely done, but until at it's at least testable.

PR #2897, opened 2024-01-02
Support for kprobe:[module]:[function]

matthew-olson-intel:
Particularly useful in situations where you need BTF symbol information for functions in a kernel module. Passes all (previously existing) tests (on my system).

##### Checklist


matthew-olson-intel:
I'm currently working on fixing/adding some unit tests; while the current state *works* in terms of providing a kernel module's BTF symbols in a kprobe, it doesn't pass the tests "properly."

danobi:
Sounds good. Obviously no rush

tests/parser.cpp
matthew-olson-intel, line 1: I was a little iffy about changing this, but my reasoning is that this test originally failed because a dangling colon indicates that there's a second (empty) argument to that probe, so it should fail because a `kprobe` should take one argument. With this change, though, the test now ensures that the parser fails on adding a _third_ (also empty) argument.

docs/reference_guide.md
```diff
@@ -1117,6 +1117,18 @@ open path: interrupts
 [...]
 ```
 
+If the traced function is in a kernel module, you can optionally specify it to use BTF data from that
+module's BTF file:
+```
+# cat kvm.bt
```
jordalgo, line 1123: What is kvm.bt?

src/ast/attachpoint_parser.cpp
```diff
@@ -325,16 +325,25 @@ AttachPointParser::State AttachPointParser::special_parser()
 
 AttachPointParser::State AttachPointParser::kprobe_parser(bool allow_offset)
 {
-  if (parts_.size() != 2)
+  if (parts_.size() != 2 && parts_.size() != 3)
```
jordalgo, line 328: nit: maybe save off a local var e.g. `auto parts_size = parts_.size()`

man/adoc/bpftrace.adoc
```diff
@@ -2391,9 +2391,9 @@ kfunc:kvm:x86_emulate_insn { @ = count(); }
 === kprobe and kretprobe
 
 .variants
-* `kprobe:fn`
-* `kprobe:fn+offset`
-* `kretprobe:fn`
+* `kprobe[:mod]:fn`
```
jordalgo, line 2394: nit: maybe just spell this out mod -> module

tests/runtime/probe
```diff
@@ -119,6 +113,18 @@ EXPECT progs: 1
...
+PROG kprobe:vmlinux:vfs_read { printf("SUCCESS %d\n", pid); exit(); }
+EXPECT SUCCESS [0-9][0-9]*
+TIMEOUT 5
+AFTER ./testprogs/syscall read
+
+NAME kprobe_module_wildcard
+PROG kprobe:vmlinu*:vfs_read { printf("SUCCESS %d\n", pid); exit(); }
+EXPECT SUCCESS [0-9][0-9]*
+TIMEOUT 5
+AFTER ./testprogs/syscall read
```
jordalgo, line 126: Is it worth adding a runtime test for a kernel module that's either not loaded and/or doesn't exist?

jordalgo:
Few small comments but otherwise LGTM!

docs/reference_guide.md
matthew-olson-intel: Oh, it's just a made-up name of a file: similar to how the other multi-line examples are in the Reference Guide.

src/ast/attachpoint_parser.cpp
matthew-olson-intel: Yeah, that does look nicer.

man/adoc/bpftrace.adoc
matthew-olson-intel: I think that would look nicer, but I was copying what the `kfunc` documentation does above -- I'll change both.

tests/runtime/probe
matthew-olson-intel: Good idea; we need to make sure that not only does the PR *work* the same way as without a module being specified, but that it also *fails* in the same way.

matthew-olson-intel: ...Adding this revealed a bit of a problem. If I specify a kernel module, attaching appears to "succeed." That is, the `bpf_attach_kprobe` call (passing in `module:function` as an argument) succeeds in creating a perf fd. I'll continue discussing on the affected line.

src/attached_probe.cpp
```diff
@@ -964,11 +964,15 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  std::string funcname = probe_.attach_point;
+  if (probe_.path.length() > 0)
+    funcname = probe_.path + ":" + funcname;
+
   resolve_offset_kprobe(safe_mode);
   int perf_event_fd = bpf_attach_kprobe(progfd_,
                                         attachtype(probe_.type),
                                         eventname().c_str(),
-                                        probe_.attach_point.c_str(),
+                                        funcname.c_str(),
```
matthew-olson-intel, line 975: The previous PR (that I based mine on) used [this](https://www.kernel.org/doc/Documentation/trace/kprobetrace.rst) as justification for this working, but that link doesn't refer to libbpf. The real question is whether the underlying `perf_event_open` call in [libbpf](https://github.com/iovisor/bcc/blob/d9645f8253d1d8aed0d43503e70dacfb88cb1c72/src/cc/libbpf.c#L1128) is able to take a `module:function` (and properly fail if either doesn't exist).

docs/reference_guide.md
jordalgo: Ha, I see them now. I wonder why those exist when they don't point to actual bpftrace scripts in the tools folder. Seem a bit weird to me - maybe I'll make another PR to remove them all 😃

src/attached_probe.cpp
jordalgo: So what happens if the module is some junk string? Is there an eventual failure somewhere else?

matthew-olson-intel: All right. After digging around in `libbpf` a little deeper, that name just gets passed to `perf_event_open` directly (e.g. it doesn't split the module out from the function name). So, we need to compare bpftrace's behavior versus directly using the kprobe sysfs.

With this PR, `bpftrace` exhibits the following behavior:

```
# Good module, good function
./src/bpftrace -e 'kprobe:kvm:x86_emulate_insn {1}'
[SUCCESS]

# Good module, bad function
./src/bpftrace -e 'kprobe:kvm:nonsense {1}'
[ERROR]

# Bad module, good function
./src/bpftrace -e 'kprobe:nonsense:x86_emulate_insn {1}'
[SUCCESS]

# Bad module, bad function
./src/bpftrace -e 'kprobe:nonsense:nonsense {1}'
[WARNING]
```

Particularly the configuration that shows a bad module *and* a bad function
made me think that this isn't getting passed along to `perf` correctly;
perhaps `bpf_attach_kprobe` doesn't properly handle `module:function`.

*But* if we use the sysfs interface directly, we get the following behavior:

```
# echo 'p:gg kvm:x86_emulate_insn' > /sys/kernel/tracing/kprobe_events
# echo 'p:gb kvm:nonsense' > /sys/kernel/tracing/kprobe_events
bash: echo: write error: No such file or directory
# echo 'p:bg nonsense:x86_emulate_insn' > /sys/kernel/tracing/kprobe_events
# echo 'p:bb nonsense:nonsense' > /sys/kernel/tracing/kprobe_events
```

... So the only configuration that fails is `gb`, when the module is good
but the function is bad, which exactly matches the new behavior of `bpftrace`.

I think it should be good, but let me know if you think we should do additional module-checking.

jordalgo: How complicated is it to check if the module is bad? Even though this mimics the sysfs behavior, I think it would be nice to give users some better messaging. Also should this succeed if the module is bad but the function is good? My gut reaction is no.

src/bpftrace.cpp
```diff
@@ -316,6 +324,12 @@ int BPFtrace::add_probe(ast::Probe &p)
         // resolved function name are used in the probe.
         target = erase_prefix(func_id);
       }
+      else if ((probetype(attach_point->provider) == ProbeType::kprobe ||
+                probetype(attach_point->provider) == ProbeType::kretprobe) &&
+               !attach_point->target.empty())
```
viktormalik, line 329: nit: merge into the previous case?

src/driver.cpp
```diff
@@ -89,7 +89,8 @@ std::set<std::string> Driver::list_modules() const
     for (auto &ap : *probe->attach_points)
     {
       auto probe_type = probetype(ap->provider);
-      if (probe_type == ProbeType::kfunc || probe_type == ProbeType::kretfunc)
+      if (probe_type == ProbeType::kfunc || probe_type == ProbeType::kretfunc ||
+          probe_type == ProbeType::kprobe || probe_type == ProbeType::kretprobe)
```
viktormalik, line 93: Should we check that `ap->target` is non-empty?

Also please update the top comment for this function.

viktormalik (commented):
Thanks for picking up this work! Just a couple of comments.

Also what about probe listing? At the moment, running `bpftace -l kprobe:kvm:*` lists all kprobes while I'd expect it to list those from the `kvm` module only. I'm ok if we add that in a separate PR, just wanted to make sure that it gets added eventually.

src/attached_probe.cpp
viktormalik: Agreed with @jordalgo here. Even though ftrace silently succeeds with probe attachment to an invalid module, it'd be nice if we could print a warning that the module is not loaded.

src/bpftrace.cpp
matthew-olson-intel: I only unmerged the conditions for aesthetic reasons; since `kprobe` and `kretprobe` require the additional `attach_point->target` check, the indentation looks off to me. Will change though.

docs/reference_guide.md
```diff
@@ -1117,6 +1117,18 @@ open path: interrupts
 [...]
 ```
 
+If the traced function is in a kernel module, you can optionally specify it to use BTF data from that
+module's BTF file:
```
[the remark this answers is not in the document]
matthew-olson-intel: Ach, forgot to mention that use-case.

matthew-olson-intel:
> Also what about probe listing? At the moment, running `bpftace -l kprobe:kvm:*` lists all kprobes while I'd expect it to list those from the `kvm` module only. I'm ok if we add that in a separate PR, just wanted to make sure that it gets added eventually.

Well, it definitely shouldn't list _all_ kprobes, but this might be a bit tricky; e.g.:
```
sudo ./src/bpftrace -vl kfunc:kvm:*
```
returns nothing on my system. Same for other kernel modules (e.g. `i915`, which definitely has kfuncs).

src/attached_probe.cpp
matthew-olson-intel: Digging a bit further, `kfunc` seems to rely on the availability of BTF for the given module/function, so fails in 3/4 cases (as it should) based on those grounds:
```
ERROR: No BTF found for kvm:nonsense
ERROR: No BTF found for nonsense:x86_emulate_insn
ERROR: No BTF found for nonsense:nonsense
```

matthew-olson-intel: It looks like the way forward might be to simply parse `/proc/modules` (for example, in a function in `src/utils.cpp`), and call that before attaching a kprobe that specifies a module. Or we could just throw a warning based on that parsing.

That seems to be how the `lsmod` implementation in busybox does it:
https://github.com/brgl/busybox/blob/abbf17abccbf832365d9acf1c280369ba7d5f8b2/modutils/lsmod.c#L78

And kmod, too:
https://github.com/kmod-project/kmod/blob/48b0c75f79f86140314b9656d3d855a515e18a73/libkmod/libkmod-module.c#L1727

jordalgo: Yah, I was pretty much going to suggest doing whatever `lsmod` does. So that works for me.

matthew-olson-intel: Great; does that solution (a function to parse `/proc/modules`) sound good to @viktormalik too?

viktormalik:
> Well, it definitely shouldn't list _all_ kprobes, but this might be a bit tricky; e.g.:
> 
> ```
> sudo ./src/bpftrace -vl kfunc:kvm:*
> ```
> 
> returns nothing on my system. Same for other kernel modules (e.g. `i915`, which definitely has kfuncs).

That is interesting, I can reproduce that but I'm pretty sure that it worked a while ago. Let me have a look.

src/attached_probe.cpp
matthew-olson-intel: @jordalgo @viktormalik My opinion is that we should mimic the `kfunc` behavior; if the module doesn't exist, bpftrace should immediately throw an error (similar to how failing in libbpf might behave).  Thoughts?

matthew-olson-intel:
Fixing wildcards; they're completely broken with this PR.

By the way, in the case where we want to attach to a function that exists in two different modules, there aren't any examples of that on my system. I just printed out the module and function (separated by a space) from `get_symbols_from_traceable_funcs` and ran it through `sort | uniq -d -f 1`. We could obviously write a kernel module that causes that, but that'd be a pain to set up in CI.

src/attached_probe.cpp
viktormalik: Agreed, parsing `/proc/modules` makes sense and throwing an error if we cannot attach does, too. For wildcarded module attachpoints, we could just print a warning.

viktormalik:
> Fixing wildcards; they're completely broken with this PR.

I fixed the problem with kfunc wildcards (#2914) and also refactored probe matching for kprobe/kfunc a bit (#2915). With those 2 PRs merged, adding support for `kprobe:module:*` listing should be easier.
 
> By the way, in the case where we want to attach to a function that exists in two different modules, there aren't any examples of that on my system. I just printed out the module and function (separated by a space) from `get_symbols_from_traceable_funcs` and ran it through `sort | uniq -d -f 1`. We could obviously write a kernel module that causes that, but that'd be a pain to set up in CI.

When I do `cat /sys/kernel/tracing/available_filter_functions | sort | uniq | cut -d" " -f1 | uniq -d`, I can see quite a lot of functions. For instance:
```
# cat /sys/kernel/tracing/available_filter_functions | grep cleanup_entry
cleanup_entry [ip_tables]
cleanup_entry [ip6_tables]
```

But I agree that adding that to CI may be quite painful (which is why I didn't do that when adding support for `kfunc:module:function` syntax).

matthew-olson-intel:
> I fixed the problem with kfunc wildcards (#2914) and also refactored probe matching for kprobe/kfunc a bit (#2915). With those 2 PRs merged, adding support for `kprobe:module:*` listing should be easier.

Sure, I'll wait until those are merged and finish up wildcard support. Working for `-l`, but not for bpftrace programs yet (but your PR should fix it).

Then I'll write some more runtime tests, since we need to additionally test listing and probing modules and functions with wildcards:

```
-l kprobe:kvm:x86_emulate_ins*
-e 'kprobe:kvm:*'
-e 'kprobe:vmlinu*:vfs_read'
...
```

matthew-olson-intel:
All the above problems are fixed, now, including:
1. Listing kprobes.
2. Wildcards (for modules and functions).
3. More unit tests.
4. Checking `/proc/modules` to throw errors/warnings if the attachment is going to fail.
5. Avoidance of the `kprobe_multi` codepaths which were previously causing attachment errors.

src/attached_probe.cpp
```diff
@@ -960,11 +962,39 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
     return;
   }
 
+  // Construct a string containing "module:function."
+  // Also throw a warning or error if the module doesn't exist,
```
jordalgo, line 966: nit
```suggestion
  // Also log a warning or error if the module doesn't exist,
```

src/attached_probe.cpp
```diff
@@ -960,11 +962,39 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  // before attempting to attach.
+  // Note that we do not pass vmlinux, if it is specified.
+  std::string funcname = probe_.attach_point;
+  std::string modname = probe_.path;
```
jordalgo, line 970:
```suggestion
  std::string& modname = probe_.path;
```

src/attached_probe.cpp
```diff
@@ -960,11 +962,39 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  if ((modname.length() > 0) && (modname.compare("vmlinux") != 0))
+  {
+    if (!is_loaded_module(modname))
+    {
+      if (probe_.orig_name != probe_.name)
+      {
+        // Wildcard usage just gets a warning
+        LOG(WARNING) << "module " << modname << " in probe " << probe_.name
```
jordalgo, line 978: If the module isn't loaded - how do we know it doesn't exist? Rather why doesn't this mirror the error message below?

src/utils.cpp
```diff
@@ -693,6 +693,32 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
...
+    return true;
+  }
+
+  // This file lists all loaded modules
+  std::ifstream modules_file("/proc/modules");
+
+  std::vector<std::string> tokens;
+  for (std::string line; std::getline(modules_file, line);)
+  {
+    tokens = split_string(line, ' ');
```
jordalgo, line 709: Maybe instead of the split we can just compare against the substring

src/utils.h
```diff
@@ -198,6 +198,7 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_hierarchy_roots();
 std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
     uint64_t cgroupid,
     std::string filter);
+bool is_loaded_module(const std::string &module);
```
jordalgo, line 201:
```suggestion
bool is_module_loaded(const std::string &module);
```

tests/runtime/btf
```diff
@@ -91,3 +91,9 @@ EXPECT Attaching 1 probe...
 TIMEOUT 2
 REQUIRES_FEATURE kfunc
 REQUIRES lsmod | grep '^kvm'
+
+NAME kprobe_kernel_module_type_fwd
+PROG kprobe:kvm:x86_emulate_insn { $ctxt = (struct x86_emulate_ctxt *) arg0; printf("%d\n", $ctxt->mode); exit(); }
+EXPECT Attaching 1 probe...
+TIMEOUT 2
+REQUIRES lsmod | grep '^kvm'
```
jordalgo, line 99: Is it possible to write a BEFORE program to load the kvm module if it's not already loaded?

jordalgo (changes requested):
Some minor comments but otherwise looks pretty good.

src/attached_probe.cpp
```diff
@@ -960,11 +962,39 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
     return;
   }
 
+  // Construct a string containing "module:function."
+  // Also throw a warning or error if the module doesn't exist,
+  // before attempting to attach.
+  // Note that we do not pass vmlinux, if it is specified.
+  std::string funcname = probe_.attach_point;
+  std::string modname = probe_.path;
+  if ((modname.length() > 0) && (modname.compare("vmlinux") != 0))
```
viktormalik, line 971: Any reason not to do this?
```suggestion
  if ((modname.length() > 0) && modname != "vmlinux"))
```

src/bpftrace.cpp
```diff
@@ -301,7 +310,10 @@ int BPFtrace::add_probe(ast::Probe &p)
                probetype(attach_point->provider) == ProbeType::uprobe ||
                probetype(attach_point->provider) == ProbeType::uretprobe ||
                probetype(attach_point->provider) == ProbeType::kfunc ||
-               probetype(attach_point->provider) == ProbeType::kretfunc)
+               probetype(attach_point->provider) == ProbeType::kretfunc ||
+               ((probetype(attach_point->provider) == ProbeType::kprobe ||
+                 probetype(attach_point->provider) == ProbeType::kretprobe) &&
+                !attach_point->target.empty()))
```
viktormalik, line 316: Also update the below comment please.

src/attached_probe.cpp
```diff
@@ -760,7 +761,8 @@ void AttachedProbe::load_prog(BPFfeature &feature)
 
       if ((probe_.type == ProbeType::kprobe ||
            probe_.type == ProbeType::kretprobe) &&
-          feature.has_kprobe_multi() && !probe_.funcs.empty())
+          feature.has_kprobe_multi() && !probe_.funcs.empty() &&
+          probe_.path.empty())
```
viktormalik, line 765: Should we explain the reason for avoiding `kprobe_multi` in this case?

src/driver.cpp
```diff
@@ -89,7 +89,10 @@ std::set<std::string> Driver::list_modules() const
     for (auto &ap : *probe->attach_points)
     {
       auto probe_type = probetype(ap->provider);
-      if (probe_type == ProbeType::kfunc || probe_type == ProbeType::kretfunc)
+      if (probe_type == ProbeType::kfunc || probe_type == ProbeType::kretfunc ||
+          ((probe_type == ProbeType::kprobe ||
+            probe_type == ProbeType::kretprobe) &&
+           !ap->target.empty()))
```
viktormalik, line 95: Future work idea: we could eventually drop the last part of the condition and parse BTF also for module functions specified in the form `kprobe:function` (as we do for kfuncs).

src/utils.cpp
```diff
@@ -693,6 +693,32 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
   return result;
 }
 
+bool is_loaded_module(const std::string &module)
+{
+  if (module.compare("vmlinux") == 0)
```
viktormalik, line 698: Same as above:
```suggestion
  if (module == "vmlinux")
```

viktormalik (commented):
Couple of additional comments/suggestions. Otherwise looks nice!

src/driver.cpp
matthew-olson-intel: Yeah, that's a good idea too. Probably wouldn't be too tough, either.

src/attached_probe.cpp
matthew-olson-intel: It should definitely mirror the below error message.

tests/runtime/btf
matthew-olson-intel: Good idea. I'll do that in another PR, since I'd rather not sneak it into this one; there may be a good reason that it's not already done.

matthew-olson-intel:
Sorry about the spam. That's what I get for clicking the "Commit Changes" button in GitHub on each of the suggestions instead of properly committing them; they bypassed my testing and my pre-commit formatting hook.

src/attached_probe.cpp
```diff
@@ -758,9 +759,12 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         opts.expected_attach_type = static_cast<::bpf_attach_type>(
             libbpf::BPF_TRACE_ITER);
 
+      // We want to avoid kprobe_multi here because the BPF_TRACE_KPROBE_MULTI
+      // link type does not currently support the `module:function` syntax.
```
viktormalik, line 763: I don't like the wording too much, what about
```suggestion
      // We want to avoid kprobe_multi when a module is specified because the BPF_TRACE_KPROBE_MULTI
      // link type does not currently support the `module:function` syntax.
```

src/attached_probe.cpp
```diff
@@ -960,11 +964,39 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
...
+  const std::string &modname = probe_.path;
+  if ((modname.length() > 0) && modname != "vmlinux")
+  {
+    if (!is_loaded_module(modname))
+    {
+      if (probe_.orig_name != probe_.name)
+      {
+        // Wildcard usage just gets a warning
+        LOG(WARNING) << "specified module " << modname << " in probe "
+                     << probe_.name << " is not loaded.";
```
viktormalik, line 981: Could we prepare the message into a string and then send it either to `LOG(WARNING)` or `LOG(ERROR)`? That way, when we want to change the message, we just do it in one place.

src/utils.cpp
```diff
@@ -693,6 +693,30 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
...
+    return true;
+  }
+
+  // This file lists all loaded modules
+  std::ifstream modules_file("/proc/modules");
+
+  for (std::string line; std::getline(modules_file, line);)
+  {
+    std::string first_token = line.substr(0, line.find(' '));
+    if (module == first_token)
```
viktormalik, line 709: Nit: this should be slightly faster as it requires just one pass through the string:
```suggestion
    if (line.compare(0, module.size() + 1, module + " ") == 0)
```

src/utils.h
viktormalik: Still applies, right?

matthew-olson-intel: Yep, for sure. Got lost in the formatting fixup that I did after clicking the "Commit suggestion" button. Sorry about that.

matthew-olson-intel:
@viktormalik Applied all suggestions. Thank you again for the review!

viktormalik (approved):
Thanks for the work!

Let's wait for at least one more approval and we'll squash the PR afterwards.

src/probe_matcher.cpp
```diff
@@ -115,7 +115,10 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     case ProbeType::kprobe:
     case ProbeType::kretprobe:
     {
-      symbol_stream = get_symbols_from_traceable_funcs();
+      if (!target.empty())
+        symbol_stream = get_symbols_from_traceable_funcs(true);
+      else
+        symbol_stream = get_symbols_from_traceable_funcs(false);
```
jordalgo, line 121:
```suggestion
      symbol_stream = get_symbols_from_traceable_funcs(!target.empty());
```

src/utils.cpp
```diff
@@ -693,6 +693,29 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
   return result;
 }
 
+bool is_module_loaded(const std::string &module)
+{
+  if (module == "vmlinux")
+  {
+    return true;
+  }
```
jordalgo, line 701: Even though I hate this kernel style convention. I think we don't use brackets for single statements under conditionals
```suggestion
  if (module == "vmlinux")
    return true;
```

tests/bpftrace.cpp
```diff
@@ -19,21 +19,27 @@ using ::testing::StrictMock;
...
 
 static const std::string kprobe_name(const std::string &attach_point,
+                                     const std::string &target,
                                      uint64_t func_offset)
 {
   auto str = func_offset ? "+" + std::to_string(func_offset) : "";
+  if (!target.empty())
+  {
+    return "kprobe:" + target + ":" + attach_point + str;
+  }
```
jordalgo, line 29:
```suggestion
  if (!target.empty())
    return "kprobe:" + target + ":" + attach_point + str;
```

jordalgo (approved):
Thanks for the fast follow-ups! Looks great. Just a few more minor style nits but I'll let the maintainers decide if you want to skip them.

src/utils.cpp
matthew-olson-intel: Ah, OK, I've got no issue with changing this. Didn't know that `clang-format` didn't support this (but a quick Google reveals it's a relatively new feature).

jordalgo: @matthew-olson-intel Actually you can ignore my comment. Seems like bpftrace doesn't require the removing of these brackets - at least according to @viktormalik

src/probe_matcher.cpp
matthew-olson-intel: I think this makes the intention of the code slightly more difficult to read, even if it's a few more lines, so I'd prefer the original.

src/utils.cpp
matthew-olson-intel: All right -- either way!

viktormalik:
Merged, thanks for the work!

PR #2900, opened 2024-01-04
Fix unit tests with kprobe_multi/uprobe_multi

viktormalik:
This updates probe addition unit tests w.r.t. presence of `(k|u)probe_multi`. Until now, tests for `kprobe_multi` checked its availability on the running system and adjusted the expected results accordingly. Tests with uprobe wildcards didn't consider `uprobe_multi` at all and therefore failed on some systems.

This PR uses `MockBPFfeature` to control the values of `(k|u)probe_multi` in the unit tests and adds separate tests cases for when the feature is available and when not.

Also changes the probe name for uprobes with wildcarded target when `uprobe_multi` is present (see the first commit for details).

##### Checklist

### DIFF
--- a/tests/bpftrace.cpp
+++ b/tests/bpftrace.cpp
@@ -222,34 +222,41 @@ TEST(bpftrace, add_probes_wildcard)
       "kprobe:sys_read,kprobe:my_*,kprobe:sys_write{}");
 
   auto bpftrace = get_strict_mock_bpftrace();
+  bpftrace->feature_ = std::make_unique<MockBPFfeature>(false);
+
   EXPECT_CALL(*bpftrace->mock_probe_matcher, get_symbols_from_traceable_funcs())
       .Times(1);
 
-  if (bpftrace->has_kprobe_multi())
-  {
-    ASSERT_EQ(0, bpftrace->add_probe(*probe));
-    ASSERT_EQ(3U, bpftrace->get_probes().size());
-    ASSERT_EQ(0U, bpftrace->get_special_probes().size());
+  ASSERT_EQ(0, bpftrace->add_probe(*probe));
+  ASSERT_EQ(4U, bpftrace->get_probes().size());
+  ASSERT_EQ(0U, bpftrace->get_special_probes().size());
 
-    std::string probe_orig_name =
-        "kprobe:sys_read,kprobe:my_*,kprobe:sys_write";
-    check_kprobe(bpftrace->get_probes().at(0), "sys_read", probe_orig_name);
-    check_kprobe(bpftrace->get_probes().at(1), "my_*", probe_orig_name);
-    check_kprobe(bpftrace->get_probes().at(2), "sys_write", probe_orig_name);
-  }
-  else
-  {
-    ASSERT_EQ(0, bpftrace->add_probe(*probe));
-    ASSERT_EQ(4U, bpftrace->get_probes().size());
-    ASSERT_EQ(0U, bpftrace->get_special_probes().size());
+  std::string probe_orig_name = "kprobe:sys_read,kprobe:my_*,kprobe:sys_write";
+  check_kprobe(bpftrace->get_probes().at(0), "sys_read", probe_orig_name);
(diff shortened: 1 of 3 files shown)


tests/bpftrace.cpp
```diff
@@ -376,20 +391,39 @@ TEST(bpftrace, add_probes_uprobe_wildcard_for_target)
...
-      bpftrace->get_probes().at(2), "/bin/sh", "second_open", probe_orig_name);
-  check_uprobe(bpftrace->get_probes().at(3),
-               "/proc/1234/exe",
-               "third_open",
-               probe_orig_name);
+  if (bpftrace->has_uprobe_multi())
+  {
+    ASSERT_EQ(3U, bpftrace->get_probes().size());
+    check_uprobe(
+        bpftrace->get_probes().at(2), "/bin/bash", "*open", probe_orig_name);
```
jordalgo, line 401: nit: why the reverse index checking order? 2, 1, 0

jordalgo:
One small nit but LGTM. Nice improvement.

jordalgo (approved):

tests/bpftrace.cpp
viktormalik: ah, my mistake, thanks

tests/mocks.h
```diff
@@ -93,6 +93,11 @@ class MockBPFtrace : public BPFtrace
     return feature_->has_kprobe_multi();
   }
 
+  bool has_uprobe_multi(void)
+  {
+    return feature_->has_uprobe_multi();
```
danobi, line 98: this reads from the running system? Is it possible to instead explicitly test both codepaths by manually setting has_uprobe_multi?

viktormalik: Yeah, I just used the same approach as kprobe_multi does but you're right that unit tests should be system-independent. Let me fix that for both uprobe_multi and kprobe_multi.

viktormalik: Fixed, please re-review, thanks! (I also updated the PR description)

tests/bpftrace.cpp
```diff
@@ -222,34 +222,41 @@ TEST(bpftrace, add_probes_wildcard)
...
-    check_kprobe(bpftrace->get_probes().at(0), "sys_read", probe_orig_name);
-    check_kprobe(bpftrace->get_probes().at(1), "my_one", probe_orig_name);
-    check_kprobe(bpftrace->get_probes().at(2), "my_two", probe_orig_name);
-    check_kprobe(bpftrace->get_probes().at(3), "sys_write", probe_orig_name);
-  }
+TEST(bpftrace, add_probes_wildcard_kprobe_multi)
+{
+  ast::Probe *probe = parse_probe(
+      "kprobe:sys_read,kprobe:my_*,kprobe:sys_write{}");
+
```
jordalgo, line 245: nit: extra line between var declarations

tests/mocks.h
```diff
@@ -114,6 +109,7 @@ class MockBPFfeature : public BPFfeature
     has_d_path_ = std::make_optional<bool>(has_features);
     has_ktime_get_boot_ns_ = std::make_optional<bool>(has_features);
     has_kprobe_multi_ = std::make_optional<bool>(has_features);
+    has_uprobe_multi_ = std::make_optional<bool>(has_features);
```
jordalgo, line 112: Not for this PR but it would be nice if setting these specific features can be more explicit in a particular test. `bpftrace->feature_ = std::make_unique<MockBPFfeature>(true);` doesn't tell me much about what is being enabled.

jordalgo (approved):

tests/bpftrace.cpp
viktormalik: Yeah, that's on purpose to visually separate the "mocks setting" below.

tests/mocks.h
viktormalik: Agreed, would be nice to have a way to simply setup `MockBPFfeature` with particular settings.

danobi (approved):
ty for the cleanup!

PR #2904, opened 2024-01-07
execsnoop: use `strftime` instead of `elapsed`

makelinux:
because `elapsed` gives relative time, and `strftime` gives absolute time which is more readable and practical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and this project adheres to
 #### Tools
 - Add PPID field to `execsnoop.bt`
   - [#2876](https://github.com/iovisor/bpftrace/pull/2876)
+- Use `strftime` instead of `elapsed` in `execsnoop.bt`
+  - [#2904](https://github.com/iovisor/bpftrace/pull/2904)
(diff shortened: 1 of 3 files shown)


viktormalik (commented):
Makes sense to me. Maybe we could just respect the original implementation and use ms instead of ns?

makelinux:
> Makes sense to me. Maybe we could just respect the original implementation and use ms instead of ns?

[strftime](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#25-strftime-formatted-timestamp)  receives integer nanoseconds and provides decimal seconds only in microseconds (six digits). It can't be changed.

danobi (commented):
seems reasonable to me. but please add changelog entry under "tools" section

viktormalik:
> > Makes sense to me. Maybe we could just respect the original implementation and use ms instead of ns?
> 
> [strftime](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#25-strftime-formatted-timestamp) receives integer nanoseconds and provides decimal seconds only in microseconds (six digits). It can't be changed.

Right, that makes sense. Ok, let's stick with us.

Issue #2809, opened 2023-10-21
silent hash map truncation to 4096


mjguzik:
I tried to find out how many chains in a hash table are unused in a certain workload and added a probe to record all computed hash values, then used it like so:

`kprobe:hashval { @hash[arg0] = count(); }`

Then I found the list is truncated to 4096 elements, without any warning. I only spotted the problem because the number is round.

Since one only gets a partial result without being told this is the case, I consider this to be a bug. It is easy to trip over it and not realize unless you count entries.

So, bare minimum this should warn entries got dropped. Best case is autosizing as RAM permits and only warning if it can't be done. I'm guessing this would require patching eBPF, but one has to start with a report somewhere.

### `bpftrace --info` output
System
  OS: Linux 6.6.0-rc6+ #367 SMP PREEMPT_DYNAMIC Sat Oct 21 06:52:28 EDT 2023
  Arch: x86_64

Build
  version: v0.17.0
  LLVM: 14.0.6
  unsafe uprobe: no
  bfd: no
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes
  skboutput: no

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  map batch: yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  iter:task: yes
  iter:task_file: yes
  kprobe_multi: no
  raw_tp_special: yes

fbs:
Have you tried setting `BPFTRACE_MAP_KEYS_MAX`

Do agree that it needs better warnings, silently 'corrupting' data is bad

danobi:
`-k` will report this: 

```
dxu@fedora~ $ sudo BPFTRACE_MAP_KEYS_MAX=1 bpftrace -e 'BEGIN { @[1] = 1; @[2] = 2; exit() }'
Attaching 1 probe...


@[1]: 1
[...]

PR #2905, opened 2024-01-07
Add 'get_helper_error_msg' utility

jordalgo:
This is for additional helpful messaging for specific errors e.g. for when the number of bpf map entries is exceeded.

Follow up to: https://github.com/iovisor/bpftrace/issues/2809

Sample output:
```
$ sudo BPFTRACE_MAP_KEYS_MAX=1 bpftrace -e 'BEGIN { @[1] = 1; @[2] = 2; exit() }' -k
Attaching 1 probe...
stdin:1:24-25: WARNING: Map full; can't update element. Try increasing MAP_KEYS_MAX.
Additional Info - helper: map_update_elem, retcode: -7
BEGIN { @[1] = 1; @[2] = 2; exit() }
                       ~
@[1]: 1

$ sudo BPFTRACE_MAP_KEYS_MAX=1 bpftrace -e 'BEGIN { @[1] = 1; @[2] = 2; exit() }' -k -f json
{"type": "helper_error", "msg": "Map full; can't update element. Try increasing MAP_KEYS_MAX.", "helper": "map_update_elem", "retcode": -7, "line": 1, "col": 24}

{"type": "map", "data": {"@": {"1": 1}}}
```

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -4,6 +4,14 @@
 #include "utils.h"
 #include <async_event_types.h>
 
+#include <bpf/libbpf.h>
+
+namespace libbpf {
+#define __BPF_NAME_FN(x) #x
+const char *bpf_func_name[] = { __BPF_FUNC_MAPPER(__BPF_NAME_FN) };
+#undef __BPF_NAME_FN
+} // namespace libbpf
+
 namespace bpftrace {
 
 namespace {
@@ -145,6 +153,30 @@ void Output::lhist_prepare(const std::vector<uint64_t> &values, int min, int max
   if (start_value == -1) {
     start_value = 0;
   }
+}
+
+std::string Output::get_helper_error_msg(int func_id, int retcode) const
+{
+  std::string msg;
+  if (func_id == libbpf::BPF_FUNC_map_update_elem && retcode == -E2BIG)
+  {
+    msg = "Map full; can't update element. Try increasing MAP_KEYS_MAX.";
+  }
+  else if (func_id == libbpf::BPF_FUNC_map_delete_elem && retcode == -ENOENT)
+  {
+    msg = "Can't delete map element because it does not exist.";
+  }
+  // bpftrace sets the return code to 0 for map_lookup_elem failures
+  // which is why we're not also checking the retcode
+  else if (func_id == libbpf::BPF_FUNC_map_lookup_elem)
+  {
+    msg = "Can't lookup map element because it does not exist.";
+  }
+  else
+  {
+    msg = strerror(-retcode);
+  }
+  return msg;
 }
 
 std::string Output::value_to_str(BPFtrace &bpftrace,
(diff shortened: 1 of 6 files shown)


viktormalik (approved):
Looks good! Do we know about other helpers that could use better error messages?

jordalgo:
> Do we know about other helpers that could use better error messages?

Good question. I'll take a look and see if anything jumps out.

danobi (approved):
nice!

danobi:
Changelog entry?

src/output.cpp
```diff
@@ -147,6 +155,16 @@ void Output::lhist_prepare(const std::vector<uint64_t> &values, int min, int max
   }
 }
 
+std::string Output::get_helper_error_msg(int func_id, int retcode) const
+{
+  std::string msg = strerror(-retcode);
+  if (func_id == libbpf::BPF_FUNC_map_update_elem && retcode == -E2BIG)
+  {
+    msg.append(" (trying increasing MAP_KEYS_MAX)");
```
ajor, line 163:
```suggestion
    msg.append(" (try increasing MAP_KEYS_MAX)");
```
Typo

ajor (commented):
Good addition! While we're adding this extra detail to this specific error message, I'm thinking that we might also want to remove unnecessary info from it too. i.e. "Failed to map_update_elem: Argument list too long"

map_update_elem is an implementation detail and "Argument list too long" doesn't mean much.

Maybe we should say something more user friendly along the lines of: "Map full, can't update element (try ...)"

jordalgo:
> map_update_elem is an implementation detail and "Argument list too long" doesn't mean much.

Yeah good call. Part of me wants to leave in 'map_update_elem' for devs debugging the issue even though it's an implementation detail. The json output separates this bpf function into a field called "helper". Maybe for non-json output we can just follow the error message with a similar format e.g.

`
Map full, can't add or update element. Try increasing MAP_KEYS_MAX. (helper: map_update_elem, retcode: 32).
`

jordalgo:
> Do we know about other helpers that could use better error messages?

Added two more for map lookup and deletion.

src/output.cpp
```diff
@@ -147,6 +155,30 @@ void Output::lhist_prepare(const std::vector<uint64_t> &values, int min, int max
   }
 }
 
+std::string Output::get_helper_error_msg(int func_id, int retcode) const
+{
+  std::string msg;
+  if (func_id == libbpf::BPF_FUNC_map_update_elem && retcode == -E2BIG)
+  {
+    msg = "Map full; can't update element. Try increasing MAP_KEYS_MAX.";
```
danobi, line 163: This is the "old" name for w/ the config changes. Sounds like there's consensus on changing the env var names - should we preemptively use BPFTRACE_MAX_MAP_KEYS here?

jordalgo: Depends on the merging order :) This is a smaller change so I imagine this will get merged first. Happy to change it with the config PR.

dxuuu (approved):

src/output.cpp
danobi: I think this is good to merge. Then the other PR can rebase on top

viktormalik:
@ajor can we merge this?

ajor (approved):
Looks good to me!

PR #2909, opened 2024-01-09
ci: Remove IRC notification

danobi:
I haven't seen this fire for broken builds in a long time. Not quite sure why it's broken. But I'm also not sure this is actually useful. Looking at GH is probably more convenient, at least for me.

##### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,22 +125,3 @@ jobs:
     - name: Build and test
       env: ${{matrix.env}}
       run: ./.github/include/ci.py
-
-  irc:
-    # Notify IRC of build failures on pushes only if we are running from
-    # the main repo. We don't want this rule to trigger from forked repos.
-    needs:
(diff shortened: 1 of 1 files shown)


jordalgo:
I agree - less IRC :)

jordalgo (approved):

viktormalik (commented):
It fired just yesterday :-)

I find it a bit useful as it notifies me that master CI is broken and I'm not sure if there's other way to get a notification on that. If I just checked GitHub, there'll be a larger delay between the failure and the fix (in 95% of cases, CI just needs a restart).

OTOH, it happens occasionally and it's probably not a big deal that master CI is broken for a short time so I'm not against removing this if others don't find it useful.

danobi:
Ah, well if it works and you find it helpful, we can keep it. I don't it costs anythign to maintain - I probably wrote that config 4 years ago and nobody's touched it since.

PR #2910, opened 2024-01-10
Use unique names for stackid maps

viktormalik:
We create a single BPF map for each `StackType`, defined by stack mode and the limit, used within the program. Until now, all these maps had the same name `stack`. While this is technically possible, it will cause problems once we start emitting BTF entries for maps.

To prevent this problem, use a unique name for each map in the form `stack_<mode>_<limit>`.

##### Checklist

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -78,6 +78,12 @@ enum class StackMode : uint8_t
   raw,
 };
 
+const std::map<StackMode, std::string> STACK_MODE_NAME_MAP = {
+  { StackMode::bpftrace, "bpftrace" },
+  { StackMode::perf, "perf" },
+  { StackMode::raw, "raw" },
+};
+
 struct StackType
 {
   uint16_t limit = DEFAULT_STACK_SIZE;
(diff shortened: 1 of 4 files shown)


src/types.h
```diff
@@ -78,6 +78,12 @@ enum class StackMode : uint8_t
   raw,
 };
 
+const std::map<StackMode, std::string> STACK_MODE_NAME_MAP = {
```
jordalgo, line 81: types used to have this map reversed but I moved it to config.h:
https://github.com/iovisor/bpftrace/blob/master/src/config.h#L71

If you want (or I can do it it another PR), just keep the one here and in the Config constructor, build the reverse of this to be used in config.cpp

viktormalik: Yeah, I noticed, but this one doesn't logically belong to `config.h`. Building the reverse one dynamically probably makes sense so that we don't forget to update both if we add new stack types. Let me add that to this PR.

viktormalik: Done.

jordalgo (approved):

src/types.h
jordalgo: Thanks!

danobi (approved):
lgtm

PR #2912, opened 2024-01-10
Introduce BpfBytecode class

viktormalik:
So far, the generated BPF bytecode was represented as a map of section names to their contents. With the upcoming movement towards libbpf-based program loading and attachment, we will need to attach more information to the bytecode, so this introduces a new class `BpfBytecode` for that purpose.

For now, the class only contains the original section map and helper methods to access it.

No functional change.

##### Checklist

### DIFF
--- a/src/bpfbytecode.cpp
+++ b/src/bpfbytecode.cpp
@@ -1,0 +1,28 @@
+#include "bpfbytecode.h"
+
+#include <stdexcept>
+
+namespace bpftrace {
+
+void BpfBytecode::addSection(const std::string &name,
+                             std::vector<uint8_t> &&data)
+{
+  sections_.emplace(name, data);
+}
+
+bool BpfBytecode::hasSection(const std::string &name) const
+{
+  return sections_.find(name) != sections_.end();
+}
+
+const std::vector<uint8_t> &BpfBytecode::getSection(
+    const std::string &name) const
+{
+  if (!hasSection(name))
+  {
+    throw std::runtime_error("Bytecode is missing section " + name);
+  }
+  return sections_.at(name);
+}
+
+} // namespace bpftrace
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -17,6 +17,7 @@ add_library(runtime
   attached_probe.cpp
   bpffeature.cpp
   bpftrace.cpp
+  bpfbytecode.cpp
   bpfprogram.cpp
   btf.cpp
   child.cpp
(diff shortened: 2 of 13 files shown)


jordalgo:
Nit: This is probably a good PR to introduce the BpfBytecode class test with maybe just a few basics testing `hasSection` and `getSection`.

viktormalik:
> Nit: This is probably a good PR to introduce the BpfBytecode class test with maybe just a few basics testing `hasSection` and `getSection`.

There is an [existing test](https://github.com/iovisor/bpftrace/blob/6a62ccecedbe308d74327279ba6857a0bd93b604/tests/codegen/general.cpp#L63-L64) for `hasSection`. We could extend it with `getSection` although I'm not sure what we'd test there as a section is just a sequence of bytes.

jordalgo:
> There is an [existing test](https://github.com/iovisor/bpftrace/blob/6a62ccecedbe308d74327279ba6857a0bd93b604/tests/codegen/general.cpp#L63-L64) for hasSection. We could extend it with getSection although I'm not sure what we'd test there as a section is just a sequence of bytes.

That exists as part of a codegen test, I was thinking just in terms of organization of unit tests it's better to put these into their own test file as I imagine this class growing (just seems like a good time since you're making a new class).

viktormalik:
> That exists as part of a codegen test, I was thinking just in terms of organization of unit tests it's better to put these into their own test file as I imagine this class growing (just seems like a good time since you're making a new class).

That's a good idea. On top of that, tests in `tests/codegen/general.cpp` are valid for all LLVM versions but only get executed for LLVM 12. I moved the relevant test to a new test file.

src/bpfbytecode.cpp
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const std::vector<uint8_t> &BpfBytecode::getSection(
+    const std::string &name) const
+{
+  if (!hasSection(name))
+  {
+    throw std::runtime_error("Bytecode is missing section " + name);
+  }
+  return sections_.at(name);
```
danobi, line 25: Not that it matters, but I believe `at()` can also throw. Technically more efficient to use `[]` operator

danobi (approved):

src/bpfbytecode.cpp
danobi: Actually I don't know of the efficient part is true. So ignore that comment I guess

jordalgo (approved):
LGTM!

src/bpfbytecode.cpp
ajor: Yes, `[]` is marginally more efficient due to not having the bounds checks

PR #2914, opened 2024-01-11
Fix attachment/listing of wildcarded module kfuncs

viktormalik:
Attachment to e.g. all functions in a module is broken:

    # bpftrace -e 'kfunc:nvme:*' { @[func] = count() }
    No probes to attach

This fixes the bug causing the above problem:

    # bpftrace -e 'kfunc:nvme:*' { @[func] = count() }
    Attaching 65 probes...
    [...]

Also adds a runtime tests to check for future regressions.

##### Checklist


src/driver.cpp
```diff
@@ -95,8 +95,9 @@ std::set<std::string> Driver::list_modules() const
         {
           for (auto &match : bpftrace_.probe_matcher_->get_matches_for_ap(*ap))
           {
-            std::string mod = match;
-            auto match_modules = bpftrace_.get_func_modules(erase_prefix(mod));
+            std::string func = match;
+            erase_prefix(func);
+            auto match_modules = bpftrace_.get_func_modules(erase_prefix(func));
```
jordalgo, line 100: Calling `erase_prefix` twice?

viktormalik: Damn, no, that's a mistake. It works b/c it's basically a no-op, but the second `erase_prefix` shouldn't be there. Good catch, fixed.

jordalgo: Ah it's because `erase_prefix` returns the prefix and not the modified string - a little prone to errors like this with a signature like that 😄

jordalgo (approved):

src/driver.cpp
viktormalik: Right, `get_and_erase_prefix` would be a better name. I'll keep that in mind and change it when I work on something related.

tests/runtime/btf
```diff
@@ -45,6 +45,16 @@ REQUIRES lsmod | grep '^nf_tables'
 REQUIRES /usr/sbin/nft --help
 CLEANUP nft delete table bpftrace
 
+NAME kernel_module_attach_wildcard
```
danobi, line 48: Would be neat to have a `NEW_NETNS` runtime test directive to provide a little more isolation. But not needed for now

danobi (approved):

PR #2879, opened 2023-12-13
Fix memory leak when listing probes

jordalgo:
Repro command with ASAN build: `sudo bpftrace -l "kprobe_seq_*"`

Fix is to also free the memory after erasing the attach point from the vector of attach points.

Error output:
```
$ sudo bpftrace -l "kprobe_seq_*"
kfunc:vmlinux:kprobe_seq_next
kfunc:vmlinux:kprobe_seq_start
kfunc:vmlinux:kprobe_seq_stop
kprobe:kprobe_seq_next
kprobe:kprobe_seq_start
kprobe:kprobe_seq_stop

=================================================================
==3018628==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 456 byte(s) in 1 object(s) allocated from:
    #0 0x7f93b34b6367 in operator new(unsigned long) (/lib64/libasan.so.6+0xb6367)
    #1 0x7c0a0f in bpftrace::Parser::parse() (/usr/local/bin/bpftrace+0x7c0a0f)
    #2 0x4c8275 in bpftrace::Driver::parse() (/usr/local/bin/bpftrace+0x4c8275)
    #3 0x426389 in main (/usr/local/bin/bpftrace+0x426389)
    #4 0x7f93aa03feaf in __libc_start_call_main (/lib64/libc.so.6+0x3feaf)

SUMMARY: AddressSanitizer: 456 byte(s) leaked in 1 allocation(s).
```

##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #2915, opened 2024-01-11
Optimize probe matcher's work with available_filter_functions

viktormalik:
Probe matcher uses data from `/sys/kernel/tracing/available_filter_functions` in 2 places:
1. when matching kprobes and
2. when matching kfuncs when BTF is not yet parsed (this is used to determine the list of modules for which we need to parse BTF).

For both use-cases, there's a number of unnecessary string transformations and parsing done. The current workflow is as follows:
1. Parse `available_filter_functions` into `BPFtrace::traceable_funcs_`, which is a function/module map.
2. For kprobe matching, read entries from `traceable_funcs_`, transform them into `available_filter_functions` format (`function [module]`), and then parse that again in probe matcher with always ignoring the module part.
3. For kfunc matching, read entries from the `available_filter_functions` file (`function [module]` format), transform them into bpftrace probes format (`module:function`), and finally run matching against that.

It's obvious that most of the points (2) and (3) is unnecessary as we already have parsed data in `traceable_funcs_` and can transform them directly into the expected format. So, this PR proposes to change the above workflow to:
1. Stays the same.
2. For kprobe, read entries from `traceable_funcs_`, transform them into `func` format (as kprobes do not support modules, yet), and use that for matching.
3. For kfunc, read entries from `traceable_funcs_`, transform them into `module:func` format, and use that for matching.

This removes a bit of code and also speeds up kfunc attachment since `available_filter_functions` is now always parsed just once.

##### Checklist

### DIFF
--- a/tests/bpftrace.cpp
+++ b/tests/bpftrace.cpp
@@ -224,7 +224,8 @@ TEST(bpftrace, add_probes_wildcard)
   auto bpftrace = get_strict_mock_bpftrace();
   bpftrace->feature_ = std::make_unique<MockBPFfeature>(false);
 
-  EXPECT_CALL(*bpftrace->mock_probe_matcher, get_symbols_from_traceable_funcs())
+  EXPECT_CALL(*bpftrace->mock_probe_matcher,
+              get_symbols_from_traceable_funcs(false))
       .Times(1);
 
   ASSERT_EQ(0, bpftrace->add_probe(*probe));
@@ -246,7 +247,8 @@ TEST(bpftrace, add_probes_wildcard_kprobe_multi)
   auto bpftrace = get_strict_mock_bpftrace();
   bpftrace->feature_ = std::make_unique<MockBPFfeature>(true);
 
-  EXPECT_CALL(*bpftrace->mock_probe_matcher, get_symbols_from_traceable_funcs())
+  EXPECT_CALL(*bpftrace->mock_probe_matcher,
+              get_symbols_from_traceable_funcs(false))
       .Times(1);
 
   ASSERT_EQ(0, bpftrace->add_probe(*probe));
@@ -265,7 +267,8 @@ TEST(bpftrace, add_probes_wildcard_no_matches)
       "kprobe:sys_read,kprobe:not_here_*,kprobe:sys_write{}");
 
   auto bpftrace = get_strict_mock_bpftrace();
-  EXPECT_CALL(*bpftrace->mock_probe_matcher, get_symbols_from_traceable_funcs())
+  EXPECT_CALL(*bpftrace->mock_probe_matcher,
+              get_symbols_from_traceable_funcs(false))
       .Times(1);
 
   ASSERT_EQ(0, bpftrace->add_probe(*probe));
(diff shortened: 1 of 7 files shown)


jordalgo:
Do we need to completely filter out functions with modules (aside from vmlinux) when matching for (or listing) kprobes? I'm thinking of kernel function which might have the same name as another function in a module.

viktormalik:
> Do we need to completely filter out functions with modules (aside from vmlinux) when matching for (or listing) kprobes? 

I don't think so. It's possible to attach to kprobes in modules by just specifying the function name and we support that.

> I'm thinking of kernel function which might have the same name as another function in a module.

To deal with such situations, we'll need the `kprobe:module:function` syntax from #2897.

jordalgo:
> I don't think so. It's possible to attach to kprobes in modules by just specifying the function name and we support that.

Ah ok. I think I was thrown off in the description "as kprobes do not support modules, yet" - but you just meant it's lacking module filtering which is added https://github.com/iovisor/bpftrace/pull/2897.

viktormalik:
> > I don't think so. It's possible to attach to kprobes in modules by just specifying the function name and we support that.
> 
> Ah ok. I think I was thrown off in the description "as kprobes do not support modules, yet" - but you just meant it's lacking module filtering which is added #2897.

Yes, that's it, the wording was not entirely correct.

tests/bpftrace.cpp
```diff
@@ -265,7 +267,8 @@ TEST(bpftrace, add_probes_wildcard_no_matches)
       "kprobe:sys_read,kprobe:not_here_*,kprobe:sys_write{}");
 
   auto bpftrace = get_strict_mock_bpftrace();
-  EXPECT_CALL(*bpftrace->mock_probe_matcher, get_symbols_from_traceable_funcs())
+  EXPECT_CALL(*bpftrace->mock_probe_matcher,
```
jordalgo, line 270: Worth adding a kfunc test to expect a `get_symbols_from_traceable_funcs(true)`

jordalgo:
One small test nit but LGTM. Nice clean up!

tests/bpftrace.cpp
viktormalik: Not really, there's too many things mocked in testing kfuncs and the test wouldn't have much value. OTOH, #2879 should also use `get_symbols_from_traceable_funcs(true)`, so the test can be added there.

viktormalik: Ah, no, I meant that they can be added as a part of that PR.

jordalgo: That was merged already (by you) :)

viktormalik: My GitHub comments today really suck :-D Sorry about that.

I meant to tag #2897 which I believe should call `get_symbols_from_traceable_funcs(true)` for attachment/listing of wildcarded `kprobe:module:function` probes. That should also add tests for this code path.

jordalgo: Haha no worries. All cleared up 👍

jordalgo (approved):

matthew-olson-intel:
This looks good to me, too; when looking at kprobe:module:function wildcards, I locally removed the whole `function [module]` syntax from the `kprobe` codepath, too, since it's no longer necessary.  Once this PR is merged, I'll merge into #2897.

danobi (approved):
not too familiar with this codepath but looks harmless enough. also well covered by tests

PR #2917, opened 2024-01-12
threadsnoop: use `strftime` instead of `elapsed`

makelinux:
because `elapsed` gives relative time,
and `strftime` gives absolute time which
is is more readable and practical.

##### Checklist


viktormalik (approved):

danobi (approved):
thanks!

PR #2919, opened 2024-01-13
docs: remove unneeded headers in example

chaiken:
The include statements cause the "Kernel Struct Tracing" example to fail.

##### Checklist

- [X ] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`

### DIFF
--- a/docs/tutorial_one_liners.md
+++ b/docs/tutorial_one_liners.md
@@ -296,6 +296,11 @@ The context of this probe is important: this fires when the I/O is issued to the
 
 ```
 # cat path.bt
+#ifndef BPFTRACE_HAVE_BTF
+#include <linux/path.h>
+#include <linux/dcache.h>
+#endif
+
 kprobe:vfs_open
 {
 	printf("open path: %s\n", str(((struct path *)arg0)->dentry->d_name.name));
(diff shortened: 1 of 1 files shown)


danobi:
Could you try enclosing the `#include`s in `#ifndef BPFTRACE_HAVE_BTF`? I am suspecting the error you're seeing is cuz your system has BTF

danobi:
Since it's the tutorial, this PR might make sense. Would like to keep it simple. And maybe enough systems have BTF now?

chaiken:
I added the ifdef as suggested.   The markdown HTML output is pretty borked, but it looks like the document is optimized for direct reading rather than HTML output.

viktormalik (approved):
It's true that a large portions of systems have BTF these days, but we're still getting issues reported from systems without BTF. That said, I'd keep the current version (with `#ifndef BPFTRACE_HAVE_BTF`) which nicely illustrates when the headers are needed.

viktormalik:
Should we also update Chinese and Japanese tutorials?

danobi (approved):
ill take a stab at updating chinese & japanese. i know a little chinese and some combination of google translate + chatgpt should help with japanese

Issue #2920, opened 2024-01-13
strerror not printing with quoted string for json format


jordalgo:
### What reproduces the bug? Provide code if possible.

```
kretfunc:__sys_bpf
{
        if (retval < 0)
        {
                print(("err", strerror(-retval)));
        }
}
```

`sudo bpftrace jordan.bt -f json`

### `bpftrace --info` output
```
{"type": "value", "data": ["err",Operation not permitted]}
```

viktormalik:
Fixed by @2921.

PR #2921, opened 2024-01-14
Fix: strerror string not quoted

jordalgo:
`sudo bpftrace -e 'kretfunc:__sys_bpf { if (retval < 0) { print((strerror(-retval))); } }' -f json`

Before:
{"type": "value", "data": No such file or directory}

After:
{"type": "value", "data": "No such file or directory"}

Fixes this issue: https://github.com/iovisor/bpftrace/issues/2920

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -11,7 +11,8 @@ bool is_quoted_type(const SizedType &ty)
 {
   return ty.IsKstackTy() || ty.IsUstackTy() || ty.IsKsymTy() || ty.IsUsymTy() ||
          ty.IsInetTy() || ty.IsUsernameTy() || ty.IsStringTy() ||
-         ty.IsBufferTy() || ty.IsProbeTy() || ty.IsCgroupPathTy();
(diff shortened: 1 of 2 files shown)


viktormalik (approved):

src/output.cpp
```diff
@@ -11,7 +11,8 @@ bool is_quoted_type(const SizedType &ty)
 {
   return ty.IsKstackTy() || ty.IsUstackTy() || ty.IsKsymTy() || ty.IsUsymTy() ||
          ty.IsInetTy() || ty.IsUsernameTy() || ty.IsStringTy() ||
-         ty.IsBufferTy() || ty.IsProbeTy() || ty.IsCgroupPathTy();
+         ty.IsBufferTy() || ty.IsProbeTy() || ty.IsCgroupPathTy() ||
+         ty.IsStrerrorTy();
```
danobi, line 15: Perhaps this could be changed to a switch/case where all the variants are explicitly listed
(without default). Then compiler would warn if new type is added without updating all sites.

jordalgo: Good idea! https://github.com/iovisor/bpftrace/pull/2925

PR #2922, opened 2024-01-15
Implement delete() with wildcards via for_each_elem

### LINKED ISSUE #2267: Filter on arbitrary level of map key
When you're collecting stacks that correspond to specific work items (e.g., workqueue items but other groupings naturally exist), you are tempted to do use `@[$itemid, kstack()]` for storage.

Unfortunately, the only way you can then print or delete this map is in its entirety. 

It would be great to have a primitive like `filter(@, 0, $itemid)` which returns `@` with the 0th level equal to `$itemid`. Then, one could `print` or `delete` that subset to better manage output and storage requirements.

This issue is related to #1470 but not the same. Batch operations require that you know the keys of all the elements you are trying to manipulate. The result of `filter` could be that subset of keys, so that subsequent `delete()` operations use batch ops underneath. 

That said, there's no real mechanism to iterate and discover all relevant keys and that's what's this issue is about.

I'm open to implementing this, if we can figure out the design. Filter expressions are relevant here though they feel like opening a bigger can of worms than strict equality matching.

viktormalik:
This implements a new syntax, `delete(@map[*, 10])` which allows partial deletes from maps, all within BPF. It uses `bpf_for_each_map_elem` helper with custom callback function.

It currently supports only integer values as concrete values.

Example usage:
```
@map[10, 10] = 1;
@map[20, 10] = 2;
@map[20, 1] = 3;
delete(@map[*, 10]);
// only @map[20, 1] is left here.
```

This is a re-submission of the last commit of #2321. The original PR contained a lot of commits for adding support for BPF subprograms (used for the custom callback). These were all merged separately in the meantime.

Resolves #2267.

##### Checklist


docs/reference_guide.md
```diff
@@ -3771,6 +3772,25 @@ it should.
...
+
+In more advanced usages, this allows for stack traces to be part of the key, while also allowing you to
+delete them at a future time. In the snippet below, we delete stacks that were taken for workqueue items,
+which finished in less than 10ms, while keeping all stacks for slower workqueue items.
+
+```
+# bpftrace -e 'tracepoint:workqueue:workqueue_execute_start { @work[tid] = nsecs; }
+    profile:ms:1 /@work[tid]/ { @stacks[tid, kstack] = count(); }
+    tracepoint:workqueue:workqueue_execute_end { if (nsecs - @work[tid] <= 10_000_000) {
+        delete(@stacks[tid, *]); } delete(@work[tid]); }'
```
danobi, line 3791: Perhaps a bit more readable if you only show contents of script and not use the one-liner invocation.

danobi (commented):
Only had time to day to think about high level - seems good to me. Hope to have time later this week/weekend to do a full review.
> This implements a new syntax, `delete(@map[*, 10])` which allows partial deletes from maps, all within BPF. It uses `bpf_for_each_map_elem` helper with custom callback function.
> 
> It currently supports only integer values as concrete values.
> 
> Example usage:
> ```
> @map[10, 10] = 1;
> @map[20, 10] = 2;
> @map[20, 1] = 3;
> delete(@map[*, 10]);
> // only @map[20, 1] is left here.
> ```
> 
> This is a re-submission of the last commit of #2321. The original PR contained a lot of commits for adding support for BPF subprograms (used for the custom callback). These were all merged separately in the meantime.
> 
> Resolves #2267.
> 
> 
> 
> ##### Checklist
> 
> - [x] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
> - [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
> - [x] The new behaviour is covered by tests
>

danobi:
Another thought: would this feature overlap with https://github.com/iovisor/bpftrace/issues/2955 ? For example, if we can loop over the key/value pairs of a map, would it also be possible to `delete()` entries in the body of the loop? If so, would this feature be needed?

viktormalik:
> Another thought: would this feature overlap with #2955 ? For example, if we can loop over the key/value pairs of a map, would it also be possible to `delete()` entries in the body of the loop? If so, would this feature be needed?

Good question. I think that #2955 could eventually support deleting the entries but I'd still argue that this syntax has its usage. After all it's much more convenient to write:
```
delete(@map[*, *, 100])
```
than
```
for ($key : @map) {
  if ($key.2 == 100) {
    delete(@map[$key]);
  }
}
```
In addition, it will take some time to get #2955 in and this seems to be an often requested capability.

ajor (changes requested):
Thanks again for resurrecting this old code - I know it wasn't you that wrote it in the first place! I've not fully gone over the semantic analyser and codegen changes, but have a few comments to start with.

One pretty major issue I've found it that this doesn't work! I haven't looked into why yet.

Delete doesn't delete:
```
$ sudo ./src/bpftrace -e 'BEGIN { @x[1,1] = 1; delete(@x[1,*]); exit(); }'
Attaching 1 probe...


@x[1, 1]: 1
```

The runtime tests should have caught this, but they are passing because they are not checking for the right thing. The `EXPECT` clause looks for a match, which it finds, and it just ignores the extra elements in the output that haven't been deleted.

danobi:
Another dimension to consider: since user-defined functions are close to merging, would it make sense to start writing a standard library? With map loops and user-defined functions, we could in theory keep the terseness of the new `delete()` overload without adding stuff to core language.

Since this PR does solve a concrete issue in the near term (the other efforts are medium term), I am still in favor of it. And it seems like it would be possible to have a migration path from moving stuff out of core language and into standard library.

src/ast/ast.h
```diff
@@ -259,11 +259,26 @@ class Map : public Expression {
...
 };
 
+class MapWildcard : public Expression
+{
+public:
+  DEFINE_ACCEPT
+  DEFINE_LEAFCOPY(MapWildcard)
+
+  explicit MapWildcard(location loc);
+  ~MapWildcard();
```
danobi, line 276: Can we `= default` the destructor?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3909,6 +3959,154 @@
...
+{
+  if (map_delete_filtered_cb_ != nullptr)
+    return map_delete_filtered_cb_;
+
+  /**
+   * The goal is to produce this code:
+   *
+   * struct ctx_t {
+   *   uint64_t offset;
+   *   uint64_t value;
```
danobi, line 3972: Perhaps good to call out somewhere in this function or the above that integers are always stored as
64bit values in maps. I see that mentioned in AssignMapStatement code but might be useful to point
it out again

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2492,6 +2553,13 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
   const std::string &map_ident = assignment.map->ident;
   auto type = assignment.expr->type;
 
+  if (assignment.map->wildcards)
+  {
+    LOG(ERROR, assignment.loc, err_)
+        << "Wildcards are only valid for delete() calls";
```
danobi, line 2559: Nit: perhaps a more maintainable error message is to say soemthing like "wildcards are not valid for assignments". So that if map wildcards gain more use cases, we don't need to update this messsage.

src/parser.yy
```diff
@@ -597,6 +598,12 @@ map_or_var:
                 var { $$ = $1; }
         |       map { $$ = $1; }
                 ;
+map_vargs:
+                map_vargs "," expr { $$ = $1; $1->push_back($3); }
+        |       map_vargs "," "*"  { $$ = $1; $1->push_back(new ast::MapWildcard(@$)); }
+        |       "*"            { $$ = new ast::ExpressionList; $$->push_back(new ast::MapWildcard(@$)); }
```
danobi, line 604: Should we use `MUL` instead of `*` here?

tests/runtime/delete_wildcard
```diff
@@ -0,0 +1,14 @@
+NAME delete_wildcard_idx_0
+PROG BEGIN { @map[10, 10] = 1; @map[2, 20] = 2; @map[10, 11] = 1; delete(@map[10, *]); print(@map); exit(); } END {}
+EXPECT @map\[2, 20\]: 2
```
danobi, line 3: This doesn't actually check that the other values are gone now, right? Cuz this would be on its own line. To check that the other entries are deleted, you'd want to probably use json output here.

tests/runtime/delete_wildcard
```diff
@@ -0,0 +1,14 @@
...
+TIMEOUT 5
+
+NAME delete_wildcard_idx_3
+PROG BEGIN { @map[10, 0, 1241, 0xcafe] = 1; @map[2, 20, 12, 0] = 2; @map[10, 192, 23, 0xcafe] = 1; delete(@map[*, *, *, 0xcafe]); print(@map); exit(); } END {}
+EXPECT @map\[2, 20, 12, 0\]: 2
+TIMEOUT 5
+
+NAME delete_wildcard_before_usage
+PROG END { delete(@map[10, *]); } BEGIN {@map[10, 20] = 1; @map[20, 10] = 2; exit();}
+EXPECT @map\[20, 10\]: 2
```
danobi, line 13: Same as above

tests/semantic_analyser.cpp
```diff
@@ -506,6 +507,23 @@ TEST(semantic_analyser, call_delete)
   test("kprobe:f { @x = 1; delete(@x) ? 0 : 1; }", 10);
 }
 
+TEST(semantic_analyser, call_delete_wildcard)
+{
+  test("kprobe:f { @x[0, 1] = 1; delete(@x[*, 1]); }", 0);
+  test("kprobe:f { @x[0, 1] = 1; $y = 1; delete(@x[* ,$y]); }", 0);
```
danobi, line 513: Add a successful test case for:

* trailing wildcard (ie `delete(@x[1, *])`)
* multiple wildcards (ie `delete(@x[*, 1, *])`)

tests/semantic_analyser.cpp
```diff
@@ -506,6 +507,23 @@
...
+  test("kprobe:f { @x[0, 1] = 1; @y = delete(@x[10, *]); }", 1);
+  test("kprobe:f { @x[0, 1] = 1; @[delete(@x[10, *])] = 1; }", 1);
+  test("kprobe:f { @x[0, 1] = 1; if(delete(@x[10, *])) { 123 } }", 10);
+  test("kprobe:f { @x[0, 1] = 1; delete(@x[10, *]) ? 0 : 1; }", 10);
```
danobi, line 517: Add a negative test case for lone wildcard (ie `delete(@x[*])`) - should error and say "use clear()` isntead ?

tests/semantic_analyser.cpp
```diff
@@ -506,6 +507,23 @@
...
+}
+
+TEST(semantic_analyser, map_wildcards)
+{
+  test("kprobe:f { @x[0, 1] = 1; @x[*, 1] = 2; }", 1);
+  test("kprobe:f { @x[0, 1] = 1; print(@x[*, 1]); }", 1);
+  test("kprobe:f { @x[0, 1] = 1; $y = @x[*, 1]; }", 10);
```
danobi, line 524: Also a test case for undeduced map type? ie a lone `delete(@[*, 1])` without any other usages to declare map type

src/parser.yy
danobi, line 1: Perhaps also add parser tests? Would be good to see some validation around negative test cases like:

```
delete(@["*"])
delete(@['*'])
delete(@[**])
```

viktormalik:
> One pretty major issue I've found it that this doesn't work! I haven't looked into why yet.
> 
> Delete doesn't delete:
> 
> ```
> $ sudo ./src/bpftrace -e 'BEGIN { @x[1,1] = 1; delete(@x[1,*]); exit(); }'
> Attaching 1 probe...
> 
> 
> @x[1, 1]: 1
> ```

Damn, stupid mistake in LLVM IR. And shame on me for not checking it manually and relying solely on the tests.

> The runtime tests should have caught this, but they are passing because they are not checking for the right thing. The `EXPECT` clause looks for a match, which it finds, and it just ignores the extra elements in the output that haven't been deleted.

This is not a good behaviour. I'm thinking about introducing a new directive `EXPECT_NOT` for checking that the given regex is *not* in the output.

tests/runtime/delete_wildcard
viktormalik: That's correct. We could either use JSON output or I'm thinking about adding a new `EXPECT_NOT` clause to runtime tests to specify regex which should not be present in the output. But using JSON is probably better.

src/parser.yy
viktormalik: The first one will parse successfully as `"*"` is a string which is a valid map key. The latter two should fail so I'll add tests for them.

viktormalik:
Thanks for the reviews @ajor and @danobi! I believe I addressed all of your comments (except for adding semantic tests for particular error messages). I didn't resolve any comments so that you can double check.

I added a bunch of new semantic analyser and parser tests and changed runtime tests to use json output to be able to verify that map entries are truly missing.

danobi:
codeql warning looks legit. Seems `is_quoted_type()` is missing map_wildcard

docs/reference_guide.md
danobi, line 1: missing man page update? I (maybe we?) should really finish that conversion...

viktormalik: Yeah, you're right, I updated it. This is really annoying though, we should make the conversion a priority. Let's discuss at the office hours.

viktormalik:
clang-format failure is fixed by #2978.

danobi:
Do we still need this now that looping over map elements is supported?

ajor:
Our users in Meta have started using the for-loop syntax now, so this isn't a priority for them.

The syntax in this PR is definitely nicer, but there are two concerns:
- From users: it hides the `O(n)` iteration over every element of the map
- From me: the current implementation opens up some big holes in semantic analysis 

I wouldn't say it's worth continuing with from my point of view.

viktormalik:
Agreed. Let's close this one and we may reiterate on the nicer syntax in future (maybe along with adding other map "batch" operations).

PR #2925, opened 2024-01-15
[output] Use a switch for 'is_quoted_type'

jordalgo:
Follow up to: https://github.com/iovisor/bpftrace/pull/2921

This will give us a compiler warning if we missed any types.

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -17,10 +17,41 @@ namespace bpftrace {
 namespace {
 bool is_quoted_type(const SizedType &ty)
 {
-  return ty.IsKstackTy() || ty.IsUstackTy() || ty.IsKsymTy() || ty.IsUsymTy() ||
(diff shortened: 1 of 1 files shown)


src/output.cpp
```diff
@@ -17,10 +17,41 @@ namespace bpftrace {
...
+    case Type::record:
+    case Type::stack_mode:
+    case Type::stats:
+    case Type::sum:
+    case Type::timestamp:
+    case Type::timestamp_mode:
+    case Type::tuple:
+      return false;
+  }
+  return false;
```
ajor, line 54: Is this duplicate return necessary?

ajor (approved):
Nice!

src/output.cpp
jordalgo: I didn't think so but the compiler complained "warning: control reaches end of non-void function [-Wreturn-type]"

viktormalik (approved):

PR #2926, opened 2024-01-16
remove btf related fd leaks in AttachedProbe::load_prog()

tyroguru:
Currently, `AttachedProbe::load_prog()` creates a pair of descriptors related to btf information when loading each BPF program. These descriptors are not subsequently required and can safely be closed after the load call has completed.

Without the fix:

```
$ cat ~/BPFTrace/fentry.bt
fentry:b*
{
  @[func] = count();
}
$ BPFTRACE_MAX_BPF_PROGS=10000 BPFTRACE_MAX_PROBES=10000 sudo src/bpftrace ~/BPFTrace/fentry.bt
Attaching 3486 probes..
<chop>

$ sudo ls -l /proc/`pgrep bpftrace`/fd | wc -l
13966
```

After the fix:

```

$ sudo ls -l /proc/`pgrep bpftrace`/fd | wc -l
6960
```

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -311,8 +311,6 @@ AttachedProbe::~AttachedProbe()
       LOG(FATAL) << "invalid attached probe type \""
                  << probetypeName(probe_.type) << "\" at destructor";
   }
-  if (btf_fd_ != -1)
-    close(btf_fd_);
 
   if (err)
     LOG(ERROR) << "failed to detach probe: " << probe_.name;
@@ -816,9 +814,9 @@ void AttachedProbe::load_prog(BPFfeature &feature)
                     .log_size = static_cast<__u32>(log_buf_size), );
 
         auto &btf = prog_.getBTF();
-        btf_fd_ = bpf_btf_load(btf.data(), btf.size(), &btf_opts);
+        auto btf_fd = bpf_btf_load(btf.data(), btf.size(), &btf_opts);
 
-        opts.prog_btf_fd = btf_fd_;
+        opts.prog_btf_fd = btf_fd;
 
         if (!func_infos.empty())
         {
@@ -830,7 +828,8 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         // Don't attempt to load the program if the BTF load failed.
         // This will fall back to the error handling for failed program load,
         // which is more robust.
-        if (btf_fd_ >= 0)
+        if (btf_fd >= 0)
+        {
           progfd_ = bpf_prog_load(static_cast<::bpf_prog_type>(prog_type),
                                   name.c_str(),
                                   license,
@@ -838,10 +837,12 @@ void AttachedProbe::load_prog(BPFfeature &feature)
                                       insns.data()),
                                   insns.size() / sizeof(struct bpf_insn),
                                   &opts);
+          close(btf_fd);
+        }
       }
 
-      if (opts.attach_prog_fd > 0)
-        close(opts.attach_prog_fd);
+      if (opts.attach_btf_obj_fd > 0)
+        close(opts.attach_btf_obj_fd);
       if (progfd_ >= 0)
         break;
     }
--- a/src/attached_probe.h
+++ b/src/attached_probe.h
@@ -89,7 +89,6 @@ private:
   int progfd_ = -1;
   uint64_t offset_ = 0;
   int tracing_fd_ = -1;
-  int btf_fd_ = -1;
   std::function<void()> usdt_destructor_;
 
   BTF &btf_;


jordalgo:
LGTM. Just needs some formatting love. I do `git clang-format upstream/master`

src/attached_probe.cpp
```diff
@@ -842,6 +842,12 @@ void AttachedProbe::load_prog(BPFfeature &feature)
 
       if (opts.attach_prog_fd > 0)
         close(opts.attach_prog_fd);
+      if (btf_fd_ >= 0) {
+        close(btf_fd_);
+        btf_fd_ = -1;
+      }
```
danobi, line 848: How about moving this to a local variable? Seems like `btf_fd_` is unused outside this function, save for closing in destructor.

danobi (commented):
thanks jon! one nit and, yeah, please fix formatting

src/attached_probe.cpp
tyroguru: Thanks @jordalgo and @danobi for the review.

Daniel, you are absolutely correct and that's me being lazy - I'll convert it to a local variable as there's no need for it to be a class member.

Re: formatting. In my defence ( :-) ) I was using the signal I got from my own branch before I created the PR and that was all green. @ajor just looked and formatting signal is only provided on PR's and not for pushes to private branches. Could a clang-format check be added to the branch pushes as this would stop thsi kind off issue?

ajor: Derailing the PR a little, but specifically this part of the CI config: https://github.com/iovisor/bpftrace/blob/20944038eef747e3c6255edb178db66bbd72737e/.github/workflows/ci.yml#L11-L14

I don't understand what the purpose is - can we remove it?

tyroguru:
> thanks jon! one nit and, yeah, please fix formatting

@danobi - fixed it as I should have done the first time (local variable) :-) . Tests out fine. Thanks!

danobi (approved):
thanks!

btw, was this causing issues somewhere? if so, might be nice to add a changelog entry. if it's just an optimization, can probably skip changelog

src/attached_probe.cpp
danobi: I think if you take that out then the job will start running on master branch. Then you'd see failures if people chose to ignore clang-format. If we are ok with making it a strict requirement (cuz in practice it is), we can delete the gate.

src/attached_probe.cpp
```diff
@@ -830,18 +827,23 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
                                   insns.size() / sizeof(struct bpf_insn),
                                   &opts);
+          close(btf_fd);
+        }
       }
 
       if (opts.attach_prog_fd > 0)
         close(opts.attach_prog_fd);
+      if (opts.attach_btf_obj_fd > 0)
+        close(opts.attach_btf_obj_fd);
```
viktormalik, line 846: It seems like we were closing the wrong fd here. Can we remove the closing of `opts.attach_prog_fd` as we don't seem to set that anywhere?

viktormalik (commented):
Good catch, thanks!

tyroguru:
> thanks!
> 
> btw, was this causing issues somewhere? if so, might be nice to add a changelog entry. if it's just an optimization, can probably skip changelog

bpftrace's usage of fd's is excessive and it causes issues when enabling large(r) amounts of probes (you run into EMFILE problems). This small fix only addresses the easy ones and I think more work is needed. This is just one of those "quality of life" things but more is needed in this area. Not sure it warrants a changelog entry but obviously willing to take input on that.

src/attached_probe.cpp
tyroguru: @viktormalik, just to clarify: the `opts.attach_btf_obj_fd` cleanup is required else we leak fd's for kfunc, kretprobe and iter probes. I agree, the existing `opts.attach_prog_fd ` close is not needed and I've no removed that. It all tests out fine locally but will need to wait for the CI to turn green.

viktormalik:
> > thanks!
> > btw, was this causing issues somewhere? if so, might be nice to add a changelog entry. if it's just an optimization, can probably skip changelog
> 
> bpftrace's usage of fd's is excessive and it causes issues when enabling large(r) amounts of probes (you run into EMFILE problems). This small fix only addresses the easy ones and I think more work is needed. This is just one of those "quality of life" things but more is needed in this area. Not sure it warrants a changelog entry but obviously willing to take input on that.

No need to make a changelog entry IMHO.

GitHub shows a merge conflict, though, could you rebase your branch onto master (and ideally drop the second commit)? Thanks!

src/attached_probe.cpp
ajor: I think it'd allow bad formatting on master anyway, because it only formats the diff between the current commit and master (which would have zero lines when on master). Opened: https://github.com/iovisor/bpftrace/pull/2931

src/attached_probe.cpp
jordalgo: For my own knowledge, what normally closes this file descriptor (in other bpf contexts)?

jordalgo (approved):

PR #2928, opened 2024-01-17
threadsnoop: increase PID field width and align to the right

makelinux:
for better readability.

Justifications:

https://ux.stackexchange.com/questions/13795/is-there-a-standard-to-left-justify-text-and-right-justify-numeric-values

wc -L /proc/sys/kernel/pid_max
7 /proc/sys/kernel/pid_max

##### Checklist


danobi (approved):
seems reasonable

viktormalik (approved):
Agreed, thanks!

PR #2929, opened 2024-01-17
CMake: Don't trigger full rebuild when a test file changes

ajor:
Swap out configure_file() for custom copy commands.

configure_file forces a CMake reconfiguration if it detects changes in the target file. This new approach sets the dependencies for test files so that only the necessary targets are rebuilt.

##### Checklist

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -145,36 +145,49 @@ foreach(testlib_source ${testlibs})
 endforeach()
 add_custom_target(testlibs DEPENDS ${compiled_testlibs})
 
+
+#
+# Runtime Tests
+#
 configure_file(runtime-tests.sh runtime-tests.sh COPYONLY)
-file(GLOB runtime_tests runtime/*)
-list(REMOVE_ITEM runtime_tests ${CMAKE_CURRENT_SOURCE_DIR}/runtime/engine)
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -145,36 +145,49 @@ foreach(testlib_source ${testlibs})
...
+  DEPENDS
+    ${compiled_testprogs}
+    ${compiled_testlibs}
+    ${CMAKE_BINARY_DIR}/src/bpftrace
+)
+add_test(NAME runtime_tests COMMAND ./runtime-tests.sh)
+
+file(GLOB_RECURSE runtime_test_files
+  RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
+  CONFIGURE_DEPENDS
```
danobi, line 165: This seems to work but in the docs (https://cmake.org/cmake/help/latest/command/file.html#filesystem) it warns about using glob

ajor: Yeah I saw that, but since we've been using `GLOB` all over the place for years then this is not a step backwards at least, and `CONFIGURE_DEPENDS` makes it an improvement with most generators.

The other (recommended) option is to explicitly list every source file in a CMakeLists.txt, which sounds like a lot of work 😅

PR #2930, opened 2024-01-17
CMake: Fix testlib dependencies

ajor:
The "testlibs" target previously depended on "compiled_testlibs", but there was no rule which told CMake how to generate these compiled testlibs from their output names.

Now we make "testlibs" depend directly on the individual testlib targets instead of their outputs.

Repro:
1. Start with a completely clean build directory: `sudo rm -rf * && cmake ../`
2. `make testlibs`

Previous output:
```
No rule to make target 'tests/testlibs/libsimple.so', needed by 'tests/CMakeFiles/testlibs'.  Stop.
```

Also made similar cleanups to testprogs, but they were not causing the same issues as testprog binary names match their CMake target names.

### DIFF
--- a/tests/testlibs/CMakeLists.txt
+++ b/tests/testlibs/CMakeLists.txt
@@ -1,0 +1,20 @@
+file(GLOB testlib_sources CONFIGURE_DEPENDS *.c)
+set(testlibtargets "")
+foreach(testlib_source ${testlib_sources})
+  get_filename_component(testlib_name ${testlib_source} NAME_WE)
+  add_library(${testlib_name} SHARED ${testlib_source})
+  set_target_properties(${testlib_name}
+    PROPERTIES
+      LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+      COMPILE_FLAGS "-g -O0")
+  if(HAVE_SYSTEMTAP_SYS_SDT_H)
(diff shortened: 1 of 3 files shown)


tests/testlibs/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+    PROPERTIES
+      LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+      COMPILE_FLAGS "-g -O0")
+  if(HAVE_SYSTEMTAP_SYS_SDT_H)
+    target_compile_definitions(${testlib_name} PRIVATE HAVE_SYSTEMTAP_SYS_SDT_H)
+  endif(HAVE_SYSTEMTAP_SYS_SDT_H)
+  # clear the executable bit
+  add_custom_command(TARGET ${testlib_name}
+    POST_BUILD
+    COMMAND chmod -x $<TARGET_FILE_NAME:${testlib_name}>)
```
ajor, line 16: I'm not sure what the purpose of this is - tempted to remove it unless anyone knows?

danobi: https://github.com/iovisor/bpftrace/commit/c3d1095a44fb51 has details on it. Looks like it's there to test a bug where +x shared libraries were not handled

danobi (approved):

PR #2931, opened 2024-01-17
CI: Run clang-format on all branches

ajor:
This will let it run on branches of forks so people can catch formatting errors before they open a PR.

I don't think this will cause failures on master even with bad formatting, as `git clang-format origin/master` will always return with "no modified files to format" when on origin/master.


jordalgo:
Let's see if we hit our CI quota :)

jordalgo (approved):

PR #2935, opened 2024-01-18
Runtime tests: Adjust meaning of BPFTRACE_RUNTIME_TEST_EXECUTABLE env var

ajor:
The previous usage was misleading - it appeared to accept a path to an executable, but actually expected a directory.

Passing in binary paths seems more flexible than directories (in case a different build system is used), so I changed the variable to expect this and kept the old name.


danobi (approved):
nice, thx

PR #2941, opened 2024-01-22
Emit maps in libbpf style (with BTF)

viktormalik:
This is another step towards delegating loading and attachment of BPF programs to libbpf. libbpf requires maps to be emitted in certain format with attached BTF information.
    
The expected format is that each map is represented by a global variable in the `.maps` section. The variable should be of a struct type with the following fields:
- `type` - map type (e.g. BPF_MAP_TYPE_HASH),
- `max_entries` - maximum number of entries,
- `key` - key type,
- `value` - value type.

The exception is the `BPF_MAP_TYPE_RINGBUF` map type which doesn't have the `key` and `value` entries.
    
The variable and its type should have appropriate BTF entries. This is done by emitting DWARF debugging information for the map which LLVM converts to BTF when emitting the program.
    
It is also necessary to fixup BTF emitted by LLVM b/c it's missing some information. This is usually done by libbpf's linker but since we're passing BTF to the BPF program directly (instead of using `bpf_object`), we must perform this step manually.

Note: there is also a change in codegen tests for `kstack` and `ustack`. #2910 updated the names of stack maps to include the stack mode and limit to be able to differentiate distinct maps by name. Since the name is now a part of codegen (it's the name of the emitted variable), it's no longer true that `kstack()` and `kstack(perf)` generate the same codegen.

##### Checklist


tests/codegen/call_kstack.cpp
```diff
@@ -10,12 +10,7 @@ TEST(codegen, call_kstack)
 {
   auto result = NAME;
 
-  // Mode doesn't directly affect codegen, so the programs below should
-  // generate the same IR
-  test("kprobe:f { @x = kstack(); @y = kstack(6) }", result);
-  test("kprobe:f { @x = kstack(perf); @y = kstack(perf, 6) }", result);
-  test("kprobe:f { @x = kstack(raw); @y = kstack(raw, 6) }", result);
```
jordalgo, line 17: curious, why remove the checks for the different stack modes?

src/ast/dibuilderbpf.h
```diff
@@ -14,17 +23,42 @@ class DIBuilderBPF : public DIBuilder
...
 private:
   struct
   {
+    DIType *int8 = nullptr;
+    DIType *int16 = nullptr;
+    DIType *int32 = nullptr;
     DIType *int64 = nullptr;
+    DIType *int128 = nullptr;
     DIType *int8_ptr = nullptr;
+    DIType *int_ = nullptr;
```
jordalgo, line 58: Why the need for this private `int_` rather than just using `int32`?

src/ast/dibuilderbpf.cpp
```diff
@@ -59,14 +86,182 @@ DIType *DIBuilderBPF::getInt64Ty()
...
   if (!types_.int8_ptr)
-    types_.int8_ptr = createPointerType(
-        createBasicType("int8", 8, dwarf::DW_ATE_signed), 64);
+    types_.int8_ptr = createPointerType(getInt8Ty(), 64);
 
   return types_.int8_ptr;
 }
 
+// Create anonymous struct with anonymous fields. It's possible that there will
+// be multiple tuples of the same (duplicated) type but BTF deduplication should
```
jordalgo, line 114: If it's not deduped, will it go over the instruction limit with duplicated types or does that not count toward that?

src/ast/dibuilderbpf.cpp
```diff
@@ -59,14 +86,182 @@ DIType *DIBuilderBPF::getInt64Ty()
...
+                                             8,
+                                             DINode::FlagZero,
+                                             nullptr,
+                                             getOrCreateArray(fields));
+  return result;
+}
+
+DIType *DIBuilderBPF::GetType(const SizedType &stype)
+{
+  if (stype.IsByteArray() || stype.IsRecordTy())
```
jordalgo, line 147: nit: would a switch be nicer here?

src/ast/dibuilderbpf.cpp
```diff
@@ -59,14 +86,182 @@ DIType *DIBuilderBPF::getInt64Ty()
...
+  }
+}
+
+DIType *DIBuilderBPF::GetMapKeyType(const MapKey &key,
+                                    const SizedType &value_type)
+{
+  // No-key maps use '0' as the key.
+  // No-key count() maps use BPF_MAP_TYPE_PERCPU_ARRAY which needs 4-byte key.
+  if (key.args_.size() == 0)
+    return value_type.IsCountTy() ? getInt32Ty() : getInt64Ty();
```
jordalgo, line 194: Just curious, why do the other no-key maps need a 64 int?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,201 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+{
+  std::string var_name = name;
+  if (var_name[0] == '@')
+    var_name = "AT_" + var_name.substr(1);
+
+  auto debuginfo = debug_.createMapEntry(
+      var_name, map_type, max_entries, key, value_type);
+
+  // It's sufficient that the global variable has the correct size (a struct
+  // with one pointer per each field). The actual inner types are defined in
```
jordalgo, line 3552: super nit
```suggestion
  // with one pointer per field). The actual inner types are defined in
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,201 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+// can be read and used by libbpf.
+//
+// Each map should be defined by a global variable of a struct type with the
+// following fields:
+// - "type"        map type (e.g. BPF_MAP_TYPE_HASH)
+// - "max_entries" maximum number of entries
+// - "key"         key type
+// - "value"       value type
+//
+// "type" and "max_entries" are integers but they must be represented as
```
jordalgo, line 3581: interesting - this is how libbpf does this under the hood?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,201 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+// Note that LLVM will generate BTF which misses some information. This is
+// normally set by libbpf's linker but since we load BTF directly, we must do
+// the fixing ourselves, until we start loading BPF programs via bpf_object.
+// See BpfBytecode::fixupBTF for details.
+void CodegenLLVM::generate_maps(const RequiredResources &resources)
+{
+  // User-defined maps
+  for (auto &map_val : resources.map_vals)
+  {
+    const std::string name = map_val.first;
```
jordalgo, line 3601: nit
```suggestion
    const std::string& name = map_val.first;
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,201 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+
+    SizedType val_type = map_val.second;
+    auto key = resources.map_keys.find(name);
```
jordalgo, line 3604: nit
```suggestion
    const auto& key = resources.map_keys.find(name);
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,201 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+    if (val_type.IsCountTy() && key->second.args_.empty())
+    {
+      max_entries = 1;
+      map_type = libbpf::BPF_MAP_TYPE_PERCPU_ARRAY;
+    }
+    else if ((val_type.IsHistTy() || val_type.IsLhistTy() ||
+              val_type.IsCountTy() || val_type.IsSumTy() ||
+              val_type.IsMinTy() || val_type.IsMaxTy() || val_type.IsAvgTy() ||
+              val_type.IsStatsTy()) &&
+             bpftrace_.feature_->has_map_percpu_hash())
```
jordalgo, line 3619: super nit: maybe move this feature check to front of this list of conditionals

tests/codegen/call_kstack.cpp
viktormalik: Because these now generate different LLVM IR (they used to generate the same) so we'd need to keep separate `.ll` files for then. It's better explained in the commit message.

src/ast/dibuilderbpf.h
viktormalik: Hm, good question. This is used for entries for integer fields of a map entry (`type` and `max_entries`) which libbpf expects to be pointers to an array of `int` (not `int32`). It's possible that it would work with `int32`, too, but I was trying to stay as close to libbpf's behaviour as possible.

src/ast/dibuilderbpf.cpp
viktormalik: I don't think that this has anything to do with instruction limit. This is just a DWARF entry which will be transformed into a BTF entry.

src/ast/dibuilderbpf.cpp
viktormalik: That's probably a good idea, but this is the way we do it everywhere else. Turning these into switches could be useful to get compiler checks but it can introduce some problems so I'd rather do that in a separate PR.

src/ast/dibuilderbpf.cpp
viktormalik: Normally, we use 64-bit ints everywhere, just BPF_MAP_TYPE_PERCPU_ARRAY needs 32-bit keys [per documentation](https://docs.kernel.org/bpf/map_array.html).

src/ast/passes/codegen_llvm.cpp
viktormalik: Indeed: https://github.com/libbpf/libbpf/blob/master/src/libbpf.c#L2163-L2169.

tests/codegen/call_kstack.cpp
viktormalik: Ah, I see. It's probably an overkill, the only change will be in names of global variables for maps.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,200 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+      map_type = libbpf::BPF_MAP_TYPE_HASH;
+    }
+
+    createMapDefinition(name, map_type, max_entries, key->second, val_type);
+  }
+
+  // bpftrace internal maps
+  for (const StackType &stack_type : resources.stackid_maps)
+  {
+    // bpf_get_stackid() is kind of broken by design.
```
jordalgo, line 3633: @viktormalik I had it on my internal todos to replace get_stackid but forgot to write up an issue until just now :) 

I stole some of your text but at least now you can replace this large comment with just a link to the issue:
https://github.com/iovisor/bpftrace/issues/2962

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3528,11 +3529,200 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+    SizedType value_type = CreateArray(value_size, CreateInt8());
+    createMapDefinition(to_string(MapManager::Type::Join),
+                        libbpf::BPF_MAP_TYPE_PERCPU_ARRAY,
+                        1,
+                        MapKey({ CreateInt32() }),
+                        value_type);
+  }
+
+  if (resources.needs_elapsed_map)
+  {
```
jordalgo, line 3674: Are we following the kernel convention of removing the brackets for single lines after conditionals?

jordalgo (commented):
The code looks good but I don't know this stuff super well yet so probably be good to get other folks to approve.

src/ast/passes/codegen_llvm.cpp
viktormalik: Thanks for creating the issue! The text is not mine, I just copied it from other place (where it will be removed in one of the future PRs).

viktormalik: Good idea on linking the issue, though, will do that!

src/ast/passes/codegen_llvm.cpp
viktormalik: I don't think that we have a rule for that. I generally omit the brackets if there's just a single line (in this case, there're multiple lines, although it's just one function call, so I'd leave the brackets).

ajor (commented):
Other than a few nits, the code looks good to me.

It'd be nice if there was a single place where we defined everything about a map (key sizes, map type etc.), but I see that it's been like this for a while with Map/IMap.

I think being split across multiple files it makes it harder to follow the logic unless you're intimately familiar with it (e.g. 0-key count using a per-cpu-array and hence needing a 4-byte key) and I'm worried that bugs could be introduced if they become desynchronised in the future. But seeing as it's been like this for a long time, let's not bother in this PR. I might have another think about it later this week.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3268,11 +3269,188 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+// normally set by libbpf's linker but since we load BTF directly, we must do
+// the fixing ourselves, until we start loading BPF programs via bpf_object.
+// See BpfBytecode::fixupBTF for details.
+void CodegenLLVM::generate_maps(const RequiredResources &resources)
+{
+  // User-defined maps
+  for (auto &map_val : resources.map_vals) {
+    const std::string &name = map_val.first;
+
+    SizedType val_type = map_val.second;
```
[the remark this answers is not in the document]
viktormalik: I think it's only for copying in function params... Good catch anyways, thanks!

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3268,11 +3269,188 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int &id)
...
+  if (!bpftrace_.feature_->has_map_ringbuf() ||
+      resources.needs_perf_event_map) {
+    createMapDefinition(to_string(MapManager::Type::PerfEvent),
+                        libbpf::BPF_MAP_TYPE_PERF_EVENT_ARRAY,
+                        get_online_cpus().size(),
+                        MapKey({ CreateInt32() }),
+                        CreateInt32());
+  }
+
+  if (bpftrace_.feature_->has_map_ringbuf()) {
```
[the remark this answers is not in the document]
viktormalik: IIUC, there are cases when we need both.

src/ast/dibuilderbpf.cpp
```diff
@@ -58,14 +85,177 @@ DIType *DIBuilderBPF::getInt64Ty()
...
+
+  // Some map types need an extra 8-byte key.
+  uint64_t extra_arg_size = 0;
+  if (value_type.IsHistTy() || value_type.IsLhistTy() || value_type.IsAvgTy() ||
+      value_type.IsStatsTy())
+    extra_arg_size = 8;
+
+  // Single map key -> use the appropriate type.
+  if (key.args_.size() == 1 && extra_arg_size == 0)
+    return GetType(key.args_[0]);
```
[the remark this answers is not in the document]
viktormalik: Yes, I believe that the LLVM types for keys and values are not important except for their size (at least not for our current usages of maps). So, at the moment it's for debugging only and to keep the LLVM IR as close to our internal type representation as possible.

src/ast/dibuilderbpf.cpp
```diff
@@ -58,14 +85,177 @@ DIType *DIBuilderBPF::getInt64Ty()
...
+                                      field.offset * 8,
+                                      DINode::FlagZero,
+                                      GetType(field.type)));
+  }
+  DICompositeType *result = createStructType(file,
+                                             "",
+                                             file,
+                                             0,
+                                             stype.GetSize() * 8,
+                                             8,
```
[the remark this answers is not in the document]
viktormalik: Shouldn't be necessary, I'll change it to 0.

src/ast/dibuilderbpf.cpp
```diff
@@ -58,14 +85,177 @@ DIType *DIBuilderBPF::getInt64Ty()
   return types_.int64;
 }
 
+DIType *DIBuilderBPF::getInt128Ty()
```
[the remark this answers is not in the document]
viktormalik: Yes

viktormalik:
> Other than a few nits, the code looks good to me.

Thanks for the review! I've addressed or at least commented all of your comments.

> 
> It'd be nice if there was a single place where we defined everything about a map (key sizes, map type etc.), but I see that it's been like this for a while with Map/IMap.
> 
> I think being split across multiple files it makes it harder to follow the logic unless you're intimately familiar with it (e.g. 0-key count using a per-cpu-array and hence needing a 4-byte key) and I'm worried that bugs could be introduced if they become desynchronised in the future. But seeing as it's been like this for a long time, let's not bother in this PR. I might have another think about it later this week.

Agreed. My ongoing work is going to make this much better, almost all map settings are going to be in `CodegenLLVM::generate_maps`. Some things will still need to be set in other places but overall the logic should be much easier to follow.

ajor (approved):
Looks good!

PR #2942, opened 2024-01-22
v0.20.0 release

viktormalik:


fbs (approved):

jordalgo (approved):

ajor (approved):

ajor:
Just noticed that CI for the Fedora build is failing on master - we should probably fix that before the release to confirm that bpftrace can still actually be built on Fedora.

ajor:
Good to know it's not our fault then, but the reason I'm a little concerned is that our Fedora build did break last week despite the other distros still working: https://github.com/iovisor/bpftrace/pull/2940

As long as bpftrace still builds on Fedora it's fine - just don't want to end up in a situation where a release doesn't compile

danobi (approved):
:fireworks:

viktormalik:
Turns out that the Fedora issue is in a package which we don't need at all. I updated our Dockerfile to only install the minimal packages necessary, see #2945. I'd like to merge that (and run CI) before doing the release.

PR #2944, opened 2024-01-23
explicitly use ast namespace when including files

chantra:
##### Checklist

##### Why

a couple of files from ast directory were included using the root namespace. To make this work, CMakeLists.txt needed to explicitely:
```
target_include_directories(runtime PRIVATE ${CMAKE_SOURCE_DIR}/src/ast)
```

This change update the couple of .cpp files that included ast header files from the root namespace to use the ast prefix, and remove the `target_include_directories` entries.

While it does not change anything fundamentally to how bpftrace is build using cmake, it does make building with other build systems such as buck, bazel... easier.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -1,8 +1,9 @@
 #include "output.h"
+
+#include "ast/async_event_types.h"
 #include "bpftrace.h"
 #include "log.h"
 #include "utils.h"
-#include <async_event_types.h>
 
 #include <bpf/libbpf.h>
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -113,7 +113,6 @@ add_definitions("-DBPFTRACE_VERSION=\"${BPFTRACE_VERSION}\"")
 
 target_include_directories(runtime PRIVATE ${CMAKE_BINARY_DIR})
 target_include_directories(runtime PRIVATE ${CMAKE_SOURCE_DIR}/src)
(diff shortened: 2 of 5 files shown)


jordalgo (approved):
Thanks for the fix. LGTM!

chantra:
got to fix a few lint errors...

src/output.cpp
```diff
@@ -2,7 +2,7 @@
 #include "bpftrace.h"
 #include "log.h"
 #include "utils.h"
-#include <async_event_types.h>
+#include <ast/async_event_types.h>
```
viktormalik, line 5:
```suggestion
#include "ast/async_event_types.h"
```
and order alphabeticlaly

viktormalik (commented):
Nice change, thanks!

danobi (approved):
clang-format doesn't look quite right... maybe ignore it in this case

chantra:
> clang-format doesn't look quite right... maybe ignore it in this case

@danobi are you referring to the fact that "output.h" was put at the top? https://github.com/iovisor/bpftrace/actions/runs/7629941448/job/20784517389#step:8:10

I think it does make sense. It signals that the files' `.h` should be first. I the updated diff, I have added an empty line between "output.h" and the block of includes so sorting is done within the block.

danobi:
D'oh, didn't realize that was in output.cpp. Tool is pretty clever :)

PR #2946, opened 2024-01-23
CMake: Use FetchContent instead of ExternalProject for vendored gtest

ajor:
FetchContent is the newer and better way of importing external code.

This also fixes a dependency issue with the old ExternalProject rules, where the gtest_byproducts were not set up correctly.

Bumped the minimum required CMake version to enable `FetchContent_MakeAvailable`. Looking at a range of distros, almost everything ships 3.20+ at this point anyway: https://pkgs.org/download/cmake


ajor:
Huh maybe you're right. I've been using for my own local development for a while but it seems to work without too now. I guess my distro updated gtest underneath me.

PR #2950, opened 2024-01-23
build: Use `llvm-config.h` to find out what LLVM_VERSION we are compiling against

chantra:
Some source files were using the values of LLVM_VERSION_* past over using a `-D` flag.
This change remove the definition addition of LLVM_VERSION_* from CMakeLists.txt and instead uses the header `llvm/Config/llvm-config.h` provided by LLVM.

This allow to build bpfdtrace using other build systems without having to re-implement the logic of llvm version detection/definition.


chantra:
clang_parser.nested_struct_no_type is legitimately failing... putting this back as a draft.

jordalgo:
LGTM!

jordalgo (approved):

viktormalik (approved):
Nice!

PR #2957, opened 2024-01-24
docs: Fix various reference_guide.md links

nick-child-ibm:
Sorry, pedantic PR.
Was following along reference_guide.md and noticed some hyperlinks were dead.

Issue #912, opened 2019-09-26
Move `ast.h` class definitions into `ast.cpp`


danobi:
We do inline definitions in a couple places but that number is starting to grow. We should really define these methods in the .cpp so it doesn't take forever to recompile bpftrace if you modify the ast code.

See `src/ast/ast.h` and `src/ast/ast.cpp`.

vladdy:
Can I grab this one?

danobi:
Sorry, taking this one. Working on #1047 and it made sense to do this first.

vladdy:
Not a problem. Thank you!

PR #2958, opened 2024-01-25
Add 'lazy_symbolication' config

jordalgo:
This can be used to lazily symbolize userspace symbols instead of loading the entire symbol table ahead of time.

This is useful for when tracing large binaries.

##### Checklist

### DIFF
--- a/docs/reference_guide.md
+++ b/docs/reference_guide.md
@@ -175,6 +175,7 @@ ENVIRONMENT:
     BPFTRACE_DEBUG_OUTPUT               [default: 0] outputs bpftrace's runtime debug messages to the trace_pipe
     BPFTRACE_KERNEL_BUILD               [default: /lib/modules/$(uname -r)] kernel build directory
     BPFTRACE_KERNEL_SOURCE              [default: /lib/modules/$(uname -r)] kernel headers directory
+    BPFTRACE_LAZY_SYMBOLICATION         [default: 0] symbolicate lazily/on-demand
     BPFTRACE_LOG_SIZE                   [default: 1000000] log size in bytes
     BPFTRACE_MAX_BPF_PROGS              [default: 512] max number of generated BPF programs
     BPFTRACE_MAX_CAT_BYTES              [default: 10k] maximum bytes read by cat builtin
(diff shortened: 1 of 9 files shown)


jordalgo:
not sure how to test this - it seems that even setting lazy_symbolize to numbers other than 1 or 0 don't cause any "noticeable" issues - maybe the errors get swallowed.

jordalgo:
The other motivation for this change is so Meta doesn't have to supply a patch to hard code this value to 1 due to the size of some of the binaries that is traces against.

viktormalik:
Should we also add this as a config option?

jordalgo:
> Should we also add this as a config option?

I started down that path but because this is more specific to the environment or binary target (so yeah maybe a case for specific uprobes), I thought it better as just an env var but I don't feel strongly.

ajor:
I think "symbolication" is the more accepted term over "symbolization". Some quick searching backs this up, e.g. https://lldb.llvm.org/use/symbolication.html

I'm also biased because I don't like the American spelling of "symbolisation" ;)

jordalgo:
Ah great. @ajor is starting a culture war...

I honestly don't feel strongly though BPFTRACE_LAZY_SYMBOLICATE is more letters AND the actual bcc field is "lazy_symbolize". That being said, in casual conversation I do tend to say "symbolication" so I yield to the wise recommendations of the bpftrace maintainers. But "analyse" is wrong.

viktormalik:
> > Should we also add this as a config option?
> 
> I started down that path but because this is more specific to the environment or binary target (so yeah maybe a case for specific uprobes), I thought it better as just an env var but I don't feel strongly.

The scripts usually contain the binary target (as a part of uprobes), so IMHO it makes sense to be able to set this in the script via a config option.

jordalgo:
Ok, changed it to 'BPFTRACE_LAZY_SYMBOLICATION' and added it to the config.

docs/reference_guide.md
ajor, line 1: There are a couple of hyperlink references to `#912-bpftrace_max_strlen` that'll need to be updated to match the new number - I only remembered because we fixed this same problem earlier today :P  #2957

ajor (approved):
Otherwise looks good!

docs/reference_guide.md
jordalgo: Good catch. I was about to put up a PR to remove these sub-section numbers so we don't have to keep doing this. But I'll fix this first and submit a second PR.

PR #2960, opened 2024-01-26
Fix build failure on 32-bit ARM

### LINKED ISSUE #2953: Build failing on 32-bit ARM
Alpine's armv7 and armhf builds are failing on 0.20.0:

```
.../bpftrace/src/bpftrace-0.20.0/src/bpftrace.cpp: At global scope:
.../bpftrace/src/bpftrace-0.20.0/src/bpftrace.cpp:2034:13: error: no declaration matches 'std::string bpftrace::BPFtrace::resolve_uid(uint64_t) const'
 2034 | std::string BPFtrace::resolve_uid(uint64_t addr) const
```

and

```
.../bpftrace/src/bpftrace-0.20.0/src/attached_probe.cpp: In member function 'void bpftrace::AttachedProbe::attach_multi_uprobe(int)':
.../bpftrace/src/bpftrace-0.20.0/src/attached_probe.cpp:1113:43: error: cannot convert 'long long unsigned int*' to 'const long unsigned int*' in assignment
 1113 |   opts.uprobe_multi.offsets = offsets.data();
      |                               ~~~~~~~~~~~~^~
      |                                           |
      |                                           long long unsigned int*
```

[armv7 build output](https://gitlab.alpinelinux.org/acj/aports/-/jobs/1253172)
[armhf build output](https://gitlab.alpinelinux.org/acj/aports/-/jobs/1253173)

I think the second error may be related to #2810.

acj:
Fixes #2953. I'm not certain that the uprobe_multi offset cast is correct/safe, but the tests seem okay with it.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -1034,7 +1034,7 @@ static void resolve_offset_uprobe_multi(const std::string &path,
                                         const std::string &probe_name,
                                         const std::vector<std::string> &funcs,
                                         std::vector<std::string> &syms,
-                                        std::vector<uint64_t> &offsets)
+                                        std::vector<unsigned long> &offsets)
 {
   struct bcc_symbol_option option = {};
   int err;
@@ -1099,7 +1099,7 @@ static void resolve_offset_uprobe_multi(const std::string &path,
 void AttachedProbe::attach_multi_uprobe(int pid)
 {
   std::vector<std::string> syms;
-  std::vector<uint64_t> offsets;
+  std::vector<unsigned long> offsets;
   unsigned int i;
 
   // Resolve probe_.funcs into offsets and syms vector
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -128,14 +128,14 @@ public:
                         StackType stack_type,
                         int indent = 0);
   std::string resolve_buf(char *buf, size_t size);
-  std::string resolve_ksym(uintptr_t addr, bool show_offset=false);
-  std::string resolve_usym(uintptr_t addr,
+  std::string resolve_ksym(uint64_t addr, bool show_offset = false);
+  std::string resolve_usym(uint64_t addr,
                            int pid,
                            int probe_id,
                            bool show_offset = false,
                            bool show_module = false);
-  std::string resolve_inet(int af, const uint8_t* inet) const;
-  std::string resolve_uid(uintptr_t addr) const;
+  std::string resolve_inet(int af, const uint8_t *inet) const;
+  std::string resolve_uid(uint64_t addr) const;
   std::string resolve_timestamp(uint32_t mode,
                                 uint32_t strftime_id,
                                 uint64_t nsecs);


ajor:
The function signature changes look good, but I don't understand what the `uprobe_multi.offsets` cast is for. Is it a necessary part of the fix or does it just resolve some compiler warnings?

Isn't casting a `uint64_t*` to a `const long unsigned int *` a no-op (even on systems where `uint64 != unsigned long`, pointers to them will be identical)?

If `sizeof(unsigned long) != sizeof(uint64)`, then the 64-bit offsets array will not have the correct format for libbpf and the cast is just tricking the compiler.

Could this be how it's working?
`uint64_t*`: `0x00000000deadbeef, 0x00000000cafef00d`
becomes
`uint32_t*`: `0xdeadbeef, 0x00000000, 0xcafef00d, 0x00000000`

So it tries to attach to the correct addresses as well as `0x00000000`.

acj:
> Is it a necessary part of the fix or does it just resolve some compiler warnings?

It fixes this compiler error on arm32:

```
.../bpftrace/src/bpftrace-0.20.0/src/attached_probe.cpp: In member function 'void bpftrace::AttachedProbe::attach_multi_uprobe(int)':
.../bpftrace/src/bpftrace-0.20.0/src/attached_probe.cpp:1113:43: error: cannot convert 'long long unsigned int*' to 'const long unsigned int*' in assignment
 1113 |   opts.uprobe_multi.offsets = offsets.data();
      |                               ~~~~~~~~~~~~^~
      |                                           |
      |                                           long long unsigned int*
```

I think your analysis is right -- the possible mismatch is what I was concerned about. Is there a better way to handle this?

viktormalik:
> I think your analysis is right -- the possible mismatch is what I was concerned about. Is there a better way to handle this?

`opts.uprobe_link.offsets` is `const unsigned long *` (see [bpf/bpf.h:399](https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L399)) so our `offsets` vector should probably respect that (so it should be `std::vector<const unsigned long>`).

@ajor @olsajiri wdyt? Does that make sense?

ajor:
> `opts.uprobe_link.offsets` is `const unsigned long *` (see [bpf/bpf.h:399](https://github.com/libbpf/libbpf/blob/master/src/bpf.h#L399)) so our `offsets` vector should probably respect that (so it should be `std::vector<const unsigned long>`).

That makes sense to me, but I'd go for `std::vector<uintptr_t>` for clarity about what we're doing - it should be the right size on all platforms.

From what I understand, we always use `uint64_t` for addresses in BPF (whether on 64 or 32-bit platforms) and therefore also for any addresses returned from BPF to userland, but as the addresses here are just used for attachment (not in BPF at all) they should be platform-specific and not follow this rule.

acj:
Using `uintptr_t` as the offset vector type:

```
.../src/attached_probe.cpp:1113:43: error: invalid conversion from 'unsigned int*' to 'const long unsigned int*' [-fpermissive]
 1113 |   opts.uprobe_multi.offsets = offsets.data();
      |                               ~~~~~~~~~~~~^~
      |                                           |
      |                                           unsigned int*
```

Using `const unsigned long`:

```
In file included from /usr/include/c++/13.2.1/regex:57,
                 from /bpftrace/src/attached_probe.cpp:14:
/usr/include/c++/13.2.1/bits/stl_vector.h: In instantiation of 'class std::vector<const long unsigned int>':
/bpftrace/src/attached_probe.cpp:1095:12:   required from here
/usr/include/c++/13.2.1/bits/stl_vector.h:437:66: error: static assertion failed: std::vector must have a non-const, non-volatile value_type
  437 |       static_assert(is_same<typename remove_cv<_Tp>::type, _Tp>::value,
      |                                                                  ^~~~~
/usr/include/c++/13.2.1/bits/stl_vector.h:437:66: note: 'std::integral_constant<bool, false>::value' evaluates to false
/usr/include/c++/13.2.1/bits/stl_vector.h:440:64: error: static assertion failed: std::vector must have the same value_type as its allocator
  440 |       static_assert(is_same<typename _Alloc::value_type, _Tp>::value,
      |                                                                ^~~~~
```

`unsigned long` builds and passes the tests, so I've updated the branch to use that.

ajor (approved):
Looks good to me!

danobi:
Should we cut a point release for this?

danobi:
Also thinking ahead: should be possible to add a CI job for `-m32` right?

acj:
Thanks, all!

> Should we cut a point release for this?

@danobi Yes, would be appreciated so that downstream packagers are unblocked and don't need to cook up temporary patches.

danobi:
@acj v0.20.1 cut: https://github.com/iovisor/bpftrace/releases/tag/v0.20.1

PR #2968, opened 2024-01-27
docs: Add coding_guidelines.md

danobi:
Initial document. If anything in the document warrants debate, let's
do it here now.

##### Checklist

### DIFF
--- a/docs/coding_guidelines.md
+++ b/docs/coding_guidelines.md
@@ -1,0 +1,51 @@
+# Coding guidelines
+
+The goal of this document is to establish some common understanding of what
+language features should and should not be used. This helps reduce mental
+overhead while reading, developing, and reviewing code. In some cases, it
+also helps present a more consistent user experience.
+
+Discussions about this document should occur in a pull request making changes
+to the text of this document. This helps keep track of why things are the way
+they are. As well as providing a structured environment to participate.
+
+If something is not called out in this document, defer to [the C++ core
+guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+In other words, this document supersedes the core guidelines.
+
+Note that the checked in code may not adhere to these guidelines yet.
+This document is intended to be referenced for new code. Ideally we also
+modify checked-in code to follow these guidelines (if/when there is time).
+
+## Error handling and exceptions
+
+If an error is recoverable (meaning downstream callers of the code should
+be able to selectively handle, ignore, or further propagate), pass the
+error through the return value of the function.
+
+Appropriate language features for this include:
+
+* `std:optional`
+* `int`
+* `bool`
+
+If an error is **not** recoverable, prefer throwing an exception. Broadly
+speaking, if you want to immediately terminate and show a message to the
+user, use an exception. Exceptions **should not** be used for recoverable
+errors.
+
+### Examples
+
+An example of a recoverable error would be failure to write an entry into a
+map. There are a number of reasons why writing to a map could fail (map is
+full, insufficient privileges, map does not exist yet, etc.) and we can't know
+for sure at the point of failure what the reason is. Additionally, this type of
+error might not be a reason fail the whole program. So we need to propagate the
+error.
+
+An example of an unrecoverable error would be if `debugfs` is not mounted.
+If debugfs isn't mounted, all bets are off b/c we cannot interact with
+the kernel debug facilities. And we probably should not be mounting
+filesystems on behalf of the user. So at the point of failure, we should
+just throw an exception and let the user know their system is not ready
+for bpftrace.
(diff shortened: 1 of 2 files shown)


docs/coding_guidelines.md
```diff
@@ -0,0 +1,35 @@
...
+
+If something is not called out in this document, defer to [the C++ core
+guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+In other words, this document supersedes the core guidelines.
+
+Note that the checked in code may not adhere to these guidelines yet.
+This document is intended to be referenced for new code. Ideally we also
+modify checked-in code to follow these guidelines (if/when there is time).
+
+## Error handling and exceptions
```
jordalgo, line 20: Nit: would be good to include a small example.

docs/developers.md
```diff
@@ -109,6 +109,14 @@ list of known such tests:
 What usually helps, is restarting the CI. This is simple on your own fork but
 requires one of the maintainers for pull requests.
 
+## Coding guidelines
+
+This is not about what the source code looks like. Rather, it's about the
```
jordalgo, line 114:
```suggestion
This is not about the format of the source code (we have clang-format). Rather, it's about the
```

docs/coding_guidelines.md
danobi: Yeah, good idea

docs/coding_guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+If an error is **not** recoverable, prefer throwing an exception. Broadly
+speaking, if you want to immediately terminate and show a message to the
+user, use an exception. Exceptions **should not** be used for recoverable
+errors.
+
+### Examples
+
+An example of a recoverable error would be failure to write an entry into
+a map. There are a number of reasons why writing to a map could fail
+(map is full, insufficient privileges, map does not exist yet, etc.) and
```
jordalgo, line 41:
```suggestion
(map is full, insufficient privileges, map does not exist yet, etc...) and
```

docs/coding_guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+we can't know for sure at the point of failure what the reason is. So
+we need to propagate the error.
+
+An example of an unrecoverable error would be if `debugfs` not mounted.
```
jordalgo, line 45: nit

```suggestion
An example of an unrecoverable error would be if `debugfs` failed to mount.
```

docs/coding_guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+speaking, if you want to immediately terminate and show a message to the
+user, use an exception. Exceptions **should not** be used for recoverable
+errors.
+
+### Examples
+
+An example of a recoverable error would be failure to write an entry into
+a map. There are a number of reasons why writing to a map could fail
+(map is full, insufficient privileges, map does not exist yet, etc.) and
+we can't know for sure at the point of failure what the reason is. So
```
jordalgo, line 42:
```suggestion
we can't know for sure at the point of failure what the reason is. Additionally, this type of error might not be a reason fail the whole program. So
```

jordalgo (approved):
Thanks for getting this started! Two more minor nits but looks good otherwise.

ajor (approved):
Looks good, thanks for starting this!

docs/coding_guidelines.md
danobi: Should probably be "is not mounted". I don't believe we try to mount debugfs in our code (nor should we)

danobi:
@viktormalik would like to get a look from you

docs/coding_guidelines.md
```diff
@@ -0,0 +1,51 @@
...
+
+If something is not called out in this document, defer to [the C++ core
+guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+In other words, this document supersedes the core guidelines.
+
+Note that the checked in code may not adhere to these guidelines yet.
+This document is intended to be referenced for new code. Ideally we also
+modify checked-in code to follow these guidelines (if/when there is time).
+
+## Error handling and exceptions
```
viktormalik, line 20: Should we also mention how and when to use the `LOG(ERROR)` facility?

viktormalik (approved):
LGTM, just one remark but I'm ok adding that in another PR.

docs/coding_guidelines.md
danobi: Do you mean in context of AST passes?

viktormalik: Well, we're using it in other places, too. But yes, in context of AST passes it's especially useful as using it makes the pass fail.

PR #2969, opened 2024-01-30
List program probes when -l is also provided

jordalgo:
Extend listing functionality to also list all the probes a program would attach to when '-l' arg is provided.

e.g.
```
$ sudo bpftrace -l -e 'kfunc:vmlinux:vfs_* { exit(); }'
kfunc:vmlinux:vfs_cancel_lock
kfunc:vmlinux:vfs_clean_context
kfunc:vmlinux:vfs_cleanup_quota_inode
kfunc:vmlinux:vfs_clone_file_range
kfunc:vmlinux:vfs_cmd_create
kfunc:vmlinux:vfs_copy_file_range
kfunc:vmlinux:vfs_create
kfunc:vmlinux:vfs_create_mount
kfunc:vmlinux:vfs_dedupe_file_range
kfunc:vmlinux:vfs_dedupe_file_range_one
kfunc:vmlinux:vfs_dentry_acceptable
kfunc:vmlinux:vfs_dup_fs_context
kfunc:vmlinux:vfs_fadvise
...
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to
 #### Added
 - Add 'lazy_symbolication' config
   - [#2958](https://github.com/bpftrace/bpftrace/pull/2958)
+- Add ability to list all probes in a program
+  - [#2969](https://github.com/bpftrace/bpftrace/pull/2969)
 #### Changed
 #### Deprecated
 #### Removed
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -134,6 +134,22 @@ RUN {{BPFTRACE}} -l "rawtracepoint:*"
 EXPECT rawtracepoint:
 TIMEOUT 1
 
+NAME it only lists probes in the program
+RUN {{BPFTRACE}} -l -e 'kfunc:vmlinux:vfs_read { exit(); }' | grep kfunc | wc -l
+EXPECT 1
+TIMEOUT 5
+
+NAME it lists uprobes in the program
+RUN {{BPFTRACE}} -l -e 'uretprobe:*:uprobeFunction* { exit(); }' -p {{BEFORE_PID}}
+EXPECT uretprobe:[\S]+uprobe_test:uprobeFunction1
+TIMEOUT 5
+BEFORE ./testprogs/uprobe_test
+
+NAME it lists multiple probes in the program
+RUN {{BPFTRACE}} -l -e 'hardware:cache-misses:10 { exit(); } tracepoint:xdp:mem_connect { exit(); }'
+EXPECT hardware:cache-misses:\ntracepoint:xdp:mem_connect
+TIMEOUT 5
+
 NAME errors on invalid character in search expression
 RUN {{BPFTRACE}} -l '\n'
 EXPECT ERROR: invalid character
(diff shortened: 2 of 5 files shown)


jordalgo:
(still in a draft because I need to do some more testing - and add more tests)

viktormalik (approved):
A very useful feature!

tests/runtime/basic
```diff
@@ -134,6 +134,22 @@ RUN {{BPFTRACE}} -l "rawtracepoint:*"
...
+NAME it lists uprobes in the program
+RUN {{BPFTRACE}} -l -e 'uretprobe:*:uprobeFunction* { exit(); }' -p {{BEFORE_PID}}
+EXPECT uretprobe:[\S]+uprobe_test:uprobeFunction1
+TIMEOUT 5
+BEFORE ./testprogs/uprobe_test
+
+NAME it lists multiple probes in the program
+RUN {{BPFTRACE}} -l -e 'hardware:cache-misses:10 { exit(); } tracepoint:xdp:mem_connect { exit(); }'
+EXPECT hardware:cache-misses:\ntracepoint:xdp:mem_connect
+TIMEOUT 5
```
danobi, line 151: Given this PR reduces the scope of the output compared to previous behavior, I don't think these tests catch any future regressions, right? We need some negative regex tests as well

danobi (changes requested):
Nice!

Although this looks more like a bug fix than a new feature, right?

```
$ sudo bpftrace -l -e 'kfunc:vmlinux:vfs_* { exit(); }' | wc -l
111518
$ sudo ./build/src/bpftrace -l -e 'kfunc:vmlinux:vfs_* { exit(); }' | wc -l
74
```

Maybe the changelog wording should be updated to reflect that. Cuz it wasn't clear to me

jordalgo:
> Although this looks more like a bug fix than a new feature, right?

I think that can probably be debated since there is nowhere in the docs saying they can be used together. It's probably a bug that passing both didn't produce an error. That being said, I don't feel strongly either way.

tests/runtime/basic
jordalgo: The only one that is superfluous is the first one as `-l` by itself doesn't show uprobes and the third test shows these two probes directly following each other in the list, which doesn't happen with just `-l`. But let me see if I can't improve these a bit more.

danobi: Ah got it. Maybe some kind of EXPECT_NOT directive?

danobi:
> > Although this looks more like a bug fix than a new feature, right?
> 
> I think that can probably be debated since there is nowhere in the docs saying they can be used together. It's probably a bug that passing both didn't produce an error. That being said, I don't feel strongly either way.

Fair enough :) As-is is fine then

CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ and this project adheres to
 
 #### Added
 - Add 'lazy_symbolication' config
-  - [#2958](https://github.com/bpftrace/bpftrace/pull/2958)
+  - [#2958](https://github.com/iovisor/bpftrace/pull/2958)
+- Add ability to list all probes in a program
+  - [#2969](https://github.com/iovisor/bpftrace/pull/2969)
```
danobi, line 15: Old URL?

danobi (approved):
ah grep | wc -l is good too

CHANGELOG.md
jordalgo: ah good catch. Old habits die hard.

tests/runtime/basic
jordalgo: Yeah I think EXPECT_NOT was mentioned on another PR. I'll add it soon if someone doesn't beat me to it.

PR #2973, opened 2024-01-30
ci: Add clang-tidy job

### LINKED ISSUE #2971: Set up clang-tidy in CI
`clang-tidy` seems like a useful linter. C++ could really use all the linters it can get.

https://github.com/marketplace/actions/clang-tidy-review looks useful but it might
be nice to be able to run clang-tidy locally as well thru nix. Perhaps this action
is better for that: https://github.com/marketplace/actions/clang-tidy-action . ie.
run clang-tidy thru nix, get the output file, then use the action to post inline comments
on the PR.

Would also be nice to only clang-tidy on changed lines, much like git-clang-format.

danobi:
This PR makes the CI start running clang-tidy. We're starting small
and over time hopefully we enable more checks. Right now this job
is structured to run over the entire codebase. So try to pair new lints
with the fixes in the same PR in the future.

This closes #2971.

##### Checklist


danobi:
I'm thinking that rather than get fancy with the GH action, we should slowly allow-list lints into `.clang-tidy.yml` paired with codemods. That way, codebase always stays clang-tidy clean and we can do a naive clang-tidy over all source files. And we can run that inside nix env so it's easy to repro locally.

A script we maintain that can both `-fix` and use in CI would be nice.

src/ast/irbuilderbpf.cpp
```diff
@@ -1605,7 +1605,8 @@ Value *IRBuilderBPF::CreateStrcontains(Value *val1,
 
 CallInst *IRBuilderBPF::CreateGetNs(TimestampMode ts, const location &loc)
 {
-  libbpf::bpf_func_id fn;
+  // Random default value to silence compiler warning
+  libbpf::bpf_func_id fn = libbpf::BPF_FUNC_ktime_get_ns;
```
jordalgo, line 1609: soo glad this is being fixed-ish

src/bpf_assembler.h
```diff
@@ -0,0 +1,168 @@
+// BPF macro assembler copied from bcc/libbpf.h
```
jordalgo, line 1: what is this file for?

jordalgo (approved):
🧹

src/bpf_assembler.h
danobi: For the test programs in bpffeature. We use assembler there

PR #2934, opened 2024-01-18
Fixup generated BTF for older kernels

viktormalik:
Older kernels missing https://github.com/torvalds/linux/commit/51c39bb1d5d1 (most importantly the stable 5.4) do not support BTF function entries with global linkage (`BTF_FUNC_GLOBAL`). This causes them to reject our generated BTF for probes and prevents from running any bpftrace program.

In order to fix this, the generated BTF must be patched by replacing the `BTF_FUNC_GLOBAL` flag by 0. This is normally done by libbpf (https://github.com/torvalds/linux/commit/2d3eb67f64ec) but only when program loading via bpf_object is used. Since we're still using direct loading using `bpf_btf_load` and `bpf_prog_load`, we need to mimic libbpf's behavior and do the patching manually. Note that we cannot fix this directly in codegen b/c LLVM would optimize all our probes away.

The BTF fixup itself is pretty simple but required to copy a portion of code from libbpf internals. The reason is that libbpf doesn't export functions for iterating BTF in a non-const way.

This also requires to feature-detect support for `BTF_FUNC_GLOBAL`, so we extend `BPFfeature` with BTF feature detection.

Fixes #2913.

##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #2977, opened 2024-02-01
Fix build issue with UAPI header of pre-v5.16 kernel

shunghsiyu:
322cdfa7 ("Fixup generated BTF for older kernels") added with PR #2934 allowed the generated BTF to work stable v5.4 kernel. However, the newly added btf_type_size() requires definition of struct btf_decl_tag (v5.16) and struct btf_enum64 (v6.0), which were not available in UAPI header of v5.4 kernel; which will cause build to fail.

Similar to libbpf/libbpf@85f8b7c, we can fix this by treating the structs as an array, and thus calculate their respective size without require their definition being present in the kernel UAPI header.

(CC @viktormalik)

##### Checklist

### DIFF
--- a/src/bpfbytecode.cpp
+++ b/src/bpfbytecode.cpp
@@ -49,7 +49,16 @@ static int btf_type_size(const struct btf_type *t)
     case BTF_KIND_ENUM:
       return base_size + vlen * sizeof(struct btf_enum);
     case BTF_KIND_ENUM64:
-      return base_size + vlen * sizeof(struct btf_enum64);
+      /* struct btf_enum64 is not available in UAPI header until v6.0,
+       * calculate its size with array instead. Its definition is:
+       *
+       * struct btf_enum64 {
+       *	__u32	name_off;
+       *	__u32	val_lo32;
+       *	__u32	val_hi32;
+       * };
+       */
+      return base_size + vlen * sizeof(__u32[3]);
     case BTF_KIND_ARRAY:
       return base_size + sizeof(struct btf_array);
     case BTF_KIND_STRUCT:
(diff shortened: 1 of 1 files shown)


shunghsiyu:
Technically the oldest supported kernel is v4.19, which would further require the definition of `struct btf_param`, `struct btf_var`, and `struct btf_var_secinfo`. But it seems better to align with the original 322cdfa and target v5.4 instead.

shunghsiyu:
Alternatively we can also add feature detection in `cmake/FindLibBcc.cmake` to check if those struct are defined. But since we're not accessing the member fields and they're part of UAPI, I opted for the simpler approach here.

src/bpfbytecode.cpp
```diff
@@ -49,7 +49,8 @@ static int btf_type_size(const struct btf_type *t)
     case BTF_KIND_ENUM:
       return base_size + vlen * sizeof(struct btf_enum);
     case BTF_KIND_ENUM64:
-      return base_size + vlen * sizeof(struct btf_enum64);
+      // btf_enum64 is not available until v6.0
```
viktormalik, line 52: Could we add the UAPI definition of `btf_enum64` to the comment? Just to make it clear why the below `sizeof` expression is used.

src/bpfbytecode.cpp
```diff
@@ -62,7 +63,8 @@ static int btf_type_size(const struct btf_type *t)
     case BTF_KIND_DATASEC:
       return base_size + vlen * sizeof(struct btf_var_secinfo);
     case BTF_KIND_DECL_TAG:
-      return base_size + sizeof(struct btf_decl_tag);
+      // btf_decl_tag is not available until v5.16
```
viktormalik, line 66: Same as above.

viktormalik (changes requested):
Looks like a reasonable approach to me. Just two remarks.

src/bpfbytecode.cpp
shunghsiyu: Sounds good, I will add their definitions.

shunghsiyu:
Added the UAPI definitions of `struct btf_decl_tag` and `struct btf_enum64` into the comment to make it clearer why the expression within sizeof() is chosen as suggested. Also drop the array syntax `[1]` that comes after `__s32` to make code a tiny bit cleaner.

danobi (approved):
thanks!

PR #2985, opened 2024-02-06
Use ELF map relocations instead of custom ones

viktormalik:
Currently, codegen encodes maps by their ids using `llvm.bpf.pseudo` instructions. LLVM transforms these into BPF instructions and we relocate them before loading the program by replacing map ids by fds.

Since we now emit global variables for maps, codegen can directly use the global variables to encode maps. LLVM will transform these accesses into BPF instructions and also will generate relocation entries to the ELF. The relocation entry will contain the symbol (map) so we can use its name to retrieve the map fd from MapManager and do the relocation.

There are several advantages to that:
- this is how libbpf does map relocations so once we switch to libbpf-based loading, we just drop the manual relocations,
- no need to keep map id->fd mapping which means a better decoupling of codegen and runtime.

Note: it'd probably be possible to clean up some MapManager code for working with map ids (b/c codegen now only works with map names) but since MapManager is going away soon, I decided not to do that now.

##### Checklist


danobi (approved):
nice! makes sense to me and looks like a good simplification.

ajor (approved):

PR #2987, opened 2024-02-08
Update tutorial_one_liners.md

next-ramses:
The example in Lesson 3 fails with ERROR: Can not access field 'filename' on type '(ctx) struct _tracepoint_syscalls_sys_enter_openat *'. Try dereferencing it first, or using '->'

Corrected to use ->

##### Checklist


viktormalik:
Since #2578, `args` is no longer a pointer and can be accessed via `.`. Please update your bpftrace.

Issue #2911, opened 2024-01-10
Improving map management


viktormalik:
I have prepared a large series with major refactoring of map management in bpftrace. To simplify reviewing, I'm going to post it in as small parts as possible but I realized that it would be good to provide an overview of all changes in advance.

The current plan is as follows:
- [x] #2910
- [x] #2912
- [x] #2941. Thanks to this, part of the map information (type, key/value size, max entries) will be stored inside the generated ELF.
- [x] #2988
- [x] Delete `MapManager`, `IMap`, `Map`, and `FakeMap` classes.

Overall, this simplifies the code (we especially get rid of the ugly `IMap`/`Map` constructors) and also is the next step towards libbpf-based loading and attachment of programs. We should be eventually able to let libbpf auto-generate the maps for us.

viktormalik:
All items are done, closing.

PR #2988, opened 2024-02-09
Move map management to `BpfBytecode`

viktormalik:
The current map management using `MapManager`/`IMap`/`Map`/`FakeMap` is quite difficult to read in the code as map settings are scattered across multiple files. This PR attempts to unify everything in `BpfBytecode`, where it logically belongs.

Maps are stored by their name (map ids are not used anymore) and represented using a new `BpfMap` class. The class contains `struct bpf_map` which is libbpf map representation containing basic map information (map type, max entries, key and value size) which libbpf is able to parse from BTF stored in the generated ELF. In addition, `BpfMap` stores map fd. Once we let libbpf auto-create maps for us in future, we will be able to drop manual map creation and fd will also be stored in `struct bpf_map`. Logically, the new `BpfMap` class replaces the previous `Imap`/`Map` classes.

The PR migrates all map usages to the new map management. The migration had to deal with two kinds of maps:
- Internal maps which usually only use fd and therefore the migration was mostly straightforward.
- Async map actions (`print`, `clear`, and `zero`) which came with two problems: (1) async actions need map ids which were dropped and (2) some information needed (e.g. key/value type) is not stored in `BpfMap`.  Since I wanted to keep the generated BPF programs decoupled from runtime as much as possible, I decided to store the above information in `RequiredResources` which are accessible by async actions. Part of the information was already there and the PR adds map name<->id mappings (for maps used inside async actions).

Finally, `MapManager`, `IMap`, `Map`, and `FakeMap` are dropped.

See individual commit messages for more details.

Note: this is the last part of #2911. It's quite big but unfortunately, I wasn't able to split it into smaller parts as it wouldn't make much sense to me. If you still prefer to post it per-commit, let me know.

##### Checklist


viktormalik:
For some reason, there are random failures of the `call.ustack_stack_mode_env*` runtime tests. Restarting the job usually helps but I don't remember seeing such failures before so they may be caused by something in this PR. I'll try to have a look.

viktormalik:
> For some reason, there are random failures of the `call.ustack_stack_mode_env*` runtime tests. Restarting the job usually helps but I don't remember seeing such failures before so they may be caused by something in this PR. I'll try to have a look.

Unfortunately, I'm not able to reproduce the issue locally (even when using Nix).

ajor:
What kinds of failures were they?

ajor (commented):
Overall looks good to me.

There do seem to be a lot of map lookups going on throughout the whole process, which might be slowing us down more than necessary.

e.g. in the output path (where performance counts), we have a multi-step process to obtain a BpfMap object:
1. Ringbuffer -> `print->mapid`
2. `resources.maps_by_id[mapid]` -> `map_name`
3. `bytecode.getMap(map_name)` -> `BpfMap`

Also `resources.map_key` and `resources.map_type` being two separate map lookups.

viktormalik:
> What kinds of failures were they?

IIRC there was no stack printed (just empty output).

src/bpfmap.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+  return bpf_map__name(bpf_map_);
+}
+
+std::string BpfMap::name() const
+{
+  return bpftrace_map_name(bpf_name());
+}
+
+__u32 BpfMap::key_size() const
```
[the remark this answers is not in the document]
viktormalik: Agreed, let's use standard `uint32_t`.

src/bpftrace.cpp
```diff
@@ -410,13 +411,18 @@ void perf_event_printer(void *cb_cookie, void *data, int size)
     return;
   } else if (printf_id == asyncactionint(AsyncAction::print)) {
     auto print = static_cast<AsyncEvent::Print *>(data);
-    IMap *map = *bpftrace->maps[print->mapid];
+    auto map_name = bpftrace->resources.maps_by_id.find(print->mapid);
+    if (map_name == bpftrace->resources.maps_by_id.end()) {
+      throw std::runtime_error("Could not find map with id " +
+                               std::to_string(print->mapid));
+    }
```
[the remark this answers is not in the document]
viktormalik: Yes, it's an internal error which should never happen.

src/required_resources.cpp
```diff
@@ -10,174 +10,16 @@
...
-#include "fake_map.h"
 #include "log.h"
 #include "utils.h"
 
 namespace bpftrace {
 
-int RequiredResources::create_maps(BPFtrace &bpftrace, bool fake)
+// All methods in this file are called by AOT only so we must explicitly define
+// at least one method called from non-AOT code to avoid linker skipping this
+// file while linking libruntime.a, which would lead to linker errors.
```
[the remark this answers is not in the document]
viktormalik: Good point. I guess that we could remove it from `libruntime` and just move it to AOT. I didn't do it b/c `required_resources.h` was included by many files in `libruntime` so I wanted to have them coupled together. But it probably doesn't make any sense.

viktormalik:
> Overall looks good to me.
> 
> There do seem to be a lot of map lookups going on throughout the whole process, which might be slowing us down more than necessary.

That's a good point, although I'm wondering if we can really avoid that.

> e.g. in the output path (where performance counts), we have a multi-step process to obtain a BpfMap object:
> 
> 1. Ringbuffer -> `print->mapid`

This is necessary b/c mapid is stored in codegen. I don't see a way to avoid this.

> 2. `resources.maps_by_id[mapid]` -> `map_name`
> 3. `bytecode.getMap(map_name)` -> `BpfMap`

Probably the most efficient would be to have mapid -> `BpfMap` mapping but the problem is that we need mapid *before* codegen but only create `BpfMap` *after* codegen. We could create an extra mapid -> `BpfMap` mapping (perhaps in `BpfBytecode`) after `BpfMap` objects are created. I can give it a try. If you have other idea how to resolve this, let me know.

> Also `resources.map_key` and `resources.map_type` being two separate map lookups.

That's a good point. We could probably aggregate them in some structure like `MapInfo` in `RequiredResources` (together with other info such as map ids, (l)hist args, etc.).

viktormalik:
Thanks for the review @ajor! I implemented all of the comments, noticeable changes in the new version are:
- moved linking of `required_resources.cpp` out of `libruntime` to `libaot` only,
- aggregated map information in `RequiredResources` to a single struct,
- created a new map `BpfBytecode::maps_by_id_` to allow efficient lookup of `BpfMap` objects by map ids.

src/bpfbytecode.cpp
```diff
@@ -26,6 +45,93 @@ const std::vector<uint8_t> &BpfBytecode::getSection(
...
+  }
+  return n;
+}
+
+void BpfBytecode::set_map_ids(RequiredResources &resources)
+{
+  for (auto &map : maps_) {
+    auto map_info = resources.maps_info.find(map.first);
+    if (map_info != resources.maps_info.end() && map_info->second.id != -1)
+      maps_by_id_.emplace(map_info->second.id, &map.second);
```
viktormalik, line 131: IIUC, this should be fine b/c references to `std::map` are not invalidated by inserting new elements. But @ajor is the C++ expert here so I'd like to highlight this so that you double check.

ajor: Yep, looks fine to me 👍

src/bpftrace.cpp
```diff
@@ -1559,18 +1561,19 @@ int BPFtrace::print_map(IMap &map, uint32_t top, uint32_t div)
...
         });
-  } else if (map.type_.IsMaxTy()) {
+  } else if (value_type.IsMaxTy()) {
     std::sort(
         values_by_key.begin(), values_by_key.end(), [&](auto &a, auto &b) {
           return max_value(a.second, nvalues) < max_value(b.second, nvalues);
         });
   } else {
-    sort_by_key(map.key_.args_, values_by_key);
+    auto map_key = map_info.key;
```
ajor, line 1575:
```suggestion
    const auto &map_key = map_info.key;
```

src/bpftrace.cpp
```diff
@@ -1650,34 +1653,34 @@ int BPFtrace::print_map_hist(IMap &map, uint32_t top, uint32_t div)
...
     return -2;
   }
   auto key(old_key);
 
   std::map<std::vector<uint8_t>, std::vector<int64_t>> values_by_key;
 
-  while (bpf_get_next_key(map.mapfd_, old_key.data(), key.data()) == 0) {
-    auto key_prefix = std::vector<uint8_t>(map.key_.size());
-    uint64_t bucket = read_data<uint64_t>(key.data() + map.key_.size());
+  auto map_key = resources.maps_info.at(map.name()).key;
```
ajor, line 1674:
```suggestion
  const auto &map_key = resources.maps_info.at(map.name()).key;
```

src/output.cpp
```diff
@@ -410,6 +411,8 @@ std::string Output::map_hist_to_str(
 {
   std::vector<std::string> elems;
   uint32_t i = 0;
+  const auto &map_info = bpftrace.resources.maps_info.at(map.name());
+  auto map_type = map_info.value_type;
```
ajor, line 415:
```suggestion
  const auto &map_type = map_info.value_type;
```

ajor (approved):
Looks great!

Just spotted few extra copies we can eliminate. I'm extra suspicious whenever I see `auto x = ` because it's easy to hide these expensive copies (not saying that I don't code like this myself!)

viktormalik:
> Just spotted few extra copies we can eliminate. I'm extra suspicious whenever I see `auto x = ` because it's easy to hide these expensive copies (not saying that I don't code like this myself!)

Thanks, those are great catches! I try to avoid `auto x = `, too, but sometimes my brain falls into a just-auto-everything mode and that's when these happen :laughing:

Anyways, I'd like to wait for @danobi's review before merging this.

danobi (commented):
only looked at first 3 commits. might get more time this weekend to finish

src/bpfbytecode.h
```diff
@@ -45,6 +53,8 @@ class BpfBytecode {
   };
   std::unique_ptr<struct bpf_object, bpf_object_deleter> bpf_object_;
 
+  std::map<std::string, BpfMap> maps_;
```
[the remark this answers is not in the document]
viktormalik: This is on purpose to get the maps in alphabetical order when printing them at the end.

src/bpfbytecode.cpp
```diff
@@ -12,6 +15,14 @@ BpfBytecode::BpfBytecode(const void *elf, size_t elf_size)
       bpf_object__open_mem(elf, elf_size, nullptr));
   if (!bpf_object_)
     throw std::runtime_error("The produced ELF is not a valid BPF object");
+
+  struct bpf_map *m;
+  bpf_map__for_each(m, bpf_object_.get())
```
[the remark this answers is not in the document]
viktormalik: Good idea, I didn't know about this feature. Done

src/bpfbytecode.h
```diff
@@ -33,9 +33,13 @@ class BpfBytecode {
   const std::vector<uint8_t> &getSection(const std::string &name) const;
 
   int create_maps();
-  bool hasMap(MapManager::Type internal_type) const;
+  bool hasMap(MapType internal_type) const;
+  bool hasMap(const StackType &stack_type) const;
   const BpfMap &getMap(const std::string &name) const;
-  const BpfMap &getMap(MapManager::Type internal_type) const;
+  const BpfMap &getMap(MapType interal_type) const;
```
danobi, line 40: `interal_type` -> `internal_type`

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3329,17 +3329,15 @@ void CodegenLLVM::createMapDefinition(const std::string &name,
 void CodegenLLVM::generate_maps(const RequiredResources &resources)
 {
   // User-defined maps
-  for (auto &map_val : resources.map_vals) {
-    const std::string &name = map_val.first;
+  for (auto &map_info : resources.maps_info) {
```
danobi, line 3338: nit: could destructure here. eg:

```
for (const auto &[name, info] : resources.maps_info) {
    ...
}
```

src/required_resources.cpp
```diff
@@ -31,40 +31,36 @@ int RequiredResources::create_maps_impl(BPFtrace &bpftrace, bool fake)
...
-    auto &key = search_args->second;
+    auto &key = map_info.second.key;
     auto max_map_keys = bpftrace.config_.get(ConfigKeyInt::max_map_keys);
 
     if (type.IsLhistTy()) {
-      auto args = lhist_args.find(map_name);
-      if (args == lhist_args.end())
+      const auto &args = map_info.second.lhist_args;
+      if (!args.has_value())
         LOG(FATAL) << "map arg \"" << map_name << "\" not found";
```
danobi: Unrelated, but wondering if we should always use `LOG(FATAL)` instead of exceptions, since they serve similar purposes as documented in our coding guidelines. Would be an easy codemod

src/required_resources.h
```diff
@@ -54,6 +54,7 @@ struct MapInfo {
   SizedType value_type;
   std::optional<LinearHistogramArgs> lhist_args;
   std::optional<int> hist_bits_arg;
+  int id = -1;
```
danobi, line 56: Need to add to `archive()` ?

src/bpfbytecode.h
```diff
@@ -58,6 +61,7 @@ class BpfBytecode {
   std::unique_ptr<struct bpf_object, bpf_object_deleter> bpf_object_;
 
   std::map<std::string, BpfMap> maps_;
+  std::map<int, BpfMap *> maps_by_id_;
```
danobi, line 64: If the map is non-nullable here, maybe better to store a reference?

src/CMakeLists.txt
```diff
@@ -25,18 +25,14 @@ add_library(runtime
...
-  imap.cpp
   log.cpp
-  map.cpp
   mapkey.cpp
   output.cpp
   probe_matcher.cpp
   procmon.cpp
   printf.cpp
   resolve_cgroupid.cpp
-  required_resources.cpp
```
danobi, line 38: This means from core bpftrace files we only use required_resources.h definitions. Meaning if future pubic methods are exposed and called from core bpftrace, then you'd get a linker error then.

What was the linker error that motivated this change? Curious cuz I feel like it should work.

src/aot/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_library(aot aot.cpp)
+add_library(aot aot.cpp ${CMAKE_SOURCE_DIR}/src/required_resources.cpp)
```
danobi, line 1: Yeah this is a bit unfortunate. I feel like cmakelists.txt should try not to reach up to parent  or sibling dirs

danobi (commented):
few comments, but great cleanup!

src/required_resources.h
viktormalik: Yes, good catch.

src/ast/passes/codegen_llvm.cpp
viktormalik: Love this syntax! Would be nice to start using it more across the codebase.

src/bpfbytecode.h
viktormalik: I don't think it's possible to store a reference in a `std::map`. We could use `std::reference_wrapper` but that feels like adding an unnecessary complexity to me.

src/CMakeLists.txt
viktormalik: Since no function from `required_resources.cpp` is used by non-aot code, the linker optimizes it away when linking `libruntime` and afterwards linking `libaot` fails with:
```
/usr/bin/ld: aot/libaot.a(aot.cpp.o): in function `bpftrace::aot::(anonymous namespace)::load_required_resources(bpftrace::BPFtrace&, unsigned char*, unsigned long)':
/home/vmalik/src/bpf/bpftrace/src/aot/aot.cpp:77: undefined reference to `bpftrace::RequiredResources::load_state(unsigned char const*, unsigned long)'
/usr/bin/ld: aot/libaot.a(aot.cpp.o): in function `bpftrace::aot::(anonymous namespace)::generate_btaot_section(bpftrace::RequiredResources const&, bpftrace::BpfBytecode const&)':
/home/vmalik/src/bpf/bpftrace/src/aot/aot.cpp:109: undefined reference to `bpftrace::RequiredResources::save_state(std::ostream&) const'
```
My original solution was to add a dummy constructor of `RequiredResrouces` to `required_resources.cpp` but @ajor had a good point that it doesn't make sense linking an object file from which no code is used. OTOH, if we ever added some code needed from `libruntime`, we'd have to revert all this.

src/aot/CMakeLists.txt
viktormalik: I don't like it either. But I couldn't come with any other solution, besides adding the dummy constructor (see the other thread).

src/required_resources.cpp
viktormalik: This is something I'm often confused about. In some places, we use `LOG(FATAL)`, in others we raise `std::runtime_exception`. It'd be nice to unify this, perhaps we can discuss on the next office hours.

src/aot/CMakeLists.txt
ajor: Should required_resources.cpp live in the aot directory, since that's where it's used?

viktormalik: Possibly but we wouldn't have `required_resources.cpp` next to `required_resources.h`.

ajor: I was thinking of moving the header too, but I suppose it's used in more than just AOT so that doesn't make sense.

How about creating a new CMake library target just for required_resources? Then both runtime and aot can link against that. Would it solve the unused symbol issues?

src/bpfbytecode.h
danobi: Ah true, yeah. Ptr is fine

src/required_resources.cpp
danobi: Sounds good!

src/aot/CMakeLists.txt
danobi: Agreed - separate target is probably cleanest solution

viktormalik: Good idea, done.

src/ast/passes/codegen_llvm.cpp
danobi: I suspect clang-tidy can probably catch this. I'll make a note in the tracking issue and take a look when I get a chance

danobi (approved):

viktormalik:
Rebased. Going to merge this to avoid more conflicts with open PRs.

PR #2989, opened 2024-02-09
Runtime tests: Don't depend on specific probe points if not required

ajor:
Use the BEGIN/END probes for testing basic functionality to insulate us from kernel changes to the probe points.

Also put in some real expected results instead of useless regexes.


tests/runtime/other
```diff
@@ -191,8 +191,8 @@ EXPECT ok
 TIMEOUT 1
 
 NAME positional lhist
-RUN {{BPFTRACE}} -e 'kretprobe:vfs_read { @bytes = lhist(retval, $1, $2, $3); exit()}' 0 10000 1000
-EXPECT @bytes: *\n[\[(].*
+RUN {{BPFTRACE}} -e 'BEGIN { @ = lhist(0, $1, $2, $3); exit()}' 0 10000 1000
+EXPECT @: *\n[\[(].*
 TIMEOUT 5
 AFTER ./testprogs/syscall read
```
danobi, line 197: delete the AFTER?

also maybe use EXPECT_FILE as well?

ajor: Removed the AFTER.

Left the expected results as a regex as it turns out that bpftrace is completely broken for this test case!

I think loads of these regex tests are either wrong or are hiding bugs like this. I want to go through every test that's using a regex and change them to properly check their results, but not right now.

jordalgo (approved):
LGTM. Nice change.

danobi:
Filed https://github.com/bpftrace/bpftrace/issues/2993

PR #2992, opened 2024-02-10
Config: update lexer to match string start

jordalgo:
To prevent this error only match against "config"
keyword if it's the start of the string:

```
bpftrace -e 'struct Foo {int config;} BEGIN { $x = (struct Foo*)0; $x->config; }' -d
stdin:1:56-66: ERROR: syntax error, unexpected config
struct Foo {int config;} BEGIN { $x = (struct Foo*)0; $x->config; }
```

This should be fine since the config block needs
to be at the top and start the line.

##### Checklist

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -73,6 +73,7 @@ oct_esc  [0-7]{1,3}
 }
 
 bpftrace|perf|raw       { return Parser::make_STACK_MODE(yytext, loc); }
+^{space}*config         { return Parser::make_CONFIG(loc); }
 {builtin}               { return Parser::make_BUILTIN(yytext, loc); }
 {call}                  { return Parser::make_CALL(yytext, loc); }
 {call_and_builtin}      { return Parser::make_CALL_BUILTIN(yytext, loc); }
@@ -143,7 +144,6 @@ bpftrace|perf|raw       { return Parser::make_STACK_MODE(yytext, loc); }
 "?"                     { return Parser::make_QUES(loc); }
 "unroll"                { return Parser::make_UNROLL(loc); }
 "while"                 { return Parser::make_WHILE(loc); }
-"config"                { return Parser::make_CONFIG(loc); }
 "for"                   { return Parser::make_FOR(loc); }
 "return"                { return Parser::make_RETURN(loc); }
 "continue"              { return Parser::make_CONTINUE(loc); }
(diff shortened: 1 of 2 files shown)


ajor (changes requested):
I probably wasn't clear enough in the original issue description, but it's actually a problem with all reserved keywords, so we'd ideally find a solution that makes them all work in user-defined types.

jordalgo:
> so we'd ideally find a solution that makes them all work in user-defined types

This change still may be a good temporary solution while we work on one that covers a wider range of keywords, especially if "config" is more common.

src/lexer.l
```diff
@@ -73,6 +73,7 @@ oct_esc  [0-7]{1,3}
 }
 
 bpftrace|perf|raw       { return Parser::make_STACK_MODE(yytext, loc); }
+^config                 { return Parser::make_CONFIG(loc); }
```
danobi, line 76: Maybe allow leading whitespace too? This would make config blocks whitespace sensitive which (IIUC) is a departure from all other syntax

ajor:
I don't think it'd be too hard to fix this properly for all keywords: we'd just need to update any rules which should accept user-defined type/field names (such as `type` and `postfix_expr`) to accept "keywords" as well as plain `ident`s, e.g.:

```
keyword:
                config     { $$ = $1; }
        |       if         { $$ = $1; }
        |       while      { $$ = $1; }
                [...]
                ;
postfix_expr:
                [...]
        |       postfix_expr DOT ident      { $$ = new ast::FieldAccess($1, $3, @2); }
        |       postfix_expr DOT keyword    { $$ = new ast::FieldAccess($1, $3, @2); }
```

ajor:
It'd probably be cleaner to create a new rule along the lines of:
```
external_name: ident | keyword;
```
And then use that in place of `ident` where necessary:
```
postfix_expr:
                [...]
        |       postfix_expr DOT external_name      { $$ = new ast::FieldAccess($1, $3, @2); }
```

jordalgo:
@ajor Great suggestion! Is something like this what you had in mind? https://github.com/bpftrace/bpftrace/pull/3004

jordalgo:
Actually, I'm realizing we might just be able to add all the other keywords to the ident definition. All the builtins are there in addition to things like STACK_MODE
```
ident:
                IDENT         { $$ = $1; }
        |       BUILTIN       { $$ = $1; }
        |       CALL          { $$ = $1; }
        |       CALL_BUILTIN  { $$ = $1; }
        |       STACK_MODE    { $$ = $1; }
                ;
 ````

PR #2997, opened 2024-02-12
Runtime tests: Make EXPECT only match exact lines of output

ajor:
The old behaviour is preserved with the new keyword: EXPECT_REGEX

Regexes make the tests harder to read and write, with special characters needing to be escaped, and are being misused in quite a number of places.

Switch the tests which use regex features over to EXPECT_REGEX. A lot of these will need to be looked at in the future as they're accepting basically any output, right or wrong.

### DIFF
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -1,16 +1,16 @@
 NAME it shows version
 RUN {{BPFTRACE}} --version
-EXPECT bpftrace v
+EXPECT_REGEX ^bpftrace v\d
 TIMEOUT 1
 
 NAME it shows usage with help flag
 RUN {{BPFTRACE}} -h
-EXPECT USAGE
+EXPECT USAGE:
 TIMEOUT 1
 
 NAME it shows usage with bad flag
 RUN {{BPFTRACE}} -idonotexist
-EXPECT USAGE
+EXPECT USAGE:
 TIMEOUT 1
 WILL_FAIL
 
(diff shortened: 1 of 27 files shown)


tests/runtime/usdt
```diff
@@ -57,7 +57,7 @@ REQUIRES ./testprogs/systemtap_sys_sdt_check
 
 NAME usdt probes - filter probes by pid and wildcard probe name
 RUN {{BPFTRACE}} -l 'usdt:*:tracetest:test*' -p {{BEFORE_PID}}
-EXPECT usdt_test:tracetest:testprobe2
+EXPECT_REGEX usdt_test:tracetest:testprobe2
```
viktormalik, line 60: Wouldn't `EXPECT` be sufficient here?

tests/runtime/usdt
```diff
@@ -192,33 +192,33 @@ REQUIRES bash -c "exit 1"
 
 NAME usdt probes - attach to probe with probe builtin and args by file
 PROG usdt:./testprogs/usdt_test:* { printf("%lld %s\n", arg0, probe ); exit(); }
-EXPECT usdt:./testprogs/usdt_test:tracetest:testprobe
+EXPECT_REGEX usdt:./testprogs/usdt_test:tracetest:testprobe
```
viktormalik, line 195: This could probably use a better regex containing the `arg0` part, too.

tests/runtime/usdt
```diff
@@ -192,33 +192,33 @@ REQUIRES bash -c "exit 1"
...
 TIMEOUT 5
 BEFORE ./testprogs/usdt_test
 REQUIRES ./testprogs/systemtap_sys_sdt_check
 
 NAME usdt probes - attach to probe with probe builtin and args by file with child
 RUN {{BPFTRACE}} -e 'usdt:./testprogs/usdt_test:* { printf("%lld %s\n", arg0, probe ); exit(); }' -c ./testprogs/usdt_test
-EXPECT usdt:./testprogs/usdt_test:tracetest:testprobe
+EXPECT_REGEX usdt:./testprogs/usdt_test:tracetest:testprobe
```
viktormalik, line 202: Same as above.

tests/runtime/usdt
```diff
@@ -192,33 +192,33 @@ REQUIRES bash -c "exit 1"
...
 TIMEOUT 5
 
 NAME usdt probes - attach to probe with probe builtin and args by pid
 RUN {{BPFTRACE}} -e 'usdt:./testprogs/usdt_test:* { printf("%lld %s\n", arg0, probe ); exit(); }' -p {{BEFORE_PID}}
-EXPECT usdt_test:tracetest:testprobe
+EXPECT_REGEX usdt_test:tracetest:testprobe
```
viktormalik, line 207: Same as above.

tests/runtime/usdt
```diff
@@ -192,33 +192,33 @@ REQUIRES bash -c "exit 1"
...
 TIMEOUT 5
 BEFORE ./testprogs/usdt_test
 REQUIRES ./testprogs/systemtap_sys_sdt_check
 
 NAME usdt probes - attach to probe with semaphore
 RUN {{BPFTRACE}} -e 'usdt::tracetest:testprobe { printf("%s\n", str(arg1) ); exit(); }' -p {{BEFORE_PID}}
-EXPECT tracetest_testprobe_semaphore: [1-9]
+EXPECT_REGEX tracetest_testprobe_semaphore: 1
```
viktormalik, line 214: Why this has `EXPECT_REGEX` but the similar tests below have `EXPECT`?
(I'm not very familiar with the USDT semaphores)

tests/runtime/engine/runner.py
```diff
@@ -201,10 +201,13 @@ def run_test(test):
...
         def get_pid_ns_cmd(cmd):
             return nsenter + [os.path.abspath(x) for x in cmd.split()]
-            
+
         def check_expect(expect, output):
             try:
-                if expect.mode == "regex":
+                if expect.mode == "text":
+                    # Raw text match on an entire line, ignoring leading/trailing whitespace
+                    return re.search(f"^\s*{re.escape(expect.expect)}\s*$", output, re.M)
```
jordalgo, line 209: Do we need a regex here at all if we're matching on an exact substring? Would a simple `return expect.expect in output` work?

ajor: I did start with something a lot simpler before I ended up with this nasty regex :P

The reason for the regex is that it lets us perform a substring match but still ensure that it's anchored to the start and end of a line, to avoid a situation like
`EXPECT val = 1` matching the string `val = 10`

tests/runtime/usdt
ajor: The expected results were missing the beginning of the line here. I've updated them to be full line matching

tests/runtime/basic
```diff
@@ -91,47 +91,53 @@ TIMEOUT 1
...
-EXPECT kretfunc
+EXPECT_REGEX kretfunc:.*
 REQUIRES_FEATURE btf
 REQUIRES_FEATURE kfunc
 TIMEOUT 1
 
 NAME it lists interval probes with regex filter
 RUN {{BPFTRACE}} -l "interval:*"
-EXPECT interval:s
+EXPECT interval:hz:
```
jordalgo, line 118: nice

tests/runtime/json-output
```diff
@@ -72,102 +72,102 @@ TIMEOUT 5
 
 NAME printf
 RUN {{BPFTRACE}} -q -f json -e 'BEGIN { printf("test %d", 5); exit(); }'
-EXPECT ^{"type": "printf", "data": "test 5"}$
+EXPECT {"type": "printf", "data": "test 5"}
```
jordalgo, line 75: These should all be EXPECT_JSON (as that tolerates different json field ordering) and we should either update EXPECT_JSON to support raw strings instead of file names or just maybe add EXPECT_JSON_FILE - I can do that in a follow-up PR if you don't want to do it here.

jordalgo (approved):
I did a quick check of all the updated runtime tests though I was starting to have EXPECT fatigue by the end...

viktormalik (approved):

tests/runtime/json-output
ajor: Good idea, but I'm afraid the EXPECT fatigue has got me too so I'll leave it out of this PR 😅

Issue #2955, opened 2024-01-24
Provide method for looping over map elements


ajor:
A looping construct to iterate over map elements would be useful:
```
BEGIN
{
  @x[1] = 100;
  @x[2] = 200;
  for ($kv : @x)
  {
    printf("key: %d\n", $kv.0);
    printf("val: %d\n", $kv.1);
  }
}
```

It would help with custom formatting when printing maps and I'm sure others will come up with more use-cases.

Related to #2807

danobi:
Also similar to https://github.com/iovisor/bpftrace/issues/2384

danobi:
Yes, this feature is often requested and would be nice. I recall discussing this at one point (maybe office hours?) - I think we mostly agreed that it would be good to transform the loop into a callback on behalf of the user. Introducing callback semantics to the language seems complex and somewhat unnecessary.

viktormalik:
Yeah, just to summarize technical details - this should use the `bpf_for_each_map_elem` helper with a custom callback (created from the `for` loop body on behalf of the user), similarly to #2922.

I'll be happy to work on this but have other things in my queue atm so if someone wants to pick this up in the meantime, feel free to.

jordalgo:
Sounds useful. The syntax also looks good. @viktormalik I'm also happy to pick it up if no one gets to it first (though similarly I have other things I need to get to before this).

viktormalik:
Thanks @jordalgo. Just assign the issue to yourself once you start working on it (I'll do the same).

danobi:
Whoever implements it ought to do a blog post. I think it's a pretty interesting topic (transforming a loop into a callback). Could help attract compiler enthusiasts when they see there's a bunch of low hanging stuff around :)

jordalgo:
Just to note, we should probably test deletion/insertion into the same map during iteration to make sure there aren't weird side effects.

thediveo:
is this released yet?

jordalgo:
@thediveo Yes, it's in 0.21.0: https://github.com/bpftrace/bpftrace/releases/tag/v0.21.0

ajor:
Closed by #3003

PR #2999, opened 2024-02-12
Loop map prep work

ajor:
This PR just contains some small bits and pieces that will be useful for implementing #2955

### DIFF
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -12,12 +12,12 @@ namespace parser {
 using Printer = ast::Printer;
 
 void test_parse_failure(BPFtrace &bpftrace,
-                        const std::string &input,
+                        std::string_view input,
                         std::string_view expected_error = {})
 {
   std::stringstream out;
   Driver driver(bpftrace, out);
-  ASSERT_EQ(driver.parse_str(input), 1);
+  EXPECT_EQ(driver.parse_str(input), 1);
 
   if (expected_error.data()) {
     if (!expected_error.empty() && expected_error[0] == '\n')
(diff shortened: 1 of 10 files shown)


tests/parser.cpp
```diff
@@ -26,35 +26,40 @@ void test_parse_failure(BPFtrace &bpftrace,
...
 }
 
 TEST(Parser, builtin_variables)
 {
-  test("kprobe:f { pid }", "Program\n kprobe:f\n  builtin: pid\n");
+  test("kprobe:f { pid }", R"(
+Program
+ kprobe:f
+  builtin: pid
+)");
```
viktormalik, line 62: Any reason not to convert the below test cases, too?

ajor: Laziness! I just converted the first one to give an example of the new functionality. What I really wanted this feature for is the new tests I'm writing for #2955

tests/semantic_analyser.cpp
```diff
@@ -54,21 +55,24 @@ void test_for_warning(const std::string &input,
...
+  if (!input.empty() && input[0] == '\n')
+    input.remove_prefix(1); // Remove initial '\n'
+
   std::stringstream out;
   std::stringstream msg;
   msg << "\nInput:\n" << input << "\n\nOutput:\n";
 
   bpftrace.safe_mode_ = safe_mode;
-  EXPECT_EQ(driver.parse_str(input), 0);
+  ASSERT_EQ(driver.parse_str(input), 0);
```
jordalgo, line 72: Why do we want `ASSERT_EQ` here rather than EXPECT_EQ?

jordalgo:
nit on commit message: "Printer: Only print types if they are set" - Maybe I'm confused but you're removing the print_types conditional right?

tests/semantic_analyser.cpp
```diff
@@ -78,31 +82,36 @@ void test(BPFtrace &bpftrace,
   // Override to mockbpffeature.
   bpftrace.feature_ = std::make_unique<MockBPFfeature>(mock_has_features);
   ast::SemanticAnalyser semantics(driver.root.get(), bpftrace, out, has_child);
-  ASSERT_EQ(expected_result, semantics.analyse()) << msg.str() + out.str();
+  if (expected_result == -1) {
+    // Accept any failure result
+    EXPECT_NE(0, semantics.analyse()) << msg.str() + out.str();
+  } else {
+    EXPECT_EQ(expected_result, semantics.analyse()) << msg.str() + out.str();
```
viktormalik, line 89: Similarly to @jordalgo's comment above, why use `EXPECT_EQ` here instead of `ASSERT_EQ`? Same below for testing the error message.

tests/semantic_analyser.cpp
ajor: In the semantic analyser tests, we use ASSERT for the earlier, non-semantic analysis, passes because we can't continue if they fail. We don't have any other error checking on them, so if we allow the test to continue after failing here then it will get a segmentation fault. (Try swapping back to EXPECT and then putting a syntax error in a semantic analysis test to confirm).

ASSERTs for early passes followed by EXPECTs for the pass we care about are the pattern that's used throughout the unit tests (see `test_for_warning()` just above this function, or `codegen/common.h`).

tests/semantic_analyser.cpp
ajor: I prefer to use EXPECT unless ASSERT is required due to later code assuming success.

In this case, it's safe to check the error message no matter what the return code is, so EXPECT is ok, and it can provide the test developer with more information about why their test is failing.

e.g. `expected_error_code = 10`, `expected_error_message = "undefined map"`
If the actual error code ends up being `1`, then it'd speed up debugging to also show what the actual error message is (e.g. "invalid timestamp mode"). Just means that the developer won't have to go and run the script manually in bpftrace to work out the problem.

viktormalik: Makes sense, thanks!

viktormalik (approved):

tests/semantic_analyser.cpp
jordalgo: thanks for detailing, that's the exact justification I was hoping for 👍

PR #3002, opened 2024-02-12
Semantic Analyser: Update map key types on the first pass

ajor:
Follow up to #2999

This is more pre-work for for-each loops over maps, as it is necessary to know map key types when looping over elements.

This change required also setting map.skip_key_validation before the map is first visited.

The tests needed to be updated as the returned error code from SemanticAnalyser changed. Took the opportunity to check the error messages for them too.

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -431,6 +431,14 @@ void SemanticAnalyser::visit(Builtin &builtin)
   }
 }
 
+namespace {
+bool skip_key_validation(const Call &call)
+{
+  return call.func == "print" || call.func == "clear" || call.func == "zero" ||
+         call.func == "len";
+}
+} // namespace
+
 void SemanticAnalyser::visit(Call &call)
 {
   // Check for unsafe-ness first. It is likely the most pertinent issue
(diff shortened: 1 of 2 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -431,6 +431,16 @@ void SemanticAnalyser::visit(Builtin &builtin)
   }
 }
 
+namespace {
+bool skip_key_validation(const Call &call)
+{
+  if (call.func == "print" || call.func == "clear" || call.func == "zero" ||
+      call.func == "len")
+    return true;
+  return false;
```
jordalgo, line 440: style nit: maybe this instead
```suggestion
  return (call.func == "print" || call.func == "clear" || call.func == "zero" ||
      call.func == "len");
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -462,10 +472,15 @@ void SemanticAnalyser::visit(Call &call)
...
   if (call.vargs) {
     for (size_t i = 0; i < call.vargs->size(); ++i) {
-      auto &expr = (*call.vargs)[i];
+      auto &expr = *(*call.vargs)[i];
       func_arg_idx_ = i;
 
-      expr->accept(*this);
+      if (expr.is_map && skip_key_validation(call)) {
+        Map &map = static_cast<Map &>(expr);
+        map.skip_key_validation = true;
```
jordalgo, line 480: I guess `call.vargs` is always true for map expressions?

ajor: Yeah it will be, because we're looking for functions which are taking a map as an argument here (e.g. `clear(@map)`) and `vargs` holds the functions arguments. There's the check for `call.vargs` on line 473 so it should be safe if that's what you were thinking about

jordalgo (approved):

Issue #2963, opened 2024-01-26
RFC: New Looping Syntax


ajor:
This ticket is just to decide on a common syntax for looping. We have number of issues for various new types of loops we want to add and it would be good if they all looked like they were designed together. Implementation details can stay in their individual issues:

- #2955 
- #2807
- #2775

### Iterate over map keys - #2955 

Multi-key map iteration provides a tuple representing the key:
```
@x[1,2] = 3;
for ($key : @x) {
  printf("%d, %d = %d\n" $key.0, $key.1, @x[$key]);
}
```

Tuple-destructuring (not needed right away):
```
@x[1,2] = 3;
for ([$key1,$key2] : @x) {
  printf("%d, %d = %d\n" $key1, $key2, @x[$key1, $key2]);
}
```

Single-key map iteration automatically destructures tuple?
```
@x[1] = 2;
for ($key : @x) {
  printf("%d = %d\n" $key, @x[$key]);
}
```

### Iterate over integers - #2807

Iterating over a range of integers provides an integer:
```
for ($i : 0..10) {
  printf("%d\n", $i);
}
```

More structured than the existing while keyword; iteration and termination is enforced at the language level (instead of BPF verifier).

Would we also want to add a way to adjust the iteration step-size?

### Iterate over kernel lists - #2775

An alternative to the `iter:*` probes, to allow iteration within other probes. Iterating over these lists provides whatever type their BPF iterators return (`task*`, `task_vma*`, etc):
```
for ($task : iter(task)) {
  printf("%s, %d\n", $task->comm, $task->pid);
}
```

(I'm not completely convinced by the `iter(task)` syntax)

viktormalik:
I like the `for (item : range)` syntax, it's common in many languages, so I'd stick to that.

> Tuple-destructuring (not needed right away):
> ```
> @x[1,2] = 3;
> for ([$key1,$key2] : @x) {
>   printf("%d, %d = %d\n" $key1, $key2, @x[$key1, $key2]);
> }
> ```

Here, I'd prefer parentheses instead of square brackets:
```
for (($key1,$key2) : @x) {
```
The reason is that this is the way we represent tuples in bpftrace. Truth is that there's no parentheses in the map definition, but I consider it "implicit" in case of compound map keys. We could eventually add tuple destructuring for normal assignments, too:
```
@x = (1, 2)
($a, $b) = @x
```

> (I'm not completely convinced by the iter(task) syntax)

[...]

PR #3003, opened 2024-02-12
Add for-each loops for iterating over map elements

ajor:
Basic support for #2955. Support for sharing variables will be added in a separate PR: #3014

For loops are implemented using the bpf_for_each_map_elem helper function, which requires them to be rewritten into a callback style.

In this first implementation, we do not allow any variables to be shared between the main probe and the loop's body. Maps are global so are not a problem.

Pseudo code for the transformation we apply:

Before:
```
PROBE {
  @map[0] = 1;
  for ($kv : @map) {
    [LOOP BODY]
  }
}
```

After:
```
PROBE {
  @map[0] = 1;
  bpf_for_each_map_elem(@map, &map_for_each_cb, 0, 0);
}
long map_for_each_cb(bpf_map *map, const void *key, void *value, void *ctx) {
  $kv = ((uint64)key, (uint64)value);
  [LOOP BODY]
}
```

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -541,6 +541,39 @@ Characters and strings support the following escape sequences:
 
 === Loops
 
+==== For
+
+With Linux 5.13 and later, `for` loops can be used to iterate over elements in a map.
+
+----
+for ($kv : @map) {
+  block;
+}
+----
+
+The variable declared in the `for` loop will be initialised on each iteration with a tuple containing a key and a value from the map, i.e. `$kv = (key, value)`.
+
+----
+@map[10] = 20;
+for ($kv : @map) {
+  print($kv.0); // key
+  print($kv.1); // value
+}
+----
+
+When a map has multiple keys, the loop variable will be initialised with nested tuple of the form: `((key1, key2, ...), value)`
+
+----
+@map[10,11] = 20;
+for ($kv : @map) {
+  print($kv.0.0); // key 1
+  print($kv.0.1); // key 2
+  print($kv.1);   // value
+}
+----
+
+==== While
+
 Since kernel 5.3 BPF supports loops as long as the verifier can prove they're bounded and fit within the instruction limit.
 
 In bpftrace, loops are available through the `while` statement.
@@ -577,6 +610,8 @@ i:s:1 {
   printf("\n");
 }
 ----
+
+==== Unroll
 
 Loop unrolling is also supported with the `unroll` statement.
 
--- a/src/ast/passes/collect_nodes.h
+++ b/src/ast/passes/collect_nodes.h
@@ -1,0 +1,45 @@
+#pragma once
+
+#include <functional>
+#include <vector>
+
+#include "ast/visitors.h"
+
+namespace bpftrace::ast {
+
+/*
+ * CollectNodes
+ *
+ * Recurses into the provided node and builds a list of all descendants of the
+ * requested type which match a predicate.
+ */
+template <typename NodeT>
+class CollectNodes : public Visitor {
+public:
+  void run(
+      Node &node,
+      std::function<bool(const NodeT &)> pred = [](const auto &) {
+        return true;
+      })
+  {
+    pred_ = pred;
+    node.accept(*this);
+  }
+  const std::vector<std::reference_wrapper<NodeT>> nodes() const
+  {
+    return nodes_;
+  }
+
+private:
+  void visit(NodeT &node) override
+  {
+    if (pred_(node)) {
+      nodes_.push_back(node);
+    }
+  }
+
+  std::vector<std::reference_wrapper<NodeT>> nodes_;
+  std::function<bool(const NodeT &)> pred_;
+};
+
+} // namespace bpftrace::ast
--- a/src/ast/passes/collect_nodes.h
+++ b/src/ast/passes/collect_nodes.h
@@ -25,7 +25,7 @@ public:
     pred_ = pred;
     node.accept(*this);
   }
-  const std::vector<std::reference_wrapper<NodeT>> nodes() const
+  const std::vector<std::reference_wrapper<NodeT>> &nodes() const
   {
     return nodes_;
   }
--- a/src/ast/passes/printer.cpp
+++ b/src/ast/passes/printer.cpp
@@ -12,8 +12,9 @@ namespace ast {
 
 void Printer::print(Node *root)
 {
-  depth_ = 0;
+  ++depth_;
   Visit(*root);
+  --depth_;
 }
 
 std::string Printer::type(const SizedType &ty)
@@ -348,6 +349,24 @@ void Printer::visit(While &while_block)
   for (Statement *stmt : *while_block.stmts) {
     stmt->accept(*this);
   }
+}
+
+void Printer::visit(For &for_loop)
+{
+  std::string indent(depth_, ' ');
+  out_ << indent << "for" << std::endl;
+
+  ++depth_;
+  out_ << indent << " decl\n";
+  print(for_loop.decl);
+  out_ << indent << " expr\n";
+  print(for_loop.expr);
+
+  out_ << indent << " stmts\n";
+  for (Statement *stmt : *for_loop.stmts) {
+    print(stmt);
+  }
+  --depth_;
 }
 
 void Printer::visit(Config &config)
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -10,6 +10,7 @@
 #include "arch/arch.h"
 #include "ast/ast.h"
 #include "ast/signal_bt.h"
+#include "collect_nodes.h"
 #include "config.h"
 #include "log.h"
 #include "printf.h"
@@ -954,7 +955,7 @@ void SemanticAnalyser::visit(Call &call)
     check_nargs(call, 0);
   } else if (call.func == "print") {
     check_assignment(call, false, false, false);
-    if (in_loop() && is_final_pass()) {
+    if (in_loop() && is_final_pass() && call.vargs->at(0)->is_map) {
       LOG(WARNING, call.loc, out_)
           << "Due to it's asynchronous nature using 'print()' in a loop can "
              "lead to unexpected behavior. The map will likely be updated "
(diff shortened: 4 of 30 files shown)


src/ast/passes/printer.cpp
```diff
@@ -12,8 +12,9 @@ namespace ast {
 
 void Printer::print(Node *root)
 {
-  depth_ = 0;
+  ++depth_;
```
jordalgo, line 15: Just curious, why increment the initial depth before visiting the root?

jordalgo: NM, I see what you're doing.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1914,6 +1915,127 @@ void SemanticAnalyser::visit(While &while_block)
   loop_depth_--;
 }
 
+void SemanticAnalyser::visit(For &f)
+{
+  bool error = false;
+
+  if (!bpftrace_.feature_->has_helper_for_each_map_elem()) {
+    LOG(ERROR, f.loc, err_)
+        << "Missing required kernel feature: for_each_map_elem";
```
jordalgo, line 1924: You mention the kernel version in the docs, want to also mention it here?

tests/semantic_analyser.cpp
```diff
@@ -3006,6 +3006,254 @@ TEST_F(semantic_analyser_btf, fentry)
...
 
+TEST(semantic_analyser, for_loop_map_one_key)
+{
+  test("BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }", R"(
+Program
+ BEGIN
+  =
+   map: @map :: [int64]
+    int: 0 :: [int64]
+   int: 1 :: [int64]
```
jordalgo, line 3017: This indentation looks off.

tests/semantic_analyser.cpp
```diff
@@ -3006,6 +3006,254 @@ TEST_F(semantic_analyser_btf, fentry)
...
+}
+
+TEST(semantic_analyser, for_loop_map)
+{
+  test("BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }");
+  test("BEGIN { @map[0] = 1; for ($kv : @map) { print($kv.0); } }");
+  test("BEGIN { @map[0] = 1; for ($kv : @map) { print($kv.1); } }");
+}
+
+TEST(semantic_analyser, for_loop_map_assign_decl_to_var)
```
jordalgo, line 3057: Should we add the negative test to show it errors with the wrong type.

tests/semantic_analyser.cpp
```diff
@@ -3006,6 +3006,254 @@ TEST_F(semantic_analyser_btf, fentry)
...
+    BEGIN {
+      $var = 0;
+      @map[0] = 1;
+      for ($kv : @map) {
+        print($var);
+      }
+      print($var);
+    })",
+             R"(
+stdin:5:9-19: ERROR: Variables defined outside of a for-loop can not be brought into the loop's scope
```
jordalgo, line 3110: nit
```suggestion
stdin:5:9-19: ERROR: Variables defined outside of a for-loop can not be accessed in the loop's scope
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1914,6 +1915,127 @@ void SemanticAnalyser::visit(While &while_block)
...
+
+  // Validate body
+  CollectNodes<Variable> vars_referenced;
+  for (auto *stmt : *f.stmts) {
+    vars_referenced.run(*stmt);
+  }
+  for (const Variable &var : vars_referenced.nodes()) {
+    if (variable_val_[probe_].find(var.ident) != variable_val_[probe_].end()) {
+      LOG(ERROR, var.loc, err_) << "Variables defined outside of a for-loop "
+                                   "can not be brought into the loop's scope";
```
jordalgo, line 1983: nit
```suggestion
                                   "can not be accessed in the loop's scope";
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1914,6 +1915,127 @@ void SemanticAnalyser::visit(While &while_block)
...
+  }
+
+  // This could be relaxed in the future:
+  CollectNodes<Jump> jumps;
+  for (auto *stmt : *f.stmts) {
+    jumps.run(*stmt);
+  }
+  for (const Jump &n : jumps.nodes()) {
+    LOG(ERROR, n.loc, err_)
+        << "'" << opstr(n) << "' statement is not allowed in a for loop";
```
jordalgo, line 1995:
```suggestion
        << "'" << opstr(n) << "' statement is not allowed in a for-loop";
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -920,7 +921,7 @@ void SemanticAnalyser::visit(Call &call)
     check_nargs(call, 0);
   } else if (call.func == "print") {
     check_assignment(call, false, false, false);
-    if (in_loop() && is_final_pass()) {
+    if (in_loop() && is_final_pass() && call.vargs->at(0)->is_map) {
```
jordalgo, line 924: JC, what is this additional check for?

tests/runtime/for
```diff
@@ -0,0 +1,33 @@
...
+EXPECT key: (16,17), val: 32
+TIMEOUT 5
+
+NAME map strings
+PROG BEGIN { @map["abc"] = "aa"; @map["def"] = "dd"; for ($kv : @map) { print($kv); } exit(); }
+EXPECT (abc, aa)
+EXPECT (def, dd)
+TIMEOUT 5
+
+NAME map create map in body
```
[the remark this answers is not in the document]
jordalgo: Can we also add a test for two maps and two loops?

man/adoc/bpftrace.adoc
```diff
@@ -856,6 +856,27 @@ if (condition) {
...
+----
+for ($kv : @map) {
+  block;
+}
+----
+
+The variable declared in the `for` loop will be initialised on each iteration with a tuple containing a key and a value from the map, i.e. `$kv = (key, value)`.
+
+----
+for ($kv : @map) {
```
jordalgo, line 872: I think also showing an example with multiple keys would be good too.

jordalgo:
Overall this is looking good! Few minor nits/questions.

tests/semantic_analyser.cpp
ajor: In the AST, the value being assigned to the map is a child of the AssignMapStatement node (shown as "=" by Printer).

This is the tree that we're trying to show:

```
     AssignMapStatement (=)
      /                \
     / LHS              \ RHS
    /                    \
  Map (@map)          Integer (1)
   |
   | key
   |
Integer (0)
```

src/ast/passes/semantic_analyser.cpp
ajor: Our feature detection logic means that this can run on older kernels if they've backported the required helper, so mentioning it in the error message could be misleading.

I wasn't sure about including it in the docs, but put it in because the docs for while loops mentioned it too. It's nice to have a rough idea of when a feature is available, but backporting happens in a few distros so having it in the docs could be misleading too... 🤷‍♂️

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1914,6 +1915,127 @@ void SemanticAnalyser::visit(While &while_block)
...
+  // Iterating over a map provides a tuple: (map_key, map_val)
+  auto *mapkey = get_map_key_type(map);
+  auto *mapval = get_map_type(map);
+
+  if (mapkey && mapkey->args_.size() == 0) {
+    LOG(ERROR, map.loc, err_)
+        << "Maps used as for-loop expressions must have keys to iterate over";
+  }
+
+  if (mapval && mapkey) {
```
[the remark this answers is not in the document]
ajor: It can happen before type-propagation is complete if a map is defined after being used in a for-loop, e.g.:

`BEGIN { for ($kv : @map) {} } END { @map[0] = 1; }`

`@map` doesn't exist on the first semantic analyser pass.

src/ast/passes/semantic_analyser.cpp
ajor: `print()` comes in two forms: print an entire map or print a single value.

Only the first form is asynchronous, so the second is safe to use inside loops. The error message below was already assuming users were printing an entire map (1st form), so really this check had been wrong for a long time, but I just noticed it when I was testing for-loops.

src/ast/visitors.cpp
```diff
@@ -157,6 +157,16 @@ void Visitor::visit(While &while_block)
   }
 }
 
+void Visitor::visit(For &for_loop)
```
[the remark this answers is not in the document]
ajor: Yeah that sounds like a reasonable optimisation to do. I think right now it'd have to be done in CodeGen. I'll leave it out of this PR as it's already complicated enough though.

This is the kind of thing where it'd be nice to have an AST rewriter: we could have a really simple pass which just deletes empty for-loops and CodeGen wouldn't need any extra logic.

tests/runtime/for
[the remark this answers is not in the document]
ajor: JSON output I'd say is orthogonal to for-loops. As long a text output works with for-loops, and JSON output works without for-loops, then JSON output will work with for-loops, so no need to test the combination.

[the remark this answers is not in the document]
ajor: Testing two loops: good idea! I discovered a bug using nested loops just now:
`BEGIN { @map1[0] = 1; @map2[0] = 1; for ($kv1 : @map1) { for ($kv2 : @map2) {} } }`

This triggers a verifier error: "call to invalid destination". Will need to look into what that means.

tests/semantic_analyser.cpp
ajor: Like JSON output, I think variable assignments are generally orthogonal to for-loops. I can't actually remember why I added this test in the first place. Will remove it if I can't work it out.

I'd think that all we need to check is that `$kv` has the correct type, and the test case `for_loop_map_one_key` is already doing that by printing out the types in the AST.

ajor:
Addressed most of the comments. Still need to add tests for multiple for-loops and fix the nested for-loop bug.

src/ast/passes/semantic_analyser.cpp
jordalgo: I think it's ok to be verbose in these scenarios e.g. mention the kernel version but it may be specific to the distro if the changes were backported.

tests/semantic_analyser.cpp
jordalgo: Ah yeah, my bad, I wasn't noticing the equal assignment being the parent.

src/ast/visitors.cpp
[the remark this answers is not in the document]
jordalgo: > This is the kind of thing where it'd be nice to have an AST rewriter: we could have a really simple pass which just deletes empty for-loops and CodeGen wouldn't need any extra logic.

Yah, that would be cool. I can maybe take a look at doing that in a follow-up (just for my own education).

src/ast/passes/semantic_analyser.cpp
jordalgo: Ah gotcha. Thanks for the explanation.

ajor:
Added tests for multiple loops in various scenarios.

Nested loops requires another relocation fix: #3020
Loops used from two different main programs is causing yet another error! 
> number of funcs in func_info doesn't match number of subprogs

This one appears to have been around for longer and affects the len() builtin as well:
`BEGIN { @map[0] = 1; print(len(@map)) } END { print(len(@map)) }`

ajor:
The loops-in-multiple-probes issue is pretty nasty: #3021

src/ast/passes/collect_nodes.h
```diff
@@ -0,0 +1,45 @@
...
+  void run(
+      Node &node,
+      std::function<bool(const NodeT &)> pred = [](const auto &) {
+        return true;
+      })
+  {
+    pred_ = pred;
+    node.accept(*this);
+  }
+  const std::vector<std::reference_wrapper<NodeT>> nodes() const
```
jordalgo, line 28: nit: I prefer `get_nodes`

jordalgo (approved):
This looks great. Nice work! 🚀

danobi:
I would like to take a look but I'll be busy until ~Tuesday.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1914,6 +1915,127 @@ void SemanticAnalyser::visit(While &while_block)
...
+    jumps.run(*stmt);
+  }
+  for (const Jump &n : jumps.nodes()) {
+    LOG(ERROR, n.loc, err_)
+        << "'" << opstr(n) << "' statement is not allowed in a for-loop";
+    error = true;
+  }
+
+  if (error)
+    return;
```
viktormalik, line 2000: Nit: you could do the same as semantic analyser does at the end:
```suggestion
  if (!err_.empty())
    return;
```
That would spare the `error` variable.

Maybe we could even introduce some wrapper like `has_error` as it may be useful in other places, too.

viktormalik (commented):
Just a couple of small comments, otherwise looks great!

viktormalik:
Also codegen tests need update after #2985 was merged.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1914,6 +1915,127 @@ void SemanticAnalyser::visit(While &while_block)
...
+
+  // Validate expr
+  if (!f.expr->is_map) {
+    LOG(ERROR, f.expr->loc, err_) << "Loop expression must be a map";
+    error = true;
+  }
+  Map &map = static_cast<Map &>(*f.expr);
+
+  // Validate body
+  CollectNodes<Variable> vars_referenced;
```
danobi, line 1976: Neat :)

danobi (commented):
overall looks pretty reasonable. didn't get a chance to look carefully let

man/adoc/bpftrace.adoc
```diff
@@ -541,6 +541,39 @@ Characters and strings support the following escape sequences:
 
 === Loops
 
+==== For
+
+With Linux 5.13 and later, `for` loops can be used to iterate over elements in a map.
+
+----
+for ($kv : @map) {
```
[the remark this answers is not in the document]
ajor: There's some discussion on syntax in #2963

src/ast/passes/collect_nodes.h
ajor: I've got to disagree on this one :)

ajor:
Addressed review comments and added extra runtime test for deleting elements while iterating over the map.

ajor:
Fixed a lifetime bug I'd introduced for tuple elements: in `CodegenLLVM::visit(Tuple &tuple)` I'm now extending the lifetime of all tuple elements until the end of the function with:
```
scoped_dels.emplace_back(std::move(accept(elem)));
```

Thanks @arthurshau for the new test case which exposed this! (`call.print_map_item_tuple`).

Will merge tomorrow if there are no objections.

viktormalik (approved):

ajor:
A lot of the error messages I'm testing for in this PR appear to have regressed: #3063

ajor:
Updated the tests to accept the incorrect expected results and left comments referring to #3063

PR #3004, opened 2024-02-12
Allow keywords to be used as idents

jordalgo:
This list of keywords that can now be used for postfix expressions and in the offsetof expression e.g.
```
$x->keyword
$x.keyword
offsetof(*curtask, keyword)
```

List:
- "break"
- "config"
- "continue"
- "else"
- "for"
- "if"
- "offsetof"
- "return"
- "sizeof"
- "unroll"
- "while"

##### Checklist

### DIFF
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -2141,6 +2141,28 @@ TEST(Parser, config_error)
   test_parse_failure("BPFTRACE_STACK_MODE=perf; i:s:1 { exit(); }");
 }
 
+TEST(Parser, keywords_as_identifiers)
+{
+  std::vector<std::string> keywords = { "break",    "config", "continue",
+                                        "else",     "for",    "if",
+                                        "offsetof", "return", "sizeof",
+                                        "unroll",   "while" };
+  for (const auto &keyword : keywords) {
+    test("BEGIN { $x = (struct Foo*)0; $x->" + keyword + "; }",
+         "Program\n BEGIN\n  =\n   variable: $x\n   (struct Foo *)\n    int: "
+         "0\n "
+         " .\n   dereference\n    variable: $x\n   " +
+             keyword + "\n");
+    test("BEGIN { $x = (struct Foo)0; $x." + keyword + "; }",
+         "Program\n BEGIN\n  =\n   variable: $x\n   (struct Foo)\n    int: 0\n "
+         " .\n   variable: $x\n   " +
+             keyword + "\n");
+    test("BEGIN { $x = offsetof(*curtask, " + keyword + "); }",
+         "Program\n BEGIN\n  =\n   variable: $x\n   offsetof: \n    "
+         "dereference\n     builtin: curtask\n");
+  }
+}
+
 } // namespace parser
 } // namespace test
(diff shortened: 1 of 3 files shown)


jordalgo:
I had trouble getting "enum", "struct", and "union" to work. I can look into it for a follow-up diff.

danobi:
If it's complicated and/or not very clean to implement, I'd vote to skip `struct`, `enum`, and `union`. Cuz you can't use those idents as struct member names anyways

tests/parser.cpp
```diff
@@ -2141,6 +2141,28 @@ TEST(Parser, config_error)
...
+         "0\n "
+         " .\n   dereference\n    variable: $x\n   " +
+             keyword + "\n");
+    test("BEGIN { $x = (struct Foo)0; $x." + keyword + "; }",
+         "Program\n BEGIN\n  =\n   variable: $x\n   (struct Foo)\n    int: 0\n "
+         " .\n   variable: $x\n   " +
+             keyword + "\n");
+    test("BEGIN { $x = offsetof(*curtask, " + keyword + "); }",
+         "Program\n BEGIN\n  =\n   variable: $x\n   offsetof: \n    "
+         "dereference\n     builtin: curtask\n");
```
ajor, line 2162: This looks like a bug in the AST Printer - it's not showing the field used in offsetof

ajor (approved):
Looks good to me!

There's still one more case I can see were keywords conflict user-defined types but they can be done later in a separate PR. Keywords used as type names:
```
# bpftrace -e 'BEGIN { (config*)0; }'
stdin:1:9-16: ERROR: syntax error, unexpected config
BEGIN { (config*)0; }
        ~~~~~~~
# bpftrace -e 'BEGIN { (comm*)0; }'
stdin:1:9-16: ERROR: syntax error, unexpected )
BEGIN { (comm*)0; }
        ~~~~~~~
```

But this has apparently been difficult to sort out before: https://github.com/bpftrace/bpftrace/pull/2560#issuecomment-1524799774

Also I was surprised to discover that this actually works: `kprobe:while { }`.
I think it's due to a happy parsing coincidence where everything in the `attach_point_def` rule takes the `attach_point_def PATH` option. Maybe means that most of the other `attach_point_def` options are redundant?

ajor:
> If it's complicated and/or not very clean to implement, I'd vote to skip `struct`, `enum`, and `union`. Cuz you can't use those idents as struct member names in C/C++ anyways

We should try and sort them out at some point though - `enum` and `union` are valid type/field names in Go, which is a language that we otherwise support.

tests/parser.cpp
jordalgo: Yeah, I was confused by that too. I'll look into it.

jordalgo:
> There's still one more case I can see were keywords conflict user-defined types but they can be done later in a separate PR.

I spent a little time looking at this but couldn't figure it out, so I'll work on it for a separate PR.

> Maybe means that most of the other attach_point_def options are redundant?

Yeah that's interesting. I'll look into that as well in a follow-up. All this stuff is still relatively new to me.

PR #3013, opened 2024-02-21
Fix offsetof ast print

jordalgo:
Print the field that we want the offsetof.

Follow up: https://github.com/bpftrace/bpftrace/pull/3004#discussion_r1490937083

##### Checklist


tests/parser.cpp
```diff
@@ -1745,7 +1745,8 @@ TEST(Parser, offsetof_type)
        "\n"
        "Program\n"
        " BEGIN\n"
-       "  offsetof: \n");
+       "  offsetof: \n"
```
danobi, line 1748: shouldn't `struct Foo` be somewhere here?

tests/parser.cpp
```diff
@@ -1762,7 +1763,8 @@ TEST(Parser, offsetof_expression)
        "    int: 0\n"
        "  offsetof: \n"
        "   dereference\n"
-       "    variable: $foo\n");
+       "    variable: $foo\n"
+       "    field: x\n");
```
ajor, line 1767: No need for the additional `++depth_` in the printer - `field` should be indented to the same level as the expr (dereference here) as it's a direct child of the `offsetof`

tests/parser.cpp
jordalgo: was also curious, why this was missing - I think it's a separate issue I need to debug

jordalgo: Fixed in the latest version.

jordalgo:
Sorry for the late-night, half-assed fix. This newest version should be better.

ajor:
> Sorry for the late-night, half-assed fix.

We all do it!

PR #3014, opened 2024-02-21
For loops: Allow sharing variables with main program

ajor:
This lets programs such as this run:
```
BEGIN
{
  @map[0] = 1; @map[1] = 2;
  $cnt = 0;
  for ($kv : @map) {
    $cnt++; // Previously would have errored with "Variables defined outside of a for-loop can not be accessed in the loop's scope"
  }
  print($cnt);
}
```

#### Implementation Details
1. Determine which variables need to be shared with the loop callback
2. Pack pointers to them into a context struct
3. Pass pointer to the context struct to the callback function
4. In the callback, override the shared variables so that they read and write through the context pointers instead of directly from their original addresses

Pseudo code for the transformation we apply:

Before:
```
PROBE {
  $str = "hello";
  $not_shared = 2;
  $len = 0;
  @map[11, 12] = "c";
  for ($kv : @map) {
    print($str);
    $len++;
  }
  print($len);
  print($not_shared);
}
```

After:
```
struct ctx_t {
  string *str;
  uint64 *len;
};
PROBE {
  $str = "hello";
  $not_shared = 2;
  $len = 0;
  @map[11, 12] = "c";

  ctx_t ctx { .str = &$str, .len = &$len };
  bpf_for_each_map_elem(@map, &map_for_each_cb, &ctx, 0);

  print($len);
  print($not_shared);
}
long map_for_each_cb(bpf_map *map,
                     const void *key,
                     void *value,
                     void *ctx) {
  $kv = (((uint64, uint64))key, (string)value);
  $str = ((ctx_t*)ctx)->str;
  $len = ((ctx_t*)ctx)->len;

  print($str);
  $len++;
}
```

##### Checklist

N/A: For-loops haven't been released yet and this is basic functionality expected of for-loops so is covered by the existing docs/changelog.
~- [ ] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`~
~- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`~

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1426,10 +1426,10 @@ void CodegenLLVM::visit(Variable &var)
   // Arrays and structs are not memcopied for local variables
   if (needMemcpy(var.type) &&
       !(var.type.IsArrayTy() || var.type.IsRecordTy())) {
-    expr_ = variables_[var.ident];
+    expr_ = variables_[var.ident].value;
   } else {
-    auto *var_alloca = variables_[var.ident];
-    expr_ = b_.CreateLoad(var_alloca->getAllocatedType(), var_alloca);
+    auto &var_llvm = variables_[var.ident];
+    expr_ = b_.CreateLoad(var_llvm.type, var_llvm.value);
   }
 }
 
@@ -2289,22 +2289,22 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
     }
 
     AllocaInst *val = b_.CreateAllocaBPFInit(alloca_type, var.ident);
-    variables_[var.ident] = val;
+    variables_[var.ident] = VariableLLVM{ val, val->getAllocatedType() };
   }
 
   if (var.type.IsArrayTy() || var.type.IsRecordTy()) {
     // For arrays and structs, only the pointer is stored
     b_.CreateStore(b_.CreatePtrToInt(expr_, b_.getInt64Ty()),
-                   variables_[var.ident]);
+                   variables_[var.ident].value);
     // Extend lifetime of RHS up to the end of probe
     scoped_del.disarm();
   } else if (needMemcpy(var.type)) {
-    auto *val = variables_[var.ident];
+    auto *val = variables_[var.ident].value;
     if (assignment.expr->type.GetSize() != var.type.GetSize())
       b_.CreateMemsetBPF(val, b_.getInt8(0), var.type.GetSize());
     b_.CREATE_MEMCPY(val, expr_, assignment.expr->type.GetSize(), 1);
   } else {
-    b_.CreateStore(expr_, variables_[var.ident]);
+    b_.CreateStore(expr_, variables_[var.ident].value);
   }
 }
 
@@ -2464,8 +2464,41 @@ void CodegenLLVM::visit(For &f)
   auto &map = static_cast<Map &>(*f.expr);
 
   Value *ctx = b_.getInt64(0);
+  llvm::Type *ctx_t = nullptr;
+
+  const auto &ctx_fields = f.ctx_type.GetFields();
+  if (!ctx_fields.empty()) {
+    // Pack pointers to variables into context struct for use in the callback
+
+#if LLVM_VERSION_MAJOR < 15
+    std::vector<llvm::Type *> ctx_field_types;
+    ctx_field_types.reserve(ctx_fields.size());
+    for (const auto &field : ctx_fields) {
+      ctx_field_types.push_back(b_.GetType(field.type)->getPointerTo());
+    }
+#else
+    std::vector<llvm::Type *> ctx_field_types(ctx_fields.size(),
+                                              b_.GET_PTR_TY());
+#endif
+    ctx_t = b_.GetStructType("ctx_t", ctx_field_types);
+    ctx = b_.CreateAllocaBPF(ctx_t, "ctx");
+
+    for (size_t i = 0; i < ctx_fields.size(); i++) {
+      const auto &field = ctx_fields[i];
+      auto *field_expr = variables_[field.name].value;
+      auto *ctx_field_ptr = b_.CreateGEP(
+          ctx_t, ctx, { b_.getInt64(0), b_.getInt32(i) }, "ctx." + field.name);
+#if LLVM_VERSION_MAJOR < 15
+      // An extra cast is required for older LLVM versions, pre-opaque-pointers
+      ctx_field_ptr = b_.CreatePointerCast(
+          ctx_field_ptr, field_expr->getType()->getPointerTo());
+#endif
+      b_.CreateStore(field_expr, ctx_field_ptr);
+    }
+  }
+
   b_.CreateForEachMapElem(
-      ctx_, map, createForEachMapCallback(map, *f.decl, *f.stmts), ctx, f.loc);
+      ctx_, map, createForEachMapCallback(f, ctx_t), ctx, f.loc);
 }
 
 void CodegenLLVM::visit(Predicate &pred)
@@ -2617,7 +2650,8 @@ void CodegenLLVM::visit(Subprog &subprog)
   for (SubprogArg *arg : *subprog.args) {
     auto alloca = b_.CreateAllocaBPF(b_.GetType(arg->type), arg->name());
     b_.CreateStore(func->getArg(arg_index + 1), alloca);
-    variables_.insert({ arg->name(), alloca });
+    variables_[arg->name()] = VariableLLVM{ alloca,
+                                            alloca->getAllocatedType() };
     ++arg_index;
   }
 
@@ -3917,14 +3951,14 @@ void CodegenLLVM::createIncDec(Unop &unop)
     b_.CreateLifetimeEnd(newval);
   } else if (unop.expr->is_variable) {
     Variable &var = static_cast<Variable &>(*unop.expr);
-    Value *oldval = b_.CreateLoad(variables_[var.ident]->getAllocatedType(),
-                                  variables_[var.ident]);
+    Value *oldval = b_.CreateLoad(variables_[var.ident].type,
+                                  variables_[var.ident].value);
     Value *newval;
     if (is_increment)
       newval = b_.CreateAdd(oldval, b_.GetIntSameSize(step, oldval));
     else
       newval = b_.CreateSub(oldval, b_.GetIntSameSize(step, oldval));
-    b_.CreateStore(newval, variables_[var.ident]);
+    b_.CreateStore(newval, variables_[var.ident].value);
 
     if (unop.is_post_op)
       expr_ = oldval;
@@ -4139,10 +4173,7 @@ Function *CodegenLLVM::createMapLenCallback()
   return callback;
 }
 
-Function *CodegenLLVM::createForEachMapCallback(
-    Map &map,
-    const Variable &decl,
-    const std::vector<Statement *> &stmts)
+Function *CodegenLLVM::createForEachMapCallback(const For &f, llvm::Type *ctx_t)
 {
   /*
    * Create a callback function suitable for passing to bpf_for_each_map_elem,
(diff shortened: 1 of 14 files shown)


src/ast/passes/printer.cpp
```diff
@@ -351,19 +352,27 @@ void Printer::visit(While &while_block)
   }
 }
 
-void Printer::visit(For &for_loop)
+void Printer::visit(For &f)
```
jordalgo, line 355: Nit: I know `f` is more terse but I much prefer the variable name `for_loop`.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2038,6 +2034,62 @@ void SemanticAnalyser::visit(For &f)
...
+   *       for ($kv : @map) {
+   *         print($str);
+   *         $len++;
+   *       }
+   *       print($len);
+   *       print($not_shared);
+   *     }
+   *
+   * After:
+   *     struct ctx_t {
```
jordalgo, line 2066: I haven't read the code thoroughly enough yet but is there a possible collision here if there are two (or more) probes that have scratch variables with the same names

src/struct.h
```diff
@@ -176,7 +182,7 @@ class StructManager {
 
 private:
   std::map<std::string, std::shared_ptr<Struct>> struct_map_;
-  std::unordered_set<std::shared_ptr<Struct>> tuples_;
+  std::unordered_set<std::shared_ptr<Struct>> anonymous_types_;
```
jordalgo, line 185: naming nit: maybe make this consistent with the method e.g. `anon_types_` or  `AddAnonymousStruct`

tests/semantic_analyser.cpp
```diff
@@ -197,12 +197,21 @@ void test(std::string_view input, std::string_view expected_ast)
...
 
-  if (!expected_ast.empty() && expected_ast[0] == '\n')
+  if (expected_ast[0] == '\n')
     expected_ast.remove_prefix(1); // Remove initial '\n'
 
   std::ostringstream out;
   ast::Printer printer(out);
   printer.print(driver.root.get());
+
+  if (expected_ast[0] == '*' && expected_ast[expected_ast.size() - 1] == '*') {
```
jordalgo, line 207: JC, what is this block doing?

src/ast/passes/codegen_llvm.h
```diff
@@ -259,7 +258,11 @@ class CodegenLLVM : public Visitor {
   int current_usdt_location_index_{ 0 };
   bool inside_subprog_ = false;
 
-  std::map<std::string, AllocaInst *> variables_;
+  struct VariableLLVM {
+    llvm::Value *value;
+    llvm::Type *type;
+  };
+  std::map<std::string, VariableLLVM> variables_;
```
jordalgo, line 265: Since this is class level is there a case of possible collision where two (or more probes) use the same scratch variable name and reference them in a loop?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2471,8 +2471,41 @@ void CodegenLLVM::visit(For &f)
...
+      ctx_field_types.push_back(b_.GetType(field.type)->getPointerTo());
+    }
+#else
+    std::vector<llvm::Type *> ctx_field_types(ctx_fields.size(),
+                                              b_.GET_PTR_TY());
+#endif
+    ctx_t = b_.GetStructType("ctx_t", ctx_field_types);
+    ctx = b_.CreateAllocaBPF(ctx_t, "ctx");
+
+    for (size_t i = 0; i < ctx_fields.size(); i++) {
```
jordalgo, line 2493: super duper nit
```suggestion
    for (size_t i = 0; i < ctx_fields.size(); ++i) {
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2038,6 +2034,62 @@ void SemanticAnalyser::visit(For &f)
...
+   * program, some extra steps are taken:
+   *
+   * 1. Determine which variables need to be shared with the loop callback
+   * 2. Pack pointers to them into a context struct
+   * 3. Pass pointer to the context struct to the callback function
+   * 4. In the callback, override the shared variables so that they read and
+   *    write through the context pointers instead of directly from their
+   *    original addresses
+   *
+   * Example transformation with context:
```
jordalgo, line 2049: I like the example!

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2082,6 +2123,35 @@ void SemanticAnalyser::visit(For &f)
...
 
+  // Collect a list of unique variables which are referenced in the loop's body
+  // and declared before the loop. These will be passed into the loop callback
+  // function as the context parameter.
+  CollectNodes<Variable> vars_referenced;
+  std::unordered_set<std::string> var_set;
+  for (auto *stmt : *f.stmts) {
+    const auto &live_vars = variable_val_[scope_];
+    vars_referenced.run(*stmt, [&live_vars, &var_set](const auto &var) {
+      if (live_vars.find(var.ident) == live_vars.end())
```
jordalgo, line 2134: Maybe I'm missing something but why is this check needed?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2082,6 +2123,35 @@ void SemanticAnalyser::visit(For &f)
...
+        return false;
+      if (var_set.find(var.ident) != var_set.end())
```
jordalgo, line 2136: I think you can safely insert the same string multiple times.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2082,6 +2123,35 @@ void SemanticAnalyser::visit(For &f)
   if (has_error())
     return;
 
+  // Collect a list of unique variables which are referenced in the loop's body
+  // and declared before the loop. These will be passed into the loop callback
+  // function as the context parameter.
+  CollectNodes<Variable> vars_referenced;
+  std::unordered_set<std::string> var_set;
+  for (auto *stmt : *f.stmts) {
```
jordalgo, line 2131: Nit: looks like you could combine these two loops to get referenced and new vars.

jordalgo (commented):
Few small questions and nits but otherwise LGTM.

src/ast/passes/codegen_llvm.h
```diff
@@ -259,7 +258,11 @@ class CodegenLLVM : public Visitor {
   int current_usdt_location_index_{ 0 };
   bool inside_subprog_ = false;
 
-  std::map<std::string, AllocaInst *> variables_;
+  struct VariableLLVM {
+    llvm::Value *value;
```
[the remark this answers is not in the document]
ajor: It's because on this line we we'll have a LoadInst instead of an AllocaInst: https://github.com/bpftrace/bpftrace/pull/3014/files#diff-28926c090177e6efbfb5a7449833e0e863bb226822abb513b7dfc6c388837d87R4257

We're loading a pointer from the ctx tuple and using that to access our variable's memory, instead of going directly through the original AllocaInst.

src/ast/passes/codegen_llvm.h
ajor: `variables_` is cleared between probes so they won't collide: https://github.com/bpftrace/bpftrace/blob/ca1b216ba92d007c1aa4aa1e46aa6270c8caadaf/src/ast/passes/codegen_llvm.cpp#L2547

src/ast/passes/semantic_analyser.cpp
ajor: I'd like to delete the "new_vars" loop in hopefully not too long (when #3017 is done), so was keeping them separate to make that easier

src/ast/passes/codegen_llvm.cpp
ajor: 🤮 

sorry, I prefer `i++`

ajor:
Addressed review comments.

Also added a new `AddrSpace::bpf` enum value to represent a pointer to data on the BPF stack (I was previously using `AddrSpace::none`).

src/ast/passes/semantic_analyser.cpp
jordalgo: Could always add a comment todo with a reference to the issue but that's minor. I won't block.

jordalgo:
Looks like the for-loop semantic analyzer tests are still failing. Fix those and this looks good to go.

src/ast/passes/semantic_analyser.cpp
ajor: The comment on the next for-loop does reference the issue

ajor:
I think I've fixed the test failures now. I was misreading them before - turns out I just needed to add "AS(bpf)" onto the expected results to take into account the new BPF address space I added.

viktormalik (approved):
LGTM, just one nit.

Also, one test on LLVM 12 is failing (not codegen) and may need some investigation.

viktormalik:
We also need to add a CHANGELOG entry since for-loops are now released, right?

src/ast/passes/codegen_llvm.h
```diff
@@ -273,7 +271,12 @@ class CodegenLLVM : public Visitor {
...
+    llvm::Type *type;
+  };
+  std::map<std::string, VariableLLVM> variables_;
```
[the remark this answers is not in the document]
ajor: Nope, just historical. I've changed it

ajor:
> Also, one test on LLVM 12 is failing (not codegen) and may need some investigation.

This is some craziness! The test shouldn't be any different across LLVM versions, and now I'm also seeing it fail on master on my local machine

ajor:
> > Also, one test on LLVM 12 is failing (not codegen) and may need some investigation.
> 
> This is some craziness! The test shouldn't be any different across LLVM versions, and now I'm also seeing it fail on master on my local machine

And the test is passing after a rebase...

viktormalik:
> > > Also, one test on LLVM 12 is failing (not codegen) and may need some investigation.
> > 
> > 
> > This is some craziness! The test shouldn't be any different across LLVM versions, and now I'm also seeing it fail on master on my local machine
> 
> And the test is passing after a rebase...

Strange. Could it be caused by a wrong generation of BTF test data? I've a suspicion that something quirky is happening there sometimes.

ajor:
> Strange. Could it be caused by a wrong generation of BTF test data? I've a suspicion that something quirky is happening there sometimes.

Yeah, I had a look at the BTF data while the test was failing and it was only showing 3 arguments for the `func_1` function instead of 4. I don't know how that happened though. After a clean build the BTF was showing 4 arguments again.

yump:
I am seeing that error message about "Variables defined outside of a for-loop can not be accessed in the loop's scope," that was not supposed to make it into a release, using version 0.21.2 as packaged in Fedora 41, and kernel 6.11.10.

It also happens with the latest v0.21.2 on the releases page:

```
> sudo /tmp/bpftrace -dd -k /tmp/test.bt
[DEBUG] libbpf: loaded kernel BTF from '/sys/kernel/btf/vmlinux'
[DEBUG] libbpf: loaded kernel BTF from '/sys/kernel/btf/vmlinux'

AST after: parser
-------------------
Program
 BEGIN
  =
   map: @map
    int: 0
   int: 1
  =
   map: @map
    int: 1
   int: 2
  =
   variable: $cnt
   int: 0
  for
   decl
    variable: $kv
   expr
    map: @map
   stmts
    variable: $cnt
     ++
  call: print
   variable: $cnt


AST after: ConfigAnalyser
-------------------
Program
 BEGIN
  =
   map: @map
    int: 0 :: [int64]
   int: 1 :: [int64]
  =
   map: @map
    int: 1 :: [int64]
   int: 2 :: [int64]
  =
   variable: $cnt
   int: 0 :: [int64]
  for
   decl
    variable: $kv
   expr
    map: @map
   stmts
    variable: $cnt
     ++
  call: print
   variable: $cnt

/tmp/test.bt:8:5-9: ERROR: Variables defined outside of a for-loop can not be accessed in the loop's scope
    $cnt++; // Previously would have errored with "Variables defined outside of a for-loop can not be accessed in the loop's scope"
    ~~~~
```

ajor:
This PR didn't make it into the 0.21 release - it'll be included in the upcoming 0.22.

PR #3019, opened 2024-02-26
docs: Add some guidelines for structs/classes/variables

danobi:
These help with maintaining some uniformity in code. I think the codebase is already pretty good about these rules, so it's just codifying it.

##### Checklist

### DIFF
--- a/docs/coding_guidelines.md
+++ b/docs/coding_guidelines.md
@@ -49,3 +49,43 @@ the kernel debug facilities. And we probably should not be mounting
 filesystems on behalf of the user. So at the point of failure, we should
 just throw an exception and let the user know their system is not ready
 for bpftrace.
+
+## Struct vs. Class
+
+Use a struct only for passive objects that carry data; everything else is a class.
+
+All fields in a struct should be public. The struct must not have invariants
+that imply relationships between different fields, since direct user access to
+those fields may break those invariants. Structs should not have any associated
+methods (including constructors and destructors).
+
+## Variable naming
+
+Variables should be in `snake_case` (all lowercase, with underscores between words).
+
+Private members should have a trailing underscore. Public members should not have
+a trailing underscore.
+
+Struct data members should _not_ have a trailing underscore.
+
+Examples:
+
+```c++
+class Foo {
+public:
+  int snake_case;      // good
+  int secondone;       // bad
+  int var_;            // bad
+  int camelCase;       // bad
+
+private:
+  int another_var_;    // good
+  int private_var;     // bad
+  int priv_;           // good
+};
+
+struct Bar {
+  int var;             // good
+  int tailing_;        // bad
+};
+```


docs/coding_guidelines.md
```diff
@@ -49,3 +49,20 @@ the kernel debug facilities. And we probably should not be mounting
...
+Use a struct only for passive objects that carry data; everything else is a class.
+
+All fields in a struct should be public. The struct must not have invariants
+that imply relationships between different fields, since direct user access to
+those fields may break those invariants. Structs should not have any associated
+methods (including constructors and destructors).
+
+## Variable naming
+
+Variables should be in `snake_case` (all lowercase, with spaces between words).
```
jordalgo, line 64:
```suggestion
Variables should be in `snake_case` (all lowercase, with underscores between words).
```

docs/coding_guidelines.md
```diff
@@ -49,3 +49,20 @@ the kernel debug facilities. And we probably should not be mounting
...
+
+Class data members should have a trailing underscore (for both public and private).
```
viktormalik, line 66: I don't want to start a dispute over coding conventions but what is the motivation for this? I always thought that the trailing underscore is to visually distinguish private members from the public ones.

jordalgo: TBH, I thought this too but then couldn't think of very many instances where we have a public class data members as the general c++ best practice is to hide them behind getter/setters.

viktormalik: We have quite a few in the `BPFtrace` class: `resources`, `maps`, `structs`, `macros_`, `enums_` (they don't even follow any rules). Same for all members of `RequiredResources` (where it doesn't really make sense to hide them behind getters/setters).

jordalgo: Sorry, my thought was completely muddled. I was trying to agree with you @viktormalik but also mention that perhaps we should abide by some best practices and try to hide these members behind getters/setters or at least abide by some general code guideline around when to make class variables public (e.g. if they are read-only they should always be const -- stuff like that).

danobi: At a high level I'd like to point out this is a purely style thing. I don't recall us ever having any issues with breaking invariants. 

The underscore part was copied from google's C++ coding guidelines. If we find trailing `_` more useful for public/private distinction, I think we should be free to do that. In fact, I'm thinking that it's more useful in general to do that.

I like the idea of getter/setters but, yeah, for RequiredResources it seems like it'd just be a lot of noise.

ajor: I like the idea of only using a trailing underscore for private members and none for public members.

This got me digging into the Google C++ Style Guide, and if I'm understanding correctly, while they do require underscores for all members, they also require that all data members be private:
https://google.github.io/styleguide/cppguide.html#Variable_Names
https://google.github.io/styleguide/cppguide.html#Access_Control

Don't think we need to go as far as requiring no public members though.

danobi:
Ready for another loo

viktormalik (approved):
Nice and clear!

If only the existing code followed these guidelines... :thinking:

jordalgo:
LGTM!

jordalgo (approved):

PR #3020, opened 2024-02-26
BpfProgram: Use correct offset when relocating .text

ajor:
The .text section appears after our main program section:

```
    [Main Program]
    [   .text    ] <- text_offset_
```

When relocating instructions from the main program to point into .text, they must add text_offset_.

When relocating instructions from .text to point to another part of .text, they should not add an offset.

This is needed to allow nested loops in #3003, in which a subprogram calls another subprogram.

### DIFF
--- a/src/bpfprogram.cpp
+++ b/src/bpfprogram.cpp
@@ -101,12 +101,11 @@ void BpfProgram::relocateInsns()
 {
   std::string relsecname = std::string(".rel") + name_;
   // Step 1: relocate our program
-  bool needs_text = relocateSection(
-      relsecname, reinterpret_cast<struct bpf_insn *>(code_.data()));
+  relocateSection(relsecname,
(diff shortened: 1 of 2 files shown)


viktormalik (approved):

src/bpfprogram.cpp
```diff
@@ -174,8 +171,16 @@ bool BpfProgram::relocateSection(const std::string &relsecname, bpf_insn *insns)
       // This is a hack. We shouldn't do this. However, we don't actually have
       // the ELF section table to determine if the relocation actually refers
       // to .text
-      needs_text = true;
-      auto target_insn = (code_.size() + sym->st_value + insn->imm) /
+      if (text_offset_ == 0)
+        text_offset_ = code_.size();
+
+      auto text_relative_offset = text_offset_;
```
jordalgo, line 177: super nits:
- could be const
- could be written as a ternary

jordalgo (approved):

PR #3022, opened 2024-02-27
cmake: Fix nix build

danobi:
Nix does the build in a hermetic environment. Git history is not available during the build. This confuses cmake. Fix by removing the explicit dependency on git history.

This should still work OK, as Version.cmake handles git shellout failures gracefully.

##### Checklist


jordalgo (approved):
(trust stamp)

PR #3023, opened 2024-02-28
CMake: Re-generate version string on every build

ajor:
The `DEPENDS` line removed in #3022 was being used to trigger a re-generation of the version string when the Git commit hash changed. Without it, the version string was only updated when the CMake files changed.

This change ensures that the version string is accurate as developers change branches or add new commits.

It will not trigger a recompilation unless the version string's value changes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -95,8 +95,8 @@ if (KERNEL_HEADERS_DIR)
   target_compile_definitions(runtime PUBLIC KERNEL_HEADERS_DIR="${KERNEL_HEADERS_DIR}")
 endif()
 
-add_custom_command(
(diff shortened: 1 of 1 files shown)


ajor:
The alternative option is that we remove the Git-based versioning and just rely on the numbers in the CMake files. I guess this is what's already happening with Nix builds.

It'd be much simpler CMake code to support and would mean native-builds and Nix-builds followed the same approach.

The disadvantage is that we would lose commit-by-commit versioning. The Git-version for this commit is `v0.20.0-51-ge05a9`, but the CMake-version would be just `v0.20.0`.

danobi (approved):
lgtm!

PR #3027, opened 2024-03-01
print: Print single map values

### LINKED ISSUE #2779: print() single map values
```
BEGIN
{
  @x[1] = 1;
  print(@x[1])
}
```
This script errors out with: `The map passed to print() should not be indexed by a key`.

`printf` is able to print a single map value - `print` should be able to as well.

arthurshau:
Allows indexing a map with key(s) to print single map values while using the print() function.

Resolves #2779.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -472,9 +472,11 @@ void SemanticAnalyser::visit(Call &call)
       auto &expr = *(*call.vargs)[i];
       func_arg_idx_ = i;
 
-      if (expr.is_map && skip_key_validation(call)) {
+      if (expr.is_map) {
         Map &map = static_cast<Map &>(expr);
-        map.skip_key_validation = true;
+        // If the map is indexed, don't skip key validation
+        if (!map.vargs && skip_key_validation(call))
+          map.skip_key_validation = true;
       }
 
       expr.accept(*this);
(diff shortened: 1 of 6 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -930,9 +932,10 @@ void SemanticAnalyser::visit(Call &call)
       if (arg.is_map) {
         Map &map = static_cast<Map &>(arg);
         if (map.vargs != nullptr) {
-          LOG(ERROR, call.loc, err_)
-              << "The map passed to " << call.func << "() should not be "
-              << "indexed by a key";
+          if (call.vargs->size() != 1)
+            LOG(ERROR, call.loc, err_)
+                << "Single-value (i.e. indexed) map print cannot take top "
+                   "or div arguments.";
```
viktormalik, line 938: I don't think that the err message should be top- and div-specific. We may add more cases requiring multiple `print` arguments in future and would have to always update the message:
```suggestion
                << "Single-value (i.e. indexed) map print cannot take additional "
                   "arguments.";
```

viktormalik (commented):
LGTM, thanks!

Just one comment about an error message.

src/ast/passes/semantic_analyser.cpp
arthurshau: Done, thanks!

arthurshau:
Addressed comments

src/ast/passes/semantic_analyser.cpp
viktormalik: It seems that you changed the wrong error message :-)

arthurshau: Whoops...should probably get to bed

viktormalik (approved):
Thanks!

ajor (approved):
Nice!

PR #3031, opened 2024-03-05
Remove more stat/fstat calls 

jordalgo:
Replace with std::filesystem.

https://github.com/bpftrace/bpftrace/issues/1263

There are still a few in utils.cpp but I think
we need to keep those because they access st_ino
and there doesn't seem to be a way to get that
from std::filesystem.

##### Checklist

### DIFF
--- a/src/aot/aot.cpp
+++ b/src/aot/aot.cpp
@@ -244,15 +244,17 @@ int load(BPFtrace &bpftrace, const std::string &in)
     return 1;
   }
 
-  struct stat sb;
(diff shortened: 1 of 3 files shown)


jordalgo:
> Seems like there's a couple more:

Ah good catch, I was only looking for "::stat"

jordalgo:
Let me know if creating a `get_file_size` helper in utils would be good too (there are two instances that would use it)

danobi (approved):
lgtm!

no strong opinion on helper; I think it's fine like this

PR #3033, opened 2024-03-06
Fix security hole checking unpacked kernel headers

jordalgo:
Make sure to check that the unpacked kheaders tar
is owned by root to prevent bpftrace from loading
compromised linux headers.

Borrowed from @brendangregg here: https://github.com/iovisor/bcc/pull/4928

Tested:
```
$ sudo chown 1000 /tmp/kheaders-6.7.4-200.fc39.aarch64
$ sudo ./src/bpftrace test-script.bt 
ERROR: header file ownership expected to be root: /tmp/kheaders-6.7.4-200.fc39.aarch64
```

##### Checklist

### DIFF
--- a/tests/utils.cpp
+++ b/tests/utils.cpp
@@ -358,6 +358,27 @@ TEST(utils, get_pids_for_program)
   ASSERT_EQ(pids.size(), 0);
 }
 
+TEST(utils, file_exists_and_ownedby_root)
+{
+  std::string tmpdir = "/tmp/bpftrace-test-utils-XXXXXX";
+  std::string file1 = "/ownedby-user";
+  std::string file2 = "/no-exists";
+  if (::mkdtemp(tmpdir.data()) == nullptr) {
+    throw std::runtime_error("creating temporary path for tests failed");
+  }
+
+  int fd;
+  fd = open((tmpdir + file1).c_str(), O_CREAT, S_IRUSR);
+  close(fd);
+  ASSERT_GE(fd, 0);
+
+  EXPECT_FALSE(file_exists_and_ownedby_root((tmpdir + file1).c_str()));
+  EXPECT_FALSE(file_exists_and_ownedby_root((tmpdir + file2).c_str()));
+  EXPECT_TRUE(file_exists_and_ownedby_root("/proc/1/maps"));
+
+  EXPECT_GT(std_filesystem::remove_all(tmpdir), 0);
+}
+
 } // namespace utils
 } // namespace test
 } // namespace bpftrace
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -108,6 +108,8 @@ const struct vmlinux_location vmlinux_locs[] = {
   { nullptr, false },
 };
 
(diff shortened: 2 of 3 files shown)


danobi (approved):

jordalgo:
> Possible to add a unit test for file_exists_and_ownedby_root()? Can easily create a non-root owned file. And could probably use /proc/1/.. for a root owned file?

@danobi Good call. Added.

tests/utils.cpp
```diff
@@ -358,6 +358,26 @@ TEST(utils, get_pids_for_program)
   ASSERT_EQ(pids.size(), 0);
 }
 
+TEST(utils, file_exists_and_ownedby_root)
+{
+  std::string tmpdir = "/tmp/bpftrace-test-utils-XXXXXX";
+  std::string file1 = "/ownedby-user";
+  std::string file2 = "/no-exists";
+  if (::mkdtemp(&tmpdir[0]) == nullptr) {
```
danobi, line 366: nit
```suggestion
  if (::mkdtemp(tmpdir.data()) == nullptr) {
```

tests/utils.cpp
```diff
@@ -358,6 +358,26 @@ TEST(utils, get_pids_for_program)
...
+    throw std::runtime_error("creating temporary path for tests failed");
+  }
+
+  int fd;
+  fd = open((tmpdir + file1).c_str(), O_CREAT, S_IRUSR);
```
danobi, line 371: ASSERT_GE(fd, 0) ?

danobi (approved):

brendangregg:
Thanks!

jeromemarchand:
I'm afraid this is not fixed yet. If a user build a valid kheaders tree in /tmp/kheaders-$(uname -r) bpftrace will show an error message, but but proceed to use use the possibly compromised headers none the less:
```
# /usr/share/bpftrace/tools/execsnoop.bt 
ERROR: header file ownership expected to be root: /tmp/kheaders-5.14.0-431.rh29064.el9.x86_64
Attaching 3 probes...
TIME            PID     PPID    ARGS
18:15:38.492036 3116    1934    ls --color=auto -l --color=auto
```

If the user has an empty keaders-$(uname -r) directory  it seems that bpftrace will overwrite it, but not if it isn't empty.
I think the problem is we don't check whether moving tpmdir to shared_path succeeds at the end of unpack_kheaders_tar_xz():
```
  int rc = ::pclose(tar);
  if (rc == 0) {
    tmpdir.move_to(shared_path);
    return shared_path;
  }
```

Or maybe we should exit as soon as wrong ownership is detected.

jordalgo:
@jeromemarchand Good catch! I think maybe there is confusion around `rename` as it seems to be implementation specific if the new file name exists. On my machine, it's doing what you observed in that it doesn't replace the compromised headers path and just re-uses. I'll add some deletion and extra checking (alternatively we can do what you said and just exit if the permissions are not root).

jordalgo:
Second attempt to fix here: https://github.com/bpftrace/bpftrace/pull/3154

PR #3042, opened 2024-03-09
Parse DWARF with liblldb instead of libdw

ttreyer:
Use liblldb instead of libdw to parse Debug Information.
LLDB has support for binaries compiled with advanced DWARF features, such as Split-DWARF and Simple Template Names, which enables using uprobes in those binaries. It will also enable support of C++ classes using inheritance, in a future PR.

The implementation is a drop-in replacement of the previous DWARF parser and uses the same test suite. The public interface of the `Dwarf` class remains the same, except for the `const` qualifier for the functions that had to be dropped, as LLDB is not `const` friendly.

##### Checklist

- [x] ~~Language changes are updated in `man/adoc/bpftrace.adoc`~~ **No language change**


src/CMakeLists.txt
```diff
@@ -172,45 +176,18 @@ if(STATIC_LINKING)
   target_link_libraries(runtime ${LIBBPF_LIBRARIES})
   target_link_libraries(runtime ${LIBBCC_LOADER_LIBRARY_STATIC})
 
+  find_package(LibLzma)
```
viktormalik, line 179: Is this always necessary? Or just for the case when LLDB is found?

src/dwarf_parser.cpp
```diff
@@ -1,250 +1,216 @@
...
+  }
 
-  Dwarf_Die *cudie = nullptr;
-  Dwarf_Addr cubias;
-  while ((cudie = dwfl_nextcu(dwfl, cudie, &cubias)) != nullptr) {
-    if (dwarf_getfuncs(cudie, get_func_die_cb, &func_info, 0) > 0)
-      return func_info.die;
+  for (uint32_t i = 0; i < functions.GetSize(); i++) {
+    auto func = functions.GetContextAtIndex(i).GetFunction();
+    if (func && func.GetName() != nullptr && func.GetName() == function) {
```
viktormalik, line 61: I think that you could omit the middle part, can't you?
```suggestion
    if (func && func.GetName() == function) {
```

src/dwarf_parser.cpp
```diff
@@ -1,250 +1,216 @@
...
-            (inner_enc == DW_ATE_signed_char ||
-             inner_enc == DW_ATE_unsigned_char))
-          result = CreateString(size);
-        else
-          result = CreateArray(size, inner_type);
-        inner_type = result;
+    case lldb::eTypeClassArray: {
+      auto inner_type = type.GetArrayElementType();
+      auto inner_stype = get_stype(inner_type);
+      auto size = type.GetByteSize() / inner_type.GetByteSize();
```
viktormalik, line 181: What about multi-dimensional arrays? IIRC, DWARF doesn't represent them as "arrays of arrays" but as "arrays with multiple dimensions". Will this handle them?

viktormalik (commented):
Nice! This is so much simpler than the libdw approach!

I'd still like to play around with the implementation a bit but won't get to it in the following week so I'm at least posting a couple of comments I have for the code.

Also, I think that we should consider making LLDB a non-optional dependency but that can be handled in a different PR.

src/dwarf_parser.cpp
ttreyer: We can't, `std::string` doesn't check the pointer when comparing against a `const char*`, so the check must be done beforehand to prevent a crash. However, I'm going to simplify it a bit.

src/dwarf_parser.cpp
```diff
@@ -1,250 +1,213 @@
...
       return CreatePointer(CreateNone());
     }
-    case DW_TAG_structure_type:
-    case DW_TAG_union_type: {
-      std::string name = dwarf_diename(&type_die);
-      name = (tag == DW_TAG_structure_type ? "struct " : "union ") + name;
+    case lldb::eTypeClassClass:
+    case lldb::eTypeClassStruct:
+    case lldb::eTypeClassUnion: {
+      std::string name = type.GetDisplayTypeName() ?: "<unknown type>";
```
ajor, line 159: How will this `<unknown type>` appear to a user?

src/dwarf_parser.cpp
```diff
@@ -256,164 +219,36 @@ void Dwarf::resolve_fields(const SizedType &type) const
   std::string type_name = type.GetName();
   if (type_name.find("struct ") == 0)
     type_name = type_name.substr(strlen("struct "));
-  auto type_die = find_type(type_name);
-  if (!type_die)
+  else if (type_name.find("class ") == 0)
+    type_name = type_name.substr(strlen("class "));
+  else if (type_name.find("union ") == 0)
+    type_name = type_name.substr(strlen("union "));
```
ajor, line 225: This code (old and new) might be wrong given that `Foo` and `struct Foo` are different types in C: #682

src/dwarf_parser.cpp
```diff
@@ -1,250 +1,213 @@
...
-  if (dwfl_nextcu(dwarf->dwfl, NULL, &bias) == NULL)
-    return nullptr;
+  debugger_ = lldb::SBDebugger::Create(false);
 
-  return dwarf;
+  lldb::SBError error;
+  target_ = debugger_.CreateTarget(
+      file_path_.c_str(), nullptr, nullptr, false, error);
+  if (!error.Success() || !target_.IsValid()) {
+    throw error;
```
ajor, line 31: I haven't worked out how to trigger this code path, but what happens if we throw `error` while `error.Success()` is true?

src/dwarf_parser.cpp
ttreyer: I've added a test for arrays, see "tests/field_analyser.cpp". It works as I would expect 🙂

src/dwarf_parser.cpp
```diff
@@ -1,250 +1,213 @@
...
-  Dwarf_Die *cudie = nullptr;
-  Dwarf_Addr cubias;
-  while ((cudie = dwfl_nextcu(dwfl, cudie, &cubias)) != nullptr) {
-    if (dwarf_getfuncs(cudie, get_func_die_cb, &func_info, 0) > 0)
-      return func_info.die;
+  auto functions = target_.FindFunctions(function.c_str());
+  for (uint32_t i = 0; i < functions.GetSize(); i++) {
+    auto sym = functions.GetContextAtIndex(i).GetSymbol();
+    if (sym.GetName() && sym.GetName() == function)
+      return functions.GetContextAtIndex(i).GetFunction();
```
ajor, line 58: Reading the docs, it look slike `FindFunctions` can take arguments to filter down the returned functions some more. Would that work instead of our own post-processing here?

src/dwarf_parser.cpp
ajor: Are multi-dimensional arrays getting flattened down to 1-D arrays? It'd be nice if users could access them in the same way as in C/C++ source code, i.e. `$foo->d[0][2]`

ajor (commented):
Very nice! This code so much shorter and more readable and it looks like it'll be far easier to extend in the future too.

src/CMakeLists.txt
ttreyer: It's always necessary for static build, liblzma is required by libbcc.
See https://github.com/bpftrace/bpftrace/actions/runs/8330340247/job/22794747512?pr=3042#step:3:623

src/dwarf_parser.h
```diff
@@ -96,17 +84,22 @@ class Dwarf {
...
   }
 
-  void resolve_fields(const SizedType &type __attribute__((unused))) const
+  void resolve_fields(const SizedType &type __attribute__((unused)))
+  {
+  }
+
+private:
+  Dwarf(BPFtrace *bpftrace __attribute__((unused)),
+        std::string file_path_ __attribute__((unused)))
```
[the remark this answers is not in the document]
ttreyer: Yes, otherwise the compiler will add a public default constructor.

ttreyer:
@ajor 

>  Are multi-dimensional arrays getting flattened down to 1-D arrays? It'd be nice if users could access them in the same way as in C/C++ source code, i.e. `$foo->d[0][2]`

Printing the `SizedType` of the test array does give me `int32[4][4]`, so the `SizedType` properly reconstruct the array.

src/dwarf_parser.cpp
```diff
@@ -1,250 +1,213 @@
...
+std::unique_ptr<Dwarf> Dwarf::GetFromBinary(BPFtrace *bpftrace,
+                                            std::string file_path)
 {
-  auto *func_info = static_cast<struct FuncInfo *>(arg);
-  if (dwarf_hasattr(func_die, DW_AT_name) &&
-      dwarf_diename(func_die) == func_info->name) {
-    func_info->die = *func_die;
-    return DWARF_CB_ABORT;
+  try {
+    return std::unique_ptr<Dwarf>(new Dwarf(bpftrace, std::move(file_path)));
```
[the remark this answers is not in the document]
ttreyer: Unfortunately, only Clang handles the friendship for `std::make_unique` and GCC doesn't, see the [CI run](https://github.com/bpftrace/bpftrace/actions/runs/8330811720/job/22796307638?pr=3042#step:6:1501).
I'll revert to the old way and double check we don't leak a `Dwarf` object in case of exception.

src/dwarf_parser.cpp
ttreyer: Printing the SizedType of the test array does give me int32[4][4], so the SizedType properly represents the array.

ajor: Ah cool! I think I'd misread the new test case you added for it.

It would be good to check the full details of the array in the test case too, to ensure it stays working this way in the future, e.g. `GetNumElements()`, `GetElementTy()` for both the outer and inner arrays

src/dwarf_parser.cpp
ttreyer: It will show the following error message to the user:
```
ERROR: Failed to parse DebugInfo: nullptr
```
Not optimal, but I'm not sure in what case we could have no error, nor target. I'll dig a little bit more into it...

src/dwarf_parser.cpp
ttreyer: There are no flag that narrow down the search to just the mangled name.

src/dwarf_parser.cpp
ttreyer: So we should never reach the second condition in the `if`. This is a quirk of the API that doesn't return a `std::expect` or Rust's `std::Result` kind of object.
Even if we reach this weird edge case, we would get a strange error message, but no crash. Would you be happy to keep it this way?

tests/runtime/dwarf
```diff
@@ -6,13 +6,13 @@ TIMEOUT 5
...
 RUN {{BPFTRACE}} -lv 'uprobe:./testprogs/uprobe_test:uprobeFunction1'
-EXPECT int* n
+EXPECT int * n
 REQUIRES_FEATURE dwarf
 TIMEOUT 5
 
 NAME list uprobe args - struct pointer type
 RUN {{BPFTRACE}} -lv 'uprobe:./testprogs/uprobe_test:uprobeFunction2'
-EXPECT struct Foo* foo1
+EXPECT Foo * foo1
```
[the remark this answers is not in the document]
ttreyer: This would be possible 🙂. We can lookup the language of a function and use it to adjust the naming scheme. Let's do it another time however.

src/dwarf_parser.cpp
ajor: Sounds ok to me then 👍

src/dwarf_parser.h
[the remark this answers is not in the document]
ajor: How about `Dwarf() = delete` instead? It's clearer about its intentions and won't have me asking the same question when I've forgotten this in a few years' time 😅

ajor:
> Following up about what we discussed during office hours: does lldb have "escape hatches" for non-C-family languages like golang or rust?

Good point - we should add some test programs for other languages to confirm what we can do already. I've opened an issue for it: #3075. Then we can confirm that they still work after this change.

src/dwarf_parser.cpp
ttreyer: I need to add tests for anonymous struct/union/enum (the case you are referring to) and anonymous fields.
For anonymous types, we could print the declaration instead of `<unknown type>`? Like `enum { A, B } foo`.
For anonymous fields, we might need to flatten the type to reproduce the behaviour of a regular compiler.

ttreyer: Following our chat, `<unknown type>` (now renamed to `<anonymous type>`) would appear when listing the argument of a function (anonymous `enum` or `union`, but we shouldn't encounter this often...), and when printing a `struct` whose fields have anonymous type.
The goal is to revisit this and to correctly print the representation of anonymous types. I'll open an issue for that.

src/dwarf_parser.cpp
```diff
@@ -1,419 +1,276 @@
...
-      return func_info.die;
+  std::string type_name = type.GetDisplayTypeName() ?: "<anonymous type>";
+
+  /* LLDB removes the struct/union/enum prefix from the type name, but C
+   * requires it! So we first need to peel off pointers/references to reach the
+   * inner type. From the inner type, we can then add the proper prefix back on.
+   */
+  while (true) {
+    switch (type.GetTypeClass()) {
+      case lldb::eTypeClassArray:
```
ajor, line 63: So I understand, is this case for handling a type name like `struct Foo[]` instead of calling it `Foo[]`? Do we have any tests for this?

ajor (approved):
Looks good to me!

For the areas you've flagged as needing work (e.g. anonymous types), would you be able to add comments to the code or tests referencing the relevant GitHub issues please?

We need to invest in more thorough testing of these type parsing code paths in general, but we've been lacking in that area for a while so don't think we need to block on that.

src/dwarf_parser.cpp
ttreyer: We don't have a test for the array case. In C, arrays passed as a parameter decay into a pointer.
In C++, it is possible to preserve the array, so I will add a test for it in my next PR.

tests/clang_parser.cpp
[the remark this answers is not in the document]
ttreyer: It looks the same and does the same steps, but add a call to `ClangParser::parse`. I'm not familiar enough with the test and the ClangParser to understand what it does.

ajor (approved):
Will be merging this soon if there are no objections

jordalgo (approved):
LGTM. Nice work @ttreyer!

tests/clang_parser.cpp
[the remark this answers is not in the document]
viktormalik: There are different code paths tested by those tests. While `field_analyser_btf` tests check for type resolution done in `FieldAnalyser` (which traverses the AST), `clang_parser_btf` check for type resolution done in `ClangParser` (which calls Clang).

The latter is triggered in cases when the script uses custom includes or user-defined types. You can see that from the fact that tests in `field_analyser_btf` use typecasts while tests in `clang_parser_btf` use user-defined types.

But, in the end of the day, both code paths use the same methods from the `BTF` class.

viktormalik (approved):

PR #3045, opened 2024-03-11
use 64bit alignment for map counter atomic add

fvdlfvdl:
For an atomic inc of a map counter (ringbuf loss counter), generate IR with 64bit alignment. This is more correct, and will avoid problems with upcoming LLVM versions, as they will emit a function call for a potentially unaligned atomicrmw. This will lead to an error like this:

# bpftrace -e 'BEGIN { print("hello")}'
error: <unknown>:0:0: in function BEGIN i64 (ptr): t15: i64 = GlobalAddress<ptr @__atomic_compare_exchange> 0 too many arguments


danobi (commented):
lgtm, thanks for the PR! but please add a changelog entry in https://github.com/bpftrace/bpftrace/blob/master/CHANGELOG.md

fvdlfvdl:
Ack - changelog update added to the commit.

PR #3046, opened 2024-03-11
Allow multiple arguments for delete()

### LINKED ISSUE #947: Variadic arguments for delete
```d
END
{
  delete(@start);
  delete(@avg);
  delete(@min);
  delete(@max);
  delete(@count);
  delete(@hist);
}
```

would become

```d
END
{
  delete(@start, @avg, @min, @max, @count, @hist);
}
```

arthurshau:
Implements passing of multiple arguments to delete().

E.g. you can now do

```
@x[1] = 1;
@x[2] = 2;
@c = 5;
delete(@x[1], @x[2], @c);
```
instead of having to call delete() separately for each individual map element.

Closes  #947.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -555,17 +555,18 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateLifetimeEnd(key);
     expr_ = nullptr;
   } else if (call.func == "delete") {
-    auto &arg = *call.vargs->at(0);
-    auto &map = static_cast<Map &>(arg);
-    auto [key, scoped_key_deleter] = getMapKey(map);
-    if (!is_bpf_map_clearable(map_types_[map.ident])) {
-      // store zero instead of calling bpf_map_delete_elem()
-      AllocaInst *val = b_.CreateAllocaBPF(map.type, map.ident + "_zero");
-      b_.CreateStore(Constant::getNullValue(b_.GetType(map.type)), val);
-      b_.CreateMapUpdateElem(ctx_, map, key, val, call.loc);
-      b_.CreateLifetimeEnd(val);
-    } else {
-      b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
+    for (const auto &arg : *call.vargs) {
+      auto &map = static_cast<Map &>(*arg);
+      auto [key, scoped_key_deleter] = getMapKey(map);
+      if (!is_bpf_map_clearable(map_types_[map.ident])) {
+        // store zero instead of calling bpf_map_delete_elem()
+        AllocaInst *val = b_.CreateAllocaBPF(map.type, map.ident + "_zero");
+        b_.CreateStore(Constant::getNullValue(b_.GetType(map.type)), val);
+        b_.CreateMapUpdateElem(ctx_, map, key, val, call.loc);
+        b_.CreateLifetimeEnd(val);
+      } else {
+        b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
+      }
     }
     expr_ = nullptr;
   } else if (call.func == "str") {
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -606,12 +606,13 @@ void SemanticAnalyser::visit(Call &call)
     call.type = CreateStats(true);
   } else if (call.func == "delete") {
     check_assignment(call, false, false, false);
-    if (check_nargs(call, 1)) {
-      auto &arg = *call.vargs->at(0);
-      if (!arg.is_map)
-        LOG(ERROR, call.loc, err_) << "delete() expects a map to be provided";
+    if (check_varargs(call, 1, std::numeric_limits<size_t>::max())) {
+      for (const auto &arg : *call.vargs) {
+        if (!arg->is_map)
+          LOG(ERROR, arg->loc, err_)
+              << "delete() only expects maps to be provided";
+      }
     }
-
     call.type = CreateNone();
   } else if (call.func == "str") {
     if (check_varargs(call, 1, 2)) {
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -220,6 +220,12 @@ PROG BEGIN { @ = 1; @a[1] = 1; delete(@); delete(@a[1]); printf("ok\n"); exit();
 EXPECT ok
 TIMEOUT 1
 
+NAME delete multiple-map
+PROG BEGIN { @ = 1; @a[1] = 1; delete(@, @a[1]); exit(); }
+EXPECT_NONE @: 1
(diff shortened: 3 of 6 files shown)


tests/runtime/basic
```diff
@@ -220,11 +220,17 @@ PROG BEGIN { @ = 1; @a[1] = 1; delete(@); delete(@a[1]); printf("ok\n"); exit();
 EXPECT ok
 TIMEOUT 1
 
+NAME delete multiple-map
+PROG BEGIN { @ = 1; @a[1] = 1; delete(@, @a[1]); printf("ok\n"); exit(); }
+EXPECT ok
```
jordalgo, line 225: Without the deletion, this would also print the map right?

arthurshau: Yes I believe that's correct.

tests/runtime/basic
```diff
@@ -220,11 +220,17 @@ PROG BEGIN { @ = 1; @a[1] = 1; delete(@); delete(@a[1]); printf("ok\n"); exit();
...
+TIMEOUT 1
+
 NAME delete count-map
 PROG BEGIN { @ = count(); @a[1] = count(); delete(@); delete(@a[1]); exit(); }
 EXPECT @: 0
 TIMEOUT 1
 
+
```
jordalgo, line 233: nit: rm extra space

src/ast/passes/semantic_analyser.cpp
```diff
@@ -606,12 +606,13 @@ void SemanticAnalyser::visit(Call &call)
     call.type = CreateStats(true);
   } else if (call.func == "delete") {
     check_assignment(call, false, false, false);
-    if (check_nargs(call, 1)) {
-      auto &arg = *call.vargs->at(0);
-      if (!arg.is_map)
-        LOG(ERROR, call.loc, err_) << "delete() expects a map to be provided";
+    if (check_varargs(call, 1, 128)) {
```
jordalgo, line 609: maybe make 128 a file const e.g. `MAX_VARGS`

src/ast/passes/semantic_analyser.cpp
```diff
@@ -606,12 +606,13 @@ void SemanticAnalyser::visit(Call &call)
...
+      for (auto iter = call.vargs->begin(); iter != call.vargs->end(); iter++) {
```
jordalgo, line 610: Can this be a simpler `for (const auto varg : call.vargs)` ?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -551,17 +551,19 @@
...
-      AllocaInst *val = b_.CreateAllocaBPF(map.type, map.ident + "_zero");
-      b_.CreateStore(Constant::getNullValue(b_.GetType(map.type)), val);
-      b_.CreateMapUpdateElem(ctx_, map, key, val, call.loc);
-      b_.CreateLifetimeEnd(val);
-    } else {
-      b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
+    for (size_t i = 0; i < call.vargs->size(); i++) {
+      auto &arg = *call.vargs->at(i);
+      auto &map = static_cast<Map &>(arg);
+      auto [key, scoped_key_deleter] = getMapKey(map);
```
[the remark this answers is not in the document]
jordalgo: We should have a lint rule that ignore this.

jordalgo:
Few small nits but otherwise LGTM

tests/runtime/basic
ajor: The problem with the `EXPECT ok` approach is that the test will still pass even if the `delete` call is omitted (or doesn't work correctly).

The existing tests in this file (and in much of the rest of the test suite) are crappy - no fault of your own, @arthurshau! It's a big problem we're going to have to deal with gradually.

I think to properly check that our code is doing what it's meant to, we'd have to do something like:
```
EXPECT_NONE @: 1
EXPECT_NONE @a[1]: 1
```

(don't bother fixing all the old tests right now, it'd just be good to have new tests be better examples to copy from in the future)

arthurshau:
Addressed comments, thanks!

jordalgo:
@arthurshau FYI: you can re-run failed jobs from the CLI page.

jordalgo (approved):
LGTM!

src/ast/passes/semantic_analyser.h
```diff
@@ -14,6 +14,8 @@
 namespace bpftrace {
 namespace ast {
 
+#define MAX_VARGS 128
```
ajor, line 17: Quoting the Cpp Core Guidelines: "Don’t use macros for constants"
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions

But also, is there any reason we need to define a maximum number of var args? Could `check_varargs` just default to having its third parameter be `std::numeric_limits<size_t>::max()`?

tests/semantic_analyser.cpp
```diff
@@ -538,6 +538,10 @@ TEST(semantic_analyser, call_stats)
 TEST(semantic_analyser, call_delete)
 {
   test("kprobe:f { @x = 1; delete(@x); }", 0);
+  test("kprobe:f { @x = 1; @y = 2; delete(@x, @y); }", 0);
+  test("kprobe:f { @x = 1; @y[5] = 5; delete(@x, @y[5]); }", 0);
+  test("kprobe:f { @x = 1; @y[5] = 5; delete(@x, @y); }", 1);
+  test("kprobe:f { @x = 1; $y = 2; delete(@x, $y); }", 1);
```
ajor, line 544: Another instance of less-than-ideal old tests being bad examples - sorry!

I'm trying to make sure all new semantic analyser tests follow the new pattern of checking the error messages instead of the return code, e.g.: https://github.com/bpftrace/bpftrace/blob/b1dd3706e81d635a342dd37a0530467ee2eb8a9f/tests/semantic_analyser.cpp#L286-L290

Checking the error would be useful for this test specifically, because it currently gives this message:
```
stdin:1:28-42: ERROR: delete() only expects maps to be provided
kprobe:f { @x = 1; $y = 2; delete(@x, $y); }
                           ~~~~~~~~~~~~~~
```

But I think that when there are lots of arguments to `delete()` it'd be more useful to highlight the particular one which is causing the problem:
```
stdin:1:39-41: ERROR: delete() only expects maps to be provided
kprobe:f { @x = 1; $y = 2; delete(@x, $y); }                                                   
                                      ~~
```

src/ast/passes/semantic_analyser.h
arthurshau: Sure, I can do that. I believe I just copied the 128 from the check_varargs(...) call for "printf", "system", "cat", etc. since I wasn't sure whether or not we had a reason to limit the number of arguments.

arthurshau:
Addressed comments

ajor (approved):

ajor:
Nice, thank you!

PR #3047, opened 2024-03-12
Better error message for args in mixed probes

viktormalik:
It is not possible to use the `args` builtin in probes with mixed probe types (e.g. tracepoints and kfuncs). When trying that, bpftrace correctly fails but the error message is not very clear:

    # bpftrace -e 'tracepoint:syscalls:sys_enter_read,kfunc:vfs_read { print(args.count); }'
    stdin:1:53-63: ERROR: The args builtin can only be used with tracepoint/kfunc/uprobeprobes (invalid used here)
    tracepoint:syscalls:sys_enter_read,kfunc:vfs_read { print(args.count); }
                                                        ~~~~~~~~~~

Improve the error message:

    # bpftrace -e 'tracepoint:syscalls:sys_enter_read,kfunc:vfs_read { print(args.count); }'
    stdin:1:53-63: ERROR: The args builtin must be used in probes with a single probe type
    tracepoint:syscalls:sys_enter_read,kfunc:vfs_read { print(args.count); }
                                                        ~~~~~~~~~~

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -423,8 +423,13 @@ void SemanticAnalyser::visit(Builtin &builtin)
 
     ProbeType type = single_provider_type(probe);
 
-    if (type == ProbeType::kfunc || type == ProbeType::kretfunc ||
-        type == ProbeType::uprobe) {
+    if (type == ProbeType::invalid) {
+      LOG(ERROR, builtin.loc, err_)
+          << "The args builtin can only be used within the context of a single "
+             "probe type, e.g. \"probe1 {args}\" is valid while "
+             "\"probe1,probe2 {args}\" is not.";
+    } else if (type == ProbeType::kfunc || type == ProbeType::kretfunc ||
+               type == ProbeType::uprobe) {
       auto type_name = probe->args_typename();
       builtin.type = CreateRecord(type_name,
                                   bpftrace_.structs.Lookup(type_name));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to
 - Add for-each loops for iterating over map elements
   - [#3003](https://github.com/bpftrace/bpftrace/pull/3003)
 #### Changed
+- Better error message for args in mixed probes
+  - [#3047](https://github.com/bpftrace/bpftrace/pull/3047)
 #### Deprecated
 #### Removed
 #### Fixed
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -3596,6 +3596,15 @@ BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }
              false);
 }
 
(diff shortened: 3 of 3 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -401,8 +401,11 @@ void SemanticAnalyser::visit(Builtin &builtin)
 
     ProbeType type = single_provider_type();
 
-    if (type == ProbeType::kfunc || type == ProbeType::kretfunc ||
-        type == ProbeType::uprobe) {
+    if (type == ProbeType::invalid) {
+      LOG(ERROR, builtin.loc, err_) << "The args builtin must be used in "
+                                       "probes with a single probe type";
```
jordalgo, line 406: How about this instead:

```suggestion
      LOG(ERROR, builtin.loc, err_) << "The args builtin can only be used in "
                                       "actions associated with a single probe type. To fix this simply create multiple action blocks; one for each probe.";
```

viktormalik: We don't use the word "action" nearly at all (except for the manual) and use "probe" everywhere instead. That being said, I'm not strictly against changing this (although I have a slight preference for "probe") but I'd argue that we should stay consistent for now.

ajor:
Could we get a test for this error message? A la: https://github.com/bpftrace/bpftrace/blob/e8d088b657221fd87b773f10a0617a5f7ed02057/tests/semantic_analyser.cpp#L286-L290

Errors messages are one of the main outputs of SemanticAnalyser, so I'd like to eventually have all of them tested.

src/ast/passes/semantic_analyser.cpp
viktormalik: I don't want to get into lengthy discussions on error messages but I'd prefer keeping them compact. What about just mentioning the actual probe types? That should give enough hint:
`The args builtin can only be used within the context of a single probe type (have "kfunc", "tracepoint").`

jordalgo: > I don't want to get into lengthy discussions on error messages

I'm trying not to honestly, but it's difficult 😄  TBH, I don't understand the `(have "kfunc", "tracepoint")` part, what does this mean? Also, why not mention the solution to this issue if it's easy enough to display to users? How about this as a more compact version of what I suggested above:
`The args builtin can only be used within the context of a single probe type e.g. "probe1 {}" is valid while "probe1,probe2 {}" is not.`

viktormalik:
> Could we get a test for this error message? A la:
> 
> https://github.com/bpftrace/bpftrace/blob/e8d088b657221fd87b773f10a0617a5f7ed02057/tests/semantic_analyser.cpp#L286-L290
> 
> Errors messages are one of the main outputs of SemanticAnalyser, so I'd like to eventually have all of them tested.

Done. I had to add it into `semantic_analyser_btf` tests to be able to mix kfunc and tracepoint and get the expected error message.

jordalgo (approved):
The most rigorously reviewed change to an error message ever!

danobi (approved):

PR #3056, opened 2024-03-14
Change field access in oneliner documentation

MaitreRenard:
```
stdin:1:39-49: ERROR: Can not access field 'bytes' on type 'struct _tracepoint_block_block_rq_issue'. Try dereferencing it first, or using '->'
tracepoint:block:block_rq_issue { @ = hist(args.bytes); }
```
I got this error a couple of times while checking out the documentation.
Indeed it seems that a pointer to the structure is passed on to the function `->` works well for the few examples there were errors on. 

##### Checklist


ajor:
I'm assuming that you're running into this issue on an older version of bpftrace (0.18 or below). `args` was changed in v0.19 to no longer require dereferencing: #2578

Older versions of the docs that correspond to your bpftrace version should be accurate for you, e.g. https://github.com/bpftrace/bpftrace/blob/0.18.1_release/docs/tutorial_one_liners.md

Side note: a number of people have come across this issue - we need to make it easier to access older versions of the docs.

PR #3062, opened 2024-03-15
Check error messages in subprog tests

ajor:
A few of them appear to have their locations offset, e.g.:
```
  fn f(): int64 { return func; }
                          ~~~~
```

This will be fixed separately.


ajor:
Opened an issue for the incorrect results: #3063

jordalgo (approved):
Since you already made an issue, you could mark the bad ones with a comment that includes that issue #

jordalgo (approved):
Sweeeettt

PR #3064, opened 2024-03-15
Parser Tests: Check all error messages on parse failure

ajor:
Most of the errors look good. Some have some minor issues but I'm just locking in the existing behaviour with this PR.

### DIFF
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -13,7 +13,7 @@ using Printer = ast::Printer;
 
 void test_parse_failure(BPFtrace &bpftrace,
                         std::string_view input,
-                        std::string_view expected_error = {})
+                        std::string_view expected_error)
 {
   std::stringstream out;
   Driver driver(bpftrace, out);
@@ -22,12 +22,11 @@ void test_parse_failure(BPFtrace &bpftrace,
   if (expected_error.data()) {
     if (!expected_error.empty() && expected_error[0] == '\n')
       expected_error.remove_prefix(1); // Remove initial '\n'
-    EXPECT_EQ(out.str(), expected_error);
+    EXPECT_EQ(expected_error, out.str());
   }
 }
 
-void test_parse_failure(std::string_view input,
-                        std::string_view expected_error = {})
+void test_parse_failure(std::string_view input, std::string_view expected_error)
 {
   BPFtrace bpftrace;
   test_parse_failure(bpftrace, input, expected_error);
@@ -36,7 +35,7 @@ void test_parse_failure(std::string_view input,
 void test(BPFtrace &bpftrace, std::string_view input, std::string_view expected)
 {
   Driver driver(bpftrace);
-  EXPECT_EQ(driver.parse_str(input), 0);
+  ASSERT_EQ(driver.parse_str(input), 0);
 
   if (expected[0] == '\n')
     expected.remove_prefix(1); // Remove initial '\n'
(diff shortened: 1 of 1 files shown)


tests/parser.cpp
```diff
@@ -22,12 +22,12 @@ void test_parse_failure(BPFtrace &bpftrace,
   if (expected_error.data()) {
     if (!expected_error.empty() && expected_error[0] == '\n')
       expected_error.remove_prefix(1); // Remove initial '\n'
-    EXPECT_EQ(out.str(), expected_error);
+    EXPECT_EQ(expected_error, out.str());
```
ajor, line 25: Flipped the expected and actual results to make GTest give nicer error messages

tests/parser.cpp
```diff
@@ -36,7 +36,7 @@ void test_parse_failure(std::string_view input,
 void test(BPFtrace &bpftrace, std::string_view input, std::string_view expected)
 {
   Driver driver(bpftrace);
-  EXPECT_EQ(driver.parse_str(input), 0);
+  ASSERT_EQ(driver.parse_str(input), 0);
```
ajor, line 39: If this fails then `driver.root` will be a null pointer and `Printer` will crash below trying to display it

tests/parser.cpp
```diff
@@ -277,11 +281,36 @@ TEST(Parser, positional_param_attachpoint)
   int: 1
 )PROG");
 
-  test_parse_failure(bpftrace, R"PROG(uprobe:$1a" { 1 })PROG");
-  test_parse_failure(bpftrace, R"PROG(uprobe:$a" { 1 })PROG");
-  test_parse_failure(bpftrace, R"PROG(uprobe:$-1" { 1 })PROG");
+  // TODO what are these tests meant to be checking for?
```
ajor, line 284: These tests confused me. Can anyone help to explain why there are trailing, unmatched quotes (`"`) in each of them?

Do they serve a purpose or could they be a mistake?

danobi: I git-blamed it and it looks like I added it. Probably a typo. Good that you're adding the expected error messages :)

ajor: Cool, removed the trailing quotes!

tests/parser.cpp
```diff
@@ -2131,17 +2422,54 @@ uprobe:asdf:Stream {} tracepoint:only_one_arg {}
...
+config = { BPFTRACE_STACK_MODE=perf } i:s:1 { BPFTRACE_MAX_PROBES=2; exit(); }
+                                              ~~~~~~~~~~~~~~~~~~~~
+)");
+  test_parse_failure("config { BPFTRACE_STACK_MODE=perf } i:s:1 { exit(); }",
+                     R"(
+stdin:1:8-9: ERROR: syntax error, unexpected {, expecting =
+config { BPFTRACE_STACK_MODE=perf } i:s:1 { exit(); }
+       ~
+)");
+  test_parse_failure("BPFTRACE_STACK_MODE=perf; i:s:1 { exit(); }", R"(
```
jordalgo, line 2468: nit: I know the rest of the file doesn't adhere to this but I like a little breathing room (line break) between the `test_parse_failure`s 🌬️

jordalgo:
@ajor is an error message extraordinaire

jordalgo (approved):

tests/parser.cpp
ajor: Good point, definitely easier to read this way

PR #3065, opened 2024-03-15
Semantic Analyser Tests: Don't explicitly check for "0" return code

ajor:
This is a first step towards having the tests not care about the exact return code value and instead only checking for success/fail.

The next step is to convert the tests with non-zero return codes to use test_error() and check their error messages.

### DIFF
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -95,19 +95,45 @@ void test(BPFtrace &bpftrace,
   }
 }
 
+void test(BPFtrace &bpftrace, std::string_view input, bool safe_mode = true)
+{
+  Driver driver(bpftrace);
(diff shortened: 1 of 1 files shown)


jordalgo (approved):
Nice cleanup.

All the overloaded `test` functions make my head hurt but I assume (with your comments) that a lot of these will go away soon.

ajor:
Yeah the overloads have gotten slightly out of hand :(

The versions that accept `int expected_result` will be going away, but we'll be left with overloads of `test()` and `test_error()` accepting each of `BPFtrace&`, `Driver&` and `MockBPFfeature&`, unless we make some more changes.

Just had a quick look through them and from what I can tell, all the `test(Driver&, ...)` overloads would be better swapped out for `test_ast()`, so there's another cleanup I might make soon.

PR #3068, opened 2024-03-16
Add user function call support

lenticularis39:
Function defined by the user can now be called in the same ways as built-in functions, for example:

fn add($a: int64, $b: int64): int64 {
  return $a + $b;
}

BEGIN {
  print(add(5, 9));
}

If a function of the same name as an existing built-in is defined, it overrides the built-in and a warning is printed.

Implementation-wise, the functionality depends on existing relocations against subprogs implemented in commit d345d06 ("Implement relocations against .text for subprogs").

Functions are generated as separate LLVM IR functions and translated into BPF subprogs via the LLVM bpf backend. Relocations make sure the resulting program can be loaded.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -472,7 +472,9 @@ void CodegenLLVM::visit(Builtin &builtin)
 
 void CodegenLLVM::visit(Call &call)
 {
-  if (call.func == "count") {
+  if (subprogs_.find(call.func) != subprogs_.end()) {
+    createSubprogCall(call);
+  } else if (call.func == "count") {
     Map &map = *call.map;
     auto [key, scoped_key_deleter] = getMapKey(map);
     b_.CreateMapElemAdd(ctx_, map, key, b_.getInt64(1), call.loc);
@@ -2553,7 +2555,20 @@ void CodegenLLVM::visit(Subprog &subprog)
                                               0);
 
   Function *func = Function::Create(
-      func_type, Function::InternalLinkage, subprog.name(), module_.get());
+      func_type, Function::ExternalLinkage, subprog.name(), module_.get());
+
+  // Mark function as NoInline. This has several reasons:
+  // - inlining takes up BPF instructions
+  // - LLVM sometimes inlines recursive calls into loops, which can be then
+  //   rejected by the BPF verifier for code where a recursive call would
+  //   not
+  AttributeList attrs;
+  attrs = attrs.addFnAttribute(b_.getContext(), Attribute::NoInline);
+  func->setAttributes(attrs);
+
+  // Create BTF information needed for program loading
+  debug_.createFunctionDebugInfo(*func);
+
   BasicBlock *entry = BasicBlock::Create(module_->getContext(), "entry", func);
   b_.SetInsertPoint(entry);
 
@@ -2574,6 +2589,11 @@ void CodegenLLVM::visit(Subprog &subprog)
   }
   if (subprog.return_type.IsVoidTy())
     createRet();
+
+  // Replace dummy function
+  if (auto dummy_func = module_->getFunction(subprog.name() + ":dummy")) {
+    dummy_func->replaceAllUsesWith(func);
+  }
 
   FunctionPassManager fpm;
   FunctionAnalysisManager fam;
@@ -2745,9 +2765,16 @@ void CodegenLLVM::visit(Probe &probe)
 
 void CodegenLLVM::visit(Program &program)
 {
-  if (program.functions)
+  if (program.functions) {
+    // first collect subprog names into set to allow them to override builtins
+    std::transform(program.functions->cbegin(),
+                   program.functions->cend(),
+                   std::inserter(subprogs_, subprogs_.begin()),
+                   [](Subprog *subprog) { return subprog->name(); });
+
     for (Subprog *subprog : *program.functions)
       auto scoped_del = accept(subprog);
+  }
   for (Probe *probe : *program.probes)
     auto scoped_del = accept(probe);
 }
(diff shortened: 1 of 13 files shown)


lenticularis39:
Second PR following recently merged #2624 to implement user-defined functions (#2516).

lenticularis39:
Fixed build on systems with older glibc that do not have BPF relocations in `elf.h`, like those used in CI.

ajor (commented):
Just had a quick play around so far - this is looking really nice!

I'm not necessarily expecting you to sort out all the stuff I'm writing here in this PR (but feel free if you want!) - it's for my future reference as much as anything :) 

#### Int Typing

There seem to be a few minor issues with the integer typing, but nothing major:
```
$ sudo ./bpftrace -e 'fn bar(): int64 { return @x } BEGIN { @x = 1; }'
stdin:1:20-29: ERROR: Function bar is of type int64, cannot return none
fn bar(): int64 { return @x } BEGIN { @x = 1; }
                   ~~~~~~~~~
```

And also a lot of confusion between `int64` and `uint64` - I haven't been able to work out when to use each. I imagine this isn't an issue specific to your work, but is just made more obvious by it.
```
$ sudo ./bpftrace -e 'fn foo(): uint64 { return 1; }'
stdin:1:21-29: ERROR: Function foo is of type unsigned int64, cannot return int64
fn foo(): uint64 { return 1; }
                    ~~~~~~~~
```

This one made me laugh:
```
$ sudo ./bpftrace -e 'fn foo(): uint64 { return @x + 1; } BEGIN { @x = 1; }'
stdin:1:21-34: ERROR: Function foo is of type unsigned int64, cannot return int64
fn foo(): uint64 { return @x + 1; } BEGIN { @x = 1; }
                    ~~~~~~~~~~~~~
```

Ok, it wants the function to return `int64` instead? Let's try that:
```
$ sudo ./bpftrace -e 'fn foo(): int64 { return @x + 1; } BEGIN { @x = 1; }'
stdin:1:20-33: ERROR: Function foo is of type int64, cannot return unsigned int64
fn foo(): int64 { return @x + 1; } BEGIN { @x = 1; }
                   ~~~~~~~~~~~~~
```

It just doesn't like me 😂

#### Other Typing

I've struggled to make non-integer types work:
```
struct Foo {
  int a;
};

fn myfunc(): struct Foo {
  return *(struct Foo*)0;
}
```
```
test.bt:6:3-26: ERROR: Function myfunc is of type struct Foo, cannot return struct Foo
  return *(struct Foo*)0;
  ~~~~~~~~~~~~~~~~~~~~~~~
```

```
$ sudo ./bpftrace -e 'fn myfunc(): str_t[15] { return "hello"; }'
stdin:1:27-41: ERROR: Function myfunc is of type string[15], cannot return string[6]
fn myfunc(): str_t[15] { return "hello"; }
                          ~~~~~~~~~~~~~~

```

Let's use the size it suggests:
```
$ sudo ./bpftrace -e 'fn myfunc(): str_t[6] { return "hello"; }'                                                                  
Segmentation fault
```


#### Verifier errors
I'm not sure what the right solution is here, but I can trigger verifier errors with recursive function calls:
```
fn foo(): uint64 {
  return foo() + 1;
}

BEGIN {
  print(foo());
}
```

I don't think we should ban recursive calls entirely as they can work:
```
fn fact($n: uint64): uint64 {
  if ($n == 0) { return (uint64)1; }
  return $n * fact($n - 1);
}

BEGIN {
  print(fact(5));
  exit();
}
```

lenticularis39:
There are still some limitations to the implementation, the biggest of which is that bpftrace functions are directly converted into BPF subprograms. That means you are not able to pass more than four 64-bit values to a function, since arguments are passed in registers, and also you can't have a return value bigger than 8 bytes. Currently, passing values through a map can be used as a workaround for that. Given that, declaring a function returning a string should not segfault; that is an unrelated bug.

The annoying typing of integers is the result of using stricter type comparisons for functions; bpftrace generally only checks the base type (integer, string etc.), not the signedness nor the size of the type. I was not sure how to approach that; opting for less strict type control seemed possibly dangerous to me so I chose to make the type control strict for the time being.

About the verifier errors: that's something we unfortunately have to live with. The verifier is expected to get better at proving the halting of recursive calls but as the halting problem is unsolvable, there will always be recursive calls that fail at the verifier. It makes little sense to replicate the complex halting checks in bpftrace.

ajor:
> About the verifier errors: that's something we unfortunately have to live with. The verifier is expected to get better at proving the halting of recursive calls but as the halting problem is unsolvable, there will always be recursive calls that fail at the verifier. It makes little sense to replicate the complex halting checks in bpftrace.

We shouldn't allow verifier errors though to the user of bpftrace though. AFAIK it's not possible to trigger one except when using `while` loops (which I'd like to deprecate eventually). In other parts of the language we generally restrict functionality so that the verifier will always be happy. I'm not so keen to do that here because there is useful functionality that'd get blocked as collateral damage.

Maybe something similar to what we do with `HelperVerifierError` might be a good solution - parse the verifier log and if it matches the pattern we expect, provide a nice message to the user explaining what's wrong and how to fix it.

jordalgo:
> Maybe something similar to what we do with HelperVerifierError might be a good solution - parse the verifier log and if it matches the pattern we expect, provide a nice message to the user explaining what's wrong and how to fix it.

This sounds good but also possibly very difficult in trying to anticipate every verifier error as we make the bpftrace language more expressive. I imagine that we wouldn't do the same for raw c code that we import and interop with.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2553,7 +2555,20 @@ void CodegenLLVM::visit(Subprog &subprog)
...
-      func_type, Function::InternalLinkage, subprog.name(), module_.get());
+      func_type, Function::ExternalLinkage, subprog.name(), module_.get());
+
+  // Mark function as NoInline. This has several reasons:
+  // - inlining takes up BPF instructions
+  // - LLVM sometimes inlines recursive calls into loops, which can be then
+  //   rejected by the BPF verifier for code where a recursive call would
+  //   not
+  AttributeList attrs;
+  attrs = attrs.addFnAttribute(b_.getContext(), Attribute::NoInline);
```
jordalgo, line 2566: Looks like this doesn't exist for llvm 12 and 13 (I think `addFnAttribute` was added in 14) so we might need a `#if LLVM_VERSION_MAJOR`

jordalgo:
@lenticularis39 Thanks for working on this! Similar to @ajor I'm starting to play with it and just documenting some random issues. 

---

This yields a verifier error for the unused `add1` function (though maybe this will go away after we finish libbpf migration). Maybe something we can catch earlier in semantic analyser.
```
fn add($a: int64, $b: int64): int64 { return $a + $b; }
fn add1($a: int64): int64 { return $a + $a; }
BEGIN { print(add(1, 1)); }
```

Also seeing it here:
```
fn add1($a: int64, $b: int64): int64 { return $a + $b; }
fn add2($a: int64): int64 { return $a + $a; }
BEGIN { print(add1(1, 1)); }
hardware:cpu-cycles:1e8 { print(add2(2)); }
```

---

I get the shadow warning but it seems like this program doesn't work:

```
fn printf($a: int64): void { print(($a)); }
BEGIN { printf(3); }
```

---

We should support some of the aggregate map types e.g. `count`, `sum`, `min`, `max`, `avg`

```
fn go($a: uint64): void { print(($a)); }
BEGIN { @ = count(); go(@); }
```

---

Should we be catching unused returns? e.g.

```
fn add1($a: int64, $b: int64): int64 { return $a + $b; }
BEGIN { add1(1, 1); }
```

---

We shouldn't allow statements after return values. Fortunately, this doesn't actually add to sum.

```
fn add1($a: int64, $b: int64): int64 { return $a + $b; @ = sum(2); }
BEGIN { @ = sum(add1(1, 1)); }
```

---

This should be caught in semantic analysis. Right now it causes a segfault
```
fn add($a: int64, $b: int64): int64 { exit(); return $a + $b; }
```

---

Questions

- Is it possible to pass strings as arguments and if so, how?
- Should we disallow some keywords as function names e.g. ` BEGIN {BEGIN(1, 1); }` ?

lenticularis39:
> Is it possible to pass strings as arguments and if so, how?

In the current implementation, no, since arguments are passed in BPF registers and thus are limited to four words. This can be improved by storing function arguments in a BPF map ("stack").

> Should we disallow some keywords as function names e.g.  BEGIN {BEGIN(1, 1); } ?

Keywords are already disallowed by the parser implementation which requires the name of a function to be an identifier. `BEGIN` is not a keyword, it is a name of an attach point; it's unusual and confusing to use it as a name of a function, but I see little reason why we should disallow that.

PR #3069, opened 2024-03-18
Fix MSAN out-of-bounds error (please ignore for the moment)

bjackman:
does this run actions...?


bjackman:
Oops, I didn't mean to make this pull request against the upstream repo 😬

I'll leave it here instead of deleting it only to (hopefully) recreate it later. But please ignore it for the moment!

bjackman:
OK, in fact since the code doesn't even build I will just delete it since I probably need to take a bit more time to actually make it not garbage :)

bjackman:
In case anyone is curious about this in the future - the issue has been fixed already. And the reason this PR didn't build is because it was just a git brainfart, mixing up code from Google's current legacy version (where there's an MSan bug) and the upstream version (where it's already fixed).

PR #3073, opened 2024-03-18
output: Flush output buffers on exit

danobi:
Unbuffer all output during runtime tests. There seems to be some kind
of buffering issue inside of glibc, although I cannot reproduce it locally.

I debugged it as far down as knowing we put the right output into
`std::cout`. But for some reason it doesn't come out in CI. Unbuffering
seems to help.

This fixes the recent flakiness in `call.ustack_*` runtime tests.

##### Checklist

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -233,53 +233,16 @@ def test():
 
     results.append(
         test_one(
-            "bpftrace_test",
-            lambda: truthy(RUN_TESTS),
-            lambda: shell(
-                ["./tests/bpftrace_test"],
-                cwd=Path(BUILD_DIR),
-                env={"GTEST_COLOR": GTEST_COLOR},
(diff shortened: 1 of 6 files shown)


danobi:
> > But for text based printf() output, we don't use std::endl
> 
> JC, should we update this too to make sure that every print is as immediate as possible?

I don't think we can cuz the output comes from the user in the program. I suppose we could try to do a transform but that feels a bit odd.

I suppose we could run a flush after each output message. I'll need to take a closer look at where to hook into.

danobi:
Also it looks like this doesn't completely fix the issue. I'll have to keep debugging.

danobi:
Towel is thrown. I still think the real bug is in the runtime test suite somewhere, but unclear why it only affects ustack

PR #3078, opened 2024-03-20
Disable flakey ustack tests in CI

danobi:
See commits for more details

##### Checklist

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -65,9 +65,9 @@ Each runtime testcase consists of multiple directives. In no particular order:
   `RUN` unless you must pass flags or create a shell pipeline.  This XOR the
   `RUN` field is required
 * `EXPECT`: The expected output. Python regular expressions are supported.
-   One or more `EXPECT` and `EXPECT_NONE` or a single `EXPECT_FILE` or 
+   One or more `EXPECT` and `EXPECT_NONE` or a single `EXPECT_FILE` or
    `EXPECT_JSON` is required.
-* `EXPECT_NONE`: The negation of `EXPECT`, which also supports Python regular 
+* `EXPECT_NONE`: The negation of `EXPECT`, which also supports Python regular
    expressions.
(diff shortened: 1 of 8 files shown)


tests/runtime/call
```diff
@@ -232,27 +232,44 @@ ENV BPFTRACE_STACK_MODE=bpftrace
...
+# [1]: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+SKIP_IF_ENV_HAS CI=true
 
 NAME ustack_stack_mode_env_perf
 PROG u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }
 ENV BPFTRACE_STACK_MODE=perf
 EXPECT_REGEX \d+ uprobeFunction1\+0 \(.*/uprobe_loop\)
 TIMEOUT 5
 AFTER ./testprogs/uprobe_loop
+# See above
```
jordalgo, line 253: Can we add a link to an existing issue here instead of `#See above` ?

tests/README.md
```diff
@@ -97,6 +97,8 @@ Each runtime testcase consists of multiple directives. In no particular order:
 * `NEW_PIDNS`: This will execute the `BEFORE`, the bpftrace (`RUN` or `PROG`),
   and the `AFTER` commands in a new pid namespace that mounts proc. At least one
   `BEFORE` is required.
+* `SKIP_IF_ENV_HAS`: Skip test case if specified environment variable is found
```
jordalgo, line 100: Maybe we should mention that only a single KEY/VALUE is per test is allowed

jordalgo:
Few small nits but overall I like this approach better, that is the skipping directive being closer to the test themselves.

tests/runtime/call
danobi: Yeah good idea

jordalgo (approved):

PR #3079, opened 2024-03-20
Ensure pcap writers are always closed

danobi:
Before, if we encountered an error, we could miss closing the pcap writers.

Also remove an unnecessary conditional.

##### Checklist


jordalgo (approved):
Groovy.

PR #3081, opened 2024-03-21
Fix inconsistent lhist bucket labels

ajor:
The 'K' and 'M' suffixes should either be applied to all buckets or to none. This can be achieved by only applying the suffixes if the step size is a multiple of 1024.

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -127,10 +127,13 @@ std::string TextOutput::hist_index_label(uint32_t index, uint32_t k)
   return label.str();
 }
 
-std::string TextOutput::lhist_index_label(int number)
+std::string TextOutput::lhist_index_label(int number, int step)
 {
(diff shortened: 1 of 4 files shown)


danobi (commented):
in theory we should be able to unit test everything in output.cpp, right?

ajor:
> in theory we should be able to unit test everything in output.cpp, right?

Damn it, I wanted to be lazy! 😛

Yeah, you're right, that'd be a better approach. I'll try and update this PR to do that.

ajor:
Updated to create and use unit tests instead of runtime tests. This now depends on #3086, so I've marked it as a draft until that gets merged.

danobi (approved):
lgtm. thanks for setting up the unit tests! next one will be much easier to add

PR #3087, opened 2024-03-22
Some Vagrantfile updates

bjackman:
Hi, I'm not sure if this is of much value. I ended up discovering that my distro packages Nix and abandoning the use of Vagrant, so this has barely been tested, I just built the code once in the new boxen. Nonetheless maybe it saves someone some time down the line...?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -69,13 +69,8 @@ Vagrant.configure("2") do |config|
       'scripts'        => [ $ubuntu_deps, ],
       'fix_console'    => 0,
     },
-    'ubuntu-21.04'     => {
-      'image'          => 'ubuntu/hirsute64',
-      'scripts'        => [ $ubuntu_deps, ],
-      'fix_console'    => 0,
-    },
-    'ubuntu-21.10'     => {
-      'image'          => 'ubuntu/impish64',
+    'ubuntu-22.04'     => {
+      'image'          => 'ubuntu/jammy64',
       'scripts'        => [ $ubuntu_deps, ],
       'fix_console'    => 0,
     },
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -69,13 +70,14 @@ Vagrant.configure("2") do |config|
       'scripts'        => [ $ubuntu_deps, ],
       'fix_console'    => 0,
     },
-    'ubuntu-21.04'     => {
-      'image'          => 'ubuntu/hirsute64',
+    'ubuntu-22.04'     => {
```
bjackman, line 73: Oh actually I forgot `llvm.sh` doesn't work on 22.04. So might as well just delete this.

viktormalik:
Thanks for opening this!

Even though adding Ubuntu 22.04 may not work, I think that it would be nice to do something about our Vagrant support - either fix it to some minimal working state or delete it completely.

fbs:
My main reason for vagrant was that a few years ago it was the easiest way to do bpftrace dev on macos. Nowdays `lima` might be a better tool (e.g. https://github.com/lima-vm/lima/blob/master/examples/ubuntu.yaml ). Or maybe even with the various docker engines as they all run in VMs anyway

bjackman:
Yeah, I did also wonder if a Dockerfile/Containerfile would make Vagrant
irrelevant here. It seems it might be easier to make it hermetic in the
ways that matter (software dependencies), and you can avoid the pointless
hermeticism that Vagrant gives you (fixed CPUs/RAM in the guest).

danobi:
Yeah vagrant may not be the best thing to support going forwards. I recall I spent around a day trying to get vagrant to work on an M1 mac and it just didn't. None of the providers supported M1, at least ~6 months ago. Not sure what current status is

viktormalik:
Agreed, supporting Vagrant with different providers is difficult. My suggestion would be to remove `Vagrantfile` completely.

danobi:
Ok, let's go with deleting vagrant then

PR #3091, opened 2024-03-28
Replace std::runtime_error with LOG(FATAL)

jordalgo:
There are still some cases where std::runtime_errors are thrown but I didn't change them because either:
- the linter got angry because the function wasn't returning a value
- the throw/catching behavior wasn't obvious enough that I felt comfortable changing it here (e.g. child.cpp)
- we are testing that an error gets thrown in a constructor - LOG(FATAL) doesn't work here because we can't catch it

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -395,35 +395,45 @@ SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
   return stype;
 }
 
-Struct BTF::resolve_args(const std::string &func, bool ret)
+std::optional<Struct> BTF::resolve_args(const std::string &func,
+                                        bool ret,
+                                        std::string &err)
 {
-  if (!has_data())
-    throw std::runtime_error("BTF data not available");
+  if (!has_data()) {
+    err = "BTF data not available";
+    return std::nullopt;
+  }
 
   auto func_id = find_id(func, BTF_KIND_FUNC);
-  if (!func_id.btf)
-    throw std::runtime_error("no BTF data for " + func);
+  if (!func_id.btf) {
+    err = "no BTF data for " + func;
+    return std::nullopt;
+  }
 
   const struct btf_type *t = btf__type_by_id(func_id.btf, func_id.id);
   t = btf__type_by_id(func_id.btf, t->type);
-  if (!t || !btf_is_func_proto(t))
-    throw std::runtime_error(func + " is not a function");
+  if (!t || !btf_is_func_proto(t)) {
+    err = func + " is not a function";
+    return std::nullopt;
+  }
 
   if (bpftrace_ && !bpftrace_->is_traceable_func(func)) {
-    if (bpftrace_->get_traceable_funcs().empty())
-      throw std::runtime_error("could not read traceable functions from " +
-                               tracefs::available_filter_functions() +
-                               " (is tracefs mounted?)");
-    else
-      throw std::runtime_error("function not traceable (probably it is "
-                               "inlined or marked as \"notrace\")");
+    if (bpftrace_->get_traceable_funcs().empty()) {
+      err = "could not read traceable functions from " +
+            tracefs::available_filter_functions() + " (is tracefs mounted?)";
+    } else {
+      err = "function not traceable (probably it is "
+            "inlined or marked as \"notrace\")";
+    }
+    return std::nullopt;
   }
 
   const struct btf_param *p = btf_params(t);
   __u16 vlen = btf_vlen(t);
   if (vlen > arch::max_arg() + 1) {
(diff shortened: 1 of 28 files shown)


src/procmon.cpp
```diff
@@ -27,7 +27,12 @@ static inline int pidfd_open(int pid, unsigned int flags)
 
 ProcMon::ProcMon(const std::string& pid)
 {
-  setup(parse_pid(pid));
+  std::string err;
+  auto maybe_pid = parse_pid(pid, err);
+  if (!maybe_pid.has_value()) {
+    throw std::runtime_error("Failed to parse pid: " + err);
```
danobi, line 33: Looks like in main() it's just a catch + exit anyways: https://github.com/bpftrace/bpftrace/blob/cf404154a06709efcd98171ed5700e3103555136/src/main.cpp#L787-L794

So I think LOG(FATAL) could be appropriate here. IMHO I'd just remove the `std::string&` overload and force the caller to parse the pid.

src/utils.cpp
```diff
@@ -1102,25 +1101,40 @@ bool symbol_has_cpp_mangled_signature(const std::string &sym_name)
...
 }
 
-pid_t parse_pid(const std::string &str)
+static std::string get_invalid_pid_message(const std::string &pid,
+                                           const std::string &msg)
 {
+  return "pid '" + pid + "' " + msg;
+}
+
+std::optional<pid_t> parse_pid(const std::string &str, std::string &err)
```
danobi, line 1110: Would be good to add unit tests for this but it's out of scope of this PR

danobi (commented):
> the linter got angry because the function wasn't returning a value

interesting, thought the `[[noreturn]]` we added on `LogStreamFatal::~LogStreamFatal()` was supposed to fix that

src/procmon.cpp
jordalgo: Just an FYI: I tried to add a bunch of `LOG(FATAL)`s to replace all the throws in this file and I ran into two issues:
- checking error behavior in unit tests isn't possible if we call `LOG(FATAL)`
- If I remove the `try/catch` in main then the error output is strange and I couldn't match against it properly in the runtime test (this one probably just needs investigation).

But yeah, I can remove the overload - that will help.

src/utils.cpp
jordalgo: We have 4 runtime tests for this but yeah, unit tests would be good too.

src/btf.cpp
```diff
@@ -395,35 +395,45 @@ SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
   return stype;
 }
 
-Struct BTF::resolve_args(const std::string &func, bool ret)
+std::optional<Struct> BTF::resolve_args(const std::string &func,
+                                        bool ret,
+                                        std::string &err)
```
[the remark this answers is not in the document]
jordalgo: Ohhh it's like an Either monad!

src/procmon.cpp
danobi: Makes sense

danobi (approved):

src/main.cpp
```diff
@@ -785,8 +785,13 @@
   bpftrace.delta_taitime_ = get_delta_taitime();
 
   if (!args.pid_str.empty()) {
+    std::string err;
```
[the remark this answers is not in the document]
viktormalik: I know that this is just a "Note" but renaming `err` to `errmsg` seems like the easiest thing to do to get rid of it.

src/probe_matcher.cpp
```diff
@@ -212,8 +212,9 @@ std::unique_ptr<std::istream> ProbeMatcher::get_symbols_from_file(
 {
   auto file = std::make_unique<std::ifstream>(path);
   if (file->fail()) {
-    throw std::runtime_error("Could not read symbols from " + path + ": " +
-                             strerror(errno));
+    LOG(WARNING) << "Could not read symbols from " << path << ": "
+                 << strerror(errno);
+    return NULL;
```
viktormalik, line 217: Nit: we prefer `nullptr` (even though the original code didn't respect that).
```suggestion
    return nullptr;
```

viktormalik (commented):
Couple of nits + agreed with @danobi that this should have a CHANGELOG entry.

src/main.cpp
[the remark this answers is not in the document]
jordalgo: Ah yeah good call. I feel like this should be a warning not a "note" as I completely didn't pay attention.

jordalgo:
Finally added a changelog entry 😉

viktormalik (approved):
Nice cleanup!

PR #3094, opened 2024-03-28
Parse C++ class and inheritance from Debug Info

ttreyer:
Extend the Dwarf parser to make class members and parents' members accessible.

This patch handles any shape of inheritance hierarchy, but does provide access to parents' shadowed members:
```
struct Parent { int abc, xyz; };
struct Child : Parent { int abc; };
...
child->abc          // Access the child's member
child->xyz          // Access the parent's member
child->Parent::abc  // Not implemented: access the parent's shadowed member
```

Multiple inheritance suffer from a similar problem: when a parent shadows the member of another parent.
```
struct GrandChild : Parent, Child {};
g_child->xyz // Access Parent's member
g_child->abc // Ambiguous! Will access Parent::abc, but it should reject and require a qualified name instead
```

##### Checklist

### DIFF
--- a/src/ast/passes/printer.cpp
+++ b/src/ast/passes/printer.cpp
@@ -207,7 +207,7 @@ void Printer::visit(Ternary &ternary)
 void Printer::visit(FieldAccess &acc)
 {
   std::string indent(depth_, ' ');
-  out_ << indent << "." << std::endl;
+  out_ << indent << "." << type(acc.type) << std::endl;
 
   ++depth_;
   acc.expr->accept(*this);
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -537,7 +537,7 @@ void SemanticAnalyser::visit(Call &call)
         map.skip_key_validation = true;
     }
 
-    expr.accept(*this);
+    Visit(call.vargs[i]);
   }
 
   if (auto probe = dynamic_cast<Probe *>(scope_)) {
@@ -1352,7 +1352,7 @@ void SemanticAnalyser::visit(Sizeof &szof)
 {
   szof.type = CreateUInt64();
   if (szof.expr) {
-    szof.expr->accept(*this);
+    Visit(szof.expr);
     szof.argtype = szof.expr->type;
   }
   resolve_struct_type(szof.argtype, szof.loc);
@@ -1362,7 +1362,7 @@ void SemanticAnalyser::visit(Offsetof &ofof)
 {
   ofof.type = CreateUInt64();
   if (ofof.expr) {
-    ofof.expr->accept(*this);
+    Visit(ofof.expr);
     ofof.record = ofof.expr->type;
   }
   resolve_struct_type(ofof.record, ofof.loc);
@@ -1462,8 +1462,8 @@ void SemanticAnalyser::visit(Map &map)
   MapKey key;
 
   for (unsigned int i = 0; i < map.vargs.size(); i++) {
+    Visit(map.vargs.at(i));
     Expression *expr = map.vargs.at(i);
-    expr->accept(*this);
 
     // Insert a cast to 64 bits if needed by injecting
     // a cast into the ast.
@@ -1473,7 +1473,7 @@ void SemanticAnalyser::visit(Map &map)
                                                   : CreateUInt64(),
                                               expr,
                                               map.loc);
-      cast->accept(*this);
+      Visit(cast);
       map.vargs.at(i) = cast;
       expr = cast;
     } else if (expr->type.IsPtrTy() && expr->type.IsCtxAccess()) {
@@ -1531,8 +1531,8 @@ void SemanticAnalyser::visit(Variable &var)
 
 void SemanticAnalyser::visit(ArrayAccess &arr)
 {
-  arr.expr->accept(*this);
-  arr.indexpr->accept(*this);
+  Visit(arr.expr);
+  Visit(arr.indexpr);
 
   SizedType &type = arr.expr->type;
   SizedType &indextype = arr.indexpr->type;
@@ -1772,8 +1772,8 @@ void SemanticAnalyser::binop_ptr(Binop &binop)
 
 void SemanticAnalyser::visit(Binop &binop)
 {
-  binop.left->accept(*this);
-  binop.right->accept(*this);
+  Visit(binop.left);
+  Visit(binop.right);
 
   auto &lht = binop.left->type;
   auto &rht = binop.right->type;
@@ -1876,7 +1876,7 @@ void SemanticAnalyser::visit(Unop &unop)
     }
   }
 
-  unop.expr->accept(*this);
+  Visit(unop.expr);
 
   auto valid_ptr_op = false;
   switch (unop.op) {
@@ -1892,6 +1892,8 @@ void SemanticAnalyser::visit(Unop &unop)
   if (is_final_pass()) {
     // Unops are only allowed on ints (e.g. ~$x), dereference only on pointers
     // and context (we allow args->field for backwards compatibility)
+    // References are not allowed, instead they get turned into pointers by
+    // the `dereference_if_needed()` function, during the `Visit()` above.
     if (!type.IsIntegerTy() &&
         !((type.IsPtrTy() || type.IsCtxAccess()) && valid_ptr_op)) {
       LOG(ERROR, unop.loc, err_)
@@ -1944,9 +1946,10 @@ void SemanticAnalyser::visit(Unop &unop)
 
 void SemanticAnalyser::visit(Ternary &ternary)
 {
-  ternary.cond->accept(*this);
-  ternary.left->accept(*this);
-  ternary.right->accept(*this);
+  Visit(ternary.cond);
+  Visit(ternary.left);
+  Visit(ternary.right);
+
   const Type &cond = ternary.cond->type.GetTy();
   const Type &lhs = ternary.left->type.GetTy();
   const Type &rhs = ternary.right->type.GetTy();
@@ -1974,7 +1977,7 @@ void SemanticAnalyser::visit(Ternary &ternary)
 
 void SemanticAnalyser::visit(If &if_block)
 {
-  if_block.cond->accept(*this);
+  Visit(if_block.cond);
 
   if (is_final_pass()) {
     const Type &cond = if_block.cond->type.GetTy();
@@ -1988,7 +1991,7 @@ void SemanticAnalyser::visit(If &if_block)
 
 void SemanticAnalyser::visit(Unroll &unroll)
 {
-  unroll.expr->accept(*this);
+  Visit(unroll.expr);
 
   auto unroll_value = bpftrace_.get_int_literal(unroll.expr);
   if (!unroll_value.has_value()) {
@@ -2011,8 +2014,9 @@ void SemanticAnalyser::visit(Jump &jump)
 {
   switch (jump.ident) {
     case JumpType::RETURN:
-      if (jump.return_value)
-        jump.return_value->accept(*this);
+      if (jump.return_value) {
+        Visit(jump.return_value);
+      }
       if (auto subprog = dynamic_cast<Subprog *>(scope_)) {
         if ((subprog->return_type.IsVoidTy() !=
              (jump.return_value == nullptr)) ||
@@ -2043,7 +2047,7 @@ void SemanticAnalyser::visit(While &while_block)
            " on LLVMs loop unroll to generate loadable code.";
   }
 
-  while_block.cond->accept(*this);
+  Visit(while_block.cond);
 
   loop_depth_++;
   accept_statements(while_block.stmts);
@@ -2173,7 +2177,7 @@ void SemanticAnalyser::visit(For &f)
   }
 
   map.skip_key_validation = true;
-  map.accept(*this);
+  Visit(&map);
 
   if (has_error())
     return;
@@ -2272,7 +2276,7 @@ void SemanticAnalyser::visit(FieldAccess &acc)
   // A field access must have a field XOR index
   assert((acc.field.size() > 0) != (acc.index >= 0));
 
-  acc.expr->accept(*this);
+  Visit(acc.expr);
 
   SizedType &type = acc.expr->type;
 
@@ -2303,7 +2307,11 @@ void SemanticAnalyser::visit(FieldAccess &acc)
       return;
     auto arg = bpftrace_.structs.GetProbeArg(*probe, acc.field);
     if (arg) {
-      acc.type = arg->type;
+      // Only set acc.type if it's not already set. [1]
+      // Overwriting the type once set could override the conversion
+      // from Reference to Pointer done by dereference_if_needed.
+      if (acc.type.IsNoneTy())
+        acc.type = arg->type;
       acc.type.SetAS(acc.expr->type.GetAS());
 
       if (is_final_pass()) {
@@ -2404,7 +2412,10 @@ void SemanticAnalyser::visit(FieldAccess &acc)
         }
       }
 
-      acc.type = field.type;
+      // Only set acc.type if it's not already set. See explanation [1].
+      if (acc.type.IsNoneTy())
+        acc.type = field.type;
+
       if (acc.expr->type.IsCtxAccess() &&
           (acc.type.IsArrayTy() || acc.type.IsRecordTy())) {
         // e.g., ((struct bpf_perf_event_data*)ctx)->regs.ax
@@ -2425,7 +2436,7 @@ void SemanticAnalyser::visit(FieldAccess &acc)
 
 void SemanticAnalyser::visit(Cast &cast)
 {
-  cast.expr->accept(*this);
+  Visit(cast.expr);
 
   // cast type is synthesised in parser, if it is a struct, it needs resolving
   resolve_struct_type(cast.type, cast.loc);
@@ -2503,8 +2514,8 @@ void SemanticAnalyser::visit(Cast &cast)
 void SemanticAnalyser::visit(Tuple &tuple)
 {
   std::vector<SizedType> elements;
-  for (auto elem : tuple.elems) {
-    elem->accept(*this);
+  for (auto &elem : tuple.elems) {
+    Visit(elem);
 
     // If elem type is none that means that the tuple contains some
     // invalid cast (e.g., (0, (aaa)0)). In this case, skip the tuple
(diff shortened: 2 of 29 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -3324,5 +3355,18 @@ Pass CreateSemanticPass()
   return Pass("Semantic", fn);
 };
 
+void SemanticAnalyser::dereference_if_needed(Expression *&expr)
+{
+  const auto &type = expr->type;
+  if (type.IsRefTy()) {
+    expr = new Unop(Operator::MUL, expr, expr->loc);
```
ajor, line 3362: If we convert the wrapped reference here into a pointer, then I think that will simplify later passes by not having to care about reference types any more. I'm hoping we can get away with zero changes to codegen and field_analyser. With luck, it might help with the crash I found too.

Here's an example to explain what I mean a bit better.
Original AST:
```
. :: [reference (struct Foo)]
  builtin: args :: [struct uprobeFunction1_args]
  foo
```

Current transformation:
```
dereference :: [struct Foo]
  . :: [reference (struct Foo)]
    builtin: args :: [struct uprobeFunction1_args]
    foo
```

AST after converting reference into pointer:
```
dereference :: [struct Foo]
  . :: [pointer (struct Foo)]
    builtin: args :: [struct uprobeFunction1_args]
    foo
```

ajor (changes requested):
I was able to cause an abort with this script when it's triggered:
```
uprobe:./tests/testprogs/uprobe_test_cxx:cpp:uprobeFunction1 { print(args.foo) }
```

This change could do with some semantic analyser unit tests as well, to confirm the AST transformation occurs as expected.

ajor:
Btw, I like how much simpler your method is for generating a DWARF data source - maybe we could switch the C data source across to that way in the future?

ttreyer:
> Btw, I like how much simpler your method is for generating a DWARF data source - maybe we could switch the C data source across to that way in the future?

We might be able to simplify it, but the C data source has the extra requirement of generating BTF information, which is not supported for C++.

tests/runtime/dwarf
```diff
@@ -24,6 +24,19 @@ REQUIRES bash -c "exit 1" # SKIP: anonymous parameter type not supported #3083
...
 
+NAME list uprobe args - class reference type
+RUN {{BPFTRACE}} -lv 'uprobe:./testprogs/uprobe_test_cxx:cpp:uprobeFunction3'
+EXPECT Child & c
+EXPECT Parent & p
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+
+NAME list uprobe args - array reference type
+RUN {{BPFTRACE}} -lv 'uprobe:./testprogs/uprobe_test_cxx:_Z11uprobeArrayRA10_i'
```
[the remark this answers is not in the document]
ttreyer: I tried, but this particular function is not recognised when using its demangled name.
I haven't searched why yet, but I'll give it a try...

src/ast/passes/semantic_analyser.cpp
ttreyer: Brilliant!

src/types.h
```diff
@@ -455,12 +466,20 @@ class SizedType {
   friend std::ostream &operator<<(std::ostream &, const SizedType &);
   friend std::ostream &operator<<(std::ostream &, Type);
 
+  SizedType &IntoPointer()
```
ajor, line 469: I feel like this function has the potential to be a bit confusing - it's modifying the type in place, but also returning a type. On my first reading of the call site I assumed that this returned type was a newly created value. Is there a benefit to this current design?

tests/runtime/dwarf
```diff
@@ -128,6 +141,61 @@ REQUIRES_FEATURE dwarf
...
+EXPECT foo.a = 1
+       foo.b = 2
+       foo.c = 3
+       foo.x = 42
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/uprobe_test_cxx
+
+NAME uprobe arg as reference - array
+PROG uprobe:./testprogs/uprobe_test_cxx:_Z11uprobeArrayRA10_i {
```
ajor, line 189:
```suggestion
PROG uprobe:./testprogs/uprobe_test_cxx:cpp:uprobeArray {
```

ajor (commented):
Looking good! Just a few small comments, and don't forget about the semantic analyser tests :)

src/types.h
ttreyer: No particular benefit, I needed to get the `SizedType` right after converting the ref into a pointer.
I've changed it to return `void` so it doesn't look like it's returning a new type anymore.

src/probe_matcher.cpp
```diff
@@ -67,7 +67,7 @@ std::set<std::string> ProbeMatcher::get_matches_in_stream(
                           [&](const std::string& token) {
                             return token.find("(") == std::string::npos;
                           })) {
-            match_line = match_line.substr(0, match_line.find_last_of("("));
+            match_line = match_line.substr(0, match_line.find_first_of("("));
```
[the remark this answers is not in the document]
ttreyer: Ping @viktormalik

tests/data/CMakeLists.txt
```diff
@@ -98,3 +99,8 @@ add_custom_command(
   DEPENDS ${BTF_DATA_HEX} ${FUNC_LIST_HEX} ${CONFIGURE_BTF_HEADERS})
 
 add_custom_target(debuginfo_btf_data DEPENDS ${BTF_DATA_H})
+
+# BTF doesn't support C++, so we only generate a data_source_cxx executable
+# to run the semantic_analyser tests on.
```
viktormalik, line 104: Nit: we're not running any semantic analyser tests, rather field analyser tests:
```suggestion
# to run the field_analyser tests on.
```

tests/data/CMakeLists.txt
```diff
@@ -98,3 +99,8 @@ add_custom_command(
   DEPENDS ${BTF_DATA_HEX} ${FUNC_LIST_HEX} ${CONFIGURE_BTF_HEADERS})
 
 add_custom_target(debuginfo_btf_data DEPENDS ${BTF_DATA_H})
+
+# BTF doesn't support C++, so we only generate a data_source_cxx executable
```
viktormalik, line 103: I understand that this is much easier to do than what we do for `data_source` but I'm not a big fan of using different approaches for generating C and C++ DWARF data. Maybe we could try to unify them somehow?

I wouldn't be against using the same "just use the compiled binary" approach for C dwarf data, too. It's unnecessarily complicated (as it was inspired by the BTF approach which has some specifics) and requires additional tools (`strip`, `xxd`) during the build.

tests/field_analyser.cpp
```diff
@@ -478,6 +480,157 @@ TEST_F(field_analyser_dwarf, parse_arrays)
...
+  std::string uprobe = "uprobe:" + std::string(cxx_bin_);
+  test(bpftrace, uprobe + ":cpp:func_2 { $x = args.lc->a; }", 0);
+
+  ASSERT_TRUE(bpftrace.structs.Has("GrandChild"));
+  auto cls = bpftrace.structs.Lookup("GrandChild").lock();
+
+  ASSERT_TRUE(cls->HasFields());
+  // The child has a total of 7 fields with its parent's,
+  // but field 'd' is shadowed by Child's field 'd'
+  // so only 6 are showned in the Child's struct
```
viktormalik, line 588: The comment is the same as above but we're using a different type here. Maybe adjust it to prevent confusion?

tests/field_analyser.cpp
```diff
@@ -478,6 +480,157 @@ TEST_F(field_analyser_dwarf, parse_arrays)
...
+}
+
+TEST_F(field_analyser_dwarf, parse_inheritance_multi)
+{
+  BPFtrace bpftrace;
+  std::string uprobe = "uprobe:" + std::string(cxx_bin_);
+  test(bpftrace, uprobe + ":cpp:func_3 { $x = args.m->abc; }", 0);
+
+  std::shared_ptr<Struct> cls;
+  if (LLVM_VERSION_MAJOR <= 12) {
```
viktormalik, line 602: Shouldn't rather use conditional compilation?
```suggestion
#if LLVM_VERSION_MAJOR <= 12
```

viktormalik (commented):
> To save me a bit of time reviewing: what was the reason for the recent changes to dwarf_parser, making resolve_fields() take a shared_ptr<Struct>?

Yeah, this got me confused, too.

Also, the individual commits are not very clean and very likely won't work on their own. I'm ok with that as long as we squash the PR in the end.

ajor:
> Also, the individual commits are not very clean and very likely won't work on their own. I'm ok with that as long as we squash the PR in the end.

Yep, we'll squash before merging. That's the approach I think we should be taking with the vast majority of PRs anyway.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -531,6 +531,7 @@ void SemanticAnalyser::visit(Call &call)
       }
 
       expr.accept(*this);
+      dereference_if_needed((*call.vargs)[i]);
```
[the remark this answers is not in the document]
ttreyer: > Requiring to call dereference_if_needed on every node that may produce an expression feels tedious and very error-prone. Sooner or later, someone will forget to add it. Maybe we could instead make accept methods do it automatically?

It is! The latest revision hides `SemanticAnalyser::Visit(Node&)` with `SemanticAnalyser::Visit(Node * const &)` and specialising `SemanticAnalyser::Visit(Expression *&)`. It enables modifying expressions "in-place" for the `Expression` case, but no other Node type.

> Or, alternatively, add a pass before semantic analysis which will add the dereferences where needed?

The dereferencing can't go in its own pass, as it relies on type propagation done by `SemanticAnalyser`.

tests/field_analyser.cpp
ttreyer: I'd use the pre-processor if it were to handle a change of API between versions.
Here it is used to handle a change in behaviour between LLDB 12 and the later versions. So we can use a regular `if/else`. Moreover, the static values enable the compiler to eliminate the dead branch

src/ast/passes/printer.cpp
```diff
@@ -210,7 +210,7 @@ void Printer::visit(Ternary &ternary)
 void Printer::visit(FieldAccess &acc)
 {
   std::string indent(depth_, ' ');
-  out_ << indent << "." << std::endl;
+  out_ << indent << "." << type(acc.type) << std::endl;
```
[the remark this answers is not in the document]
ttreyer: The `type` function makes the following output:
```
[...]
  . :: [struct Foo *, AS(user)]
    $x
    field
```

ttreyer:
> To save me a bit of time reviewing: what was the reason for the recent changes to `dwarf_parser`, making `resolve_fields()` take a `shared_ptr<Struct>`?

The new `resolve_fields()` that takes the `shared_ptr` is a `private` method that populates the fields without needing to do any lookup on `bpftrace_->structs` or LLDB's `target.FindFirstType()`.
For `record` types, we already have the `lldb::SBType`, so we can populate the `Struct` without any lookup.

The previous public version of `resolve_fields()` is still available and do the lookup, before calling the new `resolve_fields()`.

Saving a lookup is nice, but I also had a problem with the type name: I'm currently using `GetDisplayName()` which provides a user friendly name (`std::vector<int>`), but there is also `GetName()` which returns the full name of the type (`std::vector<int, std::allocator<int>>`).
LLDB needs the full name to lookup for a type by its name, so another fix could be to use the full name instead of the display name. However, the current approach with two `resolve_fields()` function solves the lookup issue I had, while preserving a user friendly type name.

Which name should we display to the user? The full one for completeness? Or the display name for friendliness?

tests/runtime/dwarf
```diff
@@ -128,6 +142,73 @@ REQUIRES_FEATURE dwarf
...
+RUN {{BPFTRACE}} --dry-run -d ast -e 'uprobe:./testprogs/uprobe_test_cxx:cpp:uprobeFunction1 { args.foo }'
+EXPECT AST after: Semantic
+       -------------------
+       Program
+        uprobe:./testprogs/uprobe_test_cxx:cpp:uprobeFunction1
+         dereference
+          . :: [struct Foo *, AS(user)]
+           builtin: args :: [struct uprobe:./testprogs/uprobe_test_cxx:cpp:uprobeFunction1_args, ctx: 1, AS(user)]
+           foo
+TIMEOUT 5
```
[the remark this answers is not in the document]
ttreyer: The issue is the test relies on debug information, so it needs to run against `tests/testprogs/uprobe_test_cxx`. I don't see how I could have a `reference` in the AST otherwise. Could I pass an AST as an input, instead of a BpfTrace script?
Otherwise, I can update the unit tests, so they can access the `testprogs`, and have it run against `uprobe_test_cxx`.

src/ast/passes/printer.cpp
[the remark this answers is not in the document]
viktormalik: Ah, that's nice! Sorry about the confusion, then.

src/ast/passes/semantic_analyser.cpp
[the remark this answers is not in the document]
viktormalik: > > Requiring to call dereference_if_needed on every node that may produce an expression feels tedious and very error-prone. Sooner or later, someone will forget to add it. Maybe we could instead make accept methods do it automatically?
> 
> It is! The latest revision hides `SemanticAnalyser::Visit(Node&)` with `SemanticAnalyser::Visit(Node * const &)` and specialising `SemanticAnalyser::Visit(Expression *&)`. It enables modifying expressions "in-place" for the `Expression` case, but no other Node type.

Do we need to update anything else than `Expression`? If yes, maybe we could change what is exposed.

> > Or, alternatively, add a pass before semantic analysis which will add the dereferences where needed?
> 
> The dereferencing can't go in its own pass, as it relies on type propagation done by `SemanticAnalyser`.

Another reason why we need to factor out type analysis into a separate pass!

tests/field_analyser.cpp
viktormalik: Ok, makes sense.

src/ast/passes/semantic_analyser.cpp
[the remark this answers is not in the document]
ttreyer: > Do we need to update anything else than Expression? If yes, maybe we could change what is exposed.

For references, we only need to change Expression.
I still need to find how to silence the warning of the new `SemanticAnalyser::Visit` functions that are shadowing `Visitor::Visit`.

tests/runtime/dwarf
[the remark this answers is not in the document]
ajor: Parsing the string representation of an AST would be very nice, but we don't have that yet.

The other option is to mock the data types you need (see `tests/mocks.(h|cpp)`). I think we'll need to insert some dummy data into `bpftrace.structs`.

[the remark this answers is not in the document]
viktormalik: This is exactly what the `test_dwarf` class from `tests/dwarf_common.h` is for. You'll probably need to introduce something like `semantic_analyser_dwarf` test class but I think that this should be doable. See e.g. `field_analyser_dwarf` test class in `tests/field_analyser.cpp` for inspiration.

ajor (commented):
I only have one suggestion on functionality: it'd seem better to me to store the fields of structs in order of their offsets, even when they're inherited fields from parents.

e.g. for these types:
```
struct Base {
  int baseMember;
};
struct Child : public Base {
  int childMember;
};
```

Printing a `Child` type would currently display:
```
{ .childMember = 123, .baseMember = 456 }
```

But should probably display this instead:
```
{ .baseMember = 456, .childMember = 123 }
```

I'm imagining that this'll be an easy fix, but feel free to open up an issue and deal with it at a later date if not.

tests/field_analyser.cpp
```diff
@@ -3,6 +3,8 @@
 #include "mocks.h"
 #include "gtest/gtest.h"
 
+#include <llvm/Config/llvm-config.h>
```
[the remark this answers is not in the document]
ttreyer: Non needed anymore 🙂

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1890,8 +1890,8 @@ void SemanticAnalyser::visit(Unop &unop)
 
   SizedType &type = unop.expr->type;
   if (is_final_pass()) {
-    // Unops are only allowed on ints (e.g. ~$x), dereference only on pointers
-    // and context (we allow args->field for backwards compatibility)
+    // Unops are only allowed on ints (e.g. ~$x), dereference only on pointers,
+    // references and context (we allow args->field for backwards compatibility)
```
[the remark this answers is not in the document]
ttreyer: The semantic analyser (`dereference_if_needed()`) automatically insert the `Unop` when necessary, so the user dosen't have to dereference a ref themselve.
So the `Unop` gets added to the AST and this comment is more about `Unop`'s internal API that will accept both pointers and references.

src/dwarf_parser.cpp
```diff
@@ -247,18 +247,49 @@ void Dwarf::resolve_fields(const SizedType &type)
...
+      subobjects.push(Subobject{ base.GetType(),
+                                 subobject.offset + base.GetOffsetInBytes() });
+    }
+
+    subobjects.pop();
   }
+
+  std::sort(str->fields.begin(),
+            str->fields.end(),
+            [](const Field &a, const Field &b) { return a.offset < b.offset; });
```
ajor, line 292: Don't forget to test this please!

We need tests for any behaviour that we want to ensure is preserved across potential future refactors or rewrites

ajor (approved):
Nice!

tests/field_analyser.cpp
```diff
@@ -478,6 +492,157 @@ TEST_F(field_analyser_dwarf, parse_arrays)
...
+  EXPECT_EQ(cls->GetField("b").type.GetSize(), 4);
+  EXPECT_EQ(cls->GetField("b").offset, 4);
+
+  EXPECT_TRUE(cls->HasField("c"));
+  EXPECT_TRUE(cls->GetField("c").type.IsIntTy());
+  EXPECT_EQ(cls->GetField("c").type.GetSize(), 4);
+  EXPECT_EQ(cls->GetField("c").offset, 8);
+
+  // The Child class also has a field 'd', which shadows the Parent's.
+  if (is_d_shadowed == false) {
```
viktormalik, line 514: Nit
```suggestion
  if (!is_d_shadowed) {
```

viktormalik (approved):
LGTM, great work!

PR #828, opened 2019-07-16
Add sargX alias to support Go function arguments

acj:
~I played with two different implementations and finally settled on this lexer spec approach. Adding something to the parser was another possibility, but it was fairly ugly and referenced a lot more types.~

~If this lexer rule sees an `sarg[0-9]` token, it swallows the token and then uses lex's `unput` to feed the stack expression back into the input stream (in reverse order). It's then lexed and parsed as normal.~

Resolves #740.

PR #3095, opened 2024-03-28
Add inline function probing

ttreyer:
Inserting a probe before the stack frame has been established can lead to issues when collecting the stack.
There might be missing frames. See the runtime test `call.ustack_elf_symtable` where `test2` is missing.

When inserting the probe after the prologue of the function, the stack walker has a better context and won't miss an entry.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,11 @@ and this project adheres to
   - [#3060](https://github.com/bpftrace/bpftrace/pull/3060)
 - Disable func builtin for kretprobes and uretprobes when `get_func_ip` feature is not available
   - [#2645](https://github.com/bpftrace/bpftrace/pull/2645)
+- Attach probe to inlined functions
+  - [#3095](https://github.com/bpftrace/bpftrace/pull/3095)
 #### Deprecated
+- Deprecate `sarg` builtin
+  - [#3095](https://github.com/bpftrace/bpftrace/pull/3095)
 #### Removed
 #### Fixed
 - Fix ability to interrupt bpftrace during probe attach
@@ -49,6 +53,8 @@ and this project adheres to
   - [#3077](https://github.com/bpftrace/bpftrace/pull/3077)
 - Fix func builtin for kretprobes and uretprobes for kernels with working `get_func_ip` feature
   - [#2645](https://github.com/bpftrace/bpftrace/pull/2645)
+- Fix ustack for functions with prologue
+  - [#3095](https://github.com/bpftrace/bpftrace/pull/3095)
 #### Docs
 #### Tools
 
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -379,7 +379,10 @@ static int sym_address_cb(const char *symname,
 {
   struct symbol *sym = static_cast<struct symbol *>(p);
 
-  if (sym->address >= start && sym->address < (start + size)) {
+  // When size is 0, then [start, start + size) = [start, start) = ø.
+  // So we need a special case when size=0, but address matches the symbol's
+  if (sym->address == start ||
+      (sym->address > start && sym->address < (start + size))) {
     sym->start = start;
     sym->size = size;
     sym->name = symname;
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -278,6 +278,32 @@ int BPFtrace::add_probe(ast::Probe &p)
             generateWatchpointSetupProbe(match_ap, p));
 
         resources.watchpoint_probes.emplace_back(std::move(probe));
+      } else if (probetype(attach_point->provider) == ProbeType::uprobe) {
+        bool locations_from_dwarf = false;
+
+        // If the user specified an address/offset, do not overwrite
+        // their choice with locations from the DebugInfo.
+        if (probe.address == 0 && probe.func_offset == 0) {
+          // Get function locations from the DebugInfo, as it skips the
+          // prologue and also returns locations of inlined function calls.
+          if (auto *dwarf = get_dwarf(probe.path)) {
+            const auto locations = dwarf->get_function_locations(
+                probe.attach_point, config_.get(ConfigKeyBool::probe_inline));
+            for (const auto loc : locations) {
+              // Clear the attach point, so the address will be used instead
+              Probe probe_copy = probe;
+              probe_copy.attach_point.clear();
+              probe_copy.address = loc;
+              resources.probes.push_back(std::move(probe_copy));
+
+              locations_from_dwarf = true;
+            }
+          }
+        }
+
+        // Otherwise, use the location from the symbol table.
+        if (!locations_from_dwarf)
+          resources.probes.push_back(std::move(probe));
       } else {
         resources.probes.push_back(probe);
       }
--- a/src/dwarf_parser.h
+++ b/src/dwarf_parser.h
@@ -23,6 +23,8 @@ public:
   static std::unique_ptr<Dwarf> GetFromBinary(BPFtrace *bpftrace,
                                               std::string file_path);
 
+  std::vector<uint64_t> get_function_locations(const std::string &function,
+                                               bool include_inlined);
   std::vector<std::string> get_function_params(const std::string &function);
   Struct resolve_args(const std::string &function);
 
@@ -68,6 +70,14 @@ public:
       LOG(WARNING) << "Cannot parse DWARF: liblldb not available";
     warned = true;
     return nullptr;
+  }
+
+  std::vector<uint64_t> get_function_locations(const std::string &function
+                                               __attribute__((unused)),
+                                               bool include_inlined
+                                               __attribute__((unused)))
+  {
+    return {};
   }
 
   std::vector<std::string> get_function_params(const std::string &function
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -59,34 +59,6 @@ EXPECT_REGEX ^SUCCESS 0x[0-9a-f]+$
 TIMEOUT 5
 AFTER ./testprogs/syscall read
 
-NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
-EXPECT SUCCESS 32 64
-ARCH x86_64
-TIMEOUT 5
-AFTER ./testprogs/stack_args
-
-NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
-EXPECT SUCCESS 128 256
-ARCH aarch64|ppc64|ppc64le
-TIMEOUT 5
-AFTER ./testprogs/stack_args
-
-NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
-EXPECT SUCCESS 16 32
-ARCH s390x
-TIMEOUT 5
-AFTER ./testprogs/stack_args
-
-NAME sarg
-PROG uprobe:./testprogs/stack_args:too_many_args { printf("SUCCESS %d %d\n", sarg0, sarg1); exit(); }
-EXPECT SUCCESS 8 16
-ARCH armv7l
-TIMEOUT 5
-AFTER ./testprogs/stack_args
-
 NAME retval
 PROG kretprobe:vfs_read { printf("SUCCESS %d\n", retval); exit(); }
 EXPECT_REGEX SUCCESS .*
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -229,7 +229,7 @@ AFTER ./testprogs/uprobe_loop
 NAME ustack_stack_mode_env_bpftrace
 PROG u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }
 ENV BPFTRACE_STACK_MODE=bpftrace
-EXPECT uprobeFunction1+0
+EXPECT_REGEX ^uprobeFunction1\+[0-9]+$
 TIMEOUT 5
 AFTER ./testprogs/uprobe_loop
 # This was debugged as far down as std::cout having badbit [0] set after a
@@ -249,7 +249,7 @@ SKIP_IF_ENV_HAS CI=true
 NAME ustack_stack_mode_env_perf
 PROG u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }
 ENV BPFTRACE_STACK_MODE=perf
-EXPECT_REGEX \d+ uprobeFunction1\+0 \(.*/uprobe_loop\)
+EXPECT_REGEX ^[0-9a-f]+ uprobeFunction1\+[0-9]+ \(.*/uprobe_loop\)$
 TIMEOUT 5
 AFTER ./testprogs/uprobe_loop
 # See https://github.com/bpftrace/bpftrace/issues/3080
@@ -276,17 +276,8 @@ SKIP_IF_ENV_HAS CI=true
 NAME ustack_elf_symtable
 ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
 PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(ustack); exit(); }
-EXPECT_REGEX ^\s+test\+0\s+main\+[1-9][0-9]*
+EXPECT_REGEX ^\s+test\+[0-9]+\s+test2\+[0-9]+\s+main\+[0-9]+
 AFTER ./testprogs/disable_aslr ./testprogs/uprobe_symres_exited_process
-ARCH x86_64
-TIMEOUT 5
-
-NAME ustack_elf_symtable
-ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
-PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(ustack); exit(); }
-EXPECT_REGEX ^\s+test\+0\s+test2\+[1-9][0-9]*\s+main\+[1-9][0-9]*
-AFTER ./testprogs/disable_aslr ./testprogs/uprobe_symres_exited_process
-ARCH ppc64le
 TIMEOUT 5
 
 NAME cat
--- a/tests/runtime/dwarf
+++ b/tests/runtime/dwarf
@@ -45,6 +45,62 @@ REQUIRES_FEATURE dwarf
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
 
+NAME uprobe skip inlined function
+PROG config = { probe_inline = 0 }
+     uprobe:./testprogs/inline_function:square { @count++ }
+     uretprobe:./testprogs/inline_function:main { exit() }
+EXPECT @count: 1
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/inline_function
+
+NAME uprobe inlined function
+PROG config = { probe_inline = 1 }
+     uprobe:./testprogs/inline_function:square { @count++ }
+     uretprobe:./testprogs/inline_function:main { exit() }
+EXPECT @count: 3
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/inline_function
+
+NAME uprobe inlined function - probe
+PROG config = { probe_inline = 1; cache_user_symbols = "PER_PROGRAM"; }
+     uprobe:./testprogs/inline_function:square {
+       printf("%s\n", probe);
+       if (++@count == 3) { exit(); }
+     }
+EXPECT uprobe:./testprogs/inline_function:square
+       uprobe:./testprogs/inline_function:square
+       uprobe:./testprogs/inline_function:square
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/inline_function
+
+NAME uprobe inlined function - func
+PROG config = { probe_inline = 1; cache_user_symbols = "PER_PROGRAM"; }
+     uprobe:./testprogs/inline_function:square {
+       printf("%s\n", func);
+       if (++@count == 3) { exit(); }
+     }
+EXPECT main
+       main
+       square
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/inline_function
+
+NAME uprobe inlined function - ustack
+PROG config = { probe_inline = 1; cache_user_symbols = "PER_PROGRAM"; }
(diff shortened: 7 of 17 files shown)


ttreyer:
I had to change the `ARG0_STACK` from 8 to 16 for x86_64, since the probe stops after the `push rbp` instruction.
See [System V ABI, page 21](https://gitlab.com/x86-psABIs/x86-64-ABI/-/jobs/artifacts/master/raw/x86-64-ABI/abi.pdf?job=build#page=21).

I still need to check the offset for other archs:
- [x] x86_64
- [x] arm
- [ ] loongarch64
- [ ] mips64
- [ ] ppc64
- [ ] riscv64
- [ ] s390

danobi:
Sweet! This is an excellent fix. Haven't had a chance to look at the code yet; hope to look at the base PR today/tomorrow

tests/testprogs/CMakeLists.txt
```diff
@@ -16,6 +16,9 @@ foreach(testprog_source ${testprog_sources})
       LINK_SEARCH_END_STATIC FALSE
       COMPILE_FLAGS "${testprog_cflags}"
       LINK_FLAGS "-no-pie")
+  if (${testprog_name} STREQUAL "inline_function")
```
jordalgo, line 19: Just curious, is there a benefit of doing this rather than just marking square as `static inline`?

tests/runtime/dwarf
```diff
@@ -45,6 +45,13 @@ REQUIRES_FEATURE dwarf
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
 
+NAME uprobe inlined function
+PROG uprobe:./testprogs/inline_function:square { printf("%d\n", ++@count); }
```
jordalgo, line 49: what about instead printing the function name via `printf("%s\n", usym(reg("ip")));` to confirm it's "square"?

src/dwarf_parser.h
```diff
@@ -70,6 +71,11 @@ class Dwarf {
     return nullptr;
   }
 
+  uint64_t get_prologue_size(const std::string &function __attribute((unused)))
```
jordalgo, line 74: JC, why is this needed?

src/bpftrace.cpp
```diff
@@ -332,6 +332,21 @@ int BPFtrace::add_probe(ast::Probe &p)
             generateWatchpointSetupProbe(func_id, *attach_point, p));
 
         resources.watchpoint_probes.emplace_back(std::move(probe));
+      } else if (probetype(attach_point->provider) == ProbeType::uprobe) {
+        bool locations_from_dwarf = false;
+        if (auto *dwarf = get_dwarf(target)) {
+          auto locations = dwarf->get_function_locations(func_id);
```
jordalgo, line 338: `const auto locations`?

src/dwarf_parser.h
ttreyer: This just need a cleanup from the first version of this PR which only skipped the prologue.

tests/testprogs/CMakeLists.txt
ttreyer: I don't like this either 😕 How can it be better implemented?
If the function is `static inline`, the compiler will be able to always inline it and won't generate an out-of-line function. So the test won't verify that both inlined and regular call are probed.

tests/runtime/dwarf
ttreyer: It is "square" for only one of the probed call, the two inlined call will resolve to "main".
I'll remove the `printf` and let bpftrace print the `@count` value at the end.

tests/testprogs/CMakeLists.txt
jordalgo: > So the test won't verify that both inlined and regular call are probed.

Don't we have other tests that already verify that we can probe non-inlined functions?

tests/runtime/dwarf
jordalgo: > It is "square" for only one of the probed call, the two inlined call will resolve to "main".

Ohh that's interesting. I guess it wouldn't be correct to return `square` for `usym(reg("ip"))` for when square is inlined but when we finally get inline info in our stack traces it should definitely come up as the leaf function for a `ustack`.

tests/testprogs/CMakeLists.txt
ttreyer: Yeah... In that case I'll go with `static inline` and we can avoid this weird exception in the CMake!

ttreyer:
GCC and Clang don't emit code with the same ABI. Clang's stack args are at the `stack pointer + 16`, but GCC stack args are at the `stack pointer + 0`. So I can't make the `stack args` test consistent between the two compilers. Is there an option or a function attribute that might force GCC to set the offset to 16?

We also discussed about `sargX` with @ajor and concluded it typically won't return the correct argument for other languages like Go. We also noticed that `kprobe/kfunc` have a maximum of 6 parameters, which fits in the AMD64 registers and the arguments never need to go on the stack.

In a later PR, we could ensure the builtin returns correct results by relying on the location expression emitted for the debugger.

viktormalik:
> We also discussed about `sargX` with @ajor and concluded it typically won't return the correct argument for other languages like Go.

Yes, Go uses their own [calling conventions](https://tip.golang.org/src/cmd/compile/abi-internal#amd64-architecture) and ignores System V ABI (on x86). I'm not sure what we can do about it other than detecting that a uprobe is inside a Go program and issuing a warning or disabling the`sargX` builtin completely for such a case.

> We also noticed that `kprobe/kfunc` have a maximum of 6 parameters, which fits in the AMD64 registers and the arguments never need to go on the stack. So we were wondering if `sarg` should remain in its current form.

FYI, kernel fprobes (i.e. our kfunc) now support up to 12 arguments so `sargX` could make sense for those (though they can by accessed by name).

tests/runtime/call
```diff
@@ -229,7 +229,7 @@ AFTER ./testprogs/uprobe_loop
 NAME ustack_stack_mode_env_bpftrace
 PROG u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }
 ENV BPFTRACE_STACK_MODE=bpftrace
-EXPECT uprobeFunction1+0
+EXPECT_REGEX uprobeFunction1\+[0-9]+
```
ajor, line 232:
```suggestion
EXPECT_REGEX ^uprobeFunction1\+[0-9]+$
```

Part of the reason `EXPECT_REGEX` is dangerous is that it doesn't enforce a full line match by default. A partial line match here would match other results in addition to what we want, e.g. a stack trace in "perf" format

tests/runtime/call
```diff
@@ -249,7 +249,7 @@ SKIP_IF_ENV_HAS CI=true
 NAME ustack_stack_mode_env_perf
 PROG u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }
 ENV BPFTRACE_STACK_MODE=perf
-EXPECT_REGEX \d+ uprobeFunction1\+0 \(.*/uprobe_loop\)
+EXPECT_REGEX [0-9a-f]+ uprobeFunction1\+[0-9]+ \(.*/uprobe_loop\)
```
ajor, line 252:
```suggestion
EXPECT_REGEX ^[0-9a-f]+ uprobeFunction1\+[0-9]+ \(.*/uprobe_loop\)$
```

tests/runtime/dwarf
```diff
@@ -45,6 +45,30 @@ REQUIRES_FEATURE dwarf
...
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/inline_function
+
+NAME uprobe inlined function with args
+WILL_FAIL
+PROG config = { probe_inline = 1 } uprobe:./testprogs/inline_function:square { print(arg0); print(sarg0); print(args.x); }
+EXPECT_REGEX ERROR: The arg0 builtin can only be used when the probe_inline config is disabled\.
+EXPECT_REGEX ERROR: The sarg0 builtin can only be used when the probe_inline config is disabled\.
+EXPECT_REGEX ERROR: The args builtin can only be used when the probe_inline config is disabled\.
```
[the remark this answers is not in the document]
ajor: Actually, if this test isn't actually reading any DWARF can it be made into a semantic_analyser unit test?

tests/runtime/dwarf
ajor, line 1: Can we add tests for "func" and "ustack" on inlined functions? Even thought they might not be returning the results we expect, it'd be good to have a reference for the future and open issues for them

ajor (commented):
Very nice! Just got a few comments about the tests.

I'm wondering if we can have a more descriptive name for the "probe_inline" config option. We don't want to keep this global setting as a long term solution though (it should be on a per-function basis), so maybe it doesn't matter too much.

Something else we should have a think about is what to do with the `func` builtin for inlined functions. Currently it will return the name of the non-inlined caller. That might be more useful than disabling it altogether though.

`ustack` for inlined functions also misses the inlined function's frame, but that'll be sorted once we have blazesym integrated.

ajor:
> > We also discussed about `sargX` with @ajor and concluded it typically won't return the correct argument for other languages like Go.
> 
> Yes, Go uses their own [calling conventions](https://tip.golang.org/src/cmd/compile/abi-internal#amd64-architecture) and ignores System V ABI (on x86). I'm not sure what we can do about it other than detecting that a uprobe is inside a Go program and issuing a warning or disabling the`sargX` builtin completely for such a case.

To clarify: `sarg` was actually added specifically for Go arguments in #828. It looks like the Go calling convention has changed since 2019 so I'm not sure `sarg` even works correctly for Go any more. I'd be in favour of deprecating it from bpftrace.

I think accessing arguments by name should be the way to go for uprobes now (it wasn't a thing back in 2019). We can't keep `sarg` up to date with changing calling conventions.

It's also possible that `sarg` will get broken for other architectures by this PR (it's update for x86_64 only), but we have no way to test them.

viktormalik:
> > > We also discussed about `sargX` with @ajor and concluded it typically won't return the correct argument for other languages like Go.
> > 
> > 
> > Yes, Go uses their own [calling conventions](https://tip.golang.org/src/cmd/compile/abi-internal#amd64-architecture) and ignores System V ABI (on x86). I'm not sure what we can do about it other than detecting that a uprobe is inside a Go program and issuing a warning or disabling the`sargX` builtin completely for such a case.
> 
> To clarify: `sarg` was actually added specifically for Go arguments in #828. It looks like the Go calling convention has changed since 2019 so I'm not sure `sarg` even works correctly for Go any more. I'd be in favour of deprecating it from bpftrace.
> 
> I think accessing arguments by name should be the way to go for uprobes now (it wasn't a thing back in 2019). We can't keep `sarg` up to date with changing calling conventions.
> 
> It's also possible that `sarg` will get broken for other architectures by this PR (it's updated for x86_64 only), but we have no way to test them.

In fact, the new Go conventions even break the `argX` builtins as those follow standard ABIs for individual architectures, e.g. System V ABI for x86. Since Go completely ignores this, `arg0` will for example return the 4th argument of the Go function (%rdi).

So, deprecating `sargX` won't resolve the above problem. I'm not saying that I'm against it but if we want to properly support tracing Go programs, we need to address the problem with `argX`, too.

As for `sargX`, I've never used them myself, but I can see them useful for tracing userspace functions with more than 6 parameters (for x86 where the 7th and further arguments are passed on stack).

ttreyer:
> I'm wondering if we can have a more descriptive name for the "probe_inline" config option. We don't want to keep this global setting as a long term solution though (it should be on a per-function basis), so maybe it doesn't matter too much.

I'd be happy to rename the config option! Any suggestion?
 
> Something else we should have a think about is what to do with the `func` builtin for inlined functions. Currently it will return the name of the non-inlined caller. That might be more useful than disabling it altogether though.

Is there a builtin returning the probe name, as the user provided it in the script?

tests/runtime/dwarf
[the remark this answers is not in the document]
ttreyer: Does the semantic_analyser unit tests take the `config` into account? I couldn't reproduce the error from there.

tests/runtime/dwarf
ttreyer: I've added tests for "func", "ustack" and "probe" in the dwarf runtime test.

ajor:
Looking good to me. I think the only remaining piece is to decide about `sarg`.

> In fact, the new Go conventions even break the `argX` builtins as those follow standard ABIs for individual architectures, e.g. System V ABI for x86. Since Go completely ignores this, `arg0` will for example return the 4th argument of the Go function (%rdi).
> 
> So, deprecating `sargX` won't resolve the above problem. I'm not saying that I'm against it but if we want to properly support tracing Go programs, we need to address the problem with `argX`, too.

Right, since Go's ABI changed we don't have a reliable way of accessing arguments with or without `sarg`. Since `sarg` was added in an attempt to support Go, I think this is an argument in favour of removing it.

> As for `sargX`, I've never used them myself, but I can see them useful for tracing userspace functions with more than 6 parameters (for x86 where the 7th and further arguments are passed on stack).

I don't like that this forces users to know the calling convention of whatever architecture they're running on. I'd say that improving our DWARF support and pushing everyone towards using named argument access (`args.foo`) is a better way to go. We'll hopefully get some limited support for DWARF locations expressions soon, which will mean that this will reliably work for accessing stack arguments. It'll work for Go arguments too, no matter how they change their calling convention in the future.

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to
   - [#3060](https://github.com/bpftrace/bpftrace/pull/3060)
 - Disable func builtin for kretprobes and uretprobes when `get_func_ip` feature is not available
   - [#2645](https://github.com/bpftrace/bpftrace/pull/2645)
+- Attache probe after function prologue and probe inlined functions
```
viktormalik, line 37: Nit: typo
```suggestion
- Attach probe after function prologue and probe inlined functions
```

tests/runtime/dwarf
```diff
@@ -45,6 +45,63 @@ REQUIRES_FEATURE dwarf
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
 
+# We check the expected output after the timeout.
```
viktormalik, line 48: What if we instead checked the number of attached probes?
```
EXPECT Attaching X probes...
```

viktormalik (commented):
Looks great! Just one suggestion for the tests.

viktormalik:
> I don't like that this forces users to know the calling convention of whatever architecture they're running on. I'd say that improving our DWARF support and pushing everyone towards using named argument access (`args.foo`) is a better way to go. We'll hopefully get some limited support for DWARF locations expressions soon, which will mean that this will reliably work for accessing stack arguments. It'll work for Go arguments too, no matter how they change their calling convention in the future.

I agree that DWARF should be the way to go. But, if the user doesn't have DWARF symbols for his binary, knows the ABI, and wants to see values of stack-passed args, why not let him do it, especially since we already have the feature implementated?

In addition, `sargX` should work for kprobes, too, right?

ajor:
> But, if the user doesn't have DWARF symbols for his binary, knows the ABI, and wants to see values of stack-passed args, why not let him do it, especially since we already have the feature implementated?

Because I think it's going to get broken by this PR and we have no way of testing it across architectures 😅

I'm not sure how we can easily ensure that it works in every situation, even on x86_64, now that we don't necessarily attach at the beginning of a function. I'd rather disallow a feature entirely than have it return wrong results.

And even if it did work correctly, I don't think it's a particularly useful feature because of how much knowledge of the application's ABI it requires in order to use. Extending `arg` to read stack arguments after the register arguments might be more useful than `sarg`.

tests/runtime/dwarf
ajor: Does this work with uprobe_multi? (I haven't got a kernel with that feature to test with)

viktormalik:
> > But, if the user doesn't have DWARF symbols for his binary, knows the ABI, and wants to see values of stack-passed args, why not let him do it, especially since we already have the feature implementated?
> 
> Because I think it's going to get broken by this PR and we have no way of testing it across architectures 😅

Ah, I somehow missed that. In that case, I'm all in for dropping the feature :-) Sorry for the confusion and thanks for clearing it up!

> And even if it did work correctly, I don't think it's a particularly useful feature because of how much knowledge of the application's ABI it requires in order to use (it was originally intended for Golang were `sarg0` corresponded to the first argument). Extending `arg` to read stack arguments after the register arguments might be more useful than `sarg`.

That sounds like a reasonable approach. Maybe we could open an issue to track that.

tests/runtime/dwarf
viktormalik: Good point, it doesn't (uprobe_multi will attach just one probe) but we could use `--no-feature=uprobe_multi` to disable it for this test.

ttreyer: Moreover, does checking the number of attached probes also checks we're able to intercept their call?

viktormalik: Maybe not but there are other tests in this suite to verify that.

src/attached_probe.cpp
```diff
@@ -379,7 +379,10 @@ static int sym_address_cb(const char *symname,
 {
   struct symbol *sym = static_cast<struct symbol *>(p);
 
-  if (sym->address >= start && sym->address < (start + size)) {
+  // When size is 0, then [start, start + size) = [start, start) = ø.
```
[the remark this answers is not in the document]
ttreyer: This can happen when the symbol has no size or an unknown size.
I'm not sure a 0-length symbol is created, but our `tests/runtime/probe - uprobe_zero_size` uses the `_init` symbol to validate bpftrace's behaviour. By default, it won't let you probe such symbol, but you can pass `--unsafe` to force it.

ttreyer:
> Ah, I somehow missed that. In that case, I'm all in for dropping the feature :-) Sorry for the confusion and thanks for clearing it up!

Following that, I marked `sarg` as deprecated and removed its tests, since this PR breaks them by skipping the function prologue.

tests/runtime/builtin
danobi, line 1: Maybe it'd be good to leave at least one or two of these in the new form to document/check that stack accesses don't regress?

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to
   - [#3060](https://github.com/bpftrace/bpftrace/pull/3060)
 - Disable func builtin for kretprobes and uretprobes when `get_func_ip` feature is not available
   - [#2645](https://github.com/bpftrace/bpftrace/pull/2645)
+- Attach probe after function prologue and probe inlined functions
```
danobi, line 37: Perhaps better to have 2 changelog entries (both pointing to same PR).

These are big features (user PoV)!

tests/runtime/dwarf
```diff
@@ -45,6 +45,63 @@ REQUIRES_FEATURE dwarf
...
+     uprobe:./testprogs/inline_function:square { @count++ }
+EXPECT @count: 1
+REQUIRES_FEATURE dwarf
+TIMEOUT 5
+AFTER ./testprogs/inline_function
+
+NAME uprobe inlined function
+PROG config = { probe_inline = 1 }
+     uprobe:./testprogs/inline_function:square { @count++ }
+EXPECT @count: 3
```
danobi, line 62: Very cool!

danobi (commented):
lgtm!

ajor (approved):
Just thought of one more changelog entry we should include: "Deprecated sarg".

Other than that, looks great and I think it's ready to merge!

tests/runtime/dwarf
viktormalik: Reminding this

viktormalik (approved):
I'd still prefer if we could avoid waiting the entire timeout for the few runtime tests but it's not a hard blocker for me.

LGTM otherwise, nice work!

tests/runtime/dwarf
ajor: How about this alternative?

Just add this line to the end of the script:
```
interval:ms:10 { if (@timeout) { exit(); } @timeout = 0; if (@count == 1) { @timeout = 1; } }
```

It'll wait 10-20ms after hitting the expected number for `@count` and then exit. That should be plenty of time for any extra function calls to execute if they're going to, and the milliseconds of test overhead will just be lost in the noise.

tests/runtime/dwarf
```diff
@@ -45,6 +45,62 @@ REQUIRES_FEATURE dwarf
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
 
+NAME uprobe skip inlined function
+PROG config = { probe_inline = 0 }
+     uprobe:./testprogs/inline_function:square { @count++ }
+     uretprobe:./testprogs/inline_function:main { exit() }
```
ttreyer, line 51: This should solve the timeout issue 😄

viktormalik: Elegant!

viktormalik (approved):

viktormalik:
@danobi @jordalgo do you want to have one more look? Otherwise we can merge.

tests/runtime/dwarf
```diff
@@ -45,6 +45,62 @@ REQUIRES_FEATURE dwarf
...
+EXPECT @count: 1
```
jordalgo, line 52: I guess bpftrace initializes this map value to 1 even though it never gets called?

viktormalik: No, it should be initialized to 0. The test checks that only one (non-inlined) instance of `square` is hit (see the `inline_function.c` source).

tests/semantic_analyser.cpp
```diff
@@ -254,6 +262,33 @@ kprobe:f { fake }
...
+  configs.set(ConfigKeyBool::probe_inline, true);
+
+  // Check argument builtins are rejected when `probe_inline` is enabled.
+  test_error(*bpftrace, "uprobe:/bin/sh:f { arg0 }", R"(
+stdin:1:20-24: ERROR: The arg0 builtin can only be used when the probe_inline config is disabled.
+uprobe:/bin/sh:f { arg0 }
+                   ~~~~
+)");
+  test_error(*bpftrace, "uprobe:/bin/sh:f { sarg0 }", R"(
+stdin:1:20-25: ERROR: The sarg0 builtin can only be used when the probe_inline config is disabled.
```
jordalgo, line 278: So will this test go away when we fully kill `sarg`?

ttreyer: Yes.

tests/testprogs/inline_function.c
```diff
@@ -0,0 +1,27 @@
...
+ */
+__attribute__((always_inline, used)) static inline int square(int x)
+{
+  volatile int res = x * x;
+  return res;
+}
+
+int main(int argc, char *argv[] __attribute__((unused)))
+{
+  printf("%d^2 = %d\n", argc, square(argc));
```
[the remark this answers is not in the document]
ttreyer: Probably not, but it prevent the compiler from completely optimising out the calls to `square` if their results are not used.

jordalgo (approved):
LGTM!

ajor:
Let's go! 🚀

PR #3100, opened 2024-04-04
Docs: Refer to "threads" and "processes" instead of "tasks"

ajor:


man/adoc/bpftrace.adoc
```diff
@@ -1021,8 +1021,8 @@ For string arguments use the `str()` call to retrieve the value.
 | comm
 | string[16]
 | 4.2
-| get_current_com
-| `comm` of the current task. Equal to the value in `/proc/<pid>/comm`
+| get_current_comm
+| `comm` of the current task. Equal to the value in `/proc/<tid>/comm`
```
jordalgo, line 1025: Why don't we just say thread name due to the confusion that pid in kernel space is a tid in user space.

danobi: Thread name makes sense to me

ajor:
Updated the docs more broadly to refer to "threads" and "processes" instead of "tasks" as they'll be more familiar terms for users who aren't kernel developers

jordalgo (approved):
Nice cleanup.

Nit: Periods don't belong at the end of most list items. These aren't sentences: "Random number" or "Name of the current probe".

ajor:
I was just going for consistency with full stops as some entries had them and some didn't.

Would you say they should be removed on some of the existing lines that have them? e.g. "Thread ID of the current task.", "	
Total amount of positional parameters passed."

Also some of the entries with multiple sentences aren't really full sentences either: "Jiffies of the kernel. [...]"

viktormalik (approved):
Nice cleanup!

I'll leave the full stop discussion to you, native speakers. I'm ok with both keeping it everywhere for consistency or omitting it in single, verb-free, items.

jordalgo:
Here is [an article on this topic](https://lawprofessors.typepad.com/legal_skills/2012/01/how-to-punctuate-bullet-points-.html#:~:text=Use%20a%20period%20(full%20stop,all%20fragments%2C%20not%20a%20mixture). 

We definitely have a mixture of fragments and sentences; though more fragments.  I would say delete periods in most cases and for something like this: 

> nth argument passed to the function being traced. These are extracted from the CPU registers. The amount of args passed in registers depends on the CPU architecture. (kprobes, uprobes, usdt).

We could help it seem more correct by making the fragment a sentence:

> This is the nth argument passed to the function being traced. These are extracted from the CPU registers. The amount of args passed in registers depends on the CPU architecture. This applies to kprobes, uprobes, and usdts.

ajor:
I took the easy route out of just removing full stops from the single-fragment lines and leaving the rest for now

jordalgo (approved):

Issue #2918, opened 2024-01-13
Proof of concept for AOT


danobi:
This is a tracking issue to break down what the AOT PoC still needs:

- [ ] Shim-side (IOW runtime-side) logic to look for bytecode
- [ ] Re-remove massive list of runtime deps from shim
    * I recall fixing this originally, but they probably crept back
    * Need to find a way to prevent backsliding. Probably some kind of CI rule or cmake step
 - [ ] Make `runtime-tests.sh --run-aot-tests` completely pass
     * This will require some relatively mechanical work in `portability_analyser.cpp`
 - [ ] Document AOT
     * Add CLI flags to `--help`
     * Manpage + reference guide

danobi:
@jordalgo if you're interested, this project is something we could fairly trivially parallelize. AOT has been a long requested feature -- many folks have told me they don't use bpftrace in prod b/c they don't want to ship a compiler into production (security reasons, etc). Obviously no pressure :) . This is something I plan to chip away at for at least a few years.

jordalgo:
@danobi Sure, would be happy to help. Let me know if one of those items above would be good to start with or if you want to hop on a VC to go over some of these things this week.

danobi:
Awesome! I'm a bit tight on time this week (in bay, actually) but next week should be good. I think after the first item is done, all further work is parallelizable. Here's the original design doc: https://dxuuu.xyz/aot-bpftrace.html . It's a little outdated now but the general idea hasn't changed

chengshuyi:
> @jordalgo if you're interested, this project is something we could fairly trivially parallelize. AOT has been a long requested feature -- many folks have told me they don't use bpftrace in prod b/c they don't want to ship a compiler into production (security reasons, etc). Obviously no pressure :) . This is something I plan to chip away at for at least a few years.

Hi, i wonder if you have considered a method like systemtap to generate c code, and then compile the c code to generate binary?

Thanks.

viktormalik:
> Hi, i wonder if you have considered a method like systemtap to generate c code, and then compile the c code to generate binary?

Hi @chengshuyi, I'm not sure how that would work but SystemTap requires kernel debuginfo so we wouldn't get much advantage in terms of dependencies there.

danobi:
[...]

PR #3104, opened 2024-04-06
aot: Enable aot built bpftrace programs to run on their own

arthurshau:
The first commit fixes a couple issues with aot built binaries that was preventing runtime from functioning.

We now look in the right place in the ELF file for the custom .btaot section that contains the metadata + generated ELF, that we then restore to create BpfBytecode. ~~Second, we now also serialize the map of BpfMaps owned by BpfBytecode into the bytecode as well, which is necessary for runtime. In order to do this, I needed to add a default constructor to BpfMap, as well as a hacky solution to be able to serialize the name of the map which is a cstring_view (which may or may not have defeated the purpose of having the cstring_view in the first place - would appreciate a review of this).~~ Finally, I fixed maps not being printed out correctly when bpftrace exits.

The second commit allows running aot binaries directly without calling having to call runtime shim (e.g. you can now just call ```./a.btaot``` instead of ```bpftrace-aotrt a.btaot```. This is possible because the aot binaries already contain all the info necessary to run themselves.

Address part 1 of #2918.

Example:
```
$ ./bpftrace -e 'profile:hz:599 { @[tid] = count(); exit(); }' --aot a.btaot
$ ./a.btaot
Attaching 1 probe...


@[679916]: 1
@[12686]: 1
@[680061]: 1
@[0]: 4
```

##### Checklist

### DIFF
--- a/src/aot/aot.cpp
+++ b/src/aot/aot.cpp
@@ -6,10 +6,9 @@
 #include <exception>
 #include <fcntl.h>
 #include <fstream>
-#include <istream>
+#include <gelf.h>
+#include <libelf.h>
 #include <optional>
-#include <ostream>
-#include <streambuf>
 #include <sys/mman.h>
 #include <sys/stat.h>
 
@@ -19,6 +18,7 @@
 #include <cereal/types/unordered_map.hpp>
 #include <cereal/types/vector.hpp>
 
+#include "ast/elf_parser.h"
 #include "filesystem.h"
 #include "log.h"
 #include "utils.h"
@@ -41,8 +41,8 @@ struct Header {
   uint64_t version;    // Hash of version string
   uint64_t rr_off;     // RequiredResources offset from start of file
   uint64_t rr_len;     // RequiredResources length
-  uint64_t bc_off;     // Bytecode offset from start of file
-  uint64_t bc_len;     // Bytecode length
+  uint64_t elf_off;    // ELF offset from start of file
+  uint64_t elf_len;    // ELF length
 };
 
 static_assert(sizeof(Header) == 48);
@@ -66,12 +66,6 @@ uint32_t rs_hash(std::string_view str)
   return hash;
 }
 
-void serialize_bytecode(const BpfBytecode &bytecode, std::ostream &out)
-{
-  cereal::BinaryOutputArchive archive(out);
-  archive(bytecode);
-}
-
 int load_required_resources(BPFtrace &bpftrace, uint8_t *ptr, size_t len)
 {
   try {
@@ -84,24 +78,10 @@ int load_required_resources(BPFtrace &bpftrace, uint8_t *ptr, size_t len)
   return 0;
 }
 
-int load_bytecode(BPFtrace &bpftrace, uint8_t *ptr, size_t len)
-{
-  try {
-    Membuf mbuf(ptr, ptr + len);
-    std::istream istream(&mbuf);
-    cereal::BinaryInputArchive archive(istream);
-    archive(bpftrace.bytecode_);
-  } catch (const std::exception &ex) {
-    LOG(ERROR) << "Failed to deserialize metadata: " << ex.what();
-    return 1;
-  }
-
-  return 0;
-}
-
 std::optional<std::vector<uint8_t>> generate_btaot_section(
     const RequiredResources &resources,
-    const BpfBytecode &bytecode)
+    void *const elf,
+    size_t elf_size)
 {
   // Serialize RuntimeResources
   std::string serialized_metadata;
@@ -114,17 +94,6 @@ std::optional<std::vector<uint8_t>> generate_btaot_section(
     return std::nullopt;
   }
 
-  // Serialize bytecode
-  std::string serialized_bytecode;
-  try {
-    std::ostringstream serialized(std::ios::binary);
-    serialize_bytecode(bytecode, serialized);
-    serialized_bytecode = serialized.str();
-  } catch (const std::exception &ex) {
-    LOG(ERROR) << "Failed to serialize bytecode: " << ex.what();
-    return std::nullopt;
-  }
-
   // Construct the header
   auto hdr_len = sizeof(Header);
   Header hdr = {
@@ -134,13 +103,13 @@ std::optional<std::vector<uint8_t>> generate_btaot_section(
     .version = rs_hash(BPFTRACE_VERSION),
     .rr_off = hdr_len,
     .rr_len = serialized_metadata.size(),
-    .bc_off = hdr_len + serialized_metadata.size(),
-    .bc_len = serialized_bytecode.size(),
+    .elf_off = hdr_len + serialized_metadata.size(),
+    .elf_len = elf_size,
   };
 
   // Resize the output buffer appropriately
   std::vector<uint8_t> out;
-  out.resize(sizeof(Header) + hdr.rr_len + hdr.bc_len);
+  out.resize(sizeof(Header) + hdr.rr_len + hdr.elf_len);
   uint8_t *p = out.data();
 
   // Write out header
@@ -151,9 +120,9 @@ std::optional<std::vector<uint8_t>> generate_btaot_section(
   memcpy(p, serialized_metadata.data(), hdr.rr_len);
   p += hdr.rr_len;
 
-  // Write out bytecode
-  memcpy(p, serialized_bytecode.data(), hdr.bc_len);
-  p += hdr.bc_len;
+  // Write out ELF
+  memcpy(p, elf, hdr.elf_len);
+  p += hdr.elf_len;
 
   return out;
 }
@@ -213,10 +182,11 @@ out:
 } // namespace
 
 int generate(const RequiredResources &resources,
-             const BpfBytecode &bytecode,
-             const std::string &out)
+             const std::string &out,
+             void *const elf,
+             size_t elf_size)
 {
-  auto section = generate_btaot_section(resources, bytecode);
+  auto section = generate_btaot_section(resources, elf, elf_size);
   if (!section)
     return 1;
 
@@ -253,16 +223,76 @@ int load(BPFtrace &bpftrace, const std::string &in)
     return 1;
   }
 
-  uint8_t *ptr = static_cast<uint8_t *>(
-      ::mmap(NULL, in_file_size, PROT_READ, MAP_PRIVATE, infd, 0));
-  if (ptr == MAP_FAILED) {
-    auto saved_err = errno;
-    LOG(ERROR) << "Failed to mmap: " << in << ": " << std::strerror(saved_err);
-    return 1;
+  // Find .btaot section
+  Elf *elf = NULL;
+  Elf_Scn *scn = NULL;
+  GElf_Shdr shdr;
+  char *secname = NULL;
+  Elf_Data *data = NULL;
+  uint8_t *btaot_section = NULL;
+  const Header *hdr;
+
+  if (elf_version(EV_CURRENT) == EV_NONE) {
+    LOG(ERROR) << "Cannot set libelf version: " << elf_errmsg(-1);
+    err = 1;
+    goto out;
   }
 
-  // Validate header
-  auto hdr = reinterpret_cast<const Header *>(ptr);
+  elf = elf_begin(infd, ELF_C_READ, NULL);
+  if (!elf) {
+    LOG(ERROR) << "Cannot read ELF file: " << elf_errmsg(-1);
+    err = 1;
+    goto out;
+  }
+
+  size_t strndx;
+  if (elf_getshdrstrndx(elf, &strndx) < 0) {
+    LOG(ERROR) << "Failed to get ELF section index of the string table: "
+               << elf_errmsg(-1);
+    err = 1;
+    goto out;
+  }
+
+  int i;
+  i = 0;
+  while ((scn = elf_nextscn(elf, scn))) {
+    i++;
+    if (!gelf_getshdr(scn, &shdr)) {
+      LOG(ERROR) << "Failed to get ELF section(" << i
+                 << ") hdr: " << elf_errmsg(-1);
+      err = 1;
+      goto out;
+    }
+
+    secname = elf_strptr(elf, strndx, shdr.sh_name);
+    if (!secname) {
+      LOG(ERROR) << "Failed to get ELF section(" << i
+                 << ") hdr name: " << elf_errmsg(-1);
+      err = 1;
+      goto out;
+    }
+
+    if (std::string_view(secname) == AOT_ELF_SECTION) {
+      data = elf_getdata(scn, 0);
+      if (!data) {
+        LOG(ERROR) << "Failed to get BTAOT ELF section(" << i
+                   << ") data: " << elf_errmsg(-1);
+        err = 1;
+        goto out;
+      }
+
+      btaot_section = static_cast<uint8_t *>(data->d_buf);
+      break;
+    }
+  }
+
+  // Validate .btaot header
+  hdr = reinterpret_cast<const Header *>(btaot_section);
+  if (!hdr) {
+    LOG(ERROR) << "Couldn't find " << AOT_ELF_SECTION << " section in " << in;
+    err = 1;
+    goto out;
+  }
   if (hdr->magic != AOT_MAGIC) {
     LOG(ERROR) << "Invalid magic in " << in << ": " << hdr->magic;
     err = 1;
@@ -285,28 +315,30 @@ int load(BPFtrace &bpftrace, const std::string &in)
     goto out;
   }
   if ((hdr->rr_off + hdr->rr_len) > static_cast<uint64_t>(in_file_size) ||
-      (hdr->bc_off + hdr->bc_len) > static_cast<uint64_t>(in_file_size)) {
+      (hdr->elf_off + hdr->elf_len) > static_cast<uint64_t>(in_file_size)) {
     LOG(ERROR) << "Corrupted AOT bpftrace file: incomplete payload";
     err = 1;
     goto out;
   }
 
   // Load payloads
-  err = load_required_resources(bpftrace, ptr + hdr->rr_off, hdr->rr_len);
+  err = load_required_resources(bpftrace,
+                                btaot_section + hdr->rr_off,
+                                hdr->rr_len);
   if (err)
     goto out;
 
-  err = load_bytecode(bpftrace, ptr + hdr->bc_off, hdr->bc_len);
+  bpftrace.bytecode_ = elf::parseBpfBytecodeFromElfObject(btaot_section +
+                                                              hdr->elf_off,
+                                                          hdr->elf_len);
   if (err)
     goto out;
 
 out:
-  if (::munmap(ptr, in_file_size)) {
-    auto saved_err = errno;
-    LOG(ERROR) << "Failed to munmap(): " << in << ": "
-               << std::strerror(saved_err);
-  }
+  if (elf)
+    elf_end(elf);
 
+  close(infd);
   return err;
 }
 
--- a/src/aot/aot.cpp
+++ b/src/aot/aot.cpp
@@ -26,7 +26,6 @@
 
 #define AOT_ELF_SECTION ".btaot"
 static constexpr auto AOT_MAGIC = 0xA07;
-static constexpr auto AOT_SHIM_NAME = "bpftrace-aotrt";
 static constexpr auto AOT_SECDATA_TEMPFILE = ".temp_btaot";
 
 // AOT payload will have this header at the beginning. We don't worry about
@@ -190,7 +189,7 @@ int generate(const RequiredResources &resources,
   if (!section)
     return 1;
 
-  auto shim = find_in_path(AOT_SHIM_NAME);
+  auto shim = find_in_path(AOT_SHIM_NAME.data());
   if (!shim) {
     LOG(ERROR) << "Failed to locate " << AOT_SHIM_NAME
                << " shim binary. Is it in $PATH?";
--- a/src/aot/aot.h
+++ b/src/aot/aot.h
@@ -9,6 +9,8 @@
 namespace bpftrace {
 namespace aot {
 
+static constexpr std::string_view AOT_SHIM_NAME = "bpftrace-aotrt";
+
 int generate(const RequiredResources &resources,
              const std::string &out,
              void *const elf,
--- a/src/aot/aot.h
+++ b/src/aot/aot.h
@@ -10,8 +10,9 @@ namespace bpftrace {
 namespace aot {
 
 int generate(const RequiredResources &resources,
-             const BpfBytecode &bytecode,
-             const std::string &out);
+             const std::string &out,
+             void *const elf,
+             size_t elf_size);
 
 int load(BPFtrace &bpftrace, const std::string &in);
 
--- a/src/aot/aot_main.cpp
+++ b/src/aot/aot_main.cpp
@@ -1,12 +1,12 @@
 #include <fstream>
 #include <getopt.h>
 #include <iostream>
-#include <memory>
 
 #include "aot.h"
 #include "bpftrace.h"
 #include "log.h"
 #include "output.h"
+#include "run_bpftrace.h"
 #include "version.h"
 
 using namespace bpftrace;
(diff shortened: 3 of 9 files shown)


src/aot/aot.cpp
```diff
@@ -253,16 +253,40 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
 
-  // Validate header
-  auto hdr = reinterpret_cast<const Header *>(ptr);
+  // Find .btaot section
+  auto *elf_header = reinterpret_cast<Elf64_Ehdr *>(ptr);
+  auto string_table_header = reinterpret_cast<Elf64_Shdr *>(
+      ptr + elf_header->e_shoff +
+      elf_header->e_shstrndx * elf_header->e_shentsize);
+  auto section_header = reinterpret_cast<Elf64_Shdr *>(ptr +
+                                                       elf_header->e_shoff);
```
danobi, line 270: I think there should probably be some error checking here. Long term, it'd be nice to not have a dependency against libelf. So that the final AOT binaries are lighter. But for now, probably best to use libelf APIs.

Here's a lazy link you can refer to for example:  https://lore.kernel.org/bpf/82ae2ddd83a8d85ae071e650cf80005c36bb1343.1711389163.git.dxu@dxuuu.xyz/   (see btf_encoder__tag_kfuncs())

src/aot/aot.cpp
```diff
@@ -253,16 +253,40 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
+  uint8_t *btaot_section = NULL;
+  for (int i = 0; i < elf_header->e_shnum; i++) {
+    auto *name = reinterpret_cast<char *>(ptr + string_table_header->sh_offset +
+                                          section_header->sh_name);
+    if (strcmp(name, ".btaot") == 0) {
```
danobi, line 275: `".btaot"` -> `AOT_ELF_SECTION`

src/bpfmap.h
```diff
@@ -56,9 +62,17 @@ class BpfMap {
 private:
   libbpf::bpf_map_type type_;
   cstring_view name_;
+  std::string tempname_;
```
danobi, line 65: Ah, I see. Rather than this, probably better to just change `name_` from being a `cstring_view` to just regular `std::string`. If copying has a functional purpose, I think it's ok.

cc @ajor ; any thoughts? Looks like BpfMap happens to be the only user of cstring_view

src/bpfmap.h
```diff
@@ -17,9 +18,13 @@ namespace bpftrace {
 
 class BpfMap {
 public:
+  BpfMap()
+  {
+  }
```
danobi, line 23: Possible to make this non-public? I suppose it depends on what cereal allows, but would be good to make clear this isn't for public consumption

src/aot/aot_main.cpp
```diff
@@ -95,7 +95,14 @@ int main(int argc, char* argv[])
...
-  if (!argv[optind]) {
+  if (argv[optind]) {
+    usage();
+    return 1;
+  }
+
+  if (strcmp(argv[0], aot::AOT_SHIM_NAME) == 0) {
+    LOG(INFO) << "Runtime shim should not be run directly, please generate a "
+                 "binary using --aot option in bpftrace";
     usage();
```
danobi, line 106: Printing `usage()` here seems confusing. This particular error is not an issue with cli flags.

src/aot/aot_main.cpp
```diff
@@ -13,7 +13,7 @@ using namespace bpftrace;
 void usage()
 {
   // clang-format off
-  std::cerr << "USAGE: bpftrace-aotrt filename" << std::endl;
+  std::cerr << "USAGE: filename [options]" << std::endl;
```
danobi, line 16: literal `"filename"` should be `argv[0]` instead right?

danobi (commented):
overall looks reasonable. i'm assuming this involved some nice debugging work :)

src/container/cstring_view.h
```diff
@@ -19,6 +19,10 @@ namespace bpftrace {
  */
 class cstring_view : public std::string_view {
 public:
+  constexpr cstring_view() noexcept
```
ajor, line 22: What's this needed for? It breaks `cstring_view`'s invariant of always pointing to a null-terminated string so we should try to find another way to solve the issue

src/bpfmap.h
arthurshau: So I spent some time trying to get this to work, but unfortunately it just didn't work out. Serialization itself wasn't an issue - it's simply a matter of manually serializing the underlying string that the ```cstring_view``` points to, which we can do since we know its size. For deserialization, although we wouldn't be able to go directly back to a ```cstring_view``` due to its properties as a view, it was possible to deserialize the data back into a regular string. And then from there you can construct a ```cstring_view``` using that string.

The problem lies with the ownership of the string's memory. With this approach, it means that when you deserialize ```BpfMap```, you need to create a local string variable to store the deserialized name. Then, you can construct a ```cstring```_view manually from that and then finally construct a new ```BpfMap ```object. But once the deserialization function exits and the local string goes out of scope, using the ```cstring_view``` becomse undefined behavior.

All in all, I think it will be difficult to avoid the need to store a copy of the string in this case. So for now, I think I'm just going to convert the ```name_``` back to a regular ```std::string``` rather than a ```cstring_view.```.  This also saves us the need to create a default constructor for ```cstring_view``` (which cereal requires), which breaks the ```cstring_view ```invariant as @ajor noted.

src/container/cstring_view.h
arthurshau: Removed - see above

src/bpfmap.h
arthurshau: Normally a private constructor would work, since we already have ```friend class cereal::access``` there, if we were serializing just ```BpfMap``` itself. But since we're serializing the ```maps_``` member of ```BpfBytecode```, which is a ```std::map<std::string, BpfMap>```, it apparently doesn't work. I think there's an explanation as to why cereal can't do this, but I didn't look super hard into it. So it might need to stay public 😔.

arthurshau:
Addressed comments, thanks!

src/aot/aot.cpp
```diff
@@ -26,9 +27,11 @@
 
 #define AOT_ELF_SECTION ".btaot"
 static constexpr auto AOT_MAGIC = 0xA07;
-static constexpr auto AOT_SHIM_NAME = "bpftrace-aotrt";
 static constexpr auto AOT_SECDATA_TEMPFILE = ".temp_btaot";
 
+#define elf_error(fmt, ...)                                                    \
+  fprintf(stderr, "%s: " fmt ": %s.\n", __func__, ##__VA_ARGS__, elf_errmsg(-1))
```
jordalgo, line 33: Curious, what is this doing? Why don't we want to use our current logging framework? Also shouldn't macros be all CAPS?

src/aot/aot.cpp
```diff
@@ -253,16 +256,62 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
+  }
+
+  size_t strndx;
+  if (elf_getshdrstrndx(elf, &strndx) < 0)
+    goto out;
+
+  while ((scn = elf_nextscn(elf, scn))) {
+    i++;
+    if (!gelf_getshdr(scn, &shdr)) {
+      elf_error("Failed to get ELF section(%d) hdr", i);
```
jordalgo, line 287: should we log the section name here instead of the index?

src/aot/aot.cpp
```diff
@@ -253,16 +256,62 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
+      btaot_section = static_cast<uint8_t *>(data->d_buf);
+    }
   }
 
-  // Validate header
-  auto hdr = reinterpret_cast<const Header *>(ptr);
+  // Validate .btaot header
+  hdr = reinterpret_cast<const Header *>(btaot_section);
+  if (!hdr) {
+    LOG(ERROR) << "Couldn't find .btaot section in " << in;
```
jordalgo, line 311:
```suggestion
    LOG(ERROR) << "Couldn't find << AOT_ELF_SECTION << section in " << in;
```

src/aot/aot_main.cpp
```diff
@@ -106,18 +111,34 @@ int main(int argc, char* argv[])
...
 
   BPFtrace bpftrace(std::move(output));
-  int err = aot::load(bpftrace, argv[optind]);
+  int err = aot::load(bpftrace, argv[0]);
   if (err) {
     LOG(ERROR) << "Failed to load AOT script";
     return err;
   }
 
+  // Signal handler that lets us know an exit signal was received.
```
jordalgo, line 120: Why do we need all these signal handlers?

src/aot/aot.cpp
arthurshau: Whoops, I copied this over from Daniel's libelf example just to get it to work, but forgot to clean it up. Will switch to LOG(ERROR); thanks for the catch.

src/aot/aot.cpp
```diff
@@ -253,16 +256,62 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
-    return 1;
+  // Find .btaot section
+  Elf *elf = NULL;
+  Elf_Scn *scn = NULL;
+  GElf_Shdr shdr;
+  char *secname = NULL;
+  Elf_Data *data = NULL;
+  uint8_t *btaot_section = NULL;
+  const Header *hdr;
+  int i = 0;
```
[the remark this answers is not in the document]
arthurshau: Yeah, I put it here (along with the others) because of the goto statements bypassing initialization errors. But I can split the initialization onto a second line to make it work.

src/aot/aot_main.cpp
arthurshau: I'm actually not too sure - I copied this code from the end of main() in main.cpp. I figured aot will need to duplicate most of the behavior surrounding these signals since we want it to behave basically the same as regular bpftrace. But I can remove them if they seem unnecessary.

arthurshau:
Address comments

src/bpfmap.h
ajor: @arthurshau I'm not familar with cereal at all - do you have an example of the kind of deserialisation function you're talking about?

ajor: Ah nevermind, I think I understand now. Makes sense that string_view won't work here 😢

src/bpfmap.h
```diff
@@ -86,9 +98,8 @@ inline std::string bpf_map_name(std::string_view bpftrace_map_name)
   return name;
 }
 
-inline std::string bpftrace_map_name(std::string_view bpf_map_name)
+inline std::string bpftrace_map_name(std::string name)
```
ajor, line 101: Let's leave the parameter as a std::string_view.

A `std::string` parameter results in two string copies being made - one to create the argument and a second to return the result. Making the copy explicitly in the function body means the compiler can take advantage of [named return value optimisation](https://en.cppreference.com/w/cpp/language/copy_elision), so there's no second copy to return the result.

A `const std::string &` parameter would also achieve the same result, but I see now that there's a third string copy that I didn't spot and eliminate the last time I worked on this function. `string::substr` makes a copy, but `string_view::substr` doesn't. The code isn't making use of this fact yet, but it could be done in the future.

Going over all of this, I can see how we ended up with so many string copies throughout bpftrace - it's so easy to accidentally copy a string!

ajor (commented):
Agreed that we can't store a cstring_view in the serialised data now - just got a couple of other places that we can still reduce string copying in

src/aot/aot_main.cpp
arthurshau: Sure - created run_bpftrace.h/cpp to handle this.

src/bpfmap.h
arthurshau: Agreed - it's something I didn't really consider before. Thanks for pointing this out!

src/run_bpftrace.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "bpftrace.h"
+
+using namespace bpftrace;
```
ajor, line 5: From the Cpp Core Guidelines: ["Don’t write using namespace at global scope in a header file"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file)

We've either got to fully qualify everything we use (`bpftrace::BPFtrace`) or maybe it'd make sense to just put this whole file in the `bpftrace` namespace as well?

src/run_bpftrace.cpp
ajor, line 1: I think we'll need to move the `#include <csignal>` line from main.cpp over to this file as well. (it's not needed there AFAICT)

ajor (commented):
Looking pretty good to me, just a couple of minor points.

What's the testing situation looking like for AOT? It'd be good to have it exercised by the runtime test suite to ensure it doesn't break again

arthurshau:
Address comments and rebase.

> What's the testing situation looking like for AOT? It'd be good to have it exercised by the runtime test suite to ensure it doesn't break again

The immediate next order of business to enable this PoC will be to get ```runtime-tests.sh --run-aot-tests``` to completely pass. As noted in #2918, this may require some work in ```portability_analyser.cpp``` to determine which features are/are not supported for AOT at the moment.

ajor (approved):
Nice!

src/aot/aot.cpp
```diff
@@ -253,16 +253,66 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
   }
 
-  // Validate header
-  auto hdr = reinterpret_cast<const Header *>(ptr);
+  size_t strndx;
+  if (elf_getshdrstrndx(elf, &strndx) < 0)
+    goto out;
+
+  int i;
+  i = 0;
```
jordalgo, line 281: super nit
```suggestion
  int i = 0;
```

jordalgo:
Maybe also for a follow-up diff but the generated binary still seems rather large
```
[parallels@fedora-linux-38 build]$ ls -la | grep "a.btaot"
-rwxr-xr-x. 1 parallels parallels 12879808 Apr 26 10:49 a.btaot
[parallels@fedora-linux-38 build]$ ls -la ./src/ | grep "bpftrace"
-rwxr-xr-x. 1 parallels parallels 14274088 Apr 26 07:25 bpftrace
-rw-r--r--. 1 parallels parallels  1191884 Apr 26 07:25 libbpftrace.a
```

jordalgo:
@arthurshau Nice work! Tested it out locally, with just some basic probes, and works just fine. 

Few follow-up questions off the top of my head (for the group):
- Are we going to ship `bpftrace-aotrt` with the normal `bpftrace` binary for all distros?
- What bpftrace features currently aren't supported in AOT? (I'm sure there are some listed here: https://dxuuu.xyz/aot-bpftrace.html)
- At what point do we add documentation?
- How do we cut down on the file size?
- Should we add functionality to build for different arch targets? (this might be a pain)

src/aot/aot.cpp
```diff
@@ -253,16 +254,66 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
+
+    if (strcmp(secname, AOT_ELF_SECTION) == 0) {
+      data = elf_getdata(scn, 0);
+      if (!data) {
+        LOG(ERROR) << "Failed to BTAOT ELF section(" << i
+                   << ") data: " << elf_errmsg(-1);
+        goto out;
+      }
+
+      btaot_section = static_cast<uint8_t *>(data->d_buf);
```
danobi, line 305: `break`?

src/aot/aot.h
```diff
@@ -9,6 +9,8 @@
 namespace bpftrace {
 namespace aot {
 
+constexpr std::string_view AOT_SHIM_NAME = "bpftrace-aotrt";
```
danobi, line 12: I think `static constexpr` is more correct but I forget why

src/aot/aot_main.cpp
```diff
@@ -70,6 +70,12 @@ int main(int argc, char* argv[])
     option{ nullptr, 0, nullptr, 0 }, // Must be last
   };
 
+  if (strcmp(argv[0], aot::AOT_SHIM_NAME.data()) == 0) {
```
danobi, line 73: I think you should be able to just `argv[0] == aot::AOT_SHIM_NAME`

danobi (commented):
just some small nits.

sorry about the delay!

src/aot/aot.cpp
arthurshau: Unfortunately, because of the three goto statements above this line, doing ```int i = 0;``` causes the compiler to complain with ```error: jump to label ‘out .... ’crosses initialization of ‘int i’```. We can work around it by splitting the declaration and assignment like so, or moving the whole thing above the gotos (which it was in a previous version of the commit, but I think we prefer to have ```int i``` closer to the while loop).

arthurshau:
> Maybe also for a follow-up diff but the generated binary still seems rather large

> How do we cut down on the file size?

Yeah, one of the items on the list in #2918 is to re-remove the large list of runtime deps from the shim, which should hopefully help a lot.

> Are we going to ship bpftrace-aotrt with the normal bpftrace binary for all distros?

I believe ```bpftrace-aotrt``` is already getting shipped? At least, I found it in my ```/usr/bin/``` alongside my normal ```bpftrace``` install. Currently, the code searches for ```bpftrace-aotrt``` in $PATH to copy the shim, so it will need to exist somewhere.

> General question: With this PR, is this feature ready enough to add to the --help output?

> At what point do we add documentation?

Probably not ready to add to ```--help``` yet, ideally we should get ```runtime-tests.sh --run-aot-tests``` to pass first I think.

danobi (approved):

src/aot/aot.cpp
jordalgo: Huh, that is surprising but ok.

jordalgo (approved):

viktormalik:
> Second, we now also serialize the map of BpfMaps owned by BpfBytecode into the bytecode as well, which is necessary for runtime.

Why is this necessary? `BpfMap` should not contain any information that is not stored inside the bytecode. At least that was the original purpose when I created it so if it does, we should figure that out before merging this PR.

src/aot/aot.cpp
```diff
@@ -253,16 +253,67 @@ int load(BPFtrace &bpftrace, const std::string &in)
...
+      LOG(ERROR) << "Failed to get ELF section(" << i
+                 << ") hdr name: " << elf_errmsg(-1);
+      goto out;
+    }
+
+    if (strcmp(secname, AOT_ELF_SECTION) == 0) {
+      data = elf_getdata(scn, 0);
+      if (!data) {
+        LOG(ERROR) << "Failed to BTAOT ELF section(" << i
+                   << ") data: " << elf_errmsg(-1);
```
viktormalik, line 301: Nit: typo?
```suggestion
        LOG(ERROR) << "Failed to get BTAOT ELF section(" << i
                   << ") data: " << elf_errmsg(-1);
```

viktormalik (changes requested):
First of all, sorry for the delayed review! I really like this work and hope that we'll be able to get it in soon. With that, I apologize if the following will cause any delays but I think that it's an important thing to mention.

I think that there's a major flaw in the way we serialize bytecode at the moment. The current workflow is:
1. codegen -> ELF -> BpfBytecode -> serialization to .btaot
2. deserialization from .btaot -> BpfBytecode -> load BPF programs

With the upcoming move to libbpf, this will not let us leverage the full power of libbpf which is able to deduce a lot of information about the BPF maps and programs from the ELF. I think that the correct workflow should be:
1. codegen -> ELF -> serialization to .btaot
2. deserialization from .btaot -> ELF -> BpfBytecode -> load BPF programs

This is also the reason why you had to serialize `BpfMap` which I specifically tried to avoid when I was implementing it. Ideally, `BpfMap` instances should be created by running libbpf parsing on the ELF produced by bpftrace.

In addition, I plan to drop `BpfBytecode.sections_` completely and use libbpf's `struct bpf_object` instead (which is created from the ELF). When that happens, the current AOT approach would stop working and we'd have to move to the workflow suggested above anyways.

So, I suggest we move to serializing the produced ELF instead of `BpfBytecode` before merging this PR.

arthurshau:
Hi Viktor, thanks for the comment. You make good points, and I forgot to consider your work moving bpftrace to use libbpf when opening this PR. Given what you said about removing ```BpfBytecode.sections_```, I agree that it makes sense to switch the AOT workflow now so we can leverage libbpf. I'll look into it.

viktormalik:
> Hi Viktor, thanks for the comment. You make good points, and I forgot to consider your work moving bpftrace to use libbpf when opening this PR. Given what you said about removing `BpfBytecode.sections_`, I agree that it makes sense to switch the AOT workflow now so we can leverage libbpf. I'll look into it.

Thanks @arthurshau! I think that it could be as simple as serializing the ELF produced by codegen (as raw binary data) and moving creation of `BpfBytecode` to `run_bpftrace`.

arthurshau:
@viktormalik Sorry for the delay, I've implemented the requested changes. Can you double check that the serialization portion in particular is done correctly? I basically just generate the ELF as a temporary file and then append it to the end after the AOT header and RequiredResources.

Also, when running the generated binary, I'm now getting these libbpf warnings for some reason:
```
libbpf: elf: skipping unrecognized data section(15) .eh_frame
libbpf: elf: skipping relo section(16) .rel.eh_frame for section(15) .eh_frame
```
Might you know why?

ajor:
Just had a quick skim, but we should try and keep the ELF file in memory until the actual serialisation step.

`CodegenLLVM::emit_elf` is just a tool for debugging bpftrace. Take a look at `CodegenLLVM::emit` to see how we can operate on the ELF file in memory instead of writing it to disk.

arthurshau:
Addressed comments, thanks! Also moved the ```libbpf_print``` and ```libbpf_print_level_string``` functions from ```main.cpp``` to the shared ```run_bpftrace.cpp``` because ```aot_main.cpp``` needs them as well - it wasn't properly setting the log callback function for libbpf warnings.

viktormalik (commented):
LGTM, I'm just not approving the PR, yet, as I'd like to take some time trying this out.

viktormalik:
> Also moved the `libbpf_print` and `libbpf_print_level_string` functions from `main.cpp` to the shared `run_bpftrace.cpp` because `aot_main.cpp` needs them as well - it wasn't properly setting the log callback function for libbpf warnings.

Yeah, makes sense.



> @viktormalik Sorry for the delay, I've implemented the requested changes. Can you double check that the serialization portion in particular is done correctly? I basically just call `emit` and then append the vector to the end after the AOT header and RequiredResources.

Yes, that should be correct.

> Also, when running the generated binary, I'm now getting these libbpf warnings for some reason:
> 
> ```
> libbpf: elf: skipping unrecognized data section(15) .eh_frame
> libbpf: elf: skipping relo section(16) .rel.eh_frame for section(15) .eh_frame
> ```
> 
> This happens when calling `elf::parseBpfBytecodeFromElfObject`. Might you know why?

I'm not sure why this is happening but I'm seeing this when running in non-AOT mode, too. It's certainly something that we should address but it shouldn't be caused by this PR.

danobi:
:rocket:

viktormalik (approved):
Basic probes work nice, some others don't but since this is still undocumented, I'm ok with merging this as-is and fixing things in the follow-up work.

My remarks:
1. Using `--aot` searches `$PATH` for `bpftrace-aotrt` which is not very convenient when doing development. Maybe we could make it also (recursively) search `$PWD` (or the directory of the `bpftrace` binary) so that developers don't have to update their `$PATH` manually?
2. BTF-based probes (e.g. `kfunc`) currently do not work b/c the AOT main doesn't run `bpftrace.parse_btf()`. Here, it'll be tricky to support kernel modules as in normal mode, we parse the probes to determine which kernel modules BTF needs to be parsed. Since we don't have that information in pre-compiled binary, we'll have to figure out a different way to get the list of modules that the program may attach to.
3. Since this will likely get merged shortly before we do the new release, I just want to make sure that we agree that we're leaving this undocumented for now.

arthurshau:
> 3. Since this will likely get merged shortly before we do the new release, I just want to make sure that we agree that we're leaving this undocumented for now.

Sounds good, I agree with that.

danobi:
> Using --aot searches $PATH for bpftrace-aotrt which is not very convenient when doing development. Maybe we could make it also (recursively) search $PWD (or the directory of the bpftrace binary) so that developers don't have to update their $PATH manually?

Yeah, need to find a better way.

> Since this will likely get merged shortly before we do the new release, I just want to make sure that we agree that we're leaving this undocumented for now.

Agreed. We don't wanna document / announce anything until at least #2918 is completed.

PR #3105, opened 2024-04-07
Update log messages

jordalgo:
- `DEBUG`: log info to the user regardless of log level
(like using stdout and comes with file and line number).
No prefix.
- `V1`: log info to the user only if verbose logging is
enabled (bt_verbose or -v) Note: -vv will be deleted as
it's only used in two places in one file. No prefix.
- `WARNING`: log info that might affect bpftrace behavior
or output but allows the run to continue (like using
stderr). Prefix: "WARNING:"
- `ERROR`: log info to indicate that the user did
something invalid which will (eventually) cause bpftrace
to exit (via `exit(1)` or `LOG(FATAL)`). This should get
used if there a multiple errors that need to get logged
to the user otherwise use `FATAL`. Prefix: "ERROR:"
- `FATAL`: abort run and log info to indicate that the
user did something invalid which caused bpftrace to exit.
Prefix: "ERROR:"
- `BUG`: abort run and log info to indicate that there
is an internal/unexpected issue (not caused by invalid
user program code or CLI use). Prefix: "BUG:"

Also update many sites in accordance with the above
schema.

However, there are still many places that utilize
LOG(ERROR) that could be switched to LOG(FATAL) or
LOG(WARNING) but some changes seemed beyond this PR.

##### Checklist

### DIFF
--- a/src/ast/passes/field_analyser.cpp
+++ b/src/ast/passes/field_analyser.cpp
@@ -155,7 +155,7 @@ void FieldAnalyser::visit(Unop &unop)
   }
 }
 
-bool FieldAnalyser::resolve_args(Probe &probe)
+void FieldAnalyser::resolve_args(Probe &probe)
 {
   // load probe arguments into a special record type "struct <probename>_args"
   Struct probe_args;
@@ -173,7 +173,7 @@ bool FieldAnalyser::resolve_args(Probe &probe)
         matches = bpftrace_.probe_matcher_->get_matches_for_ap(*ap);
       } catch (const WildcardException &e) {
         LOG(ERROR) << e.what();
-        return false;
+        return;
       }
 
       // ... and check if they share same arguments.
@@ -223,7 +223,7 @@ bool FieldAnalyser::resolve_args(Probe &probe)
             ap->func, probe_type == ProbeType::kretfunc, err);
         if (!maybe_probe_args.has_value()) {
           LOG(ERROR, ap->loc, err_) << "kfunc:" << ap->func << ": " << err;
-          return false;
+          return;
         }
         probe_args = std::move(*maybe_probe_args);
       } else // uprobe
@@ -232,7 +232,8 @@ bool FieldAnalyser::resolve_args(Probe &probe)
         if (dwarf)
           probe_args = dwarf->resolve_args(ap->func);
         else {
-          LOG(ERROR, ap->loc, err_) << "No debuginfo found for " << ap->target;
+          LOG(WARNING, ap->loc, err_)
+              << "No debuginfo found for " << ap->target;
         }
         if ((int)probe_args.fields.size() > (arch::max_arg() + 1)) {
           LOG(ERROR, ap->loc, err_) << "\'args\' builtin is not supported for "
@@ -252,7 +253,7 @@ bool FieldAnalyser::resolve_args(Probe &probe)
       bpftrace_.structs.Add(probe.args_typename(), std::move(probe_args));
     }
   }
-  return true;
+  return;
 }
 
 void FieldAnalyser::resolve_fields(SizedType &type)
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -45,7 +45,6 @@ namespace bpftrace {
 DebugLevel bt_debug = DebugLevel::kNone;
 bool bt_quiet = false;
 bool bt_verbose = false;
-bool bt_verbose2 = false;
 volatile sig_atomic_t BPFtrace::exitsig_recv = false;
 volatile sig_atomic_t BPFtrace::sigusr1_recv = false;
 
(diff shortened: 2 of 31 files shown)


ajor (commented):
We don't want to actually call `abort()` for user errors as that triggers a SIGABRT and a core dump - not a great user experience! `ABORT` might not be a good name for this error level after all. Maybe we should stick with `FATAL` internally but still label things as "ERROR" for users?

Instead of calling `abort()`, it'd be better to throw an exception and catch it in main - then we can exit from an error cleanly, e.g. with `return 1`. (we shouldn't call `exit(1)` deep within the logging code either as that would prevent using bpftrace as a library)

jordalgo:
Re-converting to draft as it seems there are still some issues to be ironed out.

danobi:
> Instead of calling abort(), it'd be better to throw an exception and catch it in main - then we can exit from an error cleanly, e.g. with return 1. (we shouldn't call exit(1) deep within the logging code either as that would prevent using bpftrace as a library)

Throwing an exception across library boundary would limit the library to C++ consumers. Which could be ok. But we'll inevitably get the question "but I want to use this from [rust|python|c]...". Could do a C wrapper around the C++ interfaces like LLVM, though

jordalgo:
@ajor I made many of the changes you suggested including changing `LOG(ABORT)` back to `LOG(FATAL)` (but utilizing the same "ERROR: " prefix as `LOG(ERROR)`. I also changed many `LOG(FATAL)`s to `LOG(BUG)` as it seemed more appropriate (e.g. at the default branch of a switch statement). 

While I still want to spend time sorting out `LOG(ERROR)` and `LOG(FATAL)` (especially in the case of exiting vs throwing) I think that can wait for a follow-up PR. I believe this one still have some valuable changes to the log scheme. Let me know if you disagree and I can just abandon.

src/probe_matcher.cpp
```diff
@@ -479,7 +479,7 @@ void ProbeMatcher::list_probes(ast::Program* prog)
         std::cout << probe_type << ":" << match_print << std::endl;
         if (bt_verbose) {
           for (auto& param : param_lists[match])
-            std::cout << "    " << param << std::endl;
+            LOG(V1) << "    " << param;
```
viktormalik, line 482: We could possibly drop the above `if (bt_verbose)` (although then the loop will always be executed).

src/main.cpp
```diff
@@ -282,105 +277,88 @@ static std::optional<struct timespec> get_delta_taitime()
...
+
+  get_uint64_env_var("BPFTRACE_MAX_CAT_BYTES", [&](uint64_t x) {
+    config_setter.set(ConfigKeyInt::max_cat_bytes, x);
+  });
 
   if (const char* env_p = std::getenv("BPFTRACE_CACHE_USER_SYMBOLS")) {
     const std::string s(env_p);
     if (!config_setter.set_user_symbol_cache_type(s))
-      return false;
+      exit(1);
```
viktormalik, line 333: Why not `LOG(FATAL)` in `ConfigSetter::set_user_symbol_cache_type` and get rid of the `if` and `exit(1)`? Same below for `set_stack_mode`.

viktormalik (commented):
I think that this PR makes sense as it nicely cleans some exit paths. Few comments/suggestions below.

src/probe_matcher.cpp
jordalgo: > although then the loop will always be executed

Yeah that's why I left it in.

src/main.cpp
jordalgo: Because then I can't easily test this code path in unit tests (AFAIK).

src/bpftrace.cpp
```diff
@@ -115,15 +114,17 @@ int BPFtrace::add_probe(ast::Probe &p)
       try {
         matches = probe_matcher_->get_matches_for_ap(*attach_point);
       } catch (const WildcardException &e) {
-        LOG(ERROR) << e.what();
+        LOG(WARNING) << e.what();
```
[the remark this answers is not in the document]
jordalgo: I think that's why I made it a warning instead of an error - because it wasn't terminating.

src/ast/passes/field_analyser.cpp
```diff
@@ -172,8 +172,8 @@ bool FieldAnalyser::resolve_args(Probe &probe)
       try {
         matches = bpftrace_.probe_matcher_->get_matches_for_ap(*ap);
       } catch (const WildcardException &e) {
-        LOG(ERROR) << e.what();
-        return false;
+        LOG(WARNING) << e.what();
```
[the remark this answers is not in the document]
jordalgo: Ah I wasn't sure if the script could continue if any of these were hit. It's a bit hard to tell TBH

viktormalik (approved):
Nice!

jordalgo:
@ajor Any objections to landing as is?

PR #3106, opened 2024-04-08
Report correct helper for error messages

ajor:
Previously, the first helper in a script would always be reported as the issue, even if it was not the cause of the verifier error.

The logic has been wrapped in a function (maybe_throw_helper_verifier_error) to allow it to be reused for other helper errors in the future.

### DIFF
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -449,9 +449,13 @@ RUN {{BPFTRACE}} -e 'BEGIN { if (strcontains("abc", "a")) { printf("OK\n"); exit
 EXPECT OK
 TIMEOUT 1
 
+# The extra prints are here to confirm that we report on the correct helper, not
+# just the first or the last one.
 NAME path_in_unsupported_kfunc
-PROG kfunc:vfs_read { print(path(args.file->f_path)); }
-EXPECT stdin:1:18-47: ERROR: helper bpf_d_path not supported in probe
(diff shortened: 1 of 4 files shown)


tests/runtime/call
```diff
@@ -449,9 +449,13 @@ RUN {{BPFTRACE}} -e 'BEGIN { if (strcontains("abc", "a")) { printf("OK\n"); exit
...
 
+# The extra prints are here to confirm that we report on the correct helper, not
+# just the first or the last one.
 NAME path_in_unsupported_kfunc
-PROG kfunc:vfs_read { print(path(args.file->f_path)); }
-EXPECT stdin:1:18-47: ERROR: helper bpf_d_path not supported in probe
+PROG kfunc:vfs_read { print("a"); print(path(args.file->f_path)); print("b"); }
+EXPECT stdin:1:30-59: ERROR: helper bpf_d_path not allowed in probe
+EXPECT kfunc:vfs_read { print("a"); print(path(args.file->f_path)); print("b"); }
+EXPECT                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ajor, line 458: This location is wrong - it should just be pointing to `path(args.file->f_path)` and not to the surrounding print. It was already like this and the issue seems to run quite deep so leaving it for now.

viktormalik (approved):
Nice cleanup!

PR #3107, opened 2024-04-08
Runtime tests: Allow multiline PROGs and EXPECTs

ajor:
The previous options had some downsides:
- Define the program/expected results in an external file: Out-of-line definitions add barriers to quickly understanding tests and to adding new tests
- Use EXPECT_REGEX: Hard to read due to escape characters and being squashed onto a single line
- Use multiple EXPECTs: Do not enforce ordering of output lines

### DIFF
--- a/tests/runtime/engine/parser.py
+++ b/tests/runtime/engine/parser.py
@@ -123,11 +123,26 @@ class TestParser(object):
         will_fail = False
         new_pidns = False
         skip_if_env_has = None
+        prev_item_name = ''
 
         for item in test:
+            if item[:len(prev_item_name) + 1].isspace():
+                # Whitespace at beginning of line means it continues from the
+                # previous line
+
+                # Remove the leading whitespace and the trailing newline
+                line = item[len(prev_item_name) + 1:-1]
+                if prev_item_name == 'PROG':
+                    prog += '\n' + line
+                    continue
+                elif prev_item_name == 'EXPECT':
+                    expects[-1].expect += '\n' + line
+                    continue
+
             item_split = item.split()
             item_name = item_split[0]
             line = ' '.join(item_split[1:])
+            prev_item_name = item_name
 
             if item_name == 'NAME':
                 name = line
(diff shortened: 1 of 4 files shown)


tests/runtime/engine/parser.py
```diff
@@ -123,20 +123,33 @@ def __read_test_struct(test, test_suite):
...
         skip_if_env_has = None
+        prev_item_name = ''
 
         for item in test:
+            if item[:len(prev_item_name) + 1].isspace():
+                # Whitespace at beginning of line means it continues from the
+                # previous line
+                line = item[len(prev_item_name) + 1:]
+                if prev_item_name == 'PROG':
+                    prog += line
```
viktormalik, line 134: Should also add newline?
```suggestion
                    prog += line + "\n"
```

tests/runtime/engine/parser.py
```diff
@@ -123,20 +123,33 @@ def __read_test_struct(test, test_suite):
...
+                    continue
+                elif prev_item_name == 'EXPECT':
+                    expects[-1].expect += line
```
viktormalik, line 137: Same as above
```suggestion
                    expects[-1].expect += line + "\n"
```

viktormalik (commented):
Nice idea! Overall looks good but there's some issue (probably with the newline chars?) which makes a lot of tests fail.

ajor:
I've sorted out the newline issues now.

Needed to make sure that there was no trailing newline at the end of the expected-text string, as that broke the regex matching in runner.py, which was already expecting a trailing newline. Double-trailing newline was bad.

jordalgo (approved):
Nice! Much easier to read.

ajor:
One more thought I've just had: would something like this be a nicer syntax instead of just requiring the correct indentation?

```
EXPECT BEGIN
     | @[a]: 1
     | @[b]: 2
     | @[c]: 3
     | @[d]: 4
     |
     | END
```

jordalgo:
@ajor Eh, I don't find that easier to read but if it's a pain to figure out the indentation when you're writing the test then maybe it's worth it.

viktormalik (approved):
Nice!

viktormalik:
> One more thought I've just had: would something like this be a nicer syntax instead of just requiring the correct indentation?

The indentation syntax is nicer. I'd use that for now and if people start complaining that it's too hard to keep track of the indentation, then we can revisit the decision.

Issue #3058, opened 2024-03-14
Can't list probes used in script file


ajor:
As of #2969, it's possible to list probes used in a script by combining `-e` and `-l`:
```
$ sudo bpftrace -l -e 't:syscalls:sys_enter_ope* {}'
tracepoint:syscalls:sys_enter_open
tracepoint:syscalls:sys_enter_open_by_handle_at
tracepoint:syscalls:sys_enter_open_tree
tracepoint:syscalls:sys_enter_openat
tracepoint:syscalls:sys_enter_openat2
```

However, this doesn't work if the script is contained in a file instead of written directly on the command line:
```
$ sudo bpftrace -l test.bt                                              
stdin:1:1-8: ERROR: Invalid probe type: test.bt
test.bt
~~~~~~~
```

This should just a matter of fixing bpftrace's command line argument parsing logic.

jordalgo:
Good catch. If no one picks it up as a starter task, I'll take it.

ikruglov:
@jordalgo I'm wondering if this is a correct fix: https://github.com/ikruglov/bpftrace/commit/65626928a74055a8aab8ad7beb4d3253634a68c0

ikruglov:
I'm looking into `--help` and doubt that my commit will do the trick. There is a conflict, it seems:

```
USAGE:
    bpftrace [options] filename
    ...

OPTIONS:
    ...
    -l [search]    list kernel probes or probes in a program
```

Based on the snippet above, a command `bpftrace -l foo` is ambiguous.
Because, `foo` can be either search pattern or a file name.

I guess we can resolve this via following logic:
1. assumes that `foo` is a file.
2. check the file existence. If yes, proceed accordingly
3. if not, treat `foo` as search command.

This approach is not ideal. A missing file would be executed as a search.

Is there any disambiguity technic which I can use?

jordalgo:
@ikruglov Good point about the ambiguity.

Another option would be to remove the positional search param for listing probes and just let people use grep. Or change search from being a positional param to a named one e.g. `--search. @ajor Any thoughts here?

ajor:
The `-l search` option is more powerful than a grep as it supports globbing, C++ name mangling and argument listing (with `-v`), so we shouldn't get rid of it.

Reworking the arguments is an option, but I'd try to avoid as much as possible so that we don't invalidate older tutorials around the web.

> 1. assume that foo is a file.
> 2. check the file existence. If yes, proceed accordingly
[...]

Issue #3057, opened 2024-03-14
Probe listing incorrect for BEGIN/END probes


ajor:
```
# bpftrace -l 'BEGIN'
special::
# bpftrace -l 'END'
special::
```

ikruglov:
@ajor, can you please help me to understand what is expected output?

ajor:
What I was thinking when I opened this issue was probably:
```
# bpftrace -l 'BEGIN'
BEGIN
# bpftrace -l 'END'
END
```

But it's probably not something people will ever use and was only really for completeness. It might conflict with your other changes in #3111 (which are definitely more useful), so maybe we just close this one

ikruglov:
ok. closing :-)

ikruglov:
ahm... I don't have permission it seems.

PR #666, opened 2019-05-22
Fix possible crash if an invalid char is used in search (#664)

caringi:
Avoid this kind of crash if by mistake a string with an invalid
character is used with '-l':

bpftrace -l 'BEGIN { printf("hello world\n"); }'
terminate called after throwing an instance of 'std::regex_error'
  what():  Unexpected escape character.

PR #3111, opened 2024-04-10
[#3058] fix bpftrace -l test.bt

ikruglov:
Fix for #3058.

The fix involves some heuristic because there is a conflict:

```
USAGE:
    bpftrace [options] filename
    ...

OPTIONS:
    ...
    -l [search]    list kernel probes or probes in a program
```

Based on the snippet above, a command `bpftrace -l foo` is ambiguous.
Because, `foo` can be either search pattern or a file name.

The PR resolves this via the following logic:
1. assume that `foo` is a file.
2. check the file existence. If yes, proceed accordingly
3. if not, treat `foo` as search command.

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -669,9 +669,13 @@ Args parse_args(int argc, char* argv[])
     if (optind == argc) {
       args.search = "*:*";
     } else if (optind == argc - 1) {
-      args.search = argv[optind];
-      if (args.search == "*") {
+      std::string_view val(argv[optind]);
+      if (val == "*") {
         args.search = "*:*";
+      } else if (val.find_first_of(":*") != std::string::npos) {
+        args.search = val;
+      } else {
+        args.filename = val;
       }
       optind++;
     } else {
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -158,6 +158,19 @@ EXPECT hardware:cache-misses:
 EXPECT tracepoint:xdp:mem_connect
 TIMEOUT 5
 
+NAME it lists probes in a given file
+RUN {{BPFTRACE}} -l runtime/scripts/interval_order.bt
+EXPECT interval:ms:
+EXPECT interval:ms:
+EXPECT interval:ms:
+TIMEOUT 5
+
+NAME errors on non existent file
+RUN {{BPFTRACE}} -l non_existent_file.bt
+EXPECT ERROR: failed to open file 'non_existent_file.bt': No such file or directory
+TIMEOUT 1
+WILL_FAIL
+
 NAME errors on invalid character in search expression
 RUN {{BPFTRACE}} -l '\n'
 EXPECT stdin:1:1-2: ERROR: invalid character '\'
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -171,12 +171,6 @@ EXPECT ERROR: failed to open file 'non_existent_file.bt': No such file or direct
 TIMEOUT 1
 WILL_FAIL
 
-NAME errors on invalid character in search expression
-RUN {{BPFTRACE}} -l '\n'
-EXPECT stdin:1:1-2: ERROR: invalid character '\'
-TIMEOUT 1
-WILL_FAIL
-
 NAME pid fails validation with leading non-number
 RUN {{BPFTRACE}} -p a1111 file.bt
 EXPECT FATAL: Failed to parse pid: pid 'a1111' is not a valid decimal number


src/utils.h
```diff
@@ -266,6 +266,15 @@ inline std::string &trim(std::string &s)
   return ltrim(rtrim(s));
 }
 
+inline bool ends_with(const std::string &s, const std::string &suffix)
```
jordalgo, line 269: Sadly, I don't think we're on C++ 20 yet otherwise we could use ends_with:
https://en.cppreference.com/w/cpp/string/basic_string/ends_with

jordalgo:
@ikruglov Mind adding a runtime test for this? After this test would be a good place:
https://github.com/bpftrace/bpftrace/blob/3b2cda69d4aeb0eb370225e8456f864a60ac9a1b/tests/runtime/basic#L155

src/main.cpp
```diff
@@ -669,9 +669,15 @@ Args parse_args(int argc, char* argv[])
     if (optind == argc) {
       args.search = "*:*";
     } else if (optind == argc - 1) {
-      args.search = argv[optind];
-      if (args.search == "*") {
+      std::string val(argv[optind]);
+      if (val == "*") {
         args.search = "*:*";
+      } else if (val.find(":") != std::string::npos) {
```
jordalgo, line 675: I think we can also check if `*` exists in the value and assume it's not a file.

src/utils.h
ikruglov: yeah, I tried std::string::ends_with but it fails compilation. So, had to do this helper.

src/main.cpp
```diff
@@ -669,9 +669,15 @@ Args parse_args(int argc, char* argv[])
...
+        args.search = val;
+      } else if (ends_with(val, ".bt") || std_filesystem::exists(val)) {
```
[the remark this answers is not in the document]
ikruglov: don't know tbh. it will likely be a single use helper function

jordalgo (changes requested):

src/main.cpp
```diff
@@ -669,9 +669,15 @@ Args parse_args(int argc, char* argv[])
     if (optind == argc) {
       args.search = "*:*";
     } else if (optind == argc - 1) {
-      args.search = argv[optind];
-      if (args.search == "*") {
+      std::string val(argv[optind]);
```
ajor, line 672:
```suggestion
      std::string_view val(argv[optind]);
```
Just to eliminate a string copy, but string copies are a big cause of overhead in bpftrace at the moment so every little helps.

Also applies to the new helper functions in utils.h

tests/runtime/basic
```diff
@@ -158,6 +158,16 @@ EXPECT hardware:cache-misses:
...
 TIMEOUT 5
 
+NAME it lists given search pattern
+RUN {{BPFTRACE}} -l hardware:cache-misses
+EXPECT hardware:cache-misses:
+TIMEOUT 5
+
+NAME it lists given file pattern
+RUN {{BPFTRACE}} -l hardware-cache-misses.bt
+EXPECT_REGEX No such file or directory
```
ajor, line 168: Nit: Could you include the full error message and `EXPECT` instead of `EXPECT_REGEX`?

We've had problems in the past where partial expected outputs were hiding real bugs, so are trying to move away from them where possible

ajor:
I think all you'll need to fix the failing tests is a `WILL_FAIL` in your test definitions, e.g. https://github.com/bpftrace/bpftrace/blob/3b2cda69d4aeb0eb370225e8456f864a60ac9a1b/tests/runtime/basic#L15

Also maybe a difference of opinion with @jordalgo here, but I don't think we should bake in assumptions about file extensions - there's nothing special about ".bt" anywhere else in the codebase. The checks for "*" and ":" are good though.

Removing the ".bt" check would only cause an issue when trying to list the "BEGIN" and "END" probes, as in #3057. But tbh I'm not sure listing those is useful after all.

jordalgo:
> Also maybe a difference of opinion with @jordalgo here, but I don't think we should bake in assumptions about file extensions - there's nothing special about ".bt" anywhere else in the codebase

I always forget that ".bt" is not an official thing :) (we should change that!) That's fine. I don't feel strongly about keeping the ".bt" check.

ikruglov:
@jordalgo @ajor I've updated the PR with suggested changes. But I don't see a conclusion on '.bt' check. Should I keep it or remove it?

ajor:
Let's get rid of the ".bt" check and just assume it's a file if it doesn't contain ":*"

ikruglov:
@ajor this should be the final version

tests/runtime/basic
```diff
@@ -158,6 +158,22 @@ EXPECT hardware:cache-misses:
 EXPECT tracepoint:xdp:mem_connect
 TIMEOUT 5
 
+NAME it lists given search pattern
+RUN {{BPFTRACE}} -l hardware:cache-misses
+EXPECT hardware:cache-misses:
+TIMEOUT 5
+
+NAME it lists given search pattern
```
jordalgo, line 166:
```suggestion
NAME it lists probes in a given file
```

tests/runtime/basic
```diff
@@ -158,6 +158,22 @@ EXPECT hardware:cache-misses:
...
+RUN {{BPFTRACE}} -l runtime/scripts/hello_world.bt
+EXPECT special::
+TIMEOUT 5
+
+NAME errors on non existent file
+RUN {{BPFTRACE}} -l non_existent_file.bt
+EXPECT ERROR: failed to open file 'non_existent_file.bt': No such file or directory
```
jordalgo, line 173: This has a different error because it's now treated as a search right (removed the ".bt" check) ?

tests/runtime/basic
```diff
@@ -158,6 +158,22 @@ EXPECT hardware:cache-misses:
 EXPECT tracepoint:xdp:mem_connect
 TIMEOUT 5
 
+NAME it lists given search pattern
```
jordalgo, line 161: I don't think we need this additional tests. The ones above it already cover this case.

tests/runtime/basic
```diff
@@ -158,6 +158,22 @@ EXPECT hardware:cache-misses:
...
+RUN {{BPFTRACE}} -l hardware:cache-misses
+EXPECT hardware:cache-misses:
+TIMEOUT 5
+
+NAME it lists given search pattern
+RUN {{BPFTRACE}} -l runtime/scripts/hello_world.bt
+EXPECT special::
```
jordalgo, line 168: Super nit: can we use a script that doesn't only have a single special probe that doesn't print quite right ? lol

src/main.cpp
```diff
@@ -669,9 +669,15 @@ Args parse_args(int argc, char* argv[])
...
-      if (args.search == "*") {
+      std::string_view val(argv[optind]);
+      if (val == "*") {
         args.search = "*:*";
+      } else if (val.find_first_of(":*") != std::string::npos) {
+        args.search = val;
+      } else if (std_filesystem::exists(val)) {
+        args.filename = val;
+      } else {
+        args.search = val;
```
ajor, line 680:
```suggestion
      } else {
        args.filename = val;
```

I was thinking we could just treat everything not containing `:*` as a file - then error messages would still look sensible for non-existent files. All probes (other than BEGIN and END) will contain one of the `:*` characters so error messages for anything resembling them would look good too

jordalgo: I do find it a bit odd that search doesn't work for substrings in the listing path e.g. this message is not great
```
$sudo bpftrace -l tracepoint
stdin:1:1-11: ERROR: tracepoint probe type requires 2 arguments, found 0
tracepoint
~~~~~~~~~~
```

ikruglov: this code breaks existing test: `-bpftrace l '\n'`. It's registered under "errors on invalid character in search expression" name

ajor: Looking at why that test was added (#666), it was for some regex matching code that couldn't handle some characters. That regex matching code was completely removed in #1549. The lexer now detects invalid characters so invalid character-detection is a parser thing now, not specific to probe listing.

I think this runtime test is unnecessary and can be deleted.

If you fancy it, you could try replacing it with a Parser unit test instead, attempting to parse an invalid character like:
```
test_parse_failure("\\", R"(
```

Here's an example of a parser test:
https://github.com/bpftrace/bpftrace/blob/8ac0a49fe9481221260147aa722b8e150f8560e6/tests/parser.cpp#L1113-L1119)

ajor (approved):
Thank you @ikruglov, looks good!

jordalgo (approved):
Sorry, forgot to unblock.

PR #3113, opened 2024-04-10
Add nr_stack_frames to stack key

jordalgo:
This will further decrease the chances of collision because even if two stacks of different sizes hash to the same id, if they are of different sizes the stack key will be different. So now two stacks of the same size will have have to hash to the same id for there to be a collision.

This also has the small benefit of being able to use the this number when iterating over the stack in userspace.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -157,7 +157,22 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
   const bool is_ustack = ident == "ustack";
   const auto uint64_size = sizeof(uint64_t);
 
-  AllocaInst *stackid = b_.CreateAllocaBPF(CreateUInt64(), "stackid");
+  std::vector<llvm::Type *> stack_key_elements = {
+    b_.getInt64Ty(), // stack id
+    b_.getInt32Ty(), // nr_stack_frames
+  };
+  StructType *stack_key_struct = b_.GetStructType("stack_key",
+                                                  stack_key_elements,
+                                                  false);
+  AllocaInst *stack_key = b_.CreateAllocaBPF(stack_key_struct, "stack_key");
+  b_.CreateStore(b_.getInt64(0),
+                 b_.CreateGEP(stack_key_struct,
+                              stack_key,
+                              { b_.getInt64(0), b_.getInt32(0) }));
+  b_.CreateStore(b_.getInt32(0),
+                 b_.CreateGEP(stack_key_struct,
+                              stack_key,
+                              { b_.getInt64(0), b_.getInt32(1) }));
 
   Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *stack_scratch_failure = BasicBlock::Create(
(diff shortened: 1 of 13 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -157,6 +157,17 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
...
+                                                   "nr_stack_frames");
+
+  std::vector<llvm::Type *> stack_key_elements = {
+    b_.getInt64Ty(), // stack id
+    b_.getInt32Ty(), // nr_stack_frames
+  };
+  StructType *stack_key_struct = b_.GetStructType("stack_key",
+                                                  stack_key_elements,
+                                                  false);
+  AllocaInst *stack_key = b_.CreateAllocaBPF(stack_key_struct, "stack_key");
```
viktormalik, line 170: Is there a reason to store `stackid` and `nr_stack_frames` both as scalars and inside the `stack_key` struct? I know that LLVM will probably optimize this away but it just seems like a waste of stack space (and is a bit difficult to read).

src/bpftrace.cpp
```diff
@@ -1757,31 +1759,33 @@ std::optional<std::vector<uint8_t>> BPFtrace::find_empty_key(
...
 
   std::ostringstream stack;
   std::string padding(indent, ' ');
 
   stack << "\n";
-  for (auto &addr : stack_trace) {
-    if (addr == 0)
-      break;
+  uint32_t i;
+  for (i = 0; i < nr_stack_frames; ++i) {
```
viktormalik, line 1787: I don't think that we need to follow kernel's "forward variable declaration" policy:
```suggestion
  for (uint32_t i = 0; i < nr_stack_frames; ++i) {
```

src/types.cpp
```diff
@@ -362,7 +363,8 @@ SizedType CreateRecord(const std::string &name, std::weak_ptr<Struct> record)
 
 SizedType CreateStack(bool kernel, StackType stack)
 {
-  auto st = SizedType(kernel ? Type::kstack : Type::ustack, kernel ? 8 : 16);
+  // These sizes are based on the stack key (see CodegenLLVM::kstack_ustack)
+  auto st = SizedType(kernel ? Type::kstack : Type::ustack, kernel ? 12 : 20);
```
viktormalik, line 367: Out of scope: I'm wondering if we could define the structs somewhere and then just use `sizeof(...)` to get the correct size (and possibly offsets, too). This applies to all kinds of codegen-defined structs so it's probably worth a separate larger refactoring.

jordalgo: Yeah agreed. Having to change these in multiple places is not great.

src/ast/passes/codegen_llvm.cpp
jordalgo: I think I originally did it because ustack returns a different struck that includes the pid and probe id but I could just use the stack_key and GEP to pull out the `stackid` and `nr_stack_frames` for this case. Let me try that and see if it makes it more readable.

jordalgo: Ok, that was easy enough to change.

viktormalik (approved):
LGTM

Issue #3017, opened 2024-02-25
Variables declared inside loop does not respect lexical scope


danobi:
### What reproduces the bug? Provide code if possible.
```
#!/usr/bin/env bpftrace

BEGIN {
        $i = 0;
        while ($i < 1) {
                $j = 0;
                $i++;
        }

        print($j);
        exit();
}
```

```
$ $ sudo ./build/src/bpftrace ~/scratch/loop_var.bt
Attaching 1 probe...
0
```

I expect to see some error about `$j` being undefined.

### `bpftrace --info` output
```
$ sudo ./build/src/bpftrace --info
System
  OS: Linux 6.7.3-arch1-2 #1 SMP PREEMPT_DYNAMIC Fri, 02 Feb 2024 17:03:55 +0000
  Arch: x86_64

Build
  version: v0.20.0-43-g6d9d8
  LLVM: 16.0.6
  unsafe probe: no
  bfd: yes
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: yes
  uprobe_multi: no
  raw_tp_special: yes
  iter: yes
```

ajor:
If-statements as well 😢 

```
$ sudo bpftrace -e 'BEGIN { if (0) { $var = 1; } print($var) }'
Attaching 1 probe...
0
^C
```

viktormalik:
> Fixing this would technically be backwards incompatible. But surely nobody is relying on this buggy behavior, right?

Right, we shouldn't maintain backwards compatibility of an obviously buggy behaviour.

[...]

PR #3116, opened 2024-04-12
Codegen: Fix variable scoping in for-loops

ajor:
Variable scoping is not just a semantics-issue with for-loops: since they are implemented as BPF subprograms, it is essential that each for-loop gets its own allocations for variables of the same name.

```
BEGIN {
  @map[0] = 1;
  for ($kv : @map) { $var = 1; }
  for ($kv : @map) { $var = 1; }
}
```

Previously, $var in the second for-loop would try to re-use the stack allocation of $var from the first, due to a lack of proper scoping. This meant that the subprogram for the second loop ended up with no valid stack space in which to put $var, resulting in LLVM generating invalid IR and marking it as unreachable.

We should be able to get rid of this code once scoping is fixed (#3017).

##### Checklist

For-loops have not made it to a bpftrace release yet, so this does not need a changelog entry.

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -37,6 +37,7 @@
 #include "ast/elf_parser.h"
 #include "ast/signal_bt.h"
 #include "bpfmap.h"
+#include "collect_nodes.h"
 #include "log.h"
 #include "tracepoint_format_parser.h"
 #include "types.h"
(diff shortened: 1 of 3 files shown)


jordalgo (approved):
Nit: do you want to add a runtime test or is that overkill?

ajor:
> Nit: do you want to add a runtime test or is that overkill?

Personally, I try to avoid adding runtime tests whenever I can - they're so slow to run.

I should have included this earlier, but the checked in IR makes it clear that it'll work now - the functions `map_for_each_cb` and `map_for_each_cb.1` are identical. Previously, the second for-loop (`map_for_each_cb.1`) was missing the first four instructions which created the `$var` variable:
```
  %"$var" = alloca i64, align 8
  %5 = bitcast i64* %"$var" to i8*
  call void @llvm.lifetime.start.p0i8(i64 -1, i8* %5)
  store i64 0, i64* %"$var", align 8
```

I look forward to being able to delete this test anyway once scoping is fixed for real!

PR #3121, opened 2024-04-16
Fix errors in tutorial_one_liners.md and tutorial_one_liners_chinese.md

ZhangYet:
When running the example code like

```bash
bpftrace -e 'tracepoint:syscalls:sys_exit_read  { @bytes = hist(args.ret); }'
```

we will get error message as below:

tracepoint:syscalls:sys_exit_read  { @bytes = hist(args.ret); }

We should use arrow operators instead of dot operators. See also tutorial_one_liners_japanese.md.

##### Checklist


jordalgo (approved):
Thanks for the cleanup!

Super nit but the error message seems to be missing in the commit message.

viktormalik (changes requested):
This has been changed in a recent version of bpftrace and using the dot operator is now the recommended way to access the `args` builtin fields.

viktormalik:
For completeness, it was changed in #2578 and released in bpftrace 0.19. I assume that you're using an older version of bpftrace.

We've had a number of such PRs so we should probably address this issue somehow.

PR #3122, opened 2024-04-16
Doc: Access member variables correctly

YeechouTang:
Now, member variables should be accessed using '->' instead of '.'.

Both the original document and the Chinese translation have this issue, but the Japanese translation does not.


YeechouTang:
> It's actually the other way around. Since #2578 (bpftrace 0.19), the recommended way to access `args` fields is via the dot operator. Using the arrow works, too, for backwards compatibility.

I see. The version of bpftrace installed on my Ubuntu is 0.15, and it throws an error when using '.' to run examples. So, perhaps a better approach would be to clarify in the documentation from which version '.' can be used?

viktormalik:
> > It's actually the other way around. Since #2578 (bpftrace 0.19), the recommended way to access `args` fields is via the dot operator. Using the arrow works, too, for backwards compatibility.
> 
> I see. The version of bpftrace installed on my Ubuntu is 0.15, and it throws an error when using '.' to run examples. So, perhaps a better approach would be to clarify in the documentation from which version '.' can be used?

We have a plan to provide "versioning" for the documentation so that it is possible to access documentation for older versions of bpftrace. That will, however, take time to implement so adding a note to the one-liners tutorial probably makes sense as many people seem to stumble on this.

In addition, I've opened #3123 which adds a warning to the PR template which should notify people to actually reproduce their change on the latest version of bpftrace.

viktormalik:
Anyways, I'm closing this PR.

viktormalik:
I opened #3124 which adds a note to the one-liners tutorial (English version only for now).

PR #3127, opened 2024-04-17
Improve unit tests for Probe creation

ajor:
- Validate uprobe "path" value
- Validate [ku]probe multi attach function lists

Want to make sure the tests are thorough before refactoring the `Probe` struct.


jordalgo (approved):
LGTM

PR #3129, opened 2024-04-18
Transform codegen test "struct_semicolon"  into a Parser unit test

### LINKED ISSUE #3115: Convert codegen.struct_semicolon test into a Parser unit test
### Problem

This test is checking the code generation for programs containing struct definitions with and without trailing semicolons: [tests/codegen/struct_semicolon.cpp](https://github.com/bpftrace/bpftrace/blob/8ac0a49fe9481221260147aa722b8e150f8560e6/tests/codegen/struct_semicolon.cpp)

Expected results are recorded in separate files:
1. [tests/codegen/llvm/struct_semicolon_1.ll](https://github.com/bpftrace/bpftrace/blob/8ac0a49fe9481221260147aa722b8e150f8560e6/tests/codegen/llvm/struct_semicolon_1.ll)
2. [tests/codegen/llvm/struct_semicolon_2.ll](https://github.com/bpftrace/bpftrace/blob/8ac0a49fe9481221260147aa722b8e150f8560e6/tests/codegen/llvm/struct_semicolon_2.ll)

This is overkill for what it's actually meant to be checking, i.e. that we can parse a struct definition with a trailing semicolon. It adds noise when changing unrelated bits of code generation, making it hard to create pull requests and code harder to review.

### Solution

Replace this codegen test with a Parser unit test.

A parser unit test just confirms that we can parse a BpfScript program successfully and that it produces the correct Abstract Syntax Tree (AST).

[Parser.cstruct](https://github.com/bpftrace/bpftrace/blob/8ac0a49fe9481221260147aa722b8e150f8560e6/tests/parser.cpp#L2056-L2064) is an example of a Parser unit test which tests structs without trailing semicolons. We should create a second test like this that includes a semicolon after the struct definition.

gystemd:
This PR transforms the `struct_semicolon` codegen test into a Parser unit test. It verifies whether a semicolon-trailed struct declaration can be correctly parsed.

List of changes:
- Introduced the new test in file [Parser.cpp](https://github.com/gystemd/bpftrace/blob/6df2374d1d593ece64ef9a571e78d029cc21f7ba/tests/parser.cpp#L2066)
- Removed old tests and related output files

fixes #3115
##### Checklist


ajor (approved):
Nice, thank you @gystemd!

PR #3130, opened 2024-04-19
Refactor attach-point expansion and creation of Probe objects

viktormalik:
Currently, we have two places where wildcard expansion is done:
- codegen for probe expansion (e.g. when the `probe` builtin is used)
- `Probe` generation for attach point expansion (for most cases)

These two places have a lot of repetitive code (both within and between them) and are sometimes very difficult to read. This PR attempts to simplify that. Ideally, we should be able to unify the expansion code eventually, this is just a first step towards that goal.

See individual commit messages for a detailed explanation of the changes.

This is also the first step towards #2334.

##### Checklist

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -587,18 +587,69 @@ std::string opstr(const Unop &unop)
   return {}; // unreached
 }
 
-std::string AttachPoint::name(const std::string &attach_target,
-                              const std::string &attach_point) const
+AttachPoint AttachPoint::create_expansion_copy(const std::string &match) const
+{
+  AttachPoint ap = *this; // copy here
+  switch (probetype(ap.provider)) {
+    case ProbeType::kprobe:
+    case ProbeType::kretprobe:
+      ap.func = match;
+      if (match.find(":") != std::string::npos)
+        ap.target = erase_prefix(ap.func);
+      break;
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+    case ProbeType::kfunc:
+    case ProbeType::kretfunc:
+    case ProbeType::tracepoint:
+      // Tracepoint, uprobe, and k(ret)func probes specify both a target
+      // (category for tracepoints, binary for uprobes, and kernel module
+      // for k(ret)func) and a function name.
+      ap.func = match;
+      ap.target = erase_prefix(ap.func);
+      break;
+    case ProbeType::usdt:
+      // USDT probes specify a target binary path, a provider, and a function
+      // name.
+      ap.func = match;
+      ap.target = erase_prefix(ap.func);
+      ap.ns = erase_prefix(ap.func);
+      break;
+    case ProbeType::watchpoint:
+    case ProbeType::asyncwatchpoint:
+      // Watchpoint probes come with target prefix. Strip the target to get the
+      // function
+      ap.func = match;
+      erase_prefix(ap.func);
+      break;
+    case ProbeType::rawtracepoint:
+      ap.func = match;
+      break;
+    case ProbeType::software:
+    case ProbeType::hardware:
+    case ProbeType::interval:
+    case ProbeType::profile:
+    case ProbeType::special:
+    case ProbeType::iter:
+    case ProbeType::invalid:
+      break;
+    default:
+      LOG(BUG) << "Unknown probe type";
+  }
+  return ap;
+}
+
+std::string AttachPoint::name() const
 {
   std::string n = provider;
-  if (attach_target != "")
-    n += ":" + attach_target;
+  if (target != "")
+    n += ":" + target;
   if (lang != "")
     n += ":" + lang;
   if (ns != "")
     n += ":" + ns;
-  if (attach_point != "") {
-    n += ":" + attach_point;
+  if (func != "") {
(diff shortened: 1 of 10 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -2709,29 +2699,20 @@
...
+          attach_point->set_index(getNextIndexForProbe());
 
-          // Ensure that the full probe name used is the resolved one for this
-          // probe.
-          attach_point->target = target;
-          attach_point->ns = ns;
-          probefull_ = attach_point->name(func_id);
+        AttachPoint match_ap = attach_point->match_override(match);
+        probefull_ = match_ap.name();
+        current_attach_point_ = &match_ap;
```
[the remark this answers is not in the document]
ajor: I think this is a genuine issue (for once!)

src/bpftrace.cpp
```diff
@@ -153,40 +161,24 @@ int BPFtrace::add_probe(ast::Probe &p)
...
         } else {
           // If we have a wildcard in the target path, we need to generate one
           // probe per expanded target.
           std::unordered_map<std::string, Probe> target_map;
           for (const auto &func : attach_funcs) {
-            std::string func_id = func;
-            std::string target = erase_prefix(func_id);
-            auto found = target_map.find(target);
+            ast::AttachPoint ap = attach_point->match_override(func);
+            ap.func = attach_point->func;
```
ajor, line 174: What's this line for? It seems like `match_override()` replaces a wildcarded function name (e.g. "read*line") with an actual matched function name ("readline"). But then this line restores the wildcarded name ("read*line")

src/ast/passes/codegen_llvm.cpp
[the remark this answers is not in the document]
viktormalik: Not really, b/c `current_attach_point_` is used from the two `generateProbe` calls in the same block (so the local variable will exist).

src/ast/ast.cpp
```diff
@@ -587,18 +587,67 @@ std::string opstr(const Unop &unop)
...
+    case ProbeType::rawtracepoint:
+      ap.func = match;
+      break;
+    case ProbeType::software:
+    case ProbeType::hardware:
+    case ProbeType::interval:
+    case ProbeType::profile:
+    case ProbeType::special:
+    case ProbeType::iter:
+    case ProbeType::invalid:
```
jordalgo, line 634: nit: Add a default to the switch that throws or does `LOG(BUG)` to account for any added PropType in the future

src/ast/ast.h
```diff
@@ -632,9 +632,9 @@ class AttachPoint : public Node {
   uint64_t func_offset = 0;
   bool ignore_invalid = false;
 
-  std::string name(const std::string &attach_point) const;
-  std::string name(const std::string &attach_target,
-                   const std::string &attach_point) const;
+  std::string name() const;
+
+  AttachPoint match_override(const std::string &match) const;
```
jordalgo, line 637: #bikeshedding: the name `match_override` sounds like it returns a bool. Maybe `create_ap_match` or just `create_match`

src/ast/passes/codegen_llvm.cpp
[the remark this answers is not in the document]
jordalgo: But why not make it less confusing/error-prone and just pass the reference directly into `generateProbe` so it doesn't have to access `current_attach_point_` member var?

jordalgo:
Overall it LGTM. Relying on the tests here to catch differences in functionality.

src/bpftrace.cpp
ajor: Is `ap.func` actually used anywhere for uprobe_multi? Might be clearer to just zero it out if not

src/ast/ast.cpp
ajor: We normally rely on the compiler to generate warnings/errors when there's a missing switch case. Although it looks like Werror has been turned off in CI since the move from Travis to GitHub actions... we need to get that back on!

src/bpftrace.cpp
viktormalik: It's used to generate the `Probe` name in `BPFtrace::generate_probe`. For now, that's only useful for debugging but the follow-up PRs will need the correct name to identify the correct BPF program.

src/ast/ast.h
viktormalik: Good point, it seems that using `match` is a bit confusing in this context. What about `expansion_copy` or `create_expansion`? Personally, I like the former as it clearly states that a copy is created.

ajor (approved):

src/ast/ast.cpp
viktormalik: Ok, I added a `LOG(BUG)` for this PR but if we enable `-Werror`, we should drop it (as well as from other enums) to let compiler statically check for missing switch cases.

src/ast/ast.h
viktormalik: Went with `create_expansion_copy` in the end. It's a bit verbose but expresses the intent the best.

jordalgo (approved):

PR #3135, opened 2024-04-23
Fix builtin includes to #include using /bpftrace/include

bernhardkaindl:
When testing bpftrace-0.19.1 on an old system based on CentOS 7, the included libclang-16.0.6 was not able to find `__btf_generated_header.h` and `clang_workarounds.h`

Prepending their directory `/bpftrace/include` (including them using their full path) fixed the issue. I wonder why this is, but I figured it might also make things clearer for the tip:

- Makes it clear that those are built-in headers
- Leaves no ambiguity which files are included.

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -663,7 +663,8 @@ bool ClangParser::parse(ast::Program *program,
   if (program->c_definitions.empty() && bpftrace.btf_set_.empty())
     return true;
 
-  input = "#include <__btf_generated_header.h>\n" + program->c_definitions;
+  input = "#include </bpftrace/include/__btf_generated_header.h>\n" +
+          program->c_definitions;
 
   input_files = getTranslationUnitFiles(CXUnsavedFile{
       .Filename = "definitions.h",
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -423,7 +423,7 @@ static std::optional<struct timespec> get_delta_taitime()
           utsname.machine, ksrc, kobj, bpftrace.kconfig);
   }
   extra_flags.push_back("-include");
-  extra_flags.push_back(CLANG_WORKAROUNDS_H);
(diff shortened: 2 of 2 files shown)


src/clang_parser.cpp
danobi, line 1: I think `args = { "-isystem", "/bpftrace/include" };` on line 675 can/should be removed, right? Would prevent backsliding.

danobi (commented):
Thanks for the PR! Sorry about the delay. This lgtm modulo a potential improvement in clang_parser.cpp

src/clang_parser.cpp
bernhardkaindl: Thanks for the quick review!
I assume that `args = { "-isystem", "/bpftrace/include" };` should not be removed:

There is a Vector of unsaved_files at line 22: https://github.com/bpftrace/bpftrace/blob/c8fd334f55e6b04a7eb3fc9a06822fc6b2a6344b/src/clang_parser.cpp#L29

I'm guessing the `args = { "-isystem", "/bpftrace/include" };` is needed to find those unsaved files.

My libclang-16.0.6 has an issue with include paths (it only seems to honor only with one `-I<includedir>` path), so including `__btf_generated_header.h` and `clang_workarounds.h` using an absolute path helped in my case. I'm in the process of upgrading the packages to llvm18, I hope that this solves these include path woes that I ran into.

Nonetheless, I wanted to submit this anyway, so the source makes clear where those files are.

danobi: I think the `-isystem` only adds `/bpftrace/include` to the header search path. But given all the `unsaved_files` use absolute paths already, I think it should be ok to remove `-isystem`. I applied https://pastes.dxuuu.xyz/k3m9io locally and ran ~1/3 of the runtime tests and it seems to be ok.

Pls update the PR with the change and let's see if the CI passes

danobi:
As-is is fine. I can follow up with my suggestion

PR #3137, opened 2024-04-25
Fix listing type definitions

viktormalik:
#3111 broke listing of struct definitions:

    # bpftrace -l 'struct task_struct'
    ERROR: failed to open file 'struct task_struct': No such file or directory

This fixes the problem and introduces a runtime test to avoid future breakage.

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,6 +39,7 @@
 #include "probe_matcher.h"
 #include "procmon.h"
 #include "tracepoint_format_parser.h"
+#include "utils.h"
 #include "version.h"
 
 using namespace bpftrace;
@@ -671,12 +672,14 @@ Args parse_args(int argc, char* argv[])
       args.search = "*:*";
     } else if (optind == argc - 1) {
       std::string_view val(argv[optind]);
-      if (val == "*") {
-        args.search = "*:*";
-      } else if (val.find_first_of(":*") != std::string::npos) {
-        args.search = val;
-      } else {
+      if (std_filesystem::exists(val)) {
         args.filename = val;
+      } else {
+        if (val == "*") {
+          args.search = "*:*";
+        } else {
+          args.search = val;
+        }
       }
       optind++;
     } else {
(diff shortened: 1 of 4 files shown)


viktormalik:
Since the breaking commit has not been released, yet, I don't think that it's necessary to pollute CHANGELOG with this fix.

ajor:
Oops, thanks for catching this!

@jordalgo and I had been discussing approaches for identifying scripts names vs probes a fair bit but didn't think about types.

We were also wondering if it might also make sense to just use "file exists" as the determinator instead of trying to pattern match probe names. With type names added in too, I think that approach makes even more sense. e.g. what should we do if someone tries running `bpftrace -l struct_sizes.bt` and that file exists?

jordalgo:
> With type names added in too, I think that approach makes even more sense. e.g. what should we do if someone tries running bpftrace -l struct_sizes.bt and that file exists?

+ 1. `std::filesystem::exists` seems like the cleanest way out of this but also we can probably discount strings that contain spaces.

viktormalik:
> what should we do if someone tries running `bpftrace -l struct_sizes.bt` and that file exists?

Fair point. I was thinking about checking for "struct " prefix (notice the trailing space) which would not eliminate but greatly decrease the probability of this happening (because who uses spaces in file names, right? ;-))

Checking for file existence makes sense, too, only the error message will have to be something like

    ERROR: search entry 'foobar' is not a valid probe or file

viktormalik:
Let's use `std::filesystem::exists`, I'll update this PR.

viktormalik:
Ok, here's the implemented idea:
- if the search entry is an existing file, it is interpreted as such,
- if the search entry starts with "struct ", "union ", or "enum " (notice the trailing spaces), it is interpreted as a type name,
- otherwise the search entry is interpreted as a probe. In this case, when the search entry contains a dot ('.'), a warning is shown that such a file does not exist.

src/main.cpp
```diff
@@ -818,15 +821,18 @@ int main(int argc, char* argv[])
...
-         args.search.find("enum") == 0)) {
+    if (is_type_name(args.search)) {
       // Print structure definitions
       bpftrace.parse_btf({});
       bpftrace.probe_matcher_->list_structs(args.search);
       return 0;
     }
 
+    if (args.search.find(".") != std::string::npos) {
+      LOG(WARNING) << "It appears that \'" << args.search
```
jordalgo, line 832: Do we have probes that contain `.` ? I'm just wondering why not exit here with an `ERROR` or `ABORT`.

viktormalik: We do, e.g. `uprobe:./a.out:*`. Now thinking about it, we should go with a stricter condition: contains `.` and does not contain neither `:` of `*`. Then, we can probably go with `LOG(ABORT)`, too.

viktormalik: In the end, I decided to stay on the safe side and continue with the execution, so kept `LOG(WARNING)` there.

ajor (approved):

src/main.cpp
jordalgo: Then maybe updating the warning message slightly would be good e.g. 
"It appears that X is a filename but the file does not exist. Proceeding as if X is a search pattern".

viktormalik: Good idea, I adjusted it a bit to be more compact:
```
WARNING: It appears that 'script.bt' is a filename but the file does not exist. Treating 'script.bt' as a search pattern.
```

jordalgo (approved):

PR #3138, opened 2024-04-25
Re-enable and fix iterator probe tests

viktormalik:
When playing around with iterator probes, I noticed that the runtime tests for them are never run, b/c the `REQUIRE_FEATURE` field was incorrect. When fixed, running the tests revealed one bug in the codegen (a missing int->ptr cast).

This PR fixes the bug and re-enables the tests.

##### Checklist


viktormalik:
The bug only reveals when the `BPFTRACE_VERIFY_LLVM_IR` env var is set or in runtime tests (where it is set by default) so this is actually not a user-facing change. So, I'm skipping the CHANGELOG entry here.

ajor (approved):

jordalgo:
Looks like we're still getting test failures for the type mismatch.

viktormalik:
> Looks like we're still getting test failures for the type mismatch.

Yup, one more pointer bitcast necessary for older LLVMs. Should be fine now.

jordalgo (approved):

PR #3149, opened 2024-05-04
Fix bpftrace for older kernels: Fallback to no func_infos, no BTF

### LINKED ISSUE #3011: bpftrace v0.20.1, Linux 4.19 - ERROR: Error loading program

The latest version (v0.20.1) of bpftrace run against a 4.19 kernel (specifying the path to the BTF file through the `BPFTRACE_BTF` environment variable) fails to load (apparently any) program. v0.19.1 works fine in the same environment. I managed to bisect the issue as being caused by https://github.com/bpftrace/bpftrace/pull/2804.

```
BPFTRACE_BTF="/boot/btf-$(uname -r)" bpftrace -e 'kprobe:do_nanosleep { printf("hi\n"); }' -v

INFO: node count: 6
Attaching 1 probe...

Error log:
magic: 0xeb9f
version: 1
flags: 0x0
hdr_len: 24
type_off: 0
type_len: 76
str_off: 76
str_len: 78
btf_total_size: 178
[1] PTR (anon) type_id=2
[2] INT int8 size=1 bits_offset=0 nr_bits=8 encoding=SIGNED
[3] Invalid kind:13
ERROR: Error loading program: kprobe:do_nanosleep
```

I'm running a 4.19 VM through https://github.com/cilium/little-vm-helper:

```
lvh images pull --cache quay.io/lvh-images/kind:4.19-20240201.165956 --dir images
lvh run --image images/kind_4.19.qcow2 --daemonize -p 2222:22 --serial-port 0 --cpu=4 --mem=6G --cpu-kind host --console-log-file /tmp/console.log
```

### `bpftrace --info` output

```
System
  OS: Linux 4.19.306 #1 SMP Thu Feb 1 17:01:42 UTC 2024
  Arch: x86_64

Build
  version: v0.20.1
  LLVM: 17.0.6
  unsafe probe: no
  bfd: no
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: no
  probe_read_user_str: no
  probe_read_kernel: no
  probe_read_kernel_str: no
  get_current_cgroup_id: yes
  send_signal: no
  override_return: yes
  get_boot_ns: no
  dpath: no
  skboutput: no
  get_tai_ns: no
  get_func_ip: no
  jiffies64: no

Kernel features
  Instruction limit: 131072
  Loop support: no
  btf: yes
  module btf: no
  map batch: no

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: no

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: no
  kprobe_multi: no
  uprobe_multi: no
  raw_tp_special: no
  iter: no
```

bernhardkaindl:
Fix #3011:

In https://github.com/bpftrace/bpftrace/issues/2913#issuecomment-1888962724, @viktormalik said:
> So the problem is that the older kernel doesn't cope well with BTF and func_infos.

He detailed the problem in https://github.com/bpftrace/bpftrace/issues/2913#issuecomment-1892053795:
> So, I was able to find the problem - older kernels do not have https://github.com/torvalds/linux/commit/51c39bb1d5d105a02e29aa7960f0a395086e6342 and therefore do not support BTF_KIND_FUNC entries with global linkage. Libbpf sanitizes this for older kernels since https://github.com/torvalds/linux/commit/2d3eb67f64ec317bbfe340cfcc8325d40a6ff317 (same series) but it only does that when the program is loaded via bpf_object (which we don't at this point).

He implemented an initial interim fix in #2934 (as loaded via bpf_object) might be a bigger change, but #3011 shows that this is not enough. In it, @giorio94 bisected that #2804 causes `bpftrace` to fail on all kernels that do not support the types that LLVM now generates. And func_infos fail with 4.19 fail too.

I now found that the issue is that current code skips loading the BPF program when loading the BTF into the kernel failed: https://github.com/bpftrace/bpftrace/blob/c8fd334f55e6b04a7eb3fc9a06822fc6b2a6344b/src/attached_probe.cpp#L831
```py
        // Don't attempt to load the program if the BTF load failed.
        // This will fall back to the error handling for failed program load,
        // which is more robust.
```
But there is no such fall back to handle this error yet, so add it. And, I found:
- As @viktormalik said, I also found that older kernels don't cope well with func_infos, so fall back no  func_infos relocation on the 2nd attempt.
- I was able to fix loading the BTF load on for my setup, but 4.19 still rejects loading the BPF program, so even if loading the BTF succeeds, loading the BPF it does not work, fall back to not loading the BTF (as a last resort) in the 3rd attempt.

These changes fix bpftrace to work on 4.19 again.

I plan also submit my BTF fixup changes to fix BTF load on 4.19, they add converting the kinds FUNC, FUNC_PROTO, VAR, and DATASEC. But as said, these only fix BTF load but passing the btf_fd for the loaded BTF still makes the BPF load fail on 4.19.

For now, this is the smallest change and only adds fallbacks for older kernels, that have issues with BTF and func_infos so it does not affect newer kernels at all.

Of course, the best solution is to load everything by using libbpf as bpf_object, but that is a bigger change.

And, I'd also like showing the error code and error message when BPF load fails (added as well).

### DIFF
--- a/src/ast/elf_parser.cpp
+++ b/src/ast/elf_parser.cpp
@@ -1,5 +1,8 @@
 #include "elf_parser.h"
 #include <elf.h>
+#ifndef EM_BPF     // EM_BPF may not be defined in older containers, e.g. RHEL7
+#define EM_BPF 247 // Could be in linux/elf.h, but it can cause conflicts
+#endif
 
 namespace bpftrace {
 namespace elf {
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -707,13 +707,32 @@ void AttachedProbe::load_prog(BPFfeature &feature)
   auto func_infos = prog_.getFuncInfos();
   const char *license = "GPL";
   int log_level = 0;
+  std::stringstream logbuf; // while stderr is silenced and for joining lines
 
   uint64_t log_buf_size = probe_.log_size;
   auto log_buf = std::make_unique<char[]>(log_buf_size);
   std::string name;
   std::string tracing_type;
 
-  {
+  /*
+   * Fallbacks for kernels that don't cope well with func_infos and BTF, and may
+   * check the program's kernel version. Try witout BTF and func_infos and when
+   * needed, also pass the kernel version (try three different methods):
+   */
+  struct load_strategy {
+    bool btf;
+    bool func_infos;
+    KernelVersionMethod kver;
+  } load_configs[] = {
+    { .btf = true, .func_infos = true, .kver = None },  // BTF and func_infos
+    { .btf = true, .func_infos = false, .kver = None }, // BTF, no func_infos
+    /* Fallbacks for Linux <5.0: No BTF and func_infos, with kernel version: */
+    { .btf = false, .func_infos = false, .kver = vDSO }, // version from vDSO
+    { .btf = false, .func_infos = false, .kver = UTS },  // version from uname
+    { .btf = false, .func_infos = false, .kver = File }, // version from header
+  };
+
+  for (auto &load_config : load_configs) {
     if (bt_debug != DebugLevel::kNone)
       log_level = 15;
     if (bt_verbose)
@@ -738,16 +757,7 @@ void AttachedProbe::load_prog(BPFfeature &feature)
     if (probe_.type == ProbeType::special && !feature.has_raw_tp_special())
       prog_type = progtype(ProbeType::uprobe);
 
-    for (int attempt = 0; attempt < 3; attempt++) {
-      auto version = kernel_version(attempt);
-      if (version == 0 && attempt > 0) {
-        // Recent kernels don't check the version so we should try to call
-        // bpf_prog_load during first iteration even if we failed to determine
-        // the version. We should not do that in subsequent iterations to avoid
-        // zeroing of log_buf on systems with older kernels.
-        continue;
-      }
-
+    {
       LIBBPF_OPTS(bpf_prog_load_opts, opts);
       opts.log_buf = log_buf.get();
       opts.log_size = log_buf_size;
@@ -802,7 +812,14 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         opts.attach_btf_id = btf_id.first;
         opts.attach_btf_obj_fd = btf_id.second;
       } else {
-        opts.kern_version = version;
+        opts.kern_version = kernel_version(load_config.kver);
+      }
+      logbuf << "load " << probe_.name << (load_config.btf ? ", with BTF" : "")
+             << (load_config.func_infos ? ", with func_infos" : "");
+      if (opts.kern_version) {
+        logbuf << ", version: " << ((opts.kern_version >> 16) & 0xFF) << "."
+               << ((opts.kern_version >> 8) & 0xFF) << "."
+               << (opts.kern_version & 0xFF);
       }
 
       {
@@ -811,35 +828,41 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         if (bt_debug == DebugLevel::kNone)
           silencer.silence();
 
-        LIBBPF_OPTS(bpf_btf_load_opts,
-                    btf_opts,
-                    .log_buf = log_buf.get(),
-                    .log_level = static_cast<__u32>(log_level),
-                    .log_size = static_cast<__u32>(log_buf_size), );
+        int btf_fd = -1;
+        if (load_config.btf) {
+          LIBBPF_OPTS(bpf_btf_load_opts,
+                      btf_opts,
+                      .log_buf = log_buf.get(),
+                      .log_level = static_cast<__u32>(log_level),
+                      .log_size = static_cast<__u32>(log_buf_size), );
 
-        auto &btf = prog_.getBTF();
-        auto btf_fd = bpf_btf_load(btf.data(), btf.size(), &btf_opts);
+          auto &btf = prog_.getBTF();
+          btf_fd = bpf_btf_load(btf.data(), btf.size(), &btf_opts);
+          if (btf_fd < 0) {
+            logbuf << ": BTF load failed";
+            goto log_error; // and continue to the next load_config fallback
+          }
+          opts.prog_btf_fd = (__u32)btf_fd;
+        }
 
-        opts.prog_btf_fd = btf_fd;
-
-        if (!func_infos.empty()) {
+        // Skip .BTF.ext func_info relocation on fallback for older kernels
+        if (load_config.func_infos && !func_infos.empty()) {
           opts.func_info_rec_size = sizeof(struct bpf_func_info);
           opts.func_info = func_infos.data();
           opts.func_info_cnt = func_infos.size() / sizeof(struct bpf_func_info);
         }
 
-        // Don't attempt to load the program if the BTF load failed.
-        // This will fall back to the error handling for failed program load,
-        // which is more robust.
-        if (btf_fd >= 0) {
+        // Load the BPF program on BTF load success or on fallback without it
+        if (!load_config.btf || btf_fd >= 0) {
           progfd_ = bpf_prog_load(static_cast<::bpf_prog_type>(prog_type),
                                   name.c_str(),
                                   license,
                                   reinterpret_cast<const struct bpf_insn *>(
                                       insns.data()),
                                   insns.size() / sizeof(struct bpf_insn),
-                                  &opts);
-          close(btf_fd);
+                                  &opts); // Configured opts are passed here
+          if (btf_fd >= 0)
+            close(btf_fd);
         }
       }
 
@@ -847,7 +870,17 @@ void AttachedProbe::load_prog(BPFfeature &feature)
         close(opts.attach_btf_obj_fd);
       if (progfd_ >= 0)
         break;
-    }
+      logbuf << " = " << progfd_ << ": " << std::strerror(-progfd_);
+    log_error:
+      if (!logbuf.str().empty()) { // log the error of the failed attempt
+        LOG(V1) << logbuf.str();
+        logbuf.str("");
+      }
+    } // obsolete indendation level, has no effect, kept for a small commitdiff
+  } // for (auto &load_config : load_configs): handle kernel features & versions
+
+  if (!logbuf.str().empty()) {
+    LOG(V1) << logbuf.str() << ": Success";
   }
 
   if (progfd_ < 0) {
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -1289,28 +1289,27 @@ static uint32_t kernel_version_from_khdr(void)
  * may not match if bpftrace is compiled on a different Linux version than it's
  * used on, e.g. if built with Docker.
  */
-uint32_t kernel_version(int attempt)
+uint32_t kernel_version(KernelVersionMethod method)
 {
   static std::optional<uint32_t> a0, a1, a2;
-  switch (attempt) {
-    case 0: {
+  switch (method) {
+    case vDSO: {
       if (!a0)
         a0 = kernel_version_from_vdso();
       return *a0;
     }
-    case 1: {
+    case UTS: {
       if (!a1)
         a1 = kernel_version_from_uts();
       return *a1;
     }
-    case 2: {
+    case File: {
       if (!a2)
         a2 = kernel_version_from_khdr();
       return *a2;
     }
-    default:
-      LOG(BUG) << "kernel_version(): Invalid attempt: "
-               << std::to_string(attempt);
+    case None:
+      return 0;
   }
 }
 
--- a/src/utils.h
+++ b/src/utils.h
@@ -280,7 +280,8 @@ T read_data(const void *src)
   return v;
 }
 
-uint32_t kernel_version(int attempt);
+enum KernelVersionMethod { vDSO, UTS, File, None };
+uint32_t kernel_version(KernelVersionMethod);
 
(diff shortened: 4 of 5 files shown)


danobi:
I think to think more deeply about edge cases here, but on the surface this seems reasonable in isolation.

However, I think @viktormalik is/has been working a lot on the libbpf migration. If it's on-track to land this release cycle, then it'd probably be better to wait for that (given this PR or the libbpf migration will ship at the same time).

viktormalik:
> I think to think more deeply about edge cases here, but on the surface this seems reasonable in isolation.
> 
> However, I think @viktormalik is/has been working a lot on the libbpf migration. If it's on-track to land this release cycle, then it'd probably be better to wait for that (given this PR or the libbpf migration will ship at the same time). Viktor, can you comment?

Yes, I've a working implementation now and will start posting PRs. It's a lot of code so I'll need to split it into multiple parts. See #2334 for a preliminary plan of updates.

I'm not sure if it's a good idea to land this in this release cycle, though. Since it's quite a disruptive change, I'd prefer having it in unreleased state to have as many people use it in their workflows as possible. On top of that, we haven't done a release in a while and already have a nice set of unreleased features, so maybe let's do a release soon and plan this for the next one?

bernhardkaindl:
Thanks @viktormalik for the insight and sharing your plans for having it in unreleased state to have as many people use it in their workflows as possible and rather do a release, already having a nice set of unreleased features without your big disruptive change soon.

That would fall right in line with my use case: I'd like to first integrate a stable release of bpftrace that works across all supported kernels, including 4.19. The reason is that 4.19 is used in production, where bpftrace would be very useful to have. If the upstream bpfstrace master branch works with 4.19, I can deploy it directly and can contribute to the bpftrace master branch code base with PRs as I'm able to address any findings that I'd see directly on master. If the upstream bptrace master branch takes the stance that it does not wish to integrate fixes that allow me and my users to use it on our production environments running 4.19, then I'm faced with having to create a bpftrace-4.19-compat fork for tracing in the actively used production environment. Of course, I'd rather prefer to "live" on master if I can use it directly.

Also, having bpftrace master work on 4.19 would be a decisive signal as it shows that there is a future outlook that working within the upstream community is viable and beneficial (I hope in both directions) to share the work. If the decision is going to delay and instead to only support 4.19 when it comes for free with using libbpf objects, it says that bpftrace upstream only tracks if it works on 4.19 or not, but does not apply fixes for it. That would send a signal to rather stay on the bpftrace-0.19.1 branch that supports 4.19 rather than master, and any upstream work would unfortunately not be used or tested by my users.

Please see this just as a user sharing his case, not more. I'm merely asking for a decision to be made in order to make my plans for using a branch or fork in 0.19.1 or work directly on upstream master regarding me and my users.

src/attached_probe.cpp
```diff
@@ -811,48 +811,56 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
+          LIBBPF_OPTS(bpf_btf_load_opts,
+                      btf_opts,
+                      .log_buf = log_buf.get(),
+                      .log_level = static_cast<__u32>(log_level),
+                      .log_size = static_cast<__u32>(log_buf_size), );
+
+          auto &btf = prog_.getBTF();
+          btf_fd = bpf_btf_load(btf.data(), btf.size(), &btf_opts);
+          if (btf_fd >= 0)
+            opts.prog_btf_fd = (__u32)btf_fd;
```
viktormalik, line 827: Could we break the outer (attempt) loop if `btf_fd < 0` here?

src/attached_probe.cpp
```diff
@@ -811,48 +811,56 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
-        if (btf_fd >= 0) {
-          progfd_ = bpf_prog_load(static_cast<::bpf_prog_type>(prog_type),
-                                  name.c_str(),
-                                  license,
-                                  reinterpret_cast<const struct bpf_insn *>(
-                                      insns.data()),
-                                  insns.size() / sizeof(struct bpf_insn),
-                                  &opts);
+        // Most bpftrace programs still work without func_infos and without BTF.
+        // See the comments above for reference(these are the only retries)
```
viktormalik, line 840: This comment doesn't make much sense here (both BTF and func_info is passed via `opts` so it's not used in the below command). I'd rather see a more general comment somewhere (maybe before the "attempt" loop header?) explaining what features are omitted in which attempt.

viktormalik (commented):
I think that it would generally be useful to have a support for the 4.19 stable kernel in master (and the next release). Then, when we switch to libbpf-based loading which should (theoretically) be backwards compatible, we'll have a baseline to compare with.

The PR itself LGTM, just a few comments below.

Also, we historically used the loading "attempts" for older kernels which required the kernel version (and we had 3 different ways of getting it). I'm wondering if this change could potentially conflict with that.

src/attached_probe.cpp
bernhardkaindl: Yes, I hadn't added the outer loop yet, but I added it now. Done!

src/attached_probe.cpp
bernhardkaindl: Yes, that's best, done now with the new outer loop!

bernhardkaindl:
Thanks @viktormalik!
> Also, we historically used the loading "attempts" for older kernels which required the kernel version (and we had 3 different ways of getting it). I'm wondering if this change could potentially conflict with that.

Thanks for the hint! I did not know, and I overlooked that! Yes, I fixed that issue with this push!
> Could we break the outer (attempt) loop if btf_fd < 0 here?

Yes, I hadn't added the outer loop yet, but I added it now. Done!

> I'd rather see a more general comment somewhere (maybe before the "attempt" loop header?) explaining what features are omitted in which attempt.

Yes, that makes total sense! Done on top of the new outer loop!

src/attached_probe.cpp
```diff
@@ -713,7 +713,17 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
-  {
+  // Fallback attempts for kernels that don't cope well with func_infos and BTF
+  struct load_strategy {
+    bool load_btf;
+    bool func_infos;
+  } load_configs[] = {
+    { .load_btf = true, .func_infos = true },   // 1st, try both
+    { .load_btf = true, .func_infos = false },  // 2nd, without func_infos
+    { .load_btf = false, .func_infos = false }, // 3rd, without btf load as well
+  };
```
viktormalik, line 724: I like this approach and I think that we should combine it with the loop for version attempts. Since kernel version checking was removed in https://github.com/torvalds/linux/commit/6c4fc209fcf9 (kernel version 5.0) and, IIUC, this PR is for 4.19, I think that the following sequence of attempts would make sense:
1. BTF + func_infos + no version
2. BTF + no func_infos + no version
3. no BTF + no func infos + version attempt 0
4. no BTF + no func infos + version attempt 1
5. no BTF + no func infos + version attempt 2

That should spare some `bpf_prog_load` calls and also make the semantics of individual attempts much clearer.

src/ast/elf_parser.cpp
```diff
@@ -1,5 +1,8 @@
 #include "elf_parser.h"
 #include <elf.h>
+#ifndef EM_BPF
+#define EM_BPF 247 // Could be in linux/elf.h, but it can cause conflicts
+#endif
```
viktormalik, line 5: Why is this necessary?

src/attached_probe.cpp
bernhardkaindl: Indeed. Done.

To make the version attempts obvious for the uninitiated, I named them using an enum, I hope you like it.

I hope you will like also logging of the retries:
I added a bit of logging for the retries and added a `std::stringstream` logbuf to append the info for each attempt into one line (and it is also needed to log inside the BTF/BTF load context as in it stderr is silenced by the silencer).

I guess that a logbuf for such situations may already exist, but I submit it now as-is. It helped me to confirm that the sequence is indeed executed as specified.

This example log is from different sequence, crafted to cover all added lines in this test sequence on the test host I used:
```
load tracepoint:raw_syscalls:sys_enter, with BTF, with func_infos: BTF load failed
load tracepoint:raw_syscalls:sys_enter, with BTF, version: 4.19.0: BTF load failed
load tracepoint:raw_syscalls:sys_enter, with func_infos = -7: Argument list too long
load tracepoint:raw_syscalls:sys_enter, version: 4.19.0: Success
```

As hinted at the end of the loop, I didn't de-indent the now context and indentation level that is now obsolete due to the removal of the inner kernel_version retriy loop:

I kept the context and indentation only to keep the commitdiff as small as possible. Of course it may be useful to de-indent that indentation level as a final step before or after the code review.

src/ast/elf_parser.cpp
bernhardkaindl: The older systems that I test on don't define it. It only used in the ELF parser to ensure that the Elf Machine is BPF. I was thinking it could be added as a related change, but I can also submit it separately. It just makes it easier for me to build the master branch without having to patch this file.
```c
+#ifndef EM_BPF // EM_BPF may not be defined in older containers, e.g. RHEL7
```

src/utils.cpp
```diff
@@ -1289,29 +1289,28 @@ static uint32_t kernel_version_from_khdr(void)
...
       if (!a2)
         a2 = kernel_version_from_khdr();
       return *a2;
     }
-    default:
-      LOG(BUG) << "kernel_version(): Invalid attempt: "
-               << std::to_string(attempt);
+    case None:;
   }
+  return 0;
```
viktormalik, line 1313: Should be under `case None:`

viktormalik (commented):
LGTM overall, just a couple of comments below.

src/ast/elf_parser.cpp
viktormalik: I see, makes sense.

src/attached_probe.cpp
```diff
@@ -707,13 +707,33 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
+  // To collect logs if stderr is silenced and to merge infos into one log line
+  std::stringstream logbuf;
 
   uint64_t log_buf_size = probe_.log_size;
   auto log_buf = std::make_unique<char[]>(log_buf_size);
   std::string name;
   std::string tracing_type;
 
-  {
+  // Fallback attempts for kernels that don't cope well with func_infos and BTF
```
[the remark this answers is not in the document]
bernhardkaindl: Thanks, applied.

src/utils.h
```diff
@@ -280,7 +280,8 @@ T read_data(const void *src)
   return v;
 }
 
-uint32_t kernel_version(int attempt);
+enum get_kernel_version { vDSO, UTS, File, None };
```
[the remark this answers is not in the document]
bernhardkaindl: Thanks, applied.

src/attached_probe.cpp
```diff
@@ -707,13 +707,33 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
+    { .btf = false, .func_infos = false, .kver = vDSO }, // version from vDSO
+    { .btf = false, .func_infos = false, .kver = UTS },  // version from uname
+    { .btf = false, .func_infos = false, .kver = File }, // version from header
+  };
+
+  for (auto &load_config : load_configs) {
+    if (!logbuf.str().empty()) { // In case of a retry, log the failed attempt
+      LOG(V1) << logbuf.str();
+      logbuf.str("");
+    }
```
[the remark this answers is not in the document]
bernhardkaindl: Yes indeed. Thanks, done.

src/utils.cpp
bernhardkaindl: Applied.

bernhardkaindl:
@viktormalik:
> LGTM overall, just a couple of comments below.

Thanks, applied the changes.

src/attached_probe.cpp
```diff
@@ -811,48 +827,66 @@ void AttachedProbe::load_prog(BPFfeature &feature)
...
+
+  if (!logbuf.str().empty()) {
+    LOG(V1) << logbuf.str() << ": Success";
   }
 
   if (progfd_ < 0) {
     if (bt_verbose) {
       std::cerr << std::endl
+                << "bpf_prog_load() failed with error " << progfd_ << ": "
+                << std::strerror(-progfd_) << std::endl
```
viktormalik, line 889: Is this necessary now when the same is printed for each failed attempt?

bernhardkaindl: Thanks, correct, that is no longer necessary as it is already logged for each failed attempt!

Removed them! - I backed out these two lines from the commit, no other changes in this push.

viktormalik (approved):
LGTM, thanks!

Let's wait for #3166 to land and rebase on top of that to make sure that the tests pass.

bernhardkaindl:
BTW: GitHub implicitly re-bases the PR git checkout for the CI builds onto the target branch automatically. That is the reason why all PRs broke when master broke CI.

bernhardkaindl:
@viktormalik, as https://github.com/bpftrace/bpftrace/releases/tag/v0.20.4 is released, and you assigned [v0.21.0](https://github.com/bpftrace/bpftrace/milestone/19) as milestone, and you reviewed it, what is the next milestone to merge this PR?

viktormalik:
@bernhardkaindl I want to get this merged asap, to get it into the upcoming 0.21 release. Just waiting for a review from @danobi.

danobi (approved):

PR #3152, opened 2024-05-06
Fix tcpconnect.bt to not need libc headers installed if BTF is available

bernhardkaindl:
Fix `tcpconnect.bt` to not need libc headers installed if BTF is available:

With BTF, BTF provides the needed types, we just need to define `AF_INET` and `AF_INET6`

`AF_INET` and `AF_INET6` are Linux ABI defines, and are not architecture-specific.

This allows `tcpconnect.bt` to work without libc headers installed at runtime and reach host to use it.


viktormalik (approved):
This is a good idea, we've had a couple of issues open for `tcpconnect.bt`.

Thanks!

bernhardkaindl:
Thanks, I thought the same. I pushed applying this update (or fix) to the other tcp tools now as well.

viktormalik (approved):

bernhardkaindl:
@ajor Of course, I'm giving permission to relicense my changes in this PR to GPL v2!

This applies to all my submissions, including #3135 and especially #3149 (that is critical to me)!

ajor (approved):
Great, thanks so much!

PR #3153, opened 2024-05-06
Make codegen compatible with libbpf

viktormalik:
This adds two changes to codegen which make the produced ELF files compatible with what libbpf expects:
- libbpf identifies BPF programs by LLVM function names rather than by section names (as we do in bpftrace now) so sanitize the function name and unify it with function name. In particular, the LLVM function name is now `<attach_point_name>_<usdt_index>_<probe_index>` (where the `_<usdt_index>` part is optional) and the section name just prepends `s_` to it.
- BPF program license (GPL) is stored in codegen as a global string variable in the `license` section:
```llvm
@LICENSE = global [4 x i8] c"GPL\00", section "license"
```

##### Checklist


chenhengqi:
Can you also make the section name compatible with libbpf ? Last time I played with veristat but it can't recognize the ELF file generated. Thanks.

viktormalik:
> Can you also make the section name compatible with libbpf ? Last time I played with veristat but it can't recognize the ELF file generated. Thanks.

Since the section name is now just `"s_" + function_name`, it is compatible, too.

chenhengqi:
> > Can you also make the section name compatible with libbpf ? Last time I played with veristat but it can't recognize the ELF file generated. Thanks.
> 
> Since the section name is now just `"s_" + function_name`, it is compatible, too.

I haven't tried this patch, but I don't think so. Libbpf expects canonical section names like "uprobe", "kprobe" etc.

viktormalik:
> I haven't tried this patch, but I don't think so. Libbpf expects canonical section names like "uprobe", "kprobe" etc.

Ah, sorry, I misunderstood your comment.

Yes, libbpf uses canonical section names for the purpose of auto-attachment of the BPF programs. Since we don't use libbpf to do the attachment, it is not necessary for us to use these. I have that on my todo list after #2334 is done.

On the other hand, if veristat requires the canonical section names and it makes sense to use veristat for bpftrace-generated ELFs, we may want to implement this independently. I wouldn't add it to this PR, though, since it may require more work than anticipated. For example, we support the `kfunc:module:function` syntax which was not supported by libbpf until very recently (and there may be more such cases). Once this PR is merged, I'll open an issue to track that.

danobi (approved):

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2506,11 +2515,14 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+
   int index = current_attach_point_->index() ?: probe.index();
-  func->setSection(
-      get_section_name_for_probe(section_name, index, usdt_location_index));
+  auto func_name = get_function_name_for_probe(name,
+                                               index,
+                                               usdt_location_index);
+  Function *func = Function::Create(
+      func_type, Function::ExternalLinkage, func_name, module_.get());
+  func->setSection("s_" + func_name);
```
jordalgo, line 2525: nit: should we have a wrapper around this so this prefix addition is centralized?

viktormalik: That's a good idea, we may want to move to libbpf canonical section names one day and this will make the change clearer. Done.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3292,13 +3304,11 @@ void CodegenLLVM::generateWatchpointSetupProbe(
...
-                                    get_watchpoint_setup_probe_name(
-                                        expanded_probe_name),
-                                    module_.get());
-  func->setSection(
-      get_section_name_for_watchpoint_setup(expanded_probe_name, index));
+  auto func_name = get_function_name_for_watchpoint_setup(expanded_probe_name,
+                                                          index);
+  Function *func = Function::Create(
+      func_type, Function::ExternalLinkage, func_name, module_.get());
+  func->setSection("s_" + func_name);
```
jordalgo, line 3311: missed one ;)

viktormalik: Many thanks for checking and fixing my careless work! Should be fixed now.

tests/codegen/llvm/struct_semicolon_1.ll
ajor, line 1: Rebase issue? This file got deleted in #3129

tests/codegen/llvm/struct_semicolon_2.ll
ajor, line 1: Also deleted

tests/codegen/llvm/struct_semicolon_1.ll
viktormalik: Indeed, removed, thanks!

ajor (approved):
Nice!

jordalgo (approved):
LGTM!

PR #3154, opened 2024-05-06
Fix bug in not overwriting compromised path

jordalgo:
The potentially compromised temporary kheaders path was not being overwritten by `rename`. This meant that the compromised path was still being used and read from.

Instead of continuing just abort and direct the user to delete this compromised path.

Reported here: https://github.com/bpftrace/bpftrace/pull/3033

##### Checklist


src/utils.cpp
```diff
@@ -717,6 +703,7 @@ struct KernelHeaderTmpDir {
 
   void move_to(const std::string &new_path)
   {
+    exec_system(("rm -rf " + new_path).c_str());
```
viktormalik, line 706: Could we use `std::filesystem::remove_all`? Feels nicer than `exec_system`.

src/utils.cpp
```diff
@@ -683,20 +683,6 @@ bool is_dir(const std::string &path)
   return std_filesystem::is_directory(buf, ec);
 }
 
-bool file_exists_and_ownedby_root(const char *f)
```
viktormalik, line 686: Why not keep the function and the test?

jordalgo: Mostly because this function is only used in one place and the control flow didn't quite work to keep this function around e.g. it's ok to continue if the file doesn't exit but not ok if it does exist and is not owned by root.

viktormalik: Right, that makes sense.

ajor (changes requested):
I think we need to remove this unpacking-header logic entirely.

As I said previously, this type of ownership detection is vulnerable to edge cases we haven't though of (as this latest fix shows).  The safest thing to do is just require users to pass in a path to unpacked headers. We shouldn't have bpftrace reading from globally-writable directories without the user explicitly telling it to or we're risking more vulnerabilities in the future.

Also shelling out to `tar` is nasty.

viktormalik:
> I think we need to remove this unpacking-header logic entirely.
> 
> As I said previously, this type of ownership detection is vulnerable to edge cases we haven't though of (as this latest fix shows). The safest thing to do is just require users to pass in a path to unpacked headers. We shouldn't have bpftrace reading from globally-writable directories without the user explicitly telling it to or we're risking more vulnerabilities in the future.
> 
> Also shelling out to `tar` is nasty.

Agreed with @ajor, let's drop this completely.

@jordalgo will you take it or should I have a look?

jordalgo:
Good call. I forgot that we actually decided to do what @ajor suggested in a past office hours. I'll close this PR out and make a new one with the agreed upon changes.

Issue #3005, opened 2024-02-14
[RFC] Probe expansion in codegen


viktormalik:
While working on #2334, I realized that we'll need to significantly change the way we do probe expansion, so I'm opening an RFC to see what other people opinions are, before I start implementing it.

Let us have a simple wildcarded probe `kfunc:vfs_* { ... }`.

At the moment, we generate **one** LLVM function, LLVM generates **one** BPF program from it, then we perform the expansion (74 probes), load the BPF program **74 times** (each time with a different BTF id of the probe), and attach each instance to a different probe.

The problem is that if we delegate probe loading to libbpf, it will need to discover the probes from the ELF object and therefore we'll need to generate **74 copies** of the same LLVM function (unless we somehow force LLVM to create multiple symbol table entries for the same BPF function). This will heavily enlarge the codegen output and the size of the ELF file.

My stand is that this is still worth it as moving to libbpf will have several advantages:
- less code (e.g. no need for custom relocations),
- possibility to use libbpf's attachment in future (removing dependency on BCC),
- a "standard" ELF produced by bpftrace, possibly usable by other loaders other than libbpf,
- access to other libbpf features.

The codegen/ELF size itself is a hidden technical detail and it'll probably only cause trouble for debugging. Also, for some probes (e.g. `kprobe`), we'll be ok with just a single LLVM function but for others (like `k(ret)func`), we'll always need to do the expansion.

ajor:
> 1. we generate one LLVM function
> 2. LLVM generates one BPF program from it
> 3. we perform the expansion (74 probes)
> 4. load the BPF program 74 times (each time with a different BTF id of the probe)
> 5. attach each instance to a different probe

What do you mean by "perform the expansion" in step 3? I'm not familiar with this code - are we copying the bytecode currently?

My concern with unnecessary probe expansion is the performance impact if we want to attach to 100k+ probes (e.g. `fentry:*, fexit:* {}`). Generating multiple symbols for the same function seems like something that should be possible to me.

viktormalik:
> 1. we generate one LLVM function
> 2. LLVM generates one BPF program from it
> 3. we perform the expansion (74 probes)
> 4. load the BPF program 74 times (each time with a different BTF id of the probe)
> 5. attach each instance to a different probe
>
> What do you mean by "perform the expansion" in step 3? I'm not familiar with this code - are we copying the bytecode currently?

[...]

PR #2757, opened 2023-09-06
Add pid and host targeting for uprobes/uretprobes

jordalgo:
This allows users to specify a single wildcard in place of the library_path to either target the whole host or a process's address space if a PID is also provided.

##### Checklist

- [x] Language changes are updated in `man/adoc/bpftrace.adoc` and if needed in `docs/reference_guide.md`
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [x] The new behaviour is covered by tests

PR #3155, opened 2024-05-07
Move probe expansion into codegen

viktormalik:
The previous probe expansion approach tried to minimize the amount of LLVM functions generated by emitting a single function for all probe matches in most cases. While this was efficient, it came with a couple of drawbacks:
- It is necessary to generate a separate LLVM function for each match (e.g. when the `probe` builtin is used). This leads to having two very similar loops for iterating matches in `BPFtrace::add_probe` and in `CodegenLLVM::visit(Probe)` which is quite confusing and hard to maintain.
- libbpf needs one BPF program (i.e. one LLVM function) per probe so if we want to delegate program loading (and possibly attachment) to libbpf (which we do), we cannot use this approach. See #3005 for more details.

This refactors probe expansion by moving most of it into codegen. Overall, we now distinguish two types of probe expansion:

- **Full expansion** - A separate LLVM function is generated for each match. This is used for most expansions now.
- **Multi expansion** - Used for k(u)probes when k(u)probe_multi is available. Generates one LLVM function and one BPF program for all matches and attaches the expanded functions via `bpf_link_create_opts`.

This allows to drop a lot of duplicated code. The expansion for "full" is done in `CodegenLLVM::visit(Probe)`, the expansion for "multi" is done in `BPFtrace::add_probe`.

A drawback of this approach is that we generate substantially larger ELF objects for expansions of probe types which do not support multi-probes (e.g. kfuncs and tracepoints) as we generate duplicate LLVM functions. This is something we can live with for now since multi-attachment is not the main use-case for these probe types (e.g. attaching to many kfuncs is very slow) and there's usually an alternative to use multi-kprobes. In addition, we will probably add a third expansion type using LLVM symbol aliases but that requires libbpf changes.

One particular area where this refactoring caused problems is unit tests in `tests/bpftrace.cpp`. Previously, it was sufficient to generate a simple `ast::Probe` and pass it to `BPFtrace::add_probe` since that was where most of the expansion was done. Now that the expansion was moved to codegen, we need to do full parser -> field analyser -> clang parser -> semantic analyser -> codegen sequence.

Also, the problem comes with USDT probes as it is not possible to easily mock `USDTHelper` which is a fully static class. Since we need to override `AttachPoint::usdt::num_locations` from tests, we allow to do that via a new internal env variable `BPFTRACE_TEST_USDT_NUM_LOCATIONS`.

##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -62,6 +62,17 @@ enum class Operator {
   BNOT,
 };
 
+// There are 2 kinds of attach point expansion:
+// - full expansion  - separate LLVM function is generated for each match
+// - multi expansion - one LLVM function and BPF program is generated for all
+//                     matches, the list of expanded functions is attached to
+//                     the BPF program using the k(u)probe.multi mechanism
+enum class ExpansionType {
+  NONE,
+  FULL,
+  MULTI,
+};
+
 class Node {
 public:
   Node() = default;
@@ -627,7 +638,9 @@ public:
   uint64_t len = 0;   // for watchpoint probes, the width of watched addr
   std::string mode;   // for watchpoint probes, the watch mode
   bool async = false; // for watchpoint probes, if it's an async watchpoint
-  bool need_expansion = false;
+
+  ExpansionType expansion = ExpansionType::NONE;
+
   uint64_t address = 0;
   uint64_t func_offset = 0;
   bool ignore_invalid = false;
--- a/src/ast/attachpoint_parser.cpp
+++ b/src/ast/attachpoint_parser.cpp
@@ -1,5 +1,6 @@
 #include "ast/attachpoint_parser.h"
 
+#include "ast.h"
 #include "ast/int_parser.h"
 #include "log.h"
 #include "types.h"
@@ -340,9 +341,17 @@ AttachPointParser::State AttachPointParser::kprobe_parser(bool allow_offset)
     ap_->func = parts_[func_idx];
   }
 
-  if (ap_->func.find('*') != std::string::npos ||
-      ap_->target.find('*') != std::string::npos)
-    ap_->need_expansion = true;
+  // kprobe_multi does not support the "module:function" syntax so in case
+  // a module is specified, always use full expansion
+  if (has_wildcard(ap_->target))
+    ap_->expansion = ExpansionType::FULL;
+  else if (has_wildcard(ap_->func)) {
+    if (ap_->target.empty() && bpftrace_.feature_->has_kprobe_multi()) {
+      ap_->expansion = ExpansionType::MULTI;
+    } else {
+      ap_->expansion = ExpansionType::FULL;
+    }
+  }
 
   return OK;
 }
@@ -440,9 +449,17 @@ AttachPointParser::State AttachPointParser::uprobe_parser(bool allow_offset,
       ap_->func = func;
   }
 
-  if (ap_->target.find('*') != std::string::npos ||
-      ap_->func.find('*') != std::string::npos)
-    ap_->need_expansion = true;
+  // As the C++ language supports function overload, a given function name
+  // (without parameters) could have multiple matches even when no
+  // wildcards are used.
+  if (has_wildcard(ap_->func) || has_wildcard(ap_->target) ||
+      ap_->lang == "cpp") {
+    if (bpftrace_.feature_->has_uprobe_multi()) {
+      ap_->expansion = ExpansionType::MULTI;
+    } else {
+      ap_->expansion = ExpansionType::FULL;
+    }
+  }
 
   return OK;
 }
@@ -477,10 +494,10 @@ AttachPointParser::State AttachPointParser::usdt_parser()
     ap_->func = parts_[3];
   }
 
-  if (ap_->target.find('*') != std::string::npos ||
-      ap_->ns.find('*') != std::string::npos || ap_->ns.empty() ||
-      ap_->func.find('*') != std::string::npos || bpftrace_.pid())
-    ap_->need_expansion = true;
+  // Always fully expand USDT probes as they may access args
+  if (has_wildcard(ap_->target) || has_wildcard(ap_->ns) || ap_->ns.empty() ||
+      has_wildcard(ap_->func) || bpftrace_.pid())
+    ap_->expansion = ExpansionType::FULL;
 
   return OK;
 }
@@ -505,7 +522,7 @@ AttachPointParser::State AttachPointParser::tracepoint_parser()
 
   if (ap_->target.find('*') != std::string::npos ||
       ap_->func.find('*') != std::string::npos)
-    ap_->need_expansion = true;
+    ap_->expansion = ExpansionType::FULL;
 
   return OK;
 }
@@ -627,7 +644,7 @@ AttachPointParser::State AttachPointParser::watchpoint_parser(bool async)
 
     ap_->func = func_arg_parts[0];
     if (ap_->func.find('*') != std::string::npos)
-      ap_->need_expansion = true;
+      ap_->expansion = ExpansionType::FULL;
 
     if (func_arg_parts[1].size() <= 3 || func_arg_parts[1].find("arg") != 0) {
       errs_ << "Invalid function argument" << std::endl;
@@ -696,7 +713,7 @@ AttachPointParser::State AttachPointParser::kfunc_parser()
 
   if (ap_->func.find('*') != std::string::npos ||
       ap_->target.find('*') != std::string::npos)
-    ap_->need_expansion = true;
+    ap_->expansion = ExpansionType::FULL;
 
   return OK;
 }
@@ -714,7 +731,7 @@ AttachPointParser::State AttachPointParser::iter_parser()
 
   if (parts_[1].find('*') != std::string::npos) {
     if (listing_) {
-      ap_->need_expansion = true;
+      ap_->expansion = ExpansionType::FULL;
     } else {
       if (ap_->ignore_invalid)
         return SKIP;
@@ -744,7 +761,7 @@ AttachPointParser::State AttachPointParser::raw_tracepoint_parser()
   ap_->func = parts_[1];
 
   if (has_wildcard(ap_->func))
-    ap_->need_expansion = true;
+    ap_->expansion = ExpansionType::FULL;
 
   return OK;
 }
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -48,8 +48,17 @@ namespace bpftrace {
 namespace ast {
 
 CodegenLLVM::CodegenLLVM(Node *root, BPFtrace &bpftrace, bool is_aot)
+    : CodegenLLVM(root, bpftrace, is_aot, std::make_unique<USDTHelper>())
+{
+}
+
+CodegenLLVM::CodegenLLVM(Node *root,
+                         BPFtrace &bpftrace,
+                         bool is_aot,
+                         std::unique_ptr<USDTHelper> usdt_helper)
     : root_(root),
       bpftrace_(bpftrace),
+      usdt_helper_(std::move(usdt_helper)),
       context_(std::make_unique<LLVMContext>()),
       module_(std::make_unique<Module>("bpftrace", *context_)),
       b_(*context_, *module_, bpftrace),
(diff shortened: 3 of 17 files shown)


viktormalik:
Update: I had to drop unit tests for single-wildcard targets for uprobe/USDT added in #2757. The reason is that tests in `tests/bpftrace.cpp` now run semantic analyser which in turn runs discovery of all running processes on the system for such probes, which is not something that should be run in unit tests (and it cannot be mocked in this case). FWIW, I don't think that those tests would add much value b/c all the code added by #2757 would be replaced by mocks anyways.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2550,6 +2550,50 @@
         func_type, func_name, current_attach_point_->address, index);
 }
 
+void CodegenLLVM::add_probe(AttachPoint &ap,
+                            Probe &probe,
+                            const std::string &name,
+                            FunctionType *func_type)
+{
+  current_attach_point_ = &ap;
```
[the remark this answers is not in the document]
viktormalik: Same situation as already discussed. `current_attach_point_` is only used by the `generateProbe` calls within this function but I added `current_attach_point_ = nullptr` to the end of the function to make sure that a pointer to a local parameter doesn't leak the function scope.

ajor (commented):
Are "full expansion" and "multi expansion" meant to be mutually exclusive? An enum instead of two bools might be clearer/safer if that's the case.

I haven't dug into the cause, but I can trigger a crash on this branch by setting both full and multi expansion to true on a single probe with this script:
```
kprobe:*:__x64_sys_rea* { print("a") }
```

src/ast/attachpoint_parser.cpp
```diff
@@ -440,9 +449,17 @@ AttachPointParser::State AttachPointParser::uprobe_parser(bool allow_offset,
       ap_->func = func;
   }
 
-  if (ap_->target.find('*') != std::string::npos ||
-      ap_->func.find('*') != std::string::npos)
-    ap_->need_expansion = true;
+  // As the C++ language supports function overload, a given function name
+  // (without parameters) could have multiple matches even when no
+  // wildcards are used.
+  if (has_wildcard(ap_->func) || has_wildcard(ap_->target) ||
```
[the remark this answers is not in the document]
viktormalik: In such a case, we do multi-expansion for each expanded path. See `BPFtrace::add_probe` for details.

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to
   - [#3060](https://github.com/bpftrace/bpftrace/pull/3060)
 - Disable func builtin for kretprobes and uretprobes when `get_func_ip` feature is not available
   - [#2645](https://github.com/bpftrace/bpftrace/pull/2645)
+- Move probe expansion into codegen
```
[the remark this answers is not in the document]
viktormalik: Good point, I'm dropping it.

viktormalik:
> Are "full expansion" and "multi expansion" meant to be mutually exclusive? An enum instead of two bools might be clearer/safer if that's the case.

Indeed they are, a great suggestion!

> 
> I haven't dug into the cause, but I can trigger a crash on this branch by setting both full and multi expansion to true on a single probe with this script:
> 
> ```
> kprobe:*:__x64_sys_rea* { print("a") }
> ```

Nice catch! It should be fixed now.

viktormalik:
While rebasing onto master, I noticed that uprobe_multi expansion doesn't work with placing uprobes after the funtion prologue. I opened #3173 to track that issue and this PR respects the current behavior (i.e. when uprobe expansion is needed, use kprobe_multi and attach to function entries, otherwise attach after function prologue).

ajor (commented):
Just a partial review so far - I'll try and get through `bpftrace.cpp` tomorrow.

I'm not keen on using environment variables to mock behaviour - it runs the risk of something dodgy happening if a user sets the env var at run time. I've recently started thinking about how we can make bpftrace a "trusted compiler" to allow it to be used by a non-root user (for use in containers with BPF token), and for that to work the compiler's behaviour has to be pretty well defined.

How about this as a way of mocking static classes: https://godbolt.org/z/37czorj84

src/bpftrace.cpp
```diff
@@ -1204,13 +1089,6 @@ int BPFtrace::run(BpfBytecode bytecode)
   if (std::getenv("__BPFTRACE_NOTIFY_PROBES_ATTACHED"))
     std::cerr << "__BPFTRACE_NOTIFY_PROBES_ATTACHED" << std::endl;
 
-#ifdef HAVE_LIBSYSTEMD
```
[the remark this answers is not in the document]
viktormalik: Yes, thanks.

tests/bpftrace.cpp
```diff
@@ -313,15 +295,34 @@ TEST(bpftrace, add_probes_wildcard_kprobe_multi)
...
+  ASSERT_EQ(2U, bpftrace->get_probes().size());
+  ASSERT_EQ(0U, bpftrace->get_special_probes().size());
+
+  std::string probe_orig_name =
+      "kprobe:sys_read,kprobe:not_here_*,kprobe:sys_write";
+  check_kprobe(bpftrace->get_probes().at(0), "sys_read", probe_orig_name);
+  check_kprobe(bpftrace->get_probes().at(1), "sys_write", probe_orig_name);
+}
+
+TEST(bpftrace, add_probes_wildcard_no_matches_kprobe_multi)
```
ajor, line 316: (naming) Isn't the test above the one using kprobe_multi and this one is without it?

tests/bpftrace.cpp
```diff
@@ -604,15 +553,15 @@ TEST(bpftrace, add_probes_uprobe_string_offset)
 TEST(bpftrace, add_probes_uprobe_cpp_symbol)
 {
   for (const std::string provider : { "uprobe", "uretprobe" }) {
-    std::string prog = provider + ":/bin/sh:cpp:cpp_mangled{}";
-    ast::Probe *probe = parse_probe(prog);
-
     auto bpftrace = get_strict_mock_bpftrace();
+    bpftrace->feature_ = std::make_unique<MockBPFfeature>(false);
```
ajor, line 557: Why disable features in this test? (also done in a few others below)

ajor (commented):
Sorry for taking so long to review! Just got a few minor points, but overall looks good to me.

It's annoying that the wildcard-target unit tests had to go, but I can see that it'd be a fair bit of work to keep them.

I think ideally we should aim to get to the point where nothing in the compilation pipeline depends on any external state, so it can all be unit tested without mocks. All external state could be collected up-front and passed in as arguments - it'd help to simplify compilation (escpecially AOT) a fair bit too. Definitely too much work to do in this PR though

tests/bpftrace.cpp
viktormalik: This is to disable uprobe_multi and check that full expansion happens in such a case.

tests/bpftrace.cpp
viktormalik: Correct, good catch!

src/bpftrace.cpp
```diff
@@ -104,237 +107,119 @@ Probe BPFtrace::generate_probe(const ast::AttachPoint &ap, const ast::Probe &p)
...
+              type == ProbeType::asyncwatchpoint) &&
+             ap.func.size()) {
+    // (async)watchpoint - generate also the setup probe
+    resources.probes.emplace_back(generateWatchpointSetupProbe(ap, p));
+    resources.watchpoint_probes.emplace_back(std::move(probe));
+  } else if (ap.expansion == ast::ExpansionType::MULTI) {
+    // k(u)probe_multi - do expansion and set probe.funcs
+    auto matches = probe_matcher_->get_matches_for_ap(ap);
+    if (!has_wildcard(ap.target)) {
+      if (!matches.empty()) {
```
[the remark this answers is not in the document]
viktormalik: This entire code path was kind of messy and not entirely correct. I refactored it, hopefully it's more readable now.

viktormalik:
> Sorry for taking so long to review! Just got a few minor points, but overall looks good to me.

Thanks for the review! While updating the PR, I discovered that there was an error in handling expansion for wildcarded `kprobe:module:function` cases so I also added a couple of unit tests for that.

src/bpftrace.cpp
```diff
@@ -104,237 +107,111 @@ Probe BPFtrace::generate_probe(const ast::AttachPoint &ap, const ast::Probe &p)
...
-        attach_funcs.push_back(attach_point->target + ":" + attach_point->func);
-      else if ((probetype(attach_point->provider) == ProbeType::kprobe ||
-                probetype(attach_point->provider) == ProbeType::kretprobe) &&
-               !attach_point->target.empty()) {
-        attach_funcs.push_back(attach_point->target + ":" + attach_point->func);
-      } else {
-        attach_funcs.push_back(attach_point->func);
-      }
+      std::vector<std::string> funcs(matches.begin(), matches.end());
+      probe.funcs = funcs;
```
ajor, line 158:
```suggestion
      probe.funcs(matches.begin(), matches.end());
```
In-place construction to avoid copying the vector and strings

src/bpftrace.cpp
```diff
@@ -104,237 +107,111 @@ Probe BPFtrace::generate_probe(const ast::AttachPoint &ap, const ast::Probe &p)
...
-      int err = resolve_uname(attach_point->func, &sym, attach_point->target);
-
-      if (attach_point->lang == "cpp") {
-        // As the C++ language supports function overload, a given function name
-        // (without parameters) could have multiple matches even when no
-        // wildcards are used.
-        matches = probe_matcher_->get_matches_for_ap(*attach_point);
+      for (auto &pair : target_map) {
+        resources.probes.push_back(std::move(pair.second));
       }
```
ttreyer, line 155: Is this ever executed? It looks like target expansion takes place in `SemanticAnalyser`, so `has_wildcard(ap.target)` is always false.

src/ast/ast.h
```diff
@@ -62,6 +62,17 @@ enum class Operator {
...
+// There are 2 kinds of attach point expansion:
+// - full expansion  - separate LLVM function is generated for each match
+// - multi expansion - one LLVM function and BPF program is generated for all
+//                     matches, the list of expanded functions is attached to
+//                     the BPF program using the k(u)probe.multi mechanism
+enum class ExpansionType {
+  NONE,
+  FULL,
+  MULTI,
+};
```
ttreyer, line 74: Could we detach the `k(u)probe.multi` feature from the `ExpansionType`?
What I'd love to be able to do is tell `bpftrace` which addresses to trap and it decides by itself if whether to use `k(u)probe.multi`, if available, or set-up the probe one-by-one for each address, if there are no multi probe support.

src/bpftrace.cpp
viktormalik: Semantic analyser only resolves the binary path. The comment there says:
https://github.com/bpftrace/bpftrace/blob/ce8f81a1945ac23ed94fe0db942c9592971a6320/src/ast/passes/semantic_analyser.cpp#L2605-L2607
That is, when wildcards are used (and multiple matches exist), the wildcard is not expanded.

I double checked and this is indeed executed when attaching to `uprobe:tests/testprogs/uprobe*:*`.

src/ast/ast.h
viktormalik: This is an excellent point. If we do AOT, we may compile the script on a machine with kprobe_multi support but then run it on a machine which doesn't have it. In such a case, the execution would probably not work.

This is quite tricky to implement, though. Since we're about to use libbpf to load the programs, for full expansion, we need to have one copy of the function for each match. I'll work on a libbpf extension to support copies via aliases (i.e. just entries in the symbol table pointing to the same code) but for now, we need full copies.

The only reasonable option that comes to mind is to replace the `MULTI` expansion by `ALIAS` (or better name if we can find one). For that expansion type, we would always emit just one copy of the program. Then, the runtime would detect if `k(u)probe_multi` is available and, if not, create the copies directly in the ELF. In future, we could replace that by just creating aliasing symbol table entries. I'll just need to check how complicated it is to do symbol copying in ELF.

In the meantime, if someone has a better idea, let me know.

src/bpftrace.cpp
viktormalik: I used `probe.funcs.assign(matches.begin(), matches.end())` which is more compact and (IIUC) should have the same complexity.

ajor (approved):
Looks good to me!

danobi (approved):
nice cleanup. tbh, didn't look too closely since it seems like most code was just moved

tests/bpftrace.cpp
```diff
@@ -1,8 +1,13 @@
 #include <cstring>
 
 #include "bpftrace.h"
+#include "clang_parser.h"
 #include "driver.h"
 #include "mocks.h"
+#include "passes/codegen_llvm.h"
+#include "passes/field_analyser.h"
+#include "passes/resource_analyser.h"
+#include "passes/semantic_analyser.h"
```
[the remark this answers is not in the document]
viktormalik: Yeah, that's better (although this compiled, too).

PR #3156, opened 2024-05-07
Don't unpack kernel headers or look in tmp

jordalgo:
Looking in shared writeable locations for kernel
headers is inherently risky even bpftrace does
the unpacking. Remove this functionality and let
the user specify the path to these headers if
we can't find them in known locations.

References:
https://github.com/bpftrace/bpftrace/pull/3033
https://github.com/bpftrace/bpftrace/pull/3154

##### Checklist

### DIFF
--- a/src/clang_parser.cpp
+++ b/src/clang_parser.cpp
@@ -660,9 +660,6 @@ bool ClangParser::parse(ast::Program *program,
   StderrSilencer silencer;
   silencer.silence();
 #endif
-  if (program->c_definitions.empty() && bpftrace.btf_set_.empty())
-    return true;
-
   input = "#include </bpftrace/include/__btf_generated_header.h>\n" +
           program->c_definitions;
 
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -108,8 +108,6 @@ const struct vmlinux_location vmlinux_locs[] = {
   { nullptr, false },
 };
 
-constexpr std::string_view PROC_KHEADERS_PATH = "/sys/kernel/kheaders.tar.xz";
-
 static bool pid_in_different_mountns(int pid);
 static std::vector<std::string> resolve_binary_path(const std::string &cmd,
                                                     const char *env_paths,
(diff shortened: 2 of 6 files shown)


viktormalik (approved):

viktormalik:
> Do we need another patch release after we land this? Considering I didn't really fix the problem before. cc @danobi

I would prefer that. Since it's a CVE, it'd make the life of us distro packagers easier.

FWIW, I think that we should also do a regular release soon, preferably with this merged.

src/utils.cpp
```diff
@@ -827,11 +727,14 @@ std::tuple<std::string, std::string> get_kernel_dirs(
...
-        return std::make_tuple(kheaders_tar_xz_path, kheaders_tar_xz_path);
-    }
+    LOG(WARNING) << "Could not find kernel headers in " << ksrc << " or "
+                 << kobj
+                 << ". To specify a particular path to kernel headers, set the "
+                    "env variables BPFTRACE_KERNEL_SOURCE and, optionally, "
+                    "BPFTRACE_KERNEL_BUILD if the kernel was built in a "
+                    "different directory than its source. To create kernel "
+                    "headers run 'modprobe kheaders', which will create a tar "
+                    "file at /sys/kernel/kheaders.tar.xz";
```
tnovak, line 737: I imagine that in most cases bpftrace doesn't need kheaders (can use BTF data instead); would it make sense to only print this warning if kernel headers end up being required?

A couple of cases come to mind:

- Bpftrace script unconditionally includes one or more kernel headers (e.g. `#include <linux/sched.h>` without gating on `BPFTRACE_HAVE_BTF`)
- Conflicts between user-defined and BTF types

Related: [commit 91db77c5](https://github.com/bpftrace/bpftrace/commit/91db77c5707c30346035330ea4fb719821254161)

cc @danobi

viktormalik: This is a good point. I agree that we should not print the warning by default as there's a number of situations where kheaders are not necessary.

jordalgo: How about as a stop-gap, I just reinstate the `!bpftrace.feature_->has_btf()` check as a conditional to show the warning?

viktormalik: I may remember it wrong (it's been a while since I wrote that code) but IMHO kheaders are needed iff ClangParser runs so we should use the same condition as in [clang_parser.cpp:663](https://github.com/bpftrace/bpftrace/blob/master/src/clang_parser.cpp#L663):
```cpp
  if (program->c_definitions.empty() && bpftrace.btf_set_.empty())
```
It'd be nice if we could reuse it.

An alternative would be to issue the warning from ClangParser when it fails but that may be even more complicated.

src/clang_parser.cpp
```diff
@@ -660,9 +660,6 @@ bool ClangParser::parse(ast::Program *program,
   StderrSilencer silencer;
   silencer.silence();
 #endif
-  if (program->c_definitions.empty() && bpftrace.btf_set_.empty())
```
jordalgo, line 663: moved this into main

src/utils.cpp
jordalgo: I moved a few things around but I'm using this same check before even calling `get_kernel_dirs` so the warning should only be shown in that case.

viktormalik (approved):
Nice, I like this approach.

Technically speaking, ClangParser may run successfully without kheaders (e.g. when user only defines custom types) but I'm ok with issuing a warning for the sake of fixing this. In most cases, BTF is present and so no warning will be shown.

jordalgo:
> ClangParser may run successfully without kheaders (e.g. when user only defines custom types)

Ah ok, do we have a way to check for this already?

viktormalik:
> > ClangParser may run successfully without kheaders (e.g. when user only defines custom types)
> 
> Ah ok, do we have a way to check for this already?

I don't think so.

viktormalik:
@jordalgo I did a backport to bpftrace 0.20.x in #3190.

PR #3157, opened 2024-05-08
Add ability to specify a script's license

ajor:
Not all code will be GPL-licensed, so provide users with a way to override the assumed GPL license.

1. Use license defined in config if provided
2. Parse SPDX License IDs from scripts

Licenses defined in a BpfScript config block take precedence over SPDX IDs.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -53,7 +53,11 @@ x86_64, arm64, s390x, arm32, loongarch64, mips64, ppc64, riscv64
 
 The license of a bpftrace script affects how it can be run, as the Linux kernel limits the functionality available to non-GPL licensed programs.
 
-A license can be specified by setting the <<license>> config variable.
+A license can be specified by setting the <<license>> config variable or by including a single-line comment containing a SPDX License ID:
+
+----
+// SPDX-License-Identifier: GPL-2.0-or-later
+----
 
 If a license is not explicitly declared, bpftrace will assume that the script is GPL v2 compatible.
 
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -49,6 +49,14 @@ List all the probes attached in the program::
 
 x86_64, arm64, s390x, arm32, loongarch64, mips64, ppc64, riscv64
 
+== Script Licenses
+
+The license of a bpftrace script affects how it can be run, as the Linux kernel limits the functionality available to non-GPL licensed programs.
+
+A license can be specified by setting the <<license>> config variable.
+
+If a license is not explicitly declared, bpftrace will assume that the script is GPL v2 compatible.
+
 == Options
 
 === *-B* _MODE_
@@ -3339,6 +3347,16 @@ This feature can be turned off by setting the value of this environment variable
 Default: 0
 
 For user space symbols, symbolicate lazily/on-demand (1) or symbolicate everything ahead of time (0).
+
+=== license
+
+Default: <unset>
+
+The license covering the script's source code.
+
+This license string is passed to the kernel where it is used to determine what functionality can be used.
+
+Licenses recognised by the Linux kernel are listed in link:https://docs.kernel.org/process/license-rules.html#id1[its documentation].
 
 === log_size
 
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -607,14 +607,12 @@ CallInst *IRBuilderBPF::CreateProbeReadStr(Value *ctx,
   // int bpf_probe_read_str(void *dst, int size, const void *unsafe_ptr)
   FunctionType *probereadstr_func_type = FunctionType::get(
       getInt64Ty(), { dst->getType(), getInt32Ty(), src->getType() }, false);
-  PointerType *probereadstr_func_ptr_type = PointerType::get(
-      probereadstr_func_type, 0);
-  Constant *probereadstr_callee = ConstantExpr::getCast(
-      Instruction::IntToPtr, getInt64(read_fn), probereadstr_func_ptr_type);
-  CallInst *call = createCall(probereadstr_func_type,
-                              probereadstr_callee,
-                              { dst, size_i32, src },
-                              probeReadHelperName(read_fn));
+
+  CallInst *call = CreateHelperCall(read_fn,
+                                    probereadstr_func_type,
+                                    { dst, size_i32, src },
+                                    probeReadHelperName(read_fn),
+                                    &loc);
   CreateHelperErrorCond(ctx, call, read_fn, loc);
   return call;
 }
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -96,12 +96,16 @@ CodegenLLVM::CodegenLLVM(Node *root, BPFtrace &bpftrace)
                          llvm::DEBUG_METADATA_VERSION);
 
   // Set license of BPF programs
-  const std::string license = "GPL";
+  const auto &license = bpftrace_.get_license();
   auto license_var = llvm::dyn_cast<GlobalVariable>(module_->getOrInsertGlobal(
       "LICENSE", ArrayType::get(b_.getInt8Ty(), license.size() + 1)));
   license_var->setInitializer(
       ConstantDataArray::getString(module_->getContext(), license.c_str()));
   license_var->setSection("license");
+
+  // XXX: Set licenses for manually loaded programs. Remove once libbpf is used
+  // instead:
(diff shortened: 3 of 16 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -49,6 +49,18 @@ List all the probes attached in the program::
 
 x86_64, arm64, s390x, arm32, loongarch64, mips64, ppc64, riscv64
 
+== Script Licenses
+
+The license of a BpfScript program affects how it can be run, as the Linux kernel limits the functionality available to non-GPL licensed programs.
```
jordalgo, line 54: I don't see anywhere else in the docs or code that refers to these programs as "BpfScript". I know we want to go in this direction but maybe for now we should refer to them as "bpftrace programs" or "bpftrace scripts".

man/adoc/bpftrace.adoc
```diff
@@ -3340,6 +3352,14 @@ Default: 0
 
 For user space symbols, symbolicate lazily/on-demand (1) or symbolicate everything ahead of time (0).
 
+=== license
+
+Default: <unset>
+
+The license covering the script's source code.
+
+This license string is passed to the kernel where it is used to determine what functionality can be used.
```
jordalgo, line 3361: Should we specify that it can only be in the set of allowed strings?
https://elixir.bootlin.com/linux/latest/source/include/linux/license.h

src/ast/passes/codegen_llvm.cpp
```diff
@@ -96,12 +96,16 @@ CodegenLLVM::CodegenLLVM(Node *root, BPFtrace &bpftrace)
...
   // Set license of BPF programs
-  const std::string license = "GPL";
+  const auto &license = bpftrace_.get_license();
   auto license_var = llvm::dyn_cast<GlobalVariable>(module_->getOrInsertGlobal(
       "LICENSE", ArrayType::get(b_.getInt8Ty(), license.size() + 1)));
   license_var->setInitializer(
       ConstantDataArray::getString(module_->getContext(), license.c_str()));
   license_var->setSection("license");
+
+  // XXX: Set licenses for manually loaded programs. Remove once libbpf is used
```
jordalgo, line 106: what is XXX?

src/driver.cpp
```diff
@@ -12,6 +12,31 @@ extern bpftrace::location loc;
...
+  auto license_pos = spdx_id_pos + spdx_id_marker.size();
+  auto eol_pos = program.find("\n", spdx_id_pos + 1);
+  if (eol_pos == program.npos)
+    return {};
+
+  auto license = program.substr(license_pos, eol_pos - license_pos);
+
+  // Try to translate some known GPL-v2-compatible licenses into a format
+  // understood by the Linux kernel
+  if (license.find("GPL-2.0") == 0)
```
jordalgo, line 33: Should we do something similar for GPL-1.0?

src/driver.cpp
```diff
@@ -12,6 +12,31 @@ extern bpftrace::location loc;
...
+    return "GPL";
```
jordalgo, line 34: Nit: but maybe this should return "GPL v2"

ajor: https://docs.kernel.org/process/license-rules.html#id1

The kernel docs description of "GPL v2" is:
> Same as “GPL”. It exists for historic reasons.

I took that to mean that "GPL" is the preferred format.

src/ast/passes/codegen_llvm.cpp
ajor: Just to draw attention to this line in the future because it shouldn't stay permanently.

Basically this: https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments

But also because it gets syntax highlighted in my editor :P

jordalgo: TIL!

src/driver.cpp
jordalgo: Ah this all stems from the fact that I thought the "GPL" string meant GPL v1 but apparently it doesn't 🙄

ajor: It's definitely confusing!

man/adoc/bpftrace.adoc
```diff
@@ -49,6 +49,18 @@ List all the probes attached in the program::
...
+
+The license of a bpftrace script affects how it can be run, as the Linux kernel limits the functionality available to non-GPL licensed programs.
+
+A license can be specified by setting the <<license>> config variable or by including a single-line comment containing a SPDX License ID:
+
+----
+// SPDX-License-Identifier: GPL-2.0-or-later
+----
+
+If a license is not explicitly declared, bpftrace will assume that the script is GPL v2 compatible.
```
jordalgo, line 62: Nit: This language feels a little weird to me. How about "If a license is not explicitly declared, bpftrace will license the script as GPL v2."

src/ast/irbuilderbpf.cpp
```diff
@@ -607,14 +607,12 @@ CallInst *IRBuilderBPF::CreateProbeReadStr(Value *ctx,
   // int bpf_probe_read_str(void *dst, int size, const void *unsafe_ptr)
   FunctionType *probereadstr_func_type = FunctionType::get(
       getInt64Ty(), { dst->getType(), getInt32Ty(), src->getType() }, false);
-  PointerType *probereadstr_func_ptr_type = PointerType::get(
-      probereadstr_func_type, 0);
-  Constant *probereadstr_callee = ConstantExpr::getCast(
```
jordalgo, line 612: Sorry, I might have missed it but what is this change?

ajor: Oops, you're right this could have done with an explanation. It's to produce more useful error messages for the `bpf_probe_read_str` helper. A lot of other helpers are already using this `CreateHelperCall` utility function, and then rest probably should be switched to it in the future as well.

`bpf_probe_read_str` is a GPL-only helper function so I ended up getting a lot of error messages with this one in particular when playing around with different licenses.

Without this change:
```
# bpftrace -e 'config = {license = "poo"; } BEGIN { str(0); }' 
WARNING: Addrspace is not set
Attaching 1 probe...
ERROR: helper bpf_probe_read_str can only be used in GPL-compatible programs
```

With this change:
```
# bpftrace -e 'config = {license = "poo"; } BEGIN { str(0); }' 
WARNING: Addrspace is not set
Attaching 1 probe...
stdin:1:38-44: ERROR: helper bpf_probe_read_str can only be used in GPL-compatible programs
config = {license = "poo"; } BEGIN { str(0); }
                                     ~~~~~~
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -96,12 +96,16 @@ CodegenLLVM::CodegenLLVM(Node *root, BPFtrace &bpftrace)
...
+  // instead:
+  bpftrace_.resources.license = license;
```
viktormalik, line 108: I know it's just temporary but this doesn't feel like a good place to set this. Why not do it in `ResourceAnalyser`?

src/config.cpp
```diff
@@ -13,6 +13,7 @@ Config::Config(bool has_cmd, bool bt_verbose) : bt_verbose_(bt_verbose)
   config_map_ = {
     { ConfigKeyBool::cpp_demangle, { .value = true } },
     { ConfigKeyBool::lazy_symbolication, { .value = false } },
+    { ConfigKeyString::license, { .value = std::nullopt } },
```
viktormalik, line 16: This is a bit strange to me. The default license is GPL but we set the default config value to *none* here and only return "GPL" from `BPFtrace::get_license`. Why not set it to "GPL" here and only override the value from SPDX/config, if specified?

If the only reason is that we don't have a config priority defined for config value coming from a script comment header (SPDX here), then I believe that we should introduce that. It will also save the introduction of `std::optional<T> Config::try_get`.

viktormalik (commented):
Overall LGTM, just some implementation details to clear up.

src/config.cpp
jordalgo: Yes. Thank you @viktormalik for articulating this. I also didn't understand why we didn't want to set the default in the config.

src/driver.cpp
```diff
@@ -12,6 +12,31 @@ extern bpftrace::location loc;
 
 namespace bpftrace {
 
+namespace {
+std::string_view parse_license()
+{
+  std::string_view program = Log::get().get_source();
+  constexpr std::string_view spdx_id_marker = "SPDX-License-Identifier: ";
+  auto spdx_id_pos = program.find(spdx_id_marker);
```
danobi, line 20: Should we ensure this is inside a comment? Seems like this would match this string anywhere in the file

danobi:
Closing due to inactivity.

PR #3158, opened 2024-05-08
Add optional support for sending systemd readiness notifications

DaanDeMeyer:
When ENABLE_SYSTEMD is enabled, we'll link against libsystemd and send readiness notifications using sd_notify() as documented in https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html.

These will only be sent if the NOTIFY_$SOCKET environment variable is set to a unix socket where systemd expects to receive readiness notifications.

This allows running bpftrace in a systemd service which will only be considered started once bpftrace has finished allocation its probes. This allows starting bpftrace from scripts and ensuring that its probes are registered before continuing.

For example, currently one has to do something like this:

"""
bpftrace <script> &
PID=$!
sleep 2 # How many seconds should we sleep???
<stuff that needs tracing>
kill $PID
"""

With the problem being that we never know how long we should sleep. Either we don't sleep long enough and miss messages, or we sleep too long and waste time doing nothing useful.

With this PR, we can do the following instead:

"""
systemd-run --unit=bpftrace --service-type=notify bpftrace <script> <stuff that needs tracing>
systemctl stop bpftrace
"""

The systemd-run command will only finish once bpftrace has sent the READY=1 notification, which is only send after probes have been registered, thus guaranteeing that we wait just long enough for the probes to be registered but not any longer.

##### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ set(VENDOR_GTEST OFF CACHE BOOL "Clone gtest from github")
 set(BUILD_FUZZ OFF CACHE BOOL "Build bpftrace for fuzzing")
 set(USE_LIBFUZZER OFF CACHE BOOL "Use libfuzzer for fuzzing")
 set(FUZZ_TARGET "codegen" CACHE STRING "Fuzzing target")
+set(ENABLE_SYSTEMD OFF CACHE BOOL "Enable systemd integration")
 
 set(ENABLE_SKB_OUTPUT ON CACHE BOOL "Enable skb_output, will include libpcap")
 
@@ -118,6 +119,11 @@ if(ENABLE_SKB_OUTPUT)
   find_package(LibPcap)
 endif()
 
+if(ENABLE_SYSTEMD)
+  find_package(PkgConfig)
+  pkg_check_modules(libsystemd REQUIRED IMPORTED_TARGET libsystemd)
+endif()
+
 if(POLICY CMP0075)
   cmake_policy(SET CMP0075 NEW)
 endif()
@@ -187,6 +193,10 @@ endif(LIBPCAP_FOUND)
 if (HAVE_LIBBPF_UPROBE_MULTI)
   set(BPFTRACE_FLAGS "${BPFTRACE_FLAGS}" HAVE_LIBBPF_UPROBE_MULTI)
 endif(HAVE_LIBBPF_UPROBE_MULTI)
+
+if(ENABLE_SYSTEMD)
+  set(BPFTRACE_FLAGS "${BPFTRACE_FLAGS}" HAVE_LIBSYSTEMD)
+endif()
 
 add_subdirectory(src)
 if (BUILD_TESTING)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ find_package(LibBfd)
 find_package(LibOpcodes)
 
 if(ENABLE_SKB_OUTPUT)
-find_package(LibPcap)
+  find_package(LibPcap)
 endif()
 
 if(POLICY CMP0075)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -3809,3 +3809,29 @@ Maps are printed by reference not by value and as the value gets updated right a
 ----
 
 Therefore, when you need precise event statistics, it is recommended to use synchronous functions (e.g. count() and hist()) to ensure more reliable and accurate results.
+
+=== Systemd support
+
+To run bpftrace in the background using systemd::
+----
+# systemd-run --unit=bpftrace --service-type=notify bpftrace -e 'kprobe:do_nanosleep { printf("%d sleeping\n", pid); }'
+----
+
+In the above example, systemd-run will not finish until bpftrace has attached
+its probes, so you can be sure that all following commands will be traced. To
+stop tracing, run `systemctl stop bpftrace`.
+
+To debug early boot issues, bpftrace can be invoked via a systemd service
+ordered before the service that needs to be traced. A basic unit file to run
+bpftrace before another service looks as follows::
+----
+[Unit]
+Before=service-i-want-to-trace.service
+
+[Service]
+Type=notify
+ExecStart=bpftrace -e 'kprobe:do_nanosleep { printf("%d sleeping\n", pid); }'
+----
+
+Similarly to the systemd-run example, the service to be traced will not start
+until bpftrace started by the systemd unit has attached its probes.
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -29,6 +29,9 @@
 #include <bcc/perf_reader.h>
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
+#ifdef HAVE_LIBSYSTEMD
+#include <systemd/sd-daemon.h>
+#endif
 
 #include "ast/async_event_types.h"
 #include "bpfmap.h"
@@ -1175,6 +1178,13 @@ int BPFtrace::run(BpfBytecode bytecode)
   if (std::getenv("__BPFTRACE_NOTIFY_PROBES_ATTACHED"))
     std::cerr << "__BPFTRACE_NOTIFY_PROBES_ATTACHED" << std::endl;
 
+#ifdef HAVE_LIBSYSTEMD
+  err = sd_notify(false, "READY=1\nSTATUS=Processing events...");
+  if (err < 0)
+    LOG(WARNING) << "Failed to send readiness notification, ignoring: "
+                 << strerror(-err);
+#endif
+
   if (has_iter_) {
     int err = run_iter();
     if (err)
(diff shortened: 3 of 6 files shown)


src/bpftrace.cpp
```diff
@@ -1175,6 +1178,13 @@ int BPFtrace::run(BpfBytecode bytecode)
   if (std::getenv("__BPFTRACE_NOTIFY_PROBES_ATTACHED"))
     std::cerr << "__BPFTRACE_NOTIFY_PROBES_ATTACHED" << std::endl;
 
+#ifdef HAVE_LIBSYSTEMD
+  err = sd_notify(false, "READY=1\nSTATUS=Processing requests...");
+  if (err < 0)
+    LOG(ERROR) << "Failed to send readiness notification, ignoring: "
```
jordalgo, line 1184: If we're not exiting or aborting this should be `LOG(WARNING)` - same goes for the one below.

src/bpftrace.cpp
```diff
@@ -1175,6 +1178,13 @@ int BPFtrace::run(BpfBytecode bytecode)
   if (std::getenv("__BPFTRACE_NOTIFY_PROBES_ATTACHED"))
     std::cerr << "__BPFTRACE_NOTIFY_PROBES_ATTACHED" << std::endl;
 
+#ifdef HAVE_LIBSYSTEMD
+  err = sd_notify(false, "READY=1\nSTATUS=Processing requests...");
```
jordalgo, line 1182: what is the "Processing requests..." message?

jordalgo (commented):
The changes look good to me. The only thing I'm thinking about is testing this functionality so we don't accidentally break it some day e.g. creating a build in the CI pipeline with this enabled and checking if we get the notify. Not sure how hard that is though.

Also, is it bad to always attempt to dynamically link against this systemd lib always, and only fire the notify is `NOTIFY_$SOCKET` is also populated?

DaanDeMeyer:
Well I assume bpftrace might be used on distributions such as Alpine Linux which do not package libsystemd, and we don't want to exclude them from being able to use bpftrace.

DaanDeMeyer:
@jordalgo I took a quick look but I have no clue where exactly I'd write a test. What the test needs to do is create a unix socket and bind it, set ``$NOTIFY_SOCKET` to the socket address, run any bpftrace program with probes and read from the socket to see if the messages were sent.

jordalgo:
> Well I assume bpftrace might be used on distributions such as Alpine Linux which do not package libsystemd, and we don't want to exclude them from being able to use bpftrace.

I'm still looking around but maybe we could use `dlopen` to dynamically load this library and fail gracefully if it doesn't exist on the distro. Then we don't have to worry about setting this flag for the different distro builds (although maybe that's not an issue).

DaanDeMeyer:
> > Well I assume bpftrace might be used on distributions such as Alpine Linux which do not package libsystemd, and we don't want to exclude them from being able to use bpftrace.
> 
> I'm still looking around but maybe we could use `dlopen` to dynamically load this library and fail gracefully if it doesn't exist on the distro. Then we don't have to worry about setting this flag for the different distro builds (although maybe that's not an issue).

dlopen() is definitely possible, but I don't think distributions will mind the extra flag. Lots of applications do it this way, where there's a build option to enable the systemd specific bits. Even if dlopen() is used, distributions would still have to list libsystemd as a recommended or required dependency of bpftrace to make sure it is pulled in when bpftrace is installed (though in practice libsystemd will always be installed)

jordalgo:
> distributions would still have to list libsystemd as a recommended or required dependency of bpftrace to make sure it is pulled in when bpftrace is installed (though in practice libsystemd will always be installed)

Ah right, good point.

jordalgo (approved):
I'm good with this change. I'll wait a few days to see if anyone else has feedback before landing.

src/bpftrace.cpp
```diff
@@ -1175,6 +1178,13 @@ int BPFtrace::run(BpfBytecode bytecode)
   if (std::getenv("__BPFTRACE_NOTIFY_PROBES_ATTACHED"))
     std::cerr << "__BPFTRACE_NOTIFY_PROBES_ATTACHED" << std::endl;
 
+#ifdef HAVE_LIBSYSTEMD
+  err = sd_notify(false, "READY=1\nSTATUS=Processing events...");
```
ajor, line 1182: Should the first argument be `0` instead? The docs seem to say it takes an integer:
> `int unset_environment`

ajor (commented):
Could you help me understand use-case for this please? Why would bpftrace be run in a systemd service?

src/bpftrace.cpp
DaanDeMeyer: The API uses int because `bool` was not a thing in C before `C99`. Every call to this function in the systemd codebase uses `true/false` so I did the same here.

ajor:
Thanks for the explanation - that does clear up how it's intended to be used.

I'm still not sure what the specific purpose is though (not that I'm saying there isn't one!). Who needs this feature and why?

DaanDeMeyer:
> Thanks for the explanation - that does clear up how it's intended to be used.
> 
> I'm still not sure what the specific purpose is though (not that I'm saying there isn't one!). Who needs this feature and why?

I need the feature to debug systemd's integration tests with bpftrace. Instead of writing the first snippet, I want to write the second snippet.

ajor:
Does the `-c` command line option not work for that purpose?

e.g.
```
# bpftrace my_script.bt -c ./test/run-integration-tests.sh
```

jordalgo:
> Does the -c command line option not work for that purpose?

Totally forgot about this option. Can't wait for the docs website :) We should definitely add an FAQ or a How-To page

DaanDeMeyer:
> Does the `-c` command line option not work for that purpose?
> 
> e.g.
> 
> ```
> # bpftrace my_script.bt -c ./test/run-integration-tests.sh
> ```

Does not work because I can not add it in the middle of an existing script. I don't want to run the entire script with bpftrace tracing, only a few commands need the tracing enabled. Having to move stuff around to accommodate `bpftrace -c` is much more work than just adding the snippet from above around the code I want to trace.

DaanDeMeyer:
Aside from the use case of using bpftrace in a script, If I want to run bpftrace during early boot to debug something then I do actually want to run it as a service so I can make sure it is started before the services that I want to trace, in which case I have the exact same problem that I don't want the services that need tracing to start before bpftrace has attached its probes. Again, the solution is the systemd notify integration so that I can drop in a service as follows:

bpftrace.service

```
[Unit]
Before=service-i-want-to-trace.service

[Service]
Type=notify
ExecStart=bpftrace <script>
```

Without the notify integration, systemd would immediately start the service I want to trace after systemd has execve()'d the bpftrace process, so you get a race condition where bpftrace might not attach its probes before the service that needs to be traced is started. The notify integration means that the service I want to trace is only started after bpftrace has attached its probes.

CMakeLists.txt
```diff
@@ -118,6 +119,11 @@ if(ENABLE_SKB_OUTPUT)
 find_package(LibPcap)
 endif()
 
+if(ENABLE_SYSTEMD)
+find_package(PkgConfig)
+pkg_check_modules(libsystemd REQUIRED IMPORTED_TARGET libsystemd)
```
danobi, line 124: Nit: pls indent (also fixing above LibPcap indend would be helpful too)

danobi (commented):
early boot use case sounds quite interesting. this change seems low maintenance cost + optional to enable so sounds reasonable to me.

maybe call out this feature in the man page? an example systemd-run one-liner or mentioning early boot debugging would be good. it'd at least help us not forget the use case

man/adoc/bpftrace.adoc
```diff
@@ -45,6 +45,30 @@ List all the probes attached in the program::
 # bpftrace -l -e 'kprobe:do_nanosleep { printf("%d sleeping\n", pid); }'
 ----
 
+Run bpftrace in the background using systemd::
```
danobi, line 48: I think it'd be better to add this to `Advanced Topics` section. Might confuse folks so early up in man page - mostly cuz it's still a default-off feature

danobi (commented):
also, could you please update src/bpffeature.cpp with an entry for systemd support?

src/bpffeature.cpp
```diff
@@ -642,6 +642,12 @@ std::string BPFfeature::report(void)
       << "  raw_tp_special: " << to_str(has_raw_tp_special())
       << "  iter: " << to_str(has_iter("task")) << std::endl;
 
+#ifdef HAVE_LIBSYSTEMD
+  buf << "Systemd support: yes" << std::endl;
+#else
+  buf << "Systemd support: no" << std::endl;
+#endif
+
```
ajor, line 650: This file is meant to be just for reporting BPF functionality detected at runtime. We've got a separate file for reporting build information: [src/build_info.cpp](https://github.com/bpftrace/bpftrace/blob/14cb69defaaf047f63c9c5fa21be27cd446f9315/src/build_info.cpp)

ajor (commented):
Sounds like a reasonable feature and looks good to me. Just need to move the feature detection code to a different file and I think we're good to go

ajor (approved):

martinetd:
Great feature, I think it'll be useful for some of the use cases described!

Before enabling it in nixos I came here to check the systemd dependency, since it's just sd_notify() I think we might as well open-code it (as in the first 90 lines of https://github.com/systemd/systemd/blob/main/man/notify-selfcontained-example.c ) -- that'll avoid the condition on the build that @jordalgo was worried about (and at worst distros like alpine linux will just never use that code as the env var won't be set)

I don't think it's a priority (especially now libsystemd no longer pulls in libxz :grin:), but perhaps as an easy "first issue" for backlog if you think it makes sense.

PR #3161, opened 2024-05-10
Fixes for watchpoint probes and runtime tests

viktormalik:
Fix two issues with watchpoint probes/tests introduced by recent PRs:
- #2997 forgot to update the `EXPECT` clauses for watchpoint runtime tests,
- #3153 broke the name of the LLVM functions/sections for watchpoint setup probes and bpftrace was not able to associate the generated BPF programs to probes.

The reason why neither of these was captured is a strange issue in CMake. During CMake run, we copy files from `tests/runtime/` to `build/tests/runtime`, however, the last file (in our case `tests/runtime/watchpoint`) is never copied. Due to this, watchpoint tests are not run in the CI and also do not get updated in local builds (it seems to have worked in the past b/c I had the file in my build directory but it was never updated).

I'm not sure if this is caused by our change or by a CMake update, however, our approach is IMHO not correct as it makes the `runtime_tests` target depend on the runtime tests files from the source directory. Changing it to depend on the (newly copied) files from the build directory eliminates the above issue. This is done by the last commit in the PR. 

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2547,7 +2547,7 @@ void CodegenLLVM::generateProbe(Probe &probe,
   if ((pt == ProbeType::watchpoint || pt == ProbeType::asyncwatchpoint) &&
       current_attach_point_->func.size())
     generateWatchpointSetupProbe(
-        func_type, func_name, current_attach_point_->address, index);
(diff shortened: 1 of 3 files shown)


jordalgo (approved):
Good catch!

danobi (approved):
Excellent catch!

PR #3164, opened 2024-05-15
Remove LOG(FATAL)

jordalgo:
Replace these with either LOG(BUG), LOG(ERROR) and exit(1) in main.cpp, or `throw std::runtime_error`.

This is to avoid calling `abort` for user errors and removes the confusion between `LOG(ERROR)` and `LOG(FATAL)`.

Issue: https://github.com/bpftrace/bpftrace/issues/3163

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -253,8 +253,7 @@ llvm::Type *IRBuilderBPF::GetType(const SizedType &stype)
         ty = getInt8Ty();
         break;
       default:
-        LOG(FATAL) << stype.GetSize()
-                   << " is not a valid type size for GetType";
+        LOG(BUG) << stype.GetSize() << " is not a valid type size for GetType";
     }
   }
   return ty;
@@ -666,8 +665,8 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
     int offset = 0;
     offset = arch::offset(argument->base_register_name);
     if (offset < 0) {
-      LOG(FATAL) << "offset for register " << argument->base_register_name
-                 << " not known";
+      LOG(BUG) << "offset for register " << argument->base_register_name
+               << " not known";
     }
 
     // bpftrace's args are internally represented as 64 bit integers. However,
@@ -682,8 +681,8 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
     if (argument->valid & BCC_USDT_ARGUMENT_INDEX_REGISTER_NAME) {
       int ioffset = arch::offset(argument->index_register_name);
       if (ioffset < 0) {
-        LOG(FATAL) << "offset for register " << argument->index_register_name
-                   << " not known";
+        LOG(BUG) << "offset for register " << argument->index_register_name
+                 << " not known";
       }
       index_offset = CreateGEP(getInt8Ty(),
                                ctx,
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -107,8 +107,8 @@ std::string progtypeName(libbpf::bpf_prog_type t)
 void check_banned_kretprobes(std::string const &kprobe_name)
 {
   if (banned_kretprobes.find(kprobe_name) != banned_kretprobes.end()) {
-    LOG(FATAL) << "kretprobe:" << kprobe_name
-               << " can't be used as it might lock up your system.";
+    throw FatalUserException("kretprobe:" + kprobe_name +
+                             " can't be used as it might lock up your system.");
   }
 }
 
@@ -118,8 +118,9 @@ void AttachedProbe::attach_kfunc(void)
     // Errors for kfunc are handled in load_prog, ignore here
     return;
   tracing_fd_ = bpf_raw_tracepoint_open(nullptr, progfd_);
-  if (tracing_fd_ < 0)
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+  if (tracing_fd_ < 0) {
+    throw FatalUserException("Error attaching probe: " + probe_.name);
+  }
 }
 
 int AttachedProbe::detach_kfunc(void)
@@ -136,7 +137,7 @@ void AttachedProbe::attach_iter(void)
                                 libbpf::BPF_TRACE_ITER),
                             NULL);
   if (linkfd_ < 0) {
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+    throw FatalUserException("Error attaching probe: " + probe_.name);
   }
 }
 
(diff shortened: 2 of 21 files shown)


jordalgo:
> Can LOG(FATAL) throw a custom exception rather than runtime_error? Having a central choke point is nice. Also it's a bit more consistent to read/write.

I think creating a custom exception is useful but I don't like the idea of `LOG(FATAL)` being responsible for throwing it as it obscures what's going on to someone reading the code. If some part of the code throws then it's immediately clear that the caller needs to handle this or let it bubble up. I also never liked the combination of `LOG(FATAL)` and `LOG(ERROR)` as it's confusing which one to use and when.

jordalgo:
Additionally, the way `log.cpp` works now we can't catch errors thrown in the destructor e.g. 
```
[[noreturn]] LogStreamBug::~LogStreamBug()
{
  throw std::runtime_error(buf_.str());
  //sink_.take_input(type_, loc_, out_, internal_location() + buf_.str());
  //abort();
}
```

src/attached_probe.cpp
```diff
@@ -1146,7 +1147,7 @@ void AttachedProbe::attach_uprobe(int pid, bool safe_mode)
                                         0);
 
   if (perf_event_fd < 0)
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+    LOG(BUG) << "Error attaching probe: " << probe_.name;
```
danobi, line 1150: This is plausible and  recoverable, right?

src/attached_probe.cpp
```diff
@@ -926,7 +926,7 @@ void AttachedProbe::attach_multi_kprobe(void)
                                 libbpf::BPF_TRACE_KPROBE_MULTI),
                             &opts);
   if (linkfd_ < 0) {
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+    LOG(BUG) << "Error attaching probe: " << probe_.name;
```
danobi, line 929: Same here

src/attached_probe.cpp
```diff
@@ -119,7 +119,7 @@ void AttachedProbe::attach_kfunc(void)
     return;
   tracing_fd_ = bpf_raw_tracepoint_open(nullptr, progfd_);
   if (tracing_fd_ < 0)
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+    LOG(BUG) << "Error attaching probe: " << probe_.name;
```
danobi, line 122: and here

src/attached_probe.cpp
```diff
@@ -136,7 +136,7 @@ void AttachedProbe::attach_iter(void)
                                 libbpf::BPF_TRACE_ITER),
                             NULL);
   if (linkfd_ < 0) {
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+    LOG(BUG) << "Error attaching probe: " << probe_.name;
```
danobi, line 139: and here

src/attached_probe.cpp
```diff
@@ -1118,7 +1119,7 @@ void AttachedProbe::attach_multi_uprobe(int pid)
                                 libbpf::BPF_TRACE_UPROBE_MULTI),
                             &opts);
   if (linkfd_ < 0) {
-    LOG(FATAL) << "Error attaching probe: " << probe_.name;
+    LOG(BUG) << "Error attaching probe: " << probe_.name;
```
danobi, line 1122: and here

src/attached_probe.cpp
```diff
@@ -1361,7 +1362,7 @@ void AttachedProbe::attach_profile()
                                               group_fd);
 
     if (perf_event_fd < 0)
-      LOG(FATAL) << "Error attaching probe: " << probe_.name;
+      LOG(BUG) << "Error attaching probe: " << probe_.name;
```
danobi, line 1365: and here (and for rest of file)

jordalgo: So should all of these be `LOG(WARNING)`?

danobi: Should probably throw runtime_error, I think. We cannot attach a subset of user requested probes. And I can't think of any way to recover higher up in the stack. If recovery is possible locally, it can be done before exception is thrown.

jordalgo: I think a `LOG(WARNING)` coupled with an early return is the same thing here as (according to our code guidelines) we don't want to throw recoverable errors, which this is.

src/utils.cpp
```diff
@@ -139,7 +139,7 @@ void StdioSilencer::silence()
     close(new_stdio);
   } catch (const std::system_error &e) {
     if (errno == EMFILE)
-      LOG(FATAL) << e.what() << ": please raise NOFILE";
+      LOG(BUG) << e.what() << ": please raise NOFILE";
```
viktormalik, line 142: Shouldn't use `throw FatalUserException`? The original code suggests that we consider `EMFILE` a user-caused error while everything else an internal bpftrace error.

src/attached_probe.cpp
viktormalik: How come these are recoverable? When probe attachment fails, bpftrace should also fail so why is this not throwing `FatalUserException`?

src/utils.cpp
jordalgo: I wasn't sure about this one. I'll change it.

src/attached_probe.cpp
jordalgo: @viktormalik History is on your side in that before I started changing the logging we used to throw a `std::runtime_error` and failing hard in [bpftrace.cpp](https://github.com/bpftrace/bpftrace/blob/4c8b262aed6399a04126d2f5199d0d588924e6d2/src/bpftrace.cpp#L946-L947). I can change these to `FatalUserException` (though I wish we had more info on the cause for the users). Alternatively, if even one probe is successful bpftrace can proceed but if there is user logic where two probes depend on one another than continuing would be worthless.

jordalgo: Ok, switched these all back to throw. Sorry for the churn.

src/ast/irbuilderbpf.cpp
```diff
@@ -253,8 +253,7 @@ llvm::Type *IRBuilderBPF::GetType(const SizedType &stype)
         ty = getInt8Ty();
         break;
       default:
-        LOG(FATAL) << stype.GetSize()
-                   << " is not a valid type size for GetType";
+        LOG(BUG) << stype.GetSize() << " is not a valid type size for GetType";
```
jordalgo, line 256: These errors all seemed like internal bugs. Please correct me if I'm wrong.

viktormalik: I believe that you're right.

viktormalik (approved):
LGTM!

PR #942, opened 2019-11-04
Add support to cast to a pointer of integer

mmisono:
An attempt to support casting to a pointer of int and then dereference. Closes #941.

example:
```
uprobe:./t.out:main+12 {
  $a = *(int8*)(reg("bp") - 1);
  printf("a=%d\n", $a);
}
```

<details>
<summary> Here is the output of `bpftrace -d` </summary>

```
Program
 uprobe:./t.out:main+12
  =
   variable: $a
   dereference
    (int8*)
     -
      call: reg
       string: bp
      int: 1
  call: printf
   string: a=%d\n
   variable: $a

; ModuleID = 'bpftrace'
source_filename = "bpftrace"
target datalayout = "e-m:e-p:64:64-i64:64-n32:64-S128"
target triple = "bpf-pc-linux"

%printf_t = type { i64, i64 }

; Function Attrs: nounwind
declare i64 @llvm.bpf.pseudo(i64, i64) #0

; Function Attrs: argmemonly nounwind
declare void @llvm.lifetime.start.p0i8(i64, i8* nocapture) #1

define i64 @"uprobe:./t.out:main+12"(i8*) local_unnamed_addr section "s_uprobe:./t.out:main+12_1" {
entry:
  %printf_args = alloca %printf_t, align 8
  %deref = alloca i8, align 1
  %1 = getelementptr i8, i8* %0, i64 32
  %reg_bp = load i64, i8* %1, align 8
  %2 = add i64 %reg_bp, -1
  call void @llvm.lifetime.start.p0i8(i64 -1, i8* nonnull %deref)
  %probe_read = call i64 inttoptr (i64 4 to i64 (i8*, i64, i8*)*)(i8* nonnull %deref, i64 1, i64 %2)
  %3 = load i8, i8* %deref, align 1
  call void @llvm.lifetime.end.p0i8(i64 -1, i8* nonnull %deref)
  %4 = bitcast %printf_t* %printf_args to i8*
  call void @llvm.lifetime.start.p0i8(i64 -1, i8* nonnull %4)
  %5 = getelementptr inbounds %printf_t, %printf_t* %printf_args, i64 0, i32 1
  %6 = bitcast %printf_t* %printf_args to i8*
  call void @llvm.memset.p0i8.i64(i8* nonnull align 8 %6, i8 0, i64 16, i1 false)
  store i8 %3, i64* %5, align 8
  %pseudo = call i64 @llvm.bpf.pseudo(i64 1, i64 1)
[...]

PR #3166, opened 2024-05-15
Fix intptrcast runtime tests

ajor:
Explicitly put numbers into an array on the stack instead of relying on architecture-specific calling conventions for stack-passed parameters.

This removes the dependency on the stack pointer's value (which changed as of #3095) and makes the tests architecture independent.

This fixes the recent test failures on master: https://github.com/bpftrace/bpftrace/actions/runs/9080541080

### DIFF
--- a/tests/runtime/intptrcast
+++ b/tests/runtime/intptrcast
@@ -1,83 +1,11 @@
 NAME Sum casted value
-PROG uprobe:./testprogs/intptrcast:fn { @=sum(*(uint16*)(reg("sp")+8)); exit();}
-EXPECT @: 3258
+PROG uprobe:./testprogs/intptrcast:fn { @=sum(*((uint16*)arg0 + 1)); exit();}
+EXPECT @: 1110
 TIMEOUT 5
-ARCH x86_64
-AFTER ./testprogs/intptrcast
-
-NAME Sum casted value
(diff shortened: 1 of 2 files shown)


tests/runtime/intptrcast
```diff
@@ -1,83 +1,11 @@
...
-PROG uprobe:./testprogs/intptrcast:fn { $a=*(uint16*)(reg("r1")+112); printf("%d\n", $a); exit();}
-EXPECT 3258
-TIMEOUT 5
-ARCH ppc64
 AFTER ./testprogs/intptrcast
 
 NAME Casting ints
-PROG uprobe:./testprogs/intptrcast:fn { $a=*(uint16*)(reg("r15")+160); printf("%d\n", $a); exit();}
-EXPECT 4660
+PROG uprobe:./testprogs/intptrcast:fn { $a=*((uint16*)arg0 + 1); printf("0x%x\n", $a); exit();}
```
viktormalik, line 8: IIUC, the only difference between the two remaining tests is that this one uses prints the number in hex form. Perhaps reflect that in the test names?

ajor: The key difference is that one test puts the casted result into `sum()` and the other into a variable. They're the original two tests added with the pointer casting feature(#942).

I just changed the printf to output as hex to make it easier for me to read the expected results. There's no way to do the same for the `sum` test though - if that's confusing I'll put the printf one back to outputting in decimal too.

viktormalik (approved):

PR #3168, opened 2024-05-15
CollectNodes: Recurse into children of the target type

ajor:
The test case `CollectNodes.nested` would have failed before this patch, with `visitor.nodes()` returning an empty list.

##### Checklist

### DIFF
--- a/tests/collect_nodes.cpp
+++ b/tests/collect_nodes.cpp
@@ -1,0 +1,118 @@
+#include "ast/passes/collect_nodes.h"
+#include "gtest/gtest.h"
+
+#include <functional>
+#include <vector>
+
+namespace bpftrace::test::collect_nodes {
+
+using namespace bpftrace::ast;
+
+template <typename T>
+void test(const std::vector<std::reference_wrapper<T>> &expected,
+          const std::vector<std::reference_wrapper<T>> &actual)
+{
+  ASSERT_EQ(expected.size(), actual.size());
+  for (size_t i = 0; i < expected.size(); i++) {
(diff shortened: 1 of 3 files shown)


tests/collect_nodes.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+  ASSERT_EQ(expected.size(), actual.size());
+  for (size_t i = 0; i < expected.size(); i++) {
+    EXPECT_EQ(&expected[i].get(), &actual[i].get());
+  }
+}
+
+TEST(CollectNodes, direct)
+{
+  auto &var = *new Variable{ "myvar", bpftrace::location{} };
```
viktormalik, line 23: Why use `new` instead of the below?
```suggestion
  auto var = Variable("myvar", bpftrace::location{});
```
Also does GTest automatically cleanup the allocated memory?

viktormalik (approved):
LGTM, just one question below.

tests/collect_nodes.cpp
ajor: I wish I didn't have to - `new` is horrible :(

It's because the AST expects nodes to have been created on the heap and tries to delete them in the destructors: https://github.com/bpftrace/bpftrace/blob/d592be8fe65d7b49aa183eb1a05625df1743ed0c/src/ast/ast.cpp#L94

I started with the other approach but got crashes from "double frees" when `delete` was run on the objects.

viktormalik: Ah, that's not too nice.

viktormalik:
Will probably need rebase to the fixed master.

PR #3174, opened 2024-05-16
Fix runtime tests using ustacks

viktormalik:
Several tests using the `ustack` builtin fail on my local setup. I found two issues:
- Some tests in `call.ustack*` do not expect the stack entries to be indented by whitespace. These tests happen to be the same as the ones disabled in the CI by #3078 so the issue was never discovered.
- The `dwarf.uprobe inlined function - ustack` gives me a slightly different stack from what is expected. It seems to expect something like:
    ```
        main+21
        0x7fdff76fb14a
    ```
    but I'm getting
    ```
        main+21
        0x7fdff76fb14a
        0x7fdff76fb20b
        _start+37
    ```
    IMHO, the most important part is to check that the "main+21" call occurs at the top of the stack and everything below is not that important.

This fixes the two issues.

##### Checklist

### DIFF
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -229,7 +229,7 @@ AFTER ./testprogs/uprobe_loop
 NAME ustack_stack_mode_env_bpftrace
 PROG u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }
 ENV BPFTRACE_STACK_MODE=bpftrace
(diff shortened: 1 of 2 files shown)


viktormalik:
@ttreyer could you please have a look at this one? I'm particularly not sure if the second fix is correct. Thanks!

ttreyer:
They both LGTM 🙂

ajor (approved):

PR #3178, opened 2024-05-16
Codegen: Memset(0) memory of variables if the data to be written is not full-size

### LINKED ISSUE #3172: Assigning string literals to variables of different size doesn't clear old data
This yields surprising results when such variables are used as map keys. For example, the following program appears to produce duplicate map entries:

```
$ sudo bpftrace -
kprobe:try_to_wake_up {
  $comm = comm;
  if (pid == 0) {
    $comm = "swapper";
    @[$comm] = count();
  }
}

Attaching 1 probe...
^C
[...]
@[swapper]: 355
@[swapper]: 402
@[swapper]: 769
```

The issue here is that the type of `$comm` is `string[16]`, but assigning a string literal of length N to it only overwrites the first N+1 bytes.

Here's the relevant part of the generated LLVM IR (note the 8-byte memcpy):
```
  %"$comm" = alloca [16 x i8], align 1
  [...]
  store [8 x i8] c"swapper\00", ptr %str, align 1
  call void @llvm.memcpy.p0.p0.i64(ptr align 1 %"$comm", ptr align 1 %str, i64 8, i1 false)
  [...]
  %lookup_elem = call ptr inttoptr (i64 1 to ptr)(ptr @AT_, ptr %"$comm")
```

Since the map key type is also `string[16]`, all lookup/update operations are performed using all 16 bytes of the `$comm` variable, which looks like "swapper\0$NUM" (i.e. it includes data from the previous call to `bpf_get_current_comm`).

A potential fix is to emit a call to `memset()` to clear the destination variable prior to assigning a string literal.

```
$ bpftrace --version
pftrace v0.20.0-gc8fd
```

ajor:
Fixes #3172

```
BEGIN
{
  $x = "xxxxx";
  $x = "a";
}
```

The value stored in `$x` at the end of the above script should be "a\0\0\0\0\0". Previously `$x` would have held "a\0xxx" as the second assignment did not overwrite all the data from the first.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@ and this project adheres to
   - [#2645](https://github.com/bpftrace/bpftrace/pull/2645)
 - Fix ustack for functions with prologue
   - [#3095](https://github.com/bpftrace/bpftrace/pull/3095)
+- Fix storing strings of differing lengths in a variable
+  - [#3178](https://github.com/bpftrace/bpftrace/pull/3178)
 #### Docs
 #### Tools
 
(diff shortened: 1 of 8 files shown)


jordalgo (approved):
Nice.

viktormalik (approved):

PR #3179, opened 2024-05-17
Use direct inc/dec for ++/-- op

jordalgo:
This will save a call to bpf_map_update_elem

Issue: https://github.com/bpftrace/bpftrace/issues/3175

Also make the add operation in CreateMapElemAdd atomic.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1519,7 +1519,9 @@ void IRBuilderBPF::CreateMapElemAdd(Value *ctx,
                                     Map &map,
                                     Value *key,
                                     Value *val,
-                                    const location &loc)
+                                    const location &loc,
+                                    AllocaInst *pre_post_val,
+                                    bool is_post_op)
 {
   CallInst *call = CreateMapLookup(map, key);
   SizedType &type = map.type;
@@ -1546,7 +1548,16 @@ void IRBuilderBPF::CreateMapElemAdd(Value *ctx,
 
   // createMapLookup  returns an u8*
   auto *cast = CreatePointerCast(call, value->getType(), "cast");
+
+  if (pre_post_val && is_post_op) {
+    CreateStore(CreateLoad(getInt64Ty(), cast), pre_post_val);
+  }
+
   CreateStore(CreateAdd(CreateLoad(getInt64Ty(), cast), val), cast);
+
+  if (pre_post_val && !is_post_op) {
+    CreateStore(CreateLoad(getInt64Ty(), cast), pre_post_val);
+  }
 
   CreateBr(lookup_merge_block);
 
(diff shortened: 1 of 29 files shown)


netoptimizer:
(Maybe for another PR)

Looking at above BPF byte-code.
- Why use a **hash-map** for storing this value?

This could easily be a simple array-map.
The hash-map comes with unnecessary overhead.

jordalgo:
> This could easily be a simple array-map.
The hash-map comes with unnecessary overhead.

Yeah true. 

Additionally, I think we might be causing data races in this implementation by updating the pointer directly instead of utilizing `__sync_fetch_and_add` for these operations to keep them atomic.

netoptimizer:
> 
> Additionally, I think we might be causing data races in this implementation by updating the pointer directly instead of utilizing `__sync_fetch_and_add` for these operations to keep them atomic.

True, I think it would be preferred that we use atomic inc/dec operations, as it is correct that data races still exists without them (still after removing the unnecessary `map_update_elem`).

As an end-user of bpftrace, I (falsely) assumed/expected these were **atomic** inc/dec operators (++/--).

danobi:
Hashmap is indeed suboptimal. We can definitely do some additional specialization here: https://github.com/bpftrace/bpftrace/blob/0d06553e98833fa014b514bd620701623ea65607/src/ast/passes/codegen_llvm.cpp#L3534

As for ++ vs count(), we have in the docs some details about that -- we can make the difference more clear though. https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#count

I'm not sure using atomic ops for `++/--` is best, though. I think you can still lose concurrent updates, right? And using atomic op in non-contended case is additional overhead I believe (HW locks cache line). To play devil's advocate, in C you wouldn't expect ++/-- to do anything special, right?

In your comment in https://github.com/xdp-project/xdp-project/commit/010d6b13cc613fb7df4cfe0a1b314514c8c67bea, I see your motivation is to be able to printf() a value, right? As a workaround, you could consider doing:

```
 $ sudo ./build/src/bpftrace -e 'BEGIN { @ = count(); print(("val", @)) }'
Attaching 1 probe...
(val, 1)
^C

@: 1

```

The reason you can't print a map with printf() yet is b/c printf() is synchronous whereas for print() the values we pull out are asynchronous. I believe it should be possible to synchronously print out count() cuz the kernel supports loops over maps elements now. We should support this.

danobi:
I've filed https://github.com/bpftrace/bpftrace/issues/3187 and https://github.com/bpftrace/bpftrace/issues/3186

jordalgo:
@danobi We already have this issue filed for our per-cpu functions: https://github.com/bpftrace/bpftrace/issues/3126

> I'm not sure using atomic ops for ++/-- is best, though. And using atomic op in non-contended case is additional overhead I believe (HW locks cache line). 

Yes, you're right. I didn't realize all the places we were calling `CreateMapElemAdd` were using per-cpu maps, which make atomic operations unnecessary. I'll probably close this PR for now and fix up inc/dec after working on https://github.com/bpftrace/bpftrace/issues/3126

PR #3183, opened 2024-05-17
config: Specify the type of str_trunc_trailer explicitly

chenhengqi:
Built bpftrace with LLVM/clang 16 and ran some random scripts, I got the following error:

    $ sudo bpftrace -e '...'
    Attaching 1 probe...
      what():  Type mismatch for config key
    [1]    1297304 abort      sudo ./bpftrace -e

The coredump reveals a failed type deduction:

    (gdb) p config_map_
    ...
      [std::variant<bpftrace::ConfigKeyBool, bpftrace::ConfigKeyInt, bpftrace::ConfigKeyString, bpftrace::ConfigKeyStackMode, bpftrace::ConfigKeyUserSymbolCacheType> [index 2] = {bpftrace::ConfigKeyString::str_trunc_trailer}] = {source = bpftrace::ConfigSource::default_,
        value = std::variant<bool, unsigned long, std::string, bpftrace::StackMode, bpftrace::UserSymbolCacheType> [index 0] = {true}},
    ...

Let's specify the type of str_trunc_trailer to avoid such error.

##### Checklist

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -23,7 +23,7 @@ Config::Config(bool has_cmd, bool bt_verbose) : bt_verbose_(bt_verbose)
     { ConfigKeyInt::max_type_res_iterations, { .value = (uint64_t)0 } },
     { ConfigKeyInt::perf_rb_pages, { .value = (uint64_t)64 } },
     { ConfigKeyStackMode::default_, { .value = StackMode::bpftrace } },
-    { ConfigKeyString::str_trunc_trailer, { .value = ".." } },
+    { ConfigKeyString::str_trunc_trailer, { .value = std::string("..") } },
     // by default, cache user symbols per program if ASLR is disabled on system
     // or `-c` option is given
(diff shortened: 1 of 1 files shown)


jordalgo:
Ah interesting. So it's treating the string as char * or something else?

jordalgo (approved):

chenhengqi:
No idea, I am not C++ expert. Here is a reproducer:

```c++
#include <string>
#include <variant>
#include <iostream>

int main()
{
	std::variant<bool, std::string> v;
	v = "..";
	auto b = std::get<bool>(v);
	std::cout << b << std::endl;
	auto s = std::get<std::string>(v); // throw std::bad_variant_access
	std::cout << s << std::endl;
	return 0;
}
```

I tried with clang++ 16 and g++ 9.4.0, both failed.

jordalgo:
Thanks for the fix!

PR #3184, opened 2024-05-17
Fix build on aarch64

jordalgo:
This fixes this error: "impossible constraint in ‘asm’" by removing the raw asm assembly call in inline_function.c.

##### Checklist


jordalgo:
Thanks @ttreyer for the code.

danobi (approved):
Looks like this works for -O0. -O3 compiler seems to figure it out https://godbolt.org/z/7cTqjj46W

jordalgo:
If we find that this test is failing in the future, we can just remove the assertions involving the non-inline function call.

PR #3185, opened 2024-05-17
Fix runtime tests for aarch64

ttreyer:
This PR fixes the following tests, that are failing when running on ARM:

* Update `CreateInteger` to accept 128-bits integers.
  The ARM specific [`struct user_fpsimd_state`](https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/uapi/asm/ptrace.h#L95) has a 128-bits integers.
* Update DWARF parser to consider arrays of `unsigned char` as strings.
  The `char`s from "tests/data/data_source.c"'s `struct Arrays::char_arr` have the Basic Type `lldb::eBasicTypeUnsignedChar`. Is this valid? Is there a bug in GCC for generating wrong DWARF?
* Force frame pointers for each `testprogs` and `testlibs`.
  The compiler doesn't have to set-up a frame pointer for leaf functions. We enforce their generation with `-fno-omit-frame-pointer` AND `-mno-omit-leaf-frame-pointer`.
* Ignore `STT_NOTYPE` symbols from the symbol table.
  ARM binaries might contain [mapping symbols](https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#mapping-symbols), which are not relevant to bpftrace. These symbols messes with symbolication, as they might have the same address as other symbols.
* Fix runtime test `other.positional ustack`.
  The expected regex would not tolerate extra stack trace. By using the `s` flag with `.*`, we accept any number of extra lines, while being strict enough to validate that `uprobeFunction1` is on top of the last stack trace.

After these fixes, the following runtime tests still fail:
* `probe.kprobe_offset_fail_size`: the function `AttachedProbe::resolve_offset_kprobe` cannot locate `vmlinux`, because my machine only has `vmlinuZ`, and thus can't check the boundaries of `vfs_read`. The Kernel didn't reject the probe `vfs_read+100000`. Shouldn't this function use `/proc/kallsyms` instead?
* `call.path`: The `path` builtin function returns path with the form `/dev/pts/4`, but never returns the expected path of the temporary file created for the test. Is this a Kernel bug?
* `call.strcontains`: This test uses the `path` builtin function and fails because of the issue above.

##### Checklist

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -144,7 +144,7 @@ std::string addrspacestr(AddrSpace as)
 std::string typestr(Type t)
 {
   switch (t) {
-    // clang-format off
+      // clang-format off
     case Type::none:     return "none";     break;
     case Type::voidtype: return "void";     break;
     case Type::integer:  return "integer";  break;
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -252,14 +252,8 @@ uint64_t asyncactionint(AsyncAction a)
 // Type wrappers
 SizedType CreateInteger(size_t bits, bool is_signed)
 {
-  // Zero sized integers are not usually valid. However, during semantic
-  // analysis when we're inferring types, the first pass may not have
-  // enough information to figure out the exact size of the integer. Later
-  // passes infer the exact size.
-  assert(bits == 0 || bits == 1 || bits == 8 || bits == 16 || bits == 32 ||
-         bits == 64);
   auto t = SizedType(Type::integer, 0, is_signed);
-  t.size_bits_ = bits;
+  t.SetIntBitWidth(bits);
   return t;
 }
 
--- a/src/types.h
+++ b/src/types.h
@@ -240,13 +240,24 @@ public:
     return size_bits_ / 8;
   }
 
-  void SetSize(size_t size)
+  void SetSize(size_t byte_size)
   {
-    size_bits_ = size * 8;
-    if (IsIntTy()) {
-      assert(size == 0 || size == 1 || size == 8 || size == 16 || size == 32 ||
-             size == 64);
-    }
+    if (IsIntTy())
+      SetIntBitWidth(byte_size * 8);
+    else
+      size_bits_ = byte_size * 8;
+  }
+
+  void SetIntBitWidth(size_t bits)
+  {
+    assert(IsIntTy());
+    // Zero sized integers are not usually valid. However, during semantic
+    // analysis when we're inferring types, the first pass may not have
+    // enough information to figure out the exact size of the integer. Later
+    // passes infer the exact size.
+    assert(bits == 0 || bits == 1 || bits == 8 || bits == 16 || bits == 32 ||
+          bits == 64 || bits == 128);
+    size_bits_ = bits;
   }
 
   size_t GetIntBitWidth() const
--- a/src/types.h
+++ b/src/types.h
@@ -251,12 +251,15 @@ public:
   void SetIntBitWidth(size_t bits)
   {
     assert(IsIntTy());
+    // Truncate integers too large to fit in BPF registers (64-bits).
+    if (bits > 64)
+      bits = 64;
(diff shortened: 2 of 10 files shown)


src/types.cpp
```diff
@@ -256,7 +256,7 @@ SizedType CreateInteger(size_t bits, bool is_signed)
   // enough information to figure out the exact size of the integer. Later
   // passes infer the exact size.
   assert(bits == 0 || bits == 1 || bits == 8 || bits == 16 || bits == 32 ||
-         bits == 64);
+         bits == 64 || bits == 128);
```
ajor, line 259: Do you know what happens in codegen if something tries to use a 128-bit integer?

ajor (commented):
Mostly looks good but I have a worry that that relaxing the assertion to allow 128-bit integers could cause problems in other places not expecting it

jordalgo:
Confirmed that `other.positional ustack` works on my Arm VM with this change.

src/types.cpp
ttreyer: Let's add a test and figure it out...

ttreyer: @ajor Following up on our chat, I modified `SizedType` to accept 128-bits integers. However, they get truncated down to only 64-bits with a warning to the user. The warning is done by the `semantic_analyser`, then the `codegen_llvm` insert the truncate for a `FieldAccess` on a user-defined type, if required.

src/types.h
```diff
@@ -240,13 +240,24 @@ class SizedType {
...
+  void SetIntBitWidth(size_t bits)
+  {
+    assert(IsIntTy());
+    // Zero sized integers are not usually valid. However, during semantic
+    // analysis when we're inferring types, the first pass may not have
+    // enough information to figure out the exact size of the integer. Later
+    // passes infer the exact size.
+    assert(bits == 0 || bits == 1 || bits == 8 || bits == 16 || bits == 32 ||
+           bits == 64 || bits == 128);
+    size_bits_ = bits;
```
ajor, line 260: Could we put the 128-bit integer truncation in here? That should guarantee that we catch every place an integer's size is set

ajor (changes requested):
Looks like it's possible to get a 128-bit integer through to codegen and have it trigger a seg-fault:
```
void uprobeFunction3(__int128_t bigint) {}
```

```
uprobe:./build/tests/testprogs/uprobe_test:uprobeFunction3 {
  print(args.bigint);
}
```

I think for simplicity, codegen shouldn't know anything about 128-bit integers at all - they should get truncated in the type system before they reach that point.

src/types.h
ttreyer: I've updated the code accordingly 🙂 
However, we currently don't handle multiple 128-bits arguments correctly. See the commented-out `EXPECT` in `tests/runtime/dwarf`'s 128-bits tests.

ttreyer: I've documented the issue in #3236.

tests/runtime/dwarf
```diff
@@ -45,6 +45,35 @@ REQUIRES_FEATURE dwarf
...
 BEFORE ./testprogs/uprobe_test
 
+NAME uprobe arg by index - 128-bits integer
+PROG uprobe:./testprogs/uprobe_test:uprobeFunctionUint128 { printf("x = %x\ny = %x\nz = %x\nw = %x\n", arg0, arg1, arg2, arg3); exit(); }
+EXPECT x = 9abcdef0
+       y = efefefef
+       z = cdcdcdcd
+       w = abababab
+REQUIRES_FEATURE dwarf
+REQUIRES bash -c "exit 1" # SKIP: 128-bits arguments are not supported, see #3236
```
jordalgo, line 55: Nit: I'm not sure if I like the use of REQUIRES for tests that we don't support at all (yet). Maybe it's better to just comment out the whole thing and add a TODO with the issue. It's more noticeable when looking at the code IMHO.

ajor (approved):
Looks good!

PR #3189, opened 2024-05-20
Add implicit cast for 'sum' and 'count'

jordalgo:
This utilizes 'map_lookup_percpu_elem' to iterate over the per-cpu maps that 'sum' and 'count' use to aggregate the values for each cpu so these maps can be used in kernel space for conditionals and arithmetic.

Additionally, add explicit casting for printing or storing just the value for these types of maps.

Issue: https://github.com/bpftrace/bpftrace/issues/3126

Note: We can do the same to support to 'avg', 'max', 'min' but those require the same user-space logic written in llvm ir.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -2415,9 +2415,12 @@ i:s:10 {
 Count how often this function is called.
 
 Using `@=count()` is conceptually similar to `@++`.
-The difference is that the `count()` function uses a map type optimized for this (PER_CPU), increasing performance and correctness.
-Multiple writers to a shared global var might lose counts as bpftrace doesn't update them atomically.
-Due to this the map cannot be accessed as a regular integer.
+The difference is that the `count()` function uses a map type optimized for 
+writing (PER_CPU), increasing performance and correctness. However, sync reads 
+can be expensive as bpftrace needs to iterate over all the cpus to collect and 
+sum these values.
+Note: In contrast to hash maps (e.g. `@++`), multiple writers to a shared 
+global var might lose counts as bpftrace doesn't update them atomically.
 
 ----
 i:ms:100 {
@@ -2425,7 +2428,12 @@ i:ms:100 {
 }
 
 i:s:10 {
+  // async read
   print(@);
+  // sync read
+  if (@ > 10) {
+    print(("hello"));
+  }
   clear(@);
 }
 ----
@@ -2574,6 +2582,30 @@ kprobe:vfs_read {
 * `sum(int64 n)`
 
 Calculate the sum of all `n` passed.
+
+Using `@=sum(5)` is conceptually similar to `@+=5`.
+The difference is that the `sum()` function uses a map type optimized for 
+writing (PER_CPU), increasing performance and correctness. However, sync reads 
+can be expensive as bpftrace needs to iterate over all the cpus to collect and 
+sum these values.
+Note: In contrast to hash maps (e.g. `@+=5`), multiple writers to a shared 
+global var might lose updates as bpftrace doesn't update them atomically.
+
+----
+i:ms:100 {
+  @ = sum(5);
+}
+
+i:s:10 {
+  // async read
+  print(@);
+  // sync read
+  if (@ > 10) {
+    print(("hello"));
+  }
+  clear(@);
+}
+----
 
 [#map-functions-zero]
 === zero
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -2,6 +2,7 @@
 
 #include <iostream>
 #include <sstream>
+#include <thread>
 
 #include <llvm/IR/DataLayout.h>
 #include <llvm/IR/Module.h>
@@ -326,6 +327,35 @@ CallInst *IRBuilderBPF::createMapLookup(const std::string &map_name,
   return createCall(lookup_func_type, lookup_func, { map_ptr, key }, name);
 }
 
+CallInst *IRBuilderBPF::createPerCpuMapLookup(const std::string &map_name,
+                                              Value *key,
+                                              Value *cpu,
+                                              const std::string &name)
+{
+  return createPerCpuMapLookup(map_name, key, cpu, GET_PTR_TY(), name);
+}
+
+CallInst *IRBuilderBPF::createPerCpuMapLookup(const std::string &map_name,
+                                              Value *key,
+                                              Value *cpu,
+                                              PointerType *val_ptr_ty,
+                                              const std::string &name)
+{
+  Value *map_ptr = GetMapVar(map_name);
+  // void *map_lookup_percpu_elem(struct bpf_map * map, void * key, u32 cpu)
+  // Return: Map value or NULL
+
+  assert(key->getType()->isPointerTy());
+  FunctionType *lookup_func_type = FunctionType::get(
+      val_ptr_ty, { map_ptr->getType(), key->getType(), getInt32Ty() }, false);
+  PointerType *lookup_func_ptr_type = PointerType::get(lookup_func_type, 0);
+  Constant *lookup_func = ConstantExpr::getCast(
+      Instruction::IntToPtr,
+      getInt64(libbpf::BPF_FUNC_map_lookup_percpu_elem),
+      lookup_func_ptr_type);
+  return createCall(lookup_func_type, lookup_func, { map_ptr, key, cpu }, name);
+}
+
 CallInst *IRBuilderBPF::CreateGetJoinMap(BasicBlock *failure_callback,
                                          const location &loc)
 {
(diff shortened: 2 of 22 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -1790,7 +1790,8 @@ void SemanticAnalyser::visit(Binop &binop)
     return;
   }
 
-  if (lht.IsIntTy() && rht.IsIntTy()) {
+  if ((lht.IsCountTy() || lht.IsSumTy() || lht.IsIntTy()) &&
+      (rht.IsCountTy() || rht.IsSumTy() || rht.IsIntTy())) {
```
danobi, line 1794: I'm wondering if we can be more generic here. For example, this doesn't quite fix Jesper's issue where he wanted to printf() the count() value. 

Perhaps we can allow casts to int64/int32/etc from Type::count? I kinda like the idea of explicit casts cuz it'll at least hint we're gonna do something magical. Eg.

```
@ = count();
printf("%d\n", (int64)@);
```

WDYT?

jordalgo:
I'll update the changelog and adoc once the code is approved.

src/ast/passes/semantic_analyser.cpp
jordalgo: Yeah, I like explicit casts - let me look into supporting that.

jordalgo:
Note: Also going to add a codegen test for this

src/ast/passes/semantic_analyser.cpp
jordalgo: Added this in the latest version.

jordalgo:
Update: I added a bit of a hack to make this work for AOT. Basically, when in AOT mode, just set an upper max on the number of cpus to iterate over (1024 - which is still below some crazy new machines but the verifier won't freak out) and exit the loop early if we fail to look up the value for that cpu. I tested this locally and it worked as expected.

src/ast/irbuilderbpf.cpp
```diff
@@ -450,6 +480,169 @@ Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
...
+  assert(ctx && ctx->getType() == GET_PTR_TY());
+
+  const std::string &map_name = map.ident;
+
+  AllocaInst *sum = CreateAllocaBPF(getInt64Ty(), "sum");
+  AllocaInst *i = CreateAllocaBPF(getInt32Ty(), "i");
+
+  // Set a large upper bound if we don't know the number of cpus
+  // when generating the instructions
+  int nr_cpus = is_aot ? 1024 : bpftrace_.get_num_possible_cpus();
```
danobi, line 551: I forget if we discussed this in person, did it not work to pass the nr_cpu from userspace and then mask (`&=`) the count?

If it didn't and we have to use this method, since 1024 is a fixed value, can bpftrace check at runtime if there's > 1024 CPUs and warn about it? This is a potential for incorrect results & and we should be really careful in these situations. Should be possible in a shared codepath between AOT and dynamic mode.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1415,7 +1416,20 @@ void CodegenLLVM::visit(Offsetof &ofof)
...
-  Value *value = b_.CreateMapLookupElem(ctx_, map, key, map.loc);
+
+  auto map_info = bpftrace_.resources.maps_info.find(map.ident);
+  if (map_info == bpftrace_.resources.maps_info.end()) {
+    LOG(BUG) << "map name: \"" << map.ident << "\" not found";
+  }
+
+  const auto &val_type = map_info->second.value_type;
+  Value *value;
+  if (val_type.IsCountTy() || val_type.IsSumTy()) {
```
danobi, line 1427: Can we add a helper method (maybe in types.h?) that groups cast-able map types? It'll help make things more clear across semantic analysis && codegen

danobi (commented):
could we add some semantic analysis test cases to check that types are coerced or casted where we expect?

tests/codegen/per_cpu_map_cast.cpp
danobi, line 1: Could we add a codegen test for `BEGIN { @ = count(); print(@) }` ? This would help ensure that we don't loop unless we actually need to.

src/ast/irbuilderbpf.cpp
```diff
@@ -326,6 +327,35 @@ CallInst *IRBuilderBPF::createMapLookup(const std::string &map_name,
...
+
+  assert(key->getType()->isPointerTy());
+  FunctionType *lookup_func_type = FunctionType::get(
+      val_ptr_ty, { map_ptr->getType(), key->getType(), getInt32Ty() }, false);
+  PointerType *lookup_func_ptr_type = PointerType::get(lookup_func_type, 0);
+  Constant *lookup_func = ConstantExpr::getCast(
+      Instruction::IntToPtr,
+      getInt64(libbpf::BPF_FUNC_map_lookup_percpu_elem),
+      lookup_func_ptr_type);
+  return createCall(lookup_func_type, lookup_func, { map_ptr, key, cpu }, name);
```
danobi, line 356: I think we should add a `CreateHelperErrorCond` call here too

src/ast/irbuilderbpf.h
```diff
@@ -89,6 +89,18 @@ class IRBuilderBPF : public IRBuilder<> {
                              Value *key,
                              SizedType &type,
                              const location &loc);
+
+  Value *CreatePerCpuMapLookupElem(Value *ctx,
+                                   const std::string &map_name,
+                                   Value *key,
+                                   Value *cpu,
+                                   SizedType &type,
+                                   const location &loc);
```
danobi, line 98: Is this unused? I deleted it from my local checkout and things still compile && work

src/ast/irbuilderbpf.cpp
```diff
@@ -450,6 +480,169 @@ Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
...
+  // If the CPU is 0 and the map lookup fails it means the key doesn't exist
+  Value *error_condition = CreateICmpEQ(CreateLoad(getInt32Ty(), i),
+                                        getInt32(0),
+                                        "error_lookup_cond");
+  CreateCondBr(error_condition, error_success_block, error_failure_block);
+
+  SetInsertPoint(error_success_block);
+
+  CreateHelperError(
+      ctx, getInt32(0), libbpf::BPF_FUNC_map_lookup_percpu_elem, loc);
```
danobi, line 626: Should `getInt32(0)` be -ENOENT ?

src/ast/irbuilderbpf.h
jordalgo: Good catch. I think this was leftover from a previous revision

src/ast/irbuilderbpf.cpp
jordalgo: This mirrors what we do for `CreateMapLookupElem` as the return value is null if the map element doesn't exist.

src/ast/irbuilderbpf.cpp
jordalgo: We don't do this for `createMapLookup` because the return value is the map value or null not an error code.

src/ast/irbuilderbpf.cpp
jordalgo: > I forget if we discussed this in person, did it not work to pass the nr_cpu from userspace and then mask (&=) the count?

I've tried other limiting techniques in other PRs and it seems like the verifier is unhappy if it doesn't know the value in advance and the type can exceed the max number of loops. You can normally get around this using libbpf by having a mutable volatile with an initial value and modifying it using the skeleton in userspace but we don't have that yet. @ajor and I discussed the option of doing this by modifying the rodata section but decided just to punt on that since AOT isn't officially supported yet anyway. 

I will add a warning if we detect the number of cpus to be greater than 1024 in userspace.

jordalgo: I also have plans to add proper loops using open coded iterators which should also fix this issue.

src/main.cpp
```diff
@@ -875,6 +875,16 @@ int main(int argc, char* argv[])
...
 
+  if (args.build_mode == BuildMode::AHEAD_OF_TIME) {
+    auto num_cpus = bpftrace.get_num_possible_cpus();
+    if (num_cpus > 1024) {
+      LOG(WARNING) << "Detected " << num_cpus
+                   << " cpus. For ahead-of-time compilation there is a max of "
+                      "1024 cpus so there may be incorrect data for 'count' "
+                      "and 'sum' aggregations.";
+    }
+  }
```
danobi, line 886: I think this warning is only relevant if the _runtime_ host has >1024 CPUs. So if build host had >1024 CPUs but runtime had <1024, it'd still be ok. So I think this should go in (or near) `aot_main.cpp`. 

Having this hard coded message is probably OK if you plan on removing it soonish. Was thinking if it makes sense to analyze RequiredResources in aot.cpp::load_required_resources. It'd be more accurate -- we could check if count() or sum() is even being used. IMO it'd be a little cleaner to have it there. Up to you.

jordalgo: > So I think this should go in (or near) aot_main.cpp.

Ah right. Good catch. I'm still getting familiar with AOT :) 

> we could check if count() or sum() is even being used. IMO it'd be a little cleaner to have it there. Up to you.

I plan on not having this hack for too long so the less additional logic the better - besides are we really worried about the case where someone is running this on a machine with over 1024 cores?

danobi: > I plan on not having this hack for too long so the less additional logic the better - besides are we really worried about the case where someone is running this on a machine with over 1024 cores?

I think it's reasonable to be worried about this case. Looks like we are getting users working on supercomputers now

danobi (approved):
would be good to get another pair of eyes on this as well.

btw, does it make sense to document this behavior? maybe not if the original behavior was surprising. perhaps just a note about how it is expensive to read count() or sum() maps, as they are optimized for writing and not reading

jordalgo:
> btw, does it make sense to document this behavior? 

Good call. I'll add something in the adoc and also make a changelog entry.

danobi:
Had some shower thoughts:

1. Can simplify semantic analyser with the type helper: https://pastes.dxuuu.xyz/4ilo7i
1. I mentioned this above; not sure if we had different opinions or if I wasn't being clear, but the special case in the binop semantic analysis is a tad odd. For example, with this: https://pastes.dxuuu.xyz/a2w36w , you can still do `BEGIN { @ = count(); if ((int64)@ == 1) { print("yes") } }`. However, the cast (and thus potentially expensive operation) is explicit and provides some symmetry with how you synchronously access map-backed types in the rest of the language. 

I do also find it odd that for-looping over a map that contains `count()` as a value immediately collapses the map into an integer. This is probably wrong IMO. A user could be doing a loop and only wanting to do expensive stuff if the key is some value they want. Collapsing the value on an entry they don't immediately care about is bad for perf.

The advantage with the explicit casts is that we can change our minds later and make it implicit. Whereas if we start with implicit, we can't really take it back.

jordalgo:
> Can simplify semantic analyser with the type helper

Good call. I'll update that.

> A user could be doing a loop and only wanting to do expensive stuff if the key is some value they care about. Collapsing the value on an entry they don't immediately care about is bad for perf. The advantage with the explicit casts is that we can change our minds later and make it implicit. Whereas if we start with implicit, we can't really take it back.

This is a good point. I also don't like the perf hit with implicit casting/collapsing - that being said I'm trying to think of use cases where a user is accessing a `count` or `sum` (or `min`, `max`, `avg`, which I plan on adding) value and just wants it for the current cpu. I feel like as long as the tradeoffs of using per-cpu map aggregations is documented (cheap writes, expensive reads) than the onus is on the user to choose the best map type for what they're doing with it. Also I feel like the auto-collapsing of a value inside a for-loop is better fixed by figuring out if the value is even being used and conditionally not collapsing it. In many ways I like when code is explicit but if we don't have good examples of using a `count` or `sum` value without collapsing then it just feels like cruft. I think this ties into your second point above but I need to think more on that.

danobi:
> > Can simplify semantic analyser with the type helper
> 
> Good call. I'll update that.
> 
> > A user could be doing a loop and only wanting to do expensive stuff if the key is some value they care about. Collapsing the value on an entry they don't immediately care about is bad for perf. The advantage with the explicit casts is that we can change our minds later and make it implicit. Whereas if we start with implicit, we can't really take it back.
> 
> This is a good point. I also don't like the perf hit with implicit casting/collapsing - that being said I'm trying to think of use cases where a user is accessing a `count` or `sum` (or `min`, `max`, `avg`, which I plan on adding) value and just wants it for the current cpu. I feel like as long as the tradeoffs of using per-cpu map aggregations is documented (cheap writes, expensive reads) than the onus is on the user to choose the best map type for what they're doing with it. 

Yeah, I buy this argument. I can't think of any use cases where you'd not want the collapsed value. To take that idea to its logical conclusion, I think https://pastes.dxuuu.xyz/iydqri would work. Basically remove all the special cases and treat reading a count() as an integer. On write side it'll still be protected:

```
$ sudo ./build/src/bpftrace -e 'BEGIN { @x = count(); $xx = @x; $xx++; print($xx) }'
Attaching 1 probe...
2
^C

@x: 1
$ sudo ./build/src/bpftrace -e 'BEGIN { @x = count(); $xx = @x; $xx++; print($xx); @x = 5 }'
[sudo] password for dxu:
stdin:1:52-54: ERROR: Type mismatch for @x: trying to assign value of type 'int64' when map already contains a value of type 'count
BEGIN { @x = count(); $xx = @x; $xx++; print($xx); @x = 5 }

```

> Also I feel like the auto-collapsing of a value inside a for-loop is better fixed by figuring out if the value is even being used and conditionally not collapsing it. In many ways I like when code is explicit but if we don't have good examples of using a `count` or `sum` value without collapsing then it just feels like cruft. I think this ties into your second point above but I need to think more on that.

Presumably any loop that does filtering is on the slow or async path. So maybe this is good enough. Maybe in the long term once we enable user defined percpu maps then users could work around this if they really needed to.

man/adoc/bpftrace.adoc
```diff
@@ -2415,17 +2415,25 @@ i:s:10 {
...
 Using `@=count()` is conceptually similar to `@++`.
-The difference is that the `count()` function uses a map type optimized for this (PER_CPU), increasing performance and correctness.
-Multiple writers to a shared global var might lose counts as bpftrace doesn't update them atomically.
-Due to this the map cannot be accessed as a regular integer.
+The difference is that the `count()` function uses a map type optimized for 
+writing (PER_CPU), increasing performance and correctness. However, sync reads 
+can be expensive as bpftrace needs to iterate over all the cpus to collect and 
+sum these values.
+Note: Multiple writers to a shared global var might lose counts as bpftrace 
+doesn't update them atomically.
```
danobi, line 2423:
```suggestion
Note: In contrast, multiple writers to a shared global var might lose counts as bpftrace 
doesn't update them atomically.
```

jordalgo: Should it be "In contrast to `@++`, which uses a global hash map, multiple writers...

danobi: Maybe add it on as "(eg `@++`)"

danobi:
Oh, I think we need a feature probe in semantic analyser as well for the map elem callback helper

jordalgo:
> Oh, I think we need a feature probe in semantic analyser as well for the map elem callback helper

Let me put up a separate PR for that.

danobi:
Chatted offline about turning count() and sum() reads into proper ints. Type system does not allow us to do this easily. So I think all outstanding issues modulo feature probing are addressed

jordalgo:
@ajor @viktormalik - Going to merge this soon unless you want more time to review.

src/ast/irbuilderbpf.cpp
```diff
@@ -450,6 +480,116 @@ Value *IRBuilderBPF::CreateMapLookupElem(Value *ctx,
   return ret;
 }
 
+Value *IRBuilderBPF::CreatePerCpuMapSumElems(Value *ctx,
```
viktormalik, line 483: This implements a non-trivial algorithm. Could we perhaps include a pseudo-code in the comment to make it clear what's the intention to do here?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2333,7 +2334,8 @@ void SemanticAnalyser::visit(Cast &cast)
   }
 
   if ((cast.type.IsIntTy() && !rhs.IsIntTy() && !rhs.IsPtrTy() &&
-       !rhs.IsCtxAccess() && !(rhs.IsArrayTy())) ||
+       !rhs.IsCtxAccess() && !rhs.IsArrayTy() && !rhs.IsCountTy() &&
+       !rhs.IsSumTy()) ||
```
viktormalik, line 2338: like this?
```suggestion
       !rhs.IsCtxAccess() && !rhs.IsArrayTy() && !rhs.IsCastableMapType()) ||
```

tests/codegen/per_cpu_map_cast.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+TEST(codegen, count_no_cast_for_print)
+{
+  test("BEGIN { @ = count(); print(@) }", NAME);
+}
+
+} // namespace codegen
+} // namespace test
+} // namespace bpftrace
```
viktormalik, line 37: Missing EOL

viktormalik (approved):
LGTM, just a few nits

Issue #3144, opened 2024-04-29
RFC: Add remediation for BPF ABBA deadlock bug caused by hash map access


jordalgo:
There was an issue discovered recently via a bpftrace script whereby a kfunc was attached on the same path used to access BPF_MAP_TYPE_HASH which created an ABBA deadlock in the kernel and a subsequent crash e.g.

```
kfunc:queued_spin_lock_slowpath /@pagefault_start[tid]/
{
	@spinlock_start[tid] = nsecs;
}
```

`@spinlock_start` was being accessed by another bpf prog on a different CPU.

Example stack trace:
```
#0  bpf_trampoline_6442467238_1+0x74/0x1000
#1  queued_spin_lock_slowpath+0x5/0x200
#2  _raw_spin_lock_irqsave+0x4e/0x60
#3  htab_map_update_elem+0xc7/0x4b0
#4  bpf_prog_717c51531607df92_kretfunc_vmlinux_down_read+0x180/0x234
#5  bpf_trampoline_6442467155_1+0xa1/0x1000
#6  down_read+0x5/0x10
#7  do_user_addr_fault+0x294/0x740
#8  exc_page_fault+0x5e/0x110
#9  asm_exc_page_fault+0x1e/0x30
```

This issue is being addressed in a [bpf kernel patch](https://git.kernel.org/pub/scm/linux/kernel/git/ast/bpf.git/commit/?h=arena_lock&id=046bbea51a8dba6cc18bdb62fd48518b8cc220ad) however because this is going into a later kernel release we should consider adding a temp fix for this in bpftrace.

One possible option, suggested by Alexei, is for a bpftrace script to create a per-cpu variable that is tested when a functional block is called. If it is already set, we exit early (and increment a missed counter?), if it is not yet set, set it, run the functional block, and clear it at exit time. We could do this conditionally if we detect that the prog is accessing a non per-cpu map type.

Another option would be to block use of certain kfuncs/kprobe if bpftrace can detect if these progs are accessing non per-cpu maps in a script with other progs accessing this map (which may end up being more code).

This issue is meant to be a place to discuss possible solutions and the priority of this fix.

ajor:
The things I'm not keen on in the proposed bpftrace mitigations are:
1. We'll have no way to determine if the running kernel has the fixes applied or is vulnerable to this deadlock
2. Therefore the bpftrace mitigation will have to be applied to all scripts going forwards

- The per-cpu variable method will result in a performance overhead for all scripts
[...]

PR #3193, opened 2024-05-21
Prevent banned kernel funcs from all probes

jordalgo:
Previously we only banned kretprobes from
using banned kernel functions but a bpftrace
script at Meta saw a crash from utilizing
one of these functions in kfunc/kretfunc.

This changes prevents any probe from attaching
to one of these functions.

Issue: https://github.com/bpftrace/bpftrace/issues/3144

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -37,7 +37,7 @@ namespace bpftrace {
  * Kernel functions that are unsafe to trace are excluded in the Kernel with
  * `notrace`. However, the ones below are not excluded.
  */
-const std::set<std::string> banned_kretprobes = {
+const std::set<std::string> banned_kernel_funcs = {
   "_raw_spin_lock",
   "_raw_spin_lock_irqsave",
   "_raw_spin_unlock_irqrestore",
@@ -104,10 +104,10 @@ std::string progtypeName(libbpf::bpf_prog_type t)
   }
 }
 
-void check_banned_kretprobes(std::string const &kprobe_name)
+void check_banned_kernel_funcs(std::string const &kernel_func)
 {
-  if (banned_kretprobes.find(kprobe_name) != banned_kretprobes.end()) {
-    LOG(FATAL) << "kretprobe:" << kprobe_name
+  if (banned_kernel_funcs.find(kernel_func) != banned_kernel_funcs.end()) {
+    LOG(FATAL) << "kernel function: " << kernel_func
                << " can't be used as it might lock up your system.";
   }
 }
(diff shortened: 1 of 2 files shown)


src/attached_probe.cpp
```diff
@@ -37,7 +37,7 @@ namespace bpftrace {
  * Kernel functions that are unsafe to trace are excluded in the Kernel with
  * `notrace`. However, the ones below are not excluded.
  */
-const std::set<std::string> banned_kretprobes = {
+const std::set<std::string> banned_kernel_funcs = {
```
4ast, line 40: That doesn't really work. There are tracepoints as well and kprobe can be inside trace_* handling code.
Also kprobe shouldn't have any problem attaching to these funcs.
fentry need per-cpu recursion counter on bpftrace side

jordalgo: The changes in this PR were meant as a small stop gap as we're still discussing how to handle the per-cpu recursion counter discussed here: https://github.com/bpftrace/bpftrace/issues/3144

We obviously don't need to land this if you think this is a step backward but it would have prevented the crash we saw at Meta.

4ast: It seems like a code churn to land "stop gap" now just to revert it later when #3144 is ready. Not like 3144 is that hard. Just a per-cpu counter for fentry progs. I'd rather remove this kprobe attach restriction, since it seems unnecessary. I'm not sure what was the decision back then when it was added.

jordalgo: > Not like 3144 is that hard

It wasn't a question of difficulty. There was some concern over how long we'd have to maintain this code and if there would be a perf impact. @danobi mentioned having a feature probe so we know when we can delete this code but I don't think that exists yet.

Also, just curious, why don't we need this for kprobe/kretprobes?

4ast: the kernel fix for ABBA is wip. Even when it lands 3144 workaround in bpftrace would have to stay. Feature probing is nice to reduce overheard, but the overhead is tiny anyway. per-cpu array lookups are all inlined now.
kprobe cannot hit this issue, because there is already global per-cpu bpf_prog_active counter that protects all kprobes, but at the same time it makes kprobes unreliable with spurious misses. we cannot make all fentry-s unreliable like kprobes in the kernel, because they're used beyond tracing. At the same time bpftrace can add this check for fentrys, since bpftrace is a tracing tool. People won't be using bpftrace for security enforcement.

jordalgo: Got it. Thanks for the explanation! I'll wait to see if anyone else chimes in but if I don't hear anything I'll work on the per-cpu counter for fentry (need it for fexit as well? - doesn't seem so).

4ast: yes. both fentry/fexit.
btw to reduce overhead of extra checks in 3144 you can combine it with this diff.
Do per-cpu cnt only if fentry/fexit are attached to banned_kernel_funcs[]

PR #3195, opened 2024-05-22
Codegen: Don't emit lifetime.end for variables used as map keys

### LINKED ISSUE #3000: LLVM lifetimes ended too early
```
BEGIN
{
  $blah = "abc";
  @x[$blah] = 1;
  @y[$blah] = 1; 
}
```

The variable `$blah` should be alive throughout the entire duration of the program, but inspecting the generated IR shows that its lifetime is ended after writing to `@x`, before it is used again when writing to `@y`. Its lifetime is then ended for a second time after its last use.

Generated IR with lines using `$blah` prefixed with `*`:
```
define i64 @BEGIN(ptr nocapture readnone %0) local_unnamed_addr section "s_BEGIN_1" !dbg !3 {
entry:
  %"@y_val" = alloca i64, align 8
  %"@x_val" = alloca i64, align 8
* %"$blah" = alloca [4 x i8], align 4
* call void @llvm.lifetime.start.p0(i64 -1, ptr nonnull %"$blah")
* store i32 6513249, ptr %"$blah", align 4
  call void @llvm.lifetime.start.p0(i64 -1, ptr nonnull %"@x_val")
  store i64 1, ptr %"@x_val", align 8
  %pseudo = tail call i64 @llvm.bpf.pseudo(i64 1, i64 0)
* %update_elem = call i64 inttoptr (i64 2 to ptr)(i64 %pseudo, ptr nonnull %"$blah", ptr nonnull %"@x_val", i64 0)
  call void @llvm.lifetime.end.p0(i64 -1, ptr nonnull %"@x_val")
* call void @llvm.lifetime.end.p0(i64 -1, ptr nonnull %"$blah")
  call void @llvm.lifetime.start.p0(i64 -1, ptr nonnull %"@y_val")
  store i64 1, ptr %"@y_val", align 8
  %pseudo1 = call i64 @llvm.bpf.pseudo(i64 1, i64 1)
* %update_elem2 = call i64 inttoptr (i64 2 to ptr)(i64 %pseudo1, ptr nonnull %"$blah", ptr nonnull %"@y_val", i64 0)
  call void @llvm.lifetime.end.p0(i64 -1, ptr nonnull %"@y_val")
* call void @llvm.lifetime.end.p0(i64 -1, ptr nonnull %"$blah")
  ret i64 0
}
```

I'm not sure what the real impact of this bug is, but I imagine that it has the potential to produce wrong results through variables being incorrectly assigned to share the same stack address.

ajor:
Fixes #3000

Previously a lifetime.end was always inserted for expressions used as map keys. This is not correct if the expression is a variable which is used again later in the program.

`$blah` should not be marked as dead after line 4, it should persist at least until after line 5:
```
  1  BEGIN
  2  {
  3    $blah = "abc";
  4    @x[$blah] = 1;
  5    @y[$blah] = 1;
  6  }
```

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2775,6 +2775,7 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
     Map &map)
 {
   Value *key;
+  bool alloca_created_here = true;
   if (map.vargs) {
     // A single value as a map key (e.g., @[comm] = 0;)
     if (map.vargs->size() == 1) {
@@ -2791,6 +2792,10 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
           key = expr_;
           // Call-ee freed
           scoped_del.disarm();
+
+          // We don't have enough visibility into where key comes from to safely
+          // end its lifetime. It could be a variable, for example.
+          alloca_created_here = false;
         }
       } else {
         key = b_.CreateAllocaBPF(expr->type, map.ident + "_key");
@@ -2813,8 +2818,8 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
     b_.CreateStore(b_.getInt64(0), key);
   }
 
-  auto key_deleter = [this, key]() {
-    if (dyn_cast<AllocaInst>(key))
+  auto key_deleter = [this, key, alloca_created_here]() {
+    if (alloca_created_here)
       b_.CreateLifetimeEnd(key);
   };
   return std::make_tuple(key, ScopedExprDeleter(std::move(key_deleter)));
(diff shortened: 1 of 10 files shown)


jordalgo (approved):
Changes look good.

Out of curiosity, what is the error that surfaces if we run out of stack space in BPF?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2791,6 +2792,10 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
           key = expr_;
           // Call-ee freed
           scoped_del.disarm();
+
+          // We don't have enough visibility into where key comes from to safely
+          // end its lifetime. It could be a variable, for example.
```
danobi, line 2797:
```suggestion
          // end its lifetime. It could be a scratch variable, for example.
```

danobi (approved):
lgtm

danobi:
> Out of curiosity, what is the error that surfaces if we run out of stack space in BPF?

Verifier error

PR #3199, opened 2024-05-26
aot: Give preference to local build shim

danobi:
This helps with development such that we don't need to mess with $PATH
when running --aot commands. Now we can just run it like a prod
invocation and it will "just work".

##### Checklist

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -268,6 +268,26 @@ std::optional<std_filesystem::path> find_in_path(std::string_view name)
   return std::nullopt;
 }
 
+std::optional<std_filesystem::path> find_near_self(std::string_view filename)
+{
+  std::error_code ec;
+  auto exe = std_filesystem::read_symlink("/proc/self/exe", ec);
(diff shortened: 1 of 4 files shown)


danobi:
cc @arthurshau

src/utils.cpp
```diff
@@ -268,6 +268,26 @@ std::optional<std_filesystem::path> find_in_path(const std::string &name)
   return std::nullopt;
 }
 
+std::optional<std_filesystem::path> find_near_self(std::string_view filename)
+{
+  std::error_code ec;
+  auto exe = std_filesystem::read_symlink("/proc/self/exe", ec);
+  if (ec) {
+    LOG(ERROR) << "Failed to resolve /proc/self/exe: " << ec;
```
jordalgo, line 276: Nit: these look like recoverable errors so how about `LOG(WARNING)`

arthurshau:
LGTM!

viktormalik (approved):

PR #3200, opened 2024-05-26
aot: Fix distro builds

danobi:
We need to link aot.o against libelf if aot.cpp wants to actually use the dependency.

##### Checklist

### DIFF
--- a/src/aot/CMakeLists.txt
+++ b/src/aot/CMakeLists.txt
@@ -6,6 +6,12 @@ target_include_directories(aot PUBLIC ${CMAKE_BINARY_DIR}/src)
 target_include_directories(aot PUBLIC ${CMAKE_BINARY_DIR})
 target_compile_definitions(aot PRIVATE ${BPFTRACE_FLAGS})
 
+if(STATIC_LINKING)
+  target_link_libraries(aot LIBELF)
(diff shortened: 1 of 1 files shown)


jordalgo (approved):

danobi:
I _think_ this is correct. It passes the static build CI check and my local docker builds. Can fix up if anyone has a way do to this better

viktormalik:
Just out of curiosity, what was the error here? And why did it work for a local build and not for a distro one?

danobi:
Oh whoops, I thought I added the error to the commit msg.

```
66.34 /usr/bin/ld: aot/libaot.a(aot.cpp.o): undefined reference to symbol 'elf_version@@ELFUTILS_1.0'
66.34 /usr/bin/ld: /usr/lib/x86_64-linux-gnu/libelf.so: error adding symbols: DSO missing from command line
66.36 collect2: error: ld returned 1 exit status
```

danobi:
Not sure why it worked for local builds. But I think this change makes sense -- AOT started directly using libelf with the latest AOT change.

viktormalik:
Strange, could be caused by the order of linking.

Anyways, the fix is IMHO correct.

PR #3201, opened 2024-05-27
Add packages required for test build

dubeyabhishek:
Modified Ubuntu docker file for test builds.

To invoke the build, run the following command in top directory:
(with dot at end)
$ docker build -f docker/Dockerfile.ubuntu .

### DIFF
--- a/docker/Dockerfile.ubuntu
+++ b/docker/Dockerfile.ubuntu
@@ -1,10 +1,17 @@
 # This Dockerfile is used to both document and test building bpftrace on the
 # development version of ubuntu. We attempt to catch bugs as early as possible
 # which is why we are using devel.
-
+#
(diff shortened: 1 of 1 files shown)


dubeyabhishek:
@viktormalik How about having similar changes to other Docker files?

viktormalik (commented):
The purpose of these Dockerfiles is to have instructions on how to build bpftrace on various distros (what dependencies are needed). We're trying to keep these as small as possible to minimize the maintenance burden so I'm not in favor of building tests there.

In addition, containers are not a good development env for bpftrace, we prefer [Nix](https://github.com/bpftrace/bpftrace/blob/master/docs/nix.md).

dubeyabhishek:
got that! Thanks @viktormalik

Issue #2852, opened 2023-11-24
Bump to C++20


danobi:
The codebase does a lot of string processing. So far, we have a mix of:

* Inefficient string concat (eg. `std::string("hello") + " world" + " !"`) which is really inefficient cuz it constructs 2 temporaries
* `std::ostringstream` which is kinda difficult to visually parse
* `snprintf` and the usual C style drawbacks
* Probably more that I'm missing

C++20 contains `std::format` which neatly solves all the above problems. We should take a look at using that if we are ok bumping to C++20.

danobi:
According to https://en.cppreference.com/w/cpp/compiler_support/20, std::format has support starting at gcc 13 and clang 17. So it's pretty bleeding edge still.

danobi:
Alternatively, we could just depend on https://github.com/fmtlib/fmt and then codemod all the calls to `fmt::*` at some later point.

jordalgo:
Switching to just use `fmt` seems like simpler/easier solution to fixing string processing issues.

ajor:
`std::format` is probably too new to use, but C++ 20 brings a lot of other nice features that have been supported by GCC and Clang for years. C++ 23 is just about to be released, so I think it's time to upgrade to C++ 20 anyway.

viktormalik:
Another nice feature is `std::set<T>::contains` which can be used instead of using `.find(...)` and comparing the result with `.end()`

ajor:
Done in #3297

PR #3203, opened 2024-05-28
Refactor verbose and debug outputs

viktormalik:
Consolidate and cleanup information that is being printed in the verbose (`-v`) and debugging output (`-d`). The main idea is:
- the verbose output is intended for users to get more information on what bpftrace is doing and where it is (possibly) failing,
- the debug output is intended for developers to help them debugging bpftrace by providing very detailed outputs of individual stages of bpftrace runtime.

The main (user-facing) change this brings is that it introduces a mandatory argument to the `-d` option which allows to pick the stage whose output should be printed.

The currently supported stages are:
- `ast` - prints the AST after each pass,
- `codegen` - prints the LLVM IR code as emit by CodegenLLVM,
- `codegen-opt` - prints the LLVM IR code after it is optimized by LLVM (i.e. what is actually compiled to BPF bytecode),
- `libbpf` - captures and prints libbpf log for all libbpf operations that we use,
- `verifier` - captures and prints the BPF verifier log.

On top of that, `-d` can be used multiple times with different arguments and the argument `all` activates all of the above.

In addition, there are more user-facing changes:
- remove the `-dd` option,
- allow to use `-v` and `-d` simultaneously,
- add `--debug` as a long version of `-d`.

Also, some minor refactorings were done:
- make sure the verbosity output always goes to stderr while the debugging output always goes to stdout,
- improve some formatting of verbose outputs.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to
 ## Unreleased
 
 #### Added
+- Add --dry-run CLI option
+  - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
 #### Changed
 - Stream output when printing maps
   - [#3264](https://github.com/bpftrace/bpftrace/pull/3264)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,14 @@ and this project adheres to
   - [#3264](https://github.com/bpftrace/bpftrace/pull/3264)
 - Only print kernel headers not found message if parsing fails
   - [#3265](https://github.com/bpftrace/bpftrace/pull/3265)
+- Add mandatory "stage" argument to the -d CLI option
+  - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
+- Allow simultaneous use of -v and -d
+  - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
 #### Deprecated
 #### Removed
+- Remove the -dd CLI option
+  - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
 #### Fixed
 - Fix incorrect buf() size calculation which could lead to verifier errors
   - [#3281](https://github.com/bpftrace/bpftrace/pull/3281)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -75,6 +75,12 @@ The child process runs with the same privileges as bpftrace itself (usually root
 Enable debug mode.
 For more details see the <<Debug Output>> section.
 
+=== *--dry-run*
+
+Terminate execution right after attaching all the probes. Useful for testing
+that the script can be parsed, loaded, and attached, without actually running
+it.
+
 === *-e* _PROGRAM_
 
 Execute _PROGRAM_ instead of reading the program from a file or stdin.
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -70,14 +70,9 @@ To avoid a race condition when using 'USDTs', the child is stopped after 'execve
 The child PID is available to programs as the 'cpid' builtin.
 The child process runs with the same privileges as bpftrace itself (usually root).
 
-=== *-d*
+=== *-d STAGE*
 
 Enable debug mode.
-For more details see the <<Debug Output>> section.
-
-=== *-dd*
-
-Enable verbose debug mode.
 For more details see the <<Debug Output>> section.
 
 === *-e* _PROGRAM_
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -38,10 +38,16 @@ and this project adheres to
   - [#3095](https://github.com/bpftrace/bpftrace/pull/3095)
 - Move error printing from debug to verbose mode
   - [#3202](https://github.com/bpftrace/bpftrace/pull/3202)
+- Add mandatory "stage" argument to the -d CLI option
+  - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
+- Allow to use -v and -d simultaneously
```
jordalgo, line 43: nit
```suggestion
- Allow simultaneous use of -v and -d
```

src/main.cpp
```diff
@@ -113,8 +114,8 @@ void usage()
   std::cerr << std::endl;
   std::cerr << "TROUBLESHOOTING OPTIONS:" << std::endl;
   std::cerr << "    -v                      verbose messages" << std::endl;
-  std::cerr << "    -d                      (dry run) debug info" << std::endl;
-  std::cerr << "    -dd                     (dry run) verbose debug info" << std::endl;
+  std::cerr << "    -d STAGE                (dry run) debug info for various stages of bpftrace execution" << std::endl;
```
jordalgo, line 117: Could just `-d` default to all?

src/ast/pass_manager.cpp
```diff
@@ -27,7 +27,7 @@ void PassManager::AddPass(Pass p)
 PassResult PassManager::Run(std::unique_ptr<Node> node, PassContext &ctx)
 {
   Node *root = node.release();
-  if (bt_debug != DebugLevel::kNone)
+  if (bt_debug.find(DebugStage::Ast) != bt_debug.end())
```
jordalgo, line 30: can we use `contains` yet?

src/bpftrace.cpp
```diff
@@ -45,7 +45,7 @@
 
 namespace bpftrace {
 
-DebugLevel bt_debug = DebugLevel::kNone;
+std::set<DebugStage> bt_debug = {};
```
jordalgo, line 48: Maybe this should go into the config class? WDYT?

src/main.cpp
```diff
@@ -113,8 +114,8 @@ void usage()
...
+  std::cerr << "                            ('all', 'ast', 'codegen', 'codegen-opt', 'libbpf', 'verifier')" << std::endl;
```
jordalgo, line 118: All this should go in the adoc too

src/ast/pass_manager.cpp
viktormalik: We're still at C++17. I left a comment in #2852 to transform these into `contains` when we move to C++20.

src/main.cpp
viktormalik: Right, forgot about the docs. I updated it, thanks!

src/bpftrace.cpp
viktormalik: I'm not a big fan of that. It's not something that affects bpftrace behaviour, it's just enabling more prints. I'm not aware of any tool which would allow passing the verbosity/debug options via a config.

src/main.cpp
viktormalik: My original idea was the same but there are problems with that. First of all, it goes against [POSIX Utility Syntax Guidelines](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02) which says:

> Option-arguments should not be optional.

Also, if we made the argument to `-d` optional, I think that we'd have to always use the argument without a space, i.e. `-dast`, `-dcodegen`, etc. Otherwise, optarg wouldn't know if the following part of the CLI is an argument or a next option.

viktormalik:
Update: the `-d` option still works as a "dry run" but now we need to stop bpftrace only after probes are attached.

src/bpftrace.cpp
jordalgo: Yeah, I guess I wasn't thinking about it being set from inside the script itself, which I also don't like. It just seemed like some level of "configuration".

man/adoc/bpftrace.adoc
```diff
@@ -3465,52 +3460,38 @@ See src/attached_probe.cpp:find_vmlinux() for details.
...
+multiple times (with different stage names) and the special value `all` prints
+the output of all the supported stages.
+
+This option does a dry run, i.e. after the information is printed, bpftrace is
+terminated.
 
 Note: This is primarily used for bpftrace developers.
 
-The output begins with `Program` and then an abstract syntax tree (AST) representation of the program.
+The supported stages are:
```
jordalgo, line 3473: Nit: I might change the table header to 'The supported options are' so you can also include 'all'.

jordalgo (approved):
LGTM.

The one additional thing we could consider doing is moving all the the stage level checking into `log.cpp` and changing `LOG(DEBUG)` to something like `LOG(DEBUG, stage)`. At the moment `LOG(DEBUG)` is only used in 1 place so I think it's fine to modify it if we want.

man/adoc/bpftrace.adoc
viktormalik: Yeah, good idea, having everything in the table is more comprehensive.

viktormalik:
> The one additional thing we could consider doing is moving all the the stage level checking into `log.cpp` and changing `LOG(DEBUG)` to something like `LOG(DEBUG, stage)`. At the moment `LOG(DEBUG)` is only used in 1 place so I think it's fine to modify it if we want.

I like this. Let's do it in a follow-up PR, though.

jordalgo:
So the `-d` flag now implicitly adds the `--dry-run` flag? Also should this new flag be mentioned in the changelog?

viktormalik:
> So the `-d` flag now implicitly adds the `--dry-run` flag? Also should this new flag be mentioned in the changelog?

Yes, it does. I added it into the changelog.

viktormalik:
> Update: the `-d` option still works as a "dry run" but now we need to stop bpftrace only after probes are attached.

Ok, this caused more trouble than I originally anticipated. The main reason is `tools-parsing-test.sh`. Originally, the test would use `-d` to do a dry run of each tool. To preserve the same behaviour, I added a new `--dry-run` CLI option and I'm using it along with verbose output `-v` in `tools-parsing-test.sh`. An advantage of this is that now we even try to load and attach the tools instead of just generating the bytecode.

The reason why I'm not using `-d all` in `tools-parsing-test.sh` is twofold:
- It is very verbose. When a tool test fails, it's impossible to navigate in the hundreds of lines of output in the GHA environment. We could potentially print just some stage but I wouldn't be sure which one. Dry run with a verbose output (which will, among others, show verifier log in case of failure) seems like a reasonable option.
- Some tools fail to load with `-d verifier`. This is strange as the only thing this does is that it sets `bpf_prog_load_opts.log_level` to 15. This may be a kernel bug (or feature, as a matter of fact) so I'll need to investigate more. Since it's just occurring in a debug run, I don't think that this is a blocker for merging this PR.

src/main.cpp
```diff
@@ -113,8 +115,9 @@ void usage()
   std::cerr << std::endl;
   std::cerr << "TROUBLESHOOTING OPTIONS:" << std::endl;
   std::cerr << "    -v                      verbose messages" << std::endl;
-  std::cerr << "    -d                      (dry run) debug info" << std::endl;
-  std::cerr << "    -dd                     (dry run) verbose debug info" << std::endl;
+  std::cerr << "    --dry-run               terminate execution right after attaching all the probes" << std::endl;
+  std::cerr << "    -d STAGE                (dry run) debug info for various stages of bpftrace execution" << std::endl;
```
jordalgo, line 119: nit: maybe add the bit about how --dry-run is automatically added at the end of the text as it reads now it feels like -d means dry-run.

viktormalik: Yeah, good idea, fixed that.

src/main.cpp
```diff
@@ -113,9 +115,10 @@ void usage()
   std::cerr << std::endl;
   std::cerr << "TROUBLESHOOTING OPTIONS:" << std::endl;
   std::cerr << "    -v                      verbose messages" << std::endl;
-  std::cerr << "    -d                      (dry run) debug info" << std::endl;
-  std::cerr << "    -dd                     (dry run) verbose debug info" << std::endl;
-  std::cerr << "    --emit-elf FILE         (dry run) generate ELF file with bpf programs and write to FILE" << std::endl;
+  std::cerr << "    --dry-run               terminate execution right after attaching all the probes" << std::endl;
+  std::cerr << "    -d STAGE                debug info for various stages of bpftrace execution (dry dun)" << std::endl;
```
jordalgo, line 119:
```suggestion
  std::cerr << "    -d STAGE                debug info for various stages of bpftrace execution (dry run)" << std::endl;
```

viktormalik: damn :grinning:

jordalgo (approved):

danobi (commented):
I like this a lot.

My only feedback is that I think it'd be nice to not make `--dry-run` and `-d` duplicate the action of early exit. I suggest to make `-d` only control debug _output_ and not affect execution. We are already departing from having `-d` emit output and not attach, so I think we are free to change it further.

That way `-d` and `--dry-run` remain orthogonal and thus more composable.

viktormalik:
> I like this a lot.
> 
> My only feedback is that I think it'd be nice to not make `--dry-run` and `-d` duplicate the action of early exit. I suggest to make `-d` only control debug _output_ and not affect execution. We are already departing from having `-d` emit output and not attach, so I think we are free to change it further.
> 
> That way `-d` and `--dry-run` remain orthogonal and thus more composable.

That's a good idea. I changed `-d` not to terminate execution.

danobi (approved):
lgtm! but CI is failing now

viktormalik:
> lgtm! but CI is failing now

That's because of #3235. I'd love to hear your opinion on that one.

viktormalik:
This needs #3246 to be merged first to fix the tools check.

viktormalik:
Rebased on top of master. In case no one has objections, I'll merge this today.

viktormalik:
> Rebased on top of master. In case no one has objections, I'll merge this today.

Or not, unfortunately `undump.bt` is broken, see #3280.

(We really need this merged ASAP to test tools loading and attachment in the CI)

danobi:
I'm looking at the regression. If you wanna `TOOLS_TEST_DISABLE` the broken one for now, I can remove it after I fix the regression

Issue #3180, opened 2024-05-17
Incorrect number of probes being attached reported with kprobe globbing


tyroguru:
### What reproduces the bug? Provide code if possible.

When using globbing to match multiple kprobes we use the multi kprobe link type to attach the probes. When doing this we report the number of probes being attached to as 1 instead of the actual number of probes being attached.

```
$ cat ~/BPF/kernelfuncs.bt
config = {
  max_probes = 50000;
}

kprobe:s*
{
  @[func] = count();
}

$ sudo ./bpftrace  ~/BPF/kernelfuncs.bt
Attaching 1 probe...
```

Obviously we can calculate up front how many probes we are trying to attach to though:

```
$ sudo ./bpftrace  -l ~/BPF/kernelfuncs.bt | wc -l
3780
```

I haven't tested but this problem probably applies to uprobe multi attach as well.

viktormalik:
Technically speaking, there's only one probe being attached (from bpftrace point of view) with the list of functions passed via `link_opts`. OTOH, I see that this may be confusing.

At the moment, we print the amount of attached functions, along with the complete list of functions, when `-v` is used:
```
# bpftrace -v -e 'kprobe:s* { @[func] = count(); }'
node count: 7
Attaching 1 probe...
[...]
Attaching kprobe:s*
Attaching to 7867 functions
 s25fs256t_late_init
 s25fs256t_post_bfpt_fixup
 [...]
```

I'd suggest we change the behaviour to report
```
Attaching 1 probe to 7876 functions...
```
for the non-verbose output and add the list of functions for the verbose output.

I'm currently working on refactoring the verbosity/debug outputs so I can add this, too.

tyroguru:
I think it's important here not to introduce anything that could be confusing to an end user as it's confusing enough as it is for them. IMO there should be no difference in the output given to a user whether they specified the probes manually or whether they obtained the list of probes to enable via globbing.  Exposing the details of a 1:N mapping that kprobe multi gives could be very confusing, especially if the output is different to the non-globbed way of specifying the probes.

viktormalik:
That's a fair point. We probably shouldn't print a different number for the same bpftrace script depending on whether the kprobe multi is used or not.

jordalgo:
This is now fixed.

PR #3204, opened 2024-05-28
Filter out kernel functions that can cause systemic instability

tyroguru:
When using globbing with kprobes it easy extremely easy to hang the system and this is because certain kernel functions react badly to being instrumented. For example, on a 5.19 based kernel, the following invocation will hang the system:

```
sudo bpftrace -e 'kprobe:*' -e 'interval:s:5{exit();}'
```

The bpf kernel selftests specify some functions that must be explicitly filtered and also some patterns that are used to match groups of functions that must be filtered. This change uses these patterns to exclude functions when the available function list is being built.
##### Checklist

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -1093,6 +1093,33 @@ std::string hex_format_buffer(const char *buf,
   return std::string(s);
 }
 
+static bool is_bad_func(std::string &func)
+{
+  /*
+   * Certain kernel functions are known to cause system stability issues if
+   * traced (but not marked "notrace" in the kernel) so they should be filtered
+   * out as the list is built. The list of functions have been taken from the
+   * bpf kernel selftests (bpf/prog_tests/kprobe_multi_test.c).
+   */
+  static const std::unordered_set<std::string> bad_funcs = {
+    "arch_cpu_idle", "default_idle", "bpf_dispatcher_xdp_func"
+  };
+
+  static const std::vector<std::string> bad_funcs_partial = {
+    "__ftrace_invalid_address__", "rcu_"
+  };
+
+  if (bad_funcs.find(func) != bad_funcs.end())
+    return true;
+
+  for (const auto &s : bad_funcs_partial) {
+    if (!std::strncmp(func.c_str(), s.c_str(), s.length()))
+      return true;
+  }
+
+  return false;
+}
+
 FuncsModulesMap parse_traceable_funcs()
 {
 #ifdef FUZZ
@@ -1117,7 +1144,9 @@ FuncsModulesMap parse_traceable_funcs()
     auto func_mod = split_symbol_module(line);
     if (func_mod.second.empty())
       func_mod.second = "vmlinux";
-    result[func_mod.first].insert(func_mod.second);
+
+    if (!is_bad_func(func_mod.first))
+      result[func_mod.first].insert(func_mod.second);
   }
 
   // Filter out functions from the kprobe blacklist.
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -141,6 +141,24 @@ EXPECT ERROR: Error attaching probe: kprobe:nonsense:vfs_read: specified module 
 TIMEOUT 5
 WILL_FAIL
 
+NAME kprobe_wildcard_all_leading_underscore
+PROG kprobe:_* { } interval:s:1{ printf("SUCCESS"); exit(); }
+EXPECT_REGEX SUCCESS
+ENV BPFTRACE_MAX_PROBES=20000
+TIMEOUT 5
+REQUIRES_FEATURE kprobe_multi
+
+NAME kprobe_disallow_rcu_functions
+PROG kprobe:rcu_* { printf("SUCCESS %d\n", pid); exit(); }
+EXPECT No probes to attach
+TIMEOUT 5
+WILL_FAIL
+
+NAME kprobe_hide_ftrace_invalid_address
+RUN {{BPFTRACE}} -l | grep -c __ftrace_invalid_address__ || true
+EXPECT 0
+TIMEOUT 5
+
 NAME kprobe_func_missing
 PROG kprobe:vmlinux:nonsense { printf("SUCCESS %d\n", pid); exit(); }
 EXPECT ERROR: Error attaching probe: kprobe:vmlinux:nonsense


src/utils.cpp
```diff
@@ -1104,6 +1104,20 @@ FuncsModulesMap parse_traceable_funcs()
...
+   * as the list is built. The list of functions have been taken from
+   * the bpf kernel selftests (bpf/prog_tests/kprobe_multi_test.c).
+   */
+  const std::unordered_set<std::string> bad_funcs = {
+    "arch_cpu_idle", "default_idle", "bpf_dispatcher_xdp_func"
+  };
+
+  const std::vector<std::string> bad_funcs_partial = {
+    "__ftrace_invalid_address__", "rcu_"
+  };
```
danobi, line 1119: Can we move these closer to https://github.com/bpftrace/bpftrace/blob/c532239f4b42eda260fcc8aa9ece56fecc1d083e/src/attached_probe.cpp#L36-L45 ? Would be good to keep all these together in the same location. Doesn't necessarily have to be in attached_probe.cpp

danobi (commented):
I'm not 100% sure, but isn't it better to annotate these in kernel? `notrace` appears to do this

tyroguru:
> I'm not 100% sure, but isn't it better to annotate these in kernel? `notrace` appears to do this

There are reasons why some of these functions are not annotated as such. For example, the discussion at https://lore.kernel.org/lkml/20220528025249.577189376@goodmis.org/T/ says why the "__ftrace_invalid_address__" functions are still effectively listed as available functions and not even in the blacklist! 

The main problem with marking functions as "notrace" @danobi is that it will take forever before kernels are available for use in most people's production environments (multiple years) even if those changes were made now. Much, much safer just to cover them off here and if someone is feeling bored they can go and "notrace" them in the kernel  as they see fit.

src/utils.cpp
tyroguru: > Can we move these closer to
> 
> https://github.com/bpftrace/bpftrace/blob/c532239f4b42eda260fcc8aa9ece56fecc1d083e/src/attached_probe.cpp#L36-L45
> 
> ? Would be good to keep all these together in the same location. Doesn't necessarily have to be in attached_probe.cpp

We can have a chat tomorrow but that mechanism doesn't seem to be a good one to me. Why not just remove the functions from the ones that are available instead of having this mechanism which is more intrusive and provides no extra benefit.

tyroguru: > > Can we move these closer to
> > https://github.com/bpftrace/bpftrace/blob/c532239f4b42eda260fcc8aa9ece56fecc1d083e/src/attached_probe.cpp#L36-L45
> > 
> > ? Would be good to keep all these together in the same location. Doesn't necessarily have to be in attached_probe.cpp
> 
> We can have a chat tomorrow but that mechanism doesn't seem to be a good one to me. Why not just remove the functions from the ones that are available instead of having this mechanism which is more intrusive and provides no extra benefit.

Just a follow on: the `banned_kretprobes` mechanism is most likely being removed in #3206 .

src/utils.cpp
```diff
@@ -1104,6 +1104,20 @@ FuncsModulesMap parse_traceable_funcs()
                                       ? path_env
                                       : tracefs::available_filter_functions();
 
+  /*
+   * Certain kernel functions are known to cause system stability issues if
+   * traced yet they are not specified in the blacklist so are filtered out
+   * as the list is built. The list of functions have been taken from
```
jordalgo, line 1110:
```suggestion
   * Certain kernel functions are known to cause system stability issues if
   * traced (but not marked "notrace" in the kernel) so they should be filtered out
   * as the list is built. The list of functions have been taken from
```

jordalgo: We could also mention that as of kernel version 6.9 these functions haven't been marked "notrace" but in the future we could potentially remove them from our internal blocklist if that changes.

src/utils.cpp
jordalgo: There is still going to be a list of functions but I can move those into utils.cpp to keep them with these.

tyroguru: > There is still going to be a list of functions but I can move those into utils.cpp to keep them with these.

Up to you @jordalgo but probably makes sense to keep them together for ease of maintenance.

jordalgo (approved):

tests/runtime/probe
```diff
@@ -141,6 +141,25 @@ EXPECT ERROR: Error attaching probe: kprobe:nonsense:vfs_read: specified module
 TIMEOUT 5
 WILL_FAIL
 
+NAME kprobe_wildcard_all_leading_underscore
```
jordalgo, line 144: This test seems to be causing an issue in CI. I think we can probably get rid of this one as it probably just takes to long to attach. The ones below are great though - thanks for adding.

tyroguru: > This test seems to be causing an issue in CI. I think we can probably get rid of this one as it probably just takes too long to attach. The ones below are great though - thanks for adding.

It looks like it's failing because it's exceeding the default maximum number of probes (512). This hasn't been an issue in testing as the kernels i work on have kprobe multi enabled and we end up reporting a small number of probes even if thousands are enabled. Victor was looking at this recently so maybe that has been fixed. I'll sort something out (either drop it or fix it).

tyroguru: @jordalgo I just upped the max probe limit for that one test to 20k (totally random value really that is ~3x the current number of probes).

jordalgo: I pulled down this PR locally and when I run this prog on my dev machine I get this error "ERROR: Error attaching probe: kprobe:_*". Looks like the errno is ESRCH - so not sure what's going on there.

Anyway, this looks like it passes CI just fine but I'll wait another day to see if anyone has any objections before I merge.

viktormalik: Beware that attaching to too many kprobes may cause kernel hang or panic (that's why the artificial limit is there). Also that's one of the reasons why kprobe_multi was introduced, despite the name, it uses a different attachment mechanism (fprobe). I'd avoid running this test when kprobe_multi is not available on the system.

src/utils.cpp
viktormalik: `utils.cpp` is already long enough. Maybe it'd be worth having a separate source file for these, to simplify maintenance.

src/utils.cpp
```diff
@@ -1114,10 +1128,20 @@ FuncsModulesMap parse_traceable_funcs()
...
       func_mod.second = "vmlinux";
-    result[func_mod.first].insert(func_mod.second);
+    if (bad_funcs.find(func_mod.first) != bad_funcs.end())
+      continue;
+    for (const auto &s : bad_funcs_partial) {
+      if (!std::strncmp(func_mod.first.c_str(), s.c_str(), s.length())) {
+        skip = true;
+        break;
+      }
+    }
```
viktormalik, line 1142: Would be nice to enclose this check in a function (e.g. `is_bad_func`).

tests/runtime/probe
```diff
@@ -141,6 +141,26 @@ EXPECT ERROR: Error attaching probe: kprobe:nonsense:vfs_read: specified module
...
+NAME kprobe_disallow_rcu_functions
+PROG kprobe:rcu_* { printf("SUCCESS %d\n", pid); exit(); }
+EXPECT No probes to attach
+TIMEOUT 5
+WILL_FAIL
+
+# WILL_FAIL is required here because 'grep' returns an error code of 1 when
+# no lines are matched.
+NAME kprobe_hide_ftrace_invalid_address
+RUN {{BPFTRACE}} -l | grep -c __ftrace_invalid_address__
```
viktormalik, line 159: Nit: it seems a bit more clear to me to use
```suggestion
RUN {{BPFTRACE}} -l | grep -c __ftrace_invalid_address__ || true
```
rather than `WILL_FAIL` but that may be just my preference.

tests/runtime/probe
jordalgo: My vote is to just remove this test but if you want to keep then you can use this to require kprobe multi `REQUIRES_FEATURE kprobe_multi`

tyroguru: > Beware that attaching to too many kprobes may cause kernel hang or panic (that's why the artificial limit is there). Also that's one of the reasons why kprobe_multi was introduced, despite the name, it uses a different attachment mechanism (fprobe). I'd avoid running this test when kprobe_multi is not available on the system.

@viktormalik Good point on the kprobe_multi feature requirement - I'll add that to that test. The point of this change (and possible subsequent changes) is to make it so that a user can light up a lot of probes via globbing without having to worry about stability issues - they simply shouldn't have to especially when dealing with kernel probes entry/return probes. There may have to be a number of iterations at this but the current limitation (512 probes) is pretty ridiculous and I think fairly arbitrary from what I remember. Of course, keep a limit in for now but removing functions with known stability issues is a start in that direction.

tyroguru: > I pulled down this PR locally and when I run this prog on my dev machine I get this error "ERROR: Error attaching probe: kprobe:_*". Looks like the errno is ESRCH - so not sure what's going on there.
> 
> Anyway, this looks like it passes CI just fine but I'll wait another day to see if anyone has any objections before I merge.

@jordalgo I'm seeing some seriously annoying behaviour on 6.4 based kernel that I have yet to get to the bottom of. I'll try to look at this again some time this week.

viktormalik: > > Beware that attaching to too many kprobes may cause kernel hang or panic (that's why the artificial limit is there). Also that's one of the reasons why kprobe_multi was introduced, despite the name, it uses a different attachment mechanism (fprobe). I'd avoid running this test when kprobe_multi is not available on the system.
> 
> @viktormalik Good point on the kprobe_multi feature requirement - I'll add that to that test. The point of this change (and possible subsequent changes) is to make it so that a user can light up a lot of probes via globbing without having to worry about stability issues - they simply shouldn't have to especially when dealing with kernel probes entry/return probes. There may have to be a number of iterations at this but the current limitation (512 probes) is pretty ridiculous and I think fairly arbitrary from what I remember. Of course, keep a limit in for now but removing functions with known stability issues is a start in that direction.

Yes, absolutely, it's a good step towards that direction. OTOH, I believe that the limit doesn't (and shouldn't) apply to kprobe_multi so if we change bpftrace to report the actual number of attached functions (as we discussed in #3180), we'll have to make sure that the limit is not applied to kprobe_multi. AFAIK, the limit is only for kprobes (and potentially uprobes) where it still applies to this day (although it's been a while since I tried to crash my kernel with so many kprobes).

src/utils.cpp
jordalgo: I agree but this is not a deal breaker for me and having utility functions only called from one place also has a bit of smell (though it would make it easier to read).

tyroguru: > Would be nice to enclose this check in a function (e.g. `is_bad_func`).

You're right @viktormalik (and @jordalgo :-) ) - it does look better with the code hoisted out into a separate function.

tests/runtime/probe
tyroguru: > Nit: it seems a bit more clear to me to use
> 
> rather than `WILL_FAIL` but that may be just my preference.

Removed the WILL_FAIL and took your suggestion - thanks.

src/utils.cpp
```diff
@@ -1093,6 +1093,33 @@ std::string hex_format_buffer(const char *buf,
...
 
+static bool is_bad_func(std::string &func)
+{
+  /*
+   * Certain kernel functions are known to cause system stability issues if
+   * traced (but not marked "notrace" in the kernel) so they should be filtered
+   * out as the list is built. The list of functions have been taken from the
+   * bpf kernel selftests (bpf/prog_tests/kprobe_multi_test.c).
+   */
+  const std::unordered_set<std::string> bad_funcs = {
```
jordalgo, line 1104: sorry more nits! 
```suggestion
  static const std::unordered_set<std::string> bad_funcs = {
```

src/utils.cpp
```diff
@@ -1093,6 +1093,33 @@ std::string hex_format_buffer(const char *buf,
...
+    "arch_cpu_idle", "default_idle", "bpf_dispatcher_xdp_func"
+  };
+
+  const std::vector<std::string> bad_funcs_partial = {
```
jordalgo, line 1108:
```suggestion
  static const std::vector<std::string> bad_funcs_partial = {
```

jordalgo: This also feels like it should be moved above the bad_funcs_partial loop

jordalgo (approved):

PR #3206, opened 2024-05-29
Prevent recursion crash for kfunc/kretfunc

jordalgo:
When attaching to certain functions via kfunc/fentry and
kretfunc/fexit there is a potential for recursion that
leads to an ABBA deadlock and subsequent kernel crash.

This change adds a PER_CPU_MAP when we detect one of these
dangerous functions and upon entry to the BPF prog attaching
to this function will set the value in this map to 1 if
the value is 0 otherwise the function will exit early. Then
upon BPF prog exit it will set the value back to 0. This
will prevent the aforementioned recursive behavior.

Simple repro:
```
kfunc:queued_spin_lock_slowpath
{
  @b[pid] = 2;
  @A[tid] = 2;
}

kretfunc:queued_spin_lock_slowpath
{
  @b[pid] = 3;
  @A[tid] = 3;
}
```

Note: there is already a mechansim in the kernel to prevent
this from happening for kprobe/kretprobe.

##### Checklist


src/ast/irbuilderbpf.cpp
```diff
@@ -543,6 +543,64 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+   * This should only happend for the functions that can lead
+   * to a recursive crash e.g. "queued_spin_lock_slowpath"
+   */
+  CreateRet(getInt64(0));
+
+  SetInsertPoint(value_is_unset_block);
+
+  AllocaInst *set_value = CreateAllocaBPF(getInt64Ty(), "set_value");
+  CreateStore(getInt64(1), set_value);
+  CreateMapUpdateElem(ctx, map_ident, key, set_value, loc);
```
4ast, line 583: Pls remove. Always update map value in place. No need for this overhead.
Also since it's filtering for kfuncs attached to spin_locks there is no guarantee that irqs are disabled,
so it's better to do __sync_fetch_and_add() instead of vanilla increment.
Only if kernel is really old fall back to ldx, +1, stx.

4ast (changes requested):

src/ast/irbuilderbpf.cpp
jordalgo: Good catch. I did this for the other per-cpu maps in bpftrace but forgot to do it here lol.

> Also since it's filtering for kfuncs attached to spin_locks there is no guarantee that irqs are disabled,
so it's better to do __sync_fetch_and_add() instead of vanilla increment.

What if we are just less strict about the comparison and just check if it's not 0, then even if there is a race condition and it ends up being 2 instead of 1 then it still counts as set.

4ast: Not doing atomic means that a cpu will read zero, assume its fine, then irq, then the same prog reads zero again and both will execute.
That's why in the kernel we do:

jordalgo: Instead of doing an addition, I'm just doing an exchange because we just care if the number is 1 or 0 and I think this should work regardless of kernel version (hopefully).
```
CREATE_ATOMIC_RMW(AtomicRMWInst::BinOp::Xchg,
                    cast,
                    getInt64(1),
                    8,
                    AtomicOrdering::SequentiallyConsistent);
```

src/ast/irbuilderbpf.cpp
```diff
@@ -543,6 +543,90 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+void IRBuilderBPF::CreateUnSetRecursion(Value *ctx, const location &loc)
+{
+  const std::string map_ident = to_string(MapType::RecursionPrevention);
+  AllocaInst *key = CreateAllocaBPF(getInt32Ty(), "lookup_key");
+  CreateStore(getInt32(0), key);
+
+  AllocaInst *unset_value = CreateAllocaBPF(getInt64Ty(), "unset_value");
+  CreateStore(getInt64(0), unset_value);
+
+  CreateMapUpdateElem(ctx, map_ident, key, unset_value, loc);
```
4ast, line 624: Pls do not use update. It's practically never necessary. Always update in place after lookup. Just map_lookup_elem followed by unconditional ST val, 0

src/ast/irbuilderbpf.cpp
```diff
@@ -543,6 +543,90 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+                                        getInt64(1),
+                                        8,
+                                        AtomicOrdering::SequentiallyConsistent);
+
+  Function *set_parent = GetInsertBlock()->getParent();
+  BasicBlock *value_is_set_block = BasicBlock::Create(module_.getContext(),
+                                                      "value_is_set",
+                                                      set_parent);
+  Value *set_condition = CreateICmpEQ(prev_value,
+                                      getInt64(1),
```
4ast, line 592: It feels comparing with 0 is more natural:
recur_enter:
prev = xchg(ptr, 1);
if (prev == 0) // good case, proceed


recur_exit:
*ptr = 0

4ast (changes requested):

src/ast/irbuilderbpf.cpp
```diff
@@ -543,6 +543,90 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+                                      "value_set_condition");
+  CreateCondBr(set_condition, value_is_set_block, merge_block);
+
+  SetInsertPoint(value_is_set_block);
+  /*
+   * The counter is set, we need to exit early from the probe.
+   * This should only happen for the functions that can lead
+   * to a recursive crash e.g. "queued_spin_lock_slowpath"
+   */
+  CreateRet(getInt64(0));
```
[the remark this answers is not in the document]
jordalgo: But this is a case of recursion right? Does that really qualify as a dropped event?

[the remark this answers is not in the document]
jordalgo: I'm also not keen on adding more logic in an escape path where we're trying to prevent doing work. Maybe I'm being overly paranoid.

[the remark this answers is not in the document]
danobi: At a high level (I and many others) believe that it should be a functional requirement that we don't ever lose an event. We may not always be able to execute code, but we need to at least report it.

Particularly in weird spots like these, it's critical we report back to user a corner case happened. Corner cases are typically why people use tracing over profiling.

Practically speaking, a percpu array access is rewritten to a direct memory access (no helpers involved). And we can do an atomic increment on the destination address. So it is a failure free codepath. If it is not a failure free codepath, the kernel must change to support this.

src/ast/irbuilderbpf.cpp
jordalgo: I'm not sure I agree (as I like seeing early exits) but I also don't mind changing it.

src/ast/irbuilderbpf.cpp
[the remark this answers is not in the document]
jordalgo: Discussed offline. I'm good to add this accounting for lost events but I think it might need to be a separate PR as I'll need to make a more general version of `handle_ringbuf_loss`.

src/ast/irbuilderbpf.cpp
```diff
@@ -701,6 +701,125 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+  CreateCondBr(condition, lookup_success_block, lookup_failure_block);
+
+  SetInsertPoint(lookup_success_block);
+
+  CreateLifetimeEnd(key);
+
+  // createMapLookup  returns an u8*
+  auto *cast = CreatePointerCast(call, getInt64Ty(), "cast");
+
+  Value *prev_value = CREATE_ATOMIC_RMW(AtomicRMWInst::BinOp::Xchg,
```
4ast, line 739: xchg logic lgtm.

4ast (approved):

src/ast/irbuilderbpf.cpp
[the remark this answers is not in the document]
jordalgo: Ok, making the ring buff loss counter generic was pretty easy. @danobi Let me know if you want me to stack the commit on top of this one in this PR. Example ouput:
```
[jordalgo@devbig1151.frc2 ~/local/bpftrace/build (kfunc-counter-3144)]$ sudo ./src/bpftrace ~/abba-crash-repro.bt
Attaching 2 probes...
Lost 13 events
Lost 28 events
Lost 7 events
Lost 8 events
Lost 322 events
Lost 14 events
Lost 1 events
Lost 31 events
...
```

src/ast/irbuilderbpf.cpp
```diff
@@ -701,6 +701,125 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+   */
+  CreateRet(getInt64(0));
+
+  SetInsertPoint(lookup_failure_block);
+
+  CreateDebugOutput(
+      "Value for per-cpu map key 0 is null. This shouldn't happen.",
+      std::vector<Value *>{},
+      loc);
+  CreateRet(getInt64(0));
```
danobi, line 768: Hmm, we cannot always return 0 -- see CodegenLLVM::createRet(). But if this is only used for kfuncs, then it's probably ok. Maybe leave a comment calling that out?

src/ast/irbuilderbpf.cpp
```diff
@@ -701,6 +701,125 @@ void IRBuilderBPF::CreateForEachMapElem(Value *ctx,
...
+      "map_lookup_cond");
+  CreateCondBr(condition, lookup_success_block, lookup_failure_block);
+
+  SetInsertPoint(lookup_success_block);
+
+  CreateLifetimeEnd(key);
+
+  // createMapLookup  returns an u8*
+  auto *cast = CreatePointerCast(call, getInt64Ty(), "cast");
+  CreateStore(getInt64(0), cast);
```
danobi, line 807: Does this need to be an atomic store?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2512,6 +2512,9 @@ void CodegenLLVM::visit(Predicate &pred)
 
   b_.CreateCondBr(expr_, pred_false_block, pred_true_block);
   b_.SetInsertPoint(pred_false_block);
+  if (need_recursion_check_) {
+    b_.CreateUnSetRecursion(current_attach_point_->loc);
+  }
```
danobi, line 2517: Let's push this down into createRet(). I have another PR that will make a call to createRet(). If someone later adds a createRet() but forgets to unset the recursion guard, it'll be disastrous.

src/ast/irbuilderbpf.cpp
jordalgo: I asked @4ast the same question above. His response: "The final ST val, 0 is fine, since on entry we got exclusive access. ST val, 0 is WRITE_ONCE equivalent. JITs must not tear the store."

danobi: Ha, shoulda read closer. Interesting, my naive understanding is that you cannot depend on that for other archs (other than x86-64). :shrug:

src/ast/passes/codegen_llvm.cpp
jordalgo: Ha.  I didn't even know `createRet` existed. I'm definitely putting up another PR to better distinguish this internal function from `CreateRet` from LLVM.

src/bpftrace.cpp
```diff
@@ -2344,5 +2344,24 @@ int BPFtrace::get_num_possible_cpus() const
 {
   return libbpf_num_possible_cpus();
 }
+// Note: it would be better if this was in resource analyzer but we need
+// probe_matcher to get the list of functions for the attach point
+void BPFtrace::kfunc_recursion_check(ast::Program *prog)
```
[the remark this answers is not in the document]
jordalgo: Good call. Thanks for writing something up, I'll add it. Though I think the NMI part is not what we're protecting against here.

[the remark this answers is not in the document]
jordalgo: So here is an explanation for this example (@4ast keep me honest here).

You have this bpftrace script:
```
kfunc:queued_spin_lock_slowpath
{
  @a[tid] = 2;
  @b[pid] = 2;
}

kretfunc:queued_spin_lock_slowpath
{
  @b[pid] = 3;
  @a[tid] = 3;
}
```

On CPU 1, kretfunc fires first and takes a lock on hash map b and in doing so triggers kfunc, which tries to take a lock for hash map a.

On CPU 2, kfunc fires first and takes a lock on hash map a and in doing so triggers kretfunc, which tries to take a lock on hash map b. 

Here is the AB - BA deadlock.

src/utils.h
```diff
@@ -157,6 +158,13 @@ static std::vector<std::string> COMPILE_TIME_FUNCS = { "cgroupid" };
 
 static std::vector<std::string> UPROBE_LANGS = { "cpp" };
 
+static const std::set<std::string> RECURSIVE_KERNEL_FUNCS = {
+  "vmlinux:_raw_spin_lock",
+  "vmlinux:_raw_spin_lock_irqsave",
+  "vmlinux:_raw_spin_unlock_irqrestore",
+  "vmlinux:queued_spin_lock_slowpath",
+};
```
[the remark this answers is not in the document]
jordalgo: IIUC - it's only kfuncs that attach to spin lock internals.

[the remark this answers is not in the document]
danobi: @4ast can you confirm if this is sufficient?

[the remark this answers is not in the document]
4ast: it's not 100%, but good enough.

src/ast/irbuilderbpf.cpp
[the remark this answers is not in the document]
danobi: Follow up is fine!

jordalgo:
Ok this is ready for another look @4ast @danobi

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2693,8 +2712,10 @@ void CodegenLLVM::createRet(Value *value)
     case ProbeType::kretfunc:
     case ProbeType::iter:
     case ProbeType::rawtracepoint:
-      b_.CreateRet(b_.getInt64(0));
-      break;
+      return 0;
+    default:
+      LOG(BUG) << "Unknown probetype";
+      return 0;
```
danobi, line 2718: Why add default? If we leave it out and cases are exhaustive, then we will get compiler warning if new probe types are not handled. Adding default silences the warning and turns a compile time warning into a run time failure

danobi (commented):
assuming you did not change irbuilderbpf.cpp bits, lgtm modulo comment about compiler warnings

src/ast/passes/codegen_llvm.cpp
jordalgo: Ha. I was avoiding another compiler warning about returning void in a non void function. I'll fix

jordalgo:
> assuming you did not change irbuilderbpf.cpp bits

Yeah nothing changed there but in codegen we now add the recursion check for all probes if we detect a dangerous function is being attached to via kfunc/kretfunc

src/ast/passes/codegen_llvm.cpp
jordalgo: I basically moved the `LOG(BUG)` out of the switch so now we have both the compiler warning and the runtime failure.

danobi (approved):

PR #3211, opened 2024-05-31
aot: Fix return code to enable aot runtime tests

arthurshau:
Currently, the ```build_binary``` function in ```aot.cpp``` returns true (1) in the successful path, and false (0) on any errors. However, it should be the other way around, so let's fix that. This enables the ```runtime-tests.sh --run-aot-tests``` flag (and any other command that generates an aot binary on the left side of a ```&&```) to work properly. While I'm at it, let's also fix the runtime test command to not need the runtime shim anymore, as the generated binaries can now run on their own.

When ran on my machine, 223 tests pass out of the 389 in the aot suite, so roughly under 60%. A lot of these are because of features that are not yet supported in AOT, which are caught by ```portability_analyser.cpp``` .


src/aot/aot.cpp
```diff
@@ -128,14 +128,14 @@ std::optional<std::vector<uint8_t>> generate_btaot_section(
...
-                  const std::vector<uint8_t> &section)
+int build_binary(const std_filesystem::path &shim,
+                 const std::string &out,
+                 const std::vector<uint8_t> &section)
 {
   std::error_code ec;
-  bool ret = false;
   char cmd[1024];
   int written;
+  int r = 0;
```
jordalgo, line 138: nit: this variable is better named 'ret' as it was IMHO.

jordalgo:
Sorry, I'm confused, why don't we need the shim for these tests? I see the path to the shim was just modified here: https://github.com/bpftrace/bpftrace/pull/3199/files

danobi:
The runtime tests no longer need to directly manipulate the shim b/c AOT generation codepath can locate the shim on its own now

arthurshau:
Yep, also with the change from #3104, it's no longer necessary to call the runtime shim, aka ```bpftrace-aotrt```, to run the generated binary. So before, after you generate with e.g. ```bpftrace -e 'BEGIN { exit(); }' --aot a.btaot```,  you would have to do ```bpftrace-aotrt a.btaot```. But now you can just run ```./a.btaot``` directly. So this change makes the runtime tests follow that pattern by omitting```bpftrace-aotrt``` as well.

tests/runtime/engine/runner.py
```diff
@@ -103,7 +102,6 @@ def prepare_bpf_call(test, nsenter=[]):
 
         if test.run:
             ret = re.sub("{{BPFTRACE}}", BPFTRACE_BIN, test.run)
-            ret = re.sub("{{BPFTRACE_AOTRT}}", AOT_BIN, ret)
```
danobi, line 106: I think tests/README.md also needs to be updated

danobi (approved):

PR #3219, opened 2024-06-04
tests: utils: Use actually non-existing binary for test

danobi:
12345 is a PID that could plausibly exist. I've seen test failures locally and in CI along the following:

```
/home/runner/work/bpftrace/bpftrace/tests/utils.cpp:374: Failure
  pids.size()
    Which is: 1
  0

[  FAILED  ] utils.get_pids_for_program (1 ms)
```

Make the test more reliable by using an actually non-existent exe.

##### Checklist


tyroguru:
I hit this earlier this afternoon in CI testing - thanks for sorting!

PR #3220, opened 2024-06-05
Semantic analyser: fix checking record map types

### LINKED ISSUE #3218: Crash when assigning a record type to map
```
$ sudo ./build/src/bpftrace -e 'BEGIN { @x = count(); @x = *curtask }'
[sudo] password for dxu:
bpftrace: /home/dxu/dev/bpftrace/src/types.h:266: const std::string bpftrace::SizedType::GetName() const: Assertion `IsRecordTy()' failed.
fish: Job 1, 'sudo ./build/src/bpftrace -e 'B…' terminated by signal SIGABRT (Abort)
```

### `bpftrace --info` output
```
 $ sudo ./build/src/bpftrace --info
System
  OS: Linux 6.8.7-arch1-2 #1 SMP PREEMPT_DYNAMIC Fri, 19 Apr 2024 09:51:31 +0000
  Arch: x86_64

Build
  version: v0.20.0-186-gbbaa-dirty
  LLVM: 17.0.6
  unsafe probe: no
  bfd: yes
  liblldb (DWARF support): yes
  libsystemd (systemd notify support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes
  for_each_map_elem: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: yes
  uprobe_multi: yes
  raw_tp_special: yes
  iter: yes
```

viktormalik:
When assigning a struct type to a map already containing a type, semantic analyser checks if the types have the same name. However, if the map contains a non-struct type, this will cause a crash:

    # bpftrace -e 'BEGIN { @x = 0; @x = *curtask }'
    bpftrace: [...]/src/types.h:266: const std::string bpftrace::SizedType::GetName() const: Assertion `IsRecordTy()' failed.

The reason is that we detect the type mismatch but only log an error and continue with follow-up checks.

Fix the crash by only doing the names check if both the existing and the newly assigned types are records.

Fixes #3218.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,8 @@ and this project adheres to
   - [#3024](https://github.com/bpftrace/bpftrace/pull/3024)
 - Fix variable corruption when used as map key
   - [#3195](https://github.com/bpftrace/bpftrace/pull/3195)
+- Semantic analyser: fix checking record map types
+  - [#3220](https://github.com/bpftrace/bpftrace/pull/3220)
 #### Docs
 #### Tools
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2392,7 +2392,7 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
   const std::string &map_ident = assignment.map->ident;
   auto type = assignment.expr->type;
 
-  if (type.IsRecordTy()) {
+  if (type.IsRecordTy() && map_val_[map_ident].IsRecordTy()) {
     std::string ty = assignment.expr->type.GetName();
     std::string stored_ty = map_val_[map_ident].GetName();
     if (!stored_ty.empty() && stored_ty != ty) {
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -379,6 +379,11 @@ stdin:1:20-22: ERROR: Type mismatch for @x: trying to assign value of type 'stri
 kprobe:f { @x = 0; @x = "a"; }
                    ~~
 )");
+  test_error("kprobe:f { @x = 0; @x = *curtask; }", R"(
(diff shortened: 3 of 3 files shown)


jordalgo:
> The reason is that we detect the type mismatch but only log an error and continue with follow-up checks.

Where is this happening? And should we exit the function early in this case or throw?

viktormalik:
> > The reason is that we detect the type mismatch but only log an error and continue with follow-up checks.
> 
> Where is this happening?

`SemanticAnalyser::assign_map_type`

> And should we exit the function early in this case or throw?

Not sure, TBH. In most cases, we try to continue to emit as many errors as possible. In this case, though, `SemanticAnalyser::visit(AssignMapStatement &assignment)` wouldn't add any more errors so it would make sense to terminate it early. OTOH, it's not consistent with the rest of the semantic analyser where we just continue.

jordalgo:
> so it would make sense to terminate it early. OTOH, it's not consistent with the rest of the semantic analyser where we just continue.

Kinda seems like we need a larger audit of `SemanticAnalyser` - at least the error parts. I'll try to have a look a bit later.

jordalgo (approved):

PR #3225, opened 2024-06-07
CHANGELOG: Cleanup summaries to be more useful to users

ajor:
Instead of listing what internal details have changed, we should explain what user-visible effects the changes have.

Added a "Security" section, as recommended by keepachangelog.com.


jordalgo (approved):
Nice improvement!

PR #3227, opened 2024-06-08
Restrict looping over some map types

jordalgo:
It doesn't make sense to loop over these types:
- avg
- stats
- hist
- lhist

This is because these maps have complex values and keys, which don't work for iteration. Example:
```
$ sudo ./src/bpftrace -e 'BEGIN { @x[1] = avg(5); for ($kv : @x) { print(($kv.0, $kv.1)); } exit(); }'
Attaching 1 probe...
(1, )
(1, )

@x[1]: 5
```

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2080,6 +2080,12 @@ void SemanticAnalyser::visit(For &f)
   }
   Map &map = static_cast<Map &>(*f.expr);
 
+  if (!map.type.IsMapIterableTy()) {
+    LOG(ERROR, f.expr->loc, err_)
+        << "Loop expression does not support type: " << map.type;
+    return;
+  }
+
   // Validate body
   CollectNodes<Variable> vars_referenced;
   for (auto *stmt : *f.stmts) {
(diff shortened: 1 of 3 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -2080,6 +2080,13 @@ void SemanticAnalyser::visit(For &f)
   }
   Map &map = static_cast<Map &>(*f.expr);
 
+  if (map.type.IsAvgTy() || map.type.IsHistTy() || map.type.IsLhistTy() ||
+      map.type.IsStatsTy()) {
```
viktormalik, line 2084: Maybe create a helper method (e.g. `IsMapIterableTy`) to simplify management of the (un)supported types?

viktormalik (approved):
I'm wondering if we could come up with a way to allow iteration for these. But in the meantime, let's disable them as they are clearly broken.

src/ast/passes/semantic_analyser.cpp
jordalgo: Good idea. Will update.

jordalgo:
> I'm wondering if we could come up with a way to allow iteration for these. 

I was working on `avg` but it's a little tricky because it keeps two keys (count and total). I'm sure it's possible for all of these, it just requires some thought on the desired behavior.

Issue #2853, opened 2023-11-27
Big string / big record support


danobi:
Take, for example, this:

      $ sudo ./build/src/bpftrace -e 'BEGIN { print(*curtask); exit() }'
      error: <unknown>:0:0: in function BEGIN i64 (ptr): A call to built-in function 'memset' is not supported.

The reason is `curtask` dereference puts the entire struct on the stack.
Which not only doesn't fit, but does not qualify for unrolled memset builtin.

One potential way to handle this is to use the new `createGetScratchMap()`
infrastructure to "allocate" memory in `unop_ptr()` instead of from the stack.

But the drawback is that we can only have a single live allocation at a time.
Meaning, something like this would not be possible:

      $one = *curtask;
      $two = *curtask;
      print(($one, $two));

However, something like this could be possible:

      $one = *curtask;
      print($one);
      $two = *curtask;
      print($two);

That is b/c in the second example the lifetimes of `$one` and `$two` are disjoint.

I think above is a good starting point for big string / big record support.
We can do some basic lifetime analysis in semantic analyser (or a new pass?)
and error out if there's overlap.

In the future we can allocate as many scratch maps as we need (perhaps up
to a certain limit).

I'm not sure I've covered all the cases. But I think this is something good
to think about.

danobi:
Could create a scratch-map per big-string/big-record. It's a bit wasteful but we won't need to do lifetime analysis.

Another option is to disallow storing big-string/big-records on the stack. So for example, only allow them to be copied into maps or output ring buffers.

danobi:
We'll tackle the memset() issue in #3229. Big record support is rather esoteric, so we can address that if users ever ask for it

PR #3228, opened 2024-06-10
Support big strings

danobi:
Previously, all strings needed to be stored on the stack. Given that BPF
has a 512B stack limit, this put serious limits on strings. In practice,
it was limited to about 200B.

This PR raises the limit much higher to around 1024. The new limit
we hit is from the LLVM memset() builtin. It only supports up to 1K. To
go above this, we need to create our own memset() routine in BPF. This
is quite easy to do and will be attempted in a later commit.

The way big strings work is by assigning each str() call in the script a
unique ID. Then we create a percpu array map such that for each CPU,
there is a percpu entry that can accomodate each str() callsite. This
makes it so scripts can keep as many strings "on the stack" as they'd
like, with the tradeoff being the amount of memory we need to
preallocate.

As long as the kernel never allows BPF programs to nest (eg execute prog
A while prog A is already on the call stack), this approach is robust to
self-corruption.

This closes https://github.com/bpftrace/bpftrace/issues/305.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -717,9 +717,9 @@ void CodegenLLVM::visit(Call &call)
     }
     expr_ = nullptr;
   } else if (call.func == "str") {
+    uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
     // Largest read we'll allow = our global string buffer size
-    Value *strlen = b_.getInt64(
-        bpftrace_.config_.get(ConfigKeyInt::max_strlen));
+    Value *strlen = b_.getInt64(max_strlen);
     if (call.vargs->size() > 1) {
       auto scoped_del = accept(call.vargs->at(1));
       expr_ = b_.CreateIntCast(expr_, b_.getInt64Ty(), true);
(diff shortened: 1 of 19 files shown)


danobi:
In another window:
```
$ /usr/bin/asdf/asdf/dsf/sdfddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd
bash: /usr/bin/asdf/asdf/dsf/sdfddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd: No such file or directory
```

```
$ sudo BPFTRACE_MAX_STRLEN=1024 ./build/src/bpftrace -e 't:syscalls:sys_enter_execve { @[str(args.filename)] = count() }'
Attaching 1 probe...
^C

@[/usr/bin/asdf/asdf/dsf/sdfddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd]: 1
@[/home/dxu/.tmux/plugins/tmux-continuum/scripts/continuum_save.sh]: 2
@[/bin/sh]: 2
@[/usr/bin/cut]: 2
```

danobi:
Related but not done: #2853

src/main.cpp
```diff
@@ -269,25 +269,6 @@ static void parse_env(BPFtrace& bpftrace)
     config_setter.set(ConfigKeyInt::max_strlen, x);
   });
 
-  // in practice, the largest buffer I've seen fit into the BPF stack was 240
-  // bytes. I've set the bar lower, in case your program has a deeper stack than
-  // the one from my tests, in the hope that you'll get this instructive error
-  // instead of getting the BPF verifier's error.
-  uint64_t max_strlen = bpftrace.config_.get(ConfigKeyInt::max_strlen);
-  if (max_strlen > 200) {
```
jordalgo, line 277: why remove this check instead of updating it?

tests/runtime/call
```diff
@@ -278,6 +278,8 @@ ENV BPFTRACE_CACHE_USER_SYMBOLS=PER_PROGRAM
 PROG uprobe:./testprogs/uprobe_symres_exited_process:test { print(ustack); exit(); }
 EXPECT_REGEX ^\s+test\+[0-9]+\s+test2\+[0-9]+\s+main\+[0-9]+
 AFTER ./testprogs/disable_aslr ./testprogs/uprobe_symres_exited_process
+# Required to skip function prologue
+REQUIRES_FEATURE dwarf
```
jordalgo, line 282: driveby fixes?

jordalgo:
This all looks great and surprisingly straightforward.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -719,37 +720,43 @@ void CodegenLLVM::visit(Call &call)
...
-                          expr_,
-                          arg0->type.GetAS(),
-                          call.loc);
-    b_.CreateLifetimeEnd(strlen);
+    b_.CreateProbeReadStr(
+        ctx_, buf, strlen, expr_, arg0->type.GetAS(), call.loc);
+    b_.CreateBr(lookup_merge_block);
+
+    // Think of this like an assert(). In practice, we cannot fail to lookup a
+    // percpu array map unless we have a coding error. Rather than have some
```
jordalgo, line 750: By "coding error" you mean we pass in a key that does not exist right?

src/main.cpp
danobi: Yeah fair enough. I'll delete it when the memset() goes in.

tests/runtime/call
danobi: Oh, that change merged already. Lemme rebase onto it

src/ast/passes/codegen_llvm.cpp
danobi: Yes

jordalgo (approved):

Issue #3021, opened 2024-02-26
Can't use subprograms from multiple probes


ajor:
The BPF verifier requires that all defined subprograms are used by the main program.

Our current approach involves appending all subprograms into `.text`, whether they are used by a given probe or not. This means that if a script contains multiple probes, each with their own subprograms, the verifier will reject it.

### What reproduces the bug?
```
BEGIN
{
  @map[0] = 1;
  print(len(@map))
}
END
{
  print(len(@map))
}
```

```
Attaching 2 probes...

Error log: 
number of funcs in func_info doesn't match number of subprogs
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

ERROR: Error loading program: BEGIN
```

The easy fix for `len()` would be to simply merge duplicate subprograms into a single `len()` subprogram.
However, we will also need to implement a proper fix for #2955.

libbpf encountered this same problem previously: https://lore.kernel.org/netdev/5f4f311d37ba8_5f3982088@john-XPS-13-9370.notmuch/T/

ajor:
@viktormalik How much of the BpfProgram logic will remain when we complete the libbpf migration? Do you think libbpf will be able to solve this issue for us?

viktormalik:
> The BPF verifier requires that all defined subprograms are used by the main program.
> 
> Our current approach involves appending all subprograms into `.text`, whether they are used by a given probe or not. This means that if a script contains multiple probes, each with their own subprograms, the verifier will reject it.

Ah, so we would need to copy just the required symbols, not the entire `.text`. That sounds tricky as there may be relative offsets which would have to be fixed.

> @viktormalik How much of the BpfProgram logic will remain when we complete the libbpf migration? Do you think libbpf will be able to solve this issue for us?

Yes, it will solve the issue completely as all of the relocations code will go away. I'll put a higher priority on that work.

viktormalik:
What we could do here is emit subprograms into special sections named e.g. `.text.<probe-name>` and then transform the relevant section into `.text` for the given probe in `BpfProgram`.

That should be fairly easy to implement and should temporarily mitigate this issue until we fully move to libbpf.

jordalgo:
>  e.g. .text.<probe-name> and then transform the relevant section into .text for the given probe in BpfProgram.

QQ: should it be probe id? I think you can have multiple of the same probe name right (different preds) ?

I can try to take a look a fixing this issue if folks point me in the right direction :)

[...]

PR #3233, opened 2024-06-11
Add user-friendly error for multiple for loops

viktormalik:
Currently, we do not allow for-loops to be used from multiple probes at the same time (see #3021 for details). If that happens, detect it in the semantic analyser and print a nice error instead of a generic `Error loading program`.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2065,6 +2065,13 @@ void SemanticAnalyser::visit(For &f)
    *     }
    */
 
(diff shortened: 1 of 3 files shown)


viktormalik:
Not updating the changelog as for loops are not released, yet.

danobi:
Semantic analyser unit test? It'd help make more clear what we are preventing.

viktormalik:
> Semantic analyser unit test? It'd help make more clear what we are preventing.

Yeah, I was reluctant to add it as this is a "missing feature" but with the test, we'll also get notified when we fix the issue. Added.

jordalgo (approved):

PR #3237, opened 2024-06-12
Add big memset() support

danobi:
This PR unlocks full support for big strings. Users can now have strings of
unlimited (until you run out of kernel memory) size. There's still the issue
of path(), buf(), etc. helpers trying to allocate BPFTRACE_MAX_STRLEN bytes
on the stack. We'll fix that next, but even if that doesn't go in, users now have
the option.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -293,6 +293,13 @@ Value *IRBuilderBPF::GetMapVar(const std::string &map_name)
   return module_.getGlobalVariable(bpf_map_name(map_name));
 }
 
+Value *IRBuilderBPF::GetNull()
+{
+  return ConstantExpr::getCast(Instruction::IntToPtr,
+                               getInt64(0),
+                               GET_PTR_TY());
+}
+
 CallInst *IRBuilderBPF::CreateMapLookup(Map &map,
                                         Value *key,
                                         const std::string &name)
@@ -410,10 +417,9 @@ CallInst *IRBuilderBPF::createGetScratchMap(const std::string &map_name,
       module_.getContext(), "lookup_" + name + "_failure", parent);
   BasicBlock *lookup_merge_block = BasicBlock::Create(
       module_.getContext(), "lookup_" + name + "_merge", parent);
-  Value *condition = CreateICmpNE(
-      CreateIntCast(call, GET_PTR_TY(), true),
-      ConstantExpr::getCast(Instruction::IntToPtr, getInt64(0), GET_PTR_TY()),
-      "lookup_" + name + "_cond");
+  Value *condition = CreateICmpNE(CreateIntCast(call, GET_PTR_TY(), true),
+                                  GetNull(),
+                                  "lookup_" + name + "_cond");
   CreateCondBr(condition, lookup_merge_block, lookup_failure_block);
 
   SetInsertPoint(lookup_failure_block);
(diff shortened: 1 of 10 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -200,6 +209,53 @@ AllocaInst *IRBuilderBPF::CreateAllocaBPF(int bytes, const std::string &name)
...
+  if (size > 512 && bpftrace_.feature_->has_helper_probe_read_kernel()) {
+    // Note we are "abusing" bpf_probe_read_kernel() by reading from NULL
+    // which triggers a call into the kernel-optimized memset().
+    //
+    // Upstream blesses this trick so we should be able to count on them
+    // to maintain these semantics.
+    //
+    // Also note we are avoiding a call to CreateProbeRead(), as it wraps
+    // calls to probe read helpers with the -kk error reporting feature.
+    // The call here will always fails and we want it that way. So avoid
```
jordalgo, line 228:
```suggestion
    // The call here will always fail and we want it that way. So avoid
```

src/ast/irbuilderbpf.cpp
```diff
@@ -200,6 +209,53 @@ AllocaInst *IRBuilderBPF::CreateAllocaBPF(int bytes, const std::string &name)
...
+    createCall(proberead_func_type,
+               proberead_func,
+               { ptr, getInt32(size), GetNull() },
+               probeReadHelperName(probe_read_id));
+  } else {
+    // Use unrolled memset for memsets less than 512 bytes mostly for
+    // correctness.
+    //
+    // It appears that helper based memsets obscure LLVM stack optimizer view
+    // into memory usage. Such that builtin memset based programs bloat with
```
jordalgo, line 249:
```suggestion
    // into memory usage such that builtin memset based programs bloat with
```

src/ast/irbuilderbpf.cpp
```diff
@@ -200,6 +209,53 @@ AllocaInst *IRBuilderBPF::CreateAllocaBPF(int bytes, const std::string &name)
...
+    // helper based enough to where LLVM BPF backend will barf.
```
jordalgo, line 250: not sure if I understand this part of the sentence "bloat with helper based enough..."

danobi: Reworded it a bit. Lemme know if it makes sense now

src/ast/irbuilderbpf.cpp
```diff
@@ -200,6 +209,54 @@ AllocaInst *IRBuilderBPF::CreateAllocaBPF(int bytes, const std::string &name)
   return CreateAllocaBPF(ty, name);
 }
 
+void IRBuilderBPF::CreateMemsetBPF(Value *ctx,
+                                   Value *ptr,
+                                   Value *val,
+                                   uint32_t size)
+{
+  assert(ctx && ctx->getType() == GET_PTR_TY());
```
jordalgo, line 217: Looks like ctx is unused in this function beside the assertion. That on purpose?

jordalgo: Looks like it's the reason you had to pass ctx around for all these functions.

src/ast/irbuilderbpf.cpp
jordalgo: Yeah, better. Thanks!

src/ast/irbuilderbpf.cpp
danobi: Ah good catch. Previously I was calling the CreateProbeRead() function which required a ctx. But eventually I realized it wasn't suitable for this use case. Forgot to delete ctx. Will fix.

jordalgo (approved):

PR #3240, opened 2024-06-13
Fix: BTF type tag resolution

jordalgo:
This adds support for 'BTF_KIND_TYPE_TAG' in that we ignore it and follow the pointer to the base type.

This fixes issue: https://github.com/bpftrace/bpftrace/issues/3221

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -348,6 +348,21 @@ const struct btf_type *BTF::btf_type_skip_modifiers(const struct btf_type *t,
   return t;
 }
 
+__u32 BTF::get_type_tags(std::unordered_set<std::string> &tags,
+                         const BTFId &btf_id) const
+{
+  __u32 id = btf_id.id;
+  const struct btf_type *t = btf__type_by_id(btf_id.btf, btf_id.id);
+
+  while (t && btf_is_type_tag(t)) {
+    tags.insert(btf_str(btf_id.btf, t->name_off));
+    id = t->type;
+    t = btf__type_by_id(btf_id.btf, t->type);
+  }
+
+  return id;
+}
+
 SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
 {
   const struct btf_type *t = btf__type_by_id(btf_id.btf, btf_id.id);
@@ -375,9 +390,12 @@ SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
     if (resolve_structs)
       resolve_fields(stype);
   } else if (btf_is_ptr(t)) {
-    // t->type is the pointee type
+    const BTFId pointee_btf_id = { .btf = btf_id.btf, .id = t->type };
+    std::unordered_set<std::string> tags;
+    auto id = get_type_tags(tags, pointee_btf_id);
     stype = CreatePointer(
-        get_stype(BTFId{ .btf = btf_id.btf, .id = t->type }, false));
+        get_stype(BTFId{ .btf = btf_id.btf, .id = id }, false));
+    stype.SetBtfTypeTags(std::move(tags));
   } else if (btf_is_array(t)) {
     auto *array = btf_array(t);
     const auto &elem_type = get_stype(
--- a/src/btf.h
+++ b/src/btf.h
@@ -103,6 +103,13 @@ private:
   std::set<std::string> get_all_structs_from_btf(const struct btf* btf) const;
   std::unordered_set<std::string> get_all_iters_from_btf(
       const struct btf* btf) const;
+  /*
+   * Similar to btf_type_skip_modifiers this returns the id of the first
+   * type that is not a BTF_KIND_TYPE_TAG while also populating the tags set
+   * with the tag/attribute names from the BTF_KIND_TYPE_TAG types it finds.
+   */
+  __u32 get_type_tags(std::unordered_set<std::string>& tags,
+                      const BTFId& btf_id) const;
 
   __s32 start_id(const struct btf* btf) const;
 
--- a/src/types.h
+++ b/src/types.h
@@ -3,6 +3,7 @@
 #include <cassert>
 #include <map>
 #include <memory>
+#include <optional>
 #include <ostream>
 #include <sstream>
 #include <string>
@@ -154,6 +155,8 @@ private:
   AddrSpace as_ = AddrSpace::none;
   bool is_signed_ = false;
   bool ctx_ = false; // Is bpf program context
+  std::unordered_set<std::string> btf_type_tags_ = {}; // Only populated for
+                                                       // Type::pointer
 
   friend class cereal::access;
   template <typename Archive>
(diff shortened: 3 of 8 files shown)


tests/runtime/btf
```diff
@@ -97,3 +97,9 @@ PROG kprobe:kvm:x86_emulate_insn { $ctxt = (struct x86_emulate_ctxt *) arg0; pri
 EXPECT Attaching 1 probe...
 TIMEOUT 2
 REQUIRES lsmod | grep '^kvm'
+
+NAME btf_type_tag_access
+PROG BEGIN { printf("SUCCESS %d\n", curtask->parent->pid); exit() }
+EXPECT_REGEX SUCCESS [0-9][0-9]*
+TIMEOUT 5
+REQUIRES_FEATURE btf
```
danobi, line 105: Note this requires a clang built linux kernel. So this is probably a noop for the reasonably close future. Perhaps leave a comment?

jordalgo: Oh? Are most kernels (I guess I also in our CI) not built with clang?

danobi: Most (if not all) kernels in wild are built with gcc. Only folks using clang built Linux I know of is meta and Google. Support came from google.

In kernel source the btf type tag annotation is ifdefd out for gcc. Look for the __has_attribute compiler built in. So it looks like a regular field in btf when kernel is compiled with gcc.

src/btf.cpp
```diff
@@ -387,6 +387,8 @@ SizedType BTF::get_stype(const BTFId &btf_id, bool resolve_structs)
     } else {
       stype = CreateArray(array->nelems, elem_type);
     }
+  } else if (btf_is_type_tag(t)) {
```
danobi, line 390: I'm conflicted whether or not we should allowlist certain known type tags. For example, rcu type tag can safely be ignored while percpu type tag cannot. The latter requires special helpers to access percpu memory.

On the other hand, only clang-built-linux will have vmlinux type tags. And clang-built-linux is not widely adopted in open source land.

Perhaps a comment calling this out will suffice for now?

jordalgo:
Here is what a tag failure looks like:
```
$ sudo ./src/bpftrace -e 'kfunc:vmlinux:__traceiter_cgroup_mkdir { print(args.cgrp->rstat_cpu->bstat); }'
stdin:1:42-68: ERROR: Attempting to access pointer field 'rstat_cpu' with unsupported tag attribute: percpu
kfunc:vmlinux:__traceiter_cgroup_mkdir { print(args.cgrp->rstat_cpu->bstat); }
                                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/types.h
```diff
@@ -154,6 +155,8 @@ class SizedType {
   AddrSpace as_ = AddrSpace::none;
   bool is_signed_ = false;
   bool ctx_ = false; // Is bpf program context
+  std::optional<std::string> btf_type_tag_ = std::nullopt; // Only populated for
```
danobi, line 158: Is optional necessary? std::string::empty() is unambiguous, right?

src/types.h
```diff
@@ -206,6 +209,17 @@ class SizedType {
     as_ = as;
   }
 
+  void SetBtfTypeTag(const std::string &tag)
+  {
+    assert(IsPtrTy());
+    btf_type_tag_ = tag;
+  }
+
+  std::optional<std::string> GetBtfTypeTag() const
```
danobi, line 218: Also assert for ptr type?

src/ast/passes/field_analyser.cpp
```diff
@@ -101,6 +101,21 @@ void FieldAnalyser::visit(FieldAccess &acc)
       if (sized_type_.HasField(acc.field))
         field_type = sized_type_.GetField(acc.field).type;
 
+      if (field_type.IsPtrTy()) {
+        auto tag = field_type.GetBtfTypeTag();
+        /*
+         * Currently only "rcu" is safe. "percpu", for example, requires special
+         * unwrapping with `bpf_per_cpu_ptr` which is not yet supported.
+         */
+        static const std::string allowed_tag = "rcu";
```
danobi, line 110: Nit: string_view more appropriate

src/ast/passes/field_analyser.cpp
danobi, line 1: IIUC field_analyser exists to do field resolution and also figure out what C definitions from BTF need to be dumped. I think semantic_analyser FieldAccess visitor is a better place for the error check.

might also be possible to add a unit test there. possibly by directly injecting a struct into bpftrace_.structs

src/types.h
danobi: Also, looking at https://www.kernel.org/doc/html/next/bpf/btf.html#btf-kind-type-tag, seems like multiple type tags are supported. so maybe this should even be a vector

jordalgo: Oh good catch. Let me look into supporting that. I guess if any in that list are not "rcu" we need to reject.

src/ast/passes/field_analyser.cpp
jordalgo: I'll look into putting it in semantic_analyser but it looks FieldAccess is too late in the process: https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L2161

Unless the type_tags should be added to the pointee type instead of the pointer type.

jordalgo:
@danobi Thanks for all the good suggestions! I think I addressed everything.

src/btf.h
```diff
@@ -103,6 +103,8 @@ class BTF {
   std::set<std::string> get_all_structs_from_btf(const struct btf* btf) const;
   std::unordered_set<std::string> get_all_iters_from_btf(
       const struct btf* btf) const;
+  __u32 get_type_tags(std::unordered_set<std::string>& tags,
```
danobi, line 106: Document that this returns the type_id with type tags removed?

tests/mocks.cpp
```diff
@@ -124,6 +125,17 @@ void setup_mock_bpftrace(MockBPFtrace &bpftrace)
   bpftrace.structs.Lookup("struct _tracepoint_file_filename")
       .lock()
       ->AddField("filename", ptr_type, 8, std::nullopt, false);
+  bpftrace.structs.Add("struct _tracepoint_btf_tag", 8);
```
danobi, line 128: maybe make this struct a bit more realistic just in case? right now you've basically made a union, right?
```suggestion
  bpftrace.structs.Add("struct _tracepoint_btf_tag", 16);
```

tests/mocks.cpp
```diff
@@ -124,6 +125,17 @@ void setup_mock_bpftrace(MockBPFtrace &bpftrace)
...
+  auto ptr_type_w_tag = CreatePointer(CreateInt8());
+  ptr_type_w_tag.SetBtfTypeTags({ "rcu" });
+  auto ptr_type_w_bad_tag = CreatePointer(CreateInt8());
+  ptr_type_w_bad_tag.SetBtfTypeTags({ "rcu", "percpu" });
+  bpftrace.structs.Lookup("struct _tracepoint_btf_tag")
+      .lock()
+      ->AddField("parent", ptr_type_w_tag, 8, std::nullopt, false);
```
danobi, line 135:
```suggestion
      ->AddField("parent", ptr_type_w_tag, 0, std::nullopt, false);
```

danobi (commented):
small nits: but lgtm otherwise

tests/mocks.cpp
jordalgo: These are supposed to be 8 and 16 - though TBH, I don't quite understand why. The only field marked as offset 0 above is one that fails the unit test e.g. `test("tracepoint:file:filename { args.common_field }", 1);`

danobi: Oh right. The first 8 bytes of tracepoint args is some internal fields for the kernel. We don't allow the users to access those

danobi (approved):

PR #3242, opened 2024-06-14
utils: fix kernel headers not found warning

martinetd:
The code would make ksrc and kobj empty then try to print them, this would always print empty strings.
Store is_dir() checks as bool instead and use these, the strings cannot be empty if the check passed.

##### Checklist

^ doesn't really seem applicable here; tested manually on nixos where directory isn't found


jordalgo (approved):
Good catch! Thanks for the fix.

PR #3243, opened 2024-06-14
clang_parser system_include_paths: allow overriding at compile time

martinetd:
While bpftrace depends on libclang it can be installed without a clang frontend, so some distributions might want to make these paths fixed as they are unlikely to change.

In particular, this is necessary to include system libraries as used by older versions of tcpaccept.bt (they now no longer require these since #3152, but that illustrate this was a recurring problem)

##### Checklist

^ There are no test for KERNEL_HEADERS_DIR either and such build-related tests are a bit o a pain, for what it's worth it's indirectly tested in nixos here: https://github.com/NixOS/nixpkgs/pull/319836

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,8 @@ set(BUILD_FUZZ OFF CACHE BOOL "Build bpftrace for fuzzing")
 set(USE_LIBFUZZER OFF CACHE BOOL "Use libfuzzer for fuzzing")
 set(FUZZ_TARGET "codegen" CACHE STRING "Fuzzing target")
 set(ENABLE_SYSTEMD OFF CACHE BOOL "Enable systemd integration")
+set(KERNEL_HEADERS_DIR "" CACHE PATH "Hard-code kernel headers directory")
+set(SYSTEM_INCLUDE_PATHS "auto" CACHE STRING "Hard-code system include paths (colon separated, the default value \"auto\" queries clang at runtime)")
 
 set(ENABLE_SKB_OUTPUT ON CACHE BOOL "Enable skb_output, will include libpcap")
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -90,11 +90,15 @@ endif()
 
 # compile definitions
 
-set(KERNEL_HEADERS_DIR "" CACHE PATH "Hard-code kernel headers directory")
 if (KERNEL_HEADERS_DIR)
   MESSAGE(STATUS "Using KERNEL_HEADERS_DIR=${KERNEL_HEADERS_DIR}")
   target_compile_definitions(runtime PUBLIC KERNEL_HEADERS_DIR="${KERNEL_HEADERS_DIR}")
 endif()
+
+if (NOT SYSTEM_INCLUDE_PATHS EQUAL "auto")
+  MESSAGE(STATUS "Using SYSTEM_INCLUDE_PATHS=${SYSTEM_INCLUDE_PATHS}")
+endif()
+target_compile_definitions(runtime PUBLIC SYSTEM_INCLUDE_PATHS="${SYSTEM_INCLUDE_PATHS}")
 
 # This is run on every build to ensure the version string is always up-to-date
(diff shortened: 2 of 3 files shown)


src/CMakeLists.txt
danobi, line 1: nit: looks like your addition and `KERNEL_HEADERS_DIR` are the only two build flags we keep apart from the rest. would be good to move both to top level cmakelists so other packagers/users can find them more easily

src/clang_parser.cpp
```diff
@@ -844,6 +844,12 @@ std::string ClangParser::get_arch_include_path()
 std::vector<std::string> ClangParser::system_include_paths()
 {
   std::vector<std::string> result;
+#ifdef SYSTEM_INCLUDE_PATHS
+  std::istringstream lines(SYSTEM_INCLUDE_PATHS);
+  std::string line;
+  while (std::getline(lines, line, ':'))
+      result.push_back(trim(line));
+#else
```
danobi, line 852: I think it'd be better to avoid preprocessor conditionals here. Better to keep it all visible to compiler so it's less likely to bitrot and break one day

danobi (commented):
ty for upstreaming the changes!

src/CMakeLists.txt
martinetd: That makes sense, I'll move the two

src/clang_parser.cpp
martinetd: Hmm, we can make the default define value be an empty string and condition on the value in the code?
It won't be possible to set the config knob to an empty string at this point but I guess it'd be equivalent to setting it to /noexist or any other missing directory

martinetd (commented):
Thanks for the review, probably won't have time today but will update this tomorrow

src/clang_parser.cpp
danobi: Maybe you could make the string `"-"` disable system include paths and document it like that. I'm guessing there's a magical cmake way to do what you want without resorting to a magic character but explicit is probably simpler in this case

martinetd: I took a slightly different approach: the keyword "auto" is hooks into the existing code, and the default value is auto, so:
- default behaviour is exactly the same
- explicit paths like /nix/store/whatever-glibc-vers/include works
- I've also tested things like auto:/foo and /foo:auto, even if I don't expect anyone to use that.
- the fallback is still there though, so empty string will yield the default /usr/include & /usr/local/include... but my first version also did that, so unless someone actually comes and requests such a behaviour I think this is fine.

CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ set(BUILD_FUZZ OFF CACHE BOOL "Build bpftrace for fuzzing")
 set(USE_LIBFUZZER OFF CACHE BOOL "Use libfuzzer for fuzzing")
 set(FUZZ_TARGET "codegen" CACHE STRING "Fuzzing target")
 set(ENABLE_SYSTEMD OFF CACHE BOOL "Enable systemd integration")
+set(KERNEL_HEADERS_DIR "" CACHE PATH "Hard-code kernel headers directory")
+set(SYSTEM_INCLUDE_PATHS "auto" CACHE STRING "Hard-code system include paths (colon separated, auto queries clang at runtime)")
```
viktormalik, line 26: It took me a while to figure out the description, maybe a slight change like this would help?
```suggestion
set(SYSTEM_INCLUDE_PATHS "auto" CACHE STRING "Hard-code system include paths (colon separated, the default value \"auto\" queries clang at runtime)")
```

viktormalik (approved):
LGTM, thanks! Just one nit below (feel free to ignore it).

CMakeLists.txt
martinetd: Sorry about that, reading back I agree it could read as "auto-queries clang.." which didn't make sense -- your suggestion looks good, applied!

viktormalik: Yep, that's exactly how my mind read it :-) Thanks!

danobi (approved):

PR #3246, opened 2024-06-17
Unify and improve ignoring of missing probes

### LINKED ISSUE #3235: Attaching to non-existing uprobes fails
#3155 broke scripts which use probes attaching to multiple uprobes when some of the target functions do not exist. Previously, we would silently ignore such attach points, now bpftrace fails with:
```
ERROR: Could not resolve symbol: ...
```

Among others, this also breaks a couple of tools (`ssllatency.bt`, `sslsnoop.bt`, `threadsnoop.bt`). CI didn't catch that since we're just parsing the tool script and do not try to attach them. This will be fixed by #3203.

While this is clearly wrong and we need to fix it, I realized that we're doing different things for uprobes and kprobes:
- for uprobes, we call `resolve_uname` and if the function doesn't exist, we silently ignore it,
- for kprobes, we always try to attach and if that fails, we issue a warning for probes with multiple attach points and continue.

I'm not sure if this is on purpose but I think that it'd be good to unify this. The options are:
1. Silently ignore non-existing attach points for both kprobes and uprobes (possibly for kfuncs, too).
2. Always issue a warning for non-existing attach points.
3. Only show warnings in verbose mode.
4. Add a CLI option or a config option to show the warnings.

All in all, any of the above options will affect the existing scripts including the tools shipped in the main repo. Either the uprobe-based tools will start showing warnings or the kprobe-based tools will stop showing them.

Personally, I'm in favor of option (3) as I find the warnings needlessly polluting the default output of tracing tools. For instance, some tools are designed in a way that there's always just one kprobe existing (b/c it was renamed in the kernel at some point) and therefore will always show the warning.

viktormalik:
For kprobes and uprobes, when a probe is attaching to multiple functions, we do not fail when some functions do not exist. The reason is that k(u)probes are dynamic events which regularly change between versions of kernel/binaries and we want to allow the tools to attach to all potential events while expecting some attachments to fail.

The problem is that we're not doing the ignoring consistently. For kprobes, we always show a warning if an attachment fails. For uprobes, bpftrace currently fails due to #3235 but even without the bug, we would just silently ignore the missing functions.

This PR unifies the approaches by always showing the warning. In addition, it adds a new config option `warn_missing_probes`, by default set to true, which allows to turn the warnings off. This is done for all the shipped tools.

Fixes #3235.

The first commit also fixes another attachment issue caused by a combination of kprobe_multi and progfd caching. See the commit message for details.

Note: one potential problem with ignoring missing probe warnings is that when all attach points of a probe are missing, the probe will not be attached at all without any warning. Normally, this is IMHO fine as the user must have requested to ignore warnings so he must be prepared that this may happen. The only problem is the shipped `xfsdist.bt` tool which behaves like this when the XFS module is not loaded but I wouldn't expect people to try to trace XFS calls without XFS being present on their system.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -2415,11 +2415,11 @@ i:s:10 {
 Count how often this function is called.
 
 Using `@=count()` is conceptually similar to `@++`.
-The difference is that the `count()` function uses a map type optimized for 
-writing (PER_CPU), increasing performance and correctness. However, sync reads 
-can be expensive as bpftrace needs to iterate over all the cpus to collect and 
+The difference is that the `count()` function uses a map type optimized for
+writing (PER_CPU), increasing performance and correctness. However, sync reads
+can be expensive as bpftrace needs to iterate over all the cpus to collect and
 sum these values.
-Note: In contrast to hash maps (e.g. `@++`), multiple writers to a shared 
+Note: In contrast to hash maps (e.g. `@++`), multiple writers to a shared
 global var might lose counts as bpftrace doesn't update them atomically.
 
 ----
@@ -2586,11 +2586,11 @@ kprobe:vfs_read {
 Calculate the sum of all `n` passed.
 
 Using `@=sum(5)` is conceptually similar to `@+=5`.
-The difference is that the `sum()` function uses a map type optimized for 
-writing (PER_CPU), increasing performance and correctness. However, sync reads 
-can be expensive as bpftrace needs to iterate over all the cpus to collect and 
+The difference is that the `sum()` function uses a map type optimized for
+writing (PER_CPU), increasing performance and correctness. However, sync reads
+can be expensive as bpftrace needs to iterate over all the cpus to collect and
 sum these values.
-Note: In contrast to hash maps (e.g. `@+=5`), multiple writers to a shared 
+Note: In contrast to hash maps (e.g. `@+=5`), multiple writers to a shared
 global var might lose updates as bpftrace doesn't update them atomically.
 
 ----
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -3427,6 +3427,19 @@ Default: 0
 Maximum number of levels of nested field accesses for tracepoint args.
 0 is unlimited.
 
+=== missing_probes
+
+Default: `warn`
+
+Controls handling of probes with multiple kprobe or uprobe attach points which
+cannot be attached to some functions because they do not exist in the kernel or
+in the traced binary.
+
+The possible options are:
+- `error` - always fail on missing probes
+- `warn` - print a warning but continue execution
+- `ignore` - silently ignore missing probes
+
 === perf_rb_pages
 
 Default: 64
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -155,17 +155,16 @@ int AttachedProbe::detach_raw_tracepoint(void)
 AttachedProbe::AttachedProbe(Probe &probe,
                              BpfProgram &&prog,
                              bool safe_mode,
-                             BPFfeature &feature,
-                             BTF &btf)
-    : probe_(probe), prog_(std::move(prog)), btf_(btf)
+                             BPFtrace &bpftrace)
+    : probe_(probe), prog_(std::move(prog)), bpftrace_(bpftrace)
 {
-  load_prog(feature);
+  load_prog(*bpftrace.feature_);
   LOG(V1) << "Attaching " << probe_.orig_name;
   switch (probe_.type) {
     case ProbeType::special:
       // If BPF_PROG_TYPE_RAW_TRACEPOINT is available, no need to attach prog
       // to anything -- we will simply BPF_PROG_RUN it
-      if (!feature.has_raw_tp_special())
+      if (!bpftrace_.feature_->has_raw_tp_special())
         attach_uprobe(getpid(), safe_mode);
       break;
     case ProbeType::kprobe:
@@ -207,15 +206,14 @@ AttachedProbe::AttachedProbe(Probe &probe,
 AttachedProbe::AttachedProbe(Probe &probe,
                              BpfProgram &&prog,
                              int pid,
-                             BPFfeature &feature,
-                             BTF &btf,
+                             BPFtrace &bpftrace,
                              bool safe_mode)
-    : probe_(probe), prog_(std::move(prog)), btf_(btf)
+    : probe_(probe), prog_(std::move(prog)), bpftrace_(bpftrace)
 {
-  load_prog(feature);
+  load_prog(*bpftrace_.feature_);
   switch (probe_.type) {
     case ProbeType::usdt:
-      attach_usdt(pid, feature);
+      attach_usdt(pid, *bpftrace_.feature_);
       break;
     case ProbeType::watchpoint:
     case ProbeType::asyncwatchpoint:
@@ -441,8 +439,7 @@ static void check_alignment(std::string &path,
   }
 }
 
-bool AttachedProbe::resolve_offset_uprobe(bool safe_mode,
-                                          bool error_on_missing_symbol)
+bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
 {
   struct bcc_symbol_option option = {};
   struct symbol sym = {};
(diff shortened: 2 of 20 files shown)


viktormalik:
There are several user-facing changes here so I ended up adding changelog entries for almost each commit (except for the second one as that fixes an unreleased bug).

src/attached_probe.cpp
```diff
@@ -478,9 +479,16 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode)
...
-    if (!sym.start)
-      throw FatalUserException("Could not resolve symbol: " + probe_.path +
-                               ":" + symbol);
+    if (!sym.start) {
+      const std::string msg = "Could not resolve symbol: " + probe_.path + ":" +
+                              symbol;
+      if (error_on_missing_symbol) {
+        throw FatalUserException(msg);
+      } else {
+        LOG(WARNING) << msg << ", skipping.";
```
danobi: I think it'd be more clear to say "skipping probe." rather than just "skipping."

src/attached_probe.cpp
```diff
@@ -478,9 +479,16 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode)
     sym.name = symbol;
     bcc_elf_foreach_sym(probe_.path.c_str(), sym_name_cb, &option, &sym);
 
-    if (!sym.start)
-      throw FatalUserException("Could not resolve symbol: " + probe_.path +
-                               ":" + symbol);
+    if (!sym.start) {
+      const std::string msg = "Could not resolve symbol: " + probe_.path + ":" +
+                              symbol;
```
danobi, line 482: Maybe add to end of message: "Cannot attach probe".

src/attached_probe.cpp
```diff
@@ -1155,7 +1163,7 @@ void AttachedProbe::attach_uprobe(int pid, bool safe_mode)
     return;
   }
 
-  if (!resolve_offset_uprobe(safe_mode))
+  if (!resolve_offset_uprobe(safe_mode, probe_.orig_name == probe_.name))
```
danobi, line 1173: So basically if multiple attachpoints are on a probe, we allow missing attachpoint with a warning. But if only one attachpoint is specified, it's an error.

I think this makes sense

danobi (commented):
overall lgtm

src/attached_probe.cpp
viktormalik: Yes, exactly. That's also how it works for kprobes.

src/attached_probe.cpp
viktormalik: Good point, I updated the error/warning messages.

man/adoc/bpftrace.adoc
```diff
@@ -2415,11 +2415,11 @@ i:s:10 {
 Count how often this function is called.
 
 Using `@=count()` is conceptually similar to `@++`.
-The difference is that the `count()` function uses a map type optimized for
```
jordalgo, line 2418: Thanks for removing the end space. I have the habit of adding them out of fear.

src/attached_probe.cpp
```diff
@@ -976,6 +986,14 @@ void AttachedProbe::attach_kprobe(bool safe_mode)
     funcname = modname + ":" + funcname;
   }
 
+  // If the user requested to ignore warnings on non-existing probes and the
```
jordalgo, line 989: Super nit to fix our comments - turn multi line into:
```
/*
 */
```

tools/gethostlatency.bt
```diff
@@ -19,6 +19,8 @@
  * 08-Sep-2018	Brendan Gregg	Created this.
  */
 
+config = { warn_missing_probes = 0; }
```
jordalgo, line 22: I'm not saying we should change this but I guess we should be aware that this will break this script on older bpftrace versions that don't support the config syntax. I guess that's fine as people can just use an older version of the repo. -- thinking out loud.

jordalgo (approved):

tools/gethostlatency.bt
viktormalik: Yeah, since we release the scripts in lockstep with the tool, I think it's fine if the tools do not work for previous versions.

src/attached_probe.cpp
viktormalik: We use both variants for multi-line comments (I think what I have is "C++ style") so I'd keep it as it is.

jordalgo: From what I understand you can use '//' for multi-line comments in C++ but I don't think that's the preferred style. I'm trying to move us to use '/**/' for multiline.

ajor:
I'm thinking that it might be useful to add the option to have missing probes trigger an error instead of a warning too. e.g. in some kind of automated tracing system it's unlikely that people will see the warnings - an error would stop them going unnoticed.

We don't have to implement that now, but it'd be good to make the config option future-proof, e.g. maybe with options like:
```
missing_probes = "ignore"
missing_probes = "warn"
missing_probes = "error"
```

jordalgo:
+1 to @ajor 's suggestion for making the config an enum of sorts

viktormalik:
@jordalgo please check 8113d594ef3ce84afb9100173d14a60f3e7e7e55, it's a bit of hacky but I didn't come with another solution which wouldn't make the code too complicated. I used the same approach for the new config option.

ajor (approved):
It looks like the warnings / errors for kprobes are a different code path than for uprobes - would it make sense to add tests for the new config option with kprobes as well?

src/attached_probe.h
```diff
@@ -23,13 +24,11 @@ class AttachedProbe {
   AttachedProbe(Probe &probe,
                 BpfProgram &&prog,
                 bool safe_mode,
-                BPFfeature &feature,
-                BTF &btf);
+                BPFtrace &bpftrace);
```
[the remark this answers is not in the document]
viktormalik: Yes, you just described my line of thought - I started with passing just `Config` and moving to `BPFtrace` once I realized that I need `is_traceable_func`, too.

viktormalik:
> It looks like the warnings / errors for kprobes are a different code path than for uprobes - would it make sense to add tests for the new config option with kprobes as well?

Yes, good idea, added.

viktormalik:
@jordalgo @danobi want to have a look once more after the latest changes?

Otherwise I'll go ahead, merge this, and craft the 0.21 release.

jordalgo:
> @jordalgo please check https://github.com/bpftrace/bpftrace/commit/8113d594ef3ce84afb9100173d14a60f3e7e7e55, it's a bit of hacky but I didn't come with another solution which wouldn't make the code too complicated. I used the same approach for the new config option.

This is just to get the location in the code for the error output right? If so, that makes sense and doesn't seem too hacky.

man/adoc/bpftrace.adoc
```diff
@@ -3427,6 +3427,22 @@ Default: 0
...
+=== missing_probes
+
+Default: `warn`
+
+Controls handling of probes with multiple kprobe or uprobe attach points which
+cannot be attached to some functions because they do not exist in the kernel or
+in the traced binary.
+
+The possible options are:
+- `error` - always fail on missing probes,
```
jordalgo, line 3439: nit: don't need commas at the end of these I think or a period at the end of the last one

man/adoc/bpftrace.adoc
```diff
@@ -3427,6 +3427,22 @@ Default: 0
...
+- `warn` - print a warning but continue execution,
+- `ignore` - silently ignore missing probes.
+
+Warning: selecting the option `ignore` may result in probes not to be attached
```
jordalgo, line 3443: meh, I don't think this warning is necessary. I think people know what "ignore" means

jordalgo (commented):
Few minor adoc nits. JC, do you want to make the config an env var as well like the others?

ajor:
> Few minor adoc nits. JC, do you want to make the config an env var as well like the others?

I don't think we need to do that for every new config option. I see env vars as mainly being for backwards compatibility

jordalgo:
> I don't think we need to do that for every new config option.

I will say that the one benefit is that they go into the `--help` output, which is nice - but maybe we should just add the config options instead of env vars to the help output.

man/adoc/bpftrace.adoc
viktormalik: Ah, that's my inner typography geek putting these in.

viktormalik:
> > I don't think we need to do that for every new config option.

Agreed with @ajor here, no need to add more env vars.

> I will say that the one benefit is that they go into the `--help` output, which is nice - but maybe we should just add the config options instead of env vars to the help output.

Putting config options to help instead of env vars sounds good to me. Let's discuss in office hours.

PR #3249, opened 2024-06-17
Support big buf()

danobi:
Similar to big str() and big path(), this PR moves buf() over to new
scratch space infra. Users can now create arbitrary sized buf()s.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -6,6 +6,7 @@
 #include <csignal>
 #include <ctime>
 #include <fstream>
+#include <limits>
 #include <llvm/IR/GlobalValue.h>
 
 #if LLVM_VERSION_MAJOR <= 16
@@ -741,10 +742,10 @@ void CodegenLLVM::visit(Call &call)
     str_id_++;
     expr_ = buf;
   } else if (call.func == "buf") {
-    Value *max_length = b_.getInt64(
-        bpftrace_.config_.get(ConfigKeyInt::max_strlen));
-    size_t fixed_buffer_length = bpftrace_.config_.get(
-        ConfigKeyInt::max_strlen);
+    const uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
+    // Subtract out metadata headroom
+    uint64_t fixed_buffer_length = max_strlen - sizeof(AsyncEvent::Buf);
+    Value *max_length = b_.getInt64(fixed_buffer_length);
     Value *length;
 
     if (call.vargs->size() > 1) {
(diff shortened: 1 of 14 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -745,10 +746,14 @@ void CodegenLLVM::visit(Call &call)
     str_id_++;
     expr_ = buf;
   } else if (call.func == "buf") {
-    Value *max_length = b_.getInt64(
-        bpftrace_.config_.get(ConfigKeyInt::max_strlen));
-    size_t fixed_buffer_length = bpftrace_.config_.get(
-        ConfigKeyInt::max_strlen);
+    uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
+    // Eventually we codegen a u32 store, so ensure we are within limits
+    assert(max_strlen <= std::numeric_limits<uint32_t>::max());
```
jordalgo, line 751: Should this be an actual runtime check with a user error message?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -745,10 +746,14 @@ void CodegenLLVM::visit(Call &call)
...
+
+    // Subtract out metadata headroom
+    uint64_t max_buffer_size = max_strlen - sizeof(AsyncEvent::Buf);
```
jordalgo, line 754: why not just name this variable `fixed_buffer_length` instead of creating a new one below?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -769,16 +774,17 @@ void CodegenLLVM::visit(Call &call)
       auto &arg = *call.vargs->at(0);
       fixed_buffer_length = arg.type.GetNumElements() *
                             arg.type.GetElementTy()->GetSize();
-      length = b_.getInt8(fixed_buffer_length);
+      length = b_.getInt32(fixed_buffer_length);
     }
 
+    Value *scratch_buf = b_.CreateGetStrScratchMap(str_id_, nullptr, call.loc);
```
jordalgo, line 780: nit: why not just do the increment in place in a post op (seems less error prone)

src/ast/passes/semantic_analyser.cpp
```diff
@@ -721,8 +722,10 @@ void SemanticAnalyser::visit(Call &call)
           << typestr(arg.type.GetTy());
     }
 
-    size_t max_buffer_size = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
-    size_t buffer_size = max_buffer_size;
+    // Subtract out metadata headroom
+    uint32_t max_buffer_size = bpftrace_.config_.get(ConfigKeyInt::max_strlen) -
```
jordalgo, line 726: nit just to distinguish these two:
```suggestion
    const uint32_t max_buffer_size = bpftrace_.config_.get(ConfigKeyInt::max_strlen) -
```

jordalgo (commented):
Couple of nits but otherwise LGTM

src/ast/passes/codegen_llvm.cpp
danobi: Yeah, let's move it to semantic analyser

src/ast/passes/codegen_llvm.cpp
danobi: Yeah helper is probably hardest to get wrong later. I'll tack on a commit to do this

danobi: Looking closer, all the other IDs are incremented like this. Probably better to keep consistency. Maybe we can create an X-macro to generate helpers as well as save/restore in create_reset_ids()?

jordalgo: > Looking closer, all the other IDs are incremented like this.

I see what you mean but I can't seem to find any reason why. In some cases the increment happens on the next line.  Anyway, I can look into cleaning that up later. No need to block this PR

jordalgo (approved):

src/ast/passes/codegen_llvm.cpp
danobi: Yeah, no reason why it has to be like this. Would be great if you could take a stab at improving this

Issue #3182, opened 2024-05-17
For-loops do not handle BPF context


ajor:
Latest bpftrace will SEGV when compiling scripts which attempt to access the BPF context from the body of a for-loop:
```
kprobe:do_nanosleep
{
  @map[0] = 1;
  for ($kv : @map) {
    arg0; // requires ctx access
  }
}
```

I think BPF context will have to be passed into the callback subprogram, similarly to variables (#3014).

viktormalik:
#3250 replaces the SEGV by a user-friendly error message.

PR #3250, opened 2024-06-18
Disallow helpers accessing context in for-loops

viktormalik:
Currently, we do not pass BPF context to the callback for the for-loop, so disable builtins which require ctx access. These are those with types marked as `IsCtxAccess` plus the `argX` and `retval` builtins.

This fixes the abort caused by #3182 and replaces it by a nicer error message.

##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -199,6 +199,13 @@ public:
   std::string ident;
   int probe_id;
 
+  // Check if the builtin is 'arg0' - 'arg9'
+  bool is_argx() const
+  {
+    return !ident.compare(0, 3, "arg") && ident.size() == 4 &&
+           ident.at(3) >= '0' && ident.at(3) <= '9';
+  }
+
 private:
   Builtin(const Builtin &other) = default;
 };
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -3757,6 +3757,16 @@ BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }
              false);
 }
 
+TEST(semantic_analyser, for_loop_no_ctx_access)
+{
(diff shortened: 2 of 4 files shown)


src/ast/ast.h
```diff
@@ -199,6 +199,12 @@ class Builtin : public Expression {
   std::string ident;
   int probe_id;
 
+  bool is_argx() const
+  {
+    return !ident.compare(0, 3, "arg") && ident.size() == 4 &&
```
jordalgo, line 204: hmm, this could use a comment as to what all these numbers are - maybe just an example - I know you didn't change this logic but would be good to add.

viktormalik: Good idea, added.

tests/semantic_analyser.cpp
```diff
@@ -3757,6 +3757,16 @@ BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }
              false);
 }
 
+TEST(semantic_analyser, for_loop_no_ctx_access)
+{
+  test_error("kprobe:f { @map[0] = 1; for ($kv : @map) { arg0 } }",
+             R"(
+stdin:1:45-49: ERROR: 'arg0' builtin is not allowed in a for-loop
+kprobe:f { @map[0] = 1; for ($kv : @map) { arg0 } }
+                                            ~~~~
```
jordalgo, line 3766: Is it a known bug that the spacing is off here?

jordalgo (approved):
🚀

tests/semantic_analyser.cpp
viktormalik: Yeah, I noticed, too. I've never realized before so it may be an unknown bug.

jordalgo: Ah here it is: https://github.com/bpftrace/bpftrace/issues/3063

ajor (approved):
Thanks!

PR #3252, opened 2024-06-18
Support printing large arguments

danobi:
This PR adds support for large arguments. Namely: you can now
`printf()` and `print()` big strings/buffers.

With this, we can raise the default on `BPFTRACE_MAX_STRLEN`
from 64 -> 1024 without large regressions. The operative word
is large, as anyone assigning a string to a scratch variable will
get new failures.

To assist with that, we add a new error for anyone trying to assign
a large string to a scratch variable along with a remediation.

**This is considered a regression.** But we have precedent in the past
where we have introduced intentional regressions if it provides a
better overall outcome **and** there is a trivial migration path.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,8 @@ and this project adheres to
   - [#3226](https://github.com/bpftrace/bpftrace/pull/3226)
 - Add config option for handling missing probes
   - [#3246](https://github.com/bpftrace/bpftrace/pull/3246)
+- Support large arguments for printf() and print()
+  - [#3252](https://github.com/bpftrace/bpftrace/pull/3252)
 #### Changed
 - Better error message for args in mixed probes
   - [#3047](https://github.com/bpftrace/bpftrace/pull/3047)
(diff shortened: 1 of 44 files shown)


ajor:
> To assist with that, we add a new error for anyone trying to assign
> a large string to a scratch variable along with a remediation.
> 
> This is considered a regression.

What exactly is the regression here? I wouldn't think that disallowing assignment of only large strings to variables was a change as that didn't work before anyway. Is assignment of small strings being banned too?

danobi:
> > To assist with that, we add a new error for anyone trying to assign
> > a large string to a scratch variable along with a remediation.
> > This is considered a regression.
> 
> What exactly is the regression here? I wouldn't think that disallowing assignment of only large strings to variables was a change as that didn't work before anyway. Is assignment of small strings being banned too?

When `BPFTRACE_MAX_STRLEN` default goes to 1024, users won't be able to do `$s = str(arg0)` anymore. B/c the string buffer is now 1024B.

However, we should still allow something like `$s = str(arg0, 55)`. B/c 55 bytes should fit on stack. I haven't figured out how to teach semantic analyser to clamp string sizes yet. I think it's doable. But that should definitely be part of this PR.

danobi:
I've split out some of the commits in this PR into #3368. I'll deal with changing the default string size in a final PR. It'll include the string clamping.

PR #3254, opened 2024-06-18
Replace usage of bcc_elf_is_*() functions

danielocfb:
In the long term, the project wants to get rid of the `bcc` dependency. As another step towards achieving this goal, replace the usage of `bcc_elf_is_*()` functions with implementations based directly on `libelf`. The `bcc` layer basically doesn't add any value here.

In `resolve_binary_path()`, we benefit additionally from not unnecessarily opening the file in question and reading its ELF header twice, if it does not reference an executable.

##### Checklist

  - N/A
  - not user visible
  - semantic preserving refactoring

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -194,7 +194,7 @@ int BPFtrace::add_probe(const ast::AttachPoint &ap,
   // Preload symbol tables if necessary
   if (resources.probes_using_usym.find(&p) !=
           resources.probes_using_usym.end() &&
-      bcc_elf_is_exe(ap.target.c_str())) {
+      is_exe(ap.target)) {
     auto user_symbol_cache_type = config_.get(
         ConfigKeyUserSymbolCacheType::default_);
     // preload symbol table for executable to make it available even if the
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -13,6 +13,7 @@
 #include <link.h>
 #include <map>
 #include <memory>
+#include <optional>
 #include <regex>
 #include <sstream>
 #include <string>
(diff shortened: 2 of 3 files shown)


danielocfb:
Will finish this up later this week. Though if there are high level comments, feel free to post them.

danobi:
Long term I was hoping AOT binaries wouldn't have to link against libelf. Since it's pretty unavoidable that AOT shim needs to link against utils.o TU, this change would force AOT to link against it. I don't have a solution to offer here. Could always pull this out into another TU. Maybe this is fine as-is and we'll deal with AOT later. Even now AOT shim uses libelf to find the bytecode section inside itself.

danielocfb:
Thanks for your comments.

> At a high level, I'd just say to try to use C++ patterns and standards over C-style code where possible (e.g. [For “out” output values, prefer return values to output parameters](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f20-for-out-output-values-prefer-return-values-to-output-parameters)). I know that's not going to always be possible when directly dealing with a C library like `libelf` though!

Massaged the code a bit.

> I see that BCC also used `access()` for the "is exectuable" check, but I'm wondering if it would be better to use a different check instead. With ahead-of-time compilation bpftrace might not always be running as root during compilation in the future, in which case this check might not be correct. What we want to know is whether something is an executable, not if our user can run it, right?

Sure, we can do that.

> What do you think about these other approaches:
> 
> * Check for executable [filesystem permissions](https://en.cppreference.com/w/cpp/filesystem/perms)
> * Just check if the ELF type is `ET_EXEC`

Changed accordingly. Please take another look when you get a chance.

danielocfb:
The
> uprobe.uprobes - attach to probe for executable in separate mount namespace

failure seems legit. Looking at it.

danielocfb:
Any suggestions for the `Uncontrolled data used in path expression` lint? I don't think there is a semantic change here, more that previously GitHub didn't have any insight into `bcc` code.

ajor (approved):
Looks good to me!

I just dismissed the lint - we actually want to be opening a user-provided file here.

Issue #3255, opened 2024-06-19
Unable to run runqlat.bt


FlakM:
Running `sudo runqlat.bt` on NixOs system fails with following error:

```
❯ sudo runqlat.bt
WARNING: Could not find kernel headers in  or . To specify a particular path to kernel headers, set the env variables BPFTRACE_KERNEL_SOURCE and, optionally, BPFTRACE_KERNEL_BUILD if the kernel was built in a different directory than its source. To create kernel headers run 'modprobe kheaders', which will create a tar file at /sys/kernel/kheaders.tar.xz
definitions.h:2:10: fatal error: 'linux/sched.h' file not found
```

Note that on the same machine it is possible to run `runqlat` from `bcc`

![image](https://github.com/bpftrace/bpftrace/assets/4190088/53ea4d8a-2d19-4060-8a1d-f97b1fd888a3)

### What reproduces the bug? Provide code if possible.

This [repository](https://github.com/FlakM/bpftrace_repro_runqlat) contains:
- instruction how to create VM to reproduce (requires nix - not NixOs, any distribution will do)
- automated test that reproduces the issue

### `bpftrace --info` output

```
❯ sudo bpftrace --info
System
  OS: Linux 6.6.33 #1-NixOS SMP PREEMPT_DYNAMIC Wed Jun 12 09:13:03 UTC 2024
  Arch: x86_64

Build
  version: v0.20.4
  LLVM: 17.0.6
  unsafe probe: no
  bfd: yes
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: no
[...]

PR #3262, opened 2024-06-21
tools/runqlat: provide TASK_RUNNING as a define

martinetd:
runqlat requires kernel headers to run even with BTF, just because of a define.
TASK_RUNNING isn't part of the stable API but it's never changed in all of the linux git history so let's pretend it's stable and just define it.

If we find a way to handle kheaders again in the future we might want to consider reverting this.

Fixes: #3255

##### Checklist

^ once again no test for this as far as I know, but tested manually.

### DIFF
--- a/tools/runqlat.bt
+++ b/tools/runqlat.bt
@@ -11,7 +11,17 @@
  * 17-Sep-2018	Brendan Gregg	Created this.
  */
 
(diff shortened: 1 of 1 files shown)


ajor (approved):
Thanks!

PR #3263, opened 2024-06-21
Refactor program loading

viktormalik:
This is the last refactoring step of #2334 before we move to libbpf-based program loading. I'm posting this separately to simplify reviewing.

This does a couple of things:
- create instances of `BpfProgram` by iterating `bpf_program` objects in `bpf_object` (instead of searching for individual sections in the ELF),
- move program loading to `BpfProgram`,
- load all BPF programs at one (from `BpfBytecode`).

See commit messages in the individual commits for more implementation details.

This should contain no functional change. The only thing is that it removes progfd caching but that will go away anyways since program fds will be managed by libbpf.

##### Checklist


jordalgo:
Looks like these two tests are broken as a result:
[  FAILED  ] probe.kprobe_wildcard
[  FAILED  ] probe.kretprobe_wildcard

src/bpfbytecode.cpp
```diff
@@ -22,6 +23,14 @@ BpfBytecode::BpfBytecode(const void *elf, size_t elf_size)
     auto map = BpfMap(m);
     maps_.emplace(name, map);
   }
+
+  struct bpf_program *p;
+  bpf_object__for_each_program(p, bpf_object_.get())
+  {
+    std::string name = bpf_program__name(p);
+    auto prog = BpfProgram(p);
+    programs_.emplace(name, std::move(prog));
```
jordalgo, line 32: nit: I think this should construct BpfProgram in-place `programs_.emplace(name, p);`.

src/bpfbytecode.cpp
```diff
@@ -44,6 +53,34 @@ const std::vector<uint8_t> &BpfBytecode::getSection(
...
 
+BpfProgram &BpfBytecode::getProgramForProbe(const Probe &probe)
+{
+  auto usdt_location_idx = (probe.type == ProbeType::usdt)
+                               ? std::make_optional<int>(
+                                     probe.usdt_location_idx)
+                               : std::nullopt;
+
+  auto prog = programs_.find(
+      get_function_name_for_probe(probe.name, probe.index, usdt_location_idx));
```
jordalgo, line 64: maybe for a follow-up diff but it feels like `get_function_name_for_probe` should be handling the usdt logic

src/bpfprogram.cpp
```diff
@@ -243,7 +232,7 @@ void BpfProgram::relocateFuncInfos()
         btfstr + info_sec->sec_name_off);
     if (text_offset_ > 0 && name == ".text")
       text_funcinfo_sec = info_sec;
-    else if (name == name_)
+    else if (name == bpf_program__section_name(bpf_prog_))
```
jordalgo, line 242: should we just store the section name as a member variable? I see `bpf_program__section_name` called from multiple methods.

viktormalik: It's stored as a member of `bpf_prog_` so the only thing this would get us is few chars on a shorter accessor :-) I don't think it's worth it.

src/bpfbytecode.cpp
viktormalik: Yeah, this is a bit messy but I'd rather refactor it separately.

viktormalik:
> Looks like these two tests are broken as a result:  
[ FAILED ] probe.kprobe_wildcard  
[ FAILED ] probe.kretprobe_wildcard

Right, that's because I dropped the progfd caching. The tests will pass on systems with kprobe_multi but not on systems without it. Since the follow-up PR will move to libbpf-based loading, there's no point of reintroducing progfd caching now so I'll just add `REQUIRES_FEATURE kprobe_multi` to these tests.

src/bpfprogram.cpp
jordalgo: kk, no worries

jordalgo:
>  The tests will pass on systems with kprobe_multi but not on systems without it. Since the follow-up PR will move to libbpf-based loading, there's no point of reintroducing progfd caching now so I'll just add REQUIRES_FEATURE kprobe_multi to these tests.

Ok, so just to be clear that we're not losing kprobe wildcard probes for kernels without kprobe-multi because they'll be fixed in your PR for libbpf-based loading?

jordalgo (approved):
One small clarifying question but code looks good.

viktormalik:
> > The tests will pass on systems with kprobe_multi but not on systems without it. Since the follow-up PR will move to libbpf-based loading, there's no point of reintroducing progfd caching now so I'll just add REQUIRES_FEATURE kprobe_multi to these tests.
> 
> Ok, so just to be clear that we're not losing kprobe wildcard probes for kernels without kprobe-multi because they'll be fixed in your PR for libbpf-based loading?

We're not loosing them at all, we're just not caching progfds for them so each match will get a separately loaded BPF program. It's done this way now and will not change with the move to libbpf.

IIUC, progfd caching was introduced as a performance optimization and was later replaced by kprobe_multi. Ofc it makes kprobe wildcard probes more efficient on systems without kprobe_multi (which is something that we lose) but we'll need to figure this out differently since we'll have less control over the loading.

jordalgo:
@viktormalik Thanks for the clarification. One more stupid question: if we're not losing kprobes with wildcards (in this PR) then why do a lot of the kprobe wildcard runtime tests require the kprobe-multi feature?

viktormalik:
> @viktormalik Thanks for the clarification. One more stupid question: if we're not losing kprobes with wildcards (in this PR) then why do a lot of the kprobe wildcard runtime tests require the kprobe-multi feature?

If I'm looking correctly, it's just three tests. Two of them affected by this PR (`kprobe_multi_wildcard` and `kretprobe_multi_wildcard`) check that only one BPF program is created (which is now only true for kprobe_multi) and the last one (`kprobe_wildcard_all_leading_underscore`) checks that we can actually attach to a lot of functions using kprobe_multi (and do not attach to functions causing instability).

viktormalik:
> Ok, well then I think you can guess where I'm going with this 😄 - Should we add a kprobe wildcard test that doesn't rely on kprobe_multi to make sure we don't break backward compat.

Now I do :grinning:. Makes sense, I added some new tests to make sure that wildcarded k(ret)probes work in every situation.

tests/runtime/probe
```diff
@@ -118,10 +118,17 @@ TIMEOUT 5
 AFTER ./testprogs/syscall read
 
 NAME kprobe_wildcard
+PROG kprobe:vmlinux:vfs_rea* { printf("SUCCESS %d\n", pid); exit(); }
```
jordalgo, line 121: Awesome. Thanks for adding!

src/bpfprogram.cpp
```diff
@@ -290,4 +286,280 @@ void BpfProgram::appendFileFuncInfos(const struct btf_ext_info_sec *src,
...
+        close(opts.attach_btf_obj_fd);
+      if (fd_ >= 0)
+        break;
+      logbuf << " = " << fd_ << ": " << std::strerror(-fd_);
+    log_error:
+      if (!logbuf.str().empty()) { // log the error of the failed attempt
+        LOG(V1) << logbuf.str();
+        logbuf.str("");
+      }
+    } // obsolete indendation level, has no effect, kept for a small commitdiff
```
ajor, line 512: I don't think we need to keep this any more

src/bpfprogram.h
```diff
@@ -10,37 +13,41 @@
...
+  void load(const Probe &probe,
+            const BpfBytecode &bytecode,
+            const BTF &btf,
+            BPFfeature &feature);
 
-  const std::vector<uint8_t> &getCode();
-  const std::vector<uint8_t> &getBTF();
-  const std::vector<uint8_t> &getFuncInfos();
+  const std::vector<uint8_t> &getCode() const;
+  const std::vector<uint8_t> &getFuncInfos() const;
```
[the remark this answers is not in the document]
viktormalik: Sure, although they'll be removed in the following PR.

src/bpfprogram.h
```diff
@@ -10,37 +13,41 @@
 
 namespace bpftrace {
 
+class BpfBytecode;
 class BPFtrace;
 
 class BpfProgram {
```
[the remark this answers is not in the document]
viktormalik: Good idea, added.

src/bpfbytecode.cpp
```diff
@@ -44,6 +49,54 @@ const std::vector<uint8_t> &BpfBytecode::getSection(
   return sections_.at(name);
 }
 
+BpfProgram &BpfBytecode::getProgramForProbe(const Probe &probe)
```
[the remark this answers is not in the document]
viktormalik: I think that it requires a bit of `const_cast`-magic:
```
const BpfProgram &BpfBytecode::getProgramForProbe(const Probe &probe) const
{
   ... original code ...
}

BpfProgram &BpfBytecode::getProgramForProbe(const Probe &probe)
{
  return const_cast<BpfProgram &>(
      const_cast<const BpfBytecode *>(this)->getProgramForProbe(probe));
}
```
but since the non-const version will go away soon, I'm ok with doing this as it makes clear that `AttachedProbe` is not changing `BpfProgram`.

src/bpfbytecode.cpp
```diff
@@ -44,6 +49,54 @@ const std::vector<uint8_t> &BpfBytecode::getSection(
...
+}
+
+void BpfBytecode::load_progs(const std::vector<Probe> &probes,
+                             BTF &btf,
+                             BPFfeature &feature)
+{
+  for (auto &probe : probes) {
+    auto &program = getProgramForProbe(probe);
+    program.assemble(*this);
+    program.load(probe, *this, btf, feature);
```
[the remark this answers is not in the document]
viktormalik: Yeah, it could be viewed that way but since both are going away in the next PR, I wouldn't bother refactoring it now.

src/bpfprogram.cpp
viktormalik: Will also go away soon but sure, let's drop it.

src/bpfbytecode.cpp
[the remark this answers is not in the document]
ajor: But if the non-const one is going away soon then I guess it's not important

ajor (approved):
Looks good!

jordalgo (approved):
👍

PR #3264, opened 2024-06-21
Output: Stream output when printing maps

### LINKED ISSUE #3251: Stream output when printing maps
Symbolicating user stack traces can be slow for large binaries and where symbol caching is not effective. This script should collect a large number of different user stack traces:
```
profile:hz:99
{
  @[ustack] = count();
}
```

After hitting Ctrl-C, bpftrace can appear to "hang" while it's stuck symbolicating all the stack traces. We then print the stacks out in a single batch once they've all been symbolicated: https://github.com/bpftrace/bpftrace/blob/b0a326a2c44e73ff0346a5fcc8b7d258faeb239b/src/output.cpp#L407

We should instead print out each stack trace as soon as it's been symbolicated, to at least show progress to the user (speeding up symbolication is a separate issue).

ajor:
Symbolicating user stack traces can be slow for large binaries and when symbol caching is not effective. When printing a large map containing stack traces, bpftrace can appear to "hang" while it's stuck symbolicating all of them, before it prints them in a single batch.

This commit changes the behaviour so that we print each stack trace as soon as it's ready, to show some progress to the user.

Fixes #3251

##### Checklist


src/output.cpp
```diff
@@ -419,17 +423,22 @@ std::string Output::map_hist_to_str(
...
     auto &key = key_count.first;
     auto &value = values_by_key.at(key);
 
     if (top && values_by_key.size() > top && i++ < (values_by_key.size() - top))
       continue;
 
+    if (first)
+      first = false;
+    else
+      map_elem_delim(map_type);
```
jordalgo, line 440: nit: can we just move this to the end of this loop and then call it again after the loop finishes? seems less ugly than keeping this `first` state var

src/output.cpp
```diff
@@ -675,24 +685,21 @@ std::string TextOutput::map_key_to_str(BPFtrace &bpftrace,
   return map.name() + map_key.argument_value_list_str(bpftrace, key);
 }
 
-std::string TextOutput::map_keyval_to_str(const SizedType &map_type,
-                                          const std::string &key,
-                                          const std::string &val) const
+void TextOutput::map_keyval(const SizedType &map_type,
```
jordalgo, line 688: naming nit to match the other snake case funcs 
```suggestion
void TextOutput::map_key_val(const SizedType &map_type,
```

jordalgo (approved):
Few small non-blocking nits but LGTM. Nice improvement!

src/output.cpp
ajor: Sorry, I'm not understanding how that'd work. We don't want to print a trailing delimiter, so we need to emit fewer delimiters than map elements - how can we do that without a variable?

viktormalik (approved):

PR #3265, opened 2024-06-22
kernel headers: only print kheaders not found warning if parsing failed

martinetd:
Current code would print kheaders not found as soon as the code has any include, even if the include worked.

This delays printing the warning until we know if parsing succeeded or not, and only prints it if parsing failed.

Also update the message to give clearer extraction instructions

(This was an annoying message on nixos that I had just shoved under `-v`-only output, but this is probably more appropriate: if parsing worked, we don't care that kernel headers weren't found.)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to
 
 #### Added
 #### Changed
+- Only print kernel headers not found message if parsing failed
```
jordalgo, line 13: nit: 
```suggestion
- Only print kernel headers not found message if parsing fails
```

src/main.cpp
```diff
@@ -408,8 +409,21 @@ static void parse_env(BPFtrace& bpftrace)
       extra_flags.push_back(file);
     }
 
-    if (!clang.parse(driver.root.get(), bpftrace, extra_flags))
+    if (!clang.parse(driver.root.get(), bpftrace, extra_flags)) {
+      if (!found_kernel_headers) {
+        LOG(WARNING)
+            << "Could not find kernel headers in /lib/modules/"
```
jordalgo, line 415: Nit: Maybe make a small utility function so we don't have to create this path twice e.g.
```
std::string get_default_kdir(const struct utsname &utsname)
{
  return std::string("/lib/modules/") + utsname.release
}
```

Then just pass that into `get_kernel_dirs`

jordalgo (commented):
Thanks for the change. Few small nits but otherwise LGTM

src/main.cpp
martinetd: I considered that and am not sure... Just `/lib/modules/(uname -r)` isn't enough without source/build - nixos in particular has a `linuxHeaders` package that has no relation to the modules path and doesn't have a `build` or `source` component, if we want to override the directory at some point I think sharing now will mostly be undone so it's not really worth it
(unfortunately, nixpkgs' linuxHeaders doesn't have a kconfig.h as it's not directly related to the current kernel build, so it wasn't usable as is when I tried and I stopped looking this direction)

OTOH I agree it's a bit of a shame to have this twice, if you really prefer having such a helper I'll fix this tomorrow

martinetd: uh, actually I now (re-)notice KERNEL_HEADERS_DIR, so part of my message doesn't make sense (it's already overridable) -- however if that's set then get_kernel_dirs() cannot fail so this message would never be printed and the inconsistency doesn't really matter either.
It might make sense to move KERNEL_HEADERS_DIR in a new get default path helper and check if it exists though? In which case I guess the helper makes more sense, do we want to check if the hardcoded path exists and allows env vars to work with it?

martinetd:
Another thing I noticed after testing a bit more just now: this works well for things like header not found or header not parsed properly (e.g. loading glibc's linux/sched.h instead of linux's), but it doesn't work when the header loader but a define is missing e.g. this will not print the message:
```
$ strace -e write -k bpftrace -e '#include <errno.h>
BEGIN { printf("ok %d\n", BITS_PER_BYTE); exit(); }'

stdin:3:27-40: ERROR: Unknown identifier: 'BITS_PER_BYTE'
BEGIN { printf("ok %d\n", BITS_PER_BYTE); exit(); }

 > /nix/store/cil1qrv5vddpxmg2r34364014ba5phxz-bpftrace-0.20.4/bin/bpftrace(bpftrace::ast::SemanticAnalyser::analyse()+0x204) [0x5c0fb4]
 > /nix/store/cil1qrv5vddpxmg2r34364014ba5phxz-bpftrace-0.20.4/bin/bpftrace(bpftrace::ast::CreateSemanticPass()::{lambda(bpftrace::ast::Node&, bpftrace::ast::PassContext&)#1}::operator()(bpftrace::ast::Node&, bpftrace::ast::PassContext&) const [clone .isra.0]+0x2cb) [0x5c2c3b]
 > /nix/store/cil1qrv5vddpxmg2r34364014ba5phxz-bpftrace-0.20.4/bin/bpftrace(std::_Function_handler<bpftrace::ast::PassResult (bpftrace::ast::Node&, bpftrace::ast::PassContext&), bpftrace::ast::CreateSemanticPass()::{lambda(bpftrace::ast::Node&, bpftrace::ast::PassContext&)#1}>::_M_invoke(std::_Any_data const&, bpftrace::ast::Node&, bpftrace::ast::PassContext&)+0x22) [0x5c2f32]
 > /nix/store/cil1qrv5vddpxmg2r34364014ba5phxz-bpftrace-0.20.4/bin/bpftrace(bpftrace::ast::PassManager::Run(std::unique_ptr<bpftrace::ast::Node, std::default_delete<bpftrace::ast::Node> >, bpftrace::ast::PassContext&)+0x87) [0x5a1a47]
 > /nix/store/cil1qrv5vddpxmg2r34364014ba5phxz-bpftrace-0.20.4/bin/bpftrace(main+0xfd9) [0x454849]
 > /nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libc.so.6(__libc_start_call_main+0x7d) [0x2a10d]
 > /nix/store/k7zgvzp2r31zkg9xqgjim7mbknryv6bs-glibc-2.39-52/lib/libc.so.6(__libc_start_main@@GLIBC_2.34+0x88) [0x2a1c8]
 > /nix/store/cil1qrv5vddpxmg2r34364014ba5phxz-bpftrace-0.20.4/bin/bpftrace(_start+0x24) [0x4b8524]
```

Since it happens in Run() it's hard to keep track of whether the linux headers addition worked or not...
I'm not sure if it's worth the effort as I don't think that's a very common failure pattern, but there might be other failure cases I can't think of right now where it'd be bad not to print we didn't find the header -- perhaps keep another message when it failed only if -v (bt_verbose global) to help debugging?

src/main.cpp
jordalgo: Ah yeah, I see what you're saying. How about this, we change the signature to something like:
```
bool get_kernel_dirs(
    const struct utsname &utsname, std::string& ksrc, std::string kobj)
```

The return bool will indicate if we failed to find the default kernel headers that you can then check if the parsing fails.

jordalgo:
> I can't think of right now where it'd be bad not to print we didn't find the header -- perhaps keep another message when it failed only if -v (bt_verbose global) to help debugging?

Yeah, this does seems a bit edge case and even the error you got says to me that `BITS_PER_BYTE` is not defined in the header that was manually included but feel free to propose a change if you have something that can improve this.

src/main.cpp
martinetd: This prototype makes sense, updated.
I've kept the cmake define first and without check to not change behaviour, but removed the #ifdef to at least make sure it builds

martinetd:
> Yeah, this does seems a bit edge case and even the error you got says to me that BITS_PER_BYTE is not defined in the header that was manually included but feel free to propose a change if you have something that can improve this.

Thanks; let's keep it that way for now instead of overcomplicating things; we can always come back to this later if there is a better idea.

src/main.cpp
```diff
@@ -382,20 +382,16 @@ static void parse_env(BPFtrace& bpftrace)
...
-      ksrc = std::get<0>(kdirs);
-      kobj = std::get<1>(kdirs);
-
-      if (ksrc != "") {
-        extra_flags = get_kernel_cflags(
-            utsname.machine, ksrc, kobj, bpftrace.kconfig);
-      }
-    }
+    uname(&utsname);
+    found_kernel_headers = get_kernel_dirs(utsname, ksrc, kobj);
```
jordalgo, line 390: nit
```suggestion
    bool found_kernel_headers = get_kernel_dirs(utsname, ksrc, kobj);
```

src/main.cpp
```diff
@@ -382,20 +382,16 @@ static void parse_env(BPFtrace& bpftrace)
 
   if (should_clang_parse) {
     ClangParser clang;
+    bool found_kernel_headers;
```
jordalgo, line 385: rm - see below
```suggestion
```

jordalgo (approved):
LGTM. One small nit.

Thanks for the update!

src/main.cpp
martinetd: Thanks, applied!

jordalgo (approved):

PR #3270, opened 2024-06-24
Fix some C++ warnings found by clang

hhoffstaette:
While packaging bpftrace-0.21.0 for Gentoo I did a test build with clang-18 and was greeted by a wall of warnings. The commits in this PR cut down the noise significantly.
There are more warnings (mostly the terrible LIBBPF_OPTS), some of which are already being addressed (e.g. [here](https://github.com/bpftrace/bpftrace/pull/3203)), so this is just a quick fix for low-hanging fruit.

No functional or user-visible changes.
The resulting binary also still works. :smile: 

##### Checklist


hhoffstaette:
man.. clang-format :sob:

src/ast/passes/codegen_llvm.cpp
```diff
@@ -986,9 +987,9 @@ void CodegenLLVM::visit(Call &call)
       buf = b_.CreateAllocaBPF(array_t, "addr4");
     }
 
-    char dst[addr_size];
+    char *dst = (char *)alloca(addr_size);
```
ajor, line 990:
```suggestion
    std::vector<uint8_t> dst(addr_size);
```

`alloca` is a very C-style way of doing things. I think the best way in C++ would be to just create a vector of the desired length.

(same comment applies to the other instances)

src/types.h
```diff
@@ -550,26 +550,54 @@ struct ProbeItem {
...
-  { "kfunc", { "f", "fentry" }, ProbeType::kfunc, .show_in_kernel_list = true },
-  { "kretfunc", { "fr", "fexit" }, ProbeType::kretfunc },
-  { "iter", { "it" }, ProbeType::iter, .show_in_kernel_list = true },
-  { "rawtracepoint",
-    { "rt" },
-    ProbeType::rawtracepoint,
+  { .name = "kprobe",
+    .aliases = { "k" },
+    .type = ProbeType::kprobe,
+    .show_in_kernel_list = true },
```
ajor, line 556: I'm guessing the problem here was that we either have to use named assignment for every field or for no fields?

I'm not sure why we have `show_in_kernel_list` and `show_in_userspace_list` as separate booleans really - it might be nicer to create an enum type that defines what lists to show a probe in, e.g.:
```
enum class ShowInList {
  None,
  Kernel,
  Userspace,
};
```

That way every probe entry can have an explicit listing definition without us having to name every field, e.g.:
```
{ "kprobe", { "k" }, ProbeType::kprobe, ShowInList::Kernel },
```

This might end up being a bigger change than I'm anticipating, so I won't block on this suggestion.

ajor (commented):
Thanks for looking into this!

src/ast/passes/codegen_llvm.cpp
hhoffstaette: I know & thought about that; eventually I ended up using alloca because I decided to stay as close to the original semantic as possible. If you want me to rewrite things with vector (which will imply a heap allocation) I'm fine with that; I don't think any of these things are on super-performance-critical paths.

src/types.h
hhoffstaette: Yes, the problem is that you need all-or-nothing in initializer lists, and it makes sense - you can use order or .tags but not really both.

src/ast/passes/codegen_llvm.cpp
ajor: Yeah, this code is just run during compilation so isn't performance-critical (although performance is nice!). Tbh I hadn't read this code too carefully before posting my comment. It was more of an instinctive reaction to seeing manual memory allocation which I'm trying to stamp out :) 

I suppose `char[16]` would also work since we know the maximum size at compile time, although I'm not sure the performance is worth the increased risks of bugs in the future from decoupling the array length from the `addr_size` variable.

src/bpftrace.cpp
```diff
@@ -1754,8 +1755,9 @@ std::string BPFtrace::resolve_timestamp(uint32_t mode,
   snprintf(usecs_buf, sizeof(usecs_buf), "%06" PRIu64, us);
   auto fmt = std::regex_replace(raw_fmt, usec_regex, usecs_buf);
 
-  char timestr[config_.get(ConfigKeyInt::max_strlen)];
-  if (strftime(timestr, sizeof(timestr), fmt.c_str(), &tmp) == 0) {
+  uint64_t timestr_size = config_.get(ConfigKeyInt::max_strlen);
+  char *timestr = (char *)alloca(timestr_size);
```
ajor, line 1759: This one I would definitely put on the heap. `max_strlen` is a user-supplied number so could potentially be pretty big. We don't have any limits on what it can be set to either... I didn't know it was being used like this actually.

`std::vector` will at least catch the out-of-memory condition and throw an error.

hhoffstaette: Already on it. :)

src/ast/passes/codegen_llvm.cpp
hhoffstaette: Replaced alloca with vectors.

src/bpftrace.cpp
hhoffstaette: Also done in the latest commit.

src/utils.cpp
```diff
@@ -1071,7 +1071,8 @@ std::string hex_format_buffer(const char *buf,
                               bool escape_hex)
 {
   // Allow enough space for every byte to be sanitized in the form "\x00"
-  char s[size * 4 + 1];
+  std::vector<char> str(size * 4 + 1);
+  char *s = str.data();
```
ajor, line 1075:
```suggestion
  std::string s(size * 4 + 1, '\0');
```

Can we just make this one a std::string from the beginning? Then there's no need to copy all the data again at the end

src/bpftrace.cpp
```diff
@@ -1754,12 +1754,13 @@ std::string BPFtrace::resolve_timestamp(uint32_t mode,
   snprintf(usecs_buf, sizeof(usecs_buf), "%06" PRIu64, us);
   auto fmt = std::regex_replace(raw_fmt, usec_regex, usecs_buf);
 
-  char timestr[config_.get(ConfigKeyInt::max_strlen)];
-  if (strftime(timestr, sizeof(timestr), fmt.c_str(), &tmp) == 0) {
+  uint64_t timestr_size = config_.get(ConfigKeyInt::max_strlen);
+  std::vector<char> timestr(timestr_size);
```
ajor, line 1758:
```suggestion
  std::string timestr(timestr_size, '\0');
```
Same. If we create it as a string then we don't have to copy it at the end

ajor (commented):
Just one last nit

src/utils.cpp
hhoffstaette: I had the same thought (no, really) and just wanted to make sure you are paying attention. :cat: 
Fixed in the latest commit.

hhoffstaette:
Finally! :zany_face:

ajor (approved):
Thanks for sorting these warnings out!

I just made a small change to replace `string.substr()` with `string.resize()`, as `substr` creates a copy of the original.

PR #3275, opened 2024-06-25
Standardize async id creation and setting

jordalgo:
This will make sure we don't forget to reset async ids in 'create_reset_ids' and also makes sure we're incrementing the id whenever we get it (instead of doing it at some later, random instruction).

This is follow up to a conversation here:
https://github.com/bpftrace/bpftrace/pull/3249#discussion_r1649215402

##### Checklist

### DIFF
--- a/src/ast/async_ids.h
+++ b/src/ast/async_ids.h
@@ -1,0 +1,54 @@
+#pragma once
+
+namespace bpftrace {
+namespace ast {
+
+// Add new ids here
+#define FOR_LIST_OF_ASYNC_IDS(DO)                                              \
+  DO(cat)                                                                      \
+  DO(cgroup_path)                                                              \
+  DO(helper_error)                                                             \
+  DO(join)                                                                     \
+  DO(mapped_printf)                                                            \
+  DO(non_map_print)                                                            \
+  DO(printf)                                                                   \
+  DO(skb_output)                                                               \
+  DO(strftime)                                                                 \
+  DO(str)                                                                      \
+  DO(system)                                                                   \
+  DO(time)                                                                     \
+  DO(watchpoint)
+
+#define DEFINE_MEMBER_VAR(id, ...) int _##id = 0;
+#define DEFINE_ACCESS_METHOD(id, ...)                                          \
+  int id()                                                                     \
+  {                                                                            \
+    return _##id++;                                                            \
+  }
+#define LOCAL_SAVE(id, ...) local_##id = _##id,
+#define LOCAL_RESTORE(id, ...) this->_##id = local_##id;
+
+class AsyncIds {
+public:
+  explicit AsyncIds() = default;
+
+  FOR_LIST_OF_ASYNC_IDS(DEFINE_ACCESS_METHOD)
+
+  /*
+   * For 'create_reset_ids' return a lambda that has captured-by-value
+   * CodegenLLVM's async id state. Running the returned lambda will restore
+   * `CodegenLLVM`s async id state back to when this function was first called.
+   */
+  std::function<void()> create_reset_ids()
+  {
+    return [FOR_LIST_OF_ASYNC_IDS(LOCAL_SAVE) this] {
+      FOR_LIST_OF_ASYNC_IDS(LOCAL_RESTORE)
+    };
+  }
+
+private:
+  FOR_LIST_OF_ASYNC_IDS(DEFINE_MEMBER_VAR)
+};
+
+} // namespace ast
+} // namespace bpftrace
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -62,7 +62,8 @@ CodegenLLVM::CodegenLLVM(Node *root,
       usdt_helper_(std::move(usdt_helper)),
       context_(std::make_unique<LLVMContext>()),
       module_(std::make_unique<Module>("bpftrace", *context_)),
-      b_(*context_, *module_, bpftrace),
+      async_ids_(AsyncIds()),
+      b_(*context_, *module_, bpftrace, async_ids_),
       debug_(*module_),
       is_aot_(is_aot)
 {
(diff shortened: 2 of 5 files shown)


jordalgo:
@danobi

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1058,18 +1057,17 @@ void CodegenLLVM::visit(Call &call)
                          b_.CreatePointerCast(data, b_.GET_PTR_TY()),
                          b_.getInt32(data_size),
                          call.loc);
-
-      mapped_printf_id_++;
     } else {
       createFormatStringCall(call,
-                             printf_id_,
+                             get_printf_id_(),
```
jordalgo, line 1062: looks like this was a bug as we weren't incrementing `printf_id_` anywhere

src/ast/passes/codegen_llvm.h
```diff
@@ -280,19 +274,74 @@ class CodegenLLVM : public Visitor {
...
+  int get_##ID()                                                               \
+  {                                                                            \
+    return ID++;                                                               \
+  }
+
+// For 'create_reset_ids' return a lambda that has captured-by-value
+// CodegenLLVM's async id state. Running the returned lambda will restore
+// `CodegenLLVM`s async id state back to when this function was first called.
+#define CREATE_ASYNC_IDS(                                                      \
+    ID1, ID2, ID3, ID4, ID5, ID6, ID7, ID8, ID9, ID10, ID11, ID12)             \
```
jordalgo, line 288: I think there is a better way to do this with boost but not sure it's worth adding.

src/ast/passes/codegen_llvm.h
```diff
@@ -280,19 +274,74 @@ class CodegenLLVM : public Visitor {
...
-  int strftime_id_ = 0;
-  uint64_t join_id_ = 0;
-  int system_id_ = 0;
-  int non_map_print_id_ = 0;
-  uint64_t watchpoint_id_ = 0;
-  int cgroup_path_id_ = 0;
-  int skb_output_id_ = 0;
-  int str_id_ = 0;
+#define ADD_ASYNC_ID(ID)                                                       \
+  int ID = 0;                                                                  \
```
danobi, line 278: Can we mangle this variable name so that direct access is more obvious during code review? Maybe mangle it like `baddirectaccess_ # ID`?

src/ast/passes/codegen_llvm.h
danobi: I think it might be cleaner to use an x-macro: https://en.wikipedia.org/wiki/X_macro 

New IDs can be added with a single line

jordalgo:
@danobi Great suggestion on using X-macros; I didn't know about them. Code looks much cleaner.

src/ast/passes/codegen_llvm.h
```diff
@@ -280,19 +274,43 @@ class CodegenLLVM : public Visitor {
...
-  int cat_id_ = 0;
-  int strftime_id_ = 0;
-  uint64_t join_id_ = 0;
-  int system_id_ = 0;
-  int non_map_print_id_ = 0;
-  uint64_t watchpoint_id_ = 0;
-  int cgroup_path_id_ = 0;
-  int skb_output_id_ = 0;
-  int str_id_ = 0;
+#define FOR_LIST_OF_ASYNC_IDS(DO)                                              \
```
danobi, line 277: IMHO kinda ugly to have all these macros defined inline. Is there a way to move the macro magic out into a separate file / class? Then in here would could store just a `AsyncIds async_ids_` member. Callers could do `async_ids_.printf()` or something like that.

jordalgo: Yeah I thought about a separate class too. I think I just got lazy 😄  - I'll update the PR

src/ast/async_ids.h
```diff
@@ -0,0 +1,57 @@
...
+  int id()                                                                     \
+  {                                                                            \
+    return _##id++;                                                            \
+  }
+#define LOCAL_SAVE(id, ...) local_##id = _##id,
+#define LOCAL_RESTORE(id, ...) this->_##id = local_##id;
+
+class AsyncIds {
+public:
+  explicit AsyncIds(){};
```
danobi, line 33:
```suggestion
  explicit AsyncIds() = default;
```

src/ast/async_ids.h
```diff
@@ -0,0 +1,57 @@
...
+  std::function<void()> create_reset_ids()
+  {
+    return [this, FOR_LIST_OF_ASYNC_IDS(LOCAL_SAVE) local_dummy = dummy_id] {
+      FOR_LIST_OF_ASYNC_IDS(LOCAL_RESTORE)
+      this->dummy_id = local_dummy;
+    };
+  }
+
+private:
+  // Exists so we don't have a trailing comma in the lambda capture list above.
```
danobi, line 51: Is this comment still accurate? Not sure I follow

src/ast/async_ids.h
```diff
@@ -0,0 +1,57 @@
...
+  int dummy_id = 0;
+  FOR_LIST_OF_ASYNC_IDS(DEFINE_MEMBER_VAR)
+};
+
+} // namespace ast
+} // namespace bpftrace
```
danobi, line 57: Is your editor respecting the `.editorconfig` at root of repo? The lack of newline will make other changes add it

src/ast/async_ids.h
jordalgo: Nevermind. I'm gonna remove this comment and the `dummy_id`. it was added because I didn't realize `this` could be last in the lambda capture list and I needed something that didn't have a trailing comma.

src/ast/async_ids.h
jordalgo: It's weird. `.editorconfig` gets applied for the local checkout of bpftrace on my laptop but when I'm editing files on my devserver checkout 😕

danobi (approved):

jordalgo:
@danobi Thanks for the review!

jordalgo:
@viktormalik Let me know if you want me to hold off on landing this as it might cause a few merge conflicts with https://github.com/bpftrace/bpftrace/pull/3285

viktormalik (approved):
> @viktormalik Let me know if you want me to hold off on landing this as it might cause a few merge conflicts with #3285

Np, I think that it should be pretty straightforward to resolve them.

I really like this, let's get it in!

Issue #3276, opened 2024-06-26
Addressing LIBBPF_OPTS vs. clang warnings vs. C++ struct initialisation


hhoffstaette:
As continuation of https://github.com/bpftrace/bpftrace/commit/65a08f4c38505d327862d52404be85469168966a the following is a request for ideas on how to best fix the C++ warnings caused by libbpf's LIBBPF_OPTS macro. This turned out to be a surprisingly deep swamp of C/C++/compiler-specific details, so sorry for the long-winded explanation.
I'm willing to make the PR but for now am just looking for the least bad option. :)

First is the inconsistent use of `DECLARE_LIBBPF_OPTS` vs. `LIBBPF_OPTS`. The former is a deprecated value and simply #defined to `LIBBPF_OPTS`, so replacing those few instances is trivial.

Second, the problem itself:
```
/tmp/bpftrace/src/bpffeature.cpp:62:5: warning: missing field 'attempts' initializer [-Wmissing-field-initializers]
   62 |     LIBBPF_OPTS(bpf_prog_load_opts, opts);
      |     ^
/usr/include/bpf/libbpf_common.h:70:3: note: expanded from macro 'LIBBPF_OPTS'
   70 |                 };                                                          \
      |                 ^
/tmp/bpftrace/src/bpffeature.cpp:134:3: warning: missing field 'log_true_size' initializer [-Wmissing-field-initializers]
  134 |   LIBBPF_OPTS(bpf_btf_load_opts,
      |   ^
/usr/include/bpf/libbpf_common.h:70:3: note: expanded from macro 'LIBBPF_OPTS'
   70 |                 };                                                          \
      |                 ^
```
(ad.inf.)

These all have the same underlying problem: the [LIBBPF_OPTS macro](https://github.com/libbpf/libbpf/blob/42065ea6627ff6e1ab4c65e51042a70fbf30ff7c/src/libbpf_common.h#L64) uses memset to zero out the declared struct (which in itself can be problematic in C++, see e.g. [here](https://stackoverflow.com/questions/2837854/initializing-an-object-to-all-zeroes), which does not impress at least clang-18 in the slightest due to the .sz designated initializer, and so it helpfully reminds us to also initialize all the other members.

This seems to be a cause of much grief everywhere. Example discussions with no useful conclusions are e.g.:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82283
https://github.com/llvm/llvm-project/issues/68933
[...]

PR #3277, opened 2024-06-26
Fix LIBBPF_OPTS noise

hhoffstaette:
- add `-Werror=missing-field-initializers` to the CI build config so that accidental use of the wrong macro fails the build
- wrap LIBBPF_OPTS into a pragma-guarded macro to silence `-Wmissing-field-initializers`

Fixes: #3276 

##### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ set (CMAKE_CXX_EXTENSIONS OFF)
 
 add_compile_options("-Wall")
 add_compile_options("-Wextra")
+add_compile_options("-Werror=missing-field-initializers")
 add_compile_options("-Werror=undef")
 add_compile_options("-Wpointer-arith")
 add_compile_options("-Wcast-align")
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -752,7 +752,7 @@ void AttachedProbe::load_prog(BPFfeature &feature)
       prog_type = progtype(ProbeType::uprobe);
 
     {
-      LIBBPF_OPTS(bpf_prog_load_opts, opts);
+      BPFTRACE_LIBBPF_OPTS(bpf_prog_load_opts, opts);
       opts.log_buf = log_buf.get();
       opts.log_size = log_buf_size;
       opts.log_level = log_level;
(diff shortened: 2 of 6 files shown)


hhoffstaette:
So it turns out the first cut works with clang but NOT with gcc because [reasons](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78657) :angry: 
However after some fiddling I found that this seems to only affect the push/pop pragmas, and removing them allows this to work after all. The only other alternative would have been to duplicate the LIBBPF_OPS definition itself.

jordalgo (approved):
LGTM, but I'll wait for @viktormalik to review.

src/bpffeature.h
```diff
@@ -54,6 +54,10 @@ public:                                                                        \
 #define DEFINE_PROG_TEST(var, progtype)                                        \
   __DEFINE_PROG_TEST(var, progtype, NULL, std::nullopt)
 
+#define BPFTRACE_LIBBPF_OPTS(TYPE, NAME, ...)                                  \
+  _Pragma("GCC diagnostic ignored \"-Wmissing-field-initializers\"")           \
+      LIBBPF_OPTS(TYPE, NAME, __VA_ARGS__)
```
danobi, line 59: nit: I think it's better to define this in `src/libbpf/bpf.h`

flake.nix
```diff
@@ -98,6 +98,7 @@
                 # Release flags
                 cmakeFlags = [
                   "-DCMAKE_BUILD_TYPE=Release"
+                  "-DCMAKE_CXX_FLAGS=-Werror=missing-field-initializers"
```
danobi, line 101: So I think the scope is wider than just CI - we'd like anyone building without nix to hit this too, right? If so, probably throw this in here? https://github.com/bpftrace/bpftrace/blob/3e915ae21d32418ac56f4e6174f96918a5926d9e/CMakeLists.txt#L36-L46

danobi (commented):
small nits, lgtm otherwise

flake.nix
hhoffstaette: I know *nothing* about nix and only found the cmakeFlags while spelunking. :)
During testing I added -DCMAKE_CXX_FLAGS to verify, but I agree it makes sense to add the -Werror bit to all builds. Will do.

src/bpffeature.h
hhoffstaette: Fine with me - will do.

danobi (approved):
thanks for the work on fixing warnings!

viktormalik (approved):
Nice, thanks!

Issue #3278, opened 2024-06-26
Character literals are not supported - ERROR: invalid character '''


giorio94:
### What reproduces the bug? Provide code if possible.

Character literals are described in the [user manual](https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#literals) as to be enclosed in single quotes, but they don't appear to work. Reproduced with bpftrace v0.20.1 and v0.21.1. 

```
cat<<"EOF" > test.bt
BEGIN {
    $c = 'c';
    printf("Test %c\n", $c);
}
EOF

sudo bpftrace test.bt

test.bt:2:11-12: ERROR: invalid character '''
    $ch = 'a';
          ~
test.bt:2:11-14: ERROR: invalid character '''
    $ch = 'a';
          ~~~
```

### `bpftrace --info` output

```
System
  OS: Linux 6.1.0-18-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01)
  Arch: x86_64

Build
  version: v0.21.1
  LLVM: 18.1.7
  unsafe probe: no
  bfd: yes
  liblldb (DWARF support): yes
  libsystemd (systemd notify support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes
  for_each_map_elem: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: no
  uprobe_multi: no
  raw_tp_special: yes
  iter: yes
```

ajor:
Character literals have never been supported as far as I know. I'm surprised they're documented!

Looks like that docs entry was added in #1900

giorio94:
[...]

PR #3283, opened 2024-06-28
docs: remove mention to unsupported character literals

giorio94:
Character literals appear to not be supported at this time. Hence, let's clarify this aspect in the documentation.

Fixes: fd0461dc ("docs: entries for new int syntax")
Fixes: #3278

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -501,7 +501,7 @@ Identifiers must match the following regular expression: `[_a-zA-Z][_a-zA-Z0-9]*
 
 === Literals
 
-Integer, char, and string literals are supported.
+Integer and string literals are supported.
 
 Integer literals can be defined in the following formats:
 
@@ -520,11 +520,17 @@ To improve the readability of big literals an underscore `_` can be used as fiel
 Integer suffixes as found in the C language are parsed by bpftrace to ensure compatibility with C headers/definitions but they're not used as size specifiers.
 `123UL`, `123U` and `123LL` all result in the same integer type with a value of `123`.
 
-Character literals can be defined by enclosing the character in single quotes e.g. `$c = 'c';`.
+Character literals are not supported at this time, and the corresponding ASCII code must be used instead:
+
+----
+BEGIN {
+  printf("Echo A: %c\n", 65);
+}
+----
 
 String literals can be defined by enclosing the character string in double quotes e.g. `$str = "Hello world";`.
 
-Characters and strings support the following escape sequences:
+Strings support the following escape sequences:
 
 [cols="~,~"]
 |===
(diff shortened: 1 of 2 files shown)


viktormalik (approved):
Thanks!

man/adoc/bpftrace.adoc
```diff
@@ -520,11 +520,11 @@ To improve the readability of big literals an underscore `_` can be used as fiel
 Integer suffixes as found in the C language are parsed by bpftrace to ensure compatibility with C headers/definitions but they're not used as size specifiers.
 `123UL`, `123U` and `123LL` all result in the same integer type with a value of `123`.
 
-Character literals can be defined by enclosing the character in single quotes e.g. `$c = 'c';`.
+Character literals are not supported at this time, and the corresponding ASCII code must be used instead.
```
jordalgo, line 523: > corresponding ASCII code must be used instead.

Maybe I'm dense but I'm not sure what this means in this context. Do you have an example?

danobi: `man ascii` - for example: `65` instead of `A`

jordalgo: Ah ok, and this only works for `printf` as you need "%c".

giorio94: Yes, exactly, I was referring to that use-case (which is admittedly a niche one).

In my specific case, I'm outputting the TCP flags with something along the lines of:

```
#define CH_S 0x53   // 'S'
#define CH_A 0x41   // 'A'
...

printf("TCP flags: %c%c%c%c\n",
        $ip4h->protocol == PROTO_TCP && $tcph->syn ? CH_S : CH_DOT,
        $ip4h->protocol == PROTO_TCP && $tcph->ack ? CH_A : CH_DOT,
        $ip4h->protocol == PROTO_TCP && $tcph->fin ? CH_F : CH_DOT,
        $ip4h->protocol == PROTO_TCP && $tcph->rst ? CH_R : CH_DOT
)
```

I had initially used strings instead, but that turned out triggering the BPF stack limit exceeded error (given that each requires 64 bytes).

jordalgo: I wouldn't be opposed to including a `printf` example just so users don't get confused (as I did) but I also won't block if someone else wants to merge.

giorio94: Last push added a very basic example.

ajor (approved):
Looks good to me, thank you!

man/adoc/bpftrace.adoc
jordalgo: Thanks! Looks great.

jordalgo (approved):

PR #3114, opened 2024-04-11
Align detect_helper with log message change

mykyta5:
"unknown func " log message, produced by  BPF verifier may change in future to  "program of this type cannot use helper ". Corresponding kernel change:
 https://lore.kernel.org/all/20240325152210.377548-1-yatsenko@meta.com/

PR #3285, opened 2024-06-28
Let libbpf load BPF programs

### LINKED ISSUE #2334: libbpf as a main part of program loading

viktormalik:
This is (finally!) the last part of #2334.

Instead of using `bpf_prog_load` to load BPF programs and `bpf_map_create` to create BPF maps, use `bpf_object__load` to load everything at once, just as libbpf does. This leverages the fact that codegen now produces an ELF compliant with libbpf.
    
Required properties of BPF programs (namely program type, expected attach type, and attach target) still need to be fixed manually (as we're not using libbpf-compatible section names), this is now done in `BpfBytecode`.
    
Verifier log is now kept per-program as the verifier overrides it for every program load. This also means that when the user sets config.log_size to `size`, the actual amount of memory allocated is (`size * number_of_probes`).
    
Thanks to this change, a lot of code could have been dropped:
- the entire ELF parsing and `BpfBytecode::sections_` (ELF is now directly passed to libbpf and all data is accessed via operations over `bpf_object` and `bpf_program`),
- custom relocations of BPF programs (now done by libbpf),
- custom fixups of BTF data (now done by libbpf).
    
One problem is kfunc probes require libbpf/libbpf@dd589c3b31c13164bdc61ed174fbae6fe76c8308. Unfortunately, it's not released, yet, so we must consider two things:
1. Do we want to have master depending on unreleased libbpf features?
2. Once the patch is released (should be in the next libbpf version 1.5.0), do we want bpftrace to require the latest libbpf (which many distros do not package)?
    
The change also broke two builtin calls: `debugf` and `printf` for iter probes so I reworked the codegen for them to also leverage libbpf. See message of the second commit for implementation details.

There should be no functional change so I didn't introduce any new tests.

Resolves #2334.

##### Checklist


danobi:
> > One problem is kfunc probes require [libbpf/libbpf@dd589c3](https://github.com/libbpf/libbpf/commit/dd589c3b31c13164bdc61ed174fbae6fe76c8308). Unfortunately, it's not released, yet, so we must consider two things:
> > 
> > 1. Do we want to have master depending on unreleased libbpf features?
> > 2. Once the patch is released (should be in the next libbpf version 1.5.0), do we want bpftrace to require the latest libbpf (which many distros do not package)?
> 
> I'm personally ok with both. My argument is that we'll be using libbpf features more and more in future so we should not overwhelm ourselves by trying to keep backwards compatibility with old libbpf versions. If a distro wants to package the latest bpftrace, they should package the latest libbpf, too. If a developer wants to build the latest bpftrace from master, they may need to build the latest libbpf, too (and we may include it in the Nix env).

Agreed. We discussed this a long time ago when debating libbpf loader migration and I think this exact situation was an expected conclusion. We also had some input from various distro packagers saying it was ok to do this.

All the infra is in place for devs on our side too. We can pin the unreleased version in flake.nix. And we can bump libbpf version requirements here: https://github.com/bpftrace/bpftrace/blob/b3997d1534bbbf87a0adeeda7afa627773021d34/CMakeLists.txt#L89-L91

src/bpfbytecode.cpp
```diff
@@ -83,51 +80,115 @@ BpfProgram &BpfBytecode::getProgramForProbe(const Probe &probe)
...
+ *     36: (b7) r3 = 64                      ; R3_w=64
+ *     37: (85) call bpf_d_path#147
+ *     helper call is not allowed in probe
+ *     [...]
+ *
+ *  In the above log, "bpf_d_path" is the helper's name and "147" is the ID.
+ */
+void maybe_throw_helper_verifier_error(std::string_view log,
+                                       std::string_view err_pattern,
+                                       const std::string &exception_msg_suffix)
```
danobi, line 99: This feels a bit unreliable. Verifier output could change - not sure it is stable.

What is the motivation for this? Doesn't feature probing prevent this?

danobi (commented):
Didn't get a chance to look closely at first commit yet.

Second commit lgtm; great cleanup!

src/bpfprogram.cpp
```diff
@@ -1,569 +1,99 @@
...
+      // one attachpoint in a multi-attachpoint (wildcard or list) probe
+      // failed, print a warning but continue
+      LOG(WARNING) << msg << ", skipping.";
     } else {
-      LOG(ERROR) << "Relocation in " << relsecname << " type " << reltype
-                 << " sym " << relsym << " type " << symtype;
-      throw std::invalid_argument(
-          "Unsupported symbol type referenced in relocation");
+      // explicit match failed, fail hard
+      throw std::runtime_error(msg);
```
danobi, line 76: Throw a FatalUserException instead?

danobi (commented):
Looks great. I don't see any other obvious issues 

> This is (finally!) the last part of #2334.
> 
> Instead of using `bpf_prog_load` to load BPF programs and `bpf_map_create` to create BPF maps, use `bpf_object__load` to load everything at once, just as libbpf does. This leverages the fact that codegen now produces an ELF compliant with libbpf.
>     
> Required properties of BPF programs (namely program type, expected attach type, and attach target) still need to be fixed manually (as we're not using libbpf-compatible section names), this is now done in `BpfBytecode`.
>     
> Verifier log is now kept per-program as the verifier overrides it for every program load. This also means that when the user sets config.log_size to `size`, the actual amount of memory allocated is (`size * number_of_probes`).
>     
> Thanks to this change, a lot of code could have been dropped:
> - the entire ELF parsing and `BpfBytecode::sections_` (ELF is now directly passed to libbpf and all data is accessed via operations over `bpf_object` and `bpf_program`),
> - custom relocations of BPF programs (now done by libbpf),
> - custom fixups of BTF data (now done by libbpf).
>     
> One problem is kfunc probes require libbpf/libbpf@dd589c3b31c13164bdc61ed174fbae6fe76c8308. Unfortunately, it's not released, yet, so we must consider two things:
> 1. Do we want to have master depending on unreleased libbpf features?
> 2. Once the patch is released (should be in the next libbpf version 1.5.0), do we want bpftrace to require the latest libbpf (which many distros do not package)?
>     
> The change also broke two builtin calls: `debugf` and `printf` for iter probes so I reworked the codegen for them to also leverage libbpf. See message of the second commit for implementation details.
> 
> There should be no functional change so I didn't introduce any new tests.
> 
> Resolves #2334.
> 
> ##### Checklist
> 
> - [ ] Language changes are updated in `man/adoc/bpftrace.adoc`
> - [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
> - [ ] The new behaviour is covered by tests
>

jordalgo:
> Once the patch is released (should be in the next libbpf version 1.5.0), do we want bpftrace to require the latest libbpf (which many distros do not package)?

Is it possible to just include libbpf as a submodule and then statically linking it so distros don't have to also package or update the latest version of libbpf?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4246,12 +4221,31 @@ Function *CodegenLLVM::createForEachMapCallback(
   return callback;
 }
 
+// BPF helpers that use fmt strings (bpf_trace_printk, bpf_seq_printf) expect
```
jordalgo, line 4224: Ha. Good comment. I didn't know libbpf did this.

src/bpfbytecode.cpp
```diff
@@ -83,51 +80,115 @@ BpfProgram &BpfBytecode::getProgramForProbe(const Probe &probe)
...
-                                   map.second.max_entries(),
-                                   &opts);
-    if (map.second.fd < 0) {
-      failed_maps++;
-      LOG(ERROR) << "failed to create map: '" << map.first
-                 << "': " << strerror(errno);
-    }
-  }
-
-  if (failed_maps > 0) {
```
jordalgo, line 125: How do we handle failing to create a map now?

src/bpftrace.cpp
```diff
@@ -1315,13 +1292,13 @@ int BPFtrace::clear_map(const BpfMap &map)
 
   // snapshot keys, then operate on them
   std::vector<std::vector<uint8_t>> keys;
-  while (bpf_get_next_key(map.fd, old_key.data(), key.data()) == 0) {
+  while (bpf_get_next_key(map.fd(), old_key.data(), key.data()) == 0) {
```
jordalgo, line 1295: Is the call to `bpf_map__fd(bpf_map_);` cheap? If not, maybe save the map fd outside these while loops into a local var.

jordalgo:
Love see all this deleted code. Nice job, @viktormalik! All looks surprisingly straightforward.

Just curious, it seems like multiple subprogs now work with this, which is AWESOME! Did you want to revert [this commit](https://github.com/bpftrace/bpftrace/pull/3233/files) in this PR and update the tests?

src/bpfbytecode.cpp
viktormalik: > This feels a bit unreliable. Verifier output could change - not sure it is stable.

This has been around for a long time now, I just moved the code. The truth is that verifier output may change (see #3114 for an example) but we don't have a better way to do this now.

> 
> What is the motivation for this? Doesn't feature probing prevent this?

No, feature probing doesn't help here, some BPF helpers are only allowed in some functions (e.g. `bpf_d_path`, see [here](https://elixir.bootlin.com/linux/latest/source/kernel/trace/bpf_trace.c#L937)).

The motivation is that we want to give the user a nice error message in case he tries to use a helper in an unsupported function so that he doesn't have to deal with the verifier output.

src/bpfprogram.cpp
viktormalik: Good point. Also I missed to respect the new `missing_probes` config.

viktormalik:
> > Once the patch is released (should be in the next libbpf version 1.5.0), do we want bpftrace to require the latest libbpf (which many distros do not package)?
> 
> Is it possible to just include libbpf as a submodule and then statically linking it so distros don't have to also package or update the latest version of libbpf?

Unfortunately not. Bundling dependencies goes against policies of most distros so we wouldn't save packagers' time at all.

I'd also prefer staying away from submodules. I believe Nix provides a much nicer solution to this problem. We've also tried static linking in the past but moved towards Nix eventually.

src/bpfbytecode.cpp
viktormalik: Good point, we don't. Thinking about it led me to realize that we now have fewer options to figure out what went wrong in prog loading or map creation in general. I tried to rework it to give user as many hints as possible, see `BpfBytecode::load_progs` for details. We can also discuss during office hours.

viktormalik:
> Just curious, it seems like multiple subprogs now work with this, which is AWESOME! Did you want to revert [this commit](https://github.com/bpftrace/bpftrace/pull/3233/files) in this PR and update the tests?

Absolutely! I'll re-enable the feature.

jordalgo (approved):
Looks great to me!

src/bpfprogram.cpp
```diff
@@ -1,569 +1,104 @@
...
-   * needed, also pass the kernel version (try three different methods):
-   */
-  struct load_strategy {
-    bool btf;
-    bool func_infos;
-    KernelVersionMethod kver;
-  } load_configs[] = {
-    { .btf = true, .func_infos = true, .kver = None },  // BTF and func_infos
-    { .btf = true, .func_infos = false, .kver = None }, // BTF, no func_infos
-    /* Fallbacks for Linux <5.0: No BTF and func_infos, with kernel version: */
```
simpleton, line 364: @viktormalik do you happen know to whether the old kernel(e.g 4.19) still work w/o these fallbacks code?

viktormalik: @simpleton now all program loading is handled by libbpf so if there's a sufficiently new libbpf on the system, it *should* work.

danobi: Anything older than 5.4 we don't support (https://github.com/bpftrace/bpftrace/blob/master/docs/dependency_support.md#linux-kernel).

PR #3287, opened 2024-06-28
Fix racy per-cpu runtime test

jordalgo:
"print_per_cpu_map_vals" is a racy test (and breaking CI) due to the use of a very active probe and per-cpu maps. Let's just use 'BEGIN' to test printing even though this doesn't really exercise the cpu aggregation path as 'BEGIN' only runs on one CPU.

##### Checklist

### DIFF
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -296,7 +296,8 @@ PROG BEGIN { @x = "xxxxx"; @x = "a"; @[@x] = 1; @y = "yyyyy"; @y = "a"; @[@y] = 
 EXPECT len: 1
 TIMEOUT 1
 
+# Re-add min/max after they are fixed: https://github.com/bpftrace/bpftrace/issues/3286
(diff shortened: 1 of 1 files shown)


danobi:
Do you happen to know what the actual race was? I can't think of anything other than the probe concurrently on different CPUs. Or maybe compiler reordered some statements.

If not, would be good to find out. Maybe it'll reveal something fundamental.

danobi (approved):

jordalgo:
> Do you happen to know what the actual race was? I can't think of anything other than the probe concurrently on different CPUs. Or maybe compiler reordered some statements. If not, would be good to find out. Maybe it'll reveal something fundamental.

Discussed a little offline but fairly certain it's just the probe executing on different CPUs while aggregating. If I slow things down (e.g. use a probe that doesn't fire as often like `hardware:cache-misses:1000000`) then the racy-ness goes away. Also confirmed the probe is running on different cpus.

Issue #3286, opened 2024-06-28
min/max aggregations are broken


jordalgo:
Since maps are 0 initialized it means that the values across all CPUs to start for `min`/`max` are 0. This means we don't register a difference between the initial value and a value set by the user e.g.

```
sudo ./src/bpftrace -e 'BEGIN { @mn = min(1); @mx = max(-10); print((@mn, @mx)); exit(); }'
```
outputs: (0, 0)

Maybe we can just store another value like we do for `avg` that is a bool to say if this val has been set by the user on this cpu.

jordalgo:
Additionally `min_value`in utils is calculating the MAX value lol:
```
int64_t min_value(const std::vector<uint8_t> &value, int nvalues)
{
  int64_t val, max = 0;
  for (int i = 0; i < nvalues; i++) {
    val = read_data<int64_t>(value.data() + i * sizeof(int64_t));
    if (val > max)
      max = val;
  }

  return max;
}
```

ajor:
min/max aggregations appear to be working to me. We can tell the difference between "set to 0" and "unset" from the return value of `map_lookup_elem`. Taking a look at the IR for this script shows that bpftrace is doing this correctly: `BEGIN { @ = min(*(0)); }`.

I think this is actually a bug in putting per-cpu values into tuples:
```
# bpftrace -e 'BEGIN { @mn = min(1); @mx = max(-10); print(@mn); print(@mx); print((@mn, @mx)); exit(); }'
Attaching 1 probe...
@mn: 1
@mx: 18446744073709551606
(0, 0)
```

(18446744073709551606 is -10 as a uint64)

ajor:
I'm not sure that min/max are actually broken - I haven't been able to trigger any wrong-results except by putting these per-cpu values into tuples (which suggests tuples are the problem):
```
# bpftrace -e 'BEGIN { @mn = min(1); @mx = max(-10); print(@mn); print(@mx); print((@mn, @mx)); exit(); }'
Attaching 1 probe...
@mn: 1                      [CORRECT]
@mx: 18446744073709551606   [CORRECT]
(0, 0)                      [WRONG]
```
(18446744073709551606 is -10 as a uint64)

I don't suppose there's any documentation on how a per-cpu map works? If a value is inserted on a single CPU, do all CPU buckets then definitely count as "filled"? My experiments are suggesting not.


> Additionally `min_value`in utils is calculating the MAX value lol

[...]

PR #3288, opened 2024-06-30
Fix min/max map functions

jordalgo:
Since these work off of per-cpu maps, it means that when a key is added the value for the key across all cpus initializes to 0. Therefore if we loop over all cpus to determine if a value is greater or less than another we need to discount all 0 values that weren't set explicitly by the user/prog (e.g. min(0)).

To do this we need to keep a separate key (similar to `avg`) that acts like a "value was set" boolean so we know if we should consider it in a less-than or greater-than operation.

This needs to be done in both userspace (for map printing) and in kernel space for min/max implicit casting.

Issue: https://github.com/bpftrace/bpftrace/issues/3286

##### Checklist


jordalgo:
This is ugly but correct. The alternative is to have min/max use hash maps and an atomic min/max write after lookup - I don't know what would be faster but I'll try to run some tests to find out.

ajor:
I don't understand the exact problem here as I can't trigger any wrong results except by using tuples. Left a comment on the issue (#3286).

jordalgo:
Responded in the issue with a more detailed repro. 

Here is the alternative solution where we just convert min/max to use a hash map, which I'm leaning towards because it's just so much simpler AND preserves the ability to loop over min/max maps. https://github.com/bpftrace/bpftrace/pull/3290

jordalgo:
Abandoning in favor of https://github.com/bpftrace/bpftrace/pull/3298, which stores a struct as the min/max map values instead of using two separate keys.

PR #3289, opened 2024-07-01
CHANGELOG: Move `buf` entry to the correct version (i.e. unreleased)

ajor:
- Remove entry for #3281, as it's a fix for a bug that hasn't made it to a release
- Quote a bunch of things with backticks to make the markdown render more nicely


jordalgo (approved):
Good catch.

PR #3296, opened 2024-07-05
Cleanups: Use "const" and reduce string copies

ajor:


jordalgo (approved):
insert JT "cleaner" gif

viktormalik (approved):
Thanks!

PR #3297, opened 2024-07-05
Bump C++ standard to C++20

ajor:
##### Checklist


ajor:
It turns out that LLVM 12 is not compatible with C++20. Maybe it's time to drop support for LLVM 12?

jordalgo:
I'm fine to move to llvm-13 as the min but there are other things that might need to be updated e.g. the codegen script. I'm also not sure of the implications on the distro side.

viktormalik:
> I'm also not sure of the implications on the distro side.

There are LTS distros which have older LLVM (e.g. Ubuntu 20.04 with LLVM 10) but I think that they generally do not update bpftrace either. So, I *think* that we should be fine here but not 100% sure.

Codegen tests are currently on LLVM 12 so they would need to be migrated. Preferably to the latest (LLVM 18) but we'd lose codegen tests for typed (non-opaque) pointers, then.

ajor:
I'm going to update the codegen tests in #3303

ajor:
LLVM 12 has been removed as of #3325 and we're ready for C++20!

jordalgo (approved):
:rocket:

PR #3300, opened 2024-07-08
Keyless maps default to BPF_MAP_TYPE_ARRAY for some types

samuelnair:
Keyless maps like `@x` or `@` when storing certain types can be special cased to BPF_MAP_TYPE_ARRAY reduce some overhead of using BPF_MAP_TYPE_HASH described here https://github.com/bpftrace/bpftrace/issues/3186.
Except Hist, LHist, Avg, Stats, Count, Sum, Max & Min all keyless maps will default to BPF_MAP_TYPE_ARRAY
##### Checklist


jordalgo:
@samuelnair From the looks of the failures, you need to [update the codegen tests](https://github.com/bpftrace/bpftrace/blob/master/tests/README.md#updating) and fix the formatting `git clang-format upstream/master`

samuelnair:
@jordalgo Why do i need to update my codegen-tests ? I have not changed anything in the tests and these are all running in the CI.

fbs:
> @jordalgo Why do i need to update my codegen-tests ? I have not changed anything in the tests and these are all running in the CI.

You've updated the code generator itself. This usually influences the generated code, which is what the codegen tests test.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3564,8 +3564,22 @@ void CodegenLLVM::generate_maps(const RequiredResources &resources)
...
+    // HASH for efficiency
     if (val_type.IsCountTy() && key.args_.empty()) {
       max_entries = 1;
+    } else if (!(val_type.IsHistTy() || val_type.IsLhistTy() ||
+                 val_type.IsAvgTy() || val_type.IsStatsTy() ||
+                 val_type.IsCountTy() || val_type.IsSumTy() ||
+                 val_type.IsMaxTy() || val_type.IsMinTy()) &&
+               key.args_.empty()) {
+      createMapDefinition(name,
+                          libbpf::BPF_MAP_TYPE_ARRAY,
```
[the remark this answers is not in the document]
viktormalik: I think that what @ajor meant is that we should just set `max_entries` here and use the `createMapDefinition` below for all cases.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3525,6 +3525,12 @@ libbpf::bpf_map_type CodegenLLVM::get_map_type(const SizedType &val_type,
               val_type.IsMinTy() || val_type.IsMaxTy() || val_type.IsAvgTy() ||
               val_type.IsStatsTy())) {
     return libbpf::BPF_MAP_TYPE_PERCPU_HASH;
+  } else if (!(val_type.IsHistTy() || val_type.IsLhistTy() ||
+               val_type.IsAvgTy() || val_type.IsStatsTy() ||
+               val_type.IsCountTy() || val_type.IsSumTy() ||
+               val_type.IsMaxTy() || val_type.IsMinTy()) &&
```
viktormalik, line 3531: This list is the same as above. Could we enclose the check into a function and use it for both?

For example a function called `needs_percpu_map` (or similar) sounds reasonable.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3513,18 +3513,25 @@ void CodegenLLVM::createMapDefinition(const std::string &name,
   var->setDSOLocal(true);
   var->addDebugInfo(debuginfo);
 }
+bool CodegenLLVM::needs_percpu_map(const SizedType &val_type)
+{
+  return ((val_type.IsHistTy() || val_type.IsLhistTy() ||
+           val_type.IsCountTy() || val_type.IsSumTy() || val_type.IsMinTy() ||
+           val_type.IsMaxTy() || val_type.IsAvgTy() || val_type.IsStatsTy()))
+             ? true
+             : false;
```
jordalgo, line 3522: nit: You don't need this ternary and these parens.
```suggestion
  return val_type.IsHistTy() || val_type.IsLhistTy() ||
           val_type.IsCountTy() || val_type.IsSumTy() || val_type.IsMinTy() ||
           val_type.IsMaxTy() || val_type.IsAvgTy() || val_type.IsStatsTy();
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3513,18 +3513,25 @@ void CodegenLLVM::createMapDefinition(const std::string &name,
...
     return libbpf::BPF_MAP_TYPE_PERCPU_ARRAY;
   } else if (bpftrace_.feature_->has_map_percpu_hash() &&
-             (val_type.IsHistTy() || val_type.IsLhistTy() ||
-              val_type.IsCountTy() || val_type.IsSumTy() ||
-              val_type.IsMinTy() || val_type.IsMaxTy() || val_type.IsAvgTy() ||
-              val_type.IsStatsTy())) {
+             needs_percpu_map(val_type)) {
     return libbpf::BPF_MAP_TYPE_PERCPU_HASH;
+  } else if (!needs_percpu_map(val_type) && key.args_.empty()) {
+    return libbpf::BPF_MAP_TYPE_ARRAY;
```
jordalgo, line 3534: I guess this should be fine if the value is a string or a tuple.

jordalgo (approved):
Code looks good to me though I don't think you need the changelog message since this is mostly internal BE - either that or rephrase as to indicate faster map access due to the new map type.

src/ast/dibuilderbpf.cpp
```diff
@@ -164,12 +165,16 @@ DIType *DIBuilderBPF::GetType(const SizedType &stype)
...
+                                    libbpf::bpf_map_type map_type)
 {
   // No-key maps use '0' as the key.
   // No-key count() maps use BPF_MAP_TYPE_PERCPU_ARRAY which needs 4-byte key.
   if (key.args_.size() == 0)
-    return value_type.IsCountTy() ? getInt32Ty() : getInt64Ty();
+    return (map_type == libbpf::BPF_MAP_TYPE_PERCPU_ARRAY ||
+            map_type == libbpf::BPF_MAP_TYPE_ARRAY)
+               ? getInt32Ty()
+               : getInt64Ty();
```
viktormalik, line 177: Needs to update the comment above?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3513,18 +3513,23 @@ void CodegenLLVM::createMapDefinition(const std::string &name,
   var->setDSOLocal(true);
   var->addDebugInfo(debuginfo);
 }
+bool CodegenLLVM::needs_percpu_map(const SizedType &val_type)
```
viktormalik, line 3516: Nit: we tend to add empty lines between functions:
```suggestion

bool CodegenLLVM::needs_percpu_map(const SizedType &val_type)
```
(I'm surprised this is not enforced by clang-format)

src/ast/passes/codegen_llvm.h
```diff
@@ -99,6 +99,7 @@ class CodegenLLVM : public Visitor {
   libbpf::bpf_map_type get_map_type(const SizedType &val_type,
                                     const MapKey &key);
   void generate_maps(const RequiredResources &resources);
+  bool needs_percpu_map(const SizedType &val_type);
```
viktormalik, line 102: It looks like this would be better placed in `types.h` as a method of `SizedType` (and probably will make sense to name it `needsPercpuMap` to comply with the rest of the class).

ajor (approved):
Just clarified the comment in DIBuilder to remove old references to count().

Otherwise looks good to me!

PR #3301, opened 2024-07-08
kprobe inlined function

ttreyer:
Extend the `probe_inline` config to also apply to `kprobe`.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -326,42 +326,6 @@ std::string AttachedProbe::eventname() const
   }
 }
 
-static int sym_name_cb(const char *symname,
-                       uint64_t start,
-                       uint64_t size,
-                       void *p)
-{
-  struct symbol *sym = static_cast<struct symbol *>(p);
-
-  if (sym->name == symname) {
-    sym->start = start;
-    sym->size = size;
-    return -1;
-  }
-
-  return 0;
-}
-
-static int sym_address_cb(const char *symname,
-                          uint64_t start,
-                          uint64_t size,
-                          void *p)
-{
-  struct symbol *sym = static_cast<struct symbol *>(p);
-
-  // When size is 0, then [start, start + size) = [start, start) = ø.
-  // So we need a special case when size=0, but address matches the symbol's
-  if (sym->address == start ||
-      (sym->address > start && sym->address < (start + size))) {
-    sym->start = start;
-    sym->size = size;
-    sym->name = symname;
-    return -1;
-  }
-
-  return 0;
-}
-
 static uint64_t resolve_offset(const std::string &path,
                                const std::string &symbol,
                                uint64_t loc)
@@ -522,69 +486,54 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
   return true;
 }
 
-// find vmlinux file containing the given symbol information
-static std::string find_vmlinux(const struct vmlinux_location *locs,
-                                struct symbol &sym)
-{
-  struct bcc_symbol_option option = {};
-  option.use_debug_file = 0;
-  option.use_symbol_type = BCC_SYM_ALL_TYPES ^ (1 << STT_NOTYPE);
-  struct utsname buf;
-
-  uname(&buf);
-
-  for (int i = 0; locs[i].path; i++) {
-    if (locs[i].raw)
-      continue; // This file is for BTF. skip
-    char path[PATH_MAX + 1];
-    snprintf(path, PATH_MAX, locs[i].path, buf.release);
-    if (access(path, R_OK))
-      continue;
-    bcc_elf_foreach_sym(path, sym_name_cb, &option, &sym);
-    if (sym.start) {
-      LOG(V1) << "vmlinux: using " << path;
-      return path;
-    }
-  }
-
-  return "";
-}
-
 void AttachedProbe::resolve_offset_kprobe()
 {
-  struct symbol sym = {};
-  std::string &symbol = probe_.attach_point;
-  uint64_t func_offset = probe_.func_offset;
-  offset_ = func_offset;
+  offset_ = probe_.func_offset;
 
-  if (func_offset == 0)
+  // If we are using only the symbol, we don't need to check the offset.
+  bool is_symbol_kprobe = !probe_.attach_point.empty();
+  if (is_symbol_kprobe && probe_.func_offset == 0)
     return;
 
-  sym.name = symbol;
-  const struct vmlinux_location *locs = vmlinux_locs;
-  struct vmlinux_location locs_env[] = {
-    { nullptr, false },
-    { nullptr, false },
-  };
-  char *env_path = std::getenv("BPFTRACE_VMLINUX");
-  if (env_path) {
-    locs_env[0].path = env_path;
-    locs = locs_env;
-  }
+  // Setup the symbol to resolve, either using the address or the name.
+  struct symbol sym = {};
+  if (is_symbol_kprobe)
+    sym.name = probe_.attach_point;
+  else
+    sym.address = probe_.address;
 
-  std::string path = find_vmlinux(locs, sym);
-  if (path.empty()) {
-    LOG(V1) << "Could not resolve symbol " << symbol
+  auto path = find_vmlinux(&sym);
+  if (!path.has_value()) {
+    if (!is_symbol_kprobe)
+      throw FatalUserException("Could not resolve address: " +
+                               std::to_string(probe_.address));
+
+    LOG(V1) << "Could not resolve symbol " << probe_.attach_point
             << ". Skipping usermode offset checking.";
     LOG(V1) << "The kernel will verify the safety of the location but "
                "will also allow the offset to be in a different symbol.";
-
     return;
   }
 
-  if (func_offset >= sym.size)
-    throw FatalUserException("Offset outside the function bounds ('" + symbol +
-                             "' size is " + std::to_string(sym.size) + ")");
+  // Populate probe_ fields according to the resolved symbol.
+  if (is_symbol_kprobe) {
+    probe_.address = sym.start + probe_.func_offset;
+  } else {
+    probe_.attach_point = std::move(sym.name);
+    if (__builtin_sub_overflow(probe_.address, sym.start, &probe_.func_offset))
+      LOG(BUG) << "Offset before the function bounds ('" << probe_.attach_point
+               << "' address is " << std::to_string(sym.start) << ")";
+    offset_ = probe_.func_offset;
+    // Set the name of the probe to the resolved symbol+offset, so that failure
+    // to attach can be ignored if the user set ConfigMissingProbes::warn.
+    probe_.name = "kprobe:" + probe_.attach_point + "+" +
+                  std::to_string(probe_.func_offset);
+  }
+
+  if (probe_.func_offset >= sym.size)
+    throw FatalUserException("Offset outside the function bounds ('" +
+                             probe_.attach_point + "' size is " +
+                             std::to_string(sym.size) + ")");
 }
 
 void AttachedProbe::attach_multi_kprobe()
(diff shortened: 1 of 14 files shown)


danobi (commented):
bit of early feedback: I don't think it's very common for vmlinux to come with dwarf. If available at all, it usually comes in a separate package.

ttreyer:
I reworked the initialisation of the DWARF parser to fail if the target doesn't have DebugInfo, so we don't query LLDB about a stripped binary. It improves the "Kernel DWARF" feature detection by not requiring a specific symbol to be present (an issue raised in #3445).
The calls to `lldb::SBDebugger::Initialize/Terminate` has also moved in its own class to benefit from RAII semantic. C++ object lifetime begins after the constructor returns, but if an exception is thrown in the constructor, the destructor won't be called. So if the `Dwarf` ctor threw an error, the `instance_counter` would not get decremented and `lldb::SBDebugger::Terminate()` would never get called.

.github/workflows/ci.yml
```diff
@@ -82,6 +82,15 @@ jobs:
     - uses: actions/checkout@v2
     - uses: DeterminateSystems/nix-installer-action@v11
     - uses: DeterminateSystems/magic-nix-cache-action@v6
+    - name: Install Kernel DebugInfo
+      # Install the Debug Info for the current Kernel,
+      # which is required for our runtime tests of kprobe inline functions.
+      run: |
+        # From Ubuntu's Wiki: https://wiki.ubuntu.com/DebuggingProgramCrash#Non-built-in_debug_symbol_packages_.28.2A-dbgsym.29
+        sudo apt-get install ubuntu-dbgsym-keyring
+        echo -e "deb http://ddebs.ubuntu.com $(lsb_release -cs) main restricted universe multiverse\ndeb http://ddebs.ubuntu.com $(lsb_release -cs)-updates main restricted universe multiverse\ndeb http://ddebs.ubuntu.com $(lsb_release -cs)-proposed main restricted universe multiverse" | sudo tee -a /etc/apt/sources.list.d/ddebs.list
```
jordalgo, line 91: I love the 'restricted universe multiverse' part 😛

tests/runtime/dwarf
```diff
@@ -205,3 +205,10 @@ PROG struct Foo { int b; } uprobe:./testprogs/uprobe_test:uprobeFunction2 { prin
 EXPECT foo1->b = 123
 REQUIRES_FEATURE dwarf
 BEFORE ./testprogs/uprobe_test
+
+NAME kprobe inlined function
+PROG config = { probe_inline = 1 } BEGIN { exit() }
+     kprobe:str_has_prefix { }
```
jordalgo, line 211: is this kernel function known to be commonly inlined?

tests/runtime/probe
```diff
@@ -210,7 +210,7 @@ WILL_FAIL
 
 NAME kprobe_warn_missing_probes
 PROG kprobe:vfs_read,kprobe:nonsense { exit(); }
-EXPECT WARNING: could not attach probe kprobe:nonsense, skipping.
+EXPECT WARNING: could not attach probe kprobe:nonsense (kprobe:vfs_read,kprobe:nonsense), skipping.
```
jordalgo, line 213: why have the warning show both the probes if only one is skipped?

.github/workflows/ci.yml
ttreyer: haha
Ideally, we'd rely on vmtests to test on the Kernel.
Until then, we have to use this solution that can sometimes be flaky 🙁

tests/runtime/dwarf
ttreyer: Yes, I looked in the Kernel source code for a function marked `always inlined` to test on.
Moreover, `str_has_prefix` seems like a stable target that won't disappear from the Kernel any time soon 🤞

tests/runtime/probe
ttreyer: Ah, I probably needed it for debugging the test. I can clean it up...

.github/workflows/ci.yml
danobi: Looks like it takes about a minute and a half to install in CI. And given that CI stability and speed is fairly critical, not sure we should rush this in. Better to direct effort towards vmtest work instead if it's really that desired. 

Kernel dwarf is a tricky to use in open source world anyways - how do users even know to install the debug info anyways? Even if they want to, it's on the order of gigabytes and maximally tricky to install (as evidenced by this block). And that's for distros that even ship debug info.

I'm not objecting to the dwarf support itself - it is a contained change that can be useful if user really needs it. But the CI bit is the uncontained change - it affects the entire project on a daily basis. I'd much rather prefer to get a fast and reliable mechanism in place. 

Perhaps we can run a test like this internally? Should be much less flaky and still provide coverage.

src/attached_probe.cpp
```diff
@@ -524,78 +488,68 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
...
+    sym.name = probe_.attach_point;
+  else
+    sym.address = probe_.address;
 
-  std::string path = find_vmlinux(locs, sym);
-  if (path.empty()) {
-    LOG(V1) << "Could not resolve symbol " << symbol
+  auto path = find_vmlinux(&sym);
+  if (!path.has_value()) {
+    if (probe_.attach_point.empty())
```
jordalgo, line 513:
```suggestion
    if (!is_symbol_kprobe)
```

src/attached_probe.cpp
```diff
@@ -524,78 +488,68 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
...
-  if (func_offset >= sym.size)
-    throw FatalUserException("Offset outside the function bounds ('" + symbol +
-                             "' size is " + std::to_string(sym.size) + ")");
+  // Populate probe_ fields according to the resolved symbol.
+  if (is_symbol_kprobe) {
+    probe_.address = sym.start + probe_.func_offset;
+  } else {
+    probe_.attach_point = std::move(sym.name);
+    if (__builtin_sub_overflow(probe_.address, sym.start, &probe_.func_offset))
+      throw FatalUserException("Offset before the function bounds ('" +
```
jordalgo, line 530: Maybe this is an opportunity to utilize the missing_probes config to either throw, log a warning, or just ignore and skip the probe.

src/attached_probe.cpp
```diff
@@ -524,78 +488,68 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
...
+                               std::to_string(sym.start) + ")");
+    // Set the name of the probe to the resolved symbol+offset, so that failure
+    // to attach can be ignored if the user set ConfigMissingProbes::warn.
+    probe_.name = "kprobe:" + probe_.attach_point + "+" +
+                  std::to_string(probe_.func_offset);
+  }
 
-  uint64_t sym_offset = resolve_offset(path, probe_.attach_point, probe_.loc);
+  if (probe_.func_offset >= sym.size)
+    throw FatalUserException("Offset outside the function bounds ('" +
```
jordalgo, line 540: same here

src/attached_probe.cpp
```diff
@@ -524,78 +488,68 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
...
+                             probe_.attach_point + "' size is " +
+                             std::to_string(sym.size) + ")");
 
-  check_alignment(
-      path, symbol, sym_offset, func_offset, safe_mode, probe_.type);
+  uint64_t sym_offset = resolve_offset(path.value(),
+                                       probe_.attach_point,
+                                       probe_.loc);
+  check_alignment(path.value(),
```
jordalgo, line 547: This check was removed because the kernel will reject the attachment anyway.

src/utils.cpp
```diff
@@ -109,6 +109,72 @@
...
+
+  for (size_t i = 0; locs[i].path; ++i) {
+    auto &loc = locs[i];
+    if (loc.raw)
+      continue; // This file is for BTF. skip
+
+    // Format the loc.path with the current release.
+    char path[PATH_MAX] = {};
+    int bytes_written = std::snprintf(
+        path, sizeof(path), loc.path, uts.release);
```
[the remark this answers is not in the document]
jordalgo: Probably needs some kind of comment to ignore this. cc @danobi

src/utils.cpp
```diff
@@ -109,6 +109,72 @@ const struct vmlinux_location vmlinux_locs[] = {
...
+      bcc_elf_symcb callback = !sym->name.empty() ? sym_name_cb
+                                                  : sym_address_cb;
+      struct bcc_symbol_option options = {
+        .use_debug_file = 0,
+        .check_debug_file_crc = 0,
+        .lazy_symbolize = 0,
+        .use_symbol_type = BCC_SYM_ALL_TYPES ^ (1 << STT_NOTYPE),
+      };
+      if (bcc_elf_foreach_sym(path, callback, &options, sym) == -1) {
+        LOG(ERROR) << "Failed to iterate over symbols in " << path;
```
jordalgo, line 164: should this be `LOG(BUG)`?

src/attached_probe.cpp
ttreyer: I'm not sure it makes sense for both those cases to use the MissingProbes config.
1. The error above should be marked "unreachable". In `find_vmlinux(&sym)`, `libbcc` looks through each symbol and returns the one containing the given `probe_.address`. So we know that `probe_.address` is `>= sym.start`, otherwise we'd have failed at the `find_vmlinux` call. So this check is more about defensive programming.
2. The second one makes more sense to me, but we have to distinguish the case where the offset was give by the user or if it comes from multiple probes (like probing an inlined function). Moreover, it also need to be changed in `resolve_offset_uprobe`.
Let me know what you think and I'll update the PR accordingly 🙂

ttreyer:
I don't understand why the runtime test `call.str_big_read` fails, but only for LLVM 15.
@danobi, do you have a clue?

jordalgo:
@ttreyer This may be broken or just incredibly flaky on trunk: https://github.com/bpftrace/bpftrace/pull/3503

src/attached_probe.cpp
jordalgo: For the error above #1, which you say is unreachable. Then it should probably just be a `LOG(BUG)` if we don't expect it to ever happen.

For #2. Yeah I agree. @ajor and I were sorta discussing similar around the alignment issue. We didn't really come to a consensus on how to handle the common case in terms of erroring or just warning and skipping attachment for the probe. We can push this to a discussion to a different PR though.

ajor (changes requested):
I don't think we're actually attaching to the correct addresses here - `bpf_attach_kprobe` picks the address based on `funcname` and `offset_`. This PR sets `funcname` in the line above, so that's correct, but `offset_` is never updated to reflect `probe.address`.

As well as sorting this out, I think we need to add a way to validate that the addresses being attached to are correct so that we can check this in tests.

src/dwarf_parser.h
```diff
@@ -43,6 +41,17 @@ class Dwarf {
...
   std::optional<Bitfield> resolve_bitfield(lldb::SBTypeMember field);
 
+  // Initialize/Terminate liblldb globally with RAII
+  class InstanceCounter {
+  private:
+    static std::atomic<size_t> count;
+
+  public:
+    InstanceCounter();
+    virtual ~InstanceCounter();
```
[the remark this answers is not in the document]
ttreyer: Good point, it doesn't need to be, as this class is not inherited from.
I've remove the `virtual`, but marked the `InstanceCounter` class as `final` to prevent any inheritance.

ttreyer:
For the vulnerability with `snprintf` receiving its format string from the environment, we could require of the user to give us the path to the Kernel directly and let them call `uname -r` to have the correct path.

For example, on my machine it would look like:
```
# Before:
$ BFPTRACE_VMLINUX='/usr/lib/modules/%1$s/vmlinux-%1$s' ...

# After:
$ BPFTRACE_VMLINUX="/usr/lib/modules/$(uname -r)/vmlinux-$(uname -r)" ...
```

So if the `BPFTRACE_VMLINUX` environment variable is set, we can use it directly, without having to do any lookup to find the Kernel image.

tests/runtime/dwarf
```diff
@@ -205,3 +205,14 @@ PROG struct Foo { int b; } uprobe:./testprogs/uprobe_test:uprobeFunction2 { prin
...
 BEFORE ./testprogs/uprobe_test
+
+# The function `str_has_prefix` is marked to be always inlined, so we'll get
+# more than one occurrence for it.
+# We first check that bpftrace attaches to more probes than the two defined in the script.
+# We also check that the offsets bpftrace attaches to are different from 0,
+# since they should be inline instances part of another function/symbol.
+NAME kprobe inlined function
+RUN {{BPFTRACE}} -ve \ 'config = { probe_inline = 1 } BEGIN { exit() } kprobe:str_has_prefix { }'
+EXPECT_NONE Attaching 2 probe...
```
ajor, line 216:
```suggestion
EXPECT_NONE Attaching 2 probes...
```
Typo. Also do you think it's worth checking against "Attaching 1 probe..." as well?

ajor (approved):
Looks good!

ajor (changes requested):
@ttreyer After reading the discussion in #3504, I gave this program a test:
```
kprobe:__x64_sys_openat {}
```

With this PR, we now appear to be using DWARF even for non-inlined kprobes and we're attaching at `__x64_sys_openat+15` instead of `__x64_sys_openat+0`.

IMO non-inlined kprobes should always resolve addresses using the symbol table.

ajor (approved):

PR #3305, opened 2024-07-10
Readme updates

jordalgo:
- Update main readme
- Add tools README.md
- Point folks to new user-tools repo for contributions
- Update CONTRIBUTING-TOOLS.md

##### Checklist


brendangregg:
I'd consider leaving a CONTRIBUTING-TOOLS here that we update, since we will still have tools here that get packaged and distributed, so this is still a location of tools. Thinking ahead, there will at some point be someone who thinks their tool should be here instead of user-tools, and having a documented stance will help the community understand it's a strategy and not a personal rejection.

The CONTRIBUTING-TOOLS here could begin with something like:

> All new tools must be submitted to the user-tools repository (and not here) where they have a chance to develop maturity and collect real world use cases. The tools that exist in this repository are small collection curated by the bpftrace maintainers.
> 
> The folllowing points describe how tools may one day be selected by the bpftrace maintainers for inclusion here:
> ...

On the same topic: I'd also edit the user-tools README (I could do a PR over there) to include some big warning that reads something like:

>  **This repository should NOT be packaged and distrubited, because it is essentially a large collection of unsupported, likely broken, sample code. This is not a typical software _project_, but this is rather a programmer's _community_, similar to a subreddit.**

Does that sound like it makes sense? Just want to be sure we're all on the same page.

README.md
```diff
@@ -4,13 +4,21 @@
...
-
-To learn more about bpftrace, see the [Manual](man/adoc/bpftrace.adoc) and [One-Liner Tutorial](docs/tutorial_one_liners.md).
-
-We are also holding regular [office hours](https://docs.google.com/document/d/1nt010RfL4s4gydhCPSJ-Z5mnFMFuD4NrcpVmUcyvu2E/edit?usp=sharing) open to the public.
-
-## One-Liners
+bpftrace is a high-level tracing language for Linux [eBPF](https://ebpf.io/what-is-ebpf/) available in recent Linux kernels (4.x). bpftrace uses LLVM as a backend to compile scripts to BPF-bytecode and makes use of [libbpf](https://github.com/libbpf/libbpf) for interacting with the Linux BPF system, as well as existing Linux tracing capabilities: kernel dynamic tracing (kprobes), user-level dynamic tracing (uprobes), tracepoints, etc. The bpftrace language is inspired by awk, C, and predecessor tracers such as DTrace and SystemTap. bpftrace was created by [Alastair Robertson](https://github.com/ajor).
```
viktormalik, line 7: Technically speaking, we're still using BCC for probe attachment. Should we leave that in the README until we change to libbpf-based attachment?

jordalgo: Ah. I thought we were only still using it for symbolication. I can add it back until we fully migrate.

jordalgo:
> Does that sound like it makes sense? Just want to be sure we're all on the same page.

@brendangregg - Yes, both suggestions make sense. I'll update this PR to re-include the CONTRIBUTING-TOOLS readme with the updated language. I'll also mention something about how the scripts in here are also used in CI to test core changes to bpftrace.

README.md
ajor: Oops, I forgot that was separate from program loading! I went around tell everyone there was no more BCC after your last PR 😅

jordalgo:
Added a warning message to the user-tools repo: https://github.com/bpftrace/user-tools/pull/15

README.md
```diff
@@ -36,98 +44,57 @@ bpftrace -e 'tracepoint:block:block_rq_issue { printf("%d %s %d\n", pid, comm, a
 # Count page faults by process
 bpftrace -e 'software:faults:1 { @[comm] = count(); }'
 
-# Count LLC cache misses by process name and PID (uses PMCs):
+# Count LLC cache misses by process and PID (uses PMCs):
```
ajor, line 47:
```suggestion
# Count LLC cache misses by thread name and PID (uses PMCs):
```
`comm` actually returns the name of the current thread. This is something that's been confusing users - didn't realise we had it wrong in the readme!

ajor (approved):

README.md
jordalgo: TBH, I wasn't sure what to do about ALL of these examples in the readme as you'll notice that a lot of them say "by process" when grouping by `comm`. That's why I just removed the "name" in this heading.

Maybe it make sense to just change all the ones that use `comm` to say  "by thread".

ajor: What's wrong with saying "name" as well? "By thread" could be interpreted to mean something more similar to "thread ID".

On a system with multiple Python processes, `comm` will return "python" for all of them, so aggregating on `comm` will group different threads and processes together because they share the same name.

jordalgo: > What's wrong with saying "name" as well? 

I was just removing for consistency with the other examples above but let's not focus on that.

> "By thread" could be interpreted to mean something more similar to "thread ID".

Ok, so how about "by thread name" so it's consistent with how we talk about `comm`?

ajor: Oh sorry, apparently I was too lazy to read your whole comment last time!

PR #3306, opened 2024-07-10
Ignore LLVM IR files when detecting repo language

viktormalik:
bpftrace uses `.ll` files in codegen tests (as expected results of codegen) but we now have so many tests that their amount is bigger than the amount of actual source files. This makes GitHub mark LLVM as the primary language of bpftrace.

Update `.gitattributes` file to tell GitHub to ignore `.ll` files during language detection.

Tested on my [fork](https://github.com/viktormalik/bpftrace), see the Languages section there for illustration.

##### Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.bt linguist-language=D
 *.bt linguist-vendored
+*.ll linguist-detectable=false


jordalgo (approved):
Neat. I didn't know you could do this.

PR #3313, opened 2024-07-12
Codegen Tests: Remove unnecessary uses of count() function

ajor:
Simplify the test cases such that each is only testing the minimal functionality that it needs to. Minimises noise from changing expected results in unrelated patches.


jordalgo (approved):
Nice.

PR #3321, opened 2024-07-16
Remove test binary created to test CFLAGs

jordalgo:
Caused by this commit:
https://github.com/bpftrace/bpftrace/commit/1021b5f6275ef1cec57ce5c3c2a04580d6020088

##### Checklist

### DIFF
--- a/tests/testprogs/CMakeLists.txt
+++ b/tests/testprogs/CMakeLists.txt
@@ -13,6 +13,7 @@ function(test_and_add_testprog_cflag flag)
   else()
     message(STATUS "${CMAKE_C_COMPILER} does not support ${flag}")
   endif()
+  execute_process(COMMAND rm -rf -)
 endfunction()
 
 test_and_add_testprog_cflag("-fno-omit-frame-pointer")


viktormalik (approved):

ajor (changes requested):
I'm sure there must be a better way we can do this - the test program shouldn't get compiled into the source directory at the very least

ajor:
[try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html) looks promising, but I can't tell if it supports specifying custom compiler flags

ajor:
This works locally at least: https://github.com/ajor/bpftrace2/commit/d42cd9c30d6a1e6476c25c496f32f6d71247afb3

The only problem is that it's using a CMake 3.25 feature and we only require 3.14 right now.

ajor:
I've put up a CMake 3.14-compatible approach here: #3323
(it uses `simple_struct.c` as a test program instead of specifying the source inline)

jordalgo:
Closing in favor of https://github.com/bpftrace/bpftrace/pull/3323

PR #3322, opened 2024-07-17
Fix stack mode for stack builtin

jordalgo:
The stack mode for kstack and ustack builtin
variables (i.e. without ()) were not respecting
the stack mode value set by the config block.

Example:
```
bpftrace -e 'config = { stack_mode = raw } kprobe:vfs_read { @ = kstack; exit() }'
bpftrace -e 'config = { stack_mode = raw } kprobe:vfs_read { @[kstack] = 0; exit() }'
bpftrace -e 'config = { stack_mode = raw } kprobe:vfs_read { $s = kstack; print($s); exit() }'
```

All of these would have the default 'bpftrace'
stack mode instead of 'raw'.

##### Checklist

### DIFF
--- a/src/ast/passes/resource_analyser.h
+++ b/src/ast/passes/resource_analyser.h
@@ -21,7 +21,9 @@ namespace ast {
 // example the helper error metadata is still being collected during codegen.
 class ResourceAnalyser : public Visitor {
 public:
-  ResourceAnalyser(Node *root, std::ostream &out = std::cerr);
+  ResourceAnalyser(Node *root,
+                   BPFtrace &bpftrace,
+                   std::ostream &out = std::cerr);
 
   std::optional<RequiredResources> analyse();
 
@@ -38,6 +40,7 @@ private:
 
   RequiredResources resources_;
   Node *root_;
+  BPFtrace &bpftrace_;
   std::ostream &out_;
   std::ostringstream err_;
   // Current probe we're analysing
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -270,3 +270,15 @@ EXPECT_REGEX SUCCESS [0-9]+
 REQUIRES_FEATURE jiffies64
 TIMEOUT 5
 MIN_KERNEL 5.9
+
+NAME ustack builtin with stack_mode config
+RUN {{BPFTRACE}} -e 'config = { stack_mode=raw } uprobe:./testprogs/uprobe_test:uprobeFunction1 { @c[ustack] = 1; exit(); }' -p {{BEFORE_PID}}
+EXPECT_REGEX ^@c\[\n[0-9a-f]+$
+TIMEOUT 5
+BEFORE ./testprogs/uprobe_test
+
+NAME kstack builtin with stack_mode config
+RUN {{BPFTRACE}} -e 'config = { stack_mode=raw } k:do_nanosleep { @c[kstack] = 1; exit(); }'
+EXPECT_REGEX ^@c\[\n[0-9a-f]+$
+TIMEOUT 5
+AFTER ./testprogs/syscall nanosleep  1e8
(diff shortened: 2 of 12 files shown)


src/ast/passes/resource_analyser.h
```diff
@@ -21,7 +21,9 @@ namespace ast {
 // example the helper error metadata is still being collected during codegen.
 class ResourceAnalyser : public Visitor {
 public:
-  ResourceAnalyser(Node *root, std::ostream &out = std::cerr);
+  ResourceAnalyser(Node *root,
+                   BPFtrace &bpftrace,
```
viktormalik, line 25: Could we just pass `Config` here?

viktormalik (commented):
LGTM, just one comment below.

Also, I don't think that this fixes stack mode specifically for maps or map keys. It rather fixes stack mode for `kstack` and `ustack` used as a builtin variable (i.e. without `()`).

For instance, all of these are currently broken and should be fixed by this PR:
```
# bpftrace -e 'config = { stack_mode = raw } kprobe:vfs_read { @ = kstack; exit() }'
# bpftrace -e 'config = { stack_mode = raw } kprobe:vfs_read { @[kstack] = 0; exit() }'
# bpftrace -e 'config = { stack_mode = raw } kprobe:vfs_read { $s = kstack; print($s); exit() }'
```
Perhaps the tests and the commit message should be adjusted to reflect that.

Also this affects released v0.21.0 so a changelog entry would make sense here.

src/ast/passes/resource_analyser.h
jordalgo: I went with that initially but that proved to be more of a pain for some reason. Also semantic_analyser has a bpftrace member var so I figured I'd stick with that pattern.

jordalgo:
> Perhaps the tests and the commit message should be adjusted to reflect that. Also this affects released v0.21.0 so a changelog entry would make sense here.

Yes, good call. Will update both.

src/ast/passes/resource_analyser.h
viktormalik: Strange but ok. It's not that important that we should block the PR on it.

tests/runtime/config
```diff
@@ -10,6 +10,12 @@ EXPECT_REGEX ^\s+[0-9a-f]+$
 TIMEOUT 5
 BEFORE ./testprogs/uprobe_test
 
+NAME config takes precedence over the default
```
viktormalik, line 13: I'm wondering if we should also add tests specifically for `kstack`/`ustack` builtins to check that they respect the config setting (presumably to `tests/runtime/builtin`).

viktormalik (approved):

tests/runtime/config
jordalgo: Yeah, I wasn't sure where this test belonged TBH. I think you're right in that this is more specific to these builtins so I'll move the test there.

tests/runtime/builtin
```diff
@@ -270,3 +270,14 @@ EXPECT_REGEX SUCCESS [0-9]+
...
+NAME ustack builtin with stack_mode config
+RUN {{BPFTRACE}} -e 'config = { stack_mode=raw } uprobe:./testprogs/uprobe_test:uprobeFunction1 { @c[ustack] = 1; exit(); }' -p {{BEFORE_PID}}
+EXPECT_REGEX ^@c\[\n[0-9a-f]+$
+TIMEOUT 5
+BEFORE ./testprogs/uprobe_test
+
+NAME kstack builtin with stack_mode config
+RUN {{BPFTRACE}} -e 'config = { stack_mode=raw } k:do_nanosleep { @c[kstack] = 1; exit(); }'
+EXPECT_REGEX ^@c\[\n[0-9a-f]+$
+TIMEOUT 5
```
viktormalik, line 283: Missing after directive to make sure the kprobe hits:
```suggestion
TIMEOUT 5
AFTER ./testprogs/syscall nanosleep 1e8
```

jordalgo: Doh. Thank you.

PR #3323, opened 2024-07-17
CMake: Use try_compile in place of execute_process

ajor:
The previous approach was leaving compilation artefacts in the source directory.


viktormalik (approved):
Nice!

jordalgo:
Better!

jordalgo (approved):

PR #3326, opened 2024-07-17
Add new runtime test suite for aggregations

ajor:
Run aggregations against a multi-threaded target program and validate that the results are correct.

Work in progress.
- I haven't worked out how to use signed integers for min/max aggregations
- I'm seeing flakey results for `hist` and `stats`. Don't understand why, as they're using per-cpu (hash) maps.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -708,9 +708,11 @@ void TextOutput::map_key_val(const SizedType &map_type,
                              const std::string &key,
                              const std::string &val) const
 {
-  out_ << key + ": ";
+  out_ << key << ":";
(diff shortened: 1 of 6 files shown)


jordalgo:
Even though this program creates multiple threads, what guarantees do we have that wherever these tests are being run will run these threads on different cpus? Doesn't this get tricky in VM-like environments where there can be pinning and maybe just a few cpus allocated for the VM to run on? I'm not an expert - just thinking out loud.

jordalgo:
Closing due to age. Feel free to reopen.

PR #3332, opened 2024-07-20
Fix build with static libbfd and dependencies

barathrm:
Related to https://github.com/bpftrace/bpftrace/issues/3331

See discussion in issue ticket. Questions I have

- if `LIBBFD_FOUND` is set to true (as was the case in my failed build), but the compile check fails, should it not be set to false again/abort?
- is zstd usually a static library when libbfd is? it a shared lib in opensuse tumbleweed
- did `LIBSFRAME_FOUND` etc ever work? what has changed? would something break for someone if we switch to simply checking whether `LIBSFRAME_LIBRARIES` is set?
- why not use `pkg_check_modules` to look for zstd? on my system that would work, as it has a pkgconf file.


ajor:
It's CMake's [`find_package_handle_standard_args`](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html) function which is meant to set the `<Package>_FOUND` variable. Doesn't seem like it would have ever worked. Seems reasonable to swap out the `LIB*_FOUND` for `LIB*_LIBRARIES` to me.

I don't understand the whole interaction between LibBfd and the zlib, sframe, zstd libs though. If they're meant to be required dependencies then I think it would make sense to set `LIBBFD_FOUND` to false if they're not found.

I'm not sure that it's ok to change the `STATIC` into `SHARED`. From `FindLibBfd.cmake`, it looks like we only try using these libraries if we're doing a static build, so we'd want to import them as static libraries. cc @viktormalik as it looks like you worked on this area of the openSUSE build most recently.

danobi:
See pointer in https://github.com/bpftrace/bpftrace/issues/3331#issuecomment-2610133171

PR #3333, opened 2024-07-21
Use struct for Avg map value type

jordalgo:
Similar to how min/max works, instead of storing
two keys (one for total and one for count) for
each user key store a struct with the total and
count. This skips an additional map lookup and
allows users to loop over avg type maps.

This also fixes implicit casting of avg in
kernel space when the avg is a negative number.

##### Checklist

### DIFF
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -425,7 +425,6 @@ PROG BEGIN { print("BEGIN"); @["a"] = avg(10); @["b"] = avg(20); @["c"] = avg(30
 EXPECT BEGIN
        @[c]: 3
        @[d]: 4
-       
        END
 TIMEOUT 1
 
@@ -436,7 +435,6 @@ EXPECT BEGIN
        @[b]: 2
        @[c]: 3
        @[d]: 4
-       
        END
 TIMEOUT 1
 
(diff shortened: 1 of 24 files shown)


tests/runtime/json-output
```diff
@@ -156,12 +156,12 @@ AFTER ./testprogs/simple_struct
 
 NAME print_avg_map_args
 RUN {{BPFTRACE}} -q -f json -e 'BEGIN { @["a"] = avg(10); @["b"] = avg(20); @["c"] = avg(30); @["d"] = avg(40); print(@, 2, 10); clear(@); exit(); }'
-EXPECT {"type": "stats", "data": {"@": {"c": 3, "d": 4}}}
```
jordalgo, line 159: This would be a breaking change so I can change it back to stats. I guess I wasn't sure if 'avg' should really be considered 'stats' even though min/min are not.

tests/runtime/call
```diff
@@ -436,7 +435,6 @@ EXPECT BEGIN
        @[b]: 2
        @[c]: 3
        @[d]: 4
-
```
jordalgo, line 439: Because 'avg' is now on the same path as other maps, this extra space was removed. I can add it back but wasn't sure if I should.

PR #3334, opened 2024-07-22
adoc updates

jordalgo:
- remove line numbers from code samples
- add example usage for kaddr
- don't use short names for probes
- add short names to probes table
- set table of contents level to 1 so it's shorter

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1,6 +1,7 @@
 = bpftrace(8)
 :doctype: manpage
 :toc: true
+:toclevels: 1
 
 ////
 Style guide:
@@ -236,10 +237,9 @@ For example, listing the sleepers.bt file using `cat -n`:
 
 ----
 # cat -n sleepers.bt
-1 tracepoint:syscalls:sys_enter_nanosleep
-2 {
-3   printf("%s is sleeping.\n", comm);
-4 }
+tracepoint:syscalls:sys_enter_nanosleep {
+  printf("%s is sleeping.\n", comm);
+}
 ----
 
 And then calling it:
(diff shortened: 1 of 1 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -2664,61 +2671,75 @@ kprobe:tcp_reset,kprobe:*socket* {
...
 |*Description*
 |*Kernel/User Level*
 
 | <<probes-begin-end, `BEGIN/END`>>
+| -
 | Built-in events
 | Kernel/User
 
 | <<probes-hardware, `hardware`>>
+| h
```
ajor, line 2687:
```suggestion
| `h`
```
I think all the short names would look better in monospace

ajor (commented):
After seeing the line numbers removed, I'm now seeing the reason that they were probably included in the first place - it's to distinguish the output of `cat` from commands typed into the terminal.

e.g. with this one the shebang could be confused with the bash prompt:
```
# cat -n sleepers.bt
#!/usr/local/bin/bpftrace

tracepoint:syscalls:sys_enter_nanosleep {
  printf("%s is sleeping.\n", comm);
}
```

I don't have any better ideas or really feel strongly about it though.

If you do want to remove the line numbers, you'll need to remove the `-n` flag from the `cat` invocations as well, as that's what causes the numbers to be printed.

jordalgo:
> you'll need to remove the -n flag from the cat invocations as well

Ah good catch. Will remove the '-n'. 

> it's to distinguish the output of cat from commands typed into the terminal

Yeah, true. I think I can just add some spacing between the two code blocks, which should help. Let me see how that looks.

jordalgo:
Re-opened here: https://github.com/bpftrace/bpftrace/pull/3337

PR #3336, opened 2024-07-22
Fix: parsing unsigned 64 bit positional params

jordalgo:
Previously either of these would fail
```
$ bpftrace -e 'BEGIN {printf("%lu", $1);}' 0xffffffffffffffff
$ bpftrace -e 'BEGIN {printf("%lu", $1);}' "18446744073709551615"
```

Because we were only using `stoll` which expects signed ints. This change falls back to using `stoull` and changes the type accordingly if `stoll` fails.

Issue:
https://github.com/bpftrace/bpftrace/issues/3335

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -2050,13 +2050,17 @@ std::optional<int64_t> BPFtrace::get_int_literal(
                  expr)) {
       if (pos_param->ptype == PositionalParameterType::positional) {
         auto param_str = get_param(pos_param->n, false);
-        if (is_numeric(param_str))
-          return std::stoll(param_str);
-        else {
+        auto param_int = get_int_from_str(param_str);
+        if (!param_int.has_value()) {
           LOG(ERROR, pos_param->loc)
               << "$" << pos_param->n << " used numerically but given \""
               << param_str << "\"";
           return std::nullopt;
+        }
+        if (std::holds_alternative<int64_t>(*param_int)) {
+          return std::get<int64_t>(*param_int);
+        } else {
+          return (int64_t)std::get<uint64_t>(*param_int);
         }
       } else
         return (int64_t)num_params();
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -1077,15 +1077,47 @@ std::string str_join(const std::vector<std::string> &list,
   return str;
 }
 
-bool is_numeric(const std::string &s)
+std::optional<std::variant<int64_t, uint64_t>> get_int_from_str(
+    const std::string &s)
 {
-  std::size_t idx;
-  try {
-    std::stoll(s, &idx, 0);
-  } catch (...) {
-    return false;
+  if (s.size() == 0) {
+    return std::nullopt;
   }
-  return idx == s.size();
+
+  if (s.starts_with("0x") || s.starts_with("0X")) {
+    // Treat all hex's as unsigned
+    std::size_t idx;
+    try {
+      uint64_t ret = std::stoull(s, &idx, 0);
+      if (idx == s.size()) {
+        return ret;
+      } else {
+        return std::nullopt;
+      }
+    } catch (...) {
+      return std::nullopt;
+    }
+  }
+
+  char *endptr;
+  const char *s_ptr = s.c_str();
+  errno = 0;
+
+  if (s.at(0) == '-') {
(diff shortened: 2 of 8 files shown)


src/utils.cpp
```diff
@@ -1077,13 +1077,19 @@ std::string str_join(const std::vector<std::string> &list,
   return str;
 }
 
-bool is_numeric(const std::string &s)
+bool is_numeric(const std::string &s, bool &is_signed)
```
ajor, line 1080: Catching exceptions can be horribly slow so we should try to avoid their use in happy-path parsing. This particular function has been on my TODO list for a long time but I've never gotten around to changing it.

jordalgo: We could use a regex to determine if the string is a hex and also look at the first character to see if it's a "-" to see if we need a signed int.

Also, I realized I need to fix some of the types for `get_int_literal`.

All that being said, I feel like perf improvements to functions not called multiple times in a hot path shouldn't block us from fixing bugs like this. We can always do follow-up PRs for speed-ups. I'll get off my soap box now.

jordalgo: Ok, I made some changes. It's a more complicated version so TBH I'm not sure if I made it better but at least it doesn't rely on exceptions to determine if the string is signed or not. Let me know what you think.

I didn't tackle the fact that `get_int_literal` only returns int64_t - mostly because it was more complicated and I wasn't sure if we want to do something similar to what I have now for `get_int_from_str` where I return a variant. Also didn't want to bloat the PR.

danobi: > All that being said, I feel like perf improvements to functions not called multiple times in a hot path shouldn't block us from fixing bugs like this. We can always do follow-up PRs for speed-ups. I'll get off my soap box now.

I agree with this. If it's not being called many times I'd prioritize correctness and maintainability instead. Also, I kinda wonder if running a regex is faster than catching an exception. At least we can make the common case not throw - then there is no cost at all.

src/utils.cpp
```diff
@@ -1077,15 +1077,49 @@ std::string str_join(const std::vector<std::string> &list,
...
+    }
+  }
+
+  char *endptr;
+  errno = 0;
+
+  if (s.at(0) == '-') {
+    int64_t ret = strtol(s.c_str(), &endptr, 0);
+    if (endptr == s.c_str() || *endptr != '\0' || errno == ERANGE ||
+        errno == EINVAL) {
```
danobi, line 1111: Maybe defensively save a copy of `errno` incase `c_str()` clobbers it? It probably doens't, but who knows

src/bpftrace.cpp
```diff
@@ -2050,14 +2050,18 @@ std::optional<int64_t> BPFtrace::get_int_literal(
...
+        if (!param_int.has_value()) {
           LOG(ERROR, pos_param->loc)
               << "$" << pos_param->n << " used numerically but given \""
               << param_str << "\"";
           return std::nullopt;
         }
+        if (std::holds_alternative<int64_t>(*param_int)) {
+          return std::get<int64_t>(*param_int);
+        } else {
+          return (int64_t)std::get<uint64_t>(*param_int);
```
danobi, line 2063: Should probably check for overflow here, right?

src/utils.cpp
```diff
@@ -1077,15 +1077,49 @@ std::string str_join(const std::vector<std::string> &list,
...
-    std::stoll(s, &idx, 0);
-  } catch (...) {
-    return false;
+  if (s.size() == 0) {
+    return std::nullopt;
+  }
+
+  static const auto hex_regex = std::regex("0[xX][0-9a-fA-F]+");
+
+  if (std::regex_match(s, hex_regex)) {
```
danobi, line 1089: Not sure regex is necessarily faster than an exception. Is it sufficient to check for `0x` or `0X` prefix?

jordalgo: > Is it sufficient to check for 0x or 0X prefix?

Yeah, good idea. I'll just do that.

src/bpftrace.cpp
jordalgo: I'm gonna tackle all that needs fixing in this function in a follow-up PR.

src/utils.cpp
jordalgo: I'll just call `c_str()` before resetting errno and save the ptr.

danobi (approved):

PR #3338, opened 2024-07-22
Use struct for Avg/Stats map value types

jordalgo:
Similar to how min/max works, instead of storing
two keys (one for total and one for count) for
each user key store a struct with the total and
count. This skips an additional map lookup and
allows users to loop over avg type maps.

This also fixes implicit casting of avg in
kernel space when the avg is a negative number.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -316,6 +316,11 @@ llvm::Type *IRBuilderBPF::GetMapValueType(const SizedType &stype)
     // The second field is the "value is set" flag
     std::vector<llvm::Type *> llvm_elems = { getInt64Ty(), getInt64Ty() };
     ty = GetStructType("min_max_val", llvm_elems, false);
+  } else if (stype.IsAvgTy() || stype.IsStatsTy()) {
+    // The first field is the total value
+    // The second is the count value
+    std::vector<llvm::Type *> llvm_elems = { getInt64Ty(), getInt64Ty() };
+    ty = GetStructType("avg_stas_val", llvm_elems, false);
   } else {
     ty = GetType(stype);
   }
@@ -605,13 +610,13 @@ Value *IRBuilderBPF::CreatePerCpuMapAggElems(Value *ctx,
   const std::string &map_name = map.ident;
 
   AllocaInst *i = CreateAllocaBPF(getInt32Ty(), "i");
-  AllocaInst *ret = CreateAllocaBPF(getInt64Ty(), "ret");
-  // used only for min/max
-  AllocaInst *is_ret_set = CreateAllocaBPF(getInt64Ty(), "is_ret_set");
+  AllocaInst *val_1 = CreateAllocaBPF(getInt64Ty(), "val_1");
+  // used for min/max/avg
+  AllocaInst *val_2 = CreateAllocaBPF(getInt64Ty(), "val_2");
 
   CreateStore(getInt32(0), i);
-  CreateStore(getInt64(0), ret);
-  CreateStore(getInt64(0), is_ret_set);
+  CreateStore(getInt64(0), val_1);
+  CreateStore(getInt64(0), val_2);
 
   Function *parent = GetInsertBlock()->getParent();
   BasicBlock *while_cond = BasicBlock::Create(module_.getContext(),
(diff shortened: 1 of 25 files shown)


tests/runtime/call
```diff
@@ -425,7 +425,6 @@ PROG BEGIN { print("BEGIN"); @["a"] = avg(10); @["b"] = avg(20); @["c"] = avg(30
 EXPECT BEGIN
        @[c]: 3
        @[d]: 4
-
```
jordalgo, line 428: Because 'avg' is now on the same path as other maps, this extra space was removed. I can add it back but wasn't sure if I should.

tests/runtime/json-output
```diff
@@ -156,12 +156,12 @@ AFTER ./testprogs/simple_struct
 
 NAME print_avg_map_args
 RUN {{BPFTRACE}} -q -f json -e 'BEGIN { @["a"] = avg(10); @["b"] = avg(20); @["c"] = avg(30); @["d"] = avg(40); print(@, 2, 10); clear(@); exit(); }'
-EXPECT {"type": "stats", "data": {"@": {"c": 3, "d": 4}}}
```
jordalgo, line 159: This would be a breaking change so I can change it back to stats. I guess I wasn't sure if 'avg' should really be considered 'stats' even though min/min are not.

ajor (commented):
Just as a high level comment: if we're changing `avg` then should we be changing `stats` at the same time? They currently share an implementation and this PR splits them so we've got two different versions.

(stats is just an average that also prints out the hidden `sum` and `count` variables)

jordalgo:
> Just as a high level comment: if we're changing avg then should we be changing stats at the same time? They currently share an implementation and this PR splits them so we've got two different versions.

Ah I didn't realize they were so close. Yes, I'll just update `stats` so it uses the same struct as `avg`. I'll also switch 'avg' back to printing out 'stats' as the map type just to not be a breaking change.

jordalgo:
@ajor Gentle ping

src/ast/irbuilderbpf.cpp
```diff
@@ -700,11 +707,60 @@ Value *IRBuilderBPF::CreatePerCpuMapAggElems(Value *ctx,
...
-  CreateLifetimeEnd(is_ret_set);
 
-  Value *ret_reg = CreateLoad(getInt64Ty(), ret);
+  Value *ret_reg;
+
+  if (type.IsAvgTy()) {
+    AllocaInst *ret = CreateAllocaBPF(getInt64Ty(), "ret");
+    // BPF doesn't yet support a signed division so we have to check if
+    // the value is negative, flip it, do an unsigned division, and then
+    // flip it back
```
ajor, line 717: 😭 thanks BPF

(nice workaround though)

src/output.cpp
```diff
@@ -475,20 +478,22 @@ void Output::map_stats_contents(
...
-  for (auto &key_count : total_counts_by_key) {
-    auto &key = key_count.first;
-    auto &value = values_by_key.at(key);
 
-    if (map_type.IsAvgTy() && top && values_by_key.size() > top &&
-        i++ < (values_by_key.size() - top))
-      continue;
+  for (auto &pair : values_by_key) {
+    auto key = pair.first;
+    auto value = pair.second;
```
ajor, line 491:
```suggestion
  for (auto &[key, value] : values_by_key) {
```
Can be shortened to a single line. Also I think the current code is making copies of the `key` and `value` vectors (`auto` instead of `auto&`)

src/output.cpp
```diff
@@ -497,21 +502,38 @@ void Output::map_stats_contents(
...
+      int64_t avg = stats_value<int64_t>(value, bpftrace.ncpus_, total, count);
+      avg_str = std::to_string(map_type.IsStatsTy() ? avg : avg / div);
+      total_str = std::to_string(total);
+      count_str = std::to_string(count);
+    } else {
+      uint64_t total = 0;
+      uint64_t count = 0;
+      uint64_t avg = stats_value<uint64_t>(
+          value, bpftrace.ncpus_, total, count);
+      avg_str = std::to_string(map_type.IsStatsTy() ? avg : avg / div);
```
ajor, line 521: Not your code (and I'm not suggesting you change it), but why on earth do we have to deal with this conditional division during output?? It should have been sorted out way earlier (probably semantic analysis) so that these later stages don't have to care 😞

src/bpftrace.cpp
```diff
@@ -1477,73 +1497,6 @@ int BPFtrace::print_map_hist(const BpfMap &map, uint32_t top, uint32_t div)
...
-      // New key - create a list of buckets for it
-      values_by_key[key_prefix] = std::vector<int64_t>(2);
-    }
-    values_by_key[key_prefix].at(bucket) = reduce_value<int64_t>(value,
-                                                                 nvalues);
-
-    old_key = key;
-  }
-
-  // Sort based on sum of counts in all buckets
```
ajor, line 1524: I spent way too long trying to understand how this related to the code before I realised the comment was just blindly copied from `print_map_hist()` 🤦

src/output.cpp
jordalgo: We probably don't as div should always be `1` for the stats type and you're right semantic analyser should catch this. I'll remove.

ajor (approved):
Nice!

PR #3340, opened 2024-07-22
appimage: Strip header files out of appimage

danobi:
Some header files libclang ships are quite huge (megabytes). Looks like they're for ARM. Regardless, we don't need any header files at runtime.

This saves us about 1M of disk space.

##### Checklist


jordalgo (approved):
Nice!

PR #3341, opened 2024-07-24
Remove useless a.out ELF file and ignore *.out

Rtoax:
In commit 32dc9e7b9fd2 ("appimage: Fix lldb in appimage builds") introduce an a.out ELF file by mistake.


jordalgo:
@Rtoax Good catch! Do we know if this a.out gets re-generated when doing the app image build? (cc @danobi)

ajor:
I'm tempted to do a force-push to get the binary file out of the repo's history. What do people think?

jordalgo:
> I'm tempted to do a force-push to get the binary file out of the repo's history. What do people think?

I don't feel strongly but I'm not opposed. I don't imagine this binary has some malware in it. I just want to make sure we didn't introduce something where the binary will return.

ajor:
I'll clean it up in a force push now then. The sooner the better to avoid more people pulling the current version

ajor:
Thanks for spotting this @Rtoax!

PR #3343, opened 2024-07-24
ci: Backstop AOT tests

danobi:
This PR backstops AOT tests in CI. Not all the AOT test pass or get
skipped yet. Rather than wait for it to happen and then add in CI,
let's just add a subset in CI now to prevent any more backsliding.

We've already had quite a few regressions due to the codepaths
not being exercised in CI.

##### Checklist

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -38,6 +38,7 @@ NIX_TARGET = os.environ.get("NIX_TARGET", "")
 CMAKE_BUILD_TYPE = os.environ.get("CMAKE_BUILD_TYPE", "Release")
 RUN_TESTS = os.environ.get("RUN_TESTS", "1")
 RUN_MEMLEAK_TEST = os.environ.get("RUN_MEMLEAK_TEST", "0")
+RUN_AOT_TESTS = os.environ.get("RUN_AOT_TESTS", "0")
 CC = os.environ.get("CC", "cc")
 CXX = os.environ.get("CXX", "c++")
 GTEST_COLOR = os.environ.get("GTEST_COLOR", "auto")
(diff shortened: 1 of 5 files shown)


.github/include/ci.py
```diff
@@ -274,6 +276,34 @@ def test():
...
+                        "./tests/runtime-tests.sh",
+                        "--run-aot-tests",
+                        "--filter",
+                        "aot.*",
+                    ]
+                    + [
+                        "--skiplist_file",
+                        f"{root()}/{AOT_SKIPLIST_FILE}",
+                    ]
+                    if AOT_SKIPLIST_FILE
```
jordalgo, line 295: nit: why do we need this check here? Are there cases when it's ok if CI doesn't find this file even though we specify it.

tests/runtime/engine/runner.py
```diff
@@ -93,6 +95,8 @@ def skip_reason(test, status):
             return "disabled by environment variable"
         elif status == Runner.SKIP_AOT_NOT_SUPPORTED:
             return "aot does not yet support this"
+        elif status == Runner.SKIP_IN_SKIPLIST:
+            return "disabled by entry in skiplist"
```
jordalgo, line 99: super nit

Would also be cool to print the file name or path but maybe that's too verbose
```suggestion
            return "disabled by entry in skiplist file"
```

jordalgo (approved):
Minor question and nit but otherwise LGTM

jordalgo:
Side note: In terms of test output, should we just show the number of tests skipped at the end instead of re-listing them all?

danobi:
> Side note: In terms of test output, should we just show the number of tests skipped at the end instead of re-listing them all?

My personal preference would be to keep it as-is. I find the summary helpful - esp for when I build the skiplist which was just copy paste

PR #3347, opened 2024-07-25
AST: Delete unused Mutator class

ajor:
As it's unused and untested, it's been rotting for a few years and does not correctly do the job it claims to do.

Additionally, it is getting in the way of refactoring the AST away from manual memory management.

There is value in having a Mutator class, but due to bugs and upcoming refactors it will be easier to write it again from scratch.


jordalgo (approved):
Okey dokey.

PR #3350, opened 2024-07-26
AST: Switch to non-owning pointers in nodes

ajor:
Nodes are now owned by the ASTContext class.

This lets us get rid of the custom node destructors and eliminates a source of memory leaks (not all destructors were complete).

AST lists are embedded directly in the nodes that own them, but the pointers-to-lists are maintained in this commit for API compatibility. The pointers will be removed entirely in a separate, API-breaking, commit. Changing them adds a lot of churn (especially in semantic_analyser.cpp) so I've left them out of this PR for clarity.

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -121,7 +121,6 @@ public:
   explicit PositionalParameter(PositionalParameterType ptype,
                                long n,
                                location loc);
-  ~PositionalParameter() = default;
 
   PositionalParameterType ptype;
   long n;
@@ -136,7 +135,6 @@ public:
   DEFINE_ACCEPT
 
   explicit String(const std::string &str, location loc);
-  ~String() = default;
 
   std::string str;
 
@@ -149,7 +147,6 @@ public:
   DEFINE_ACCEPT
 
   explicit StackMode(const std::string &mode, location loc);
-  ~StackMode() = default;
 
   std::string mode;
 
@@ -162,7 +159,6 @@ public:
   DEFINE_ACCEPT
 
   explicit Identifier(const std::string &ident, location loc);
-  ~Identifier() = default;
 
   std::string ident;
 
@@ -175,7 +171,6 @@ public:
   DEFINE_ACCEPT
 
   explicit Builtin(const std::string &ident, location loc);
-  ~Builtin() = default;
 
   std::string ident;
   int probe_id;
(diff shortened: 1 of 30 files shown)


ajor:
Oh the irony - the memleak tests are complaining about my code. I missed a few places where `new` was called directly:
- https://github.com/bpftrace/bpftrace/blob/e44032809e01e548401fe09cd1d4b016d24833f2/src/ast/attachpoint_parser.cpp#L169
- https://github.com/bpftrace/bpftrace/blob/e44032809e01e548401fe09cd1d4b016d24833f2/src/ast/passes/semantic_analyser.cpp#L555
- https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L1458

I'll try to sort them out.

src/ast/ast.h
```diff
@@ -719,6 +691,35 @@ std::string opstr(const Jump &jump);
 
 SizedType ident_to_record(const std::string &ident, int pointer_level = 0);
 
+template <typename T>
+concept NodeType = std::derived_from<T, Node>;
```
danobi, line 695: :O

src/ast/ast.h
```diff
@@ -719,6 +691,35 @@ std::string opstr(const Jump &jump);
...
+
+/*
+ * Manages the lifetime of AST nodes.
+ *
+ * Nodes allocated by an ASTContext will be kept alive for the duration of the
+ * owning ASTContext object.
+ */
```
danobi, line 702: Neat idea. Basically an arena, right?

danobi (approved):
PR was a bit too large to do a detailed review, but looking over it seems mostly mechanical. test cases remain intact and still pass so that's great

src/ast/ast.h
ajor: Yep, a poor man's arena for now :P 
One day we might want to move it over to a more efficient allocation scheme

src/main.cpp
```diff
@@ -361,26 +361,26 @@ static void parse_env(BPFtrace& bpftrace)
 
   err = driver.parse();
   if (err)
-    return nullptr;
+    return {};
```
jordalgo, line 364: nit: I prefer returning `std::nullopt` in these cases.

jordalgo (commented):
Yay for removing `new`!

PR #3354, opened 2024-07-29
AST: Finish the refactoring of Lists from pointers to inline members

ajor:
Follow up to #3350

```
1. Remove the pointer-to-vector members
2. Rename the inline vector members to the pointers' old names
3. Update users of the AST to handle std::vector instead of pointers
  - This is mainly just replacing "->" with "."
  - The bigger changes come in places where we used to check for null
    pointers. Empty vectors are now used in place of null pointers, so
    these checks are not necessary before loops.
```

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -86,17 +86,16 @@ PositionalParameter::PositionalParameter(PositionalParameterType ptype,
 }
 
 Call::Call(const std::string &func, location loc)
-    : Expression(loc), func(is_deprecated(func)), vargs(nullptr)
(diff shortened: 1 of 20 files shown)


ajor:
Just one more PR to come after this one (replacing pointers which can never be null with references (or equivalents)

jordalgo (approved):
Did a surface glance. Mostly relying on our tests for this one.

danobi (approved):

jordalgo:
@ajor No rush or presh but when you fix the merge conflict and merge this, I can put up the PR for the lexical scope fix.

ajor:
Go go go! :D

PR #3355, opened 2024-07-29
SemanticAnalyser: Only run check_arg() if check_nargs() has passed

ajor:
It is expected that `check_arg()` is only called after the number of arguments have passed validation with `check_nargs().` Due to a quirk of the implementation, it wasn't necessary to do this when calling `check_arg()` for argument 0. However, this quirk will no longer exist after the ongoing AST refactoring work.

Add `[[nodiscard]]` to `check_nargs()` and `check_varargs()` to try to prevent this validation being missed again.

This PR is necessary for the refactoring in #3354.


jordalgo (approved):
LGTM

PR #3356, opened 2024-07-29
AttachedProbe: Remove references to load_prog()

ajor:
The load_prog() function was deleted in 0c3a9492ab8d2b2e0dc9d16802a97e88dd8f1b13.

I don't really understand the purpose of the code I'm deleting here so this might not be the right "fix". It was introduced in 5a84f46212aeffa4223a721eee3c2b6988163ab8 to improve error messages. If it's still needed, should we be checking `progfd_ < 0` for all probe types?

@viktormalik probably has the most context.

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -95,9 +95,6 @@ std::string progtypeName(libbpf::bpf_prog_type t)
 
 void AttachedProbe::attach_kfunc(void)
 {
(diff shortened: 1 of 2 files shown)


viktormalik (approved):
This is a correct change. These situations should never happen now as program loading should fail if there are any `progfd_ < 0`.

ajor:
Thanks!

PR #3358, opened 2024-07-30
Add 'has_key' builtin for maps

jordalgo:
This is so we can check if a map has a key
instead of relying on the wrong way by seeing
if the value is 0.

```
kprobe:dummy {
  @scalar = 1;
  @associative[1,2] = 1;

  if (has_key(@scalar)) {
    print(("hello"));
  }

  if (!has_key(@associative[1,3])) {
    print(("bye"));
  }
}
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to
   - [#3285](https://github.com/bpftrace/bpftrace/pull/3285)
 - For-loops: Allow sharing variables between the main probe and the loop's body
   - [#3014](https://github.com/bpftrace/bpftrace/pull/3014)
+- Add `has_key` builtin for maps
+  - [#1111](https://github.com/bpftrace/bpftrace/pull/1111)
 #### Changed
 - Stream output when printing maps
   - [#3264](https://github.com/bpftrace/bpftrace/pull/3264)
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -771,6 +771,51 @@ void CodegenLLVM::visit(Call &call)
       }
     }
     expr_ = nullptr;
+  } else if (call.func == "has_key") {
+    auto &arg = *call.vargs->at(0);
+    auto &map = static_cast<Map &>(arg);
+    auto [key, scoped_key_deleter] = getMapKey(map);
+
+    CallInst *lookup = b_.CreateMapLookup(map, key);
+
+    AllocaInst *has_key = b_.CreateAllocaBPF(b_.getInt8Ty(), "has_key");
+
+    Function *parent = b_.GetInsertBlock()->getParent();
+    BasicBlock *lookup_success_block = BasicBlock::Create(module_->getContext(),
+                                                          "lookup_success",
+                                                          parent);
+    BasicBlock *lookup_failure_block = BasicBlock::Create(module_->getContext(),
+                                                          "lookup_failure",
+                                                          parent);
+    BasicBlock *lookup_merge_block = BasicBlock::Create(module_->getContext(),
+                                                        "lookup_merge",
+                                                        parent);
+
+    Value *lookup_condition = b_.CreateICmpNE(
+        b_.CreateIntCast(lookup, b_.GET_PTR_TY(), true),
+        b_.GetNull(),
+        "lookup_cond");
+    b_.CreateCondBr(lookup_condition,
+                    lookup_success_block,
+                    lookup_failure_block);
+
+    b_.SetInsertPoint(lookup_success_block);
+
+    b_.CreateStore(b_.getInt8(1), has_key);
+
+    b_.CreateBr(lookup_merge_block);
+
+    b_.SetInsertPoint(lookup_failure_block);
+
+    b_.CreateStore(b_.getInt8(0), has_key);
+
+    b_.CreateBr(lookup_merge_block);
+
+    b_.SetInsertPoint(lookup_merge_block);
+
+    expr_ = b_.CreateLoad(b_.getInt8Ty(), has_key);
+
+    b_.CreateLifetimeEnd(has_key);
   } else if (call.func == "str") {
     uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
     // Largest read we'll allow = our global string buffer size
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -2347,6 +2347,10 @@ Functions that are marked *async* are asynchronous which can lead to unexpected 
 | Delete a single key from a map. For a single value map this deletes the only element. For an associative-array the key to delete has to be specified. Multiple arguments can be passed to delete many keys at once.
 | Sync
 
+| <<map-functions-has_key, `has_key(mapkey k)`>>
+| Return true if the key exists in this map. Otherwise return false. For a single value map this returns true. For an associative-array the key has to be specified.
+| Sync
+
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to
   - [#3285](https://github.com/bpftrace/bpftrace/pull/3285)
 - For-loops: Allow sharing variables between the main probe and the loop's body
   - [#3014](https://github.com/bpftrace/bpftrace/pull/3014)
+- Add `has_key` builtin for maps
+  - [#1111](https://github.com/bpftrace/bpftrace/pull/1111)
```
jordalgo, line 21: will update this before merging

src/ast/passes/codegen_llvm.cpp
```diff
@@ -771,6 +771,51 @@ void CodegenLLVM::visit(Call &call)
...
+                                                          "lookup_success",
+                                                          parent);
+    BasicBlock *lookup_failure_block = BasicBlock::Create(module_->getContext(),
+                                                          "lookup_failure",
+                                                          parent);
+    BasicBlock *lookup_merge_block = BasicBlock::Create(module_->getContext(),
+                                                        "lookup_merge",
+                                                        parent);
+
+    Value *lookup_condition = b_.CreateICmpNE(
```
fbs, line 794: Its been a while but do we need to branch here? Cant we just type cast the Icmp result into an i8? LLVM probably optimizes it (dont have the setup to check atm) but it saves some code.

fbs: something like `expr_ = b_.CreateZext(lookup_condition, b_.getInt8Ty())` i believe

ajor:
This is definitely functionality that we need, but I'm not sold on the syntax.

The name `has_key` doesn't make sense on a scalar - a scalar by definition will never have a key. Is some kind of has-value / is-set functionality needed for scalars in the first place? A scalar map is just like a global variable in other languages and they don't have any way of determining if the value has been updated from the default value of 0.

`has_key(@associative[1,3])` - this is replicating the `delete()` syntax, but I kind of consider that a bad design choice too (although one that we have to keep for compatibility). It looks as if it's looking up a value in the `@associative` map and then passing that value to the `has_key` function.

I did a quick survey of how some other systems programming languages handle maps and there seem to be three main approaches.

C: `has_key(mymap, mykey)`

Languages using methods on maps:
C++ `mymap.contains(mykey)`
Rust: `mymap.contains_key(mykey)`
Nim: `mymap.hasKey(mykey)`

Languages which return some sort of optional value on looking up a key:
Zig: `if (mymap.get(mykey)) |v| { got value v } else { key doesn't exist }`
Swift: `if let v = dict[key] { got value v } else { key doesn't exist }`
Go: `value, exists := mymap[mykey]`

I'd favour going with the "map methods" approach. I don't think the C method is very ergonomic (backed up by the fact that no other modern language has chosen it). The optional method isn't bad, but might involve changing the return value of our map-lookups, which would break backwards compatibility. The "map methods" can be a pure language extension without breaking existing behaviour.

We'll also want to add some more functions to operate on maps in the future too, e.g. "insert_if_doesnt_exist" and "update_if_already_exists", to expose more features to the user and make BpfScript capable of expressing more complex programs.

src/ast/passes/codegen_llvm.cpp
jordalgo: Yeah, good call. Let me look into that.

jordalgo:
> I'd favour going with the "map methods" approach. 

In general, I'm a big fan of methods on objects BUT that's a big language change for bpftrace. If we added something like `@my_map.contains(mykey)` I think we'd also need to provide methods for current functions like `len`, `clear`, `delete` and `zero` - and perhaps eventually deprecate the current map functions that take a map as the first arg. If there is consensus that we should go that direction, then I'm happy to work on that change. That being said, I think that providing a `has_key(mymap, mykey)` (C-like) builtin would be valuable for the time-being, as it aligns with the current map functions - then perhaps we can deprecate all at once (after we add map methods).

> The name has_key doesn't make sense on a scalar

I agree. Perhaps I can add a semantic_analyser check to not allow scalar maps.

> We'll also want to add some more functions to operate on maps in the future too, e.g. "insert_if_doesnt_exist" and "update_if_already_exists", to expose more features to the user and make BpfScript capable of expressing more complex programs.

@ajor - I feel like I'm channeling you when I say this but I think those additional methods just add language bloat and won't be necessary once we land a `has_key` method/function as folks can just use `has_key` themselves before they insert or update.

ajor:
> In general, I'm a big fan of methods on objects BUT that's a big language change for bpftrace. If we added something like `@my_map.contains(mykey)` I think we'd also need to provide methods for current functions like `len`, `clear`, `delete` and `zero` - and perhaps eventually deprecate the current map functions that take a map as the first arg. If there is consensus that we should go that direction, then I'm happy to work on that change. That being said, I think that providing a `has_key(mymap, mykey)` (C-like) builtin would be valuable for the time-being, as it aligns with the current map functions - then perhaps we can deprecate all at once (after we add map methods).

Agreed that we'd want to provide methods for the other existing map functions too. I think we should sort out the syntax before merging new functions though, as once they're in, they're pretty much in forever. A C-like function would also be somewhat out-of-place, given that the `delete()` function's syntax is so unique and we don't have any other function which accepts a map key as a separate argument.


> @ajor - I feel like I'm channeling you when I say this but I think those additional methods just add language bloat and won't be necessary once we land a `has_key` method/function as folks can just use `has_key` themselves before they insert or update.

Haha fair point. I was just thinking about exposing the functionality that the underlying BPF maps already provide - I guess the advantage of these functions would be efficiency (1 function call instead of 2), but they're not essential.

jordalgo:
> I think we should sort out the syntax before merging new functions though, as once they're in, they're pretty much in forever.

I'll create an RFC issue with some options though I think `@.method(...)` seems like the obvious choice. But just curious, why is stuff "in forever" - breaking changes are a thing right? They just require a major version bump though since we're not even at version 1 yet, the API is not considered stable and according to the semver docs: "Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable." I know this isn't a good user experience but I think some warnings for deprecated functions followed by removal is legit.

> I guess the advantage of these functions would be efficiency (1 function call instead of 2),

Yeah true. I guess we can add as requested/needed.

ajor:
Ok not really forever, I just meant it in the sense that while theoretically we can (and I'd say should!) deprecate and remove some things, in practice we've very rarely done it.

jordalgo:
Methods RFC: https://github.com/bpftrace/bpftrace/issues/3361

> I just meant it in the sense that while theoretically we can (and I'd say should!) deprecate and remove some things, in practice we've very rarely done it.

Let's start making some issues! I'm happy to start deprecating things - I love deleting 🔥

danobi:
> The name has_key doesn't make sense on a scalar - a scalar by definition will never have a key. Is some kind of has-value / is-set functionality needed for scalars in the first place? A scalar map is just like a global variable in other languages and they don't have any way of determining if the value has been updated from the default value of 0.

Agreed - not sure what `has_key()` would do on a scalar given that they're now backed by an array map.

danobi:
If we decide to introduce new syntax for map methods, then the most pragmatic way would be to introduce new functionality and deprecate the old functionality over some number of releases. It sounds like we are in agreement over that.

So I don't see the harm in introducing a potentially "old style" `has_key()` consistent with the design of `delete()`. Better to solve real user problems, be consistent, and then maybe deprecate later IMO.

ajor:
For anyone coming here from the changelog entry: `has_key` was implemented in PR #3485

PR #3359, opened 2024-07-31
Disable liblldb scripting

ttreyer:
LLDB can be customised (with its .lldbinit file) and extended with scripting (typically in Python).
These files are not necessarily write-protected, so an adversary could exploit them to alter the generated BPF byte-code.

We don't rely on Python extension in our use of LLDB. Ideally we would link against a liblldb with no scripting support, but we can't control what features distributions enable in their `liblldb-devel` package. So this PR is an attempt at mitigating this risk by documenting the issue and disabling the scripting engine.

##### Checklist

### DIFF
--- a/src/dwarf_parser.cpp
+++ b/src/dwarf_parser.cpp
@@ -22,11 +22,15 @@ Dwarf::Dwarf(BPFtrace *bpftrace, std::string file_path)
   if (instance_count++ == 0)
     lldb::SBDebugger::Initialize();
 
-  debugger_ = lldb::SBDebugger::Create(false);
+  // Create a "hardened" debugger instance with no scripting, nor .lldbinit.
+  // We don't want a Python extension or .lldbinit to influence the byte-code
+  // that will get executed by the Kernel. It would be a security risk!
+  debugger_ = lldb::SBDebugger::Create(/* source_init_file = */ false);
+  debugger_.SetScriptLanguage(lldb::ScriptLanguage::eScriptLanguageNone);
 
   lldb::SBError error;
   target_ = debugger_.CreateTarget(
-      file_path_.c_str(), nullptr, nullptr, false, error);
+      file_path_.c_str(), nullptr, nullptr, true, error);
   if (!error.Success() || !target_.IsValid()) {
     throw error;
   }


src/dwarf_parser.cpp
```diff
@@ -22,11 +22,15 @@ Dwarf::Dwarf(BPFtrace *bpftrace, std::string file_path)
...
+  // Create a "hardened" debugger instance with no scripting, nor .lldbinit.
+  // We don't want a Python extension or .lldbinit to influence the byte-code
+  // that will get executed by the Kernel. It would be a security risk!
+  debugger_ = lldb::SBDebugger::Create(/* source_init_file = */ false);
+  debugger_.SetScriptLanguage(lldb::ScriptLanguage::eScriptLanguageNone);
 
   lldb::SBError error;
   target_ = debugger_.CreateTarget(
-      file_path_.c_str(), nullptr, nullptr, false, error);
+      file_path_.c_str(), nullptr, nullptr, true, error);
```
ajor, line 33: What's the meaning of this `add_dependent_modules` attribute you've set to true?

ref: https://lldb.llvm.org/cpp_reference/classlldb_1_1SBDebugger.html#a0e423c608ce554a133d8611578da5f7f

ttreyer: This will prompt liblldb to load debug information of the shared libraries, alongside the main binary DebugInfo.

ajor: Thinking about this some more, do we definitely need to load debug info from shared libraries?

Won't types defined in libraries also be represented in the debug info of the main executable, since it included the libraries' headers?
And when attaching to a function in a library, the probe definition would have the library's path rather than the executables, e.g. `uprobe:/lib64/libc.so:chmod`

danobi (approved):

danobi:
Can you also revert 0e71958ce4a38bd5aa5fb7d090da9a167547d797 and see if the appimage works again?

src/dwarf_parser.cpp
ttreyer: As a counter-example, a *library* could hide its implementation details with a pointer:
```
# file include/API/Foo.h
class private_Foo;

class Foo {
  private_Foo *impl;

public:
  void func1();
  void func2();
};
```

In that case, the *binary*'s Debug Info would contains info for `Foo`, but would not know anything about `private_Foo`. You'd need to load the *library*'s Debug Info in order to poke around `private_Foo`.
This is how `liblldb`'s API is implemented: `lldb::SBType` is just a wrapper around a shared pointer to `lldb_private::TypeImpl`.

ttreyer:
> Can you also revert [0e71958](https://github.com/bpftrace/bpftrace/commit/0e71958ce4a38bd5aa5fb7d090da9a167547d797) and see if the appimage works again?

Unfortunately, this change does not solve the appimage issue.

danobi:
Ah, ok. Unfortunate. But this seems like a good change anyways

ajor (approved):
Nice 👌

PR #3362, opened 2024-07-31
codegen: Remove unused MCJIT.h include

MaskRay:
The usage was removed by #2376.

MCJIT will be removed from LLVM:
https://discourse.llvm.org/t/rfc-removing-mcjit-and-runtimedyld/80464


danobi (approved):
thanks!

PR #3363, opened 2024-08-01
Add `-d` option in aot

ZhangYet:
Add the -d CLI option to AOT-compiled programs.

To resolve https://github.com/bpftrace/bpftrace/issues/3357

##### Checklist

- [*] Language changes are updated in `man/adoc/bpftrace.adoc`
- [*] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [*] The new behaviour is covered by tests


src/aot/aot_main.cpp
```diff
@@ -18,6 +18,7 @@ void usage(std::string_view filename)
   std::cerr << "USAGE: " << filename << " [options]" << std::endl;
   std::cerr << std::endl;
   std::cerr << "OPTIONS:" << std::endl;
+  std::cerr << "    -d STAGE       debug info for various stages of bpftrace execution" << std::endl;
```
danobi, line 21: Nit: probably better to place this down right below `-v` line. B/c debug is less likely to be used regularly

src/aot/aot_main.cpp
```diff
@@ -98,6 +99,16 @@ int main(int argc, char* argv[])
       case 'v':
         bt_verbose = true;
         break;
+      case 'd':
+        if (std::strcmp(optarg, "libbpf") == 0)
+          bt_debug.insert(DebugStage::Libbpf);
+        else if (std::strcmp(optarg, "verifier") == 0)
+          bt_debug.insert(DebugStage::Verifier);
+        else {
```
danobi, line 107: Perhaps also add support for "all" ?

danobi (changes requested):
thanks for the PR!

src/aot/aot_main.cpp
ZhangYet: Added.

src/aot/aot_main.cpp
ZhangYet: Done, thanks.

danobi (approved):
thanks!

PR #3365, opened 2024-08-01
codegen: Only initialize BPF target

danobi:
Previously, we were initializing all targets. We only codegen BPF code so we should be good to scope it down. In some basic tests, I don't see any performance wins. But nonetheless good to do this.

This also fixes the ubuntu distro build. Looks like there's some packaging issue where ubuntu configures with VE target but doesn't ship it in the shared library or something. This was the error:

```
/usr/bin/ld: ast/libast.a(codegen_llvm.cpp.o): undefined reference to symbol 'LLVMInitializeVETargetInfo@@LLVM_18.1'
```

##### Checklist


jordalgo (approved):
😮

PR #3367, opened 2024-08-02
Fix variable lexical scoping

jordalgo:
This implements a `scope` stack to prevent blocks
from accessing variables outside of their lexical scope.

Example, which is no longer valid:
```
BEGIN { if (0) { $var = 1; } print($var) }
```

We need to do this for:
- While
- For
- If (both the if and else blocks)
- Unroll
- Probe
- Subprog

Issue: https://github.com/bpftrace/bpftrace/issues/3017

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -941,13 +941,27 @@ fentry:tcp_connect {
 
 bpftrace knows two types of variables, 'scratch' and 'map'.
 
-'scratch' variables are kept on the BPF stack and only exists during the execution of the action block and cannot be accessed outside of the program.
-Scratch variable names always start with a `$`, e.g. `$myvar`.
+'scratch' variables are kept on the BPF stack and their names always start
+with a `$`, e.g. `$myvar`.
+'scratch' variables cannot be accessed outside of their lexical block e.g.
+```
+$a = 1;
+if ($a == 1) {
+  $b = "hello"
+  $a = 2;
+}
+// $b is not accessible here
+```
 
 'scratch' variables can also declared before or during initialization with `let` e.g.
 ```
-let $x;
-let $y = 11;
+let $a = 1;
+let $b;
+if ($a == 1) {
+  $b = "hello"
+  $a = 2;
+}
+// $b IS accessible here and would be an empty string if the condition wasn't true
 ```
 
 If no assignment is specified variables will initialize to 0.
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -318,7 +318,7 @@ AddrSpace SemanticAnalyser::find_addrspace(ProbeType pt)
 void SemanticAnalyser::visit(Builtin &builtin)
 {
   if (builtin.ident == "ctx") {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     ProbeType pt = probetype(probe->attach_points[0]->provider);
@@ -396,7 +396,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
                                                   "struct task_struct")),
                                  AddrSpace::kernel);
   } else if (builtin.ident == "retval") {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     ProbeType type = single_provider_type(probe);
@@ -434,7 +434,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
     // Case: @=comm and strncmp(@, "name")
     builtin.type.SetAS(AddrSpace::kernel);
   } else if (builtin.ident == "func") {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     for (auto *attach_point : probe->attach_points) {
@@ -463,7 +463,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
       }
     }
   } else if (builtin.is_argx()) {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     ProbeType pt = probetype(probe->attach_points[0]->provider);
@@ -490,7 +490,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
   } else if (!builtin.ident.compare(0, 4, "sarg") &&
              builtin.ident.size() == 5 && builtin.ident.at(4) >= '0' &&
              builtin.ident.at(4) <= '9') {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     ProbeType pt = probetype(probe->attach_points[0]->provider);
@@ -519,7 +519,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
     builtin.type = CreateUInt64();
     builtin.type.SetAS(addrspace);
   } else if (builtin.ident == "probe") {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     builtin.type = CreateProbe();
@@ -533,7 +533,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
     }
     builtin.type = CreateUInt32();
   } else if (builtin.ident == "args") {
-    auto probe = get_probe_from_scope(scope_, builtin.loc, builtin.ident);
+    auto probe = get_probe(builtin.loc, builtin.ident);
     if (probe == nullptr)
       return;
     for (auto *attach_point : probe->attach_points) {
@@ -645,7 +645,7 @@ void SemanticAnalyser::visit(Call &call)
     Visit(call.vargs[i]);
   }
 
-  if (auto probe = dynamic_cast<Probe *>(scope_)) {
+  if (auto probe = dynamic_cast<Probe *>(top_level_node_)) {
     for (auto *ap : probe->attach_points) {
       if (!check_available(call, *ap)) {
         LOG(ERROR, call.loc, err_) << call.func << " can not be used with \""
@@ -1074,7 +1074,7 @@ void SemanticAnalyser::visit(Call &call)
       }
     }
     call.type = CreateUInt64();
-    if (auto probe = dynamic_cast<Probe *>(scope_)) {
+    if (auto probe = dynamic_cast<Probe *>(top_level_node_)) {
       ProbeType pt = single_provider_type(probe);
       // In case of different attach_points, Set the addrspace to none.
       call.type.SetAS(find_addrspace(pt));
@@ -1089,7 +1089,7 @@ void SemanticAnalyser::visit(Call &call)
     call.type = CreateUInt64();
     call.type.SetAS(AddrSpace::kernel);
   } else if (call.func == "uaddr") {
-    auto probe = get_probe_from_scope(scope_, call.loc, call.func);
+    auto probe = get_probe(call.loc, call.func);
     if (probe == nullptr)
       return;
 
@@ -1337,7 +1337,7 @@ void SemanticAnalyser::visit(Call &call)
           << "signal only accepts string literals or integers";
     }
   } else if (call.func == "path") {
-    auto probe = get_probe_from_scope(scope_, call.loc, call.func);
+    auto probe = get_probe(call.loc, call.func);
     if (probe == nullptr)
       return;
 
@@ -1412,7 +1412,7 @@ void SemanticAnalyser::visit(Call &call)
     }
     call.type = CreateUInt64();
   } else if (call.func == "override") {
-    auto probe = get_probe_from_scope(scope_, call.loc, call.func);
+    auto probe = get_probe(call.loc, call.func);
     if (probe == nullptr)
       return;
 
@@ -1628,11 +1628,9 @@ void SemanticAnalyser::check_stack_call(Call &call, bool kernel)
   call.type = CreateStack(kernel, stack_type);
 }
 
-Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
-                                              const location &loc,
-                                              std::string name)
+Probe *SemanticAnalyser::get_probe(const location &loc, std::string name)
 {
-  auto probe = dynamic_cast<Probe *>(scope);
+  auto probe = dynamic_cast<Probe *>(top_level_node_);
   if (probe == nullptr) {
     // Attempting to use probe-specific feature in non-probe context
     if (name.empty()) {
@@ -1713,20 +1711,20 @@ void SemanticAnalyser::visit(Map &map)
 
 void SemanticAnalyser::visit(Variable &var)
 {
-  if (auto search_val = variables_[scope_].find(var.ident);
-      search_val != variables_[scope_].end()) {
-    if (search_val->second.was_assigned) {
-      var.type = search_val->second.type;
+  if (auto *found = find_variable(var.ident)) {
+    if (found->was_assigned) {
+      var.type = found->type;
     } else {
       LOG(ERROR, var.loc, err_)
           << "Variable used before it was assigned: " << var.ident;
       var.type = CreateNone();
     }
-  } else {
-    LOG(ERROR, var.loc, err_)
-        << "Undefined or undeclared variable: " << var.ident;
-    var.type = CreateNone();
+    return;
   }
+
+  LOG(ERROR, var.loc, err_)
+      << "Undefined or undeclared variable: " << var.ident;
+  var.type = CreateNone();
 }
 
 void SemanticAnalyser::visit(ArrayAccess &arr)
@@ -2184,18 +2182,18 @@ void SemanticAnalyser::visit(Ternary &ternary)
   }
 }
 
-void SemanticAnalyser::visit(If &if_block)
+void SemanticAnalyser::visit(If &if_node)
 {
-  Visit(if_block.cond);
+  Visit(if_node.cond);
 
   if (is_final_pass()) {
-    const Type &cond = if_block.cond->type.GetTy();
+    const Type &cond = if_node.cond->type.GetTy();
     if (cond != Type::integer)
-      LOG(ERROR, if_block.loc, err_) << "Invalid condition in if(): " << cond;
+      LOG(ERROR, if_node.loc, err_) << "Invalid condition in if(): " << cond;
   }
 
-  accept_statements(if_block.stmts);
-  accept_statements(if_block.else_stmts);
+  Visit(&if_node.if_block);
+  Visit(&if_node.else_block);
 }
 
 void SemanticAnalyser::visit(Unroll &unroll)
@@ -2216,7 +2214,7 @@ void SemanticAnalyser::visit(Unroll &unroll)
     LOG(ERROR, unroll.loc, err_) << "unroll minimum value is 1";
   }
 
-  accept_statements(unroll.stmts);
+  Visit(&unroll.block);
 }
 
 void SemanticAnalyser::visit(Jump &jump)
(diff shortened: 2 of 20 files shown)


ajor (commented):
#### AST
Can we do without having these Node types inherit from `Scope` and introducing multiple inheritance in the process?

- I'm thinking that it might make more sense for a loop to "have a" `Scope`, rather than for it to be an "is a" relationship. That'd remove the difference in implementation between `If` and the other node types
- Should this `Scope` node be called `Block` instead? Other things, e.g. a for-loop's header, can contribute to the scope but aren't part of this current `Scope` type.

#### Semantic Analyser
If `Scope` / `Block` is made into its own `Node` type, then pushing onto the `scope_stack_` member could be wrapped away in a `visit(Block&)`.

Instead of iterating over the entire scope stack each time we need to look up a variable, could we keep a cached set of valid variable names for the current scope? The cached set could be updated when entering/leaving a scope. It'd simplify looking up variables and I think also speed up semantic analysis.

#### Language Design
I'm not 100% against it, but it's the mixture of block-scoping and variables being hoisted out of if-statements that scares me. It's not clear in my mind yet what the rules are and I worry that it'll be confusing for users too. I probably need to go through some more examples.

jordalgo:
> I'm thinking that it might make more sense for a loop to "have a" Scope, rather than for it to be an "is a" relationship.

Yeah, I like this. Good call. Will update.

> Should this Scope node be called Block instead? 

I went back and forth on this and TBH I'm not sure what the correct answer is. Scope seems more explicit to me and self-explanatory, especially if we include it as a member of for, while, if, etc... I like saying "While statements have a Scope" rather than "While statements have a block".

> If Scope / Block is made into its own Node type, then pushing onto the scope_stack_ member could be wrapped away in a visit(Block&).

But what else would this node do? We still have to explicitly `pop` outside this node's `visit` right? 

> Instead of iterating over the entire scope stack each time we need to look up a variable, could we keep a cached set of valid variable names for the current scope? The cached set could be updated when entering/leaving a scope. It'd simplify looking up variables and I think also speed up semantic analysis.

Maybe we can amortize this in some way but yeah I agree with the sentiment.

BTW, I'll probably hold off on changes until we resolve https://github.com/bpftrace/bpftrace/issues/3017

ajor:
> > Should this Scope node be called Block instead?
> 
> I went back and forth on this and TBH I'm not sure what the correct answer is. Scope seems more explicit to me and self-explanatory, especially if we include it as a member of for, while, if, etc... I like saying "While statements have a Scope" rather than "While statements have a block".

I think `scope == block` for most block-types, but not for-loops or functions.
```
for ($kv : @map) { ... }
fn add($a: int64): int64 { ... }
```
`$kv` and `$a` are both in the scope of the above nodes, but not in their blocks.

> > If Scope / Block is made into its own Node type, then pushing onto the scope_stack_ member could be wrapped away in a visit(Block&).
> 
> But what else would this node do? We still have to explicitly `pop` outside this node's `visit` right?

I was imagining something like this:
```
visit(Scope &scope)
{
  scope_stack_.push_back(scope);
  for (Statement &stmt : scope.stmts) {
    accept(stmt);
  }
  scope_stack_.pop_back();
}
```

It'd also come of more use to centralise the logic for caching of valid variables in the current scope if we added that optimisation.

jordalgo:
```
visit(Scope &scope)
{
  scope_stack_.push_back(scope);
  for (Statement &stmt : scope.stmts) {
    accept(stmt);
  }
  scope_stack_.pop_back();
}
```
Looking again at the individual places where we push on to the scope_stack, there is logic specific to the block nodes that happens in between the push and pop e.g.
```
void SemanticAnalyser::visit(While &while_block)
{
  scope_stack_.push_back(&while_block.scope);

  if (is_final_pass() && !bpftrace_.feature_->has_loop()) {
    LOG(WARNING, while_block.loc, out_)
        << "Kernel does not support bounded loops. Depending"
           " on LLVMs loop unroll to generate loadable code.";
  }

  while_block.cond->accept(*this);

  loop_depth_++;
  accept_statements(while_block.scope.stmts);
  loop_depth_--;
  scope_stack_.pop_back();
}
```
So I'm not sure how we can easily make `Scope` a node without bloating it.

ajor:
That could just become this, no?
```
void SemanticAnalyser::visit(While &while_block)
{
  if (is_final_pass() && !bpftrace_.feature_->has_loop()) {
    LOG(WARNING, while_block.loc, out_)
        << "Kernel does not support bounded loops. Depending"
           " on LLVMs loop unroll to generate loadable code.";
  }

  while_block.cond->accept(*this);

  loop_depth_++;
  while_block.scope->accept(*this);
  loop_depth_--;
}
```
I don't think there's any need for the warning and loop condition to occur within the scope of the loop's body.

jordalgo:
> I don't think there's any need for the warning and loop condition to occur within the scope of the loop's body.

I think the condition *should* exist within the scope of the loop's body. Let's look at a `for` loop for example:
```
for ($kv : @map) {
  print($kv.0); // key
  print($kv.1); // value
}
```
`$kv` should not be accessible outside of the for loop. I believe this is true for other languages as well though C++ directly comes to mind e.g. these are both errors
```
std::vector<int> my_vec{1, 2};
for (auto i : my_vec) {
  std::cout << i;
}
std::cout << i; // undefined error
```
```
if (int i = some_condition()) {
  ...
}
std::cout << i; // undefined error
```

jordalgo:
> Instead of iterating over the entire scope stack each time we need to look up a variable, could we keep a cached set of valid variable names for the current scope? The cached set could be updated when entering/leaving a scope. It'd simplify looking up variables and I think also speed up semantic analysis.

@ajor I looked into this but I don't think there is any additional optimization here. At the time of scope iteration, we're only already iterating over eligible scopes and all the variables in those are valid. Also the scopes are all unique and are looking up maps of variables. Let me know if I'm missing something obvious.

ajor:
> I think the condition _should_ exist within the scope of the loop's body. Let's look at a `for` loop for example:

Agreed about for-loops, and that's why I think `Scope` is a misleading name for this node! The node doesn't include a for-loop's header, despite the header forming part of the lexical scope.

I still think it should be a proper AST `Node` type though. The way the AST is built is that everything is either a `Node` or a list of nodes and it would be odd to violate that for one particular type. If we're struggling to make it fit, then maybe we've got the wrong abstractions.

What about having a separate, non-AST data type to represent the abstract lexical scope. Then the AST node can just be a `Block`, which most of the time is directly used to build a `Scope,` but not always (e.g. in for-loops).

ajor:
> @ajor I looked into this but I don't think there is any additional optimization here. At the time of scope iteration, we're only already iterating over eligible scopes and all the variables in those are valid. Also the scopes are all unique and are looking up maps of variables. Let me know if I'm missing something obvious.

In this script:
```
$a = 1;
{
  $b = 1;
  {
    $c = 1;
    {
      $d = $1;
      print($d);
    }
  }
}
```

The scope stack will look something like this:
```
{$a}
{$b}
{$c}
{$d}
```

When we come to the `print($d)` statement, the current implementation will iterate over each level of the scope stack and perform a map-lookup in each one until it finds `$d`. That makes it an `O(s)` operation where `s` is the size of the scope stack.

What I'm suggesting is to instead build the scope stack like this, where all live variables are present in the top level of the stack:
```
{$a}
{$a,$b}
{$a,$b,$c}
{$a,$b,$c,$d}
```

Then the `print($d)` statement only ever needs to do one map-lookup into the top level of the scope stack - an `O(1)` operation.

jordalgo:
> When we come to the print($d) statement, the current implementation will iterate over each level of the scope stack and perform a map-lookup in each one until it finds $d. That makes it an O(s) operation where s is the size of the scope stack.

Right but I don't think we have any real life examples where the scope stack is ever bigger than 10 (and that's being generous).

> What I'm suggesting is to instead build the scope stack like this, where all live variables are present in the top level of the stack:

In this case we're paying for operations up front (that is the inserting, or removing, of references, or copies, of the variables into a new set/map when we create a new scope) when we're not sure we're going to need them.

jordalgo:
> Agreed about for-loops, and that's why I think Scope is a misleading name for this node! 

Still not sure if I understand why the name is misleading for this example but I understand that if you want to see this in the AST then it probably reads better as Block.

> What about having a separate, non-AST data type to represent the abstract lexical scope. Then the AST node can just be a Block, which most of the time is directly used to build a Scope, but not always (e.g. in for-loops).

I'm not sure if I understand what you mean here. Do you mind adding an example? I think I'm having trouble here because I don't see a easy way to generalize the logic that exists between pushing a scope to the stack and popping it off. Or do you mean, we can have a `BlockStart` and `BlockEnd` node that the parent Node can `visit` at different logic points that controls the scope?

jordalgo:
@ajor and I cleared some things up offline. Going to put up a few minor changes and then I think this should be good to go.

jordalgo:
Please vote here if you want this diff to move forward. *Note*: we will first tally the votes from only maintainers and very active contributors but also take into account votes from the larger community to make this decision.

- **Thumbs up** for this PR (or some version of it): **block/lexical scoping** 👍 
- **Thumbs down** for keeping things as they are: **Probe scoping** 👎 
- Any other random emoji if you don't feel strongly.

tools/tcpdrop.bt
```diff
@@ -80,7 +73,17 @@ tracepoint:skb:kfree_skb
...
+        
+    if ($inet_family == AF_INET) {
+      $daddr = ntop($sk->__sk_common.skc_daddr);
+      $saddr = ntop($sk->__sk_common.skc_rcv_saddr);
+      printf("%39s:%-6d %39s:%-6d %-10s\n", $saddr, $lport, $daddr, $dport, $statestr);
+    } else {
+      $daddr = ntop($sk->__sk_common.skc_v6_daddr.in6_u.u6_addr8);
+      $saddr = ntop($sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr8);
+      printf("%39s:%-6d %39s:%-6d %-10s\n", $saddr, $lport, $daddr, $dport, $statestr);
+    }
```
viktormalik, line 85: While I'm a fan of block scoping in general, the fact that `$daddr` is not accessible after the `if`/`else` block is quite a big backwards compatibility breakage. I can imagine a lot of scripts out there which rely on this functionality.

I may have missed something in the discussions but do we plan to introduce variable declarations so that the only update people need to do is add an explicit declaration before the condition? Or do we even consider declaration hoisting?

fbs: This was discussed on IRC a bit but I dont think anything came out of it.

I'm not a fan of hoisting as it basically undoes the scoping. You're lifting the var to the outer scope, might as well stick with probe scope. Nested if statements also become confusing

If we do block scoping we also need a way to do upfront declaration. Either the ugly way :

```
$daddr = ntop(dummy)
if ...
  $daddr = ntop(real)
 ```
 
 or by adding the option to define upfront.
 
 This probably shouldn't land until a decision has been made here so good to bring it up

Comment (jordalgo) [reply tools/tcpdrop.bt:85]: I think the general consensus is that we want to introduce variable declarations, which seem to coincide with both the [user-defined functions PR](https://github.com/bpftrace/bpftrace/pull/3068) and this [map declarations RFC](https://github.com/bpftrace/bpftrace/issues/2954). If we land this PR without waiting for variable declarations then to @fbs 's point, users have to use the ugly way in that they need to initialize with a junk or 0 value. 

IMHO, variable declarations don't seem very controversial and are probably not that hard to implement, so I'm fine doing that work either stacked on top of this PR or to follow closely after before we release another bpftrace version with this in it.

Comment (viktormalik) [reply tools/tcpdrop.bt:85]: So, to make things clear, with variable declarations but no hoisting (which is roughly the plan), the above will become
```
let $daddr, $saddr: inet_t;
if ($inet_family == AF_INET) {
  $daddr = ntop($sk->__sk_common.skc_daddr);
  $saddr = ntop($sk->__sk_common.skc_rcv_saddr);
} else {
  $daddr = ntop($sk->__sk_common.skc_v6_daddr.in6_u.u6_addr8);
  $saddr = ntop($sk->__sk_common.skc_v6_rcv_saddr.in6_u.u6_addr8);
}
printf("%39s:%-6d %39s:%-6d %-10s\n", $saddr, $lport, $daddr, $dport, $statestr);
```
and semantic analyser will check that the value is defined in all branches, otherwise print a warning.

From users perspective, they will just need to add explicit declarations to their scripts. It's still a backwards compat breakage but certainly something we can live with.

Comment (jordalgo) [reply tools/tcpdrop.bt:85]: @viktormalik That's pretty much how I imagine it - maybe without the `let` keyword 😉

Comment (jordalgo) [reply tools/tcpdrop.bt:85]: @fbs @viktormalik I've got a first draft of variable declarations (still needs more tests and a fix that hasn't been merged yet) just to give you an idea of what it looks like. It's pretty simple, which is good:
https://github.com/jordalgo/bpftrace/commit/2bcb96a35287e39fb20e12cfb121a1e334474aab

Comment (viktormalik) [reply tools/tcpdrop.bt:85]: Looks pretty straightforward, nice!

Comment (fbs) [reply tools/tcpdrop.bt:85]: IIRC some other MR/proposal added a let syntax. Should we centralize that work in an rfc like issue and block this MR on that?

Comment (jordalgo) [reply tools/tcpdrop.bt:85]: @fbs Are you talking about the map declarations proposal? https://github.com/bpftrace/bpftrace/issues/2954 (cc @JakeHillion)

File: src/ast/passes/semantic_analyser.cpp [hunk 2]
```diff
@@ -2969,45 +2976,53 @@ void SemanticAnalyser::visit(VarDeclStatement &decl)
...
 
-  // This should be the first time we're seeing this variable
-  if (auto decl_search = variable_decls_[scope_].find(var_ident);
-      decl_search != variable_decls_[scope_].end()) {
-    if (decl_search->second != decl.loc) {
-      // Note this will change if/when proper block/lexical scoping is added
-      LOG(ERROR, decl.loc, err_)
-          << "Variable " << var_ident << " was already declared.";
-      LOG(ERROR, decl_search->second, err_) << "Initial declaration";
+  if (is_first_pass()) {
```
viktormalik, line 2979: How come this condition was not necessary before?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2969,45 +2976,53 @@ void SemanticAnalyser::visit(VarDeclStatement &decl)
...
-      LOG(ERROR, decl.loc, err_)
-          << "Variable declarations need to occur before variable usage or "
-             "assignment. Variable: "
-          << var_ident;
-    } else if (is_final_pass()) {
-      // Update the declaration type if it was either not set e.g. `let $a;`
-      // or the type is ambiguous or resizable e.g. `let $a: string;`
-      decl.var->type = search->second.type;
-    }
+  if (is_first_pass() || is_final_pass()) {
```
viktormalik, line 2994: Why do this specifically on the first and the last pass? Can we add a comment to explain?

src/ast/passes/semantic_analyser.h
```diff
@@ -159,7 +159,8 @@ class SemanticAnalyser : public Visitor {
   };
   void accept_statements(StatementList &stmts);
 
-  Scope *scope_;
+  std::vector<Node *> scope_stack_;
```
viktormalik, line 162: Representing the scope stack with `std::vector` means that we're always iterating it in the "opposite" order, i.e. the outermost scope first. It shouldn't matter now as we don't support variable shadowing but if we ever decide to do so, this may become a problem so I'd at least explicitly state the fact here.

tests/semantic_analyser.cpp
```diff
@@ -4304,6 +4306,86 @@ stdin:1:34-47: ERROR: Type mismatch for $a: trying to assign value of type 'stru
...
+                         ~~~~~~
+)");
+}
+
+TEST(semantic_analyser, block_scoping)
+{
+  // if/else
+  test("BEGIN { $a = 1; if (1) { $b = 2; print(($a, $b)); } }");
+  test("BEGIN { $a = 1; if (1) { print(($a)); $b = 2; if (1) { print(($a, "
+       "$b)); } else { print(($a, $b)); } } }");
```
viktormalik, line 4322: I know it's just a test but could we have a nicer indentation of programs which span multiple lines? Same for the other test cases in this block.

viktormalik (commented):
I like how straightforward this change mostly is. Logic looks good to me, I just have a few code readability remarks.

man/adoc/bpftrace.adoc
```diff
@@ -938,13 +938,26 @@ kfunc:tcp_connect {
 
 bpftrace knows two types of variables, 'scratch' and 'map'.
 
-'scratch' variables are kept on the BPF stack and only exists during the execution of the action block and cannot be accessed outside of the program.
-Scratch variable names always start with a `$`, e.g. `$myvar`.
+Scratch variable names always start with a `$`, e.g. `$myvar`.
```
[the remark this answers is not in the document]
jordalgo: Cause they won't always be -> all of @ryantimwilson's work to sometimes put things in a global scratch buffer. It seems like an implementation detail that users don't need to know about.

src/ast/passes/semantic_analyser.h
jordalgo: Good call.

man/adoc/bpftrace.adoc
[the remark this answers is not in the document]
viktormalik: Yeah, but I thought that at the moment, scratch variables are always stored on stack. My point is that I wouldn't change this in the docs before the actual implementation changes.



> It seems like an implementation detail that users don't need to know about.

I don't think it's an implementation detail. It inherently limits the overall size of scratch variables which is something that users can run into.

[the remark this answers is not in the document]
jordalgo: kk, I'll change it back for now.

src/ast/passes/semantic_analyser.cpp
jordalgo: Since variables are no longer probe scoped. This is now valid `BEGIN { if (1) { let $x; } else { let $x; } let $x; }` - See how we define the last $x in the outer scope at the end. Now on subsequent passes, it's viewed as variable shadowing in the if/else blocks.

viktormalik: Right, makes sense. A comment would be helpful, though.

src/ast/passes/semantic_analyser.cpp
jordalgo: The only error checks below are for last or first pass so no need to iterate over the scope stack in other passes. I'm adding a comment as to why we check for `first_pass` - it's similar to above. For cases like this: `BEGIN { if (1) { let $x; } $x = 2; }`

jordalgo:
@viktormalik I think I addressed all your comments. 🤞

man/adoc/bpftrace.adoc
```diff
@@ -941,13 +941,27 @@ fentry:tcp_connect {
...
 ```
-let $x;
-let $y = 11;
+let $a = 1;
+let $b;
+if ($a == 1) {
+  $b = "hello"
+  $a = 2;
+}
+// $b IS accessible here
```
viktormalik, line 964: Should we say that it will be 0 if the condition was not true?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2389,27 +2387,17 @@ void SemanticAnalyser::visit(For &f)
   // and declared before the loop. These will be passed into the loop callback
   // function as the context parameter.
   CollectNodes<Variable> vars_referenced;
-  std::unordered_set<std::string> var_set;
+  std::unordered_set<std::string> found_vars;
   for (auto *stmt : f.stmts) {
-    const auto &live_vars = variables_[scope_];
-    vars_referenced.run(*stmt, [&live_vars, &var_set](const auto &var) {
-      if (live_vars.find(var.ident) == live_vars.end())
+    vars_referenced.run(*stmt, [this, &f, &found_vars](const auto &var) {
```
viktormalik, line 2392: Do we still need to pass `&f`?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2969,45 +2959,62 @@ void SemanticAnalyser::visit(VarDeclStatement &decl)
...
-      // Update the declaration type if it was either not set e.g. `let $a;`
-      // or the type is ambiguous or resizable e.g. `let $a: string;`
-      decl.var->type = search->second.type;
-    }
+  if (is_first_pass() || is_final_pass()) {
+    if (auto *scope = find_variable_scope(var_ident)) {
+      auto &foundVar = variables_[scope][var_ident];
+      // Checking the first pass only for cases like this:
+      // `BEGIN { if (1) { let $x; } $x = 2; }`
+      // Again, this is legal and their is no ambiguity but `$x = 2` gets
```
viktormalik, line 2988: typo?
```suggestion
      // Again, this is legal and there is no ambiguity but `$x = 2` gets
```

viktormalik (commented):
> @viktormalik I think I addressed all your comments. 🤞

You did, I just found 3 very small things :-)

jordalgo:
@viktormalik with the sharp eyes 👀

jordalgo:
Fixed 👍

viktormalik (approved):
Nice!

jordalgo:
@fbs Any objections to merging?

PR #3370, opened 2024-08-02
codegen: NFC: Rename onStack() and shouldBeOnStackAlready()

danobi:
The two functions are misnamed. The intent is to check for types that are directly accessible from the program. Meaning no probe reads. Now that we are making heavier use of perpcu scratch maps, we should rename the helper so the code is more clear.

##### Checklist

### DIFF
--- a/src/ast/codegen_helper.h
+++ b/src/ast/codegen_helper.h
@@ -10,7 +10,12 @@ inline bool needMemcpy(const SizedType &stype)
   return stype.IsAggregate() || stype.IsTimestampTy() || stype.IsCgroupPathTy();
 }
 
-inline bool shouldBeOnStackAlready(const SizedType &type)
+// BPF memory is memory that the program can access with a regular
+// dereference. This could mean the value is on the stack, a map, or
+// maybe something else (like BPF arenas) in the future.
+//
+// This means that a bpf_probe_read_*() is _NOT_ required.
+inline bool shouldBeInBpfMemoryAlready(const SizedType &type)
 {
   return type.IsStringTy() || type.IsBufferTy() || type.IsInetTy() ||
          type.IsUsymTy() || type.IsKstackTy() || type.IsUstackTy() ||
@@ -18,14 +23,14 @@ inline bool shouldBeOnStackAlready(const SizedType &type)
          type.IsCgroupPathTy();
 }
 
-inline bool onStack(const SizedType &type)
+inline bool inBpfMemory(const SizedType &type)
 {
-  return type.is_internal || shouldBeOnStackAlready(type);
+  return type.is_internal || shouldBeInBpfMemoryAlready(type);
 }
 
 inline AddrSpace find_addrspace_stack(const SizedType &ty)
 {
-  return (shouldBeOnStackAlready(ty)) ? AddrSpace::kernel : ty.GetAS();
+  return (shouldBeInBpfMemoryAlready(ty)) ? AddrSpace::kernel : ty.GetAS();
 }
 
 } // namespace ast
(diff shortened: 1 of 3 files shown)


src/ast/codegen_helper.h
```diff
@@ -10,22 +10,27 @@ inline bool needMemcpy(const SizedType &stype)
   return stype.IsAggregate() || stype.IsTimestampTy() || stype.IsCgroupPathTy();
 }
 
-inline bool shouldBeOnStackAlready(const SizedType &type)
+// BPF memory is memory that the program can access with a regular
+// dereference. This could mean the value is on the stack, a map, or
+// maybe something else (like BPF arenas) in the future.
+//
+// The necessarily means that a bpf_probe_read_*() is _NOT_ required.
```
jordalgo, line 17:
```suggestion
// This means that a bpf_probe_read_*() is _NOT_ required.
```

src/ast/codegen_helper.h
```diff
@@ -10,22 +10,27 @@ inline bool needMemcpy(const SizedType &stype)
...
          type.IsUsymTy() || type.IsKstackTy() || type.IsUstackTy() ||
          type.IsTupleTy() || type.IsTimestampTy() || type.IsMacAddressTy() ||
          type.IsCgroupPathTy();
 }
 
-inline bool onStack(const SizedType &type)
+inline bool inBpfMemory(const SizedType &type)
 {
-  return type.is_internal || shouldBeOnStackAlready(type);
+  return type.is_internal || shouldBeInBpfMemoryAlready(type);
```
jordalgo, line 28: The next thing to clear up is this 'is_internal' flag. I don't understand why it only applies to `Type::inet` and `Type::mac_address`. But not for this PR.

danobi: Yeah, there's probably some stuff we should clean up around this. It was probably written around the time when we had a pretty sharp transition in people working on this project, and the new guys (me and Viktor) were just figuring stuff out as we went

src/ast/codegen_helper.h
```diff
@@ -10,22 +10,27 @@ inline bool needMemcpy(const SizedType &stype)
   return stype.IsAggregate() || stype.IsTimestampTy() || stype.IsCgroupPathTy();
 }
 
-inline bool shouldBeOnStackAlready(const SizedType &type)
+// BPF memory is memory that the program can access with a regular
+// dereference. This could mean the value is on the stack, a map, or
+// maybe something else (like BPF arenas) in the future.
+//
+// The means that a bpf_probe_read_*() is _NOT_ required.
```
jordalgo, line 17: Sorry, one more minor typo

```suggestion
// This means that a bpf_probe_read_*() is _NOT_ required.
```

danobi: Ha, working too fast...

jordalgo (approved):

src/ast/codegen_helper.h
ajor: Definitely some messiness with this.

`is_internal` is only hardcoded to true for `inet` and `mac_address` (I don't know why they don't just use `shouldBeInBpfMemoryAlready()`), but it's also set dynamically for other types, e.g. structs when assigned to a map: https://github.com/bpftrace/bpftrace/blob/898ae5e10a3e5f5c14290623b21507da776348cc/src/ast/passes/semantic_analyser.cpp#L2532

jordalgo: I'll try to make some time to clean it up.

ajor: Maybe we could move across to just using the address space markers? `AddrSpace::bpf` means `is_internal` / `inBpfMemory()` and `AddrSpace::kernel` or `AddrSpace::user` otherwise

danobi: That makes sense to me. And it probably opens up some other interesting optimizations in the future. I bet we're emitting bpf_probe_read_kernel()s where they aren't strictly necessary at the moment.

PR #3371, opened 2024-08-02
Increase default string size to 1024

danobi:
This is the final PR for big-X support, where X is strings, buffers, etc.

In this PR we change the default size of these types to 1024. Hopefully
that is enough to make truncation happen very rarely in practice. Even
if it does occur, the user has the option to change BPFTRACE_MAX_STRLEN
to whatever they want now.

The biggest potential regression is that these types will be difficult to
store to scratch variables, as scratch variables go onto the stack and
we only have 512B of stack to work with. However, the user has 3 fairly
easy workarounds:

1. Assign it to a map instead
2. Save the str() call for the print()/printf() argument
3. Clamp the string size

Given that big strings have been asked about for years, it's probably
worth the small amount of pain it causes to make big strings work
out of the box for everyone else.

I've also improved semantic analyser diagnostics to nudge users
to those 3 remediation if they need it.

##### Checklist


src/ast/passes/semantic_analyser.cpp
```diff
@@ -2677,6 +2678,18 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
     if (ty == Type::none)
       LOG(ERROR, assignment.expr->loc, err_)
           << "Invalid expression for assignment: " << ty;
+
+    // Scratch variables are on stack. The BPF stack is only 512 bytes at time
```
jordalgo, line 2682: nit
```suggestion
    // Scratch variables are on the BPF stack, which is only 512 bytes at time
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2677,6 +2678,18 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
...
+    // of writing, so large values do not fit on there. 200 is a ballpark of
+    // what probably won't work.
+    auto expr_size = assignTy.GetSize();
+    if (expr_size > 200) {
+      LOG(ERROR, assignment.loc, err_)
+          << "Value is too big "
+          << "(" << expr_size << " bytes) for the stack. "
+          << "Try reducing its size, storing it in a map, or creating it in "
```
jordalgo, line 2690: Maybe a small example for the the last option wouldn't be bad.

jordalgo: Actually, an example for the size reduction would be good to.

danobi: Updated. I originally didn't wanna get too specific as multiple types can trigger this (tuples included). But probably ok to optimize for common case strings. It's hopefully obvious enough

jordalgo: I like it! I always find examples super helpful.

jordalgo:
Ah I forgot. The adoc looks like it needs updating - it has the old MAX_STRLEN value and the few mentions of that config probably could use some additional context now.

danobi:
> Ah I forgot. The adoc looks like it needs updating - it has the old MAX_STRLEN value and the few mentions of that config probably could use some additional context now.

Good call. I did a grep for the env var but forgot to check lower case config name

man/adoc/bpftrace.adoc
```diff
@@ -3443,15 +3443,12 @@ system.
...
 
-Default: 64
+Default: 1024
 
-Number of bytes allocated on the BPF stack for the string returned by str().
-Make this larger if you wish to read bigger strings with str().
-Beware that the BPF stack is small (512 bytes), and that you pay the toll again inside printf() (whilst
-it composes a perf event output buffer).
-So in practice you can only grow this to about 200 bytes.
+The maximum length (in bytes) for strings created by `str()`.
```
jordalgo, line 3448: nit
```suggestion
The maximum length (in bytes) for strings created by `str()` and `buf()`.
```

danobi: and path() too

man/adoc/bpftrace.adoc
```diff
@@ -3443,15 +3443,12 @@ system.
...
 
-Number of bytes allocated on the BPF stack for the string returned by str().
-Make this larger if you wish to read bigger strings with str().
-Beware that the BPF stack is small (512 bytes), and that you pay the toll again inside printf() (whilst
-it composes a perf event output buffer).
-So in practice you can only grow this to about 200 bytes.
+The maximum length (in bytes) for values created by `str()`, `buf()` and `path()`.
 
-Support for even larger strings is being discussed: https://github.com/bpftrace/bpftrace/issues/305.
+This limit is necessary in order to size buffers data must be staged in before they can be interacted with.
```
jordalgo, line 3450: Sorry, now you got me in editing mode. I won't block but how about
```suggestion
This limit is necessary because BPF requires the size of all program strings to be declared up front. This is the size for all string variables in bpftrace unless specified at the call site.
```

jordalgo (approved):

man/adoc/bpftrace.adoc
danobi: Edited a little to account for the other types, but overall taken

PR #3375, opened 2024-08-03
Use second param for `str` to set length

jordalgo:
If this value is a literal then we can set the string length up front and avoid the explosion of branches created by codegen in the `strcontains` function.

We want to change BPFTRACE_MAX_STRLEN up from 64 to 1024.
In an effort to break as few existing scripts as possible,
we need to clamp string sizes when possible (namely when a
literal size param is provided) so that if the user tries
to assign a string to a scratch variable, we use as little
stack space as we can get away with.

Without this, then all scratch strings would take 1024
bytes on stack which LLVM or the verifier would reject
out of hand.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -685,7 +685,33 @@ void SemanticAnalyser::visit(Call &call)
             << call.func << "() expects an integer or a pointer type as first "
             << "argument (" << t << " provided)";
       }
-      call.type = CreateString(bpftrace_.config_.get(ConfigKeyInt::max_strlen));
+
+      auto strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
+
+      if (call.vargs.size() == 2 && check_arg(call, Type::integer, 1, false)) {
+        auto &size_arg = *call.vargs.at(1);
+        if (size_arg.is_literal) {
+          auto &integer = static_cast<Integer &>(size_arg);
+          long value = integer.n;
+          if (value < 0) {
+            if (is_final_pass())
+              LOG(ERROR, call.loc, err_)
+                  << call.func << "cannot use negative length (" << value
+                  << ")";
+          } else if (value > static_cast<int64_t>(strlen)) {
+            if (is_final_pass())
+              LOG(WARNING, call.loc, out_)
+                  << "length param (" << value
+                  << ") is too long and will be shortened to " << strlen
+                  << " bytes (see BPFTRACE_MAX_STRLEN)";
+          } else {
+            strlen = value;
+          }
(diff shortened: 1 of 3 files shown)


jordalgo:
Follow up from discussion here: https://github.com/bpftrace/bpftrace/discussions/3373

danobi:
This approach makes sense to me. Better than what I came up with in #3371. I'm trying it out there now

src/ast/passes/semantic_analyser.cpp
```diff
@@ -685,7 +685,27 @@ void SemanticAnalyser::visit(Call &call)
...
+        if (size_arg.is_literal) {
+          auto &integer = static_cast<Integer &>(size_arg);
+          long value = integer.n;
+          if (value < 0) {
+            if (is_final_pass())
+              LOG(ERROR, call.loc, err_)
+                  << call.func << "cannot use negative length (" << value
+                  << ")";
+          } else {
+            strlen = value;
```
danobi, line 702: Do we want the value to include NUL terminator or exclude?

Eg, if `str($1, 4)` is called, should we size a 4 byte buffer or 5?

danobi: Should also probably check if `value` is bigger than max_strlen

danobi:
Yep, this approach did the trick. Let's get this in before my PRs. Also perhaps add something about the big string use case to the commit msg here:

```
We want to change BPFTRACE_MAX_STRLEN up from 64 to 1024. In an effort
to break as few existing scripts as possible, we need to clamp string
sizes when possible (namely when a literal size param is provided) so
that if the user tries to assign a string to a scratch variable, we use
as little stack space as we can get away with.

Without this, then all scratch strings would take 1024 bytes on stack
which LLVM or the verifier would reject out of hand.
```

src/ast/passes/semantic_analyser.cpp
jordalgo: > Do we want the value to include NUL terminator or exclude?

Doesn't this happen automatically in codegen: https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/codegen_llvm.cpp#L782

jordalgo:
> A semantic analyser test would be nice, showing the correct length is set in the AST if you've got the time.

No problem. Added.

src/ast/passes/semantic_analyser.cpp
danobi: Yes, kinda. The generated conditional is mostly to handle dynamic len args. But if user provides a literal one I think it'd be better to error out. Principle of least surprise IMO

danobi (approved):

PR #3377, opened 2024-08-04
Add root check to AOT binaries

jordalgo:
Currently running the binary without root will result in a verifier error, which is not a good user experience.

##### Checklist

### DIFF
--- a/src/aot/aot_main.cpp
+++ b/src/aot/aot_main.cpp
@@ -123,6 +123,8 @@ int main(int argc, char* argv[])
     return 1;
   }
 
+  check_is_root();
+
   libbpf_set_print(libbpf_print);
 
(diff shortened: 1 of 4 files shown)


jordalgo:
> Utils.cpp is becoming a dumping ground for all kinds of different things and is a bit of a mess

I think that's what a file named is "utils" is for but NBD, I moved it to `run_bpftrace.cpp`

ajor:
I think a file called "utils.cpp" existing is a code smell by itself 😛

ajor (approved):

danobi (approved):
not a huge fan of `exit()` calls in non-main function but I don't think it matters practically

jordalgo:
Merging before this devolves into a conversation about code organization.

PR #3385, opened 2024-08-08
AST: Replace non-optional pointers with reference types

ajor:
Introduce a new class `ref` which holds references / non-nullable pointers.

-----

This'll cause **a lot** of conflicts so I'm happy to rebase other PRs on top of these changes to ease the process for others.

Instructions for rebasing:
1. Rebase, resolving conflicts by taking the changes from other PRs and discarding the stuff from this/master: `git rebase -i upstream/master -Xtheirs` (yes, `theirs`, not `ours`, even though it sounds wrong)
2. Try to compile: `ninja` / `make`
3. Fix build errors by removing `*`, replacing `->` with `.`, etc.

### DIFF
--- a/src/ast/passes/printer.cpp
+++ b/src/ast/passes/printer.cpp
@@ -11,10 +11,10 @@
 namespace bpftrace {
 namespace ast {
 
-void Printer::print(Node *root)
+void Printer::print(Node &root)
 {
   ++depth_;
-  Visit(*root);
+  Visit(root);
   --depth_;
 }
 
@@ -110,8 +110,8 @@ void Printer::visit(Call &call)
   out_ << indent << "call: " << call.func << type(call.type) << std::endl;
 
   ++depth_;
-  for (Expression *expr : call.vargs) {
-    expr->accept(*this);
+  for (Expression &expr : call.vargs) {
+    expr.accept(*this);
   }
   --depth_;
 }
@@ -152,8 +152,8 @@ void Printer::visit(Map &map)
   out_ << indent << "map: " << map.ident << type(map.type) << std::endl;
 
   ++depth_;
-  for (Expression *expr : map.vargs) {
-    expr->accept(*this);
+  for (Expression &expr : map.vargs) {
+    expr.accept(*this);
   }
   --depth_;
 }
@@ -170,8 +170,8 @@ void Printer::visit(Binop &binop)
   out_ << indent << opstr(binop) << type(binop.type) << std::endl;
 
   ++depth_;
-  binop.left->accept(*this);
-  binop.right->accept(*this);
+  binop.left().accept(*this);
+  binop.right().accept(*this);
   --depth_;
 }
 
@@ -180,14 +180,14 @@ void Printer::visit(Unop &unop)
   if (unop.is_post_op) {
     std::string indent(depth_ + 1, ' ');
 
-    unop.expr->accept(*this);
+    unop.expr().accept(*this);
     out_ << indent << opstr(unop) << std::endl;
   } else {
     std::string indent(depth_, ' ');
     out_ << indent << opstr(unop) << std::endl;
 
     ++depth_;
-    unop.expr->accept(*this);
+    unop.expr().accept(*this);
     --depth_;
   }
 }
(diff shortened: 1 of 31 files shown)


ajor:
The clang-format failure is annoying - my local clang-format puts it in the way I've got it but the CI wants it laid out differently.

danobi:
Maybe you have a different clang-format version installed. You could try using the CI version: v

src/ast/passes/printer.cpp
```diff
@@ -321,12 +321,12 @@ void Printer::visit(Unroll &unroll)
   out_ << indent << "unroll" << std::endl;
 
   ++depth_;
-  unroll.expr->accept(*this);
+  unroll.expr().accept(*this);
```
danobi, line 324: the reason we need `operator()` here and not for `unroll.stmts` is the binding in the latter implicitly converts the `ref`?

danobi:
IMO the use of `operator()` makes everything look too much like a function call. Which I suppose on a really technical level is true. But kinda misleading to read still. WDYT about this?

```diff
diff --git a/src/ast/passes/printer.cpp b/src/ast/passes/printer.cpp
index 05eb023c..8c9e1a62 100644
--- a/src/ast/passes/printer.cpp
+++ b/src/ast/passes/printer.cpp
@@ -321,7 +321,7 @@ void Printer::visit(Unroll &unroll)
   out_ << indent << "unroll" << std::endl;

   ++depth_;
-  unroll.expr().accept(*this);
+  unroll.expr->accept(*this);
   out_ << indent << " block" << std::endl;

   ++depth_;
diff --git a/src/container/ref.h b/src/container/ref.h
index 2eb2bfc4..94a5b744 100644
--- a/src/container/ref.h
+++ b/src/container/ref.h
@@ -17,6 +17,11 @@ public:
   {
   }

+  constexpr T *operator->() const
+  {
+    return &this->get();
+  }
+
   constexpr T &operator()() const
   {
     return this->get();
```

IIUC it's still completely pointer-free despite appearances. It's using the smart pointer interface, so the compiler is turning something like `x->member` into `x.operator->()->member`.

src/ast/passes/printer.cpp
ajor: yeah, exactly

ajor:
> IMO the use of `operator()` makes everything look too much like a function call. Which I suppose on a really technical level is true. But kinda misleading to read still. WDYT about this?

I deliberately wanted to avoid pointer-like syntax, because it gives the impression that there's the possibility of a null value. I wanted it to be clear which fields must be checked for null before use (e.g. `probe.pred`) and which are guaranteed to hold a value.

It's not quite accurate, but I was thinking that `operator()` could make it look as if the AST nodes provided getters instead of direct access to their members. This would have the same-ish interface:
```
class Unop {
public:
  Expression &expr() { return expr_; }
private:
  ...
};
```

`std::reference_wrapper`'s `get()` is another option, but it's so verbose :( 
I would definitely have preferred to make it completely transparent if overloading operator `.` was a thing.

jordalgo:
> std::reference_wrapper's get() is another option, but it's so verbose :(

TBH, I think I prefer the verbosity and the familiar `get` syntax over the API of the container ref. Just less to keep in my poor little brain when I'm reading.

ajor:
I suppose I could refactor it into this form which would get rid of the need for the custom operators:
```
class Unop {
public:
  Expression &expr() { return expr_; }
private:
  std::reference_wrapper<Expression> expr_;
};
```

Loops over `vector<reference_wrapper<Expression>>` can still be written to avoid the `.get()`:
```
for (Expression &expr : args) { ... }

Then we'd only be left

jordalgo:
> So I'm fine with going for this approach instead of having a custom operator if you think it's clearer.

I prefer that approach but if other folks feel strongly about keeping the custom operator I won't die on this hill 😄

danobi:
> > IMO the use of `operator()` makes everything look too much like a function call. Which I suppose on a really technical level is true. But kinda misleading to read still. WDYT about this?
> 
> I deliberately wanted to avoid pointer-like syntax, because it gives the impression that there's the possibility of a null value. I wanted it to be clear which fields must be checked for null before use (e.g. `probe.pred`) and which are guaranteed to hold a value.

Good point.

danobi:
Closing due to inactivity.

PR #3389, opened 2024-08-09
clang-tidy: Make codebase bugprone-unchecked-optional-access clean

danobi:
This lint checks that optionals are checked before access. Trying to access a empty optional is an abort which is bad.

Unfortunately clang-tidy doesn't seem to know that gtest ASSERT_*() terminates the test case so I've had to annotate a few locations.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -821,8 +821,9 @@ void CodegenLLVM::visit(Call &call)
       length = b_.CreateSelect(
           cmp, proposed_length, max_length, "length.select");
 
-      if (arg.is_literal)
-        fixed_buffer_length = *bpftrace_.get_int_literal(&arg);
+      auto literal_length = bpftrace_.get_int_literal(&arg);
+      if (literal_length)
+        fixed_buffer_length = *literal_length;
     } else {
       auto &arg = *call.vargs.at(0);
       fixed_buffer_length = arg.type.GetNumElements() *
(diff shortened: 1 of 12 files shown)


danobi:
Not sure if this one is too much noise to enable. It did find 2 legit issues. The rest was false positive.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1307,8 +1308,11 @@ void CodegenLLVM::visit(Call &call)
   } else if (call.func == "strerror") {
     auto scoped_del = accept(call.vargs.front());
   } else if (call.func == "strncmp") {
+    // Checked in semantic analysis
```
jordalgo, line 1311: This isn't on the hot path, I almost feel like the extra optional check is better than this comment. Also it protects if, for some reason, the semantic analysis check gets removed.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1417,8 +1421,11 @@ void CodegenLLVM::visit(Call &call)
     expr_ = ret;
   } else if (call.func == "nsecs") {
     if (call.type.ts_mode == TimestampMode::sw_tai) {
+      // Checked in semantic analyser
```
jordalgo, line 1424: Same as above

jordalgo:
I think it's good to enable this 👍

src/ast/passes/codegen_llvm.cpp
danobi: It's not done this way for performance. The reason is b/c codegen is not responsible for error detection - that occurs in semantic analysis.

jordalgo: So why not something like:
```
auto int_literal = bpftrace_.get_int_literal(call.vargs.at(2));
if (!int_literal) {
	LOG(BUG) << "Int literal should have been checked in semantic_analyser";
}
```

danobi: Ah yeah that's better

jordalgo (approved):

PR #3390, opened 2024-08-11
RFC: Convert associative array map keys to tuples

jordalgo:
This is for a few reasons:
- it aligns with the type that is created when looping over a map that has an associative array key
- it opens the door to fixing/changing the API for `delete` to `delete(map_ref, key)` as we currently don't have support for associative array literals
- it opens the door for a `has_key` map function that doesn't mimic the current `delete` API

These are now equivalent:
```
@my_map[1, "hi"] = 1;
@my_map[(1, "hi")] = 1;
```

##### Checklist


jordalgo:
@danobi I had to add more tests to aot_skip list because AOT seems to have an issue with tuple printing.

danobi:
```
The map above is now printed:

@my_map[(1, hi)]: 1

Instead of:

@my_map[1, hi]: 1
```

Why change this? Semantically they are identical but adding `()` adds more visual noise. On top of that, this is no longer backwards compatible for anyone parsing output.

It should be straightforward to keep the output the same but only change internals right?

danobi:
Yeah I think if we did this clean slate I would've leaned towards keeping the parenthesis in just for implementation simplicity. But I would also said it's b/c I'm lazy and not b/c I think it's a better result. Given that the parenthesis is really only a detail for output.cpp (I think), I'd lean towards keeping the output fixed here

src/ast/dibuilderbpf.cpp
```diff
@@ -227,8 +227,8 @@ DIType *DIBuilderBPF::GetMapKeyType(const MapKey &key,
     extra_arg_size = 8;
 
   // Single map key -> use the appropriate type.
-  if (key.args_.size() == 1 && extra_arg_size == 0)
-    return GetType(key.args_[0]);
+  if (extra_arg_size == 0)
+    return GetType(key.arg_);
 
   // Multi map key -> use byte array.
```
danobi, line 233: I need to read the rest of this PR more carefully, but "multi map key" logic can be deleted. After this change, all maps will have 0 or 1 keys. My hope was that it'll get rid of a lot of redundant code.

viktormalik: This alone would be a good reason for going forward with this :-)

viktormalik:
I didn't get a chance to review the code so far but I'm in favor of this change. There's no point in special-casing multi-value keys when we can make them tuples and maintain just one implementation of things.

I think that it's better to print the keys without parentheses (i.e. `@my_map[1, hi]: 1`). My argument would be that bpfscript is a high-level language so we should slightly prefer readability and simplicity over pedantic technical consistency. In this case, there's no point in adding the visual noise (just for the sake of being consistent with the implementation) and the documentation will probably use the no-parenthesis form, too.

src/ast/dibuilderbpf.cpp
jordalgo: Maybe. Right now we still have multi-map keys for `hist` and `lhist`. I can look at refactoring that but it might be adding more scope to this change.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -184,7 +184,7 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
 
   std::vector<llvm::Type *> stack_key_elements = {
     b_.getInt64Ty(), // stack id
-    b_.getInt32Ty(), // nr_stack_frames
+    b_.getInt64Ty(), // nr_stack_frames
```
jordalgo, line 187: I was having some issues with the stack id when setting the map key type as a tuple with a int64 and a int32 - maybe an alignment thing?

jordalgo: Yeah, it's because of padding. I could use an array here instead of a padded struct for the key if we wanted to save 4 bytes, which maybe is worth it for when we have a lot of stacks.

ajor (commented):
In my recent AST refactorings I've moved us away from representing empty lists with null pointers in an attempt to simplify the code. An empty ExpressionList is now just represented by an empty vector - this means that we don't need to check for nullptr before using them and can just try iterating over the list. If it's empty the loop will run 0 times.

From a quick skim, it looks like the approach in this PR is:
- 0 keys: nullptr
- 1 key: non-tuple Expression
- 2+ keys: Tuple

But how about this instead:
- 0 keys: 0-element tuple
- 1 key: 1-element tuple
- 2+ keys: n-element tuple

man/adoc/bpftrace.adoc
```diff
@@ -2465,15 +2477,15 @@ interval:s:10 {
...
+For a map with multiple keys, the map and key are combined in the same argument.
 Multiple arguments can be passed to delete many keys at once.
 
 ```
 kprobe:dummy {
   @scalar = 1;
   @associative[1,2] = 1;
   delete(@scalar);
-  delete(@associative[1,2]);
+  delete(@associative[(1,2)]); // This is equivalent delete(@associative[1,2]);
```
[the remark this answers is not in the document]
jordalgo: what if I just flip it around so the tuple syntax is in the comments? I'm generally pro more documentation when it comes to stuff that can work both ways like this.

jordalgo:
> In my recent AST refactorings I've moved us away from representing empty lists with null pointers in an attempt to simplify the code. An empty ExpressionList is now just represented by an empty vector - this means that we don't need to check for nullptr before using them and can just try iterating over the list. If it's empty the loop will run 0 times.
> But how about this instead:
> 0 keys: 0-element tuple
> 1 key: 1-element tuple
> 2+ keys: n-element tuple

This was a good suggestion and I tried VERY hard (many, many hours) to make it work as well as this current solution but I ran into some edge cases that were problematic. In short, when you attempt to wrap everything in a Tuple, even if it's a single value, you have to do some unwrapping if that single value is already a tuple e.g. `@map[(1, 2)]` or `$a = (1, 2); @map[$a]`. And we want to do this so that these two expressions are equivalent `@map[(1, 2)]`  and `@map[1, 2]`. 

However in the case of `$a = (1, 2); @map[$a]`, you don't have access to the underlying tuple expressions because the actual AST object is a Variable (not a Tuple). 

Additionally, when everything is wrapped in a Tuple, it's more difficult to manipulate the Tuple elements when you want to account for larger strings e.g. `@map["he", 2] = 3; @map[("hellasdfasdfo", 3)] = 4;`, which we already do for map keys.

Here are some example programs that my all tuple implementation can't handle:
```
BEGIN { for ($kv : @map) { @map[$kv.0] } @map[0] = 1; }
```
```
BEGIN { @map[1, 2] = 1; for ($kv : @map) { @map[$kv.0] = 2; } }
```
```
BEGIN { $a = (3, "hi"); @map[1, "by"] = 1; @map[$a] = 2; exit(); }
```

[Here is the all tuple implementation](https://github.com/jordalgo/bpftrace/tree/map_tuple_keys_all_tuples) if you want to check it out.

What's accidentally nice about this current implementation is that single values that are both tuples, as variables and tuple literals, don't get wrapped and are just handled as if they were a standard multi-map key. This was a good exercise though and I added a few more changes to this PR to make it work a little more smoothly - it also made me want to audit our AST -> SizeType -> SemanticAnalysis pipeline.

src/ast/ast.cpp
```diff
@@ -117,13 +117,11 @@ Map::Map(const std::string &ident, location loc) : Expression(loc), ident(ident)
   is_map = true;
 }
 
-Map::Map(const std::string &ident, ExpressionList &&vargs_, location loc)
-    : Expression(loc), ident(ident), vargs(std::move(vargs_))
+Map::Map(const std::string &ident, Expression *key_expr, location loc)
+    : Expression(loc), ident(ident), key_expr(key_expr)
```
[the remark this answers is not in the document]
jordalgo: This didn't seem to work for me. I also didn't see any other examples in the ast file aside from things that take an r-value.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1464,29 +1464,73 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+  bool is_tuple_literal = false;
+
+  if (map.key_expr) {
+    Visit(map.key_expr);
+    Expression *expr = map.key_expr;
+
+    if (expr->type.IsTupleTy()) {
+      // We can only do this expression and size type manipulation
+      // if we have access to the underlying expressions in a tuple literal
+      // e.g. @map[(1, 2)] vs $a = (1, 2) @ma[$a]
```
danobi, line 1476:
```suggestion
      // e.g. @map[(1, 2)] vs $a = (1, 2); @map[$a]
```

src/mapkey.h
danobi, line 1: Is `MapKey` abstraction still useful? Seems like all uses can be replaced with `SizedType`. Right now I think we have 3 places where values are deserialized: `Output::value_to_str`, `BPFtrace::get_arg_values`, and here. Would be good to eventually only have 1. And surely one of the two remaining can be repurposed for map keys which is now either `nullptr` or `SizedType`

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1464,29 +1464,73 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+        std::vector<SizedType> elements;
+        Tuple *tuple = static_cast<Tuple *>(map.key_expr);
+        // NOLINTBEGIN(modernize-loop-convert)
+        for (unsigned int i = 0; i < tuple->elems.size(); i++) {
+          // NOLINTEND(modernize-loop-convert)
+          Visit(tuple->elems.at(i));
+          Expression *elem = tuple->elems.at(i);
+
+          if (elem->type.IsNoneTy() || elem->type.GetSize() == 0)
+            return;
```
danobi, line 1488: Is this an error? And why the return?

src/ast/ast.cpp
```diff
@@ -181,6 +179,7 @@ Cast::Cast(SizedType cast_type, Expression *expr, location loc)
 Tuple::Tuple(ExpressionList &&elems, location loc)
     : Expression(loc), elems(std::move(elems))
 {
+  is_literal = true;
```
danobi, line 182: Not sure this is correct. Isn't this saying all tuples are literal? Perhaps you want to check that all fields in the tuple are literal as well? But I'm unsure what the use case is.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3402,32 +3447,70 @@ void SemanticAnalyser::accept_statements(StatementList &stmts)
...
+        } else {
+          for (size_t i = 0; i < keyFields.size(); i++) {
+            SizedType &key_type = keyFields[i].type;
+            const SizedType &new_key_type = newKeyFields[i].type;
+            if (key_type.GetTy() != new_key_type.GetTy()) {
+              valid = false;
+              break;
+              // We can't exchange the keys if we don't have access to the
+              // underlying expressions
+            } else if (key_type.IsStringTy() && is_tuple_literal) {
```
danobi, line 3475: Related to comment about all tuples being literal, I'm not sure why a literal check is necessary here.

danobi (commented):
took a mostly high level look at code. some questions and suggestions. will probably want to approach this with multi-pass review process

> This was a good suggestion and I tried VERY hard (many, many hours) to make it work as well as this current solution but I ran into some edge cases that were problematic

big TY for giving it a shot. Worth a try :p

man/adoc/bpftrace.adoc
```diff
@@ -940,28 +940,40 @@ All valid identifiers can be used as `name`.
 
 The data type of a variable is automatically determined during first assignment and cannot be changed afterwards.
 
-==== Associative Arrays
+==== Map Keys
```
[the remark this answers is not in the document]
jordalgo: Yeah good call.

src/ast/passes/semantic_analyser.cpp
jordalgo: I guess the comment above this statement didn't make it clearer 😞 Only in the case of a tuple literal (not a variable that is a tuple type) can we get access to the underlying expressions and manipulate their types. This is for cases like this:
```
@[(1, "hi")] = 1;
@[(2, "hello")] = 2;
```
We have to increase the string size of "hi" and, because of tuple alignment (comment below), we just replace the whole stored tuple type with the larger one.

src/ast/ast.cpp
jordalgo: It's just saying that the tuple this expression: `(1, 2, 3)` is a literal, whereas this `$a`, which was previously initialized as a tuple `$a = (1,2,3)` is not. This is used for type manipulation in semantic_analyser for concrete expressions like `@map[(1, 2)] = 1;` vs `@map[$a] = 1;`

danobi: But my reading of this code is that something like `(1, 2 $three)` will be marked as a literal. Is there something I'm missing?

Perhaps you actually want a `dyn_cast` to check that the AST node is directly a tuple. `is_literal` is more for integers/strings

src/mapkey.h
jordalgo: This is a good question. Let me look into removing it.

src/ast/ast.cpp
jordalgo: > Perhaps you actually want a dyn_cast to check that the AST node is actually a tuple. is_literal is more for integers/strings

Ah ok. Yeah, this is better. Will change.

src/ast/passes/semantic_analyser.cpp
jordalgo: Added a comment which I borrowed from `visit(Tuple)`

tests/bpftrace.cpp
```diff
@@ -1021,48 +1021,19 @@ TEST(bpftrace, sort_by_key_int)
   EXPECT_THAT(values_by_key, ContainerEq(expected_values));
 }
 
-TEST(bpftrace, sort_by_key_int_int)
```
viktormalik, line 1024: Sorting by multiple keys is now broken, right? Is that expected?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1463,30 +1463,78 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+
+    if (expr->type.IsTupleTy()) {
+      // We can only do this expression and size type manipulation
+      // if we have access to the underlying expressions in the tuple ast
+      // e.g. @map[(1, 2)] vs $a = (1, 2) @map[$a]
+      auto tuple_ast = dynamic_cast<Tuple *>(expr);
+      if (tuple_ast) {
+        is_tuple_ast = true;
+        std::vector<SizedType> elements;
+        Tuple *tuple = static_cast<Tuple *>(map.key_expr);
```
viktormalik, line 1481: Why not use `tuple_ast` directly?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1463,30 +1463,78 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+          if (is_final_pass() && elem->type.IsNoneTy())
+            LOG(ERROR, elem->loc, err_)
+                << "Invalid expression for assignment: ";
 
-  for (unsigned int i = 0; i < map.vargs.size(); i++) {
-    Visit(map.vargs.at(i));
-    Expression *expr = map.vargs.at(i);
+          SizedType keytype = elem->type;
+          if (keytype.IsIntTy()) {
+            keytype = CreateUInt(keytype.GetSize() * 8);
```
viktormalik, line 1516: What does this do?

viktormalik (commented):
Overall looks like a useful change. I've a couple of comments and questions, though.

src/ast/passes/semantic_analyser.cpp
jordalgo: This matches what we currently do for map keys, we increase the size of all ints to 8 bytes large - this probably needs further fixing though. https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L1499

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3639,7 +3641,7 @@ void CodegenLLVM::generate_maps(const RequiredResources &resources)
     createMapDefinition(stack_type.name(),
                         libbpf::BPF_MAP_TYPE_LRU_HASH,
                         128 << 10,
-                        MapKey({ CreateUInt64(), CreateUInt32() }),
+                        debug_.CreateByteArrayType(12),
```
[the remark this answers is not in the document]
jordalgo: I think I had this as a tuple originally but they are padded and will be larger than 12 bytes if we have a int64 and a int32, which is what makes up the internal stack id. But I was able to make this an Array sized type e.g. `CreateArray(12, CreateInt8())` so I can get rid of the overload of `createMapDefinition`.

tests/bpftrace.cpp
jordalgo: I think I forgot about this as I was deleting. Let me fix.

jordalgo: Ok, re-added. Thanks for catching this.

src/ast/passes/semantic_analyser.cpp
danobi: Ah ok. I was confused with the literal check. with dyn_cast this all makes sense

src/ast/passes/semantic_analyser.cpp
viktormalik: But this just converts all int keys to uint, doesn't it? I was wondering why do we need that.

(the above link is probably outdated and points to some unrelated code so I cannot really compare)

jordalgo: > But this just converts all int keys to uint, doesn't it? I was wondering why do we need that.

Yeah I was scratching my head about that too. I think our whole int conversion strategy needs re-evaluation but it also puts us in this situation of potentially breaking backward compatibility like with this recent change: https://github.com/bpftrace/bpftrace/pull/3416

I updated the link above.

viktormalik: I see, ok, it's strange but it seems to be working so let's keep it for now.

viktormalik (approved):
Nice work and kudos for another simplification of the codebase!

jordalgo:
@viktormalik Thanks for the review! I'll wait at least a day or two before merging to give other folks a chance to voice concerns.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1529,30 +1529,77 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+      // We can only do this expression and size type manipulation
+      // if we have access to the underlying expressions in the tuple ast
+      // e.g. @map[(1, 2)] vs $a = (1, 2) @map[$a]
+      auto tuple_ast = dynamic_cast<Tuple *>(expr);
+      if (tuple_ast) {
+        is_tuple_ast = true;
+        std::vector<SizedType> elements;
+        // NOLINTBEGIN(modernize-loop-convert)
+        for (unsigned int i = 0; i < tuple_ast->elems.size(); i++) {
+          // NOLINTEND(modernize-loop-convert)
```
ajor, line 1549: Why not accept the lint and use a range-based for-loop?

tests/semantic_analyser.cpp
```diff
@@ -439,10 +440,37 @@ BEGIN { @x[1] = 0; @x[2,3]; }
...
+                                              ~~~~~~~~
+)");
+
+  test_error(
+      R"(BEGIN { $a = (3, "hello"); $b = (4, "hi"); @map[$a] = 1; @map[$b] = 2; })",
+      R"(
+stdin:1:58-66: ERROR: Argument mismatch for @map: trying to access with arguments: '(int64,string[3])' when map expects arguments: '(int64,string[6])'
+BEGIN { $a = (3, "hello"); $b = (4, "hi"); @map[$a] = 1; @map[$b] = 2; }
+                                                         ~~~~~~~~
+)");
```
ajor, line 473: These expected failures are a bit annoying - the sizes of strings doesn't make a difference anywhere else so it's a bit strange to require them to exactly match here

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1529,30 +1529,77 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+  bool is_tuple_ast = false;
+
+  if (map.key_expr) {
+    Visit(map.key_expr);
+    Expression *expr = map.key_expr;
+
+    if (expr->type.IsTupleTy()) {
+      // We can only do this expression and size type manipulation
+      // if we have access to the underlying expressions in the tuple ast
+      // e.g. @map[(1, 2)] vs $a = (1, 2) @map[$a]
```
ajor, line 1542: Is this definitely true? We have `expr->type` which holds the type information of the tuple, no matter whether it's a literal tuple or a variable / map holding a tuple. I'd have thought we'd be able to get the type information out of that without looking at the AST directly.

This'd let us get rid of that incompatible string-length limitation when using non-literal tuples.

src/bpftrace.h
```diff
@@ -195,7 +195,7 @@ class BPFtrace {
   bool need_recursion_check_ = false;
 
   static void sort_by_key(
-      std::vector<SizedType> key_args,
+      SizedType key,
```
ajor, line 198:
```suggestion
      const SizedType &key,
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3531,35 +3576,75 @@ void SemanticAnalyser::accept_statements(StatementList &stmts)
...
+            SizedType &key_type = keyFields[i].type;
+            const SizedType &key_field_type = newKeyFields[i].type;
+            if (key_type.GetTy() != key_field_type.GetTy()) {
+              valid = false;
+              break;
+              // We can't exchange the keys if we don't have access to the
+              // underlying expressions
+            } else if (key_type.IsStringTy() && is_tuple_ast) {
+              // We have to replace the whole key type because of tuple
+              // alignment
```
ajor, line 3606: Why restrict this to literal tuples in the AST? If we relaxed this then that incompatible string-length limitation when using non-literal tuples would be gone.

ajor (changes requested):
If we can possibly get rid of the string-length matching requirement then we should do. Up until now, the type of a string is just "string" and the length doesn't matter to users.

src/ast/passes/semantic_analyser.cpp
jordalgo: Leftover from older code. I'll switch it to the range-based

src/ast/passes/semantic_analyser.cpp
jordalgo: Good catch. I think this was leftover from when I was attempting to manually update individual keys. I'll remove the `is_tuple_ast` check.

janca-ucdavis:
> I think that it's better to print the keys without parentheses (i.e. `@my_map[1, hi]: 1`). My argument would be that bpfscript is a high-level language so we should slightly prefer readability and simplicity over pedantic technical consistency. In this case, there's no point in adding the visual noise (just for the sake of being consistent with the implementation) and the documentation will probably use the no-parenthesis form, too.

What about JSON output?  My preference would be backwards compatibility, supporting multiple versions of bpftrace is a hassle if one parser has to handle different output formats.

jordalgo:
> What about JSON output? My preference would be backwards compatibility, supporting multiple versions of bpftrace is a hassle if one parser has to handle different output formats.

@janca-ucdavis Right now it is 99% backward compatible. The only small difference is that for json output there is now a space between the keys, which actually makes the json output now consistent with the standard output.

e.g. this `-e 'BEGIN { @a[1, 2] = 1; exit(); }' -f json`

Now outputs this `{"type": "map", "data": {"@a": {"1, 2": 1}}}` instead of this `{"type": "map", "data": {"@a": {"1,2": 1}}}` (notices the space between "1," and "2".

jordalgo:
> Is there any performance delta? I'm using 5-dimensional maps now and expect to go to 7 soon.

There shouldn't be. I think the only difference is a slight increase in required memory for multi-keys as they are now stored in padded structs.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3531,35 +3570,72 @@ void SemanticAnalyser::accept_statements(StatementList &stmts)
...
+            } else if (key_type.IsStringTy()) {
+              // We have to replace the whole key type because of tuple
+              // alignment/padding
+              if (key_field_type.GetSize() > key_type.GetSize()) {
+                replace_key = true;
+              }
+            } else if (key_type.GetSize() != key_field_type.GetSize()) {
+              valid = false;
+              break;
+            }
```
ajor, line 3604: This type checking currently only checks two levels deep, but needs to be recursive to catch mismatched nested types.

I can break it with this script:
```
BEGIN
{
  $a = ((uint8)1,((uint8)123,(uint64)1234));
  $b = ((uint8)4,((uint64)1234,(uint8)123));
  @map[$a] = 1;
  @map[$b] = 2;
}
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2285,21 +2327,18 @@ void SemanticAnalyser::visit(For &f)
   auto *mapkey = get_map_key_type(map);
   auto *mapval = get_map_type(map);
 
-  if (mapkey && mapkey->args_.size() == 0) {
-    LOG(ERROR, map.loc, err_)
-        << "Maps used as for-loop expressions must have keys to iterate over";
+  if (!mapkey || mapkey->IsNoneTy()) {
```
ajor, line 2330:
```suggestion
  if (mapkey && mapkey->IsNoneTy()) {
```
This logic can be kept similar to the old code, right? (I might be missing something)

I try to return error messages as soon as possible and only resort to `is_final_pass` when absolutely necessary. That way the most useful error messages tend to get printed out first, and fewer bogus messages will be displayed to the user (i.e. those errors which were deferred to the final pass, if we error out before the final pass)

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1529,30 +1529,71 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+
+  if (map.key_expr) {
+    Visit(map.key_expr);
+    Expression *expr = map.key_expr;
+
+    if (expr->type.IsTupleTy()) {
+      // We can only do this expression and size type manipulation
+      // if we have access to the underlying expressions in the tuple ast
+      // e.g. @map[(1, 2)] vs $a = (1, 2) @map[$a]
+      auto tuple_ast = dynamic_cast<Tuple *>(expr);
```
ajor, line 1542: What's this manipulation needed for?

If it's necessary, isn't it necessary for non-literal tuples as well? If it's not, then can this code just be deleted?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1529,30 +1529,71 @@ Probe *SemanticAnalyser::get_probe_from_scope(Scope *scope,
...
+            return;
+
+          // Insert a cast to 64 bits if needed by injecting
+          // a cast into the ast.
+          if (elem->type.IsIntTy() && elem->type.GetSize() < 8) {
+            Expression *cast = ctx_.make_node<Cast>(elem->type.IsSigned()
+                                                        ? CreateInt64()
+                                                        : CreateUInt64(),
+                                                    elem,
+                                                    map.loc);
```
ajor, line 1561: This only works two-levels deep as well. This script breaks it:
```
BEGIN { @map[(1, ((uint16)2,3))] = 1; }
```

But see my comment above - is it even necessary?

src/ast/passes/semantic_analyser.cpp
jordalgo: This is to maintain backward compat with the current multi-key implementation where we insert casts and increase the size of ints to 64 bits. We can't do this for non-literal tuples (as the comment says) because we don't have access to the underlying expressions for non-literal tuples.

tests/codegen/llvm/call_path.ll
```diff
@@ -33,7 +33,7 @@ lookup_str_merge:                                 ; preds = %entry
   %1 = ptrtoint ptr %0 to i64
   %2 = getelementptr i64, ptr %0, i64 0
   %filp = load volatile i64, ptr %2, align 8
-  %3 = add i64 %filp, 152
+  %3 = add i64 %filp, 16
```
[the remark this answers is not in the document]
jordalgo: Ah weird. I'll change it back.

src/ast/passes/semantic_analyser.cpp
jordalgo: Wrangling nested tuples and attempting to modify their nested expressions/types is very complicated and out of scope for this PR so I just added a restriction on using nested tuples as map keys - also added a test for it.

src/ast/passes/semantic_analyser.cpp
jordalgo: I actually think this isn't necessary - it was here before this PR but I yanked out this cast as we can just set the map key integer type to Int64 and everything works fine.

src/ast/passes/semantic_analyser.cpp
jordalgo: Related: https://github.com/bpftrace/bpftrace/issues/3439

tests/semantic_analyser.cpp
```diff
@@ -3328,9 +3364,9 @@ TEST(semantic_analyser, string_size)
        0);
   stmt = driver.ctx.root->probes.at(0)->stmts.at(0);
   map_assign = dynamic_cast<ast::AssignMapStatement *>(stmt);
-  ASSERT_EQ(map_assign->map->key_type.size(), 14UL);
-  ASSERT_TRUE(map_assign->map->key_type.args_.at(0).IsStringTy());
-  ASSERT_EQ(map_assign->map->key_type.args_.at(0).GetSize(), 6UL);
+  ASSERT_EQ(map_assign->map->key_type.GetSize(), 16UL);
+  ASSERT_TRUE(map_assign->map->key_type.IsTupleTy());
+  ASSERT_EQ(map_assign->map->key_type.GetSize(), 16UL);
```
ajor, line 3369:
```suggestion
  ASSERT_EQ(map_assign->map->key_type.GetSize(), 16UL);
```
This line is checking the same as the line two above it now instead of checking the string size

tests/semantic_analyser.cpp
ajor, line 1: What's changed all the integer types from `unsigned int64` to `int64` in these tests?

ajor:
Preventing nested tuples from being used as map keys seems a reasonable first step, but I really do think we need to make strings in tuples work transparently, regardless of length. Let's talk outside of GitHub.

Also just sharing this strange error message I got while playing around with strings in tuples:
```
# bpftrace -e 'BEGIN { $a = (3, "hello"); @m[$a] = 1; $a = (1,"aaaaaaaaaa"); @m[$a] = 2; }'
stdin:1:28-34: ERROR: Argument mismatch for @m: trying to access with arguments: '(int64,string[11])' when map expects arguments: '(int64,string[11])'
BEGIN { $a = (3, "hello"); @m[$a] = 1; $a = (1,"aaaaaaaaaa"); @m[$a] = 2; }
                           ~~~~~~
stdin:1:63-69: ERROR: Argument mismatch for @m: trying to access with arguments: '(int64,string[11])' when map expects arguments: '(int64,string[11])'
BEGIN { $a = (3, "hello"); @m[$a] = 1; $a = (1,"aaaaaaaaaa"); @m[$a] = 2; }
                                                              ~~~~~~
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2285,21 +2343,18 @@ void SemanticAnalyser::visit(For &f)
...
+  if (!mapval)
     return;
 
-  auto keytype = CreateNone();
-  if (mapkey->args_.size() == 1) {
-    keytype = mapkey->args_[0];
-  } else {
-    keytype = CreateTuple(bpftrace_.structs.AddTuple(mapkey->args_));
+  if (mapkey && mapkey->IsNoneTy()) {
+    if (is_final_pass()) {
```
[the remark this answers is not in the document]
jordalgo: I think maybe I was confused by your original comment. I'll change it back to check if the mapkey is null and I'll remove the `is_final`.

tests/semantic_analyser.cpp
jordalgo: (discussed offline) I made the default key type Int64 (instead of Uint64) to match the default for integers everywhere else.

src/output.cpp
```diff
@@ -99,6 +99,121 @@ std::ostream &operator<<(std::ostream &out, MessageType type)
   return out;
 }
 
+std::string map_key_str(BPFtrace &bpftrace,
```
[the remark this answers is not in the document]
jordalgo: Good call. Mind if I do this in a follow-up PR?

jordalgo:
- Fixed the issue with this and added a semantic_analyser test for it: `bpftrace -e 'BEGIN { $a = (3, "hello"); @m[$a] = 1; $a = (1,"aaaaaaaaaa"); @m[$a] = 2; }'`.
- Stacked this PR on top of this one as it has fixes for tuple string resizing: https://github.com/bpftrace/bpftrace/pull/3443
- Simplified the Map ast class and am now just using `key_expr`

CHANGELOG.md
```diff
@@ -36,6 +38,8 @@ and this project adheres to
   - [#3249](https://github.com/bpftrace/bpftrace/pull/3249)
 - Faster map access for keyless maps by using BPF_MAP_TYPE_ARRAY
   - [#3300](https://github.com/bpftrace/bpftrace/pull/3300)
+- Multi map keys are stored as tuples
+  - [#3390](https://github.com/bpftrace/bpftrace/pull/3390/)
```
ajor, line 42: Is this something users can see? I'd not include it if it's just an implementation detail. The other changelog entry is definitely something users would care about though

src/ast/passes/semantic_analyser.cpp
ajor: Now that #3443 has been merged (and might have fixed #3439), is this code needed any more?

jordalgo: I tried removing it and some tests are still failing e.g. `Casting retval should work`. Mind if I work on fixing this in a follow-up PR? It might require extended investigation.

jordalgo: Nevermind. I figured it out and removed the casting.

src/output.cpp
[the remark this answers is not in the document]
jordalgo: Nm, fixed it here.

jordalgo:
@ajor I think I finally addressed all the concerns 🤞

src/ast/passes/semantic_analyser.cpp
jordalgo: The example above works now.

danobi:
I will take another look this week

src/types.h
```diff
@@ -241,6 +241,7 @@ class SizedType {
   bool operator==(const SizedType &t) const;
   bool operator!=(const SizedType &t) const;
   bool IsSameType(const SizedType &t) const;
+  bool IsSameSizeRecursive(const SizedType &t) const;
```
danobi, line 244: Leave a comment about why the recursive bit is necessary?

src/output.cpp
```diff
@@ -236,163 +236,236 @@ std::string Output::get_helper_error_msg(int func_id, int retcode) const
...
+      for (auto &field : arg.GetFields()) {
+        std::vector<uint8_t> elem_data(data.begin() + field.offset,
+                                       data.begin() + field.offset +
+                                           field.type.GetSize());
+        elems.push_back(
+            value_to_str(bpftrace, field.type, elem_data, false, 1, true));
+      }
+      return tuple_to_str(elems, true);
+    }
+    default:
```
danobi, line 463: Instead of default, list out all the other types? Compiler will warn if new type is added

src/output.cpp
```diff
@@ -236,163 +236,236 @@ std::string Output::get_helper_error_msg(int func_id, int retcode) const
...
+    case Type::strerror: {
+      return strerror(read_data<uint64_t>(value.data()));
+    }
+    case Type::pointer:
+    case Type::reference: {
+      std::ostringstream res;
+      res << "0x" << std::hex << read_data<uint64_t>(value.data());
+      return res.str();
+    }
+    default: {
```
danobi, line 423: ditto

danobi (commented):
looks quite reasonable. tests and ast changes look good. I got a bit lost in semantic analysis and codegen changes. hard to tell apart refactor changes with actual changes.

jordalgo:
@danobi - Thanks for the review! I made the updates.

> I got a bit lost in semantic analysis and codegen changes. hard to tell apart refactor changes with actual changes.

Sorry about that. This change ended up being a lot more complicated than I anticipated. Hopefully the new added tests make it feel safer to merge.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3698,7 +3718,7 @@ void CodegenLLVM::generate_maps(const RequiredResources &required_resources,
     createMapDefinition(stack_type.name(),
                         libbpf::BPF_MAP_TYPE_LRU_HASH,
                         128 << 10,
-                        MapKey({ CreateUInt64(), CreateUInt32() }),
+                        CreateArray(12, CreateInt8()),
```
ajor, line 3721: Nit: would something like this work to make it more descriptive?
```suggestion
                        CreateTuple({ CreateUInt64(), CreateUInt32() }),
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2948,10 +2960,7 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
 AllocaInst *CodegenLLVM::getMultiMapKey(Map &map,
                                         const std::vector<Value *> &extra_keys)
 {
-  size_t size = 0;
-  for (auto &key_type : map.key_type.args_) {
-    size += key_type.GetSize();
-  }
+  size_t size = map.key_expr->type.GetSize();
```
ajor, line 2963: This previously read from `map.key_type` (the computed size of the map key across all AST nodes) but now uses `map.key_expr` (the size of a specific AST node).

I haven't triggered any obviously bad behaviour, but this script shows some bad generated code which I imagine could lead to issues with a bit more fiddling:
```
BEGIN { @[1,"aaaaaaaaaaaaaaaaaaaaaaaaaaa"] = hist(0); @[1,"a"] = hist(0);}
```
Different key sizes on each access:
```
  %"@_key2" = alloca [24 x i8], align 1
  %"@_key" = alloca [48 x i8], align 1
```
Out of bounds indexing:
```
  %14 = getelementptr [24 x i8], ptr %"@_key2", i64 0, i64 40
```

ajor (commented):
Sema looks good to me but I'll have to finish going over codegen tomorrow

tests/semantic_analyser.cpp
```diff
@@ -3354,9 +3381,11 @@ TEST(semantic_analyser, string_size)
        0);
   stmt = driver.ctx.root->probes.at(0)->stmts.at(0);
   map_assign = dynamic_cast<ast::AssignMapStatement *>(stmt);
-  ASSERT_EQ(map_assign->map->key_type.size(), 14UL);
-  ASSERT_TRUE(map_assign->map->key_type.args_.at(0).IsStringTy());
-  ASSERT_EQ(map_assign->map->key_type.args_.at(0).GetSize(), 6UL);
+  ASSERT_TRUE(map_assign->map->key_expr->type.IsTupleTy());
+  ASSERT_TRUE(map_assign->map->key_expr->type.GetField(0).type.IsStringTy());
+  ASSERT_EQ(map_assign->map->key_expr->type.GetField(0).type.GetSize(), 3UL);
```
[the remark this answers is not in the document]
jordalgo: Keen eyes you have! Good catch, going to add this line right under this one:
```
ASSERT_EQ(map_assign->map->key_type.GetField(0).type.GetSize(), 6UL);
```

src/ast/passes/codegen_llvm.cpp
jordalgo: I tried that at first but there was an issue with resolving the stack id. Let me work on fixing this in a follow-up PR.

src/types.cpp
```diff
@@ -559,6 +569,32 @@ std::weak_ptr<const Struct> SizedType::GetStruct() const
...
+
+    for (ssize_t i = 0; i < GetFieldCount(); i++) {
+      if (GetField(i).type.GetSize() != t.GetField(i).type.GetSize()) {
+        return false;
+      }
+      if (GetField(i).type.IsTupleTy() && t.GetField(i).type.IsTupleTy()) {
+        if (!GetField(i).type.IsSameSizeRecursive(t.GetField(i).type)) {
+          return false;
+        }
+      }
```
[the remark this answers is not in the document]
jordalgo: duh - good catch.

src/output.cpp
```diff
@@ -236,161 +236,256 @@ std::string Output::get_helper_error_msg(int func_id, int retcode) const
...
+    case Type::inet:
+    case Type::username:
+    case Type::probe:
+    case Type::string:
+    case Type::buffer:
+    case Type::pointer:
+    case Type::array:
+    case Type::mac_address:
+    case Type::record:
+      return value_to_str(bpftrace, arg, data, false, 1, true);
```
ajor, line 459: Just checking: any reason for the last argument to always be true?

(if I'm understanding correctly, it doesn't make a difference what value it has)

ajor (approved):
Looks good to me!

src/output.cpp
jordalgo: This is needed for json output mode e.g. for this program
`bpftrace -e 'BEGIN { @["hello"] = 1; }' -f json`

without setting this to true you get this:
`{"type": "map", "data": {"@": {"\"hello\"": 1}}}`

whereas you want this:
`{"type": "map", "data": {"@": {"hello": 1}}}`

jordalgo:
Thanks for the reviews @viktormalik @danobi @ajor ! Going to merge shortly 🚀

PR #3398, opened 2024-08-15
Minor refactoring of BpfBytecode to support loading third-party BPF objects

ajor:
This is the first step towards interop with externally written BPF programs.

```
BpfBytecode: Accept std::span input parameter instead of pointer and size
```
```
BpfBytecode: Move updating global variables into its own function

To support linking against third-party BPF libraries, it would be useful
to have the constructor do nothing but parse a BPF bytecode blob.
```
```
BpfBytecode: Don't require log_size parameter in ctor

The log size is only relevant when loading programs, so move the
parameter onto the load_progs() function.

This lets us more easily use BpfBytecode to represent bytecode which we
don't intend to load, e.g. when analysing third-party BPF libraries.
```

### DIFF
--- a/src/aot/aot.cpp
+++ b/src/aot/aot.cpp
@@ -340,7 +340,7 @@ int load(BPFtrace &bpftrace, const std::string &in)
 
   bpftrace.bytecode_ = BpfBytecode(
       std::span<uint8_t>{ btaot_section + hdr->elf_off, hdr->elf_len },
-      bpftrace);
+      bpftrace.config_.get(ConfigKeyInt::log_size));
   if (err)
     goto out;
 
--- a/src/aot/aot.cpp
+++ b/src/aot/aot.cpp
@@ -338,9 +338,9 @@ int load(BPFtrace &bpftrace, const std::string &in)
   if (err)
     goto out;
 
-  bpftrace.bytecode_ = BpfBytecode(btaot_section + hdr->elf_off,
-                                   hdr->elf_len,
-                                   bpftrace);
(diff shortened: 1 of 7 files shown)


danobi (approved):
looks like a good cleanup to me.

one thing I did notice though is libbpf is able to internally allocate and resize program log buffers. should we just use that functionality instead? can probably delete all the log buffer wrangling stuff as well as the env var.

see https://github.com/libbpf/libbpf/commit/90910812b59d841904418fa7eaba388eadf68c2f

viktormalik (approved):

viktormalik:
> one thing I did notice though is libbpf is able to internally allocate and resize program log buffers. should we just use that functionality instead? can probably delete all the log buffer wrangling stuff as well as the env var.
> 
> see [libbpf/libbpf@9091081](https://github.com/libbpf/libbpf/commit/90910812b59d841904418fa7eaba388eadf68c2f)

As long as we can access the log buffer after loading to analyse its contents, I'm all in for using libbpf-allocated buffers.

ajor:
> one thing I did notice though is libbpf is able to internally allocate and resize program log buffers. should we just use that functionality instead? can probably delete all the log buffer wrangling stuff as well as the env var.
> 
> see [libbpf/libbpf@9091081](https://github.com/libbpf/libbpf/commit/90910812b59d841904418fa7eaba388eadf68c2f)

From a skim of that commit, it looks like if libbpf owns the log buffer then it'll print out its own debug on errors and will free the buffer on success, which wouldn't be suitable for bpftrace.

PR #3405, opened 2024-08-19
nix: Disable hardening flags for nix dev builds

danobi:
Hardening is not really useful for dev builds and for CI. It causes pointless spew, so disable it.

##### Checklist


jordalgo (approved):

ajor:
The Nix build is also used for the release packages we publish ourselves, right? We should probably keep hardening for them

danobi:
Yes, but we don't do the release build in the nix shell. It goes thru the regular package build which keeps the hardening flags

PR #3408, opened 2024-08-20
Dry run: terminate bpftrace as soon as possible

### LINKED ISSUE #3330: `--dry-run` should not run programs
`--dry-run` does cause bpftrace to exit, but only after its probes have fired thousands of times in this example:

```
# bpftrace --dry-run -e 't:sched:sched_switch { @ = count(); }' -d ast

AST after: parser
-------------------
Program
 tracepoint:sched:sched_switch
  =
   map: @
   call: count

[...]

AST after: ReturnPath
-------------------
Program
 tracepoint:sched:sched_switch
  =
   map: @ :: [count]
   call: count :: [count]

Attaching 1 probe...


@: 5396
```

viktormalik:
The --dry-run option is supposed to terminate bpftrace before actually running the program. This is a tricky thing to do as the probes start to hit as soon as BPF programs are attached to them.

Despite that, we can still minimize the amount of time that the probes are running by terminating bpftrace immediately after all probes are attached instead of doing it only after we start polling perf events.

Resolves #3330.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1014,6 +1014,11 @@ int BPFtrace::run(BpfBytecode bytecode)
     }
   }
 
+  if (dry_run) {
+    request_finalize();
+    return 0;
+  }
+
   // Kick the child to execute the command.
   if (child_) {
     try {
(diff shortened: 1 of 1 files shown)


viktormalik:
Not updating the changelog as `--dry-run` is unreleased.

danobi (approved):

danobi:
I think @jordalgo 's comment in the issue makes sense. But that can be addressed in a separate PR.

PR #3410, opened 2024-08-21
Dry run: do not print maps at the end

viktormalik:
Do not print contents of maps when running with `--dry-run` as the probes should not hit at all and the maps will not contain any meaningful data and just pollute the output.

##### Checklist


jordalgo:
Love it. Thank you!

jordalgo (approved):
I know I'm pushing it but... runtime test? I won't block.

viktormalik:
> I know I'm pushing it but... runtime test? I won't block.

Good point, added one.

PR #3411, opened 2024-08-21
.gitignore: add clangd files

viktormalik:
I recently started using the clangd LSP and it uses some files and dirs in the project root dir which should not be checked to the repo. I'd appreciate if we could add them to `.gitignore` so that they don't pollute my `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ tests/**/*.bc
 
 result
 ignored/
+
+# clangd files
+compile_commands.json
+.cache/


jordalgo (approved):

ajor:
I don't have a serious objection to this, but IMO the bpftrace `.gitignore` file should be for ignoring files specifically related to the bpftrace project. I'd put these clangd rules (and all the vim, idea, etc. rules we already have) in my global gitignore file so it applies to all projects on my machine.

danobi:
I use LSP as well and I don't have these issues. No objection to putting this in here, but maybe your configuration is off? This works quite well for me: https://github.com/mattn/vim-lsp-settings

viktormalik:
> I'd put these clangd rules (and all the vim, idea, etc. rules we already have) in my global gitignore file so it applies to all projects on my machine.

I knew I'd learn something new when I open this PR :-) I didn't know you could have a global `.gitignore`, thanks for the tip!

Closing the issue.

viktormalik:
> I use LSP as well and I don't have these issues. No objection to putting this in here, but maybe your configuration is off? This works quite well for me: https://github.com/mattn/vim-lsp-settings

Most LSPs I use (in Neovim) do not require any files in the repo but AFAIK clangd specifically needs `compile_commands.json` to work properly. But truth is that it can be placed in the `build/` directory in which case our current `.gitignore` will already ignore it.

PR #3416, opened 2024-08-22
Fix variable type mismatches for scratch vars

jordalgo:
Right now the script below prints 0, when it
should print 2.
```
BEGIN { $a = (int16)0; $a = 2; print(($a)); exit(); }
```

This is because the type of '2' is an int64.
This change will cast the literal to the smaller
type if possible, otherwise it causes an error.
This also adds additional integer error checking
to prevent incorrect results.

Fixes: https://github.com/bpftrace/bpftrace/issues/3415

##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -106,9 +106,10 @@ class Integer : public Expression {
 public:
   DEFINE_ACCEPT
 
-  explicit Integer(int64_t n, location loc);
+  explicit Integer(int64_t n, location loc, bool is_negative = true);
 
   int64_t n;
+  bool is_negative;
 
 private:
   Integer(const Integer &other) = default;
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -37,9 +37,61 @@ static const std::map<std::string, std::tuple<size_t, bool>> &getIntcasts()
   return intcasts;
 }
 
+static std::pair<uint64_t, uint64_t> getUIntTypeRange(const SizedType &ty)
+{
+  assert(ty.IsIntegerTy());
+  auto size = ty.GetSize();
+  switch (size) {
+    case 1:
+      return { 0, std::numeric_limits<uint8_t>::max() };
+    case 2:
+      return { 0, std::numeric_limits<uint16_t>::max() };
+    case 4:
+      return { 0, std::numeric_limits<uint32_t>::max() };
+    case 8:
+      return { 0, std::numeric_limits<uint64_t>::max() };
+    default:
+      LOG(BUG) << "Unrecognized int type size: " << size;
+      return { 0, 0 };
+  }
+}
+
+static std::pair<int64_t, int64_t> getIntTypeRange(const SizedType &ty)
+{
+  assert(ty.IsIntegerTy());
+  auto size = ty.GetSize();
+  switch (size) {
+    case 1:
+      return { std::numeric_limits<int8_t>::min(),
+               std::numeric_limits<int8_t>::max() };
+    case 2:
+      return { std::numeric_limits<int16_t>::min(),
+               std::numeric_limits<int16_t>::max() };
+    case 4:
+      return { std::numeric_limits<int32_t>::min(),
+               std::numeric_limits<int32_t>::max() };
+    case 8:
+      return { std::numeric_limits<int64_t>::min(),
+               std::numeric_limits<int64_t>::max() };
+    default:
+      LOG(BUG) << "Unrecognized int type size: " << size;
+      return { 0, 0 };
+  }
+}
+
 void SemanticAnalyser::visit(Integer &integer)
 {
-  integer.type = CreateInt64();
+  if (integer.is_negative) {
+    integer.type = CreateInt64();
+  } else {
+    // Default to signed unless the original value is too large
+    uint64_t val = static_cast<uint64_t>(integer.n);
+    if (val > std::numeric_limits<int32_t>::max()) {
+      integer.type = CreateUInt64();
+    } else {
+      integer.type = CreateInt64();
+    }
+  }
 }
 
 void SemanticAnalyser::visit(PositionalParameter &param)
@@ -2639,6 +2691,69 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
           << "trying to assign value of type '" << assignTy
           << "' when variable already contains a value of type '"
           << search->second << "'";
+    } else if (search->second.IsIntegerTy()) {
+      if (assignment.expr->is_literal) {
+        auto integer = static_cast<Integer *>(assignment.expr);
+        if (!search->second.IsEqual(assignTy)) {
+          int64_t value = integer->n;
+          bool can_fit = false;
+          if (integer->is_negative) {
+            if (!search->second.IsSigned()) {
+              LOG(ERROR, assignment.loc, err_)
+                  << "Type mismatch for " << var_ident << ": "
+                  << "trying to assign value of type '" << assignTy
+                  << "' when variable already contains a value of type '"
+                  << search->second << "'";
+              return;
+            } else {
+              auto min_max = getIntTypeRange(search->second);
+              can_fit = value >= min_max.first;
+            }
+          } else {
+            if (!search->second.IsSigned()) {
+              auto min_max = getUIntTypeRange(search->second);
+              can_fit = static_cast<uint64_t>(value) <= min_max.second;
+            } else {
+              // Casting to a uint64 here because the assign 'value'
+              // might be larger than the max signed int64 e.g.
+              // `$x = -1; $x = 10223372036854775807;`
+              auto min_max = getIntTypeRange(search->second);
+              can_fit = static_cast<uint64_t>(value) <=
+                        static_cast<uint64_t>(min_max.second);
+            }
+          }
+          if (can_fit) {
+            Expression *cast = ctx_.make_node<Cast>(
+                CreateInteger(search->second.GetSize() * 8,
+                              search->second.IsSigned()),
+                assignment.expr,
+                assignment.loc);
+            Visit(cast);
+            assignment.expr = cast;
+          } else {
+            LOG(ERROR, assignment.loc, err_)
+                << "Type mismatch for " << var_ident << ": "
+                << "trying to assign value '"
+                << (integer->is_negative ? integer->n
+                                         : static_cast<uint64_t>(integer->n))
+                << "' which does not fit into the variable of type '"
+                << search->second << "'";
+          }
+        }
+      } else if (search->second.IsSigned() != assignTy.IsSigned()) {
+        LOG(ERROR, assignment.loc, err_)
+            << "Type mismatch for " << var_ident << ": "
+            << "trying to assign value of type '" << assignTy
+            << "' when variable already contains a value of type '"
+            << search->second << "'";
+      } else {
+        if (!assignTy.FitsInto(search->second)) {
+          LOG(ERROR, assignment.loc, err_)
+              << "Integer size mismatch. Assignment type '" << assignTy
+              << "' is larger than the variable type '" << search->second
+              << "'.";
+        }
+      }
     }
   } else {
     // This variable hasn't been seen before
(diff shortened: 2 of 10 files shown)


src/types.cpp
```diff
@@ -567,6 +567,9 @@ bool SizedType::FitsInto(const SizedType &t) const
   if (IsStringTy() && t.IsStringTy())
     return GetSize() <= t.GetSize();
 
+  if (IsIntegerTy() && t.IsIntegerTy())
+    return GetSize() <= t.GetSize();
```
ajor, line 571: Should we also check that the signed-ness matches?

e.g. I feel like this shouldn't be allowed:
```
$a = (uint32)1; $a = (int32)-1
```

ajor (commented):
Nice - I feel like integer sizes will finally make sense to me once this is merged!

tests/codegen/intcast_assign_var.cpp
```diff
@@ -10,6 +10,12 @@ TEST(codegen, intcast_retval)
   test("kretprobe:f { @=(int32)retval }", NAME);
 }
 
+TEST(codegen, intcast_implicit_truncation)
+{
+  // Make sure the result is truncated to 8 bit and sign extended
+  test("kretprobe:f { $a = (uint8)1; $a = 200; print(($a)); }", NAME);
```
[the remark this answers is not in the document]
jordalgo: Ah yeah, I think that's what I was originally looking for. Will update, thanks!

src/ast/passes/semantic_analyser.cpp
```diff
@@ -37,6 +37,38 @@ static const std::map<std::string, std::tuple<size_t, bool>> &getIntcasts()
...
+    if (ty.IsSigned()) {
+      return { INT_MIN, INT_MAX };
+    } else {
+      return { 0, UINT_MAX };
+    }
+  } else if (size == 8) {
+    if (ty.IsSigned()) {
+      return { LONG_MIN, LONG_MAX };
+    } else {
+      return { 0, ULONG_MAX };
```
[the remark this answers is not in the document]
jordalgo: Ugh, yes. Good catch.

src/types.cpp
jordalgo: I'm checking the signed-ness before this gets called.

jordalgo:
@ajor I realized I could use the fact that we know if an integer literal is negative to improve automatic downcasting e.g. this is no longer an issue `BEGIN { $x = (uint8)1; $x = 200; }` 200 gets casted to a uint8 even though it starts as a int64.

src/types.cpp
ajor: I think it needs to be checked in this function, otherwise stuff like this is still possible:
```
$x = (0, (uint32)123); $x = (0, (int32)-123);
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -37,9 +37,53 @@ static const std::map<std::string, std::tuple<size_t, bool>> &getIntcasts()
...
+
 void SemanticAnalyser::visit(Integer &integer)
 {
-  integer.type = CreateInt64();
+  if (integer.is_negative) {
+    integer.type = CreateInt64();
+  } else {
+    // Default to signed unless the original value is too large
+    uint64_t val = static_cast<uint64_t>(integer.n);
+    if (val > LONG_MAX) {
```
ajor, line 81: To get consistent behaviour across platforms:
```suggestion
    if (val > std::numeric_limits<int64_t>::max()) {
```

`LONG_MAX` depends on `long`, which can be either 32-bit or 64-bit.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -37,9 +37,53 @@ static const std::map<std::string, std::tuple<size_t, bool>> &getIntcasts()
   return intcasts;
 }
 
+template <typename T>
+static std::pair<T, T> getIntTypeRange(const SizedType &ty)
```
ajor, line 41: Right now we're checking `IsSigned()` before the call to `getIntTypeRange` to determine which template parameter to use, and then checking it again inside this function to find the actual range. Also, the function `getIntTypeRange<int64_t>` sounds like it might return the range specifically for 64-bit integers (and not 8, 16, 32 bit ints).

How about just having two functions that don't have to check `IsSigned` inside them:
```
std::pair<int64_t, int64_t> getIntTypeRange(...)
std::pair<uint64_t, uint64_t> getUIntTypeRange(...)
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -37,9 +37,53 @@ static const std::map<std::string, std::tuple<size_t, bool>> &getIntcasts()
...
+    if (ty.IsSigned()) {
+      return { INT_MIN, INT_MAX };
+    } else {
+      return { 0, UINT_MAX };
+    }
+  } else if (size == 8) {
+    if (ty.IsSigned()) {
+      return { LONG_MIN, LONG_MAX };
+    } else {
+      return { 0, ULONG_MAX };
```
ajor, line 67: We'd also be better off using `std::numeric_limits<T>::max` in place of the platform-dependent C macros here

src/ast/passes/semantic_analyser.cpp
jordalgo: I like it. Good idea 💡

src/types.cpp
jordalgo: Ah good point. Will also add a test for this case.

src/ast/ast.h
```diff
@@ -106,9 +106,10 @@ class Integer : public Expression {
 public:
   DEFINE_ACCEPT
 
-  explicit Integer(int64_t n, location loc);
+  explicit Integer(int64_t n, location loc, bool is_negative = true);
```
danobi, line 109: naming nit: is_signed instead of is_negative? naively I'd think you could just check n < 0 for is_negative.

tests/runtime/basic
```diff
@@ -300,3 +300,8 @@ NAME print_per_cpu_map_vals
 PROG BEGIN { @a = avg(5); @c = count(); @s = sum(5); @mn = min(1); @mx = max(-1); print((@a, @c, @s, @mn, @mx)); exit(); }
 EXPECT (5, 1, 5, 1, -1)
 TIMEOUT 3
+
+NAME print large int
+PROG BEGIN { $a = 10223372036854775807; print(($a)); }
```
danobi, line 305: adding an exit() at the end will make this run faster

danobi (commented):
looks quite reasonable.

looks like maps represent all integers as i64. is that why maps don't have this issue?

do we have any desire to unify the behavior?

src/ast/ast.h
jordalgo: > naively I'd think you could just check n < 0 for is_negative

The 'is_negative', which matches the parser name, is the key to understanding if we are dealing with an actual negative number or a really large unsigned integer e.g. `10223372036854775807` already comes in as casted to an `int64_t` and would show up as negative.

jordalgo:
> looks like maps represent all integers as i64. is that why maps don't have this issue? do we have any desire to unify the behavior?

Good call. We should absolutely unify this behavior. Let me look into doing that in a follow-up PR.

danobi:
> > looks like maps represent all integers as i64. is that why maps don't have this issue? do we have any desire to unify the behavior?
> 
> Good call. We should absolutely unify this behavior. Let me look into doing that in a follow-up PR.

unless we want to use the map approach here as well, right? cuz making all ints 64 bits should simplify code here. But whether that is the right thing to do, I don't know

src/ast/ast.h
jordalgo: I guess I don't find it that bad because we're still at the AST level which interacts with the parser which has specific logic for MINUS. We could overload Integer to accept a uint64 in the case there is no MINUS but we still need to pass this information some way to the semantic_analyser which ultimately is deciding the type.

jordalgo:
I think we need to fix maps. See the below examples, which should all cause type errors IMHO.
```
'BEGIN { @a = (uint16)1; @a = -2; exit(); }'

@a: 18446744073709551614
```

```
BEGIN { @a = 1; @a = 10223372036854775807; exit(); }

@a: -8223372036854775809
```

src/types.cpp
```diff
@@ -567,6 +567,9 @@ bool SizedType::FitsInto(const SizedType &t) const
   if (IsStringTy() && t.IsStringTy())
     return GetSize() <= t.GetSize();
 
+  if (IsIntegerTy() && t.IsIntegerTy() && (IsSigned() == t.IsSigned()))
+    return GetSize() <= t.GetSize();
```
viktormalik, line 571: Nit: the comment for this method is outdated (it already was before this PR), should be adjusted or removed (I think that the code is quite self-documenting in this case).

src/ast/passes/semantic_analyser.cpp
```diff
@@ -37,9 +37,57 @@ static const std::map<std::string, std::tuple<size_t, bool>> &getIntcasts()
   return intcasts;
 }
 
+static std::pair<uint64_t, uint64_t> getUIntTypeRange(const SizedType &ty)
+{
+  assert(ty.IsIntegerTy());
+  auto size = ty.GetSize();
+  if (size == 1) {
```
viktormalik, line 44: Nit: it feels like a `switch` statement would make the code more readable here but that may be just my preference. Same in the other function below.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2639,6 +2687,66 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
...
+              can_fit = value >= min_max.first;
+            }
+          } else {
+            if (!search->second.IsSigned()) {
+              auto min_max = getUIntTypeRange(search->second);
+              can_fit = static_cast<uint64_t>(value) <= min_max.second;
+            } else {
+              auto min_max = getIntTypeRange(search->second);
+              can_fit = static_cast<uint64_t>(value) <=
+                        static_cast<uint64_t>(min_max.second);
```
viktormalik, line 2715: Why not like this?
```suggestion
              can_fit = value >= min_max.first && value <= min_max.second;
```

tools/runqlen.bt
```diff
@@ -33,7 +33,7 @@ profile:hz:99
 {
 	$task = (struct task_struct *)curtask;
 	$my_q = (struct cfs_rq *)$task->se.cfs_rq;
-	$len = $my_q->nr_running;
+	$len = (int64)$my_q->nr_running;
```
viktormalik, line 36: What would happen if the typecast was not added here? An error or an incorrect value assigned to `$len`?

viktormalik (commented):
Just a couple of nits, otherwise makes sense and looks good.

src/ast/passes/semantic_analyser.cpp
jordalgo: We already know the value is non negative at this point and the lowest uint64 is 0.

viktormalik: I must be missing something but I still don't understand why cast the values to `uint64_t` when this branch deals with signed ints.

tools/runqlen.bt
viktormalik: Ok, makes sense. But maybe this would be a candidate for the new "Breaking" changelog section as it may cause previously working scripts fail.

jordalgo: My hope is there aren't many of these cases but I can add an entry if you prefer.

viktormalik: Yeah, I'm not sure either. Let's leave it without an entry for now and discuss on the office hours how we want to treat breaking changes in general.

src/ast/passes/semantic_analyser.cpp
jordalgo: Sure, will add a comment.

jordalgo:
I think I addressed all comments. Going to merge soon unless anyone has any objections.

viktormalik (approved):
Nice!

jordalgo:
As a bit of added info, I tested the 63 internal Meta bpftrace scripts and 2 had a integer type error as a result of this.

PR #3417, opened 2024-08-23
nix: appimage: Get LLDB to stop running python

danobi:
This PR overlays https://github.com/llvm/llvm-project/pull/105757 into our nix build.

This means that the appimages we ship will stop running python. Which lets us
delete all the python files in the appimage and makes the final result slightly smaller
(216MB -> 214MB). But more importantly, it'll make bpftrace faster since we start
doing python stuff. Less of an attack vector as well.

This does have the downside of requiring an LLDB rebuild. It's not that big of a deal
in CI, as the result will be cached. But for local users this can be annoying. I'm open
to feedback on whether this is worthwhile or not. It _is_ useful, but just unsure how
useful relative to the additional compute for new users.

##### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -71,9 +71,9 @@
 
           pkgs = import nixpkgs { inherit system; };
 
-          # Define lambda that returns a derivation for bpftrace given llvm package as input
+          # Define lambda that returns a derivation for bpftrace given llvm version as input
           mkBpftrace =
-            llvmPackages:
+            llvmVersion:
               with pkgs;
               pkgs.stdenv.mkDerivation rec {
                 name = "bpftrace";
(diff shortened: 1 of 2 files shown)


viktormalik:
No hard opinions on this, just a couple of thoughts below.

The change LGTM although I can't really decide if I'm for including this or not. It's not just the longer Nix build time but also another complexity in the flake file which we need to maintain. On top of that, local Nix builds are primarily aimed at new users who want to experiment with the latest bpftrace so I'm not sure if the benefits are really worth it here.

I'd probably say that it'd be better to work with LLDB upstream to solve this on their side but that may be a lot of work.

danobi:
Yeah, that makes sense. Not running python is only kinda a small benefit anyways.

Let's just merge the first two commits then? I think those are still a good cleanup.

viktormalik:
> Let's just merge the first two commits then? I think those are still a good cleanup.

Sure, those two are nice, let's take them. Maybe just update the commit messages as they mention LLDB.

danobi:
Moved to https://github.com/bpftrace/bpftrace/pull/3425

Issue #2979, opened 2024-02-01
Move type propagation out of semantic analyser


viktormalik:
Currently, semantic analyser does 10 passes to make sure that map types for maps defined after their first usage are correctly propagated. This has several problems:
- semantic analyser needs to do a lot of checks in the final pass, requiring the `is_final_pass()` call,
- for most scripts, 10 passes are not necessary,
- it can shadow some bugs,
- it can still be broken.

To clean this up, let's split out type propagation from semantic analyser into a separate pass. The new pass can loop until all types are resolved. Then, semantic analyser can avoid `is_final_pass()` calls as it will always run after the final type resolution pass.

ajor:
Once this is done, we can clean up the `dereference_if_needed` stuff in #3094. It would be better as a separate pass that sits between type propagation and the true semantic analysis.

ajor:
The new, separate type propagation pass should also try to fix FieldAnalyser, if possible. See #3422. FieldAnalyser currently runs its own incomplete type propagation logic that results in bugs like this:
```
# bpftrace -e 'BEGIN { @y = @x->dfl_cgrp; @x = curtask->cgroups; }'
stdin:1:14-26: ERROR: Struct/union of type 'struct css_set' does not contain a field named 'dfl_cgrp'
BEGIN { @y = @x->dfl_cgrp; @x = curtask->cgroups; }
             ~~~~~~~~~~~~
```

PR #3422, opened 2024-08-27
FieldAnalyser: Resolve fields for structs inside arrays

ajor:
Previously this script would have errored out:
```
# bpftrace -e 'BEGIN { @x = curtask->cgroups->dfl_cgrp->ancestors[1]->kn->name; }'
stdin:1:14-64: ERROR: Struct/union of type 'struct kernfs_node' does not contain a field named 'name'
BEGIN { @x = curtask->cgroups->dfl_cgrp->ancestors[1]->kn->name; }
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The immediate bug is fixed by this PR, but the whole FieldAnalyser pass needs reworking really. A slight modification of the script causes it to break again:
```
# bpftrace -e 'BEGIN { @y = @x->dfl_cgrp; @x = curtask->cgroups; }'
stdin:1:14-26: ERROR: Struct/union of type 'struct css_set' does not contain a field named 'dfl_cgrp'
BEGIN { @y = @x->dfl_cgrp; @x = curtask->cgroups; }
             ~~~~~~~~~~~~
```
This is because FieldAnalyser attempts to do some form of type propagation (via. it's `sized_type_` member), but it does not perform the full type propagation done by SemanticAnalyser. We shouldn't have two separate implementations of type propagation at all. #2979.

##### Checklist

### DIFF
--- a/src/ast/passes/field_analyser.cpp
+++ b/src/ast/passes/field_analyser.cpp
@@ -90,22 +90,20 @@ void FieldAnalyser::visit(FieldAccess &acc)
       sized_type_ = arg->type;
 
     has_builtin_args_ = false;
-  } else if (!sized_type_.IsNoneTy()) {
-    // If the struct type or the field type has not been resolved, add the type
-    // to the BTF set to let ClangParser resolve it
-    if (bpftrace_.has_btf_data() && sized_type_.IsRecordTy()) {
-      SizedType field_type = CreateNone();
-      if (sized_type_.HasField(acc.field))
-        field_type = sized_type_.GetField(acc.field).type;
+  } else if (sized_type_.IsRecordTy()) {
+    SizedType field_type = CreateNone();
+    if (sized_type_.HasField(acc.field))
+      field_type = sized_type_.GetField(acc.field).type;
 
-      if (!field_type.IsNoneTy())
-        sized_type_ = field_type;
-      else {
-        bpftrace_.btf_set_.insert(sized_type_.GetName());
-        auto field_type_name = bpftrace_.btf_->type_of(sized_type_.GetName(),
-                                                       acc.field);
-        bpftrace_.btf_set_.insert(field_type_name);
-      }
+    if (!field_type.IsNoneTy()) {
+      sized_type_ = field_type;
+    } else if (bpftrace_.has_btf_data()) {
+      // If the struct type or the field type has not been resolved, add the
+      // type to the BTF set to let ClangParser resolve it
+      bpftrace_.btf_set_.insert(sized_type_.GetName());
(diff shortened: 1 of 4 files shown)


src/ast/passes/field_analyser.cpp
```diff
@@ -90,22 +90,16 @@ void FieldAnalyser::visit(FieldAccess &acc)
...
+  } else if (sized_type_.IsRecordTy()) {
+    if (sized_type_.HasField(acc.field)) {
+      sized_type_ = sized_type_.GetField(acc.field).type;
+    } else if (bpftrace_.has_btf_data()) {
+      // If the struct type or the field type has not been resolved, add the
+      // type to the BTF set to let ClangParser resolve it
+      bpftrace_.btf_set_.insert(sized_type_.GetName());
+      auto field_type_name = bpftrace_.btf_->type_of(sized_type_.GetName(),
+                                                     acc.field);
+      bpftrace_.btf_set_.insert(field_type_name);
```
ajor, line 102: This change is to make sure we can still resolve fields from DWARF types when there is no BTF present, as in the new unit test. This logic was previously gated on `bpftrace_.has_btf_data()`.

I think the existing unit tests didn't hit this as they were just not trying to access fields nested inside multiple structs.

jordalgo:
Looks like this broke some things in the tools.

ajor:
> Looks like this broke some things in the tools.

The FieldAnalyser/BTF code's even more broken than I'd realised! It can't handle anonymous types at all, this is what broke the tcpconnect tool:
```
struct in6_addr {
        union {
                __u8 u6_addr8[16];
                __be16 u6_addr16[8];
                __be32 u6_addr32[4];
        } in6_u;
};
```

`in6_u` is parsed as `Type::none`:
https://github.com/bpftrace/bpftrace/blob/85e5d653f3bc46ae2d7cff6db26f81138a1dd811/src/btf.cpp#L385-L386

It works around this by getting libbpf to dump the BTF data in C-format and parsing that with ClangParser:
https://github.com/bpftrace/bpftrace/blob/85e5d653f3bc46ae2d7cff6db26f81138a1dd811/src/ast/passes/field_analyser.cpp#L104-L107
https://github.com/bpftrace/bpftrace/blob/85e5d653f3bc46ae2d7cff6db26f81138a1dd811/src/clang_parser.cpp#L722-L730

Massively overcomplicated, but I'm not going to fix it now because I think FieldAnalyser needs a complete rewrite anyway.

jordalgo (approved):

PR #3424, opened 2024-08-28
Printing improvements to help with debugging

ajor:
```
Print unsigned integers as "uintX" instead of "unsigned intX"

This matches the BpfScript syntax and cleans up error messages and AST
debug output.
```
```
Printer: Print types of all expression nodes

This improves the debugging experience and will hopefully help with unit
tests in the future.

Plus another minor change to print post-op Unop nodes in AST order
(like all other nodes) and distinguish them with the string "(post)"
```
```
PassManager: When AST debug is requested, print it for failed passes too

The AST might be in an inconsistent state when it is dumped from a
failed pass, but it is this state at the point of failure that is
often the most useful for understanding issues.
```

##### Checklist

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -354,9 +354,13 @@ std::string opstr(const Unop &unop)
     case Operator::MUL:
       return "dereference";
     case Operator::INCREMENT:
-      return "++";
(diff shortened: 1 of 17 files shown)


jordalgo (approved):
LGTM.

PR #3426, opened 2024-08-29
Move stack allocations to bpftrace stack

danobi:
##### Checklist


danobi:
this approach will have issues when we land calling user defined functions. b/c while all probes can share same percpu stack (as only 1 probe can run on a cpu), subprog calls technically require a frame push. and list of frames need to be stored somewhere. there are probably other hidden complexities. I think surmountable, but we haven't fully exploited (or hit the limits of) percpu scratch maps yet. so I'm going to fully exploit that approach before re-considering this one.

also Yonghong is working on private stacks at kernel level. so stacks will be arbitrarily sized soon. and we can disable percpu scratch maps rather easily if we see private stacks are available.

PR #3428, opened 2024-08-29
Make bpftrace work correctly inside containers with PID namespacing

ajor:
This has been a long time coming - I presented it at eBPF Summit 2023 - but it's finally ready!

This fixes `pid`, `tid` and `ustack` when running bpftrace inside a PID namespace. Previously, these builtins would return PIDs as-seen from the init (global/root) namespace, but this meant that they did not correspond to the correct processes from the bpftrace runtime's point-of-view, as it would see namespaced PIDs.

The core of the PR is covered by this commit:
```
Codegen: Get correct PIDs/TIDs relative to bpftrace's PID namespace

The bpf_get_current_pid_tgid helper only returns PIDs in the root namespace.

The bpf_get_ns_current_pid_tgid helper only returns PIDs in a given
namespace for processes running in that same namespace.

We must switch between the two helpers based on where bpftrace is
running in order to cover the most scenarios.

The following tables show whether we are able to get the correct PID for
a target process using each helper, when bpftrace and the target are and
are-not in PID namespaces.

Always using bpf_get_current_pid_tgid (old behaviour):
        \ target |                |
bpftrace \       | root namespace | child namespace
-----------------+----------------+-----------------
root namespace   | YES            | YES
child namespace  | N/A            | NO

Always using bpf_get_ns_current_pid_tgid:
        \ target |                |
bpftrace \       | root namespace | child namespace
-----------------+----------------+-----------------
root namespace   | YES            | NO
child namespace  | N/A            | YES

Switching between the two helpers (new behaviour):
        \ target |                |
bpftrace \       | root namespace | child namespace
-----------------+----------------+-----------------
root namespace   | YES            | YES
child namespace  | N/A            | YES

This approach is still not complete, however. If the target is running
in a namespace, and bpftrace is running in a different namespace which
still has visibility over the target's namespace, we will not get the
correct PID for the target. This would require kernel changes to work.
```

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -303,7 +303,7 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
         b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
     // store pid
     b_.CreateStore(
-        b_.CreateGetPid(loc),
+        b_.CreateGetPid(ctx_, loc),
         b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(2) }));
     // store probe id
     b_.CreateStore(
@@ -346,9 +346,9 @@ void CodegenLLVM::visit(Builtin &builtin)
   } else if (builtin.ident == "kstack" || builtin.ident == "ustack") {
     kstack_ustack(builtin.ident, builtin.type.stack_type, builtin.loc);
   } else if (builtin.ident == "pid") {
-    expr_ = b_.CreateGetPid(builtin.loc);
+    expr_ = b_.CreateGetPid(ctx_, builtin.loc);
   } else if (builtin.ident == "tid") {
-    expr_ = b_.CreateGetTid(builtin.loc);
+    expr_ = b_.CreateGetTid(ctx_, builtin.loc);
   } else if (builtin.ident == "cgroup") {
     expr_ = b_.CreateGetCurrentCgroupId(builtin.loc);
   } else if (builtin.ident == "uid" || builtin.ident == "gid" ||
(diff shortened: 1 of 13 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -305,7 +305,7 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
         b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
     // store pid
     b_.CreateStore(
-        b_.CreateTrunc(b_.CreateGetPidTgid(loc), b_.getInt32Ty()),
+        b_.CreateGetPid(ctx_, loc),
```
ajor, line 308: The old code here used to extract `tid`, not `pid`, despite the comment. I don't understand why it'd need to be `tid` instead of `pid,` so I think it was just a mistake right from the beginning: 6f5b3fa2798b79acaebc7b3a77b6a6723e7ba48a

tests/runtime/engine/parser.py
```diff
@@ -136,7 +136,7 @@ def __read_test_struct(test, test_suite):
                 if prev_item_name == 'PROG':
                     prog += '\n' + line
                     continue
-                elif prev_item_name == 'EXPECT':
+                elif prev_item_name in ('EXPECT', 'EXPECT_REGEX'):
```
ajor, line 139: To allow multi-line `EXPECT_REGEX` clauses

ajor:
Bump - review please :)

jordalgo (approved):
Sorry for the long delay in reviewing 🕐 Looks good to me. The merge conflicts just need fixing. Did you want to target this for 0.22.x?

ajor:
> Sorry for the long delay in reviewing 🕐 Looks good to me. The merge conflicts just need fixing. Did you want to target this for 0.22.x?

No, let's leave it out of 0.22 so it gets some testing time before a release

MikeWalrus:
Has the manual been changed to reflect this?

Also, sometimes the global pids and tids are useful. e.g., for using tracepoints like sched:sched_switch. Can we add more builtins for getting them?

ajor:
Sorry, thanks for working to fix this!

peckto:
Thanks a lot for solving this issue!
Does this fix also include a method to get the correct parent PID?

PR #3429, opened 2024-08-30
irbuilder: NFC: Add asserts for big allocas

danobi:
First we remove the unused arraysize parameter for allocas. It was a Value so we can't statically know the real value. Since it was unused, delete some of the overloads.

With arraysize deleted, we can now statically assert on alloca size. The new asserts help ensure that we don't try to allocate something too big on the stack. All such allocations need to go onto a scratch map.

##### Checklist


viktormalik (approved):
Makes sense to me.

ajor (approved):

PR #3430, opened 2024-08-30
Check if base path is a directory before iterating

jordalgo:
`std_filesystem::recursive_directory_iterator` will throw if the passed path is not a directory.

##### Checklist

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -636,7 +636,9 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_hierarchy_roots()
   for (std::string line; std::getline(mounts_file, line);) {
     std::smatch match;
     if (std::regex_match(line, match, cgroup_mount_regex)) {
-      result.push_back({ match[1].str(), match[2].str() });
+      if (std_filesystem::is_directory(match[2].str())) {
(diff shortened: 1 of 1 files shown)


danobi (commented):
how did this issue manifest?

only get_cgroup_paths() calls this and it already does some filtering. should we just fix the filter there?

jordalgo:
> how did this issue manifest?

As an error surfaced by running a script via Meta's observability daemon, Strobelight. I couldn't repro the issue running the same script on it's own for some reason but confirmed this fixes the issue.

>only get_cgroup_paths() calls this and it already does some filtering. should we just fix the filter there?

I considered that. You mean doing the same is_directory check here? https://github.com/bpftrace/bpftrace/blob/master/src/utils.cpp#L639

danobi:
> > only get_cgroup_paths() calls this and it already does some filtering. should we just fix the filter there?
> 
> I considered that. You mean doing the same is_directory check here? https://github.com/bpftrace/bpftrace/blob/master/src/utils.cpp#L639

yep

jordalgo:
> yep

Done. Moved.

danobi (approved):

PR #3432, opened 2024-08-30
codegen: ternary: Move allocations off stack

danobi:

Part of https://github.com/bpftrace/bpftrace/issues/3431.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1897,9 +1897,6 @@ void CodegenLLVM::visit(Ternary &ternary)
                                                parent);
   BasicBlock *done = BasicBlock::Create(module_->getContext(), "done", parent);
   // ordering of all the following statements is important
-  Value *result = ternary.type.IsNoneTy()
-                      ? nullptr
-                      : b_.CreateAllocaBPF(ternary.type, "result");
   AllocaInst *buf = ternary.type.IsNoneTy()
                         ? nullptr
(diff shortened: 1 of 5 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -1881,16 +1881,17 @@ void CodegenLLVM::visit(Ternary &ternary)
                                                "right",
                                                parent);
   BasicBlock *done = BasicBlock::Create(module_->getContext(), "done", parent);
+
+  Value *buf = nullptr;
+  if (ternary.type.IsStringTy()) {
```
jordalgo, line 1886: Maybe I'm missing something but why can't this logic be nested inside the `ternary.type.IsStringTy()` on line 1921?

danobi: Yeah, I tried that initially but LLVM barfs during IR validation. It's b/c there's a CreateCondBr on line 1896, and if we put this initialization code before the insert points for left_block and right_block, we strand code between basic blocks. I wanted to keep the current code structure so I put the stuff up here

jordalgo: Ah ok. Maybe just move this new code under the "// ordering of all the following statements is important" comment - as it seems to be important for this too

danobi: Yeah good call

jordalgo (approved):

PR #3434, opened 2024-09-01
Fix `call.path*` tests in newer kernel

ZhangYet:
The runtimetests of `call.path` and `call.path_with_optional_size` will timeout on newer kernels.

See kernel 021a160abf62 ("fs: use __fput_sync in close(2)"). `close` stopped using `flip_close` since kernel 6.5-rc1.

Just like kernel d11ae1b16b0a ("selftests/bpf: Fix d_path test"), using `close_range` to replace `close` is reasonable.

##### Checklist

### DIFF
--- a/tests/testprogs/syscall.c
+++ b/tests/testprogs/syscall.c
@@ -1,6 +1,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <linux/version.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


jordalgo (approved):
Makes sense to me. Thanks!

viktormalik (commented):
`close_range` was added in [kernel 5.9](https://lkml.iu.edu/hypermail/linux/kernel/2008.0/02649.html), could we add `MIN_KERNEL` directive to the relevant tests?

ZhangYet:
Hi, @viktormalik  I added a `KERNEL_VERSION(5, 9, 0)` check.

The reason I didn't use `MIN_KERNEL` is that I'm not sure what kernel is using when running ci. The `KERNEL_VERSION(5, 9, 0)` check can ensure the `flip_close` relative tests to be run.

viktormalik (approved):
> The reason I didn't use `MIN_KERNEL` is that I'm not sure what kernel is using when running ci. The `KERNEL_VERSION(5, 9, 0)` check can ensure the `flip_close` relative tests to be run.

Yeah, this is even better, thanks!

PR #3436, opened 2024-09-03
Support bool type in ClangParser and codegen

### LINKED ISSUE #3435: bpftrace isn't handling bool from tracepoint correctly

In my script `cgroup_rstat_tracepoint.bt` I notice that tracepoint decoding `bool` was returning wrong values (e.g. 129).

-  See [this commit](https://github.com/xdp-project/xdp-project/pull/102/commits/46b330e6dd600f16) for reproducing code.

My workaround is switching to use rawtracepoint:
 - https://github.com/xdp-project/xdp-project/pull/102

### `bpftrace --info` output
System
  OS: Linux 6.10.0-rc5-tj-cgroup+ #44 SMP PREEMPT_DYNAMIC Fri Aug 30 16:14:23 CEST 2024
  Arch: x86_64

Build
  version: v0.20.1
  LLVM: 17.0.6
  unsafe probe: no
  bfd: yes
  libdw (DWARF support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: yes
  uprobe_multi: no
  raw_tp_special: yes
  iter: yes

viktormalik:
Until now, `ClangParser` represented bools as 1-byte ints. This makes us always read 8 bits instead of 1 which may return incorrect values when those remaining 7 bits contain some trash. This is especially an issue for bool-typed tracepoint args as tracepoint definitions are parsed by `ClangParser` (other types are typically parsed from BTF these days).

Fix the issue by properly representing bools with 1-bit type.

This also required a small update in codegen to emit LLVM's `Int1Ty` for our bool type.

Fixes #3435.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -299,6 +299,8 @@ llvm::Type *IRBuilderBPF::GetType(const SizedType &stype)
     ty = getInt64Ty();
   } else if (stype.IsVoidTy()) {
     ty = getVoidTy();
+  } else if (stype.IsBoolTy()) {
+    ty = getInt1Ty();
   } else {
     switch (stype.GetSize()) {
       case 16:
--- a/tests/clang_parser.cpp
+++ b/tests/clang_parser.cpp
@@ -30,16 +30,18 @@ static void parse(const std::string &input,
 TEST(clang_parser, integers)
 {
   BPFtrace bpftrace;
-  parse("struct Foo { int x; int y, z; }", bpftrace);
+  parse("#include <stdbool.h>\nstruct Foo { int x; int y, z; bool b; }",
+        bpftrace);
 
   ASSERT_TRUE(bpftrace.structs.Has("struct Foo"));
   auto foo = bpftrace.structs.Lookup("struct Foo").lock();
 
-  EXPECT_EQ(foo->size, 12);
-  ASSERT_EQ(foo->fields.size(), 3U);
+  EXPECT_EQ(foo->size, 16);
+  ASSERT_EQ(foo->fields.size(), 4U);
   ASSERT_EQ(foo->HasField("x"), true);
   ASSERT_EQ(foo->HasField("y"), true);
   ASSERT_EQ(foo->HasField("z"), true);
+  ASSERT_EQ(foo->HasField("b"), true);
 
   EXPECT_TRUE(foo->GetField("x").type.IsIntTy());
   EXPECT_EQ(foo->GetField("x").type.GetSize(), 4U);
@@ -52,6 +54,10 @@ TEST(clang_parser, integers)
   EXPECT_TRUE(foo->GetField("z").type.IsIntTy());
   EXPECT_EQ(foo->GetField("z").type.GetSize(), 4U);
   EXPECT_EQ(foo->GetField("z").offset, 8);
+
+  EXPECT_TRUE(foo->GetField("b").type.IsBoolTy());
+  EXPECT_EQ(foo->GetField("b").type.GetSize(), 1U);
+  EXPECT_EQ(foo->GetField("b").offset, 12);
 }
 
 TEST(clang_parser, c_union)
@@ -553,11 +559,12 @@ TEST_F(clang_parser_btf, btf)
   auto foo3 = bpftrace.structs.Lookup("struct Foo3").lock();
   auto task_struct = bpftrace.structs.Lookup("struct task_struct").lock();
 
-  EXPECT_EQ(foo1->size, 16);
-  ASSERT_EQ(foo1->fields.size(), 3U);
+  EXPECT_EQ(foo1->size, 24);
+  ASSERT_EQ(foo1->fields.size(), 4U);
   ASSERT_TRUE(foo1->HasField("a"));
   ASSERT_TRUE(foo1->HasField("b"));
   ASSERT_TRUE(foo1->HasField("c"));
+  ASSERT_TRUE(foo1->HasField("d"));
 
   EXPECT_TRUE(foo1->GetField("a").type.IsIntTy());
   EXPECT_EQ(foo1->GetField("a").type.GetSize(), 4U);
@@ -571,7 +578,11 @@ TEST_F(clang_parser_btf, btf)
   EXPECT_EQ(foo1->GetField("c").type.GetSize(), 8U);
   EXPECT_EQ(foo1->GetField("c").offset, 8);
 
-  EXPECT_EQ(foo2->size, 24);
+  EXPECT_TRUE(foo1->GetField("d").type.IsBoolTy());
+  EXPECT_EQ(foo1->GetField("d").type.GetSize(), 1U);
+  EXPECT_EQ(foo1->GetField("d").offset, 16);
+
+  EXPECT_EQ(foo2->size, 32);
   ASSERT_EQ(foo2->fields.size(), 3U);
   ASSERT_TRUE(foo2->HasField("a"));
   ASSERT_TRUE(foo2->HasField("f"));
@@ -582,7 +593,7 @@ TEST_F(clang_parser_btf, btf)
   EXPECT_EQ(foo2->GetField("a").offset, 0);
 
   EXPECT_TRUE(foo2->GetField("f").type.IsRecordTy());
-  EXPECT_EQ(foo2->GetField("f").type.GetSize(), 16U);
+  EXPECT_EQ(foo2->GetField("f").type.GetSize(), 24U);
   EXPECT_EQ(foo2->GetField("f").offset, 8);
 
   EXPECT_TRUE(foo2->GetField("g").type.IsIntTy());
--- a/tests/runtime/other
+++ b/tests/runtime/other
@@ -279,3 +279,13 @@ NAME dry run empty output
 RUN {{BPFTRACE}} --dry-run -e 'BEGIN { printf("hello\n"); @ = 0; }'
 EXPECT_NONE hello
 EXPECT_NONE @: 0
+
+NAME bool type proberead
+PROG struct Foo { int a; bool b; } uprobe:./testprogs/uprobe_test:uprobeFunction2 { printf("value: %d\n", ((struct Foo *)arg1)->b); exit() }
+AFTER ./testprogs/uprobe_test
+EXPECT value: 1
+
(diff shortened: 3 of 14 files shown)


jordalgo:
LGTM. Do we want a codegen test?

tests/clang_parser.cpp
```diff
@@ -30,16 +30,17 @@ static void parse(const std::string &input,
 TEST(clang_parser, integers)
 {
   BPFtrace bpftrace;
-  parse("struct Foo { int x; int y, z; }", bpftrace);
+  parse("struct Foo { int x; int y, z; _Bool b; }", bpftrace);
```
danobi, line 33: does regular `bool` work? would be good to test that. not sure if libbpf generates `_Bool` or `bool`

viktormalik:
> LGTM. Do we want a codegen test?

So I started adding one and it turned out that our bool handling was broken on many fronts :smile: I tried to fix it everywhere (codegen, `ClangParser`, `DwarfParser`, `BTF`) and added various tests. Check the updated commit message for details.

tests/clang_parser.cpp
viktormalik: Interestingly it doesn't here in the tests unless I explicitly include `stdbool.h` (which I did). When running bpftrace from CLI, using `bool` works without the include (as `ClangParser` includes `stdbool.h` implicitly). I don't understand why that doesn't work in the tests, though.

netoptimizer:
I compiled bpftrace from this branch
and unfortunately [my reproducer](https://github.com/xdp-project/xdp-project/pull/102/commits/46b330e6dd600f1689c65f01bf9acdabcadb47a8) cannot run and fails with:

```
ERROR: Error loading BPF program for kfunc_vmlinux_cgroup_rstat_flush_locked_5. Use -v for full kernel error log.
```
...so, I cannot test if this fixed the issue :crying_cat_face:

netoptimizer:
I compiled bpftrace from latest git:
```
/usr/local/bin/bpftrace --version
bpftrace v0.21.0-127-g2a3e
```
And I get same error:
```
$ sudo /usr/local/bin/bpftrace -v -e 'kfunc:cgroup_rstat_flush_locked {@cnt[probe]=count()}' 
AST node count: 7
Attaching 1 probe...
ERROR: Error loading BPF program for kfunc_vmlinux_cgroup_rstat_flush_locked_1.
Kernel error log: 
Tracing programs must provide btf_id
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0


```
Maybe this is another unrelated bug?

ajor:
@netoptimizer That error message is very unhelpful, but the most likely cause is that you're compiling with an old(ish) version of libbpf. Latest bpftrace needs an as-yet-unreleased libbpf to work properly: https://github.com/bpftrace/bpftrace/blob/2a3e2497c518bfd5e0cc7bb6437c17e7a71faca7/flake.nix#L36-L40

[Building with Nix](https://github.com/bpftrace/bpftrace/blob/master/docs/nix.md) is probably the easiest solution, but you can equally build and install libbpf from source.

src/ast/irbuilderbpf.cpp
```diff
@@ -299,6 +299,8 @@ llvm::Type *IRBuilderBPF::GetType(const SizedType &stype)
     ty = getInt64Ty();
   } else if (stype.IsVoidTy()) {
     ty = getVoidTy();
+  } else if (stype.IsBoolTy()) {
+    ty = getInt1Ty();
```
ajor, line 303: What's the purpose of this, given that you're overriding it with `getInt8Ty()` above and below in this PR?

It would be nice if we could just use this `getInt1Ty()` and not have the overrides though. I'm surprised that LLVM doesn't make this work. Maybe it's just a problem with the BPF backend that converts `i1` into `i8`? (I'm just guessing here)

viktormalik: The `getInt8Ty()` override is only necessary when doing proberead as it requires to read the entire byte. For all other purposes, this `getInt1Ty()` should do the job.

But I agree that the code was confusing with the `getInt8Ty()` overrides spread all over the place so I moved everything related to the proberead special case to one place.

tests/runtime/other
```diff
@@ -279,3 +279,13 @@ NAME dry run empty output
 RUN {{BPFTRACE}} --dry-run -e 'BEGIN { printf("hello\n"); @ = 0; }'
 EXPECT_NONE hello
 EXPECT_NONE @: 0
+
+NAME bool type proberead
+PROG struct Foo { int a; bool b; } uprobe:./testprogs/uprobe_test:uprobeFunction2 { printf("value: %d\n", ((struct Foo *)arg1)->b); exit() }
```
jordalgo, line 284: dumb question: why does printf cause a proberead?

tests/clang_parser.cpp
```diff
@@ -553,11 +559,12 @@ TEST_F(clang_parser_btf, btf)
   auto foo3 = bpftrace.structs.Lookup("struct Foo3").lock();
   auto task_struct = bpftrace.structs.Lookup("struct task_struct").lock();
 
-  EXPECT_EQ(foo1->size, 16);
-  ASSERT_EQ(foo1->fields.size(), 3U);
+  EXPECT_EQ(foo1->size, 24);
```
jordalgo, line 562: Why did foo1's size increase?

jordalgo (approved):
LGTM

ajor (commented):
Thanks, the new code is clearer but I'm still a bit unsure about whether we should be special-casing probe-reading for bools at all. I guess it comes down to what we want to do when we encounter a boolean value that isn't 0 or 1.

Here's a program I wrote to experiment with this:
```
#include <stdio.h>
#include <stdint.h>
#include <stdbool.h>

struct Foo {
  uint8_t a;
  bool b;
  uint8_t c;
};

void foo(struct Foo *f) {
  printf("%d %d %d\n", f->a, f->b, f->c);
}

int main() {
  struct Foo f = {.a=0xaa, .b = 1, .c = 0xcc};
  *((uint8_t*)&f.b) = 0xb0;
  foo(&f);
  return 0;
}
```

The probe-read for the boolean is special-cased, so the value `0xb0` becomes "false" when `foo.b` is read on it's own, but when the entire struct is read, `foo.b` is interpreted as "true" since it's non-zero:
```
# bpftrace -e 'uprobe:/tmp/test/a.out:foo { @f = *args.f; @b = args.f->b; exit(); }'
Attaching 1 probe...

@b: 0
@f: { .a = 170, .b = 1, .c = 204 
```

Should we follow the C-convention of zero being `false` and non-zero being `true`, or do we just want to say that behaviour is undefined for values other than 0 or 1 (that might be easier)?

viktormalik:
> Thanks, the new code is clearer but I'm still a bit unsure about whether we should be special-casing probe-reading for bools at all. I guess it comes down to what we want to do when we encounter a boolean value that isn't 0 or 1.
> 
> Here's a program I wrote to experiment with this:

This is a good example, thanks for coming with it.

> [...]
> 
> The probe-read for the boolean is special-cased, so the value `0xb0` becomes "false" when `foo.b` is read on it's own, but when the entire struct is read, `foo.b` is interpreted as "true" since it's non-zero:

Correct. To explain in detail: when we do (special-cased) probe read via `args.f->b`, we read `0xb0` and then do `0xb0 & 0x01 = 0x00`. For the other case, the conversion from `i8` to `i1` is done in `Output::value_to_str` which this PR updates for the bool type:
```cpp
      case 1:
        return std::to_string(reduce_value<bool>(value, nvalues) / div);
```
so it basically interprets the value as `bool` and is assigned 1 as it's not zero.

> Should we follow the C-convention of zero being `false` and non-zero being `true`, or do we just want to say that behaviour is undefined for values other than 0 or 1 (that might be easier)?

I would say that yes, we should. The other option would be to do what we do for the special-cased proberead here, i.e. apply the `0x01` mask but in hindsight, that doesn't seem like a correct solution as it doesn't comply with the C standard.

In fact, I'm surprised by the bug reported in #3435 as only the 0/1 value should be stored for the `bool` tracepoint arg anyways. I start to feel like the actual problem is related to other fields not being read correctly either.

Still, I think that it's worth making the implicit conversion to `bool` by comparing the value to 0. I think that LLVM `i8` -> `i1` conversion should do exactly that, though, so I'm gonna try to use it.

viktormalik:
> Still, I think that it's worth making the implicit conversion to `bool` by comparing the value to 0. I think that LLVM `i8` -> `i1` conversion should do exactly that, though, so I'm gonna try to use it.

Unfortunately, this doesn't seem to be the case.

For the above @ajor's example, using a simpler script:
```
$ bpftrace -e 'uprobe:/tmp/test:foo { printf("%d\n", args.f->b); exit(); }'
```
will print `176` (i.e. `0xb0`).

Codegen contains this code:
```llvm
%"struct Foo.b" = alloca i1, align 1
[...]
%probe_read_user = call i64 inttoptr (i64 112 to ptr)(ptr %"struct Foo.b", i32 1, i64 %6)
%7 = load i1, ptr %"struct Foo.b", align 1
```
however the `load i1` doesn't perform implicit conversion to 0/1.

We could add an implicit `icmp eq %X, 0` but there's quite a lot of places in codegen where we'd have to add it. So, I'm starting to be in favor of leaving things as they were (representing `bool` with `Int8Ty`) and saying that we just read whatever is in the memory.

viktormalik:
Closing as this is not the correct way to handle external bools. We should just use `int8` to represent boolean fields (which we do now) and read whatever is stored in the memory. We may want to introduce special handling of internal boolean types but that's tracked in #3502.

PR #3437, opened 2024-09-03
Runtime tests: Set default timeout if none is specified

ajor:
It looks like there was the idea to do this years ago, but it wasn't implemented throughout the test framework so the parser still rejected tests without a TIMEOUT clause.

This is just to make it a little easier to write runtime tests. The vast majority shouldn't need to specify a timeout.


jordalgo:
Nice! What do you think about auditing the tests and removing all the hardcoded timeouts if they are the same as the default?

jordalgo (approved):

ajor:
> Nice! What do you think about auditing the tests and removing all the hardcoded timeouts if they are the same as the default?

Good idea. That shouldn't take too long - I'll do it as a second commit in this PR

jordalgo (approved):

PR #3438, opened 2024-09-03
WIP: Support cgroup_path as a mapkey in output

ZhangYet:
To resolve https://github.com/bpftrace/bpftrace/issues/3421

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ and this project adheres to
   - [#3249](https://github.com/bpftrace/bpftrace/pull/3249)
 - Faster map access for keyless maps by using BPF_MAP_TYPE_ARRAY
   - [#3300](https://github.com/bpftrace/bpftrace/pull/3300)
+- Support `cgroup_path` as a mapkey in output
+  - [#3438](https://github.com/bpftrace/bpftrace/pull/3438)
 #### Deprecated
 #### Removed
 - Remove the `-dd` CLI option
--- a/src/mapkey.cpp
+++ b/src/mapkey.cpp
@@ -148,6 +148,11 @@ std::string MapKey::argument_value(BPFtrace &bpftrace,
       auto p = static_cast<const uint8_t *>(data);
       return bpftrace.resolve_mac_address(p);
     }
+    case Type::cgroup_path: {
+      return bpftrace.resolve_cgroup_path(
+          read_data<uint64_t>(data),
+          read_data<uint64_t>(static_cast<const uint8_t *>(data) + 8));
+    }
     default:
       LOG(BUG) << "invalid mapkey argument type";
   }
--- a/tests/runtime/call
+++ b/tests/runtime/call
@@ -498,6 +498,12 @@ EXPECT_REGEX path: ([a-z]*:\/.*;)*[a-z]*:\/.*
 REQUIRES grep -q '^cgroup2' /proc/mounts
 MIN_KERNEL 4.18
 
+NAME cgroup_path map_key
+PROG BEGIN { @[cgroup_path(cgroup & ((1 << 32) - 1))] = 1; exit(); }
+EXPECT_REGEX @\[.*\]\: 1
+REQUIRES grep -q '^cgroup2' /proc/mounts
+MIN_KERNEL 4.18
+
 NAME strerror
 RUN {{BPFTRACE}} --include "errno.h" -e 'BEGIN { print(strerror(EPERM)); exit(); }'
 EXPECT Operation not permitted


danobi:
cc @jordalgo - this intersects with your multi-key -> tuple refactor

jordalgo:
> this intersects with your multi-key -> tuple refactor

Thanks for the heads up. I plan on landing that in the next day or too.

ZhangYet:
Hi, @danobi @jordalgo 

I found a problem: 

```shell
sudo build/src/bpftrace -e ' BEGIN{ @[cgroup_path(1)] = 1; print(@[cgroup_path(1)]); exit(); }'

Attaching 1 probe...
0 # which should be 1


@[unified:/]: 1
```

ajor:
`cgroup_path` isn't something I've looked at before, but the docs make it sound like we'd have a problem storing it in a map: https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#functions-cgroup_path

> Convert cgroup id to cgroup path. This is done asynchronously in userspace when the cgroup_path value is printed, therefore it can resolve to a different value if the cgroup id gets reassigned. This also means that the returned value can only be used for printing.

I'm not sure how true that really is though - if it consists of a `cgroup_path_id` and a `cgroup_id` that later get resolved in userspace, it'd still be fine to do stuff with it in BPF space, we just wouldn't have access to the string representation. If it could only be used for printing, the semantic analyser should have been made to reject its use in any other situation.

It having an `AsyncEvent` entry is also doesn't make sense: https://github.com/bpftrace/bpftrace/blob/2a3e2497c518bfd5e0cc7bb6437c17e7a71faca7/src/ast/async_event_types.h#L104-L109
It should be purely a function that returns a type, not an event, unless I'm missing something. These oddities plus your problem make me think that there might be more work involved in fixing `cgroup_path` than expected.

ZhangYet:
@ajor 

I have tested this script today:
```shell
bpftrace -e 'tracepoint:cgroup:cgroup_rstat_cpu_locked_fastpath { @[cgroup_path(cgroup)]=count(); }'
```

It did give outputs like:

```
...
@[unified:/system.slice/udisks2.service]: 1
@[unified:/system.slice/system-gpg\x2dagent.slice/gpg-agent@etc-pacman.d-gnupg.service]: 1
@[unified:/user.slice/user-1000.slice/user@1000.service/background.slice/plasma-powerdevil.service]: 2
@[unified:/system.slice/rtkit-daemon.service]: 3
@[unified:/user.slice/user-1000.slice/user@1000.service/app.slice/tmux-spawn-2aed1393-8978-49dc-9f52-8b9513f0ed9b.scope]: 3
...
```
So, I think the problem maybe in `print`.

ZhangYet:
@ajor 

I think this is the bug of `print`, see: 
```shell
bpftrace -e 'BEGIN { @[kstack]=42; print(@[kstack]); exit(); }'
Attaching 1 probe...
0


@[
    bpf_prog_6deef7357e7b4530_sd_fw_ingress+14784
    bpf_prog_6deef7357e7b4530_sd_fw_ingress+14784
    __bpf_prog_test_run_raw_tp+106
    bpf_prog_test_run_raw_tp+301
    __sys_bpf+2566
    __x64_sys_bpf+37
    do_syscall_64+130
    entry_SYSCALL_64_after_hwframe+118
]: 42
```

I found that when in `Output::value_to_str` the parameter `type` was passed as `integer`. 

Maybe we should raise a bug issue for this?

ZhangYet:
Since both `print` and `cgroup_path` are async, I think that `print(@[cgroup_path(cgroup_id)]` may be a complicated problem but not relevant to this PR. Maybe we can move on? @ajor @danobi

ajor:
@ZhangYet Sorry, I haven't looked into what's going wrong, but we can't merge support for cgroup_path map keys if they give wrong-results in some scenarios.

I don't think it's related to them being async though - AFAIK they aren't actually async at all, despite what the code's saying! All the async-ness around cgroup_path should probably be deleted for clarity.

Here's another case where they don't work, avoiding any async-ness:
```
$ sudo ./src/bpftrace -e 'BEGIN{ @[cgroup_path(1)] = 1; @a = @[cgroup_path(1)]; }'
Attaching 1 probe...
^C

@[unified:/]: 1
@a: 0
```

ZhangYet:
The same scenario:

```
sudo bpftrace -e 'BEGIN { @[pid]=42; @a=@[pid]; exit(); }'

Attaching 1 probe...


@[544265]: 42
@a: 0
```

Still working on it.

ZhangYet:
I think I found the direct cause of this scenario, below is asm generated by (the v0.21.2 version will generate same elf)
```shell
bpftrace -e 'BEGIN { @[cgroup_path(13)]=42; @debug_var=@[cgroup_path(13)]; }' --emit-elf cgroup_elf
```
As we can see line 0-3, the prog save a `cgroup_path_t` <13, 0> in `r10-0x30` which will be used in line 7 as the key to call `bpf_map_update_elem`(line 13).

But in line 14-16, the prog save a `cgroup_path_t` <13, 1> in `r0-0x20` which will be used in line 17 as the key to call `bpf_map_lookup_elem`(line 21).

@jordalgo Could you help to locate the root cause? 


```
cgroup_elf:	file format elf64-bpf

Disassembly of section s_BEGIN_1:

0000000000000000 <BEGIN_1>:
       0:	b7 07 00 00 0d 00 00 00	r7 = 0xd
       1:	7b 7a d8 ff 00 00 00 00	*(u64 *)(r10 - 0x28) = r7
       2:	b7 06 00 00 00 00 00 00	r6 = 0x0
       3:	7b 6a d0 ff 00 00 00 00	*(u64 *)(r10 - 0x30) = r6
       4:	b7 01 00 00 2a 00 00 00	r1 = 0x2a
       5:	7b 1a e0 ff 00 00 00 00	*(u64 *)(r10 - 0x20) = r1
       6:	bf a2 00 00 00 00 00 00	r2 = r10
       7:	07 02 00 00 d0 ff ff ff	r2 += -0x30
       8:	bf a3 00 00 00 00 00 00	r3 = r10
       9:	07 03 00 00 e0 ff ff ff	r3 += -0x20
      10:	18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00	r1 = 0x0 ll
      12:	b7 04 00 00 00 00 00 00	r4 = 0x0
      13:	85 00 00 00 02 00 00 00	call 0x2
      14:	7b 7a e8 ff 00 00 00 00	*(u64 *)(r10 - 0x18) = r7
      15:	b7 01 00 00 01 00 00 00	r1 = 0x1
      16:	7b 1a e0 ff 00 00 00 00	*(u64 *)(r10 - 0x20) = r1
      17:	bf a2 00 00 00 00 00 00	r2 = r10
      18:	07 02 00 00 e0 ff ff ff	r2 += -0x20
      19:	18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00	r1 = 0x0 ll
      21:	85 00 00 00 01 00 00 00	call 0x1
      22:	b7 01 00 00 00 00 00 00	r1 = 0x0
      23:	15 00 01 00 00 00 00 00	if r0 == 0x0 goto +0x1 <LBB0_2>
      24:	79 01 00 00 00 00 00 00	r1 = *(u64 *)(r0 + 0x0)

00000000000000c8 <LBB0_2>:
      25:	7b 6a f0 ff 00 00 00 00	*(u64 *)(r10 - 0x10) = r6
      26:	7b 1a f8 ff 00 00 00 00	*(u64 *)(r10 - 0x8) = r1
      27:	bf a2 00 00 00 00 00 00	r2 = r10
      28:	07 02 00 00 f0 ff ff ff	r2 += -0x10
      29:	bf a3 00 00 00 00 00 00	r3 = r10
      30:	07 03 00 00 f8 ff ff ff	r3 += -0x8
      31:	18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00	r1 = 0x0 ll
      33:	b7 04 00 00 00 00 00 00	r4 = 0x0
      34:	85 00 00 00 02 00 00 00	call 0x2
      35:	b7 00 00 00 00 00 00 00	r0 = 0x0
      36:	95 00 00 00 00 00 00 00	exit
```

jordalgo:
@ZhangYet The `cgroup_path_id` is used to track the second argument (the optional string) for `cgroup_path`:
https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/resource_analyser.cpp#L198

So I don't think we can get rid of it . You'll need to find a way to unique it so we're not using the incremented async id counter we use to identify different `cgroup_path` calls.

jordalgo:
```
sudo bpftrace -e 'BEGIN { @[pid]=42; @a=@[pid]; exit(); }'

Attaching 1 probe...


@[544265]: 42
@a: 0
```
This looks like a bug but I can't reproduce it. Maybe file an issue with details if you see it again?

ZhangYet:
> So I don't think we can get rid of it .

After reading the code all the weekend, I agree with you. I'm trying to figure out a way to resolve it.


> This looks like a bug but I can't reproduce it. Maybe file an issue with details if you see it again?

I cannot reproduce it too. Focusing on the `cgroup_path` problem these days, I'll raise a new issue once I find a way to reproduce it.

ZhangYet:
@jordalgo 

I don't think I can resolve this problem. 

If we want to use `cgroup_path(cgrou_id, filter)` as a map key, we need the `cgroup_path` call to generate a consistent map key, but it can't because the `cgroup_path_id` changes.

I have considered that using `cgroup_id` as mapkey only. Then the problem is

1. using the value of `cgroup_id` as mapkey, how can we know the address of the `cgroup_path_struct` so that we can know the `cgroup_paht_id`. 
2. using the address of `cgroup_id` as mapkey, it also changes.

Maybe we add another `cgroup_path_with_filter()` builtin? But I think this cost too much.

jordalgo:
@ZhangYet My apologies, I should have looked at this closer sooner. Since we're doing the path resolution in user space, can we just store the cgroup id as the map key and then iterate over the map to resolve it to the path e.g.
```
BEGIN {
  @mp[cgroup] = 42; exit(); 
}

END { 
  for ($kv : @mp) { 
    print((cgroup_path($kv.0), $kv.1)); 
   } 
}
```

ZhangYet:
@jordalgo This script works well after my commit. I got lost in the idea that using `cgroup_path` as mapkey.

```
sudo build/src/bpftrace test.bt
Attaching 2 probes...
(unified:/user.slice/user-1000.slice/user@1000.service/app.slice/tmux-spawn-ca84bd3e-9208-4c5f-a6b6-a2013be387e4.scope, 42)


@mp[8929]: 42
```

I'll fix the unit test tomorrow.

jordalgo:
What additional changes are needed? My script should work fine on master.

ZhangYet:
Maybe we should go back to the init [issue](https://github.com/bpftrace/bpftrace/issues/3421).

This issue requires to use `cgroup_path` as a mapkey. But after working on it, I think it's very hard(or maybe impossible) to implement.

jordalgo:
> This issue requires to use cgroup_path as a mapkey. But after working on it, I think it's very hard(or maybe impossible) to implement.

I think it might be possible but at this point, I think it's more unnecessary since you can iterate over maps now and can call `cgroup_path` with the cgroup id as a map key. Again, sorry for not noticing this sooner.

ZhangYet:
@jordalgo  Thank you anyway, you helped a lot and I learned a lot when debugging this issue.

> I think it might be possible but at this point, I think it's more unnecessary since you can iterate over maps now and can call cgroup_path with the cgroup id as a map key. Again, sorry for not noticing this sooner.

I agree with that, if you don't mind, I will close this PR.

PR #3441, opened 2024-09-05
Consistently store pid, tid and probe id as int32_t

ajor:
Previously, they were mostly 64-bit but had a few places where they were treated as 32-bit.
    
This saves some memory and may marginally improve performance when pushing usyms onto the ring buffer, etc., but it's mainly just a nice little clean up for consistency. Maybe it even eliminates some undiscovered lurking bugs.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -73,6 +73,20 @@ libbpf::bpf_func_id IRBuilderBPF::selectProbeReadHelper(AddrSpace as, bool str)
   return fn;
 }
 
+Value *IRBuilderBPF::CreateGetPid(const location &loc)
+{
+  Value *pidtgid = CreateGetPidTgid(loc);
+  Value *pid = CreateLShr(pidtgid, 32, "pid");
+  return pid;
+}
+
+Value *IRBuilderBPF::CreateGetTid(const location &loc)
+{
+  Value *pidtgid = CreateGetPidTgid(loc);
+  Value *tid = CreateAnd(pidtgid, 0xffffffff, "tid");
+  return tid;
+}
+
 AllocaInst *IRBuilderBPF::CreateUSym(llvm::Value *val,
                                      int probe_id,
                                      const location &loc)
@@ -85,7 +99,7 @@ AllocaInst *IRBuilderBPF::CreateUSym(llvm::Value *val,
   StructType *usym_t = GetStructType("usym_t", elements, false);
   AllocaInst *buf = CreateAllocaBPF(usym_t, "usym");
 
-  Value *pid = CreateLShr(CreateGetPidTgid(loc), 32);
+  Value *pid = CreateGetPid(loc);
   Value *probe_id_val = Constant::getIntegerValue(getInt64Ty(),
                                                   APInt(64, probe_id));
 
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -76,14 +76,14 @@ libbpf::bpf_func_id IRBuilderBPF::selectProbeReadHelper(AddrSpace as, bool str)
 Value *IRBuilderBPF::CreateGetPid(const location &loc)
 {
   Value *pidtgid = CreateGetPidTgid(loc);
-  Value *pid = CreateLShr(pidtgid, 32, "pid");
+  Value *pid = CreateTrunc(CreateLShr(pidtgid, 32), getInt32Ty(), "pid");
   return pid;
 }
 
 Value *IRBuilderBPF::CreateGetTid(const location &loc)
 {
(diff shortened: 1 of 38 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -305,7 +305,7 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
         b_.CreateGEP(stack_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
     // store pid
     b_.CreateStore(
-        b_.CreateTrunc(b_.CreateGetPidTgid(loc), b_.getInt32Ty()),
+        b_.CreateGetPid(loc),
```
ajor, line 308: The old code here used to extract `tid`, not `pid`, despite the comment. I don't understand why it'd need to be `tid` instead of `pid,` so I think it was just a mistake right from the beginning: 6f5b3fa2798b79acaebc7b3a77b6a6723e7ba48a

src/ast/irbuilderbpf.cpp
```diff
@@ -73,21 +73,35 @@ libbpf::bpf_func_id IRBuilderBPF::selectProbeReadHelper(AddrSpace as, bool str)
...
 AllocaInst *IRBuilderBPF::CreateUSym(llvm::Value *val,
                                      int probe_id,
                                      const location &loc)
 {
   std::vector<llvm::Type *> elements = {
     getInt64Ty(), // addr
-    getInt64Ty(), // pid
-    getInt64Ty(), // probe id
+    getInt32Ty(), // pid
+    getInt32Ty(), // probe id
```
jordalgo, line 97: Drive by fix?

jordalgo (commented):
Changes look good. 

I'm wondering if the changelog entry should be in our new "Breaking" section though. Consider this contrived example:

```
BEGIN { $a = (pid, tid); $a = ((uint64)2, (uint64)3); exit(); }
```

These tuples are compatible at the moment but after this change they will not be and surface a semantic_analyser error. There might be other cases like this that I'm not considering.

danobi (approved):

viktormalik:
> Changes look good.
> 
> I'm wondering if the changelog entry should be in our new "Breaking" section though. Consider this contrived example:
> 
> ```
> BEGIN { $a = (pid, tid); $a = ((uint64)2, (uint64)3); exit(); }
> ```
> 
> These tuples are compatible at the moment but after this change they will not be and surface a semantic_analyser error. There might be other cases like this that I'm not considering.

I wouldn't want to overwhelm the "breaking" section with too many entries so we should probably clarify what kinds of entries should go in there. For instance, this one does break compatibility but it can be also viewed as a "fixed bug" as pid/tid are `int32` in the kernel.

viktormalik (approved):

jordalgo:
> I wouldn't want to overwhelm the "breaking" section with too many entries so we should probably clarify what kinds of entries should go in there. 

Clarification would be good but right now we have 0 entries in the "breaking" section (which doesn't exist yet either) and the only other one that I see going in there soon is the lexical/block scoping PR. I think this PR is a good candidate for the "breaking" section but I don't feel strongly enough to fight beyond this comment 😄

viktormalik:
> > I wouldn't want to overwhelm the "breaking" section with too many entries so we should probably clarify what kinds of entries should go in there.
> 
> Clarification would be good but right now we have 0 entries in the "breaking" section (which doesn't exist yet either) and the only other one that I see going in there soon is the lexical/block scoping PR. I think this PR is a good candidate for the "breaking" section but I don't feel strongly enough to fight beyond this comment 😄

Sure, pretty much same feeling here, I just wanted to say that we don't need to block this PR on it.

Let's discuss properly during the next office hours.

src/ast/irbuilderbpf.cpp
ajor: This is the core of the change - pid and probe id from 64 to 32 bits

jordalgo: Yeah, I don't know what was going through my head when I asked that question.

ajor:
The previous example of a broken script with tuples no longer breaks with all of @jordalgo's tuple and integer size improvements. Here's an updated example of a broken script:
```
$a = pid;
$a = 10000000000;
```

It may now fail to compile with the following error:
```
ERROR: Type mismatch for $a: trying to assign value '10000000000' which does not fit into the variable of type 'uint32'
```

To restore the old behaviour, simply cast the value of `pid` to a `uint64`:
```
$a = (uint64)pid;
$a = 10000000000;
```

I'm going to merge this PR without checking in this example because I didn't fully understand what the plan was and I'm running out of time before I'm off 😅

jordalgo:
@ajor Ack. I'm fine with merging.

PR #3443, opened 2024-09-06
Fix resizing for tuples

jordalgo:
This fixes these issues: 
- https://github.com/bpftrace/bpftrace/issues/3439
- https://github.com/bpftrace/bpftrace/issues/3444

The fix involves creating individual memcpys
for each of the tuple elements if the variable type
is a different size from the expression type -
do this recursively for nested tuples.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2245,6 +2245,33 @@ AllocaInst *CodegenLLVM::createTuple(
   return buf;
 }
 
+void CodegenLLVM::createTupleCopy(const SizedType &expr_type,
+                                  const SizedType &var_type,
+                                  Value *dst_val,
+                                  Value *src_val)
+{
+  assert(expr_type.IsTupleTy() && var_type.IsTupleTy());
+  auto *array_ty = ArrayType::get(b_.getInt8Ty(), expr_type.GetSize());
+  auto *tuple_ty = b_.GetType(var_type);
+  for (size_t i = 0; i < expr_type.GetFields().size(); ++i) {
+    SizedType &t_type = expr_type.GetField(i).type;
+    Value *offset_val = b_.CreateGEP(
+        array_ty,
+        b_.CreatePointerCast(src_val, array_ty->getPointerTo()),
+        { b_.getInt64(0), b_.getInt64(expr_type.GetField(i).offset) });
+    Value *dst = b_.CreateGEP(tuple_ty,
+                              b_.CreatePointerCast(dst_val,
+                                                   tuple_ty->getPointerTo()),
+                              { b_.getInt32(0), b_.getInt32(i) });
+    if (t_type.IsTupleTy() &&
+        t_type.GetSize() != var_type.GetField(i).type.GetSize()) {
+      createTupleCopy(t_type, var_type.GetField(i).type, dst, offset_val);
+    } else {
+      b_.CreateMemcpyBPF(dst, offset_val, t_type.GetSize());
+    }
+  }
+}
+
 void CodegenLLVM::visit(Tuple &tuple)
 {
   // Store elements on stack
@@ -2283,29 +2310,33 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
   Value *val, *expr;
   expr = expr_;
   auto [key, scoped_key_deleter] = getMapKey(map);
-  if (shouldBeInBpfMemoryAlready(assignment.expr->type)) {
-    if ((assignment.expr->type.IsStringTy() ||
-         assignment.expr->type.IsTupleTy()) &&
-        assignment.expr->type.GetSize() != map.type.GetSize()) {
+  auto &expr_type = assignment.expr->type;
+  if (shouldBeInBpfMemoryAlready(expr_type)) {
+    if (expr_type.GetSize() != map.type.GetSize()) {
       val = b_.CreateAllocaBPF(map.type, map.ident + "_val");
       b_.CreateMemsetBPF(val, b_.getInt8(0), map.type.GetSize());
-      b_.CreateMemcpyBPF(val, expr, assignment.expr->type.GetSize());
-      self_alloca = true;
-    } else
+      if (expr_type.IsTupleTy()) {
+        createTupleCopy(expr_type, map.type, val, expr);
+        self_alloca = true;
+      } else if (expr_type.IsStringTy()) {
+        b_.CreateMemcpyBPF(val, expr, expr_type.GetSize());
+        self_alloca = true;
+      } else {
+        LOG(BUG) << "Type size mismatch. Map Type Size: " << map.type.GetSize()
+                 << " Expression Type Size: " << expr_type.GetSize();
+      }
+    } else {
       val = expr;
+    }
   } else if (map.type.IsRecordTy() || map.type.IsArrayTy()) {
-    if (assignment.expr->type.is_internal) {
+    if (expr_type.is_internal) {
       val = expr;
     } else {
       // expr currently contains a pointer to the struct or array
       // We now want to read the entire struct/array in so we can save it
       AllocaInst *dst = b_.CreateAllocaBPF(map.type, map.ident + "_val");
-      b_.CreateProbeRead(ctx_,
-                         dst,
-                         map.type,
-                         expr,
-                         assignment.loc,
-                         assignment.expr->type.GetAS());
+      b_.CreateProbeRead(
+          ctx_, dst, map.type, expr, assignment.loc, expr_type.GetAS());
       val = dst;
       self_alloca = true;
     }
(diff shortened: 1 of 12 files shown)


tests/semantic_analyser.cpp
ajor, line 1: Could you swap out the new `test(..., 10)` cases for `test_error()` please?

I'm trying to deprecate the semantic analyser tests which check for an explicit return code in favour of checking the error message. Just haven't managed to do the entirety of this file yet

src/types.cpp
```diff
@@ -561,13 +561,14 @@ std::weak_ptr<const Struct> SizedType::GetStruct() const
 
 bool SizedType::FitsInto(const SizedType &t) const
 {
-  if (IsStringTy() && t.IsStringTy())
-    return GetSize() <= t.GetSize();
```
ajor, line 565: This logic looks pretty sensible - why get rid of it?

ajor (changes requested):
I'm not sure that I agree with the approach of updating the type of an assignment's right-hand-side. It creates some strange looking ASTs like this (using the improved printing from #3424):
```
BEGIN { $t = ("a", "b"); $t = ("aa", "bb"); }
```
```
Program
 BEGIN
  =
   variable: $t :: [(string[3],string[3])]
** tuple: :: [(string[3],string[3])]
    string: a :: [string[2], AS(kernel)]
    string: b :: [string[2], AS(kernel)]
  =
   variable: $t :: [(string[3],string[3])]
   tuple: :: [(string[3],string[3])]
    string: aa :: [string[3], AS(kernel)]
    string: bb :: [string[3], AS(kernel)]
```

The line I marked with `**` is saying that a tuple literal consisting of two `string[2]` nodes has a type of `(string[3],string[3])`, which just seems wrong to me. It also means that type-back-propagation is now also applying to variables, whereas before it was exclusive to maps, making variables more complicated to think about. Maybe it'll also cause some issues when we try to split out type-propagation from semantic analyser.

In my mind, it'd make more sense to allow assigning a `(string[2],string[2])` to a `(string[3],string[3])` and have codegen work it out like we do for assigning an `int32` to an `int64` or a `string[2]` to a `string[3]` already. This doesn't work currently because codegen assumes that tuple assignment can always be a memcpy instead of needing a field-by-field copy: https://github.com/bpftrace/bpftrace/blob/2a3e2497c518bfd5e0cc7bb6437c17e7a71faca7/src/ast/codegen_helper.h#L10

Alternatively, we could insert a cast node like this:
```
Program
 BEGIN
  =
   variable: $t :: [(string[3],string[3])]
   ((string[3],string[3]))
    tuple: :: [(string[2],string[2])]
     string: a :: [string[2], AS(kernel)]
     string: b :: [string[2], AS(kernel)]
  =
   variable: $t :: [(string[3],string[3])]
   tuple: :: [(string[3],string[3])]
    string: aa :: [string[3], AS(kernel)]
    string: bb :: [string[3], AS(kernel)]
```
There's some elegance to this casting approach, but the disadvantage is that it introduces a new assignment mechanism and doesn't follow the same approach already taken for mismatched int or string types (i.e. fix in codegen).

src/types.cpp
jordalgo: Nothing invokes this call path.

tests/semantic_analyser.cpp
jordalgo: Ah sure. Will do.

src/types.cpp
ajor: This script does go through it, but passes the check: `BEGIN { $a = ("a", 1); $a = ("abc", 1);}`

Maybe there's no script that can fail the check though? Still might be nice to keep as a safety net

jordalgo: Ah right, it happens for the inner tuple case. I think I got rid of it because of the type manipulation on both sides. Either way, I'll put it back because I'm not going to do that.

jordalgo:
@ajor Thanks for the detailed explanation and advice. I went the codegen route, which seemed cleaner. I'm going to look into fixing nested tuples but it might be more complicated and shouldn't block this fix IMHO.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2245,6 +2245,32 @@ AllocaInst *CodegenLLVM::createTuple(
   return buf;
 }
 
+void CodegenLLVM::createTupleCopy(const SizedType &expr_type,
+                                  const SizedType &var_type,
+                                  Value *dst_val,
+                                  Value *src_val)
+{
```
viktormalik, line 2252: Perhaps add `assert(expr_type.IsTupleTy() && var_type.IsTupleTy())` to the beginning of this method to make sure that all the below getters succeed?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2245,6 +2245,32 @@ AllocaInst *CodegenLLVM::createTuple(
...
+  auto *array_ty = ArrayType::get(b_.getInt8Ty(), expr_type.GetSize());
+  auto tuple_ty = b_.GetType(var_type);
```
viktormalik, line 2254: Super nit: why `array_type` has `auto *` while `tuple_ty` has just `auto`?

tests/semantic_analyser.cpp
```diff
@@ -3103,19 +3105,43 @@ TEST(semantic_analyser, tuple)
...
-  test(R"_(BEGIN { $t = ((int32)1, (int64)2); $t = ((int64)1, (int32)2); })_",
-       10);
 
   test(R"_(BEGIN { @t = (1, 2); @t = (4, "other"); })_", 10);
   test(R"_(BEGIN { @t = (1, 2); @t = 5; })_", 1);
   test(R"_(BEGIN { @t = (1, count()) })_", 1);
+
+  test(R"_(BEGIN { $t = (1, (2, 3)); $t = (4, ((int8)5, 6)); })_");
+
+  test_error(R"_(BEGIN { $t = (1, ((int8)2, 3)); $t = (4, (5, (int8)6)); })_",
```
viktormalik, line 3129: I'm a bit lost here. Why does the first one pass but the second one trigger an error?

viktormalik (commented):
Generally LGTM, nice and clean approach. I've one thing to clarify, though (+ a couple of nits).

tests/semantic_analyser.cpp
jordalgo: Because the '5' in the first position (of the nested tuple) of the second assignment doesn't fit into the `int8` in the same position in the first assignment. The second int8 cast for the 6 is kinda superfluous, I'll remove it.

viktormalik: Right, that's probably what confused me. Thanks for the explanation!

viktormalik (approved):
Looks good!

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2283,29 +2310,32 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
...
-    } else
+      if (expr_type.IsTupleTy()) {
+        createTupleCopy(expr_type, map.type, val, expr);
+        self_alloca = true;
+      } else if (expr_type.IsStringTy()) {
+        b_.CreateMemcpyBPF(val, expr, expr_type.GetSize());
+        self_alloca = true;
+      } else {
+        val = expr;
+      }
```
ajor, line 2326: I know the logic is equivalent, but the new layout highlights a potential issue more clearly: it wasn't expected that any non-tuple, non-string type could have a mismatched size. i.e. this `else` branch probably won't work right.

I'm thinking that a `buf` could break it, but should we just turn it into a `LOG(BUG)` for now?

tests/runtime/variable
```diff
@@ -63,7 +63,25 @@ PROG BEGIN { @["hello", 0] = 0; } i:ms:1 { @["hi", 1] = 1; exit(); }
 EXPECT @[hi, 1]: 1
 TIMEOUT 2
 
-NAME tuple string resize
-PROG BEGIN { @ = ("hello", 0); } i:ms:1 { @ = ("hi", 1); exit(); }
-EXPECT @: (hi, 1)
+NAME map tuple string resize
+PROG BEGIN { @[1] = ("hi", 1); @[1] = ("hellolongstr", 2); @[1] = ("by", 3); exit(); }
```
ajor, line 67: Nit: The third assignment goes back to the size of the first - couldn't that hide potential issues with the second assignment?

tests/codegen/tuple.cpp
```diff
@@ -11,6 +11,28 @@ TEST(codegen, tuple)
...
+TEST(codegen, tuple_map_val_different_sizes)
+{
+  test(R"_(k:f { @a = (1, "hi"); @a = (1, "hellolongstr"); })_",
+
+       NAME);
+}
+
+TEST(codegen, tuple_variable_different_sizes)
+{
+  test(R"_(k:f { $t = (1, "hi"); $t = (1, "hellolongstr"); print($t); })_",
```
ajor, line 23: Nit: We can cut down on irrelevant generated code by removing the `print()`

tests/codegen/tuple.cpp
```diff
@@ -11,6 +11,28 @@ TEST(codegen, tuple)
...
+
+       NAME);
+}
+
+TEST(codegen, nested_tuple_different_sizes)
+{
+  test(
+      R"_(k:f { $t = (1, ("hi", 3)); $t = (1, ("hellolongstr", 4)); print($t); })_",
```
ajor, line 31: Same

ajor (approved):
Looks good! Just a few nits

src/ast/passes/codegen_llvm.cpp
jordalgo: I wonder if it would make it through the semantic_analyser pass. I'm fine to turn it into a `LOG(BUG)` or a `LOG(WARNING)` if we're worried it might break existing scripts.

tests/runtime/variable
jordalgo: I think I can get rid of the first assignment as I want to show that a shorter string doesn't have issues when the map type holds a longer string.

PR #3450, opened 2024-09-11
build all internal libraries as static

eli-schwartz:
A bunch of utility libraries are created as internal implementation details of the build process. None of them are installed by running `make install`.

By default, cmake builds static libraries, but this is a global setting. Correct use of cmake is to annotate internal utility libaries as explicitly static. Without doing this, the user configuration may end up creating shared libraries, which then break the software since none of these "shared libraries" with highly generic names such as "ast" or "arch" or "runtime" are actually installed.


danobi (approved):
Makes sense. Thanks!

PR #3453, opened 2024-09-16
BTF: Handle btf__parse_raw() errors

danobi:
btf__parse_raw() can fail and result in NULL pointer dereference
somewhere down the road, check its return value.

Without this, I get a segmentation fault when accidentally passing an
incorrect BTF file using BPFTRACE_BTF.

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -51,6 +51,10 @@ BTF::BTF(const std::set<std::string> &modules) : state(NODATA)
     btf_objects.push_back(
         BTFObj{ .btf = btf__parse_raw(path), .id = 0, .name = "" });
     vmlinux_btf = btf_objects.back().btf;
+    if (!vmlinux_btf) {
+      LOG(WARNING) << "BTF: failed to parse BTF from " << path;
+      return;
+    }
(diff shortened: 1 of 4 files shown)


src/btf.cpp
```diff
@@ -51,6 +51,10 @@ BTF::BTF(const std::set<std::string> &modules) : state(NODATA)
     btf_objects.push_back(
         BTFObj{ .btf = btf__parse_raw(path), .id = 0, .name = "" });
     vmlinux_btf = btf_objects.back().btf;
+    if (!vmlinux_btf) {
+      LOG(V1) << "BTF: failed to parse BTF";
```
jordalgo, line 55: nit: I feel like this should be a LOG(WARNING) as it can affect the program behavior.

danobi: Good catch - not nit IMO. If user explicitly requested this BTF and we cannot parse it, we really should let user know. The non-requested V1 log below is fine IMO. We opportunistically parse BTF in the non-requested case.

jordalgo (approved):

PR #3456, opened 2024-09-18
Use scratch global variable for script-dependent tuple allocations

ryantimwilson:
This PR starts fixing https://github.com/bpftrace/bpftrace/issues/3431, ensuring tuples use scratch maps vs stack allocation. This will allow large strings to be stored in tuples and improve user experience.

This PR starts migrating script-dependent data structures to global variables, starting with tuples and implicit tuples generated from for loop variables. This will allow tuples larger than the BPF stack size, which will specifically allow increasing default max string size to 1024.

To deal with nested tuples we create a scratch global variable with size = number of tuples explored via AST - similar to string buffers. This is done to avoid overwriting multiple tuples used simultaneously by the same CPU.

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -498,10 +498,14 @@ Value *IRBuilderBPF::createScratchBuffer(
   auto cmp = CreateICmp(CmpInst::ICMP_ULE, cpu_id, max, "cpuid.min.cmp");
   auto select = CreateSelect(cmp, cpu_id, max, "cpuid.min.select");
 
+  // Note the 1st index is 0 because we're pointing to
+  // ValueType var[MAX_CPU_ID + 1][num_elements]
+  // 2nd/3rd/4th indexes actually index into the array
+  // See https://llvm.org/docs/LangRef.html#id236
   return CreateGEP(GetType(type),
                    CreatePointerCast(module_.getGlobalVariable(config.name),
                                      GetType(type)->getPointerTo()),
-                   { select, getInt64(key), getInt64(0) });
+                   { getInt64(0), select, getInt64(key), getInt64(0) });
 }
 
 /*
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -479,6 +479,13 @@ Value *IRBuilderBPF::CreateGetFmtStringArgsScratchBuffer(const location &loc)
       bpftrace::globalvars::GlobalVar::FMT_STRINGS_BUFFER, loc);
 }
 
+Value *IRBuilderBPF::CreateTupleScratchBuffer(const location &loc, int key)
+{
+  return createScratchBuffer(bpftrace::globalvars::GlobalVar::TUPLE_BUFFER,
+                             loc,
+                             key);
+}
+
 Value *IRBuilderBPF::createScratchBuffer(
     bpftrace::globalvars::GlobalVar globalvar,
     const location &loc,
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2215,20 +2215,22 @@ void CodegenLLVM::compareStructure(SizedType &our_type, llvm::Type *llvm_type)
 /*
  * createTuple
  *
- * Constructs a tuple on the stack from the provided values.
+ * Constructs a tuple on the scratch buffer from the provided values.
  */
-AllocaInst *CodegenLLVM::createTuple(
+Value *CodegenLLVM::createTuple(
     const SizedType &tuple_type,
     const std::vector<std::pair<llvm::Value *, const location *>> &vals,
-    const std::string &name)
+    const location &loc)
 {
   auto tuple_ty = b_.GetType(tuple_type);
   size_t tuple_size = datalayout().getTypeAllocSize(tuple_ty);
-  AllocaInst *buf = b_.CreateAllocaBPF(tuple_ty, name);
+  auto buf = b_.CreatePointerCast(
+      b_.CreateTupleScratchBuffer(loc, async_ids_.tuple()),
+      tuple_ty->getPointerTo());
   b_.CreateMemsetBPF(buf, b_.getInt8(0), tuple_size);
 
   for (size_t i = 0; i < vals.size(); ++i) {
-    auto [val, loc] = vals[i];
+    auto [val, vloc] = vals[i];
     SizedType &type = tuple_type.GetField(i).type;
 
     Value *dst = b_.CreateGEP(tuple_ty,
@@ -2238,7 +2240,7 @@ AllocaInst *CodegenLLVM::createTuple(
     if (inBpfMemory(type))
       b_.CreateMemcpyBPF(dst, val, type.GetSize());
     else if (type.IsArrayTy() || type.IsRecordTy())
-      b_.CreateProbeRead(ctx_, dst, type, val, *loc);
+      b_.CreateProbeRead(ctx_, dst, type, val, *vloc);
     else
       b_.CreateStore(val, dst);
   }
@@ -2280,7 +2282,6 @@ void CodegenLLVM::createTupleCopy(const SizedType &expr_type,
 
 void CodegenLLVM::visit(Tuple &tuple)
 {
-  // Store elements on stack
   llvm::Type *tuple_ty = b_.GetType(tuple.type);
 
   compareStructure(tuple.type, tuple_ty);
@@ -2293,10 +2294,11 @@ void CodegenLLVM::visit(Tuple &tuple)
     scoped_dels.emplace_back(accept(elem));
     vals.push_back({ expr_, &elem->loc });
   }
-  AllocaInst *buf = createTuple(tuple.type, vals, "tuple");
+  auto buf = createTuple(tuple.type, vals, tuple.loc);
 
+  // No need to end buf lifetime since tuple allocation is done via scratch maps
+  // and not on stack
   expr_ = buf;
-  expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
 }
 
 void CodegenLLVM::visit(ExprStatement &expr)
@@ -4402,11 +4404,10 @@ Function *CodegenLLVM::createForEachMapCallback(const For &f, llvm::Type *ctx_t)
   }
 
   // Create decl variable for use in this iteration of the loop
-  AllocaInst *tuple = createTuple(f.decl->type,
-                                  { { key, &f.decl->loc },
-                                    { val, &f.decl->loc } },
-                                  f.decl->ident);
-  variables_[f.decl->ident] = VariableLLVM{ tuple, tuple->getAllocatedType() };
+  auto tuple = createTuple(f.decl->type,
+                           { { key, &f.decl->loc }, { val, &f.decl->loc } },
+                           f.decl->loc);
+  variables_[f.decl->ident] = VariableLLVM{ tuple, b_.GetType(f.decl->type) };
 
   // 1. Save original locations of variables which will form part of the
   //    callback context
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -55,6 +55,14 @@ std::optional<RequiredResources> ResourceAnalyser::analyse()
         bpftrace::globalvars::GlobalVar::MAX_CPU_ID);
   }
 
+  if (resources_.max_tuple_size > 0) {
+    assert(resources_.tuple_buffers > 0);
+    resources_.needed_global_vars.insert(
+        bpftrace::globalvars::GlobalVar::TUPLE_BUFFER);
+    resources_.needed_global_vars.insert(
+        bpftrace::globalvars::GlobalVar::MAX_CPU_ID);
+  }
+
   return std::optional{ std::move(resources_) };
 }
 
@@ -241,6 +249,25 @@ void ResourceAnalyser::visit(Map &map)
   auto &map_info = resources_.maps_info[map.ident];
   map_info.value_type = map.type;
   map_info.key_type = map.key_type;
+}
+
+void ResourceAnalyser::visit(Tuple &tuple)
+{
+  Visitor::visit(tuple);
+
+  resources_.tuple_buffers++;
+  resources_.max_tuple_size = std::max(resources_.max_tuple_size,
+                                       tuple.type.GetSize());
+}
+
+void ResourceAnalyser::visit(For &f)
+{
+  Visitor::visit(f);
+
+  // Need tuple per for loop to store key and value
+  resources_.tuple_buffers++;
+  resources_.max_tuple_size = std::max(resources_.max_tuple_size,
+                                       f.decl->type.GetSize());
 }
 
 bool ResourceAnalyser::uses_usym_table(const std::string &fun)
--- a/src/ast/passes/resource_analyser.h
+++ b/src/ast/passes/resource_analyser.h
@@ -33,6 +33,8 @@ private:
   void visit(Builtin &map) override;
   void visit(Call &call) override;
   void visit(Map &map) override;
+  void visit(Tuple &tuple) override;
+  void visit(For &f) override;
 
   // Determines whether the given function uses userspace symbol resolution.
   // This is used later for loading the symbol table into memory.
--- a/src/required_resources.h
+++ b/src/required_resources.h
@@ -103,6 +103,10 @@ public:
   // rather than duplicating it in CodegenResources.
   uint64_t max_fmtstring_args_size = 0;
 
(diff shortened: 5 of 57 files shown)


ajor:
Thanks for working on this!

I haven't been through the code but this seems like a sensible approach. A couple of high level points:
1. There might be some conflicts with #3390. With the increased number of branches from map-lookup null checks that these two PRs will introduce when combined, we should confirm that there aren't any verifier issues with larger tracing scripts. If they work, it'd also be good to know how much headroom common scripts still have. I know @danobi's planning to do some kernel work to eliminate the need for branches in the future.
  For context, some bpftrace-based games (e.g. https://github.com/amiremohamadi/bpfsnake) have started hitting verifier limits due to these branches. They use lots of unrolled loops so do generate tons of code, but I'd just like to confirm that no real tools get broken.
2. This PR makes temporary big tuples work, but what's the plan for handling tuples in maps and variables? I hit an assertion failure when running this script on this PR: `config = {max_strlen = 1000} BEGIN { @x = (1,str(0)); @y = @x }`

src/ast/passes/codegen_resources.h
```diff
@@ -13,6 +13,9 @@ struct CodegenResources {
   bool needs_elapsed_map = false;
   bool needs_join_map = false;
   uint32_t str_buffers = 0;
+  size_t max_tuple_size = 0;
+  size_t curr_tuple_depth = 0;
+  size_t max_tuple_depth = 0;
```
danobi, line 18: Maybe make `curr_tuple_depth` private?

danobi (commented):
minor nit but lgtm!

danobi:
> With the increased number of branches from map-lookup null checks that these two PRs will introduce when combined, we should confirm that there aren't any verifier issues with larger tracing scripts

At least for the tools checked into this repo, `tools-parsing-test.sh` is confirming they compile and load still

> it'd also be good to know how much headroom common scripts still have

In-kernel verifier keeps track of this "headroom" as `BPF_COMPLEXITY_LIMIT_JMP_SEQ - env->stack_size`. Not sure if it's reported anywhere.

src/ast/passes/codegen_resources.h
ryantimwilson: Fixed in latest version!

ryantimwilson:
Thanks @ajor @danobi for all the comments!

> There might be some conflicts with https://github.com/bpftrace/bpftrace/pull/3390.There might be some conflicts with https://github.com/bpftrace/bpftrace/pull/3390.

Its possible. Its a pretty long PR and I'm not super familiar with bpftrace codebase yet. I'll leave up to you and @danobi to determine which one we should merge first.

> For context, some bpftrace-based games (e.g. https://github.com/amiremohamadi/bpfsnake) have started hitting verifier limits due to these branches. They use lots of unrolled loops so do generate tons of code, but I'd just like to confirm that no real tools get broken.

Thanks for the pointer! Indeed, bpftrace aborts on my Centos 6.9 machine with this PR's bpftrace...but it also aborts with the same error with stable `bpftrace v0.21.0-2a3e`. It does seem related to branching but happening in LLVM. Do you know what kernel + bpftrace version snake.bt works with?
```
$ sudo ./result/bin/bpftrace ~/bpfsnake/snake.bt
LIBCLANG FATAL ERROR: Branch target out of insn range
Aborted
$ sudo bpftrace ~/bpfsnake/snake.bt
LIBCLANG FATAL ERROR: Branch target out of insn range
Aborted
```

> I don't think it's likely that most scripts will come close to the limit, but perhaps something that initialises a lot of map entries at startup, like the script described in https://github.com/bpftrace/bpftrace/issues/3257.I don't think it's likely that most scripts will come close to the limit, but perhaps something that initialises a lot of map entries at startup, like the script described in https://github.com/bpftrace/bpftrace/issues/3257.

I tried a program with the following repeated 1000 times.
```
@x = ("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", 1);
```

You're right. On this PR, this hits the `LLVM ERROR: Branch target out of insn range` error. On stable bpftrace, it does not as there is no branching.

At smaller values < 1000 (starting around 500), I stop hitting the LLVM error and start getting a kernel error:
```
$ sudo result/bin/bpftrace tuple_repro.bt -v
AST node count: 1016
Attaching 1 probe...
ERROR: Error loading BPF program for BEGIN_1.
Kernel error log:
processed 17397 insns (limit 1000000) max_states_per_insn 0 total_states 204 peak_states 204 mark_read 202

```

Doing some manual binary search, I found the limit is roughly 130 for attaching the probe to be successful, meaning 130 `br` instructions in LLVM (vs. 0 on stable).

So I need to dig more into this prior to merging, it seems pretty bad to have a 10x regression in # of variables to support large strings.

> In-kernel verifier keeps track of this "headroom" as BPF_COMPLEXITY_LIMIT_JMP_SEQ - env->stack_size. Not sure if it's reported anywhere.

Looks like BPF_COMPLEXITY_LIMIT_JMP_SEQ is 8192. However, I wonder why I see a limit of 130 br - even with 3 jumps per br (-> success, ->failure, -> merge), that's only 390 jumps.

> Similar plan. Described in https://github.com/bpftrace/bpftrace/issues/3431Similar plan. Described in https://github.com/bpftrace/bpftrace/issues/3431

Yup I plan to split changes between tuples, map keys and map values.

danobi:
@ajor @viktormalik, and I chatted offline at LPC yesterday. The plan to address this is to move the percpu scratch maps to global data. Global variables don't require a null check and have been supported for a long time: https://github.com/torvalds/linux/commit/6316f78306c171f5a857a2442dbeebc7baab3566#diff-4f15343cf34fde0cbb97aba726a48c63a4c320ed91fce341450c547f60ddb5ab

Merged in 5.2. And 4.19 LTS (https://www.kernel.org/category/releases.html) is EOL'd in december of this year. So close enough where I'd feel comfortable relying on global data being supported everywhere.

The aborts are probably b/c BPF chokes on having so many branches. Seems like both LLVM backend and verifier are kinda sensitive in these extreme cases.

I plan on working on the global data conversion. I think it should be fairly straightforward and it's pretty high on my priority list now. Don't let that block you on the stack allocation conversions.

ryantimwilson:
Thanks for the update @danobi @ajor @viktormalik!

So I figured out why the program above wasn't working with `sudo BPFTRACE_LOG_SIZE=100000000 result/bin/bpftrace tuple_repro.bt -d verifier -v`. Its due to:
```
insn 12275 cannot be patched due to 16-bit range
```

Looks like it is indeed due to too many jumps: https://patchwork.ozlabs.org/project/netdev/patch/20190402042749.3670015-6-ast@kernel.org/

Although curiously, I can now get up to ~325 lines before hitting the verifier error and ~350 lines before hitting the LLVM instruction error.

Talking to @danobi on IM, I'd hate to merge something which introduces this regression, so I will start moving tuple scratch map to global data which should remove the branching.

ryantimwilson:
@danobi @ajor @viktormalik Here is the PR to start migrating to global data maps, starting with fmt strings scratch map for simplicity: https://github.com/bpftrace/bpftrace/pull/3460

src/ast/irbuilderbpf.cpp
```diff
@@ -498,10 +505,14 @@ Value *IRBuilderBPF::createScratchBuffer(
...
 
+  // Note the 1st index is 0 because we're pointing to
+  // ValueType var[MAX_CPU_ID + 1][num_elements]
+  // 2nd/3rd/4th indexes actually index into the array
+  // See https://llvm.org/docs/LangRef.html#id236
   return CreateGEP(GetType(type),
                    CreatePointerCast(module_.getGlobalVariable(config.name),
                                      GetType(type)->getPointerTo()),
-                   { select, getInt64(key), getInt64(0) });
+                   { getInt64(0), select, getInt64(key), getInt64(0) });
```
ryantimwilson, line 515: TL;DR: There was a bug in my createScratchBuffer() function but my analysis indicates it doesn't cause bad effects for fmtstr as it only affects where key > 0. So I think a forward fix is fine. But let me know if I'm wrong.

@danobi @jordalgo this was a bug in my initial fmtstrings implementation. Basically because the type was a pointer to an array, the first index needs to be 0 to get the actual array itself. See https://llvm.org/docs/LangRef.html#id236

In theory, this should cause out-of-bounds access in my previous PR for CPU IDs > 0

The only reason I don't think it did is because the LLVM IR instruction:
```
%12 = getelementptr [1 x [1 x [24 x i8]]], ptr @fmt_str_buf, i64 %cpuid.min.select, i64 0, i64 0
```

Generates something like:
```
r9 = <global var address>
16: (85) call bpf_get_smp_processor_id#8
17: (2d) if r6 > r0 goto pc+1         ; R0_w=55 R6=55
18: (bf) r0 = r6                      ; R0=55 R6=55
19: (27) r0 *= 48                     ; R0_w=2640
20: (0f) r9 += r0
```

Instruction 19 is basically multiplying bpf_get_smp_processor_id() * sizeof([1 x [1 x [24 x i8]]]) which is 24.

So even with multiple CPUs involved, the BPF assembly still works because the real data structure is [N x [1 x [24 x i8]]] as key is always 0 in fmt strings buffer.

But for this PR where key >= 0, without this fix, `getInt64(key)` thinks it indexing into `[1 x [2 x [[24 x i8]]])` and not `[2 x [[24 x i8]]` so we end up overrunning the buffer. For example, I saw stuff like:
```
r1 = 6
*(u64 *)(r9 +56) = r1
```

danobi: Nice catch! GEP is tricky like that :P

src/required_resources.h
```diff
@@ -103,6 +103,10 @@ class RequiredResources {
   // rather than duplicating it in CodegenResources.
   uint64_t max_fmtstring_args_size = 0;
 
+  // Required for sizing of tuple scratch buffer
+  size_t max_tuple_size = 0;
+  size_t max_tuple_depth = 0;
+
```
danobi, line 109: Better fit for codegen_resources.h? Since you are using initial value actual_size trick, so these values do not need to be serialized for AOT

ryantimwilson: Can't use codegen_resources.h because I need to set `needed_global_variables` based on these values.

I suppose I could introduce a dependency between codegen_resources and required_resources where required_resources checks codegen_resources and adds necessary needed_global_variables. What do you think about that or leaving it as it is?

src/ast/irbuilderbpf.cpp
ryantimwilson: Sounds good, I'll split this into 2 commits

src/required_resources.h
danobi: Sorry, not quite following. The only references I see to `max_tuple_size` and `max_tuple_depth` are from `globalvars::get_type()`. But `get_type()` is called like:

```
CodegenLLVM::generate_ir()
  CodegenLLVM::generate_global_vars()
    bpftrace::globalvars::get_type()
```

`CodegenResources` is present as a local variable in `CodegenLLVM::generate_ir()`, so it's possible to thread both `RequiredResources` and `CodegenResources` through to `get_type()` just like we do for `CodegenLLVM::generate_maps()` right?

ryantimwilson: See this code in ResourceAnalyser::analyse() in my PR:
```
  if (resources_.max_tuple_size > 0) {
    assert(resources_.max_tuple_depth > 0);
    resources_.needed_global_vars.insert(
        bpftrace::globalvars::GlobalVar::TUPLE_BUFFER);
    resources_.needed_global_vars.insert(
        bpftrace::globalvars::GlobalVar::MAX_CPU_ID);
  }
```

We need to add MAX_CPU_ID and TUPLE_BUFFER to needed_global_vars for AOT to let AOT know to resize these read-write .data.* ELF regions.

Yea I suppose we could append to bpftrace.resources in CodegenLLVM::generate_global_vars() based on codegen_resoureces...but it feels a bit unclean to me as its introducing a dependency between those objects buried in CodegenLLVM module.

danobi: Ah, got it. Missed the `> 0` checks. Let's leave it as-is then.

danobi (approved):

src/types.h
```diff
@@ -717,6 +717,8 @@ enum class GlobalVar {
   MAX_CPU_ID,
   // Scratch buffer used for format strings to avoid BPF stack allocation limits
```
jordalgo, line 718: nit: can probably combine these two comments so you don't have to keep repeating it

ryantimwilson:
@jordalgo addressed your comments!

src/ast/passes/resource_analyser.h
```diff
@@ -47,6 +48,7 @@ class ResourceAnalyser : public Visitor {
   Probe *probe_;
 
   int next_map_id_ = 0;
+  size_t curr_tuple_depth_ = 0;
```
jordalgo, line 51: Nit: why not use the same member var name as in codegen `tuple_depth_` ?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2280,23 +2280,47 @@ void CodegenLLVM::createTupleCopy(const SizedType &expr_type,
...
+  for (size_t i = 0; i < tuple.elems.size(); i++) {
+    // Evaluate expression using DFS
+    auto elem = tuple.elems[i];
     scoped_dels.emplace_back(accept(elem));
-    vals.push_back({ expr_, &elem->loc });
+
+    // Then copy or store child expression
+    // For child tuples, to avoid the child overwriting the parent's tuple
+    // scratch map, we use a separate scratch map key according to the tuple
+    // depth e.g. child key = N + 1 and parent key = N
```
jordalgo, line 2302: Feel free to ignore this but it took me a second to understand what this was saying. Maybe slight rewording:

```suggestion
    // For child tuples, to avoid the child overwriting the parent's tuple
    // scratch map, we use tuple_depth_ as the scratch map key 
    // e.g. for ("a", ("b", ("c", "d"))) the top tuple key (containing "a") is 0, then the inner tuple key (containing "b") is 1, then the tuple containing "c" and "d", it's key is 2 and so forth.
```

ryantimwilson:
So there is a bug:

```
$ sudo build-nix/src/bpftrace -e 'BEGIN { @map[(1, 2)] = (3,4); print(@map); }'
Attaching 1 probe...
@map[1, 2]: (1, 2)
```

Seems what's happening is:
1. We recurse into map value and write tuple (3,4) to scratch buffer
2. We recurse into map key and write tuple (1,2) to scratch buffer
3. We then evaluate assignment and read from buffer for both key and value which both read (1,2)

This PR has an implicit dependency that the tuple is memcpy-ed out of the scratch location before another tuple takes its place. But that's not necessarily true for all callers

Proper fix seems to be to trace allocation lifetimes. But this seems hard so we're gonna do what @danobi suggested and just use a buffer per tuple like we do for strings

src/ast/passes/codegen_llvm.cpp
ryantimwilson: With my new update we're no longer using this approach :) So resolving but lets re-review the new version and see where I can add comments!

src/ast/passes/resource_analyser.h
ryantimwilson: No longer needed in current version

src/ast/passes/resource_analyser.cpp
```diff
@@ -243,6 +251,25 @@ void ResourceAnalyser::visit(Map &map)
...
+  resources_.tuple_buffers++;
+  resources_.max_tuple_size = std::max(resources_.max_tuple_size,
+                                       tuple.type.GetSize());
+}
+
+void ResourceAnalyser::visit(For &f)
+{
+  Visitor::visit(f);
+
+  // Need tuple per for loop to store key and value
```
jordalgo, line 267: Does this handle the case where the key has a nested tuple?

jordalgo (approved):
LGTM!

src/ast/passes/resource_analyser.cpp
ryantimwilson: Yup! Because Visitor::visit(f) will recurse into the declared expression which will recurse into Visitor::visit(tuple)

tests/runtime/tuples
danobi, line 1: can you add an additional test for implicit conversion of multi key maps to tuples?
eg. `@["asdf", 2] = (3, 4)` works as expected

danobi: Happens here: https://github.com/bpftrace/bpftrace/blob/1874688980a9a63d9ec8447b815c724b3f7acc1e/src/parser.yy#L660-L666

danobi:
LGTM, but PR description and second commit msg are stale. Please update. And lemme think on it some more while driving in - see if there's any more edge cases

tests/runtime/tuples
ryantimwilson: Done!

danobi (approved):

PR #3457, opened 2024-09-19
Fix distro builds

danobi:
See commit msgs and discussion in https://github.com/bpftrace/bpftrace/pull/3433

##### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,7 +151,15 @@ else()
 endif()
 
 set(MIN_LLVM_MAJOR 13)
-set(MAX_LLVM_MAJOR 18)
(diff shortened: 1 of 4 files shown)


ajor (approved):
🚀

ajor:
Oh just thought of one small thing - maybe add a comment to the Dockerfiles explaining that `CMAKE_BUILD_TYPE=Release` can be used instead of "Debug" if the user isn't trying to develop bpftrace?

We had this problem years ago where everyone would end up running debug builds unnecessarily because they just copied the commands directly from places like this. We do actually tell people to follow the Dockerfile instructions when building bpftrace themselves: https://github.com/bpftrace/bpftrace/blob/master/INSTALL.md#building-bpftrace

Issue #63, opened 2018-09-07
lhist() 0 entry has 0M


brendangregg:
extra "M" on 0:

```
# ./src/bpftrace -e 'BEGIN { @ = lhist(0, 0, 100, 10); }'
Attaching 1 probe...
^C

@:
[0M, 10)               1 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
```

brendangregg:
fixed by #216

PR #3460, opened 2024-09-20
Migrate fmt string scratch map to global variable

ryantimwilson:
To resolve the issues with too much branching introduced from scratch maps in https://github.com/bpftrace/bpftrace/pull/3456, this PR attempts to use writable array global variables in .data.* section to avoid jumps (since the verifier can guarantee the pointer points to real memory): https://github.com/torvalds/linux/commit/6316f78306c171f5a857a2442dbeebc7baab3566#diff-4f15343cf34fde0cbb97aba726a48c63a4c320ed91fce341450c547f60ddb5ab

To demonstrate this, I migrate fmt strings to this approach since it is the simplest map to migrate.

To simulate per-CPU maps, which aren't available in global variables, we need to do the following:
1. For AOT compatibility, re-initialize size for variable to num_cpus * max_fmt_string_size via bpf_map__set_initial_value() on program load.
2. We give each array its own custom .data.<name> section in ELF since bpf_map__set_initial_value() can only resize the last array variable in the section.
3. When accessing the global array, we call bpf_get_smp_processor_id to get the CPU offset into the array.

Now running a program with the following line repeated N times:
```
printf("hello %s %d", "a", 123);
```

At ~350 lines, master will crash due to verifier errors due to long jumps.

On this PR, I tried 1000 lines and it worked. I'm guessing we can go infinitely long now!

Notes:
1. We do not do a bounds check to verify cpu_id <= num_cpus. It seems kernel 6.9 verifier is smart enough to know this but I need to check 5.2 and above.
2. AOT does not seem backwards compatible as I needed to add `max_fmtstring_args_size` to AOT serialization. Is this okay? If not, we can use scratch map around if max_fmtstring_args_size == 0

### DIFF
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -288,11 +288,12 @@ DIGlobalVariableExpression *DIBuilderBPF::createMapEntry(
       file, name, "global", file, 0, map_entry_type, false);
 }
 
-DIGlobalVariableExpression *DIBuilderBPF::createGlobalInt64(
-    std::string_view name)
+DIGlobalVariableExpression *DIBuilderBPF::createGlobalVariable(
+    std::string_view name,
+    const SizedType &stype)
 {
   return createGlobalVariableExpression(
-      file, name, "global", file, 0, getInt64Ty(), false);
+      file, name, "global", file, 0, GetType(stype), false);
 }
 
 } // namespace bpftrace::ast
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -473,15 +473,35 @@ CallInst *IRBuilderBPF::CreateGetStrScratchMap(int idx,
                              idx);
 }
 
-CallInst *IRBuilderBPF::CreateGetFmtStringArgsScratchMap(
-    BasicBlock *failure_callback,
-    const location &loc)
+Value *IRBuilderBPF::CreateGetFmtStringArgsScratchBuffer(const location &loc)
 {
-  return createGetScratchMap(to_string(MapType::FmtStringArgs),
-                             "fmtstr",
-                             GET_PTR_TY(),
-                             loc,
-                             failure_callback);
+  return createScratchBuffer(
+      bpftrace::globalvars::GlobalVar::FMT_STRINGS_BUFFER, loc);
+}
+
+Value *IRBuilderBPF::createScratchBuffer(
+    bpftrace::globalvars::GlobalVar globalvar,
+    const location &loc,
+    size_t key)
+{
+  auto config = globalvars::get_config(globalvar);
+  // ValueType var[MAX_CPU_ID + 1][num_elements]
+  auto type = globalvars::get_type(globalvar, bpftrace_.resources);
+
+  // Get CPU ID
+  auto cpu_id = CreateGetCpuId(loc);
+  auto max = CreateLoad(getInt64Ty(),
+                        module_.getGlobalVariable(to_string(
+                            bpftrace::globalvars::GlobalVar::MAX_CPU_ID)));
+  // Verify CPU ID is between 0 and MAX_CPU_ID to ensure we pass BPF
+  // verifer on older 5.2 kernels
+  auto cmp = CreateICmp(CmpInst::ICMP_ULE, cpu_id, max, "cpuid.min.cmp");
+  auto select = CreateSelect(cmp, cpu_id, max, "cpuid.min.select");
+
+  return CreateGEP(GetType(type),
+                   CreatePointerCast(module_.getGlobalVariable(config.name),
+                                     GetType(type)->getPointerTo()),
+                   { select, getInt64(key), getInt64(0) });
 }
 
 /*
@@ -656,12 +676,13 @@ Value *IRBuilderBPF::CreatePerCpuMapAggElems(Value *ctx,
   CreateBr(while_cond);
   SetInsertPoint(while_cond);
 
-  auto *cond = CreateICmp(CmpInst::ICMP_ULT,
-                          CreateLoad(getInt32Ty(), i),
-                          CreateLoad(getInt32Ty(),
-                                     module_.getGlobalVariable(
-                                         bpftrace::globalvars::NUM_CPUS)),
-                          "num_cpu.cmp");
+  auto *cond = CreateICmp(
+      CmpInst::ICMP_ULT,
+      CreateLoad(getInt32Ty(), i),
+      CreateLoad(getInt32Ty(),
+                 module_.getGlobalVariable(
+                     to_string(bpftrace::globalvars::GlobalVar::NUM_CPUS))),
+      "num_cpu.cmp");
   CreateCondBr(cond, while_body, while_end);
 
   SetInsertPoint(while_body);
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -3422,7 +3422,7 @@ void CodegenLLVM::createFormatStringCall(Call &call,
                << " does not match LLVM offset=" << expected_offset;
   }
 
-  Value *fmt_args = b_.CreateGetFmtStringArgsScratchMap(nullptr, call.loc);
+  Value *fmt_args = b_.CreateGetFmtStringArgsScratchBuffer(call.loc);
   // The struct is not packed so we need to memset it
   b_.CreateMemsetBPF(fmt_args, b_.getInt8(0), struct_size);
 
@@ -3566,7 +3566,7 @@ void CodegenLLVM::createPrintNonMapCall(Call &call, int id)
   StructType *print_struct = b_.GetStructType(struct_name.str(),
                                               elements,
                                               true);
-  Value *buf = b_.CreateGetFmtStringArgsScratchMap(nullptr, call.loc);
+  Value *buf = b_.CreateGetFmtStringArgsScratchBuffer(call.loc);
   size_t struct_size = datalayout().getTypeAllocSize(print_struct);
 
   // Store asyncactionid:
@@ -3791,28 +3791,21 @@ void CodegenLLVM::generate_maps(const RequiredResources &required_resources,
                       1,
                       CreateInt(loss_cnt_key_size),
                       CreateInt(loss_cnt_val_size));
-
-  if (required_resources.max_fmtstring_args_size > 0) {
-    createMapDefinition(to_string(MapType::FmtStringArgs),
-                        libbpf::BPF_MAP_TYPE_PERCPU_ARRAY,
-                        1,
-                        CreateInt32(),
-                        CreateArray(required_resources.max_fmtstring_args_size,
-                                    CreateInt8()));
-  }
 }
 
 void CodegenLLVM::generate_global_vars(const RequiredResources &resources)
 {
-  for (const auto &name : resources.needed_global_vars) {
+  for (const auto global_var : resources.needed_global_vars) {
+    auto config = bpftrace::globalvars::get_config(global_var);
+    auto type = bpftrace::globalvars::get_type(global_var, resources);
     auto var = llvm::dyn_cast<GlobalVariable>(
-        module_->getOrInsertGlobal(name, b_.getInt64Ty()));
-    var->setInitializer(b_.getInt64(1));
-    var->setConstant(true);
-    var->setSection(bpftrace::globalvars::SECTION_NAME);
+        module_->getOrInsertGlobal(config.name, b_.GetType(type)));
+    var->setInitializer(ConstantAggregateZero::get(b_.GetType(type)));
+    var->setConstant(config.read_only);
+    var->setSection(config.section);
     var->setExternallyInitialized(true);
     var->setDSOLocal(true);
-    var->addDebugInfo(debug_.createGlobalInt64(name));
+    var->addDebugInfo(debug_.createGlobalVariable(config.name, type));
   }
 }
 
--- a/src/globalvars.cpp
+++ b/src/globalvars.cpp
@@ -13,9 +13,47 @@
 
 namespace bpftrace::globalvars {
 
-void update_global_vars(const struct bpf_object *bpf_object,
-                        struct bpf_map *global_vars_map,
-                        BPFtrace &bpftrace)
+const GlobalVarConfig &get_config(GlobalVar global_var)
+{
+  auto it = GLOBAL_VAR_CONFIGS.find(global_var);
+  if (it == GLOBAL_VAR_CONFIGS.end()) {
+    LOG(BUG) << "Global variable enum not found in GLOBAL_VAR_CONFIGS";
+  }
+  return it->second;
+}
+
+static void verify_maps_found(
+    const std::unordered_map<std::string, struct bpf_map *>
+        &section_name_to_global_vars_map,
+    const BPFtrace &bpftrace)
+{
+  for (const auto global_var : bpftrace.resources.needed_global_vars) {
+    auto config = get_config(global_var);
+    if (!section_name_to_global_vars_map.count(config.section)) {
+      LOG(BUG) << "No map found for " << config.section
+               << " which is needed to set global variable " << config.name;
+    }
+  }
+}
+
+static std::unordered_set<GlobalVar> get_global_vars_for_section(
+    std::string_view target_section,
+    const BPFtrace &bpftrace)
+{
+  std::unordered_set<GlobalVar> ret;
+  for (const auto global_var : bpftrace.resources.needed_global_vars) {
+    auto config = get_config(global_var);
+    if (config.section == target_section) {
+      ret.insert(global_var);
+    }
+  }
+  return ret;
+}
+
+static std::map<GlobalVar, int> find_btf_var_offsets(
+    const struct bpf_object *bpf_object,
+    std::string_view section_name,
+    const std::unordered_set<GlobalVar> &needed_global_vars)
 {
   struct btf *self_btf = bpf_object__btf(bpf_object);
 
@@ -24,27 +62,23 @@ void update_global_vars(const struct bpf_object *bpf_object,
   }
 
   __s32 section_id = btf__find_by_name(self_btf,
-                                       std::string(SECTION_NAME).c_str());
+                                       std::string(section_name).c_str());
   if (section_id < 0) {
-    LOG(BUG) << "Failed to find section " << SECTION_NAME
+    LOG(BUG) << "Failed to find section " << section_name
              << " to update global vars";
   }
 
   const struct btf_type *section_type = btf__type_by_id(
       self_btf, static_cast<__u32>(section_id));
   if (!section_type) {
-    LOG(BUG) << "Failed to get BTF type for section " << SECTION_NAME;
+    LOG(BUG) << "Failed to get BTF type for section " << section_name;
   }
 
   // First locate the offsets of each global variable in the section with btf
-  std::map<std::string_view, int> vars_and_offsets;
+  std::map<GlobalVar, int> vars_and_offsets;
 
-  for (const auto &name : GLOBAL_VAR_NAMES) {
-    if (bpftrace.resources.needed_global_vars.find(name) ==
-        bpftrace.resources.needed_global_vars.end()) {
-      continue;
-    }
-    vars_and_offsets[name] = -1;
+  for (const auto global_var : needed_global_vars) {
+    vars_and_offsets[global_var] = -1;
   }
 
   int i;
@@ -59,13 +93,29 @@ void update_global_vars(const struct bpf_object *bpf_object,
     }
 
     std::string_view name = btf__name_by_offset(self_btf, type_id->name_off);
+    vars_and_offsets[from_string(name)] = member->offset;
+  }
 
-    if (vars_and_offsets.find(name) != vars_and_offsets.end()) {
-      vars_and_offsets[name] = member->offset;
-    } else {
-      LOG(BUG) << "Unknown global variable " << name;
+  for (const auto [global_var, offset] : vars_and_offsets) {
+    if (offset < 0) {
+      LOG(BUG) << "Global variable " << to_string(global_var)
+               << " has not been added to the BPF code "
+                  "(codegen_llvm)";
     }
   }
+  return vars_and_offsets;
+}
+
+static void update_global_vars_rodata(
+    const struct bpf_object *bpf_object,
+    std::string_view section_name,
+    struct bpf_map *global_vars_map,
+    const std::unordered_set<GlobalVar> &needed_global_vars,
+    const BPFtrace &bpftrace)
+{
+  auto vars_and_offsets = find_btf_var_offsets(bpf_object,
+                                               section_name,
+                                               needed_global_vars);
 
   size_t v_size;
   char *global_vars_buf = reinterpret_cast<char *>(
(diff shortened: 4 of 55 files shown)


src/globalvars.cpp
```diff
@@ -90,4 +139,78 @@
...
+  }
+}
+
+uint32_t get_custom_rw_section_desired_size(std::string_view section_name,
+                                            BPFtrace &bpftrace)
+{
+  if (section_name == FMT_STRINGS_BUFFER_SECTION_NAME) {
+    return bpftrace.ncpus_ * bpftrace.resources.max_fmtstring_args_size;
+  }
+  LOG(BUG) << "Unknown section " << section_name;
```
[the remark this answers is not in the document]
ryantimwilson: Fixed in latest version

jordalgo:
@ryantimwilson Not sure if this is specific to my machine but seeing a verifier error for the following program:
```
i:ms:1 { printf("hello %s %d", "a", 123); }
```
Error:
```
AST node count: 8
Attaching 1 probe...
ERROR: Error loading BPF program for interval_ms_1_1.
Kernel error log:
reg type unsupported for arg#0 function interval_ms_1_1#22
0: R1=ctx(off=0,imm=0) R10=fp0
;
0: (85) call bpf_get_smp_processor_id#8       ; R0_w=scalar()
1: (27) r0 *= 24                      ; R0_w=scalar(smax=9223372036854775800,umax=18446744073709551608,var_off=(0x0; 0xfffffffffffffff8),s32_max=2147483640,u32_max=-8)
2: (18) r2 = 0xffffc90020726000       ; R2_w=map_value(off=0,ks=4,vs=1920,imm=0)
4: (0f) r2 += r0
math between map_value pointer and register with unbounded min value is not allowed


```

I'm on a 5.19 kernel.

ryantimwilson:
@jordalgo argh, I have 6.9 kernel and verifier is smart enough for this to pass there.

@danobi mentioned that older kernels could have this issue. I will add instructions to verify 0 <= mult and downgrade kernel to 5.2 to verify

Also, I will do some refactoring of globalvars module...configuration is a bit all over the place now

danobi:
> AOT does not seem backwards compatible as I needed to add max_fmtstring_args_size to AOT serialization. Is this okay? If not, we can use scratch map around if max_fmtstring_args_size == 0

AOT is design such that it's always ok to add/remove/change any of the serialized fields. The AOT shim binary is shipped alongside main bpftrace binary. They are expected to be built from the same source.

src/ast/irbuilderbpf.cpp
```diff
@@ -473,15 +473,20 @@ CallInst *IRBuilderBPF::CreateGetStrScratchMap(int idx,
...
-{
-  return createGetScratchMap(to_string(MapType::FmtStringArgs),
-                             "fmtstr",
-                             GET_PTR_TY(),
-                             loc,
-                             failure_callback);
+Value *IRBuilderBPF::CreateGetFmtStringArgsScratchBuffer(const location &loc)
+{
+  // Get CPU ID
+  auto cpu_id = CreateGetCpuId(loc);
```
danobi, line 479: Just calling out that helper call here is fine. Verifier will inline the call so at runtime function call overhead is avoided.

Also, when we add the masking (to prove to older verifiers we are inbounds), we might wanna consider loading value into a register and doing the masking once at the beginning of the probe. That way we avoid re-doing the same work over and over if there's a lot of scratch map lookups.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3771,28 +3771,41 @@ void CodegenLLVM::generate_maps(const RequiredResources &required_resources,
...
+      var->setConstant(false);
+      var->setSection(bpftrace::globalvars::FMT_STRINGS_BUFFER_SECTION_NAME);
+      var->setExternallyInitialized(true);
+      var->setDSOLocal(true);
+      var->addDebugInfo(
+          debug_.createGlobalVariable(globalvars::FMT_STRINGS_BUFFER,
+                                      array_type));
+    } else {
+      LOG(BUG) << "Unknown global variable: " << name;
+    }
```
danobi, line 3808: Nit: maybe time for an enum? That way if we switch/case all the variants (leaving out default), compiler will warn if we forget to account for a new variant

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3771,28 +3771,41 @@ void CodegenLLVM::generate_maps(const RequiredResources &required_resources,
...
+      var->setExternallyInitialized(true);
+      var->setDSOLocal(true);
+      var->addDebugInfo(debug_.createGlobalInt64(name));
+    } else if (name == globalvars::FMT_STRINGS_BUFFER) {
+      assert(resources.max_fmtstring_args_size > 0);
+      // Assume 1 CPU to ensure codegen tests pass across multiple machines and
+      // so LLVM doesn't incorrectly optimize code (which happens if array size
+      // = 1)
+      auto array_type = CreateArray(resources.max_fmtstring_args_size,
+                                    CreateInt8());
```
danobi, line 3794: Seems reasonable. Typically how we do it in other places is there's some member function in a struct that returns the value. And for codegen we'll override the implementation.

But probably simpler to do it like this.

src/globalvars.cpp
```diff
@@ -90,4 +139,77 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+    const std::unordered_set<std::string> &needed_global_vars,
+    BPFtrace &bpftrace)
+{
+  auto desired_size = get_custom_rw_section_desired_size(section_name,
+                                                         bpftrace);
+  auto err = bpf_map__set_value_size(global_vars_map, desired_size);
+  if (err != 0) {
+    LOG(BUG) << "Failed to set size to " << desired_size << " for section "
+             << section_name;
+  }
```
danobi, line 167: Not necessarily a bug, right? Maybe libbpf or kernel thinks it's too big. Probably want to throw `FatalUserException`: https://github.com/bpftrace/bpftrace/blob/master/docs/coding_guidelines.md#error-handling-and-exceptions

src/globalvars.cpp
```diff
@@ -90,4 +139,77 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+
+  size_t actual_size;
+  auto buf = bpf_map__initial_value(global_vars_map, &actual_size);
+  if (!buf) {
+    LOG(BUG) << "Failed to get size for section " << section_name;
+  }
+  if (actual_size != desired_size) {
+    LOG(BUG) << "Failed to set size from " << actual_size << " to "
+             << desired_size << " for section " << section_name;
+  }
```
danobi, line 177: Same as above

danobi (commented):
nit: please place description/explanation into commit message. We'd like git-blame to work well offline (ie without going to GH)

overall, codegen bits lgtm. global vars stuff is starting to get messy, though. Would appreciate a good refactor

src/ast/passes/codegen_llvm.cpp
ryantimwilson: agreed, im gonna refactor globalvars stuff into enum and more configuration-driven

src/ast/passes/codegen_llvm.cpp
ryantimwilson: Ooh that's a good idea, that's a little cleaner, I'll try that if its easy enough

src/ast/irbuilderbpf.cpp
ryantimwilson: Will try that!

Masking didn't work as array size isn't always guaranteed to be a power of 2 - 1. I did what was suggested in the wiki and this worked - tested with 1000 printfs:
```
  auto max = getInt64(bpftrace_.resources.max_fmtstring_args_size * (bpftrace_.ncpus_ - 1));
  // check whether expr_ is less-than-or-equal-to maximum
  auto cmp = CreateICmp(CmpInst::ICMP_ULE, mult, max, "globalvar.min.cmp");
  // Select will contain expr_ if expr_ is sufficiently low, otherwise it will contain max
  auto select = CreateSelect(cmp, mult, max, "globalvar.min.select");
```

ryantimwilson: Hmmm so I tried two things:
1. Bounding CPU ID between 0 and maxCpuId, then multiplying by max_fmt_args_size
2. Multiplying CPU ID by max_fmt_arg_size, then applying bound between 0 and maxCpuId*max_fmt_args_size

#1 didn't work on 5.2, verifier still barfed. #2 did.

Problem is for #2, we'll need to store a value per scratch map as each array is different sizes. With 11 registers and ~5 scratch maps, that takes up a good amount of registers.

For now, I'll leave it un-cached but happy to discuss further.

src/ast/passes/codegen_llvm.cpp
ryantimwilson: Sorry I ended up not trying out (spent a lot of time refactoring globalvars), can you send me a code ref if you have one?

src/ast/irbuilderbpf.cpp
ryantimwilson: I was unable to get caching into registers to work, I kept hitting exceptions when loading the BPF program. For now, I'm not gonna cache it, but happy to discuss offline if I'm missing something.

src/globalvars.cpp
```diff
@@ -77,17 +134,123 @@
...
+static SizedType get_type(GlobalVar global_var,
+                          const RequiredResources &resources)
+{
+  switch (global_var) {
+    case GlobalVar::NUM_CPUS:
+      return CreateInt64();
+    case GlobalVar::FMT_STRINGS_BUFFER:
+      return CreateArray(resources.max_fmtstring_args_size, CreateInt8());
+  }
+  LOG(BUG) << "Unknown global variable enum";
```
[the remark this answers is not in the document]
ryantimwilson: Added fake return to make linter happy

src/globalvars.cpp
```diff
@@ -77,17 +134,123 @@
...
+  return get_metadata(global_var).name;
+}
+
+GlobalVar from_string(std::string_view name)
+{
+  for (const auto &[global_var, metadata] : GLOBAL_VAR_METADATA) {
+    if (metadata.name == name)
+      return global_var;
+  }
+  LOG(BUG) << "Unknown global variable " << name;
```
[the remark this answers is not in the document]
ryantimwilson: Added fake return to make linter happy

ryantimwilson:
Recent changes:
1. Refactoring globalvars to centralize configuration, use enums, etc. Note I didn't migrate `resources.needed_global_vars` from string -> enum to still allow for AOT backwards compatibility
2. Move from using ncpus -> max cpu ID for global var sizing since its possible the CPU ID is higher than the number of CPUs enabled on the computer (e.g. imagine a host with 64 CPUs where only CPU ID = 63 is available. Then ncpus = 1 but max_cpu_id = 63). As a result, needed to add a new read-only global variable MAX_CPU_ID similar to NUM_CPUS
3. Bound CPU ID between 0 and MAX_CPU_ID using select in LLVM IR. This allows us to pass BPF verifier on older 5.2 kernels
4. Move to FatalUserException for BPF global var resizing

src/ast/dibuilderbpf.cpp
```diff
@@ -292,9 +292,16 @@ DIGlobalVariableExpression *DIBuilderBPF::createMapEntry(
 
 DIGlobalVariableExpression *DIBuilderBPF::createGlobalInt64(
```
jordalgo, line 293: Nit: we probably don't need this wrapper function now. It's easy enough to write `createGlobalVariable(name, CreateInt64());`

jordalgo:
> Note I didn't migrate resources.needed_global_vars from string -> enum to still allow for AOT backwards compatibility

Do you mean AOT for older bpftrace versions? If so, I don't think you need to worry about that as AOT is still very experimental/use-at-your-own-risk (but I'll let @danobi decide).

src/globalvars.cpp
```diff
@@ -77,17 +134,130 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+{
+  switch (global_var) {
+    case GlobalVar::NUM_CPUS:
+      return CreateInt64();
+    case GlobalVar::MAX_CPU_ID:
+      return CreateInt64();
+    case GlobalVar::FMT_STRINGS_BUFFER:
+      return CreateArray(resources.max_fmtstring_args_size, CreateInt8());
+  }
+  return {}; // unreachable
```
jordalgo, line 233: It's annoying that the linter can't figure this out.

src/ast/irbuilderbpf.cpp
```diff
@@ -473,15 +473,27 @@ CallInst *IRBuilderBPF::CreateGetStrScratchMap(int idx,
...
+  auto config = globalvars::get_config(
+      bpftrace::globalvars::GlobalVar::FMT_STRINGS_BUFFER, bpftrace_.resources);
+
+  // Get CPU ID
+  auto cpu_id = CreateGetCpuId(loc);
+  auto max = CreateLoad(getInt64Ty(),
+                        module_.getGlobalVariable(to_string(
+                            bpftrace::globalvars::GlobalVar::MAX_CPU_ID)));
+  auto cmp = CreateICmp(CmpInst::ICMP_ULE, cpu_id, max, "cpuid.min.cmp");
+  auto select = CreateSelect(cmp, cpu_id, max, "cpuid.min.select");
```
jordalgo, line 487: Maybe add a small comment about why this comparison is necessary

src/globalvars.cpp
```diff
@@ -77,17 +134,130 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+    const std::unordered_set<std::string> &needed_global_vars,
+    BPFtrace &bpftrace)
+{
+  if (needed_global_vars.size() > 1) {
+    LOG(BUG) << "Multiple read-write global variables are in same section "
+             << section_name;
+  }
+  auto name = *needed_global_vars.begin();
+  auto config = get_config(name, bpftrace.resources);
+  auto desired_size = (bpftrace.max_cpu_id_ + 1) * config.type.GetSize();
```
jordalgo, line 161: Just throwing this out there: from your example if there were 64 cpus but only one is available/online #63 - then we're allocating 63 additional slots for cpus that we'll never hit. Maybe we should only allocated based on num cpus and the create a way to map the CPU id to an open slot index.

src/utils.cpp
```diff
@@ -507,6 +507,13 @@ std::vector<int> get_possible_cpus()
   return read_cpu_range("/sys/devices/system/cpu/possible");
 }
 
+int get_max_cpu_id()
+{
+  const auto cpus = get_possible_cpus();
+  auto it = std::max_element(cpus.begin(), cpus.end());
+  return it != cpus.end() ? *it : 0;
```
jordalgo, line 514: Is this check necessary, if we can't get any possible cpus then I don't know how bpftrace itself is running :)

ryantimwilson:
> Do you mean AOT for older bpftrace versions? If so, I don't think you need to worry about that as AOT is still very experimental/use-at-your-own-risk (but I'll let @danobi decide).

Ah I didn't know that! Yea I was concerned about building AOT with bpftrace version N-1, then running with bpftrace N (likely a common pattern if this was widely used).

But if experimental, then yea no need to over-index on that. We can convert needed_global_vars to enum. Will make sure to add comment that enum vals should be stable (aka we shouldn't reorder them between versions)

src/globalvars.cpp
ryantimwilson: > Just throwing this out there: from your example if there were 64 cpus but only one is available/online https://github.com/bpftrace/bpftrace/issues/63 - then we're allocating 63 additional slots for cpus that we'll never hit.

Agreed, that is annoying.

> Maybe we should only allocate based on num cpus and the create a way to map the CPU id to an open slot index.

Ah I see, we could allocate a global uint8 array of size N = max_cpu_id + 1, then in each index, we have the open slot index. So in that example, we map 0-62 -> -1 (invalid) and 63 -> 0.

That could work! It certainly avoids the memory over-allocation but involves another array lookup and likely a CreateSelect to bound the result.

Do you know if per-CPU arrays in BPF already allocate memory for each possible CPU or only the online ones? If the former, then this is equivalent for what we have now and I would prefer to do it in another PR. According to this comment thread, it is the former: https://github.com/iovisor/bcc/issues/893. Happy to discuss of course!

src/utils.cpp
ryantimwilson: Yea not really :/ We could probably just throw a FatalUserException instead

src/ast/passes/codegen_llvm.cpp
danobi: Oh, that's not what I meant but also a good idea. I was more referring to ensuring the codegen unit tests are stable. ie we wouldn't want the test output to change depending on how many CPUs your laptop has.

Here you're essentially setting a random (but stable) size in codegen. Which ensures the `.ll` output is stable. At runtime it'll be resized to the real size. Which I think is probably the simplest way to do it.

In other places we use mocking to ensure stability. Eg: https://github.com/bpftrace/bpftrace/blob/3cd6d50e3597dfd826583453c318c39002044572/tests/codegen/array_integer_equal_comparison.cpp#L20-L30

But perhaps overkill for this

danobi:
> > Do you mean AOT for older bpftrace versions? If so, I don't think you need to worry about that as AOT is still very experimental/use-at-your-own-risk (but I'll let @danobi decide).
> 
> Ah I didn't know that! Yea I was concerned about building AOT with bpftrace version N-1, then running with bpftrace N (likely a common pattern if this was widely used).
> 
> But if experimental, then yea no need to over-index on that. We can convert needed_global_vars to enum. Will make sure to add comment that enum vals should be stable (aka we shouldn't reorder them between versions)

Both not possible and not supported. AOT is designed where there is never any compatibility concerns.

src/utils.cpp
jordalgo: I would argue you can just blindly do the dereference. We could add something more friendly if we get reports that bpftrace is actually dying here.

src/ast/passes/codegen_llvm.cpp
ryantimwilson: Ah I see, agreed, this approach is simpler so I'll go with this for now!

src/ast/dibuilderbpf.cpp
ryantimwilson: Fixed in latest PR update (coming soon)

src/utils.cpp
ryantimwilson: Fixed in latest PR update (coming soon)

src/ast/irbuilderbpf.cpp
ryantimwilson: Fixed in latest PR update (coming soon)

ryantimwilson:
Updates:
1. Migrate needed_global_vars from string -> enum
2. Make createScratchBuffer (e.g. support > 1 elements) sufficiently generic to make migrating more global vars easier
3. Remove need for max_fmt_strings_size or anything saved in AOT ELF. Now we simply get .data.* section, get initial size, multiply initial size by MAX_CPU_ID

@danobi @jordalgo I think this PR is probably ready to go as a first step. Let me know what you think. Only thing I didn't implement is the optimizations mentioned above, would prefer to do those in other PRs if you don't mind (see below).

Next PRs:
1. Update all current existing scratch maps to use global vars
2. Migrate tuple + map key + map val from stack allocation to use global var scratch buffer (finally solve big string problem!)
3. Optimization to cache bounded CPU ID in register at beginning of probe
4. Optimization to allocate global var memory only for online CPUs

ryantimwilson:
Looks like 2 builds failed due to infrastructure errors, how do I retry them?

src/globalvars.h
```diff
@@ -4,25 +4,42 @@
...
+  { GlobalVar::NUM_CPUS, { "num_cpus", std::string(RO_SECTION_NAME), true } },
+  { GlobalVar::MAX_CPU_ID,
+    { "max_cpu_id", std::string(RO_SECTION_NAME), true } },
+  { GlobalVar::FMT_STRINGS_BUFFER,
+    { "fmt_str_buf", std::string(FMT_STRINGS_BUFFER_SECTION_NAME), false } },
+};
+
+void update_global_vars(
+    const struct bpf_object *obj,
+    std::unordered_map<std::string, struct bpf_map *> global_vars_map,
```
jordalgo, line 38:
```suggestion
    const std::unordered_map<std::string, struct bpf_map *>
                            &section_name_to_global_vars_map,
```

src/globalvars.cpp
```diff
@@ -76,18 +125,137 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+  }
+}
+
+void update_global_vars(const struct bpf_object *bpf_object,
+                        std::unordered_map<std::string, struct bpf_map *>
+                            section_name_to_global_vars_map,
+                        BPFtrace &bpftrace)
+{
+  verify_maps_found(section_name_to_global_vars_map, bpftrace);
+  for (auto &[section_name, global_vars_map] :
```
jordalgo, line 200:
```suggestion
  for (const auto &[section_name, global_vars_map] :
```

src/globalvars.cpp
```diff
@@ -13,9 +13,46 @@
...
+{
+  auto it = GLOBAL_VAR_CONFIGS.find(global_var);
+  if (it == GLOBAL_VAR_CONFIGS.end()) {
+    LOG(BUG) << "Unknown global variable " << to_string(global_var);
+  }
+  return it->second;
+}
+
+static void verify_maps_found(std::unordered_map<std::string, struct bpf_map *>
+                                  section_name_to_global_vars_map,
```
jordalgo, line 26:
```suggestion
static void verify_maps_found(const std::unordered_map<std::string, struct bpf_map *>
                                  &section_name_to_global_vars_map,
```

src/globalvars.cpp
```diff
@@ -13,9 +13,46 @@
...
+                              BPFtrace &bpftrace)
```
jordalgo, line 27:
```suggestion
                              const BPFtrace &bpftrace)
```

src/globalvars.cpp
```diff
@@ -13,9 +13,46 @@
...
+{
+  for (const auto global_var : bpftrace.resources.needed_global_vars) {
```
jordalgo, line 29:
```suggestion
  for (const auto &global_var : bpftrace.resources.needed_global_vars) {
```

src/globalvars.cpp
```diff
@@ -13,9 +13,46 @@
...
+    auto config = get_config(global_var);
+    if (!section_name_to_global_vars_map.count(config.section)) {
+      LOG(BUG) << "No map found for " << config.section
+               << " which is needed to set global variable " << config.name;
+    }
+  }
+}
+
+static std::unordered_set<GlobalVar> get_needed_global_variable_for_section(
```
jordalgo, line 38: naming nit:
```suggestion
static std::unordered_set<GlobalVar> get_global_vars_for_section(
```

src/globalvars.cpp
```diff
@@ -76,18 +125,137 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
-    }
-
+  for (const auto [global_var, offset] : vars_and_offsets) {
     int64_t *var = reinterpret_cast<int64_t *>(global_vars_buf + offset);
 
-    if (name == NUM_CPUS) {
+    if (global_var == GlobalVar::NUM_CPUS) {
       *var = bpftrace.ncpus_;
+    } else if (global_var == GlobalVar::MAX_CPU_ID) {
+      *var = bpftrace.max_cpu_id_;
```
jordalgo, line 134: maybe for a follow-up PR but I feel like we should have a map in `bpftrace` of globalVar to these values instead of them being individual member vars.

src/globalvars.cpp
```diff
@@ -59,13 +92,29 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
   }
+  return vars_and_offsets;
+}
+
+static void update_global_vars_rodata(
+    const struct bpf_object *bpf_object,
+    std::string_view section_name,
+    struct bpf_map *global_vars_map,
+    const std::unordered_set<GlobalVar> &needed_global_vars,
+    BPFtrace &bpftrace)
```
jordalgo, line 113:
```suggestion
    const BPFtrace &bpftrace)
```

src/globalvars.cpp
```diff
@@ -76,18 +125,137 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+    }
+  }
+}
+
+static void update_global_vars_custom_rw_section(
+    const struct bpf_object *bpf_object,
+    const std::string &section_name,
+    struct bpf_map *global_vars_map,
+    const std::unordered_set<GlobalVar> &needed_global_vars,
+    BPFtrace &bpftrace)
```
jordalgo, line 144:
```suggestion
    const BPFtrace &bpftrace)
```

src/globalvars.cpp
```diff
@@ -76,18 +125,137 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+                                       " for section " + section_name);
+  }
+
+  buf = bpf_map__initial_value(global_vars_map, &actual_size);
+  if (!buf) {
+    LOG(BUG) << "Failed to get size for section " << section_name
+             << " after resizing";
+  }
+  if (actual_size != desired_size) {
+    throw bpftrace::FatalUserException(
```
jordalgo, line 176: Remind me, why is this a user error instead of a bug?

src/types.h
```diff
@@ -703,6 +703,16 @@ uint64_t asyncactionint(AsyncAction a);
 
 enum class PositionalParameterType { positional, count };
 
+namespace globalvars {
```
jordalgo, line 706: nit: do we need this additional namespace?

src/globalvars.cpp
```diff
@@ -13,9 +13,46 @@
...
 namespace bpftrace::globalvars {
 
-void update_global_vars(const struct bpf_object *bpf_object,
-                        struct bpf_map *global_vars_map,
-                        BPFtrace &bpftrace)
+const GlobalVarConfig &get_config(GlobalVar global_var)
+{
+  auto it = GLOBAL_VAR_CONFIGS.find(global_var);
+  if (it == GLOBAL_VAR_CONFIGS.end()) {
+    LOG(BUG) << "Unknown global variable " << to_string(global_var);
```
jordalgo, line 20: `to_string` calls `get_config` right so this wouldn't work

jordalgo (commented):
Overall looking good - just a few small issues and nits.

src/types.h
```diff
@@ -703,6 +703,16 @@ uint64_t asyncactionint(AsyncAction a);
...
+
+enum class GlobalVar {
+  NUM_CPUS,
+  MAX_CPU_ID,
+  FMT_STRINGS_BUFFER,
+};
```
danobi, line 712: This seems like a good place to leave some comments about purpose of each var

src/globalvars.cpp
```diff
@@ -76,18 +125,137 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+  }
+  LOG(BUG) << "Unknown global variable " << name;
+  return {}; // unreachable
+}
+
+static SizedType make_rw_type(size_t num_elements,
+                              const SizedType &element_type)
+{
+  auto subtype = CreateArray(num_elements, element_type);
+  // For 1 CPU, will be adjusted at runtime
```
danobi, line 242:
```suggestion
  // For 1 CPU, will be adjusted to actual CPU count at runtime
```

danobi (commented):
some nits; lgtm otherwise

src/globalvars.cpp
ryantimwilson: yup, will fix, thanks!

src/types.h
ryantimwilson: Well...all the other globalvar stuff is in bpftrace::globalvars namespace in globalvars.h so feels nice to be consistent. We only need to move the enum here to avoid a circular dependency between resources.h <-> globalvars.h

With that said, I don't feel strongly

src/globalvars.cpp
ryantimwilson: You could argue it both ways:

User's fault because they don't have enough memory or on a computer with too many CPUs

vs. 

Our bug because we designed global var system

I don't feel strongly, I can see why this could be our bug but @danobi suggested a user facing error

src/globalvars.cpp
ryantimwilson: Agreed, the only tough part is read-write vs. read-only needs to be handled differently.

For how, let me change this to a switch statement for clarity as well

ryantimwilson:
@danobi @jordalgo thanks for all the comments, addressed them in recent revision

> Didn't fully follow discussion above b/c strapped for time this week, but CPUs can be dynamically onlined while a script is running. So probably better to always allocate num possible to account for that case?

Oh good call. Then yea doing that optimization isn't a good idea. cc @jordalgo

src/globalvars.cpp
danobi: IMO anything that can fail which has an outside element should be handled gracefully. BUG (ie coredump) should be things firmly inside our control.

danobi (approved):

src/globalvars.cpp
```diff
@@ -76,18 +126,151 @@ void update_global_vars(const struct bpf_object *bpf_object,
...
+SizedType get_type(bpftrace::globalvars::GlobalVar global_var,
+                   const RequiredResources &resources)
+{
+  switch (global_var) {
+    case bpftrace::globalvars::GlobalVar::NUM_CPUS:
+    case bpftrace::globalvars::GlobalVar::MAX_CPU_ID:
+      return CreateInt64();
+    case bpftrace::globalvars::GlobalVar::FMT_STRINGS_BUFFER:
+      assert(resources.max_fmtstring_args_size > 0);
+      return make_rw_type(
```
jordalgo, line 261: Question: so in future PRs, there will be other things calling `make_rw_type` with more than 1 num_elements?

ryantimwilson: Yup!

2 use cases:
1. String scratch map already in code: https://github.com/bpftrace/bpftrace/blob/f6d6345934d92be18de3fc79a16d5163f9d93f6a/src/ast/passes/codegen_llvm.cpp#L3782
2. Tuple scratch map from PR (aka size = max tuple depth): https://github.com/bpftrace/bpftrace/pull/3456

jordalgo (approved):

PR #3461, opened 2024-09-21
Feature: scratch variable declarations

jordalgo:
This commit implements scratch variable declarations,
which will be useful when fixing/adding block/lexical scoping.

As per this RFC/discussion: https://github.com/bpftrace/bpftrace/issues/3423
this new syntax is all valid:
```
let $x;
let $x: uint8;
let $x = 11;
let $x: uint8 = 7;
```

Additionally, there are some restrictions:
- variable declarations, if used, have to come before variable use
- variable declarations, if used, have to come before variable assignment
- the same variable name can't be declared twice
- for variable declarations that specify the type, this type
is not resizeable e.g. this is not valid
`let $a: str_t[5] = "hiya"; $a = "longerstr";` but this is
`let $a: str_t[5] = "hiya"; $a = "by";`
- declared variables must be assigned to

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -434,9 +434,10 @@ inside a script config block.
 
 === Data Types
 
-The following fundamental integer types are provided by the language.
-Integers are internally represented as 64 bit signed.
-If you need another representation, you may cast to the following built in types:
+The following fundamental types are provided by the language.
+Note: Integers are by default represented as 64 bit signed but that can be 
+changed by either casting them or, for scratch variables, explicitly specifying 
+the type upon declaration.
 
 [cols="~,~"]
 |===
@@ -939,6 +940,21 @@ bpftrace knows two types of variables, 'scratch' and 'map'.
 
 'scratch' variables are kept on the BPF stack and only exists during the execution of the action block and cannot be accessed outside of the program.
 Scratch variable names always start with a `$`, e.g. `$myvar`.
+
+'scratch' variables can also declared before or during initialization with `let` e.g.
+```
+let $x;
+let $y = 11;
+```
+
+If no assignment is specified variables will initialize to 0.
+
+You can also specify the type in the declaration e.g.
+```
+let $x: uint8;
+let $y: uint8 = 7;
+let $a: string = "hiya";
+```
 
 'map' variables use BPF 'maps'.
 These exist for the lifetime of `bpftrace` itself and can be accessed from all action blocks and user-space.
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -35,6 +35,7 @@ MAKE_ACCEPT(ExprStatement)
 MAKE_ACCEPT(AssignMapStatement)
 MAKE_ACCEPT(AssignVarStatement)
 MAKE_ACCEPT(AssignConfigVarStatement)
+MAKE_ACCEPT(VarDeclStatement)
 MAKE_ACCEPT(Predicate)
 MAKE_ACCEPT(AttachPoint)
 MAKE_ACCEPT(If)
@@ -201,12 +202,35 @@ AssignVarStatement::AssignVarStatement(Variable *var,
   expr->var = var;
 }
 
+AssignVarStatement::AssignVarStatement(VarDeclStatement *var_decl_stmt,
+                                       Expression *expr,
+                                       location loc)
+    : Statement(loc),
+      var_decl_stmt(var_decl_stmt),
+      var(var_decl_stmt->var),
+      expr(expr)
+{
+  expr->var = var;
+}
+
 AssignConfigVarStatement::AssignConfigVarStatement(
     const std::string &config_var,
     Expression *expr,
     location loc)
     : Statement(loc), config_var(config_var), expr(expr)
 {
+}
+
+VarDeclStatement::VarDeclStatement(Variable *var, SizedType type, location loc)
+    : Statement(loc), var(var), set_type(true)
+{
+  var->type = std::move(type);
+}
+
+VarDeclStatement::VarDeclStatement(Variable *var, location loc)
+    : Statement(loc), var(var)
+{
+  var->type = CreateNone();
 }
 
 Predicate::Predicate(Expression *expr, location loc) : Node(loc), expr(expr)
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2421,27 +2421,33 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
     b_.CreateLifetimeEnd(val);
 }
 
+void CodegenLLVM::maybeAllocVariable(const std::string &var_ident,
+                                     const SizedType &var_type)
+{
+  if (variables_.find(var_ident) != variables_.end()) {
+    return;
+  }
+  SizedType alloca_type = var_type;
+  // Arrays and structs need not to be copied when assigned to local variables
+  // since they are treated as read-only - it is sufficient to assign
+  // the pointer and do the memcpy/proberead later when necessary
+  if (var_type.IsArrayTy() || var_type.IsRecordTy()) {
+    auto &pointee_type = var_type.IsArrayTy() ? *var_type.GetElementTy()
+                                              : var_type;
+    alloca_type = CreatePointer(pointee_type, var_type.GetAS());
+  }
+
+  AllocaInst *val = b_.CreateAllocaBPFInit(alloca_type, var_ident);
+  variables_[var_ident] = VariableLLVM{ val, val->getAllocatedType() };
+}
+
 void CodegenLLVM::visit(AssignVarStatement &assignment)
 {
   Variable &var = *assignment.var;
 
   auto scoped_del = accept(assignment.expr);
 
-  if (variables_.find(var.ident) == variables_.end()) {
-    SizedType alloca_type = var.type;
-
-    // Arrays and structs need not to be copied when assigned to local variables
-    // since they are treated as read-only - it is sufficient to assign
-    // the pointer and do the memcpy/proberead later when necessary
-    if (var.type.IsArrayTy() || var.type.IsRecordTy()) {
-      auto &pointee_type = var.type.IsArrayTy() ? *var.type.GetElementTy()
-                                                : var.type;
-      alloca_type = CreatePointer(pointee_type, var.type.GetAS());
-    }
-
-    AllocaInst *val = b_.CreateAllocaBPFInit(alloca_type, var.ident);
-    variables_[var.ident] = VariableLLVM{ val, val->getAllocatedType() };
-  }
+  maybeAllocVariable(var.ident, var.type);
 
   if (var.type.IsArrayTy() || var.type.IsRecordTy()) {
     // For arrays and structs, only the pointer is stored
@@ -2465,6 +2471,16 @@ void CodegenLLVM::visit(AssignVarStatement &assignment)
   } else {
     b_.CreateStore(expr_, variables_[var.ident].value);
   }
+}
+
+void CodegenLLVM::visit(VarDeclStatement &decl)
+{
+  Variable &var = *decl.var;
+  if (var.type.IsNoneTy()) {
+    // unused and has no type
+    return;
+  }
+  maybeAllocVariable(var.ident, var.type);
 }
 
 void CodegenLLVM::visit(If &if_block)
--- a/src/ast/passes/printer.cpp
+++ b/src/ast/passes/printer.cpp
@@ -262,12 +262,20 @@ void Printer::visit(AssignMapStatement &assignment)
 void Printer::visit(AssignVarStatement &assignment)
 {
   std::string indent(depth_, ' ');
-  out_ << indent << "=" << std::endl;
 
-  ++depth_;
-  assignment.var->accept(*this);
-  assignment.expr->accept(*this);
-  --depth_;
+  if (assignment.var_decl_stmt) {
+    assignment.var_decl_stmt->accept(*this);
+    ++depth_;
+    assignment.expr->accept(*this);
+    --depth_;
+  } else {
+    out_ << indent << "=" << std::endl;
+
+    ++depth_;
+    assignment.var->accept(*this);
+    assignment.expr->accept(*this);
+    --depth_;
+  }
 }
 
 void Printer::visit(AssignConfigVarStatement &assignment)
@@ -279,6 +287,15 @@ void Printer::visit(AssignConfigVarStatement &assignment)
   std::string indentVar(depth_, ' ');
   out_ << indentVar << "config var: " << assignment.config_var << std::endl;
   assignment.expr->accept(*this);
+  --depth_;
+}
+
+void Printer::visit(VarDeclStatement &decl)
+{
+  std::string indent(depth_, ' ');
+  out_ << indent << "decl" << std::endl;
+  ++depth_;
+  decl.var->accept(*this);
   --depth_;
 }
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -83,6 +83,48 @@ static std::pair<int64_t, int64_t> getIntTypeRange(const SizedType &ty)
   }
 }
 
+// These are types which aren't valid for scratch variables
+// e.g. this is not valid `let $x: sum_t;`
+static bool IsValidVarDeclType(const SizedType &ty)
+{
+  switch (ty.GetTy()) {
+    case Type::avg_t:
+    case Type::count_t:
+    case Type::hist_t:
+    case Type::lhist_t:
+    case Type::max_t:
+    case Type::min_t:
+    case Type::stats_t:
+    case Type::sum_t:
+    case Type::stack_mode:
+    case Type::voidtype:
+    case Type::probe:
+      return false;
+    case Type::integer:
+    case Type::kstack_t:
+    case Type::ustack_t:
+    case Type::timestamp:
+    case Type::ksym_t:
+    case Type::usym_t:
+    case Type::inet:
+    case Type::username:
+    case Type::string:
+    case Type::buffer:
+    case Type::pointer:
+    case Type::array:
+    case Type::mac_address:
+    case Type::record:
+    case Type::tuple:
+    case Type::cgroup_path_t:
+    case Type::strerror_t:
+    case Type::none:
+    case Type::reference:
+    case Type::timestamp_mode:
+      return true;
+  }
+  return false; // unreachable
+}
+
 void SemanticAnalyser::visit(Integer &integer)
 {
   if (integer.is_negative) {
@@ -1665,9 +1707,15 @@ void SemanticAnalyser::visit(Map &map)
 
 void SemanticAnalyser::visit(Variable &var)
 {
-  auto search_val = variable_val_[scope_].find(var.ident);
-  if (search_val != variable_val_[scope_].end()) {
-    var.type = search_val->second;
+  if (auto search_val = variables_[scope_].find(var.ident);
+      search_val != variables_[scope_].end()) {
+    if (search_val->second.was_assigned) {
+      var.type = search_val->second.type;
+    } else {
+      LOG(ERROR, var.loc, err_)
+          << "Variable used before it was assigned: " << var.ident;
+      var.type = CreateNone();
+    }
   } else {
     LOG(ERROR, var.loc, err_)
         << "Undefined or undeclared variable: " << var.ident;
@@ -2302,7 +2350,7 @@ void SemanticAnalyser::visit(For &f)
 
   // Validate decl
   const auto &decl_name = f.decl->ident;
-  if (variable_val_[scope_].find(decl_name) != variable_val_[scope_].end()) {
+  if (variables_[scope_].find(decl_name) != variables_[scope_].end()) {
     LOG(ERROR, f.decl->loc, err_)
         << "Loop declaration shadows existing variable: " + decl_name;
   }
@@ -2343,7 +2391,7 @@ void SemanticAnalyser::visit(For &f)
   CollectNodes<Variable> vars_referenced;
   std::unordered_set<std::string> var_set;
   for (auto *stmt : f.stmts) {
-    const auto &live_vars = variable_val_[scope_];
+    const auto &live_vars = variables_[scope_];
     vars_referenced.run(*stmt, [&live_vars, &var_set](const auto &var) {
       if (live_vars.find(var.ident) == live_vars.end())
         return false;
@@ -2359,7 +2407,7 @@ void SemanticAnalyser::visit(For &f)
   // real thing (#3017).
   CollectNodes<Variable> new_vars;
   for (auto *stmt : f.stmts) {
-    const auto &live_vars = variable_val_[scope_];
+    const auto &live_vars = variables_[scope_];
     new_vars.run(*stmt, [&live_vars](const auto &var) {
       return live_vars.find(var.ident) == live_vars.end();
     });
@@ -2381,7 +2429,9 @@ void SemanticAnalyser::visit(For &f)
 
   f.decl->type = CreateTuple(bpftrace_.structs.AddTuple({ *mapkey, *mapval }));
 
-  variable_val_[scope_][decl_name] = f.decl->type;
+  variables_[scope_][decl_name] = { .type = f.decl->type,
+                                    .can_resize = true,
+                                    .was_assigned = true };
 
   loop_depth_++;
   accept_statements(f.stmts);
@@ -2402,11 +2452,11 @@ void SemanticAnalyser::visit(For &f)
   }
 
   // Decl variable is not valid beyond this for-loop
-  variable_val_[scope_].erase(decl_name);
+  variables_[scope_].erase(decl_name);
 
   // Variables declared in a for-loop are not valid beyond it
   for (const Variable &var : new_vars.nodes()) {
-    variable_val_[scope_].erase(var.ident);
+    variables_[scope_].erase(var.ident);
   }
 
   // Finally, create the context tuple now that all variables inside the loop
@@ -2763,23 +2813,27 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
 void SemanticAnalyser::visit(AssignVarStatement &assignment)
 {
   Visit(assignment.expr);
+  if (assignment.var_decl_stmt) {
+    Visit(assignment.var_decl_stmt);
+  }
 
   std::string var_ident = assignment.var->ident;
-  auto search = variable_val_[scope_].find(var_ident);
+  auto search = variables_[scope_].find(var_ident);
   auto &assignTy = assignment.expr->type;
 
-  if (search != variable_val_[scope_].end()) {
-    auto &storedTy = search->second;
+  if (search != variables_[scope_].end()) {
+    auto &storedTy = search->second.type;
+    bool type_mismatch_error = false;
     if (storedTy.IsNoneTy()) {
       storedTy = assignTy;
     } else if (!storedTy.IsSameType(assignTy)) {
-      LOG(ERROR, assignment.loc, err_)
-          << "Type mismatch for " << var_ident << ": "
-          << "trying to assign value of type '" << assignTy
-          << "' when variable already contains a value of type '" << storedTy
-          << "'";
+      type_mismatch_error = true;
     } else if (assignTy.IsStringTy()) {
-      update_string_size(storedTy, assignTy);
+      if (search->second.can_resize) {
+        update_string_size(storedTy, assignTy);
+      } else if (!assignTy.FitsInto(storedTy)) {
+        type_mismatch_error = true;
+      }
     } else if (storedTy.IsIntegerTy()) {
       if (assignment.expr->is_literal) {
         auto integer = static_cast<Integer *>(assignment.expr);
@@ -2788,12 +2842,7 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
           bool can_fit = false;
           if (integer->is_negative) {
             if (!storedTy.IsSigned()) {
-              LOG(ERROR, assignment.loc, err_)
-                  << "Type mismatch for " << var_ident << ": "
-                  << "trying to assign value of type '" << assignTy
-                  << "' when variable already contains a value of type '"
-                  << storedTy << "'";
-              return;
+              type_mismatch_error = true;
             } else {
               auto min_max = getIntTypeRange(storedTy);
               can_fit = value >= min_max.first;
@@ -2818,7 +2867,7 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
                 assignment.loc);
             Visit(cast);
             assignment.expr = cast;
-          } else {
+          } else if (!type_mismatch_error) {
             LOG(ERROR, assignment.loc, err_)
                 << "Type mismatch for " << var_ident << ": "
                 << "trying to assign value '"
@@ -2829,25 +2878,13 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
           }
         }
       } else if (storedTy.IsSigned() != assignTy.IsSigned()) {
-        LOG(ERROR, assignment.loc, err_)
-            << "Type mismatch for " << var_ident << ": "
-            << "trying to assign value of type '" << assignTy
-            << "' when variable already contains a value of type '" << storedTy
-            << "'";
+        type_mismatch_error = true;
       } else {
         if (!assignTy.FitsInto(storedTy)) {
           LOG(ERROR, assignment.loc, err_)
               << "Integer size mismatch. Assignment type '" << assignTy
               << "' is larger than the variable type '" << storedTy << "'.";
         }
-      }
-    } else if (assignTy.IsRecordTy()) {
-      if (assignTy.GetName() != storedTy.GetName()) {
-        LOG(ERROR, assignment.loc, err_)
-            << "Type mismatch for " << var_ident << ": "
-            << "trying to assign value of type '" << assignTy.GetName()
-            << "' when variable already contains a value of type '" << storedTy
-            << "'";
       }
     } else if (assignTy.IsBufferTy()) {
       auto var_size = storedTy.GetSize();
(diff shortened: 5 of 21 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -2827,6 +2844,46 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+    if (!areLocationsEqual(decl_search->second, decl.loc)) {
+      // Note this will change with proper block/lexical scoping
+      LOG(ERROR, decl.loc, err_)
+          << "Variable " << var_ident << " was already declared.";
+    }
+  }
+
+  auto search = variable_val_[scope_].find(var_ident);
+  bool var_found = search != variable_val_[scope_].end();
+  if (var_found) {
```
viktormalik, line 2865: Can be merged into a single line?
```suggestion
  if (search != variable_val_[scope_].end()) {
```

src/ast/passes/printer.cpp
```diff
@@ -262,6 +262,11 @@ void Printer::visit(AssignMapStatement &assignment)
 void Printer::visit(AssignVarStatement &assignment)
 {
   std::string indent(depth_, ' ');
+
+  if (assignment.var_decl_stmt) {
+    assignment.var_decl_stmt->accept(*this);
```
viktormalik, line 267: Just for the reference, this will print the variable twice - in the declaration and in the assignment:
```
# bpftrace -e 'BEGIN { let $x = 10 }' -d ast

AST after: parser
-------------------
Program
 BEGIN
  decl
   variable: $x
  =
   variable: $x
   int: 10
```
I'm personally ok with this, just highlighting it.

Perhaps we may want to add an AST printer test for that?

viktormalik (commented):
Just a couple of small remarks, otherwise looks nice and simple.

man/adoc/bpftrace.adoc
```diff
@@ -932,6 +932,14 @@ bpftrace knows two types of variables, 'scratch' and 'map'.
 'scratch' variables are kept on the BPF stack and only exists during the execution of the action block and cannot be accessed outside of the program.
 Scratch variable names always start with a `$`, e.g. `$myvar`.
 
+'scratch' variables can also declared before or during initialization with `let` e.g.
```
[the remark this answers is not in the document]
jordalgo: Good idea. Will add.

src/ast/passes/printer.cpp
jordalgo: > this will print the variable twice

Right, that was my intention to separate the declaration statement from the assignment.

> Perhaps we may want to add an AST printer test for that?

Good idea. Will do.

src/ast/passes/semantic_analyser.h
```diff
@@ -159,7 +160,13 @@ class SemanticAnalyser : public Visitor {
   // SemanticAnalyser.
   int func_arg_idx_ = -1;
 
-  std::map<Scope *, std::map<std::string, SizedType>> variable_val_;
+  struct variable_val {
+    SizedType type;
+    bool can_resize;
+  };
+
+  std::map<Scope *, std::map<std::string, variable_val>> variable_val_;
```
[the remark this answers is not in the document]
jordalgo: Ha yes. I'll put that in a separate commit on top of this one.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2803,6 +2820,44 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+
+  auto search = variable_type_[scope_].find(var_ident);
+  if (search != variable_type_[scope_].end()) {
+    if (!declared_variables_[scope_].contains(var_ident)) {
+      LOG(ERROR, decl.loc, err_)
+          << "Variable declarations need to occur before variable usage or "
+             "assignment. Variable: "
+          << var_ident;
+    } else {
+      decl.var->type = search->second.type;
```
viktormalik, line 2845: IIUC, this is to assign type to untyped declarations after we have visited the assignment. Shouldn't we check that `decl.var->type` is not already set to a different (incompatible) type? Or that cannot happen?

Anyways, a comment here would help a lot, it took me a while to figure out the purpose of this line.

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,59 @@ kfunc:foo { $p = path((uint8 *)0) }
...
 
+TEST(semantic_analyser, variable_declarations)
+{
+  test("BEGIN { let $a; $a = 1; }");
+  test("BEGIN { let $a: int16; $a = 1; }");
+  test("BEGIN { let $a = 1; }");
+  test("BEGIN { let $a: int16 = 1; }");
+  test("BEGIN { let $a: str_t[5] = \"hiya\"; }");
+  // If the type is specified it's strict in that future assignments
+  // need to fit into that type
```
viktormalik, line 4046: Could we have a positive case test for this for integers, too? Something like:
```
let $a: int16 = 1; $a = (int8)2;
```

viktormalik (commented):
Couple more remarks.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -79,6 +79,15 @@ static std::pair<int64_t, int64_t> getIntTypeRange(const SizedType &ty)
   }
 }
 
+static bool areLocationsEqual(const location &loc1, const location &loc2)
```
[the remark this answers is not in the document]
jordalgo: ohhh nice! Thanks!

src/ast/passes/semantic_analyser.h
```diff
@@ -161,7 +162,13 @@ class SemanticAnalyser : public Visitor {
   // SemanticAnalyser.
   int func_arg_idx_ = -1;
 
-  std::map<Scope *, std::map<std::string, SizedType>> variable_val_;
+  struct variable_val {
```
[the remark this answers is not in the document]
jordalgo: ooops. good catch.

src/ast/passes/semantic_analyser.cpp
jordalgo: > Shouldn't we check that decl.var->type is not already set to a different (incompatible) type? Or that cannot happen?

That can't happen because if we haven't seen this variable before (which we also validate in this function) then when the declaration happens we insert into the, newly renamed, `variable_type_` map. Then when var assignment  happens it will check against the type in that map and fail if they're not same (unless the store type is a None type, in which case it updates it). But yes, will add a comment :)

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2803,6 +2820,44 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+{
+  std::string var_ident = decl.var->ident;
+
+  // This should be the first time we're seeing this variable
+  auto decl_search = declared_variables_[scope_].find(var_ident);
+  if (decl_search != declared_variables_[scope_].end()) {
+    if (!areLocationsEqual(decl_search->second, decl.loc)) {
+      // Note this will change with proper block/lexical scoping
+      LOG(ERROR, decl.loc, err_)
+          << "Variable " << var_ident << " was already declared.";
```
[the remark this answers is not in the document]
jordalgo: I like this idea. Will add.

man/adoc/bpftrace.adoc
```diff
@@ -466,6 +467,33 @@ If you need another representation, you may cast to the following built in types
...
+|Kernel symbol. Returned by <<functions-ksym, `ksym`>>
+
+|macaddr_t
+|Macaddress. Returned by <<functions-macaddr, `macaddr`>>
+
+|strerror_t
+|String represenation of errno code. Returned by <<functions-strerror, `strerror`>>
+
+|str_t[]
+|String: Null terminated array of characters
```
ajor, line 490: I'm a bit wary of documenting these builtin types as I'm not sure that they're implemented correctly right now. Strings, for example, shouldn't be an array type - they're variable length - and I'm not sure that we should specify that they're null terminated as that's an implementation detail that we might want to change.

Also do `inet_t` and `buf_t` need to be array types?

src/ast/ast.cpp
```diff
@@ -209,6 +221,18 @@ AssignConfigVarStatement::AssignConfigVarStatement(
 {
 }
 
+VarDeclStatement::VarDeclStatement(Variable *var, SizedType type, location loc)
+    : Statement(loc), var(var), set_type(true)
+{
+  var->type = type;
```
ajor, line 227:
```suggestion
  var->type = std::move(type);
```

src/ast/passes/printer.cpp
ajor: I do think it looks nice to print out decl-assignments like this, but it's not a true representation of the AST's structure, which is what the AST printer is meant to show.

In this script we can't differentiate between the two different AST nodes by looking at the AST's debug output:
```
let $x = 1;
let $y;
$y = 1
```
```
  decl
   variable: $x :: [int64]
  =
   variable: $x :: [int64]
   int: 1 :: [int64]
  decl
   variable: $y :: [int64]
  =
   variable: $y :: [int64]
   int: 1 :: [int64]
```

For comparison, here's Clang's AST for a similar program:
```
int x = 1;
int y;
y = 1;
```
```
    |-DeclStmt <line:2:5, col:14>
    | `-VarDecl <col:5, col:13> col:9 x 'int' cinit
    |   `-IntegerLiteral <col:13> 'int' 1
    |-DeclStmt <line:3:5, col:10>
    | `-VarDecl <col:5, col:9> col:9 used y 'int'
    `-BinaryOperator <line:4:5, col:9> 'int' lvalue '='
      |-DeclRefExpr <col:5> 'int' lvalue Var 0xc824780 'y' 'int'
      `-IntegerLiteral <col:9> 'int' 1
```

src/parser.yy
```diff
@@ -411,6 +414,7 @@ expr_stmt:
  * this avoids a r/r conflict
  */
         |       assign_stmt        { $$ = $1; }
+        |       var_decl_stmt      { $$ = $1; }
```
ajor, line 417: Is this needed here? It's not really an expression statement and doesn't seem to cause conflicts if I delete it. Maybe it should be a new top-level statement type? i.e. in the `stmt_list` rule

man/adoc/bpftrace.adoc
viktormalik: I agree but we should document them somehow before we land this PR.

Another point - all all built-in types usable in a cast? If not, should we mention that in the docs?

It's starting to look like we may need a separate PR to clarify types documentation :grinning:

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2803,6 +2811,48 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
   Visit(assignment.expr);
 }
 
+void SemanticAnalyser::visit(VarDeclStatement &decl)
+{
+  std::string var_ident = decl.var->ident;
```
ajor, line 2816:
```suggestion
  const std::string &var_ident = decl.var->ident;
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2367,27 +2367,32 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
     b_.CreateLifetimeEnd(val);
 }
 
-void CodegenLLVM::visit(AssignVarStatement &assignment)
+void CodegenLLVM::updateVariableType(const std::string &var_ident,
```
ajor, line 2370: This function name's a bit confusing as it won't actually update the variable's type if the variable is already defined

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2367,27 +2367,32 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
...
+                                     const SizedType &var_type)
 {
-  Variable &var = *assignment.var;
-
-  auto scoped_del = accept(assignment.expr);
-
-  if (variables_.find(var.ident) == variables_.end()) {
-    SizedType alloca_type = var.type;
-
+  if (variables_.find(var_ident) == variables_.end()) {
```
ajor, line 2373: Nit: inverse the condition and return early to reduce nesting
```suggestion
  if (variables_.find(var_ident) != variables_.end())
    return
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2803,6 +2811,48 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+
+  // This should be the first time we're seeing this variable
+  auto decl_search = declared_variables_[scope_].find(var_ident);
+  if (decl_search != declared_variables_[scope_].end()) {
```
ajor, line 2820: C++17 baby! To keep the search variable scoped to where it's needed:
```suggestion
  if (auto decl_search = declared_variables_[scope_].find(var_ident);
      decl_search != declared_variables_[scope_].end()) {
```

(a lot of the old code was written pre-C++17 which is why it doesn't do this)

ajor (commented):
Love the test cases and error reporting - very clear!

src/ast/passes/printer.cpp
jordalgo: So perhaps this:
```
let $x = 1;
```
Would get changed to something like this:
```
decl
   variable: $y
   int: 1 :: [int64]
```

man/adoc/bpftrace.adoc
jordalgo: > Also do inet_t and buf_t need to be array types?

They do.

> all built-in types usable in a cast?

No. All the aggregate map types like `sum_t` and `max_t` are not allowed. However, it seems like I need to add an additional semantic_analyser check for those. So good question!

> It's starting to look like we may need a separate PR to clarify types documentation 😀

I feel like it doesn't make sense as a PR **before** this one - so maybe we just leave the adoc alone for now. I'll remove the additional types I added from the Data Types section. Then I'll put up a separate PR where we can discuss how we want to handle type documentation.

src/parser.yy
jordalgo: > It's not really an expression statement

You're right it doesn't really produce a value which seems to be the requirement for an expression. 

> Maybe it should be a new top-level statement type? i.e. in the stmt_list rule

Yeah, that works. I'll move it there.

src/ast/passes/semantic_analyser.cpp
jordalgo: See @ryantimwilson - I don't practice what I preach 😄

src/ast/passes/printer.cpp
ajor:
```
decl
 variable: $y
 int: 1 :: [int64]
```
I think this looks reasonable - I'd vote for this one

JakeHillion (commented):
Few comments, I think only the one about the additional test needs to go in with this, the others are about errors which could improve a UX a little but shouldn't be blockers.

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,69 @@ kfunc:foo { $p = path((uint8 *)0) }
...
+  test(R"(BEGIN { let $a: str_t[5] = "hiya"; $a = "bye"; })");
+  test(R"(BEGIN { let $a = "hiya"; $a = "longerstr"; })");
+  test("BEGIN { let $a: int16 = 1; $a = (int8)2; }");
+
+  test_error("BEGIN { let $a; }", R"(
+stdin:1:9-15: ERROR: Variable $a never assigned to.
+BEGIN { let $a; }
+        ~~~~~~
+)");
```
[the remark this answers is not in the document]
jordalgo: It's a good question. I went back and forth on it. I don't feel strongly other than I kinda like it when I get yelled at for doing silly things. That being said, it can be annoying if you're iterating fast and run into stuff like this. Maybe I will remove it - scripting languages are supposed to be flexible. Unless anyone else has strong feelings...

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,69 @@ kfunc:foo { $p = path((uint8 *)0) }
...
+stdin:1:26-33: ERROR: Type mismatch for $a: trying to assign value of type 'int64' when variable already contains a value of type 'uint16'
+BEGIN { let $a: uint16; $a = -1; }
+                         ~~~~~~~
+)");
+
+  test_error("BEGIN { let $a; $a = (uint8)1; $a = -1; }", R"(
+stdin:1:32-39: ERROR: Type mismatch for $a: trying to assign value of type 'int64' when variable already contains a value of type 'uint8'
+BEGIN { let $a; $a = (uint8)1; $a = -1; }
+                               ~~~~~~~
+)");
```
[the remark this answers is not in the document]
jordalgo: Yeah good call - though as you mentioned - maybe for a separate PR.

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,69 @@ kfunc:foo { $p = path((uint8 *)0) }
...
+stdin:1:9-15: ERROR: Initial declaration
+BEGIN { let $a; let $a; }
+        ~~~~~~
+)");
+
+  test_error("BEGIN { let $a: uint16; $a = -1; }", R"(
+stdin:1:26-33: ERROR: Type mismatch for $a: trying to assign value of type 'int64' when variable already contains a value of type 'uint16'
+BEGIN { let $a: uint16; $a = -1; }
+                         ~~~~~~~
+)");
```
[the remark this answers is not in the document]
jordalgo: Yeah, you're right. It should say something like "when variable already has a type 'uint16'". Let me see about improving this.

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,69 @@ kfunc:foo { $p = path((uint8 *)0) }
...
+  test(R"(BEGIN { let $a: str_t[5] = "hiya"; $a = "bye"; })");
+  test(R"(BEGIN { let $a = "hiya"; $a = "longerstr"; })");
+  test("BEGIN { let $a: int16 = 1; $a = (int8)2; }");
+
+  test_error("BEGIN { let $a; }", R"(
+stdin:1:9-15: ERROR: Variable $a never assigned to.
+BEGIN { let $a; }
+        ~~~~~~
+)");
+
```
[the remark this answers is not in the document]
jordalgo: Yes, I wanted to do this but it was more complicated than I thought. Let me look into it again because this is a good check to have.

tests/semantic_analyser.cpp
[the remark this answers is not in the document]
jordalgo: I made it a warning instead of an error.

src/ast/passes/semantic_analyser.h
```diff
@@ -161,7 +162,14 @@ class SemanticAnalyser : public Visitor {
...
 
-  std::map<Scope *, std::map<std::string, SizedType>> variable_val_;
+  struct variable_type {
+    SizedType type;
+    bool can_resize;
+    bool was_assigned;
+  };
+
+  std::map<Scope *, std::map<std::string, variable_type>> variable_type_;
+  std::map<Scope *, std::map<std::string, location>> declared_variables_;
```
viktormalik, line 172: Sorry about being such a nitpicker but with the new `was_assigned` field, `variable_type` does not make sense as the type name.

What about this:
```suggestion
  struct variable {
    SizedType type;
    bool can_resize;
    bool was_assigned;
  };

  std::map<Scope *, std::map<std::string, variable>> variables_;
  std::map<Scope *, std::map<std::string, location>> variable_decls_;
```
or, maybe, we could even add `location` to `struct variable` and have just one map (not sure if that would make sense)?

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,78 @@ kfunc:foo { $p = path((uint8 *)0) }
...
+  test_error(R"(BEGIN { let $a; print(($a)); $a = 1; })", R"(
+stdin:1:17-26: ERROR: Variable used before it was assigned: $a
+BEGIN { let $a; print(($a)); $a = 1; }
+                ~~~~~~~~~
+)");
+
+  test_error(R"(BEGIN { let $a: sum_t; })", R"(
+stdin:1:9-23: ERROR: Invalid variable declaration type: sum
```
viktormalik, line 4106: Shouldn't be
```suggestion
stdin:1:9-23: ERROR: Invalid variable declaration type: sum_t
```

tests/semantic_analyser.cpp
```diff
@@ -4035,4 +4035,78 @@ kfunc:foo { $p = path((uint8 *)0) }
...
+  test("BEGIN { let $a: int16 = 1; }");
+  test("BEGIN { let $a: str_t[5] = \"hiya\"; }");
+  // If the type is specified it's strict in that future assignments
+  // need to fit into that type
+  test(R"(BEGIN { let $a: str_t[5] = "hiya"; $a = "bye"; })");
+  test(R"(BEGIN { let $a = "hiya"; $a = "longerstr"; })");
+  test("BEGIN { let $a: int16 = 1; $a = (int8)2; }");
+  // Test more types
+  test("BEGIN { let $a: struct x; }");
+  test("BEGIN { let $a: struct x *; }");
```
viktormalik, line 4052: This caught my attention (only now, sorry!) as it can introduce a lot of tricky things.

When I tried it, though, I'm unable to get it past semantic analyser:
```
# bpftrace -e 'BEGIN { let $f: struct task_struct; }'
stdin:1:17-38: ERROR: syntax error, unexpected end of file, expecting identifier
BEGIN { let $f: struct task_struct; }
                ~~~~~~~~~~~~~~~~~~~~~

# bpftrace -e 'BEGIN { let $f: struct task_struct *; }'
stdin:1:17-40: ERROR: syntax error, unexpected end of file, expecting identifier
BEGIN { let $f: struct task_struct *; }
                ~~~~~~~~~~~~~~~~~~~~~~~
```
Is that the expected behaviour? And if so, how come that the tests pass?

jordalgo: @viktormalik Are you on an older version of this PR? This was the behavior before I updated the lexer.

tests/semantic_analyser.cpp
jordalgo: arrrgghhhh - the only time these types are available is in the parser. Our `typestr` function in types.cpp doesn't have any `_t` postfix for any types.

src/ast/passes/semantic_analyser.h
jordalgo: > Sorry about being such a nitpicker but with the new was_assigned field, variable_type does not make sense as the type name.

No worries! You're right, `variable` is a better name.

> or, maybe, we could even add location to struct variable and have just one map (not sure if that would make sense)?

We'd have to add in another field like `was_declared`. Also I think adding location to all the variables is unnecessary at this point.

viktormalik: > We'd have to add in another field like `was_declared`. Also I think adding location to all the variables is unnecessary at this point.

Or make it `std::optional<location> decl_location` (with default value `std::nullopt`) but that's probably just an implementation detail.

tests/semantic_analyser.cpp
viktormalik: Indeed I was and when I tried to update, I accidentally broke your PR :sob: I'm so sorry about that, you'll need to force-push the latest version.

viktormalik: Anyways, with the latest version of this PR, I revealed a couple of bugs:

1. Variable of struct type:
```
# bpftrace -e 'BEGIN { let $f: struct task_struct; $f = *curtask; print($f.pid); }'
Segmentation fault (core dumped)
```
While without the declaration, it correctly reports the size of `task_struct` exceeding the stack:
```
# bpftrace -e 'BEGIN { $f = *curtask; print($f.pid); }'
stdin:1:9-22: ERROR: Value is too big (13696 bytes) for the stack. Try reducing its size, storing it in a map, or creating it in argument position to a helper call.

Examples:
    `$s = str(..);` => `$s = str(.., 32);`
    `$s = str(..);` => `@s = str(..);`
    `$s = str(..); print($s);` => `print(str(..));`

BEGIN { $f = *curtask; print($f.pid); }
        ~~~~~~~~~~~~~

```

2. Variable of struct pointer type:
```
bpftrace -e 'BEGIN { let $f: struct task_struct *; $f = curtask; print($f->pid); }'
stdin:1:40-52: ERROR: Type mismatch for $f: trying to assign value of type 'struct task_struct *' when variable already has a type 'struct task_struct *'
BEGIN { let $f: struct task_struct *; $f = curtask; print($f->pid); }
                                       ~~~~~~~~~~~~
stdin:1:54-62: ERROR: Variable used before it was assigned: $f
BEGIN { let $f: struct task_struct *; $f = curtask; print($f->pid); }
                                                     ~~~~~~~~
```

viktormalik: For the record, I'm perfectly ok with disabling declarations of struct-typed variables now and fixing them later.

jordalgo: Good catches! I found fixes for both.

jordalgo:
I'm just going to mash these into the same commit. I can't keep track of my changes and small name change to `variable_val_`

tests/semantic_analyser.cpp
viktormalik: Yep, looks good, thanks!

viktormalik (commented):
Two more small comments, otherwise LGTM.

The type names are still to to be solved but I think that we can do that async.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2761,18 +2802,32 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
...
+                             ? "when variable already contains a value of type"
+                             : "when variable already has a type";
+      LOG(ERROR, assignment.loc, err_)
+          << "Type mismatch for " << var_ident << ": "
+          << "trying to assign value of type '" << assignTy << "' "
+          << err_segment << " '" << storedTy << "'";
+    } else {
+      if (!search->second.was_assigned) {
+        // The assign type is more complete as it may resolve builtins
+        // like `curtask` which is actually of type 'struct task_struct *'
```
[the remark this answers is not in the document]
jordalgo: Sorry, yes, more complete than the storedTy. Will update.

viktormalik (approved):
LGTM, I'm ok with merging this as it is.

Nice work!

src/lexer.l
```diff
@@ -208,13 +209,13 @@ struct|union|enum       {
...
 }
 <STRUCT_AFTER_COLON>{
-  "{"|","|")"           {
-                          yy_pop_state(yyscanner);
-                          unput(yytext[0]);
-                          return Parser::make_IDENT(struct_type + " " + trim(buffer), loc);
-                        }
-  .                     { buffer += yytext[0]; }
-  \n                    buffer += '\n'; loc.lines(1); loc.step();
+  "{"|","|")"|"*"|";"|"[" {
```
ajor, line 212: Should we invert this and instead of trying to think of all characters that aren't allow, just say only `ident` characters are allowed (with whitespace before the ident)?

ajor (commented):
I got a crash with this script:
```
BEGIN { let $x: struct Foo[1]; }
```

src/lexer.l
jordalgo: Good idea. Will update.

jordalgo:
> I got a crash with this script: `BEGIN { let $x: struct Foo[1]; }`

Good catch. I know what's wrong. Will add a fix + a runtime test.

tests/runtime/scripts/struct_let.bt
ajor, line 1: Nit: the missing newline at end of the file will mess up some CLI tools

src/types.h
```diff
@@ -159,6 +159,7 @@ class SizedType {
   bool ctx_ = false;                                   // Is bpf program context
   std::unordered_set<std::string> btf_type_tags_ = {}; // Only populated for
                                                        // Type::pointer
+  size_t num_elements_ = 0; // Only populated for array types
```
danobi, line 162: See `CreateArray()`. It looks like this information is already present, except perhaps not super clearly. Basically `num_elements_` exists as `size_bits_ / 8 / element_type_->GetSize()`. I think we should only have one way to encode array size.

src/types.h
```diff
@@ -292,7 +293,11 @@ class SizedType {
   size_t GetNumElements() const
   {
     assert(IsArrayTy() || IsStringTy());
-    return IsStringTy() ? size_bits_ : size_bits_ / element_type_->size_bits_;
+    // For array's we can't just do size_bits_ / element_type.GetSize()
+    // because we might not know the size of the element type (it may be 0)
+    // if it's being resolved in a later AST pass e.g. for an imported type:
+    // `let $x: struct Foo[10];`
+    return IsStringTy() ? size_bits_ : num_elements_;
```
danobi, line 300: Need to return bytes

```suggestion
    return IsStringTy() ? GetSize() : num_elements_;
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2381,27 +2381,33 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
...
+  // Arrays and structs need not to be copied when assigned to local variables
+  // since they are treated as read-only - it is sufficient to assign
+  // the pointer and do the memcpy/proberead later when necessary
+  if (var_type.IsArrayTy() || var_type.IsRecordTy()) {
+    auto &pointee_type = var_type.IsArrayTy() ? *var_type.GetElementTy()
+                                              : var_type;
+    alloca_type = CreatePointer(pointee_type, var_type.GetAS());
+  }
+
+  AllocaInst *val = b_.CreateAllocaBPFInit(alloca_type, var_ident);
```
danobi, line 2400: Ok, so it's zeroed. Let's document that. Would be good if this was contract.

man/adoc/bpftrace.adoc
```diff
@@ -434,9 +434,10 @@ inside a script config block.
 
 === Data Types
 
-The following fundamental integer types are provided by the language.
-Integers are internally represented as 64 bit signed.
-If you need another representation, you may cast to the following built in types:
+The following fundamental types are provided by the language.
+Note: Integers are by default represented as 64 bit signed but that can be 
+changed by either casting them or, for scratch variables, explicitly specifying 
+the type upon declaration.
```
danobi, line 440: String is a fundamental data type too, right? Perhaps document that as well

src/ast/passes/semantic_analyser.cpp
```diff
@@ -83,6 +83,46 @@ static std::pair<int64_t, int64_t> getIntTypeRange(const SizedType &ty)
   }
 }
 
+static bool IsValidVarDeclType(const SizedType &ty)
```
danobi, line 86: What is the rationale behind the selection? All the percpu ones cannot be declared? If so, can `ty.NeedsPercpuMap()` be used instead? 

Whatever it is, it should probably be documented. Will help inform the next person who adds/removes a type (hopefully nobody is adding more).

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2898,6 +2946,54 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+          << "Variable " << var_ident << " never assigned to.";
+    }
+
+    return;
+  }
+
+  bool can_resize = !(decl.set_type && decl.var->type.GetSize() > 0);
+
+  variables_[scope_].insert(
+      { var_ident, { decl.var->type, can_resize, false } });
```
danobi, line 2993: Nit: use designated initializer here to improve readability

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2898,6 +2946,54 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+
+    if (is_final_pass() && !search->second.was_assigned) {
+      LOG(WARNING, decl.loc, out_)
+          << "Variable " << var_ident << " never assigned to.";
+    }
+
+    return;
+  }
+
+  bool can_resize = !(decl.set_type && decl.var->type.GetSize() > 0);
```
danobi, line 2990: Nit: De Morgan this to improve readability 
```suggestion
  bool can_resize = !decl.set_type || decl.var->type.GetSize() == 0;
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2898,6 +2946,54 @@ void SemanticAnalyser::visit(AssignConfigVarStatement &assignment)
...
+  if (!IsValidVarDeclType(decl.var->type)) {
+    LOG(ERROR, decl.loc, err_)
+        << "Invalid variable declaration type: " << decl.var->type;
+  }
+
+  // This should be the first time we're seeing this variable
+  if (auto decl_search = variable_decls_[scope_].find(var_ident);
+      decl_search != variable_decls_[scope_].end()) {
+    if (decl_search->second != decl.loc) {
+      // Note this will change with proper block/lexical scoping
```
danobi, line 2962: Will be confusing if we never end up adding new scoping rules
```suggestion
      // Note this will change if/when proper block/lexical scoping is added to language
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2373,7 +2419,7 @@ void SemanticAnalyser::visit(For &f)
 
   f.decl->type = CreateTuple(bpftrace_.structs.AddTuple({ *mapkey, *mapval }));
 
-  variable_val_[scope_][decl_name] = f.decl->type;
+  variables_[scope_][decl_name] = { f.decl->type, true, true };
```
danobi, line 2422: Nit: desginated initializer to improve readability

danobi (commented):
really cool!

src/types.h
danobi: Scratch that - looks like you did exactly this :)

src/types.h
jordalgo: This was the previous logic. I'll change it in a follow up PR. I just changed the right side of the ternary.

src/ast/passes/semantic_analyser.cpp
jordalgo: Yeah it's all the map aggregation types that can't go into scratch variables. I'll see if they're the same as `NeedsPercpuMap` and document either way.

man/adoc/bpftrace.adoc
jordalgo: I'm going to update docs for type stuff in a follow up PR, there was some discussion in this PR or another (I can't recall).

src/ast/passes/semantic_analyser.cpp
jordalgo: They're not quite the same. stack_mode is not a valid var decl type at the moment - and actually I should probably add 'void' and 'none' for now.

viktormalik:
I hate to reopen the naming discussion again but nevertheless I will.

@fbs has made a good point in #3509 that we should only use `type X = ...` to specify user-defined types (i.e. without `let`). Considering that, wouldn't it make more sense to use `var` instead of `let` here (for consistency)?

danobi:
@viktormalik you're suggesting just a naming change, not semantic change, right?

viktormalik:
> @viktormalik you're suggesting just a naming change, not semantic change, right?

Yes, but that was just an idea. After more thoughts, I realized that we may use the new keyword to declare not only variables but also maps so `let` is more general in this context.

jordalgo:
Cool. So anyone opposed to me merging this - anyone want to do a final review?

man/adoc/bpftrace.adoc
```diff
@@ -932,6 +933,19 @@ bpftrace knows two types of variables, 'scratch' and 'map'.
 'scratch' variables are kept on the BPF stack and only exists during the execution of the action block and cannot be accessed outside of the program.
 Scratch variable names always start with a `$`, e.g. `$myvar`.
 
+'scratch' variables can also declared before or during initialization with `let` e.g.
+```
+let $x;
```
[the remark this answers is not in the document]
viktormalik: Does it matter? You still cannot use the variable before assigning to it.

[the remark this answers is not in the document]
jordalgo: It does matter (I also added documentation). Think of this program, which I added a runtime test for:
`BEGIN { let $a; if (false) { $a = 1; } @b = $a; exit();}` - semantic analyzer won't catch this (and it probably never could, except in silly cases like this, because it would need to evaluate the conditional).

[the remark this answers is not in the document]
ajor: Shouldn't this be a semantic analysis error - using a potentially undefined variable? We can tell that it's not set on all control flow paths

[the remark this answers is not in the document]
jordalgo: Oops sorry, didn't see this before I merged. Let me look into fixing this.

PR #3465, opened 2024-09-23
Don't use BTF checking for indexed pointers

fvdlfvdl:
The BPF verifier can use BTF type information to check the validity of pointer / array accesses for kfunc probes. However, it can't do any BTF-based bounds checking on pointers indexed as an array.  So, a simple probe like:

kfunc:kfree_skb_reason
{
	printf("%hx\n", args.skb->head[0]);
}

..would then fail with a verifier error.

67629c1927f4 ("Fix error in dereferencing kernel double pointers") fixed a similar case for double pointers, let's fix it for this case too.

### DIFF
--- a/.github/include/aot_skip.txt
+++ b/.github/include/aot_skip.txt
@@ -72,6 +72,7 @@ aot.probe.kprobe_ignore_missing_probes
 aot.probe.kretfunc
 aot.probe.uprobe_error_missing_probes
 aot.probe.uprobe_ignore_missing_probes
+aot.regression.kfunc array access via pointer
 aot.regression.kfunc double pointer array access
 aot.regression.kfunc double pointer dereference
 aot.tuples.basic tuple map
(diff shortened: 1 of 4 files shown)


fvdlfvdl:
I ran the test case I added manually, as in: used a command line with it to verify the before / after behavior. But apparently the automated AOT test fails with the one I added to the automated tests. I'll check it out.

fvdlfvdl:
> I ran the test case I added manually, as in: used a command line with it to verify the before / after behavior. But apparently the automated AOT test fails with the one I added to the automated tests. I'll check it out.



> I ran the test case I added manually, as in: used a command line with it to verify the before / after behavior. But apparently the automated AOT test fails with the one I added to the automated tests. I'll check it out.

Done.

danobi:
Change LGTM, thanks for that.

But for changelog, perhaps word it something like "Fix verifier error when array indexing through pointer". More of a hint for anyone else hitting this issue

fvdlfvdl:
> Fix verifier error when array indexing through pointer

Yes, good point. Done.

danobi (approved):

viktormalik (approved):
Good catch, thanks!

PR #3467, opened 2024-09-25
Utilize MissingProbes config for offset errors

jordalgo:
Instead of auto aborting if uprobe/kprobe alignment is invalid or if there is a kprobe offset error, use the value of the MissingProbes config to determine if we should abort, warn, or ignore.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -161,10 +161,10 @@ AttachedProbe::AttachedProbe(Probe &probe,
         attach_uprobe(getpid(), safe_mode);
       break;
     case ProbeType::kprobe:
-      attach_kprobe(safe_mode);
+      attach_kprobe();
       break;
     case ProbeType::kretprobe:
-      attach_kprobe(safe_mode);
+      attach_kprobe();
       break;
     case ProbeType::tracepoint:
       attach_tracepoint();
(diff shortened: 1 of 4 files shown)


src/attached_probe.cpp
```diff
@@ -519,8 +503,19 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
...
+  bool alignment_valid = is_alignment_valid(
+      probe_.path, symbol, sym_offset, func_offset, probe_.type, alignment_err);
+
+  if (!alignment_valid) {
+    if (missing_probes == ConfigMissingProbes::error) {
+      throw FatalUserException(alignment_err);
+    } else if (missing_probes == ConfigMissingProbes::warn) {
+      LOG(WARNING) << alignment_err << ". Skipping probe.";
+    }
+    return false;
```
viktormalik, line 516: Using `switch` here would probably be more readable and we'd get some compiler checks.

src/attached_probe.cpp
```diff
@@ -585,17 +576,38 @@ void AttachedProbe::resolve_offset_kprobe(bool safe_mode)
...
+
+  if (func_offset >= sym.size) {
+    std::string offset_err = "Offset outside the function bounds ('" + symbol +
+                             "' size is " + std::to_string(sym.size) + ")";
+    if (missing_probes == ConfigMissingProbes::error) {
+      throw FatalUserException(offset_err);
+    } else if (missing_probes == ConfigMissingProbes::warn) {
+      LOG(WARNING) << offset_err << ". Skipping probe.";
+    }
+    return false;
```
viktormalik, line 592: Same as above (use `switch` statement).

src/attached_probe.cpp
```diff
@@ -585,17 +576,38 @@ void AttachedProbe::resolve_offset_kprobe(bool safe_mode)
...
+  bool alignment_valid = is_alignment_valid(
+      path, symbol, sym_offset, func_offset, probe_.type, alignment_err);
+
+  if (!alignment_valid) {
+    if (missing_probes == ConfigMissingProbes::error) {
+      throw FatalUserException(alignment_err);
+    } else if (missing_probes == ConfigMissingProbes::warn) {
+      LOG(WARNING) << alignment_err << ". Skipping probe.";
+    }
+    return false;
```
viktormalik, line 607: Same as above (use `switch` statement).

src/attached_probe.h
```diff
@@ -42,11 +42,11 @@ class AttachedProbe {
 private:
   std::string eventprefix() const;
   std::string eventname() const;
-  void resolve_offset_kprobe(bool safe_mode);
+  bool resolve_offset_kprobe();
   bool resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps);
   void attach_multi_kprobe(void);
   void attach_multi_uprobe(int pid);
-  void attach_kprobe(bool safe_mode);
```
viktormalik, line 49: Why are we getting rid of `safe_mode` here but not for uprobes?

viktormalik (commented):
LGTM, just a couple of nits and one question

src/attached_probe.cpp
```diff
@@ -381,54 +381,38 @@ static uint64_t resolve_offset(const std::string &path,
   return bcc_sym.offset;
 }
 
-static void check_alignment(std::string &path,
-                            std::string &symbol,
-                            uint64_t sym_offset,
-                            uint64_t func_offset,
-                            bool safe_mode,
```
ajor, line 388: I think we'd still want to keep the `safe_mode` option for this function. Normally we don't want to allow unaligned probes (although we shouldn't `abort()`), but there might be cases where we can't correctly determine instruction boundaries and users should be able to override our checks.

jordalgo: >  Normally we don't want to allow unaligned probes

This PR doesn't change that. If the alignment isn't valid we don't attach the probe. This is just utilizing the MissingProbes config to allow the user to specify how they want to handle this failure. The default now, instead of aborting, is to `LOG(WARN)` and not attach the probe. This is so that when attaching to multiple probes with a glob the whole script doesn't fail.

jordalgo:
Maybe a different fix here is to just not force safe_mode in the alignment check. It already defaults to true in most cases so if users want to override they can just pass `--unsafe`

jordalgo:
Upon closer inspection. I think I might abandon this PR and just remove the `HAVE_UNSAFE_PROBE` compiler flag, which doesn't make a ton of sense to me since we already default to safe_mode. It also feels better as an explicit command line flag instead of a config.

jordalgo:
Really there are a few options we could expose to users in the case of invalid alignment:
- warn and force attach anyway
- user failure (similar to abort)
- warn and just skip probe attachment
- silently skip probe attachment

viktormalik:
> Upon closer inspection. I think I might abandon this PR and just remove the `HAVE_UNSAFE_PROBE` compiler flag, which doesn't make a ton of sense to me since we already default to safe_mode. It also feels better as an explicit command line flag instead of a config.

Agreed. IIUC, not compiling with `-DALLOW_UNSAFE_PROBE` (which I believe most distros don't use) will force the safe mode for probe alignment checks but not for any other case (e.g. usage of `system()`). I'm not sure what the motivation for the flag was but it feels sort of redundant to me.

I also remember some discussion where someone claimed that kernel will not allow unaligned kprobes anyways. Maybe we could verify that and get rid of the check entirely (at least for kprobes)?

> Really there are a few options we could expose to users in the case of invalid alignment:
> 
> * warn and force attach anyway
> * user failure (similar to abort)
> * warn and just skip probe attachment
> * silently skip probe attachment

I think that it's generally a good idea to respect what we do for missing probes here.

viktormalik:
> > I think that it's generally a good idea to respect what we do for missing probes here.
> 
> That means we would never attempt attachment if the alignment is invalid. Do we see any use cases where we want to allow users to force attachment? If not, then I'll keep this PR and just delete the compiler flag.

You're right, allowing to force the attachment is the reason why unsafe flag exists for this case. I'm not very familiar with its motivation, though.

jordalgo:
> I also remember some discussion where someone claimed that kernel will not allow unaligned kprobes anyways. 

Yup, this seems to be the case - at least on my 5.12 kernel.

```
ERROR: Possible attachment attempt in the middle of an instruction, try a different offset.
ERROR: Error attaching probe: kprobe:do_nanosleep+2
```

jordalgo:
Abandoning in favor of this: https://github.com/bpftrace/bpftrace/pull/3476

PR #3470, opened 2024-09-26
bug fix: stderr to stdout

aaliyahnl:
##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -89,65 +89,65 @@ void usage()
   std::cout << "USAGE:" << std::endl;
   std::cout << "    bpftrace [options] filename" << std::endl;
   std::cout << "    bpftrace [options] - <stdin input>" << std::endl;
-  std::cerr << "    bpftrace [options] -e 'program'" << std::endl;
-  std::cerr << std::endl;
-  std::cerr << "OPTIONS:" << std::endl;
-  std::cerr << "    -B MODE        output buffering mode ('full', 'none')" << std::endl;
-  std::cerr << "    -f FORMAT      output format ('text', 'json')" << std::endl;
-  std::cerr << "    -o file        redirect bpftrace output to file" << std::endl;
-  std::cerr << "    -e 'program'   execute this program" << std::endl;
-  std::cerr << "    -h, --help     show this help message" << std::endl;
-  std::cerr << "    -I DIR         add the directory to the include search path" << std::endl;
-  std::cerr << "    --include FILE add an #include file before preprocessing" << std::endl;
-  std::cerr << "    -l [search|filename]" << std::endl;
-  std::cerr << "                   list kernel probes or probes in a program" << std::endl;
-  std::cerr << "    -p PID         enable USDT probes on PID" << std::endl;
-  std::cerr << "    -c 'CMD'       run CMD and enable USDT probes on resulting process" << std::endl;
-  std::cerr << "    --usdt-file-activation" << std::endl;
-  std::cerr << "                   activate usdt semaphores based on file path" << std::endl;
-  std::cerr << "    --unsafe       allow unsafe builtin functions" << std::endl;
-  std::cerr << "    -q             keep messages quiet" << std::endl;
-  std::cerr << "    --info         Print information about kernel BPF support" << std::endl;
-  std::cerr << "    -k             emit a warning when a bpf helper returns an error (except read functions)" << std::endl;
-  std::cerr << "    -kk            check all bpf helper functions" << std::endl;
-  std::cerr << "    -V, --version  bpftrace version" << std::endl;
-  std::cerr << "    --no-warnings  disable all warning messages" << std::endl;
-  std::cerr << std::endl;
-  std::cerr << "TROUBLESHOOTING OPTIONS:" << std::endl;
-  std::cerr << "    -v                      verbose messages" << std::endl;
-  std::cerr << "    --dry-run               terminate execution right after attaching all the probes" << std::endl;
-  std::cerr << "    -d STAGE                debug info for various stages of bpftrace execution" << std::endl;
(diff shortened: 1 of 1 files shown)


jordalgo (changes requested):
Thanks for the contribution!

Few additional places where we can change the output:
- https://github.com/bpftrace/bpftrace/blob/master/src/bpftrace.cpp#L1037
- https://github.com/bpftrace/bpftrace/blob/master/src/aot/aot_main.cpp#L18-L29

ajor:
The reason `usage()` prints to standard error is that it's often the only message displayed when invalid command line arguments are used, e.g.: https://github.com/bpftrace/bpftrace/blob/1517d1cc9b7436069412d764ae32d34c9e58afd7/src/main.cpp#L633-L636

We should always print some form of message to stderr when there is an error. Whether this means adding new error messages at each point we currently call `usage()`, or just continuing to have `usage()` print to stderr, I don't really mind.

`--help` was later introduced and also calls ends up calling `usage()` - I agree that in this instance it shouldn't go to stderr.

aaliyahnl:
I replaced more err statements as requested by jordalgo. 

@ajor could you clarify the next step?

Since I replaced the std:cerr statements, I see I can now add error statements each time usage() is called. 
1) Is this format acceptable?
std:cerr << "message related to error conditions" << endl;

2) Should do this for every call to usage()?
 You mention the call to usage() in --help as an exception, but I don't see a place that matches this description. I may not be clear on what --help means.

ajor:
After looking at what other CLI tools do (e.g. `nm`, `ps`) and searching online [[1]](http://courses.cms.caltech.edu/cs11/material/general/usage.html#:~:text=It's%20important%20to%20always%20print,rather%20than%20to%20a%20terminal.) [[2]](https://unix.stackexchange.com/questions/8813/should-the-usage-message-go-to-stderr-or-stdout), I think we should continue to print usage to `stderr` when there is an error, but print it to `stdout` when `--help` is requested.

This is what I meant by `--help` calling `usage()`:
https://github.com/bpftrace/bpftrace/blob/1517d1cc9b7436069412d764ae32d34c9e58afd7/src/main.cpp#L613-L615

It's the only caller of `usage()` which doesn't exit with an error code and should print to `stdout`. All the other callers of `usage()` report errors and should print to `stderr`.

If we're going to keep `usage()` going to `stderr` when there are errors, adding extra error messages shouldn't be required for this PR (but would still be nice to have at some point). For reference, here's a good example of how we can output error messages with `LOG(ERROR)`: https://github.com/bpftrace/bpftrace/blob/1517d1cc9b7436069412d764ae32d34c9e58afd7/src/main.cpp#L638-L641

aaliyahnl:
Thank you for the detailed feedback. I can return my changes in usage() to stderr. 

For the --help condition, I am thinking to replace the call to usage() with the contents of the function. Then modify the contents to have stout.

I can add LOG(ERROR) to the other calls to usage() too. 

Two follow up questions:

1) Should I return the change in info() too?
2) Would it be easier to make another issue for adding LOG(ERROR) or would it be better to continue this here?

ajor:
> Thank you for the detailed feedback. I can return my changes in usage() to stderr.
> 
> For the --help condition, I am thinking to replace the call to usage() with the contents of the function. Then modify the contents to have stout.

Sorry if I'm misunderstanding, but I think we might be able to do it with less code duplication by having `usage()` accept a `std::ostream&` argument (`usage(std::ostream &out)`) like Printer does: https://github.com/bpftrace/bpftrace/blob/1517d1cc9b7436069412d764ae32d34c9e58afd7/src/ast/passes/printer.h#L12

Then we could call it as either `usage(std::cout)` or `usage(std::cerr)` depending on where we want the output to go to.

> 1. Should I return the change in info() too?

Your info change is good - that'll never be printed in response to an error so `stdout` always makes sense.

> 2. Would it be easier to make another issue for adding LOG(ERROR) or would it be better to continue this here?

Entirely up to you :)

aaliyahnl:
I made the requested change, but it seems to be failing many tests. Any recommendation on how to proceed? I can make a new issue for the error statements after this one. Thank you again.

jordalgo:
@aaliyahnl From the error messages it seems like these are all valid errors that need to be fixed (e.g. calls to `usage` that need to be updated) and a formatting error.

src/aot/aot_main.cpp
```diff
@@ -12,21 +12,21 @@
...
-  std::cerr << "    -f FORMAT      output format ('text', 'json')" << std::endl;
-  std::cerr << "    -o file        redirect bpftrace output to file" << std::endl;
-  std::cerr << "    -q,            keep messages quiet" << std::endl;
-  std::cerr << "    -v,            verbose messages" << std::endl;
-  std::cerr << "    -d STAGE       debug info for various stages of bpftrace execution" << std::endl;
-  std::cerr << "                   ('all', 'libbpf', 'verifier')" << std::endl;
-  std::cerr << "    -h, --help     show this help message" << std::endl;
-  std::cerr << "    -V, --version  bpftrace version" << std::endl;
-  std::cerr << std::endl;
+  std::out << "USAGE: " << filename << " [options]" << std::endl;
```
ajor, line 18:
```suggestion
  out << "USAGE: " << filename << " [options]" << std::endl;
```
`out` is just a local variable - it isn't in namespace `std`

aaliyahnl:
Oh, thank you! I've been trouble getting compiling, and wasn't able to see the error message. I'll work on this as soon as I can

danobi:
Closing due to inactivity

jordalgo:
I feel like we might have closed this too aggressively. @aaliyahnl Any desire to continue work on this? If not, I'll hand it off.

aaliyahnl:
Thank you for checking. Please feel free to hand it off! I thought I would be able to continue, but no longer am. Glad someone else is willing.

PR #3474, opened 2024-09-27
[Draft][Unfinished] Use blazesym for symbolization/inspection matters

danielocfb:
[Draft][Unfinished] Use blazesym for symbolization/inspection matters

Use blazesym for symbolization/inspection matters. Blazesym [0] is a
library that handles address symbolization and related tasks. It can
work with various formats including ELF, DWARF, Gsym, and Breakpad and
supports mapping from address to symbol (+ meta data) as well as from
symbol to address (+ meta data).

This pull request proposes adding support for using blazesym for these
tasks to bpftrace. Doing so is beneficial for a multitude of reasons,
including but not limited to:
- it will allow bpftrace to incorporate DWARF symbol information into
  its various workflows
- it will switch away more parts of bpftrace from relying on BCC;
  removal of BCC as a dependency is a long-term goal of the project

At this point, this is not a full replacement of the existing
functionality. Rather, blazesym usage is guarded by a build time switch
and opt-in. The reason for this are two fold:
- currently we pull the blazesym source and compile it, because
  inclusion in various distributions is still limited
  - this means a Rust toolchain is currently required as well
  - also, that means that we currently only support linking the library
    statically, which may not be preferred in all cases
- we do anticipate some corner cases to be handled differently and we
  want to have a way to easily switch between the two "backends" to
  quickly compare results

[0] https://github.com/libbpf/blazesym

##### Checklist


danielocfb:
The main purpose of the PR, at this point, is to get feedback on the idea and to motivate the inclusion of some of the preparatory changes (that are improvements and can stand on their own) into main line via separate pull requests.
Far from every code path has been converted over.

danobi:
Do you want us to take a look now? Or do you expect to make more changes still in draft mode?

danielocfb:
> Do you want us to take a look now? Or do you expect to make more changes still in draft mode?

It's not yet ready for review, mostly just in "FYI" state. I plan to update everything as I go along. That being said, I'd hope that this PR can motivate some early inclusion of some internal refactorings. E.g., I think both https://github.com/bpftrace/bpftrace/pull/3474/commits/53b0fd9e45d0fa6696838f702394b9de6ce9dfc7 and https://github.com/bpftrace/bpftrace/pull/3474/commits/92c972560e176cd9e5f238c7ed82d088029c9623 would make sense to land, but without the context of this PR it's not obvious where they'd lead. So, would it be okay to create pull requests for those changes to minimize rebasing efforts for the `blazesym` work down the line?

> I don't get notifications when PRs are opened in draft, fwiw

Oh, didn't know that. I'll tag @ajor directly then, FYI.

danielocfb:
Also, as a general question: I see that `bcc` supports pulling debug info/binaries using `debuginfod`. Blazesym doesn't support `debuginfod` directly. Rather, it's a building block on top ([example](https://github.com/libbpf/blazesym/tree/main/examples/sym-debuginfod)).

Is the `debuginfod` logic being relied upon by users (I saw it was a compile time conditional in `bcc`) and a requirement for the conversion?

ajor:
I'm fine without debuginfod - my only experience with it is it hanging bpftrace occasionally when it can't get a network connection! If people really do want it, we can add it back at a later point.

danielocfb:
Closing this PR in favor of #3760.

PR #3475, opened 2024-09-28
Fix: json formatting for hex values

amiremohamadi:
hex values are not supported in json format, so they should be output as a string when json formatting is selected.

```sh
$ bpftrace -q -f json -e 'BEGIN { @=(int16*) 0x32; exit(); }' | jq .
```

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -414,12 +414,6 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
     case Type::strerror: {
       return strerror(read_data<uint64_t>(value.data()));
     }
-    case Type::pointer:
-    case Type::reference: {
-      std::ostringstream res;
-      res << "0x" << std::hex << read_data<uint64_t>(value.data());
-      return res.str();
-    }
     case Type::none: {
       return "";
     }
@@ -428,6 +422,8 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
     case Type::lhist:
     case Type::stack_mode:
     case Type::stats:
+    case Type::pointer:
+    case Type::reference:
     case Type::timestamp_mode: {
       LOG(BUG) << "Invalid value type: " << type;
     }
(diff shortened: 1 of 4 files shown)


jordalgo (approved):
LGTM. Thanks for the fix!

viktormalik (changes requested):
Wouldn't it be better to use a decimal number instead of a hex string? JSON output is mainly used for machine processing and this way, we're forcing users to add additional string parsing into the pipeline.

Blocking the PR until we resolve this.

jordalgo:
> Wouldn't it be better to use a decimal number instead of a hex string? JSON output is mainly used for machine processing and this way, we're forcing users to add additional string parsing into the pipeline.

Yeah, perhaps this is the better way to go. Though if people want hex, you're adding an additional step into their processing pipeline as well. I don't feel strongly either way.

viktormalik:
> > Wouldn't it be better to use a decimal number instead of a hex string? JSON output is mainly used for machine processing and this way, we're forcing users to add additional string parsing into the pipeline.
> 
> Yeah, perhaps this is the better way to go. Though if people want hex, you're adding an additional step into their processing pipeline as well. I don't feel strongly either way.

The step is always there unless people want hex strings which doesn't seem like something people usually want (at least I don't see a point in having an "address string" as a value of a pointer unless you want to print it straight away).

viktormalik (approved):

amiremohamadi:
actually switching to a decimal number instead of a hex string makes sense to me — as @viktormalik mentioned JSON is mainly used for machine processing.

I've made some updates, could you please take another look please?

@viktormalik 
@jordalgo

src/output.cpp
```diff
@@ -1112,8 +1112,18 @@ std::string JsonOutput::value_to_str(BPFtrace &bpftrace,
                                      uint32_t div,
                                      bool is_map_key) const
 {
-  auto str = Output::value_to_str(
-      bpftrace, type, value, is_per_cpu, div, is_map_key);
+  std::string str;
+
+  switch (type.GetTy()) {
```
jordalgo, line 1117: this is super nitpicky so I won't block but I think perhaps moving the pointer and reference printing out of the base Output class `value_to_str` and putting them (like you have it here) in `TextOutput::value_to_str` so it's clear that there is a difference in the output type.

src/output.cpp
```diff
@@ -431,8 +425,10 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
     case Type::timestamp_mode: {
```
jordalgo, line 425: Looking good overall! One last small thing. Can you add `case Type::pointer` and `case Type::reference` right above `timestamp_mode` so the switch is exhaustive and will log a bug if anything tries to call `value_to_str` with these two types on the base class?

jordalgo (approved):
Thanks for the fix!

PR #3476, opened 2024-09-28
Update 'check_alignment'

jordalgo:
Remove HAVE_UNSAFE_PROBE compiler flag. This seems redundant at best since safe mode is the default and I can't seem to find reasoning for it when it was added: https://github.com/bpftrace/bpftrace/commit/18b96356408392e7330d3e95ef871088c00c6c97

Also remove 'check_alignment' for kprobes as the
kernel gives a nice error message if you try to add a probe in the middle of an instruction:
```
ERROR: Possible attachment attempt in the middle of an instruction, try a different offset.
ERROR: Error attaching probe: kprobe:do_nanosleep+2
```

##### Checklist


viktormalik:
@olsajiri do you remember what was the original purpose of 18b96356408392e7330d3e95ef871088c00c6c97? Can we remove the `HAVE_UNSAFE_PROBE` compile flag and just keep the `--unsafe` CLI option?

jordalgo:
Just to add a bit. The compiler flag also doesn't protect against anything because if you can run as sudo you can always just write raw libbpf to attach to uprobes in the middle of an instruction -- just sayin 😉

danobi:
I suspect the intention was so let packagers have a say in unsafe attachment. It doesn't look like the build flag is documented. Currently for my arch linux install it's set to default value of off. Not sure anyone even knows about it. Probably safe to remove.

src/attached_probe.cpp
```diff
@@ -393,19 +393,14 @@ static void check_alignment(std::string &path,
...
-#ifndef HAVE_UNSAFE_PROBE
-  safe_mode = true;
-#endif
-
   switch (aligned) {
     case AlignState::Ok:
       return;
     case AlignState::NotAlign:
       if (safe_mode)
         LOG(BUG) << "Could not add " << type
```
danobi, line 401: not actually a bug right? LOG(BUG) triggers coredump, probably want to throw `FatalUserException`.

probably in a different PR though

danobi (approved):

src/attached_probe.cpp
jordalgo: Ah true. Good catch. Maybe same PR, but a separate commit.

jordalgo:
@viktormalik Going to merge this unless you have any objections. I want to unblock someone who is running into an issue with uprobe glob matching.

viktormalik (approved):
No objections, let's remove it.

PR #3481, opened 2024-10-02
Use 'missing_probes' config for check_alignment

jordalgo:
Instead of attemping to attach uprobes if they fail `check_alignment` never attach and use the `missing_probes` config to determine if we error, warn, or ignore.

This removes the ability to use the "--unsafe" flag
and force attachment anyway. This was done because
attaching in the middle of an instruction can be very unsafe
and cause data corruption or crashes. Even the kernel
doesn't allow this for kprobes.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -381,12 +381,12 @@ static uint64_t resolve_offset(const std::string &path,
   return bcc_sym.offset;
 }
 
-static void check_alignment(std::string &path,
+static bool check_alignment(std::string &path,
                             std::string &symbol,
                             uint64_t sym_offset,
                             uint64_t func_offset,
-                            bool safe_mode,
-                            ProbeType type)
+                            ProbeType type,
+                            std::string &err_msg)
 {
   Disasm dasm(path);
   AlignState aligned = dasm.is_aligned(sym_offset, func_offset);
@@ -395,36 +395,23 @@ static void check_alignment(std::string &path,
 
   switch (aligned) {
     case AlignState::Ok:
-      return;
+      return true;
     case AlignState::NotAlign:
-      if (safe_mode)
-        throw FatalUserException("Could not add " + probetypeName(type) +
-                                 " into middle of instruction: " + tmp +
-                                 " (use --unsafe to force attachment)");
-      else
-        LOG(WARNING) << "Unsafe " << type
-                     << " in the middle of the instruction: " << tmp;
-      break;
-
+      err_msg = "Could not add " + probetypeName(type) +
+                " into middle of instruction: " + tmp;
+      return false;
     case AlignState::Fail:
-      if (safe_mode)
-        throw FatalUserException("Failed to check if " + probetypeName(type) +
-                                 " is in proper place: " + tmp +
-                                 " (use --unsafe to force attachment)");
-      else
-        LOG(WARNING) << "Unchecked " << type << ": " << tmp;
-      break;
-
+      err_msg = "Failed to check if " + probetypeName(type) +
+                " is in proper place: " + tmp;
+      return false;
     case AlignState::NotSupp:
-      if (safe_mode)
-        throw FatalUserException("Can't check if " + probetypeName(type) +
-                                 " is in proper place (compiled without "
-                                 "(k|u)probe offset support): " +
-                                 tmp + " (use --unsafe to force attachment)");
-      else
-        LOG(WARNING) << "Unchecked " << type << " : " << tmp;
-      break;
+      err_msg = "Can't check if " + probetypeName(type) +
+                " is in proper place (compiled without "
+                "(k|u)probe offset support): " +
+                tmp;
+      return false;
   }
+  return false;
 }
 
 bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
@@ -433,6 +420,7 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
   struct symbol sym = {};
   std::string &symbol = probe_.attach_point;
   uint64_t func_offset = probe_.func_offset;
+  auto missing_probes = bpftrace_.config_.get(ConfigKeyMissingProbes::default_);
 
   sym.name = "";
   option.use_debug_file = 1;
@@ -467,8 +455,6 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
     if (!sym.start) {
       const std::string msg = "Could not resolve symbol: " + probe_.path + ":" +
                               symbol;
-      auto missing_probes = bpftrace_.config_.get(
-          ConfigKeyMissingProbes::default_);
       if (!has_multiple_aps || missing_probes == ConfigMissingProbes::error) {
         throw FatalUserException(msg + ", cannot attach probe.");
       } else {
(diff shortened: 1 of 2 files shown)


src/attached_probe.cpp
```diff
@@ -515,8 +501,20 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
...
+                       symbol,
+                       sym_offset,
+                       func_offset,
+                       probe_.type,
+                       alignment_err_msg)) {
+    if (missing_probes == ConfigMissingProbes::error) {
+      throw FatalUserException(alignment_err_msg + ", cannot attach probe.");
+    } else if (missing_probes == ConfigMissingProbes::warn) {
+      LOG(WARNING) << alignment_err_msg << ", skipping probe.";
+    }
```
danobi, line 515: switch/case all variants?

src/attached_probe.cpp
```diff
@@ -515,8 +501,20 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
   if (func_offset == 0)
     return true;
 
-  check_alignment(
-      probe_.path, symbol, sym_offset, func_offset, safe_mode, probe_.type);
```
danobi, line 519: What about safe_mode? 

I think this PR takes away ability for `--unsafe` to force an attachment. Is that the intention?

danobi: Good call, then :)

Maybe point it out in commit msg in case someone wonders down the line where it went. Changelog entry about that would be good too

danobi (commented):
Outputting fact that we're attaching probes but then not is a bit unfortunate:

```
$ sudo ./build/src/bpftrace -e 'uprobe:./hw:main+5 {}'
Attaching 1 probe...
WARNING: Could not add uprobe into middle of instruction: ./hw:main+5, skipping probe.
^C
```

Not sure it's worth it to rearrange all the code to improve it a little, though. Especially since this is such a niche case.

Can we add a few runtime tests for this? Should be possible to use hello world test_prog:
```
dxu@kashmir~/d/bpftrace (check_alignment) $ cat /tmp/hw.c
#include <stdio.h>
int main()
{
  printf("hello\n");
  return 0;
}
dxu@kashmir~/d/bpftrace (check_alignment) $ gcc /tmp/hw.c -o hw
dxu@kashmir~/d/bpftrace (check_alignment) $ sudo ./build/src/bpftrace -e 'config = { missing_probes="error" } uprobe:./hw:main+5 {}'
Attaching 1 probe...
ERROR: Could not add uprobe into middle of instruction: ./hw:main+5, cannot attach probe.
```
```

**Review by ajor — CHANGES_REQUESTED** [review]
I don't agree with removing unsafe attachment entirely - our alignment calculation could be wrong or someone could be tracing some kind of JIT'd code which we can't properly analyse. `--unsafe` means it's up to the user to check themselves - they just need to be very careful about using it.

I didn't realise when we talked about this before that implementing this change would change the default for unaligned probe attachment from an error to a warning. Not sure that this is a good thing. Also there seem to be a couple more cases above where we can fail attachment in various circumstances. Should be maybe put all of these non-missing-probe errors behind a new config flag that defaults to "error"?

**Comment by jordalgo** [comment]
> I didn't realise when we talked about this before that implementing this change would change the default for unaligned probe attachment from an error to a warning. Not sure that this is a good thing. 

To be clear it's a warning with no attachment. Whereas in trunk/master you could pass "--unsafe" and get a warning with attachment.

> Also there seem to be a couple more cases above where we can fail attachment in various circumstances. Should be maybe put all of these non-missing-probe errors behind a new config flag that defaults to "error"?

I would be good for a new config flag if we wanted to still allow attachment. "missing_probes" is not a great config name for this anyway. Maybe something like "misaligned_probes" where the options are "error", "attach_warn", "skip_warn", "skip_silent"?

**Comment by ajor** [comment]
> To be clear it's a warning with no attachment. Whereas in trunk/master you could pass "--unsafe" and get a warning with attachment.

Yeah, but I meant in the default case without setting any config options and without using `--unsafe`, we now only warn instead of error when failing to attach these probes.

> I would be good for a new config flag if we wanted to still allow attachment. "missing_probes" is not a great config name for this anyway. Maybe something like "misaligned_probes" where the options are "error", "attach_warn", "skip_warn", "skip_silent"?

Sounds reasonable

**Comment by jordalgo** [comment]
> We could just use the same options as "missing_probes"

I'm hesitant to add a very similar probe that has the exact same options.  How about we go back to not involving a config option and instead change the default behavior to skip attachment if there is misalignment and log a warning to the user that if they want to force attachment they need --unsafe (whereby we'll still log a warning). This way scripts won't crash in the case of uprobe globbing but the user is notified of the issue.

**Comment by ajor** [comment]
> I'm hesitant to add a very similar probe that has the exact same options.

Why?

> How about we go back to not involving a config option and instead change the default behavior to skip attachment if there is misalignment and log a warning to the user that if they want to force attachment they need --unsafe (whereby we'll still log a warning). This way scripts won't crash in the case of uprobe globbing but the user is notified of the issue.

This downgrades a misaligned probe from an error to a warning by default, which I'm reluctant to do as this is a real error that indicates something is seriously wrong. Missing probes defaulting to a warning kind of makes sense for cross-kernel compatibility, as function names can change and break scripts, but really it's also legacy behaviour that we need to keep for compatibility. A misaligned probe should never happen unless something else has gone very wrong or the user is doing something silly.

What we have now is:
1. default to erroring out when there's a serious error (unaligned probe)
2. let users tell bpftrace they know better and attach to unaligned probes with `--unsafe`

I don't think either of these should change. I was thinking that this PR would provide a way to ignore the errors from (1) without attaching, but maybe it's not the right solution after all? The original reason for this was LLDB picking the wrong offset to attach to - maybe we should instead add a way to tell bpftrace to not use LLDB's location discovery and just attach to the symbol's address?

**Comment by jordalgo** [comment]
> Why?

As you said recently [here](https://github.com/bpftrace/bpftrace/issues/3293#issuecomment-2391998148), which I agree with: "New config options will form part of our stable interface, so I want to be careful about adding new ones to avoid carrying cruft forward forever." I also think it's confusing if we have two config options that are both similar in functionality and have the same options.

> This downgrades a misaligned probe from an error to a warning by default, which I'm reluctant to do as this is a real error that indicates something is seriously wrong.

I'm not sure if it's a downgrade. We're warning and not attaching (as per my new proposal) that's not exposing the user to more harm IMHO. 

> A misaligned probe should never happen unless something else has gone very wrong or the user is doing something silly.

Right. The reason this PR was opened was because LLDB was doing something silly and we were punishing the user by aborting their script. I'm saying we should just notify and do the safe thing of not attaching.

> maybe we should instead add a way to tell bpftrace to not use LLDB's location discovery and just attach to the symbol's address?

Yeah, at this point I'd probably prefer giving this a shot.

**Comment by ajor** [comment]
> I'm not sure if it's a downgrade. We're warning and not attaching (as per my new proposal) that's not exposing the user to more harm IMHO.

The reason I describe it as a downgrade is that it's a lot easier to miss warnings than errors. e.g. In this contrived script, it'd be very hard to notice a warning from the uprobe being misaligned:
```
uprobe:/bin/bash:main+2
{
}

profile:hz:99
{
  print(kstack);
}
```

jordalgo:
Maybe if we detect if the user is explicitly adding an offset vs the one we get from LLDB. In the former case we could error and the latter case, warn and skip.

jordalgo:
Abandoning in favor of this solution which @ajor and I discussed offline: https://github.com/bpftrace/bpftrace/issues/3493

Issue #1, opened 2018-08-31
lhist truncate zero ranges


brendangregg:
lhist() currently works like this:

```
# ./src/bpftrace -e 'kprobe:do_nanosleep { @ = lhist(pid, 0, 65000, 1000); }'
Attaching 1 probe...
^C

@:
(...,0]                0 |                                                    |
[0, 1000)              0 |                                                    |
[1000, 2000)           4 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@|
[2000, 3000)           0 |                                                    |
[3000, 4000)           0 |                                                    |
[4000, 5000)           0 |                                                    |
[5000, 6000)           0 |                                                    |
[6000, 7000)           0 |                                                    |
[7000, 8000)           0 |                                                    |
[8000, 9000)           0 |                                                    |
[9000, 10000)          2 |@@@@@@@@@@@@@@@@@@@@@@@@@@                          |
[10000, 11000)         0 |                                                    |
[11000, 12000)         0 |                                                    |
[12000, 13000)         0 |                                                    |
[13000, 14000)         0 |                                                    |
[14000, 15000)         0 |                                                    |
[15000, 16000)         0 |                                                    |
[16000, 17000)         0 |                                                    |
[17000, 18000)         0 |                                                    |
[18000, 19000)         0 |                                                    |
[19000, 20000)         0 |                                                    |
[20000, 21000)         0 |                                                    |
[21000, 22000)         0 |                                                    |
[22000, 23000)         0 |                                                    |
[23000, 24000)         0 |                                                    |
[24000, 25000)         0 |                                                    |
[25000, 26000)         0 |                                                    |
[26000, 27000)         0 |                                                    |
[27000, 28000)         0 |                                                    |
[28000, 29000)         0 |                                                    |
[29000, 30000)         0 |                                                    |
[30000, 31000)         0 |                                                    |
[31000, 32000)         0 |                                                    |
[32000, 33000)         0 |                                                    |
[33000, 34000)         0 |                                                    |
[34000, 35000)         0 |                                                    |
[35000, 36000)         0 |                                                    |
[36000, 37000)         0 |                                                    |
[37000, 38000)         0 |                                                    |
[38000, 39000)         0 |                                                    |
[39000, 40000)         0 |                                                    |
[40000, 41000)         0 |                                                    |
[41000, 42000)         0 |                                                    |
[42000, 43000)         0 |                                                    |
[...]

PR #3483, opened 2024-10-03
Use global scratch buffers for reading and writing map values

ryantimwilson:
This PR continues fixing https://github.com/bpftrace/bpftrace/issues/3431, migrating map values to scratch global buffers. This will allow large strings in map values, one of the primary use cases for large strings.

We introduce a write buffer of size 1 and read buffer with size N as its possible multiple values can be read in tandem.

Note in many cases, ResourceAnalyser does not exactly replicate CodegenLLVM meaning it is possible to over-allocate memory at the benefit of code simplicity. However, we did 2 optimizations in ResourceAnalyser which seem to reduce most unnecessary memory based on codegen tests:
1. Don't recurse into map from AssignmentMap to avoid an unnecessary read buffer
2. Check if we actually need to do a copy during AssignmentMap to avoid an unnecessary write buffer

After this, we see no additional global buffers in current codegen tests.

We also add new codegen and runtime tests for big strings in map values.

### DIFF
--- a/src/ast/async_ids.h
+++ b/src/ast/async_ids.h
@@ -12,6 +12,7 @@ namespace ast {
   DO(bpf_print)                                                                \
   DO(non_map_print)                                                            \
   DO(printf)                                                                   \
+  DO(read_map_value)                                                           \
   DO(skb_output)                                                               \
   DO(strftime)                                                                 \
   DO(str)                                                                      \
--- a/src/ast/codegen_helper.cpp
+++ b/src/ast/codegen_helper.cpp
@@ -1,0 +1,33 @@
+#include "codegen_helper.h"
+
+namespace bpftrace::ast {
+
+/* needAssignMapStatementAllocation determines if a map assignment requires a
+ * new memory allocation. This happens only in a few cases e.g. if there are
+ * two map assignments of tuples of different sizes e.g.
+ *   @x = ("xxx", 1); @x = ("xxxxxxx", 1);
+ * which requires a 0 memsetting and copying of each tuple element into the
+ * new allocation before calling bpf_map_update_elem.
+ *
+ * Another case when we need an allocation is for a external struct e.g.
+ *   $v = (struct task_struct *)arg1; @ = *$v;.
+ *
+ * Most cases we can reuse existing BPF memory and not create a new allocation.
+ *
+ * Note this function does NOT determine if an allocation should use scratch
+ * buffer or the stack, that logic is in
+ * IRBuilderBPF::CreateWriteMapValueAllocation
+ */
+bool needAssignMapStatementAllocation(const AssignMapStatement &assignment)
+{
+  const auto &map = *assignment.map;
+  const auto &expr_type = assignment.expr->type;
+  if (shouldBeInBpfMemoryAlready(expr_type)) {
+    return !expr_type.IsSameSizeRecursive(map.type);
+  } else if (map.type.IsRecordTy() || map.type.IsArrayTy()) {
+    return !expr_type.is_internal;
+  }
+  return true;
+}
+
+} // namespace bpftrace::ast
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -769,10 +769,13 @@ void CodegenLLVM::visit(Call &call)
                                          : getMapKey(map);
     if (!is_bpf_map_clearable(map_types_[map.ident])) {
       // store zero instead of calling bpf_map_delete_elem()
-      AllocaInst *val = b_.CreateAllocaBPF(map.type, map.ident + "_zero");
+      auto val = b_.CreateWriteMapValueAllocation(map.type,
+                                                  map.ident + "_zero",
+                                                  call.loc);
       b_.CreateStore(Constant::getNullValue(b_.GetType(map.type)), val);
       b_.CreateMapUpdateElem(ctx_, map.ident, key, val, call.loc);
-      b_.CreateLifetimeEnd(val);
+      if (dyn_cast<AllocaInst>(val))
+        b_.CreateLifetimeEnd(val);
     } else {
       b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
     }
@@ -812,7 +815,7 @@ void CodegenLLVM::visit(Call &call)
       strlen = b_.CreateSelect(Cmp, proposed_strlen, strlen, "str.min.select");
     }
 
-    Value *buf = b_.CreateGetStrAllocation("str", call.loc, async_ids_);
+    Value *buf = b_.CreateGetStrAllocation("str", call.loc);
     b_.CreateMemsetBPF(buf, b_.getInt8(0), max_strlen);
     auto arg0 = call.vargs.front();
     auto scoped_del = accept(call.vargs.front());
@@ -851,9 +854,7 @@ void CodegenLLVM::visit(Call &call)
       length = b_.getInt32(fixed_buffer_length);
     }
 
-    Value *stack_or_scratch_buf = b_.CreateGetStrAllocation("buf",
-                                                            call.loc,
-                                                            async_ids_);
+    Value *stack_or_scratch_buf = b_.CreateGetStrAllocation("buf", call.loc);
     auto elements = AsyncEvent::Buf().asLLVMType(b_, fixed_buffer_length);
     std::ostringstream dynamic_sized_struct_name;
     dynamic_sized_struct_name << "buffer_" << fixed_buffer_length << "_t";
@@ -887,7 +888,7 @@ void CodegenLLVM::visit(Call &call)
     if (dyn_cast<AllocaInst>(buf))
       expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
   } else if (call.func == "path") {
-    Value *buf = b_.CreateGetStrAllocation("path", call.loc, async_ids_);
+    Value *buf = b_.CreateGetStrAllocation("path", call.loc);
     b_.CreateMemsetBPF(buf,
                        b_.getInt8(0),
                        bpftrace_.config_.get(ConfigKeyInt::max_strlen));
@@ -1955,7 +1956,7 @@ void CodegenLLVM::visit(Ternary &ternary)
   // ordering of all the following statements is important
   Value *buf = nullptr;
   if (ternary.type.IsStringTy()) {
-    buf = b_.CreateGetStrAllocation("buf", ternary.loc, async_ids_);
+    buf = b_.CreateGetStrAllocation("buf", ternary.loc);
     uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
     b_.CreateMemsetBPF(buf, b_.getInt8(0), max_strlen);
   }
@@ -2282,9 +2283,7 @@ Value *CodegenLLVM::createTuple(
 {
   auto tuple_ty = b_.GetType(tuple_type);
   size_t tuple_size = datalayout().getTypeAllocSize(tuple_ty);
-  auto buf = b_.CreatePointerCast(
-      b_.CreateTupleAllocation(tuple_type, name, loc, async_ids_),
-      tuple_ty->getPointerTo());
+  auto buf = b_.CreateTupleAllocation(tuple_type, name, loc);
   b_.CreateMemsetBPF(buf, b_.getInt8(0), tuple_size);
 
   for (size_t i = 0; i < vals.size(); ++i) {
@@ -2368,7 +2367,6 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
 {
   Map &map = *assignment.map;
   auto scoped_del = accept(assignment.expr);
-  bool self_alloca = false;
 
   if (!expr_) // Some functions do the assignments themselves
     return;
@@ -2377,53 +2375,39 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
   expr = expr_;
   auto [key, scoped_key_deleter] = getMapKey(map);
   auto &expr_type = assignment.expr->type;
+  const auto self_alloca = needAssignMapStatementAllocation(assignment);
+  val = self_alloca ? b_.CreateWriteMapValueAllocation(map.type,
+                                                       map.ident + "_val",
+                                                       assignment.loc)
+                    : expr;
   if (shouldBeInBpfMemoryAlready(expr_type)) {
     if (!expr_type.IsSameSizeRecursive(map.type)) {
-      val = b_.CreateAllocaBPF(map.type, map.ident + "_val");
       b_.CreateMemsetBPF(val, b_.getInt8(0), map.type.GetSize());
       if (expr_type.IsTupleTy()) {
         createTupleCopy(expr_type, map.type, val, expr);
-        self_alloca = true;
       } else if (expr_type.IsStringTy()) {
         b_.CreateMemcpyBPF(val, expr, expr_type.GetSize());
-        self_alloca = true;
       } else {
         LOG(BUG) << "Type size mismatch. Map Type Size: " << map.type.GetSize()
                  << " Expression Type Size: " << expr_type.GetSize();
       }
-    } else {
-      val = expr;
     }
   } else if (map.type.IsRecordTy() || map.type.IsArrayTy()) {
-    if (expr_type.is_internal) {
-      val = expr;
-    } else {
+    if (!expr_type.is_internal) {
       // expr currently contains a pointer to the struct or array
       // We now want to read the entire struct/array in so we can save it
-      AllocaInst *dst = b_.CreateAllocaBPF(map.type, map.ident + "_val");
       b_.CreateProbeRead(
-          ctx_, dst, map.type, expr, assignment.loc, expr_type.GetAS());
-      val = dst;
-      self_alloca = true;
+          ctx_, val, map.type, expr, assignment.loc, expr_type.GetAS());
     }
-  } else if (map.type.IsPtrTy()) {
-    // expr currently contains a pointer to the struct
-    // and that's what we are saving
-    AllocaInst *dst = b_.CreateAllocaBPF(map.type, map.ident + "_ptr");
-    b_.CreateStore(expr, dst);
-    val = dst;
-    self_alloca = true;
   } else {
     if (map.type.IsIntTy()) {
       // Integers are always stored as 64-bit in map values
       expr = b_.CreateIntCast(expr, b_.getInt64Ty(), map.type.IsSigned());
     }
-    val = b_.CreateAllocaBPF(map.type, map.ident + "_val");
     b_.CreateStore(expr, val);
-    self_alloca = true;
   }
   b_.CreateMapUpdateElem(ctx_, map.ident, key, val, assignment.loc);
-  if (self_alloca)
+  if (self_alloca && dyn_cast<AllocaInst>(val))
     b_.CreateLifetimeEnd(val);
 }
 
@@ -3510,19 +3494,17 @@ void CodegenLLVM::createFormatStringCall(Call &call,
   // The struct is not packed so we need to memset it
   b_.CreateMemsetBPF(fmt_args, b_.getInt8(0), struct_size);
 
-  Value *id_offset = b_.CreateGEP(
-      fmt_struct,
-      b_.CreatePointerCast(fmt_args, fmt_struct->getPointerTo()),
-      { b_.getInt32(0), b_.getInt32(0) });
+  Value *id_offset = b_.CreateGEP(fmt_struct,
+                                  fmt_args,
+                                  { b_.getInt32(0), b_.getInt32(0) });
   b_.CreateStore(b_.getInt64(id + asyncactionint(async_action)), id_offset);
 
   for (size_t i = 1; i < call.vargs.size(); i++) {
     Expression &arg = *call.vargs.at(i);
     auto scoped_del = accept(&arg);
-    Value *offset = b_.CreateGEP(
-        fmt_struct,
-        b_.CreatePointerCast(fmt_args, fmt_struct->getPointerTo()),
-        { b_.getInt32(0), b_.getInt32(i) });
+    Value *offset = b_.CreateGEP(fmt_struct,
+                                 fmt_args,
+                                 { b_.getInt32(0), b_.getInt32(i) });
     if (needMemcpy(arg.type))
       b_.CreateMemcpyBPF(offset, expr_, arg.type.GetSize());
     else if (arg.type.IsIntegerTy() && arg.type.GetSize() < 8)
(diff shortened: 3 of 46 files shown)


ryantimwilson:
Per offline discussion with @ajor and @danobi, PR in current form fails to compile bpfsnake due to kernel verifier saying too many jumps >= 8192 due to `select` LLVM IR instructions

To fix this, I did 2 things:
1. Cached bounded CPU ID at beginning of probe to minimize # of jumps. Unfortunately, this caused LLVM to barf with `BPF program is too large. Processed 1000001 insn` errors due to instruction exploration space (e.g. verifier assumes value could be MAX_CPU_ID or get_snmp_processor_id so has to 2x explore every path using scratch buffers)
2. To fix #1, I just used `assert(snmp_processor_id <= MAX_CPU_ID)`. Previously, this caused too many jumps but because we now do it once per probe, its fine. We can also move to CPUv4 to support long jumps which I will do a bit later but holding off on that now

I will productionize local changes and put them up in a separate commit

danobi:
Playing with snake some more, a carefully placed loop:

```diff
diff ~/scratch/snake.bt ~/scratch/snake.bt.optimized
58c58
<         unroll(50) {
---
>         while($y < 50) {
```

brings insn count down from 400K -> 20K.

So I think unroll() is the true issue here. Program is just too complex.

Let's do the cpuid caching optimization since that will be necessary sooner or later. But let's avoid the complicated changes - we are too deep in complex territory already to start tugging on more threads.

src/ast/async_ids.h
```diff
@@ -12,6 +12,7 @@ namespace ast {
   DO(bpf_print)                                                                \
   DO(non_map_print)                                                            \
   DO(printf)                                                                   \
+  DO(read_map_value)                                                           \
```
ajor, line 15: I'm not familiar with this file, but something feels wrong here. Async events are things that we pass over the ring buffer for processing in userspace.

`read_map_value`, `tuple` and `str` aren't async events as they're all handled synchronously in BPF-space. Possibly also not `bpf_print`.

tests/codegen/llvm/variable.ll
```diff
@@ -13,11 +13,13 @@ target triple = "bpf-pc-linux"
 @AT_y = dso_local global %"struct map_t.0" zeroinitializer, section ".maps", !dbg !25
 @ringbuf = dso_local global %"struct map_t.1" zeroinitializer, section ".maps", !dbg !27
 @event_loss_counter = dso_local global %"struct map_t.2" zeroinitializer, section ".maps", !dbg !41
+@max_cpu_id = dso_local externally_initialized constant i64 zeroinitializer, section ".rodata", !dbg !48
+@read_map_val_buf = dso_local externally_initialized global [1 x [2 x [16 x i8]]] zeroinitializer, section ".data.read_map_val_buf", !dbg !50
```
ajor, line 17: These global variables aren't used in this script. Also in a number of other test cases.

The condition for "required global variable" seems to be a little off.

src/ast/async_ids.h
ryantimwilson: AsyncIDs class provides functionality to keep a running counter. This allows us to use this to assign memory slots for tuple/str/read_map_value scratch buffers, so that we make sure one usage of the buffer doesn't overwrite the buffer. See my comment here: https://github.com/bpftrace/bpftrace/pull/3456#issuecomment-2387264451

I am happy to rename this class to `Counter` or `IdGenerator` or something like that

tests/codegen/llvm/map_assign_int.ll
ajor, line 1: It seems unnecessary to use the scratch buffer for writing small values like integers. Could we set a limit on when it's used? I'm imagining that there's a performance penalty for using these scratch maps, so we should only use them when required.

src/ast/codegen_helper.cpp
```diff
@@ -0,0 +1,43 @@
+#include "codegen_helper.h"
+
+namespace bpftrace::ast {
+
+bool shouldUseWriteMapValueScratchBuffer(const AssignMapStatement &assignment)
```
ajor, line 5: > We unnecessarily allocate the write_map_value_buf for things like @ = avg(2)

To address your point, how about having this function return `false` for aggregate

ajor (commented):
> 1. Cached bounded CPU ID at beginning of probe to minimize # of jumps. Unfortunately, this caused LLVM to barf with `BPF program is too large. Processed 1000001 insn` errors due to instruction exploration space (e.g. verifier assumes value could be MAX_CPU_ID or get_snmp_processor_id so has to 2x explore every path using scratch buffers)

The `BPF program is too large. Processed 1000001 insn` error comes from the kernel, not LLVM, right?

Second question due to my lack of knowledge of the verifier: why does caching the CPU ID trigger this expanded exploration space, but fetching it before each string lookup doesn't? (Or does it, but we used to hit the 8192 branch limit before the 1000000 instruction limit?)

> 2. To fix 1, I just used `assert(snmp_processor_id <= MAX_CPU_ID)`. Unfortunately, this caused long jump errors.

Sorry, I don't understand this - what you mean exactly by adding an `assert`? An assert in the bpftrace source code won't appear in the generated bytecode and we don't have support for asserts in BPF AFAIK. 

> 3. To fix 2, I moved to CPUv4 to support long jumps. But I needed to write my own probing function similar to https://github.com/llvm/llvm-project/blob/efb583178d74b2174e8b9660b67ba7295527b09f/llvm/lib/TargetParser/Host.cpp#L464 to ensure the kernel and LLVM support V4.

In case you haven't seen it already, we have similar code in bpffeature.cpp for detecting what the kernel supports: https://github.com/bpftrace/bpftrace/blob/25376f126f200f4b2620923b7a98307f430354d6/src/bpffeature.cpp#L236-L252

-----

Unrelated point: There's a lot of unnecessary changes going on in the test cases, mostly because they're calling `print` unnecessarily (i.e. when that's not the thing they're meant to be testing). I'll go through at least some of them today and try to simplify them down so that they're not impacted by your PR.

ryantimwilson:
> Let's do the cpuid caching optimization since that will be necessary sooner or later. 

@danobi Thanks for finding this!

Converting unroll(50) -> while ($y < 50) indeed works with my CPU ID caching local patch makes this PR work with bpfsnake. So I will implement it as a commit in this PR.

However, I still see # of instructions at 400K, not 20K:
```
processed 402613 insns (limit 1000000) max_states_per_insn 19 total_states 23892 peak_states 2587 mark_read 1302
```

On bpftrace master, I actually get slightly more instructions:
```
processed 416021 insns (limit 1000000) max_states_per_insn 25 total_states 24735 peak_states 2792 mark_read 1405
```

Curious how you're getting 20K. I'm on kernel 6.9

I also tried converting unroll (12) -> while ($x < 12) to further reduce instruction count but that got me too many jumps 8193 >= limit of 8192. I get ~70K instructions there though

src/ast/async_ids.h
ajor: Ok yeah, if this doesn't represent anything "async", then the name is very misleading!

It looks like the class originally was just for async events IDs, which have to stay in sync with the corresponding code in resource_analyser.cpp and are used in bpftrace.cpp. Some comments explaining the purpose of this class would be nice.

If the class is no longer just for async events, then a name change would make sense.

src/ast/codegen_helper.cpp
ryantimwilson: Sounds good to me. Ah I see what I can do now...I can look at `assignment.expr` and see if its an aggregate function

Yea I'll add `is_aggregate_function` or something like that to unify logic in CodegenLLVM and ResourceAnalyser

tests/codegen/llvm/map_assign_int.ll
ryantimwilson: We can but then there's a penalty to pay for code complexity imho as you now have to track whether you allocate on stack vs scratch buffer (and stack allocation need lifetimes). I can look into it but I'm not super convinced its worth it, especially if we do CPU ID caching optimization to minimize instructions.

tests/codegen/llvm/variable.ll
ryantimwilson: Ah this is because Visitors.cpp recurses into the map AND expression:
```
void Visitor::visit(AssignMapStatement &assignment)
{
  Visit(*assignment.map);
  Visit(*assignment.expr);
}
```

While codegen LLVM does not recurse into the map under certain circumstances (aka only when map.key_expr is set AND not using an aggregate function):
```
  Map &map = *assignment.map;
  auto scoped_del = accept(assignment.expr);

  if (!expr_) // Some functions do the assignments themselves
    return;
```

This kinda further emphasizes the point I make in the summary that CodegenLLVM and ResourceAnalyser are starting to be very tied together. It may make sense to extract some logic out to a base class or do a 2 pass codegen algorithm.

ryantimwilson:
@ajor you can generally ignore my comment except 1 per @danobi's investigation

> Second question due to my lack of knowledge of the verifier: why does caching the CPU ID trigger this expanded exploration space, but fetching it before each string lookup doesn't? (Or does it, but we used to hit the 8192 branch limit before the 1000000 instruction limit?)

I think the reason is bpfsnake doesn't use fmt strings or tuples - what I used before. I did fairly simple load tests, not complex programs like bpfsnake.

> Sorry, I don't understand this - what you mean exactly by adding an assert? An assert in the bpftrace source code won't appear in the generated bytecode and we don't have support for asserts in BPF AFAIK.

I meant generating something like this in bytecode, similar to what previous scratch maps did:
```
if (snmp_processor_id > MAX_CPU_ID)
  return 0;
else
  // continue with rest of program
  ...
```

> In case you haven't seen it already, we have similar code in bpffeature.cpp for detecting what the kernel supports:

Sounds good, I'll consider using it. I may also upstream a LLVM patch for cpuv4 detection into "probe" since this isn't urgent now.

> Unrelated point: There's a lot of unnecessary changes going on in the test cases, mostly because they're calling print unnecessarily (i.e. when that's not the thing they're meant to be testing). I'll go through at least some of them today and try to simplify them down so that they're not impacted by your PR.

I'm confused by this comment...don't we need to print things to read the output and verify its what we expect. Or do you propse just using exit()?

ajor:
> I'm confused by this comment...don't we need to print things to read the output and verify its what we expect. Or do you propse just using exit()?

I've made the change in #3486. The codegen tests don't actually get run, so they don't need to output anything at all. They're meant to be minimal unit tests that just exercise specific parts of codegen_llvm.cpp.

ryantimwilson:
@ajor @danobi I implemented the CPU ID caching in its own PR since IMHO it was non-trivial (story of this whole task lol): https://github.com/bpftrace/bpftrace/pull/3487

Lets follow up there, then once that's merged, I can continue on this PR

tests/codegen/llvm/map_assign_int.ll
ryantimwilson: @ajor The more I think about this, the more I think you're right.

Unfortunately as we've seen this with this patch series, scratch buffers aren't free. They may use up more memory, more jumps/branches, more get_cpu_id calls, etc. While stack is limited in size but much cheaper.

So we should only use scratch space when necessary (or when the user wants it)

So I propose:
1. Create a `scratch_buffer_threshold` config where anything >= gets allocated on scratch. Anything less gets allocated on stack
2. Default it to some sane value (64?) but allow it to be configurable

I will put up yet another PR for this.

danobi: >  I'm imagining that there's a performance penalty for using these scratch maps, so we should only use them when required.

Other than extra memory used, what is the performance impact? These are direct memory reads/writes, effectively no different than the stack.

With the cpu id caching in place, I don't think there's a performance hit. Any non trivial script will need the cpuid already.

While I agree that smaller allocations should probably go on stack, this seems like something we can make a note about and revisit (you can put me on the hook for it) after the big string stuff is done. We're pretty close to seeing that be functionally complete and I don't see any obvious benefit by doing the optimization immediately.

ryantimwilson:
Not ready for review yet, code is still quite sloppy, just putting up to make sure tests work across all environments before I start refactoring

ryantimwilson:
Okay after enough iterations locally, the logic is so complex around whether we read/write from scratch buffer in CodegenAnalyser. Its not worth trying to replicate it in ResourceAnalyser, otherwise, the code will be super brittle.

We need to experiment with ways to run CodegenAnalyser twice. First time, we don't actually generate code, we just count resources e.g. when CreateMapValueScratchBuffer is called, instead we run a callback that bumps # of buffers needed and max size needed. Then the next time we actually generate the code. I will try to prototype this but also need to focus on other work, so might take me a bit.

src/ast/async_ids.h
ryantimwilson: @ajor if you don't mind, I will do the rename in a separate PR, I prefer smaller PRs and this one already has some non-trivial complexities :)

src/ast/codegen_helper.cpp
ryantimwilson: @ajor I took a different approach here and ended up using your suggestion of only use scratch buffer > 16 bytes. This ends up solving the aggregate problem mostly since aggregate maps use <= 16 byte values. See the PR description for more details.

tests/codegen/llvm/map_assign_int.ll
ryantimwilson: FYI I ended up going with @ajor's threshold strategy not due to BPF but due to trying to replicate the complexity of CodegenLLVM (aka propagation of expr_ variable) in ResourceAnalyser. As opposed to that, I implement a 16 byte threshold to use scratch buffer for map values. This ends up dealing with unnecessary allocations from lhist/hist/etc. See PR description for more details.

ryantimwilson:
@ajor @danobi @jordalgo this is ready for review. FWIW, I expect some controversy here, I made some tradeoffs between code complexity and accurate buffer accounting - I think I found a nice medium ground but let me know what you think.

ryantimwilson:
Sigh my most recent changes to make memory accounting more accurate didn't work...still failing many tests due to complexity of CodegenLLVM. I think the only option is the blunt approach which may over-allocate memory

src/ast/passes/resource_analyser.cpp
```diff
@@ -242,6 +258,14 @@ void ResourceAnalyser::visit(Call &call)
 
     resources_.skboutput_args_.emplace_back(file.str, offset.n);
     resources_.needs_perf_event_map = true;
+  } else if (call.func == "delete") {
```
jordalgo, line 261: Is this to support the old/deprecated `delete` API where the map and key are passed as one arg e.g. `delete(@mymap[1])`? If so, I don't think we should add this logic here or else heavily document that this should get removed.

The new API has a single value as the second arg and should already be accounted for (in terms of global buffer vs stack).

src/ast/passes/resource_analyser.cpp
```diff
@@ -303,6 +333,33 @@ void ResourceAnalyser::visit(For &f)
   }
 }
 
+void ResourceAnalyser::visit(AssignMapStatement &assignment)
+{
+  // Unfortunately, CodegenLLVM doesn't recurse into assignment.map
```
jordalgo, line 338: I'll be honest. I read this comment twice and I'm still not sure what's happening here - maybe I just need a brain restart.

jordalgo (commented):
Dumb question: why do we even need  a write map buffer? Shouldn't the assignment values already be allocated to the stack or global buffer depending on their type and size?

src/ast/passes/resource_analyser.cpp
ryantimwilson: I will fix this to not loop through and only look at the first arg. I just copied codegenllvm module previously

src/ast/passes/resource_analyser.cpp
ryantimwilson: 1. CodegenLLVM module traverses AST via accept() method
2. Visitor module traverse AST via visit() method
3. CodegenLLVM will traverse AssignmentMap -> accept(assignment.expr) and accept(assignment.map.key_expr) BUT will not call accept(assignment.map)
4. However, Visitor will traverse AssignmentMap -> visit(assignment.expr) and visit(assignment.map) 
5. CodegenLLVM module will do a map read lookup when traversing AST into assignment.map

Thus, to avoid over-counting read buffer, we need to do this hack above. Unfortunately, this optimization is pretty important, otherwise a bunch of codegen tests start allocating buffers needlessly. So I'll try to figure out how to clarify it.

ryantimwilson:
> Dumb question: why do we even need a write map buffer? Shouldn't the assignment values already be allocated to the stack or global buffer depending on their type and size?

@jordalgo my understanding is a few cases:
1. Tuples/strings where they are smaller than the map value size
2. Arrays/structs that may not be read into BPF memory yet
3. Delete API where we need to zero an element

There's not a lot but a smattering of smaller use cases

src/ast/codegen_helper.cpp
```diff
@@ -0,0 +1,17 @@
...
+namespace bpftrace::ast {
+
+bool needAssignMapStatementAllocation(const AssignMapStatement &assignment)
+{
+  const auto &map = *assignment.map;
+  const auto &expr_type = assignment.expr->type;
+  if (shouldBeInBpfMemoryAlready(expr_type)) {
+    return !expr_type.IsSameSizeRecursive(map.type);
+  } else if (map.type.IsRecordTy() || map.type.IsArrayTy()) {
+    return !expr_type.is_internal;
```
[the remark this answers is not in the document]
ryantimwilson: This method is refactoring and transitioning the usage of `is_internal` from CodegenLLVM::visit(AssigmentMap&) to this function. So shouldn't be a new usage altogether

src/ast/passes/resource_analyser.cpp
ryantimwilson: @jordalgo updated the comment with diagrams, let me know if its not clear still or you have suggestions

src/ast/passes/resource_analyser.cpp
ryantimwilson: Fixed in latest

src/ast/codegen_helper.cpp
```diff
@@ -0,0 +1,17 @@
+#include "codegen_helper.h"
+
+namespace bpftrace::ast {
+
+bool needAssignMapStatementAllocation(const AssignMapStatement &assignment)
+{
+  const auto &map = *assignment.map;
+  const auto &expr_type = assignment.expr->type;
+  if (shouldBeInBpfMemoryAlready(expr_type)) {
+    return !expr_type.IsSameSizeRecursive(map.type);
```
jordalgo, line 10: I realized what was bothering me about this check. It only applies in the case of the resource_analyser when you're trying to get the max write value size. However when you call this from `AssignMapStatement` in codegen you shouldn't be checking this e.g. this returns false:
```
BEGIN { @a[1] = ("hellotherelongstr", 5); @a[2] = ("hellotherelongstr", 6); exit(); }
```
Further more I think some of the types in `shouldBeInBpfMemoryAlready` don't need an additional scratch buffer e.g. tuples and strings.

src/ast/passes/resource_analyser.cpp
jordalgo: Is the map.type already updated at this point in the case the second arg is larger then what's currently stored in the map? Example:
```
@a[1, "by"] = 1;
delete(@a, (1, "hellolongerstr"));
```

src/ast/passes/resource_analyser.cpp
```diff
@@ -303,6 +332,43 @@ void ResourceAnalyser::visit(For &f)
   }
 }
 
+void ResourceAnalyser::visit(AssignMapStatement &assignment)
+{
+  /* CodegenLLVM traverses the AST like:
```
jordalgo, line 337: Thank you! Much clearer comment. Love it.

tests/codegen/scratch_buffer.cpp
```diff
@@ -65,6 +65,20 @@ TEST(codegen, str_stack)
                                LARGE_ON_STACK_LIMIT);
 }
 
+TEST(codegen, map_value_scratch_buf)
```
jordalgo, line 68: Maybe we should add some tests for some different types as they seem to determine if they get a buffer or not.

jordalgo (changes requested):

src/ast/codegen_helper.cpp
ryantimwilson: Hmmm I was just copying the existing logic in CodegenLLVM::AssignMapStatement. I'm fine to change it but admittedly I don't understand the effects well.

Is there an urgency to do this in this PR? Could we do this in another PR? I'm hoping to minimize side effects

ryantimwilson: Discussed this offline with @jordalgo - there was confusion on what this function did. Will clarify with comments why this function is extracted out and what its purpose is

tests/codegen/scratch_buffer.cpp
ryantimwilson: Added tuple (with different sizes to trigger the !isSameSizeRecursive condition) and ints. Let me know if you'd like more types

src/ast/codegen_helper.cpp
jordalgo: Yeah, my bad. I wasn't reading close enough.

tests/codegen/scratch_buffer.cpp
```diff
@@ -65,6 +65,36 @@ TEST(codegen, str_stack)
...
+}
+
+TEST(codegen, map_value_int_stack)
+{
+  test_stack_or_scratch_buffer("kprobe:f { @x = 1; @y = @x }",
+                               NAME,
+                               LARGE_ON_STACK_LIMIT);
+}
+
+TEST(codegen, map_value_tuple_scratch_buf)
```
jordalgo, line 82: nit: Maybe add a comment about the fact that these need an allocation and maybe global scratch buf because these are different sized tuples.

ryantimwilson: Fixed in latest

src/ast/codegen_helper.cpp
```diff
@@ -0,0 +1,25 @@
+#include "codegen_helper.h"
+
+namespace bpftrace::ast {
+
+/* needAssignMapStatementAllocation determines if a map assignment requires
```
jordalgo, line 5: Hope you don't mind. Here is a rewrite of this comment - more for the sanity of my future self:
"needAssignMapStatementAllocation determines if a map assignment requires a new memory allocation. This happens only in a few cases e.g. if there are two map assignments of tuples of different sizes (`@x = ("xxx", 1); @x = ("xxxxxxx", 1);` which requires a 0 memsetting and copying of each tuple element into the new allocation before calling `bpf_map_update_elem`.

Another case when we need an allocation is for a external struct e.g. `$v = (struct task_struct *)arg1; @ = *$v;`. 

Most cases we can reuse existing BPF memory and not create a new allocation.
"

I wouldn't change your 2nd paragraph "Note this function does NOT determine..."

jordalgo (approved):
💥

src/ast/codegen_helper.cpp
ryantimwilson: Fixed!

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2382,53 +2384,41 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
...
+  const auto self_alloca = needAssignMapStatementAllocation(assignment);
+  val = self_alloca ? b_.CreatePointerCast(
+                          b_.CreateWriteMapValueAllocation(map.type,
+                                                           map.ident + "_val",
+                                                           assignment.loc),
+                          b_.GetType(map.type)->getPointerTo())
+                    : expr;
   if (shouldBeInBpfMemoryAlready(expr_type)) {
     if (!expr_type.IsSameSizeRecursive(map.type)) {
-      val = b_.CreateAllocaBPF(map.type, map.ident + "_val");
```
danobi, line 2387: Nice. I feel like the code here was simplified a lot

danobi (commented):
small nits but lgtm otherwise

danobi (approved):

PR #3484, opened 2024-10-03
Remove array syntax need for string/buffer/inet

jordalgo:
These types can, and should be, used without an explicit size. The size will get inferred in later AST passes.

##### Checklist

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -212,6 +212,15 @@ type:
                     };
                     $$ = type_map[$1];
                 }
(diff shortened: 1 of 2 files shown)


jordalgo:
Note: a lot of this type logic will get removed from the lexer/parser but I still want to support strings without specifying size before landing the `let` syntax PR: https://github.com/bpftrace/bpftrace/pull/3461

ajor (approved):

PR #3485, opened 2024-10-04
Add 'has_key' builtin for maps

jordalgo:
This is so we can check if a map has a key
instead of relying on the wrong way by seeing
if the value is 0.

```
kprobe:dummy {
  @associative[1,2] = 1;

  if (!has_key(@associative, (1,3))) {
    print(("bye"));
  }
}
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to
   - [#3401](https://github.com/bpftrace/bpftrace/pull/3401)
 - Allow tuples to be used as map keys
   - [#3390](https://github.com/bpftrace/bpftrace/pull/3390/)
+- Add `has_key` function for maps
+  - [#3358](https://github.com/bpftrace/bpftrace/pull/3358)
 #### Changed
 - Merge output into `stdout` when `-lv`
   - [#3383](https://github.com/bpftrace/bpftrace/pull/3383)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -989,7 +989,7 @@ kprobe:do_nanosleep {
   @start[tid] = nsecs;
 }
 
-kretprobe:do_nanosleep /@start[tid] != 0/ {
+kretprobe:do_nanosleep /has_key(@start, tid)/ {
   printf("slept for %d ms\n", (nsecs - @start[tid]) / 1000000);
   delete(@start, tid);
 }
@@ -2371,6 +2371,10 @@ See <<Advanced Topics>> for more information on <<Map Printing>>.
 | Delete a single key from a map.
 | Sync
 
+| <<map-functions-has_key, `has_key(map m, mapkey k)`>>
+| Return true (1) if the key exists in this map. Otherwise return false (0).
+| Sync
+
 | <<map-functions-hist, `hist(int64 n[, int k])`>>
 | Create a log2 histogram of n using buckets per power of 2, 0 <= k <= 5, defaults to 0.
 | Sync
@@ -2509,6 +2513,34 @@ kprobe:dummy {
   // deprecated but ok
   delete(@single["hello"]);
   delete(@associative[1, 2]);
+}
+```
+
+[#map-functions-has_key]
+=== has_key
+
+.variants
+* `int has_key(map m, mapkey k)`
+
+Return true (1) if the key exists in this map.
+Otherwise return false (0).
+Error if called with a map that has no keys (aka scalar map).
+Return value can also be used for scratch variables and map keys/values.
+
+```
+kprobe:dummy {
+  @associative[1,2] = 1;
+  if (!has_key(@associative, (1,3))) { // ok
+    print(("bye"));
+  }
+
+  @scalar = 1;
+  if (has_key(@scalar)) { // error
+    print(("hello"));
+  }
+  
+  $a = has_key(@associative, (1,2)); // ok
+  @b[has_key(@associative, (1,2))] = has_key(@associative, (1,2)); // ok
 }
 ```
 
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -768,6 +768,19 @@ void CodegenLLVM::visit(Call &call)
       }
     }
     expr_ = nullptr;
+  } else if (call.func == "has_key") {
+    auto &arg = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg);
+    auto [key, scoped_key_deleter] = getMapKey(map, call.vargs.at(1));
+
+    CallInst *lookup = b_.CreateMapLookup(map, key);
+    expr_ = b_.CreateICmpNE(b_.CreateIntCast(lookup, b_.GET_PTR_TY(), true),
+                            b_.GetNull(),
+                            "has_key");
+#if LLVM_VERSION_MAJOR <= 14
+    // Got to cast to the correct type for earlier versions of LLVM
+    expr_ = b_.CreateIntCast(expr_, b_.getInt8Ty(), false);
+#endif
   } else if (call.func == "str") {
     uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
     // Largest read we'll allow = our global string buffer size
@@ -2941,24 +2954,30 @@ int CodegenLLVM::getNextIndexForProbe()
 std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
     Map &map)
 {
+  return getMapKey(map, map.key_expr);
+}
+
+std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
+    Map &map,
+    Expression *key_expr)
+{
   Value *key;
   bool alloca_created_here = true;
-  if (map.key_expr) {
-    Expression *expr = map.key_expr;
-    auto scoped_del = accept(expr);
-    if (inBpfMemory(expr->type)) {
+  if (key_expr) {
+    auto scoped_del = accept(key_expr);
+    if (inBpfMemory(key_expr->type)) {
       auto &key_type = map.key_type;
-      if (!expr->type.IsSameSizeRecursive(key_type)) {
+      if (!key_expr->type.IsSameSizeRecursive(key_type)) {
         key = b_.CreateAllocaBPF(key_type, map.ident + "_key");
         b_.CreateMemsetBPF(key, b_.getInt8(0), key_type.GetSize());
-        if (expr->type.IsTupleTy()) {
-          createTupleCopy(expr->type, key_type, key, expr_);
-        } else if (expr->type.IsStringTy()) {
-          b_.CreateMemcpyBPF(key, expr_, expr->type.GetSize());
+        if (key_expr->type.IsTupleTy()) {
+          createTupleCopy(key_expr->type, key_type, key, expr_);
+        } else if (key_expr->type.IsStringTy()) {
+          b_.CreateMemcpyBPF(key, expr_, key_expr->type.GetSize());
         } else {
           LOG(BUG) << "Type size mismatch. Key Type Size: "
                    << key_type.GetSize()
-                   << " Expression Type Size: " << expr->type.GetSize();
+                   << " Expression Type Size: " << key_expr->type.GetSize();
         }
       } else {
         key = expr_;
@@ -2973,15 +2992,16 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
       key = b_.CreateAllocaBPF(map.key_type, map.ident + "_key");
       // Integers are always stored as 64-bit in map keys
       b_.CreateStore(
-          b_.CreateIntCast(expr_, b_.getInt64Ty(), expr->type.IsSigned()), key);
+          b_.CreateIntCast(expr_, b_.getInt64Ty(), key_expr->type.IsSigned()),
+          key);
     } else {
-      key = b_.CreateAllocaBPF(expr->type, map.ident + "_key");
-      if (expr->type.IsArrayTy() || expr->type.IsRecordTy()) {
+      key = b_.CreateAllocaBPF(key_expr->type, map.ident + "_key");
+      if (key_expr->type.IsArrayTy() || key_expr->type.IsRecordTy()) {
         // We need to read the entire array/struct and save it
-        b_.CreateProbeRead(ctx_, key, expr->type, expr_, expr->loc);
+        b_.CreateProbeRead(ctx_, key, key_expr->type, expr_, key_expr->loc);
       } else {
         b_.CreateStore(
-            b_.CreateIntCast(expr_, b_.getInt64Ty(), expr->type.IsSigned()),
+            b_.CreateIntCast(expr_, b_.getInt64Ty(), key_expr->type.IsSigned()),
             b_.CreatePointerCast(key, expr_->getType()->getPointerTo()));
       }
     }
--- a/src/ast/passes/codegen_llvm.h
+++ b/src/ast/passes/codegen_llvm.h
@@ -173,6 +173,9 @@ private:
 
   [[nodiscard]] ScopedExprDeleter accept(Node *node);
   [[nodiscard]] std::tuple<Value *, ScopedExprDeleter> getMapKey(Map &map);
+  [[nodiscard]] std::tuple<Value *, ScopedExprDeleter> getMapKey(
+      Map &map,
+      Expression *key_expr);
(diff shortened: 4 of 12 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to
   - [#3401](https://github.com/bpftrace/bpftrace/pull/3401)
 - Allow tuples to be used as map keys
   - [#3390](https://github.com/bpftrace/bpftrace/pull/3390/)
+- Add `has_key` builtin for maps
```
jordalgo, line 26: note to self:
```suggestion
- Add `has_key` function for maps
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -771,6 +771,51 @@ void CodegenLLVM::visit(Call &call)
...
     }
     expr_ = nullptr;
+  } else if (call.func == "has_key") {
+    auto &arg = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg);
+    auto [key, scoped_key_deleter] = getMapKey(map, call.vargs.at(1));
+
+    CallInst *lookup = b_.CreateMapLookup(map, key);
+
+    AllocaInst *has_key = b_.CreateAllocaBPF(b_.getInt8Ty(), "has_key");
```
danobi, line 781: CreateBool() in types.cpp defines a boolean as a 1 bit integer. This should probably be a `b_.getInt1Ty()` right?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -771,6 +771,51 @@ void CodegenLLVM::visit(Call &call)
...
+        b_.CreateIntCast(lookup, b_.GET_PTR_TY(), true),
+        b_.GetNull(),
+        "lookup_cond");
+    b_.CreateCondBr(lookup_condition,
+                    lookup_success_block,
+                    lookup_failure_block);
+
+    b_.SetInsertPoint(lookup_success_block);
+
+    b_.CreateStore(b_.getInt8(1), has_key);
```
danobi, line 804: Likewise, `b_.getTrue()` here and `b_.getFalse()` below

src/ast/passes/codegen_llvm.cpp
danobi: Also, use a phi node here instead of a stack temporary. See https://github.com/bpftrace/bpftrace/commit/caed2944717f985d6295a84e3ff166bc7038e76b for example

danobi (commented):
need to run soon, so skipped over semantic analysis bits

src/ast/passes/codegen_llvm.cpp
jordalgo: Ok, I tried. Let me know if I did this right.

danobi:
(btw, what do you think about just `has()` over `has_key()`?

src/ast/passes/codegen_llvm.h
```diff
@@ -172,7 +172,9 @@ class CodegenLLVM : public Visitor {
                  FunctionType *func_type);
 
   [[nodiscard]] ScopedExprDeleter accept(Node *node);
-  [[nodiscard]] std::tuple<Value *, ScopedExprDeleter> getMapKey(Map &map);
+  [[nodiscard]] std::tuple<Value *, ScopedExprDeleter> getMapKey(
+      Map &map,
+      Expression *key_expr);
```
viktormalik, line 177: Could we rather overload `getMapKey` to have both one- and two-parameter variants? It looks like `getMapKey(map, map.map_expr)` is used quite often.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -768,6 +768,46 @@ void CodegenLLVM::visit(Call &call)
...
+    b_.SetInsertPoint(lookup_failure_block);
+
+    b_.CreateBr(lookup_merge_block);
+
+    b_.SetInsertPoint(lookup_merge_block);
+
+    auto phi = b_.CreatePHI(b_.getInt1Ty(), 2, "result");
+    phi->addIncoming(b_.getTrue(), lookup_success_block);
+    phi->addIncoming(b_.getFalse(), lookup_failure_block);
+    expr_ = phi;
```
viktormalik, line 810: Why don't just do
```suggestion
    expr_ = b_.CreateICmpNE(
        b_.CreateIntCast(lookup, b_.GET_PTR_TY(), true),
        b_.GetNull(),
        "has_key");
```
and skip all the branching?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -769,6 +769,31 @@ void SemanticAnalyser::visit(Call &call)
...
+      Map &map = static_cast<Map &>(arg0);
+      auto *mapkey = get_map_key_type(map);
+      if (mapkey) {
+        if (mapkey->IsNoneTy()) {
+          LOG(ERROR, arg0.loc, err_)
+              << "has_key() only accepts maps that have keys to iterate over";
+        } else {
+          auto &arg1 = *call.vargs.at(1);
+          SizedType new_key_type = create_key_type(arg1.type, arg1.loc);
+          update_key_type(*mapkey, new_key_type, map.ident, arg1.loc);
```
viktormalik, line 789: Do we really want to update the stored key type based on the argument used in the `has_map` call?

jordalgo:
> (btw, what do you think about just has() over has_key()?)

Eh. Seems to ambiguous for me. has... what? has value? has a cool name?

src/ast/passes/semantic_analyser.cpp
jordalgo: We need to do this for string resizing e.g. we don't want this to fail `@a["hi"] = 1; if (has_key(@a, "longerstr"))`

viktormalik: Ok, makes sense.

src/ast/passes/codegen_llvm.cpp
jordalgo: Cause that would be much simpler and straightforward 🤕

src/ast/passes/codegen_llvm.cpp
```diff
@@ -768,6 +768,15 @@ void CodegenLLVM::visit(Call &call)
...
     expr_ = nullptr;
+  } else if (call.func == "has_key") {
+    auto &arg = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg);
+    auto [key, scoped_key_deleter] = getMapKey(map, call.vargs.at(1));
+
+    CallInst *lookup = b_.CreateMapLookup(map, key);
+    expr_ = b_.CreateICmpNE(b_.CreateIntCast(lookup, b_.GET_PTR_TY(), true),
+                            b_.GetNull(),
+                            "lookup_cond");
```
viktormalik, line 779: Nit: a different name would make the expression easier to spot in codegen
```suggestion
                            "has_key");
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3553,60 +3601,45 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
   return new_key_type;
 }
 
-void SemanticAnalyser::update_key_type(const Map &map,
-                                       const SizedType &new_key_type)
+void SemanticAnalyser::update_key_type(SizedType &current_key_type,
```
viktormalik, line 3604: This method deserves a comment. IIUC, originally, it not only updated `current_key_type` but also inserted entries into `map_key_` (which is a bit confusing tbh). Documenting the purpose of the method would make it easier to understand when it should be used.

tests/semantic_analyser.cpp
```diff
@@ -1067,6 +1068,61 @@ TEST(semantic_analyser, call_len)
...
+{
+  test("kprobe:f { @x[1] = 0; if (has_key(@x, 1)) {} }");
+  test("kprobe:f { @x[1, 2] = 0; if (has_key(@x, (3, 4))) {} }");
+  test("kprobe:f { @x[1, (int8)2] = 0; if (has_key(@x, (3, 4))) {} }");
+  test(R"(kprobe:f { @x[1, "hi"] = 0; if (has_key(@x, (2, "bye"))) {} })");
+  test(
+      R"(kprobe:f { @x[1, "hi"] = 0; if (has_key(@x, (2, "longerstr"))) {} })");
+  test(
+      R"(kprobe:f { @x[1, "longerstr"] = 0; if (has_key(@x, (2, "hi"))) {} })");
+  test("kprobe:f { @x[1, 2] = 0; $a = (3, 4); if (has_key(@x, $a)) {} }");
```
viktormalik, line 1081: Maybe also a unit test for passing the key via a map?
```suggestion
  test("kprobe:f { @x[1, 2] = 0; @a = (3, 4); if (has_key(@x, @a)) {} }");
```

src/ast/passes/codegen_llvm.cpp
jordalgo: Good catch.

src/ast/passes/semantic_analyser.cpp
jordalgo: I split this out into two separate methods to make it more understandable, which resulted in some reduced code.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3553,60 +3603,56 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
   return new_key_type;
 }
 
-void SemanticAnalyser::update_key_type(const Map &map,
-                                       const SizedType &new_key_type)
+void SemanticAnalyser::maybe_update_current_key(SizedType &current_key_type,
```
viktormalik, line 3606: Nit: is the `maybe_` prefix necessary? It's sort of expected that the key won't be updated if not necessary.

I'd even go with `update_key` or `update_map_key`.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3553,60 +3603,56 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
...
+                                                const SizedType &new_key_type)
 {
-  if (map.skip_key_validation) {
+  if (current_key_type.IsSameType(new_key_type) &&
+      (current_key_type.IsStringTy() || current_key_type.IsTupleTy())) {
+    update_string_size(current_key_type, new_key_type);
+  }
+}
+
+void SemanticAnalyser::validate_new_key(SizedType &current_key_type,
```
viktormalik, line 3615:
```suggestion
void SemanticAnalyser::validate_new_key(const SizedType &current_key_type,
```

viktormalik (commented):
Two more small nits ;-)

src/ast/passes/semantic_analyser.cpp
jordalgo: Pretty bike-sheddy 🖌️ 🚲 🏠  but I would argue my function name (though more verbose) is the most accurate/descriptive.

viktormalik: No problem, I'm certainly not going to fight for it :grinning: Feel free to keep it as it is.

jordalgo: Ok fine, I'll get rid of the `maybe`.

tests/runtime/basic
```diff
@@ -300,3 +300,23 @@ NAME print large int
...
+EXPECT_NONE ok
+TIMEOUT 1
+
+NAME has_key complex tuple
+PROG BEGIN { @a[1, ("hello", (int8)5)] = 0; if (has_key(@a, (1, ("hello", 5)))) { printf("ok\n"); exit(); } }
+EXPECT ok
+TIMEOUT 1
+
+NAME has_key map value
+PROG BEGIN { @g = 2; @a[2] = 0; if (has_key(@a, @g)) { printf("ok\n"); exit(); } }
```
ajor, line 320: Can the `exit()` calls go outside of the if-block? Some of these will be relying on timeouts to end right now

Also the `TIMEOUT`s aren't required any more - I wouldn't include them unless there's a specific reason you want "TIMEOUT 1"

tests/runtime/variable
```diff
@@ -40,7 +40,7 @@ EXPECT openflags: 64
 AFTER ./testprogs/syscall openat
 
 NAME tracepoint arg casts in predicates
-RUN {{BPFTRACE}} -e 'tracepoint:syscalls:sys_enter_wait4 /args.ru/ { @ru[tid] = args.ru; } tracepoint:syscalls:sys_exit_wait4 /@ru[tid]/ { @++; exit(); }' -c ./testprogs/wait4_ru
+RUN {{BPFTRACE}} -e 'tracepoint:syscalls:sys_enter_wait4 /args.ru/ { @ru[tid] = args.ru; } tracepoint:syscalls:sys_exit_wait4 /has_key(@ru, tid)/ { @++; exit(); }' -c ./testprogs/wait4_ru
```
ajor, line 43:
```suggestion
RUN {{BPFTRACE}} -e 'tracepoint:syscalls:sys_enter_wait4 /args.ru/ { @ru[tid] = args.ru; } tracepoint:syscalls:sys_exit_wait4 /@ru[tid]/ { @++; exit(); }' -c ./testprogs/wait4_ru
```

This change looks wrong to me. The test is specifically checking that pointer values (such as `@ru[tid]`) can be used in predicates

Also, we don't need to swap out instances of this pattern everywhere - it's not going to be deprecated and isn't incorrect.

tests/semantic_analyser.cpp
```diff
@@ -1067,6 +1068,62 @@ TEST(semantic_analyser, call_len)
...
+  test_error("kprobe:f { @x[1] = 1;  if (has_key(@x[1], 1)) {} }",
+             R"(
+stdin:1:27-41: ERROR: has_key() expects the first argument to be a map. Not a map value expression.
+kprobe:f { @x[1] = 1;  if (has_key(@x[1], 1)) {} }
+                          ~~~~~~~~~~~~~~
+)");
+
+  test_error("kprobe:f { @x = 1;  if (has_key(@x, 1)) {} }",
+             R"(
+stdin:1:24-35: ERROR: has_key() only accepts maps that have keys to iterate over
```
ajor, line 1100: "maps that have keys to iterate over" is odd phrasing - we're not going to iterate over any keys with this function

man/adoc/bpftrace.adoc
```diff
@@ -2510,6 +2514,30 @@ kprobe:dummy {
...
+* `has_key(map m, mapkey k)`
+
+Return true if the key exists in this map.
+Otherwise return false.
+Error if called with a map that has no keys (aka scalar map).
+
+```
+kprobe:dummy {
+  @associative[1,2] = 1;
+  if (!has_key(@associative[1,3])) { // ok
```
ajor, line 2530:
```suggestion
  if (!has_key(@associative, (1,3))) { // ok
```

ajor (commented):
Since `has_key` returns a bool, assign it to a map or variable should work:
```
# bpftrace -e 'BEGIN { @a[0] = 1; @x = has_key(@a, 0); }'
stdin:1:25-39: ERROR: has_key() should not be used in an assignment or as a map key
BEGIN { @a[0] = 1; @x = has_key(@a, 0); }
                        ~~~~~~~~~~~~~~
```

Printing its value should work too:
```
# bpftrace -e 'BEGIN { @a[0] = 1; print(has_key(@a, 0)); }'
Attaching 1 probe...
ERROR: Error loading BPF program for BEGIN_1. Use -v for full kernel error log.
```

tests/runtime/variable
jordalgo: oops, got a little replacement happy here. Will change it back.

tests/semantic_analyser.cpp
jordalgo: copy 🍝

viktormalik (approved):

viktormalik:
Nice!

jordalgo:
@ajor This now supports assigning the return value of `has_key` to a variable, map value, and map key.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3553,60 +3606,56 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
...
-            << "trying to access with no arguments when map expects arguments: "
-               "'"
-            << key->second << "'";
+  bool valid = true;
+  if (current_key_type.IsSameType(new_key_type)) {
+    if (current_key_type.IsTupleTy() || current_key_type.IsIntegerTy() ||
+        current_key_type.IsStringTy()) {
+      // This should always be true as map keys default to 64 bits
+      // but keep this here just in case we add larger ints and need to
+      // update the map int logic
```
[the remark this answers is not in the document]
jordalgo: Good call. Will update and will mention strings and tuples.

jordalgo:
@viktormalik @ajor - Thanks for the review!

PR #3487, opened 2024-10-04
Cache bounded CPU ID in register for global scratch buffers

ryantimwilson:
This PR implements CPU ID caching based on discussion in https://github.com/bpftrace/bpftrace/pull/3483

Background: When using scratch buffers, we need to fetch the CPU ID via BPF helper bpf_get_smp_processor_id to find out which slot the CPU should use. That's all well and good except on older kernels (e.g 5.2) where the BPF verifier is not smart enough to bound bpf_get_smp_processor_id <= MAX_CPU_ID. So the verifier will throw errors unless we explicitly bound bpf_get_smp_processor_id. To do that, we use LLVM's select instruction which essentially does `min(bpf_get_smp_processor_id, MAX_CPU_ID)`. 

Unfortunately, if we do enough scratch lookups, this causes too many jumps >= 8192 kernel limit. This can be seen when migrating map values to scratch buffers in the above PR for bpfsnake.

To avoid this problem, we cache the bounded CPU ID in a register at the beginning of each probe, subprogram and bpf_for_each_map_cb. We need to do it at function level since registers are changed upon calling functions. Luckily, there aren't too many other than probes/subprogs and map callbacks, so this should be fairly future proof. Note we cannot cache the CPU ID in memory since memory is shared by all CPUs.

To deal with for loop nesting, we need to implement a stack s.t. we reference the correct CPU ID on the top of the stack (aka within the function call)

I tested this on 5.2 and 6.9 and verified standard test cases and stress test cases for tuples and fmt strings still pass.

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2216,18 +2216,17 @@ void CodegenLLVM::compareStructure(SizedType &our_type, llvm::Type *llvm_type)
  */
 Value *CodegenLLVM::createTuple(
     const SizedType &tuple_type,
-    const std::vector<std::pair<llvm::Value *, const location *>> &vals,
-    const location &loc)
+    const std::vector<std::pair<llvm::Value *, const location *>> &vals)
 {
   auto tuple_ty = b_.GetType(tuple_type);
   size_t tuple_size = datalayout().getTypeAllocSize(tuple_ty);
   auto buf = b_.CreatePointerCast(
-      b_.CreateTupleScratchBuffer(loc, async_ids_.tuple()),
+      b_.CreateTupleScratchBuffer(cpu_id_stack_.top(), async_ids_.tuple()),
       tuple_ty->getPointerTo());
   b_.CreateMemsetBPF(buf, b_.getInt8(0), tuple_size);
 
   for (size_t i = 0; i < vals.size(); ++i) {
-    auto [val, vloc] = vals[i];
+    auto [val, loc] = vals[i];
     SizedType &type = tuple_type.GetField(i).type;
 
     Value *dst = b_.CreateGEP(tuple_ty,
@@ -2237,7 +2236,7 @@ Value *CodegenLLVM::createTuple(
     if (inBpfMemory(type))
       b_.CreateMemcpyBPF(dst, val, type.GetSize());
     else if (type.IsArrayTy() || type.IsRecordTy())
-      b_.CreateProbeRead(ctx_, dst, type, val, *vloc);
+      b_.CreateProbeRead(ctx_, dst, type, val, *loc);
     else
       b_.CreateStore(val, dst);
   }
@@ -2291,7 +2290,7 @@ void CodegenLLVM::visit(Tuple &tuple)
     scoped_dels.emplace_back(accept(elem));
     vals.push_back({ expr_, &elem->loc });
   }
-  auto buf = createTuple(tuple.type, vals, tuple.loc);
+  auto buf = createTuple(tuple.type, vals);
 
   // No need to end buf lifetime since tuple allocation is done via scratch maps
   // and not on stack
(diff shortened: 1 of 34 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -2635,6 +2634,35 @@ void CodegenLLVM::visit(AttachPoint &)
...
+// BPF verifier on older kernels.
+//
+// However, for programs that do many scratch buffer lookups, this causes
+// many jumps which means a complex program may be rejected by the BPF
+// verifier for >= 8192 jumps (current limit).
+//
+// To minimize jumps, we bound the CPU ID at the beginning of each function
+// and cache the value in a register. This ensures we only add one
+// additional jump per function. Note we cannot cache this value in memory
+// since memory is shared between all CPUs
```
danobi, line 2647: > cache this value in memory

What do you mean by memory?

tests/codegen/llvm/avg_cast_loop.ll
```diff
@@ -24,31 +24,35 @@ define i64 @kprobe_f_1(ptr %0) section "s_kprobe_f_1" !dbg !72 {
 entry:
   %avg_struct = alloca %avg_stas_val, align 8
   %"@x_key" = alloca i64, align 8
+  %get_cpu_id = call i64 inttoptr (i64 8 to ptr)()
+  %1 = load i64, ptr @max_cpu_id, align 8
+  %2 = icmp ule i64 %get_cpu_id, %1
+  %3 = select i1 %2, i64 %get_cpu_id, i64 %1
```
danobi, line 30: It's a bit unfortunate this is still generated even though cpuid is not used from main prog. It's being generated in all progs b/c the map cb needs it. I'm not sure it's worth the complexity to fix this case, as basically all non-trivial scripts will end up using printf() or tuples() or map accesses (etc) in main prog. Just calling it out here.

danobi (commented):
lgtm overall

tests/codegen/llvm/avg_cast_loop.ll
ryantimwilson: Yea I saw that too.

We could implement a Probe/Subprog -> "does this need cpu_id caching" in ResourceAnalyser. But I agree...feels like a lot of complexity.

The good news is LLVM optimizer realizes this and only keeps the get_cpu_id instruction in the optimized codegen (because it doesn't know if the BPF helper has side effects). But this instruction is cheap.
```
define noundef i64 @kprobe_f_1(ptr nocapture readnone %0) local_unnamed_addr section "s_kprobe_f_1" !dbg !72 {
entry:
  %avg_struct = alloca %avg_stas_val, align 8
  %"@x_key" = alloca i64, align 8
  %get_cpu_id = tail call i64 inttoptr (i64 8 to ptr)()
  call void @llvm.lifetime.start.p0(i64 -1, ptr nonnull %"@x_key")
  store i64 1, ptr %"@x_key", align 8
  %lookup_elem = call ptr inttoptr (i64 1 to ptr)(ptr nonnull @AT_x, ptr nonnull %"@x_key")
  %lookup_cond.not = icmp eq ptr %lookup_elem, null
  br i1 %lookup_cond.not, label %lookup_failure, label %lookup_success
```

src/ast/passes/codegen_llvm.cpp
ryantimwilson: To be clear, I mean we can't call `CreateStore()` and save it to some location in memory. Because memory of the BPF script is shared between all CPUs (at least AFAIK), so we can't have CPU A write the bounded CPU ID and CPU B load the bounded CPU ID because it'd yield the wrong result.

Basically saying we have to keep it in a register

danobi: I see. Right, unless it was a percpu array. In which case it's recursive. Maybe replace "memory" with "global variable"? I think that would make it less confusing

danobi (approved):

src/ast/passes/codegen_llvm.cpp
ryantimwilson: Fixed in latest version!

jordalgo:
Is there any perf issue in dedicating a register for the CPU ID - are we more likely to run out of these?

danobi:
This doesn’t necessarily dedicate a register - llvm is allowed to spill it to stack

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2635,6 +2634,37 @@ void CodegenLLVM::visit(AttachPoint &)
...
+          globalvars::GlobalVar::MAX_CPU_ID)) {
+    return nullptr;
+  }
+
+  auto cpu_id = b_.CreateGetCpuId(loc);
+  auto max = b_.CreateLoad(b_.getInt64Ty(),
+                           module_->getGlobalVariable(to_string(
+                               bpftrace::globalvars::GlobalVar::MAX_CPU_ID)));
+  auto cmp = b_.CreateICmp(CmpInst::ICMP_ULE, cpu_id, max);
+  return b_.CreateSelect(cmp, cpu_id, max);
```
ajor, line 2665: Why get rid of the descriptive names that were used for these in IRBuilder? They make the generated IR a lot easier to read (and it seems that 99% of the expected results changed in this PR are due to the names disappearing)

ryantimwilson: @ajor the reason I got rid of this is when I hit the a bug re: for loops that looked like `Instruction does not dominate all use cases`: https://www.reddit.com/r/LLVM/comments/vd5urh/runtimeerror_instruction_does_not_dominate_all/ (it was due to not using stack + for loop)

With named variables, it was very hard to debug because all the names looked the same. When I removed the names, it was obvious that function call 1 was trying to use a register from function call 2.

I will add them back in another PR, sorry for the confusion

ryantimwilson:
> This doesn’t necessarily dedicate a register - llvm is allowed to spill it to stack

@danobi Ah interesting. I suppose the reason this works is bpftrace programs/probes are non-preemptible so another CPU won't read the stack variable while the probe is executing.

We would probably have to revisit this if bpftrace supports preemptible programs

danobi:
> > This doesn’t necessarily dedicate a register - llvm is allowed to spill it to stack
> 
> @danobi Ah interesting. I suppose the reason this works is bpftrace programs/probes are non-preemptible so another CPU won't read the stack variable while the probe is executing.
> 
> We would probably have to revisit this if bpftrace supports preemptible programs

Each bpf program gets its own stack (provided by kernel). It's unique to each invocation of the program. So no worries about preemption. The reason we are going through the whole exercise of off-stack allocations is b/c the kernel-provided stack is too small (512B).

PR #3489, opened 2024-10-06
Add `-d dis` debug output option

danobi:
See commits for more details.

##### Checklist


danobi:
This is the backport: https://pastes.dxuuu.xyz/whuenq

src/bpfbytecode.cpp
```diff
@@ -34,6 +34,7 @@ static std::optional<std::string> get_global_var_section_name(
 }
 
 BpfBytecode::BpfBytecode(std::span<const std::byte> elf)
```
jordalgo, line 36: Would it be simpler if the constructor took a vector arg and we just moved it into `elf_` ?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -45,6 +45,10 @@
 #include "types.h"
 #include "usdt.h"
 
+#define _BT_DISASM_TMPFILE ".temp_bt_disassemble"
+static constexpr auto BT_DISASM_TMPFILE = _BT_DISASM_TMPFILE;
+static constexpr auto BT_DISASM_CMD = "llvm-objdump -d " _BT_DISASM_TMPFILE;
```
jordalgo, line 50: Do we need to check if this is available on the host and perhaps generate a nicer error if it's not?

jordalgo (approved):
Nice catch. LGMT - just a few small questions.

src/bpfbytecode.cpp
danobi: Doesn't look like it. Looks like llvm::SmallVector doesn't easy convert to std::vector. And in aot::load() it's working with pointer + len, so span is more natural there as well. So as-is is probably simple enough

src/ast/passes/codegen_llvm.cpp
danobi: Error seems reasonable enough IMO:
```
sudo PATH= ./build/src/bpftrace -e 'BEGIN { print(*curtask) }' -d dis -dcodegen-opt --test codegen 2>&1 | tail
!49 = distinct !DISubprogram(name: "BEGIN_1", linkageName: "BEGIN_1", scope: !2, file: !2, type: !50, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !46, retainedNodes: !53)
!50 = !DISubroutineType(types: !51)
!51 = !{!35, !52}
!52 = !DIDerivedType(tag: DW_TAG_pointer_type, baseType: !43, size: 64)
!53 = !{!54}
!54 = !DILocalVariable(name: "ctx", arg: 1, scope: !49, file: !2, type: !52)

Disassembled bytecode
---------------------------
sh: line 1: llvm-objdump: No such file or directory
```

danobi:
Updated man page

viktormalik (approved):
Oops, I didn't realize `bpf_object__open_mem` doesn't copy the memory (though it makes sense).

Thanks for fixing this! We should certainly backport.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -45,6 +45,10 @@
 #include "types.h"
 #include "usdt.h"
 
+#define _BT_DISASM_TMPFILE ".temp_bt_disassemble"
```
ajor, line 48: Could we put this in TMPDIR instead or CWD? I know we try to clean up, but it still feels wrong to write files to the random directory we happen to be running it.
- Security software might be monitoring for writes to certain directories. We wouldn't want to trigger it by just debugging bpftrace in that directory.
- Maybe the filesystem metadata matters to some people. We're modifying the last updated time on the current directory by doing this.

ajor (changes requested):
I don't think there's actually a UAF bug here. It seems like libbpf/elfutils is already making a copy of the buffer that we pass in.

I can see that the original buffer isn't being used, as I can overwrite it all and still read out valid data from `bpf_object_`: https://pastebin.com/raw/XAsrw6kR

I haven't dug all the way through the elfutils source code, but there's this slightly cryptic comment in libbpf that suggests that it doesn't need the buffer after it's finished loading the ELF file: https://github.com/libbpf/libbpf/blob/caa17bdcbfc58e68eaf4d017c058e6577606bf56/src/libbpf.c#L1485-L1490

src/ast/passes/codegen_llvm.cpp
jordalgo: Eh... that's not such a graceful failure. Looks like a bug in our code TBH.

danobi: Fair enough. But I'd also rather avoid overcomplicating something that'll have like N=3 users

src/ast/passes/codegen_llvm.cpp
danobi: Fair enough. Updated

src/ast/passes/codegen_llvm.cpp
jordalgo: Kk, don't feel strongly enough and you're right, it's probably not a highly used feature.

viktormalik:
> Yeah, you're right. libbpf takes care to not hold onto the elf file past call to bpf_object__open_mem(). So it's not a UAF.
> 
> That being said, the change is still needed for the second commit. Lemme reword the commit msg

Now we don't need to backport the first commit, right?

ajor:
> Now we don't need to backport the first commit, right?

Right.

> That being said, the change is still needed for the second commit. Lemme reword the commit msg

Could we enable this debug output without always making a copy of the ELF file? e.g. do it as part of `emit()` or a new `emit_and_disas()` function.

The debug option will be rarely used so it'd seem nicer to add overhead (CPU + memory) to this new code path rather than to the common code path.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3965,6 +3967,38 @@ BpfBytecode CodegenLLVM::compile()
...
+  sprintf(tmpfile, "%s/.tmp_bt_dis.%d", tmpfs.c_str(), getpid());
+
+  // Write contents out to disk temporarily
+  std::ofstream file(tmpfile, std::ios_base::binary);
+  file.write(reinterpret_cast<const char *>(elf.data()), elf.size());
+  file.close();
+
+  char cmd[120];
+  sprintf(cmd, "llvm-objdump -d %s", tmpfile);
+  std::system(cmd);
```
ajor, line 3996: While this code might be ok, it feels like it's on the edge of being exploitable: buffer overflows with `sprintf` or shell injection with `system`.

How about only making this debug output available in dev builds?

danobi: Ack

danobi: Well, sure. Which is why I agreed to the `NDEBUG` thing. So where are we with this, then? Do we want to harden a debug tool?

danobi: So turns out llvm-objdump supports `-` as input file (read from stdin). Which is really convenient - I recall looking for that earlier but I think I looked at `objdump` instead. I actually did try creating a memfd and sending in /proc/self/fd/...` - didn't work. So updated PR to avoid temp file altogether.

ajor (approved):
Sorry, I misunderstood your ack as a nack! I was only looking at the `disassemble` function and didn't spot the `NDEBUG` in bpftrace.h. The latest update looks really nice though and looks like it'd be fine to have it available in release builds as well if you'd prefer.

danobi:
I think debug builds only is fine. Shelling out is, of course, ugly. Probably good to not get into habit of asking users to run that.

PR #3490, opened 2024-10-07
debug: Support comma separated stages

danobi:
Something like `-d codegen,dis` is more convenient than `-d codegen -d
dis`. It's easy enough to support both, so might as well.

##### Checklist

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -469,6 +469,25 @@ struct Args {
   std::vector<std::string> debug_stages;
 };
 
+static bool parse_debug_stages(const std::string& arg)
+{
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -469,6 +469,25 @@ struct Args {
   std::vector<std::string> debug_stages;
```
jordalgo, line 469: nit: Should this be a set since we're potentially inserting the same stage twice?

jordalgo (approved):
Nice. LGTM

danobi:
Fixed accidental use of `optarg` in the helper

PR #3491, opened 2024-10-07
Warn about potential for data corruption with unaligned probes

ajor:
We should alert users to the dangers of "--unsafe" if we're going to suggest it to them as an option.


jordalgo:
Good clarification.

jordalgo (approved):

PR #3492, opened 2024-10-07
Introduce migration guide for breaking changes

viktormalik:
The idea is that the document will contain one section for each breaking change with:
- link to the PR introducing the change
- brief description of the change
- example of an error message
- simple guide to fix existing scripts

As an example, add entry for the recent change of the type of the `pid` and `tid` builtins.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 #### Breaking Changes
 - Return `uint32` instead of `uint64` for `pid` and `tid` builtins
   - [#3441](https://github.com/bpftrace/bpftrace/pull/3441)
+  - [Migration guide](docs/migration_guide.md#pid-and-tid-builtins-return-uint32)
 #### Added
 - Add `--dry-run` CLI option
   - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ bpftrace is a high-level tracing language for Linux. bpftrace uses LLVM as a bac
 - [Contribute](#contribute)
 - [Development](#development)
 - [Support](#support)
+- [Migration guide](docs/migration_guide.md)
 - [Probe types](#probe-types)
 - [Plugins](#plugins)
 - [License](#license)
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -1,0 +1,31 @@
+# Breaking changes (migration guide)
+
+This document lists changes introduced in bpftrace which break backwards
+compatibility in some way. Each entry should contain:
+- a link to the PR introducing the change
+- a brief description of the change
+- an example of an error message
+- a simple guide to fix existing scripts
+
+## Versions 0.21.x (or earlier) to 0.22.x (or later)
+
+### `pid` and `tid` builtins return `uint32`
+
+https://github.com/bpftrace/bpftrace/pull/3441
+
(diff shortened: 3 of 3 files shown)


docs/breaking.md
```diff
@@ -0,0 +1,29 @@
...
+
+### `pid` and `tid` builtins return `uint32`
+
+https://github.com/bpftrace/bpftrace/pull/3441
+
+Previously, `pid` and `tid` builtins returned `uint64` so it is now possible to
+get an error when storing the builtin in a variable and overriding it with
+`uint64` later:
+```
+# bpftrace -e 'BEGIN { $x = pid; $x = uid; }'   # uid is uint_64
```
ajor, line 18: `uid` might not be a very good example - it really should be a uint32 like pid and we could change that in the future too

ajor (approved):

docs/breaking.md
viktormalik: Good point, changed to `cgroup` which is `uint64` (returned by the BPF helper so very unlikely to change).

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to
 #### Breaking Changes
 - Return `uint32` instead of `uint64` for `pid` and `tid` builtins
   - [#3441](https://github.com/bpftrace/bpftrace/pull/3441)
+  - [Migration guide](docs/breaking.md#pid-and-tid-builtins-return-uint32)
```
jordalgo, line 14: Nit: maybe this should just go as a paragraph under the breaking changes header.

docs/breaking.md
```diff
@@ -0,0 +1,29 @@
+# Breaking changes (migration guide)
+
+This document lists changes introduced in bpftrace which break backwards
+compatibility in some way. Each entry should contain:
+- a link to the PR introducing the change
+- a brief description of the change
+- an example of an error message
+- a simple guide to fix existing scripts
+
+### `pid` and `tid` builtins return `uint32`
```
jordalgo, line 10: We should add the versions as a `##` sub header. Maybe something like
```
## Migrating from 0.21.2 or earlier to 0.22.x or later
```

docs/breaking.md
```diff
@@ -0,0 +1,29 @@
...
+
+https://github.com/bpftrace/bpftrace/pull/3441
```
jordalgo, line 12: Maybe additional subheads for these to make it super clear e.g.
```
#### PR

#### Error

#### Fix
```

jordalgo:
Yay for new migration guide! Can we also add a link from the main readme? I might also change the file name to "migration_guide" instead of "breaking" or "breaking_changes".

docs/breaking.md
viktormalik: Too much visual noise for my liking, I'd prefer not to add it.

viktormalik:
> Yay for new migration guide! Can we also add a link from the main readme? I might also change the file name to "migration_guide" instead of "breaking" or "breaking_changes".

Sure, good idea, added.

CHANGELOG.md
jordalgo: Oh sorry, guess I missed that.

jordalgo (approved):
📝

docs/breaking.md
jordalgo: Yeah, doesn't look like it needs when viewing the rendered markdown.

PR #3494, opened 2024-10-07
Add reference to the, now deprecated, delete API

jordalgo:
This is so users, who are running an older version of bpftrace don't get confused when looking at the updated adoc.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -2484,10 +2484,14 @@ interval:s:10 {
 
 .variants
 * `delete(map m, mapkey k)`
+* deprecated `delete(mapkey k, ...)`
 
 Delete a single key from a map.
 For scalar maps (e.g. no explicit keys), the key is omitted and is equivalent to calling `clear`.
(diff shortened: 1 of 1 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -2499,6 +2503,9 @@ kprobe:dummy {
   delete(@associative, (1,2)); // ok
   delete(@associative); // error
   delete(@associative, 1); // error
+
+  // deprecated but ok
+  delete(@associative[1, 2], @single["hello"]);
```
ajor, line 2508:
```suggestion
  delete(@associative[1, 2]);
  delete(@single["hello"]);
```
Could we show the single map variant instead? It's much older than the multi-one and is what most people will be familiar with.

We'll also be maintaining backwards compatibility with the single variant for a while, but the multi-one I think will have to go soon as it conflicts with the new syntax.

jordalgo: Sure. Will update.

ajor (approved):

PR #3501, opened 2024-10-08
allow custom error code on exit()

### LINKED ISSUE #3497: Add the ability to pass a custom error code to `exit`

Right now bpftrace always exits with a 0 error code. However it would be nice if users could pass a custom error code to exit e.g. `exit(1)` if they want to indicate some sort of error/failure.

Have the `exit` function take an optional int (https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html) and have bpftrace use this on script exit.

amiremohamadi:
Fix #3497

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1257,8 +1257,8 @@ Tracing block I/O sizes > 0 bytes
 | Convert cgroup id to cgroup path
 | Sync
 
-| <<functions-exit, `exit()`>>
-| Quit bpftrace
+| <<functions-exit, `exit([int code])`>>
+| Quit bpftrace with an optional exit code
 | Async
 
 | <<functions-join, `join(char *arr[] [, char *delim])`>>
@@ -1499,12 +1499,27 @@ BEGIN {
 === exit
 
 .variants
-* `void exit()`
+* `void exit([int code])`
 
 *async*
 
 Terminate bpftrace, as if a `SIGTERM` was received.
 The `END` probe will still trigger (if specified) and maps will be printed.
+An optional exit code can be provided.
+
+----
+BEGIN {
+  exit();
+}
+----
+
+Or
+
+----
+BEGIN {
+  exit(1);
+}
+----
 
 [#functions-join]
 === join
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1167,15 +1167,31 @@ void CodegenLLVM::visit(Call &call)
                            "cat",
                            AsyncAction::cat);
   } else if (call.func == "exit") {
-    /*
-     * perf event output has: uint64_t asyncaction_id
-     * The asyncaction_id informs user-space that this is not a printf(), but is
-     * a special asynchronous action. The ID maps to exit().
-     */
-    AllocaInst *perfdata = b_.CreateAllocaBPF(b_.getInt64Ty(), "perfdata");
-    b_.CreateStore(b_.getInt64(asyncactionint(AsyncAction::exit)), perfdata);
-    b_.CreateOutput(ctx_, perfdata, sizeof(uint64_t), &call.loc);
-    b_.CreateLifetimeEnd(perfdata);
+    auto elements = AsyncEvent::Exit().asLLVMType(b_);
+    StructType *exit_struct = b_.GetStructType("exit_t", elements, true);
+    AllocaInst *buf = b_.CreateAllocaBPF(exit_struct, "exit");
+    size_t struct_size = datalayout().getTypeAllocSize(exit_struct);
+
+    // Fill in exit struct.
+    b_.CreateStore(
+        b_.getInt64(asyncactionint(AsyncAction::exit)),
+        b_.CreateGEP(exit_struct, buf, { b_.getInt64(0), b_.getInt32(0) }));
+
+    Value *code = b_.getInt8(0);
+    if (call.vargs.size() == 1) {
+      auto scoped_del = accept(call.vargs.at(0));
+#if LLVM_VERSION_MAJOR <= 14
+      expr_ = b_.CreateIntCast(expr_, b_.getInt8Ty(), false);
+#endif
+      code = expr_;
+    }
+    b_.CreateStore(
+        code,
+        b_.CreateGEP(exit_struct, buf, { b_.getInt64(0), b_.getInt32(1) }));
+
+    b_.CreateOutput(ctx_, buf, struct_size, &call.loc);
+    b_.CreateLifetimeEnd(buf);
+
     expr_ = nullptr;
     createRet();
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -1139,7 +1139,12 @@ void SemanticAnalyser::visit(Call &call)
     call.type = CreateNone();
   } else if (call.func == "exit") {
     check_assignment(call, false, false, false);
-    (void)check_nargs(call, 0);
+
+    if (!check_varargs(call, 0, 1))
+      return;
+
+    if (call.vargs.size() == 1)
+      check_arg(call, Type::integer, 0);
   } else if (call.func == "print") {
     check_assignment(call, false, false, false);
     if (in_loop() && is_final_pass() && call.vargs.at(0)->is_map) {
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -51,6 +51,7 @@ std::set<DebugStage> bt_debug = {};
 bool bt_quiet = false;
 bool bt_verbose = false;
 bool dry_run = false;
+int BPFtrace::exit_code = 0;
 volatile sig_atomic_t BPFtrace::exitsig_recv = false;
 volatile sig_atomic_t BPFtrace::sigusr1_recv = false;
 
@@ -261,6 +262,8 @@ void perf_event_printer(void *cb_cookie, void *data, int size)
 
   // async actions
   if (printf_id == asyncactionint(AsyncAction::exit)) {
+    auto exit = static_cast<AsyncEvent::Exit *>(data);
+    BPFtrace::exit_code = exit->exit_code;
     bpftrace->request_finalize();
     return;
   } else if (printf_id == asyncactionint(AsyncAction::print)) {
--- a/tests/codegen/call_exit.cpp
+++ b/tests/codegen/call_exit.cpp
@@ -9,6 +9,11 @@ TEST(codegen, call_exit)
   test("kprobe:f { exit(); @=10 }", NAME);
 }
 
+TEST(codegen, call_exit_with_error_code)
+{
+  test("kprobe:f { exit(1); }", NAME);
+}
+
 } // namespace codegen
 } // namespace test
 } // namespace bpftrace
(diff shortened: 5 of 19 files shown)


amiremohamadi:
trying to add `runtime` test later. apparently `tests/runtime/engine/runner.py` needs to be changed in order to be able to check the exit code.

man/adoc/bpftrace.adoc
```diff
@@ -1257,7 +1257,7 @@ Tracing block I/O sizes > 0 bytes
 | Convert cgroup id to cgroup path
 | Sync
 
-| <<functions-exit, `exit()`>>
+| <<functions-exit, `exit([int code])`>>
```
jordalgo, line 1260: Let's add a note about the code being optional

tests/semantic_analyser.cpp
```diff
@@ -948,7 +948,8 @@ kprobe:f { @x = 1; $y = 2; $c = 3; delete(@x, $y, $c); }
 TEST(semantic_analyser, call_exit)
 {
   test("kprobe:f { exit(); }");
-  test("kprobe:f { exit(1); }", 1);
+  test("kprobe:f { exit(1); }");
+  test("kprobe:f { exit(1, 2); }", 1);
```
jordalgo, line 952: let's use `test_error` to test the error message

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1104,7 +1104,12 @@ void SemanticAnalyser::visit(Call &call)
     call.type = CreateNone();
   } else if (call.func == "exit") {
     check_assignment(call, false, false, false);
-    (void)check_nargs(call, 0);
+
+    if (!check_varargs(call, 0, 1))
+      return;
+
+    if (call.vargs.size() == 1)
+      check_arg(call, Type::integer, 0);
```
jordalgo, line 1112: I need to read through this a bit more but maybe we need to limit the size of the exit code to an int8
https://pubs.opengroup.org/onlinepubs/9699919799/functions/_exit.html

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1154,15 +1154,29 @@ void CodegenLLVM::visit(Call &call)
...
+    AllocaInst *buf = b_.CreateAllocaBPF(exit_struct, "exit");
+    size_t struct_size = datalayout().getTypeAllocSize(exit_struct);
+
+    // Fill in perf event struct.
+    b_.CreateStore(
+        b_.getInt64(asyncactionint(AsyncAction::exit)),
+        b_.CreateGEP(exit_struct, buf, { b_.getInt64(0), b_.getInt32(0) }));
+    auto code = 0;
+    if (call.vargs.size() == 1) {
+      auto code_opt = bpftrace_.get_int_literal(call.vargs.at(0));
```
jordalgo, line 1168: so can this not be a variable?

jordalgo (commented):
Looking good so far. Thanks for working on this.

amiremohamadi:
> Oh also, can we add a codegen test and some runtime tests (when ready)

@jordalgo 
Thanks for the review. I think some changes need be applied on `tests/runtime/engine/` in order to be able to check the exit code in runtime tests. right? any other ideas?

src/ast/passes/semantic_analyser.cpp
amiremohamadi: good catch! actually we're using int8 in other parts of the [code](https://github.com/bpftrace/bpftrace/blob/c67683752655258a0b730840b95dcdf84ef2f8ec/src/ast/passes/codegen_llvm.cpp#L1174). will change it.

src/ast/passes/codegen_llvm.cpp
amiremohamadi: this will return an `std::optional` which should be checked in the latter line. we can't directly assign it to the `code` variable.

ajor: I think what @jordalgo meant was a script using a variable like this:
```
$exit_code = 1;
exit($exit_code);
```

We should be able to make this work. `get_int_literal` only lets us get the value of a literal integer, but if you call `accept(call.vargs.at(0))` instead, that that'll recurse the AST and populate `expr_` with an expression representing either a literal or a non-literal value. Take a look at the implementation of the `signal` function for a real example.

jordalgo:
>  I think some changes need be applied on tests/runtime/engine/ in order to be able to check the exit code in runtime tests. right? 

Yeah that sounds correct. Let me know if you want some help in adding this.

amiremohamadi:
rebased my changes. would u please do another review?
@jordalgo

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1167,15 +1167,28 @@ void CodegenLLVM::visit(Call &call)
...
-    AllocaInst *perfdata = b_.CreateAllocaBPF(b_.getInt64Ty(), "perfdata");
-    b_.CreateStore(b_.getInt64(asyncactionint(AsyncAction::exit)), perfdata);
-    b_.CreateOutput(ctx_, perfdata, sizeof(uint64_t), &call.loc);
-    b_.CreateLifetimeEnd(perfdata);
+    auto elements = AsyncEvent::Exit().asLLVMType(b_);
+    StructType *exit_struct = b_.GetStructType("exit_t", elements, true);
+    AllocaInst *buf = b_.CreateAllocaBPF(exit_struct, "exit");
+    size_t struct_size = datalayout().getTypeAllocSize(exit_struct);
+
+    // Fill in perf event struct.
```
jordalgo, line 1175: Nit:
```suggestion
    // Fill in exit struct.
```
Or just remove this comment - it's not super useful

src/ast/passes/semantic_analyser.cpp
jordalgo: Did this get missed in an update?

I think for a int literal we can probably check the range and for a variable we can insert a cast to int8 perhaps.

tests/codegen/call_exit.cpp
```diff
@@ -9,6 +9,11 @@ TEST(codegen, call_exit)
   test("kprobe:f { exit(); @=10 }", NAME);
 }
 
+TEST(codegen, call_exit_with_error_code)
+{
+  test("kprobe:f { exit(1); @=10 }", NAME);
```
jordalgo, line 14: why the added map assignment at the end?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1167,15 +1167,28 @@ void CodegenLLVM::visit(Call &call)
...
+    b_.CreateStore(
+        b_.getInt64(asyncactionint(AsyncAction::exit)),
+        b_.CreateGEP(exit_struct, buf, { b_.getInt64(0), b_.getInt32(0) }));
+
+    Value *code = b_.getInt8(0);
+    if (call.vargs.size() == 1) {
+      auto scoped_del = accept(call.vargs.at(0));
+      code = b_.CreateIntCast(expr_, b_.getInt8Ty(), false /* unsigned */);
```
jordalgo, line 1183: If we know the expression is going to be a int8 do we need this IntCast?

tests/runtime/basic
```diff
@@ -324,3 +324,9 @@ EXPECT @b[1]: 0
 NAME has_key as a variable
 PROG BEGIN { @a[1] = 1; $b = has_key(@a, 1); $c = has_key(@a, 2); print(($b, $c)); exit(); }
 EXPECT (1, 0)
+
+NAME exit code
+PROG BEGIN { exit(69); }
+EXPECT_NONE 69
```
jordalgo, line 330: What's this EXPECT_NONE for?

jordalgo (changes requested):

src/ast/passes/semantic_analyser.cpp
amiremohamadi: I guess we don't need to check the range as it says
> only the least significant 8 bits (that is, status & 0377) shall be available from wait() and waitpid()

tests/runtime/basic
amiremohamadi: at least one variation of EXPECT was required in runtime tests. modified it in a way to only require expect when no RETURN_CODE is provided.

amiremohamadi:
one more thing. currently if we exit with return codes other than 0, `print_maps` won't be called ([L52](https://github.com/bpftrace/bpftrace/blob/2edd37e6aabc627726da9e7960110a9c9a3ce213/src/run_bpftrace.cpp#L52)). what's the correct behavior? because we can use the exit code at [this line](https://github.com/bpftrace/bpftrace/blob/2edd37e6aabc627726da9e7960110a9c9a3ce213/src/run_bpftrace.cpp#L74) instead.

tests/runtime/engine/parser.py
```diff
@@ -124,6 +126,8 @@ def __read_test_struct(test, test_suite):
         will_fail = False
         new_pidns = False
         skip_if_env_has = None
+        return_code = 0
+        has_return_code = False
```
ajor, line 130:
```suggestion
        return_code = None
```
My Python's rusty, but I think we can do something like this to avoid using two variables when one will do. Then check `return_code is not None` instead of `has_return_code`

tests/runtime/engine/runner.py
```diff
@@ -108,7 +108,7 @@ def prepare_bpf_call(test, nsenter=[]):
 
             return nsenter_prefix + ret
         else:  # PROG
-            use_json = "-q -f json" if test.expects[0].mode == "json" else ""
+            use_json = "-q -f json" if (not test.has_return_code and test.expects[0].mode == "json") else ""
```
ajor, line 111:
```suggestion
            use_json = "-q -f json" if (len(test.expects) > 0 and test.expects[0].mode == "json") else ""
```
To allow using both `RETURN_CODE` and `EXPECT_JSON` at the same time

src/bpftrace.cpp
```diff
@@ -1016,7 +1019,7 @@ int BPFtrace::run(BpfBytecode bytecode)
 
   if (dry_run) {
     request_finalize();
-    return 0;
+    return exit_code;
```
ajor, line 1022: I'd say this should remain hardcoded to `0` for clarity. A dry run can never fail.

ajor (commented):
> one more thing. currently if we exit with return codes other than 0, `print_maps` won't be called ([L52](https://github.com/bpftrace/bpftrace/blob/2edd37e6aabc627726da9e7960110a9c9a3ce213/src/run_bpftrace.cpp#L52)). what's the correct behavior?

Good point! I'd say that the default printing of maps should always occur regardless of the exit code.

If people want to suppress the map printing, they can use an `END` probe to clear the maps as normal.

src/ast/passes/semantic_analyser.cpp
jordalgo: Ok, but you'll still need a cast somewhere (this is why you're getting errors on LLVM 13/14. You can add the conditional codegen cast like I did here:
https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/codegen_llvm.cpp#L782

ajor (approved):
Nice, thank you!

jordalgo (approved):
Thanks for the contribution!

amiremohamadi:
@ajor @jordalgo thanks for the reviews. is there any chance you can add a `hacktoberfest-accepted` label to this PR?
https://hacktoberfest.com/participation/#pr-mr-details

Issue #646, opened 2019-05-16
ustack missing stack frame second from top


danobi:
```
$ sudo ./build/src/bpftrace -e 'uprobe:/lib64/libc.so.6:usleep { printf("%s\n", ustack); @[ustack(1)] = count();
 }'                                                                                                                                          
Attaching 1 probe...                                                                                                                         
                                                                                                                                             
        usleep+0                                                                                                                             
        pathBBBB()+9                                                                                                                         
        pathBBB()+9                                                                                                                          
        pathBB()+9                                                                                                                           
        pathB()+9                                                                                                                            
        main+48                                                                                                                              
        __libc_start_main+235                                                                                                                
        0x41fd89415541f689                                                                                                                   
                                                                                                                                             
                                                                                                                                             
        usleep+0                                                                                                                             
        pathAAAA()+9                                                                                                                         
        pathAAA()+9                                                                                                                          
        pathAA()+9                                                                                                                           
        pathA()+9                                                                                                                            
        main+41                                                                                                                              
        __libc_start_main+235                                                                                                                
        0x41fd89415541f689                                                                                                                   
                                                                                                                                             
^C                                                                                                                                           
                                                                                                                                             
@[                                                                                                                                           
    usleep+0
]: 2

$ cat ~/scratch/memcpy/main.cpp
#include <iostream>
#include <unistd.h>
#include <stdio.h>

__attribute__((noinline)) void do_sleep() {
  ::usleep(100000);
}

__attribute__((noinline)) void pathAAAA() {
  do_sleep();
}

__attribute__((noinline)) void pathAAA() {
  pathAAAA();
}

__attribute__((noinline)) void pathAA() {
  pathAAA();
}

__attribute__((noinline)) void pathA() {
  pathAA();
}

__attribute__((noinline)) void pathBBBB() {
  do_sleep();
}

__attribute__((noinline)) void pathBBB() {
  pathBBBB();
}

__attribute__((noinline)) void pathBB() {
  pathBBB();
}

__attribute__((noinline)) void pathB() {
  pathBB();
}

int main() {
  int i = 0;
  while (1) {
    if (i++ % 2) {
      pathA();
    } else {
      pathB();
    }
  }
}
```

Notice how `do_sleep` is missing in both stack traces.

brendangregg:
Interesting. What does perf record show?

```
perf probe -x /lib64/libc.so.6 usleep
perf record -e probe_libc:usleep -a -g
perf script
```

[...]

PR #3504, opened 2024-10-09
Add 'symbol_source' config

### LINKED ISSUE #3493: Add a config variable to use symbol addresses instead of LLDB location resolution

We saw an issue where LLDB was generating a uprobe location that was falling in between two instructions. This is bad and can cause memory corruption in the targeted process if the uprobe is attached (via use of the `--unsafe` flag).

Create a [config variable](https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#config-variables) that will default to using LLDB location resolution but provides the option of disabling this and using symbol addresses for all probes.

Deciding not to go with this option: https://github.com/bpftrace/bpftrace/pull/3481

ttreyer:
Let the user decide whether to use DebugInfo to determine `uprobe` location, or use the Symbol Table.

The new `symbol_source` config takes the following two values:
- `dwarf`: uses the DebugInfo, which yield more accurate stack traces. However, if the DebugInfo was rewritten by a post-linkage optimisation tool (like BOLT or AutoFDO), we might get wrong addresses for the probe location.
- `symbol_table`: uses the target's Symbol Table, which won't suffer from improperly rewritten DebugInfo.

This will close #3493.

##### Checklist

  - [x] Test setting the config
  - [x] Test the config impact

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ and this project adheres to
   - [#3515](https://github.com/bpftrace/bpftrace/pull/3515)
 - Configuration option to suppress printing maps by default at program exit
   - [#3547](https://github.com/bpftrace/bpftrace/pull/3547)
+- Add `symbol_source` config to source uprobe locations from either DWARF or the Symbol Table
+  - [#3504](https://github.com/bpftrace/bpftrace/pull/3504/)
 #### Changed
 - Merge output into `stdout` when `-lv`
   - [#3383](https://github.com/bpftrace/bpftrace/pull/3383)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -3703,6 +3703,20 @@ Default: 1
 
 Controls whether maps are printed on exit. Set to `0` in order to change the default behavior and not automatically print maps at program exit.
 
+=== symbol_source
+
+Default: `dwarf` if `bpftrace` is compiled with LLDB, `symbol_table` otherwise
+
+Choose how bpftrace will resolve all `uprobe` symbol locations.
+
+Available options:
+
+- `dwarf` - locate uprobes using DebugInfo, which yields more accurate stack traces (`ustack`). Fall back to the Symbol Table if it can't locate the probe using DebugInfo.
+- `symbol_table` - don't use DebugInfo and rely on the ELF Symbol Table instead.
+
+If the DebugInfo was rewritten by a post-linkage optimisation tool (like BOLT or AutoFDO), it might yield an incorrect address for a probe location.
+This config can force using the Symbol Table, for when the DebugInfo returns invalid addresses.
+
 == Environment Variables
 
 These are not available as part of the standard set of <<Config Variables>> and can only be set as environment variables.
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -349,7 +349,11 @@ static constexpr std::string_view hint_unsafe =
     "\nUse --unsafe to force attachment. WARNING: This option could lead to "
     "data corruption in the target process.";
 
-static void check_alignment(std::string &path,
+static constexpr std::string_view hint_symbol_source =
+    "\nUse config 'symbol_source = \"symbol_table\"' in case of bad DebugInfo.";
+
+static void check_alignment(std::string &orig_name,
+                            std::string &path,
                             std::string &symbol,
                             uint64_t sym_offset,
                             uint64_t func_offset,
@@ -365,13 +369,20 @@ static void check_alignment(std::string &path,
     case AlignState::Ok:
       return;
     case AlignState::NotAlign:
-      if (safe_mode)
-        throw FatalUserException("Could not add " + probetypeName(type) +
-                                 " into middle of instruction: " + tmp +
-                                 std::string{ hint_unsafe });
-      else
+      if (safe_mode) {
+        auto msg = "Could not add " + probetypeName(type) +
+                   " into middle of instruction: " + tmp +
+                   std::string{ hint_unsafe };
+        if (orig_name.find('*') != std::string::npos)
+          msg += hint_symbol_source;
+        throw FatalUserException(std::move(msg));
+      } else {
+        std::string_view hint;
+        if (orig_name.find('*') != std::string::npos)
+          hint = hint_symbol_source;
         LOG(WARNING) << "Unsafe " << type
-                     << " in the middle of the instruction: " << tmp;
+                     << " in the middle of the instruction: " << tmp << hint;
+      }
       break;
 
     case AlignState::Fail:
@@ -481,8 +492,13 @@ bool AttachedProbe::resolve_offset_uprobe(bool safe_mode, bool has_multiple_aps)
   if (func_offset == 0)
     return true;
 
-  check_alignment(
-      probe_.path, symbol, sym_offset, func_offset, safe_mode, probe_.type);
+  check_alignment(probe_.orig_name,
+                  probe_.path,
+                  symbol,
+                  sym_offset,
+                  func_offset,
+                  safe_mode,
+                  probe_.type);
   return true;
 }
 
--- a/tests/runtime/dwarf
+++ b/tests/runtime/dwarf
@@ -74,6 +74,22 @@ EXPECT foo1->d = 9abcdef0
 REQUIRES_FEATURE dwarf
 BEFORE ./testprogs/uprobe_test
 
+NAME uprobe without dwarf
+PROG config = { symbol_source = "symbol_table"; cache_user_symbols = "PER_PROGRAM"; }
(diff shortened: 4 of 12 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ and this project adheres to
   - [#3390](https://github.com/bpftrace/bpftrace/pull/3390/)
 - Add `has_key` function for maps
   - [#3358](https://github.com/bpftrace/bpftrace/pull/3358)
+- Add `symbol_source` config to source uprobe locations from either DWARF or the Symbol Table
```
jordalgo, line 32: Naming question: is `uprobe_symbol_source` to long?

man/adoc/bpftrace.adoc
```diff
@@ -3563,6 +3563,16 @@ Default: `..`
 Trailer to add to strings that were truncated.
 Set to empty string to disable truncation trailers.
 
+=== symbol_source
+
+Default: `dwarf`
+
+Choose how bpftrace will resolve `uprobe`s' location.
```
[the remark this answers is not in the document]
jordalgo: I would also call out that we default to `symbol_table` if LLDB is not available.

man/adoc/bpftrace.adoc
```diff
@@ -3563,6 +3563,16 @@ Default: `..`
...
+
+Available options:
+- `dwarf` - locate probe using DebugInfo, it will give better stack trace output. bpftrace will fall back to the Symbol Table if it can't locate the probe using DebugInfo.
```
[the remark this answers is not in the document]
jordalgo: > fall back to the Symbol Table if it can't locate the probe using DebugInfo.

Where does this happen?

jordalgo (changes requested):
Code looks pretty good. Maybe good to also add a runtime test (if that's possible).

CHANGELOG.md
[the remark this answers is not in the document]
ttreyer: It would be fine, but this settings is also going to apply to kprobes in #3301. Do you have a suggestion fitting for both?

man/adoc/bpftrace.adoc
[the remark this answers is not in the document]
ttreyer: Right [here](https://github.com/bpftrace/bpftrace/pull/3504/files#diff-f6bea5bb780ac5ca59678079c8010b81f5b8eb83cbe90f77e134fd6f2a742a32R168-R170).

CHANGELOG.md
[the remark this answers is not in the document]
jordalgo: Oh, ok. Didn't know kprobes were going to be using it too. This is fine then.

jordalgo:
Looks good! One last thing (and maybe this needs to be a separate PR) but can we detect if users are setting an offset themselves or using a glob for their uprobe and, in the latter case, suggest using this new config if `check_alignment` fails (https://github.com/bpftrace/bpftrace/blob/2fa7894baf76934b761772962a8154485c6ba44c/src/attached_probe.cpp#L405).

Also was a runtime test too much of a PITA?

ttreyer:
> Looks good! One last thing (and maybe this needs to be a separate PR) but can we detect if users are setting an offset themselves or using a glob for their uprobe and, in the latter case, suggest using this new config if `check_alignment` fails (
> 
> https://github.com/bpftrace/bpftrace/blob/2fa7894baf76934b761772962a8154485c6ba44c/src/attached_probe.cpp#L405
> 
> ).

Good suggestion! I've added it as a hint.

> Also was a runtime test too much of a PITA?

My bad, I've added two to test both with and without `probe_inline` enabled.

man/adoc/bpftrace.adoc
```diff
@@ -3563,6 +3563,19 @@ Default: `..`
...
 Set to empty string to disable truncation trailers.
 
+=== symbol_source
+
+Default: `dwarf` if `bpftrace` is compiled with LLDB, `symbol_table` otherwise
+
+Choose how bpftrace will resolve all `uprobe` symbol locations.
+
+Available options:
+- `dwarf` - locate uprobes using DebugInfo, which yields more accurate stack traces (`ustack`). Fall back to the Symbol Table if it can't locate the probe using DebugInfo.
```
jordalgo, line 3573: nit: Insert a line break above this list, it's not getting formatted correctly in the markdown.

src/attached_probe.cpp
```diff
@@ -401,13 +405,17 @@ static void check_alignment(std::string &path,
...
-      if (safe_mode)
-        throw FatalUserException("Could not add " + probetypeName(type) +
-                                 " into middle of instruction: " + tmp +
-                                 std::string{ hint_unsafe });
-      else
+      if (safe_mode) {
+        auto msg = "Could not add " + probetypeName(type) +
+                   " into middle of instruction: " + tmp +
+                   std::string{ hint_unsafe };
+        if (orig_name.find('*') != std::string::npos)
```
jordalgo, line 412: Should we also add this hint to the other failing case statements below?

tests/runtime/dwarf
```diff
@@ -74,6 +74,22 @@ EXPECT foo1->d = 9abcdef0
...
+EXPECT_REGEX_NONE ^\s*main\+\d+$
+REQUIRES_FEATURE dwarf
+AFTER ./testprogs/uprobe_test
+
+NAME uprobe inline without dwarf
+PROG config = { symbol_source = "symbol_table"; probe_inline = 1; }
+     uprobe:./testprogs/inline_function:square { @count++ }
+     uretprobe:./testprogs/inline_function:main { exit() }
+EXPECT @count: 1
+REQUIRES_FEATURE dwarf
```
jordalgo, line 90: why does this one require dwarf?

ttreyer: It requires `liblldb` to test that the `symbol_source` config takes precedence over `probe_inline`.

src/attached_probe.cpp
jordalgo: @ttreyer I meant using `hint_symbol_source` for the AlignState::Fail and AlignState::NotSupp - or do those not apply?

jordalgo:
Looks like maybe one of the new added tests needs to fixed for AOT or added to the skip list.

CHANGELOG.md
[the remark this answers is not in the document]
ajor: I'm not so bothered about the name but just want to clarify what the plans are for these config options.

Inlined kprobes are already going to be controlled by the probe_inline config option, right? For non-inlined kprobes, I'm not sure there's any benefit to using LLDB's locations. The reason for uprobes was that the kernel doesn't provide correct stack traces otherwise, but that's not the case for kprobes.

ttreyer:
I've added it to the AOT skip list, as Dwarf is currently not compatible with AOT.

CHANGELOG.md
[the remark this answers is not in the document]
viktormalik: To second what @ajor said, would it make sense to extend the `probe_inline` option to choose whether to use DWARF or symbol table for placing the probes? And use symbol table for non-inlined uprobes?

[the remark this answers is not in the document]
ajor: There is still a reason to have two separate config options - we want to control these two things independently:
- Uprobes/kprobes tracing inlined functions
- Uprobes using debug info to pick attach points for non-inlined functions (to fix ustacks: #646)

The second one is needed because we've come across a case where BOLT (or similar) messes up the DWARF info for non-inlined functions, so we need a way to fall back to the symbol table in that case.

We shouldn't be using DWARF info for non-inlined kprobes anyway, so this config option only needs to apply to uprobes.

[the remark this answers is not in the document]
viktormalik: Ah, ok, thanks for the clarification.

src/attached_probe.cpp
ttreyer: I understand `AlignState::Fail` and `AlignState::NotSupp` to be errors returned by the disassembler, not necessary an issue with the address itself. That's why I'm not convinced that the hint applies to both of them.

jordalgo: Would these failure occur if the user switched to `symbol_table` mode?

fbs:
Somewhat uninformed question:

Does it make sense for this to be a 'global' toggle, or would you want to be able to do this per probe? Or maybe a global and per probe?

src/attached_probe.cpp
ttreyer: Yes, the failure would still occur.

ttreyer:
> Does it make sense for this to be a 'global' toggle, or would you want to be able to do this per probe? Or maybe a global and per probe?

It would make sense to have config per-probe instead of always being global. But `bpftrace` doesn't support per-probe config currently.

src/attached_probe.cpp
jordalgo: Ok cool. I'm satisfied then 😄

jordalgo (approved):

PR #3506, opened 2024-10-09
Remove multi-map delete API

jordalgo:
This is no longer valid:
```
@a = 1;
@b[1] = 2;

delete(@a, @b[1]);
```

This also fixes map key resizing for the `delete` function: this is no longer an error: `@y["hi"] = 5; delete(@y, "longerstr");`

https://github.com/bpftrace/bpftrace/issues/3495
https://github.com/bpftrace/bpftrace/issues/3496

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -756,18 +756,25 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateLifetimeEnd(key);
     expr_ = nullptr;
   } else if (call.func == "delete") {
-    for (const auto &arg : call.vargs) {
-      auto &map = static_cast<Map &>(*arg);
-      auto [key, scoped_key_deleter] = getMapKey(map);
-      if (!is_bpf_map_clearable(map_types_[map.ident])) {
-        // store zero instead of calling bpf_map_delete_elem()
-        AllocaInst *val = b_.CreateAllocaBPF(map.type, map.ident + "_zero");
-        b_.CreateStore(Constant::getNullValue(b_.GetType(map.type)), val);
-        b_.CreateMapUpdateElem(ctx_, map.ident, key, val, call.loc);
-        b_.CreateLifetimeEnd(val);
-      } else {
-        b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
-      }
+    auto &arg0 = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg0);
+    // Current API: delete accepts two arguments except in the case of scalar
+    // maps (maps with no keys) in which case it you can just pass it the map
+    // and it will act similar to `clear` e.g. `delete(@scalar);`
+    // Legacy API: delete accepts a single argument that is the map with a
+    // key expression e.g. `delete(@mymap[1, 2]);` or no key if the map
+    // is a scalar
+    auto [key, scoped_key_deleter] = call.vargs.size() > 1
+                                         ? getMapKey(map, call.vargs.at(1))
+                                         : getMapKey(map);
+    if (!is_bpf_map_clearable(map_types_[map.ident])) {
+      // store zero instead of calling bpf_map_delete_elem()
+      AllocaInst *val = b_.CreateAllocaBPF(map.type, map.ident + "_zero");
+      b_.CreateStore(Constant::getNullValue(b_.GetType(map.type)), val);
+      b_.CreateMapUpdateElem(ctx_, map.ident, key, val, call.loc);
+      b_.CreateLifetimeEnd(val);
+    } else {
+      b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
     }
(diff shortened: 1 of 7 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -754,18 +754,19 @@ void CodegenLLVM::visit(Call &call)
...
-        b_.CreateMapUpdateElem(ctx_, map.ident, key, val, call.loc);
-        b_.CreateLifetimeEnd(val);
-      } else {
-        b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
-      }
+    auto &arg0 = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg0);
+    auto [key, scoped_key_deleter] = call.vargs.size() > 1
+                                         ? getMapKey(map, call.vargs.at(1))
+                                         : getMapKey(map);
```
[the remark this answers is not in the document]
jordalgo: Cool. Will add.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -549,7 +549,7 @@ namespace {
 bool skip_key_validation(const Call &call)
 {
   return call.func == "print" || call.func == "clear" || call.func == "zero" ||
-         call.func == "len" || call.func == "has_key";
+         call.func == "len" || call.func == "has_key" || call.func == "delete";
```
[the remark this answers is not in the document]
jordalgo: Good catch. Fixed in the latest.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -756,18 +756,22 @@ void CodegenLLVM::visit(Call &call)
...
-        b_.CreateMapUpdateElem(ctx_, map.ident, key, val, call.loc);
-        b_.CreateLifetimeEnd(val);
-      } else {
-        b_.CreateMapDeleteElem(ctx_, map, key, call.loc);
-      }
+    auto &arg0 = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg0);
+    // Note: delete accepts two arguments except in the case of scalar maps
+    // (maps with no keys) in which case it you can just pass it the map
+    // and it will act similar to `clear` e.g. `delete(@scalar);`
```
ajor, line 763: Also supports one argument for the legacy delete syntax!
`delete(@map[123])`

jordalgo: Haha fine. But that's going away.

ajor: I'm not sure it should go away - it's been part of the language for so long that all existing scripts are using it. Should we really break them?

jordalgo: We can create a separate issue to discuss but my 2cents - yes we should deprecate and eventually remove. From someone else reading the script the syntax makes no sense. `delete(@map[123])` looks like we're deleting the value and deleting it from what? Also do we want to set a precedence that we can't delete anything from the bpftrace API? We could do this in phases e.g. begin by one release issuing a warning then the next release it gets deleted.

jordalgo:
> Sorry, found another case that shouldn't be accepted:

Don't be sorry! I wish I had thought of these cases. Anyway, fixed the latest one and added a test case.

src/ast/passes/codegen_llvm.cpp
jordalgo: Either way, I updated the comment language in the codegen file.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -724,46 +724,48 @@ void SemanticAnalyser::visit(Call &call)
...
+              Map &map_as_key = static_cast<Map &>(arg1);
+              auto *map_as_key_key_type = get_map_key_type(map_as_key);
+              if (!map_as_key.key_expr && map_as_key_key_type &&
+                  !map_as_key_key_type->IsNoneTy()) {
+                LOG(ERROR, map_as_key.loc, err_)
+                    << "Argument mismatch for " << map_as_key.ident << ": "
+                    << "trying to access with no arguments when map expects "
+                       "arguments: "
+                       "'"
+                    << *map_as_key_key_type << "'";
```
ajor, line 762: It looks like we're duplicating some code from `visit(Map &map)` here. It might be cleaner to keep all key validation logic in that function.

The issue seems to be that we're setting `map.skip_key_validation` for delete when we don't actually want to skip validation for the second argument. What about changing the call to `skip_key_validation(const Call &call)` to only run on the first argument? https://github.com/bpftrace/bpftrace/blob/84cfa952f5c4d34e6b7771fdf42e1b7dd422b011/src/ast/passes/semantic_analyser.cpp#L592

jordalgo: Good call. This made me realize I also need to apply this same logic for `has_key`.

ajor (approved):
Nice!

PR #3510, opened 2024-10-10
Reduce jumps needed from scratch buffer lookup

ryantimwilson:
We currently use global scratch variables to allow for storing big strings. It looks like this in BPF bytecode:
```
ValueType global_var [NUM_CPUS][NUM_SLOTS]

buf = global_var[bpf_get_smp_processor_id()][slot_id]
```

On newer kernels, BPF verifier knows how to bound bpf_get_smp_processor_id. However, on older kernels, the BPF verifier this value is unbounded. Currently, to bound the value of bpf_get_smp_processor_id, we do:
```
bounded_cpu_id = min(bpf_get_smp_processor_id(), NUM_CPUS)
```

However, this min call causes the BPF verifier to do additional jumps and in some complex programs, we hit the kernel complexity limit of 8192 jumps:
```
The sequence of 8193 jumps is too complex.
```

To fix this, we use an AND instruction instead:
```
bounded_cpu_id = bpf_get_smp_processor_id() & (NUM_CPUS - 1)
```

Thus, we need to round up NUM_CPUS to the next largest power of 2 e.g. 55 CPUs will be rounded up to 64 CPUs.

The con is that this allocates more memory than necessary (in the worst case, 2x scratch variable space). But the whole big strings project is about trading off memory increase with better usability.

We previously tried to cache the bounded CPU ID value in a LLVM register. However, LLVM15 started aggressively dumping registers to the stack and the BPF verifier in older kernels (6.5) do not know how to bound values read from the stack.

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -514,10 +514,10 @@ Value *IRBuilderBPF::createScratchBuffer(
   auto max = CreateLoad(getInt64Ty(),
                         module_.getGlobalVariable(to_string(
                             bpftrace::globalvars::GlobalVar::MAX_CPU_ID)));
-  // Verify CPU ID is between 0 and MAX_CPU_ID to ensure we pass BPF
-  // verifer on older 5.2 kernels
-  auto cmp = CreateICmp(CmpInst::ICMP_ULE, cpu_id, max, "cpuid.min.cmp");
-  auto select = CreateSelect(cmp, cpu_id, max, "cpuid.min.select");
+  // Mask CPU ID by MAX_CPU_ID to ensure BPF verifier knows CPU ID is bounded
+  // on older kernels. Note this means MAX_CPU_ID must be 2^N - 1 for some N.
+  // See get_max_cpu_id() for more details.
+  auto bounded_cpu_id = CreateAnd(cpu_id, max, "cpu.id.bounded");
 
   // Note the 1st index is 0 because we're pointing to
   // ValueType var[MAX_CPU_ID + 1][num_elements]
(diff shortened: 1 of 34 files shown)


ajor (approved):
Nice!

Also got a new top score during my celebration game of snake:
```
##################################################
# $@@                                            #
#  @                                             #
#  @                                             #
#  @                                             #
#  @                                             #
#  @                                             #
#  @                                             #
#  @                                             #
#                                                #
#                                                #
##################################################
```

src/ast/irbuilderbpf.cpp
```diff
@@ -514,10 +514,10 @@ Value *IRBuilderBPF::createScratchBuffer(
   auto max = CreateLoad(getInt64Ty(),
                         module_.getGlobalVariable(to_string(
                             bpftrace::globalvars::GlobalVar::MAX_CPU_ID)));
-  // Verify CPU ID is between 0 and MAX_CPU_ID to ensure we pass BPF
-  // verifer on older 5.2 kernels
-  auto cmp = CreateICmp(CmpInst::ICMP_ULE, cpu_id, max, "cpuid.min.cmp");
-  auto select = CreateSelect(cmp, cpu_id, max, "cpuid.min.select");
+  // Mask CPU ID by MAX_CPU_ID to ensure BPF verifier knows CPU ID is bounded
+  // on older kernels. Note this means MAX_CPU_ID must be N^2 - 1 for some N.
```
danobi, line 518: You mean this right?
```suggestion
  // on older kernels. Note this means MAX_CPU_ID must be 2^N - 1 for some N.
```

src/utils.cpp
```diff
@@ -91,6 +91,21 @@ std::vector<std::string> expand_wildcard_paths(
...
+  }
+  n--;
+  n |= n >> 1;
+  n |= n >> 2;
+  n |= n >> 4;
+  n |= n >> 8;
+  n |= n >> 16;
+  return n + 1;
+}
+
```
danobi, line 108: Might be good to add some unit tests for posterity. At least it'll be documentation

src/ast/irbuilderbpf.cpp
ryantimwilson: Yes, math is hardz

src/utils.cpp
ryantimwilson: Done, even found I had to deal with edge cases where n > 2^31 :)

danobi (approved):
a robust copy pasta!

edit: oh, I see the assert was added

PR #3512, opened 2024-10-11
Migrate strings from scratch maps to global scratch variables

ryantimwilson:
Continuing to iterate on https://github.com/bpftrace/bpftrace/issues/3431

Global scratch maps allows users to use large strings in bpftrace. However, we've seen they can cause many jumps/branches in large programs. Global scratch variables allow both large strings and support complex programs via read-write global variables.

This PR is in preparation for adding a scratch buffer vs. stack threshold as described in https://github.com/bpftrace/bpftrace/pull/3483#issuecomment-2407858401


danobi (approved):
lgtm!

jordalgo (approved):

PR #3513, opened 2024-10-11
Add config to store objects on stack vs. scratch buffer

ryantimwilson:
This PR adds a configurable limit `on_stack_limit` - default = 32 - to allow the user to specify whether tuples/fmt strings/str calls (and soon map keys/values) should be stored on the stack or in a scratch buffer. This allows us to balance supporting large strings/objects while avoiding making the ResourceAnalyser code too complex by trying to accurately estimating memory for map values/keys: https://github.com/bpftrace/bpftrace/pull/3483

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -3572,6 +3572,14 @@ The possible options are:
 - `warn` - print a warning but continue execution
 - `ignore` - silently ignore missing probes
 
+=== on_stack_limit
+
+Default: 32
+
+The maximum size (in bytes) of individual objects that will be stored on the BPF stack. If they are larger than this limit they will be stored in pre-allocated memory.
+
+This exists because the BPF stack is limited to 512 bytes and large objects make it more likely that we'll run out of space. bpftrace can store objects that are larger than the `on_stack_limit` in pre-allocated memory to prevent this stack error. However, storing in pre-allocated memory may be less memory efficient. Lower this default number if you are still seeing a stack memory error or increase it if you're worried about memory consumption.
+
 === perf_rb_pages
 
 Default: 64
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -475,24 +475,57 @@ CallInst *IRBuilderBPF::CreateGetStackScratchMap(StackType stack_type,
                              failure_callback);
 }
 
-Value *IRBuilderBPF::CreateGetStrScratchBuffer(const location &loc, int key)
+Value *IRBuilderBPF::CreateGetStrAllocation(const std::string &name,
+                                            const location &loc,
+                                            AsyncIds &async_ids)
 {
-  return createScratchBuffer(bpftrace::globalvars::GlobalVar::GET_STR_BUFFER,
-                             loc,
-                             key);
+  const auto max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
+  const auto str_type = CreateArray(max_strlen, CreateInt8());
+  return createAllocation(bpftrace::globalvars::GlobalVar::GET_STR_BUFFER,
+                          GetType(str_type),
+                          name,
+                          loc,
+                          [&async_ids] { return async_ids.str(); });
 }
 
-Value *IRBuilderBPF::CreateGetFmtStringArgsScratchBuffer(const location &loc)
+Value *IRBuilderBPF::CreateGetFmtStringArgsAllocation(StructType *struct_type,
+                                                      const std::string &name,
+                                                      const location &loc)
 {
-  return createScratchBuffer(
-      bpftrace::globalvars::GlobalVar::FMT_STRINGS_BUFFER, loc);
+  return createAllocation(bpftrace::globalvars::GlobalVar::FMT_STRINGS_BUFFER,
+                          struct_type,
+                          name,
+                          loc);
 }
 
-Value *IRBuilderBPF::CreateTupleScratchBuffer(const location &loc, int key)
+Value *IRBuilderBPF::CreateTupleAllocation(const SizedType &tuple_type,
+                                           const std::string &name,
+                                           const location &loc,
+                                           AsyncIds &async_ids)
 {
-  return createScratchBuffer(bpftrace::globalvars::GlobalVar::TUPLE_BUFFER,
-                             loc,
-                             key);
+  return createAllocation(bpftrace::globalvars::GlobalVar::TUPLE_BUFFER,
+                          GetType(tuple_type),
+                          name,
+                          loc,
+                          [&async_ids] { return async_ids.tuple(); });
+}
+
+Value *IRBuilderBPF::createAllocation(
+    bpftrace::globalvars::GlobalVar globalvar,
+    llvm::Type *obj_type,
+    const std::string &name,
+    const location &loc,
+    std::optional<std::function<size_t()>> gen_async_id_cb)
+{
+  const auto obj_size = module_.getDataLayout().getTypeAllocSize(obj_type);
+  const auto on_stack_limit = bpftrace_.config_.get(
+      ConfigKeyInt::on_stack_limit);
+  if (obj_size > on_stack_limit) {
+    return createScratchBuffer(globalvar,
+                               loc,
+                               gen_async_id_cb ? (*gen_async_id_cb)() : 0);
+  }
+  return CreateAllocaBPF(obj_type, name);
 }
 
 Value *IRBuilderBPF::createScratchBuffer(
--- a/src/ast/passes/codegen_llvm.h
+++ b/src/ast/passes/codegen_llvm.h
@@ -95,6 +95,7 @@ public:
   Value *createTuple(
       const SizedType &tuple_type,
       const std::vector<std::pair<llvm::Value *, const location *>> &vals,
+      const std::string &name,
       const location &loc);
   void createTupleCopy(const SizedType &expr_type,
                        const SizedType &var_type,
--- a/tests/codegen/scratch_buffer.cpp
+++ b/tests/codegen/scratch_buffer.cpp
@@ -1,0 +1,70 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+static constexpr uint64_t SMALL_ON_STACK_LIMIT = 0;
+static constexpr uint64_t LARGE_ON_STACK_LIMIT = 128;
+static constexpr uint64_t MAX_STRLEN = 64;
+
+static void test_stack_or_scratch_buffer(const std::string &input,
+                                         const std::string &name,
+                                         uint64_t on_stack_limit)
+{
+  auto bpftrace = get_mock_bpftrace();
+  bpftrace->feature_ = std::make_unique<MockBPFfeature>(true);
+  auto configs = ConfigSetter(bpftrace->config_, ConfigSource::script);
+  configs.set(ConfigKeyInt::on_stack_limit, on_stack_limit);
+  configs.set(ConfigKeyInt::max_strlen, MAX_STRLEN);
+
+  bpftrace->safe_mode_ = true;
+
+  test(*bpftrace, input, name);
+}
+
+TEST(codegen, tuple_scratch_buf)
+{
+  test_stack_or_scratch_buffer("kprobe:f { (1, \"xxxx\") }",
+                               NAME,
+                               SMALL_ON_STACK_LIMIT);
+}
+
+TEST(codegen, tuple_stack)
+{
+  test_stack_or_scratch_buffer("kprobe:f { (1, \"xxxx\") }",
+                               NAME,
+                               LARGE_ON_STACK_LIMIT);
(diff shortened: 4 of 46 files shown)


src/main.cpp
```diff
@@ -138,6 +138,7 @@ void usage()
   std::cerr << "    BPFTRACE_MAX_TYPE_RES_ITERATIONS  [default: 0] number of levels of nested field accesses for tracepoint args" << std::endl;
   std::cerr << "    BPFTRACE_PERF_RB_PAGES            [default: 64] pages per CPU to allocate for ring buffer" << std::endl;
   std::cerr << "    BPFTRACE_STACK_MODE               [default: bpftrace] Output format for ustack and kstack builtins" << std::endl;
+  std::cerr << "    BPFTRACE_STACK_THRESHOLD          [default: 32] max bytes to store objects on BPF stack vs. global variable" << std::endl;
```
jordalgo, line 141:
```suggestion
  std::cerr << "    BPFTRACE_STACK_THRESHOLD          [default: 32] max bytes of individual objects to store them on the BPF stack vs. a global variable" << std::endl;
```

man/adoc/bpftrace.adoc
```diff
@@ -3571,6 +3571,15 @@ Available modes/formats:
 
 This can be overwritten at the call site.
 
+=== stack_threshold
+
+Default: 32
+
+The maximum length (in bytes) where objects will be stored on the BPF stack instead of a scratch global variable.
```
jordalgo, line 3578:
```suggestion
The maximum size (in bytes) of individual objects that will be stored on the BPF stack. If they are larger than this threshold they will be stored in a scratch global variable.
```

tests/codegen/scratch_buffer.cpp
```diff
@@ -0,0 +1,24 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, big_tuple_scratch_buf)
+{
+  test("kprobe:f { (1, \"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\") }",
```
jordalgo, line 9: Maybe these tests should explicitly set the config variable so if we change the default they don't change. And maybe have two codegens (under and over threshold) for each fmt and tuples?

jordalgo (commented):
Few small things but overall looks good.

src/main.cpp
danobi: "global variable" kinda leaks an unnecessary implementation detail. Maybe say "preallocated memory"?

danobi: nit: "stack threshold" seems backwards (implying a minimum for the stack). maybe "offstack threshold"?

src/ast/passes/codegen_llvm.h
```diff
@@ -248,6 +249,13 @@ class CodegenLLVM : public Visitor {
 
   Value *createFmtString(int print_id);
 
+  Value *createTupleAllocation(const SizedType &tuple_type,
+                               const std::string &name,
+                               const location &loc);
+  Value *createGetFmtStringArgsAllocation(StructType *struct_type,
+                                          const std::string &name,
+                                          const location &loc);
+
```
danobi, line 258: Wondering if there's a way to prevent proliferation of wrappers like this. Seems like it would be good to keep the "stack or offstack" decision in a central place so new offstack types can't easily forget to check on stack_threshold.

I can't think of any better ideas at the moment so just consider this a non-blocking idea to ponder.

danobi (commented):
overall lgtm.

tests/codegen/scratch_buffer.cpp
danobi: +1, testing both paths is good idea

src/main.cpp
jordalgo: I think maybe both are ambiguous. How about "on_stack_limit" or "on_stack_max" or "off_stack_min"

src/ast/passes/codegen_llvm.h
ryantimwilson: Yea the problem is we depend on iterating `async_ids` if using scratch, else not iterating it. So its hard to move into irbuilderbpf.cpp. I guess we could do this with a callback...let me experiment with that approach.

src/main.cpp
ryantimwilson: Lets discuss naming offline

ryantimwilson: on_stack_limit seems to be the choice, lets go with that

ryantimwilson:
@danobi @jordalgo 

Updates:
1. Changed name to `off_stack_limit` and updated relevant descriptions
2. Updated strings to use stack if max_strlen > on_stack_limit
3. Added codegen tests for stack vs. scratch buffer
4. Refactored to use common stack vs. scratch buffer function in irbuilderbpf

src/ast/passes/codegen_llvm.h
ryantimwilson: Addressed in latest version using callbacks

tests/codegen/scratch_buffer.cpp
ryantimwilson: Done in latest!

man/adoc/bpftrace.adoc
```diff
@@ -3556,6 +3556,16 @@ The possible options are:
 - `warn` - print a warning but continue execution
 - `ignore` - silently ignore missing probes
 
+=== on_stack_limit
+
+Default: 32
+
+The maximum size (in bytes) of individual objects that will be stored on the BPF stack. If they are larger than this threshold they will be stored in
```
jordalgo, line 3563: nit
```suggestion
The maximum size (in bytes) of individual objects that will be stored on the BPF stack. If they are larger than this limit they will be stored in
```

man/adoc/bpftrace.adoc
```diff
@@ -3556,6 +3556,16 @@ The possible options are:
...
+pre-allocated memory.
+
+This threshold is necessary because the BPF stack is limited to 512 bytes and large objects make it more likely for the stack to run out of space. Instead,
+bpftrace can store these large objects in read-write global variables. However, this involves pre-allocating memory so it may be less memory efficient.
+
```
jordalgo, line 3568: nit: slight re-wording

```suggestion
This exists because the BPF stack is limited to 512 bytes and large objects make it more likely that we'll run out of space. bpftrace can store objects that are larger than the `on_stack_limit` in pre-allocated memory to prevent this stack error. However, storing in pre-allocated memory may be less memory efficient. Lower this default number if you are still seeing a stack memory error or increase it if you're worried about memory consumption.`

```

src/ast/irbuilderbpf.cpp
```diff
@@ -475,24 +475,61 @@ CallInst *IRBuilderBPF::CreateGetStackScratchMap(StackType stack_type,
                              failure_callback);
 }
 
-Value *IRBuilderBPF::CreateGetStrScratchBuffer(const location &loc, int key)
-{
-  return createScratchBuffer(bpftrace::globalvars::GlobalVar::GET_STR_BUFFER,
-                             loc,
-                             key);
+Value *IRBuilderBPF::CreateGetStrStackOrScratchBufferAllocation(
```
jordalgo, line 478: oofa these function names - what is this Java?. Is something like `CreateGetStrAllocation` not accurate or descriptive enough?

src/ast/irbuilderbpf.cpp
```diff
@@ -475,24 +475,61 @@ CallInst *IRBuilderBPF::CreateGetStackScratchMap(StackType stack_type,
...
+      [&async_ids] { return async_ids.tuple(); });
 }
 
-Value *IRBuilderBPF::CreateTupleScratchBuffer(const location &loc, int key)
+Value *IRBuilderBPF::createStackOrScratchBufferAllocation(
+    bpftrace::globalvars::GlobalVar globalvar,
+    llvm::Type *obj_type,
+    const std::string &name,
+    const location &loc,
+    std::optional<std::function<size_t()>> key_cb)
```
jordalgo, line 524: naming nit
```suggestion
    std::optional<std::function<size_t()>> gen_async_id_cb)
```

jordalgo (commented):
Code looks good, just a few small doc updates and naming nits/questions.

jordalgo:
> Changed name to off_stack_limit and updated relevant descriptions

Also, you mean `on_stack_limit`

src/ast/irbuilderbpf.cpp
ryantimwilson: lol yes these are too wordy...i debated changing it and I should've. Yes lets go with Allocation

jordalgo (approved):
LGTM. Nice work!

danobi (approved):

src/main.cpp
```diff
@@ -136,6 +136,7 @@ void usage()
   std::cerr << "    BPFTRACE_MAX_PROBES               [default: 512] max number of probes" << std::endl;
   std::cerr << "    BPFTRACE_MAX_STRLEN               [default: 64] bytes on BPF stack per str()" << std::endl;
   std::cerr << "    BPFTRACE_MAX_TYPE_RES_ITERATIONS  [default: 0] number of levels of nested field accesses for tracepoint args" << std::endl;
+  std::cerr << "    BPFTRACE_ON_STACK_LIMIT           [default: 32] max bytes of individual objects to store them on the BPF stack vs. pre-allocated memory" << std::endl;
```
ajor, line 139: I don't think this needs to be an environment variable as well as a config option. It seems unlikely that someone would want to change this setting globally for all scripts on their system - it's a thing that'd be very specific to a single script. (this is true for a lot of our env vars, but they exist as historical baggage since they were around before config options were a thing)

tests/runtime/call
```diff
@@ -109,6 +109,24 @@ AFTER ./testprogs/syscall execve /$(python3 -c "print('X'*5555)")
...
+NAME str_scratch_buf
+PROG t:syscalls:sys_enter_execve { print(str(args.filename)) }
+ENV BPFTRACE_ON_STACK_LIMIT=0
+EXPECT /XXXX
+AFTER ./testprogs/syscall execve /$(python3 -c "print('XXXX')")
+REQUIRES_FEATURE probe_read_kernel
+TIMEOUT 1
+
+NAME fmt_str_scratch_buf
+PROG BEGIN { printf("%s %d", "XXXX", 1) }
```
ajor, line 126:
```suggestion
PROG BEGIN { printf("%s %d\n", "XXXX", 1) }
```
Nit: newline to avoid buffering issues (it caught me out when running this script locally at the least)

tests/runtime/call
```diff
@@ -109,6 +109,24 @@ AFTER ./testprogs/syscall execve /$(python3 -c "print('X'*5555)")
...
+NAME str_tuple_scratch_buf
+PROG BEGIN { print((1, (2, "XXXX"))) }
+ENV BPFTRACE_ON_STACK_LIMIT=0
+EXPECT (1, (2, XXXX))
+
+NAME str_scratch_buf
+PROG t:syscalls:sys_enter_execve { print(str(args.filename)) }
+ENV BPFTRACE_ON_STACK_LIMIT=0
+EXPECT /XXXX
+AFTER ./testprogs/syscall execve /$(python3 -c "print('XXXX')")
```
ajor, line 121:
```suggestion
AFTER ./testprogs/syscall execve /XXXX
```
Is the Python part necessary?

tests/runtime/call
```diff
@@ -109,6 +109,24 @@ AFTER ./testprogs/syscall execve /$(python3 -c "print('X'*5555)")
...
+REQUIRES_FEATURE probe_read_kernel
+TIMEOUT 1
+
+NAME fmt_str_scratch_buf
```
ajor, line 125: Sorry, I've probably picked the wrong PR to bring this up on, but the name "fmt_str_scratch_buf" is confusing me. From playing around with it a little, it seems like it's not the format string that's being put into a scratch buf, but rather all the other arguments to `printf`. The format string never appears in BPF code IIUC.

In the case where we're keeping data on the stack, the name "printf_args" is used in place of "fmtstr", which makes more sense to me.

tests/runtime/call
ryantimwilson: No I will fix

tests/runtime/call
ryantimwilson: hmm interesting, I didn't hit it but will fix, thanks!

src/main.cpp
ryantimwilson: ack will remove env var

tests/runtime/call
ryantimwilson: technically its called fmt_str_args_scratch_buf, I just fudged the naming here: https://github.com/bpftrace/bpftrace/blob/57b355c95b5406742edeb4b99330e2253703615b/src/required_resources.h#L104

I'll fix the naming here.

I kinda like format string args to printf args but if we want to change it, I prefer we do in another PR

ryantimwilson:
@ajor addressed your comments, mind taking another look?

PR #3515, opened 2024-10-13
Support symbolizing enums

### LINKED ISSUE #2772: Support symbolizing enum values

Consider:
```
$ sudo bpftrace -e 'tracepoint:skb:kfree_skb { print(args->reason) }'
Attaching 1 probe...
50
50
```

bpftrace knows the type of `args->reason`:
```
$ sudo bpftrace -lv tracepoint:skb:kfree_skb
tracepoint:skb:kfree_skb
    void * skbaddr
    void * location
    unsigned short protocol
    enum skb_drop_reason reason
```

as well as the names of the values:
```
 $ sudo bpftrace -lv 'enum skb_drop_reason' | grep 50
        SKB_DROP_REASON_UNHANDLED_PROTO = 50,
```

Would be nice if bpftrace could be printing "SKB_DROP_REASON_UNHANDLED_PROTO" instead of integer values.

Ideally we just do it and tell users who want the integer values to cast the enum to an int. But that might not be backwards compatible.

 Could reuse `str()` but overload on identifiers. That's kinda ugly though IMO. Could create a new helper. Also prefer to avoid that. Wonder how many people are depending on enums being printed as an integer. Seems like it's just an annoyance for everyone.
Maybe we can have an environ variable to control this? Not sure.

danobi:
This PR allows the user to symbolize enums in userspace by
using `%s` printf() specifier. This is a fully backwards compatible
change, as trying to use `%s` on an enum was a compile error
before.

Example:
```
# bpftrace -e 'tracepoint:skb:kfree_skb { $r = args->reason; printf("%d, %s\n", $r, $r); exit() }'
Attaching 1 probe...
2, SKB_DROP_REASON_NOT_SPECIFIED
```

Note this PR does not support casting integers to enums. While
this is technically possible (and described in the linked ticket), it's
probably not necessary as users can simply perform arithmetic
operations on an enum and get the same result. There's currently
a limitation but it's being fixed in #3517.

This is easier to review per-commit.

This closes #2772.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1949,7 +1949,27 @@ The non-standard ones can be found in the table below:
 
 |===
 
-Supported escape sequences
+`printf()` can also symbolize enums as strings. User defined enums as well as enums
+defined in the kernel are supported. For example:
+
+----
+enum custom {
+  CUSTOM_ENUM = 3,
+};
+
+BEGIN {
+  $r = SKB_DROP_REASON_SOCKET_FILTER;
+  printf("%d, %s, %s\n", $r, $r, CUSTOM_ENUM);
+  exit();
+}
+----
+
+yields:
+
+----
+6, SKB_DROP_REASON_SOCKET_FILTER, CUSTOM_ENUM
+----
+
 
 Colors are supported too, using standard terminal escape sequences:
 
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -182,8 +182,8 @@ public:
   BpfBytecode bytecode_;
   StructManager structs;
   std::map<std::string, std::string> macros_;
-  // Map of enum variant_name to variant_value
-  std::map<std::string, uint64_t> enums_;
+  // Map of enum variant_name to (variant_value, enum_name)
+  std::map<std::string, std::tuple<uint64_t, std::string>> enums_;
   // Map of enum_name to map of variant_value to variant_name
   std::map<std::string, std::map<uint64_t, std::string>> enum_defs_;
   std::map<libbpf::bpf_func_id, location> helper_use_loc_;
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -182,7 +182,10 @@ public:
   BpfBytecode bytecode_;
   StructManager structs;
   std::map<std::string, std::string> macros_;
+  // Map of enum variant_name to variant_value
   std::map<std::string, uint64_t> enums_;
+  // Map of enum_name to map of variant_value to variant_name
+  std::map<std::string, std::map<uint64_t, std::string>> enum_defs_;
   std::map<libbpf::bpf_func_id, location> helper_use_loc_;
   const FuncsModulesMap &get_traceable_funcs() const;
   KConfig kconfig;
--- a/src/format_string.cpp
+++ b/src/format_string.cpp
@@ -4,6 +4,7 @@
 #include "utils.h"
 
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 namespace bpftrace {
@@ -105,16 +106,26 @@ std::string validate_format_string(const std::string &fmt,
   }
 
   for (int i = 0; i < num_args; i++) {
+    std::unordered_set<Type> arg_types;
     Type arg_type = args.at(i).type.GetTy();
     if (arg_type == Type::ksym_t || arg_type == Type::usym_t ||
         arg_type == Type::probe || arg_type == Type::username ||
         arg_type == Type::kstack_t || arg_type == Type::ustack_t ||
         arg_type == Type::inet || arg_type == Type::timestamp ||
         arg_type == Type::mac_address || arg_type == Type::cgroup_path_t ||
-        arg_type == Type::strerror_t)
-      arg_type = Type::string; // Symbols should be printed as strings
-    if (arg_type == Type::pointer)
-      arg_type = Type::integer; // Casts (pointers) can be printed as integers
+        arg_type == Type::strerror_t) {
+      arg_types.insert(Type::string); // Symbols should be printed as
+                                      // strings
+    } else if (arg_type == Type::pointer) {
+      arg_types.insert(Type::integer); // Casts (pointers) can be
+                                       // printed as integers
+    } else if (args.at(i).type.IsEnumTy()) {
+      // Symbolizing enum
+      arg_types.insert(Type::string);
+      arg_types.insert(Type::integer);
+    } else {
+      arg_types.insert(arg_type);
+    }
 
     auto token = std::get<0>(token_types[i]);
     auto token_type = std::get<1>(token_types[i]);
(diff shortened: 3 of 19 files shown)


ajor (changes requested):
Going back to our discussions in the issue: I'm not sure that this PR solves enough of the problem to be useful, and by defining enums as integers, it risks locking us into behaviour that we won't be able to change to fix it in the future.

We need to be able to symbolicate enums everywhere that they can appear:
- as a field in a struct: `{ .myenum = ABC }`
- in tuples: `(SKB_CONSUMED, 123)`
- as map keys or values: `@[SKB_CONSUMED] = ABC`

danobi:
> Going back to our discussions in the issue: I'm not sure that this PR solves enough of the problem to be useful,

I think it's a good start. A few users have asked me for this already. Let me follow up and collect their feedback on this approach.

> and by defining enums as integers, it risks locking us into behaviour that we won't be able to change to fix it in the future.

I think it might be too late for enums as they exist today. Plenty of people are using today's enums as ints. But the good news is I haven't seen it be a problem yet in the ~5 years they've been in: https://github.com/bpftrace/bpftrace/commit/e4cb6ce47018bf857ed0dbd0f0e01ce4e9cb63c9

> We need to be able to symbolicate enums everywhere that they can appear:

For sure, agreed. But I think we need to start somewhere and printf() is the highest impact for now. We'll need the metadata collected in commits 1-6 to keep building on top.

danobi:
@ajor and I chatted offline. Decision is that this approach is ok. We'll decide how to extend enum symbolication to struct fields, tuples, and map key/value next (or even at some other later point). Gut feeling is map key is next most useful.

It won't be hard to do when the infra in this PR is landed. Just a question of how much backwards compat we want. For example, changing human readable output is probably fine. But we probably want to keep JSON output as-is so tooling doesn't silently break.

src/format_string.cpp
```diff
@@ -39,6 +41,37 @@ ArgumentType get_expected_argument_type(const std::string &fmt)
   return ArgumentType::UNKNOWN;
 }
 
+// Returns a vector of (token, type) tuples given a format string
+std::vector<std::tuple<std::string, Type>> get_token_types(
```
jordalgo, line 45: nit: Maybe a few unit tests for this would be good.

src/format_string.h
```diff
@@ -34,8 +37,12 @@ class FormatString {
    */
   FormatString() = default;
 
-  FormatString(const char *s) : fmt_(s){};
-  FormatString(std::string &s) : fmt_(s){};
+  FormatString(const char *s) : fmt_(s)
+  {
```
jordalgo, line 41: why this change?

src/format_string.h
```diff
@@ -82,14 +89,15 @@ class FormatString {
...
 
   friend class cereal::access;
 
   template <typename Archive>
   void serialize(Archive &ar)
   {
-    // NOTE: parts_ is not constructed until first use, so no point in
-    // serializing it
+    // NOTE: parts_, expected_types_, and tokens_ are not constructed until
+    // first use, so no point in serializing it
```
jordalgo, line 100: "so no point in serializing them"

tests/runtime/call
```diff
@@ -30,6 +30,28 @@ NAME printf_char
...
+
+NAME printf_enum_symbolize_map
+PROG enum Foo { ONE = 1, TWO = 2, OTHER = 99999 }; BEGIN { @one = ONE; @two = TWO; @other = OTHER; printf("%d %s %d %s %d %s\n", @one, @one, @two, @two, @other, @other); exit() }
+EXPECT 1 ONE 2 TWO 99999 OTHER
+
+NAME printf_enum_symbolize_width
+PROG enum Foo { A, B, C, }; BEGIN { printf("%-5s %5s %s\n", A, B, C); exit() }
+EXPECT A         B C
+
+NAME printf_enum_symbolize_tracepoint
```
jordalgo, line 49: should we add a test for trying to symbolize an int that isn't an enum?

man/adoc/bpftrace.adoc
```diff
@@ -1922,7 +1922,26 @@ The non-standard ones can be found in the table below:
...
+  CUSTOM_ENUM = 3,
+};
+
+tracepoint:skb:kfree_skb {
+  $r = args->reason;
+  printf("%d, %s, %s\n", $r, $r, CUSTOM_ENUM);
+}
+----
+
+may yield:
```
jordalgo, line 1938: This feels confusing. Why not just use a `BEGIN` probe so you can say what it definitely will print out.

jordalgo:
> But we probably want to keep JSON output as-is so tooling doesn't silently break.

Would be nice if we could provide a way to print the symbolized enums in JSON output though - env variable?

src/format_string.h
danobi: dunno. clang-format really wants to do this and it won't let me leave it unchanged :P

tests/runtime/call
danobi: Yeah makes sense. I'll add it to semantic analyser tests cuz it should be a compile failure

man/adoc/bpftrace.adoc
danobi: Yeah, agreed. Simplified.

src/format_string.cpp
danobi: Took a look. Could probably be done, but would need to pull this out of anon namespace. We also already have quite a bit of coverage in semantic analyser tests. See `semantic_analyser.printf_*` set of tests. So I think coverage is likely sufficient here. WDYT?

danobi:
> > But we probably want to keep JSON output as-is so tooling doesn't silently break.
> 
> Would be nice if we could provide a way to print the symbolized enums in JSON output though - env variable?

That was referring to changing the output for map keys/values, tuples, and structs. I think changing the human output is fine but JSON is dicey. As folks may be parsing it in prod already.

For regular printf() (added in this PR), JSON output changes "correctly":

```
$ sudo ./build/src/bpftrace -e 'BEGIN { printf("%s\n", SKB_DROP_REASON_NOT_SPECIFIED); exit() }' -f json
{"type": "printf", "data": "SKB_DROP_REASON_NOT_SPECIFIED\n"}
```

jordalgo:
> I think changing the human output is fine but JSON is dicey. As folks may be parsing it in prod already. 

I guess we'll see if we get requests from users to make this configurable.

src/bpftrace.h
```diff
@@ -182,7 +182,10 @@ class BPFtrace {
   BpfBytecode bytecode_;
   StructManager structs;
   std::map<std::string, std::string> macros_;
-  std::map<std::string, uint64_t> enums_;
+  // Map of enum variant_name -> (variant_value, enum_name)
```
jordalgo, line 185: super nit (just language consistency)
```suggestion
  // Map of enum variant_name to (variant_value, enum_name)
```

src/types.h
```diff
@@ -302,7 +303,7 @@ class SizedType {
 
   const std::string GetName() const
   {
-    assert(IsRecordTy());
+    assert(IsRecordTy() || IsIntTy());
```
jordalgo, line 306: Nit. Does this work?
```suggestion
    assert(IsRecordTy() || IsEnumTy());
```

jordalgo (approved):
One super nit and one question but otherwise LGTM.

src/types.h
danobi: Ah yeah that's better

PR #3517, opened 2024-10-13
ints: Support subword binary operations

danobi:
Previously, all binary operations result in a 64 bit signed integer.
This causes issues when you try to do things like:

    $x = (uint32)5; $x |= 1;

Even explicitly casting the literal does not help:

    $x = (uint32)5; $x |= (uint32)1;

Both result in the error:

    Integer size mismatch. Assignment type 'uint64' is larger than the
    variable type 'uint32'.

This commit teaches semantic analyser and codegen to promote to only the
necessary size and not always 64 bits.

This closes https://github.com/bpftrace/bpftrace/issues/3516.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,8 @@ and this project adheres to
   - [#3453](https://github.com/bpftrace/bpftrace/pull/3453)
 - Fix json formatting for hex values
   - [#3475](https://github.com/bpftrace/bpftrace/pull/3475)
+- Fix binary operations on integers always returning 64 bit values
+  - [#3517](https://github.com/bpftrace/bpftrace/pull/3517)
 #### Security
 #### Docs
 - Remove mention of unsupported character literals
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -675,7 +675,15 @@ The following operators are available for integer arithmetic:
 
 |===
 
-// TODO: Words about integer conversion when types mismatch
+Operations between a signed and an unsigned integer are allowed providing
+bpftrace can statically prove a safe conversion is possible. If safe conversion
+is not guaranteed, the operation is undefined behavior and a corresponding
+warning will be emitted.
+
+If the two operands are different size, the smaller integer is implicitly
+promoted to the size of the larger one. Sign is preserved in the promotion.
+For example, `(uint32)5 + (uint8)3` is converted to `(uint32)5 + (uint32)3`
+which results in `(uint32)8`.
 
 ==== Logical Operators
 
@@ -2553,7 +2561,7 @@ kprobe:dummy {
   if (has_key(@scalar)) { // error
     print(("hello"));
   }
-  
+
   $a = has_key(@associative, (1,2)); // ok
   @b[has_key(@associative, (1,2))] = has_key(@associative, (1,2)); // ok
 }
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1624,6 +1624,18 @@ void CodegenLLVM::binop_buf(Binop &binop)
   expr_ = b_.CreateStrncmp(left_string, right_string, len, inverse);
 }
 
+// Looks like LLVM <= 14 IR verifier doesn't like when you assign an i1 into an
+// i64. So zero extend there but don't penalize newer LLVM releases.
+//
+// Just LSB needs to be set for bools. If we do a sign extend then all bits
+// will be set (which performs correctly) but printing out the value will be -1
+// which is confusing. So zero extend instead.
+#if LLVM_VERSION_MAJOR <= 14
+#define MAYBE_ZERO_EXTEND(e) b_.CreateIntCast(e, b_.getInt64Ty(), false)
+#else
+#define MAYBE_ZERO_EXTEND(e) e
+#endif
+
 void CodegenLLVM::binop_int(Binop &binop)
 {
   Value *lhs, *rhs;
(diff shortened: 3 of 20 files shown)


CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ and this project adheres to
   - [#3453](https://github.com/bpftrace/bpftrace/pull/3453)
 - Fix json formatting for hex values
   - [#3475](https://github.com/bpftrace/bpftrace/pull/3475)
+- Fix binary operations on integers always returning 64 bit values
```
jordalgo, line 84: If this PR was the cause then it hasn't been release yet and this probably doesn't need a changelog entry: https://github.com/bpftrace/bpftrace/pull/3416

tests/semantic_analyser.cpp
```diff
@@ -2804,6 +2826,11 @@ kprobe:f { $x = -1; $x = 10223372036854775807; }
 stdin:1:35-56: ERROR: Tuple type mismatch: (int64,uint32) != (int64,int32).
 kprobe:f { $x = (0, (uint32)123); $x = (0, (int32)-123); }
                                   ~~~~~~~~~~~~~~~~~~~~~
+)");
+  test_error("kprobe:f { $x = (uint32)5; $x += 1; }", R"(
```
jordalgo, line 2830: This should not be an error IMHO. We are able to tell statically that `1` is within the range of an unsigned 32 bit int and case it to the correct integer size (this is what we do for variable assignments).

CHANGELOG.md
danobi: While the referenced PR does cause the breakage, I think this was already the case. So a Fixes: entry makes sense to me:

```
$ sudo bpftrace -e 'BEGIN { print(sizeof((uint32)5 + (uint32)5)) }'
Attaching 1 probe...
8
^C
```

tests/semantic_analyser.cpp
danobi: I agree this ideally should be supported. But the issue is the "can fit" checks are all behind `assignment.expr->is_literal`. And while this looks like a literal, it's actually not. The parser desugars this to `AssignVarStatement(Variable, Binop(Variable, Integer(1)))`. So we've lost that data. Unclear to me if it can be easily recovered. Maybe worth a try? But gut feeling is it may  not be worthwhile.

jordalgo: > And while this looks like a literal, it's actually not.

Ah that's a bummer. Ok, I won't block on this then. I'll also poke around and see if I can make integers more flexible: https://github.com/bpftrace/bpftrace/issues/3518

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1706,7 +1721,13 @@ void CodegenLLVM::binop_int(Binop &binop)
     default:
       LOG(BUG) << "\"" << opstr(binop) << "\" was handled earlier";
   }
-  // Using signed extension will result in -1 which will likely confuse users
+
+  return;
+
+sign_extend_bool:
+  // Just the bottom bit needs to be set for bools. If we do a signed sign
```
jordalgo, line 1728: a signed sign?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1706,7 +1721,13 @@ void CodegenLLVM::binop_int(Binop &binop)
     default:
       LOG(BUG) << "\"" << opstr(binop) << "\" was handled earlier";
   }
-  // Using signed extension will result in -1 which will likely confuse users
+
+  return;
+
+sign_extend_bool:
```
jordalgo, line 1727: Does it have to be a goto? Why not just do this in place of the goto?
`expr_ = b_.CreateIntCast(expr_, b_.getInt64Ty(), false);`

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1706,7 +1721,13 @@ void CodegenLLVM::binop_int(Binop &binop)
     default:
       LOG(BUG) << "\"" << opstr(binop) << "\" was handled earlier";
   }
-  // Using signed extension will result in -1 which will likely confuse users
+
+  return;
```
jordalgo, line 1725: Also seems like this early return changes the expression for other operations. Is that intentional?

tests/codegen/binop_int_promotion.cpp
```diff
@@ -0,0 +1,14 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, binop_int_promotion)
+{
+  test("kretprobe:f { $x = (uint32)5; $x += (uint32)1 }", NAME);
```
viktormalik, line 9: Technically speaking, there's no promotion here. What about doing
```suggestion
  test("kretprobe:f { $x = (uint32)5; $x += (uint16)1 }", NAME);
```
to check that the upcast is correctly generated?

viktormalik (commented):
Maybe also add `(uint32)5 + (uint16)1` runtime test?

src/ast/passes/codegen_llvm.cpp
danobi: Fixed other typo too. And tightened up the language

src/ast/passes/codegen_llvm.cpp
danobi: Ack. Will just inline it.

tests/codegen/binop_int_promotion.cpp
danobi: Yeah good point. I think I just named it poorly - idea was to test the regression. But this catches more.

tests/semantic_analyser.cpp
danobi: Gonna move this to a commented out "positive" test. Having a test like this encodes wrong intention. It'll seem to suggest we want it this way. Rather than it being a limitation.

man/adoc/bpftrace.adoc
```diff
@@ -675,7 +675,15 @@ The following operators are available for integer arithmetic:
 
 |===
 
-// TODO: Words about integer conversion when types mismatch
+Operations between a signed and an unsigned integer are allowed providing
+bpftrace can statically prove a safe conversion is possible. If safe conversion
+is not guaranteed, the operation is undefined behavior and a corresponding
+warning will be emitted.
```
jordalgo, line 681: Do we mean error here instead of warning?

danobi: No, mixing signed and unsigned currently emits a warning:

```
$ sudo ./build/src/bpftrace -e 'BEGIN { $s = -5; $u = (uint64) 5; $s == $u }'
stdin:1:38-40: WARNING: comparison of integers of different signs: 'int64' and 'uint64' can lead to undefined behavior
BEGIN { $s = -5; $u = (uint64) 5; $s == $u }
                                     ~~
Attaching 1 probe...
^C
```

jordalgo (approved):
LGTM

viktormalik (approved):
LGTM

PR #3520, opened 2024-10-14
has_key: only skip key validation for first arg

jordalgo:
Similar to the new `delete` API, `has_key` expects only a map reference as the first argument but the second arg should be treated normally and have map key validation.

##### Checklist


viktormalik (approved):
Makes sense based on the test.

ajor (approved):

Issue #3500, opened 2024-10-08
Add Signal Handler Probe Type


ajor:
In #2203, a signal handler for SIGUSR1 was added, which prints all maps when the signal is received. Note: it appears that the docs for this feature have been deleted - see the original PR for reference instead.

This is a proposal to extend the signal handler with the ability to run user-defined BpfScript code. Instead of printing all maps, we can attach a new probe type to the signal, for example (exact syntax TBD):
```
signal:sigusr1
{
  printf("custom signal handler\n");
}
```

Now when bpftrace receives a SIGUSR1, it'll print the custom message.

If a custom signal handling probe is not defined then we'll fall back to the current behaviour of printing all maps.

### Why?

Impatient users might want to see some sort of status from their script. The options for this now are:
- `END` - not useful if we want our script to continue running after displaying its status
- `interval` - fixed intervals can be either too frequent or to infrequent to be useful to an interactive user. It'd be nicer to have a way to get output exactly when it's wanted.

jordalgo:
Being able to interact with a live script sounds cool.

ajor:
It'd also be nice to be able to do something like this if someone presses the space bar, for example. I'm not sure how we'd make that work though.

viktormalik:
> It'd also be nice to be able to do something like this if someone presses the space bar, for example. I'm not sure how we'd make that work though.

We'd have to hook the kernel function which handles key presses. It won't be probably stable enough to hard-code it in bpftrace but we could provide a code snippet in the docs. Something like "if you want to print maps on space bar press, use this". Eventually, it could become a part of the standard library.

fbs:
> > It'd also be nice to be able to do something like this if someone presses the space bar, for example. I'm not sure how we'd make that work though.
> 
> We'd have to hook the kernel function which handles key presses. It won't be probably stable enough to hard-code it in bpftrace but we could provide a code snippet in the docs. Something like "if you want to print maps on space bar press, use this". Eventually, it could become a part of the standard library.

why not just stdin?

ajor:
Stdin would be nice, but we'd need to define some form of probe type that could execute arbitrary BpfScript code on the input event. Just printing all maps is a fine default, but too limiting in general. For example (not that I've thought about much the exact syntax):
```
stdin:space
{
  printf("space pressed!\n");
}
```

amiremohamadi:
the stdin probe would kind of like attaching to `kprobe:pty_write`?

ajor:
[...]

PR #3522, opened 2024-10-16
Add custom signal handling probe type

amiremohamadi:
#3500

##### Checklist


man/adoc/bpftrace.adoc
```diff
@@ -2841,6 +2846,21 @@ END {
...
 ----
 
+[#probes-signal]
+=== signal
+
+.variants
+* `signal:sigusr1`
+
+Signal probes are special built-in events provided by the bpftrace runtime.
+These are triggered when a specified signal is received. Currently, only SIGUSR1 is supported.
```
viktormalik, line 2856: The probe is triggered only when the **bpftrace process** receives a signal, right? Perhaps it's worth mentioning here to prevent confusion.

src/ast/attachpoint_parser.cpp
```diff
@@ -276,13 +276,22 @@ AttachPointParser::State AttachPointParser::lex_attachpoint(
...
+    if (parts_.size() == 2 && parts_[1] == "*")
+      parts_.pop_back();
+    if (parts_.size() != 1) {
+      return argument_count_error(0);
+    }
+  } else if (ap_->provider == "signal") {
+    if (parts_.size() != 2) {
+      return argument_count_error(1);
+    }
+    // TODO: allow wildcard for listing?
```
viktormalik, line 293: We could even enable wildcard attachments in future.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3155,6 +3155,13 @@ void SemanticAnalyser::visit(AttachPoint &ap)
         has_end_probe_ = true;
       }
     }
+  } else if (ap.provider == "signal") {
+    if (ap.target == "")
+      LOG(ERROR, ap.loc, err_) << ap.provider << " should have a target";
+    if (ap.func != "")
+      LOG(ERROR, ap.loc, err_) << ap.provider << " should only have a target";
```
viktormalik, line 3162: These should be caught by `AttachpointParser`, shouldn't they?

viktormalik, line 3162: Also, could we get unit tests for these errors?

src/bpftrace.cpp
```diff
@@ -962,13 +950,30 @@ int BPFtrace::run(BpfBytecode bytecode)
...
-  // XXX: cast is a workaround for a gcc bug that causes an "invalid conversion"
-  // error here in case the trigger function has the `target` attribute set (see
-  // triggers.h): https://gcc.godbolt.org/z/44b5MjdbT
-  if (run_special_probe("BEGIN_trigger",
-                        bytecode_,
-                        reinterpret_cast<trigger_fn_t>(BEGIN_trigger)))
-    return -1;
+  for (auto &special_probe :
+       std::ranges::reverse_view(resources.special_probes)) {
+    if (special_probe.attach_point == "BEGIN_trigger") {
```
viktormalik, line 955: It looks like it's time to start storing special probes in a map hashed by the probe name. That could remove these loops and simplify the code a lot.

viktormalik (commented):
Nice feature! Just a few comments below.

ajor:
To throw out another idea for probe naming that combines two cases we'd eventually want to support:
```
trigger:signal:SIGUSR1 {}
trigger:keypress:space {}
```

src/ast/passes/semantic_analyser.cpp
amiremohamadi: oh right, checking the number of arguments in AttachpointParser seems enough.

amiremohamadi:
> To throw out another idea for probe naming that combines two cases we'd eventually want to support:
> 
> ```
> trigger:signal:SIGUSR1 {}
> trigger:keypress:space {}
> ```

looks more clear. if you all agree, we can use this naming.

amiremohamadi:
or maybe using `event` instead of `trigger`?

```
event:signal:SIGUSR1 {}
```

viktormalik:
What is the purpose of using the `trigger:` or `event:` prefix? To logically group probe types which fire when the bpftrace process is a target of an event (signal, keypress)? Are we expecting more such events in future?

ajor:
> What is the purpose of using the `trigger:` or `event:` prefix? To logically group probe types which fire when the bpftrace process is a target of an event (signal, keypress)? Are we expecting more such events in future?

The logical grouping was the reason in my mind. I can't think of any classes of event like this right now other than signals and keypresses.

I guess another reason for grouping them behind a single probe type would be that it makes it clear that it's not tracing any signal or keypress systemwide - only ones sent to the bpftrace runtime. Another idea based on this:
```
bpftrace:signal:SIGUSR
bpftrace:keypress:space
```

src/bpftrace.cpp
amiremohamadi: how about using a function instead? if we store special probes in a map, in different parts across the code (E.g [this](https://github.com/bpftrace/bpftrace/blob/57b355c95b5406742edeb4b99330e2253703615b/src/bpfbytecode.cpp#L190)) we need to use templates or convert the std::map to a std::vec.

viktormalik: Sure, it was just an idea. We're talking units of probes here so storing them in a map will not bring any real performance benefits and, as you notice, complicate other code paths. Function sounds like a way to go, the only problem is that sometimes we need to *attach and trigger* a probe (BEGIN, END) and sometimes *just attach* it (signals).

src/bpftrace.cpp
```diff
@@ -119,10 +119,12 @@ int BPFtrace::add_probe(const ast::AttachPoint &ap,
...
 
   // Add the new probe(s) to resources
-  if (ap.provider == "BEGIN" || ap.provider == "END") {
+  if (ap.provider == "BEGIN" || ap.provider == "END" ||
+      ap.provider == "bpftrace") {
     // special probes
     probe.path = "/proc/self/exe";
-    probe.attach_point = ap.provider + "_trigger";
+    auto target = ap.target.empty() ? "" : "_" + ap.target;
+    probe.attach_point = ap.provider + target + "_trigger";
```
viktormalik, line 127: Not valid ATM but once we support multiple signals, we'll need to add the `_<func>` part to the attach point name, right?

src/types.h
```diff
@@ -597,6 +597,7 @@ const std::vector<ProbeItem> PROBE_LIST = {
     .show_in_userspace_list = true },
   { .name = "BEGIN", .aliases = { "BEGIN" }, .type = ProbeType::special },
   { .name = "END", .aliases = { "END" }, .type = ProbeType::special },
+  { .name = "bpftrace", .aliases = { "bpftrace" }, .type = ProbeType::special },
```
viktormalik, line 600: Shall we make a `bt` alias? It'd be unified with the common extension of bpfscript files.

viktormalik (commented):
Nice changes! A few more comments but I think that it's almost ready.

jordalgo:
Nit: The commit message could use some more detail and perhaps an example.

man/adoc/bpftrace.adoc
```diff
@@ -2865,6 +2870,21 @@ END {
...
 ----
 
+[#probes-bpftrace]
+=== bpftrace
+
+.variants
+* `bpftrace:signal:SIGUSR1`
+
+These are special built-in events provided by the bpftrace runtime.
+They are triggered when the bpftrace process receives specific events, such as a `SIGUSR1` signal.
```
fbs, line 2880: does this need more words about how this overrides the built in behaviour (of printing maps) and how the bpf prog trigger is indirct (i.e. it depends on the runtime calling a trigger function). Also, what happens if you define two of these probes?

jordalgo:
I don't mean to hold up this PR but I got confused when I first saw this naming `bpftrace:signal:SIGUSR`. I read the conversation above and understand the reasoning but naming a probe after the name of the tool feels wrong. I'm still thinking about alternatives but what about something like `self:signal:SIGUSR` ?

src/ast/attachpoint_parser.cpp
```diff
@@ -276,13 +276,23 @@ AttachPointParser::State AttachPointParser::lex_attachpoint(
...
+    if (parts_.size() == 2 && parts_[1] == "*")
+      parts_.pop_back();
+    if (parts_.size() != 1) {
+      return argument_count_error(0);
+    }
+  } else if (ap_->provider == "bpftrace") {
+    if (parts_.size() != 3) {
+      return argument_count_error(2);
+    }
+    // TODO: Allow wildcard
```
jordalgo, line 293: Eh, I'm not sure we should add wildcard support - and it definitely doesn't apply to BEGIN/END

viktormalik: True, I'm not sure either. It would make it possible to attach a single probe to all signals received by the bpftrace runtime but I don't know if that's any useful.

Once we add the `bpftrace:keypress:...` events, wildcards could be more useful.

jordalgo:
Should this also deprecate the ability to print maps when issuing a SIGUSR to bpftrace? Seems strange to have both.
https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#map-printing

amiremohamadi:
hmm. while defining a signal probe changes the default behavior, still it's not a breaking change for those who don’t use it. does it really need a migration guide entry?

viktormalik:
> hmm. while defining a signal probe changes the default behavior, still it's not a breaking change for those who don’t use it. does it really need a migration guide entry? @viktormalik @jordalgo

But it is a breaking change for those who do. If people have scripts which run a bpftrace command, periodically send SIGUSR1, and process the results, this change will break them. Which is why I think that it is a breaking change.

viktormalik:
Also cc @andreasgerstmayr who added the original "print maps on SIGUSR1" feature.

src/bpftrace.cpp
```diff
@@ -1237,10 +1261,18 @@ void BPFtrace::poll_output(bool drain)
...
-      print_maps();
+
+      if (sigusr1_attached_probe_) {
+        if (run_special_probe(*sigusr1_attached_probe_,
+                              reinterpret_cast<trigger_fn_t>(signal_trigger))) {
+          LOG(ERROR) << "Failed to run signal probe";
+          return;
+        }
+      } else {
+        print_maps();
```
viktormalik, line 1274: I don't think that we should preserve the default behaviour here. It was just a convenience b/c some people wanted to inspect map contents during bpftrace script runtime but now that we have proper signal probes, they can just define it themselves.

I now realized that I confused you a lot b/c I missed this line in my earlier review and I though that we were dropping this behaviour. Which is why I insisted on the migration guide entry. My apologies about that.

Still, I think that we should drop it.

fbs: Isn't it nice to keep it? For simple scripts it can be a nice bit of utility, you dont have to add the `signal{print(@)}` manually, which makes it a lttle bit simpler, and the code complexity of keeping this seems low.

viktormalik: To me, it's a bit of a surprising behaviour, even though it is documented. Why should the tool print maps upon receiving some random signal? I understand that it was a convenient way to provide a requested feature but now that we have proper signal probes, I'd argue that we should stick to the principle of the least surprise.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3262,15 +3262,14 @@ void SemanticAnalyser::visit(AttachPoint &ap)
...
-          << "fentry/fexit not available for your kernel version.";
+  } else if (ap.provider == "bpftrace") {
+    if (ap.target == "signal") {
+      if (SIGNALS.find(ap.func) == SIGNALS.end())
+        LOG(ERROR, ap.loc, err_) << ap.func << " is not a supported signal";
       return;
     }
-
-    if (ap.func == "")
-      LOG(ERROR, ap.loc, err_) << "fentry should specify a function";
```
danobi, line 3273: Looks like #3524 caused this duplicated code.

Will need a rebase past #3538. Let's do a clean fix to intent is saved into git.

danobi:
Apologies in advance if this sends you on bit of a side quest, but BPF_PROG_TYPE_RAW_TRACEPOINT was added in 4.17 kernel: https://github.com/torvalds/linux/commit/a0fe3e574b50461c4811ce81811f0eaefda62871

bpftrace's policy w.r.t. old kernels is we support up to the oldest LTS kernel still in service: https://www.kernel.org/category/releases.html

The oldest LTS kernel still in service is 4.19, so we are free to assume BPF_PROG_TYPE_RAW_TRACEPOINT is always available.

So let's delete all the trigger probe stuff. It was always a bit of a hack. It should greatly simplify some of the stuff you're doing in this PR. Also means we can delete the feature probe (`has_raw_tp_special()`)

amiremohamadi:
@danobi just to make sure i got your point correctly, now we don't need to attach the special probes to anyting? calling `bpf_prog_test_run_opts` would be enough, right?

src/bpftrace.h
```diff
@@ -190,7 +190,7 @@ class BPFtrace {
   const FuncsModulesMap &get_traceable_funcs() const;
   KConfig kconfig;
   std::vector<std::unique_ptr<AttachedProbe>> attached_probes_;
-  std::unique_ptr<AttachedProbe> sigusr1_attached_probe_;
+  std::optional<int> sigusr1_prog_fd_;
```
amiremohamadi, line 193: using optional instead of initializing the `prog_fd` with a value because `bpf_prog_test_run_opts` will raise an error for errored fds (similiar to BEGIN/END)

danobi:
> @danobi just to make sure i got your point correctly, now we don't need to attach the special probes to anyting? calling `bpf_prog_test_run_opts` would be enough, right?

Yes, that's right. And sorry about late reply - must've missed notification

danobi:
> @viktormalik - Making sure you saw my comment above. What do you think about `self:signal:SIGUSR` instead of `bpftrace:signal:SIGUSR`?

Hmm, I bpftrace prefix is ok. Maybe I write too much c++/rust. In both languages you end up having a top level namespace w/ same name as project. `self` is probably ok too.

danobi:
:P sure. Sounds like `self` is shorter and less ambiguous for future use cases.

`<joke>or we could do "this"</joke>`

amiremohamadi:
thanks folks. I’d like to know @ajor and @viktormalik opinion too before doing any modifications.

viktormalik:
Yeah, `self` sounds good and probably expresses the purpose better (the "target" is the current bpftrace process). No strong opinion on my side.

jordalgo (approved):
LGTM! Thanks for the contribution!

This does make me crave the ability for bpftrace to be more interactive and be able to handle custom commands from stdin e.g. "print_maps", "lost_event_count", "print_map_value(@a, 1)", etc.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3333,6 +3333,13 @@ void SemanticAnalyser::visit(AttachPoint &ap)
         has_end_probe_ = true;
       }
     }
+  } else if (ap.provider == "self") {
+    if (ap.target == "signal") {
+      if (SIGNALS.find(ap.func) == SIGNALS.end())
+        LOG(ERROR, ap.loc, err_) << ap.func << " is not a supported signal";
```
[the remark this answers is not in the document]
amiremohamadi: can you elaborate more? currently It looks like this 
https://github.com/bpftrace/bpftrace/blob/611450e2d362fd29543fafffefd81b05555ae39c/tests/semantic_analyser.cpp#L2197-L2200

amiremohamadi:
> This does make me crave the ability for bpftrace to be more interactive and be able to handle custom commands from stdin e.g. "print_maps", "lost_event_count", "print_map_value(@A, 1)", etc.

if I’m not mistaken, we can also [receive data via signals](https://man7.org/linux/man-pages/man2/sigaction.2.html). while this approach isn’t as convenient as sending data through stdin, it can still be implemented relatively easily in the future

src/ast/passes/semantic_analyser.cpp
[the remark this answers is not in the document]
jordalgo: Don't mind my passing comment. It was just about the variable name "func" not really being a function. Not blocking anything.

jordalgo:
@viktormalik @danobi - One last look before I merge?

src/required_resources.h
```diff
@@ -132,7 +132,7 @@ class RequiredResources {
   // Probe metadata that codegen creates. Ideally ResourceAnalyser pass should
   // be collecting this, but it's complex to move the logic.
   std::vector<Probe> probes;
-  std::vector<Probe> special_probes;
+  std::map<std::string, Probe> special_probes;
```
viktormalik, line 135: Prefer `std::unordered_map` for string keys.

src/bpfbytecode.cpp
```diff
@@ -186,8 +186,11 @@ void BpfBytecode::load_progs(const RequiredResources &resources,
     bpf_program__set_log_buf(prog.bpf_prog(), log_buf.data(), log_buf.size());
   }
 
+  std::vector<Probe> special_probes;
+  for (auto probe : resources.special_probes)
+    special_probes.push_back(probe.second);
```
viktormalik, line 191: Not very nice that we have to copy here but I don't have a better solution at the moment.

danobi:
Did a quick skim - didn't spot any major issues

simpleton:
> BPF_PROG_TYPE_RAW_TRACEPOINT was added in 4.17 kernel

this's not sufficient, due to the test_run on raw_tracepoint is not available util 5.10 kernel. https://github.com/torvalds/linux/commit/1b4d60ec162f82ea29a2e7a907b5c6cc9f926321

jordalgo:
@simpleton Sadly, our docs are a bit out of date. I think we're right now looking at a min version of 5.15 as our support backstop - though we're still investigating what's broken on that.

PR #3523, opened 2024-10-18
Optimise `len()` by using `bpf_map_sum_elem_count` kfunc (introduce kfuncs)

### LINKED ISSUE #3511: Optimise `len()` by using `bpf_map_sum_elem_count` kfunc
The `len()` function currently works by iterating over all elements in a map. The kernel provides a kfunc `bpf_map_sum_elem_count` which does the same, more efficiently.

It looks like this kfunc has a simple interface, so it might be a good candidate for our first kfunc. @viktormalik

viktormalik:
Optimise `len()` by calling the `bpf_map_sum_elem_count` kernel function if it is available. The function just reads and sums per-cpu counters so it's much faster than iterating through all map elements. We still keep that code path, though, for older kernels that do not have the kfunc (it was added in https://github.com/torvalds/linux/commit/803370d3d375).

One problem is that not all map types support per-cpu counters. For those that don't, `bpf_map_sum_elem_count` returns 0. From map types used in bpftrace, this applies to (per-cpu) array maps which are typically used for keyless maps. Such maps always have a single element so we can directly return 1 upon `len()` call in most such cases.

Closes #3511.

This is the first instance of calling kernel functions (so-called "kfuncs") from bpftrace bytecode. This required some tweaks, especially to improve how we generate debug info for function prototypes. See commit message of the first patch for details.

The support for kfuncs is rather straightforward, the kfunc type is retrieved from BTF and a weak declaration of the prototype is emitted to the `.ksyms` section. Then, libbpf takes care of the rest.

This also introduces a way to write codegen tests which need BTF which also required some tweaks (patches 2 and 3).

##### Checklist


src/ast/dibuilderbpf.cpp
```diff
@@ -39,12 +42,29 @@ void DIBuilderBPF::createFunctionDebugInfo(Function &func)
                                          DINode::FlagPrototyped,
                                          flags);
 
-  createParameterVariable(
-      subprog, "ctx", 1, file, 0, static_cast<DIType *>(types[1]), true);
+  for (size_t i = 0; i < args.fields.size(); i++) {
```
jordalgo, line 48: nit: maybe this reads better:
`for (size_t i = 1; i < args.fields.size() + 1; i++) {`

src/bpffeature.cpp
```diff
@@ -739,4 +739,18 @@ bool BPFfeature::has_kernel_dwarf()
   return dwarf->has_debug_info();
 }
 
+bool BPFfeature::has_kernel_func(const std::string& name)
+{
+  if (!has_btf())
```
jordalgo, line 744: Is it not possible for us to even attempt to call a kfunc if we dont' have BTF?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3742,6 +3757,22 @@ libbpf::bpf_map_type CodegenLLVM::get_map_type(const SizedType &val_type,
...
+bool CodegenLLVM::is_array_map(const SizedType &val_type,
+                               const SizedType &key_type)
+{
+  auto map_type = get_map_type(val_type, key_type);
+  return map_type == libbpf::BPF_MAP_TYPE_ARRAY ||
+         map_type == libbpf::BPF_MAP_TYPE_PERCPU_ARRAY;
+}
+
+// Check if we can special-case the map to have a single element. This is done
+// for keyless maps (e.g. @x or @++) of BPF_MAP_TYPE_(PERCPU_)ARRAY type.
```
jordalgo, line 3769: nit: (e.g. `@x = 1;` or `@++;`)

jordalgo (commented):
Overall this looks great! Very nice, detailed commit messages. Just a few small nits.

src/struct.h
```diff
@@ -91,9 +91,9 @@ struct Struct {
   const Field &GetField(const std::string &name) const;
   void AddField(const std::string &field_name,
```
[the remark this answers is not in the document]
viktormalik: I'd prefer keeping the changelog of this PR minimal. Also don't see much benefit from overloading here.

src/bpffeature.cpp
viktormalik: No, b/c we don't know the signature and therefore are not able to construct the call. In addition, kfuncs are "called" by their BTF ID so the kernel where this is run has to have BTF.

src/bpffeature.h
```diff
@@ -146,6 +148,8 @@ class BPFfeature {
   std::optional<bool> has_btf_func_global_;
   std::optional<bool> has_kernel_dwarf_;
 
+  std::unordered_map<std::string, bool> has_kernel_func_;
```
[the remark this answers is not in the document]
viktormalik: Good point. What about `available_kernel_funcs_`? That sounds quite self-describing to me.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1283,16 +1283,31 @@ void CodegenLLVM::visit(Call &call)
     auto &arg = *call.vargs.at(0);
     auto &map = static_cast<Map &>(arg);
 
-    AllocaInst *len = b_.CreateAllocaBPF(b_.getInt64Ty(), "len");
-    b_.CreateStore(b_.getInt64(0), len);
+    // Some map types (BPF_MAP_TYPE_(PERCPU_)ARRAY from those that we use) do
```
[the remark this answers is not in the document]
viktormalik: Nice, thanks.

src/bpffeature.h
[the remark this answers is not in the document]
jordalgo: Works for me.

src/ast/dibuilderbpf.cpp
jordalgo: Ah yeah, you're right. Keep as is.

src/struct.h
```diff
@@ -12,6 +12,8 @@
 
 namespace bpftrace {
 
+static constexpr std::string retval_field_name = "$retval";
```
jordalgo, line 15: (food for though) We have a mix of constexpr names that are all caps and ones that are like this. What do you think about making file level consts all caps as a style rule?

viktormalik: Yeah, makes sense, let me change this one.

src/ast/dibuilderbpf.cpp
```diff
@@ -14,12 +14,15 @@ DIBuilderBPF::DIBuilderBPF(Module &module) : DIBuilder(module)
...
+void DIBuilderBPF::createFunctionDebugInfo(Function &func,
+                                           const SizedType &ret_type,
+                                           const Struct &args)
 {
-  // BPF probe function has:
-  // - int return type
-  // - single parameter (ctx) of a pointer type
-  SmallVector<Metadata *, 2> types = { getInt64Ty(), getInt8PtrTy() };
+  // Return type should be at index 0
+  SmallVector<Metadata *> types;
```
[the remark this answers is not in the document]
jordalgo: "I won't block but I think you forgot about me" - tiny optimization

jordalgo (approved):
Provided the re-run job passes. This looks good to me! Hooray for kfuncs!

@viktormalik - Can we get this one merged first? https://github.com/bpftrace/bpftrace/pull/3524

viktormalik:
> @viktormalik - Can we get this one merged first? #3524

Sure, let's get it in :rocket:

src/ast/dibuilderbpf.cpp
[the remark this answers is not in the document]
viktormalik: I didn't, see the `reserve` call one line below. Or did I miss something?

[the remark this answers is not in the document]
jordalgo: Doh. I'm blind 🙄

src/ast/passes/codegen_llvm.h
```diff
@@ -257,6 +257,12 @@ class CodegenLLVM : public Visitor {
   void maybeAllocVariable(const std::string &var_ident,
                           const SizedType &var_type);
 
+  Function *DeclareKernelFunc(const std::string &func_name);
+
+  CallInst *CreateKernelFuncCall(const std::string &func_name,
```
danobi, line 265: Can we do enums for kfuncs instead? Strings are easy to typo. Furthermore, kfuncs are unstable API. So I can see us ending up in a situation where we need `v1`, `v2`, ... `vN` variants for the same kfunc. Enums would help make that more manageable as well.

src/bpffeature.h
```diff
@@ -104,6 +104,8 @@ class BPFfeature {
   bool has_iter(std::string name);
   bool has_kernel_dwarf();
 
+  bool has_kernel_func(const std::string& name);
```
danobi, line 107: Same comment about enum

src/bpffeature.h
```diff
@@ -146,6 +148,8 @@ class BPFfeature {
   std::optional<bool> has_btf_func_global_;
   std::optional<bool> has_kernel_dwarf_;
 
+  std::unordered_map<std::string, bool> available_kernel_funcs_;
```
danobi, line 151: Same

tests/mocks.h
```diff
@@ -116,6 +116,11 @@ class MockBPFfeature : public BPFfeature {
     has_loop_ = std::make_optional<bool>(has);
   }
 
+  void mock_missing_kernel_func(const std::string &name)
```
danobi, line 119: Same

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1291,11 +1291,26 @@ void CodegenLLVM::visit(Call &call)
...
+    // Some map types used in bpftrace (BPF_MAP_TYPE_(PERCPU_)ARRAY) do not
+    // implement per-cpu counters and bpf_map_sum_elem_count would always return
+    // 0 for them. In our case, those maps typically have a single element so we
+    // can return 1 straight away.
+    // For the rest, use bpf_map_sum_elem_count if available and map supports
+    // it, otherwise fall back to bpf_for_each_map_elem with a custom callback.
+    if (map_has_single_elem(map.type, map.key_type)) {
+      expr_ = b_.getInt64(1);
+    } else if (bpftrace_.feature_->has_kernel_func("bpf_map_sum_elem_count") &&
+               !is_array_map(map.type, map.key_type)) {
```
danobi, line 1303: Don't we statically know size of array map? Should be possible to just hard code the result into bytecode, right? So another branch like single elem case above

src/ast/irbuilderbpf.h
```diff
@@ -41,7 +41,7 @@ class IRBuilderBPF : public IRBuilder<> {
   AllocaInst *CreateAllocaBPF(int bytes, const std::string &name = "");
   void CreateMemsetBPF(Value *ptr, Value *val, uint32_t size);
   void CreateMemcpyBPF(Value *dst, Value *src, uint32_t size);
-  llvm::Type *GetType(const SizedType &stype);
```
danobi, line 44: Might be good to leave some of the commit message about meaning of `emit_codegen_types` as a comment in here. Not obvious to me from code alone what the param does.

And I'd suggest rewording:

> we want to use i64 for pointers and byte arrays for structs as that's how we actually store them,

to something about how it's what libbpf expects. That's the real reason, right?

danobi (commented):
very exciting change!

src/ast/passes/codegen_llvm.cpp
viktormalik: We do but this case checks if the map is **not** an array.

src/ast/irbuilderbpf.h
viktormalik: > Might be good to leave some of the commit message about meaning of `emit_codegen_types` as a comment in here. Not obvious to me from code alone what the param does.

Good idea, will do.


> And I'd suggest rewording:
> 
> 
> 
> > we want to use i64 for pointers and byte arrays for structs as that's how we actually store them,
> 
> 
> 
> to something about how it's what libbpf expects. That's the real reason, right?
> 

I don't think so. IIUC, we store pointers as ints and structs as byte arrays b/c it's easier to store them in maps that way (maybe BPF maps didn't even allow storing raw pointers in maps back in the time when bpftrace was created).

OTOH, libbpf requires kfunc declarations to have types matching those in the kernel which is why I had to introduce the `emit_codegen_types` param. I'll try to reword it in the comment and commit message to make it clearer.

src/ast/passes/codegen_llvm.h
viktormalik: Good idea, I'll give it a shot.

src/ast/passes/codegen_llvm.cpp
danobi: Right, cuz array map is statically sized. I was suggesting we could have above a:

```
else if (is_array_map(...) {
  expr_ = b_.getInt64(map.type.GetNumElements());
}
```

then the `!is_array_map(..)` check in the has_kernel_func() branch could be dropped as well.

src/ast/irbuilderbpf.h
viktormalik: Updated both the commit message and the relevant function comments.

src/ast/passes/codegen_llvm.h
viktormalik: Done, ready for another review.

danobi (approved):
small nits, but lgtm!

sorry about delay, and let's not round trip it again. feel free to follow up w/ another PR. the kfunc decl tag thing is probably ok to omit if it's gonna be complex to mock it up in data_source.c

danobi:
Nvm on the string_view thing. I see it just complicates things more. Here's the followup for the enum class fix: https://github.com/bpftrace/bpftrace/pull/3556

PR #3524, opened 2024-10-18
Deprecate k(ret)func probes in favor of fentry/fexit

viktormalik:
Originally, when bpftrace introduced support for BPF trampolines (called `fentry`/`fexit` in the kernel), these were called `kfunc`/`kretfunc` to make a connection to `kprobe`/`kretprobe`. Unfortunately later, kernel introduced [BPF Kernel Functions](https://docs.kernel.org/bpf/kfuncs.html) which started to be abbreviated as "kfuncs".

This creates a lot of confusion as the same name "kfunc" is used for two entirely different concepts in the BPF world. This will be even more evident when bpftrace starts calling kernel kfuncs (coming in #3523).

Since kernel is not going to change their naming, let us align on bpftrace side. We already have `fentry`/`fexit` as aliases for `k(ret)func` so deprecate the usage of `kfunc`/`kretfunc` by removing it from all places in the project (documentation, source code, tests). Since there are many places where the probe name faces users (documentation, error messages, etc.), the most reasonable thing to do is to just replace it everywhere.

The old probe names `kfunc`/`kretfunc` are kept for backwards compatibility but they should not be mentioned anywhere (except one place in the docs) anymore.

The change is mostly mechanical, with several manual adjustments where necessary. I check every replacement manually but another thorough check will do no harm.

cc @olsajiri 

##### Checklist


tests/codegen/llvm/kretfunc_dereference.ll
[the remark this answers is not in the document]
viktormalik: Yes, I just forgot to delete it.

viktormalik:
> Nice change!
> 
> There's just a test failure in `call.path_in_unsupported_fentry` and I think an extra test file needs deleting/renaming.

Good catch! I wonder why the GHA run on my fork didn't catch that :thinking:

man/adoc/bpftrace.adoc
```diff
@@ -3013,50 +3013,52 @@ iter:task_file:/sys/fs/bpf/files {
...
-[#probes-kfunc]
-=== kfunc and kretfunc
+[#probes-fentry]
+=== fentry and fexit
 
 .variants
-* `kfunc[:module]:fn`
 * `fentry[:module]:fn`
-* `kretfunc[:module]:fn`
+* `kfunc[:module]:fn`
```
jordalgo, line 3021: Why do we want to leave these references in the adoc? I would probably just put a little note that these are deprecated but not offer them as valid variants.

jordalgo (approved):
Yes,  yes, yes. Great change! One small comment on the adoc but I won't let it be a blocker.

man/adoc/bpftrace.adoc
viktormalik: That note is already there few lines below, so I guess that we can just drop this.

danobi:
Yeah, this was probably overdue

Issue #3308, opened 2024-07-11
Verifier Error / Incorrect comparisons against result of len() function


ajor:
This script should print "true":
```
BEGIN
{
  @[1] = 1;
  @[2] = 2;

  $count = len(@);
  // Uncomment this line and it works
  //$count = 2;

  printf("count: %d\n", $count);

  if ($count > 1) {
    print("true");
  } else {
    print("false");
  }

  printf("count: %d\n", $count);

  clear(@);
  exit();
}
```

Output:
```
Attaching 1 probe...
count: 2
false
count: 2
```

We can see that `$count` holds the value 2 (as expected), but this value does not evaluate as greater than 1 for some reason.
If we use the literal `2` instead of the result of `len(@)`, then we get the expected result.

viktormalik:
Hm, my output on latest master (unless I did some mistake):
```
# src/bpftrace test.bt
[...]
BPF program is too large. Processed 1000001 insn
processed 1000001 insns (limit 1000000) max_states_per_insn 65 total_states 75003 peak_states 25025 mark_read 5


ERROR: Loading BPF object(s) failed.
```

ajor:
> Hm, my output on latest master (unless I did some mistake):
> 
> ```
> # src/bpftrace test.bt
> [...]
> BPF program is too large. Processed 1000001 insn
> processed 1000001 insns (limit 1000000) max_states_per_insn 65 total_states 75003 peak_states 25025 mark_read 5
> 
> 
> ERROR: Loading BPF object(s) failed.
> ```

That's odd - my bpftrace compiles the script down to 168 instructions and I don't think there are any loops in there (except the for-each-map-elem).

Things that could be relevant are that I'm on LLVM 17 and my kernel is 5.19 but with lots of newer BPF features backported.

Here's my full `--info`:
```
System
  OS: Linux 5.19.0-0_fbk11_hardened_11538_g61e82533119f #1 SMP Wed Oct 4 09:20:00 PDT 2023
  Arch: x86_64
Build
  version: v0.21.0-33-g32ca
  LLVM: 17.0.6
  unsafe probe: no
  bfd: yes
  liblldb (DWARF support): yes
  libsystemd (systemd notify support): no
Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
[...]

PR #3533, opened 2024-10-20
fix verifier Error / Incorrect comparisons against result of len() function

mannkafai:
Use the return value of `bpf_for_each_map_elem` as `len()` function result 
since it means the number of traversed map elements for success. 
Modify the `map_len_cb` to a dummy function and only return 0.

That will close [#3308](https://github.com/bpftrace/bpftrace/issues/3308)

### DIFF
--- a/src/ast/irbuilderbpf.h
+++ b/src/ast/irbuilderbpf.h
@@ -74,12 +74,11 @@ public:
                            Map &map,
                            Value *key,
                            const location &loc);
-  void CreateForEachMapElem(Value *ctx,
-                            Map &map,
-                            Value *callback,
-                            Value *callback_ctx,
-                            Value *return_ctx,
-                            const location &loc);
+  Value *CreateForEachMapElem(Value *ctx,
+                              Map &map,
+                              Value *callback,
+                              Value *callback_ctx,
+                              const location &loc);
   void CreateProbeRead(Value *ctx,
                        Value *dst,
                        llvm::Value *size,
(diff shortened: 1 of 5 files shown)


src/ast/irbuilderbpf.h
```diff
@@ -78,6 +78,7 @@ class IRBuilderBPF : public IRBuilder<> {
                             Map &map,
                             Value *callback,
                             Value *callback_ctx,
+                            Value *return_ctx,
```
danobi, line 81: Rather than an out-param, how about just making `CreateForEachMapElem()` return a `Value *`? `CallInst *`  is a valid `Value *`.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1297,7 +1297,7 @@ void CodegenLLVM::visit(Call &call)
     if (!map_len_func_)
       map_len_func_ = createMapLenCallback();
 
-    b_.CreateForEachMapElem(ctx_, map, map_len_func_, len, call.loc);
+    b_.CreateForEachMapElem(ctx_, map, map_len_func_, len, len, call.loc);
```
danobi, line 1300: I think the `len` allocated on stack is no longer necessary. Can you try using a null ptr instead?

danobi:
Just confirming: does this fix the linked issue? If so, please add a runtime test to prevent future regression

src/ast/passes/codegen_llvm.cpp
mannkafai: Yes, you are right.  I'll revise it.

jordalgo (approved):
Nice! LGTM

danobi (approved):

danobi:
For next time, please add a CHANGELOG.md entry. To save a round trip for this PR, I'll push one up.

But thanks for the contribution!

Issue #3488, opened 2024-10-04
RFC: bpftrace Types


jordalgo:
This is mostly documenting decisions from a discussion at the most recent office hours but please feel free to comment, offer up suggestions, etc.

Current internal types will remain named as they are (following this [PR](https://github.com/bpftrace/bpftrace/pull/3478)). We will try our best not add any additional internal types (as we hope to rely on user-defined types or those defined in a standard library). However, if an additional type needs to be added, we will add the "_t" suffix if it conflicts with an existing internal function. If this type happens to conflict with a script that has already used this type name, the script will probably break but we will try to direct the user to a resolution in both the error message and in bpftrace documentation/release notes.

In cases where there is a type conflict within a script e.g. if the user, library, or kernel header defines the same type name as an internal type then the internal type will be chosen (and most likely a type mismatch error in one of the AST passes will occur). Users can resolve this issue by adding a namespace (the exact namespace is TBD) to the beginning of their type e.g. `mylibrary::ksym_t`.

Additional implementation detail: there are plans to remove hardcoded types from the lexer/parser and handle these in an AST pass.

batu3H:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

jordalgo:
Note to self: Improve type hinting when we pull hard-coded types out of the lexer: https://github.com/bpftrace/bpftrace/pull/3534

jordalgo:
Closing this as we're re-thinking our internal types right now.

PR #3534, opened 2024-10-21
Hint the correct integer type for casts

ttreyer:
The naming scheme for integers in `bpftrace` is different from `C`: `int` becomes `int32`, `unsigned char` becomes `uint8`. It might cause confusion for first time users to not have these basic types. They would not be able to pretty print the value of a `counter()`, if they can't cast it to an integer.

This PR adds a hint for type casts, suggesting the name of the `bpftrace` integer type matching the `C` type.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2634,6 +2634,16 @@ void SemanticAnalyser::visit(FieldAccess &acc)
   }
 }
 
+// We can't hint for unsigned types. It is a syntax error,
+// because the word "unsigned" is not allowed in a type name.
+static std::unordered_map<std::string_view, std::string_view>
+    KNOWN_TYPE_ALIASES{
+      { "char", "int8" },   /* { "unsigned char", "uint8" }, */
+      { "short", "int16" }, /* { "unsigned short", "uint16" }, */
+      { "int", "int32" },   /* { "unsigned int", "uint32" }, */
+      { "long", "int64" },  /* { "unsigned long", "uint64" }, */
+    };
+
 void SemanticAnalyser::visit(Cast &cast)
 {
   Visit(cast.expr);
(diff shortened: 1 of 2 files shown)


ttreyer:
Should this hint be for _type cast_ only? In `SemanticAnalyser::visit(Cast &cast)`?
Or should it be also for resolving the type in `SemanticAnalyser::resolve_struct_type(SizedType &type, const location &loc)`?

jordalgo:
> Should this hint be for type cast only? In SemanticAnalyser::visit(Cast &cast)?
Or should it be also for resolving any type? In SemanticAnalyser::resolve_struct_type(SizedType &type, const location &loc)?

Good question. IMHO it should be for resolving any type. This would include our new variable declaration syntax e.g. hint on `let $x: int;`

ttreyer:
> Sorry if I'm missing something, but how do integer type hints apply to `resolve_struct_type()`? `resolve_struct_type()` only errors if `inner_type->IsRecordTy()`, which integers are not.

My bad, its error message also appear alongside the cast error and I thought `resolve_struct_type()` was called on all types.

> Also please add some semantic analysis tests. Would be good to check that output is formatted how we expect (and so it doesn't regress)

Added 🙂 

> Good question. IMHO it should be for resolving any type. This would include our new variable declaration syntax e.g. hint on `let $x: int`;

The problem is that the `type` rule in the syntax is strict and won't even accept an invalid type! For instance, trying your `let $x: int` gives me the following parser error:
```
stdin:1:17-21: ERROR: syntax error, unexpected identifier, expecting struct or integer type or builtin type or sized type
BEGIN { let $x: int; }
                ~~~~
```
So hinting on `let` will require adding custom error message to our parser, or changing the `type` rule to accept identifier and catching the type issue in `semantic_analyser`.

jordalgo:
> Sorry if I'm missing something, but how do integer type hints apply to resolve_struct_type()? resolve_struct_type() only errors if inner_type->IsRecordTy(), which integers are not.

I think @ttreyer demonstrated how this is happening in the newly added semantic tests. And I believe it's because of this line in the parser: https://github.com/bpftrace/bpftrace/blob/master/src/parser.yy#L619

danobi (approved):
lgtm! straightforward improvement.

didn't get a chance to think on what @jordalgo is suggesting, though. so pls get his ack

jordalgo (approved):
This is an improvement so I'm fine to merge. That being said, I feel like the hint should go into the first error message and not the second.

```
test_error("BEGIN { (int)cpu }", R"(
stdin:1:9-14: ERROR: Cannot resolve unknown type "int"
BEGIN { (int)cpu }
        ~~~~~
stdin:1:9-14: ERROR: Cannot cast to "int", did you mean "int32"?
BEGIN { (int)cpu }
        ~~~~~
)");
```

ttreyer:
> This is an improvement so I'm fine to merge. That being said, I feel like the hint should go into the first error message and not the second.

I agree, it would be easier to understand.
Let's make an issue about types being in the lexer (or update #3488), make a note to improve this PR hints and add hint for typed `let`.

PR #3536, opened 2024-10-22
Fix map type back propagation

jordalgo:
Issue: https://github.com/bpftrace/bpftrace/issues/3448

Both of these should be valid:
```
BEGIN
{
  @stacks[@integer] = kstack;
  @integer = 1;
}
```
And
```
BEGIN
{
  @map1[@map2] = 1;
  @map2 = 1;

  for ($kv : @map1) {
  }
}
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,8 @@ and this project adheres to
   - [#3517](https://github.com/bpftrace/bpftrace/pull/3517)
 - Fix verifier error when comparing result of len()
   - [#3308](https://github.com/bpftrace/bpftrace/issues/3308)
+- Fix type back propagation for map keys
+  - [#3536](https://github.com/bpftrace/bpftrace/pull/3536)
 #### Security
 #### Docs
 - Remove mention of unsupported character literals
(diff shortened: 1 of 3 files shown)


danobi (approved):
seems reasonable to me.

tbh the is_final_pass() stuff and the multiple passes is insanely confusing for me to reason about. I think long term we need to start thinking about how to extract and contain type propagation logic into its own pass. ie https://github.com/bpftrace/bpftrace/issues/2979

we probably just need to bite the bullet and allocate time to take care of this tech debt. maybe an 2025H1 roadmap item

jordalgo:
> tbh the is_final_pass() stuff and the multiple passes is insanely confusing for me to reason about. I think long term we need to start thinking about how to extract and contain type propagation logic into its own pass. ie https://github.com/bpftrace/bpftrace/issues/2979
> 
> we probably just need to bite the bullet and allocate time to take care of this tech debt. maybe an 2025H1 roadmap item

Yes, completely agree and I'm happy to take this on (unless someone else really wants to).

danobi:
Please do! I expect the work to probably look similar to big strings in that it’ll have a lot of unanticipated complexity. So we should keep that in mind when budgeting time

PR #3539, opened 2024-10-25
Symbolize enums in maps

danobi:
This PR teaches bpftrace to symbolize enums in map keys. I didn't add map value symbolization as I haven't seen any requests for it. It's also a little weird in that map values can have a divisor (see `div` arg in `print()`). Not a huge deal - we can special case it all. But I figure it wouldn't hurt to add it as requested in the future. Map keys are much more useful as people tend to aggregate based on enums fairly often.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -3731,7 +3731,8 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
     // Store all integer values as 64-bit in map keys, so that there will
     // be space for any integer in the map key later
     // This should have a better solution.
-    new_key_type = CreateInt64();
+    new_key_type.SetSign(true);
+    new_key_type.SetIntBitWidth(64);
   }
 
   validate_map_key(new_key_type, loc);
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -341,33 +341,68 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
       return std::to_string(read_data<uint64_t>(value.data()) / div);
     }
     case Type::integer: {
-      auto sign = type.IsSigned();
-      switch (type.GetIntBitWidth()) {
-          // clang-format off
-        case 64:
-          if (sign)
-            return std::to_string(reduce_value<int64_t>(value, nvalues) / static_cast<int64_t>(div));
-          return std::to_string(reduce_value<uint64_t>(value, nvalues) / div);
-        case 32:
-          if (sign)
-            return std::to_string(
-                reduce_value<int32_t>(value, nvalues) / static_cast<int32_t>(div));
-          return std::to_string(reduce_value<uint32_t>(value, nvalues) / div);
-        case 16:
-          if (sign)
-            return std::to_string(
-                reduce_value<int16_t>(value, nvalues) / static_cast<int16_t>(div));
-          return std::to_string(reduce_value<uint16_t>(value, nvalues) / div);
-        case 8:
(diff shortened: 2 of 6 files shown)


danobi:
So this PR doesn't work for enums inside tuples. Lemme think on it a bit. Should probably  handle that case.

danobi:
Alright I added support for tuples too

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3731,7 +3731,8 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
     // Store all integer values as 64-bit in map keys, so that there will
     // be space for any integer in the map key later
     // This should have a better solution.
-    new_key_type = CreateInt64();
+    new_key_type.SetSign(true);
```
jordalgo, line 3734: Is this because the underlying int type might be an enum/have a name?

src/output.cpp
```diff
@@ -341,33 +341,76 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
...
-            return std::to_string(
-                reduce_value<int8_t>(value, nvalues) / static_cast<int8_t>(div));
-          return std::to_string(reduce_value<uint8_t>(value, nvalues) / div);
-          // clang-format on
-        default:
-          LOG(BUG) << "value_to_str: Invalid int bitwidth: "
-                   << type.GetIntBitWidth() << "provided";
-          return {};
+      // Note we make the a conscious choice to only symbolize enums in
+      // map keys. The biggest concern is that it makes sense to do arith
```
jordalgo, line 345: Did a 14 year old on twitter write this?
```suggestion
      // map keys. The biggest concern is that it makes sense to do arithmetic
```

src/output.cpp
```diff
@@ -341,33 +341,76 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
...
+      // on enums sometimes. So if user does `SOME_ENUM + 1`, we could
```
jordalgo, line 346: Hmm, I'm a little skeptical of this happening. Would be curious to see a real-world example but I won't block on it.

src/ast/passes/semantic_analyser.cpp
danobi: Yep

src/output.cpp
danobi: Thinking again I don't feel strongly about it either way. Hard to think of a case where you'd keep an enum in value of map, though. It's like a ~2 line change either way.

src/output.cpp
danobi: :D

src/output.cpp
danobi: One edge case: when we enable configurable map types and the user chooses a percpu map, then symbolization makes no sense. If each CPU had a different enum val, what do we do?

jordalgo: Exposing per-cpu maps to users I think poses some unique challenges we'll have to discuss.

We really don't have a good solution for how we would print any value that's kept in a percpu map that's not intended to be summed in some way (e.g. count, sum, avg). Maybe we just need to provide some utility that allows users to either print maps with the cpuid paired with the value (like a tuple) or to be able to iterate over cpus and print map values specific to that cpu.

danobi: Good point. Percpu presents challenge for aggregation. So you're right it doesn't intersect with this PR. Let's symbolize map values, then. And continue discussion on percpu on the map type declarations issue.

jordalgo (approved):

PR #3542, opened 2024-10-26
Add checks to prevent assignment of some map types

jordalgo:
These are special map aggregation types that cannot be assigned to scratch variables or from one map
to another e.g. these are both invalid:
`@a = hist(10); let $b = @a;`
`@a = hist(10); @b = @a;`

https://github.com/bpftrace/bpftrace/issues/3540

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -125,6 +125,18 @@ static bool IsValidVarDeclType(const SizedType &ty)
   return false; // unreachable
 }
 
+// These are special map aggregation types that cannot be assigned
+// to scratch variables or from one map to another e.g. these are both invalid:
+// `@a = hist(10); let $b = @a;`
+// `@a = hist(10); @b = @a;`
+static bool IsValidAssignment(const SizedType &ty, bool is_map)
+{
+  if (is_map && (ty.IsHistTy() || ty.IsLhistTy() || ty.IsStatsTy())) {
+    return false;
+  }
+  return true;
+}
+
 void SemanticAnalyser::visit(Integer &integer)
 {
   if (integer.is_negative) {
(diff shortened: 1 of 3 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -125,6 +125,18 @@ static bool IsValidVarDeclType(const SizedType &ty)
   return false; // unreachable
 }
 
+// These are special map aggregation types that cannot be assigned
+// to scratch variables or from one map to another e.g. these are both invalid:
+// `@a = hist(10); let $b = @a;`
+// `@a = hist(10); @b = @a;`
+static bool IsValidAssignment(const SizedType &ty, bool is_map)
+{
+  if (is_map && (ty.IsHistTy() || ty.IsLhistTy() || ty.IsStatsTy())) {
```
fbs, line 134: Should this be inverted? So that new types are not automatically assignable by default?

Or maybe adding an `IsAssignable` method to on the `type` itself?

jordalgo: > Or maybe adding an IsAssignable method to on the type itself?

I like this and will update. But also think that by default types should be assignable. I don't see us really adding more map aggregation types in the future (seems like something our future standard library should provide).

jordalgo: Oh actually, I realized that we need the `is_map` part which is part of the expression so we can't do this on the type itself.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2731,10 +2743,30 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
...
   Visit(assignment.expr);
 
+  auto type = assignment.expr->type;
+
+  if (!IsValidAssignment(type, assignment.expr->is_map)) {
+    auto hint = "";
+    if (type.IsHistTy()) {
+      hint = "hist(retval)";
+    } else if (type.IsLhistTy()) {
+      hint = "lhist(rand %10, 0, 10, 1)";
```
danobi, line 2753: `rand %10`?

src/ast/passes/semantic_analyser.cpp
danobi: Very interesting that avg(), count(), sum(), etc. all work. Intentional or just lucky?

danobi (commented):
lgtm module nit

src/ast/passes/semantic_analyser.cpp
jordalgo: It's the same example for lhist that we have in the adoc

jordalgo:
> Very interesting that avg(), count(), sum(), etc. all work. Intentional or just lucky?

Intentional. Min and max also work because we generate single values for all of these by looping over the per-cpu map e.g. https://github.com/bpftrace/bpftrace/pull/3268

src/ast/passes/semantic_analyser.cpp
jordalgo: Intentional. Min and max also work because we generate single values for all of these by looping over the per-cpu map e.g. https://github.com/bpftrace/bpftrace/pull/3268

danobi (approved):

PR #3544, opened 2024-10-27
Use global scratch buffers for variables

ryantimwilson:
This PR migrates variables to use global scratch variables if size > on_stack_limit. This will allow users to store large strings/tuples/objects in local variables.

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -183,6 +183,15 @@ AllocaInst *IRBuilderBPF::CreateAllocaBPF(const SizedType &stype,
   return CreateAllocaBPF(ty, name);
 }
 
+void IRBuilderBPF::CreateAllocationInit(const SizedType &stype, Value *alloc)
+{
+  if (needMemcpy(stype)) {
+    CreateMemsetBPF(alloc, getInt8(0), stype.GetSize());
+  } else {
+    CreateStore(ConstantInt::get(GetType(stype), 0), alloc);
+  }
+}
+
 AllocaInst *IRBuilderBPF::CreateAllocaBPFInit(const SizedType &stype,
                                               const std::string &name)
 {
@@ -194,11 +203,7 @@ AllocaInst *IRBuilderBPF::CreateAllocaBPFInit(const SizedType &stype,
     llvm::Type *ty = GetType(stype);
     alloca = CreateAlloca(ty, nullptr, name);
     CreateLifetimeStart(alloca);
-    if (needMemcpy(stype)) {
-      CreateMemsetBPF(alloca, getInt8(0), stype.GetSize());
-    } else {
-      CreateStore(ConstantInt::get(ty, 0), alloca);
-    }
+    CreateAllocationInit(stype, alloca);
   });
   return alloca;
 }
@@ -551,6 +556,28 @@ Value *IRBuilderBPF::CreateWriteMapValueAllocation(const SizedType &value_type,
       GetType(value_type),
       name,
       loc);
+}
+
+Value *IRBuilderBPF::CreateVariableAllocationInit(const SizedType &value_type,
+                                                  const std::string &name,
+                                                  const location &loc)
+{
+  /* Hoist variable declaration and initialization to entry point of
+   * probe/subprogram. While we technically do not need this as variables
+   * are properly scoped, it eases debugging and is consistent with previous
+   * stack-only variable implementation. */
+  Value *alloc;
+  hoist([this, &value_type, &name, &loc, &alloc] {
+    alloc = createAllocation(bpftrace::globalvars::GlobalVar::VARIABLE_BUFFER,
+                             GetType(value_type),
+                             name,
+                             loc,
+                             [](AsyncIds &async_ids) {
+                               return async_ids.variable();
+                             });
+    CreateAllocationInit(value_type, alloc);
+  });
+  return alloc;
 }
 
 Value *IRBuilderBPF::createAllocation(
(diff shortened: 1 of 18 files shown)


src/ast/passes/resource_analyser.cpp
```diff
@@ -387,6 +396,31 @@ void ResourceAnalyser::visit(Ternary &ternary)
   }
 }
 
+void ResourceAnalyser::update_variable_info(Variable &var)
+{
+  const auto on_stack_limit = bpftrace_.config_.get(
+      ConfigKeyInt::on_stack_limit);
+  if (var.type.GetSize() > on_stack_limit) {
```
jordalgo, line 403: This similar logic now seems to be in multiple places in this file. Should we have a helper like `exceedStackLimit(size_t size);`

jordalgo (approved):
One small nit. Otherwise LGTM!

src/ast/passes/resource_analyser.cpp
ryantimwilson: Fixed in latest!

ryantimwilson:
@jordalgo there was a bug in my original implementation when I claimed that we did not need to hoist the variable initialization to the probe/subprog's entry point.

Consider the program:
```
kprobe:f { if (arg0 > 0) { $x = 1; } else { $x = 2 } }
```

This is a valid program with two locally scoped variables $x. However, based on maybeAllocVariable, we will reuse the same variable allocation for different scopes: https://github.com/bpftrace/bpftrace/blob/c14a2ce8b7afa3950e7263d509bd869cb1284b97/src/ast/passes/codegen_llvm.cpp#L2417-L2419

Thus, CodegenLLVM will:
1. Generate the variable declaration - either stack or scratch buffer - and hoist it to the entry point
2. With the previous implementation, it would only initialize the variable to 0 in the `{ $x = 1 }` scope. In `{ $x = 2 }`, it would be uninitialized. In this example, it doesn't matter but I could probably come up with an example where it might.

To fix this, I also hoisted the initialization to the entry point, which also has the nice property that there are less codegen changes.

I requested your review again if you want to take another look.

jordalgo:
@ryantimwilson Thanks for the clarification. If you can find a runtime test that would have an error with the older implementation and want to add it that would be cool but no worries if it's too time consuming. It's probably better to have less codegen churn anyway.

ryantimwilson:
@jordalgo couldn't find an implementation that made the old code fail...maybe its not possible since a variable always has to be initialized in a scope per semantic analyser. Nonetheless, keeping the hoist to reduce codegen churn

Unfortunately, I found other issues with:
1. Same variable name in multiple subprograms/probes
2. Same variable name in multiple for loops

Thus, I simplified the code (aka stopped checking if we've seen a variable before or not before bumping resources_..variable_buffers) at the cost of over-allocating memory and added test cases for both cases

jordalgo:
@ryantimwilson Good catch! And thanks for adding runtime tests for the issues. I'm surprised that we didn't have anything in place before that would have caught such things but I guess that's probably because nothing in previous tests is over the default on_stack_limit.

For the test failure, you need to add this test to the aot skip list `variable_for_loop_scratch_buf`

src/ast/irbuilderbpf.cpp
```diff
@@ -533,6 +538,32 @@ Value *IRBuilderBPF::CreateWriteMapValueAllocation(const SizedType &value_type,
       loc);
 }
 
+Value *IRBuilderBPF::CreateVariableAllocationInit(const SizedType &value_type,
+                                                  const std::string &name,
+                                                  const location &loc)
+{
+  /* Variables are lexically scoped but the same variable name may reuse
```
jordalgo, line 545: Does this comment still apply? e.g. the two variables below are not the same (as they have a different scope, as per your logic below). Also, just to be clear, the hoisting doesn't behavior right because the slots/ids are different for these two correct?

ryantimwilson: So this comment does apply because the variables do use the same slot:
```
if (arg0 > 0) { $x = 1 } else { $x = 2 }
```

But...these variables do NOT use the same slot:
```
@[1] = 1; for ($kv : @) { $y = $kv } for ($kv : @) { $y = $kv } 
```

Its basically because when using for loops, we do a bunch of logic to add/remove variables from the current scope while for if/else, we do not alter the `variables_` map.

But to my point above, I can't find a repro where not hoisting would cause bugs. So maybe I'll just remove it.

jordalgo: `if (arg0 > 0) { $x = 1 } else { $x = 2 }` - but these two variable have different entries in `variables_` right because they have different scopes or am I missing some implementation detail?

jordalgo: @ryantimwilson Ah, sorry, I was looking at `variables_` in semantic analyser. That being said, it seems we still have an issue with using only the variable names, which is broken in trunk - probably something I missed with the addition of block scoping.

Here is an example of the issue:
```
BEGIN {
	if (1) {
		$x = 1;
	}
	
	if (1) { 
		let $x; 
		if (0) {
			$x = 5;
		}
		print($x)
	} 
	exit(); 
}
```

This prints 1, when it should print 0.

Let me know if you want to try to fix forward with this PR or if I should put up one first that fixes this (as it's probably touching the same code).

ryantimwilson: @jordalgo go ahead and fix this in another PR, would appreciate it! Gives me time to debug some other stuff

jordalgo: This is almost ready: https://github.com/bpftrace/bpftrace/pull/3550

ryantimwilson: Looks like your PR is merged, I will update mine

ryantimwilson:
@jordalgo can you take another look? Implementation is the same but updated comments

jordalgo (approved):
LGTM! (pending fixing of the len_map_sum codegen)

danobi (approved):

PR #3547, opened 2024-10-28
Add config option to control maps printed at exit

### LINKED ISSUE #3545: Add the ability for users to not print maps at the end of program exit
bpftrace currently prints all maps/globals by default upon program termination and the only way to prevent this is to explicitly call `clear` on these maps in an `END` probe (though it doesn't work for scalar maps).

It would be great if users didn't have to write `clear()` multiple times if they don't want to print these maps.

My current preference would be a config variable e.g. `print_maps_on_exit`, which would default to true to not disturb the current behavior, which is great for one-liners.

Another option would be a command line arg but I think we probably want to keep these to minimum. Any other suggestions are welcome!

amscanne:
Fixes #3545

This change adds a standard configuration option, including relevant environment variables, documentation, to control whether prints are printed.

The default behavior is the current behavior.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ and this project adheres to
   - [#3461](https://github.com/bpftrace/bpftrace/pull/3461)
 - Support symbolizing enum values using `%s` specifier in `printf()`
   - [#3515](https://github.com/bpftrace/bpftrace/pull/3515)
+- Configuration option to suppress printing maps by default at program exit
+  - [#3547](https://github.com/bpftrace/bpftrace/pull/3547)
 #### Changed
 - Merge output into `stdout` when `-lv`
   - [#3383](https://github.com/bpftrace/bpftrace/pull/3383)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -3697,6 +3697,12 @@ Default: `..`
 Trailer to add to strings that were truncated.
 Set to empty string to disable truncation trailers.
 
+=== print_maps_on_exit
+
+Default: 1
+
+Controls whether maps are printed on exit. Set to `0` in order to change the default behavior and not automatically print maps at program exit.
+
 == Environment Variables
 
 These are not available as part of the standard set of <<Config Variables>> and can only be set as environment variables.
@@ -4049,8 +4055,21 @@ Therefore, when you need precise event statistics, it is recommended to use sync
 === Map Printing
 
 By default when a bpftrace program exits it will print all maps to stdout.
-If you don't want this to happen you can specify an `END` probe and `clear` the maps you don't want printed (unfortunately this doesn't work on scalar maps) e.g. this script:
+If you don't want this, you can either override the `print_maps_on_exit` configuration option or you can specify an `END` probe and `clear` the maps you don't want printed (this second method doesn't work for scalar maps).
 
+For example, this script:
+```
+config = {
+  print_maps_on_exit=0
+}
+
+BEGIN {
+  @a = 1;
+  @b[1] = 1;
+}
+```
+
+would print nothing on exit, while this script:
 ```
 BEGIN {
   @a = 1;
@@ -4068,8 +4087,7 @@ will print this on exit:
 @a: 0
 ```
 
-Additionally, if you want to get a snapshot of your maps before script exit, you can send the running bpftrace process a `SIGUSR1` signal. bpftrace will then print all maps to stdout. Example:
-
+If you want to get a snapshot of your maps while the script is running, you can send the running bpftrace process a `SIGUSR1` signal. bpftrace will then print all maps to stdout. Example:
 ```
 # bpftrace -e 'BEGIN { @b[1] = 2; }' & kill -s USR1 $(pidof bpftrace)
 ...
--- a/tests/runtime/config
+++ b/tests/runtime/config
@@ -25,3 +25,19 @@ RUN {{BPFTRACE}} -e 'config = { debug_output=1 } uprobe:./testprogs/uprobe_test:
 EXPECT stdin:1:12-25: ERROR: debug_output can only be set as an environment variable
 BEFORE ./testprogs/uprobe_test
 WILL_FAIL
+
+NAME maps are printed by default
+PROG BEGIN { @["test"] = count(); exit(); }
+EXPECT @[test]: 1
+
+NAME maps can be disabled
+PROG config = { print_maps_on_exit=0 } BEGIN { @["test"] = count(); exit(); }
+EXPECT_NONE @[test]: 1
+
+NAME scalar maps are printed by default
+PROG BEGIN { @test = 1; exit(); }
+EXPECT @test: 1
+
+NAME scalar maps can be disabled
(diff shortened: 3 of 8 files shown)


src/main.cpp
```diff
@@ -306,6 +306,10 @@ static void parse_env(BPFtrace& bpftrace)
     config_setter.set(ConfigKeyInt::max_cat_bytes, x);
   });
 
+  get_bool_env_var("BPFTRACE_PRINT_MAPS_ON_EXIT", [&](bool x) {
+    config_setter.set(ConfigKeyBool::print_maps_on_exit, x);
+  });
+
```
danobi, line 312: I view this config as a property of the source code, not the environment it's run in. ie. should only be configurable using `config = ` block. So I'd vote for omitting the env var here.

tests/runtime/config
```diff
@@ -25,3 +25,13 @@ RUN {{BPFTRACE}} -e 'config = { debug_output=1 } uprobe:./testprogs/uprobe_test:
 EXPECT stdin:1:12-25: ERROR: debug_output can only be set as an environment variable
 BEFORE ./testprogs/uprobe_test
 WILL_FAIL
+
+NAME maps are printed by default
+RUN {{BPFTRACE}} -e build/src/bpftrace -e 'uprobe:./testprogs/uprobe_test:uprobeFunction1 { @[comm] = count(); exit(); }' -p {{BEFORE_PID}}
+EXPECT @[uprobe_test]: 1
+BEFORE ./testprogs/uprobe_test
```
danobi, line 32: nit: simplify to:
```suggestion
PROG BEGIN { @[comm] = count(); exit(); }'
EXPECT @[bpftrace]: 1
```
uprobe is not necessary, so using BEGIN will help make the uprobe stuff less load bearing.

same for below

danobi (commented):
lgtm overall!

src/main.cpp
amscanne: Funny, I had the opposite view because it actually doesn't affect the probes at all, and just affects what the tool does at the end. (I first considered just adding as a command line flag, which is a further step in that direction.) Either way, I sort of feel like it's good to be consistent and just have these settable uniformly via environment variables? If you feel strongly I can remove this.

tests/runtime/config
amscanne: Thanks! Much cleaner, I will update shortly

src/main.cpp
jordalgo: I kind of agree with @amscanne though I don't feel strongly. It seems like a property or preference for whoever is running the script. That being said I know we're trying to not add more ENV vars... so 🤷

amscanne: Not relying on env variables is definitely a good thing IMHO, but personally I value consistency over most other things :) If any decider wants to call it, I'm happy to do whatever with this, I just wanted to note the differences in perspective first

danobi: I'm kinda looking at it from the view of pre-written scripts that rely on the default printing behavior to display output to user. In that case I think config only makes sense. I think there's already more config vars than env vars (can double check me). FWIW I don't have a strong preference either, probably fine to go as-is. If so, pls update the `-h` message. At some point we should think about culling the list or pointing to the docs.

danobi:
Ah, AOT has different binary name. So either change it to EXPECT_REGEX and glob over key name or just hard code a key name

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ and this project adheres to
   - [#3461](https://github.com/bpftrace/bpftrace/pull/3461)
 - Support symbolizing enum values using `%s` specifier in `printf()`
   - [#3515](https://github.com/bpftrace/bpftrace/pull/3515)
+- Configuration option to suppress map printing
```
jordalgo, line 44: Nit:
```suggestion
- Configuration option to suppress printing maps by default at program exit
```

man/adoc/bpftrace.adoc
```diff
@@ -3697,6 +3697,12 @@ Default: `..`
 Trailer to add to strings that were truncated.
 Set to empty string to disable truncation trailers.
 
+=== print_maps_on_exit
+
+Default: 1
+
+Controls whether maps are printed on exit. Set to `0` in order to change the default behavior.
```
jordalgo, line 3704: nit:
```suggestion
Controls whether maps are printed on exit. Set to `0` in order to change the default behavior and not automatically print maps at program exit.
```

tests/runtime/config
```diff
@@ -25,3 +25,11 @@ RUN {{BPFTRACE}} -e 'config = { debug_output=1 } uprobe:./testprogs/uprobe_test:
 EXPECT stdin:1:12-25: ERROR: debug_output can only be set as an environment variable
 BEFORE ./testprogs/uprobe_test
 WILL_FAIL
+
+NAME maps are printed by default
+PROG BEGIN { @["test"] = count(); exit(); }
+EXPECT @[test]: 1
+
+NAME maps can be disabled
+PROG config = { print_maps_on_exit=0 } BEGIN { @["test"] = count(); exit(); }
```
jordalgo, line 34: super nit but mind also adding a scalar map e.g. `@a = 2` and confirming that this also doesn't get printed ?

man/adoc/bpftrace.adoc
amscanne: Done

CHANGELOG.md
amscanne: Done

src/main.cpp
amscanne: I've dropped the environment variable. Even the use is ambiguous, better to not allow something that folks might depend on. It can always be revisited later.

amscanne:
Ah, I think the AOT pass configuration does not include the config pass for some reason. My instinct would be to just unify the pass configurations (parse the config) and anything that configuration needs would need to manifest in the resources.

I think the second part is already mostly true (everything needed manifesting in the resource configuration), but because the AOT configuration pass is missing, the resource calculations are exclusively based on the default configuration values? As a user, it would seem like a bit of a bug to me that the config is ignored for AOT mode, but perhaps I should fix this first separately (happy to split that unification of the passes into a separate pull request) or I could be completely misunderstanding what is happening.

jordalgo:
@amscanne Don't worry about fixing the config pass for AOT just yet (at least it's not a pre requisite for this PR). Just add the tests to the aot skip list.

tests/runtime/config
amscanne: Done

danobi (approved):

jordalgo:
@amscanne Changes look good but can I trouble you to update this adoc section to mention the new config as option for not printing maps: https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#map-printing

amscanne:
> @amscanne Changes look good but can I trouble you to update this adoc section to mention the new config as option for not printing maps: https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#map-printing

Done

jordalgo (approved):

jordalgo:
Thanks for contributing!

PR #3550, opened 2024-10-28
Fix variable scoping issues:

jordalgo:
Issue 1:
Don't save variables to the for loop context
that were declared AFTER the loop in the script.
For example:
```
@map[16] = 32;
for ($kv : @map) {
  $x = 4;
}

let $x = 5;
```

Issue 2:
Add a scope_stack similar to the one in semantic
analyser, which is needed to make sure variables
are truly isolated.

For example:
```
BEGIN {
	if (1) {
		$x = 1;
	}

	if (1) {
		let $x;
		if (0) {
			$x = 5;
		}
		print($x)
	}
}
```
On trunk this will print "1" because the `$x` variable in the first `if` is re-used in the lower scope when it shouldn't be.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1540,9 +1540,9 @@ void CodegenLLVM::visit(Variable &var)
   // Arrays and structs are not memcopied for local variables
   if (needMemcpy(var.type) &&
       !(var.type.IsArrayTy() || var.type.IsRecordTy())) {
-    expr_ = variables_[var.ident].value;
+    expr_ = getVariable(var.ident).value;
   } else {
-    auto &var_llvm = variables_[var.ident];
+    auto &var_llvm = getVariable(var.ident);
     expr_ = b_.CreateLoad(var_llvm.type, var_llvm.value);
   }
 }
@@ -2430,9 +2430,11 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
 void CodegenLLVM::maybeAllocVariable(const std::string &var_ident,
                                      const SizedType &var_type)
 {
-  if (variables_.find(var_ident) != variables_.end()) {
+  if (maybeGetVariable(var_ident) != nullptr) {
+    // Already been allocated
     return;
   }
+
   SizedType alloca_type = var_type;
   // Arrays and structs need not to be copied when assigned to local variables
   // since they are treated as read-only - it is sufficient to assign
@@ -2444,7 +2446,31 @@ void CodegenLLVM::maybeAllocVariable(const std::string &var_ident,
   }
 
   AllocaInst *val = b_.CreateAllocaBPFInit(alloca_type, var_ident);
-  variables_[var_ident] = VariableLLVM{ val, val->getAllocatedType() };
+  variables_[scope_stack_.back()][var_ident] = VariableLLVM{
+    val, val->getAllocatedType()
+  };
+}
+
+VariableLLVM *CodegenLLVM::maybeGetVariable(const std::string &var_ident)
+{
+  for (auto scope : scope_stack_) {
+    if (auto search_val = variables_[scope].find(var_ident);
+        search_val != variables_[scope].end()) {
+      return &search_val->second;
+    }
+  }
+  return nullptr;
+}
+
+VariableLLVM &CodegenLLVM::getVariable(const std::string &var_ident)
+{
+  auto *variable = maybeGetVariable(var_ident);
+  if (!variable) {
+    LOG(BUG) << "Can't find variable: " << var_ident
+             << " in this or outer scope";
+    __builtin_unreachable();
+  }
+  return *variable;
 }
 
 void CodegenLLVM::visit(AssignVarStatement &assignment)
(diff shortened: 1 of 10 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -4521,16 +4558,11 @@ Function *CodegenLLVM::createForEachMapCallback(const For &f, llvm::Type *ctx_t)
...
   for (const auto &[ident, expr] : orig_ctx_vars) {
-    variables_[ident].value = expr;
+    getVariable(ident).value = expr;
   }
 
   // Decl variable is not valid beyond this for loop
-  variables_.erase(f.decl->ident);
-
-  // Variables declared in a for-loop are not valid beyond it
-  for (const Variable &var : new_vars.nodes()) {
```
jordalgo, line 4531: No need for this as we have proper scoping now.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2428,7 +2432,25 @@ void CodegenLLVM::maybeAllocVariable(const std::string &var_ident,
...
   AllocaInst *val = b_.CreateAllocaBPFInit(alloca_type, var_ident);
-  variables_[var_ident] = VariableLLVM{ val, val->getAllocatedType() };
+  variables_[scope_stack_.back()][var_ident] = VariableLLVM{
+    val, val->getAllocatedType()
+  };
+}
+
+VariableLLVM &CodegenLLVM::getVariable(const std::string &var_ident)
+{
+  for (auto scope : scope_stack_) {
```
ryantimwilson, line 2442: Correct me if I'm wrong but don't we need to iterate through scope_stack_ in reverse order e.g. from deepest scope -> top-level scope.

For example:
```
{
  $ x= 1
  if (1) {
    let $x;
    $x = 2;
  }
}
```

In that case, I think the current code would return the variable `$x` in the top-level scope in all cases.

ryantimwilson (commented):
Looks great, other than 1 question!

src/ast/passes/codegen_llvm.cpp
jordalgo: It's a good question but we don't need to do this because bpftrace doesn't support variable shadowing. So that script above should throw an error.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2428,7 +2432,25 @@ void CodegenLLVM::maybeAllocVariable(const std::string &var_ident,
...
+        search_val != variables_[scope].end()) {
+      return search_val->second;
+    }
+  }
+  LOG(BUG) << "Can't find variable: " << var_ident << " in this or outer scope";
+
+  // unreachable because of the LOG(BUG) above
+  AllocaInst *val = b_.CreateAllocaBPFInit(CreateUInt16(), "none");
+  static VariableLLVM empty = { val, val->getAllocatedType() };
+  return empty;
```
danobi, line 2453:
```suggestion
  __builtin_unreachable();
```

tests/runtime/variable
```diff
@@ -121,3 +121,13 @@ EXPECT Attaching 1 probe...
...
+
+NAME variable doesn't escape scope
+RUN {{BPFTRACE}} runtime/scripts/variable_scope.bt
+EXPECT 0
+
+NAME late variable declaration
+RUN {{BPFTRACE}} runtime/scripts/variable_scope_late_decl.bt
+EXPECT 1
+EXPECT 2
+EXPECT 0
```
danobi, line 133: Unsure if ordering is guaranteed here. Maybe you wanna do it all in one line with `\n` separating values?

src/ast/passes/codegen_llvm.cpp
jordalgo: That's better 👍

tests/runtime/variable
jordalgo: I'm going to pair the `$x` variable with a string in a tuple to confirm they are coming from the expected places e.g.
```
EXPECT (a, 1)
EXPECT (b, 2)
EXPECT (c, 0)
```

danobi (approved):

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2430,9 +2430,13 @@ void CodegenLLVM::visit(AssignMapStatement &assignment)
 void CodegenLLVM::maybeAllocVariable(const std::string &var_ident,
                                      const SizedType &var_type)
 {
-  if (variables_.find(var_ident) != variables_.end()) {
-    return;
+  for (auto scope : scope_stack_) {
+    if (auto search_val = variables_[scope].find(var_ident);
+        search_val != variables_[scope].end()) {
```
viktormalik, line 2435: If `getVariable` returned a pointer (or `nullptr` if the variable doesn't exist), you could also use it here.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4520,7 +4545,13 @@ Function *CodegenLLVM::createForEachMapCallback(const For &f, llvm::Type *ctx_t)
   CollectNodes<Variable> new_vars;
   for (auto *stmt : f.stmts) {
     new_vars.run(*stmt, [this](const auto &var) {
-      return variables_.find(var.ident) == variables_.end();
+      for (auto scope : scope_stack_) {
+        if (auto search_val = variables_[scope].find(var.ident);
+            search_val != variables_[scope].end()) {
```
viktormalik, line 4550: and here

tests/codegen/late_variable_decl.cpp
```diff
@@ -0,0 +1,37 @@
...
+      for ($kv : @map) {
+        $x = 4;
+      }
+			
+      let $x = 5;
+    })",
+       NAME);
+}
+
+} // namespace bpftrace::test::codegen
```
viktormalik, line 37: Missing EOL

src/ast/passes/codegen_llvm.cpp
jordalgo: I thought about that but I don't want to be sprinkling checks and error message logic all over codegen. I could add a `hasVariable` function that `getVariable` could call.

viktormalik: Right, yeah, that would make sense.

viktormalik (approved):
Nice!

PR #3558, opened 2024-10-30
enums: Do not symbolize enums in json output

danobi:
https://github.com/bpftrace/bpftrace/pull/3539 inadvertently supported enum symbolization everywhere. Which is actually fine for text output, as that's what we wanted to do long term. So we add some tests to seal in that behavior there.

But we actually don't want enum symbolization for non-printf use cases in JSON output. So fix that and add tests there as well.

##### Checklist


jordalgo (approved):
LGTM. Just need to add a few more entries to our ever-growing aot skip list.

PR #3563, opened 2024-10-31
Fix typo in FindLibPcap.cmake

caringi:
libcap and libpcap are different libraries:

libcap: Library for getting and setting POSIX.1e capabilities
libpcap: A system-independent interface for user-level packet capture

The one needed by bpftrace is libpcap.


jordalgo (approved):
Thanks for the fix!

PR #3564, opened 2024-11-01
tools/bashreadline: fix probe for dynamically linked readline

ailiop-suse:
When bash links against libreadline dynamically, the probe fails as the readline symbol is undefined until runtime. Fix this issue by adding a fallback probe that hooks directly into libreadline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,8 @@ and this project adheres to
 - Remove mention of unsupported character literals
   - [#3283](https://github.com/bpftrace/bpftrace/pull/3283)
 #### Tools
+- Fix bashreadline tool probe for dynamically linked readline
+  - [#3564](https://github.com/bpftrace/bpftrace/pull/3564)
 
 ## [0.21.0] 2024-06-21
 
--- a/tools/bashreadline.bt
+++ b/tools/bashreadline.bt
@@ -15,13 +15,17 @@
  * 06-Sep-2018	Brendan Gregg	Created this.
  */
 
+config = { missing_probes = "ignore" }
+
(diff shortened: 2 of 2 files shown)


danobi (commented):
lgtm. mind updating the changelog? https://github.com/bpftrace/bpftrace/blob/master/CHANGELOG.md#tools

pevik:
Thanks a lot, fixes the issue in Tumbleweed.

danobi (approved):

danobi:
Ah, conflict in changelog. While you're at it, mind moving it to the `tools` section too?

pevik:
OT: I wonder how often you get conflicts due updating `CHANGELOG.md` as part of the change. I understand it's nice to have human readable changelog (generated git log isn't really a human readable changelog) always ready for release, but OTOH that reminds me old glibc changelog, which was part of the commit as well and it looked to me not really practical.

Maybe using github milestones which would track pull requests and issues, use as changelog the description for milestone as a temporary changelog - maintainers would manually add new entry to it. And that would be then copied into the release description and whole changelog would be out of git. A bit more work for the maintainer, but you would not bother contributers with changelog, no conflicts and you would add new items to the sections you wants. I'm actually thinking to adapt similar approach for https://github.com/iputils/iputils/ (moving away from changelog in git).

danobi:
Conflicts don't happen too often but they do happen. I think more annoying than that is you don't know your PR number until you create it, so you've gotta come back and update the changelog after PR is up.

Maybe we should just have maintainers update the changelog or something. I generally don't like too much complicated process or over-reliance on GH specific things.

PR #3569, opened 2024-11-01
Restrict use some map aggregation types

jordalgo:
lhist/hist/stats have some restrictions in that
they cannot be referenced inside tuples and single key values cannot be printed; the entire map must
be printed. This latter restriction could be relaxed; it just requires the development work to support it.

Issue: https://github.com/bpftrace/bpftrace/issues/3540

##### Checklist

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -478,8 +478,15 @@ public:
   }
   bool IsMapIterableTy() const
   {
-    return !(type_ == Type::hist_t || type_ == Type::lhist_t ||
-             type_ == Type::stats_t);
+    return !IsMultiOutputMapTy();
+  }
+
+  // These are special map value types that can't be reduced to a single value
+  // and output multiple lines when printed
+  bool IsMultiOutputMapTy() const
+  {
+    return type_ == Type::hist_t || type_ == Type::lhist_t ||
+           type_ == Type::stats_t;
   }
 
   bool NeedsPercpuMap() const;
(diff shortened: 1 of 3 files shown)


src/types.h
```diff
@@ -478,8 +478,13 @@ class SizedType {
   }
   bool IsMapIterableTy() const
   {
-    return !(type_ == Type::hist_t || type_ == Type::lhist_t ||
-             type_ == Type::stats_t);
+    return !IsRestrictedMapTy();
+  }
+
+  bool IsRestrictedMapTy() const
```
danobi, line 484: "restricted" is a bit ambiguous. what restriction? can the existing `IsPrintableTy()` be extended for this?

danobi (changes requested):

src/types.h
jordalgo: Yeah I couldn't figure out a better name. Unfortunately we can't use `IsPrintableTy` because this restriction also applies to adding these types to tuples.

jordalgo: Maybe I can utilize the new semantic_analyser function `IsValidAssignment` which checks these same types.

danobi: Can't you call IsPrintableTy() form inside the Tuple check?

jordalgo: Ok, changed the name to "IsMultiOutputMapTy" - WDYT?

danobi (approved):
lgtm.

fwiw, this is the kinda stuff I'm referring to when I say adding new fundamental types results in a long tail complexity explosion. so many different interactions in different contexts

PR #3574, opened 2024-11-06
Migrate map keys to scratch buffers

ryantimwilson:
This completes the migration of script-dependent allocations to scratch buffers by migrating map keys > 32 bytes (on_stack_limit). Once this PR is merged, all script-dependent allocations can dynamically determine to use the stack vs. scratch buffer meaning we can support large strings across all data structures.

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -712,9 +712,10 @@ void CodegenLLVM::visit(Call &call)
                              b_.getInt64Ty(),
                              call.vargs.front()->type.IsSigned());
     Value *log2 = b_.CreateCall(log2_func_, { expr_, k }, "log2");
-    AllocaInst *key = getHistMapKey(map, log2);
+    auto key = getHistMapKey(map, log2, call.loc);
     b_.CreateMapElemAdd(ctx_, map, key, b_.getInt64(1), call.loc);
-    b_.CreateLifetimeEnd(key);
+    if (dyn_cast<AllocaInst>(key))
+      b_.CreateLifetimeEnd(key);
     expr_ = nullptr;
   } else if (call.func == "lhist") {
     if (!linear_func_)
@@ -750,10 +751,10 @@ void CodegenLLVM::visit(Call &call)
                                   { value, min, max, step },
                                   "linear");
 
-    AllocaInst *key = getHistMapKey(map, linear);
+    auto key = getHistMapKey(map, linear, call.loc);
     b_.CreateMapElemAdd(ctx_, map, key, b_.getInt64(1), call.loc);
-
-    b_.CreateLifetimeEnd(key);
+    if (dyn_cast<AllocaInst>(key))
+      b_.CreateLifetimeEnd(key);
     expr_ = nullptr;
   } else if (call.func == "delete") {
     auto &arg0 = *call.vargs.at(0);
@@ -3018,13 +3019,19 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
     Expression *key_expr)
 {
   Value *key;
-  bool alloca_created_here = true;
+  const auto alloca_created_here = needMapKeyAllocation(map, key_expr);
+
   if (key_expr) {
     auto scoped_del = accept(key_expr);
+    const auto &key_type = map.key_type;
+    // Allocation needs to be done after recursing via accept(key_expr)
+    // so expr_ will be set properly
+    key = alloca_created_here ? b_.CreateMapKeyAllocation(key_type,
+                                                          map.ident + "_key",
+                                                          key_expr->loc)
+                              : expr_;
     if (inBpfMemory(key_expr->type)) {
-      auto &key_type = map.key_type;
       if (!key_expr->type.IsSameSizeRecursive(key_type)) {
-        key = b_.CreateAllocaBPF(key_type, map.ident + "_key");
         b_.CreateMemsetBPF(key, b_.getInt8(0), key_type.GetSize());
         if (key_expr->type.IsTupleTy()) {
           createTupleCopy(key_expr->type, key_type, key, expr_);
@@ -3036,25 +3043,23 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
                    << " Expression Type Size: " << key_expr->type.GetSize();
         }
       } else {
-        key = expr_;
         // Call-ee freed
         scoped_del.disarm();
-
-        // We don't have enough visibility into where key comes from to safely
-        // end its lifetime. It could be a variable, for example.
-        alloca_created_here = false;
       }
     } else if (map.key_type.IsIntTy()) {
-      key = b_.CreateAllocaBPF(map.key_type, map.ident + "_key");
       // Integers are always stored as 64-bit in map keys
       b_.CreateStore(
           b_.CreateIntCast(expr_, b_.getInt64Ty(), key_expr->type.IsSigned()),
           key);
     } else {
-      key = b_.CreateAllocaBPF(key_expr->type, map.ident + "_key");
       if (key_expr->type.IsArrayTy() || key_expr->type.IsRecordTy()) {
         // We need to read the entire array/struct and save it
-        b_.CreateProbeRead(ctx_, key, key_expr->type, expr_, key_expr->loc);
+        b_.CreateProbeRead(ctx_,
+                           b_.CreatePointerCast(
+                               key, b_.GetType(key_expr->type)->getPointerTo()),
+                           key_expr->type,
+                           expr_,
+                           key_expr->loc);
       } else {
         b_.CreateStore(
             b_.CreateIntCast(expr_, b_.getInt64Ty(), key_expr->type.IsSigned()),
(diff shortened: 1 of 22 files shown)


src/ast/passes/resource_analyser.cpp
```diff
@@ -289,6 +302,41 @@ void ResourceAnalyser::visit(Call &call)
       resources_.str_buffers++;
   }
 
+  if (call.func == "count" || call.func == "sum" || call.func == "max" ||
```
jordalgo, line 305: Curious, why not `hist` and `lhist` ?

src/ast/passes/resource_analyser.cpp
```diff
@@ -289,6 +302,41 @@ void ResourceAnalyser::visit(Call &call)
...
+    if (needMapKeyAllocation(map) && map_key_exceeds_stack_limit(map)) {
+      resources_.map_key_buffers++;
+      resources_.max_map_key_size = std::max(resources_.max_map_key_size,
+                                             get_map_key_size(map));
+    }
+  } else if (call.func == "has_key") {
+    auto &arg0 = *call.vargs.at(0);
+    auto &map = static_cast<Map &>(arg0);
+    // has_key does not work on maps with default key (aka no key_type), so we
+    // don't need to check if map.key_expr is set
```
jordalgo, line 317: nit:
```suggestion
    // has_key does not work on scalar maps (e.g. @a = 1), so we
    // don't need to check if map.key_expr is set
```

src/ast/passes/resource_analyser.cpp
```diff
@@ -417,6 +476,12 @@ bool ResourceAnalyser::exceeds_stack_limit(size_t size)
   return size > bpftrace_.config_.get(ConfigKeyInt::on_stack_limit);
 }
 
+bool ResourceAnalyser::map_key_exceeds_stack_limit(const Map &map)
+{
+  return get_map_key_size(map) >
+         bpftrace_.config_.get(ConfigKeyInt::on_stack_limit);
```
jordalgo, line 482: nit: can you use your other helper?
```suggestion
  return exceeds_stack_limit(get_map_key_size(map));
```

tests/runtime/call
```diff
@@ -163,6 +163,14 @@ NAME variable_probe_subprog_scratch_buf
 PROG config = { on_stack_limit = 0 } BEGIN { $x = "XXXX" } i:ms:100 { $x = "YYYY"; exit(); }END { $x = "ZZZZ"; print($x) }
 EXPECT ZZZZ
 
+NAME map_default_key_scratch_buf
+PROG config = { on_stack_limit = 0 } BEGIN { @x = "xxxx"; print(@x) }
```
jordalgo, line 167: nit: add an `exit()` to both of these

src/ast/passes/resource_analyser.cpp
ryantimwilson: Oops I forgot them since they use getMultiMapKey / getHistKey, I have fixed in latest version

src/ast/passes/resource_analyser.cpp
```diff
@@ -289,6 +302,53 @@ void ResourceAnalyser::visit(Call &call)
       resources_.str_buffers++;
   }
 
+  if (call.func == "count" || call.func == "sum" || call.func == "max" ||
```
jordalgo, line 305: @ryantimwilson Out of curiosity, can we get away with removing all this code in the `visit(Call &call)` function if we add checks for the map key in `visit(Map &map)`, which you have and in `visit(AssignMapStatement &assignment)` ?

ryantimwilson: @jordalgo unfortunately no for 2 reasons:
1. All calls except has_key and delete set `call.map`. But `Visitors::visit(Call &call)` does not recurse into `call.map` AST node: https://github.com/bpftrace/bpftrace/blob/941391743d58a418cc075983760e79bc8732ee95/src/ast/visitors.cpp#L31-L36. And changing the recursion would likely increase allocations for other buffers, so I recommend we don't change that.
2. For has_key and delete, arg0 is the map. However, these functions have multiple arguments and whether the buffer size depends on the additional argument that we can't access within `visit(Map &map)` and in `visit(AssignMapStatement &assignment)`. For example, for `delete(@, 2)`, `map.key_expr` will be set to nullptr but we will call `needMapKeyAllocaton(@, 2)`

I can add comments to explain this.

jordalgo: Ok, so what am I missing here. I pulled down this PR and I removed this bit:
```
if (call.func == "count" || call.func == "sum" || call.func == "max" ||
      call.func == "min" || call.func == "avg" || call.func == "stats") {
    Map &map = *call.map;
    if (needMapKeyAllocation(map) && map_key_exceeds_stack_limit(map)) {
      resources_.map_key_buffers++;
      resources_.max_map_key_size = std::max(resources_.max_map_key_size,
                                             get_map_key_size(map));
    }
  } 
```
And I still got the key scratch buffer because the key_expr exists in `visit(AssignMapStatement)`.

ryantimwilson: Oh yea good point. Its because for these functions, they are always called like `@ = count()`, so it will always trigger AssignMapStatement AST node where we already add map key buffer.

I'll remove this and add a comment why they're not here.

Note we need to keep has_key/delete for the issue I mentioned above and lhist/hist because the buffer is specially sized to include log2/linear function pointer.

src/ast/passes/resource_analyser.cpp
```diff
@@ -307,6 +374,11 @@ void ResourceAnalyser::visit(Map &map)
     resources_.max_read_map_value_size = std::max(
         resources_.max_read_map_value_size, map.type.GetSize());
   }
+  if (needMapKeyAllocation(map) && map_key_exceeds_stack_limit(map)) {
+    resources_.map_key_buffers++;
```
jordalgo, line 378: Nit: these three lines are repeated. Could be extracted into a helper.

tests/codegen/scratch_buffer.cpp
```diff
@@ -115,6 +115,20 @@ TEST(codegen, variable_stack)
...
 }
 
+TEST(codegen, map_key_scratch_buf)
+{
+  test_stack_or_scratch_buffer("kprobe:f { @x[1] = 1; @y[\"yyyy\"] = @x[1]; }",
+                               NAME,
+                               SMALL_ON_STACK_LIMIT);
+}
+
+TEST(codegen, map_key_stack)
```
jordalgo, line 125: Is it worth adding codegen tests for some of the function calls e.g. `hist` or `has_key` ?

jordalgo (commented):
One more nit and one more question but otherwise LGTM!

src/ast/passes/resource_analyser.cpp
ryantimwilson: Sounds good, I extracted AssignMapStatement and Map parts out to a common function.

lhist/hist/has_key/delete are hard to extract out since they all have slight differences and I found extracting out those parts made it harder to read. Let me know if you disagree though.

jordalgo: Might just have to see it first but it's not a big deal either way.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3063,19 +3068,23 @@ std::tuple<Value *, CodegenLLVM::ScopedExprDeleter> CodegenLLVM::getMapKey(
     }
   } else {
     // No map key (e.g., @ = 1;). Use 0 as a key.
-    key = b_.CreateAllocaBPF(CreateUInt64(), map.ident + "_key");
+    assert(alloca_created_here);
+    key = b_.CreateMapKeyAllocation(CreateUInt64(),
+                                    map.ident + "_key",
+                                    map.loc);
     b_.CreateStore(b_.getInt64(0), key);
```
danobi, line 3075: This doesn't scale with the types used in the script, so perhaps better to keep this a stack allocation? That way we don't need unnecessary preallocation in scratch map.

src/ast/passes/resource_analyser.cpp
```diff
@@ -289,6 +297,60 @@ void ResourceAnalyser::visit(Call &call)
...
 
+  /* Aggregation functions like count/sum/max are always called like:
+   *   @ = count()
+   * Thus, we visit AssignMapStatement AST node which visits the map and
+   * assigns a map key buffer. Thus, there is no need to assign another
+   * buffer here.
+   *
+   * The exceptions are:
+   * 1. lhist/hist because the map key buffer includes both the key itself
+   *    and a function pointer to the linear/log2 functions.
```
danobi, line 308: Not sure it's a function pointer. Looks like the result of a call to log2 - ie the bucket. right?

src/ast/passes/codegen_llvm.cpp
ryantimwilson: I thought about this too but I propose we keep it this way because this makes the definition of on_stack_limit consistent e.g. if I set on_stack_limit = 0, all allocations of map keys will be on the scratch buffer. This also makes testing easier as I can run standard runtime test cases with on_stack_limit = 0 instead of having to know that default maps don't use the scratch buffer.

The good news is in real usage, this will always be on the stack if on_stack_limit >= 8 - by default, its 32.

src/ast/passes/resource_analyser.cpp
ryantimwilson: Ah I think you're right, I misread the code, I will fix

src/ast/passes/codegen_llvm.cpp
danobi: Makes sense. Consistency and the fact that on_stack_limit makes it a stack allocation in practice are valid reasons

danobi: > I'm also not quite sure what you mean by this.

Oh, I just meant that it's not script dependent. ie it'll always be a 64 bit allocation regardless of what type the user puts in the map.

danobi (approved):

PR #3575, opened 2024-11-06
Fix build on clang 17

danobi:
We're seeing a clang 17 build failure:

```
... struct.h:15:30: error: constexpr variable cannot have non-literal type 'const std::string' (aka 'const basic_string<char>')
   15 | static constexpr std::string RETVAL_FIELD_NAME = "$retval";
      |                              ^
... bits/basic_string.h:85:11: note: 'basic_string<char>' is not literal because it is not an aggregate and has no constexpr constructors other than copy or move constructors
   85 |     class basic_string
      |           ^
1 error generated.
```

Seems correct - std::string is not very const. Fix by changing it to raw string which should construct temporary std::string& correctly.

##### Checklist

### DIFF
--- a/src/struct.h
+++ b/src/struct.h
@@ -12,7 +12,7 @@
 
 namespace bpftrace {
 
-static constexpr std::string RETVAL_FIELD_NAME = "$retval";
+static constexpr auto RETVAL_FIELD_NAME = "$retval";
 
 struct Bitfield {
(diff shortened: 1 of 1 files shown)


danobi:
Harmless change that unblocks build stuff, so merging. Feel free to yell at me if it goes pear shaped

viktormalik:
Thanks!

PR #3576, opened 2024-11-06
cmake: Use stricter include paths for src/ast

danobi:
The target_include_directories() on parser target was leaking into other cmake targets. Making include paths stricter keeps the code more uniform. It also matters for stricter build systems like buck.

##### Checklist


jordalgo (approved):
Nice!

PR #3577, opened 2024-11-06
cmake: Simplify handling of include directories

danobi:
Simplify cmake a bit. See each commit for details.

There's some churn between commits 1 and 3, but since they don't completely
cancel each other out, I figured I'd leave both in so git history is a bit more
descriptive.

##### Checklist


jordalgo (approved):
I don't know much about Cmake so this is a trust-you stamp. Feel free to wait for other, more-qualified people to review.

danobi:
Waiting for a cmake review, then :)

src/aot/CMakeLists.txt
```diff
@@ -1,9 +1,6 @@
 add_library(aot STATIC aot.cpp)
-add_dependencies(aot parser version_h)
-target_link_libraries(aot required_resources)
-target_include_directories(aot PUBLIC ${CMAKE_SOURCE_DIR}/src)
-target_include_directories(aot PUBLIC ${CMAKE_BINARY_DIR}/src)
-target_include_directories(aot PUBLIC ${CMAKE_BINARY_DIR})
+add_dependencies(aot version_h)
+target_link_libraries(aot required_resources parser)
```
viktormalik, line 3: Why is `parser` moved here?

viktormalik (commented):
Nice! I was always confused how some includes got into some subdirs. This makes things much simpler.

Just one question below.

src/aot/CMakeLists.txt
danobi: add_dependencies was not necessary for `parser`: https://stackoverflow.com/questions/74043615/when-should-i-use-cmakes-add-dependencies-instead-of-target-link-libraries

The link dependency is enough. However for version_h, it's just a header file, so need to manually set the dep

viktormalik: Makes sense, thanks.

viktormalik (approved):

Issue #3147, opened 2024-05-01
Don't print out maps on exit if END probe is defined


ajor:
If the `END` probe is defined, don't print out maps automatically when bpftrace exits.

This will change the behaviour of some existing scripts, so could do with a discussion.

I often see the `END` probe being used solely to `clear()` all maps so that they're not printed by bpftrace, e.g.: https://github.com/bpftrace/bpftrace/blob/c8fd334f55e6b04a7eb3fc9a06822fc6b2a6344b/tools/biosnoop.bt#L48-L54

It's tedious to write this in most scripts and keep the list of maps to be cleared up-to-date as the script evolves.

Analysing the current state of the bpftrace tools, 18 scripts use `END` to clear maps and 18 rely on the implicit printing without specifying an `END` probe. 3 scripts define and `END` probe and also rely on the implicit printing. They all take the form:
https://github.com/bpftrace/bpftrace/blob/c8fd334f55e6b04a7eb3fc9a06822fc6b2a6344b/tools/bitesize.bt#L26-L29
But I think this would be better if they explicitly printed out the relevant map anyway, instead of relying on the implicit print-on-exit behaviour.

danobi:
Agreed that current behavior is kinda annoying.

But could be risky. There might be automated users (eg https://github.com/iovisor/kubectl-trace and others) that rely on the exit printing. If we change the behavior they could lose stats.

danobi:
Actually, current behavior is beneficial for one liners. Which is probably why it’s like this in the first place

viktormalik:
I agree with @danobi here. The "feature" has been around for a long time and many automated users of bpftrace may rely on it.

What about we introduced a config option (and possibly a CLI option) that would auto-clear all the maps prior to exiting?

jordalgo:
I agree that I don't like auto printing of maps at the end AND that maybe too many people might rely on this default behavior to change it (at least without a lot of notice).

What about instead of implicit map clearing with `END`, which feels a little magical. What about introducing a single new function `clear_all_maps`.

lenticularis39:
I think we could even do both `clear_all_maps` and a config option/command line switch. The config option way seems more natural to me, since it is not an actual bpftrace statement which prints the maps, but bpftrace itself, thus, it feels like it should be controlled by bpftrace. Clearing all maps so they are not displayed feels more like a hack.

[...]

PR #3578, opened 2024-11-06
Don't print out scalar maps if clear()ed in END

danobi:
Since b4a70de7 ("Keyless maps default to BPF_MAP_TYPE_ARRAY for some types (#3300)"), scalar maps are backed by BPF_MAP_TYPE_ARRAY. Which means their entries cannot be cleared. This has started causing confusing issues where scalar map entries are printed out with value 0.

This commit closes the hole somewhat by teaching the runtime to reliably keep track of which maps are cleared in the END probe.

Note that we technically still do the right thing even given something like this:

    BEGIN { @s = 5 } END { clear(@s); @s = 1 }

That is b/c clear() is defined as an async helper. Which means that the user is racing with clear() and we are free to choose either outcome. So we happen to choose the cleared outcome.

However, it does not solve the case where scalar maps are cleared in non-END probes like this:

    BEGIN { @s = 5; clear(@s) }

But at least it doesn't make the problem worse.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,8 @@ and this project adheres to
   - [#3542](https://github.com/bpftrace/bpftrace/pull/3542)
 - Fix field access and offsetof for strings that are builtin types
   - [#3565](https://github.com/bpftrace/bpftrace/pull/3565)
+- Fix printing of cleared scalar maps as 0 value even if cleared in END probe
+  - [#3578](https://github.com/bpftrace/bpftrace/pull/3578)
 #### Security
 #### Docs
 - Remove mention of unsupported character literals
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -4090,9 +4090,9 @@ Therefore, when you need precise event statistics, it is recommended to use sync
 === Map Printing
 
 By default when a bpftrace program exits it will print all maps to stdout.
-If you don't want this, you can either override the `print_maps_on_exit` configuration option or you can specify an `END` probe and `clear` the maps you don't want printed (this second method doesn't work for scalar maps).
+If you don't want this, you can either override the `print_maps_on_exit` configuration option or you can specify an `END` probe and `clear` the maps you don't want printed.
 
-For example, this script:
+For example, these two scripts are equivalent and will print nothing on exit:
 ```
 config = {
   print_maps_on_exit=0
@@ -4104,7 +4104,6 @@ BEGIN {
 }
 ```
 
-would print nothing on exit, while this script:
 ```
 BEGIN {
   @a = 1;
@@ -4115,11 +4114,6 @@ END {
   clear(@a);
   clear(@b);
 }
-```
-
-will print this on exit:
-```
-@a: 0
 ```
 
 === Systemd support
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1068,6 +1068,7 @@ int BPFtrace::run(BpfBytecode bytecode)
 #endif
 
   attached_probes_.clear();
+  cleared_map_fds_.clear();
   // finalize_ and exitsig_recv should be false from now on otherwise
   // perf_event_printer() can ignore the `END` events.
   finalize_ = false;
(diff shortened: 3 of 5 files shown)


danobi:
Also on the table is to just revert the scalar map specialization change. This PR is basically a workaround. I don't wanna dig this hole any deeper than this, so I'd say if we see any more issues after this change we revert the whole lot.

jordalgo:
> However, it does not solve the case where scalar maps are cleared in non-END probes like this:

It does feel like this is a little bit of hack but I also feel that way about `clear` being tied to map printing. Also our docs only talk about calling `clear` in `END` probes: https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#map-printing

So I think I'm fine with this change. This also removes one of the blockers for the 0.22 release. I'll wait to see if @ajor @viktormalik chime in.

jordalgo:
@danobi I know you didn't want to dig any deeper but would it make this a little more correct if we also checked the contents of the scalar map before choosing to print or not print?

danobi:
> @danobi I know you didn't want to dig any deeper but would it make this a little more correct if we also checked the contents of the scalar map before choosing to print or not print?

Given that 0 is a valid value, I'd say it's probably a hack too far :P

danobi:
> > However, it does not solve the case where scalar maps are cleared in non-END probes like this:
> 
> It does feel like this is a little bit of hack but I also feel that way about `clear` being tied to map printing. Also our docs only talk about calling `clear` in `END` probes: https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#map-printing
> 
> So I think I'm fine with this change. This also removes one of the blockers for the 0.22 release. I'll wait to see if @ajor @viktormalik chime in.

Oh, good call on the docs. I'll update that here. But shall wait for more discussion on this.

jordalgo:
> But shall wait for more discussion on this.

Yeah let's wait a few days. No rush on this anyway.

jordalgo:
> Given that 0 is a valid value, I'd say it's probably a hack too far :P

Is it? We would at least know that if the value is not 0 then the map was assigned after the clear.

danobi:
Ah interesting idea. So you mean for the `clear(@s); @s = 1` case?

jordalgo:
> Ah interesting idea. So you mean for the clear(@s); @s = 1 case?

Yeah exactly.

danobi:
Hmm, thinking again, seems not worthwhile to handle. Cuz why not just do `@s = 1` alone instead of clearing then setting? Doesn't seem realistic for user to do.

jordalgo:
> Cuz why not just do @s = 1 alone instead of clearing then setting? Doesn't seem realistic for user to do.

Yeah true. Ignore my suggestion, it's probably unnecessary.

ajor:
Is the PR description out of date or am misunderstanding the implementation? What's the issue with this:

> However, it does not solve the case where scalar maps are cleared in non-END probes like this:
> 
> ```
> BEGIN { @s = 5; clear(@s) }
> ```

The change does feel kind of hacky but I'm not against it - it just seems to preserve v0.21 behaviour as far as I understand(?)

danobi:
@ajor b/c after we detach normal probes and before we run END probe we run `cleared_map_fds_.clear();`

ajor:
Ah I get it now. That does make me less keen on this solution then - `clear()` doing a different thing in `END` probes vs other probes seems wrong to me.

I'm going to push for #3147 as the solution again. It would be a breaking change, but it'd be consistent and I think it makes more sense from a user's point of view - overriding default, compiler-defined functions is a thing in other languages.

danobi:
I would much rather revert the arraymap optimization than go that route.

viktormalik:
> > I would much rather revert the arraymap optimization than go that route.
> 
> In order to unblock the new release, while we're still discussing this issue, I agree with @danobi here.

Short term, I agree. I'd really like to get that release done so reverting the optimization seems like the best option for now.

Long term, I think that it makes sense to give users an opportunity to disable auto-printing of the maps at the end or to let them print whatever they want (in whatever format). A default print-all-maps END probe which can be overridden sounds like quite an elegant solution to achieve that.

danobi:
We already have `print_maps_on_exit` config: https://github.com/bpftrace/bpftrace/commit/cdf5e10ed18c4ab36373ead4cef765b4960d035d

danobi:
Going with a revert

PR #3585, opened 2024-11-13
Support `fn` keyword for external names

### LINKED ISSUE #3177: subprogram support causes problems accessing fentry args named 'fn'

On a 5.19 kernel with bpftrace HEAD commit == 742f0cf1 we get a naming clash when attempting to access an fentry argument named 'fn'.

```
$ sudo ./bpftrace -lv fentry:vmlinux:module_kallsyms_on_each_symbol
kfunc:vmlinux:module_kallsyms_on_each_symbol
    const char * modname
    int (*)(void *, const char *, struct module *, unsigned long) fn
    void * data
    int retval

$ sudo ./bpftrace -e 'fentry:vmlinux:module_kallsyms_on_each_symbol{ $t = args.fn; }'
stdin:1:53-60: ERROR: syntax error, unexpected subprog
fentry:vmlinux:module_kallsyms_on_each_symbol{ $t = args.fn; }
                                                    ~~~~~~~
```

@ajor suggested adding the SUBPROG token as a known keyword in the grammar as so:

```
diff --git a/src/parser.yy b/src/parser.yy
index 85c8a20e..89efca28 100644
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -628,6 +628,7 @@ keyword:
         |       SIZEOF        { $$ = $1; }
         |       UNROLL        { $$ = $1; }
         |       WHILE         { $$ = $1; }
+        |       SUBPROG       { $$ = $1; }
         ;
```

The above modification makes bpftrace happier but whether it is a complete fix is another thing.

```
$ sudo ./bpftrace -e 'fentry:vmlinux:module_kallsyms_on_each_symbol{ $t = args.fn; }'
Attaching 1 probe...
ERROR: Error loading program: kfunc:vmlinux:module_kallsyms_on_each_symbol (try -v)
```

(The error here is a verifier problem and should be ignored for the purpose of this bug).

amscanne:
Fixes #3177

This change adds `fn` to the list of external names that may be used to index structures, etc. I had originally considered allowing `fn` (and `config`) arbitrarily as an identifier without the body of a scope (esentially extending `ident` to include CONFIG and SUBPROG within the parser, and limiting `attach_point_def` accordingly), but since there is already a strong split between internal and external identifiers, I went with the simpler proposed solution.

##### Checklist

- [*] User-visible and non-trivial changes updated in `CHANGELOG.md`

* Trivial bug fix, not really CHANGELOG worthy


danobi (approved):
makes sense to me

man/adoc/bpftrace.adoc
```diff
@@ -978,6 +978,7 @@ These exist for the lifetime of `bpftrace` itself and can be accessed from all a
 Map names always start with a `@`, e.g. `@mymap`.
 
 All valid identifiers can be used as `name`.
+ Note that there are restrictions on the use of identifiers beyond C, e.g. `config` is reserved as a language keyword and may not be used directly.
```
jordalgo, line 981: This seems to work for me:
```
sudo bpftrace -e 'BEGIN { @config = 1; exit(); }'
```

danobi: I guess it depends on your read of "directly". `@config` is namespaced by `@`

jordalgo: This is in a section about map naming so I think this addition is confusing as no identifiers without @ can be used for maps.

amscanne: Yeah, you're right... the problem only exists for bare identifiers, so you couldn't name a `struct config` or subprogram `config`. As this change isn't really changing any of those semantics, I'll drop this line here for now but might be worthwhile to have a kind of language spec/guide (both for bring forward some clear decisions and for the user). Happy to help on that.

Issue #1452, opened 2020-08-04
Allow variable printf format


albancrommer:
### bpftrace --info

Debian Buster version 0.8+git60-gccac69c2239b-2

### Use Case

Say you want to write a BT program with many probes, all of them printing identically formated lines, for example "TIME PID SRC COMM ARGS" with sources like tcp, blk, etc. 

To avoid repeating yourself for every probe, you want to have the `printf` format as a variable used for every printf statement, like

```
BEGIN
{
  @fmtstr = "%-8s\t%-6s\t%-16s\t%-16s\t%-32s\n";
}

tracepoint:syscalls:sys_enter_open {
  printf(@fmtstr, ... );
}


```

### What reproduces the bug?

Using a string variable as first parameter (i.e. the format) with printf.

**Expected result**
The string variable should be accepted.

**Actual result**
BT returns the error `printf() expects a string literal (string provided)`

```
#!/usr/bin/env bpftrace

BEGIN
{
  @fmtstr = "%s\n";
  printf( @fmtstr, "This does not work.");
}

```

fbs:
For your use case a define seems better:

```
$ sudo bpftrace -e '#define FMT "%d\n"
>
> i:ms:100 { printf(FMT, @); @++ }'
Attaching 1 probe...
0
1
2
3
4
```

Probe wildcards might  also work in your case, avoids having to dupe a lot of code.

Having a way to define global constants (which we can optimize at compile time) sounds like a good thing to add, something like 

```
let VAR="%d\n"

k:f { printf(VAR, @) }
```

albancrommer:
Thanks for yout feedback.

Sadly, the `define` way doesn't seem to work on my end...
```
$ sudo bpftrace -e '#define FMT "%d\n"
> i:ms:100 { printf(FMT, @); @++ }'

2.22: syntax error, unexpected ",", expecting (
```
Made different attempts in scripts, I always get that syntax error.

Am I doing something wrong? Is it a build flag or anything related to my BT version?

danobi:
> Having a way to define global constants (which we can optimize at compile time) sounds like a good thing to add, something like

Yeah, and functions would be nice too.

Issue #430, opened 2019-02-19
ustackf() and kstackf()


brendangregg:
Follow on from #287 by @mmarchini 

I'd like to print the stack trace in different ways. Eg:
- The library name in the stack output
- Folded format, as used by flame graphs: the stack on one line, semi-colon delimitered, with a count at the end. (Would be great to strip semicolons from symbols, which can work their way in there from perf-map-agent).

Also, I'm wondering if we can make ustack, ustack(), kstack, and kstack() print if used stand-alone. So this:

```
bpftrace -e 'k:do_nanosleep { kstack }
```

does this:

```
bpftrace -e 'k:do_nanosleep { prinf("%s", kstack) }
```

We can file that as a separate issue, as it may grow to include other data types as well. Eg the `{ arg0 }` program could print it as an int.

For this issue, I'd like a ustack formatted and kstack formatted, and ways to emit them with the library name and with folded format. @mmarchini already has the #287 PR with much of the work.

I get that we can invent a printf()-like format string, with %'s, but if the input is a string then could it also support aliases? Eg:

```
kstackf("perf")
kstackf("folded")
```

perf == Linux perf-inspired output
folded == folded format

In fact, we could do that to start with, and only do the %'s customizations later if really necessary (remembering we can always defer to bcc for exotic stuff).

To make it future proof, it may need a special character to denote aliases, just like % denotes fields. Since in the future we may want it to support mixed strings with % fields. Or maybe we have it like this:

```
kstackf(perf)
kstackf(folded)
kstackf("...format string...")
```

So if quote marks are used, it's a format string, but otherwise it's an alias. Doable?

If that's possible, then we may not even need kstackf()/ustackf(). If it can detect the type of argument as either an integer, string, or alias (perf/folded), then maybe we just add the ability to the existing kstack() ustack(), and make it support multiple options. Eg:

```
kstack(5)
kstack(perf)
kstack(perf, 5)
kstack("...format string''', 5)
```

That would be pretty nice.

mmarchini:
> I get that we can invent a printf()-like format string, with %'s, but if the input is a string then could it also support aliases?

I like this idea! I think I can tweak #287 to implement as `kstack(perf)`/`ustack(perf)`. I'll try to look into it later this week.

jonhoo:
[...]

Issue #404, opened 2019-02-07
RFC: -i 1


brendangregg:
An idea to consider, -i:

```
bpftrace -i 1 -e 'k:do_nanosleep { @[comm] = count(); }'
```

would do

```
bpftrace -e 'k:do_nanosleep { @[comm] = count(); } interval:s:1 { print(@); clear(@); }'
```

so -i takes an interval in seconds, and prints all maps and clears them. Because I type that interval thing in too much.

tyroguru:
@brendangregg - I completely agree that some mechanism that provided a shorthand version of the whole interval/print/clear construct would be very welcome. As you say, it's something that you tend to do all the time. 

I'm not sure on how this should be achieved yet though as I'm generally against using command line parameters for specific functionality like this as it just feels like this should be a language feature. Would be super useful however it ends up getting there. I'll ponder it.

ajor:
Closing as won't-do in this form.

As much logic as possible should remain in the BpfScript language for increased flexibility. The command line option only works for this one particular scenario. e.g. what if someone wants to print all maps every second, but not clear them?

We also already have three command line flags that begin with "i"; adding more could get confusing.

Alternatives could be considered if there is still a desire for this shorthand functionality. For example, one approach could be to add a BpfScript mechanism to iterate over all maps in a program (metaprogramming?!), define a function that does the requested printing/clearing operations, put that function in a library, then just import that library into scripts as required.

PR #3588, opened 2024-11-15
feature: add automatic printing for bare identifiers

### LINKED ISSUE #431: By-default printing
Just an idea. Imagine this program:

```
bpftrace -e 'k:vfs_read { arg0; kstack; };
```

Did this:

```
bpftrace -e 'k:vfs_read { printf("%d\n", arg0); printf("%s\n", kstack); }
```

What does a bare `arg0` and `kstack` do in bpftrace anyway? Nothing useful. Can it be detected by lexer/parser and turned into a printf based on its type?

It'd also lead to one-liners like this:

```
bpftrace -e 'k:do_nano* { probe }'
```

To find out which probe is firing.

This is a follow on from #430 .

amscanne:
This change adds a pass that expands bare identifier statements into print statements. Since bare identifier statements were side-effect free and would have been optimized out, this shouldn't break any existing functionality or programs.

Fixes #431

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to
   - [#3522](https://github.com/bpftrace/bpftrace/pull/3522)
   - [Migration guide](docs/migration_guide.md#default-sigusr1-handler-removed)
 #### Added
+- Bare identifier statements (variables, maps, builtins) are implicit prints
+  - [#3588](https://github.com/bpftrace/bpftrace/pull/3588)
 - Add `--dry-run` CLI option
   - [#3203](https://github.com/bpftrace/bpftrace/pull/3203)
 - Enable avg map reads in kernel space (implicit casting)
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1947,6 +1947,21 @@ Changing the print call to `print(@, 5, 2)` will take the top 5 values and scale
 @[10]: 5
 ----
 
+As a convenience, variables, maps and builtins will be automatically printed when they are appear alone as a statement.
+The following two programs are equivalent:
+
+----
+kprobe:f {
+  print(kstack)
+}
+
+kprobe:f {
+  kstack
+}
+----
+
+This does not work for more complex expressions, tuples, calls, or anything else that may have side-effects.
+
 [#functions-printf]
 === printf
 
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -12,6 +12,7 @@ add_library(ast STATIC
   signal.cpp
   visitors.cpp
 
+  passes/auto_print.cpp
   passes/codegen_resources.cpp
   passes/config_analyser.cpp
   passes/field_analyser.cpp
--- a/src/ast/passes/auto_print.cpp
+++ b/src/ast/passes/auto_print.cpp
@@ -1,0 +1,44 @@
+#include "auto_print.h"
+
+#include "log.h"
+
+namespace bpftrace::ast {
+
+AutoPrintAnalyser::AutoPrintAnalyser(ASTContext &ctx) : ctx_(ctx)
+{
+}
+
+template <typename T>
+void promote(ASTContext &ctx, ExprStatement &statement)
+{
+  if (auto n = dynamic_cast<T *>(statement.expr)) {
+    statement.expr = ctx.make_node<ast::Call>("print",
+                                              ExpressionList{ n },
+                                              n->loc);
+  }
+}
+
+void AutoPrintAnalyser::visit(ExprStatement &statement)
+{
+  // If this statement is a bare identifier, without any side-effects, then
+  // automatically promote into a print statement. Note that this was
+  // previously guarranteed to be optimized away during compilation, so it is
+  // unlikely that any programs existed with a vestigal identifier statement.
+  promote<Builtin>(ctx_, statement);
+  promote<Variable>(ctx_, statement);
+  promote<Map>(ctx_, statement);
+  Visitor::visit(statement);
+}
+
+Pass CreateAutoPrintPass()
+{
+  auto fn = [](Node &n, PassContext &ctx) {
+    AutoPrintAnalyser analyser(ctx.ast_ctx);
+    analyser.Visit(n);
+    return PassResult::Success();
+  };
+
+  return Pass("AutoPrintAnalyser", fn);
+}
+
+} // namespace bpftrace::ast
--- a/src/ast/passes/auto_print.h
+++ b/src/ast/passes/auto_print.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include <iostream>
+#include <sstream>
+
+#include "ast/pass_manager.h"
+#include "ast/visitors.h"
+
+namespace bpftrace {
+namespace ast {
+
+// Automatically promotes bare identifier statements to print statements.
+class AutoPrintAnalyser : public Visitor {
+public:
+  AutoPrintAnalyser(ASTContext &ctx);
+
+private:
+  void visit(ExprStatement &statement) override;
+
+  ASTContext &ctx_;
+};
+
+Pass CreateAutoPrintPass();
+
+} // namespace ast
+} // namespace bpftrace
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -15,6 +15,7 @@
 #include "aot/aot.h"
 #include "ast/pass_manager.h"
 
+#include "ast/passes/auto_print.h"
 #include "ast/passes/codegen_llvm.h"
 #include "ast/passes/config_analyser.h"
 #include "ast/passes/field_analyser.h"
@@ -425,6 +426,7 @@ ast::PassManager CreateDynamicPM()
 {
   ast::PassManager pm;
   pm.AddPass(ast::CreateConfigPass());
+  pm.AddPass(ast::CreateAutoPrintPass());
   pm.AddPass(ast::CreateSemanticPass());
   pm.AddPass(ast::CreateCounterPass());
   pm.AddPass(ast::CreateResourcePass());
@@ -436,6 +438,7 @@ ast::PassManager CreateDynamicPM()
 ast::PassManager CreateAotPM()
 {
   ast::PassManager pm;
+  pm.AddPass(ast::CreateAutoPrintPass());
   pm.AddPass(ast::CreateSemanticPass());
   pm.AddPass(ast::CreatePortabilityPass());
   pm.AddPass(ast::CreateResourcePass());
--- a/tests/codegen/auto_print.cpp
+++ b/tests/codegen/auto_print.cpp
@@ -1,0 +1,24 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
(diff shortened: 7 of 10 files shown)


danobi:
This is very neat. I can't think of any potential issues at the moment (btw only glanced at code). Makes me think of implicit returns in rust.

One thought is we probably want to update the one-liner tutorial with an example of this: https://github.com/bpftrace/bpftrace/blob/master/docs/tutorial_one_liners.md#lesson-2-hello-worldhttps://github.com/bpftrace/bpftrace/blob/master/docs/tutorial_one_liners.md#lesson-2-hello-world

And maybe even the tools in `tools/`.

But probably good to first have a discussion on this feature.

jordalgo:
+1 This is great for the one-liner use case! I definitely want to take some time to think about the implications of allowing this syntax though. Like @danobi, I can't think of any immediate issues. 

Why is it that tuples are not supported e.g. `BEGIN { (1, "hello");` or the result of a function call? Is that just out of scope for this initial implementaiton?

jordalgo:
Also some additional unit and runtime tests would be good to add.

amscanne:
I'm good with letting this sit and marinate for a little while, I also think it's good to be careful about what's allowed and what's not. I'm just going through older issues and knocking out a fix or implementation to learn more about the internals. I'll circle back on this one in a few weeks and see how everyone feels (including myself!), and then update docs and tests more thoroughly.

danobi:
The more I sit on this the more I like it

jordalgo:
I like it too. Any thoughts @fbs @viktormalik @ajor ? 

I'm trying to think of the potential downsides or what future syntax this might get in the way of. We would, of course, feature gate this to start if we added it.

viktormalik:
Is this much of an improvement from using `print`? I'm not strictly against this feature but I'm wondering if the benefit we get (saving 7 characters) is worth potential negatives (especially confusion of users as it's not super-intuitive).

jordalgo:
@viktormalik As per our [language design doc](https://docs.google.com/document/d/1a5vla623n0Z9oaUCNLaaDux6X6FO4qYfBs9zLJ6nzhA/edit?tab=t.0) the top priority item is "conciseness / one-liners" so I think saving 7 characters for EACH print call fits nicely into that item.

> especially confusion of users as it's not super-intuitive

At the moment just having bare identifiers does nothing e.g. `BEGIN { $a = 1; $a; }`. This almost feels like the more confusing behavior.

tyroguru:
Sorry for the delay in commenting on this. TBH, I'm not a massive fan and my thoughts are along the lines that @viktormalik has just written. I'm personally not in favour of conciseness if it's just to support one liners and I'm not even sure that having a language priority behind that is a good thing. I feel this change will lead to scripts that look more confusing and that's not a good thing - we should be aiming for improving readability and not the other way.

viktormalik:
> At the moment just having bare identifiers does nothing e.g. `BEGIN { $a = 1; $a; }`. This almost feels like the more confusing behavior.

I'd argue that it doesn't as it's quite a common behaviour in most programming languages. A bare identifier is a statement with no effect. It usually triggers a compiler warning but it is not forbidden and has no implicit behaviour (at least I'm not aware of any language where it does).

jordalgo:
> I'm personally not in favour of conciseness if it's just to support one liners and I'm not even sure that having a language priority behind that is a good thing.

As we've seen, one of bpftrace's largest use-cases (if not the largest) is one-liners - I would imagine this is true for awk (one of the inspiration languages). awk also has implicit printing (on top of an explicit `print` function) e.g. `awk 'NR % 2 == 0' data` prints the even numbered lines in `data`.

The reason I like automatic/implicit printing is because bpftrace's goal is to print things (rather than drive a backend service or UI application) so making that easier to do seems like a good thing (as long as it doesn't cause any future issues with the syntax, which I haven't heard any examples of yet). In general, I feel like a language/tool like bpftrace should be providing bits of sugar like this - feels similar to the forward slashes in front of probes to indicate a conditional.

danobi:
FWIW rust has implicit returns but a quick search on the internet reveals some level of unhappiness with that decision.

https://doc.rust-lang.org/book/ch03-03-how-functions-work.html#functions-with-return-values

amscanne:
Personally, I actually agree with the points @tyroguru and @viktormalik have expressed. I think it just adds to the list of mini-surprises where the language semantics depart from expectations. (In this regard, it's actually worse than the rust implicit return, since it's not "the last expression", it just "naked identifiers".)

BUT what's interesting to me is that it definitely does align with the language design doc which has "conciseness / one-liners" as the top priority and @jordalgo has a good argument. There's some obvious tension between those values, as "conciseness" is often the enemy of readability & maintainability (*cough* Perl *cough*). It makes sense that you want to support both interactive use cases and building tools, so perhaps a reframing there as *simple* and/or *intuitive* would be better words for what I think those values are trying to express? I think there's a good argument against this feature through a "simple" or "intuitive" lens, but an argument in favor through a pure "conciseness" one.

I did intentionally pick an issue from ~6 years ago (presuming that it wasn't an obvious "we're not doing this"), so it's really interesting to see this discussion. If it were up to me, I'd say that the right thing would be to sort out some of the directional/values questions, clean up the backlog/roadmap based on that, and the eventual fate of this particular pull request should be clear :)

viktormalik:
> As we've seen, one of bpftrace's largest use-cases (if not the largest) is one-liners - I would imagine this is true for awk (one of the inspiration languages). awk also has implicit printing (on top of an explicit `print` function) e.g. `awk 'NR % 2 == 0' data` prints the even numbered lines in `data`.

I also had AWK in mind but it's slightly different - it has a *default action block* `{ print $0 }` which is slightly different from what we do here and that is implicit wrapping of bare identifiers by a `print()` call. In this sense, it's more similar to Rust's (and also Nim's or JavaScript's) implicit returns but those are (at least to me) much more intuitive than what we try to do here. And they obviously still have problems as seen in @danobi's above comment.

> The reason I like automatic/implicit printing is because bpftrace's goal is to print things (rather than drive a backend service or UI application) so making that easier to do seems like a good thing (as long as it doesn't cause any future issues with the syntax, which I haven't heard any examples of yet). In general, I feel like a language/tool like bpftrace should be providing bits of sugar like this - feels similar to the forward slashes in front of probes to indicate a conditional.

I do agree here so I'm wondering if we could introduce a shorter way for printing things, e.g. a specific operator? For instance:
```
#kstack
```
could just print `kstack`. We don't use `#` for anything (except for preprocessor directives and `$#` for the number of positional arguments) and it doesn't have any "standard" meaning in programming so we could probably overload it for the purpose of printing (which is arguably bpftrace's most used action).

> BUT what's interesting to me is that it definitely does align with the language design doc which has "conciseness / one-liners" as the top priority and @jordalgo has a good argument. There's some obvious tension between those values, as "conciseness" is often the enemy of readability & maintainability (cough Perl cough). It makes sense that you want to support both interactive use cases and building tools, so perhaps a reframing there as simple and/or intuitive would be better words for what I think those values are trying to express? I think there's a good argument against this feature through a "simple" or "intuitive" lens, but an argument in favor through a pure "conciseness" one.
> 
> I did intentionally pick an issue from ~6 years ago (presuming that it wasn't an obvious "we're not doing this"), so it's really interesting to see this discussion. If it were up to me, I'd say that the right thing would be to sort out some of the directional/values questions, clean up the backlog/roadmap based on that, and the eventual fate of this particular pull request should be clear :)

These are very good points @amscanne, thanks for opening this discussion! I'd say that we should discuss on the office hours.

I agree that we should resolve this conflict between conciseness and intuitiveness in our language priorities. Perhaps it would just suffice to say that intuitiveness has priority over conciseness or that following the conciseness principle should not introduce counter-intuitive behaviour.

ajor:
I just want to point out that the language design doc doesn't dictate what decisions we have to make and we are free to ignore it or change it where it doesn't make sense. I'm not wholly convinced that it's possible to define an ordered list of priorities that we obey without exception.

Conciseness is good, but taken to the extreme, valuing it over all else results in languages like [GolfScript](https://www.golfscript.com/golfscript/examples.html) or [Uiui](https://tankorsmash.unison-services.cloud/s/uiuisms-service/), which obviously isn't what we want.

The majority of bpftrace users will be people who need it a few times a year. They won't put a lot of effort into learning the language and they won't remember most of what they do learn by the next time they need to use it. For this reason, I'd say that understandability should win out over conciseness here and in general (but there might well be other cases where prioritising readability above all else results in bloat that we won't want).

Also, as someone who writes quite a lot of throwaway bpftrace scripts, I've never felt annoyed by having to type out `print()`. In fact, I rarely use more than one `print` call in a script. I either pack multiple values into a single `print` call by using a tuple, or, when I'm really printing loads, I switch to `printf` to annotate my output.

jordalgo:
I'll wait for @danobi to chime in with any additional thoughts but I think this can be closed. The general consensus is that introducing this behavior is too confusing and that no one is currently asking for it. I will shed a single tear for bare identifier printing but no more than that 😢

amscanne:
> I'll wait for @danobi to chime in with any additional thoughts but I think this can be closed.

Just as a reminder, this is all a proxy discussion for #431, which this pull request is implementing. If there's a conclusion, I'd suggest that the issue should be closed first. The reason I suggested taking a pass at principles/having a rough roadmap is that there seem to be quite a few speculative issues that could similarly benefit from executive decisions (e.g. #404, #430, #1452 to pick a few at random from the last couple pages). In my mind at least, older issues tend to have an implicit directional acknowledgement (as they have been kept open for many years).

jordalgo:
> I'd suggest that the issue should be closed first. 

Agreed but let's wait a few days to do both.

> The reason I suggested taking a pass at principles/having a rough roadmap

We actually have a roadmap, which I'm sure needs updating, but I think we want to focus on the vision/strategy first, which was discussed in the last office hours. Here are links to the three planning docs:

[Project roadmap](https://docs.google.com/document/d/17729Rlyo1xzlJObzHpFLDzeCVgvwRh0ktAmMEJLK-EU/edit?usp=sharing)
[Language Design Document](https://docs.google.com/document/d/1a5vla623n0Z9oaUCNLaaDux6X6FO4qYfBs9zLJ6nzhA/edit?usp=sharing)
[Vision and Strategy](https://docs.google.com/document/d/1lMUnToY7eV5pJuqnWMTk1v6rF9XN6xS1-YbhM6J8abw/edit?usp=sharing) (In Review)

ajor:
> there seem to be quite a few speculative issues that could similarly benefit from executive decisions (e.g. #404, #430, #1452 to pick a few at random from the last couple pages). In my mind at least, older issues tend to have an implicit directional acknowledgement (as they have been kept open for many years).

Executive decisions made on these :) 

If anything, I'd say that the older issues have the least buy-in as they've not been acted on for years. If there's no discussion then they've likely just been missed amongst the masses.

danobi:
> I'll wait for @danobi to chime in with any additional thoughts but I think this can be closed. The general consensus is that introducing this behavior is too confusing and that no one is currently asking for it. I will shed a single tear for bare identifier printing but no more than that 😢

I think folks have reasonable points. I actually tend to do `print("XX")` a lot for bisecting new codepaths in the kernel, so I would've liked this feature. I've probably written too much rust

jordalgo:
Details above and here: https://github.com/bpftrace/bpftrace/issues/431

PR #3594, opened 2024-11-19
Fix crash using some per-cpu maps as map keys

jordalgo:
'count', 'sum', 'avg', 'min', 'max' are all castable per-cpu map aggregation types so they should be valid as map keys when printing.

Issue: https://github.com/bpftrace/bpftrace/issues/3591

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -452,6 +452,11 @@ std::string Output::map_key_str(BPFtrace &bpftrace,
     case Type::array:
     case Type::mac_address:
     case Type::record:
+    case Type::count_t:
+    case Type::avg_t:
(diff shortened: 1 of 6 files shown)


src/output.cpp
danobi, line 1: just double checking that I read the code right - `arg` will reflect the correct "collapsed" size of the integer, right? 8 bytes.

danobi (approved):
lgtm.

src/output.cpp
jordalgo: Yes, as `value_to_str` handles these types already.

danobi (approved):

PR #3595, opened 2024-11-19
Fix binop type check

jordalgo:
If either the left or the right is a none type
then wait till the final pass before throwing
a mismatch error.

Issue: https://github.com/bpftrace/bpftrace/issues/3592

##### Checklist

### DIFF
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -4265,6 +4265,11 @@ fentry:func_1,tracepoint:sched:sched_one { args }
 )");
 }
 
+TEST_F(semantic_analyser_btf, binop_late_ptr_resolution)
+{
+  test(R"(fentry:func_1 { if (@a[1] == args.foo1) { } @a[1] = args.foo1; })");
+}
+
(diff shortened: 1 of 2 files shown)


tests/semantic_analyser.cpp
```diff
@@ -2832,6 +2832,12 @@ TEST(semantic_analyser, binop_sign)
   }
 }
 
+TEST(semantic_analyser, binop_late_ptr_resolution)
+{
+  test("kfunc:mem_cgroup_charge_skmem { if (@a[1] == args.memcg) { print((1)); "
```
jordalgo, line 2837: not sure if we should use a real fentry function in this test but I'm not sure how else to test. Maybe with a uprobe and our own program?

viktormalik: Use `semantic_analyser_btf` fixture and test on functions from `tests/data/data_source.c`. Also prefer `fentry` over `kfunc`.

jordalgo: Thanks @viktormalik - that works.

danobi (approved):

viktormalik (approved):

PR #3596, opened 2024-11-19
Introduce builtin to access percpu kernel data

### LINKED ISSUE #1837: Support accessing percpu kernel data
There's a helper to access percpu data now: https://github.com/torvalds/linux/blob/0aa099a312b6323495a23d758009eb7fc04a7617/include/uapi/linux/bpf.h#L4489

However, it's not so simple to use. IIUC, there needs to be BTF ID to the data in question
in the right register when the helper is called. libbpf does these fixups and they look
somewhat complicated. So before we can support this we'll need to have libbpf do
our loading for us.

viktormalik:
Introduce new builtin `percpu_kaddr` to access kernel percpu variables. It leverages the `bpf_per_cpu_ptr` helper which accepts a pointer to a percpu ksym and a CPU number. The ksym is an extern variable with a BTF entry matching the BTF of the corresponding kernel variable. Since we now use libbpf to do the loading, it is sufficient to emit a global variable declaration with a proper BTF (debug info) and libbpf will take care of the rest.

The helper has two forms:

    percpu_kaddr("symbol_name")
    percpu_kaddr("symbol_name", N)

where N is the CPU number. The former variant retrieves the value for the current CPU.

For instance, this allows to access the `process_counts` percpu variable:
```
$ bpftrace -e 'BEGIN { printf("%d processes running on CPU %d\n", *percpu_kaddr("process_counts"), cpu); exit() }'
Attaching 1 probe...
88 processes running on CPU 1
```

 A tricky part is that `bpf_per_cpu_ptr` may return NULL if the supplied CPU number is higher than the number of the CPUs. The BPF program should perform a NULL-check on the returned value, otherwise it is rejected by the verifier. In practice, this only happens if pointer arithmetics is used (i.e. a struct field is accessed). Since it is quite complex to detect a missing NULL-check in bpftrace, we instead let verifier do it and just display the potential verifier error in a nicer manner:

```
$ bpftrace -e 'BEGIN { printf("%d\n", ((struct rq *)percpu_kaddr("runqueues"))->nr_running); exit() }'
Attaching 1 probe...
stdin:1:24-64: ERROR: helper bpf_per_cpu_ptr: result needs to be null-checked before accessing fields
BEGIN { printf("%d\n", ((struct rq *)percpu_kaddr("runqueues"))->nr_running); exit() }
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Resolves #1837.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1364,6 +1364,10 @@ Tracing block I/O sizes > 0 bytes
 | Return full path
 | Sync
 
+| <<functions-percpu-kaddr, `percpu_kaddr(const string name [, int cpu])`>>
+| Resolve percpu kernel symbol name
+| Sync
+
 | <<functions-print, `print(...)`>>
 | Print a non-map value with default formatting
 | Async
@@ -1845,6 +1849,39 @@ Return full path referenced by struct path pointer in argument. If `size` is set
 the path will be clamped by `size` otherwise `BPFTRACE_MAX_STRLEN` is used.
 
 This function can only be used by functions that are allowed to, these functions are contained in the `btf_allowlist_d_path` set in the kernel.
+
+[#functions-percpu-kaddr]
+=== percpu_kaddr
+
+.variants
+* `void *percpu_kaddr(const string name)`
+* `void *percpu_kaddr(const string name, int cpu)`
+
+*sync*
+
+Get the address of the percpu kernel symbol `name` for CPU `cpu`. When `cpu` is
+omitted, the current CPU is used.
+
+----
+interval:s:1 {
+  $proc_cnt = percpu_kaddr("process_counts");
+  printf("% processes are running on CPU %d\n", *$proc_cnt, cpu);
+}
+----
+
+The second variant may return NULL if `cpu` is higher than the number of
+available CPUs. Therefore, it is necessary to perform a NULL-check on the result
+when accessing fields of the pointed structure, otherwise the BPF program will
+be rejected.
+
+----
+interval:s:1 {
+  $runqueues = (struct rq *)percpu_kaddr("runqueues", 0);
+  if ($runqueues != 0) {         // The check is mandatory here
+    print($runqueues->nr_running);
+  }
+}
+----
 
 [#functions-print]
 === print
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -213,10 +213,12 @@ DIType *DIBuilderBPF::GetType(const SizedType &stype, bool emit_codegen_types)
 {
   if (!emit_codegen_types && stype.IsRecordTy()) {
     std::string name = stype.GetName();
-    if (name.find("struct ") == 0)
-      name = name.substr(std::string("struct ").length());
-    else if (name.find("union ") == 0)
-      name = name.substr(std::string("union ").length());
+    static constexpr std::string struct_prefix = "struct ";
+    static constexpr std::string union_prefix = "union ";
+    if (name.find(struct_prefix) == 0)
+      name = name.substr(struct_prefix.length());
+    else if (name.find(union_prefix) == 0)
+      name = name.substr(union_prefix.length());
 
     return createStructType(file,
                             name,
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -212,8 +212,14 @@ DIType *DIBuilderBPF::CreateByteArrayType(uint64_t num_bytes)
 DIType *DIBuilderBPF::GetType(const SizedType &stype, bool emit_codegen_types)
 {
   if (!emit_codegen_types && stype.IsRecordTy()) {
+    std::string name = stype.GetName();
+    if (name.find("struct ") == 0)
+      name = name.substr(std::string("struct ").length());
+    else if (name.find("union ") == 0)
+      name = name.substr(std::string("union ").length());
+
     return createStructType(file,
-                            stype.GetName(),
+                            name,
                             file,
                             0,
                             stype.GetSize() * 8,
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -347,7 +347,7 @@ DIGlobalVariableExpression *DIBuilderBPF::createGlobalVariable(
     const SizedType &stype)
 {
   return createGlobalVariableExpression(
-      file, name, "global", file, 0, GetType(stype), false);
+      file, name, "global", file, 0, GetType(stype, false), false);
 }
 
 } // namespace bpftrace::ast
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -2141,6 +2141,39 @@ CallInst *IRBuilderBPF::CreateGetFuncIp(Value *ctx, const location &loc)
                           &loc);
 }
 
+CallInst *IRBuilderBPF::CreatePerCpuPtr(Value *var,
+                                        Value *cpu,
+                                        const location &loc)
+{
+  // void *bpf_per_cpu_ptr(const void *percpu_ptr, u32 cpu)
+  // Return:
+  //    A pointer pointing to the kernel percpu variable on
+  //    cpu, or NULL, if cpu is invalid.
+  FunctionType *percpuptr_func_type = FunctionType::get(
+      GET_PTR_TY(), { GET_PTR_TY(), getInt64Ty() }, false);
+  return CreateHelperCall(libbpf::BPF_FUNC_per_cpu_ptr,
+                          percpuptr_func_type,
+                          { var, cpu },
+                          "per_cpu_ptr",
+                          &loc);
+}
+
+CallInst *IRBuilderBPF::CreateThisCpuPtr(Value *var, const location &loc)
+{
+  // void *bpf_per_cpu_ptr(const void *percpu_ptr)
+  // Return:
+  //    A pointer pointing to the kernel percpu variable on
+  //    this cpu. May never be NULL.
+  FunctionType *percpuptr_func_type = FunctionType::get(GET_PTR_TY(),
+                                                        { GET_PTR_TY() },
+                                                        false);
+  return CreateHelperCall(libbpf::BPF_FUNC_this_cpu_ptr,
+                          percpuptr_func_type,
+                          { var },
+                          "this_cpu_ptr",
+                          &loc);
+}
+
 void IRBuilderBPF::CreateGetCurrentComm(Value *ctx,
                                         AllocaInst *buf,
                                         size_t size,
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -927,6 +927,18 @@ void CodegenLLVM::visit(Call &call)
     if (!addr)
       throw FatalUserException("Failed to resolve kernel symbol: " + name);
     expr_ = b_.getInt64(addr);
+  } else if (call.func == "percpu_kaddr") {
+    auto name = bpftrace_.get_string_literal(call.vargs.at(0));
+    auto var = b_.CreatePointerCast(DeclareKernelVar(name), b_.GET_PTR_TY());
+    Value *percpu_ptr;
+    if (call.vargs.size() == 1) {
+      percpu_ptr = b_.CreateThisCpuPtr(var, call.loc);
+    } else {
+      auto scoped_del = accept(call.vargs.at(1));
+      Value *cpu = expr_;
+      percpu_ptr = b_.CreatePerCpuPtr(var, cpu, call.loc);
+    }
+    expr_ = b_.CreatePtrToInt(percpu_ptr, b_.getInt64Ty());
   } else if (call.func == "uaddr") {
     auto name = bpftrace_.get_string_literal(call.vargs.at(0));
     struct symbol sym = {};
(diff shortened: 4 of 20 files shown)


ajor:
Since I don't generally like adding new builtins, I've got to ask: what are the benefits of this over reading from the kernel PERCPU data manually, a la: https://github.com/hodgesds/scx/blob/912d6e01c1538cb5ec7e6363332caef66f7765ba/scheds/rust/scx_layered/integration/layer_llc.bt#L22-L27

No bpftrace changes requires and with user defined functions, the above approach could be wrapped in a simple stdlib function to make it easier to use.

viktormalik:
> Since I don't generally like adding new builtins, I've got to ask: what are the benefits of this over reading from the kernel PERCPU data manually, a la: https://github.com/sched-ext/scx/blob/912d6e01c1538cb5ec7e6363332caef66f7765ba/scheds/rust/scx_layered/integration/layer_llc.bt#L22-L27
> 
> No bpftrace changes requires and with user defined functions, the above approach could be wrapped in a simple stdlib function to make it easier to use.

Mostly convenience, at the moment. It will take a while until we get user-defined functions and a standard library and writing this manually is much more code.

On top of that, we get some verifier checks for free, e.g. the mentioned check that users cannot do pointer arithmetic on the returned value without a prior null-check. For stdlib bpftrace function, we'd have to define those checks on our own.

That being said, I understand you reluctance to add new helpers but I feel like the standard library is still quite some time away and, in the meantime, we'll need to address some real-world use-cases which require new builtins. I'm wondering if we could establish a plan that once the standard library is in place, we will remove some (potentially many) builtin functions in favor of stdlib functions? That would allow us to add builtins for the time being but we would also be transparent about the fact that they will go away from the language eventually.

jordalgo:
> I'm wondering if we could establish a plan that once the standard library is in place, we will remove some (potentially many) builtin functions in favor of stdlib functions? 

Is it possible for this to be seamless for users? Maybe we could detect what "builtin" functions are being used and automatically include them from the standard library (if there are no conflicts with other user-defined functions)

viktormalik:
> > I'm wondering if we could establish a plan that once the standard library is in place, we will remove some (potentially many) builtin functions in favor of stdlib functions?
> 
> Is it possible for this to be seamless for users? Maybe we could detect what "builtin" functions are being used and automatically include them from the standard library (if there are no conflicts with other user-defined functions)

Yeah, something like that. It very much depends on the include model that we end up with but I'm sure we'll be able to find a solution which is transparent for users.

src/ast/irbuilderbpf.cpp
```diff
@@ -2141,6 +2141,23 @@ CallInst *IRBuilderBPF::CreateGetFuncIp(Value *ctx, const location &loc)
...
+  //    cpu, or NULL, if cpu is invalid.
+  FunctionType *percpuptr_func_type = FunctionType::get(
+      GET_PTR_TY(), { GET_PTR_TY(), getInt64Ty() }, false);
+  return CreateHelperCall(libbpf::BPF_FUNC_per_cpu_ptr,
+                          percpuptr_func_type,
+                          { var, cpu },
+                          "per_cpu_ptr",
+                          &loc);
+}
+
```
danobi, line 2160: Looks like the `bpf_this_cpu_ptr()` variant can never return null according to docs. Given that user needs to do null checks if accessing fields, probably worthwhile to use the `this` variant when possible. So less null checks necessary. WDYT?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4697,4 +4710,37 @@ CallInst *CodegenLLVM::CreateKernelFuncCall(Kfunc kfunc,
...
+  if (auto *sym = module_->getGlobalVariable(var_name))
+    return sym;
+
+  std::string err;
+  auto type = bpftrace_.btf_->get_var_type(var_name);
+  if (type.IsNoneTy()) {
+    throw FatalUserException("Could not resolve variable \"" + var_name +
+                             "\" from BTF");
+  }
+
```
danobi, line 4734: Can this be done in semantic analyser so there's highlighting?

tests/runtime/call
danobi, line 1: test missing symbol also?

src/ast/irbuilderbpf.cpp
viktormalik: Yeah, absolutely, I missed that one.

src/ast/passes/codegen_llvm.cpp
viktormalik: Right, makes sense.

man/adoc/bpftrace.adoc
```diff
@@ -1846,6 +1850,39 @@ the path will be clamped by `size` otherwise `BPFTRACE_MAX_STRLEN` is used.
...
+* `void *percpu_kaddr(const string name, int cpu)`
+
+*sync*
+
+Get the address of the percpu kernel symbol `name` for CPU `cpu`. When `cpu` is
+omitted, the current CPU is used.
+
+----
+interval:s:1 {
+  $proc_cnt = kaddr("process_counts");
```
jordalgo, line 1867:
```suggestion
  $proc_cnt = percpu_kaddr("process_counts");
```

man/adoc/bpftrace.adoc
```diff
@@ -1846,6 +1850,39 @@ the path will be clamped by `size` otherwise `BPFTRACE_MAX_STRLEN` is used.
...
+
+The function may return NULL if `cpu` is higher than the number of available
+CPUs. Therefore, it is necessary to perform a NULL-check on the result when
+accessing fields of the pointed structure, otherwise the BPF program will be
+rejected.
+
+----
+interval:s:1 {
+  $runqueues = (struct rq *)percpu_kaddr("runqueues");
+  if ($runqueues != 0) {         # The check is mandatory here
```
jordalgo, line 1880: Should we add a NULL type ? 🤔

man/adoc/bpftrace.adoc
```diff
@@ -1846,6 +1850,39 @@ the path will be clamped by `size` otherwise `BPFTRACE_MAX_STRLEN` is used.
...
+interval:s:1 {
+  $proc_cnt = kaddr("process_counts");
+  printf("% processes are running on CPU %d\n", *$proc_cnt, cpu);
+}
+----
+
+The function may return NULL if `cpu` is higher than the number of available
+CPUs. Therefore, it is necessary to perform a NULL-check on the result when
+accessing fields of the pointed structure, otherwise the BPF program will be
+rejected.
```
jordalgo, line 1875: What happens if the kernel symbol is invalid? e.g. `percpu_kaddr("potato")`

src/ast/irbuilderbpf.cpp
jordalgo: I wonder if we can bake this into a semantic_analyser check. It's a bit confusing to users that they have to null check one variant and not the other.

man/adoc/bpftrace.adoc
viktormalik: Then we fail in the semantic analyser, see one of the unit tests.

tests/codegen/llvm/call_len_map_sum_elem_count.ll
```diff
@@ -117,5 +117,5 @@ attributes #1 = { nocallback nofree nosync nounwind willreturn memory(argmem: re
 !61 = !DISubroutineType(types: !62)
 !62 = !{!18, !63}
 !63 = !DIDerivedType(tag: DW_TAG_pointer_type, baseType: !64, size: 64)
-!64 = !DICompositeType(tag: DW_TAG_structure_type, name: "struct bpf_map", scope: !2, file: !2, elements: !65)
+!64 = !DICompositeType(tag: DW_TAG_structure_type, name: "bpf_map", scope: !2, file: !2, elements: !65)
```
jordalgo, line 120: this a safe change?

src/ast/dibuilderbpf.cpp
```diff
@@ -212,8 +212,14 @@ DIType *DIBuilderBPF::CreateByteArrayType(uint64_t num_bytes)
 DIType *DIBuilderBPF::GetType(const SizedType &stype, bool emit_codegen_types)
 {
   if (!emit_codegen_types && stype.IsRecordTy()) {
+    std::string name = stype.GetName();
+    if (name.find("struct ") == 0)
+      name = name.substr(std::string("struct ").length());
```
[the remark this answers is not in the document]
jordalgo: Yeah. I think const static local variables would be fine. Save on additional string creation every time we hit this function.

src/ast/irbuilderbpf.cpp
viktormalik: It's not easy, I think. First, I wouldn't put it to semantic analyser but rather to a separate pass. And there, you'd have to introduce a context-sensitive analysis - you'd have to check that all field accesses of the pointer returned by the helper are done within a context where the pointer is not equal to NULL. I can't see a simple way of doing that.

Alternatively, we could inject the check ourselves into the AST or in codegen. But then, what would be the value for the other branch? Zero? That's a valid value for some fields so we would basically give users incorrect values.

The good thing here is that bpftrace will immediately inform you about the missing check so users don't need to remember anything, just add the check when the tool requires it.

man/adoc/bpftrace.adoc
jordalgo: What if NULL was an alias for `0`?

viktormalik: That would be probably reasonable. I would do it in a separate PR, though.

tests/codegen/llvm/call_len_map_sum_elem_count.ll
viktormalik: Interestingly yes, it looks like the verifier doesn't care about the name for function parameters as much as for global variables. In any case, now the name matches the name in kernel BTF.

man/adoc/bpftrace.adoc
viktormalik: Ouch, good catch.

src/ast/dibuilderbpf.cpp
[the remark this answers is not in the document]
viktormalik: Ok, will fix.

src/ast/irbuilderbpf.cpp
danobi: This seems like something we can improve on in the verifier. If the CPU number is statically known and in the range of valid CPUs, verifier should be able to let user omit null check.

man/adoc/bpftrace.adoc
```diff
@@ -1846,6 +1850,39 @@ the path will be clamped by `size` otherwise `BPFTRACE_MAX_STRLEN` is used.
...
+
+The second variant may return NULL if `cpu` is higher than the number of
+available CPUs. Therefore, it is necessary to perform a NULL-check on the result
+when accessing fields of the pointed structure, otherwise the BPF program will
+be rejected.
+
+----
+interval:s:1 {
+  $runqueues = (struct rq *)percpu_kaddr("runqueues");
+  if ($runqueues != 0) {         # The check is mandatory here
```
danobi, line 1880: Comment in bpftrace is `//` or `/* */`, not `#`, right?

src/ast/irbuilderbpf.cpp
```diff
@@ -2141,6 +2141,39 @@ CallInst *IRBuilderBPF::CreateGetFuncIp(Value *ctx, const location &loc)
...
+CallInst *IRBuilderBPF::CreateThisCpuPtr(Value *var, const location &loc)
+{
+  // void *bpf_per_cpu_ptr(const void *percpu_ptr)
+  // Return:
+  //    A pointer pointing to the kernel percpu variable on
+  //    this cpu.
```
danobi, line 2166: Maybe worth calling out "can never be NULL" here

src/ast/passes/codegen_llvm.cpp
```diff
@@ -927,6 +927,18 @@ void CodegenLLVM::visit(Call &call)
...
+    auto var = b_.CreatePointerCast(DeclareKernelVar(name), b_.GET_PTR_TY());
+    Value *percpu_ptr;
+    if (call.vargs.size() == 1) {
+      percpu_ptr = b_.CreateThisCpuPtr(var, call.loc);
+    } else {
+      auto scoped_del = accept(call.vargs.at(1));
+      Value *cpu = expr_;
+      percpu_ptr = b_.CreatePerCpuPtr(var, cpu, call.loc);
+    }
+    expr_ = b_.CreatePtrToInt(percpu_ptr, b_.getInt64Ty());
```
danobi, line 941: Is pointer to u64 right here? semantic analyser seems to mark the return type as u64

tests/semantic_analyser.cpp
```diff
@@ -1366,6 +1366,24 @@ TEST(semantic_analyser, call_kaddr)
   test("kprobe:f { kaddr(123); }", 1);
 }
 
+TEST(semantic_analyser, call_percpu_kaddr)
+{
+  test("kprobe:f { percpu_kaddr(\"process_counts\"); }");
```
danobi, line 1371: is `process_counts` from host? I don't see it in btf data source

man/adoc/bpftrace.adoc
viktormalik: Yes, of course :man_facepalming:

tests/semantic_analyser.cpp
viktormalik: Yes, it's from the host. It exists on every Linux system but it's still probably better to have it in data_source.

src/ast/passes/codegen_llvm.cpp
viktormalik: It creates a cast from a ptr to an integer, not to a "pointer to integer".

viktormalik:
All comments should be addressed. Adding `process_counts` to `data_source` was quite a fight as pahole puts only percpu variables into BTF. So, I had to put the variable to the `.data..percpu` section and do some tweaks to building BTF and DWARF data in CMake (BTF needs relocatable file to preserve the custom section while DWARF needs executable file to be able to strip DWARF data into a separate file).

jordalgo (approved):
LGTM. Nice work!

danobi (approved):
yay, great to finally have this!

PR #3600, opened 2024-11-21
Automatically cast map function when assigning to int map

ttreyer:
Assigning a map containing a castable map function (like `count()`) to another map containing integers is allowed, but requires a cast. However, the cast is not required for other operations, like comparison. This is inconsistent and particularly confusing for new users.

This PR removes this requirement for assignments.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -127,11 +127,23 @@ static bool IsValidVarDeclType(const SizedType &ty)
 // These are special map aggregation types that cannot be assigned
 // to scratch variables or from one map to another e.g. these are both invalid:
 // `@a = hist(10); let $b = @a;`
-// `@a = hist(10); @b = @a;`
-static bool IsValidAssignment(const SizedType &ty, bool is_map)
+// `@a = count(); @b = @a;`
+// However, if the assigned map already contains integers, we implicitly cast
+// the aggregation into an integer to retrieve its value, so this is valid:
+// `@a = count(); @b = 0; @b = @a`
+bool SemanticAnalyser::is_valid_assignment(const Expression *target,
+                                           const Expression *expr)
 {
-  if (is_map && (ty.IsMultiOutputMapTy())) {
-    return false;
+  if (expr->is_map) {
+    // Prevent assigning multi-output aggregations to another map.
+    if (expr->type.IsMultiOutputMapTy())
+      return false;
+    // Prevent declaring a map copying another aggregate map.
+    if (auto *target_map = dynamic_cast<const Map *>(target)) {
+      bool map_has_type = get_map_type(*target_map);
+      if (expr->type.IsCastableMapTy() && map_has_type == false)
+        return false;
+    }
   }
   return true;
 }
@@ -2824,35 +2836,61 @@ void SemanticAnalyser::visit(ExprStatement &expr)
   Visit(expr.expr);
 }
 
+static const std::unordered_map<Type, std::string_view> AGGREGATE_HINTS{
+  { Type::count_t, "count()" },
+  { Type::sum_t, "sum(retval)" },
+  { Type::min_t, "min(retval)" },
+  { Type::max_t, "max(retval)" },
+  { Type::avg_t, "avg(retval)" },
+  { Type::hist_t, "hist(retval)" },
+  { Type::lhist_t, "lhist(rand %10, 0, 10, 1)" },
+  { Type::stats_t, "stats(arg2)" },
+};
+
 void SemanticAnalyser::visit(AssignMapStatement &assignment)
 {
   Visit(assignment.map);
   Visit(assignment.expr);
 
-  auto type = assignment.expr->type;
-
-  if (!IsValidAssignment(type, assignment.expr->is_map)) {
-    auto hint = "";
-    if (type.IsHistTy()) {
-      hint = "hist(retval)";
-    } else if (type.IsLhistTy()) {
-      hint = "lhist(rand %10, 0, 10, 1)";
-    } else if (type.IsStatsTy()) {
-      hint = "stats(arg2)";
-    } else {
+  const auto *map_type_before = get_map_type(*assignment.map);
+  if (!is_valid_assignment(assignment.map, assignment.expr)) {
+    const auto &type = assignment.expr->type;
+    auto hint = AGGREGATE_HINTS.find(type.GetTy());
+    if (hint == AGGREGATE_HINTS.end())
       LOG(BUG) << "Missing assignment hint";
-    }
+
     LOG(ERROR, assignment.loc, err_)
         << "Map value '" << type
-        << "' cannot be assigned from one map to another. The function that "
-           "returns this "
-           "type must be called directly e.g. `@a = "
-        << hint << ";`.";
+        << "' cannot be assigned from one map to another. "
+           "The function that returns this type must be called directly e.g. `"
+        << assignment.map->ident << " = " << hint->second << ";`.";
+
+    if (auto *expr_map = dynamic_cast<const Map *>(assignment.expr)) {
+      if (type.IsCastableMapTy()) {
+        LOG(HINT, err_)
+            << "Add a cast to integer if you want the value of the aggregate, "
+               "e.g. `"
+            << assignment.map->ident << " = (int64)" << expr_map->ident
+            << ";`.";
+      }
+    }
+  }
+
+  // Add an implicit cast when copying the value of an aggregate map to an
+  // existing map of int. Enables the following: `@x = 1; @y = count(); @x = @y`
+  const bool map_contains_int = map_type_before && map_type_before->IsIntTy();
+  const bool expr_is_map_with_castable_agg =
+      assignment.expr->is_map && assignment.expr->type.IsCastableMapTy();
+  if (map_contains_int && expr_is_map_with_castable_agg) {
+    assignment.expr = ctx_.make_node<Cast>(*map_type_before,
+                                           assignment.expr,
+                                           assignment.expr->loc);
   }
 
   assign_map_type(*assignment.map, assignment.expr->type);
 
-  const std::string &map_ident = assignment.map->ident;
+  const auto &map_ident = assignment.map->ident;
+  const auto &type = assignment.expr->type;
 
   if (type.IsRecordTy() && map_val_[map_ident].IsRecordTy()) {
     std::string ty = assignment.expr->type.GetName();
@@ -2928,16 +2966,23 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
     Visit(assignment.var_decl_stmt);
   }
 
-  std::string var_ident = assignment.var->ident;
-  auto &assignTy = assignment.expr->type;
-
-  if (!IsValidAssignment(assignTy, assignment.expr->is_map)) {
+  if (!is_valid_assignment(assignment.var, assignment.expr)) {
     LOG(ERROR, assignment.loc, err_)
-        << "Map value '" << assignTy
+        << "Map value '" << assignment.expr->type
         << "' cannot be assigned to a scratch variable.";
   }
 
+  const bool expr_is_map_with_castable_agg =
+      assignment.expr->is_map && assignment.expr->type.IsCastableMapTy();
+  if (expr_is_map_with_castable_agg) {
+    assignment.expr = ctx_.make_node<Cast>(CreateInt64(),
+                                           assignment.expr,
+                                           assignment.expr->loc);
+  }
+
   Node *var_scope = nullptr;
+  const auto &var_ident = assignment.var->ident;
+  const auto &assignTy = assignment.expr->type;
 
   if (auto *scope = find_variable_scope(var_ident)) {
     auto &foundVar = variables_[scope][var_ident];
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -2349,6 +2349,80 @@ TEST(semantic_analyser, map_cast_types)
 
 TEST(semantic_analyser, map_aggregations_implicit_cast)
 {
+  // When assigning an aggregation to a map containing integers,
+  // the aggregation is implicitly cast to an integer.
+  test("kprobe:f { @x = 1; @y = count(); @x = @y; }", R"(*
+  =
+   map: @x :: [int64]
+   (int64)
+    map: @y :: [count_t]
+*)");
+  test("kprobe:f { @x = 1; @y = sum(5); @x = @y; }", R"(*
+  =
+   map: @x :: [int64]
+   (int64)
+    map: @y :: [sum_t]
+*)");
+  test("kprobe:f { @x = 1; @y = min(5); @x = @y; }", R"(*
+  =
+   map: @x :: [int64]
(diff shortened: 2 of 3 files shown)


tests/semantic_analyser.cpp
```diff
@@ -533,8 +533,9 @@ kprobe:f { @x = pid < 10000 ? "lo" : 2 }
 
 TEST(semantic_analyser, mismatched_call_types)
 {
+  // This assignment requires a cast, hint the requirement in the error message
   test_error("kprobe:f { @x = 1; @x = count(); }", R"(
-stdin:1:20-22: ERROR: Type mismatch for @x: trying to assign value of type 'count_t' when map already contains a value of type 'int64'
+stdin:1:20-22: ERROR: Type mismatch for @x: trying to assign value of type 'count_t' when map already contains a value of type 'int64', cast the 'count_t' value to 'int64'
 kprobe:f { @x = 1; @x = count(); }
```
jordalgo, line 539: hmm, I'm not sure this is valid even with the cast. Doesn't the count operation need to be linked to a per_cpu map - meaning it doesn't really live as an isolated function. The map in this case is already a hash map.

ttreyer: Very good point, we don't want to confuse our users with an incorrect hint!

jordalgo (approved):
LGTM!

tests/semantic_analyser.cpp
```diff
@@ -533,6 +533,14 @@ kprobe:f { @x = pid < 10000 ? "lo" : 2 }
 
 TEST(semantic_analyser, mismatched_call_types)
 {
+  // This assignment requires a cast, hint the requirement in the error message
+  test_error("kprobe:f { @x = 1; @y = count(); @x = @y; }", R"(
+stdin:1:34-36: ERROR: Type mismatch for @x: trying to assign value of type 'count_t' when map already contains a value of type 'int64', cast the 'count_t' value to 'int64'
```
danobi, line 538: Bit of a long sentence. How about this instead?

```suggestion
stdin:1:34-36: ERROR: Type mismatch for @x: trying to assign value of type 'count_t' when map already contains a value of type 'int64'. Maybe cast the 'count_t' value to 'int64'?
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3761,10 +3765,15 @@ void SemanticAnalyser::assign_map_type(const Map &map, const SizedType &type)
       else
         *maptype = type;
     } else if (maptype->GetTy() != type.GetTy()) {
+      std::stringstream hint;
+      if (maptype->IsIntegerTy() && type.IsCountTy() && expr.is_map)
```
danobi, line 3769: More than just count() right?

```suggestion
      if (maptype->IsIntegerTy() && type.IsCastableMapTy() && expr.is_map)
```

jordalgo: Good call.

jordalgo:
@ttreyer This got me thinking. Maybe what we need to do is remove the semantic_analyser error instead of hinting. These aggregation types are treated as int64 types and should be automatically compatible without the cast.

ttreyer:
> @ttreyer This got me thinking. Maybe what we need to do is remove the semantic_analyser error instead of hinting. These aggregation types are treated as int64 types and should be automatically compatible without the cast.

That would make a lot of sense, as this is already the case for `Binopt`.
One reason to keep it around is to highlight the cost of reading the per-CPU map. What do you think?

jordalgo:
> One reason to keep it around is to highlight the cost of reading the per-CPU map. What do you think?

At this point, it feels like a bug since other usages of these map aggregation types don't require a cast e.g.
https://github.com/bpftrace/bpftrace/blob/master/tests/semantic_analyser.cpp#L2565
https://github.com/bpftrace/bpftrace/blob/master/tests/semantic_analyser.cpp#L2347

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2819,7 +2819,15 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
 
   const std::string &map_ident = assignment.map->ident;
 
-  if (type.IsRecordTy() && map_val_[map_ident].IsRecordTy()) {
+  if (type.IsCastableMapTy() && map_val_[map_ident].IsIntTy()) {
+    if (auto *call_expr = dynamic_cast<Call *>(assignment.expr)) {
```
jordalgo, line 2823: Can we add a semantic analyser test that triggers this error?

ttreyer: This is covered by `semantic_analyser.mismatched_call_types`.

jordalgo, line 2823: nit: to remove the unused variable warning
```suggestion
    if (dynamic_cast<Call *>(assignment.expr)) {
```

ajor:
It's not clear to me what the intended behaviour should be after this change.

```
@a = count();
@b = @a;
```
I would expect `@b` hold a `count()` map, but from the comments I'm guessing that it holds an integer instead? I'm not sure I agree that this should be the behaviour, if that is the case.

Also, it'd be good to have the semantic analyser tests validate the AST instead of just checking that there are no errors. That would document what the intended behaviour is and also lock it in so it can't get silently changed in the future.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2819,7 +2819,15 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
 
   const std::string &map_ident = assignment.map->ident;
 
-  if (type.IsRecordTy() && map_val_[map_ident].IsRecordTy()) {
+  if (type.IsCastableMapTy() && map_val_[map_ident].IsIntTy()) {
+    if (dynamic_cast<Call *>(assignment.expr)) {
+      LOG(ERROR, assignment.map->loc, err_)
```
ajor, line 2824: Unit tests for the error case would also be good

jordalgo: @ajor See my resolved comment above. These are already covered by a semantic analyser test.

ajor: Got confused by the error message staying the same!

jordalgo:
@ajor Actually, I think I was confused by what was happening. I thought `@b` ended up as a hashmap storing an int64 but you're right, it operates like a `count_t`, which is broken for other castable aggregation types (e.g. `min`, `max`, `avg`). These do special things when these functions are called and you can't simply assign values to the maps that contain these types e.g.

```
@a = avg(rand % 100);
@b = @a; // this is equivalent to @b = some_int, which actually causes a crash.
```

I also feel like it only happens to work for `sum` and `count` because they're sums because what is actually happening is that the value of `@a` is aggregated across cpus and then assigned as the value of the current cpu for `@b`. This feels funky bad and I think the type of `@b` SHOULD be an integer in this case:

```
@a = count();
@b = @a;
```

ajor (changes requested):
Requesting follow up as it turns out the behaviour change isn't completely straight-forward:
- add a changelog entry to explain the behaviour change
- extend semantic analysis unit tests to include ASTs to verify what's happening

@jordalgo: Yep, looks like you're right - after playing around some more there do seem to be loads of issues with assigning aggregation maps to other aggregation maps (although we disagree on the solution ;) ). I've opened an issue for it: #3620.

From what I understand, this PR is just for assigning an aggregation map to an existing integer map, so this discussion (that I started) isn't directly related. Testing the PR out myself, it does seem to give correct results for its intended use cases.

jordalgo:
@ajor and I chatted offline and thought that it's best if we just disallow this behavior:
```
@a = count();
@b = @a;
```
It's too confusing as to what should happen. Though I know @ttreyer is arguing, in the commit message, that it's confusing that other assignments of map aggregation types don't require a cast. It's also too difficult (at least in the short term) to copy the aggregation at the point of assignment. This however would be allowed as it's explicit what is happening.

```
@a = count();
@b = (int64)@a;
```
In this case `@b` is a hash map (non per-cpu) that contains an int64 value with the total count of `@a` at that moment of assignment.

@ttreyer - So sorry for the back and forth, how do you feel about this and perhaps updating the PR to disallow this kind of assignment?

ttreyer:
If I understand correctly, we should allow for assigning an aggregation map to an existing integer map:
```
@a = count();
@b = 0;
@b = @a; // Contains the integer value of the counter
```

But reject assigning aggregation map to a new map:
```
@a = count();
@b = @a; // Not allowed!
```

Let me know if that's correct, I'll update the PR accordingly and add the tests/CHANGELOG needed!

jordalgo:
> If I understand correctly, we should allow for assigning an aggregation map to an existing integer map:

This is a good question. I think if there is no ambiguity and the map already has an integer type as per your example then it should be allowed. We would also allow this (explicit casting):
```
@a = count();
@b = (int64)@a; // Allowed even though @b does not yet have a type.
```

tests/semantic_analyser.cpp
```diff
@@ -2349,6 +2349,80 @@ TEST(semantic_analyser, map_cast_types)
...
+  test("kprobe:f { @x = sum(5); @y = (uint64)@x; }");
+  test("kprobe:f { @x = min(5); @y = (uint64)@x; }");
+  test("kprobe:f { @x = max(5); @y = (uint64)@x; }");
+  test("kprobe:f { @x = avg(5); @y = (uint64)@x; }");
+
+  // However, if there is no explicit cast,
+  // the assignment is rejected and casting is suggested.
+  test_error("kprobe:f { @y = count(); @x = @y; }", R"(
+stdin:1:26-33: ERROR: Map value 'count_t' cannot be assigned from one map to another. The function that returns this type must be called directly e.g. `@x = count();`.
+Add a cast to integer if you want the value of the aggregate.
```
jordalgo, line 2397: I feel like this hint should go after the error, perhaps with a small example of the cast.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2824,35 +2834,59 @@ void SemanticAnalyser::visit(ExprStatement &expr)
...
-    auto hint = "";
-    if (type.IsHistTy()) {
-      hint = "hist(retval)";
-    } else if (type.IsLhistTy()) {
-      hint = "lhist(rand %10, 0, 10, 1)";
-    } else if (type.IsStatsTy()) {
-      hint = "stats(arg2)";
-    } else {
+  const auto *map_type = get_map_type(*assignment.map);
+  const bool is_map_declared = map_type != nullptr;
```
jordalgo, line 2854: maybe this is better or `map_has_type`?
```suggestion
  const bool is_map_assigned = map_type != nullptr;
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2824,35 +2834,59 @@ void SemanticAnalyser::visit(ExprStatement &expr)
...
-  if (!IsValidAssignment(type, assignment.expr->is_map)) {
-    auto hint = "";
-    if (type.IsHistTy()) {
-      hint = "hist(retval)";
-    } else if (type.IsLhistTy()) {
-      hint = "lhist(rand %10, 0, 10, 1)";
-    } else if (type.IsStatsTy()) {
-      hint = "stats(arg2)";
-    } else {
+  const auto *map_type = get_map_type(*assignment.map);
```
jordalgo, line 2853: Let's change this variable name to avoid the warnings below.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2824,35 +2834,59 @@ void SemanticAnalyser::visit(ExprStatement &expr)
...
+        << "' cannot be assigned from one map to another. "
+           "The function that returns this type must be called directly e.g. `"
+        << assignment.map->ident << " = " << hint_usage->second << ";`."
+        << hint_cast;
+  }
+
+  // Add an implicit cast when copying the value of an aggregate map to an
+  // existing map of int. Enables the following: `@x = 1; @y = count(); @x = @y`
+  const bool map_contains_int = map_type != nullptr && map_type->IsIntTy();
+  const bool expr_is_map_with_castable_agg =
```
jordalgo, line 2879: Do we need `assignment.expr->is_map` ?

ttreyer: Yes, otherwise it would allow the following:
```
@x = 0;         // Init @x with 0, so map_contains_int == true
@x = count();   // expr.type.IsCastableMapTy() == true, but expr.is_map == false
```

tests/semantic_analyser.cpp
ttreyer: Is that something we can do with our current `LOG` facilities?

jordalgo: Try `LOG(HINT, cast.loc, err_) `

tests/semantic_analyser.cpp
```diff
@@ -2349,6 +2349,90 @@ TEST(semantic_analyser, map_cast_types)
...
+  test("kprobe:f { @x = avg(5); @y = (uint64)@x; }");
+
+  // However, if there is no explicit cast,
+  // the assignment is rejected and casting is suggested.
+  test_error("kprobe:f { @y = count(); @x = @y; }", R"(
+stdin:1:26-33: ERROR: Map value 'count_t' cannot be assigned from one map to another. The function that returns this type must be called directly e.g. `@x = count();`.
+kprobe:f { @y = count(); @x = @y; }
+                         ~~~~~~~
+stdin:1:26-33: HINT: Add a cast to integer if you want the value of the aggregate, e.g. `@x = (uint64)@y;`.
+kprobe:f { @y = count(); @x = @y; }
```
jordalgo, line 2400: We probably don't need to repeat the offending line if it's in the error above but I don't see a way of omitting the line in `LOG(HINT)`, I'll fix that later.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2824,35 +2836,61 @@ void SemanticAnalyser::visit(ExprStatement &expr)
...
+        << "' cannot be assigned from one map to another. "
+           "The function that returns this type must be called directly e.g. `"
+        << assignment.map->ident << " = " << hint->second << ";`.";
+
+    if (auto *expr_map = dynamic_cast<const Map *>(assignment.expr)) {
+      if (type.IsCastableMapTy()) {
+        LOG(HINT, assignment.loc, err_)
+            << "Add a cast to integer if you want the value of the aggregate, "
+               "e.g. `"
+            << assignment.map->ident << " = (uint64)" << expr_map->ident
```
jordalgo, line 2873: super nit
```suggestion
            << assignment.map->ident << " = (int64)" << expr_map->ident
```

jordalgo (approved):

jordalgo:
@ajor Want to take a look before your leave? If not, I'll bypass branch protections muahahhaha

tests/semantic_analyser.cpp
jordalgo: My bad, looks like you can just do `LOG(HINT, out_)` without the loc (let's do that)

jordalgo:
@ttreyer Changes look good, just have to update the semantic analyser tests.

ttreyer:
> @ttreyer Changes look good, just have to update the semantic analyser tests.

Ah! Silly me...

PR #3602, opened 2024-11-22
CMake: fix (re)generation of BTF and DWARF data

viktormalik:
When `data_source.c` is changed, BTF and DWARF data (`btf_data.h` and `dwarf_data.h`) is not regenerated by CMake.

The reason for this is that `DEPENDS data_source_o` doesn't add a file-level dependency (only target ordering dependency) b/c `data_source_o` is not a file but just a custom target. That means that even if `data_source.o` is rebuilt by running the custom command, the `data_source_o` target stays up-to-date and doesn't make targets depending on it be rebuilt.

At the same time, we don't want to use just `data_source.o` as the dependency of the following commands as it will rebuild `data_source.o` multiple times (once for each depending target), due to parallelization. This was fixed in 31bdc1a2.

The solution here is to use **both** `data_source_o` and `data_source.o` (stored in the `${DATA_SOURCE_O}` variable) dependencies of the following custom commands. The `data_source_o` dependency will prevent `data_source.o` from being built multiple times by injecting a target-level ordering and the `${DATA_SOURCE_O}` dependency will make sure that targets are rebuilt whenever `data_source.o` (and transitively `data_source.c`) changes.

The exact same situation is discussed in [1].

[1] https://discourse.cmake.org/t/add-custom-command-with-target-dependency-doesnt-rebuild/7029/2

##### Checklist

### DIFF
--- a/tests/data/CMakeLists.txt
+++ b/tests/data/CMakeLists.txt
@@ -21,14 +21,26 @@ add_custom_command(
   # We must hack it like this b/c cmake does not support setting env vars at build time
   COMMAND bash -c "LLVM_OBJCOPY=${LLVM_OBJCOPY} pahole -J ${DATA_SOURCE_O}"
   DEPENDS ${DATA_SOURCE_C})
+
+# We don't want to use just ${DATA_SOURCE_O} as the dependency of the below
+# commands as that would make data_source.o be regenerated for every command
+# which is not only inefficient but also prone to race conditions.
+# So, we introduce a custom target data_source_o but unfortunately, it is not
+# sufficient to use solely that either as it just creates a target-ordering
+# dependency and not a file dependency, which causes the below commands not be
+# rerun when data_source.o changes.
+# The solution here is to use **both** data_source_o (to ensure correct target
+# ordering) and ${DATA_SOURCE_O} (to create file dependency) targets. We tie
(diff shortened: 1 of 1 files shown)


tests/data/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ set(BTF_DATA_FILE ${CMAKE_CURRENT_BINARY_DIR}/btf_data)
 add_custom_command(
   OUTPUT ${BTF_DATA_FILE}
   COMMAND ${LLVM_OBJCOPY} --dump-section .BTF=${BTF_DATA_FILE} ${DATA_SOURCE_O}
-  DEPENDS data_source_o)
+  DEPENDS data_source_o ${DATA_SOURCE_O})
```
danobi, line 30: Is it possible to put both these values into a variable? 1) to reduce duplication. 2) so you can add a comment describing the issue it's fixing. 

great find for something extremely non-obvious. I recall I tried to fix this when @ryantimwilson hit this a while ago. Didn't manage to find it.

viktormalik: Yeah, good idea, I added a new variable.

I'm still amazed that this behaviour is actually *by design* of CMake.

danobi (approved):

PR #3604, opened 2024-11-25
dibuilder: Use string_view for constexpr string

danobi:
std::string is not very const. The build fails with the following on clang17:

```
src/ast/dibuilderbpf.cpp:216:34: error: constexpr variable cannot have non-literal type 'const std::string' (aka 'const basic_string<char>')
  216 |     static constexpr std::string struct_prefix = "struct ";
      |                                  ^
libgcc/include/c++/trunk/bits/basic_string.h:85:11: note: 'basic_string<char>' is not literal because it is not an aggregate and has no constexpr constructors other than copy or move constructors
   85 |     class basic_string
      |           ^
src/ast/dibuilderbpf.cpp:217:34: error: constexpr variable cannot have non-literal type 'const std::string' (aka 'const basic_string<char>')
  217 |     static constexpr std::string union_prefix = "union ";
      |                                  ^
libgcc/include/c++/trunk/bits/basic_string.h:85:11: note: 'basic_string<char>' is not literal because it is not an aggregate and has no constexpr constructors other than copy or move constructors
   85 |     class basic_string
      |           ^
2 errors generated.
```

Fix by using string_view which accomplishes the same thing.

##### Checklist

### DIFF
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -1,11 +1,13 @@
 #include "dibuilderbpf.h"
 
+#include <string_view>
+
+#include <llvm/IR/Function.h>
+
 #include "libbpf/bpf.h"
 #include "log.h"
 #include "struct.h"
 #include "utils.h"
-
-#include <llvm/IR/Function.h>
 
 namespace bpftrace::ast {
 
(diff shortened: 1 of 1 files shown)


viktormalik (approved):
Thanks!

PR #3606, opened 2024-11-26
GlobalVars: Ignore global vars from external BPF programs

ajor:


viktormalik (approved):

src/globalvars.cpp
```diff
@@ -238,14 +243,13 @@ std::string to_string(GlobalVar global_var)
...
 }
 
-GlobalVar from_string(std::string_view name)
+std::optional<GlobalVar> from_string(std::string_view name)
 {
   for (const auto &[global_var, config] : GLOBAL_VAR_CONFIGS) {
     if (config.name == name)
       return global_var;
   }
-  LOG(BUG) << "Unknown global variable " << name;
```
jordalgo, line 247: This was a developer safety net from what I remember but I guess there are probably other places where it will fail if we don't reference the right global var.

PR #3607, opened 2024-11-26
Logging: Add HINT log level

ajor:
This can be used to provide suggestions to the user when an error is encountered. For example:
```
ERROR: No match for function: myfunc(int64)
HINT: Candidates:
        myfunc(string)
```

### DIFF
--- a/docs/coding_guidelines.md
+++ b/docs/coding_guidelines.md
@@ -97,6 +97,9 @@ and line number)
 - `V1`: log info only if verbose logging is enabled (-v); use this (among
   others) for printing warnings which may occur on every bpftrace execution
   (like BTF is not available)
+- `HINT`: log info that could give tips on how the user can resolve a problem
+they have encountered. This would normally be used immediately after a
+LOG(WARNING) or LOG(ERROR).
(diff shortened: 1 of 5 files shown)


viktormalik (approved):
Sounds useful.

Do we already have some cases where this could be used (maybe even as a part of this PR)?

ajor:
> Do we already have some cases where this could be used (maybe even as a part of this PR)?

I'm sure I've wanted it in the past, but I can't remember any of them right now. I'd be happy to add a usage in this PR if someone can think of something!

jordalgo (changes requested):
Mind updating the coding guidelines?
https://github.com/bpftrace/bpftrace/blob/master/docs/coding_guidelines.md#log-messages

jordalgo:
I like a hint log message. @ttreyer just added a hint type thing here: https://github.com/bpftrace/bpftrace/pull/3534

ajor:
Done and done :) 

Also expanded the expected results of a few related tests while I fiddled around in there.

jordalgo (approved):

danobi (approved):

viktormalik (approved):
Nice!

PR #3609, opened 2024-11-26
Store Block through pointers

ttreyer:
AST Nodes typically store references between themselves using pointers.
However, the newly introduced `Block` node was stored as part of the following nodes: `If`, `Unroll`, `While` and `Probe`.

Since the `Block` class has its copy constructor kept `private`, the nodes above could not copy their `Block`s and thus had their own copy constructor deleted. This didn't lead to any issue in bpftrace, but generates lots of warnings when compiling with `clang`.

This PR fixes the issue by storing reference to `Block`s using pointers. The `Block`s are then created like any other AST node by the parser.

##### Checklist

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -247,29 +247,18 @@ Block::Block(StatementList &&stmts) : stmts(std::move(stmts))
 {
 }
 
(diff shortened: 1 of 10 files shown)


jordalgo (approved):
Great cleanup! Thanks @ttreyer

Issue #3554, opened 2024-10-30
Make `--info` more compact


danobi:
### Is your feature request related to a problem? Please describe.

`bpftrace --info` is getting quite long. Would be good to make better use of screen real estate. Maybe multiple key: value entries per line?

One thing to make sure of is to keep runtime test feature parser in sync: https://github.com/bpftrace/bpftrace/blob/acc409c19be21d2a09077b2bd0c098a5babbea1b/tests/runtime/engine/runner.py#L124-L155

fbs:
I quite like it being verbose but something like `--info -v` to get verbose would also be ok

danobi:
@fbs Oh, I don't mean reduce amount of information. I meant make the information more visually compact. So something like:

```
Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
```

to

```
Kernel helpers
  probe_read: yes          probe_read_str: yes
  probe_read_user: yes     probe_read_user_str: yes
  probe_read_kernel: yes   probe_read_kernel_str: yes
```

sikehish:
I'd like to take this up :)

fbs:
@danobi I like it not being compact. Find it easier to read through / grep. But  its not a big deal either.

Maybe we should have a json output for it, as a 'machine' format? That way the test parser can also be simpler.

danobi:
@fbs if we use my example, grepping continues to work. And personally I don't find having two aligned columns any more visually difficult to parse. JSON output / machine output is reasonable, although that's not the problem I'm trying to solve.

Perhaps I should've led with this, but the issue I'm trying to solve is that `bpftrace --info` doesn't fit in one terminal screen anymore. And since we ask for `--info` output to report bugs, it adds friction to bug reporting process. If you work on remote servers a lot, then scrollback on things like tmux while trying to copy/paste is kinda klunky. I think reducing bug report friction is quite important.

Perhaps long term we want something like https://bun.sh/blog/bun-report-is-buns-new-crash-reporter. That's quite slick but it seems like a good amount of work, including a website with a backend.

danobi:
@fbs any objections before we let @sikehish have a go at this?

fbs:
oh sorry missed it.  

Let's give it a go, I don't use --info that often anyway.

go for it @sikehish 👍

sikehish:
[...]

PR #3616, opened 2024-12-01
feat(report): bpftrace --info made more compact

sikehish:
- Added a separator width to ensure consistent alignment of key-value pairs in the report function

This PR addresses issue #3554 by making the output of bpftrace --info more compact. Key-value pairs are now displayed two per line, utilizing horizontal space efficiently and maintaining readability. Column widths and separator widths were used to ensure proper alignment.

Screenshot of the output:
image

##### Checklist

### DIFF
--- a/src/bpffeature.cpp
+++ b/src/bpffeature.cpp
@@ -564,87 +564,91 @@ bool BPFfeature::has_skb_output()
   return *has_skb_output_;
 }
 
+void tabulate(std::stringstream& buf,
+              std::vector<std::pair<std::string, std::string>>& data)
+{
+  size_t len = data.size();
+  constexpr int width = 35;
+  for (size_t i = 0; i < len; i += 2) {
+    buf << std::setw(width) << std::left
+        << " " + data[i].first + ": " + data[i].second << std::setw(width);
+    if (i + 1 < len) {
+      buf << data[i + 1].first + ": " + data[i + 1].second << std::endl;
+    } else {
+      buf << std::endl;
+    }
+  }
+}
+
 std::string BPFfeature::report()
 {
   std::stringstream buf;
 
-  auto to_str = [](bool f) -> std::string { return f ? " yes" : " no"; };
+  auto to_str = [](bool f) -> std::string { return f ? "yes" : "no"; };
 
-  constexpr int width = 35;
+  std::vector<std::pair<std::string, std::string>> helpers = {
+    { "probe_read", to_str(has_helper_probe_read()) },
+    { "probe_read_str", to_str(has_helper_probe_read_str()) },
+    { "probe_read_user", to_str(has_helper_probe_read_user()) },
+    { "probe_read_user_str", to_str(has_helper_probe_read_user_str()) },
+    { "probe_read_kernel", to_str(has_helper_probe_read_kernel()) },
+    { "probe_read_kernel_str", to_str(has_helper_probe_read_kernel_str()) },
+    { "get_current_cgroup_id", to_str(has_helper_get_current_cgroup_id()) },
+    { "send_signal", to_str(has_helper_send_signal()) },
+    { "override_return", to_str(has_helper_override_return()) },
+    { "get_boot_ns", to_str(has_helper_ktime_get_boot_ns()) },
+    { "dpath", to_str(has_d_path()) },
+    { "skboutput", to_str(has_skb_output()) },
+    { "get_tai_ns", to_str(has_helper_ktime_get_tai_ns()) },
+    { "get_func_ip", to_str(has_helper_get_func_ip()) },
+    { "jiffies64", to_str(has_helper_jiffies64()) },
+    { "for_each_map_elem", to_str(has_helper_for_each_map_elem()) }
+  };
 
-  buf << "Kernel helpers" << std::endl
-      << std::setw(width) << std::left
-      << "  probe_read:" + to_str(has_helper_probe_read()) << std::setw(width)
-      << "probe_read_str:" + to_str(has_helper_probe_read_str()) << std::endl
-      << std::setw(width) << std::left << std::left
-      << "  probe_read_user:" + to_str(has_helper_probe_read_user())
-      << std::setw(width)
-      << "probe_read_user_str:" + to_str(has_helper_probe_read_user_str())
-      << std::endl
-      << std::setw(width) << std::left
-      << "  probe_read_kernel:" + to_str(has_helper_probe_read_kernel())
-      << std::setw(width)
-      << "probe_read_kernel_str:" + to_str(has_helper_probe_read_kernel_str())
-      << std::endl
-      << std::setw(width) << std::left
-      << "  get_current_cgroup_id:" + to_str(has_helper_get_current_cgroup_id())
-      << std::setw(width) << "send_signal:" + to_str(has_helper_send_signal())
-      << std::endl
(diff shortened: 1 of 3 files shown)


danobi:
Hi @sikehish , thanks for working on this! 

A few pieces of feedback:
1. Rather than right aligning the value of each feature, I think it'd be better to keep the value closer to the feature name. It's easier to eyeball without trying to line up the row visually. This also ensures anyone parsing the output is less likely to be broken. Which includes our own test suite. ie.

```
Kernel helpers
  probe_read: yes              probe_read_str: yes
  probe_read_user: yes         probe_read_user_str: yes
  probe_read_kernel: yes       probe_read_kernel_str: yes
  get_current_cgroup_id: yes   send_signal: yes
```
2. The current code works but is a bit verbose and ugly to read even if we disabled formatting for that section (via `// clang-format off` and `// clang-format on`. How about maintaining each section in an ordered list of 2-tuples? Something like:

```
std::vector<std::pair<std::string, std::string>> helpers = {
    {"helper1", "val"},
    {"helper2", "val2"},
    ...
};
```

Then a helper function that you call to tabulate each section. That'll keep formatting separate from the data.

sikehish:
> std::vector<std::pair<std::string, std::string>> helpers

Thanks for the feedback. I'm making the changes.

sikehish:
> Hi @sikehish , thanks for working on this!
> 
> A few pieces of feedback:
> 
> 1. Rather than right aligning the value of each feature, I think it'd be better to keep the value closer to the feature name. It's easier to eyeball without trying to line up the row visually. This also ensures anyone parsing the output is less likely to be broken. Which includes our own test suite. ie.
> 
> ```
> Kernel helpers
>   probe_read: yes              probe_read_str: yes
>   probe_read_user: yes         probe_read_user_str: yes
>   probe_read_kernel: yes       probe_read_kernel_str: yes
>   get_current_cgroup_id: yes   send_signal: yes
> ```
> 
> 2. The current code works but is a bit verbose and ugly to read even if we disabled formatting for that section (via `// clang-format off` and `// clang-format on`. How about maintaining each section in an ordered list of 2-tuples? Something like:
> 
> ```
> std::vector<std::pair<std::string, std::string>> helpers = {
>     {"helper1", "val"},
>     {"helper2", "val2"},
>     ...
> };
> ```
> 
> Then a helper function that you call to tabulate each section. That'll keep formatting separate from the data. ie.
> 
> ```
> buf << "Kernel helpers" << std::endl;
> buf << tabulate(helpers) << std::endl << std::endl;
> buf << "Kernel features" << std::endl;
> buf << tabulate(features) << std::endl << std::endl;
> ...
> ```

image

I've made this change. Does this look fine? If yes, then I'll proceed with the other changes(which are based on this).

danobi:
The alignment on the second column for "Map types" seems off. And also please leave a space after the `:`s so feature parsing doesn't break. But otherwise looks good!

sikehish:
> The alignment on the second column for "Map types" seems off. And also please leave a space after the `:`s so feature parsing doesn't break. But otherwise looks good!

Hi @danobi .
I've made the requested changes. Attaching a screenshot of the output.
I've used different vectors for different sections and hopefully it's more readable and less verbose now :)

image

src/bpffeature.cpp
```diff
@@ -564,59 +564,91 @@ bool BPFfeature::has_skb_output()
   return *has_skb_output_;
 }
 
+void tabulate(std::stringstream& buf,
```
danobi, line 567:
```suggestion
static void tabulate(std::stringstream& buf,
```

src/bpffeature.cpp
```diff
@@ -564,59 +564,91 @@ bool BPFfeature::has_skb_output()
...
+  };
+
+  std::vector<std::pair<std::string, std::string>> features = {
+    { "Instruction limit", std::to_string(instruction_limit()) },
+    { "Loop support", to_str(has_loop()) },
+    { "btf", to_str(has_btf()) },
+    { "module btf", to_str(has_module_btf()) },
+    { "Kernel DWARF", to_str(has_kernel_dwarf()) },
+    { "map batch", to_str(has_map_batch()) },
+    { "uprobe refcount", to_str(has_uprobe_refcnt()) }
```
danobi, line 615: Dropping the `(depends on Build:bcc bpf_attach_uprobe refcount)` is probably fine. But perhaps leave it as a comment in the code? The dependency is a little tricky and would be good to keep track of it here.

danobi (commented):
for the runtime test failures:
* looks like `basic.info flag` needs to change from EXPECT to EXPECT_REGEX (EXPECT is matching full line I think)
* `tests/runtime/engine/runner.py` needs to be updated to parse new `uprobe refcount` feature name

sikehish:
> for the runtime test failures:
> 
>     * looks like `basic.info flag` needs to change from EXPECT to EXPECT_REGEX (EXPECT is matching full line I think)
> 
>     * `tests/runtime/engine/runner.py` needs to be updated to parse new `uprobe refcount` feature name

I didn't see/encounter any runtime test failures. Am I missing something? I used to run "nix build" and then execute the resulting executable.

src/bpffeature.cpp
sikehish: > Dropping the `(depends on Build:bcc bpf_attach_uprobe refcount)` is probably fine. But perhaps leave it as a comment in the code? The dependency is a little tricky and would be good to keep track of it here.
Sure, I'll do that right away!

src/bpffeature.cpp
```diff
@@ -564,59 +564,91 @@ bool BPFfeature::has_skb_output()
...
+  };
+
+  std::vector<std::pair<std::string, std::string>> features = {
+    { "Instruction limit", std::to_string(instruction_limit()) },
+    { "Loop support", to_str(has_loop()) },
+    { "btf", to_str(has_btf()) },
+    { "module btf", to_str(has_module_btf()) },
+    { "Kernel DWARF", to_str(has_kernel_dwarf()) },
+    { "map batch", to_str(has_map_batch()) },
+    { "uprobe refcount", to_str(has_uprobe_refcnt()) } //depends on Build:bcc bpf_attach_uprobe refcount:
```
danobi, line 615:
```suggestion
    // Depends on BCC's bpf_attach_uprobe refcount feature
    { "uprobe refcount", to_str(has_uprobe_refcnt()) } 
```

tests/runtime/engine/runner.py
```diff
@@ -139,7 +139,7 @@ def __get_bpffeature():
         bpffeature["fentry"] = output.find("fentry: yes") != -1
         bpffeature["dpath"] = output.find("dpath: yes") != -1
         bpffeature["uprobe_refcount"] = \
-            output.find("uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes") != -1
+            output.find("uprobe refcount") != -1
```
danobi, line 142: Nit: keep it all on a single line please

danobi (commented):
minor nits but LGTM otherwise

sikehish:
@danobi I've made those changes. Anything else that needs to be changd?

danobi:
Please fix the clang format errors

sikehish:
> Please fix the clang format errors

I've applied clang-format. Any other changes?

danobi (approved):
thanks for the contribution!

PR #3433, opened 2024-08-31
cmake: Bump max LLVM version to 19

kraj:
##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc`
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #3618, opened 2024-12-02
Nix: update nixpkgs to 24.11

viktormalik:
The 24.11 release contains LLVM 19 so we can finally merge #3433.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -679,7 +679,11 @@ void SemanticAnalyser::visit(Call &call)
       if (!check_arg(call, Type::integer, 1, true))
         return;
       const auto bits = bpftrace_.get_int_literal(call.vargs.at(1));
-      if (bits < 0 || bits > 5) {
+      if (!bits.has_value()) {
+        // Bug here as the validity of the integer literal is already checked by
+        // check_arg above.
+        LOG(BUG) << call.func << ": invalid bits value";
+      } else if (*bits < 0 || *bits > 5) {
         LOG(ERROR, call.loc, err_)
             << call.func << ": bits " << *bits << " must be 0..5";
       }
@@ -1361,7 +1365,9 @@ void SemanticAnalyser::visit(Call &call)
       }
     } else if (arg.type.IsIntTy() && arg.is_literal) {
       auto sig = bpftrace_.get_int_literal(&arg);
-      if (!sig.has_value() || *sig < 1 || *sig > 64) {
+      if (!sig.has_value())
+        LOG(ERROR, call.loc, err_) << call.func << ": invalid signal value";
+      else if (*sig < 1 || *sig > 64) {
         LOG(ERROR, call.loc, err_)
             << std::to_string(*sig)
             << " is not a valid signal, allowed range: [1,64]";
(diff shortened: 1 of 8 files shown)


ajor (approved):

jordalgo (approved):

jordalgo:
Hmm, not sure what all the build errors are about but those probably need fixing 😜

viktormalik:
> Hmm, not sure what all the build errors are about but those probably need fixing 😜

Yup, working on those.

viktormalik:
> Do we have a policy on dropping older LLVMs yet? Could drop 13 while adding 19

We don't but it's my plan to drop 13, at least from the CI. Supporting last 6 versions (i.e. 3 years back) sounds reasonable to me.

Btw, I'm working on putting down a release policy/schedule so this could be a part of it.

viktormalik:
It took a bit more work to get the CI pass so I'm re-requesting the reviews. See the first 3 commits.

jordalgo (approved):
LGTM.

One minor nit: Not sure if there should be new semantic_analyser tests for the added error cases where `get_int_literal` returns a nullopt.

viktormalik:
> One minor nit: Not sure if there should be new semantic_analyser tests for the added error cases where `get_int_literal` returns a nullopt.

That should never happen as the same is checked by `check_arg`. So, this new condition is just to silence the warning (and cover cases when we change/misuse `check_arg`).

jordalgo:
> That should never happen as the same is checked by check_arg. So, this new condition is just to silence the warning (and cover cases when we change/misuse check_arg).

Ah ok, so then should you use `LOG(BUG)` instead?

viktormalik:
> Ah ok, so then should you use `LOG(BUG)` instead?

TBH, I don't know. I think both ways are fine as technically it's just double checking the same semantic error. But if people prefer `LOG(BUG)`, I can change it.

jordalgo:
> TBH, I don't know. I think both ways are fine as technically it's just double checking the same semantic error. But if people prefer LOG(BUG), I can change it.

If users are not supposed to see the error then it should be `LOG(BUG)`. https://github.com/bpftrace/bpftrace/blob/master/docs/coding_guidelines.md#log-messages
But if it's a gray area then this shouldn't block.

src/bpffeature.cpp
```diff
@@ -513,7 +513,7 @@ bool BPFfeature::has_uprobe_multi()
 #else
   has_uprobe_multi_ = false;
 #endif // HAVE_LIBBPF_UPROBE_MULTI
-  return *has_uprobe_multi_;
+  return *has_uprobe_multi_; // NOLINT
```
danobi, line 516: Nit: specify the lint:

```
// NOLINT(bugprone-unchecked-optional-access)
```

danobi:
Yeah if it's a double check I'd lean towards LOG(BUG). Makes the intent more clear that it should've already been checked before reaching the double check.

src/bpffeature.cpp
viktormalik: Good idea, thanks.

viktormalik:
Ok, changed to `LOG(BUG)`.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1361,7 +1365,9 @@ void SemanticAnalyser::visit(Call &call)
       }
     } else if (arg.type.IsIntTy() && arg.is_literal) {
       auto sig = bpftrace_.get_int_literal(&arg);
-      if (!sig.has_value() || *sig < 1 || *sig > 64) {
+      if (!sig.has_value())
+        LOG(ERROR, call.loc, err_) << call.func << ": invalid signal value";
```
viktormalik, line 1369: I had to go with `LOG(ERROR)` here b/c while the above `get_int_literal` does `LOG(ERROR)`, it doesn't print to `err_` so semantic analyser doesn't properly fail.

danobi (approved):

PR #3625, opened 2024-12-03
Expand COMM and ADDR columns in tcplive

klasen:
Expand `COMM` to 16 characters to match `TASK_COMM_LEN`. 
Expand `LADDR` and `RADDR` to 39 characters to support IPv6.

### DIFF
--- a/tools/tcplife.bt
+++ b/tools/tcplife.bt
@@ -30,7 +30,7 @@
 
 BEGIN
 {
-	printf("%-5s %-10s %-15s %-5s %-15s %-5s ", "PID", "COMM",
+	printf("%-5s %-16s %-39s %-5s %-39s %-5s ", "PID", "COMM",
 	    "LADDR", "LPORT", "RADDR", "RPORT");
 	printf("%5s %5s %s\n", "TX_KB", "RX_KB", "MS");
 }
@@ -91,7 +91,7 @@ kprobe:tcp_set_state
 			$daddr = ntop(AF_INET6,
 			    $sk->__sk_common.skc_v6_daddr.in6_u.u6_addr8);
 		}
-		printf("%-5d %-10.10s %-15s %-5d %-15s %-6d ", $pid,
+		printf("%-5d %-16.16s %-39s %-5d %-39s %-6d ", $pid,
 		    $comm, $saddr, $lport, $daddr, $dport);
 		printf("%5d %5d %d\n", $tp->bytes_acked / 1024,
 		    $tp->bytes_received / 1024, $delta_ms);


jordalgo:
@klasen Just curious, is there any reason to limit the length of these at all?

brendangregg:
No thanks, they are short for a reason. From the bcc contributor's guide (we need to copy a bunch of these points into bpftrace):

> Default output <80 chars wide. Try hard to keep the output less than 80 characters wide, especially the default output of the tool. That way, the output not only fits on the smallest reasonable terminal, it also fits well in slide decks, blog posts, articles, and printed material, all of which help education and adoption. Publishers of technical books often have templates they require books to conform to: it may not be an option to shrink or narrow the font to fit your output.

I frequently use the output of tcplife in presentations to promote bpftrace, and this change would make the output no longer fit on a slide. These are primarily short examples that are also tools. The strategy in bcc is to move wide-output to options, like -v for verbose, however, I'd be reclutant to add switches to tcplife for this minor reason as it would inflate the code and make it a worse example.

jordalgo:
> I'd be reclutant to add switches to tcplife for this minor reason as it would inflate the code and make it a worse example.

How about adding switches to a version of tcplife kept in the [user-tools repo](https://github.com/bpftrace/user-tools) ?

brendangregg:
Good idea, I can imagine a version where someone is experimenting with a bunch of extra options, and that would suit the user-tools repo.

klasen:
Maybe we should split this into parts:

# COMM

`TASK_COMM_LEN` is defined to be 16 chars. In the output, `COMM` is **exactly**10 chars wide (`%-10.10s`).  Values get truncated to 10 chars max :

```
PID   COMM       LADDR           LPORT RADDR           RPORT TX_KB RX_KB MS
6941  https-jsse 192.18.0.5      8543  172.30.1.1      58816      4     0 12
```

The current header is 75 chars wide. Expanding COMMS to 15 (16 - 1 null byte) would stay within the 80 columns limit. So we would get:

```
PID   COMM            LADDR           LPORT RADDR           RPORT TX_KB RX_KB MS
6941  https-jsse-nio- 192.18.0.5      8543  172.30.1.1      50232      4     0 10
```

# ADDR

In the output, ADDR has a minimum width of 15 chars but is is not truncated (`%-15s`). If you have IPv6 connections, the output will be wider than 80 chars:

```log
PID   COMM       LADDR           LPORT RADDR           RPORT TX_KB RX_KB MS
16567 curl       127.0.0.1       51948 127.0.0.1       8743       0    15 6
11472 https-jsse 127.0.0.1       8743  127.0.0.1       51948     15     0 6
0     swapper/0  fdd9:3434:bb14:200:20c:29ff:feee:792 9100  fdd9:3434:bb14:200:20c:29ff:fea3:5e75 41766      0     0 11
```

Expanding the width for these fields was to make the table line up nicely:
```log
PID   COMM            LADDR                                   LPORT RADDR                                   RPORT TX_KB RX_KB MS
11472 https-jsse-nio- 127.0.0.1                               8743  127.0.0.1                               51948     15     0 6
0     swapper/0       fdd9:3434:bb14:200:20c:29ff:feee:792    9100  fdd9:3434:bb14:200:20c:29ff:fea3:5e75   41766      0     0 11
```

# DPORT

The $dport format string is `%-6d`. That should be reduced to 5 to match the header and the maximum width of port values.

Issue #3603, opened 2024-11-23
Verifier error caused by llvm transformation


jordalgo:
### What reproduces the bug? Provide code if possible.
The script below triggers this verifier error:
```
...
39: (85) call bpf_map_update_elem#2   ; R0=scalar()
40: (07) r6 += 16                     ; R6_w=ctx(off=16)
41: (79) r7 = *(u64 *)(r6 +0)
dereference of modified ctx ptr R6 off=16 disallowed
```

```

BEGIN { @args[1] = (uint64)1; }
tracepoint:syscalls:sys_enter_kill
{
  // other if statements don't cause this error
  if (strcontains(comm, "fb-oomd")) {
    @test[args.pid] = 1;
  }
  // using this variable in place of the two lookups
  // below doesn't trigger the error
  // $x = @args[1];
  if (args.pid == @args[1]) {
    print((1));
  }
  if (args.pid == @args[1]) {
    print((1));
  }
}
```

@4ast believes it can be fixed by this though I'm not sure where to place that yet in the codegen.
`__attribute__((preserve_static_offset))`
https://clang.llvm.org/docs/AttributeReference.html#preserve-static-offset

### `bpftrace --info` output

System
  OS: Linux 5.19.0
  Arch: x86_64

Build
  version: v0.21.0-1b7c
  LLVM: 17.0.4
  bfd: yes
  liblldb (DWARF support): yes
  libsystemd (systemd notify support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes
  for_each_map_elem: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  Kernel DWARF: yes
  map batch: yes
  uprobe refcount (depends on Build:bcc bpf_attach_uprobe refcount): yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  fentry: yes
  kprobe_multi: yes
[...]

PR #3629, opened 2024-12-06
codegen: use preserve_static_offset intrinsic

amscanne:
Currently argument access uses pointer conversions with manual integer
arithmetic. This means that using the same values will often result in
LLVM choosing to save these intermediate values, which violates the
rules imposed by the verified if these are based on the context.

This situation is increasingly likely to happen as programs get larger.

Instead, we change to use proper pointer arithmetic and protect context
accesses using the `preserve_static_offsets` instrinsic, which was
introduced for Clang to prevent the same set of optimizations.

This results in optimized IR being generated that no longer uses manual
pointer casts and integer arithmetic, and ends up being converted to the
`llvm.bpf.getelementptr.and.load` intrinsic:

```
-  %14 = ptrtoint ptr %0 to i64
-  %15 = add i64 %14, 16
-  %16 = inttoptr i64 %15 to ptr
-  %17 = load volatile i64, ptr %16, align 8
+  %14 = call i64 (ptr, i1, i8, i8, i8, i1, ...) @llvm.bpf.getelementptr.and.load.i64(ptr elementtype(i8) %0, i1 true, i8 0, i8 1, i8 3, i1 false, i64 immarg 16)
```

The types used during IR generation are still very confusing, because
pointers are often treated as integer types. This could be significant
improved in the future, and types coerced only at the point where it is
needed.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1421,10 +1421,10 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
     // bpftrace's args are internally represented as 64 bit integers. However,
     // the underlying argument (of the target program) may be less than 64
     // bits. So we must be careful to zero out unused bits.
-    Value *reg = CreateGEP(getInt8Ty(),
-                           ctx,
-                           getInt64(offset * sizeof(uintptr_t)),
-                           "load_register");
+    Value *reg = CreateSafeGEP(getInt8Ty(),
+                               ctx,
+                               getInt64(offset * sizeof(uintptr_t)),
+                               "load_register");
     AllocaInst *dst = CreateAllocaBPF(builtin.type, builtin.ident);
     Value *index_offset = nullptr;
     if (argument->valid & BCC_USDT_ARGUMENT_INDEX_REGISTER_NAME) {
@@ -1433,10 +1433,10 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
         LOG(BUG) << "offset for register " << argument->index_register_name
                  << " not known";
       }
-      index_offset = CreateGEP(getInt8Ty(),
-                               ctx,
-                               getInt64(ioffset * sizeof(uintptr_t)),
-                               "load_register");
+      index_offset = CreateSafeGEP(getInt8Ty(),
+                                   ctx,
+                                   getInt64(ioffset * sizeof(uintptr_t)),
+                                   "load_register");
       index_offset = CreateLoad(getInt64Ty(), index_offset);
       if (argument->valid & BCC_USDT_ARGUMENT_SCALE) {
         index_offset = CreateMul(index_offset, getInt64(argument->scale));
@@ -2479,15 +2479,14 @@ CallInst *IRBuilderBPF::CreateSkbOutput(Value *skb,
   return call;
 }
 
-Value *IRBuilderBPF::CreatKFuncArg(Value *ctx,
-                                   SizedType &type,
-                                   std::string &name)
+Value *IRBuilderBPF::CreateKFuncArg(Value *ctx,
+                                    SizedType &type,
+                                    std::string &name)
 {
   assert(type.IsIntTy() || type.IsPtrTy());
-  ctx = CreatePointerCast(ctx, getInt64Ty()->getPointerTo());
   Value *expr = CreateLoad(
       GetType(type),
-      CreateGEP(getInt64Ty(), ctx, getInt64(type.funcarg_idx)),
+      CreateSafeGEP(getInt64Ty(), ctx, getInt64(type.funcarg_idx)),
       name);
 
   // LLVM 7.0 <= does not have CreateLoad(*Ty, *Ptr, isVolatile, Name),
@@ -2503,9 +2502,9 @@ Value *IRBuilderBPF::CreateRawTracepointArg(Value *ctx,
   int offset = atoi(builtin.substr(3).c_str());
   llvm::Type *type = getInt64Ty();
 
-  ctx = CreatePointerCast(ctx, type->getPointerTo());
+  // All arguments are coerced into Int64.
   Value *expr = CreateLoad(type,
-                           CreateGEP(type, ctx, getInt64(offset)),
+                           CreateSafeGEP(type, ctx, getInt64(offset)),
                            builtin);
 
   return expr;
@@ -2564,12 +2563,9 @@ Value *IRBuilderBPF::CreateRegisterRead(Value *ctx,
   // Bitwidth of register values in struct pt_regs is the same as the kernel
   // pointer width on all supported architectures.
   llvm::Type *registerTy = getKernelPointerStorageTy();
-  Value *ctx_ptr = CreatePointerCast(ctx, registerTy->getPointerTo());
-  // LLVM optimization is possible to transform `(uint64*)ctx` into
-  // `(uint8*)ctx`, but sometimes this causes invalid context access.
-  // Mark every context access to suppress any LLVM optimization.
+
   Value *result = CreateLoad(registerTy,
-                             CreateGEP(registerTy, ctx_ptr, getInt64(offset)),
+                             CreateSafeGEP(registerTy, ctx, getInt64(offset)),
                              name);
   // LLVM 7.0 <= does not have CreateLoad(*Ty, *Ptr, isVolatile, Name),
   // so call setVolatile() manually
@@ -2716,11 +2712,10 @@ void IRBuilderBPF::CreateSeqPrintf(Value *ctx,
       getInt64(libbpf::BPF_FUNC_seq_printf),
       seq_printf_func_ptr_type);
 
-  ctx = CreatePointerCast(ctx, getInt8Ty()->getPointerTo());
-  Value *meta = CreateLoad(getInt64Ty()->getPointerTo(),
-                           CreateGEP(getInt8Ty(), ctx, getInt64(0)),
-                           "meta");
-  dyn_cast<LoadInst>(meta)->setVolatile(true);
+  LoadInst *meta = CreateLoad(getInt64Ty()->getPointerTo(),
+                              CreateSafeGEP(getInt64Ty(), ctx, getInt64(0)),
+                              "meta");
+  meta->setVolatile(true);
 
   Value *seq = CreateLoad(getInt64Ty(),
                           CreateGEP(getInt64Ty(), meta, getInt64(0)),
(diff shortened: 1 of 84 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -4221,11 +4255,9 @@ void CodegenLLVM::probereadDatastructElem(Value *src_data,
   } else {
     // Read data onto stack
     if (data_type.IsCtxAccess() || data_type.is_btftype) {
+      // Types have already been suitable casted; just do the access.
```
jordalgo, line 4258:
```suggestion
      // Types have already been suitably casted; just do the access.
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4197,9 +4211,29 @@ void CodegenLLVM::probereadDatastructElem(Value *src_data,
                                           location loc,
                                           const std::string &temp_name)
 {
-  Value *src = b_.CreateAdd(src_data, offset);
+  // FIXME(amscanne): Pointers are often casted into integer types, and
+  // therefore we need to handle this case. Unfortunately, this means that we
+  // need to rewrap in the intrinsic to ensure that these continue to preserve
+  // offsets during access.  The solution to this to have have the local types
```
jordalgo, line 4217:
```suggestion
  // offsets during access.  The solution to this is to have the local types
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4197,9 +4211,29 @@ void CodegenLLVM::probereadDatastructElem(Value *src_data,
...
+  // accurately reflect the true types, rather than using `codegen_types`, and
+  // simply apply appropriate conversions at the moment needed.
```
jordalgo, line 4219: Not sure if I understand what this means. What is a 'codegen_type' ?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2126,17 +2125,29 @@ void CodegenLLVM::visit(FieldAccess &acc)
...
-          raw = b_.CreateLoad(field_type,
-                              b_.CreateIntToPtr(src,
-                                                field_type->getPointerTo()),
-                              true);
-        else {
+        if (type.IsCtxAccess()) {
+          // We preserve pointer semantics. See `probereadDatastructElem` where
+          // the logic is essentially constrained in the same way.
+          if (expr_->getType() != b_.GET_PTR_TY()) {
+            if (!expr_->getType()->isPointerTy())
```
jordalgo, line 2135: I'm confused here - what's the difference between this if check and the one above `if (expr_->getType() != b_.GET_PTR_TY()) {`

jordalgo:
Is it worth having a runtime test using a similar repro as in the issue: https://github.com/bpftrace/bpftrace/issues/3603

src/ast/passes/codegen_llvm.cpp
amscanne: Ah, the `emit_codegen_types` parameter for `GetType`, which coerces pointer types into `Int64` by default (https://github.com/bpftrace/bpftrace/blob/master/src/ast/irbuilderbpf.cpp#L335).

I dropped the code that coerces the context immediately into a `Int64` as a builtin (and all tests are passing, so hopefully this is fine), but unfortunately that integer cast is still going to happen with any kind of variable assignment, e.g.

```
BEGIN {
  @test = 1;
}

tracepoint:syscalls:sys_enter_kill {
  $a = args;
  print($a.pid);
  exit();
}
```

Therefore it necessary to coerce back to a pointer and rewrap in the intrinsic.

src/ast/passes/codegen_llvm.cpp
amscanne: If the type is not the correct pointer type, then it is either a) an integer, or b) a pointer of another type. Coercing into the correct pointer type is one of two different operations, depending on which is the case.

src/ast/passes/codegen_llvm.cpp
amscanne: Fixed.

src/ast/passes/codegen_llvm.cpp
amscanne: Fixed.

amscanne:
> Is it worth having a runtime test using a similar repro as in the issue: #3603 ?

I did add a codegen test that is similar to the reproducer (https://github.com/bpftrace/bpftrace/pull/3629/files#diff-2ef6096262f7a2a7c30b2ba1e4e4057e2dcfe77cd954109da249c143f795146e), but it's probably not that helpful given that it affects a *lot* of generated code (fundamentally moving from the integer add op to the getelementptr op). Given the coverage, I could just convert this to a runtime test? Seems like it might be more likely to catch regressions if the codegen tests are updated w/o paying attention to generated code (which obviously, no one would do).

src/ast/passes/codegen_llvm.h
```diff
@@ -327,6 +327,11 @@ class CodegenLLVM : public Visitor {
     return module_->getDataLayout().getTypeAllocSize(s);
   }
 
+  Value *getContext()
+  {
+    return b_.preserveStaticOffset(ctx_);
```
viktormalik, line 332: Could we wrap `ctx_` in `preserve_static_offsets` when assigning to it? That should lead to less code changes and prevent future regressions by accidentally using `ctx_` instead of `getContext()`.

src/ast/passes/codegen_llvm.cpp
viktormalik: The question is whether we should completely drop this "store pointers as integers" thing. I'm not sure if has any benefits at all. I know it's probably out of the scope of this PR but it would allow to simplify things here. I'm just wondering how much work it would be to convert everything to use raw pointers and not ints.

src/ast/irbuilderbpf.cpp
```diff
@@ -2479,12 +2479,22 @@ CallInst *IRBuilderBPF::CreateSkbOutput(Value *skb,
...
+Value *IRBuilderBPF::CreateKFuncArg(Value *ctx,
+                                    SizedType &type,
+                                    std::string &name)
 {
   assert(type.IsIntTy() || type.IsPtrTy());
+
+#if LLVM_VERSION_MAJOR <= 14
+  // Older versions of LLVM check that the type matches the GEP. Newer versions
+  // do not use this cast because we need to specifically preserve the pattern
+  // `GEP(preserveStaticOffsets(ctx))`, which doesn't exist on older versions.
```
viktormalik, line 2491: So the pointer cast would drop the `preserve_static_offsets` attribute off `ctx`, right?

What if we used `llvm.bpf.getelementptr.and.load` directly? Wouldn't that simplify the code as we wouldn't need these ifdefs? I'm just curious, not sure if that's even possible.

viktormalik (commented):
Looks good to me overall, just some high-level comments to clarify some things.

Also, it would be nice to reference some resource for the `preserve_static_offset` feature, both from the commit message and from the code. Either the commit https://github.com/llvm/llvm-project/commit/030b8cb1561db4161b108b59044717d89026cf70 or the [Phabricator link](https://reviews.llvm.org/D133361).

src/ast/passes/codegen_llvm.cpp
ajor: Does the pointer being the wrong type only apply for LLVM < 15? For newer LLVM versions there's only one pointer type, right?

If that's right, could you wrap it in a `#if LLVM_VERSION_MAJOR <= 14` check please? That'll make it easier to find and remove when we eventually deprecate the older LLVM versions.

src/ast/passes/codegen_llvm.cpp
amscanne: I'm strongly in favor of that, but I think that's a bigger exercise (and would probably be much easier with some broader cleanup first, to make it easier to reason about the types). I've refactored this change to try to reduce the scope a bit, as I'm looking for a relatively narrow fix. (Happy to push separately on the broader bits.)

src/ast/passes/codegen_llvm.cpp
amscanne: I've refactored this change a bit (might need to iterate to ensure it's working well for all versions), which I think makes this a lot easier to understand and centralizes all these conversions. It keeps the LLVM14-specific stuff, but they are generally less scattered anyways.

viktormalik (commented):
Just a couple of nits and one test-related change suggestion.

tests/codegen/preserve_static_offsets.cpp
```diff
@@ -0,0 +1,37 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, preserve_static_offsets)
```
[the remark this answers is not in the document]
amscanne: Agreed, @jordalgo also had this suggestion and I forgot to convert it. This is done now, it is a runtime regression test.

src/ast/irbuilderbpf.cpp
```diff
@@ -2807,6 +2800,40 @@ llvm::Value *IRBuilderBPF::CreatePtrOffset(const SizedType &type,
   return CreateMul(index, getInt64(elem_size));
 }
 
+llvm::Value *IRBuilderBPF::SafeGEP(llvm::Type *ty,
```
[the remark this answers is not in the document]
amscanne: Done.

src/ast/passes/codegen_llvm.h
amscanne: No longer relevant, since I've move things around to make this a bit cleaner and eliminate this method. (I swear I had already responded to this chain, but it must have gotten lost?)

src/ast/irbuilderbpf.cpp
```diff
@@ -2807,6 +2802,40 @@ llvm::Value *IRBuilderBPF::CreatePtrOffset(const SizedType &type,
   return CreateMul(index, getInt64(elem_size));
 }
 
+llvm::Value *IRBuilderBPF::CreateSafeGEP(llvm::Type *ty,
```
jordalgo, line 2805: Nice and clean with good comments. Thanks!

src/ast/irbuilderbpf.h
```diff
@@ -266,6 +266,19 @@ class IRBuilderBPF : public IRBuilder<> {
                                llvm::Value *index,
                                AddrSpace as);
 
+  // Safely handle pointer references by wrapping the address with the
+  // intrinsic `preserve_static_offset` [1], which will ensure that LLVM does
+  // not apply certain basic optimizations (notably, saving any intermediate
+  // offset from this pointer). This is required for the context pointer, which
+  // cannot be modified. This method also automatically handles casts from
```
jordalgo, line 273: nit:

```suggestion
  // offset from this pointer). This is required for the context pointer, which,
  // if modified, will trigger an error in the verifier. This method also automatically handles casts from
```

tests/runtime/regression
```diff
@@ -108,3 +108,12 @@ NAME unaligned key with aligned value
 PROG BEGIN { @mapA["aaaabbb", 0] = 1; @mapA["ccccdddd", 0] = 1; }
 EXPECT Attaching 1 probe...
 TIMEOUT 1
+
+# In some cases, LLVM can generate code that violates the BPF verifier
+# constraint that the context pointer is not modified. #3629 introduces the use
+# of intrinsics that should ensure that this doesn't happen, but this runtime
+# test exercises a pattern that previously produced code that failed to verify.
```
jordalgo, line 115: Nit: you could also add a comment to the original issue.

jordalgo (approved):
few nits but otherwise this looks great!

viktormalik (approved):
Nice!

tests/runtime/regression
amscanne: Added link to both

jordalgo:
One last thing to consider is a changelog entry. It's a pretty edge case bug at the moment and these changes are pretty deep into the internals so not sure it's worth it.

amscanne:
> One last thing to consider is a changelog entry. It's a pretty edge case bug at the moment and these changes are pretty deep into the internals so not sure it's worth it.

Yeah, I'd just consider this to be part of the standard set of bug fixes, and would have a soft preference to avoid calling out that in the changelog.

PR #3631, opened 2024-12-06
Static builds: update Alpine to 3.21

viktormalik:
We now require libbpf 1.5 for fentry probes to work so we need to update `Dockerfile.static` to use Alpine 3.21 which contains libbpf 1.5.

This comes with a couple of problems:

- bcc-static needs LLVM 18, however the latest one in Alpine is LLVM 19. It is not a problem to install Clang/LLVM 18, which we do, however, CMake then reports errors that it cannot find files in `/usr/lib/cmake/clang/`. The reason is that it's a symlink for `/usr/lib/cmake/clangXX/` which is only created when the latest Clang is installed. To fix this, create the symlink manually in Dockerfile.

- bcc in Alpine 3.21 needs an additional Clang library, `libclangCodeGen`. This, in turn, requires a number of additional LLVM libs so add all of them to our CMakeLists.txt.

##### Checklist

### DIFF
--- a/docker/Dockerfile.static
+++ b/docker/Dockerfile.static
@@ -1,6 +1,6 @@
 # This Dockerfile is used to test STATIC_LINKING=ON builds in the CI
 
-FROM alpine:3.18
+FROM alpine:3.21
 
 RUN apk add --update \
   asciidoctor \
@@ -12,8 +12,9 @@ RUN apk add --update \
   bzip2-static \
   build-base \
   cereal \
-  clang-dev \
-  clang-static \
+  clang18-dev \
+  clang18-extra-tools \
+  clang18-static \
   cmake \
   elfutils-dev \
   flex-dev \
@@ -24,8 +25,9 @@ RUN apk add --update \
   libpcap-dev \
   libc6-compat \
   linux-headers \
-  llvm16-dev \
-  llvm16-static \
+  llvm18-dev \
+  llvm18-gtest \
+  llvm18-static \
   musl-obstack-dev \
   openssl-dev \
   pahole \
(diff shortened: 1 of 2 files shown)


danobi (approved):

tnovak:
> This, in turn, requires a number of additional LLVM libs so add all of them to our CMakeLists.txt.

@viktormalik I think this change effectively bumps the minimum required LLVM version in statically-linked builds.

My local Android build that uses 14.0.6 fails with this error:
```
[ 98%] Linking CXX executable bpftrace
ld.lld: error: unable to find library -lLLVMfrontenddriver
ld.lld: error: unable to find library -lLLVMfrontendhlsl
```

AFAICT FrontendHLSL was added in LLVM 16, and FrontendDriver in 18.

danobi:
@tnovak Is there a way to get a newer llvm on android? 

More broadly, we don't have a policy on supported LLVM versions for static builds yet. IMO it would be nice if we could meet halfway there, wherever halfway is. The static stuff is not at all easy to debug (as I'm sure you know)

tnovak:
@danobi: that's fair; just wanted to flag in case it's worth documenting.

I've been using [ExtendedAndroidTools](https://github.com/facebookexperimental/ExtendedAndroidTools) to build bpftrace for Android (it pulls all the dependencies and sets up the build environment).  It currently uses LLVM 14.0.6 ([build.mk](https://github.com/facebookexperimental/ExtendedAndroidTools/blob/93e273d61a323185d3fcd99d235e956652de208c/projects/llvm/build.mk#L78)), but it should be straightforward to replace that with a newer version.

PR #3634, opened 2024-12-09
RFC: Add ability to cast probe builtin as a string

jordalgo:
This allows string comparison in BPF space which is useful if there is a glob matching against multiple probes e.g. `kprobe:vfs_*`. Utilizing the `probe` builtin enforces probe expansion so we get all the probe names in codegen.

Sytnax: `(string)probe`

Issue: https://github.com/bpftrace/bpftrace/issues/3598

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2260,6 +2260,17 @@ void CodegenLLVM::visit(Cast &cast)
     auto v = b_.CreateAllocaBPF(expr_->getType());
     b_.CreateStore(expr_, v);
     expr_ = b_.CreatePointerCast(v, b_.GetType(cast.type)->getPointerTo());
+  } else if (cast.type.IsStringTy() && cast.expr->type.IsProbeTy()) {
+    Value *buf = b_.CreateGetStrAllocation("probe", cast.loc);
+    b_.CreateMemsetBPF(buf,
+                       b_.getInt8(0),
+                       bpftrace_.config_.get(ConfigKeyInt::max_strlen));
+
+    Constant *const_str = ConstantDataArray::getString(module_->getContext(),
+                                                       probefull_,
+                                                       true);
+    b_.CreateStore(const_str, buf);
+    expr_ = buf;
   }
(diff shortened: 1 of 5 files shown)


jordalgo:
I considered allowing this syntax: `str(probe)` but went with string casting e.g. `(string)probe` as the function definition of `str` specifies a char * param, which `probe` is not.

I also considered just changing the `probe` builtin to a string type but then we lose the optimization of just passing a single int (probe id) into the perf buffer for printing.

Looking for feedback on both of these decisions. Also, if this implementation looks good I'll:

- update the change log
- add semantic analyser tests
- update the adoc

danobi:
Looks like the link to the issue is an unrelated PR.

> I also considered just changing the probe builtin to a string type but then we lose the optimization of just passing a single int (probe id) into the perf buffer for printing.

I'd say one less type in `enum class Type` is a big win. It also seems a bit more obvious than requiring a cast. Losing the integer optimization is unfortunate, but I'm inclined to prefer simplifying user experience as well as our codebase over having a relatively small optimization.

Note that the optimization is more than just for printing - it probably makes keying a map more efficient as well. Less hashing work for kernel.

jordalgo:
> Looks like the link to the issue is an unrelated PR.

Oops, left off a character: https://github.com/bpftrace/bpftrace/issues/3598 😜 

> I'm inclined to prefer simplifying user experience as well as our codebase over having a relatively small optimization.

Just to play devil's advocate. In the example I gave above (`kprobe:vfs_*`) there are 70 static strings that need to get added to our BPF program if someone uses the `probe` builtin and it's a string by default. Additionally, I don't think this PR makes the code around probe types too complicated - it does feel a little hacky though. Not offering any strong opinions here - I also like getting rid of another internal type.

danobi:
> Additionally, I don't think this PR makes the code around probe types too complicated

Right, not saying the PR makes situation any worse. Just that we have a chance here to remove some ambient complexity.

jordalgo:
Closing in favor of this: https://github.com/bpftrace/bpftrace/pull/3638

PR #3636, opened 2024-12-10
RFC: remove double dispatch from pass plumbing

amscanne:
This is an experimental branch refactoring the pass organization and some of the structures used for code generation. The intention is to simplify the code and remove the reliance on RTTI and dynamic dispatch when not needed, and generally prepare for larger changes to code generation. An attempt has been made to stage the early parts, in order to 1) gather early feedback and drop the changes if not palatable, 2) ensure that the code remains functional at each intermediate step, and 3) ensure that the changes are understandable.

Currently in the series there are two changes:
* Structure the pass manager around the `Program` type, rather than the generic `Node*`. More likely there was an early attempt to structure passes such that they could be run at any part of the AST (giving fine-grained control), but in reality they are only used at the whole program level. Leaning in to this provides more robust types in the APIs.
* Convert the `Visitor` to use a statically-defined template-based dispatch rather than relying on the double virtual dispatch and manually-constructed vtable. For now, this continues to rely on RTTI for the `Expression` and `Statement` base types and has a couple of kludges, which will be removed subsequently.

The "coming soon" changes are:
* Clean the `CodegenLLVM` pass to use a structured return value rather than `expr_` and `expr_deleter_` during code generation. This uses the new Visitor pattern, which allows for such a return and will make the code more robust and easy to understand (using the control flow for return values, rather than a lookaside value in the object).
* Convert the `Expression*` and `Statement*` base types to wrappers of an exhaustive `variant`. Prior to this, separate changes may be required to clean-up the use of `is_literal` and other metadata that has been hidden away in these classes. This will remove any need for RTTI, and clean the kludges introduced above.

All feedback and opinions welcome!


fbs:
Interesting approach, seems like a nice cleanup on first glance although 

How does this impact compiletime, size, execution speed (if measurable on the few passes we have)

amscanne:
I'm not going for the performance angle, but I added a benchmark for basic compilation (will send separate pull req for all these commits once they start to make sense, that commit is now in the stack).

Measuring before:

```
-----------------------------------------------------------------------
Benchmark                             Time             CPU   Iterations
-----------------------------------------------------------------------
BM_compile/parse_while_loop    89932842 ns     43255410 ns           30
BM_compile/passes_while_loop  183938901 ns     78489637 ns           19
```

And after:

```
-----------------------------------------------------------------------
Benchmark                             Time             CPU   Iterations
-----------------------------------------------------------------------
BM_compile/parse_while_loop    84139538 ns     39948835 ns           31
BM_compile/passes_while_loop  157490936 ns     70854450 ns           18
```

There's quite a bit of variability there (and the divergence between Time and CPU means something else is going on), but the CPU time is probably the best metric here, and I generally observed it being 5-10% faster.

I didn't observe any specific difference in the binary sizes (just checked the benchmark binary, which includes the passes). I'm not sure if compile time is affected, but I didn't really notice anything outrageous.

amscanne:
Closing in favor of #3698, #3699 and #3700.

PR #3639, opened 2024-12-11
Fix fentry probes wrt. missing_probes config

viktormalik:
Currently, fentry/fexit probes do not correctly respect the `missing_probes` config, especially for the "ignore" and "warn" settings. The reason is that while a warning is correctly displayed (or not for "ignore"), the probe is still attempted to be loaded which causes the loading of the entire BPF object fail as libbpf fails on not finding the symbol in BTF.

Fix this issue by disabling auto-load for missing probes. Note that this means that the code for the missing probes is still generated into the BPF object. This is a bit inefficient but will be necessary for AOT as we will not know which symbols exist on the target machine in advance.

Also fix the error/warning message a bit to show `symbol` instead of `:symbol` in case when module name is not given.

##### Checklist

### DIFF
--- a/src/bpfprogram.cpp
+++ b/src/bpfprogram.cpp
@@ -63,24 +63,29 @@ void BpfProgram::set_attach_target(const Probe &probe,
 
   const std::string &mod = probe.path;
   const std::string &fun = probe.attach_point;
+  const std::string attach_target = !mod.empty() ? mod + ":" + fun : fun;
 
   const std::string &btf_fun = probe.type == ProbeType::iter ? "bpf_iter_" + fun
                                                              : fun;
   if (btf.get_btf_id(btf_fun, mod) < 0) {
-    std::string msg = "No BTF found for " + mod + ":" + fun;
+    const std::string msg = "No BTF found for " + attach_target;
     if (probe.orig_name != probe.name &&
         config.get(ConfigKeyMissingProbes::default_) !=
             ConfigMissingProbes::error) {
+      // One attach point in a multi-attachpoint probe failed and the user
+      // requested not to error out. Show a warning (if requested) and continue
+      // but disable auto-loading of the program as it would make the entire BPF
+      // object loading fail.
       if (config.get(ConfigKeyMissingProbes::default_) ==
           ConfigMissingProbes::warn)
         LOG(WARNING) << msg << ", skipping.";
+      bpf_program__set_autoload(bpf_prog_, false);
     } else {
       // explicit match failed, fail hard
       throw FatalUserException(msg);
     }
   }
 
-  const std::string &attach_target = !mod.empty() ? mod + ":" + fun : fun;
   bpf_program__set_attach_target(bpf_prog_, 0, attach_target.c_str());
 }
 
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -68,6 +68,19 @@ REQUIRES_FEATURE fentry
 REQUIRES_FEATURE btf
 AFTER ./testprogs/fentry_args 1111 2222
 
+NAME fentry_warn_missing_probes
+PROG fentry:vfs_read,fentry:nonsense { exit(); }
+EXPECT WARNING: No BTF found for nonsense, skipping.
+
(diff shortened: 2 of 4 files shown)


viktormalik:
Not updating the changelog as this was introduced by #3285 which has not been released, yet.

tests/runtime/probe
```diff
@@ -68,6 +68,19 @@ REQUIRES_FEATURE fentry
 REQUIRES_FEATURE btf
 AFTER ./testprogs/fentry_args 1111 2222
 
+NAME fentry_warn_missing_probes
+PROG fentry:vfs_read,fentry:nonsense { exit(); }
+EXPECT WARNING: No BTF found for nonsense, skipping.
+
+NAME fentry_ignore_missing_probes
+PROG config = { missing_probes = "ignore" } fentry:vfs_read,fentry:nonsense { exit(); }
```
jordalgo, line 76: Does this also assert that there was no error?

jordalgo:
LGTM. Thanks for the fix! Looks like you just need to update the AOT skip list

jordalgo (approved):

src/bpfprogram.cpp
```diff
@@ -63,24 +63,29 @@ void BpfProgram::set_attach_target(const Probe &probe,
 
   const std::string &mod = probe.path;
   const std::string &fun = probe.attach_point;
+  const std::string &attach_target = !mod.empty() ? mod + ":" + fun : fun;
 
   const std::string &btf_fun = probe.type == ProbeType::iter ? "bpf_iter_" + fun
                                                              : fun;
   if (btf.get_btf_id(btf_fun, mod) < 0) {
-    std::string msg = "No BTF found for " + mod + ":" + fun;
+    const std::string &msg = "No BTF found for " + attach_target;
```
jordalgo, line 71: qq: what good does it do making these temporary strings a ref type?

viktormalik: None, I'm just sometimes too lazy to think about it and just use const ref everywhere.

Fixed, thanks.

src/bpfprogram.cpp
```diff
@@ -63,24 +63,29 @@ void BpfProgram::set_attach_target(const Probe &probe,
 
   const std::string &mod = probe.path;
   const std::string &fun = probe.attach_point;
+  const std::string &attach_target = !mod.empty() ? mod + ":" + fun : fun;
 
   const std::string &btf_fun = probe.type == ProbeType::iter ? "bpf_iter_" + fun
                                                              : fun;
   if (btf.get_btf_id(btf_fun, mod) < 0) {
-    std::string msg = "No BTF found for " + mod + ":" + fun;
+    std::string msg = "No BTF found for " + attach_target;
```
jordalgo, line 71: The `const` was good (just the ref needed to go).
```suggestion
    const std::string msg = "No BTF found for " + attach_target;
```

src/bpfprogram.cpp
```diff
@@ -63,24 +63,29 @@ void BpfProgram::set_attach_target(const Probe &probe,
 
   const std::string &mod = probe.path;
   const std::string &fun = probe.attach_point;
+  const std::string &attach_target = !mod.empty() ? mod + ":" + fun : fun;
```
jordalgo, line 66:
```suggestion
  const std::string attach_target = !mod.empty() ? mod + ":" + fun : fun;
```

jordalgo:
These are super nits, so feel free to ignore and merge if you can't be bothered 🙈

src/bpfprogram.cpp
viktormalik: Here, I think the ref makes sense as for the false case (`mod.empty()`), it doesn't have to copy `fun`.

jordalgo: I wonder how the compiler handles this. I think because the type is either a new string or a ref it might have to default to making a copy (non ref) because it doesn't know at compile time.

viktormalik: Yeah, it probably does that. Let me change to non-ref.

Issue #3446, opened 2024-09-09
Look up addresses of non exported variables in DWARF


ajor:
### Is your feature request related to a problem? Please describe.

From discussion #3445

Provide a way to access global-but-not-exported symbols. For example, `user_alpha2` (https://github.com/torvalds/linux/blob/d1f2d51b711a3b7f1ae1b46701c769c1d580fa7f/net/wireless/reg.c#L288)

### Describe the solution you'd like

Use LLDB / DWARF to look up addresses of variables if they are not found in kallsyms or an executable's symbol table.

These function calls should work if the variables are defined in DWARF:
```
kaddr("user_alpha2");
uaddr("my_non_exported_userland_variable");
```

### Describe alternative solutions or features you've considered

Require users to manually look up variable addresses from DWARF and then use those raw-addresses in their scripts:
```
$user_alpha2 = *(char[2]*)0xdeadbeef;
```

PR #3640, opened 2024-12-11
bpftrace: resolve kaddr in module

mannkafai:
Resolve symbol address in kernel module. 

It will access address whit `kaddr("user_alpha2")`  if `cfg80211`  is load as described in #3446.

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1690,9 +1690,12 @@ uint64_t BPFtrace::resolve_kname(const std::string &name) const
   std::string line;
 
   while (std::getline(file, line) && addr == 0) {
+    // https://elixir.bootlin.com/linux/v2.5.71/source/kernel/kallsyms.c#L227
+    // https://github.com/torvalds/linux/blob/v6.12/kernel/kallsyms.c#L748
     auto tokens = split_string(line, ' ');
+    auto names = split_string(tokens[2], '\t', true);
 
-    if (name == tokens[2]) {
+    if (name == names[0]) {
       addr = read_address_from_output(line);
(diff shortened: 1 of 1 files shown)


jordalgo:
@mannkafai Can you add a little more detail to this commit message and a test, if possible.

mannkafai:
I lookup address of `do_sys_openat2` and `user_alpha2`  in `/proc/kallsyms`, and access whit `kaddr` in bpftrace, it works. 

image

The `user_alpha2` is a symbol in module `cfg80211` , it‘s a `\t` between `user_alpha2` and `[cfg80211]`  in `/proc/kallsyms` .
https://github.com/torvalds/linux/blob/adc218676eef25575469234709c2d87185ca223a/kernel/kallsyms.c#L739-L753

viktormalik (commented):
Couple of remarks:
- What if there are multiple symbols of the same name in different modules (or vmlinux)? Should we just randomly pick whichever occurs first in `/proc/kallsyms`? I think it would make sense to at least issue a warning in such a case as it may be surprising to users. Also, we may want to consider adding a new optional parameter to `kaddr`.
- As @jordalgo mentioned, it would be nice to add a test. We already have a couple of runtime tests depending on kernel modules, for instance [here](https://github.com/bpftrace/bpftrace/blob/master/tests/runtime/btf#L32-L39).
- Again already mentioned by @jordalgo, the commit message could use some explanation of the change.

mannkafai:
Sorry, bpftrace can access address of symbol with module name like `kaddr("user_alpha2\t[cfg80211]")` .  So, I close the pr.

PR #3610, opened 2024-11-26
Add support for calling external BPF programs

ajor:
This is a messy draft of my work to support calling out to pre-compiled BPF Libraries. I will be splitting out most of these commits into separate PRs for review.

Work remaining (aside from TODO comments in the code):
- Add qualified identifiers syntax to support namespaced functions, e.g. `mylib::myfunc()`
- Parse types from BTF into appropriate namespaces

Would be nice (but not essential for first implementation):
- Rewrite symbol table to namespace dlopened functions
- Rewrite BTF strings to namespace functions and types before linking

#### Example 1: t-digest
```
import tdigest_lib;

BEGIN
{
  @x = str(0); // hack to allocate a buffer

  tdigest_lib_t_digest_add(@x, 2);
  tdigest_lib_t_digest_add(@x, 3);
  tdigest_lib_t_digest_add(@x, 4);
  tdigest_lib_t_digest_add(@x, 5);
  tdigest_lib_t_digest_add(@x, 6);

  printf("0th percentile: %d\n", tdigest_lib_t_digest_value_at(@x, 0));
  printf("50th percentile: %d\n", tdigest_lib_t_digest_value_at(@x, 50));
  printf("100th percentile: %d\n", tdigest_lib_t_digest_value_at(@x, 100));

  exit();
}
```
```
0th percentile: 2
50th percentile: 3
100th percentile: 6
```

#### Example 2: Python Stack Traces
```
import pystacks;

profile:hz:97
{
  @x = str(0); // hack to allocate a buffer

  $res = pystacks_read_py_stacks((pt_regs*)ctx, (task_struct*)curtask, @x, (uint64)0, (uint64)512);
  if ($res < 0) {
    return;
  }

  $msg = pystacks_get_py_msg();
  print(*$msg);
}
```
```
I1126 08:33:33.055516 1874353 pystacks.cpp:268] BPF symbols FD table contains 19 symbols.
I1126 08:33:33.055526 1874353 pystacks.cpp:281] Duplicate 19 BPF symbols, skipped.

    foo[17][-1]
    foo2[18][-1]
    <module>[19][-1]

I1126 08:33:33.065824 1874353 pystacks.cpp:268] BPF symbols FD table contains 19 symbols.
I1126 08:33:33.065835 1874353 pystacks.cpp:281] Duplicate 19 BPF symbols, skipped.

    foo[17][-1]
    foo2[18][-1]
    <module>[19][-1]
```

##### Checklist

[...]

PR #3641, opened 2024-12-13
Add `FunctionRegistry` for keeping track of functions

ajor:
This will eventually be used by builtin functions, user-defined functions from scripts and those imported from external BPF programs.  It handles duplicate function definitions by preferring non-builtin functions over builtins.

See #3610 for an example of how `FunctionRegistry` can be used. 

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -196,7 +196,7 @@ void CodegenLLVM::kstack_ustack(const std::string &ident,
                               stack_key,
                               { b_.getInt64(0), b_.getInt32(1) }));
 
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *stack_scratch_failure = BasicBlock::Create(
       module_->getContext(), "stack_scratch_failure", parent);
   BasicBlock *merge_block = BasicBlock::Create(module_->getContext(),
@@ -523,7 +523,7 @@ void CodegenLLVM::visit(Call &call)
 
     llvm::Type *mm_struct_ty = b_.GetMapValueType(map.type);
 
-    Function *parent = b_.GetInsertBlock()->getParent();
+    llvm::Function *parent = b_.GetInsertBlock()->getParent();
     BasicBlock *lookup_success_block = BasicBlock::Create(module_->getContext(),
                                                           "lookup_success",
                                                           parent);
@@ -631,7 +631,7 @@ void CodegenLLVM::visit(Call &call)
 
     llvm::Type *avg_struct_ty = b_.GetMapValueType(map.type);
 
-    Function *parent = b_.GetInsertBlock()->getParent();
+    llvm::Function *parent = b_.GetInsertBlock()->getParent();
     BasicBlock *lookup_success_block = BasicBlock::Create(module_->getContext(),
                                                           "lookup_success",
                                                           parent);
@@ -959,7 +959,7 @@ void CodegenLLVM::visit(Call &call)
     auto scoped_del = accept(arg0);
     auto addrspace = arg0->type.GetAS();
 
-    Function *parent = b_.GetInsertBlock()->getParent();
+    llvm::Function *parent = b_.GetInsertBlock()->getParent();
     BasicBlock *failure_callback = BasicBlock::Create(module_->getContext(),
                                                       "failure_callback",
                                                       parent);
@@ -1462,9 +1462,9 @@ void CodegenLLVM::visit(Call &call)
     assert(arg->type.IsIntegerTy());
     if (arg->type.GetSize() > 1) {
       llvm::Type *arg_type = b_.GetType(arg->type);
-      Function *swap_fun = Intrinsic::getDeclaration(module_.get(),
-                                                     Intrinsic::bswap,
-                                                     { arg_type });
+      llvm::Function *swap_fun = Intrinsic::getDeclaration(module_.get(),
+                                                           Intrinsic::bswap,
+                                                           { arg_type });
 
       expr_ = b_.CreateCall(swap_fun, { expr_ });
     }
@@ -1975,7 +1975,7 @@ void CodegenLLVM::visit(Unop &unop)
 
 void CodegenLLVM::visit(Ternary &ternary)
 {
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *left_block = BasicBlock::Create(module_->getContext(),
                                               "left",
                                               parent);
@@ -2537,7 +2537,7 @@ void CodegenLLVM::visit(VarDeclStatement &decl)
 
 void CodegenLLVM::visit(If &if_node)
 {
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *if_true = BasicBlock::Create(module_->getContext(),
                                            "if_body",
                                            parent);
@@ -2638,7 +2638,7 @@ void CodegenLLVM::visit(Jump &jump)
   //   br label %while_cond
   //
 
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *unreach = BasicBlock::Create(module_->getContext(),
                                            "unreach",
                                            parent);
@@ -2650,7 +2650,7 @@ void CodegenLLVM::visit(While &while_block)
   if (!loop_metadata_)
     loop_metadata_ = createLoopMetadata();
 
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *while_cond = BasicBlock::Create(module_->getContext(),
                                               "while_cond",
                                               parent);
@@ -2726,7 +2726,7 @@ void CodegenLLVM::visit(For &f)
 
 void CodegenLLVM::visit(Predicate &pred)
 {
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *pred_false_block = BasicBlock::Create(module_->getContext(),
                                                     "pred_false",
                                                     parent);
@@ -2780,8 +2780,8 @@ void CodegenLLVM::generateProbe(Probe &probe,
   auto func_name = get_function_name_for_probe(name,
                                                index,
                                                usdt_location_index);
-  Function *func = Function::Create(
-      func_type, Function::ExternalLinkage, func_name, module_.get());
+  auto *func = llvm::Function::Create(
+      func_type, llvm::Function::ExternalLinkage, func_name, module_.get());
   func->setSection(get_section_name(func_name));
   debug_.createProbeDebugInfo(*func);
 
@@ -2869,8 +2869,10 @@ void CodegenLLVM::visit(Subprog &subprog)
                                               arg_types,
                                               0);
 
-  Function *func = Function::Create(
-      func_type, Function::InternalLinkage, subprog.name(), module_.get());
+  auto *func = llvm::Function::Create(func_type,
+                                      llvm::Function::InternalLinkage,
+                                      subprog.name(),
+                                      module_.get());
   BasicBlock *entry = BasicBlock::Create(module_->getContext(), "entry", func);
   b_.SetInsertPoint(entry);
 
@@ -3197,7 +3199,7 @@ Value *CodegenLLVM::createLogicalAnd(Binop &binop)
   assert(binop.left->type.IsIntTy() || binop.left->type.IsPtrTy());
   assert(binop.right->type.IsIntTy() || binop.right->type.IsPtrTy());
 
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *lhs_true_block = BasicBlock::Create(module_->getContext(),
                                                   "&&_lhs_true",
                                                   parent);
@@ -3246,7 +3248,7 @@ Value *CodegenLLVM::createLogicalOr(Binop &binop)
   assert(binop.left->type.IsIntTy() || binop.left->type.IsPtrTy());
   assert(binop.right->type.IsIntTy() || binop.right->type.IsPtrTy());
 
-  Function *parent = b_.GetInsertBlock()->getParent();
+  llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *lhs_false_block = BasicBlock::Create(module_->getContext(),
                                                    "||_lhs_false",
                                                    parent);
@@ -3290,7 +3292,7 @@ Value *CodegenLLVM::createLogicalOr(Binop &binop)
   return b_.CreateLoad(b_.getInt64Ty(), result);
 }
 
-Function *CodegenLLVM::createLog2Function()
+llvm::Function *CodegenLLVM::createLog2Function()
 {
   auto ip = b_.saveIP();
   // Arguments: VAL (int64), K (0..5)
@@ -3332,8 +3334,8 @@ Function *CodegenLLVM::createLog2Function()
 
   FunctionType *log2_func_type = FunctionType::get(
       b_.getInt64Ty(), { b_.getInt64Ty(), b_.getInt64Ty() }, false);
-  Function *log2_func = Function::Create(
-      log2_func_type, Function::InternalLinkage, "log2", module_.get());
+  auto *log2_func = llvm::Function::Create(
+      log2_func_type, llvm::Function::InternalLinkage, "log2", module_.get());
   log2_func->addFnAttr(Attribute::AlwaysInline);
   log2_func->setSection("helpers");
   BasicBlock *entry = BasicBlock::Create(module_->getContext(),
@@ -3412,7 +3414,7 @@ Function *CodegenLLVM::createLog2Function()
   return module_->getFunction("log2");
 }
 
-Function *CodegenLLVM::createLinearFunction()
+llvm::Function *CodegenLLVM::createLinearFunction()
 {
   auto ip = b_.saveIP();
   // lhist() returns a bucket index for the given value. The first and last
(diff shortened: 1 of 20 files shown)


src/functions.cpp
```diff
@@ -0,0 +1,204 @@
...
+                                std::vector<Param> params)
+{
+  all_funcs_.push_back(std::make_unique<Function>(
+      origin, std::string{ name }, returnType, params));
+  Function &newFunc = *all_funcs_.back().get();
+  std::string fq_name;
+  if (ns.empty())
+    fq_name = name;
+  else
+    fq_name = std::string{ ns } + "::" + newFunc.name();
```
danobi, line 62: Not a huge fan of using strings with implicit structure. Probably better to have a struct that can be used as a key (hashable and equality methods). All the colon stuff in probe attachpoints got really messy and hard to understand after a while.

src/functions.cpp
```diff
@@ -0,0 +1,204 @@
...
+ * 1. If a single valid candidate function exists, pick that
+ * 2. If one of the candidates is an exact type match, pick that
+ * 3. If multiple candidates are exact type matches, prefer script/external
+ *    functions over builtin functions
+ * 4. If there are multiple, non-exact matches, return no-match
+ *
+ * Valid functions have the correct name and all arguments can be implicitly
+ * casted into all parameter types.
+ */
+const Function *FunctionRegistry::get(const std::string &name,
```
danobi, line 126: Building on above comment, likely better to take namespace and name as separate fields. Too much implicit structure in string, otherwise

src/functions.cpp
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * Returns either a single function or nullptr, when no such function exists or
+ * a single best match can not be found.
+ *
+ * Summary of overload resolution rules:
+ * 1. If a single valid candidate function exists, pick that
+ * 2. If one of the candidates is an exact type match, pick that
+ * 3. If multiple candidates are exact type matches, prefer script/external
+ *    functions over builtin functions
+ * 4. If there are multiple, non-exact matches, return no-match
```
danobi, line 121: Can we punt on function overloading? I am very much not a fan of function overloading. That aside, I think it'd be better to add it later and after some more discussion. Nothing on the project roadmap needs it yet.

danobi (dismissed):

tests/function_registry.cpp
```diff
@@ -0,0 +1,789 @@
...
+TEST_F(TestFunctionRegistryPopulated, unique_exact_args)
+{
+  test("unique_no_args", {}, unique_no_args_);
+}
+
+TEST_F(TestFunctionRegistryPopulated, unique_int8)
+{
+  test("unique_int8", { CreateInt8() }, unique_int8_);
+
+  test("unique_int8", { CreateInt16() }, R"(
```
jordalgo, line 258: I know dynamic test creation is sometimes frowned upon but I feel like this could be a loop over a vector of pairs where you make sure you get the right error message for each - I only say that because this is kind of a lot of boilerplate for pretty much the same test.

tests/function_registry.cpp
```diff
@@ -0,0 +1,789 @@
...
+}
+
+TEST_F(TestFunctionRegistryPopulated, unique_int16)
+{
+  test("unique_int16", { CreateInt8() }, unique_int16_);
+  test("unique_int16", { CreateInt16() }, unique_int16_);
+
+  test("unique_int16", { CreateUInt8() }, unique_int16_);
+
+  test("unique_int16", { CreateInt32() }, R"(
```
jordalgo, line 308: same suggestion as above.

tests/function_registry.cpp
```diff
@@ -0,0 +1,789 @@
...
+)");
+
+  test("unique_int16", { CreateUInt64() }, R"(
+ERROR: Cannot call function 'unique_int16' using argument types: (uint64)
+HINT: Candidate functions:
+  unique_int16(int16)
+)");
+}
+
+TEST_F(TestFunctionRegistryPopulated, unique_int32)
```
jordalgo, line 339: I dunno - feels like a lot of boiler plate to test every iteration of `fits_into`.

tests/function_registry.cpp
```diff
@@ -0,0 +1,789 @@
...
+  test("unique_tuple", { tuple1 }, unique_tuple_);
+  test("unique_tuple", { tuple2 }, unique_tuple_);
+
+  test("unique_tuple", { tuple3 }, R"(
+ERROR: Cannot call function 'unique_tuple' using argument types: ((int64,string[64]))
+HINT: Candidate functions:
+  unique_tuple((int32,string[64]))
+)");
+
+  // Can't pass deconstructed tuple fields as multiple arguments
```
jordalgo, line 568: why the hell not?! We do it for map keys. :trollface:

tests/function_registry.cpp
```diff
@@ -0,0 +1,789 @@
...
+  auto &foo = reg.add(Function::Origin::Script, "ns", "foo", CreateNone(), {});
+  EXPECT_EQ(nullptr, reg.get("foo", {}));
+  EXPECT_EQ(&foo, reg.get("ns::foo", {}));
+}
+
+TEST(TestFunctionRegistry, add_duplicate_of_builtin)
+{
+  FunctionRegistry reg;
+  reg.add(Function::Origin::Builtin, "foo", CreateNone(), {});
+  reg.add(Function::Origin::Script, "foo", CreateNone(), {});
```
jordalgo, line 774: this just tests that it doesn't throw?

src/functions.cpp
```diff
@@ -0,0 +1,204 @@
...
+  // functions over builtin functions
+  if (exact_matches.size() == 2) {
+    if (exact_matches[0].get().origin() == Function::Origin::Builtin)
+      return &exact_matches[1].get();
+    return &exact_matches[0].get();
+  }
+
+  // We disallow duplicate functions other than for builtins, so expect at most
+  // two exact matches.
+  if (exact_matches.size() > 2)
```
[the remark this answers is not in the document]
ajor: `FunctionRegistry::add` doesn't allow two non-builtins with the same name and types

src/types.cpp
```diff
@@ -657,7 +651,11 @@ bool SizedType::FitsInto(const SizedType &t) const
     return GetSize() <= t.GetSize();
 
   if (IsIntegerTy()) {
-    return (IsSigned() == t.IsSigned()) && (GetSize() <= t.GetSize());
+    if (IsSigned() == t.IsSigned())
+      return GetSize() <= t.GetSize();
+    if (!IsSigned())
+      return GetSize() < t.GetSize();
```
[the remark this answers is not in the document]
ajor: Unsigned into signed requires the destination to be bigger than the source, e.g. uint32 -> int64.  uint32 doesn't fit into int32 so we can't use `<=`

I'll add a comment saying this

tests/function_registry.cpp
ajor: Yep, I'll add a comment

tests/function_registry.cpp
ajor: But how do you encode the expected results of which cases are valid and which error out? I don't want to duplicate the logic from `types.cpp` because then we're not really testing anything.

What do you think of this approach:
```
  // columns represent the type of the parameter being passed in, in order:
  // int8, int16, int32, int64, uint8, uint16, uint32, uint64
  constexpr std::vector<std::vector<int>> expected = {
    {1, 0, 0, 0, 0, 0, 0, 0}, // to int8
    {1, 1, 0, 0, 1, 0, 0, 0}, // to int16
    {1, 1, 1, 0, 1, 1, 0, 0}, // to int32
    {1, 1, 1, 1, 1, 1, 1, 0}, // to int64

    {0, 0, 0, 0, 1, 0, 0, 0}, // to uint8
    {0, 0, 0, 0, 1, 1, 0, 0}, // to uint16
    {0, 0, 0, 0, 1, 1, 1, 0}, // to uint32
    {0, 0, 0, 0, 1, 1, 1, 1}, // to uint64
  };
```
I'll have a play around with it myself now

ajor: Done it this way

src/functions.cpp
```diff
@@ -0,0 +1,204 @@
...
+  for (const Function &func : funcs_by_fq_name_[fq_name]) {
+    if (func.origin() == Function::Origin::Builtin)
+      continue; // Allow duplicates of builtins
+    if (func.params().size() != params.size())
+      continue;
+    bool is_dup = true;
+    for (size_t i = 0; i < func.params().size(); i++) {
+      if (func.params()[i].type() != params[i].type()) {
+        is_dup = false;
+        break;
```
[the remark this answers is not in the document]
ajor: I think it's the reverse of what you're suggesting.

We have to check every parameter to confirm if something is a duplicate and throw. We can early exit from the loop if it's not a duplicate, in which case we don't want to throw.

src/functions.cpp
ajor: Yeah fair point. This comes from a hack I was using in #3610 to support namespacing without updating the parser (I will do the parser work at some point before merging that).

src/functions.h
```diff
@@ -0,0 +1,120 @@
...
+#include "types.h"
+
+namespace bpftrace {
+
+/**
+ * A parameter for a BpfScript function
+ */
+class Param {
+public:
+  Param(std::string name, const SizedType &type)
```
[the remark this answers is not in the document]
ajor: Because `SizedType` is a beast of a class which is pretty expensive to both copy and move.

As far as I can remember, the pattern of taking ctor parameters by value is to replace an rvalue copy with a move, at the expense of introducing an additional move for both lvalues and rvalues:

| | By const reference | By value |
|---|---|---|
| **lvalue** | copy | copy + move |
| **rvalue** | copy | move + move |

When moves are almost as expensive as copies, this tradeoff doesn't seem worth it.

src/functions.cpp
danobi: @jordalgo I think it's unneeded complexity. Why bother with exactness? Reading the discussion in #2516, it seems sufficient to only consider name and origin. Why bring types into this?

src/functions.cpp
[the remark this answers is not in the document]
jordalgo: Ah you're right. I misread.

src/functions.cpp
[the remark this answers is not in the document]
jordalgo: Ah ok, thanks for explaining.

src/functions.cpp
```diff
@@ -0,0 +1,204 @@
...
+                               param_types_str(params));
+    }
+  }
+
+  funcs_by_fq_name_[fq_name].push_back(newFunc);
+  return newFunc;
+}
+
+namespace {
+bool can_implicit_cast(const SizedType &from, const SizedType &to)
```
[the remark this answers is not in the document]
ajor: I put it here because the logic in this function about converting strings into pointers is only relevant to function calls, not string types in general. I can move it if you feel strongly though

ajor:
Updated to remove function overloading. Stashed the code changes for overloading here for future reference: https://github.com/ajor/bpftrace2/tree/function-registry-overloading

src/functions.h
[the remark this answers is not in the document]
danobi: But you copy it during construction, right? IIUC you are preventing RVO/NRVO if you do it this way. Could be totally wrong, though... cuz C++

src/ast/passes/codegen_llvm.cpp
```diff
@@ -49,6 +49,8 @@
 
 namespace bpftrace::ast {
 
+using Function = llvm::Function;
```
jordalgo, line 52: super nit (non blocking) but I might rather just adding the namespace to all instances of `Function` (like you do in some of the other files below) to distinguish at first glance. Same for the other file where you add a `using`

src/functions.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+const Function &FunctionRegistry::add(Function::Origin origin,
+                                      std::string_view ns,
+                                      std::string_view name,
+                                      const SizedType &return_type,
+                                      const std::vector<Param> &params)
+{
+  all_funcs_.push_back(std::make_unique<Function>(
+      origin, std::string{ name }, return_type, params));
+  Function &new_func = *all_funcs_.back().get();
```
jordalgo, line 57: nit: 
```suggestion
  const Function &new_func = *all_funcs_.back().get();
```

src/functions.cpp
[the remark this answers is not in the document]
jordalgo: There are just other very similar functions in types.h. I feel semi-strongly but not enough to block.

src/functions.cpp
```diff
@@ -0,0 +1,169 @@
...
+  // functions cannot be defined to accept string types (they'd take char*)
+  if (from.IsStringTy() && to.IsStringTy())
+    return true;
+
+  return false;
+}
+} // namespace
+
+/*
+ * Find the best function called 'name' for the given argument types.
```
jordalgo, line 101:
```suggestion
 * Find the best function by name for the given argument types.
```
The function's name is not 'name'

src/functions.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                      const std::vector<SizedType> &arg_types,
+                                      std::optional<location> loc) const
+{
+  FqName fq_name = {
+    .ns = std::string{ ns },
+    .name = std::string{ name },
+  };
+  auto it = funcs_by_fq_name_.find(fq_name);
+  if (it == funcs_by_fq_name_.end()) {
+    LOG(ERROR, out_) << "Function not found: '" << name << "'";
```
jordalgo, line 122: Maybe a silly question but a user won't actually get here right? I assume we'd surface an error in one of the passes with a bit of underlined code about the not found function. Long winded way of asking if this should be `LOG(BUG)`.

src/functions.h
```diff
@@ -0,0 +1,153 @@
...
+  std::string name_;
+  SizedType return_type_;
+  std::vector<Param> params_;
+  Origin origin_;
+};
+
+/**
+ * Registry of callable functions
+ *
+ * Multiple functions may share the same name, in which case overload resolution
```
jordalgo, line 94: This comment need updating?

jordalgo (commented):
Few minor nits but otherwise LGTM

src/functions.cpp
ajor: The intention is that this code gives the error for functions that don't exist. It'd avoid duplication of error messages if there are multiple callers of `FunctionRegistry::get`, and `FunctionRegistry` also has some more information that it can use to give more useful error messages in the future, e.g. something like:
```
ERROR: Function not found: "prin"
HINT: Did you mean: "print"?
```

I did miss out the `loc` argument to `LOG` though, which I've added now.

jordalgo (approved):
Nice work!

src/functions.h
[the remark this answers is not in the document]
ajor: Might be misunderstanding, sorry, but are we talking about these two (simplified) variants?
```
ByVal(SizedType type) : type_(std::move(type)) {}
ByRef(const SizedType &type) : type_(type) {}
```

Creating a situation where RVO might apply:
```
ByVal obj1{CreateVoid()};
ByRef obj2{CreateVoid()};
```

Empirically, I see "ctor + move" for `ByVal` and "ctor + copy" for `ByRef`. RVO might be applying in the `ByVal` case, but is constructing the `SizedType` in the location of the `type` argument - it still needs an expensive move to put it into the `ByVal::type_` member variable.

src/functions.cpp
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * When there are multiple candidate functions with the same name, prefer the
+ * non-builtin over the builtin function.
+ *
+ * Valid functions have the correct name and all arguments can be implicitly
+ * casted into all parameter types.
+ */
+const Function *FunctionRegistry::get(std::string_view ns,
+                                      std::string_view name,
+                                      const std::vector<SizedType> &arg_types,
```
danobi, line 113: Similar to above, I feel like it'd make more sense for caller to validate arg matching. By doing the check here (and evidenced by the error messages about "calling"), you've coupled lookup with a function call.

src/functions.cpp
```diff
@@ -0,0 +1,169 @@
...
+  };
+
+  // Check for duplicate function definitions
+  // The assumption is that builtin functions are all added to the registry
+  // before any user-defined functions.
+  // Builtin functions can be duplicated. Other functions can not.
+  for (const Function &func : funcs_by_fq_name_[fq_name]) {
+    if (func.origin() != Function::Origin::Builtin) {
+      throw std::runtime_error("Cannot redefine existing function: " +
+                               fq_name.str() + param_types_str(params));
```
danobi, line 67: Feels a bit odd to do policy check here. Seems like it should belong in caller.  But assuming we really want to keep it here, I think you should write the error to `out_` and return a nullptr. Reference type return means infallability - seems like you have a very possible, not abnormal failure possible here (eg user defines two `foo()`s).

src/functions.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    if (valid)
+      return candidate;
+  }
+
+  LOG(ERROR, loc, out_) << "Cannot call function '" << name
+                        << "' using argument types: "
+                        << arg_types_str(arg_types);
+  LOG(HINT, out_) << "Candidate function:\n  " << candidate->name()
+                  << param_types_str(candidate->params());
```
danobi, line 164: Also seems quite odd to have an `out_` on the registry. Policy probably should not exist at this level.

danobi (commented):
I think this could be further simplified by moving policy out to the caller. The `out_` seems like an unnecessarily complex side channel for failure information. My gut feeling is the API could be tighter.

danobi:
I can kinda see where you're going with the `out_` thing - presumably SemanticAnalyser will construct a single registry and pass in its `err_` ostringstream. The lookaside nature feels.. a bit unobvious though.

src/functions.cpp
ajor: There'll be multiple callers (at least one for each of builtin, UDF and external functions), so I think it makes sense to centralise the logic here to ensure `FunctionRegistry` can never contain duplicates.

ajor:
I've moved the `out` `ostream` to a parameter on `FunctionRegistry::get`, which is hopefully a bit clearer.

The reason for having the `out` stream is actually just to enable unit testing the error messages. The intended way for callers to detect an error is still from the return code, not by looking at the `ostream`.

It's my preference to checking argument types within `FunctionRegistry` so that it provides a single unified code path for function validation. `FunctionRegistry` either provides you with the exact function you're looking for or no function at all. `FunctionRegistry` also has more information available than the callers will do on the available function candidates which weren't selected, which can let it give more detailed error messages in the future. e.g.:
```
ERROR: Function not found: "prin"
HINT: Did you mean: "print"?
```
or
```
ERROR: Cannot call function "cat" with argument types: (string)
HINT: Builtin function cat(string) has been overridden by a user defined function, defined here: ...
```

Seeing as this is just an early check-in of code which will only really be used in future pull requests, I don't think we need to worry too much about the API details right now - I'm sure it'll change again anyway as it starts being used!

danobi (approved):
Fair enough - we will probably iterate on the internal API over time anyways.

I didn't get a chance to look closely at the unit tests yet, but it looks like you've given it quite some thought :)

PR #3643, opened 2024-12-13
cmake: Fix version.h build rule

tnovak:
If cmake is invoked from a separate directory, running "git describe" to get the version info fails:

```
  $ cd /tmp/build && cmake [...] ~/bpftrace && make version_h
  -- Could not obtain version string from Git. Falling back to CMake version string.
```

That's because `cmake -P` sets `CMAKE_CURRENT_SOURCE_DIR` to the current working directory[1], which may be outside of the git repository.

Fix it by explicitly passing `CMAKE_SOURCE_DIR` to `Version.cmake`.

[1] https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html

### DIFF
--- a/cmake/Version.cmake
+++ b/cmake/Version.cmake
@@ -13,7 +13,7 @@
 # Try and get a version string from Git tags
 execute_process(
   COMMAND git describe --abbrev=4 --dirty --tags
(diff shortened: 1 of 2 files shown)


danobi (approved):
thanks!

PR #3645, opened 2024-12-16
Fix alignment for multi-map keys

### LINKED ISSUE #3644: misaligned stack access when using kstack
### Minimal reproduction

```
> sudo bpftrace -ve 'profile:hz:1 { @[pid,kstack]=count(); }'
node count: 8
Attaching 1 probe...
load profile:hz:1, with BTF, with func_infos = -13: Permission denied
load profile:hz:1, with BTF = -13: Permission denied
load profile:hz:1, version: 6.11.10 = -13: Permission denied
load profile:hz:1, version: 6.11.10 = -13: Permission denied
load profile:hz:1, version: 6.11.3 = -13: Permission denied

Error log: 
0: R1=ctx() R10=fp0
0: (bf) r7 = r1                       ; R1=ctx() R7_w=ctx()
1: (85) call bpf_get_current_pid_tgid#14      ; R0_w=scalar()
2: (77) r0 >>= 32                     ; R0_w=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff))
3: (7b) *(u64 *)(r10 -44) = r0
misaligned stack access off 0+0+-44 size 8

ERROR: Error loading program: profile:hz:1
```

Using binaries from the releases page, this problem first appeared in 0.21.0. Version 0.20.4 is unaffected.

### `bpftrace --info` output

```
System
  OS: Linux 6.11.10-300.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Nov 23 00:51:20 UTC 2024
  Arch: x86_64

Build
  version: v0.21.2
  LLVM: 19.1.0
  unsafe probe: no
  bfd: yes
  liblldb (DWARF support): yes
  libsystemd (systemd notify support): no

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: no
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes
  for_each_map_elem: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: yes
  uprobe_multi: yes
  raw_tp_special: yes
  iter: yes
```

amscanne:
The allocation call for multi-map keys is currently a byte array, which has no alignment restrictions. This is converted to a set of int64 values, which will correctly enforce alignment.

Fixes #3644

##### Checklist


amscanne:
I'm happy to close this, but a bug from #3646 is still present, I believe. To be honest, I'm not actually entirely sure why your change fixed it on the release branch. Perhaps there was a separate problem there (that at least is no longer present in master), where there's a mismatch between the `map_key_type` (which is a tuple with coerced elements) and the set of expressions (which will not be coerced in the same way), and your change *did* switch the aligned check to use the `map_key_type` elements.

I think the problem which remains is the fact that the allocation gives you no guaranteed alignment (since it is a byte-array), but then we assume that can be aligned. If we are on the same page here, I can clean up the change. This is not the most elegant solution (and it's not high-priority) -- I'd rather have the allocation actually use the real underlying type, but I don't want to just forget about this either.

tests/runtime/regression
```diff
@@ -118,3 +118,10 @@ NAME preserve context pointer
 PROG BEGIN { @test[1] = (uint64)1; } tracepoint:syscalls:sys_enter_kill { if (strcontains(comm, "test")) { @test[(uint64)args.pid] = 1; } if (args.pid == @test[1]) { print((1)); } if (args.pid == @test[1]) { print((1)); exit(); } }
 EXPECT Attaching 2 probes...
 TIMEOUT 1
+
+# In the past, codegen would generate misaligned access for multi-key maps in
+# certain conditions. Make sure that this does not happen. See #3644.
+NAME unaligned multi-key map
+PROG profile:hz:1 { @[pid,kstack]=count(); }
```
jordalgo, line 125: But this test does not fail on master currently. Do we have a test that does as a result of this change not being in?

jordalgo:
> Perhaps there was a separate problem there 

Yes, I believe this was the case as map keys were refactored into tuples since that release.

> but then we assume that can be aligned. 

Not sure what you mean here. We handle tuple and record alignment before codegen here: https://github.com/bpftrace/bpftrace/blob/d543ce871e88ec06a1d17beb49a4bf2b772e99a2/src/struct.cpp#L62

BTW, I'm not against this change, I'm just trying to understand it.

amscanne:
@jordalgo With all your work on Tuples, can you just see if we still have special alignment rules that apply uniquely there? I'm just going to close this off either way, but this would be one advantage to unifying all these paths and removing the special handling in codegen.

jordalgo:
@amscanne For multi-map keys (kind of a misnomer), we still have special handling because the other parts of the keys are hidden from the user - basically all the tuple refactoring didn't account for this bit so it could still use cleaning up.

PR #3646, opened 2024-12-16
Fix alignment issues for multiple map keys

jordalgo:
This is caused by not aligning all the keys
when one of them has a size that is not a multiple of 8 bytes.

Issue: https://github.com/bpftrace/bpftrace/issues/3644

Partial fix on master: https://github.com/bpftrace/bpftrace/pull/3295

##### Checklist

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -309,7 +309,7 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
   else if (type.IsIntTy()) {
     auto sign = type.IsSigned();
     switch (type.GetIntBitWidth()) {
-      // clang-format off
+        // clang-format off
       case 64:
         if (sign)
           return std::to_string(reduce_value<int64_t>(value, nvalues) / (int64_t)div);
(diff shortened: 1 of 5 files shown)


jordalgo:
This fixes verifier complaints for the following scripts:
- `profile:hz:1 { @[0, "asdfasdf"] = 0; }`
- `profile:hz:1 { @["asdfasdf", 0] = 0; }`
- `profile:hz:1 { @[pid, kstack] = 0; }`
- `profile:hz:1 { @[kstack, pid] = 0; }`

src/output.cpp
```diff
@@ -309,7 +309,7 @@ std::string Output::value_to_str(BPFtrace &bpftrace,
   else if (type.IsIntTy()) {
     auto sign = type.IsSigned();
     switch (type.GetIntBitWidth()) {
-      // clang-format off
+        // clang-format off
```
danobi, line 312: Previous formatting looked more correct IMO

jordalgo: This is just to make clang format happy. It was a failure before this change.

danobi: ha, I see

PR #3650, opened 2024-12-21
Add basic benchmark for parsing and passes

amscanne:
This is a standalone executable for now, not wired into any tests. In
the future we may want to expand the set of benchmarks and add more
structure, to ensure that we can track parse times, compile times, etc.
and catch any massive regressions. For now however, this is a simple
tool that can be used to manually find regressions during refactors.

Here is an example of the `hello_world` benchmark:

```
------------------------------------------------------------------------
Benchmark                              Time             CPU   Iterations
------------------------------------------------------------------------
hello_world/parser                 45554 ns        45543 ns        15300
hello_world/field_analyser           970 ns          961 ns       768297
hello_world/clang_parser        18220105 ns      4284873 ns          125
hello_world/semantic_analyser      23912 ns        23898 ns        29148
hello_world/resource_analyser       4006 ns         4004 ns       171757
hello_world/codegen               846969 ns       845485 ns          790
```

Interestingly, for trivial programs the `clang_parser` is completely dominant.
However, this will need to be built out a bit further before making any hard
conclusions.

##### Checklist

- [-] Language changes are updated in `man/adoc/bpftrace.adoc`
- [-] The new behaviour is covered by tests

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -56,6 +56,12 @@ add_executable(bpftrace_test
 )
 add_test(NAME bpftrace_test COMMAND bpftrace_test)
 
+add_executable(bpftrace_benchmark
+  benchmarks.cpp
+  mocks.cpp
+  tracepoint_format_parser.cpp
+)
+
 add_subdirectory(data)
 if(${LLDB_FOUND})
   add_dependencies(bpftrace_test debuginfo_dwarf_data)
@@ -64,9 +70,12 @@ endif()
 add_dependencies(bpftrace_test debuginfo_btf_data)
 target_include_directories(bpftrace_test PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 target_include_directories(bpftrace_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+target_include_directories(bpftrace_benchmark PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+target_include_directories(bpftrace_benchmark PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 
 target_compile_definitions(bpftrace_test PRIVATE TEST_CODEGEN_LOCATION="${CMAKE_SOURCE_DIR}/tests/codegen/llvm/")
 target_link_libraries(bpftrace_test libbpftrace)
+target_link_libraries(bpftrace_benchmark libbpftrace)
 
 if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
   target_compile_definitions(bpftrace_test PRIVATE ARCH_AARCH64)
@@ -92,11 +101,15 @@ endif()
 # modules don't respect minimum requested version.
 find_package(GTest REQUIRED)
 find_package(GMock REQUIRED)
+find_package(benchmark REQUIRED)
 include_directories(SYSTEM ${GTEST_INCLUDE_DIRS} ${GMOCK_INCLUDE_DIRS})
 target_link_libraries(bpftrace_test ${GTEST_BOTH_LIBRARIES} ${GMOCK_LIBRARIES})
+target_link_libraries(bpftrace_benchmark ${GTEST_BOTH_LIBRARIES} ${GMOCK_LIBRARIES})
+target_link_libraries(bpftrace_benchmark benchmark::benchmark)
 
 find_package(Threads REQUIRED)
 target_link_libraries(bpftrace_test ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(bpftrace_benchmark ${CMAKE_THREAD_LIBS_INIT})
 
 add_subdirectory(testprogs)
 add_subdirectory(testlibs)
--- a/tests/benchmarks.cpp
+++ b/tests/benchmarks.cpp
@@ -1,0 +1,147 @@
+#include "benchmark/benchmark.h"
+#include "gmock/gmock.h"
+
+#include "mocks.h"
+
+#include "ast/passes/clone.h"
+#include "ast/passes/codegen_llvm.h"
+#include "ast/passes/field_analyser.h"
+#include "ast/passes/resource_analyser.h"
+#include "ast/passes/semantic_analyser.h"
+
+#include "bpffeature.h"
+#include "bpftrace.h"
+#include "clang_parser.h"
+#include "driver.h"
+
+namespace bpftrace::test {
+
+// We break the benchmarks into several phases, in order to track issues
+// separately. Note that currently each meaningful pass has its own phase, but
+// these could be logically merged as "optimizations" or some other passes with
+// some minor code restructuring.
+enum class Phase {
+  Parse,
+  FieldAnalyser,
+  ClangParser,
+  SemanticAnalyser,
+  ResourceAnalyser,
+  Codegen,
+};
+
+template <Phase P, Phase Is, typename Func>
+void measure(benchmark::State &state, Driver &driver, Func fn)
+{
+  // Run once only if we are not in the benchmark.
+  if constexpr (P != Is) {
+    fn();
+    return;
+  }
+
+  auto orig = std::move(driver.ctx);
+  for (auto _ : state) {
+    {
+      // Clone the AST in its current exact form. This should be relatively
+      // inexpensive (at least on the order of the pass itself), but we need to
+      // exclude this time.
+      state.PauseTiming();
+      if (orig.root != nullptr) {
+        bpftrace::ast::ASTContext newctx;
+        bpftrace::ast::Clone cloner(newctx);
+        newctx.root = orig.root;
+        cloner.clone(&newctx.root);
+        driver.ctx = std::move(newctx);
+      }
+      state.ResumeTiming();
+    }
+
+    // Run the function.
+    fn();
+  }
+}
+
+template <Phase P>
+void BM_compile(benchmark::State &state, const std::string &input)
+{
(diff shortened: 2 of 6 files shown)


jordalgo:
Very cool!

docker/Dockerfile.static
```diff
@@ -8,6 +8,7 @@ RUN apk add --update \
   bash \
   bcc-dev \
   bcc-static \
+  benchmark \
```
danobi, line 11: Dockerfile.static has a fairly specific purpose of exercising statically linked build configuration. Is there a reason this test-only dep needs to be added?

tests/CMakeLists.txt
```diff
@@ -55,6 +55,12 @@ add_executable(bpftrace_test
   ${CODEGEN_SRC}
 )
 
+add_executable(bpftrace_benchmark
+  benchmarks.cpp
+  mocks.cpp
+  tracepoint_format_parser.cpp
```
danobi, line 61: Would it make sense to extract these 2 TUs out and share with bpftrace_test target? To avoid compiling them twice

tests/benchmarks.cpp
```diff
@@ -0,0 +1,148 @@
...
+    case Phase::ResourceAnalyser:
+      return "resource_analyser";
+    case Phase::CodegenGenerateIR:
+      return "codegen_generate_ir";
+    case Phase::CodegenOptimize:
+      return "codegen_optimize";
+    case Phase::CodegenEmit:
+      return "codegen_emit";
+    default:
+      return "unknown";
```
danobi, line 54: nit: should be ok to drop default. We have the compiler flags on to warn if switch is not exhaustive. Adding `default` makes it easier for bugs to slip in

docker/Dockerfile.static
amscanne: Does it build all targets statically? E.g. `gtest-dev` is also in this list (as well as llvm18-gtest).

danobi: No, it disables tests: https://github.com/bpftrace/bpftrace/blob/870547c1a4a4974318d9fa2036429328da9fd11f/.github/include/static.sh#L20

The gtest packages might just be vestigial, then

tests/benchmarks.cpp
amscanne: Done.

tests/CMakeLists.txt
amscanne: Sure, I was on the fence. Done.

amscanne:
> > So `0.9/s` would represent something that is responsible for 90% of wall-clock time.
> 
> I'm a bit confused on the output format. 0.9/s is 90%, fine. But what is `9.76501m/s`? 9ms out of a second, so 0.976%?

Correct. I struggled with the way to structure this, since the different phases are not idempotent and you essentially need to parse, do analysis, etc. to get to codegen. But then it would take waaaaay too long to run any benchmark, since you need to spend 90x the time doing setup that the actual run takes. So I decided on just tracking iterations for the full thing, and the "rate" for each sub-phase was the best compromise that I came up with.

docker/Dockerfile.static
amscanne: Updated.

amscanne:
I need to figure out a better solution to the codeQL false positive.

jordalgo:
> I'm a bit confused on the output format. 0.9/s is 90%, fine. But what is 9.76501m/s? 9ms out of a second, so 0.976%?

@amscanne (not to speak for @danobi) but I think my issue with the format was the combination of seconds and milliseconds (or microseconds) in some but not others. I'd almost prefer we just stick to a single unit e.g. milliseconds so the above would be `900 ms` and `9.7 ms` - IIUC

tests/benchmarks.cpp
```diff
@@ -0,0 +1,148 @@
...
+                                              benchmark::Counter::kIsRate);
+  }
+}
+
+void BM_compile(benchmark::State &state, const std::string &input)
+{
+  auto bpftrace = get_mock_bpftrace();
+  bpftrace->feature_ = std::make_unique<MockBPFfeature>(true);
+
+  for (auto _ : state) {
```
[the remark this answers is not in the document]
amscanne: This is the recommended way to call the state loop. Since this is just a notice, I will ignore this (as well as the other notice.)

amscanne:
> @amscanne (not to speak for @danobi) but I think my issue with the format was the combination of seconds and milliseconds (or microseconds) in some but not others. I'd almost prefer we just stick to a single unit e.g. milliseconds so the above would be 900 ms and 9.7 ms - IIUC

Okay, I think it is simpler to just mandate that the units are in milliseconds, and that the counters are non-normalized time. Unfortunately I can't provide a counter for the unit ("ms"), but I think this is pretty readable now. Unfortunately you'll still get a *bit* of the automatic human suffixes (e.g. 5 m will mean 5 microseconds in that context) but hopefully it's still an improvement overall.

danobi (dismissed):
I don't think you pushed up the cmake changes for TU reuse. But I'm now on the fence as well :P. Sometimes it gets ugly to split out that stuff.

danobi:
By the way, might be good to update tests/README (https://github.com/bpftrace/bpftrace/blob/master/tests/README.md) with this new tool? For discoverability

jordalgo:
@amscanne Sorry, maybe I'm being dense or I've forgotten how to think over the holidays but I'm still unclear as to why we need to mix microseconds ('m') with milliseconds in the output. Am I just interpreting this wrong? I'm not sure what you mean by "non-normalized time".

```
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations clang_parser codegen_emit codegen_generate_ir codegen_optimize field_analyser      parse resource_analyser semantic_analyser
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
BM_compile/hello_world/process_time/real_time       41.5 ms         17.0 ms           34      37.2604      2.28517            0.408326          1.17408       5.57397m   0.154968         0.0316903          0.109188
```

Can this example above not be turned into this?

```
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                              Time             CPU   Iterations clang_parser codegen_emit codegen_generate_ir codegen_optimize field_analyser      parse resource_analyser semantic_analyser
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
BM_compile/hello_world/process_time/real_time       41.5 ms         17.0 ms           34   37.2604 ms   2.28517 ms         0.408326 ms       1.17408 ms     0.00557 ms  0.1549 ms      0.0316903 ms       0.109188 ms
```

amscanne:
> I don't think you pushed up the cmake changes for TU reuse. But I'm now on the fence as well :P. Sometimes it gets ugly to split out that stuff.

LOL, yeah it got ugly. The way the `tracepoint_format_parser` symbols are mixed in means they are used to override symbols in dependencies. If you put that in a library, then now you're subject to the order specified at link time, and you get nasty cycles for resolution. I just kept it as is for now since it works.

> Can this example above not be turned into this?

Correct, it cannot with the existing benchmark library. We would either need to use something else, patch it, or roll our own (and all seem more unpleasant than the alternative).

The passes are all stateful and the process doesn't really have side-effects captured, so I'm just doing a hack here to capture individual phases as *counters*. (This is not at all what they are for, which is why I don't have control over time units for the counters.) In an ideal case, we would have an `O(1)` setup routine, and then be able to run the same pass over and over again, N times to properly benchmark it. Unfortunately, it's not feasible to have a setup routine that is `O(N)` and takes 99% of the time, as the benchmark will take forever to converge.

I'm happy to put a pin in this, but it will require some refactoring of the internal pass plumbing and other work to actually capture these as they should be (separate benchmarks for the separate passes).

jordalgo:
> Correct, it cannot with the existing benchmark library. We would either need to use something else, patch it, or roll our own (and all seem more unpleasant than the alternative).

Ah ok. Sorry if I missed this point. Let's move forward as is. This will be very helpful when I start removing `is_final_pass`.

jordalgo:
> By the way, might be good to update tests/README

+1 to this.

jordalgo:
@amscanne What about finishing this up with a simple how-to doc so we can merge it. I want to use it for one of my PRs 😄

tests/benchmarks.cpp
```diff
@@ -0,0 +1,147 @@
...
+    fn();
+    return;
+  }
+
+  auto orig = std::move(driver.ctx);
+  for (auto _ : state) {
+    {
+      // Clone the AST in its current exact form. This should be relatively
+      // inexpensive (at least on the order of the pass itself), but we need to
+      // exclude this time.
```
danobi, line 46: Implicit is that the pass may be making modifications. And that we want to capture the cost of those modifications, right? Might be worth documenting this in the comment

danobi (commented):
How do we wanna order this w.r.t. the CRTP change? I've got no preference.

amscanne:
I'd rather push on continuing tidying ahead of merging this. It really *should* be quite trivial after the refactor, but I'm also aiming to burn down some other bits. (E.g. the AST isn't currently a tree, but a DAG,because of the extra bits in the expression nodes.) I'd also like to clean up the build a bit (as part of having fuzzing run) so that benchmarks can run too.

I've marked as draft so that's clear. Please yell if you consider this a blocker. I don't think it should be, since clang parser is so dominant that it's not really worth worrying about anything else, performance-wise, at least for now.

amscanne:
Abandoning in favor of #3998, which generalizes over all passes.

PR #3653, opened 2024-12-24
static: Remove gtest from dockerfile

danobi:
Shouldn't be necessary, as we do static builds with BUILD_TESTING=OFF.

##### Checklist


jordalgo (approved):
okey dokey

PR #3654, opened 2024-12-24
Turn known string allocations into globals

jordalgo:
This saves us from doing a stack allocation (using up valuable BPF stack space) and saves some instructions.

This applies to AST strings and positional params that are converted to strings.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -134,14 +134,12 @@ void CodegenLLVM::visit(PositionalParameter &param)
           expr_ = b_.getInt64(std::stoull(pstr, nullptr, 0));
         }
       } else {
-        Constant *const_str = ConstantDataArray::getString(
-            module_->getContext(), pstr, true);
-        AllocaInst *buf = b_.CreateAllocaBPF(
-            ArrayType::get(b_.getInt8Ty(), pstr.length() + 1), "str");
-        b_.CreateMemsetBPF(buf, b_.getInt8(0), pstr.length() + 1);
-        b_.CreateStore(const_str, buf);
-        expr_ = b_.CreatePtrToInt(buf, b_.getInt64Ty());
-        expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
+        auto string_param = llvm::dyn_cast<GlobalVariable>(
+            module_->getOrInsertGlobal(
+                pstr, ArrayType::get(b_.getInt8Ty(), pstr.length() + 1)));
+        string_param->setInitializer(
+            ConstantDataArray::getString(module_->getContext(), pstr));
+        expr_ = b_.CreatePtrToInt(string_param, b_.getInt64Ty());
       }
     } break;
     case PositionalParameterType::count:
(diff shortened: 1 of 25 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -134,14 +134,12 @@ void CodegenLLVM::visit(PositionalParameter &param)
...
-            module_->getContext(), pstr, true);
-        AllocaInst *buf = b_.CreateAllocaBPF(
-            ArrayType::get(b_.getInt8Ty(), pstr.length() + 1), "str");
-        b_.CreateMemsetBPF(buf, b_.getInt8(0), pstr.length() + 1);
-        b_.CreateStore(const_str, buf);
-        expr_ = b_.CreatePtrToInt(buf, b_.getInt64Ty());
-        expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
+        auto string_param = llvm::dyn_cast<GlobalVariable>(
+            module_->getOrInsertGlobal(
+                pstr, ArrayType::get(b_.getInt8Ty(), pstr.length() + 1)));
```
danobi, line 139: thoughts on a more generic name of the global var other than the string value itself? the contents are already visible on the same line in IR

could use something like "const_string" and let llvm generate disambiguators

tests/codegen/llvm/optional_positional_parameter.ll
```diff
@@ -37,12 +37,7 @@ entry:
   %cpu.id.bounded = and i64 %get_cpu_id, %1
   %2 = getelementptr [1 x [1 x [64 x i8]]], ptr @get_str_buf, i64 0, i64 %cpu.id.bounded, i64 0, i64 0
   call void @llvm.memset.p0.i64(ptr align 1 %2, i8 0, i64 64, i1 false)
-  call void @llvm.lifetime.start.p0(i64 -1, ptr %str)
-  call void @llvm.memset.p0.i64(ptr align 1 %str, i8 0, i64 1, i1 false)
-  store [1 x i8] zeroinitializer, ptr %str, align 1
-  %3 = ptrtoint ptr %str to i64
-  %probe_read_kernel_str = call i64 inttoptr (i64 115 to ptr)(ptr %2, i32 64, i64 %3)
-  call void @llvm.lifetime.end.p0(i64 -1, ptr %str)
+  %probe_read_kernel_str = call i64 inttoptr (i64 115 to ptr)(ptr %2, i32 64, i64 ptrtoint (ptr @0 to i64))
```
danobi, line 40: depending on how deep you wanna go on surgical optimizations, but seems like probe_read_kernel_str should not be necessary here. should be able to do memcpy

danobi (commented):
sweet optimization!

src/ast/passes/codegen_llvm.cpp
jordalgo: What are the benefits of a more generic name? At least when the global var is used later, you also see the string e.g.
```
call void @llvm.memcpy.p0.p0.i64(ptr align 1 %2, ptr align 1 @abc, i64 4, i1 false)
```

tests/codegen/llvm/optional_positional_parameter.ll
jordalgo: Yeah, good point. Let me look into that for another PR.

jordalgo:
> sweet optimization!

You inspired it 👍

src/ast/passes/codegen_llvm.cpp
danobi: Yeah good point. As is makes sense

danobi (approved):
love to see `+179 −305 ` on a codegen change

PR #3655, opened 2024-12-26
bpftrace: Support bash-completion

Rtoax:
Based on bash-completion [1], this commit enables bpftrace to support command autocompletion.

Also modified INSTALL.md and add '-B line' to 'bpftrace --help'.

Link: https://github.com/scop/bash-completion [1]

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,9 @@ if (ENABLE_MAN)
   add_subdirectory(man)
 endif(ENABLE_MAN)
 
+set(BASH_COMPLETION_PATH ${CMAKE_CURRENT_SOURCE_DIR}/scripts/bash-completion/bpftrace)
+install(FILES ${BASH_COMPLETION_PATH} DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/bash-completion/completions)
+
 if(NOT TARGET uninstall)
   configure_file(
     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/CmakeUninstall.cmake.in"
--- a/scripts/bash-completion/bpftrace
+++ b/scripts/bash-completion/bpftrace
@@ -1,0 +1,67 @@
+# bash completion of bpftrace(8)
+
+_bpftrace_filedir()
+{
+  # bash-completion 2.11-1091-g6e8a1546bde2 rename _filedir to _comp_compgen_filedir
+  if [[ ${BASH_COMPLETION_VERSINFO[0]} == 2 ]] && \
+     [[ ${BASH_COMPLETION_VERSINFO[1]} > 11 ]]; then
+    _comp_compgen_filedir "${@}"
+  else
+    _filedir "${@}"
+  fi
+}
+
+_bpftrace()
+{
+  local cur prev words
+
+  _init_completion -- "$@" || return
+
+  local all_args='-B -f -o -e -h --help -I --include -l -p -c
+                  --usdt-file-activation --unsafe -q --info -k -kk
+                  -V --version --no-warnings -v --dry-run -d
+                  --emit-elf --emit-llvm'
+
+  case ${prev} in
+  -B)
+    COMPREPLY=( $(compgen -W "line full none" -- ${cur}) )
+    return 0
+    ;;
+  -f)
+    COMPREPLY=( $(compgen -W "text json" -- ${cur}) )
+    return 0
+    ;;
+  -c)
+    COMPREPLY=( $(compgen -c -- ${cur}) )
+    return 0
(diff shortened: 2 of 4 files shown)


scripts/bash-completion/bpftrace
```diff
@@ -0,0 +1,90 @@
...
+                 iter:netlink
+                 iter:sockmap
+                 iter:task
+                 iter:task_file
+                 iter:task_vma
+                 iter:tcp
+                 iter:udp
+                 iter:unix"
+    COMPREPLY=( $(compgen -W "${PROBES} ${ITERS}" -- ${cur}) )
+    return 0
```
danobi, line 68: This is very interesting. Few thoughts:

1. Completing `-e` seems quite tricky. But maybe adds a lot of value for 1-liner use case, where there's only a single attachpoint. I wonder if we could do some caching to speed things up? And if bpftrace is slow at this, maybe profile the listing codepath and see if there's any easy wins to be had.
2. Completing `-l` could be quite useful as well. Probably no extra code if we're already doing `-e`
3. Hard coding stuff here does not seem like a good idea. Could confuse users, especially if we complete stuff that is not available on running kernel.

scripts/bash-completion/bpftrace
```diff
@@ -0,0 +1,90 @@
...
+    ;;
+  -p)
+    local PIDS=$(cd /proc && echo [0-9]*)
+    COMPREPLY=( $(compgen -W "$PIDS" -- ${cur}) )
+    return 0
```
danobi, line 73: Cool idea

danobi (commented):
FWIW I did not test the bash-completion bits yet.

CMakeLists.txt
```diff
@@ -220,6 +220,9 @@ if (ENABLE_MAN)
   add_subdirectory(man)
 endif(ENABLE_MAN)
 
+set(BASH_COMPLETION_PATH ${CMAKE_CURRENT_SOURCE_DIR}/scripts/bash-completion/bpftrace)
+install(FILES ${BASH_COMPLETION_PATH} DESTINATION /usr/share/bash-completion/completions/)
```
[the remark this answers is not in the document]
Rtoax: Fix it , thanks!

INSTALL.md
```diff
@@ -281,6 +281,7 @@ Use specific OS build sections listed earlier if available.
 - Libpcap
 - Systemtap SDT headers
 - Zlib development package
+- Base Completion (optional)
```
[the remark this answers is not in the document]
Rtoax: Move it to README, thanks

scripts/bash-completion/bpftrace
Rtoax: :)

scripts/bash-completion/bpftrace
Rtoax: I just fix other problems, Maybe we can use this method first and modify it later when we have a better solution, is that ok?

Rtoax: And, i just add `MISCS` for complete `-c`. By the way, like `bash -c "ls -l"`, bash-completion do nothing for `bash -c`, or we just ignore `-c` here?

scripts/bash-completion/bpftrace
```diff
@@ -0,0 +1,91 @@
...
+                 iter:sockmap
+                 iter:task
+                 iter:task_file
+                 iter:task_vma
+                 iter:tcp
+                 iter:udp
+                 iter:unix"
+    local MISCS="hardware software interval"
+    COMPREPLY=( $(compgen -W "${PROBES} ${ITERS} ${MISCS}" -- ${cur}) )
+    return 0
```
danobi, line 69: Thinking about this again, I think trying to complete contents of `-e`  causes more problems than it solves. One problem is keeping the list here in sync.

Another practical one is if you type `bpftrace -e "inte<TAB>` you get completed to `bpftrace -e "interval"`. Which is not quite right. And then it only works for the first probe.

So IMO it's better to just drop `-e` content completion.

Rtoax: just drop -e content completion, please review, thanks!

Rtoax:
Rebase to master(https://github.com/bpftrace/bpftrace/commit/36d72f648d36727f2d435b39b3f03a268a5adc2a)

danobi (approved):

danobi:
Seems like nobody has any comments to add. Mind rebasing? I'll merge after

Rtoax:
> Seems like nobody has any comments to add. Mind rebasing? I'll merge after

Hi, Thanks, Just rebase to master.

PR #3657, opened 2024-12-27
ci: Run runtime tests under controlled kernel

danobi:
This PR teaches CI how to run tests under different and tightly controlled
kernels. Under the hood it uses vmtest.

More details about vmtest can be found here:
* https://github.com/danobi/vmtest
* https://dxuuu.xyz/vmtest.html

Right now this PR adds a new job that runs runtime tests under the
current latest released kernel (6.12). I'm not sure that this is the best
long term testing policy - we'll want to iterate on it. But for now it
seems useful to do this.

TODO:
- [x] Address https://github.com/danobi/vmtest/issues/79
- [x] Pull out GHA kvm perms + package install into separate action
- [x] Configure this for a single kernel (preferably new one)
- [x] Fix absolute path binaries in runtime tests (#3659)
- [x] Fix now broken filp_close tests (#3661)
- [x] Fix broken kprobe offset test (#3662)
- [x] Fix all the broken/timed out uprobe_test runtime tests (https://github.com/bpftrace/bpftrace/pull/3666)
- [x] Build kernels in bpftrace org (probably a new repo) (https://github.com/bpftrace/kernels)
- [x] Figure out how this interacts with kernel module tracing tests
- [x] Add SETUP directive for netns usage (https://github.com/bpftrace/bpftrace/pull/3674)
- [x] Hook up nix to bpftrace/kernels
- [x] Load relevant kernel modules before running runtime tests
- [x] Figure out how to get color back on runtime tests (https://github.com/danobi/vmtest/pull/110)
- [x] Teach vmtest a way to disable fancy UI (https://github.com/danobi/vmtest/pull/110)
- [x] Adjust kprobe module offsets
- [x] Fix strftime test flakiness (https://github.com/bpftrace/bpftrace/pull/3704)
- [x] Install bpftool to unlock multi probe tests (https://github.com/bpftrace/bpftrace/pull/3703)
- [x] Explicitly set TMPDIR (https://github.com/bpftrace/bpftrace/pull/3706)

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -42,6 +42,7 @@ RUN_AOT_TESTS = os.environ.get("RUN_AOT_TESTS", "0")
 CC = os.environ.get("CC", "cc")
 CXX = os.environ.get("CXX", "c++")
 CI = os.environ.get("CI", "false")
+NIX_TARGET_KERNEL = os.environ.get("NIX_TARGET_KERNEL", "")
 TOOLS_TEST_OLDVERSION = os.environ.get("TOOLS_TEST_OLDVERSION", "")
 TOOLS_TEST_DISABLE = os.environ.get("TOOLS_TEST_DISABLE", "")
 AOT_SKIPLIST_FILE = os.environ.get("AOT_SKIPLIST_FILE", "")
@@ -219,6 +220,43 @@ def tests_finish(results: List[TestResult]):
         print(output.getvalue())
 
 
+def run_runtime_tests():
+    """Runs runtime tests, under a controlled kernel if requested"""
+    script = "./tests/runtime-tests.sh"
+
+    if NIX_TARGET_KERNEL:
+        # Bring kernel into nix store then grab the path
+        subprocess.run([nix(), "build", NIX_TARGET_KERNEL], check=True)
+        eval = subprocess.run(
+            [nix(), "eval", "--raw", NIX_TARGET_KERNEL],
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+
+        # nf_tables and xfs are necessary for testing kernel modules BTF support
+        modules = ["kvm", "nf_tables", "xfs"]
+        modprobe = f"modprobe -d {eval.stdout} -a {' '.join(modules)}"
+
+        c = f"{modprobe} && {script}"
+        cmd = ["vmtest", "-k", f"{eval.stdout}/bzImage", c]
+    else:
+        cmd = [script]
+
+    shell(
+        cmd,
+        # Don't need root if running tests in a VM
+        as_root=not NIX_TARGET_KERNEL,
+        cwd=Path(BUILD_DIR),
+        env={
+            "CI": CI,
+            "RUNTIME_TEST_COLOR": "yes",
+            # Disable UI to make CI and manual runs behave identically
+            "VMTEST_NO_UI": "1",
+        },
+    )
+
+
 def test():
     """
     Run all requested tests
@@ -246,15 +284,7 @@ def test():
         test_one(
             "runtime-tests.sh",
             lambda: truthy(RUN_TESTS),
-            lambda: shell(
-                ["./tests/runtime-tests.sh"],
-                as_root=True,
-                cwd=Path(BUILD_DIR),
-                env={
-                    "CI": CI,
-                    "RUNTIME_TEST_COLOR": "yes",
-                },
-            ),
+            run_runtime_tests,
         )
     )
     results.append(
--- a/flake.nix
+++ b/flake.nix
@@ -71,6 +71,23 @@
 
           pkgs = import nixpkgs { inherit system; };
 
+          # Download statically linked vmtest binary
+          arch = pkgs.lib.strings.removeSuffix "-linux" system;
+          vmtestVersion = "0.18.0";
+          vmtest = pkgs.stdenv.mkDerivation {
+            name = "vmtest";
+            version = vmtestVersion;
+            src = builtins.fetchurl {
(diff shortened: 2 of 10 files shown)


danobi:
Ready for a review now.

tests/runtime/usdt
```diff
@@ -74,6 +74,10 @@ RUN {{BPFTRACE}} -e 'usdt:*:tracetest:testprobe { printf("here\n" ); exit(); }'
 EXPECT here
 BEFORE ./testprogs/usdt_test
 NEW_PIDNS
+# On 9pfs, is_elf() hangs on a syscall to fstat64. We suspect it's
```
jordalgo, line 77: nit: should we make an issue to re-enable?

flake.nix
```diff
@@ -71,6 +71,23 @@
 
           pkgs = import nixpkgs { inherit system; };
 
+          # Download statically linked vmtest binary
```
jordalgo, line 74: Maybe a question not necessarily related to this PR but is it worth have separate nix flakes depending on the kind of environment we want e.g. in my local development I don't need these packages unless I want to run a vm test right?

flake.nix
```diff
@@ -214,6 +233,12 @@
                 "... libLLVM-11.so"
               ];
             };
+
+            # Kernels to run runtime tests against
```
jordalgo, line 237: nit: maybe include a comment as to why these versions

jordalgo (commented):
Overall this looks awesome! Nice work.

What do you think about adding some docs on how to run tests inside a VM locally? Maybe at least a little update here: https://github.com/bpftrace/bpftrace/blob/f8c0211da6bf4380bfe62f83a5be5126a42be362/docs/developers.md?plain=1#L54

tests/runtime/usdt
danobi: Probably not necessary. When vmtest switches over to virtiofs this will automatically be re-enabled.

And I don't know that 9pfs has any plans to fix this. Nor do I plan on fixing it. So the issue probably doesn't serve any purpose.

flake.nix
danobi: Possibly - let's wait longer and see. Right now anything that could possibly be used by CI is included in flake. I think there's a lot of value in keeping things consistent across CI and developer env.  But if it gets too heavy in the long term we can reconsider how we handle this.

danobi:
Ack. Will update docs - good idea

jordalgo (approved):
Awesome stuff! I tried the local vm test run and it worked nicely.

danobi:
Would be good to get one more review. Cc @viktormalik

.github/include/ci.py
```diff
@@ -219,6 +220,43 @@ def tests_finish(results: List[TestResult]):
...
+        subprocess.run([nix(), "build", NIX_TARGET_KERNEL], check=True)
+        eval = subprocess.run(
+            [nix(), "eval", "--raw", NIX_TARGET_KERNEL],
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+
+        # nf_tables and xfs are necessary for testing kernel modules BTF support
+        modules = ["kvm", "nf_tables", "xfs"]
```
viktormalik, line 238: Would be nice to somehow unify the list of necessary modules to a single place so that we don't have to maintain two lists (one here and one in `ci.yml`).

flake.nix
```diff
@@ -71,6 +71,42 @@
...
+            phases = [ "installPhase" ];
+            installPhase = ''
+              install -m755 -D $src $out/bin/vmtest
+            '';
+          };
+
+          # Define lambda that returns a derivation for a kernel given kernel version and SHA as input
+          mkKernel = kernelVersion: sha256:
+            with pkgs;
+            pkgs.stdenv.mkDerivation rec {
```
viktormalik, line 94: Nit: don't need `pkgs.` with the `with pkgs;` above
```suggestion
            stdenv.mkDerivation rec {
```

flake.nix
```diff
@@ -71,6 +71,42 @@
...
+                url = "https://github.com/bpftrace/kernels/releases/download/assets/linux-v${kernelVersion}.tar.zst";
+                sha256 = sha256;
+              };
+              # Remove all other phases b/c we already have a prebuilt binary
+              phases = [ "installPhase" ];
+              installPhase = ''
+                mkdir -p $out
+                tar xvf $src --strip-components=1 -C $out
+              '';
+              nativeBuildInputs = [ pkgs.gnutar pkgs.zstd ];
```
viktormalik, line 107: Same as above
```suggestion
              nativeBuildInputs = [ gnutar zstd ];
```

viktormalik (approved):
LGTM, just three nits but not going to block on those.

Great addition!

.github/include/ci.py
danobi: Right. I'm still thinking on that a bit. My feeling is these modules tests are reaching kinda deep into the kernel. So maybe it'd be a good idea to only test under vmtest. Easier to develop / reproduce locally, as it's not easy to run a GHA runner kernel locally.

If we only run module tests under vmtest, then we can drop the modprobes in ci.yml. Lemme follow up on that when we solidify a vmtest-ing policy.

flake.nix
danobi: Lemme drop the `with pkgs`. In #3751 we are moving to be more explicit. Would be good to keep it all consistent.

danobi:
I added commit 9d843987f779fe9322a0552dae65b687df762c1d to work around the softlockups. Should be harmless.

.github/include/ci.py
viktormalik: Right, that makes sense b/c we can't be sure the modules are compiled in the CI runner kernel. To me, these tests are kernel-dependent and therefore it makes sense to run them under vmtest only.

PR #3659, opened 2024-12-28
tests: runtime: Avoid /usr/sbin paths

danobi:
Avoid using absolute paths in runtime tests, as in CI under nix nothing will be reliably be present there. Instead rely on PATH search.

##### Checklist


jordalgo (approved):
Makes sense. Nice clean up.

PR #3660, opened 2024-12-28
ci: Move color configuration into ci.py

danobi:
There's no need for the configuration to be wired through the YAML. We always want the CI in GHA to be as similar to local execution as possible. So move the configuration directly into ci.py.

##### Checklist


jordalgo (approved):
Trust stamp on this one.

PR #3662, opened 2024-12-28
tests: runtime: Delete probe.kprobe_offset_fail_size

danobi:
This test has been disabled for a long time (years). It's not easily
fixed, so just delete. To reliably test this, we would need to do some
binary analysis of vmlinux and find a suitable symbol and offset to
outside of it. Too much effort for minimal gain.

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -178,16 +178,6 @@ RUN {{BPFTRACE}} --unsafe -e 'kprobe:ksys_* { @ = probe; printf("progs: "); syst
 EXPECT_REGEX progs: [1-9][0-9]+
 REQUIRES bpftool
 
-# Note: this test may fail if you've installed a new kernel but not rebooted
-# yet. Reason is b/c bpftrace will look for a vmlinux based on the running kernel's
-# version.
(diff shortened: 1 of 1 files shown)


tests/runtime/probe
```diff
@@ -178,16 +178,6 @@ RUN {{BPFTRACE}} --unsafe -e 'kprobe:ksys_* { @ = probe; printf("progs: "); syst
 EXPECT_REGEX progs: [1-9][0-9]+
 REQUIRES bpftool
 
-# Note: this test may fail if you've installed a new kernel but not rebooted
```
jordalgo, line 181: Do we not want to leave some kind of record/note as to why we don't have  test for this (perhaps with some of the details in your commit message) ?

danobi: Eh, on the fence. Hard to discover that information. And we have lots of untested stuff already.

jordalgo: Yeah kinda feels like we need some kind of special section that details why we explicitly left out some tests. Make a mental note of this one if you choose to make that PR 😉

jordalgo (approved):

PR #3664, opened 2024-12-29
Fix: for loops with per-cpu aggregations

jordalgo:
For a map that had multiple keys or a single key
that was `inBpfMemory` and contained a per-cpu
aggregation as a value (e.g. `count` or `sum`)
then retrieving the value from the `$kv` tuple was broken e.g.
```
@map[1, 2] = count();
for ($kv : @map) {
  print($kv.1); // this would always be 0
}
```

The fix is to just use the callback arg ptr
instead of allocating a new value (as the key)
and trying to store something in it.

Previously, for values inBpfMemory (like tuples) the
key in the line `b_.CreateStore(key, key_ptr);` was
populated with a pointer instead of the actual value
so key_ptr was storing a pointer and then we were
taking a pointer to a pointer for the map lookup,
which was wrong.
##### Checklist

### DIFF
--- a/tests/codegen/llvm/avg_cast_loop.ll
+++ b/tests/codegen/llvm/avg_cast_loop.ll
@@ -72,10 +72,7 @@ define internal i64 @map_for_each_cb(ptr %0, ptr %1, ptr %2, ptr %3) section ".t
   %val_2 = alloca i64, align 8
   %val_1 = alloca i64, align 8
   %i = alloca i32, align 4
-  %lookup_key = alloca i64, align 8
   %key = load i64, ptr %1, align 8
-  call void @llvm.lifetime.start.p0(i64 -1, ptr %lookup_key)
-  store i64 %key, ptr %lookup_key, align 8
   call void @llvm.lifetime.start.p0(i64 -1, ptr %i)
   call void @llvm.lifetime.start.p0(i64 -1, ptr %val_1)
   call void @llvm.lifetime.start.p0(i64 -1, ptr %val_2)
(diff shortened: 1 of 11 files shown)


jordalgo:
This might need to go into the release branch. Will check tomorrow.

tests/runtime/for
```diff
@@ -72,3 +72,11 @@ EXPECT_REGEX ^abc\ndef\ndef$
 NAME variable context multiple
 PROG BEGIN { @map[0] = 0; $var1 = 123; $var2 = "abc"; for ($kv : @map) { print(($var1, $var2)); } exit(); }
 EXPECT (123, abc)
+
+NAME map two keys with a per cpu aggregation
```
jordalgo, line 76: yes, both of these tests don't pass on master 😞

tests/codegen/llvm/avg_cast_loop.ll
danobi, line 1: why does this fix work? looking at this codegen change, seems like the PR is a noop.

(although avoiding a stack allocation is still better, but that's different)

jordalgo: I'll add a codegen test too but basically for values `inBpfMemory` (like tuples) the `key` in the line `b_.CreateStore(key, key_ptr);` was populated with a pointer instead of the actual value so `key_ptr` was storing a pointer and then we were taking a pointer to a pointer for the map lookup (see line 4557)

danobi: Ah ok. This detail would be good to include in PR and commit message

jordalgo: @danobi Updated.

danobi (approved):

danobi:
Does this need a fixes entry in changelog? Or was buggy behavior not released yet?

jordalgo:
@danobi I'm going to make a PR for the release branch with the changelog entry and then another PR for master with the changelog entry under the new release branch section. 🏋️

PR #3666, opened 2024-12-29
tests: runtime: Remove races with uprobe_test

danobi:
There was a `sleep(1)` in the beginning of uprobe_test. This was a racey
way to synchronize between `BEFORE` clause and bpftrace execution.

Not only was it racey, but it introduces unnecessary delay in the event
that bpftrace runs fast. We can shave off some CI time by doing more
precise synchronization.

##### Checklist


jordalgo (approved):
Nice fix up! 🔧

Issue #3423, opened 2024-08-27
RFC: Introduce a keyword for map/var definitions/declarations.


jordalgo:
This issue is just to consolidate some discussions here: https://github.com/bpftrace/bpftrace/pull/3367, https://github.com/bpftrace/bpftrace/issues/2954

Doing a bit of research, here is how other languages break down in terms of keywords on the left side of variable declarations/initializations.

## Languages that don't have any space-separate label/keyword:
- Python
- Ruby (global variables are prefixed with `$`)
- R
- Haskell
- Perl (scalar variables are prefixed with `$`)
- PHP/Hack (prefixed with `$`)
- C / C++ / C# / Java (require a type on the left)

## Languages that have a label/keyword
- Clojure (`def`)
- Go (`var`)
- Lua  / Bash (global variables have no label but local variables have `local` label)
- Scala / Kotlin (`var` and `val`)
- Dart (has `var` but also can skip that by specifying the type before the variable name)
- Zig (`var` and `const`)
- Rust (`let` and `const`)
- Javascript ((`let` and `const` - `var` is deprecated)
- Swift (`let` and `var`, though the latter is mutable)

**For a quick vote (but feel free to comment):**
Thumbs up this comment if you are in favor of ADDING a new label/keyword 👍 
Thumbs down this comment if you are in favor of NOT ADDING a new label/keyword 👎

fbs:
Thanks for opening this. Useful to get a consistent experience around all types of variables.

As we already have `$` and `@` to identify variables I don't think we have a need for let from a parsing perspective. One place where `let` can be useful is if we want to introduce [variable shadowing](https://en.wikipedia.org/wiki/Variable_shadowing) in scopes. E.g.

```
probe {
  $x = 1
  for (...) {
     $x = "abc"
     print($x)
  }
}
```

This would be a type error right now. Although we can probably also get away without shadowing?

jordalgo:
> One place where let can be useful is if we want to introduce [variable shadowing](https://en.wikipedia.org/wiki/Variable_shadowing) in scopes

I think even then I might prefer the keyword `local` as `let` seems to be a bit ambiguous across languages.

viktormalik:
Agreed with the argument of @fbs here.

jordalgo:
[...]

PR #3668, opened 2024-12-30
Unambiguously adopt zero value semantics

amscanne:
Rather than relying on semantic analysis to detect cases where variables are unassigned, we can simplify the code by simply specifying that variables are always initialized with their respect "zero values", that is the value specified by an in-memory representation of all zero.

In fact, this is already the case today, both for implementation and documentation.

For the implementation, during code generation variables of all sizes are zeroed. We do not make any attempt to optimize for the case where there is a different initializer, and the generated bytecode often looks like `store 0`, followed by `store <x>`. This is perfectly reasonable, and we can defer such analysis and optimization to LLVM. Doing a complete analysis for whether a variable is assigned can become arbitrarily complex, and in all likelihood we would need to generate this pattern to avoid undefined behavior (which I suspect the BPF verified would reject anyways).

Just to provide an example of this analysis being complex, it is already trivial to escape the existing analysis by simply having the variable defined in the first branch of an `if` statement:
```
BEGIN {
   let $x: int32;
   let $y: int32;
   $y = 1;
   if ($y == 0) {
     $x = 1;
   } else {
     if ($x == 0) {
        print("x is 0");
     } else {
        print("x is 1");
     }
   }
   exit();
}
```

For the documentation, the man page already states:
```
If no assignment is specified variables will initialize to 0.
```

This change merely drops the associated semantic analyzer checks and metadata, as well as the test against that specific error.

##### Checklist

- [-] The new behaviour is covered by tests

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -822,7 +822,8 @@ $b = --$a; // a = 9; b = 9
 ----
 
 Note that maps will be implicitly declared and initialized to 0 if not already declared or defined.
-Scratch variables must be initialized before using these operators.
+
+Scratch variables must be declared or initialized before using these operators.
 
 Note `{plus}{plus}`/`--` on a shared global variable can lose updates. See <<map-functions-count, `count()`>> for more details.
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -1756,13 +1756,7 @@ void SemanticAnalyser::visit(Map &map)
 void SemanticAnalyser::visit(Variable &var)
 {
   if (auto *found = find_variable(var.ident)) {
-    if (found->was_assigned) {
-      var.type = found->type;
-    } else {
-      LOG(ERROR, var.loc, err_)
-          << "Variable used before it was assigned: " << var.ident;
-      var.type = CreateNone();
-    }
+    var.type = found->type;
     return;
   }
 
@@ -2492,8 +2486,7 @@ void SemanticAnalyser::visit(For &f)
   scope_stack_.push_back(&f);
 
   variables_[scope_stack_.back()][decl_name] = { .type = f.decl->type,
-                                                 .can_resize = true,
-                                                 .was_assigned = true };
+                                                 .can_resize = true };
 
   loop_depth_++;
   accept_statements(f.stmts);
@@ -3008,30 +3001,23 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
       }
     }
     if (type_mismatch_error) {
-      auto err_segment = foundVar.was_assigned
-                             ? "when variable already contains a value of type"
-                             : "when variable already has a type";
       LOG(ERROR, assignment.loc, err_)
           << "Type mismatch for " << var_ident << ": "
           << "trying to assign value of type '" << assignTy << "' "
-          << err_segment << " '" << storedTy << "'";
+          << "when variable has type '" << storedTy << "'";
     } else {
-      if (!foundVar.was_assigned) {
-        // The assign type is possibly more complete than the stored type,
-        // which could come from a variable declaration. The assign type may
-        // resolve builtins like `curtask` which also specifies the address
-        // space.
+      // The assign type is possibly more complete than the stored type, which
+      // could come from a variable declaration. The assign type may resolve
+      // builtins like `curtask` which also specifies the address space.
+      if (assignTy.IsPtrTy() || assignTy.IsArrayTy())
         foundVar.type = assignTy;
-        foundVar.was_assigned = true;
-      }
       var_scope = scope;
     }
   }
 
   if (var_scope == nullptr) {
     variables_[scope_stack_.back()].insert(
-        { var_ident,
-          { .type = assignTy, .can_resize = true, .was_assigned = true } });
+        { var_ident, { .type = assignTy, .can_resize = true } });
     var_scope = scope_stack_.back();
   }
 
@@ -3099,21 +3085,14 @@ void SemanticAnalyser::visit(VarDeclStatement &decl)
         decl.var->type = foundVar.type;
       }
 
-      if (is_final_pass() && !foundVar.was_assigned) {
-        LOG(WARNING, decl.loc, out_)
-            << "Variable " << var_ident << " never assigned to.";
-      }
-
       return;
     }
   }
 
   bool can_resize = !decl.set_type || decl.var->type.GetSize() == 0;
 
-  variables_[scope_stack_.back()].insert({ var_ident,
-                                           { .type = decl.var->type,
-                                             .can_resize = can_resize,
-                                             .was_assigned = false } });
+  variables_[scope_stack_.back()].insert(
+      { var_ident, { .type = decl.var->type, .can_resize = can_resize } });
   variable_decls_[scope_stack_.back()].insert({ var_ident, decl.loc });
 }
 
@@ -3434,8 +3413,7 @@ void SemanticAnalyser::visit(Subprog &subprog)
   top_level_node_ = &subprog;
   for (SubprogArg *arg : subprog.args) {
     variables_[scope_stack_.back()].insert(
-        { arg->name(),
-          { .type = arg->type, .can_resize = true, .was_assigned = true } });
+        { arg->name(), { .type = arg->type, .can_resize = true } });
   }
   Visitor::visit(subprog);
   scope_stack_.pop_back();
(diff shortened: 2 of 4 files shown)


jordalgo:
It's true, as per the example, that you can get past the "use before assign" checks but I think for the cases that it DOES catch are still more helpful to the user than ambiguous/hurtful - and worth the added complexity IMHO. We have a lot of internal types that, yes, initialize to a 0 value but don't really make sense as such e.g. `macaddr_t`, `timestamp`, `inet`. Using variables of these types before assigning to them is a bad/confusing experience. 

I'll think on this a bit more but I might rather work on improving the "use before assign" rather than removing it entirely. I do think we should update the adoc though to clarify this behavior if we don't end up merging this.

ajor:
> I might rather work on improving the "use before assign" rather than removing it entirely

+1. I'd say that the escapes in the OP are bugs in the semantic analyser and would prefer to fix them rather than change the language design again. It should be possible detect uninitialised variables 100% of the time - other languages do this already.

@amscanne We did have zero values semantics until very recently (the upcoming v0.22 will be the first release without them). There was a conscious decision to instead error out when using uninitialised variables, to help reduce bugs when writing larger scripts.

amscanne:
Doing a full analysis to determine whether something is set at the point of user is taking a on a lot of complexity (at the extreme, you're essentially tracking an arbitrary computation). I believe that most languages just adopt a warning when something is never set or never used, which is must simpler. You *could* just require that the variable is set in all paths, but now you're breaking code that previously worked and throwing a different kind of friction at the user. For example, this works today:

```
BEGIN {
  let $x : int32;
  $y = 1000;
  if ($y < 100) {
    $x = 2;
  }
  // $x may or may not be set
  if ($y > 100) {
    print($x);
  }
}
```

Now if you make that break by testing whether X is set in *all* branches of the if, you can get a warning. But that breaks this piece of code:

```
BEGIN {
  let $x : int32;
  if (curtask) {
    $x = 2;
  }
  ...
  if (curtask) {
    print($x);
  }
}
```

Regarding the "zero values don't make sense" point, that seems like a separate source of friction. To the best of my knowledge, there are well-defined zero values for all types in C/C++ (e.g. initialized with '{}'), even if they aren't always useful. Is there another way to produce the zero value if I e.g. want to produce a "null" entry for something? There would have to be another builtin (e.g. `zero(type)`) otherwise you're potentially forcing a call to another builtin just to have a throw-away value. I get that it's trying to protect the user from certain edges, but it feels like it's outsmarting them a bit.

Anyways, I'm not trying to relitigate whatever decision was made (though I'm curious, is there a issue# or link?). Just airing my opinion. Should I fix the existing documentation regarding initialization to zero?

jordalgo:
> You could just require that the variable is set in all paths, but now you're breaking code that previously worked and throwing a different kind of friction at the user.

This isn't true YET. The let syntax was only recently introduced on master and hasn't gone out in a release (though we're planning one shortly that includes `let`).

> I believe that most languages just adopt a warning when something is never set or never used, which is must simpler. 

I can get behind a WARNING instead of an ERROR (for the reasons you stated). We're just trying to let them know that using something that they haven't explicitly initialized is sort of UB (e.g. with the types I mentioned). If other folks are ok with that change (cc @ajor) I think it should go into the new release branch and master.

> Anyways, I'm not trying to relitigate whatever decision was made

Please continue to engage and ask questions!

ajor:
> You _could_ just require that the variable is set in all paths, but now you're breaking code that previously worked and throwing a different kind of friction at the user.

This is what I was thinking. You're right that these code snippets would stop working, but I think that them passing the analysis right now is a bug.

> though I'm curious, is there a issue# or link?

#3017 looks the most relevant. There's also #3423. It doesn't look like these contain everything though - we might have discussed the rest during office hours or in person at Linux Plumbers.

> Should I fix the existing documentation regarding initialization to zero?

Yeah, as far as I can remember we didn't want zero-initialised variables any more, so this'd be good thanks :) 

> > I believe that most languages just adopt a warning when something is never set or never used, which is must simpler.
> 
> I can get behind a WARNING instead of an ERROR (for the reasons you stated). We're just trying to let them know that using something that they haven't explicitly initialized is sort of UB (e.g. with the types I mentioned). If other folks are ok with that change (cc @ajor) I think it should go into the new release branch and master.

I think @amscanne's point here wasn't warning vs error, but rather "variable is never assigned anywhere" vs "variable is not assigned in all branches (or some particular subset of branches)".

amscanne:
I think there are some mixed messages here. I would argue strongly that undefined behavior is bad. Hopefully this is uncontroversial, and there is a shared desire to avoid undefined behavior. Therefore, there are two options:

1) Disallow the use of unassigned variables and declare those invalid programs. This is an ERROR. The challenge is capturing that set, ideally without excluding valid programs.

2) Define the semantics for unassigned variables. You may freely choose to add WARNINGs for such cases, but this does not affect whether a program is valid or not.

Right now neither of these are true (if zero semantics are defined then valid programs are rejected, and if they are not valid then invalid programs are accepted), and I'm not sure which is intended to be the case. Based on the documentation, I assumed that it was (2), and that the error was out-of-date. My personal bias is towards (2) because I think it's simplest for the user & to implement, and there's no real performance argument (you're going to need to guarantee initialization for the verifier, so being clever in (1) is unlikely to allow you to generate better code anyways). However, there are precedents for both decisions in different languages (e.g. most languages for (1) and rust for (2)).

However, either way a decision needs to be made that it's (1) or (2), and the following done:

1) Disallow all invalid programs. This could be trivially done by checking that every variable is defined in both branches of an if, at the cost of disallowing some valid programs (example in my comment above).

2) Keep the documented zero value semantics, but downgrade to a WARNING.

None of this is urgent. Perhaps the best thing would be to produce a 1.0 spec that encapsulates the various outstanding RFCs? Having a single spec makes potential inconsistencies clear. Just to give an example, declared maps (and all globals) are likely to require zero value semantics regardless because you have no control over the order probes fire, e.g.:

```
BEGIN { print(@); } END { @ = ntop(0); }
```

jordalgo:
> Keep the documented zero value semantics, but downgrade to a WARNING if you want to keep it (although there would be some false positive rate).

This is my vote for the reasons stated above. Also just feels like we should do our best not to add more friction for a tool like bpftrace so opting for allowing some invalid programs rather than possibly rejecting valid ones seems like the better option. I also agree this is not urgent so we can wait for other maintainers to add their input here or add it as a topic for the next office hours.

> Perhaps the best thing would be to produce a 1.0 spec that encapsulates the various outstanding RFCs?

Let's discuss this offline. We already have many documents floating around that few are reading, editing or commenting on. We can consolidate but I want to understand better what you mean by "spec".

viktormalik:
FWIW, even though at the office hours I suggested that using ERROR instead of WARNING is better for future backwards compatibility breakage, after reading (and hopefully understanding) the entire discussion here, I'm in favor of

> Keep the documented zero value semantics, but downgrade to a WARNING if you want to keep it (although there would be some false positive rate).

In was never a fan of languages not zeroing out unintialized variables so to me, this is the most intuitive behaviour. Adding a WARNING (even though it was false in few corner cases) sounds like a nice addition to prevent user confusion.

And, we already have zero value semantics for maps (as per @amscanne's example above) so it makes sense to keep it consistent for variables as well.

jordalgo:
@danobi @fbs - Any opinions/thoughts on this?

@ajor - Can you be convinced? I think a WARNING is good enough at conveying to users that there may be an issue with some un-assigned types while at the same time allowing to slowly improve semantic_analyser checking and not blocking potentially valid scripts.

jordalgo:
@amscanne - I talked with @ajor offline and he's good with adopting zero value semantics and changing the current ERRORs into WARNINGs. Let me know if you want to update this PR otherwise, I can open a new one that steals some of your changes.

danobi:
I need to give this thread a read still. But to add some color, some 0 values are kinda meaningless. For example the timestamp type. It's really an offset from an opaque point in time (by default nsecs since boot). And runtime will resolve that into a localized timestamp. 0 for that is kinda invalid.

jordalgo:
> And runtime will resolve that into a localized timestamp. 0 for that is kinda invalid.

@danobi I mentioned some of these problematic types above (it's a long thread) but ultimately I think a WARNING is doing enough of a job to let users know of UB and doesn't get in the way of potentially valid scripts.

amscanne:
I will revisit this and cleanup after the visitor refactor. I'd like to keep the meat of the change here (simplify the logic in the semantic analyzer slightly) and just capture in a simple pass just for one purpose (e.g. WarnUninitializedPass).

jordalgo:
Let's land this instead: https://github.com/bpftrace/bpftrace/pull/3832

PR #3671, opened 2024-12-30
runtime: engine: Prevent hangs after sending SIGTERM

danobi:
SIGTERM only sets BPFtrace::exitsig_recv which the event loop reads. If bpftrace is not in the event loop (for example computing wildcarded probes), SIGTERM will not be immediately respected and will instead hang until the blocking work is done.

To better respect TIMEOUT directive, we must limit the amount of time the engine actually waits.

Also fix up other subprocess.communicate() calls to retry after a timeout. The python docs say this is required for reliable output fetching.

##### Checklist

### DIFF
--- a/tests/runtime/engine/runner.py
+++ b/tests/runtime/engine/runner.py
@@ -420,7 +420,16 @@ class Runner(object):
             if p:
                 if p.poll() is None:
                     os.killpg(os.getpgid(p.pid), signal.SIGTERM)
-                output += p.stdout.read()
+                try:
+                    # SIGTERM only sets the exit flag in bpftrace event loop.
+                    # If bpftrace is stuck somewhere else, it won't exit. So we
+                    # have to set a timeout here to prevent hangs.
+                    output += p.communicate(timeout=1)[0]
+                except subprocess.TimeoutExpired:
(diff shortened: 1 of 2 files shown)


jordalgo:
Change looks fine. I guess I would have expected some test TIMEOUTs also needing adjustment with this change if it was an issue.

danobi:
Fortunately I don't think we don't have any overtly buggy tests relying on TIMEOUT leniency. But when run in a VM in #3657 this starts to happen sometimes.

danobi:
I'm reading thru the docs on https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate, and I think this PR needs a change to drain output after SIGKILL. Lemme update with ath

jordalgo (approved):
Ok, let's give this a shot.

danobi:
I've added a selftest that hangs indefinitely on master but not after this PR. Was probably about time we started writing tests for the test engine :)

jordalgo:
> I've added a selftest that hangs indefinitely on master but not after this PR. 

Nice!

> Was probably about time we started writing tests for the test engine :)

Do you mind also writing a test for the tests that's testing the test engine? 🐢 🐢 🐢

PR #3674, opened 2024-12-31
tests: Make call.skboutput more reliable

danobi:
Loopback device may not always be up. On most systems it's there, but in
a minimal VM, it could not be.  Make this test more reliable by setting
up a loopback device in a new netns.

The choice was made to not add NEW_NETNS (like with NEW_PIDNS), as netns
stuff is unlikely to be reused in the future.

Also make the test case easier to debug by always printing the return
value.

##### Checklist

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -75,10 +75,13 @@ Each runtime testcase consists of multiple directives. In no particular order:
    converting the output and the file to a dict (thus ignoring field order).
 * `TIMEOUT`: The timeout for the testcase (in seconds). This field is required.
 * `BEFORE`: Run the command in a shell before running bpftrace. The command
-  will be terminated after testcase is over. Can be used multiple times,
-  commands will run in parallel.
+  will run while bpftrace is running and be terminated after the test case
+  finishes. Can be used multiple times, commands will run in parallel.
 * `AFTER`: Run the command in a shell after running bpftrace (after the probes
   are attached). The command will be terminated after the testcase is over.
+* `SETUP`: Run the command in a shell before the test is run. This differs from
+  the `BEFORE` directive in that setup commands are expected to exit before
+  bpftrace is executed.
 * `CLEANUP`: Run the command in a shell after test is over. This holds any
   cleanup command to free resources after test completes.
 * `MIN_KERNEL`: Skip the test unless the host's kernel version is >= the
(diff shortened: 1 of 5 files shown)


tests/README.md
```diff
@@ -76,9 +76,13 @@ Each runtime testcase consists of multiple directives. In no particular order:
 * `TIMEOUT`: The timeout for the testcase (in seconds). This field is required.
 * `BEFORE`: Run the command in a shell before running bpftrace. The command
   will be terminated after testcase is over. Can be used multiple times,
-  commands will run in parallel.
+  commands will run in parallel. `BEFORE` commands are expected to run while
```
jordalgo, line 79: nit: how about just changing the second sentence to "The command will run while bpftrace is running and be terminated after the test case finishes."

tests/README.md
```diff
@@ -76,9 +76,13 @@ Each runtime testcase consists of multiple directives. In no particular order:
...
+  bpftrace is running.
 * `AFTER`: Run the command in a shell after running bpftrace (after the probes
   are attached). The command will be terminated after the testcase is over.
+* `SETUP`: Run the command in a shell before test is run. This differs from
```
jordalgo, line 83:
```suggestion
* `SETUP`: Run the command in a shell before the test is run. This differs from
```

tests/README.md
```diff
@@ -76,9 +76,13 @@ Each runtime testcase consists of multiple directives. In no particular order:
...
+  `BEFORE` directive in that setup commands are expected to exit before bpftrace
```
jordalgo, line 84:
```suggestion
  the `BEFORE` directive in that setup commands are expected to exit before bpftrace
```

jordalgo:
TBH, I don't love adding the `SETUP` directive only because I don't (at first look) understand how it differs from `BEFORE` but I guess since we have `CLEANUP` and `AFTER` it's ok.

jordalgo (commented):
Few text nits but otherwise LGTM.

danobi:
> TBH, I don't love adding the SETUP directive only because I don't (at first look) understand how it differs from BEFORE but I guess since we have CLEANUP and AFTER it's ok.

I think the main thing is BEFORE/AFTER implies concurrency while SETUP/CLEANUP implies a serialized event. I also generally find it more useful to document intent as precisely as possible.

jordalgo (approved):

PR #3676, opened 2025-01-01
Hide pcap writer lifecycle

danobi:
Right now class BPFTrace owns the pcap writers but the lifecycle is
managed outside of the class in two different places. Simplify the logic
by privatizing the members and moving all management into class BPFTrace.

This also makes skboutput() helper work for AOT. All create_pcaps()
needs is the skboutput async args. And those are stored in
RequiredResources, which AOT shim will load back up.

##### Checklist


jordalgo (approved):
> This also makes skboutput() helper work for AOT.

Nice!

PR #3677, opened 2025-01-01
tests: runtime: engine: Get rid of attachment timeout

danobi:
Runtime tests have the explicit TIMEOUT directive in each test case. But there is also a hidden timeout for attachment. Previously it was 10s for waiting to attach that did not count towards TIMEOUT directive.

IMO that's confusing - better to just have a single timeout that does what it suggests. If any tests are relying on that additional 10s, just increase their timeout.

##### Checklist


jordalgo (approved):
SGTM

PR #3680, opened 2025-01-02
test: Ensure bpftrace_test runs in `nix develop` environment

amscanne:
This change cleans up minor issues with testing. One is that the probe expansion `/bin/*sh` implicitly depends on multiple paths matching in order to call the mocks. A small code change prevents inline expansion until later on when this will follow the correct code path.

Another introduces local testprogs to replace the use of `/bin/ls` and `/bin/sleep` (which are not guaranteed to be available). These are used in the child and procmon tests, as well as the runtime tests.

##### Checklist

- [-] Language changes are updated in `man/adoc/bpftrace.adoc`

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -3249,7 +3249,11 @@ void SemanticAnalyser::visit(AttachPoint &ap)
                                  << "' does not exist or is not executable";
         break;
       case 1:
-        ap.target = paths.front();
+        // Replace the glob at this stage only if this is *not* a wildcard,
+        // otherwise we rely on the probe matcher. This is not going through
+        // any interfaces that can be properly mocked.
+        if (ap.target.find("*") == std::string::npos)
+          ap.target = paths.front();
         break;
       default:
         // If we are doing a PATH lookup (ie not glob), we follow shell
@@ -3277,12 +3281,12 @@ void SemanticAnalyser::visit(AttachPoint &ap)
                                    << "' does not exist or is not executable";
           break;
         case 1:
-          ap.target = paths.front();
+          // See uprobe, above.
+          if (ap.target.find("*") == std::string::npos)
+            ap.target = paths.front();
           break;
         default:
-          // If we are doing a PATH lookup (ie not glob), we follow shell
-          // behavior and take the first match.
-          // Otherwise we keep the target with glob, it will be expanded later
+          // See uprobe, above.
           if (ap.target.find("*") == std::string::npos) {
             LOG(WARNING, ap.loc, out_)
                 << "attaching to usdt target file '" << paths.front()
--- a/tests/child.cpp
+++ b/tests/child.cpp
@@ -21,11 +21,30 @@ namespace bpftrace::test::child {
 
 using ::testing::HasSubstr;
 
-#define TEST_BIN "/bin/ls"
-#define TEST_BIN_ERR "/bin/ls /does/not/exist/abc"
-#define TEST_BIN_SLOW "/bin/sleep 10"
+class childproc : public ::testing::Test {
+protected:
+  void find(std::string &out, const char *path)
+  {
+    std::error_code ec;
+    auto self = std::filesystem::read_symlink("/proc/self/exe", ec);
+    ASSERT_FALSE(ec);
+    auto parent_dir = self.parent_path();
+    out = parent_dir / std::filesystem::path(path);
+  }
 
-TEST(childproc, exe_does_not_exist)
+  void SetUp() override
+  {
+    find(TEST_BIN, "testprogs/true");
+    find(TEST_BIN_ERR, "testprogs/false");
+    find(TEST_BIN_SLOW, "testprogs/wait10");
+  }
+
+  std::string TEST_BIN;
+  std::string TEST_BIN_ERR;
+  std::string TEST_BIN_SLOW;
+};
+
+TEST_F(childproc, exe_does_not_exist)
 {
   try {
     ChildProc child("/does/not/exist/abc/fed");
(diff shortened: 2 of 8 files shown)


tests/child.cpp
```diff
@@ -20,11 +20,30 @@ namespace bpftrace::test::child {
...
 
-#define TEST_BIN "/bin/ls"
-#define TEST_BIN_ERR "/bin/ls /does/not/exist/abc"
-#define TEST_BIN_SLOW "/bin/sleep 10"
+class childproc : public ::testing::Test {
+protected:
+  void find(std::string &out, const char *path)
+  {
+    std::error_code ec;
+    auto self = std::filesystem::read_symlink("/proc/self/exe", ec);
```
danobi, line 28: See `src/filesystem.h`. Not sure if this is still necessary since we c++20 now, but probably good to keep it consistent and clean it all up in one PR.
```suggestion
    auto self = std_filesystem::read_symlink("/proc/self/exe", ec);
```

tests/child.cpp
```diff
@@ -20,11 +20,30 @@ namespace bpftrace::test::child {
...
+    ASSERT_FALSE(ec);
+    auto parent_dir = self.parent_path();
+    out = parent_dir / std::filesystem::path(path);
```
danobi, line 31:
```suggestion
    out = parent_dir / std_filesystem::path(path);
```

tests/procmon.cpp
```diff
@@ -26,7 +26,12 @@ TEST(procmon, no_such_proc)
 
 TEST(procmon, child_terminates)
 {
-  auto child = getChild("/bin/ls");
+  std::error_code ec;
+  auto self = std::filesystem::read_symlink("/proc/self/exe", ec);
```
danobi, line 30:
```suggestion
  auto self = std_filesystem::read_symlink("/proc/self/exe", ec);
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3181,7 +3181,11 @@ void SemanticAnalyser::visit(AttachPoint &ap)
                                  << "' does not exist or is not executable";
         break;
       case 1:
-        ap.target = paths.front();
+        // Replace the glob at this stage only if this is *not* a wildcard,
+        // otherwise we rely on the probe matcher. This is not going through
+        // any interfaces that can be properly mocked.
+        if (ap.target.find("*") == std::string::npos)
+          ap.target = paths.front();
```
danobi, line 3188: What's the implication of this change? Fixes a broken test on a more sterile system?

amscanne: This will resolve the wildcard ahead of where the mocking takes place. The wildcard is handled perfectly fine later, so I've just made this non-destructive (match using PATH if needed, but don't remove the wildcard that will be expanded later). This matches the semantics below as well, which defers the expansion when there are multiple matches.

tests/child.cpp
amscanne: I'd rather just resolve this bit of debt. That header doesn't make sense anymore. Will send  #3722 for review shortly.

tests/child.cpp
```diff
@@ -20,11 +20,30 @@ namespace bpftrace::test::child {
 
 using ::testing::HasSubstr;
 
-#define TEST_BIN "/bin/ls"
-#define TEST_BIN_ERR "/bin/ls /does/not/exist/abc"
-#define TEST_BIN_SLOW "/bin/sleep 10"
+class childproc : public ::testing::Test {
+protected:
+  void find(std::string &out, const char *path)
```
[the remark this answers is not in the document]
amscanne: We have the tests configured to fail by raising an exception where there is an assertion failure (via `tests/main.cc`) but you can only use `ASSERT_...` in a void function (like the test main). I figured since this was just a local helper for setup, these were the best ergnomics. So I change it, but it would be something like this:

```
std::string find(const char *path) {
   ...
   if (ec) {
      return "";
   }
}
void SetUp() override {
  TEST_BIN = find("testprogs/true");
  ASSERT(!TEST_BIN.empty());
  ...
}
```

Happy to do so, I just find the current one cleaner.

tests/semantic_analyser.cpp
```diff
@@ -355,7 +355,7 @@ TEST(semantic_analyser, builtin_functions)
   test("kprobe:f { kstack(1) }");
   test("kprobe:f { ustack(1) }");
   test("kprobe:f { cat(\"/proc/uptime\") }");
-  test("uprobe:/bin/bash:main { uaddr(\"glob_asciirange\") }");
+  test("uprobe:/bin/sh:main { uaddr(\"glob_asciirange\") }");
```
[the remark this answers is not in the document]
amscanne: Doesn't always exist => did not exist in our own NIX flake environment (for whatever reason) :)

Yes, I don't think `/bin/sh` is specifically mandated by POSIX, but I've not seen a system without it (e.g. https://serverfault.com/questions/865874/bin-sh-vs-bin-bash-for-maximum-portability). I'm not using a testprog here because we're specifically checking some external function (`glob_asciirange`), and my goal is the minimum set of changes to make the tests pass in our own NIX flake. It basically means making it a different test.

I agree that it would probably be better to depend on nothing outside on our binaries, but if you're going to look at one external binary then `/bin/sh` is probably the right one.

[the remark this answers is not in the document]
viktormalik: Ok, it makes sense to minimize the amount of changes here.

tests/child.cpp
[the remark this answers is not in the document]
viktormalik: Right, that makes sense, I didn't know the constraint for `ASSERT`. Let's keep it as it is.

viktormalik (approved):
Nice!

PR #3682, opened 2025-01-03
Rm hard-coded number of semantic_analyser passes

jordalgo:
Instead of doing 10 passes by default for every
script, have the places that know that another
pass is required increment a counter (num_unresolved_).

We can then use this number to see if we need
an additional pass AND make sure we're making progress
(resolving types) with additional passes.

For example, in this valid script:
```
interval:s:1 { @j = @i; @i = @h; @h = @g; @g = @f; @f = @e; @e = @d; @d = @c; @c = @b; @b = @a; }
interval:s:1 { @a = 1; }'
```
each pass resolves the type for a single map,
starting at the end of the chain. Without this change,
this script will error because it will exceed
the hard-coded 10 passes.

And in the case of this error script:
```
BEGIN { @a = @b; }
```
We stop making progress after 2 passes
and know we can stop making additional passes.

In local testing I mostly saw 2 or 3 passes for all
scripts, which includes the first and final pass.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -1726,7 +1726,7 @@ void SemanticAnalyser::visit(Map &map)
         update_current_key(key->second, new_key_type);
         validate_new_key(key->second, new_key_type, map.ident, map.loc);
       } else {
-        if (!key->second.IsNoneTy() && is_final_pass()) {
+        if (!key->second.IsNoneTy()) {
           LOG(ERROR, map.loc, err_)
               << "Argument mismatch for " << map.ident << ": "
               << "trying to access with no arguments when map expects "
@@ -1748,9 +1748,12 @@ void SemanticAnalyser::visit(Map &map)
   if (search_val != map_val_.end()) {
     map.type = search_val->second;
   } else {
-    if (is_final_pass()) {
+    // If there is no record of any assignment after the first pass
+    // then it's safe to say this map is undefined
+    if (!is_first_pass()) {
       LOG(ERROR, map.loc, err_) << "Undefined map: " << map.ident;
     }
+    pass_tracker_.inc_num_unresolved();
     map.type = CreateNone();
   }
 
@@ -2917,12 +2920,6 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
           << "Array type mismatch: " << map_type << " != " << expr_type << ".";
     }
   }
-
-  if (is_final_pass()) {
-    if (type.IsNoneTy())
-      LOG(ERROR, assignment.expr->loc, err_)
-          << "Invalid expression for assignment: " << type;
-  }
 }
 
 void SemanticAnalyser::visit(AssignVarStatement &assignment)
@@ -2950,7 +2947,11 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
     if (storedTy.IsNoneTy()) {
       storedTy = assignTy;
     } else if (!storedTy.IsSameType(assignTy)) {
-      type_mismatch_error = true;
+      if (!assignTy.IsNoneTy() || is_final_pass()) {
+        type_mismatch_error = true;
+      } else {
+        pass_tracker_.inc_num_unresolved();
+      }
     } else if (assignTy.IsStringTy()) {
       if (foundVar.can_resize) {
         update_string_size(storedTy, assignTy);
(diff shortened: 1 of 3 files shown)


jordalgo:
I'm semi-frightened that this just mostly works. I will spend more time going through the changes and adding more tests where I see holes but overall this is a nice run time win (10 required passes to mostly 2 or 3).

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2904,12 +2907,6 @@ void SemanticAnalyser::visit(AssignMapStatement &assignment)
           << "Array type mismatch: " << map_type << " != " << expr_type << ".";
     }
   }
-
-  if (is_final_pass()) {
```
jordalgo, line 2908: This was an extra, unneeded check.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2937,7 +2934,11 @@ void SemanticAnalyser::visit(AssignVarStatement &assignment)
     if (storedTy.IsNoneTy()) {
       storedTy = assignTy;
     } else if (!storedTy.IsSameType(assignTy)) {
-      type_mismatch_error = true;
+      if (!assignTy.IsNoneTy() || is_final_pass()) {
```
jordalgo, line 2937: This fails on master (unit test added below)
```
kprobe:f { $b = "hi"; $b = @b; } kprobe:g { @b = "bye"; }
```
but is fixed in this change.

jordalgo:
This is ready for a look.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1739,9 +1739,12 @@ void SemanticAnalyser::visit(Map &map)
   if (search_val != map_val_.end()) {
     map.type = search_val->second;
   } else {
-    if (is_final_pass()) {
+    // If there is no record of any assignment after the first pass
+    // then it's safe to say this map is undefined
+    if (!is_first_pass()) {
       LOG(ERROR, map.loc, err_) << "Undefined map: " << map.ident;
     }
+    num_unresolved_++;
```
danobi, line 1747: Wrap this in a private method? Would make it symmetric with `is_final_pass()`/`is_first_pass()`. Also makes contract a bit more obvious. eg. you cannot decrement a function call.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3462,25 +3463,41 @@ void SemanticAnalyser::visit(Program &program)
...
+
+    if (num_unresolved_ > 0 &&
+        (last_num_unresolved == 0 || num_unresolved_ < last_num_unresolved)) {
+      // If we're making progress, keep making passes
+      last_num_unresolved = num_unresolved_;
+    } else {
+      is_final_pass_ = true;
+    }
+
+    pass_++;
```
danobi, line 3494: Maybe `LOG(BUG)` if pass_ goes above 1000 or something? In case there's a resolution but it'd be better to fail hard than spin forever. Failing louder, esp. in automated environments is probably better here.

src/ast/passes/semantic_analyser.h
```diff
@@ -110,8 +110,9 @@ class SemanticAnalyser : public Visitor {
   std::ostream &out_;
   std::ostringstream err_;
   int pass_;
-  const int num_passes_ = 10;
   bool listing_;
+  bool is_final_pass_ = false;
+  int num_unresolved_;
```
danobi, line 115: Building on above comment, break these two vars out into its own grouping and comment to not directly modify these?

danobi (commented):
very nice! love how the diff is so small for such a (probably) big perf impact.

src/ast/passes/semantic_analyser.h
danobi: Would be great to find some way to have an "even private-er" way to hide these vars. A small helper class?

danobi: Also defensive `num_unresolved_ = 0` would be good

src/ast/passes/semantic_analyser.cpp
jordalgo: Actually, I think the progress checking functionality should handle this case. Even if we start with a HUGE number of unresolved types, it needs to make progress or else it exists with an error.

jordalgo:
@danobi Good suggestion. I made a little private class to provide better encapsulation

src/ast/passes/semantic_analyser.h
```diff
@@ -105,11 +105,47 @@ class SemanticAnalyser : public Visitor {
...
+    }
+    void inc_num_unresolved()
+    {
+      num_unresolved_++;
+    }
+    void reset_num_unresolved()
+    {
+      num_unresolved_ = 0;
+    }
+    int get_num_unresovled() const
```
danobi, line 126:
```suggestion
    int get_num_unresolved() const
```

src/ast/passes/semantic_analyser.h
```diff
@@ -105,11 +105,47 @@ class SemanticAnalyser : public Visitor {
   int analyse();
 
 private:
+  class PassTracker {
```
danobi, line 108: nit: Move out of line of SemanticAnalyser definition? Just to keep the class stuff more compact

src/ast/passes/semantic_analyser.cpp
danobi: Ah, yeah. Progress checks should be catching it.

danobi (approved):
some nits; lgtm otherwie

src/ast/passes/semantic_analyser.h
jordalgo: spelling is importnat

Issue #3075, opened 2024-03-19
Add non-C test programs


ajor:
We currently have a range of C-language programs for testing bpftrace against: https://github.com/bpftrace/bpftrace/tree/19239ad80d0daac1f064ca2e441b2d54120a2f05/tests/testprogs

We should also have a few test programs written in the other languages that we claim to support. I think these would be:
- C++
- Go
- Rust

amscanne:
I'd be happy to do this, but I think pulling in all the separate toolchains to the native build will bloat everything quite a lot for some quick hello-world probe tests. Any objections to having this configuration separate workflows? (So you're just testing against some canonical pre-existing application image, rather than needing all the toolchains from scratch)

danobi:
It looks like already have c++: https://github.com/bpftrace/bpftrace/blob/master/tests/testprogs/uprobe_test_cxx.cpp

Do we have a good idea of the test cases we want to add? 

In general I'd prefer to not get too fancy w/ GH workflows and stuff. We do all our heavy lifting with nix and it's great for reproducibility. Should be ok IMO to add in `rustc` or `go` packages into flake.nix and then have cmake check for those binaries. That way CI will always have it and developers can opt-into nix and get it or not install those packages and just skip the relevant tests.

PR #3687, opened 2025-01-07
test: Add skeleton for additional languages/runtimes

amscanne:
This provides simple rust and Go programs, that are compiled if the toolchains are available. The toolchains are added to the flake, so these should be built. The runtime tests will be skipped if they are not built.

Updates #3075 

##### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -140,6 +140,9 @@
                   strace
                   unixtools.ping
                   util-linux
+                  # For runtime tests
+                  rustc
+                  go
                 ] ++ pkg.nativeBuildInputs ++ pkg.buildInputs;
 
                 # Some hardening features (like _FORTIFY_SOURCE) requires building with
--- a/tests/runtime/languages
+++ b/tests/runtime/languages
@@ -1,0 +1,16 @@
+# This only tests for the presence of mangled rust probe names. We need to add
+# support for automatically unmangling the names (and include the 'rust'
+# language type).
+NAME uprobes - list rust uprobes (mangled)
+REQUIRES testprogs/hello_rust
+RUN {{BPFTRACE}} -l 'uprobe:./testprogs/hello_rust:*'
+EXPECT_REGEX uprobe:./testprogs/hello_rust.*fun1
+
+# This only tests for the presence of the builtin symbol `runtime.schedule`.
+# Argument unpacking and other things will *not* be supported in Go for a
+# while, since the calling convention is different. But useful analysis can
+# still be done with the basics.
+NAME uprobes - list Go uprobes (no arguments)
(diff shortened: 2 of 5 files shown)


danobi (commented):
I think it should be possible to avoid package management for both these languages right? IMO it'd be cleaner to just have `foo.go` and `bar.rs` style files in `tests/testprogs`. 

It'd be consistent with c and c++ testprog handling. And less chance of external deps being introduced -- all internet downloads should happen in a download step (eg `nix develop`) and not in the build step.

`go build foo.go` and `rustc bar.rs` seem to work exactly as expected.

amscanne:
I can go that way. My goal was just to produce binaries that are as canonical as possible (using the standardized paths for the toolchains), but I suspect you're right that the raw tools are good enough for these two cases.

amscanne:
Done. This is a nice simplification.

danobi:
Missing testprogs?

danobi (commented):
lgtm!

any thoughts on explicitly turning optimizations off?

https://doc.rust-lang.org/rustc/codegen-options/index.html#opt-level

https://pkg.go.dev/cmd/compile

amscanne:
Hm, it seems to be the default for rust, and I'm somewhat opposed to doing this for Go -- just because I don't think anyone outside the Go team has ever used `-N` :) It seems like you actually need to bypass the build tool to even use that flag, e.g. `go build -gcflags=-N`. The Go binary has debug info by default and non-inlined symbols are defined.

I think there are lots of questions on what support level is provided for various languages, and documenting the different paths. This is really just a "it doesn't immediately blow up" test. (No new functionality or user documentation.)

As an aside, there's lots of neat future work that could be done here (esp. for languages where there's a standardized runtime and therefore more amenable to standardized tools), but I just don't want to get ahead of ourselves. For example, with any dwarf debug info, you can already import as BTF and get those types for your probes. Just playing around, but amazingly the following "works" for Go:
* `pahole --btf_encode_detached hello_go`
* `BPFTRACE_BTF=hello_go.btf sudo -E bpftrace -e 'uprobe:./hello_go:runtime.schedule { $schedt = (struct runtime.schedt*)(uaddr("runtime.sched")); printf("nmidle: %d, npidle: %d\n", $schedt->nmidle, $schedt->npidle.value) }'`

BUT the calling convention for Go is different, and there's bound to be a bunch of other stuff that will break in weird ways (unless this is already magically resolved via dwarf info?). I suspect that once there are clear project rails, these would be good to solicit as cohesive projects. This is just the babiest of steps.

danobi (approved):

danobi:
Cool demo! Something cooperative with language runtimes would be quite interesting.

PR #3688, opened 2025-01-07
Add support for rust demangling

amscanne:
Adds basic support for rust demangling, similar to existing C++ demangling. Note that unlike C++, there is no good stdlib fallback for the V0 mangling scheme, so this is not generally supported.

Updates #3687 #3075 

##### Checklist

### DIFF
--- a/src/cxxdemangler/cxxdemangler_stdlib.cpp
+++ b/src/cxxdemangler/cxxdemangler_stdlib.cpp
@@ -4,9 +4,15 @@
 
 namespace bpftrace {
 
-char* cxxdemangle(const char* mangled)
+std::string cxxdemangle(const char *mangled)
 {
-  return abi::__cxa_demangle(mangled, nullptr, nullptr, nullptr);
+  std::string s;
+  char *d = abi::__cxa_demangle(mangled, nullptr, nullptr, nullptr);
+  if (!d)
+    return s;
+  s = std::string(d);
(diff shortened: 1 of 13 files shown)


jordalgo:
Looks like there is an issue on llvm 14 - I'm happy to drop support for this version (which we discussed offline a bit) but then you just might be waiting a bit.

src/cxxdemangler/cxxdemangler_stdlib.cpp
```diff
@@ -4,9 +4,15 @@
 
 namespace bpftrace {
 
-char* cxxdemangle(const char* mangled)
+std::string cxxdemangle(const char *mangled)
```
jordalgo, line 7: Nit: feels like the updates to the cxxdemangle paths should be a separate commit.

tests/runtime/languages
```diff
@@ -6,6 +6,12 @@ REQUIRES testprogs/hello_rust
 RUN {{BPFTRACE}} -l 'uprobe:./testprogs/hello_rust:*'
 EXPECT_REGEX uprobe:./testprogs/hello_rust.*fun1
 
+# This test adds support for mangled names.
```
jordalgo, line 9: This comment seems odd/unneccessary.

danobi:
Closing due to inactivity. Feel free to reopen later.

PR #3692, opened 2025-01-08
Fix json output for none type

jordalgo:
For non-json output we print an empty string
("") for a none type, however for json output
we need to "quote" this empty string so we
don't get invalid json e.g.

```
bpftrace -e 'kprobe:vfs_write {print(*(struct file *)arg0); exit()}' -f json
```

Before snippet:
```
{"type": "value", "data": { "f_u": , "f_path": { "mnt": 18446612686587658272, ...
```

After snippet:
```
{"type": "value", "data": { "f_u": "", "f_path": { "mnt": 18446612686587658272, ...
```

This is ultimately caused by bpftrace treating
anon union and structs as none types.
https://github.com/bpftrace/bpftrace/blob/master/src/btf.cpp#L389

Issue: https://github.com/bpftrace/bpftrace/issues/3543

##### Checklist


jordalgo:
@fbs That's what I had originally but I think because null is an actual state for some of these fields (like in the example I provided above) I didn't want to confuse users. Our none type is more like "this is unsupported, we don't know what to do here". It also matches the current none output for plain text.

tests/runtime/json-output
```diff
@@ -178,3 +178,10 @@ EXPECT {"type": "value", "data": 333}
 NAME enum_symbolized_printf
 RUN {{BPFTRACE}} -q -f json -e 'enum Foo { ONE = 1, TWO = 2, OTHER = 99999 }; BEGIN { printf("%d %s %d %s %d %s\n", ONE, ONE, TWO, TWO, OTHER, OTHER); exit() }'
 EXPECT {"type": "printf", "data": "1 ONE 2 TWO 99999 OTHER\n"}
+
+# Just check that the json is well formatted
+# Note: if we fix the underlying issue we might need to change/delete this test
+NAME none type
+RUN {{BPFTRACE}} -q -f json -e 'kprobe:vfs_write {print((*(struct file *)arg0).f_owner); exit()}' | python3 -c 'import sys,json; print(json.load(sys.stdin)["type"])'
```
danobi, line 185: kprobe:vfs_write is not guaranteed to fire, especially on a quiet host. Might need a call to `testprog/syscall`. Or a more reliable attachpoint.

danobi, line 185: also, why is this expected to be `none` type?

jordalgo: Because we don't yet support this anon struct btf type (as per the summary)

jordalgo: Good point. Let me add a syscall in a test prog.

danobi: Oh sorry, missed that part of summary.

But `f_owner` is not an anon struct or union?

```
$ sudo bpftrace -lv 'struct file'
struct file {
        union {
                struct callback_head f_task_work;
                struct llist_node f_llist;
                unsigned int f_iocb_flags;
        };
        spinlock_t f_lock;
        fmode_t f_mode;
        atomic_long_t f_count;
        struct mutex f_pos_lock;
        loff_t f_pos;
        unsigned int f_flags;
        struct fown_struct f_owner;
        const struct cred *f_cred;
        struct file_ra_state f_ra;
        struct path f_path;
        struct inode *f_inode;
        const struct file_operations *f_op;
        u64 f_version;
        void *f_security;
        void *private_data;
        struct hlist_head *f_ep;
        struct address_space *f_mapping;
        errseq_t f_wb_err;
        errseq_t f_sb_err;
};```

jordalgo: `f_owner` is a struct that contains an anon struct `kuid_t`

danobi: Ah, I see. This seems unreliable in the long term, though. Can you do a user defined C type and try printing that instead?

danobi:
Bigger picture: seems like exposing `none` type to user is always a mistake. Should we error out the access / print if it's not actually possible to print it? Printing empty string seems... wrong

tests/runtime/json-output
jordalgo: oooooookkkayayyyyy

jordalgo:
>  IMO exposing none type to user is always a mistake. Should we error out the access / print if it's not actually possible to print it? Printing empty string seems... wrong

I'm not sure what the right answer is here. Like for the example in summary, do we really want to error if only some of the fields in the struct are none/not-supported. I'll try to think on it a bit.

danobi:
> > IMO exposing none type to user is always a mistake. Should we error out the access / print if it's not actually possible to print it? Printing empty string seems... wrong
> 
> I'm not sure what the right answer is here. Like for the example in summary, do we really want to error if only some of the fields in the struct are none/not-supported. I'll try to think on it a bit.

I was thinking the error criteria should be "using an unsupported field". Where "using" could mean accessing. But there a lot of edge cases. In this case, it's printing the entire struct. Another edge case could be using the entire struct as a map key.

This could get tricky...

danobi (approved):
Mind following up with an issue about not exposing none types to the user?

jordalgo:
> Mind following up with an issue about not exposing none types to the user?

Done. Still needs more fleshing out though: https://github.com/bpftrace/bpftrace/issues/3696

PR #3698, opened 2025-01-09
Accept ASTContext& for Visitors and pass manager entries

amscanne:
In the future, it will be useful to allow the `Visitor` to rewrite parts of the AST in a consistent way. This requires access to the `ASTContext` in order to allocate new nodes. Standardize this as required to instantiate new passes.

This is plumbed through the pass, and allows us to generally remove the use of `Program` since it is already specified in the `ASTContext`. This is a minor cleanup in preparation for restructuring drafted in #3636.

##### Checklist

- ~~[ ] The new behaviour is covered by tests~~

### DIFF
--- a/src/ast/pass_manager.cpp
+++ b/src/ast/pass_manager.cpp
@@ -8,12 +8,12 @@
 namespace bpftrace::ast {
 
 namespace {
-void print(Node *root, const std::string &name, std::ostream &out)
+void print(PassContext &ctx, const std::string &name, std::ostream &out)
 {
   out << "\nAST after: " << name << std::endl;
   out << "-------------------\n";
-  ast::Printer printer(out);
-  printer.print(root);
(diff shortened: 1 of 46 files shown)


danobi (commented):
Overall, looks quite reasonable to me

I think you missed one AST node creation conversion in sematic_analyser:
```
Unop *deref_expr = new Unop(Operator::MUL, ptr_expr, ptr_expr->loc);
```


also, 
> allows us to generate remove the use

wording typo?

amscanne:
Ah, good catch. Since this is independent, I'll just send #3714 to fix that case. (Don't want to mix the refactor with a change.)

danobi:
Mind updating the commit msg? That's the history that actually matters (IM)

amscanne:
Whoops, I thought I did. Wrong branch name.

danobi (approved):

PR #3699, opened 2025-01-09
Prefer static types over double dynamic dispatch

amscanne:
Currently, the visitor used for all the AST validation, rewriting and code
generation has two separate layers of virtual function indirection (and
occasionally makes use of a manually-crafted vtable-like hashtable keyed by the
RTTI typeids for hand-rolled dynamic dispatch). This is more complex than
necessary, and limits the opportunities for the compiler to optimize these
passes by e.g. folding a visitor into only the things that it actually cares
about visiting.

This change is intended to be non-functional, and makes use of the C++
Curiously Recurring Template Pattern (CRTP) to make each visitor implementation
fully non-virtual and self-contained. This commit does not quite eliminate the
need for RTTI at all, which is left to a subsequent change just to keep each
commit manageable. The `Expression*` and `Statement*` types now have explicit
types that are dispatched, which will be subsequently eliminated.

This is a follow-on from https://github.com/bpftrace/bpftrace/pull/3698.

##### Checklist

- ~~[ ] The new behaviour is covered by tests~~


jordalgo:
Looks like this has a few conflicts that need resolving :)

amscanne:
Just a rebase due to the manual stacking. No conflicts on this one.

src/ast/passes/callback_visitor.h
```diff
@@ -2,50 +2,23 @@
...
-  void visit(Unroll &unroll) override;
-  void visit(While &while_block) override;
-  void visit(Jump &jump) override;
-  void visit(Predicate &pred) override;
-  void visit(AttachPoint &ap) override;
-  void visit(Probe &probe) override;
-  void visit(Config &config) override;
-  void visit(Block &block) override;
-  void visit(Program &program) override;
+  void preVisit(Node &node)
```
jordalgo, line 18: It seems like this is only really used by the node counting (unless I'm missing something). Should we be explicit with the naming and make it more generic in the future if needed?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1490,11 +1490,14 @@ void CodegenLLVM::visit(Call &call)
     b_.CreateStore(b_.CreateGetNs(TimestampMode::boot, call.loc), time_addr);
 
     auto &arg_skb = *call.vargs.at(1);
-    arg_skb.accept(*this);
+    // FIXME(amscanne): this is a likely bug, as we are note appropriately
```
jordalgo, line 1493:
```suggestion
    // FIXME(amscanne): this is likely a bug, as we are not appropriately
```

jordalgo: You could also create an issue for these two lifetime bugs and add it to the comment.

src/ast/visitor.h
```diff
@@ -0,0 +1,325 @@
...
+  }
+  template <typename T>
+  R visit(T *ptr)
+  {
+    if (ptr)
+      return _visit(*ptr);
+    return _def();
+  }
+
+  // There are the runtime-type adaptors that are currently required for
```
jordalgo, line 243:
```suggestion
  // These are the runtime-type adaptors that are currently required for
```

src/ast/visitor.h
```diff
@@ -0,0 +1,325 @@
...
+  // Expression and Statement, but can be removed by encoding this type
+  // information into the AST directly.
+  template <typename T, typename... Ts>
+  R tryVisit(Node *node)
+  {
+    if (auto *t = dynamic_cast<T>(node))
+      return _visit(*t);
+    if constexpr (sizeof...(Ts) != 0) {
```
jordalgo, line 251: So the compiler automatically unwinds all this? Neat.

src/ast/passes/codegen_llvm.h
```diff
@@ -280,6 +280,26 @@ class CodegenLLVM : public Visitor {
...
   GlobalVariable *DeclareKernelVar(const std::string &name);
 
+  template <typename T>
+  ScopedExprDeleter accept(T *node)
+  {
+    // This wraps the visit by essentially stacking return values in this
+    // object, and popping them. This should be converted in the future to
+    // using the structured visit (e.g. Visitor<CodegenLLVM, SomeThing>. The
+    // special handling of Expression and Statement will also be removed once
+    // we are no longer relying on RTTI for object type information.
```
jordalgo, line 290:
```suggestion
    // we are no longer relying on RTTI for objects.
```

src/ast/passes/collect_nodes.h
```diff
@@ -14,36 +14,36 @@ namespace bpftrace::ast {
...
-    node.accept(*this);
-  }
   const std::vector<std::reference_wrapper<NodeT>> &nodes() const
   {
     return nodes_;
   }
 
-private:
-  void visit(NodeT &node) override
+  using Visitor<CollectNodes<NodeT>>::visit;
```
jordalgo, line 30: JC, why is this `using` neccessary?

src/ast/passes/printer.cpp
```diff
@@ -364,15 +358,20 @@ void Printer::visit(For &for_loop)
   }
 
   out_ << indent << " decl\n";
-  print(for_loop.decl);
+  ++depth_;
```
jordalgo, line 361: Why were these additional depth inc/dec added?

src/ast/passes/printer.cpp
```diff
@@ -416,24 +411,14 @@ void Printer::visit(AttachPoint &ap)
...
-    ap->accept(*this);
-  }
+  visit(probe.attach_points);
 
   ++depth_;
-  if (probe.pred) {
-    probe.pred->accept(*this);
-  }
-  probe.block->accept(*this);
+  if (probe.pred)
```
jordalgo, line 419: JC, if there are checks in the `visit` do we need this probe.pred conditional?

src/ast/passes/semantic_analyser.h
```diff
@@ -156,7 +139,7 @@ class SemanticAnalyser : public Visitor {
   void binop_ptr(Binop &op);
   void binop_int(Binop &op);
   void binop_array(Binop &op);
-  void dereference_if_needed(Expression *&expr);
+  Expression *dereference_if_needed(Expression *expr);
```
jordalgo, line 142: Yeah this is much better

jordalgo:
Overall this looks good! I didn't know about CRTP before this - very cool. I'll never use virtual functions again 😄

amscanne:
Thanks for the review! I will take a pass.

CRTP is super common, esp. within LLVM itself. Dynamic polymorphism still has a place (virtual methods are not strictly RTTI), but you there are plenty of cases where static polymorphism makes more sense (and you get a surface a bunch of issues at compile time).

src/ast/visitor.h
```diff
@@ -0,0 +1,325 @@
...
+    // Before dispatching to the specific visit method, we invoke the walk hook
+    // which can be used to enumerate nodes, dump nodes, etc.
+    Impl *impl = static_cast<Impl *>(this);
+    if constexpr (std::is_base_of_v<Node, T>) {
+      Node *node = static_cast<Node *>(&t);
+      impl->preVisit(*node);
+    }
+    return impl->visit(t);
+  }
+  R _def()
```
danobi, line 313: Nit: rename to `_default()`? Thought it was "define" for a bit.

src/ast/passes/callback_visitor.h
```diff
@@ -2,50 +2,23 @@
...
-class CallbackVisitor : public Visitor {
+class CallbackVisitor : public Visitor<CallbackVisitor> {
 public:
   explicit CallbackVisitor(ASTContext &ctx, callback func)
-      : Visitor(ctx), func_(func)
+      : Visitor<CallbackVisitor>(ctx), func_(func)
   {
   }
-
-  void visit(Integer &integer) override;
```
danobi, line 19: Seems like we're going to lose `override` catching us on typos. Any known tricks we can apply here for some more guardrails? Regardless, probably fine in exchange for the performance and simplicity win.

src/ast/passes/return_path_analyser.h
```diff
@@ -1,32 +1,27 @@
...
-  bool visit(Program &prog) override;
-  bool visit(Subprog &subprog) override;
-  bool visit(Jump &jump) override;
-  bool visit(If &if_stmt) override;
-
-  // For nodes that neither directly return a value or have children with
-  // semantics such that the code represented by the node always returns
-  // a value, false is returned
-  bool default_visitor(Node &node) override;
+  using Visitor<ReturnPathAnalyser, bool>::visit;
```
danobi, line 16: Just for my understanding, this brings into scope all the overloaded `visit()` methods, right?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -665,7 +665,7 @@ void SemanticAnalyser::visit(Call &call)
       }
     }
 
-    Visit(call.vargs[i]);
+    call.vargs[i] = dereference_if_needed(call.vargs[i]);
```
danobi, line 668: Is there a way to hide `dereference_if_needed()` in a more central place like before? Or do you think it's better to open code it like this?

danobi (commented):
Overall I think this is quite elegant. Do you happen to have performance numbers yet?

src/ast/visitor.h
```diff
@@ -0,0 +1,325 @@
...
+
+namespace bpftrace {
+namespace ast {
+
+/**
+   Visitor for fully-static visitation
+
+   This uses CRTP to make all calls static, while still allowing the basic
+   entrypoint for a single visitor to be dispatched dynamically.
+*/
```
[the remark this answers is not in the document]
amscanne: Consistency with what? FWIW, I really hate this style but I used it because this was the style in the original `visitors.h`. Is there a style guide or conventions that I can converge with? (And happy to clean up older stuff)

I will personally advocate for the C++ style `//` comments all day long, the C style blocks just have too many variations (exactly like this).

src/ast/visitor.h
```diff
@@ -0,0 +1,325 @@
...
+    }
+    return impl->visit(t);
+  }
+  R _def()
+  {
+    if constexpr (!std::is_void<R>::value) {
+      return R();
+    }
+  }
+
```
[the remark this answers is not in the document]
amscanne: LOL, the code base has something funny where there are some really newish features sprinkled around (e.g. the NodeType concept in ast.h? I had not seen that before), but fairly minimal use of templates generally. I think this pattern has been around since C++11, so I'm hoping it's not too out there :)

There are couple weird holdovers that would be nicer with C++20 though (e.g. https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/config_analyser.cpp#L19), I may send some cleanups.

src/ast/passes/callback_visitor.h
amscanne: I believe that the compiler should still be able to resolve the functions statically (even if marked virtual), and I actually started with that. I ultimately decided that the likelihood of typos wasn't worth the extra characters (& potential confusion from readers), since it's just "visit" and you're likely to declare in one block up front. WDYT?

src/ast/passes/return_path_analyser.h
amscanne: Right. In C++ if you declare one method with the same name, then it immediately shadows all other methods unless there is an explicit using statement.

src/ast/passes/semantic_analyser.cpp
amscanne: I will take a think on this when I rebase. Since this is potentially mutating the value, I like this to be explicit. But there probably needs to be a way for the visitor to mutate the AST during passes, so maybe I will need to tweak these interfaces.

src/ast/visitor.h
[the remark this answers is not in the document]
danobi: IIRC right now the codebase is split between `/* .. */` and `//`. Didn't realize previous visitors.h used a new third style.

If you're up for it, could unify the style in one big commit (and then .git-blame-ignore-revs` it) and update the coding guide: https://github.com/bpftrace/bpftrace/blob/master/docs/coding_guidelines.md

src/ast/visitor.h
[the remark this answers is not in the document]
danobi: Whenever someone gets excited stuff gets sprinkled :)

src/ast/passes/callback_visitor.h
danobi: Yeah, I suppose only thing you could typo is "visit". Anything that manipulates AST is unit testable, so risk is low. Let's keep it as is, then.

src/ast/passes/semantic_analyser.cpp
amscanne: I've updated the pass and will push shortly, here's my thinking:

YES, but I don't want to do it in the pass. I've updated the visitor to allow for mutating the graph (e.g. define "visit" if you want to just walk the graph, and "replace" for the node type if you want to consider replacing it). I think we should try to organize the passes and just break this out into a separate "AutomaticDereference" pass (that could basically be defined for the generic expression and be a single method). My general feeling on a lot of the passes (esp. codegen, but that's another story) is that to the extent they can be made easier-to-understand & self-contained, the better. That's part of the reason for doing this Visitor change (so I can start adding passes that are very small and self-contained).

src/ast/passes/printer.cpp
amscanne: The output is the same.

src/ast/passes/semantic_analyser.h
amscanne: Per @danobi's comment, there may be some opportunities to simplify this further, but I don't want to understand much in the way of semantic changes.

src/ast/visitor.h
amscanne: I've used `default_value` just to drop the ugly `_` prefix.

src/ast/passes/semantic_analyser.cpp
danobi: Ack. +1 on smaller self-contained passes

src/ast/visitor.h
```diff
@@ -0,0 +1,369 @@
...
+template <typename Impl, typename R = void>
+class Visitor {
+public:
+  Visitor(ASTContext &ctx) : ctx_(ctx)
+  {
+  }
+
+  // See above; specific replace methods may be defined.
+  template <typename T>
+  T *replace(T *node, [[maybe_unused]] R *result)
```
danobi, line 33: Took me a bit to figure out the API. Now that I see it, I notice the type signature makes it unambiguous - `result` is an in-out param first containing what the `visit` returned and then containing the (optionally) modified value.

Perhaps worthwhile to spell it out a bit more in a comment and provide an example. In case anyone else gets confused.

src/ast/visitor.h
```diff
@@ -0,0 +1,369 @@
...
+#include "ast/ast.h"
+
+namespace bpftrace::ast {
+
+// Visitor for fully-static visitation.
+//
+// This uses CRTP to make all calls static, while still allowing the entrypoint
+// for a single visitor to be dispatched dynamically. The implementation may
+// optionally provide individual `visit` methods (matching either pointers or
+// references, the latter preferred), or `replace` methods (matching just the
```
danobi, line 14: Am I getting it right in that you intend to follow up this PR with one that moves `dereference_if_needed()` out into its own pass? And that would allow you to make `visit` method take const ref? Or do we still want `visit` to be able to mutate the AST?

src/ast/visitor.h
amscanne: Yeah, let me fill this in a bit more and test it. I'm sort of iterating on this gently while a few other changes churn.

My intent is to allow for complete replacement of nodes in the graph without needing to have a visitor for every single node type. For example, I'd like to be able to rewrite an expression like this:

```
public StrCast : public Visitor<StrCast, Expression*> {
...
   Expression *replace(Expression *expr, Expression **result) {
     if (*result) { // Was str(*result)
        return ctx_.make_node<StrCast>(*result);
     }
     return expr;
   }
   Expression *visit(Call &call) {
      if (call.ident == "str") {
         return call.args[0];
      }
      return nullptr; // No replacement.
   }
};
```

(Just an example, not necessarily advocating for a separate op. It just comes to mind because of all the special handling around this specific call, esp. wrt. parameters.) I think that we'll generally need this kind of replacement in order to implement the "syntactic sugar" transformations.

src/ast/visitor.h
amscanne: Visit is allowed to mutate the AST, it is just not allowed to replace a node in the graph. I didn't have plans otherwise, but it's actually an interesting idea to have *nothing* be allowed to mutate via visit (replace could also pass a const ref). The mutation would be passed up the graph automatically. But maybe this is getting a bit hairy and complicated, and I'd rather keep everything as simple as possible. WDYT?

src/ast/visitor.h
danobi: Cool, looks neat. 

> without needing to have a visitor for every single node type

Yeah that would help remove a bunch of friction from adding more passes.

src/ast/visitor.h
danobi: > Visit is allowed to mutate the AST, it is just not allowed to replace a node in the graph.

Not allowed to replace any node in the graph, or just not allowed to replace the _current_ node in the graph? Seems like it should be possible to replace children nodes, right?

src/ast/visitor.h
```diff
@@ -0,0 +1,369 @@
...
+      auto rval = visitModify(&ptr);
+      assert(ptr == &t); // Should not be modified.
+      return rval;
+    } else {
+      visitModify(&ptr);
+      assert(ptr == &t); // See above.
+    }
+  }
+
+  // Automatically unpacked and dispatch all variant and vector types into the
```
jordalgo, line 235:
```suggestion
  // Automatically unpack and dispatch all variant and vector types into the
```

src/ast/visitor.h
```diff
@@ -0,0 +1,369 @@
...
+  //
+  // In order to automatically replace a variant, e.g. change from type A to
+  // type B, it is necessary to provide a replace method that accepts that
+  // variant type directly. This could still dispatch via the standard visit
+  // function, which could e.g. return the replacement pointer, but this would
+  // be a single specialized pass for this case.
+  template <typename... Ts>
+  R visit(std::variant<Ts *...> var)
+  {
+    return std::visit([this](auto &value) -> R { return visitModify(&value); },
```
jordalgo, line 246: Naming question: should this be `visitReplace` ?

src/ast/visitor.h
jordalgo: Just curious what is the advantage of having a separate `replace` method as opposed to returning a different expression in the `visit` ? Is it mostly just to be explicit about when a replacement is happening?

src/ast/visitor.h
jordalgo: If I'm reading this right, `visit` is allowed to mutate and `replace` is allowed to completely replace. I think I would prefer if `visit` wasn't allowed to mutate and we only did explicit replacement instead of mutation otherwise I'm not sure why we should have both methods.

src/ast/visitor.h
amscanne: Some existing passes want to be able to propagate information during the visit, e.g. codegen will propagate a `ScopedExpr` that may insert lifetime ends, etc. The replace works in tandem with that, e.g. you may use a visit to identify the nodes that you are qualifying, then the replace will produce the substitutes.

src/ast/visitor.h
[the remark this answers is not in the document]
amscanne: I've posted #3749 and we can follow up there.

src/ast/passes/semantic_analyser.cpp
amscanne: Closing on this, it would be a follow-up item.

src/ast/visitor.h
amscanne: Done.

src/ast/visitor.h
amscanne: I've made this `visitAndReplace` for the internal case, I think this makes the most sense.

src/ast/visitor.h
jordalgo: Makes sense. Not a requirement for this PR, just thinking about the future state of things.

jordalgo (approved):
Let's get this in! 🚀

Issue #3529, opened 2024-10-18
Colorize semantic analyser output


danobi:
### Is your feature request related to a problem? Please describe.
We recently held an in-person hands on lab for teaching bpftrace. As part of that, I was collecting first time user experience reactions.

On thing that came up was that when an earlier error causes a following warning, it could be a bit misleading if you're not reading carefully. For example:

```
$ sudo ./build/src/bpftrace -e 'uprobe:/asdf:func { print(*args) }'
[sudo] password for dxu:
ERROR: Failed to parse DebugInfo: unable to find executable for '/asdf'
stdin:1:1-18: WARNING: No debuginfo found for /asdf
uprobe:/asdf:func { print(*args) }
```

In this case, the error triggered the warning. But it's not clear that the error is more important.

### Describe the solution you'd like

To help with this, we could colorize output if we detect attachment to a terminal. Errors are red, warnings are yellow, everything else stays as-is.

### Describe alternative solutions or features you've considered

We could silence warnings if errors are present. But generally speaking we'd like to make errors one-shot. Meaning you get as many details as possible in one run. Most compilers choose to do this to reduce unnecessary iteration. I think that is reasonable for us as well.

batu3H:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

Phoenix500526:
@danobi I'm interested in this issue. Can you assign it to me?

danobi:
Sure. Happy hacking!

PR #3710, opened 2025-01-15
enhance: colorize semantic analyser output

Phoenix500526:
Close issue #3529 

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ and this project adheres to
   - [#3522](https://github.com/bpftrace/bpftrace/pull/3522)
   - [Migration guide](docs/migration_guide.md#default-sigusr1-handler-removed)
 #### Added
+- Add env var to colorize bpftrace log output
+  - [#3710](https://github.com/bpftrace/bpftrace/pull/3710)
 - Bump max supported LLVM version to 19
   - [#3433](https://github.com/bpftrace/bpftrace/pull/3433)
 - Add `--dry-run` CLI option
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -5,7 +5,7 @@ namespace bpftrace {
 std::string logtype_str(LogType t)
 {
   switch (t) {
-    // clang-format off
+      // clang-format off
     case LogType::DEBUG   : return "";
     case LogType::V1      : return "";
     case LogType::HINT    : return "HINT: ";
@@ -39,26 +39,43 @@ void Log::take_input(LogType type,
                      std::ostream& out,
                      std::string&& input)
 {
-  auto print_out = [&]() { out << logtype_str(type) << input << std::endl; };
-
   if (loc) {
     if (src_.empty()) {
       std::cerr << "Log: cannot resolve location before calling set_source()."
                 << std::endl;
-      print_out();
+      out << log_format_output(type, std::move(input));
     } else if (loc->begin.line == 0) {
       std::cerr << "Log: invalid location." << std::endl;
-      print_out();
+      out << log_format_output(type, std::move(input));
     } else if (loc->begin.line > loc->end.line) {
       std::cerr << "Log: loc.begin > loc.end: " << loc->begin << ":" << loc->end
                 << std::endl;
-      print_out();
+      out << log_format_output(type, std::move(input));
     } else {
       log_with_location(type, loc.value(), out, input);
     }
   } else {
-    print_out();
+    out << log_format_output(type, std::move(input));
   }
+}
+
+std::string Log::log_format_output(LogType type, std::string&& input)
+{
+  if (!is_colorize_) {
+    return logtype_str(type) + std::move(input) + '\n';
+  }
+  std::string color;
+  switch (type) {
+    case LogType::ERROR:
+      color = LogColor::RED;
+      break;
+    case LogType::WARNING:
+      color = LogColor::YELLOW;
+      break;
+    default:
+      return logtype_str(type) + std::move(input) + '\n';
+  }
+  return color + logtype_str(type) + std::move(input) + LogColor::RESET + '\n';
 }
 
 const std::string Log::get_source_line(unsigned int n)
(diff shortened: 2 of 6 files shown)


danobi:
Hi, thanks for working on this!

I think we need some kind of environment variable or CLI flag to control colorization. For example, an env var like `BPFTRACE_COLOR=[always|never|auto]`, where `auto` is the default. You'd check if bpftrace is attached to a terminal under auto, and if so, enable colors.

For tests you'd want to disable colors.

Phoenix500526:
> Hi, thanks for working on this!
> 
> I think we need some kind of environment variable or CLI flag to control colorization. For example, an env var like `BPFTRACE_COLOR=[always|never|auto]`, where `auto` is the default. You'd check if bpftrace is attached to a terminal under auto, and if so, enable colors.
> 
> For tests you'd want to disable colors.

Done. Please review this pr, thanks. @danobi

src/log.cpp
```diff
@@ -34,31 +35,75 @@ Log& Log::get()
...
+bool Log::is_terminal(std::ostream& out)
+{
+  if (&out == &std::cout) {
+    return isatty(STDOUT_FILENO);
+  }
+  if (&out == &std::cerr) {
+    return isatty(STDERR_FILENO);
+  }
+  return false;
+}
```
danobi, line 47: At least for semantic analyser, this won't work b/c semantic analysis buffers error output in an `std::ostringstream`. So it'll never directly be attached to a terminal.

I think terminal checking should probably be done from `main()`. And can set a flag on whether or not to colorize using `Log::get().set_color(..)`, or something like that.

That way it'll never be on for unit tests. And we can more reliably look at stdout/stderr properties.

danobi (commented):
Few other procedural bits:
* Please update CHANGELOG.md
* Please update `--help` text (in main.cpp) with new env var

src/log.cpp
Phoenix500526: Done

src/log.cpp
```diff
@@ -39,26 +40,44 @@ void Log::take_input(LogType type,
...
     }
   } else {
-    print_out();
+    out << log_format_output(type, std::move(input));
+  }
+}
+
+std::string Log::log_format_output(LogType type, std::string&& input)
+{
+  if (is_colorize_) {
```
jordalgo, line 65: nit: you could avoid some nesting if you return early if `is_colorize` is false

CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ and this project adheres to
   - [#3522](https://github.com/bpftrace/bpftrace/pull/3522)
   - [Migration guide](docs/migration_guide.md#default-sigusr1-handler-removed)
 #### Added
+- Colorize bpftrace log output
```
jordalgo, line 55:
```suggestion
- Add env var to colorize bpftrace log output
```

jordalgo (commented):
Overall looks good. Just a few nits and maybe fix the changelog language.

src/main.cpp
```diff
@@ -692,10 +693,27 @@ Args parse_args(int argc, char* argv[])
...
+  } else if (mode == "never") {
+    return false;
+  }
+  return isatty(STDOUT_FILENO) && isatty(STDERR_FILENO);
+}
+
 int main(int argc, char* argv[])
 {
+  bool colorize = is_colorize();
+  Log::get().set_colorize(colorize);
```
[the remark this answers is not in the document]
Phoenix500526: Done

src/log.cpp
Phoenix500526: Done

CHANGELOG.md
Phoenix500526: Done

jordalgo (approved):

src/log.cpp
```diff
@@ -1,11 +1,12 @@
 #include "log.h"
+#include <unistd.h>
```
danobi, line 2: Unnecessary include?

src/log.h
```diff
@@ -67,7 +78,9 @@ class Log {
                          const location&,
                          std::ostream&,
                          const std::string&);
+  std::string log_format_output(LogType, std::string&&);
   std::unordered_map<LogType, bool> enabled_map_;
+  bool is_colorize_;
```
danobi, line 83: Be good to default it in case it's being run in unit tests.

```suggestion
  bool is_colorize_ = false;
```

src/main.cpp
```diff
@@ -692,10 +693,26 @@ Args parse_args(int argc, char* argv[])
...
 }
 
+bool is_colorize()
+{
+  const char* color_env = std::getenv("BPFTRACE_COLOR");
+  if (!color_env) {
+    return isatty(STDOUT_FILENO) && isatty(STDERR_FILENO);
+  }
+
+  std::string mode(color_env);
```
danobi, line 703: Nit: non-allocating view is probably better here
```suggestion
  std::string_view mode(color_env);
```

src/main.cpp
```diff
@@ -692,10 +693,26 @@ Args parse_args(int argc, char* argv[])
...
+  if (mode == "always") {
+    return true;
+  } else if (mode == "never") {
+    return false;
+  }
```
danobi, line 708: Please do an explicit check for "auto" and handle invalid values. Should probably error to the user if they typo'd it.

danobi (commented):
It doesn't look like this works for most of semantic analyser output - you're missing color wrapping around Log::log_with_location() in Log::take_input().

Please add unit tests to `tests/log.cpp` to cover that case, along with the other 4 branches in Log::take_input().

src/log.cpp
Phoenix500526: Done

src/log.h
Phoenix500526: Done

src/main.cpp
Phoenix500526: Done

src/main.cpp
Phoenix500526: Done

Phoenix500526:
> It doesn't look like this works for most of semantic analyser output - you're missing color wrapping around Log::log_with_location() in Log::take_input().
> 
> Please add unit tests to `tests/log.cpp` to cover that case, along with the other 4 branches in Log::take_input().

Done

src/log.cpp
```diff
@@ -82,6 +99,19 @@ void Log::log_with_location(LogType type,
...
                             const std::string& m)
 {
+  if (is_colorize_) {
+    switch (type) {
+      case LogType::ERROR:
+        out << LogColor::RED;
+        break;
+      case LogType::WARNING:
+        out << LogColor::YELLOW;
+        break;
```
danobi, line 109: Nit: set a local variable for reset code here? Default is empty string. That would simplify and deduplicate the duplicated LogType checks below

Phoenix500526: Done

src/log.cpp
```diff
@@ -144,6 +181,9 @@ void Log::log_with_location(LogType type,
       out << marker;
     }
   }
+  if (is_colorize_ && (type == LogType::ERROR || type == LogType::WARNING)) {
+    out << LogColor::RESET;
+  }
```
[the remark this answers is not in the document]
Phoenix500526: Done

danobi (approved):
lgtm. thanks for iterating on this!

PR #3714, opened 2025-01-15
Fix simple leak from semantic_analyser pass

amscanne:
This should be allocated in the context, and therefore freed when the ASTContext is also freed. Currently it is orphaned.

##### Checklist


jordalgo:
I wonder why our ASAN build didn't catch this.

danobi:
> I wonder why our ASAN build didn't catch this.

The memory leaks tests are restricted to a few "known good" scripts. There were previously way too many leaks to run it always. Would be nice to have it on for all our tests. Good reliability project.

danobi (approved):

danobi:
@amscanne FYI this PR is in draft mode. Is that intentional?

jordalgo (approved):

PR #3715, opened 2025-01-16
Do not generate functions for empty attach points

viktormalik:
When a probe contains a mix of wildcarded and non-wildcarded attach points which are not fully expanded (e.g. `kprobe:fun,kprobe:vfs*` as the latter is multi-expanded) and some wildcarded attach point has no matches, we must avoid generating an LLVM function for that attach point as libbpf would turn it into a BPF program for which we wouldn't have a corresponding Probe object and loading of such program would fail.

This fixes the dcsnoop.bt tool as it contains:

    kprobe:lookup_fast,
    kprobe:lookup_fast.constprop.*

and on some kernels, the second probe expands to nothing if `lookup_fast` is not affected by compiler's constant propagation. Without this fix, the tool would fail to load.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,13 @@ and this project adheres to
   - [#3704](https://github.com/bpftrace/bpftrace/pull/3704)
 - Fix `pid`, `tid` and `ustack` when running bpftrace in containers with PID namespacing
   - [#3428](https://github.com/bpftrace/bpftrace/pull/3428)
+- Do not generate functions for empty attach points
+  - [#3715](https://github.com/bpftrace/bpftrace/pull/3715)
 #### Security
 #### Docs
 #### Tools
+- Fix dcsnoop.bt on newer kernels
+  - [#3715](https://github.com/bpftrace/bpftrace/pull/3715)
 
 ## [0.22.0] 2025-01-07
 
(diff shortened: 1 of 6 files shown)


viktormalik:
I think that this should be backported to the 0.22.x branch and we should release 0.22.1 as it fixes a tool which many distros ship.

jordalgo:
@viktormalik The test failure on llvm 18 looks related "codegen.builtin_func_wild"

jordalgo:
> I think that this should be backported to the 0.22.x branch and we should release 0.22.1 as it fixes a tool which many distros ship.

Sounds good to me. I can do the patch release once the two additions are merged.

viktormalik:
> @viktormalik The test failure on llvm 18 looks related "codegen.builtin_func_wild"

Yes, the test needs to attach to a different function, will fix it.

viktormalik:
Fixed, should be ready now.

jordalgo (approved):

viktormalik:
> > I think that this should be backported to the 0.22.x branch and we should release 0.22.1 as it fixes a tool which many distros ship.
> 
> Sounds good to me. I can do the patch release once the two additions are merged.

That would be great, thanks!

PR #3716, opened 2025-01-16
tests: Do not parse system BTF by default

danobi:
We have gtest fixtures in place for synthetic BTF data. For example: codegen_btf and test_btf. Tests should explicltly use that if they want BTF mocked.

By default we should prevent unit tests from reading host state - unit tests should be hermetic.

##### Checklist


danobi:
This is finding some legit host dependencies. Here's the full denylist for getting the suite to work:

```
$ ./build/tests/bpftrace_test --gtest_filter='-semantic_analyser.array_in_map:semantic_analyser.array_compare:semantic_analyser.intarray_to_int_cast:semantic_analyser_btf.*:codegen.call_percpu_kaddr:codegen.call_percpu_kaddr_this_cpu:codegen.fentry_dereference:codegen.fexit_dereference:codegen.iter_dereference:codegen_btf.call_len_map_sum_elem_count'
```

Gonna background this. Someone else is free to pick this up. But this would be great to get done (helps w/ test performance as well as hermeticity)

PR #3717, opened 2025-01-16
BTF: Print textual error as opposed to mere error number

danielocfb:
Print a textual representation of the error as opposed to a mere number that the user has to decode when logging an error in the BTF logic.


jordalgo:
Thanks for the contribution. LGTM, just need to fix the formatting.

danielocfb:
Thanks for the review. Should be fixed now.

danobi (approved):

PR #3721, opened 2025-01-17
Simplify max nodes check

amscanne:
This simplifies the max nodes check, which was added for fuzzing.

First, the node visitor can be replaced by a trivial `node_count` check on the ASTContext, which will directly determine the number of nodes allocated for the AST without requiring a full walk.

Second, since this check is already implemented (along with the required environment variable), there is no need to do this with the fuzzing parse separately -- parsing will already fail.

This allows us to eliminate *both* the visitors (NodeCounter and CallbackVisitor) and simplify the pass plumbing.

##### Checklist


src/fuzz_main.cpp
```diff
@@ -102,20 +101,6 @@ int fuzz_main(const char* data, size_t sz)
   if (err)
     return err;
 
-  // Limit node size
-  uint64_t max_ast_nodes = DEFAULT_NODE_MAX;
```
jordalgo, line 106: the define of DEFAULT_NODE_MAX can also be removed. But is it the case that sometimes this fuzzing test sets this ENV var and sometimes it doesn't?

jordalgo (commented):
Great cleanup!

src/fuzz_main.cpp
amscanne: The use of the variable is emphasized in the docs, I don't think fuzzing should have a different default behavior than the program. I've dropped the #define.

amscanne:
I think the LLVM 14 failure is an existing flake (and I have no way to rerun w/o updating the commit).

jordalgo:
@amscanne I got you ;)

jordalgo (approved):
LGTM. @viktormalik or @danobi - quick sanity check before I merge?

amscanne:
Once this is merged I will simplify and rebase #3699.

PR #3725, opened 2025-01-21
doc: use dash (minus) instead colon

makelinux:
Quote:

The colon is a powerful punctuation mark and should be used sparingly.

https://blog.wordvice.com/when-to-use-commas-colons-semicolons-and-dashes/

For example, dash (minus) is used in man pages:
https://man7.org/linux/man-pages/dir_all_by_section.html


jordalgo:
I don't feel strongly but I'm fine with a dash.

jordalgo (approved):

PR #3729, opened 2025-01-22
cleanup: update explicit C++20 TODOs

amscanne:
This change updates C++20 TODOs with a reasonable implementation. There may be other code that could be updated, but this at least removes the explicit cases that are marked for improvement with C++20.

##### Checklist

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -775,7 +775,7 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
   auto paths_v1 = filtered_v1 | get_path_func;
   auto paths_v2 = filtered_v2 | get_path_func;
 
-  // Return paths with v2 first, then v2 sorted lexically by name.
+  // Return paths with v2 first, then v1 sorted lexically by name.
   std::vector<std::pair<std::string, std::string>> sorted(paths_v2.begin(),
                                                           paths_v2.end());
   std::vector<std::pair<std::string, std::string>> sorted_v1(paths_v1.begin(),
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -10,6 +10,7 @@
 #include <map>
 #include <memory>
 #include <optional>
+#include <ranges>
 #include <regex>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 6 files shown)


src/utils.cpp
```diff
@@ -735,55 +736,53 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
...
+                     if (root.first == "cgroup") {
+                       return {
+                         std::filesystem::path(root.second).filename().string(),
+                         root.second
+                       };
+                     } else if (root.first == "cgroup2") {
+                       return { "unified", root.second };
+                     }
+                     LOG(BUG) << "Unknown cgroup filesystem: " << root.first;
+                     __builtin_unreachable();
```
jordalgo, line 755: Maybe we should be throwing here so we just have a single statement?

amscanne: I actually just changed the structure here to eliminate the need for any kind of branch & assertion in this path.

src/utils.cpp
```diff
@@ -735,55 +740,50 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
...
-    if (pair1.first == "unified")
-      return true;
-    return pair1.first < pair2.first;
-  });
-
-  return result;
+  std::vector<std::pair<std::string, std::string>> sorted_v1(paths_v1.begin(),
+                                                             paths_v1.end());
+  std::sort(sorted_v1.begin(), sorted_v1.end());
+  std::vector<std::pair<std::string, std::string>> sorted(paths_v2.begin(),
```
jordalgo, line 783: readability nit: I think this line should go under the comment above about having "unified" come first.

Also (and I haven't tried this) do we have to create the temporary `sorted_v1` vector as opposed to sorting paths_v1 in place and then inserting into `sorted`?

src/utils.cpp
jordalgo: > This is a bigger thread, I will file a separate bug.

Definitely 😄  Saw the new thread. Thanks for filing and thinking about! I'll comment there.

jordalgo (approved):

src/utils.cpp
amscanne: The views are a bit weird. You can actually sort using the view itself and then concat, but this is a C++25 thing. You *can* solve it if you built your own type erasure bits, but I didn't really want to go down that path. I'm happy to reorganize to make this more readable.

src/utils.cpp
```diff
@@ -735,55 +740,49 @@ std::vector<std::pair<std::string, std::string>> get_cgroup_paths(
...
+  // Get cgroup path for each root
+  auto get_path_func = std::views::transform(
+      [&](auto pair) -> std::pair<std::string, std::string> {
+        return { pair.first,
+                 get_cgroup_path_in_hierarchy(cgroupid, pair.second) };
+      });
+  auto paths_v1 = filtered_v1 | get_path_func;
+  auto paths_v2 = filtered_v2 | get_path_func;
+
+  // Return paths with v2 first, then v2 sorted lexically by name.
```
jordalgo, line 778:
```suggestion
  // Return paths with v2 first, then v1 sorted lexically by name.
```

jordalgo (approved):

PR #3730, opened 2025-01-23
RFC: Remove type signatures from adoc tables

jordalgo:
This is all in the name of making it easier
to convert our asciidoc to markdown
and have it be rendered properly on
docs.bpftrace.org

Brackets ('[]') and pipes ('|'), even escaped
give the tool (downdoc) a lot of issues
inside the table columns.

This is a bit of a usability regression
but I think it might be ok as we want
people reading the docs on the above
website.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1293,149 +1293,149 @@ Tracing block I/O sizes > 0 bytes
 
 [%header]
 |===
-| Function Name | Description | Sync/Async/Compile Time
+| Name | Description | Sync/Async/Compile Time
 
-| <<functions-bswap, `bswap(uint[8\|16\|32\|64] n)`>>
+| <<functions-bswap, `bswap`>>
 | Reverse byte order
 | Sync
 
-| <<functions-buf, `buf(void *d [, int length])`>>
+| <<functions-buf, `buf`>>
 | Returns a hex-formatted string of the data pointed to by d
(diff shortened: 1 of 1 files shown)


danobi:
I think the types there are pretty important :(

Is it possible to render the asciidoc to html and then embed it? 

Worst case is an iframe, right? https://stackoverflow.com/questions/68091978/is-there-any-way-to-render-a-html-file-in-a-markdown-post-with-docusaurus

fbs:
Can't we chanage the syntax of them? Its a bit ugly, but better than removing them completely.

Or maybe some intermediate processing step?

danobi:
One thing I just realized is that the type signatures are available further down in the expanded helper docs. So removing them from table might be an improvement - cleaner and less duplication. Note between the table and the expanded docs we already have skew on our "types". Eg 

> [strftime(char *format, int nsecs)](https://docs.bpftrace.org/latest#strftime)

> timestamp strftime(const string fmt, int64 timestamp_ns)

jordalgo:
Ok it seems from comments and conversation that it's good to remove these signatures from the tables: they're ugly and duplicated (sometimes incorrectly).

danobi (approved):

danobi:
@fbs  wdyt?

fbs:
ah yes, lets remove those then

fbs (approved):

PR #3735, opened 2025-01-24
Drop clang_workarounds.h

viktormalik:
The `clang_workarounds.h` header was introduced to allow including and parsing kernel headers with Clang back in the days when Clang didn't support `asm goto` and `asm inline`.

`asm goto` has been added in Clang 9 [1] and `asm inline` in Clang 11 [2]. Since the minimum LLVM version supported by bpftrace is 14, we can finally drop this code.

One of the problems that this header has is that it includes `<linux/types.h>` which causes redefinition errors when `<sys/types.h>` is included from the bpftrace script. Since `<linux/types.h>` is not needed when BTF is available, this should prevent such conflicts in a majority of cases.

[1] https://github.com/llvm/llvm-project/commit/784929d0454c
[2] https://github.com/llvm/llvm-project/commit/246398ece711

##### Checklist


jordalgo (approved):
LGTM. Good catch!

amscanne:
Interesting. Aren't both of these headers (clang_workarounds and linux/types) omitted if BTF data is present? I was recently wondering what the support surface is for cases where there is no BTF -- is this a specific edge case or is there a common path where this is the case?

viktormalik:
> Interesting. Aren't both of these headers (clang_workarounds and linux/types) omitted if BTF data is present? 

Yes, that should be the case but these kinds of errors just keep popping up so it may be that we have a wrong condition somewhere.

> I was recently wondering what the support surface is for cases where there is no BTF -- is this a specific edge case or is there a common path where this is the case?

My understanding is that we include `<linux/types.h>` in 2 places:
- From `clang_workarounds.h` which is included every time, however, we define `-D_LINUX_TYPES_H` when BTF is available so the header should not be parsed.
- When tracepoint probes are used and BTF is not available. This is b/c we use ClangParser to parse tracepoint format definitions which contain linux internal types.

This PR removes the first case.

amscanne:
Is this from a use case where BTF is not available or you think that it is being incorrectly included when BTF *is* available?

The thing that I'm wondering is if/when you'd be able to say that BTF is actually a requirement. AFAIK, BTF has been available prior to all available LTS releases (5.2 vs 5.4) and seems to have been widely available in most distributions for years (e.g. https://github.com/aquasecurity/btfhub/blob/main/docs/supported-distros.md). I'm thinking about the opportunity to simplify by relying on BTF more directly for the types.

viktormalik:
> Is this from a use case where BTF is not available or you think that it is being incorrectly included when BTF _is_ available?

So this problem I was looking into is from a case when BTF was available, see #3744.

> The thing that I'm wondering is if/when you'd be able to say that BTF is actually a requirement. AFAIK, BTF has been available prior to all available LTS releases (5.2 vs 5.4) and seems to have been widely available in most distributions for years (e.g. https://github.com/aquasecurity/btfhub/blob/main/docs/supported-distros.md). I'm thinking about the opportunity to simplify by relying on BTF more directly for the types.

Interestingly, requiring BTF would not drop some large amount of code. The main reason is that we cannot drop ClangParser as we still support user includes. The only remaining place where we _internally_ use ClangParser when BTF is not available is for parsing tracepoint arguments.

That being said, I'm not strictly against making BTF a requirement, however, we've seen people using bpftrace on systems where BTF was not available (partially due to an old kernel but partially for other reasons, e.g. saving space).

amscanne:
> Interestingly, requiring BTF would not drop some large amount of code. The main reason is that we cannot drop ClangParser as we still support user includes. The only remaining place where we internally use ClangParser when BTF is not available is for parsing tracepoint arguments.

Yep, it's just some blue sky thinking, which is why I'm wondering how much this is used. There are interesting alternatives there, e.g. we could use debug info from the binaries (directly or users can actually attach BTF there, e.g. `pahole -J <mybinary>`). Obviously that's a potential lift for users, but so is having headers where you are running `bpftrace` (and it's not actually guaranteed to *match* the binary itself, unlike debug info or type information embedded there).

Interestingly, I think this does relate to the issue you were working on it. Right now we basically marshall all the types through a single ClangParser instance and have to deal with conflicts. Essentially we have to construct a single unified type system for everything in the script. In reality, each probe really has its own types (e.g. if you probe two different binaries simultaneously, they could have completely different and conflicting types), and the current `#include` model can't really capture that. (But having the type system sucked in from the individual binaries/kernel directly for each probe would resolve that. This is all notwithstanding what happens when a kernel probe wants to reach up into userspace.)

Anyways, it's just all blue sky thinking. Very curious to hear your perspective.

> That being said, I'm not strictly against making BTF a requirement, however, we've seen people using bpftrace on systems where BTF was not available (partially due to an old kernel but partially for other reasons, e.g. saving space).

Totally, but you've got to draw the line somewhere. (Also, it's worth noting that if people have an ancient or esoteric system, there's nothing stopping them from using older versions.) I'm not really pushing for anything, it really only makes sense if there's a significant upside & savings in complexity, maintainability, features, etc. which is what I'm wondering about above.

viktormalik:
> Yep, it's just some blue sky thinking, which is why I'm wondering how much this is used. There are interesting alternatives there, e.g. we could use debug info from the binaries (directly or users can actually attach BTF there, e.g. `pahole -J <mybinary>`). Obviously that's a potential lift for users, but so is having headers where you are running `bpftrace` (and it's not actually guaranteed to _match_ the binary itself, unlike debug info or type information embedded there).

If I understand correctly, this should be what I added in #2034 (I hope it survived the refactoring to use lldb). The problem is that people typically trace binaries provided by the distro, i.e. without debug info and stripped. Most distros provide debug info in separate packages and I'm not sure if bpftrace can currently parse those. If not, it's certainly a nice addition but even then, we must let people know that it's better to install the -debuginfo package than the -devel package and include headers. And I think that this will be the real challenge as most tutorials and bpftrace examples out there just use includes. Perhaps we could add a hint to do that when ClangParser fails.

Also, AFAIK, BTF is not designed to be used for userspace binaries so we rely on DWARF there.

> Interestingly, I think this does relate to the issue you were working on it. Right now we basically marshall all the types through a single ClangParser instance and have to deal with conflicts. Essentially we have to construct a single unified type system for everything in the script. In reality, each probe really has its own types (e.g. if you probe two different binaries simultaneously, they could have completely different and conflicting types), and the current `#include` model can't really capture that. (But having the type system sucked in from the individual binaries/kernel directly for each probe would resolve that. This is all notwithstanding what happens when a kernel probe wants to reach up into userspace.)

As mentioned above, I think we already have that (at least to some extent) but need to convince users to switch from the include model to the debuginfo model. Another story would be to allow per-probe header includes but I'm not sure if it's worth the work.

> > That being said, I'm not strictly against making BTF a requirement, however, we've seen people using bpftrace on systems where BTF was not available (partially due to an old kernel but partially for other reasons, e.g. saving space).
> 
> Totally, but you've got to draw the line somewhere. (Also, it's worth noting that if people have an ancient or esoteric system, there's nothing stopping them from using older versions.) I'm not really pushing for anything, it really only makes sense if there's a significant upside & savings in complexity, maintainability, features, etc. which is what I'm wondering about above.

Agreed, it would be nice to have a better view into how many non-BTF users of bpftrace there are. I suspect it will be a vast minority these days - maybe just some embedded devices, I'm also not sure about Android but I know people use bpftrace there.

viktormalik:
While the discussion here was useful, I believe that nothing prevents us from merging this PR, right?

cc @amscanne

amscanne (approved):

amscanne:
> While the discussion here was useful, I believe that nothing prevents us from merging this PR, right?
> 
> cc @amscanne

Totally, the thread is just me trying to understand long-term direction. Thanks for engaging :)

PR #3738, opened 2025-01-24
cleanup: drop `ENABLE_MAN` support

amscanne:
In an effort to reduce build optionality, remove the `ENABLE_MAN` flag. This was added in ef39ed0f0196cd418266721d0a122670b98bda9c, but does not appear in a spot check of upstream packaging rules [1][2], and the only that I can see appears abandoned [3].

It's probably a good idea to advise folks to simply install the bits they want through their own scripts or build wrappers, rather than pushing this business logic down into the cmake configuration.

[1] https://git.launchpad.net/ubuntu/+source/bpftrace/tree/debian/rules
[2] https://git.centos.org/rpms/bpftrace/blob/bd8a9c838039cbe306b0377963aed4cb16676b5d/f/SPECS/bpftrace.spec#_60
[3] https://github.com/ce-bu/meta-observability/commits/main/

##### Checklist


amscanne:
I'm happy to provide a changelog entry for this, and explain the alternative. I think pushing configuration down into cmake via bespoke flags isn't the best pattern, and it would be good to have a general recommended. My intuition is something like:

* Make configuration opportunistic in all cases possible, e.g. don't configure test targets if `gtest` is not available. Our CI environments will have these things (and CI entrypoint is doing something trivial to ensure that tests are all built). This is already true for *most* things
* If you only care about specific targets, e.g. the `bpftrace` binary and the man page, just do `make bpftrace man` and extract those targets. Alternatively, we can group into cmake components, e.g. bpftrace-binaries, bpftrace-libraries, bpftrace-data, to allow the install of just those things

But I can think about this a bit because I'll be following up with some similar PRs

danobi:
I'm kinda on the fence about opportunistic. It can get kinda confusing when targets are missing b/c of missing deps. Maybe being good about emitting warnings on missing optional deps is enough to address that, though.

Cmake components could make sense for making `make install` more structured. But I don't think there's a builtin way to dump the list of components. So same issue as with targets: discover-ability.

danobi:
Closing due to inactivity. Feel free to reopen later.

PR #3741, opened 2025-01-25
Add feature check for map_lookup_percpu_elem

jordalgo:
This feature wasn't added till 5.19
so we need to check if it's available.

https://github.com/bpftrace/bpftrace/issues/3732

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2468,6 +2468,12 @@ void SemanticAnalyser::visit(For &f)
   map.skip_key_validation = true;
   Visit(&map);
 
+  if (map.type.IsCastableMapTy() &&
+      !bpftrace_.feature_->has_helper_map_lookup_percpu_elem()) {
+    LOG(ERROR, f.expr->loc, err_)
+        << "Missing required kernel feature: map_lookup_percpu_elem";
+  }
+
   if (has_error())
     return;
 
--- a/tests/semantic_analyser.cpp
+++ b/tests/semantic_analyser.cpp
@@ -4371,6 +4371,20 @@ BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }
              false);
 }
 
+TEST(semantic_analyser, for_loop_castable_map_missing_feature)
+{
+  test_error("BEGIN { @map = count(); for ($kv : @map) { print($kv); } }",
+             R"(
+stdin:1:25-28: ERROR: Missing required kernel feature: for_each_map_elem
+BEGIN { @map = count(); for ($kv : @map) { print($kv); } }
+                        ~~~
+stdin:1:36-41: ERROR: Missing required kernel feature: map_lookup_percpu_elem
+BEGIN { @map = count(); for ($kv : @map) { print($kv); } }
+                                   ~~~~~
+)",
+             false);
+}
+
 TEST(semantic_analyser, for_loop_no_ctx_access)
 {
   test_error("kprobe:f { @map[0] = 1; for ($kv : @map) { arg0 } }",
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -831,6 +831,18 @@ Value *IRBuilderBPF::CreatePerCpuMapAggElems(Value *ctx,
 
   const std::string &map_name = map.ident;
 
+  if (!bpftrace_.feature_->has_helper_map_lookup_percpu_elem()) {
+    /*
+     * Unfortunately detecting this in semantic_analyser would be very ugly &
+     * complicated for all implicit casting, so error here until we drop
+     * support for kernel 5.19
+     */
+    throw FatalUserException("Missing required kernel feature "
+                             "(map_lookup_percpu_elem) to read map: " +
+                             map_name);
+    exit(-1);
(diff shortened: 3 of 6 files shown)


tests/semantic_analyser.cpp
```diff
@@ -4371,6 +4410,20 @@ BEGIN { @map[0] = 1; for ($kv : @map) { print($kv); } }
              false);
 }
 
+TEST(semantic_analyser, for_loop_castable_map_missing_feature)
+{
+  test_error("BEGIN { @map[0] = count(); for ($kv : @map) { print($kv); } }",
```
jordalgo, line 4415: I thought it might be overkill to test all castable map types here (I check them all above in 'castable_map_missing_feature') but I can add if people want. I can also try something dynamic.

tests/semantic_analyser.cpp
```diff
@@ -2451,6 +2451,45 @@ kprobe:f { @ = hist(5); print((1, (uint16)@)); }
 )");
 }
 
+TEST(semantic_analyser, castable_map_missing_feature)
+{
+  test_error("BEGIN { @a = count(); print(@a); }",
```
danobi, line 2456: I don't think `print(@)` forces a collapse on percpu type into scalar. Rather, it's an async map print. For example, compare outputs of these two:

```
sudo bpftrace -e 'BEGIN { @ = count(); print(@) }' -dcodegen
```

```
sudo bpftrace -e 'BEGIN { @ = count(); print((int64)@) }' -dcodegen
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2454,6 +2461,12 @@ void SemanticAnalyser::visit(For &f)
     return;
   }
 
+  if (map.type.IsCastableMapTy() &&
+      !bpftrace_.feature_->has_helper_map_lookup_percpu_elem()) {
+    LOG(ERROR, f.expr->loc, err_)
+        << "Missing required kernel feature: map_lookup_percpu_elem";
+  }
```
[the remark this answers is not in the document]
jordalgo: It shouldn't that's why I have the `map.type.IsCastableMapTy()` check (unless I overlooked something)

tests/semantic_analyser.cpp
jordalgo: Ah ok, I'll fix.

src/ast/passes/semantic_analyser.cpp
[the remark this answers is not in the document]
danobi: Ah, right. My bad.

tests/semantic_analyser.cpp
jordalgo: As per my updated comment, it seems that checking for this case in semantic analyser would involve a lot of ugly code so I just dropped in a check on the codegen path, which isn't great because we lose the line number. But I don't think all the extra code in semantic analyser is worth it to detect this feature on older kernels. WDYT?

danobi: What was the complication? 5.15 is still somewhat new - I expect it'll be around for 2-4 more years.

And returning error from codegen is not very nice, so it'd be preferable to find another way.

jordalgo: @danobi Here is what this looks like checked in semantic analyser. It's not too bad but my main concern is adding another property of the Map ast type AND potentially missing setting this property to false on a case I didn't account for because the default needs to be 'true'. https://github.com/bpftrace/bpftrace/pull/3752

danobi: Hmm, agreed it's not the prettiest. But given it's allowlist style, actually not too bad? If you miss a case for some new feature, then user is blocked from doing something that should work. Which is a better outcome than being allowed to do something that won't work.

That seems like something that would be more easily surfaced in tests, especially once we add an "oldest supported kernel" vmtest job.

Long term maybe we can move all the feature checking out into its own pass (cc @amscanne)? That could help contain the complexity and make details like this more obvious.

jordalgo: > That seems like something that would be more easily surfaced in tests, especially once we add an "oldest supported kernel" vmtest job.

I was thinking the same thing. I tried to run the runtime test locally with vmtest but they all kept timing out - I was probably doing something wrong.

> Long term maybe we can move all the feature checking out into its own pass

Makes sense.

So would you rather I abandon this PR for the other one? My slight preference is for this one to reduce complexity in semantic analyser and know that I'm always erroring when we're going to actually call the bpf function but I'll have some solace if we land the vmtest stuff and we can run the runtime tests on a 5.15 kernel.

danobi: I prefer the other PR (ie doing it in semantic analyser). Codegen is not designed to return errors. And we can emit better diagnostics in semantic analyser. Imagine having a large script with lots of operations on a map. The codegen exception would not be super helpful.

Maybe just document the new map field with a comment and tag an issue that moves all the feature check stuff out into its own pass?

I could see it being a lot simpler in its own pass. For example, you don't need to modify the ast - you could have an `std::set <Node *> maps_that_are_read`  that keeps track of the information. Doing that kind of collection in semantic analyser is more annoying.

jordalgo:
Closing in favor of: https://github.com/bpftrace/bpftrace/pull/3752

PR #3743, opened 2025-01-27
child.cpp: Add missing include of sched.h

bjackman:
An build failure was reported with Google's internal toolchain. I don't know how to reproduce that build failure here, but it seems legit to me, this file calls clone() but doesn't include sched.h.

So, include it.

##### Checklist

(All N/A)

### DIFF
--- a/src/child.cpp
+++ b/src/child.cpp
@@ -2,6 +2,7 @@
 #include <cerrno>
 #include <fcntl.h>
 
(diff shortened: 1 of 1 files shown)


viktormalik (approved):
Looks reasonable, thanks!

jordalgo (approved):

PR #3745, opened 2025-01-27
cleanup: run include-what-you-use on existing headers

amscanne:
This removes a few headers and mostly reflects that there are no implicit transitive dependencies. Developer documentation is added for how to run this in the future. It is not necessarily intended to be a continuous process (as it requires some degree of human judgement).

##### Checklist


amscanne:
`clang-tidy` is now running on headers and there was a significant cleanup.

PR #3746, opened 2025-01-27
cleanup: remove trailing whitespace

amscanne:
The histogram printing code was generating a line with trailing whitespace, which required the test to also match against this trailing whitespace. This change fixes this formatting, and updates the associated tests. It also removes a few incidental cases of trailing whitespace.

I encountered this issue assumingly because my editor would automatically strip trailing whitespace when editing `output.cpp`. This is not uncommon, and I was strangely enough not able to edit this file without breaking tests w/o resorting to manual trickery to restore these spaces (`awk 'NR==45{print $0 " "} NR!=45;'`). Presumably they are not intentional, so may as well quickly fix this and avoid anyone stumbling over this issue in the future.

##### Checklist


danobi (approved):
Very nice

jordalgo (approved):
Sooo you deleted nothing 😉  ?

PR #3751, opened 2025-01-29
nix: Remove unnecessary overlays

danobi:
Nix overlays are most useful when you want to operate on the entire dependency graph. For our use case, we only want to modify libbpf and bcc.

The canonical way (AFAIK) to do that is by using overrides. This commit removes an unnecessary level of indirection and achieves the same thing.

##### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -38,11 +38,18 @@
               owner = "libbpf";
               repo = "libbpf";
               rev = "v${libbpfVersion}";
-              # If you don't know the hash the first time, set:
-              # hash = "";
+              # Nix uses the hash to do lookups in its cache as well as check that the
+              # download from the internet hasn't changed. Therefore, it's necessary to
+              # update the hash every time you update the source. Failure to update the
+              # hash in a cached environment (warm development host and CI) will cause
+              # nix to use the old source and then fail at some point in the future when
+              # the stale cached content is evicted.
+              #
+              # If you don't know the hash, set:
+              #   sha256 = "";
               # then nix will fail the build with such an error message:
-              # hash mismatch in fixed-output derivation '/nix/store/m1ga09c0z1a6n7rj8ky3s31dpgalsn0n-source':
-              # specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
+              #   hash mismatch in fixed-output derivation '/nix/store/m1ga09c0z1a6n7rj8ky3s31dpgalsn0n-source':
+              #   specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
               # got:    sha256-173gxk0ymiw94glyjzjizp8bv8g72gwkjhacigd1an09jshdrjb4
               sha256 = "sha256-+L/rbp0a3p4PHq1yTJmuMcNj0gT5sqAPeaNRo3Sh6U8=";
             };
@@ -57,6 +64,7 @@
               owner = "iovisor";
               repo = "bcc";
               rev = "v${bccVersion}";
+              # See above
               sha256 = "sha256-6dT3seLuEVQNKWiYGLK1ajXzW7pb62S/GQ0Lp4JdGjc=";
             };
           };
--- a/flake.nix
+++ b/flake.nix
@@ -64,7 +64,7 @@
           # Define lambda that returns a derivation for bpftrace given llvm version as input
           mkBpftrace =
             llvmVersion:
-              pkgs.stdenv.mkDerivation rec {
+              pkgs.stdenv.mkDerivation {
                 name = "bpftrace";
 
                 src = self;
--- a/flake.nix
+++ b/flake.nix
@@ -74,33 +74,39 @@
           # Define lambda that returns a derivation for bpftrace given llvm version as input
           mkBpftrace =
             llvmVersion:
-              with pkgs;
               pkgs.stdenv.mkDerivation rec {
                 name = "bpftrace";
 
                 src = self;
 
-                nativeBuildInputs = [ cmake ninja bison flex gcc clang ];
+                nativeBuildInputs = [
+                  pkgs.bison
+                  pkgs.clang
+                  pkgs.cmake
+                  pkgs.flex
+                  pkgs.gcc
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -25,51 +25,41 @@
...
+              # then nix will fail the build with such an error message:
+              # hash mismatch in fixed-output derivation '/nix/store/m1ga09c0z1a6n7rj8ky3s31dpgalsn0n-source':
+              # specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
+              # got:    sha256-173gxk0ymiw94glyjzjizp8bv8g72gwkjhacigd1an09jshdrjb4
+              sha256 = "sha256-+L/rbp0a3p4PHq1yTJmuMcNj0gT5sqAPeaNRo3Sh6U8=";
+            };
+          };
 
-          pkgs = import nixpkgs { inherit system; };
+          # Override to specify build should use the specific bcc we want.
```
jordalgo, line 51:
```suggestion
          # Override to specify the bcc build we want.
```

flake.nix
```diff
@@ -25,51 +25,41 @@
     flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ]
       (system:
         let
+          pkgs = import nixpkgs { inherit system; };
+
           # The default LLVM version is the latest supported release
           defaultLlvmVersion = 19;
 
-          # Overlay to specify build should use the specific libbpf we want
+          # Override to specify build should use the specific libbpf we want
```
jordalgo, line 33:
```suggestion
          # Override to specify the libbpf build we want
```

flake.nix
```diff
@@ -25,51 +25,41 @@
...
-
-          # We need to use two overlays so that bcc inherits the our pinned libbpf
-          overlayedPkgs = import nixpkgs { inherit system; overlays = [ libbpfOverlay bccOverlay ]; };
+          libbpfOverride = pkgs.libbpf.overrideAttrs {
+            version = libbpfVersion;
+            src = pkgs.fetchFromGitHub {
+              owner = "libbpf";
+              repo = "libbpf";
+              rev = "v${libbpfVersion}";
+              # If you don't know the hash the first time, set:
```
jordalgo, line 41: I know this was from before but can you explain this comment?

viktormalik (approved):
LGTM

flake.nix
```diff
@@ -25,51 +25,41 @@
     flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ]
       (system:
         let
+          pkgs = import nixpkgs { inherit system; };
```
amscanne, line 28: I am finally learning about nix. Can this just be `pkgs = nixpkgs.${system}`?

flake.nix
```diff
@@ -84,21 +74,21 @@
 
                 buildInputs = [
                   asciidoctor
+                  bccOverride
```
amscanne, line 77: I can't comment above, but I think there's some extra complexity here that maybe could be cut out as well. Purely optional, I can also do it later (I've spent a *bit* of time learning nix, and need to do a bit more).

* The "with pkgs" above means that all packages are in the local scope, but then *some* packages are actually prefixed by `pkgs` explicitly. Maybe just individual pkgs here should just be prefixed with `pkgs.` like the LLVM versions? Then you should drop the `with pkgs;` above, and the overrides stand out (you could even just call them "libbpf" and "bcc").
* AFAICT, there is no need for this to be recursive set. Does it work without the `rec` above?

This rec comment applies below, except I think you'd need `drv = self.packages.${system}.default` instead of `drv = default`.

amscanne (approved):

flake.nix
amscanne: I misunderstood what this type is. The `nixpkgs` input is actually a set that represents that given `nixpkgs` module on disk (and it has `outPath` set to something like "/nix/store/...". It needs the system parameter set, and the import makes perfect sense.

flake.nix
danobi: > The "with pkgs" above means that all packages are in the local scope, but then some packages are actually prefixed by pkgs explicitly. Maybe just individual pkgs here should just be prefixed with pkgs. like the LLVM versions? Then you should drop the with pkgs; above, and the overrides stand out (you could even just call them "libbpf" and "bcc").

Right. Originally I wanted to keep the list more terse, but for string interpolation to work on the llvmPackages I needed the prefix anyways.

> AFAICT, there is no need for this to be recursive set. Does it work without the rec above?

Nice, yep. Must've removed the recursive bit in there at some point.

flake.nix
danobi: Yeah, I'll leave a note about a footgun as well

flake.nix
amscanne: I don't think so, beyond `rec` just being a little more complex to interpret. I think that the only difference would be conceptually you're referring to the default package of the fix point versus the default package specified above, so maybe the recursive case is actually more correct.

flake.nix
```diff
@@ -25,82 +25,86 @@
...
-          overlayedPkgs = import nixpkgs { inherit system; overlays = [ libbpfOverlay bccOverlay ]; };
+          libbpf = pkgs.libbpf.overrideAttrs {
+            version = libbpfVersion;
+            src = pkgs.fetchFromGitHub {
+              owner = "libbpf";
+              repo = "libbpf";
+              rev = "v${libbpfVersion}";
+              # Nix uses the hash to do lookups in its cache as well as check that the
+              # download from the internet hasn't changed. Therefore, it's necessary to
+              # update the hash every time you update the source. Failure update the hash
```
jordalgo, line 43:
```suggestion
              # update the hash every time you update the source. Failure to update the hash
```

flake.nix
jordalgo: Thanks. The note helps a lot.

flake.nix
danobi: I need to learn more nix :P. If you think it's more correct to leave it like this then I will

JakeHillion:
Sorry I missed this, you went into a side box! I think this makes sense.

I wonder whether it'd be better to move these dependencies from scoped to that let expression to being packages of the flake? That would make them more accessible outside of `nix {develop,build,run}` and might be a bit neater.

PR #3757, opened 2025-01-30
Warn when probing a partially inlined function

ttreyer:
Functions can exist both as symbol in the program and as inlined instances in the body of another function, at the same time. When probing such a function, the user expects to attach to all calls/instances and gets confused if calls are missing.

This commit adds a warning for uprobes and kprobes that are partially inlined. It looks at the number of locations returned by `dwarf->get_function_locations()`, with `include_inline` set to `True`, and display a warning if there are more than __one__ location. The warning is raised only when DWARF is available and the config `probe_inline` is disabled.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -185,6 +185,20 @@ int BPFtrace::add_probe(const ast::AttachPoint &ap,
 
           locations_from_dwarf = true;
         }
+
+        // Check if the function is inlined and warn
+        // the user about missing calls to inlined instances.
+        if (!config_.get(ConfigKeyBool::probe_inline)) {
+          auto inst_count =
+              dwarf->get_function_locations(probe.attach_point, true).size();
+          if (inst_count > 1) {
+            LOG(WARNING) << "Function \"" << probe.attach_point
+                         << "\" is partially inlined, bpftrace will miss calls "
+                            "to inlined instances.";
+            LOG(HINT)
+                << "Set `probe_inline = 1` config to probe all instances.";
+          }
+        }
       }
     }
 
--- a/.github/include/aot_skip.txt
+++ b/.github/include/aot_skip.txt
@@ -154,6 +154,7 @@ aot.variable.map nested tuple string resize
 aot.variable.map key nested tuple string resize
 aot.variable.map key tuple with casted ints
 aot.dwarf.uprobe without dwarf
+aot.dwarf.uprobe inline without dwarf
 aot.dwarf.uprobe inlined function - func
 aot.dwarf.uprobe inlined function - probe
 aot.dwarf.uprobe inlined function - ustack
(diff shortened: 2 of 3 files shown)


src/bpftrace.cpp
```diff
@@ -185,6 +185,20 @@ int BPFtrace::add_probe(const ast::AttachPoint &ap,
...
+        // the user about missing calls to inlined instances.
+        if (!config_.get(ConfigKeyBool::probe_inline)) {
+          auto inst_count =
+              dwarf->get_function_locations(probe.attach_point, true).size();
+          if (inst_count > 1) {
+            LOG(WARNING) << "Function \"" << probe.attach_point
+                         << "\" is partially inlined, bpftrace will miss calls "
+                            "to inlined instances.";
+            LOG(HINT)
+                << "Set `probe_inline = 1` config to probe all instances.";
```
danobi, line 199: Didn't we leave `probe_inline` undocumented b/c it was incomplete? And do you mind filing that issue about what is missing from `probe_inline`?

src/bpftrace.cpp
```diff
@@ -185,6 +185,20 @@ int BPFtrace::add_probe(const ast::AttachPoint &ap,
...
           locations_from_dwarf = true;
         }
+
+        // Check if the function is inlined and warn
+        // the user about missing calls to inlined instances.
+        if (!config_.get(ConfigKeyBool::probe_inline)) {
+          auto inst_count =
+              dwarf->get_function_locations(probe.attach_point, true).size();
+          if (inst_count > 1) {
+            LOG(WARNING) << "Function \"" << probe.attach_point
```
jordalgo, line 195: Is it easy to get numbers on the number of call sites and use the number of inline locations to indicate a percentage of inlined-ness?

src/bpftrace.cpp
jordalgo: My lingering PR: https://github.com/bpftrace/bpftrace/pull/3628

src/bpftrace.cpp
ttreyer: You'd like to compare how many time the function is inlined vs how many time it is called? What info does that give us?
I'm not sure about getting the number of call sites using lldb, I'll have a look...

jordalgo: > You'd like to compare how many time the function is inlined vs how many time it is called? What info does that give us?

No, I'm just interested in call sites so I can understand, if possible, on how many call sites was this function inlined but it's not a blocker.

ttreyer:
I'm worried about the failing AOT test: I was not expecting the resulting AOT binary to be linked against liblldb, so the Debug Information are read at run-time 🙁 
I'm not familiar with AOT, would you have some pointer on how the AOT binary is created?

danobi:
We are rethinking how we approach inline function probing. Let's punt on this.

PR #3760, opened 2025-01-31
Use `blazesym` for kernel address symbolization

danielocfb:
(supersedes #3474)

Hook up `blazesym` as a way to symbolize kernel addresses. This change puts in the plumbing for building and linking the library and wires kernel address symbolization up. Both a compile time and runtime switch are provided to fall back to BCC as necessary. Currently `blazesym` usage is entirely opt-in.

Please see individual commits for more detailed descriptions.

Will work on user space address symbolization once these changes are in.

--------

Can someone please help with the CI setup? I'd think running the test suite with `USE_BLAZESYM` defined on LLVM 19 or so would be a good first step, but open to recommendations.

--------

##### Checklist


jordalgo:
@danielocfb Sorry for the delay in reviewing. Someone will get to it this week.

man/adoc/bpftrace.adoc
```diff
@@ -3791,6 +3791,12 @@ Available options:
 If the DebugInfo was rewritten by a post-linkage optimisation tool (like BOLT or AutoFDO), it might yield an incorrect address for a probe location.
 This config can force using the Symbol Table, for when the DebugInfo returns invalid addresses.
 
+=== use_blazesym
+
+Default: `0` if built without blazesym support, `1` otherwise
```
jordalgo, line 3796: Hmm, I wonder if we shouldn't just decide to build with blazesym by default and then this can default to 1.

src/ksyms.cpp
```diff
@@ -23,8 +42,66 @@ std::string Ksyms::resolve(uint64_t addr, bool show_offset)
   return symbol.str();
 }
 
-Ksyms::~Ksyms()
+#ifdef USE_BLAZESYM
```
jordalgo, line 45: eh, I don't think we need to guard the creation of these functions for something we're fairly sure we're going to switch to.

src/ksyms.cpp
```diff
@@ -23,8 +42,66 @@ std::string Ksyms::resolve(uint64_t addr, bool show_offset)
...
+std::optional<std::string> Ksyms::resolve_blazesym_int(uint64_t addr,
+                                                       bool show_offset)
 {
-  if (ksyms_)
-    bcc_free_symcache(ksyms_, -1);
+  if (symbolizer_ == nullptr) {
+    symbolizer_ = blaze_symbolizer_new();
+    return std::nullopt;
```
jordalgo, line 51: why do we want to return here?

danielocfb: I think intention was to check `symbolizer_` for `NULL` again. Will fix. Thanks for catching.

src/main.cpp
```diff
@@ -337,6 +338,17 @@ static void parse_env(BPFtrace& bpftrace)
                     "BPFTRACE_MAX_MAP_KEYS instead.";
     config_setter.set(ConfigKeyInt::max_map_keys, x);
   });
+
+  get_bool_env_var("BPFTRACE_USE_BLAZESYM", [&](bool x) {
```
jordalgo, line 342: We don't need t make this an env var too. Let's just leave it as a config.

man/adoc/bpftrace.adoc
danielocfb: Not sure we should make it the default right away. Why not do it as a second step?

jordalgo (commented):
I don't think we should surface the config to users yet (in either the adoc or the help menu). Ultimately it will just be used as an escape hatch if blazesym is doing something weird that we need to debug.

src/main.cpp
jordalgo: See my comment below, I don't plan on us going back and forth on symbolication strategy. I seriously doubt that blazesym won't be significantly better than what we have now.

man/adoc/bpftrace.adoc
jordalgo: Sure, we can wait. No big deal.

src/main.cpp
danielocfb: Okay, but somebody may do; at the very least I know I do. Given that we will have both in place for a while, I very much believe that it makes sense to make it easy to switch to compare, from an engineering perspective. Put it differently: what is the problem with these five odd lines of code that makes them so troublesome to include?

jordalgo: Ok, if it's helping with development (and not surfaced to users) then we can leave it. Let's just make a note to yank it out when blazesym is fully integrated and the default.

src/ksyms.cpp
```diff
@@ -28,9 +42,67 @@ std::string Ksyms::resolve(uint64_t addr, bool show_offset)
...
+std::optional<std::string> Ksyms::resolve_blazesym_int(uint64_t addr,
+                                                       bool show_offset)
 {
-  if (ksyms_)
-    bcc_free_symcache(ksyms_, -1);
+  if (symbolizer_ == nullptr) {
+    symbolizer_ = blaze_symbolizer_new();
+    if (symbolizer_ == nullptr)
+      return std::nullopt;
+  }
```
viktormalik, line 53: Note: if `blaze_symbolizer_new` fails for some reason, this will keep calling it for every symbolization request. I'm not sure how big of an issue that could be, just wanted to point it out.

src/ksyms.cpp
```diff
@@ -23,8 +42,67 @@ std::string Ksyms::resolve(uint64_t addr, bool show_offset)
...
+}
+
+std::string Ksyms::resolve_blazesym(uint64_t addr, bool show_offset)
+{
+  if (auto sym = resolve_blazesym_int(addr, show_offset)) {
+    return *sym;
+  } else {
+    std::ostringstream symbol;
+    symbol << reinterpret_cast<void *>(addr);
+    return symbol.str();
```
viktormalik, line 94: This is the same fallback as in `resolve_bcc`. Would it make sense to unify them (e.g. for the case we want to update it)?

viktormalik (commented):
Looks great!

Just a few comments, mostly regarding code style.

src/CMakeLists.txt
```diff
@@ -69,6 +69,44 @@ add_executable(${BPFTRACE}
...
+      Corrosion
+      GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
+      GIT_TAG v0.5.1
+  )
+  FetchContent_MakeAvailable(Corrosion)
+
+  FetchContent_Declare(
+    blazesym
+    GIT_REPOSITORY https://github.com/libbpf/blazesym.git
+    GIT_TAG ccc02c14cc79b86018d2f56764448fb86a984a2a
```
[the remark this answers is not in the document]
danobi: Let's also move the cmake stuff into nix. I will help with that.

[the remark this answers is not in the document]
viktormalik: Absolutely, also will require less CMake changes when we switch to system-provided blazesym.

src/ksyms.h
```diff
@@ -19,5 +21,15 @@ class Ksyms {
 private:
   const Config &config_;
   void *ksyms_{ nullptr };
+
+#ifdef USE_BLAZESYM
+  blaze_symbolizer *symbolizer_{ nullptr };
```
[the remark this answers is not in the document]
danielocfb: Agreed; my thinking was that if the bcc logic gets removed soon-ish anyway, this is acceptable. We could prefix things with the library as you suggest, but it's not like there really is much of a chance for confusion in a 100 line module, especially with an unambiguously typed new member like this,. And if we rename it will just be a bunch of churn now and once we revert back due to convergence to a single symbolizer.

Anyway, if you feel strongly I am happy to change it; just let me know.

src/CMakeLists.txt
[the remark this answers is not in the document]
danobi: I'm a bit backed up behind a few other items. Feel free to land with cmake if that's the only blocker. I'll come by later at some point to nix-ify it.

src/CMakeLists.txt
```diff
@@ -69,6 +69,43 @@ add_executable(${BPFTRACE}
...
 )
 
+# TODO: Add support for using "system" blazesym and linking dynamically. Then,
+#       also honor `STATIC_LINKING` properly.
+if (USE_BLAZESYM)
+  if (NOT STATIC_LINKING)
+    # It's not a strict requirement (we could just link *only* blazesym
+    # statically), but for now we enforce program-wide static linking.
+    message(FATAL_ERROR "blazesym usage requires static linking; consider setting STATIC_LINKING")
+  endif()
```
danobi, line 79: Why is only static linking supported? Aside from the fact users generally don't have a statically linked installation (which your comments allude to), we won't get coverage in CI. I don't think it's a good idea to land without CI coverage - untested code quickly bitrots.

src/CMakeLists.txt
danobi, line 1: Both compile time and runtime switching seems overkill to me. The last time we did it, it was for the `--unsafe` uprobe thingy and it got somewhat confusing by the end. Let's try to keep config simple if we can. The commit message says it's for easily switching back and forth at runtime, but that's not strictly true. You have to restart the tracing session. At that point, why not just have two cmake build directories? One for blazesym and one for bcc.

src/blazesym_int.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#ifdef USE_BLAZESYM
+#include <blazesym.h>
+
+#define BLAZESYM_OPT                                                           \
+  _Pragma("GCC diagnostic ignored \"-Wmissing-field-initializers\"")
```
danobi, line 7: Use pragma push and pragma pop instead? This seems unbounded - a transitive include of this file could add this for a bunch of TUs. Would be good to tightly scope where this warning is disabled so it's obvious what it's there for.

src/ksyms.cpp
```diff
@@ -2,35 +2,110 @@
...
+  uint64_t addrs[1] = { addr };
+
+  const blaze_syms *syms = blaze_symbolize_kernel_abs_addrs(
+      symbolizer_, &src, addrs, ARRAY_SIZE(addrs));
+  if (syms == NULL)
+    return std::nullopt;
+
+  const blaze_sym *sym = &syms->syms[0];
+  if (sym->name == nullptr) {
+    blaze_syms_free(syms);
```
danobi, line 81: Use newly added SCOPE_EXIT scopeguard instead: https://github.com/bpftrace/bpftrace/blob/db6e23dc2427cfac1a5fe94616af281d4924a661/src/scopeguard.h#L39

src/CMakeLists.txt
jordalgo: > Both compile time and runtime switching seems overkill to me.

I thought that too but I think it's only temporary while we do the switch over. I don't imagine we're going to want to support bcc symbolication for very long once blazesym can  do it for user and kernel addresses. I guess all I'm saying is that I agree with you but shouldn't be a blocker.

src/ksyms.h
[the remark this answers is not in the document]
viktormalik: Right, ok, if you plan to remove the bcc part in foreseeable future then it makes sense not to introduce more churn than necessary.

src/CMakeLists.txt
danielocfb: Sounds good, I can remove that part.

src/CMakeLists.txt
danielocfb: What Jordan said. Two build directories is yet another thing that can be forgotten to be kept up-to-date.

src/CMakeLists.txt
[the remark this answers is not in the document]
viktormalik: If we stick with CMake for now, my original request still stands - can we bind this to some numbered release of blazesym?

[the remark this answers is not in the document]
danielocfb: It's a bit of a chicken and egg problem. I don't want to cut a new `blazesym` release right now, because we *know* that we want https://github.com/libbpf/blazesym/pull/991 (+ C API bindings), meaning we need a new release *after* that lands. Then I'd want to make sure that this functionality works properly with `bpftrace`, which means implementing the user address symbolization bits that will use it. Now, I don't really want to get that started until the kernel symbolization bits have settled, to not have to constantly rebase and resolve merge conflicts while we change the underlying plumbing.

So, basically, the plan was to get this PR in using the currently pinned SHA, implement and land the user space symbolization bits, and then cut a new `blazesym` release containing everything `bpftrace` needs. Then we can pull in the tagged/numbered release. Does this make sense? Or why is it so important to have a tagged release right here right now?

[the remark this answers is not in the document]
jordalgo: @danielocfb Makes sense to me. I also think it's low risk since it's currently opt-in anyway.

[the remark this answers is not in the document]
viktormalik: Sounds good to me, thanks for the explanation!

viktormalik (approved):

jordalgo (approved):
Great work!

src/CMakeLists.txt
jordalgo: @danobi Going to merge later today unless you think this is a big issue that needs to be addressed now.

danobi: If the plan is to remove it after migration, then that's fine

PR #3762, opened 2025-02-01
man: bpftrace: Fix name of sample script 'bsize.bt'

carnil:
The manual uses he bsize.bt script as example but when naming it the first time it is mentioned as bsize.d. Fix the naming to be consistent with the usual suffix for bpftrace scripts and as used just in the following paragraph.

##### Checklist


jordalgo (approved):
Good catch. Thanks for the fix.

PR #3763, opened 2025-02-02
ci: Run entire test suite under ASAN

### LINKED ISSUE #3748: Plug memory leaks
At one point we were somewhat ASAN clean: #1047. But since then we've regressed.

Let's fix this again and prevent future backslides.

This issue is to:

1. Fix existing memory leaks. See #3747 for all the current failures. There seems to be only a few sources, so hopefully it's not too difficult.
2. Run all existing tests under ASAN. The diff in #3747 should be sufficient.
3. Delete `memleak-tests.sh` and associated CI jobs (https://github.com/bpftrace/bpftrace/blob/431466574eb54694f2de080584726739c398df93/.github/workflows/ci.yml#L60-L69)

The existing tests in `memleak-tests.sh` should be more than compensated for in the runtime test suite.

danobi:
This closes #3748.

Note we are disabling LLDB in CI (and by extension appimages) here. LLDB is quite a problematic dependency. Its problems (in order of severity):

1. It runs the python interpreter at startup.
    * Yes, we are running python any time a uprobe is used. No, this cannot be disabled. I've tried to [work with LLDB folks](https://github.com/llvm/llvm-project/pull/105757) on this. End result is general agreement LLDB is not the right dependency to take on.
2. It leaks memory like crazy.
    * We're talking thousands, at ~150K memory. Many of the leaks [come from python](https://github.com/llvm/llvm-project/blob/e566313a1fac1b290c98454cc52b485ae4f644c5/lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp#L3170-L3184). There are a handful more from elsewhere, but the bottom line here is there are assumptions within lldb about how it's built and where it's run. It is not designed to be a general purpose dependency. The leaks themselves are minor. But ASAN clean status is a huge boon for reliability. 
3. It adds ~100 MB compressed to the appimage.
    * The appimage is quite popular actually. While 100MB is not necessarily that much on a modern system, it's quite a lot to carry for what amounts to a handful of undocumented features.

I plan to follow up with a plan for the future of LLDB in bpftrace. But for now, I think the benefits of being ASAN clean outweigh the limited set of features LLDB affords.

##### Checklist


jordalgo (approved):
fASANtastic!

danobi:
I have somewhat strong opinions on this so I am merging. If there's dissenting opinion, we can discuss during OH next week.

danobi:
Just did some tests. I was mistaken earlier - lldb only adds ~10 MB to compressed appimage size. But I think point still stands.

PR #3765, opened 2025-02-04
Add support for all types in ternary

jordalgo:
Previously only int, string, and none
types were supported results of ternary
expressions. This adds support for the
others.

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1977,6 +1977,9 @@ void CodegenLLVM::visit(Ternary &ternary)
     buf = b_.CreateGetStrAllocation("buf", ternary.loc);
     uint64_t max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
     b_.CreateMemsetBPF(buf, b_.getInt8(0), max_strlen);
+  } else if (!ternary.type.IsIntTy() && !ternary.type.IsNoneTy()) {
+    buf = b_.CreateAllocaBPF(ternary.type);
+    b_.CreateMemsetBPF(buf, b_.getInt8(0), ternary.type.GetSize());
   }
 
   auto scoped_del = accept(ternary.cond);
@@ -2007,22 +2010,7 @@ void CodegenLLVM::visit(Ternary &ternary)
     phi->addIncoming(left_expr, left_block);
     phi->addIncoming(right_expr, right_block);
     expr_ = phi;
-  } else if (ternary.type.IsStringTy()) {
-    b_.SetInsertPoint(left_block);
-    auto scoped_del_left = accept(ternary.left);
-    b_.CreateMemcpyBPF(buf, expr_, ternary.type.GetSize());
-    b_.CreateBr(done);
-
-    b_.SetInsertPoint(right_block);
-    auto scoped_del_right = accept(ternary.right);
-    b_.CreateMemcpyBPF(buf, expr_, ternary.type.GetSize());
-    b_.CreateBr(done);
-
-    b_.SetInsertPoint(done);
-    expr_ = buf;
-    if (dyn_cast<AllocaInst>(buf))
-      expr_deleter_ = [this, buf]() { b_.CreateLifetimeEnd(buf); };
-  } else {
+  } else if (ternary.type.IsNoneTy()) {
     // Type::none
     b_.SetInsertPoint(left_block);
     {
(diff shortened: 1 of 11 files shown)


danobi (commented):
lgtm. but missing codegen test?

tests/semantic_analyser.cpp
```diff
@@ -500,6 +500,8 @@ TEST(semantic_analyser, ternary_expressions)
   test(R"(kprobe:f { @x = pid < 10000 ? "lo" : "high" })");
   test("kprobe:f { pid < 10000 ? printf(\"lo\") : exit() }");
   test("kprobe:f { curtask ? printf(\"lo\") : exit() }");
+  test(R"(kprobe:f { $x = pid < 10000 ? buf("mystr", 5) : buf("mystr", 4) })");
+  test(R"(kprobe:f { $x = pid < 10000 ? macaddr(arg0) : macaddr(arg1); })");
```
danobi, line 504: Maybe table driven test with all the types?

jordalgo:
> but missing codegen test?

Ah good call. Will add.

tests/semantic_analyser.cpp
jordalgo: Feel like a bit of overkill since we don't have separate branches for everything that's not a string, int, or none but I'll look into it.

danobi: Yeah, perhaps overkill. But upshot is it encodes intent to support all types

jordalgo: Added a lot more types and support for different sized tuples.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -4021,6 +4040,31 @@ bool SemanticAnalyser::update_string_size(SizedType &type,
...
 }
 
+SizedType SemanticAnalyser::create_merged_tuple(const SizedType &left,
+                                                const SizedType &right)
+{
+  assert(left.IsTupleTy() && right.IsTupleTy() &&
+         (left.GetFieldCount() == right.GetFieldCount()));
+
+  SizedType type;
+  type.SetSize(std::max(left.GetSize(), right.GetSize()));
```
viktormalik, line 4050: This is IMO incorrect. You can have two tuples like `((int64)0, (int32)1)` and `((int32)1, (int64)0)` which will both have size 12 but the merged tuple should have the size 16 (`(int64, int64)`).

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2054,6 +2057,33 @@ void CodegenLLVM::visit(Ternary &ternary)
     b_.CreateBr(done);
     b_.SetInsertPoint(done);
     expr_ = nullptr;
+  } else {
```
viktormalik, line 2060: This new branch could be merged with the branch for `Type::string`, couldn't it?

tests/semantic_analyser.cpp
```diff
@@ -496,29 +496,93 @@ kprobe:f / @mymap / { @mymap = "str" }
...
-  test("kprobe:f { curtask ? printf(\"lo\") : exit() }");
   test(R"(kprobe:f { @x = pid < 10000 ? printf("lo") : cat("/proc/uptime") })",
        3);
+  test("struct Foo { int x; } kprobe:f { curtask ? (struct Foo)*arg0 : (struct "
+       "Foo)*arg1 }",
+       1);
+  test("struct Foo { int x; } kprobe:f { curtask ? (struct Foo*)arg0 : (struct "
+       "Foo*)arg1 }");
+  test(
+      R"(kprobe:f { pid < 10000 ? ("a", "hellolongstr") : ("hellolongstr", "b") })");
```
viktormalik, line 532: How is this test different from the `{ "\"lo\"", "\"high\"" },` variant above?

jordalgo: It's testing tuple resizing (the `create_merged_tuple` function)

src/ast/passes/semantic_analyser.cpp
jordalgo: Good catch. I think I can get rid of this line entirely as it gets ignored anyway for the case above, as the tuple gets resized based on it's elements anyway.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3987,6 +4006,30 @@ bool SemanticAnalyser::update_string_size(SizedType &type,
...
+    assert(leftTy.GetTy() == rightTy.GetTy());
+    if (leftTy.IsTupleTy()) {
+      new_elems.push_back(create_merged_tuple(leftTy, rightTy));
+    } else {
+      new_elems.push_back(leftTy.GetSize() > rightTy.GetSize() ? leftTy
+                                                               : rightTy);
+    }
+  }
+  type = CreateTuple(bpftrace_.structs.AddTuple(new_elems));
+  return type;
```
viktormalik, line 4030: Nit, you can probably just do
```suggestion
  return CreateTuple(bpftrace_.structs.AddTuple(new_elems));
```
and drop the `SizedType type;` decl above the loop.

Just cleaner code.

viktormalik (approved):
LGTM, just one code style nit.

PR #3768, opened 2025-02-05
Fix extant memory leaks

danobi:
This fixes all memory leaks that can be exposed by our test suite.
IOW, I ran the entire test suite under BUILD_ASAN=1.

This excludes the multiple thousands of leaks from LLDB. I will
follow up with a broader plan for dealing with that after.

The end goal is we run all CI jobs with ASAN and then delete
the memleak-tests.sh jobs from CI to repurpose that capacity
for vmtest jobs.

##### Checklist

### DIFF
--- a/src/scopeguard.h
+++ b/src/scopeguard.h
@@ -1,0 +1,39 @@
+#pragma once
+
+#include <functional>
+
+// Need two levels of indirection here for __LINE__ to correctly expand
+#define _CONCAT2(a, b) a##b
+#define _CONCAT(a, b) _CONCAT2(a, b)
+#define _ANON_VAR(str) _CONCAT(str, __LINE__)
+
+namespace bpftrace {
+
+enum class ScopeGuardExit {};
+
+class ScopeGuard {
+public:
+  explicit ScopeGuard(std::function<void()> fn)
+  {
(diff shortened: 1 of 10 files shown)


src/scopeguard.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <functional>
```
jordalgo, line 3: yay! I've been craving a scope guard but was too lazy to do it myself. Is this implementation mostly copied from std::experimental::scope_exit ?

tests/scopeguard.cpp
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    [&]() {
+      SCOPE_EXIT
+      {
+        x++;
+      };
+
+      throw std::runtime_error("exception");
+    }();
+  } catch (const std::exception &) {
```
jordalgo, line 54: super nit. I might add an ordering part of the test here where you expect x to be 6 and then increment it and expect x to be 7 below.

jordalgo (approved):
Nice work fixing all the 💧 s!

src/scopeguard.h
danobi: No, I wrote this

PR #3771, opened 2025-02-06
ci: Bring back nix caching

danobi:
Due to GHA changing the cache API [0], we had to delete the magic-nix-cache-action. As a result, we saw about a 20% regression in CI times.

Now that we have access to the open source tier of flakehub cache, let's bring it back. Flakehub cache is a bit more interesting than GHA cache as developers/users can potentially connect to it and access a pre-warmed cache built from CI.

[0]: https://determinate.systems/posts/magic-nix-cache-free-tier-eol/

##### Checklist

### DIFF
--- a/.github/workflows/binary.yml
+++ b/.github/workflows/binary.yml
@@ -14,9 +14,17 @@ jobs:
   build-and-upload:
     runs-on: ubuntu-latest
 
+    # For flakehub cache
+    permissions:
+      id-token: write
+      contents: read
+
     steps:
     - uses: actions/checkout@v4
-    - uses: DeterminateSystems/nix-installer-action@v11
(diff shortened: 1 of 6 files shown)


danobi:
I'm seeing:

> FlakeHub Cache is disabled due to missing or invalid token
> If you're signed up for FlakeHub Cache, make sure that your Actions config has a `permissions` block with `id-token` set to "write" and `contents` set to "read"

in the cache setup step. But I presume it's b/c this isn't checked in yet. Likely for security reasons a random PR can't start pushing to cache. Either way, should be fairly safe to land - a cache miss is a noop w.r.t. current configuration.

jordalgo (approved):

Issue #3525, opened 2024-10-18
RFC: session probes


viktormalik:
### Is your feature request related to a problem? Please describe.

Kernel 6.10 has introduced so-called "kprobe session probes" ([patch series](https://lore.kernel.org/bpf/20240430112830.1184228-1-jolsa@kernel.org/)) which are an extension of kprobe-multi that allow a single BPF program attached to both entry and exit of the function. The `bpf_session_is_return` kfunc can be used to determine if the current attachment is in the entry or the exit probe. In addition, the entry and the exit probes can share data via a "cookie" (available via the `bpf_session_cookie` kfunc).

Once we have kfuncs support (#3523), we should be able to use session probes. This RFC is to discuss the necessary syntax decisions.

### Describe the solution you'd like

There are two solutions I currently have in mind:

1. Introduce a new probe type

    This should allow this simple program to compute average runtime of each VFS function:
    ```
    ksession:vfs_*
    {
        if (is_return())
            cookie = nsecs;
        else
            @runtime[func] = avg(nsecs - cookie);
    }
    ```

2. Detect if user uses `kprobe:X` and `kretprobe:X` and transparently merge his probes into one session probe. This is purely for peformance benefits as session probes are faster, especially when using wildcard attachment.

Overall, I can see benefits in both approaches. Session probes are mostly about performance and the second one uses it transparently without user even noticing, however, it fails to leverage all the benefits (sharing code between the probes, usage of cookie) so the first approach can be even better but requires the user to explicitly use it.

ajor:
I'll put my vote behind the second option (transparently use session probes where applicable). It'll benefit all existing scripts and we won't have to tell users to forget the old kprobe/kretprobe patterns and to use this new probe type on newer kernels.

Code sharing we'll soon support through user defined functions anyway. I suppose the cookie will perform better, but we can already achieve similar things with maps, plus the cookie is more limited (u64 only).

jordalgo:
I'm also for 2nd option for now (for the reasons @ajor mentioned). I think we can punt on the decision to add a new probe type later when there is more of an explicit need and it's available on more kernels/distros.

batu3H:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

viktormalik:
[...]

PR #3772, opened 2025-02-06
Introduce automatic session kprobes

viktormalik:
This implements the option (2) from #3525.

Introduce session kprobes to bpftrace by automatically using them when possible. This is done by detecting whether the script contains wildcarded kprobe and kretprobe with the same attach points and, if so, merging their codes into a single session probe.

In pseudo-code, for the following probes:
    
    kprobe:vfs_* { XXX }
    kretprobe:vfs_* { YYY }
    
a single session probe is created:

    kprobe:vfs_* { if (bpf_session_is_return()) { YYY } else { XXX } }
    
Here, `bpf_session_is_return` is a kernel function (kfunc).

The implementation introduces a new `SESSION` probe expansion type and a new AST pass `ProbeAnalyser` which tries to find kprobe/kretprobe with matching attach points.

##### Checklist


viktormalik:
I'm not sure whether to add a changelog item here. It is fully transparent for users, however, power users may observe differences - faster attachment for a large number of probes, single BPF program/link created, and maybe few others.

src/ast/passes/probe_analyser.cpp
```diff
@@ -0,0 +1,57 @@
...
+        continue;
+      auto &other_ap = *other_probe->attach_points[0];
+      if (probetype(other_ap.provider) != ProbeType::kretprobe ||
+          other_ap.expansion != ExpansionType::MULTI) {
+        continue;
+      }
+
+      if (ap.target == other_ap.target && ap.func == other_ap.func) {
+        ap.expansion = ExpansionType::SESSION;
+        ap.ret_probe = other_probe;
```
amscanne, line 34: I really think that it would be better to do an actual transformation here (e.g. remove the probe, replace with another AST node that reflects a `SessionProbe` or other such specialized case) rather than leaving breadcrumbs and an implicit requirements that codegen do the right thing with respect to new and old probes.

I know that the current structure doesn't actually let us do this (and I need to finish and submit my own contribution there), but I'm just trying to raise awareness.

viktormalik: Good point. We used to have a `Mutator` class but it was removed in #3347 as it was kind of broken.

I think we would be able to identify quite a number of cases where we could do actual AST transformations instead of having classes with a lot of fields specific to certain use-cases only (like `Probe` or so).

jordalgo:
> I'm not sure whether to add a changelog item here.

I think you should. I always default to being transparent about these kinds of things and you get to claim perf improvements.

jordalgo:
@viktormalik Just curious, why is this change only implemented for kprobes with wildcards? Why couldn't the same transformation be done for specific kprobes that have both an entry/exit? Is it cause it's linked to the multi implementation?

viktormalik:
> @viktormalik Just curious, why is this change only implemented for kprobes with wildcards? Why couldn't the same transformation be done for specific kprobes that have both an entry/exit? Is it cause it's linked to the multi implementation?

Good point, I had the same thing in mind. The main reason is that I realized that we could also switch generic entry/exit kprobes (i.e. no addresses/offsets) to use multi-probes. In theory, they should be slightly faster (using fprobe rather than `int3`). I need to give it a bit of thought but if we go down that road, the current implementation of this PR will also enable auto-session probes. So, I'd keep session probes bound to multi-probes (which is also convenient as they share many implementation traits).

tests/runtime/probe
```diff
@@ -278,6 +278,18 @@ REQUIRES bpftool
 REQUIRES_FEATURE uprobe_multi
 BEFORE ./testprogs/uprobe_test
 
+NAME kprobe_session_progs
+RUN {{BPFTRACE}} --unsafe -e 'kprobe:ksys_* { printf("progs: "); system("bpftool prog | grep kprobe | grep ksys_ | wc -l"); exit(); } kretprobe:ksys_* { exit(); }'
```
jordalgo, line 282: We don't want to print something in the exit probe here and confirm it was actually printed?

tests/runtime/probe
```diff
@@ -278,6 +278,18 @@ REQUIRES bpftool
...
+EXPECT progs: 1
+REQUIRES bpftool
+REQUIRES_FEATURE kprobe_session
+
+NAME kprobe_session_links
+RUN {{BPFTRACE}} --unsafe -e 'kprobe:ksys_* { printf("link: "); system("bpftool link | grep kprobe_multi | wc -l"); exit(); } kretprobe:ksys_* { exit(); }'
```
jordalgo, line 288: also why they use of the system function?

jordalgo (commented):
Overall this looks great. Few small questions/nits.

Nit: on one of the commit messages: "Add a new parameter to BTF::resolve_args which allows to disable the check and disable it when resolving args of kfuncs." - Maybe just rephrase to "Add a new parameter to BTF::resolve_args which allows disabling of the check, which will be used for kfuncs."

jordalgo:
> The main reason is that I realized that we could also switch generic entry/exit kprobes (i.e. no addresses/offsets) to use multi-probes.

IIUC, it is because of the commonality to multi probes, which we might eventually want to use for single kprobe attachment at some point (not in this PR).

src/main.cpp
```diff
@@ -544,7 +544,7 @@ Args parse_args(int argc, char* argv[])
       case Options::NO_FEATURE: // --no-feature
         if (args.no_feature.parse(optarg)) {
           LOG(ERROR) << "USAGE: --no-feature can only have values "
-                        "'kprobe_multi,uprobe_multi'.";
+                        "'kprobe_multi,kprobe_session,uprobe_multi'.";
```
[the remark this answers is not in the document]
viktormalik: Not at the moment. And since it's an internal-only option (not documented) for testing purposes, I wouldn't bother changing it.

tests/runtime/probe
viktormalik: Because that's currently the only way to run bpftool and check that the amount of BPF progs/links is as expected. I'm considering adding a new kind of check, `EXPECT_CMD`, which will execute a command and check the return code but that will be done in a different PR.

jordalgo: >  check that the amount of BPF progs/links is as expected

Ah, that part totally went over my head. Maybe a comment wouldn't be a bad idea - for us slow folk 😄

tests/runtime/probe
jordalgo: > it should have the same system call in the kretprobe part.

But we should still assert that both the entry/exit paths were called.

tests/runtime/probe
viktormalik: Sure, I added the comment.

tests/runtime/probe
viktormalik: So I ended up using one large tests for all - checking that both probes fire and that the number of progs/links is 1. We could theoretically split it into multiple cases but IMO it's not that important feature that it would need multiple test. And I plan to drop the `system()` calls in future so the test will be much simplified.

jordalgo: Nice! LGTM - A super nit would be to put the `RUN` part into a separate script for easier readability but not a blocker.

viktormalik: Let's just keep it here for now. If I manage to move the `system` calls into separate `EXPECT` checks, the remainder program will be very simple.

jordalgo (approved):
Nice work!

(although looks like there are still some test failures)

viktormalik:
> (although looks like there are still some test failures)

Yeah, ASAN is complaining, working on it.

viktormalik:
So that memleak comes from LLVM and only happens for version <=16. There *may be* a bug in our code but since there's no memleak on newer LLVM versions, it suggests that it was more of a bug in LLVM which was fixed in LLVM 17.

I'm not sure how to proceed in this case. Since the memleak comes from LLVM, it's not possible to skip ASAN for the problematic function. Perhaps it would make sense to run tests under ASAN only for the latest LLVM to minimize the probability of getting memleaks from LLVM?

cc @danobi

viktormalik:
> Perhaps it would make sense to run tests under ASAN only for the latest LLVM to minimize the probability of getting memleaks from LLVM?

Or we can turn off ASAN for LLVM <= 16 which will cover this case and still get us better ASAN coverage: #3782.

viktormalik:
The first commit https://github.com/bpftrace/bpftrace/pull/3772/commits/b9c1470eb50e3d84d679752a56e65fb89f52e07c adds a new way to suppress LeakSanitizer warnings in CI (inspired by https://github.com/danobi/bpftrace/commit/fd9ee029e02e719ccf996bd6354f4707e2c074a2). See commit message for details - it adds both a global suppression file `.github/include/leaks.txt` as well as a way to pass additional suppression files per-CI-job via `LEAK_SUPPRESS`.

The latter option is used to suppress the above leak on LLVM <=16.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2799,10 +2799,38 @@ void CodegenLLVM::generateProbe(Probe &probe,
...
+                    exit_probe,
+                    entry_probe);
+
+    // Entry probe
+    b_.SetInsertPoint(entry_probe);
+    auto scoped_del_entry = accept(probe.block);
+    createRet();
 
-  createRet();
+    variables_.clear();
```
jordalgo, line 2824: Super nit: but might be good to have a codegen or runtime tests that exercises this `clear` as it's pretty important.

jordalgo (approved):

src/ast/passes/codegen_llvm.cpp
viktormalik: Indeed, it looks like this is not even needed as (1) there's always just a single branch executed and (2) even if it was possible to execute both, since we have two separate probes, semantic analyser will make sure that each probe first assigns a scratch variable before reading it so the value from the other branch cannot get there.

I think that we would be fine using just one set of variables but it still feels a bit cleaner and safer to me to do the `clear()` here.

jordalgo: I think if we have 2 reasons why it's not needed we should remove it (it's code smell) or add a comment as to why we're keeping it.

jordalgo: If it ends up biting us in the ass, I'll buy you a beer 🍺

src/ast/passes/probe_analyser.cpp
danobi: The new mutation infra is in now. Is it worth trying to use in this PR? Agreed a few other places could use it as well. But in general I'd agree it's better to lower the AST into structurally unambiguous representations when possible.

jordalgo: Might be nice but I don't see that as being a blocker - could be done as part of a larger, follow-up refactor.

viktormalik:
I was debugging the failure in #3683 and noticed that the issue prevents us from using **any kfunc** on LLVM <= 16. It has not been captured by CI here as the runner VMs probably don't have session probes enabled. It seems to be a bug in LLVM (very much related to the above memleak) so I'm marking this as do-not-merge until I investigate the issue better.

danobi:
Good find. I forgot about that. Let's make sure whatever fix we add is generic for all kfuncs.

viktormalik:
> Good find. I forgot about that. Let's make sure whatever fix we add is generic for all kfuncs.

This turned out to be more tricky than I expected. It is fixed by #3809 for now but I had to disable LLVM IR verification on LLVM <=15 as I was hitting an error which I couldn't get rid of. But good news is that the fix in #3809 resolves the memleak so we can drop memleak suppression from this PR.

viktormalik:
#3809 has resolved the memleak so I dropped the code for memleak suppression and rebased. Should be now ready for review and merge.

src/ast/passes/probe_analyser.cpp
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+}
+
+void ProbeAnalyser::analyse()
+{
+  visit(ctx_.root);
+}
+
+Pass CreateProbesPass()
```
jordalgo, line 46: naming nit: 
```suggestion
Pass CreateProbePass()
```

jordalgo (approved):

src/ast/passes/probe_analyser.cpp
viktormalik: Fixed.

src/ast/passes/probe_analyser.cpp
viktormalik: @amscanne @danobi are you ok with merging this in the current form?

jordalgo:
@viktormalik @danobi Any reason we're waiting on merging this?

viktormalik:
@jordalgo not from my side, I just rebased to the new pass infra.

jordalgo:
might need another rebase 😬

viktormalik:
> might need another rebase 😬

Done. With all the changes in the pass infrastructure going on, let's merge this ASAP to avoid further conflicts.

Still, it would be good to check if I did the new pass correctly.

jordalgo (approved):
@viktormalik I'm fine for you to merge this first since you've been waiting longer and then @amscanne can update the pass logic as necessary.

PR #3778, opened 2025-02-10
Use an AOT allow list instead of a skip list

jordalgo:
This is so we don't have to keep adding tests
to the skip list.

##### Checklist


danobi (approved):
thanks for doing this

PR #3782, opened 2025-02-11
CI: do not build ASAN for LLVM <=16

viktormalik:
While working on #3772, I ran into what looks like a memleak in older LLVM versions.

It looks like LLVM versions <=16 contain a memleak in `DIBuilder` which only manifests when creating debug info for kfuncs. We now only have a single kfunc (`bpf_map_sum_elem_count`) which is only used for LLVM >=17 so the issue did not appear, yet, however, adding new kfuncs will trigger it.

It looks like this is a bug in older LLVM versions so the simplest thing for us is to disable ASAN on these versions.

This adds a new CI env variable `BUILD_ASAN` (turned on by default) which allows to disable ASAN for specific jobs.

##### Checklist

### DIFF
--- a/.github/include/ci.py
+++ b/.github/include/ci.py
@@ -45,6 +45,7 @@ NIX_TARGET_KERNEL = os.environ.get("NIX_TARGET_KERNEL", "")
 TOOLS_TEST_OLDVERSION = os.environ.get("TOOLS_TEST_OLDVERSION", "")
 TOOLS_TEST_DISABLE = os.environ.get("TOOLS_TEST_DISABLE", "")
 AOT_ALLOWLIST_FILE = os.environ.get("AOT_ALLOWLIST_FILE", "")
+BUILD_ASAN = os.environ.get("BUILD_ASAN", "1")
 
 
 class TestStatus(Enum):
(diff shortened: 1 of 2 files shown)


danobi:
Looking at your other PR, here is the failure:
```
==14163==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x7ffff78dcb88 in operator new(unsigned long) (/nix/store/ybjcla5bhj8g1y84998pn4a2drfxybkv-gcc-13.3.0-lib/lib/libasan.so.8+0xdcb88)
    #1 0x7fffebc7d205 in llvm::MDNode::operator new(unsigned long, unsigned long, llvm::Metadata::StorageType) (/nix/store/15ydnad55xckaq7rcbm56hmhp0zm80di-llvm-16.0.6-lib/lib/libLLVM-16.so+0x127d205) (BuildId: a969df209e8e44fc0b99612ef41af4faccd21eb8)
```

Mind running the tests with: `ASAN_OPTIONS=fast_unwind_on_malloc=0`? That'll give you the full stack. Let's make sure it's indeed a bug in LLVM before using the bigger hammer of disabling ASAN.

viktormalik:
Rebased on top of master.

viktormalik:
> Thanks for checking. In that case, how about doing a surgical suppression for LLVM<=16 of the leak instead? Or just disable LSAN? ASAN has a bunch of useful memory safety checks as well. Would be a shame to lose that just cuz of an LLVM bug.
> 
> Example of suppression here: [danobi@fd9ee02](https://github.com/danobi/bpftrace/commit/fd9ee029e02e719ccf996bd6354f4707e2c074a2)

Ah, I was wondering if that's possible, thanks.

I took the liberty and used (and expanded) your commit. Since it is now specific to the particular memleak, I added it directly to #3772 so this one can be closed.

PR #3784, opened 2025-02-12
Remove '-kk' flag and surface BPF errors

jordalgo:
We should not be hiding these errors
from users by default. They inform if
something has gone wrong and will actually
affect program behavior. If users don't
like them they can use the `--no-warnings`
flag.

This also changes the '-k' flag to surface
errors from probe read helpers instead of
'-kk'.

##### Checklist


jordalgo:
Note: I debated about just surfacing all errors include bpf probe read ones but thought that might be too extreme - though I'm not against it if others feel differently; it certainly seems that errors from probe reads and `map_lookup_elem` can also have a bad affect on program behavior.

mmisono:
Making error reporting default seems reasonable to me as it helped me a lot troubleshoot several issues I had. Thins to notes are 

- Error checking increases code size and computation cost. In some critical cases, the overhead might be negligible. If we have `--no-warnings`, then I think it's okay.
- bpftrace stores zero when read-related helper function fails. And it's common pattern to rely on it, e.g., `/ @[tid] / { ... }`. IIRC, I found reporting read-related errors was a bit too much, and many interesting errors were non-read-related helpers errors when implementing `-k`, and that's the reason I added `-kk`. I haven't been involved in bpftrace recently, so I don't have a strong opinion here, but I feel including read-related errors by default is a bit too extreme.

(The original PR adding the error checking: #1276)

ajor:
> Yeah this is the one thing I'm worried about but I still think it might be a good tradeoff to swallowing important errors. We could also make an additional config for `no_warnings` so it can be embedded in a script to reduce this code size.

Error reporting by default with an option to disable seems like a good tradeoff to me. For most people, getting correct results is more important than performance.

> > reporting read-related helper function errors are sometimes really useful to debug elusive errors, e.g., detecting non-present page access
> 
> As you can tell I'm also on the fence. However for the map read case, maybe users should switch over to using `has_key` anyway, which wouldn't yield an error.

I don't think we should error by default when reading from a map with a non-existant key. Other reads might be fine to error on, if the error cases are separable.
I'm strongly against forcing people off of the old coding patterns - it'd render examples/tutorials incorrect (on lots of blogs we don't control), break most existing scripts and make users have to relearn the syntax for one-liners they might be typing out frequently.

jordalgo:
> I'm strongly against forcing people off of the old coding patterns  - it'd render examples/tutorials incorrect

Makes sense though I think this is an interesting topic to discuss (perhaps in a separate issue). For this, in particular, even if we bubbled up the error for map reads, it's still just a warning in the output so nothing "breaks" right? (aside from folks who output JSON and don't have any handling for warning messages). Either way, I think I'm fine to leave the changes as is and start by surfacing the errors we feel strongly about not suppressing - we can follow up with other kinds of reads in a later PR.

danobi:
Makes sense to me to do a gradual surfacing, starting with the obvious error ones (like -EFAULT with reading paged out memory). There's a tradeoff with generated overhead.

jordalgo:
Seems like folks are overall in favor of this PR as is (unless there are other errors that should be surfaced by default):
https://github.com/bpftrace/bpftrace/blob/master/src/ast/irbuilderbpf.cpp#L2669

jordalgo:
Very recent example of how surfacing probe read errors would have helped: 
- https://github.com/bpftrace/bpftrace/issues/3802

danobi:
Code changes lgtm.

But for codegen tests, what do you think about disabling helper error codegen by default? Long term I think we want to make the codegen tests be more single purpose (ie check for a single specific thing) and generally higher signal. So injecting a cross cutting feature like helper error checks moves away from that goal.

We already have `tests/codegen/runtime_error_check.cpp` which does some basic checks. How about augmenting that more if necessary instead of adding it to all codegen tests?

jordalgo:
> But for codegen tests, what do you think about disabling helper error codegen by default? 

Great suggestion. I'll do that and also add a few more tests runtime_error_check.cpp.

jordalgo:
@danobi Updated to remove all the changes from the codegen tests. Also added a bunch more codegen tests explicitly for the error helper.

danobi (approved):

PR #3788, opened 2025-02-12
Fix flaky tracepoint_multiattach_missing test

jordalgo:
This is because it was using printf, which
doesn't automatically insert a line break.
So sometimes the output would look like:
"hithit".

Also changed the other instances of printf("hit")
even though they are error cases.

This is the error in CI:
```
[  FAILED  ] probe.tracepoint_multiattach_missing (405 ms)
	Command: /home/runner/work/bpftrace/bpftrace/build-ci/src/bpftrace  -e 't:syscalls:sys_exit_nanosleep,t:syscalls:nonsense { printf("hit"); exit(); }'
	Expected: hit
	Found:
stdin:1:31-50: WARNING: tracepoint not found: syscalls:nonsense\nt:syscalls:sys_exit_nanosleep,t:syscalls:nonsense { printf("hit"); exit(); }\n                              ~~~~~~~~~~~~~~~~~~~\nAttaching 2 probes...\nopen(/sys/kernel/debug/tracing/events/syscalls/nonsense/id): No such file or directory\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\nhithit\n\n
```

##### Checklist

### DIFF
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -371,12 +371,12 @@ RUN {{BPFTRACE}} -e 'tracepoint:syscalls:sys_*_nanosleep { printf("hit\n"); }' -
 EXPECT hit
 
 NAME tracepoint_missing
-PROG t:syscalls:nonsense { printf("hit"); exit(); }
+PROG t:syscalls:nonsense { print("hit"); exit(); }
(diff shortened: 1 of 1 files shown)


danobi (approved):
good catch!

PR #3797, opened 2025-02-14
bpftrace: Store Config as a unique_ptr

danobi:
Previously, we were storing config by value and then giving references out to Usyms and Ksyms. If bpftrace instance were to ever be copied or moved, it would be UB, as those references are pointing to invalid memory.

Switch to storing Config as a unique_ptr so that its location is stable.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -563,7 +563,7 @@ CallInst *IRBuilderBPF::CreateGetStackScratchMap(StackType stack_type,
 Value *IRBuilderBPF::CreateGetStrAllocation(const std::string &name,
                                             const location &loc)
 {
-  const auto max_strlen = bpftrace_.config_.get(ConfigKeyInt::max_strlen);
+  const auto max_strlen = bpftrace_.config_->get(ConfigKeyInt::max_strlen);
   const auto str_type = CreateArray(max_strlen, CreateInt8());
   return createAllocation(bpftrace::globalvars::GlobalVar::GET_STR_BUFFER,
                           GetType(str_type),
@@ -661,7 +661,7 @@ Value *IRBuilderBPF::createAllocation(
     std::optional<std::function<size_t(AsyncIds &)>> gen_async_id_cb)
 {
   const auto obj_size = module_.getDataLayout().getTypeAllocSize(obj_type);
-  const auto on_stack_limit = bpftrace_.config_.get(
+  const auto on_stack_limit = bpftrace_.config_->get(
       ConfigKeyInt::on_stack_limit);
   if (obj_size > on_stack_limit) {
     return CreatePointerCast(
@@ -683,7 +683,7 @@ Value *IRBuilderBPF::createScratchBuffer(
   // ValueType var[MAX_CPU_ID + 1][num_elements]
   auto type = globalvars::get_type(globalvar,
                                    bpftrace_.resources,
-                                   bpftrace_.config_);
+                                   *bpftrace_.config_);
 
   // Get CPU ID
   auto cpu_id = CreateGetCpuId(loc);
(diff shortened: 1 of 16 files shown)


viktormalik (approved):

jordalgo:
Not disputing this is a good change but I *think* because bpftrace (the class) doesn't define move semantics or a copy constructor, you actually can't copy or move it because it already contains unique_ptrs as member variables. But yeah, in the future if we define those than the previous behavior would be a problem.

amscanne:
> Not disputing this is a good change but I think because bpftrace (the class) doesn't define move semantics or a copy constructor, you actually can't copy or move it because it already contains unique_ptrs as member variables. But yeah, in the future if we define those than the previous behavior would be a problem.

`unique_ptr` is movable, so it won't stop `BPFtrace` from being movable by default. Maybe there's something else stopping it from being movable, but you get a move and copy constructor by default unless there's something preventing it.

jordalgo:
@amscanne If you define a destructor for your class, you also need to provide a copy constructor in this case or you get an error.

amscanne:
> @amscanne If you define a destructor for your class, you also need to provide a copy constructor in this case or you get an error.

Maybe I'm misinterpreting what you're saying, but this is not true. You're only going to get an error if the copy constructor is being invoked -- which good since is it not copyable!. In fact, unique_ptr itself is an example of a move-only type.

jordalgo:
@amscanne - quoting google's AI here :) 

"A C++ class containing a std::unique_ptr member and a user-declared destructor can still be moved, but the move operation will not be implicitly defined by the compiler. The presence of a user-declared destructor prevents the automatic generation of a move constructor and move assignment operator. "

Here is a simple repro to illustrate this:

Works (no destructor):
```
class Dog {
public:
    Dog(std::unique_ptr<std::string> dog_name) : name(std::move(dog_name)) {}
    std::unique_ptr<std::string> name;
};

int main() {
    auto name = std::make_unique<std::string>("River");
    auto river = Dog(std::move(name));
    auto heidi = std::move(river);
    return 0;
}
```

Error (yes, destructor)
```
class Dog {
public:
    Dog(std::unique_ptr<std::string> dog_name) : name(std::move(dog_name)) {}
    ~Dog() {}
    std::unique_ptr<std::string> name;
};

int main() {
    auto name = std::make_unique<std::string>("River");
    auto river = Dog(std::move(name));
    auto heidi = std::move(river);
    return 0;
}
```
```
/Users/jordalgo/repos/cppstuff/hello.cpp:15:10: error: call to implicitly-deleted copy constructor of 'typename remove_reference<Dog &>::type' (aka 'Dog')
   15 |     auto heidi = std::move(river);
      |          ^       ~~~~~~~~~~~~~~~~
/Users/jordalgo/repos/cppstuff/hello.cpp:7:34: note: copy constructor of 'Dog' is implicitly deleted because field 'name' has a deleted copy constructor
    7 |     std::unique_ptr<std::string> name;
      |                                  ^
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/__memory/unique_ptr.h:213:3: note: copy constructor is implicitly deleted because 'unique_ptr<std::string>' has a user-declared move constructor
  213 |   unique_ptr(unique_ptr&& __u) _NOEXCEPT
      |   ^
1 error generated.
```

amscanne:
Right, that makes sense. I didn't know that a destructor alone stopped that. Apparently [so does](https://en.cppreference.com/w/cpp/language/move_constructor) having a user-defined copy constructor, copy assignment operator or move assignment operator. But it all makes sense and your original comment is correct. Who doesn't love C++ :)

danobi:
My thinking was that it's probably ok today but maybe in the future we break up the mega BPFtrace class and something slips through. Seems like that's a real possibility given the weird rules around this.

PR #3799, opened 2025-02-15
ast: make AST nodes consistent

amscanne:
Currently *most* AST nodes have a location, some have a location optionally passed, and some have none. To unify diagnostics and reporting, make all the AST nodes consistent and leverage the type system: all nodes must have a location. This fixes several gaps in the parser.

If a location is not available, then a location can be constructed somehow from the original source locations. It is simply not helpful to produce errors that tie in no way to the source file.

Additionally, make AST nodes non-copyable. This was done in part for *only* the AttachPoint, which is expanded. In the future, this should either be done 1) as a true expansion of the AST, so that subsequent passes can operate on the newly expanded probes as expanded or 2) not a property of the AST itself, and existing in a side-structure. I suspect that a mix a both will be required, but right now it lives in a strange ambiguity within the AST.

This change is a non-functional change in preparation for additional internal improvements.

##### Checklist

- ~[X] The new behaviour is covered by tests~

### DIFF
--- a/tests/parser.cpp
+++ b/tests/parser.cpp
@@ -284,9 +284,9 @@ TEST(Parser, positional_param_attachpoint)
 stdin:1:1-12: ERROR: Found trailing text 'a' in positional parameter index. Try quoting the trailing text.
 uprobe:$1a { 1 }
 ~~~~~~~~~~~
-stdin:1:1-1: ERROR: No attach points for probe
+stdin:1:1-18: ERROR: No attach points for probe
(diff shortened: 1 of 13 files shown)


tests/parser.cpp
danobi, line 1: sweet!

danobi (approved):
minor nit; lgtm otherwise

jordalgo (commented):
Looks great!

PR #3800, opened 2025-02-15
Add pid targeting/filtering for all probes

jordalgo:
Currently the `-p` (pid) CLI option
only applies to uprobes/uretprobes/usdts.
However this causes confusion as you can
get events from all pids for any other
kind of probe[1].

This change adds pid filtering at the attach
point when possible or, for probes that
do not support a pid filter when attaching,
modifies the AST by adding this to the start:
```
if (pid != TARGETED_PID) {
  return;
}
```

[1] https://github.com/bpftrace/bpftrace/issues/254

##### Checklist

### DIFF
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -20,6 +20,7 @@ add_library(ast STATIC
   passes/semantic_analyser.cpp
   passes/codegen_llvm.cpp
   passes/return_path_analyser.cpp
+  passes/pid_filter_pass.cpp
 )
 
 target_compile_definitions(ast PRIVATE ${BPFTRACE_FLAGS})
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -110,9 +110,9 @@ int AttachedProbe::detach_fentry()
   return 0;
 }
 
-void AttachedProbe::attach_iter(int pid)
+void AttachedProbe::attach_iter(std::optional<int> pid)
 {
-  if (pid == 0) {
+  if (!pid.has_value()) {
     linkfd_ = bpf_link_create(progfd_,
                               0,
                               static_cast<enum ::bpf_attach_type>(
@@ -122,7 +122,7 @@ void AttachedProbe::attach_iter(int pid)
     BPFTRACE_LIBBPF_OPTS(bpf_link_create_opts, opts);
     union bpf_iter_link_info linfo;
     memset(&linfo, 0, sizeof(linfo));
-    linfo.task.pid = pid;
+    linfo.task.pid = *pid;
     opts.iter_info = &linfo;
     opts.iter_info_len = sizeof(linfo);
     linkfd_ = bpf_link_create(progfd_,
@@ -166,7 +166,7 @@ int AttachedProbe::detach_raw_tracepoint()
 
 AttachedProbe::AttachedProbe(Probe &probe,
                              const BpfProgram &prog,
-                             int pid,
+                             std::optional<int> pid,
                              BPFtrace &bpftrace,
                              bool safe_mode)
     : probe_(probe), progfd_(prog.fd()), bpftrace_(bpftrace)
@@ -759,7 +759,7 @@ static void resolve_offset_uprobe_multi(const std::string &path,
   }
 }
 
-void AttachedProbe::attach_multi_uprobe(int pid)
+void AttachedProbe::attach_multi_uprobe(std::optional<int> pid)
 {
   std::vector<std::string> syms;
   std::vector<unsigned long> offsets;
@@ -778,8 +778,8 @@ void AttachedProbe::attach_multi_uprobe(int pid)
   opts.uprobe_multi.flags = probe_.type == ProbeType::uretprobe
                                 ? BPF_F_UPROBE_MULTI_RETURN
                                 : 0;
-  if (pid != 0) {
-    opts.uprobe_multi.pid = pid;
+  if (pid.has_value()) {
+    opts.uprobe_multi.pid = *pid;
   }
 
   if (bt_verbose) {
@@ -804,7 +804,7 @@ void AttachedProbe::attach_multi_uprobe(int)
 }
 #endif // HAVE_LIBBPF_UPROBE_MULTI
 
-void AttachedProbe::attach_uprobe(int pid, bool safe_mode)
+void AttachedProbe::attach_uprobe(std::optional<int> pid, bool safe_mode)
 {
   if (!probe_.funcs.empty()) {
     attach_multi_uprobe(pid);
@@ -819,7 +819,7 @@ void AttachedProbe::attach_uprobe(int pid, bool safe_mode)
                                         eventname().c_str(),
                                         probe_.path.c_str(),
                                         offset_,
-                                        pid == 0 ? -1 : pid,
+                                        pid.has_value() ? *pid : -1,
                                         0);
 
   if (perf_event_fd < 0) {
@@ -913,7 +913,7 @@ int AttachedProbe::usdt_sem_up([[maybe_unused]] BPFfeature &feature,
   return usdt_sem_up_manual_addsem(pid, fn_name, ctx);
 }
 
-void AttachedProbe::attach_usdt(int pid, BPFfeature &feature)
+void AttachedProbe::attach_usdt(std::optional<int> pid, BPFfeature &feature)
 {
   struct bcc_usdt_location loc = {};
   int err;
@@ -922,13 +922,13 @@ void AttachedProbe::attach_usdt(int pid, BPFfeature &feature)
   // probe:
   std::string fn_name = "probe_" + probe_.attach_point + "_1";
 
-  if (pid) {
+  if (pid.has_value()) {
     // FIXME when iovisor/bcc#2064 is merged, optionally pass probe_.path
-    ctx = bcc_usdt_new_frompid(pid, nullptr);
+    ctx = bcc_usdt_new_frompid(*pid, nullptr);
     if (!ctx)
       throw FatalUserException(
           "Error initializing context for probe: " + probe_.name +
-          ", for PID: " + std::to_string(pid));
+          ", for PID: " + std::to_string(*pid));
   } else {
     ctx = bcc_usdt_new_frompath(probe_.path.c_str());
     if (!ctx)
@@ -964,7 +964,7 @@ void AttachedProbe::attach_usdt(int pid, BPFfeature &feature)
   //
   // NB: Do *not* use `ctx` after this call. It may either be open or closed,
   // depending on which path was taken.
-  err = usdt_sem_up(feature, pid, fn_name, ctx);
+  err = usdt_sem_up(feature, pid.value_or(0), fn_name, ctx);
 
   if (err) {
     throw FatalUserException(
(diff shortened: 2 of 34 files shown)


jordalgo:
Before I add tests, I wanted a check on the implementation. I don't think semantic_analyser is a great place for this modification but I didn't want to create a whole new pass. Also, I still have to investigate if we can pass a pid for other attachments (I think I can for perf events).

src/attached_probe.cpp
```diff
@@ -1029,7 +1032,7 @@ void AttachedProbe::attach_profile()
                                               PERF_COUNT_SW_CPU_CLOCK,
                                               period,
                                               freq,
-                                              pid,
+                                              pid == 0 ? -1 : pid,
```
viktormalik, line 1035: Wouldn't it make more sense to use -1 as the "no PID" value everywhere? Technically speaking, 0 is a valid PID.

viktormalik (commented):
Code looks mostly good to me.

> I don't think semantic_analyser is a great place for this modification but I didn't want to create a whole new pass.

Agreed, we should definitely have this outside of the semantic analyser. As I already wrote in https://github.com/bpftrace/bpftrace/pull/3772#discussion_r1957810821, it might make sense to first refactor the existing code to the new mutating infra so that we have a better idea whether it makes sense to create a new pass for this or merge it to some existing one.

src/attached_probe.cpp
jordalgo: Soooo.... I started to do this but found out there is some weird behavior when using pid 0 at the attach point here (namely it doesn't seem to work - I've been doing some light poking around in the kernel to see what's going on). 
Basically, I agree with you but it would be mean utilizing the added AST filter approach for all these probes: hardware, software, interval, etc. instead of passing the pid at the attachpoint.

viktormalik: Hm, there may be some misunderstanding here. I meant to change the semantics of `BPFtrace::pid()` to always return -1 when no PID is specified. Then, you can use it directly here (if 0 doesn't work as you claim) and you'll just need to update places which check for 0 as the "no PID" special case to check for -1.

jordalgo: @viktormalik I understand but maybe I did a bad job of explaining the issue I ran into. So I started doing exactly as you described but when I set the pid to 0 on the command line e.g. `bpftrace -e 'hardware:cpu-cycles:10000000 { print(("hi", pid)); }' -p 0`. I was NOT getting events from only pid 0 (in fact, I wasn't getting any). However, if I did the filtering (via the added AST I added) instead of at the attach point for the hardware probe (which uses `bpf_attach_perf_event`) then it worked as expected.

viktormalik: Ok, I'm still probably missing something. What did you expect to see from using `-p 0`? PID 0 is the swapper so it's not supposed to run any events so getting nothing is IMO actually correct.

The way I understand the semantics of `bpf_attach_perf_event` is:
- `pid == -1` -> trace all events system-wide
- `pid != -1` -> trace only events from the given PID

viktormalik: Ah, I see, my mistake.

Still, the current solution is not correct, is it? When someone wants to filter events from PID 0 (i.e. uses `-p 0`), we will pass -1 here so all events will be captured, won't they? So users still need to apply manual filtering as you did in the above case.

But it's probably still better than my suggestion as we'd still need these ternary operators to replace 0 by -1 to at least get some events.

jordalgo: > Still, the current solution is not correct, is it?

Yes, I'm still trying to figure out the best solution. Maybe we just don't pass the pid at the attach point for these other probes (until i have more time to investigate) and utilize the injected filter (as I'm doing for kfuncs/fentry) as that seems to yield more correct results. 

I still agree about allowing pid `-p` to be 0, which requires some additional changes.

jordalgo: @viktormalik I'm going to look into allowing pid 0 in a follow-up PR as I need to think on it a bit more. After some discussion the use case for filtering on the swapper thread seems a little thin.

jordalgo:
Went with a separate analyser pass instead of adding this to semantic analyser as per the recommendation from @amscanne and @danobi

src/ast/passes/pid_filter_analyser.h
```diff
@@ -0,0 +1,39 @@
...
+#include "bpffeature.h"
+#include "bpftrace.h"
+#include "collect_nodes.h"
+#include "config.h"
+#include "types.h"
+
+namespace bpftrace {
+namespace ast {
+
+class PidFilterAnalyser : public Visitor<PidFilterAnalyser> {
```
amscanne, line 18: Minor nits, mostly to solicit feedback on the nits themselves rather than have you address them (since it's just as easy to address in one swoop).

I'm trying to clear up some of the pass plumbing to make this a bit simpler, and for visibility here are the things that I'm intending to do:
* With the exception of helpers like "CollectNodes", there is no need for the Visitors to be defined in the headers. They can defined inline in the `.cpp` files, and the only exposed function through the headers needs to be `CreatePidFilterPass` below. You basically have that and `#include <ast/pass_manager.h>` above.
* I'm not sure what a "Analyser" is broadly, but there's a lot of cargo culting with the `analyse` method. It is not necessary. Just have a simple visitor defined directly and roll your pass logic into the function you return from `CreatePidFilterPass`.
* I have a change that I'll be sending later today after the Diagnostics, which removes the `PassResult` completely, since it's unnecessary. With that and the above, I think the boilerplate woudl be cut down considerably.

src/attached_probe.cpp
amscanne: For the internal APIs, I would recommend just using `std::optional<int>` for the arguments and plumbing. This avoids all the ambiguity over what the special sentinel value is. It seems reasonably fine to apply some constraints on this pid value at the top-level, but this shouldn't bring special consideration to the internal plumbing, because `std::optional` is using the type system to do the lifting :).

tests/pid_filter_analyser.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  std::ostringstream out;
+  ast::Printer printer(driver.ctx, out);
+  printer.print();
+
+  if (has_filter) {
+    EXPECT_THAT(out.str(), HasSubstr(expected_ast));
+  } else {
+    EXPECT_THAT(out.str(), Not(HasSubstr(expected_ast)));
+  }
```
amscanne, line 53: Thinking out loud: I don't want to boil the ocean, but it would be great to move away from "stringly typed" tests in general. We are doing this across the board for errors and other things, but since this is some new ground I will give a suggestion.

You could write a [custom matcher](https://google.github.io/googletest/reference/matchers.html) for the pid filter check in the AST.

```
MATCHER_P(HasPidFilter, pid, "") {
  if (arg->block->stmts.size() == 0)
    return false;
  auto if = dynamic_cast<ast::If*>(arg->block->stmts[0]);
  if (if == nullptr)
     return false;
  ... // check if condition, and that statement is return
}

for (const auto &probe : driver.ctx.root->probes) {
   EXPECT_THAT(probe, HasPidFilter(pid));
}
```

Alternately, you could just have an inline visitor that verifies this. Neither is a huge life, and I think it makes the general test more sustainable.

tests/pid_filter_analyser.cpp
```diff
@@ -0,0 +1,94 @@
...
+    "usdt:sh:probe",
+    "watchpoint:0x0:8:rw",
+    "asyncwatchpoint:func1+arg2:8:rw",
+    "profile:ms:1",
+    "interval:s:1",
+    "software:faults:1000",
+    "hardware:cache-references:1000000",
+  };
+
+  for (auto& probe : no_filter_probes) {
```
amscanne, line 89: Nit: I think it's also murky with table-drive tests, but these seem like two separate tests at least.

jordalgo: Good call to separate the tests but I think the table-driven tests is actually better/more-readable in this case.

amscanne: I'm not objecting to the table-driven, just the "matches" and "no matches" two logical tests in a single test. I'm advocating for making this two cases.

(As an aside, you can do table-driven natively with gtest also, but that's probably overkill)

src/ast/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ add_library(ast STATIC
   passes/semantic_analyser.cpp
   passes/codegen_llvm.cpp
   passes/return_path_analyser.cpp
+  passes/pid_filter_analyser.cpp
```
viktormalik, line 23: Naming flame alert :warning: 

I don't think `_analyser` is a good suffix in this context as the pass doesn't just analyze but also mutates the AST.

LLVM distinguishes [Analysis and Transform Passes](https://llvm.org/docs/Passes.html) so I think we should do the same as people may be already familiar with them. So, I'd go with `passes/pid_filter_transformer.cpp`.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2878,7 +2878,8 @@ void CodegenLLVM::add_probe(AttachPoint &ap,
       return;
   }
   if (probetype(ap.provider) == ProbeType::usdt) {
-    auto usdt = usdt_helper_->find(bpftrace_.pid(), ap.target, ap.ns, ap.func);
+    auto usdt = usdt_helper_->find(
+        bpftrace_.pid().value_or(0), ap.target, ap.ns, ap.func);
```
viktormalik, line 2882: Why not use `std::optional<pid_t>` in `USDTHelper`, too, to keep consistency?

src/bpftrace.cpp
```diff
@@ -1493,8 +1494,8 @@ std::optional<std::string> BPFtrace::get_watchpoint_binary_path() const
     auto args = split_string(cmd_, ' ', /* remove_empty */ true);
     assert(!args.empty());
     return resolve_binary_path(args[0]).front();
-  } else if (pid())
-    return "/proc/" + std::to_string(pid()) + "/exe";
+  } else if (pid().has_value())
+    return "/proc/" + std::to_string(pid().value_or(0)) + "/exe";
```
viktormalik, line 1498: No need for the `.value_or(0)` part IMO, since `pid()` must already have value.
```suggestion
    return "/proc/" + std::to_string(*pid()) + "/exe";
```

viktormalik (commented):
Couple more comments, otherwise looks good.

src/bpftrace.cpp
viktormalik: Makes sense.

tests/pid_filter_analyser.cpp
jordalgo: Let me look into doing this in a follow-up PR. Not opposed to the suggestion. I just have to time box how much I spend on this "minor" fix 😄

src/ast/CMakeLists.txt
jordalgo: Renamed it to PidFilterPass. WDYT?

jordalgo:
Ugh, seems I'm having an issue with the codegen test I added. My local machine is producing slightly different IR. I'll try to debug a bit but I might just end up removing it.

amscanne:
> Ugh, seems I'm having an issue with the codegen test I added. My local machine is producing slightly different IR. I'll try to debug a bit but I might just end up removing it.

That's curious, the IR should be invariant with respect to the specific environment. (Perhaps just the printing is subject to some minor version specifics.) Just make sure that you've rebased to avoid the PR catching an additional change? I wouldn't roll over this without investigating a bit.

jordalgo:
> That's curious, the IR should be invariant with respect to the specific environment. (Perhaps just the printing is subject to some minor version specifics.) Just make sure that you've rebased to avoid the PR catching an additional change? I wouldn't roll over this without investigating a bit.

I was able to fix it by rebasing off my change to always mock BPFfeature. Still not exactly sure what was causing the issue by disabling all the features.

viktormalik (approved):

PR #3805, opened 2025-02-18
bpffeature: Fix detection for tracing programs

kkdwivedi:
Eric Hagberg from Jane Street reports:

After the upstream fix [0] for tracing programs rejecting narrow loads for pointer arguments in context, bpftrace began failing feature detection for dpath and skboutput.

For instance, bpftrace --info would say no for both cases, and attempting to attach a program like so:

  bpftrace -e 'kfunc:security_file_open { printf("%s", path(args->file->f_path)); }'

would yield an error:

  stdin:1:41-64: ERROR: BPF_FUNC_d_path not available for your kernel

Fix this by doing the right thing and performing 64-bit loads into the context to form a pointer in R1.

  [0]: https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf.git/commit/?id=659b9ba7cb2d

##### Checklist

### DIFF
--- a/src/bpffeature.cpp
+++ b/src/bpffeature.cpp
@@ -365,7 +365,7 @@ bool BPFfeature::has_d_path()
     return *has_d_path_;
 
   struct bpf_insn insns[] = {
-    BPF_LDX_MEM(BPF_W, BPF_REG_1, BPF_REG_1, 0),
+    BPF_LDX_MEM(BPF_DW, BPF_REG_1, BPF_REG_1, 0),
     BPF_MOV64_REG(BPF_REG_2, BPF_REG_10),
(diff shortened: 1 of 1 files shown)


viktormalik (commented):
Thanks @kkdwivedi!

Am I correct that this is backwards-compatible, i.e. the feature detection will work on kernels without [0]?

kkdwivedi:
Yes, the verifier check was too relaxed, so both u64 and u32 (and lower) access sizes passed on older kernels, now it's simply u64 going forward/

viktormalik (approved):

jordalgo:
+1 thanks for the fix! 

@viktormalik Think this is worth a patch for 0.22?

jordalgo (approved):

danobi:
Changelog + backport seems good

PR #3809, opened 2025-02-20
Re-enable bpf_map_sum_elem_count() on LLVM <17

viktormalik:
#3683 disabled usage of the `bpf_map_sum_elem_count` kfunc on LLVM <17 as it was failing LLVM IR code verification. There were two issues reported:

1. ```
    Call parameter type does not match function signature!
    %"struct map_t"* @AT_
     i8*  %len = call i64 @bpf_map_sum_elem_count(%"struct map_t"* @AT_)
    ```
    This one can be solved by simply adding a pointer cast.

2. ```
    Expected no forward declarations!
    !75 = <temporary!> !{}
    ```
    The reason for this failure is a bug in LLVM <17 in `DIBuilder::createFunction` when called for a function declaration (which we do for kfuncs). It creates an empty temporary `MDTuple` for `RetainedNodes` inside `DISubprogram` which is, in addition, never freed and therefore introduces a memory leak. The bug was fixed by https://github.com/llvm/llvm-project/commit/ed506dd6cecd9653cf9202bfe195891a33482852.

    This can be solved by deleting the temporary `MDTuple`.

Unfortunately, resolving the above two problems was not sufficient and LLVM IR verification kept complaining:
```
inlinable function call in a function with debug info must have a !dbg location
  %len = call i64 @bpf_map_sum_elem_count(i8* bitcast (%"struct map_t"* @AT_ to i8*))
```

This one turned out to be tricky to resolve. I tried to add a debug location to the call but then I was hitting a segfault in LLVM which I was not able to get rid of. So, in the end, I decided to disable LLVM IR verification for runtime tests using kfuncs on affected LLVM versions (only 14 and 15 as this check is relaxed in LLVM 16 by https://github.com/llvm/llvm-project/commit/041ec822421aae33f064d96a21607e550766dd4d). 

This is not an ideal solution as it will require us to explicitly disable IR verification for each test using kfunc which we add but I don't have anything better at the moment. Good thing is that we will stop supporting LLVM 15 once LLVM 21 is out (per our policy to support the last 6 versions) which will happen in September this year.

Note that the above issues would happen to any kfunc call so keeping the feature disabled is not a viable solution as we will eventually want to use more kfuncs.

##### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,12 +35,6 @@ jobs:
     strategy:
       matrix:
         env:
-        - NAME: LLVM 14
-          CMAKE_BUILD_TYPE: Debug
-          NIX_TARGET: .#bpftrace-llvm14
-        - NAME: LLVM 15
-          CMAKE_BUILD_TYPE: Debug
-          NIX_TARGET: .#bpftrace-llvm15
         - NAME: LLVM 16
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm16
--- a/flake.nix
+++ b/flake.nix
@@ -228,8 +228,6 @@
             bpftrace-llvm18 = mkBpftrace 18;
             bpftrace-llvm17 = mkBpftrace 17;
             bpftrace-llvm16 = mkBpftrace 16;
-            bpftrace-llvm15 = mkBpftrace 15;
-            bpftrace-llvm14 = mkBpftrace 14;
 
             # Self-contained static binary with all dependencies
             appimage = nix-appimage.mkappimage.${system} {
@@ -286,8 +284,6 @@
             bpftrace-llvm18 = mkBpftraceDevShell self.packages.${system}.bpftrace-llvm18;
             bpftrace-llvm17 = mkBpftraceDevShell self.packages.${system}.bpftrace-llvm17;
             bpftrace-llvm16 = mkBpftraceDevShell self.packages.${system}.bpftrace-llvm16;
-            bpftrace-llvm15 = mkBpftraceDevShell self.packages.${system}.bpftrace-llvm15;
-            bpftrace-llvm14 = mkBpftraceDevShell self.packages.${system}.bpftrace-llvm14;
           };
         });
 }
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -47,6 +47,29 @@ void DIBuilderBPF::createFunctionDebugInfo(llvm::Function &func,
                                          0,
                                          DINode::FlagPrototyped,
                                          flags);
+#if LLVM_VERSION_MAJOR < 17
+  // There's a bug in LLVM <17 in DIBuilder::createFunction when called for a
+  // function declaration. It creates an empty temporary MDTuple for
+  // RetainedNodes inside DISubprogram which is, in addition, never freed.
+  //
+  // We generate function declaration debug info for kfuncs so this causes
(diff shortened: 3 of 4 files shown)


jordalgo (approved):
Nice investigating! This looks good to me. I agree that disabling LLVM IR validation is not great but since it's contained to a few run time tests for older LLVM versions, I think it's a good tradeoff for unblocking kfuncs and session probes as well as fixing a mem leak.

tests/runtime/basic
```diff
@@ -285,11 +285,13 @@ NAME variable strings are memset
 PROG BEGIN { $x = "xxxxx"; $x = "a"; @[$x] = 1; $y = "yyyyy"; $y = "a"; @[$y] = 1; printf("len: %d\n", len(@)); exit(); }
 EXPECT len: 1
 TIMEOUT 1
+SKIP_VERIFY_LLVM_IR 14,15
```
jordalgo, line 288: Nit: maybe it would be good to add a comment here with either an issue or some details as to why this is needed (re: the segfault and dbg info)

danobi:
> Good thing is that we will stop supporting LLVM 15 once LLVM 21 is out (per our policy to support the last 6 versions) which will happen in September this year.

I don't think we've made a firm commitment to support 6 versions:
https://github.com/bpftrace/bpftrace/blob/master/docs/dependency_support.md#llvm-dynamically-linked

In contrast to kernel support, I suspect in practice it's unlikely that a new bpftrace is paired with a 3 years old LLVM. So I feel like it'd be reasonable to just drop LLVM 14 and 15 if they're giving us trouble. 16 is more borderline but I feel like even at ~2 years old it's fine to drop.

jordalgo:
> So I feel like it'd be reasonable to just drop LLVM 14 and 15 if they're giving us trouble. 16 is more borderline but I feel like even at ~2 years old it's fine to drop.

I agree that being said, I don't want to keep blocking @viktormalik. I think it could be a follow-up task to remove support for those 3 versions but I'll leave it to you all to decide. I already approved 👍

viktormalik:
> > So I feel like it'd be reasonable to just drop LLVM 14 and 15 if they're giving us trouble. 16 is more borderline but I feel like even at ~2 years old it's fine to drop.
> 
> 
> 
> I agree that being said, I don't want to keep blocking @viktormalik. I think it could be a follow-up task to remove support for those 3 versions but I'll leave it to you all to decide. I already approved 👍 

As I wrote it somewhere, the proposed solution is far from being good as it will require to explicitly skip IR verification for each runtime test using kfuncs.

So, if other maintainers don't mind, I'm in for dropping LLVM 14 and 15 from CI. No need to drop 16 as it works fine with the fix in this PR.

tests/runtime/engine/parser.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, expect, mode):
         'will_fail',
         'new_pidns',
         'skip_if_env_has',
+        'skip_verify_llvm_ir',
```
danobi, line 49: Please document in tests/README.md

danobi (approved):
Ack. Let's follow up in separate PR then.

But please add docs on new runtime test directive

viktormalik:
> Ack. Let's follow up in separate PR then.

Ah, I meant we should drop LLVM 14 and 15 directly in this PR. Then we don't even need the new test directive.

danobi:
+1 on dropping!

viktormalik:
> +1 on dropping!

Done. I only dropped the CI jobs and Nix targets for now and will follow-up with a PR for dropping conditionally compiled code.

danobi (approved):

PR #3811, opened 2025-02-21
Overhaul strcontains() implementation

danobi:
strcontains() unrolls a nested loop. With default big strings, this easily
triggers an overflow of instructions.

This PR overhauls the implementation to not unroll nested loops. See
individual commits for more details.

##### Checklist


tests/codegen/array_integer_equal_comparison.cpp
```diff
@@ -16,28 +16,9 @@ constexpr auto PROG = "struct Foo { int arr[4]; }"
...
-
-  // Force unroll fallback
-  auto feature = std::make_unique<MockBPFfeature>(true);
-  feature->has_loop(false);
-  bpftrace->feature_ = std::move(feature);
-
-  test(*bpftrace, PROG, NAME);
-}
-
 TEST(codegen, array_integer_equal_comparison_no_unroll)
```
jordalgo, line 19: nit: can probably rename this test to just `array_integer_equal_comparison` as there is no longer an unroll option.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1404,29 +1404,31 @@ ScopedExpr CodegenLLVM::visit(Call &call)
   } else if (call.func == "strerror") {
     return visit(call.vargs.front());
   } else if (call.func == "strncmp") {
+    auto &left_arg = *call.vargs.at(0);
+    auto &right_arg = *call.vargs.at(1);
     auto size_opt = bpftrace_.get_int_literal(call.vargs.at(2));
```
jordalgo, line 1409: JC, is it possible and better to set the call expression size to be this min of the arg, left, and right in a earlier pass (semantic_analyser) ?

tests/codegen/llvm/literal_strncmp.ll
```diff
@@ -11,6 +11,7 @@ target triple = "bpf-pc-linux"
 @AT_ = dso_local global %"struct map_t" zeroinitializer, section ".maps", !dbg !0
```
jordalgo, line 11: Do we still need this codegen test at all now that string handling is all the same and string_equal_comparision.cpp also uses a literal?

tests/README.md
```diff
@@ -103,6 +103,10 @@ Each runtime testcase consists of multiple directives. In no particular order:
 * `MIN_KERNEL`: Skip the test unless the host's kernel version is >= the
   provided kernel version. Try not to use this directive as kernel versions may
   be misleading (backported kernel features, for example)
+* `MAX_KERNEL`: Skip the test unless the host's kernel version is <= the
```
jordalgo, line 106: Nit: shouldn't this addition be in the commit message as well?

Edit: Ah, you're just adding details about it. It already existed. Carry on.

tests/codegen/strcontains.cpp
```diff
@@ -0,0 +1,15 @@
+#include "../mocks.h"
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, strcontains)
+{
+  test(R"(k:foo { strcontains("hello-test-world", "test") })", NAME);
```
jordalgo, line 10: Worth adding a test with non literal string(s) ?

tests/runtime/probe
```diff
@@ -464,8 +464,11 @@ PROG hardware:cache-misses:10 { print(probe); exit();}
 EXPECT hardware:cache-misses:10
 
 NAME probe_builtin_string_compare
-RUN {{BPFTRACE}} -e 'uprobe:testprogs/uprobe_test:uprobeFunction* { if (strcontains(probe, "uprobeFunction2")) { print(probe); } }' -c ./testprogs/uprobe_test
+PROG uprobe:testprogs/uprobe_test:uprobeFunction* { if (strcontains(probe, "uprobeFunction2")) { print(probe); exit(); } }
+AFTER ./testprogs/uprobe_test
 EXPECT uprobe:testprogs/uprobe_test:uprobeFunction2
+# For some reason LLVM versions older than just generate code that does not pass BPF verifier
+MIN_LLVM 18
```
jordalgo, line 471: Language nit: "For some reason older LLVM versions generated code that does not pass the BPF verifier".

Larger question: isn't this a problem? You're saying strcontains will not work on llvm versions less than 18?

tests/runtime/strcontains
```diff
@@ -0,0 +1,6 @@
+NAME strcontains no literals
```
jordalgo: Does this need it's own file. Why not under the test in "tests/runtime/call" ?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to
   - [#3787](https://github.com/bpftrace/bpftrace/pull/3787)
 - Published aarch64 appimage builds from master
   - [#3795](https://github.com/bpftrace/bpftrace/pull/3795)
+- Added warning when strcontains() is used on too-big strings (which can cause verifier issues)
```
jordalgo, line 24: nit: "...used on strings that are too big and may cause verifier issues"

tests/runtime/probe
danobi: The new implementation for some reason doesn't verify if built with older llvms. I plan on asking about this tomorrow. It is a problem, but a somewhat small one IMHO. The previous impl had silent wrong results + frequent LLVM crashes. The new one is correct, but under certain builds doesn't load. I'd argue new impl is strictly better b/c at least it's correct and loud if wrong.

jordalgo: And we don't want to just maintain the old way for llvm versions < 18 ? When you say the previous impl had silent wrong results - do we know how often and for what cases? Maybe having something a little broken on these older LLVM version is better than nothing at all.

danobi: Included in this PR is a runtime test that had silent wrong results for the old impl. I couldn't debug it - the old impl was too confusing. I think there's nothing worse than silent wrong results. It is much better to not do it at all than to silently do it wrong.

tests/runtime/engine/runner.py
```diff
@@ -70,6 +70,7 @@ class Runner(object):
     SKIP_AOT_NOT_SUPPORTED = 7
     SKIP_KERNEL_VERSION_MAX = 8
     SKIP_NOT_IN_ALLOWLIST = 9
+    SKIP_LLVM_VERSION_MIN = 9
```
viktormalik, line 73: Should use a different number?
```suggestion
    SKIP_LLVM_VERSION_MIN = 10
```

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  //
+  //   return false;
+  // }
+  //
+  // clang-format on
 
   llvm::Function *parent = GetInsertBlock()->getParent();
-  AllocaInst *store = CreateAllocaBPF(getInt1Ty(), "strcontains.result");
+  BasicBlock *empty_check = BasicBlock::Create(module_.getContext(),
+                                               "strcontains.outer.cond",
```
viktormalik, line 1681:
```suggestion
                                               "strcontains.empty.check",
```

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
                                              parent);
   BasicBlock *done_false = BasicBlock::Create(module_.getContext(),
                                               "strcontains.false",
                                               parent);
+  BasicBlock *done = BasicBlock::Create(module_.getContext(),
+                                        "strcontains.done",
+                                        parent);
 
-  CreateStore(getInt1(!inverse), store);
+  AllocaInst *i = CreateAllocaBPF(getInt64Ty(), "strcontain.i");
```
viktormalik, line 1714: To keep consistent naming
```suggestion
  AllocaInst *i = CreateAllocaBPF(getInt64Ty(), "strcontains.i");
```

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  CreateStore(getInt64(0), i);
+  AllocaInst *j = CreateAllocaBPF(getInt64Ty(), "strcontain.j");
```
viktormalik, line 1716: Same as above
```suggestion
  AllocaInst *j = CreateAllocaBPF(getInt64Ty(), "strcontains.j");
```

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  Value *needle_first = CreateLoad(
+      getInt8Ty(), CreateGEP(getInt8Ty(), needle, { getInt64(0) }));
+  Value *needle_first_null = CreateICmpEQ(needle_first, null_byte);
+  CreateCondBr(needle_first_null, done_true, outer_cond);
+
+  SetInsertPoint(outer_cond);
+  Value *i_val = CreateLoad(getInt64Ty(), i);
+  Value *haystack_inbounds = CreateICmpULT(i_val, getInt64(haystack_sz));
+  Value *haystack_char = CreateLoad(
+      getInt8Ty(), CreateGEP(getInt8Ty(), haystack, { i_val }));
```
viktormalik, line 1732: Are we sure that LLVM optimizes this to short-circuit evaluation? Otherwise you may be doing an out-of-bounds read here.

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  Value *haystack_cmp_char = CreateLoad(
+      getInt8Ty(), CreateGEP(getInt8Ty(), haystack, { haystack_cmp_idx }));
+  Value *haystack_cmp_null = CreateICmpEQ(haystack_cmp_char, null_byte);
+  Value *haystack_cmp_needle = CreateICmpNE(haystack_cmp_char, needle_char);
+  Value *haystack_cmp_cond = CreateOr(
+      { haystack_cmp_outbounds, haystack_cmp_null, haystack_cmp_needle });
+  CreateCondBr(haystack_cmp_cond, outer_incr, inner_incr);
+
+  SetInsertPoint(inner_incr);
+  CreateStore(CreateAdd(CreateLoad(getInt64Ty(), j), getInt64(1)), j);
```
viktormalik, line 1766: Note: this could be replaced by a PHI node to save stack space. Not sure if it's worth the work, though. And maybe LLVM will do it for us.

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,146 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  CreateBr(inner_cond);
+
+  SetInsertPoint(outer_incr);
+  CreateStore(CreateAdd(CreateLoad(getInt64Ty(), i), getInt64(1)), i);
```
viktormalik, line 1770: Same as above.

tests/semantic_analyser.cpp
```diff
@@ -3253,6 +3253,27 @@ TEST(semantic_analyser, strconrtains)
   test("i:s:1 { strcontains(\"a\",1) }", 2);
 }
 
+TEST(semantic_analyser, strcontains_large_warnings)
+{
+  test_for_warning("kprobe:f { $s1 = "
+                   "str(arg0); $s2 = str(arg1); strcontains($s1, $s2) }",
```
viktormalik, line 3259: A bit hard to read, improve line splits?
```suggestion
  test_for_warning("kprobe:f {
                   "$s1 = str(arg0); $s2 = str(arg1);
                   "strcontains($s1, $s2) }",
```

tests/semantic_analyser.cpp
```diff
@@ -3253,6 +3253,27 @@ TEST(semantic_analyser, strconrtains)
...
+                   "both string sizes is larger");
+
+  test_for_warning("kprobe:f { $s1 = str(arg0, 64); "
+                   "$s2 = str(arg1, 16); strcontains($s1, $s2) }",
```
viktormalik, line 3263: Same as above
```suggestion
  test_for_warning("kprobe:f {
                   "$s1 = str(arg0, 64); $s2 = str(arg1, 16);
                   "strcontains($s1, $s2) }",
```

viktormalik (commented):
A couple of comments from my side.

src/ast/passes/codegen_llvm.cpp
danobi: The call to strncmp() returns an int64, so probably can't do it like that. But it is a good point that we should be trying to find a way to lower the AST into more unambiguous representations.

tests/codegen/strcontains.cpp
danobi: Good idea. Done.

src/ast/irbuilderbpf.cpp
danobi: Oops, thanks for catching these.

src/ast/irbuilderbpf.cpp
danobi: Excellent point, I assumed LLVM had short circuit semantics. Let me look into this. Maybe this is why verifier is puking.

danobi: I don't see special handling for if and while conditionals. So if it's broken here, they're all broken.

viktormalik: Correct me if I'm wrong but I believe that short circuit semantics needs to be implemented on the level of LLVM IR.

For instance, C code `if (A && B) { X } else { Y }` would be compiled (by Clang) into something like
```llvm
cond_1_bb:
  %a = ...eval A...
  br %a, label %cond_2_bb, label %Y_bb

cond_2_bb:
  %b = ...eval B...
  br %b, label %X_bb, label %Y_bb

X_bb:
  ...do X...

Y_bb:
  ...do Y...
```
So, we need to do something similar here. For the `strcontains.ll` test case, it should be (roughly):
```llvm
strcontains.outer.cond.1:
  %3 = load i64, ptr %strcontains.i, align 8
  %4 = icmp ult i64 %3, 17
  br %4, label %strcontains.outer.cond.2, label %strcontains.false

strcontains.outer.cond.2:
  %5 = getelementptr i8, ptr @hello-test-world, i64 %3
  %6 = load i8, ptr %5, align 1
  %7 = icmp ne i8 %6, 0
  br %7, label %strcontains.outer.body, label %strcontains.false
```
This way, you make sure that the load inst is not executed if `i` (`%3`) is out of bounds.

src/ast/irbuilderbpf.cpp
danobi: Ah I see https://llvm.org/docs/LangRef.html#id319 . Trying to wrap my head around that example :P

src/ast/irbuilderbpf.cpp
danobi: Yeah, you're absolutely right. FWIW, I wasn't disagreeing - just pointing out we may have "miscompiles" elsewhere.

danobi:
Changes:
* Fixed the 2 eagerly evaluated conditionals in irbuilder
* Simplified inner conditional (`haystack[i+j] == '\0'` was unnecessary, as the `needle[j]` is already non-nul and there's a subsequent `haystack[i+j] != needle[j]`)

tests/codegen/llvm/literal_strncmp.ll
jordalgo: Sounds good to me.

tests/runtime/probe
danobi: With @viktormalik 's help, figured it out. I was generating bad code. Fixed to work on older LLVMs.

src/ast/irbuilderbpf.cpp
danobi: Lemme follow up on this in a separate PR.

tests/codegen/strcontains.cpp
```diff
@@ -0,0 +1,20 @@
...
+}
+
+TEST(codegen, strcontains_one_literal)
```
jordalgo, line 13: nit addtion: stringcontains_no_literals ?

jordalgo (approved):
Nice work!

tests/codegen/strcontains.cpp
danobi: Ack. good idea

src/ast/irbuilderbpf.cpp
viktormalik: Yeah, I got that. Could be the case, although this is quite a specific case when we explicitly rely on the short-circuit evaluation, otherwise we get UB.

src/ast/irbuilderbpf.cpp
```diff
@@ -1669,143 +1644,156 @@ Value *IRBuilderBPF::CreateStrncmp(Value *str1,
...
+  CreateCondBr(haystack_cmp_outbounds, outer_incr, inner_cmp);
+
+  // Inner conditional matching haystack char with needle char
+  SetInsertPoint(inner_cmp);
+  Value *haystack_cmp_char = CreateLoad(
+      getInt8Ty(), CreateGEP(getInt8Ty(), haystack, { haystack_cmp_idx }));
+  Value *haystack_cmp_needle = CreateICmpNE(haystack_cmp_char, needle_char);
+  Value *haystack_cmp_cond = CreateOr(haystack_cmp_outbounds,
+                                      haystack_cmp_needle);
+  CreateCondBr(haystack_cmp_cond, outer_incr, inner_incr);
```
viktormalik, line 1773: The `or` is no longer needed (already checked few lines above).
```suggestion
  CreateCondBr(haystack_cmp_needle, outer_incr, inner_incr);
```

viktormalik (commented):
One more comment for the IR, otherwise LGTM.

src/ast/irbuilderbpf.cpp
danobi: Oops, good catch

viktormalik (approved):
Nice!

PR #3816, opened 2025-02-24
POC: Hygienic macros

danobi:
This is a proof of concept for hygienic macros.

##### Checklist

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -702,12 +702,12 @@ call:
         |       CALL "(" vargs ")"           { $$ = driver.ctx.make_node<ast::Call>($1, std::move($3), @$); }
         |       CALL_BUILTIN  "(" ")"        { $$ = driver.ctx.make_node<ast::Call>($1, @$); }
         |       CALL_BUILTIN "(" vargs ")"   { $$ = driver.ctx.make_node<ast::Call>($1, std::move($3), @$); }
-        |       IDENT "(" ")"                { error(@1, "Unknown function: " + $1); YYERROR;  }
-        |       IDENT "(" vargs ")"          { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       BUILTIN "(" ")"              { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       BUILTIN "(" vargs ")"        { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       STACK_MODE "(" ")"           { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       STACK_MODE "(" vargs ")"     { error(@1, "Unknown function: " + $1); YYERROR;  }
+        |       IDENT "(" ")"                { $$ = driver.ctx.make_node<ast::Call>($1, @$); }
+        |       IDENT "(" vargs ")"          { $$ = driver.ctx.make_node<ast::Call>($1, std::move($3), @$); }
                 ;
 
 map:
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -326,7 +326,7 @@ macros:
         |       %empty       { $$ = ast::MacroList{}; }
 
 macro:
-                MACRO IDENT "(" macro_args ")" "{" expr "}" { $$ = driver.ctx.make_node<ast::Macro>($2, std::move($4), $7, @$); }
+                MACRO IDENT "(" macro_args ")" block_expr { $$ = driver.ctx.make_node<ast::Macro>($2, std::move($4), $6, @$); }
 
 macro_args:
                 macro_args "," map { $$ = std::move($1); $$.push_back($3); }
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -107,6 +107,7 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %token <std::string> INT_TYPE "integer type"
 %token <std::string> BUILTIN_TYPE "builtin type"
 %token <std::string> SUBPROG "subprog"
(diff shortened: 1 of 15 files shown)


danobi:
The code started getting sloppier as the weekend went on. But seems to work alright for a POC

src/parser.yy
```diff
@@ -180,8 +185,8 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %%
 
 program:
-                c_definitions config probes_and_subprogs END {
-                    driver.ctx.root = driver.ctx.make_node<ast::Program>($1, $2, std::move($3.second), std::move($3.first));
+                c_definitions config macros probes_and_subprogs END {
```
jordalgo, line 188: Do we care where macros are defined? Meaning above/below config and c_definitions. I was thinking about this as well for global map declarations.

jordalgo:
Should allow and maybe remove semi-colons at the end of the single expression? e.g. this fails to parse:
`macro add($a) { $a + 1; } BEGIN { $a=2; $b = add($a); print($b); exit(); }`
as does this `macro add($a) { $a + 1; } BEGIN { $a=2; $b = add($a) print($b); exit(); }`

jordalgo:
First off. This is very cool! Thanks @danobi for getting this started.

Just doing some red-team analysis here. These are scripts I tried that didn't work - I'm sure with some expected failures like map assignment

// no variable assignments
```
macro add($a) { $b = $a + 1 }
BEGIN { $a=2; $b = add($a); print($b); exit(); }
```

// no semicolons at the end of macro expression
```
macro add($a) { $a + 1; }
BEGIN { $a=2; $b = add($a); print($b); exit(); }
```

// macros can't call macros
```
macro sub($a) { $a - 5 }
macro add($a) { sub($a - 1) }
BEGIN { $a=2; print(add($a)); exit(); }
```

// adding and assigning to a map should get the unhygenic error
// right now it's a syntax error
```
macro add($a, @b) { @b += $a }
BEGIN { $a=2; @b = 5; add($a, @b); exit(); }
```

// no for loops in macros
```
macro printall(@b) { for ($kv : @b) { print($kv.1) } }
BEGIN { @b[1] = 5; printall(@b); exit(); }
```

// causes a segfault
```
macro getone() { 1 }
BEGIN { $a = getone($a); print($a); exit(); }
```

// causes a segfault
```
macro getone() { 1 }
BEGIN { $a = getstack($b); print($a); exit(); }
```

// Incorrect error
```
macro sprint($a) { print((1, $a)) }
BEGIN { $b = "avenrun"; sprint($a); exit(); }
```

src/parser.yy
danobi: I don't think we care. I'll leave it as-is for now, but we should definitely come by later and loosen restrictions for all constructs.

danobi:
> First off. This is very cool! Thanks @danobi for getting this started.
> 
> Just doing some red-team analysis here. These are scripts I tried that didn't work - I'm sure with some expected failures like map assignment
> 
> // no variable assignments
> 
> ```
> macro add($a) { $b = $a + 1 }
> BEGIN { $a=2; $b = add($a); print($b); exit(); }
> ```

Macro body is not an expression

> 
> // no semicolons at the end of macro expression
> 
> ```
> macro add($a) { $a + 1; }
> BEGIN { $a=2; $b = add($a); print($b); exit(); }
> ```

Ditto

> 
> // macros can't call macros
> 
> ```
> macro sub($a) { $a - 5 }
> macro add($a) { sub($a - 1) }
> BEGIN { $a=2; print(add($a)); exit(); }
> ```

Ack. I left this out of the PoC. Not terribly difficult to implement - just need to create some kind of a graph and resolve it.

> 
> // adding and assigning to a map should get the unhygenic error // right now it's a syntax error
> 
> ```
> macro add($a, @b) { @b += $a }
> BEGIN { $a=2; @b = 5; add($a, @b); exit(); }
> ```

Macro body not an expression.

> 
> // no for loops in macros
> 
> ```
> macro printall(@b) { for ($kv : @b) { print($kv.1) } }
> BEGIN { @b[1] = 5; printall(@b); exit(); }
> ```

Ditto.

> 
> // causes a segfault
> 
> ```
> macro getone() { 1 }
> BEGIN { $a = getone($a); print($a); exit(); }
> ```
> 

Fixed now.

> // causes a segfault
> 
> ```
> macro getone() { 1 }
> BEGIN { $a = getstack($b); print($a); exit(); }
> ```

Also fixed now.
> 
> // Incorrect error
> 
> ```
> macro sprint($a) { print((1, $a)) }
> BEGIN { $b = "avenrun"; sprint($a); exit(); }
> ```

Ack.

danobi:
But thanks a lot for trying it out. It's clear to me now that improving the error messages are critical. Especially for the body-is-expr rule. Should be possible to get bison to help with that.

danobi:
I'd wager that most users are going to want to use a block expression in the macro body 99% of the time. And even if they wanted a different kind of expression, they can still put it in a block expression. So how about instead of defining a macro as:

```
macro foo() { <EXPR> }
```

we instead define it as:

```
macro foo() <BLOCK-EXPR>
```

It's not quite as "clean" as the former w.r.t. language design, but it'll solve a real usability issue.

jordalgo:
@danobi Can you give some examples of what is allowed or not allowed if we define a macro that way? Since a block_expr ends with an expr - I'm not sure what we're not allowing with this syntax `macro foo() { <EXPR> }` ?

jordalgo:
Continuing this work here: https://github.com/bpftrace/bpftrace/pull/4037

PR #3821, opened 2025-02-24
Add structured passes with intermediate values

amscanne:
This is a series of logical commits (please do not flatten), which implements a series of changes to the pass plumbing. Currently, it represents about half the changes described in #3827.

##### Checklist

- ~[ ] Language changes are updated in `man/adoc/bpftrace.adoc`~
- ~[ ] User-visible and non-trivial changes updated in `CHANGELOG.md`~
- ~[ ] The new behaviour is covered by tests~

PR #3823, opened 2025-02-25
Add diagnostics helper class

amscanne:
The `Diagnostic` class encapsulates an error with a source location. The `Diagnostics` class contains a collection of different `Diagnostic` instances at different severity levels. This can be used to collect errors and warnings from passes in a structured way for later emission.

##### Checklist

### DIFF
--- a/src/ast/diagnostic.h
+++ b/src/ast/diagnostic.h
@@ -1,0 +1,124 @@
+#pragma once
+
+#include <cassert>
+#include <functional>
+#include <memory>
+#include <sstream>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "location.hh"
+
+namespace bpftrace {
+namespace ast {
+
+// Diagnostic reflects a single error at a single source location. This is a
+// simple wrapper around a string for that message, and the location class.
+class Diagnostic {
+public:
+  enum class Severity {
+    Warning,
+    Error,
+  };
+  Diagnostic(location loc) : loc_(loc) {};
+  const std::string msg() const
+  {
+    return msg_.str();
+  }
+  const std::string hint() const
+  {
+    return hint_.str();
+  }
+  const location& loc() const
+  {
+    return loc_;
+  }
+
+  // Each diagnostic can potentially have a hint attached, which is how to
+  // effectively resolve this issue.
+  std::stringstream& addHint()
+  {
+    return hint_;
+  }
+
+  template <typename T>
+  Diagnostic& operator<<(const T& t)
+  {
+    msg_ << t;
+    return *this;
+  }
+
+private:
+  std::stringstream msg_;
+  std::stringstream hint_;
+  location loc_;
+};
+
+class Diagnostics {
+public:
+  using Severity = Diagnostic::Severity;
+
+  template <typename... Args>
+  Diagnostic& add(Severity severity, Args... args)
+  {
+    auto index = static_cast<long unsigned int>(severity);
+    if (diagnostics_.size() <= index) {
+      diagnostics_.resize(index + 1);
+    }
+    auto& diags = diagnostics_[index];
+    auto& p = diags.emplace_back(std::make_unique<Diagnostic>(args...));
+    return *p.get();
+  }
+
+  template <typename... Args>
+  Diagnostic& addError(Args... args)
+  {
+    return add(Severity::Error, args...);
+  }
+  template <typename... Args>
+  Diagnostic& addWarning(Args... args)
+  {
+    return add(Severity::Warning, args...);
+  }
+  bool has(Severity severity) const
+  {
+    auto index = static_cast<long unsigned int>(severity);
+    return diagnostics_.size() > index && diagnostics_[index].size() > 0;
+  }
+  void clear()
+  {
+    for (auto& diag : diagnostics_) {
+      diag.clear();
+    }
+  }
+
+  // emit implements a default formatter of all diagnostics to a given stream.
+  // The use of `emit` should be generally discouraged, especially by tests,
+  // who should use more structured checks and avoid matching against the
+  // specific format here.
+  void emit(std::ostream& out);
+
+private:
+  void foreach(Severity severity,
+               std::function<void(const Diagnostic&)> fn) const
+  {
+    auto index = static_cast<long unsigned int>(severity);
+    if (diagnostics_.size() <= index) {
+      return;
+    }
+    for (const auto& diag : diagnostics_[index]) {
+      fn(*diag.get());
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


amscanne:
This is one part of a larger structured diagnostics change. See #3821.

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+namespace bpftrace {
+namespace ast {
+
+// Diagnostic reflects a single error at a single source location. This is a
+// simple wrapper around a string for that message, and the location class.
+class Diagnostic {
+public:
+  enum class Severity {
+    Warning,
+    Error,
```
viktormalik, line 22: Do we also have (or want to have) `Info` severity?

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+  // emit implements a default formatter of all diagnostics to a given stream.
+  // The use of `emit` should be generally discouraged, especially by tests,
+  // who should use more structured checks and avoid matching against the
+  // specific format here.
+  void emit(std::ostream& out);
+
+private:
+  // N.B. we store diagnostics as a pointer because the lifetime is returned
+  // early above, so they must not be moving at any point.
+  std::vector<std::vector<std::unique_ptr<Diagnostic>>> diagnostics_;
```
viktormalik, line 118: It would be also worth explaining the semantics of the vector indices here (IIUC, especially the first vector groups by severity which is not immediately clear).

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+    return hint_.str();
+  }
+  const location& loc() const
+  {
+    return loc_;
+  }
+
+  // Each diagnostic can potentially have a hint attached, which is how to
+  // effectively resolve this issue.
+  std::stringstream& addHint()
```
viktormalik, line 40: Judging by the name, shouldn't this write to `hint_`?

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+  Diagnostic& add(Severity severity, Args... args)
+  {
+    auto index = static_cast<long unsigned int>(severity);
+    if (diagnostics_.size() <= index) {
+      diagnostics_.resize(index + 1);
+    }
+    auto& diags = diagnostics_[index];
+    int which = diags.size();
+    diags.emplace_back(std::make_unique<Diagnostic>(args...));
+    return *diags[which].get();
```
viktormalik, line 73: I think `emplace_back` returns a reference to the inserted element since C++17. Since we're on C++20 already, it should be possible to use it instead of `which`.

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+public:
+  using Severity = Diagnostic::Severity;
+
+  template <typename... Args>
+  Diagnostic& add(Severity severity, Args... args)
+  {
+    auto index = static_cast<long unsigned int>(severity);
+    if (diagnostics_.size() <= index) {
+      diagnostics_.resize(index + 1);
+    }
```
viktormalik, line 69: Since we have only 2 severity levels, wouldn't it make more sense to resize `diagnostics_` upon creation instead of doing it lazily? I think it would allow to simplify this and some other methods below.

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+  {
+    return add(Severity::Warning, args...);
+  }
+  bool has(Severity severity) const
+  {
+    auto index = static_cast<long unsigned int>(severity);
+    return diagnostics_.size() > index && diagnostics_[index].size() > 0;
+  }
+  void foreach(Severity severity,
+               std::function<void(const Diagnostic&)> fn) const
```
viktormalik, line 92: I'm not a big fan of functor-style iteration instead of good old `for` loop which is IMO much more readable in the code. Do we anticipate this to be used outside of this class? If not, then at least let's make it private.

viktormalik (commented):
It's quite hard to review without seeing actual usage but I've added a couple of comments.

src/ast/diagnostic.h
jordalgo: We don't have `LOG(INFO)` at the moment but instead `LOG(V1)`. I don't think we need either location or hints for these types of messages.

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+    return *this;
+  }
+
+private:
+  std::stringstream msg_;
+  std::stringstream hint_;
+  location loc_;
+};
+
+// Diagnostics contains a collection of diagnostics.
```
jordalgo, line 58: Not a very helpful comment IMHO :)

src/ast/diagnostic.h
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+  void clear()
+  {
+    for (auto& diag : diagnostics_) {
+      diag.clear();
+    }
+  }
+
+  // emit implements a default formatter of all diagnostics to a given stream.
```
jordalgo, line 109: Little confused by this comment. Why make it public then?

src/ast/diagnostic.h
jordalgo: Is it silly for this function to just accept a string and add it to the stream itself?

src/ast/diagnostic.h
jordalgo: I find it useless. You could re-word it but as is, I would remove.

src/ast/diagnostic.h
amscanne: The existing code for adding a hint uses a stream, and a stream makes the most sense. Otherwise I would be reworking that code to use a `std::stringstream` and then making a call to add hint?

See the subsequent changes:
Screenshot 2025-02-25 at 2 27 45 PM

src/ast/diagnostic.h
amscanne: Candidly, I don't think it was worth the time to iterate on this, but the comment is now gone.

amscanne:
I've written up the broader intent here: #3827 (note that this is just one part of one bullet, so there's a lot stacked behind this).

src/ast/diagnostic.h
viktormalik: Sure, no problem, makes sense.

src/ast/diagnostic.h
viktormalik: Yeah, stream makes sense. I can see now that this is consistent with `addError()` etc. below so let's keep `addHint()` (even though it still reads a bit weird to me).

viktormalik (approved):
I like this!

jordalgo (approved):

PR #3824, opened 2025-02-25
Fix flex scanner push-back overflow issue

### LINKED ISSUE #3215: flex scanner push-back overflow

```bpftrace
#include <linux/watch_queue.h>

tracepoint:syscalls:sys_enter_ioctl
/args->cmd == IOC_WATCH_QUEUE_SET_FILTER/
{
    @ioctl_count[pid]++;
}

tracepoint:syscalls:sys_enter_socket
{
    @socket_count[pid]++;
}

tracepoint:syscalls:sys_enter_pipe2
{
    @pipe2_count[pid]++;
}

tracepoint:syscalls:sys_enter_msgget
{
    @msgget_count[pid]++;
}

tracepoint:syscalls:sys_enter_msgsnd
{
    @msgsnd_count[pid]++;
}

tracepoint:syscalls:sys_enter_msgrcv
{
    @msgrcv_count[pid]++;
}

tracepoint:syscalls:sys_enter_read
{
    @read_count[pid]++;
}

tracepoint:syscalls:sys_enter_write
{
    @write_count[pid]++;
}

tracepoint:syscalls:sys_enter_close
{
    @close_count[pid]++;
}

tracepoint:syscalls:sys_enter_execve
{
    delete(@ioctl_count[pid]);
    delete(@socket_count[pid]);
    delete(@pipe2_count[pid]);
    delete(@msgget_count[pid]);
    delete(@msgsnd_count[pid]);
    delete(@msgrcv_count[pid]);
    delete(@read_count[pid]);
    delete(@write_count[pid]);
    delete(@close_count[pid]);
}

```

```
❯ sudo bpftrace test.bt
flex scanner push-back overflow
```

### `bpftrace --info` output

```
System
  OS: Linux 6.5.13-7-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Dec 20 07:15:58 UTC 2023
  Arch: x86_64

Build
  version: v0.20.4
  LLVM: 17.0.6
  unsafe probe: no
  bfd: yes
  libdw (DWARF support): yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes
  get_boot_ns: yes
  dpath: yes
  skboutput: yes
  get_tai_ns: yes
  get_func_ip: yes
  jiffies64: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes
  module btf: yes
  map batch: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes
  ringbuf: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
  kprobe_multi: yes
  uprobe_multi: no
  raw_tp_special: yes
  iter: yes

```

wokron:
Fixes #3215 

This issue was caused by two factors:
1. Unlike `yy_create_buffer()`, the buffer created by calling `yy_scan_string()` was only the size of the string, not `YY_BUF_SIZE`, which makes it easy to encounter buffer overflow.
2. During macro processing, `YY_BUF_SIZE` was assumed to be the **remaining** buffer size, leading to calls to `unput()` even when the buffer was full. (#2313)

I defined the `yy_scan_string_with_size()` function to create a buffer of a specified size (`MAX_BUF_SIZE`). And defined the `yy_is_buffer_full()` function to check if the buffer is full, preventing calls to `unput()` when the buffer is already full.

##### Checklist

### DIFF
--- a/src/driver.cpp
+++ b/src/driver.cpp
@@ -1,10 +1,7 @@
 #include "driver.h"
 #include "parser.tab.hh"
 
-struct yy_buffer_state;
-
-extern struct yy_buffer_state *yy_scan_string(const char *yy_str,
-                                              yyscan_t yyscanner);
+extern void set_source_string(const std::string *s);
 extern int yylex_init(yyscan_t *scanner);
 extern int yylex_destroy(yyscan_t yyscanner);
 extern bpftrace::location loc;
@@ -22,7 +19,7 @@ void Driver::parse()
   if (debug) {
     parser.set_debug_level(1);
   }
-  yy_scan_string(ctx.source_->contents.c_str(), scanner);
+  set_source_string(&ctx.source_->contents);
   parser.parse();
   yylex_destroy(scanner);
 }
--- a/src/driver.cpp
+++ b/src/driver.cpp
@@ -4,14 +4,15 @@
 extern void set_source_string(const std::string *s);
 extern int yylex_init(yyscan_t *scanner);
 extern int yylex_destroy(yyscan_t yyscanner);
-extern bpftrace::location loc;
 
 namespace bpftrace {
 
 void Driver::parse()
 {
-  // Reset source location info on every pass.
+  // Reset state on every pass.
   loc.initialize();
+  struct_type.clear();
+  buffer.clear();
 
   yyscan_t scanner;
   yylex_init(&scanner);
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -18,6 +18,18 @@ static std::string buffer;
 #define yyterminate() return bpftrace::Parser::make_END(loc)
 
 using namespace bpftrace;
+
+// Since `YY_INPUT` cannot access the `driver` variable, `source` and `curr` 
+// are defined as global variables. They are marked as thread_local to 
+// ensure thread safety during lexical analysis.
+static thread_local const std::string *source;
+static thread_local size_t curr;
+
+void set_source_string(const std::string *s);
+static int read_from_source(char* buf, size_t max_size);
+
+#define YY_INPUT(buf,result,max_size) \
+  result = read_from_source(buf, max_size);
 %}
 
 /* https://en.cppreference.com/w/cpp/language/integer_literal#The_type_of_the_literal */
@@ -266,18 +278,16 @@ struct|union|enum       {
                             }
                             else
                             {
-                              int leng = YY_BUF_SIZE;
                               std::string original_s(s);
                               std::string original_yytext(yytext);
-                              for (z=strlen(s) - 1; z >= 0; z--){
-                                if (unput_count >= 1000 || leng <= 0) {
+                              for (z=strlen(s) - 1; z >= 0; z--) {
+                                if (unput_count >= 1000) {
                                   driver.error(loc, std::string("Macro recursion limit reached: ")
                                                     + original_yytext + ", " + original_s);
                                   yyterminate();
                                 }
                                 unput(s[z]);
                                 unput_count++;
-                                leng--;
                               }
                             }
                           } else {
@@ -290,3 +300,21 @@ struct|union|enum       {
                                             std::string(yytext) + std::string("'")); }
 
 %%
+
+void set_source_string(const std::string *s) {
+  source = s;
+  curr = 0;
+}
+
+// Here we replaced the original YY_INPUT with the read_from_source() function,
+// allowing flex to read the source code from a string rather than a file. In
+// this case, flex uses a buffer size of YY_BUF_SIZE (16384) and reads up to
+// YY_READ_BUF_SIZE (8192) at a time. This gives us enough space for macro
+// expansion. Additionally, just like reading from a file, flex's internal
+// buffer management can handle cases where the string size exceeds YY_BUF_SIZE.
+int read_from_source(char* buf, size_t max_size) {
+  size_t num_to_copy = std::min(source->size() - curr, max_size);
+  source->copy(buf, num_to_copy, curr);
+  curr += num_to_copy;
+  return num_to_copy;
+}
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -10,12 +10,8 @@
 #include "util/int_parser.h"
 #include "util/format.h"
 
-bpftrace::location loc;
-static std::string struct_type;
-static std::string buffer;
-
-#define YY_USER_ACTION loc.columns(yyleng);
-#define yyterminate() return bpftrace::Parser::make_END(loc)
+#define YY_USER_ACTION driver.loc.columns(yyleng);
+#define yyterminate() return bpftrace::Parser::make_END(driver.loc)
 
 using namespace bpftrace;
 
@@ -73,8 +69,8 @@ oct_esc  [0-7]{1,3}
 
 %%
 
-{hspace}+               { loc.step(); }
-{vspace}+               { loc.lines(yyleng); loc.step(); }
+{hspace}+               { driver.loc.step(); }
+{vspace}+               { driver.loc.lines(yyleng); driver.loc.step(); }
 
 ^"#!".*$                // executable line
 "//".*$                 // single-line comments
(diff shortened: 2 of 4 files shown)


src/driver.cpp
```diff
@@ -7,8 +7,12 @@
 
 struct yy_buffer_state;
 
-extern struct yy_buffer_state *yy_scan_string(const char *yy_str,
-                                              yyscan_t yyscanner);
+#ifndef YY_BUF_SIZE
+#define YY_BUF_SIZE 16384
```
jordalgo, line 11: Isn't this the default size? When is the case that it's undefined?

src/lexer.l
```diff
@@ -267,18 +270,20 @@ struct|union|enum       {
...
-                                  driver.error(loc, std::string("Macro recursion limit reached: ")
-                                                    + original_yytext + ", " + original_s);
+                                if (unput_count >= 1000 || yy_is_buffer_full(yyg)) {
+                                  if (unput_count >= 1000)
+                                    driver.error(loc, std::string("Macro recursion limit reached: ")
+                                            + original_yytext + ", " + original_s);
+                                  else
+                                    driver.error(loc, std::string("Buffer overflow while expanding macro: ")
+                                          + original_yytext + ", " + original_s);
                                   yyterminate();
```
jordalgo, line 283: I think it reads better this way:
```suggestion
                                if (yy_is_buffer_full(yyg)) {
                                  driver.error(loc, std::string("Buffer overflow while expanding macro: ")
                                          + original_yytext + ", " + original_s);
                                    yyterminate();
                                } else if (unput_count >= 1000) {
                                  driver.error(loc, std::string("Macro recursion limit reached: ")
                                            + original_yytext + ", " + original_s);
                                  yyterminate();
                                }
                                    
                                 
```

jordalgo (changes requested):
Thanks for working on this! I see in the issue that @ajor has a minimal repo. Do you mind adding a test that failed before and now passes with this change?

src/driver.cpp
wokron: Since we only generated lex.yy.cc, we cannot include the `YY_BUF_SIZE` defined within it (which is a macro). Therefore, `YY_BUF_SIZE` is actually always undefined in driver.cpp. So I copied the definition of `YY_BUF_SIZE` here. Adding `#ifndef` allows users to configure the size of `YY_BUF_SIZE`, just like in lex.yy.cc.

wokron: In fact, the `YY_BUF_SIZE` here does not need to be consistent with the `YY_BUF_SIZE` in flex, so perhaps we can change the name. Considering that the length of the source string may be longer than `YY_BUF_SIZE`, I set the buffer size to `max(source.size(), YY_BUF_SIZE)` in my recent changes. Perhaps we can rename `YY_BUF_SIZE` here to `MIN_YY_BUF_SIZE`?

jordalgo: > Perhaps we can rename YY_BUF_SIZE here to MIN_YY_BUF_SIZE?

Yeah that makes sense to me. Thanks for explaining

tests/parser.cpp
```diff
@@ -2872,4 +2872,36 @@ BEGIN { $x: int8 = 1; }
 )");
 }
 
+TEST(Parser, macro_expansion_error)
+{
+  BPFtrace bpftrace;
+  // #define M M+1
+  bpftrace.macros_.emplace("M", "M+1");
```
jordalgo, line 2879: I know you're saving code here but can we just inline the macro in both tests?

Also is it possible to get a buffer overflow error that doesn't involve a macro that causes a recursion limit error?

wokron: > I know you're saving code here but can we just inline the macro in both tests?

At first, I did inline the macro into the source string, but I found that the `Driver` actually ignores my macro definition because the parsing of macro definitions is actually done in the `ClangParser`. So I directly insert the macro into the `macro_` map to mock the behavior of `ClangParser`. But I agree that the macro definition should be kept in the source string.

wokron: > Also is it possible to get a buffer overflow error that doesn't involve a macro that causes a recursion limit error?

Sure, but it only occurs when using macros from header files, just like the minimal repo in #3215. I'll change the test to this case.

jordalgo: > because the parsing of macro definitions is actually done in the ClangParser

Ah right. I forgot about that 😞

jordalgo (dismissed):

src/driver.cpp
```diff
@@ -7,8 +7,12 @@
 
 struct yy_buffer_state;
 
-extern struct yy_buffer_state *yy_scan_string(const char *yy_str,
-                                              yyscan_t yyscanner);
+#ifndef MIN_YY_BUF_SIZE
```
danobi, line 10: This is not some preprocessor definition exported by flex, right? IMO better to drop the `#ifndef` - it's confusing b/c it has no users (overrides) in our codebase.

src/driver.cpp
```diff
@@ -46,7 +50,11 @@ int Driver::parse()
   if (debug_) {
     parser.set_debug_level(1);
   }
-  yy_scan_string(Log::get().get_source().c_str(), scanner);
+  auto &source = Log::get().get_source();
+  yy_scan_string_with_size(source.c_str(),
+                           std::max(source.size(),
+                                    static_cast<size_t>(MIN_YY_BUF_SIZE)),
```
danobi, line 56: If you need this to be `size_t`, maybe just declare it as `static constexpr size_t`, then? Instead of a macro

src/lexer.l
```diff
@@ -291,3 +296,20 @@ struct|union|enum       {
...
+  b->yy_buf_size = size;
+
+  return b;
+}
+
+bool yy_is_buffer_full(struct yyguts_t *yyg) {
+  char *yy_cp = yyg->yy_c_buf_p;
+  return (yy_cp < YY_CURRENT_BUFFER_LVALUE->yy_ch_buf + 2 &&
+          YY_CURRENT_BUFFER_LVALUE->yy_n_chars == YY_CURRENT_BUFFER_LVALUE->yy_buf_size);
+}
```
danobi, line 315: While I'll grant this looks like it works (for now at least), it seems like this reaches way into the internals of flex. I suspect this will cause maintenance issues sooner or later, as I don't see any of these fields/types reference in the flex API.

Please consult https://westes.github.io/flex/manual/Index-of-Variables.html#Index-of-Variables and only use things in public API (or prove that it's indeed intended to be public but has no listing).

danobi (changes requested):
Appreciate the fix, but the biggest issue IMO is it reaches too deep into flex internals. I think we need a way to do this with public API.

src/lexer.l
wokron: You are right, I may have focused too much on the flex-generated `lex.yy.cc` file. So I reviewed the flex documentation and came up with another approach (2c44af97d7782e2b865e0504e6f7ac53436c17f2). This method is documented ([YY_INPUT, overriding](https://westes.github.io/flex/manual/Generated-Scanner.html#index-YY_005fINPUT_002c-overriding)). If it is better, I can squash this commit.

wokron: Basically, the new approach is to use a custom function to override the behavior of `YY_INPUT`, changing the input from a file to the source string we provide. In this case, the buffer size is the default `YY_BUF_SIZE`, and each read will copy up to `YY_READ_BUF_SIZE` bytes (half of the buffer size). This gives us enough space for macro expansion (larger than the macro recursion limit of 1000 bytes).

danobi:
Sorry about the delay - this slipped off my radar. I will take a look this weekend.

src/lexer.l
```diff
@@ -291,3 +298,15 @@ struct|union|enum       {
                                             std::string(yytext) + std::string("'")); }
 
 %%
+
+void yy_set_source_string(const std::string& s) {
+  source = s;
```
danobi, line 303: Technically a copy of `source` isn't necessary, right? Can read directly from driver instance? All we needed was an indirection.

tests/parser.cpp
```diff
@@ -2872,4 +2872,23 @@ BEGIN { $x: int8 = 1; }
...
+  test_parse_failure(bpftrace,
+                     "#define M M+1\n"
+                     "BEGIN { M; }",
+                     R"(
+stdin:2:9-343: ERROR: Macro recursion limit reached: M, M+1
+BEGIN { M; }
+        ~~~~
+stdin:2:9-343: ERROR: syntax error, unexpected end of file
+BEGIN { M; }
+        ~~~~
```
danobi, line 2890: I think parser inserts this, right? So not necessary to do anything fancy.
```suggestion
  test_parse_failure("#define M M+1\n"
                     "BEGIN { M; }",
                     R"(
stdin:2:9-343: ERROR: Macro recursion limit reached: M, M+1
BEGIN { M; }
        ~~~~
stdin:2:9-343: ERROR: syntax error, unexpected end of file
BEGIN { M; }
        ~~~~
```

src/lexer.l
```diff
@@ -291,3 +298,15 @@ struct|union|enum       {
...
+  curr = 0;
+}
+
+int yy_read_from_source(char* buf, size_t max_size) {
```
danobi, line 307: Would be good to document with a comment why this exists here. From your PR description, seems like the magic is:

> Unlike yy_create_buffer(), the buffer created by calling yy_scan_string() was only the size of the string, not YY_BUF_SIZE, which makes it easy to encounter buffer overflow.

So basically force flex to pessimistically allocate the buffer by not letting it see the string. Which makes sense and should work well enough for most cases. Although I wonder what happens if the input file is too big for the default buffer size...

tests/parser.cpp
wokron: Actually, it doesn't :joy:. Macro parsing is done by `ClangParser` (https://github.com/bpftrace/bpftrace/pull/3824#discussion_r1974983783). I have added a comment to clarify this.

src/lexer.l
```diff
@@ -19,6 +19,15 @@ static std::string buffer;
 #define yyterminate() return bpftrace::Parser::make_END(loc)
 
 using namespace bpftrace;
+
+static std::string source;
+static size_t curr;
```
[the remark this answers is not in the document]
wokron: For `source` and `curr`, it seems they cannot be moved to the `Driver` instance because `YY_INPUT` is used by an internal function of Flex (`yy_get_next_buffer`), and we have no way to pass `driver` into it. However, I have marked `source` and `curr` as `thread_local`, so this will not affect the thread safety of lexer.

As for the other global variables in `lexer.l`, I have moved them to the `Driver` class.

src/lexer.l
wokron: I have added comments explaining why `YY_INPUT` is used.

> Although I wonder what happens if the input file is too big for the default buffer size...

Actually, flex can handle this. It only reads a portion of the data at a time. I have added a test case about this.

tests/parser.cpp
danobi: Ah, got it. How about writing a new wrapper, maybe `test_macro_parse_failure(...)`? We recently overhauled the pass manager, so should be straightforward. Example of a I recently added: https://github.com/bpftrace/bpftrace/blob/882c1836aebb082995bda421cbfe30b43e5430aa/tests/deprecated.cpp#L11-L27

Reason is it'd be good to move the tests towards stricter input -> output and avoid reaching into classes to mock.

src/lexer.l
danobi: Great, thanks for that!

danobi (approved):
thanks for working on this!

viktormalik:
Do we want to backport this to the 0.23.x branch?

jordalgo:
@viktormalik Good reminder. I'll put up a backport PR.

PR #3825, opened 2025-02-25
Drop support for LLVM 14 and 15

viktormalik:
#3809 removed support for LLVM 14 and 15 and therefore we can drop all the code conditionally compiled for these versions.

In addition, LLVM 14 was the last one which required typed (non-opaque) pointers so we can now remove all pointer casts from codegen. This removes all usages of `CreatePointerCast`. In some cases, it was doing int-to-ptr cast so I explicitly replaced it by `CreateIntToPtr`. In general, the transformations were done such that codegen tests do not change.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ and this project adheres to
   - [#3808](https://github.com/bpftrace/bpftrace/pull/3808)
 #### Deprecated
 #### Removed
+- Drop support for LLVM 14 and 15
+  - [#3825](https://github.com/bpftrace/bpftrace/pull/3825)
 #### Fixed
 - Fix json output for none type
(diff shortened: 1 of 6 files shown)


jordalgo (approved):
Niiiiice!

danobi (commented):
Need to update `MIN_LLVM_MAJOR` in cmakelists.txt

Changelog entry couldn't hurt. But should be fairly obvious when cmake is updated - build will just hard error with a message

viktormalik:
> Need to update `MIN_LLVM_MAJOR` in cmakelists.txt
> 
> Changelog entry couldn't hurt. But should be fairly obvious when cmake is updated - build will just hard error with a message

Ok, good point. I added a changelog entry and bumped min supported LLVM version to 16.

Also updating the PR title as it now officially drops support for LLVM 14 and 15.

PR #3826, opened 2025-02-25
Add checks for adding the EventLoss map

jordalgo:
This seems to get added by default even
when nothing is writing to it. This adds
checks in resource_analyser if it's needed.

Additionally, don't increment the event
loss counter when we're writing error messages
to the perf buffer as this is potentially
confusing to users.

##### Checklist


tests/codegen/call_kstack.cpp
```diff
@@ -36,7 +36,7 @@ TEST(codegen, call_kstack_mapids)
   ast::CodegenLLVM codegen(driver.ctx, *bpftrace);
   bpftrace->bytecode_ = codegen.compile();
 
-  ASSERT_EQ(bpftrace->bytecode_.maps().size(), 8);
+  ASSERT_EQ(bpftrace->bytecode_.maps().size(), 7);
```
amscanne, line 39: I'm not sure I understand the point of this assert in the test. What if the spill rules change, and we need to update things. Why is `7` the right number here? I'm not really advocating the removal at this time, but I just want to gently introduce the idea of "`[change-detector tests](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html)`", which I think can add friction without a ton of value.

src/ast/passes/resource_analyser.cpp
```diff
@@ -359,6 +364,13 @@ void ResourceAnalyser::visit(Call &call)
     // and symbols resolved even when unavailable at resolution time
     resources_.probes_using_usym.insert(probe_);
   }
+
+  if (call.func == "print" || call.func == "printf" || call.func == "time" ||
+      call.func == "unwatch" || call.func == "join" || call.func == "exit" ||
+      call.func == "clear" || call.func == "zero" || call.func == "system" ||
+      call.func == "cat") {
```
amscanne, line 371: Ideally this is an attribute in the function registry that applies to builtins? It doesn't need to happen at this time, but it would be good to think about incrementally moving towards a strategy where data on the builtins isn't scattered and hard-coded ;)

tests/codegen/call_kstack.cpp
jordalgo: > I'm not sure I understand the point of this assert in the test.

Me neither. I'm fine to remove it. Also if the right maps aren't being created we'll hear about it in a runtime or codegen test.

src/ast/passes/resource_analyser.cpp
jordalgo: > moving towards a strategy where data on the builtins isn't scattered and hard-coded ;)

💯 - It pained me to write this if statement.

src/required_resources.h
```diff
@@ -146,6 +146,8 @@ class RequiredResources {
   // List of probes using userspace symbol resolution
   std::unordered_set<const ast::Probe *> probes_using_usym;
 
+  bool needs_event_loss_map = false;
```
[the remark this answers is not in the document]
jordalgo: Great idea. I actually was able to move the fentry recursion check into resource analyser so now it's all in one place.

src/ast/passes/resource_analyser.cpp
danobi: This list seems risky and feels complicated. There's no real runtime cost to having an unused single entry 8 byte map, right? I doubt any non-trivial programs will be missing all of these, so we'll have added a bunch of conditionals and complexity for not much. So what is this commit optimizing for?

jordalgo: I think this is a fair comment and TBH I was originally skeptical of submitting this PR because it didn't seem worth the tradeoff. I might still be in that camp but I also think that even though this conditional is ugly, it should be fairly safe as you get a nice verifier error or segfault if you try to reference this map in codegen and it doesn't exist so any runtime tests would catch this.

I'm fine to abandon and make a smaller PR just moving the recursion entry check to the resource analyser but I also think it's nice getting rid of unnecessary codegen (like unused maps).

jordalgo: >  we should probably not land the first commit

Fine with me - this is not a hill I care to die on.

>  Recursion analysis is only a codegen concern, right? Would it make more sense to throw it into codegen_llvm.cpp? Or codegen_helper.cpp?

Let me try to find a good home for it. Maybe I'll wait until after @amscanne 's pass refactor lands.

PR #3833, opened 2025-02-26
Wire `Diagnostics` to AST nodes

amscanne:
Stacked PRs:
 * #3835
 * #3834
 * __->__#3833

--- --- ---

### Wire `Diagnostics` to AST nodes

This connects the newly introduced `Diagnostics` and `Diagnostic` class
to the AST nodes.

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -4,7 +4,7 @@
 #include <string>
 #include <vector>
 
-#include "location.hh"
+#include "diagnostic.h"
 #include "types.h"
 #include "usdt.h"
 #include "utils.h"
@@ -61,7 +61,7 @@ enum class ExpansionType {
 
 class Node {
 public:
-  Node(location loc) : loc(loc) {};
+  Node(Diagnostics &d, location loc) : diagnostics_(d), loc(loc) {};
   virtual ~Node() = default;
 
   Node(const Node &) = delete;
@@ -69,6 +69,32 @@ public:
   Node(Node &&) = delete;
   Node &operator=(Node &&) = delete;
 
+  template <typename... Args>
+  Diagnostic &addError(Args &...args) const
+  {
+    if constexpr (sizeof...(Args) == 0) {
+      return diagnostics_.addError(loc);
+    } else {
+      return diagnostics_.addError(loc + (args.loc + ...));
+    }
+  }
+  template <typename... Args>
+  Diagnostic &addWarning(Args &...args) const
+  {
+    if constexpr (sizeof...(Args) == 0) {
+      return diagnostics_.addWarning(loc);
+    } else {
+      return diagnostics_.addWarning(loc + (args.loc + ...));
+    }
+  }
+
+private:
+  Diagnostics &diagnostics_;
+
+public:
+  // This is temporarily accessible by other classes because we don't have a
+  // clear `clone` operation at this time. Eventually this should be made
+  // private and we should rely on a clear model for cloning nodes.
   location loc;
 };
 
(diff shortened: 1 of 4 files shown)


src/ast/context.h
```diff
@@ -35,8 +36,19 @@ class ASTContext {
     return nodes_.size();
   }
 
+  // Callers should avoid mutating diagnostics through this method. It is
+  // non-const to allow for tests to clear the set, but this should be avoided
+  // except in the context of a test.
+  Diagnostics &diagnostics() const
```
jordalgo, line 42: Can this be private/protected except to the test friend classes?

src/ast/ast.h
```diff
@@ -61,22 +61,49 @@ enum class ExpansionType {
...
   Node(Node &&) = delete;
   Node &operator=(Node &&) = delete;
 
+  template <typename... Args>
+  Diagnostic &addError(Args &...args) const
+  {
+    if constexpr (sizeof...(Args) == 0) {
+      return diagnostics_.addError(loc);
+    } else {
+      return diagnostics_.addError(loc + (args.loc + ...));
```
jordalgo, line 78: This looks neat (automatically adding the location). Help my sad little brain understand the else branch - you add the node's location and then an additional args location? Mind pointing me to an example of this?

amscanne: It just automatically merges all the locations for any additional nodes that you pass. For example,

```
foo.addError(bar)
```

I just made this require a second node to keep a consistent API (and allow for future metadata, if we prefer to do that via some diagnostic structure instead of the location). I believe that there are one or two instances in the next change, which is why this was added. See `binop_ptr` in the semantic analyser, e.g. `binop.left->addWarning(*binop.right)`

src/ast/context.h
jordalgo: What about a test-only `clear_diagnostics` method so you can make that const?

amscanne: Given the choice, I'd rather just tweak the comment here. I don't object to mutations, it's just that it shouldn't be the path that callers find. I really just want to say something like:

```
Callers should avoid using this method for adding individual diagnostics. Where possible, these should
be added directly through the corresponding `Node` and its own `addError` and `addWarning` methods.
```

Does this sound good?

jordalgo: What was wrong with my suggestion? You said you'd prefer to make it const so until there is a use case outside of the tests having a separate method only exposed to tests seems to be the best option -- it could even be public and at least that way the single mutation is clear.  I don't particularly care for comments on usage sitting in header files. People just look for other examples around the code base on usage and don't bother reading such things.

amscanne: I think the current test behavior in these cases is not particularly desirable (running a pass, adding some diagnostic, clearing diagnostics, doing other things, rechecking) and I'm pushing back because I don't want to codify that by adding a `clear_diagnostics()` method that I just go and remove later. I think the desired end state is:
* `const Diagnostics& diagnostics() const`, and no extra methods.
* Tests that want to do multiple things have a clear `ast.clone()` method or whatever. (So that they mutate and check diagnostics on something that they then throw away.)

I'd rather just make it `const` later than introduce intermediate plumbing. It's a simple change that I can go back and do if you feel strongly, I'm only really objecting because it candidly feels like arranging deck chairs.

amscanne: I've added this as a [later cleanup](https://github.com/bpftrace/bpftrace/pull/3821/commits/cf79ad7250bf6527069fcaa76e8dad4c91f22759) in the queue. I can tweak this in the meantime, but there's a lot of churn from ordering changes.

jordalgo: > I've added this as a [later cleanup](https://github.com/bpftrace/bpftrace/pull/3821/commits/cf79ad7250bf6527069fcaa76e8dad4c91f22759) in the queue.

Appreciate that. I know the stack of things can make smaller changes a pain but at the same time I want to optimize for the health of the code base and things can often get lost/forgotten.

jordalgo (approved):

danobi (approved):

Issue #3731, opened 2025-01-23
8 million ways to die


amscanne:
<img src="https://m.media-amazon.com/images/M/MV5BZDVhYWVmZDMtYTJiZS00MmJjLWI5YjQtZDBiZDliMTBmY2Q3XkEyXkFqcGc@._V1_.jpg" width="200x"/>

Okay, there are only three that I'm aware of:
* `LOG(BUG)` seems to be used in some unreachable conditions, more commonly in codegen (86 times).
* `assert` is used liberally in codegen and the irbuilder, but also elsewhere (143 times).
* runtime exceptions are used mostly in the internals and around tooling (143 times).

The coding guidelines recommend the use of exceptions, stating:

> If an error is recoverable (meaning downstream callers of the code should be able to selectively handle, ignore, or further propagate), pass the error through the return value of the function.
> ...
> If an error is not recoverable, prefer throwing FatalUserException. Exceptions should not be used for recoverable errors.

There are also 36 try/catch blocks for exceptions.

This can be confusing, since there are lots of options to reach from.

I would propose the following:
* Eliminating unnecessary assertions where possible (e.g. adding -Werror=switch-enum and removing unnecessary `default:` clauses)
* Sticking to explicit `assert` statements (with the (portable `&& "this is the error"` style)[https://en.cppreference.com/w/cpp/error/assert]) for "should never happen" errors
* Adding error propagation using a suitable mechanism (e.g. std::optional, llvm::ErrorOr for llvm-bound code, etc.)
* Limiting the use of exceptions strictly to high-level runtime errors (in functions that would never conceivable be used transitively in a destructor). That fact that they may throw should be clearly documented (and documented transitively!). Exceptions should be avoided in low-level internals because it is not always clear whether callers would consider it recoverable or not.

Practically, I think this means slowly eliminating "LOG(BUG)", as those cases would be forced into one of the above (an assert, error propagation, or very rarely, a runtime exception).

I believe that this aligns with the existing coding guidelines against using exceptions if the error is recoverable. None of this is urgent, but I'm filing this bug in order to discuss and ensure that there is consensus. The existing `LOG(BUG)` and exception mechanisms have the following thorny issues:
* Exceptions can be thrown from anywhere, and therefore exception safety can be challenging. If exceptions are thrown from a stack unwinding context (e.g. a destructor) then behavior is not defined (and the program is typically immediately terminated).
[...]

PR #3834, opened 2025-02-26
Simplify pass plumbing by using diagnostics

amscanne:
Stacked PRs:
 * #3855
 * #3854
 * #3853
 * #3852
 * #3835
 * __->__#3834

--- --- ---

### Simplify pass plumbing by using diagnostics

This is the first pass in aligning some of the pass internals, and
eventually restructuring many of the tests to use the existing pass
plumbing rather than manually constructing each pass separately.


src/ast/irbuilderbpf.cpp
```diff
@@ -1440,16 +1440,16 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
...
-               << " this case is not yet handled";
+    builtin.addError() << "index register set without base register;"
+                       << " this case is not yet handled";
   }
   Value *result = nullptr;
   if (argument->valid & BCC_USDT_ARGUMENT_BASE_REGISTER_NAME) {
     int offset = 0;
     offset = arch::offset(argument->base_register_name);
     if (offset < 0) {
-      LOG(BUG) << "offset for register " << argument->base_register_name
```
jordalgo, line 1451: So there is a difference here between `LOG(BUG)` and `LOG(ERROR)` as per our coding_guidelines. The former is intended for developers (like an assert) that we did something wrong and the latter is for users - they did something wrong. How do we distinguish between the two in the new diagnostics world?

amscanne: Both still exist. In this specific case, I was a bit confused about which case it was. I don't think *bugs* should actually preserve meaningful call site information (i.e. you typically want the stack & internals, and if anything want the whole program to reproduce, not the single expression). In this case, I thought that you actually *do* want the specific site so that users could work around the issue, even if it is ultimately a bug. I think that often, we may choose to surface issues *as* diagnostics ("can't support X", "unable to do Y"), just so the user knows what is triggering it and can work around it. I think this is part of the intent of #3731 -- to avoid the squishy middle as much as possible, and try to separate into standard error paths (e.g. diagnostics like this, returning `Error` and handling it, etc.) and `assert` when things truly should "never happen".

I am happy to revert this line, I can definitely get a bit over-jealous is trying to normalize. Just say the word.

amscanne: I am reverting this (will be the good old `LOG(BUG)` when the queue is updated).

amscanne:
Gentle ping. The stack on top of this is a bit deeper, but I'm trying to avoid churn.

src/ast/irbuilderbpf.cpp
jordalgo: Sorry for the delayed response. Glad to hear both still exist for now. In terms of converting specific `LOG(BUG)`s - probably worth doing in a separate PR.

src/ast/diagnostic.h
```diff
@@ -93,6 +93,12 @@ class Diagnostics {
     }
   }
 
+  // ok is the recommended short-hand for `has(Severity::Error)`.
+  bool ok() const
```
jordalgo, line 97: hmmm, nitpicky but I worry about this now getting confused with the `OK` in result.

tests/semantic_analyser.cpp
```diff
@@ -30,9 +30,13 @@ void test_for_warning(BPFtrace &bpftrace,
   clang.parse(driver.ctx.root, bpftrace);
 
   ASSERT_EQ(driver.parse_str(input), 0);
-  std::stringstream out;
-  ast::SemanticAnalyser semantics(driver.ctx, bpftrace, out);
+  // Override to mockbpffeature.
```
jordalgo, line 33: unnecessary comment

tests/semantic_analyser.cpp
```diff
@@ -2597,9 +2608,7 @@ BEGIN { (long)cpu }
 stdin:1:9-15: ERROR: Cannot cast to "long"
 BEGIN { (long)cpu }
         ~~~~~~
-stdin:1:9-15: HINT: Did you mean "int64"?
-BEGIN { (long)cpu }
```
jordalgo, line 2601: Love that this isn't repeated anymore

jordalgo (approved):
Nice cleanup! 🧹

tests/semantic_analyser.cpp
amscanne: Hm, I think this actually got copied up from this comment in the other `test()` in this file. Removed. Not sure what it means in the other case :)

tests/semantic_analyser.cpp
amscanne: This should be the ~only functional change here.

src/ast/diagnostic.h
amscanne: Did you want me to capitalize it? It still means the same thing. You will just have:

```
return OK();
```

And:

```
ast.diagnostics().ok()
```

But you'll also have lots of:

```
auto ok = doThing();
if (!ok) {
   ...
 }
 ```
 
 Conceptually it's all the same, and has the same bool-like semantics (OK == good!). Where is the confusion coming in?

jordalgo: Nevermind me, I think I was jet-lagged when I made this comment. `ast.diagnostics().ok()` is fine.

PR #3836, opened 2025-02-27
Add `ErrorOr` and associated types

amscanne:
In order to avoid reinventing the wheel, these are imported directly from LLVM, with a header providing a layer of indirection.

These errors must be checked and consumed in order to proceed, providing a layer of safety and enforcing good practices for error checking.

##### Checklist

- ~[ ] Language changes are updated in `man/adoc/bpftrace.adoc`~ No language changes.
- ~[ ] User-visible and non-trivial changes updated in `CHANGELOG.md`~ No user visible changes.
- ~[ ] The new behaviour is covered by tests~ Aliases and documentation primarily.

### DIFF
--- a/src/util/result.h
+++ b/src/util/result.h
@@ -1,0 +1,109 @@
+#pragma once
+
+#include <iostream>
+
+#include "llvm/Support/Error.h"
+#include "llvm/Support/ErrorOr.h"
+#include "llvm/Support/raw_os_ostream.h"
+
+namespace bpftrace {
+
+// OK is used in order to indicate success without a value type.
+//
+// It is effectively a monostate class, but using this instead of the default
+// llvm `ErrorSuccess` or `Error` means that the bool operator is consistent.
+//
+// To indicate success, you can merely do `return OK()`.
+class OK {};
+
+// Despite the name, in LLVM `ErrorOr` corresponds to a `std::error_code`
+// whilst `Expected` corresponds to a richer LLVM `Error`. The latter is
+// broadly preferred for newer code, as it allows a much richer expression of
+// error information. In an attempt to avoid confusion, we standardize on only
+// the modern versions, but call it `Result` to make its purpose clear. In the
+// future, this could map to a newer standard, such as the standard library
+// `std::expected<T, Error>`, independently of the names in this namespace. The
+// purpose of this file is to approximately insulate against such changes (but
+// not to provide a perfect abstraction, a little bit of leakage is okay).
+//
+// For now, the LLVM `Error` class provides mandatory checking (errors cannot
+// be ignored, they must be explicitly forward or consumed), which is useful.
+//
+// You can use Result as follows:
+//
+//     Result<> foo() {
+//       return OK();
+//     }
+//
+//     Result<Value> bar() {
+//       return Value();
+//     }
+//
+template <typename T = OK>
+using Result = llvm::Expected<T>;
+
+// For error types, one can define custom error type by inheriting from
+// `ErrorInfo`. For example, you may construct an error as follows:
+//
+//    class BpfVerifierError : public ErrorInfo<BpfVerifierError> {
+//    public:
+//      static char ID;
+//      void log(llvm::raw_ostream &OS) const override {
+//        OS << "End of file reached.";
+//      }
+//    };
+//
+template <typename T>
+class ErrorInfo : public llvm::ErrorInfo<T> {
+public:
+  std::error_code convertToErrorCode() const override
+  {
+    return llvm::inconvertibleErrorCode();
+  }
+};
+
+// LLVM has its own set of stream classes and wrappers. Provide a simple
+// wrapper that supports `std::ostream` for convenience.
+std::ostream& operator<<(std::ostream& out, const llvm::Error& err);
+
+// All errors are constructed using `make_error<...>` with the error class.
+template <typename E, typename... Ts>
+auto make_error(Ts... args)
+{
+  return llvm::make_error<E>(std::forward<Ts>(args)...);
+}
+
+// For error handling there are several cases to consider:
+//
+// (1) If you want to propagate the error, you can return as expected:
+//
+//   auto ok = doAThing();
+//   if (!ok) {
+//     return ok.takeError();
+//   }
+//
+// (2) If you need to handle some cases, you can use `handleErrors`. Note that
+// you will still need to propagate unhandled cases via (1).
+//
+//   auto ok = doAThing();
+//   if (!ok) {
+//     auto nowOk = handleErrors(ok.takeError(),
+//                               [](const BpfVerifierError&) { ... });
+//     if (!nowOk) {
+//       return nowOk.takeError();
+//     }
+//  }
+template <typename Err, typename... Ts>
+Result<> handleErrors(Err err, Ts... args)
+{
+  // This changes the semantics of the error propagation to return the wrapper
+  // above. This keeps the `operator bool` behavior consistent.
+  auto llvmErr = llvm::handleErrors(std::forward<Err>(err),
+                                    std::forward<Ts>(args)...);
+  if (!llvmErr) {
+    return OK();
+  }
+  return llvmErr;
+}
+
+}; // namespace bpftrace
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -47,6 +47,7 @@ add_library(runtime STATIC
   debugfs.cpp
   usdt.cpp
   utils.cpp
+  util/result.cpp
   pcap_writer.cpp
   ksyms.cpp
   usyms.cpp
--- a/src/util/result.cpp
+++ b/src/util/result.cpp
@@ -1,0 +1,13 @@
+#include "util/result.h"
+
+namespace bpftrace {
+
+std::ostream& operator<<(std::ostream& out, const llvm::Error& err)
+{
+  llvm::raw_os_ostream raw_ostream(out);
+  raw_ostream << err;
+  raw_ostream.flush();
+  return out;
+}
+
+} // namespace bpftrace


src/util/error.h
```diff
@@ -0,0 +1,96 @@
...
+// this, but use the `Expected` type. In the future, this could map to
+// something like `std::expected<T, Error>`, and we could replace the LLVM
+// `Error` with our own.
+//
+// Out of the box, the LLVM `Error` class provides mandatory checking (errors
+// cannot be ignored, they must be explicitly forward or consumed).
+template <typename T>
+using ErrorOr = llvm::Expected<T>;
+
+// success is used locally in order to indicate success.
```
jordalgo, line 23: This sentence feels tautological to me.

src/util/error.h
```diff
@@ -0,0 +1,96 @@
...
+class ErrorInfo : public llvm::ErrorInfo<T> {
+public:
+  std::error_code convertToErrorCode() const override
+  {
+    return llvm::inconvertibleErrorCode();
+  }
+};
+
+// LLVM has its own set of stream classes and wrappers. Provide a simple
+// wrappers that supports `std::ostream` using the standardized wrappers.
```
jordalgo, line 53:
```suggestion
// wrapper that supports `std::ostream` using the standardized wrappers.
```

src/util/error.h
```diff
@@ -0,0 +1,96 @@
...
+using ErrorOr = llvm::Expected<T>;
+
+// success is used locally in order to indicate success.
+//
+// It is effectively a monostate class, but using this instead of the default
+// llvm `ErrorSuccess` or `Error` means that the bool operator is consistent.
+//
+// To indicate success, you can merely do `return success()`.
+class success {};
+using Error = ErrorOr<success>;
```
jordalgo, line 30: I feel like this `using` is confusing when reading the code below.

src/util/error.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include <iostream>
+
+#include "llvm/Support/Error.h"
+#include "llvm/Support/ErrorOr.h"
+#include "llvm/Support/raw_os_ostream.h"
+
+namespace bpftrace {
+
+// In LLVM, ErrorOr maps a std::error_code, and Expected maps a richer Error
```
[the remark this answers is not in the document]
amscanne: Does something like this help:

```
Despite the name, in LLVM `ErrorOr` corresponds to a `std::error_code` whilst `Expected`
corresponds to a richer LLVM `Error`. The latter is broadly preferred for newer code, as it
allows a much richer expression of error information. In an attempt to avoid confusion, we
standardize on only the modern versions, but still call it `Error` and `ErrorOr` to make its
purpose clear.
```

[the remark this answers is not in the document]
jordalgo: Much better!

src/util/error.h
jordalgo: Yes! Much better. 🚢  it

src/util/error.h
jordalgo: I feel like `Error` and `ErrorOr` are two distinct types. The former being ONLY an error (in that it contains an err code and/or additional metadata about said error) while the latter is like an optional that may contain an error or a success - it feels like renaming Optional to `Value`. Am I missing something?

amscanne: I'm quite confused. It seems like you're describing exactly what it is. Are you saying that you'd expect functions to explicitly declare `ErrorOr<success> foo()` instead of `Error foo()`? Or you want `std::optional<Error> foo()`?

One of the things that I'm trying to do with the aliases here is align the bool conventions are well. Consider the following two snippets:

```
auto ok = foo();
if (!ok) {
   return ok.takeError();
}
```

```
auto err = foo();
if (err) {
  return std::move(*err);
}
```

Which one is correct? Well, if you have `std::optional<Error>` then it's the second one. If you have `ErrorOr<T>` then it's the first one. I think callers should be immune to the value returned, if `foo` happens to change. That realy sucks and it makes it confusing because a good chunk of functions may have a return value, and others won't... leading to weird conventions.

What I'm trying to do here is simply define `Error` (which *can* represent a success, which is a fairly standard thing even for the LLVM error types) to be `ErrorOr<success>` so that the usage is consistent. Regardless of whether foo returns a value (`ErrorOr<T>`) or not (`Error`), you will *always* do:

```
auto ok = foo();
if (!ok) {
   return ok.takeError();
}
```

Does this help explain why I'm trying to achieve?

amscanne: Just for reference, the idea of a success sentinel within the Error type itself is what happens in LLVM, and sort of what my mental model is (https://llvm.org/doxygen/classllvm_1_1ErrorSuccess.html#details).

src/util/error.h
amscanne: Done.

src/util/error.h
jordalgo: Sorry for bringing `optional` into this. I was trying to use it as an example but it doesn't have anything to do with my point, which, for the record, is minor.

> Are you saying that you'd expect functions to explicitly declare ErrorOr<success> foo() instead of Error foo()?

This is more or less what I'm trying to say. I'd rather, if possible, something like `using ErrorSuccess ErrorOr<success>` to distinguish this special type of error by readers who don't know that `Error` can actually represent a success.

amscanne: > This is more or less what I'm trying to say. I'd rather, if possible, something like using ErrorSuccess ErrorOr<success> to distinguish this special type of error by readers who don't know that Error can actually represent a success.

Ahh, I understand now. I think I'm used to interpreting `Error` as a *possible* error (i.e. like `errno`, `std::error_code` or the raw `int` generally used at the moment). I think the always-error object here is `ErrorInfo`. In any case, I don't object to using `ErrorOr<Success>` explicitly and dropping the alias. It's just a bit verbose? What about `ErrorOr<OK>` and return `OK()`? That's a bit more terse, plus it more directly maps to the pattern that I'm advocating for.

I'd love to slowly move towards structured errors over exceptions (and am introducing this to use it). A lot of this is simply stylistic, so curious to hear thoughts/feedback/support/objections from others too!

amscanne:
Related: #3731. This is being added ahead of some additional internal plumbing, and I'd like to use structured errors (e.g. `ErrorOr`) and figured it's a reasonable to introduce generally.

src/util/error.h
jordalgo: I like `ErrorOr<OK>`. It's explicit. If you really want an alias it could be `ErrorOk`. I do understand your comparison to `std::error_code` and `errno` -  I think the lack of an "int"-ness in the name is what threw me.

Thinking out loud but what about using `Result` (like Rust) as an alias to `std::expected<T, Error>` or `llvm::Expected<T>`? I just think the name is more elegant than `ErrorOr` and it avoids the conflict/confusion with LLVM's `ErrorOr`.

amscanne: The alias was only for brevity. I'm also fine with `Result`. I think it makes the abstraction less bound, which is good. I've updated the PR with this approach.

Note that we could use `std::expected` in the future, but it's not C++20. The point of the wrapper here is to avoid the types being viral internally (easy swap in the future). Plus `llvm::Expected` has the lovely property that it forces error checking (you'll get an assertion failure if an error is destroyed without being checked).

jordalgo (approved):
This PR is `OK`!

src/util/result.h
```diff
@@ -0,0 +1,109 @@
...
+//
+template <typename T = OK>
+using Result = llvm::Expected<T>;
+
+// For error types, one can define custom error type by inheriting from
+// `ErrorInfo`. For example, you may construct an error as follows:
+//
+//    class BpfVerifierError : public ErrorInfo<BpfVerifierError> {
+//    public:
+//      static char ID;
```
danobi, line 50: In case anyone was wondering: this snippet from llvm source explains why this exists. It's quite clever IMO:
```
template <typename ThisErrT, typename ParentErrT = ErrorInfoBase>
class ErrorInfo : public ParentErrT {
public:
  using ParentErrT::ParentErrT; // inherit constructors

  static const void *classID() { return &ThisErrT::ID; }

  const void *dynamicClassID() const override { return &ThisErrT::ID; }

  bool isA(const void *const ClassID) const override {
    return ClassID == classID() || ParentErrT::isA(ClassID);
  }
};
```

Some kind of light weight RTTI

danobi (approved):
very nice. let's give this a try

src/util/result.h
```diff
@@ -0,0 +1,109 @@
...
+//   auto ok = doAThing();
+//   if (!ok) {
+//     auto nowOk = handleErrors(ok.takeError(),
+//                               [](const BpfVerifierError&) { ... });
+//     if (!nowOk) {
+//       return nowOk.takeError();
+//     }
+//  }
+template <typename T, typename... Ts>
+Result<> handleErrors(Result<T> err, Ts... args)
```
danobi, line 97: why by value? not perfect forwarding?

amscanne: Ah, for the error. Yeah it's because I made this into the `Result`. I can fix that.

src/util/result.h
```diff
@@ -0,0 +1,109 @@
...
+//   auto ok = doAThing();
+//   if (!ok) {
+//     auto nowOk = handleErrors(ok.takeError(),
+//                               [](const BpfVerifierError&) { ... });
+//     if (!nowOk) {
+//       return nowOk.takeError();
+//     }
+//  }
+template <typename Err, typename... Ts>
+Result<> handleErrors(Err err, Ts... args)
```
danobi, line 97: Don’t you need rval ref for perfect forwarding? 
```suggestion
Result<> handleErrors(Err &&err, Ts... args)
```

danobi:
Merged to unblock, but please check me on the rval ref thing

amscanne:
> Merged to unblock, but please check me on the rval ref thing

Totally. I wrote a test to validate also and tweaked the API a bit as a result (just for `handleErrors`). See #3856.

PR #3837, opened 2025-02-27
Fix ustack/kstack sizes

jordalgo:
Since making these LLVM struct types[1]
these types now have added padding
so the sizes in CreateStack need to
account for this.

[1] d543ce871e88ec06a1d17beb49a4bf2b772e99a2

##### Checklist


jordalgo:
I'll backport this to the release branch

amscanne (approved):

src/types.cpp
```diff
@@ -439,9 +439,9 @@ SizedType CreateRecord(const std::string &name, std::weak_ptr<Struct> record)
 SizedType CreateStack(bool kernel, StackType stack)
 {
   // These sizes are based on the stack key (see
-  // IRBuilderBPF::GetStackStructType)
+  // IRBuilderBPF::GetStackStructType) but include struct padding
   auto st = SizedType(kernel ? Type::kstack_t : Type::ustack_t,
-                      kernel ? 12 : 20);
+                      kernel ? 16 : 24);
```
amscanne, line 444: I don't know why my comment got lost somehow. Doh. Trying to recreate.

While you should submit this, I think that the having two separate type systems tied together by ad-hoc constants is pretty fraught. Especially when the *primary* type system is not the one that is actually doing the code generation. I don't doubt there are many subtle bugs lurking related to this disconnect.

I think improving this would be good for robustness and reliability in the ong-term, and I'd like to think that we could tie the internal type system into LLVM in some canonical way, rather than having those types being generated independently. E.g. I'm imagining the size of a `SizedType` as something like `size_t size() { return LLVMStoreSizeOfType(target_, llvm_type_); }` wrapping a stable LLVM type (rather than being constructed in each instance). Nothing wrong with having the wrappers, it's ensuring that the mapping is stable and sound that is the trick :)

(There's a lot of yak shaving and details to be sorted here, and I'm just adding this for some discussion & thought. After all, are really *three* type systems here (BTF/Dwarf, SizedType, LLVM). Ideally if we move to defining the internal types in native C and BPF C, the wrapper around these gets thinner and essentially the internal type classes are effectively the stable mapping between BTF & LLVM types. I think constants start disappearing at that point.)

chenhengqi (commented):
LGTM.

chenhengqi:
@jordalgo One minor question, does the following error related to map key size limit on old kernel:

```
invalid access to map value, value_size=1016 off=1016 size=8
R5 min value is outside of the allowed memory range
```

src/types.cpp
jordalgo: >  I think that the having two separate type systems tied together by ad-hoc constants is pretty fraught.

No disagreements here. Let's start fixing these up and shaving these yaks

jordalgo:
>  does the following error related to map key size limit on old kernel:

@chenhengqi I haven't seen that error recently, do you mind filing a separate issue with details?

src/types.cpp
danobi: We have `src/ast/async_event_types.h` that helps with this kinda thing. But that's for events. As a stop gap, we could have one for internal types, just to keep everything centralized and less error prone.

Longer term having a more rigorous treatment of internal/external types is definitely necessary.

PR #3841, opened 2025-02-27
Add support for LLVM 20

viktormalik:
- Fix compilation errors in codegen
- Bump nixpkgs to the latest unstable which contains LLVM 20.1.0-rc2
- Add LLVM 20 support

##### Checklist


jordalgo (approved):
LGTM!

I think it would also be good to reference this PR on the release docs as an example. I'll try to remember to do that later.

viktormalik:
> I think it would also be good to reference this PR on the release docs as an example. I'll try to remember to do that later.

Good idea. I also have a couple small things to update in the doc so let me handle it all together.

PR #3847, opened 2025-02-27
simpify fuzzing infrastructure

amscanne:
This change returns the fuzzing infrastructure to a building state (from not building), and dramatically simplifies the overall setup.

First, documentation regarding `libfuzzer` is removed, as this is two generations out of date. (Libfuzzer become Centipede, which in turn became fuzztest. There's a legitimate use case for fuzztest, but these are unit tests. A sample is added for this new style of test for tracepoint parsing, simply as a reference.

This can be re-added in the future as FuzzTest, which essentially extends the gtest plumbing more seamlessly.

Second, the dedicated fuzzing binary is removing and instead the actual main program can be built with relevant santizing. The existing `-testmode=codegen` is used to run a fuzz test, and the existing ability to read an input file is also used. This reduces duplication and redundancy.

Third, the `#ifdef FUZZ` statements are removed from the code. All of these statements are either a) never reached with the changes above, and b) easily mitigated by turning some other knob (such as redirecting stderr during fuzzing).

Finally, this commit adds a GitHub workflow that does a basic fuzzing smoke test, to ensure that the now-simplified fuzzing infrastructure remains working.

##### Checklist

- ~[ ] Language changes are updated in `man/adoc/bpftrace.adoc`~ No language changes.
- ~[ ] User-visible and non-trivial changes updated in `CHANGELOG.md`~ No user visible changes.


.github/include/ci.py
```diff
@@ -256,6 +254,57 @@ def run_runtime_tests():
     )
 
 
+def fuzz():
+    """
+    Run a basic fuzz smoke test.
+    """
+    # Setup the core pattern.
+    shell(cmd=["sh", "-c", "echo core | tee /proc/sys/kernel/core_pattern"], as_root=True)
```
danobi, line 262: Can we get away without this? It'd be nice to not modify host state too much, as `ci.py` is intended to be run locally as well as in CI. Given I don't see any slurping of the cores (so they're lost anyways), maybe just set `AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES`?

docs/fuzzing.md
```diff
@@ -12,58 +12,53 @@ generate input efficiently.
...
 When doing fuzzing, it is important to limit the number of AST nodes because otherwise, a fuzzer might
 keep generating a very long program that causes a stack overflow.  `BPFTRACE_MAX_AST_NODES` environment
 variable controls the maximum number of AST nodes.
 
 ## Fuzzing with AFL
+
 Here, I briefly describe the way to fuzz bpftrace with AFL. I highly recommend reading the documentation
 in the AFL's repository for further information.
 
 ### Install AFL (or AFLPlusPlus)
```
danobi, line 34: Point to nix instructions instead?

docs/fuzzing.md
```diff
@@ -12,58 +12,53 @@ generate input efficiently.
...
-CC=/path/to/AFLplusplus/afl-clang-fast \
-CXX=/path/to/AFLplusplus/afl-clang-fast++ \
-AFL_USE_ASAN=1 \
-cmake .. \
--DBUILD_FUZZ \
--DFUZZ_TARGET=codegen \
--DCMAKE_BUILD_TYPE=Debug \
--DBUILD_TESTING=0 \
--DBUILD_ASAN=1
+nix develp #.bpftrace-fuzz
```
danobi, line 47:
```suggestion
nix develop #.bpftrace-fuzz
```

docs/fuzzing.md
```diff
@@ -77,72 +72,30 @@ Then, start fuzzing! AFL and AddressSanitizer have a lot of settings, so please
...
-  the first argument is the script file name. If the argument is not given, it reads a script from stdin.
 - `-i` is the input directory, and `-o` is the output directory. In the input directory, you need to put
   something to start fuzzing. The most simple example is `echo a > input/a`. More sophisticated inputs can
   be created by extracting the bpftrace program from the source code directory (especially from tests
   directory. See the "Input corpus creation" section below for the details.) If some inputs that cause a
   program crash is found, `output/crashes` contains them.
-- bpftrace has known that it has several memory leaks. Therefore `ASAN_OPTIONS=detect_leaks=0` is needed.
-  Otherwise, fuzzer thinks each memory leak as a crash and report it. Also,
-  `abort_on_error=1: symbolize=0` is required for fuzzing.
+- bpftrace has known that it has several memory leaks. Therefore `ASAN_OPTIONS=detect_leaks=0` may also needed.
```
danobi, line 91: I've plugged all the leaks. Any new leaks are push blocking IMO

flake.nix
```diff
@@ -214,6 +214,60 @@
...
+                --replace-fail 'getenv("AFL_PATH")' "(getenv(\"AFL_PATH\") ? getenv(\"AFL_PATH\") : \"$out/lib/afl\")"
+              substituteInPlace src/afl-ld-lto.c \
+                --replace-fail 'LLVM_BINDIR' '"/nixpkgs-patched-does-not-exist"'
+              sed -i 's|LLVM_BINDIR = .*|LLVM_BINDIR = |' utils/aflpp_driver/GNUmakefile
+              substituteInPlace utils/aflpp_driver/GNUmakefile \
+                --replace-fail 'LLVM_BINDIR = ' 'LLVM_BINDIR = ${pkgs.clang_18}/bin/'
+              substituteInPlace GNUmakefile.llvm \
+                --replace-fail "\$(LLVM_BINDIR)/clang" "${pkgs.clang_18}/bin/clang"
+            '';
+          };
```
danobi, line 256: Was there an issue with https://pastes.dxuuu.xyz/x5zobb ?

danobi (commented):
Nice cleanup!

.github/include/ci.py
amscanne: LOL, apparently you should only [use that](https://aflplus.plus/docs/env_variables/) if you are Jakub.

But this sounds fine, I will refresh and plumb `AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES` with a comment explaining it.

docs/fuzzing.md
amscanne: Done.

flake.nix
amscanne: I honestly couldn't remember and was trying to page back in some details. I've taken a pass at this, and it should now be easy to tie together the LLVM versions. It seems to work (let's see if CI passes!).

docs/fuzzing.md
amscanne: Dropped.

docs/fuzzing.md
amscanne: Fixed.

.github/include/ci.py
amscanne: Done, I'm just going to verify that CI passes before resolving.

amscanne: Seems to be working.

flake.nix
amscanne: I'm happy with this now. It's pretty straight-forward but isn't hard-coded to the specific version. (Although that pinning still occurs, it's at the top-level with the single environment definition.)

danobi (approved):
nice!

PR #3849, opened 2025-02-27
codegen: str: Fix off-by-one error for clamped string size

danobi:
In the bpfscript type system, string types have known sizes. The known size is understood by the rest of the type system to include the NUL terminator. The bpftrace.adoc documentation confirms this:

    In case the string is longer than the specified length only `length
    - 1` bytes are copied and a NULL byte is appended at the end.

However for the current code, if the clamped size was provided, codegen injected an extra byte to make room for the NUL terminator.

This is wrong, as it causes C strings that are larger than the clamped size to have its NUL terminator past the end of the bpfscript string. One way this breaks things is that now strcontains() fails to see the NUL terminator during its search, which causes a miss when it otherwise should've been a match. See attached runtime test for example.

Fix by removing the extra byte.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,8 @@ and this project adheres to
   - [#3805](https://github.com/bpftrace/bpftrace/pull/3805)
 - Fix strcontains() correctness bug where matches could be lost if both strings are non-literal
   - [#3811](https://github.com/bpftrace/bpftrace/pull/3811)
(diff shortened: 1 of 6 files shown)


danobi:
Should probably backport this to release branch.

jordalgo (approved):
LGTM - I'm surprised this hasn't come up before.

danobi:
Right. This is kinda bad.

PR #3855, opened 2025-03-01
Invert the `Driver`/`ASTContext` relationship

amscanne:
Stacked PRs:
 * #3869
 * #3868
 * __->__#3855

--- --- ---

### Invert the `Driver`/`ASTContext` relationship

Right now, the global singleton logger keeps all the source context
(filename and source contents), and all logging is resolved with this.

In order to eventually support imports, additional files, etc. it is
necessary to break this hard dependency. This change makes a number of
minor adjustments, relying on the newly introduced diagnostics.

First, the `ASTContext` gains an `ASTSource` member which is effectively
the immutable source filename and contents that will be parsed. It is
useful to store this directly in the context, as it means that the
context carries all the information required to be reset and reparsed
(which is something done after clang parsing is done).

Next, the driver is made to *not* own the context, and the context is
simply passed to a temporary driver. In the future, the driver could be
implemented as a pass over the `ASTContext` containing the source (and
the pass could be done twice: once before clang parsing and once after).
This change greatly simplifies the driver, and will allow it to be used
as part of a standardized pass.

Next, the log plumbing is updated to require references to all the
needed contextual information: filename, contents and location. The
logger supports two modes: one needing this information completely, and
one that emits bare messages.

Finally, all the tests and various ergonomics of these classes are
updated. This is by far the largest contributor to this commit; it is
primarily a boilerplate change. Asuste observers may notice that a lot
of the plumbing and passes in both `main.cpp` and the individual tests
are starting to look similar; this is by design, over time these things
will converge into passes, and the amount of replicated boilerplate can
be greatly reduced.
There are some minor incidental changes that are note worthy as part of
this change:

* Since locations are only defined within the context of a single
  `ASTContext`, when helper information is recorded, those locations are
  translated to a filename, line and column number in the form of the
  `HelperErrorInfo`. This is fine, since the full source context for a
  helper error is unlikely to be useful.

* This *now* applies to both the runtime errors and verifier errors
  related to helpers. The verifier errors now record *all* instances of
  a helper, rather than just the first one, to provide high-quality
  errors.

### DIFF
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_library(ast_defs STATIC
   ast.cpp
   context.cpp
+  int_parser.cpp
   location.cpp
 )
 add_dependencies(ast_defs parser)
@@ -11,7 +12,6 @@ add_library(ast STATIC
   codegen_helper.cpp
   diagnostic.cpp
   dibuilderbpf.cpp
-  int_parser.cpp
   irbuilderbpf.cpp
   pass_manager.cpp
   signal.cpp
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -55,6 +55,7 @@ extern bool bt_verbose;
 extern bool dry_run;
 
 enum class DebugStage {
+  Parse,
   Ast,
   Codegen,
   CodegenOpt,
@@ -64,6 +65,8 @@ enum class DebugStage {
 };
 
 const std::unordered_map<std::string_view, DebugStage> debug_stages = {
+  // clang-format off
+  { "parse", DebugStage::Parse },
   { "ast", DebugStage::Ast },
   { "codegen", DebugStage::Codegen },
   { "codegen-opt", DebugStage::CodegenOpt },
@@ -72,6 +75,7 @@ const std::unordered_map<std::string_view, DebugStage> debug_stages = {
 #endif
   { "libbpf", DebugStage::Libbpf },
   { "verifier", DebugStage::Verifier },
+  // clang-format on
 };
 
 class WildcardException : public std::exception {
@@ -190,11 +194,12 @@ public:
   StructManager structs;
   FunctionRegistry functions;
   std::map<std::string, std::string> macros_;
-  // Map of enum variant_name to (variant_value, enum_name)
+  // Map of enum variant_name to (variant_value, enum_name).
   std::map<std::string, std::tuple<uint64_t, std::string>> enums_;
-  // Map of enum_name to map of variant_value to variant_name
+  // Map of enum_name to map of variant_value to variant_name.
   std::map<std::string, std::map<uint64_t, std::string>> enum_defs_;
-  std::map<libbpf::bpf_func_id, ast::Location> helper_use_loc_;
+  // For each helper, list of all generated call sites.
+  std::map<libbpf::bpf_func_id, std::vector<HelperErrorInfo>> helper_use_loc_;
   const FuncsModulesMap &get_traceable_funcs() const;
   KConfig kconfig;
   std::vector<std::unique_ptr<AttachedProbe>> attached_probes_;
--- a/src/lexer.h
+++ b/src/lexer.h
@@ -1,9 +1,0 @@
-#pragma once
-
-#include "driver.h"
-#include "parser.tab.hh"
-
-#define YY_DECL                                                                \
-  bpftrace::Parser::symbol_type yylex(bpftrace::Driver &driver,                \
-                                      yyscan_t yyscanner)
-YY_DECL;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -13,6 +13,7 @@
 #include <unistd.h>
 
 #include "aot/aot.h"
+#include "ast/attachpoint_parser.h"
 #include "ast/diagnostic.h"
 #include "ast/pass_manager.h"
 #include "ast/passes/codegen_llvm.h"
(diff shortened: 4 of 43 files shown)


tests/runtime/call
```diff
@@ -505,8 +505,8 @@ AFTER TMPDIR=/tmp ./testprogs/syscall open
 NAME path_in_unsupported_fentry
 PROG fentry:vfs_read { print("a"); print(path(args.file->f_path)); print("b"); }
 EXPECT stdin:1:31-60: ERROR: helper bpf_d_path not allowed in probe
-EXPECT fentry:vfs_read { print("a"); print(path(args.file->f_path)); print("b"); }
```
amscanne, line 508: Is this needed? Please revert :)

amscanne: (This is just a note to self for the next iteration, I don't think these lines have changes to no need to drop the `EXPECT` prefix.)

danobi (commented):
High level looks good and makes sense. Only skimmed some of the details. Would be good to split out the boilerplate changes into separate commits in future - that way review can just lean on compiler and tests

src/ast/CMakeLists.txt
```diff
@@ -11,7 +12,6 @@ add_library(ast STATIC
   codegen_helper.cpp
   diagnostic.cpp
   dibuilderbpf.cpp
-  int_parser.cpp
```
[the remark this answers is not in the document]
amscanne: There's some link ordering here that is funky. IMO we should just move `int_parser.cpp` and others into the logical `utils` (and organize that a bit), and then add that as a proper dependency everywhere that needs it. I think I'm being a bit lazy here, so you should probably ask me to fix this first :)

src/lexer.h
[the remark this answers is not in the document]
amscanne: I think it can be dropped. I will update this.

amscanne:
> High level looks good and makes sense. Only skimmed some of the details. Would be good to split out the boilerplate changes into separate commits in future - that way review can just lean on compiler and tests

Honestly, this is the second iteration of this change (or maybe it's third, I can't remember). I'm sure I could do a better job with time, but there was a lot of entanglement around some of these pieces, so I decided just to pull out the logical pieces that were clean and tolerate one more entangled change. I'm with you that it's not ideal.

With the bottom change in, let me see if there's something I can pull out of here and split a bit.

src/ast/CMakeLists.txt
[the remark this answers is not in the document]
amscanne: I started a separate change to break up `utils` and move things into a logical linking order. Bigger job than I thought, so I'm not going to block on it for now.

tests/semantic_analyser.cpp
```diff
@@ -15,238 +16,251 @@ namespace bpftrace::test::semantic_analyser {
...
 {
-  Driver driver(bpftrace);
-  test(bpftrace, has_features, driver, input, -1, expected_error, true, false);
+  return test(bpftrace, has_features, input, -1, expected_error, true, false);
 }
 
-void test_error(std::string_view input,
-                std::string_view expected_error,
-                bool has_features = true)
+ast::ASTContext test_error(const std::string &input,
```
danobi, line 258: Just curious, what was the issue with string_view?

danobi (approved):

danobi:
Needs a rebase

src/main.cpp
```diff
@@ -807,7 +818,12 @@ int main(int argc, char* argv[])
     }
   }
 
-  // Listing probes when there is no program
+  // This is our primary program AST context. Initially it is empty, not
+  // filename set or source file. The way we set it up depends on the mode of
```
jordalgo, line 822:
```suggestion
  // This is our primary program AST context. Initially it is empty i.e. there is no
  // filename set or source file. The way we set it up depends on the mode of
```

src/parser.yy
```diff
@@ -10,7 +10,7 @@
 %define define_location_comparison
 %define parse.assert
 %define parse.trace
-%expect 5
+%expect 4
```
jordalgo, line 13: nice! 1 closer to 0.

src/required_resources.h
```diff
@@ -19,9 +19,39 @@ namespace bpftrace {
 
 class BPFtrace;
 
-struct HelperErrorInfo {
-  int func_id = -1;
-  ast::Location loc;
+class HelperErrorInfo {
+public:
+  // This classe effectively wraps a location, but preserves only the parts
```
jordalgo, line 24:
```suggestion
  // This class effectively wraps a location, but preserves only the parts
```

src/required_resources.h
```diff
@@ -19,9 +19,39 @@ namespace bpftrace {
...
+  // that are needed to emit the error in a useful way. This is because it may
+  // be serialized and used by a separate runtime.
```
jordalgo, line 26: wait what separate runtime?

jordalgo (commented):
This was A LOT of code to review in a single commit. The commit message also seems to indicate several changes so in the future I would absolutely prefer separate commits even if they belong to the same PR (if possible). 

In general it looks good and I'm relying on the tests a bit - just a few minor comment nits and a question or two.

tests/semantic_analyser.cpp
amscanne: Since the `ASTContext` requires a `const std::string&` input, I just made these consistent. If they aren't being modified, then there is no material difference between a `const std::string &` input and a `string_view`, the `string_view` can just be adjusted. (But this wasn't something we were doing anyways.)

src/required_resources.h
amscanne: AOT

src/bpftrace.h
```diff
@@ -64,6 +65,8 @@ enum class DebugStage {
 };
 
 const std::unordered_map<std::string_view, DebugStage> debug_stages = {
+  // clang-format off
```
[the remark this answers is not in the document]
amscanne: See comment above from @danobi and the context, he asked that this was added. It made two per line.

src/main.cpp
```diff
@@ -349,43 +350,42 @@ static void parse_env(BPFtrace& bpftrace)
...
+           BPFtrace& bpftrace,
+           const std::vector<std::string>& include_dirs,
+           const std::vector<std::string>& include_files)
 {
-  Driver driver(bpftrace);
-  driver.source(name, program);
-  int err;
+  Driver driver(ast,
+                bpftrace,
+                /*debug=*/bt_debug.find(DebugStage::Parse) != bt_debug.end());
```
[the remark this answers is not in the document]
amscanne: I will drop this, but it is pretty common to provide a hint with the argument name, e.g.:
* https://github.com/bpftrace/bpftrace/blob/master/src/ast/irbuilderbpf.cpp#L1182
* https://github.com/bpftrace/bpftrace/blob/master/src/dwarf_parser.cpp#L127

src/required_resources.h
jordalgo: That's what I thought - maybe good to be specific but NBD.

src/main.cpp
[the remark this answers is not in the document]
jordalgo: The code itself seems obvious enough and the comment doesn't seem to add anything IMHO.

src/bpftrace.h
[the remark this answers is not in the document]
danobi: Yeah, IMO a single list is easier to visually parse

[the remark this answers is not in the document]
jordalgo: Ah ok, I missed the comment.

src/required_resources.h
amscanne: I try to avoid being too prescriptive, the important bits are that it is an isolated unit which is serialized. (As is the rest of the resources in this file.)

jordalgo: Obviously non-blocking but since it seems you like discussing the color of the bike shed 😄  I think being vague has the downside that developers reading this comment will have the exact same question in their head as I did and right now there isn't a separate runtime aside from AOT, which is barely supported at this point. So I don't find it so much "prescriptive" as "accurate" and "clarifying".

jordalgo (approved):

PR #3856, opened 2025-03-01
fix: add test for result and tweak APIs

amscanne:
As suggested, this implements perfect forwarding. When implementing the test, it was also apparent that collisions with `llvm::handleErrors` might cause some issues, so the signature was changed to allow for the result itself to move directly (and constrained against only these). This also makes the inline handling cleaner.

##### Checklist

### DIFF
--- a/tests/result.cpp
+++ b/tests/result.cpp
@@ -1,0 +1,94 @@
+#include "gtest/gtest.h"
+
+#include "util/result.h"
+
+namespace bpftrace::test::result {
+
+class NotOK : public ErrorInfo<NotOK> {
+public:
+  static char ID;
+  void log(llvm::raw_ostream &OS) const override
+  {
+    OS << "Not OK.";
+  }
+};
+
+class ReallyNotOK : public ErrorInfo<ReallyNotOK> {
+public:
+  static char ID;
+  void log(llvm::raw_ostream &OS) const override
+  {
+    OS << "Really not OK.";
+  }
+};
+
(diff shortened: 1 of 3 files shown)


tests/result.cpp
```diff
@@ -0,0 +1,94 @@
...
+    auto nowOk = handleErrors(std::move(ok), [](const NotOK &) {});
+    EXPECT_TRUE(bool(nowOk)); // Should now be fine.
+  }
+}
+
+TEST(result, handle_err_missing)
+{
+  auto ok = maybeOK(false, true);
+  if (!ok) {
+    auto nowOk = handleErrors(std::move(ok), [](const NotOK &) {});
```
jordalgo, line 83: So the result of this is false because the lambda is accepting a `NotOK` instead of a `ReallyNotOK` right? Just sanity checking myself.

jordalgo (approved):

tests/result.cpp
amscanne: Correct, we are handling `NotOK` but not `ReallyNotOK`.

amscanne: You could provide a second handler (or third, or fourth, etc.) if you wanted to. But I elected *not* provide an API for `handleAllErrors` since I think it's fair to require that no caller is ever *sure* that they handle every error case.

(You can also check it and just print it and ignore it or whatever. But `handleErrors` lets you unpack the actual error information.)

danobi (approved):

PR #3862, opened 2025-03-02
Refactor HelperErrorInfo to include the line & column

amscanne:
Stacked PRs:
 * #3855
 * __->__#3862
 * #3854
 * #3853
 * #3852

--- --- ---

### Refactor HelperErrorInfo to include the line & column

This allows it to be properly serialized, and doesn't significantly
detracted from the way the information is printed. This is being done
because in the future, the `location` will only be interpretable in the
context of a specific `ASTContext` which contains the source. (Which is
necessary to support the possibility of parsing multiple files, etc.)

### DIFF
--- a/src/ast/context.h
+++ b/src/ast/context.h
@@ -3,6 +3,8 @@
 #include <memory>
 #include <vector>
 
+#include "ast/diagnostic.h"
+
 namespace bpftrace {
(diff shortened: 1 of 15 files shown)


amscanne:
This has been rolled into #3855 for simplicity.

PR #3863, opened 2025-03-02
Add Map Declaration syntax

jordalgo:
This gives users the ability to both
specify a BPF map type and the
number of max entries.

Example:
```
let @a = hash(10);
let @b = lruhash(100);

BEGIN {
  @a[1] = "hi";
  @b[1] = "bye";
}
```

The available BPF map types:
- BPF_MAP_TYPE_HASH
- BPF_MAP_TYPE_LRU_HASH
- BPF_MAP_TYPE_PERCPU_HASH
- BPF_MAP_TYPE_PERCPU_ARRAY
- BPF_MAP_TYPE_LRU_PERCPU_HASH

##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -214,6 +214,18 @@ public:
   std::vector<std::string> field;
 };
 
+class MapDeclStatement : public Expression {
+public:
+  explicit MapDeclStatement(Diagnostics &d,
+                            std::string ident,
+                            std::string bpf_type,
+                            int max_entries,
+                            Location &&loc);
+  std::string ident;
+  std::string bpf_type;
+  int max_entries;
+};
+
 class Map : public Expression {
 public:
   explicit Map(Diagnostics &d, std::string ident, Location &&loc);
@@ -315,6 +327,8 @@ public:
 
   Expression *expr = nullptr;
 };
+
+using MapDeclList = std::vector<MapDeclStatement *>;
 
 class VarDeclStatement : public Statement {
 public:
@@ -594,6 +608,7 @@ public:
   Program(Diagnostics &d,
           std::string c_definitions,
           Config *config,
+          MapDeclList &&map_decls,
           SubprogList &&functions,
           ProbeList &&probes,
           Location &&loc);
@@ -602,6 +617,7 @@ public:
   Config *config = nullptr;
   SubprogList functions;
   ProbeList probes;
+  MapDeclList map_decls;
 };
 
 std::string opstr(const Binop &binop);
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -5,6 +5,7 @@
 #include "ast/async_event_types.h"
 #include "ast/codegen_helper.h"
 #include "bpftrace.h"
+#include "log.h"
 #include "struct.h"
 
 namespace bpftrace::ast {
@@ -345,6 +346,18 @@ void ResourceAnalyser::visit(Call &call)
   }
 }
 
+void ResourceAnalyser::visit(MapDeclStatement &decl)
+{
+  Visitor<ResourceAnalyser>::visit(decl);
+
+  auto bpf_type = get_bpf_map_type(decl.bpf_type);
+  if (!bpf_type) {
+    LOG(BUG) << "No bpf type from string: " << decl.bpf_type;
+    return;
+  }
+  map_decls_.insert({ decl.ident, { *bpf_type, decl.max_entries } });
+}
+
 void ResourceAnalyser::visit(Map &map)
 {
   Visitor<ResourceAnalyser>::visit(map);
@@ -476,6 +489,25 @@ void ResourceAnalyser::update_map_info(Map &map)
   auto &map_info = resources_.maps_info[map.ident];
   map_info.value_type = map.type;
   map_info.key_type = map.key_type;
+
+  auto decl = map_decls_.find(map.ident);
+
+  // hist() and lhist() transparently create additional elements in whatever
+  // map they are assigned to. So even if the map looks like it has no keys,
+  // multiple keys are necessary.
+  if (map_info.key_type.IsNoneTy() && !map_info.value_type.IsHistTy() &&
+      !map_info.value_type.IsLhistTy()) {
+    map_info.max_entries = 1;
+    map_info.bpf_type = get_bpf_map_type(map_info.value_type,
+                                         map_info.key_type);
+  } else if (decl != map_decls_.end()) {
+    map_info.bpf_type = decl->second.first;
+    map_info.max_entries = decl->second.second;
+  } else {
+    map_info.bpf_type = get_bpf_map_type(map_info.value_type,
+                                         map_info.key_type);
+    map_info.max_entries = bpftrace_.config_->get(ConfigKeyInt::max_map_keys);
+  }
 }
 
 void ResourceAnalyser::maybe_allocate_map_key_buffer(const Map &map)
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -1720,6 +1720,38 @@ void SemanticAnalyser::validate_map_key(const SizedType &key, Node &node)
   }
 }
 
+void SemanticAnalyser::visit(MapDeclStatement &decl)
+{
+  if (!bpftrace_.config_->get(ConfigKeyBool::unstable_map_decl)) {
+    decl.addError() << "Map declarations are not enabled by default. To enable "
+                       "this unstable feature, set this config flag to 1 "
+                       "e.g. unstable_map_decl=1";
+  }
+
+  const auto bpf_type = get_bpf_map_type(decl.bpf_type);
+  if (!bpf_type) {
+    auto &err = decl.addError();
+    err << "Invalid bpf map type: " << decl.bpf_type;
+    auto &hint = err.addHint();
+    add_bpf_map_types_hint(hint);
+  } else {
+    bpf_map_type_.insert({ decl.ident, *bpf_type });
+
+    if (decl.max_entries != 1 &&
+        *bpf_type == libbpf::BPF_MAP_TYPE_PERCPU_ARRAY) {
+      decl.addError() << "Max entries can only be 1 for map type "
+                      << decl.bpf_type;
+    }
+  }
+
+  if (is_final_pass()) {
+    auto map_key_search_val = map_key_.find(decl.ident);
+    if (map_key_search_val == map_key_.end()) {
+      decl.addWarning() << "Unused map: " << decl.ident;
+    }
+  }
+}
+
 void SemanticAnalyser::visit(Map &map)
 {
   SizedType new_key_type = CreateNone();
@@ -1778,6 +1810,21 @@ void SemanticAnalyser::visit(Map &map)
     map.key_type = map_key_search_val->second;
   } else {
     map.key_type = CreateNone();
+  }
+
+  if (is_final_pass()) {
+    auto found_kind = bpf_map_type_.find(map.ident);
+    if (found_kind != bpf_map_type_.end()) {
+      if (!bpf_map_types_compatible(map.type,
+                                    map.key_type,
+                                    found_kind->second)) {
+        auto map_type = get_bpf_map_type(map.type, map.key_type);
+        map.addError() << "Incompatible map types. Type from declaration: "
+                       << get_bpf_map_type_str(found_kind->second)
+                       << ". Type from value/key type: "
+                       << get_bpf_map_type_str(map_type);
+      }
+    }
   }
 }
 
--- a/src/bpfmap.h
+++ b/src/bpfmap.h
@@ -1,10 +1,13 @@
 #pragma once
 
+#include <optional>
 #include <string>
 #include <string_view>
 
 #include <bpf/libbpf.h>
 #include <linux/bpf.h>
+
+#include "types.h"
 
 namespace libbpf {
 #include "libbpf/bpf.h"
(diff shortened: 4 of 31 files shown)


jordalgo:
If this implementation looks good, I'll put this behind a config flag, add tests and documentation, etc.

src/parser.yy
```diff
@@ -181,8 +183,8 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %%
 
 program:
-                c_definitions config probes_and_subprogs END {
-                    driver.ctx.root = driver.ctx.make_node<ast::Program>($1, $2, std::move($3.second), std::move($3.first), @$);
+                c_definitions config map_decl_list probes_and_subprogs END {
```
amscanne, line 186: Can we allow the map_decls to be interspersed with the probes? Seems like you might want the declaration close to usage if you have multiple.

Optionally, before submission I would just have a bug as a "good first issue" and leave a `FIXME(#bug)` here. You don't need this right away, but it's an obvious future improvement.

src/parser.yy
```diff
@@ -479,6 +481,20 @@ assign_stmt:
...
+                map_decl_list map_decl_stmt ";"      { $$ = std::move($1); $$.push_back($2); }
+        |        %empty                              { $$ = ast::MapDeclList{}; }
+        ;
+
+map_decl_stmt:
+                LET MAP ASSIGN map_initializer { $$ = driver.ctx.make_node<ast::Map>($2, $4, @$); }
+        ;
+
+map_initializer:
+                IDENT LPAREN RPAREN { $$ = driver.ctx.make_node<ast::MapInitializer>($1, @$); }
```
amscanne, line 494: There was some discussion about this, ... why allow the size to be optional? I think we should force the user to select the map size in these functions.

src/ast/ast.h
```diff
@@ -211,10 +211,26 @@ class Offsetof : public Expression {
...
+
+  std::string bpf_type;
+  std::optional<int> max_entries = std::nullopt;
+};
+
 class Map : public Expression {
 public:
   explicit Map(Diagnostics &d, const std::string &ident, location loc);
   Map(Diagnostics &d, const std::string &ident, Expression &expr, location loc);
+  Map(Diagnostics &d,
```
amscanne, line 230: This is *somewhat* a style thing, but I wouldn't parse this using the `Map` node. It doesn't make sense -- the `Map` node is an expression value.

I would just parse as a `MapInitializer` that has the map identity, the map type and the size all as required parameters in string form. The various passes need to look at this MapInitializer specifically anyways, and this doesn't create any ambiguity for other passes (since it is decided *not* a Map expression).

Using the expression also creates a more complexity variant for that node that everything has to handle, e.g. there are effectively three types in one:

```
MapInitializer != nullptr && key_expr == nullptr => the initializer, not an expression
MapInitializer == nullptr && key_expr != nullptr => the `@[...]` expression with a key
MapInitializer == nullptr && key_expr == nullptr => the `@` expression with no key
```

src/ast/passes/resource_analyser.cpp
```diff
@@ -479,6 +479,27 @@ void ResourceAnalyser::update_map_info(Map &map)
   auto &map_info = resources_.maps_info[map.ident];
   map_info.value_type = map.type;
   map_info.key_type = map.key_type;
+
+  auto initializer = dynamic_cast<MapInitializer *>(map.initializer);
+  map_info.max_entries = initializer && initializer->max_entries
+                             ? *initializer->max_entries
+                             : bpftrace_.config_->get(
+                                   ConfigKeyInt::max_map_keys);
```
amscanne, line 487: Per comment elsewhere, I think this new map syntax should be preferred (for non-trivial programs) and should explicitly require the number of keys in order to avoid the foot gun.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1767,7 +1788,11 @@ void SemanticAnalyser::visit(Map &map)
     // If there is no record of any assignment after the first pass
     // then it's safe to say this map is undefined
     if (!is_first_pass()) {
-      map.addError() << "Undefined map: " << map.ident;
+      if (initializer) {
+        map.addError() << "Unused map: " << map.ident;
```
amscanne, line 1792: I think this should be `addWarning`. I don't object to it being error for now, but in the future, such a map declaration could be for the purposes of an external object, etc.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1722,6 +1722,27 @@ void SemanticAnalyser::validate_map_key(const SizedType &key, Node &node)
 
 void SemanticAnalyser::visit(Map &map)
 {
+  auto initializer = dynamic_cast<MapInitializer *>(map.initializer);
+  if (initializer) {
+    auto found = BPF_MAP_TYPES.find(initializer->bpf_type);
+    if (found == BPF_MAP_TYPES.end()) {
+      auto &err = initializer->addError();
+      err << "Invalid bpf map type: " << initializer->bpf_type;
+      std::stringstream hint;
```
amscanne, line 1731: The hint is a stream, intentionally to avoid this boilerplate, remember? :)  Just do:
```
auto &hint = err.addHint();
hint << "Valid map types: ";
....
   hint << bpf_type.first;
   if (num_types != 0) {
      hint << ", ";
   }
```

And eliminate the `err.addHint()` line below.

src/parser.yy
jordalgo: > Can we allow the map_decls to be interspersed with the probes? 

I'd rather not in this iteration (and maybe ever?). I think @danobi and @JakeHillion feel similarly - only for the reason that it's confusing in how it's not really influencing control flow.

src/ast/ast.h
jordalgo: I had this originally but realized I would have to keep a map of ident to BPF map type in semantic analyser, which, for whatever reason, I thought was more complicated. I'll revisit it though.

src/ast/passes/semantic_analyser.cpp
jordalgo: > such a map declaration could be for the purposes of an external object

Yeah i forgot about that. I'm fine with making it a warning.

src/ast/passes/semantic_analyser.cpp
jordalgo: I forget so quickly! Will fix.

src/parser.yy
amscanne: I'm personally of the somewhat strong opinion that we should not allow this to be optional in the declarations. If we end up going back later and relaxing it (allow no argument) then it's low friction. However, you don't want to end up with a bunch of `lruhashmap()` statements because the other way is higher friction.

My opinion is primarily based on the idea that users should be forced to make an intentional choice for each map. The default is a foot gun not just because the default size is perhaps wrong (which the warning goes a ways to address), but also because **it tricks users into thinking that the map is like an unlimited C++ `std::map` (or name your hash table of choice)**. Given that you already have implicit maps for the one-off case, so I think it's reasonable that explicit maps need this knob.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1715,6 +1715,46 @@ void SemanticAnalyser::validate_map_key(const SizedType &key, Node &node)
   }
 }
 
+void SemanticAnalyser::visit(MapDeclStatement &decl)
+{
+  if (!bpftrace_.config_->get(ConfigKeyBool::unstable_map_decl)) {
+    decl.addError() << "Map declarations are not enabled by default. To enable "
+                       "this unstable feature, set this config flag to true "
+                       "e.g. unstable_map_decl=true";
```
amscanne, line 1723: Nit: maybe the second sentence here is actually a `HINT`?

tests/codegen/map_declarations.cpp
```diff
@@ -0,0 +1,29 @@
+#include "common.h"
+
+namespace bpftrace::test::codegen {
+
+TEST(codegen, map_declarations)
```
amscanne, line 5: Maybe break into separate tests to make catching strange issues a bit easier? E.g. if *one* of the map types change, then something is probably up.

tests/runtime/variable
```diff
@@ -135,3 +135,35 @@ EXPECT (c, 0)
...
+EXPECT @a[2]: 2
+
+NAME map declaration percpuhash
+PROG let @a = percpuhash(1); BEGIN { @a[1] = count(); @a[2] = count(); exit(); }
+EXPECT_REGEX .*WARNING: Map full; can't update element.*
+EXPECT @a[1]: 1
+
+NAME map declaration percpuhash default max entries
+PROG let @a = percpuhash(); BEGIN { @a[1] = count(); @a[2] = count(); exit(); }
+EXPECT_REGEX_NONE .*WARNING: Map full; can't update element.*
```
amscanne, line 162: I still object to the functionality and the test highlights the reason :)

Now this test implicitly depends on the default size configuration, for example. We know in practice that it won't be less than 3, but this would theoretically bite us in exactly the same way it would bite a user -- they'll just find out their code is broken one day while they are trying to debug something (and they are lucky if they discover the problem quickly). No one should have to "discover" that there's a limit because that will never be a good experience -- it should be apparent and explicit up front

amscanne (approved):
Notwithstanding the `max_entries` debate ;)

src/parser.yy
jordalgo: > it tricks users into thinking that the map is like an unlimited C++ std::map

I think this is a bit extreme. We're not intentionally hiding this limitation from users; it's in the documentation and they are free to set the global max entries with a config variable. I think it's less of a footgun now with the warning.
That being said, and as I mentioned, I DO NOT feel strongly but I am curious what the other maintainers think.

src/ast/passes/semantic_analyser.cpp
jordalgo: Hmm... It feels like all one message to me.

amscanne: We don't do this and it doesn't matter to me, but the reason I'm calling out the distinction is that often you have an error, and then a richer diagnostic that has a recommended fix that you can apply. (Like we apply the `clang-tidy` fixes, which are even more structured.) This is definitely fitting that pattern, which is why it's already two separate sentences. I would do it just to set a good example for others since everything gets cargo-culted, and this structure generally makes this more testable and stable (because the tests don't need to overfit on the hint text!), e.g.

```
auto &err = decl.addError();
err << "Map declarations are not enabled by default.";
err.addHint() << "To enable this feature, set the config flag unstable_map_decl=true.";
```

jordalgo: This is another hill I don't wish to die on but IMHO this isn't really an "error" or an indication that the user did something wrong but rather a notification that the feature they are using requires an explicit opt-in. The error message doesn't exist without the information on how to opt-in, which is why it isn't a hint.

I think we have plenty of examples in the code base of proper `addHint` usage.

tests/runtime/variable
danobi, line 1: new file for explicit maps?

src/parser.yy
jordalgo: This was resolved offline in office hours. It was agreed that we'll start by requiring users to specify max entries for map declarations and then follow-up with users for feedback.

jordalgo:
I think this is ready to go for a final review @viktormalik @amscanne @danobi

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1715,6 +1715,46 @@ void SemanticAnalyser::validate_map_key(const SizedType &key, Node &node)
...
+    bpf_map_type_.insert({ decl.ident, found->second });
+
+    if (decl.max_entries != 1 &&
+        found->second == libbpf::BPF_MAP_TYPE_PERCPU_ARRAY) {
+      decl.addError() << "Max entries can only be 1 for map type "
+                      << decl.bpf_type;
+    }
+  }
+
+  if (!is_first_pass()) {
```
amscanne, line 1750: Since warnings won't block progress, I think maybe this should be `is_final_pass` instead of `!is_first_pass()`? Otherwise, I think this warning could be added multiple times.

src/bpfmap.h
```diff
@@ -14,6 +17,23 @@ namespace libbpf {
 
 namespace bpftrace {
 
+static const std::unordered_map<std::string, libbpf::bpf_map_type>
```
amscanne, line 20: Nit: Unless these are used in the header, I would place these definitions inside the implementation file.. If they are used in multiple compilation units, I would place the declaration here but the actual definition (non-static) in a reasonable home.

src/main.cpp
```diff
@@ -781,8 +781,12 @@ static ast::ASTContext buildListProgram(const std::string& search)
   auto ap = ast.make_node<ast::AttachPoint>(search, true, location());
   auto probe = ast.make_node<ast::Probe>(
       ast::AttachPointList({ ap }), nullptr, nullptr, location());
-  ast.root = ast.make_node<ast::Program>(
-      "", nullptr, ast::SubprogList(), ast::ProbeList({ probe }), location());
+  ast.root = ast.make_node<ast::Program>("",
+                                         nullptr,
+                                         std::vector<ast::MapDeclStatement*>{},
```
amscanne, line 786: Did you want to define a using alias for this type for consistent? E.g. `MapDeclList`? Then here you just have `ast::MapDeclList()`, which I think is cleaner.

src/output.cpp
```diff
@@ -218,7 +218,9 @@ std::string Output::get_helper_error_msg(int func_id, int retcode) const
 {
   std::string msg;
   if (func_id == libbpf::BPF_FUNC_map_update_elem && retcode == -E2BIG) {
-    msg = "Map full; can't update element. Try increasing max_map_keys config";
+    msg = "Map full; can't update element. Try increasing max_map_keys config "
```
amscanne, line 221: *cough* Hint *cough*

src/required_resources.h
```diff
@@ -83,13 +88,22 @@ struct MapInfo {
   std::optional<LinearHistogramArgs> lhist_args;
   std::optional<int> hist_bits_arg;
   int id = -1;
+  int max_entries = -1;
+  bool has_decl_bpf_type = false;
```
amscanne, line 92: What is the significance of this? It is not serialized. I would argue that this class should hold the info for serialization purposes only, so it's a bit of a flag that there's something here that's not being serialized. (Maybe this is an accidental omission, or maybe the class is being used for multiple purposes)

amscanne (approved):

src/output.cpp
jordalgo: Let me refactor this a little in a follow-up PR. I'll have to pass a tuple/pair/struct up to the caller.

src/required_resources.h
jordalgo: Leftover from an older version. Will remove. Good catch.

src/parser.yy
JakeHillion: Ah, I missed this again, sorry! I think my initial implementation did interesperse and I do think that's the slightly more friendly option. As @amscanne says, you might have a map only used in two probes and want to declare it close. But I don't have any issue with this implementation because it's really easy to extend to that behaviour in the future without breaking backwards compatibility.

jordalgo: > it's really easy to extend to that behaviour in the future without breaking backwards compatibility

Agreed and we can wait for feedback from users if this is something they find necessary.

jordalgo:
Going to merge this tomorrow if I don't hear any complaints.

PR #3867, opened 2025-03-03
ast: add `Location` to wrap parser location

amscanne:
ast: add `Location` to wrap parser location

In the future, we will no longer rely on the global singleton to contain
the filename and contents, since we want to support multiple files. This
change introduces an `ASTSource` class which can be used to contain this
metadata, and wraps a reference to this in a new `Location` class.

For now, the existing filename and contents are used by the `LogSink`,
but this will change very shortly.

One related note with this change is the requirement that `Location` be
passed to nodes as a r-value. This is a minor change, but means that
calls to `make_node` internally must explicitly copy or construct a new
location, which makes it clear what is happening. In general, the
`Location` type is cheap and can be carried by value.


src/ast/context.h
```diff
@@ -49,8 +72,22 @@ class ASTContext {
   Program *root = nullptr;
 
 private:
+  // wrap potentially converts external types to internal ones. At the moment,
+  // this automatically converts the parser `location` to the `Location` class
+  // bound to the current source file.
+  template <typename T>
+  auto wrap(T &&t) -> decltype(t)
```
jordalgo, line 79: JC, what else do we intend on wrapping?

amscanne: Nothing at the moment, but this is also the simplest way to automatically decay all the regular cases to the same type.

danobi (approved):

PR #3868, opened 2025-03-03
ast: restructure the pass manager

amscanne:
Stacked PRs:
 * #3888
 * #3885
 * #3869
 * __->__#3868

--- --- ---

### ast: restructure the pass manager

Currently the pass manager does little other than run a sequence of
functions in order. This has led to a few problems:

1) The `BPFtrace` object has logically morphed in a kind of "god object",
   where passes set and mutate global state, which is difficult to
   encapsulate, track or mock.

2) Because of (1), tests for individual passes depend on the implicit
   behavior from all previous passes. Therefore, the same pass plumbing
   is replicated in many places.

Finding myself needing to stash some data resulting from passes
(deserialized bitcode from imports), I decided to take a pass at this
problem rather than adding another field to the AST.

This change introduces a pass context and pass manager which operate
using a `PassContext`. Each pass can require some set of known inputs,
which will be automatically extracted from the `PassContext`.
Furthermore, each pass may produce some known output, which is
automatically injected into to the `PassContext` when the pass is
finished running.

Attempting to add a pass which does not have the required inputs
produced by an earlier pass will generate an error at add time, not at
runtime. Attempting to set an output which has already been set by a
previous pass will also generate an error.

The core idea for this change is that the intermediate state required
(probe attachment info, resources, etc.) will be manifest as these
objects which live in the context. Known passes can produce these
objects, which will then be consumed subsequently.

This makes it easier to design tests around specific passes: rather than
needing to recreate the known universe, you need only manually inject
the specific bits of context that you care about, and validate that the
intended outputs are successfully produced.

Note that this change still leaves the AST and all other intermediate
values as mutable. It would be good to allow for incremental
transformations there as well that could be statically validated, but
this is a much broader task.  Implementing a simple static storage for
passes provides an opportunity to refactor and simplify a lot of the
existing plumbing.


amscanne:
I have a patch pending in the stack which will split out `ASTContext` from the `PassManager`, because lots of the later passes do not depend on the AST. Will be updated shortly, but it should simplify this patch slightly.

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
-  PassFPtr fn_;
+  template <typename Return, typename... Inputs>
+  static Pass createFn(const std::string &name,
+                       std::function<Result<Return>(Inputs &...)> fn)
+  {
+    return createImpl(name, fn);
+  }
+  template <typename Return, typename... Inputs>
+  static Pass createFn(const std::string &name,
+                       std::function<Return(Inputs...)> fn)
```
danobi, line 241: Can't do `Inputs &...` like above? Would be better to do that if possible than to have the static_assert? Or maybe static_assert is better error, as otherwise it'll be some crazy template mismatch error

danobi:
Quick thoughts before I disappear for the day:
* The use of template metaprogramming worries me in terms of us lay persons understanding and modifying it. However, it seems to be contained enough where the big hammer of a rewrite isn't out of the question (assuming you are hit by a bus or something). I'm sure I can fully understand it given enough time.
* It's actually not super crazy now that I'm looking at it. Seems like the trickiest bits are the parts dealing with lambdas. Which I can mostly nod through

src/ast/pass_manager.cpp
```diff
@@ -1,47 +1,75 @@
...
-#include "bpftrace.h"
+#include "ast/context.h"
+#include "log.h"
+#include "util/result.h"
 
 namespace bpftrace::ast {
 
-namespace {
-void print(PassContext &ctx, const std::string &name, std::ostream &out)
+std::atomic<int> PassContext::next_type_id_;
```
jordalgo, line 8: JC, do we have multiple threads accessing this or is it just for future proofing?

src/ast/pass_manager.cpp
```diff
@@ -1,47 +1,75 @@
...
-    // If the number of errors is non-zero, stop and return.
-    if (!ctx.ast_ctx.diagnostics().ok()) {
-      ctx.ast_ctx.diagnostics().emit(out_);
-      return 1;
+  // Check that the inputs are all available.
+  for (const int type_id : pass.inputs()) {
+    if (!outputs_.contains(type_id)) {
+      auto type_name = PassContext::lookup_type(type_id);
+      LOG(BUG) << "Pass " << pass.name() << " requires output " << type_name
+               << ", which is not available.";
```
jordalgo, line 35: WDYT?
```suggestion
               << ", which is not available. Prior passes need to provide this type as an output";
```

src/ast/pass_manager.cpp
```diff
@@ -1,47 +1,75 @@
...
+  int type_id = TypeId<ASTContext>::type_id();
+  if (state_.contains(type_id)) {
+    return get<ASTContext>().diagnostics().ok();
+  }
+  return true;
 }
-} // namespace
 
-void PassManager::AddPass(Pass p)
+void PassContext::fail(int type_id)
```
jordalgo, line 20: this fail seems a bit specific - what about re-naming to something like `no_object_fail` or just inlining this logic where it's called.

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
+    type_names_[type_id] = std::move(type_name);
+    return type_id;
+  }
+
+  // Returns the underlying type for this type_id. Note that the type must
+  // exist, or this will have undefined behavior. This is called exclusively
+  // by the pass manager, when statically checking pass types.
+  static const std::string &lookup_type(int type_id)
+  {
+    return type_names_[type_id];
```
jordalgo, line 127: Does this really justify it's own function and comment?

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
+class State : public PassContext::State {
+public:
+  static std::string type_name()
+  {
+    return name.str();
+  }
+};
 
-// Base pass
+// Pass is a arbitrary pass.
```
jordalgo, line 169:
```suggestion
// Pass is an arbitrary pass.
```

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
+//
+// It is permitted to produce a single output, but may consume any number of
+// inputs.
```
jordalgo, line 172:
```suggestion
// It is permitted to produce a single output, but may consume multiple inputs.
```

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
-  Pass() = delete;
-  Pass(std::string name, PassFPtr fn)
-      : fn_(std::move(fn)), name(std::move(name)) {};
+  Pass(const Pass &) = delete;
+  Pass(Pass &&) = default;
+  Pass &operator=(const Pass &) = delete;
+  Pass &operator=(Pass &&) = default;
 
-  virtual ~Pass() = default;
+  // create creates the given pass.
```
jordalgo, line 193: I would have assumed that `create` destroys 😛

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
+  static Pass create(const std::string &name, Func func)
+  {
+    std::function fn(func);
+    return createFn(name, fn);
+  }
 
-  void Run(PassContext &ctx)
+  // run executes the pass.
+  //
+  // Before we reach this point, the pass manager will have statically verified
```
jordalgo, line 212: You said this above - any need to repeat it here?

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
+  {
+  }
+
+  std::string name_;
+  std::vector<int> inputs_;
+  std::vector<int> outputs_;
+  Fn fn_;
 };
 
+// PassManager hold passes in sequence.
```
jordalgo, line 318:
```suggestion
// PassManager holds passes in sequence.
```

src/ast/pass_manager.h
```diff
@@ -1,64 +1,371 @@
...
+//
+// The PassManager is responsible for verifying the inputs and outputs of the
```
jordalgo, line 320: This is now said a third time - maybe just a summary at the top of PassManager would be better.

jordalgo:
> The use of template metaprogramming worries me in terms of us lay persons understanding and modifying it.

I'm also scared of too much meta programming. However, and maybe because of the in-person review @amscanne provided at Scale, I'm not as worried here. That being said, I think it all comes down to when things don't compile and I get cryptic messages that I can't resolve easily - then in a fit of rage I may curse all this stuff and attempt to rip it out. As @danobi suggested - how much would we lose by having a static list of passes and how much less meta-programming code would there be?

All that aside, these changes look pretty good to me and aside from some nits above, this is pretty much ready to go.

src/ast/pass_manager.h
amscanne: Yeah, I wound up doing this to provide a cleaner error message. I'm sure that I could use concepts and constraints in some complicated way but I found the `static_assert` lets you spell out what's wrong.

Some sample errors:

If you don't pass a reference:
```
In file included from /home/amscanne/bpftrace/tests/pass_manager.cpp:1:
/home/amscanne/bpftrace/src/ast/pass_manager.h: In instantiation of 'static bpftrace::ast::Pass bpftrace::ast::Pass::createFn(const std::string&, std::function<_Res(_ArgTypes ...)>) [with Return = llvm::Expected<bpftrace::test::passes::Test3Output>; Inputs = {bpftrace::test::passes::Test2Output}; std::string = std::__cxx11::basic_string<char>]':
/home/amscanne/bpftrace/src/ast/pass_manager.h:197:20:   required from 'static bpftrace::ast::Pass bpftrace::ast::Pass::create(const std::string&, Func) [with Func = bpftrace::test::passes::CreateTest3Pass()::<lambda(bpftrace::test::passes::Test2Output)>; std::string = std::__cxx11::basic_string<char>]'
  197 |     return createFn(name, fn);
      |            ~~~~~~~~^~~~~~~~~~
/home/amscanne/bpftrace/tests/pass_manager.cpp:56:22:   required from here
   56 |   return Pass::create("test3", [](Test2Output) -> Result<Test3Output> {
      |          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   57 |     return Test3Output();
      |     ~~~~~~~~~~~~~~~~~~~~~
   58 |   });
      |   ~~                  
/home/amscanne/bpftrace/src/ast/pass_manager.h:234:25: error: static assertion failed: all arguments must be references
  234 |     static_assert((std::is_reference_v<Inputs> && ...),
      |                    ~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

If you pass a type that isn't an `ast::State` derived type:
```
In file included from /home/amscanne/bpftrace/tests/pass_manager.cpp:1:
/home/amscanne/bpftrace/src/ast/pass_manager.h: In instantiation of 'static int bpftrace::ast::PassContext::TypeId<T>::type_id() [with T = bpftrace::test::passes::Foo]':
/home/amscanne/bpftrace/src/ast/pass_manager.h:254:69:   required from 'static bpftrace::ast::Pass bpftrace::ast::Pass::createImpl(const std::string&, std::function<llvm::Expected<T>(Inputs& ...)>) [with Return = bpftrace::test::passes::Test3Output; Inputs = {bpftrace::test::passes::Foo}; std::string = std::__cxx11::basic_string<char>]'
  254 |     std::vector<int> inputs = { PassContext::TypeId<Inputs>::type_id()... };
      |                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
/home/amscanne/bpftrace/src/ast/pass_manager.h:227:22:   required from 'static bpftrace::ast::Pass bpftrace::ast::Pass::createFn(const std::string&, std::function<llvm::Expected<T>(Inputs& ...)>) [with Return = bpftrace::test::passes::Test3Output; Inputs = {bpftrace::test::passes::Foo}; std::string = std::__cxx11::basic_string<char>]'
  227 |     return createImpl(name, fn);
      |            ~~~~~~~~~~^~~~~~~~~~
/home/amscanne/bpftrace/src/ast/pass_manager.h:198:20:   required from 'static bpftrace::ast::Pass bpftrace::ast::Pass::create(const std::string&, Func) [with Func = bpftrace::test::passes::CreateTest3Pass()::<lambda(bpftrace::test::passes::Foo&)>; std::string = std::__cxx11::basic_string<char>]'
  198 |     return createFn(name, fn);
      |            ~~~~~~~~^~~~~~~~~~
/home/amscanne/bpftrace/tests/pass_manager.cpp:58:22:   required from here
   58 |   return Pass::create("test3", [](Foo&) -> Result<Test3Output> {
      |          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   59 |     return Test3Output();
      |     ~~~~~~~~~~~~~~~~~~~~~
   60 |   });
      |   ~~                  
/home/amscanne/bpftrace/src/ast/pass_manager.h:53:26: error: static assertion failed: type must be derived from State<...>
   53 |       static_assert(std::is_base_of_v<State, T>);
      |                     ~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

If you try to return an unregistered type:
```
In file included from /home/amscanne/bpftrace/tests/pass_manager.cpp:1:
/home/amscanne/bpftrace/src/ast/pass_manager.h: In instantiation of 'static bpftrace::ast::Pass bpftrace::ast::Pass::createImpl(const std::string&, std::function<llvm::Expected<T>(Inputs& ...)>) [with Return = bpftrace::test::passes::Foo; Inputs = {}; std::string = std::__cxx11::basic_string<char>]':
/home/amscanne/bpftrace/src/ast/pass_manager.h:247:24:   required from 'static bpftrace::ast::Pass bpftrace::ast::Pass::createFn(const std::string&, std::function<_Res(_ArgTypes ...)>) [with Return = bpftrace::test::passes::Foo; Inputs = {}; std::string = std::__cxx11::basic_string<char>]'
  247 |       return createImpl(name, wrap);
      |              ~~~~~~~~~~^~~~~~~~~~~~
/home/amscanne/bpftrace/src/ast/pass_manager.h:198:20:   required from 'static bpftrace::ast::Pass bpftrace::ast::Pass::create(const std::string&, Func) [with Func = bpftrace::test::passes::CreateTest3Pass()::<lambda()>; std::string = std::__cxx11::basic_string<char>]'
  198 |     return createFn(name, fn);
      |            ~~~~~~~~^~~~~~~~~~
/home/amscanne/bpftrace/tests/pass_manager.cpp:58:22:   required from here
   58 |   return Pass::create("test3", []() -> Foo {
      |          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
   59 |     return Foo(); // Test3Output();
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   60 |   });
      |   ~~                  
/home/amscanne/bpftrace/src/ast/pass_manager.h:277:40: error: static assertion failed: return value must be derived from State<...>
  277 |                     static_assert(std::is_base_of_v<PassContext::State, Return>);
      |                                   ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/ast/pass_manager.cpp
amscanne: Static initializers have historically been fraught. I know that you have thread-safe guarantees for the *same* static initializer (multiple threads hitting the same static initializer, magically one goes and others block). You *may* have hard guarantees around concurrency for *different* static initializers, but I just prefer to be defensive for this case.

src/ast/pass_manager.cpp
amscanne: I've tweaked this message, PTAL.

src/ast/pass_manager.cpp
amscanne: Done.

src/ast/pass_manager.h
amscanne: Dropped.

src/ast/pass_manager.h
amscanne: Done.

src/ast/pass_manager.h
amscanne: I've tweaked this comment to be more precise regarding inputs and outputs. PTAL.

src/ast/pass_manager.h
amscanne: I've clarified that this is the intended constructor.

src/ast/pass_manager.h
amscanne: Done.

src/ast/pass_manager.h
amscanne: Dropped two.

amscanne:
I don't think there's really any meta-programming here. It's mostly a few partial specializations and the one `TypeId` trick.

I'm not really sure that an enum is any cleaner. Do you mean something like this?

```
enum StateType {
   BpfTrace,
   AstContext,
   FooBar,
   ...
 }
 
 class PassContext {
    ...
    template <typename T>
    T& get(StateType id) {
       return *static_cast<T*>(...);
    }
 };
 ```
 
 This seems to have a lot of consistency issues and redundancy issues, e.g. you need the type and ID to correspond (`ctx.get<BPFtrace>(StateType::BpfTrace)`).
 
 You could make it so the context just have a bunch of objects directly, e.g.
 
 ```
 class PassContext {
    BPFtrace& getBpftrace() {
       return bpftrace_wrapper_.get();
    }
    std::optional<std::reference_wrapper<BPFTrace>> bpftrace_wrapper_;
  };
  ```
  
But then you just have an everything-is-about-everything class again. Really the point of this mechanism is to allow for loose couplings, e.g. you get to change some intermediate data w/o having to worry about a bunch of plumbing internals.
 
 I'm open to any other way, I'd just want a little more clarity on what you mean by a static list of passes.

**Comment by jordalgo** [comment]
> I don't think there's really any meta-programming here. 

C'mon now. There is even a named technique called Template Metaprogramming, which is what is used here. You also work at Meta so it's double!

> I'm not really sure that an enum is any cleaner.

It might not be. I haven't thought about it too deeply but the thought wasn't about it being cleaner necessarily just less magical (more explicit) for folks not as familiar with the techniques being used here. Like I said, now that we've gone over it a bit in person and I reviewed this code I'm fine with things as is. 

I know @danobi said he was going to be busy this week but I'll still give him half a day (I'm so generous) to review before merging.

**Review by jordalgo — APPROVED** [review]

**Comment by amscanne** [comment]
> C'mon now. There is even a named technique called Template Metaprogramming, which is what is used here. You also work at Meta so it's double!

Touché. I just mean that I don't think of it as meta programming because that's more computation or code generation through templates. I think I managed to avoid a lot of that.

> It might not be. I haven't thought about it too deeply but the thought wasn't about it being cleaner necessarily just less magical (more explicit) for folks not as familiar with the techniques being used here. Like I said, now that we've gone over it a bit in person and I reviewed this code I'm fine with things as is.
> 
> I know @danobi said he was going to be busy this week but I'll still give him half a day (I'm so generous) to review before merging.

I don't want to foist anything, so I'm open to exploring other avenues. I recognize that templates can be used in good taste or bad. I tend to think it's best when it's compartmentalized to some core functionality, but then it enables a broad set of patterns (which aren't directly loaded with magic). This is what I tried to do here. I did a pass at this last night to simplify (dropped all the `TypeName` stuff, for example), but ultimately felt that could hurt debuggability.

In terms of how this all manifests, I think it's best to look ahead at the [codegen passes](https://github.com/bpftrace/bpftrace/pull/3885/files#diff-28926c090177e6efbfb5a7449833e0e863bb226822abb513b7dfc6c388837d87R4524) to evaluate if it's worth it.

File: src/ast/pass_manager.h [hunk 12]
```diff
@@ -1,64 +1,366 @@
...
+  Pass(const Pass &) = delete;
+  Pass(Pass &&) = default;
+  Pass &operator=(const Pass &) = delete;
+  Pass &operator=(Pass &&) = default;
 
-  virtual ~Pass() = default;
+  // create is the standard Pass constructor.
+  //
+  // The passed function may accept an arbitrary set of inputs. Note that the
+  // `type_identity_t` basically makes it so this template will always match,
```
danobi, line 188: type_identity_t? I don't see any further references to it

src/ast/pass_manager.h
```diff
@@ -1,64 +1,366 @@
...
+  // and the types are resolved later.
+  //
+  // For convenience, this template is set up to match against lambdas, and the
+  // partial specialization that is used below will only work with lambdas.
```
danobi, line 192: Not sure I got the "only work with lambdas" bit. Specifically which parts?

src/ast/pass_manager.h
```diff
@@ -1,64 +1,366 @@
...
+  // run all registered passes.
+  //
+  // This is a canonical example of how to use `foreach` above, in many other
+  // cases a direct call will be suitable.
+  Result<PassContext> run()
+  {
+    PassContext ctx;
+    auto err = foreach([&](auto &pass) -> Result<> {
+      if (!ctx.ok())
+        return OK(); // Skip the pass.
```
danobi, line 347:
```suggestion
        return OK(); // Previous errors diagnostics, skip the pass.
```

danobi (commented):
nits but lgtm otherwise

danobi (approved):

viktormalik:
> In terms of how this all manifests, I think it's best to look ahead at the [codegen passes](https://github.com/bpftrace/bpftrace/pull/3885/files#diff-28926c090177e6efbfb5a7449833e0e863bb226822abb513b7dfc6c388837d87R4524) to evaluate if it's worth it.

I like how this allows to split these large passes we have into a number of smaller, interconnected passes while also being able to check that the dependencies among them are contained.

I wish it could be done without the template magic under the hood but I'm ok to trade off the complexity in the core part for simplicity to use, which seems to be the case here.

src/ast/pass_manager.h
```diff
@@ -1,64 +1,366 @@
...
+      return static_cast<T &>(extern_it->second.get());
+    }
+    no_object_failure(type_id);
+  }
+
+private:
+  // for the failed lookup path, see above.
+  [[noreturn]] static void no_object_failure(int type_id);
+
+  // put emplaces the the value into the map; it must not exist already. In
```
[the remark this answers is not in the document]
amscanne: Whoops, fixed.

src/ast/pass_manager.h
amscanne: Yeah, this got rewritten a few times for sure. The `type_identity_t` was the original name for `TypeId` I think :) 

Anyways, I've simplified this documentation and fixed the below as well. Should now make sense.

src/ast/pass_manager.h
amscanne: Fixed. It now explains this well I think. It's just "lambdas" encouraged (since otherwise you're doing extra copying)

src/ast/pass_manager.h
amscanne: Fixed. I added a richer explanation to the actual method documentation, and just a note on the if here.

PR #3869, opened 2025-03-04
ast: use new PassManager when possible

amscanne:
Stacked PRs:
 * #3888
 * #3885
 * __->__#3869

--- --- ---

### ast: use new PassManager when possible

This change introduces trivial pass wrappers for the `Driver`,
`TracepointFormatParser`, `AttachpointParser` and `BTF` parser. These
are then collected into a function `AllParsePasses` which has a
canonical order for these passes used by `main`.

The passes are then used across the board by the various tests and
`main` itself, in order to take the first step towards standardizing
this plumbing.

### DIFF
--- a/src/ast/attachpoint_parser.cpp
+++ b/src/ast/attachpoint_parser.cpp
@@ -780,4 +780,12 @@ AttachPointParser::State AttachPointParser::raw_tracepoint_parser()
   return OK;
 }
 
+Pass CreateParseAttachpointsPass()
+{
+  return Pass::create("attachpoints", [](ASTContext &ast, BPFtrace &b) {
+    AttachPointParser ap_parser(ast, b, false);
+    ap_parser.parse();
+  });
+}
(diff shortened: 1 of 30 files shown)


danobi (commented):
Didn't look very closely at details but this is very nice.

viktormalik (approved):
Great cleanup!

It would be nice to unify and simplify the lists of passes in different test suites but I assume that it will be done in follow-up PRs.

amscanne:
> > BTW, please don't click merge on this specific PR until the base PR is merged and it is rebased. (You can see that it is a PR to the other branch below, since it depends on those changes.)
> 
> Why is the 'Mergable' check green?

I'm having trouble with it. I don't know why the GitHub base ref is wrong. I think it doesn't get rerun when the PR gets reparented. I'll take a look at it later this week, and perhaps just delete it if it's too finicky.

PR #3870, opened 2025-03-04
Fix size mismatch for `stack_key`s

ttreyer:
When printing a stack (either `ustack` or `kstack`), we pass a stack key through the ring buffer and `bpftrace` can retrieve the stack trace using it.

This stack key structure is defined twice, with different sizes:
1. as a `bpftrace` SizedType, which is 16 bytes and account for padding,
2. as a `llvm` StructType, which is 12 bytes and without padding. This lead to generating a `memcpy` of 16 bytes, instead of 12, triggering a verifier error for accessing the unitialized padding.

This patch fixes the issue by explicitly adding the padding to the `u|kstack_key` to fix the size mismatch.
We can't fix the issue by changing `src/types.c SizedType::CreateStack()` as the padding needs to be accounted for in case the stack is part of a tuple.

Somehow, the problem appears on Kernel 5.15, but not when I try on 5.19.
There must have been a change in the eBPF verifier between those two versions?

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -161,12 +161,14 @@ StructType *IRBuilderBPF::GetStackStructType(bool is_ustack)
       getInt32Ty(), // nr_stack_frames
       getInt32Ty(), // pid
       getInt32Ty(), // probe id
+      getInt32Ty(), // __padding, see #3870
     };
     return GetStructType("ustack_key", elements, false);
   } else {
     std::vector<llvm::Type *> elements{
       getInt64Ty(), // stack id
       getInt32Ty(), // nr_stack_frames
+      getInt32Ty(), // __padding, see #3870
     };
     return GetStructType("kstack_key", elements, false);
   }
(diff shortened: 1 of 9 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -161,12 +161,14 @@ StructType *IRBuilderBPF::GetStackStructType(bool is_ustack)
       getInt32Ty(), // nr_stack_frames
       getInt32Ty(), // pid
       getInt32Ty(), // probe id
+      getInt32Ty(), // __padding
```
amscanne, line 164: Just leave an issue # pointing to this PR? Or some other breadcrumb?

@jordalgo I'm pretty sure we just talked about sneaky bits from the "type system mismatch" very recently, do you remember the bug and/or if there's a canonical bug for this generally?

jordalgo: I don't think one was filed as this is only happening on earlier kernels but yeah a comment/breadcrumb would be good.

I think this bug was introduced here:
https://github.com/bpftrace/bpftrace/pull/3769

jordalgo:
@ttreyer Looks like you need to update the codegen

src/ast/irbuilderbpf.cpp
```diff
@@ -161,12 +161,14 @@ StructType *IRBuilderBPF::GetStackStructType(bool is_ustack)
...
     };
     return GetStructType("ustack_key", elements, false);
   } else {
     std::vector<llvm::Type *> elements{
       getInt64Ty(), // stack id
       getInt32Ty(), // nr_stack_frames
+      getInt32Ty(), // __padding
```
danobi, line 171: Instead of inserting padding, how about increasing the size of `nr_stack_frames` to i64?

amscanne (approved):
LGTM

PR #3871, opened 2025-03-04
mips: use offset in `pt_regs` as `pc_offset`

amscanne:
Instead of finding the current PC, compute the expected PC offset into `pt_regs`, which is what this function is actually supposed to do.

##### Checklist

- ~[ ] The new behaviour is covered by tests~ -- is there a recommended way to structure arch tests?

### DIFF
--- a/src/arch/mips64.cpp
+++ b/src/arch/mips64.cpp
@@ -47,7 +47,7 @@ static std::array<std::string, 32> registers = {
 };
 
 // Alternative register names that match struct pt_regs
-static std::array<std::string, 32> ptrace_registers = {
+static std::array<std::string, 38> ptrace_registers = {
   "regs[0]",
   "regs[1]",
   "regs[2]",
@@ -80,6 +80,15 @@ static std::array<std::string, 32> ptrace_registers = {
   "regs[29]",
   "regs[30]",
   "regs[31]",
+  // This layout supports only MIP64, which does not have the option
+  // `CONFIG_CPU_HAS_SMARTMIPS`. See the `pt_regs` defintion for MIP64 [1].
+  // [1] https://github.com/torvalds/linux/blob/848e076317446f9c663771ddec142d7c2eb4cb43/arch/mips/include/asm/ptrace.h#L28
+  "cp0_status";
+  "hi",
+  "lo",
+  "cp0_badvaddr",
+  "cp0_cause",
+  "cp0_epc",
 };
 
 static std::array<std::string, 8> arg_registers = {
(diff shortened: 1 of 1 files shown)


amscanne:
I'd like to add some tests here, but I think it's a general question for the arch bits. Just aiming for the point fix on the MIPS `pc_offset`.

jordalgo (commented):
I endorse this diff but want another maintainer to stamp.

src/arch/mips64.cpp
```diff
@@ -80,6 +80,12 @@ static std::array<std::string, 32> ptrace_registers = {
   "regs[29]",
   "regs[30]",
   "regs[31]",
+  "lo",
+  "hi",
+  "epc",
+  "badvaddr",
+  "status",
+  "cause",
```
danobi, line 88: The internet says they're supposed to have "cp0_" prefix. Should we keep that?

amscanne: Welp, I've made this change, but I am currently suspecting that that the whole arch/registers mechanism is broken anyways. (Not I'm not sure the implications in practice, or for other architectures.)

This layout is for [user space](https://github.com/torvalds/linux/blob/48a5eed9ad584315c30ed35204510536235ce402/arch/mips/include/uapi/asm/ptrace.h#L34) `pt_regs`, and the offsets are valid and stable. But to the kernel, these are `user_pt_regs`, and there's an extra step of serialization (for ptrace and other similar APIs). The [actual](https://github.com/torvalds/linux/blob/48a5eed9ad584315c30ed35204510536235ce402/arch/mips/include/uapi/asm/ptrace.h#L34) `pt_regs` can depend on config options, etc.

I haven't dug in, but I would assume that `pt_regs` in BPF is actually the kernel `pt_regs`, and that the stability for references there needs to come from BTF annotations. We should actually be dynamically resolving as the suitable `pt_regs` type in the kernel rather than relying on a `regs("rax")` option (e.g. it would look like `regs.rax` and, with time, needs to encode the appropriate CO-RE annotations).

My preference would be to merge this with these names as is (since it is changing something obviously broken, to something that possibly still broken but now it's probably broken in the same way as everything else), and I will file a separate follow-up issue.

amscanne: Okay, so for `amd64` and `aarch64`, they are effectively the same layout and not dependent on kernel configurations. So the offsets produced should be correct, although I don't think they can strictly *rely* on a fixed offset (and should be resolved against BTF for `pt_regs`).

This one is a bit broken, and there is simply no way to make it truly correct without having the actual `pt_reg` type information, barring reading the kernel config here. It's because of this thing in the middle of the struct definition, *before* the `epc`.

```
#ifdef CONFIG_CPU_HAS_SMARTMIPS
	unsigned long acx;
#endif
```

Let me file an issue for this, add a note here. I will make the offset dependent on whatever the default is for that config, and the `ip` will *sometimes* work for MIPS :) (And we can work to broadly fix types)

amscanne: I don't believe `CONFIG_CPU_HAS_SMARTMIPS` is defined for MIPS64, so the order and names should now be "correct". Long term fix is to map to `pt_regs` and rely on BTF to resolve offsets correctly.

amscanne:
Filed #3873 and this is ready to merge.

PR #3872, opened 2025-03-04
nix: Move blazesym rev to flake.lock

danobi:
Only updating the rev in flake.nix and *not also* running `nix flake lock --update-input $THE_INPUT` causes nix to use a stale revision, at least until the old revision is evicted from cache and you get an error.

In the meantime, you simply get old results without any warning.

Make this less likely to happen again for blazesym by moving the rev into flake.lock. This forces the updater to use the correct tooling.

Also update nix.md with some helpful examples.

##### Checklist

### DIFF
--- a/docs/nix.md
+++ b/docs/nix.md
@@ -107,6 +107,37 @@ dxu@kashmir bpftrace]$ CC=afl-clang-fast CXX=afl-clang-fast++ cmake -B build-fuz
 
 This section has a few examples on how to interact with the Nix configuration.
 
+### Update flake inputs
+
+Flakes have external inputs in the `inputs = { ... }` section of `flake.nix`.
+
+To update a single input:
+```
+$ nix flake lock --update-input blazesym
+warning: updating lock file '/home/dxu/dev/bpftrace/flake.lock':
+• Updated input 'blazesym':
+    'github:libbpf/blazesym/6beb39ebc8e3a604c7b483951c85c831c1bbe0d1' (2025-02-14)
+  → 'github:libbpf/blazesym/285b17f15a12885544b21f1ae352928910656767' (2025-03-04)
+```
+
+To update a single input to a specific revision:
+```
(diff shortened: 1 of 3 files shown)


docs/nix.md
```diff
@@ -107,6 +107,37 @@ dxu@kashmir bpftrace]$ CC=afl-clang-fast CXX=afl-clang-fast++ cmake -B build-fuz
 
 This section has a few examples on how to interact with the Nix configuration.
 
+### Update flake inputs
+
+Flakes have external inputs in the `inputs = { ... }` section of `flake.nix`.
+
+To update a single input:
+```
+$ nix flake lock --update-input blazesym
```
danielocfb, line 116: This warns me with:
```
> warning: '--update-input' is a deprecated alias for 'flake update' and will be removed in a future version.
```

danielocfb (approved):
Tried it out and it works fine, thanks! It even updated `narHash` on its own..

docs/nix.md
danobi: Interesting. Can't seem to find an equivalent, at least not in the help page. Tried a few things:

```
dxu@kashmir~/d/bpftrace (blazesym_update) $ nix flake update blazesym
error: cannot find flake 'flake:blazesym' in the flake registries
dxu@kashmir~/d/bpftrace (blazesym_update) $ nix flake update --update-input blazesym
error: unrecognised flag '--update-input'
Try 'nix --help' for more information.
dxu@kashmir~/d/bpftrace (blazesym_update) $ nix flake update --input blazesym
error: unrecognised flag '--input'

```

danielocfb: Weird. I suppose you may have to update `nix` itself?

For me the following works (in the sense that it doesn't error out), but produces the same warning:
```
$ nix flake update --update-input blazesym
warning: '--update-input' is a deprecated alias for 'flake update' and will be removed in a future version.
```
...
Anyway, no idea.

danobi: Yeah, I've got nix 2.17 which is a bit old. Since it looks like a new thing, maybe let's leave it as is for now, so it works for the most people.

jordalgo (approved):

PR #3874, opened 2025-03-06
clang-tidy all the things

amscanne:
This commit does the following, primarily for headers, which were not
previously checked:

* Concatenate nested namespaces (modernize-concat-nested-namespaces)
* Drop void from `foo(void)` (modernize-redundant-void-arg)
* Remove any unused includes (misc-include-cleaner)
* Use `override` instead of `virtual` (modernize-use-override)
* Try to ensure a consistent include order:
  => All "system" libraries (which we include llvm, and bbc)
  => All "internal" libraries (which includes gtest, gmock and others)
* Remove unused variables and ensure dead code is blocked by
  `#if` appropriately, to prevent compiler warnings
* Use `DISABLED_` in favor of `GTEST_SKIP()` if the test is skipped
  100% of the time; this removes the dead code warnings
* Remove `std::move` that prevents temporary copy elision

The `.clang-tidy` configuration is adjusted to check headers.

##### Checklist

### DIFF
--- a/src/util/result.cpp
+++ b/src/util/result.cpp
@@ -1,3 +1,6 @@
+
+#include <llvm/Support/raw_os_ostream.h>
+
 #include "util/result.h"
 
 namespace bpftrace {
--- a/tests/field_analyser.cpp
+++ b/tests/field_analyser.cpp
@@ -1,5 +1,6 @@
 #include "ast/passes/field_analyser.h"
 #include "driver.h"
+#include "dwarf_common.h"
 #include "mocks.h"
 #include "gtest/gtest.h"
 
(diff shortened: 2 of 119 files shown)


danobi:
Update `.clang-tidy` to get CI to prevent backslides?

amscanne:
> Update `.clang-tidy` to get CI to prevent backslides?
> 
> I also presume you found `scripts/clang_tidy.sh`

LOL, no I didn't notice it. I got VSCode set up with clang-tidy (which I guess ignores .clang-tidy?) and decided to fix all the warnings.

But I think this file is actually not working properly right now, since these checks are already there, and do not result in errors. I'm investing this and will include here.

amscanne:
I think that the checks weren't previously enabled for headers, which explains why most of the fixes here are in headers. This is included in this change.

danobi (approved):
Let’s add some of the rules you listed in the pr description to the .clang-tidy. I can do that tmr if you dont beat me to it

tests/field_analyser.cpp
```diff
@@ -702,10 +699,9 @@ TEST_F(field_analyser_dwarf, parse_inheritance_multi)
   EXPECT_EQ(cls->GetField("rabc").offset, 24);
 }
 
-TEST_F(field_analyser_dwarf, parse_struct_anonymous_fields)
+// Disable because anonymous fields not supported #3084.
+TEST_F(field_analyser_dwarf, DISALBED_parse_struct_anonymous_fields)
```
danobi, line 703:
```suggestion
TEST_F(field_analyser_dwarf, DISABLED_parse_struct_anonymous_fields)
```

src/util/result.cpp
```diff
@@ -1,3 +1,6 @@
+
```
danobi, line 1:
```suggestion
```

PR #3875, opened 2025-03-06
Fix calls to ConfigSetter constructor

viktormalik:
#3797 changed the constructor of `ConfigSetter` to accept a reference to `BPFtrace` rather than a pointer but failed to update three calls to the constructor in tests which are only executed if `HAVE_LIBLLDB` is set.

Fix the problematic calls.

##### Checklist


jordalgo:
Good catch.

jordalgo (approved):

PR #3879, opened 2025-03-07
Even more clang-tidy!

danobi:
See commits. Even more tidiness! 

##### Checklist


danobi:
The codeql warning is legit - there are indeed codepaths where we use env vars as a format string. But it's pre-existing and requires an attacker to control env vars for root user to abuse. At that point attacker should just hijack $PATH? Low severity IMO

src/bpffeature.h
```diff
@@ -58,14 +58,12 @@ class BPFfeature;
...
-  BPFnofeature() : kprobe_multi_(false), uprobe_multi_(false)
-  {
-  }
+  BPFnofeature() = default;
   int parse(const char* optarg);
 
 protected:
-  bool kprobe_multi_;
-  bool uprobe_multi_;
+  bool kprobe_multi_{ false };
```
jordalgo, line 65: yay for fixing potentially uninitialized member vars

jordalgo (approved):

PR #3880, opened 2025-03-09
Remove padding from stack struct

jordalgo:
Just make the number of frames an int64.
Issue: https://github.com/bpftrace/bpftrace/issues/3876

##### Checklist


jordalgo:
cc @ttreyer @amscanne

tests/codegen/llvm/builtin_kstack.ll
```diff
@@ -56,11 +56,12 @@ lookup_stack_scratch_merge:                       ; preds = %entry
 get_stack_success:                                ; preds = %lookup_stack_scratch_merge
   %2 = udiv i32 %get_stack, 8
   %3 = getelementptr %kstack_key, ptr %stack_key, i64 0, i32 1
-  store i32 %2, ptr %3, align 4
-  %4 = trunc i32 %2 to i8
-  %murmur_hash_2 = call i64 @murmur_hash_2(ptr %lookup_stack_scratch_map, i8 %4, i64 1)
-  %5 = getelementptr %kstack_key, ptr %stack_key, i64 0, i32 0
-  store i64 %murmur_hash_2, ptr %5, align 8
+  %4 = zext i32 %2 to i64
```
danobi, line 59: Seems like an unnecessary zero extension here. IRBuilderBPF::CreateGetStack() is defined to return an i32 but in reality the underlying get_stack() helper call returns a long. So I think the prototype in CreateGetStack() is wrong.

jordalgo: `long` is 32 bits on some systems - how do we decide if it should be 32 vs 64? It seems like the `size` param is 32 and the return value is "...buf length equal to or less than size on success"

danobi: Hmm, that's an interesting point. I'm not sure what happens if we assume i64 on a 32-bit system. But seems like the current behavior of assuming 32-bit on a 64-bit system is only accidentally correct (assuming I am even thinking about it right).

Maybe the most robust solution is to look at llvm target triple? Maybe best documented in a ticket and handled later.

jordalgo: It seems there are two instances where we treat a returned long as 32 bits (this one included) and 13 instances where we treat it as 64 bits. I'll file a ticket and look into it.

amscanne: Even if the helper is returning a long and/or 32-bits, the `zext` here is just for the value to be stored in the structure (In other words, this is an implicit cast). Isn't this the number of stack frames? No disagreement that there are lots of type consistency issues, just wondering if it matters for this particular case.

jordalgo: Yeah this is the number of stack frames which became 64 bits in this change to avoid the arbitrary padding in the struct.

jordalgo:
If there are no objections, I'm going to merge this. Then add this and @ttreyer's fix onto the release branch.

jordalgo:
Confirmed this worked on 5.15 kernel `vmtest -k $(nix build --print-out-paths .#kernel-5_15)/bzImage -- ./build/src/bpftrace -e "'BEGIN { print(kstack); exit(); }'"`

Issue #18, opened 2018-08-31
scripts should accept positional parameters


brendangregg:
Fix #17 beforehand.

We should be able to provide parameters to the script:

```
# ./src/bpftrace fsslower.bt 10 file1
# ./fsslower.bt 10 file1
```

They should be available as the variables `$1`, `$2`, etc.

I see this could be processed in at least one of two ways:

### A) type detection

In this case, "10" would available as the integer `$1`, and "file1" would be available as the string `$2`. Integers can be identified as those that match the regexp `/^[0-9]+$/`

If `$1` and `$2` (etc) are accessed and were not supplied on the command line, they should default to 0 or "", depending on their use as an integer or a string. That may be tricky to implement. So,

### B) integers

All positional parameters, `$1`, `$2`, etc, are provided as integer types (uint64_t). Strings are provided as the pointer to the string. If they are not provided, they are zero.

In the script with the invocation example above, we can then use:

- `$1` to refer to "10"
- `str($2)` to refer to "file1"

I think this (B) approach is easier and probably more sane.

brendangregg:
Added priority flag, as people keep asking. I think option (B) is the way to go.

PR #3883, opened 2025-03-11
global vars: don't always use an aggregate as initializer

fvdlfvdl:
Global variables are always initialized using the method ConstantAggregateZero::get. But, they are not always compound types (array, vector, struct), like e.g.
num_cpus.

With assertions enabled in LLVM, this will lead to the following failure whenever num_cpus is used (which is always for certain types of maps):

assertion failed at
llvm-project/llvm/lib/IR/Constants.cpp:1674 in static ConstantAggregateZero *llvm::ConstantAggregateZero::get(Type *): (Ty->isStructTy() || Ty->isArrayTy() || Ty->isVectorTy()) && "Cannot create an aggregate zero of non-aggregate type!"
*** Check failure stack trace: ***
    @     0x560b06286229  absl::log_internal::LogMessage::SendToLog()
    @     0x560b0628619e  absl::log_internal::LogMessage::Flush()
    @     0x560b06233d84  __assert_fail
    @     0x560b07d8fb08  llvm::ConstantAggregateZero::get()
    @     0x560b02928a53
bpftrace::ast::CodegenLLVM::generate_global_vars()
    @     0x560b02926a4c  bpftrace::ast::CodegenLLVM::generate_ir()
    @     0x560b027c4360  main
    @     0x7faf97e8d3d4  __libc_start_main
    @     0x560b027bc02a  _start

Fix this by using Constant::getNullValue instead.

##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc` (N/A)
- [ ] User-visible and non-trivial changes updated in `CHANGELOG.md`. (N/A)
- [ ] The new behaviour is covered by tests    (N/A)


danobi:
lgtm. But please update codegen tests: https://github.com/bpftrace/bpftrace/blob/master/tests/README.md#updating

fvdlfvdl:
Hm, the LLVM #18 test failure shows that the expected output in the .ll files for tests should all be changed from:

```externally_initialized constant i64 zeroinitializer```

to

```externally_initialized constant i64 0```

..since it appears LLVM prints 'zeroinitializer' for aggregate values, but just 0, obviously, for others.

Not sure why this only is flagged on LLVM 18, though.

jordalgo:
@fvdlfvdl - Trying to understand why our CI debug builds didn't catch this issue. Can you share your repro steps?

jordalgo:
>  I think you need an assert-enabled LLVM library build to trigger this. Without it, you'll probably won't see it.

I was thinking this too. Since we're pulling LLVM from nix it's probably not enabled by default.

Anyway, thanks for the fix!

PR #3884, opened 2025-03-11
Use blazesym for user space address symbolization

danielocfb:
Commit fdf50acb2f2a ("Use blazesym for kernel address symbolization matters") introduced the means for using the `blazesym` library (as opposed to `bcc`) for the symbolization of kernel addresses. With this change we add the necessary logic for also using `blazesym` for the symbolization of user space addresses. Code structure is kept similar, with different paths taken for `blazesym` vs. `bcc`. The intention is that eventually `bcc` functionality will be removed and `blazesym` be made the default.

##### Checklist


danielocfb:
Based on my understanding of what bcc does, there may be some differences in functionality. For example, it seems to support lookup of debug information via symfs (https://github.com/iovisor/bcc/blob/82f9d1cb633aa3b4ebcbbc5d8b809f48d3dfa222/src/cc/bcc_elf.c#L721-L722. `blazesym` currently doesn't have this support. bcc also seems to support VDSO symbolization, which is still work in progress for `blazesym` (let me know if that's a blocker).

On the other hand, `blazesym` supports symbolization of BPF kernel addresses (this is actually more relevant to https://github.com/bpftrace/bpftrace/commit/fdf50acb2f2a4185cb2f8e2be736d4bc36894f7c, but I figured I bring it up). I don't know whether that has any relevance for `bpftrace` specifically, though.

Also, I wanted to give a heads up that I'll be out the entirety of April. I don't know if you folks intend to cut a new release in this time frame. Personally, I'd probably exclude `blazesym` support if that were the case, or make it explicitly opt-in. It's not clear to me whether the usage of `Nix` to get this support is guardrail enough, but leave that up to you to decide.

danielocfb:
Still looking at the one test failure.

danielocfb:
> Still looking at the one test failure.

Seems to be an issue with pivot root handling.

danielocfb:
> > Also, I wanted to give a heads up that I'll be out the entirety of April. I don't know if you folks intend to cut a new release in this time frame.
> 
> We already cut a release branch and it will go out March 25th as per our [new schedule](https://github.com/bpftrace/bpftrace/blob/master/docs/release_process.md). I agree that we shouldn't enable blazesym for that release but definitely the next one. I also think we can enable it internally at Meta to get some usage before the next external release.

That sounds good to me.

Also, still working on addressing the last test failure of the test requiring a slightly different output than what `blazesym` currently produces.

danielocfb:
Updated to most recent `blazesym` commit, which includes https://github.com/libbpf/blazesym/pull/1063, which addresses the previous test failure.

jordalgo (commented):
Minor nits but overall this looks good.

Curious, is it possible to add some additional symbolication tests that shows the advantage of using blazesym over bcc or is it mostly about perf/speed?

d-e-s-o:
> Curious, is it possible to add some additional symbolication tests that shows the advantage of using blazesym over bcc or is it mostly about perf/speed?

Performance is certainly one aspect, though I haven't done any performance evalation. I think the main benefit down the line will be that `blazesym` supports DWARF and so we can get source code location information. However, this PR has not yet hooked that up. It is focused on getting on-par in terms of functionality.

jordalgo:
@danielocfb Mind rebasing off master and re-pushing?

src/usyms.cpp
```diff
@@ -2,12 +2,47 @@
...
+#endif
+
 #include "config.h"
 #include "scopeguard.h"
 #include "usyms.h"
 #include "util/symbols.h"
 #include "util/system.h"
 
+namespace {
+std::string stringify_addr(uint64_t addr, bool show_module)
```
jordalgo, line 16: looks like this can go inside the `HAVE_BLAZESYM` ifdef

src/usyms.h
```diff
@@ -37,7 +37,30 @@ class Usyms {
...
+                          bool show_offset,
+                          bool show_module);
   struct bcc_symbol_option& get_symbol_opts();
+
+#ifdef HAVE_BLAZESYM
+  struct blaze_symbolizer* symbolizer_{ nullptr };
+
+  struct blaze_symbolizer* create_symbolizer() const;
+  void cache_blazesym(const std::string& elf_file);
+  std::optional<std::string> resolve_blazesym_int(uint64_t addr,
```
jordalgo, line 53: I know the name is the same in ksyms but maybe this is slightly better `resolve_blazesym_inst` ? Non blocking

src/usyms.cpp
```diff
@@ -47,11 +87,79 @@ void Usyms::cache(const std::string &elf_file)
       pid_sym_[pid] = bcc_symcache_new(pid, &get_symbol_opts());
 }
 
-std::string Usyms::resolve(uint64_t addr,
-                           int32_t pid,
-                           const std::string &pid_exe,
-                           bool show_offset,
-                           bool show_module)
+#ifdef HAVE_BLAZESYM
+struct blaze_symbolizer *Usyms::create_symbolizer() const
```
jordalgo, line 91: Can this be extracted out into a helper used by both this and ksyms?

jordalgo (commented):
Looking good but needs another rebase 😢 

For next steps are we going to enable blazesym by default and remove the ifdefs? I think we can add some runtime tests at this point for both symbolizer types.

src/usyms.cpp
```diff
@@ -47,8 +87,71 @@ void Usyms::cache_bcc(const std::string &elf_file)
...
+    return;
+
+  if (symbolizer_ == nullptr) {
+    symbolizer_ = create_symbolizer();
+    if (symbolizer_ == nullptr)
+      return;
+  }
+
+  // preload symbol table for executable to make it available even if the
+  // binary is not present at symbol resolution time
```
danobi, line 117: "binary is not present" meaning binary is deleted or process is not running?

src/usyms.h
danielocfb: No idea what "inst" signifies but I suppose you will say the same thing about "int". So sure, why not.

src/usyms.cpp
danielocfb: It could. Where do you suggest it would live, though?

src/usyms.cpp
```diff
@@ -142,12 +245,107 @@ std::string Usyms::resolve_bcc(uint64_t addr,
...
+
+        const blaze_sym *sym = &syms->syms[0];
+        if (sym->name != nullptr)
+          return stringify_sym(sym, addr, show_offset, show_module);
+      }
+    }
+  }
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
```
danobi, line 298: Since there's a bunch of these in this file, maybe just do it once at the top?

danielocfb:
> For next steps are we going to enable blazesym by default and remove the ifdefs?

My understand is that it will already be used once this pull request gets merged, right? If so, then I think we should leave the option in place to switch back for a while.

> I think we can add some runtime tests at this point for both symbolizer types.

Can you elaborate some more, please? Doesn't `bpftrace` already have tests for the symbolization functionality?

danobi (commented):
nice!

src/usyms.cpp
danielocfb: It's just a copy of the same blurb from the bcc path -- I anticipated its upcoming removal and figured we'd loose it otherwise. Anyway, it refers to the binary no longer being available on the file system, i.e., having been deleted.

danobi:
> > For next steps are we going to enable blazesym by default and remove the ifdefs?
> 
> My understanding is that it will already be used once this pull request gets merged, right? If so, then I think we should leave the option in place to switch back for a while (i.e., keep the ifdefs around).

It'll be used by default in CI and probably by anyone using nix environment for development.

I think what Jordan means to say is we should start requiring blazesym for the build. Which I agree with. We can still support bcc for a bit (potentially 1 more release? tbd) in case there are any show stoppers in production.

The upside of requiring blazesym for the build is it'll get the distro packaging movement started. And we're more likely to shake out issues if we force everyone developing bpftrace to use it. If there's a show stopper for blazesym packaging we can put the ifdefs back in - it's a simple revert.

src/usyms.h
jordalgo: Hah - I thought "int" was short for "instruction" was it not?

src/usyms.cpp
jordalgo: Could be in a new file or there is a symbols.h helper. Just an idea - this isn't a blocker.

jordalgo:
Yes @danobi correctly translated my ramblings.

> Yes, that's fine. Let's make this contingent on a stable release of blazesym-c first, though. I will work on that this week.

Sure, that works for me.

src/usyms.h
d-e-s-o: I typically use it to mean "internal". As in, an implementation detail not visible outside the module.

src/usyms.cpp
d-e-s-o: I just noticed that it actually can't be moved. The point of this method was that it accesses the `Usym` type's `config_` value, meaning that it needs to be defined within this type's purview.

src/usyms.h
jordalgo: Ah, the convention I'm used to for such things is "impl" e.g. `resolve_blazesym_impl` but NBD.

d-e-s-o: Okay, will rename then.

jordalgo: @danielocfb Sorry to be a pain but can we at least make it consistent with ksyms? I'm fine changing it back to "int" or "impl" or whatever.

d-e-s-o: Yep, done.

jordalgo (approved):

PR #3885, opened 2025-03-11
codegen: break into separate passes

amscanne:
Stacked PRs:
 * #3888
 * __->__#3885

--- --- ---

### codegen: break into separate passes

This change uses the new pass manager to break the codegen phases into
many separate passes.


danobi (approved):

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2917,6 +2917,11 @@ ScopedExpr CodegenLLVM::visit(Probe &probe)
                                               { b_.getPtrTy() }, // ctx
                                               false);
 
+  // Skip if we've generated too many.
+  uint64_t max_bpf_progs = bpftrace_.config_->get(ConfigKeyInt::max_bpf_progs);
+  if (probe_count_ > max_bpf_progs)
```
jordalgo, line 2922: Why was this added here and why don't we want the same error as below?

src/ast/passes/codegen_llvm.h
```diff
@@ -363,4 +353,67 @@ class CodegenLLVM : public Visitor<CodegenLLVM, ScopedExpr> {
...
+// by the Link pass below, and can be dumped.
+Pass CreateObjectPass();
+
+// DumpASM may be used to dump the produced binary.
+Pass CreateDumpASMPass(std::ostream &out);
+
+// Link produces the final output `BpfBytecode` object.
+Pass CreateLinkPass();
+
+// AllCompilePasses returns a `MultiPass` representing all compile passes, in
```
jordalgo, line 403: nit: there is some minor inconsistency in the comments here as some are prefixed with part of the function name e.g. "DumpASM" and some are not. I tend to think we don't need to repeat the name if it's right above the function.

tests/bpfbytecode.cpp
```diff
@@ -20,11 +18,11 @@ BpfBytecode codegen(const std::string &input)
...
                 .add(ast::AllParsePasses())
                 .add(ast::CreateSemanticPass())
+                .add(ast::AllCompilePasses())
                 .run();
   EXPECT_TRUE(ok && ast.diagnostics().ok());
-
-  ast::CodegenLLVM codegen(ast, *bpftrace);
-  return codegen.compile();
+  auto &output = ok->get<BpfBytecode>();
+  return std::move(output);
```
jordalgo, line 25:
```suggestion
  return ok->get<BpfBytecode>();
```
Why does this not work? The return type doesn't include a reference.

tests/runtime/basic
```diff
@@ -277,7 +277,9 @@ WILL_FAIL
 
 NAME kaddr fails
 PROG BEGIN { print(kaddr("asdfzzzzzzz")) }
-EXPECT ERROR: Failed to compile: Failed to resolve kernel symbol: asdfzzzzzzz
+EXPECT stdin:1:9-35: ERROR: Failed to resolve kernel symbol: asdfzzzzzzz
+       BEGIN { print(kaddr("asdfzzzzzzz")) }
```
jordalgo, line 281: niiiiicceeee

jordalgo (commented):
Minor nits and a few questions but otherwise LGTM

src/ast/passes/codegen_llvm.cpp
amscanne: The error below will have already been emitted. This was previously relying on an exception being thrown which goes all the way to main. Now the pass will complete, and actually put a diagnostic on the probe where we encountered the error.

tests/bpfbytecode.cpp
amscanne: I actually want to move out the value. (Essentially we leave an empty `BpfBytecode` inside the context to be destroyed.)

I could have a `std::move(ok->get<BpfBytecode>())` instead, but for some reason I thought that was less clear? Should I change this?

You definitely don't want to return a reference here, since that will be deleted once the context goes out of scope.

amscanne: Oh, I think I see your question now. I think what you're saying *would* be fine, if `BpfBytecode` had a suitable copy constructor. Unfortunately that is deleted (https://github.com/bpftrace/bpftrace/blob/master/src/bpfbytecode.h#L29), presumably to avoid accidental expensive copies.

src/ast/passes/codegen_llvm.h
amscanne: I'm open to whatever, I would just like to be consistent.

I'd like to think that *one day*, we could clean up comments to the extent that documentation gets generated. In that case, it's nice to have full sentences. I typically used the name because docs should be regular, grammatically correct sentences, and they need a subject, e.g. `Returns a multi-pass representing...` is not a sentence :)

Honestly, I think I was doing `Link blah blah` in some weird way to appease your desire to avoid repetition, but still sneak in a full sentence 😆 . I think I just need to make them consistent, e.g. maybe the compile pass becomes "The compile pass ...." ?

src/ast/passes/codegen_llvm.cpp
jordalgo: Maybe a comment here would be good but non-blocking.

src/ast/passes/codegen_llvm.h
jordalgo: Personally, I think sentence fragments are fine but if you want full sentences, just include the whole function name for now to be consistent - at least in this file.

tests/bpfbytecode.cpp
jordalgo: something something copy elision if you just do `ok->get<BpfBytecode>();` ? I need to read into this more.

amscanne: That only applies to temporary local values. You definitely won't be able to avoid copy/move elision in this case, since you're moving it out of the context.

tests/runtime/basic
amscanne: You get it for free!

src/ast/passes/codegen_llvm.h
amscanne: I stuck with fragments here and made them consistent. I don't want to get hung up on this, but it's nice to have something uniform at some point. (Having automated developer docs in the future would be a good forcing function for that.)

src/ast/passes/codegen_llvm.cpp
amscanne: Added clarifying comment.

jordalgo:
A rebase from master will fix that flaky runtime test ;)

viktormalik:
> A rebase from master will fix that flaky runtime test 😉

`kretprobe_order` will need fixing, too :grimacing:

jordalgo:
oh haha, I thought that one was fixed but the title was just wrong on this one: https://github.com/bpftrace/bpftrace/pull/3890

amscanne:
> oh haha, I thought that one was fixed but the title was just wrong on this one: #3890

I think I just fixed the wrong thing. Didn't realize that there were several there and got confused (it was late). Ah well, they should all have the same fix now.

jordalgo (approved):

PR #3887, opened 2025-03-12
test: Don't require a single instance of `bpftrace_test`

amscanne:
This updates the util tests to avoid failing when more than one instance of `bpftrace_test` (or whatever test binary) happens to be running at the same time. Instead, it merely requires that *at least* one instance of this test is running is suitable checks that the current process is one of them.

This enables the use of tools such as `gtest-parallel` to easily run the suite in parallel, taking a time of several minutes down to just a few seconds:

```
$ time gtest-parallel tests/bpftrace_test
[825/825] semantic_analyser.struct_member_keywords (9426 ms)

real	0m9.548s
user	4m21.008s
sys	4m35.737s
```

##### Checklist

### DIFF
--- a/tests/utils.cpp
+++ b/tests/utils.cpp
@@ -16,6 +16,7 @@
 #include "util/symbols.h"
 #include "util/system.h"
 #include "util/wildcard.h"
+#include "gmock/gmock-matchers.h"
 #include "gtest/gtest.h"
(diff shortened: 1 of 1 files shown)


danobi (approved):
nice; parallel is a huge win

Issue #3827, opened 2025-02-26
RFC: implement structured passes


amscanne:
There are a few things within `bpftrace` that I'd like to improve (hopefully they are viewed as improvements), and this proposal lays out (in general terms) what steps I plan to take. While this is not a feature, it is generally broad enough that it warrants an RFC. I don't intend to make any breaking changes, and this will likely be a longer-term effort.

First, the issues that I'm aiming to improve:
* While there was an attempt at some point to standardize passes, the pass plumbing is used minimally. Most tests and other places manually instantiate the most important passes (ResourceAnalyser, FieldAnalyser, SemanticAnalyser, Codegen) and the `PassManager` is used only in a limited capacity. It's not clear that the `PassResult` and other types are providing value at the moment given their very limited use.
* The passes have no way to produce or consume intermediate outputs, therefore these generally get stuffed into the `BPFtrace` object directly or somewhere in the AST (not representing any actual AST state, but representing some intermediate state from some pass). This limits proper encapsulation, and makes reasoning about passes difficult because they are defined by their side-effects. The same set of passes are copied in different tests (not always clear what's needed and what's not needed).
* Since there's no way to represent intermediate state, tests simply pass all passes and then check against either mocked side-effects in the `BPFtrace` object, or check for the full, formatted program output (e.g. errors). Ideally, one would run a well-defined pass with inputs and outputs.

In order to hopefully improve on these things, I'm planning to do the following and would like feedback. Note that these points are not laid out strictly in linear order, as some from the second bullet may precede some from the first bullet.
* Simplify the pass infrastructure, as follows:
  - [ ] Introduce "diagnostics" that can be added directly to the AST. Then passes can attach errors directly to the nodes, instead of each pass having to do its own error stream (or e.g. emitting errors directly to stderr). This standardizes how passes can produce results/a known class of side-effects (diagnostics).
  - [ ] Add a general-purpose `ErrorOr`-style class, and use this instead of the `PassResult` plumbing for a pass. This also relates to #3731, since this should be used elsewhere over time.
  - [ ] Enable a new pass manager, which will automatically validate and provide any inputs required for a pass function (e.g. if the function needs `BPFtrace&, FooBar&`, then the pass manager will validate that previous passes have produced those outputs when the pass is registered), and automatically register any outputs from the pass function (e.g. if the function returns `Config`, then any future pass that requires `Config&` will have this parameter provided). This will make it simpler to split intermediate state into separate objects and reduce associated boilerplate.
[...]

PR #3888, opened 2025-03-12
passes: Move implementations into implementation files

amscanne:
Stacked PRs:
 * __->__#3888

--- --- ---

### passes: Move implementations into implementation files

It is not longer necessary to have the full Visitor class specifications
inside the headers. This reduces the dependency surface, and makes the
abstractions cleaner. This is part of the plan outlined in #3827.


danobi (approved):

viktormalik (approved):
Depending on whether this gets merged before #3772, the pass in the other PR will have to be updated again.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -76,6 +95,334 @@ static auto getTargetMachine()
...
+public:
+  // Neither a value nor a deletion method.
+  explicit ScopedExpr() = default;
+
+  // Value only.
+  explicit ScopedExpr(Value *value) : value_(value)
+  {
+  }
+
+  // Value with an explicit deletion method.
```
jordalgo, line 120: ugghhh, I really tried not to comment on these comments but they are just not valuable at ALL. If I read the larger comment at the top of the class, which is good, looking at the constructors makes perfect sense. These additional bits above each constructor are just repetitive crud.

jordalgo: I know you're going to mention something about comment consistency and eventual doc blocks yada yada - so I won't block but just know my eyes hurt reading these.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -76,6 +95,334 @@ static auto getTargetMachine()
...
+  void createRet(Value *value = nullptr);
+  int getReturnValueForProbe(ProbeType probe_type);
+
+  // Every time we see a watchpoint that specifies a function + arg pair, we
+  // generate a special "setup" probe that:
+  //
+  // * sends SIGSTOP to the tracee
+  // * pulls out the function arg
+  // * sends an asyncaction to the bpftrace runtime and specifies the arg value
+  //   and which of the "real" probes to attach to the addr in the arg
```
jordalgo, line 320:
```suggestion
  // * sends an asyncaction to the bpftrace runtime and specifies the arg value
  //   which the "real" probes attach to, via the arg value's addr
```

jordalgo (approved):

src/ast/passes/codegen_llvm.cpp
amscanne: That's fine & fair, but I'm not changing anything here. This is strictly a move from the header, and it would be bad to make adjustments. Same for the comments below.

BUT as an aside, I feel like you're optimizing for the wrong thing with comments. They aren't supposed to be minimal and elegant, they are supposed to be expository & error correcting. Redundancy isn't strictly a bad thing.

PR #3889, opened 2025-03-12
Standardize long ret values

jordalgo:
For all BPF helpers that return long
we set that type to int64 except in
two cases: bpf_skb_output, bpf_get_stack.

This updates those two cases to also
treat long as 64 bits instead of 32.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -2042,7 +2042,7 @@ CallInst *IRBuilderBPF::CreateGetStack(Value *ctx,
   // Return: The non-negative copied *buf* length equal to or less than
   // *size* on success, or a negative error in case of failure.
   FunctionType *getstack_func_type = FunctionType::get(
-      getInt32Ty(),
+      getInt64Ty(),
       { getPtrTy(), getPtrTy(), getInt32Ty(), getInt64Ty() },
       false);
   CallInst *call = CreateHelperCall(libbpf::BPF_FUNC_get_stack,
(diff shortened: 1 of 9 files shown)


jordalgo:
Tested on 5.15 kernel via `vmtest -k $(nix build --print-out-paths .#kernel-5_15)/bzImage -- ./build/src/bpftrace -e "'BEGIN { print(kstack); exit(); }'"`

jordalgo:
> No codegen test for skboutout?

Ha, I guess not. I'll add one here.

danobi:
Are we sure it's sound to treat `long` on 32-bit systems as 64-bit? It's gotta be some kind of ABI mismatch, no?

jordalgo:
> Are we sure it's sound to treat long on 32-bit systems as 64-bit?

@danobi I'm not sure but it's how we've been treating all the other helpers that return long (aside from these two). This change just makes it consistent.

jordalgo:
> No codegen test for skboutout?

I'm gonna try this in a follow-up PR. I tried a few times here but my local codegen doesn't match the one generated in the CI and I'm not sure why.

eddyz87:
> Ack - consistent is a good first step. No objections on that. Was just thinking ahead to if we need a more thorough fix.
> 
> cc @eddyz87, perhaps you have some more wisdom to offer

Both functions are defined to have `bpf_func_proto->ret_type == RET_INTEGER` in the kernel. The `verifier.c:check_helper_call()` marks r0 to be a scalar of unknown 64-bit range in such cases. => As far as I understand, kernel does not really care about how return values of these functions are used. So, no wisdom to share :)

PR #3890, opened 2025-03-13
tests: fix kretprobe_order

amscanne:
It is possible that the first probe fires multiple times before the second probe is attached. The test should use a regular expression to allow for this possibility.

##### Checklist


jordalgo:
Thank you for fixing! I saw this recently and was going to put it on my todo list.

jordalgo (approved):

PR #3895, opened 2025-03-14
misc: add default configuration for VSCode devcontainers

amscanne:
This wraps the existing nix configuration by using `direnv`. Essentially we build a devcontainer that has Nix for the current user, and immediately enter the default devshell.

##### Checklist

- [x] ~Language changes are updated in `man/adoc/bpftrace.adoc`~ Developer documentation is updated

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,9 @@
+FROM mcr.microsoft.com/vscode/devcontainers/base:bullseye
+
+# Required nix install dependencies.
+RUN apt-get update && apt-get install -y curl bzip2 adduser direnv vim
+RUN su - vscode -c 'sh <(curl -L https://releases.nixos.org/nix/nix-2.26.3/install) --no-daemon'
+RUN su - vscode -c 'mkdir -p ~/.config/nix && echo "extra-experimental-features = nix-command flakes" > ~/.config/nix/nix.conf'
+RUN su - vscode -c 'nix-env -iA nixpkgs.direnv'
+RUN su - vscode -c 'direnv hook bash >> ~/.bashrc'
+USER vscode
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,13 @@
+{
+  "build": { "dockerfile": "Dockerfile" },
+   "customizations": {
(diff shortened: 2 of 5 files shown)


.devcontainer/Dockerfile
danobi, line 1: Put this in `docker/` dir?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/vscode/devcontainers/base:bullseye
+
+# Required nix install dependencies.
+RUN apt-get update && apt-get install -y curl bzip2 adduser direnv
+RUN su - vscode -c 'sh <(curl -L https://releases.nixos.org/nix/nix-2.26.3/install) --no-daemon'
```
danobi, line 5: If you use determinate nix (https://github.com/DeterminateSystems/nix-installer), I think you can skip all the experimental flakes stuff. Determinate's flavor enables it by default.

danobi (commented):
Seems fine to add. But do you plan on using it, though? If not (and assuming nobody else does), I imagine it'll just start bitrotting immediately.

.devcontainer/Dockerfile
amscanne: I would still need the other two. I looked at the determinate systems installer for a bit, it seems nice. But I've only ever used the upstream installer, and I got it all working (and even pinned). Unless there's a hard reason, I'd rather not switch over just to drop the one line setting the configuration :)

.devcontainer/Dockerfile
amscanne: It has a fixed location, which is found by VSCode.

amscanne: Or at least, the Dockerfile here may be movable (I could specify a relative path possibly?), but there's a strong convention here.

amscanne:
> Seems fine to add. But do you plan on using it, though? If not (and assuming nobody else does), I imagine it'll just start bitrotting immediately.

I've been using it occasionally. It is written in such a way that as long as Nix is working, the dev container should be working (mostly). E.g. we don't need to maintain or update dependencies here, just in the Nix flake.

My preference would be to merge, and if it stops to bitrot then just drop it.

danobi (approved):

PR #3897, opened 2025-03-14
update cmake for zlib libraries

jordalgo:
When this PR landed[1] it caused two
of our distro builds to fail wit this error:
`90.10 /usr/bin/ld: /usr/lib/x86_64-linux-gnu/libbpf.so: undefined reference to symbol 'gzgets'`

I think it's because the CMakeLists file
for utils no longer includes the zlib libraries.

This change moves them to the right CMakeLists file.

[1] https://github.com/bpftrace/bpftrace/pull/3881

Failing jobs:
https://github.com/bpftrace/bpftrace/actions/runs/13845989045/job/38744400466

##### Checklist

### DIFF
--- a/src/util/CMakeLists.txt
+++ b/src/util/CMakeLists.txt
@@ -21,4 +21,4 @@ if (KERNEL_HEADERS_DIR)
 endif()
 target_compile_definitions(util PUBLIC KERNEL_HEADERS_DIR="${KERNEL_HEADERS_DIR}")
 
-target_link_libraries(util debugfs)
+target_link_libraries(util debugfs ${ZLIB_LIBRARIES})


jordalgo:
I still need to test on docker unless someone is more confident that me about this change.

danobi:
It's weird the linker complains about libbpf.so's use of gzgets(). I think in reality it's libutil.a:
```
$ rg gzgets
src/util/kernel.cpp
297:    // gzopen/gzgets handle both uncompressed and compressed files
303:    while (gzgets(file, buf, sizeof(buf))) {
```

So I think this is the more correct fix. You still probably want libbpf to explicitly link against libz.

```diff
diff --git a/src/util/CMakeLists.txt b/src/util/CMakeLists.txt
index 1376456c..57d0b4e8 100644
--- a/src/util/CMakeLists.txt
+++ b/src/util/CMakeLists.txt
@@ -21,4 +21,4 @@ if (KERNEL_HEADERS_DIR)
 endif()
 target_compile_definitions(util PUBLIC KERNEL_HEADERS_DIR="${KERNEL_HEADERS_DIR}")
 
-target_link_libraries(util debugfs)
+target_link_libraries(util debugfs ${ZLIB_LIBRARIES})
```

amscanne (approved):
Thanks for fixing this, I figured the lack of cmake tweaking was too good to be true. Hopefully that's it.

danobi (approved):

PR #3900, opened 2025-03-14
log: fix reference binding for `LogStream`

amscanne:
Not sure how this snuck in there during the refactor, but the reference binding for two log constructors was not correct. This is unused at the moment, except for the `DEBUG` and `BUG` cases (the latter of which goes through a separate constructor).

##### Checklist


jordalgo (approved):

danobi (approved):
nice catch

PR #3914, opened 2025-03-18
clang-tidy: Turn on most readability checks

danobi:
Turns out most of them are pretty good.

Some of the bigger changes I pulled into separate commits. Final commit switches to deny-list, similar to modernize checks.

For the final commit, here's are the lints associated with most of the generated changes: https://gist.github.com/danobi/38197d1842a749a471754953c15f6351

##### Checklist

### DIFF
--- a/src/aot/aot_main.cpp
+++ b/src/aot/aot_main.cpp
@@ -29,7 +29,6 @@ void usage(std::ostream& out, std::string_view filename)
   out << "    -V, --version  bpftrace version" << std::endl;
   out << std::endl;
   // clang-format on
-  return;
 }
 
 std::unique_ptr<Output> prepare_output(const std::string& output_file,
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1238,7 +1238,7 @@ void IRBuilderBPF::CreateCheckSetRecursion(const Location &loc,
   // to a crash e.g. "queued_spin_lock_slowpath" but it can also happen
   // for nested probes e.g. "page_fault_user" -> "print".
   CreateAtomicIncCounter(to_string(MapType::EventLossCounter),
-                         bpftrace_.event_loss_cnt_key_);
+                         bpftrace::BPFtrace::event_loss_cnt_key_);
   CreateRet(getInt64(early_exit_ret));
 
   SetInsertPoint(lookup_failure_block);
@@ -2170,7 +2170,7 @@ void IRBuilderBPF::CreateRingbufOutput(Value *data,
 
   SetInsertPoint(loss_block);
   CreateAtomicIncCounter(to_string(MapType::EventLossCounter),
-                         bpftrace_.event_loss_cnt_key_);
+                         bpftrace::BPFtrace::event_loss_cnt_key_);
   CreateBr(merge_block);
 
   SetInsertPoint(merge_block);
@@ -2225,7 +2225,6 @@ void IRBuilderBPF::CreateMapElemInit(Value *ctx,
   CreateStore(val, initValue);
   CreateMapUpdateElem(ctx, map.ident, key, initValue, loc, BPF_NOEXIST);
   CreateLifetimeEnd(initValue);
-  return;
 }
 
 void IRBuilderBPF::CreateMapElemAdd(Value *ctx,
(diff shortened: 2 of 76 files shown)


src/aot/aot_main.cpp
```diff
@@ -63,7 +62,8 @@ std::unique_ptr<Output> prepare_output(const std::string& output_file,
 
 int main(int argc, char* argv[])
 {
-  std::string output_file, output_format;
```
jordalgo, line 66: why is this bad?

src/ast/irbuilderbpf.cpp
```diff
@@ -2170,7 +2172,7 @@ void IRBuilderBPF::CreateRingbufOutput(Value *data,
 
   SetInsertPoint(loss_block);
   CreateAtomicIncCounter(to_string(MapType::EventLossCounter),
-                         bpftrace_.event_loss_cnt_key_);
+                         bpftrace::BPFtrace::event_loss_cnt_key_);
```
jordalgo, line 2175: This kinda freaks me out. Should this member var not be static?

src/config.h
```diff
@@ -203,8 +203,7 @@ class Config {
     }
   }
 
-private:
```
jordalgo, line 206: Why this removal?

src/format_string.h
```diff
@@ -45,23 +45,23 @@ class FormatString {
   std::string format_str(std::vector<std::unique_ptr<IPrintable>> &args);
 
   // length returns the length of the format string
-  inline size_t length() const noexcept
```
jordalgo, line 48: Are these compiler directives considered bad?

jordalgo (commented):
Overall looks good but I have some questions

src/aot/aot_main.cpp
danobi: I have no preference. Just what clang-tidy seems to prefer.

This is the check: https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html

src/ast/irbuilderbpf.cpp
danobi: It's the key value and value value (heh) for the map it writes into for the loss counter.

The impl should probably be refactored to use global vars instead of an arraymap.

src/ast/passes/config_analyser.cpp
```diff
@@ -102,7 +102,7 @@ void ConfigAnalyser::set_config(AssignConfigVarStatement &assignment,
 }
 
 void ConfigAnalyser::set_config(AssignConfigVarStatement &assignment,
-                                [[maybe_unused]] ConfigKeyStackMode)
+                                [[maybe_unused]] ConfigKeyStackMode /*unused*/)
```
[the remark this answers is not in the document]
danobi: Will fix. I'll just give it a name like the other functions in the file

src/format_string.h
danobi: `inline` doesn't do anything if used on a member defined in the class definition.

https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more

src/config.h
danobi: There's a private 2 functions up already. Caught by https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-access-specifiers.html

src/format_string.h
jordalgo: Ohh TIL.

src/config.h
jordalgo: Oooohhh, nice!

src/ast/irbuilderbpf.cpp
jordalgo: No, I meant that we're using a static class variable to read and increment.

src/aot/aot_main.cpp
jordalgo: Eh, we work in the kernel code base where this is not only allowed but encouraged. I'm sure there are reasons to prefer clang-tidy's way but I feel like it might be too prescriptive - like would we want to block a PR because of this?

danobi: Good point. I will revert the isolate-declaration check

src/ast/irbuilderbpf.cpp
jordalgo: Oh sorry, I was thrown off by the IR instruction: `CreateAtomicIncCounter`

PR #3917, opened 2025-03-19
config: add `unstable_import` as option to enable imports

amscanne:
This adds the statement to the parser, the basic configuration and a dummy pass that checks that configuration and emits a warning.

##### Checklist

Note: I have decided to leave this undocumented in the man page until the feature actually does something. This is just adding the basic plumbing for the nodes.

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -608,11 +608,26 @@ private:
 };
 using SubprogList = std::vector<Subprog *>;
 
+class Import : public Node {
+public:
+  Import(Diagnostics &d, std::string name, Location &&loc);
+
+  const std::string &name() const
+  {
+    return name_;
+  }
+
+private:
+  std::string name_;
+};
+using ImportList = std::vector<Import *>;
+
 class Program : public Node {
 public:
   Program(Diagnostics &d,
           std::string c_definitions,
           Config *config,
+          ImportList &&imports,
           MapDeclList &&map_decls,
           SubprogList &&functions,
           ProbeList &&probes,
@@ -620,6 +635,7 @@ public:
 
   std::string c_definitions;
   Config *config = nullptr;
+  ImportList imports;
   SubprogList functions;
   ProbeList probes;
   MapDeclList map_decls;
--- a/src/ast/passes/resolve_imports.cpp
+++ b/src/ast/passes/resolve_imports.cpp
@@ -1,0 +1,53 @@
+#include <iostream>
+#include <sstream>
+#include <unordered_set>
+
+#include "ast/passes/resolve_imports.h"
+#include "ast/visitor.h"
+#include "bpftrace.h"
+#include "log.h"
+
+namespace bpftrace::ast {
+
+class ResolveImports : public Visitor<ResolveImports> {
+public:
+  ResolveImports(BPFtrace &bpftrace, const std::vector<std::string> &paths)
+      : bpftrace_(bpftrace), paths_(paths) {};
+
+  using Visitor<ResolveImports>::visit;
+  void visit(Import &imp);
+
+  // Should be called after the visit, this will move the result.
+  Imports imports()
+  {
+    return std::move(imports_);
+  }
+
+private:
+  BPFtrace &bpftrace_;
+  const std::vector<std::string> &paths_;
+  Imports imports_;
+};
+
+void ResolveImports::visit(Import &imp)
+{
+  if (!bpftrace_.config_->get(ConfigKeyBool::unstable_import)) {
+    imp.addError() << "Imports are not enabled by default. To enable "
+                      "this unstable feature, set this config flag to 1 "
+                      "e.g. unstable_import=1";
+    return;
+  }
+
+  imp.addWarning() << "Imports are not yet implemented.";
+}
+
+Pass CreateResolveImportsPass(std::vector<std::string> &&import_paths)
+{
+  return Pass::create("ResolveImports",
+                      [import_paths](ASTContext &ast, BPFtrace &b) {
+                        ResolveImports analyser(b, import_paths);
+                        analyser.visit(ast.root);
+                      });
+}
+
+} // namespace bpftrace::ast
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -131,6 +131,7 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %token <std::string> SIZEOF "sizeof"
 %token <std::string> OFFSETOF "offsetof"
 %token <std::string> LET "let"
+%token <std::string> IMPORT "import"
 
 
 %type <ast::Operator> unary_op compound_op
@@ -157,6 +158,8 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %type <ast::Probe *> probe
 %type <std::pair<ast::ProbeList, ast::SubprogList>> probes_and_subprogs
 %type <ast::Config *> config
+%type <ast::Import *> import_stmt
+%type <ast::ImportList> imports
 %type <ast::Statement *> assign_stmt block_stmt expr_stmt if_stmt jump_stmt loop_stmt config_assign_stmt for_stmt
 %type <ast::MapDeclList> map_decl_list
 %type <ast::VarDeclStatement *> var_decl_stmt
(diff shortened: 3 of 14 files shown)


src/parser.yy
```diff
@@ -186,8 +189,8 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %%
 
 program:
-                c_definitions config map_decl_list probes_and_subprogs END {
-                    driver.ctx.root = driver.ctx.make_node<ast::Program>($1, $2, std::move($3), std::move($4.second), std::move($4.first), @$);
+                c_definitions config imports map_decl_list probes_and_subprogs END {
```
jordalgo, line 192: Why imports here as opposed to at the top?

amscanne: Simply because of the fact that you need a config option to enable it. Therefore, I think the config has logical precedence. (I'm not a huge fan of the config block if we wanted to start making these command-line switches? Then I don't think there's any issues with precedence)

jordalgo: > Simply because of the fact that you need a config option to enable it.

But the check for that config option doesn't happen until a later pass so I don't think that should be influencing placement. I think the expectation for most users is that imports happen at the top of a file.

> I'm not a huge fan of the config block if we wanted to start making these command-line switches?

The config block was added explicitly so we don't need to add more command line switches and also make the script more portable. I don't like the idea that you have to remember or lookup these switches to use a script that is using an experimental feature (as an example). It's nice that it goes with the script automatically.

amscanne: > But the check for that config option doesn't happen until a later pass so I don't think that should be influencing placement. I think the expectation for most users is that imports happen at the top of a file.

My conceptualization there has nothing to do with the pass order, I just mean as a user. My expectation for the "config" to the extent that there is one is the tippy top of the file. (I'm thinking about go generate or arch tags, C pragmas, etc. Our own `#pragma once` appears *before* any includes.)

I think this is a fine thing to discuss and debate for the final feature, but I don't see it as a blocker for the `unstable_` variant of the feature, no?

> The config block was added explicitly so we don't need to add more command line switches and also make the script more portable. I don't like the idea that you have to remember or lookup these switches to use a script that is using an experimental feature (as an example). It's nice that it goes with the script automatically.

As an aside, isn't this an argument against these as unstable options inside the file? Now the script could break and change depending on the version you're running it with. The point of the experimental flags is to require the user to opt-in and be aware that the script is unsupported, now that's no longer true. Anything using an `unstable_` is non-portable by design, and now you're just hiding a little booby-trap inside by making it "portable".

jordalgo: > I think this is a fine thing to discuss and debate for the final feature, but I don't see it as a blocker for the unstable_ variant of the feature, no?

I'm not sure TBH. Also it's not just the config which is above the imports but also C definitions, which also says to me (as a user) that whatever C definitions exist in the imported file take precedence (because they come after) if there was a conflict over the ones in the file that are doing the importing. @danobi @viktormalik - any thoughts on this?

jordalgo: > As an aside, isn't this an argument against these as unstable options inside the file? Now the script could break and change depending on the version you're running it with. The point of the experimental flags is to require the user to opt-in and be aware that the script is unsupported, now that's no longer true. Anything using an unstable_ is non-portable by design, and now you're just hiding a little booby-trap inside by making it "portable".

I don't understand this argument. The script WILL break/change depending on the version regardless of if the config is within the script or passed as a CLI arg because the feature is embedded into the functionality of the script itself. We may have the option to ignore some syntax if we pull out an experimental feature but that is probably worse than erroring out as it would have unexpected/surprising consequences.

amscanne: I don't understand this argument. The script WILL break/change depending on the version regardless of if the config is within the script or passed as a CLI arg because the feature is embedded into the functionality of the script itself. We may have the option to ignore some syntax if we pull out an experimental feature but that is probably worse than erroring out as it would have unexpected/surprising consequences.

Definitely not wanting to get into a whole thing about this specific tangent, I just mean the point of the `unstable_` flag is to make sure that users are aware of the dependency on an unstable feature. But the argument that it's better in the script is essentially "that way users don't have to know about it". I think it just depends on whether you think it's important that script **authors* know, or script **users** know. (They are usually the same person, but not always. Scripts get checked in and have a lift of their own.) The point is to force the user to type `--feature=unstable_imports` (at least once) and mentally go (okay, this is using an unstable feature). I mean, we all now know that Nix flakes and subcommands are "experimental" right? :) If `use_experimental_flakes=True` was just in the `flake.nix` script, I would have never known that.

danobi: Regarding user opt-in and author opt-in: I think there are at least 2 schools of thought on this. Nix has user-opt in AFAIK and rust kinda has both. For rust you have to be on the unstable channel as well as using the source level directives.

Maybe it's sufficient to emit a warning if an unstable feature being used? IMO that's a pragmatic middle ground.

danobi: Regarding ordering of things at top level, I think practically speaking we can't be enforcing the order for 5 kinds of constructs. We've got C definitions, config block, imports, map decls, and soon macros. 

My immediate thought (and preference) is to not enforce ordering at all for stuff that can be unambiguously parsed. I suspect over time there'll be a convention but I don't see why we need to enforce it at parser level.

danobi: Yeah let's hash that out separately. No need to do that here

jordalgo: I do think we probably want to be prescriptive about the ordering of some of these things but not all but sure - let's talk in a separate thread.

src/ast/ast.h
```diff
@@ -608,18 +608,34 @@ class Subprog : public Node {
...
+public:
+  Import(Diagnostics &d, std::string name, Location &&loc);
+
+  const std::string &name() const
+  {
+    return name_;
+  }
+
+private:
+  std::string name_;
```
jordalgo, line 621: subprog ast has a private `name_` but all other AST classes exposes their member strings as public. This just to protect the node after creation?

src/ast/passes/resolve_imports.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <unordered_set>
+
+#include "ast/passes/resolve_imports.h"
+#include "ast/visitor.h"
+#include "bpftrace.h"
+#include "log.h"
+
+namespace bpftrace::ast {
+
+// Load imports emit unprocessed source, objects, etc.
```
jordalgo, line 12: "Load imports then emit unprocessed source, objects, etc." ?

src/ast/passes/resolve_imports.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+private:
+  BPFtrace &bpftrace_;
+  const std::vector<std::string> &paths_;
+  Imports imports_;
+};
+
+void ResolveImports::visit(Import &imp)
+{
+  if (!bpftrace_.config_->get(ConfigKeyBool::unstable_import)) {
```
jordalgo, line 35: I'm going to make a helper for this sort of thing that passes can call.

jordalgo (approved):
One small comment nit and ordering of statements aside, this looks good to me.

src/ast/passes/resolve_imports.cpp
amscanne: I'll just drop this given that it's just a skeleton.

src/ast/ast.h
amscanne: We should probably make all of these consistent. Given that Subprog and SubprogArg are already different, I can do that as a separate thing. I have a pending branch with some other consistency issues (e.g. the use of explicit constructors, other private member).

src/ast/passes/resolve_imports.cpp
amscanne: Sure, I think it's fine though. I'd still want to check it within the context of the pass, although maybe you're talking about just standardizing the message? That sounds fine.

jordalgo (dismissed):

jordalgo (dismissed):

jordalgo:
The only bit that gives me pause is the changelog addition - it does specify "syntax" but the functionality doesn't do anything yet (at least in this PR)

amscanne:
> The only bit that gives me pause is the changelog addition - it does specify "syntax" but the functionality doesn't do anything yet (at least in this PR)

Should I remove this? I don't really have a strong preference

jordalgo:
> Should I remove this? I don't really have a strong preference

I don't feel strongly either but maybe it's best to remove for now until it's something we want users to know about.

amscanne:
I've dropped the CHANGELOG update.

jordalgo (approved):

PR #3918, opened 2025-03-19
Update rawtracepoints to use correct types and utilize BTF

jordalgo:
This updates rawtracepoints to use the correct prog and attach types, which will give rawtracepoints access to functions
like 'skboutput' and others.

This also utilizes BTF to read rawtracepoint arguments to make the `args` builtin available to rawtracepoints and allows users to see their params when listing in verbose mode.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1482,29 +1482,22 @@ profile:hz:99 { @[tid] = count(); }
 .short name
 * `rt`
 
-The hook point triggered by `tracepoint` and `rawtracepoint` is the same.
-`tracepoint` and `rawtracepoint` are nearly identical in terms of functionality.
-The only difference is in the program context.
-`rawtracepoint` offers raw arguments to the tracepoint while `tracepoint` applies further processing to the raw arguments.
-The additional processing is defined inside the kernel.
+Raw tracepoints are attached to the same tracepoints as normal tracepoint programs.
+The reason why you might want to use raw tracepoints over normal tracepoints is due to the performance improvement - https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_RAW_TRACEPOINT/[Read More].
+
+`rawtracepoint` arguments can be accessed via the `argN` builtins AND via the `args` builtin.
 
 ----
-rawtracepoint:block_rq_insert {
-  printf("%llx %llx\n", arg0, arg1);
+rawtracepoint:kfree_skb {
+  printf("%llx %llx\n", arg0, args.skb);
 }
 ----
+`arg0` and `args.skb` will print the same address.
 
-Tracepoint arguments are available via the `argN` builtins.
-Each arg is a 64-bit integer.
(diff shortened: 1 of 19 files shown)


danobi:
Runtime test for skboutput in rawtracepoint probe?

jordalgo:
> Runtime test for skboutput in rawtracepoint probe?

Good call though I might just add a test to confirm it can attach as it seems the current skboutput runtime test is a bit ... involved and I don't think it's necessary to repeat all that logic.

src/ast/passes/field_analyser.cpp
```diff
@@ -214,7 +214,7 @@ void FieldAnalyser::resolve_args(Probe &probe)
 
     auto probe_type = probetype(ap->provider);
     if (probe_type != ProbeType::fentry && probe_type != ProbeType::fexit &&
-        probe_type != ProbeType::uprobe)
```
jordalgo, line 217: From what i could tell, having uprobe here was a mistake and didn't do anything.

jordalgo:
@danobi I added one more larger commit on top of this so rawtracepoints can utilize the `args` builtin.

jordalgo:
@viktormalik You like BTF stuff right? Mind having a look?

src/btf.h
```diff
@@ -29,6 +30,15 @@ struct btf_type;
 
 namespace bpftrace {
 
+// Note: there are several prefixes for raw tracepoint BTF functions.
+// "__probestub_" seems to be the most accurate in terms of getting the params
+// but it wasn't added until May 2023 so older kernels might not have it,
+// which is why we also check "__traceiter_" (as needed).
```
viktormalik, line 36: Could we add links to some kernel docs here? Or at least the patch sets that added the `__probestub_` and `__traceiter_` prefixes?

viktormalik (commented):
Looks great! Just a few comments below.

man/adoc/bpftrace.adoc
```diff
@@ -1482,29 +1482,22 @@ profile:hz:99 { @[tid] = count(); }
...
+Raw tracepoints are attached to the same tracepoints as normal tracepoint programs.
+The reason why you might want to use raw tracepoints over normal tracepoints is due to the performance improvement - https://docs.ebpf.io/linux/program-type/BPF_PROG_TYPE_RAW_TRACEPOINT/[Read More].
+
+`rawtracepoint` arguments can be accessed via the `argN` builtins AND via the `args` builtin.
 
 ----
-rawtracepoint:block_rq_insert {
-  printf("%llx %llx\n", arg0, arg1);
+rawtracepoint:kfree_skb {
+  printf("%llx "%llx\n", arg0, args.skb);
```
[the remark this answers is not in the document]
jordalgo: Good catch

src/btf.h
jordalgo: Yeah, I'll try to find something to add as a reference.

jordalgo:
@viktormalik - Comments addressed

viktormalik (approved):

PR #3919, opened 2025-03-19
ast: Split `TupleAccess` into its own node

amscanne:
Since this is unamibiguously parseable (`$x.0` vs `$x.arg`), there is no need for the `FieldAccess` node to be doing double duty. This reduces the complexity of the resulting code.

##### Checklist


jordalgo:
Got some conflicts but in general, I like this change.

jordalgo (approved):

PR #3920, opened 2025-03-19
ast: Fix MapDeclStatement; not an expression

amscanne:
##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -214,7 +214,7 @@ public:
   std::vector<std::string> field;
 };
 
-class MapDeclStatement : public Expression {
+class MapDeclStatement : public Node {
 public:
   explicit MapDeclStatement(Diagnostics &d,
                             std::string ident,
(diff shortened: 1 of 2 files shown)


src/ast/ast.h
```diff
@@ -214,7 +214,7 @@ class Offsetof : public Expression {
   std::vector<std::string> field;
 };
 
-class MapDeclStatement : public Expression {
+class MapDeclStatement : public Node {
```
jordalgo, line 217: @amscanne Shouldn't this be a `Statement` instead of a `Node`?  Second question: should we just remove `Statement` as it seems to be the same as `Node`.

jordalgo (approved):

src/ast/ast.h
amscanne: No, statements appear in blocks. This isn't something that can be parsed as part of a block. Perhaps it should be named `MapDecl`, but alas. This stuff will probably keep changing quite a lot, no need to fuss, I just noticed that this wasn't quite right and fired off a PR.

Issue #1743, opened 2021-03-19
Add support for statement probes 


ema:
### Is your feature request related to a problem? Please describe.
Probing function calls and return statements can go a long way, but it is often necessary to probe somewhere inside the body of a function. The only way of doing that in bpftrace, as far as I'm aware, is by disassembling, looking at the assembly and [specifying the right offset in the probe definition](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#3-uprobeuretprobe-dynamic-tracing-user-level).

### Describe the solution you'd like

SystemTap supports probing by statement: users can list all probe points available within the body of a function, and probe by line number.

```
$ sudo stap -L 'process("/bin/ls").statement("print_name_with_quoting@src/ls.c:*")'
process("/bin/ls").statement("print_name_with_quoting@src/ls.c:4731") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t
process("/bin/ls").statement("print_name_with_quoting@src/ls.c:4734") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t
process("/bin/ls").statement("print_name_with_quoting@src/ls.c:4735") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t $name:char const*
process("/bin/ls").statement("print_name_with_quoting@src/ls.c:4737") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t $name:char const* $color:struct bin_str const*
process("/bin/ls").statement("print_name_with_quoting@src/ls.c:4740") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t $name:char const* $used_color_this_time:_Bool
process("/bin/ls").statement("print_name_with_quoting@src/ls.c:4741") $f:struct fileinfo const* $symlink_target:_Bool $stack:struct obstack* $start_col:size_t $name:char const* $used_color_this_time:_Bool
```
It would be very useful to have the same feature in bpftrace.

danobi:
We have support for kernel arguments: https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#16-kfunckretfunc-kernel-functions-tracing-arguments

User arguments would require some cost/benefit analysis. Seems like it'd require deep language support (for anything other than C).

PR #3921, opened 2025-03-19
Remove LLDB

danobi:
DWARF support, particularly as delivered by LLDB, was fraught with
landmines. More details about why LLDB was troublesome is available in
PR https://github.com/bpftrace/bpftrace/pull/3763.

This commit finishes off what that PR started by deleting the now
untested code.

This closes https://github.com/bpftrace/bpftrace/issues/3392, https://github.com/bpftrace/bpftrace/issues/3446, https://github.com/bpftrace/bpftrace/issues/3173, https://github.com/bpftrace/bpftrace/issues/3083, https://github.com/bpftrace/bpftrace/issues/3084.

##### Checklist


chenhengqi:
cc @curu

danobi:
This PR isn't an indication that we're abandoning richer userspace tracing forever. Rather, the DWARF based approach isn't working. We're already on our second attempt (first libdw, now liblldb) and it's not looking promising.

We have new ideas about approaching userspace tracing with BTF. One potential idea is to library-ify pahole (eg libpahole) and generate BTF from DWARF on the fly. Or ship applications with BTF prebuilt. That way the ingress point into bpftrace is C, a very well understood FFI boundary. It keep bpftrace simpler (and hopefully more robust). And all the DWARF handling is contained to pahole, a place where DWARF is very well understood, well maintained, and has support for multiple languages. A cleaner separation of concerns.

src/config.cpp
```diff
@@ -35,13 +35,7 @@ Config::Config(bool has_cmd)
...
     { ConfigKeyString::str_trunc_trailer, { .value = std::string("..") } },
     { ConfigKeySymbolSource::default_,
-      { .value =
-#ifdef HAVE_LIBLLDB
-            ConfigSymbolSource::dwarf
-#else
-            ConfigSymbolSource::symbol_table
-#endif
-      } },
+      { .value = ConfigSymbolSource::symbol_table } },
```
jordalgo, line 38: I think this entire config can go and perhaps we should issue a warning if we see it but not break the script if it's set to "symbol_table"

jordalgo (approved):
One small thing with the SymbolSource config but otherwise LGTM. Nice cleanup. It's a little sad we'll be losing this functionality but I agree that a different approach is needed.

danobi:
> Hm, sorry for being late here (was off for a week) but do I understand it correctly that we're just dropping a number of DWARF-based features, especially access to uprobe args by name (https://github.com/bpftrace/bpftrace/pull/1994) and type resolution from DWARF (https://github.com/bpftrace/bpftrace/pull/2034, https://github.com/bpftrace/bpftrace/pull/2505), just because LLDB is full of bugs troublesome to work with?

I put some thoughts above (https://github.com/bpftrace/bpftrace/pull/3921#issuecomment-2738943951), but yes, one reason was LLDB is not a good dependency. The other was that I think DWARF is not quite the right ingress point for userspace tracing. I think it's too complex to work with on top of being generally unavailable by default for userspace applications. The libdw implementation more or less only worked with C, right? Which is quite limiting (not too many applications are done in C these days). Long term I was thinking we could handle userspace the same way as we do the kernel - with BTF. That seems to me a more scalable solution.

That being said, if you feel strongly (and please feel free to push) that we should not regress the libdw-based functionality, I can take on the task of putting the libdw bits back in. At least until the next gen BTF approach is proven (if at all). Since I am the one removing the stuff, after all.

viktormalik:
> The other was that I think DWARF is not quite the right ingress point for userspace tracing. I think it's too complex to work with 

I beg to differ here. I believe that DWARF has been specifically *designed for* userspace tracing (in the end, debugging and tracing are not that much different concepts). Sure, it has its flaws and design issues, some of which are addressed by other formats like BTF (one could say that BTF is just deduplicated DWARF), but that doesn't change the fact that it's the most common format (and standard) for storing debugging information.

> on top of being generally unavailable by default for userspace applications.

That depends. For distro-shipped applications, distros usually provide `-debuginfo` packages which allow users to install DWARF data for pretty much any application. For user-built applications, the situation is probably different and I assume that people usually do not build with DWARF.

> The libdw implementation more or less only worked with C, right? Which is quite limiting (not too many applications are done in C these days).

I've never tested it on anything else but DWARF supports (and is by default used by) many other languages - C++, Rust, Go, etc.

> Long term I was thinking we could handle userspace the same way as we do the kernel - with BTF. That seems to me a more scalable solution. We can more easily handle more languages (and better) by leveraging and enhancing pahole.

Agreed, but it seems like a long way to go. And, on top of that, BTF is still generated from DWARF so conceptually, wherever you can use BTF, you can use DWARF either. You just need a different (and potentially more complex due to the nature of the format) way for parsing it.

It would be nice if one day, applications embedded BTF just like the kernel does but until we get there, we'll have to use pahole to generate it and that means getting DWARF first.

> That being said, if you feel strongly (and please feel free to push) that we should not regress the libdw-based functionality

I'm not sure how many people actually used that functionality but to me, it seemed like a nice set of features. On top of that, there are other not-yet-implemented DWARF-enabled features like attachment of uprobes by source line number (#1743). From what I've heard, this is one of the functionalities users very much miss from bpftrace compared to other tracing tools like SystemTap. And I'm not sure if BTF even has the information which would enable such an extension.

So, overall, I don't really see a reason to remove DWARF support from bpftrace.

viktormalik:
And it looks like people are using it: #3939.

jordalgo:
@viktormalik Ack on all your points. Not to speak for @danobi but he and I chatted yesterday and he agreed to bring back libdw so as not to drop these features. I'll also volunteer myself to bring it back if @danobi is too busy.

viktormalik:
@jordalgo thanks!

It's a shame we have to let LLDB go as it provided some nice functionality (especially attachment to inlined functions). It should be still possible to achieve the same with libdw but it may a lot of additional work :disappointed:

danobi:
> > The other was that I think DWARF is not quite the right ingress point for userspace tracing. I think it's too complex to work with
> 
> I beg to differ here. I believe that DWARF has been specifically _designed for_ userspace tracing (in the end, debugging and tracing are not that much different concepts). Sure, it has its flaws and design issues, some of which are addressed by other formats like BTF (one could say that BTF is just deduplicated DWARF), but that doesn't change the fact that it's the most common format (and standard) for storing debugging information.

My issue with DWARF isn't that it's complex (although it is), it's that the scope is boundless. I don't think we'll ever fully support everything in DWARF. BTF (or perhaps CTF after speaking with Arnaldo this week) is nice b/c we can say "we fully support everything in BTF, now meet us in the middle and make your language work for BTF".

The line is much fuzzier when DWARF is involved - we cannot easily say which subset we support. And application writers likely do not have the mental model to write code that causes specific DWARF generation. Contrast that to BTF/CTF where a C FFI boundary is not only common but well understood - all popular languages have some form of C interop.

> 
> > on top of being generally unavailable by default for userspace applications.
> 
> That depends. For distro-shipped applications, distros usually provide `-debuginfo` packages which allow users to install DWARF data for pretty much any application. For user-built applications, the situation is probably different and I assume that people usually do not build with DWARF.

Sure, fair enough. And there's debuginfod which makes it more available.

> 
> > The libdw implementation more or less only worked with C, right? Which is quite limiting (not too many applications are done in C these days).
> 
> I've never tested it on anything else but DWARF supports (and is by default used by) many other languages - C++, Rust, Go, etc.

Right, but I wonder how much of C++/Rust/etc. map to any assumptions we may have made. 

> 
> > Long term I was thinking we could handle userspace the same way as we do the kernel - with BTF. That seems to me a more scalable solution. We can more easily handle more languages (and better) by leveraging and enhancing pahole.
> 
> Agreed, but it seems like a long way to go. And, on top of that, BTF is still generated from DWARF so conceptually, wherever you can use BTF, you can use DWARF either. You just need a different (and potentially more complex due to the nature of the format) way for parsing it.
> 
> It would be nice if one day, applications embedded BTF just like the kernel does but until we get there, we'll have to use pahole to generate it and that means getting DWARF first.
> 
> > That being said, if you feel strongly (and please feel free to push) that we should not regress the libdw-based functionality
> 
> I'm not sure how many people actually used that functionality but to me, it seemed like a nice set of features. On top of that, there are other not-yet-implemented DWARF-enabled features like attachment of uprobes by source line number (#1743). From what I've heard, this is one of the functionalities users very much miss from bpftrace compared to other tracing tools like SystemTap. And I'm not sure if BTF even has the information which would enable such an extension.

FWIW, BTF does support line numbers: https://docs.kernel.org/bpf/btf.html#bpf-prog-load. But regardless, I generally agree with your point - DWARF does have a lot more stuff. But the larger question I'd pose is why someone would want to use bpftrace for userspace tracing over gdb/lldb. It has been possible to script gdb for at least a decade (https://sdimitro.github.io/post/scripting-gdb/). The lack of wide spread usage (at least from my PoV) says to me it's not solving a big enough pain point. My take is that userspace is crowded with good options, a very good option being just recompile with printf()s. 

I think it'd be a different story if we were bringing something otherwise not possible to the table - in that case I think it makes sense to invest. But in lieu of that, my thinking is how we enable motivated users to "meet us in the middle" with the least amount of effort from our side. And that looked to me like BTF support.

> 
> So, overall, I don't really see a reason to remove DWARF support from bpftrace.

Ack. While I'm skeptical we'll ever start getting new language support for free using DWARF approach, I have no objections to leaving the libdw implementation in. It did not cause maintenance issues like the lldb one.

I'll put the libdw resurrection at the top of my todo list.

danobi:
> @jordalgo thanks!
> 
> It's a shame we have to let LLDB go as it provided some nice functionality (especially attachment to inlined functions). It should be still possible to achieve the same with libdw but it may a lot of additional work 😞

Please see the latest LSFMM talk about inline function probing. Should be up in a few days I think. We're working on attacking that from the kernel side. Hopefully it solves the inline tracing problem for the entire ecosystem, all at once.

PR #3922, opened 2025-03-20
ast: Introduce transforms for map functions

amscanne:
Rather than relying on look-aside information in the AST, use transformations for these map functions to pass the map as an argument.

##### Checklist


amscanne:
Now part of #3988.

PR #3923, opened 2025-03-20
Don't issue warning if initial map value exists

jordalgo:
Seeing a warning that users can't do anything about
and shouldn't be possible.

```
scuba_queries_oom_profiler.bt:65:26-33: WARNING: File exists
Additional Info - helper: map_update_elem, retcode: -17
  @agg_counts[$agg_id] = count();
                         ~~~~~~~
```

I'm not sure how this warning is even happening.
These are non-sleepable probes so how can the element
not exist in the bpf_map_lookup_elem call and then
exist on that same cpu moments later.

Either way, we don't need to guard against
insertion, in the case of per-cpu maps if we try
to look up the element and it doesn't exist
then we shouldn't worry about this warning.

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1092,15 +1092,14 @@ void IRBuilderBPF::CreateMapUpdateElem(Value *ctx,
                                        const std::string &map_ident,
                                        Value *key,
                                        Value *val,
-                                       const Location &loc,
-                                       int64_t flags)
+                                       const Location &loc)
 {
   Value *map_ptr = GetMapVar(map_ident);
   assert(ctx && ctx->getType() == getPtrTy());
   assert(key->getType()->isPointerTy());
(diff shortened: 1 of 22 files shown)


jordalgo:
damn, looks like I need a more nuanced solution

amscanne:
> I think this is simply due to concurrency issues
> and is not really a problem. Let's stop warning.

I don't think it's a major problem, but it *is* a problem, no? You'll just get lost events. The syntax for `count` is a specialized and esoteric, so I would be expecting it to handle this case (the result should be 2 and no warning, instead of 1 and no warning, no?) :)

jordalgo:
> The syntax for count is a specialized

In this case it's all map aggregation functions that end up calling `CreateMapElemAdd`. 

Truth be told, I'm not sure how this warning is even happening. These are non-sleepable probes so how can the element not exist in the bpf_map_lookup_elem call and then exist on that same cpu moments later.

PR #3929, opened 2025-03-24
Fix not showing pass warnings

jordalgo:
##### Checklist


src/main.cpp
```diff
@@ -1028,6 +1028,8 @@ int main(int argc, char* argv[])
     return 1;
   }
 
+  ast.diagnostics().emit(std::cout);
```
amscanne, line 1031: Add a comment, indicating that this is required to show warnings and that if there were errors they would have been emitted above?

amscanne (dismissed):

PR #3930, opened 2025-03-24
Add warning for use of unstable features

jordalgo:
This is to notify users of a script using unstable features. Inspired by conversation here:
https://github.com/bpftrace/bpftrace/pull/3917#discussion_r2010025715

##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -377,11 +377,11 @@ public:
 class AssignConfigVarStatement : public Statement {
 public:
   AssignConfigVarStatement(Diagnostics &d,
-                           std::string config_var,
+                           Identifier *config_var,
                            Expression *expr,
                            Location &&loc);
 
-  std::string config_var;
+  Identifier *config_var = nullptr;
   Expression *expr = nullptr;
 };
 
--- a/src/ast/passes/config_analyser.cpp
+++ b/src/ast/passes/config_analyser.cpp
@@ -50,9 +50,9 @@ void ConfigAnalyser::log_type_error(SizedType &type,
                                     Type expected_type,
                                     AssignConfigVarStatement &assignment)
 {
-  assignment.addError() << "Invalid type for " << assignment.config_var
-                        << ". Type: " << type.GetTy()
-                        << ". Expected Type: " << expected_type;
+  assignment.expr->addError()
+      << "Invalid type for " << assignment.config_var->ident
+      << ". Type: " << type.GetTy() << ". Expected Type: " << expected_type;
 }
 
 void ConfigAnalyser::set_config(AssignConfigVarStatement &assignment,
@@ -82,8 +82,9 @@ void ConfigAnalyser::set_config(AssignConfigVarStatement &assignment,
   } else if (val == 1) {
     config_setter_.set(key, true);
   } else {
-    assignment.addError() << "Invalid value for " << assignment.config_var
-                          << ". Needs to be 0 or 1. Value: " << val;
+    assignment.expr->addError()
+        << "Invalid value for " << assignment.config_var->ident
+        << ". Needs to be 0 or 1. Value: " << val;
   }
 }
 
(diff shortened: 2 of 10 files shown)


src/config.h
```diff
@@ -98,7 +100,8 @@ const std::map<std::string, ConfigKey> CONFIG_KEY_MAP = {
   { "missing_probes", ConfigKeyMissingProbes::default_ },
   { "print_maps_on_exit", ConfigKeyBool::print_maps_on_exit },
   { "use_blazesym", ConfigKeyBool::use_blazesym },
-  { "unstable_map_decl", ConfigKeyBool::unstable_map_decl },
+  { std::string{ UNSTABLE_PREFIX } + "map_decl",
```
danobi, line 103: This makes the key un-greppable. I think it's probably ok to just rely on convention? Once we get the unstable features really going it'll be hard to mess up (due to copy paste effect)

tests/config_analyser.cpp
```diff
@@ -38,6 +38,47 @@ void test(BPFtrace &bpftrace,
...
+                                 bool invert)
+{
+  ast::ASTContext ast("stdin", input);
+
+  // N.B. No tracepoint expansion, but we will do the double parse to enable
+  // macro expansion.
+  auto ok = ast::PassManager()
+                .put(ast)
+                .put(bpftrace)
+                .add(ast::AllParsePasses())
```
danobi, line 53: Just CreateParsePass(), not AllParsePasses() is sufficient

tests/config_analyser.cpp
```diff
@@ -38,6 +38,47 @@ void test(BPFtrace &bpftrace,
...
 
+ast::ASTContext test_for_warning(BPFtrace &bpftrace,
+                                 const std::string &input,
+                                 const std::string &warning,
+                                 bool invert)
+{
+  ast::ASTContext ast("stdin", input);
+
+  // N.B. No tracepoint expansion, but we will do the double parse to enable
+  // macro expansion.
```
danobi, line 49: Not relevant to config analyzer tests?

danobi (commented):
nice second commit!

tests/config_analyser.cpp
jordalgo: Too much copy/pasta.

src/ast/ast.h
```diff
@@ -377,11 +377,11 @@ class AssignVarStatement : public Statement {
 class AssignConfigVarStatement : public Statement {
 public:
   AssignConfigVarStatement(Diagnostics &d,
-                           std::string config_var,
+                           Identifier *config_var,
                            Expression *expr,
                            Location &&loc);
 
-  std::string config_var;
+  Identifier *config_var = nullptr;
```
amscanne, line 384: Just as aside: we do have a weird habit of providing default values for these things, despite them not being optional nor having a default constructor. I'm not saying you should do anything different here, but it's been on my list for a long time because it creates ambiguity.

src/ast/passes/config_analyser.cpp
```diff
@@ -165,26 +166,33 @@ void ConfigAnalyser::visit(StackMode &mode)
...
   }
 
   auto configKey = maybeConfigKey.value();
 
   std::visit([&](auto key) { set_config(assignment, key); }, configKey);
+
+  if (bpftrace_.config_->is_unstable(configKey)) {
+    assignment.config_var->addWarning()
+        << "Script is using an unstable feature: "
+        << raw_ident.erase(0, UNSTABLE_PREFIX.size());
```
amscanne, line 194: Why is the logic for `is_unstable` split across translation units like this? I would either make the analyzer aware of these things, or limit it to the config file itself. The export of `UNSTABLE_PREFIX` is the part that I'm complaining about here, since it seems firmly in the "implementation detail" realm, with `is_unstable` being a public API.

jordalgo: Yeah maybe I'll just remove the prefix erasing here.

src/config.cpp
```diff
@@ -120,6 +120,23 @@ std::optional<ConfigKey> Config::get_config_key(const std::string &str,
...
+bool Config::is_unstable(ConfigKey key)
+{
+  static const std::map<ConfigKey, std::string> reversed_map =
+      reverse_config_map();
+  const std::string &config_str = reversed_map.at(key);
+  if (config_str.find(UNSTABLE_PREFIX) != std::string::npos) {
+    try {
+      // This could be updated to not only require bool keys
+      auto bool_key = std::get<ConfigKeyBool>(key);
+      return get(bool_key);
```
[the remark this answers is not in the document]
jordalgo: > Honestly, just allow any config key with unstable_ as a prefix to be reported by convention.

That's probably simpler but it feels weird to issue a warning if the flag is set to 0, which I admit is probably rare to never.

amscanne (approved):

danobi (approved):

PR #3932, opened 2025-03-24
ast: Add basic linking pass

amscanne:
Stacked PRs:
 * __->__#3932
 * #3931

--- --- ---

### ast: Add basic linking pass

### DIFF
--- a/src/ast/passes/link.cpp
+++ b/src/ast/passes/link.cpp
@@ -1,0 +1,91 @@
+#include <cstring>
+#include <fstream>
+
+#include "ast/passes/codegen_llvm.h"
+#include "ast/passes/link.h"
+#include "bpfbytecode.h"
+#include "scopeguard.h"
+#include "util/temp.h"
+
+namespace bpftrace::ast {
+
+char LinkError::ID;
+void LinkError::log(llvm::raw_ostream &OS) const
+{
+  OS << "linking " << origin_ << ": " << strerror(err_);
+}
+
+Pass CreateExternObjectPass()
+{
+  return Pass::create("extern", []() { return BpfExternObjects(); });
+}
+
+Pass CreateLinkPass()
+{
+  return Pass::create(
+      "link", [](BpfObject &obj, BpfExternObjects &ext) -> Result<BpfBytecode> {
+        // If there are no other objects to link, then just return our own.
+        if (ext.objects.empty()) {
+          return BpfBytecode{ obj.data };
+        }
+
+        // Otherwise, dump the intermediate object.
+        auto object = util::TempFile::create();
+        auto ok = object->write_all(obj.data);
+        if (!ok) {
+          return ok.takeError();
+        }
+
+        // Create an output file on disk. In the future, we may want to accept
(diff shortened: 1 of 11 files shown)


src/ast/passes/link.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+namespace bpftrace::ast {
+
+char LinkError::ID;
+void LinkError::log(llvm::raw_ostream &OS) const
+{
+  OS << "linking " << origin_ << ": " << strerror(err_);
+}
+
+Pass CreateExternObjectPass()
```
jordalgo, line 17: Why does this pass live here?

src/ast/passes/link.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        // Link in our own program.
+        int rc = bpf_linker__add_file(linker, object->path().c_str(), nullptr);
+        if (rc != 0) {
+          int err = errno;
+          bpf_linker__free(linker);
+          return make_error<LinkError>(output->path().string(), err);
+        }
+
+        // Next, we iterate through the link of link targets that we collected
```
jordalgo, line 57:
```suggestion
        // Next, we iterate through the list of link targets that we collected
```

src/ast/passes/link.cpp
amscanne: It is a placeholder. This will be exported in the future by the pass that resolves imports.

src/ast/passes/link.cpp
amscanne: Done.

jordalgo (approved):

src/ast/passes/link.cpp
```diff
@@ -0,0 +1,88 @@
...
+        // Create an output file on disk. In the future, we may want to accept
+        // some flags that allow this file to persist.
+        auto output = util::TempFile::create();
+
+        // In order to craft the final output, since we may have external maps
+        // and probes, we delegate the heavy lifting to libbpf. First, we open a
+        // new memfd as output, and add our top-level output to the linker.
+        struct bpf_linker *linker = bpf_linker__new(output->path().c_str(),
+                                                    nullptr);
+        assert(linker != nullptr);
```
danobi, line 47: Two things:
1. I think Alastair added in-memory support to avoid on-disk stuff here: https://github.com/libbpf/libbpf/commit/d2f1f4490ba4a95efbeb6c62a632aef59a94437d
2. Why assert and not error check? There seems to be a lot going on in libbpf construction side.

src/ast/passes/link.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+        // Link in our own program.
+        int rc = bpf_linker__add_file(linker, object->path().c_str(), nullptr);
+        if (rc != 0) {
+          int err = errno;
+          bpf_linker__free(linker);
```
danobi, line 53: Use `SCOPE_EXIT` immediately after successful construction instead?

amscanne: Done.

danobi (approved):

Issue #3898, opened 2025-03-14
runqlen.bt: struct cfs_rq field `nr_running` renamed to `nr_queued`


Avinesh:
`struct cfs_rq` field `nr_running` has been renamed to `nr_queued` in kernel v6.14-rc1
(https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=736c55a02c477ad31c57ae4c69130f437855e051)

which is causing the runqlen.bt to fail with - 
`ERROR: Struct/union of type 'struct cfs_rq' does not contain a field named 'nr_running'`

I want to ask how such kernel changes are handled?

jordalgo:
> I want to ask how such kernel changes are handled?

Sadly, I don't think we have a good way at the moment. There was some discussion about being able to provide a builtin or ifdef to detect the kernel version and do different things in the script but none of that has been fleshed out yet. For this case, I wonder if something like `has_field` could work.

viktormalik:
> > I want to ask how such kernel changes are handled?

What we did in the past for these cases is that we created a copy of the tool with the new field and copied the old version into the `tools/old/` directory.

> For this case, I wonder if something like `has_field` could work.

Yeah, I think it would. We should be able to determine whether the field exists during bpfscript compilation from BTF.

viktormalik:
Fixed by #3934.

PR #3934, opened 2025-03-25
Adapt to the renaming of nr_running to nr_queued in `struct cfs_rq`

shunghsiyu:
With ["sched/fair: Fix statistics with delayed dequeue" series](https://lore.kernel.org/all/20241202174606.4074512-1-vincent.guittot@linaro.org/) (more specifically torvalds/linux@736c55a02c47) several fields in "struct cfs_rq" is renamed, leading to runqlen.bt to fail at runtime with the following error:

```
  ERROR: Struct/union of type 'struct cfs_rq' does not contain a field named 'nr_running'
      $len = $my_q->nr_running;
```

Move existing runqlen script to tools/old/, and create a copy of the original with with nr_running changed to nr_queued in its place.

While at it, also touch up the comment regarding BTF.

This address issue report by @Avinesh in #3898.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,26 +41,33 @@ jobs:
         - NAME: LLVM 16
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm16
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: LLVM 17
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm17
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: LLVM 18
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm18
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: LLVM 19
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm19
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: LLVM 20 Release
           CMAKE_BUILD_TYPE: Release
           NIX_TARGET: .#bpftrace-llvm20
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: LLVM 20 Debug
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm20
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: LLVM 20 Clang Debug
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm20
           CC: clang
           CXX: clang++
+          TOOLS_TEST_OLDVERSION: runqlen.bt
         - NAME: Fuzzing
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-fuzz
@@ -70,6 +77,7 @@ jobs:
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm20
           NIX_TARGET_KERNEL: .#kernel-6_12
+          TOOLS_TEST_OLDVERSION: runqlen.bt # Need 6.14 kernel
         - NAME: AOT (LLVM 20 Debug)
           CMAKE_BUILD_TYPE: Debug
           NIX_TARGET: .#bpftrace-llvm20
--- a/tools/old/runqlen.bt
+++ b/tools/old/runqlen.bt
@@ -5,6 +5,8 @@
  *
  * This is a bpftrace version of the bcc tool of the same name.
  *
+ * For Linux < 6.14.
+ *
  * Copyright 2018 Netflix, Inc.
  * Licensed under the Apache License, Version 2.0 (the "License")
  *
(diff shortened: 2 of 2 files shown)


shunghsiyu:
Looks like I might need to add `TOOLS_TEST_OLDVERSION` for runqlen,

viktormalik:
> Looks like I might need to add `TOOLS_TEST_OLDVERSION` for runqlen, not sure if I need to drop the `pre-6.14` postfix to get it work though.

Yeah, you need to drop the suffix, it must be called `tools/old/runqlen.bt` for the script to pick it up.

tests/runtime/call
```diff
@@ -661,13 +661,13 @@ PROG BEGIN { if (*percpu_kaddr("process_counts") == *percpu_kaddr("process_count
 EXPECT SUCCESS
 
 NAME percpu_kaddr field access
-PROG BEGIN { $runq = ((struct rq *)percpu_kaddr("runqueues", 0)); if ($runq != 0) { printf("nr_running: %d\n", $runq->nr_running) } exit() }
-EXPECT_REGEX nr_running: [0-9]+
+PROG BEGIN { $runq = ((struct rq *)percpu_kaddr("runqueues", 0)); if ($runq != 0) { printf("nr_queued: %d\n", $runq->nr_queued) } exit() }
```
viktormalik, line 664: GHA runners certainly don't have kernels with this change at the moment. Ideally, we should change this test to use some different field of `runqueues` which is stable across kernels.

tests/runtime/call
```diff
@@ -661,13 +661,13 @@ PROG BEGIN { if (*percpu_kaddr("process_counts") == *percpu_kaddr("process_count
...
+EXPECT_REGEX nr_queued: [0-9]+
 
 NAME percpu_kaddr field access no NULL check
-PROG BEGIN { $runq = ((struct rq *)percpu_kaddr("runqueues", 0)); printf("nr_running: %d\n", $runq->nr_running); exit() }
+PROG BEGIN { $runq = ((struct rq *)percpu_kaddr("runqueues", 0)); printf("nr_queued: %d\n", $runq->nr_queued); exit() }
```
viktormalik, line 668: Same as above.

tools/old/runqlen-pre-6.14.bt
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bpftrace
+/*
+ * runqlen.bt	CPU scheduler run queue length as a histogram.
+ *		For Linux, uses bpftrace, eBPF.
+ *
+ * This is a bpftrace version of the bcc tool of the same name.
+ *
+ * For Linux <= 6.14.
```
viktormalik, line 8: Shouldn't be?
```suggestion
 * For Linux < 6.14.
```
The change appears in 6.14.

tools/runqlen.bt
```diff
@@ -14,13 +14,13 @@
 #ifndef BPFTRACE_HAVE_BTF
 #include <linux/sched.h>
 
-// Until BTF is available, we'll need to declare some of this struct manually,
+// Without BTF, we'll need to declare some of this struct manually,
```
viktormalik, line 17: Nit: we could also do this change in the old version.

man/adoc/bpftrace.adoc
```diff
@@ -2610,7 +2610,7 @@ be rejected.
 interval:s:1 {
   $runqueues = (struct rq *)percpu_kaddr("runqueues", 0);
   if ($runqueues != 0) {         // The check is mandatory here
-    print($runqueues->nr_running);
+    print($runqueues->nr_queued);
```
viktormalik, line 2613: Same as for the tests - it would be good to use a different field here, if possible, so that the example works across different kernels.

tools/old/runqlen-pre-6.14.bt
shunghsiyu: Absolutely, thanks for spotting that.

viktormalik (approved):
LGTM, thanks!

The old field is still used in the tests and the manual, right? Would you mind updating them within this PR? I think just changing to a different field from `runqueues` would suffice.

But I'm not going to block on this and we can do it in a follow-up PR.

tests/runtime/call
shunghsiyu: My bad, I confused `struct cfs_rq` with `struct rq` (my lack of scheduler knowledge is evident here), the latter do still have the `nr_running`.

Those can be left as-is.

viktormalik:
@jordalgo since this patch is not in any release, most distros with the latest kernel (e.g. Fedora) currently ship a broken version of runqlen. Should we craft a new patch release to fix that?

jordalgo:
@viktormalik We can. Are there any other fixes/patches we can include while we're at it? 
Maybe these:
https://github.com/bpftrace/bpftrace/issues/4131
https://github.com/bpftrace/bpftrace/issues/4111
https://github.com/bpftrace/bpftrace/issues/4104
https://github.com/bpftrace/bpftrace/pull/4100
https://github.com/bpftrace/bpftrace/pull/4074

viktormalik:
I would certainly include #4100 and #4074 as those fix bugs which started appearing on 0.23.0.

The rest look like the bugs have been there for a long time so it's IMHO ok to postpone them for the next release.

jordalgo:
@viktormalik Cool. I'll draft a PR for the release branch

viktormalik:
Thanks! I think we can do a single PR for all the backported commits and the release commit.

jordalgo:
https://github.com/bpftrace/bpftrace/pull/4177

PR #3935, opened 2025-03-25
v0.23.0 release

viktormalik:
##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.23.0] 2025-03-25
 
 #### Breaking Changes
 - Remove '-kk' command line opt, surface some BPF errors by default, and make '-k' surface probe read errors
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,18 @@ and this project adheres to
 
 ## Unreleased
 
+#### Breaking Changes
```
jordalgo, line 11: I'm confused. Why is this going into the release branch?

viktormalik: Because we may still backport important fixes to the branch and we will want to mark them in the changelog.

But I'm also ok with omitting it for now and adding the titles when necessary.

jordalgo: But then wouldn't those fixes/additions go into the "0.23.0" section? There shouldn't be an "Unreleased" section at all for the release branch as it's only for master right?

viktormalik: No, they will initially go to the "Unreleased" section which will then become "0.23.1" once a new patch release is done.

viktormalik: Of course we can directly create the "0.23.1" section once we find out that there are patches that need backporting.

viktormalik: Ok, makes sense. I dropped this part, then.

jordalgo (approved):

PR #3943, opened 2025-03-26
Fix iter probe listing with a wildcard

jordalgo:
args.listing wasn't being passed to the
attachpoint parser.

Issue: https://github.com/bpftrace/bpftrace/issues/3942

##### Checklist

### DIFF
--- a/src/ast/attachpoint_parser.h
+++ b/src/ast/attachpoint_parser.h
@@ -63,6 +63,6 @@ private:
 };
 
 // The attachpoints are expanded in their own separate pass.
-Pass CreateParseAttachpointsPass();
+Pass CreateParseAttachpointsPass(bool listing = false);
 
 } // namespace bpftrace::ast
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -867,9 +867,9 @@ int main(int argc, char* argv[])
     auto ok = ast::PassManager()
                   .put(ast)
                   .put(bpftrace)
-                  .add(ast::CreateParseAttachpointsPass())
+                  .add(ast::CreateParseAttachpointsPass(args.listing))
                   .add(CreateParseBTFPass())
-                  .add(ast::CreateSemanticPass(true))
+                  .add(ast::CreateSemanticPass(args.listing))
                   .run();
     if (!ok || !ast.diagnostics().ok()) {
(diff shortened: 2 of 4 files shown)


src/ast/passes/parser.h
```diff
@@ -14,20 +14,21 @@ namespace bpftrace::ast {
 // AllParsePasses returns a vector of passes representing all parser passes, in
 // the expected order. This should be used unless there's a reason not to.
 inline std::vector<Pass> AllParsePasses(
-    std::vector<std::string> &&extra_flags = {})
+    std::vector<std::string> &&extra_flags = {},
+    bool listing = false)
```
amscanne, line 18: I'm trying to remove "listing" as a parameter to the passes in general, as it's really a "please do a completely different thing" flag, rather than one that controls the behavior more narrowly.

I think you can just drop this from `AllParsePasses` and keep it, since the listing behavior is a manual construction in `main` anyways.

src/ast/attachpoint_parser.h
```diff
@@ -63,6 +63,6 @@ class AttachPointParser {
 };
 
 // The attachpoints are expanded in their own separate pass.
-Pass CreateParseAttachpointsPass();
+Pass CreateParseAttachpointsPass(bool listing = false);
```
amscanne, line 66: Maybe add a warning that `listing` will change the parsing semantics for attach points?

src/main.cpp
```diff
@@ -950,7 +950,7 @@ int main(int argc, char* argv[])
     }
   };
   // Start with all the basic parsing steps.
-  for (auto& pass : ast::AllParsePasses(std::move(flags))) {
+  for (auto& pass : ast::AllParsePasses(std::move(flags), args.listing)) {
```
amscanne, line 953: The use of `listing` is pretty confusing, because it's an entirely different op. I'd love for us to fix these abstractions and be able to list without the parsers.

As evidence of this confusion, the correct/previous behavior here is actually to pass false (which aligns with eliminating this parameter from AllParsePasses), because otherwise the semantics of parsing are changed and you can actually pass `iter:*` and match with full expansion against all iterators (:cry:).

amscanne (approved):

PR #3946, opened 2025-03-28
cleanup: remove unnecessary exceptions

amscanne:
##### Checklist

### DIFF
--- a/src/arch/s390.cpp
+++ b/src/arch/s390.cpp
@@ -1,9 +1,8 @@
-#include "arch.h"
-#include "util/exceptions.h"
-
 #include <array>
 #include <set>
 #include <vector>
+
+#include "arch.h"
 
 #define ARG_REGISTERS 5
 // For s390x, r2-r6 registers are used as function arguments, then the extra
@@ -90,10 +89,9 @@ std::string name()
   return std::string("s390x");
 }
 
-std::vector<std::string> invalid_watchpoint_modes()
+const std::unordered_set<std::string> &watchpoint_modes()
 {
-  throw util::FatalUserException(
-      "Watchpoints are not supported on this architecture");
+  return {}; // Not supported.
 }
 
 int get_kernel_ptr_width()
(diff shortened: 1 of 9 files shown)


src/arch/s390.cpp
```diff
@@ -90,10 +89,9 @@ std::string name()
   return std::string("s390x");
 }
 
-std::vector<std::string> invalid_watchpoint_modes()
+bool is_watchpoint_mode_valid(const std::string &mode)
 {
-  throw FatalUserException(
```
jordalgo, line 95: Not that this was the correct place for this but should we add a check in one of the passes to error that watchpoints are not supported on X architecture? It seems users will now get an error "invalid watchpoint mode", which is not correct.

amscanne: I can restructure further to return the list of valid watchpoints, and add a hint.

jordalgo (dismissed):
If you want to be super duper awesome you'll add some unit tests (but I know that might be a pain to mock the arch utility so won't block).

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3598,11 +3598,28 @@ void SemanticAnalyser::visit(AttachPoint &ap)
...
+        hint << "supported modes: ";
+        bool first = true;
+        for (const auto &mode : modes) {
+          if (!first) {
+            hint << ", " << mode;
+          } else {
+            hint << mode;
+            first = false;
+          }
+        }
```
viktormalik, line 3620: We have `utils::str_join`, can we use it?

viktormalik (commented):
LGMT.

One suggestion and also I'd love to see at least some explanation of the change in the commit message but I'm not going to block.

src/ast/passes/semantic_analyser.cpp
amscanne: Using it now. Although it hurts my soul a bit to see an `O(n^2)` string join :) We should fix that separately

amscanne:
> LGMT.
> 
> One suggestion and also I'd love to see at least some explanation of the change in the commit message but I'm not going to block.

Done

jordalgo (approved):

PR #3948, opened 2025-03-29
Simplify hashing multiple fields

danobi:
No need to manually swizzle hashes. Just use the existing util::hash_combine() we have.

Also fix hash.h to have the pragma once.

##### Checklist

### DIFF
--- a/src/functions.h
+++ b/src/functions.h
@@ -9,6 +9,7 @@
 #include "ast/ast.h"
 #include "ast/location.h"
 #include "types.h"
+#include "util/hash.h"
 
 namespace bpftrace {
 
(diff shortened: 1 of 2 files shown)


src/functions.h
```diff
@@ -127,8 +128,10 @@ class FunctionRegistry {
   public:
     size_t operator()(const FqName &fq_name) const
     {
-      return std::hash<std::string>()(fq_name.ns) ^
-             std::hash<std::string>()(fq_name.name);
+      std::size_t seed = 0;
```
jordalgo, line 131: maybe this is better if the starting seed is the length of ns or name.

danobi: hmm, that's not a pattern I've typically seen. Given that this is not security sensitive (only used for hash table insertion, where collisions are properly handled), not sure I see the value of doing anything non-obvious.

maybe there are security implications? rather just cargo cult it tbh

jordalgo: kk, I think just looking at this part of `hash_combine` made me nervous:
```
(seed << 6) + (seed >> 2);
```
but you're right I think it's fine for this.

jordalgo (approved):

PR #3949, opened 2025-03-29
tests: config: Remove unnecessary headers

danobi:
Config analyser tests only use bare minimum passes now. So these headers are no longer used.

##### Checklist


jordalgo:
Flaky test on the latest kernel run - but I haven't seen it before: "watchpoint - absolute address"

jordalgo (approved):

PR #3954, opened 2025-03-31
[BACKPORT] nix: Fix vmtest for aarch64

danobi:
Previously, we had one sha for both the x86-64 and aarch64 vmtest binaries. Obviously one hash cannot work for both. So add a hash for each arch.

##### Checklist


danobi:
This should fix the flakehub push. Hopefully, as I don't see a good way to test a flakehub push.

jordalgo (approved):

PR #3955, opened 2025-03-31
docs: Update CONTRIBUTING.md to use `unstable_` prefix

danobi:
Looks like we're using `unstable_` prefix for map declarations, so update the docs to it.

##### Checklist


jordalgo (approved):
🍰

PR #3958, opened 2025-04-01
cleanup: Remove unused `std::hash` for `SizedType`

amscanne:
Stacked PRs:
 * #3962
 * #3960
 * #3959
 * __->__#3958
 * #3964

--- --- ---

### cleanup: Remove unused `std::hash` for `SizedType`

Given that `SizedType` is a rather complex structure, and involves
references to other structures, it would be better to not have this to
hashable and instead treat it as a value type. Since this is no longer
generally required for anonymous structures in the `StructRegistry`, we
can drop the hashing support.


jordalgo (approved):
Okey dokey

danobi (approved):

PR #3959, opened 2025-04-01
cleanup: simplify SemanticAnalyzer treatment of maps

amscanne:
Stacked PRs:
 * #3962
 * #3960
 * __->__#3959
 * #3958
 * #3964

--- --- ---

### cleanup: simplify SemanticAnalyzer treatment of maps

This change originally began as a way to avoid generating spurious
types, and prevent the failure of `GetTupleCnt` tests. However, this was
removed in an earlier change in this stack since the complexity of
juggling the anonymous type registry didn't seem worth it.

Rather than throwing out this change, it has been adapted to be a
general improvement on the SemanticAnalyzer's treatment of maps: instead
of setting look-aside failures on the AST that changes treatment of the
`Map` values (or lookaside values that record the parameter index, which
then changes how the visit works!), this logic is encoded directly in
the visit for the map assignments, for loops and calls.


amscanne:
I am going to wait until #3988 is in, and drop this. The only approach achieves the same thing, and I think it is the better long-term evolution.

amscanne:
Closing per previous comment.

PR #3960, opened 2025-04-01
cleanup: remove `tp_args_struct_level`

amscanne:
cleanup: remove `tp_args_struct_level`

This field is used to determine the maximium depth to traverse when
resolving types via clang. However, an arbitrary depth is easily
achievable via cast operations or other mechanisms (such as using
`current`). This also adds complexity to the type resolution, which can
use a simple convergence check instead of a fixed limit on iterations.


danobi (approved):
not an expert at the type resolution bits but seems correct to me

amscanne:
It was added in #1571 purely as an optimization. I think we can limit it just by checking if things don't converge.

amscanne:
@viktormalik Any concerns? Since most of these types are coming from BTF, I'm trying to clean stuff around the types to support the interop efforts and converge to BTF.

amscanne:
> @viktormalik Any concerns? Since most of these types are coming from BTF, I'm trying to clean stuff around the types to support the interop efforts and converge to BTF.

Just shout if there's something objectionable or something missed. Happy to revert/rollback later.

PR #3961, opened 2025-04-01
cleanup: simplify AST by introducing `PositionalParameterCount`

amscanne:
Stacked PRs:
 * #3962
 * #3960
 * #3959
 * #3958
 * #3964
 * __->__#3961

--- --- ---

### cleanup: simplify AST by introducing `PositionalParameterCount`

The current class is essentially a switch on the type, which fails to
capture the value of having different types and a structured visitor.
The code structure is simplified by using a proper type for this case.


jordalgo (approved):
I like it!

PR #3962, opened 2025-04-01
cleanup: split `types.h` into logical subfiles

amscanne:
cleanup: split `types.h` into logical subfiles

Similar to utilities, the header had likely grown over time organically
and become a bit of a kitchener sink. This commit splits this header
into logical locations, including new `probe_types.(cpp|h)` files.


src/bpftrace.cpp
```diff
@@ -427,16 +429,21 @@ void perf_event_printer(void *cb_cookie, void *data, int size)
...
 
-  bpftrace->out_->message(MessageType::printf,
-                          fmt.format_str(arg_values),
-                          false);
+    bpftrace->out_->message(MessageType::printf,
+                            fmt.format_str(arg_values),
+                            false);
+    return;
+  } else {
+    LOG(WARNING) << "Unknown printf_id: " << static_cast<int64_t>(printf_id);
```
danobi, line 445: `LOG(BUG)` instead? Or at least warn that an event is dropped here?

danobi (approved):

viktormalik (approved):
Nice cleanup!

amscanne:
Sorry, I missed that this was approved. I will rebase and merge.

src/bpftrace.cpp
amscanne: Sure, `LOG(BUG)` works.

PR #3963, opened 2025-04-01
util: move `operator<<` for `llvm::Error` into llvm namespace

amscanne:
The resolution rules for `operator<<` means that it needs to be either in the same namespace as the user or in the same namespace as the type in order to be found reliably. We can simply put this into the `llvm` namespace.

##### Checklist

### DIFF
--- a/src/util/result.cpp
+++ b/src/util/result.cpp
@@ -2,7 +2,7 @@
 
 #include "util/result.h"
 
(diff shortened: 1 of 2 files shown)


jordalgo (approved):
Makes sense

danobi (approved):

PR #3964, opened 2025-04-01
cleanup: simplify anonymous type plumbing

amscanne:
Stacked PRs:
 * #3962
 * #3960
 * #3959
 * #3958
 * __->__#3964

--- --- ---

### cleanup: simplify anonymous type plumbing

Currently, a `weak_ptr` is used for type references internally. The
canonical ownership rests with the `StructManager`. This works well
enough for types that may be recursively-defined, e.g. those that
include pointers to themselves.

For anonymous types, this adds additional complexity without much
benefit. Primarily, there are two issues this commit aims to address:

* The anonymous types must be dedupped and stored centrally, even though
  they are not addressable. Although this may seem useful, in order to
  do this dedupping the objects need to be fully allocated and
  constructed anyways (and then compared against the others in the set),
  so there is actually *extra* work. A small amout of memory may be
  saved, but unreferenced anonymous types can also pile up in the
  central registry (because these tend to go through adjustment, based
  on string sizes, etc.).

* The ambiguity around the validity of `weak_ptr` leaks out from the
  SizedType and into other code.

For simplicity, the `StructManager` will continue to return `weak_ptr`
objects, but hide the details of ownership inside the `SizedType`.
Anonymous types are owned directly and will be freed directly, with the
pointers owned by `SizedType`. The methods on `SizedType` return
`std::shared_ptr` for clarity and convenience. There is no possibility
of cycles, because there is no way to look up a reference to an
anonymous type.


src/types.h
```diff
@@ -149,16 +151,19 @@ class SizedType {
   std::shared_ptr<SizedType> element_type_; // for "container" and pointer
                                             // (like) types
   std::string name_; // name of this type, for named types like struct and enum
-  std::weak_ptr<Struct> inner_struct_; // inner struct for records and tuples
-                                       // the actual Struct object is owned by
-                                       // StructManager
+  std::variant<std::shared_ptr<Struct>, std::weak_ptr<Struct>>
+      inner_struct_; // inner struct for records and tuples: if a shared_ptr, it
```
jordalgo, line 155: This is a comment I like 👍

jordalgo (dismissed):
Nice optimization. I don't quite understand the initial motivation for deduping anon types in the StructManager - maybe so they'd all be in one place but not sure what the benefit of that is (as pointed out by your commit msg).

amscanne:
> Nice optimization. I don't quite understand the initial motivation for deduping anon types in the StructManager - maybe so they'd all be in one place but not sure what the benefit of that is (as pointed out by your commit msg).

For a bit of extra context: I'm doing this to eliminate the need to hash types and put them in the anonymous set. Obviously this won't work with an `AbstractType` class that can have different implementations. It is a benefit that we get to delete some code and complexity. (Although I thought it would be more initially, and the `variant` stuff is a bit more complicated than I like. On balance, I think it's probably a reasonable simplification, at least for anonymous types.)

jordalgo (approved):

PR #3965, opened 2025-04-01
Fix double arg listing for rawtracepoints

jordalgo:
Because we're checking two FUNC_PROTO prefixes
for rawtracepoints, we need to make sure we only
use one. So check for duplicates when getting BTF
params.

Before:
```
rawtracepoint:vmlinux:consume_skb
  struct sk_buff * skb
  void * location
  void * __data
  struct sk_buff * skb
  void * location
  int retval
```

After:
```
rawtracepoint:vmlinux:consume_skb
  struct sk_buff * skb
  void * location
```

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -656,7 +656,7 @@ std::map<std::string, std::vector<std::string>> BTF::get_params_from_btf(
       func_name = btf_obj.name + ":" + str;
     }
 
-    if (!funcs.contains(func_name))
+    if (!funcs.contains(func_name) || params.contains(func_name))
       continue;
 
     t = btf__type_by_id(btf_obj.btf, t->type);
@@ -684,7 +684,7 @@ std::map<std::string, std::vector<std::string>> BTF::get_params_from_btf(
       params[func_name].push_back(type + " " + arg_name);
     }
 
-    if (!t->type)
+    if (!t->type || is_raw_tracepoint)
       continue;
 
     // set by dump_printf callback
--- a/tests/runtime/basic
+++ b/tests/runtime/basic
@@ -63,6 +63,16 @@ EXPECT_REGEX [ ]+[a-zA-Z_\*\s]+
 REQUIRES_FEATURE btf
 TIMEOUT 1
 
+# The BTF FUNC_PROTOs for rawtracepoint params have an unnecessary initial
+# void * and a retval that can't be used so make sure they are not
+# included in the listing output
+NAME it doesn't list the first rawtracepoint param or retval
+RUN {{BPFTRACE}} -lv "rawtracepoint:consume_skb"
+EXPECT_NONE void * __data
+EXPECT_NONE int retval
+REQUIRES_FEATURE btf
+TIMEOUT 1
+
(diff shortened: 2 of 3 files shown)


tests/runtime/basic
```diff
@@ -63,6 +63,13 @@ EXPECT_REGEX [ ]+[a-zA-Z_\*\s]+
 REQUIRES_FEATURE btf
 TIMEOUT 1
 
+NAME it doesn't list the first rawtracepoint param or retval
```
danobi, line 66: Not sure I understand. What is the full expected output here?

jordalgo: I didn't want to rely on the exact param output in case the API changes so instead I'm just asserting that the retval and void * (first param) are not listed.
Before this change this is what you would see:
```
rawtracepoint:vmlinux:consume_skb
    struct sk_buff * skb
    void * location
    void * __data
    struct sk_buff * skb
    void * location
    int retval
```
What you want to see is this:
```
rawtracepoint:vmlinux:consume_skb
    struct sk_buff * skb
    void * location
```

src/btf.cpp
```diff
@@ -684,7 +684,7 @@ std::map<std::string, std::vector<std::string>> BTF::get_params_from_btf(
       params[func_name].push_back(type + " " + arg_name);
     }
 
-    if (!t->type)
+    if (!t->type || is_raw_tracepoint)
```
amscanne, line 687: As an aside, the `is_raw_tracepoint` parameter is a flag for me. You should just submit and get the bug fixed, but in terms of API design, this doesn't make sense. It's undocumented and leaking unexpected semantics across the object boundary. If possible, I would try to encode this in some way that makes sense within the class API, not that is meaningful to the caller, e.g. `omit_retval` or something. (Otherwise the reader of this function needs to work way, way harder to understand if the behavior is correct, if they are able to at all.)

(I realize that it's also used above, but you could provide separate parameters, e.g. one could also provide an optional function that maps from the BTF function name to `func_name`, although I'm not sure why you wouldn't do this from the caller?)

amscanne (approved):

src/btf.cpp
jordalgo: @amscanne You mean **red** flag 😸  - but yes, I agree - I think I had it this way originally and I changed it back for some reason - let me fix that in a follow-up PR.

tests/runtime/basic
viktormalik: It should probably go to the comment for this test case in some form.

viktormalik (changes requested):
The change looks ok but is desperately undocumented.

At the minimum, an explanation why this is needed should be in the commit message. Ideally, a comment why special-casing for raw tracepoints is needed in btf.cpp but that can be postponed to the follow-up PR which will replace `is_tracepoint`.

jordalgo:
> The change looks ok but is desperately undocumented.

My apologies. Updated both the commit message and added a comment to the run time test.

viktormalik (approved):
Thanks!

PR #3966, opened 2025-04-01
Remove unneeded BTF funcs

jordalgo:
These BTF funcs aren't needed, cause additional
processing, and cause confusion.

Before BTF is loaded we read
/sys/kernel/tracing/available_filter_functions

And match them against the probes in the
bpftrace program to determine what modules
we need to load and parse for BTF.

However, later on when other parts of bpftrace
call `get_matches_for_probetype` we then
utilize the BTF we loaded to AGAIN search
for the same set of functions instead of just
relying on the ones we already used for module
loading.

Additionally, in the func that's removed here
`get_all_funcs_from_btf` there is a call to
`is_traceable_func` which utilizes the same
list from the "available_filter_functions" file.

##### Checklist


viktormalik (changes requested):
I'm all in for removing these functions or otherwise simplifying this code path but IIRC the problem here is that for a function to be traceable with fentry/fexit, it needs to fulfill 2 conditions: (1) have a BTF entry and (2) be in `available_filter_functions`.

By removing these functions and using, we're effectively changing the set of functions that *appear* to be traceable from bpftrace:
```
// On master
# bpftrace -l 'fentry:*' | wc -l
83841

// On this PR
# bpftrace -l 'fentry:*' | wc -l
87774
```

I didn't try it but I think that those extra functions will fail to attach.

That's the reason why we had two functions:
- `get_symbols_from_traceable_funcs` is an imprecise heuristics (b/c it doesn't cross-check with BTF) but it is good enough to get us the list of BTF modules to parse,
- `BTF::get_all_funcs` gets the actual list of traceable functions by iterating BTF and cross-checking it with `available_filter_functions` (hence the "extra" `is_traceable_func` call).

I would love to simplify/optimize it somehow but so far I didn't figure out how.

jordalgo:
I think you're right. Oh well, it would have been nice to remove these but doesn't look like we should. Thanks for checking @viktormalik

PR #3967, opened 2025-04-02
BTF cleanup

jordalgo:
Combination of 7 clean-up commits.

- [Clean up get_params APIs in BTF class](https://github.com/bpftrace/bpftrace/pull/3967/commits/0251759d3a9e5b8fa930214304dfa238a2327152)
- [Reuse FuncParamLists alias in BTF class](https://github.com/bpftrace/bpftrace/pull/3967/commits/81cde5475038d70c335d3a9584bba48a506f8145)
- [Add helpful comment in get_matches_for_probetype](https://github.com/bpftrace/bpftrace/pull/3967/commits/a78ecaf4fa24511af66ede73cc54f4607b8bc2d7)
- [[BTF] cache calls to get_all funcs](https://github.com/bpftrace/bpftrace/pull/3967/commits/50844e2cca81b39fa03bf63bea1419302508d98b)
- [Don't re-parse BTF](https://github.com/bpftrace/bpftrace/pull/3967/commits/0014d53a434a00bb206bd1fa7189ebc73129a8a0)
- [Remove unused 'id' field from BTFObj](https://github.com/bpftrace/bpftrace/pull/3967/commits/8d815ccb9126603f2814817aca4d801e237193e7)
- [Pass callback to get_params_impl instead of bool](https://github.com/bpftrace/bpftrace/pull/3967/commits/b410d089346748c54748ac9885fd60e853636fd3)

##### Checklist

### DIFF
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -48,8 +48,7 @@ BTF::BTF(const std::set<std::string> &modules)
   // Try to get BTF file from BPFTRACE_BTF env
   char *path = std::getenv("BPFTRACE_BTF");
   if (path) {
-    btf_objects.push_back(
-        BTFObj{ .btf = btf__parse_raw(path), .id = 0, .name = "" });
+    btf_objects.push_back(BTFObj{ .btf = btf__parse_raw(path), .name = "" });
     vmlinux_btf = btf_objects.back().btf;
     if (!vmlinux_btf) {
       LOG(WARNING) << "BTF: failed to parse BTF from " << path;
@@ -81,8 +80,7 @@ void BTF::load_kernel_btfs(const std::set<std::string> &modules)
     LOG(V1) << "BTF: failed to find BTF data for vmlinux: " << strerror(errno);
     return;
   }
-  btf_objects.push_back(
-      BTFObj{ .btf = vmlinux_btf, .id = 0, .name = "vmlinux" });
+  btf_objects.push_back(BTFObj{ .btf = vmlinux_btf, .name = "vmlinux" });
 
   if (bpftrace_ && !bpftrace_->feature_->has_module_btf())
     return;
@@ -124,12 +122,9 @@ void BTF::load_kernel_btfs(const std::set<std::string> &modules)
     if (!info.kernel_btf)
       continue;
 
-    if (mod_name == "vmlinux") {
-      btf_objects.front().id = id;
-    } else if (modules.contains(mod_name)) {
+    if (modules.contains(mod_name)) {
       btf_objects.push_back(
           BTFObj{ .btf = btf__load_from_kernel_by_id_split(id, vmlinux_btf),
-                  .id = id,
                   .name = mod_name });
     }
   }
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -611,7 +611,7 @@ std::unique_ptr<std::istream> BTF::get_all_raw_tracepoints() const
   return funcs;
 }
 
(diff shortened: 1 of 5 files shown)


jordalgo:
I can separate this PR into two if needed. Unfortunately the commit "Have bpffeature use bpftrace's BTF object" turned out to be a real pain, exposing a lot of weird issues.

src/btf.h
```diff
@@ -124,10 +122,13 @@ class BTF {
                                  std::unordered_set<std::string>& types) const;
   std::string get_all_funcs_from_btf(const BTFObj& btf_obj) const;
   std::string get_all_raw_tracepoints_from_btf(const BTFObj& btf_obj) const;
-  std::map<std::string, std::vector<std::string>> get_params_from_btf(
+  FuncParamLists get_params_impl(const std::set<std::string>& funcs,
+                                 bool is_rawtracepoints) const;
```
viktormalik, line 126: I don't want to over-complicate things but how about we pass a callback (either `get_params` or `get_rawtracepoint_params`) instead of `is_tracepoints` here?

src/btf.cpp
```diff
@@ -707,11 +693,91 @@ std::map<std::string, std::vector<std::string>> BTF::get_params_from_btf(
...
+  dump = dump_new(btf_obj.btf, dump_printf, &type);
+  err = libbpf_get_error(dump);
+  if (err) {
+    libbpf_strerror(err, err_buf, sizeof(err_buf));
+    LOG(ERROR) << "BTF: failed to initialize dump (" << err_buf << ")";
+    return {};
+  }
+
+  FuncParamLists params;
+  for (id = start_id(btf_obj.btf); id <= max; id++) {
```
viktormalik, line 715: The entire BTF iteration is cumbersome, it would be great if we could simplify it one day. Definitely out of the scope of this PR, though.

viktormalik (commented):
> I can separate this PR into two if needed. Unfortunately the commit "Have bpffeature use bpftrace's BTF object" turned out to be a real pain, exposing a lot of weird issues.

Yeah, the rest looks good (just one suggestion below) so I'd suggest drop that one for now and we can merge this as these are nice improvements. It will be also easier to review the problematic commit by itself.

jordalgo:
> so I'd suggest drop that one for now and we can merge this as these are nice improvements

Good idea. Removed.

src/btf.cpp
jordalgo: Yeah, it's crazy how many odd bugs and things I ran into trying to refactor BTF loading

src/btf.h
jordalgo: Good idea. Done.

src/btf.cpp
```diff
@@ -720,13 +788,31 @@ std::map<std::string, std::vector<std::string>> BTF::get_params(
...
 
   return params;
 }
 
+FuncParamLists BTF::get_params(const std::set<std::string> &funcs) const
+{
+  return get_params_impl(
+      funcs, [this](const BTFObj &btf_obj, const std::set<std::string> &funcs) {
+        return get_params_from_btf(btf_obj, funcs);
+      });
```
viktormalik, line 803: Doesn't this work?
```suggestion
  return get_params_impl(funcs, get_params_from_btf);
```

jordalgo: No. It's a non-static member function. I think the other option is to use `std::bind`

viktormalik (approved):
Nice cleanup!

PR #3968, opened 2025-04-02
nix: fix references to incorrect versions of clang

JakeHillion:
The Nix flake got updated a while ago and it seems the Nix develop
environments are now providing the wrong versions of clang and some of
the clang tools. Update to use the llvmVersion in more places.

v2: took clang tidy fixes with `./scripts/clang_tidy.sh -u` and fixed
them. It's selecting a different version of clang now.

v3: drop clang tidy fixes again. set the clang tidy script back to
llvm-19 which is what it was using accidentally before, and leave this
to be fixed properly in the future.

v4: remove duplicated version in mkBpftraceDevShell.

##### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -146,7 +146,7 @@
 
                 nativeBuildInputs = [
                   pkgs.bison
-                  pkgs.clang
+                  pkgs."llvmPackages_${toString llvmVersion}".clang
                   pkgs.cmake
                   pkgs.flex
                   pkgs.gcc
@@ -188,9 +188,12 @@
               };
 
           # Define lambda that returns a devShell derivation with extra test-required packages
-          # given the bpftrace package derivation as input
+          # given the bpftrace LLVM version as input
           mkBpftraceDevShell =
-            pkg:
+            llvmVersion:
+            let
+              pkg = self.packages.${system}."bpftrace-llvm${toString llvmVersion}";
+            in
               with pkgs;
               pkgs.mkShell {
                 buildInputs = [
@@ -198,7 +201,7 @@
                   binutils
                   bpftools
                   coreutils
-                  clang-tools  # Needed for the nix-aware "wrapped" clang-tidy
+                  pkgs."llvmPackages_${toString llvmVersion}".clang-tools # Needed for the nix-aware "wrapped" clang-tidy
                   gawk
                   git
                   gnugrep
@@ -206,7 +209,7 @@
                   iproute2
                   kmod
                   # For git-clang-format
-                  libclang.python
+                  pkgs."llvmPackages_${toString llvmVersion}".libclang.python
                   nftables
                   procps
                   python3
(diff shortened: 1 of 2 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -2040,7 +2044,7 @@ void SemanticAnalyser::visit(ArrayAccess &arr)
         auto index = bpftrace_.get_int_literal(arr.indexpr);
         if (index.has_value()) {
           size_t num = type.GetNumElements();
-          if (num != 0 && static_cast<size_t>(*index) >= num)
+          if (num != 0 && std::cmp_greater_equal(*index, num))
```
ttreyer, line 2047: This one is causing the issue with the `semantic_analyser.array_access` test: when `*index` is negative, the comparison always returns `false`, since `num` is unsigned.

src/btf.cpp
```diff
@@ -214,7 +215,7 @@ std::string BTF::dump_defs_from_btf(
     if (btf_is_enum(t)) {
       const struct btf_enum *p = btf_enum(t);
       uint16_t vlen = btf_vlen(t);
-      for (int e = 0; e < vlen; ++e, ++p) {
+      for (int e = 0; std::cmp_less(e, vlen); ++e, ++p) {
```
ttreyer, line 218: Could we change `e` to be a `uint16_t` instead?

ttreyer (changes requested):
Thanks for fixing Nix!

The `contains` cleanup is really nice 🙂 

The use of `std::cmp_*` has introduced a bug (see `SemAn::visit(ArrayAccess&)`), so I'll take some time tomorrow to review each of them more thoroughly.

Could you change the type of each loop index to match what they're compared with? Instead of using `std::cmp_*`.

scripts/clang_tidy.sh
```diff
@@ -16,7 +16,7 @@ SCRIPT_NAME=$0
 FIX=
 
 function run() {
-  nix develop --command "$@"
+  nix develop ".#bpftrace-llvm19" --command "$@"
```
danobi, line 19: Mind following up with clang-tidy fixes to drop this?

flake.nix
```diff
@@ -187,6 +187,7 @@
           # Define lambda that returns a devShell derivation with extra test-required packages
           # given the bpftrace package derivation as input
           mkBpftraceDevShell =
+            llvmVersion:
```
danobi, line 190: Try to drop `pkg` below as well? Can do fmt string selection using `llvmVersion`. Would be good to have non-duplicated params.

flake.nix
```diff
@@ -187,23 +187,26 @@
           # Define lambda that returns a devShell derivation with extra test-required packages
           # given the bpftrace package derivation as input
```
danobi, line 188:
```suggestion
          # given the bpftrace LLVM version as input
```

JakeHillion: Whoops, fixed!

scripts/clang_tidy.sh
JakeHillion: I can definitely take a look. The `readability-container-contains` fixes are fine and worth taking, but the `modernize-use-integer-sign-comparison` honestly seem wrong. For example:

    ap.address > static_cast<uint64_t>(arch::max_arg())

This current code is fine. `ap.address` is a `uint64_t`, and the cast on the right is nice and explicit. The recommended fix is:

    std::cmp_greater(ap.address, arch::max_arg())

This is supposed to always do the right thing, but it's different behaviour. I think it would be okay in new code but this isn't a clear fix in any case. What do you think? I'm a bit worried about taking them when not familiar with the code, though @ttreyer  is right that a lot of these go away if you correct the type of a loop variable.

danobi: Agreed that `std::cmp_greater()` stuff is much less readable. I think it'd be better to avoid the signed-unsigned comparison altogether if possible. And if this lint is too noisy, add it to the block list.

danobi (approved):
ty!

PR #3970, opened 2025-04-03
Have bpffeature use bpftrace's BTF object

jordalgo:
Instead of having bpffeature create it's own
BTF object and parsing vmlinux BTF again,
just pass bpftrace's BTF object when the bpftrace
object is created so vmlinux BTF is parsed
one less time.

This also separate BTF module loading
from the initial vmlinux BTF parsing
as that depends on the program probes.

There is also a commit to search all modules when listing types;
this is a drive-by fix I noticed when refactoring BTF.

##### Checklist

### DIFF
--- a/src/bpffeature.h
+++ b/src/bpffeature.h
@@ -70,7 +70,8 @@ protected:
 
 class BPFfeature {
 public:
-  BPFfeature(BPFnofeature& no_feature) : no_feature_(no_feature)
+  BPFfeature(BPFnofeature& no_feature, BTF& btf)
+      : no_feature_(no_feature), btf_(btf)
   {
   }
   BPFfeature() = default;
@@ -95,13 +96,13 @@ public:
   bool has_kprobe_multi();
   bool has_kprobe_session();
   bool has_uprobe_multi();
-  bool has_fentry();
   bool has_skb_output();
   bool has_prog_fentry();
   bool has_module_btf();
   bool has_iter(std::string name);
-
-  bool has_kernel_func(Kfunc kfunc);
+  // These are virtual so they can be overridden in tests by the mock
+  virtual bool has_fentry();
+  virtual bool has_kernel_func(Kfunc kfunc);
 
   std::string report();
 
@@ -165,9 +166,8 @@ private:
       int* outfd = nullptr);
   bool try_load_btf(const void* btf_data, size_t btf_size);
 
-  BTF btf_ = BTF({ "vmlinux" });
-
   BPFnofeature no_feature_;
+  BTF& btf_;
 };
 
 #undef DEFINE_PROG_TEST
--- a/src/btf.cpp
+++ b/src/btf.cpp
@@ -43,7 +43,11 @@ __s32 BTF::start_id(const struct btf *btf) const
   return btf == vmlinux_btf ? 1 : vmlinux_btf_size + 1;
 }
 
-BTF::BTF(const std::set<std::string> &modules)
+BTF::BTF() : BTF(nullptr)
+{
+}
+
+BTF::BTF(BPFtrace *bpftrace) : bpftrace_(bpftrace)
 {
   // Try to get BTF file from BPFTRACE_BTF env
   char *path = std::getenv("BPFTRACE_BTF");
@@ -54,8 +58,15 @@ BTF::BTF(const std::set<std::string> &modules)
       LOG(WARNING) << "BTF: failed to parse BTF from " << path;
       return;
     }
-  } else
-    load_kernel_btfs(modules);
+  } else {
+    vmlinux_btf = btf__load_vmlinux_btf();
+    if (libbpf_get_error(vmlinux_btf)) {
+      LOG(V1) << "BTF: failed to find BTF data for vmlinux: "
+              << strerror(errno);
+      return;
+    }
+    btf_objects.push_back(BTFObj{ .btf = vmlinux_btf, .name = "vmlinux" });
+  }
 
   if (btf_objects.empty()) {
     LOG(V1) << "BTF: failed to find BTF data";
@@ -73,16 +84,9 @@ BTF::~BTF()
     btf__free(btf_obj.btf);
 }
 
-void BTF::load_kernel_btfs(const std::set<std::string> &modules)
+void BTF::load_module_btfs(const std::set<std::string> &modules)
 {
-  vmlinux_btf = btf__load_vmlinux_btf();
-  if (libbpf_get_error(vmlinux_btf)) {
-    LOG(V1) << "BTF: failed to find BTF data for vmlinux: " << strerror(errno);
-    return;
-  }
-  btf_objects.push_back(BTFObj{ .btf = vmlinux_btf, .name = "vmlinux" });
-
-  if (bpftrace_ && !bpftrace_->feature_->has_module_btf())
+  if ((bpftrace_ && !bpftrace_->feature_->has_module_btf()) || state != OK)
     return;
 
   // Note that we cannot parse BTFs from /sys/kernel/btf/ as we need BTF object
@@ -128,6 +132,8 @@ void BTF::load_kernel_btfs(const std::set<std::string> &modules)
                   .name = mod_name });
     }
   }
+
+  state = OK_MODULES_LOADED;
 }
 
 static void dump_printf(void *ctx, const char *fmt, va_list args)
@@ -1073,7 +1079,7 @@ ast::Pass CreateParseBTFPass()
 {
   return ast::Pass::create(
       "btf", []([[maybe_unused]] ast::ASTContext &ast, BPFtrace &b) {
-        b.parse_btf(b.list_modules(ast));
+        b.parse_module_btf(b.list_modules(ast));
       });
 }
 
--- a/src/btf.h
+++ b/src/btf.h
@@ -53,6 +53,7 @@ class BTF {
   enum state {
     NODATA,
     OK,
+    OK_MODULES_LOADED,
   };
 
   // BTF object for vmlinux or a kernel module.
@@ -70,18 +71,17 @@ class BTF {
   };
 
 public:
-  BTF(const std::set<std::string>& modules);
-  BTF(BPFtrace* bpftrace, const std::set<std::string>& modules) : BTF(modules)
-  {
-    bpftrace_ = bpftrace;
-  };
+  BTF();
+  BTF(BPFtrace* bpftrace);
   ~BTF();
 
   bool has_data() const;
+  bool modules_loaded() const;
   size_t objects_cnt() const
   {
     return btf_objects.size();
   }
+  void load_module_btfs(const std::set<std::string>& modules);
   std::string c_def(const std::unordered_set<std::string>& set) const;
   std::string type_of(const std::string& name, const std::string& field);
   std::string type_of(const BTFId& type_id, const std::string& field);
@@ -110,7 +110,6 @@ public:
                  __u32 kind = BTF_KIND_FUNC) const;
 
 private:
-  void load_kernel_btfs(const std::set<std::string>& modules);
   SizedType get_stype(const BTFId& btf_id, bool resolve_structs = true);
   void resolve_fields(const BTFId& type_id,
                       std::shared_ptr<Struct> record,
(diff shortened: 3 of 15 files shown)


src/probe_matcher.cpp
```diff
@@ -138,20 +138,12 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     // Then we check the BTF to filter out functions from that list that don't
     // have any BTF definitions.
     case ProbeType::rawtracepoint: {
-      if (bpftrace_->has_btf_data()) {
```
jordalgo, line 141: moved this logic into a helper method so it can be mocked in tests

src/bpffeature.h
```diff
@@ -165,9 +165,8 @@ class BPFfeature {
       int* outfd = nullptr);
   bool try_load_btf(const void* btf_data, size_t btf_size);
 
-  BTF btf_ = BTF({ "vmlinux" });
-
   BPFnofeature no_feature_;
+  BTF* btf_;
```
amscanne, line 169: Having this be a bare pointer creates ambiguity about ownership, since the `BPFfeature` type can be instantiated independently. Maybe just make the BTF type a `shared_ptr` across the board?

Otherwise, you might want to a way to bind the lifetime here. (If none of those are nice suggestions, than I'd say at minimum you'd want to document the semantics on the constructor and here.)

src/btf.h
```diff
@@ -70,18 +70,16 @@ class BTF {
   };
 
 public:
-  BTF(const std::set<std::string>& modules);
-  BTF(BPFtrace* bpftrace, const std::set<std::string>& modules) : BTF(modules)
-  {
-    bpftrace_ = bpftrace;
-  };
+  BTF(BPFtrace* bpftrace);
```
amscanne, line 73: Nit: Since this is a cleanup, can this be decomposed into what `BTF` actually needs? Right now you have:

```
Bpftrace <----> BTF
```

Which suggests that there's some composition problem, and ideally you have:

```
        Bpftrace ------> BTF
             \          /
              \        /
               v      v
             SomethingData
```

As a general rule, I think that the hierarchy shouldn't have cycles unless you need to.

src/main.cpp
```diff
@@ -88,6 +88,9 @@ enum Options {
   DEBUG,
   DRY_RUN,
 };
+
+constexpr auto FULL_SEARCH = "*:*";
```
amscanne, line 92: Nit: you can declare this in the scope where it is used. (You can keep it at the top if used broadly within the file, but I think the reasonable principle here is "smallest scope possible", e.g. if it was specific to a class you would keep it in the class, etc.

amscanne (approved):

src/btf.h
jordalgo: I know but that is a larger change. It references bpftrace's bpffeature object, which is fine; I can pass that as a dependency but it also updates bpftrace's structs and calls some of its helper functions.

src/bpffeature.h
jordalgo: Yeah, a shared ptr is a good idea - let me look into that.

src/main.cpp
jordalgo: Note: it's used in different functions in main, so I think file level is appropriate here.

src/btf.cpp
```diff
@@ -1072,7 +1073,7 @@ ast::Pass CreateParseBTFPass()
 {
   return ast::Pass::create(
       "btf", []([[maybe_unused]] ast::ASTContext &ast, BPFtrace &b) {
-        b.parse_btf(b.list_modules(ast));
+        b.parse_module_btf(b.list_modules(ast));
```
viktormalik, line 1076: Correct me if I'm wrong but with this change, this pass no longer parses the entire BTF, just the modules BTF. The vmlinux BTF is parsed in `BTF` constructor which must be called before this pass is executed (otherwise `btf_` in `BPFtrace::parse_module_btf` is null).

Is that the expected behaviour? Are we sure that it's safe to do?

tests/dwarf_common.h
```diff
@@ -29,5 +31,28 @@ class test_dwarf : public ::testing::Test {
     std::remove(bin_);
   }
 
+  void SetUp() override
+  {
+    // BTF data file
+    char *btf_path = strdup("/tmp/btf_dataXXXXXX");
```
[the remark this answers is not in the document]
jordalgo: yeah good call. let me look into this more

src/bpffeature.h
```diff
@@ -165,9 +165,8 @@ class BPFfeature {
       int* outfd = nullptr);
   bool try_load_btf(const void* btf_data, size_t btf_size);
 
-  BTF btf_ = BTF({ "vmlinux" });
-
   BPFnofeature no_feature_;
+  std::shared_ptr<BTF> btf_;
```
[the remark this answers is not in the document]
jordalgo: I think a shared ptr is less risky and also would cause an issue I believe if a BPFtrace object was ever moved/copied.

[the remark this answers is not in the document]
viktormalik: Hm, I don't see many risks in using `unique_ptr<BTF>` in `BPFtrace` (as it is the actual owner of the `BTF` instance) and passing a const reference to `BPFfeature`. I believe that we use that pattern throughout the codebase a lot and it should handle the case of moving the `BPFtrace` object.

But I'm not strongly opinionated either way, it's just that we don't use `shared_ptr` a lot (as we don't have "shared ownership" in many places and IMO neither we have it here).

src/btf.cpp
jordalgo: > I don't think so, it was created by

Ah you're right. I should have actually looked instead of trying to recall from memory 🧠 

> So I guess that the only real change here is that vmlinux BTF is always parsed as opposed to the previous implementation

Actually it was always parsed in the previous implementation because a BTF object was constructed (and vmlinux parsed) by BPFfeature when constructing the BPFtrace object. Very tricky! But I agree with your point that we can possibly stop this in tests that don't need it. Let me look into that in a follow-up commit/PR.

src/bpffeature.h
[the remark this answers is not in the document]
jordalgo: I don't feel strongly either so I'll just update that to use a unique_ptr and references

src/btf.cpp
viktormalik: Absolutely, thanks!

src/bpffeature.h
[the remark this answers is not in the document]
jordalgo: @viktormalik Do you mind if I do this in a follow-up PR - I think it ties closely to not parsing the vmlinux BTF in our tests because the MockBPFFeature will then need a reference to a BTF object

tests/dwarf_common.h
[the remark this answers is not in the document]
jordalgo: Removed. It now extends `test_btf` and it's necessary because it looks at the fake task_struct in data_source.c

src/bpffeature.h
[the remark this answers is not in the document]
jordalgo: @viktormalik Gentle ping :)

tests/dwarf_common.h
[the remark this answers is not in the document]
viktormalik: Ah, but I don't think that it's correct that way.

I assume that the reason for this change is the `field_analyser_dwarf:dwarf_types_bitfield` test which uses `task_struct`. IIUC, the goal of that test is to check whether DWARF parser is able to parse bitfields so we need to use the definition of `task_struct` from DWARF (as `task_struct` from data_source.c has bitfields).

I'm not 100% sure but with this change, it looks to me as if we were using `task_struct` from BTF so we are not testing DWARF bitfield parsing at all.

src/bpffeature.h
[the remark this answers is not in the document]
viktormalik: Sure, we can change it in another PR.

tests/dwarf_common.h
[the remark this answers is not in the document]
jordalgo: @viktormalik Thanks for pushing on this. I clearly didn't understand what was going on in this test. I think it made it more confusing because this test is exercising a strange codepath whereby we're using DWARF to resolve types for builtins (e.g. `curtask`) so BTF had to be completely unavailable: https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/field_analyser.cpp#L104

I'll probably work on a follow-up PR to actually utilize a uprobe function argument to test bitfields instead of using `curtask`.

src/bpffeature.h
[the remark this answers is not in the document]
jordalgo: FYI: I just rolled the change into this PR.

tests/dwarf_common.h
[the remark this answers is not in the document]
jordalgo: @viktormalik I think this is ready for another look (no rush, just putting it back on your queue)

src/bpffeature.h
```diff
@@ -95,13 +96,12 @@ class BPFfeature {
...
   bool has_uprobe_multi();
-  bool has_fentry();
   bool has_skb_output();
   bool has_prog_fentry();
   bool has_module_btf();
   bool has_iter(std::string name);
-
-  bool has_kernel_func(Kfunc kfunc);
+  virtual bool has_fentry();
+  virtual bool has_kernel_func(Kfunc kfunc);
```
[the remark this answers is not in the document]
amscanne: Why not just make all these mockable functions virtual? You can ignore this because I’m being pedantic, but now I see this comment and I’m asking myself “what’s special about these specific functions?”. I’m reasonably sure the answer is that we just don’t have tests that need to mock the other things yet, but that now requires a small leap :)

src/btf.cpp
```diff
@@ -43,7 +43,7 @@ __s32 BTF::start_id(const struct btf *btf) const
   return btf == vmlinux_btf ? 1 : vmlinux_btf_size + 1;
 }
 
-BTF::BTF(const std::set<std::string> &modules)
+BTF::BTF(BPFtrace *bpftrace) : bpftrace_(bpftrace)
```
amscanne, line 46: It’s too much for this change (this is an aside), but I’m confused about this. Why does the BTF have a pointer back to the god object? It’s  owned by the same object, so this is a design flag for me.

Looking at it, it seems there are three things:
- Using the structmanager. This is just a composition problem, and the BTF class should just have a shared reference to this.
- Asking “is BTF enabled”. This is a strange pattern — why does the BPFtrace object instantiate the BTF object if it is not enabled? It can also just pass this as a boolean, or you can have a hierarchy of types here ( e.g. NoBTF as an implementation of the class with a well-defined API)
- checking if a tracepoint exists. I think we gotta split out some of that stuff from “BPFtrace” and have a “ProbeManager” or something similar. (Or maybe this is already the ProbeMatcher, and can we just pass a reference to this object?)

src/btf.h
```diff
@@ -156,13 +153,19 @@ class BTF {
...
 };
 
 inline bool BTF::has_data() const
 {
   return state == OK;
 }
 
+inline bool BTF::modules_loaded() const
+{
+  return modules_loaded_;
```
amscanne, line 166: Nit: we have the “state”, why not encode this in there? It’s always nice when the types are not capable of representing an invalid state, e.g. currently you may have uninitialized and modules_loaded_=true as an impossible state? If you encode MODULES_LOADED in the state then you model as a state machine, can assert regarding current state, etc.

src/main.cpp
```diff
@@ -177,6 +180,8 @@ static void info(BPFnofeature no_feature)
   struct utsname utsname;
   uname(&utsname);
 
+  auto btf = bpftrace::BTF(nullptr);
```
amscanne, line 183: Nit: I’d love for the backref to get burned down, but if this is valid please just add a default constructor. Looking at this line, I have no idea what nullptr is, and it takes a lot of context chasing for the reader.

amscanne (approved):
No concerns from me

src/btf.cpp
jordalgo: Agree that BTF should not have a ref back to bpftrace. Working on cleaning that up but it's a bit of a pain.

src/btf.h
jordalgo: I'll look into fixing this but modules can't be loaded if the state is not already OK.

tests/dwarf_common.h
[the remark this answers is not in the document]
viktormalik: Sorry for the delay in the review. It looks great, thanks!

PR #3976, opened 2025-04-06
Don't crash if kernel is built without PID namespaces

tbodt:
Like the android kernel.

##### Checklist

Not sure how to test this?

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1943,14 +1943,17 @@ std::unordered_set<std::string> BPFtrace::get_raw_tracepoint_modules(
   return mod != rts.end() ? mod->second : std::unordered_set<std::string>();
 }
 
-const struct stat &BPFtrace::get_pidns_self_stat() const
+const std::optional<struct stat> &BPFtrace::get_pidns_self_stat() const
 {
-  static struct stat pidns = []() -> auto {
+  static std::optional<struct stat> pidns = []() -> std::optional<struct stat> {
     struct stat s;
-    if (::stat("/proc/self/ns/pid", &s))
+    if (::stat("/proc/self/ns/pid", &s)) {
+      if (errno == ENOENT)
+        return std::nullopt;
       throw std::runtime_error(
           std::string("Failed to stat /proc/self/ns/pid: ") +
           std::strerror(errno));
+    }
     return s;
   }();
 
--- a/tests/mocks.h
+++ b/tests/mocks.h
@@ -74,17 +74,17 @@ public:
     return { "mock_vmlinux" };
   }
 
-  const struct stat &get_pidns_self_stat() const override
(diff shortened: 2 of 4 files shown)


src/bpftrace.cpp
```diff
@@ -1943,14 +1943,17 @@ std::unordered_set<std::string> BPFtrace::get_raw_tracepoint_modules(
   return mod != rts.end() ? mod->second : std::unordered_set<std::string>();
 }
 
-const struct stat &BPFtrace::get_pidns_self_stat() const
+const std::optional<struct stat> &BPFtrace::get_pidns_self_stat() const
```
jordalgo, line 1946: I think it's more important that the value inside the optional is const rather than the optional itself, which still means a caller could still modify the stat struct

tbodt: damn i've been writing too much rust huh

tests/mocks.h
```diff
@@ -74,17 +74,17 @@ class MockBPFtrace : public BPFtrace {
...
 
-  const struct stat &get_pidns_self_stat() const override
+  const std::optional<struct stat> &get_pidns_self_stat() const override
   {
-    static const struct stat init_pid_namespace = []() {
+    static const std::optional<struct stat> init_pid_namespace = []() {
       struct stat s {};
       s.st_ino = 0xeffffffc; // PROC_PID_INIT_INO
-      return s;
+      return std::optional{ s };
```
danobi, line 82: Should be ok to just return `s` like this, right? Same for below

```suggestion
      return s;
```

src/bpftrace.cpp
danobi: According to this: https://stackoverflow.com/questions/55060223/is-the-content-of-a-const-stdoptional-always-const
(which I did not test) I think what we have here is what we want. `std::optional<const struct stat>` would allow wholesale replacement of the value inside the optional.

danobi (commented):
overall lgtm. just a small nit

src/bpftrace.cpp
jordalgo: @danobi why do we want to protect the optional instead of the thing inside it which was (prior to this change) a const ref?

danobi: Const on the outside protects both the optional and the contained value. Const on the inside does not protect the optional - I think it is strictly worse.

jordalgo: Ugh, you're right. I read that overflow answer too fast. Sorry for the noise.

tests/mocks.h
tbodt: That didn't compile (i guess the closure return type is inferred from the type of the object in the return statement not the type of the variable the result is assigned ti

tbodt:
I gotta say I appreciate rust significantly more after this code review 😁

danobi (approved):

PR #3978, opened 2025-04-07
fix: stats_t is not a multi-output map

amscanne:
Stacked PRs:
 * #3988
 * #3984
 * #3983
 * #3982
 * #3981
 * #3980
 * #3979
 * __->__#3978

--- --- ---

### fix: stats_t is not a multi-output map

Stats do not require multiple keys. By request, they cannot be casted or
aggregated and must be kept in a map for printing.

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -459,15 +459,20 @@ public:
   }
   bool IsMapIterableTy() const
   {
-    return !IsMultiOutputMapTy();
+    if (IsMultiKeyMapTy()) {
+      return false;
+    }
+    if (NeedsPercpuMap() && !IsCastableMapTy()) {
+      return false;
+    }
+    return true;
   }
 
-  // These are special map value types that can't be reduced to a single value
-  // and output multiple lines when printed
-  bool IsMultiOutputMapTy() const
+  // These are special map value types that use multiple keys to store a single
+  // logical value (from the user perspective).
+  bool IsMultiKeyMapTy() const
   {
-    return type_ == Type::hist_t || type_ == Type::lhist_t ||
-           type_ == Type::stats_t;
+    return type_ == Type::hist_t || type_ == Type::lhist_t;
   }
 
   bool NeedsPercpuMap() const;
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -330,8 +330,8 @@ bool SemanticAnalyser::is_valid_assignment(const Expression *target,
                                            const Expression *expr)
 {
   if (expr->is_map) {
-    // Prevent assigning multi-output aggregations to another map.
-    if (expr->type.IsMultiOutputMapTy())
+    // Prevent assigning aggregations to another map.
+    if (expr->type.IsMultiKeyMapTy())
       return false;
     // Prevent declaring a map copying another aggregate map.
     if (const auto *target_map = dynamic_cast<const Map *>(target)) {
@@ -339,6 +339,9 @@ bool SemanticAnalyser::is_valid_assignment(const Expression *target,
       if (expr->type.IsCastableMapTy() && !map_has_type)
         return false;
     }
+    // Preventing folding of non-castable maps.
+    if (!expr->type.IsCastableMapTy() && expr->type.NeedsPercpuMap())
+      return false;
   }
   return true;
 }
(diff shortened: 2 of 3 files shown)


jordalgo (changes requested):
It's true that stats_t does not require two keys but I think being able to cast it, assign it to a scratch variable, or use it in a for loop makes for strange/unexpected behavior. If the casted value is a struct with named fields then I think it makes more sense.

amscanne:
> It's true that stats_t does not require two keys but I think being able to cast it, assign it to a scratch variable, or use it in a for loop makes for strange/unexpected behavior. If the casted value is a struct with named fields then I think it makes more sense.

Some context:

I'm trying to reduce the custom behavior around types here, in order to support de-sugaring maps (#3988) and simplifying analysis and code generation.

The histogram types are a separate issue. I had assumed that they were parametric types (e.g. they defined an array type based on the number of buckets), but the whole "multi-output" thing is really subtle and broken as a user: you can easily end up with different partial histograms, rather than a small number of complete histograms. (And then the partial histograms are presented as complete in the output!) As a first step, I'm just trying to isolate these actual multi-map types. I think that the stats type was marked as multi-output just to avoid having it be castable, which seems to be the complement of the multi-output types.

Unless I'm mistaken, we don't really have any examples of other types which allow you to materialize as a named structure, partially because the plumbing for that just isn't there. (I don't think it should be built, because I'd rather reduce the number of special internal-only types and shift these to general external types, and have this "just work"). I'm happy to adjust the casting behavior for `stats_t` to be whatever you like, but I don't want to invent something new.

jordalgo:
> I'm happy to adjust the casting behavior for stats_t to be whatever you like, but I don't want to invent something new.

Fair enough. Personally, I'd prefer that we leave the current restrictions around this map type as they are but I'm fine with some internal changes to reduce custom behavior. That being said, looking at this PR I'm not entirely sure what those would be as the method `IsMultiOutputMapTy` is exactly for the purpose of preventing the user from doing these things with stats_t.

amscanne:
I've made the change. This is now a new category of type: not one that lives in multiple keys, but one that we do not allow casting and aggregation.

src/types.h
```diff
@@ -459,15 +459,15 @@ class SizedType {
...
   {
-    return !IsMultiOutputMapTy();
+    return !IsMultiOutputMapTy() && (IsCastableMapTy() || !NeedsPercpuMap());
   }
 
   // These are special map value types that can't be reduced to a single value
-  // and output multiple lines when printed
+  // and output multiple lines when printed. In essence, once conceptual value
+  // written is actually multiple map keys.
   bool IsMultiOutputMapTy() const
```
jordalgo, line 468: I think you've now changed the meaning of this function. `stats_t` IS a multi output map as it prints multiple things from whatever structure it keeps under the hood. I know you're trying to separate maps that keep multiple keys per value from ones that don't. In that case, I would say to rename this function to something like `IsMultiKeyMapTy` - just to remove the "output" part.

amscanne: You're right that there are ambiguous interpretations, but I don't think the previous definition holds either. Any aggregate type would then be a "multi output map", e.g.

```
struct foo { int x; int y; } BEGIN { @ = *(struct foo*)(0); print(@); }
```

But in this case `@` is not `IsMultiOutputMapTy`, despite being irreducible when printing.

I can change it per your suggestion, but not sure what the actual intent is here.

src/types.h
```diff
@@ -459,15 +459,14 @@ class SizedType {
   }
   bool IsMapIterableTy() const
   {
-    return !IsMultiOutputMapTy();
+    return !IsMultiKeyMapTy() && (IsCastableMapTy() || !NeedsPercpuMap());
```
jordalgo, line 462: Honestly (and maybe because it's Monday) but my brain hurts trying to read this.

How about something like this, which is more verbose but requires less thinking:
```
if (IsMultiKeyMapTy()) {
  return false;
}
if (NeedsPercpuMap() && !IsCastableMapTy()) {
  return false;
}
return true;
```

src/types.h
amscanne: > I disagree with that example. That prints 1 struct not multiple.

I don't understand. I'm saying that it's identical to `stats_t`. It is one struct, not multiple. You were saying that this did apply to `stats_t` because it is an aggregate type ("can't be reduced to a single value"), but this is true for all composite types and they were not considered "multi-output".

> The intent was just to be able to localize these odd map types that couldn't be used for things like casting, for loops, var assignment.

IMHO the problem is too many basic kinds and no way to build on top of them. Not trying to fix this here. But it's clear there's a problem when you need a category system on the top of them :)

src/types.h
amscanne: Done

jordalgo (approved):

src/types.h
jordalgo: uh oh - did I give bad logical advice here ?

amscanne:
Sorry was stupidly missing the `return true` branch. Apparently needs re-approval.

jordalgo (approved):

PR #3981, opened 2025-04-07
ast: support automatic scalar maps

amscanne:
Stacked PRs:
 * #3998
 * #3997
 * #3996
 * #3988
 * #3984
 * __->__#3981

--- --- ---

### ast: support automatic scalar maps

Instead of relying on the `NoneType` for maps, we can support
automatically printing maps as scalars when they are exactly one entry
in size. In the future, this will be move the generation of default keys
up from code generation and effectively "desugar" map functions.

This commit preserves the existing `NoneType` behavior, but allows for
automatic detection of scalar maps, and propagates this information via
the `RequiredResources`. In the future, all maps will have a type after
being desugared, and this detection mechanism will be the only way that
scalar maps work.


tests/runtime/map
```diff
@@ -1,23 +1,23 @@
 NAME map declaration lruhash
-PROG config = { unstable_map_decl=1 } let @a = lruhash(1); BEGIN { @a[1] = 1; @a[2] = 2; exit(); }
```
jordalgo, line 2: Help me understand why there is a need to change this test? Will an lruhash map of 1 entry be turned into a scalar? Does this work for the per-cpu variety? Is the scalar local to the CPU?

amscanne: Yes — if you explicitly access an LRU hash map using only literal index 0, it will be turned into a scalar. This has two implications:
- It will only allocate 1 element for this map.
- It will print it with the index at the end. (This is the only perceptible change from the user perspective, and it applies only in this extremely odd usage case.)

Excellent point re: per CPU maps. Let me take a pass and make sure that’s sane. I will probably just exclude this behavior on those maps, but I need to figure out the current behavior there.

I think it’s in the commit message, but the point of this is that the desugaring can then reduce scalar maps to the single constant expression, and we don’t need special cases later in semantic analysis and codegen.

amscanne: I've added a few tests. I've also added a CHANGELOG note in the breaking section (although I think it is a very small change in behavior, always good to note it).

tests/runtime/map
```diff
@@ -1,13 +1,41 @@
...
+
+NAME map scalar
+PROG BEGIN { @a = 0; exit(); }
+EXPECT @a: 0
+
+NAME map auto scalar
+PROG BEGIN { @a[0] = 0; exit(); }
+EXPECT @a: 0
+
+# Unfortunately there are some issues when evicting lruhash entries. We can't
```
jordalgo, line 14: This whole comment should be updated now that we know this is expected behavior and will never be "fixed"

tests/runtime/map
```diff
@@ -1,13 +1,41 @@
...
-PROG config = { unstable_map_decl=1 } let @a = hash(1); BEGIN { @a[1] = 1; @a[2] = 2; exit(); }
+PROG config = { unstable_map_decl=1 } let @a = hash(2); BEGIN { @a[0] = 0; @a[1] = 1; @a[2] = 2; exit(); }
 EXPECT_REGEX .*WARNING: Map full; can't update element.*
+EXPECT @a[0]: 0
 EXPECT @a[1]: 1
 
+NAME map declaration hash (auto scalar)
+PROG config = { unstable_map_decl=1 } let @a = hash(1); BEGIN { @a[1] = 1; @a[2] = 2; exit(); }
+EXPECT_REGEX .*WARNING: Map full; can't update element.*
+EXPECT @a: 1
```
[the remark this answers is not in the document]
amscanne: You're right, the output side was inconsistent. This is now serialized with the `map_info` so the output side treats 1-sized maps the way it does before.

tests/runtime/map
amscanne: Done.

tests/runtime/map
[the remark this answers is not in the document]
amscanne: The original tests have been restored.

tests/runtime/map
```diff
@@ -1,13 +1,55 @@
...
+# Unfortunately there are some issues when evicting lruhash entries. We can't
+# know exactly which elements will be present. See #3992 for more information.
 NAME map declaration lruhash
-PROG config = { unstable_map_decl=1 } let @a = lruhash(1); BEGIN { @a[1] = 1; @a[2] = 2; exit(); }
+PROG config = { unstable_map_decl=1 } let @a = lruhash(2); BEGIN { @a[0] = 0; @a[1] = 1; @a[2] = 2; exit(); }
 EXPECT_REGEX_NONE .*WARNING: Map full; can't update element.*
-EXPECT @a[2]: 2
+EXPECT_REGEX @a\[.\]: .
+
+NAME map declaration lruhash
```
jordalgo, line 33: This seems like almost a copy of the test above. Remove?

amscanne: It is not -- it is a single element and the above is two elements. This is ensuring that the map isn't automatically treated as a scalar map when the keys may be non-zero (per the very valid point that you had)

jordalgo: > per the very valid point that you had

I forget easily.

amscanne:
@jordalgo I remember now why I didn't do the **only syntactic scalar maps work this way** thing. That's a bit of a different pass, which comes later. Now, this is limited **only** to cases where you use only literal 0 and nothing else.

jordalgo (approved):
Just an FYI to other reviewers: the plan is to have an additional PR after this which adds an output from the map desugar pass and drops the whole bounds check (thereby removing the auto scalar conversion).

amscanne:
Fingers crossed that the series here is smooth sailing. We can always roll-back if it gets too complex. Going to merge this one now, and get the others rebased.

PR #3982, opened 2025-04-07
config: refactor and simplify configuration plumbing

amscanne:
Stacked PRs:
 * #3998
 * #3997
 * #3996
 * #3988
 * #3984
 * #3981
 * __->__#3982

--- --- ---

### config: refactor and simplify configuration plumbing

This change started as a way to remove the `StackMode` processing from
the lexer and parser, in order to simplify the literal mechanisms.
Unfortunately, it became quickly apparent that the AST types that these
gave rise to were somewhat tied in to the configuration itself. In order
to achieve this simpliciation, this change refactors and rewrites the
`Config` object.

Instead of a enum class of keys, the `Config` is changed to be a
plain-old C object. This allows us to more throughly document each
member, and reference the configuration is an easily understood and
type-checked way. This also makes the defaults incredibly obvious.

The parsing of different types is delegated to extensible `ConfigParser`
classes, that can be declared for any type that you wish to add to the
configuration. The mapping from configuration "keys" (human strings) is
owned by `config.cpp`, which automatically defines parsers for each of
the associated fields. Amazingly, this is the only place that some
amount of template magic occurs, in order to automatically construct the
appropriate parser based on the field type.

This change does drop some functionality:

* The "ConfigSource" parts are dropped, since they were never referenced
  beyond checking if a configuration could be overriden. There is a much
  simpler way to do this: load the environment variables and override
  configuration values *after* they have been parsed from the script.

* The potential to allow arbitrary expressions as configuration values
  is removed. Arguably, this could have been useful in the future for
  doing evaluation, but this was not done yet. This will be reintroduced
  with proper literal plumbing, coming shortly.

On of the upsides of this change is that parsing is unified for each
type, and we can accept a wider variety of options. For example, in the
configuration block we treat identifiers as strings, so quoting is
necessary only for values containing a space or other characters.
Similarly, values like `true` and `false` can be accepted for boolean
values, in addition to `0` and `1`.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1,59 +1,30 @@
 #include <algorithm>
 #include <cstring>
 #include <fstream>
+#include <unordered_set>
 
 #include "config.h"
 #include "log.h"
 #include "types.h"
+#include "util/int_parser.h"
 
 namespace bpftrace {
 
-Config::Config(bool has_cmd)
+char ParseError::ID;
+char RenameError::ID;
+
+void ParseError::log(llvm::raw_ostream &OS) const
 {
-  config_map_ = {
-    { ConfigKeyBool::cpp_demangle, { .value = true } },
-    { ConfigKeyBool::lazy_symbolication, { .value = false } },
-    { ConfigKeyBool::print_maps_on_exit, { .value = true } },
-    { ConfigKeyBool::unstable_import, { .value = false } },
-    { ConfigKeyBool::unstable_map_decl, { .value = false } },
-#ifndef HAVE_BLAZESYM
-    { ConfigKeyBool::use_blazesym, { .value = false } },
-#else
-    { ConfigKeyBool::use_blazesym, { .value = true } },
-#endif
-    { ConfigKeyInt::log_size, { .value = static_cast<uint64_t>(1000000) } },
-    { ConfigKeyInt::max_bpf_progs, { .value = static_cast<uint64_t>(1024) } },
-    { ConfigKeyInt::max_cat_bytes, { .value = static_cast<uint64_t>(10240) } },
-    { ConfigKeyInt::max_map_keys, { .value = static_cast<uint64_t>(4096) } },
-    { ConfigKeyInt::max_probes, { .value = static_cast<uint64_t>(1024) } },
-    { ConfigKeyInt::max_strlen, { .value = static_cast<uint64_t>(1024) } },
-    { ConfigKeyInt::max_type_res_iterations,
-      { .value = static_cast<uint64_t>(0) } },
-    { ConfigKeyInt::on_stack_limit, { .value = static_cast<uint64_t>(32) } },
-    { ConfigKeyInt::perf_rb_pages, { .value = static_cast<uint64_t>(64) } },
-    { ConfigKeyStackMode::default_, { .value = StackMode::bpftrace } },
-    { ConfigKeyString::license, { .value = std::string("GPL") } },
-    { ConfigKeyString::str_trunc_trailer, { .value = std::string("..") } },
-    { ConfigKeyMissingProbes::default_,
-      { .value = ConfigMissingProbes::warn } },
-    // by default, cache user symbols per program if ASLR is disabled on system
-    // or `-c` option is given
-    { ConfigKeyUserSymbolCacheType::default_,
-      { .value = (has_cmd || !is_aslr_enabled())
-                     ? UserSymbolCacheType::per_program
-                     : UserSymbolCacheType::per_pid } },
-  };
+  OS << key_ << ": " << detail_;
 }
 
-bool Config::can_set(ConfigSource prevSource, ConfigSource source)
+void RenameError::log(llvm::raw_ostream &OS) const
 {
-  return prevSource == ConfigSource::default_ ||
-         (prevSource == ConfigSource::script &&
-          source == ConfigSource::env_var);
+  OS << "key has been renamed to '" << name_ << "'";
 }
 
 // /proc/sys/kernel/randomize_va_space >= 1
-bool Config::is_aslr_enabled()
+static bool is_aslr_enabled()
 {
   std::string randomize_va_space_file = "/proc/sys/kernel/randomize_va_space";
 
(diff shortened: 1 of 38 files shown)


src/types.h
```diff
@@ -84,6 +86,33 @@ const std::map<StackMode, std::string> STACK_MODE_NAME_MAP = {
   { StackMode::raw, "raw" },
 };
 
+template <>
+struct ConfigParser<StackMode> {
```
jordalgo, line 90: I'm a bit confused as to why this ConfigParser lives here instead of config.cpp

tests/codegen/license.cpp
```diff
@@ -7,8 +7,7 @@ namespace codegen {
 TEST(codegen, license)
 {
   auto bpftrace = get_mock_bpftrace();
-  auto configs = ConfigSetter(*bpftrace->config_, ConfigSource::script);
-  configs.set(ConfigKeyString::license, "Dual BSD/GPL");
+  bpftrace->config_->license = "Dual BSD/GPL";
```
jordalgo, line 10: I have to admit this kind of setting is one thing I liked about the old Config class implementation - being able to accidentally do this on non-test code paths is a bit sad 😢

tests/config_analyser.cpp
```diff
@@ -116,50 +117,56 @@ TEST(config_analyser, config)
...
   test("config = { BAD_CONFIG=1 } BEGIN { }",
-       R"(stdin:1:12-22: ERROR: Unrecognized config variable: BAD_CONFIG
+       R"(stdin:1:12-24: ERROR: BAD_CONFIG: not a known configuration option
 config = { BAD_CONFIG=1 } BEGIN { }
-           ~~~~~~~~~~
+           ~~~~~~~~~~~~
 )",
        false);
   test(
       "config = { BPFTRACE_MAX_PROBES = \"hello\" } BEGIN { }",
```
jordalgo, line 126: Is this now legal?
`BPFTRACE_MAX_PROBES = hello `

src/types.h
amscanne: Ideally they just live next to whatever types they are supposed to parse, if you want to support configuration variables of that type. The `config.cpp` ones should just be the basic ones (integer, bool), and unfortunately the `ConfigMissingProbes` type is defined only there. I'd rather move that elsewhere to be close to where it is semantically meaningful.

tests/codegen/license.cpp
amscanne: There are language features for this, we should be passing around a `const Config&` to places that don't need mutability :)

tests/config_analyser.cpp
amscanne: I'm not sure what you mean by legal. It will now parse, but the config analyzer is going to add an error, as indicated below because it fails to parse as the intended type. So the failure is pushed back a bit in the pipeline, into a pass. From the user perspective, it is an error just as it was before.

It was always legal to have `bpftrace = "hello"` or other non-sensical assignments, which would behave exactly the same way. If we want `stack_mode = kernel` to parse but not `stack_mode = hello` then we have no choice but to have `kernel` as a first-class lex token. This change just accepts arbitrary identifiers, and allows the actual configuration parser to determine if it makes sense for the expected type of the configuration variable.

Essentially now all configuration type-checking is done in one spot, during the configuration pass, rather than being done half in the parser and half in the pass.

src/config.cpp
```diff
@@ -73,110 +44,302 @@ bool Config::is_aslr_enabled()
...
-{
-  static auto stack_mode_map = get_stack_mode_map();
-  auto found = stack_mode_map.find(s);
-  if (found != stack_mode_map.end()) {
-    return std::make_optional(found->second);
+template <>
+struct ConfigParser<bool> {
+  Result<OK> parse(const std::string &key, bool *target, const std::string &s)
+  {
+    if (s == "1" || s == "true" || s == "TRUE" || s == "on" || s == "ON" ||
```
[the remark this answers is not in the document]
amscanne: This only operates per character. You'd basically need this:

```
std::string v(s);
std::ranges::transform(s, s.begin(), [](unsigned char c) {
  return std::tolower(c);
});
```

I was aiming just to keep it simple, and support the existing set of keys.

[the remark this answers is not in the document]
jordalgo: A little utility function is pretty simple I think. I mean there is no template meta programming involved so maybe it's not cool enough for ya 😉

tests/config_analyser.cpp
jordalgo: Ah sorry, my question was confusing. What I meant to ask was for config variables that accept a string e.g. "license" can people just pass an un-quoted string / raw identifier?

tests/codegen/license.cpp
jordalgo: That's fair. I guess there really is only one or two places that needs to mutate the Config object.

src/types.h
jordalgo: > Ideally they just live next to whatever types they are supposed to parse

Where would this be except for `types.h` ? I understand what you're saying but I think the logical place for all ConfigParsers at the moment is `config.cpp`. I won't block on this but having this little one-off seems strange when it could easily live with all the other organizationally in `config.cpp`

src/config.cpp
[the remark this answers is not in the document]
jordalgo: I won't block because it's not changing current behavior but I still think it would be nice, simple addition.

jordalgo (dismissed):

src/types.h
amscanne: I would argue that "how do you parse a `StackMode`" should live with the `StackMode`, and not a collection of "here's how you parse every different type of thing".

We only have a few basic config types at the moment, but one could imagine this extending a bit. For example, imagine you add a basic `License` type that still wraps a string. But perhaps it knows how to parse common licenses and convert them into the suitable representation that the kernel understands. That logic should live on its own, with other license-related logic, no? Perhaps it lives with a pass that also knows the license of kfuncs. (I will say that in practice this is a bit of a straw man because I'm not so much in favor of the `config` as a thing, but I do think separation of concerns is a good general design pattern.)

I'm just trying to have a bit less of an everything-about-everything style :) Obviously `types.h` is it's own thing (and very much has that going on), so I'm not advocating moving things there specifically, but to that specific detail, maybe in the future we just need a `types/stack_mode.h` with just those definitions :)

src/config.cpp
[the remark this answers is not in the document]
amscanne: I've added this.

src/types.h
amscanne: Turns out that we already need it. I was moving this as suggestion and realized that this is now what is used to parse the random literal stack mode literal for `ustack(raw)` elsewhere. (Since this is no longer being lexed, this site just says `ConfigParser<StackMode>().parse(...)`.)

jordalgo (approved):
```
config = {
    accepted=TrUe;
}
```

PR #3983, opened 2025-04-07
ast: allow `replace` to operate on `Expression*` and `Statement*` types

amscanne:
Stacked PRs:
 * #3988
 * #3984
 * __->__#3983

--- --- ---

### ast: allow `replace` to operate on `Expression*` and `Statement*` types

The visitor is starting to get a bit complex, but we can keep using the
existing `replace` pattern for now. This extends it to ensure that it
works on cross-type nodes (e.g. `Expression*` and `Statement*`), which
will allow us to replace some of those node for desugaring.


jordalgo (approved):
Okey dokey

PR #3984, opened 2025-04-07
refactor: drop `is_literal` and add folding pass

amscanne:
Stacked PRs:
 * #3998
 * #3997
 * #3996
 * #3988
 * __->__#3984

--- --- ---

### refactor: drop `is_literal` and add folding pass

In order to reduce the complexity inside the more complex passes (the
semantic analyser and code generation), add a folding pass for literals
and constants. Note that this is not intended to do the work that LLVM
will do in the backend, rather it is merely folding constants that must
be interpreted by semantic analysis.

A classic example is `str($1)`, which can be folded into a suitable
`String` node. This eliminates the need for special cases in semantic
analysis and the `is_in_str` field within the AST.

Note that many tests are updated mostly to allow these new literal
cases, but there are a few noteworthy cases:

* `buf(arg0, arg1)` remains legal, because the code for this check is
  broken. The buffer will always be sized at the maximum size.

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -97,27 +97,25 @@ public:
   Map *key_for_map = nullptr;
   Map *map = nullptr;      // Only set when this expression is assigned to a map
   Variable *var = nullptr; // Set when this expression is assigned to a variable
-  bool is_literal = false;
 };
 using ExpressionList = std::vector<Expression *>;
 
 class Integer : public Expression {
 public:
-  explicit Integer(ASTContext &ctx,
-                   int64_t n,
-                   Location &&loc,
-                   bool is_negative = true);
+  explicit Integer(ASTContext &ctx, int64_t n, Location &&loc);
+  const uint64_t value;
+};
 
-  int64_t n;
-  bool is_negative;
+class NegativeInteger : public Expression {
+public:
+  explicit NegativeInteger(ASTContext &ctx, int64_t n, Location &&loc);
+  const int64_t value;
 };
 
 class PositionalParameter : public Expression {
 public:
   explicit PositionalParameter(ASTContext &ctx, long n, Location &&loc);
-
-  long n;
-  bool is_in_str = false;
+  const long n;
 };
 
 class PositionalParameterCount : public Expression {
@@ -128,8 +126,7 @@ public:
 class String : public Expression {
 public:
   explicit String(ASTContext &ctx, std::string str, Location &&loc);
-
-  std::string str;
+  const std::string value;
 };
 
 class Identifier : public Expression {
@@ -280,7 +277,7 @@ public:
   TupleAccess(ASTContext &ctx, Expression *expr, ssize_t index, Location &&loc);
 
   Expression *expr = nullptr;
-  ssize_t index;
+  size_t index;
 };
 
 class Cast : public Expression {
@@ -399,7 +396,6 @@ class Unroll : public Statement {
 public:
   Unroll(ASTContext &ctx, Expression *expr, Block *block, Location &&loc);
 
-  long int var = 0;
   Expression *expr = nullptr;
   Block *block = nullptr;
 };
--- a/src/ast/passes/fold_literals.cpp
+++ b/src/ast/passes/fold_literals.cpp
@@ -1,0 +1,452 @@
+#include "ast/passes/fold_literals.h"
+#include "ast/ast.h"
+#include "ast/visitor.h"
+#include "bpftrace.h"
+#include "log.h"
+#include "util/int_parser.h"
+
+namespace bpftrace::ast {
+
+namespace {
+
+class LiteralFolder : public Visitor<LiteralFolder, Expression *> {
+public:
+  LiteralFolder(ASTContext &ast, BPFtrace &bpftrace)
+      : ast_(ast), bpftrace_(bpftrace) {};
+
+  using Visitor<LiteralFolder, Expression *>::visit;
+
+  template <typename T>
+  T *replace(T *node, [[maybe_unused]] Expression **result)
+  {
+    if constexpr (std::is_same_v<T, Expression>) {
+      if (*result != nullptr && *result != node) {
+        return *result;
+      }
+    }
+    return node;
+  }
+
+  Expression *visit(Cast &cast);
+  Expression *visit(Unop &op);
+  Expression *visit(Binop &op);
+  Expression *visit(PositionalParameterCount &param);
+  Expression *visit(PositionalParameter &param);
+  Expression *visit(Call &call);
+
+private:
+  ASTContext &ast_;
+  BPFtrace &bpftrace_;
+};
+
+} // namespace
+
+template <typename T>
+static std::optional<std::variant<uint64_t, int64_t>> eval_binop(T left,
+                                                                 T right,
+                                                                 Operator op)
+{
+  auto clamp = [](auto v) -> std::variant<uint64_t, int64_t> {
+    if constexpr (std::is_same_v<T, int64_t>) {
+      if (v >= 0) {
+        return static_cast<uint64_t>(v);
+      }
+    }
+    return v;
+  };
+
+  switch (op) {
+    case Operator::EQ:
+      return static_cast<uint64_t>(left == right);
+    case Operator::NE:
+      return static_cast<uint64_t>(left != right);
+    case Operator::LE:
+      return static_cast<uint64_t>(left <= right);
+    case Operator::GE:
+      return static_cast<uint64_t>(left >= right);
+    case Operator::LT:
+      return static_cast<uint64_t>(left < right);
+    case Operator::GT:
+      return static_cast<uint64_t>(left > right);
+    case Operator::PLUS:
+      if (left == 0 || right == 0) {
+        return clamp(left + right);
+      }
+      if (left > 0 && right > 0) {
+        auto res = static_cast<uint64_t>(left) + static_cast<uint64_t>(right);
+        if (res > static_cast<uint64_t>(left) &&
+            res > static_cast<uint64_t>(right)) {
+          return res;
+        }
+      }
+      if constexpr (std::is_same_v<T, int64_t>) {
+        if ((left > 0 && right < 0) || (left < 0 && right > 0)) {
+          return clamp(left + right);
+        }
+        if (left < 0 && right < 0) {
+          auto res = left + right;
+          if (res < left && res < right) {
+            return res;
+          }
+          return std::nullopt;
+        }
+      }
+      return std::nullopt;
+    case Operator::MINUS:
+      if constexpr (std::is_same_v<T, uint64_t>) {
+        if (left >= right) {
+          return left - right;
+        } else {
+          // Ensure that this is representable during conversion. We know that
+          // right > left, therefore the delta here is guaranteed to be greater
+          // than zero.
+          uint64_t delta = right - left - 1;
+          auto max = static_cast<uint64_t>(std::numeric_limits<int64_t>::max());
+          if (delta > max) {
+            return std::nullopt;
+          } else {
+            return -static_cast<int64_t>(delta) - 1;
+          }
+        }
+      } else {
+        if (right == 0) {
+          return clamp(left);
+        } else if (right < 0) {
+          auto res = left - right;
+          if (res < left) {
+            return std::nullopt;
+          }
+          return clamp(res);
+        } else {
+          auto res = left - right;
+          if (res > left) {
+            return std::nullopt;
+          }
+          return clamp(res);
+        }
+      }
+    case Operator::MUL:
+      if (left == 0 || right == 0) {
+        return static_cast<uint64_t>(0);
+      } else {
+        if constexpr (std::is_same_v<T, int64_t>) {
+          // Spill into unsigned.
+          if (left < 0 && right < 0) {
+            return eval_binop(static_cast<uint64_t>(-(left + 1)) + 1,
+                              static_cast<uint64_t>(-(right + 1)) + 1,
+                              op);
+          }
+        }
+        auto result = left * right;
+        if (result / left != right) {
+          return std::nullopt; // Overflow.
+        }
+        return clamp(result);
+      }
+    case Operator::DIV:
+      if (right == 0) {
+        return std::nullopt;
+      }
+      return clamp(left / right);
+    case Operator::MOD:
+      if (right == 0) {
+        return std::nullopt;
+      }
+      return clamp(left % right);
+    case Operator::BAND:
+      return clamp(left & right);
+    case Operator::BOR:
+      return clamp(left | right);
+    case Operator::BXOR:
+      return clamp(left ^ right);
+    case Operator::LEFT:
+      return clamp(left << right);
+    case Operator::RIGHT:
+      return clamp(left >> right);
+    case Operator::LAND:
+      return static_cast<uint64_t>(left && right);
+    case Operator::LOR:
+      return static_cast<uint64_t>(left || right);
+    case Operator::INVALID:
+    case Operator::ASSIGN:
+    case Operator::INCREMENT:
+    case Operator::DECREMENT:
+    case Operator::LNOT:
+    case Operator::BNOT:
+      break;
+  }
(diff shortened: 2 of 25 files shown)


src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+private:
+  ASTContext &ast_;
+  BPFtrace &bpftrace_;
+};
+
+} // namespace
+
+template <typename T>
+static std::optional<std::variant<uint64_t, int64_t>> eval_binop(T left,
```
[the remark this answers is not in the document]
amscanne: @jordalgo 👀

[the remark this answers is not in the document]
jordalgo: @amscanne Looks like the kinda comments you would add 😝 - j/k!

tests/semantic_analyser.cpp
```diff
@@ -770,7 +773,7 @@ kprobe:f { lhist(-10, -10, 10, 1); }
            ~~~~~~~~~~~~~~~~~~~~~~
 )");
   test_error("kprobe:f { @ = lhist(-10, -10, 10, 1); }", R"(
-stdin:1:16-38: ERROR: lhist() min must be non-negative (provided min -10)
```
jordalgo, line 773: nit: I do think the little "(provided min -10)" is helpful

src/ast/passes/fold_literals.cpp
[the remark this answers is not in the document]
amscanne: No, it definitely does.

tests/semantic_analyser.cpp
amscanne: Because the negative literals here are a separate kind, this is now folded into the case where anything at all is provided other than a non-negative integer. I could specifically re-add a case that checks to see if you've provided a negative number, but it doesn't feel worth having a special case just for that in `lhist`.

Once we converge on something generic for checking against prototypes, we could certainly include the value of the actual provided expression.

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,450 @@
...
+      if (integer->value >= static_cast<uint64_t>(str->value.size())) {
+        op.addWarning() << "literal string will always be empty";
+        return ast_.make_node<String>("", Location(op.loc));
+      }
+      return ast_.make_node<String>(str->value.substr(integer->value),
+                                    Location(op.loc));
+    }
+
+    // Check for another string.
+    auto *rs = dynamic_cast<String *>(other);
```
jordalgo, line 263: Might be more cruft but since you checked above already you could have a variable indicating if you already know the other is not a string.

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+  // Check for string cases.
+  auto *str = dynamic_cast<String *>(op.left);
+  auto *other = op.right;
+  if (str == nullptr) {
+    str = dynamic_cast<String *>(op.right);
+    other = op.left;
+  }
+  if (str) {
+    // For whatever reason you are allowed to fold "foo"+3.
```
jordalgo, line 251: Wait what? You mean you can already do this in bpftrace?

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,450 @@
...
+    if (ru->value <= std::numeric_limits<int64_t>::max()) {
+      result = eval_binop(ls->value, static_cast<int64_t>(ru->value), op.op);
+    }
+  } else {
+    // This is not an integer expression at all.
+    return nullptr;
+  }
+
+  if (!result) {
+    // This is not a valid expression.
```
jordalgo, line 335: Had trouble figuring that out from the added error and return value 🤪

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,450 @@
...
+    // If this is an integer of some kind, then fold it into a string.
+    if (auto *n = dynamic_cast<Integer *>(call.vargs.at(0))) {
+      std::stringstream ss;
+      ss << n->value;
+      s = ss.str();
+    } else if (auto *n = dynamic_cast<NegativeInteger *>(call.vargs.at(0))) {
+      std::stringstream ss;
+      ss << n->value;
+      s = ss.str();
+      return ast_.make_node<String>(ss.str(), Location(call.loc));
```
jordalgo, line 413: why don't we need to truncate the string if it's negative?

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,450 @@
...
+#include "ast/visitor.h"
+#include "bpftrace.h"
+#include "log.h"
+#include "util/int_parser.h"
+
+namespace bpftrace::ast {
+
+namespace {
+
+class LiteralFolder : public Visitor<LiteralFolder, Expression *> {
```
jordalgo, line 12: I feel like there should be a ton of unit tests for this class. Any reason not to have them?

src/ast/ast.h
```diff
@@ -97,27 +97,25 @@ class Expression : public Node {
...
 };
 using ExpressionList = std::vector<Expression *>;
 
 class Integer : public Expression {
 public:
-  explicit Integer(ASTContext &ctx,
-                   int64_t n,
-                   Location &&loc,
-                   bool is_negative = true);
+  explicit Integer(ASTContext &ctx, int64_t n, Location &&loc);
```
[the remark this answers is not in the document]
amscanne: It is technically `NonNegativeInteger`, I just felt that was overly verbose. Should I use `UnsignedInteger`?

src/ast/passes/printer.cpp
```diff
@@ -26,7 +26,14 @@ std::string Printer::type(const SizedType &ty)
 void Printer::visit(Integer &integer)
 {
   std::string indent(depth_, ' ');
-  out_ << indent << "int: " << integer.n << type(integer.type) << std::endl;
+  out_ << indent << "int: " << integer.value << type(integer.type) << std::endl;
```
jordalgo, line 29:
```suggestion
  out_ << indent << "unsigned int: " << integer.value << type(integer.type) << std::endl;
``` ?

Comment (amscanne) [reply src/ast/passes/printer.cpp:29]: Uh, I had this and had to update tons of change detector tests. I ended up reverting because I felt that the `int` case was by far more common and I didn't want to get into the weeds. WDYT? Should I make this universally `unsigned int`?

File: tests/semantic_analyser.cpp [hunk 10]
```diff
@@ -1249,7 +1252,7 @@ TEST(semantic_analyser, call_str)
   test("kprobe:f { str(arg0); }");
   test("kprobe:f { @x = str(arg0); }");
   test("kprobe:f { str(); }", 1);
-  test("kprobe:f { str(\"hello\"); }", 1);
```
jordalgo, line 1252: I know it's not causing any issues to allow this but should we add a warning to tell the user this is unneccessary?

src/ast/passes/fold_literals.cpp
amscanne: No, I will add them to this change shortly. I just want to get the architecture bits for the stack out of the way.

tests/semantic_analyser.cpp
```diff
@@ -2892,24 +2895,24 @@ TEST(semantic_analyser, signed_int_division_warnings)
 {
   bool invert = true;
   std::string msg = "signed operands";
-  test_for_warning("kprobe:f { @ = -1 / 1 }", msg);
-  test_for_warning("kprobe:f { @ = 1 / -1 }", msg);
+  test_for_warning("kprobe:f { @x = -1; @y = @x / 1 }", msg);
```
jordalgo, line 2898: why do we need all these extra map assignments for these tests?

amscanne: Because literals are untyped and will be folded without any warnings as long as there is no overflow/underflow. They only because typed when they are actually assigned. This test is checking that the actual semantic analyzer **type checks** are triggered correctly.

src/ast/ast.h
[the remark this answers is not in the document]
jordalgo: Oh right, because of 0 not being positive. Hmm, yeah `UnsignedInteger` I think is better.

src/ast/passes/printer.cpp
jordalgo: I mean, this is just for AST debugging. I think it's better if we have a 'signed int' to have an 'unsigned int' but that could be a follow-up PR (or a "good first issue" issue).

tests/semantic_analyser.cpp
jordalgo: **this** would be a good place for a comment :)

tests/semantic_analyser.cpp
amscanne: It gets complicated. What if the user has `str("foo", 3)`? I'd rather just let it be a no-op than have complex logic around whether it is a warning or not (e.g. if the resulting fold is bounded or not).

src/ast/passes/fold_literals.cpp
amscanne: Good catch, this is definitely a mistake. Fixed in my local, will fix on update.

tests/semantic_analyser.cpp
jordalgo: Yeah that's fair. Not a big deal.

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,450 @@
...
+    } else if (op.op == Operator::MINUS) {
+      // Ensure that it is representable as a negative value.
+      if (integer->value >
+          static_cast<uint64_t>(std::numeric_limits<int64_t>::max()) + 1) {
+        op.addError() << "negative value will underflow";
+        return nullptr;
+      }
+      // Carefully make the conversion. We need to ensure that this does not
+      // overflow/underflow while doing the math.
+      if (integer->value <= 2) {
```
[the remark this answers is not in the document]
amscanne: I could actually use `1` instead of `2` (but it's the same).

The problem is the `static_cast<int64_t>(integer->value)` below. The value `-std::numeric_limits<int64_t>::min()` is not representable in an `int64_t`, but obviously `std::numeric_limits<int64_t>::min()` is a legal negative number. Therefore we cannot just do `-static_cast<int64_t>(integer->value)`. We need to shrink that value by at least one to do the cast, and then subtract to get to the final form.

src/ast/passes/fold_literals.cpp
amscanne: Yep.

src/ast/passes/printer.cpp
amscanne: I'd rather leave this bit alone for this change. Certainly a valid clean-up later on, but it would also introduce a bunch of unrelated churn renamed `Integer` to `UnsignedInteger`. (And there's plenty of churn already.)

src/ast/passes/fold_literals.cpp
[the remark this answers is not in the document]
jordalgo: I think using `1` is better just so you don't have people like me look at it and wonder why it's `2` lol

src/ast/passes/fold_literals.cpp
jordalgo: Umm, can you give me an example?

src/ast/passes/printer.cpp
jordalgo: Mind just making an issue for it?

src/ast/passes/fold_literals.cpp
amscanne: https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L1327

jordalgo: I think it's something to do with char * offset

tests/semantic_analyser.cpp
amscanne: Done

src/ast/passes/fold_literals.cpp
amscanne: I've added a bunch of tests around this. They look like the other tests for now, for simplicity.

src/ast/passes/fold_literals.cpp
```diff
@@ -0,0 +1,459 @@
...
+    case Operator::LOR:
+      return static_cast<uint64_t>(left || right);
+    case Operator::INVALID:
+    case Operator::ASSIGN:
+    case Operator::INCREMENT:
+    case Operator::DECREMENT:
+    case Operator::LNOT:
+    case Operator::BNOT:
+      break;
+  }
```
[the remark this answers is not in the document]
jordalgo: wow what a useful comment from our robot friends...

src/ast/passes/fold_literals.cpp
jordalgo: Ah, my bad - it was for positional params only 🙄

src/ast/passes/fold_literals.cpp
jordalgo: The tests look good!

jordalgo:
I think this is almost ready to go - there is a clang-tidy issue and one or two unresolved comments. Overall, great work!

src/ast/ast.h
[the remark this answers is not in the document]
amscanne: I will merge with the below and file an issue when this is landed. (I don't want to file an issue ahead of this, just because it's a bit weird for a pending PR.)

src/ast/passes/printer.cpp
amscanne: I will when it lands, if that's okay.

tests/semantic_analyser.cpp
```diff
@@ -2741,20 +2744,19 @@ TEST(semantic_analyser, positional_parameters)
...
   // Pointer arithmetic in str() for parameters
-  // Only str($1 + CONST) where CONST <= strlen($1) should be allowed
   test(bpftrace, "kprobe:f { printf(\"%s\", str($1 + 1)); }");
   test(bpftrace, "kprobe:f { printf(\"%s\", str(1 + $1)); }");
-  test(bpftrace, "kprobe:f { printf(\"%s\", str($1 + 4)); }", 2);
-  test(bpftrace, "kprobe:f { printf(\"%s\", str($1 * 2)); }", 2);
-  test(bpftrace, "kprobe:f { printf(\"%s\", str($1 + 1 + 1)); }", 1);
+  test(bpftrace, "kprobe:f { printf(\"%s\", str($1 + 4)); }");
+  test(bpftrace, "kprobe:f { printf(\"%s\", str($1 * 2)); }");
+  test(bpftrace, "kprobe:f { printf(\"%s\", str($1 + 1 + 1)); }");
```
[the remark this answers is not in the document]
amscanne: I've added this test to the `fold_literals.cpp` test suite.

src/ast/passes/fold_literals.cpp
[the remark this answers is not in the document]
amscanne: They're not wrong, but also just not that helpful.

src/ast/passes/fold_literals.cpp
amscanne: I don't love extra state, so I'd rather leave as is. Easier for me to understand, at least personally.

src/ast/passes/printer.cpp
jordalgo: Sure.

src/ast/passes/fold_literals.cpp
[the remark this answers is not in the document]
jordalgo: > They're not wrong

In that ANY code has potential problems?

tests/semantic_analyser.cpp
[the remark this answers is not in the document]
amscanne: Sorry, it was coming just with the latest export. Takes a minute to work changes up and rebase.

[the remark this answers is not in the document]
amscanne: (It's line 169 in the test file)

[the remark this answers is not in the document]
jordalgo: I see it now. Thanks!

PR #3988, opened 2025-04-07
ast: Implement map "de-sugaring" as a separate pass

amscanne:
Stacked PRs:
 * #3998
 * #3997
 * #3996
 * __->__#3988

--- --- ---

### ast: Implement map "de-sugaring" as a separate pass

In order to simplify and standardize code generation, implement an early
pass that desugars map syntax, and starts to handle code paths are
currently pushed back to code generation.

There are three major things done here, which are somewhat entangled.

1) Always fill in default keys.

Currently, code generation will transparently fill in the keys for
scalar maps, e.g. `@x` becomes an array map of size 1.

Since this will be automatically recognized as a scalar map, we simply
desugar this statement by doing `@x[0]` for that first case. All error
messages will be unified aroud paths where maps diverge for any reason,
e.g. `@x = 1; @x["foo"] = 1`.

2) Split map accesses into explicit reads `MapAccess` and writes
   `AssignMapStatement`. Both of these require explicit keys.

(Note that in the future we could make assignments an expression also,
but this should come after fixing up some of the code generation
semantics to make this a bit cleaner.)

This removes a lot of the special cases in the semantic analyzer (e.g.
`skip_map_key_validation`, `is_read`) because these are simply different
node in the tree explicitly.

3) Turn `@x[0] = avg()` into `avg(@x, 0)`.

The only way to avoid the map `@x` being desugared into `@x[0]` is to
use it in the context of a special assignment. These will be split into
the map argument and the key expression, and passed to the function.

The semantic analyser is able to infer the map type from the hard-coded
function prototypes for these special functions. In the future, we can
extend this inference to non-hard-coded functions.

This has the biggest effect on internals, and allows us to drop the
inverted pointers (e.g. `Expression->map`), since code generation will
have all the context required to emit the necessary code.


tests/runtime/call
```diff
@@ -657,7 +657,10 @@ TIMEOUT 3
 
 NAME map len keyless
 PROG BEGIN { @ = 0; printf("map len: %d\n", len(@)); exit(); }
-EXPECT map len: 1
+EXPECT stdin:1:40-45: ERROR: @ used as a non-scalar map, previously used as a scalar map
```
jordalgo, line 660:
```suggestion
EXPECT stdin:1:40-45: ERROR: @ used as a non-scalar map. 'len' only accepts maps that have keys
```

tests/runtime/map
```diff
@@ -7,21 +7,17 @@ NAME map scalar
 PROG BEGIN { @a = 0; exit(); }
 EXPECT @a: 0
 
-NAME map auto scalar
+NAME map no auto scalar
```
jordalgo, line 10: should we just get rid of these 'auto scalar' runtime tests now that nothing is being auto-scalared

tests/semantic_analyser.cpp
```diff
@@ -384,9 +387,9 @@ BEGIN { @x[1] = 0; @x; }
   test("BEGIN { @x[1, ((int8)2, ((int16)3, 4))] = 0; @x[5, (6, (7, 8))]; }");
 
   test_error("BEGIN { @x[1,2] = 0; @x[3]; }", R"(
-stdin:1:22-27: ERROR: Argument mismatch for @x: trying to access with arguments: 'int64' when map expects arguments: '(int64,int64)'
+stdin:1:22-26: ERROR: Argument mismatch for @x: trying to access with arguments: 'int64' when map expects arguments: '(int64,int64)'
 BEGIN { @x[1,2] = 0; @x[3]; }
-                     ~~~~~
+                     ~~~~
```
jordalgo, line 392: darn - we lost the last squiggly

tests/semantic_analyser.cpp
```diff
@@ -778,34 +772,22 @@ kprobe:f { @ = lhist(-10, -10, 10, 1); }
                ~~~~~~~~~~~~~~~~~~~~~~
 )");
   test_error("kprobe:f { $x = lhist(); }", R"(
-stdin:1:17-24: ERROR: lhist() should be directly assigned to a map
-kprobe:f { $x = lhist(); }
-                ~~~~~~~
-stdin:1:17-24: ERROR: lhist() requires 4 arguments (0 provided)
```
jordalgo, line 784: A bit sad that we're losing this error message when two things are wrong but I understand that this might be just the price we have to pay for this change.

tests/semantic_analyser.cpp
```diff
@@ -908,25 +890,34 @@ TEST(semantic_analyser, call_delete)
...
 
   test_error("kprobe:f { delete(1); }", R"(
-stdin:1:12-20: ERROR: delete() expects a map for the first argument and a key for the second argument e.g. `delete(@my_map, 1);`
+stdin:1:12-21: ERROR: delete() requires 2 arguments (1 provided)
 kprobe:f { delete(1); }
+           ~~~~~~~~~
+)");
+
+  test_error("kprobe:f { delete(1, 1); }", R"(
+stdin:1:12-20: ERROR: delete() expects a map argument
```
jordalgo, line 899: can we include the argument index in the error message? e.g. `expects a map as the first argument` or `as argument number 1`

src/ast/ast.h
```diff
@@ -322,14 +325,32 @@ class VarDeclStatement : public Statement {
   bool set_type = false;
 };
 
+// Scalar map assignment is purely a syntax sugar that is removed by the pass
```
jordalgo, line 328:
```suggestion
// Scalar map assignment is purely syntax sugar that is removed by the pass
```

src/ast/passes/map_sugar.cpp
```diff
@@ -0,0 +1,274 @@
...
+  using Visitor<MapDefaultKey>::visit;
+  void visit(Call &call);
+  void visit(For &for_loop);
+  void visit(Map &map);
+  void visit(MapAccess &acc);
+  void visit(AssignScalarMapStatement &assign);
+  void visit(AssignMapStatement &assign);
+  void check(Map &map, bool indexed);
+
+  MapMetadata metadata()
```
jordalgo, line 51: why move the metadata out of here? maybe the function should be named `extract_metadata`?

src/ast/passes/map_sugar.cpp
```diff
@@ -0,0 +1,274 @@
...
+public:
+  using Visitor<MapAssignmentCheck>::visit;
+  void visit(Call &call);
+};
+
+} // namespace
+
+// These are special functions which are part of the map API, and operate
+// independently of any key. We allow the first argument to be a pure map,
+// and process therefore don't expand a default key in these cases.
```
jordalgo, line 102:
```suggestion
// independently of any key. We allow the first argument to be a pure map
// and therefore don't expand a default key in these cases.
```

src/ast/passes/map_sugar.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+void MapDefaultKey::visit(For &for_loop)
+{
+  check(*for_loop.map, true);
+  Visitor<MapDefaultKey>::visit(for_loop.stmts);
+}
+
+void MapFunctionAliases::visit(Call &call)
+{
+  // If this is specfically `delete(@)` then we remap this to `clear`. We
```
jordalgo, line 190: The problem with that is you've turned a sync function into an async function. Can't we just desugar it to `delete(@, key)` or for scalars `delete(@, 0)` ?

jordalgo (commented):
Looking pretty good so far - I still have to review the semantic_analyser and codegen_llvm changes...

src/ast/passes/map_sugar.cpp
amscanne: Yikes, I didn't even realize that `clear` was async. Yes, I can desugar as `delete(..., 0)`.

amscanne: So I originally got this from the docs, which I guess are wrong:

```
Delete a single key from a map. For scalar maps (e.g. no explicit keys), the key is omitted and is equivalent to calling clear.
```

It is *not* equivalent to `clear`.

src/ast/passes/map_sugar.cpp
```diff
@@ -0,0 +1,274 @@
...
+void MapDefaultKey::check(Map &map, bool indexed)
+{
+  bool scalar = !indexed;
+  auto val = metadata_.scalar.find(map.ident);
+  if (val == metadata_.scalar.end()) {
+    metadata_.scalar.emplace(map.ident, scalar);
+  } else if (val->second != scalar) {
+    if (!scalar) {
+      map.addError() << map.ident
+                     << " used as a non-scalar map, previously used as a "
```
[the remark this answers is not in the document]
amscanne: I just tried to pick the term used in the docs. Happy to adjust and be more explicit, e.g.

```
map.ident used as a map without an explicit key (scalar map), previously used with an explicit key
```

WDYT?

src/ast/passes/map_sugar.cpp
amscanne: I can just make it public and move it out from the pass.

tests/semantic_analyser.cpp
amscanne: I can make this annotation on the actual first argument.

tests/semantic_analyser.cpp
amscanne: Yeah, we just stop processing if there are errors after a given pass, since we don't necessarily want to overload the user. They'll still get this error, but only once `lhist` is used correctly.

src/ast/passes/map_sugar.cpp
jordalgo: > It is not equivalent to clear.

It certainly is not.

src/ast/ast.h
amscanne: Done

src/ast/passes/printer.cpp
```diff
@@ -147,12 +147,20 @@ void Printer::visit(MapDeclStatement &decl)
 
 void Printer::visit(Map &map)
 {
+  // Use a slightly customized format for the map type here, since it is never
+  // going to be marked as `is_ctx`, not have an associated address space.
```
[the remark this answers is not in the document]
amscanne: Done

src/ast/passes/map_sugar.cpp
amscanne: Done

src/ast/passes/map_sugar.cpp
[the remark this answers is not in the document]
amscanne: Updated all tests and references

tests/runtime/call
amscanne: I've updated this these messages for all calls to include the specific function and context.

tests/semantic_analyser.cpp
amscanne: Done

src/ast/passes/map_sugar.cpp
amscanne: This transformation is no longer being done.

src/ast/passes/map_sugar.cpp
amscanne: Done.

tests/runtime/map
amscanne: I've left them in place but made them a bit saner. I do think they are good rails with respect to the intended behavior either way.

amscanne: (They no longer say "auto scalar", they just say "scalar-like")

src/ast/passes/codegen_llvm.cpp
```diff
@@ -3337,87 +3322,70 @@ int CodegenLLVM::getNextIndexForProbe()
...
-                             scoped_key_expr.value(),
-                             key_expr->type.GetSize());
-        } else {
-          LOG(BUG) << "Type size mismatch. Key Type Size: "
-                   << key_type.GetSize()
-                   << " Expression Type Size: " << key_expr->type.GetSize();
-        }
+  auto scoped_key_expr = visit(key_expr);
+  const auto &key_type = map.key_type;
+  // Allocation needs to be done after recursing via vist(key_expr) so that
```
jordalgo, line 3331:
```suggestion
  // Allocation needs to be done after recursing via visit(key_expr) so that
```

tests/semantic_analyser.cpp
```diff
@@ -944,7 +935,7 @@ kprobe:f { @x = 1; delete(@x) ? 0 : 1; }
 )");
 
   test_error("kprobe:f { @y[5] = 5; delete(@y[5], 5); }", R"(
-stdin:1:23-35: ERROR: delete() expects a map with no keys for the first argument
+stdin:1:23-35: ERROR: delete() expects a map argument
```
jordalgo, line 938: This error messaging still isn't great. I won't block but we should address this in a follow up PR. It should say "delete() expects a map argument. Got a map access" or something like that.

tests/semantic_analyser.cpp
```diff
@@ -956,13 +947,13 @@ kprobe:f { @y[(3, 4, 5)] = 5; delete(@y, (1, 2)); }
 )");
 
   test_error("kprobe:f { @y[1] = 2; delete(@y); }", R"(
-stdin:1:23-32: ERROR: delete() expects a map for the first argument and a key for the second argument e.g. `delete(@my_map, 1);`
+stdin:1:23-32: ERROR: delete() expects a map without explicit keys (scalar map)
```
jordalgo, line 950: this is just wrong. it should be something like "delete(), when called with 1 argument, expects a map without explicit keys (scalar map)."

tests/semantic_analyser.cpp
```diff
@@ -956,13 +947,13 @@ kprobe:f { @y[(3, 4, 5)] = 5; delete(@y, (1, 2)); }
...
 kprobe:f { @y[1] = 2; delete(@y); }
                       ~~~~~~~~~
 )");
 
   test_error("kprobe:f { @a[1] = 1; delete(@a, @a); }", R"(
-stdin:1:34-36: ERROR: Argument mismatch for @a: trying to access with no arguments when map expects arguments: 'int64'
+stdin:1:34-36: ERROR: @a used as a map without an explicit key (scalar map), previously used without an explicit key (non-scalar map)
```
jordalgo, line 956:
```suggestion
stdin:1:34-36: ERROR: @a used as a map without an explicit key (scalar map), previously used with an explicit key (non-scalar map)
```

tests/semantic_analyser.cpp
```diff
@@ -974,9 +965,9 @@ kprobe:f { @a[1] = 1; delete(@a, @a); }
   test(R"(kprobe:f { @y[1, "longerstr"] = 5; delete(@y[1, "hi"]); })");
 
   test_error("kprobe:f { @x = 1; @y = 5; delete(@x, @y); }", R"(
-stdin:1:39-41: ERROR: Argument mismatch for @x: trying to access with arguments: 'int64' when map expects no arguments
+stdin:1:28-37: ERROR: delete() expects a map with explicit keys (non-scalar map)
```
jordalgo, line 968: Again, this should read something like "delete(), when called with 2 arguments, expects a map with explicit keys (non-scalar map) as it's first argument"

tests/semantic_analyser.cpp
```diff
@@ -985,16 +976,22 @@ kprobe:f { @x[1, "hi"] = 1; delete(@x["hi", 1]); }
                             ~~~~~~~~~~~~~~~~~~
 )");
 
+  test_error("kprobe:f { @x[0] = 1; @y[5] = 5; delete(@x, @y[5], @y[6]); }", R"(
+stdin:1:34-58: ERROR: delete() requires 2 arguments (3 provided)
```
jordalgo, line 980: Also incorrect. "delete() requires 1 or 2 arguments (3 provided)"

tests/semantic_analyser.cpp
```diff
@@ -4795,21 +4812,30 @@ BEGIN { unroll(1) { $a = 1; } print(($a)); }
...
+BEGIN { @a = lhist(123, 0, 123, 1); let $b = @a; }
+                                    ~~~~~~
 )");
 
   test_error("BEGIN { @a = stats(10); let $b = @a; }", R"(
-stdin:1:25-36: ERROR: Map value 'stats_t' cannot be assigned to a scratch variable.
+stdin:1:25-36: ERROR: Value 'stats_t' cannot be assigned to a scratch variable.
 BEGIN { @a = stats(10); let $b = @a; }
                         ~~~~~~~~~~~
+stdin:1:25-31: WARNING: Variable $b never assigned to.
```
jordalgo, line 4836: This is kinda funny. Side effect of the map sugar pass?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -76,13 +77,49 @@ class PassTracker {
...
+
+struct map_type_spec {
+  // This indicates that the argument must be a map type. The given function
+  // may be called to determine the map type.
+  std::function<SizedType(const Call &call)> type;
+};
+
+struct map_key_spec {
+  // This indicates that the argument must be a key. The index provided here
+  // indicates whether argument this is a key expression for, so the types can
```
jordalgo, line 97: This could use a re-wording :)

tests/semantic_analyser.cpp
amscanne: This error is generated by the semantic analyzer by your call spec. The `delete` call for a scalar is expanded into `delete(@x, 0)`. How should we fix this?

tests/semantic_analyser.cpp
amscanne: I think it's just because the type is still none at the end. I can chase this down, but as long as the error messages are sensible I'm not generally fighting the errors.

tests/semantic_analyser.cpp
amscanne: Can I get away with something like "call to delete() expects ..." implying that it is just this call? Otherwise everything gets special cased around `delete` vs. all these other map calls (e.g. `has_key`, `len`, etc.).

tests/semantic_analyser.cpp
amscanne: The error message is for the current form. Per below, can I do `call to delete() expects a map without explicit keys ...`

tests/semantic_analyser.cpp
jordalgo: The current language is "delete() takes up to 2 arguments". I see that you changed `.min_args` from 1 to 2 for "delete". Can we just leave it as 1 or does that cause other issues?

tests/semantic_analyser.cpp
jordalgo: So what would be the final error message?

tests/semantic_analyser.cpp
jordalgo: It's not big enough to block, it's just weird.

tests/semantic_analyser.cpp
amscanne: I've backported this (pass wise) in the same format used by the semantic analyser (e.g. `requires 1 or 2 arguments`) as a special case for `delete`.

tests/semantic_analyser.cpp
amscanne: Fixed, whoops. Good catch.

tests/semantic_analyser.cpp
amscanne: There will never be a case where it is given 1 argument, it will always be desugared and provided a key (even if it is the default `0` key). This message is now encoded in the desugar pass.

tests/semantic_analyser.cpp
amscanne:
```
stdin:1:28-37: ERROR: call to delete() expects a map with explicit keys (non-scalar map)
kprobe:f { @x = 1; @y = 5; delete(@x, @y); }
                           ~~~~~~~~~
```

src/ast/passes/semantic_analyser.cpp
amscanne: Reworded.

jordalgo (approved):
🚀

PR #3991, opened 2025-04-09
ast: allow chain of errors

amscanne:
This updates the diagnostic structure to support chain of locations (not an arbitrary graph, just a chain to eventually allow showing expansion sites). It is necessary to support a chain rather than a span because with macro expansion, the source locations will be able to (eventually) live across different source files. Even within a single file, it is not useful for a span to cover the entire file from the macro to the probe.

The plumbing to updated to remove the operator overloading, and move the implementation for `addError` out of the header, since it is no longer a template. We can also update the type passed through to be the `ASTContext` (since the implementation is in the header, this can be used), in order to lay the groundwork for the clone operation.

##### Checklist

### DIFF
--- a/src/ast/diagnostic.cpp
+++ b/src/ast/diagnostic.cpp
@@ -1,8 +1,21 @@
-#include "diagnostic.h"
+#include <algorithm>
+#include <ranges>
 
+#include "diagnostic.h"
 #include "log.h"
 
 namespace bpftrace::ast {
+
+std::stringstream& Diagnostic::addContext(Location loc)
+{
+  // This must not modify the existing location, instead we add
+  // a new link to the location chain. We don't inject the full
+  // chain if one has been provided, but take only that node.
+  auto nloc = std::make_shared<LocationChain>(loc->current);
+  auto& nlink = nloc->parent.emplace(LocationChain::Parent(std::move(loc_)));
+  loc_ = std::move(nloc);
+  return nlink.msg;
+}
 
 void Diagnostics::emit(std::ostream& out) const
 {
@@ -18,16 +31,40 @@ void Diagnostics::emit(std::ostream& out, Severity s) const
 
 void Diagnostics::emit(std::ostream& out, Severity s, const Diagnostic& d) const
 {
-  const auto& loc = d.loc();
+  // Build our set of messages.
+  std::vector<std::pair<std::string, SourceLocation>> msgs;
+  auto loc = d.loc();
+  while (loc) {
+    auto& parent = loc->parent;
+    if (parent) {
+      msgs.emplace_back(parent->msg.str(), loc->current);
+      loc = parent->loc;
+    } else {
+      msgs.emplace_back(d.msg(), loc->current);
+      break;
+    }
+  }
+  std::ranges::reverse(msgs);
+
   switch (s) {
     case Severity::Warning:
-      LOG(WARNING, loc.source_location(), loc.source_context(), out) << d.msg();
+      if (msgs.empty()) {
+        LOG(WARNING, out) << d.msg();
+      }
+      for (const auto& [msg, loc] : msgs) {
+        LOG(WARNING, loc.source_location(), loc.source_context(), out) << msg;
+      }
       if (auto msg = d.hint(); !msg.empty()) {
         LOG(HINT, out) << msg;
       }
       break;
     case Severity::Error:
-      LOG(ERROR, loc.source_location(), loc.source_context(), out) << d.msg();
+      if (msgs.empty()) {
+        LOG(ERROR, out) << d.msg();
+      }
+      for (const auto& [msg, loc] : msgs) {
+        LOG(ERROR, loc.source_location(), loc.source_context(), out) << msg;
+      }
       if (auto msg = d.hint(); !msg.empty()) {
         LOG(HINT, out) << msg;
       }
--- a/src/ast/location.h
+++ b/src/ast/location.h
@@ -2,6 +2,8 @@
 
 #include <cassert>
 #include <memory>
+#include <optional>
+#include <sstream>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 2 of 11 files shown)


amscanne:
@danobi @jordalgo

src/ast/diagnostic.cpp
```diff
@@ -18,16 +20,37 @@ void Diagnostics::emit(std::ostream& out, Severity s) const
...
+      break;
+    }
+  }
+  std::ranges::reverse(msgs);
+
   switch (s) {
     case Severity::Warning:
-      LOG(WARNING, loc.source_location(), loc.source_context(), out) << d.msg();
+      if (msgs.empty()) {
+        LOG(WARNING, out) << d.msg();
```
jordalgo, line 38: JC, when does it happen that the diagnostic has an error without a location?

src/ast/location.h
```diff
@@ -84,4 +63,49 @@ class Location {
   std::shared_ptr<ASTSource> source_;
 };
 
+class LocationChain;
+using Location = std::shared_ptr<LocationChain>;
+
+// LocationChain is a list of locations.
```
jordalgo, line 69: You know I'm gonna comment on this. It is more of a LINKED list than a "list". Chain describes it well enough for me so this comment should go byebye.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2258,9 +2259,11 @@ void SemanticAnalyser::binop_ptr(Binop &binop)
...
         const auto *re = rht.GetPointeeTy();
         if (*le != *re) {
-          binop.left->addWarning(*binop.right)
-              << "comparison of distinct pointer types ('" << *le << ", '"
-              << *re << "')";
+          auto &warn = binop.addWarning();
+          warn << "comparison of distinct pointer types ('" << *le << ", '"
+               << *re << "')";
+          warn.addContext(binop.left->loc) << "left side";
+          warn.addContext(binop.right->loc) << "right side";
```
jordalgo, line 2266: This just to demonstrate how it can be used?

src/ast/diagnostic.cpp
amscanne: There are tests that construct a location with Location(). This used to result in a non-sensical line/column which was ignored by the log. It’s actually better like this, since the the empty location is much more explicit (and maybe I should add asserts that the locations logged are valid, and drop the default SingleLocation constructor).

src/ast/passes/semantic_analyser.cpp
amscanne: This is to replace the existing span, which will no longer be possible.

I just threw this message together though, and we can tweak them. We probably want to attach just one context, and annotate them with the types directly. Eg. annotate the left with “invalid addition for type XXX” and annotate the right with “other type is YYY”

WDYT?

src/ast/location.h
amscanne: A linked list is a list :)I don’t feel strongly, I can drop the comment.

src/ast/location.h
```diff
@@ -12,14 +14,15 @@ namespace bpftrace::ast {
 
 class ASTSource;
 
-// Location is a source location. This effectively wraps an `ASTSource` and
-// allows for a full source context to be produced.
-class Location {
+// SingleLocation refers to a single location the source.
+//
+// It does not contain any additional context.
+class SingleLocation {
```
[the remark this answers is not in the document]
amscanne: I did it that way because I don’t want anyone to really care about those details, it’s fine to still reason about a “Location” and that’s it. For macro nodes, etc, it’s just that a location may include a pointer to the node it was expanded from, and this will be included in diagnostic messages.

src/ast/location.h
jordalgo: drop it like it's hot!

src/ast/location.h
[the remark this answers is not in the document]
jordalgo: meh, I'm not really a fan of that kind of obscurity. I think it just makes for confusion down the line - that is I don't think "Location" should be an alias for a  shared pointer to a LocationChain

src/ast/passes/semantic_analyser.cpp
jordalgo: I think we can leave the main message where it is and just add the types to the left and right e.g.
```
stdin:1:31-32: ERROR: Type mismatch for '>': comparing 'hist_t' with 'int64'
kprobe:f { @ = hist(5); if (@ > 0) { print((1)); } }

stdin:1:28-30: ERROR: left (hist_t)
kprobe:f { @ = hist(5); if (@ > 0) { print((1)); } }

stdin:1:33-34: ERROR: right (int64)
kprobe:f { @ = hist(5); if (@ > 0) { print((1)); } }

```

src/ast/location.h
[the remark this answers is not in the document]
amscanne: I don't think you need to do it most of the time, but it would result in some much boilerplate. Callers don't need to care about this detail. `Location loc` is so much clearer than `std::shared_ptr<LocationChain>`. Conceptually, it's a location: the fact that it may be a list under the hood is an implementation detail.

src/ast/location.h
amscanne: I wound up updating the comment to make it useful.

src/ast/location.h
```diff
@@ -84,4 +63,49 @@ class Location {
...
+  unsigned int column() const
+  {
+    return current.column();
+  }
+
+  // This may be modified for a node, typically when copying. For example, when
+  // copying a node you may automatically modify the location in order to
+  // inject a parent that has "expanded from <...>". When these are displayed as
+  // diagnostics, they are unpacked in reverse.
+  std::optional<Link> parent;
```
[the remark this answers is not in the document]
amscanne: Done.

src/ast/passes/semantic_analyser.cpp
amscanne: Done.

src/ast/location.h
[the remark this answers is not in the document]
jordalgo: It's an implementation detail that will cause confusion because you're changing the meaning of Location. How about "Locations" (plural) as an alias for LocationChain?

src/ast/location.h
jordalgo: Much better. Thank you.

src/ast/location.h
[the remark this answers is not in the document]
amscanne: I think it's still a conceptual Location, it is just carrying extra context. I'm not sure how `Locations` addressed your concern about the abstraction of `shared_ptr`.

If this is about naming, can I call it `SourceLocation`? Then `Location` is effectively an abstraction of a `SourceLocation` which may carry context referencing other `SourceLocation`s?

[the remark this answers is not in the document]
jordalgo: Sure, `SourceLocation` works for me. And yeah, it was mostly a concern around naming confusion.

jordalgo (approved):

PR #3994, opened 2025-04-09
Lazy load btf

jordalgo:
This is primarily for tests that don't
need vmlinux BTF loaded.

Number of times vmlinux is loaded when
running bpftrace_test.
Before: 2162
After: 1626

I'm sure this number can be futher reduced
by finding all tests where we don't
need to have vmlinux BTF loaded.

This PR also included a minor refactoring to
remove a circular depedency.

##### Checklist


viktormalik (approved):
Nice! I especially love that all BTF is now loaded inside the `CreateParseBTFPass` :-)

PR #3995, opened 2025-04-10
stdlib: Add basic build harness for embedding files

amscanne:
This will be used by to hold the standard library files in the future. For now, it is used in practice to generate test files.

The first case for this will be to support an `always_true` function with an import of `stdlib/base.bc`. This change will not be merged until there is considerable movement on some of the other pieces.

##### Checklist

### DIFF
--- a/cmake/BuildBPF.cmake
+++ b/cmake/BuildBPF.cmake
@@ -1,0 +1,153 @@
+# Functions to build intermediate BPF modules.
+#
+# The `bpf` function will produce bitcode, an object as well as the BTF type
+# information for a given file. The `btf_header` function can be used to
+# produce a C header for a given BTF source file (often the kernel, but it
+# can be the output from any `bpf` rule, for example).
+
+find_program(GCC gcc REQUIRED)
+find_program(BPFTOOL bpftool REQUIRED)
+find_program(PAHOLE pahole REQUIRED)
+find_program(NM nm REQUIRED)
+find_program(AWK awk REQUIRED)
+find_program(LLVM_OBJCOPY
+  NAMES llvm-objcopy llvm-objcopy-${LLVM_VERSION_MAJOR} llvm${LLVM_VERSION_MAJOR}-objcopy
+  REQUIRED)
+find_program(CLANG
+  NAMES clang-${LLVM_VERSION_MAJOR}
+  REQUIRED)
+
+function(btf_header NAME)
+  cmake_parse_arguments(
+    ARG
+    ""
+    "OUTPUT;SOURCE"
+    ""
+    ${ARGN}
+  )
+  if (NOT DEFINED ARG_SOURCE)
+    set(ARG_SOURCE "/sys/kernel/btf/vmlinux")
+  endif ()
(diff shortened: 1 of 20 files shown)


cmake/BuildBPF.cmake
```diff
@@ -0,0 +1,153 @@
...
+  # `pahole` conversion. Let's do that.
+  #
+  # But hold on a second, we said that we wanted to build a BPF target? Turns
+  # out that `gcc` can't do that very well. So `clang` can't build our BPF
+  # target because of incorrect BTF, and `gcc` can't build the BPF target. So
+  # we need to build the source natively, use `pahole` to generate the BTF
+  # data, and then extract that separately. This makes some sense, as it is the
+  # kernel happy path.
+  #
+  # But at least what we built above has the DWARF data, right? Wrong! If were
```
jordalgo, line 96:
```suggestion
  # But at least what we built above has the DWARF data, right? Wrong! If we were
```

jordalgo (commented):
I endorse this PR but someone with more knowledge of CMake should also review.

cmake/BuildBPF.cmake
amscanne: Fixed

amscanne:
@danobi Did you want to take a look? I believe this should be uncontroversial cmake

cmake/BuildBPF.cmake
```diff
@@ -0,0 +1,153 @@
...
+  # you'll realize that clang seems to generate proper annotations and
+  # relocations for loadable programs (the happy path for that!), but fails to
+  # generate the same types for base types. In fact, in doesn't even generate
+  # named arguments without an explicit `-O2`.  Oh well, I suppose that the
+  # common path for the kernel is to build with `gcc` and then rely on the
+  # `pahole` conversion. Let's do that.
+  #
+  # But hold on a second, we said that we wanted to build a BPF target? Turns
+  # out that `gcc` can't do that very well. So `clang` can't build our BPF
+  # target because of incorrect BTF, and `gcc` can't build the BPF target. So
```
danobi, line 91: Did you try with these magic flags? https://github.com/libbpf/libbpf-rs/blob/f9ec2c444032454c2ca791c8058fc3499ee51818/libbpf-cargo/src/build.rs#L66

Building with clang should work...

amscanne: Yeah, the building was not the problem. It just had some weird issues with the per-CPU attributes on the BTF. I'm happy to rip this out and revisit it, but for right now I'm fine with it working as-is (since it's how it works currently with your version).

It could have been the case that it was a bug with `clang-18` (that I was testing with here), because the change pins that version (since it is necessary for the bitcode). But this is actually an argument in favor of just leaving this as `gcc`.

danobi (approved):

jordalgo:
@amscanne FYI this broke our docker distro builds in trunk:
https://github.com/bpftrace/bpftrace/actions/runs/14545328765/job/40810007404

jordalgo:
Codeql also broke: https://github.com/bpftrace/bpftrace/actions/runs/14545328769/job/40810007379

PR #3996, opened 2025-04-10
refactor: move to static types

amscanne:
Stacked PRs:
 * #3998
 * #3997
 * __->__#3996

--- --- ---

### refactor: move to static types

This change makes AST node consistent, and adopts static types rather
than relying on RTTI.

The major motivation for this is to move away from the `Expression`
inheritence model, whereby instances reach into the `Expression` base
class and access the `SizedType` directly. Instead, a layer of
abstraction is added here to allow individual node to provide access to
their type information via a returned reference. Instead of being
specifically a `SizedType` object, this reference can be an abstract
type of some kind, e.g. originating outside of our own type system.

Consider a future pass which examines `Call` nodes, and replaces them
with `ExternalCall` nodes based on the resolve imports. The types for
those new `ExternalCall` nodes should be BTF types sourced from the
impored type information. Currently we would need to figure out a way to
convert every type exhaustively. While this is done currently within the
`BTF` class, it is all driven by the `ClangParser`. Reinvoking the
parser again and again in order to introduce these types to our unified
type universe is not ideal: we should be able to operate exclusively and
directly on the type information contained in the external objects. Over
time, our internal implementations will leverage these mechanisms,
allowing us to remove the special cases from `SizedType`.

With this change, individual node are still typed, but they have their
own field, appropriately named. Direct access to these fields are
encouraged, and this is where type inference will operate. This reduces
ambiguity, as node without fundamental type information (e.g. a
PositionalParameter, a Block, etc.) need not store an invalid value.
They can implement a `type` function which simply references to the
appropriate sub-type.

Note that the `type` method returns immutable references, which means
that type inference and code generation can only mutate types on node
that it understands, which is a desirable safety property. This is on
top of the additional safety properties that no more `static_cast`s or
`dynamic_cast`s are permitted, reducing the possibility of subtle memory
errors.

Most of this change itself is boilerplate in two forms:
* Making AST node consistent in structure and moving the `type`.
* Changing all call-sites to use this new consistent structure.

There should be no semantic change.


tests/runtime/builtin
```diff
@@ -157,7 +157,7 @@ EXPECT ERROR: failed to open file '/does/not/exist/file': No such file or direct
 
 NAME sizeof
 PROG struct Foo { int x; char c; } BEGIN { $x = 1; printf("%d %d %d %d %d\n", sizeof(struct Foo), sizeof((*(struct Foo*)0).x), sizeof((*(struct Foo*)0).c), sizeof(1 == 1), sizeof($x)); exit(); }
-EXPECT 8 4 1 1 8
```
jordalgo, line 160: So the size of our boolean turned from 1 into 8? This by design?

tests/semantic_analyser.cpp
```diff
@@ -510,13 +511,13 @@ Program
   ?: :: [(string[13],string[13])]
    < :: [uint64]
     builtin: pid :: [uint32]
-    int: 10000 :: [int64]
+    int: 10000
    tuple: :: [(string[2],string[13])]
-    string: a :: [string[2], AS(kernel)]
```
jordalgo, line 515: Maybe you mentioned it somewhere but what happened to the address space?

tests/semantic_analyser.cpp
```diff
@@ -2361,47 +2365,47 @@ TEST(semantic_analyser, map_aggregations_implicit_cast)
   test("kprobe:f { @x = 1; @y = count(); @x = @y; }", R"(*
   =
    map: @x :: [int64]int64
-    int: 0 :: [int64]
```
jordalgo, line 2364: seems like a nice bit of removed redundancy

jordalgo (approved):
Mostly relying on our tests here as this is a lot of boilerplate to review but overall this seems like a nice cleanup.

tests/runtime/builtin
amscanne: It went through a bit of a journey. In general, this is still a literal which is always treated as size 8. I can't remember exactly why this was changed during the fold literals pass, but this is the original behavior, so I wasn't worried: https://github.com/bpftrace/bpftrace/blob/86f98d9095e7c21f81c1dee17ce254592b0f8bad/tests/runtime/builtin#L160

(In other words, before this series and after this series it is still 8)

PR #3999, opened 2025-04-11
Add config to enable debug info for blazesym

jordalgo:
This adds information about dir, file, line
and inlined functions if debug info is available
and blazesym is being used.

This config defaults to true if blazesym exists.

##### Checklist

### DIFF
--- a/src/bpftrace.h
+++ b/src/bpftrace.h
@@ -138,12 +138,8 @@ public:
                         StackType stack_type,
                         int indent = 0);
   std::string resolve_buf(const char *buf, size_t size);
-  std::string resolve_ksym(uint64_t addr, bool show_offset = false);
-  std::string resolve_usym(uint64_t addr,
-                           int32_t pid,
-                           int32_t probe_id,
-                           bool show_offset = false,
-                           bool show_module = false);
+  std::string resolve_ksym(uint64_t addr);
+  std::string resolve_usym(uint64_t addr, int32_t pid, int32_t probe_id);
   std::string resolve_inet(int af, const uint8_t *inet) const;
   std::string resolve_uid(uint64_t addr) const;
   std::string resolve_timestamp(uint32_t mode,
@@ -212,7 +208,6 @@ public:
   std::unique_ptr<BTF> btf_;
   std::unique_ptr<BPFfeature> feature_;
 
-  bool resolve_user_symbols_ = true;
   bool safe_mode_ = true;
   bool has_usdt_ = false;
   bool usdt_file_activation_ = false;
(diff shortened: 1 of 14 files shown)


jordalgo:
Small thing to consider - maybe we shouldn't show any debug info for the `perf` stack mode as that already has module (or just not show the module if we also have dir and file). Module can be different from dir and file but it's just possibly repetitive data.

src/config.cpp
```diff
@@ -252,6 +252,7 @@ const std::map<std::string, AnyParser> CONFIG_KEY_MAP = {
   { "missing_probes", CONFIG_FIELD_PARSER(missing_probes) },
   { "print_maps_on_exit", CONFIG_FIELD_PARSER(print_maps_on_exit) },
   { "use_blazesym", CONFIG_FIELD_PARSER(use_blazesym) },
+  { "show_debug_info", CONFIG_FIELD_PARSER(show_debug_info) },
```
danobi, line 255: missing adoc change?

jordalgo: It seems like we're not advertising blazesym in the adoc yet. Maybe we're waiting till it becomes the default?

amscanne (approved):
No hard objections, I do think it is worthwhile to consider adding some structure rather than evolving to hand-rolled `std::vector`, as this would likely make a lot of things easier to understand

This would also make it easier to add unit tests for that specific class and various stack formats that you expect, which AFAIK don't exist today?

src/bpftrace.h
```diff
@@ -278,6 +273,24 @@ class BPFtrace : public ast::State<"bpftrace"> {
...
+  std::vector<std::string> resolve_ksym_impl(uint64_t addr,
+                                             bool show_offset,
+                                             bool perf_mode,
+                                             bool show_debug_info);
+  std::vector<std::string> resolve_usym_impl(uint64_t addr,
+                                             int32_t pid,
+                                             int32_t probe_id,
+                                             bool show_offset,
+                                             bool perf_mode,
+                                             bool show_debug_info);
```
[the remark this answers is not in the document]
jordalgo: Good call. I got rid of the impl variants but I'm not yet sold on the class abstraction since stack mode is only used in one place for formatting. I'll think on it a bit more.

jordalgo:
Going to merge this unless anyone has objections

danobi:
@jordalgo we're not advertising blazesym b/c so far it's been a transparent change. but a new config knob should be documented

jordalgo:
@danobi kk, I'll add it to the adoc.

jordalgo:
@danobi https://github.com/bpftrace/bpftrace/pull/4039

PR #3987, opened 2025-04-07
Fix build failures due to missing location.hh

viktormalik:
ast/location.h does `#include "location.hh"` and location.hh is generated by the parser so any CMake target whose source includes ast/location.h needs to have a dependency on the `parser` target, otherwise the compilation may fail due to incorrect ordering of build targets.

compiler_core is an example of such a target, however, in this case, we can just drop the include of ast/location.h as it's not used.

Fixes #3986.

##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc`
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #4000, opened 2025-04-11
Fix build failures due to missing location.hh

viktormalik:
location.hh is generated by the parser so any CMake target whose source includes it needs to have a dependency on the `parser` target, otherwise the compilation may fail due to incorrect ordering of build targets. This also applies to targets which include location.hh transitively via other headers.

To avoid such errors, add an explicit dependency on the `parser` target for all CMake targets including location.hh.

##### Checklist


viktormalik:
@acj I backported #3987 to the 0.23 release here, however, the code is quite different (a number of changes are not applicable). Could you please test this PR to see if it succeeds to build on Alpine? This is the one we'll use to cut a new patch release.

Thanks!

jordalgo (approved):

Issue #4001, opened 2025-04-11
cstring_view will not guarantee null termination


snambakam:
### What reproduces the bug?

The cstring_view bpftrace test fails when asserting the null termination.

### `bpftrace --info` output
I was trying bpftrace-v0.21.3

ttreyer:
Thanks for the report!

I was able to reproduce on the current head by defining `_GLIBCXX_DEBUG`.
We should probably enable these extra checks on our CI. I'll fix the test in the meantime.

PR #4003, opened 2025-04-11
Remove `cstring_view`

ttreyer:
The `cstring_view` class was a `std::string_view` with the added guarantee that it is NULL-terminated.
It was only used by BpfMap.

This commit replaces it by a `std::string`

Fixes: #4001


jordalgo (changes requested):
@ttreyer The reason cstring_view exists seems to be to guarantee null termination. I think we need to fix that if it's broken.

ttreyer:
> @ttreyer The reason cstring_view exists seems to be to guarantee null termination. I think we need to fix that if it's broken.

I'll fix cstring_view, but what size should we report then?
e.g.: "abc" currently has a size of 3. Do we want to keep it like this?

jordalgo:
> I'll fix cstring_view, but what size should we report then? e.g.: "abc" currently has a size of 3. Do we want to keep it like this?

This is a good question and I'm not sure the implications of changing this in the bpftrace codebase. In C `sizeof` includes the single null terminal byte but `strlen` does not include this as a character toward the length of the string.

ttreyer:
> `strlen` does not include this as a character toward the length of the string.

The unit tests specifically check that `cstring_view("abc").length() == 3`, so let's keep this semantic.

Would you restart the two failed checks? I don't have the rights and it's a GitHub failure.

jordalgo:
@ttreyer You can just try force pushing again and it should re-trigger the builds

amscanne:
> @ttreyer The reason cstring_view exists seems to be to guarantee null termination. I think we need to fix that if it's broken.

Is it even needed at all? 

It seems to exist only for `BpfMap`, to capture the `const char *` from `bpf_map__name`. But why not just capture using a `std::string`? It will be a few bytes at most (almost certainly stored inline within the `std::string`). And then you can return `const std::string &` from `name()` and `c_str()` already implements the guaranteed semantics for NULL-termination.

The best code is no code :)

danobi (approved):

jordalgo (approved):

PR #4004, opened 2025-04-11
Run CI with hardened stdlib

ttreyer:
Enable hardened-mode for the standard libraries, both C and C++:
- [glibc](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html#fortify-sources-for-unsafe-libc-usage-and-buffer-overflows)
- [libstdc++](https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++.html#precondition-checks-for-c-standard-library-calls)
- [libcxx](https://libcxx.llvm.org/Hardening.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ set(bpftrace_VERSION_PATCH 0)
 include(GNUInstallDirs)
 
 set(WARNINGS_AS_ERRORS OFF CACHE BOOL "Build with -Werror")
+set(HARDENED_STDLIB OFF CACHE BOOL "Enable hardened definitions for standard library (for development use only)")
 set(STATIC_LINKING OFF CACHE BOOL "Build bpftrace as a statically linked executable")
 
 set(BUILD_ASAN OFF CACHE BOOL "Build bpftrace with -fsanitize=address")
@@ -41,6 +42,12 @@ add_compile_options("-Wcast-qual")
 add_compile_options("-Wunreachable-code")
 #add_compile_options("-Wformat=2")
 add_compile_options("-Wdisabled-optimization")
+
+if (HARDENED_STDLIB)
+  add_compile_definitions("-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG")
+  add_compile_definitions("-D_GLIBCXX_ASSERTIONS")
+  add_compile_definitions("-D_FORTIFY_SOURCE=1")
+endif()
 
 if (WARNINGS_AS_ERRORS)
   add_compile_options("-Werror")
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -2,6 +2,13 @@ if(HAVE_BFD_DISASM)
   set(BFD_DISASM_SRC bfd-disasm.cpp)
 endif()
 
+if (HARDENED_STDLIB)
+  add_compile_definitions("-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG")
+  add_compile_definitions("-D_GLIBCXX_ASSERTIONS")
+  add_compile_definitions("-D_FORTIFY_SOURCE=3")
+  add_compile_options("-O1")
+endif()
```
danobi, line 10: seems like this is better placed over here with other compile options: https://github.com/bpftrace/bpftrace/blob/a30d61b56368ef33092fa261c168424168b1bcd3/CMakeLists.txt#L43

src/CMakeLists.txt
```diff
@@ -2,6 +2,13 @@ if(HAVE_BFD_DISASM)
   set(BFD_DISASM_SRC bfd-disasm.cpp)
 endif()
 
+if (HARDENED_STDLIB)
+  add_compile_definitions("-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_DEBUG")
+  add_compile_definitions("-D_GLIBCXX_ASSERTIONS")
+  add_compile_definitions("-D_FORTIFY_SOURCE=3")
+  add_compile_options("-O1")
```
danobi, line 9: Not sure forcing `-O1` here is right. This messes with `CMAKE_BUILD_TYPE` and associated CI matrix for it. Seems like `-O1` is required for `_FORTIFY_SOURCE` > 1. Maybe just stick to `=1`?

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set(bpftrace_VERSION_PATCH 0)
 include(GNUInstallDirs)
 
 set(WARNINGS_AS_ERRORS OFF CACHE BOOL "Build with -Werror")
+set(HARDENED_STDLIB OFF CACHE BOOL "Enable hardened definitions for standard library")
```
danobi, line 14: seems like these options are for development use only? not sure `_LIBCPP_HARDENING_MODE_DEBUG` is suitable for production usage according to the docs.
```suggestion
set(HARDENED_STDLIB OFF CACHE BOOL "Enable hardened definitions for standard library (for development use only)")
```

src/CMakeLists.txt
ttreyer: The `-O1` option was breaking the tests. I'll remove it as you suggested below and move the options to the top level!

danobi (approved):

PR #4005, opened 2025-04-13
Add map "spec" for call validation

jordalgo:
This creates a map that lays out a simple spec
for `Call` checks. It removes boilerplate calls
to `check_assignment`, `check_varargs`,
`check_nargs`, and `check_arg`.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -135,10 +135,8 @@ private:
   bool is_final_pass() const;
   bool is_first_pass() const;
 
-  bool check_assignment(const Call &call,
-                        bool want_map,
-                        bool want_var,
-                        bool want_map_key);
+  void check_assignment(const Call &call, bool want_map, bool want_var);
+  [[nodiscard]] bool check_call(const Call &call);
   [[nodiscard]] bool check_nargs(const Call &call, size_t expected_nargs);
   [[nodiscard]] bool check_varargs(const Call &call,
                                    size_t min_nargs,
@@ -146,8 +144,7 @@ private:
   bool check_arg(const Call &call,
                  Type type,
                  int arg_num,
-                 bool want_literal = false,
-                 bool fail = true);
+                 bool want_literal = false);
   bool check_symbol(const Call &call, int arg_num);
   bool check_available(const Call &call, const AttachPoint &ap);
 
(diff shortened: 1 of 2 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -135,19 +135,19 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
   bool is_final_pass() const;
   bool is_first_pass() const;
 
-  bool check_assignment(const Call &call,
```
jordalgo, line 138: This return value was never checked

src/ast/passes/semantic_analyser.cpp
```diff
@@ -135,19 +135,19 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+  [[nodiscard]] bool check_call(const Call &call);
   [[nodiscard]] bool check_nargs(const Call &call, size_t expected_nargs);
   [[nodiscard]] bool check_varargs(const Call &call,
                                    size_t min_nargs,
                                    size_t max_nargs);
   bool check_arg(const Call &call,
                  Type type,
                  int arg_num,
-                 bool want_literal = false,
-                 bool fail = true);
```
jordalgo, line 150: There was only one call site using this last `fail` param. The site no longer calls this function.

tests/semantic_analyser.cpp
```diff
@@ -787,9 +787,6 @@ stdin:1:12-21: ERROR: lhist() should be directly assigned to a map
 kprobe:f { @[lhist()] = 1; }
            ~~~~~~~~~
 stdin:1:12-21: ERROR: lhist() requires 4 arguments (0 provided)
-kprobe:f { @[lhist()] = 1; }
-           ~~~~~~~~~
-stdin:1:12-21: ERROR: lhist_t cannot be used as a map key
```
jordalgo, line 792: this error is not needed as the one above it "lhist() should be directly assigned to a map" covers it

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +216,303 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+struct arg_type_spec {
+  Type type = Type::integer;
+  bool literal = false;
+  // This means that this check is only valid if there are
+  // this exact number of args in the call
+  size_t nargs = 0;
+  // This indicates that this is just a placeholder
+  // as we use the index in the vector of arg_type_spec
+  // as the number argument to check
+  bool no_check = false;
```
viktormalik, line 260: Why not use the arg index here directly? IIUC, it's not used for anything else than the check itself and it would allow us to only specify and iterate indices which need checking.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +216,303 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+};
+
+struct arg_spec {
+  size_t min;
+  size_t max;
+  // NOLINTBEGIN(readability-redundant-member-init)
+  std::vector<arg_type_spec> types = {};
+  // NOLINTEND(readability-redundant-member-init)
```
viktormalik, line 268: Is the initializer that important that using it is worth the `NOLINTBEGIN`/`NOLINTEND` lines?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +216,303 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+  { "time", {} },
+  { "zero", {} },
+};
+
+struct arg_type_spec {
+  Type type = Type::integer;
+  bool literal = false;
+  // This means that this check is only valid if there are
+  // this exact number of args in the call
+  size_t nargs = 0;
```
viktormalik, line 256: This is a bit confusing (although the comment helps). I'd either consider renaming the field or dropping it completely as it's only applicable in 3 cases. And we still have other places where we need to call `check_arg` explicitly (like kstack/ustack) so maybe let's just handle these cases where the check depends on the call context separately.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +216,303 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+  { "len",
+    { .min=1,
+      .max=1 } },
+  { "lhist",
+    { .min=4,
+      .max=4,
+      .types={ { .type=Type::integer },
+        { .type=Type::integer, .literal=true },
+        { .type=Type::integer, .literal=true },
+        { .type=Type::integer, .literal=true } } } },
```
viktormalik, line 363:
```suggestion
      .types={
        { .type=Type::integer },
        { .type=Type::integer, .literal=true },
        { .type=Type::integer, .literal=true },
        { .type=Type::integer, .literal=true } } } },
```

viktormalik (commented):
Love the refactoring! Just a few comments.

I'm also thinking about a nicer formatting of the specs but haven't been able to come up with something substantially better than what you have :grinning:

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +216,303 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
 
 } // namespace
 
+struct assign_checks {
+  bool want_map = false;
+  bool want_var = false;
+  bool want_map_key = false;
```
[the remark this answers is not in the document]
jordalgo: Yup, looks like it. Good catch.

src/ast/passes/semantic_analyser.cpp
jordalgo: @viktormalik Can you explain a bit more. I don't quite understand what you're saying. Also if you look at the spec, you'll see where `no_check` is used (which is only a few places). I had originally tried to define the types field as `std::vector<std::optional<arg_type_spec>> types` to avoid this field but then I would have to wrap all non `nullopt` values in `make_optional`, which was gross. I also originally had a `num` field in this struct to indicate the arg index but thought it was also kind of extra cruft when I could just use the index of into the array.

src/ast/passes/semantic_analyser.cpp
viktormalik: :facepalm:

src/ast/passes/semantic_analyser.cpp
viktormalik: Yeah, I meant having an explicit `num` field in the struct.

It's just that to me, it'd easier to "parse" that large initializer in my head if every argument had its index explicitly specified. But I understand that using the vector index is less code and that we don't have that many places where `no_check` is used so I'm not going to push this as it may be just my perception.

Also, if we keep it, maybe `skip_check` would be a bit better?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +216,303 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+      .max=2,
+      .types={
+        { .no_check=true },
+        { .type=Type::string, .literal=true } } } },
+  { "kaddr",
+    { .min=1,
+      .max=1,
+      .types={
+        { .type=Type::string, .literal=true },
+        { .type=Type::integer, .literal=true } } } },
```
viktormalik, line 341: How come `kaddr` has 1 argument but 2 constraints?

src/ast/passes/semantic_analyser.cpp
jordalgo: I like `skip_check`.

src/ast/passes/semantic_analyser.cpp
jordalgo: Bad copy pasta :) good catch.

jordalgo:
@viktormalik I decided to combine the two maps into one.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +213,300 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+  // as we use the index in the vector of arg_type_spec
+  // as the number argument to check
+  bool skip_check = false;
+};
+
+struct arg_spec {
+  size_t min = 0;
+  size_t max = 0;
+  bool check_assign = false;
+  AssignWant assign = AssignWant::NONE;
```
viktormalik, line 236: Naming super-duper-nit (I just cannot help myself :grinning:):

Since the struct is called `arg_spec`, i.e. a specification of a builtin, having the word "check" there is kind of confusing. How about something like:
```suggestion
  bool has_return = false;
  ReturnKind return_kind = ReturnKing::NONE;
```

viktormalik (commented):
Looks even better now!

Just one note, otherwise LGTM.

src/ast/passes/semantic_analyser.cpp
jordalgo: Ah, well I'm also going to change `arg_spec` to `call_spec`. And yeah, good call, I'll make those fields refer to the return.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +213,300 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+  bool literal = false;
+  // This indicates that this is just a placeholder
+  // as we use the index in the vector of arg_type_spec
+  // as the number argument to check
+  bool skip_check = false;
+};
+
+struct call_spec {
+  size_t min = 0;
+  size_t max = 0;
```
viktormalik, line 234: Sorry for more comments but now we should probably change these
```suggestion
  size_t args_min = 0;
  size_t args_max = 0;
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +213,300 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+  bool has_return = false;
+  ReturnKind return_kind = ReturnKind::NONE;
+  // NOLINTBEGIN(readability-redundant-member-init)
+  std::vector<arg_type_spec> types = {};
```
viktormalik, line 238: and this one
```suggestion
  std::vector<arg_type_spec> arg_types = {};
```

viktormalik (approved):

viktormalik:
Great job!

jordalgo:
@viktormalik Thanks for the thorough review. You always catch good stuff!

src/ast/passes/semantic_analyser.cpp
```diff
@@ -216,6 +213,300 @@ class SemanticAnalyser : public Visitor<SemanticAnalyser> {
...
+struct arg_type_spec {
+  Type type = Type::integer;
+  bool literal = false;
+  // This indicates that this is just a placeholder
+  // as we use the index in the vector of arg_type_spec
+  // as the number argument to check
+  bool skip_check = false;
+};
+
+struct call_spec {
```
amscanne, line 232: Can we reconcile this with the `FunctionRegistry`?

jordalgo: @amscanne Are we still planning on keeping the FunctionRegistry?

Issue #3986, opened 2025-04-07
0.23.0 fails to build on Alpine


acj:
I'm currently seeing [this build error](https://github.com/acj/aport-builds/actions/runs/14305129235/job/40087271141) on both 0.23.0 and the tip of master (currently d5653cf):

```
ninja: job failed: /usr/bin/g++ -DKERNEL_HEADERS_DIR=\"\" -I/home/builder/package/src/bpftrace-d5653cf8c1e2519cbbf37978d17f855b6323e438/src -I/home/builder/package/src/bpftrace-d5653cf8c1e2519cbbf37978d17f855b6323e438/build -isystem /usr/lib/llvm18/include -Os -fstack-clash-protection -Wformat -Werror=format-security -D_GLIBCXX_ASSERTIONS=1 -D_LIBCPP_ENABLE_THREAD_SAFETY_ANNOTATIONS=1 -D_LIBCPP_ENABLE_HARDENED_MODE=1 -fno-plt -Os -DNDEBUG -std=c++20   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -Wall -Wextra -Werror=missing-field-initializers -Werror=undef -Wpointer-arith -Wcast-align -Wwrite-strings -Wcast-qual -Wunreachable-code -Wdisabled-optimization -fdiagnostics-color=always -MD -MT src/util/CMakeFiles/util.dir/paths.cpp.o -MF src/util/CMakeFiles/util.dir/paths.cpp.o.d -o src/util/CMakeFiles/util.dir/paths.cpp.o -c /home/builder/package/src/bpftrace-d5653cf8c1e2519cbbf37978d17f855b6323e438/src/util/paths.cpp
In file included from /home/builder/package/src/bpftrace-d5653cf8c1e2519cbbf37978d17f855b6323e438/src/log.h:11,
                 from /home/builder/package/src/bpftrace-d5653cf8c1e2519cbbf37978d17f855b6323e438/src/util/paths.cpp:10:
/home/builder/package/src/bpftrace-d5653cf8c1e2519cbbf37978d17f855b6323e438/src/ast/location.h:9:10: fatal error: location.hh: No such file or directory
    9 | #include "location.hh"
      |          ^~~~~~~~~~~~~
compilation terminated.
```

There are a few old issues referring to location.hh as an ephemeral file that's generated during the build. Is that still the case? If so, maybe there's an order of operations issue?

[...]

PR #4006, opened 2025-04-13
Fix build error on Alpine 32-bit arches

acj:
Following #3986, which addressed a build error for x86_64, I'm also seeing errors on Alpine for armhf and armv7 architectures:

```
src/ast/passes/codegen_llvm.cpp: In member function 'bpftrace::ast::ScopedExpr bpftrace::ast::CodegenLLVM::visit(bpftrace::ast::Call&)':
src/ast/passes/codegen_llvm.cpp:1373:29: error: no matching function for call to 'min(<brace-enclosed initializer list>)'
 1373 |     uint64_t size = std::min({ static_cast<uint64_t>(*size_opt),
      |                     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1374 |                                left_arg.type.GetSize(),
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~
 1375 |                                right_arg.type.GetSize() });
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This PR adds a couple of casts to make sure all the args have the same type, which then satisfies the template function. My C++ fu is fairly weak these days, so please let me know if there's a different type of cast I should be using.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ and this project adheres to
 #### Fixed
 - Fix build failures due to missing location.hh
   - [#3987](https://github.com/bpftrace/bpftrace/pull/3987)
+- Fix 32-bit build failures due to missing cast
+  - [#4006](https://github.com/bpftrace/bpftrace/pull/4006)
 #### Security
 #### Docs
 #### Tools
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -1715,9 +1715,10 @@ ScopedExpr CodegenLLVM::visit(Call &call)
     auto size_opt = bpftrace_.get_int_literal(call.vargs.at(2));
     if (!size_opt.has_value())
       LOG(BUG) << "Int literal should have been checked in semantic analysis";
-    uint64_t size = std::min({ static_cast<uint64_t>(*size_opt),
-                               left_arg.type.GetSize(),
-                               right_arg.type.GetSize() });
+    uint64_t size = std::min(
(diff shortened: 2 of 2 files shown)


jordalgo:
This change looks good to me. I wonder if we can enable stricter checks on these things so you don't see as many downstream build errors on Alpine.

jordalgo (approved):

jordalgo:
@acj Thanks for the fix!

jordalgo:
@viktormalik Think we should make another patch release for this? The problem is also on the 0.23 release branch

viktormalik:
> @viktormalik Think we should make another patch release for this? The problem is also on the 0.23 release branch

Yeah, since bpftrace is built for 32-bit arches on Alpine, we should make another patch release.

viktormalik:
> This change looks good to me. I wonder if we can enable stricter checks on these things so you don't see as many downstream build errors on Alpine.

Should we at least add Alpine to our [distros CI workflow](https://github.com/bpftrace/bpftrace/blob/master/.github/workflows/distros.yml#L19-L21)? That would only catch the x86_64 build errors but it's at least something (if I recall correctly, we've had issues with Alpine builds on older releases, too).

As for the 32-bit builds, I assume that the same failure could happen on other 32-bit arches. Perhaps we could add a 32-bit x86 CI job if Nix supports it.

jordalgo:
> Yeah, since bpftrace is built for 32-bit arches on Alpine, we should make another patch release.

I can take care of that.

PR #4011, opened 2025-04-14
clang_tidy.sh: update to clang-tidy 20 and disable modernize-use-integer-sign-comparison

JakeHillion:
The clang-tidy script was recently held back to LLVM-19 in #3968 because it provided some annoying lints. Disable `modernize-use-integer-sign-comparison` which is pretty noisy in the codebase and has non obvious fixes, and apply the rest (`.contains` and `.ends_with`).

Did start work on the `modernize-use-integer-sign-comparison`, but I really think we should leave it disabled. It would be good if the things it complains about weren't used in new code, but every change to the old code to make it happy seems unnecessarily risky, especially given most of the cases are perfectly valid (if ugly) static casts anyway.

##### Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -16,6 +16,7 @@ Checks:
   - modernize-*
   - -modernize-avoid-c-arrays
   - -modernize-use-nodiscard
+  - -modernize-use-integer-sign-comparison
   - -modernize-use-trailing-return-type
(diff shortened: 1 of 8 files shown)


jordalgo (approved):
LGTM!

viktormalik (approved):

PR #3622, opened 2024-12-03
Tools: switch all bio* tools to tracepoints

viktormalik:
Most `bio*` tools attach to `__blk_account_io_start`/`__blk_account_io_done` kprobes which were removed in https://github.com/torvalds/linux/commit/e165fb4dd698 and https://github.com/torvalds/linux/commit/06965037ce94. On top of that, the only remaining functions `blk_account_io_start`/`blk_account_io_done` are typically inlined so there are no more kprobes to attach the tools to.

Fortunately, since https://github.com/torvalds/linux/commit/5a80bd075f3b, there are tracepoints in all of the functions that we trace so we can switch to use them.

The main problem with this switch is that `biosnoop.bt` no longer has access to disk names, only to disk ids. BCC does the translation via `/proc/diskstats` which is not possible in bpftrace without introducing a new async builtin. So, in order not to complicate things, this switches `biosnoop.bt` to print disk ids rather than disk names and lets users do the translation themselves.

The `biolatency-kp` is left to use kprobes (as that's the point), just adding a warning that it may not work on new kernels.

See individual commits messages for more details.

Resolves #3621.

##### Checklist

- [ ] Language changes are updated in `man/adoc/bpftrace.adoc`
- [x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ ] The new behaviour is covered by tests

PR #4013, opened 2025-04-15
biosnoop.bt: Print comm from block_io_start probe

### LINKED ISSUE #4010: biosnoop.bt and args.comm
biosnoop.bt was changed recently to use tracepoints instead of kprobes

```
tracepoint:block:block_io_start
tracepoint:block:block_io_done

```
both tracepoints have an arg called 'comm', but in my testing, 'block_io_start' and 'block_io_done' show a different 'comm' (for the same block io request).

An example 'block_io_start' shows 'kworker/u48:2', but 'block_io_done' shows 'swapper/7'
As it stands, 'biosnoop.bt' does not show the 'comm' for 'block_io_start', so everything says "swapper/x".
The BCC version of the tool shows "kworker/x".

Testing in RHEL 9.5

viktormalik:
#3622 has changed biosnoop.bt to use tracepoints instead of unstable kprobes. However, it also changed the way process name (`comm`) is printed - instead of printing `comm` from the `block_io_start` probe, it prints the one from the `block_io_done` probe. Apparently, these two tracepoints can be called by different kernel threads and therefore the thread name can be different.

Return to printing `comm` from the `block_io_start` probe to keep consistent with the previous (documented) version as well as with the BCC and libbpf versions of the same tool.

Fixes #4010.

##### Checklist

### DIFF
--- a/tools/biosnoop.bt
+++ b/tools/biosnoop.bt
@@ -20,10 +20,13 @@ tracepoint:block:block_io_start
 	$key = (args.dev, args.sector);
 	@start[$key] = nsecs;
 	@iopid[$key] = pid;
+	@iocomm[$key] = comm;
 }
 
 tracepoint:block:block_io_done
-/@start[args.dev, args.sector] != 0 && @iopid[args.dev, args.sector] != 0 && args.comm != ""/
+/@start[args.dev, args.sector] != 0 &&
+ @iopid[args.dev, args.sector] != 0 &&
+ @iocomm[args.dev, args.sector] != ""/
 
 {
 	$key = (args.dev, args.sector);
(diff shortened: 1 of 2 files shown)


viktormalik:
@jordalgo since the tool was broken in 0.23.0, I think that we should include this fix in 0.23.2.

jordalgo:
@viktormalik Darn, wish I saw this before I just merged the release commit on that branch. Should be ok though, I'll just cherry-pick this change and add it to the release changelog.

tools/biosnoop.bt
```diff
@@ -20,26 +20,31 @@ tracepoint:block:block_io_start
...
 	@start[$key] = nsecs;
 	@iopid[$key] = pid;
+	@iocomm[$key] = comm;
 }
 
 tracepoint:block:block_io_done
-/@start[args.dev, args.sector] != 0 && @iopid[args.dev, args.sector] != 0 && args.comm != ""/
+/@start[args.dev, args.sector] != 0 &&
+ @iopid[args.dev, args.sector] != 0 &&
+ @iocomm[args.dev, args.sector] != ""/
```
jordalgo, line 29: supernit while you're here, can we use the `has_key` func for these two?

viktormalik:
> @viktormalik Darn, wish I saw this before I just merged the release commit on that branch. Should be ok though, I'll just cherry-pick this change and add it to the release changelog.

Yeah, since the release is not out, yet, it's IMO ok.

tools/biosnoop.bt
viktormalik: Good idea, will do.

viktormalik: Or, you know what, let's just do it in a separate PR (which doesn't have to go to the release branch) as `has_key` can be used in a bunch of other tools.

jordalgo: Sure - sounds good.

jordalgo (approved):

PR #4028, opened 2025-04-17
codegen: disable unwind generation

amscanne:
codegen: disable unwind generation

This causes issues during BPF linking. There's no way to use this
currently, so no harm in disabling.


jordalgo (approved):
Seems reasonable.

PR #4031, opened 2025-04-17
fix: include information when loading fails

amscanne:
Stacked PRs:
 * #4034
 * #4033
 * #4032
 * __->__#4031
 * #4030
 * #4029
 * #4028

--- --- ---

### fix: include information when loading fails

Signed-off-by: Adin Scannell <amscanne@meta.com>

PR #4032, opened 2025-04-17
fix: ensure error output when not in diagnostic form

amscanne:
Stacked PRs:
 * #4034
 * #4033
 * __->__#4032
 * #4031
 * #4030
 * #4029
 * #4028

--- --- ---

### fix: ensure error output when not in diagnostic form

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -787,7 +787,10 @@ int main(int argc, char* argv[])
                   .add(ast::CreateMapSugarPass())
                   .add(ast::CreateSemanticPass(args.listing))
                   .run();
-    if (!ok || !ast.diagnostics().ok()) {
+    if (!ok) {
+      std::cerr << ok.takeError() << "\n";
+      return 2;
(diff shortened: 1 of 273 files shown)


src/main.cpp
```diff
@@ -787,7 +787,10 @@ int main(int argc, char* argv[])
                   .add(ast::CreateMapSugarPass())
                   .add(ast::CreateSemanticPass(args.listing))
                   .run();
-    if (!ok || !ast.diagnostics().ok()) {
+    if (!ok) {
+      std::cerr << ok.takeError() << "\n";
+      return 2;
```
jordalgo, line 792: why return 2?

jordalgo (approved):

src/main.cpp
amscanne: **¯\_(ツ)_/¯** I can make it one, I just wanted to distinguish the error cases (user error vs. internal error). I think it would be good to take a separate pass, make sure this was consistent and just document somewhere.

PR #4034, opened 2025-04-17
ast: add basic import resolution and object linking

amscanne:
Stacked PRs:
 * #4051
 * #4050
 * #4049
 * __->__#4034

--- --- ---

### ast: add basic import resolution and object linking

Currently this only works on paths that are relative to the script
itself as well as the embedded standard library. Additional search paths
can be provided from the caller, but this is not done at the moment.

The `stdlib` package is always implicitly imported, but when implicitly
imported it will be only be sourced from the embedded contents. It can
be overriden by an external file/directory, but only with an explicit
import.


danobi:
Overall looks good. Looked at everything except actual import search logic. Will take a closer look at that tonight/tmr

src/ast/passes/resolve_imports.cpp
```diff
@@ -1,29 +1,242 @@
...
+
+  // Import supported extensions.
+  std::filesystem::path path(name);
+  if (path.extension() == ".bt") {
+    return importNative(node, name, data);
+  } else if (path.extension() == ".bc") {
+    return importBitcode(node, name, data);
+  } else if (path.extension() == ".btf") {
+    // Ignore for now: these are BTF types.
+    return OK();
```
danobi, line 182: If this is a TODO note, error instead?

src/ast/passes/resolve_imports.cpp
```diff
@@ -1,29 +1,242 @@
...
+        continue; // No file found.
+      }
+
+      // Attempt the import.
+      return importAny(node, name, path);
+    }
+  }
+
+  // See if this matches a set of builtins. Note that we do the "directory"
+  // expansion here, importing anything that is matching as a path.
```
danobi, line 211: Should this be inserted into `checked`?

danobi (commented):
Quite a bit of logic in ResolveImports resolutions. Should be unit testable with some TempDirs and fixtures, right? Would be good to have some coverage on this logic. We've already had one CVE on bad import logic: https://nvd.nist.gov/vuln/detail/CVE-2024-2313

src/ast/passes/resolve_imports.cpp
amscanne: This is the set of files that are effectively built in to the binary itself, so there's nothing under the user control. I thought about including the list of builtins, but I think we would want to surface that in another way.

src/ast/passes/resolve_imports.cpp
amscanne: Yep, just dropped this case.

amscanne:
> Quite a bit of logic in ResolveImports resolutions. Should be unit testable with some TempDirs and fixtures, right? Would be good to have some coverage on this logic. We've already had one CVE on bad import logic: https://nvd.nist.gov/vuln/detail/CVE-2024-2313

Yeah, I think we probably want to have a simple abstract interface for accessing the file system broadly.  Could be fed as an input to the pass manager, used directly for header resolution, etc.

I will put something simple together for this specific change to ensure that there's some basic coverage, but I think we need something bigger for testing with imports as a general feature.

src/ast/passes/resolve_imports.cpp
danobi: I was thinking more if someone tries to include the standard library but typos, the following error would be a good hint about the correct spelling.

danobi (commented):
ty for the tests!

tests/imports.cpp
```diff
@@ -0,0 +1,198 @@
...
+  bool set_perms(const std::string &path,
+                 std::optional<std::filesystem::perms> perms = std::nullopt)
+  {
+    if (perms) {
+      std::error_code ec = {};
+      std::filesystem::permissions(dir->path() / path,
+                                   *perms,
+                                   std::filesystem::perm_options::replace,
+                                   ec);
+      EXPECT_FALSE(bool(ec));
```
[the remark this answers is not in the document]
amscanne: `ASSERT_*` will yield a naked `return` and can only be used in `void` contexts. It should only be used at the top-level test function (and it is in this case). I think maybe I should just drop the `EXPECT` here and annotate with `nodiscard`?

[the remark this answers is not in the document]
danobi: Ah, ok. And yeah, that works. Just keep all the expects/assertions in top level is clean.

src/ast/passes/resolve_imports.cpp
amscanne: I've added `utils/similar.h`, which includes a helper to give a "similarly score" for two strings. If they are similar, they will be included and added as a hint. To do this, I also added support for multiple hints.

tests/imports.cpp
```diff
@@ -0,0 +1,198 @@
...
+  void TearDown() override
+  {
+    files.clear();
+    dir.reset();
+  }
+
+  bool set_perms(const std::string &path,
+                 std::optional<std::filesystem::perms> perms = std::nullopt)
+  {
+    if (perms) {
```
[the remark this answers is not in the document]
amscanne: Done

tests/imports.cpp
```diff
@@ -0,0 +1,198 @@
...
+      parent = parent.parent_path();
+      bool ok = create_dir(parent.string());
+      EXPECT_TRUE(ok);
+      if (!ok) {
+        return false;
+      }
+    }
+
+    // Create the file at the given path.
+    auto ok = dir->create_file(path, false);
```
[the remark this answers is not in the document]
amscanne: I've clarified the doc comment there. This is now an allowed usage ;)

amscanne:
@jordalgo @danobi This one should now have issues addressed and be ready to go

src/ast/passes/resolve_imports.cpp
```diff
@@ -1,40 +1,326 @@
...
+  if (imports.scripts.contains(name)) {
+    return OK(); // Already added.
+  }
+
+  // Load the file.
+  std::ifstream file(path);
+  if (file.fail()) {
+    node.addError() << "error reading import '" << path
+                    << "': " << std::strerror(errno);
+    return OK();
```
jordalgo, line 95: JC, why is a failure returning OK?

src/ast/passes/resolve_imports.cpp
```diff
@@ -1,40 +1,326 @@
...
+}
+
+Result<OK> ResolveImports::importAny(Node &node,
+                                     const std::string &name,
+                                     const std::filesystem::path &path,
+                                     bool ignore_unknown)
+{
+  if (!checkPerms(path)) {
+    node.addError()
+        << "cowardly refusing to import global writable permissions: " << path;
```
jordalgo, line 173: lol

src/ast/passes/resolve_imports.cpp
```diff
@@ -1,40 +1,326 @@
...
+      if (!ok) {
+        return ok.takeError();
+      }
+    }
+    return OK();
+  } else {
+    // Import any support file-based extensions.
+    if (path.extension() == ".bt") {
+      return importScript(node, name, path);
+    } else if (path.extension() == ".o") {
```
jordalgo, line 193: Is it silly to be even more strict here and only look for something like `bpf.o` or `bt.o` ?

src/ast/passes/resolve_imports.cpp
jordalgo: > I was thinking more if someone tries to include the standard library but typos

Are we not including the standard library by default for all scripts? Or am I confused.

tests/imports.cpp
```diff
@@ -0,0 +1,192 @@
...
+  // same result as importing the standard library implicitly.
+  test(R"(import "stdlib"; BEGIN {})", {}, fn);
+}
+
+TEST_F(ImportTest, stdlib_implicit_rules)
+{
+  auto dir = TempDir::create();
+  ASSERT_TRUE(bool(dir));
+
+  // If we have a `stdlib` directory, it should override the explicit import.
```
jordalgo, line 130: i'm confused about this logic - is this how things are normally done in terms of override priority?

src/ast/passes/resolve_imports.cpp
amscanne: Correctly -- if it is included implicitly, then it will always be sourced from the embedded version. If users import explicitly, then it will do the standard file system search (same as used for all explicit imports).

src/ast/passes/resolve_imports.cpp
amscanne: I failure means that the pass has actually failed, and it will super-ceded any other diagnostics. I don't think there are clear guidelines for when to overtly fail vs. adding an error diagnostic. Broadly, diagnostics should probably be a user-error with the script, whereas an error should be "something not part of the script went wrong". This is kind of a weird borderline case. WDYT? I'm happy to do either.

src/ast/passes/resolve_imports.cpp
amscanne: Sure, I can do `bpf.o`. But `.bt` is the correct one, for bitcode files. It's not an object.

src/ast/passes/resolve_imports.cpp
```diff
@@ -1,40 +1,326 @@
...
+{
+  std::vector<std::string> checked;
+  if (allow_paths) {
+    for (const auto &import_path : paths_) {
+      // Check to see if this specific file exists.
+      auto path = import_path / name;
+      checked.emplace_back(path.string());
+      std::error_code ec;
+      if (!std::filesystem::exists(path, ec)) {
+        continue; // No file found.
```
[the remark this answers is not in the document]
amscanne: Okay, this made me laugh quite a lot. I will remove the comment. Just to explain myself: I just have a habit of having comments on control flow changes to help with readability, because my eyes hit those quickly and you don't need to absorb details on exactly which function is being called, etc. This one might be lower-information than usual, but the `// Already added` comment above is an example of where this is quite useful, despite this also being quite obvious if you read the code carefully :)

tests/imports.cpp
amscanne: Sorry, this should say `implicit`. I just tweaked the comment.

If you have an explicit `import "stdlib"` then it use the regular search for the library. If you don't have an explicit `import "stdlib"` then it will never search the filesystem. This is to prevent strange surprises and unexpected behavior.

src/ast/passes/resolve_imports.cpp
amscanne: Thinking about it a bit more, if you can strongly associate the error with a specific AST node, then it probably makes sense to keep it as a diagnostic (and not fail the pass).

In this case, it is a permission error directly tied to that import. If it were less relevant, e.g. an error bubbling up from `TempDir::create` in building the import, I'd say that returning an error would make more sense (since it has nothing to do with that specific node).

src/ast/passes/resolve_imports.cpp
jordalgo: Yeah let's maybe do `bpf.o`

src/ast/passes/resolve_imports.cpp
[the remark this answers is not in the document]
jordalgo: > I just have a habit of having comments on control flow changes to help with readability,

I understand. I like when things make it easy to scan code - this is why I like early returns in a function (as opposed to if nesting). My biggest beef with comments is this requirement to keep them updated as the code changes - I especially love it when they get out of sync with the actual code and then you can't tell which one is correct. Anyway, this also is not blocking. Just being a wiseass.

src/ast/passes/resolve_imports.cpp
amscanne: Yes, if an error is returned, all passes cease and the error is propagated back up to the top-level. Since it's not a diagnostic, I think the logical thing would be to print it and exit. Put another way: the error is essentially "I was not able to run this pass due to some externality" (so the diagnostics should not be considered complete) vs. running the pass and successfully producing errors.

Diagnostics are pretty flexible, we stop passes right now in order to avoid overwhelming the user with errors. But we could imagine just running the next pass until we have >N diagnostics instead. But if a pass produces a error, we would *not* be able to produce because the pass did not produce its required output.

I am borderline on this case, because the import is going to make the program non-functional for other reasons. I think in this case I'd bias towards making it a diagnostic and allowing the pass to succeed, just to highlight the specific import that led to the error, but happy to change.

src/ast/passes/resolve_imports.cpp
amscanne: Done. Note that we need a lot more tests for this broadly, but I don't really want to load it all into this PR. I think it's reasonable to gate having imports be an "official" feature on this, I've really just added some minimal tests to ensure that the path resolution has some of the intended behaviors.

src/ast/passes/resolve_imports.cpp
jordalgo: Sounds good to me. Thanks for explaining.

jordalgo (approved):

PR #4037, opened 2025-04-18
Add hygienic macros

jordalgo:
Hygienic macros (usually just "macros") provide
a way for you to structure your script. They
can be useful when you want to factor out code
into smaller, more understandable parts. Or if
you want to share code between probes.

At a high level, macros can be thought of as
semantic aware text replacement. They accept
(optional) variable, map, and some expression
arguments. The body of the macro may only
access maps and variables passed in through
the arguments (the hygiene). The body of the
macro is exactly one block expression.

Example:
```
macro set($x) { $x += 1; $x }

BEGIN { $a = 1; set($a); print($a); exit(); }
```

This will print "2";

This feature is gated behind a config flag:
"unstable_macro".

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -978,6 +978,91 @@ The following snippets create a map with key signature `(int64, string[16])` and
 @[(pid, comm)]++
 ----
 
+==== Macros
+
+bpftrace macros (as opposed to C macros) provide a way for you to structure your script.
+They can be useful when you want to factor out code into smaller, more understandable parts.
+Or if you want to share code between probes.
+
+At a high level, macros can be thought of as semantic aware text replacement.
+They accept (optional) variable, map, and some expression arguments.
+The body of the macro may only access maps and external variables passed in through the arguments, which is why these are often referred to as "hygienic macros".
+The body of the macro is exactly one block expression.
+
+For example, these are valid usages of macros:
+
+----
+config = {
+  unstable_macro=1;
+}
+
+macro one() {
+  1
+}
+
+macro add_one($other_name) {
+  $other_name + 1
+}
+
+macro add_one_to_each($a, @b) {
+  $a += 1;
+  @b += 1;
+  $a + @b
+}
+
+macro side_effect($x) {
+  print($x)
+}
+
+macro add_two($x) {
+  add_one($x) + 1
+}
+
+BEGIN {
+  print(one());                   // prints 1
+
+  $a = 42;
+  print(add_one($a));             // prints 43
+
+  @b = 3;
+  print(add_one_to_each($a, @b)); // prints 47
+
+  side_effect(5)                  // prints 5
+
+  print(add_two(1));              // prints 3
+}
+----
+
+Some examples of invalid macro usage:
+
+----
+config = {
+  unstable_macro=1;
+}
+
+macro not_expression() {
+  $var = 1;                    // BAD: Not an expression
+}
+
+macro unhygienic_access() {
+  @x++                         // BAD: @x not passed in
+}
+
+macro wrong_parameter_type($x) {
+  $x++
+}
+
+BEGIN {
+  @x = 1;
+  unhygienic_access();
+
+  wrong_parameter_type(@x);    // BAD: macro expects a scratch variable
+}
+----
+
+**Warning** this feature is experimental and may be subject to changes.
+It also requires the 'unstable_macro' config being set to 1.
+
 ==== Per-Thread Variables
 
 These can be implemented as a map keyed on the thread ID. For example, `@start[tid]`:
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -1175,6 +1175,29 @@ public:
 };
 using ImportList = std::vector<Import *>;
 
+class Macro : public Node {
+public:
+  Macro(ASTContext &ctx,
+        std::string name,
+        ExpressionList &&vargs,
+        Block *block,
+        Location &&loc)
+      : Node(ctx, std::move(loc)),
+        name(std::move(name)),
+        vargs(std::move(vargs)),
+        block(block) {};
+  explicit Macro(ASTContext &ctx, const Macro &other, const Location &loc)
+      : Node(ctx, loc + other.loc),
+        name(other.name),
+        vargs(clone(ctx, other.vargs, loc)),
+        block(clone(ctx, other.block, loc)) {};
+
+  std::string name;
+  ExpressionList vargs;
+  Block *block;
+};
+using MacroList = std::vector<Macro *>;
+
 class Program : public Node {
 public:
   explicit Program(ASTContext &ctx,
@@ -1182,6 +1205,7 @@ public:
                    Config *config,
                    ImportList &&imports,
                    MapDeclList &&map_decls,
+                   MacroList &&macros,
                    SubprogList &&functions,
                    ProbeList &&probes,
                    Location &&loc)
@@ -1190,6 +1214,7 @@ public:
         config(config),
         imports(std::move(imports)),
         map_decls(std::move(map_decls)),
+        macros(std::move(macros)),
         functions(std::move(functions)),
         probes(std::move(probes)) {};
   explicit Program(ASTContext &ctx, const Program &other, const Location &loc)
@@ -1198,6 +1223,7 @@ public:
         config(clone(ctx, other.config, loc)),
         imports(clone(ctx, other.imports, loc)),
         map_decls(clone(ctx, other.map_decls, loc)),
+        macros(clone(ctx, other.macros, loc)),
         functions(clone(ctx, other.functions, loc)),
         probes(clone(ctx, other.probes, loc)) {};
 
@@ -1205,6 +1231,7 @@ public:
   Config *config = nullptr;
   ImportList imports;
   MapDeclList map_decls;
+  MacroList macros;
   SubprogList functions;
   ProbeList probes;
 };
--- a/src/ast/passes/macro_expansion.cpp
+++ b/src/ast/passes/macro_expansion.cpp
@@ -1,0 +1,318 @@
+#include <format>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+
+#include "ast/ast.h"
+#include "ast/context.h"
+#include "ast/passes/macro_expansion.h"
+#include "ast/visitor.h"
+#include "bpftrace.h"
+
+#include "log.h"
+
+namespace bpftrace::ast {
+
+class MacroExpander : public Visitor<MacroExpander> {
+public:
+  MacroExpander(ASTContext &ast,
+                std::string macro_name,
+                const std::unordered_map<std::string, Macro *> &macros,
+                std::vector<std::string> &&macro_stack);
+
+  using Visitor<MacroExpander>::visit;
+
+  void visit(AssignVarStatement &assignment);
+  void visit(Variable &var);
+  void visit(Map &map);
+  void visit(Expression &expr);
+
+  std::optional<Block *> expand(Macro &macro, const Call &call);
+  bool check_recursive_call(const std::string &macro_name, const Call &call);
+
+private:
+  ASTContext &ast_;
+  const std::string macro_name_;
+  const std::unordered_map<std::string, Macro *> &macros_;
+
+  std::string get_new_var_ident(std::string original_ident);
+
+  // Maps of macro map/var names -> callsite map/var names
+  std::unordered_map<std::string, std::string> maps_;
+  std::unordered_map<std::string, std::string> vars_;
+  std::unordered_set<Variable *> temp_vars_;
+  const std::vector<std::string> macro_stack_;
+};
+
+class MacroCollection : public Visitor<MacroCollection> {
+public:
+  MacroCollection(ASTContext &ast, BPFtrace &b);
+  void collect_macros();
+
+  using Visitor<MacroCollection>::visit;
+
+  void visit(Expression &expr);
+
+private:
+  ASTContext &ast_;
+  BPFtrace &bpftrace_;
+  std::unordered_map<std::string, Macro *> macros_;
+};
+
+MacroExpander::MacroExpander(
+    ASTContext &ast,
+    std::string macro_name,
+    const std::unordered_map<std::string, Macro *> &macros,
+    std::vector<std::string> &&macro_stack)
+    : ast_(ast),
+      macro_name_(std::move(macro_name)),
+      macros_(macros),
+      macro_stack_(std::move(macro_stack))
+{
+  assert(!macro_stack_.empty());
+}
+
+void MacroExpander::visit(AssignVarStatement &assignment)
+{
+  auto *var = assignment.var();
+  if (temp_vars_.contains(var)) {
+    // Don't change variable names if this is a variable assignment
+    // that we created to represent the passed in expression e.g.
+    // macro add_one($x) { $x + 1 } BEGIN { $a = 1; print(add_one($a + 1)); }
+    // will become
+    // BEGIN { $a = 1; print({let $$add_one_$a = $a + 1; $$add_one_$a + 1}); }
+    return;
+  }
+  if (std::holds_alternative<VarDeclStatement *>(assignment.var_decl)) {
+    visit(std::get<VarDeclStatement *>(assignment.var_decl));
+  } else {
+    visit(std::get<Variable *>(assignment.var_decl));
+  }
+  visit(assignment.expr);
+}
+
+void MacroExpander::visit(Variable &var)
+{
+  if (auto it = vars_.find(var.ident); it != vars_.end()) {
+    var.ident = it->second;
+  } else if (!temp_vars_.contains(&var)) {
+    var.ident = get_new_var_ident(var.ident);
+  }
+}
+
+void MacroExpander::visit(Map &map)
+{
+  if (auto it = maps_.find(map.ident); it != maps_.end()) {
+    map.ident = it->second;
+  } else {
+    map.addError() << "Unhygienic access to map: " << map.ident
+                   << ". Maps must be passed into the macro as arguments.";
+  }
+}
+
+bool MacroExpander::check_recursive_call(const std::string &macro_name,
+                                         const Call &call)
+{
+  for (size_t i = 0; i < macro_stack_.size(); ++i) {
+    if (macro_stack_.at(i) == macro_name) {
+      auto &err = call.addError();
+      err << "Recursive macro call detected. Call chain: ";
+      for (; i < macro_stack_.size(); ++i) {
+        err << macro_stack_.at(i) << " > ";
+      }
+      err << macro_name;
+      return true;
+    }
+  }
+  return false;
+}
+
+void MacroExpander::visit(Expression &expr)
+{
+  auto *call = expr.as<Call>();
+  if (!call) {
+    // Recursively expand the new expression, which may again contain macros...
+    Visitor<MacroExpander>::visit(expr);
+    return;
+  }
+
+  for (auto &varg : call->vargs) {
+    visit(varg);
+  }
+
+  if (auto it = macros_.find(call->func); it != macros_.end()) {
+    Macro *macro = it->second;
+
+    if (check_recursive_call(macro->name, *call)) {
+      return;
+    }
+
+    auto next_macro_stack = macro_stack_;
+    next_macro_stack.push_back(macro->name);
+
+    auto r = MacroExpander(
+                 ast_, macro->name, macros_, std::move(next_macro_stack))
+                 .expand(*macro, *call);
+    if (r) {
+      expr.value = *r;
+    }
+  }
+}
+
+std::string MacroExpander::get_new_var_ident(std::string original_ident)
+{
+  return std::format("$${}_{}", macro_name_, original_ident);
+}
+
+std::optional<Block *> MacroExpander::expand(Macro &macro, const Call &call)
+{
+  if (macro.vargs.size() != call.vargs.size()) {
+    call.addError() << "Call to macro has wrong number arguments. Expected: "
+                    << macro.vargs.size() << " but got " << call.vargs.size();
+    return std::nullopt;
+  }
+
+  StatementList stmt_list;
+
+  for (size_t i = 0; i < call.vargs.size(); i++) {
+    if (auto *cvar = call.vargs.at(i).as<Variable>()) {
+      if (auto *mvar = macro.vargs.at(i).as<Variable>()) {
+        vars_[mvar->ident] = cvar->ident;
+      } else {
+        call.addError()
+            << "Mismatched arg to macro call. Macro expects a map for arg "
+            << macro.vargs.at(i).as<Map>()->ident << " but got a variable.";
+      }
+    } else if (auto *cmap = call.vargs.at(i).as<Map>()) {
+      if (auto *mmap = macro.vargs.at(i).as<Map>()) {
+        maps_[mmap->ident] = cmap->ident;
+      } else {
+        call.addError()
+            << "Mismatched arg to macro call. Macro expects a variable for arg "
+            << macro.vargs.at(i).as<Variable>()->ident << " but got a map.";
+      }
+    } else {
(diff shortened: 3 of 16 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -977,6 +977,131 @@ The following snippets create a map with key signature `(int64, string[16])` and
...
 ----
 
+==== Hygienic macros
+
+Hygienic macros (usually just "macros") provide a way for you to structure your script. They can be
+useful when you want to factor out code into smaller, more understandable parts. Or if you want to
+share code between probes.
+
+At a high level, macros can be thought of as semantic aware text replacement. They accept (optional)
+variable and map arguments - the body of the macro may only access maps and variables passed in through
```
amscanne, line 987: You say `maps and variables` here, but actually have an example with an expression `add_two(1)`. I think it would be good to nail that up front, e.g. maps and expressions.

man/adoc/bpftrace.adoc
```diff
@@ -977,6 +977,131 @@ The following snippets create a map with key signature `(int64, string[16])` and
...
+config = {
+    unstable_macro=1;
+}
+
+macro not_expression() {
+  $var = 1;                    // BAD: Not an expression
+}
+
+macro unhygienic_access() {
+  @x++;                        // BAD: @x not passed in
```
amscanne, line 1058: Maybe remove the `;` because it is also not an expression ;) For this one and below

man/adoc/bpftrace.adoc
```diff
@@ -977,6 +977,131 @@ The following snippets create a map with key signature `(int64, string[16])` and
...
+
+BEGIN {
+  @x = 1;
+  unhygienic_access();
+
+  wrong_parameter_type(@x);    // BAD: macro expects a scratch variable
+}
+----
+
+Macros also allow variable declarations but not if they shadow variables
```
amscanne, line 1073: Ohhhh, this one is tricky. Do we need to automatically rewrite these variables? If you are authoring a macro, you can't really be subject to constraints about what's defined in the *caller* context, e.g. this would make defining macros for the standard library nearly impossible.

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,274 @@
...
+}
+
+std::optional<Expression> MacroExpander::visit(Call &call)
+{
+  for (size_t i = 0; i < call.vargs.size(); ++i) {
+    visit(call.vargs.at(i));
+  }
+
+  if (auto it = macros_.find(call.func); it != macros_.end()) {
+    if (macro_level_ > 5) {
```
amscanne, line 80: I wouldn't impose a limit by an arbitrary constant here. Just have a separate visitor beforehand that checks for any kind of recursion, and then you will be able to expand without fear. E.g. build a graph with `std::map<std::string, std::unordered_set<std::string>>` that maps macros to calls to other macros. Then iteratively prune empty sets until it's all gone, or you have a loop, e.g.

```
std::map<std::string, std::unordered_set<std::string>> macro_calls;
auto remove_one = [&]() {
  for (const auto &[name, calls] : macro_calls) {
    if (calls.empty()) {
      for (const auto &[other_name, other_calls] : macro_calls) {
          other_calls.erase(name);
      }
      macro_calls.erase(name);
      return true;
    }
  }
  return false;
};
while (remove_one());
if (!macro_calls.empty()) {
   // print all the cycles you just found
}
```

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+private:
+  ASTContext &ast_;
+  const std::string macro_name_;
+  const std::unordered_map<std::string, Macro *> &macros_;
+  int macro_level_;
+
+  std::string make_var_literal_ident(std::string original_ident);
+
+  // Maps of macro map/var names -> callsite map/var names
```
amscanne, line 41: One thing we had talked about re: the semantics of `$x` in the macro is wether it is actually passed by l-value or not. Personally, I'd advocate for the idea that you always bind a local for the expression, e.g.

```
{
  let $__local_macro_x = $x;
  ...
}
```

In order to disallow mutability, but I know this was a topic discussed on the thread. I'm just bringing it up because. I think either way, it is something that should have some exposition here :)

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,274 @@
...
+        .specialize(*macro, call);
+    StatementList stmt_list;
+  }
+
+  return std::nullopt;
+}
+
+std::optional<Expression> MacroExpander::visit(For &f)
+{
+  for_vars_.insert(f.decl->ident);
```
amscanne, line 95: Do these need to be separate sets?

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,274 @@
...
+  }
+  if (auto it = maps_.find(map.ident); it != maps_.end()) {
+    map.ident = it->second;
+  } else {
+    map.addError() << "Unhygienic access to map";
+  }
+  return std::nullopt;
+}
+
+std::optional<Expression> MacroExpander::visit(Expression &expr)
```
amscanne, line 135: Macros are always calls that are expanded. Instead of having a visitor that exclusively returns `std::optional<Expression>`, you could just do the following:

```
class MacroExpander : Visitor<MacroExpander> {
public:
  void visit(Variable &var);
  ...
};
class MacroMatcher : Visitor<MacroMatcher> {
public:
  void visit(Expression &expr);
};

void MacroMatcher::visit(Expression &expr) {
   if (auto *call = expr.as<Call>()) {
      // Check if it's a macro...
      auto *macro = <lookup the macro>;
      // If yes...
      auto *r = clone(ast, macro->expr, expr.node().loc); // Will be a block expression.
      MacroExpander(macro).visit(macro, call.vargs);       // Expand the cloned expression.
      expr.value = r;                                                               // Replace the original call.
   }
   
   // Recursively expand the new expression, which may again contain macros...
   Visitor<MacroMatcher>::visit(expr);
}
```

I think having a structure where you split responsibilities (in addition to a visitor that checks for non-recursion of macros up front), might make it easier to understand, but this is just an idea.

src/ast/passes/parser.h
```diff
@@ -19,6 +21,8 @@ inline std::vector<Pass> AllParsePasses(
 {
   std::vector<Pass> passes;
   passes.emplace_back(CreateParsePass());
+  passes.emplace_back(CreateConfigPass());
```
amscanne, line 24: Ah yes, I have a separate change pending that does this for imports :)

src/ast/visitor.h
```diff
@@ -209,6 +209,11 @@ class Visitor {
     visitImpl(block.expr);
     return default_value();
   }
+  // A default visitor for Macro is not provided, as alone they are not
+  // valid code - they must be expanded in their callsites. Thus, Macro
+  // visitation should be specially handled by the macro expansion pass.
+  //
+  // R visit(Macro &macro) { .. }
```
amscanne, line 216: Please provide a default visitor that does nothing, and include this comment inside. Then you should have the visitor with `program.macros` uncommented out below.

If a visitor defines `visito(Macro &macro)` then it should visit them. We shouldn't expect them to re-define the `Program` level visitor. But the comment is perfectly fair that unless this is provided explicitly, the default visitor for macros will not descend into that expression.

src/parser.yy
```diff
@@ -710,12 +729,12 @@ call:
...
         |       CALL_BUILTIN  "(" ")"        { $$ = driver.ctx.make_node<ast::Call>($1, @$); }
         |       CALL_BUILTIN "(" vargs ")"   { $$ = driver.ctx.make_node<ast::Call>($1, std::move($3), @$); }
-        |       IDENT "(" ")"                { error(@1, "Unknown function: " + $1); YYERROR;  }
-        |       IDENT "(" vargs ")"          { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       BUILTIN "(" ")"              { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       BUILTIN "(" vargs ")"        { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       STACK_MODE "(" ")"           { error(@1, "Unknown function: " + $1); YYERROR;  }
         |       STACK_MODE "(" vargs ")"     { error(@1, "Unknown function: " + $1); YYERROR;  }
+        |       IDENT "(" ")"                { $$ = driver.ctx.make_node<ast::Call>($1, @$); }
+        |       IDENT "(" vargs ")"          { $$ = driver.ctx.make_node<ast::Call>($1, std::move($3), @$); }
```
amscanne, line 737: This reminds me: I had a change to strip the list of builtins and "calls" out of the lexer, which I think can now be done without issue. Did you want to do that? (Once you add macros the distinction being made between these things is now not useful, so we can decompose everything into either `Call` or `Builtin`)

amscanne:
I know it's draft, I just noticed it and was very excited :)

man/adoc/bpftrace.adoc
jordalgo: Yeah, re-writing is a good idea and I'm already doing something like that when passing literals. I could also just re-write variable assignments if we detect that the variable isn't included in the macro params, which would prevent un-hygienic errors. Let me look into doing that.

src/ast/passes/macro_expansion.cpp
jordalgo: Yeah a hard limit is no good. I have another idea to prevent recursion.

src/ast/passes/macro_expansion.cpp
jordalgo: I mentioned this on a chat to @danobi who thought we should allow variable mutability (as we allow map mutability). He was worried about restricting the functionality of these macros too much. I'm on the fence - in general this kind of mutability always feels like a footgun but I also see the argument that it restricts some of the macro functionality and we should just trust users to read the docs.

This could also be an interesting case to introduce something like `const` to the language, which would be cool for maps as well in that we know we can transform them into global static variables.

src/parser.yy
jordalgo: > I had a change to strip the list of builtins and "calls" out of the lexer, which I think can now be done without issue. Did you want to do that? 

Sure, that sounds like a fun cleanup.

src/ast/passes/macro_expansion.cpp
jordalgo: Nope. I'm removing this set anyway

src/ast/passes/macro_expansion.cpp
amscanne: Map mutability seems like it's more of a "pass by reference" situation. E.g., imagining that we supported writing memory, it is essentially writing through to the value of a pointer variable. You certainly can't change *what* the map points to. I think that the syntax of scalar maps does a bit of a dis-service here with respect to semantics (makes it seem like a global being mutated). But I digress, not intending to litigate anything here.

I think `const` makes sense for variables (and allow expressions for those cases), but gets difficult for maps (because of the above, you can pass them to functions which then mutate the values inside the maps). We probably don't want to get into the business of another language feature.

The best path is probably just to get something that you're happy with, and then do a bunch of porting internals to understand where the most friction comes from. We should assume that macros will be "unstable" for a while.

src/parser.yy
jordalgo: https://github.com/bpftrace/bpftrace/pull/4043

jordalgo:
@danobi In the latest version, I allow maps to be read in macros even if they are not passed in explicitly as arguments but assignment of those maps is not allowed. I think this cuts down on boilerplate and is still considered "hygienic" (at least from the perspective of bpftrace).

jordalgo:
@viktormalik @fbs Would love your thoughts here too

viktormalik:
> @viktormalik @fbs Would love your thoughts here too

I'm off this week so will only have time to get to this next week. But I'll try to prioritize it, then.

man/adoc/bpftrace.adoc
viktormalik: If I understand the code correctly, the re-writing is already implemented so this statement no longer holds, does it?

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+  return std::nullopt;
+}
+
+MacroCollection::MacroCollection(ASTContext &ast, BPFtrace &b)
+    : ast_(ast), bpftrace_(b)
+{
+}
+
+void MacroCollection::visit(Expression &expr)
```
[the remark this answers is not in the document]
jordalgo: I had that originally but that implementation had to keep some annoying bit of state in the MacroExpander to indicate/check for being at the top level of the AST for the other visit methods e.g. we don't want to rename any variables that aren't inside a macro. I thought this was cleaner and easier to grok.

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,326 @@
...
+        call.addError()
+            << "Mismatched arg to macro call. Macro expects a variable at arg "
+            << i << " but got a map";
+      }
+    } else {
+      if (auto *mvar = macro.vargs.at(i).as<Variable>()) {
+        auto temp_ident = get_new_var_ident(mvar->ident);
+        decl_vars_.insert(temp_ident);
+        stmt_list.emplace_back(ast_.make_node<AssignVarStatement>(
+            ast_.make_node<VarDeclStatement>(
```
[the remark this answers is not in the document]
jordalgo: Yeah I think I had that originally, I can change it back as I can't remember why I made it a Decl

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,326 @@
...
+{
+  int j = 0;
+  for (int i = static_cast<int>(macro_stack_.size()) - 1; i >= 0; --i) {
+    ++j;
+    if (macro_stack_.at(i) == macro_name) {
+      auto &err = call.addError();
+      err << "Recursive macro call detected. Call chain: ";
+      for (int k = 0; k < j; ++k) {
+        err << macro_stack_.at(k) << " > ";
+      }
```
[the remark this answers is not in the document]
jordalgo: If you look at the unit test, it makes sense. I basically want to print the proper call chain:
```
test_error(
      "macro add1($x) { $x + add3($x) } macro add2($x) { $x + add1($x) } macro "
      "add3($x) { $x + add2($x) } BEGIN { print(add3(1)); }",
      "Recursive macro call detected. Call chain: add3 > add2 > add1 > add3");
}
```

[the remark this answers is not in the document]
jordalgo: Ok, I cleaned up this logic - it was unnecessarily complicated lol

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,320 @@
...
+      macros_(macros),
+      macro_stack_(std::move(macro_stack))
+{
+  assert(!macro_stack_.empty());
+}
+
+void MacroExpander::visit(AssignVarStatement &assignment)
+{
+  auto *var = assignment.var();
+  if (var->ident.starts_with(get_new_var_prefix())) {
```
danobi, line 78: Doing a variable name check seems fragile and leaky. Can you store a std::unordered_set of pointers to variables that this pass created and check that instead?

tests/macro_expansion.cpp
```diff
@@ -0,0 +1,113 @@
...
+  auto mock_bpftrace = get_mock_bpftrace();
+  BPFtrace& bpftrace = *mock_bpftrace;
+  bpftrace.config_->unstable_macro = true;
+
+  // The input provided here is embedded into an expression.
+  ast::ASTContext ast("stdin", input);
+  std::stringstream msg;
+  msg << "\nInput:\n" << input << "\n\nOutput:\n";
+
+  // N.B. No macro or tracepoint expansion.
```
danobi, line 24: Probably best to clarify this, assuming it is accurate
```suggestion
  // N.B. No C macro or tracepoint expansion.
```

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+  if (auto it = maps_.find(map.ident); it != maps_.end()) {
+    map.ident = it->second;
+  } else {
+    map.addError() << "Unhygienic access to map: " << map.ident
+                   << ". Maps must be passed into the macro as arguments.";
+  }
+}
+
+bool MacroExpander::check_recursive_call(const std::string &macro_name,
```
danobi, line 113: Lemme think on this and see if I can break it. But I think this approach is correct

danobi (commented):
out of time for now - didn't look at the tests. but code looks reasonable to me overall

man/adoc/bpftrace.adoc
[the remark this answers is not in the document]
jordalgo: Good idea. I'll write something up.

src/ast/ast.h
```diff
@@ -1175,13 +1175,37 @@ class Import : public Node {
...
+        std::string name,
+        ExpressionList &&vargs,
+        Block *block,
+        Location &&loc)
+      : Node(ctx, std::move(loc)),
+        name(std::move(name)),
+        vargs(std::move(vargs)),
+        block(block) {};
+  explicit Macro(ASTContext &ctx, const Macro &other, const Location &loc)
+      : Node(ctx, loc + other.loc),
```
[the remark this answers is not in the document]
jordalgo: No, this doesn't really work as we want it to yet. Example:
```
bpftrace -e 'config = { unstable_macro=1; } macro add_one($x) { $x = 2; $x } BEGIN { $a = "string"; add_one($a); exit(); }'

stdin:1:52-58: ERROR: Type mismatch for $a: trying to assign value of type 'int64' when variable already contains a value of type 'string[7]'
config = { unstable_macro=1; } macro add_one($x) { $x = 2; $x } BEGIN { $a = "string"; add_one($a); exit(); }
                                                   ~~~~~~
stdin:1:52-58: ERROR: expanded from
config = { unstable_macro=1; } macro add_one($x) { $x = 2; $x } BEGIN { $a = "string"; add_one($a); exit(); }
                                                   ~~~~~~
```

I'm trying to figure out exactly what we want this to look like.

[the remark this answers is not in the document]
danobi: Broader thought: for each unstable flag we add, we should probably have an associated issue for:
1. why it's unstable
2. what's missing to get to stability
3. other notes and what not

man/adoc/bpftrace.adoc
```diff
@@ -977,6 +977,134 @@ The following snippets create a map with key signature `(int64, string[16])` and
...
+
+Some examples of invalid macro usage:
+
+----
+config = {
+    unstable_macro=1;
+}
+
+macro not_expression() {
+  $var = 1;                    // BAD: Not an expression
```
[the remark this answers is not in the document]
viktormalik: Right, ok, probably something to look into but certainly not in this PR.

src/ast/passes/macro_expansion.cpp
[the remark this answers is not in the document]
viktormalik: I see. Well, it got me confused for a while but that would probably be the case in the other implementation, too. Maybe just add a comment to explain the difference between the two?

src/ast/passes/macro_expansion.cpp
danobi: I think it's correct

src/ast/ast.h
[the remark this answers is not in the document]
jordalgo: > Broader thought: for each unstable flag we add, we should probably have an associated issue

Good idea. I'll make one for this and map declarations.

tests/runtime/macro
```diff
@@ -0,0 +1,99 @@
...
+PROG config = { unstable_macro=1; } macro print_me($x) { print(("me", $x)) } BEGIN { $a = 1; print_me($a); exit(); }
+EXPECT (me, 1)
+TIMEOUT 1
+
+NAME it can exit early
+PROG config = { unstable_macro=1; } macro early() { exit() } BEGIN { early(); print(1); }
+EXPECT_NONE 1
+TIMEOUT 1
+
+NAME it evaluates the block expression with a variable
```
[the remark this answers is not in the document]
jordalgo: Maybe this was leftover from some testing, it doesn't seem to do anything different from other tests - going to remove it.

src/ast/passes/macro_expansion.cpp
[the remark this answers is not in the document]
jordalgo: Comment added

man/adoc/bpftrace.adoc
```diff
@@ -978,6 +978,91 @@ The following snippets create a map with key signature `(int64, string[16])` and
...
+
+  $a = 42;
+  print(add_one($a)); // prints 43
+
+  @b = 3;
+  print(add_one_to_each($a, @b)); // prints 47
+
+  side_effect(5) // prints 5
+
+  print(add_two(1)); // prints 3
```
danobi, line 1032: ultra nit: align comments to same column like the `BAD: ` comments below

src/ast/passes/macro_expansion.cpp
```diff
@@ -0,0 +1,319 @@
...
+                 .expand(*macro, *call);
+    if (r) {
+      expr.value = *r;
+    }
+  }
+}
+
+std::string MacroExpander::get_new_var_ident(std::string original_ident)
+{
+  return "$$" + macro_name_ + "_" + original_ident;
```
danobi, line 163: nit: avoids unnecessary intermediate string instances:

```diff
diff --git a/src/ast/passes/macro_expansion.cpp b/src/ast/passes/macro_expansion.cpp
index c2f8186d..8b3b4894 100644
--- a/src/ast/passes/macro_expansion.cpp
+++ b/src/ast/passes/macro_expansion.cpp
@@ -1,3 +1,4 @@
+#include <format>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
@@ -160,7 +161,7 @@ void MacroExpander::visit(Expression &expr)

 std::string MacroExpander::get_new_var_ident(std::string original_ident)
 {
-  return "$$" + macro_name_ + "_" + original_ident;
+  return std::format("$${}_{}", macro_name_, original_ident);
 }

 std::optional<Block *> MacroExpander::expand(Macro &macro, const Call &call)
```

danobi (approved):

viktormalik (commented):
Looks great, just one nit.

Also, I noticed that when a macro body ends with a semicolon (i.e. is not an expression), the error message is not particularly helpful:
```
# bpftrace -e 'macro foo() { printf("hello\n"); } BEGIN { hello(); }'
stdin:1:34-35: ERROR: syntax error, unexpected }
macro foo() { printf("hello\n"); } BEGIN { hello(); }
                                 ~
```
Not sure if we can do something about it, though.

jordalgo:
> Not sure if we can do something about it, though.

Yeah, that's something that happens way down in parser and would create more shift-reduce conflicts if we allowed a trailing semi-colon in the block expr.

jordalgo:
Going to merge this unless anyone has any last concerns.

viktormalik (approved):
Exciting stuff!

jordalgo:
Appreciate the reviews @amscanne @danobi @viktormalik!

jordalgo:
Stabilize issue here: https://github.com/bpftrace/bpftrace/issues/4079

smagnani96:
@jordalgo Nice PR and feature! 💯
Just a curiosity: is there interest into supporting including macro definitions from different `.bt` files?
Trying with normal `#include` doesn't seem to work. I was able to workaround this with `sudo bpftrace  -e "$(cat main.bt macros.bt)"`, but wanted to see if there is (or will be) a native solution. Thanks in advance 😃

jordalgo:
@smagnani96 Thanks!

What you're asking is actually supported on master (though it's behind an unstable feature flag, which means you can use it but you get a warning), e.g.,
add.bt
```
macro add1($x) { $x + 1 }
```

test.bt
```
import "add.bt";

begin {
  $a = 1;
  print(add1($a));
}
```

smagnani96:
Just tried it out and works smoothly. I was not aware of that, many thanks again Jordan! 🙏🏼

PR #4043, opened 2025-04-21
Remove call and call_builtin from lexer/parser 

jordalgo:
Also remove STACK_MODE in parser which was unused

Semantic analyser already takes care of checking
if the ident of the `Call` is an expected name.

##### Checklist

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -44,16 +44,12 @@ hspace   [ \t]
 vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#$+\*])+
-builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies
-call     avg|buf|cat|cgroupid|clear|count|delete|exit|hist|join|kaddr|kptr|ksym|len|lhist|macaddr|max|min|ntop|override|print|printf|cgroup_path|reg|signal|stats|str|strerror|strftime|strncmp|strcontains|sum|system|time|uaddr|uptr|usym|zero|path|unwatch|bswap|skboutput|pton|debugf|has_key|percpu_kaddr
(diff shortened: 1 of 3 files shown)


src/lexer.l
```diff
@@ -44,17 +44,13 @@ hspace   [ \t]
 vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#$+\*])+
-builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies
-call     avg|buf|cat|cgroupid|clear|count|delete|exit|hist|join|kaddr|kptr|ksym|len|lhist|macaddr|max|min|ntop|override|print|printf|cgroup_path|reg|signal|stats|str|strerror|strftime|strncmp|strcontains|sum|system|time|uaddr|uptr|usym|zero|path|unwatch|bswap|skboutput|pton|debugf|has_key|percpu_kaddr
+builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
```
amscanne, line 47: This is now equivalent to IDENT for the purposes of parsing, will this be removed subsequently?

It's actually getting near the point where we could also remove the second parsing pass for C macro expansion, by automatically converting each defined C macro into a BPFtrace macro (IFF it is used). WDYT?

jordalgo: > This is now equivalent to IDENT for the purposes of parsing, will this be removed subsequently?

Yes, working on removing these builtins in the lexer/parser as well but running into some weird edge cases.

> It's actually getting near the point where we could also remove the second parsing pass for C macro expansion, by automatically converting each defined C macro into a BPFtrace macro (IFF it is used). WDYT?

Sounds good to me. Let me look into that after this.

viktormalik (approved):

PR #4049, opened 2025-04-22
ast: Add native import pass

amscanne:
Stacked PRs:
 * #4051
 * #4050
 * __->__#4049

--- --- ---

### ast: Add native import pass

### DIFF
--- a/src/ast/diagnostic.cpp
+++ b/src/ast/diagnostic.cpp
@@ -84,4 +84,16 @@ void Diagnostics::emit(std::ostream& out, Severity s, const Diagnostic& d) const
   }
 }
 
+void Diagnostics::add(Diagnostics&& other)
+{
+  if (diagnostics_.size() < other.diagnostics_.size()) {
+    diagnostics_.resize(other.diagnostics_.size());
+  }
+  for (size_t i = 0; i < other.diagnostics_.size(); i++) {
+    for (auto& d : other.diagnostics_[i]) {
+      diagnostics_[i].emplace_back(std::move(d));
+    }
+  }
+}
+
 } // namespace bpftrace::ast
--- a/src/ast/location.cpp
+++ b/src/ast/location.cpp
@@ -89,6 +89,9 @@ Location operator+(const Location &orig, const Location &expansion)
   if (expansion == nullptr) {
     return orig;
   }
+  if (orig == nullptr) {
+    return expansion;
+  }
   auto nlink = std::make_shared<LocationChain>(expansion->current);
   nlink->parent.emplace(LocationChain::Context(Location(orig)));
   nlink->parent->msg << "expanded from";
--- a/src/ast/passes/import_scripts.cpp
+++ b/src/ast/passes/import_scripts.cpp
@@ -1,0 +1,36 @@
+#include "ast/passes/import_scripts.h"
+#include "ast/ast.h"
+#include "ast/passes/resolve_imports.h"
+
+namespace bpftrace::ast {
+
+Pass CreateImportScriptsPass()
+{
+  return Pass::create("ImportScripts", [](ASTContext &ast, Imports &imports) {
+    // Clone all map declarations, subfunctions, etc. into the primary AST.
+    // Note that we may choose not to inline all definitions in the future, and
+    // define that namespace-based resolution is used for e.g. macro expansion,
+    // function matching, etc. But for now, we just support a trivial
+    // expansion.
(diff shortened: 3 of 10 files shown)


src/ast/location.cpp
```diff
@@ -89,6 +89,9 @@ Location operator+(const Location &orig, const Location &expansion)
   if (expansion == nullptr) {
     return orig;
   }
+  if (orig == nullptr) {
+    return expansion;
+  }
```
danobi, line 94: Wait, how can these references be null?

src/ast/diagnostic.cpp
```diff
@@ -72,4 +72,16 @@ void Diagnostics::emit(std::ostream& out, Severity s, const Diagnostic& d) const
   }
 }
 
+void Diagnostics::absorb(Diagnostics& other)
```
jordalgo, line 75: cute name but maybe something more idiomatic? `merge` or `join`?

jordalgo (commented):
Overall looks good.

src/ast/diagnostic.cpp
amscanne: `drain`? I am trying to come up with a way to express that they are being moved from `other` also. I was not happy with absorb, could also flip the relationship, e.g. `move_into`?

src/ast/location.cpp
amscanne: It's actually a null `std::shared_ptr`, so perhaps a bit confusing. Like if you did:

```
auto new_loc = foo.loc + Location()
```

In this case, `expansion == nullptr`.

I think I can drop this case actually, but I should probably just keep it and add a comment that explains it regardless.

src/ast/diagnostic.cpp
jordalgo: `move_into` isn't bad or maybe `take` ?

src/ast/passes/import_scripts.cpp
```diff
@@ -0,0 +1,31 @@
...
+      ast.root->c_definitions += other.root->c_definitions;
+      for (const auto &decl : other.root->map_decls) {
+        ast.root->map_decls.push_back(clone(ast, decl));
+      }
+      for (const auto &fn : other.root->functions) {
+        ast.root->functions.push_back(clone(ast, fn));
+      }
+      for (const auto &probe : other.root->probes) {
+        ast.root->probes.push_back(clone(ast, probe));
+      }
```
danobi, line 26: Need to clone macros now as well, right?

src/ast/location.cpp
danobi: ahh, shared_ptr, got it

src/ast/passes/import_scripts.cpp
```diff
@@ -0,0 +1,33 @@
...
+    // function matching, etc. But for now, we just support a trivial
+    // expansion.
+    for (const auto &[name, other] : imports.scripts) {
+      ast.diagnostics().add(std::move(other.diagnostics()));
+      if (other.root) {
+        ast.root->c_definitions += other.root->c_definitions;
+        for (const auto &decl : other.root->map_decls) {
+          ast.root->map_decls.push_back(clone(ast, decl));
+        }
+        for (const auto &fn : other.root->functions) {
```
jordalgo, line 22: Is this for the functions we don't even really support yet?

src/ast/passes/import_scripts.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "ast/passes/resolve_imports.h"
+
+namespace bpftrace::ast {
+
+Pass CreateImportScriptsPass()
+{
+  return Pass::create("ImportScripts", [](ASTContext &ast, Imports &imports) {
+    // Clone all map declarations, subfunctions, etc. into the primary AST.
+    // Note that we may choose not to inline all definitions in the future, and
+    // define that namespace-based resolution is used for e.g. macro expansion,
```
jordalgo, line 12: Just curious, are macros more complicated?

jordalgo (approved):
LGTM

src/ast/diagnostic.cpp
amscanne: I've made it so that you can `move` Diagnostics objects, and added an `add` that requires a moved Diagnostics. This seems like the cleanest semantics.

src/ast/passes/import_scripts.cpp
amscanne: Yep, added.

PR #4050, opened 2025-04-22
codegen: support linking imported bitcode files

amscanne:
codegen: support linking imported bitcode files

Note that this includes minor fixes to ensure that embeded files with
NUL bytes are corrected encoded as strings.


src/ast/passes/codegen_llvm.cpp
```diff
@@ -63,7 +69,7 @@ namespace bpftrace::ast {
 
 using namespace llvm;
 
-static constexpr char LLVMTargetTriple[] = "bpf-pc-linux";
+static constexpr char LLVMTargetTriple[] = "bpf";
```
jordalgo, line 72: JC, why this change?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4870,6 +4897,57 @@ Pass CreateCompilePass(
...
+                            if (fn.isDSOLocal()) {
+                              fn.removeFnAttr(Attribute::NoInline);
+                              fn.removeFnAttr(Attribute::OptimizeNone);
+                              fn.addFnAttr(Attribute::AlwaysInline);
+                              fn.addFnAttr(Attribute::NoUnwind);
+                            }
+                          }
+
+                          // Link into the original source module, consume the
+                          // new one. This returns `false` on success, and
```
jordalgo, line 4929: Not sure if I understand this comment. It returns `false` on success?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4870,6 +4897,57 @@ Pass CreateCompilePass(
...
+                          // Parse the underlying bitcode.
+                          auto buf = llvm::MemoryBuffer::getMemBuffer(
+                              StringRef(bc.data));
+                          assert(buf);
+                          auto mod = llvm::parseBitcodeFile(*buf, *ctx.context);
+                          if (!mod) {
+                            return mod.takeError();
+                          }
+
+                          // Modify to ensure everything is inlined. Note that
```
jordalgo, line 4916: Nit: I know I have the reputation of being anti-comment but might be good to also explain why the other removed/added attributes are important.

jordalgo (approved):
Few small questions but non-blocking.

src/ast/passes/codegen_llvm.cpp
amscanne: I believe that this is the new standard triple for BPF, it is no longer limited to just Linux :) In serious, it's because the compiled bitcode and codegen bits need the same triple. This was the triple used for `clang -target bpf ...` so I just made this change.

src/ast/passes/codegen_llvm.cpp
amscanne: I've clarified. Yes, it returns a `bool`, which is `false` on success. I hate it. LLVM is so inconsistent with this. That's why I wanted to do the `util::Result` class up front :)

src/ast/passes/codegen_llvm.cpp
amscanne: Added.

src/ast/passes/codegen_llvm.cpp
jordalgo: got it.

PR #4051, opened 2025-04-22
Sample call to `always_true`

amscanne:
Stacked PRs:
 * __->__#4051
 * #4050

--- --- ---

### Sample call to `always_true`

This is generated in bitcode. We need to automatically build out the
support for checking against types and the registry, but this works as
long as those checks are hand-coded (and correct) in the semantic
analyser.


amscanne:
Replaced by the stack in #4340.

Issue #3711, opened 2025-01-15
Clean up all the spew in unit tests


danobi:
### Is your feature request related to a problem? Please describe.
When you run `bpftrace_test`, if the tests pass, you should get no extraneous output other than the gtest decorations. 

### Describe the solution you'd like
Go to each place where there is spew and figure out how to silence it.

### Describe alternative solutions or features you've considered

n/a

Phoenix500526:
I ran the `bpftrace_test` and pasted its output here:
```bash
./bpftrace_test 
[==========] Running 548 tests from 35 test suites.
[----------] Global test environment set-up.
...
[ RUN      ] bpftrace.empty_attachpoints
stdin:1:1-1: ERROR: No attach points for probe
{}

[       OK ] bpftrace.empty_attachpoints (116 ms)
....
[----------] 1 test from bpftrace_bad_btf
[ RUN      ] bpftrace_bad_btf.parse_invalid_btf
WARNING: BTF: failed to parse BTF from /tmp/btf_datahuLEss
WARNING: BTF: failed to parse BTF from /tmp/btf_datahuLEss
[       OK ] bpftrace_bad_btf.parse_invalid_btf (6 ms)
[----------] 1 test from bpftrace_bad_btf (7 ms total)
...
[ RUN      ] clang_parser.parse_fail
definitions.h:2:28: error: field has incomplete type 'struct b'
definitions.h:2:26: note: forward declaration of 'struct b'
[       OK ] clang_parser.parse_fail (466 ms)
[ RUN      ] clang_parser.btf_unresolved_typedef
[       OK ] clang_parser.btf_unresolved_typedef (894 ms)
[ RUN      ] clang_parser.struct_typedef
[       OK ] clang_parser.struct_typedef (622 ms)
[ RUN      ] clang_parser.struct_qualifiers
[       OK ] clang_parser.struct_qualifiers (499 ms)
[ RUN      ] clang_parser.redefined_types
definitions.h:3:8: error: redefinition of 'a'
definitions.h:2:8: note: previous definition is here
WARNING: Cannot take type definitions from BTF since there is a redefinition conflict with user-defined types.
definitions.h:3:8: error: redefinition of 'a'
definitions.h:2:8: note: previous definition is here
WARNING: Cannot take type definitions from BTF since there is a redefinition conflict with user-defined types.
[       OK ] clang_parser.redefined_types (769 ms)

...
----------] 3 tests from clang_parser_btf
[...]

PR #4056, opened 2025-04-23
Revert "enhancement: suppress extraneous output in bpftrace_test when tests pass"

jordalgo:
This reverts commit a30d61b56368ef33092fa261c168424168b1bcd3.

This is causing 2 issues:
1) Our test output supression is swallowing LOG(BUG) because this leads to an abort which the test listener is not catching

2) LOG(ERROR) no longer leads to a program termination so any places where this is called are also getting swallowed by output suppression because the tests don't fail

1 - This is probably address-able by adding a signal handler to OutputSuppressorListener.

2 - This is a larger problem with the codebase and needs to be addressed.

##### Checklist

### DIFF
--- a/tests/main.cpp
+++ b/tests/main.cpp
@@ -1,11 +1,7 @@
 #include "gtest/gtest.h"
-#include <cstdlib>
-#include <string>
-#include <sys/mman.h>
-#include <unistd.h>
 
 class ThrowListener : public testing::EmptyTestEventListener {
-  void OnTestPartResult(const testing::TestPartResult& result) override
+  void OnTestPartResult(const testing::TestPartResult &result) override
   {
     if (result.type() == testing::TestPartResult::kFatalFailure) {
       throw testing::AssertionException(result);
@@ -13,76 +9,9 @@ class ThrowListener : public testing::EmptyTestEventListener {
   }
 };
 
-class OutputSuppressorListener : public testing::EmptyTestEventListener {
-public:
-  void OnTestStart(const testing::TestInfo& /*test_info*/) override
-  {
-    // stash original stderr
-    original_stderr_ = dup(STDERR_FILENO);
-    if (original_stderr_ == -1) {
-      return;
-    }
-
-    // create a temp file
-    constexpr unsigned int NO_FLAGS = 0;
-    memfd_ = memfd_create("stderr_capture", NO_FLAGS);
-    if (memfd_ == -1) {
-      close(original_stderr_);
-      original_stderr_ = -1;
-      return;
-    }
-
-    // redirect stderr to temp file
-    dup2(memfd_, STDERR_FILENO);
-  }
-
-  void OnTestEnd(const testing::TestInfo& test_info) override
-  {
-    if (original_stderr_ != -1) {
-      // restore output if test failed
-      if (test_info.result()->Failed()) {
-        restore_output();
-      }
-
-      if (original_stderr_ != -1) {
-        dup2(original_stderr_, STDERR_FILENO);
-        close(original_stderr_);
-        original_stderr_ = -1;
-      }
-    }
-  }
-
-private:
-  void restore_output()
-  {
-    if (memfd_ && original_stderr_ != -1) {
-      // temporarily switch back to original stderr
-      int old = dup(STDERR_FILENO);
-      dup2(original_stderr_, STDERR_FILENO);
-
-      // reset file pointer and output content
-      lseek(memfd_, 0, SEEK_SET);
-      char buffer[1024];
-      ssize_t n;
-      while ((n = read(memfd_, buffer, sizeof(buffer))) > 0) {
-        std::cerr << buffer;
-      }
-
-      // restore to temp file
-      dup2(old, STDERR_FILENO);
-      close(old);
-    }
-  }
-
-  int memfd_ = -1;
-  int original_stderr_ = -1;
-};
-
-int main(int argc, char* argv[])
+int main(int argc, char *argv[])
 {
   ::testing::InitGoogleTest(&argc, argv);
   ::testing::UnitTest::GetInstance()->listeners().Append(new ThrowListener);
-  ::testing::UnitTest::GetInstance()->listeners().Append(
-      new OutputSuppressorListener);
   return RUN_ALL_TESTS();
 }


jordalgo:
cc @danobi @Phoenix500526

Phoenix500526:
Can we register a signal handler in the `OutputSuppressorListener` to temporarily unblock the output suppression, then safely re-raise the same signal after ensuring no recursive triggering?   @jordalgo

jordalgo:
I'll let @danobi weigh in but TBH, I don't think what's required for this to work well and not hide errors is worth the trade off of not seeing some bit of log spew. Even if we add a signal handler, we still don't solve issue #2 which is more systemic.

amscanne:
It would be good to enumerate the cases of log spew that this was fixing. I suspect that we could just wrap individual cases where they is spew that we don't control (e.g. some underlying library writing to stderr) in an RAII variant of this class (or merge this into [StderrSilencer](https://github.com/bpftrace/bpftrace/blob/master/src/util/io.h#L22)).

danobi:
Hmm, I'm still a fan of hooking into the test harness. I think quite a few other test frameworks (albeit in other languages) do this. Catching signals should be reasonably robust - this is what folly::installFatalSignalHandler() does (https://github.com/facebook/folly/blob/1257705272de61783fbdfb6dc0f701f699322243/folly/debugging/symbolizer/SignalHandler.cpp#L507). 

Granted, it's probably overkill to actually implement all that. Maybe it's better to just go and fix all the tests that spew? I think it's good to fix the spew - on principle it's hard to converge to quality if you selectively let ugly things through.

jordalgo:
> Maybe it's better to just go and fix all the tests that spew?

I'm a fan of that. In truth the spew doesn't bother me all that much but I do feel strongly about not accidentally suppressing important things that lead to test failure.

Phoenix500526:
> 2\. LOG(ERROR) no longer leads to a program termination so any places where this is called are also getting swallowed by output suppression because the tests don't fail

Shouldn't it be LOG(BUG)?

Phoenix500526:
IMO, the commit a30d61b is not the culprit of the issue you mentioned above. I did some experiments. 

I prepared two experimental environments:

- master, whose git HEAD is the latest commit. (ci: Fold output for successful test suites - b0d177)
- test_env, whose git HEAD is the commit 071f29b9 (ast: add basic `clone` plumbing)

And then I added some test cases like :
```C++
TEST(LogDeath, bug_test) {
  EXPECT_DEATH(LOG(BUG) << "I’m gonna die", ".*");
}

TEST(LogDeath, error_test) {
  EXPECT_DEATH(LOG(ERROR) << "I’m gonna die", ".*");
}
```


The results on the master branch：
```bash
bt --gtest_filter=LogDeath.*
Note: Google Test filter = LogDeath.*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from LogDeath
[ RUN      ] LogDeath.bug_test
[       OK ] LogDeath.bug_test (79 ms)
[ RUN      ] LogDeath.error_test
/bpftrace/tests/log.cpp:129: Failure
Death test: bpftrace::LogStream("/bpftrace/tests/log.cpp", 129, bpftrace::LogType::ERROR) << "I’m gonna die"
    Result: failed to die.
 Error msg:
[  DEATH   ] ERROR: I’m gonna die
[  DEATH   ] 

[  FAILED  ] LogDeath.error_test (3 ms)
[----------] 2 tests from LogDeath (82 ms total)

[==========] 2 tests from 1 test suite ran. (82 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] LogDeath.error_test

 1 FAILED TEST
```

The results on the test_env branch：
```bash
bt --gtest_filter=LogDeath.*
Note: Google Test filter = LogDeath.*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from LogDeath
[ RUN      ] LogDeath.bug_test
[       OK ] LogDeath.bug_test (76 ms)
[ RUN      ] LogDeath.error_test
/bpftrace/tests/log.cpp:129: Failure
Death test: bpftrace::LogStream("/bpftrace/tests/log.cpp", 129, bpftrace::LogType::ERROR) << "I’m gonna die"
    Result: failed to die.
 Error msg:
[  DEATH   ] ERROR: I’m gonna die
[  DEATH   ] 

[  FAILED  ] LogDeath.error_test (2 ms)
[----------] 2 tests from LogDeath (79 ms total)

[==========] 2 tests from 1 test suite ran. (79 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] LogDeath.error_test

 1 FAILED TEST
```

These results show that the SIGABRT won’t be ignored even if we introduce the `OutputSuppressorListener`


I also added some other test cases: 
```C++
TEST(LogDeath, bug_output) {
  LOG(BUG) << "I’m gonna die";
}

TEST(LogDeath, error_output) {
  LOG(ERROR) << "I’m gonna die";
}
```

The results on the master branch：
```bash
$ bt --gtest_filter=LogDeath.*
Note: Google Test filter = LogDeath.*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from LogDeath
[ RUN      ] LogDeath.bug_output

$ 
```

The results on the test_env branch：
```bash
$ bt --gtest_filter=LogDeath.*
Note: Google Test filter = LogDeath.*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from LogDeath
[ RUN      ] LogDeath.bug_output
BUG: [/bpftrace/tests/log.cpp:125] I’m gonna die

$
```

Their behaviors are almost the same. The only difference is that the former captured all log output. None of them think their tests are failed.  So their behaviors are correct, and it’s also what we expected.

jordalgo:
@Phoenix500526 

> Shouldn't it be LOG(BUG)?

No, `LOG(BUG)` aborts the program and indicates that there is an error that the user can't do anything about (it's almost equivalent to an assert) - it's something for bpftrace developers. `LOG(ERROR)` is an indication the user did something wrong or there is something wrong with their environment. More info here: https://github.com/bpftrace/bpftrace/blob/master/docs/coding_guidelines.md#log-messages

> Their behaviors are almost the same. The only difference is that the former captured all log output. 

This is the the big difference. One is swallowing the `LOG(BUG)` output, which is needed for debugging. Additionally, in my own A/B tests `LOG(ERROR)` in the codebase (which didn't eventually lead to failed tests or program termination) were also being swallowed. As mentioned this is a larger problem in the code base but the commit this PR is reverting just made it worse.

Phoenix500526:
@jordalgo 

> [@danobi](https://github.com/danobi) I think I'd want a way to turn this on/off (maybe env var) because if I'm running these locally I might want output from out/err.


OK, I think I get your points. 
1. The first point essentially states: If a user invokes `LOG(BUG)` during testing (even if the test doesn't ultimately fail), the log message should still be printed to stderr. We can add a signal handler in `OutputSuppressorListener` to handle the `SIGABRT`.
2. The second point indicates: When you attempted to use `LOG(ERROR)` for debugging purposes, the `OutputSuppressorListener` intercepted those messages (instead of displaying them on stderr), which complicated your debugging process.  Maybe we can add an env to turn on/off this `OutputSuppressorListener` as you commented in the issue #3711 




Do I get your points right? Please correct any misunderstandings.

jordalgo:
@Phoenix500526 Yes, you understand both points. But my ultimate larger point is that the code and considerations required to just suppress some log spew is not worth it e.g. I don't want to have to remember this flag and re-run tests if something in CI fails and swallows something important.

Phoenix500526:
> I don't want to have to remember this flag and re-run tests if something in CI fails and swallows something important.

@jordalgo 

We can disable the OutputSuppressorListener by default, enabling its functionality only when the specific environment variable is set. Users shouldn’t pay the cost for unused features, so you don’t need to keep track of the flag. BTW, if something in CI fails, all the logs will be restored to stderr. It won’t swallow anything. 

Can you provide a concrete example where, with `OutputSuppressorListener` manually enabled, a test fails but critical logs are swallowed?

Phoenix500526:
Any updates? @jordalgo

danobi (approved):

PR #4057, opened 2025-04-23
tests/README: document comments in runtime tests

ttreyer:
Document comments in runtime tests.

##### Checklist

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -139,6 +139,14 @@ separate minimal programs to test tracing functionality, and argument passing
 hasn't been required. If test programs need arguments, a more sophisticated
 approach will be necessary.
 
+### Comments
+
+Lines starting with a `#` character are treated as comments and are ignored by
+the test framework. This allows you to add explanatory notes or documentation
+within your test files without affecting test execution.
+
+Trailing comments are are not supported.
+
 ### Runtime variables
 
 Runtime variables are placeholders that the runtime test engine will fill out


tests/runtime/engine/parser.py
```diff
@@ -181,9 +187,9 @@ def __read_test_struct(test, test_suite):
             elif item_name == 'CLEANUP':
                 cleanup = line
             elif item_name == 'MIN_KERNEL':
-                kernel_min = line
+                kernel_min = TestParser.strip_comments(line)
             elif item_name == 'MAX_KERNEL':
-                kernel_max = line
+                kernel_max = TestParser.strip_comments(line)
```
danobi, line 192: How about doing on `line` before the giant if/elif block? Seems useful to do generically

ttreyer: It cannot be applied to every item; the test [other.positional arg count](https://github.com/ttreyer/bpftrace/blob/ttreyer/rt-trim-comments/tests/runtime/other#L162-L164) has a `#` in the BpfScript. There's also a few tests in `basic` that use `#include` and would get their script stripped.

danobi: Hmm, fair. But it's still weird to make MAX_KERNEL and MIN_KERNEL special w.r.t. comments. How about making PROG and RUN special instead?

danobi: Or just move the trailing comments to their own line?

ttreyer: Something like [this](https://github.com/bpftrace/bpftrace/pull/4057/commits/55b77d1f7f0481010d8628b4289a51a23e5df8df)?
I don't think that comment should be trimmed for any `EXPECT_…`, their values are arbitrary text/regex.
Same for `BEFORE`, `AFTER`, `SETUP` and `CLEANUP`, which contain shell commands.
How about `ENV`? The value might contain '#'.

danobi: So this is getting a bit complex. Let's back up - what is this needed for? Why doesn't putting comments on their own line suffice? I think having consistency is good in general.

When I search for trailing comments I don't see any existing ones. Am I missing something?

```
$ rg '.+#'
other
163:RUN {{BPFTRACE}} -e 'BEGIN { printf("got %d args: %s %d\n", $#, str($1), $2); exit();}' "one" 2

map
23:# know exactly which elements will be present. See #3992 for more information.

regression
1:# https://github.com/bpftrace/bpftrace/pull/566#issuecomment-487295113
114:# #3603). #3629 introduces the use of intrinsics that should ensure that this

builtin
80:# (https://github.com/iovisor/bcc/pull/4319#issuecomment-1321731687)

basic
221:RUN {{BPFTRACE}} -e "$(echo "#include <sys/xattr.h>"; echo "BEGIN { exit(); }")" 2>&1
227:RUN {{BPFTRACE}} -e "$(echo "#include <lol/no.h>"; echo "BEGIN { exit(); }")" 2>&1
233:RUN {{BPFTRACE}} -e "$(echo '#define _UNDERSCORE 314'; echo 'BEGIN { printf("%d\n", _UNDERSCORE); exit(); }')"

usdt
104:#  - Fix https://github.com/bpftrace/bpftrace/issues/565#issuecomment-496731112
```

ttreyer: Oh no 🙃 I'm the one who added trailing comments in my changes...
I'll move the comments to their own line and document commenting in the tests' README.

tests/README.md
```diff
@@ -54,6 +54,14 @@ Runtime tests will call the bpftrace executable. These are located in `tests/run
 Runtime tests are grouped into "suites". A suite is usually a single file. The
 name of the file is the name of the suite.
 
+### Comments
```
danobi, line 57: Put this after "Runtime test directives" section? Better to order this from most to least important

danobi (approved):

PR #4065, opened 2025-04-24
Enable some unstable features by default

jordalgo:
Enable some unstable features by default

Instead of having users explicitly opt-in to
most unstable features, enable them by default.
Obviously unstable features that are
half-baked or more risky should be disabled
by default but this can be on a case-by-case
basis. In general though we want to reduce
friction for users to use new, cool things.

This changes the config value type for unstable
features to an enum: enable, warn, and error.
For macros and map declarations it will be warn,
which means users get a warning they are using
an unstable feature. They can suppress this warning
by setting this config to 'enable'.

The import feature will stay as 'error'.

##### Checklist

### DIFF
--- a/src/ast/passes/unstable_feature.cpp
+++ b/src/ast/passes/unstable_feature.cpp
@@ -1,0 +1,96 @@
+#include <cstring>
+#include <format>
+#include <string>
+
+#include "ast/ast.h"
+#include "ast/passes/unstable_feature.h"
+#include "ast/visitor.h"
+#include "bpftrace.h"
+#include "config.h"
+
+namespace bpftrace::ast {
+
+namespace {
+
+std::string get_warning(const std::string &feature)
+{
+  return std::format("Script is using an unstable feature. To prevent this "
+                     "warning you must explicitly enable the unstable "
+                     "feature in the config e.g. {}=enable",
+                     feature);
+}
+
+std::string get_error(std::string &&feature)
+{
+  return std::format(
+      "Feature not enabled by default. To enable "
+      "this unstable feature, set the config flag to enable. {}=enable",
+      feature);
+}
+
+class UnstableFeature : public Visitor<UnstableFeature> {
+public:
+  explicit UnstableFeature(BPFtrace &bpftrace) : bpftrace_(bpftrace) {};
+
+  using Visitor<UnstableFeature>::visit;
+  void visit(MapDeclStatement &decl);
+  void visit(Import &imp);
+  void visit(Macro &macro);
+
+private:
+  BPFtrace &bpftrace_;
+  // This set is so we don't warn multiple times for the same feature.
+  std::unordered_set<std::string> warned_features;
+};
+
+} // namespace
+
+void UnstableFeature::visit(MapDeclStatement &decl)
+{
+  if (bpftrace_.config_->unstable_map_decl == ConfigUnstable::error) {
+    decl.addError() << get_error(UNSTABLE_MAP_DECL);
+    return;
+  }
+  if (bpftrace_.config_->unstable_map_decl == ConfigUnstable::warn &&
+      !warned_features.contains(UNSTABLE_MAP_DECL)) {
+    decl.addWarning() << get_warning(UNSTABLE_MAP_DECL);
+    warned_features.insert(UNSTABLE_MAP_DECL);
+  }
+}
+
+void UnstableFeature::visit(Import &imp)
+{
+  if (bpftrace_.config_->unstable_import == ConfigUnstable::error) {
+    imp.addError() << get_error(UNSTABLE_IMPORT);
+    return;
+  }
+
+  if (bpftrace_.config_->unstable_import == ConfigUnstable::warn &&
+      !warned_features.contains(UNSTABLE_IMPORT)) {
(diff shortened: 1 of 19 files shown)


jordalgo:
This is an idea that @amscanne and I had to make it easier to use unstable features but still make users aware of their instability.

src/ast/passes/unstable_feature.cpp
```diff
@@ -0,0 +1,83 @@
...
+void UnstableFeature::visit(MapDeclStatement &decl)
+{
+  if (!bpftrace_.config_->unstable_map_decl) {
+    decl.addError() << "Map declarations are not enabled by default. To enable "
+                       "this unstable feature, set this config flag to 1 "
+                       "e.g. unstable_map_decl=1";
+    return;
+  }
+  if (!enabled_unstable_features.contains("unstable_map_decl")) {
+    decl.addWarning() << "Script is using an unstable feature. To prevent this "
```
amscanne, line 53: Did you want to just make the unstable_map_decl a new type to capture the TriState? (You could parse as `true`, `false`, `warn` and assign sensible mappings to 0 and 1 that match the current defaults. We'd effectively be changing the default from `false` to `warn`).

amscanne: If you do that you also might not need a dedicated pass. Individual passes would just need to check that config flag and see if it was e.g. UnstableFlag::Disable, UnstableFlag::Enable, UnstableFlag::Warn.

jordalgo: What is the use-case for setting this config flag to `warn`? Couldn't users just not include the config flag at all and that's the default behavior?

jordalgo:
@viktormalik Any thoughts on this one? Sounds like maybe it's ok to proceed.

jordalgo (commented):
@danobi @amscanne - Gentle ping

src/ast/passes/unstable_feature.cpp
```diff
@@ -0,0 +1,107 @@
+#include <cstring>
```
amscanne, line 1: Optional nit: maybe we want to roll this and the deprecated pass into one "feedback" pass. Both are just providing hints to the user about old or new functionality, so maybe they are logically the same.

I don't actually feel there's a particularly strong reason one way or the other, it's just a logical alternative that I thought I'd mention. Totally your preference.

src/ast/passes/unstable_feature.cpp
```diff
@@ -0,0 +1,107 @@
...
+  BPFtrace &bpftrace_;
+  std::unordered_set<std::string> enabled_unstable_features;
+};
+
+} // namespace
+
+void UnstableFeature::visit(AssignConfigVarStatement &assignment)
+{
+  // If set explicitly, we don't need to warn of use
+  if (bpftrace_.config_->is_unstable(assignment.var)) {
```
amscanne, line 50: Can we just explicitly make this configuration variable a tri-state? (false/warn/true) And then just refer to the config here in this analyser, rather than introducing a new separate map that basically tracks whether it is explicitly set or not.

src/config.h
```diff
@@ -32,9 +36,9 @@ class Config {
   bool cpp_demangle = true;
   bool lazy_symbolication = true;
   bool print_maps_on_exit = true;
-  bool unstable_macro = false;
-  bool unstable_map_decl = false;
-  bool unstable_import = false;
+  bool unstable_macro = true;
```
amscanne, line 39: See comment elsewhere. I'm into the change in spirit, I just think you can introduce a basic enum type and parser here and make this behavior explicit rather than relying on an implicit (it is actually set in the config) value.

e.g.
```
UnstableFeature unstable_macro = UnstableFeature::warn;
```

src/ast/passes/unstable_feature.cpp
jordalgo: I'm not opposed but let me look at doing that in a follow-up PR.

src/ast/passes/unstable_feature.cpp
jordalgo: I made it a tri-state but I still need the map to avoid issuing the same warning multiple times.

src/config.h
jordalgo: Done.

jordalgo:
Another thought, even though this enables all our current unstable features by default, this doesn't have to be the case for all new ones. A particularly risky feature could be disabled by default if we're worried about unexpected behavior or potential bugs.

amscanne (approved):
LGTM

src/ast/passes/unstable_feature.cpp
amscanne: Yeah SG. We could also mutate the config value on first warning, but is undesirable for other reasons. This seems reasonable to me.

amscanne:
> Another thought, even though this enables all our current unstable features by default, this doesn't have to be the case for all new ones. A particularly risky feature could be disabled by default if we're worried about unexpected behavior or potential bugs.

Yeah, did you want to update the default for just macros and imports for now? I don't think imports are ready for this yet, but it's nice to change it to `warn` when it's done.

jordalgo:
@amscanne Yeah, let me change it to error for imports.

PR #4066, opened 2025-04-25
fix: colorized LogStreamBug output

Phoenix500526:
This pr enable colorized feature for LogStreamBug.

##### Checklist

### DIFF
--- a/tests/log.cpp
+++ b/tests/log.cpp
@@ -120,7 +120,7 @@ TEST(Log, disable_log_type)
   ss.str({});
 }
 
-TEST(LogBugStream, log_bug_should_be_aborted)
+TEST(LogStreamBug, log_stream_bug_should_be_aborted)
 {
   const std::string content = "I'm gonna die";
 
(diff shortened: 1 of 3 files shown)


tests/log.cpp
```diff
@@ -132,4 +132,21 @@ TEST(LogBugStream, log_bug_should_be_aborted)
...
+    const std::string default_color = "\033\\[0m";
+    const std::string filename = __FILE__;
+    return bug_color + "BUG: \\[" + filename + ":" + std::to_string(line) +
+           "\\] " + content + default_color;
+  };
+  EXPECT_DEATH(LOG(BUG) << content, output(__LINE__));
+  Log::get().set_colorize(false);
+}
+
+} // namespace bpftrace::test::log
```
jordalgo, line 152: Nit: missing empty line at the end of the file

tests/log.cpp
```diff
@@ -132,4 +132,21 @@ TEST(LogBugStream, log_bug_should_be_aborted)
   EXPECT_DEATH(LOG(BUG) << content, output(__LINE__));
 }
 
-} // namespace bpftrace::test::log
+TEST(LogStreamBug, log_stream_bug_colorized)
```
jordalgo, line 135: I dunno. This extra test (as well as any testing for the colorization of logs) feels like overkill.

tests/log.cpp
jordalgo: Nice. Thanks!

tests/log.cpp
jordalgo: I won't block but the reason I think it's overkill is because log colorization is not business critical functionality for bpftrace. IMHO it's nice but not important enough to have and maintain unit tests for. Other people might feel differently. I just think there are always maintenance concerns when adding new code (unit tests are no exception).

Phoenix500526:
Hi, @danobi. Could you give this pr a look? Thanks 😄

jordalgo (approved):
As mentioned, I'm not going to block because I think a unit test is a little overkill.

PR #4070, opened 2025-04-26
doc: update join function's description

Phoenix500526:
This pr modifies the description of the join function from `Print the array` to `Join all the string array into one string` in `bpftrace.adoc`

##### Checklist


man/adoc/bpftrace.adoc
```diff
@@ -2029,7 +2029,7 @@ Tracing block I/O sizes > 0 bytes
 | Async
 
 | <<functions-join, `join`>>
-| Print the array
+| Join all the string array into one string
```
jordalgo, line 2032: So from the detailed description it looks like both of these are correct:
> `join` joins all the string array arr with sep as separator into one string. This string will be printed to stdout directly"

So maybe a better short description is:
`Combine an array of char* into one string and print it.`

Also the long description could also be updated to:

`join` joins a `char *` arr with `sep` as separator into one string. This string will be printed to stdout directly

Phoenix500526: Done

jordalgo (approved):
LGTM. Thanks for the fix!

PR #4071, opened 2025-04-27
refactor: refactor the join function logic

Phoenix500526:
I added the `AsyncAction::Join` struct with its asLLVMType method and updated the `createJoinCall` function to support the new join logic.

I run the bpftrace manually, and tested the refactored result by running the shell command `ls -l /tmp`. The tested results are pasted below:

Before refactor:
```bash
# original bpftrace
$ bpftrace -e 'tracepoint:syscalls:sys_enter_execve { join(args.argv) }'
Attaching 1 probe...
# omit some output ...
ls --color=auto -l /tmp
# omit some output ...
...
```

After refactor:
```bash
# refactored bpftrace
$ ./build/src/bpftrace -e 'tracepoint:syscalls:sys_enter_execve { join(args.argv) }'
Attaching 1 probe...
# omit some output ...
ls --color=auto -l /tmp
# omit some output ...
```

It proves that, from the user’s point of view, the refactored join function behaves the same way as the original join function. 

So far, the refactored implementation currently doesn't satisfy the existing test cases (`codegen.call_join` and `codegen.call_join_with_debug`) because I haven’t updated the `tests/codegen/llvm/call_join.ll` and `tests/codegen/llvm/call_join_with_debug.ll` yet. 

Therefore, I have two questions:
1. How to update these two files?
2. If we update these two files, we may not ensure that the refactored join is the same as its previous version, because their test cases become different. Do we have a better way to test it?

@jordalgo @danobi 

##### Checklist


jordalgo:
@Phoenix500526 Looks like you figured out how to update the codegen tests 😄 
To answer your second question, the codegen tests are mostly there to prevent unintentional codegen changes and allow reviewers to see how codegen changes when the tests are updated as part of a PR.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4003,6 +3948,71 @@ void CodegenLLVM::createPrintMapCall(Call &call)
...
+
+    b_.CreateProbeReadStr(ctx_,
+                          str_ptr,
+                          bpftrace_.join_argsize_,
+                          b_.CreateLoad(b_.getInt64Ty(), arr),
+                          addrspace,
+                          call.loc);
+  }
+
+  size_t total_size = 8 + 8 + content_size; // action_id + join_id + content
```
jordalgo, line 4009: Can you get these two `8` sizes from the first two elements of the LLVM type?

jordalgo (commented):
One small nit but otherwise LGTM

src/ast/passes/codegen_llvm.cpp
Phoenix500526: Done

jordalgo:
@Phoenix500526 Do you think it's worth updating the Join BPF map to be the same type as this new struct (and I guess make it a per_cpu_hash?
https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/codegen_llvm.cpp#L4140

Phoenix500526:
> @Phoenix500526 Do you think it's worth updating the Join BPF map to be the same type as this new struct 

Are you suggesting replacing the hardcoded offset `8 + 8` with `offsetof(AsyncEvent::Join, content)`?

> I guess make it a per_cpu_hash? https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/codegen_llvm.cpp#L4140

The current BPF map for join operations uses `BPF_MAP_TYPE_PERCPU_ARRAY`. What are the potential advantages of switching to `BPF_MAP_TYPE_PERCPU_HASH`?

jordalgo:
> The current BPF map for join operations uses BPF_MAP_TYPE_PERCPU_ARRAY. What are the potential advantages of switching to BPF_MAP_TYPE_PERCPU_HASH?

I don't think it makes much of a difference in terms of performance but you could unify the types here in that the map value type would be your newly added join struct instead of this byte array.

Phoenix500526:
> > The current BPF map for join operations uses BPF_MAP_TYPE_PERCPU_ARRAY. What are the potential advantages of switching to BPF_MAP_TYPE_PERCPU_HASH?
> 
> I don't think it makes much of a difference in terms of performance but you could unify the types here in that the map value type would be your newly added join struct instead of this byte array.

I think it would not be worth the effort. Although we can set AsyncEvent::Join as a value type in a BPF_MAP_TYPE_PERCPU_HASH, we still need to generate a BitCast instruction to reinterpret the void pointer returned by `CreateGetJoinMap` as an AsyncEvent::Join pointer.

jordalgo:
> I think it would not be worth the effort. 

Yeah, upon closer inspection, it looks like more of a pain than I thought it was.

jordalgo (approved):

danobi:
Nice work!

PR #4082, opened 2025-05-01
cmake: simplify `BuildBPF`, eliminate race

amscanne:
Amazingly, `cmake` manages to generate Makefiles that execute the same
rule multiple times in parallel if the same `OUTPUT` is used in
multiple `DEPENDS` stanzas. See [1] for more information.
    
[1] https://discourse.cmake.org/t/how-to-avoid-parallel-build-race-conditions/727
    
This change simplifies some of the structure in `BuildBPF`, using
relative paths in the local rules, which should trigger the heuristics
internally:
    
```
If any dependency is an OUTPUT of another custom command in the same
directory (CMakeLists.txt file), CMake automatically brings the other
custom command into the target in which this command is built.
```

This does not fix this issue, but should make it easier to see.
    
For the actual fix, we group the cases where the same output is used
for multiple rules into a target. This should prevent the concurrent
execution of those rules.

### DIFF
--- a/cmake/Embed.cmake
+++ b/cmake/Embed.cmake
@@ -7,7 +7,7 @@ function(embed NAME SOURCE)
     ARG
     ""
     "OUTPUT;HEX;VAR"
-    ""
+    "DEPENDS"
     ${ARGN}
   )
   if (NOT DEFINED ARG_HEX)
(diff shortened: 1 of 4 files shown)


jordalgo:
@amscanne What is the actual race/bug being fixed here - just the running in parallel of the same build?

cmake/Embed.cmake
```diff
@@ -36,13 +36,14 @@ function(embed NAME SOURCE)
       -DTEMPLATE=${CMAKE_SOURCE_DIR}/cmake/Embed.tmpl
       -P ${CMAKE_SOURCE_DIR}/cmake/EmbedRun.cmake
     DEPENDS
-      ${CMAKE_CURRENT_BINARY_DIR}/${ARG_HEX}
+      ${ARG_HEX}
       ${CMAKE_SOURCE_DIR}/cmake/EmbedRun.cmake
       ${CMAKE_SOURCE_DIR}/cmake/Embed.tmpl
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
```
danobi, line 42: Is this right? Doesn't it mean ${ARG_OUTPUT} goes into src/ rather than build/?

danobi (commented):
ah, you have re-discovered the same fix in https://github.com/bpftrace/bpftrace/commit/6ce6693a61e7a6a9d1c8a97622a0ddef6674d5d0

I suspected this was it... but I guess I failed to recreate the incantation

cmake/Embed.cmake
amscanne: `-DOUPTUT=...` is still specified in absolute terms, so the working directory doesn't actually matter here. I thought I had stripped these changes though, so I'll drop this.

amscanne:
> ah, you have re-discovered the same fix in [6ce6693](https://github.com/bpftrace/bpftrace/commit/6ce6693a61e7a6a9d1c8a97622a0ddef6674d5d0)
> 
> I suspected this was it... but I guess I failed to recreate the incantation

Indeed, WTF cmake

I vaguely recall having dealt with this before, and the pattern of 'just have a target for everything' was familiar. The solution here and actually slightly different, I am just defining an explicit target for every generated file, rather than having those custom rules rely on both a base target and the base file.

I'm gonna fully digest the details and make sure this works for incremental builds as well

danobi (approved):

danobi:
Since this is more correct than what's on master (and currently failing on every commit), I will land this now. If you find issues, we can fix forward or revert.

danobi:
@jordalgo depending on a file OUTPUT of a add_custom_command() can cause all higher level targets to "inline" the command. So the command is being run multiple times in parallel. Which can confuse pahole if the object file is being overwritten in the middle of a pahole encoding

PR #4084, opened 2025-05-01
Use `blaze_sym::offset` attribute for reporting offsets

danielocfb:
Use the blaze_sym::offset attribute for reporting function offsets instead of attempting to calculate the offset by subtracting the symbol address from the input address. The latter will only work when dealing with non-relocatable symbols, which at least in user space is the exception rather than the rule. Without this change, we could end up displaying unrealistically large offsets.

##### Checklist


jordalgo (approved):
LGTM. Thanks @danielocfb

PR #4088, opened 2025-05-02
parser: add ability to parse bare expressions

amscanne:
Stacked PRs:
 * #4089
 * __->__#4088

--- --- ---

### parser: add ability to parse bare expressions

Using a documented Bison trick, we can add the ability to parse some
source string as a full program or simply an expression. This will allow
us to add macro expansion as a later pass without needing a reparse.


jordalgo (approved):
LGTM

PR #4089, opened 2025-05-02
ast: remove need for macro expansion in the lexer

amscanne:
Stacked PRs:
 * __->__#4089

--- --- ---

### ast: remove need for macro expansion in the lexer

This change has a few tightly coupled parts.

First, we add `CDefinitions` as state produced by `ClangParse` instead
of having this generation in the `BPFtrace` object.

Next, the `Output` needs to be decoupled from `BPFtrace` also, because
the `CDefinitions` will only be available after `ClangParse`, but it is
needed by the output in order to pretty-print enum values.  The `Output`
is now plumbed explicitly when required. This is a general good anyways,
since untangling things from `BPFtrace` is an overall improvement.

Finally, we add a `CMacroExpansion` pass that expands the macros in
place, without the need for reparsing the full source.

### DIFF
--- a/src/ast/passes/c_macro_expansion.cpp
+++ b/src/ast/passes/c_macro_expansion.cpp
@@ -1,0 +1,79 @@
+#include <algorithm>
+
+#include "ast/ast.h"
+#include "ast/context.h"
+#include "ast/passes/c_macro_expansion.h"
+#include "ast/visitor.h"
+#include "clang_parser.h"
+#include "driver.h"
+#include "util/format.h"
+
+namespace bpftrace::ast {
+
+class CMacroExpander : public Visitor<CMacroExpander> {
+public:
+  CMacroExpander(ASTContext &ast, CDefinitions &c_definitions)
+      : ast_(ast), c_definitions_(c_definitions) {};
+
+  using Visitor<CMacroExpander>::visit;
+  void visit(Expression &expr);
+
+private:
+  ASTContext &ast_;
+  CDefinitions &c_definitions_;
+  std::vector<std::string> active_;
+};
+
+void CMacroExpander::visit(Expression &expr)
+{
+  // N.B. We only support raw identifier macros. The way expansion works is
+  // that we see if an expression is a bare identifier, then attempt expansion
+  // recurisvely.
+  if (auto *ident = expr.as<Identifier>()) {
+    if (c_definitions_.macros.contains(ident->ident)) {
+      const auto &value = c_definitions_.macros[ident->ident];
+
+      // Check for recursion.
+      if (std::ranges::find(active_, ident->ident) != active_.end()) {
+        ident->addError() << "Macro recursion: "
+                          << util::str_join(active_, "->");
+        return;
+      }
+
+      // Parse just the macro as an expression.
+      ASTContext macro(ident->ident, value);
+      Driver driver(macro);
+      auto expanded = driver.parse_expr();
+      if (!expanded) {
+        ident->addError() << "unable to expand macro as an expression: "
+                          << value;
+        return;
+      }
+
+      // Expand the macro expression in place.
+      expr.value = clone(ast_, expanded->value, ident->loc);
+
+      // Recursively visit the potentially expanded expression, ensuring that
+      // we can catch recursive expansion, per above.
+      active_.emplace_back(ident->ident);
+      Visitor<CMacroExpander>::visit(expr);
+      active_.pop_back();
+      return;
+    }
+  }
+
+  // Expand normally.
+  Visitor<CMacroExpander>::visit(expr);
+}
+
+Pass CreateCMacroExpansionPass()
+{
+  auto fn = [](ASTContext &ast, CDefinitions &c_definitions) {
+    CMacroExpander expander(ast, c_definitions);
+    expander.visit(ast.root);
+  };
+
+  return Pass::create("CMacroExpansion", fn);
+}
+
+} // namespace bpftrace::ast
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -687,17 +687,6 @@ int main(int argc, char* argv[])
     os = &outputstream;
   }
 
-  std::unique_ptr<Output> output;
-  if (args.output_format.empty() || args.output_format == "text") {
-    output = std::make_unique<TextOutput>(*os);
-  } else if (args.output_format == "json") {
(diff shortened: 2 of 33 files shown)


tests/portability_analyser.cpp
```diff
@@ -19,10 +19,13 @@ void test(BPFtrace &bpftrace, const std::string &input, int expected_result = 0)
   std::stringstream msg;
   msg << "\nInput:\n" << input << "\n\nOutput:\n";
 
+  CDefinitions nocdefs; // Output from clang parser.
```
jordalgo, line 22: nit
```suggestion
  CDefinitions no_c_defs; // Output from clang parser.
```

tests/codegen/regression.cpp
```diff
@@ -14,10 +14,13 @@ TEST(codegen, regression_957)
   ast::ASTContext ast("stdin", "t:sched:sched_one* { cat(\"%s\", probe); }");
   auto bpftrace = get_mock_bpftrace();
 
+  CDefinitions nocdefs; // Output from clang parser.
```
jordalgo, line 17:
```suggestion
  CDefinitions no_c_defs; // Output from clang parser.
```

src/output.h
```diff
@@ -268,9 +277,10 @@ class TextOutput : public Output {
 
 class JsonOutput : public Output {
 public:
-  explicit JsonOutput(std::ostream &out = std::cout,
+  explicit JsonOutput(CDefinitions &cdefinitions,
```
jordalgo, line 280:
```suggestion
  explicit JsonOutput(CDefinitions &c_definitions,
```

src/output.h
```diff
@@ -38,14 +39,20 @@ std::ostream &operator<<(std::ostream &out, MessageType type);
...
+                  std::ostream &out = std::cout,
+                  std::ostream &err = std::cerr)
+      : cdefinitions_(cdefinitions), out_(out), err_(err)
   {
   }
   Output(const Output &) = delete;
   Output &operator=(const Output &) = delete;
   virtual ~Output() = default;
 
+  virtual const CDefinitions &cdefinitions() const
```
jordalgo, line 52:
```suggestion
  virtual const CDefinitions &c_definitions() const
```

src/main.cpp
```diff
@@ -980,6 +973,20 @@ int main(int argc, char* argv[])
...
+  auto& cdefinitions = pmresult->get<CDefinitions>();
+  std::unique_ptr<Output> output;
+  if (args.output_format.empty() || args.output_format == "text") {
+    output = std::make_unique<TextOutput>(cdefinitions, *os);
+  } else if (args.output_format == "json") {
+    output = std::make_unique<JsonOutput>(cdefinitions, *os);
+  } else {
+    LOG(ERROR) << "Invalid output format \"" << args.output_format << "\"\n"
+               << "Valid formats: 'text', 'json'";
+    exit(1);
```
jordalgo, line 987:
```suggestion
    usage(std::cerr);
    exit(1);
```
?

src/ast/passes/c_macro_expansion.cpp
```diff
@@ -0,0 +1,76 @@
...
+};
+
+void CMacroExpander::visit(Expression &expr)
+{
+  // N.B. We only support raw identifier macros. The way expansion works is
+  // that we see if an expression is a bare identifier, then attempt expansion
+  // recurisvely.
+  if (auto ident = expr.as<Identifier>()) {
+    if (cdefinitions_.macros.contains(ident->ident)) {
+      auto &value = cdefinitions_.macros[ident->ident];
```
jordalgo, line 31: ?
```suggestion
      const auto &value = cdefinitions_.macros[ident->ident];
```

src/ast/passes/c_macro_expansion.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+      // Check for recursion.
+      if (active_.contains(ident->ident)) {
+        ident->addError() << "Macro recursion limit reached: " << ident->ident
```
jordalgo, line 35: There really isn't a "limit" right? It's just the detection of any recursion.

src/ast/passes/c_macro_expansion.cpp
```diff
@@ -0,0 +1,76 @@
...
+                          << ", " << value;
+        return;
+      }
+
+      // Parse just the macro as an expression.
+      ASTContext macro(ident->ident, value);
+      Driver driver(macro);
+      auto expanded = driver.parse_expr();
+      if (!expanded) {
+        ident->addError() << "unable to expand macro as an expression: "
```
jordalgo, line 45: should we add a test for this?

src/ast/passes/c_macro_expansion.cpp
```diff
@@ -0,0 +1,76 @@
...
+                          << value;
+        return;
+      }
+
+      // Expand the macro expression in place.
+      expr.value = clone(ast_, expanded->value, ident->loc);
+
+      // Recursively visit the potentially expanded expression, ensuring that
+      // we can catch recursive expansion, per above.
+      active_.emplace(ident->ident);
```
jordalgo, line 55: if you use a vector here like I did with the bpftrace macros then you can print the "path" to the recursion in the error.

jordalgo (approved):
Overall looks good. Is this on the way to doing pass before hand that separates idents for c-macros, c types, sized types, and builtins? ;)

src/ast/passes/c_macro_expansion.cpp
amscanne: Yes, exactly. I just kept the message the same. The "limit" part is just because the current code expands until it has expanded 10000 characters, and then decides that it is recursion :P

I can change the message and tests.

src/ast/passes/c_macro_expansion.cpp
amscanne: Done

src/ast/passes/c_macro_expansion.cpp
amscanne: Sure, done

src/ast/passes/c_macro_expansion.cpp
amscanne: I've clarified this. It's actually much cleaner now with the "expanded from" context that you fixed :D

jordalgo (approved):

Phoenix500526 (commented):
LGTM

PR #4090, opened 2025-05-02
util: add `TempDir` to create temporary directories

amscanne:
Stacked PRs:
 * #4051
 * #4050
 * #4049
 * #4034
 * __->__#4090

--- --- ---

### util: add `TempDir` to create temporary directories

In order to add a common working directory, add a `TempDir` utility
which can collect a common set of temporary files.

### DIFF
--- a/src/util/temp.cpp
+++ b/src/util/temp.cpp
@@ -3,6 +3,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 
+#include "log.h"
 #include "util/result.h"
 #include "util/temp.h"
 
@@ -14,7 +15,11 @@ void TempFileError::log(llvm::raw_ostream &OS) const
   OS << "temporary file " << origin_ << ": " << strerror(err_);
 }
 
-Result<TempFile> TempFile::create(std::string pattern)
+// Generic helper for providing a default pattern and ensuring that it is
+// mutable, in order to call the standard `mktemp` et al. Note that the
+// function `fn` must return `-errno` in the case of failure.
+static Result<std::pair<std::string, int>> mktemp(std::string pattern,
+                                                  std::function<int(char *)> fn)
 {
   if (pattern.empty()) {
     // Attempt to extract the best temporary directory. If the environment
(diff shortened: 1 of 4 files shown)


src/util/temp.cpp
```diff
@@ -32,13 +34,29 @@ Result<TempFile> TempFile::create(std::string pattern)
   // hold this result. This is used below as the actual filename.
   std::vector<char> mutable_pattern(pattern.size() + 1);
   ::strncpy(mutable_pattern.data(), pattern.c_str(), mutable_pattern.size());
-  int fd = ::mkostemp(mutable_pattern.data(), O_CLOEXEC);
+  int fd = fn(mutable_pattern.data());
   if (fd < 0) {
     int err = errno;
```
danobi, line 39: Seems fragile to rely on `fn()` to set errno. How about making `fn` return -errno to make it all in-band?

tests/temp.cpp
```diff
@@ -37,4 +38,31 @@ TEST(util, tempfile_good_pattern)
...
+{
+  auto d = TempDir::create();
+  ASSERT_TRUE(bool(d));
+  auto f1 = d->create_file("foo", false);
+  auto f2 = d->create_file("bar", false);
+  ASSERT_TRUE(bool(f1));
+  ASSERT_TRUE(bool(f2));
+  EXPECT_EQ(f1->path().filename().string(), "foo");
+  EXPECT_EQ(f2->path().filename().string(), "bar");
+  EXPECT_FALSE(bool(d->create_file("foo", false)));
```
danobi, line 65: Check for -EEXIST?

src/util/temp.cpp
amscanne: Sounds reasonable to me, I will make that change.

jordalgo: Reminder for this change.

src/util/temp.h
```diff
@@ -60,4 +62,41 @@ class TempFile {
...
+    return *this;
+  }
+
+  const std::filesystem::path &path()
+  {
+    return path_;
+  }
+
+  // Creates a temporary file in this directory. Note that `name` need not
+  // contain any `X` characters if `pattern` is true, as these will be appended
```
jordalgo, line 92: not sure I understand the 'X' characters thing...

jordalgo (approved):

src/util/temp.cpp
amscanne: Done

tests/temp.cpp
amscanne: Done

Issue #3712, opened 2025-01-15
Set up test harness for bpftrace language runtime


danobi:
### Is your feature request related to a problem? Please describe.
bpftrace has two major architectural components: compiler and language runtime. The compiler bits are fairly well represented in unit tests -- see `tests/semantic_analyser.cpp`, `tests/codegen/`, etc.

However, the language runtime bit has traditionally been tested in runtime tests (unfortunate naming coincidence). The runtime tests are actually end to end tests. They tend to be slow and more flaky. They'll be even slower after #3657. 

However, ever since introduction of `RequiredResources` abstraction, it should be fairly straightforward to add language runtime unit tests. See https://github.com/bpftrace/bpftrace/commit/e2192500f291777642af76f347744fb80c51c19b as example. That might be the first test of its kind as far as I can remember.

### Describe the solution you'd like

Set up a basic test harness for the language runtime bits. The entry point for the bulk of the logic is `BPFtrace::perf_event_printer`. Ideally we could mock out all the dependencies (including RequiredResources) and just unit test that API. Might be tricky, but worth a shot.

Alternatively, we could test the individual helpers `perf_event_printer` calls. Like the reference commit.

The new harness should live in its own file in `tests/`.

### Describe alternative solutions or features you've considered

Keep adding runtime tests when it could've been a unit test :(

danobi:
@Phoenix500526 I saw you looking at another unit testing related issue. This is a bigger one that would really help with our reliability efforts. Feel free to pick it up (or not).

Phoenix500526:
Hi，@danobi！ I understand this issue focuses on improving UT coverage for runtime components. Any existing e2e test in the runtime test suite should remain unchanged.

Could you help me locate ​the source code of these runtime components. Is there any documentation specifying their code locations? Or, could you indicate which directories/key files to begin with??

danobi:
Hi @Phoenix500526, sorry about the delay.

https://github.com/bpftrace/bpftrace/blob/b0d177d4cfc7749260158fad44d98654abaaa99f/src/bpftrace.cpp#L192 is the main runtime callback. This callback is passed data read out of the ring buffer and then parses it and takes some kind of action. 

[...]

PR #4092, opened 2025-05-04
Ut/perf_join

Phoenix500526:
Refactor the join implementation in the `perf_event_printer` and add unit tests to cover it.

##### Checklist


jordalgo:
Nit: would be good to add the PR summary to the umbrella issue for tracking.

src/asyncaction_handler.h
```diff
@@ -0,0 +1,50 @@
...
+#include "bpftrace.h"
+#include "log.h"
+#include <memory>
+#include <string>
+
+namespace bpftrace {
+
+const static size_t MAX_TIME_STR_LEN = 64;
+
+inline void asyncaction_join_handler(BPFtrace *bpftrace, void *data)
```
jordalgo, line 13: What about just having a regular header file with function definitions and an associated cpp file? Also maybe for the naming just put this into an "asyn_action" namespace and then you have just have `join_handler` as the function name or even just `join`?

src/asyncaction_handler.h
```diff
@@ -0,0 +1,50 @@
...
+    joined << arg;
+  }
+  bpftrace->out_->message(MessageType::join, joined.str());
+}
+
+inline void asyncaction_time_handler(BPFtrace *bpftrace, void *data)
+{
+  char timestr[MAX_TIME_STR_LEN]; // not respecting
+                                  // config_->get(ConfigKeyInt::max_strlen)
+  time_t t;
```
jordalgo, line 34:
```suggestion
  // not respecting config_->get(ConfigKeyInt::max_strlen)
  char timestr[MAX_TIME_STR_LEN];
  time_t t;
```

tests/asyncaction_handler.cpp
```diff
@@ -0,0 +1,100 @@
+#include "asyncaction_handler.h"
+#include "ast/async_event_types.h"
+#include "bpftrace.h"
+#include "mocks.h"
+#include "types.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace bpftrace::test::asyncaction_handler {
```
jordalgo, line 9: nit
```suggestion
namespace bpftrace::test::async_action_handler {
```

jordalgo:
Overall looks good. A few minor nits. Also the commit history is a bit weird in that the second to last creates a language_runtime test file and then the commit after that deletes it (maybe squash the last two ?)

src/asyncaction_handler.h
Phoenix500526: Done

tests/asyncaction_handler.cpp
Phoenix500526: Done

src/asyncaction_handler.h
Phoenix500526: It seems I cannot inlne these functions if I split these code into a regular header file and associated cpp file.

jordalgo: Personally, I'm fine with that. We can let the compiler of bpftrace make inlining decisions. One extra function call for the `join` output I think is acceptable.

jordalgo (commented):
Is there a larger plan to move all the async actions into this new `asyncaction_handler` class/file ? If so maybe it would be good to move the AsyncAction enum into this header file as well.

src/asyncaction_handler.h
Phoenix500526: Done

Phoenix500526:
> Is there a larger plan to move all the async actions into this new `asyncaction_handler` class/file ? If so maybe it would be good to move the AsyncAction enum into this header file as well.

The AsyncAction is also used in other *.cpp files, such as irbuilderbpf.cpp and codegen_llvm.cpp. These files shouldn't care about the async action handler function. I think maintaining the status quo may be a better choice.

jordalgo:
> The AsyncAction is also used in other *.cpp files, such as irbuilderbpf.cpp and codegen_llvm.cpp. These files shouldn't care about the async action handler function. I think maintaining the status quo may be a better choice.

The reason I like moving the types to asyncaction_handler is because the types.h file is just a monolith right now. Another options would be a different header like `async_actions.h`. This is not blocking this PR, just thinking longer term.

> Is there a larger plan to move all the async actions into this new asyncaction_handler class/file ?

Also what about this? I'm asking only because it feels a bit weird having an asyncaction_handler class where only some of the async actions are handled there and the others stay in bpftrace.cpp. If it's just for testability, these join/time functions could live as static functions in bpftrace.cpp right?

Phoenix500526:
> The reason I like moving the types to asyncaction_handler is because the types.h file is just a monolith right now. Another options would be a different header like `async_actions.h`. This is not blocking this PR, just thinking longer term.

OK, I'll leave a todo comment in issue #3712 

>  Is there a larger plan to move all the async actions into this new asyncaction_handler class/file ?

Yes, all the async action handlers will be moved into this new asyncaction_handler file.

jordalgo:
> Yes, all the async action handlers will be moved into this new asyncaction_handler file.

Ok great. Please also include that as a comment and/or in the issue.

Also, one last nit: can you rename the files to match the namespace "asyncaction_handler" -> "async_action_handler"

Phoenix500526:
> Ok great. Please also include that as a comment and/or in the issue.
> 
> Also, one last nit: can you rename the files to match the namespace "asyncaction_handler" -> "async_action_handler"

How about renaming async_action_handler to async_action and appending _handler as a suffix to the method? This would avoid a naming collision between async_action::time and the standard library's time function.

jordalgo:
> How about renaming async_action_handler to async_action and appending _handler as a suffix to the method? 

Sounds good to me.

Phoenix500526:
> Sounds good to me.

Done

Phoenix500526:
> @Phoenix500526 I still see the 'asyncaction_handler' instead of 'async_action' for the file names (and test names)

Done

tests/async_action.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "mocks.h"
+#include "types.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace bpftrace::test::async_action {
+
+using namespace bpftrace::async_action;
+
+TEST(asyncaction_handler, join)
```
jordalgo, line 13:
```suggestion
TEST(async_action_handler, join)
```
And for the rest of them, please 🙏

Phoenix500526: Done

jordalgo (approved):
LGTM. Nice work!

PR #4096, opened 2025-05-06
lexer: add support for adjacent string literal concatenation (#2458)

Egor-OSSRevival:
**This change introduces a new lexer state, STR_CONCAT, to handle string concatenation more effectively. It modifies the STR state to transition to STR_CONCAT after encountering a closing quote ("), allowing additional handling for concatenated strings or whitespace. New rules in STR_CONCAT manage spaces, return to STR for further strings, or finalize the string token. This improves support for multi-line strings, location tracking, and error handling in the lexer.**

##### Checklist

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -56,6 +56,7 @@ hex_esc  (x|X)[0-9a-fA-F]{1,2}
 oct_esc  [0-7]{1,3}
 
 %x STR
+%x STR_CONCAT
 %x STRUCT
 %x ENUM
 %x BRACE
@@ -171,7 +172,10 @@ oct_esc  [0-7]{1,3}
 
 \"                      { yy_push_state(STR, yyscanner); driver.buffer.clear(); }
 <STR>{
-  \"                    { yy_pop_state(yyscanner); return Parser::make_STRING(driver.buffer, driver.loc); }
+  \"                    { 
+                          yy_pop_state(yyscanner); 
+                          yy_push_state(STR_CONCAT, yyscanner);
+                        }
   [^\\\n\"]+            driver.buffer += yytext;
   \\n                   driver.buffer += '\n';
   \\t                   driver.buffer += '\t';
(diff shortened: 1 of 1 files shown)


jordalgo:
@Egor-OSSRevival Mind adding some examples of why this is needed? Also looks like tests are failing so those will need to all get fixed.

src/lexer.l
```diff
@@ -193,6 +197,24 @@ oct_esc  [0-7]{1,3}
...
+  {hspace}+             { driver.loc.step(); }
+  {vspace}+             { driver.loc.lines(yyleng); driver.loc.step(); }
+  \"                    { 
+                          yy_pop_state(yyscanner);
+                          yy_push_state(STR, yyscanner);
+                        }
+  .                     { 
+                          unput(yytext[0]); /* Put the character back */
+                          yy_pop_state(yyscanner);
+                          return Parser::make_STRING(driver.buffer, driver.loc);
```
amscanne, line 210: The change makes sense to me, I believe that the reason the tests are failing is because the location of the string has been extended to be the next non-whitespace character past the end of the string. I haven't dug into the details, maybe there's a relatively simple way to preserve the current location?

Egor-OSSRevival:
> @Egor-OSSRevival Mind adding some examples of why this is needed? Also looks like tests are failing so those will need to all get fixed.

You're right — I'll add examples to the description to clarify the motivation. In short, this change allows adjacent string literals to be concatenated, which aligns bpftrace more closely with C-style syntax and improves the ergonomics of writing longer or formatted strings without explicit concatenation.

For example, this currently fails:

```
printf("line 1: "
       "line 2\n");
```
With this change, it would work as expected, just like in C.

As for the failing tests, I’ll go through them and push fixes shortly. Let me know if you’d like to see any additional examples or edge cases covered.

src/lexer.l
Egor-OSSRevival: I think you're right — looks like the string location range is going too far because of how I'm handling the tokens. I’ll dig into it and try to preserve the original behavior while still supporting the new syntax.

Planning to work on it over the next couple days and will update the PR once it's sorted.

Egor-OSSRevival:
Hi all! @danobi @jordalgo @amscanne 
I wanted to clarify the current status of this PR in light of the related issue (#2458) being closed.

This PR (https://github.com/bpftrace/bpftrace/pull/4096) was opened two days before the issue was closed, and it directly addresses the request for supporting adjacent string literal concatenation — as described in both #2458 and earlier discussion around #2464.

From what I see, #3984 added support for compile-time string literal concatenation, but it's unclear whether that completely covers the original use case — particularly for adjacent quoted strings, like in:

`printf("Hello " "world\n");`

This PR enables exactly that behavior, with a dedicated lexer state (STR_CONCAT) for tracking such sequences, while preserving location tracking and parser behavior.

Should I continue refining this (e.g., fixing the failing tests), or is the functionality now considered fully covered by other changes? Just want to avoid duplicating effort if it's no longer needed.

Thanks!

jordalgo:
@Egor-OSSRevival Thanks for following up. I'm not opposed to the idea of supporting adjacent quoted strings but I'm not sure if it's worth a ton of additional complexity in the lexer, which we'd like to trim down. So I think it all depends on the implementation. TBH I've never heard this as being a problem for users but YMMV. cc @amscanne and @viktormalik

amscanne:
I think it's still a useful feature, since it mirrors what is present in C. It is slightly different than the compile-time evaluation (which can also fold through macros, etc.). I agree that as long as it's simple and tests are passing, it's worth merging.

PR #4097, opened 2025-05-07
Always fail if probes can't be attached

jordalgo:
This changes the default behavior for probes
that have multiple attach points or a wildcard.
Formerly we would just warn but this behavior
is inconsistent with single attach point probes,
which error. Now we error by default and users
have to explicity set the 'missing_probes'
config to 'warn'.

This also includes an improvement to the count
of actual probes being attached in the out message.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -35,6 +35,13 @@
 
 namespace bpftrace {
 
+char AttachError::ID;
+
+void AttachError::log(llvm::raw_ostream &OS) const
+{
+  OS << msg_;
+}
+
 bpf_probe_attach_type attachtype(ProbeType t)
 {
   // clang-format off
@@ -95,15 +102,16 @@ std::string progtypeName(libbpf::bpf_prog_type t)
   }
 }
 
-void AttachedProbe::attach_fentry()
+Result<> AttachedProbe::attach_fentry()
 {
   if (progfd_ < 0)
-    return;
+    return make_error<AttachError>();
 
   tracing_fd_ = bpf_raw_tracepoint_open(nullptr, progfd_);
   if (tracing_fd_ < 0) {
-    throw util::FatalUserException("Error attaching probe: " + probe_.name);
+    return make_error<AttachError>();
   }
+  return OK();
 }
 
 int AttachedProbe::detach_fentry()
@@ -112,9 +120,9 @@ int AttachedProbe::detach_fentry()
   return 0;
 }
 
-void AttachedProbe::attach_iter(std::optional<int> pid)
+Result<> AttachedProbe::attach_iter()
 {
-  if (!pid.has_value()) {
+  if (!pid_.has_value()) {
     linkfd_ = bpf_link_create(progfd_,
                               0,
                               static_cast<enum ::bpf_attach_type>(
@@ -124,7 +132,7 @@ void AttachedProbe::attach_iter(std::optional<int> pid)
     BPFTRACE_LIBBPF_OPTS(bpf_link_create_opts, opts);
     union bpf_iter_link_info linfo;
     memset(&linfo, 0, sizeof(linfo));
-    linfo.task.pid = *pid;
+    linfo.task.pid = *pid_;
     opts.iter_info = &linfo;
     opts.iter_info_len = sizeof(linfo);
     linkfd_ = bpf_link_create(progfd_,
@@ -135,8 +143,9 @@ void AttachedProbe::attach_iter(std::optional<int> pid)
   }
 
   if (linkfd_ < 0) {
-    throw util::FatalUserException("Error attaching probe: " + probe_.name);
+    return make_error<AttachError>();
   }
+  return OK();
 }
 
 int AttachedProbe::detach_iter()
@@ -145,20 +154,20 @@ int AttachedProbe::detach_iter()
   return 0;
 }
 
-void AttachedProbe::attach_raw_tracepoint()
+Result<> AttachedProbe::attach_raw_tracepoint()
 {
   tracing_fd_ = bpf_raw_tracepoint_open(nullptr, progfd_);
   if (tracing_fd_ < 0) {
-    if (tracing_fd_ == -ENOENT)
-      throw util::FatalUserException("Probe does not exist: " + probe_.name);
-    else if (tracing_fd_ == -EINVAL)
-      throw util::FatalUserException(
-          "Error attaching probe: " + probe_.name +
-          ", maybe trying to access arguments beyond "
-          "what's available in this tracepoint");
-    else
-      throw util::FatalUserException("Error attaching probe: " + probe_.name);
+    if (tracing_fd_ == -ENOENT) {
+      return make_error<AttachError>("Probe does not exist: " + probe_.name);
+    } else if (tracing_fd_ == -EINVAL) {
+      return make_error<AttachError>(
+          "Maybe trying to access arguments beyond what's available in "
+          "this tracepoint");
+    }
+    return make_error<AttachError>();
   }
+  return OK();
 }
 
 int AttachedProbe::detach_raw_tracepoint()
@@ -167,60 +176,36 @@ int AttachedProbe::detach_raw_tracepoint()
   return 0;
 }
 
+Result<std::unique_ptr<AttachedProbe>> AttachedProbe::make(
+    Probe &probe,
+    const BpfProgram &prog,
+    std::optional<int> pid,
+    BPFtrace &bpftrace,
+    bool safe_mode)
+{
+  // Cannot use std::make_unique here b/c it won't have access to
+  // private constructor.
+  auto attached = std::unique_ptr<AttachedProbe>(
+      new AttachedProbe(probe, prog, pid, bpftrace, safe_mode));
+
+  auto result = attached->attach();
+  if (!result)
+    return result.takeError();
+
+  return attached;
+}
+
 AttachedProbe::AttachedProbe(Probe &probe,
                              const BpfProgram &prog,
                              std::optional<int> pid,
                              BPFtrace &bpftrace,
                              bool safe_mode)
-    : probe_(probe), progfd_(prog.fd()), bpftrace_(bpftrace)
+    : probe_(probe),
+      progfd_(prog.fd()),
+      bpftrace_(bpftrace),
+      pid_(pid),
+      safe_mode_(safe_mode)
 {
-  LOG(V1) << "Attaching " << probe_.orig_name;
-  switch (probe_.type) {
-    case ProbeType::kprobe:
-      attach_kprobe();
-      break;
-    case ProbeType::kretprobe:
-      attach_kprobe();
-      break;
-    case ProbeType::tracepoint:
-      attach_tracepoint();
-      break;
-    case ProbeType::profile:
-      attach_profile(pid);
-      break;
-    case ProbeType::interval:
-      attach_interval(pid);
-      break;
-    case ProbeType::software:
-      attach_software(pid);
-      break;
-    case ProbeType::hardware:
-      attach_hardware(pid);
-      break;
-    case ProbeType::fentry:
-    case ProbeType::fexit:
-      attach_fentry();
-      break;
-    case ProbeType::iter:
-      attach_iter(pid);
-      break;
-    case ProbeType::rawtracepoint:
-      attach_raw_tracepoint();
-      break;
-    case ProbeType::usdt:
-      attach_usdt(pid, *bpftrace_.feature_);
-      break;
-    case ProbeType::watchpoint:
-    case ProbeType::asyncwatchpoint:
-      attach_watchpoint(pid, probe.mode);
-      break;
-    case ProbeType::uprobe:
-    case ProbeType::uretprobe:
-      attach_uprobe(pid, safe_mode);
-      break;
-    default:
-      LOG(BUG) << "invalid attached probe type \"" << probe_.type << "\"";
-  }
 }
 
 AttachedProbe::~AttachedProbe()
@@ -282,6 +267,44 @@ AttachedProbe::~AttachedProbe()
     close(progfd_);
 }
 
+Result<> AttachedProbe::attach()
+{
+  LOG(V1) << "Trying to attach probe: " << probe_.orig_name;
+  switch (probe_.type) {
+    case ProbeType::kprobe:
+    case ProbeType::kretprobe:
+      return attach_kprobe();
+    case ProbeType::tracepoint:
+      return attach_tracepoint();
+    case ProbeType::profile:
+      return attach_profile();
+    case ProbeType::interval:
+      return attach_interval();
+    case ProbeType::software:
+      return attach_software();
+    case ProbeType::hardware:
+      return attach_hardware();
+    case ProbeType::fentry:
+    case ProbeType::fexit:
+      return attach_fentry();
+    case ProbeType::iter:
+      return attach_iter();
+    case ProbeType::rawtracepoint:
+      return attach_raw_tracepoint();
+    case ProbeType::usdt:
+      return attach_usdt(*bpftrace_.feature_);
+    case ProbeType::watchpoint:
+    case ProbeType::asyncwatchpoint:
+      return attach_watchpoint(probe_.mode);
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+      return attach_uprobe(safe_mode_);
+    default:
+      LOG(BUG) << "invalid attached probe type \"" << probe_.type << "\"";
+  }
+  return OK();
+}
+
 const Probe &AttachedProbe::probe() const
 {
   return probe_;
@@ -329,16 +352,17 @@ std::string AttachedProbe::eventname() const
   }
 }
 
-static uint64_t resolve_offset(const std::string &path,
-                               const std::string &symbol,
-                               uint64_t loc)
+Result<uint64_t> AttachedProbe::resolve_offset(const std::string &path,
+                                               const std::string &symbol,
+                                               uint64_t loc)
 {
   bcc_symbol bcc_sym;
 
   if (bcc_resolve_symname(
-          path.c_str(), symbol.c_str(), loc, 0, nullptr, &bcc_sym))
-    throw util::FatalUserException("Could not resolve symbol: " + path + ":" +
+          path.c_str(), symbol.c_str(), loc, 0, nullptr, &bcc_sym)) {
+    return make_error<AttachError>("Could not resolve symbol: " + path + ":" +
                                    symbol);
+  }
 
   // Have to free sym.module, see:
   // https://github.com/iovisor/bcc/blob/ba73657cb8c4dab83dfb89eed4a8b3866255569a/src/cc/bcc_syms.h#L98-L99
@@ -352,12 +376,12 @@ static constexpr std::string_view hint_unsafe =
     "\nUse --unsafe to force attachment. WARNING: This option could lead to "
     "data corruption in the target process.";
 
-static void check_alignment(std::string &path,
-                            std::string &symbol,
-                            uint64_t sym_offset,
-                            uint64_t func_offset,
-                            bool safe_mode,
-                            ProbeType type)
+Result<> AttachedProbe::check_alignment(std::string &path,
+                                        std::string &symbol,
+                                        uint64_t sym_offset,
+                                        uint64_t func_offset,
+                                        bool safe_mode,
+                                        ProbeType type)
 {
   Disasm dasm(path);
   AlignState aligned = dasm.is_aligned(sym_offset, func_offset);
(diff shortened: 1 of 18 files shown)


jordalgo:
I also believe this makes the behavior of bpftrace more consistent as the number of attachpoints or syntax of a single attachpoint no longer changes how we handle probes not attaching.

src/tracepoint_format_parser.cpp
```diff
@@ -90,16 +90,9 @@ bool TracepointFormatParser::parse(ast::ASTContext &ctx, BPFtrace &bpftrace)
...
-            // Do not fail if trying to attach to multiple tracepoints
-            // (at least one of them could succeed)
-            bool fail = probe->attach_points.size() == 1;
             auto msg = "tracepoint not found: " + category + ":" + event_name;
             auto select = [&]() -> ast::Diagnostic & {
-              if (fail)
-                return ap->addError();
-              else
-                return ap->addWarning();
+              return ap->addWarning();
```
amscanne, line 95: The only thing `select` was doing is basically adding an error or warning diagnostic. Just drop this in favor of `auto &warn = ap->addWarning();`

tests/runtime/probe
```diff
@@ -137,7 +137,8 @@ AFTER ./testprogs/syscall read
 
 NAME kprobe_module_missing
 PROG kprobe:nonsense:vfs_read { printf("SUCCESS %d\n", pid); exit(); }
-EXPECT ERROR: Error attaching probe: kprobe:nonsense:vfs_read: specified module nonsense in probe kprobe:nonsense:vfs_read is not loaded.
+EXPECT WARNING: specified module nonsense in probe kprobe:nonsense:vfs_read is not loaded.
+EXPECT ERROR: Attachment failed for all (non-special) probes.
```
amscanne, line 141: `(non-special)` is weird, maybe we want to just call these "real probes"? "actual probes"? Worthwhile to standardize on a name here and in docs, possibly.

amscanne (approved):
The behavior seems reasonable, but I would:

- Get someone else to review also,
- Actually ensure that this is actually documented (ideally in the man page, but it's also something that should be explained pretty early for users). Having precise documentation might also shape the exact feature here, since having something easy-to-explain is probably an indication that it provides good ergonomics for the user

tests/runtime/probe
jordalgo: Yeah "non-special" is crappy. Let me marinate on some better wording. I could also just remove this as an error. If only `BEGIN` attaches - oh well. At least we told the user we couldn't attach these other ones.

jordalgo: Just removed this special case.

jordalgo:
>  I think we can all agree that silent data loss is probably the worst outcome for users, even if we technically did warn. It's on us to make it hard to lose data. 

I agree but I think, with this change, we're doing a pretty good job of notifying the users of the attachment failure. I also updated the "Attaching" count to indicate only one was attached e.g.

```
# sudo bpftrace -e 'kprobe:hrtimer_nanosleep { print(1); } kprobe:bad { print(2); }'

stdin:1:39-50: WARNING: bad is not traceable (either non-existing, inlined, or marked as "notrace"); attaching to it will likely fail
kprobe:hrtimer_nanosleep { print(1); } kprobe:bad { print(2); }
                                      ~~~~~~~~~~~
WARNING: Unable to attach probe: kprobe:bad. Skipping.
Attaching 1 probe...
```

I will also say that perhaps we're not giving users enough credit to actually test their script and read the output. Perhaps the big hammer here is the erroring by default of unattached probes, especially considering users don't have a way to disable that (for single attachpoint probes).

> They can organize the probe body with a `if (probe == "tracepoint:foo:bar") { tp_impl() } else { kprobe_impl() }.`

This is a good point that I hadn't considered before. I will say, one thing I like about this PR is that it makes probe attachment failure consistent and removes the special casing about when we warn vs error.

> So if we wanna go with the current approach, I'd suggest switching the default to error.

I would be ok with this but it would be a breaking change. I do like that it forces users to be explicit about how to handle probes that don't attach. Being consistent also paves a better path to supplying users with a syntax for conditional probe attachment (either with probe attributes or something like a `?` in front as @amscanne suggested).

jordalgo:
@viktormalik @fbs - Any thoughts on this?

jordalgo:
@danobi @viktormalik - Thanks for the feedback. I updated this PR so we now error by default for any probes that can't be attached - regardless of context (well, mostly, because watchpoint attach failures have some special casing). I think we can continue discussion of conditional probe attachment here: https://github.com/bpftrace/bpftrace/issues/3637

src/tracepoint_format_parser.cpp
```diff
@@ -90,33 +90,19 @@ bool TracepointFormatParser::parse(ast::ASTContext &ctx, BPFtrace &bpftrace)
           if (format_file.fail()) {
             // Errno might get clobbered by LOG().
             int saved_errno = errno;
-
-            // Do not fail if trying to attach to multiple tracepoints
-            // (at least one of them could succeed)
-            bool fail = probe->attach_points.size() == 1;
```
viktormalik, line 96: I assume that we will just print a warning here and fail elsewhere, is that correct?

tests/runtime/probe
```diff
@@ -68,17 +68,18 @@ REQUIRES_FEATURE fentry
...
 
 NAME fentry_ignore_missing_probes
 PROG config = { missing_probes = "ignore" } fentry:vfs_read,fentry:nonsense { exit(); }
-EXPECT_NONE WARNING: No BTF found for nonsense, skipping.
+EXPECT_NONE WARNING: No BTF found for nonsense.
 
 NAME fentry_error_missing_probes
 PROG config = { missing_probes = "error" } fentry:vfs_read,fentry:nonsense { exit(); }
-EXPECT ERROR: No BTF found for nonsense
+EXPECT ERROR: No BTF found for nonsense.
```
viktormalik, line 82: We should change this one to use `missing_probes = "warn"` to cover all values of the config.

tests/runtime/probe
```diff
@@ -353,7 +342,7 @@ AFTER ./testprogs/uprobe_test
 
 NAME uprobe_error_missing_probes
 PROG config = { missing_probes = "error" } uprobe:./testprogs/uprobe_test:uprobeFunction1,uprobe:./testprogs/uprobe_test:uprobeFunction3 { exit(); }
-EXPECT ERROR: Could not resolve symbol: ./testprogs/uprobe_test:uprobeFunction3, cannot attach probe.
+EXPECT ERROR: Could not resolve symbol: ./testprogs/uprobe_test:uprobeFunction3
```
viktormalik, line 345: Same here, we should use a warning in this test.

tests/runtime/watchpoint
```diff
@@ -32,7 +32,8 @@ REQUIRES_FEATURE signal
 
 NAME many_function_probes
 RUN {{BPFTRACE}} -e 'watchpoint:increment+arg0:4:w { printf("hit!\n") }' -c ./testprogs/watchpoint_func_many_probes
-EXPECT Failed to attach watchpoint probe. You are out of watchpoint registers.
+EXPECT ERROR: No more HW registers left
+EXPECT ERROR: Unable to attach probe: watchpoint:./testprogs/watchpoint_func_many_probes:increment:4:w. If this is expected set the 'missing_probes' config variable to 'warn'.
```
viktormalik, line 36: This is a bit tricky. IIUC, we will not fail in this situation, will we? So issuing an error and the hint is a bit strange.

viktormalik (commented):
This is a very nice refactoring, nice work!

Just a couple of comments and suggestions.

tests/runtime/watchpoint
jordalgo: Yeah, this one was weird. Maybe we should just change the behavior to fail as normal if there are no more hardware registers left (and warn if that what the user specified in the config). WDYT?

viktormalik: That sounds very much reasonable to me but I assume that there was a reason why it was done this way. Would be good to find it out.

jordalgo: @viktormalik was added here: https://github.com/bpftrace/bpftrace/commit/be64c5ba2c592f509befca11fa341b59340659df

I assume we're not aborting if there are no more hardware registers because it's happening in the middle of bpftrace execution (as opposed to at attach time) and perhaps that's a bad experience. It still could potentially cause odd behavior (lost data) in the script so erroring/aborting seems reasonable.

src/tracepoint_format_parser.cpp
jordalgo: Yeah exactly. This should fail to attach and warn/error/ignore as per the config.

src/attached_probe.cpp
```diff
@@ -352,12 +353,13 @@ static constexpr std::string_view hint_unsafe =
...
 
-static void check_alignment(std::string &path,
+static bool check_alignment(std::string &path,
                             std::string &symbol,
                             uint64_t sym_offset,
                             uint64_t func_offset,
                             bool safe_mode,
-                            ProbeType type)
+                            ProbeType type,
+                            std::string &err_msg)
```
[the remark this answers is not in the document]
jordalgo: > Or, maybe even better, why not make all these static functions members of AttachedProbe and write to err_msg_ directly?

Great idea. Done.

tests/runtime/watchpoint
viktormalik: I agree that erroring out makes sense but I'd like to get @danobi's opinion.

src/attached_probe.h
```diff
@@ -31,12 +31,28 @@ class AttachedProbe {
...
   int progfd() const;
   int linkfd_ = -1;
+  bool did_attach()
+  {
+    return attached_;
+  }
+  const std::string &err_msg() const
+  {
+    return err_msg_;
+  }
```
danobi, line 41: This out of band status / error mechanism feels odd. How about unifying around `Result<T>` now that we're using it elsewhere? It'll take a little more refactoring, as the AttachedProbe constructor can't return anything, but that could easily be turned into `AttachedProbe() = delete;` and then have a static factory method that's something like `static Result<AttachedProbe> attach(...);`

`BPFtrace::attach_probe()` can return a vector of `Result<AttachedProbe>` and then still have `BPFtrace::run()` do all the error handling.

WDYT?

tests/runtime/watchpoint
danobi: Yeah erroring out makes sense to me

src/attached_probe.h
jordalgo: Sounds like a bit of a pain TBH 😄 but I agree that it's a better pattern. Mind if I look at doing that refactoring in a follow-up PR? I think it'll be a cleaner diff.

jordalgo: @danobi ok, it wasn't too bad. I made this change, let me know if it's what you had in mind.

src/bpftrace.cpp
```diff
@@ -63,6 +64,25 @@ int BPFtrace::exit_code = 0;
 volatile sig_atomic_t BPFtrace::exitsig_recv = false;
 volatile sig_atomic_t BPFtrace::sigusr1_recv = false;
 
+void log_probe_attach_failure(const std::string &err_msg,
```
danobi, line 67:
```suggestion
static void log_probe_attach_failure(const std::string &err_msg,
```

src/attached_probe.cpp
```diff
@@ -282,6 +248,44 @@ AttachedProbe::~AttachedProbe()
...
+    case ProbeType::usdt:
+      return attach_usdt(*bpftrace_.feature_);
+    case ProbeType::watchpoint:
+    case ProbeType::asyncwatchpoint:
+      return attach_watchpoint(probe_.mode);
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+      return attach_uprobe(safe_mode_);
+    default:
+      LOG(BUG) << "invalid attached probe type \"" << probe_.type << "\"";
```
danobi, line 284: Nit: delete default arm? Should be exhaustive here

src/attached_probe.cpp
```diff
@@ -172,55 +181,12 @@ AttachedProbe::AttachedProbe(Probe &probe,
                              std::optional<int> pid,
                              BPFtrace &bpftrace,
                              bool safe_mode)
-    : probe_(probe), progfd_(prog.fd()), bpftrace_(bpftrace)
+    : probe_(probe),
+      progfd_(prog.fd()),
+      bpftrace_(bpftrace),
+      pid_(pid),
+      safe_mode_(safe_mode)
```
danobi, line 188: Why allow construction of `AttachedProbe` that hasn't been attached? IOW this is partial construction, as destructor depends on full construction. See `detach_iter()`, `detach_raw_tracepoint()`, etc.

It'd be more RAII friendly to have a static factory method that returns a fully attached `AttachedProbe`, as the name would imply. ie.

```
class AttachedProbe {
public:
  static Result<AttachedProbe> make_attached_probe(Probe &probe, ...);
  ~AttachedProbe();
  ...
private:
  AttachedProbe(Probe &probe, ...);
  ...
};
```

This removes the need for everything in the AttachedProbe destructor to consider whether or not attachment has been done. It also saves the `ap->attach()` calls from the various call sites. It also avoids the semantic drift in naming (see below).

src/bpftrace.cpp
```diff
@@ -650,38 +674,22 @@ std::vector<std::unique_ptr<AttachedProbe>> BPFtrace::attach_probe(
...
+    for (auto &ap : aps)
+      ret.emplace_back(std::move(ap));
+
+  } else if (probe.type == ProbeType::uprobe ||
+             probe.type == ProbeType::uretprobe) {
+    ret.emplace_back(std::make_unique<AttachedProbe>(
+        probe, program, pid, *this, safe_mode_));
+  } else {
+    ret.emplace_back(
+        std::make_unique<AttachedProbe>(probe, program, pid, *this));
```
danobi, line 692: Continuing from above, `attach_probe()` does not actually attach the probe anymore. Callers have to iterate through the returned vector and call `->attach()`. So probably better to go with static factory method and keep current code structure.

src/attached_probe.cpp
```diff
@@ -329,16 +333,17 @@ std::string AttachedProbe::eventname() const
...
+                                               uint64_t loc)
 {
   bcc_symbol bcc_sym;
 
   if (bcc_resolve_symname(
-          path.c_str(), symbol.c_str(), loc, 0, nullptr, &bcc_sym))
-    throw util::FatalUserException("Could not resolve symbol: " + path + ":" +
+          path.c_str(), symbol.c_str(), loc, 0, nullptr, &bcc_sym)) {
+    return make_error<AttachError>("Could not resolve symbol: " + path + ":" +
                                    symbol);
```
danobi, line 345: Aside: excellent place to use std::format and modernize the code in this file...

src/bpftrace.cpp
```diff
@@ -845,18 +852,28 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
...
     auto &begin_prog = bytecode_.getProgramForProbe((*begin_probe).second);
     if (::bpf_prog_test_run_opts(begin_prog.fd(), nullptr))
       return -1;
 
     LOG(V1) << "Attaching BEGIN";
+    ++num_special_attached;
+  }
+
+  // This is a little bit of a lie as we don't actually attach the END
+  // special probe until the end
```
danobi, line 868: I don't think it's a lie. In our abstract world of attachments, the probe is indeed attached and waiting for the END event to happen.

jordalgo: Ok ok - perhaps the language should be we don't actually RUN the END program until the end. Anyway, I can remove.

src/attached_probe.cpp
jordalgo: If we don't remove std::format usage ;)

src/attached_probe.cpp
jordalgo: > This removes the need for everything in the AttachedProbe destructor to consider whether or not attachment has been done.

What conditional logic in the destructor are you referring to? The closing of potentially negative or un-initialized file descriptors?

danobi: Right, we have those right now. But in future if detachment logic is more complex we don't want to consider attached and unattached cases.

src/attached_probe.cpp
jordalgo: Going to look at all the format conversion in another PR as soon as we decide what we want to do.

src/bpftrace.cpp
jordalgo: See comment above.

src/attached_probe.cpp
```diff
@@ -904,317 +1071,243 @@
...
+  return [perf_event_fd, probe_name = probe.name, usdt_sem_cleanup]() {
+    if (bpf_close_perf_event_fd(perf_event_fd))
+      LOG(WARNING) << "failed to close perf event FD for probe: " << probe_name;
+
+    usdt_sem_cleanup();
+  };
 }
 
-void AttachedProbe::attach_profile(std::optional<int> pid)
+Result<std::unique_ptr<AttachedProbe>> AttachedProbe::make_attached_probe(
```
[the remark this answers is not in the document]
amscanne: 🤣

src/attached_probe.cpp
jordalgo: @danobi and I discussed offline - going to save the refactor for a separate PR.

danobi (commented):
Sorry about the scrutiny, but IMO this is an exceptionally important codepath. I think it's worth the additional effort to keep it as maintainable as possible

src/bpftrace.cpp
```diff
@@ -510,13 +530,22 @@ std::vector<std::unique_ptr<AttachedProbe>> BPFtrace::attach_usdt_probe(
     Probe &probe,
     const BpfProgram &program,
     std::optional<int> pid,
-    bool file_activation)
+    bool file_activation,
+    bool &has_attach_errors)
```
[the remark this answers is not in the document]
jordalgo: Good call. Will update.

src/bpftrace.cpp
```diff
@@ -510,13 +530,22 @@ std::vector<std::unique_ptr<AttachedProbe>> BPFtrace::attach_usdt_probe(
...
   if (feature_->has_uprobe_refcnt() || !file_activation || probe.path.empty()) {
-    ret.emplace_back(
-        std::make_unique<AttachedProbe>(probe, program, pid, *this));
+    auto ap = AttachedProbe::make(probe, program, pid, *this, safe_mode_);
+    if (!ap) {
+      auto missing_probes = config_->missing_probes;
+      has_attach_errors = true;
+      auto ok = handleErrors(std::move(ap), [&](const AttachError &err) {
+        log_probe_attach_failure(err.msg(), probe.name, missing_probes);
+      });
```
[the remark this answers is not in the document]
jordalgo: Why do we need do that? Can't we just do `return make_error<AttachError>();` on the next line. Seems to work for me locally.

tests/runtime/probe
```diff
@@ -341,22 +329,23 @@ RUN {{BPFTRACE}} --unsafe -e 'uprobe:./testprogs/uprobe_test:_init { printf("arg
...
-EXPECT WARNING: Could not resolve symbol: ./testprogs/uprobe_test:uprobeFunction3, skipping probe.
+EXPECT ERROR: Could not resolve symbol: ./testprogs/uprobe_test:uprobeFunction3
+AFTER ./testprogs/uprobe_test
+WILL_FAIL
+
+NAME uprobe_warn_missing_probes
+PROG config = { missing_probes = "warn" } uprobe:./testprogs/uprobe_test:uprobeFunction1,uprobe:./testprogs/uprobe_test:uprobeFunction3 { exit(); }
+EXPECT WARNING: Could not resolve symbol: ./testprogs/uprobe_test:uprobeFunction3
 AFTER ./testprogs/uprobe_test
+WILL_FAIL
```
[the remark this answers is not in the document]
jordalgo: Oh, that's because `WILL_FAIL` is broken lol. Will fix in a separate PR but good catch!

jordalgo:
> Sorry about the scrutiny, but IMO this is an exceptionally important codepath. I think it's worth the additional effort to keep it as maintainable as possible

No worries! Thanks for the meticulous review.

src/bpftrace.cpp
```diff
@@ -861,6 +886,23 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
...
+
+  auto total_attached = num_attached + num_special_attached;
+
+  if (total_attached == 0) {
+    LOG(ERROR) << "Attachment failed for all probes.";
+    return -1;
+  }
+
+  if (!bt_quiet)
+    out.attached_probes(total_attached);
```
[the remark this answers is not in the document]
jordalgo: Blah - seems like "Attaching" is all over our docs and codebase. I can move it back if you and other folks feel strongly but I think it's a more valuable message reporting the actual number of probes that were attached instead of the number we're attempting to attach.

tests/runtime/probe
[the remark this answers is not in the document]
jordalgo: https://github.com/bpftrace/bpftrace/pull/4167

viktormalik (commented):
Great job on refactoring the `AttachedProbe` class, I really like it.

Just two comments below.

src/attached_probe.cpp
```diff
@@ -592,35 +615,27 @@ void AttachedProbe::attach_kprobe()
...
     }
     funcname = modname + ":" + funcname;
   }
 
-  // If the user requested to ignore warnings on non-existing probes and the
-  // function is not traceable, do not even try to attach as that would yield
-  // warnings from BCC which we don't want to see.
-  if (bpftrace_.config_->missing_probes == ConfigMissingProbes::ignore &&
-      probe_.name != probe_.orig_name &&
+  if (probe_.name != probe_.orig_name &&
```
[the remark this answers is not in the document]
jordalgo: Ah yeah, good catch.

tests/runtime/probe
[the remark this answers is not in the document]
jordalgo: Removed the WILL_FAIL

src/attached_probe.cpp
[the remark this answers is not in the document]
jordalgo: Edit: ok, fixed it. Just needed to move this check earlier in the function.

src/bpftrace.cpp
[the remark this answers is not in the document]
jordalgo: @ajor Mind if I do that in a follow-up PR? This one already has a lot of churn.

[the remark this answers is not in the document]
ajor: Yeah, it's not a big deal really

jordalgo:
Ok, anything I missed here? I want to merge so I can move on with my life 🚶

viktormalik (approved):
Let's go :rocket:

jordalgo:
Thanks everyone for the great review!

PR #4098, opened 2025-05-07
Refactor `helpererror` in `perf_event_printer`

Phoenix500526:
Extract helpererror logic into a separate function and add a unit test to cover it. This is the second pr of issue #3712 

##### Checklist

### DIFF
--- a/src/async_action.cpp
+++ b/src/async_action.cpp
@@ -45,4 +45,13 @@ void time_handler(BPFtrace *bpftrace, void *data)
   bpftrace->out_->message(MessageType::time, timestr, false);
 }
 
+void helper_error_handler(BPFtrace *bpftrace, void *data)
+{
+  auto *helper_error = static_cast<AsyncEvent::HelperError *>(data);
+  auto error_id = helper_error->error_id;
+  auto return_value = helper_error->return_value;
+  auto &info = bpftrace->resources.helper_error_info[error_id];
+  bpftrace->out_->helper_error(return_value, info);
+}
+
 } // namespace bpftrace::async_action
--- a/tests/async_action.cpp
+++ b/tests/async_action.cpp
@@ -1,6 +1,7 @@
 #include "async_action.h"
 #include "ast/async_event_types.h"
 #include "bpftrace.h"
+#include "location.hh"
 #include "mocks.h"
 #include "types.h"
 #include "gmock/gmock.h"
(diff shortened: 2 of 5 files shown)


tests/async_action.cpp
```diff
@@ -1,6 +1,7 @@
 #include "async_action.h"
 #include "ast/async_event_types.h"
 #include "bpftrace.h"
+#include "location.hh"
```
viktormalik, line 4: Not sure but I think that we need to add a dependency between the `bpftrace_test` and `parser` CMake targets to prevent failures like in #3986.

tests/async_action.cpp
```diff
@@ -99,4 +100,79 @@ TEST(async_action, time_invalid_format)
...
+  for (const auto &tc : test_cases) {
+    std::stringstream out;
+    auto bpftrace = get_mock_bpftrace(out);
+
+    auto src_loc = ast::SourceLocation(
+        location(&tc.filename, tc.line, tc.column));
+    auto location_chain = std::make_shared<ast::LocationChain>(src_loc);
+    HelperErrorInfo info(tc.func_id, location_chain);
+
+    bpftrace->resources.helper_error_info.emplace(tc.func_id, std::move(info));
```
viktormalik, line 156: This seems unnecessarily complicated with all the `make_shared` and `std::move`.

Wouldn't this be possible?
```suggestion
    auto src_loc = ast::SourceLocation(
        location(&tc.filename, tc.line, tc.column));
    auto location_chain = LocationChain(src_loc);

    bpftrace->resources.helper_error_info.emplace(tc.func_id, HelperErrorInfo info(tc.func_id, location_chain));
```

viktormalik (commented):
Just a couple of comments/suggestions for the tests, otherwise looks good.

tests/async_action.cpp
Phoenix500526: That's because the `HelperErrorInfo` requires a `ast::Location` type parameter and the `ast::Location` is a type alias of `std::shared_ptr<LocationChain>`

FYI: 
https://github.com/bpftrace/bpftrace/blob/master/src/required_resources.h#L30-L41
https://github.com/bpftrace/bpftrace/blob/master/src/ast/location.h#L67

tests/async_action.cpp
Phoenix500526: Done

Phoenix500526: The `bpftrace_test` depends on the `libbpftrace`, and the `libbpftrace` depends on the `parser`. So, do we really need to add such an explicit dependency between the `bpftrace_test` and `parser`? 🤔

src/async_action.cpp
```diff
@@ -45,4 +45,13 @@ void time_handler(BPFtrace *bpftrace, void *data)
   bpftrace->out_->message(MessageType::time, timestr, false);
 }
 
+void helpererror_handler(BPFtrace *bpftrace, void *data)
```
jordalgo, line 48: nit
```suggestion
void helper_error_handler(BPFtrace *bpftrace, void *data)
```
As you can probably tell I'm a stickler for readability 👓

src/async_action.cpp
```diff
@@ -45,4 +45,13 @@ void time_handler(BPFtrace *bpftrace, void *data)
...
+{
+  auto *helpererror = static_cast<AsyncEvent::HelperError *>(data);
```
jordalgo, line 50:
```suggestion
  auto *helper_error = static_cast<AsyncEvent::HelperError *>(data);
```

tests/async_action.cpp
```diff
@@ -99,4 +100,79 @@ TEST(async_action, time_invalid_format)
   EXPECT_THAT(log, testing::HasSubstr("strftime returned 0"));
 }
 
+TEST(async_action, helpererror)
```
jordalgo, line 103:
```suggestion
TEST(async_action, helper_error)
```

src/async_action.cpp
Phoenix500526: Done. 😄

src/async_action.cpp
Phoenix500526: Done

tests/async_action.cpp
Phoenix500526: Done

jordalgo (approved):
LGTM. Thanks!
I'll wait a day or two to see if @viktormalik has any follow-ups before merging

Phoenix500526:
Close as #4103 merged

PR #4102, opened 2025-05-08
Fix macro location expansion

jordalgo:
- Separate parent location chaining from added context. Now LocationChain has a vector of 'contexts' that can be added too in addition to a parent chain.
- Pass call location to cloned nodes in macro_expansion

##### Checklist


amscanne (approved):
LGTM, notwithstanding the parent ordering discussion but I think this can be revisited when macros are made "non-experimental"

Issue #2807, opened 2023-10-20
[RFC] Support for big loops


viktormalik:
It is currently not possible to have a loop with more than 1000 iterations:
```
# bpftrace -e 'BEGIN { $i = 0; while ($i < 10000) { @++; $i++ } exit() }' -v
[...]
The sequence of 8193 jumps is too complex.
[...]
```

This is because the verifier must enumerate all paths through the loop and it fails to do so if the loop has too many iterations.

The solution to this is to use the [bpf_loop](https://lwn.net/Articles/877062/) helper which takes a number of iterations and a callback function and executes that function the specified number of times. It is then simple for the verifier to see that if the callback function terminates, the loop terminates, too.

Once we have support for loading subprograms (part of #2710), we will be able to use the helper. I think that, despite the fact that big loops are probably not used often, it'd be a nice addition and I'd be willing to implement it. I just wanted to open a discussion on the syntax that we'll use.

Couple of options come to mind:
1. Wait for user-defined functions to land (#2624) and then use syntax similar to libppf's: `loop(iterations, callback_fn)`. This would be straightforward to implement but doesn't feel too nice for a high-level language that bpftrace is.
2. Add a new control-flow statement `loop` and use it similarly to `while`:
    ```
    BEGIN {
        loop(10000) {
            ...
        }
    }
    ```
    This should be still reasonably easy to implement. I'd require to cope with `break` and `continue` statements but they should correspond to `return 1` and `return 0` in the callback function, respectively.

    The disadvantage is that this introduces an additional language construct for implementing loops.
3. Use `bpf_loop` under the hood for `while` loops (where possible). This would be the cleanest from users point of view but could be tricky to implement as we'd have to deal with modification of the iteration counter. 

Of course it is possible to implement multiple approaches at the same time.

Personally, I prefer option 2 as it is quite common for languages to have separate statements for counted and non-counted loops. We could also detect the above verifier error and suggest the user to use `loop` instead of `while` in such a case.

What are other's thoughts on this?

jordalgo:
I do agree that option 2 seems more idiomatic to bpftrace but with user functions on the horizon, maybe option 1 is more re-usable. Is there a consideration for anonymous functions? Also would the function passed to the `loop` statement be required to take at least one argument being the counter?

viktormalik:
> Is there a consideration for anonymous functions? 

[...]

PR #4105, opened 2025-05-09
Add ncpus builtin to get the number of CPUs

Rtoax:
If we want to count data indexed by CPU, we cannot print out the statistics of the number of CPUs now. To use unroll() to iterate over all CPUs, we can only pass in the $[0-9] parameter or use a constant.

For example, to get the softirqs information of all CPUs:

    tracepoint:irq:softirq_entry {
            @vec_count[cpu, args->vec] += 1;
    }
    interval:s:1 {
            system("clear");
            $i = 0;
            unroll($1) {
                    /* Do the print */
                    printf("cpu %d\n", $i);
                    $i += 1;
            }
    }

If running without pass parameter:
    $ sudo bpftrace --unsafe script.bt
    10:13-23: ERROR: unroll minimum value is 1
            unroll($1) {
            ~~~~~~~~~~

We need pass number of cpus:
    $ sudo bpftrace --unsafe script.bt $(nproc)
    # works fine

This submit introduce 'nproc' to provide the number of processing units online.

    $i = 0;
    unroll(nproc) {
            /* Do something */
            $i += 1;
    }

Whole script:

```
#!/usr/bin/bpftrace --unsafe
/**
 * Display softirqs statistic on CPUs.
 */

#include <linux/interrupt.h>

BEGIN
{
	@softirqname[HI_SOFTIRQ] = "HI";
	@softirqname[TIMER_SOFTIRQ] = "TIMER";
	@softirqname[NET_TX_SOFTIRQ] = "NET_TX";
	@softirqname[NET_RX_SOFTIRQ] = "NET_RX";
	@softirqname[BLOCK_SOFTIRQ] = "BLOCK";
	@softirqname[IRQ_POLL_SOFTIRQ] = "IRQ_POLL";
	@softirqname[TASKLET_SOFTIRQ] = "TASKLET";
	@softirqname[SCHED_SOFTIRQ] = "SCHED";
	@softirqname[HRTIMER_SOFTIRQ] = "HRTIMER";
	@softirqname[RCU_SOFTIRQ] = "RCU";

	$nr_cpu = *kaddr("nr_cpu_ids");

	if ($# < 1) {
		printf("Must specify CPU number, max %d.\n", $nr_cpu);
		exit();
	}

	if ($1 > $nr_cpu) {
		printf("ERROR: too much CPU %d, max %d\n", $1, $nr_cpu);
		exit(1);
	}

	printf("Tracing softirq_entry, hit ctrl+c to end.\n");
}

tracepoint:irq:softirq_entry
{
	$vec_nr = args->vec;

	@vec_count[cpu, @softirqname[$vec_nr]] += 1;
}

interval:s:1
{
	$nr_cpu = *kaddr("nr_cpu_ids");

	system("clear");

	printf("Statistic Softirq... %d out of %d\n", $1, $nr_cpu);
	printf("\033[1;32m%-4s\033[m\033[1;33m %-10s %-10s %-10s %-10s %-10s ",
		"CPU", "HI", "TIMER", "NET_TX", "NET_RX", "BLOCK");
	printf("%-10s %-10s %-10s %-10s %-10s\033[m\n",
		"IRQ_POLL", "TASKLET", "SCHED", "HRTIMER", "RCU");

	$i = 0;
	unroll($1) {
	printf("%-4d %-10ld %-10d %-10d %-10d %-10d ",
		$i,
		@vec_count[$i, "HI"],
		@vec_count[$i, "TIMER"],
		@vec_count[$i, "NET_TX"],
		@vec_count[$i, "NET_RX"],
		@vec_count[$i, "BLOCK"]);
	printf("%-10d %-10d %-10d %-10d %-10d\n",
		@vec_count[$i, "IRQ_POLL"],
		@vec_count[$i, "TASKLET"],
		@vec_count[$i, "SCHED"],
		@vec_count[$i, "HRTIMER"],
		@vec_count[$i, "RCU"]);

		$i = $i + 1;
	}
}

END
{
	clear(@softirqname);
	clear(@vec_count);
	printf("Bye.\n");
}
```

$ sudo ./softirq_entry.bt $(nproc)

```
Statistic Softirq... 12 out of 12
CPU  HI         TIMER      NET_TX     NET_RX     BLOCK      IRQ_POLL   TASKLET    SCHED      HRTIMER    RCU       
0    8          667        0          0          0          0          0          6299       0          1710      
1    193        485        0          0          0          0          0          1050       0          981       
2    358        1067       0          0          0          0          0          1057       0          1409      
3    5          225        0          0          0          0          0          445        0          757       
4    404        2319       0          0          0          0          0          1944       0          2374      
5    164        1634       0          0          0          0          0          1220       0          1643      
6    521        134        0          0          0          0          0          874        0          1293      
7    50         378        0          0          0          0          0          586        0          1006      
8    53         262        0          0          0          0          0          369        0          892       
9    5178       1179       0          0          0          0          0          1850       0          2411      
10   216        442        0          0          28         0          0          831        0          1137      
11   427        864        0          83         0          0          0          997        0          1360
```

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1916,6 +1916,12 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | raw_smp_processor_id
 | ID of the processor executing the BPF program
 
+| ncpus
+| uint64
+| n/a
+| n/a
+| Number of CPUs
+
 | curtask
 | uint64
 | 4.8
--- a/tests/runtime/builtin
+++ b/tests/runtime/builtin
@@ -30,6 +30,10 @@ NAME cpu
 PROG i:ms:1 { printf("SUCCESS %lu\n", cpu); exit(); }
 EXPECT_REGEX SUCCESS [0-9]+
 
+NAME ncpus
+PROG BEGIN { printf("%ld\n", ncpus); exit(); }
+EXPECT_REGEX ^[1-9][0-9]*$
+
 NAME comm
 PROG BEGIN { printf("SUCCESS %s\n", comm); exit(); }
 EXPECT SUCCESS bpftrace
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to
 - Ustack and kstack symbols are automatically enhanced with debug info if available
   - [#3999](https://github.com/bpftrace/bpftrace/pull/3999)
 #### Added
+- Add ncpus builtin to get the number of CPUs.
+  - [#4105](https://github.com/bpftrace/bpftrace/pull/4105)
 - Use blazesym for user space address symbolization
   - [#3884](https://github.com/bpftrace/bpftrace/pull/3884)
 - Add simple block expressions
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -654,6 +654,12 @@ ScopedExpr CodegenLLVM::visit(Builtin &builtin)
   } else if (builtin.ident == "cpu") {
     Value *cpu = b_.CreateGetCpuId(builtin.loc);
     return ScopedExpr(b_.CreateZExt(cpu, b_.getInt64Ty()));
+  } else if (builtin.ident == "ncpus") {
+    return ScopedExpr(
+        b_.CreateLoad(b_.getInt64Ty(),
+                      module_->getGlobalVariable(
+                          to_string(bpftrace::globalvars::GlobalVar::NUM_CPUS)),
+                      "num_cpu.cmp"));
   } else if (builtin.ident == "curtask") {
     return ScopedExpr(b_.CreateGetCurrentTask(builtin.loc));
   } else if (builtin.ident == "rand") {
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -171,6 +171,9 @@ void ResourceAnalyser::visit(Builtin &builtin)
     // mark probe as using usym, so that the symbol table can be pre-loaded
     // and symbols resolved even when unavailable at resolution time
     resources_.probes_using_usym.insert(probe_);
+  } else if (builtin.ident == "ncpus") {
+    resources_.needed_global_vars.insert(
+        bpftrace::globalvars::GlobalVar::NUM_CPUS);
   }
 }
 
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -857,7 +857,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
              builtin.ident == "cgroup" || builtin.ident == "uid" ||
              builtin.ident == "gid" || builtin.ident == "cpu" ||
              builtin.ident == "rand" || builtin.ident == "numaid" ||
-             builtin.ident == "jiffies") {
+             builtin.ident == "jiffies" || builtin.ident == "ncpus") {
     builtin.builtin_type = CreateUInt64();
     if (builtin.ident == "cgroup" &&
         !bpftrace_.feature_->has_helper_get_current_cgroup_id()) {
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -44,7 +44,7 @@ hspace   [ \t]
 vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#$+\*])+
-builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
+builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ncpus|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
 
 int_type        bool|(u)?int(8|16|32|64)
 builtin_type    void|(u)?(min|max|sum|count|avg|stats)_t|probe_t|username|lhist_t|hist_t|usym_t|ksym_t|timestamp|macaddr_t|cgroup_path_t|strerror_t|kstack_t|ustack_t
--- a/tests/codegen/builtin_ncpus.cpp
+++ b/tests/codegen/builtin_ncpus.cpp
@@ -1,0 +1,14 @@
+#include "common.h"
+
+namespace bpftrace {
+namespace test {
+namespace codegen {
+
+TEST(codegen, builtin_ncpus)
+{
+  test("BEGIN { @x = ncpus; exit(); }", NAME);
+}
+
+} // namespace codegen
+} // namespace test
+} // namespace bpftrace
--- a/tests/codegen/llvm/builtin_ncpus.ll
+++ b/tests/codegen/llvm/builtin_ncpus.ll
@@ -1,0 +1,141 @@
+; ModuleID = 'bpftrace'
+source_filename = "bpftrace"
+target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n32:64-S128"
+target triple = "bpf-pc-linux"
+
+%"struct map_t" = type { ptr, ptr, ptr, ptr }
+%"struct map_t.0" = type { ptr, ptr }
+%"struct map_t.1" = type { ptr, ptr, ptr, ptr }
+%exit_t = type <{ i64, i8 }>
+
+@LICENSE = global [4 x i8] c"GPL\00", section "license", !dbg !0
+@AT_x = dso_local global %"struct map_t" zeroinitializer, section ".maps", !dbg !7
+@ringbuf = dso_local global %"struct map_t.0" zeroinitializer, section ".maps", !dbg !22
+@event_loss_counter = dso_local global %"struct map_t.1" zeroinitializer, section ".maps", !dbg !36
+@num_cpus = dso_local externally_initialized constant i64 0, section ".rodata", !dbg !48
+
+; Function Attrs: nounwind
+declare i64 @llvm.bpf.pseudo(i64 %0, i64 %1) #0
+
+; Function Attrs: nounwind
+define i64 @BEGIN_1(ptr %0) #0 section "s_BEGIN_1" !dbg !54 {
+entry:
+  %key = alloca i32, align 4
+  %exit = alloca %exit_t, align 8
+  %"@x_val" = alloca i64, align 8
+  %"@x_key" = alloca i64, align 8
+  %num_cpu.cmp = load i64, ptr @num_cpus, align 8
+  call void @llvm.lifetime.start.p0(i64 -1, ptr %"@x_key")
+  store i64 0, ptr %"@x_key", align 8
+  call void @llvm.lifetime.start.p0(i64 -1, ptr %"@x_val")
+  store i64 %num_cpu.cmp, ptr %"@x_val", align 8
+  %update_elem = call i64 inttoptr (i64 2 to ptr)(ptr @AT_x, ptr %"@x_key", ptr %"@x_val", i64 0)
+  call void @llvm.lifetime.end.p0(i64 -1, ptr %"@x_val")
+  call void @llvm.lifetime.end.p0(i64 -1, ptr %"@x_key")
+  call void @llvm.lifetime.start.p0(i64 -1, ptr %exit)
+  %1 = getelementptr %exit_t, ptr %exit, i64 0, i32 0
+  store i64 30000, ptr %1, align 8
+  %2 = getelementptr %exit_t, ptr %exit, i64 0, i32 1
+  store i8 0, ptr %2, align 1
+  %ringbuf_output = call i64 inttoptr (i64 130 to ptr)(ptr @ringbuf, ptr %exit, i64 9, i64 0)
+  %ringbuf_loss = icmp slt i64 %ringbuf_output, 0
+  br i1 %ringbuf_loss, label %event_loss_counter, label %counter_merge
+
+event_loss_counter:                               ; preds = %entry
+  call void @llvm.lifetime.start.p0(i64 -1, ptr %key)
+  store i32 0, ptr %key, align 4
+  %lookup_elem = call ptr inttoptr (i64 1 to ptr)(ptr @event_loss_counter, ptr %key)
+  %map_lookup_cond = icmp ne ptr %lookup_elem, null
+  br i1 %map_lookup_cond, label %lookup_success, label %lookup_failure
+
+counter_merge:                                    ; preds = %lookup_merge, %entry
+  call void @llvm.lifetime.end.p0(i64 -1, ptr %exit)
+  ret i64 0
+
+lookup_success:                                   ; preds = %event_loss_counter
+  %3 = atomicrmw add ptr %lookup_elem, i64 1 seq_cst, align 8
+  br label %lookup_merge
+
+lookup_failure:                                   ; preds = %event_loss_counter
+  br label %lookup_merge
+
+lookup_merge:                                     ; preds = %lookup_failure, %lookup_success
+  call void @llvm.lifetime.end.p0(i64 -1, ptr %key)
+  br label %counter_merge
+
+deadcode:                                         ; No predecessors!
+  ret i64 0
+}
+
+; Function Attrs: nocallback nofree nosync nounwind willreturn memory(argmem: readwrite)
+declare void @llvm.lifetime.start.p0(i64 immarg %0, ptr nocapture %1) #1
+
+; Function Attrs: nocallback nofree nosync nounwind willreturn memory(argmem: readwrite)
+declare void @llvm.lifetime.end.p0(i64 immarg %0, ptr nocapture %1) #1
+
+attributes #0 = { nounwind }
+attributes #1 = { nocallback nofree nosync nounwind willreturn memory(argmem: readwrite) }
+
+!llvm.dbg.cu = !{!50}
+!llvm.module.flags = !{!52, !53}
+
+!0 = !DIGlobalVariableExpression(var: !1, expr: !DIExpression())
+!1 = distinct !DIGlobalVariable(name: "LICENSE", linkageName: "global", scope: !2, file: !2, type: !3, isLocal: false, isDefinition: true)
+!2 = !DIFile(filename: "bpftrace.bpf.o", directory: ".")
+!3 = !DICompositeType(tag: DW_TAG_array_type, baseType: !4, size: 32, elements: !5)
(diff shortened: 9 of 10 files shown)


src/ast/ast.h
```diff
@@ -130,6 +130,7 @@ class Integer;
 class NegativeInteger;
 class PositionalParameter;
 class PositionalParameterCount;
+class CpuProcessorCount;
```
fbs, line 133: haven't kept up with development too much, but does this need to be its own ast node instead 'just another' builtin?

man/adoc/bpftrace.adoc
```diff
@@ -1916,6 +1916,12 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | raw_smp_processor_id
 | ID of the processor executing the BPF program
 
+| nproc
+| int64
+| n/a
+| n/a
+| The number of CPU processing units online.
```
fbs, line 1923: maybe good to add that this is (if i read correctly) a compile time constant and not runtime. If something happens to the cpu count the program has to be restarted?

jordalgo:
It might be a little silly but I'm trying to prevent adding new special idents to the lexer/parser. Could this be a function instead e.g. `nproc()` or `ncpus()` ?

Also I don't think we should compute this at compile time as that would have issues for Ahead of Time compilation. We currently have a global variable called `NUM_CPUS` that gets updated at runtime (for the reason I just mentioned). Can we just re-use that global?

src/ast/ast.h
Rtoax: Thanks for the reply, actually, at the first time i try new builtin:

```patch
diff --git a/src/ast/passes/codegen_llvm.cpp b/src/ast/passes/codegen_llvm.cpp
index e848bbdd..94cc3f5c 100644
@@ -654,6 +654,8 @@ ScopedExpr CodegenLLVM::visit(Builtin &builtin)
   } else if (builtin.ident == "cpu") {
     Value *cpu = b_.CreateGetCpuId(builtin.loc);
     return ScopedExpr(b_.CreateZExt(cpu, b_.getInt64Ty()));
+  } else if (builtin.ident == "nproc") {
+    return ScopedExpr(b_.getInt64(util::get_online_cpus().size()));
   } else if (builtin.ident == "curtask") {
     return ScopedExpr(b_.CreateGetCurrentTask(builtin.loc));
   } else if (builtin.ident == "rand") {
diff --git a/src/ast/passes/semantic_analyser.cpp b/src/ast/passes/semantic_analyser.cpp
index 5bfe8f69..09d8f55e 100644
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -857,7 +857,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
              builtin.ident == "cgroup" || builtin.ident == "uid" ||
              builtin.ident == "gid" || builtin.ident == "cpu" ||
              builtin.ident == "rand" || builtin.ident == "numaid" ||
-             builtin.ident == "jiffies") {
+             builtin.ident == "jiffies" || builtin.ident == "nproc") {
     builtin.builtin_type = CreateUInt64();
     if (builtin.ident == "cgroup" &&
         !bpftrace_.feature_->has_helper_get_current_cgroup_id()) {
diff --git a/src/lexer.l b/src/lexer.l
index c0829a3b..d3dea6d0 100644
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -44,7 +44,7 @@ hspace   [ \t]
 vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#$+\*])+
-builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
+builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|nproc|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
 
 int_type        bool|(u)?int(8|16|32|64)
 builtin_type    void|(u)?(min|max|sum|count|avg|stats)_t|probe_t|username|lhist_t|hist_t|usym_t|ksym_t|timestamp|macaddr_t|cgroup_path_t|strerror_t|kstack_t|ustack_t
```

# However

```
$ more b.bt 

BEGIN {
	$i = 0;
	unroll(nproc) {
		printf("%d\n", $i);
		$i += 1;
	}
}
```

```
$ sudo bpftrace b.bt
b.bt:4:2-15: ERROR: invalid unroll value
    unroll(nproc) {
    ~~~~~~~~~~~~~
```

Rtoax: Just as the following you said, `This is a compile time constant and not runtime`

Rtoax:
> It might be a little silly but I'm trying to prevent adding new special idents to the lexer/parser. Could this be a function instead e.g. `nproc()` or `ncpus()` ?
> 
> Also I don't think we should compute this at compile time as that would have issues for Ahead of Time compilation. We currently have a global variable called `NUM_CPUS` that gets updated at runtime (for the reason I just mentioned). Can we just re-use that global?

Thanks for the advise, could you give me more hint, need help!

jordalgo:
@Rtoax No problem.

Here is an example of a builtin function that is not part of the lexer/parser. The type is defined in semantic_analyser: https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/semantic_analyser.cpp#L1271

And then in codegen you'll also match on the function string:
https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/codegen_llvm.cpp#L1250

For this particular function, you'll load this global in codegen:https://github.com/bpftrace/bpftrace/blob/master/src/ast/irbuilderbpf.cpp#L844-L846

Rtoax:
@jordalgo i tried two different way:

# 1

```patch
diff --git a/src/ast/passes/codegen_llvm.cpp b/src/ast/passes/codegen_llvm.cpp
index e848bbdd..94cc3f5c 100644
@@ -654,6 +654,8 @@ ScopedExpr CodegenLLVM::visit(Builtin &builtin)
   } else if (builtin.ident == "cpu") {
     Value *cpu = b_.CreateGetCpuId(builtin.loc);
     return ScopedExpr(b_.CreateZExt(cpu, b_.getInt64Ty()));
+  } else if (builtin.ident == "nproc") {
+    return ScopedExpr(b_.getInt64(util::get_online_cpus().size()));
   } else if (builtin.ident == "curtask") {
     return ScopedExpr(b_.CreateGetCurrentTask(builtin.loc));
   } else if (builtin.ident == "rand") {
diff --git a/src/ast/passes/semantic_analyser.cpp b/src/ast/passes/semantic_analyser.cpp
index 5bfe8f69..09d8f55e 100644
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -857,7 +857,7 @@ void SemanticAnalyser::visit(Builtin &builtin)
              builtin.ident == "cgroup" || builtin.ident == "uid" ||
              builtin.ident == "gid" || builtin.ident == "cpu" ||
              builtin.ident == "rand" || builtin.ident == "numaid" ||
-             builtin.ident == "jiffies") {
+             builtin.ident == "jiffies" || builtin.ident == "nproc") {
     builtin.builtin_type = CreateUInt64();
     if (builtin.ident == "cgroup" &&
         !bpftrace_.feature_->has_helper_get_current_cgroup_id()) {
diff --git a/src/lexer.l b/src/lexer.l
index c0829a3b..d3dea6d0 100644
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -44,7 +44,7 @@ hspace   [ \t]
 vspace   [\n\r]
 space    {hspace}|{vspace}
 path     :(\\.|[_\-\./a-zA-Z0-9#$+\*])+
-builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
+builtin  arg[0-9]|args|cgroup|comm|cpid|numaid|cpu|nproc|ctx|curtask|elapsed|func|gid|pid|probe|rand|retval|sarg[0-9]|tid|uid|username|jiffies|nsecs|kstack|ustack
 
 int_type        bool|(u)?int(8|16|32|64)
 builtin_type    void|(u)?(min|max|sum|count|avg|stats)_t|probe_t|username|lhist_t|hist_t|usym_t|ksym_t|timestamp|macaddr_t|cgroup_path_t|strerror_t|kstack_t|ustack_t
```

# 2

```patch
diff --git a/src/ast/passes/codegen_llvm.cpp b/src/ast/passes/codegen_llvm.cpp
index e848bbdd3c1e..cb697fdbfefa 100644
@@ -1255,6 +1255,12 @@ ScopedExpr CodegenLLVM::visit(Call &call)
     return ScopedExpr(
         b_.CreateUSym(ctx_, scoped_arg.value(), get_probe_id(), call.loc),
         std::move(scoped_arg));
+  } else if (call.func == "nproc") {
+    return ScopedExpr(
+        b_.CreateLoad(b_.getInt64Ty(),
+                      module_->getGlobalVariable(
+                          to_string(bpftrace::globalvars::GlobalVar::NUM_CPUS)),
+                      "num_cpu.cmp"));
   } else if (call.func == "ntop") {
     // struct {
     //   int af_type;
diff --git a/src/ast/passes/resource_analyser.cpp b/src/ast/passes/resource_analyser.cpp
index 1c9418717502..5840adaa18f9 100644
--- a/src/ast/passes/resource_analyser.cpp
+++ b/src/ast/passes/resource_analyser.cpp
@@ -240,7 +240,7 @@ void ResourceAnalyser::visit(Call &call)
                                        : " ";
     resources_.join_args.push_back(delim);
   } else if (call.func == "count" || call.func == "sum" || call.func == "min" ||
-             call.func == "max" || call.func == "avg") {
+             call.func == "max" || call.func == "avg" || call.func == "nproc") {
     resources_.needed_global_vars.insert(
         bpftrace::globalvars::GlobalVar::NUM_CPUS);
   } else if (call.func == "hist") {
diff --git a/src/ast/passes/semantic_analyser.cpp b/src/ast/passes/semantic_analyser.cpp
index 5bfe8f69cec6..5d413457c2c4 100644
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -408,6 +408,9 @@ static const std::map<std::string, call_spec> CALL_SPEC = {
       .max_args=1,
       .arg_types={
         arg_type_spec{ .type=Type::timestamp_mode } } } },
+  { "nproc",
+    { .min_args=0,
+      .max_args=0 } },
   { "ntop",
     { .min_args=1,
       .max_args=2 } },
@@ -1279,6 +1282,8 @@ void SemanticAnalyser::visit(Call &call)
       call.return_type = CreateKSym();
     else if (call.func == "usym")
       call.return_type = CreateUSym();
+  } else if (call.func == "nproc") {
+    call.return_type = CreateInteger(64, 0);
   } else if (call.func == "ntop") {
     int index = 0;
     if (call.vargs.size() == 2) {
```
# invalid unroll value
```
	$i = 0;
	unroll(nproc()) { // or nproc
		printf("%d\n", $i);
		$i += 1;
	}
```

both:

```
./nproc.bt:6:2-17: ERROR: invalid unroll value
    unroll(nproc()) {
    ~~~~~~~~~~~~~~~
```

amscanne:
`unroll` is a bit special, because it strictly requires a constant value. You *could* support this with your approach "1" (essentially treat `nproc` as something evaluated purely at compile-time), but then I think you're creating a future issue for AOT.

So I think it makes sense to go ahead with "2" and just use a simple loop, e.g.:

```
$n = nprocs();
$i = 0;
while ($i < $n && $i < 1024) {
   printf("%d\n", $i);
   $i++;
}
```

(The `1024` there is just to help out the verifier and prove that it is bounded, not sure that it's strictly necessary)

Then maybe we can figure out the fit for macros to support something like `for_each_possible_cpu` that wraps the basic `nproc`? This is an interesting use case for that. (Doesn't need to be an extra block like that, but certainly something where we'd ideally be providing these helpers via a stdlib rather than adding more codegen.)

Rtoax:
Thanks, @fbs @jordalgo @amscanne , i just add `nproc()`, please review.

man/adoc/bpftrace.adoc
```diff
@@ -2579,6 +2583,14 @@ interval:s:1 {
  */
 ----
 
+[#functions-nproc]
+=== nproc
+
+.variants
+* `int64 nproc()`
```
jordalgo, line 2590: nit: should this be an unsigned int64?

jordalgo (commented):
Overall this looks good. I think an added runtime test would be great where you confirm the printed `nproc()` value is the same as the nproc on the system.

Also looks like the codegen needs to be regenerated.

jordalgo:
For `unroll`, we could allow this for the non-AOT case (even with version 2) and just do the lookup in a pass (maybe in fold_literals). In general, there is a lot of improvements (read utilization of BPF features) we can make to our handling of loops that would make this work for both AOT and not: https://docs.ebpf.io/linux/concepts/loops/

amscanne:
> For `unroll`, we could allow this for the non-AOT case (even with version 2) and just do the lookup in a pass (maybe in fold_literals). In general, there is a lot of improvements (read utilization of BPF features) we can make to our handling of loops that would make this work for both AOT and not: https://docs.ebpf.io/linux/concepts/loops/

I don't think we should introduce such cases that diverge. It is making our future lives much more complex.

There's a reasonable argument that perhaps you'd want to extend `for` to accept some kind of synthetic generator/set (i.e. a sequence) beyond maps, but I think this is a separate language discussion. We would want this to map logically into the `bpf_iter` framework, and be the preferred mechanism for creating loops generally.

amscanne:
(Just link the issues, it seems like #2963 and #2807 and pretty well-aligned proposals. I think that would let you do something like:

```
for ($cpu : 0..nprocs()) {
   ...
}
```

which seems pretty clean, and just builds on `nprocs` directly)

man/adoc/bpftrace.adoc
Rtoax: Fixed, thanks.

amscanne:
If you happen to have nix, updating the codegen tests is pretty easy with this script: https://github.com/bpftrace/bpftrace/blob/master/tests/codegen-tests.sh

ajor:
> It might be a little silly but I'm trying to prevent adding new special idents to the lexer/parser. Could this be a function instead e.g. `nproc()` or `ncpus()` ?

I don't think we should let implementation details inform the syntax. New idents in the lexer are nasty, but are something that we can fix at a later point (as has been done with functions at last, thank you!). Syntax is pretty much permanent.

`nproc` seems more consistent with the existing builtins/functions than `nproc()`, as it just returns a fixed value without any side effects or options.

Rtoax:
> > It might be a little silly but I'm trying to prevent adding new special idents to the lexer/parser. Could this be a function instead e.g. `nproc()` or `ncpus()` ?
> 
> I don't think we should let implementation details inform the syntax. New idents in the lexer are nasty, but are something that we can fix at a later point (as has been done with functions at last, thank you!). Syntax is pretty much permanent.
> 
> `nproc` seems more consistent with the existing builtins/functions than `nproc()`, as it just returns a fixed value without any side effects or options.

How about we keep it `nproc()` this patch, in future, we could consider `nproc(1)`'s parameters `--ignore=N`,`--all` and `sysconf(3)`:

```
$ getconf -a | grep NPROC # or sysconf(3)
_NPROCESSORS_CONF                  12
NPROCESSORS_CONF                   12
_NPROCESSORS_ONLN                  12
NPROCESSORS_ONLN                   12
```

jordalgo:
> I don't think we should let implementation details inform the syntax. 

Yeah, I agree. For `nproc`, if we do plan to add parameters in the future we should keep it a function but if not, then you're right it should probably be a builtin. `nproc(online)` does seem like it might be a good candidate...

ajor:
> How about we keep it `nproc()` this patch, in future, we could consider `nproc(1)`'s parameters `--ignore=N`,`--all` and `sysconf(3)`, like `nproc(conf)`,`nproc(online)`,`nproc(all)`:

The default should still not be a function call though - it behaviour is similar to `cpu`, `nsecs` and `kstack`, and should be syntactically similar too.

Looking at `man nproc`, I'm not sure we should be associating our builtin with the `nproc` CLI tool at all. It doesn't have the same semantics as our builtin:

> Print the number of processing units available to the current process, which may be less than the number of online processors

Our builtin is reading from `/sys/devices/system/cpu/possible`, so isn't accounting for any kind of CPU masking. I'm not sure if this file is cgroup-aware.

What about `ncpus` instead?

Rtoax:
> > How about we keep it `nproc()` this patch, in future, we could consider `nproc(1)`'s parameters `--ignore=N`,`--all` and `sysconf(3)`, like `nproc(conf)`,`nproc(online)`,`nproc(all)`:
> 
> The default should still not be a function call though - it behaviour is similar to `cpu`, `nsecs` and `kstack`, and should be syntactically similar too.
> 
> Looking at `man nproc`, I'm not sure we should be associating our builtin with the `nproc` CLI tool at all. It doesn't have the same semantics as our builtin:
> 
> > Print the number of processing units available to the current process, which may be less than the number of online processors
> 
> Our builtin is reading from `/sys/devices/system/cpu/possible`, so isn't accounting for any kind of CPU masking. I'm not sure if this file is cgroup-aware.
> 
> What about `ncpus` instead?

Thanks for your replay, it's makes sense. @jordalgo @amscanne @fbs  What do you think? If everyone agrees, I will redevelop the code according to `ncpus`. thanks everybody again :)

tests/README.md
```diff
@@ -156,6 +156,7 @@ not known until test time. The following runtime variables are available for the
 
 * `{{BPFTRACE}}`: Path to bpftrace executable
 * `{{BEFORE_PID}}`: Process ID of the process in the first `BEFORE` directive
+* `{{EXPECT_NCPUS}}`: The number of CPUs in all `EXPECT*` directive
```
jordalgo, line 159: we shouldn't make this a new directive, let's just use the output from a command line tool in the runtime test. If that's too much of a pain then we can forego the runtime test as this is a pretty straightforward addition.

tests/runtime/builtin
```diff
@@ -30,6 +30,10 @@ NAME cpu
 PROG i:ms:1 { printf("SUCCESS %lu\n", cpu); exit(); }
 EXPECT_REGEX SUCCESS [0-9]+
 
+NAME ncpus
+PROG BEGIN { printf("ncpus = %ld\n", ncpus); exit(); }
+EXPECT ncpus = {{EXPECT_NCPUS}}
```
jordalgo, line 35: If the assertion of exact number of cpus is too clunky, we can just do a regex test that the output is greater than 0 e.g. `^[1-9][0-9]*$`

jordalgo (commented):
Overall this looks good. Just should remove the new directive and fix-up the runtime test.

jordalgo (approved):
LGTM! Thanks for the contribution. Will merge once the tests pass.

PR #4108, opened 2025-05-10
Add a config to suppress warning messages

jordalgo:
Add log_warnings config variable

This is so script writers can suppress noisy or irrelevant output.

##### Checklist


src/ast/passes/config_analyser.cpp
```diff
@@ -62,6 +63,13 @@ void ConfigAnalyser::visit(AssignConfigVarStatement &assignment)
     assignment.addWarning()
         << assignment.var << " has been renamed, please use " << var;
   }
+
+  // Need this in a fairly early pass to make sure we don't get warnings
+  // in later passes.
+  if (!bpftrace_.config_->log_warnings) {
+    bpftrace_.helper_check_level_ = 0;
+    DISABLE_LOG(WARNING);
+  }
```
amscanne, line 72: I think the pass mutating the global log state is a bit of a no-no. Can we just put this into `main` after the passes have been run, but prior to actually running the pipelines? It should essentially be in the same place as the `--no-warnings` handler.

amscanne (approved):

jordalgo:
Note: this probably won't catch all instances of `LOG(WARNING)` if they occur before we disable them but most. @amscanne This is the downside of checking it after the pass runs.

src/ast/passes/config_analyser.cpp
jordalgo: > but ideally we would have a model where those are pushed to the individual probes

Do you mean something like probe attributes? 

The specifics on this change are that there were runtime warnings/errors for map functions in scripts where these errors were somewhat expected. So not really "probe" specific in that case. Maybe another solution might be named arguments for some of these map functions e.g. `delete(@a, 1, warnings=false)`.

src/main.cpp
```diff
@@ -966,6 +966,10 @@ int main(int argc, char* argv[])
     return 1;
   }
 
+  if (!bpftrace.config_->log_warnings) {
+    bpftrace.helper_check_level_ = 0;
+    DISABLE_LOG(WARNING);
+  }
```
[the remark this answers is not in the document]
jordalgo: I want the config to work in the same way as `--no-warnings` so suppress both compile-time and run-time warnings. We don't really make a distinction between the two at the moment - not to say we shouldn't but it probably needs more thought on the goals for treating the two separately.

man/adoc/bpftrace.adoc
```diff
@@ -3744,6 +3745,12 @@ Default: 1000000
 
 Log size in bytes.
 
+==== log_warnings
+
+Default: 1
+
+Print warnings to the log. If set to 0, this is similar to using the `--no-warnings` CLI option.
```
danobi, line 3752: Similar or the same?

tools/runqlat.bt
```diff
@@ -22,6 +22,10 @@
 #define TASK_RUNNING 0
 #endif
 
+config = {
+	log_warnings=0;
```
danobi, line 26: Document what warnings are suppressed? 

Same for below

danobi: Aside: in rust there is `#![allow(..)]` attribute syntax to suppress warnings/errors on specific parts of the source. This might compose well with our probe attributes desire, as attributes could be attached to arbitrary points of the AST. Suppression could be propagated down using a visitor and then diganostic class could check for suppression. It'd work best if we had warning categories just like rust, so the suppression is more targetted.

man/adoc/bpftrace.adoc
jordalgo: well... I hope it's the same but because of where they get read/applied we might miss some warnings. I'll just change it to the same.

tools/runqlat.bt
jordalgo: Yeah, I need to give our whole warning thing more thought. There is quite a mix (as @amscanne mentioned above) of runtime, compile time, and even developer-targeted warnings.

jordalgo:
Not going to land this yet. Still discussing options: https://github.com/bpftrace/bpftrace/issues/4169

PR #4112, opened 2025-05-11
Add check for signal probes

jordalgo:
Add a semantic analyser check for the
case when there are multiple signal
probes with the same signal.

Issue: https://github.com/bpftrace/bpftrace/issues/4110

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1230,7 +1230,7 @@ END {
 
 These are special built-in events provided by the bpftrace runtime.
 The trigger function is called by the bpftrace runtime when the bpftrace process receives specific events, such as a `SIGUSR1` signal.
-When multiple signal handlers are attached to the same signal, only the first one is used.
+Multiple signal probes for the same signal are not allowed.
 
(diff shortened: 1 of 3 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -1230,7 +1230,7 @@ END {
 
 These are special built-in events provided by the bpftrace runtime.
 The trigger function is called by the bpftrace runtime when the bpftrace process receives specific events, such as a `SIGUSR1` signal.
-When multiple signal handlers are attached to the same signal, only the first one is used.
+Multiple signal probes for the same signal are not allowed.
```
amscanne, line 1233: Why don't just we just add a pass that merges these probes into a sequence of blocks? Once you have imports, etc. that may add probes, it may start to get complex. This is a fairly simple, isolated pass.

jordalgo: Going with @viktormalik 's suggestion for now just to pave the way for supporting other signals. https://github.com/bpftrace/bpftrace/pull/4117

PR #4115, opened 2025-05-12
Make `blazesym` a required dependency

d-e-s-o:
Make `blazesym` a required dependency, lack of which causes build failure. Please see individual commits for descriptions.

##### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,7 +141,14 @@ if(ENABLE_SYSTEMD)
   pkg_check_modules(libsystemd REQUIRED IMPORTED_TARGET libsystemd)
 endif()
 
-find_package(LibBlazesym)
+find_package(LibBlazesym REQUIRED)
+if("${LIBBLAZESYM_VERSION}" VERSION_LESS 0.1.1)
+  message(SEND_ERROR "bpftrace requires libblazesym 0.1.1 or greater")
+endif()
+
+if("${LIBBLAZESYM_VERSION}" VERSION_GREATER_EQUAL 0.2.0)
+  message(SEND_ERROR "bpftrace requires libblazesym 0.1.x")
+endif()
 
 if(POLICY CMP0075)
   cmake_policy(SET CMP0075 NEW)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,10 +227,6 @@ if(ENABLE_SYSTEMD)
   set(BPFTRACE_FLAGS "${BPFTRACE_FLAGS}" HAVE_LIBSYSTEMD)
 endif()
 
-if(LIBBLAZESYM_FOUND)
-  set(BPFTRACE_FLAGS "${BPFTRACE_FLAGS}" HAVE_BLAZESYM)
-endif()
-
 add_subdirectory(src)
 if (BUILD_TESTING)
   add_subdirectory(tests)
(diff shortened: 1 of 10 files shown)


danielocfb:
Seems for the `static` CI workflow we require a package on Alpine. Any idea if somebody can help with that?

jordalgo:
@d-e-s-o Meaning we need someone to package up blazesym for Alpine?

danielocfb:
> @d-e-s-o Meaning we need someone to package up blazesym for Alpine?

It's what is used by the workflow. I suppose we can also change the workflow to use a different distribution or manually clone and build the library in the docker image. But I suspect getting it packaged is probably the most sensible way all things considered, yeah.

jordalgo:
@acj - Anyway we can convince you to package up blazesym in Alpine? I can send some bpftrace stickers and possibly some swag when I order more :gift:

acj:
Yep, happy to do that. The new package process usually takes a few days and needs to start in the `testing` repo before moving into `community`.

I see that v0.2.0 is cooking but not yet released. The Alpine maintainers usually prefer to have a proper release version rather than an RC, so would it be okay to use v0.1.1 as the Alpine package version and build from the [capi-v0.1.1](https://github.com/libbpf/blazesym/releases/tag/capi-v0.1.1) tag?

> bpftrace stickers and possibly some swag

I'll definitely take some when they're available :-) ty!

d-e-s-o:
> so would it be okay to use v0.1.1 as the Alpine package version and build from the [capi-v0.1.1](https://github.com/libbpf/blazesym/releases/tag/capi-v0.1.1) tag?

Yep, `blazesym-c` in version `0.1.1` is what we are looking for. Thanks!

jordalgo:
@acj You're awesome. Thanks so much. 🥇

acj:
First step: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/84179

When that lands, we can install the package from the edge repo to make sure it has what we need.

danielocfb:
> First step: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/84179
> 
> When that lands, we can install the package from the edge repo to make sure it has what we need.

Thanks @acj! I see the patch has been merged. I've updated the pull request to make use of the new package.

jordalgo:
@acj I think it's passing in our CI, is that enough to move this to `community` ?

acj:
I think so. Opened that MR: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/85232

Another step will be to add it to the 3.22 stable branch so that we don't need to depend on the edge (unstable) repo, but that can happen async and doesn't need to hold up this PR.

jordalgo (approved):

jordalgo:
Ok, great. Let's get this in. I added this task to remind us to stop using edge:
- https://github.com/bpftrace/bpftrace/issues/4231

viktormalik:
Sorry I'm late here but I think this was merged a bit prematurely. For the very least, it broke all of our distro build jobs, see here for the current master:
https://github.com/bpftrace/bpftrace/actions/runs/15509168370

jordalgo:
@viktormalik I've been tracking this but yeah, I didn't realize we were missing packages on both Fedora and Debian. Here is the issue I created:
- https://github.com/bpftrace/bpftrace/issues/4235

I'm currently chatting with folks to fix both. We can revert this if you want and wait till those are both stable, I don't feel strongly since we're not on the cusp of a release yet but maybe this hurts the development experience for people on Fedora or Debian (though they should be able to use nix).

viktormalik:
> I'm currently chatting with folks to fix both

That's good to hear!

> but maybe this hurts the development experience for people on Fedora or Debian (though they should be able to use nix).

Yeah, that's how I found it. I'm used to do a local build and now it fails and installing the currently provided blazesym Fedora package doesn't help. But sure, I can use Nix or install blazesym from source so definitely not a blocker.

jordalgo:
@viktormalik Then let's revert. It's silly to force this to be in master when it's not ready yet and causing issues. I'll make a PR (cc @d-e-s-o )

PR #4117, opened 2025-05-13
Add support for multiple signal probes

jordalgo:
##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -118,11 +118,15 @@ int BPFtrace::add_probe(ast::ASTContext &ctx,
   auto probe = generate_probe(ap, p, usdt_location_idx);
 
   // Add the new probe(s) to resources
-  if (ap.provider == "BEGIN" || ap.provider == "END" || ap.provider == "self") {
+  if (ap.provider == "BEGIN" || ap.provider == "END") {
     // special probes
     auto target = ap.target.empty() ? "" : "_" + ap.target;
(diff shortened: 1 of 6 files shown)


jordalgo:
https://github.com/bpftrace/bpftrace/issues/4110

src/bpftrace.cpp
```diff
@@ -118,11 +118,13 @@ int BPFtrace::add_probe(ast::ASTContext &ctx,
...
 
   // Add the new probe(s) to resources
-  if (ap.provider == "BEGIN" || ap.provider == "END" || ap.provider == "self") {
+  if (ap.provider == "BEGIN" || ap.provider == "END") {
     // special probes
     auto target = ap.target.empty() ? "" : "_" + ap.target;
     auto name = ap.provider + target;
     resources.special_probes[name] = std::move(probe);
+  } else if (ap.provider == "self") {
+    resources.signal_probes.emplace_back(std::move(probe));
```
viktormalik, line 127: For clarity, I would prefer
```suggestion
  } else if (ap.provider == "self") {
    if (ap.target == "signal")
      resources.signal_probes.emplace_back(std::move(probe));
```
or at least leave a comment that currently, `self` only supports signal probes.

viktormalik (approved):
Just one nit but not going to block on it.

src/bpftrace.cpp
jordalgo: Good call. Will update.

PR #4123, opened 2025-05-14
Ut/skboutput and exit

Phoenix500526:
This pr extracts `watchpoint_detach`, `exit`, and `skboutput` arm implementation in `perf_event_printer`.

Unit tests don't cover these three functions on the following grounds:
1. The `watchpoint_detach` is very simple. It just removes some elements in `bpftrace.attached_probes_`
2. The only test point in `exit` is the `request_finalize` method, whose test point has already been covered by `childproc.terminate`
3. The only test point of `skboutput` is the `write_pcaps` function and it's a simple wrapper of a third library function `pcap_dump` 

##### Checklist

### DIFF
--- a/src/async_action.h
+++ b/src/async_action.h
@@ -10,6 +10,7 @@ void join_handler(BPFtrace &bpftrace, Output &out, void *data);
 void time_handler(BPFtrace &bpftrace, Output &out, void *data);
 void helper_error_handler(BPFtrace &bpftrace, Output &out, void *data);
 void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data);
+void skboutput_handler(BPFtrace &bpftrace, void *data, int size);
 void syscall_handler(BPFtrace &bpftrace,
                      Output &out,
                      AsyncAction printf_id,
--- a/src/async_action.h
+++ b/src/async_action.h
@@ -11,6 +11,7 @@ void join_handler(BPFtrace &bpftrace, Output &out, void *data);
 void time_handler(BPFtrace &bpftrace, Output &out, void *data);
 void helper_error_handler(BPFtrace &bpftrace, Output &out, void *data);
 void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data);
+void watchpoint_detach_handler(BPFtrace &bpftrace, void *data);
 void skboutput_handler(BPFtrace &bpftrace, void *data, int size);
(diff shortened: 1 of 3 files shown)


src/async_action.h
```diff
@@ -6,10 +6,13 @@
 namespace bpftrace::async_action {
 
 const static size_t MAX_TIME_STR_LEN = 64;
+void exit_handler(BPFtrace &bpftrace, void *data);
 void join_handler(BPFtrace &bpftrace, Output &out, void *data);
 void time_handler(BPFtrace &bpftrace, Output &out, void *data);
 void helper_error_handler(BPFtrace &bpftrace, Output &out, void *data);
 void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data);
+void watchpoint_detach_handler(BPFtrace &bpftrace, void *data);
+void skboutput_handler(BPFtrace &bpftrace, void *data, int size);
```
amscanne, line 15: (Not an ask for this specific PR, just thinking out loud. Please don't change anything.)

In the distant future, I was thinking that we could make async dispatch purely dynamic, e.g. register a unique id for the required "async" function (perhaps defined directly in the bpfscript itself), and dispatch to a handler (that is either JITed or bundled into the AOT binary).

This refactoring is a great first step towards standardizing, and potentially enabling other pieces that can build towards something like this. What do you think of standardizing the dispatch functions? E.g. in this case, this is the only one which is accepting a size for the given buffer.

amscanne (approved):
LGTM, but @jordalgo or someone with authority will need to approve :)

src/async_action.h
Phoenix500526: Here is my understanding of your proposal:

A user can define a custom async function (e.g. foo) in a BPF script (e.g. user_func.bt) and use a built-in function like `register_async_func()` to register `foo`.

When executing `bpftrace user_func.bt`, the compiler compiles foo into either AOT (Ahead-of-Time) or JIT (Just-in-Time) code. And then the `register_async_func()`  generates a unique function ID for `foo` and registers the compiled (AOT/JIT) foo with an async dispatcher.
And the dispatcher Invokes the registered foo when its required input data becomes available and finally returns the result to the user.


Do I get your point correctly? I think we can file an issue for some further discussions.

jordalgo: @amscanne Has a little detail in this larger proposal but a separate issue would be great: https://github.com/bpftrace/bpftrace/issues/3773

PR #4124, opened 2025-05-14
parser: fix shift/reduce conflicts

amscanne:
Stacked PRs:
 * #4126
 * #4125
 * __->__#4124

--- --- ---

### parser: fix shift/reduce conflicts

This eliminates all expected shift/reduce conflicts which should make it
easy to make changes in the parser without unexpected changes (e.g.
where the number of conflicts stays the same, but they move around).

To do this, we introduce a pseudo-token, which allows us to explicitly
lower the precedence of conflicting reduce rules. The shift variant is
already being used in these cases, so this should not change parsing.

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -10,7 +10,7 @@
 %define define_location_comparison
 %define parse.assert
 %define parse.trace
-%expect 4
+%expect 0
 
 %define parse.error verbose
 
@@ -99,6 +99,9 @@ void yyerror(bpftrace::Driver &driver, const char *s);
   PTR        "->"
   STRUCT     "struct"
   UNION      "union"
+
+  // Pseudo token; see below.
+  LOW "low-precedence"
 ;
 
 %token <std::string> BUILTIN "builtin"
@@ -169,9 +172,19 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %type <ast::ConfigStatementList> config_assign_stmt_list config_block
 %type <SizedType> type int_type pointer_type struct_type
 %type <ast::Variable *> var
-%type <ast::Identifier *> raw_ident
 %type <ast::Program *> program
 
+
+// A pseudo token, which is the lowest precedence among all tokens.
+//
+// This helps us explicitly lower the precedence of a given rule to force shift
+// vs. reduce, and make the grammar explicit (still ambiguous, but explicitly
+// ambiguous). For example, consider the inherently ambiguous `@foo[..]`, which
+// could be interpreted as accessing the `@foo` non-scalar map, or indexing
+// into the value of the `@foo` scalar map, e.g. `(@foo)[...]`. We lower the
+// precedence of the associated rules to ensure that this is shifted, and the
+// longer `map_expr` rule will match over the `map` rule in this case.
+%left LOW
 
 %left COMMA
 %right ASSIGN LEFTASSIGN RIGHTASSIGN PLUSASSIGN MINUSASSIGN MULASSIGN DIVASSIGN MODASSIGN BANDASSIGN BORASSIGN BXORASSIGN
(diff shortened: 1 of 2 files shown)


src/parser.yy
```diff
@@ -173,6 +176,16 @@ void yyerror(bpftrace::Driver &driver, const char *s);
...
 
 
+// A pseudo token, which is the lowest precedence among all tokens.
+//
+// This helps us explicitly lower the precedence of a given rule to force shift
+// vs. reduce, and make the grammar explicit (still ambiguous, but explicitly
+// ambiguous). For example, consider the inherently ambiguous `@foo[..]`, which
+// could be interpreted as accessing foo, or indexing into the value of the
+// `@foo`, e.g.  (@foo)[...]. We lower the precedence of the brackets to ensure
+// that this is shifted, and the longer `mep_expr` rule will match.
```
jordalgo, line 186:
```suggestion
// that this is shifted, and the longer `map_expr` rule will match.
```

src/parser.yy
```diff
@@ -173,6 +176,16 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %type <ast::Program *> program
 
 
+// A pseudo token, which is the lowest precedence among all tokens.
+//
+// This helps us explicitly lower the precedence of a given rule to force shift
+// vs. reduce, and make the grammar explicit (still ambiguous, but explicitly
+// ambiguous). For example, consider the inherently ambiguous `@foo[..]`, which
+// could be interpreted as accessing foo, or indexing into the value of the
```
jordalgo, line 184:
```suggestion
// could be interpreted as an array access of `foo`, or indexing into the value of the
```

src/parser.yy
```diff
@@ -173,6 +176,16 @@ void yyerror(bpftrace::Driver &driver, const char *s);
...
+// `@foo`, e.g.  (@foo)[...]. We lower the precedence of the brackets to ensure
```
jordalgo, line 185:
```suggestion
// `@foo` map, e.g.  (@foo)[...]. We lower the precedence of the brackets to ensure
```

src/parser.yy
```diff
@@ -722,7 +737,7 @@ ident:
                 ;
 
 raw_ident:
```
jordalgo, line 739: how about just removing `raw_ident` all together then you can put `map %prec LOW` and `IDENT %prec LOW` next to each other in `primary_expr`?

src/parser.yy
amscanne: Hm, I think you have misunderstood this. I tried to clarify, along with the other one.

src/parser.yy
amscanne: Done.

src/parser.yy
amscanne: Sure, I will do this.

amscanne:
(Updating shortly, just finishing some changes in the stack.)

src/parser.yy
```diff
@@ -169,10 +172,20 @@ void yyerror(bpftrace::Driver &driver, const char *s);
...
 %type <ast::Program *> program
 
 
+// A pseudo token, which is the lowest precedence among all tokens.
+//
+// This helps us explicitly lower the precedence of a given rule to force shift
+// vs. reduce, and make the grammar explicit (still ambiguous, but explicitly
+// ambiguous). For example, consider the inherently ambiguous `@foo[..]`, which
+// could be interpreted as accessing the `@foo` non-scalar map, or indexing
+// into the value of the `@foo` scalar map, e.g. `(@foo)[...]`. We lower the
```
jordalgo, line 184: I'm still a bit confused. Is this clarification correct?
```suggestion
// could be interpreted as accessing the `@foo` non-scalar map (updating or reading the value at that key), or indexing
// into the value of the `@foo` scalar map, e.g. `(@foo)[...]` if it contained an array or pointer. We lower the
```

amscanne: Yes. I think whether the map contains an array or pointer is actually irrelevant, the syntax here is evaluated prior to any kind of type semantics. This is valid but ambiguous:
```
{ @val = curtask; print(@val[0]); }'
```

Presently it is the definition that is distinct from this:
```
{ @val = curtask; print((@val)[0]); }'
```

danobi (approved):

PR #4125, opened 2025-05-14
codegen: fix flushing of bitcode

amscanne:
Stacked PRs:
 * #4126
 * __->__#4125
 * #4124

--- --- ---

### codegen: fix flushing of bitcode

If a crash occurs during optimization, we may not see the full module.
This caused me to waste several hours things that the module was somehow
causing the crash. Nope, just `std::ostream` buffering.


amscanne:
The `ret` was correct (https://llvm.org/doxygen/Verifier_8cpp_source.html#l07308), and signifies `Broken`. There is just no output string. I will update this change to be just the flush.

viktormalik (approved):

jordalgo (approved):

PR #4126, opened 2025-05-14
codegen: implement `For` ranges

amscanne:
codegen: implement `For` ranges

This adds the `..` syntax for `For` loops, allows two expressions (as
long as they are unsigned types).


src/bpffeature.h
```diff
@@ -124,6 +124,7 @@ class BPFfeature {
   DEFINE_HELPER_TEST(for_each_map_elem, libbpf::BPF_PROG_TYPE_KPROBE);
   DEFINE_HELPER_TEST(get_ns_current_pid_tgid, libbpf::BPF_PROG_TYPE_KPROBE);
   DEFINE_HELPER_TEST(map_lookup_percpu_elem, libbpf::BPF_PROG_TYPE_KPROBE);
+  DEFINE_HELPER_TEST(loop, libbpf::BPF_PROG_TYPE_KPROBE);
```
jordalgo: Can we start adding comments to features with the kernel version where it was added? Will make it easier to do feature cleanup in the future.

tests/parser.cpp
```diff
@@ -2931,6 +2931,57 @@ BEGIN { for (@kv : @map) { } }
 )");
 }
 
+TEST(Parser, for_range)
```
jordalgo, line 2929: I feel like this could use some more tests but don't have anything specific in mind at the moment.

tests/semantic_analyser.cpp
```diff
@@ -4535,6 +4535,118 @@ BEGIN { @map[0] = count(); for ($kv : @map) { print($kv); } }
...
+stdin:1:19-23: ERROR: Can not access index '0' on expression of type 'int64'
+BEGIN { for ($i : 0.0..5) { printf("%d", $i); } }
+                  ~~~~
+stdin:1:19-26: ERROR: Loop range requires integer start
+BEGIN { for ($i : 0.0..5) { printf("%d", $i); } }
+                  ~~~~~~~
+)");
+}
+
+TEST(semantic_analyser, for_range_control_flow)
```
jordalgo, line 4603: Do we plan on adding these control flow bits later? Looks like these should be supported for `bpf_loop` and `bpf_for_each_map_elem`: "If callback_fn returns 0, the helper will continue to the next loop. If return value is 1, the helper will skip the rest of the loops and return. Other return values are not used now, and will be rejected by the verifier."

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2563,30 +2570,41 @@ void SemanticAnalyser::visit(For &f)
...
+  if (auto *map = f.iterable.as<Map>()) {
+    if (!map->type().IsMapIterableTy()) {
+      map->addError() << "Loop expression does not support type: "
+                      << map->type();
+      return;
+    }
+  } else if (auto *range = f.iterable.as<Range>()) {
+    if (is_final_pass()) {
+      if (!range->start.type().IsIntTy()) {
+        range->addError() << "Loop range requires integer start";
```
jordalgo: nit
```suggestion
        range->addError() << "Loop range requires an integer for the start";
```
(same for the end below)

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4687,22 +4699,136 @@ llvm::Function *CodegenLLVM::createForEachMapCallback(For &f, llvm::Type *ctx_t)
...
+  // Decl variable is not valid beyond this for loop.
   variables_[scope_stack_.back()].erase(f.decl->ident);
 
   b_.restoreIP(saved_ip);
   return callback;
 }
 
+ScopedExpr CodegenLLVM::visit(For &f, Range &range)
+{
+  // Evaluate our starting an endpoint values.
```
jordalgo:
```suggestion
  // Evaluate our starting and endpoint values.
```

src/bpffeature.h
amscanne: Sure, will do.

tests/semantic_analyser.cpp
amscanne: Sure, I'll add it to the stack here. It does seem pretty trivial to support break/continue. Supporting `return` is a bit more complex, so it'll be another separate change (maybe post-poned a bit more).

src/ast/passes/codegen_llvm.cpp
amscanne: Done.

src/ast/passes/semantic_analyser.cpp
amscanne: Done.

tests/semantic_analyser.cpp
amscanne: This will be separate.

tests/parser.cpp
amscanne: Let me know. I'll have more tests with the control flow stuff coming shortly.

jordalgo:
@amscanne FYI: I ran into a SEGV on my ASAN build for this PR when running this program:
```
BEGIN { @a = 1; }
i:s:1 { 
  for ($i : 0..@a) {
    print($i);
    @a += 1;
  }
}
```
Can you repro?

Ok found the issue. You need to update the map_sugar visit. Here is the updated code:
```
void MapDefaultKey::visit(For &for_loop)
{
  if (auto *map = for_loop.iterable.as<Map>()) {
    if (!check(*map, true)) {
      map->addError() << map->ident
                      << " has no explicit keys (scalar map), and "
                         "cannot be used for iteration";
    }
  } else if (auto *range = for_loop.iterable.as<Range>()) {
    visit(range->start);
    visit(range->end);
  }
  visit(for_loop.stmts);
}
```

Probably good to add a runtime test for this too.

jordalgo:
Also, probably good to call out in the docs that even though `$i` is being mutated the range is a snapshot and will not be updated during iteration e.g. this prints 0 through 9 instead of 0 through 4.
```
i:s:1 {
  $a = 10;
   for ($i : 0..$a) {
     print($i); 
     $a--;
    }
}
```

amscanne:
> @amscanne FYI: I ran into a SEGV on my ASAN build for this PR when running this program:
> 
> ```
> BEGIN { @a = 1; }
> i:s:1 { 
>   for ($i : 0..@a) {
>     print($i);
>     @a += 1;
>   }
> }
> ```
> 
> Can you repro?
> 
> Ok found the issue. You need to update the map_sugar visit. Here is the updated code:
> 
> ```
> void MapDefaultKey::visit(For &for_loop)
> {
>   if (auto *map = for_loop.iterable.as<Map>()) {
>     if (!check(*map, true)) {
>       map->addError() << map->ident
>                       << " has no explicit keys (scalar map), and "
>                          "cannot be used for iteration";
>     }
>   } else if (auto *range = for_loop.iterable.as<Range>()) {
>     visit(range->start);
>     visit(range->end);
>   }
>   visit(for_loop.stmts);
> }
> ```
> 
> Probably good to add a runtime test for this too.

I think I saw this just before you edited. I also found it and fixed it, and added a variant of your test. Good catch.

jordalgo (approved):
Had a final docs comment but non-blocking. Nice work!

amscanne:
I've updated the docs with information about mutability of the bounds and an example.

PR #4127, opened 2025-05-15
c++20: don't use std::format()

Rtoax:
On debian12(gcc 12.2.0, llvm 19.1.14):

    ....
    [ 18%] Building CXX object src/ast/CMakeFiles/ast.dir/passes/macro_expansion.cpp.o
    /home/rongtao/Git/bpftrace/bpftrace/src/ast/passes/macro_expansion.cpp:1:10:
    fatal error: format: No such file or directory
        1 | #include <format>
          |          ^~~~~~~~
    compilation terminated.

    ....
    [ 82%] Building CXX object src/ast/CMakeFiles/ast.dir/passes/unstable_feature.cpp.o
    /home/rongtao/Git/bpftrace/bpftrace/src/ast/passes/unstable_feature.cpp:2:10:
    fatal error: format: No such file or directory
        2 | #include <format>
          |          ^~~~~~~~
    compilation terminated.

gcc-13-3936-g1d9454aba615 and llvmorg-13-init-172-g081c1db02dd2 introduce format header. Because the clang/gcc version judgment conditions are more complicated, it is a good solution to directly replace std::format with std::string concatenation.

Link: https://github.com/gcc-mirror/gcc/commit/1d9454aba615
Link: https://github.com/llvm/llvm-project/commit/081c1db02dd2

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -34,6 +34,12 @@ Checks:
   - -readability-make-member-function-const
   - -readability-suspicious-call-argument
 
+CheckOptions:
+  # gcc and llvm did not introduce <format> when supporting C++20, so using
+  # std::format is temporarily not allowed.
+  # see issue https://github.com/bpftrace/bpftrace/issues/4185
+  modernize-use-std-format.ReplacementFormat: "DISALLOW"
+
 WarningsAsErrors: "*"
 UseColor: true
 HeaderFilterRegex: ".*"
--- a/src/ast/passes/macro_expansion.cpp
+++ b/src/ast/passes/macro_expansion.cpp
@@ -1,4 +1,3 @@
-#include <format>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
@@ -171,7 +170,7 @@ void MacroExpander::visit(Expression &expr)
 
 std::string MacroExpander::get_new_var_ident(std::string original_ident)
 {
-  return std::format("$${}_{}", macro_name_, original_ident);
+  return std::string("$$") + macro_name_ + std::string("_") + original_ident;
 }
 
 std::optional<Block *> MacroExpander::expand(Macro &macro, Call &call)
--- a/src/ast/passes/unstable_feature.cpp
+++ b/src/ast/passes/unstable_feature.cpp
@@ -1,5 +1,4 @@
 #include <cstring>
-#include <format>
 #include <string>
 
 #include "ast/ast.h"
@@ -14,18 +13,17 @@ namespace {
 
 std::string get_warning(const std::string &feature)
 {
-  return std::format("Script is using an unstable feature. To prevent this "
+  return std::string("Script is using an unstable feature. To prevent this "
                      "warning you must explicitly enable the unstable "
-                     "feature in the config e.g. {}=enable",
-                     feature);
+                     "feature in the config e.g. ") +
+         feature + std::string("=enable");
 }
 
 std::string get_error(std::string &&feature)
 {
-  return std::format(
-      "Feature not enabled by default. To enable "
-      "this unstable feature, set the config flag to enable. {}=enable",
-      feature);
+  return std::string("Feature not enabled by default. To enable "
+                     "this unstable feature, set the config flag to enable. ") +
+         feature + std::string("=enable");
 }
 
 class UnstableFeature : public Visitor<UnstableFeature> {


jordalgo:
Hmm, the lack of `<format>` also bit us internally at Meta when we imported bpftrace into our internal build system. @danobi @viktormalik - any thoughts on this?

danobi:
What's the use case for building upstream packages with such an old gcc? While std::format isn't really a huge deal, it puts us in an odd position of never being able to use newer features. We're already fairly conservative (staying on c++20) as it is.

Rtoax:
> What's the use case for building upstream packages with such an old gcc? While std::format isn't really a huge deal, it puts us in an odd position of never being able to use newer features. We're already fairly conservative (staying on c++20) as it is.

I completely agree. Unfortunately, while gcc supports C++20, it has not introduced the <format> header file. The problem is that the version of gcc that supports C++20 does not have the format header file. This patch attempts to resolve the issue without using std::format.

danobi:
Sure, I totally get that. I'm not necessarily opposed to removing std::format, but would first like to know why you need to use this older compiler. Is it just your development setup?

danobi:
Here's another place where we'd like to use std::format: https://github.com/bpftrace/bpftrace/pull/4097#discussion_r2094160550

Concatenating strings requires mental interpolation and uglier code. It's also less efficient, as it creates a bunch of string temporaries. The perf issue isn't a huge deal, we we're probably not doing string stuff in any fast paths, but it's really a 2 birds 1 stone kind of deal

Rtoax:
> Sure, I totally get that. I'm not necessarily opposed to removing std::format, but would first like to know why you need to use this older compiler. Is it just your development setup?

For example, on debian12 (newest released debian), gcc version is 12.11, and bcc could not be compiled on it.

```
rongtao@NUC11-Deb12:~$ cat /etc/debian_version 
12.11
rongtao@NUC11-Deb12:~$ gcc --v^C
rongtao@NUC11-Deb12:~$ sudo apt upgrade 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
rongtao@NUC11-Deb12:~$ gcc --version
gcc (Debian 12.2.0-14+deb12u1) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

danobi:
We synced on this and the current decision is to pull in fmtlib as a dependency. It should be packaged and available everywhere. fmtlib should be able to handle old compilers and thus let us use the new and improved formatting in C++ while also letting us migrate cleanly to std::format in the future when all compilers have it

Rtoax:
> @Rtoax Is this impacting your ability to build bpftrace for development or does this affect packaging on Debian?

No, it is not affected by packaging. I compile the source code directly.

Rtoax:
> We synced on this and the current decision is to pull in fmtlib as a dependency. It should be packaged and available everywhere. fmtlib should be able to handle old compilers and thus let us use the new and improved formatting in C++ while also letting us migrate cleanly to std::format in the future when all compilers have it

How do we introduce fmtlib? Should we use https://github.com/fmtlib/fmt as a submodule, or check whether fmtlib is supported in cmake? If it is clear, I can take on this part of the work.

viktormalik:
> How do we introduce fmtlib? Should we use https://github.com/fmtlib/fmt as a submodule, or check whether fmtlib is supported in cmake? If it is clear, I can take on this part of the work.

@Rtoax certainly not submodule.

I'd say that we should follow the same pattern as for the other libraries - introduce cmake/FindLibFmt.cmake and then call `find_package(LibFmt REQUIRED)` from CMakeLists.txt.

Rtoax:
@danobi @jordalgo @viktormalik Introduce https://github.com/fmtlib/fmt, please review, thanks :)

danobi:
Lack of static fmtlib is tough. It's MIT license and have native vendoring support:

```
Embedded: You can add the {fmt} source tree to your project and include it in your CMakeLists.txt file:

add_subdirectory(fmt)
target_link_libraries(<your-target> fmt::fmt)
```

https://fmt.dev/11.1/get-started/

How do we feel about just copy pasting the source?

Rtoax:
> Lack of static fmtlib is tough. It's MIT license and have native vendoring support:
> 
> ```
> Embedded: You can add the {fmt} source tree to your project and include it in your CMakeLists.txt file:
> 
> add_subdirectory(fmt)
> target_link_libraries(<your-target> fmt::fmt)
> ```
> 
> https://fmt.dev/11.1/get-started/
> 
> How do we feel about just copy pasting the source?

Thanks for your reply, how about add libfmt as submodule, because it has too much code, and in order to maintain consistency with the upstream, it is difficult to do code reduction.

```bash
# under fmt dir
$ cloc .
     132 text files.
     124 unique files.                                          
       9 files ignored.

github.com/AlDanial/cloc v 2.02  T=0.21 s (602.6 files/s, 402646.6 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
C/C++ Header                    24           5184           9935          26427
C++                             46           3525           4500          18474
Markdown                        12           1868             41           6942
YAML                            12            197            153           2281
Python                           6            237            190           1002
CMake                           12            148            197            901
Text                             1             47              0            243
Gradle                           1             11             28             93
Bourne Shell                     1             21             15             54
CSS                              1             11              0             50
Starlark                         3              3              2             26
CUDA                             1              6              9             13
make                             1              6              0              9
JavaScript                       1              0              0              4
SVG                              1              0              0              1
XML                              1              0              0              1
-------------------------------------------------------------------------------
SUM:                           124          11264          15070          56521
-------------------------------------------------------------------------------
```

Rtoax:
clang-tidy error, try to skip libfmt check:

```
[ 21/135][22.3s] /nix/store/zyx11fzfcq2id8fmfzcvnmipf6ifp8qg-clang-tools-20.1.1/bin/clang-tidy -p=build-clang-tidy -quiet --config-file=.clang-tidy /home/runner/work/bpftrace/bpftrace/src/ast/passes/unstable_feature.cpp
/home/runner/work/bpftrace/bpftrace/src/fmt/include/fmt/base.h:24:1: error: replace macro with enum [modernize-macro-to-enum,-warnings-as-errors]
   24 | #define FMT_VERSION 110200
      | ^~~~~~~
      |                     =
   25 | 
/home/runner/work/bpftrace/bpftrace/src/fmt/include/fmt/base.h:24:9: error: macro 'FMT_VERSION' defines an integral constant; prefer an enum instead [modernize-macro-to-enum,-warnings-as-errors]
   24 | #define FMT_VERSION 110200
      |         ^
```

danobi:
@Rtoax the reason we don't submodule in bpftrace is b/c at least a few distros have no-networking policies when building. And a submodule requires networking. I think submodules are fine in general (if not a little confusing to work with in git) for more proprietary/private projects, but for us we want to be packaged widely and it's just less friction overall to not submodule.

W.r.t. consistency with upstream, as long as we never modify code in the vendored subdirectory, it should be trivial to update.

And this is not a forever solution. It's just until compilers are updated more widely with full c++20 support. I expect in 2-ish years we'll delete it and move on.

Rtoax:
@danobi Thank you very much for your explanation. I have changed the `submodule` to `import the source code`. Please review it again.

jordalgo:
Is it possible to limit the files we import to the ones we actually need to make fmt work?

Rtoax:
> Is it possible to limit the files we import to the ones we actually need to make fmt work?

Makes sense, I'll try to remove Changelog/Doc/test.

Rtoax:
> > Is it possible to limit the files we import to the ones we actually need to make fmt work?
> 
> Makes sense, I'll try to remove Changelog/Doc/test. Maybe some content cannot be deleted without modifying CMakeLists.txt. https://github.com/fmtlib/fmt/blob/master/CMakeLists.txt#L308

remove CONTRIBUTING.md,doc,test. and set `FMT_INSTALL to OFF`.

jordalgo:
@Rtoax So sorry for the back and forth. This fmt import is looking a bit too messy for my taste. I think I'd rather (and I spoke with @danobi offline) we go back to your original solution of just not using `std::format` or any `format` and relying on string concatenation for now. We can make a task to look into switching over in 6 months / 1 year if there is wide enough support for `std::format`

Rtoax:
> @Rtoax So sorry for the back and forth. This fmt import is looking a bit too messy for my taste. I think I'd rather (and I spoke with @danobi offline) we go back to your original solution of just not using `std::format` or any `format` and relying on string concatenation for now. We can make a task to look into switching over in 6 months / 1 year if there is wide enough support for `std::format`

It's ok, Good things take time. Pushed, please review.

.clang-tidy
```diff
@@ -34,6 +34,10 @@ Checks:
   - -readability-make-member-function-const
   - -readability-suspicious-call-argument
 
+CheckOptions:
+  # std::format()
```
jordalgo, line 38: Maybe good to add a small explanation about the fact that `<format>` isn't included in some gcc versions that support c++20. And link to a todo issue.

jordalgo (commented):
@Rtoax Thanks for working on this. Just a few minor things and then we'll get this merged.

Rtoax:
> @Rtoax Thanks for working on this. Just a few minor things and then we'll get this merged.

Pushed, thanks, please review again.

src/ast/passes/macro_expansion.cpp
```diff
@@ -171,7 +170,7 @@ void MacroExpander::visit(Expression &expr)
 
 std::string MacroExpander::get_new_var_ident(std::string original_ident)
 {
-  return std::format("$${}_{}", macro_name_, original_ident);
+  return std::string("$$") + macro_name_ + std::string("_") + original_ident;
```
jordalgo, line 173: Not going to make you re-submit this PR but in the future you can just do `return "$$" + macro_name_ + "_" + original_ident;`

Rtoax: Thanks!

PR #4128, opened 2025-05-15
Make strings unsized from a user's point of view

### LINKED ISSUE #4120: String sizes should not be exposed to users
Strings are unsized from a user's point of view. It is fine to assign a larger string to a variable holding a smaller string, for example:
```
$x = "abc";
$x = "abcdefg";
```

The size tracking bpftrace does under the hood is an implementation detail that users shouldn't see.

```
# bpftrace -e 'BEGIN { $x = "abc"; $x = 1; }'
stdin:1:21-27: ERROR: Type mismatch for $x: trying to assign value of type 'int64' when variable already contains a value of type 'string[4]'
```
The error message exposes the size of the string variable as: `string[4]`. The error message should instead just talk about `string` (unsized).

I think this should also apply to variable and subprog declarations, which currently accept sized-strings.

Sized-string output should be kept for debug output, e.g. `-d ast`

ajor:
Fixes #4120

1. Don't allow users to specify string sizes for variables or subprogs
2. Hide string sizes from user-facing error messages

Sized-string output is kept in developer-facing messages, e.g. `-d ast`

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -967,7 +967,7 @@ tuple above:
 Just like with any variable the type is determined on first use and cannot be modified afterwards.
 This applies to both the key(s) and the value type.
 
-The following snippets create a map with key signature `(int64, string[16])` and a value type of `int64`:
+The following snippets create a map with key signature `(int64, string)` and a value type of `int64`:
 
 ----
 @[pid, comm]++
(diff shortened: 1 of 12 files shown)


jordalgo (approved):
Nice!

viktormalik (approved):

PR #4136, opened 2025-05-16
Add signed checks for map keys

jordalgo:
This makes sure we can print all uint64
map keys and that there is signed check
if the key is a integer (similar to how
we handle map value types).

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -212,7 +212,7 @@ private:
   void validate_new_key(const SizedType &current_key_type,
                         const SizedType &new_key_type,
                         const std::string &map_ident,
-                        const Node &node);
+                        const Expression &key_expr);
   bool update_string_size(SizedType &type, const SizedType &new_type);
   SizedType create_merged_tuple(const SizedType &left, const SizedType &right);
   void validate_map_key(const SizedType &key, Node &node);
@@ -2791,7 +2791,7 @@ void SemanticAnalyser::reconcile_map_key(Map *map, const Expression &key_expr)
 
   if (const auto &key = map_key_.find(map->ident); key != map_key_.end()) {
     update_current_key(key->second, new_key_type);
-    validate_new_key(key->second, new_key_type, map->ident, key_expr.node());
+    validate_new_key(key->second, new_key_type, map->ident, key_expr);
   } else {
     if (!new_key_type.IsNoneTy()) {
       map_key_.insert({ map->ident, new_key_type });
@@ -4037,7 +4037,7 @@ SizedType SemanticAnalyser::create_key_type(const SizedType &expr_type,
     // Store all integer values as 64-bit in map keys, so that there will
     // be space for any integer in the map key later
     // This should have a better solution.
-    new_key_type.SetSign(true);
+    new_key_type.SetSign(expr_type.IsSigned());
     new_key_type.SetIntBitWidth(64);
   }
 
@@ -4057,7 +4057,7 @@ void SemanticAnalyser::update_current_key(SizedType &current_key_type,
 void SemanticAnalyser::validate_new_key(const SizedType &current_key_type,
                                         const SizedType &new_key_type,
                                         const std::string &map_ident,
-                                        const Node &node)
+                                        const Expression &key_expr)
 {
   // Map keys can get resized/updated across multiple passes
   // wait till the end to log an error if there is a key mismatch.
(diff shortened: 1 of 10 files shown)


tests/runtime/map
```diff
@@ -19,6 +19,10 @@ NAME map indexed, scalar-like, with decl
 PROG let @a = hash(1); BEGIN { @a[0] = 0; exit(); }
 EXPECT @a[0]: 0
 
+NAME uint64 key
+PROG BEGIN { @a[10223372036854775807] = 0; exit(); }
```
danobi, line 23: just use `UINT64_MAX` (aka 18446744073709551615) to make it completely clear it's a u64?

tools/old/tcpdrop.bt
```diff
@@ -35,7 +35,7 @@ BEGIN
   printf("%-8s %-8s %-16s %-21s %-21s %-8s\n", "TIME", "PID", "COMM", "SADDR:SPORT", "DADDR:DPORT", "STATE");
 
   // See https://github.com/torvalds/linux/blob/master/include/net/tcp_states.h
-  @tcp_states[1] = "ESTABLISHED";
+  @tcp_states[(uint64)1] = "ESTABLISHED";
```
danobi, line 38: Reads a bit odd. How about using map declarations instead? Same for other scripts.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -4077,7 +4076,32 @@ void SemanticAnalyser::validate_new_key(const SizedType &current_key_type,
         valid = false;
       }
     } else if (!current_key_type.IsEqual(new_key_type)) {
-      valid = false;
+      if (current_key_type.IsIntegerTy()) {
+        auto *integer = key_expr.as<Integer>();
+        if (integer) {
+          uint64_t value = integer->value;
+          bool can_fit = false;
+          if (!current_key_type.IsSigned()) {
```
danobi, line 4084: Nit: flip branches to avoid negation

src/ast/passes/semantic_analyser.cpp
```diff
@@ -4077,7 +4076,32 @@ void SemanticAnalyser::validate_new_key(const SizedType &current_key_type,
...
+            auto min_max = getUIntTypeRange(current_key_type);
+            can_fit = value <= min_max.second;
+          } else {
+            auto min_max = getIntTypeRange(current_key_type);
+            can_fit = value <= static_cast<uint64_t>(min_max.second);
```
danobi, line 4089: I'm confused. For the signed case, why are we not also looking at the minimum value?

danobi: For example, this isn't expected, right?

```
$ sudo ./build/src/bpftrace -e 'BEGIN { let $x: int8 = -5; $x = -99999; print($x) }'
[sudo] password for dxu:
Attaching 1 probe...
-99999
^C
```

tools/old/tcpdrop.bt
jordalgo: you can't declare key/value types for map declarations.. yet

src/ast/passes/semantic_analyser.cpp
jordalgo: Map keys and values are automatically up-casted to 64 bit ints so even if you cast a value to a smaller int, it still sets it to 64 bits. Also your example is for variables.

tests/runtime/map
jordalgo: `UINT64_MAX` isn't available in our runtime framework as a const

tools/old/tcpdrop.bt
danobi: ah, right

src/ast/passes/semantic_analyser.cpp
danobi: Ah right, there's the NegativeInteger AST node now. Never mind. 

But I was showing the scratch var example b/c the code paths look similar. It's a bug above, right? 

Also relevant since I know about NegativeInteger node, should that be considered in these checks (separate PR territory)

tests/runtime/map
danobi: Right, I was just saying if you're gonna use a huge number as a magic value, might as well use UINT64_MAX to make it obvious it's a uint64

danobi (approved):

src/ast/passes/semantic_analyser.cpp
jordalgo: > It's a bug above, right?

Yeah, looks like it. I'll look into that.

> Also relevant since I know about NegativeInteger node, should that be considered in these checks

We actually don't have to worry about that at this point because of the underflow checks that happen earlier (and because of the 64 bit upcasting).

PR #4137, opened 2025-05-16
Parse implicit module BTF

danobi:
If a kprobe did not have an explicit module in the attachpoint, we would fail to parse that module's BTF while at the same point allowing the attachment. This leads to user confusion, as they expect types defined in that module to be available.

For example:

    kprobe:tun_net_xmit {
        $dev = (struct net_device *)arg1;
        $tun = (struct tun_struct *)$dev->priv;
    }
does not parse b/c `struct tun_struct` is unrecognized. However, the same probe using fentry worked just fine:

    fentry:tun_net_xmit {
        $tun = (struct tun_struct *)args->dev->priv;
        print($tun->numqueues);
    }

That was b/c fentry/fexit attachpoint parsing normalized the attachpoint and always inserted module information.

Ideally we'd fix kprobe by doing the same thing in attachpoint parser, but that ends up being somewhat tricky as a bunch of code expects there to not be a module name for the default vmlinux module. It is also backwards incompatible in a risky way, as `probe` builtin will start returning a different string.

So fix by teach BPFtrace::list_modules() to always try to extract module name for any kprobe/kretprobe function. Note we delete the calls to get_raw_tracepoint_modules() and get_func_modules(). Those are extraneous - the probe matcher already takes care of that work for us when returning us the attachpoint with the modules baked in.

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1956,20 +1956,24 @@ std::set<std::string> BPFtrace::list_modules(const ast::ASTContext &ctx)
       auto probe_type = probetype(ap->provider);
       if (probe_type == ProbeType::fentry || probe_type == ProbeType::fexit ||
           probe_type == ProbeType::rawtracepoint ||
-          ((probe_type == ProbeType::kprobe ||
-            probe_type == ProbeType::kretprobe) &&
-           !ap->target.empty())) {
-        if (ap->expansion != ast::ExpansionType::NONE) {
-          for (const auto &match : probe_matcher_->get_matches_for_ap(*ap)) {
-            std::string func = match;
-            util::erase_prefix(func);
-            auto match_modules = probe_type == ProbeType::rawtracepoint
-                                     ? get_raw_tracepoint_modules(func)
-                                     : get_func_modules(func);
-            modules.insert(match_modules.begin(), match_modules.end());
-          }
-        } else
-          modules.insert(ap->target);
+          probe_type == ProbeType::kprobe ||
+          probe_type == ProbeType::kretprobe) {
+        // Force module name into listing
+        bool clear_target;
+        if (ap->target.empty()) {
+          ap->target = "*";
+          clear_target = true;
+        } else {
+          clear_target = false;
+        }
+
+        for (std::string match : probe_matcher_->get_matches_for_ap(*ap)) {
+          auto module = util::erase_prefix(match);
+          modules.insert(module);
+        }
+
+        if (clear_target)
+          ap->target.clear();
       } else if (probe_type == ProbeType::tracepoint) {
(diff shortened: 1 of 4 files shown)


src/bpftrace.cpp
```diff
@@ -1956,20 +1956,24 @@ std::set<std::string> BPFtrace::list_modules(const ast::ASTContext &ctx)
       auto probe_type = probetype(ap->provider);
       if (probe_type == ProbeType::fentry || probe_type == ProbeType::fexit ||
           probe_type == ProbeType::rawtracepoint ||
-          ((probe_type == ProbeType::kprobe ||
-            probe_type == ProbeType::kretprobe) &&
-           !ap->target.empty())) {
-        if (ap->expansion != ast::ExpansionType::NONE) {
```
jordalgo, line 1962: Do we know what this `ast::ExpansionType::NONE` was doing?

jordalgo:
Looks good to me though maybe it would be good to add a runtime test to check that your failing example can now attach?

src/bpftrace.cpp
danobi: I think the idea was to avoid making a call to ProbeMatcher if the probe didn't need expansion (as presumably there would be no additional matches). But to fix this bug, it was simpler to put everything on the ProbeMatcher codepath rather than carve out further exceptions in the name of performance.

danobi:
> Looks good to me though maybe it would be good to add a runtime test to check that your failing example can now attach?

Runtime test for this would require loading a new module or finding something similar enough in the modules we already load in CI. Seemed flakey to me anyways. Unit test ought to be sufficient. I'd like us in the long run to avoid runtime tests where possible - slow, flaky, not parallelizable.

src/bpftrace.cpp
jordalgo: kk - makes sense.

jordalgo:
> Runtime test for this would require loading a new module or finding something similar enough in the modules we already load in CI. Seemed flakey to me anyways. 

There isn't a vmlinux example?

> I'd like us in the long run to avoid runtime tests where possible - slow, flaky, not parallelizable.

I feel like we've addressed most of the flakyness, no? I agree that we should abide by the test pyramid and have unit tests be the bulk but I find runtime tests pretty valuable for a tool like bpftrace. Perhaps we need an audit to see what runtime tests can be converted to unit tests (or removed as duplicates). Anyway, won't block on a runtime test here.

jordalgo (approved):

danobi:
No, it can't be vmlinux as that's always parsed. Has to be a symbol defined in a loadable module.

We addressed the flakiness but that was a whackamole effort. Being judicious about adding new possibly flaky ones would be good, especially now that we've kinda learned the common failure patterns.

PR #4140, opened 2025-05-17
cmake: handle the target LLVM not available gracefully

simpleton:
In PR #3580, it uses `llvm_config` to migrate the LLVM linking issue, but the `llvm_config` directly link the target LLVM[0], which break the bpftrace build in [facebookexperimental/ExtendedAndroidTools](https://github.com/facebookexperimental/ExtendedAndroidTools).

This PR falls back to original logic if LLVM target is not available.

[0] https://github.com/llvm/llvm-project/blob/7a8fe0f83c4ba03d07aef9243596d67af74a3b87/llvm/cmake/modules/LLVM-Config.cmake#L95

### DIFF
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -100,9 +100,15 @@ if(STATIC_LINKING)
   target_link_libraries(ast PUBLIC ${clang_libs})
   target_link_libraries(ast PUBLIC ${llvm_libs})
 else()
-  # llvm_config macro comes from the LLVM toolchain and will auto-resolve component
-  # names to library names. USE_SHARED option will tell llvm_config macro to prefer
-  # shared library / DLL on the system over the static libraries
-  llvm_config(ast USE_SHARED bpfcodegen ipo irreader mcjit orcjit)
+  if (TARGET LLVM)
+    # llvm_config macro comes from the LLVM toolchain and will auto-resolve component
+    # names to library names. USE_SHARED option will tell llvm_config macro to prefer
+    # shared library / DLL on the system over the static libraries
+    llvm_config(ast USE_SHARED bpfcodegen ipo irreader mcjit orcjit)
+  else()
+    # some build environment(e.g facebookexperimental/ExtendedAndroidTools) doesn't
+    # have libLLVM.so, drop `USE_SHARED` to avoid forcing link to `-lLLVM`.
+    llvm_config(ast bpfcodegen ipo irreader mcjit orcjit)
+  endif()
   target_link_libraries(ast PUBLIC libclang)
 endif()


src/ast/CMakeLists.txt
```diff
@@ -100,9 +100,17 @@ if(STATIC_LINKING)
   target_link_libraries(ast PUBLIC ${clang_libs})
   target_link_libraries(ast PUBLIC ${llvm_libs})
 else()
-  # llvm_config macro comes from the LLVM toolchain and will auto-resolve component
-  # names to library names. USE_SHARED option will tell llvm_config macro to prefer
-  # shared library / DLL on the system over the static libraries
-  llvm_config(ast USE_SHARED bpfcodegen ipo irreader mcjit orcjit)
+  find_package(LLVM QUIET CONFIG)
```
danobi, line 103: Why is this config mode search necessary? Wouldn't it also be necessary for above `llvm_map_components_to_libnames()`, as it's defined in the same source file as `llvm_config()`?

src/ast/CMakeLists.txt
```diff
@@ -100,9 +100,17 @@ if(STATIC_LINKING)
...
+  if (TARGET LLVM)
```
danobi, line 104: What's the significance of this check? I presume it means the config mode search succeeded, but that wouldn't explain why it's still ok to use `llvm_map_components_to_libnames()` and `llvm_expand_dependencies()` in the failure branch below, as they're defined in the same file as `llvm_config()`.

danobi (commented):
Also, does this mean android is not using static build anymore? I vaguely recall we've kept it around just for android.

src/ast/CMakeLists.txt
simpleton: The `llvm_config()` links the LLVM lib which causes the below error when I build bpftrace for Android.
```
ld.lld: error: unable to find library -lLLVM
```
code ref[0]:
```
target_link_libraries(${executable} PRIVATE LLVM)
```
`llvm_map_components_to_libnames ` and `llvm_expand_dependencies` are old code before https://github.com/bpftrace/bpftrace/pull/3580, which works fine.
image


[0] https://github.com/llvm/llvm-project/blob/7a8fe0f83c4ba03d07aef9243596d67af74a3b87/llvm/cmake/modules/LLVM-Config.cmake#L95

simpleton:
>  does this mean android is not using static build anymore? I vaguely recall we've kept it around just for android.

it supports both

src/ast/CMakeLists.txt
danobi: Ah ok, so your build environment doesn't have libLLVM.so. And the use of `llvm_config(.. USE_SHARED ..)` is forcing linking to the libLLVM.so. The `if (TARGET LLVM)` check seems more correct. How about dropping the `USE_SHARED` arg in the else branch instead of bringing back the old logic?

src/ast/CMakeLists.txt
```diff
@@ -100,9 +100,16 @@ if(STATIC_LINKING)
...
+    # llvm_config macro comes from the LLVM toolchain and will auto-resolve component
+    # names to library names. USE_SHARED option will tell llvm_config macro to prefer
+    # shared library / DLL on the system over the static libraries
+    llvm_config(ast USE_SHARED bpfcodegen ipo irreader mcjit orcjit)
+  else()
+    # Some build environment(e.g facebookexperimental/ExtendedAndroidTools) doesn't
+    # have libLLVM.so, drop `USE_SHARED` to avoid forcking link the `-lLLVM`.
```
danobi, line 111:
```suggestion
    # have libLLVM.so, drop `USE_SHARED` to avoid forcing link to `-lLLVM`.
```

danobi (approved):
lgtm, nit on comment

src/ast/CMakeLists.txt
danobi: Looking again, what happens if you delete this line?

Obviously nothing happens for me on my machine (due to what I'm describing in above comment). What about your setup?

Asking b/c once this goes in nobody else who develops this project will be able to reproduce what you're seeing, so unless we understand why it is the way it is it'll probably regress again eventually.

simpleton: sry for missing this comment. 

> Why is this config mode search necessary?

I believe LLVM was loaded by another CMake file earlier. it's not necessary here. thanks for pointing it out.

PR #4141, opened 2025-05-17
util: drop unused integer parsers

amscanne:
Merge the existing integer parsers into one simplified version. There is no longer any need to support parsing different types, only unsigned integers with the various qualifiers.

##### Checklist


tests/runtime/basic
```diff
@@ -195,25 +195,25 @@ WILL_FAIL
...
 RUN {{BPFTRACE}} -p 111a1 file.bt
-EXPECT ERROR: Failed to parse pid: pid '111a1' is not a valid decimal number
+EXPECT ERROR: Failed to parse pid: invalid trailing bytes: 111a1
 TIMEOUT 1
 WILL_FAIL
 
 NAME pid fails validation with non-numeric argument
 RUN {{BPFTRACE}} -p not_a_pid file.bt
-EXPECT ERROR: Failed to parse pid: pid 'not_a_pid' is not a valid decimal number
+EXPECT ERROR: Failed to parse pid: invalid integer: notapid
```
jordalgo, line 210: not super important but what happened to the underscores?

tests/parser.cpp
```diff
@@ -2479,20 +2468,20 @@ k:f { print(5e-9); }
...
   test_parse_failure("k:f { print(12e4); }", R"(
-stdin:1:7-17: ERROR: Coefficient part of scientific literal must be in range (0,9), got: 12
+stdin:1:7-17: ERROR: coefficient part of scientific literal must be 1-9: 12e4
 k:f { print(12e4); }
       ~~~~~~~~~~
 )");
 
   test_parse_failure("k:f { print(1_1e100); }", R"(
-stdin:1:7-20: ERROR: Coefficient part of scientific literal must be in range (0,9), got: 11
+stdin:1:7-20: ERROR: coefficient part of scientific literal must be 1-9: 11e100
```
jordalgo, line 2484: we have quite a mix of capitalization at the beginning of ERROR sentences - even in this PR 😉

tests/parser.cpp
```diff
@@ -2479,20 +2468,20 @@ k:f { print(5e-9); }
       ~~~~~~~~
 )");
 
-  test_parse_failure("k:f { print(1e17); }", R"(
-stdin:1:7-17: ERROR: Exponent will overflow integer range: 17
-k:f { print(1e17); }
+  test_parse_failure("k:f { print(1e21); }", R"(
+stdin:1:7-17: ERROR: overflow error, maximum value is 18446744073709551615: 1e21
```
jordalgo, line 2472: nit:
```suggestion
stdin:1:7-17: ERROR: overflow error, maximum value is 18446744073709551615. Found: 1e21
```

tests/parser.cpp
```diff
@@ -1538,35 +1535,31 @@ TEST(Parser, watchpoint_probe)
        "  int: 1\n");
 
   test_parse_failure("watchpoint:1b:8:w { 1 }", R"(
-stdin:1:1-18: ERROR: Found trailing non-numeric characters
-Invalid function/address argument
+stdin:1:1-18: ERROR: Invalid function/address argument: invalid trailing bytes: 1b
```
jordalgo, line 1538: How much of a pain is it to change it to this? (And some of the others below)
```suggestion
stdin:1:1-18: ERROR: Invalid function/address argument: 1b. Invalid trailing bytes.
```

jordalgo (commented):
Overall looks good. Nice cleanup. Just a few nits.

tests/runtime/basic
amscanne: Fixed.

tests/parser.cpp
amscanne: I'd rather keep it consistent and just have "error: <number>" as much as possible.

tests/parser.cpp
amscanne: I tend to anchor on always lower case for errors, but should capitalize log messages (and use complete sentences) when you are actually controlling the output. I'm used to this standard from Go (e.g. https://staticcheck.dev/docs/checks#ST1005).  This lets you chain errors effectively. Anyways, call me out if I'm doing something inconsistent from *that* angle (e.g. capitalizing the errors).

tests/parser.cpp
jordalgo: Would still be nice to fix this but it's not a blocker.

jordalgo (approved):

tests/parser.cpp
amscanne: It's tough to frame chained errors in this way, i.e. it assumes that you can return an error with multiple complete sentences (versus a sentence fragment that someone else can chain). I will submit as is, and we can leave the broader question of clean errors.

PR #4154, opened 2025-05-19
refactor: clean up print arm implementation in `perf_event_printer`

Phoenix500526:
This is the last pr of issue #3712. 

### Refactored BpfMap

* move some map relevant operations from `Bpftrace` to `BpfMap`

### Extracted handler functions and added comprehensive tests

* Moved `clear_map`, `zero_map`, and `print_map` handling from `perf_event_printer` to a dedicated handler function
* Created tests for `print_map` with various map types (`avg`, `max`, `lhist`, `hist`)
* Added tests for key sorting, `top` parameter, and `div` parameter behavior

## Not covered
The `clear_map` and `zero_map` handlers weren't specifically tested due to their simple logic.

##### Checklist

### DIFF
--- a/src/async_action.cpp
+++ b/src/async_action.cpp
@@ -77,6 +77,28 @@ void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data)
   out.value(bpftrace, ty, bytes);
 }
 
+void zero_map_handler(BPFtrace &bpftrace, void *data)
+{
+  auto *mapevent = static_cast<AsyncEvent::MapEvent *>(data);
+  const auto &map = bpftrace.bytecode_.getMap(mapevent->mapid);
+
+  auto err = bpftrace.zero_map(map);
+  if (err)
+    LOG(BUG) << "Could not zero map with ident \"" << map.name()
+             << "\", err=" << std::to_string(err);
+}
+
+void clear_map_handler(BPFtrace &bpftrace, void *data)
+{
+  auto *mapevent = static_cast<AsyncEvent::MapEvent *>(data);
+  const auto &map = bpftrace.bytecode_.getMap(mapevent->mapid);
+
+  auto err = bpftrace.clear_map(map);
+  if (err)
+    LOG(BUG) << "Could not clear map with ident \"" << map.name()
+             << "\", err=" << std::to_string(err);
+}
+
 void watchpoint_attach_handler(BPFtrace &bpftrace, Output &output, void *data)
 {
   auto *watchpoint = static_cast<AsyncEvent::Watchpoint *>(data);
--- a/src/async_action.cpp
+++ b/src/async_action.cpp
@@ -11,21 +11,21 @@
 
 namespace bpftrace::async_action {
 
-void exit_handler(BPFtrace &bpftrace, void *data)
+void exit_handler(BPFtrace &bpftrace, const void *data)
 {
-  auto *exit = static_cast<AsyncEvent::Exit *>(data);
+  const auto *exit = static_cast<const AsyncEvent::Exit *>(data);
   BPFtrace::exit_code = exit->exit_code;
   bpftrace.request_finalize();
 }
 
-void join_handler(BPFtrace &bpftrace, Output &out, void *data)
+void join_handler(BPFtrace &bpftrace, Output &out, const void *data)
 {
-  auto *join = static_cast<AsyncEvent::Join *>(data);
+  const auto *join = static_cast<const AsyncEvent::Join *>(data);
   uint64_t join_id = join->join_id;
   const auto *delim = bpftrace.resources.join_args[join_id].c_str();
   std::stringstream joined;
   for (unsigned int i = 0; i < bpftrace.join_argnum_; i++) {
-    auto *arg = join->content + (i * bpftrace.join_argsize_);
+    const auto *arg = join->content + (i * bpftrace.join_argsize_);
     if (arg[0] == 0)
       break;
     if (i)
@@ -35,7 +35,7 @@ void join_handler(BPFtrace &bpftrace, Output &out, void *data)
   out.message(MessageType::join, joined.str());
 }
 
-void time_handler(BPFtrace &bpftrace, Output &out, void *data)
+void time_handler(BPFtrace &bpftrace, Output &out, const void *data)
 {
   // not respecting config_->get(ConfigKeyInt::max_strlen)
   char timestr[MAX_TIME_STR_LEN];
@@ -46,7 +46,7 @@ void time_handler(BPFtrace &bpftrace, Output &out, void *data)
     LOG(WARNING) << "localtime_r: " << strerror(errno);
     return;
   }
-  auto *time = static_cast<AsyncEvent::Time *>(data);
+  const auto *time = static_cast<const AsyncEvent::Time *>(data);
   const auto *fmt = bpftrace.resources.time_args[time->time_id].c_str();
   if (strftime(timestr, sizeof(timestr), fmt, &tmp) == 0) {
     LOG(WARNING) << "strftime returned 0";
@@ -55,18 +55,18 @@ void time_handler(BPFtrace &bpftrace, Output &out, void *data)
   out.message(MessageType::time, timestr, false);
 }
 
-void helper_error_handler(BPFtrace &bpftrace, Output &out, void *data)
+void helper_error_handler(BPFtrace &bpftrace, Output &out, const void *data)
 {
-  auto *helper_error = static_cast<AsyncEvent::HelperError *>(data);
+  const auto *helper_error = static_cast<const AsyncEvent::HelperError *>(data);
   auto error_id = helper_error->error_id;
-  auto return_value = helper_error->return_value;
-  auto &info = bpftrace.resources.helper_error_info[error_id];
+  const auto return_value = helper_error->return_value;
+  const auto &info = bpftrace.resources.helper_error_info[error_id];
   out.helper_error(return_value, info);
 }
 
-void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data)
+void print_non_map_handler(BPFtrace &bpftrace, Output &out, const void *data)
 {
-  auto *print = static_cast<AsyncEvent::PrintNonMap *>(data);
+  const auto *print = static_cast<const AsyncEvent::PrintNonMap *>(data);
   const SizedType &ty = bpftrace.resources.non_map_print_args.at(
       print->print_id);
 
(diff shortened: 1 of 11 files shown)


Phoenix500526:
> In this PR, I modified bytecode_ in bpftrace from value semantics to pointer semantics, which caused some runtime test cases in CI to fail. The log output is pasted below:
> 
> ```shell
>  [ RUN      ] aot.call.ustack_stack_mode_env_bpftrace
>   [  FAILED  ] aot.call.ustack_stack_mode_env_bpftrace (1743 ms)
>   	Command: /home/runner/work/bpftrace/bpftrace/build-ci/src/bpftrace  -e 'u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }' --aot /tmp/tmpprog.btaot && /tmp/tmpprog.btaot
>   	Expected REGEX: ^\s+uprobeFunction1\+[0-9]+@[\w\/]+.c:[0-9]+$
>   	Found:
>   WARNING: Using development shim ("/home/runner/work/bpftrace/bpftrace/build-ci/src/aot/bpftrace-aotrt"). Ensure you are a developer!\nAttaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\n\n        0\n\n\n\n
>   	After cmd output: 
>   [ RUN      ] aot.call.ustack_stack_mode_env_override
>   [       OK ] aot.call.ustack_stack_mode_env_override (1885 ms)
>   [ RUN      ] aot.call.ustack_stack_mode_env_perf
>   [  FAILED  ] aot.call.ustack_stack_mode_env_perf (1737 ms)
>   	Command: /home/runner/work/bpftrace/bpftrace/build-ci/src/bpftrace  -e 'u:./testprogs/uprobe_loop:uprobeFunction1 { printf("%s\n", ustack(1)); exit(); }' --aot /tmp/tmpprog.btaot && /tmp/tmpprog.btaot
>   	Expected REGEX: ^\s+[0-9a-f]+ uprobeFunction1\+[0-9]+ \(.*/uprobe_loop\)$
>   	Found:
>   WARNING: Using development shim ("/home/runner/work/bpftrace/bpftrace/build-ci/src/aot/bpftrace-aotrt"). Ensure you are a developer!\nAttaching 1 probe...\n__BPFTRACE_NOTIFY_PROBES_ATTACHED\n\n\t0 0 ([unknown])\n\n\n\n
>   	After cmd output: 
>   [ RUN      ] aot.call.ustack_stack_mode_env_raw
>   [       OK ] aot.call.ustack_stack_mode_env_raw (1726 ms)
> ```
> 
> I'm unsure how to solve this problem, and concerned that fixes might introduce breaking changes. What do you think? @danobi @jordalgo @amscanne

I think that the commit `1aba9592d21a76e1f4a177f2e8522cb971e10b46` may be the culprit. In this commit, I moved some map relevant operations from Bpftrace to BpfMap. It's pretty simple and straightforward.

src/async_action.cpp
```diff
@@ -77,9 +77,45 @@ void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data)
...
+    LOG(BUG) << "Could not zero map with ident \"" << map.name()
+             << "\", err=" << std::to_string(err);
+}
+
+void clear_map_handler(BPFtrace &bpftrace, const void *data)
+{
+  const auto *mapevent = static_cast<const AsyncEvent::MapEvent *>(data);
+  const auto &map = bpftrace.bytecode_.getMap(mapevent->mapid);
+
+  auto err = bpftrace.clear_map(map);
```
jordalgo, line 108: From what I could tell of `clear_map` and `zero_map` (and perhaps `print_map`) they don't need to live in the `bpftrace` class. Maybe they'd be better suited in `bpfmap` instead. Especially since you moved most of the logic there anyway e.g. `zero_out`. WDYT?

jordalgo (commented):
I still need to spend more time with this PR but left a few comments/thoughts.

src/util/exceptions.h
```diff
@@ -30,4 +30,11 @@ class FatalUserException : public std::runtime_error {
   using std::runtime_error::runtime_error;
 };
 
+class BpfMapElemException : public std::runtime_error {
```
amscanne, line 33: I know that there are a lot that still exist in the codebase, but I'm on a personal mission to burn down exceptions. This isn't a hard ask, but hopefully this comment can sell you a bit on this approach :)

Some of this is covered here: https://github.com/bpftrace/bpftrace/blob/master/docs/coding_guidelines.md#error-handling-and-exceptions

Basically exceptions are silent in the API, and make it much harder to test and reason about.

When an error is possible, I've added a basic `Result` wrapper, and am trying to slowly push green field cases (and some updated cases) to use this. You see this here: https://github.com/bpftrace/bpftrace/blob/master/src/util/result.h#L17

These types can wrap a value, and be easily propagated through a call chain. In this case, it would be something like:

bpfmap.h
```
class BpfMapError : public ErrorInfo<BpfMapError> {
public:
  static char ID;
  void log(llvm::raw_ostream &OS) const override {
    OS << "BPF map error... <other information in the object as needed>";
  }
};

class BpfMap {
   Result<> some_function();
   Result<SomeType> other_function();
};
```

bpfmap.cpp
```
char BpfMapError::ID;

Result<> BpfMap::some_function() {
   auto st = other_function();
   if (!st) {
      return st.takeError();
   }
   // Do something with *st
   return OK();
}

Result<SomeType> BpfMap::other_function() {
   ...
   if (err) {
      return make_error<BpfMapError>(...);
   }
   return SomeType();
}
```

amscanne: (I see that @jordalgo also suggested this above, sorry for the noise! Hopefully the examples there are helpful, if you go this route.)

src/bpfmap.cpp
```diff
@@ -71,6 +64,138 @@ bool BpfMap::is_printable() const
...
+void BpfMap::lookup_elem(const void *key, void *value) const
+{
+  auto err = bpf_map_lookup_elem(fd(), key, value);
+  if (err != 0)
+    throw util::BpfMapElemException(
+        "failed to lookup elem: " + std::to_string(err) + " (" +
+        std::strerror(-err) + ")");
+}
+
+KVPairVec BpfMap::collect_kvs(int nvalues) const
```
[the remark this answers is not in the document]
Phoenix500526: Done

src/bpftrace.cpp
```diff
@@ -1146,28 +1124,13 @@ int BPFtrace::clear_map(const BpfMap &map)
...
-  std::vector<uint8_t> zero(value_size, 0);
-  for (auto &k : keys) {
-    int err = bpf_update_elem(map.fd(), k.data(), zero.data(), BPF_EXIST);
-
-    if (err && err != -ENOENT) {
-      LOG(ERROR) << "failed to look up elem: " << err;
-      return -1;
-    }
+  auto keys = map.collect_keys();
+  try {
```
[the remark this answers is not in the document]
Phoenix500526: Done

src/async_action.cpp
Phoenix500526: I've moved `clear_map` and `zero_map` from the `bpftrace` class to the `BpfMap` class, but kept `print_map` in its original location. This decision was made after careful consideration of the different responsibilities of these functions.

The `print_map` function differs from `clear_map` and `zero_map` in two key aspects:

1. **Dependency on Output Class**: `print_map` relies on the `Output` class for formatting and displaying map contents. Moving it to `BpfMap` would create an unnecessary coupling between these two classes, which could make the code less maintainable and more difficult to test.
2. **Extended Functionality**: `print_map` encompasses more than just map operations - it includes sorting, dividing, and formatting logic. These are presentation concerns rather than core map functionality. Keeping these responsibilities separate from `BpfMap` better adheres to the single responsibility principle.

src/util/exceptions.h
Phoenix500526: Done

src/bpftrace.cpp
```diff
@@ -774,10 +753,12 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
...
-
-    if (bpf_update_elem(
-            bytecode_.getMap(MapType::Elapsed).fd(), &key, &nsec, 0) < 0) {
-      perror("Failed to write start time to elapsed map");
+    auto map = bytecode_.getMap(MapType::Elapsed);
+    auto ok = map.update_elem(&key, &nsec);
+    if (!ok) {
+      auto error = ok.takeError();
+      LOG(ERROR) << "Failed to write start time to elapsed map: "
+                 << toString(std::move(error));
```
amscanne, line 761: You can just use ok.takeError() here directly, util/result.h defines this operator<<. (Here and elsewhere.) That saves the line above, the std::move and the toString().

src/bpfmap.h
```diff
@@ -5,16 +5,38 @@
...
+  static char ID;
+  std::string message;
+
+  BpfMapError(std::string msg) : message(std::move(msg))
+  {
+  }
+
+  void log(llvm::raw_ostream &OS) const override
+  {
+    OS << "BPF map error... <" << message << ">";
```
amscanne, line 28: Optional: my "..." was just an error :) you can frame this message how you think is reasonable/correct.

src/async_action.cpp
```diff
@@ -77,9 +78,48 @@ void print_non_map_handler(BPFtrace &bpftrace, Output &out, void *data)
...
+             << "\", err=" << std::to_string(err);
+}
+
+void zero_map_handler(BPFtrace &bpftrace, const void *data)
+{
+  const auto *mapevent = static_cast<const AsyncEvent::MapEvent *>(data);
+  const auto &map = bpftrace.bytecode_.getMap(mapevent->mapid);
+  uint64_t nvalues = map.is_per_cpu_type() ? bpftrace.ncpus_ : 1;
+
+  [[maybe_unused]] auto ok = handleErrors(
```
amscanne, line 99: Minor: I believe that this may cause an assertion failure if the error is unhandled.

Honestly, since you're not really using the error, I wouldn't bother with the handleErrors. I would just do:

```
auto ok = map.zero_out(nvalues);
if(!ok) {
  LOG(BUG) << ... << ok.takeError();
}
```

This is a pretty simple and it's a common pattern, unless you specifically want to extract some data and handle that case.

(Applies here and related functions.)

src/bpfmap.cpp
```diff
@@ -60,17 +54,155 @@ bool BpfMap::is_per_cpu_type() const
...
+                           static_cast<size_t>(nvalues));
+    int err = bpf_map_lookup_elem(fd(), key.data(), value.data());
+    if (err == -ENOENT) {
+      // key was removed by the eBPF program during bpf_map_get_next_key() and
+      // bpf_map_lookup_elem(), let's skip this key
+      continue;
+    } else if (err) {
+      return make_error<BpfMapError>(
+          "failed to look up elem: " + std::to_string(err) + " (" +
+          std::strerror(-err) + ")");
```
amscanne, line 151: Just for context (feel free to do this or not) for future errors, you could just put the structure in the info class and format once in the log() function.

For example, the class could have "std::string op", and "int err" and you could construct `make_error<BpfMapError>("lookup", -err)`. It's just pushing the structure into the types. Makes it so there's less boilerplate formatting code around, etc.

amscanne (approved):
Just some minor stuff, overall seems reasonable to me (but you'll still need stamp from @jordalgo or someone else)

BTW I had missed the earlier question re: Bytecode -- sorry about that, but it seems you found your way around that?

src/bpfmap.h
Phoenix500526: Done

src/bpftrace.cpp
Phoenix500526: Done

src/async_action.cpp
Phoenix500526: Done

src/bpfmap.cpp
```diff
@@ -60,17 +54,143 @@ bool BpfMap::is_per_cpu_type() const
...
+  // snapshot keys, then operate on them
+  KeyVec keys;
+  while (bpf_map_get_next_key(fd(), old_key, key.data()) == 0) {
+    keys.push_back(key);
+    old_key = key.data();
+  }
+  return keys;
+}
+
+Result<OK> BpfMap::zero_out(int nvalues) const
```
jordalgo, line 77: nit
```suggestion
Result<> BpfMap::zero_out(int nvalues) const
```
As was just pointed out to me, the OK is the default so you don't need to type it out explicitly.

jordalgo: Same for the ones below.

jordalgo:
Just one small nit otherwise this is good to go. Nice work!

src/bpfmap.cpp
Phoenix500526: Done

jordalgo (approved):

PR #4158, opened 2025-05-20
codegen: allow filtering on current cgroup

amscanne:
codegen: allow filtering on current cgroup

This is a fairly minimal version, and could be made much better simply
by plumbing the cgroup ancestor level through to the `__in_cgroup` call.


src/ast/passes/filter_pass.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  switch (type) {
+    case ProbeType::kprobe:
+    case ProbeType::kretprobe:
+    case ProbeType::fentry:
+    case ProbeType::fexit:
+    case ProbeType::tracepoint:
+    case ProbeType::rawtracepoint:
+      return true;
+    // These probe types support passing the pid during attachment
```
viktormalik, line 41: Is this also true for cgroup filtering? Shouldn't we use a different `cgroup_needs_filter` function?

PR #4164, opened 2025-05-20
ci: adding `gtest-parallel` as a nix flake, using for ci.py

amscanne:
ci: adding `gtest-parallel` as a nix flake, using for ci.py

Let's see how much this speeds up the `bpftrace_test` portion of tests.


jordalgo (approved):
🚤

PR #4168, opened 2025-05-21
Fix WILL_FAIL directive

danobi:
See commits for more details

##### Checklist

### DIFF
--- a/tests/runtime/engine/runner.py
+++ b/tests/runtime/engine/runner.py
@@ -519,13 +519,16 @@ class Runner(object):
             print(warn("[   SKIP   ] ") + label)
             return Runner.SKIP_AOT_NOT_SUPPORTED
 
-        if p and p.returncode != test.return_code and not test.will_fail and not timeout:
-            print(f"{fail("[  FAILED  ] ")} {label}")
-            print(f"\tCommand: {bpf_call}")
-            print(f"\tUnclean exit code: {p.returncode}")
-            print(f"\tOutput: {to_utf8(output)}")
-            print_befores_and_after_output()
-            return Runner.FAIL
+        if p:
+            return_code_good = p.returncode == test.return_code
+            will_succeed = not test.will_fail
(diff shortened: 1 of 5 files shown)


tests/runtime/engine/runner.py
```diff
@@ -519,13 +519,16 @@ def print_befores_and_after_output():
...
             return Runner.SKIP_AOT_NOT_SUPPORTED
 
-        if p and p.returncode != test.return_code and not test.will_fail and not timeout:
-            print(fail("[  FAILED  ] ") + label)
-            print('\tCommand: ' + bpf_call)
-            print('\tUnclean exit code: ' + str(p.returncode))
-            print('\tOutput: ' + to_utf8(output))
-            print_befores_and_after_output()
-            return Runner.FAIL
+        if p:
```
jordalgo, line 522: I feel like it's less confusing to just disallow the use of the `RETURN_CODE` directive and the `WILL_FAIL` directive e.g. here is our one test that uses `RETURN_CODE`

```
NAME exit code
PROG BEGIN { exit(69); }
RETURN_CODE 69
TIMEOUT 1
```
If we add `WILL_FAIL`, it fails the test and gets this error:
```
Unexpected exit code: 69 (expected != 69)
```

tests/runtime/selftest
```diff
@@ -5,3 +5,13 @@ NAME command doesn't respond to SIGTERM
 RUN trap '' TERM; echo __BPFTRACE_NOTIFY_PROBES_ATTACHED && sleep 999 && echo "shouldn't echo"
 TIMEOUT 1
 EXPECT_NONE shouldn't echo
+
+NAME WILL_FAIL allows failure
```
jordalgo, line 9: I'm not terribly keen on adding runtime tests that test the runtime framework.

tests/runtime/engine/runner.py
danobi: Yeah makes sense. Let's keep it NAND

PR #4170, opened 2025-05-21
types: drop `is_internal`

amscanne:
types: drop `is_internal`

This can use the address space, which seems to be mostly a no-op
replacement. The only exception is that some codegen tests now have
an extra `probe_read` in certain places. But this appears to be the
intended behavior, according to the semantic analyzer, which
intentionally neuters the addrspace for any pointers:

```
  // BPF verifier cannot track BTF information for double pointers so we
  // cannot propagate is_btftype for arrays of pointers and we need to reset
  // it on the array type as well. Indexing a pointer as an array also can't
  // be verified, so the same applies there.
```

This change brings the test inline with the code and documented
behavior, and we can subsequently remove this hack if BTF does allow us
to track the type effectively.


ajor:
Re the "BPF verifier cannot track BTF information for double pointers" comment - it looks like that was originally talking about `is_btftype` and not `is_internal`. This only got changed in #3996 - I couldn't work out how to link to a specific line of a PR, so here's a screenshot instead:

image

What was the reason this was changed in that PR? If we don't override the addr space here can we get rid of those extra probe read calls?

amscanne:
This will be removed in #4891.

PR #4171, opened 2025-05-21
types: drop `is_btftype`

amscanne:
types: drop `is_btftype`

This can also be replaced by the `AddrSpace::bpf` marker.


amscanne:
This was removed in #4285.

Issue #4162, opened 2025-05-20
SEGV When an error is encountered during probe listing


ajor:
Repro on lastest master:
```
# bpftrace -l 'fentry:'
Segmentation fault
```

From a quick debugging session, I think this is caused by some lifetime problems with `ASTContext`.

In `buildListProgram`, a local `ASTContext` is created and `make_node` is called on it. The newly created nodes will store a reference to this local `ASTContext` variable. When the function returns, this local variable's address is no longer valid, and `node.ctx_` will point to this invalid address:
https://github.com/bpftrace/bpftrace/blob/5b075d2a8a32bdcc2420a74c40cbdddcd2a1cdaf/src/main.cpp#L662-L663

ajor:
cc @amscanne as this seems to have originated in the series of patches #3855 & #3991

jordalgo:
Yup, I noticed this too and just forgot to make an issue 👎

amscanne:
Whoops, let me kick up a fix.

PR #4172, opened 2025-05-21
test: fix regression test for #4162

amscanne:
Update test for #4162 and ensure description is accurate.


jordalgo:
@amscanne Making commits from the github UI? DCO failing :)

jordalgo:
Also the regression test is failing lol. I think it needs to be 'fentry:*'

PR #4173, opened 2025-05-21
Refactor AttachedProbe class

jordalgo:
Create separate derived classes for each probe type.
This allows us to removing forking logic from the
AttachedProbe destructor and makes it very explicit
what member variables are required for each probe type.

This removes a lot of bloat from the AttachedProbe.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -14,6 +14,7 @@
 #include <sys/syscall.h>
 #include <sys/utsname.h>
 #include <unistd.h>
+#include <utility>
 
 #include <bcc/bcc_elf.h>
 #include <bcc/bcc_syms.h>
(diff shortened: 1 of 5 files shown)


jordalgo:
Also the regression test is failing lol. I think it needs to be 'fentry:*'

jordalgo:
Most of the logic in the `attach` functions is unchanged except that they no longer reference member variables and they return this `cleanup` function that references the fds they open.

src/bpftrace.cpp
```diff
@@ -682,7 +682,7 @@ int BPFtrace::run_iter()
   }
 
   auto &ap = *attached_probes_.begin();
-  int link_fd = ap->linkfd_;
+  int link_fd = static_cast<AttachedIterProbe *>(ap.get())->iter_link_fd;
```
danobi, line 685: Safer to dynamic_cast here? I think if there's a bug static_cast is UB but dynamic_cast will null ptr exception. cc @amscanne who probably knows the right thing here

src/attached_probe.cpp
```diff
@@ -982,267 +1113,172 @@ Result<> AttachedProbe::attach_usdt(BPFfeature &feature)
...
-                                              period,
-                                              freq,
-                                              pid_.has_value() ? *pid_ : -1,
-                                              cpu,
-                                              group_fd);
-
-    if (perf_event_fd < 0) {
-      return make_error<AttachError>();
+  LOG(V1) << "Trying to attach probe: " << probe.orig_name;
+  int progfd = prog.fd();
```
danobi, line 1157: Can you pls propagate down `prog` into the attach_* functions? We'll need it to do libbpf attachment

src/attached_probe.h
```diff
@@ -40,84 +40,42 @@ class AttachedProbe {
...
-
-  std::vector<int> perf_event_fds_;
-  uint64_t offset_ = 0;
-  int tracing_fd_ = -1;
-  std::function<void()> usdt_destructor_;
-  USDTHelper usdt_helper;
+  AttachedWatchpointProbe(int progfd,
+                          std::function<void()> cleanup,
+                          uint64_t probe_addr);
+  friend class AttachedProbe;
```
danobi, line 78: I'm not a fan of the use of `friend` here - to me it indicates the class hierarchy is not properly designed. As an alternative perhaps consider `protected` members? But see my overall comment as well

src/attached_probe.h
```diff
@@ -40,84 +40,42 @@ class AttachedProbe {
...
+  const std::function<void()> cleanup_;
+  friend class AttachedIterProbe;
+  friend class AttachedWatchpointProbe;
+};
+
+class AttachedIterProbe : public AttachedProbe {
+public:
+  ~AttachedIterProbe() override = default;
+
+  const int iter_link_fd;
```
danobi, line 59: For the kprobe nmissed work, we'll need to access the link fd. This is generally applicable to all probe types, so exposing a getter for link fd is reasonable and useful. But to do that with your proposed design/refactor, we'll need to create classes for all probe types, no? So see overall comment.

danobi (commented):
Overall I'm not convinced this approach is cleaner than having separate subclasses for each probe type. I understand separate subclasses has more boiler plate, but it's not like we're adding new probe types all the time. Rather, over time I expect each probe type to more subtly different from the others, with the example of needing to pull nmissed stats from kprobe link. So having the separate subclasses is an opportunity to cleanly hold separate logic. Right now this PR has the different probe types smushed together into one cleanup_ callback with opportunities for extension limited, and requiring a criss-crossing of `friend` declarations.

src/attached_probe.h
```diff
@@ -40,84 +40,208 @@ class AttachedProbe {
...
+class AttachedWatchpointProbe : public AttachedProbe {
+public:
+  static Result<std::unique_ptr<AttachedWatchpointProbe>> make(
+      Probe &probe,
+      const BpfProgram &prog,
+      std::optional<int> pid,
+      const std::string &mode);
+  ~AttachedWatchpointProbe() override;
+
+  const uint64_t probe_addr;
```
danobi, line 238: Instead of divergent subclasses, how about wiring this through a virtual getter on the base class? On the base class, you could supply a default EOPNOTSUPP (some equivalent, of course) implementation. Then relevant subclasses can override. That way callers don't need to dynamic cast and we can hide all these class definitions in attached_probe.cpp and only expose `class AttachedProbe` in `attached_probe.h`.

Also, it might make more sense to just add a `get_probe()` that returns `const Probe&`?

src/attached_probe.h
```diff
@@ -40,84 +40,208 @@ class AttachedProbe {
...
+};
+
+class AttachedIterProbe : public AttachedProbe {
+public:
+  static Result<std::unique_ptr<AttachedIterProbe>> make(
+      const BpfProgram &prog,
+      std::optional<int> pid);
+  ~AttachedIterProbe() override;
+
+  const int iter_link_fd;
```
danobi, line 223: Similar to AttachedWatchpoingProbe, this could be moved to something like `get_link_fd()` on the base class. This gets quite clean for my follow-on work (as well as @Phoenix500526's), as we can slowly implement `get_link_fd()` on subclasses that are migrated to libbpf attachment.

danobi (commented):
ty for iterating!

src/attached_probe.h
jordalgo: > Also, it might make more sense to just add a get_probe() that returns const Probe&?

I think I prefer this. Will update.

jordalgo:
@danobi Made the suggested updates.

danobi:
Move the subclass definitions into attached_probe.cpp?

PR #4174, opened 2025-05-21
Add code of conduct

jordalgo:
This is the same one used in the linux kernel.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8a104f8b5867c682d994ffa7a74093c54469c11f

##### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ bpftrace is a high-level tracing language for Linux. bpftrace uses LLVM as a bac
 - [Probe types](#probe-types)
 - [Plugins](#plugins)
 - [License](#license)
+- [Code of Conduct](docs/code_of_conduct.md)
 
 ## Example One-Liners
 
--- a/docs/code_of_conduct.md
+++ b/docs/code_of_conduct.md
@@ -1,0 +1,132 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the overall
+  community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email address,
+  without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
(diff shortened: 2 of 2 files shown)


jordalgo:
Before this gets merged we need an email accessible to all the maintainers where people can report conduct violations. Or we can just have people make sure to email all the maintainers to ensure visibilty

docs/code_of_conduct.md
```diff
@@ -0,0 +1,132 @@
...
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official email address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at BPFTRACE_EMAIL.
```
amscanne, line 63: I think you should just create a Google group or mailing list that anyone can send mail to, but only maintainers will be on (as part of the governance model). This is also where maintainers would discuss and vote on maintainer-only issues. It would exist forever as a record of the decisions made, etc.

This would also be a suitable place to report code of conduct violations. You could also note that any maintainer can be emailed individually if the reporter doesn't feel comfortable sending to that list. (And acting on code of conduct violations would be part of the expectations for a maintainer)

README.md
```diff
@@ -25,6 +25,7 @@ bpftrace is a high-level tracing language for Linux. bpftrace uses LLVM as a bac
 - [Probe types](#probe-types)
 - [Plugins](#plugins)
 - [License](#license)
+- [Code of Conduct](docs/code_of_conduct.md)
```
amscanne, line 28: Why is the code of conduct in `docs` but the contributing guide is in the base directory? Both are acceptable, but I'm wondering why they are different: https://github.com/joelparkerhenderson/github-special-files-and-paths?tab=readme-ov-file#contributing

I would personally recommend just putting all the "special" files in the root, but I think it should be consistent at least.

ajor (changes requested):
A code of conduct seems unnecessary to me. What problem is it trying to solve?

I don't see a need for a formal set of rules on a small project like bpftrace which has never had any of the issues described.

amscanne:
It is a standard practice now, and is primarily a signaling mechanism that the maintainers care about the community and will take complaints seriously. I'm not sure which formal rules are being objected to, but the contributor convenant is pretty in-objectionable as far as these things go.

Is there a specific reason "why not"? The code of conduct is best added long before any issues arise, not in response to them and does carry some amount of signaling power.

ajor:
Codes of conduct are used in lots of projects, but there are also plenty of projects larger than us which don't use them. [A search on Hacker News](https://hn.algolia.com/?dateRange=all&page=0&prefix=true&query=code%20of%20conduct&sort=byPopularity&type=story) shows that codes of conduct are far from being universally accepted.

My personal objections are:

### Bureaucracy
- It adds unnecessary bureaucracy, formal processes and duties for maintainers that take time away from doing real project work.
- An over-focus on bureaucracy can take away from the pleasure of writing code for fun.
- A new communication channel (email or google group) would have to be regularly checked, and I can picture it getting forgotten about if it goes unused for years.

### Wording
The wording is not precise enough to be a proper set of rules, and two-thirds of the document is about enforcement. I have a few concerns about this:
- The vague wording about what's unacceptable leaves a lot open to interpretation, while still providing enough detail to point at as evidence of a rule violation. This is the kind of thing that I worry could lead to selective enforcement of the rules while giving a veneer of legitimacy.
  One clause I feel is unclear, for example, is: "conduct which could reasonably be considered inappropriate in a professional setting". What professional setting? In what country? There isn't a fixed worldwide "professional setting". I feel like this code of conduct was written from a corporate American point of view.
- A focus on enforcement guidelines could encourage heavy-handed enforcement of the code of conduct, where simply telling someone to be nicer would be enough. In particular, I don't like the clause setting out the possibility of forced public apologies.

### Culture
This document feels like an encroachment of corporate culture into a non-corporate project. When I'm coding for free in my spare time, I don't want to deal with more rules. Simply put: I don't like being told what to do :)

-----

In order to get behind a code of conduct, I'd like to see an argument for adopting one in bpftrace specifically. For example, what are the risks of not having a code of conduct?
- How likely is it that we will encounter some bad behaviour in this project in the next couple of years?
- What would the impact to bpftrace be if some bad behaviour were to occur and we did not have a formal code of conduct in place?

In particular, is the risk of not having a code of conduct large enough to override other concerns and be worth the time required to draft and implement it?

PR #4175, opened 2025-05-22
FindLibOpcodes.cmake: Fix typos

Rtoax:
libbfd should be libopcodes.

Fix: commit 542f2b90f076 ("Add support to check for libbfd/libopcodes libraries")


jordalgo (approved):
Yup, looks like some bad copy-pasta. Good catch.

PR #4179, opened 2025-05-22
Attach watchpoint probe to threads

### LINKED ISSUE #826: Memory watchpoint does not work with threaded targets
Looks like `perf_event_open` does not work if applications use threads. It seems like only the main TID (ie the PID) is traced:

```
$ sudo ./build/src/bpftrace ~/watchpoint.bt -c "/home/dlxu/scratch/bpftrace/mmap_addr/a.out 'pthread on'"
Attaching 1 probe...
Mapped 2097152 bytes at 268435456 (0x10000000)
^C

$  ~/dev/bpftrace git:(master) sudo ./build/src/bpftrace ~/watchpoint.bt -c "/home/dlxu/scratch/bpftrace/mmap_addr/a.out"
Attaching 1 probe...
Mapped 2097152 bytes at 268435456 (0x10000000)
hit!
hit!
hit!
hit!
hit!
hit!
^C

$ cat /home/dlxu/scratch/bpftrace/main.cpp
at /home/dlxu/scratch/bpftrace/mmap_addr/main.cpp
#include <cstdint>
#include <stdio.h>
#include <sys/mman.h>
#include <thread>
#include <unistd.h>
#include <pthread.h>

using namespace std::chrono_literals;

void* work(void* arg) {
  void* addr = ::mmap(
      reinterpret_cast<void*>(0x10000000),
      2 << 20,
      PROT_READ | PROT_WRITE,
      MAP_PRIVATE | MAP_ANONYMOUS,
      -1,
      0);

  if (addr < 0) {
    ::perror("mmap");
    return nullptr;
  }

  printf("Mapped %d bytes at %d (%p)\n", 2 << 20, addr, addr);

  uint8_t count = 0;
  while (count < 20) {
    *reinterpret_cast<uint8_t*>(addr) = count++;
    std::this_thread::sleep_for(1s);
  }

  return nullptr;
}

int main(int argc, char** argv) {
  if (argc > 1) {
    pthread_t tid;
    if (::pthread_create(&tid, nullptr, work, 0)) {
      ::perror("pthread_create");
    }

    if (::pthread_join(tid, nullptr)) {
      ::perror("pthread_join");
    }
  } else {
    work(nullptr);
  }

  ::printf("Exiting...\n");
  return 0;
}
```

Will possibly require a kernel change. Looking into it with @yonghong-song

danobi:
Previously, watchpoints only worked on the thread group leader. Threads were not considered.

This commit makes the situation better but does not fully solve it. It appears there's a limitation with perf_event_open() where only new threads can be attached to. See inline comments for more details.

This closes #826.

##### Checklist


jordalgo:
@danobi I had tried fixing this last week but ran into an issue wit this example program:
```
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

__attribute__((noinline)) void increment(__attribute__((unused)) int _, int *i)
{
  (*i)++;
}

__attribute__((noinline)) void *work()
{
  int *i = malloc(sizeof(int));
  while (1) {
    increment(0, i);
    (*(int *)i)++;
    usleep(10000);
  }
  pthread_exit(NULL);
}

int main()
{
  usleep(10000000);

  pthread_t tid;
  if (pthread_create(&tid, NULL, work, NULL)) {
    perror("pthread_create");
  }

  if (pthread_join(tid, NULL)) {
    perror("pthread_join");
  }

  return 0;
}
```

And here was the runtime test I couldn't get to pass:
```
NAME jordan
RUN {{BPFTRACE}} -e 'watchpoint:increment+arg1:4:w { printf("hit!\n"); exit() }' -c ./testprogs/watchpoint_threaded
EXPECT hit!
ARCH aarch64|x86_64
REQUIRES_FEATURE signal
```
It was able to attach but I never got the `hit`. Do you know what I'm doing wrong?

danobi:
I can repro but didn't get a chance to debug. I wonder if it's related to where the setup probe fires

danobi:
@jordalgo what do you wanna do with this? Should we land and then track the issue above with a new issue?

jordalgo:
@danobi I'm on the fence. I would like to debug a little more before declaring this fixed. If you think it would still be helpful to land this as is, then that's fine with me. Adding `inherit = 1` is probably good regardless.

danobi:
I think this is probably strictly helpful. Not sure I have enough time left to finish debugging this any time soon

jordalgo:
@danobi kk - just fix the merge conflict and we can get this in. Let's keep issue 826 open and make a note of the problem.

ajor:
This PR is definitely an improvement and is worth getting in even if it's not a full fix.

Just a quick point though, in case it's relevant, is this code not likely to be an issue:
```
    int perf_event_fd = syscall(__NR_perf_event_open,
                                &attr,
                                pid_.has_value() ? *pid_ : -1,   <<<<<< this line
                                cpu,
                                -1,
                                PERF_FLAG_FD_CLOEXEC);
```

When we're attaching our watchpoint to a given PID we pass it over to perf_events. Is perf_events treating this passed in PID as a kernel "pid", i.e. a thread ID rather than a process ID (tgid). That would explain it only catching the thread group parent at launch.

jordalgo:
Ok, I figured out why my example doesn't work and @danobi was right, it had to do with the setup probe. Basically the `increment` function never fires until the new thread is created and calls it. This initial call is what triggers the setup probe (which is a uprobe) to fire and ultimately attach the actual watchpoint probe. At this point the child thread is already created, which means the watchpoint probe won't count events for it. As per the 'inherit' docs: "This applies only to new children, not to any existing children at the time the counter is created (nor to any new children of existing children)".

If I change the example to call `increment` once before the new thread is created then it works as expected.

This is interesting in that my example (as is) should work but because of how we have the setup probe trigger the watchpoint probe it does not. I'll try to think a bit on how to solve this (any suggestions are welcome).

man/adoc/bpftrace.adoc
```diff
@@ -1845,6 +1845,9 @@ int main()
 # bpftrace -e 'watchpoint:increment+arg1:4:w { printf("hit!\n"); exit() }' -c ./wpfunc
 ----
 
+Note that threads are monitored, but only for threads created after watchpoint attachment.
+The is a limitation from the kernel.
```
amscanne, line 1849: Nit: if there's anything that can be linked here for more context (an issue, source control, anything) that would be helpful if someone hits that limitation and just wants to understand more.

jordalgo:
Closing in favor of this, which preserved the original commit:
https://github.com/bpftrace/bpftrace/pull/4239

Issue #3916, opened 2025-03-19
Make ringbuf loss counter use global variable


danobi:
The ringbuf loss counter currently uses an array map. Arraymaps (on all but the most bleeding edge kernels) require null checks, and thus the possibility of a map lookup failure.

It's more reliable to use global variables, which do not need null checks. It'll also simplify this codepath, both for us and users at runtime.

## Quote 
             It's the key value and value value (heh) for the map it writes into for the loss counter.

The impl should probably be refactored to use global vars instead of an arraymap.

_Originally posted by @danobi in https://github.com/bpftrace/bpftrace/pull/3914#discussion_r2002187547_

Phoenix500526:
I'm interested in this issue. Plz assign this issue to me.

jordalgo:
@Phoenix500526 Done. Thanks for picking this up!

Phoenix500526:
Hi, @jordalgo @danobi . I got a question. In the current implementation, `event_loss_counter` is a BPF array map. The BPF program updates this map during execution, and `bpftrace` then reads the updated value from it. 
If we change `event_loss_counter` from a BPF map to a global variable, how would the BPF program and `bpftrace` interact each other using this global variable?

Phoenix500526:
Hi @danobi @jordalgo 

PR #4181  is ready for an initial look. Most tests are green, but there's one CI check that's failing due to its dependency on the new implementation details.

To make the review process efficient, I wanted to get your input. Would you be open to reviewing the core refactoring first? Once we're aligned on the main approach, I'll update the affected test case. Or, if you prefer, I can address the test case now and then you can review the complete set of changes.

Looking forward to your feedback!

Phoenix500526:
Yes, only the codegen tests didn't pass. And I've already regenerated it and all tests are passed now.  @jordalgo

PR #4181, opened 2025-05-23
refactor: make ringbuf loss counter use global variable

Phoenix500526:
This pr replaces the `EventLossCounter` map with a global variable `EventLossCounter`. Please read the issue #3916 for more details. 

##### Checklist


src/ast/irbuilderbpf.cpp
```diff
@@ -1223,8 +1223,7 @@ void IRBuilderBPF::CreateCheckSetRecursion(const Location &loc,
   // Most of the time this will happen for the functions that can lead
   // to a crash e.g. "queued_spin_lock_slowpath" but it can also happen
   // for nested probes e.g. "page_fault_user" -> "print".
-  CreateAtomicIncCounter(to_string(MapType::EventLossCounter),
-                         bpftrace::BPFtrace::event_loss_cnt_key_);
+  CreateAtomicIncCounter();
```
jordalgo, line 1226: Seems like this function is no longer generic so perhaps this is a better name.
```suggestion
  CreateIncEventLossCounter();
```

src/globalvars.cpp
```diff
@@ -316,4 +319,55 @@ std::unordered_set<std::string> get_section_names()
   return ret;
 }
 
+uint64_t get_event_loss_counter(
```
jordalgo, line 322: Not for this PR but looks like we could make this generic by passing in the section name we care about to get the global var.

jordalgo (commented):
One small naming nit but otherwise this is good to go!

src/globalvars.cpp
Phoenix500526: Done

src/ast/irbuilderbpf.cpp
Phoenix500526: Done

src/globalvars.cpp
jordalgo: This function now is half-way to being generic as it's still named `get_event_loss_counter`. Maybe it should be something like `get_global_var`?

Also there are still local variables in this function named `event_loss_cnt`

jordalgo (approved):
Great work! Thanks for the contribution.

PR #4186, opened 2025-05-24
Add return value to delete

jordalgo:
Additionally, check if the return value
for delete is discarded. If it is discarded
issue a warning if there was an error deleting.
This allows users to handle deletion errors
and prevent expected warnings.

Issue: https://github.com/bpftrace/bpftrace/issues/4169

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1109,9 +1109,10 @@ void IRBuilderBPF::CreateMapUpdateElem(const std::string &map_ident,
   CreateHelperErrorCond(call, libbpf::BPF_FUNC_map_update_elem, loc);
 }
 
-void IRBuilderBPF::CreateMapDeleteElem(Map &map,
-                                       Value *key,
-                                       const Location &loc)
+CallInst *IRBuilderBPF::CreateMapDeleteElem(Map &map,
+                                            Value *key,
+                                            bool ret_val_discarded,
+                                            const Location &loc)
 {
   assert(key->getType()->isPointerTy());
   Value *map_ptr = GetMapVar(map.ident);
@@ -1127,7 +1128,10 @@ void IRBuilderBPF::CreateMapDeleteElem(Map &map,
       delete_func_ptr_type);
   CallInst *call = createCall(
       delete_func_type, delete_func, { map_ptr, key }, "delete_elem");
-  CreateHelperErrorCond(call, libbpf::BPF_FUNC_map_delete_elem, loc);
+  CreateHelperErrorCond(
+      call, libbpf::BPF_FUNC_map_delete_elem, loc, !ret_val_discarded);
+
+  return call;
 }
 
 Value *IRBuilderBPF::CreateForEachMapElem(Map &map,
@@ -2459,14 +2463,12 @@ void IRBuilderBPF::CreateHelperError(Value *return_value,
   CreateLifetimeEnd(buf);
 }
 
-// Report error if a return value < 0 (or return value == 0 if compare_zero is
-// true)
 void IRBuilderBPF::CreateHelperErrorCond(Value *return_value,
                                          libbpf::bpf_func_id func_id,
                                          const Location &loc,
-                                         bool compare_zero)
+                                         bool suppress_error)
 {
-  if (bpftrace_.helper_check_level_ == 0 ||
+  if (bpftrace_.helper_check_level_ == 0 || suppress_error ||
       (bpftrace_.helper_check_level_ == 1 && return_zero_if_err(func_id)))
     return;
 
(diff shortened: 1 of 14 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -2489,14 +2493,12 @@ void IRBuilderBPF::CreateHelperError(Value *return_value,
   CreateLifetimeEnd(buf);
 }
 
-// Report error if a return value < 0 (or return value == 0 if compare_zero is
-// true)
 void IRBuilderBPF::CreateHelperErrorCond(Value *return_value,
                                          libbpf::bpf_func_id func_id,
                                          const Location &loc,
-                                         bool compare_zero)
+                                         bool suppress_error)
```
danobi, line 2499: Rather than push suppression logic down into codegen, how about a new "unused return value" pass on AST? It'll

1. Get users in habit of checking return values
2. Avoid hyper-specific new flags on `Call` AST node
3. Keep codegen dumb

danobi: I meant to say that we should stop emitting warnings from codegen for this error case completely. And then just create a pass that, for an allowlist of calls with important return values, warns if the return value is unused. Maybe these warnings can be silenced with a cast to void. I think it's less complicated and points users towards a direction we want to be going anyways.

> What specific new flags for Call nodes are you worried about?
```
bool ret_val_discarded = false;
```

danobi: > This is a runtime warning not a compiler warning but maybe you're aware of that and saying we shouldn't issue a runtime warning even on failed deletes that don't handle the return value. If you are saying that then it's an interesting idea.

Right, I was suggesting delete the runtime error and make it strictly a compile time check.

> The part about that I don't love is just introducing even more warnings as all current scripts use delete as just an expression statement.

True, we are probably going to introduce a bunch of warnings to people's scripts. But like you said, perhaps it's important enough to do it? I'm not strictly opposed to the current approach in the PR - just surfacing a potential alternative.

jordalgo: I was going to introduce a `update` method for maps so users can handle the return value, perhaps for that one we issue a warning if the return value is ignored (as a start).

viktormalik: For the sake of terseness, I'd avoid printing a compile-time warning when the return value of `delete` is unused. For many scripts and one-liners, users will delete a key which always exists and they will want to use plain `delete(@map, X)`.

On top of that, I can see value in the runtime warning itself - many people will not realize that an update/delete may fail, would just suppress the compile-time warning, and then be surprised why the script is behaving strangely.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1082,10 +1082,14 @@ ScopedExpr CodegenLLVM::visit(Call &call)
                                                    call.loc);
       b_.CreateStore(Constant::getNullValue(b_.GetType(map.value_type)), val);
       b_.CreateMapUpdateElem(map.ident, scoped_key.value(), val, call.loc);
-      return ScopedExpr();
+      // Assume delete for these kinds of maps always succeeds.
```
viktormalik, line 1085: Are we sure with this?

src/ast/irbuilderbpf.cpp
```diff
@@ -2511,11 +2513,7 @@ void IRBuilderBPF::CreateHelperErrorCond(Value *return_value,
   // return int and we need to use Int32Ty value to check if a return
   // value is negative. TODO: use Int32Ty as a return type
```
viktormalik, line 2514: Is this TODO still relevant?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2930,7 +2939,11 @@ void SemanticAnalyser::visit(Tuple &tuple)
 
 void SemanticAnalyser::visit(ExprStatement &expr)
 {
-  visit(expr.expr);
+  if (auto *call = expr.expr.as<Call>()) {
+    visit_call(*call, true);
+  } else {
+    visit(expr.expr);
+  }
```
viktormalik, line 2946: Wouldn't it be easier to do
```suggestion
  if (auto *call = expr.expr.as<Call>()) {
    call->ret_val_discarded = true;
  }
  visit(expr.expr);
```
and no more modifications to the current code would be necessary?

Also a comment would be good as it took me some time to figure out why this is correct.

src/ast/passes/codegen_llvm.cpp
jordalgo: It's obviously not 100% but probably not something the user can do anything about. These scalar maps only have a single entry so no max-entries issues. Not sure what the other cases would be that would cause this to fail.

src/ast/irbuilderbpf.cpp
jordalgo: Yeah, we're still setting the ret for most helpers to Int64Ty

src/ast/passes/semantic_analyser.cpp
jordalgo: Ha yeah, that's better.

jordalgo: > Also a comment would be good as it took me some time to figure out why this is correct.

Done.

viktormalik (approved):
Nice!

PR #4189, opened 2025-05-27
Remove find_addrspace_stack

jordalgo:
##### Checklist

### DIFF
--- a/src/ast/codegen_helper.cpp
+++ b/src/ast/codegen_helper.cpp
@@ -21,7 +21,7 @@ bool needAssignMapStatementAllocation(const AssignMapStatement &assignment)
 {
   const auto &map = *assignment.map;
   const auto &expr_type = assignment.expr.type();
-  if (shouldBeInBpfMemoryAlready(expr_type)) {
+  if (expr_type.is_internal) {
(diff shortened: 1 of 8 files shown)


amscanne (commented):
Hm wait, I submitted the approval without fully processing. Why are all these types `AddrSpace::kernel`? They are bpftrace types and will generally be `AddrSpace::bpf` since the buffers will be natively addressable without probes.

I don't think this carries any risk because of the fact that this likely all handled in bespoke ways in codegen, but this may actually be adding to the address space confusion. I would try to get a few more deep before submitting, personally.

ajor:
As Adin says, I think a lot of these should be `AddrSpace::bpf` rather than `kernel`. `AddrSpace::bpf` is a relatively new addition bpftrace's codebase, so a lot of legacy code was using `AddrSpace::kernel` in combination with `is_internal = true` to indicate what can now be represented by `AddrSpace::bpf`.

Also, I'm shocked at how few tests needed changes! Is that because the code is doing more-or-less the same thing after your patch, or because all this address space stuff is untested? (I'm worried it's the latter 😅)

PR #4191, opened 2025-05-27
Add return value discarded warning

jordalgo:
Most functions that return a value should
issue a warning if that return value is not used.

Also a few drive-by fixes:
- Fixed percpu_kaddr ret type in adoc
- Added call_spec checks for `offsetof` and `sizeof`

##### Checklist


src/ast/passes/semantic_analyser.cpp
```diff
@@ -3703,6 +3738,11 @@ bool SemanticAnalyser::check_call(const Call &call)
     return true;
   }
 
+  if (is_final_pass() && call.ret_val_discarded &&
+      spec->second.discard_ret_warn) {
+    call.addWarning() << "Return value discarded";
```
amscanne, line 3743: Maybe a bit of exposition? I.e. explain a bit more about what the problem is:

`call.ident << " return value is discarded, it should be used";`

amscanne (approved):

PR #4194, opened 2025-05-27
Fix attaching probe count and change attaching language

jordalgo:
Two commits:

## #1
Report correct number of probes attached

This is specific to kprobe-multi and uprobe-multi,
which are attaching to multiple functions under the
hood. This is transparent to users and we should
report the number of attached functions correctly.

Issue: https://github.com/bpftrace/bpftrace/issues/3180

## #2

Change "Attaching X probes..." to "Attached X probes"

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ and this project adheres to
   - [#4137](https://github.com/bpftrace/bpftrace/pull/4137)
 - Fix execution watchpoints
   - [#4139](https://github.com/bpftrace/bpftrace/pull/4139)
+- Fix incorrect reporting of attached count for multi probes
+  - [#4194](https://github.com/bpftrace/bpftrace/pull/4194)
 #### Security
 #### Docs
 #### Tools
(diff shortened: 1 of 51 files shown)


amscanne (commented):
Aside: how hard would it be to plumb an API that allows incremental progress updates on attach? (Like an abstract `Progress&` that knows the total expected, and has methods called before and after things attach). This was a significant bit of user feedback from the workshop ("how do I know if something is happening?") and an opportunity to have the tool be a bit magical (everyone loves a sweet progress bar).

CHANGELOG.md
```diff
@@ -83,6 +83,8 @@ and this project adheres to
   - [#4137](https://github.com/bpftrace/bpftrace/pull/4137)
 - Fix execution watchpoints
   - [#4139](https://github.com/bpftrace/bpftrace/pull/4139)
+- Fix incorrect reporting of attached probe count
```
viktormalik, line 86: Maybe add that this is specific to multi-probes?

viktormalik (commented):
Looks good, how about we add a changelog entry for the second commit, too? I can imagine users grepping for `Attaching X probes...`.

jordalgo:
@viktormalik Updated the changelog with both suggestions.

jordalgo:
> an opportunity to have the tool be a bit magical (everyone loves a sweet progress bar)

Love this idea. Let me look into it or at least make a task and see if anyone wants to pick it up.

danobi (approved):

Issue #2064, opened 2021-11-14
Fail to run bpftrace with netif_receive_skb tracepoint


alpha-li:
### What reproduces the bug?

When I run bpftrace with
 ```
bpftrace -e 't:net:netif_receive_skb { @[str(args->name)] = lhist(cpu, 0, 128, 1); }'
```
it failed with the error messge:
```
stdin:1:27-43: ERROR: Struct/union of type 'struct _tracepoint_net_netif_receive_skb' does not contain a field named 'name'
t:net:netif_receive_skb { @[str(args->name)] = lhist(cpu, 0, 128, 1); }
                          ~~~~~~~~~~~~~~~~
```
Use bpftrace to show t:net:netif_receive_skb info:
```root@lizhao-VirtualBox:/usr# bpftrace -lv t:net:netif_receive_skb
BTF: using data from /sys/kernel/btf/vmlinux
tracepoint:net:netif_receive_skb
    void * skbaddr;
    unsigned int len;
    __data_loc char[] name;
```
### bpftrace --info
```
root@lizhao-VirtualBox:/usr# bpftrace --info
System
  OS: Linux 5.13.0-19-generic #19-Ubuntu SMP Thu Oct 7 21:58:00 UTC 2021
  Arch: x86_64

Build
  version: v0.11.3
  LLVM: 11
  foreach_sym: yes
  unsafe uprobe: no
  bfd: no
  bpf_attach_kfunc: yes
  bcc_usdt_addsem: yes
  bcc bpf_attach_uprobe refcount: yes
  libbpf: yes
  libbpf btf dump: yes
  libbpf btf dump type decl: yes

Kernel helpers
  probe_read: yes
  probe_read_str: yes
  probe_read_user: yes
  probe_read_user_str: yes
  probe_read_kernel: yes
  probe_read_kernel_str: yes
  get_current_cgroup_id: yes
  send_signal: yes
  override_return: yes

Kernel features
  Instruction limit: 1000000
  Loop support: yes
  btf: yes

Map types
  hash: yes
  percpu hash: yes
  array: yes
  percpu array: yes
  stack_trace: yes
  perf_event_array: yes

Probe types
  kprobe: yes
  tracepoint: yes
  perf_event: yes
  kfunc: yes
```

fbs:
Hi

v11 is quite old, we're on 14 now and iirc this has been fixed :). Can you try with 0.14 or newer? You can use the docker image for a static build

alpha-li:
[...]

PR #4196, opened 2025-05-28
usdt: update context initialization to use probe path

Phoenix500526:
Pass probe_.path to `bcc_usdt_new_frompid` since iovisor/bcc #2064 has already merged.
FYI: https://github.com/iovisor/bcc/pull/2064

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -1,8 +1,5 @@
 #include "ast/irbuilderbpf.h"
 
-#include <llvm/IR/DataLayout.h>
-#include <llvm/IR/Module.h>
-
 #include "arch/arch.h"
 #include "ast/async_event_types.h"
 #include "ast/codegen_helper.h"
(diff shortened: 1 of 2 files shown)


src/ast/irbuilderbpf.cpp
```diff
@@ -1467,8 +1467,11 @@ Value *IRBuilderBPF::CreateUSDTReadArgument(Value *ctx,
   void *usdt;
 
   if (pid.has_value()) {
-    // FIXME use attach_point->target when iovisor/bcc#2064 is merged
-    usdt = bcc_usdt_new_frompid(*pid, nullptr);
+    std::string real_path;
+    if (!attach_point->target.empty()) {
+      real_path = std::filesystem::absolute(attach_point->target).string();
+    }
+    usdt = bcc_usdt_new_frompid(*pid, real_path.c_str());
```
viktormalik, line 1474: I'm not sure this is correct. When `attach_point->target` is empty, `real_path.c_str()` will not return NULL but a pointer to an empty string instead. And looking at the [implementation of `bcc_usdt_new_frompid`](https://github.com/iovisor/bcc/blob/master/src/cc/usdt/usdt.cc#L460), it expects either NULL or a valid path.

Phoenix500526: Done

PR #4197, opened 2025-05-28
Add task_from_pid(pid) and task_release(task)

Rtoax:
linux v6.1 commit 3f0e6f2b41d3 ("bpf: Add bpf_task_from_pid() kfunc") introduce bpf_task_from_pid() kfunc to get 'struct task_struct *' from pid.

For example, in killsnoop.bt, the 'struct task_struct *' pointer of the target process can be obtained through the pid information of the target process to obtain more information.

Example:

    tracepoint:syscalls:sys_enter_kill,
    tracepoint:syscalls:sys_enter_tkill,
    tracepoint:syscalls:sys_enter_tgkill
    {
        $pid = args.pid;
        printf("%-15s %-8d %-16s %-8d",
            strftime("%H:%M:%S.%f", nsecs),
            pid, comm, $pid);

        $ttask = task_from_pid($pid);
        if ($ttask) {
            printf(" %-16s", $ttask->comm);
            task_release($ttask);
        }

        printf(" %-4d\n", args.sig);
    }

    16:48:59.238166 1755268  tgkill           1755269  target-thread    15
    16:49:48.287715 1755507  tkill            1755508  target-thread    15

Test C:

```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/syscall.h>
#include <pthread.h>
#include <signal.h>

static pid_t child_tid;

void *thread_func(void* arg)
{
	child_tid = syscall(SYS_gettid);
	while (1) {
		sleep(1);
		printf("child running...\n");
	}
	return NULL;
}

int main(void)
{
	pthread_t thread;
	pthread_create(&thread, NULL, thread_func, NULL);
	pthread_setname_np(thread, "target-thread");
	sleep(1);
#ifdef TGKILL
	tgkill(getpid(), child_tid, SIGTERM);
#else
	syscall(SYS_tkill, child_tid, SIGTERM);
#endif
	pthread_join(thread, NULL);
	return 0;
}
```

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -3132,6 +3132,41 @@ tracepoint:syscalls:sys_enter_execve {
 }
 ----
 
+[#functions-task_from_pid]
+=== task_from_pid
+
+.variants
+* `struct task_struct *task_from_pid(pid)`
+
+*compile time*
+
+Get 'struct task_struct *' pointer from pid.
+
+`task_from_pid()` needs to be used together with `task_release()`, otherwise it will fail the verifier check.
+
+----
+tracepoint:syscalls:sys_enter_kill,
+tracepoint:syscalls:sys_enter_tkill,
+tracepoint:syscalls:sys_enter_tgkill
+{
+    $ttask = task_from_pid(args.pid);
+    if ($ttask) {
+        /* do something */
+        task_release($ttask);
+    }
+}
+----
+
+[#functions-task_release]
+=== task_release
+
+.variants
+* `void task_release(TASK_STRUCT)`
+
+*compile time*
+
+`task_release()` releases the 'struct task_struct *' pointer obtained by `task_from_pid()`.
+
 [#functions-time]
 === time
 
--- a/src/ast/dibuilderbpf.cpp
+++ b/src/ast/dibuilderbpf.cpp
@@ -285,6 +285,9 @@ DIType *DIBuilderBPF::GetType(const SizedType &stype, bool emit_codegen_types)
                                                           emit_codegen_types),
                                                   64);
 
+  if (stype.IsVoidTy())
+    return createUnspecifiedType("void");
+
   // Integer types and builtin types represented by integers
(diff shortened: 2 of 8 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -3132,6 +3132,41 @@ tracepoint:syscalls:sys_enter_execve {
...
+=== task_from_pid
+
+.variants
+* `struct task_struct *task_from_pid(pid)`
+
+*compile time*
+
+Get 'struct task_struct *' pointer from pid.
+
+`task_from_pid()` needs to be used together with `task_release()`, otherwise it will fail the verifier check.
```
viktormalik, line 3145: Currently, our general strategy is to minimize the possibilities to trigger verifier errors from bpftrace scripts as those may be quite hard to debug for users.

So, I want to avoid merging this in the current form. However, there's clearly value in exposing these (and other ptr-acquiring) kfuncs so let's discuss how we do it in #4198.

viktormalik (changes requested):

man/adoc/bpftrace.adoc
Rtoax: Thanks, If we have a conclusion, I still want to do the work of `task_from_pid`, because I want to use this kfunc, thanks again.

amscanne: If we changed the `Variable` types in codegen to a `ScopedExpr`-style object, we could allow for the possibility of attached a "destructor" method for when those are reassigned to a new value or the scope ends. (Codegen would need to explicitly flush all of these at the end of the probe also, to ensure these are run.)

The resulting expression of this function call could carry around its own destructor, which would be the call to the release function. WDYT?

viktormalik: Yeah, using a destructor could be nice, however I think that we must make sure not to trigger the destructor when the pointer is assigned to a map.

Anyways, let's move the discussion to #4198, there are other ideas already proposed there.

PR #4201, opened 2025-05-28
attached_probe: kprobe: Use libbpf attachment API

### LINKED ISSUE #4145: Migrate kprobe attachment from bcc to libbpf API
`AttachedProbe::attach_kprobe()` uses bcc API for attaching non-multi and non-session kprobes. Such kprobes represent the majority of kprobe usage (IMO).

Migrate from bcc's `bpf_attach_kprobe()` to libbpf's `bpf_program__attach_kprobe[_opts]()`.

danobi:
Use libbpf API instead of bcc. libbpf is not only the maintained path forward, but it allows us to access APIs only available to link based kernel API.

This closes #4145.

##### Checklist

### DIFF
--- a/src/attached_probe.cpp
+++ b/src/attached_probe.cpp
@@ -537,31 +537,31 @@ public:
       BPFtrace &bpftrace);
   ~AttachedKprobeProbe() override;
 
+  int link_fd() override;
+
 private:
-  AttachedKprobeProbe(const Probe &probe,
-                      int progfd,
-                      int perf_event_fd,
-                      std::string event_name);
-  int perf_event_fd_;
-  const std::string event_name_;
+  AttachedKprobeProbe(const Probe &probe, int progfd, struct bpf_link *link);
+  struct bpf_link *link_;
 };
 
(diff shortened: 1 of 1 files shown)


src/attached_probe.cpp
```diff
@@ -601,26 +601,25 @@ Result<std::unique_ptr<AttachedKprobeProbe>> AttachedKprobeProbe::make(
...
-          << ", 0)";
-  int perf_event_fd = bpf_attach_kprobe(prog.fd(),
-                                        attachtype(probe.type),
-                                        eventname(probe, offset).c_str(),
-                                        funcname.c_str(),
-                                        offset,
-                                        0);
+  struct bpf_kprobe_opts opts;
+  ::memset(&opts, 0, sizeof(opts));
+  opts.sz = sizeof(struct bpf_kprobe_opts);
```
viktormalik, line 606: I know I'm late to the party but why don't we use `BPFTRACE_LIBBPF_OPTS` which we have for that purpose?

danobi: Ah yes. Forgot about that!

danobi: #4209

PR #4208, opened 2025-05-29
Allow statement lists in macro bodies

jordalgo:
This is to help the awkward case when macros
just have side effects and don't "return"
anything like block expressions.

##### Checklist


amscanne:
Seems reasonable to me, I started with the opinion that we could auto-parameterize *functions* and have LLVM inline them :)

PR #4210, opened 2025-05-29
build: vendor generated header files

amscanne:
This adds a CI check that ensures these headers are up to date, but means that this can be built without requiring the regeneration of all the bitcode data. This makes it easier to integrate with other build systems.

##### Checklist


ajor:
I don't like checking in generated files like this, particularly for what is executable binary code. Maybe we can come up with a different way to integrate with other build systems. I'm guessing Buck?

- It's adding unnecessary bloat to the repo. This bloat could grow a fair bit as the binary diffs will not be small.
- It's impossible to review changes to binary files.
- It's a security risk. The XZ Utils backdoor from a few years ago was partially hidden within binary files checked into the repo. I'm not saying you've hidden stuff in there, but the inability to review changes would make it hard to spot rogue additions in the future.

jordalgo:
@ajor Yeah, it's for BUCK. We can try to find another solution...

amscanne:
I'm not sure I agree with the points against it. It is mostly a philosophical difference (but checking in the generated code is the standard practice for the Go community, e.g. `go generate`, and it works pretty well). For the security angle, we'd be explicitly checking that those generated files are actually directly derived from the source, and they are pre-generated only for build convenience. No one would be expecting a review in these files, you review in the source files and then CI ensures that they are actually as expected. It is not the same as the checked-in binary that is not verifiable.

BUT it doesn't matter, because this is kinda broken anyways. The bitcode files are dependent on the LLVM major version, which is fine for the case where we build (because that implicit dependency already exists), but not for generating static files, unless we do so for every version. We need to think of some alternatives.

Issue #4151, opened 2025-05-19
Migrate usdt attachment from bcc to libbpf


danobi:
`AttachedProbe::attach_usdt()` uses bcc's `bpf_attach_uprobe()` to attach usdt probes (which are actually just backed by uprobes). This can be migrated to libbpf's `bpf_program__attach_uprobe()`.

Note that `attach_usdt()` makes use of a few other bcc APIs. This issue does not intend to migrate those usages to libbpf ones. That task would be much harder. It should be doable, though. We'll eventually do it, but sooner is better than later. Many bonus points if you can figure that out. This might need to be done in sync with other parts of the codebase like codegen.

Phoenix500526:
I'm interesting in this issue, please assign it to me. @danobi

Phoenix500526:
I noticed that there is an API named `bpf_program__attach_usdt` in libbpf. Why don't we migrate the `bpf_attach_uprobe` to the `bpf_program__attach_usdt`? @danobi @jordalgo

Phoenix500526:
# TL;DR
The BCC and libbpf frameworks provide different methods for attaching USDT probes:

* BCC: Uses the `bpf_attach_uprobe` function, requiring manual calculation of the offset, handling of semaphores, and allows custom event names.

* libbpf: Provides the `bpf_program__attach_usdt` function, which automatically extracts USDT information from ELF files and handles offsets and semaphores, but does not support custom event names.

Their underlying implementations are similar, as both create probes using either the `perf_event_open` system call or tracefs. Libbpf is simpler and easier to use, but it does not support custom event names. I am unsure if this will impact previous releases.

---

The BCC framework doesn't provide a `bpf_attach_usdt` API. Therefore, if we want to use the BCC framework to attach a USDT probe, `bpf_attach_uprobe` is our only choice. However, libbpf provides an API called bpf_program__attach_usdt for USDT probe attachment. These two APIs are pasted as follows:

```C++
// BCC: bpf_attach_usdt
int bpf_attach_uprobe(int progfd, enum bpf_probe_attach_type attach_type,
                      const char *ev_name, const char *binary_path,
                      uint64_t offset, pid_t pid, uint32_t ref_ctr_offset);

// libbpf: bpf_program__attach_usdt
struct bpf_link *bpf_program__attach_usdt(const struct bpf_program *prog,
					  pid_t pid, const char *binary_path,
					  const char *usdt_provider, const char *usdt_name,
[...]

PR #4217, opened 2025-05-30
attached_probe: usdt: Use libbpf attachment API

Phoenix500526:
Use libbpf API instead of bcc. libbpf is not only the maintained path 
forward, but it allows us to access APIs only available to link based
kernel API.

Currently, this PR uses `bpf_program__attach_uprobe_opts` instead 
of `bpf_program__attach_usdt` because BCC employs `bpf_attach_uprobe` 
to attach USDT probes, which requires integration with other `bpf_usdt_xxx` 
APIs. Libbpf's `bpf_program__attach_uprobe_opts` is more similar to 
`bpf_attach_uprobe`, so I have opted to use `bpf_program__attach_uprobe_opts` 
as a replacement for `bpf_attach_uprobe`. I will gradually transition to using 
libbpf to replace BCC in `AttachedUSDTProbe::make`, and at that point, I will 
revisit whether `bpf_program__attach_uprobe_usdt` can be used instead of 
`bpf_program__attach_uprobe_opts`.

Closes: #4151

##### Checklist


Phoenix500526:
> @Phoenix500526 So the plan is to add an additional commit to this PR for the `make` method?

As it mention in #4151 , I prefer to make another pr to solve this problem. Should we open a new issue?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1864,6 +1866,35 @@ If you're seeing errors, try clamping the string sizes. For example:
       return;
     }
     call.return_type = CreateUInt64();
+  } else if (call.func == "__usdt_arg") {
```
amscanne, line 1886: I can't remember quite what was required here, but with the C interop checks in the semantic analyzer now at HEAD, perhaps the USDT pass could simply inject the suitable casts?

Phoenix500526: This commit is just a temporary submission to save my code changes — it's not ready for review yet. It may be removed later - I haven't had time to clean up my changes yet.  Although I’ve managed to compile the BPF program for USDT, I'm currently encountering some errors(EBADF, EINVAL) when attaching it due to kernel version issues. Specifically, libbpf depends on the BPF_TOKEN_CREATE command, which was only introduced in kernel version 6.9 and later. I'm working on resolving this issue.

jordalgo: @Phoenix500526 Are you running bpftrace inside a container in a different user-namespace? I didn't think you needed any token creation/management aside from in that situation - I don't see any of the libbpf-bootstrap programs doing token creation: https://github.com/libbpf/libbpf-bootstrap

Phoenix500526: @jordalgo Neither in a container nor in different user-namespace. I got this conclusion from the strace output:
```bash
sudo strace -e trace=bpf,perf_event_open LIBBPF_BPF_TOKEN_PATH="" src/bpftrace -e 'usdt:tests/testprogs/usdt_args:usdt_args:reg_u64 { printf("hit\n"); }' -c tests/testprogs/usdt_args 
... 
# 0x24 is BPF_TOKEN_CREATE
bpf(0x24 /* BPF_??? */, 0x7ffcefa77db0, 8) = -1 EINVAL (Invalid argument)
...
bpf(BPF_LINK_CREATE, {link_create={prog_fd=3, target_fd=-1, attach_type=BPF_TRACE_UPROBE_MULTI, flags=0}, ...}, 64) = -1 EBADF (Bad file descriptor)
bpf(BPF_LINK_CREATE, {link_create={prog_fd=3, target_fd=-1, attach_type=BPF_TRACE_UPROBE_MULTI, flags=0}, ...}, 64) = -1 EINVAL (Invalid argument)
bpf(BPF_MAP_UPDATE_ELEM, {map_fd=5, key=0x7ffcefa77eb0, value=0x5e779fd6e148, flags=BPF_ANY}, 32) = 0
bpf(BPF_LINK_CREATE, {link_create={prog_fd=10, target_fd=0, attach_type=BPF_TRACE_UPROBE_MULTI, flags=0}, ...}, 64) = -1 EINVAL (Invalid argument)
ERROR: Failed to attach usdt probe: Invalid argument
ERROR: Unable to attach probe: usdt:tests/testprogs/usdt_args:usdt_args:reg_u64. If this is expected, set the 'missing_probes' config variable to 'warn'.
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_KILLED, si_pid=73770, si_uid=0, si_status=SIGKILL, si_utime=0, si_stime=0} ---
+++ exited with 255 +++
```
Setting the env `LIBBPF_BPF_TOKEN_PATH` to an empty value can disable this implicit BPF token creation. 

FYI: https://lwn.net/Articles/959350/

Phoenix500526: I’ve identified that the direct cause of the usdt probe attach failure is an error returned from the `bpf_link_create` function. The call chain that leads to this function is:

```markdown
bpf_program__attach_usdt
  → usdt_manager_attach_usdt
    → bpf_program__attach_uprobe_multi
      → bpf_link_create (returns -1)
```

To investigate and resolve this issue, I’ve tried several approaches:
1. Upgraded the kernel to 6.12 – no effect; the issue persists.
2. Used bpftrace to trace some internal helpers in `__sys_bpf` syscall, like `bpf_check_uarg_tail_zero`, `security_bpf`, and `bpf_prog_get`. Both `bpf_check_uarg_tail_zero` and `security_bpf` were triggered and returned successfully. However, `bpf_prog_get` was never triggered. Given the kernel source of `link_create`(https://elixir.bootlin.com/linux/v6.8/source/kernel/bpf/syscall.c#L5003), it’s possible that `CHECK_ATTR` returned a non-NULL value and short-circuited the process. Unfortunately, since `memchr_inv` is untraceable, I couldn’t gather further information along this direction.
3. Created two minimal demo programs based on [libbpf-bootstrap](https://github.com/libbpf/libbpf-bootstrap):
* One attaches a usdt probe using bpf_program__attach_usdt, which succeeds.
* The other attaches a uprobe using bpf_program__attach_uprobe_multi, which fails.

Intriguingly, the usdt probe correctly links to the target binary, while the uprobe probe fails with the same behavior and return value (bpf_link_create → -1) as the bpftrace-based usdt probe.

At this point, I’m out of ideas and would appreciate any suggestions or debugging tips.
Below is the minimal reproducible example using `bpf_program__attach_uprobe_multi`, which seems to consistently reproduce the failure.

The `uprobe.bpf.c` file:
```c
#include <linux/ptrace.h>
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_tracing.h>

char LICENSE[] SEC("license") = "Dual BSD/GPL";

SEC("uprobe")
int uprobe_add(struct pt_regs *ctx)
{
    int a = (int)PT_REGS_PARM1(ctx);
    int b = (int)PT_REGS_PARM2(ctx);
    bpf_printk("uprobed_add ENTRY: a = %d, b = %d", a, b);
    return 0;
}

SEC("uretprobe")
int uretprobe_add(struct pt_regs *ctx)
{
    int ret = (int)PT_REGS_RC(ctx);
    bpf_printk("uprobed_add EXIT: return = %d", ret);
    return 0;
}

SEC("uprobe//proc/self/exe:uprobed_sub")
int uprobe_sub(struct pt_regs *ctx)
{
    int a = (int)PT_REGS_PARM1(ctx);
    int b = (int)PT_REGS_PARM2(ctx);
    bpf_printk("uprobed_sub ENTRY: a = %d, b = %d", a, b);
    return 0;
}

SEC("uretprobe//proc/self/exe:uprobed_sub")
int uretprobe_sub(struct pt_regs *ctx)
{
    int ret = (int)PT_REGS_RC(ctx);
    bpf_printk("uprobed_sub EXIT: return = %d", ret);
    return 0;
}
```
The `uprobe.c` file:
```c
#include <errno.h>
#include <stdio.h>
#include <unistd.h>
#include <sys/resource.h>
#include <bpf/libbpf.h>
#include "uprobe.skel.h"
static int libbpf_print_fn(enum libbpf_print_level level, const char *format, va_list args)
{
	return vfprintf(stderr, format, args);
}
__attribute__((noinline)) int uprobed_add(int a, int b)
{
	asm volatile ("");
	return a + b;
}
__attribute__((noinline)) int uprobed_sub(int a, int b)
{
	asm volatile ("");
	return a - b;
}
int main(int argc, char **argv)
{
	struct uprobe_bpf *skel;
	int err, i;
	LIBBPF_OPTS(bpf_uprobe_opts, uprobe_opts);
	libbpf_set_print(libbpf_print_fn);

	skel = uprobe_bpf__open_and_load();
	if (!skel) {
		fprintf(stderr, "Failed to open and load BPF skeleton\n");
		return 1;
	}

	/* 
	 * Method 2: More efficient - use one BPF program to handle multiple functions
	 * This requires modifying the BPF program to be more generic
	 */
	
	/* Define all functions we want to trace */
	const char *entry_syms[] = {"uprobed_add", "uprobed_sub"};
	size_t entry_cnt = sizeof(entry_syms) / sizeof(entry_syms[0]);
	
	/* If you have a generic uprobe program that can handle multiple functions */
	LIBBPF_OPTS(bpf_uprobe_multi_opts, entry_opts);
	entry_opts.syms = entry_syms;
	entry_opts.cnt = entry_cnt;
	entry_opts.retprobe = false;
	
	/* Note: This assumes you have a generic uprobe program.
	 * For the original separate programs, use individual attachments as shown above.
	 */
	
	/* Individual attachments (more practical for this specific case) */
	
	/* Attach uprobe for uprobed_add */
	const char *add_sym[] = {"uprobed_add"};
	LIBBPF_OPTS(bpf_uprobe_multi_opts, add_uprobe_opts);
	add_uprobe_opts.syms = add_sym;
	add_uprobe_opts.cnt = 1;
	add_uprobe_opts.retprobe = false;
	
	skel->links.uprobe_add = bpf_program__attach_uprobe_multi(skel->progs.uprobe_add,
								  0, "/proc/self/exe", NULL,
								  &add_uprobe_opts);
	if (!skel->links.uprobe_add) {
		err = -errno;
		fprintf(stderr, "Failed to attach uprobe_multi for uprobe_add: %d\n", err);
		goto cleanup;
	}

	add_uprobe_opts.retprobe = true;
	skel->links.uretprobe_add = bpf_program__attach_uprobe_multi(skel->progs.uretprobe_add,
								     -1, "/proc/self/exe", NULL,
								     &add_uprobe_opts);
	if (!skel->links.uretprobe_add) {
		err = -errno;
		fprintf(stderr, "Failed to attach uprobe_multi for uretprobe_add: %d\n", err);
		goto cleanup;
	}

	const char *sub_sym[] = {"uprobed_sub"};
	LIBBPF_OPTS(bpf_uprobe_multi_opts, sub_uprobe_opts);
	sub_uprobe_opts.syms = sub_sym;
	sub_uprobe_opts.cnt = 1;
	sub_uprobe_opts.retprobe = false;
	
	skel->links.uprobe_sub = bpf_program__attach_uprobe_multi(skel->progs.uprobe_sub,
								  0, "/proc/self/exe", NULL,
								  &sub_uprobe_opts);
	if (!skel->links.uprobe_sub) {
		err = -errno;
		fprintf(stderr, "Failed to attach uprobe_multi for uprobe_sub: %d\n", err);
		goto cleanup;
	}
	sub_uprobe_opts.retprobe = true;
	skel->links.uretprobe_sub = bpf_program__attach_uprobe_multi(skel->progs.uretprobe_sub,
								     0, "/proc/self/exe", NULL,
								     &sub_uprobe_opts);
	if (!skel->links.uretprobe_sub) {
		err = -errno;
		fprintf(stderr, "Failed to attach uprobe_multi for uretprobe_sub: %d\n", err);
		goto cleanup;
	}

	printf("Successfully started! Please run `sudo cat /sys/kernel/debug/tracing/trace_pipe` "
	       "to see output of the BPF programs.\n");

	for (i = 0;; i++) {
		fprintf(stderr, ".");
		uprobed_add(i, i + 1);
		uprobed_sub(i * i, i);
		sleep(1);
	}

cleanup:
	uprobe_bpf__destroy(skel);
	return -err;
}
```
The Makefile:
```Makefile
CC = gcc
CLANG = clang
BPFTOOL = bpftool
CFLAGS = -Wall -Wextra -O0 -g
LDFLAGS = -lbpf -lelf -lz
TARGET = uprobe
BPF_TARGET = uprobe.bpf.o
SOURCE = uprobe.c
BPF_SOURCE = uprobe.bpf.c
UPROBE_SKEL_H = uprobe.skel.h
LIBBPF_CFLAGS = $(shell pkg-config --cflags libbpf)
LIBBPF_LDFLAGS = $(shell pkg-config --libs libbpf)

BPF_CFLAGS = -target bpf -D__TARGET_ARCH_x86 -I/usr/include/x86_64-linux-gnu -O0 -g

version:
	@echo "System libbpf version:"
	@pkg-config --modversion libbpf
	@echo "libbpf cflags: $(LIBBPF_CFLAGS)"
	@echo "libbpf ldflags: $(LIBBPF_LDFLAGS)"

all: $(TARGET)

$(BPF_TARGET): $(BPF_SOURCE)
	$(CLANG) $(BPF_CFLAGS) -c $< -o $@
	$(BPFTOOL) gen skeleton $@ > $(UPROBE_SKEL_H)

$(TARGET): $(SOURCE) $(BPF_TARGET)
	$(CC) $(CFLAGS) $(LIBBPF_CFLAGS) -o $@ $< $(LIBBPF_LDFLAGS)

test: $(TARGET)
	sudo ./$(TARGET)

clean:
	rm -f $(TARGET) $(BPF_TARGET) $(UPROBE_SKEL_H)

.PHONY: version test clean 
```

jordalgo:
@Phoenix500526 - Sorry for the delay. @amscanne and I will take a look this week if you haven't figured it out already.

Phoenix500526:
> @Phoenix500526 - Sorry for the delay. @amscanne and I will take a look this week if you haven't figured it out already.

@jordalgo  Thanks! I've solved the most problems. There are three failed test cases so far and I'm taking 4 days off from tomorrow. Please drop me a message if anything comes up.

src/ast/passes/parser.h
```diff
@@ -34,6 +35,8 @@ inline std::vector<Pass> AllParsePasses(
   passes.emplace_back(CreateMacroExpansionPass());
   passes.emplace_back(CreateDeprecatedPass());
   passes.emplace_back(CreateParseAttachpointsPass());
+  passes.emplace_back(CreateUSDTImportPass()); // Import USDT stdlib if needed
+  passes.emplace_back(CreateProbeExpansionPass());
```
jordalgo:
```suggestion
```
Rm: this was causing an issue after rebasing

Phoenix500526: Done

jordalgo:
So I'm still investigating the test failures and should have a few suggestions for fixes up soon.

Phoenix500526:
> So I'm still investigating the test failures and should have a few suggestions for fixes up soon.

@jordalgo  Oh, thanks, that's really great. I also have some new findings. One of the reason causing the failure of index_*-type USDT probe is that the current libbpf doesn't support SIB(Scale Index Base) addressing mode. The libbpf doesn't handle the SIB argument, like "-1@-96(%rbp,%rax,8)", correctly. I've submit a [patch](https://patchwork.kernel.org/project/netdevbpf/list/?series=986784&state=*) to solve this issue.

jordalgo:
@Phoenix500526 Funny, I was about to post the same finding but you're ahead of me and already posted a patch, which is awesome! IMHO, we don't have to wait for this patch to land before migrating as it feels like a pretty edge-case usage of USDTs (@amscanne or @viktormalik - feel to disagree).

jordalgo:
@Phoenix500526 So the Scale Index Base issue is accounting for 2 of the 3 test failures. The last one `duplicated markers` is failing because we're attaching two probes when we should just be attaching one. I believe that because this PR now uses uprobe-multi a single bpf program will get events from both tracepoints if they have the same name/marker.

If I run `./testprogs/usdt_inlined` in the background and then run this:
```
$ ninja && sudo ./src/bpftrace -e 'usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe { printf("%d\n", arg1); }' -v
Trying to attach probe: usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe
Trying to attach probe: usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe
Attached 2 probes
999
999
100
100
999
999
100
100
```
You see we have two probes both getting both USDT calls. However, if I hack a bit of code to only make sure to attach one then it works as expected:
```
$ ninja && sudo ./src/bpftrace -e 'usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe { printf("%d\n", arg1); }' -v
Trying to attach probe: usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe
Attached 1 probes
999
100
999
100
```

EDIT:
So you might be able to get away with hard-coding num_locations to 1:
https://github.com/bpftrace/bpftrace/blob/master/src/usdt.cpp#L39

But that also begs the question of if we can remove the `bcc_usdt_new_frompid` and `bcc_usdt_new_frompath` calls.

Phoenix500526:
> @Phoenix500526 So the Scale Index Base issue is accounting for 2 of the 3 test failures. The last one `duplicated markers` is failing because we're attaching two probes when we should just be attaching one. I believe that because this PR now uses uprobe-multi a single bpf program will get events from both tracepoints if they have the same name/marker.
> 
> If I run `./testprogs/usdt_inlined` in the background and then run this:
> 
> ```
> $ ninja && sudo ./src/bpftrace -e 'usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe { printf("%d\n", arg1); }' -v
> Trying to attach probe: usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe
> Trying to attach probe: usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe
> Attached 2 probes
> 999
> 999
> 100
> 100
> 999
> 999
> 100
> 100
> ```
> 
> You see we have two probes both getting both USDT calls. However, if I hack a bit of code to only make sure to attach one then it works as expected:
> 
> ```
> $ ninja && sudo ./src/bpftrace -e 'usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe { printf("%d\n", arg1); }' -v
> Trying to attach probe: usdt:./tests/testprogs/usdt_inlined:tracetest:testprobe
> Attached 1 probes
> 999
> 100
> 999
> 100
> ```
> 
> EDIT: So you might be able to get away with hard-coding num_locations to 1: https://github.com/bpftrace/bpftrace/blob/master/src/usdt.cpp#L39
> 
> But that also begs the question of if we can remove the `bcc_usdt_new_frompid` and `bcc_usdt_new_frompath` calls.

I'm trying to remove these two calls. I noticed a strange thing. IIRC, the libbpf is static linked currently. So the `bpf_program__attach_usdt` which comes from libbpf.a should be defined in bpftrace. But when I used nm to find this symbol, I got this:
```bash
$ nm src/bpftrace | grep bpf_program__attach_usdt
                 U bpf_program__attach_usdt
```
Actually, it should be `T bpf_program__attach_usdt`. And as it happens, the libbcc_bpf.so defines a same symbol. 
```bash
$ nm /lib/x86_64-linux-gnu/libbcc_bpf.so.0 | grep bpf_program__attach_usdt
00000000000dfe10 T bpf_program__attach_usdt
```
It's obvious that the symbol `bpf_program__attach_usdt` in my bpftrace is come from libbcc_bpf.so. Can you help me to check this? Just nm your bpftrace to see whether the `bpf_program__attach_usdt` symbol is T or U.  I guess it may be some environment issues on my PC. Thanks.  @jordalgo

tests/runtime/usdt
```diff
@@ -114,19 +114,25 @@ NAME usdt probes - attach to probes by wildcard file with child
 RUN {{BPFTRACE}} -e 'usdt:./testprogs/usdt_test*::* { printf("here\n" ); exit(); }' -c ./testprogs/usdt_test
 EXPECT Attached 3 probes
 
-NAME usdt probes - attach to probe on multiple files by wildcard
-PROG usdt:./testprogs/usdt*::* { printf("here\n" ); exit(); }
-EXPECT Attached 49 probes
+# TODO(Jiawei Zhao): re-enable this test when libbpf can handle SIB usdt argument spec correctly
```
jordalgo, line 117: nit: did you want to add your patch link?

tests/runtime/usdt
```diff
@@ -114,19 +114,25 @@ NAME usdt probes - attach to probes by wildcard file with child
...
+
+# When uprobe_multi is supported, a single ebpf program can be attached to multiple
+# identical usdt probes, so we only need to attach to 3 different usdt probes.
+# For systems without uprobe_multi support, we need to manually attach an ebpf
+# program to each probe location, resulting in 6 probes being attached.
 NAME usdt probes - attach to multiple probes with different number of locations by wildcard
 PROG usdt:./testprogs/usdt_multiple_locations:tracetest:testprobe* { printf("here\n" ); exit(); }
 BEFORE ./testprogs/usdt_multiple_locations
-EXPECT Attached 6 probes
+EXPECT_REGEX Attached (3|6) probes
```
jordalgo: What do you think about having an additional test that utilizes the `REQUIRES_FEATURE uprobe_multi` directive ?

src/stdlib/usdt/usdt.bpf.c
```diff
@@ -0,0 +1,10 @@
+#define __VMLINUX_H__
+#include <vmlinux.h>
+#include <bpf/usdt.bpf.h>
+
+long __usdt_arg(long ctx, long arg_num)
+{
+  long _x;
+  bpf_usdt_arg((void *)ctx, arg_num, &_x);
+  return _x;
+}
```
jordalgo: nit: missing empty end line

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2088,6 +2088,45 @@ ScopedExpr CodegenLLVM::visit(Call &call)
     auto scoped_arg = visit(call.vargs.at(0));
 
     return ScopedExpr(b_.CreateGetSocketCookie(scoped_arg.value(), call.loc));
+  } else if (call.func == "__usdt_arg") {
```
jordalgo, line 2091: Since we're creating the AST for this function do we really need this codegen part at all? (cc @amscanne) Can't we do the necessary casts there too?

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1881,6 +1883,36 @@ If you're seeing errors, try clamping the string sizes. For example:
...
+    // Special handling for __usdt_arg function in USDT context
+    // This function expects (long, long) but we pass (struct pt_regs *, int64)
+    // The ctx parameter should be treated as a pointer value (long)
+    if (call.vargs.size() != 2) {
```
jordalgo, line 1890: The semantic analyser checks here seem a bit unnecessary since we're the ones creating this call. Is this just for folks who might find the internal `__usdt_arg` function directly?

jordalgo (commented):
@Phoenix500526 Thanks much for continuing to work on this. Just a few more questions/comments.

src/ast/passes/codegen_llvm.cpp
amscanne: I think this will all happen automatically now. You could even add a macro in `stdlib/usdt/usdt.bt` that does whatever casting is necessary, e.g.

```
// usdt_arg returns the ...
macro usdt_arg($x) {
   __usdt_arg((int64)ctx, (int64)$x)
}
```

(You would use the `usdt_arg` macro in the transform, not the underlying C function. This would also allow you document the `usdt_arg` macro in case someone wants to just use this explicitly?)

src/ast/passes/semantic_analyser.cpp
amscanne: These should also be covered automatically now. I recommend that you just wrap in a macro to do whatever casting is needed, per above.

tests/runtime/usdt
Phoenix500526: Done

src/stdlib/usdt/usdt.bpf.c
Phoenix500526: Done

tests/runtime/usdt
Phoenix500526: Done

src/ast/passes/semantic_analyser.cpp
Phoenix500526: @amscanne  In the latest commit 37a5265c, I remove the checks logic in `codegen_llvm.cpp` and `semantic_analyser.cpp`, and define a macro to instead of calling the C function directly. But I got some errors:
```bash
Command: /home/runner/work/bpftrace/bpftrace/build-ci/src/bpftrace -e 'config = { show_debug_info=1 } usdt:./testprogs/usdt_inlined:tracetest:testprobe { if (arg1 == 100) { printf("%s\n", ustack()); exit(); } }' -c ./testprogs/usdt_inlined
  	Unclean exit code: 1
  	Output: stdin:1:31-140: ERROR: Unknown function: 'usdt_arg'\nconfig = { show_debug_info=1 } usdt:./testprogs/usdt_inlined:tracetest:testprobe { if (arg1 == 100) { printf("%s\\n", ustack()); exit(); } }\n                              
```

Did I make something wrong?

jordalgo: @Phoenix500526  This error is because your new pass (`passes.emplace_back(CreateUSDTImportPass());`) needs to go before the macro expansion pass in `parser.h`

Phoenix500526: @jordalgo I remember that I've tried that before, but it doesn't work.

jordalgo: @Phoenix500526 That's the only was it can work. I pulled down your branch yesterday and tried it - it gets rid of the error about a not found `usdt_arg` function but it seems like there is another issue with attachment now. I can take a look today and do some debugging.

jordalgo: @Phoenix500526 Ok, I debugged a bit and there is both an ordering issue with the parser.h and something weird with the conditional importing of the usdt.bt file, which I still have to dig into but solved it by just moving the `usdt_arg` macro to `base.bt`. Try these changes: https://github.com/jordalgo/bpftrace/commit/39ec6f87c450ecb507b54e1d3ea8a944d64fc5bf

Phoenix500526: @jordalgo Thanks much for debugging. I've tried it but there are some other errors.

amscanne:
Crazy, it seems like codegen isn't handling the pointer cast correctly. I assume that we rely on the implicit casting during map assignment and other cases. I have all the runtime errors go ahead on my system with this patch:

```
diff --git a/src/ast/passes/codegen_llvm.cpp b/src/ast/passes/codegen_llvm.cpp
index d831f9d5..af971b2d 100644
@@ -2868,6 +2868,8 @@ ScopedExpr CodegenLLVM::visit(Cast &cast)
         array = buf;
       }
       return ScopedExpr(b_.CreateLoad(int_ty, array, true));
+    } else if (cast.expr.type().IsPtrTy()) {
+      return ScopedExpr(b_.CreatePtrToInt(scoped_expr.value(), int_ty));
     } else {
       return ScopedExpr(
           b_.CreateIntCast(scoped_expr.value(),
```

But I wasn't able to test the multi-attach paths, and there may be some other things there. I will push that as a separate PR.

amscanne:
Summarizing discussion from meeting. @Phoenix500526 will drop the last commit with the macro migration and keep the manual C types. This is apparently all working. We will fix a couple of issues and then separately refactor this code.

Issue #4152, opened 2025-05-19
Migrate uprobe attachment from bcc to libbpf API


danobi:
`AttachedProbe::attach_uprobe()` uses bcc's `bpf_attach_uprobe()`. Migrate this to libbpf's `bpf_program__attach_uprobe[_opts]()`.

Phoenix500526:
I'm interesting in this issue, please assign it to me. @danobi

PR #4218, opened 2025-05-30
attached_probe: uprobe: Use libbpf attachment API

Phoenix500526:
Use libbpf API instead of bcc. libbpf is not only the maintained path forward, but it allows us to access APIs only available to link based kernel API.

Closes: #4152.

##### Checklist


jordalgo (approved):

PR #4223, opened 2025-05-31
Add support for bare block statements

jordalgo:
See individual commits but this essentially adds support for bare block statements e.g.
```
BEGIN {
  $a = 1;
  {
    $b = 2;
    print($b);
    {
       $c = 3;
       print($c);
    }
   }
   // $b and $c are not accessible
  }
```

This makes a distinction between a Block and a BlockExpression, where the latter evaluates to an expression and the former does not.

This also enables macros to produce a Block or a BlockExpression, which fixes some confusing behavior where macros need to end in a dangling expression even if it only produces side effects e.g. `macro print_me() { print("me"); }`

##### Checklist

### DIFF
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -140,7 +140,7 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 
 %type <ast::AttachPoint *> attach_point
 %type <ast::AttachPointList> attach_points
-%type <ast::Block *> block_expr
+%type <ast::BlockExpr *> block_expr
 %type <ast::Call *> call
 %type <ast::Sizeof *> sizeof_expr
 %type <ast::Offsetof *> offsetof_expr
@@ -164,7 +164,7 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %type <ast::Config *> config
 %type <ast::Import *> import_stmt
 %type <ast::ImportList> imports
-%type <ast::Statement> assign_stmt block_stmt expr_stmt if_stmt jump_stmt loop_stmt for_stmt
+%type <ast::Statement> assign_stmt block_stmt expr_stmt if_stmt jump_stmt loop_stmt for_stmt bare_block
 %type <ast::MapDeclList> map_decl_list
 %type <ast::VarDeclStatement *> var_decl_stmt
 %type <ast::StatementList> block block_or_if stmt_list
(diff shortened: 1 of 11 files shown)


src/parser.yy
```diff
@@ -362,7 +362,9 @@ macros:
 
 macro:
                 MACRO IDENT "(" macro_args ")" block_expr { $$ = driver.ctx.make_node<ast::Macro>($2, std::move($4), $6, @$); }
-        |       MACRO IDENT "(" ")" block_expr { $$ = driver.ctx.make_node<ast::Macro>($2, ast::ExpressionList{}, $5, @$); }
+        |       MACRO IDENT "(" ")" block_expr            { $$ = driver.ctx.make_node<ast::Macro>($2, ast::ExpressionList{}, $5, @$); }
+        |       MACRO IDENT "(" macro_args ")" "{" stmt_list "}" { $$ = driver.ctx.make_node<ast::Macro>($2, std::move($4), driver.ctx.make_node<ast::Block>(std::move($7), @7), @$); }
```
amscanne: Why not use the `bare_block` here?

src/parser.yy
```diff
@@ -362,7 +362,9 @@ macros:
 
 macro:
                 MACRO IDENT "(" macro_args ")" block_expr { $$ = driver.ctx.make_node<ast::Macro>($2, std::move($4), $6, @$); }
-        |       MACRO IDENT "(" ")" block_expr { $$ = driver.ctx.make_node<ast::Macro>($2, ast::ExpressionList{}, $5, @$); }
+        |       MACRO IDENT "(" ")" block_expr            { $$ = driver.ctx.make_node<ast::Macro>($2, ast::ExpressionList{}, $5, @$); }
```
amscanne: I believe the function args have a node that produces an optionally empty list. Did you want to have a `macro_args` node? It would make this easier to understand IMO.

amscanne (commented):
Minor feedback, but seems reasonable to me. I think it's generally a good thing to split the types into precise subtypes, e.g. `BlockExpr`.

danobi:
High level I think this change makes sense. It'll be confusing to users why macro bodies must be an expression.

Btw, 3rd commit message doesn't seem to match commit contents? Also missing a rationale.

jordalgo:
> Btw, 3rd commit message doesn't seem to match commit contents? Also missing a rationale.

Ugh, I must have messed up my rebasing. I'll just combine it into the parent commit.

src/parser.yy
jordalgo: Cause I was doing something dumb.

src/parser.yy
jordalgo: Fixed.

src/parser.yy
jordalgo: Fixed.

jordalgo:
I think this is good to go.

amscanne (approved):

PR #4226, opened 2025-06-03
build: move all generated files to the same structure

amscanne:
build: move all generated files to the same structure


jordalgo (approved):
Okey dokey.

src/stdlib/CMakeLists.txt
```diff
@@ -32,10 +32,30 @@ embed(
...
 # The standard library holds these definitions internally via the headers
 # above. This can be made more dynamic in the future (e.g. using dynamic
 # initializers and registration), but given the size and scope the simpler
 # solution is more desirable for now.
 
 add_library(stdlib STATIC stdlib.cpp)
-add_dependencies(stdlib base_bitcode base_btf base_script)
+add_dependencies(stdlib
+  base_btf
+  base_bitcode
```
viktormalik: Super-nit - why swapping the order of these? They are no longer alphabetically ordered.

src/clang_parser.cpp
```diff
@@ -119,45 +119,20 @@ class ClangParser {
...
+      const auto &n = names.emplace_back("/bpftrace/" + name);
+      unsaved_files.push_back(CXUnsavedFile{
+          .Filename = n.c_str(),
+          .Contents = view.data(),
+          .Length = view.length(),
+      });
+    }
+    return std::make_pair(std::move(names), std::move(unsaved_files));
+  }();
+  return cached.second; // Just the map part.
```
viktormalik: Why do we return the names part, then?

src/clang_parser.cpp
```diff
@@ -119,45 +119,20 @@ class ClangParser {
...
-        .Contents = stdint_h.data(),
-        .Length = stdint_h.length(),
-    },
-  };
-
-  return unsaved_files;
+  static auto cached = [] {
+    std::vector<std::string> names;
+    std::vector<CXUnsavedFile> unsaved_files;
+    for (const auto &[name, view] : stdlib::Stdlib::files) {
```
viktormalik, line 130: Shouldn't we filter just the header files here?

amscanne: Good call, I will do that.

src/stdlib/CMakeLists.txt
amscanne: Oh, I was just matching the order above. I will swap it here and above.

src/clang_parser.cpp
amscanne: Because you have a `.c_str()` in there and need to preserve that `std::string` for the lifetime of the possible usage (basically cached.second points to memory inside cached.first). I will add a comment here that explains it.

PR #4228, opened 2025-06-03
ast: add ClangBuilder

amscanne:
Stacked PRs:
 * #4050
 * __->__#4228

--- --- ---

### ast: add ClangBuilder

This is a pass which will build `llvm::Module` using the linked
`libclang` for any C sources which have been imported. Since bitcode is
not stable, we need to use this mechanism or jump through very annoying
build hoops to have any kind of stability.

Instead, we can efficiently memoize compiled files (e.g. in the standard
`~/.cache/bpftrace` directory) based on a unique version of a) bpftrace,
b) our linked LLVM libraries, and c) the contents of the C file and all
imported headers.

Using C files directly allows us to remove the bitcode and BTF
generation that is currently part of the build. The build paths are
still neccessary for tests, but are not necessary in the main path.

### DIFF
--- a/src/ast/passes/clang_build.cpp
+++ b/src/ast/passes/clang_build.cpp
@@ -1,0 +1,215 @@
+#include <clang/CodeGen/CodeGenAction.h>
+#include <clang/Driver/Driver.h>
+#include <clang/Frontend/CompilerInstance.h>
+#include <clang/Frontend/FrontendActions.h>
+#include <clang/Frontend/TextDiagnosticPrinter.h>
+#include <fcntl.h>
+#include <fstream>
+#include <llvm/ADT/IntrusiveRefCntPtr.h>
+#include <llvm/Bitcode/BitcodeWriter.h>
+#include <llvm/Support/MemoryBuffer.h>
+#include <llvm/Support/VirtualFileSystem.h>
+#include <llvm/Support/raw_ostream.h>
+#include <llvm/TargetParser/Host.h>
+#include <sstream>
+
+#include "ast/ast.h"
+#include "ast/passes/clang_build.h"
+#include "ast/passes/resolve_imports.h"
+#include "stdlib/stdlib.h"
+#include "util/result.h"
+
+namespace bpftrace::ast {
+
+char ClangBuildError::ID;
+
+void ClangBuildError::log(llvm::raw_ostream &OS) const
+{
+  OS << msg_;
+}
+
+namespace {
+
+class PipeFds {
+public:
+  PipeFds(int rfd, int wfd)
+      : rfd_(rfd),
+        wfd_(wfd),
+        read_("/dev/fd/" + std::to_string(rfd)),
+        write_("/dev/fd/" + std::to_string(wfd)) {};
+  PipeFds(PipeFds &&other)
+      : rfd_(other.rfd_),
+        wfd_(other.wfd_),
+        read_(other.read_),
+        write_(other.write_)
+  {
+    other.rfd_ = -1;
+    other.wfd_ = -1;
+  }
+  PipeFds(const PipeFds &other) = delete;
+  ~PipeFds()
+  {
+    close_read();
+    close_write();
+  }
+
+  const std::string &write_file()
+  {
+    return write_;
+  }
+
+  std::string read_all()
+  {
+    close_write();
+    std::ifstream file(read_);
+    if (file.fail()) {
+      return ""; // Nothing to read.
+    }
+    std::stringstream contents;
+    contents << file.rdbuf();
+    return contents.str();
+  }
+
+  void close_read()
+  {
+    if (rfd_ >= 0) {
+      close(rfd_);
+      rfd_ = -1;
+    }
+  }
+
+  void close_write()
+  {
+    if (wfd_ >= 0) {
+      close(wfd_);
+      wfd_ = -1;
+    }
+  }
+
+private:
+  int rfd_ = -1;
+  int wfd_ = -1;
+  std::string read_;
+  std::string write_;
+};
+
+Result<PipeFds> create_pipe()
+{
+  int fds[2];
+  if (pipe2(fds, O_CLOEXEC) < 0) {
+    return make_error<ClangBuildError>("failed to create pipes");
+  }
+  return PipeFds(fds[0], fds[1]);
+}
+
+} // namespace
+
+static Result<> build(const std::string &name,
+                      LoadedObject &obj,
+                      Imports &imports,
+                      BitcodeModules &result)
+{
+  llvm::IntrusiveRefCntPtr<llvm::vfs::InMemoryFileSystem> vfs(
+      new llvm::vfs::InMemoryFileSystem());
+  vfs->addFile(name, 0, llvm::MemoryBuffer::getMemBuffer(obj.data()));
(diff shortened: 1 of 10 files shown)


ajor:
This sounds like a big change that could do with some discussion. Am I right in understanding that the intention is to compile the C stdlib on the first launch of bpftrace, rather than compiling and embedding it into bpftrace during bpftrace's compilation?

jordalgo:
> This sounds like a big change that could do with some discussion. 

This is why it's still in draft mode. @amscanne and I are still working through some options.

> Am I right in understanding that the intention is to compile the C stdlib on the first launch of bpftrace, rather than compiling and embedding it into bpftrace during bpftrace's compilation?

Ultimately, yes. And then those compiled C stdlib files will be cached so they don't unnecessarily re-compile for further bpftrace invocations.

amscanne:
To compile and cache C extensions dynamically, yes. I agree that this is fraught, and would only make sense if this can be fully hermetic (i.e. not depending on anything in the environment). It is still just an idea to explore, among some others.

ajor:
I don't think I fully understand the reason this is needed though. Right now (if I understand correctly), we compile the stdlib as part of bpftrace's compilation, using LLVM-version-xx. If bpftrace is linked against the same LLVM-version-xx and uses that to compile scripts, then the bitcode should be compatible, right?

amscanne:
> I don't think I fully understand the reason this is needed though. Right now (if I understand correctly), we compile the stdlib as part of bpftrace's compilation, using LLVM-version-xx. If bpftrace is linked against the same LLVM-version-xx and uses that to compile scripts, then the bitcode should be compatible, right?

Right, but I think it is very difficult to guarantee this without a ton of yak shaving for internal BUCK-based builds. You have (a) the toolchains for BUCK, and (b) the LLVM libs, and there doesn't seem to be a good way to ensure that (a) == (b) (or maybe we haven't spent enough time, my intuition is that even if we contrive something, it will be hard to maintain). I'm totally open to alternate paths here, and would also like to talk through concerns on the late-generation from C (as long as it can be made to be hermetic).

ajor:
Ah, makes sense. 

The concern I have with late-compilation of the stdlib is just the impact it'd have on the startup time of scripts. As the stdlib becomes bigger this will become more of an issue. Thoughts on some possible solutions:
- Caching (as suggested above). I have a strong aversion to caching things on the filesystem for a few of reasons:
  - It can lead to security issues (see our [CVE](https://nvd.nist.gov/vuln/detail/CVE-2024-2313)).
  - I like bpftrace to "leave no trace". I often plop a statically-compiled bpftrace executable onto a machine, run a couple of commands and then delete the executable. Having to delete cached files is something extra to remember.
  - Some systems might not have a writable filesystem. We'd have to take this into account in the caching system and fall back to a potentially slow compilation on every run.
- Selectively compile only required parts of stdlib:
  - If a script is simply `fentry:foo { print("hi") }`, then bpftrace could avoid compiling the stdlib entirely and the startup remains fast.
  - A more advanced system could determine the exact stdlib functions required by a script and only compile them.
  - Disadvantages: adds complexity and could potentially still be slow if we had to compile some very large stdlib functions (how slow? I don't know).
- Rely on [LLVM IR backwards compatibility](https://llvm.org/docs/DeveloperPolicy.html#ir-backwards-compatibility)
  - They claim that bitcode compiled with LLVM 3.0 can still be loaded by LLVM 21. I don't know what exactly they mean by this though, as they say that older features can be ignored by newer versions.
  - Don't know if this would work for us.
  - Might also be difficult to make work with Buck depending on the versions available.
-  Complex build processes. Some kind of setup like this could ensure that we compile the stdlib with the exact same LLVM version as used by codegen:
    1. Compile ClangBuilder with the available libLLVM. Make this a standalone executable rather than something we include in bpftrace.
    2. Add a build rule to compile the stdlib using the just-built ClangBuilder.
    3. Embed the stdlib bitcode into bpftrace.

My favourite is the last approach as it pushes all stdlib compilation to bpftrace compile-time, but maybe it'd be trickier to implement than I'm imagining.

Since the stdlib is so small right now, I'm guessing that it's pretty fast to compile (I'm curious if you've got any numbers), so a simple solution will probably do for a little while.

amscanne:
I've done some basic testing, and it is currently <2ms for a trivial import. It is also hermetic, and does not search any paths on the host system, just the internal VFS. (For reference the clang parser for a trivial program is ~19ms on the same machine.) I'm sure this will go up, but I think this is fine for now and lets us explore the space a little bit.

amscanne:
Summarizing, we will allow inline compiling for now, start adding things and see how it works/falls over. The "complex build processes" alternative will work for the cases we care about, but it's worth seeing the problems we experience before over-engineering that. There are advantages to being able to import `.bpf.c` directly.

src/ast/passes/clang_build.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+  // We create a temporary pipe that we can use to splurp the output,
+  // since the ClangDriver API is framed in terms of filenames. Perhaps
+  // we could use the internals here, but that carries other risks.
+  auto pipefds = create_pipe();
+  if (!pipefds) {
+    return pipefds.takeError();
+  }
+
+  // Create the compiler invocation.
```
jordalgo: Nit: Not a terrible place to explain why we need the `-02` or link to something that explains it.

src/ast/passes/clang_build.cpp
```diff
@@ -0,0 +1,212 @@
...
+  // Generate the object file, which should include the required BTF
+  // debug information. This also generates the module as a
+  // side-effect, which is what we actually extract for linking.
+  std::unique_ptr<clang::CodeGenAction> action =
+      std::make_unique<clang::EmitObjAction>();
+  if (!ci.ExecuteAction(*action)) {
+    // This is likely a build failure, we can surface this directly
+    // into the user context. We first highlight the location of the
+    // original import, then include the C message as a "hint".
+    auto &e = obj.node.addError();
```
jordalgo, line 171: Should we differentiate here from build errors from user-provided bpf.c files and our internal standard libraries?

src/ast/passes/clang_build.cpp
```diff
@@ -0,0 +1,212 @@
...
+    e << "failed to build";
+    e.addHint() << errstr;
+    return OK();
+  }
+  if (!errstr.empty()) {
+    // If the compilation didn't fail, then these weren't errors but we
+    // can surface them as compilation warnings.
+    auto &e = obj.node.addWarning();
```
jordalgo, line 179: Related to above ^ - I'm sure we'll try to sort out all build warnings for the stdlib but I don't think users should see these if they come from the stdlib.

amscanne: We shouldn't ship C code that generates warnings :) I think that is the simplest and best possible here, rather than trying to introduce some mechanism that separates those.

src/ast/passes/clang_build.cpp
amscanne: See below: I think we should just not have warnings in our own stdlib. I wouldn't want to suppress them or treat them differently, since I think they would be easy to miss in that case.

src/ast/passes/clang_build.cpp
amscanne: Done.

PR #4229, opened 2025-06-03
CMake: Fix dependency tracking for "embed" rule

ajor:
Have it depend on source files to trigger a rebuild when they change. The dependency on the CMake target does not trigger rebuilds, but is left to enforce ordering.

The dependency chain is roughly as follows (using data_source/BTF as an example but applies to all targets and formats):
1. data_source.c
2. data_source.o
3. data_source.btf
4. data_source_btf.hex
5. data_source_btf.h
6. bpftrace_test

An update to (1) triggers rebuilds of (2) and (3), but not (4) or later.

This change makes it so that (4) and also later targets are rebuilt when the source files change.

Tested with Make and Ninja build systems.

### DIFF
--- a/cmake/Embed.cmake
+++ b/cmake/Embed.cmake
@@ -23,7 +23,7 @@ function(embed NAME SOURCE)
     OUTPUT ${ARG_HEX}
     COMMAND ${XXD} -i < ${SOURCE} > ${ARG_HEX}
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-    DEPENDS ${ARG_DEPENDS}
+    DEPENDS ${SOURCE} ${ARG_DEPENDS}
     VERBATIM
   )
   add_custom_command(


amscanne (approved):
I struggled with the way the custom commands are embedded in the dependent rules. Does this cause the same commands to be generated multiple times in different rules? E.g. the custom commands generated files derived from `data_source.o` can potentially stomp on each other.

I *think* the canonical way to support this is to depend on the genrule (to avoid ordering) and the original source (to retrigger). E.g.

```
embed(
  data_source_btf
  data_source.btf
  OUTPUT  data_source_btf.h
  VAR     btf_data
  DEPENDS data_source data_source.c
)
```

But if this works and expresses the expected result, then LGTM!

ajor:
Ah yeah, I see what you mean about the custom command dependencies. Just pasting this for reference: https://cmake.org/cmake/help/latest/command/add_custom_command.html

> Do not list the output in more than one independent target that may build in parallel or the instances of the rule may conflict. Instead, use the [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target) command to drive the command and make the other targets depend on that one.

I haven't noticed any conflicts with my file-based dependencies. I'm not sure what the symptoms would be though.

What I don't understand is why the old dependency tracking based solely on a custom target didn't work.

amscanne:
> What I don't understand is why the old dependency tracking based solely on a custom target didn't work.

Yeah, depending on the output of an `add_custom_command` is a foot-gun, because then if someone else depends on it, you'll get multiple rules that stomp on each other (generating the `add_custom_command` themselves, since that's how it works). Since the files to be embedded here are the result of custom commands, I believe that adding that dependency does add this potential foot-gun.

The recommendation is to depend on the *rule* for ordering constraints (e.g. have these depends on `base`), and then just depend on the specific source files (e.g. `base.c`) in order to get the rule to re-evaluate when it changes. Basically you're constructing a separate build graph that enforces the file will be regenerated as a side-effect before you need it (gross), and that your rule will retrigger when it needs to. This is why I had the explicit `ARG_DEPENDS` there, and I would recommend just adding `base.c` to the derived rules to trigger that rebuild. However, if you're confident that it's correct (and hopefully not creating a foot-gun), then by all means proceed :)

ajor:
Updated to depend on both the source file and the custom target.

Just for my own future reference, the reason the target dependency wasn't triggering rebuilds is explained in the [add_custom_command : DEPENDS](https://cmake.org/cmake/help/latest/command/add_custom_command.html) docs:

> If the argument is the name of a target (created by the [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target), [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable), or [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) command) a target-level dependency is created to make sure the target is built before any target using this custom command. Additionally, if the target is an executable or library, a file-level dependency is created to cause the custom command to re-run whenever the target is recompiled.

i.e. targets created with `add_executable` and `add_library` create both ordering and file-level dependencies (and trigger rebuilds). Targets created with `add_custom_target` only do ordering.

PR #4234, opened 2025-06-06
tools/killsnoop.bt: display signal name string

Rtoax:
Printing the signal name directly is more intuitive than printing the signal value. However, the real-time signal is relatively complex, so it is omitted and the signal value is printed directly.

```
$ sudo ./killsnoop.bt 
Attached 8 probes
Tracing kill(),tkill(),tgkill() signals... Hit Ctrl-C to end.
TIME                 PID COMM                    SIG     TPID RESULT
17:09:33.259550   592754 smbd                SIGKILL   592757      0
17:09:33.270556   592751 systemctl           SIGTERM   592752      0
17:09:33.270567   592751 systemctl           SIGCONT   592752      0
```


tools/killsnoop.bt
```diff
@@ -29,8 +29,49 @@
...
+	@SIG_STR[28] = "SIGWINCH";
+	@SIG_STR[29] = "SIGIO";
+	@SIG_STR[30] = "SIGPWR";
+	@SIG_STR[31] = "SIGSYS";
+
+	/**
+	 * The real-time signal is relatively complex, so it is omitted and the
+	 * signal value is printed directly.
+	 *
+	 * POSIX SIGRTMIN is 34 or 35, see signal(7)
```
jordalgo, line 70: Ah so you mean it can be 32, 34, or 35?

tools/killsnoop.bt
```diff
@@ -47,9 +88,20 @@ tracepoint:syscalls:sys_exit_tkill,
 tracepoint:syscalls:sys_exit_tgkill
 /@tpid[tid]/
 {
-	printf("%-15s %8d %-16s %4d %8d %6d\n",
-		strftime("%H:%M:%S.%f", nsecs),
-		pid, comm, @tsig[tid], @tpid[tid], args.ret);
+	printf("%-15s %8d %-16s", strftime("%H:%M:%S.%f", nsecs), pid, comm);
+	if (@SIG_STR[@tsig[tid]] == "") {
```
jordalgo:
```suggestion
	if (!has_key(@SIG_STR, @tsig[tid])) {
```

amscanne: Thinking out loud, we had discussed previously improving the options around maps to allow you to use the different lookup semantics effectively. Go has two forms of this, `x = m[...]` or `x, ok = m[...]` where `ok` is a boolean that indicates whether the value was there or not. (And x would be the default zero value in that case.) That gets lowered to different intrinsics, depending on which is used.

Obviously this is a nice improvement as is, but it would be cool if we had:

```
$name, $exists = @SIG_STR[@tsig[tid]];
...
```

tools/killsnoop.bt
Rtoax: Yes, for example:

- linux kernel SIGRTMIN is 32 https://elixir.bootlin.com/linux/v6.15.1/source/arch/x86/include/uapi/asm/signal.h#L61
- POSIX SIGRTMIN is 34 or 35, see signal(7)
  - glibc is 34: sysdeps/unix/sysv/linux/internal-signals.h:#define RESERVED_SIGRT  2
  - musl-libc is 35: https://git.musl-libc.org/cgit/musl/tree/src/signal/sigrtmin.c

If directly displaying SIGRTMIN would cause confusion, can we display "SIGNAL-32" instead?

tools/killsnoop.bt
Rtoax: @jordalgo Use `has_key`, thanks. @amscanne Thanks.

jordalgo: @amscanne That's not bad but feels a bit premature for bpftrace. We'd discussed trying to keep the syntax as simple as possible (at least in the near term) and that feel a bit much. I do like that it would save a map lookup call.

On an even separate tangent, this PR got me thinking about `const`s and how we can save hash map lookups if we converted `@SIG_STR` to a global object of some sort or even a per-cpu hash.

tools/killsnoop.bt
```diff
@@ -4,17 +4,17 @@
  *		For Linux, uses bpftrace and eBPF.
  * 
  * Example of usage:
- * 
- * # ./killsnoop.bt
```
jordalgo, line 8: Nit: let's bring back the invocation comment. It indicates this has no positional params and can be invoked directly because of the shebang.

Rtoax: Sorry, oops. Please review. thanks

jordalgo (approved):
Thanks for the contribution!

PR #4237, opened 2025-06-06
Add CLI opts syntax

jordalgo:
This is the first PR in a series that will
add a syntax that allows users to specify
named command line arguments rather than
relying on positional params.

Issue:
https://github.com/bpftrace/bpftrace/issues/1945

##### Checklist

### DIFF
--- a/src/lexer.l
+++ b/src/lexer.l
@@ -165,6 +165,7 @@ oct_esc  [0-7]{1,3}
 "unroll"                { return Parser::make_UNROLL(yytext, driver.loc); }
 "while"                 { return Parser::make_WHILE(yytext, driver.loc); }
 "config"                { return Parser::make_CONFIG(yytext, driver.loc); }
+"opts"                  { return Parser::make_OPTS(yytext, driver.loc); }
 "for"                   { return Parser::make_FOR(yytext, driver.loc); }
 "return"                { return Parser::make_RETURN(yytext, driver.loc); }
 "continue"              { return Parser::make_CONTINUE(yytext, driver.loc); }
--- a/src/parser.yy
+++ b/src/parser.yy
@@ -121,6 +121,7 @@ void yyerror(bpftrace::Driver &driver, const char *s);
 %token <std::string> PARAM "positional parameter"
 %token <uint64_t> UNSIGNED_INT "integer"
 %token <std::string> CONFIG "config"
+%token <std::string> OPTS "opts"
 %token <std::string> UNROLL "unroll"
 %token <std::string> WHILE "while"
 %token <std::string> FOR "for"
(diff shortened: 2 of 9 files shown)


src/parser.yy
```diff
@@ -220,8 +224,8 @@ start:          START_PROGRAM program END { driver.result = $2; }
                 ;
 
 program:
-                c_definitions config imports map_decl_list macros body {
-                    $$ = driver.ctx.make_node<ast::Program>($1, $2, std::move($3), std::move($4), std::move($5), std::move($6.second), std::move($6.first), @$);
+                c_definitions config opts imports map_decl_list macros body {
```
amscanne, line 227: We probably need to allow for flexibility in the ordering of these blocks. Or at least grouping them, e.g. (c_definitions|config|opts|imports)*(map_decl_list|macros|body)*

src/lexer.l
```diff
@@ -165,6 +165,7 @@ oct_esc  [0-7]{1,3}
 "unroll"                { return Parser::make_UNROLL(yytext, driver.loc); }
 "while"                 { return Parser::make_WHILE(yytext, driver.loc); }
 "config"                { return Parser::make_CONFIG(yytext, driver.loc); }
+"opts"                  { return Parser::make_OPTS(yytext, driver.loc); }
```
amscanne, line 168: This will need to be added as an extern keyword in the parser.

tests/parser.cpp
```diff
@@ -2747,6 +2747,115 @@ BPFTRACE_STACK_MODE=perf; i:s:1 { exit(); }
...
+Program
+ opts
+  opt
+   =
+    var: blah
+    int: 5
+ BEGIN
+)");
+
+  test("opts = [{ blah = 5; type = string }] BEGIN {}", R"(
```
amscanne, line 2762: Thinking out loud. I'm not sure I understand the general syntax. Why not just have a list, e.g.

```
opts = {
   foo = "bar"; // type is a string, default is "bar"
   zap = 5; // type is an integer, default is 5
}
```

You could special case `foo = 0` and accept `--foo` to set to 1 (and `foo = 1` to accept `--no-foo`, if you wanted). Nothing prevents extending beyond this, but I'd just be curious to understand more advanced user examples.

src/parser.yy
jordalgo: I was planning on doing that in a follow-up PR.

src/lexer.l
jordalgo: Ah good catch.

tests/parser.cpp
jordalgo: Please see the attached RFC. We already have cases we want to support that require more keys for each opt e.g. 'required' (which may not have a default) and 'short_name'. I also think the shorter syntax does make it harder to extend this in the future.

amscanne: Got it. I thought about this today, and put a bit of a straw man there as a simplifying alternative.

My 2c is that it would be good to have some concrete examples, look at existing scripts a bit, and maybe let this stack up a bit to land. I don't object to the change, I just think it's worthwhile to be specifically cautious for new syntax.

jordalgo:
Shelving this for now in favor of `getopt`

PR #4238, opened 2025-06-06
codegen: Fix __data_loc reads

tnovak:
A __data_loc field is an (offset, len) pair; its value is determined by adding the offset (in bytes) to the address of the tracepoint struct.

Commit 06a8972510 ("codegen: use preserve_static_offset intrinsic") replaced the add instruction with getelementptr, but used i32 as the GEP type without adjusting the offset. This causes the value to be incorrectly set to `ctx + 4*offset`.

For example, this currently prints an empty or incomplete filename:

```
  bpftrace -e 't:sched:sched_process_exec { print(str(arg.filename)) }'
```

Fix it by changing the getelementptr type argument to i8.

##### Checklist


amscanne (approved):
Nice catch. I'm looking through the tests to understand why nothing failed, but perhaps all the tracepoint runtime tests are using the first argument? Did you want to add your simple example as a runtime test (e.g. in tests/runtime/...)? Otherwise, I can follow up and add this.

amscanne:
> Nice catch. I'm looking through the tests to understand why nothing failed, but perhaps all the tracepoint runtime tests are using the first argument? Did you want to add your simple example as a runtime test (e.g. in tests/runtime/...)? Otherwise, I can follow up and add this.

Correcting myself, that doesn't matter. I think I understand how this got through. We have some tests for arguments (e.g. https://github.com/bpftrace/bpftrace/blob/535f5ad8738cfdf4058ea20f98fc0b8f39ac9db2/tests/runtime/call#L92), but presumably there's nothing specifically using a `__data_loc` field. Having a similar case with a `__data_loc` field should catch this path, let me know if you want to add with this change or I can send separately.

jordalgo:
+1 good catch! And yes, we should absolutely have a runtime test for this.

jordalgo (approved):
LGTM. Thanks!

PR #4242, opened 2025-06-09
Allow interleaving of macros and map decls

jordalgo:
This changes allows the following AST classes
to be interleaved at the top level:
- probes
- subprogs
- macros
- map declarations

This allows users to more strategically place
these nodes in their bpftrace program instead
of forcing them to be at the beginning.

##### Checklist


src/ast/ast.h
```diff
@@ -216,6 +216,19 @@ class Statement : public VariantNode<ExprStatement,
...
+class Probe;
+class Subprog;
+
+class RootStatement
+    : public VariantNode<Probe, Subprog, Macro, MapDeclStatement> {
+public:
+  using VariantNode::VariantNode;
+  RootStatement() : RootStatement(static_cast<Probe *>(nullptr)) {};
+};
+using RootStatements = std::vector<RootStatement>;
```
amscanne, line 230: I don't specifically have a program with this construction, but it's a bit weird because `RootStatement` is purely a byproduct of the parser AFAICT.

One thought that I had in the past is factoring the driver/context to hold a program as a non-pointer reference, e.g.

```
class ASTContext {
   ...
   Program root;
   ...
}
```

And then just have the parser add to the various blocks as it encounters the top-level statements, rather than needing the intermediate node for Bison. This also solves the issue of whether root is `nullptr`.

I'm good to approve as is, I'm just raising an alternative that you may want to consider.

jordalgo: > factoring the driver/context to hold a program as a non-pointer reference

Ah neat. Let me look into that in a follow-up PR.

PR #4244, opened 2025-06-09
Fix kstack/ustack on big-endian systems

viktormalik:
Commits 3b2cda69 ("Replace 'bpf_get_stackid' (#3060)") and b5db2533 ("Add nr_stack_frames to stack key (#3113)") changed the layout of the internal stack map. Now, the key consists of two parts - hashed stack id and the number of stack frames. The problem is that the number of stack frames was originally a 32-bit integer which caused various problems, which were fixed by 14dd0e3 ("Fix size mismatch for `stack_key`s") and 3e66227c ("Remove padding from stack struct").

Unfortunately, problems still exist on big-endian systems, causing errors like:

    # bpftrace -e 'fentry:vfs_read { print(kstack); exit() }'
    Attached 1 probe
    ERROR: failed to look up stack id: 2027306984000431560 stack length: 6 (pid -1): BPF map operation lookup failed: No such file or directory [map = stack_bpftrace_127]

The problem here is that there are two places where the number of stack frames is treated as 32-bit int:
- when defining the BPF map key type (the key has 12 bytes)
- in the internal stack_key type which is used for retrieving values from the internal stack map

Fix the type in both places.

Note: this only affects big-endian systems as for little-endian, the number of stack frames always fits the first half of the 64-bit value reserved for the stack key so it doesn't matter if it's treated as 32-bit in some places.

##### Checklist


viktormalik:
@jordalgo I'm afraid we'll need to craft a new release as `kstack`/`ustack` are currently broken on big-endian systems in the 0.23 release.

I can prepare the release if you want.

jordalgo:
> I can prepare the release if you want.

Up to you. I also don't mind seeing as how I originally broke it 🔧

viktormalik:
> > I can prepare the release if you want.
> 
> Up to you. I also don't mind seeing as how I originally broke it 🔧

No problem, I'll take this one and ping you when it's ready for review. Should be soon.

PR #4254, opened 2025-06-10
add support for init namespace to `pid` and `tid`

### LINKED ISSUE #4253: Support getting pids and tids from the initial namespace

Before #3428, we can use builtins `pid` and `tid` to get the global PIDs and TIDs. They are useful for using tracepoints like `sched_switch`, whose arguments are in the initial namespace.

Add an argument to `pid` and `tid` to specify whether to use the initial PID namespace.

A configuration variable for switching the behavior globally.

MikeWalrus:
fixes #4253.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@ and this project adheres to
   - [#4126](https://github.com/bpftrace/bpftrace/pull/4126)
 - For loops now support `break` and `continue`
   - [#4250](https://github.com/bpftrace/bpftrace/pull/4250)
+- Add `pid` and `tid` functions for choosing between the initial or the current namespace
+  - [#4254](https://github.com/bpftrace/bpftrace/pull/4254)
 #### Changed
 - `-p` CLI flag now applies to all probes (except BEGIN/END)
   - [#3800](https://github.com/bpftrace/bpftrace/pull/3800)
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -24,6 +24,14 @@ config = {
 }
 ```
 
+## Versions 0.22.x (or earlier) to 0.24.x (or later)
+After https://github.com/bpftrace/bpftrace/pull/3428 `pid` and `tid` are no
+longer from the initial namespace.
+
+https://github.com/bpftrace/bpftrace/pull/3428 introduced `pid(init)` and
+`tid(init)` which have the old behavior.
+Unfortunately, there are no equivalent workarounds in 0.23.x.
+
 ## Versions 0.21.x (or earlier) to 0.22.x (or later)
 
 ### Added block scoping for scratch variables
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1988,7 +1988,7 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | `pid`
 | uint32
 | 4.2
-| get_current_pid_tgid / bpf_get_ns_current_pid_tgid
+| get_current_pid_tgid / get_ns_current_pid_tgid
 | Process ID of the current thread (aka thread group ID), as seen from the PID namespace of bpftrace
 
 | `probe`
@@ -2018,7 +2018,7 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | `tid`
 | uint32
 | 4.2
-| get_current_pid_tgid / bpf_get_ns_current_pid_tgid
+| get_current_pid_tgid / get_ns_current_pid_tgid
 | Thread ID of the current thread, as seen from the PID namespace of bpftrace
 
 | `uid`
@@ -2140,6 +2140,14 @@ Tracing block I/O sizes > 0 bytes
 | <<functions-cat, `cat`>>
 | Print file content
 | Async
+
+| <<functions-pid, `pid`>>
+| Process ID of the current thread
+| Sync
+
+| <<functions-tid, `tid`>>
+| Thread ID of the current thread
+| Sync
 
 | <<functions-cgroupid, `cgroupid`>>
 | Resolve cgroup ID
@@ -2349,6 +2357,28 @@ tracepoint:syscalls:sys_enter_execve {
 55f683edf000-55f683ee2000 rw-p 00021000 08:01 1843399                    /usr/bin/ls
 55f683ee2000-55f683ee3000 rw-p 00000000 00:00 0
 ----
+
+[#functions-pid]
+=== pid
+.variants
+* `uint32 pid([curr_ns|init])`
+
+Returns the process ID of the current thread.
+Defaults to `curr_ns`.
+
+- `pid(curr_ns)` - The process ID as seen from the PID namespace of bpftrace.
+- `pid(init)` - The process ID as seen from the initial PID namespace.
+
+[#functions-tid]
+=== pid
+.variants
+* `uint32 tid([curr_ns|init])`
+
+Returns the thread ID of the current thread.
+Defaults to `curr_ns`.
+
+- `tid(curr_ns)` - The thread ID as seen from the PID namespace of bpftrace.
+- `tid(init)` - The thread ID as seen from the initial PID namespace.
 
 [#functions-cgroupid]
 === cgroupid
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1988,8 +1988,8 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | `pid`
 | uint32
 | 4.2
-| get_current_pid_tgid
-| Process ID of the current thread (aka thread group ID), as seen from the init namespace
+| get_current_pid_tgid / bpf_get_ns_current_pid_tgid
+| Process ID of the current thread (aka thread group ID), as seen from the PID namespace of bpftrace
 
 | `probe`
 | string
@@ -2018,8 +2018,8 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | `tid`
 | uint32
 | 4.2
-| get_current_pid_tgid
-| Thread ID of the current thread, as seen from the init namespace
+| get_current_pid_tgid / bpf_get_ns_current_pid_tgid
+| Thread ID of the current thread, as seen from the PID namespace of bpftrace
 
 | `uid`
 | uint64
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -8,6 +8,7 @@
 #include "bpftrace.h"
 #include "globalvars.h"
 #include "log.h"
+#include "types.h"
 #include "util/exceptions.h"
 #include <filesystem>
 #include <llvm/IR/DataLayout.h>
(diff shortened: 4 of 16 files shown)


tests/runtime/pid_namespace
```diff
@@ -2,17 +2,18 @@
 # They should both be "1" as bpftrace is the first and only process in this
 # temporary namespace.
 NAME pid_tid_inside
-RUN unshare --fork --pid --mount-proc {{BPFTRACE}} -e 'BEGIN { printf("pid: %d, tid: %d\n", pid, tid); exit() }'
-EXPECT pid: 1, tid: 1
+RUN unshare --fork --pid --mount-proc {{BPFTRACE}} -e 'BEGIN { printf("pid: %d, gpid: %d, tid: %d, gtid: %d\n", pid, gpid, tid, gtid); exit() }'
+EXPECT_REGEX ^pid: 1, gpid: \d+, tid: 1, gtid: \d+$
+EXPECT_REGEX_NONE ^pid: 1, gpid: 1, tid: 1, gtid: 1$
```
amscanne: Nice!

src/ast/irbuilderbpf.cpp
```diff
@@ -74,10 +74,10 @@ libbpf::bpf_func_id IRBuilderBPF::selectProbeReadHelper(AddrSpace as, bool str)
 // It represents the inode of the initial (global) PID namespace
 constexpr uint32_t PROC_PID_INIT_INO = 0xeffffffc;
 
-Value *IRBuilderBPF::CreateGetPid(const Location &loc)
+Value *IRBuilderBPF::CreateGetPid(const Location &loc, bool force_initial)
 {
   const auto &pidns = bpftrace_.get_pidns_self_stat();
-  if (pidns && pidns->st_ino != PROC_PID_INIT_INO) {
+  if (!force_initial && pidns && pidns->st_ino != PROC_PID_INIT_INO) {
```
amscanne: As a related aside (not an issue for this PR): the logic here is broken for any kind of AOT compilation. We are generating code that uses the inode of the pid namespace **during compile time**. We'd have to lift this into the runtime and a some kind of map/const global to have it work for those cases.

I will file an issue for this.

amscanne: #4256 filed for this specific issue.

jordalgo:
> Seems reasonable to me. Is the gpid short-hand for "global" pid? 

Yeah that was the idea. @MikeWalrus also suggested overloading pid/tid to be functions that accept a param that allows you to get them in the global/root namespaces. I'm not sure which I prefer TBH.

ajor:
I think I prefer something like `pid(xxx)` instead of an entirely separate builtin - it'd conceptually link the functionalities more closely and would be consistent with `nsecs`, which also switches between various different BPF helpers. Theoretically, with this approach we could also add a config option to change the default functionality of `pid` within a script (if that's a good idea). Any reason you had in mind to have them separate as builtins (`gpid`/`gtid`) @jordalgo?

I'm not sure what the argument to `pid()` would be though. Taking an inode number of a PID namespace might be too complicated/difficult for users to remember, so maybe `pid(init)`/`pid(curr_ns)`?

MikeWalrus:
> so maybe `pid(init)`/`pid(curr_ns)`?

I have an implementation of that as well. I'll just have to add some tests and docs on that branch.

jordalgo:
> Any reason you had in mind to have them separate as builtins

Not really anything specific. I think I agree with you that we should just overload pid/tid because this isn't a different concept rather just a different flavor of pid/tid. @MikeWalrus sorry for making you implement this other builtin path.

Just FYI: I think whatever arguments we have for pid/tid should not be in the parser/lexer. `init` and `current_ns` as an ident enum works for me for now.

CHANGELOG.md
```diff
@@ -55,6 +55,9 @@ and this project adheres to
   - [#4126](https://github.com/bpftrace/bpftrace/pull/4126)
 - For loops now support `break` and `continue`
   - [#4250](https://github.com/bpftrace/bpftrace/pull/4250)
+- Add `pid` and `tid` functions for choosing between the initial or the current
+  namespace
```
jordalgo: super nit: let's keep this on the same line

docs/migration_guide.md
```diff
@@ -24,6 +24,14 @@ config = {
 }
 ```
 
+## Versions 0.22.x (or earlier) to 0.24.x (or later)
+After https://github.com/bpftrace/bpftrace/pull/3428 `pid` and `tid` are no
+longer from the initial namespace.
+
+https://github.com/bpftrace/bpftrace/pull/3428 introduced `pid(init)` and
+`tid(init)` which have the old behavior.
+Unfortunately, there are no equivalent workarounds in 0.23.x.
```
jordalgo, line 33: I mean we COULD do a patch release (cc @viktormalik)

man/adoc/bpftrace.adoc
```diff
@@ -1988,8 +1988,8 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | `pid`
 | uint32
 | 4.2
-| get_current_pid_tgid
-| Process ID of the current thread (aka thread group ID), as seen from the init namespace
+| get_current_pid_tgid / bpf_get_ns_current_pid_tgid
```
jordalgo:
```suggestion
| get_current_pid_tgid / get_ns_current_pid_tgid
```

man/adoc/bpftrace.adoc
```diff
@@ -2018,8 +2018,8 @@ For string arguments in an action block use the `str()` call to retrieve the val
 | `tid`
 | uint32
 | 4.2
-| get_current_pid_tgid
-| Thread ID of the current thread, as seen from the init namespace
+| get_current_pid_tgid / bpf_get_ns_current_pid_tgid
```
jordalgo:
```suggestion
| get_current_pid_tgid / get_ns_current_pid_tgid
```

CHANGELOG.md
MikeWalrus: Done.

src/types.h
```diff
@@ -170,6 +173,7 @@ class SizedType {
   bool is_funcarg = false;
   bool is_btftype = false;
   TimestampMode ts_mode = TimestampMode::boot;
+  PidNamespaceMode pid_ns_mode = PidNamespaceMode::curr_ns;
```
jordalgo: I'm personally not a fan of adding any more to SizedType if we can avoid it. Can we not just check if the ident (or string) is one of the values we expect in semantic analyser and then we know if can only be one of two when we do the check again in codegen? Then we don't have to plumb this enum all the way through.

man/adoc/bpftrace.adoc
MikeWalrus: Also fixed the other occurrence.

man/adoc/bpftrace.adoc
MikeWalrus: Done.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1729,6 +1746,12 @@ If you're seeing errors, try clamping the string sizes. For example:
       call.addError() << "Failed to initialize sw_tai in "
                          "userspace. This is very unexpected.";
     }
+  } else if (call.func == "pid" || call.func == "tid") {
```
jordalgo, line 1743: why is this logic split from the added logic above?

src/types.h
MikeWalrus: I've removed the type to use ident directly. I can squash the two commits if needed.

src/ast/passes/semantic_analyser.cpp
MikeWalrus: Now this is for handling cases where the argument is in the wrong shape.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -839,6 +841,11 @@ ScopedExpr CodegenLLVM::visit(Builtin &builtin)
   }
 }
 
+bool CodegenLLVM::shouldForceInitPidNs(const ExpressionList &args)
```
jordalgo: Nit: This could just be a free function at the top of this file and not a method.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -685,6 +690,7 @@ static bool IsValidVarDeclType(const SizedType &ty)
     case Type::strerror_t:
     case Type::none:
     case Type::timestamp_mode:
+    case Type::pid_namespace_mode:
```
jordalgo: Leftover from the last change?

src/ast/passes/codegen_llvm.cpp
MikeWalrus: I agree.

src/types.h
jordalgo: Great, thanks. We can squash when we merge.

man/adoc/bpftrace.adoc
```diff
@@ -2350,6 +2358,28 @@ tracepoint:syscalls:sys_enter_execve {
 55f683ee2000-55f683ee3000 rw-p 00000000 00:00 0
 ----
 
+[#functions-pid]
+=== pid
+.variants
+* `uint32 pid([PidNamespaceMode mode])`
```
jordalgo: ah sorry, this should be updated as well as `PidNamespaceMode` no longer exists. 

Maybe just `uint32 pid([curr_ns|init])` or `uint32 pid([ident])` to keep with the type style.

jordalgo:
One small adoc update and then I think this is good to go.

man/adoc/bpftrace.adoc
MikeWalrus: Yes, I was going to ask what I should change it to.

jordalgo: See my edited comment.

MikeWalrus: Yeah, that's what I thought as well.

docs/migration_guide.md
viktormalik: Well, I'm not a big fan of doing patch releases all the time, however, this looks like it could be a regression.

IIUC, when running bpftrace from a non-init namespace, there's no way to obtain PID/TID from the init namespace, right? Is that something that we expect users to do? I thought that `pid`/`tid` always working on the init namespace was viewed as a bug (for containerized environments) which was fixed by #3428.

jordalgo: > IIUC, when running bpftrace from a non-init namespace, there's no way to obtain PID/TID from the init namespace, right?

Correct. 

> Is that something that we expect users to do?

I'm not sure, there is at least one user who sees this as a regression 😉  Any thoughts on this @MikeWalrus? How niche is your use case.

jordalgo (approved):

docs/migration_guide.md
MikeWalrus: I don't know about others, but when dealing with scheduler-related tracing, there's going to be some global TIDs in the kfunc arguments or tracepoints. It's nice to be able to figure out who's who and calculate stuff even if bpftrace is run within something like a container.

PR #4255, opened 2025-06-10
pass: add `LoopReturn` pass

amscanne:
Stacked PRs:
 * __->__#4255

--- --- ---

### pass: add `LoopReturn` pass

Since all the context building supports scratch variables, supporting
loop returns involved injecting logic directly into the program
structure to support chain-breaking out of loops. The variables use
non-`$` prefixes, and therefore cannot collide with user variables.

They will not be injected unless they are needed.

Note that this commit also tweaks the address space for the context,
ensuring that it is correctly marked as `none` (or BPF).

### DIFF
--- a/src/ast/passes/loop_return.cpp
+++ b/src/ast/passes/loop_return.cpp
@@ -1,0 +1,232 @@
+#include "ast/passes/loop_return.h"
+#include "ast/ast.h"
+#include "ast/visitor.h"
+
+namespace bpftrace::ast {
+
+namespace {
+
+class LoopReturn : public Visitor<LoopReturn> {
+public:
+  LoopReturn(ASTContext &ast) : ast_(ast) {};
+
+  using Visitor<LoopReturn>::visit;
+
+  template <typename T>
+  void inject(T &node);
+
+  void visit(Subprog &subprog);
+  void visit(Probe &probe);
+  void visit(For &for_loop);
+  void visit(While &while_loop);
+  void visit(Statement &stmt);
+
+private:
+  Variable *get_return_var(const Node &node)
+  {
+    return ast_.make_node<Variable>(node.loc, "__return_value");
+  }
+  Variable *get_return_set_var(const Node &node)
+  {
+    return ast_.make_node<Variable>(node.loc, "__return_value_set");
+  }
+
+  ASTContext &ast_;
+  int loop_depth_ = 0;
+  std::optional<bool> loop_return_;
+};
+
+} // namespace
+
+void LoopReturn::visit(For &for_loop)
+{
+  loop_depth_++;
+  Visitor<LoopReturn>::visit(for_loop);
+  loop_depth_--;
+}
+
+void LoopReturn::visit(While &while_loop)
+{
+  loop_depth_++;
+  Visitor<LoopReturn>::visit(while_loop);
+  loop_depth_--;
+}
+
+void LoopReturn::visit(Statement &stmt)
+{
+  // Recursively visit the block statements.
+  Visitor<LoopReturn>::visit(stmt);
+
+  // First see if we have any loop statements, because we will inject an
+  // additional return into those statements if needed. This may be rewritten
+  // below, but only if this is within an inner loop.
+  //
+  // Consider that this takes:
+  //
+  // for (...) { }
+  //
+  // And turns it into:
+  //
+  // { for (...) { } if (__return_value_set) { return __return_value; } }
+  //
+  // The return may be subsequently rewritten below, if inside a loop.
+  bool is_loop = stmt.is<For>() || stmt.is<While>();
+  if (is_loop && loop_return_.has_value()) {
+    std::vector<Statement> ret_stmts;
+    if (loop_depth_ > 0) {
+      // We're still in a loop, so just break.
+      ret_stmts.emplace_back(
+          ast_.make_node<Jump>(stmt.node().loc, JumpType::BREAK));
+    } else if (loop_return_.value()) {
+      // The value was set within the loop, so return it.
+      ret_stmts.emplace_back(ast_.make_node<Jump>(stmt.node().loc,
+                                                  JumpType::RETURN,
+                                                  get_return_var(stmt.node())));
+    } else {
+      // The value was not set within the loop, so skip it.
+      ret_stmts.emplace_back(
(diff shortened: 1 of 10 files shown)


amscanne:
I will also add a test case for nested loops before submission.

src/ast/passes/loop_return.cpp
```diff
@@ -0,0 +1,164 @@
...
+          ast_.make_node<Jump>(JumpType::BREAK, Location(jmp->loc)));
+      auto *block = ast_.make_node<Block>(
+          std::vector<Statement>(std::move(ret_stmts)), Location(jmp->loc));
+      stmt = Statement(block);
+    }
+  }
+}
+
+void LoopReturn::visit(Probe &probe)
+{
```
amscanne, line 190: I need to do two things here:

* Clear `loop_return_`
* Apply the same logic below to `SubProg` (factoring this out)

I should also add add tests for these cases, since nothing caught these problems.

tests/runtime/for
```diff
@@ -92,6 +92,10 @@ NAME map with continue
 PROG begin { @map["foo"] = 1; @map["bar"] = 2; print("start"); for ($i : @map) { if ($i.0 == "bar") { continue; } print($i.0); }  }
 EXPECT_REGEX .*start\nfoo\n\n
 
+NAME map with return
```
jordalgo, line 95: I think this test can be simplified with a range based for loop and just a single `begin` probe. Also you need an EXPECT_NONE for the print statement after the return in the loop.

tests/runtime/for
```diff
@@ -136,6 +140,10 @@ NAME range with continue
 PROG begin { for ($i : 1..4) { if ($i == 2) { continue; } print($i); }  }
 EXPECT_REGEX .*\n1\n3\n\n
 
+NAME range with nested return
```
jordalgo, line 144: Same comment as above.

tests/control_flow_analyser.cpp
```diff
@@ -67,9 +67,10 @@ TEST(control_flow_analyser, if_without_else)
 
 TEST(control_flow_analyser, while_loop)
 {
-  // This is allowed, as we can detect that this will always return. However,
-  // the verifier may reject it.
-  test("fn test($x: int64): int64 { while ($x) { return 0; } }", 0);
+  // This was previously allowed, but is no longer allowed. All loops
+  // are treated equally, and therefore we require all paths to return
+  // appropriately ahead of any folding, etc.
```
jordalgo, line 72: By "return appropriately" you mean all branches must return the same type, yes? Right now this is a mix of int and void.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3088,7 +3088,7 @@ void SemanticAnalyser::visit(For &f)
   auto [iter, _] = for_vars_referenced_.try_emplace(&f);
   auto &collector = iter->second;
   for (const Variable &var : collector.nodes()) {
-    ctx_types.push_back(CreatePointer(var.var_type, AddrSpace::kernel));
+    ctx_types.push_back(CreatePointer(var.var_type, AddrSpace::none));
```
jordalgo, line 3090: Why this change?

src/ast/passes/loop_return.cpp
```diff
@@ -0,0 +1,232 @@
...
+void LoopReturn::inject(T &node)
+{
+  loop_return_.reset();
+  Visitor<LoopReturn>::visit(node);
+
+  if (loop_return_.has_value()) {
+    // Declare the return variables at the top of the probe. These actually
+    // need to supercede all the other statements in the block.
+    if (loop_return_.value()) {
+      auto *ret_var_decl = ast_.make_node<VarDeclStatement>(
```
jordalgo, line 198: Why do you need a decl here if you're not setting the type?

src/ast/passes/semantic_analyser.cpp
amscanne: This was wrong, the context arguments are in BPF and do not require probe reads.

I can split this into a separate commit if it's useful.

src/ast/passes/loop_return.cpp
amscanne: To ensure that the scope is correct. It will be first assigned in a subscope.

tests/runtime/for
amscanne: The double new line is printed only at the end. Ideally this is just a test probe.

I'll clean this up, it predates exiting automatically :)

tests/control_flow_analyser.cpp
amscanne: Right, it's not clear to me why this was ever allowed. It's weird because the comment above indicates that it's probably broken anyways.

src/ast/passes/semantic_analyser.cpp
jordalgo: That or just a note in the current commit.

src/ast/passes/loop_return.cpp
jordalgo: Ah got it.

jordalgo (approved):
LGTM. Also please update the table in the language doc for supported control flow statements for these loops:
- https://bpftrace.org/docs/release_024/language#for

src/ast/passes/loop_return.cpp
```diff
@@ -0,0 +1,232 @@
...
+  // additional return into those statements if needed. This may be rewritten
+  // below, but only if this is within an inner loop.
+  //
+  // Consider that this takes:
+  //
+  // for (...) { }
+  //
+  // And turns it into:
+  //
+  // { for (...) { } if (__return_value_ret) { return __return_value; } }
```
viktormalik: Shouldn't be?
```suggestion
  // { for (...) { } if (__return_value_set) { return __return_value; } }
```

viktormalik (approved):
Looks good, just one nit in a comment.

src/ast/passes/semantic_analyser.cpp
amscanne: Done.

src/ast/passes/loop_return.cpp
amscanne: Fixed.

PR #4258, opened 2025-06-10
Add: `getopt` function

jordalgo:
This function provides a way to get named command
line arguments passed to bpftrace after a double dash.

Example:
```
bpftrace -e 'BEGIN { print((getopt("aa", 10), getopt("bb", "hello"), getopt("cc"))); }' -- --cc --bb=bye
```
This prints (10, bye, 1)

The `getopt` function call is always replaced
before semantic analysis in the cli_opts pass
with a literal AST node (string, integer, or negative integer). This will be changed for AOT in a follow-up PR,
whereby we will make these global values that
get set and looked up at runtime.

##### Checklist


src/util/format.cpp
```diff
@@ -97,4 +98,13 @@ std::string hex_format_buffer(const char *buf,
   return str;
 }
 
+std::string to_lower(const std::string &original)
```
amscanne, line 101: IIRC we have this elsewhere (maybe it's `to_upper`?)? Can you refactor and standardize so we don't have duplication, maybe this just means a stacked change to centralize the existing `to_upper`/`to_lower` in this file.

src/util/str_to_bool.cpp
```diff
@@ -0,0 +1,19 @@
+#include "util/str_to_bool.h"
+#include "util/format.h"
+#include <string>
+
+namespace bpftrace::util {
+
+bool is_str_bool_truthy(const std::string &value)
```
amscanne: We have this functionality in the `ConfigParser<bool>`. I think as a separate change, can you generalize from `ConfigParser<...>` to `Parser<...>` and we can just have a `util/parsers.h` or `parsers/bool.h` with these implementation? Just trying to avoid a return to the old `util` where there's a lot of overlapping/duplicate functionality.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1655,6 +1655,8 @@ If you're seeing errors, try clamping the string sizes. For example:
       }
     }
     call.return_type = CreateUInt64();
+  } else if (call.func == "getopt") {
+    LOG(BUG) << "getopt function call should have been replaced at this point.";
```
amscanne: I don't think you need a dedicated `BUG` here as this should fail when it hits the final `else`. We should be okay with slowly eliminating function cases from the semantic analyzer, and not replacing everything with a call to `BUG` :)

src/ast/passes/cli_opts.cpp
```diff
@@ -0,0 +1,127 @@
...
+                     << named_args_.at(arg_name->value);
+  } catch (const std::out_of_range &) {
+    call->addError() << "Value for command line option '" << arg_name->value
+                     << "' is out of range. Got: "
+                     << named_args_.at(arg_name->value);
+  }
+}
+
+Pass CreateCLIOptsPass(
+    const std::unordered_map<std::string, std::string> &named_args)
```
amscanne, line 127: This is good for a POC, but I think the actual pass needs to be able to handle AOT properly.

Just as a possible suggestion: IMO, this pass could "desugar" the getopt function and replace with the use of scalar maps. Then it could return some metadata about those maps (including help message, default values, etc.):

```
class OptionMaps : public ast::State<"option-maps"> {
public:
   std::map<std::string, Option<std::string>> strings;
   std::map<std::string, Option<bool>> bools;
   std::map<std::string, Option<int64_t>> integers;
};
```

That metadata would be consumed by the resource analyzer / semantic analyzer, which would be able to associate types with those maps. It would also be consumed at the end (extracted from the pass) for the runtime to parse those arguments and set the values. Whether these actually materialize as maps and/or consts is totally an implementation decision later in the pipeline (because we're probably want to be able to optimize *all* read-only maps into const values, and just have that apply universally to these maps as well).

src/util/format.cpp
jordalgo: There is no `to_upper`. And this change already consolidates this logic with the `Config` class (just not in a separate commit -  I can split it out if desired).

src/util/str_to_bool.cpp
jordalgo: See above. ConfigParser calls this function now. I can change the name to parsers.h though.

src/ast/passes/cli_opts.cpp
jordalgo: Yeah, as per my commit message, AOT support will be in a follow-up PR but I'll try out what you suggested. I'll also work on a follow-up PR that automatically converts all read-only maps into const values.

jordalgo:
@amscanne I'm going to work on the map decl stuff for `getopt` in a separate PR as that's a bit more complicated. We have to create the map and then pre-populate it with the argument value (or default) after program load but before attachment (which isn't hard but we just don't have a mechanism for that yet -- I don't think).

src/util/format.cpp
amscanne: Oh, not sure why I didn't grok that. Amazing!

src/main.cpp
```diff
@@ -296,10 +297,39 @@ std::vector<std::string> extra_flags(
   return extra_flags;
 }
 
-void CreateDynamicPasses(std::function<void(ast::Pass&& pass)> add)
+struct Args {
+  std::string pid_str;
```
amscanne, line 315: Thinking out loud: I guess one downside to this specific proposal is that we did lose the unification with existing args (e.g. `opts.pid` as a thing). But honestly I think that with the pid filtering working as is, it would be worth thinking about that some more and seeing if it still makes sense.

src/ast/passes/cli_opts.cpp
amscanne: I don't think you should land it without working with AOT, IMO it's just part of the design, unless there's some urgent need. The optimization (converting read-only maps) can come later. I wouldn't block anything, I just think you won't flesh out the feature properly until you have those parts.

src/main.cpp
```diff
@@ -624,11 +628,63 @@ Args parse_args(int argc, char* argv[])
...
+    // We want to load positional params before the driver runs so they can be
+    // used inside attach point definitions.
+    std::string named_arg;
+    bool found_named_arg_start = false;
     while (optind < argc) {
-      args.params.emplace_back(argv[optind]);
+      auto new_arg = std::string(argv[optind]);
       optind++;
+
+      if (new_arg.starts_with("--")) {
```
amscanne: This seems backwards to me.

It seems that you are requiring it to be parsed as:

```
bpftrace <opts> mytool.bt first second third --foo=bar --bar=baz
```

But the following is how you would conventionally do this:

```
bpftrace <opts> mytool.bt --foo=bar --bar=baz first second third
```

In fact, I would not expect the `--foo=...` to actually parse as an option in that first case, they should be `$4` and `$5` respectively. If I do the natural thing (second case), then I would have an error like: `Multiple values for a single named argument are not yet supported` and be thoroughly confused.

I think there's strong conceptual alignment that what *getopt* should be doing is allowing the parsing of options from the effective "program arguments", and reducing the set of positional arguments to the "real" positional arguments. There are some choices there, but I don't think it should invert the traditional understanding of positional arguments vs. options and behavior the opposite of a classic `getopt`.

Here is a strawman for how to handle the ambiguity here that may be more natural, I'd be curious for feedback:

* I **doubt** that anyone is trying to use a positional argument of the form `--...`. Maybe we can just parse all "program arguments" (everything from `optind`) as potential named arguments. If they start with `--` they get parsed, otherwise positional. If there is an explicit `--`, then parsing stops and all subsequent arguments are positional. (This also lets you simplify a lot of the code below, since you don't need to remember the last arg, nor do you have the error edge cases.)
* The way to deal with `--foo bar` vs `--foo --bar` is simple: require the arguments to use `=`. I think it's justified since we don't know what they are when we're parsing, so just require the `--foo=bar` form.
* As an additional backstop, if an argument is provided (e.g. `--foo=bar`) but it either a) isn't referred to by any `getopt` call, or b) has a value with the wrong type (can't be parsed, etc.), that's an error. This makes it hard to mess up.

src/util/str_to_bool.cpp
amscanne: Whatever your prefer. `str_to_bool` seems *overly* specific IMO, but I don't object. Maybe `format` just needs renaming to `stings` or something with a bunch of string functions.

man/adoc/bpftrace.adoc
```diff
@@ -188,6 +188,16 @@ Print bpftrace version information.
 Enable verbose messages.
 For more details see the <<Verbose Output>> section.
 
+=== Program Options
+
+You can also pass custom options to a bpftrace program/script itself via positional or named parameters.
+All positional parameters need to come before any named parameters e.g.
```
amscanne: I noticed this while reading the parsing code, and have a long comment/proposal below. We can discuss there, I just wanted to note that this isn't really how positional parameters normally work, and therefore we would be introducing a rather sharp edge right away.

src/main.cpp
jordalgo: Good call - I didn't realize there was this convention of having named arguments come first. From what I was reading everything after a double dash is considered a positional param anyway even if it is prefixed with "--" (as you alluded to).

> require the arguments to use =

Sounds good. However, I think it works if we allow boolean flags e.g. this should be fine `--foo --bar`. Both would be set to true.

>  either a) isn't referred to by any getopt call, or b) has a value with the wrong type (can't be parsed, etc.), that's an error

Ack. Will fix.

src/util/str_to_bool.cpp
jordalgo: kk, made a single `strings` util that merges this and `format`

src/ast/passes/cli_opts.cpp
jordalgo: What is the concern about the current implementation that you think won't be compatible with AOT? They are going to be different code paths whereby we do compile time replacement for non-AOT and global setting/lookup for AOT. Waiting for AOT, which users can't even use right now, seems counter to our current "move fast" goal - especially as this is a feature we've wanted to ship for a long time.

src/main.cpp
jordalgo: Why couldn't we make those available (in the future) to `getopt` e.g. `getopt("pid")` ? Also what other current long options do you think users need to access in their programs?

amscanne:
> @amscanne I'm going to work on the map decl stuff for `getopt` in a separate PR as that's a bit more complicated. We have to create the map and then pre-populate it with the argument value (or default) after program load but before attachment (which isn't hard but we just don't have a mechanism for that yet -- I don't think).

This is basically the AOT design ;) I would roll it into that if you're going to do it separately, don't have two separate code paths for how this works internally.

jordalgo:
Working on an alternative solution using global vars so it works for AOT as well.
- https://github.com/bpftrace/bpftrace/pull/4267

Issue #4017, opened 2025-04-16
proposal: fold `src/arch` into main code


amscanne:
This is a relatively low priority proposal, but I just wanted to get it down as an idea. CC @danobi

Having conditional compilation units without any build smoke tests for other architectures can make it difficult to catch dumb mistakes (e.g. #4016). It would be difficult and expensive to set up and maintain smoke tests for all the different systems that we would like to support.

I believe that this layout was inspired by the kernel layout (top-level `arch`), but I don't think it's strictly necessary. One of the main purposes of the functions in `arch` is to determine the register offsets (in `pt_regs`), but in some cases this is actually wrong. We should add an early pass that translates `regs("reg_name")` to `ctx.reg_name` and allow the standard field analyzer to compute the offset. In the future, when we have appropriate BTF relocation information to have this access be portable, this will also benefit from that. Similarly, this pass could translate known builtins (e.g. `retval`), and even provide a builtin (e.g. `arch`) that reduces to a string for the current architecture.

For other arch-specific functions, (e.g. `invalid_watchpoint_modes`), these can be moved to simple inline `#ifdef`) blocks within an actual translation unit. This would have the effect of minimizing the amount of "uncompiled" state, but also would serve to reduce duplication (e.g. many architectures do not support any valid watchpoint modes, and therefore could share a common `#else` clause).

jordalgo:
Proposal sounds good to me and I like @danobi 's suggestion of not using ifdefs

PR #4260, opened 2025-06-13
arch: ensure all code is built all the time

amscanne:
Each architecture is a specialization of the generic `Arch` class. The only `#ifdef` is used for a `constexpr` function which returns the current architecture, which can be used to instantiate (in a non-virtual way) the intended archiecture helpers.

In general, I expect that all unused implementations will be stripped out by the linker as they are unreferenced. However, we can easily create a parameterized test that *does* use all these implementations to provide a uniform sanity test.

Updates #4017

##### Checklist

### DIFF
--- a/src/arch/x86_64.cpp
+++ b/src/arch/x86_64.cpp
@@ -1,92 +1,86 @@
-#include <algorithm>
-#include <array>
+#include <unordered_map>
 
 #include "arch.h"
 
-// SP + 8 points to the first argument that is passed on the stack
-enum { ARG0_STACK = 8 };
-
 namespace bpftrace::arch {
 
-// clang-format off
-static std::array<std::string, 27> registers = {
-  "r15",
-  "r14",
-  "r13",
-  "r12",
-  "bp",
-  "bx",
-  "r11",
-  "r10",
-  "r9",
-  "r8",
-  "ax",
-  "cx",
-  "dx",
-  "si",
-  "di",
-  "orig_ax",
-  "ip",
-  "cs",
-  "flags",
-  "sp",
-  "ss",
-};
-
-static std::array<std::string, 6> arg_registers = {
-  "di",
-  "si",
-  "dx",
-  "cx",
-  "r8",
-  "r9",
-};
-// clang-format on
-
-int offset(std::string reg_name)
+template <>
+size_t Arch<Machine::X86_64>::kernel_ptr_width()
 {
-  auto *it = std::ranges::find(registers, reg_name);
-  if (it == registers.end())
-    return -1;
-  return distance(registers.begin(), it);
+  return 64;
 }
 
-int max_arg()
+template <>
+std::optional<size_t> Arch<Machine::X86_64>::register_to_pt_regs_offset(
+    const std::string& name)
 {
-  return arg_registers.size() - 1;
+  static const std::unordered_map<std::string, size_t> register_offsets = {
+    { "r15", 0 },  { "r14", 8 },  { "r13", 16 },    { "r12", 24 },
+    { "bp", 32 },  { "bx", 40 },  { "r11", 48 },    { "r10", 56 },
+    { "r9", 64 },  { "r8", 72 },  { "ax", 80 },     { "cx", 88 },
+    { "dx", 96 },  { "si", 104 }, { "di", 112 },    { "orig_rax", 120 },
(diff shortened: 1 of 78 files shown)


jordalgo (approved):
Bit of an "I trust you" stamp as it looks like most of this is just moving things around. The added unit tests are nice though.

amscanne:
> Bit of an "I trust you" stamp as it looks like most of this is just moving things around. The added unit tests are nice though.

There shouldn't be any functional changes, however I have added the ground work for evaluating registers via BTF and semantic analysis, but will leave that to a later change and pass.

I did take a look to see if anything got accidentally changed, if you twenty minutes it's always good to get another set of eyes on that.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -3551,8 +3552,9 @@ void SemanticAnalyser::visit(AttachPoint &ap)
       if (!bpftrace_.pid().has_value() && !has_child_)
         ap.addError() << "-p PID or -c CMD required for watchpoint";
 
-      if (ap.address > static_cast<uint64_t>(arch::max_arg()))
-        ap.addError() << arch::name() << " doesn't support arg" << ap.address;
+      if (ap.address >= static_cast<uint64_t>(arch::Host::arguments().size()))
+        ap.addError() << arch::Host::Machine << " doesn't support arg"
+                      << ap.address;
```
viktormalik, line 3557: Nit: add curly braces for multi-line if body?

src/arch/x86_64.cpp
```diff
@@ -1,92 +1,86 @@
...
 }
 
-int pc_offset()
+template <>
+size_t Arch<Machine::X86_64>::argument_stack_offset()
 {
-  return offset("ip");
+  // The return address is pushed on the frame, so we need to reach up
+  // the stack further to find the first argument.
+  return 8;
```
viktormalik, line 61: IIUC, the original semantics of `arg_stack_offset` is that it returns the number of stack entries, not the amount of bytes. Same for other arches with non-zero offset.

amscanne:
@jordalgo and I did a pass over register offsets, arguments, etc. and it seemed sane. Sending it.

viktormalik (commented):
Damn, I had a review pending but wanted to fully finish it before sending it.

Anyways, `sargX` is IMO broken by this (if we want to still support it, I remember we discussed getting rid of it).

jordalgo:
@viktormalik No worries. We can revert or fix forward as needed. I'll let @amscanne respond

src/ast/passes/semantic_analyser.cpp
amscanne: I **try** not to make unrelated changes, so this is just the existing structure.

But I would **love** to have all `if` statements in braces :) I think we could probably update the clang-fmt settings and format everything, is this something you'd be in favor of?

viktormalik: I kind of like the kernel guidelines for this: no braces if all branches within a single `if`/`else` expression have a single line, otherwise braces for all the branches. I'm not sure if clang-format can be configured in such a way.

Also here you changed a single-line block to multi-line, that's why introducing braces is IMO a related change.

src/arch/x86_64.cpp
amscanne: Yeah, there was a mistake in the codegen calculation (good catch, sent a PR to fix) but the move to byte-based offsets for this and the `pt_regs` offsets was intentional. I don't think anything guarantees the specific layout here, so byte offsets feels more justified to me. (But I'm aiming to completely remove the offsets anyways, in favor of replacing with the expression which will be evaluated via BTF.)

I'm happy to revisit this if you feel strongly, I just did it to allow the unified `Arch` API to return the same units (which make more sense in my head).

viktormalik: Sure, byte offsets make more sense to me, too (IMO that's why they are called "offsets" and not "indices"). I just wanted to make sure that we don't break the existing behavior. Thanks for sending the fix PR!

src/ast/passes/semantic_analyser.cpp
jordalgo: My 2 cents - I also like having braces for ALL `if` statements regardless of number of following lines.

viktormalik: Even though I like compact, single-line, `if` statements, I'm obviously not going to push against having braces everywhere :-)

PR #4263, opened 2025-06-13
build: Fix linking for Fedora build

amscanne:
It seems Fedora packages the clang libraries differently. When building the Docker image, set the `cmake` flag appropriately.

##### Checklist

### DIFF
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -77,15 +77,7 @@ if(STATIC_LINKING)
   unlink_transitive_dependency("${CLANG_EXPORTED_TARGETS}" "LLVM")
   unlink_transitive_dependency("${CLANG_EXPORTED_TARGETS}" "$<LINK_ONLY:clang-cpp>")
 
-  if(TARGET libclang_static)
-    target_link_libraries(ast PUBLIC libclang_static)
-  else()
-    # old LLVM versions don't export libclang_static in ClangTargets.cmake; fall back to
-    # libclang.a in that case
-    target_link_libraries(ast PUBLIC libclang.a)
-  endif()
-
-  target_link_libraries(ast PUBLIC clangDriver clangFrontend clangCodeGen)
+  target_link_libraries(ast PUBLIC libclang_static clangDriver clangFrontend clangCodeGen)
   target_link_libraries(ast PUBLIC ${llvm_libs})
 else()
   if (TARGET LLVM)
@@ -98,5 +90,6 @@ else()
     # have libLLVM.so, drop `USE_SHARED` to avoid forcing link to `-lLLVM`.
     llvm_config(ast bpfcodegen ipo irreader mcjit orcjit)
   endif()
-  target_link_libraries(ast PUBLIC libclang clangDriver clangFrontend clangCodeGen)
+
+  target_link_libraries(ast PUBLIC libclang clang-cpp)
 endif()
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -23,6 +23,7 @@ RUN apt-get update && apt-get install -y \
     libpcap-dev \
     llvm-dev \
     libclang-dev \
+    libclang-cpp-dev \
     pahole \
     systemtap-sdt-dev \
     xxd \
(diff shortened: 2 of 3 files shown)


src/ast/CMakeLists.txt
```diff
@@ -48,11 +48,17 @@ target_link_libraries(ast PRIVATE debugfs tracefs util stdlib)
...
 
 if (NOT SYSTEM_INCLUDE_PATHS EQUAL "auto")
-  MESSAGE(STATUS "Using SYSTEM_INCLUDE_PATHS=${SYSTEM_INCLUDE_PATHS}")
+  message(STATUS "Using SYSTEM_INCLUDE_PATHS=${SYSTEM_INCLUDE_PATHS}")
 endif()
 target_compile_definitions(ast PUBLIC SYSTEM_INCLUDE_PATHS="${SYSTEM_INCLUDE_PATHS}")
 
 if(STATIC_LINKING)
+  if(USE_CLANG_CPP)
+    # The `USE_CLANG_CPP` exists to support linking for Fedora, which is
```
jordalgo: wait, why does Fedora not require static linking?

jordalgo (approved):

src/ast/CMakeLists.txt
```diff
@@ -98,5 +104,9 @@ else()
     # have libLLVM.so, drop `USE_SHARED` to avoid forcing link to `-lLLVM`.
     llvm_config(ast bpfcodegen ipo irreader mcjit orcjit)
   endif()
-  target_link_libraries(ast PUBLIC libclang clangDriver clangFrontend clangCodeGen)
+  if(USE_CLANG_CPP)
+    target_link_libraries(ast PUBLIC libclang libclang-cpp)
+  else()
+    target_link_libraries(ast PUBLIC libclang clangDriver clangFrontend clangCodeGen)
+  endif()
```
viktormalik: I think we're not doing the right thing here. IIUC, `clangDriver`, `clangFrontend`, and `clangCodeGen` are typically built as static libraries so we should not be using them here when linking dynamically. Instead, I think that we should always use `libclang-cpp.so` which other distros package, too.

amscanne: Hm, yes I think you're right. I was assuming too much. Let me click down and see what's going on there. I see that the static libraries are included in the regular `-dev` package on debian (https://packages.debian.org/sid/amd64/libclang-19-dev/filelist), so perhaps it's just that Fedora has a stricter separation there (and we generally don't want to pick up the static libraries, so it's helpful in that way).

viktormalik: Yeah, Fedora typically puts static libraries into a `-static` package, however, I don't see `clang-static` anywhere. It is possible that they are not packaged at all, I'll dig more.

amscanne: I'm not worried about the static case for Fedora. I'm not sure why the `CompilerInstance` class is in `clang-cpp` but it does seem to be the case. I believe that this is now correct, but I'll test with each distro individually prior to submission, assuming this looks reasonable.

src/ast/CMakeLists.txt
```diff
@@ -98,5 +90,6 @@ else()
     # have libLLVM.so, drop `USE_SHARED` to avoid forcing link to `-lLLVM`.
     llvm_config(ast bpfcodegen ipo irreader mcjit orcjit)
   endif()
-  target_link_libraries(ast PUBLIC libclang clangDriver clangFrontend clangCodeGen)
+
+  target_link_libraries(ast PUBLIC libclang clang-cpp)
```
viktormalik, line 94: Nit: I think that we could do with
```suggestion
  target_link_libraries(ast PUBLIC clang clang-cpp)
```

viktormalik (approved):
Yeah, this looks good now, thanks!

src/ast/CMakeLists.txt
amscanne: Resolved separately.

src/ast/CMakeLists.txt
amscanne: This doesn't work as expected, I believe because this is the cmake target name not `find_library` name.

viktormalik: Ah, ok, then sorry for the confusion.

PR #4264, opened 2025-06-13
cmake: link against static `libbpf`

amscanne:
Stacked PRs:
 * #4276
 * #4275
 * #4274
 * #4273
 * #4272
 * #4271
 * #4265
 * __->__#4264

--- --- ---

### cmake: link against static `libbpf`

See #4151 for more detail.

### DIFF
--- a/cmake/FindLibBpf.cmake
+++ b/cmake/FindLibBpf.cmake
@@ -1,16 +1,9 @@
-# - Try to find libbpf
-# Once done this will define
-#
 #  LIBBPF_FOUND - system has libbpf
 #  LIBBPF_INCLUDE_DIRS - the libbpf include directory
 #  LIBBPF_LIBRARIES - Link these to use libbpf
 #  LIBBPF_DEFINITIONS - Compiler switches required for using libbpf
 #  LIBBPF_VERSION_MAJOR - Libbpf major version
 #  LIBBPF_VERSION_MINOR - Libbpf minor version
-
-#if (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIRS)
-#  set (LibBpf_FIND_QUIETLY TRUE)
-#endif (LIBBPF_LIBRARIES AND LIBBPF_INCLUDE_DIRS)
 
 find_path (LIBBPF_INCLUDE_DIRS
   NAMES
@@ -20,9 +13,11 @@ find_path (LIBBPF_INCLUDE_DIRS
   PATHS
     ENV CPATH)
 
+# We require that `libbpf` be linked statically, as we will implicitly vendor
+# the header into embedded code. See #4151 for more information.
 find_library (LIBBPF_LIBRARIES
   NAMES
-    bpf
+    libbpf.a
   PATHS
     ENV LIBRARY_PATH
     ENV LD_LIBRARY_PATH)
@@ -30,7 +25,7 @@ set(LIBBPF_ERROR_MESSAGE "Please install the libbpf development package")
 
 include (FindPackageHandleStandardArgs)
 
-# handle the QUIETLY and REQUIRED arguments and set LIBBPF_FOUND to TRUE if all listed variables are TRUE
+# handle the QUIETLY and REQUIRED arguments and set LIBBPF_FOUND to TRUE if all listed variables are TRUE.
 FIND_PACKAGE_HANDLE_STANDARD_ARGS(LibBpf ${LIBBPF_ERROR_MESSAGE}
   LIBBPF_LIBRARIES
   LIBBPF_INCLUDE_DIRS)
--- a/docker/Dockerfile.fedora
+++ b/docker/Dockerfile.fedora
@@ -22,6 +22,7 @@ RUN dnf install -y \
         gcc-c++ \
         libpcap-devel \
         libbpf-devel \
+        libbpf-static \
         llvm-devel \
         make \
         pahole \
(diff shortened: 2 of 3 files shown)


jordalgo (approved):
Should be ok. cc @viktormalik and @acj - for distro packaging as some distros frown upon static linking.

Phoenix500526 (commented):
That looks greate.

viktormalik:
> Should be ok. cc @viktormalik and @acj - for distro packaging as some distros frown upon static linking.

Yeah, I'm not particularly happy about this. Fedora (and I believe other distros, too) strongly discourage bundling libraries for good reasons.

But if I understand the problem correctly, there's not much we can do here. Unless libbpf is able to make USDT argument packing/unpacking a part of their ABI, I don't see an option other than statically linking against libbpf.

So I guess before committing to this change, my question is - is it completely out of the table to persuade libbpf maintainers to keep backwards compatibility when it comes to USDT arg management? Shouldn't it be considered a part of the libbpf's ABI? Or, if not, are we perhaps able to do something else on libbpf side (like separate USDT management to a separate library)? I know I may be over-complicating things but I just want to make sure that we've considered all the options before committing to this practice going against distro packaging principles.

Also, I'm not saying that we should block this change until we explore all the options. I'm all good with merging this now (we're still far away from the next release) but it would be good if we could at least try to find some better option long-term.

jordalgo:
I pinged @anakryiko offline to get his thoughts as well

anakryiko:
I don't see any problems with linking libbpf statically. If anything, that's what I always recommend. Distros' insistence on dynamically linked libraries, particularly for libbpf use cases, has never been convincing (security argument is usually brought up, but the need for it hasn't been demonstrated so far, and if anything, libbpf shouldn't be used with untrusted inputs anyways, it's root-level library; and bpftrace as a project can easily take care of updating its linked libbpf with any necessary fixes as fast if not faster, as any distro, IMO).

On the upside, bpftrace can get any necessary unreleased feature or bug fix as soon as it gets synced to Github. And given tight relationships between bpftrace and libbpf projects, it might be possible to rely on some non-stable (UAPI-wise) functionality, if absolutely necessary (within reasons, of course).

In short, I think for project like bpftrace statically linking libbpf makes all the sense, regardless of USDTs.

For USDT specifically. I'm forgetting the specifics of what internal implementation details bpftrace will be relying on. But before committing to adding some more UAPI stability, it would be nice to have bpftrace do full transition and identify what exactly they need w.r.t. USDT support. So either way I think this step is a good one, and we should come back to enhancing USDT APIs in libbpf once the switch to libbpf is done and some real-world experience is gained.

viktormalik:
> I don't see any problems with linking libbpf statically. If anything, that's what I always recommend. Distros' insistence on dynamically linked libraries, particularly for libbpf use cases, has never been convincing (security argument is usually brought up, but the need for it hasn't been demonstrated so far, and if anything, libbpf shouldn't be used with untrusted inputs anyways, it's root-level library; and bpftrace as a project can easily take care of updating its linked libbpf with any necessary fixes as fast if not faster, as any distro, IMO).

This has been discussed many times, I don't think we're about to find an agreement here.

> On the upside, bpftrace can get any necessary unreleased feature or bug fix as soon as it gets synced to Github. And given tight relationships between bpftrace and libbpf projects, it might be possible to rely on some non-stable (UAPI-wise) functionality, if absolutely necessary (within reasons, of course).
> 
> In short, I think for project like bpftrace statically linking libbpf makes all the sense, regardless of USDTs.

Until we really need this, I'd stay out of using libbpf via a submodule. It just adds additional maintenance need for no added value (at least for the moment).
 
> For USDT specifically. I'm forgetting the specifics of what internal implementation details bpftrace will be relying on. But before committing to adding some more UAPI stability, it would be nice to have bpftrace do full transition and identify what exactly they need w.r.t. USDT support. So either way I think this step is a good one, and we should come back to enhancing USDT APIs in libbpf once the switch to libbpf is done and some real-world experience is gained.

I like this approach. Let's switch to static linking for now and see if we're able to stabilize the libbpf UAPI in future, based on our needs.

amscanne:
Aside: it appears that `libbcc_bpf` is dynamically linked against `libbpf`, so `bpftrace` will continue to have an effective runtime dependency on `libbpf.so` until that migration is complete anyways.

amscanne:
I will submit shortly. Doesn't mean that it will be forever, nor a general statement on static linking. :)

(We can also change/rollback if it remains contentious. Just want to enable libbpf USDT migration.)

jordalgo:
cc @vincentbernat (sorry, Vincent, pinged you on the wrong thread previously)

vincentbernat:
In Debian, this is also frowned upon, but it is allowed for good technical reasons. libbpf-dev package provides libbpf.a, so it should be OK. However, we may end up in a situation where we are dynamically linked with libbpf version A and statically linked with libbpf version B. Would it be a problem?

amscanne:
> In Debian, this is also frowned upon, but it is allowed for good technical reasons. libbpf-dev package provides libbpf.a, so it should be OK. However, we may end up in a situation where we are dynamically linked with libbpf version A and statically linked with libbpf version B. Would it be a problem?

I don't believe that this would be a problem, for two reasons:
* I'm not aware of any shared global state or reasons that transitive uses of `libbpf` would pose compatibility issues, we just need the static binding to ensure that the implementation of the attachpoints that **we** use is the same as the headers that ultimately get vendored and embedded (that define the associated maps use by `libbpf`). This is a good thing to keep an eye out for though, although hopefully the second item solves this...
* I hope that this situation is very short-lived, as the reason for this is ultimately motivated by a migration to `libbpf` for the attachment APIs, and we would no longer have the dynamic dependency.

In any case, I'll move forward for now and we can revisit if there are problems. The input is much appreciated!

PR #4265, opened 2025-06-13
cmake: Build the `stdlib` dynamically

amscanne:
cmake: Build the `stdlib` dynamically

This does not change the external structure, but generates the standard
library dynamically based on all available files. It also includes
standardized headers from `libbpf`.

This change also shifts to using `string_view` to allow us to reference
the objects embedded in binary data without needing new allocations.


src/stdlib/CMakeLists.txt
```diff
@@ -1,42 +1,84 @@
...
+    OUTPUT  ${PREFIX}/${header_name}.h
+    VAR     ${rule_name}
   )
+
+  # Add the file to the list of rules we need, and add the original
+  # source file to the list of sources we need to depend on.
+  list(APPEND STDLIB_TARGETS "${rule_name}")
+  set(STDLIB_TARGETS "${STDLIB_TARGETS}" PARENT_SCOPE)
+
+  # Write out our meta-includes.
```
jordalgo: meta includes?

jordalgo (approved):

src/stdlib/CMakeLists.txt
```diff
@@ -1,42 +1,84 @@
...
-  VAR     base_bt
-)
-
-foreach(header in float limits stdarg stdbool stddef __stddef_max_align_t stdint)
+# The standard library is effectively a bunch of embedded files.
+#
+# The data layout is standardized by `stdlib.h`, but we generate
+# stdlib.c which includes the relevant generated headers.
+#
+# * Basically, everything here that is `*.c` or `*.h` goes into
```
viktormalik: `*.bt` files are apparently taken, too, reflect that in the comment?

src/stdlib/CMakeLists.txt
```diff
@@ -1,42 +1,84 @@
...
+#include \"${PREFIX}/${header_name}.h\"
+")
+
+  # Write out our translation unit.
+  file(APPEND ${CMAKE_CURRENT_BINARY_DIR}/stdlib.cpp "
+  { \"${PREFIX}/${name}\", make_view(${rule_name}, sizeof(${rule_name})) },
+")
+endfunction()
+
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/stdlib.cpp "
```
viktormalik: I'm a bit on the fence here but how about we generated this via `configure_file`? That would allow us to have the C++ code in a separate file (stdlib.cpp.in) since embedding it in CMakeLists.txt is quite hard to read. OTOH it may be more difficult to grasp the entire context when it's spread among multiple files. But I'm slightly in favor of using `configure_file`.

amscanne: I did look for something that would generate from a template rather than having it bespoke, but didn't find any good options initially because `configure_file` didn't have a good way to expand lists. I'll take another look. I agree that it's a bit janky as is.

amscanne: Ugh, the problem here is that I basically want a trivial templating engine (e.g. similar to the request in https://discourse.cmake.org/t/configure-file-generate-code-dependant-on-an-if-clause/3269), but I'd rather avoid introducing something else into the build system. Are you okay if I just separate this into a distinct cmake module & function? Then at least it's encapsulated.

src/stdlib/CMakeLists.txt
amscanne: It includes all the includes.

src/stdlib/CMakeLists.txt
viktormalik: I'm probably missing something here but why we cannot have `stdlib.cpp.in` like this:
```cpp
#include "stdlib/stdlib.h"
#include "stdlib/stdlib_includes.h"
namespace bpftrace::stdlib {
inline std::string_view make_view(const unsigned char *v, size_t sz)
{
  return { reinterpret_cast<const char *>(v), sz };
}
const std::map<std::string, std::string_view> Stdlib::files = {
    ${STDLIB_FILES}
};
} // namespace bpftrace::stdlib
```
and then populate `STDLIB_FILES` from CMakeLists.txt (using `add_stdlib_source`) and call `configure_file(stdlib.cpp.in stdlib.cpp)`?

amscanne: Ah, I just didn't want to have that because it's still just constructing a magic string (and the cmake bits in `CMakeLists.txt` implicitly depends on the structure in `stdlib.cpp.in`). But I agree that what you've proposed here is an improvement, so I'll just do that if there's nothing obviously better :)

jordalgo:
cc @vincentbernat

src/stdlib/CMakeLists.txt
amscanne: Done.

src/stdlib/CMakeLists.txt
amscanne: I've implemented your suggestion!

src/stdlib/CMakeLists.txt
amscanne: This is no longer used.

PR #4266, opened 2025-06-14
Refactor Global Vars

jordalgo:
This is in preparation to create dynamic global
vars based on the bpftrace script itself e.g.
from `getopt` and read-only map optimization.

This makes GlobalVars a class whereby parts
of bpftrace can add global vars instead of there
being a static list - though we do maintain a list
of "known" global variables.

##### Checklist


src/globalvars.h
```diff
@@ -17,27 +18,19 @@ class RequiredResources;
...
-  WRITE_MAP_VALUE_BUFFER,
-  VARIABLE_BUFFER,
-  MAP_KEY_BUFFER,
-  EVENT_LOSS_COUNTER,
-};
-
-std::string to_string(GlobalVar global_var);
-std::optional<GlobalVar> from_string(std::string_view name);
+// Known global variables
+constexpr std::string_view NUM_CPUS = "__bt__num_cpus";
```
jordalgo, line 22: Prefixing these with "\_\_bt\_\_" will make more sense in the following PR but it's essentially to avoid collisions with user-provided globals (e.g. via `getopt` or map names)

src/globalvars.h
```diff
@@ -52,71 +45,99 @@ constexpr std::string_view MAP_KEY_BUFFER_SECTION_NAME = ".data.map_key_buf";
...
 
-void update_global_vars(
-    const struct bpf_object *bpf_object,
-    const std::unordered_map<std::string, struct bpf_map *> &global_vars_map,
-    const BPFtrace &bpftrace);
+const std::unordered_map<std::string_view, GlobalVarConfig>
+    GLOBAL_VAR_CONFIGS = {
+      { NUM_CPUS,
+        { .section = std::string(RO_SECTION_NAME),
+          .type = GlobalVarType::integer } },
```
amscanne, line 67: Is there a good reason to have custom sections for everything? Or can we just have a `read_only` flag and have the generation record this as either `data` or `rodata`?

jordalgo: Actually yes, because we resize (via `bpf_map__set_value_size`) these globals/array-maps they need their own elf sections. From the `bpf_map__set_value_size` docs: "This attempt is best-effort and can only succeed if the last variable of the data section map is an array." This resizing is done for AOT.

PR #4267, opened 2025-06-15
Getopt

jordalgo:
##### Checklist


jordalgo:
I split this into 3 commits so hopefully it's a bit easier to review now.

man/adoc/bpftrace.adoc
```diff
@@ -188,6 +188,17 @@ Print bpftrace version information.
 Enable verbose messages.
 For more details see the <<Verbose Output>> section.
 
+=== Program Options
+
+You can also pass custom options to a bpftrace program/script itself via positional or named parameters.
+All positional and named parameters should come after a double dash e.g.
```
amscanne: ... should come after the script filename or a double dash if running an inline script with `-e`

man/adoc/bpftrace.adoc
```diff
@@ -188,6 +188,17 @@ Print bpftrace version information.
...
+```
+# bpftrace -e 'BEGIN { ... }' -- p1 --aa=20 --bb p2
+```
+In this example there are two positional parameters (p1, p2) and two named parameters (aa - set to 20 and bb - set to true).
```
amscanne: "aa - set to 20 and bb - set to true" has very confusing punctuation. I would just use a full sentence, e.g. "`aa`, which is set to 20, and `bb`, which is set to true".

(Consider using backticks for the variables to make them stand out there also)

man/adoc/bpftrace.adoc
```diff
@@ -2451,6 +2466,33 @@ BEGIN {
 }
 ----
 
+[#functions-getopt]
+=== getopt
+
+.variants
+* `int getopt(string arg_name)`
```
amscanne: Consider just requiring a default value?

src/aot/aot_main.cpp
```diff
@@ -6,9 +6,11 @@
```
amscanne: Is there a reason we can't push the changes required here into `run_bpftrace` or some core runtime layer? May require some gentle refactoring.

src/ast/ast.h
```diff
@@ -271,6 +273,41 @@ class NegativeInteger : public Node {
   const int64_t value;
 };
 
+class NamedParameter : public Node {
```
amscanne: I wouldn't add a new AST structure here. You are desugaring to something new that doesn't otherwise exist, which is strange.

I would desugar directly to the map or whatever the underlying implementation is, and then have the namedparameter pass emit a pass output without whatever named parameter metadata you need to consume by e.g. the resource pass.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -531,6 +532,28 @@ ScopedExpr CodegenLLVM::visit(NegativeInteger &integer)
   return ScopedExpr(b_.getInt64(integer.value));
 }
 
+ScopedExpr CodegenLLVM::visit(NamedParameter &named_param)
```
amscanne: I'm not a huge fan of needing new custom codegen for this case. Why can't this be handled as standard map after desugaring? (And in codegen maybe we add some map metadata that says whether soemthing is read-only, and we reject writes to "read-only" maps in the semantic analyzer.)

src/ast/passes/named_param.cpp
```diff
@@ -0,0 +1,128 @@
...
+                                                Location(call->loc));
+  } else if (auto *default_value = call->vargs.at(1).as<NegativeInteger>()) {
+    // integer
+    expected_default_val = std::to_string(default_value->value);
+    expr.value = ast_.make_node<NamedParameter>(arg_name->value,
+                                                CreateInt64(),
+                                                expected_default_val,
+                                                false,
+                                                Location(call->loc));
+  } else if (auto *default_value = call->vargs.at(1).as<Identifier>()) {
```
amscanne: I'm not a huge fan of the magic identifier cases (there are only a few, and this is adding a new one). Maybe you could file an issue to make true/false actual values that are passed as a boolean type, and we could at least leave a FIXME here? I wouldn't advertise support for `yes`/`no` as part of the API.

src/globalvars.cpp
```diff
@@ -258,19 +303,99 @@ void GlobalVars::verify_maps_found(
...
+        return make_error<NamedParamError>(
+            "Command line option --" + name +
+            " is expecting an integer. Got: " + val);
+      } catch (const std::out_of_range &) {
+        return make_error<NamedParamError>("Value for command line option --" +
+                                           name +
+                                           " is out of range. Got: " + val);
+      }
+    } else if (config.type == GlobalVarType::boolean) {
+      if (val == "--") {
```
amscanne: Huh? This is `--foo=--` ? Why is this not an error?

src/globalvars.cpp
```diff
@@ -258,19 +303,99 @@ void GlobalVars::verify_maps_found(
...
+    auto &config = added_global_vars_.at(name);
+    auto it = named_params.find(name);
+    std::string val = default_val;
+
+    if (it != named_params.end()) {
+      val = it->second;
+      named_params.erase(name);
+    }
+
+    if (config.type == GlobalVarType::integer) {
```
amscanne: Why is the `config.type` distinct from just the type held by the variant?

src/globalvars.cpp
amscanne: Oh, I see that this is "TRUTH_VALUE". I think it is actually a presence value.

Why do you need this? Can't presence in the map be an indicator of presence (e.g. just make `--foo` and `--foo=` equivalent in terms of representation).

src/globalvars.cpp
```diff
@@ -258,19 +303,99 @@ void GlobalVars::verify_maps_found(
...
-  for (const auto &[name, config] : global_var_map_) {
+  for (const auto &[name, config] : added_global_vars_) {
     if (!section_name_to_global_vars_map.contains(config.section)) {
       LOG(BUG) << "No map found in " << config.section
                << " which is needed to set global variable " << name;
     }
   }
 }
 
+Result<GlobalVarVals> GlobalVars::get_named_param_vals(
```
amscanne: I think you could probably isolate all the parsing state and logic and add unit tests that makes it easier to grok?

src/aot/aot_main.cpp
jordalgo: The challenge there was the invoking of the `usage` function in both `main.cpp` and `aot_main.cpp`. I could pass a lambda down through a few layers that is invoked upon these specific `Result` errors or just not show the `usage` menu upon unexpected long args. WDYT?

Edit: Ok, I actually just passed a lambda down one layer and was able to consolidate some logic.

src/ast/passes/codegen_llvm.cpp
jordalgo: Yeah, that's not a bad idea. Let me look into that.

src/ast/passes/named_param.cpp
jordalgo: Yeah, I didn't love it either. Maybe I'll just add a boolean type (as I've wanted to do that for a while now) before landing this.

src/globalvars.cpp
jordalgo: > just make --foo and --foo= equivalent in terms of representation

Meaning the value in the map that I pass around is just an empty string? That should work - will update.

src/globalvars.cpp
jordalgo: That's why I made this function separate and added unit tests for it :)

src/globalvars.cpp
jordalgo: Not sure what you mean. The `named_params` come in as just a map of string to string. I use the config type to determine the type of the variant held in `GlobalVarVals` and do things like `stoll`.

src/ast/passes/codegen_llvm.cpp
jordalgo: Done.

src/ast/ast.h
jordalgo: Done.

man/adoc/bpftrace.adoc
jordalgo: I like that we provide a terse option but I won't die on this hill. Though maybe I should change this behavior to not assume a boolean flag but rather we can just try to deduce the type from semantic analysis and if it's ambiguous we throw an error. This leaves the door open for a named argument `required` e.g. `getopt("hello", required=true)` that doesn't force a default value when the option is required.

Edit: actually trying to deduce the type and back propagate in semantic analyser seems like a pain.

src/aot/aot_main.cpp
```diff
@@ -132,8 +137,25 @@ int main(int argc, char* argv[])
...
+        auto unrecognized_opt = std::string(argv[optind - 1]);
+        if (unrecognized_opt.starts_with("--")) {
+          auto arg = unrecognized_opt.substr(2);
+          if (arg.find("=") != std::string::npos) {
+            auto split = util::split_string(arg,
+                                            '=',
+                                            /* remove_empty= */ true);
+            named_params[split[0]] = split[1];
+          } else {
+            named_params[arg] = "";
```
amscanne: Nit: you can distinguish between the string and no string by having the map be a `std::map<std::string, std::optional<std::string>>`. Then you would still have strict checking for `--foo` and `--foo=`.

src/ast/ast.h
```diff
@@ -513,10 +513,18 @@ using MapDeclList = std::vector<MapDeclStatement *>;
 class Map : public Node {
 public:
   explicit Map(ASTContext &ctx, std::string ident, Location &&loc)
-      : Node(ctx, std::move(loc)), ident(std::move(ident)) {};
+      : Node(ctx, std::move(loc)), ident(std::move(ident)), read_only(false) {};
+  explicit Map(ASTContext &ctx,
+               std::string ident,
+               bool read_only,
```
amscanne, line 519: I don't think the ast is the ideal place for this, but won't make a fuss.

src/ast/passes/named_param.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "ast/pass_manager.h"
+#include "types.h"
+
+namespace bpftrace::ast {
+
+struct NamedParamDefault {
+  std::string value;
+  bool is_bool = false;
```
amscanne: Please, no. 😮‍💨 

Just do this properly, and have value be a `std::variant`. E.g.

```
std::variant<std::string, int64_t, uint64_t, bool> value;
```

Then you can also ensure that the type matches. The `type` can be carried implicitly by the variant that you've set, rather than requiring yet-another link to `SizedType` and including `types.h`.

src/ast/passes/named_param.cpp
```diff
@@ -0,0 +1,124 @@
...
+  if (call->vargs.size() == 1) {
+    // boolean
+    map_node->value_type = CreateBool();
+    expected_default_val = "false";
+    np_default.is_bool = true;
+  } else if (auto *default_value = call->vargs.at(1).as<Boolean>()) {
+    // boolean
+    map_node->value_type = CreateBool();
+    expected_default_val = default_value->value ? "true" : "false";
+    np_default.is_bool = true;
```
amscanne: See my comment elsewhere, but this is really hacky. You don't need to create a new sized type, just store this as a variant and unpack it appropriately later.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2841,7 +2841,7 @@ void SemanticAnalyser::visit(MapAccess &acc)
   } else {
     // If there is no record of any assignment after the first pass
     // then it's safe to say this map is undefined.
-    if (!is_first_pass()) {
+    if (!is_first_pass() && !acc.map->read_only) {
```
amscanne, line 2844: This feels a bit strange, and suggests that you want to actually insert this in the known set of maps for semantic analysis or do something else.

src/globalvars.cpp
```diff
@@ -232,23 +265,40 @@ std::unordered_set<std::string> get_section_names()
...
+
+void GlobalVars::add_named_param(const std::string &name,
+                                 Type type,
+                                 bool is_bool,
+                                 const std::string &default_val)
+{
+  if (added_global_vars_.contains(name)) {
+    return;
+  }
+  added_global_vars_[name] = GlobalVarConfig({
```
amscanne, line 303: Per comment elsewhere, this is actually much cleaner if you just pack the variant and unpack it here. There's a lot of extra information being carried here that isn't supported, and a lot of invalid states (e.g. is_bool = true & type != Type::integer) that indicates you're not using types sensible. As an alternative to the variant, you could wrap the variant in a `NamedParamType` simple class without a `global_var_type()` method that returns this too.

src/globalvars.cpp
```diff
@@ -258,19 +308,99 @@ void GlobalVars::verify_maps_found(
...
+    std::string val = default_val;
+
+    if (it != named_params.end()) {
+      val = it->second;
+      named_params.erase(name);
+    }
+
+    if (config.type == GlobalVarType::integer) {
+      try {
+        named_param_vals[name] = std::stoll(val);
```
amscanne: Building on the same comment elsewhere. If you store this as an actual primitive, you won't need to unparse (convert to a string) and then reparse.

src/globalvars.cpp
```diff
@@ -258,19 +308,99 @@ void GlobalVars::verify_maps_found(
...
+        named_param_vals[name] = 0;
+      } else {
+        return make_error<NamedParamError>(
+            "Command line option --" + name +
+            " is expecting a boolean (e.g. 'true'). Got: " + val);
+      }
+    } else if (config.type == GlobalVarType::string) {
+      if (val.empty()) {
+        return make_error<NamedParamError>(
+            "Command line option --" + name +
```
amscanne: I know there's been some discussion on this, but errors are not paragraphs. Don't capitalize, and don't use complete sentences that end with periods. It is a fragment that describes the error, if you want richer structure, you need to compose using the type itself, e.g. `NamedParamError` shouldn't just be a string, it can have a parameter name **and** and error message, and those can be composed in it's `log` function

src/globalvars.h
```diff
@@ -9,15 +9,56 @@
 #include <unordered_set>
 
 #include "types.h"
+#include "util/result.h"
 
 namespace bpftrace {
 
+class NamedParamError : public ErrorInfo<NamedParamError> {
```
amscanne, line 22: Why not put this in the globalvars namespace? I'm just not clear if there's a specific reason for isolating this within the context of this header.

src/run_bpftrace.cpp
```diff
@@ -34,10 +34,34 @@ void check_is_root()
   }
 }
 
-int run_bpftrace(BPFtrace &bpftrace, Output &output, BpfBytecode &bytecode)
+int run_bpftrace(BPFtrace &bpftrace,
+                 Output &output,
+                 BpfBytecode &bytecode,
+                 std::unordered_map<std::string, std::string> &&named_params,
```
amscanne: Why not just push down the parsing into this layer? Just pass the remaining arg pointer (or a std::vector<args> that are unprocessed). Then you can reduce duplication with the AOT case, which essentially has the same code right now.

src/ast/ast.h
jordalgo: I'm confused. Isn't this what you suggested? We were also going to mark maps that were only written to once as read-only and turn them into array maps/globals. `read_only` could also be another field in `struct MapInfo`

src/ast/passes/named_param.cpp
jordalgo: I think maybe this was leftover from before I had the bool type and forgot to fix it.

src/globalvars.h
jordalgo: Ack.

src/run_bpftrace.cpp
jordalgo: Because AOT and main parse the raw args slightly differently e.g. there is no `--` needed for AOT binaries - everything is in the same arg namespace.

src/aot/aot_main.cpp
jordalgo: Ack. Though now it's valid to pass an empty string if `--foo` expects a string.

src/globalvars.cpp
jordalgo: Made some updates here, let me know if it's what you had in mind.

src/run_bpftrace.cpp
jordalgo: > or a std::vector that are unprocessed

Ok, this is what I did. Good call.

src/ast/ast.h
amscanne: Right, this has nothing to do with the actual AST. It is a property that gets inferred during some set of passes, and belongs in the intermediate state. It's a minor thing, I just don't want to backslide and stuff a lot of things here (we still have a problem where the attachpoints carry way too much non-AST state).

src/ast/passes/named_param.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+  bool is_arg_error = false;
+  if (call->vargs.size() == 2) {
+    if (!call->vargs.at(1).as<Integer>() &&
+        !call->vargs.at(1).as<NegativeInteger>() &&
+        !call->vargs.at(1).as<String>() && !call->vargs.at(1).as<Boolean>()) {
+      is_arg_error = true;
+    }
+    if (is_arg_error) {
+      call->vargs.at(1).node().addError()
```
amscanne, line 48: Nit: you could roll this into the `else` case below and it will simplify the control flow a bit. Up to you.

src/globalvars.cpp
```diff
@@ -96,19 +106,30 @@ void update_global_vars_rodata(
...
+      if (config.type == Type::integer) {
+        auto *var = reinterpret_cast<int64_t *>(global_vars_buf + offset);
+        *var = std::get<int64_t>(it->second);
+      } else if (config.type == Type::boolean) {
+        auto *var = reinterpret_cast<int8_t *>(global_vars_buf + offset);
+        *var = std::get<bool>(it->second) ? 1 : 0;
+      } else if (config.type == Type::string) {
+        auto *var = reinterpret_cast<char *>(global_vars_buf + offset);
+        const auto &val = std::get<std::string>(it->second);
+        strncpy(var, val.data(), val.size());
```
amscanne: You probably want `val.c_str()` and `val.size() + 1` to ensure that it is null-terminated, no? I'm not sure how string globals are sized, but just for safety and to ensure that things like `strlen` would give you the intended results (if you call this on the value in the kernel for some reason).

src/globalvars.cpp
```diff
@@ -220,6 +240,20 @@ SizedType GlobalVars::get_sized_type(const std::string &global_var_name,
                         CreateArray(resources.max_map_key_size, CreateInt8()));
   }
 
+  if (config.type == Type::integer) {
```
amscanne, line 258: See comment below.

src/globalvars.cpp
```diff
@@ -232,23 +266,38 @@ std::unordered_set<std::string> get_section_names()
...
+{
+  if (added_global_vars_.contains(name)) {
     return;
   }
-  global_var_map_[std::move(str_name)] = GLOBAL_VAR_CONFIGS.at(name);
+  added_global_vars_[name] = GlobalVarConfig({
+      .section = std::string(RO_SECTION_NAME),
+      .type = std::holds_alternative<std::string>(default_val) ? Type::string
+              : std::holds_alternative<bool>(default_val)      ? Type::boolean
+                                                               : Type::integer,
```
amscanne: Per comment above.

You have this type now `GlobalVarValue`, and are maintaining a bijective mapping between this and `Type` in a couple places manually. You could just have this type be a class that wraps the variant and implement these methods in one place for clarity, e.g.

```
class GlobalVarValue {
public:
  using Variant = std::variant<...>;
  
  template <typename T>
  GlobalVarValue(T&& t)
    requires std::constructible_from<Variant, T>
      : variant_(std::forward<T>(t)){};

   Type get_type() const {
         ...
    }
    ...
private:
   Variant variant_;
}
```

Or at least perhaps I'm confused why there seem to be multiple ways to construct the type for the global variable? (One just the Type, and the other the full SizedType?)

(This is an optional nit, only if you feel this is helpful)

src/globalvars.cpp
```diff
@@ -258,19 +307,110 @@ void GlobalVars::verify_maps_found(
...
+        named_param_vals[name] = true;
+      }
+      continue;
+    }
+
+    const std::string &val = *it->second;
+
+    if (config.type == Type::integer) {
+      try {
+        named_param_vals[name] = std::stoll(val);
```
amscanne: Did you want to parse using the util::to_uint? (You'd need to explicitly check for negative and handle that case, but it's simple.) That way you could pass `--foo=43ns` for example.

This also lets you avoid the exception-based flow, which I think we should try to do whenever possible.

amscanne (approved):

src/ast/passes/named_param.cpp
jordalgo: I don't need the whole `is_arg_error`. Must have been a relic of a previous version.

src/globalvars.cpp
jordalgo: Ack

src/globalvars.cpp
jordalgo: I see what you mean though that class looks a bit intense just for this mapping. I'll just create a free helper function in this file.

> One just the Type, and the other the full SizedType?

I didn't want to have to serialize an entire SizedType when all I needed was the Type enum.

src/globalvars.cpp
jordalgo: Ack. I like it

src/ast/ast.h
jordalgo: Let me think more on this. I have some ideas but maybe one will emerge in the map optimization I want to make.

man/adoc/bpftrace.adoc
```diff
@@ -188,6 +188,22 @@ Print bpftrace version information.
 Enable verbose messages.
 For more details see the <<Verbose Output>> section.
 
+=== Program Options
+
+You can also pass custom options to a bpftrace program/script itself via positional or named parameters.
+All positional and named parameters should come after the script filename or a double dash if running an inline script with `-e` e.g.
```
viktormalik: This sounds like we're requiring the double dash to separate the script options from the bpftrace ones but this:
```
# bpftrace -e 'BEGIN { print(getopt("str", "")); }' --str="hello"
Attached 1 probe
hello
```
shows otherwise.

So, what's the expected behaviour here?

man/adoc/bpftrace.adoc
```diff
@@ -188,6 +188,22 @@ Print bpftrace version information.
...
+```
+# bpftrace -e 'BEGIN { ... }' -- p1 --aa=20 --bb p2
+```
+or
+```
+# bpftrace myscript.bt p1 --aa=20 --bb p2
+```
+
+In these examples there are two positional parameters (`p1`, `p2`) and two named parameters (`aa`, which is set to `20`, and `bb`, which is set to `true`).
+Named program parameters require the `=` unless they are boolean parameters (like 'bb' above).
```
viktormalik: This (requiring the `=`) is quite an unexpected requirement. Can we lift it?

man/adoc/bpftrace.adoc
viktormalik: Note that also this works which is IMO quite confusing:
```
bpftrace --str="hello" -e 'BEGIN { print(getopt("str", "")); }'
Attached 1 probe
hello
```

viktormalik (changes requested):
Didn't review the code, yet, but I think the style of specifying script args needs some clarification. See below for comments.

Also, I was able to trigger a segfault with:
```
# bpftrace -e 'BEGIN { print(getopt("str", "")); }' --str=
fish: Job 1, 'sudo build/src/bpftrace -e 'BEG…' terminated by signal SIGSEGV (Address boundary error)
```

jordalgo:
> I was able to trigger a segfault

Ah good catch. Fixed

man/adoc/bpftrace.adoc
jordalgo: So the "should" in the adoc is doing a lot of heavy lifting. This change allows users to pass named params before a double dash but encourages them to pass them after to differentiate the arg namespace. @amscanne and I [discussed this a bit](https://github.com/bpftrace/bpftrace/pull/4258#discussion_r2141692292). That being said, I think you're right and that we should update this language to remove this recommendation.

Here are some reasons for allowing script named params before a double dash:
- current positional script params (e.g. `$1`) don't have this requirement
- we can detect which named params are not supported by bpftrace itself and by the script so we can error appropriately (there are tests for this) so why force users to add a double dash
- if you have a compiled AOT binary/script the bpftrace params and the script params should be in the same namespace e.g. `$ ./my_cool_script --str=hello -f json`

@viktormalik Do you have strong objections to allowing this or is it more about the documentation needing to be made clearer?

man/adoc/bpftrace.adoc
jordalgo: We made this requirement in order to differentiate between named params and positional params e.g. this parses two ways `--foo hello`:
- one named param `foo` that is a boolean flag and one positional param `hello`
- one named param `foo` that is a string with a value `hello`

And we need to know the CLI parsing before we run the script passes that might surface what type `foo` is.

viktormalik: Right, I see why it would be difficult to get rid of the `=`. Yet, I expect users to try to use `--foo hello` as a way to pass the `hello` value to `--foo` (as that's a standard way to do so), so we should think about making their lives easier (with at least some smart warning).

I'll need to think about this more, we can discuss on the office hours.

man/adoc/bpftrace.adoc
viktormalik: > @amscanne and I [discussed this a bit](https://github.com/bpftrace/bpftrace/pull/4258#discussion_r2141692292). 

Oh, sorry, I missed that discussion.

> Here are some reasons for allowing script named params before a double dash:
> 
> * current positional script params (e.g. `$1`) don't have this requirement
> * we can detect which named params are not supported by bpftrace itself and by the script so we can error appropriately (there are tests for this) so why force users to add a double dash
> * if you have a compiled AOT binary/script the bpftrace params and the script params should be in the same namespace e.g. `$ ./my_cool_script --str=hello -f json`

Does this mean that we do not allow defining script options which conflict with bpftrace options?

jordalgo: > Does this mean that we do not allow defining script options which conflict with bpftrace options?

You can have ones that conflict but they need to come after the double dash e.g. `bpftrace -e 'BEGIN { print(getopt("help")); }' -- --help`

viktormalik: Would the same apply to AOT, i.e. would need `$ ./my_cool_script -- --help`?

jordalgo: > Would the same apply to AOT

Good question! I see that I don't even parse args after a double dash for AOT (so I need to fix that), but yes I think the same should apply. The only two long args I see for AOT are `--help` and `--version`.

man/adoc/bpftrace.adoc
jordalgo: Another option is that we don't allow boolean flags (no value) for named params e.g. we parse anything after `--arg` as belonging to the previous named param. So if you wanted a boolean you'd have to do `--foo true` or `--foo 1` or `foo=true` (this would only apply to named params defined in the script)

man/adoc/bpftrace.adoc
viktormalik: Oh, so this only applies to long args since we don't allow short script args? What if we ever want to allow them?

I feel like this needs a bit more thinking (but maybe it's been already done and a better way doesn't exist). Let's discuss at the office hours.

jordalgo:
Implemented the changes discussed offline during office hours:
- [x] require that named parameters always come after a double dash (for all cases)
- [x] require an `=` after named parameters for string and integer values
- [x] updated the adoc 

This also slightly simplified the implementation as we don't need to pass a lambda that calls `usage` into `run_bpftrace`.

man/adoc/bpftrace.adoc
```diff
@@ -188,6 +188,24 @@ Print bpftrace version information.
...
+
+// (p1, p2, 20, true) is printed
+```
+or
+```
+# bpftrace myscript.bt -- p1 --aa=20 --bb p2
+```
+
+In these examples there are two positional parameters (`p1`, `p2`) and two named parameters (`aa`, which is set to `20`, and `bb`, which is set to `true`).
+Named program parameters require the `=` unless they are boolean parameters (like 'bb' above).
```
viktormalik: Slightly better wording?
```suggestion
Named program parameters require the `=` to set their value unless they are boolean parameters (like 'bb' above).
```

src/ast/passes/named_param.h
```diff
@@ -0,0 +1,16 @@
...
+#include "types.h"
+
+namespace bpftrace::ast {
+
+class NamedParamDefaults : public ast::State<"named_params_defaults"> {
+public:
+  std::unordered_map<std::string, globalvars::GlobalVarValue> defaults;
+};
+
+Pass CreateNamedParamPass();
```
viktormalik: Super-nit:
```suggestion
Pass CreateNamedParamPass();
```
```suggestion
Pass CreateNamedParamsPass();
```
Since there may be multiple named params.

jordalgo:
> Mind moving some of the refactoring parts to separate commits?

If you really feel strongly, I will try to tease it apart but at this point it feels like a bit of a pain and late to do so. This PR has been thoroughly reviewed by @amscanne, discussed in multiple office hours, and active for over 3 weeks (I'm including [this older one](https://github.com/bpftrace/bpftrace/pull/4258)). I've also published and landed supporting PRs to simplify this one:
- https://github.com/bpftrace/bpftrace/pull/4266
- https://github.com/bpftrace/bpftrace/pull/4280

I realize there are a lot of changes but is it possible for me to add some exposition in the commit message instead that might help review?

viktormalik:
> If you really feel strongly, I will try to tease it apart but at this point it feels like a bit of a pain and late to do so. This PR has been thoroughly reviewed by @amscanne, discussed in multiple office hours, and active for over 3 weeks (I'm including [this older one](https://github.com/bpftrace/bpftrace/pull/4258)). I've also published and landed supporting PRs to simplify this one:
> 
> * [Refactor Global Vars #4266](https://github.com/bpftrace/bpftrace/pull/4266)
> * [Add boolean value #4280](https://github.com/bpftrace/bpftrace/pull/4280)
> 
> I realize there are a lot of changes but is it possible for me to add some exposition in the commit message instead that might help review?

Ok, no problem, I'm not going to push. I'll try to review the PR but may miss things as there's just too many changes. But that should be ok since it was already reviewed by @amscanne.

src/aot/aot_main.cpp
```diff
@@ -6,9 +6,11 @@
 
 #include "aot.h"
 #include "bpftrace.h"
+#include "globalvars.h"
```
viktormalik: Unnecessary import?

src/ast/passes/named_param.cpp
```diff
@@ -0,0 +1,124 @@
...
+  BPFtrace &bpftrace_;
+};
+
+void NamedParamPass::visit(Expression &expr)
+{
+  auto *call = expr.as<Call>();
+  if (!call || call->func != "getopt") {
+    Visitor<NamedParamPass>::visit(expr);
+    return;
+  }
```
viktormalik, line 35: I don't understand this code. Why not use `NamedParamPass::visit(Call &call)`instead of this? And what's the recursion for?

src/globalvars.cpp
```diff
@@ -96,19 +121,30 @@ void update_global_vars_rodata(
...
   for (const auto &[global_var, offset] : vars_and_offsets) {
-    auto *var = reinterpret_cast<int64_t *>(global_vars_buf + offset);
-
-    auto it = known_global_var_values.find(global_var);
-    if (it != known_global_var_values.end()) {
-      *var = it->second;
+    auto it = global_var_vals.find(global_var);
+    if (it != global_var_vals.end()) {
+      const auto &config = added_global_vars.at(global_var);
+      if (config.type == Type::integer) {
```
viktormalik, line 127: Maybe use `switch` instead of this `if`-`else if`-`else`? We'd get some compiler checks for free.

src/globalvars.cpp
```diff
@@ -220,6 +255,20 @@ SizedType GlobalVars::get_sized_type(const std::string &global_var_name,
                         CreateArray(resources.max_map_key_size, CreateInt8()));
   }
 
+  if (config.type == Type::integer) {
```
viktormalik, line 258: Same as above, use `switch`? I think that would allow to get rid of the unused `return CreateInt64()` at the end.

src/globalvars.cpp
```diff
@@ -258,19 +320,143 @@ void GlobalVars::verify_maps_found(
...
+        if (!int_val) {
+          return make_error<NamedParamError>(name, val, "expects an integer");
+        }
+        // Converting without overflow.
+        auto neg = -static_cast<int64_t>(*int_val - 1) - 1;
+        if (neg > 0) {
+          return make_error<NamedParamError>(name,
+                                             val,
+                                             "value is out of range");
+        }
```
viktormalik: This should be IMHO moved to `util::to_int`.

tests/globalvars.cpp
```diff
@@ -0,0 +1,105 @@
...
+  EXPECT_FALSE(bad_values1);
+  bool has_unknown_opts_error = false;
+  auto ok2 = handleErrors(
+      std::move(bad_values1), [&](const UnknownParamError& uo_err) {
+        EXPECT_THAT(uo_err.err(),
+                    HasSubstr(
+                        "unexpected program command line options: --random"));
+        has_unknown_opts_error = true;
+      });
+  EXPECT_TRUE(has_unknown_opts_error);
```
viktormalik, line 74: This test pattern repeats multiple times below. How about factoring it out to a function?

viktormalik (commented):
Couple of ideas for improvement, otherwise looks good to me.

src/ast/passes/named_param.cpp
jordalgo: We can't use `visit(Call` because we might be returning an expression of a different type e.g. `MapAccess`. And we need the recursion because Calls can exist as params of Calls e.g. things inside of `print`

src/globalvars.cpp
jordalgo: I considered that but we're only using a small subset of the `Type` enum from sized type and so we'd have A LOT of boilerplate to get some checks about non-exhaustive type checks.

src/globalvars.cpp
jordalgo: Mentioned this above but most of the types are unused.

src/globalvars.cpp
jordalgo: Yeah, I agree. Let me see about refactoring.

tests/globalvars.cpp
jordalgo: Ack.

src/ast/passes/named_param.cpp
viktormalik: > We can't use `visit(Call` because we might be returning an expression of a different type e.g. `MapAccess`.

Ah, right, I see it now.

> And we need the recursion because Calls can exist as params of Calls e.g. things inside of `print`

This still confuses me a bit but I guess I just need to look at the entire visitor code better.

Thanks!

src/globalvars.cpp
viktormalik: My idea was that we'd handle the rest via `default` but in that case, it probably doesn't matter if we use `if` chain or `switch`.

tests/globalvars.cpp
jordalgo: Done. Well, specifically, the part I refactored was the NamedParamError

src/globalvars.cpp
jordalgo: Pulled it into the int_parser util

viktormalik (approved):
Nice feature!

jordalgo:
Thanks for the review @amscanne and @viktormalik ! I think we landed on something that fits nicely into our language/tool goals.

PR #4272, opened 2025-06-17
codegen: Use a single `LLVMContext`

amscanne:
Stacked PRs:
 * #4276
 * #4275
 * #4274
 * #4273
 * __->__#4272

--- --- ---

### codegen: Use a single `LLVMContext`

I was struggling with various failures during linking modules, which
were likely a result of the two modules have two different type systems
and two different sets of intrinsics. Using the same `LLVMContext`
resolves these problems, and was a likely user error.

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -5101,16 +5101,20 @@ Pass CreateLinkBitcodePass()
   return Pass::create(
       "LinkBitcode", [](BitcodeModules &bm, CompiledModule &cm) -> Result<> {
         for (auto &mod : bm.modules) {
-          // Modify to ensure everything is inlined. Note that
-          // this is also marking all these functions for
-          // below, which will adjust their linkage.
+          // Make a copy of the module, to ensure this is not modifying the
+          // original. The link step must consume the module below.
+          auto copy = llvm::CloneModule(*mod);
+
+          // Modify to ensure everything is inlined. Note that this is also
+          // marking all these functions for below, which will adjust their
+          // linkage.
           //
-          // We also want to ensure that we remove any
-          // attributes that prevent any subsequent inline
-          // (such as "OptimizeNone"), and suitably tag these
-          // functions are "NoUnwind", like the rest.
-          for (auto &fn : mod->functions()) {
-            if (fn.isDSOLocal()) {
+          // We also want to ensure that we remove any attributes that prevent
+          // any subsequent inline (such as "OptimizeNone"), and suitably tag
+          // these functions are "NoUnwind", like the rest. There is no
+          // unwinding in BPF, and these attributes are just ensuring that.
+          for (auto &fn : copy->functions()) {
+            if (fn.isDSOLocal() && !fn.isIntrinsic()) {
               fn.removeFnAttr(Attribute::NoInline);
               fn.removeFnAttr(Attribute::OptimizeNone);
               fn.addFnAttr(Attribute::AlwaysInline);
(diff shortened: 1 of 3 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -5101,16 +5101,22 @@ Pass CreateLinkBitcodePass()
   return Pass::create(
       "LinkBitcode", [](BitcodeModules &bm, CompiledModule &cm) -> Result<> {
         for (auto &mod : bm.modules) {
-          // Modify to ensure everything is inlined. Note that
-          // this is also marking all these functions for
-          // below, which will adjust their linkage.
+          // Make a copy of the module, to ensure that this
+          // specific is not modifying the original. The link
+          // step must consume the module below.
```
jordalgo: nit
```suggestion
          // Make a copy of the module, to ensure this
          // is not modifying the original. The link
          // step must consume the module below.
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -5101,16 +5101,22 @@ Pass CreateLinkBitcodePass()
...
-          // (such as "OptimizeNone"), and suitably tag these
-          // functions are "NoUnwind", like the rest.
-          for (auto &fn : mod->functions()) {
-            if (fn.isDSOLocal()) {
+          //  We also want to ensure that we remove any
+          //  attributes that prevent any subsequent inline
+          //  (such as "OptimizeNone"), and suitably tag these
+          //  functions are "NoUnwind", like the rest.
+          for (auto &fn : copy->functions()) {
+            std::cerr << fn.getName().str() << ": " << fn.isIntrinsic() << "\n";
```
jordalgo: You want to keep this err log msg?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -5101,16 +5101,22 @@ Pass CreateLinkBitcodePass()
...
-          // We also want to ensure that we remove any
-          // attributes that prevent any subsequent inline
-          // (such as "OptimizeNone"), and suitably tag these
-          // functions are "NoUnwind", like the rest.
-          for (auto &fn : mod->functions()) {
-            if (fn.isDSOLocal()) {
+          //  We also want to ensure that we remove any
+          //  attributes that prevent any subsequent inline
+          //  (such as "OptimizeNone"), and suitably tag these
+          //  functions are "NoUnwind", like the rest.
```
jordalgo: Just curious, why do we want "NoUnwind" for our stdlib funcs?

src/ast/passes/codegen_llvm.cpp
```diff
@@ -5122,23 +5128,14 @@ Pass CreateLinkBitcodePass()
...
         }
 
-        // Remark all defined functions as having internal
-        // linkage. This is using all the functions marked
-        // above. It doesn't really make sense to have
-        // `always_inline` but be an external linkage.
-        for (auto &fn : cm.module->functions()) {
-          if (fn.hasFnAttribute(Attribute::AlwaysInline)) {
-            fn.setLinkage(llvm::Function::InternalLinkage);
-            llvm::stripDebugInfo(fn);
```
jordalgo, line 5138: We don't want to continue stripping debuginfo for our inlined funcs or are we just not generating it anymore?

amscanne: This was one of the crashes that was caused by the different contexts, I believe. We actually need debug info for BTF.

src/ast/passes/codegen_llvm.cpp
amscanne: Added a bit of exposition there.

src/ast/passes/codegen_llvm.cpp
amscanne: Nice catch, removed.

src/ast/passes/codegen_llvm.cpp
amscanne: Done.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -5101,16 +5101,23 @@ Pass CreateLinkBitcodePass()
...
           //
           // We also want to ensure that we remove any
           // attributes that prevent any subsequent inline
           // (such as "OptimizeNone"), and suitably tag these
-          // functions are "NoUnwind", like the rest.
-          for (auto &fn : mod->functions()) {
-            if (fn.isDSOLocal()) {
+          // functions are "NoUnwind", like the rest. There
+          // is no unwinding in BPF, and this is just ensuring
+          // that the attributes of these functions.
```
jordalgo: super nit
```suggestion
          // is no unwinding in BPF, and these attributes are just ensuring that.
```

PR #4274, opened 2025-06-17
stdlib: add required headers for `libbpf` headers

amscanne:
Stacked PRs:
 * #4276
 * #4332
 * #4331
 * __->__#4274

--- --- ---

### stdlib: add required headers for `libbpf` headers

We need to generate the type headers from BPF, but `usdt.bpf.h` imports
some standard system headers. As long as these are limited, it seems
reasonable and we can simply vendor them and make them available.

### DIFF
--- a/src/stdlib/include/asm-generic/errno-base.h
+++ b/src/stdlib/include/asm-generic/errno-base.h
@@ -1,0 +1,40 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+#ifndef _ASM_GENERIC_ERRNO_BASE_H
+#define _ASM_GENERIC_ERRNO_BASE_H
+
+#define	EPERM		 1	/* Operation not permitted */
+#define	ENOENT		 2	/* No such file or directory */
+#define	ESRCH		 3	/* No such process */
+#define	EINTR		 4	/* Interrupted system call */
+#define	EIO		 5	/* I/O error */
+#define	ENXIO		 6	/* No such device or address */
+#define	E2BIG		 7	/* Argument list too long */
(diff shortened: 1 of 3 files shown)


jordalgo:
Should we add a script that can sync/update these and perhaps a bit in a readme

src/stdlib/include/linux/bpf.h
```diff
@@ -0,0 +1,5060 @@
...
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of version 2 of the GNU General Public
+ * License as published by the Free Software Foundation.
+ */
+#ifndef __LINUX_BPF_H__
+#define __LINUX_BPF_H__
+
+#include <linux/types.h>
+#include <linux/bpf_common.h>
```
viktormalik: How come we don't need to vendor these two as well?

amscanne: Hm, yes let me take a pass at this. I think that our Nix build may be vendoring some things transitively from LIBBPF. I want to vendor the same basic set, and I'll need to add something like the [sync_kernel.sh](https://github.com/libbpf/libbpf/blob/master/scripts/sync-kernel.sh) from libbpf.

amscanne: I'll limit the vendoring to `errno.h` (needed for `usdt.bpf.h`) and tackle the broader vendoring/sync bits in a later change.

viktormalik: Sure, looks good now!

viktormalik (approved):

amscanne:
I've stripped this down to the minimum header. I can revisit the script question if/when there's more that needs to be vendored. I want to unblock the migration.

PR #4277, opened 2025-06-18
Make EVENT_LOSS_COUNTER like other rw globals

jordalgo:
I couldn't see any performance difference, even
though I expected one, by using a "slotted"
(for each CPU) array map rather than atomically
updating a single global but this removes some
bespoke code to handle the EVENT_LOSS_COUNTER
in globalvars.

##### Checklist


src/globalvars.cpp
```diff
@@ -255,6 +255,10 @@ SizedType GlobalVars::get_sized_type(const std::string &global_var_name,
                         CreateArray(resources.max_map_key_size, CreateInt8()));
   }
 
+  if (global_var_name == EVENT_LOSS_COUNTER) {
+    return make_rw_type(1, CreateArray(8, CreateInt8()));
```
viktormalik: Why the `[8 x i8]` instead of `i64`?

src/bpftrace.cpp
```diff
@@ -994,7 +994,13 @@ int BPFtrace::poll_skboutput_events()
 
 void BPFtrace::poll_event_loss(Output &out)
 {
-  auto current_value = bytecode_.get_event_loss_counter(*this);
+  uint64_t current_value = 0;
+  uint64_t *current_values = bytecode_.get_event_loss_counter(*this);
+  for (int i = 0; i < max_cpu_id_; ++i) {
+    current_value += *current_values;
+    current_values++;
+  }
```
viktormalik: It feels like the aggregation should be done in `bytecode_.get_event_loss_counter`, shouldn't it?

viktormalik (commented):
I'm not very familiar with the global vars code so please forgive me if the answers to the questions below are obvious.

src/globalvars.cpp
jordalgo: It's because I didn't want to have forking logic in `createScratchBuffer` in irbuilder but that's perhaps a silly reason.  I'll update this so it's a i64 and just add the check in `createScratchBuffer`.

viktormalik (approved):
I still feel like we're missing a better comment in `createScratchBuffer` explaining all the per-cpu magic but otherwise this looks reasonable to me. Another review would probably be useful, though.

jordalgo:
@viktormalik There is this bit: https://github.com/bpftrace/bpftrace/blob/master/src/ast/irbuilderbpf.cpp#L688-L691
Do you think that needs more clarification?

viktormalik:
> @viktormalik There is this bit: https://github.com/bpftrace/bpftrace/blob/master/src/ast/irbuilderbpf.cpp#L688-L691 Do you think that needs more clarification?

Well, that one needs some clarification by itself, like what's the trailing 0 index for? Your updated comment explains it a bit better - I assume that `ValueType` is either a string (by default) or a scalar int (for event loss counter). So the trailing zero is to get a pointer to the first character in the string (if it is a string).

But what I miss more is the motivation *why* global values are represented the way they are. The comment says we use `ValueType var[MAX_CPU_ID + 1][num_elements]` by why such a structure was chosen? Perhaps it's explained somewhere and we just need a pointer here.

jordalgo:
@viktormalik I tried to improve the comment a bit in `createScratchBuffer`. Let me know if it makes sense.

viktormalik:
> @viktormalik I tried to improve the comment a bit in `createScratchBuffer`. Let me know if it makes sense.

Oh yes, this is much better, thank you!

PR #4278, opened 2025-06-18
Add issue metrics workflow

jordalgo:
Taken from here:
https://github.com/github/issue-metrics

##### Checklist


jordalgo:
Experimenting with this.

.github/workflows/issue-metrics.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Get dates for last month
+        shell: bash
+        run: |
+          # Calculate the first day of the previous month
+          first_day=$(date -d "last month" +%Y-%m-01)
+
+          # Calculate the last day of the previous month
+          last_day=$(date -d "$first_day +1 month -1 day" +%Y-%m-%d)
+
+          #Set an environment variable with the date range
```
amscanne, line 25: :scream:

.github/workflows/issue-metrics.yml
```diff
@@ -0,0 +1,42 @@
...
+          echo "$first_day..$last_day"
+          echo "last_month=$first_day..$last_day" >> "$GITHUB_ENV"
+
+      - name: Run issue-metrics tool
+        uses: github/issue-metrics@v3
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SEARCH_QUERY: 'repo:bpftrace/bpftrace created:${{ env.last_month }} -reason:"not planned"'
+
+      - name: Create issue
```
amscanne: I wouldn't create issue spam, that will make everyone hate this. Maybe make this the output of the job: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary

I think this would be a custom step that does `cp issue_metrics.md $GITHUB_STEP_SUMMARY`

jordalgo: If we made this an output of the job instead of an issue, how would we know it got run/generated at all? You'd have to go search for it no? Also, I don't consider running it once a month to be spammy

amscanne: You would just go to "Actions" and click this specific workflow.

I don't have hard objections to the issue, but maybe let's just start by generating and clicking, and if it's a good/useful report we can make it publish.

jordalgo: Ok fine. I can do that. 

Edit: Done.

amscanne (approved):

viktormalik (approved):
Agreed with @amscanne, I prefer not to create an issue.

PR #4279, opened 2025-06-19
Codegen: Mark some helpers as "pure" so that LLVM can optimise calls to them

ajor:
If a BPF helper has no side effects and always returns the same value when called repeatedly within a single probe's context, we can get LLVM to optimise out repeated calls to the same helper.

Most of these helpers are not inlined in the current latest kernel, making the cost of repeated helper calls higher.

One common pattern which is helped is recording both the `pid` and the `tid`. Since `pid` and `tid` are both implemented with the `get_current_pid_tgid` helper, this script would have previously required two identical helper calls, but now only one is needed: `print((pid, tid, ...))`

Concise example script:
```
BEGIN { pid; tid; }
```

Old IR (optimised):
```
define noundef i64 @BEGIN_1(ptr nocapture readnone %0) local_unnamed_addr #0 section "s_BEGIN_1" !dbg !29 {
entry:
  %get_pid_tgid = tail call i64 inttoptr (i64 14 to ptr)() #0
  %get_pid_tgid1 = tail call i64 inttoptr (i64 14 to ptr)() #0
  ret i64 0
}
attributes #0 = { nounwind }
```

New IR (optimised):
```
define noundef i64 @BEGIN_1(ptr nocapture readnone %0) local_unnamed_addr #0 section "s_BEGIN_1" !dbg !29 {
entry:
  %get_pid_tgid = tail call i64 inttoptr (i64 14 to ptr)() #1
  ret i64 0
}
attributes #0 = { nofree nosync nounwind memory(none) }
attributes #1 = { nounwind memory(none) }
```

##### Checklist

### DIFF
--- a/src/ast/irbuilderbpf.cpp
+++ b/src/ast/irbuilderbpf.cpp
@@ -440,9 +440,26 @@ llvm::Type *IRBuilderBPF::GetMapValueType(const SizedType &stype)
   return ty;
 }
 
+/// Creates a call to a BPF helper function
+///
+/// A call to a helper function can be marked as "pure" to allow LLVM to
+/// optimise around it.
+///
+/// ** BE VERY CAREFUL when marking a helper function as pure - marking an
+/// impure helper as pure can result in undefined behaviour. **
+///
+/// Guidelines for deciding if a helper can be considered pure:
+/// - It must always return the same value when called repeatedly with the same
+///   arguments within a single probe's context
+/// - It must not read or write any memory in the BPF address space (e.g. it
+///   mustn't take any pointers to BPF memory as parameters)
+/// - It must not have any intentional side effects outside of the BPF address
+///   space, otherwise these may be optimised out, e.g. pushing to the ring
+///   buffer, signalling a process
 CallInst *IRBuilderBPF::CreateHelperCall(libbpf::bpf_func_id func_id,
                                          FunctionType *helper_type,
                                          ArrayRef<Value *> args,
+                                         bool is_pure,
                                          const Twine &Name,
                                          const Location &loc)
 {
@@ -451,7 +468,14 @@ CallInst *IRBuilderBPF::CreateHelperCall(libbpf::bpf_func_id func_id,
   Constant *helper_func = ConstantExpr::getCast(Instruction::IntToPtr,
                                                 getInt64(func_id),
                                                 helper_ptr_type);
-  return createCall(helper_type, helper_func, args, Name);
+  CallInst *call = createCall(helper_type, helper_func, args, Name);
+
+  // When we tell LLVM that this function call "does not access memory", this
+  // only refers to BPF memory. Accessing kernel or user memory is fine within
+  // a "pure helper", as we can consider kernel/user memory as constant.
+  if (is_pure)
+    call->setDoesNotAccessMemory();
+  return call;
 }
 
 CallInst *IRBuilderBPF::createCall(FunctionType *callee_type,
@@ -1803,7 +1827,7 @@ CallInst *IRBuilderBPF::CreateGetNs(TimestampMode ts, const Location &loc)
   // u64 ktime_get_*ns()
   // Return: current ktime
   FunctionType *gettime_func_type = FunctionType::get(getInt64Ty(), false);
-  return CreateHelperCall(fn, gettime_func_type, {}, "get_ns", loc);
+  return CreateHelperCall(fn, gettime_func_type, {}, false, "get_ns", loc);
 }
 
 CallInst *IRBuilderBPF::CreateJiffies64(const Location &loc)
@@ -1811,8 +1835,12 @@ CallInst *IRBuilderBPF::CreateJiffies64(const Location &loc)
   // u64 bpf_jiffies64()
   // Return: jiffies (BITS_PER_LONG == 64) or jiffies_64 (otherwise)
   FunctionType *jiffies64_func_type = FunctionType::get(getInt64Ty(), false);
-  return CreateHelperCall(
-      libbpf::BPF_FUNC_jiffies64, jiffies64_func_type, {}, "jiffies64", loc);
+  return CreateHelperCall(libbpf::BPF_FUNC_jiffies64,
+                          jiffies64_func_type,
+                          {},
+                          false,
+                          "jiffies64",
+                          loc);
 }
 
 Value *IRBuilderBPF::CreateIntegerArrayCmp(Value *val1,
@@ -1935,11 +1963,13 @@ CallInst *IRBuilderBPF::CreateGetPidTgid(const Location &loc)
   // u64 bpf_get_current_pid_tgid(void)
   // Return: current->tgid << 32 | current->pid
   FunctionType *getpidtgid_func_type = FunctionType::get(getInt64Ty(), false);
-  return CreateHelperCall(libbpf::BPF_FUNC_get_current_pid_tgid,
-                          getpidtgid_func_type,
-                          {},
-                          "get_pid_tgid",
-                          loc);
+  auto *res = CreateHelperCall(libbpf::BPF_FUNC_get_current_pid_tgid,
+                               getpidtgid_func_type,
+                               {},
+                               true,
+                               "get_pid_tgid",
+                               loc);
+  return res;
 }
 
 void IRBuilderBPF::CreateGetNsPidTgid(Value *dev,
@@ -1964,6 +1994,7 @@ void IRBuilderBPF::CreateGetNsPidTgid(Value *dev,
   CallInst *call = CreateHelperCall(libbpf::BPF_FUNC_get_ns_current_pid_tgid,
                                     getnspidtgid_func_type,
                                     { dev, ino, ret, getInt32(struct_size) },
+                                    false,
                                     "get_ns_pid_tgid",
                                     loc);
   CreateHelperErrorCond(call, libbpf::BPF_FUNC_get_ns_current_pid_tgid, loc);
@@ -1987,6 +2018,7 @@ CallInst *IRBuilderBPF::CreateGetCurrentCgroupId(const Location &loc)
   return CreateHelperCall(libbpf::BPF_FUNC_get_current_cgroup_id,
                           getcgroupid_func_type,
                           {},
+                          true,
                           "get_cgroup_id",
                           loc);
 }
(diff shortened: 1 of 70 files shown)


ajor:
Anyone have any ideas for a better way to test this particular change?

jordalgo:
Well it seems our current runtime tests are catching something 😄

src/ast/irbuilderbpf.cpp
```diff
@@ -451,7 +452,15 @@ CallInst *IRBuilderBPF::CreateHelperCall(libbpf::bpf_func_id func_id,
...
                                                 helper_ptr_type);
-  return createCall(helper_type, helper_func, args, Name);
+  CallInst *call = createCall(helper_type, helper_func, args, Name);
+
+  // Tell LLVM if a BPF helper is a pure function (i.e. has no side effects and
+  // always returns the same value when called repeatedly within a single
+  // probe's context), so that it can optimise out repeated calls to the same
+  // helper.
+  if (is_pure)
+    call->setDoesNotAccessMemory();
```
amscanne, line 477: Thinking out loud: as an alternative to doing this at the codegen level, did you consider just having a pass that lifts a set of builtins/calls and memoizes their values? E.g. a pass that simply turns

```
$x = pid;
$y = pid;
```

Into

```
let __pid = pid;
$x = __pid; // N.B. Var name is not usable by user
$y = __pid;
```

The reason is that then we could potentially have a way to annotate any macro/builtin as memoizable, even if it does complex operations. For example: this is already true for pid/tid, which take that value and applies a mask (it's a simple example, but I'm sure there are others). I believe that this is a more powerful/general ability than marking specific functions in codegen (but does require that we introduce wrappers).

I'd be a bit hestitant to change what would be the canonical prototype of the helper from the C side. In the future, it would be good if we could link some C code and automatically get all the helpers/kfuncs available in the namespace without hard-coding any specifications, and I wouldn't really want to go and tweak attrs on a specific set of them.

Admittedly this doesn't capture the specific `pid`/`tid` case (dedupping the call to `get_pid_tgid`), but you could imagine this specific mechanism catching this via composability e.g. (assuming that you can specify builtins as no-arg macros, and memoize will automatically lift and capture those at callsites)

```
memoize macro pid_tgid() {
   __get_pid_tgid()
}

memoize macro pid() {
   (pid_tgid() & 0xffff0000) >> 32
}

memoize macro tid() {
  pid_tgid() & 0xffff
}
```

jordalgo:
Looks like the `comm` based codegen tests need to be updated.

src/ast/irbuilderbpf.cpp
```diff
@@ -443,6 +443,7 @@ llvm::Type *IRBuilderBPF::GetMapValueType(const SizedType &stype)
 CallInst *IRBuilderBPF::CreateHelperCall(libbpf::bpf_func_id func_id,
                                          FunctionType *helper_type,
                                          ArrayRef<Value *> args,
+                                         bool is_pure,
```
jordalgo, line 462: nit: should we just make this the last argument and have it default to false?

src/ast/irbuilderbpf.cpp
jordalgo: Related question, other than the `pid`/`tid` case do we see a lot of instances in the wild where people are calling these builtins/funcs multiple times without saving the result into a scratch variable? What about just staring with something simple like a `get_pid_tid` function or builtin that returns a tuple of pid and tid?

ajor: > Thinking out loud: as an alternative to doing this at the codegen level, did you consider just having a pass that lifts a set of builtins/calls and memoizes their values?

Yeah, that was actually what I was thinking of doing when I first came across this issue. Decided against it in the end because it'd be far more code than the LLVM approach and doesn't easily handle the `pid`/`tid` case.

> The reason is that then we could potentially have a way to annotate any macro/builtin as memoizable, even if it does complex operations.

After teaching LLVM about these opaque BPF helpers, I don't think this is necessary. It can propagate "pure-ness" and other attributes up through function calls and it will automatically memoise anything it can. [Example of attribute propagation](https://godbolt.org/z/8Gd7P3e9f). Here I manually tag `foo()` with `memory(none)` and we can see that `bar()` is also tagged with the same attribute. `bar()` is then automatically memoised when called twice in `baz()`.

Doesn't seem worth re-inventing optimisations that LLVM already implements.

ajor: > Related question, other than the pid/tid case do we see a lot of instances in the wild where people are calling these builtins/funcs multiple times without saving the result into a scratch variable?

`tid` being called multiple times in a probe is a very common thing - it's part of the standard pattern of implementing per-thread variables in BpfScript. e.g. this snippet calls `tid` twice:
```
if (@myvar[tid]) { print(@myvar[tid]) }
```


As this isn't a combined `pid`/`tid` lookup, it wouldn't be helped by a combined builtin.

Here are a couple of concrete examples I found in our repo:
https://github.com/bpftrace/bpftrace/blob/ae53e838054f5b462cdc46dd6b6f643d53a1feca/tools/opensnoop.bt#L58-L66
https://github.com/bpftrace/bpftrace/blob/ae53e838054f5b462cdc46dd6b6f643d53a1feca/tools/setuids.bt#L55-L59

I can also see a couple of script's in Meta's internal collection which call the `cpu` builtin multiple times in a probe.

> What about just staring with something simple like a get_pid_tid function or builtin that returns a tuple of pid and tid?

What's the benefit? bpftrace is all about providing an easy way to do the common things and hiding the nasty implementation details from the user (like `pid` and `tid` being bundled together by the BPF helper).

ajor:
Updated to mark a couple of extra helpers as impure (`get_ns_current_pid_tgid` and `get_stack`) as they write their results to a provided pointer parameter. Tests are passing now!

src/ast/irbuilderbpf.cpp
ajor: Of the 18 helpers currently, 9 are pure and 9 aren't. Doesn't seem like it'd save anything to make this a default argument. I stuck it in the middle of the list to try to keep the convention of `name` and `loc` being the final two parameters.

src/ast/irbuilderbpf.cpp
jordalgo: > tid being called multiple times in a probe is a very common thing 

Yeah true. I hadn't thought about that case.

> Decided against it in the end because it'd be far more code than the LLVM approach and doesn't easily handle the pid/tid case.

I think I agree here. Introducing a memoized concept into macros just for this change seems premature and overly complicated. As long as we're pretty sure these marked functions are actually pure in the context of a BPF prog function call then this change is fine with me.

amscanne: My concern with the marking of helper declarations as pure is that we are effectively doing something that won't align with what is declared by imported C code (so C usage of kfuncs won't link in the expected way against internal usage of kfuncs), and will need to maintain this in the future. This means frequently reaching down into codegen to tweak the magic parameter that has the potential to explode if used incorrectly (just like it did with the tests failing, but it's easy to imagine that mistake in paths not covered by tests, when e.g. a new contributor adds support for some other random kfunc).

> As long as we're pretty sure these marked functions are actually pure in the context of a BPF prog

I don't believe that this is actually possible, as these are all external functions. Very few of them are actually pure, as they often need to acquire locks and do other things, even if they aren't generally mutating state. The question is whether the surface can be well-understood enough that it's justified (I think it can, but I'm concerned about doing so via ad-hoc parameters in codegen internals).

Maybe this kind of thing would be more supportable if we pulled out and generalized kfuncs & helpers so we could annotate the "pureness" of these cases and modify it if this changes for any cases (and also patch up the attributes on imported modules). If owned a layer that was doing kfunc/helper defs for linked C code and our own stuff, we could unify this flag and make changes more prominent. (It doesn't have to be at the macro level, but I'm advocating for something at a more obvious layer than just where the declaration is generated)

jordalgo: Going to give @amscanne a chance to voice any strong objections otherwise I'll stamp tomorrow.

ajor: > My concern with the marking of helper declarations as pure is that we are effectively doing something that won't align with what is declared by imported C code (so C usage of kfuncs won't link in the expected way against internal usage of kfuncs), and will need to maintain this in the future.

I don't think I follow you here. How is marking helper calls as pure incompatible with imported C code?

> This means frequently reaching down into codegen to tweak the magic parameter that has the potential to explode if used incorrectly (just like it did with the tests failing, but it's easy to imagine that mistake in paths not covered by tests, when e.g. a new contributor adds support for some other random kfunc).

The magic parameter only needs to be set for magic builtins, which aren't something that are frequently added.

> > As long as we're pretty sure these marked functions are actually pure in the context of a BPF prog
> 
> I don't believe that this is actually possible, as these are all external functions. Very few of them are actually pure, as they often need to acquire locks and do other things, even if they aren't generally mutating state.

They only need to be pure from the point of view of the BPF bytecode that we're generating. Any kernel resources they acquire/mutate that aren't visible to a BPF program don't cause problems with this LLVM optimisation.

> Maybe this kind of thing would be more supportable if we pulled out and generalized kfuncs & helpers so we could annotate the "pureness" of these cases and modify it if this changes for any cases (and also patch up the attributes on imported modules). If owned a layer that was doing kfunc/helper defs for linked C code and our own stuff, we could unify this flag and make changes more prominent. (It doesn't have to be at the macro level, but I'm advocating for something at a more obvious layer than just where the declaration is generated)

The original intention of IRBuilderBPF was to simply be a layer that hid away the complexities of calling BPF helpers, nothing more. (although it has since been extended with stuff that I think should probably be in Codegen rather than IRBuilder)

ajor:
Added an expanded comment about the dangers of marking helpers as pure

jordalgo (approved):

ajor:
I was wondering about them myself, but in what circumstances can that memory change within the context of a single BPF program?

jordalgo:
> I was wondering about them myself, but in what circumstances can that memory change within the context of a single BPF program?

I'm not sure myself. These were just the two @anakryiko flagged. Probably better to just be safe and not mark them as pure for the time being, we can always add them later when we feel more confident.

EDIT: I talked with Andrii again and it seems he's ok with those two functions being marked as pure so I'll re-add my approval.

jordalgo (approved):

PR #4280, opened 2025-06-20
Add boolean value

jordalgo:
This is so users can actually write out:
`true` and `false` instead of `1` or `0`.

This value is represented as a 8 bit int
in codegen.

This adds additional supporting for casting
to a bool type e.g. for ints, strings and
pointers.

This also fixes functions returning boolean
values e.g. `strcontains` and `has_key`,
which is technicially a breaking change but
should hopefully not have any real-world issues.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -347,6 +347,9 @@ the type upon declaration.
 |*Type*
 |*Description*
 
+|bool
+|`true` or `false`
+
 |uint8
 |Unsigned 8 bit integer
 
@@ -3813,17 +3816,17 @@ If there are many processes running, it will consume a lot of a memory.
 
 ==== cpp_demangle
 
-Default: 1
+Default: true
 
 C++ symbol demangling in userspace stack traces is enabled by default.
 
-This feature can be turned off by setting the value of this environment variable to `0`.
+This feature can be turned off by setting the value of this variable to `false`.
 
 ==== lazy_symbolication
 
-Default: 0
+Default: false
 
-For user space symbols, symbolicate lazily/on-demand (1) or symbolicate everything ahead of time (0).
+For user space symbols, symbolicate lazily/on-demand (`true`) or symbolicate everything ahead of time (`false`).
 
 ==== license
 
@@ -3907,7 +3910,7 @@ The tradeoff is that bpftrace will use more memory.
 
 ==== show_debug_info
 
-This is only available if the link:https://github.com/libbpf/blazesym[Blazesym] library is available at build time. If it is available this defaults to 1, meaning that when printing ustack and kstack symbols bpftrace will also show (if debug info is available) symbol file and line ('bpftrace' stack mode) and a label if the function was inlined ('bpftrace' and 'perf' stack modes).
+This is only available if the link:https://github.com/libbpf/blazesym[Blazesym] library is available at build time. If it is available this defaults to `true`, meaning that when printing ustack and kstack symbols bpftrace will also show (if debug info is available) symbol file and line ('bpftrace' stack mode) and a label if the function was inlined ('bpftrace' and 'perf' stack modes).
 There might be a performance difference when symbolicating, which is the only reason to disable this.
 
 ==== stack_mode
@@ -3932,9 +3935,9 @@ Set to empty string to disable truncation trailers.
 
 ==== print_maps_on_exit
 
-Default: 1
+Default: true
 
-Controls whether maps are printed on exit. Set to `0` in order to change the default behavior and not automatically print maps at program exit.
+Controls whether maps are printed on exit. Set to `false` in order to change the default behavior and not automatically print maps at program exit.
 
 ==== unstable_macro
 
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -116,6 +116,40 @@ std::string opstr(const Unop &unop)
   return {}; // unreached
 }
 
+bool is_comparison_op(Operator op)
+{
+  switch (op) {
+    case Operator::EQ:
+    case Operator::NE:
+    case Operator::LE:
+    case Operator::GE:
+    case Operator::LT:
+    case Operator::GT:
+    case Operator::LAND:
+    case Operator::LOR:
+      return true;
+    case Operator::PLUS:
+    case Operator::MINUS:
+    case Operator::MUL:
+    case Operator::DIV:
+    case Operator::MOD:
+    case Operator::BAND:
+    case Operator::BOR:
+    case Operator::BXOR:
+    case Operator::LEFT:
+    case Operator::RIGHT:
+    case Operator::INVALID:
+    case Operator::ASSIGN:
+    case Operator::INCREMENT:
+    case Operator::DECREMENT:
+    case Operator::LNOT:
+    case Operator::BNOT:
+      return false;
+  }
+
+  return false; // unreached
+}
+
 AttachPoint &AttachPoint::create_expansion_copy(ASTContext &ctx,
                                                 const std::string &match) const
 {
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -214,6 +214,7 @@ public:
   using Visitor<CodegenLLVM, ScopedExpr>::visit;
   ScopedExpr visit(Integer &integer);
   ScopedExpr visit(NegativeInteger &integer);
+  ScopedExpr visit(Boolean &boolean);
   ScopedExpr visit(String &string);
   ScopedExpr visit(Identifier &identifier);
   ScopedExpr visit(Builtin &builtin);
@@ -529,6 +530,11 @@ ScopedExpr CodegenLLVM::visit(Integer &integer)
 ScopedExpr CodegenLLVM::visit(NegativeInteger &integer)
 {
   return ScopedExpr(b_.getInt64(integer.value));
+}
+
+ScopedExpr CodegenLLVM::visit(Boolean &boolean)
+{
+  return ScopedExpr(b_.getInt8(boolean.value ? 1 : 0));
 }
 
 ScopedExpr CodegenLLVM::visit(String &string)
@@ -2246,7 +2252,7 @@ ScopedExpr CodegenLLVM::unop_ptr(Unop &unop)
 ScopedExpr CodegenLLVM::visit(Unop &unop)
 {
   const SizedType &type = unop.expr.type();
-  if (type.IsIntegerTy()) {
+  if (type.IsIntegerTy() || type.IsBoolTy()) {
     return unop_int(unop);
   } else if (type.IsPtrTy() || type.IsCtxAccess()) // allow dereferencing args
   {
@@ -2597,6 +2603,19 @@ ScopedExpr CodegenLLVM::visit(Cast &cast)
     auto *v = b_.CreateAllocaBPF(scoped_expr.value()->getType());
     b_.CreateStore(scoped_expr.value(), v);
     return ScopedExpr(v, [this, v] { b_.CreateLifetimeEnd(v); });
+  } else if (cast.cast_type.IsBoolTy()) {
+    if (cast.expr.type().IsStringTy()) {
+      auto *first_char = b_.CreateGEP(b_.getInt8Ty(),
+                                      scoped_expr.value(),
+                                      { b_.getInt32(0) });
+      Value *cond = b_.CreateICmpNE(b_.CreateLoad(b_.getInt8Ty(), first_char),
+                                    b_.getInt8(0),
+                                    "bool_cast");
+      return ScopedExpr(cond);
+    }
+    Value *zero_value = Constant::getNullValue(scoped_expr.value()->getType());
+    Value *cond = b_.CreateICmpNE(scoped_expr.value(), zero_value, "bool_cast");
+    return ScopedExpr(cond);
   } else {
     // FIXME(amscanne): The existing behavior is to simply pass the existing
     // expression back up when it is neither an integer nor an array.
@@ -3440,6 +3459,9 @@ ScopedExpr CodegenLLVM::getMapKey(Map &map, Expression &key_expr)
                                     b_.getInt64Ty(),
                                     key_type.IsSigned()),
                    key);
+  } else if (map.key_type.IsBoolTy()) {
+    b_.CreateStore(
+        b_.CreateIntCast(scoped_key_expr.value(), b_.getInt8Ty(), false), key);
   } else {
     if (key_expr.type().IsArrayTy() || key_expr.type().IsRecordTy()) {
       // We need to read the entire array/struct and save it
@@ -4347,7 +4369,7 @@ ScopedExpr CodegenLLVM::readDatastructElemFromStack(ScopedExpr &&scoped_src,
 
   Value *src = b_.CreateGEP(data_type, src_data, { b_.getInt32(0), index });
 
-  if (elem_type.IsIntegerTy() || elem_type.IsPtrTy()) {
+  if (elem_type.IsIntegerTy() || elem_type.IsPtrTy() || elem_type.IsBoolTy()) {
     // Load the correct type from src
     return ScopedExpr(
         b_.CreateDatastructElemLoad(elem_type, src, true, elem_type.GetAS()));
--- a/src/ast/passes/fold_literals.cpp
+++ b/src/ast/passes/fold_literals.cpp
@@ -36,6 +36,58 @@ private:
 } // namespace
 
 template <typename T>
+static Expression make_boolean(ASTContext &ast, T left, T right, Binop &op)
+{
+  bool value = true;
+  switch (op.op) {
+    case Operator::EQ:
+      value = left == right;
+      break;
+    case Operator::NE:
+      value = left != right;
+      break;
+    case Operator::LE:
+      value = left <= right;
+      break;
+    case Operator::GE:
+      value = left >= right;
+      break;
+    case Operator::LT:
+      value = left < right;
+      break;
+    case Operator::GT:
+      value = left > right;
+      break;
+    case Operator::LAND:
+      value = left && right;
+      break;
+    case Operator::LOR:
+      value = left || right;
+      break;
+    case Operator::PLUS:
+    case Operator::MINUS:
+    case Operator::MUL:
+    case Operator::DIV:
+    case Operator::MOD:
+    case Operator::BAND:
+    case Operator::BOR:
+    case Operator::BXOR:
+    case Operator::LEFT:
+    case Operator::RIGHT:
+    case Operator::INVALID:
+    case Operator::ASSIGN:
+    case Operator::INCREMENT:
+    case Operator::DECREMENT:
+    case Operator::LNOT:
+    case Operator::BNOT:
+      LOG(BUG) << "binary operator is not a comparison: "
+               << static_cast<int>(op.op);
+  }
+
+  return ast.make_node<Boolean>(value, Location(op.loc));
+}
+
+template <typename T>
 static std::optional<std::variant<uint64_t, int64_t>> eval_binop(T left,
                                                                  T right,
                                                                  Operator op)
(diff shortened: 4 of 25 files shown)


src/parser.yy
```diff
@@ -547,6 +548,7 @@ var_decl_stmt:
 
 primary_expr:
                 UNSIGNED_INT       { $$ = driver.ctx.make_node<ast::Integer>($1, @$); }
+        |       BOOL               { $$ = driver.ctx.make_node<ast::Boolean>($1, @$); }
```
amscanne, line 552: Just mentioning: an alternative here would be parse true/false directly as an untype integer, e.g. `driver.ctx.make_node<ast::Integer>(1 or 0 @$)`, but this approach also seems reasonable (as you building the plumbing for a new class of literal that aligns with `_Bool` and you can get the nice "true"/"false" printing, etc.).

If you are adding a new top-level AST node though, you will need to ensure that it is handled in the [literal folding pass](https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/fold_literals.cpp). We'd need to define whether there are implicit conversions from boolean to integers, e.g. what is "true + 1". I think you've just implicitly defined this as invalid (which is good IMO, but you'll want a bit more exposition in the docs), and you could pretty simply define an implicit integer and string conversion to bool using the standard "truthiness" measures. This could actually simplify some of that constant plumbing a bit as all the logical operators would fold into an `ast::Boolean`.

src/lexer.l
```diff
@@ -33,6 +33,7 @@ int_size (([uU])|([uU]?[lL]?[lL]))
 
 /* Number with underscores in it, e.g. 1_000_000 */
 int      [0-9]([0-9_]*[0-9])?{int_size}?
+bool     true|false
```
amscanne, line 36: Given that bool is currently a built-in type that is an integer, how does this affect casting, etc. between these types? We should have well-defined semantics for this that align with the C99 _Bool type: https://en.cppreference.com/w/c/language/arithmetic_types.html#Boolean_type

src/ast/passes/semantic_analyser.cpp
```diff
@@ -2405,7 +2403,8 @@ void SemanticAnalyser::visit(Ternary &ternary)
     return;
   }
 
-  if (is_final_pass() && cond != Type::integer && cond != Type::pointer) {
+  if (is_final_pass() && cond != Type::integer && cond != Type::pointer &&
```
amscanne, line 2406: Thinking out loud: we could a separate visitor ahead of semantic analyzer, and it can inject a cast to boolean for all conditions. Then we should add a second pass that runs after semantic analyzer that strips out all redundant casts. This can simplify some of these conditions in the semantic analyzer and codegen, and allow us to be more strict while still preserving the user behavior.

src/types.h
```diff
@@ -350,7 +351,7 @@ class SizedType {
     return element_type_.get();
   }
 
-  bool IsBoolTy() const
+  bool IsBitIntTy() const
```
amscanne: Can you actually exist? It seems like it is only used for a check in codegen, but should now be impossible to construct? An array of bools is perfectly fine that way you have defined them.

If it can't then prefer just to delete this pass. AFIAK, non-byte sized integers can only exist as part of a packed struct (because they are not addressable on their own).

amscanne:
Very nice!

Some minor comments -- I know it's in draft, I'm just interested ;)

src/parser.yy
jordalgo: > and you can get the nice "true"/"false" printing

Yeah, that's why I did it this way. I want actual boolean types to be printed as true/false, which is supported in JSON output. If people don't like it they can easily cast their bool to one of our ints.

> you'll want a bit more exposition in the docs

Ack.

> you could pretty simply define an implicit integer and string conversion to bool

~I'm not sure about this TBH - at least as a first pass for this new value. I'd rather keep it simple and only allow casting from string, int, and ptr to bool and from bool to int (as I have now). I'll take a look at what I might need in the literal folding pass though.~ Never mind, I think I misunderstood what you were saying (jet lag is rough).

src/lexer.l
jordalgo: I semi answered above but I think we have an easier time of this because we don't support floats. At the moment for an int anything that evaluates to 0 is false and everything else is true. For a string, we check if the first character in the char array is a null terminator and if so that's false, everything else is true. And for a ptr, `null` is false, everything else is true. I can document this. Anything you think we should add/remove?

src/ast/passes/semantic_analyser.cpp
jordalgo: Interesting... let me poke at that.

src/types.h
jordalgo: Ah good call. Let me look into this.

jordalgo:
@amscanne This PR is already getting a little large. I'm going to do the conditional optimization in a follow-up PR. I'm also going to update the data type section of the adoc as well in a follow-up PR/Commit.

src/types.h
jordalgo: Done.

src/parser.yy
jordalgo: Updated the fold_literals pass to return ast:boolean instead of integer/negative integer.

src/ast/passes/fold_literals.cpp
```diff
@@ -35,6 +35,42 @@ class LiteralFolder : public Visitor<LiteralFolder, std::optional<Expression>> {
...
+    case Operator::GT:
+      value = left > right;
+      break;
+    case Operator::LAND:
+      value = left && right;
+      break;
+    case Operator::LOR:
+      value = left || right;
+      break;
+    default:
```
amscanne: Do you need this? I thought we had compile errors for non-exhaustive enum selects? This is actually preventing those from working.

(You can keep the BUG as a failsafe, but maybe return directly and put it outside the switch? The presence of the `default` here means that you won't get a warning if you do add a new operator)

src/ast/passes/fold_literals.cpp
```diff
@@ -157,10 +181,17 @@ static std::optional<std::variant<uint64_t, int64_t>> eval_binop(T left,
...
+    case Operator::LE:
+    case Operator::GE:
+    case Operator::LT:
+    case Operator::GT:
     case Operator::LAND:
-      return static_cast<uint64_t>(left && right);
     case Operator::LOR:
-      return static_cast<uint64_t>(left || right);
+      LOG(BUG) << "Unexpected comparison binary operator: "
+               << static_cast<int>(op);
```
amscanne: Maybe leave a comment that says these are captured by the `is_comparison_op` check at the top-level?

Is there separate `LOG(BUG)` needed here versus just falling through to the one below?

src/ast/passes/fold_literals.cpp
```diff
@@ -270,28 +301,28 @@ std::optional<Expression> LiteralFolder::visit(Binop &op)
```
amscanne, line 326: You are missing the actual handling of `ast::Boolean` in terms of folding. If you have `true && false` then it won't be folded at all AFAICT.

This needs to handle the case of `Boolean [op] anything` (or vice-versa) with the implicit "truthiness" rules that you have defined elsewhere. Then all of this logic can be pushed into the `Cast` operator and handled as an `Unop`.

src/ast/passes/semantic_analyser.cpp
amscanne: The semantic analyzer itself could choose to inject such a cast if it detects e.g. a non-matching binop. Then we can simplify all the logic around binops to remove anything with casting, and focus exclusively on that logic in the cast operator? Just thinking out loud. (Leaving that cast there also simplifies codegen, since it doesn't need to replicate the casting logic. I believe that today we just hope that LLVM is doing the same thing as us there are implicit conversions.)

src/lexer.l
amscanne: Yeah, that all sounds great.

src/ast/passes/semantic_analyser.cpp
jordalgo: That sounds good. Going to stack that in a PR on-top.

src/ast/passes/fold_literals.cpp
jordalgo: Added handling boolean literals in the folding pass.

src/ast/ast.cpp
```diff
@@ -116,6 +116,40 @@ std::string opstr(const Unop &unop)
...
+    case Operator::INVALID:
+    case Operator::ASSIGN:
+    case Operator::INCREMENT:
+    case Operator::DECREMENT:
+    case Operator::LNOT:
+    case Operator::BNOT:
+      return false;
+  }
+
+  return false; // unreached
```
amscanne, line 150: You could do a `LOG(BUG)` and `__builtin_unreachable()` here if you prefer, just don't add a `default:` clause to the switch :)

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2597,6 +2604,22 @@ ScopedExpr CodegenLLVM::visit(Cast &cast)
...
+      if (cast.expr.type().GetSize() == 0) {
+        return ScopedExpr(b_.getInt8(0));
+      }
+      auto *empty_str = llvm::dyn_cast<GlobalVariable>(
+          module_->getOrInsertGlobal(EMPTY_STR,
+                                     ArrayType::get(b_.getInt8Ty(), 1)));
+      empty_str->setInitializer(
+          ConstantDataArray::getString(module_->getContext(), "\0"));
+      return ScopedExpr(
+          b_.CreateStrncmp(scoped_expr.value(), empty_str, 1, false));
```
amscanne: Why not just check if the pointer is not null and the loaded value is not zero? The need to introduce a global and call to strcmp is a bit strange. (I know that you're trying to avoid some codegen, but I would still err towards doing this in the expected way)

src/ast/passes/fold_literals.cpp
```diff
@@ -228,6 +271,10 @@ std::optional<Expression> LiteralFolder::visit(Unop &op)
       return ast_.make_node<Integer>(static_cast<uint64_t>(value) + 1,
                                      Location(op.loc));
     }
+  } else if (auto *boolean = op.expr.as<Boolean>()) {
+    if (op.op == Operator::LNOT) {
```
amscanne, line 277: In C99, there is implicit conversion from booleans to integers, e.g. `_Bool foo = false; printf("%d", ~foo);` actually prints "-1" :)

I think it's fine to not support any kind of bitwise operation on the `bool` you're building in here, but probably worth explaining this in a comment.

src/ast/passes/fold_literals.cpp
```diff
@@ -302,6 +349,13 @@ std::optional<Expression> LiteralFolder::visit(Binop &op)
     }
   }
 
+  // Handle boolean cases.
+  auto *boolean_left = op.left.as<Boolean>();
+  auto *boolean_right = op.right.as<Boolean>();
+  if (boolean_left && boolean_right) {
+    return make_boolean(ast_, boolean_left->value, boolean_right->value, op);
+  }
```
amscanne, line 388: You will need to also handle the "truthyiness" values that you've spec'ed out. E.g. `true && "foo"` will evaluate to true at runtime, and is a perfectly foldable expression. You will probably need logic like the above that checks and folds the other value into a boolean if it can be folded, probably that the operation is legal.

tests/fold_literals.cpp
```diff
@@ -352,19 +362,27 @@ TEST(fold_literals, binary)
...
+
+  test("1 || 1", "bool: true");
+  test("0 || 1", "bool: true");
+  test("0 || 0", "bool: false");
+  test("-1 || 0", "bool: true");
+  test("1 || -1", "bool: true");
+  test("true || true", "bool: true");
+  test("true || false", "bool: true");
+  test("false || false", "bool: false");
+  test("\"foo\" || 1", "||");    // Left as is
```
amscanne: This is a bug, this should be folded. If it's a bunch of constants and can be evaluated at runtime, then it should be folded.

src/ast/ast.cpp
jordalgo: I think for this function it's fine as is because we have the compiler warnings and if we add a new comparison op, we'll probably add tests for it.

src/ast/passes/codegen_llvm.cpp
jordalgo: Ah yeah, I actually meant to fix this up to do that and forgot.

src/ast/passes/fold_literals.cpp
jordalgo: Yeah, we can revisit if it's requested but for now I'll add a comment.

src/ast/passes/fold_literals.cpp
jordalgo: Ack. Will fix.

src/ast/ast.cpp
```diff
@@ -116,6 +116,40 @@ std::string opstr(const Unop &unop)
   return {}; // unreached
 }
 
+bool is_comparsion_op(Operator op)
```
ajor:
```suggestion
bool is_comparison_op(Operator op)
```
:D

src/parser.yy
```diff
@@ -333,6 +334,7 @@ config_assign_stmt:
                 IDENT ASSIGN UNSIGNED_INT { $$ = driver.ctx.make_node<ast::AssignConfigVarStatement>($1, $3, @$); }
         |       IDENT ASSIGN IDENT        { $$ = driver.ctx.make_node<ast::AssignConfigVarStatement>($1, $3, @$); }
         |       IDENT ASSIGN STRING       { $$ = driver.ctx.make_node<ast::AssignConfigVarStatement>($1, $3, @$); }
+        |       IDENT ASSIGN BOOL       { $$ = driver.ctx.make_node<ast::AssignConfigVarStatement>($1, $3, @$); }
```
ajor: nit: alignment
```suggestion
        |       IDENT ASSIGN BOOL         { $$ = driver.ctx.make_node<ast::AssignConfigVarStatement>($1, $3, @$); }
```

src/types.cpp
```diff
@@ -245,7 +247,7 @@ SizedType CreateInteger(size_t bits, bool is_signed)
 
 SizedType CreateBool()
 {
-  return CreateInteger(1, false);
+  return { Type::boolean, 8 };
```
ajor: Isn't this creating an 8-byte, not 8-bit, type?

tests/runtime/bool
```diff
@@ -0,0 +1,49 @@
...
+EXPECT (true, false)
+TIMEOUT 1
+
+NAME cast string to bool
+PROG BEGIN{ print(((bool)"hello", (bool)"")); exit(); }
+EXPECT (true, false)
+TIMEOUT 1
+
+NAME cast ptr to bool
+PROG BEGIN{ $a = (int64*)0; print(((bool)percpu_kaddr("process_counts", 0), (bool)$a)); exit(); }
```
ajor: nit: is the `percpu_kaddr` necessary here? It'd just be nice to not have a dependency on the kernel symbol `process_counts` if we don't need it. Would something like `$a = (int64*)0; $b = (int64*)1; print((bool)$a, (bool)$b))` be enough?

Also it looks like `percpu_kaddr` returns a `uint64`, not a pointer type after all

tests/semantic_analyser.cpp
```diff
@@ -3039,7 +3064,6 @@ TEST(semantic_analyser, intarray_cast_types)
   test("kretprobe:f { @ = (int8[8])retval }");
 
   test("kprobe:f { @ = (int8[4])1 }", 1);
-  test("kprobe:f { @ = (bool[64])1 }", 1);
```
ajor, line 3042: Are bool arrays meant to work now? I tried running this script on your branch and it fails with:
```
bpftrace: /home/ajor/src/bpftrace/src/types.h:333: const string& bpftrace::SizedType::GetName() const: Assertion `IsRecordTy() || IsEnumTy()' failed.
```

man/adoc/bpftrace.adoc
```diff
@@ -3813,17 +3816,17 @@ If there are many processes running, it will consume a lot of a memory.
...
 
 C++ symbol demangling in userspace stack traces is enabled by default.
 
-This feature can be turned off by setting the value of this environment variable to `0`.
+This feature can be turned off by setting the value of this variable to `false`.
 
 ==== lazy_symbolication
 
-Default: 0
+Default: false
```
ajor, line 3827: Just checking: integer values will still parse correctly for config and env vars, right?

Is it worth updating main.cpp to show these as booleans as well?

e.g.:
https://github.com/bpftrace/bpftrace/blob/f3e3455f876307d5ea5516af802c21d32062c552/src/main.cpp#L148

ajor (commented):
Nice, I've been looking forward to booleans for years!

src/ast/ast.cpp
jordalgo: This is the American spelling 😉

man/adoc/bpftrace.adoc
jordalgo: Yes, integer values still work.

> Is it worth updating main.cpp to show these as booleans as well?

Good call. Will update.

src/types.cpp
jordalgo: Whoops. Good catch.

tests/runtime/bool
jordalgo: Yup, what you suggested is better. Will change.

man/adoc/bpftrace.adoc
jordalgo: Actually, these are env variables that only take 1 or 0 (though we should probably fix that).

tests/semantic_analyser.cpp
jordalgo: Oops, fixed now and added a few tests.

ajor:
Just one more issue with bool arrays - this program triggers a verifier error:
```
BEGIN { $a = (bool[8])1; print($a)  }
```

UPDATE: Ah, it seems like this is an issue with arrays in general, not specific to bool arrays. Same error:
```
BEGIN { $a = (uint32[2])1; print($a)  }
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2597,6 +2603,22 @@ ScopedExpr CodegenLLVM::visit(Cast &cast)
     auto *v = b_.CreateAllocaBPF(scoped_expr.value()->getType());
     b_.CreateStore(scoped_expr.value(), v);
     return ScopedExpr(v, [this, v] { b_.CreateLifetimeEnd(v); });
+  } else if (cast.cast_type.IsBoolTy()) {
+    if (cast.expr.type().IsStringTy()) {
+      if (cast.expr.type().GetSize() == 0) {
```
ajor: Is this something that can happen?

jordalgo:
> Ah, it seems like this is an issue with arrays in general, not specific to bool arrays.

Issue filed. I can take a look if no one else wants to investigate:
- https://github.com/bpftrace/bpftrace/issues/4295

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2690,7 +2712,8 @@ void CodegenLLVM::createTupleCopy(const SizedType &expr_type,
                               { b_.getInt32(0), b_.getInt32(i) });
     if (t_type.IsTupleTy() && !t_type.IsSameSizeRecursive(var_type)) {
       createTupleCopy(t_type, var_type.GetField(i).type, dst, offset_val);
-    } else if (t_type.IsIntTy() && t_type.GetSize() < 8) {
+    } else if ((t_type.IsIntTy() || t_type.IsBoolTy()) &&
+               t_type.GetSize() < 8) {
```
ajor: I think this check is wrong - it doesn't match the corresponding check on line 3446, which stores bools as 8-bits, not 64-bits.

Here's a program that (I think) fails verification because of this:
```
@mmm[true, true, 123] = 1111; @mmm[true, true, (int32)456] = 2222;
```

src/ast/passes/codegen_llvm.cpp
jordalgo: Probably not. Just being overly paranoid I guess. I can remove.

src/ast/passes/codegen_llvm.cpp
jordalgo: Ack. Good catch. Fixed now.

jordalgo:
@ajor @amscanne - How is this looking now? Baked enough to merge?

PR #4281, opened 2025-06-21
Update issue metrics workflow

jordalgo:
Add separate metrics for PRs and issues.
Also hide some fields that are always empty.

##### Checklist

### DIFF
--- a/.github/workflows/issue-metrics.yml
+++ b/.github/workflows/issue-metrics.yml
@@ -1,4 +1,4 @@
-name: Generate Issue Metrics
+name: Generate Issue and PR Metrics
 on:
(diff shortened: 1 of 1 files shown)


jordalgo:
> Did you want to put this into a different workflow? Just makes it easier to find separately as a distinct thing.

Maybe? I'm just tinkering right now. We can always improve as we go.

> IMHO, I don't think you need to wait for reviews for this kind of scaffolding and should just go ahead.

I was thinking the same. The only thing I was worried about making ugly was the commit history. Not sure if we need any kind of rules because we, as maintainers, have the ability to by-pass the merge rules.

PR #4285, opened 2025-06-22
Remove is_btftype

jordalgo:
##### Checklist


tests/codegen/llvm/fexit_dereference.ll
```diff
@@ -27,7 +28,10 @@ entry:
   %5 = getelementptr i8, ptr %4, i64 0
   %6 = call ptr @llvm.preserve.static.offset(ptr %5)
   %7 = getelementptr i8, ptr %6, i64 0
-  %8 = load volatile i32, ptr %7, align 4
+  call void @llvm.lifetime.start.p0(i64 -1, ptr %"struct sock_common.skc_daddr")
+  %probe_read_kernel = call i64 inttoptr (i64 113 to ptr)(ptr %"struct sock_common.skc_daddr", i32 4, ptr %7)
```
amscanne, line 32: Did this previously work because fentry/fexit have BTF annotations on the input context and the pointer is non-nullable?

jordalgo: I think it's far more simple. `is_btftype`, which was added in semantic analyser, was determining if we need a probe read call instead of just a de-ref for field access into fentry/fexit args. I'm going to look into fixing this - at the same time this additional probe read call still works, it's just a bit more overhead.

amscanne: The question was dumber than that :)

The original code for this is:
```
fentry:tcp_sendmsg { @[args->sk->__sk_common.skc_daddr] = 1; }
```

We were generating raw dereferences for `args->sk` (that's okay, it's the context), and `sk->__sk_common`. What is the mechanism that allows the verifier to accept this, i.e. guaranteeing that `sk` itself is not null in this path? I know there's some magic around this, I'd just like to understand it since our internal types should eventually map into the same way this magic is allowed/represented.

jordalgo: According to Alexei, all fentry/fexit args are untrusted but you don't need a null check because a read from NULL will return zero. So args->sk->__sk_common.skc_daddr won't crash, but results could be misleading.

More details:
BPF sets up a handler to catch memory access errors when load instructions for untrusted pointers are processed. At verification time the "untrusted" bit for a pointer is tracked and is sticky. At jit time, a set of BPF program addresses corresponding to these loads is remembered, and at execution time faults at these addresses would be caught and silenced. The type for the e.g. sk is "pointer to btf id", when verifier follows loads with offsets from such pointers, it tracks the BTF type of the fields, thus allowing the chain of dereferences.

amscanne: Got it, so it's "safe" because it is typed and therefore the verifier can track that, but loads through nulls will silently return nulls.

For now, I'd consider the new code generated here to actually be more correct. Once we have a self-consistent type system, we can peel out the probe reads as an optimization. (I would assume that these get inlined anyways in a way that is very similar to the direct dereference?)

amscanne (approved):

jordalgo:
@viktormalik @ajor - Mind taking a look too?

tests/codegen/llvm/fexit_dereference.ll
jordalgo: > I would assume that these get inlined anyways in a way that is very similar to the direct dereference?

That I'm not sure of - hopefully that's the case 🤞 

Just as some added info, the script below yields the same probe read call on master because we saved off part of this access to a variable:
```
fentry:tcp_sendmsg { $a = args->sk->__sk_common; @[$a.skc_daddr] = 1; }
```

jordalgo:
@viktormalik Done and thanks for sanity checking.

PR #4296, opened 2025-06-25
Remove AddrSpace::bpf

jordalgo:
This enum value wasn't being checked for.

##### Checklist

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -2706,7 +2706,7 @@ void SemanticAnalyser::visit(For &f)
   auto [iter, _] = for_vars_referenced_.try_emplace(&f);
   auto &collector = iter->second;
   for (const Variable &var : collector.nodes()) {
-    ctx_types.push_back(CreatePointer(var.var_type, AddrSpace::bpf));
+    ctx_types.push_back(CreatePointer(var.var_type, AddrSpace::kernel));
     ctx_idents.push_back(var.ident);
   }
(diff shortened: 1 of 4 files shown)


amscanne (approved):
I'm strongly in favor of taking things out that are unused and distilling current functionality as much as possible. But I'll explicitly note that my approval is not judgement on whether such a designation is **useful** or not, which we can figure out in the future :)

ajor:
Why was this removed? Kernel and BPF are separate address spaces and it doesn't make sense to mark pointers to things on the BPF stack as being in the kernel address space.

Prior to introducing `AddrSpace;:bpf` they would have just been marked as `AddrSpace::none`. It would make more sense to go back to that rather than marking as `AddrSpace::kernel`, although I disagree with it being removed anyway

jordalgo:
@ajor If the enum value is not being used in any way, why keep it?

ajor:
Because the `AddrSpace` enum is incomplete without it. `AddrSpace::kernel` isn't correct for these instances, and `AddrSpace::none` just makes it look like we don't know which address space the memory belongs to.

I thought your intention was to move away from `is_internal` and start using `AddrSpace::bpf` instead anyway?

jordalgo:
> I thought your intention was to move away from is_internal and start using AddrSpace::bpf instead anyway?

I'm just trying to remove some cruft and this being unused falls into that category. Let me look into using `AddrSpace::bpf` as a replacement for `is_internal` (formerly I was just trying to remove the latter if possible) as I would prefer that as well (if it means roughly the same thing)

PR #4297, opened 2025-06-26
async: refactor handles into a stateful class

amscanne:
async: refactor handles into a stateful class

In the future, this will allow that class to load information about
external types, and dispatch appropriately.


tests/bpftrace.cpp
```diff
@@ -1441,9 +1441,8 @@ TEST(bpftrace, print_basic_map)
     std::string expected_output;
   };
 
-  auto keys = std::vector<uint64_t>{ 1, 3, 5, 7, 9 };
-  auto values = std::vector<uint64_t>{ 5, 10, 4, 11, 7 };
-  auto returned_kvs = generate_kv_pairs(keys, values);
+  const auto keys = std::vector<uint64_t>{ 1, 3, 5, 7, 9 };
+  const auto values = std::vector<uint64_t>{ 5, 10, 4, 11, 7 };
```
viktormalik, line 1445: Nit: changes in this file look unrelated to the rest. Maybe move them to a separate commit?

jordalgo (approved):
Aside from @viktormalik's suggestion, it LGTM.

tests/bpftrace.cpp
amscanne: Sorry, accidental flatten. Thanks for catching this. These will appear in another patch.

PR #4298, opened 2025-06-26
util: split TypeName into separate file

amscanne:
Stacked PRs:
 * #4302
 * #4301
 * #4300
 * #4299
 * __->__#4298
 * #4304

--- --- ---

### util: split TypeName into separate file

In order to use this trick for other cases (registering classes that
correspond to standard library types), make this a separate util file.

### DIFF
--- a/src/util/type_name.h
+++ b/src/util/type_name.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include <algorithm>
+#include <string>
+
+namespace bpftrace::util {
+
+// TypeName is used as a template parameter to work around restrictions on
+// literals passed as template parameters. This is essentially a string
+// literal used to name the type.
+template <size_t N>
+struct TypeName {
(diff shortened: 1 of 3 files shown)


src/util/type_name.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <algorithm>
+#include <string>
+
+namespace bpftrace::util {
+
+// TypeName is used as a template parameter to work around restrictions on
+// literals passes as template parameters. This is essentially a string
```
viktormalik: Typo?
```suggestion
// literals passed as template parameters. This is essentially a string
```

jordalgo (approved):

src/util/type_name.h
amscanne: Fixed. I've updated in my local stack, and it will be refreshed prior to submit.

PR #4301, opened 2025-06-26
output: add output APIs and libraries

amscanne:
Stacked PRs:
 * #4323
 * #4322
 * #4321
 * #4320
 * #4302
 * __->__#4301

--- --- ---

### output: add output APIs and libraries

In order to implement extensible types, it is necessary to decouple the
output from the definition of internal types. This change adds a layer
of abstraction for output: there are high-level primitives for types to
be printed, and high-level operations associated with the current set of
"MessageTypes" for an abstract output.

The core idea is that in the future, an extensible type can be asked for
its representation in terms of the `output::Value` rather than having
the output driver need to understand all of the different internal type
representations. This provides a clear separation of concerns, and
should allow us to build new types, functions, etc. without having to
make changes to all the various output drivers.

This specific change contains only the implementation of those two
libraries: while it should build, nothing will use this yet. The
subsequent change will effectively be a switch over to using these
output drivers instead of the existing one.


amscanne:
@jrife FYI, this stack is a significant refactoring of the way the output modules work. The goal is to work towards non-hard-coded types (e.g. allow extensibility from a standard library). These types will need to map to some set of "primitives" which the output plugins understand. That way, you can add new types w/o needing to implement support in the individual outputs.

However, there are some special types that the outputs understand: for example Histograms (although they don't understand the different between histograms, nor the values, nor the steps, etc.). The `TimeSeries` is also added as one of these special types in this commit. I tried to port your code and preserve the exact output as much as possible, but given that it's a very different structure, there were changes (intentional and likely unintentional, which I'll need to find).

The next change in the stack (#4302) changes the backend to generate the data from existing types and wires up these output implementations. So essentially the output for the `tseries` is split into two parts. (This is why I wanted the change to land, so you didn't have to deal with this large refactor in the middle.) Subsequent changes also affect how these types are extracted and build (moving away from `util::read_data` and towards a bounds-checked `OpaqueValue` class with suitable accessors). I'll add you as these are ready for review in case you want to take a look. I just wanted to let you know as I don't want to break tseries immediately :)

src/output/text.cpp
```diff
@@ -0,0 +1,604 @@
...
+    const auto [unit, scale] = util::duration_str(interval);
+    std::string time_fmt = "hh:mm:ss";
+    size_t time_size = time_fmt.size();
+    switch (unit) {
+      case util::DisplayUnit::ns:
+        time_fmt += ".ns";
+        time_size += 1 + 9;
+        break;
+      case util::DisplayUnit::us:
+        time_fmt += ".μs";
```
jrife: μnicode? Nice. I have a PoC that uses unicode box drawing characters and such for tseries and hist output. Maybe we can explore it a bit more after all the dust settles from the output refactoring work..

Screenshot from 2025-07-01 21-58-49

Screenshot from 2025-07-09 23-05-10

Anyway, just took a cursory glance at this. Will look in more detail at the TimeSeries parts tomorrow.

jrife:
> I tried to port your code and preserve the exact output as much as possible, but given that it's a very different structure, there were changes (intentional and likely unintentional, which I'll need to find).

Thanks for the heads up. Do you have examples of the output before and after? Is this branch runnable if I wanted to play around with it?

src/output/text.cpp
amscanne: > μnicode? Nice. I have a PoC that uses unicode box drawing characters and such for tseries and hist output. Maybe we can explore it a bit more after all the dust settles from the output refactoring work..

The output **should** be the same. I tried to preserve it. The only difference is that there was an existing bug when emitting multiple nested objects under a single key. This has been fixed.

(You can see this encoded in the t-series test. Your test has `is_scalar = true` and yeah the output has two maps emitted with the same name. The `is_scalar` was meaningful, and the keys were missing the key. This was the case for histograms as well.)

The point of the series is to simply disentangle the output from the internal types, and all the map details. The code ends up being substantially different (because it's basically a rewrite to add a module boundary there), but I tried to not overtly change the output. The net result of the full series is a small reduction in code size, but IMO a lot more structure.

> Screenshot from 2025-07-01 21-58-49
> 
> Screenshot from 2025-07-09 23-05-10

Oh my gawd this looks so good. Once this has landed I'd be strongly supportive of switching the `TextOutput` to using this kind of unicode output. It will become a very simple, isolated change. And it's `CURRENT_YEAR`, unicode works everywhere :)

> Anyway, just took a cursory glance at this. Will look in more detail at the TimeSeries parts tomorrow.

src/output/output.h
```diff
@@ -0,0 +1,162 @@
...
+  // Specialized messages during execution.
+  virtual void printf(const std::string& str) = 0;
+  virtual void time(const std::string& time) = 0;
+  virtual void cat(const std::string& cat) = 0;
+  virtual void join(const std::string& join) = 0;
+  virtual void syscall(const std::string& syscall) = 0;
+  virtual void lost_events(uint64_t lost) = 0;
+  virtual void attached_probes(uint64_t num_probes) = 0;
+  virtual void helper_error(int retcode, const HelperErrorInfo& info) = 0;
+  virtual void end() = 0;
```
jrife, line 161: Would we need another member like `benchmark` for printing benchmark results for https://github.com/bpftrace/bpftrace/pull/4343? I can wait for this PR to land and just rebase the `BENCH` later to add this.

src/output/text.cpp
```diff
@@ -0,0 +1,604 @@
...
+    // Figure out the interval.
+    const auto interval = (ts.values.back().first - ts.values.front().first) /
+                          (ts.values.size() > 1 ? ts.values.size() - 1 : 1);
+    const auto [unit, scale] = util::duration_str(interval);
+    std::string time_fmt = "hh:mm:ss";
+    size_t time_size = time_fmt.size();
+    switch (unit) {
+      case util::DisplayUnit::ns:
+        time_fmt += ".ns";
+        time_size += 1 + 9;
```
jrife: nit: not sure I love the magic numbers here. Should there be some named consts? Or maybe a method like `max_length()` for `DisplayUnit`?

src/output/text.cpp
```diff
@@ -0,0 +1,604 @@
...
+      }
+      if (v > max_value) {
+        max_value = v;
+      }
+    }
+    if (min_value == max_value) {
+      // Generally prefer if we can add some buffer on both sides of the
+      // points in the graph, but don't overflow if our min or max is at the
+      // upper bound of the range.
+      try_add<uint64_t>(max_value);
```
jrife: nit: Personally I would prefer if the type check was open-coded at this level instead of being buried in `try_add`, as I think it makes the intent slightly clearer.

```cpp
  if (std::holds_alternative<uint64_t>(min_value)) {
    try_sub<uint64_t>(min_value);
  } else {
    try_sub<int64_t>(min_value);
  }

  if (std::holds_alternative<uint64_t>(max_value)) {
    try_add<uint64_t>(max_value);
  } else {
    try_add<int64_t>(max_value);
  }
```

or

```cpp
  if (std::holds_alternative<uint64_t>(min_value)) {
    try_add<uint64_t>(max_value);
    try_sub<uint64_t>(min_value);
  } else {
    try_add<int64_t>(max_value);
    try_sub<int64_t>(min_value);
  }
```

src/output/text.cpp
```diff
@@ -0,0 +1,604 @@
...
+      try_add<int64_t>(max_value);
+      try_sub<uint64_t>(min_value);
+      try_sub<int64_t>(min_value);
+    }
+
+    // Figure out the interval.
+    const auto interval = (ts.values.back().first - ts.values.front().first) /
```
jrife, line 377: I don't see the code that populates `ts.values`, but one thing to watch out for is that the time series is sparsely populated, so deriving the interval in this way may not be accurate. That's why I carry it through as part of `map_info` in the current implementation.

amscanne: It includes a value for each interval, which may be a `std::monostate` Primitive. This is handled below.

src/output/text.cpp
amscanne: Sure, I will do this on the next rebase & iteration.

src/output/text.cpp
amscanne: Yeah, I'm up for anything. I was just getting it to look exactly the same.

src/output/output.h
amscanne: Let's see how this goes. It will probably take a while to land this (but it's just for the two). Then again, it may take a little while to work through the `BENCH` decisions, for different reasons.

src/output/text.cpp
jrife: Ah, I see now. In that case, carry on.

src/output/json.cpp
```diff
@@ -0,0 +1,461 @@
+#include <iomanip>
+#include <string>
+
+#include "output/json.h"
```
jrife, line 4: Translating data structures to JSON output seems like something there should be a standard library for, but maybe I have been coddled by Go's JSON handling features. Is constructing JSON strings by hand the easiest way here?

amscanne: LLVM has JSON types and encoding/decoding, but I don't think it would be significantly less complex than just having these bits here. At least it's structured, so objects will be spit out in the same way. (Versus each map type have its own open-coded JSON generation.)

That said, my goal is to decouple the output from the internal types & map details and not change them in the process. Once this refactor lands, we can just go hog wild on whatever outputs we want. I'd like to add an `--output=none` which just the empty implementation which can be the default for `--bench`. But I think that changing *just* the output becomes way simpler and self-contained either way.

src/output/text.cpp
amscanne: I've done a different iteration without the templates that personally I think is easiest to understand. PTAL.

src/output/text.cpp
amscanne: I've updated the base to do something with the enum units. PTAL. Maybe it's too clever by half.

jrife:
```cpp
enum class DisplayUnit {
  s = 0,
  ms = 3,
  us = 6,
  ns = 9,
};
```

This is clever. Part of me thinks the `static_cast<size_t>` is a little unclear at first glance and a named helper function would be clearer, but no strong feeling.

src/output/text.cpp
jrife: Looks nicer. Thanks.

src/output/json.cpp
jrife: Makes sense, thanks.

src/output/text.cpp
amscanne: LOL. Yeah but is it *too* clever?

I just didn't want a naked function. Too bad no methods for enums. I tried to make up for the cleverness with documentation ;)

jrife: > LOL. Yeah but is it too clever?

On one hand, I feel like this is something I would come up with and fall in love with because I like being clever. On the other hand, it's one of those things I'd fully expect to be met with a code review comment like "Just use a function and a switch statement". I'd say "let your conscience decide" :).

src/output/output.h
```diff
@@ -0,0 +1,162 @@
...
+  virtual void value(const Value& value) = 0;
+
+  // Specialized messages during execution.
+  virtual void printf(const std::string& str) = 0;
+  virtual void time(const std::string& time) = 0;
+  virtual void cat(const std::string& cat) = 0;
+  virtual void join(const std::string& join) = 0;
+  virtual void syscall(const std::string& syscall) = 0;
+  virtual void lost_events(uint64_t lost) = 0;
+  virtual void attached_probes(uint64_t num_probes) = 0;
```
jordalgo, line 157: Can't these two be made more general? Feels like they could use some abstraction being they are just an int and an associated string message. I guess I'm just looking to reduce the total number of specialized cases we have in this list.

jordalgo (commented):
This feels like the right commit to add tests for the new json.cpp and text.cpp classes as there is a fair bit of biz logic in each. I know it's a large stack but adding tests to new classes strikes me as being part of this "Boy Scout" rule (I guess implying that all Girl Scouts do is make cookies - which seems wrong).

src/output/output.h
```diff
@@ -0,0 +1,162 @@
...
+
+  // Print an arbitrary value.
+  virtual void value(const Value& value) = 0;
+
+  // Specialized messages during execution.
+  virtual void printf(const std::string& str) = 0;
+  virtual void time(const std::string& time) = 0;
+  virtual void cat(const std::string& cat) = 0;
+  virtual void join(const std::string& join) = 0;
+  virtual void syscall(const std::string& syscall) = 0;
```
jordalgo, line 155: Is it possible to smush these all together into a method like `special` that takes an enum to account for these one-offs? Might be cleaner than these methods.

src/output/text.cpp
```diff
@@ -0,0 +1,594 @@
...
+  }
+  // bpftrace sets the return code to 0 for map_lookup_elem failures
+  // which is why we're not also checking the retcode
+  else if (info.func_id == libbpf::BPF_FUNC_map_lookup_elem) {
+    msg = "Can't lookup map element because it does not exist.";
+  } else {
+    msg = strerror(-retcode);
+  }
+
+  LOG(WARNING,
```
jordalgo, line 583: Umm, why do we only log warnings for these errors when producing text output?

src/output/output.h
amscanne: This is codifying the existing API exactly, I'm trying not to make any changes to the API and the underlying implementations.

src/output/output.h
amscanne: I absolutely agree and hate the structure of this API. However, there's enough changing at once, so I'm not changing the "message" protocol here, and just doing a 1:1 mapping from the `MessageType` to virtual methods on the output side to preserve the exact behavior.

src/output/text.cpp
amscanne: I don't understand the question. The JSON output has structured output for helper errors (in exactly the same format that they exist currently).

If you're asking why json output does not *also* emit text output to the console via a separate stream, I would ask: do you expect it to? Isn't JSON output implying that you are consuming via a machine? Is the expectation that you want *both* textual output and JSON output whenever it is an error?

This is why global singletons for logging are bad. Now there's a bunch of confusion regarding the role of the output classes. Right here, because `out_` is being passed, I am merely using the `LOG` for the formatting. The output types themselves do not have extra side-effects (logging to some hidden global). There is no way to preserve that property for the JSON type and using `LOG(...)` because you'd just be stuffing some text into the JSON stream.

src/output/output.h
jordalgo: Fair enough but perhaps let's add a TODO or issue to clean this up after the fact.

src/output/text.cpp
jordalgo: > If you're asking why json output does not also emit text output to the console via a separate stream, I would ask: do you expect it to? Isn't JSON output implying that you are consuming via a machine? Is the expectation that you want both textual output and JSON output whenever it is an error?

Yeah this is what I was asking (sorry for the ambiguity). I always forget that JSON output is intended to be for machine consumption. That being said every `LOG(WARNING)` in the codebase obviously doesn't show up in the json output. It shows up on the command line as text. These warnings seem to be unique in this way. This is not for this PR to deal with but maybe we should think more about if we even want these warnings going to the json output like they are now. (cc @viktormalik)

> This is why global singletons for logging are bad. Now there's a bunch of confusion regarding the role of the output classes. Right here, because out_ is being passed, I am merely using the LOG for the formatting. 

Indeed. I was confused 😕 It's also interesting when folks specify the output file command line option. Most things go to the file but `LOG(WARNING)` does not. How do other tools handle stderr messages when there is an option for a file output?

amscanne: I don't know if there's a good general answer to this. Right now I'm just trying to structure those the output drivers in a more self-contained way.

src/output/output.h
amscanne: I've added this.

src/output/output.h
amscanne: I'm honestly not sure in which way I'd want to clean this up. For now, for better or worse, we have codified this series of messages in the JSON output. The methods here correspond 1:1 to the `type` emitted. I don't really want to break that `1:1` correspondence and have one special case for 9/10th of the messages. Either a set of virtual methods or a `enum` are fine, having a hybrid of **both** seems a bit silly. I just elected for the first one, since I think it is clearer this way (and yes, the kludginess of having a dedicated "printf" message type is apparent, but this is also by design).

What I *would* like to do is actually eliminate most of these messages, but I think that's bigger than a `TODO`.

jordalgo: > that's bigger than a TODO

That's what issues are for 😉 but fine, I'm not blocking on this. Just another yak that needs shaving.

jordalgo (approved):
Nice clean up and new API! Feel free to merge before my modulo fix PR so you don't have to rebase again.

PR #4302, opened 2025-06-26
output: change to using generic output drivers

amscanne:
Stacked PRs:
 * #4323
 * #4322
 * #4321
 * #4320
 * __->__#4302

--- --- ---

### output: change to using generic output drivers

The previous change introduced the `output` library, which included a
generic type schema and drivers that were capable of displaying that
schema is a way that aligns with the current output.

This change removes the current output classes & output generation, and
replaces it with a version that constructs objects from the output
library and then calls the relevant methods. This forces a separation of
concerns and modularity for the output drivers.

Note that this change is rather large, and difficult to break up. There
are a few pieces that are entangled:

* The existing `output.cpp` file, which understands `SizedType` is
  converted into a `types_format.cpp` file, which also understands
  `SizedType` and how to unpack data into an "outputable"
  representation. Some of the `print_maps` methods are collected from
  `BPFtrace` in order to separate concerns more effectively.

* The paths for creating outputs, which were duplicated in the AOT and
  regular cases are aggregated into `run_bpftrace`.

* The async actions are updated to call the `format` method which
  constructs the object objects. These are then passed to the associated
  virtual methods, which generate the required messages.

* A very small number of tests are updated, only to make the output
  consistent where capturing the inconsistencies would increase
  complexities. In general, great effort was taken to support the
  existing schema and format wherever possible.


src/bpftrace.cpp
```diff
@@ -1279,24 +1115,46 @@ std::string BPFtrace::resolve_uid(uint64_t addr) const
...
+    const std::string &raw_fmt,
+    bool utc)
 {
   static const auto nsec_regex = std::regex("%k");
   static const auto usec_regex = std::regex("%f");
   static const auto msec_regex = std::regex("%l");
-  uint64_t ns = 0;
-  time_t time = time_since_epoch(mode, nsecs, &ns);
+  time_t time = std::chrono::system_clock::to_time_t(time_point);
+  auto ns = (time_point - std::chrono::floor<std::chrono::seconds>(time_point))
```
jordalgo, line 1203: JC, why flooring instead of rounding?

src/bpftrace.h
```diff
@@ -141,13 +141,16 @@ class BPFtrace : public ast::State<"bpftrace"> {
...
   std::string resolve_inet(int af, const uint8_t *inet) const;
   std::string resolve_uid(uint64_t addr) const;
-  std::string resolve_timestamp(uint32_t mode,
-                                uint32_t strftime_id,
-                                uint64_t nsecs);
-  std::string resolve_timestamp(uint32_t mode,
-                                uint64_t nsecs,
-                                const std::string &raw_fmt,
-                                bool utc);
+  std::chrono::time_point<std::chrono::system_clock> resolve_timestamp(
```
jordalgo, line 144: nit: Should these time stamp methods go into a utility class?

src/helpers_format.cpp
```diff
@@ -0,0 +1,43 @@
+#include <bpf/libbpf.h>
```
jordalgo: nit: the filename is too generic, it should be `bpf_helpers_format` - also maybe be part of `util` folder?

tests/runtime/signed_ints
```diff
@@ -1,6 +1,6 @@
 NAME stats with negative values
 PROG BEGIN { @=stats(-10); @=stats(-5); @=stats(5); exit() }
-EXPECT @: count 3, average -3, total -10
+EXPECT @: { .count = 3, .average = -3, .total = -10 }
```
jordalgo, line 3: ha - yeah, this is better.

tests/runtime/json-output
```diff
@@ -183,7 +183,7 @@ EXPECT {"type": "printf", "data": "1 ONE 2 TWO 99999 OTHER\n"}
 # or anon structs, then we will need to change/delete this test
 NAME none type
 RUN {{BPFTRACE}} -q -f json -e 'union N { int i; float f; }; struct Foo { int m; union N n; }; uprobe:./testprogs/struct_with_union:func { print(*((struct Foo *) arg0)); exit(); }'
-EXPECT {"type": "value", "data": { "m": 2, "n": { "i": 5, "f": "" } }}
+EXPECT {"type": "value", "data": {"m": 2, "n": {"i": 5, "f": null}}}
```
jordalgo, line 186: I know this is tiny but maybe this is a breaking change?

src/util/tseries.cpp
```diff
@@ -2,10 +2,92 @@
...
+  if (agg == TSeriesAggFunc::avg) {
+    latest.first = static_cast<T>(latest.first / latest.second);
+  }
+
+  latest.second = latest_epoch;
+
+  return latest;
+}
+
 std::pair<std::vector<uint8_t>, uint64_t> reduce_tseries_value(
```
jordalgo, line 91: Do we still need this wrapping function or can we just do the check at the call site and avoid memcpy?

jordalgo (commented):
Overall looks good. Few minor nits and questions.

src/bpftrace.cpp
amscanne: This is taking the floor and taking the nanoseconds separately, in order to print in this specific way (preserving the existing behavior).

src/bpftrace.h
amscanne: They certainly don't belong in `BPFtrace` IMO :) However, I'm aiming to minimize unnecessary moves/refactors since this change is rather massive. I'd like to reserve that for a future cleanup.

src/helpers_format.cpp
amscanne: This file appears to be accidentally added! The individual outputs have this internally. I will delete it.

tests/runtime/json-output
amscanne: You mean the `null`? Or that map spacing is now consistent?

I don't think the second one is an issue, provided that the user is actually deserializing JSON.

The second is more subtle: most JSON serialization libraries should handle this appropriately and will simply set to the zero value of the intended type. (E.g. if it's a string, it's the empty string, a number, then it's zero, etc.). I can add a changelog entry that notes voids/unserializable values are now serialized as null just in case?

src/util/tseries.cpp
amscanne: This is rewritten with `OpaqueValue`, so I'd rather not touch in this context.

tests/runtime/json-output
jordalgo: Yeah, I meant the `null`. 

> I can add a changelog entry that notes voids/unserializable values are now serialized as null just in case?

Yeah that's probably fine.

src/util/tseries.cpp
jordalgo: kk

PR #4303, opened 2025-06-26
tests: fix key data for map tests

amscanne:
tests: fix key data for map tests

It is unclear why this is working currently. The key type is an int64,
and the `util::read_data` function *will* read a full 8 bytes from the
starting point of this vector. Since the type of the key vector is
`std::vector<uint8_t>`, we need to have 8 entries for this to work.

We should consider updating the `util` helpers to use the bounds of the
vector, rather than running off the end.


amscanne:
Somehow this got duplicated. I will dedup to #4304.

PR #4300, opened 2025-06-26
stats: handle zero counts

amscanne:
Stacked PRs:
 * #4302
 * #4301
 * __->__#4300

--- --- ---

### stats: handle zero counts

Signed-off-by: Adin Scannell <amscanne@meta.com>

PR #4304, opened 2025-06-26
tests: fix key data for map tests

amscanne:
Stacked PRs:
 * #4302
 * #4301
 * #4300
 * #4299
 * #4298
 * __->__#4304

--- --- ---

### tests: fix key data for map tests

It is unclear why this is working currently. The key type is an int64,
and the `util::read_data` function *will* read a full 8 bytes from the
starting point of this vector. Since the type of the key vector is
`std::vector<uint8_t>`, we need to have 8 entries for this to work.

We should consider updating the `util` helpers to use the bounds of the
vector, rather than running off the end.

Note that this is little-endian specific, but this change is not
adjusting the current semantics. I intend to replace these vectors with
a mechanism that is safer and works independently of the endianness.

### DIFF
--- a/tests/bpftrace.cpp
+++ b/tests/bpftrace.cpp
@@ -1792,8 +1792,8 @@ lhist_map_3:
     auto mock_map = std::make_unique<MockBpfMap>(libbpf::BPF_MAP_TYPE_HASH,
                                                  tc.name);
     HistogramMap values_by_key = {
-      { { 0 }, { 0, 10, 20, 30, 40, 50, 0 } },
-      { { 1 }, { 0, 2, 2, 2, 2, 2, 0 } },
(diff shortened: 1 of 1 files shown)


viktormalik (approved):
Nice catch!

Note: this only works on little-endian but that was the case even before this change. I also expect that we have more places in our tests where we assume little-endian.

amscanne:
> Nice catch!
> 
> Note: this only works on little-endian but that was the case even before this change. I also expect that we have more places in our tests where we assume little-endian.

Yes, I thought about this as well. My conclusion was the same: I didn't worry about the arch-specifcness for now, since there are likely other tests that will fail.

I intend to refactor these factors to not require an array and instead operate on an opaque value that you can construct from any type.

PR #4306, opened 2025-06-26
tests: fix `collect_elements`-related mock calls

amscanne:
Stacked PRs:
 * #4302
 * #4301
 * #4300
 * #4299
 * #4298
 * #4304
 * #4297
 * __->__#4306

--- --- ---

### tests: fix `collect_elements`-related mock calls

Since these now return a `Result`, it should ideally return a value
which is moved. While this works at the moment, it results in errors
using newer versions of clang. This fixes it to properly specify the
type and construct the new value without an implicit copy.

### DIFF
--- a/tests/bpftrace.cpp
+++ b/tests/bpftrace.cpp
@@ -1441,9 +1441,8 @@ TEST(bpftrace, print_basic_map)
     std::string expected_output;
   };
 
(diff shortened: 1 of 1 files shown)


jordalgo (approved):
Clean!

PR #4307, opened 2025-06-26
tidy: run `clang-tidy` against codegen tests

amscanne:
##### Checklist


jordalgo:
I guess clang tidy doesn't run against these files normally because they're not copied into the build directory. TBH, I'm a little unclear why we need to build to have clang-tidy run.

amscanne:
Yeah, I just happened to run it manually. I'm not going to think too hard about it, but may as well fix the errors.

PR #4310, opened 2025-06-27
tools/opensnoop: support openat2(2)

Rtoax:
kernel v5.6 introduces fddb5d430ad9 ("open: introduce openat2(2) syscall").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,8 @@ and this project adheres to
 #### Security
 #### Docs
 #### Tools
+- opensnoop.bt: support openat2 syscall
+  - [#4310](https://github.com/bpftrace/bpftrace/pull/4310)
 - killsnoop.bt: display signal name instead of value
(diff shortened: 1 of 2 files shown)


jordalgo (commented):
@Rtoax Thanks for the addition. Do you mind adding this to the top just in case any of these tracepoints are not available:
```
config = { missing_probes=warn }
```

Rtoax:
> @Rtoax Thanks for the addition. Do you mind adding this to the top just in case any of these tracepoints are not available:
> 
> ```
> config = { missing_probes=warn }
> ```

Fixed, thanks.

jordalgo (approved):

jordalgo:
@Rtoax I was about to merge but thinking maybe we should add a change log entry in the tools section. WDYT?

Rtoax:
@jordalgo Yes, you're right, added, thanks.

PR #4312, opened 2025-06-27
Make `get_pid_exe` failures not crash bpftrace

jordalgo:
As reported in the issue below a thrown error
in `get_pid_exe` was causing bpftrace to crash.
This makes sure to handle it more gracefully
by just returning a `Result`.

Issue:
https://github.com/bpftrace/bpftrace/issues/4309

##### Checklist

### DIFF
--- a/src/bpftrace.cpp
+++ b/src/bpftrace.cpp
@@ -1446,7 +1446,11 @@ std::vector<std::string> BPFtrace::resolve_usym_stack(uint64_t addr,
                                                       bool perf_mode,
                                                       bool show_debug_info)
 {
-  std::string pid_exe = util::get_pid_exe(pid);
+  std::string pid_exe;
+  auto res = util::get_pid_exe(pid);
+  if (res) {
+    pid_exe = *res;
+  }
   if (pid_exe.empty() && probe_id != -1) {
     // sometimes program cannot be determined from PID, typically when the
     // process does not exist anymore; in that case, try to get program name
--- a/src/util/system.cpp
+++ b/src/util/system.cpp
@@ -13,7 +13,14 @@
 
 namespace bpftrace::util {
 
-std::string get_pid_exe(const std::string &pid)
+char GetPidError::ID;
+
+void GetPidError::log(llvm::raw_ostream &OS) const
+{
+  OS << "Error code: " << err_;
+}
+
+Result<std::string> get_pid_exe(const std::string &pid)
 {
   std::error_code ec;
   std::filesystem::path proc_path{ "/proc" };
@@ -23,15 +30,11 @@ std::string get_pid_exe(const std::string &pid)
   try {
     return std::filesystem::read_symlink(proc_path).string();
   } catch (const std::filesystem::filesystem_error &e) {
-    auto err = e.code().value();
-    if (err == ENOENT || err == EINVAL)
-      return {};
-    else
-      throw e;
+    return make_error<GetPidError>(e.code().value());
   }
 }
 
-std::string get_pid_exe(pid_t pid)
+Result<std::string> get_pid_exe(pid_t pid)
 {
   return get_pid_exe(std::to_string(pid));
 }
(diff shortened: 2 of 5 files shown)


src/util/system.cpp
```diff
@@ -23,15 +30,11 @@ std::string get_pid_exe(const std::string &pid)
   try {
     return std::filesystem::read_symlink(proc_path).string();
   } catch (const std::filesystem::filesystem_error &e) {
-    auto err = e.code().value();
-    if (err == ENOENT || err == EINVAL)
-      return {};
-    else
-      throw e;
```
viktormalik, line 30: Aren't we losing some behaviour here? Previously, we would return an empty string on ENOENT and EINVAL and throw an error otherwise. Now, we always return an error, which we silently ignore. Shouldn't we at least print a warning on errors other than ENOENT and EINVAL?

jordalgo: > Shouldn't we at least print a warning on errors other than ENOENT and EINVAL?

I was debating about that. Perhaps it should be up to the caller to determine where to place the warning. The only place that looks useful (and not noisy) is `resolve_usym_stack`

viktormalik: Caller vs. callee is another discussion, my point was if we want to have unified behaviour for all error codes or not. But looking at the linked issue, it looks like we want to have it unified as there may be many different reasons why a link cannot be read.

jordalgo: > But looking at the linked issue, it looks like we want to have it unified as there may be many different reasons why a link cannot be read.

Right and in master we either return an empty string or crash, neither of which are very useful.

src/util/system.cpp
```diff
@@ -121,9 +124,10 @@ std::vector<std::string> get_mapped_paths_for_pid(pid_t pid)
   std::vector<std::string> paths;
 
   // start with the exe
-  std::string pid_exe = get_pid_exe(pid);
-  if (!pid_exe.empty() && pid_exe.find("(deleted)") == std::string::npos)
-    paths.push_back(get_pid_exe(pid));
+  auto pid_exe = get_pid_exe(pid);
+  if (pid_exe && !pid_exe->empty() &&
```
viktormalik: IIUC, we can drop the `!pid_exe->empty()` condition as we either return a path or an error. Occurs in multiple places.

src/util/system.cpp
viktormalik: It would be nice to print a warning with the error code (e.g. under `-v`) but I understand that it may be quite noisy in some cases.

jordalgo: I was also thinking a message to `LOG(V1)`, let's try it.

src/bpftrace.cpp
```diff
@@ -1449,8 +1449,12 @@ std::vector<std::string> BPFtrace::resolve_usym_stack(uint64_t addr,
...
                                                       bool show_debug_info)
 {
-  std::string pid_exe = util::get_pid_exe(pid);
-  if (pid_exe.empty() && probe_id != -1) {
+  std::string pid_exe;
+  auto res = util::get_pid_exe(pid);
+  if (res) {
+    pid_exe = *res;
+  }
+  if (!res && probe_id != -1) {
```
viktormalik: Better like this?
```suggestion
  } else if (probe_id != -1) {
```

jordalgo: Ha. I was looking at that thinking to myself "this doesn't look right". Thanks for fixing.

src/util/system.cpp
```diff
@@ -23,15 +30,13 @@ std::string get_pid_exe(const std::string &pid)
   try {
     return std::filesystem::read_symlink(proc_path).string();
```
amscanne, line 31: You shouldn't have the exception here at all. You can use the exceptionless variant, e.g. pass an error_code (https://en.cppreference.com/w/cpp/filesystem/read_symlink.html). @jordalgo

jordalgo: Ah neat. Will fix.

PR #4317, opened 2025-06-29
Add support for duration literals

jordalgo:
This includes: ns, us, ms, s, m, h, d

This also allows users to pass duration
literals into interval and profile probes
as the second argument e.g.

`interval:1s` instead of `interval:s:1`

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -428,6 +428,12 @@ To improve the readability of big literals an underscore `_` can be used as fiel
 Integer suffixes as found in the C language are parsed by bpftrace to ensure compatibility with C headers/definitions but they're not used as size specifiers.
 `123UL`, `123U` and `123LL` all result in the same integer type with a value of `123`.
 
+These duration suffixes are also supported: `ns`, `us`, `ms`, `s`, `m`, `h`, and `d`. All get turned into integer values in nanoseconds, e.g.
+```
+$a = 1m;
+print($a); // prints 60000000000
+```
+
 Character literals are not supported at this time, and the corresponding ASCII code must be used instead:
 
 ----
@@ -1315,6 +1321,7 @@ hardware:cache-misses:1e6 { @[pid] = count(); }
 === interval
 
 .variants
+* `interval:count`
 * `interval:us:count`
 * `interval:ms:count`
 * `interval:s:count`
@@ -1325,12 +1332,13 @@ hardware:cache-misses:1e6 { @[pid] = count(); }
 
 The interval probe fires at a fixed interval as specified by its time spec.
 Interval fires on one CPU at a time, unlike <<probes-profile>> probes.
+If a unit of time is not specified in the second position, the number is interpreted as nanoseconds; e.g., `interval:1s`, `interval:1000000000`, and `interval:s:1` are all equivalent.
 
 This prints the rate of syscalls per second.
 
 ----
 tracepoint:raw_syscalls:sys_enter { @syscalls = count(); }
-interval:s:1 { print(@syscalls); clear(@syscalls); }
+interval:1s { print(@syscalls); clear(@syscalls); }
 ----
 
 [#probes-iterator]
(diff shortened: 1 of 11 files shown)


man/adoc/bpftrace.adoc
```diff
@@ -428,6 +428,12 @@ To improve the readability of big literals an underscore `_` can be used as fiel
 Integer suffixes as found in the C language are parsed by bpftrace to ensure compatibility with C headers/definitions but they're not used as size specifiers.
 `123UL`, `123U` and `123LL` all result in the same integer type with a value of `123`.
 
+These duration suffixes are also supported: `ns`, `us`, `ms`, `s`, and `min`. All get turned into integer values in nanoseconds e.g.
```
jrife: nit: Maybe just `m` for minutes? `min` feels a little out of place contrasted with the second and subsecond abbreviation. Maybe I'm biased, since I'm used to seeing [Go style duration strings](https://pkg.go.dev/time#ParseDuration), but single letters feels more natural.

jordalgo: Yeah, I thought about that too. I'm fine with `m` and I know @amscanne is also partial to Go. I'll see if anyone else feels strongly.

viktormalik: `m` sounds good to me and feels more consistent with the rest.

amscanne: Although I'm partial to Go, I suspect that the reason C++ also chose `min` for duration literals is the chance of an accidental fat-finger of `ms` (which is likely to be the most common suffix).

I don't feel strongly though, and would also have a purely esthetic preference for `m` :)

I would add `h` and `d` just for completeness though, even if we don't expect them to be used as much I can imagine many scenarios where `h` is appropriate and it would be just annoying to have the others and not that.

man/adoc/bpftrace.adoc
```diff
@@ -1318,13 +1324,15 @@ hardware:cache-misses:1e6 { @[pid] = count(); }
 * `interval:us:count`
 * `interval:ms:count`
 * `interval:s:count`
+* `interval:count`
```
amscanne: Which option should users prefer? Here and below. It's unclear if there's a recommendation.

amscanne (approved):
LGTM notwithstanding the `m`/`min` question

man/adoc/bpftrace.adoc
jordalgo: I'll change to `m` and also add `h` and `d`.

man/adoc/bpftrace.adoc
jordalgo: Probably this new way. I find it more intuitive (and shorter by a char) to write `1s` over `s:1`

man/adoc/bpftrace.adoc
```diff
@@ -428,6 +428,12 @@ To improve the readability of big literals an underscore `_` can be used as fiel
 Integer suffixes as found in the C language are parsed by bpftrace to ensure compatibility with C headers/definitions but they're not used as size specifiers.
 `123UL`, `123U` and `123LL` all result in the same integer type with a value of `123`.
 
+These duration suffixes are also supported: `ns`, `us`, `ms`, `s`, `m`, `h`, and `d`. All get turned into integer values in nanoseconds e.g.
```
jrife: super nit: I think should be a comma before "e.g.": nanoseconds, e.g.

man/adoc/bpftrace.adoc
```diff
@@ -1325,12 +1332,13 @@ hardware:cache-misses:1e6 { @[pid] = count(); }
 
 The interval probe fires at a fixed interval as specified by its time spec.
 Interval fires on one CPU at a time, unlike <<probes-profile>> probes.
+If a unit of time is not specified in the second position, the number is interpreted as nanoseconds e.g. `interval:1s`, `interval:1000000000`, and `interval:s:1` are all equivalent.
```
jrife: "e.g." -> ", e.g.," or "; e.g." ?

jordalgo: I think it's "; e.g.,"

Issue #4305, opened 2025-06-26
Only cache symbols from targeted processes


ajor:
When symbol caching is set to "per_pid", we currently cache symbols from all running instances of a given process:
https://github.com/bpftrace/bpftrace/blob/2ab422d7e389bb3efad8bb33ad37d57bed13b5fd/src/usyms.cpp#L152-L153
https://github.com/bpftrace/bpftrace/blob/2ab422d7e389bb3efad8bb33ad37d57bed13b5fd/src/usyms.cpp#L194-L202

When processes are targeted using `-p` (e.g. `bpftrace -p 1234 myscript.bt`), we should only cache symbols from these targeted processes. This will improve startup time and memory usage.

pantaifeng:
> When symbol caching is set to "per_pid", we currently cache symbols from all running instances of a given process:
> 
> [bpftrace/src/usyms.cpp](https://github.com/bpftrace/bpftrace/blob/2ab422d7e389bb3efad8bb33ad37d57bed13b5fd/src/usyms.cpp#L152-L153)
> 
> Lines 152 to 153 in [2ab422d](/bpftrace/bpftrace/commit/2ab422d7e389bb3efad8bb33ad37d57bed13b5fd)
> 
>  for (int pid : util::get_pids_for_program(elf_file)) 
>    pid_sym_[pid] = bcc_symcache_new(pid, &get_symbol_opts()); 
> 
> [bpftrace/src/usyms.cpp](https://github.com/bpftrace/bpftrace/blob/2ab422d7e389bb3efad8bb33ad37d57bed13b5fd/src/usyms.cpp#L194-L202)
> 
> Lines 194 to 202 in [2ab422d](/bpftrace/bpftrace/commit/2ab422d7e389bb3efad8bb33ad37d57bed13b5fd)
> 
>  for (int pid : util::get_pids_for_program(elf_file)) { 
>    blaze_cache_src_process cache = { 
>      .type_size = sizeof(cache), 
>      .pid = static_cast<uint32_t>(pid), 
>      .cache_vmas = true, 
>    }; 
>   
>    blaze_symbolize_cache_process(symbolizer_, &cache); 
>  } 
> When processes are targeted using `-p` (e.g. `bpftrace -p 1234 myscript.bt`), we should only cache symbols from these targeted processes. This will improve startup time and memory usage.

Looking forward to this feature, I hope you can release it soon

PR #4318, opened 2025-06-29
Limit per-pid symbol caching to explicitly targeted PIDs (#4305)

Aevil1:
This PR updates the behavior of symbol caching when --pid is specified and user_symbol_cache_type is set to per_pid. Previously, bpftrace would cache symbols for all running instances of a given ELF file. With this change, only the explicitly targeted PIDs will have their symbols cached. This applies to both BCC and Blazesym symbolizers.

Benefits:

-Reduces startup time
-Decreases memory usage
-Avoids unnecessary symbol cache creation for unrelated processes

##### Checklist

- [ x] Language changes are updated in `man/adoc/bpftrace.adoc`
- [ x] User-visible and non-trivial changes updated in `CHANGELOG.md`
- [ x] The new behaviour is covered by tests

### DIFF
--- a/src/usyms.cpp
+++ b/src/usyms.cpp
@@ -115,6 +115,11 @@ Usyms::Usyms(const Config &config) : config_(config)
 {
 }
 
+Usyms::Usyms(const Config &config, const std::set<int> &targeted_pids)
+    : config_(config), targeted_pids_(targeted_pids), has_targeted_pids_(!targeted_pids.empty())
+{
+}
+
 Usyms::~Usyms()
 {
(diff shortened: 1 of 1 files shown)


jordalgo:
@Aevil1 Thanks for looking into this but we just landed a fix for this:
- https://github.com/bpftrace/bpftrace/pull/4315

Here are some other good starter issues if you're interested:
- https://github.com/bpftrace/bpftrace/issues/3549
- https://github.com/bpftrace/bpftrace/issues/4316
- https://github.com/bpftrace/bpftrace/issues/4314

PR #4320, opened 2025-06-30
util: add OpaqueValue to replace `std::vector<uint8_t>`

amscanne:
Stacked PRs:
 * #4323
 * #4322
 * #4321
 * __->__#4320

--- --- ---

### util: add OpaqueValue to replace `std::vector<uint8_t>`

This can be used to add safety and reduce copying. It is intended to
replace the ad-hoc ues of the `util::read_data` and other similar
methods which do not have any inherit bounds checks.

### DIFF
--- a/src/util/opaque.cpp
+++ b/src/util/opaque.cpp
@@ -1,0 +1,56 @@
+#include <cstring>
+#include <functional>
+#include <iomanip>
+
+#include "util/opaque.h"
+
+namespace bpftrace::util {
+
+bool OpaqueValue::operator==(const OpaqueValue &other) const
+{
+  if (size() != other.size()) {
+    return false;
+  }
+  return std::memcmp(data(), other.data(), size()) == 0;
+}
+
+bool OpaqueValue::operator!=(const OpaqueValue &other) const
+{
+  return !(*this == other);
+}
+
+bool OpaqueValue::operator<(const OpaqueValue &other) const
+{
+  size_t min_len = std::min(size(), other.size());
+  int cmp = std::memcmp(data(), other.data(), min_len);
+  if (cmp != 0) {
+    return cmp < 0;
+  }
+  return size() < other.size();
+}
+
+size_t OpaqueValue::hash() const
+{
+  const char *backing_data = data();
+  // We don't expect hashing to be common, but we do use the OpaqueValue as the
+  // key for some map. We just use the simple djb2 algorithm, which should be
+  // fine across a range of different data.
+  size_t hash_value = 5381;
+  for (size_t i = 0; i < size(); i++) {
+    hash_value = ((hash_value << 5) + hash_value) +
+                 static_cast<unsigned char>(backing_data[i]);
+  }
+  return hash_value;
+}
+
+std::ostream &operator<<(std::ostream &out, const OpaqueValue &value)
+{
+  const char *data = value.data();
+  for (size_t i = 0; i < value.size(); i++) {
+    out << std::hex << std::setfill('0') << std::setw(2)
+        << static_cast<unsigned int>(static_cast<unsigned char>(data[i]));
+  }
+  return out;
+}
+
+} // namespace bpftrace::util
--- a/src/util/opaque.h
+++ b/src/util/opaque.h
@@ -1,0 +1,205 @@
+#pragma once
+
+#include <cstdlib>
+#include <cstring>
+#include <functional>
+#include <memory>
+#include <optional>
+#include <utility>
+#include <variant>
+
+namespace bpftrace::util {
+
+// OpaqueValue corresponds to a blob of type-erased memory. It owns its own
+// underlying memory, taking a copy if necessary. This allows it to be used
+// to read maps, rings, and other locations which may not be permanent.
+//
+// This class does not really provide safety beyond first-order memory safety.
+class OpaqueValue {
+private:
+  struct OwnedBuffer {
+    OwnedBuffer(char *data) : data(data) {};
+    OwnedBuffer(const OwnedBuffer &other) = delete;
+    OwnedBuffer(OwnedBuffer &&other) = delete;
+    OwnedBuffer &operator=(const OwnedBuffer &other) = delete;
+    OwnedBuffer &operator=(OwnedBuffer &&other) = delete;
+    ~OwnedBuffer()
+    {
+      free(data);
+    }
+    char *data;
+  };
+  using SharedMemory = std::shared_ptr<OwnedBuffer>;
+  using Storage = std::variant<uintptr_t, SharedMemory>;
+
+public:
+  // Allows for the creation of an OpaqueValue.
+  static OpaqueValue alloc(size_t length, std::function<void(char *)> init)
+  {
+    return { length, init };
+  }
+
+  // Create a zerod OpaqueValue.
+  static OpaqueValue alloc(size_t length)
+  {
+    return { length, [&](char *data) { memset(data, 0, length); } };
+  }
+
+  // Create a copy of a string.
+  static OpaqueValue string(const std::string &s, size_t sz)
+  {
+    return { sz, [&](char *data) {
+              size_t s_sz = std::min(s.size() + 1, sz);
+              memcpy(data, s.data(), s_sz);
+              memset(data + s_sz, 0, sz - s_sz);
+            } };
+  }
+
+  // Merge two OpaqueValues together.
+  OpaqueValue operator+(const OpaqueValue &other) const
+  {
+    return { size() + other.size(), [&](char *new_data) {
+              memcpy(new_data, data(), size());
+              memcpy(new_data + size(), other.data(), other.size());
+            } };
+  }
+
+  // Creates a copy of any type that has an iterator.
+  //
+  // This requires that the underlying type has a simple `from`, so we can't do
+  // complex types like strings in this way. This is intended for use in tests
+  // only and is therefore not optimized for performance in any way.
+  template <std::ranges::range T>
+  static OpaqueValue from(const T &other)
+  {
+    OpaqueValue res;
+    for (const auto &v : other) {
+      res = res + from(v);
+    }
+    if constexpr (std::is_same_v<T, std::string>) {
+      res = res + OpaqueValue::from<char>(0);
+    }
+    return res;
+  }
+
+  // Creates an OpaqueValue from an existing region.
+  template <typename T>
+  static OpaqueValue from(const T *orig_data, size_t length)
+  {
+    return { sizeof(T) * length, [&](char *data) {
+              memcpy(data,
+                     reinterpret_cast<const char *>(orig_data),
+                     sizeof(T) * length);
+            } };
+  }
+
+  // Creates a new OpaqueValue from any trivial type.
+  template <typename T>
+    requires(std::is_trivially_copyable_v<T>)
+  static OpaqueValue from(const T &other)
+  {
+    return { sizeof(other),
+             [&](char *data) { memcpy(data, &other, sizeof(other)); } };
+  }
+
+  // Returns bytes available.
+  size_t size() const
+  {
+    return length_ - offset_;
+  }
+
+  // Return the number of values that may fit.
+  template <typename T>
+  size_t count() const
+  {
+    return size() / sizeof(T);
+  }
+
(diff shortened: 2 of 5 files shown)


src/util/opaque.h
```diff
@@ -0,0 +1,197 @@
...
+
+  // Creates an OpaqueValue from an existing region.
+  template <typename T>
+  static OpaqueValue from(const T *orig_data, size_t length)
+  {
+    return OpaqueValue(sizeof(T) * length, [&](char *data) {
+      memcpy(data,
+             reinterpret_cast<const char *>(orig_data),
+             sizeof(T) * length);
+    });
```
viktormalik: Use the same style as elsewhere?
```suggestion
    return { sizeof(T) * length, [&](char *data) {
             memcpy(data,
                    reinterpret_cast<const char *>(orig_data),
                    sizeof(T) * length);
           } };
```

src/util/opaque.h
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  // Creates a new OpaqueValue from any trivial type.
+  template <typename T>
+    requires(std::is_trivial_v<T>)
+  static OpaqueValue from(const T &other)
+  {
+    return OpaqueValue(sizeof(other), [&](char *data) {
+      memcpy(data, &other, sizeof(other));
+    });
```
viktormalik: Same as above
```suggestion
    return { sizeof(other), [&](char *data) {
             memcpy(data, &other, sizeof(other));
            } };
```

src/util/opaque.h
```diff
@@ -0,0 +1,197 @@
...
+  {
+    return { sz, [&](char *data) {
+              size_t s_sz = std::min(s.size() + 1, sz);
+              memcpy(data, s.data(), s_sz);
+              memset(data + s_sz, 0, sz - s_sz);
+            } };
+  }
+
+  // Merge two OpaqueValues together.
+  OpaqueValue operator|(const OpaqueValue &other) const
```
viktormalik: Does this operator need to be public? I'd prefer hiding as much complexity (templates, overloaded operators, etc.) as possible inside the classes.

amscanne: There are tests that need to construct the "per-cpu" variant of data, which is the reason this operator was added. I can make it a proper method or name it something else, but it exists to construct values like:

```
auto a = OpaqueValue::from<int64_t>(1);
auto b = OpaqueValue::from<int64_t>(1):
auto c = OpaqueValue::from<int64_t>(1):
auto agg = a|b|c;
```

(I'm not sure why I picked an operator specifically, just looking for something with good ergonomics. More than happy to just make it a method, maybe `concat`?)

viktormalik: I'm not sure TBH. I kind of like the ergonomics of the operator but `|` is a bit confusing to me. I'm probably bike-shedding here but how about `+` which is normally used for string concatenation? That would at least resemble something existing.

src/util/opaque.h
amscanne: Done.

src/util/opaque.h
amscanne: Done.

src/util/opaque.h
amscanne: I've replaced it with `operator+` which is more standardized, I'm good with that.

amscanne:
@jordalgo Any hard objections to `OpaqueValue`? I'd also like to wrap up this stack ASAP given how long it's been around (and how much rebasing is needed).

src/util/opaque.h
```diff
@@ -0,0 +1,196 @@
...
+// OpaqueValue corresponds to a blob of type-erased memory. It owns its own
+// underlying memory, taking a copy if necessary. This allows it to be used
+// to read maps, rings, and other locations which may not be permanent.
+//
+// This class does not really provide safety beyond first-order memory safety.
+class OpaqueValue {
+private:
+  struct OwnedBuffer {
+    OwnedBuffer(char *data) : data(data) {};
+    ~OwnedBuffer()
```
jordalgo, line 26: Do we need to delete the copy and move constructors for this to prevent accidental double frees?

src/util/opaque.h
```diff
@@ -0,0 +1,196 @@
...
+    {
+      free(data);
+    }
+    char *data;
+  };
+  using SharedMemory = std::shared_ptr<OwnedBuffer>;
+  using Variant = std::variant<uintptr_t, SharedMemory>;
```
jordalgo: Not really a fan of this name `Variant` just to have an alias for this small `std::variant` - bad for searches/groking

src/util/opaque.h
```diff
@@ -0,0 +1,196 @@
...
+    return { sizeof(T) * length, [&](char *data) {
+              memcpy(data,
+                     reinterpret_cast<const char *>(orig_data),
+                     sizeof(T) * length);
+            } };
+  }
+
+  // Creates a new OpaqueValue from any trivial type.
+  template <typename T>
+    requires(std::is_trivial_v<T>)
```
jordalgo: Looks like `is_trivial` is being deprecated - can we use an alternative?
https://stackoverflow.com/questions/79222674/why-is-stdis-trivial-deprecated-in-c26

src/util/opaque.h
```diff
@@ -0,0 +1,196 @@
...
+  //
+  // This does not copy any memory, and relies on the original ownership
+  // semantics for the current OpaqueValue.
+  OpaqueValue slice(size_t offset,
+                    std::optional<size_t> length = std::nullopt) const
+  {
+    size_t l = size() - offset;
+    if (length) {
+      l = *length;
+    }
```
jordalgo: style nit
```suggestion
    size_t l = length ? *length : size() - offset;
```

src/util/opaque.h
```diff
@@ -0,0 +1,196 @@
...
+  OpaqueValue(Variant mem, size_t offset, size_t length)
+      : mem_(std::move(mem)), offset_(offset), length_(length) {};
+
+  void check(size_t offset, size_t length) const
+  {
+    // Check for overflow and underflow.
+    if (offset_ + offset < offset_ ||
+        offset_ + offset + length < offset_ + offset ||
+        offset_ + offset + length > length_) {
+      throw std::bad_alloc();
```
jordalgo, line 184: JC, why is `check` throwing a `bad_alloc` here?

src/util/opaque.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+bool OpaqueValue::operator<(const OpaqueValue &other) const
+{
+  size_t min_len = std::min(size(), other.size());
+  int cmp = std::memcmp(data(), other.data(), min_len);
+  if (cmp != 0) {
+    return cmp < 0;
+  }
+  return size() < other.size();
```
jordalgo, line 29: JC, why not do the size comparison first before the actual byte comparison?

src/util/opaque.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+size_t OpaqueValue::hash() const
+{
+  size_t hash_value = 0;
+  const char *backing_data = data();
+  for (size_t i = 0; i < size(); i++) {
+    hash_value = hash_value * 31 + static_cast<unsigned char>(backing_data[i]);
```
jordalgo: From a quick google, looks like the common DJB2 hash multiplies by 33. 31 is obviously also prime so it probs doesn't matter but maybe good for a quick comment so other noobs like me don't have to ask the robots.

jordalgo (commented):
Overall looks good. A few minor nits and questions.

src/util/opaque.cpp
amscanne: It's a basic lexicographical ordering. The length only matters if the prefix of the longer one is equal to the shorter one.

src/util/opaque.cpp
amscanne: Interesting, I'm not sure what you mean by "also prime", since 33 is not prime at all :) It seems like 33 is commonly used for strings and English text since it works well in practice. However, since it is not prime there are some collisions during overflow, so the hash range is guaranteed to be smaller.

But it really doesn't matter much, and it having a reference point is useful, so I'll just use DJB2.

src/util/opaque.h
amscanne: This is called on slicing, etc. I didn't want to have a error-based API for this case and while I'm somewhat staunchly opposed to exceptions it feels like this is case where it makes sense. This is what is permitted to be throw by allocation functions (i.e. the thing called by `operator new`, etc.). Since the `slice` operator is logically slicing/allocating a new object, I throw this when you are doing so on an invalid range.

src/util/opaque.h
amscanne: Done, will be changing on new rebase.

src/util/opaque.h
amscanne: Done, will be changing on new rebase.

src/util/opaque.h
amscanne: It's not strictly needed because this is a private internal class, but I'm always for explicitness and safety. Added.

src/util/opaque.h
jordalgo: Is anything setup to catch these exceptions or will that be added later? I'm just curious as to why you wouldn't use `LOG(BUG)` - like do we foresee this (or rather the actual malloc) happening occasionally on systems with low memory (e.g. embedded) ?

src/util/opaque.cpp
jordalgo: 🧠

src/util/opaque.cpp
jordalgo: > since 33 is not prime at all 

I told you I was a noob!

src/util/opaque.h
amscanne: Done.

src/util/opaque.h
amscanne: `malloc` never actually fails, since virtual memory exists ;)

I'm not a huge fan of `LOG(BUG)` because of the dependency inversion problem -- this is an library that can be tested in isolation, and it doesn't need to depend on `libbpftrace` (incorrectly). I just want it to die if you try to do undefined behavior, rather than that undefined behavior manifesting later on as e.g. a SEGV (current behavior). An exception is a reasonable path for these cases, as it can be debugged & caught easily (whether exceptions are enabled or not). Actually must easier than even `LOG(BUG)`.

The viable alternative would be an assertion failure, which I would be fine switching to if you I shouldn't have an exception, e.g. `assert(!(....) && "attempted slice beyond bounds of existing value");`

I have added a pile of tests that are also testing this behavior.

src/util/opaque.cpp
amscanne: Nah, your point was valid (even if your prime number detector is broken ;)

amscanne:
@jordalgo Anything else outstanding that I should resolve? Happy to follow up also

src/util/opaque.h
jordalgo: >  this is an library that can be tested in isolation

Yeah this is a bit what I was alluding to. For things that don't exist as "libraries" in the bpftrace code base the convention is to use `LOG(BUG)` - I'm fine to change that (as it seems you're making a case that assertions and uncaught exceptions are better) - we should just make a task for it so we can change it uniformly. 

> I have added a pile of tests that are also testing this behavior.

Thanks for adding. They look great.

jordalgo (approved):

src/util/opaque.h
amscanne: > Yeah this is a bit what I was alluding to. For things that don't exist as "libraries" in the bpftrace code base the convention is to use `LOG(BUG)` - I'm fine to change that (as it seems you're making a case that assertions and uncaught exceptions are better) - we should just make a task for it so we can change it uniformly.

I filed #3731 a while ago, and I think it generally describes my position here. Note that I'm not advocating for uncaught exceptions widely, only in very rare cases where they map into an existing runtime exception in a very clear and obvious way. (Like in this case, where it is arguably a low-level memory error, as the caller has failed to bounds check correctly. An exception makes sense here as it's certainly *possible*, but would be considered undefined behavior to proceed, while an assertion should be used for the *this can never happen* checks)

As I think on it, I am tempted to just make this an assert to bias against exceptions, but I don't want to spend time rebasing on this. We should just make it consistent as part of #3731 over time.

PR #4321, opened 2025-06-30
core: update internals to `OpaqueValue` from `util::read_data`

amscanne:
core: update internals to `OpaqueValue` from `util::read_data`

This ensures that all accesses to memory can be suitably bounds-check
(with the exception of direct calls to `data`).


tests/bpftrace.cpp
```diff
@@ -1755,6 +1707,10 @@ TEST(bpftrace, print_lhist_map)
     std::string expected_output;
   };
 
+  const HistogramMap values_by_key = {
```
amscanne, line 1785: @viktormalik FYI, `OpaqueValue` is the follow through that is fixing the endianness problems :) No more `std::vector<uint8_t>` everywhere with copies and unbounded accesses!

src/types_format.cpp
```diff
@@ -37,61 +36,59 @@ Result<output::Primitive> format(BPFtrace &bpftrace,
...
                                 8);
     }
     case Type::ustack_t: {
-      return bpftrace.get_stack(util::read_data<uint64_t>(value.data()),
-                                util::read_data<uint64_t>(value.data() + 8),
-                                util::read_data<int32_t>(value.data() + 16),
-                                util::read_data<int32_t>(value.data() + 20),
+      return bpftrace.get_stack(value.bitcast<uint64_t>(0),
+                                value.bitcast<uint64_t>(1),
+                                value.slice(16, 4).bitcast<int32_t>(),
```
viktormalik: It's a bit more verbose but how about we use
```suggestion
                                value.slice(2 * sizeof(uint64_t), sizeof(int32_t)).bitcast<int32_t>(),
```
That makes the intent super-clear.

Same would apply to multiple occurences of `slice` below.

src/bpfmap.h
```diff
@@ -33,16 +33,14 @@ class BpfMapError : public ErrorInfo<BpfMapError> {
...
-using KeyType = std::vector<uint8_t>;
-using KeyVec = std::vector<KeyType>;
-using ValueType = std::vector<uint8_t>;
-using MapElements = std::vector<std::pair<KeyType, ValueType>>;
+using util::OpaqueValue;
+using MapElements = std::vector<std::pair<OpaqueValue, OpaqueValue>>;
 using BucketUnit = uint64_t;
-using BucketType = std::vector<BucketUnit>;
-using HistogramMap = std::map<KeyType, BucketType>;
+using BucketType = std::vector<uint64_t>;
```
viktormalik: I'd either keep `BucketUnit` here or drop it completely.

src/bpfmap.cpp
```diff
@@ -175,55 +168,48 @@
...
-      values_by_key[key_prefix] = TSeries();
-    }
-
-    auto v = util::reduce_tseries_value(
-        value, nvalues, tseries_args.value_type, tseries_args.agg);
-    values_by_key[key_prefix][v.second] = v.first;
-    old_key = key.data();
+    auto prefix = key.slice(0, map_info.key_type.GetSize());
+    auto bucket = key.slice(map_info.key_type.GetSize(), sizeof(uint64_t));
+    auto [tseries, _] = values_by_key.try_emplace(prefix);
```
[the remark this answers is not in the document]
viktormalik: This looks like a reasonable suggestion.

viktormalik (commented):
`OpaqueValue` is really handy! Just a few comments below.

src/types_format.cpp
amscanne: Updated.

src/bpfmap.h
amscanne: Dropped.

src/bpfmap.cpp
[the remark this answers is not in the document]
amscanne: Done.

PR #4322, opened 2025-06-30
core: use `OpaqueValue` for async dispatch

amscanne:
Stacked PRs:
 * #4323
 * __->__#4322

--- --- ---

### core: use `OpaqueValue` for async dispatch

This effectively adds suitable bounds checking to all async dispatch
functions and standardizes the way casting and pointer munging works.

### DIFF
--- a/tests/codegen/llvm/call_printf.ll
+++ b/tests/codegen/llvm/call_printf.ll
@@ -4,7 +4,8 @@ target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n32:64-S128"
 target triple = "bpf"
 
 %"struct map_t" = type { ptr, ptr }
-%printf_t = type { i64, i64, i64 }
+%printf_t = type { i64, %printf_args_t }
+%printf_args_t = type { i64, i64 }
 
 @LICENSE = global [4 x i8] c"GPL\00", section "license", !dbg !0
 @ringbuf = dso_local global %"struct map_t" zeroinitializer, section ".maps", !dbg !7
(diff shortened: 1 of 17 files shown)


tests/codegen/llvm/call_printf.ll
```diff
@@ -31,39 +32,40 @@ entry:
...
+  %9 = load i8, ptr %"struct Foo.c", align 1
   call void @llvm.lifetime.end.p0(i64 -1, ptr %"struct Foo.c")
-  %9 = getelementptr %printf_t, ptr %printf_args, i32 0, i32 1
-  %10 = sext i8 %8 to i64
-  store i64 %10, ptr %9, align 8
-  %11 = load i64, ptr %"$foo", align 8
-  %12 = inttoptr i64 %11 to ptr
-  %13 = call ptr @llvm.preserve.static.offset(ptr %12)
-  %14 = getelementptr i8, ptr %13, i64 8
+  %10 = getelementptr %printf_args_t, ptr %4, i32 0, i32 0
```
jordalgo, line 44: seems now we have `printf_args` and `printf_args_t` what is the diff?

amscanne: The former was a dynamically-constructed type that included the ID (and there was code that would add the ID, then add all the arguments and compute offsets, then *remove* the ID from the `SizedType` just so that the offsets match the expectation).

Now there are two types: the actual arguments being passed, and the arguments pushed to the ring buf (which has the arguments as a member). Effectively:

```
printf_args {
  uint64_t id;
  printf_args_t args;
};
printf_args_t {
   ... actual tuple constructed ..
 };
 ```

jordalgo: Ah gotcha. Thanks!

jordalgo (approved):

PR #4323, opened 2025-06-30
printf: use `OpaqueValue` for formatting

amscanne:
Stacked PRs:
 * __->__#4323

--- --- ---

### printf: use `OpaqueValue` for formatting

This aligns the intermediate output types with the intermediate format
types. Note that the support is provided by the output's native ability
to print the various primitives. So `%s%`, for example, is now able to
handle almost any object type.

An incidental change in this commit is that overflow is now correctly
handled on the kernel side. We read an extra character, and whether the
string is overflowed or not is indicated by whether `str[N]` contains a
NULL. (If it does and is "well-formed", then no overflow has occured).


tests/codegen/llvm/str_scratch_buf.ll
```diff
@@ -21,11 +21,11 @@ entry:
   %1 = load i64, ptr @__bt__max_cpu_id, align 8
   %cpu.id.bounded = and i64 %get_cpu_id, %1
   %2 = getelementptr [1 x [1 x [64 x i8]]], ptr @__bt__get_str_buf, i64 0, i64 %cpu.id.bounded, i64 0, i64 0
-  call void @llvm.memset.p0.i64(ptr align 1 %2, i8 0, i64 64, i1 false)
+  call void @llvm.memset.p0.i64(ptr align 1 %2, i8 -1, i64 65, i1 false)
```
jordalgo, line 24: This `-1` for the memset looks sus

tests/semantic_analyser.cpp
```diff
@@ -2971,13 +2972,6 @@ TEST_F(SemanticAnalyserTest, enums)
   test("enum named { a = 1, b } kprobe:f { printf(\"%s\", a); }");
   test("enum named { a = 1, b } kprobe:f { $e = a; printf(\"%s\", $e); }");
   test("enum named { a = 1, b } kprobe:f { printf(\"%15s %-15s\", a, a); }");
-
-  // Cannot symbolize a non-enum
-  test("kprobe:f { $x = (uint8)1; printf(\"%s\", $x) }", Error{ R"(
```
jordalgo, line 2984: Should we replace this with a positive test?

tests/codegen/llvm/str_scratch_buf.ll
amscanne: It is intentional. Prior to `probe_read_str` we now poison the memory. See here: https://github.com/bpftrace/bpftrace/pull/4323/files#diff-28926c090177e6efbfb5a7449833e0e863bb226822abb513b7dfc6c388837d87R1403

I could actually make this `0` and it won't make a different to the semantics of detecting actual truncation, but without the poison we don't have the ability to tell the different between a failed read and empty string.

tests/semantic_analyser.cpp
amscanne: You can now `%s` any type, as it is merely the `TextOutput` representation of that primitive. I'm not inclined to add every one of those cases individually as tests? But open to suggestions.

jordalgo: We don't have to be exhaustive but a few positive tests with some primitives that didn't previously work when trying to be printed as strings would be nice.

tests/codegen/llvm/str_scratch_buf.ll
jordalgo: Got it. Thanks for explaining (the comment in the code also helps)

tests/semantic_analyser.cpp
amscanne: Will do

jordalgo:
Seeing a seemingly related test failure: `strcontains.no literals`

amscanne:
> Seeing a seemingly related test failure: `strcontains.no literals`

Yeah, I think the +1 in `str` size (ensuring `max_strlen=1` will have a valid string of length 1) caused the complexity of `strcontains` to spill over. I've just lowered this by 1 for this test, but I think we can separately fix this `strcontains` problem with loops and macros now. (I've pushed a separate PR but it will need to land after a bunch of the cleanup PRs.)

jordalgo (approved):

PR #4325, opened 2025-06-30
Use non-throwing version of read_symlink

jordalgo:
For `get_pid_exe`.
Also remove the Result return value and
just return an optional since no callers
are consuming the error.

##### Checklist

### DIFF
--- a/src/util/system.cpp
+++ b/src/util/system.cpp
@@ -13,30 +13,23 @@
 
 namespace bpftrace::util {
 
-char GetPidError::ID;
-
-void GetPidError::log(llvm::raw_ostream &OS) const
-{
-  OS << "Error code: " << err_;
-}
-
-Result<std::string> get_pid_exe(const std::string &pid)
+std::optional<std::string> get_pid_exe(const std::string &pid)
 {
   std::error_code ec;
   std::filesystem::path proc_path{ "/proc" };
   proc_path /= pid;
   proc_path /= "exe";
+  auto res = std::filesystem::read_symlink(proc_path, ec);
 
-  try {
-    return std::filesystem::read_symlink(proc_path).string();
-  } catch (const std::filesystem::filesystem_error &e) {
-    int err = e.code().value();
-    LOG(V1) << "Error reading " << proc_path << ". EC: " << err;
-    return make_error<GetPidError>(err);
+  if (ec) {
+    LOG(V1) << "Error reading " << proc_path << " - " << ec.message();
+    return std::nullopt;
   }
+
+  return res.string();
 }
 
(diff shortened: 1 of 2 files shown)


src/util/system.cpp
```diff
@@ -13,30 +13,23 @@
...
 
-  try {
-    return std::filesystem::read_symlink(proc_path).string();
-  } catch (const std::filesystem::filesystem_error &e) {
-    int err = e.code().value();
-    LOG(V1) << "Error reading " << proc_path << ". EC: " << err;
-    return make_error<GetPidError>(err);
+  if (ec) {
+    LOG(V1) << "Error reading " << proc_path << " - " << ec.message();
+    return std::nullopt;
```
amscanne, line 26: If you are logging the error, then just return it. I understand that it's prevalent, but IMHO it's an anti-pattern to have libraries generate log spam via global singletons. Typically you want to be able to wrap this and stub at least for testing. From the user's POV, then `Result<>` is the same as the optional, in that they can just check and ignore it OR they get the option to log the error.

The `GetPidError` was overly specific though, you could just do a `SystemError` or whatever that takes an `errno` and use it very broadly.

jordalgo: I don't think we need to be this prescriptive about internal libraries. If we return the error than every call site will have to do the check and log it even though we want this behavior globally. If at some point in the future we want to be more targeted about where we show this error then yes, I agree that it should just return it in a Result. At the moment, we have multiple util functions calling `LOG` - how is this different?

amscanne: I'm not advocating that we stop everything and go and fix those cases, but it would be good to have the gradient pointed the right way :) You're replacing a well-structured library API & error with something that eats the error here... why? The caller would literally even have the same code, e.g. `if (res) { .... }` and could just do: `if (!res) { LOG(WARNING) << ....; }` if they wanted to.

Logging deep in the bowels of libraries is a kind of technical debt IMO, because then we end up throwing extra stuff to capture output, turn it off just for tests, etc. If we wanted to make these things link together in a sane way, we're already in a bad way, since `libbpftrace => util => libbpftrace` here.

In general, if you don't need to use a global singleton, you shouldn't and we should just have libraries that are independently testable and modularized. (If there's a reason that this is deeply complicated for some case, that's fine... but when it's not, why not?)

jordalgo: The point is we don't know which callers we want to have log this error, which is why it's global. The time spent litigating such a small log to verbose-only output is just not worth it IMHO. I can close this PR if you'd like to follow-up with a proper fix and investigate the different call sites and evaluate if they should individually be logging this error message - just let me know.

PR #4326, opened 2025-06-30
Remove AST location from unstable feature warning

jordalgo:
This is mostly so we can add new macros to base.bt and have the output be less noisy. Seems also like good general practice to remove the source line
for this warning as these features are enabled by
default and we don't need to beat our users over the head with some additional details of where the
unstable feature is being used.

Also made the warning and error message a little better.

##### Checklist

### DIFF
--- a/src/ast/passes/unstable_feature.cpp
+++ b/src/ast/passes/unstable_feature.cpp
@@ -11,19 +11,24 @@ namespace bpftrace::ast {
 
 namespace {
 
-std::string get_warning(const std::string &feature)
+const auto MAP_DECL = "map declarations";
+const auto IMPORTS = "imports";
+const auto MACROS = "macros";
+
+std::string get_warning(const std::string &feature, const std::string &config)
 {
-  return std::string("Script is using an unstable feature. To prevent this "
-                     "warning you must explicitly enable the unstable "
-                     "feature in the config e.g. ") +
-         feature + std::string("=enable");
+  return std::string("Script is using an unstable feature: " + feature +
+                     ". To prevent this warning you must explicitly enable it "
+                     "in the config e.g. ") +
+         config + std::string("=enable");
 }
 
-std::string get_error(std::string &&feature)
+std::string get_error(const std::string &feature, const std::string &config)
 {
-  return std::string("Feature not enabled by default. To enable "
+  return std::string(feature +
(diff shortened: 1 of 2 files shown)


src/ast/passes/unstable_feature.cpp
```diff
@@ -11,19 +11,25 @@ namespace bpftrace::ast {
...
+std::string get_warning(const std::string &feature, const std::string &config)
 {
-  return std::string("Script is using an unstable feature. To prevent this "
-                     "warning you must explicitly enable the unstable "
-                     "feature in the config e.g. ") +
-         feature + std::string("=enable");
+  return std::string("Script is using an unstable feature: " + feature +
+                     ". To prevent this "
+                     "warning you must explicitly enable it "
+                     "in the config e.g. ") +
```
viktormalik: Nit: this can be probably wrapped better and fit two lines.

src/ast/passes/unstable_feature.cpp
```diff
@@ -46,39 +52,39 @@ class UnstableFeature : public Visitor<UnstableFeature> {
...
 {
   if (bpftrace_.config_->unstable_map_decl == ConfigUnstable::error) {
-    decl.addError() << get_error(UNSTABLE_MAP_DECL);
+    decl.addError() << get_error(MAP_DECL, UNSTABLE_MAP_DECL);
     return;
   }
   if (bpftrace_.config_->unstable_map_decl == ConfigUnstable::warn &&
       !warned_features.contains(UNSTABLE_MAP_DECL)) {
-    decl.addWarning() << get_warning(UNSTABLE_MAP_DECL);
+    LOG(WARNING) << get_warning(MAP_DECL, UNSTABLE_MAP_DECL);
```
viktormalik, line 64: Ah, so we replace `.addWarning()`, which is associated with an AST node, i.e. implicitly contains a location, by `LOG(WARNING)` which just prints the string without additional data, is that right?

If so, can we add this kind of an explanation to the commit message? Took me some time to figure out.

jordalgo: Yeah, that's exactly right. Sorry if I didn't spell it out more. I'll add a comment.

jordalgo: Added.

viktormalik (approved):

PR #4328, opened 2025-07-01
Update kernel dependency support

jordalgo:
Currently bpftrace fails in multiple places when
run on 5.4 (the oldest LTS kernel) and even on 5.10. It's possible we can fix some of these issues by
vendoring kernel headers from these versions but
until we do so our dependency support policy needs to be updated.

Picking 5.15 to be conservative and making a general policy of supporting the 4 most recent LTS versions.

##### Checklist

### DIFF
--- a/docs/dependency_support.md
+++ b/docs/dependency_support.md
@@ -6,17 +6,18 @@ usually at distro/user request.
 
 ## Linux kernel
 
-The linux kernel is bpftrace's biggest runtime dependency. Our stance on kernel
-support is that for **cross cutting** implementation details (ie. something
-every reasonably sophisticated user is expected to depend on), we'll support up
-to and including the oldest LTS kernel still in service.
+**Minimum kernel version: 5.15**
+
+The linux kernel is bpftrace's biggest runtime dependency.
+While we would like to support the oldest LTS kernel version, we also want to make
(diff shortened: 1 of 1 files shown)


viktormalik (approved):
Ok with me.

docs/dependency_support.md
```diff
@@ -6,18 +6,19 @@ usually at distro/user request.
...
+**Minimum kernel version: 5.15**
 
-For features that have clear boundaries (eg. new builtins or helpers), bpftrace
-is free to opportunistically depend on newer kernels as long as there is a
-reasonable runtime fallback strategy. An error message is a reasonable fallback
-strategy in this case.
+The linux kernel is bpftrace's biggest runtime dependency.
+While we would like to support the oldest LTS kernel version, we also want to make
+sure bpftrace is keeping up with the latest features. As a compromise bpftrace
+supports stable kernels and the 4 most recent LTS kernels.
```
amscanne: Optional: you could also just say something like "we support stable kernels and as many LTS kernels as we can, provided it doesn't cause significant complications (currently 4)".

You could also add explicitly: "users that require support for older kernels than that can often simply use older versions of bpftrace".

jordalgo: I took the last suggestion but I like that we have some kind of concrete baseline (even if we have to adjust it in the future).

PR #4276, opened 2025-06-17
codegen: add USDT arguments pass (proof-of-concept)

amscanne:
Stacked PRs:
 * __->__#4276

--- --- ---

### codegen: add USDT arguments pass (proof-of-concept)

Signed-off-by: Adin Scannell <amscanne@meta.com>

PR #4331, opened 2025-07-01
codegen: generate `vmlinux.h` automatically for BPF C code

amscanne:
Stacked PRs:
 * #4276
 * #4332
 * __->__#4331

--- --- ---

### codegen: generate `vmlinux.h` automatically for BPF C code

Before building external sources, a single `vmlinux.h` file is generated
from the current BTF data using the `c_def` method. Note that building
this set required changing the internal generation mechanism from a
`std::string` to a `std::stringstream`, and using a `std::string`
results in endless memory churn due to the simple concatentation (it's
effectively `O(n^2)` in the length of the generated header, which is
quite substantial). Using a `std::stringstream` means that the internal
buffer is managed in a more sensible way (still taking a bit of time,
but much less noticeable).

If there are no external sources imported, then this generation does not
occur, in order to ensure that there are no regressions.

### DIFF
--- a/src/ast/passes/clang_build.cpp
+++ b/src/ast/passes/clang_build.cpp
@@ -22,6 +22,7 @@
 #include "ast/passes/clang_build.h"
 #include "ast/passes/codegen_llvm.h"
 #include "ast/passes/resolve_imports.h"
+#include "bpftrace.h"
 #include "stdlib/stdlib.h"
 #include "util/result.h"
 
@@ -113,6 +114,7 @@ Result<PipeFds> create_pipe()
 static Result<> build(CompileContext &ctx,
                       const std::string &name,
                       LoadedObject &obj,
+                      const llvm::MemoryBufferRef &vmlinux_h,
                       Imports &imports,
                       BitcodeModules &result)
 {
@@ -129,6 +131,7 @@ static Result<> build(CompileContext &ctx,
                  0,
                  llvm::MemoryBuffer::getMemBufferCopy(other.data(), name));
   }
+  vfs->addFileNoOwn("include/vmlinux.h", 0, vmlinux_h);
 
   // Create the diagnostic options and client. We emit the error to
   // a string, which we can then capture and associate with the import.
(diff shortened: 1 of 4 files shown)


jordalgo (commented):
We've discussed offline but mind adding a little more to the commit msg as to why this is necessary.

src/ast/passes/clang_build.cpp
```diff
@@ -217,22 +220,36 @@ static Result<> build(CompileContext &ctx,
...
+      "ClangBuilder",
+      [](BPFtrace &bpftrace,
+         CompileContext &ctx,
+         ast::Imports &imports) -> Result<BitcodeModules> {
+        BitcodeModules result;
+
+        // Construct our kernel headers.
+        std::string vmlinux_h;
+        if (!imports.c_sources.empty()) {
+          vmlinux_h = bpftrace.btf_->c_def();
```
jordalgo: Looking at this confused me and I think it has to do with the `c_def` function which returns the vmlinux btf if no set is provided. How about changing that a bit where we expose a method e.g. `vmlinux_defs()` that you can call here and that `c_def` can call as well?

src/btf.cpp
```diff
@@ -181,21 +182,21 @@ void BTF::load_module_btfs(const std::set<std::string> &modules)
 
 static void dump_printf(void *ctx, const char *fmt, va_list args)
 {
-  auto *ret = static_cast<std::string *>(ctx);
+  auto *ret = static_cast<std::stringstream *>(ctx);
```
jordalgo, line 185: nit: These minor stream changes feel like they belong in their own commit.

src/ast/passes/clang_build.cpp
amscanne: Everything is returning the `vmlinux` defs, it's just a question of which types are generated. I've trying cleaning up the methods and added some much needed exposition for a commit messaged. PTAL and see if it is more reasonable now.

src/ast/passes/clang_parser.cpp
```diff
@@ -796,7 +796,12 @@ void ClangParser::resolve_unknown_typedefs_from_btf(BPFtrace &bpftrace)
 
 CXUnsavedFile ClangParser::get_btf_generated_header(BPFtrace &bpftrace)
 {
-  btf_cdef = bpftrace.btf_->c_def(bpftrace.btf_set_);
+  // Note that `c_def` will provide the full set of types if an empty set is
+  // used here. We only want the types in `btf_set_` or nothing at all, so
+  // don't generate anything if this set is empty.
+  if (!bpftrace.btf_set_.empty()) {
+    btf_cdef = bpftrace.btf_->c_def(bpftrace.btf_set_);
```
viktormalik, line 803: If full vmlinux.h is generated in ClangBuildPass, how about we reuse it here to save some time?

src/ast/passes/clang_build.cpp
jordalgo: > Everything is returning the vmlinux defs

It seems like the current one will only return the defs in `btf_set_` if they exist, which could be just module BTF right instead of the whole vmlinux BTF - or am I misunderstanding?

Either way, the commit message is much improved and the comments help. Thanks!

src/ast/passes/clang_parser.cpp
amscanne: Ah, I believe because we can't currently. The reason for the whole "iterative" header building is that you want these defs to be the minimal satisfying set, because there may be conflicts with user headers and other things. I think there are opportunities to improve this, but I'd like to separate this from the critical path for USDTs.

src/ast/passes/clang_build.cpp
amscanne: Candidly, the`BTF` object is a bit of a mess and the way the source there is selected makes no sense to me. I believe that the module is likely a rebased on top of the full vmlinux, I'm not actually sure what will get dumped.

I'm trying to avoid going in and fixing too much there because it will quickly become yak shaving. My attempt here is to do the bare minimum to make it work for now, but I will test some cases with modules and see if it falls over.

jordalgo: No worries, this conversation is non-blocking :)

src/ast/passes/clang_parser.cpp
viktormalik: Ah, good point on the potential conflicts. Makes sense.

viktormalik (approved):

src/ast/passes/clang_build.cpp
amscanne: There's going to be a lot of churn in the types, so I'm just going to submit this one to unblock the USDT work at least

src/btf.cpp
amscanne: I'd ordinarily split this out, I just have too much building up here. The commit is pretty trivial (and this is needed to make it not dog-slow when generating `vmlinux.h`), so I'm going to move ahead. There's going to be a lot of churn in BTF anyways.

PR #4332, opened 2025-07-01
imports: refactor to allow direct import calls in later passes

amscanne:
Stacked PRs:
 * #4276
 * __->__#4332

--- --- ---

### imports: refactor to allow direct import calls in later passes

This change is primarily a small refactoring of the `Imports` object and
the associated `ResolveImports` visitor. It makes the `import_any`
method available on the imports object so that later passes can do
direct imports without needing to modify the AST. If the import has
already happened, then it will still be correctly memoized.

Note that there is some minor header shuffling, primarily due to the
lack of consistent IWYU.

### DIFF
--- a/src/ast/passes/resolve_imports.cpp
+++ b/src/ast/passes/resolve_imports.cpp
@@ -1,11 +1,9 @@
 #include <filesystem>
 #include <fstream>
 #include <sstream>
-#include <unordered_set>
 
 #include "ast/passes/resolve_imports.h"
 #include "ast/visitor.h"
-#include "bpftrace.h"
 #include "driver.h"
 #include "stdlib/stdlib.h"
 #include "util/result.h"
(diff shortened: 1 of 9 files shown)


src/ast/passes/resolve_imports.cpp
```diff
@@ -96,44 +49,26 @@ bool ResolveImports::checkPerms(const std::filesystem::path &path)
...
 
-Result<OK> ResolveImports::importScript(Node &node,
-                                        const std::string &name,
-                                        const std::filesystem::path &path)
+static Result<OK> importScript([[maybe_unused]] Node &node,
+                               Imports &imports,
+                               const std::string &name,
+                               const std::string &&contents,
+                               const std::vector<std::filesystem::path> &paths,
+                               std::map<std::string, ASTContext> &where)
```
jordalgo: Don't love this variable name... why not call it `scripts` ?

src/ast/passes/resolve_imports.cpp
```diff
@@ -275,29 +246,26 @@ Result<OK> ResolveImports::importAny(Node &node, const std::string &name)
...
-                          << path;
-        continue;
-      }
+  for (const auto &import_path : paths) {
+    // Check to see if this specific file exists.
+    auto path = import_path / name;
+    checked.emplace_back(path.string());
+    std::error_code ec;
+    if (!std::filesystem::exists(path, ec)) {
+      continue; // No file found.
```
jordalgo, line 252: I know this wasn't changed in this PR but should we do `node.addWarning` here?

src/driver.h
```diff
@@ -45,6 +45,6 @@ class Driver {
   void parse(Parser::symbol_type first_token);
 };
 
-ast::Pass CreateParsePass(bool debug = false);
+ast::Pass CreateParsePass(bool debug = false, uint32_t max_ast_nodes = 0);
```
jordalgo: nit: I feel like these two params should be reversed

src/ast/passes/resolve_imports.cpp
```diff
@@ -340,15 +308,15 @@ Result<OK> ResolveImports::importAny(Node &node, const std::string &name)
 void ResolveImports::visit(Import &imp)
 {
   // Ensure that the explicit statement is allowed.
-  if (bpftrace_.config_->unstable_import == ConfigUnstable::error) {
+  if (enabled_ == ConfigUnstable::error) {
```
jordalgo: Do we need to do this check here rather than just in `UnstableFeature` class?

amscanne: Let me check, and I'll remove this if it's not needed. That would be great.

src/driver.h
amscanne: I had the same feeling when I added it but ignored it :) I'll swap em

src/ast/passes/resolve_imports.cpp
amscanne: It's an error if the path can't be resolved, and we list all the paths we checked in that case (see `checked` above). I don't think you want a warning for every search path checked.

src/ast/passes/resolve_imports.cpp
amscanne: I'll just push back gently -- this is the map where the contents are placed. I could call it `out`? I would just like to keep it consistent with the other names for the other cases, and `scripts` is not suitable for those ones.

jordalgo: What about `cache` or `contents` ?

src/ast/passes/resolve_imports.cpp
jordalgo: Ah I see. Nevermind.

src/ast/passes/resolve_imports.cpp
amscanne: I'm fine with `contents`, I'll make them all that. (It's not really a `cache`)

src/ast/passes/resolve_imports.cpp
amscanne: Dropped.

src/driver.h
amscanne: Swapped.

PR #4338, opened 2025-07-03
btf: add library to support basic BTF types

amscanne:
Stacked PRs:
 * #4340
 * #4339
 * __->__#4338

--- --- ---

### btf: add library to support basic BTF types

This library is intended to be a thin wrapper around `libbpf`, and is
not entangled with other parts of bpftrace. It is being used to load BTF
from the compiled C modules, and match against the called types.


src/btf/compat.h
```diff
@@ -0,0 +1,63 @@
...
+Result<SizedType> getCompatType(const Pointer &type);
+Result<SizedType> getCompatType(const Array &type);
+Result<SizedType> getCompatType(const Struct &type);
+Result<SizedType> getCompatType(const Enum &type);
+Result<SizedType> getCompatType(const Enum64 &type);
+Result<SizedType> getCompatType(const TypeTag &type);
+Result<SizedType> getCompatType(const Typedef &type);
+
+template <typename T>
+Result<SizedType> getCompatType(const T &type)
```
jordalgo, line 57: This is neat. So any BTF types we don't have a `getCompatType` for will return the `CompatTypeError` ?

src/btf/compat.cpp
```diff
@@ -0,0 +1,116 @@
...
+Result<SizedType> getCompatType(const Struct &type)
+{
+  std::vector<SizedType> fields;
+  std::vector<std::string_view> idents;
+  auto f = type.fields();
+  if (!f) {
+    return f.takeError();
+  }
+  // We have no ability to control how things are packed,
+  // so this may or may not generate the correct type.
```
jordalgo, line 58: oof 😢

src/btf/compat.cpp
```diff
@@ -0,0 +1,116 @@
...
+  return CreateEnum(8 * (*size), type.name());
+}
+
+Result<SizedType> getCompatType(const Enum64 &type)
+{
+  auto size = type.size();
+  if (!size) {
+    return size.takeError();
+  }
+  return CreateEnum(8 * (*size), type.name());
```
jordalgo, line 85: Do we need to get the size here? Don't we know it's a 64 bit type?

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+  std::string content = buffer.str();
+  auto btf = Types::parse(content.c_str(), content.size());
+  ASSERT_TRUE(bool(btf));
+
+  // Want to see all the functions available?
+  std::unordered_set<std::string> funcs;
+  for (const Function &func : *btf | filter<Function>()) {
+    EXPECT_FALSE(funcs.contains(func.name()));
+    funcs.insert(func.name());
+  }
```
jordalgo, line 39: Add an ASSERT that `funcs` (and `vars` below) is not empty

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+{
+  Types btf;
+  auto int8 = btf.add<Integer>("int8", 1, 1);
+  ASSERT_TRUE(bool(int8));
+  EXPECT_EQ(to_str(AnyType(*int8)), "int8");
+}
+
+TEST(btf, kernel_types)
+{
+  std::ifstream file("/sys/kernel/btf/vmlinux");
```
jordalgo: Is this file always available everywhere this test might get run? What about just vendoring a smaller btf file for the test?

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+    vars.insert(var.name());
+  }
+}
+
+TEST(btf, map_types)
+{
+  Types btf;
+  auto int8 = btf.add<Integer>("int8", 1, 1);
+  ASSERT_TRUE(bool(int8));
+  EXPECT_EQ(to_str(AnyType(*int8)), "int8");
```
jordalgo: These two lines are already in `to_str` test above; no need to repeat.

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+  auto m = createMap(btf,
+                     MapInfo{
+                         .map_type = BPF_MAP_TYPE_HASH,
+                         .key = *int8,
+                         .value = *int8,
+                         .nr_elements = 1,
+                     });
+  ASSERT_TRUE(bool(m));
+  EXPECT_EQ(to_str(AnyType(*m)), "struct (anon)");
```
jordalgo: We can't detect it's a map and print something nicer?

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+  auto ptr_int32 = btf.add<Pointer>(AnyType(*int32));
+  ASSERT_TRUE(bool(ptr_int32));
+
+  auto element = ptr_int32->element_type();
+  ASSERT_TRUE(bool(element));
+  EXPECT_TRUE(element->is<Integer>());
+  EXPECT_EQ(to_str(*element), "int32");
+
+  auto ptr_ptr_int32 = btf.add<Pointer>(AnyType(*ptr_int32));
+  ASSERT_TRUE(bool(ptr_ptr_int32));
```
jordalgo, line 113:
```suggestion
  ASSERT_TRUE(bool(ptr_ptr_int32));
  EXPECT_TRUE(ptr_ptr_int32->is<Pointer>());
```

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+
+  auto looked_up = btf.lookup<Integer>("int32");
+  ASSERT_TRUE(bool(looked_up));
+  EXPECT_EQ(looked_up->type_id(), int32->type_id());
+
+  auto looked_up_by_id = btf.lookup<Integer>(int32->type_id());
+  ASSERT_TRUE(bool(looked_up_by_id));
+  EXPECT_EQ(looked_up_by_id->type_id(), int32->type_id());
+
+  auto any_looked_up = btf.lookup("int32");
```
jordalgo, line 382: Add a failure case were the lookup returns null?

tests/btf.cpp
```diff
@@ -0,0 +1,529 @@
...
+  auto int32 = btf.add<Integer>("int32", 4, 1);
+  auto int64 = btf.add<Integer>("int64", 8, 1);
+  auto ptr_int32 = btf.add<Pointer>(AnyType(*int32));
+
+  ASSERT_TRUE(bool(int32));
+  ASSERT_TRUE(bool(int64));
+  ASSERT_TRUE(bool(ptr_int32));
+
+  std::vector<std::string> integer_names;
+  for (const Integer &integer : btf | filter<Integer>()) {
```
jordalgo, line 426: a pipe syntax in cpp? Now i've seen everything

src/btf/irbuilder.cpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+    fields.push_back(*ft);
+  }
+  return llvm::StructType::create(
+      base.getContext(), fields, type.name(), packed);
+}
+
+Result<llvm::Type *> getType(llvm::IRBuilderBase &base, const Union &type)
+{
+  // There no unions in LLVM IR. Therefore, we need to construct a
```
jordalgo:
```suggestion
  // There are no unions in LLVM IR. Therefore, we need to construct a
```

src/btf/irbuilder.cpp
```diff
@@ -0,0 +1,181 @@
...
+  }
+  auto rtt = getType(base, rt);
+  if (!rtt) {
+    return rtt.takeError();
+  }
+  auto args = type.argument_types();
+  if (!args) {
+    return args.takeError();
+  }
+  bool varargs = false;
```
jordalgo:
```suggestion
  bool var_args = false;
```

src/btf/compat.cpp
amscanne: The packing of `SizedType` is an implementation detail, although it *should* align with the LLVM layout. So they *should* be the same. The packing of the BTF type is part of the representation, but we have no ability to pass that through when creating the `SizedType` here.

(Obviously removing the load-bearing "should"s there is part of the motivation for eventually aligning on a single consistent type representation with BTF)

src/btf/compat.cpp
amscanne: Maybe? I don't think it's any less correct to just get the size. The enum64 is defined as capable of holding 64-bit values and it should always return 8 here, but I'd rather just rely on that underlying BTF call then hard-code something and outsmart it :)

src/btf/compat.h
amscanne: Yup, and variant types will be automatically unpacked.

Actually, I can extend this to whatever... vectors of types, etc. I think I should do this?

Then we'd have less boilerplate, and you could just do:

```
auto args = getCompatType(func.args());
...
```

WDYT?

src/btf/irbuilder.cpp
amscanne: Why bad? You no hear English prime? ;)

(I'll fix)

tests/btf.cpp
amscanne: Will do.

tests/btf.cpp
amscanne: Will do.

tests/btf.cpp
amscanne: Will add.

tests/btf.cpp
amscanne: Will drop.

tests/btf.cpp
amscanne: There's another test for `mapInfo` which can extract this from the type. The default `operator<<` will print the C literal type, but you're right -- I can adjust this and do something nicer.

tests/btf.cpp
amscanne: This is how all the [C++20 ranges bits](https://en.cppreference.com/w/cpp/ranges.html#Range_adaptors) are meant to look. We're just. not using it very much.

I can drop this and do a custom type, I just figured I'd take a stab at implementing the iterator interfaces and make it "standard". I don't have strong feelings either way.

amscanne (commented):
Thanks for the review! I will update in the morning. It would be great to pipeline the ones above it, I'm happy to iterate on this in parallel once there's a clearer model of the use, etc.

viktormalik (commented):
This is a LOT of code and I'm not able to do a detailed review in a reasonable time so I'll rely on @jordalgo here.

But from high-level perspective, this looks reasonable (although a bit too complex but that's probably just due to my limited understanding of things).

I guess my main question is - do we want this infrastructure to eventually replace *all* BTF parsing within bpftrace, i.e. especially the current `BTF` class in src/btf.cpp?

src/btf/compat.h
jordalgo: > vectors of types

Yeah, makes sense. We can also add it later if you feel like it'll be more churn in this large PR.

src/btf/compat.cpp
jordalgo: > I don't think it's any less correct to just get the size

Sure. I would just lose my mind if the size was ever something other than 8 for enum64

tests/btf.cpp
jordalgo: Nah, leave it. I like learning and doing the idiomatic c++ things if we can.

src/btf/btf.h
```diff
@@ -0,0 +1,879 @@
...
+  Handle &operator=(Handle &other) = delete;
+
+  struct btf *btf_library() const
+  {
+    return btf_;
+  }
+
+  void rebase(HandleRef base)
+  {
+    base_ = std::move(base);
```
jordalgo, line 49: I see `base_` getting assigned to but not being used. Am I missing something?

src/btf/btf.h
```diff
@@ -0,0 +1,879 @@
...
+  UnknownType(const std::string &name) : name_(name) {};
+  UnknownType(uint32_t type_id) : name_(type_id) {};
+  void log(llvm::raw_ostream &OS) const override;
+
+private:
+  std::variant<uint32_t, std::string> name_;
+};
+
+// Indicates that a type was found, but it did not match the expected type.
+class KindMismatch : public ErrorInfo<KindMismatch> {
```
jordalgo, line 101: Just for conversation. It feel like there is a lot of boilerplate for these ErrorInfos we have scattered around the codebase now. Think it's worth creating something more generic or having some kinda factory?

src/btf/btf.h
```diff
@@ -0,0 +1,879 @@
...
+  using BaseType::BaseType;
+
+  // Used by the variant above, to determine if this accepts the given kind.
+  static bool is(uint32_t kind)
+  {
+    return kind == Kind;
+  }
+
+  // Finds the type with the required kind and name, constructing a new object.
+  // Note that this returns a const version of this object, preventing any
```
jordalgo: Does the const come with the `T`?

amscanne:
> This is a LOT of code and I'm not able to do a detailed review in a reasonable time so I'll rely on @jordalgo here.

There's no extreme urgency, so I'm happy to wait. That said, the code is also mostly boilerplate (simple wrappers around the libbpf handles).

> But from high-level perspective, this looks reasonable (although a bit too complex but that's probably just due to my limited understanding of things).
> 
> I guess my main question is - do we want this infrastructure to eventually replace _all_ BTF parsing within bpftrace, i.e. especially the current `BTF` class in src/btf.cpp?

I think how we evolve the type systems internally is still hazy, but yes, eventually. There are a couple problems with the existing `btf.cpp`:

* It represents a single namespace, regardless of whether you use multiple `BTF` instances. This is *mostly* okay with just vmlinux and a module, say, but isn't strictly correct (because the module can have types that are conflicting with internal vmlinux types, and you can even have conflicting structs and other types within vmlinux itself... this all gets much worse and hairier with many modules and interop code). You could extract this but the code is all tied in to `bpftrace_->structs` (a global namespace of structs) and it's effectively a rewrite anyways.
* It doesn't really understand BTF types. The ability to parse the actual BTF type is pretty limited, and it tends to trip over the namespace issue above. As an example see [get_stype](https://github.com/bpftrace/bpftrace/blob/master/src/btf.cpp#L434) -- it resolves structs purely by name, and turns all anonymous structs into `none`. The existing `btf.cpp` is essentially doing one thing: it takes a set of type names, and then **generates C definitions for those types**, which are subsequently parsed by `ClangParser`, which is the actual current source of type definitions (also defined in the single struct manager namespace).

That said, the existing thing does **mostly work** for the current cases, and I don't want to mess with that. So this is why I opted to build a separate BTF system that doesn't have the above problems: it supports multiple BTF type systems (mapped to the object itself), and fully understands the types. I will use this for the external object case, and we can figure out how to evolve the type system. At some point in the future, yes, I would like to remove `btf.cpp`, but that will likely come after a lot of other work to lift `SizedType`.

I can understand how the current thing came into being (since BTF was not everywhere, it was an adding as an adapter to the existing type system), but now that BTF is everywhere we can also depend on that type model more directly. Happy to talk through all these things, and none of this is a done deal.

src/btf/compat.h
amscanne: I added it but didn't like it. The code constructing that vector is also doing type checking in the loop anyways, so I feel like it just added complexity.

src/btf/btf.h
amscanne: It is a `shared_ptr`, so this is ensuring that the `btf_library` object is not being freed while this handle is in use.

src/btf/btf.h
amscanne: I agree we should have a generic `errno` one, but these errors are all different and emit different messages.

It probably just feels like a lot because they didn't exist, and we've had to add them in many cases. It will slow down and converge to a reasonable set. Having a few different errors for a library of this size seems reasonable to me.

src/btf/btf.h
amscanne: Hm, I may have changed this but I can't remember why. I will take a look and fix this or the comment.

src/btf/btf.h
jordalgo: Ahhh right. Maybe a small comment in the header file would be good but non-blocking.

tests/btf.cpp
amscanne: This is already true. `add<Pointer>` will return a `Pointer`.

src/btf/irbuilder.cpp
amscanne: Done.

src/btf/btf.h
amscanne: Done.

viktormalik:
> > This is a LOT of code and I'm not able to do a detailed review in a reasonable time so I'll rely on @jordalgo here.
> 
> There's no extreme urgency, so I'm happy to wait. That said, the code is also mostly boilerplate (simple wrappers around the libbpf handles).

Yeah, most of it is boilerplate so no problem with merging this straight away. Also, it looks very reasonable and if we manage to use it for our entire type system one day (see below), it will be great.

> > I guess my main question is - do we want this infrastructure to eventually replace _all_ BTF parsing within bpftrace, i.e. especially the current `BTF` class in src/btf.cpp?
> 
> I think how we evolve the type systems internally is still hazy, but yes, eventually. There are a couple problems with the existing `btf.cpp`:
> 
> * It represents a single namespace, regardless of whether you use multiple `BTF` instances. This is _mostly_ okay with just vmlinux and a module, say, but isn't strictly correct (because the module can have types that are conflicting with internal vmlinux types, and you can even have conflicting structs and other types within vmlinux itself... this all gets much worse and hairier with many modules and interop code). You could extract this but the code is all tied in to `bpftrace_->structs` (a global namespace of structs) and it's effectively a rewrite anyways.
> * It doesn't really understand BTF types. The ability to parse the actual BTF type is pretty limited, and it tends to trip over the namespace issue above. As an example see [get_stype](https://github.com/bpftrace/bpftrace/blob/master/src/btf.cpp#L434) -- it resolves structs purely by name, and turns all anonymous structs into `none`. The existing `btf.cpp` is essentially doing one thing: it takes a set of type names, and then **generates C definitions for those types**, which are subsequently parsed by `ClangParser`, which is the actual current source of type definitions (also defined in the single struct manager namespace).

Just a note - AFAIK, `btf.cpp` also does BTF -> `SizedType` translation which is used by `FieldAnalyser` to resolve types of field accesses. If all accesses can be resolved in such a way, `ClangParser` is not even used, IIRC. But that doesn't make any of the above false, I just wanted to complete the picture.

> That said, the existing thing does **mostly work** for the current cases, and I don't want to mess with that. So this is why I opted to build a separate BTF system that doesn't have the above problems: it supports multiple BTF type systems (mapped to the object itself), and fully understands the types. I will use this for the external object case, and we can figure out how to evolve the type system. At some point in the future, yes, I would like to remove `btf.cpp`, but that will likely come after a lot of other work to lift `SizedType`.

Right, I think this is a good approach. Now thinking about it, it's actually nice to have a full independent type system which we will use for a while (and hopefully figure out most of its problems) and which we can then use to replace the current one.

> I can understand how the current thing came into being (since BTF was not everywhere, it was an adding as an adapter to the existing type system), but now that BTF is everywhere we can also depend on that type model more directly. Happy to talk through all these things, and none of this is a done deal.

Sure, I'll be happy to talk, too, but I think that I now have a better understanding of the motivation behind this.

amscanne:
> Just a note - AFAIK, `btf.cpp` also does BTF -> `SizedType` translation which is used by `FieldAnalyser` to resolve types of field accesses. If all accesses can be resolved in such a way, `ClangParser` is not even used, IIRC. But that doesn't make any of the above false, I just wanted to complete the picture.

Yeah, I believe that this is only done directly for a few builtins: `struct pt_regs`, `struct task_struct`, etc. (See the `visit(Builtin&)` call in `FieldAnalyser`, which is the only use of `get_stype` that I know of). I believe that most of the time the FieldAnalyzer is adding to the `btf_set`, which is iteratively used to generate the types from indirectly via `ClangParser` (BTF generates C, which is then parsed and turned into a SizedType via that route). I suppose that the `struct task_struct` is holding up quite a lot though (since any fields read would also have types defined in this way), so it's a good call out.

viktormalik:
> > Just a note - AFAIK, `btf.cpp` also does BTF -> `SizedType` translation which is used by `FieldAnalyser` to resolve types of field accesses. If all accesses can be resolved in such a way, `ClangParser` is not even used, IIRC. But that doesn't make any of the above false, I just wanted to complete the picture.
> 
> Yeah, I believe that this is only done directly for a few builtins: `struct pt_regs`, `struct task_struct`, etc. (See the `visit(Builtin&)` call in `FieldAnalyser`, which is the only use of `get_stype` that I know of). I believe that most of the time the FieldAnalyzer is adding to the `btf_set`, which is iteratively used to generate the types from indirectly via `ClangParser` (BTF generates C, which is then parsed and turned into a SizedType via that route). I suppose that the `struct task_struct` is holding up quite a lot though (since any fields read would also have types defined in this way), so it's a good call out.

Not really, we also call `BTF::resolve_args` (which calls `get_stype` internally) from `FieldAnalyser` when we encounter the `args` builtin on fentry/fexit probes. Similarly, we call `get_stype` from `FieldAnalyser::resolve_type` used for resolving type casts. So, IMHO, for most common cases of accessing kernel types (builtins, fentry/fexit args, typecasted kernel pointers such as kprobe args), we resolve everything in `FieldAnalyser` and don't add to `btf_set` at all (and, consequently, `ClangParser` is skipped).

amscanne:
> Not really, we also call `BTF::resolve_args` (which calls `get_stype` internally) from `FieldAnalyser` when we encounter the `args` builtin on fentry/fexit probes. Similarly, we call `get_stype` from `FieldAnalyser::resolve_type` used for resolving type casts. So, IMHO, for most common cases of accessing kernel types (builtins, fentry/fexit args, typecasted kernel pointers such as kprobe args), we resolve everything in `FieldAnalyser` and don't add to `btf_set` at all (and, consequently, `ClangParser` is skipped).

Interesting, I never see `ClangParser` skipped for the most trivial program. Perhaps it is not doing anything useful in the most common case.

viktormalik:
> > Not really, we also call `BTF::resolve_args` (which calls `get_stype` internally) from `FieldAnalyser` when we encounter the `args` builtin on fentry/fexit probes. Similarly, we call `get_stype` from `FieldAnalyser::resolve_type` used for resolving type casts. So, IMHO, for most common cases of accessing kernel types (builtins, fentry/fexit args, typecasted kernel pointers such as kprobe args), we resolve everything in `FieldAnalyser` and don't add to `btf_set` at all (and, consequently, `ClangParser` is skipped).
> 
> Interesting, I never see `ClangParser` skipped for the most trivial program. Perhaps it is not doing anything useful in the most common case.

I believe that's the case. It runs but it actually doesn't do anything (since `BPFtrace::btf_set_` is empty).

src/btf/btf.h
amscanne: I've fixed this. The objects can't really be modified anyways, and you can't return the `Result<const T>` because then you can't move it, etc. So I just dropped this sentence.

tests/btf.cpp
jordalgo: Seems like these ASSERTs are still missing

src/btf/btf.h
```diff
@@ -0,0 +1,884 @@
...
+  Handle(HandleRef base);
+  Handle(struct btf *btf) : btf_(btf) {};
+  Handle(struct btf *btf, HandleRef base)
+      : btf_(btf), base_(std::move(base)) {};
+  ~Handle();
+
+  static Result<HandleRef> parse(const void *data, size_t sz);
+  static Result<HandleRef> parse(HandleRef base, const void *data, size_t sz);
+
+  Handle(Handle &&other) = default;
```
jordalgo: By leaving the default move constructor and assignment, do we need to null out the `btf_` pointer it has to prevent doubling calling `btf__free(btf_);` (if that's an issue)  ?

amscanne: Yikes, these should all be deleted. It is not used, but the Handle is never intended to be moved, only accessed through `HandleRef`.

tests/btf.cpp
amscanne: Weird, not sure what happened. There are there now.

jordalgo (approved):
LGTM. Nice addition!

PR #4339, opened 2025-07-03
ast: add a `TypeSystem` pass which builds a BTF type system

amscanne:
Stacked PRs:
 * #4340
 * __->__#4339

--- --- ---

### ast: add a `TypeSystem` pass which builds a BTF type system

This includes only external types, but these can be reconciled for calls
(at least in one direction), later on. Since this will be consumed by
the semantic analyzer, it means that `ClangBuild` also becomes a
required pass in most cases.


src/ast/passes/type_system.h
```diff
@@ -0,0 +1,22 @@
...
+#include "btf/btf.h"
+
+namespace bpftrace::ast {
+
+// TypeMetadata contains metadata related to the set of external types that are
+// available to each probe. Note that this does not currently cover any of the
+// existing `SizedType` implementations.
+class TypeMetadata : public ast::State<"type-metadata"> {
+public:
+  std::unordered_map<Probe *, btf::Types> types;
```
viktormalik: Are we sure we are going to have separate type sets for different probes? Wouldn't just using a single `btf::Types` suffice for the moment (IIUC, that's the way it's implemented now)?

src/main.cpp
```diff
@@ -783,6 +788,7 @@ int main(int argc, char* argv[])
     // probe. The raw contents of the program are the initial search provided.
     ast = buildListProgram(is_search_a_type ? FULL_SEARCH : args.search);
     ast::CDefinitions no_c_defs; // No external C definitions may be used.
+    ast::TypeMetadata no_types;  // No exteranl types may be used.
```
viktormalik: Typo
```suggestion
    ast::TypeMetadata no_types;  // No external types may be used.
```

src/ast/passes/type_system.h
jordalgo: I think having a type set for different probes is to disambiguate names that might be the same in two different probes but have different types. Correct me if I'm wrong here @amscanne

src/main.cpp
```diff
@@ -920,8 +930,9 @@ int main(int argc, char* argv[])
       break;
   }
 
-  pm.add(ast::CreateLLVMInitPass());
-  pm.add(ast::CreateClangBuildPass());
+  if (bt_debug.contains(DebugStage::Types)) {
```
jordalgo, line 928: nit: I think the `CreateDumpTypesPass` could have been a separate commit but I won't block the stack.

jordalgo:
> Perhaps we should at least add a test for it?

+1

src/ast/passes/type_system.h
amscanne: Yes, I would like to move in this direction so type resolution happens per-probe. However, I have simplified at the moment to just `global` since we aren't loading anything probe-specific (yet). We will probably end up retaining `global` as the set of linked BPF global functions, etc. but may prefer to resolve to the probe-specific types (when those are introduced).

src/main.cpp
amscanne: Done.

src/main.cpp
amscanne: I wanted a way to debug it interactively. I've also added a test.

amscanne:
Test added.

PR #4340, opened 2025-07-03
btf: add external type support for semantic analysis

amscanne:
Stacked PRs:
 * __->__#4340

--- --- ---

### btf: add external type support for semantic analysis

With this metadata, there is no longer any need to manually specify
functions in the semantic analyzer for external cases. We can start to
peel away static helpers into standard library files and pull code out
of codegen.

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -443,6 +443,7 @@ private:
   }
 
   Value *ctx_;
+  llvm::DILocalScope *scope_ = nullptr;
   AttachPoint *current_attach_point_ = nullptr;
   std::string probefull_;
   std::string tracepoint_struct_;
@@ -2097,7 +2098,14 @@ ScopedExpr CodegenLLVM::visit(Call &call)
       arg_values.push_back(args.back().value());
     }
 
-    return ScopedExpr(b_.CreateCall(func, arg_values, call.func));
+    auto *inst = b_.CreateCall(func, arg_values, call.func);
+    return ScopedExpr(inst, [this, inst, &call] {
+      // We set the debug location on the call instructions only after the
+      // scoped expression is not longer used. Otherwise the instruction emitter
+      // seems to use this location for everything, which results in problems.
+      inst->setDebugLoc(
+          debug_.createDebugLocation(llvm_ctx_, scope_, call.loc));
+    });
   }
 }
 
@@ -3369,7 +3377,7 @@ void CodegenLLVM::generateProbe(Probe &probe,
       func_type, llvm::Function::ExternalLinkage, func_name, module_.get());
   func->setSection(util::get_section_name(func_name));
   func->addFnAttr(Attribute::NoUnwind);
-  debug_.createProbeDebugInfo(*func);
+  scope_ = debug_.createProbeDebugInfo(*func);
 
   BasicBlock *entry = BasicBlock::Create(module_->getContext(), "entry", func);
   b_.SetInsertPoint(entry);
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -14,6 +14,7 @@
 #include "ast/passes/semantic_analyser.h"
 #include "ast/passes/type_system.h"
 #include "ast/signal_bt.h"
+#include "btf/compat.h"
 #include "collect_nodes.h"
 #include "config.h"
 #include "log.h"
@@ -115,12 +116,14 @@ public:
                             BPFtrace &bpftrace,
                             CDefinitions &c_definitions,
                             MapMetadata &map_metadata,
+                            TypeMetadata &type_metadata,
                             bool has_child = true,
                             bool listing = false)
       : ctx_(ctx),
         bpftrace_(bpftrace),
         c_definitions_(c_definitions),
         map_metadata_(map_metadata),
+        type_metadata_(type_metadata),
         listing_(listing),
         has_child_(has_child)
   {
@@ -170,6 +173,7 @@ private:
   BPFtrace &bpftrace_;
   CDefinitions &c_definitions_;
   MapMetadata &map_metadata_;
+  TypeMetadata &type_metadata_;
   bool listing_;
 
   bool is_final_pass() const;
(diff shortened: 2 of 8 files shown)


amscanne:
@jordalgo This is the basic bridge from C interop into bpftrace. With this in place, you're able to write C functions and then just use them *without adding anything to semantic analyzer*. I expect that things will still have plenty of rough edges and require time to refine (and I'd like to figure out how to migrate from `SizedType` being the source of truth here to potentially BTF being the source of truth, so that things like type maps, etc. work as expected). BUT after this lands, in theory we could start taking existing helper functions and migrating them to C, or better: a macro wrapper + C implementation.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -1900,7 +1901,14 @@ ScopedExpr CodegenLLVM::visit(Call &call)
       arg_values.push_back(args.back().value());
     }
 
-    return ScopedExpr(b_.CreateCall(func, arg_values, call.func));
+    auto *inst = b_.CreateCall(func, arg_values, call.func);
+    return ScopedExpr(inst, [this, inst, &call] {
+      // We set the debug location on the call instructions only after the
+      // scoped expression is not longer used. Otherwise the instruction emitter
+      // seems to use this location for everything, which results in problems.
```
viktormalik, line 2105: Nice trick, I was also fighting this behavior in the past.

viktormalik (commented):
Could we add unit tests for the introduced semantic analyser errors?

amscanne:
> Could we add unit tests for the introduced semantic analyser errors?

Yeah, I will add some tests.

src/ast/passes/codegen_llvm.cpp
amscanne: Yeah, I believe the correct solution is to actually emit the correct debug location on every new node, that we could do with a generic override of the `visit` template for `CodegenLLVM`. E.g. something like:

```
template <typename T>
ScopedExpr visit(T &t)
{
    if (std::is_derived_from_v<Node, T>) {
       b_.SetDebugLocation(createDebugLocation(t.loc(), scope_));
    }
    return codegen(t);
}
```

(We'd also need to change the scope_stack_ to actually generate a new `DILocalScope` for every scope, and pass this in.) But my goal is to actually unlock interop, we can come and clean this up later.

amscanne:
A basic test has been added to the semantic analyzer suite as well.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1853,7 +1857,72 @@ If you're seeing errors, try clamping the string sizes. For example:
     }
     call.return_type = CreateUInt64();
   } else {
-    call.addError() << "Unknown function: '" << call.func << "'";
+    // Check here if this corresponds to an external functions. We convert the
```
jordalgo: nit
```suggestion
    // Check here if this corresponds to an external function. We convert the
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1853,7 +1857,72 @@ If you're seeing errors, try clamping the string sizes. For example:
...
+      }
+      auto compat_return_type = getCompatType(*return_type);
+      if (!compat_return_type) {
+        call.addError() << "Unable to convert return type: "
+                        << compat_return_type.takeError();
+        continue;
+      }
+      std::vector<SizedType> args;
+      for (size_t i = 0; i < argument_types->size(); i++) {
+        const auto &[name, type] = argument_types->at(i);
```
jordalgo:
```suggestion
        const auto &[_, type] = argument_types->at(i);
```

tests/runtime/interop
```diff
@@ -0,0 +1,3 @@
+NAME interop return types
+PROG config = { unstable_import=true; } import "stdlib/test"; BEGIN { print(__always_true()); exit(); }
```
jordalgo, line 2: It's happening!!!

tests/semantic_analyser.cpp
```diff
@@ -5420,4 +5420,46 @@ TEST_F(SemanticAnalyserTest, warning_for_discared_return_value)
...
+      "foo", btf::Function::Linkage::Global, *add_proto);
+  ASSERT_TRUE(bool(add_func));
+
+  // Test that calling this function works.
+  test("kprobe:f { foo((int32)1, (int64)2); }", Types{ types });
+  test("kprobe:f { print(foo((int32)1, (int64)2)); }", Types{ types });
+
+  // Test that calling with the wrong types fails.
+  test("kprobe:f { foo((int64)1, (int64)2); }", Types{ types }, Error{ R"(
+stdin:1:12-35: ERROR: Expected int32, got int64
```
jordalgo: Nit: I feel like this error message could be better e.g. "Function `foo` expects two arguments of type: int32, int32 and got int64, int64"

src/ast/passes/semantic_analyser.cpp
amscanne: I don't know how I feel about this. The analyzer will still complain about it, even if the variable is named `_`. It wants you to do `const auto &type = argument_types->at(i).second;` but I think this is ugly and less clear. I like the unpacking syntax since you mentally reason that it's a pair, and they are `[name, type]`. WDYT?

tests/runtime/interop
amscanne: Yup, that's a C call. No special type checking.

src/ast/passes/semantic_analyser.cpp
jordalgo: For me it's just about reading convention, if the variable is named `_` then I know it's not going to be used. But NBD either way.

amscanne: I'm with you, if the analyzer didn't complain anyways I would use `_`. I'll probably just use `name` in the updating error message anyways.

amscanne: I'm now using the name in the error message when it is available.

tests/semantic_analyser.cpp
amscanne: I've added both, you will get a message for the specific argument (including the name), and you will get an additional message (either as a hint or a separate error, depending if you just have the wrong number of arguments or one of them is wrong) with the full set of arguments required.

jordalgo: Looks great. Thanks for fixing

src/ast/passes/semantic_analyser.cpp
amscanne: Done

Issue #4114, opened 2025-05-12
[RFC] Move probe expansion into its own AST pass(es)


viktormalik:
Currently, code for probe expansion is scattered all over the place.

The detection of whether (and what kind of) probe expansion is needed is in two places:
- most of the logic is in `AttachpointParser` which checks for wildcarded attach-points and sets different expansion kinds based on the probe type,
- `SemanticAnalyser` sets full expansion when either the `probe` builtin is used or the `args` builtin is used in a `tracepoint` probe.

A somewhat similar situation is for the actual expansion:
- full expansion (i.e. probe copying) is done in codegen,
- multi/session expansion is done on the level of probe attachment but some accompanying code is in the `BPFtrace` class,
- we were also discussing a new kind of expansion for fentry/fexit probes (these are now fully expanded every time) in #3005.

Now that we have AST mutating infrastructure in place, it should be possible to unify all the expansion code in AST passes. I'd suggest two passes:
1. `ExpansionAnalyser` which would detect which kind of probe expansion is needed.
2. `ProbeExpander` which would perform especially full expansion by creating copies of the expanded probes.

Running these before `FieldAnalyser` would also allow us to resolve issues like #2676 or #3590 more easily.

viktormalik:
@amscanne you brought this up in the office hours, would you mind if I took it (unless you already started working on it)?

amscanne:
> [@amscanne](https://github.com/amscanne) you brought this up in the office hours, would you mind if I took it (unless you already started working on it)?

It would give me great joy to have you take this :) 

My one comment/feedback on the proposal is on the two pass suggestion. I think that historically, we tended to treat passes as implementing one specific visitor, but IMHO you should have the pass be a logical boundary and use whatever set of visitors internally as an implementation detail. So perhaps your two passes are just two separate visitors within a single logical pass? (You can generate intermediate state within the actual pass implementation, and have whatever number of visitors or other structure that you need.)

viktormalik:
> My one comment/feedback on the proposal is on the two pass suggestion. I think that historically, we tended to treat passes as implementing one specific visitor, but IMHO you should have the pass be a logical boundary and use whatever set of visitors internally as an implementation detail. So perhaps your two passes are just two separate visitors within a single logical pass? (You can generate intermediate state within the actual pass implementation, and have whatever number of visitors or other structure that you need.)

[...]

PR #4342, opened 2025-07-04
Move expansion analysis to a separate pass

viktormalik:
This is the first part of #4114.

Move all code related to analysing what kind of expansion is needed for individual probes into a new pass `ProbeExpansionPass`. So, far, the new pass has two visitors - one for analysing most of the expansions (code moved from `AttachPointParser` and `SemanticAnalyser`) and one for analysing session expansion since it depends on the information computed by the first visitor (code moved from `ProbeAnalyser`, which was removed).

Now, all the code for assigning into `ast::AttachPoint::expansion` is in the new pass. This also removes the `ast::Probe::need_expansion` field by setting full expansion on all the probe's attach points (which has the same semantics).

The ultimate goal is to get rid of `ast::AttachPoint::expansion`, too, but since it is used by the code performing the actual expansion, it is left there for now. My idea is that these two analysis visitors will return some internal state, which will be used by a follow-up visitor performing the actual expansion. But I believe that this refactoring makes sense on its own, so I'm posting it separately to simplify reviewing (it's already big enough).

See commit messages for more details.

##### Checklist

### DIFF
--- a/src/ast/passes/probe_expansion.cpp
+++ b/src/ast/passes/probe_expansion.cpp
@@ -10,8 +10,7 @@ using ast::ExpansionType;
 
 class ExpansionAnalyser : public Visitor<ExpansionAnalyser> {
 public:
-  ExpansionAnalyser(BPFtrace &bpftrace, bool listing)
-      : bpftrace_(bpftrace), listing_(listing)
+  ExpansionAnalyser(BPFtrace &bpftrace) : bpftrace_(bpftrace)
   {
   }
   void analyse(Program &program);
@@ -25,7 +24,6 @@ private:
   Probe *probe_ = nullptr;
 
   BPFtrace &bpftrace_;
-  bool listing_;
 };
 
 void ExpansionAnalyser::analyse(Program &program)
@@ -98,12 +96,8 @@ void ExpansionAnalyser::visit(AttachPoint &ap)
       break;
 
     case ProbeType::iter:
-      if (util::has_wildcard(ap.func)) {
-        if (listing_)
-          expansion = ExpansionType::FULL;
-        else
-          ap.addError() << "iter probe type does not support wildcards";
-      }
+      if (util::has_wildcard(ap.func))
+        expansion = ExpansionType::FULL;
 
     default:
       // No expansion support for the rest of the probe types
@@ -192,10 +186,10 @@ void SessionAnalyser::visit(Probe &probe)
   }
 }
 
-Pass CreateProbeExpansionPass(bool listing)
+Pass CreateProbeExpansionPass()
 {
-  auto fn = [listing](ASTContext &ast, BPFtrace &bpftrace) {
-    ExpansionAnalyser analyser(bpftrace, listing);
+  auto fn = [](ASTContext &ast, BPFtrace &bpftrace) {
+    ExpansionAnalyser analyser(bpftrace);
     analyser.analyse(*ast.root);
     SessionAnalyser sessions(ast, bpftrace);
     sessions.analyse();
--- a/src/ast/passes/probe_expansion.h
+++ b/src/ast/passes/probe_expansion.h
@@ -4,6 +4,6 @@
 
 namespace bpftrace::ast {
 
-Pass CreateProbeExpansionPass(bool listing = false);
+Pass CreateProbeExpansionPass();
 
 } // namespace bpftrace::ast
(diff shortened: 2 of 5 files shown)


src/ast/passes/probe_expansion.cpp
```diff
@@ -0,0 +1,143 @@
...
+        else
+          expansion = ExpansionType::FULL;
+      }
+      break;
+
+    case ProbeType::uprobe:
+    case ProbeType::uretprobe:
+      // As the C++ language supports function overload, a given function name
+      // (without parameters) could have multiple matches even when no
+      // wildcards are used.
```
amscanne, line 66: (This must have existed before, but it's worth asking)

Don't these things mean that we actually need FULL expansion? They are different functions with different arguments, not just different attachment sites.

src/ast/passes/probe_expansion.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+void ExpansionAnalyser::visit(Builtin &builtin)
+{
+  if (!probe_)
+    return;
+
+  if (builtin.ident == "args") {
+    for (auto *ap : probe_->attach_points) {
+      if (probetype(ap->provider) == ProbeType::tracepoint) {
```
amscanne, line 118: Per my comment above, is there a reason that this is true only for tracepoints? This is a kind of soft optimization, but in theory the same soft optimization would apply to every probe type that has well-typed args. I suppose I'm just confused by all the fragmentation.

(This is all pre-existing, and I suspect you intend to tackle this subsequently, as expansion is removed from codegen. This all makes sense, I'm just thinking out loud.)

src/ast/passes/probe_expansion.cpp
```diff
@@ -0,0 +1,143 @@
...
+      break;
+
+    case ProbeType::watchpoint:
+      if (util::has_wildcard(ap.func))
+        expansion = ExpansionType::FULL;
+      break;
+
+    case ProbeType::iter:
+      if (util::has_wildcard(ap.func)) {
+        if (listing_)
```
amscanne: Aside: this seems to be the only use of `listing_`. I'd love to remove that, if possible. (Getting rid of it completely might need a bit of an overhaul, but maybe we can remove it here.)

One possibility might be to define the behavior here: we could simply ignore wildcards on the iter probe and add as a warning. The listing code path doesn't care about warnings. Up to you.

viktormalik:
> This looks great. Very clear and nicely separated into logical commits.
> 
> QQ: Did you want to move or add some unit tests into a `probe_expansion` test ?

Great idea. Thinking about it, I believe it makes more sense to move the relevant tests from tests/bpftrace.cpp when the pass does the actual expansion. So, I'd postpone this to the follow-up PR.

jordalgo:
> I believe it makes more sense to move the relevant tests from tests/bpftrace.cpp when the pass does the actual expansion. So, I'd postpone this to the follow-up PR.

Sounds good to me.

jordalgo (approved):

src/ast/passes/probe_expansion.cpp
viktormalik: Depends whether there are some arguments accessed by name. For kprobes and uprobes, the standard way to access arguments is position-based (`argN` builtins) so we can use multi-expansion here as the probe always accesses the same register from the context.

For when `args` is used, see below.

src/ast/passes/probe_expansion.cpp
viktormalik: Correct, we may want to use this for uprobes, too, since we allow accessing arguments by name (using `args`) when DWARF is available. The problem is that adding uprobes to this condition would not help since the field analyser only allows attach points with exactly the same prototype, even when full expansion is used. We actually have the same problem with fentries.

This is going to be resolved by moving the probe expansion prior to the field analyser so let me handle uprobes here in the follow-up PR.

src/ast/passes/probe_expansion.cpp
viktormalik: Oh but we don't want to ignore wildcards for iter probes when listing. We want to do full expansion so that we can do things like:
```
# bpftrace -l 'iter:*'
iter:bpf_link
iter:bpf_map
[...]
```
We could do the opposite though. Always mark full expansion here and fail later when trying to expand an iter probe in a non-listing mode. Let me give it a try.

viktormalik: Turns out it was pretty simple, I just moved the error message to semantic analyser. For simplicity, I added it as a new commit.

cc @jordalgo if he wants to have another look.

jordalgo: LGTM. 🚢

viktormalik:
@amscanne do you still want to clarify something here or can we merge? I've another follow-up refactoring PR prepared.

amscanne:
> @amscanne do you still want to clarify something here or can we merge? I've another follow-up refactoring PR prepared.

Nope, I was good from the approval :) My comments were mostly discussion-oriented/forward looking.

PR #4343, opened 2025-07-05
benchmark: Add `BENCH` probe type

jrife:
Introduce `BENCH`, a special probe type that allows you to benchmark a bpftrace code snippet. All `BENCH` probes are executed sequentially on startup in the order they are defined before all other probes types (even `BEGIN`) using `bpf_prog_test_run_opts`. Use `BPF_PROG_TYPE_XDP` instead of `BPF_PROG_TYPE_RAW_TRACEPOINT` for `BENCH` probes, since the latter [does not support the `.repeat` option](https://github.com/jrife/linux/blob/90b83efa6701656e02c86e7df2cb1765ea602d07/net/bpf/test_run.c#L769).

For bpftrace developers, `BENCH` makes it easy to understand the effects of performance optimizations, and may help bpftrace users as they try to understand the performance implications of certain probes that execute in the hot path of performance-sensitive code.

#### Background
Originally, I set out to try optimizing codegen for certain operations, but I needed a reliable way to assess the effects of my changes. I first considered writing some throwaway script to do this, but quickly realized a probe that let me benchmark arbitrary code snippets would be extremely handy. I guess I was loosely inspired by [Go's benchmark facilities](https://gobyexample.com/testing-and-benchmarking).
    
#### Example
    
```
// script.bt
BENCH:lhist {
  @a = lhist(5, 0, 10, 1);
}

BENCH:hist {
   @b = hist(rand);
}

BENCH:assign {
  @c = 0;
}

BENCH:count {
   @d = count();
}

BEGIN {
  exit();
}
```
    
```
jordan@t14:~/code/bpftrace$ sudo ./result/bin/bpftrace script.bt
[sudo] password for jordan:
Attached 5 probes

+-----------+-------------+
| BENCHMARK | NANOSECONDS |
+-----------+-------------+
| assign    | 33          |
| count     | 39          |
| hist      | 92          |
| lhist     | 40          |
+-----------+-------------+
```

#### Example: Optimizing `AssignMapStatement`
Do a map lookup first instead of unconditionally updating the map key to improve performance in the common case.

```
BEFORE                      AFTER
+-----------+-------------+ +-----------+-------------+
| BENCHMARK | NANOSECONDS | | BENCHMARK | NANOSECONDS |
+-----------+-------------+ +-----------+-------------+
| assign    | 153         | | assign    | 66          |
+-----------+-------------+ +-----------+-------------+
```

##### Before
```
jordan@t14:~/code/bpftrace$ sudo ./result/bin/bpftrace -e 'BENCH:assign { @c = 0; } BEGIN { exit(); }' -d verifier
...
BPF verifier log for BENCHMARK_assign_1:
--------------------------------------
func#0 @0
0: R1=ctx() R10=fp0
;
0: (b7) r6 = 0                        ; R6_w=0
1: (7b) *(u64 *)(r10 -40) = r6        ; R6_w=0 R10=fp0 fp-40_w=0
2: (7b) *(u64 *)(r10 -32) = r6        ; R6_w=0 R10=fp0 fp-32_w=0
3: (bf) r2 = r10                      ; R2_w=fp0 R10=fp0
4: (07) r2 += -40                     ; R2_w=fp-40
5: (bf) r3 = r10                      ; R3_w=fp0 R10=fp0
6: (07) r3 += -32                     ; R3_w=fp-32
7: (18) r1 = 0xffff8dc44e372000       ; R1_w=map_ptr(map=AT_c,ks=8,vs=8)
9: (b7) r4 = 0                        ; R4_w=0
10: (85) call bpf_map_update_elem#2   ; R0_w=scalar()
11: (bf) r1 = r0                      ; R0_w=scalar(id=1) R1_w=scalar(id=1)
12: (67) r1 <<= 32                    ; R1_w=scalar(smax=0x7fffffff00000000,umax=0xffffffff00000000,smin32=0,smax32=umax32=0,var_off=(0x0; 0xffffffff00000000))
13: (c7) r1 s>>= 32                   ; R1_w=scalar(smin=0xffffffff80000000,smax=0x7fffffff)
14: (65) if r1 s> 0xffffffff goto pc+17       ; R1_w=scalar(smin=0xffffffff80000000,smax=smax32=-1,umin=0xffffffff80000000,umin32=0x80000000,var_off=(0xffffffff80000000; 0x7fffffff))
15: (63) *(u32 *)(r10 -8) = r0        ; R0_w=scalar(id=1) R10=fp0 fp-8=????mmmm
16: (b7) r1 = 30006                   ; R1_w=30006
17: (7b) *(u64 *)(r10 -24) = r1       ; R1_w=30006 R10=fp0 fp-24_w=30006
18: (7b) *(u64 *)(r10 -16) = r6       ; R6_w=0 R10=fp0 fp-16_w=0
19: (bf) r2 = r10                     ; R2_w=fp0 R10=fp0
20: (07) r2 += -24                    ; R2_w=fp-24
21: (18) r1 = 0xffff8dc459523400      ; R1_w=map_ptr(map=ringbuf,ks=0,vs=0)
23: (b7) r3 = 20                      ; R3_w=20
24: (b7) r4 = 0                       ; R4_w=0
25: (85) call bpf_ringbuf_output#130
mark_precise: frame0: last_idx 25 first_idx 0 subseq_idx -1
mark_precise: frame0: regs=r3 stack= before 24: (b7) r4 = 0
mark_precise: frame0: regs=r3 stack= before 23: (b7) r3 = 20
26: R0=scalar()
26: (6d) if r6 s> r0 goto pc+1        ; R0=scalar(smin=0,umax=0x7fffffffffffffff,var_off=(0x0; 0x7fffffffffffffff)) R6=0
27: (05) goto pc+4
32: (b7) r0 = 0                       ; R0_w=0
33: (95) exit

...

Attached 2 probes

+-----------+-------------+
| BENCHMARK | NANOSECONDS |
+-----------+-------------+
| assign    | 153         |
+-----------+-------------+

@c: 0
```

##### After
```
jordan@t14:~/code/bpftrace$ sudo ./result/bin/bpftrace -e 'BENCH:assign { @c = 0; } BEGIN { exit(); }' -d verifier

...

BPF verifier log for BENCHMARK_assign_1:
--------------------------------------
func#0 @0
0: R1=ctx() R10=fp0
;
0: (b7) r6 = 0                        ; R6_w=0
1: (7b) *(u64 *)(r10 -40) = r6        ; R6_w=0 R10=fp0 fp-40_w=0
2: (7b) *(u64 *)(r10 -32) = r6        ; R6_w=0 R10=fp0 fp-32_w=0
3: (bf) r2 = r10                      ; R2_w=fp0 R10=fp0
4: (07) r2 += -40                     ; R2_w=fp-40
5: (18) r1 = 0xffff8dc44c4bdc00       ; R1_w=map_ptr(map=AT_c,ks=8,vs=8)
7: (85) call bpf_map_lookup_elem#1    ; R0_w=map_value_or_null(id=1,map=AT_c,ks=8,vs=8)
8: (15) if r0 == 0x0 goto pc+9        ; R0_w=map_value(map=AT_c,ks=8,vs=8)
9: (73) *(u8 *)(r0 +7) = r6           ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
10: (73) *(u8 *)(r0 +6) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
11: (73) *(u8 *)(r0 +5) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
12: (73) *(u8 *)(r0 +4) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
13: (73) *(u8 *)(r0 +3) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
14: (73) *(u8 *)(r0 +2) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
15: (73) *(u8 *)(r0 +1) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
16: (73) *(u8 *)(r0 +0) = r6          ; R0_w=map_value(map=AT_c,ks=8,vs=8) R6_w=0
17: (05) goto pc+12
30: (b7) r0 = 0                       ; R0_w=0
31: (95) exit

from 8 to 18: R0_w=0 R6_w=0 R10=fp0 fp-32_w=0 fp-40_w=mmmmmmmm
18: R0_w=0 R6_w=0 R10=fp0 fp-32_w=0 fp-40_w=mmmmmmmm
18: (bf) r2 = r10                     ; R2_w=fp0 R10=fp0
19: (07) r2 += -40                    ; R2_w=fp-40
20: (bf) r3 = r10                     ; R3_w=fp0 R10=fp0
21: (07) r3 += -32                    ; R3_w=fp-32
22: (18) r1 = 0xffff8dc44c4bdc00      ; R1_w=map_ptr(map=AT_c,ks=8,vs=8)
24: (b7) r4 = 0                       ; R4_w=0
25: (85) call bpf_map_update_elem#2   ; R0=scalar()
26: (bf) r1 = r0                      ; R0=scalar(id=2) R1_w=scalar(id=2)
27: (67) r1 <<= 32                    ; R1_w=scalar(smax=0x7fffffff00000000,umax=0xffffffff00000000,smin32=0,smax32=umax32=0,var_off=(0x0; 0xffffffff00000000))
28: (c7) r1 s>>= 32                   ; R1_w=scalar(smin=0xffffffff80000000,smax=0x7fffffff)
29: (6d) if r6 s> r1 goto pc+2        ; R1_w=scalar(smin=smin32=0,smax=umax=umax32=0x7fffffff,var_off=(0x0; 0x7fffffff)) R6=0
30: safe

...

Attached 2 probes

+-----------+-------------+
| BENCHMARK | NANOSECONDS |
+-----------+-------------+
| assign    | 66          |
+-----------+-------------+

@c: 0
```

##### Checklist


jrife:
@jordalgo This is still in the works, but I thought it was cool, so I wanted to share. At first I just wanted to try out some optimizations, but then `BENCHMARK` emerged as a tool to help with that. It's been quite handy as I test out some of the performance improvement ideas I had.

jordalgo:
Cool idea! I think this could be very useful for developers and users. The output table is also nice. Here are some initial thoughts on the implementation and syntax:

Usually a benchmark is something that is executed many many times. Is there a way, maybe utilizing the `profile` probe, to execute the code snippets across all cpus in some random order multiple times? TBH, I'm not sure if a single execution of a BPF code snippet is representative of the timing - maybe it is but there might be magical JIT stuff that happens and CPU instruction pipelining, etc. Also if we do execute multiple times should we include other stats like 'avg', 'p50', 'p99', 'min', 'max'?

What about just `BENCH` for short? We like to keep things terse here 😄 

Should we allow the `BENCH` probe interspersed with other probes? Maybe it should only exist on it's own.

amscanne:
> Usually a benchmark is something that is executed many many times. Is there a way, maybe utilizing the profile probe, to execute the code snippets across all cpus in some random order multiple times? TBH, I'm not sure if a single execution of a BPF code snippet is representative of the timing - maybe it is but there might be magical JIT stuff that happens and CPU instruction pipelining, etc. Also if we do execute multiple times should we include other stats like 'avg', 'p50', 'p99', 'min', 'max'?

I think @jordalgo is just looking for something that will help prove that the per-CPU variants have savings over global atomics ;)

My perspective on this is that the while something more sophisticated may be useful, this is actually achieving an amazing grail for **micro**-benchmarks. It is exploiting a kernel feature that reduces all this noise and synthesizes a single number. I *love* this as a driver for the simple cases you've outlined, i.e. testing very basic code generation. The fact that we don't have min/max/avg, etc. is great, and "we just say what the kernel tells us". (I did [something basic](https://github.com/bpftrace/bpftrace/blob/master/src/benchmark.cpp#L50) for the pass benchmark, but it's always a crapshoot and the statistical measures you end up making up are rarely useful. I'd rather not have that at all.)

We could build something for more sophisticated later for effectively testing full programs (basically testing the distributed system you're building, with all its side effects, etc.) but as a micro-benchmark for BPF programs, this is great IMO.

> What about just BENCH for short? We like to keep things terse here 😄

+1

> Should we allow the BENCH probe interspersed with other probes? Maybe it should only exist on it's own.

Thinking out loud: now that we have macros and other things, we may want to introduce a pure "test" probe type, e.g. `TEST:foo`, that just executes and could maybe cause assert failures and other things. This would basically be a wrapper around BPF `prog_run` without the repeats.

Both the `TEST` probes and `BENCH` probes would run only if you ran the script a certain way, e.g. `bpftrace --test script.bt` or whatever (and only those probes would run).

I wouldn't hold this change based on anything like that, but it's an interesting enhancement. The tools tests could essentially turn into `bpftrace --test tools/*.bt` and we could add additional assertions there, if we wanted to.

jrife:
> Usually a benchmark is something that is executed many many times. 
> TBH, I'm not sure if a single execution of a BPF code snippet is representative of the timing - maybe it is but there might be magical JIT stuff that happens and CPU instruction pipelining, etc

Currently, each `BENCHMARK` code snippet runs 1000000 times (the `repeat` parameter in `bpf_prog_test_run_opts`) and the average is reported back through `bpf_test_run_opts.duration`, so there are repeated trials of each benchmark. From observation, I can say that with this many trials the results are fairly consistent and repeatable. No micro benchmark is perfect, but the numbers seem to behave how I'd expect with the handful of optimizations I've tried.

>  Is there a way, maybe utilizing the profile probe, to execute the code snippets across all cpus in some random order multiple times?

Maybe. My feeling is that results would be fairly noisy and hard to interpret unless I'm misunderstanding something here. How would this look?

> Also if we do execute multiple times should we include other stats like 'avg', 'p50', 'p99', 'min', 'max'?

This would be nice, but is harder. I'm essentially just surfacing built-in functionality from `bpf_test_run_opts`, which only reports back the average. This might be a good kernel feature to add though which could then be surfaced through bpftrace where supported. We could try to simulate things by executing `bpf_test_run_opts` 1000000 times in a loop from userspace, but again, I think this would be subject to noise compared to the loop being closer to the kernel.

> What about just BENCH for short? We like to keep things terse here

Sure, I like the shorter name!

> Should we allow the BENCH probe interspersed with other probes? Maybe it should only exist on it's own.

I was split on this. In the end, it felt cleaner to just not add extra restrictions and rules around whether or not you can combine `BENCH` and other probes. It seems strange that someone would want to put `BENCH` and `kprobe` in the same script, but I guess there's no harm in doing so? What would be your concern in allowing this?

jordalgo:
> Currently, each BENCHMARK code snippet runs 1000000 times (the repeat parameter in bpf_prog_test_run_opts)

Ooohh. Totally didn't pick up on that. Very cool. Ignore my suggestion about using the `profile` probe.

> We could build something for more sophisticated later for effectively testing full programs

+1 

> would run only if you ran the script a certain way, e.g. `bpftrace --test script.bt`

Building off of this idea. What if we don't add any new probe types (`BENCH` or `TEST`) and just add two new command line params `--test` and `--bench` (we can start with the latter) where by we just run all the current probe programs in the script in test or benchmark mode? We can re-use the current probe names as the test and benchmark names. This saves the  user/developer the trouble of repeating logic in a separate probe that ultimately shouldn't run in a real usage of the script.

amscanne:
I've written up the `TEST`/`BENCH` proposal in #4347. @jrife not sure if you wanted to own this, but it's also a light lift from this. I think that it's a great general feature as we stabilize macros, etc.

jrife:
@amscanne Yeah seems interesting. I can take that on as well. I should have some time later in the week to dive deeper on this.

jrife:
> Building off of this idea. What if we don't add any new probe types (BENCH or TEST) and just add two new command line params --test and --bench (we can start with the latter) where by we just run all the current probe programs in the script in test or benchmark mode? We can re-use the current probe names as the test and benchmark names. This saves the user/developer the trouble of repeating logic in a separate probe that ultimately shouldn't run in a real usage of the script.

One problem I see with this is that, given an arbitrary program with an arbitrary set of probes, you don't really know what order they're meant to run in. For example, it may be expected that two `kprobe` probes are run one after the other in practice, and the code in a script may be written expecting this. The benefit of having dedicated probe types like `BENCH` and `TEST` is that there's a well-defined order they're meant to execute in, and the intent is clear. Otherwise, what's actually being tested/benchmarked would be somewhat ambiguous.

I suppose one way around this is to just document that "in benchmark/test mode probes are executed sequentially in the order they are defined" or something like that. WDYT? Although, now that I think of it, you may run into issues still given that `repeat` is only supported for certain attachment types.

jordalgo:
> One problem I see with this is that, given an arbitrary program with an arbitrary set of probes, you don't really know what order they're meant to run in.

I think probe order is rarely a concern in prod bpftrace scripts (unless we're talking about k/u probes and their 'ret' equivalents - or BEGIN/END probes) as they can fire at anytime depending on the event. Also, I don't see how probe order matters in a benchmark or in a test - the isolated program is what is being benchmarked or tested not the entirety of the script itself.

> Although, now that I think of it, you may run into issues still given that repeat is only supported for certain attachment types.

In 'bench' mode we wouldn't respect the attach types and just use the one you're using for `BENCH` now. Though I'm realizing this won't work for any probes that utilize outside args -- we can detect this and not run those programs but that's not as cool.

amscanne:
Right, the existing probes also depend on arguments and other context (registers, calling kernel functions that give you external state) that would not be available to the `TEST` or `BENCH` probe. There's no way around it: users need to do some work to factor their code to be testable. We've only now just provided some basic abilities to do that (e.g. with macros and imports), provide the ability to define generic test/benchmark entrypoints is just completing that story. I don't think there's a silver bullet that will automatically test the unmodified probe logic (and if we try, IMO we'll just be fighting a million little problems and the feature won't be useful).

jrife:
In addition to the dedicated `BENCH` and `TEST` probe types, some future work could add support for calculating runtime averages from BPF program stats (`sysctl kernel.bpf_stats_enabled=1`), which AFAICT would allow you to calculate at least an average from a count and total time spent metric. This would work well with any probe type, since you're measuring the runtime of programs running in the context they're meant to be run in. It might make a good compliment to `BENCH`, which would primarily be a microbenchmarking facility.

jordalgo:
Chatted with @amscanne offline and he made a compelling case for the `BENCH` probe (as opposed to my suggestion) so I'm good with this addition. 

I'm on the fence about requiring a `--bench` or `--test` flag to run the probes. It is simpler/cleaner to just run them by default as @jrife pointed out. So maybe we start with that. I wouldn't consider it a breaking change if we wanted to require it later as `BENCH` probes are specifically for bpftrace analysis.

jrife:
Agh, looks like `bpf_prog_test_run_opts` returns `-EINVAL` on the CI kernel, so the runtime tests fail. I'll need to poke around a bit to see what's going on there. Possibly a difference in kernel config or difference in input validations on newer kernel versions, etc. I may need to build my own kernel locally using the same config for some deeper inspection.

Edit: Note to me: Looks like [this check](https://github.com/jrife/linux/commit/6b3d638ca897e099fa99bd6d02189d3176f80a47) was added recently and is where the `-EINVAL` is coming from. I'll need to explicitly initialize `data_in` and `data_size_in` for the call to `bpf_prog_test_run_opts`.

jrife:
@jordalgo @amscanne I'm not sure where we landed on requiring the user to explicitly pass a `--bench` or `--test` parameter to run `BENCH` probes vs just running them by default if present, but it should be easy to adjust this. Anyway, marking this as ready to review. Feel free to leave some feedback when you get a chance.

CHANGELOG.md
```diff
@@ -69,6 +69,8 @@ and this project adheres to
   - [#4345](https://github.com/bpftrace/bpftrace/pull/4345)
 - Introduce `tseries` for capturing time series data
   - [#3838](https://github.com/bpftrace/bpftrace/pull/3838)
+- Add `BENCH` for creating microbenchmarks
```
jordalgo, line 74:
```suggestion
- Add `BENCH` probe for creating micro benchmarks
```

man/adoc/bpftrace.adoc
```diff
@@ -1202,6 +1202,11 @@ Most providers also support a short name which can be used instead of the full n
 | Built-in events
 | Kernel/User
 
+| <<probes-bench, `BENCH`>>
+| -
+| Microbenchmarks
```
jordalgo:
```suggestion
| Micro benchmarks
```

man/adoc/bpftrace.adoc
```diff
@@ -1285,6 +1290,37 @@ END {
...
+execution time of the contained code. If multiple `BENCH` probes exist
+in a script, bpftrace executes them sequentially in the order they are
+specified. `BENCH` probes execute before `BEGIN`.
+
+----
+BENCH:lhist {
+    @a = lhist(rand % 10, 1, 10, 1);
+}
+
+BENCH:count {
```
jordalgo, line 1294: Maybe it's good to add an example that doesn't use one of bpftrace's predefined names so users don't think it's a limitation e.g. just change `BENCH:count` to `BENCH:my_count` or something.

man/adoc/bpftrace.adoc
```diff
@@ -4221,6 +4257,17 @@ The possible options are:
 - `warn` - enable feature but print a warning
 - `enable` - enable feature
 
+==== unstable_testing
```
jordalgo: Curious to hear what @viktormalik and @amscanne think but I don't think we need to gate this feature. It's so benign and doesn't really pertain to actually debugging/investigations with bpftrace. If in the future we wanted to trash it, we could just issue a warning and not run them.

src/ast/attachpoint_parser.cpp
```diff
@@ -288,6 +290,22 @@ AttachPointParser::State AttachPointParser::special_parser()
   return OK;
 }
 
+AttachPointParser::State AttachPointParser::benchmark_parser()
```
jordalgo, line 293: hmm, maybe this should instead go in the `special_parser`

src/attached_probe.cpp
```diff
@@ -66,6 +67,7 @@ libbpf::bpf_prog_type progtype(ProbeType t)
   switch (t) {
       // clang-format off
     case ProbeType::special:    return libbpf::BPF_PROG_TYPE_RAW_TRACEPOINT; break;
+    case ProbeType::benchmark:  return libbpf::BPF_PROG_TYPE_XDP; break;
```
jordalgo, line 70: why is this probe an XDP type?

src/attached_probe.cpp
```diff
@@ -1629,7 +1631,8 @@ Result<std::unique_ptr<AttachedProbe>> AttachedProbe::make(
       return AttachedUprobeProbe::make(probe, prog, pid, safe_mode);
     }
     case ProbeType::invalid:
-    case ProbeType::special: {
+    case ProbeType::special:
+    case ProbeType::benchmark: {
```
jordalgo, line 1635: As mentioned above, I think this should qualify as a special probe.

src/output.cpp
```diff
@@ -1041,6 +1041,43 @@ void TextOutput::helper_error(int retcode, const HelperErrorInfo &info) const
...
 
+void TextOutput::benchmark_results(
+    const std::map<std::string, uint32_t> &results) const
+{
+  if (results.empty()) {
+    return;
+  }
+
+  const std::string BENCHMARK = "BENCHMARK";
+  const std::string NANOSECONDS = "NANOSECONDS";
```
jordalgo: should these be file level static consts?

tests/parser.cpp
```diff
@@ -1393,6 +1393,26 @@ END:path:f { 1 }
...
+       "  int: 1\n");
+
+  test_parse_failure("BENCH{ 1 }", R"(
+stdin:1:1-6: ERROR: BENCH probe type requires 1 arguments, found 0
+BENCH{ 1 }
+~~~~~
+)");
+
+  test_parse_failure("BENCH:a:f { 1 }", R"(
+stdin:1:1-10: ERROR: BENCH probe type requires 1 arguments, found 2
```
jordalgo, line 1420: nit: 
```suggestion
stdin:1:1-10: ERROR: BENCH probe type requires 1 argument, found 2
```

tests/runtime/json-output
```diff
@@ -190,3 +190,14 @@ NAME print bool type
 RUN {{BPFTRACE}} -q -f json -e 'BEGIN { print((true, false)); exit(); }'
 EXPECT {"type": "value", "data": [true,false]}
 TIMEOUT 1
+
+NAME BENCH
+RUN {{BPFTRACE}} -q -f json -e 'BENCH:a { @ = count(); } BEGIN { print(@); clear(@); exit(); }'
```
jordalgo: why do we need the `BEGIN` probe?

jordalgo (changes requested):
I was trying this out a bit and I think we should indeed only run the `BENCH` probes if we pass a flag `--bench` and these will be the ONLY ones we run. It's a footgun if your `BENCH` probe modifies a map that is used in the rest of your script e.g.
```
BENCH:hello { @a = count(); }
BEGIN { print(@a); }
```
Maybe we need some magic to auto-clear the maps after every bench run?

I also think we should swallow all the resulting output. Take this unfortunate example: `BENCH:a { print(1); }`. This almost borks my terminal. There is also this example: `BENCH:hello { @a[rand] = rand; }`, which issues a ton of warnings about the map being full and then prints a crazy long map at the end of execution. I don't know how you would selectively swallow the ring buffer output from a certain set of probes - unless you wrote to a separate buffer. This is much easier in a `--bench` mode.

A few more notes:
- we need to prevent against name collision e.g. `BENCH:hello { @a = count(); } BENCH:hello { @b = sum(1); }`
- I think we should implicitly exit in `bench` mode - users shouldn't need to write `exit()` in a BEGIN or END probe

CHANGELOG.md
jrife: I'm not sure about this. I googled this a bit beforehand trying to find an authoritative source on whether it's "micro benchmark" "micro-benchmark" or "microbenchmark" and all three seemed to be present in various places. "Microbenchmark" seemed the most common.

man/adoc/bpftrace.adoc
jrife: Sure, will do.

src/ast/attachpoint_parser.cpp
jrife: This is where I started but ended up separating them to avoid conditionals special-casing "BENCH" probes through the stack. The two key differences are parsing out the benchmark name and using `BPF_PROG_TYPE_XDP` instead of `BPF_PROG_TYPE_RAW_TRACEPOINT`.

jrife: This is where I started but ended up separating them to avoid conditionals special-casing "BENCH" probes through the stack. The two key differences are parsing out the benchmark name and using `BPF_PROG_TYPE_XDP` instead of `BPF_PROG_TYPE_RAW_TRACEPOINT`. Also, `special_probes` is currently an `unordered_map`. I wanted something ordered that would store the `BENCH` probes in the order they are defined. Changing the `special_probes` data structure to accomodate `BENCH` felt wrong.

src/attached_probe.cpp
jrife: Unfortunately, `BPF_PROG_TYPE_RAW_TRACEPOINT` and the other probe types here do not support the `.repeat` option [in the kernel](https://github.com/jrife/linux/blob/90b83efa6701656e02c86e7df2cb1765ea602d07/net/bpf/test_run.c#L769).  `BPF_PROG_TYPE_XDP` happens to be one type that does.

I know this seems gross, but I've seen Cilium abuse `BPF_PROG_TYPE_XDP` and `bpf_prog_test_run` to create BPF unit tests, so there's at least some precedent for using `BPF_PROG_TYPE_XDP` for purposes it wasn't intended for.

tests/runtime/json-output
jrife: This is just to assert that the `BENCH` probes run before `BEGIN`. Also, I needed somewhere to `exit()`.

tests/parser.cpp
jrife: Yeah, this comes from [here](https://github.com/bpftrace/bpftrace/blob/566a4d233f63adae07458f33dac0b9de13d1d6a9/src/ast/attachpoint_parser.cpp#L33) which just generates a generic message without a special case to differentiate between "argument" and "arguments". It's an easy fix. I can open a different PR to address this to keep this one focused on the `BENCH` changes if that works.

jrife:
> I was trying this out a bit and I think we should indeed only run the BENCH probes if we pass a flag --bench and these will be the ONLY ones we run. It's a footgun if your BENCH probe modifies a map that is used in the rest of your script e.g.

I mean it *is* working as intended in that case according to the wording in the docs, but people don't always read the docs, so sure. 

> Maybe we need some magic to auto-clear the maps after every bench run?

I think my intuition would be that values are preserved between the `BENCH` probes. I think keeping the behavior consistent with all the other probe types would result in fewer surprises. If someone wants to clear and reuse the same map between benchmark runs it is *slightly* annoying though. If you don't want to benchmark the `clear()` operation itself you'd need to add a "dummy benchmark" which would do the map clear 1000000 times in a row that runs before your next benchmark. Then again, you can just use different map names for each benchmark.

> I also think we should swallow all the resulting output. Take this unfortunate example: BENCH:a { print(1); }. This almost borks my terminal.

Yeah, this bit me the other day actually while writing the runtime tests. It seems simple enough to make an exception for `print()` `printf()` output to disable it.

> There is also this example: BENCH:hello { @a[rand] = rand; }, which issues a ton of warnings about the map being full and then prints a crazy long map at the end of execution. I don't know how you would selectively swallow the ring buffer output from a certain set of probes - unless you wrote to a separate buffer. This is much easier in a --bench mode.

Maybe this adds weight to the argument that we should clear maps automatically, at least at the very end. I don't think we should do anything about warnings though.

I guess where I'm torn is that there's not anything *that* special about a probe running 1000000 times in quick succession. There are plenty of events in the kernel that could happen just as frequently and if someone puts code in such a probe that will result in full maps, warnings, or lots of prints, that's on them. Putting a bunch of guardrails in seems like more trouble than it's worth considering the alternative is just letting people run a script, seeing that they messed up, and letting them fix it so it doesn't spew out a bunch of stuff.

> we need to prevent against name collision e.g. BENCH:hello { @a = count(); } BENCH:hello { @b = sum(1); }

Hmmmm, sure. At first I didn't really see a reason to block this, but I also don't see a legitimate use case for duplicate names. So, it probably is more helpful to call it out, since it's probably a mistake on a user's part.

> I think we should implicitly exit in bench mode - users shouldn't need to write exit() in a BEGIN or END probe

Yeah, this makes sense if we have `--bench` and aren't running any other probes.

tests/runtime/json-output
jrife: Note: after adding the `--bench` parameter and defaulting to exit after running the benchmarks, the `BEGIN` can be dropped.

amscanne: Shouldn't you still run `BEGIN` first? My assumption would be that `BEGIN` could still be used to set up some common state.

man/adoc/bpftrace.adoc
amscanne: +1 We could just not document it or include it in the CHANGELOG until it is relatively hashed out. (I'm not opposed to documenting in a lightweight form either, as long as we can hash it out relatively quickly which I feel is pretty tractable.)

amscanne:
> I also think we should swallow all the resulting output. Take this unfortunate example: BENCH:a { print(1); }. This almost borks my terminal.

> Yeah, this bit me the other day actually while writing the runtime tests. It seems simple enough to make an exception for print() printf() output to disable it.

What if we just add an option to [suppress output](https://github.com/bpftrace/bpftrace/issues/4347#issuecomment-3058646888). I don't think we should outsmart ourselves by having special codegen for benchmarks.

src/bpftrace.cpp
```diff
@@ -688,6 +691,43 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
 
   int num_special_attached = 0;
 
+  for (auto &probe : resources.benchmark_probes) {
+    auto &benchmark_prog = bytecode_.getProgramForProbe(probe);
+    const int trials = 1'000'000;
```
amscanne: I think this could be an optional property of the benchmark, e.g.

`BENCH:foo:10000000 { .... }`

(And you use the default if it is not provided.)

An alternative would be a command-line argument, but IMHO given the nature of the benchmarks it makes more sense to provide with the probe itself.

src/bpftrace.cpp
```diff
@@ -688,6 +691,43 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
...
+    // Note: on newer kernels you must provide a data_in buffer at least
+    // ETH_HLEN bytes long to make sure input validation works for
+    // opts. Otherwise, bpf_prog_test_run_opts will return -EINVAL for
+    // BPF_PROG_TYPE_XDP.
+    //
+    // https://github.com/torvalds/linux/commit/6b3d638ca897e099fa99bd6d02189d3176f80a47
+    constexpr size_t ETH_HLEN = 14;
+    char data_in[ETH_HLEN];
+    struct ::bpf_test_run_opts opts = {
```
amscanne: Nit: I'm not sure, but I believe that if you provide an empty initializer, it will be zeroed. Then you can set just the fields you care about, e.g.

```
struct ::bpf_test_run_opts opts = {};
opts.sz = sizeof(opts);
opts.data_in = data_in,
opts.data_size_in = sizeof(data_in);
opts.repeat = trials;
```

src/bpftrace.cpp
```diff
@@ -688,6 +691,43 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
...
+    if (auto ret = ::bpf_prog_test_run_opts(benchmark_prog.fd(), &opts)) {
+      LOG(ERROR) << "bpf_prog_test_run_opts failed: " << ret;
+      return -1;
+    }
+
+    benchmark_results[probe.path] = opts.duration;
+    ++num_special_attached;
+  }
+
   auto begin_probe = resources.special_probes.find("BEGIN");
```
amscanne: Per the comment elsewhere, my intuition is that `BEGIN` is still run first (and becomes like `SetUp` for the tests, effectively). Curious for others intuition on this.

src/output.cpp
amscanne: I'm pulling out some common bits here: #4356

Let's just have the time printing use the same standards either way

tests/runtime/probe
```diff
@@ -514,6 +514,30 @@ EXPECT Hello
...
+EXPECT_REGEX @: 1000000\n\n
+             \+-----------\+-------------\+
+             \| BENCHMARK \| NANOSECONDS \|
+             \+-----------\+-------------\+
+             \| a         \| \d+[ ]+\|
+             \+-----------\+-------------\+
+TIMEOUT 2
+
+NAME BENCH multiple
+PROG BENCH:a { @a++; } BENCH:b { if (@b == 0) { print(@a); } @b++; } BENCH:c { if(@c == 0) { print(@b);  } @c++; } BEGIN { print(@c); clear(@a); clear(@b); clear(@c); exit(); }
```
amscanne: Per discussion elsewhere, I think we need an independent feature first where we automatically exit if there's nothing attached. There's no sense hanging there with nothing going on. With the `--bench` / `--test` proposal (that disables regular attachment), this would be the standard path. (Unless the user changes the `--test-mode`, in which case you can run tests and then attach as normal, and you don't exit.)

src/bpftrace.cpp
jrife: I was thinking more of a `config` option if you wanted to adjust the global default as opposed to a per-`BENCH` setting. I'm not sure I see the use case for wanting to adjust it so granularly. Do you see a reason someone might want to do this?

src/bpftrace.cpp
jrife: While it's legal c++, I think it was something clang was complaining about, which is why I did the explicit initialization. I'll play around with it again and see if I can get some configuration that isn't so verbose.

jrife:
> I'm pulling out some common bits here: https://github.com/bpftrace/bpftrace/pull/4356
> 
> Let's just have the time printing use the same standards either way

Sure. I will just call this column `AVERAGE TIME` and let `duration_str` generate the unix suffix in the table.

src/bpftrace.cpp
amscanne: Sure, no worries. I know that clang-tidy doesn't like partial initializer lists, I thought that the empty set `{}` was a special case. Anyways, whatever works.

src/bpftrace.cpp
amscanne: I would assume that the default global value is sufficient except for a few probes (e.g. maybe you have something rather heavyweight, calling external kfuncs, etc. and you don't want it to spin 1000000 times). Maybe you have both? Make the global default configurable, and just allow overrides. It's also something that can be added, so I don't think you need to do it as part of this initial PR. I'm just thinking out loud.

src/attached_probe.cpp
jordalgo: Ah ok. I still wonder if it’s easier to make this a special probe type and just check for the name when setting the attach type but I won’t block on this detail.

jordalgo:
> What if we add an option to suppress the output. I don't think we should outsmart ourselves by having special codegen for benchmarks.

I'm good with having an additional CLI option. I also wasn't suggesting different codegen (as presumably you'd want to measure the time it takes to put stuff in the ring buffer) but rather just change the behavior on the user-space side where by we consume the contents of the ring buffer but just not output anything to the console from `print` or `printf`.

tests/runtime/json-output
jordalgo: > My assumption would be that BEGIN could still be used to set up some common state.

Yeah this is a good point. `END` can also be used to clear maps you don't want printed to the console.

src/bpftrace.cpp
jordalgo: +1 for punting on this and just having a global value. We can add a config or another param to the probe later as needed.

jordalgo:
>  Putting a bunch of guardrails in seems like more trouble than it's worth considering the alternative is just letting people run a script, seeing that they messed up, and letting them fix it so it doesn't spew out a bunch of stuff.

Sure. It's a valid point that @amscanne was also advocating for. We can punt on the decision to clear maps or do anything different for `BENCH` probes until after we see how they're used more.  As mentioned above I still like the idea of a CLI arg to suppress print output.

tests/runtime/probe
jordalgo: Trying this: 
- https://github.com/bpftrace/bpftrace/pull/4358

amscanne:
> I'm good with having an additional CLI option. I also wasn't suggesting different codegen (as presumably you'd want to measure the time it takes to put stuff in the ring buffer) but rather just change the behavior on the user-space side where by we consume the contents of the ring buffer but just not output anything to the console from `print` or `printf`.

After my refactor lands (i.e. the bottom two changes at #4301), I will add `--output none` and it will be a trivial implementation of the output interface. 😀 We don't even need a new command line flag.

src/bpftrace.cpp
jrife: OK. I'll keep it as-is for now then. I'd like to err on the side of simplicity for now and expand it later as use cases arise. We could probably then make a more informed decision on which direction to take the config.

src/attached_probe.cpp
jrife: I may try that configuration and see how messy the code feels. If it looks gross, I'll just go with the different probe types. Since this is all internal, I guess they could be merged later if need be. Overall, I just want to avoid bunch of special-casing around `BENCH`.

src/attached_probe.cpp
jrife: Resolving this, since we're discussing this topic in the thread above.

src/output.cpp
jrife: So that this isn't blocked on https://github.com/bpftrace/bpftrace/pull/4356, I can hand-code the `ns` suffix for now and switch to `DisplayUnit` when that lands. WDYT?

tests/runtime/json-output
jrife: Makes sense to me. I'll make it so `BEGIN` and `END` always run first and last, even in benchmark mode.

src/bpftrace.cpp
jrife: Also mentioned above, I'll go ahead and make it so `BEGIN` runs at the start even in benchmark/test mode. I think this makes sense.

tests/runtime/probe
jrife: Nice. With this, I think we don't need any special handling. The `BENCH` probes won't count towards keeping bpftrace running and it should just follow the quick exit path when in benchmark mode.

jrife:
Waiting for https://github.com/bpftrace/bpftrace/pull/4358 to land to do a rebase. This should clean up some hacks I put in place inside `src/bpftrace.cpp` to exit without running probes in bench mode. Trying to get the behavior working with the `--test-bench` flag discussed [here](https://github.com/bpftrace/bpftrace/issues/4347#issuecomment-3064141399).

man/adoc/bpftrace.adoc
jrife: Done.

man/adoc/bpftrace.adoc
jrife: Removed the `unstable_testing` config in the latest push.

tests/runtime/json-output
jrife: The latest push makes it so `BEGIN` runs first and `END` runs last. I added some hacks in `src/bpftrace.cpp` to get the behavior to be that bpftrace exits immediately in `bench` mode to get the tests working, but that will get cleaned up after I rebase on https://github.com/bpftrace/bpftrace/pull/4358.

src/bpftrace.cpp
jrife: Done.

src/bpftrace.cpp
jrife: This works -_-. I'm not sure what I saw the first time around that led me to this belief. Anyway, I just zero-initialize it with `= {}` in the latest push.

man/adoc/bpftrace.adoc
```diff
@@ -1270,6 +1275,45 @@ END {
...
 ----
 
+[#probes-bench]
+=== BENCH
+
+`BENCH` is a special built-in probe type for creating microbenchmarks.
+bpftrace executes each `BENCH` probe repeatedly to measure the average
+execution time of the contained code. If multiple `BENCH` probes exist
+in a script, bpftrace executes them sequentially in the order they are
+specified. `BENCH` probes execute before `BEGIN`.
```
jordalgo:
```suggestion
specified. 
```
I guess you'll update this language more when adding the details about the CLI flag.

src/attached_probe.cpp
jordalgo: > Overall, I just want to avoid bunch of special-casing around BENCH.

Yeah I understand. I guess it just depends how in common the rest of the code is with the "special"probe type

src/main.cpp
```diff
@@ -472,13 +473,16 @@ Args parse_args(int argc, char* argv[])
...
-          args.test_mode = TestMode::BENCHMARK;
-        else {
-          LOG(ERROR) << "USAGE: --test can only be 'codegen'.";
+        } else if (std::strcmp(optarg, "compiler-bench") == 0) {
+          args.test_mode = TestMode::COMPILER_BENCHMARK;
+        } else if (std::strcmp(optarg, "bench") == 0) {
+          args.test_mode = TestMode::BPF_BENCHMARK;
+        } else {
+          LOG(ERROR) << "USAGE: --test can only be 'codegen', "
+                        "'compiler-bench', 'bench', or 'test'.";
```
jordalgo:
```suggestion
          LOG(ERROR) << "USAGE: --test-mode can only be 'codegen', "
                        "'compiler-bench', or 'bench'";
```
We don't have the `test` mode yet.

tests/runtime/json-output
```diff
@@ -190,3 +190,14 @@ NAME print bool type
 RUN {{BPFTRACE}} -q -f json -e 'BEGIN { print((true, false)); exit(); }'
 EXPECT {"type": "value", "data": [true,false]}
 TIMEOUT 1
+
+NAME BENCH
+RUN {{BPFTRACE}} -q -f json --test-mode bench -e 'BENCH:a { @ = count(); }'
```
jordalgo, line 195: Maybe add a new test with a superfluous probe to verify that it did not attach and maybe add a `BEGIN` or `END` to make sure they do run and in the right order.

src/output.cpp
amscanne: I'll submit and you can use it if you like. I'm not concerned either way.

man/adoc/bpftrace.adoc
jrife: Done.

CHANGELOG.md
jrife: Done.

man/adoc/bpftrace.adoc
jrife: Done.

src/main.cpp
jrife: Removed this.

tests/runtime/probe
jrife: The new behavior is that in bench mode (`--test-mode bench`) bpftrace always exits immediately and prints the benchmark results. If `BEGIN` or `END` probes are present, these are run before and after the benchmarks, respectively.

src/output.cpp
jrife: I ended up using `util::duration_str` to generate the time string.

src/bpftrace.cpp
```diff
@@ -809,7 +838,8 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
     if (err)
       return err;
   } else {
-    bool should_drain = num_begin_end_attached > 0 &&
+    bool should_drain = (num_begin_end_attached > 0 ||
+                         !benchmark_results.empty()) &&
```
jordalgo: Nit: Maybe this is slightly more grokable
```suggestion
    bool should_drain = (num_begin_end_attached > 0 ||
                        run_benchmarks_) &&
```

src/bpftrace.cpp
```diff
@@ -838,16 +868,6 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
 
   poll_output(out, /* drain */ true);
 
-  uint64_t total_lost_events = bytecode_.get_event_loss_counter(*this,
```
jordalgo: This shouldn't have been deleted.

src/run_bpftrace.cpp
```diff
@@ -81,8 +81,10 @@ int run_bpftrace(BPFtrace &bpftrace,
 
   std::cout << "\n\n";
 
+  bpftrace.print_benchmark_results(output);
```
jordalgo: I know this won't print anything if there are no benchmark results but maybe it reads easier if it's gated on `run_benchmarks` then you can just do
```
if (bpftrace.run_benchmarks_) {
  bpftrace.print_benchmark_results(output);
} else if (bpftrace.config_->print_maps_on_exit) {
  err = bpftrace.print_maps(output);
}
```

tests/semantic_analyser.cpp
```diff
@@ -2311,6 +2311,16 @@ TEST_F(SemanticAnalyserTest, begin_end_probes)
   test("END { 1 } END { 2 }", Error{});
 }
 
+TEST_F(SemanticAnalyserTest, bench_probes)
+{
+  test("BENCH:a { 1 } BENCH:b { 2 }");
+  test("BENCH: { 1 }", Error{ R"(
```
jordalgo, line 2317: Should this check for a non-empty name be in the attachpoint_parser?

jordalgo (commented):
Looking good! Just a few minor nits/fixes as well as the Changelog conflict (my fault).

Also, looks like this doc also needs updating:
https://github.com/bpftrace/bpftrace/blob/ce5236138c3617fcb462795ab4e007091d06d851/docs/developers.md?plain=1#L49

src/run_bpftrace.cpp
jrife: Done.

src/bpftrace.cpp
jrife: Oops, don't know how I dropped these lines. Fixed!

src/bpftrace.cpp
jrife: Changed.

tests/semantic_analyser.cpp
jrife: That makes more sense to me (I don't know why `abc:` or `abc::adsf` should be valid syntax at all). However, it might make more sense to change this in its own PR; I see a few more checks in `semantic_analyser.cpp` that check for empty `target` fields, etc.

jrife:
> Also, looks like this doc also needs updating:

Ack. I will add some blurb here and change the flag name(s).

Edit: Done.

jordalgo (approved):
Great work! Thanks for the contribution.

I'll wait a day to see if @viktormalik or @amscanne want to review otherwise I'll merge (we can always fix forward).

jrife:
@jordalgo It looks like there was a flake on one of the watchpoint tests. Is there any way to retrigger this? I would make a dummy change and force push to my branch again but thought maybe there is a more elegant way :).

jordalgo:
@jrife You can ignore that. That failure won't stop me from merging :)

PR #4344, opened 2025-07-06
codegen: Prefer *ARRAY* to *HASH* for scalar maps

jrife:
*TODO*

##### Checklist

### DIFF
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -213,6 +213,7 @@ AttachPoint &AttachPoint::create_expansion_copy(ASTContext &ctx,
     case ProbeType::interval:
     case ProbeType::profile:
     case ProbeType::special:
+    case ProbeType::benchmark:
     case ProbeType::iter:
     case ProbeType::invalid:
       break;
@@ -242,6 +243,7 @@ bool AttachPoint::check_available(const std::string &identifier) const
         return true;
       case ProbeType::invalid:
       case ProbeType::special:
+      case ProbeType::benchmark:
       case ProbeType::tracepoint:
       case ProbeType::fentry:
       case ProbeType::fexit:
@@ -257,6 +259,7 @@ bool AttachPoint::check_available(const std::string &identifier) const
         return true;
       case ProbeType::invalid:
       case ProbeType::special:
+      case ProbeType::benchmark:
       case ProbeType::kprobe:
       case ProbeType::kretprobe:
       case ProbeType::tracepoint:
@@ -287,6 +290,7 @@ bool AttachPoint::check_available(const std::string &identifier) const
         return true;
       case ProbeType::invalid:
       case ProbeType::special:
+      case ProbeType::benchmark:
       case ProbeType::interval:
       case ProbeType::software:
       case ProbeType::hardware:
--- a/src/ast/attachpoint_parser.cpp
+++ b/src/ast/attachpoint_parser.cpp
@@ -157,6 +157,8 @@ AttachPointParser::State AttachPointParser::parse_attachpoint(AttachPoint &ap)
   switch (probetype(ap.provider)) {
     case ProbeType::special:
       return special_parser();
+    case ProbeType::benchmark:
+      return benchmark_parser();
     case ProbeType::kprobe:
       return kprobe_parser();
     case ProbeType::kretprobe:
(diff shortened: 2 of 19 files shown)


amscanne:
This is great! I've wanted to do the array change since the map desugar bits. I poked at it for a bit, and IIRC you may need to work through some edge cases; array maps only support `uint32_t` key types, so I'm not sure if any of the code that automatically "upgrades" integers used for keys/values will need wrangling. Hopefully it goes smoothly :)

I know this is just a draft (and you were probably intending to do this anyways), but if the `BENCHMARK` change is split into a separate PR it could likely be landed independently.

amscanne:
> I know this is just a draft (and you were probably intending to do this anyways), but if the `BENCHMARK` change is split into a separate PR it could likely be landed independently.

LOL, I just saw that this is already a separate PR. Disregard the above.

jrife:
> LOL, I just saw that this is already a separate PR. Disregard the above.

Yep. Was going to try getting `BENCHMARK` in first then work through some edge cases in this PR. Just wanted to get an initial idea of the performance difference going from `*ARRAY*` to `*HASH*`. The numbers so far look good!

```
BEFORE                      AFTER                       DIFFERENCE
+-----------+-------------+ +-----------+-------------+ +----------+
| BENCHMARK | NANOSECONDS | | BENCHMARK | NANOSECONDS | | % CHANGE |
+-----------+-------------+ +-----------+-------------+ +----------+
| assign    | 152         | | assign    | 64          | | -57%     |
| avg       | 67          | | avg       | 39          | | -41%     |
| count     | 41          | | count     | 39          | |          |
| hist      | 93          | | hist      | 92          | |          |
| lhist     | 71          | | lhist     | 71          | |          |
| max       | 68          | | max       | 39          | | -41%     |
| min       | 67          | | min       | 40          | | -40%     |
| sum       | 66          | | sum       | 39          | | -41%     |
+-----------+-------------+ +-----------+-------------+ +----------+
```

```
BEFORE                      AFTER                       DIFFERENCE
+-----------+-------------+ +-----------+-------------+ +----------+
| BENCHMARK | NANOSECONDS | | BENCHMARK | NANOSECONDS | | % CHANGE |
+-----------+-------------+ +-----------+-------------+ +----------+
| lhist     | 72          | | lhist     | 39          | | -46%     |
+-----------+-------------+ +-----------+-------------+ +----------+
```

jrife:
> This is great! I've wanted to do the array change since the map desugar bits. I poked at it for a bit, and IIRC you may need to work through some edge cases; array maps only support uint32_t key types, so I'm not sure if any of the code that automatically "upgrades" integers used for keys/values will need wrangling. Hopefully it goes smoothly :)

With `BENCH` merged, I've been revisiting this. In its current state, this PR "works" for a few manual test cases, but I think that's more of a happy accident. It would be more proper to ensure that uint32 keys are being used in codegen when switching to `*ARRAY`s. This is slightly more tricky, since the assumption that integer map key = 64 bits seems somewhat baked in. There's also an assumption previously that `@` and `@[0]` are represented equivalently (mostly in the map_sugar pass, but this has consequences further down the line).

Right now, I'm exploring some changes to map sugar which would hopefully make this fit in well. I'm trying to find the right way to represent this that doesn't feel super hacky.

amscanne:
> > This is great! I've wanted to do the array change since the map desugar bits. I poked at it for a bit, and IIRC you may need to work through some edge cases; array maps only support uint32_t key types, so I'm not sure if any of the code that automatically "upgrades" integers used for keys/values will need wrangling. Hopefully it goes smoothly :)
> 
> With `BENCH` merged, I've been revisiting this. In its current state, this PR "works" for a few manual test cases, but I think that's more of a happy accident. It would be more proper to ensure that uint32 keys are being used in codegen when switching to `*ARRAY`s. This is slightly more tricky, since the assumption that integer map key = 64 bits seems somewhat baked in. There's also an assumption previously that `@` and `@[0]` are represented equivalently (mostly in the map_sugar pass, but this has consequences further down the line).
> 
> Right now, I'm exploring some changes to map sugar which would hopefully make this fit in well. I'm trying to find the right way to represent this that doesn't feel super hacky.

Yeah, the intent for the map desugar pass was to make the representation of all maps uniform. After that point, they should all be clearly keyed (e.g. `@` is no longer legal, only `@[0]` and all operations pass a key and can be dealt with uniformly). And then you operate and collapse them in ways that apply regardless of whether you have accessed as a "scalar" or not.

FWIW, my original approach was to look at [map bounds in the resource analyzer](https://github.com/bpftrace/bpftrace/commit/137b9022f0630b40409668be752a9d5ca35cbe8d#diff-dd0acf958e44461e5cac571093f6ed97ad4a0b86c5abe753b50e95fc16536b16L26). This was dropped just to keep it a bit simpler initially. Thinking out loud, I'd look at something like:

* Compute the map bounds on all maps in the resource analyzer, similar to the original change. Anything that can be said to use a constant range (say using at least 50% within some fixed range?) would get nominated to be an array-based map.
* All the key types for this map could be automatically marked as `Uint32`, ahead of semantic analysis or anything else. You might need to make the `Integer` type mutable, which is fine (defaulting to 64-bit).
* It *seems* like the semantic analyzer is okay with this, and won't mess with those types? But codegen does lots of messing with map keys and types. So ideally this can be fixed first, and we get rid of the transparent "upgrade" that integers get. Hopefully you can just remove some of this code without issue now, but it will be a journey.

PR #4345, opened 2025-07-06
Add: `socket_cookie` function

### LINKED ISSUE #4314: Add function that calls bpf_get_socket_cookie

This is so users can monitor per socket networking traffic statistics as it provides a global socket identifier that can be assumed unique e.g. they can use it in a map.

Discussion:
- https://github.com/bpftrace/bpftrace/discussions/4313

Perhaps just a new function `socket_cookie` or `skcookie` for brevity. 

This could be a new helper wrapper added to your stdlib in bpf.c (cc @amscanne)

mannkafai:
This function provides a way to retrieve the cookie (generated by the kernel) of a socket.

This will close #4314.

##### Checklist


src/ast/passes/semantic_analyser.cpp
```diff
@@ -595,6 +595,11 @@ static const std::map<std::string, call_spec> CALL_SPEC = {
       .max_args=1,
       .discard_ret_warn = true },
   },
+  { "socket_cookie",
+    { .min_args=1,
+      .max_args=1,
+      .arg_types={
+        arg_type_spec{ .type=Type::pointer } } } },
```
jrife: nit: Add a comment like the one on line 495 to document the pointer type: `struct sock *`.

tests/runtime/call
```diff
@@ -731,3 +731,16 @@ AFTER ./testprogs/syscall nanosleep  1e8
 NAME block expression with map assignment
 PROG BEGIN { @ = { let $a = 100; avg($a) }; exit(); }
 EXPECT @: 100
+
+NAME socket_cookie
```
jrife: nit: I feel like these tests belong near those for `skboutput`, since they're both networking-related helpers.

man/adoc/bpftrace.adoc
```diff
@@ -3119,6 +3123,19 @@ dropped privs to tcpdump
...
+
+.variants
+* `uint64 socket_cookie(struct sock *sk)`
+
+*Helper* `bpf_get_socket_cookie`
+
+Retrieve the cookie (generated by the kernel) of the socket.
+If no cookie has been set yet, generate a new cookie. Once generated, the socket cookie remains stable for the life of the socket.
+
+This function returns a 8-byte long unique number on success, or 0 if *sk* is NULL.
```
jordalgo: Think it's worth adding an example where by we get the `struct sock *sk` from an fentry arg?

src/ast/irbuilderbpf.cpp
```diff
@@ -2188,6 +2188,21 @@ void IRBuilderBPF::CreateGetCurrentComm(AllocaInst *buf,
   CreateHelperErrorCond(call, libbpf::BPF_FUNC_get_current_comm, loc);
 }
 
+CallInst *IRBuilderBPF::CreateGetSocketCookie(Value *var, const Location &loc)
+{
+  // u64 bpf_get_socket_cookie(struct sock *sk)
+  // Return:
+  //    A 8-byte long unique number or 0 if *sk* is NULL.
+  FunctionType *getsocketcookie_func_type = FunctionType::get(
```
jordalgo: nit
```suggestion
  FunctionType *get_socket_cookie_func_type = FunctionType::get(
```

tests/runtime/call
```diff
@@ -731,3 +731,16 @@ AFTER ./testprogs/syscall nanosleep  1e8
...
+SETUP ip netns add bpftrace-test && ip -netns bpftrace-test link set lo up
+AFTER ip netns exec bpftrace-test ./testprogs/syscall connect 127.0.0.1 80
+CLEANUP ip netns delete bpftrace-test
+EXPECT_REGEX ret: [0-9]+
+REQUIRES_FEATURE fentry
+
+# Just test that the verifier doesn't reject this
+NAME socket_cookie rawtracepoint
+RUN {{BPFTRACE}} -e 'rawtracepoint:tcp_probe { $ret = socket_cookie(args->sk); printf("ret: %llu\n", $ret); exit(); }'
+EXPECT Attached 1 probe
```
jordalgo: missing a new line

jordalgo (commented):
Should we also add a check for the argument type passed to `socket_cookie` in semantic analyser?

src/ast/passes/semantic_analyser.cpp
mannkafai: I'll add it.

man/adoc/bpftrace.adoc
mannkafai: how about this one? 

```
fentry:tcp_rcv_established
{
	$cookie = socket_cookie(args->sk);
	@psize[$cookie] = hist(args->skb->len);
}
```

jordalgo: LGTM

man/adoc/bpftrace.adoc
```diff
@@ -3119,6 +3123,43 @@ dropped privs to tcpdump
...
```
jordalgo: nit - framed in terms of bpftrace int types
```suggestion
This function returns a `uint64` unique number on success, or 0 if *sk* is NULL.
```

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1747,6 +1753,22 @@ If you're seeing errors, try clamping the string sizes. For example:
                         << arg.type().GetTy() << " provided)";
       }
     }
+  } else if (call.func == "socket_cookie") {
+    const auto &type = call.vargs.at(0).type();
+    if (!type.IsPtrTy() || !type.GetPointeeTy() ||
+        !type.GetPointeeTy()->IsRecordTy()) {
+      call.addError() << call.func
```
jordalgo: nit: you can just make a small lambda at the top that sets the error so you don't have to repeat this string

jordalgo (commented):
Few more small nits but otherwise this looks great.

src/ast/passes/semantic_analyser.cpp
```diff
@@ -1169,6 +1169,26 @@ void SemanticAnalyser::visit(Call &call)
...
+          return false;
+        }
+        if (!type.GetPointeeTy()->IsSameType(CreateRecord(name))) {
+          call.addError() << call.func << "() only supports '" << name
+                          << " *' as the argument ('"
+                          << type.GetPointeeTy()->GetName() << " *' provided)";
+          return false;
+        }
+        return true;
+      };
```
jordalgo: Sorry, what I meant was, don't make this whole thing a lambda, just the error logging part and please move it back to inside the `call.func == "socket_cookie"` block e.g.
```suggestion
  auto CheckArgIsRecordPtr =
      [](Call &call, size_t index, const std::string &name) {
        assert(index < call.vargs.size());
        const auto &type = call.vargs.at(index).type();
        auto logError = [&]<typename T>(T provided) {
           call.addError() << call.func << "() only supports '" << name
                          << " *' as the argument (" << provided
                          << " provided)";
        };
        if (!type.IsPtrTy() || !type.GetPointeeTy() ||
            !type.GetPointeeTy()->IsRecordTy()) {
           logError(type.GetTy());
           return;
        }
        if (!type.GetPointeeTy()->IsSameType(CreateRecord(name))) {
          logError(type.GetPointeeTy()->GetName());
          return;
        }
        return;
      };
```

mannkafai: I misunderstood.

PR #4346, opened 2025-07-07
codegen: add headers commonly needed for C code

amscanne:
Stacked PRs:
 * __->__#4346

--- --- ---

### codegen: add headers commonly needed for C code

These are the headers recursively vendored by `libbpf`, and therefore
commonly used by BPF C code. A script is added `sync_headers.sh` which
can synchronize these to the local system. These headers have been
synchronized from `e34a79b96ab9d49ed8b605fee11099cf3efbb428`.


jordalgo (approved):
LGTM. Do you mind just adding this line to the top of sync_headers.sh ? "These are the headers recursively vendored by libbpf, and therefore commonly used by BPF C code."

amscanne:
I've added a little explanation block. This is indeed the current description, but I also added that we may evolve this over time and there are certain classes of headers that we should *not* add.

amscanne:
I noticed that there are a couple headers that are aarch64-specific (coming from my machine). I need to fix up these cases, and potentially sync from the actual kernel repo (sigh) to ensure that it's structured in a fully generic way. So not merging just yet, hopefully will fix that by EOW.

amscanne:
The script and plumbing got a bit more complicated to handle the multi-arch case (mapping the right `asm` directory, effectively), but is all done now. The binary will carry the other arches in its `data` section and only map the ones that matter. I don't really want to optimize the build bits right now (feels premature), this can be done in the future along with a host of other improvements to this setup. This is just enough to get things working and provide some basic headers.

amscanne:
@jordalgo @viktormalik I'll submit in the morning, unless there are objections.

jordalgo (commented):
I appreciate the irony of what I'm about to say but I think the sync_headers script needs more comments e.g. what kernel version are we currently syncing/using, why are we invoking clang, what's the deal with rwonce?

amscanne:
> I appreciate the irony of what I'm about to say but I think the sync_headers script needs more comments e.g. what kernel version are we currently syncing/using, why are we invoking clang, what's the deal with rwonce?

I rewrote the script (again), and added more comments. I don't think it will be perfect regardless, so hopefully it's a "good enough for now" state.

PR #4348, opened 2025-07-08
Use bytes to create ints and enums

jordalgo:
Small refactor so we don't have to multiply
or divide values by 8 around the codebase
as much.

I also don't think we actually support any
int types of 1 bit.

##### Checklist


amscanne:
This is perfectly reasonable and good, but it's also the kind of change that will bite you on an automatic rebase because the semantics have changed but the types and parameters haven't. Can we sit on it a bit until it's quiet on the types front?

jordalgo:
> Can we sit on it a bit until it's quiet on the types front?

Sure. This was something I noticed randomly and thought would be nice to improve but no rush at all.

PR #4351, opened 2025-07-09
tests: refactor semantic_analyzer tests

amscanne:
tests: refactor semantic_analyzer tests

I just couldn't bring myself to add one more overloaded `test` method or
one more vague set of parameters. This commit refactors the semantic
analyzer tests, and reduces the number of `test` methods to just one.

This new method uses a very basic pattern that is similar to the
gtest-style matchers, where you pass objects that you'd like to use for
validation into the `test` method. If you pass an `Error` argument, then
the program must emit an error, and if there is a string provided, it
must be contained with the error. Similarly for `Warning` and
`NoWarning`. If you provide `Mock` the test will be run with your custom
`BPFtrace` object, and this will be shortly extensible to `Types`.

The `test` method is also constant-parameterized by a set of options,
notable `features`, `safe_mode` and `with_child`. These are ways of
configuration the mock that will be created for the test.

In order to ensure that individual test cases don't need to create mocks
just for lifetime issues, the tests are wrapped in a simple harness.
Maybe one day this can be eliminated, but not today. Its purpose is
explained in a comment.

Finally, there is one related code change. I can't for the life of me
understand why this refactor caused `type_ctx` to start failing, but it
does. That test was not changed, and now triggers an assertion failure.
I suspect that previously the tests were loading `BTF` data as a
side-effect somewhere, and therefore the lookup of `struct pt_regs` was
successfully. Because all the instances of `BPFtrace` have now been
replaced with the mocks in this test, it is likely that these are no
longer loading. Therefore this assertion failure code path is updated by
leaving the type of `ctx` as `None`.


tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
+// fixture, and allows us to have a single harness.
+class SemanticAnalyserHarness {
+public:
+  template <Options options = Options{}, typename... Ts>
+    requires((std::is_same_v<std::decay_t<Ts>, Mock> ||
+              std::is_same_v<std::decay_t<Ts>, Warning> ||
+              std::is_same_v<std::decay_t<Ts>, NoWarning> ||
+              std::is_same_v<std::decay_t<Ts>, Error> ||
+              std::is_same_v<std::decay_t<Ts>, ExpectedAST>) &&
+             ...)
```
viktormalik: This looks neat although I'm not much familiar with the syntax.

Do I understand correctly that this places restrictions on the types of values passed to `args` and if a different type is passed, the compiler will complain?

Also, what's the `&& ...` part for?

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
-  } else {
-    // Accept any failure result.
-    EXPECT_FALSE(ok && ast.diagnostics().ok()) << msg.str();
-  }
-  if (expected_error.data() && !expected_error.empty()) {
-    if (!expected_error.empty() && expected_error[0] == '\n')
-      expected_error.remove_prefix(1); // Remove initial '\n'
-    EXPECT_EQ(out.str(), expected_error);
+  if constexpr (sizeof...(Ts) != 0) {
+    return extract<T, Ts...>(std::forward<Ts>(rest)...);
```
viktormalik, line 60: What's this recursion for?

viktormalik (approved):
This looks very nice!

Just two questions below, mainly to expand my C++ templates knowledge :sweat_smile:

tests/semantic_analyser.cpp
amscanne: Right, if you try to pass something else, then it will not compile. `std::decay_t` is just removing any qualifiers on the passed type (e.g. `const`, references, etc.).

The `&& ...` is a [fold expression](https://en.cppreference.com/w/cpp/language/fold.html) over the [parameter pack](https://en.cppreference.com/w/cpp/language/parameter_pack.html) Ts. Basically it is just requires that same expression `&&`ed for all types in `Ts`.

tests/semantic_analyser.cpp
amscanne: The `extract` function is just looking for the first argument over the given type. It checks the `First` type, and if it doesn't match, it recursively calls `extract` on the subsequent arguments.

I don't think it's possible to do this kind of metaprogramming using iteration (@jordalgo I concede this is metaprogramming), because you can't match the types individual in a parameter pack. So this recursion is a common pattern (e.g. with `First` type and then a pack for the rest). FWIW, it doesn't actually build a recursive function, it's just recursive expansion :)

I would normally avoid adding this kind of thing, but given that there were **fourteen** different variants of the `test` function (including `test_error` and `test_warning` variants), all with slightly different overloaded arguments and semantics, I actually feel this is an improvement. You just need to accept a bit of template magic, but then there's only a single `test` function and I believe all the calls are more understandable in general.

tests/semantic_analyser.cpp
viktormalik: Thanks for the explanation!

tests/semantic_analyser.cpp
viktormalik: > The `extract` function is just looking for the first argument over the given type. It checks the `First` type, and if it doesn't match, it recursively calls `extract` on the subsequent arguments.
> 
> I don't think it's possible to do this kind of metaprogramming using iteration (@jordalgo I concede this is metaprogramming), because you can't match the types individual in a parameter pack. So this recursion is a common pattern (e.g. with `First` type and then a pack for the rest). FWIW, it doesn't actually build a recursive function, it's just recursive expansion :)

Oh, right, I didn't notice we're using `extract` for the entire `args` list so we actually need to traverse it. The recursive expansion makes perfect sense, then.

> I would normally avoid adding this kind of thing, but given that there were **fourteen** different variants of the `test` function (including `test_error` and `test_warning` variants), all with slightly different overloaded arguments and semantics, I actually feel this is an improvement. You just need to accept a bit of template magic, but then there's only a single `test` function and I believe all the calls are more understandable in general.

Agreed. My first thought was "aren't we complicating things too much?" but the more I looked at it, the more sense it made. This is definitely an improvement from my perspective.

jordalgo:
I'm not stamping this until I understand it. To be clear, I'm not blocking but I want to spend some time, like @viktormalik did, to understand this new fangled (probably 20 years old) C++ magic.

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
-
-  return ast;
+  return std::nullopt;
 }
 
-ast::ASTContext test(BPFtrace &bpftrace,
-                     const std::string &input,
-                     bool safe_mode = true)
+template <typename T>
+std::optional<T> extract()
```
jordalgo, line 66: Why do we need this extract that accepts no params? Is it the case when `...args` passed to `test` doesn't exist that this is called when invoking extract?

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
+    ast::ASTContext ast("stdin", std::string(clean_prefix(input)));
 
-ast::ASTContext test(const std::string &input)
-{
-  auto bpftrace = get_mock_bpftrace();
-  return test(*bpftrace, true, input, 0, {}, true, false);
-}
+    // Reset for each iteration. We only guarantee that the types remain
+    // valid after the ASTContext has been returned.
+    bpftrace_ = get_mock_bpftrace();
```
jordalgo: why do we need this private `bpftrace_` object if we're using the mock? Also shouldn't `get_mock_bpftrace` only be called if the `mock` below is empty?

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
-         "Use test(const std::string&) for expected successes");
-  auto bpftrace = get_mock_bpftrace();
-  return test(*bpftrace, true, input, expected_result, {}, true, false);
-}
+// This exists as a test fixture because the types may refer to `bpftrace`, so
+// this objects lifetime must exceed the tests lifetime. This is easier with a
+// fixture, and allows us to have a single harness.
+class SemanticAnalyserHarness {
+public:
+  template <Options options = Options{}, typename... Ts>
```
jordalgo: JC, why aren't the options handled in the same way as the other args? Because of the default value?

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
 
-  if (expected_ast[0] == '\n')
-    expected_ast.remove_prefix(1); // Remove initial '\n'
+    if (!mock) {
+      mock.emplace(*bpftrace_);
+    }
+    mock->bpftrace.safe_mode_ = options.safe_mode;
+    mock->bpftrace.feature_ = std::make_unique<MockBPFfeature>(
+        options.features);
+    if constexpr (options.with_child) {
```
jordalgo: neat - I should do this more

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
+                  .add(ast::CreateCMacroExpansionPass())
+                  .add(ast::CreateFoldLiteralsPass())
+                  .add(ast::CreateMapSugarPass())
+                  .add(ast::CreateSemanticPass())
+                  .run();
+    EXPECT_TRUE(bool(ok));
+
+    std::stringstream out;
+    ast.diagnostics().emit(out, ast::Diagnostics::Severity::Warning);
+    if (warning && !warning->str.empty()) {
```
jordalgo: I didn't see any cases where the warning string was empty. Should we just assert that it's not (same for nowarning below) ?

Maybe we just add a check in `extract` when the type has a `str` field that it's not empty?

jordalgo:
OK, I went through it and it wasn't that bad (even for meta programming from a Meta programmer). Just had a few small questions.

tests/semantic_analyser.cpp
```diff
@@ -8,237 +11,150 @@
...
+      EXPECT_THAT(out.str(), HasSubstr(clean_prefix(warning->str)))
+          << out.str();
+    }
+    if (nowarning && !nowarning->str.empty()) {
+      EXPECT_THAT(out.str(), Not(HasSubstr(clean_prefix(nowarning->str))))
+          << out.str();
+    }
+    out.str("");
+    ast.diagnostics().emit(out, ast::Diagnostics::Severity::Error);
+    if (error && !error->str.empty()) {
```
jordalgo: Actually I do see `Error{}` used below but then if `error->str` is empty shouldn't we assert that `out.str()` is not empty?

tests/semantic_analyser.cpp
amscanne: Yeah, I guess I could make them all uniform. I started trying to use exclusively templates, but you can't constify a string easily for use as a parameter. I don't have a super strong opinion.

tests/semantic_analyser.cpp
amscanne: The `Error` is empty *most* of the time, but you're right -- the warning isn't. I can make this an error.

tests/semantic_analyser.cpp
amscanne: Ah, good point.

tests/semantic_analyser.cpp
amscanne: Correct.

tests/semantic_analyser.cpp
amscanne: I've made this consistent and they are just extracted as arguments in whatever order.

tests/semantic_analyser.cpp
amscanne: Okay, I've made this conditionally initialized and added a bit in the comment. It will be unilaterally cleared but only created if you don't provide your own mock.

tests/semantic_analyser.cpp
amscanne: It's gone now :) I've gone with your suggestion and will just be using a single form of metaprogramming, which makes it a bit simpler.

tests/semantic_analyser.cpp
jordalgo: Nice. I like it better.

jordalgo (approved):
Looks great to me. Nice cleanup!

PR #4353, opened 2025-07-09
Refactor naming of `Probe` objects

viktormalik:
The `Probe` class, which we use to represent a single "BPF probe" has two "name" properties - `name` which represents the name of the probe after the expansion has been done and `orig_name` which should correspond to the original user input. At the moment, `orig_name` is used in 3 places:
1. Reporting the name of the probe that bpftrace is trying to attach to in verbose mode.
2. Checking if the `Probe` object was created from an expanded attach point and potentially reporting the user the original name of the attach point in case of errors.
3. Internal naming of some entities.

Point 1 may lead to some not very informative outputs:
```
# bpftrace -v -e 'fentry:vfs_read* { exit() }'
Trying to attach probe: fentry:*:vfs_read*
Trying to attach probe: fentry:*:vfs_read*
Trying to attach probe: fentry:*:vfs_read*
Attached 3 probes
```
so the first patch changes it to use `name` instead.

Another problem is that once we move the probe expansion prior to codegen, the way `orig_name` is constructed will not correspond to the original user input, which will be a problem for the user-facing point 2 above. So, the second patch of this PR refactors the creation of `orig_name` to preserve the original user input.

See commit messages for more details and examples.

This is another refactoring needed for #4114.

##### Checklist


viktormalik:
There are some unrelated failures in the codegen tests which need to be fixed on master.

jordalgo:
@viktormalik Yup, probably my fault when I merged the EVENT_LOSS_COUNTER change. Will put up a fix.

amscanne:
A bit of an aside: we have a problem wherein probe names are actually bound to the specific files e.g. in the case of usdt/uprobe names. This makes our tests fragile. For example, codegen tests actually depend on writing a file to `/tmp/dwarf_data` 🤦 

I'd love if we could move towards a model where the "name" internally was just a incrementing unique ID per probe instance (potentially stemming from the `ASTContext`). This is actually the only blocker for #4164. We can still keep an expanded name, but move towards a `id()` or something else (ideally numeric!) when we actually need a unique thing to do naming in LLVM modules, etc.

CHANGELOG.md
```diff
@@ -90,6 +90,8 @@ and this project adheres to
   - [#4194](https://github.com/bpftrace/bpftrace/pull/4291)
 - Only cache symbols from targeted process
   - [#4315](https://github.com/bpftrace/bpftrace/pull/4315)
+- Refactor naming of Probe objects
```
jordalgo: Nit: I think this qualifies as a fix but NBD.

jordalgo:
@amscanne I assume we'd still use the expanded (non numeric) probe name for actual program loading/attachment...

amscanne:
> @amscanne I assume we'd still use the expanded (non numeric) probe name for actual program loading/attachment...

Yeah that's fine. I think we sort of have three things that are a occasionally conflated, and perhaps it would be good to make them explicit:

* `display_name` -- the thing you could show to users, no uniqueness guarantees
* `id` (does not exist formally at the moment, we sort of use "name" in some places for this) -- a unique probe identifier for use as symbol names, looking up the probe eetc.
* `args_typename` -- Does this need to be a property of the probe? Shouldn't this be a property of the attachpoint? And then we split if this is unequal, no? (And after split, you can assume that all attachpoints are the same, e.g. `probe.attachpoints().front().args_typename()` is fine. No harm in having a convenience accessor, but why not make this call through to a property of the attachpoint (and validate) rather than juggling state?

jordalgo:
> id (does not exist formally at the moment, we sort of use "name" in some places for this) -- a unique probe identifier for use as symbol names, looking up the probe etc

Do we have cases where `name` wouldn't be unique and therefore a bad `id` or is it just an internal naming convention where we should be using an `id` field which would contain the name?

viktormalik:
Yeah, agreed, using `name` (which is potentially non-unique) for internal probe identification is cumbersome and forces us to have things like `ast::Probe::index_` and `ast::AttachPoint::index_`.

> > @amscanne I assume we'd still use the expanded (non numeric) probe name for actual program loading/attachment...
> 
> Yeah that's fine. I think we sort of have three things that are a occasionally conflated, and perhaps it would be good to make them explicit:
> 
> * `display_name` -- the thing you could show to users, no uniqueness guarantees

We'll still need two variants - non-expanded and expanded one - as each is used in different situations.

> * `id` (does not exist formally at the moment, we sort of use "name" in some places for this) -- a unique probe identifier for use as symbol names, looking up the probe eetc.
> * `args_typename` -- Does this need to be a property of the probe? Shouldn't this be a property of the attachpoint? And then we split if this is unequal, no? (And after split, you can assume that all attachpoints are the same, e.g. `probe.attachpoints().front().args_typename()` is fine. No harm in having a convenience accessor, but why not make this call through to a property of the attachpoint (and validate) rather than juggling state?

You're probably right. I'm not sure this name is even exposed to users (shouldn't be) so we may be able to replace it by something `id`-based.

Anyways, I think this all will be better refactored when we move expansion to a separate pass so let me reiterate then. The changes in this PR are still going to be needed so I'm going to merge later today if there are no objections.

viktormalik:
> > id (does not exist formally at the moment, we sort of use "name" in some places for this) -- a unique probe identifier for use as symbol names, looking up the probe etc
> 
> Do we have cases where `name` wouldn't be unique and therefore a bad `id` or is it just an internal naming convention where we should be using an `id` field which would contain the name?

IIUC, that shouldn't happen since we have `ast::Probe::index_`, `ast::AttachPoint::index_`, and `Probe::index`. IMO one of the problems is that the first two are set from codegen while they should be set from `ASTContext` instead.

amscanne:
> Do we have cases where `name` wouldn't be unique and therefore a bad `id` or is it just an internal naming convention where we should be using an `id` field which would contain the name?

The `id` needs a few properties:

* Uniqueness -- exactly identifies a single probe
* Stability -- does not depend on local paths, etc.

The `name` can easily satisfy the first, but does not satisfy the second. We should use the second in LLVM identifiers, etc., so that the generated bitcode is not generally not bound to the specific paths. (This is the case currently, which is why `/tmp/dwarf` appears in the codegen tests bitcode, and it can't be parallelized.)

Anyways, it seems like we'd generally on the same page with this and I'm not hung up on the particulars. The space of valid solutions is reasonable large, I just wanted to call attention to it if you're refactoring this area :)

viktormalik:
> Anyways, it seems like we'd generally on the same page with this and I'm not hung up on the particulars. The space of valid solutions is reasonable large, I just wanted to call attention to it if you're refactoring this area :)

Yeah, it's a good call, I'll definitely look into that.

CHANGELOG.md
viktormalik: Fixed

PR #4354, opened 2025-07-09
Add ability to attach to BPF progs and sub progs

jordalgo:
Examples:
- `fentry:bpf:prog_name { ... }`
- `fentry:bpf:prog_id:prog_name { ... }`

This requires obtaining the fd for the BPF
program and passing it to `bpf_program__set_attach_target`.

Code for `get_prog_full_name` was shamelessly
taken from bpftool.

Getting a list of running BPF programs and sub programs is
also supported:
```
bpftrace -l 'fentry:bpf:*'
fentry:bpf:4217779:dnswatch_kprobe_udp_recvmsg
fentry:bpf:4217779:populate_cmdline
fentry:bpf:4217780:armr_env_get_len
fentry:bpf:4217780:armr_get_cgroup_from_task
```

You'll notice the names are prefixed with
their program ids.

##### Checklist

### DIFF
--- a/man/adoc/bpftrace.adoc
+++ b/man/adoc/bpftrace.adoc
@@ -1441,6 +1441,8 @@ iter:task_file:/sys/fs/bpf/files {
 .variants
 * `fentry[:module]:fn`
 * `fexit[:module]:fn`
+* `fentry:bpf[:prog_id]:prog_name`
+* `fexit:bpf[:prog_id]:prog_name`
 
 .short names
 * `f` (`fentry`)
@@ -1460,6 +1462,15 @@ The original names are still supported for backwards compatibility.
 This removes the need for manual type casting and makes the code more resilient against small signature changes in the kernel.
 The function arguments are available in the `args` struct which can be inspected by doing verbose listing (see <<Listing Probes>>).
 These arguments are also available in the return probe (`fexit`), unlike `kretprobe`.
+
+The bpf variants (e.g. `fentry:bpf[:prog_id]:prog_name`) allow attaching to running BPF programs and sub-programs.
+For example, if bpftrace was already running with a script like `uprobe:./testprogs/uprobe_test:uprobeFunction1 { print("hello"); }` then you could attach to this program with `fexit:bpf:uprobe___testprogs_uprobe_test_uprobeFunction1_1 { print("bye"); }` and this probe would execute after (because it's `fexit`) the `print("hello")` probe executes.
+You can specify just the program name, and in this case bpftrace will attach to all running programs and sub-programs with that name.
+You can differentiate between them using the `probe` builtin.
+You can also specify the program id (e.g. `fentry:bpf:123:*`) to attach to a specific running BPF program or sub-programs called in that running BPF program.
+To see a list of running, valid BPF programs and sub-programs use `bpftrace -l 'fentry:bpf:*'`.
+Note: only BPF programs with a BTF Id can be attached to.
+Also, the `args` builtin is not yet available for this variant.
 
 ----
 # bpftrace -lv 'fentry:tcp_reset'
--- a/src/ast/attachpoint_parser.cpp
+++ b/src/ast/attachpoint_parser.cpp
@@ -655,11 +655,39 @@ AttachPointParser::State AttachPointParser::watchpoint_parser(bool async)
 AttachPointParser::State AttachPointParser::fentry_parser()
 {
   // fentry[:module]:function
-  if (parts_.size() != 2 && parts_.size() != 3) {
+  // fentry:bpf:[:prog_id]:prog_name
+  if (parts_.size() != 2 && parts_.size() != 3 && parts_.size() != 4) {
     if (ap_->ignore_invalid)
       return SKIP;
 
-    return argument_count_error(1, 2);
+    return argument_count_error(1, 3);
+  }
+
+  if (parts_[1] == "bpf") {
+    ap_->target = parts_[1];
+    if (parts_.size() == 2) {
+      errs_ << "the 'bpf' variant of this probe requires a bpf program name "
+               "and optional bpf program id";
+      return INVALID;
+    } else if (parts_.size() == 3) {
+      ap_->func = parts_[2];
+    } else {
+      ap_->func = parts_[3];
+      if (parts_[2] != "*") {
+        auto uint_res = util::to_uint(parts_[2]);
+        if (!uint_res) {
+          errs_ << "bpf program id must be a number or '*'";
+          return INVALID;
+        }
+        ap_->bpf_prog_id = *uint_res;
+      }
+    }
+    return OK;
+  }
+
+  if (parts_.size() == 4) {
+    errs_ << "Only the 'bpf' variant of this probe supports 4 arguments";
+    return INVALID;
   }
 
   if (parts_.size() == 3) {
--- a/src/bpfprogram.cpp
+++ b/src/bpfprogram.cpp
@@ -7,6 +7,7 @@
 #include "bpfprogram.h"
 #include "log.h"
 #include "util/exceptions.h"
+#include "util/fd.h"
 
 namespace bpftrace {
 
@@ -67,7 +68,7 @@ void BpfProgram::set_attach_target(const Probe &probe,
 
   const std::string &mod = probe.path;
   const std::string &fun = probe.attach_point;
-  const std::string attach_target = !mod.empty() ? mod + ":" + fun : fun;
+  std::string attach_target = !mod.empty() ? mod + ":" + fun : fun;
 
   std::string btf_fun;
   __u32 btf_kind = BTF_KIND_FUNC;
@@ -81,17 +82,34 @@ void BpfProgram::set_attach_target(const Probe &probe,
     btf_fun = fun;
   }
 
-  if (btf.get_btf_id(btf_fun, mod, btf_kind) < 0) {
-    const std::string msg = "No BTF found for " + attach_target + ".";
+  std::string err_msg;
+
+  if ((probe.type == ProbeType::fentry || probe.type == ProbeType::fexit) &&
+      mod == "bpf") {
+    int raw_fd = bpf_prog_get_fd_by_id(static_cast<__u32>(probe.bpf_prog_id));
+    if (raw_fd < 0) {
+      err_msg = "No valid BPF program found with name: " + fun +
+                " and id: " + std::to_string(probe.bpf_prog_id) + ".";
+    } else {
+      bpf_prog_fd_ = util::FD(raw_fd);
+      attach_target = fun;
+    }
+  } else if (btf.get_btf_id(btf_fun, mod, btf_kind) < 0) {
+    err_msg = "No BTF found for " + attach_target + ".";
+  }
+
+  if (!err_msg.empty()) {
     if (config.missing_probes == ConfigMissingProbes::error) {
-      LOG(ERROR) << msg;
+      LOG(ERROR) << err_msg;
     } else if (config.missing_probes == ConfigMissingProbes::warn) {
-      LOG(WARNING) << msg;
+      LOG(WARNING) << err_msg;
     }
     bpf_program__set_autoload(bpf_prog_, false);
   }
 
-  bpf_program__set_attach_target(bpf_prog_, 0, attach_target.c_str());
+  bpf_program__set_attach_target(bpf_prog_,
+                                 bpf_prog_fd_ ? *bpf_prog_fd_ : 0,
+                                 attach_target.c_str());
 }
 
 void BpfProgram::set_no_autoattach()
--- a/src/bpfprogram.h
+++ b/src/bpfprogram.h
@@ -6,6 +6,7 @@
 #include "btf.h"
 #include "config.h"
 #include "probe_types.h"
+#include "util/fd.h"
 
 namespace bpftrace {
 
@@ -35,6 +36,7 @@ public:
 
 private:
   struct bpf_program *bpf_prog_;
+  std::optional<util::FD> bpf_prog_fd_ = std::nullopt;
 };
 
 } // namespace bpftrace
--- a/src/probe_matcher.cpp
+++ b/src/probe_matcher.cpp
@@ -13,6 +13,7 @@
 #include "probe_matcher.h"
 #include "scopeguard.h"
 #include "tracefs/tracefs.h"
+#include "util/bpf_progs.h"
 #include "util/paths.h"
 #include "util/strings.h"
 #include "util/symbols.h"
@@ -143,7 +144,11 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     }
     case ProbeType::fentry:
     case ProbeType::fexit: {
-      symbol_stream = get_fentry_symbols();
+      if (target == "bpf") {
+        symbol_stream = get_running_bpf_programs();
+      } else {
+        symbol_stream = get_fentry_symbols();
+      }
       break;
     }
     case ProbeType::usdt: {
@@ -248,6 +253,16 @@ std::unique_ptr<std::istream> ProbeMatcher::get_fentry_symbols() const
   else {
     return get_symbols_from_traceable_funcs(true);
   }
+}
+
+std::unique_ptr<std::istream> ProbeMatcher::get_running_bpf_programs() const
+{
+  std::string funcs;
+  auto ids_and_syms = util::get_bpf_progs();
+  for (const auto& [id, symbol] : ids_and_syms) {
+    funcs += "bpf:" + std::to_string(id) + ":" + symbol + "\n";
+  }
+  return std::make_unique<std::istringstream>(funcs);
 }
 
 std::unique_ptr<std::istream> ProbeMatcher::get_raw_tracepoint_symbols() const
@@ -542,14 +557,24 @@ std::set<std::string> ProbeMatcher::get_matches_for_ap(
       search_input = attach_point.func;
       break;
     }
+    case ProbeType::fentry:
+    case ProbeType::fexit: {
+      if (attach_point.target == "bpf") {
+        search_input = "bpf:" +
+                       (attach_point.bpf_prog_id
+                            ? std::to_string(attach_point.bpf_prog_id)
+                            : "*") +
+                       ":" + attach_point.func;
+        break;
+      }
+      [[fallthrough]];
+    }
     case ProbeType::special:
     case ProbeType::uprobe:
     case ProbeType::uretprobe:
     case ProbeType::watchpoint:
     case ProbeType::asyncwatchpoint:
-    case ProbeType::tracepoint:
-    case ProbeType::fentry:
-    case ProbeType::fexit: {
+    case ProbeType::tracepoint: {
       // Do not expand "target:" as that would match all functions in target.
       // This may occur when an absolute address is given instead of a function.
       if (attach_point.func.empty())
--- a/src/util/bpf_progs.cpp
+++ b/src/util/bpf_progs.cpp
@@ -1,0 +1,138 @@
+#include <bpf/bpf.h>
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-qual"
+#include <bpf/btf.h>
+#pragma GCC diagnostic pop
+#include <limits>
+#include <linux/btf.h>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+#include "log.h"
+#include "scopeguard.h"
+#include "util/fd.h"
+
+namespace bpftrace::util {
+
+std::string get_prog_full_name(const struct bpf_prog_info *prog_info,
+                               int prog_fd)
+{
+  const char *prog_name = prog_info->name;
+  const struct btf_type *func_type;
+  struct bpf_func_info finfo = {};
+  struct bpf_prog_info info = {};
+  __u32 info_len = sizeof(info);
+
+  std::string name = std::string(prog_name);
+
+  if (!prog_info->btf_id || prog_info->nr_func_info == 0) {
+    return name;
+  }
+
+  info.nr_func_info = 1;
+  info.func_info_rec_size = prog_info->func_info_rec_size;
+  info.func_info_rec_size = std::min<unsigned long>(info.func_info_rec_size,
+                                                    sizeof(finfo));
+  info.func_info = reinterpret_cast<__u64>(&finfo);
+
+  if (bpf_prog_get_info_by_fd(prog_fd, &info, &info_len)) {
+    return name;
+  }
+
+  struct btf *prog_btf = btf__load_from_kernel_by_id(info.btf_id);
+  if (!prog_btf) {
+    return name;
+  }
+
+  func_type = btf__type_by_id(prog_btf, finfo.type_id);
+  if (!func_type || !btf_is_func(func_type)) {
+    btf__free(prog_btf);
+    return name;
+  }
+
+  prog_name = btf__name_by_offset(prog_btf, func_type->name_off);
+  name = std::string(prog_name);
+  btf__free(prog_btf);
+  return name;
+}
+
+std::vector<std::pair<__u32, std::string>> get_bpf_progs()
+{
+  std::vector<std::pair<__u32, std::string>> ids_and_syms;
+  __u32 id = 0;
+  while (bpf_prog_get_next_id(id, &id) == 0) {
+    int raw_fd = bpf_prog_get_fd_by_id(id);
+
+    if (raw_fd < 0) {
+      continue;
+    }
+
+    auto fd = FD(raw_fd);
+
+    struct bpf_prog_info info = {};
+    __u32 info_len = sizeof(info);
+
+    if (bpf_obj_get_info_by_fd(fd, &info, &info_len) != 0) {
+      continue;
+    }
+
+    if (!info.btf_id) {
+      // BPF programs that don't have a BTF id won't load
+      continue;
+    }
+
+    ids_and_syms.emplace_back(id, get_prog_full_name(&info, fd));
+
+    // Now let's look at the subprograms for this program
+    // if they exist
+    if (info.nr_func_info == 0) {
+      continue;
+    }
+
+    size_t nr_func_info = info.nr_func_info;
+    size_t rec_size = info.func_info_rec_size;
+
+    if (rec_size > std::numeric_limits<std::size_t>::max() / nr_func_info) {
+      // This shouldn't happen
+      continue;
+    }
+
+    std::vector<char> fi_mem(nr_func_info * rec_size);
+
+    struct btf *btf = btf__load_from_kernel_by_id(info.btf_id);
+    if (!btf) {
+      continue;
+    }
+
+    SCOPE_EXIT
+    {
+      btf__free(btf);
+    };
+
+    info = {};
+    info.nr_func_info = nr_func_info;
+    info.func_info_rec_size = rec_size;
+    info.func_info = reinterpret_cast<__u64>(fi_mem.data());
+
+    if (bpf_prog_get_info_by_fd(fd, &info, &info_len) != 0) {
+      continue;
+    }
+
+    auto *func_info = reinterpret_cast<struct bpf_func_info *>(fi_mem.data());
+
+    for (__u32 i = 0; i < nr_func_info; i++) {
+      const struct btf_type *t = btf__type_by_id(btf, (func_info + i)->type_id);
+      if (!t) {
+        continue;
+      }
+
+      const char *func_name = btf__name_by_offset(btf, t->name_off);
+      ids_and_syms.emplace_back(id, std::string(func_name));
+    }
+  }
+
+  return ids_and_syms;
+}
+
+} // namespace bpftrace::util
--- a/src/util/bpf_progs.h
+++ b/src/util/bpf_progs.h
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace bpftrace::util {
+
+// This includes all BPF programs and subprograms
+// the pair is (prog id, symbol)
+std::vector<std::pair<__u32, std::string>> get_bpf_progs();
+
+} // namespace bpftrace::util
--- a/src/util/fd.h
+++ b/src/util/fd.h
@@ -43,8 +43,7 @@ public:
 
   operator int() const noexcept
   {
-    assert(fd_ >= 0);
-    return fd_;
+    return get();
   }
 
 private:
--- a/tests/runtime/probe
+++ b/tests/runtime/probe
@@ -615,3 +615,12 @@ EXPECT WARNING: Unable to attach probe: tracepoint:syscalls:nonsense. Skipping.
 EXPECT WARNING: Unable to attach probe: uprobe:./testprogs/uprobe_test:uprobeFunction3. Skipping.
 EXPECT ERROR: Attachment failed for all probes.
 WILL_FAIL
+
+# This is hacky because we can't easily run a separate bpftrace program in BEFORE
+# because it would require more hacks in the runtime runner.py
+# including waiting for probe attachment in the BEFORE program
+NAME attach_to_bpf_program
+RUN {{BPFTRACE}} runtime/scripts/bpf_attach.bt & sleep 3 && {{BPFTRACE}} -e 'fentry:bpf:interval_us_* { print(probe); if (@c >= 1) { exit(); } @c = count(); }'
+EXPECT_REGEX fentry:bpf:[0-9]+:interval_us_100000_1
+EXPECT_REGEX fentry:bpf:[0-9]+:interval_us_100000_2
+TIMEOUT 10
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,8 @@ and this project adheres to
   - [#4345](https://github.com/bpftrace/bpftrace/pull/4345)
 - Introduce `tseries` for capturing time series data
   - [#3838](https://github.com/bpftrace/bpftrace/pull/3838)
+- Add ability to attach to running BPF programs and sub-programs via `fentry:bpf:prog_name` or `fentry:bpf:prog_id:prog_name`
+  - [#4354](https://github.com/bpftrace/bpftrace/pull/4354)
 #### Changed
 - kprobe: support verbose mode listing
   - [#4362](https://github.com/bpftrace/bpftrace/pull/4362)
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -5,6 +5,7 @@
 #include "ast/context.h"
 #include "attached_probe.h"
 #include "log.h"
+#include "util/int_parser.h"
 #include "util/strings.h"
 
 namespace bpftrace::ast {
@@ -182,10 +183,24 @@ AttachPoint &AttachPoint::create_expansion_copy(ASTContext &ctx,
       if (match.find(":") != std::string::npos)
         ap.target = util::erase_prefix(ap.func);
       break;
+    case ProbeType::fentry:
+    case ProbeType::fexit: {
+      if (match.starts_with("bpf:")) {
+        auto parts = util::split_string(match, ':');
+        ap.target = parts[0];
+        auto prog_id = util::to_uint(parts[1]);
+        if (!prog_id) {
+          LOG(BUG) << "Invalid bpf prog id: " << parts[1];
+        } else {
+          ap.bpf_prog_id = *prog_id;
+        }
+        ap.func = parts[2];
+        break;
+      }
+      [[fallthrough]];
+    }
     case ProbeType::uprobe:
     case ProbeType::uretprobe:
-    case ProbeType::fentry:
-    case ProbeType::fexit:
     case ProbeType::tracepoint:
     case ProbeType::rawtracepoint:
       // Tracepoint, raw tracepoint, uprobe, and fentry/fexit probes specify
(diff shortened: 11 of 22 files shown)


src/bpfprogram.cpp
```diff
@@ -81,17 +81,31 @@ void BpfProgram::set_attach_target(const Probe &probe,
...
+      err_msg = "No BPF program found with name: " + fun + ".";
+    }
+  } else if (btf.get_btf_id(btf_fun, mod, btf_kind) < 0) {
+    err_msg = "No BTF found for " + attach_target + ".";
+  }
+
+  if (!err_msg.empty()) {
     if (config.missing_probes == ConfigMissingProbes::error) {
-      LOG(ERROR) << msg;
+      LOG(ERROR) << err_msg;
```
amscanne, line 103: This is a bit weird. We don't propagate the error up the stack? Can we fix this to return `Result`? Then the `main` should look at the intended outcome. (Doesn't need to be in this commit.)

src/bpfprogram.cpp
```diff
@@ -81,17 +81,31 @@ void BpfProgram::set_attach_target(const Probe &probe,
...
     } else if (config.missing_probes == ConfigMissingProbes::warn) {
-      LOG(WARNING) << msg;
+      LOG(WARNING) << err_msg;
     }
     bpf_program__set_autoload(bpf_prog_, false);
   }
 
-  bpf_program__set_attach_target(bpf_prog_, 0, attach_target.c_str());
+  bpf_program__set_attach_target(bpf_prog_,
```
amscanne, line 110: How do we call this when an error has occurred the lookup? We set the attach fd to **zero** in this case? I think it would be best just to fix the control flow here.

src/bpfprogram.h
```diff
@@ -35,6 +41,8 @@ class BpfProgram {
 
 private:
   struct bpf_program *bpf_prog_;
+  // This fd is specifically for attaching to other running BPF programs.
+  int attach_fd_ = -1;
```
amscanne: Aside: It would be ideal if we had an auto-releasing FD in util. I'd say about half our SCOPEGUARD use is just this case, and it would be a pretty simple thing.

src/btf.cpp
```diff
@@ -970,6 +1013,181 @@
...
+  __u32 info_len = sizeof(info);
+  while (bpf_prog_get_next_id(id, &id) == 0) {
+    int fd = bpf_prog_get_fd_by_id(id);
+    // Don't close the fd if we return it;
+    // it needs to stay alive until after program attaching/loading
+    bool close_fd = true;
+
+    SCOPE_EXIT
+    {
+      memset(&info, 0, sizeof(info));
```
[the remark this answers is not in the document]
amscanne: Why are you bothering to clear the info at all?

src/btf.cpp
```diff
@@ -970,6 +1013,181 @@ int BTF::get_btf_id(std::string_view func,
...
+int BTF::get_fd_for_bpf_prog(const std::string &bpf_prog_name)
+{
+  struct bpf_prog_info info = {};
+  __u32 id = 0;
+  __u32 info_len = sizeof(info);
+  while (bpf_prog_get_next_id(id, &id) == 0) {
+    int fd = bpf_prog_get_fd_by_id(id);
+    // Don't close the fd if we return it;
+    // it needs to stay alive until after program attaching/loading
+    bool close_fd = true;
```
amscanne: Mirroring my other comment, if we had a move-only `Fd` class this would eliminate a lot of code like this.

src/btf.cpp
```diff
@@ -970,6 +1013,181 @@
...
+    // Now let's look at the subprograms for this program
+    // if they exist
+    if (info.nr_func_info == 0) {
+      continue;
+    }
+
+    struct btf *btf = nullptr;
+    auto nr_func_info = info.nr_func_info;
+    auto rec_size = info.func_info_rec_size;
+    void *fi_mem = malloc(nr_func_info * rec_size);
```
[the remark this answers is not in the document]
amscanne: Manual memory management like this with `malloc` and `free` in C++ is a red flag IMO. Just add the right minimal abstractions. You may think it's a simple case, but you have a leak here with the path on line 1069 already.

src/btf.cpp
```diff
@@ -970,6 +1013,181 @@ int BTF::get_btf_id(std::string_view func,
   return -1;
 }
 
+int BTF::get_fd_for_bpf_prog(const std::string &bpf_prog_name)
```
amscanne: I would recommend splitting the whole `bpg_prog_...` bits into their own class. Perhaps they are static members of `BpgProgram`. They are not BTF related.

*Then* you have a `type_id` that you can look up in BTF, and I think you already have the methods you need for that in the existing BTF object.

src/btf.cpp
```diff
@@ -46,6 +46,49 @@ __u32 BTF::start_id(const struct btf *btf) const
...
+    return name;
+  }
+
+  info.nr_func_info = 1;
+  info.func_info_rec_size = prog_info->func_info_rec_size;
+  info.func_info_rec_size = std::min<unsigned long>(info.func_info_rec_size,
+                                                    sizeof(finfo));
+  info.func_info = reinterpret_cast<__u64>(&finfo);
+
+  if (bpf_prog_get_info_by_fd(prog_fd, &info, &info_len)) {
```
amscanne: Aside: It's definitely strange to me that we have an abstraction over BTF (e.g. the `BTF` class), but we just use the raw objects in this case instead. But I understand why you've done this, because the current class does not really give you access to what you need. An option would be to merge the new BTF library, and just add a `from_id` static constructor and then everything gets cleaned up, you can query by ID and use the types, etc.

src/bpfprogram.h
jordalgo: Agreed. Let me look at doing that in a follow up.

src/btf.cpp
jordalgo: > Perhaps they are static members of BpfProgram.

I want these methods together as they do mostly the same thing. One is accessed from BpfProgram and one from ProbeMatcher. They can be a utility class. I just thought they semi-fit here but I don't feel strongly - I'll move them.

src/bpfprogram.cpp
jordalgo: Sure, let me look into fixing that.

Edit: Actually this will take a lot of reworking so going to save it for a follow-up PR.

src/bpfprogram.cpp
jordalgo: We set it to 0 and set the autoload to false. In terms of the control flow, I'll test a bit but pretty sure this is what we want to happen.

src/bpfprogram.h
amscanne: Zero is a valid FD though, which is why you have this separate `close_fd` bool. (This is despite initializing the FD to -1, which is a distinct convention and it's easy to imagine someone making a change in the future which assumes *that* the is convention you're using).

For example, if this class instead had a `std::optional<Fd>` and that was a move-only type that automatically closed on destruction, you'd have no invalid states (impossible to hold it wrong).

amscanne: Wrong place -- I intended to reply to the "we set it to zero" above. Apologies.

src/bpfprogram.cpp
amscanne: Is there a reason we can't do the reworking required for the feature first?

I ask because unless there's overriding urgency, I believe that this should be the default mode. Because while I believe you will follow up, reworking has a habit of not happening. (I believe this pattern is pretty common in bpftrace, and as a leader in the project I think you should be held to the highest standards ;)

Anyways, I'm not intending to block anything and can do another review w/ approval if you want. But I don't think the refactoring/reworking is huge here, so I think it's a great opportunity to exercise the Boy Scout rule while delivering a nice feature!

src/bpfprogram.cpp
jordalgo: Actually yeah, we don't need to call `bpf_program__set_attach_target` in the failure case. Just need to set autoload to false.

src/bpfprogram.cpp
jordalgo: > so I think it's a great opportunity to exercise the Boy Scout rule while delivering a nice feature!

Stop trying to butter me up in order to convince me to shave some yaks! Fine Fine. I'll look at fixing this failure case up first.

src/bpfprogram.h
```diff
@@ -35,6 +36,8 @@ class BpfProgram {
 
 private:
   struct bpf_program *bpf_prog_;
+  // This fd is specifically for attaching to other running BPF programs.
+  util::FD attach_fd_ = util::FD(-1);
```
amscanne: The `-1` here is a leaky abstraction (knowledge of the internal `FD` implementation). Just make this type `std::optional<FD> attach_fd_` and then there is no valid state. You can just do things like:

```
attach_fd_.emplace(fd);
...
attach_fd_ ? attach_fd_->get() : 0 /* libbpf weirdness */
```

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
+#include <string>
+#include <unistd.h>
+#include <unordered_set>
+#include <vector>
+
+#include "log.h"
```
amscanne, line 12: Please keep the header blocks tidy:

```
<external headers>

<internal headers>
```

This is mixing both together.

(This is just the most common internal convention and I cleaned up most files to be consistent. As long as it's consistent I think it's easier to read.)

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+      }
+
+      const char *func_name = btf__name_by_offset(btf, t->name_off);
+      if (std::string(func_name) == bpf_prog_name) {
+        return fd;
+      }
+    }
+  }
+
+  return FD(-1);
```
amscanne: This is not using the type as designed. You're getting back a valid `FD` object and need to reach it (through a leaky abstraction) to see whether it is valid. Just make this function return a `Result<FD>`, and callers will be forced to check if it's correct. Then you can also record errors instead of logging. E.g. I could imagine something like:

```
class BpfProgError : public ErrorInfo<BpfProgError> {
   BpfProgError(std::vector<std::string> &&errs) : errs_(std::move(errs)) {};
   ...
   void log(llvm::raw_ostream &OS) {
       if (errs_.size() == 1) {
            OS << errs_[0];
       } else {
           OS << "multiple errors occured: " << ...
       }
    }
...

Result<FD> get_fd_for_bpf_prog(...) {
   std::vector<std::string> errs;
   ...
     if (error) {
        errs.push_back("Failed to load BTF for BPF program ID " + str::to_string(info.btf_id));
        continue;
     }
     ...
     return FD(fd);
   ...
   return make_error<BpfProgError>(std::move(errs));
}
```

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+std::unordered_set<std::string> get_bpf_program_symbols()
+{
+  std::unordered_set<std::string> symbols;
+  __u32 id = 0;
+  while (bpf_prog_get_next_id(id, &id) == 0) {
+    auto fd = FD(bpf_prog_get_fd_by_id(id));
+
+    if (fd < 0) {
```
amscanne: Same complaint re: leaky abstraction. Why does this object's contract have this? If I have a valid FD object, it should be an FD not *maybe* an FD :)

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+    // Now let's look at the subprograms for this program
+    // if they exist
+    if (info.nr_func_info == 0) {
+      continue;
+    }
+
+    struct btf *btf = nullptr;
+    auto nr_func_info = info.nr_func_info;
+    auto rec_size = info.func_info_rec_size;
+    std::vector<char> fi_mem(nr_func_info * rec_size);
```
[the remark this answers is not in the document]
amscanne: Please check that the size isn't unreasonable before this call, just as a sanity check.

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    btf = btf__load_from_kernel_by_id(info.btf_id);
```
amscanne: Why not use `struct btf *btf = btf__load...` here? I had to go up looking for why it was declared before, but there is no reason as far as I can tell.

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+    info = {};
+    info.nr_func_info = nr_func_info;
+    info.func_info_rec_size = rec_size;
+    info.func_info = reinterpret_cast<__u64>(fi_mem.data());
+
+    if (bpf_prog_get_info_by_fd(fd, &info, &info_len) != 0) {
+      continue;
+    }
+
+    auto *func_info = reinterpret_cast<struct bpf_func_info *>(info.func_info);
```
amscanne: This is not a variable-sized structure. Do we need to read the whole thing?

I think it would be better to reinterpret `fi_mem.data()` here, since presumably it is not subject to the call but this is a minor thing.

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+    if (!func_info) {
```
amscanne: Can this happen? The kernel zeros out this field if it is invalid? Just asking re: my question about for `fi_mem.data()`.

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,207 @@
...
+      continue;
+    }
+
+    const struct btf_type *t = nullptr;
```
amscanne: Any reason to not declare this in the loop scope below? It just reduces the possibility of a persistent `t` and mutation of the value.

src/util/bpf_progs.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "util/fd.h"
```
amscanne: Nit: header order.

src/bpfprogram.cpp
```diff
@@ -81,17 +82,31 @@ void BpfProgram::set_attach_target(const Probe &probe,
...
-  if (btf.get_btf_id(btf_fun, mod, btf_kind) < 0) {
-    const std::string msg = "No BTF found for " + attach_target + ".";
+  std::string err_msg;
+
+  if ((probe.type == ProbeType::fentry || probe.type == ProbeType::fexit) &&
+      mod == "bpf") {
+    attach_target = fun;
+    attach_fd_ = std::move(util::get_fd_for_bpf_prog(fun));
+    if (attach_fd_ < 0) {
+      err_msg = "No BPF program found with name: " + fun + ".";
```
amscanne: I still think you should just make this function return an error, instead of potentially logging something and just returning `void`. It just feels like an anti-pattern to me, but I won't block for this one thing (most of the rest of my complaints are much improved).

src/util/bpf_progs.cpp
jordalgo: I think a Result in this file over-complicates things. Any of these errors where we `continue` would be quite unexpected and not something the user would be able to fix. Also this particular function can not return the `FD` and that's valid so I think it should just be an `optional<FD>`.

src/util/bpf_progs.cpp
jordalgo: Hmm, let me try that.

src/util/bpf_progs.cpp
jordalgo: I'm not sure tbh...

src/bpfprogram.cpp
jordalgo: Working on a refactor: https://github.com/jordalgo/bpftrace/commit/d112b53c5017c168c3aff3dd3faabafe747bc9f5

src/util/bpf_progs.cpp
jordalgo: Ack. Wish the formatter just did this automatically.

man/adoc/bpftrace.adoc
```diff
@@ -1476,6 +1478,9 @@ This removes the need for manual type casting and makes the code more resilient
 The function arguments are available in the `args` struct which can be inspected by doing verbose listing (see <<Listing Probes>>).
 These arguments are also available in the return probe (`fexit`), unlike `kretprobe`.
 
+The last variant `fentry:bpf:prog_name` allows attaching to running BPF programs and sub programs.
```
jrife: Just curious, so I have a few questions. 

1. What happens if there are multiple BPF programs with the same name? Does it attach to all of them? Is there any way to select a BPF program by its ID instead? I imagine it could be useful to reference a particular program by ID or pin path similar to bpftool. 
2. Can this attach to any BPF program even if its attach type is different? E.g., could I attach a tracing program to a `BPF_PROG_TYPE_SCHED_CLS` program which itself is attached to a network interface? Can I inspect the context of the second program this way?

jrife (commented):
This seems cool! I just have a few questions.

man/adoc/bpftrace.adoc
jordalgo: > What happens if there are multiple BPF programs with the same name? Does it attach to all of them? 

Good question. At the moment, it will attach to the first one that matches. It's probably better to attach to all of them but I'd have to think about how that would work.

> Is there any way to select a BPF program by its ID instead? 

This is a great idea. Let me look into making this possible. Seems very reasonable and do-able.

> Can this attach to any BPF program even if its attach type is different?

I believe so. I don't see any limitations on the type of BPF program you can attach to (but maybe there is some fine print somewhere).

> Can I inspect the context of the second program this way?

Ideally yes, but I think this is tricky and I'm still thinking of how this will work.

src/bpfprogram.h
jordalgo: Done.

src/util/bpf_progs.cpp
amscanne: I acquiesce, but I think logging to global singletons is not a great pattern in general. We do have open issues about how we silence test log spam, etc. However, I do think `std::optional<FD>` is acceptable ;)

src/util/bpf_progs.cpp
amscanne: It's weird. If you just use your original cast data then I don't think you need to re-validate this field.

jordalgo: Probably not. I'm always nervous around raw pointers 😨

man/adoc/bpftrace.adoc
jordalgo: @jrife I added the ability to attach to BPF programs via their program id. Thanks for the suggestion. I also discovered that you can't attach to some BPF programs (they need to have a BTF id).

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,254 @@
...
+      continue;
+    }
+
+    if (!info.btf_id) {
+      // BPF programs that don't have a BTF id won't load
+      continue;
+    }
+
+    if (get_prog_full_name(&info, fd) == bpf_prog_name) {
+      return fd;
```
jrife: This seems like it could get confusing. A common scenario for me in working with Cilium is to have several programs with the same name on the same host, sometimes many with the same name. Trying to imagine how I might use this, I'd either want to add some tracing to *all* of them or one of them. The selection by ID is good if I really wanted to target a specific instance of a program, but if I'm using a name then either:

1. I really don't care which one and just want one to be selected automatically without me needing to look up an ID beforehand.
2. I want to do some tracing on all of them.

In the first scenario, I'd probably want some way to identify which program has been selected after the atachment, in which case I'd probably want some builtin like `bpfid`: `printf("BPF program ID is %d\n", bpfid);`

In the second scenario, I'd want attachment to all matching programs and still probably want the ability to differentiate between instances of programs with the same name by ID for, e.g., histogram keys.

man/adoc/bpftrace.adoc
```diff
@@ -1476,6 +1480,9 @@ This removes the need for manual type casting and makes the code more resilient
 The function arguments are available in the `args` struct which can be inspected by doing verbose listing (see <<Listing Probes>>).
 These arguments are also available in the return probe (`fexit`), unlike `kretprobe`.
 
+The bpf variants (e.g. `fentry:bpf:prog_name` and `fentry:bpf:prod_id`) allow attaching to running BPF programs and sub programs.
```
jrife: `prod_id` -> `prog_id`

man/adoc/bpftrace.adoc
jrife: @jordalgo Looks great. Thanks for adding this. Left a few more comments after some thought.

jrife:
Also thinking out loud, since this is targeting tracing of individual BPF programs, I wonder if there might be a use case for working with existing maps. This is definitely beyond the scope of this change, but it seems like it could be useful to bind a map name in bpftrace to an existing map and have the type information inferred somehow so you could inspect the map state, etc.

src/util/bpf_progs.cpp
jordalgo: > This seems like it could get confusing. A common scenario for me in working with Cilium is to have several programs with the same name on the same host

Thanks for calling this out. I didn't realize this was a common use case. The latest version of this PR now supports attaching to multiple BPF programs of the same name (as a default if you only supply the name). However you can also specify the name and prog id or just the prog id to just attach to one.

Also, you can use the `probe` builtin to see which of the different (same named) BPF programs you're hitting as this include the name and id. We can probably add a `bpfid` later on if needed but `probe` should be an ok start.

jordalgo:
> Also thinking out loud, since this is targeting tracing of individual BPF programs, I wonder if there might be a use case for working with existing maps. 

Read my mind. I think we'll be able to do this soon now that @amscanne is working on a better BTF internal interface. I'll make a task to look into this.

src/util/bpf_progs.cpp
jrife: > Also, you can use the probe builtin to see which of the different (same named) BPF programs you're hitting as this include the name and id. We can probably add a bpfid later on if needed but probe should be an ok start.

I see. This would probably work fine then.

man/adoc/bpftrace.adoc
```diff
@@ -1461,6 +1463,11 @@ This removes the need for manual type casting and makes the code more resilient
 The function arguments are available in the `args` struct which can be inspected by doing verbose listing (see <<Listing Probes>>).
 These arguments are also available in the return probe (`fexit`), unlike `kretprobe`.
 
+The bpf variants (e.g. `fentry:bpf[:prog_name]:prog_id`) allow attaching to running BPF programs and sub programs.
+You can specify just the program name and in this case bpftrace will attach to all running programs with that name (differentiate between them using the `probe` builtin).
```
jrife: small nit: `program name and` -> `program name, and`.

man/adoc/bpftrace.adoc
```diff
@@ -1461,6 +1463,11 @@ This removes the need for manual type casting and makes the code more resilient
 The function arguments are available in the `args` struct which can be inspected by doing verbose listing (see <<Listing Probes>>).
 These arguments are also available in the return probe (`fexit`), unlike `kretprobe`.
 
+The bpf variants (e.g. `fentry:bpf[:prog_name]:prog_id`) allow attaching to running BPF programs and sub programs.
```
viktormalik: Nit: subprogram is a single word, isn't it?
```suggestion
The bpf variants (e.g. `fentry:bpf[:prog_name]:prog_id`) allow attaching to running BPF programs and subprograms.
```
or "sub-program"

Same below.

src/ast/attachpoint_parser.cpp
```diff
@@ -655,11 +655,45 @@ AttachPointParser::State AttachPointParser::watchpoint_parser(bool async)
...
+    ap_->target = "bpf";
+    if (parts_.size() == 2) {
+      if (listing_) {
+        ap_->func = "*";
+        return OK;
+      } else {
+        errs_ << "the 'bpf' variant of this probe requires a bpf program name, "
+                 "a bpf program id, or both ";
+        return INVALID;
+      }
```
viktormalik: I think that we should have a unified behavior for both listing and non-listing here. So, either always require an explicit wildcard (`fentry:bpf:*`) or insert it implicitly. I'd go with the former as it's consistent with other probe types as well as with the documentation above.

man/adoc/bpftrace.adoc
viktormalik: Also we should clarify here what "attaching to a running BPF program" means from user perspective as not all users may be familiar with this concept.

man/adoc/bpftrace.adoc
```diff
@@ -1441,6 +1441,8 @@ iter:task_file:/sys/fs/bpf/files {
 .variants
 * `fentry[:module]:fn`
 * `fexit[:module]:fn`
+* `fentry:bpf[:prog_name]:prog_id`
+* `fexit:bpf[:prog_name]:prog_id`
```
viktormalik: Is there any difference between `fentry` and `fexit`? We should clarify it here.

src/ast/attachpoint_parser.cpp
```diff
@@ -655,11 +655,45 @@ AttachPointParser::State AttachPointParser::watchpoint_parser(bool async)
...
+        ap_->func = "*:" + parts_[2];
+      } else {
+        ap_->func = parts_[2] + ":*";
+      }
+    } else {
+      if (!util::to_uint(parts_[3])) {
+        errs_ << "bpf program id must be a number";
+        return INVALID;
+      }
+      ap_->func = parts_[2] + ":" + parts_[3];
```
viktormalik: I'm not a big fan of this, the id should be parsed here. We already have a number of probe types which have more than 3 parts and a there's a couple of integer fields in `ast::AttachPoint` which could be used (e.g. `ast::AttachPoint::address`).

src/probe_matcher.cpp
```diff
@@ -143,6 +144,10 @@ std::set<std::string> ProbeMatcher::get_matches_for_probetype(
     }
     case ProbeType::fentry:
     case ProbeType::fexit: {
+      if (target == "bpf") {
+        symbol_stream = get_running_bpf_program_symbols();
+        break;
+      }
       symbol_stream = get_fentry_symbols();
       break;
```
viktormalik: How about
```suggestion
      if (target == "bpf")
        symbol_stream = get_running_bpf_program_symbols();
      else
        symbol_stream = get_fentry_symbols();
      break;
```

src/util/fd.h
```diff
@@ -0,0 +1,62 @@
...
+  int get() const noexcept
+  {
+    assert(fd_ >= 0);
+    return fd_;
+  }
+
+  operator int() const noexcept
+  {
+    assert(fd_ >= 0);
+    return fd_;
```
viktormalik: Use `get` here?
```suggestion
  return get();
```
Also do we need two same methods?

tests/runtime/probe
```diff
@@ -615,3 +615,11 @@ EXPECT WARNING: Unable to attach probe: tracepoint:syscalls:nonsense. Skipping.
 EXPECT WARNING: Unable to attach probe: uprobe:./testprogs/uprobe_test:uprobeFunction3. Skipping.
 EXPECT ERROR: Attachment failed for all probes.
 WILL_FAIL
+
+# This is hacky because we can't easily run a separate bpftrace program in BEFORE
+# because it would require more hacks in the runtime runner.py
+# including waiting for probe attachment in the BEFORE program
+NAME attach_to_bpf_program
+RUN {{BPFTRACE}} runtime/scripts/bpf_attach.bt & sleep 3 && {{BPFTRACE}} -e 'fentry:bpf:interval_us_* { print(1); exit(); }'
+EXPECT 1
```
viktormalik: If we did `print(probe)` in the program, we could check that each prog was correctly attached to.

src/bpfprogram.cpp
```diff
@@ -81,17 +84,43 @@ void BpfProgram::set_attach_target(const Probe &probe,
...
+      LOG(BUG) << "BPF function should contain a colon: " << fun;
+    }
+    attach_target = pair[0];
+    auto prog_id = util::to_uint(pair[1]);
+    if (!prog_id) {
+      // This shouldn't happen
+      err_msg = "Invalid BPF program id: " + pair[1];
+    } else {
+      attach_fd_ = util::get_fd_for_bpf_prog(pair[0],
+                                             static_cast<__u32>(*prog_id));
```
viktormalik: Why do we need to pass both the name and the id to the function? Shouldn't id be unique?

src/util/bpf_progs.cpp
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  return std::nullopt;
+}
+
+std::unordered_set<std::string> get_bpf_program_symbols()
+{
+  std::unordered_set<std::string> symbols;
+  __u32 id = 0;
+  while (bpf_prog_get_next_id(id, &id) == 0) {
```
viktormalik, line 64: A big portion of this loop just duplicates the loop from `get_fd_from_bpf_prog`. I wish there was a way to simplify this.

src/bpfprogram.cpp
viktormalik: So, IIUC, when we attach to multiple progs using wildcard, we will:
1. Iterate all progs using `get_bpf_program_symbols` and do full expansion (i.e. one BPF program per match).
2. For each generated BPF program, iterate all progs to find its fd.
That's some quadratic complexity for something that could be done in a linear way. Could this be a performance problem on systems with many BPF programs running? Perhaps we should have some id->fd caching?

viktormalik (commented):
A very nice functionality!

There's a few things that could be cleaned up and some questions from my side.

man/adoc/bpftrace.adoc
jordalgo: It's probably a gray area. I'm good with the dash.

src/ast/attachpoint_parser.cpp
jordalgo: Ack

man/adoc/bpftrace.adoc
jordalgo: I think it's the same as the differences between normal fentry/fexit. What were you thinking?

src/ast/attachpoint_parser.cpp
jordalgo: hmm, I'm on the fence. I'm not a big fan of overloading `address` here as this BPF program id is something totally different. I'd rather just bite the bullet and add a new field.

src/util/fd.h
jordalgo: > Also do we need two same methods?

I think it's nice to provide both: an explicit getter and then another for auto conversion to int but it's a minor detail either way.

tests/runtime/probe
jordalgo: I like it.

src/bpfprogram.cpp
jordalgo: Good point. Will fix.

Actually this allows me to remove most of the code in `get_fd_for_bpf_prog` so there is no more duplicated logic.

Edit: removed `get_fd_for_bpf_prog`

src/util/bpf_progs.cpp
jordalgo: see above.

src/bpfprogram.cpp
jordalgo: You're right in that it's not the most efficient but even on Meta servers that run A LOT of bpf programs, we're talking about numbers below 200 so I think a caching solution at this point would be a premature optimization. Also, after your suggestion above, `get_fd_for_bpf_prog` is now much simpler and just iterates looking for the id (nothing else).

Edit: Sorry, I realized I didn't need `get_fd_for_bpf_prog` at all and I can just look up the FD directly. So now we're not looping at all when we attach.

man/adoc/bpftrace.adoc
jordalgo: I added an example - let me know if that helps.

man/adoc/bpftrace.adoc
viktormalik: So, the `fentry` will execute before the target BPF program and `fexit` will execute after it? That's what I would expect so I'm just checking if that's the case.

I'm mostly asking because the below docs always say that this probe will execute **when** the target BPF program executes but in some cases the ordering matters.

man/adoc/bpftrace.adoc
viktormalik: Yes, thank you.

src/bpfprogram.cpp
```diff
@@ -81,17 +85,44 @@ void BpfProgram::set_attach_target(const Probe &probe,
     btf_fun = fun;
   }
 
-  if (btf.get_btf_id(btf_fun, mod, btf_kind) < 0) {
-    const std::string msg = "No BTF found for " + attach_target + ".";
+  std::string err_msg;
+
+  if ((probe.type == ProbeType::fentry || probe.type == ProbeType::fexit) &&
+      mod == "bpf") {
```
viktormalik, line 88: Hopefully no one comes with a kernel module called `bpf`, that will break this :laughing:

src/bpfprogram.cpp
```diff
@@ -81,17 +85,44 @@ void BpfProgram::set_attach_target(const Probe &probe,
...
+    auto pair = util::split_string(fun, ':');
```
viktormalik: I'm not going to block on this but I don't like that this relies on always doing full expansion of `fentry:bpf` probes, even when the probe is fully specified (like `fentry:bpf:my_prog:1234`). Ideally, what we should do here is introduce a new field on `bpftrace::Probe` and update it from `BPFtrace::generate_probe` and `AttachPoint::create_expansion_copy`. It's some amount of additional code so I understand why we'd want to avoid it. I'm just pointing this out b/c if someone decides to not do expansion for the explicit case in future, the implementation will break.

src/bpfprogram.cpp
```diff
@@ -81,17 +85,44 @@ void BpfProgram::set_attach_target(const Probe &probe,
...
+    if (pair.size() != 2) {
+      LOG(BUG) << "BPF function should contain a colon: " << fun;
+    }
+    attach_target = pair[0];
+    auto prog_id = util::to_uint(pair[1]);
+    if (!prog_id) {
+      // This shouldn't happen
+      err_msg = "Invalid BPF program id: " + pair[1];
```
viktormalik: Issue a `BUG` then?

man/adoc/bpftrace.adoc
jordalgo: > So, the fentry will execute before the target BPF program and fexit will execute after it?

Yes, just did some local testing to confirm.

> I'm mostly asking because the below docs always say that this probe will execute when the target BPF program executes but in some cases the ordering matters.

Let me fix up that language to be less ambiguous.

src/bpfprogram.cpp
jordalgo: That seems unlikely...

src/bpfprogram.cpp
jordalgo: I think we can be more specific but I'm not quite sure if it solves the problem you're describing. We can check if bpf_prog_id is 0 or if there is a wild card in the func name - either of those indicate we need expansion and don't need it when the probe is fully specified.

viktormalik: IMHO, the cause of the problem that I'm describing is that we only read the new `AttachPoint::bpf_prog_id` field in `ProbeMatcher`, i.e. we always have to go through it, otherwise the id is not assigned anywhere. And in `ProbeMatcher`, the id is made a part of `func` so that's why we need to do the split on colon here.

jordalgo: I see what you mean. Yeah I was bothered a bit by that too but wasn't sure if it was better to add a new field to `Probe` that was so specific to BPF program attachment. The other issue is all the custom code you'd need in `ProbeMatcher` to pass around the found `bpf_prog_id` when you just have a program or sub-program name. TBH, I thought it was kinda nice (except for this part) that you could just make it part of the function name and everything else just worked.

src/bpfprogram.cpp
jordalgo: If it makes you feel better it lists `[bpf]` as the module already in kallsyms e.g.
```
ffffffffa00019b4 t bpf_prog_6deef7357e7b4530_sd_fw_ingress	[bpf]
```

viktormalik: Nice, that's good to know.

src/bpfprogram.cpp
viktormalik: Yeah, that's why I say that it may not be worth doing it. Although I'm not sure how well this will play along with my refactoring of probe expansion (it probably won't) but I can work on that later.

jordalgo: @viktormalik My apologies, I got completely confused between the `ast::Probe` and the `struct Probe`. I added a `bpf_prog_id` to `struct Probe` and now the only place we have to do the string split is in `create_expansion_copy`. Let me know if this is what you had in mind.

jordalgo:
@viktormalik I also made a small change because I realized I had the probe argument ordering wrong. It should be `fentry:bpf:[:prog_id]:prog_name` this is because the prog_id is the higher level "bucket". You can still attach to multiple bpf programs and sub-programs across multiple running prog ids but now the listing makes a lot more sense. It's grouped by program id instead of by function name e.g.
```
bpftrace -l 'fentry:bpf:*'
fentry:bpf:4217779:dnswatch_kprobe_udp_recvmsg
fentry:bpf:4217779:populate_cmdline
fentry:bpf:4217780:armr_env_get_len
fentry:bpf:4217780:armr_get_cgroup_from_task
```

src/bpfprogram.cpp
viktormalik: Yes, this is exactly what I had in mind (sorry if I wasn't clear enough). It looks much better now, thanks!

viktormalik:
> @viktormalik I also made a small change because I realized I had the probe argument ordering wrong. It should be `fentry:bpf:[:prog_id]:prog_name` this is because the prog_id is the higher level "bucket". You can still attach to multiple bpf programs and sub-programs across multiple running prog ids but now the listing makes a lot more sense. It's grouped by program id instead of by function name e.g.

Good point, I think you're right.

jordalgo:
> It would be nice to add a few more unit tests for the expansion (currently in tests/bpftrace.cpp, will be moved soon).

Done.

PR #4361, opened 2025-07-12
Log the total lost event count

jordalgo:
In high frequency output the single messages
about how many lost events there were can
easily be drowned out. Let's log the total
number of lost events at the end.

Also log it to std:err via `LOG(WARNING)`
so it also doesn't get potentially dropped
if stdout is flooded.

##### Checklist

### DIFF
--- a/tests/runtime/other
+++ b/tests/runtime/other
@@ -229,7 +229,7 @@ EXPECT_REGEX_NONE @
 TIMEOUT 1
 
 NAME sigint under heavy load
-RUN {{BPFTRACE}} --unsafe -e 'tracepoint:sched:sched_switch { system("echo foo"); } END { printf("end"); }'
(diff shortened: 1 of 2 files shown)


tests/runtime/other
```diff
@@ -229,7 +229,7 @@ EXPECT_REGEX_NONE @
 TIMEOUT 1
 
 NAME sigint under heavy load
-RUN {{BPFTRACE}} --unsafe -e 'tracepoint:sched:sched_switch { system("echo foo"); } END { printf("end"); }'
+RUN {{BPFTRACE}} --unsafe -e 'tracepoint:sched:sched_switch { system("echo foo"); } END { print("end"); }'
```
viktormalik, line 232: Why this change?

viktormalik (approved):
Looks reasonable to me.

tests/runtime/other
jordalgo: There is no new line after `printf` so the new added warning was getting placed on the same line and causing this test to fail.

PR #4362, opened 2025-07-13
kprobe: support list verbose (-lv)

Rtoax:
This patch make kprobe/kretprobe list kernel function declaration for verbose, this allows us to write scripts directly without having to look at the Linux kernel source code every time.

For example:

    $ sudo bpftrace -l kprobe:schedule
    kprobe:schedule
    $ sudo bpftrace -l kprobe:do_sys_openat2
    kprobe:do_sys_openat2
    $ sudo bpftrace -l kprobe:vmlinux:do_sys_openat2
    kprobe:vmlinux:do_sys_openat2
    $ sudo bpftrace -l kprobe:kvm:x86_emulate*
    kprobe:kvm:x86_emulate_insn
    kprobe:kvm:x86_emulate_instruction

    # show the verbose of kprobe

    $ sudo bpftrace -lv kprobe:schedule
    kprobe:schedule
        void schedule();
    $ sudo bpftrace -lv kprobe:do_sys_openat2
    kprobe:do_sys_openat2
        int do_sys_openat2(int dfd, const char * filename, struct open_how * how);
    $ sudo bpftrace -lv kprobe:vmlinux:do_sys_openat2
    kprobe:vmlinux:do_sys_openat2
        int vmlinux:do_sys_openat2(int dfd, const char * filename, struct open_how * how);
    $ sudo bpftrace -lv kprobe:kvm:x86_emulate*
    kprobe:kvm:x86_emulate_insn
        int kvm:x86_emulate_insn(struct x86_emulate_ctxt * ctxt);
    kprobe:kvm:x86_emulate_instruction
        int kvm:x86_emulate_instruction(struct kvm_vcpu * vcpu, gpa_t cr2_or_gpa, int emulation_type, void * insn, int insn_len);

It should be noted that, like fentry/fexit, kprobe and kretprobe view the same verbose information, which are both function declarations.


CHANGELOG.md
```diff
@@ -70,6 +70,8 @@ and this project adheres to
 - Introduce `tseries` for capturing time series data
   - [#3838](https://github.com/bpftrace/bpftrace/pull/3838)
 #### Changed
+- kprobe: support list verbose
```
jordalgo: nit
```suggestion
- kprobe: support verbose mode listing
```

src/btf.cpp
```diff
@@ -727,6 +727,91 @@ FuncParamLists BTF::get_params_from_btf(
   return params;
 }
 
+FuncParamLists BTF::get_kprobes_params_from_btf(
```
jordalgo, line 730: naming nit
```suggestion
FuncParamLists BTF::get_kprobe_params_from_btf(
```

src/btf.cpp
```diff
@@ -727,6 +727,91 @@ FuncParamLists BTF::get_params_from_btf(
...
+    }
+
+    // set by dump_printf callback
+    type.str("");
+    if (btf_dump__emit_type_decl(dump, t->type, &decl_opts)) {
+      LOG(ERROR) << "failed to dump return type for: " << func_name;
+      break;
+    }
+
+    params[func_name].push_back(type.str());
```
jordalgo: Not sure how clever we should get here but I think we know of some types that typically won't be in registers and therefore not accessible via the `argN` builtins e.g. double/float

https://github.com/bpftrace/bpftrace/blob/master/man/adoc/bpftrace.adoc#kprobe-and-kretprobe

src/btf.cpp
```diff
@@ -727,6 +727,91 @@ FuncParamLists BTF::get_params_from_btf(
...
+    if (!t)
+      continue;
+
+    BPFTRACE_LIBBPF_OPTS(btf_dump_emit_type_decl_opts,
+                         decl_opts,
+                         .field_name = "");
+
+    const auto *p = btf_params(t);
+    for (__u16 j = 0, len = btf_vlen(t); j < len; j++, p++) {
+      const char *arg_name = btf__name_by_offset(btf_obj.btf, p->name_off);
```
jordalgo: For bpftrace kprobes the arg name doesn't help when accessing arguments perhaps we should just use the same convention as you would to access them e.g. `arg0`, `arg1`, etc.

src/probe_matcher.cpp
```diff
@@ -486,10 +486,14 @@ void ProbeMatcher::list_probes(ast::Program* prog)
           param_lists = get_iters_params(matches);
         else if (probe_type == ProbeType::uprobe)
           param_lists = get_uprobe_params(matches);
+        else if (probe_type == ProbeType::kprobe ||
+                 probe_type == ProbeType::kretprobe)
```
jordalgo:
```suggestion
        else if (probe_type == ProbeType::kprobe)
```
This only works for kprobes

src/probe_matcher.cpp
```diff
@@ -505,12 +509,35 @@ void ProbeMatcher::list_probes(ast::Program* prog)
...
           match_print = target + ":" + ap->lang + ":" + func;
+          func_name = func;
         }
 
         std::cout << probe_type << ":" << match_print << std::endl;
         if (bt_verbose) {
-          for (auto& param : param_lists[match])
-            std::cout << "    " << param << std::endl;
+          if (probe_type == ProbeType::kprobe ||
+              probe_type == ProbeType::kretprobe) {
```
jordalgo:
```suggestion
          if (probe_type == ProbeType::kprobe) {
```

src/probe_matcher.cpp
```diff
@@ -505,12 +509,35 @@ void ProbeMatcher::list_probes(ast::Program* prog)
...
 
           match_print = target + ":" + ap->lang + ":" + func;
+          func_name = func;
         }
 
         std::cout << probe_type << ":" << match_print << std::endl;
         if (bt_verbose) {
-          for (auto& param : param_lists[match])
-            std::cout << "    " << param << std::endl;
+          if (probe_type == ProbeType::kprobe ||
```
jordalgo: I'm a bit confused. Why do we need all this added logic for kprobe params? Maybe for the other probes it's just kept in a separate file/class...

CHANGELOG.md
Rtoax: Fixed, thanks.

src/btf.cpp
Rtoax: Indeed, thanks.

src/probe_matcher.cpp
Rtoax: How about we keep it same as fentry/fexit, like:

```
$ sudo bpftrace -lv fentry:vmlinux:do_sys_openat2
fentry:vmlinux:do_sys_openat2
    int dfd
    const char * filename
    struct open_how * how
    int retval
$ sudo bpftrace -lv fexit:vmlinux:do_sys_openat2
fexit:vmlinux:do_sys_openat2
    int dfd
    const char * filename
    struct open_how * how
    int retval
```

After all, kretprobe also needs to know the return value.

```
$ sudo bpftrace -lv kprobe:do_sys_openat2
kprobe:do_sys_openat2
    int do_sys_openat2(int arg0, const char * arg1, struct open_how * arg2);
$ sudo bpftrace -lv kretprobe:do_sys_openat2
kretprobe:do_sys_openat2
    int do_sys_openat2(int arg0, const char * arg1, struct open_how * arg2);
```

right?

src/btf.cpp
Rtoax: How about we keep it `_kprobes_`, because it's not only one kprobe?

src/probe_matcher.cpp
Rtoax: same as above, please review!

src/probe_matcher.cpp
Rtoax: Indeed, but I think that displaying function declarations for kprobe is more comprehensive. For the consistency of displaying with fentry/tracepoint, it seems that it should indeed be shown in a list format. What do you think?

src/btf.cpp
Rtoax: I try to search in linux source tree, functions with float/double  parameters pass have no btf? 

```
$ sudo bpftrace -lv kprobe:math_*
kprobe:math_ceil
kprobe:math_ceil2
kprobe:math_error
    void math_error(struct pt_regs * arg0, int arg1);
kprobe:math_fabs
kprobe:math_floor
kprobe:math_floor2
kprobe:math_log
kprobe:math_log2
kprobe:math_log2_approx
    unsigned int math_log2_approx(unsigned int arg0);
kprobe:math_max2
kprobe:math_max3
kprobe:math_max4
kprobe:math_max5
kprobe:math_min2
kprobe:math_mod
kprobe:math_pow
kprobe:math_round
```

src/probe_matcher.cpp
jordalgo: > After all, kretprobe also needs to know the return value.

For kretprobe and uretprobe, its type is always uint64. The reason it exists for fexit is because you can access the `args` in an fexit probe. I think for `kretprobe` it just might be misleading to list the function signature.

src/btf.cpp
jordalgo: > functions with float/double parameters pass have no btf?

That's odd. Maybe we should add a comment and/or a check for the Float BTF type anyway. WDYT?

src/btf.cpp
jordalgo: Sure, not a big deal either way.

src/probe_matcher.cpp
jordalgo: Yeah, I think it should be in a list like the other probes for consistency. Here is where you could, though maybe unnecessary, have a verbose mode for kretprobe e.g.
```
$ sudo bpftrace -lv kretprobe:do_sys_openat2
kretprobe:do_sys_openat2
    int retval
```

Rtoax: Done!

src/btf.cpp
Rtoax: I thinks you're right, done, please review.

src/btf.cpp
```diff
@@ -727,6 +727,106 @@
...
+        }
+
+        // Note that floating point arguments are typically passed in special
+        // registers which don’t count as argN arguments. e.g. on x86_64 the
+        // first 6 non-floating point arguments are passed in registers and
+        // all following arguments are passed on the stack
+        const auto *pt = btf__type_by_id(btf_obj.btf, p->type);
+        if (pt && BTF_INFO_KIND(pt->info) == BTF_KIND_FLOAT) {
+          LOG(WARNING) << "BTF: skip a float/double param of " << func_name;
+          j--;
```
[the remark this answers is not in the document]
Rtoax: oops, fixed.

src/btf.cpp
```diff
@@ -727,6 +727,106 @@ FuncParamLists BTF::get_params_from_btf(
...
+    const BTFObj &btf_obj,
+    const std::set<std::string> &funcs,
+    const ProbeType &probe_type) const
+{
+  std::stringstream type;
+  auto *dump = dump_new(btf_obj.btf, dump_printf, &type);
+  if (auto err = libbpf_get_error(dump)) {
+    char err_buf[256] = {};
+    libbpf_strerror(err, err_buf, sizeof(err_buf));
+    LOG(ERROR) << "BTF: failed to initialize dump (" << err_buf << ")";
```
jordalgo: Unless we want to completely crash here maybe we just use `LOG(V1)` instead since users probably can't do anything about this error.
```suggestion
    LOG(ERROR) << "BTF: failed to initialize dump (" << err_buf << ")";
```

src/btf.cpp
```diff
@@ -727,6 +727,106 @@ FuncParamLists BTF::get_params_from_btf(
...
+    if (probe_type == ProbeType::kprobe) {
+      const auto *p = btf_params(t);
+
+      for (__u16 j = 0, argN = 0, len = btf_vlen(t); j < len; j++, p++) {
+        const std::string arg_name = "arg" + std::to_string(argN);
+
+        // set by dump_printf callback
+        type.str("");
+        if (btf_dump__emit_type_decl(dump, p->type, &decl_opts)) {
+          LOG(ERROR) << "failed to dump argument: " << arg_name;
```
jordalgo: same here
```suggestion
          LOG(V1) << "failed to dump argument: " << arg_name;
```

src/btf.cpp
```diff
@@ -727,6 +727,106 @@ FuncParamLists BTF::get_params_from_btf(
...
+          break;
+        }
+
+        // Note that floating point arguments are typically passed in special
+        // registers which don’t count as argN arguments. e.g. on x86_64 the
+        // first 6 non-floating point arguments are passed in registers and
+        // all following arguments are passed on the stack
+        const auto *pt = btf__type_by_id(btf_obj.btf, p->type);
+        if (pt && BTF_INFO_KIND(pt->info) == BTF_KIND_FLOAT) {
+          LOG(WARNING) << "BTF: skip a float/double param of " << func_name;
```
jordalgo: I don't think we should have a log here. Might be noisy in the future.

src/btf.cpp
```diff
@@ -727,6 +727,106 @@ FuncParamLists BTF::get_params_from_btf(
...
+        LOG(ERROR) << "failed to dump return type for: " << func_name;
+        break;
+      }
+
+      params[func_name].push_back(type.str() + " retval");
+    }
+  }
+
+  if (id != (max + 1))
+    LOG(ERROR) << "BTF data inconsistency " << id << "," << max;
```
jordalgo: This should maybe be a `LOG(BUG)`

src/btf.cpp
```diff
@@ -727,6 +727,106 @@ FuncParamLists BTF::get_params_from_btf(
   return params;
 }
 
+FuncParamLists BTF::get_kprobes_params_from_btf(
+    const BTFObj &btf_obj,
+    const std::set<std::string> &funcs,
+    const ProbeType &probe_type) const
```
jordalgo: Super nit but it feels weird to pass the probe type when this will only ever be called by krpobe/kretprobe. Let's make it more clear why we need this info.
```suggestion
    bool is_ret_probe) const
```

jordalgo (commented):
Few small nits but otherwise this is looking great! WDYT about adding a small runtime test for verbose mode for kprobes?

src/btf.cpp
Rtoax: Fixed, thanks.

Rtoax:
Fix all, and add simple '-lv kprobe:*' runtime test in `basic`. please review.

jordalgo (approved):

jordalgo:
@Rtoax Sorry, could you fix the clang-tidy error?

Rtoax:
@jordalgo fixed, 

```
/home/runner/work/bpftrace/bpftrace/src/btf.h:149:46: error: parameter 'is_kretprobe' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
  149 |                                              const bool is_kretprobe) const;
      |                                              ^~~~~
11192 warnings generated.
```

PR #4366, opened 2025-07-15
Macro library

jordalgo:
This is a combination of 3 commits to add some new macros to the standard library and produce documentation from the doc strings in stdlib.bt

##### Checklist

### DIFF
--- a/.github/workflows/stdlib-docs.yml
+++ b/.github/workflows/stdlib-docs.yml
@@ -1,0 +1,36 @@
+name: Standard Library Documentation Check
+
+on:
+  push:
+    branches: [ master, release/* ]
+  pull_request:
+    paths:
+      - "src/stdlib/**/*.bt"
+      - "scripts/generate_stdlib_docs.py"
+
+# Cancel previous run if a new one is started
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check-stdlib-docs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Generate stdlib documentation
+      run: ./scripts/generate_stdlib_docs.py
+    - name: Check for changes in docs/stdlib.md
+      run: |
+        if git diff --exit-code docs/stdlib.md; then
+          echo "✅ docs/stdlib.md is up to date"
+        else
+          echo "❌ docs/stdlib.md is out of date. Please run './scripts/generate_stdlib_docs.py src/stdlib/base.bt' and commit the changes."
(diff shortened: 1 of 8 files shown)


.github/workflows/stdlib-docs.yml
```diff
@@ -0,0 +1,36 @@
+name: Standard Library Documentation Check
+
+on:
+  push:
+    branches: [ master, release/* ]
+  pull_request:
+    paths:
+      - "src/stdlib/base.bt"
```
amscanne: `src/stdlib/**/*.bt`?

.github/workflows/stdlib-docs.yml
```diff
@@ -0,0 +1,36 @@
...
+  check-stdlib-docs:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Generate stdlib documentation
+      run: ./scripts/generate_stdlib_docs.py src/stdlib/base.bt
```
amscanne: I recommend you just encode the paths and pretty-printing into the script directly, to avoid encoding any logic in GitHub actions. (You could have a `-u` flag to update the docs, and exit with a non-zero exit code if they are different, plus dumping the diff just like `clang_tidy.sh`)

src/stdlib/base.bt
```diff
@@ -0,0 +1,23 @@
+/*
+Simple assertion macro that will exit the entire script with an error code if the condition is not met.
+
+:param (bool) $cond: The condition to check
+:param (string) $msg: The message to print if the condition is not met
+*/
+macro assert($cond, $msg) {
+  if (!$cond) {
+    printf("Assertion failed. Msg: %s", $msg);
```
amscanne, line 9: It would be trivial for us to add a `line` and `file` builtin. Then you could just make this look like a `cassert` message.

.github/workflows/stdlib-docs.yml
jordalgo: Sounds good.

src/stdlib/base.bt
jordalgo: Neat idea. Let me look into that.

.github/workflows/stdlib-docs.yml
jordalgo: Actually, I like the first suggestion but I'd rather do the git diffing and subsequent failing here - feels like a lack of composition to shell out to `git diff` in the python script (especially seeing that we need to update the file anyway in order to do the diff - if using `git` that is).

src/stdlib/base.bt
jordalgo: - https://github.com/bpftrace/bpftrace/pull/4374

jordalgo: I'll update the `assert` macro if/when we add the file/line builtins/functions.

PR #4368, opened 2025-07-18
Move full expansion to a separate pass

viktormalik:
Instead of doing probe expansion in codengen, do it on the AST level using a newly introduced `ProbeExpander` visitor as a part of the `ProbeExpansion` pass. The information about what kind of expansion is needed for each attach point is passed from the `ExpansionAnalyser` visitor using an internal state.

For each attach point which needs full expansion, we replace it by a new set of expanded attach points. If we end up with probes with no attach points (due to no expansion matches), we remove the probes from the AST straight away.

After this pass, there is a one-to-one correspondence between attach points and generated BPF programs. This allows to drop expansion code from all subsequent passes, most importantly semantic analyser and tracepoint format parser.

Overall, this change required some specific changes in tests, the reasons are explained in the related commits.

See commit messages for more implementation details.

This is another piece of #4114. After this, two main things remain:
- moving multi and session expansion to the new pass
- dropping expansion code from field analyser

Both of these could be bigger so I'm posting this part separately.

There should be no user-facing change (apart from one error message change).

##### Checklist


src/ast/passes/codegen_llvm.cpp
```diff
@@ -3477,12 +3471,12 @@ void CodegenLLVM::add_probe(AttachPoint &ap,
       reset_ids();
 
       std::string full_func_id = name + "_loc" + std::to_string(i);
```
jordalgo: Nit:
```suggestion

```

src/ast/passes/probe_expansion.cpp
```diff
@@ -2,33 +2,73 @@
...
+  {
+    auto it = session_ret_probes.find(&ap);
+    if (it == session_ret_probes.end()) {
+      LOG(BUG) << "Requesting return probe for non-session attach-point";
+    }
+    return *it->second;
+  }
+
+private:
+  std::map<AttachPoint *, ExpansionType> expansions;
```
jordalgo: super nit - can these be unordered maps?

tests/semantic_analyser.cpp
```diff
@@ -219,7 +219,7 @@ TEST_F(SemanticAnalyserTest, builtin_variables)
   test("kprobe:f { func }");
   test("uprobe:/bin/sh:f { func }");
   test("kprobe:f { probe }");
-  test("tracepoint:a:b { args }");
+  test("tracepoint:sched:sched_one { args }", Mock{ *get_mock_bpftrace() });
```
jordalgo: JC, why does this have to be wrapped in a new `Mock`?

tests/semantic_analyser.cpp
```diff
@@ -4367,8 +4369,8 @@ stdin:1:1-10: ERROR: Only single iter attach point is allowed.
 iter:task,f:func_1 { 1 }
 ~~~~~~~~~
 )" });
-  test("iter:task* { }", Error{ R"(
-stdin:1:1-11: ERROR: iter probe type does not support wildcards
```
jordalgo: Nit. I do prefer this older error message. Don't we have the probe string at the time of issuing this message and can therefore make the distinction between `iter:task,iter:task_vma` and `iter:task*` ?

src/ast/passes/probe_expansion.cpp
```diff
@@ -1,34 +1,75 @@
...
 #include "ast/visitor.h"
 #include "bpftrace.h"
+#include "log.h"
 #include "util/wildcard.h"
 
 namespace bpftrace::ast {
 
 using ast::ExpansionType;
 
+class ExpansionResult {
```
jordalgo, line 13: Super nit: An extra class here seems a bit overkill instead of just using a pair or struct to hold these two maps but YMMV

jordalgo (approved):
Few minor nits but otherwise nice cleanup!

tests/semantic_analyser.cpp
viktormalik: It doesn't, I just missed that the `test` method already creates a `MockBPFTrace` if it's not passed. Dropped.

src/ast/passes/probe_expansion.cpp
viktormalik: I'm not a big fan of using `std::pair` just to collate two things together. To me, `first` and `second` read much harder than the actual names expressing the semantics of the fields.

Also here, the class has some advantages - we can delete the copy constructor to ensure a single instance and can return `Expansion::NONE` for attach points which don't have an entry in the `expansions` map.

tests/semantic_analyser.cpp
viktormalik: We have `ap->raw_input` so we could check if the current attach point has been expanded. Restored the original error message.

src/ast/passes/probe_expansion.cpp
jordalgo: > we can delete the copy constructor to ensure a single instance and can return Expansion::NONE for attach points which don't have an entry in the expansions map.

Yeah true. Comment retracted 😄

jordalgo:
Looks like the last commit was done in the github UI perhaps - missing the signed-off but you know I won't block on that. Thanks for the fixes.

tests/semantic_analyser.cpp
```diff
@@ -4367,7 +4369,7 @@ stdin:1:1-10: ERROR: Only single iter attach point is allowed.
 iter:task,f:func_1 { 1 }
 ~~~~~~~~~
 )" });
-  test("iter:task* { }", Error{ R"(
+  test("iter:task* { }", Mock{ *get_mock_bpftrace() }, Error{ R"(
```
jordalgo: nit: missed an extra wrap here too
```suggestion
  test("iter:task* { }", Mock{ *get_mock_bpftrace() }, Error{ R"(
```

viktormalik:
> Looks like the last commit was done in the github UI perhaps - missing the signed-off but you know I won't block on that. Thanks for the fixes.

Ouch, forgot to squash, let me fix that.

tests/semantic_analyser.cpp
viktormalik: Again, no explicit mock needed here, dropped.

src/ast/ast.h
```diff
@@ -1239,6 +1239,8 @@ class Probe : public Node {
 };
 using ProbeList = std::vector<Probe *>;
 
+void clear_empty_probes(ProbeList &probes);
```
amscanne: Nit: this could be a method on the `Program`, since it is always used in that context.

src/ast/ast.cpp
```diff
@@ -151,50 +151,50 @@ bool is_comparison_op(Operator op)
...
+  ap->lang = lang;
+  ap->ns = ns;
+  ap->func = func;
+  ap->pin = pin;
+  ap->usdt = usdt;
+  ap->freq = freq;
+  ap->len = len;
+  ap->mode = mode;
+  ap->async = async;
+  ap->expansion = expansion;
```
amscanne, line 175: Based on a cursory look, it seems like this is no longer being relied upon for the actual expansion decision? (Which uses the `ExpansionResult`.) Is the intent to drop this once everything is cleaned up, or what's the plan?

src/ast/ast.h
viktormalik: Good idea, will do that.

src/ast/ast.cpp
viktormalik: That's exactly the plan but the `AttachPoint::expansion` field is still relied on for multi and session expansion. I'll migrate those in follow-up PRs.

src/ast/ast.h
viktormalik: Done

PR #4371, opened 2025-07-21
Make probe names be case insensitive

jordalgo:
It's not a valuable restriction to force
users to remeber that special probes like
`BEGIN`, `END`, and `BENCH` need to be all
uppercase. Furthermore, since bpftrace aims
to be a quick debugging tool we should also
give users the leeway to mix cases in their
probe names. For example, we don't need to fail
if the probe name looks like this `bEgIn`.

##### Checklist


amscanne:
It's a bit hard to review as is, so I have a favor to ask: can you split this into two changes?

1. A change that supports case-insensitive names, handling collisions, likely flattening codegen names, etc.
2. A change that updates our own use of `BEGIN` everywhere (and changes all the defaults in the docs), since I think this is added the most noise.

jordalgo:
> It's a bit hard to review as is, so I have a favor to ask: can you split this into two changes?

Sure. I think I was partially submitting to get a temperature check on this idea - shoulda marked it with RFC.

amscanne:
> > It's a bit hard to review as is, so I have a favor to ask: can you split this into two changes?
> 
> Sure. I think I was partially submitting to get a temperature check on this idea - shoulda marked it with RFC.

As far as a temperature check goes: I think that *providers* can be case-insensitive. We still need case-sensitive targets because that depends on the underlying model.

viktormalik:
> As far as a temperature check goes: I think that _providers_ can be case-insensitive. We still need case-sensitive targets because that depends on the underlying model.

Agreed, this makes sense to me.

jordalgo:
Ok cool. Then I think this is ready for a review though most of it is just find/replace.

tests/codegen/builtin_ctx_field.cpp
```diff
@@ -4,7 +4,7 @@ namespace bpftrace::test::codegen {
 
 TEST(codegen, builtin_ctx_field)
 {
-  std::string prog = R"END(
```
jordalgo, line 7: this didn't need to change but don't think it makes a difference

jordalgo:
@amscanne @viktormalik - Review ping

src/probe_types.cpp
```diff
@@ -15,12 +16,13 @@ std::ostream &operator<<(std::ostream &os, ProbeType type)
 ProbeType probetype(const std::string &probeName)
 {
   ProbeType retType = ProbeType::invalid;
+  const std::string lowerProbeName = util::to_lower(probeName);
 
   auto v = std::ranges::find_if(PROBE_LIST,
```
amscanne, line 21: Can we drop this extra line?

src/probe_types.h
```diff
@@ -60,10 +60,10 @@ const std::vector<ProbeItem> PROBE_LIST = {
     .aliases = { "U" },
     .type = ProbeType::usdt,
     .show_in_userspace_list = true },
-  { .name = "BEGIN", .aliases = { "BEGIN" }, .type = ProbeType::special },
-  { .name = "END", .aliases = { "END" }, .type = ProbeType::special },
+  { .name = "begin", .aliases = { "begin" }, .type = ProbeType::special },
+  { .name = "end", .aliases = { "end" }, .type = ProbeType::special },
   { .name = "self", .aliases = { "self" }, .type = ProbeType::special },
-  { .name = "BENCH", .aliases = { "BENCH" }, .type = ProbeType::benchmark },
+  { .name = "bench", .aliases = { "bench" }, .type = ProbeType::benchmark },
```
amscanne: Nit: "bench" is an alias of "bench"? I think you just want to drop the alias here

src/bpftrace.cpp
```diff
@@ -693,17 +693,17 @@ int BPFtrace::run(Output &out, BpfBytecode bytecode)
...
 
-  auto begin_probe = resources.special_probes.find("BEGIN");
+  auto begin_probe = resources.special_probes.find("begin");
   if (begin_probe != resources.special_probes.end()) {
     auto &begin_prog = bytecode_.getProgramForProbe((*begin_probe).second);
     if (::bpf_prog_test_run_opts(begin_prog.fd(), nullptr))
       return -1;
 
-    LOG(V1) << "Attaching BEGIN";
+    LOG(V1) << "Attaching begin";
```
amscanne: Nit: here and below it's not clear if this is the probe name or just bad grammar. Can we quote the probe? E.g. "Attaching `begin` probe"

tests/codegen/builtin_ctx_field.cpp
amscanne: I've been removing these cases and just standing on the multi-line literal strings, since many are different. Can you just make this `R"(`?

amscanne (approved):

src/probe_types.cpp
jordalgo: Done.

src/probe_types.h
jordalgo: Done.

src/bpftrace.cpp
jordalgo: Done.

tests/codegen/builtin_ctx_field.cpp
jordalgo: Done.

PR #4372, opened 2025-07-21
Fix builds on PowerPC

viktormalik:
Fix two issues which occurred when trying to build bpftrace on PowerPC:
1. The `Machine` enum entry for PowerPC is `PPC64`, not `POWERPC`.
2. Some GCC versions define `__powerpc64__` rather than `__ppc64__`.

##### Checklist

### DIFF
--- a/src/arch/arch.h
+++ b/src/arch/arch.h
@@ -93,8 +93,8 @@ constexpr Machine current()
   return Machine::ARM;
 #elif defined(__s390x__)
   return Machine::S390X;
-#elif defined(__ppc64__)
-  return Machine::POWERPC;
+#elif defined(__ppc64__) || defined(__powerpc64__)
+  return Machine::PPC64;
 #elif defined(__mips64)
   return Machine::MIPS64;
 #elif defined(__riscv) && (__riscv_xlen == 64)


CHANGELOG.md
```diff
@@ -131,6 +131,8 @@ and this project adheres to
   - [#4244](https://github.com/bpftrace/bpftrace/pull/4244)
 - Report expanded probe name when trying to attach
   - [#4353](https://github.com/bpftrace/bpftrace/pull/4353)
+- Fix builds on PowerPC
```
jordalgo: @viktormalik Is this something broken on the current version or is it just in trunk after @amscanne 's refactoring

viktormalik: Oh, right, just in trunk. I'll drop the changelog entry, sorry.

src/arch/arch.h
```diff
@@ -93,8 +93,8 @@ constexpr Machine current()
   return Machine::ARM;
 #elif defined(__s390x__)
   return Machine::S390X;
-#elif defined(__ppc64__)
-  return Machine::POWERPC;
+#elif defined(__ppc64__) || defined(__powerpc64__)
```
amscanne, line 96: Do you know of the best authoritative source for this? I think I was using [this one](https://github.com/cpredef/predef/blob/master/Architectures.md), but would like to change if there's a better reference since I know this can vary quite a lot

src/arch/arch.h
```diff
@@ -93,8 +93,8 @@ constexpr Machine current()
...
+  return Machine::PPC64;
```
amscanne, line 97: 🤦 

The goal of the change was to get rid of as much `#ifdef` as possible, so clearly the one that was left ended up biting. Do you think we should just make this fully dynamic? It's not in the hot path, this function could literally use `uname`, it would just no longer be `constexpr`. WDYT?

src/arch/arch.h
viktormalik: Not one that I know of, unfortunately. That one looks quite complete, the question is whether we want to check for all of the possible identifiers.

src/arch/arch.h
viktormalik: I thought it must be `constexpr` so that it can be used as `Arch<current()>`. But if not, then probably using `uname` would be the safest option.

amscanne: Yes, it must be. Sorry if I was unclear: that path would be changed (and it just would no longer be a compile-time constant, but it doesn't really matter for performance anyways).

I can do this in a follow-up if it's friendly, I just wanted to get thoughts.

viktormalik: Now thinking about it, parsing `uname` output will have a similar problem - some arches have different valid values (like `ppc`, `ppc64`, `ppc64le`, `ppcle` for PowerPC to give an example) so there will still be space for errors. So, maybe let's just leave it the way it is, I kind of like the elegance of the current implementation.

amscanne: Probably the best thing here would be just to set up an emulated build & perhaps execution of some unit tests for all supported architectures. We can do it with GitHub actions, just need some qemu trickery.

PR #4373, opened 2025-07-21
Tests: vendor the sdt.h header

viktormalik:
When `<sys/sdt.h>` is not present on the system, each test prog defines a dummy `DTRACE_PROBEXY` macro to at least enable compiling the tests.

`usdt_multi_args.c` defines the wrong macro variant so fix it to the one actually used inside the test prog.

##### Checklist


amscanne:
Where does this header come from? Can we make this dependency mandatory if `BUILD_TESTS` is set? (And presumably this could be just a build-time dependency, rather than any kind of runtime dependency for the packaged versions)

viktormalik:
`systemtap-sdt-devel`, at least on Fedora. Not sure what was the original reason to make it optional, perhaps it's not available on all the systems. But I think we even might be able to ship it, I'll check that.

amscanne:
> `systemtap-sdt-devel`, at least on Fedora. Not sure what was the original reason to make it optional, perhaps it's not available on all the systems. But I think we even might be able to ship it, I'll check that.

If it's test-only, I'd rather not make it a runtime dependency. But seems very reasonable to have as a requirement if you want to build/run tests, and we already have `BUILD_TESTS`

viktormalik:
Vendoring the header works, I think that this is the best path forward as we can completely drop the SystemTap dependency and all the ifdefs for handling the case when it is not present.

tests/testprogs/usdt_inlined.c
```diff
@@ -1,8 +1,5 @@
-#ifdef HAVE_SYSTEMTAP_SYS_SDT_H
-#include <sys/sdt.h>
-#else
-#define DTRACE_PROBE2(a, b, c, d) (void)0
-#endif
+#include "sdt.h"
```
amscanne: Nit: for consistency, I have been trying to leave header blocks in the same layout. I picked the most common one, which I think is probably used in the vast majority of the code base now, which is:

```
<all system includes and dependencies>

"local includes"
```

Since this file is local (and correctly using `""`), I would put it in the block beneath the system block.

.clang-format
```diff
@@ -89,4 +89,7 @@ UseTab: Never
 ---
 Language: Json
 DisableFormat: true
+---
```
amscanne, line 92: Just curious, what was the reason for this?

tests/testprogs/usdt_inlined.c
viktormalik: Ack, I can do that.

.clang-format
viktormalik: Apparently, `std.h` contains some ObjectiveC code and clang-format was failing on it.

tests/testprogs/usdt_inlined.c
viktormalik: Should be done now.

.clang-format
amscanne: Yikes!

PR #4374, opened 2025-07-21
Add file and line builtins

jordalgo:
This is mostly in support of the new
`assert` macro so we can also show file and line
so make sure the source_expansion pass happens
AFTER macro expansion.

##### Checklist


src/ast/passes/source_expansion.cpp
```diff
@@ -0,0 +1,54 @@
...
+void SourceExpansionPass::visit(Expression &expr)
+{
+  auto *builtin = expr.as<Builtin>();
+  if (!builtin || (builtin->ident != "file" && builtin->ident != "line")) {
+    Visitor<SourceExpansionPass>::visit(expr);
+    return;
+  }
+
+  if (builtin->ident == "file") {
+    expr.value = ast_.make_node<String>(ast_.source()->filename,
```
amscanne: Interesting, this will always use the main source file. I believe the location keeps a reference to the source, did you not want to have this align with the way you do it below?

src/ast/passes/source_expansion.cpp
```diff
@@ -0,0 +1,54 @@
...
+                                        Location(builtin->loc));
+    return;
+  }
+
+  auto line = builtin->loc->parent ? builtin->loc->parent->loc->line()
```
amscanne: IIUC, if this is a macro, it will use the parent of depth 1. (If that is also a macro, that is the line that will be used.) If it is not a macro, it will use the current line.

I don't think it's crazy behavior, but I think you want to get precise about this (and make sure that it is using the same thing for the file above).

src/ast/passes/source_expansion.cpp
```diff
@@ -0,0 +1,54 @@
...
+      : ast_(ast), bpftrace_(bpftrace)
+  {
+  }
+
+  using Visitor<SourceExpansionPass>::visit;
+  void visit(Expression &expr);
+
+private:
+  ASTContext &ast_;
+  BPFtrace &bpftrace_;
```
amscanne: This member is not needed AFAICT.

src/ast/passes/source_expansion.cpp
jordalgo: Ack.

src/ast/passes/source_expansion.cpp
jordalgo: Good call. I'm gonna follow the parent to the root.

src/ast/passes/source_expansion.cpp
jordalgo: Ack.

src/ast/passes/source_expansion.cpp
amscanne: Do you want to reflect on this for a day? The code is fine, I'm just wondering if we *always* want the root. Sometimes you just want the caller. 

For example, Go has `runtime.Callers(skip int)`. The equivalent here might be `line` or `line(1)` (e.g. get the root line, or get the line from 1 caller up). Now it's getting complicated, and I regret suggesting it was so simple 🤣 

.... Maybe what we actually want is `errorf`, `warningf` that are like `printf` but emit with the location chain? Anyways, you can submit as is, but I do think we should think about this a bit. I think that we can use these mechanisms in the future to generate errors, etc.

amscanne (approved):
LGTM but I think we should think on it a bit

src/ast/passes/source_expansion.cpp
jordalgo: > I'm just wondering if we always want the root.

I thought about this too and figured that we could always make this the default behavior (via the builtin) and if we wanted to allow different behavior then we could add a `line` or `file` function (like we have for `pid` and `tid`) that accepts params. 

Either way, I'll think on it and wait for @viktormalik to have a look before merging. 

> Maybe what we actually want is errorf, warningf that are like printf

Yeah that could work too.

jordalgo:
@viktormalik I un-resolved a comment where I addressed the expanding of this.

amscanne:
> I'm not really sure if we want to always follow the parents. It will be beneficial for cases like `assert` but if you want to debug a macro, you may want to see the actual file and line of the macro body.

To this point, I think it would be common to have an `assert` in a user macro.

Maybe we just skip the built-in, and require `file(N)` and `line(N)` with integer literals that indicate the number of expansions to skip. So `line(0)` and `file(0)` are the current location, `line(1)` and `file(1)` are the parent, etc. Anything negative would be any number of expansions, e.g. `line(-1)` and `file(-1)` would be the probe always.

jordalgo:
I could get behind a function but maybe we also add a builtin `#parents` (or `#srcs`) which would allow users of `file(N)` and `line(N)` to get the top or the bottom (I'm not terribly keen on the negative number). They could also create a loop to walk up the parent stack (although now it's getting even more complicated).

viktormalik:
> I could get behind a function but maybe we also add a builtin `#parents` (or `#srcs`) which would allow users of `file(N)` and `line(N)` to get the top or the bottom (I'm not terribly keen on the negative number). They could also create a loop to walk up the parent stack (although now it's getting even more complicated).

Agreed. It feels like mostly 0, 1, and -1 would be used. We could introduce some names for them like `line(here)`, `line(parent)`, `line(probe)`.

jordalgo:
Another idea is that maybe we just make the builtins `lines` and `files` which are both 1-D arrays (int and string) with the most local src/location being first and the top level parent being last.

amscanne:
> Agreed. It feels like mostly 0, 1, and -1 would be used. We could introduce some names for them like `line(here)`, `line(parent)`, `line(probe)`.

We have this in a few places (special enum identifiers as function arguments). I'd like to think through a generalized model for this without having to hard-code all these cases.

I'll start with a strawman proposal. Suppose we have the following (as a C prototype):

```
int foo(enum fooparam x);
```

Then we call:

```
foo(Something);
```

This implicitly becomes:

```
foo(fooparam::Something);
```

I think that for enum parameters, naked identifiers should always be evaluated as a member of the enum. Then all these special cases reduce to that. This will require some support in the semantic analyzer, but e.g. for C functions, we have the BTF type information for each parameter and know the full value space, and can automatically apply this when `Identifier`s are passed as parameters.

I don't think this is possible for `SizedType` because enum values are not enumerated, but it's possible for extern functions.  We could add a pass ahead of the full semantic analyzer pass that looks at extern functions specifically and applies types to identifiers based on the Enum value. Then in a C interop file we could define something like:

```
enum Location {
   local,
   parent,
   root,
 };
 
 bool location_equal(Location l1, Location l2) {
    return l1 == l2;
 }
 ```
 
Then we could add a macro that looks something like:

```
macro line($x) {
   let $line;
   if (location_equal($x, local)) {
       $line = local_line; // Builtin injected by LocationPass.
   } else if (location_equal($x, root)) {
       $line = root_line; // Same.
   } else if (location_equal($x, parent)) {
       $line = parent_line; // Same.
   }
   $line
}
```

And we would define three trivial builtins: `local_line`, `root_line`, `parent_line`.

**Comment by jordalgo** [comment]
I know we want things to be composable and whatnot but I'm coming around to something similar to what @amscanne suggested above. Maybe we just expose a single new function `print_source` the dumps out like what we'd see for an ast node error e.g.
```
filename:1:21-22
macro add2($x) { $x + 1 } macro add1($x) { add2($x) } BEGIN { $a = "string"; add1($a); }
                    ~
filename:1:78-86
macro add2($x) { $x + 1 } macro add1($x) { add2($x) } BEGIN { $a = "string"; add1($a); }
                                                                             ~~~~~~~~
filename:1:44-52
macro add2($x) { $x + 1 } macro add1($x) { add2($x) } BEGIN { $a = "string"; add1($a); }
                                           ~~~~~~~~
```
It's probably what we want for the `assert` macro anyway and doesn't involve any changes to the lexer.

jordalgo:
I'm going to close this in favor of adding a `print_error` function:
- https://github.com/bpftrace/bpftrace/pull/4409

PR #4380, opened 2025-07-24
Move session expansion to AST

viktormalik:
When doing session expansion, we need to take two probes and merge them into one. For instance:

    kprobe:* { XXX } kretprobe:* { YYY }

become

    kprobe:* { if (bpf_session_is_return()) { YYY } else  { XXX } }

Until now, we were doing it in codegen but now that we have proper AST mutation passes, we can do it directly in AST.

This changes the `SessionAnalyser` visitor in the `ProbeExpansion` pass to `SessionExpander` which directly does the above transformation. For the condition, we introduce an internal builtin `__session_is_return` which codegen transforms into a call to the `bpf_session_is_return` kfunc.

This also introduces a new test suite for probe expansions and a test which verifies what the AST looks like after the session expansion.

##### Checklist

### DIFF
--- a/src/ast/passes/probe_expansion.cpp
+++ b/src/ast/passes/probe_expansion.cpp
@@ -1,5 +1,7 @@
 #include "ast/passes/probe_expansion.h"
 
+#include <algorithm>
+
 #include "ast/ast.h"
 #include "ast/visitor.h"
 #include "bpftrace.h"
@@ -29,22 +31,9 @@ public:
       return ExpansionType::NONE;
     return exp->second;
   }
-  void set_session_ret_probe(AttachPoint &ap, Probe &ret_probe)
-  {
-    session_ret_probes[&ap] = &ret_probe;
-  }
-  Probe &get_session_ret_probe(AttachPoint &ap)
-  {
-    auto it = session_ret_probes.find(&ap);
-    if (it == session_ret_probes.end()) {
-      LOG(BUG) << "Requesting return probe for non-session attach-point";
-    }
-    return *it->second;
-  }
 
 private:
   std::unordered_map<AttachPoint *, ExpansionType> expansions;
-  std::unordered_map<AttachPoint *, Probe *> session_ret_probes;
 };
 
 class ExpansionAnalyser : public Visitor<ExpansionAnalyser> {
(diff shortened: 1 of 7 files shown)


src/ast/passes/probe_expansion.cpp
```diff
@@ -168,68 +157,87 @@ void ExpansionAnalyser::visit(Builtin &builtin)
...
+        return other->attach_points.size() == 1 && other->pred == nullptr &&
+               probetype(other->attach_points[0]->provider) ==
+                   ProbeType::kretprobe &&
+               expansion_result_.get_expansion(*other->attach_points[0]) ==
+                   ExpansionType::MULTI &&
+               other->attach_points[0]->target == ap->target &&
+               other->attach_points[0]->func == ap->func;
+      });
+
+  // If there's not exactly one match, we don't know how to do session expansion
```
amscanne, line 202: Thinking out loud: the rules for merge session probes here can be difficult to understand in practice and they are sensitive to the rest of the program (leading to cases like... "hey I added a totally unrelated attach point and now I'm losing events"). I feel it would be good long-term to have a unified way that ensures this is always applied. I'll propose something a bit radical just to discuss (and it's also fine if this is the long term goal, and we are merely taking incremental steps).

* We unilaterally expand programs ahead of semantic analysis, ignoring specifics of whether they are using specific builtins, etc. For example: `kprobe:sched*` would be aggressively expanded into many probes
* Add `operator=` support for the AST to allow deep equals, taking into account types, etc.
* Do a few things in a pass after semantic analysis:
1) turns instances of the same program type+attach point into **functions**, with a new probe definition that calls those N functions in sequence (make unique)
2) merges every kretprobe into a session probe, optionally folding the at-most-one kprobe if it exists (fold)
3) collapse all identical probes (with matching program type and equal programs, per above) into probes with multiple attach points (dedup)

This way retprobes are always folded into session probes, and they will always capture the kprobes regardless of how many exist. WDYT?

amscanne (approved):
I didn't dig too deeply, but all seems reasonable.

src/ast/passes/probe_expansion.cpp
viktormalik: I may need to think about this more to fully understand. Some comments below in the meantime:

> - We unilaterally expand programs ahead of semantic analysis, ignoring specifics of whether they are using specific builtins, etc. For example: `kprobe:sched*` would be aggressively expanded into many probes

Do you mean to fully expand as in generate many AST probes, each with a single attach point? Because now, "full" expansion just expands the attach points, not the probe itself.

(note: I'll probably introduce full AST probe expansion soon since we'll need it to address some issues)

> 1. turns instances of the same program type+attach point into functions, with a new probe definition that calls those N functions in sequence (make unique)

So after all, we'd introduce functions to bpfscript?

> This way retprobes are always folded into session probes, and they will always capture the kprobes regardless of how many exist. WDYT?

Overall, this seems to be a bit complicated to me. Perhaps I'm wrong and we need the complexity but I think that we could do with just smarter attach point analysis. That should be sufficient to be able to generate multi and session probes as efficiently as possible.

Also, one thing I don't like too much about too many AST transformations is that the resulting set of BPF programs would be completely detached from the original set of probes in bpfscript. Power users may get confused by it when they want to, e.g. inspect the BPF programs generated by bpftrace using bpftool.

Anyways, this looks like a longer discussion so if you don't mind, I'll go ahead and merge this.

jordalgo: +1 - I also need to digest this suggestion on handling probe expansion and subsequent folding/de-deduping.

amscanne: > I may need to think about this more to fully understand. Some comments below in the meantime:
> 
> 
> 
> > - We unilaterally expand programs ahead of semantic analysis, ignoring specifics of whether they are using specific builtins, etc. For example: `kprobe:sched*` would be aggressively expanded into many probes
> 
> 
> 
> Do you mean to fully expand as in generate many AST probes, each with a single attach point? Because now, "full" expansion just expands the attach points, not the probe itself.

Right, and there's no need to carry through state around expansion types, etc. In unusual edges cases (e.g. where you glob a ton of stuff) it'll be a bit slower. But a) you could optimize in the future, once you have the basic model and b) IMO those cases are not actually useful for building real tools in practice, and even our demos end up getting expanding prior to codegen (because they use "probe" or whatever) which is more expensive than our passes anyways.

> (note: I'll probably introduce full AST probe expansion soon since we'll need it to address some issues)
> 
> 
> 
> > 1. turns instances of the same program type+attach point into functions, with a new probe definition that calls those N functions in sequence (make unique)
> 
> 
> 
> So after all, we'd introduce functions to bpfscript?

Functions already exist and somewhat work (every probe is just a function!) the main issue was the business of expressing and parsing types (the bpfscript part, which I'm not really getting into here). Moving a probe to a subprog as part of a pass (carrying over types, etc.) is more of a bookkeeping mechanism, and just changes how libbpf is going to load and link things (in a sensible way).

> > This way retprobes are always folded into session probes, and they will always capture the kprobes regardless of how many exist. WDYT?
> 
> 
> 
> Overall, this seems to be a bit complicated to me. Perhaps I'm wrong and we need the complexity but I think that we could do with just smarter attach point analysis. That should be sufficient to be able to generate multi and session probes as efficiently as possible.

We can think about it. It seems complicated, but my goal would be to make things simpler by having the ability to reason about those steps on the whole graph. I worry about "smarter" attach point analysis because it ends up being much more fragile logic.

> Also, one thing I don't like too much about too many AST transformations is that the resulting set of BPF programs would be completely detached from the original set of probes in bpfscript. Power users may get confused by it when they want to, e.g. inspect the BPF programs generated by bpftrace using bpftool.

This is an interesting philosophical question. IMO when people use bpftrace, they just want to solve their problem and not necessarily be prescriptive about the details. If you want to do that, you could use always BPF C. Even having to understand all the probe types could be a bother to the user, and ideally we would provide the semantic meaning without having the user need to understand all the technical details.

Ironically, your automatic session probes are absolutely doing just that ;). This is why I want to transparently upgrade all retprobes, because no one actually *wants* to learn about `max_active`, and we can totally save our users the pain!

> Anyways, this looks like a longer discussion so if you don't mind, I'll go ahead and merge this.

Absolutely!

PR #4381, opened 2025-07-24
Move multi-expansion to ProbeExpansion

viktormalik:
When using kprobe_multi, uprobe_multi, or kprobe_session, we need to expand a wildcarded attach point and add all expanded target functions to `Probe::funcs`. At the moment, we do it in the `BPFtrace` class but there's no reason not to do it in the `ProbeExpansion` pass. This just needs to introduce a new `ast::AttachPoint::expanded_funcs` field which will be used to fill `Probe::funcs` later.
    
This allows to drop the remaining expansion code from `BPFtrace` and codegen. It also removes one call to `ProbeMatcher` (reflected in tests).

##### Checklist

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -52,22 +52,6 @@ enum class Operator {
   BNOT,
 };
 
-// There are 2 kinds of attach point expansion:
-// - full expansion  - separate LLVM function is generated for each match
-// - multi expansion - one LLVM function and BPF program is generated for all
-//                     matches, the list of expanded functions is attached to
-//                     the BPF program using the k(u)probe.multi mechanism
-// - session expansion - extension of the multi expansion when a single BPF
-//                       program is shared for both the entry and the exit probe
-//                       (when they are both attached to the same attach points)
-//                       using the kprobe.session mechanism
-enum class ExpansionType {
-  NONE,
-  FULL,
-  MULTI,
-  SESSION,
-};
-
 class Node {
 public:
   Node(ASTContext &ctx, Location &&loc) : state_(*ctx.state_), loc(loc) {};
@@ -1137,7 +1121,6 @@ public:
         len(other.len),
         mode(other.mode),
         async(other.async),
-        expansion(other.expansion),
         address(other.address),
         func_offset(other.func_offset),
         ignore_invalid(other.ignore_invalid),
@@ -1168,8 +1151,6 @@ public:
   uint64_t len = 0;   // for watchpoint probes, the width of watched addr
   std::string mode;   // for watchpoint probes, the watch mode
   bool async = false; // for watchpoint probes, if it's an async watchpoint
-
-  ExpansionType expansion = ExpansionType::NONE;
 
   uint64_t address = 0;
   uint64_t func_offset = 0;
--- a/src/ast/passes/probe_expansion.cpp
+++ b/src/ast/passes/probe_expansion.cpp
@@ -2,39 +2,12 @@
 
 #include <algorithm>
 
-#include "ast/ast.h"
 #include "ast/visitor.h"
 #include "bpftrace.h"
 #include "log.h"
 #include "util/wildcard.h"
 
 namespace bpftrace::ast {
-
-using ast::ExpansionType;
-
-class ExpansionResult {
-public:
-  ExpansionResult() = default;
-  ExpansionResult(const ExpansionResult &) = delete;
-  ExpansionResult &operator=(const ExpansionResult &) = delete;
-  ExpansionResult(ExpansionResult &&) = default;
-  ExpansionResult &operator=(ExpansionResult &&) = default;
-
-  void set_expansion(AttachPoint &ap, ExpansionType type)
-  {
-    expansions[&ap] = type;
-  }
-  ExpansionType get_expansion(AttachPoint &ap)
-  {
-    auto exp = expansions.find(&ap);
-    if (exp == expansions.end())
-      return ExpansionType::NONE;
-    return exp->second;
-  }
-
-private:
-  std::unordered_map<AttachPoint *, ExpansionType> expansions;
-};
 
 class ExpansionAnalyser : public Visitor<ExpansionAnalyser> {
 public:
@@ -243,28 +216,27 @@ void SessionExpander::visit(Probe &probe)
 
 class ProbeExpander : public Visitor<ProbeExpander> {
 public:
-  ProbeExpander(ASTContext &ast, BPFtrace &bpftrace)
-      : ast_(ast), bpftrace_(bpftrace)
+  ProbeExpander(ASTContext &ast, BPFtrace &bpftrace, ExpansionResult &result)
+      : ast_(ast), bpftrace_(bpftrace), result_(result)
   {
   }
 
-  void expand(ExpansionResult result);
+  void expand();
 
   using Visitor<ProbeExpander>::visit;
   void visit(Program &prog);
   void visit(AttachPointList &aps);
 
 private:
-  ExpansionResult result_;
   uint64_t probe_count_ = 0;
 
   ASTContext &ast_;
   BPFtrace &bpftrace_;
+  ExpansionResult &result_;
 };
 
-void ProbeExpander::expand(ExpansionResult result)
+void ProbeExpander::expand()
 {
-  result_ = std::move(result);
   visit(*ast_.root);
 }
 
@@ -308,7 +280,6 @@ void ProbeExpander::visit(AttachPointList &aps)
       case ExpansionType::SESSION:
       case ExpansionType::MULTI:
       case ExpansionType::NONE: {
-        ap->expansion = expansion;
         new_aps.push_back(ap);
         break;
       }
@@ -327,8 +298,10 @@ Pass CreateProbeExpansionPass()
     SessionExpander session_expander(ast, bpftrace, result);
     session_expander.visit(*ast.root);
 
-    ProbeExpander expander(ast, bpftrace);
-    expander.expand(std::move(result));
+    ProbeExpander expander(ast, bpftrace, result);
+    expander.expand();
+
+    return result;
   };
 
   return Pass::create("ProbeExpansion", fn);
--- a/src/ast/passes/probe_expansion.cpp
+++ b/src/ast/passes/probe_expansion.cpp
@@ -253,6 +253,7 @@ void ProbeExpander::visit(AttachPointList &aps)
 
   AttachPointList new_aps;
   for (auto *ap : aps) {
+    auto probe_type = probetype(ap->provider);
     auto expansion = result_.get_expansion(*ap);
     switch (expansion) {
       case ExpansionType::FULL: {
@@ -278,7 +279,33 @@ void ProbeExpander::visit(AttachPointList &aps)
       }
 
       case ExpansionType::SESSION:
-      case ExpansionType::MULTI:
+      case ExpansionType::MULTI: {
+        auto matches = bpftrace_.probe_matcher_->get_matches_for_ap(*ap);
+        if (util::has_wildcard(ap->target)) {
+          // If we have a wildcard in the target path, we need to generate one
+          // attach point per expanded target
+          assert(probe_type == ProbeType::uprobe ||
+                 probe_type == ProbeType::uretprobe);
+
+          std::unordered_map<std::string, AttachPoint *> new_aps_by_target;
+          for (const auto &func : matches) {
+            auto *match_ap = ap->create_expansion_copy(ast_, func);
+            // Reset the original (possibly wildcarded) function name
+            auto expanded_func = match_ap->func;
+            match_ap->func = ap->func;
+
+            auto new_ap = new_aps_by_target.emplace(match_ap->target, match_ap);
+            result_.add_expanded_func(*new_ap.first->second,
+                                      match_ap->target + ":" + expanded_func);
+          }
+          for (auto &[_, new_ap] : new_aps_by_target)
+            new_aps.push_back(std::move(new_ap));
+        } else if (!matches.empty()) {
+          result_.set_expanded_funcs(*ap, std::move(matches));
+          new_aps.push_back(ap);
+        }
+        break;
+      }
       case ExpansionType::NONE: {
         new_aps.push_back(ap);
         break;
--- a/tests/pid_filter_pass.cpp
+++ b/tests/pid_filter_pass.cpp
@@ -2,6 +2,7 @@
 #include "ast/attachpoint_parser.h"
 #include "ast/passes/field_analyser.h"
 #include "ast/passes/printer.h"
+#include "ast/passes/probe_expansion.h"
 #include "driver.h"
 #include "mocks.h"
 #include "gtest/gtest.h"
@@ -29,6 +30,7 @@ void test(const std::string& input, bool has_pid, bool has_filter)
                 .put(bpftrace)
                 .add(CreateParsePass())
                 .add(ast::CreateParseAttachpointsPass())
+                .add(ast::CreateProbeExpansionPass())
                 .add(ast::CreateFieldAnalyserPass())
                 .add(ast::CreatePidFilterPass())
                 .run();
@@ -62,7 +64,7 @@ TEST(pid_filter_pass, add_filter)
     "fentry:f",
     "fexit:f",
     "tracepoint:category:event",
-    "rawtracepoint:event",
+    "rawtracepoint:module:event",
   };
 
   for (auto& probe : filter_probes) {
--- a/src/ast/ast.cpp
+++ b/src/ast/ast.cpp
@@ -172,7 +172,6 @@ AttachPoint *AttachPoint::create_expansion_copy(ASTContext &ctx,
   ap->len = len;
   ap->mode = mode;
   ap->async = async;
-  ap->expansion = expansion;
   ap->address = address;
   ap->func_offset = func_offset;
 
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -3414,13 +3414,6 @@ void CodegenLLVM::add_probe(AttachPoint &ap,
 {
   current_attach_point_ = &ap;
   probefull_ = ap.name();
-  auto expansion = expansions_.get_expansion(ap);
-  if (expansion != ExpansionType::NONE && expansion != ExpansionType::FULL) {
-    // For non-full expansion, we need to avoid generating the code for attach
-    // points with no matches as the BPF program would fail to load.
-    if (bpftrace_.probe_matcher_->get_matches_for_ap(ap).empty())
-      return;
-  }
   if (probetype(ap.provider) == ProbeType::usdt) {
     auto usdt = usdt_helper_.find(bpftrace_.pid(), ap.target, ap.ns, ap.func);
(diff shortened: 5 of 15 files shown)


src/ast/ast.h
```diff
@@ -1176,6 +1177,8 @@ class AttachPoint : public Node {
   uint64_t bpf_prog_id = 0;
   bool ignore_invalid = false;
 
+  std::set<std::string> expanded_funcs; // for multi-attachment probes
```
viktormalik: I'm on the fence with this one. I really don't like introducing more fields to AST nodes but the only reasonable alternative that I could think of is storing `ExpansionResult` in the `BPFtrace` class which is basically the same.

IMHO ideally, the `ProbeExpansion` pass should emit `ExpansionResult` and it should be used by the subsequent passes in `PassManager`. The problem is that it needs to be used in `BPFtrace::add_probe` which is not implemented within any pass.

Thoughts? cc @amscanne

src/ast/passes/probe_expansion.cpp
```diff
@@ -306,7 +307,36 @@ void ProbeExpander::visit(AttachPointList &aps)
       }
 
       case ExpansionType::SESSION:
-      case ExpansionType::MULTI:
+      case ExpansionType::MULTI: {
+        auto matches = bpftrace_.probe_matcher_->get_matches_for_ap(*ap);
+        if (util::has_wildcard(ap->target)) {
```
amscanne, line 284: Why is there a separate check for `has_wildcard`? I thought `get_matches_for_ap` returns all matches, both for wildcard cases and cases where there may be multiple matches even without a wildcard (e.g. overloading in C++).

(I see this existed already, I just don't understand it!)

src/ast/ast.h
amscanne: (Apologies, as this turned into a bit of a rant. But hopefully it makes sense. TL;DR: It's fine, but it's a symptom of structural problems with `AttachPoint`)

My naive assumption/intuition is that post-"expansion", you would have probes with a set of attach points that are the places where you intend to attach the associated program (may still be symbolic, e.g. a USDT target, but no longer have wildcards, etc.). Furthermore: past the expansion point, I would assume that programs need only be compiled once (not compiled per attach point), and you could use multi-attach where available for the set of attach points associated with a given program. (So any merge/dedup pass would need to preserve this property and merge two probes into one only when the programs were semantically the same.)

However, I think that we either lost this property, or never had it in the first place. We treat the `AttachPoint` not as a place where the program gets bound, but as a "compile context and **set** of attach points" for each program, and we possibly re-generate the probe for each "attach point" (hence the `add_probe(ap, probe)`). This is the reason you're needing to stuff the functions there, it's not an "attach point" at all! At best, it's a "spec" at the AST level, and we stuff on top of this spec a these-are-likely-similar "attach points" model (made more confusing by the fact that the spec is called `AttachPoint`, oh and we replicate the spec itself when doing expansion).

I'll stop ranting. You've been poking at this and know it better than me, so I certainly don't object to some intermediate state that stuffs a set of strings. I do feel like this is certainly an incremental improvement over having the logic in codegen. However, I don't know if there's a better way to workaround this issue unless we fix `AttachPoint` to actually *be* an attach point for the associated program (that doesn't need to have that program compiled separately for each "attach point", which makes it a totally different thing. It also doesn't need to be materialized in that way in the AST itself, that's just an implementation detail,.. but conceptually you get the picture). BUT this gets into  the long-term proposal in the other PR, i.e. fully expand everything, do all semantic analysis, do optimizations like the ksession rewrite, then dedup/merge all applicable cases (by comparing the programs, not complex logic around expansion types).

The last point on this (then rant done, I promise) is that for all this groups-of-similar-stuff-on-the-spec complexity, it doesn't even work that well. For example `kprobe:f {} kprobe:g {}` results in two separate identical programs could be trivially merged and multi-attached, but this isn't captured by our manual-globbing-expansion rules. Heck even `kprobe:f*, kprobe:g* {}` will generate two separate programs because the only multiplexing is on `AttachPoint`, not the probe.

viktormalik: > (Apologies, as this turned into a bit of a rant. But hopefully it makes sense. TL;DR: It's fine, but it's a symptom of structural problems with `AttachPoint`)

No problem, go ahead and rant, I think it will eventually lead us to improving the stuff. Here, it made me better understand your vision of full expansion and subsequent deduplication.

In fact, I think I'll give it a try to emit the `ExpansionResult` from `ProbeExpansion` and then use it from codegen (which should be possible as both are managed by `PassManager`). Then, we'll just pass the list of functions to `BPFtrace::add_probe` from `CodegenLLVM::add_probe`.

> My naive assumption/intuition is that post-"expansion", you would have probes with a set of attach points that are the places where you intend to attach the associated program (may still be symbolic, e.g. a USDT target, but no longer have wildcards, etc.). Furthermore: past the expansion point, I would assume that programs need only be compiled once (not compiled per attach point), and you could use multi-attach where available for the set of attach points associated with a given program. (So any merge/dedup pass would need to preserve this property and merge two probes into one only when the programs were semantically the same.)
> 
> However, I think that we either lost this property, or never had it in the first place. We treat the `AttachPoint` not as a place where the program gets bound, but as a "compile context and **set** of attach points" for each program, and we possibly re-generate the probe for each "attach point" (hence the `add_probe(ap, probe)`). This is the reason you're needing to stuff the functions there, it's not an "attach point" at all! At best, it's a "spec" at the AST level, and we stuff on top of this spec a these-are-likely-similar "attach points" model (made more confusing by the fact that the spec is called `AttachPoint`, oh and we replicate the spec itself when doing expansion).

This is a very good point, I never thought about it that way. You're right, currently an AST `AttachPoint` is a specification of where the corresponding probe will attach to so there's not a 1-1 correspondence between AST probes and logical probes (i.e. BPF programs) but instead there's a 1-1 correspondence between an AST (probe + attach point) and logical probes. I'm not sure that model is inherently wrong, though. Maybe what makes it confusing here is that the top-level entity of a bpftrace program is called a *probe*. If it was called a *function* (or anything else), then we could say that (function + attach point) makes a probe and things would be much clearer, I believe.

> I'll stop ranting. You've been poking at this and know it better than me, so I certainly don't object to some intermediate state that stuffs a set of strings. I do feel like this is certainly an incremental improvement over having the logic in codegen. However, I don't know if there's a better way to workaround this issue unless we fix `AttachPoint` to actually _be_ an attach point for the associated program (that doesn't need to have that program compiled separately for each "attach point", which makes it a totally different thing. It also doesn't need to be materialized in that way in the AST itself, that's just an implementation detail,.. but conceptually you get the picture). BUT this gets into the long-term proposal in the other PR, i.e. fully expand everything, do all semantic analysis, do optimizations like the ksession rewrite, then dedup/merge all applicable cases (by comparing the programs, not complex logic around expansion types).

Obviously, one advantage with this model is that there may be thousands of functions to attach to so we'd initially end up with thousands of copies in the AST (until the dedup). I'm not sure what the performance consequences could be. But ideally, I hope one day kernel improves sufficiently enough such that it is possible to multi-attach to **all** kernel functions which would be some ~100000 functions.

> The last point on this (then rant done, I promise) is that for all this groups-of-similar-stuff-on-the-spec complexity, it doesn't even work that well. For example `kprobe:f {} kprobe:g {}` results in two separate identical programs could be trivially merged and multi-attached, but this isn't captured by our manual-globbing-expansion rules. Heck even `kprobe:f*, kprobe:g* {}` will generate two separate programs because the only multiplexing is on `AttachPoint`, not the probe.

Yeah, I realized the same problem. Once I'm done with this work, I wanted to follow with resolving this at least for multiple (potentially non-wildcarded) attach points on a single probe.

Maybe one last thought from my side. In practice, I don't see many cases where a user would write two different probes with completely same semantics. It may happen for some super-simple probes but still, I'd expect not to be very usual. So, instead of copying and de-duping AST probes, we could probably do some smart analysis of attach points. Something like expand and collect all attach points for each probe, cross-check kprobe and kretprobe probes to see if any can be merged via session probes, and then use multi-attach points in a way similar to what we use today.

src/ast/passes/probe_expansion.cpp
viktormalik: That's just an optimization to decide whether we need to do the complicated *split by expanded target* work or we can just go ahead and use `matches` directly. Since the latter is much more common, we can avoid running the below loop for a majority of cases.

src/ast/ast.h
viktormalik: So it turned out that propagating `ExpansionResult` via `PassManager` is a viable solution which is not even that complicated (see the now first commit in the PR). And it allows to drop all expansion-related fields from AST nodes.

@amscanne since the implementation is now quite different, I'm resetting your approval.

src/ast/passes/probe_expansion.cpp
```diff
@@ -306,9 +279,34 @@ void ProbeExpander::visit(AttachPointList &aps)
...
+            match_ap->func = ap->func;
+
+            auto new_ap = new_aps_by_target.emplace(match_ap->target, match_ap);
+            result_.add_expanded_func(*new_ap.first->second,
+                                      match_ap->target + ":" + expanded_func);
+          }
+          for (auto &[_, new_ap] : new_aps_by_target)
+            new_aps.push_back(std::move(new_ap));
+        } else if (!matches.empty()) {
+          result_.set_expanded_funcs(*ap, matches);
```
jordalgo: This is a long, non-blocking nit.

It seems like we can move `matches` into `set_expanded_funcs`.
Also it seems like `get_expanded_funcs` can just return a set instead of a vector as the `funcs` field in the `struct Probe` should ultimately be a set of strings instead of vector (since it seems like they are de-duped anyway). It can also be an unordered set (I think).

tests/pid_filter_pass.cpp
```diff
@@ -62,7 +64,7 @@ TEST(pid_filter_pass, add_filter)
     "fentry:f",
     "fexit:f",
     "tracepoint:category:event",
-    "rawtracepoint:event",
```
jordalgo, line 65: Not sure if I understand the need for this change.

jordalgo (commented):
Overall code looks good to me. I bet you'll want to wait for @amscanne to review anyway.

As a side thought, can we merge `ExpansionType::FULL` and `ExpansionType::NONE`? I mention this because even if a user doesn't include a wildcard in their module or function we still end up doing a full lookup for the attach point anyway e.g.
```
$ sudo bpftrace -e 'kprobe:tomato { print(1); }'
stdin:1:1-14: WARNING: tomato is not traceable (either non-existing, inlined, or marked as "notrace"); attaching to it will likely fail
kprobe:tomato { print(1); }
```
Are there cases where this lookup might fail for something like `tomato` but still be able to attach?

src/ast/ast.h
amscanne: > In fact, I think I'll give it a try to emit the `ExpansionResult` from `ProbeExpansion` and then use it from codegen (which should be possible as both are managed by `PassManager`). Then, we'll just pass the list of functions to `BPFtrace::add_probe` from `CodegenLLVM::add_probe`.
> 
> > My naive assumption/intuition is that post-"expansion", you would have probes with a set of attach points that are the places where you intend to attach the associated program (may still be symbolic, e.g. a USDT target, but no longer have wildcards, etc.). Furthermore: past the expansion point, I would assume that programs need only be compiled once (not compiled per attach point), and you could use multi-attach where available for the set of attach points associated with a given program. (So any merge/dedup pass would need to preserve this property and merge two probes into one only when the programs were semantically the same.)
> > However, I think that we either lost this property, or never had it in the first place. We treat the `AttachPoint` not as a place where the program gets bound, but as a "compile context and **set** of attach points" for each program, and we possibly re-generate the probe for each "attach point" (hence the `add_probe(ap, probe)`). This is the reason you're needing to stuff the functions there, it's not an "attach point" at all! At best, it's a "spec" at the AST level, and we stuff on top of this spec a these-are-likely-similar "attach points" model (made more confusing by the fact that the spec is called `AttachPoint`, oh and we replicate the spec itself when doing expansion).
> 
> This is a very good point, I never thought about it that way. You're right, currently an AST `AttachPoint` is a specification of where the corresponding probe will attach to so there's not a 1-1 correspondence between AST probes and logical probes (i.e. BPF programs) but instead there's a 1-1 correspondence between an AST (probe + attach point) and logical probes. I'm not sure that model is inherently wrong, though. Maybe what makes it confusing here is that the top-level entity of a bpftrace program is called a _probe_. If it was called a _function_ (or anything else), then we could say that (function + attach point) makes a probe and things would be much clearer, I believe.

Yeah, there's nothing inherent wrong here, it's just confusing. Name changes could help, but I'm also not so convinced that it's the best model. It does mostly work, and that is the most important thing so while worthy of a rant, I'm not arguing that this is a P0.

> Obviously, one advantage with this model is that there may be thousands of functions to attach to so we'd initially end up with thousands of copies in the AST (until the dedup). I'm not sure what the performance consequences could be. But ideally, I hope one day kernel improves sufficiently enough such that it is possible to multi-attach to **all** kernel functions which would be some ~100000 functions.

My opinion is that this is an unhelpful thing to optimize for. If you are attaching to tens of thousands of functions, and the compile takes O(10s) of seconds as a result,... okay? If we simplify things, we should also be able to make our various analyses and passes faster (i.e. semantic analysis also currently needs to be aware of all the attach points and does things like iterating over all possible types, which disappears). I'd rather optimize for simplicity of the model and compilation, because that lets us make the best optimizations and user experience (e.g. like the `ksessionprobe` merge, which IMO should happen always with all `kretprobe`s and work every time without needing to reason about globs or attachpoints :))
 
> Yeah, I realized the same problem. Once I'm done with this work, I wanted to follow with resolving this at least for multiple (potentially non-wildcarded) attach points on a single probe.
> 
> Maybe one last thought from my side. In practice, I don't see many cases where a user would write two different probes with completely same semantics. It may happen for some super-simple probes but still, I'd expect not to be very usual. So, instead of copying and de-duping AST probes, we could probably do some smart analysis of attach points. Something like expand and collect all attach points for each probe, cross-check kprobe and kretprobe probes to see if any can be merged via session probes, and then use multi-attach points in a way similar to what we use today.

The two different probes with same semantics case isn't optimized for that specifically, it's just a side effect of solving the general problem. Right now we're implicitly using a bunch of logic and analysis to determine which probes may or may not be able to be equal (they are from the same globbed target, whether they use args or other builtins, etc.). The proposal that I had was to just expand everything, then just merge all the programs that are the same. Very similar to how a classic block-level dedup or CAS might work; we don't care *how* those programs got there, if they are identical than they can be attached together. The primary benefit is simplification: we don't care how it was derived, we can just observe that it is the same and then merge.

> So it turned out that propagating ExpansionResult via PassManager is a viable solution which is not even that complicated (see the now first commit in the PR). And it allows to drop all expansion-related fields from AST nodes.

SGTM, I'll take a pass.

src/ast/passes/probe_expansion.cpp
amscanne: Dumb question: isn't it wrong since a C++ function can be overloaded and require the split even without a wildcard?

(Please don't change this in the context of the change, since it is the existing behavior, I just don't understand why we need to split only in the case of a target wildcard).

src/ast/passes/probe_expansion.cpp
viktormalik: > It seems like we can move `matches` into `set_expanded_funcs`.

Correct, will do.

> Also it seems like `get_expanded_funcs` can just return a set instead of a vector as the `funcs` field in the `struct Probe` should ultimately be a set of strings instead of vector (since it seems like they are de-duped anyway).

Yeah, good point, changing it to `std::set` in `Probe` should work. Let me give it a try.

> It can also be an unordered set (I think).

I think it's better to have the functions alphabetically ordered. Both for the purpose of reporting and testing.

tests/pid_filter_pass.cpp
viktormalik: I introduced a new behavior - when an expanded attach point has no matches, it is removed in `ProbeExpansion`. And since `rawtracepoint:event` is now expanded to `rawtracepoint:*:event` in `AttachPointParser`, the `ProbeExpansion` pass would essentially remove the attach point and the test would fail. So, we need the test to use a form which doesn't expand so that the attach point is not dropped.

viktormalik:
> As a side thought, can we merge `ExpansionType::FULL` and `ExpansionType::NONE`?

In the current design, we can't. As mentioned in one of the answers above, we now remove wildcarded attach points with no matches in `ProbeExpansion`. OTOH, for non-wildcarded attach points, we want the attachment to explicitly fail.

IMHO this behavior is correct - when a user uses a wildcard and there are no matches, we don't even try to attach (we'd fail with `No probes to attach`) but when a user specifies an exact attach point, we try to attach and fail with an error message specifying that exact attach point.

> I mention this because even if a user doesn't include a wildcard in their module or function we still end up doing a full lookup for the attach point anyway e.g.
> 
> ```
> $ sudo bpftrace -e 'kprobe:tomato { print(1); }'
> stdin:1:1-14: WARNING: tomato is not traceable (either non-existing, inlined, or marked as "notrace"); attaching to it will likely fail
> kprobe:tomato { print(1); }
> ```

Yeah, this is something we do for kprobes specifically because the generic error message for failed attachment was not very nice. IIRC, for other probe types, we do something similar in other places, maybe we could unify it somehow in a follow-up PR.
 
> Are there cases where this lookup might fail for something like `tomato` but still be able to attach?

AFAIK no, when a function is not in `available_filter_functions` (that's what the above check does), it cannot be attached to.

src/ast/ast.h
viktormalik: > Yeah, there's nothing inherent wrong here, it's just confusing. Name changes could help, but I'm also not so convinced that it's the best model. It does mostly work, and that is the most important thing so while worthy of a rant, I'm not arguing that this is a P0.

Likewise, I'm not arguing that the current model is the best that we can come up with.

In fact, your arguments are very good and I'm starting to lean towards your proposal. Let's discuss it more in the next office hours.

src/ast/passes/probe_expansion.cpp
viktormalik: Oh but `target` is the traced binary, not the function. So, I believe that it has to be wildcarded in order for expansion to take place.

src/ast/passes/probe_expansion.cpp
viktormalik: Changed everything to `std::set`.

jordalgo:
> IMHO this behavior is correct - when a user uses a wildcard and there are no matches, we don't even try to attach (we'd fail with No probes to attach) but when a user specifies an exact attach point, we try to attach and fail with an error message specifying that exact attach point.

Yeah this is a good point I hadn't considered. Suggestion to remove `ExpansionType::FULL` cheerfully withdrawn 😄 

> Yeah, this is something we do for kprobes specifically because the generic error message for failed attachment was not very nice. IIRC, for other probe types, we do something similar in other places, maybe we could unify it somehow in a follow-up PR.

+1 - and/or make an issue for someone else to pick up.

viktormalik:
> > Yeah, this is something we do for kprobes specifically because the generic error message for failed attachment was not very nice. IIRC, for other probe types, we do something similar in other places, maybe we could unify it somehow in a follow-up PR.
> 
> +1 - and/or make an issue for someone else to pick up.

Ack, I'll try to figure out what can be done and will open an issue.

jordalgo (approved):

PR #4382, opened 2025-07-24
Fix up Binop checks considering new bool type

jordalgo:
This fixes two issues (that were a bit coupled):
1.Treat bool types similar to int types when validating binops

These are now valid
```
$a = true;
$b = false;
$c = $a && $b;
$d = $a > $b;
$e = $a >> 1;
$f = $a + $b;
```

This wasn't caught due to lack of test coverage
and too much reliance on fold_literals code
which, obviously, doesn't fold variables.

2.Make the result of all comparison binops of type Bool

The previous type was an integer but this is incorrect, e.g., these should all evaluate to a boolean.
```
$a = 1 && 2;
$b = 1 > 2;
$c = 1 != 2;
$d = true != false;
```

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2245,7 +2245,8 @@ ScopedExpr CodegenLLVM::binop_int(Binop &binop)
   bool do_signed = lsign && rsign;
 
   // Promote operands if necessary
-  auto size = binop.result_type.GetSize();
+  auto size = std::max(binop.left.type().GetSize(),
+                       binop.right.type().GetSize());
   lhs = b_.CreateIntCast(lhs, b_.getIntNTy(size * 8), lsign);
   rhs = b_.CreateIntCast(rhs, b_.getIntNTy(size * 8), rsign);
 
@@ -3742,8 +3743,10 @@ ScopedExpr CodegenLLVM::getMultiMapKey(Map &map,
 
 ScopedExpr CodegenLLVM::createLogicalAnd(Binop &binop)
 {
-  assert(binop.left.type().IsIntTy() || binop.left.type().IsPtrTy());
-  assert(binop.right.type().IsIntTy() || binop.right.type().IsPtrTy());
+  assert(binop.left.type().IsIntTy() || binop.left.type().IsPtrTy() ||
+         binop.left.type().IsBoolTy());
+  assert(binop.right.type().IsIntTy() || binop.right.type().IsPtrTy() ||
+         binop.right.type().IsBoolTy());
 
   llvm::Function *parent = b_.GetInsertBlock()->getParent();
   BasicBlock *lhs_true_block = BasicBlock::Create(module_->getContext(),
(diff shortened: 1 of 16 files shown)


src/ast/passes/codegen_llvm.cpp
```diff
@@ -3770,8 +3770,10 @@ ScopedExpr CodegenLLVM::getMultiMapKey(Map &map,
 
 ScopedExpr CodegenLLVM::createLogicalAnd(Binop &binop)
 {
-  assert(binop.left.type().IsIntTy() || binop.left.type().IsPtrTy());
-  assert(binop.right.type().IsIntTy() || binop.right.type().IsPtrTy());
+  assert(binop.left.type().IsIntTy() || binop.left.type().IsPtrTy() ||
```
amscanne, line 3746: Something I've thought about a few times is the fact that there's a lot of implicit casting in codegen and handling those cases. It feels like semantic analyzer should inject explicit casts where it allows them explicitly, and maybe we could get rid of some paths. (I don't think this applies to logical and, which just compares against the zero value for the input type, but I'm just releasing this general thought into the world.)

jordalgo: Yeah, I agree that could be some good yak shaving and improvement to the codebase. Want to make an issue?

amscanne: Hm, not sure that it's a well-formed thought yet. I'll file it soon :)

src/ast/passes/fold_literals.cpp
```diff
@@ -65,23 +65,42 @@ static Expression make_boolean(ASTContext &ast, T left, T right, Binop &op)
...
+      value = left + right;
+      break;
     case Operator::MINUS:
+      value = left - right;
+      break;
     case Operator::MUL:
+      value = left && right;
+      break;
     case Operator::DIV:
+      value = static_cast<bool>(left && right);
```
amscanne, line 82: This is just for literal folding, so the value will always be known. For both `DIV` and `MOD`, I think you need to surface as an error on the binop. (You can probably just leave the value here but add the error to the node passed in as a parameter if `!right`).

tests/semantic_analyser.cpp
```diff
@@ -3121,23 +3121,49 @@ TEST_F(SemanticAnalyserTest, cast_sign)
...
 
-TEST_F(SemanticAnalyserTest, binop_sign)
+TEST_F(SemanticAnalyserTest, binop_bool_and_int)
+{
+  std::string operators[] = {
+    "==", "!=", "<", "<=", ">",  ">=", "&&", "||", "+",
+    "-",  "/",  "*", ">>", "<<", "&",  "|",  "^",
+  };
+
+  // Making a single variable below so as not to trigger fold_literals code
```
amscanne, line 3131: Can you add tests to the fold literals suite to ensure that bool is covered? (Specifically it would be good to make sure the overflow/underflow cases are handled.)

amscanne (approved):
Mostly lightweight review, just a couple things to be fixed

tests/semantic_analyser.cpp
jordalgo: Done.

src/ast/passes/fold_literals.cpp
jordalgo: Done.

PR #4383, opened 2025-07-25
Fix bad modulo optimization

jordalgo:
Some compilers will assume that the RHS of a modulo operation can never be null/0 as that is undefined behavior so it will
remove the null check on a map lookup, which the BPF verifier does not like.

This PR has three commits:

1. Add a a zero check for the RHS of a modulo, which means the compiler can't optimize away the null ref check as it's no longer going straight to the modulo operation.

2. Create a new runtime error for divide by zero and rename HelperError to RuntimeError

3. Add a zero check for the RHS of a divide operation. Wasn't seeing the same compiler issue but good have the same behavior as modulo.

Issue: https://github.com/bpftrace/bpftrace/issues/4379

##### Checklist

### DIFF
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2317,6 +2317,7 @@ ScopedExpr CodegenLLVM::binop_int(Binop &binop)
       b_.CreateCondBr(cond, is_zero, not_zero);
       b_.SetInsertPoint(is_zero);
       b_.CreateStore(b_.getInt64(1), mod_result);
+      b_.CreateRuntimeError(RuntimeErrorId::DIVIDE_BY_ZERO, binop.loc);
       b_.CreateBr(zero_merge);
 
       b_.SetInsertPoint(not_zero);
@@ -2324,10 +2325,9 @@ ScopedExpr CodegenLLVM::binop_int(Binop &binop)
       b_.CreateBr(zero_merge);
 
       b_.SetInsertPoint(zero_merge);
-      return ScopedExpr(b_.CreateLoad(b_.getInt64Ty(), mod_result),
-                        [this, mod_result, d = std::move(del)] {
-                          b_.CreateLifetimeEnd(mod_result);
-                        });
+      auto *result = b_.CreateLoad(b_.getInt64Ty(), mod_result);
+      b_.CreateLifetimeEnd(mod_result);
+      return ScopedExpr(result, std::move(del));
     }
     case Operator::BAND:
       return ScopedExpr(b_.CreateAnd(lhs, rhs), std::move(del));
--- a/src/ast/passes/codegen_llvm.cpp
+++ b/src/ast/passes/codegen_llvm.cpp
@@ -2283,51 +2283,12 @@ ScopedExpr CodegenLLVM::binop_int(Binop &binop)
     case Operator::MUL:
       return ScopedExpr(b_.CreateMul(lhs, rhs), std::move(del));
     case Operator::DIV:
-      return ScopedExpr(b_.CreateUDiv(lhs, rhs), std::move(del));
     case Operator::MOD: {
       // Always do an unsigned modulo operation here even if `do_signed`
       // is true. bpf instruction set does not support signed division.
       // We already warn in the semantic analyser that signed modulo can
       // lead to undefined behavior (because we will treat it as unsigned).
-      // Additionally, we need to do a 0 check or else a compiler optimization
-      // will assume that the value can't ever be null and remove the null check
-      // which is problematic if the right side of the modulo is a map value
-      // e.g. (10 % @my_map[1]) - From Google's AI: "LLVM, like other optimizing
-      // compilers, is allowed to make assumptions based on the absence of
-      // undefined behavior. If a program's code, after optimization, would
-      // result in undefined behavior (like division by zero by CreateURem), the
-      // compiler is free to make transformations that assume such a situation
-      // will never occur."
-      AllocaInst *mod_result = b_.CreateAllocaBPF(b_.getInt64Ty(),
-                                                  "mod_result");
-
-      llvm::Function *parent = b_.GetInsertBlock()->getParent();
-      BasicBlock *is_zero = BasicBlock::Create(module_->getContext(),
-                                               "is_zero",
-                                               parent);
-      BasicBlock *not_zero = BasicBlock::Create(module_->getContext(),
-                                                "not_zero",
-                                                parent);
-      BasicBlock *zero_merge = BasicBlock::Create(module_->getContext(),
-                                                  "zero_merge",
-                                                  parent);
-
-      Value *cond = b_.CreateICmpEQ(rhs, b_.getInt64(0), "zero_cond");
-
-      b_.CreateCondBr(cond, is_zero, not_zero);
-      b_.SetInsertPoint(is_zero);
-      b_.CreateStore(b_.getInt64(1), mod_result);
-      b_.CreateRuntimeError(RuntimeErrorId::DIVIDE_BY_ZERO, binop.loc);
-      b_.CreateBr(zero_merge);
-
-      b_.SetInsertPoint(not_zero);
-      b_.CreateStore(b_.CreateURem(lhs, rhs), mod_result);
-      b_.CreateBr(zero_merge);
-
-      b_.SetInsertPoint(zero_merge);
-      auto *result = b_.CreateLoad(b_.getInt64Ty(), mod_result);
-      b_.CreateLifetimeEnd(mod_result);
-      return ScopedExpr(result, std::move(del));
+      return ScopedExpr(b_.CreateCheckedBinop(binop, lhs, rhs), std::move(del));
     }
     case Operator::BAND:
       return ScopedExpr(b_.CreateAnd(lhs, rhs), std::move(del));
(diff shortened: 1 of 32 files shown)


jordalgo:
If this looks good, I'll add a check for the divide operator.

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2282,12 +2282,52 @@ ScopedExpr CodegenLLVM::binop_int(Binop &binop)
...
     case Operator::DIV:
       return ScopedExpr(b_.CreateUDiv(lhs, rhs), std::move(del));
-    case Operator::MOD:
+    case Operator::MOD: {
       // Always do an unsigned modulo operation here even if `do_signed`
       // is true. bpf instruction set does not support signed division.
       // We already warn in the semantic analyser that signed modulo can
       // lead to undefined behavior (because we will treat it as unsigned).
-      return ScopedExpr(b_.CreateURem(lhs, rhs), std::move(del));
+      // Additionally, we need to do a 0 check or else a compiler optimization
```
amscanne: You should push this down to IRbuilder, because the same behavior has to apply to DIV. (Technically signed integer underflow/overflow is also UB, but not sure we want to bother with that.)

For example, I'm expecting it here and in DIV to be pretty simple just operating on the value, e.g.:

```
return ScopedExpr(b_.CreateURem(lhs, b_.EnsureNonZero(rhs)), std::move(del));
```

src/ast/passes/codegen_llvm.cpp
```diff
@@ -2282,12 +2282,52 @@ ScopedExpr CodegenLLVM::binop_int(Binop &binop)
...
+      // will assume that the value can't ever be null and remove the null check
+      // which is problematic if the right side of the modulo is a map value
+      // e.g. (10 % @my_map[1]) - From Google's AI: "LLVM, like other optimizing
+      // compilers, is allowed to make assumptions based on the absence of
+      // undefined behavior. If a program's code, after optimization, would
+      // result in undefined behavior (like division by zero by CreateURem), the
+      // compiler is free to make transformations that assume such a situation
+      // will never occur."
+      AllocaInst *mod_result = b_.CreateAllocaBPF(b_.getInt64Ty(),
```
amscanne: This is more complicated than it should be. You don't want an allocation for this value, you just want a `Phi` node that joins the two branches and gives you either the value, or the constant value `1` (with the Ub Error logged to the ring buffer)

For example, see the ternary case for integers: https://github.com/bpftrace/bpftrace/blob/master/src/ast/passes/codegen_llvm.cpp#L2542

src/output.cpp
```diff
@@ -247,6 +247,23 @@ std::string Output::get_helper_error_msg(libbpf::bpf_func_id func_id,
   return msg;
 }
 
+std::string Output::get_ub_error_msg(UbErrorId ub_id) const
+{
+  std::string msg;
+  switch (ub_id) {
+    case UbErrorId::DIVIDE_BY_ZERO: {
+      msg = "Divide or modulo by 0 detected. This can lead to unexpected "
+            "results.";
```
amscanne: Note that `1` has been used instead, and that the user should handle this case explicitly?

src/output.cpp
```diff
@@ -247,6 +247,23 @@ std::string Output::get_helper_error_msg(libbpf::bpf_func_id func_id,
...
+      break;
+    }
+    case UbErrorId::INVALID: {
```
amscanne: Why define `INVALID`? I think you only want sentinels if there's a concrete reason, e.g. you want a sensible value for zero memory, you need the know the specific number of enums, etc.

src/required_resources.h
```diff
@@ -51,7 +48,50 @@ class HelperErrorInfo {
   template <typename Archive>
   void serialize(Archive &archive)
   {
-    archive(func_id, filename, line, column, source_location, source_context);
+    archive(filename, line, column, source_location, source_context);
+  }
+};
+
+class HelperErrorInfo : public RuntimeErrorInfo {
```
amscanne: The split into `HelperErrorInfo` and `RuntimeErrorInfo` is good, I'm just wondering if we need it? Maybe you want to just generalized over `RuntimeError` and define specific enums for the various helpers rather than using the `bpf_func_id` directly? I can't imagine we'll have too many distinct categories, so you're just forcing an enumeration earlier rather than supporting `RuntimeError` for arbitrary helpers (which we don't need anyways). WDYT?

src/ast/passes/codegen_llvm.cpp
jordalgo: Ack

src/ast/passes/codegen_llvm.cpp
jordalgo: Ack.

jordalgo: @amscanne I could be wrong (still looking) but I don't think I can call out to `b_.CreateUbError` in a phi node branch. I get this error:
```
ERROR: PHI node entries do not match predecessors
```

with this code:
```
llvm::Function *parent = b_.GetInsertBlock()->getParent();
BasicBlock *is_zero = BasicBlock::Create(module_->getContext(),
                                         "is_zero",
                                         parent);
BasicBlock *not_zero = BasicBlock::Create(module_->getContext(),
                                          "not_zero",
                                          parent);
BasicBlock *zero_merge = BasicBlock::Create(module_->getContext(),
                                            "zero_merge",
                                            parent);

Value *cond = b_.CreateICmpEQ(rhs, b_.getInt64(0), "zero_cond");

b_.CreateCondBr(cond, is_zero, not_zero);
b_.SetInsertPoint(is_zero);
b_.CreateUbError(UbErrorId::DIVIDE_BY_ZERO, binop.loc);
auto *is_zero_expr = b_.getInt64(1);
b_.CreateBr(zero_merge);

b_.SetInsertPoint(not_zero);
auto *mod_expr = b_.CreateURem(lhs, rhs);
b_.CreateBr(zero_merge);

b_.SetInsertPoint(zero_merge);
auto *phi = b_.CreatePHI(b_.getInt64Ty(), 2, "result");
phi->addIncoming(is_zero_expr, is_zero);
phi->addIncoming(mod_expr, not_zero);

return ScopedExpr(phi, std::move(del));
 ```

Edit: Ok, the issue is because `CreateRingbufOutput` creates additional blocks for conditionally updating the event loss counter, which I guess the compiler is not happy with (but I don't yet understand why).

Comment (amscanne) [reply src/ast/passes/codegen_llvm.cpp]: If you split the Phi itself into a simpler block, can you emit the error via a separate conditional?

Note that using `1` for the value here also seems to be a bit wrong, because I actually think you want `1` for the `rhs`, always producing a value of `0` after the operation. (It's all undefined, but having n % 0 => 0 seems at least less likely to trigger surprising paths)

Comment (jordalgo) [reply src/ast/passes/codegen_llvm.cpp]: > If you split the Phi itself into a simpler block, can you emit the error via a separate conditional?

Why would we want to do that? That's two separate conditionals with two sets of blocks in order to preserve a phi node and save a stack allocation?

> (It's all undefined, but having n % 0 => 0 seems at least less likely to trigger surprising paths)

Ehh, not sure I agree. For the issue where this was surfaced, it was waiting for the mod operation to equal 0 before performing the operation, which I think is common. If we return 1 that is usually the failure case.

Comment (amscanne) [reply src/ast/passes/codegen_llvm.cpp]: > Why would we want to do that? That's two separate conditionals with two sets of blocks in order to preserve a phi node and save a stack allocation?

My assumption would be that the compiler is better able to optimize around a PHI and blocks using the same value (i.e. side-effect-free control flow) than it is to optimize around extraneous loads and stores. If you're feeling adventurous, you could implement both and see what gets produced? It will probably do both in this case, so perhaps whatever is logically simplest is fine (but this is why I prevent not to have the allocation). If you do keep the allocation, I don't think you need to extend the lifetime past the actual load.

> Ehh, not sure I agree. For the issue where this was surfaced, it was waiting for the mod operation to equal 0 before performing the operation, which I think is common. If we return 1 that is usually the failure case.

LOL, good point -- we could also just return `0xBAAAAAAD`. That's the most likely to raise attention. My reasoning was that the limit of `x mod y` as `y -> 0` for integers is `0`, but all reasoning is equally invalid.

Comment (jordalgo) [reply src/ast/passes/codegen_llvm.cpp]: The optimized codegen looks equivalent to me (even with the stack allocation, a phi node is created) - let me know if I'm missing something.

- [With stack allocation](https://gist.github.com/jordalgo/e01d137b717b63cb68a135ca50054f30)
- [With explicit phi node and two conditionals](https://gist.github.com/jordalgo/96011a472c916efa52bd34e3f918aca0)

File: src/required_resources.h (hunk 5 above)
Comment (jordalgo) [reply src/required_resources.h]: Sounds good. Made RuntimeError the thing that both helper errors and divide-by-zero errors use.

File: src/output/json.cpp [hunk 6]
```diff
@@ -434,17 +434,31 @@ void JsonOutput::attached_probes(uint64_t num_probes)
...
+      JsonEmitter<std::string>::emit(out_, ss.str());
+      out_ << R"(, "retcode": )";
+      JsonEmitter<int64_t>::emit(out_, retcode);
+      break;
+    }
+    case RuntimeErrorId::DIVIDE_BY_ZERO: {
+      out_ << R"({"type": "runtime_error")";
+      out_ << R"(, "msg": )";
+      std::stringstream ss;
+      ss << DIVIDE_BY_ZERO_MSG;
```
amscanne: Why not have `RuntimeErrorInfo` implement an `operator<<(std::ostream&)` and then it could be responsible for serializing this message? That would allow you to move the `constexpr` string out of the header and make it an implementation detail for most cases.

(I think you'd still need to preserve the special case handling for the helper errors, but at least then we could add new Runtime errors without modifying the output handlers at all.)

jordalgo: Done.

PR #4385, opened 2025-07-25
build: use memfd instead of pipes for `ClangBuild`

amscanne:
Stacked PRs:
 * __->__#4385

--- --- ---

### build: use memfd instead of pipes for `ClangBuild`

For simplicity, we construct a file to pass through the clang argument
layers (which is later reopened) that does not correspond to any on-disk
file. Currently this is a pipe. However, the pipe can fill out with
output in some cases, which results in the program blocking. Replace the
pipe with a memfd to simplify this logic and eliminate that case.

### DIFF
--- a/src/ast/passes/clang_build.cpp
+++ b/src/ast/passes/clang_build.cpp
@@ -12,6 +12,7 @@
 #include <llvm/Support/raw_ostream.h>
 #include <llvm/TargetParser/Host.h>
 #include <sstream>
+#include <sys/mman.h>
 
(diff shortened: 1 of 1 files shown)


jordalgo:
LGTM - one small nit on the commit message: " that does correspond to any on-disk file" -> "does NOT"

jordalgo (approved):

amscanne:
> LGTM - one small nit on the commit message: " that does correspond to any on-disk file" -> "does NOT"

Fixed.

PR #4386, opened 2025-07-26
fn: the context is no longer needed

amscanne:
fn: the context is no longer needed

And we can use external linkage for functions.


amscanne:
I will fix this in #4791.

PR #4390, opened 2025-07-26
benchmark: Prevent duplicate `BENCH` probe names

jrife:
Currently, `BENCH` probe names are allowed to be reused, but the benchmark results only show one row per unique name:

```
BENCH:my_bench {
    @a = 1;
}

BENCH:my_bench {
    @b = 1;
}
```

```
Attached 1 probe

+-----------+--------------+
| BENCHMARK | AVERAGE TIME |
+-----------+--------------+
| my_bench  | 155ns        |
+-----------+--------------+
```

This is incorrect and confusing. To ensure consistency between the script and the output, we have two options:

1. Continue to allow duplicate `BENCH` probe names and teach the output logic how to display one row per `BENCH` probe instead of one row per unique name.
2. Prevent users from creating two `BENCH` probes with the same name.

Since there doesn't seem to be much legitimate reason to allow for duplicate names, go with option two and check for duplicate names in semantic analysis.

```
script.bt:3-5: ERROR: "my_bench" was used as the name for more than one BENCH probe
}

BENCH:my_bench {
script.bt:1:1-15: ERROR: this is the other instance
BENCH:my_bench {
~~~~~~~~~~~~~~
```

##### Checklist

- [x] Language changes are updated in `docs/language.md`, `docs/stdlib.md`, or `man/adoc/bpftrace.adoc`

### DIFF
--- a/src/ast/passes/semantic_analyser.cpp
+++ b/src/ast/passes/semantic_analyser.cpp
@@ -264,6 +264,7 @@ private:
   bool has_begin_probe_ = false;
   bool has_end_probe_ = false;
   bool has_child_ = false;
+  std::unordered_map<std::string, Location> benchmark_locs_;
 };
 
 } // namespace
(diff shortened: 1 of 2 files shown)


src/ast/passes/semantic_analyser.cpp
```diff
@@ -3816,6 +3817,13 @@ void SemanticAnalyser::visit(AttachPoint &ap)
   } else if (ap.provider == "BENCH") {
     if (ap.target.empty())
       ap.addError() << "BENCH probes must have a name";
+    if (is_final_pass()) {
+      if (has_benchmark_.contains(ap.target))
+        ap.addError() << "\"" + ap.target + "\""
```
amscanne: You can make the `std::unordered_set` above a `std::map<std::string, Location>`. Then here, do something like (supposing it is called `benchmark_locs_`):

```
auto it = benchmark_locs_.find(ap.target);
if (it != benchmark_locs_.end()) {
  auto &err = ap.addError();
  err << ... (original error);
  err.addContext(it->second) << "this is the other instance";
}
benchmark_locs_.emplace(ap.target, ap.loc);
```

That will give a message that includes both locations.

amscanne (approved):
LGTM, I have just suggested a way to include the other location in the error, for more clarity

src/ast/passes/semantic_analyser.cpp
jrife: Done. Updated the PR description and commit message with example output.

jordalgo (approved):
LGTM. Thanks for the fix!

PR #4392, opened 2025-07-27
codegen: use unique name for internal map type

amscanne:
Stacked PRs:
 * #4403
 * #4401
 * #4397
 * #4402
 * #4393
 * #4395
 * #4399
 * __->__#4392

--- --- ---

### codegen: use unique name for internal map type

The name `map_t` conflicts with types in the kernel, so it gets de-dup
in an ugly way (and potentially unstable, since it depends on the kernel
types). This doesn't have any functional effect, but may make the
codegen types more fragile. Just use a longer name for this type to
avoid this collision.


src/ast/passes/codegen_llvm.cpp
```diff
@@ -4392,7 +4392,7 @@ void CodegenLLVM::createMapDefinition(const std::string &name,
     elems.push_back(b_.getPtrTy());
     elems.push_back(b_.getPtrTy());
   }
-  auto *type = StructType::create(elems, "struct map_t", false);
+  auto *type = StructType::create(elems, "struct map_interal_repr_t", false);
```
jordalgo: I know it doesn't matter much but why not just call it `bpftrace_map_t`?

jordalgo (approved):

src/ast/passes/codegen_llvm.cpp
amscanne: Meh, cause this is the map type required for BTF? It's not bpftrace-specific. I can use any name, I just decided "internal representation" was sufficiently vague?

jordalgo: > It's not bpftrace-specific.

They're the maps that bpftrace creates right? Anyway, this is not at all blocking.

amscanne: Sure. But everything in that file is bpftrace-created. I just don't like using the name `bpftrace` in types, etc. since it is just a useless specifier. Imagine if the Linux kernel types were all like `linux_task_struct` ;)

src/ast/passes/codegen_llvm.cpp
```diff
@@ -4397,7 +4397,7 @@ void CodegenLLVM::createMapDefinition(const std::string &name,
...
```
viktormalik: Typo?
```suggestion
  auto *type = StructType::create(elems, "struct map_internal_repr_t", false);
```

amscanne: I had one job... fixing.

PR #4393, opened 2025-07-27
docs: Use `//` for generating stdlib docs

amscanne:
Stacked PRs:
 * #4397
 * #4402
 * __->__#4393

--- --- ---

### docs: Use `//` for generating stdlib docs

This allows for a simpler parser, since we don't have to take into
account any of the common C-style multi-line strings, e.g.

```
/*
 * Something.
 */
```

Further, this change adds support for a longer description in addition
to the table based summary. The summary is used in the table, and both
the summary and description are used in the detailed markdown section.

### DIFF
--- a/docs/stdlib.md
+++ b/docs/stdlib.md
@@ -1220,6 +1220,7 @@ Simple assertion macro that will exit the entire script with an error code if th
 
 #### Last Expression
 - **None**
+
 ### ppid
 Get the pid of the parent process
 
--- a/scripts/generate_stdlib_docs.py
+++ b/scripts/generate_stdlib_docs.py
@@ -7,22 +7,26 @@ import glob
 import os
 import sys
 import re
-from typing import NamedTuple, Optional
+from typing import Optional
 
-class Param(NamedTuple):
-    name: str
-    type: str
-    description: str
+class Param:
+    name: str = ""
+    type: str = "none"
+    description: str = ""
 
-class LastExpr(NamedTuple):
-    type: str
-    description: str
+class LastExpr:
+    type: str = "none"
+    description: str = ""
 
-class Macro(NamedTuple):
-    name: str
-    params: list[Param]
-    last_expr: LastExpr
-    description: str
+class Macro:
+    name: str = ""
+    params: list[Param] = None
+    last_expr: LastExpr = LastExpr()
+    summary: str = ""
+    description: str = ""
+
+    def __init__(self):
+      self.params = []
 
 def parse_param_string(param_string: str) -> Optional[Param]:
     """
@@ -33,11 +37,11 @@ def parse_param_string(param_string: str) -> Optional[Param]:
 
     match = re.match(pattern, param_string.strip())
     if match:
-        param_type = match.group(1).strip()
-        name = match.group(2).strip()
-        description = match.group(4).strip()
-
-        return Param(name=name, type=param_type,description=description)
+        param = Param()
+        param.type = match.group(1).strip()
+        param.name = match.group(2).strip()
+        param.description = match.group(4).strip()
+        return param
 
     return None
 
@@ -50,10 +54,10 @@ def parse_last_expr_string(param_string: str) -> Optional[LastExpr]:
 
     match = re.match(pattern, param_string.strip())
     if match:
-        expr_type = match.group(1).strip()
-        description = match.group(2).strip()
-
-        return LastExpr(type=expr_type, description=description)
+        last_expr = LastExpr()
+        last_expr.type = match.group(1).strip()
+        last_expr.description = match.group(2).strip()
+        return last_expr
 
     return None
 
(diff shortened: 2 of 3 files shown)


jordalgo (changes requested):
What about allowing both since both style of comments are supported in bpftrace?
Also I don't see how the `//` is a more condensed view, it adds more characters to each line and IMHO doesn't make anything more readable. `/*` is also preferred for multi-line comments.

> Further, this change adds support for a longer description in addition
to the table based summary.

Maybe I missed this but how?

amscanne:
> `/*` is also preferred for multi-line comments.

Are we advocating for this? I personally loathe the `/*` style of comments, since it creates all kinds of parsing ambiguities for any kind of external tooling (and IMO, creates immediate style ambiguities and looks generally worse). Your own parser here wasn't handling them at all requiring the `*/` to be on its own on a single line so they weren't actually supported at all!

Let me propose the following: let's support `///` as the doc format, which is what is used by Rust, Javadoc and Doxygen. It is very standardized. Then there's no mixing with regular comments either. Does this seem reasonable?

> > Further, this change adds support for a longer description in addition
> > to the table based summary.
> 
> Maybe I missed this but how?

The first line is the summary (which is the only thing included in the table), and everything from the first non-empty line onwards is the full description (both the summary and description are included in the full table).

amscanne:
BTW, for context: I didn't just do this for the esthetics. I did this change because I had to support the "summary" bits (because the `socket_cookie` docs were large, and embedded markdown and an example and didn't work in the table), and the parsing was already complex. So I simplified the parser by moving to the simpler comment model (the esthetics are just a fringe benefit), but I think that `///` makes sense as it is standardized for many other languages as well (and is already our more common comment form).

jordalgo:
I don't intend to die on this hill but choosing `//` or `/*` or `////////` seems like we're not really addressing the problem, which is just that the script needs to be changed (not the convention). Why don't we just be flexible about using either of our two comment options in bpftrace (even if it means that the script is a little more complicated) and either:
- add new prefixes for the two fields we want to target directly (e.g. `short_des` and  `long_des`)
OR
- update the script to do what docstrings does and treat the first line as the short description and anything that follows (if followed by an empty line) as the longer description (that would not go in the table).

> /// makes sense as it is standardized for many other languages as well (and is already our more common comment form).

I see it in some places around the code base but it definitely doesn't seem to be the most common (`//` is the clear winner from my search).

amscanne:
> I don't intend to die on this hill but choosing `//` or `/*` or `////////` seems like we're not really addressing the problem, which is just that the script needs to be changed (not the convention). Why don't we just be flexible about using either of our two comment options in bpftrace (even if it means that the script is a little more complicated) and either:
> 
> * add new prefixes for the two fields we want to target directly (e.g. `short_des` and  `long_des`)
>   OR
> * update the script to do what docstrings does and treat the first line as the short description and anything that follows (if followed by an empty line) as the longer description (that would not go in the table).

This is how it works. It just includes the first line in the longer section. Sort of like seeing a commit message summary and the full commit.

I'm confused about what you're asking for. The current version of this patch is not accepting any random comments, it is only accepting the comment strings that are in the special `///` form. Is it really worth correctly parsing `/*` just to avoid using `///` prefix for documentation? Who needs this flexibility? Regardless of the form, all of our stdlib documentation should all be written using one.

If the idea is that we should encode special parameters `:short_des` and `:long_des`, it seems rather silly. Why not just make this more natural? Going back to the commit message -- you don't need special metadata for the summary in that case either, and it's quite natural.

> > /// makes sense as it is standardized for many other languages as well (and is already our more common comment form).
> 
> I see it in some places around the code base but it definitely doesn't seem to be the most common (`//` is the clear winner from my search).

Right. I'm not saying that we change our accepted comment style to `///`, just that only `///` generates doc strings. (As is the case for Rust, Javadoc, Doxygen, etc.). And obviously `///` is a form of `//` comments, which is the more common form, so it seems more in place.

jordalgo:
> Is it really worth correctly parsing /* just to avoid using /// prefix for documentation? Who needs this flexibility? Regardless of the form, all of our stdlib documentation should all be written using one.

Would this cover it? What edge cases is this missing? https://github.com/bpftrace/bpftrace/pull/4400

amscanne:
> > Is it really worth correctly parsing /* just to avoid using /// prefix for documentation? Who needs this flexibility? Regardless of the form, all of our stdlib documentation should all be written using one.
> 
> Would this cover it? What edge cases is this missing? #4400

There are cases not covered by that:
- The comment can be ended and started multiple times on a single line, so you can have `/* foo */ xyz /* bar */`
- Your specific PR is including text past the end of the comment in the line itself, e.g. `/* foo */ macro x { }` would apply "foo macro x { }` as the summary for the **next** macro.

Is that you don't want a special syntax at all? While I'd rather support the doc standard `///`, I would say a middle ground is just supporting `//` until we build this support & generation into the lexer & parser (at some point)? Then we don't need the separate script that parses these cases. But supporting`/*` only in a specific layout *is* a special syntax (i.e. you must start `/*` at the beginning of the line or it doesn't work -- that seems less user friendly than requiring `///`).

jordalgo:
> that seems less user friendly than requiring

How so? The user now has to add more characters to every line in order to adhere to the proposed `///` syntax, which is also more visually noisy.

jordalgo:
> There are cases not covered by that

Yes, that's true. The user would have to make some level of effort to make their new comments look like the current ones in the same way they would have to remember to put a `///`, which is not even a "supported" comment syntax at all in bpftrace, or even a `//` at the start of every line.

amscanne:
> How so? The user now has to add more characters to every line in order to adhere to the proposed /// syntax, which is also more visually noisy.

It's less user friendly because you're requiring a non-standard comment format. No one writes `/*` comments like you have because it is lacks the visual cue that the line is a comment and related block. That's why people usually have:

```
/* 
 * Foo.
 *
 * Foo is a bar.
 */
```

The `*` is included to provide the visual cue on each line that is is part of the same block. (Although these can be solved using syntax highlighting and fancy editors, not everyone has these and most appreciated the fact that the extra text makes the block appear more coherent.)

Meanwhile the `///` syntax is also not the common form but it *is* a standard convention across some of the most popular languages and tools.

> > There are cases not covered by that
> 
> Yes, that's true. The user would have to make some level of effort to make their new comments look like the current ones in the same way they would have to remember to put a `///`, which is not even a "supported" comment syntax at all in bpftrace, or even a `//` at the start of every line.

What do you mean? `///` is just a `//` comment with an extra `/` which is absolutely supported. It's just a special way of marking some subset of comments as being used to general formal documentation.

jordalgo (approved):
Discussed offline - no stars - we're double slashing

PR #4395, opened 2025-07-27
interop: relax type checks in some cases

amscanne:
Stacked PRs:
 * #4403
 * #4401
 * #4397
 * #4402
 * #4393
 * __->__#4395
 * #4399
 * #4392

--- --- ---

### interop: relax type checks in some cases

Unfortunately `SizedType` is not able to represent many types: 32-bit
enums, unions, floating point, etc. Therefore it cannot be compatible
in many cases with the required argument. For example, it is not
possible to pass a `SizedType` which is equivalent to `struct sock*`,
only one that *appears* to be similar to one level. The BTF types are
fully structured and refuse to create a compat type.

In this case, we relax the constrain and allow the user to pass any
pointer type as long as a pointer type is expected. This is only done
when the compat type cannot be constructed, if it can, then we do
enforce that the types are the same.


jordalgo (commented):
ooof, this makes me a bit sad - is there a future here where we can actually do a type check on the pointee type?

amscanne:
> ooof, this makes me a bit sad - is there a future here where we can actually do a type check on the pointee type?

Of course, if our internal type system is using some variant of BTF then this relaxation can go away. Problem is that our internal type system can't represent the type.

viktormalik (approved):
Looks reasonable to me although I'm not very familiar with this part of code.

